PR #3156, opened 2024-11-15
Fix bugs in idle timeout logic (C#)

bernardnormier:
This PR fixes two small bugs in the idle timeout logic (in C# - will port to other languages in a follow-up PR).

PR #3159, opened 2024-11-16
Idle timeout fixes in C++/Java/JS

bernardnormier:
This PR "ports" the fixes from #3156 to C++, Java and JS.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/IdleTimeoutTransceiverDecorator.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/IdleTimeoutTransceiverDecorator.java
@@ -9,7 +9,7 @@ import java.util.concurrent.TimeUnit;
 // Decorates Transceiver to send heartbeats and optionally detect when no byte is received/read for
 // a while.
 // This decorator must not be applied on UDP connections.
-class IdleTimeoutTransceiverDecorator implements Transceiver {
+final class IdleTimeoutTransceiverDecorator implements Transceiver {
     private final Transceiver _decoratee;
     private final int _idleTimeout;
     private final ScheduledExecutorService _scheduledExecutorService;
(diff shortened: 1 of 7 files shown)


js/src/Ice/IdleTimeoutTransceiverDecorator.js
```diff
@@ -11,22 +11,20 @@ export class IdleTimeoutTransceiverDecorator {
         this._decoratee = decoratee;
         this._idleTimeout = idleTimeout * 1000; // Convert seconds to milliseconds
         this._timer = timer;
-        this._enableIdleCheck = enableIdleCheck;
         this._connection = connection;
+
+        // _idleCheckEnabled is initially enableIdleCheck (by default, true) unlike C++/C#/Java.
```
bernardnormier, line 16: The overall logic in JS is simpler/different because (i) there is no Holding state and (ii) the connection cannot be backpressured; that's why there is no enable/disable idleCheck methods in JS.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/IdleTimeoutTransceiverDecorator.java
```diff
@@ -126,24 +120,30 @@ public IdleTimeoutTransceiverDecorator(
...
-    public void disableIdleCheck() {
+    void disableIdleCheck() {
         if (_idleCheckEnabled && _idleCheck != null) {
             cancelReadTimer();
             _idleCheckEnabled = false;
         }
     }
 
+    void scheduleHeartbeat() {
+        // Reschedule because the connection establishment may have already written to the
```
bernardnormier, line 142: Keep in mind that a "write" can occur during connection establishment. This pre-Active write schedules a heartbeat, but this heartbeat will be ignored if we attempt to send it before the ice connection is Active.

pepone (approved):

js/src/Ice/IdleTimeoutTransceiverDecorator.js
```diff
@@ -11,22 +11,20 @@ export class IdleTimeoutTransceiverDecorator {
...
+        this._idleCheckEnabled = enableIdleCheck;
     }
 
     setCallbacks(connectedCallback, bytesAvailableCallback, bytesWrittenCallback) {
         this._decoratee.setCallbacks(connectedCallback, bytesAvailableCallback, bytesWrittenCallback);
     }
 
     initialize(readBuffer, writeBuffer) {
-        const op = this._decoratee.initialize(readBuffer, writeBuffer);
-        if (op == SocketOperation.None) {
```
pepone, line 24: We can remove the `SocketOperation` import, seems it is no longer used.

InsertCreativityHere (approved):

PR #3164, opened 2024-11-18
Fix JavaScript connection scheduling of heartbeats

pepone:
This PR fixes JavaScript connection to ensure hearbeats are scheduled even if the write completes asynchronously.

It also reworks how the connection checks if the transceiver write call consumed the completed buffer. There is no need for a callback or separate parameter as the transceiver advances the buffer and the call can check if there is any remaining data.

### DIFF
--- a/js/src/Ice/ConnectionI.js
+++ b/js/src/Ice/ConnectionI.js
@@ -415,14 +415,26 @@ export class ConnectionI {
         let info = null;
         let response = null;
         try {
-            if ((operation & SocketOperation.Write) !== 0 && this._writeStream.buffer.remaining > 0) {
-                DEV: console.assert(this._sendStreams.length > 0);
-                const currentMessage = this._sendStreams[0];
-                if (!this.write(this._writeStream.buffer, () => (currentMessage.isSent = true))) {
-                    DEV: console.assert(!this._writeStream.isEmpty());
-                    return;
+            if ((operation & SocketOperation.Write) !== 0) {
+                if (this._writeStream.buffer.remaining > 0) {
+                    DEV: console.assert(this._sendStreams.length > 0);
+                    const completedSynchronously = this.write(this._writeStream.buffer);
+                    if (this._writeStream.buffer.remaining === 0) {
+                        // If the write call consumed the complete buffer, we assume the message is sent now for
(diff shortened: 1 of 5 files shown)


js/src/Ice/ConnectionI.js
```diff
@@ -415,14 +415,26 @@ export class ConnectionI {
...
-                    return;
+            if ((operation & SocketOperation.Write) !== 0) {
+                if (this._writeStream.buffer.remaining > 0) {
+                    DEV: console.assert(this._sendStreams.length > 0);
+                    const completedSynchronously = this.write(this._writeStream.buffer);
+                    if (this._writeStream.buffer.remaining === 0) {
+                        // If the write call consumed the complete buffer, we assume the message is sent now for
+                        // at-most-once semantics.
+                        this._sendStreams[0].isSent = true;
+                    }
```
pepone, line 426: This replaces the previous callback passed to write.

js/src/Ice/ConnectionI.js
```diff
@@ -415,14 +415,26 @@ export class ConnectionI {
...
+
+                    if (!completedSynchronously) {
+                        DEV: console.assert(!this._writeStream.isEmpty());
+                        return;
+                    }
+                    DEV: console.assert(this._writeStream.buffer.remaining === 0);
+                } else if (this._transceiver instanceof IdleTimeoutTransceiverDecorator) {
+                    // The writing of the current message completed, schedule a heartbeat in case the connection has
+                    // nothing more to write.
+                    this._transceiver.scheduleHeartbeat();
```
pepone, line 436: This ensure that a heartbeat is scheduled even if write completes asynchronously. In this case Transceiver.write is not called and the decorator would not schedule one.

externl (approved):

bernardnormier (approved):
Looks good.

A key part of the logic here is the "write" callback set by ConnectionI using setCallbacks:
```js
  this._transceiver.setCallbacks(
                () => this.message(SocketOperation.Write), // connected callback
                () => this.message(SocketOperation.Read), // read callback
                () => this.message(SocketOperation.Write), // write callback
            );
``` 

This ensures `message(SocketOperation.Write)` is called when a write completes.

PR #3067, opened 2024-11-05
Add a General Purpose 'MetadataValidator' to the Slice Parser

InsertCreativityHere:
This PR lays the groundwork for implementing #117.

Currently, each language has it's own completely separate machinery for validating metadata, if it has any at all.
This is not a great setup.

To fix this, I added a new `MetadataValidator` class to the parser, which all the individual languages can easily re-use.
When you construct it, you pass in the language prefix to check, and a list of validation functions it can use.
It will automatically visit all the Slice definitions, and call the provided validation functions when appropriate.

These validation functions take the metadata to be validated, and a reference to what it's applied to.
They all return an `optional<string>`.`nullopt` indicates the validation succeeded. If a string is returned, we emit it to the user, and remove the offending metadata (filter it out).

----

Note that this PR doesn't actually _use_ this new class in the language compilers yet.
I first wanted to make sure that the initial concept was good and sounded okay to everyone.
If it is, then I'll start switching over the languages, a few at a time to try and keep the PRs from being too huge.

PR #3165, opened 2024-11-18
Centralize and Improve Metadata Validation in the Slice Parser

InsertCreativityHere:
This PR introduces all the heavy machinery discussed in #3067, which it is a replacement for.

It still accomplishes the same goal of centralizing the validation of Slice metadata into the Slice library (the parser), instead of making each language do it all from scratch.

But, whereas #3067 used inheritance and method overriding to do the validation, this PR uses a more declarative approach.
It adds a new `MetadataInfo` struct (to the `MetadataValidator` header file) which fully describes the information needed to validate a piece of metadata. Our metadata is _pretty flexible_, so this struct needed quite a couple fields in it to be complete.

----

The general idea is there is a single visitor in the parser named `MetadataValidator`.
It takes a `map<string, MetadataInfo>` (where the key is the directive's name) at construction.

And when the validator is ran, if visits each Slice element the parser has found, goes through their lists of metadata and checks them against the map. If there is no matching key, we report a warning (unless it's for another language of course). If we did find an entry in the map, it will automatically perform all the validation using the information in `MetadataInfo`.

For some weirder metadata, there is no straightforward way to declaratively state how to validate a piece of metadata, so the struct also has a special `std::function extraValidation` field, for running truly custom validation after all the normal validation has finished.

I think I was pretty good with writing comments, so I'll stop rambling and let those do the explaining instead.
I would recommend you first review the `MetadataValidator.h` header file before anything else.


cpp/src/Slice/msbuild/slice.vcxproj.filters
```diff
@@ -17,381 +17,20 @@
   <ItemGroup>
     <ClCompile Include="..\FileTracker.cpp" />
     <ClCompile Include="..\Grammar.cpp" />
+    <ClCompile Include="..\MetadataValidator.cpp" />
     <ClCompile Include="..\Parser.cpp" />
     <ClCompile Include="..\Preprocessor.cpp" />
     <ClCompile Include="..\Scanner.cpp" />
     <ClCompile Include="..\SliceUtil.cpp" />
     <ClCompile Include="..\StringLiteralUtil.cpp" />
-    <ClCompile Include="..\FileTracker.cpp" />
```
InsertCreativityHere, line 25: All of these files were either duplicates or references to files that no longer existed.
I'm not sure how this happened, or why they haven't been automatically removed.
So I removed them manually.

cpp/src/slice2cpp/CPlusPlusUtil.cpp
```diff
@@ -136,11 +136,11 @@ namespace
             {
                 string_view directive = meta->directive();
                 string_view arguments = meta->arguments();
-                if (directive == "cpp:view-type" && !arguments.empty())
```
InsertCreativityHere, line 139: We should no longer be performing "on-the-spot" metadata validation like this.

We trust that the metadata validator has done it's thing and any invalid metadata was removed.
No need to double check it's work.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -772,44 +773,19 @@ Slice::Gen::generate(const UnitPtr& p)
     }
 
     // Emit #include statements for any 'cpp:include' metadata directives in the top-level Slice file.
```
InsertCreativityHere, line 775: Same idea, don't double check the metadata validator.
This also has the added benefit of greatly simplifying the code.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -898,289 +874,184 @@ Slice::Gen::writeExtraHeaders(IceInternal::Output& out)
 void
 Slice::Gen::validateMetadata(const UnitPtr& u)
 {
```
InsertCreativityHere, line 876: Now, instead of creating it's own visitor, `validateMetadata` just creates a map detailing the metadata we care about,
and let's the core `MetadataValidator` handle the rest.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -898,289 +874,184 @@ Slice::Gen::writeExtraHeaders(IceInternal::Output& out)
...
+    MetadataInfo viewTypeInfo = {
+        {&typeid(Sequence), &typeid(Dictionary)},
+        MetadataArgumentKind::RequiredTextArgument,
+        nullopt,
+        false,
+        true,
+    };
+    metadataInfo.emplace("cpp:view-type", std::move(viewTypeInfo));
+
+    // "cpp:type"
```
InsertCreativityHere, line 987: The only "messy" thing in this PR is trying to validate the `cpp:type` metadata.
But it's more a product of the metadata being messy itself I think.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -898,289 +874,184 @@ Slice::Gen::writeExtraHeaders(IceInternal::Output& out)
...
+            if (argument != "string" && argument != "wstring")
             {
-                continue;
+                return "invalid argument '" + argument + "' supplied to 'cpp:type' metadata in this context.";
             }
+            return nullopt;
         }
+        else if (dynamic_pointer_cast<Sequence>(appliedTo) || dynamic_pointer_cast<Dictionary>(appliedTo) || dynamic_pointer_cast<ClassDecl>(appliedTo))
+        {
+            return nullopt; // TODO: I see no reason to support 'cpp:type' on class declarations.
```
InsertCreativityHere, line 1044: I really wanted to avoid changing any behavior in this PR, since it also changes alot of other stuff.
After this PR is merged, I'll open a separate PR to remove this, and any of the other weird metadata things that are easier to see now.

cpp/src/Slice/MetadataValidator.cpp
```diff
@@ -0,0 +1,410 @@
...
+        // Otherwise if there's nothing special going on, then the metadata was applied directly to `p`.
+        appliedTo = p;
+    }
+
+    if (appliedTo)
+    {
+        const list<const type_info*>& validOn = info.validOn;
+        auto found =
+            std::find_if(validOn.begin(), validOn.end(), [&](const type_info* t) { return *t == typeid(*appliedTo); });
+        if (!validOn.empty() && found == validOn.end())
```
pepone, line 355: Can we check `validOn.empty()` before with `if (appliedTo`?

cpp/src/Slice/MetadataValidator.cpp
```diff
@@ -0,0 +1,410 @@
...
+        else if (auto dm = dynamic_pointer_cast<DataMember>(p); dm && info.isTypeMetadata)
+        {
+            appliedTo = dm->type();
+        }
+        // Otherwise this metadata is just invalid.
+        else
+        {
+            string msg = '\'' + directive + "' metadata cannot be applied to definitions and declarations" +
+                         ", it can only be applied directly to types";
+            p->unit()->warning(metadata->file(), metadata->line(), InvalidMetadata, msg);
```
pepone, line 340: Is this warning required, wouldn't this be handled below by appliedTo checks?

cpp/src/Slice/MetadataValidator.cpp
```diff
@@ -0,0 +1,410 @@
...
+        if (!wasInserted)
+        {
+            string msg = "ignoring duplicate metadata: '" + directive + "' has already been applied in this context";
+            p->unit()->warning(metadata->file(), metadata->line(), InvalidMetadata, msg);
+            isValid = false;
+        }
+    }
+
+    // Finally, if a custom validation function is specified for this metadata, we run it.
+    if (info.extraValidation.has_value())
```
pepone, line 377: For a nullable types like `std::function` I'lll just use ` nullptr` as not set instead of an optional?

cpp/src/Slice/MetadataValidator.cpp
InsertCreativityHere: Sure!
We do the check up-front now.
I also moved the `find_if` in the `if` statement now that there was extra room.

cpp/test/Slice/errorDetection/WarningInvalidMetadata.err
```diff
@@ -1,21 +1,44 @@
...
+WarningInvalidMetadata.ice:34: warning: 'amd' metadata cannot be specified as file metadata
+WarningInvalidMetadata.ice:40: warning: the 'cpp:header-ext' metadata only accepts one argument but a list was provided
+WarningInvalidMetadata.ice:40: warning: 'cpp:header-ext' metadata cannot be applied to sequences
+WarningInvalidMetadata.ice:44: warning: ignoring duplicate metadata: 'deprecated' has already been applied in this context
+WarningInvalidMetadata.ice:48: warning: ignoring unknown metadata: 'unknown'
+WarningInvalidMetadata.ice:48: warning: ignoring unknown metadata: 'cpp:unknown'
+WarningInvalidMetadata.ice:48: warning: ignoring unknown metadata: 'bad:unknown'
+WarningInvalidMetadata.ice:56: warning: the 'protected' metadata does not take any arguments
+WarningInvalidMetadata.ice:56: warning: the 'cpp:array' metadata does not take any arguments
+WarningInvalidMetadata.ice:62: warning: 'cpp:type' metadata cannot be applied operations with void return type
```
bernardnormier, line 31: The message needs to be grammatically correct

cpp/test/Slice/errorDetection/WarningInvalidMetadata.err
```diff
@@ -1,21 +1,44 @@
...
+WarningInvalidMetadata.ice:65: warning: 'cpp:array' metadata cannot be applied operations with void return type
+WarningInvalidMetadata.ice:68: warning: invalid argument 'my_string' supplied to 'cpp:type' metadata in this context.
```
bernardnormier, line 33: warning ends with period or not? Need to be consistent.

cpp/src/Slice/MetadataValidator.h
```diff
@@ -0,0 +1,104 @@
...
+        /// Indicates whether it's valid and meaningful for this metadata to appear multiple times on the same thing.
+        bool mustBeUnique = true;
+
+        /// A function used to run additional validation for this metadata.
+        /// If this field is set, it will always be run.
+        /// @see ValidationFunc
+        ValidationFunc extraValidation = nullptr;
+    };
+
+    class MetadataValidator final : public ParserVisitor
```
bernardnormier, line 72: This is not the correct API for validation. This visitor is an implementation detail that should not be exposed to "clients" (slice2cpp, slice2java).

A much better API is a simple validateMetadata function on Unit:
```cpp
// That's what slice2cpp calls:
unit->validateMetadata(...);
```

Then this MetadataValidator is used in the _implementation_ of `Unit::validateMetadata`.

cpp/src/Slice/MetadataValidator.cpp
```diff
@@ -0,0 +1,410 @@
...
+{
+    // We want to perform all the metadata validation in the same pass, to keep all the diagnostics in order.
+    // So, we add all the language-agnostic metadata validation into the provided list:
+
+    // "amd"
+    MetadataInfo amdInfo = {
+        {&typeid(InterfaceDef), &typeid(Operation)},
+        MetadataArgumentKind::NoArguments,
+    };
+    _metadataInfo.emplace("amd", std::move(amdInfo));
```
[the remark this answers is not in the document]
externl: Nvm, this way is correct. since we need to construct pairs.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -898,289 +874,190 @@ Slice::Gen::writeExtraHeaders(IceInternal::Output& out)
...
-    // Validate file metadata in the top-level file and all included files.
-    for (const string& file : unit->allFiles())
-    {
-        DefinitionContextPtr dc = unit->findDefinitionContext(file);
-        MetadataList fileMetadata = dc->getMetadata();
-        assert(dc);
-        bool seenHeaderExtension = false;
-        bool seenSourceExtension = false;
-        bool seenDllExport = false;
+    map<string, MetadataInfo> metadataInfo;
```
bernardnormier, line 877: The name of the variable is poorly chosen. A variable name should match a type name only when the variable is of that type (with possibly qualification/pointer and the like).

Here it should be something more like `validationMap`.

cpp/src/slice2cpp/Gen.cpp
InsertCreativityHere: I did spend some time thinking about it, but decided against adding any mechanisms to handle it because:

A) I believe that this is the _only_ metadata with two disjoint uses-cases like this.
So it's not worth complicating everything else to handle a single strange case.

B) I don't want to encourage this kind of metadata re-use. Realistically, these should just be two separate directives.
But they aren't, and we shouldn't break existing Slice definitions just to make the validation logic cleaner IMO.

The truly painful part is that the string flavors can be placed on containers, but the non-string flavors cannot.
Compare this to `cs:generic`. Yes `cs:generic` has a couple 'special' values like `cpp:type` does, but there's still consistent rules for where/when `cs:generic` can be used. Unlike `cpp:type` which has 2 separate rule sets.

cpp/src/Slice/MetadataValidator.cpp
InsertCreativityHere: I reworked this whole section of checks.

cpp/src/Slice/MetadataValidator.h
```diff
@@ -0,0 +1,104 @@
...
+{
+    enum class MetadataArgumentKind
+    {
+        /// The metadata takes no arguments.
+        NoArguments,
+        /// The metadata must have exactly 1 argument.
+        SingleArgument,
+        /// This metadata can have any number of arguments [0, ∞).
+        AnyNumberOfArguments,
+
```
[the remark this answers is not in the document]
InsertCreativityHere: Ideally, I think having these visual groups helps readability, so you can tell that they kind of go together in a certain way.
In lieu of blank lines, should I add a comment above each group of 3 like:
```
        // Something something arguments

        /// The metadata takes no arguments.
        NoArguments,

        /// The metadata must have exactly 1 argument.
        SingleArgument,

        /// This metadata can have any number of arguments [0, ∞).
        AnyNumberOfArguments,

        // Something something text

        /// This metadata must have an argument, but that argument is raw text and not validated further.
        RequiredTextArgument,

        /// This metadata can optionally have an argument, but that argument is raw text and not validated further.
        OptionalTextArgument,
```
Or what's the correct way/style to group related enumerators in C++?

cpp/src/Slice/MetadataValidator.h
```diff
@@ -0,0 +1,104 @@
...
+        /// This metadata must have an argument, but that argument is raw text and not validated further.
+        RequiredTextArgument,
+        /// This metadata can optionally have an argument, but that argument is raw text and not validated further.
+        OptionalTextArgument,
+    };
+
+    /// Typedef for functions which can provide additional validation to certain metadata directives.
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed.

cpp/src/Slice/MetadataValidator.h
```diff
@@ -0,0 +1,104 @@
...
+        /// If this list is empty we don't perform any automatic checking of whether this metadata is validly applied.
+        /// Usually, this is because determining validity isn't as straightforward as matching against a list,
+        /// and requires a more complex approach, which is achieved through providing an `extraValidation` function.
+        std::list<const std::type_info*> validOn;
+
+        /// Specifies how many, and what kinds of arguments, this metadata accepts.
+        MetadataArgumentKind acceptedArguments;
+
+        /// This field stores the specific values that can be provided as arguments to this metadata.
+        /// If this field is unset, then we perform no validation of the arguments (ie. arguments can have any value).
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes, this is the formally correct spelling.
I've always spelt it the colloquial way of `ie.`.
So I went ahead and changed all the `ie.`s I found in the Slice compiler's files to `i.e.` to be fully consistent.

cpp/test/Slice/errorDetection/WarningInvalidMetadata.err
InsertCreativityHere: Another good catch. And another one fixed.

cpp/test/Slice/errorDetection/WarningInvalidMetadata.err
InsertCreativityHere: They shouldn't end with periods, I inserted them by habit I guess.
Good catch. I removed this one, and there were other places where I also accidentally added them (also removed).

cpp/src/Slice/MetadataValidator.h
```diff
@@ -0,0 +1,104 @@
...
+    using ValidationFunc = std::function<std::optional<std::string>(const MetadataPtr&, const SyntaxTreeBasePtr&)>;
+
+    struct MetadataInfo
+    {
+        /// A list of types that this metadata can validly be applied to.
+        ///
+        /// If this list is empty we don't perform any automatic checking of whether this metadata is validly applied.
+        /// Usually, this is because determining validity isn't as straightforward as matching against a list,
+        /// and requires a more complex approach, which is achieved through providing an `extraValidation` function.
+        std::list<const std::type_info*> validOn;
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes, this would save us the `&` address operator from all the lists.
Changed as suggested.

cpp/src/Slice/MetadataValidator.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+#include <algorithm>
+#include <cassert>
+#include <iterator>
+
+using namespace std;
+using namespace Slice;
+
+Slice::MetadataValidator::MetadataValidator(string language, map<string, MetadataInfo> metadataInfo)
+    : _language(language),
```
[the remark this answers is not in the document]
InsertCreativityHere: Yep, good catch, fixed!

cpp/src/Slice/MetadataValidator.cpp
```diff
@@ -0,0 +1,410 @@
...
+        }
+        else if (auto dm = dynamic_pointer_cast<DataMember>(p); dm && info.isTypeMetadata)
+        {
+            appliedTo = dm->type();
+        }
+        // Otherwise this metadata is just invalid.
+        else
+        {
+            string msg = '\'' + directive + "' metadata cannot be applied to definitions and declarations" +
+                         ", it can only be applied directly to types";
```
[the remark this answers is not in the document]
InsertCreativityHere: I reworked this entire section, along with the messages.

cpp/src/Slice/MetadataValidator.cpp
```diff
@@ -0,0 +1,410 @@
...
+Slice::MetadataValidator::misappliedMetadataMessage(const MetadataPtr& metadata, const SyntaxTreeBasePtr& p)
+{
+    string message = '\'' + metadata->directive() + "' metadata cannot be ";
+    if (dynamic_pointer_cast<Unit>(p))
+    {
+        message += "specified as file metadata";
+    }
+    else if (dynamic_pointer_cast<Builtin>(p))
+    {
+        message += "applied to primitive types";
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed!

cpp/src/Slice/MetadataValidator.cpp
```diff
@@ -0,0 +1,424 @@
...
+    }
+
+    // Fifth we check if this metadata is a duplicate, i.e. has the same directive already been applied in this context?
+    if (info.mustBeUnique)
+    {
+        // 'insert' only returns `true` if the value wasn't already present in the set.
+        bool wasInserted = _seenDirectives.insert(directive).second;
+        if (!wasInserted)
+        {
+            string msg = "ignoring duplicate metadata: '" + directive + "' has already been applied in this context";
```
pepone, line 384: Is this behavior new? Does 3.7 ignore duplicates, or let them override existing entries?

pepone (approved):

cpp/src/Slice/MetadataValidator.h
InsertCreativityHere: What I ended up doing was making the visitor inheritance private. It is an implementation detail, and indeed shouldn't be exposed. In it's place I added a single "entry-point" function, but instead of `Unit::validateMetadata(MetadataValidator)`, I added `MetadataValidator::validateMetadataWithin(Unit)`.

For two reasons:
1) "Calling a function on a validator, and passing in the thing to validate" feels more natural than
"calling a function on the thing you want to validate, and passing in the validator".
2) The includes are more logical this way. Currently, `MetadataValidator.h` includes `Parser.h`, and that's it (to get all the grammar types). Implementing your suggestions means that we'd have a circular includes (since `Unit` lives in `Parser.h` and now needs to take a `MetadataValidator` parameter). There's ways to address this, but I think keeping the including in a single direction is nicer.

cpp/src/Slice/MetadataValidation.h
```diff
@@ -0,0 +1,101 @@
...
+        /// Usually, this is because determining validity isn't as straightforward as matching against a list,
+        /// and requires a more complex approach, which is achieved through providing an `extraValidation` function.
+        std::list<std::reference_wrapper<const std::type_info>> validOn;
+
+        /// Specifies how many, and what kinds of arguments, this metadata accepts.
+        MetadataArgumentKind acceptedArguments;
+
+        /// This field stores the specific values that can be provided as arguments to this metadata.
+        /// If this field is unset, then we perform no validation of the arguments (i.e. arguments can have any value).
+        std::optional<StringList> validArgumentValues = std::nullopt;
```
bernardnormier, line 74: Clarify that validArgumentValues has an effect only when acceptedArguments is SingleArgument or AnyNumberOfArguments.

cpp/src/Slice/MetadataValidation.h
```diff
@@ -0,0 +1,101 @@
...
+    {
+        /// A list of types that this metadata can validly be applied to.
+        ///
+        /// If this list is empty we don't perform any automatic checking of whether this metadata is validly applied.
+        /// Usually, this is because determining validity isn't as straightforward as matching against a list,
+        /// and requires a more complex approach, which is achieved through providing an `extraValidation` function.
+        std::list<std::reference_wrapper<const std::type_info>> validOn;
+
+        /// Specifies how many, and what kinds of arguments, this metadata accepts.
+        MetadataArgumentKind acceptedArguments;
```
bernardnormier, line 70: I would pick a field name closer to the field type, e.g. argumentKind.

cpp/src/Slice/MetadataValidation.h
```diff
@@ -0,0 +1,101 @@
...
+    };
+
+    enum class MetadataApplicationContext
+    {
+        /// The metadata can only be applied to definitions and declarations.
+        /// Ex: `["deprecated"] enum A {...}` is valid, but `sequence<["deprecated"] A> S;` is not.
+        Definitions,
+
+        /// The metadata can be applied to definitions of types and to where those types are referenced.
+        /// Ex: both of these are valid: `["java:buffer"] sequence<string> S;` and `["java:buffer"] StringSeq op();`.
```
bernardnormier, line 35: You should include an example for fields, since that's an example of what's beyond Definitions and ParameterTypeReferences.

cpp/src/Slice/MetadataValidation.h
```diff
@@ -0,0 +1,101 @@
...
+    std::string misappliedMetadataMessage(const MetadataPtr& metadata, const SyntaxTreeBasePtr& p);
+
+    /// Validates all the metadata in the provided Unit against a map of known metadata.
+    /// @param p The unit who's metadata should be validated.
+    /// @param language Which language's metadata should be validated. Any metadata that starts with a different
+    ///                 language prefix than this one will be ignored. Note that metadata without any language prefix
+    ///                 (i.e. parser metadata) is always checked.
+    /// @param knownMetadata A map containing the directives that should be validated, and information describing
+    ///                      the various constraints and conditions that should be upheld for it and its arguments.
+    void validateMetadata(const UnitPtr& p, std::string language, std::map<std::string, MetadataInfo> knownMetadata);
```
bernardnormier, line 99: const& since the implementation only reads these parameters.

bernardnormier (approved):

cpp/src/Slice/MetadataValidation.h
InsertCreativityHere: Fixed. I changed my operation example to showcase a field instead.

cpp/src/Slice/MetadataValidation.h
InsertCreativityHere: Also fixed!

cpp/src/Slice/MetadataValidation.h
InsertCreativityHere: Fixed.

cpp/src/Slice/MetadataValidation.h
InsertCreativityHere: I changed the `language` to a `string_view` (since it's preferred over `const string&`)

But the `knownMetadata` map was left as-is, since the implementation does actually modify it, and I'd like to avoid making a working copy. Especially since the caller has no use for the map; it exists solely to be consumed by this function.

And for completeness, the function as a whole shouldn't be `const` because it has side-effects (invalid metadata is filtered out, so the code-gen is assured it will only encounter valid metadata).

PR #3166, opened 2024-11-19
DataStorm fixes for configuration, session retries, and more

pepone:


config/PropertyNames.xml
```diff
@@ -409,6 +409,11 @@
         <property name="Node.Server" class="ObjectAdapter" languages="cpp" />
         <property name="Node.Server.Enabled" default="1" languages="cpp" />
         <property name="Node.Server.ForwardDiscoveryToMulticast" default="0" languages="cpp" />
+        <property name="Topic.ClearHistory" default="OnAll" languages="cpp" />
+        <property name="Topic.DiscardPolicy" default="Never" languages="cpp" />
+        <property name="Topic.Priority" default="0" languages="cpp" />
+        <property name="Topic.SampleCount" default="-1" languages="cpp" />
+        <property name="Topic.SampleLifetime" default="0" languages="cpp" />
```
pepone, line 416: Topic properties where not added before, but they are public and documented in

https://doc.zeroc.com/datastorm/latest/property-reference/datastorm-topic

I updated the docs to include `-1` default for SampleCount this was missing in the docs.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -517,7 +517,6 @@ SessionI::connected(SessionPrx session, const Ice::ConnectionPtr& connection, co
     lock_guard<mutex> lock(_mutex);
     if (_destroyed || _session)
     {
-        assert(_connectedCallbacks.empty());
```
pepone, line 520: Removed unused `_connectedCallbacks` we never add any callbacks.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -605,21 +601,22 @@ SessionI::disconnected(const Ice::ConnectionPtr& connection, exception_ptr ex)
...
             out << (_connection ? _connection->toString() : "<no connection>") << "\n";
             out << e.what();
         }
     }
 
+    auto self = shared_from_this();
     for (auto& t : _topics)
     {
-        runWithTopics(t.first, [&](TopicI* topic, TopicSubscriber&) { topic->detach(t.first, shared_from_this()); });
+        runWithTopics(t.first, [id = t.first, self](TopicI* topic, TopicSubscriber&) { topic->detach(id, self); });
```
pepone, line 619: minor change, no need to call shared_from_this in the loop or capture everything by ref.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -655,20 +650,17 @@ SessionI::retry(NodePrx node, exception_ptr exception)
         }
     }
 
-    if (node->ice_getEndpoints().empty() && node->ice_getAdapterId().empty())
+    // Cancel any pending retry task, before we start a new one below.
+    if (_retryTask)
```
pepone, line 654: Previous code only canceled the retry task when there is no endpoints. Here we are about to schedule a new task I think is clear to cancel any pending task.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -678,16 +670,15 @@ SessionI::retry(NodePrx node, exception_ptr exception)
                 << " (ms) for peer to reconnect";
         }
 
+        // Schedule a timer to remove the session if the peer doesn't reconnect.
         _retryTask = make_shared<IceInternal::InlineTimerTask>([self = shared_from_this()] { self->remove(); });
```
pepone, line 674: Here `SessionI::remove` would call retry again, and we schedule a new call to remove.

As far as I see we have and endless loop of calling retry to print this message and scheduling the remove call that just calls retry.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -703,13 +694,14 @@ SessionI::retry(NodePrx node, exception_ptr exception)
...
             }
+
             if (exception)
             {
                 try
                 {
                     rethrow_exception(exception);
                 }
-                catch (const Ice::LocalException& ex)
+                catch (const std::exception& ex)
```
pepone, line 704: This is to match what we do above, when we handle what exceptions to retry.

cpp/src/DataStorm/TopicI.cpp
```diff
@@ -13,24 +13,6 @@ using namespace DataStormContract;
...
-        return seq;
-    }
-
-    int toInt(const string& v, int value = 0)
-    {
-        istringstream is(v);
-        is >> value;
-        return value;
-    }
-
```
pepone, line 33: unused after updating handling of configurations.

cpp/src/DataStorm/TopicI.cpp
```diff
@@ -57,6 +39,54 @@ namespace
...
+            throw Ice::ParseException(__FILE__, __LINE__, "Invalid clear history policy: " + value);
+        }
+    }
+
+    DataStorm::DiscardPolicy parseDiscardPolicy(const std::string& value)
+    {
+        if (value == "Never")
+        {
+            return DataStorm::DiscardPolicy::None;
+        }
```
pepone, line 76: We have this discrepancy, documented value is "Never", but the enum value is None.

cpp/src/DataStorm/TopicI.cpp
```diff
@@ -272,24 +302,24 @@ TopicI::getElementSpecs(int64_t topicId, const ElementInfoSeq& infos, const shar
 }
 
 void
-TopicI::attach(int64_t id, const shared_ptr<SessionI>& session, SessionPrx prx)
+TopicI::attach(int64_t id, shared_ptr<SessionI> session, SessionPrx peerSession)
 {
-    auto p = _listeners.find(ListenerKey{session});
```
pepone, line 277: ListenerKey was not required for this map, we were using the default shared_ptr operator<

cpp/src/DataStorm/TopicI.cpp
```diff
@@ -876,8 +868,7 @@ TopicReaderI::TopicReaderI(
           std::move(name),
           id)
 {
-    _defaultConfig = {-1, 0, DataStorm::ClearHistoryPolicy::OnAll, DataStorm::DiscardPolicy::None};
```
pepone, line 879: Move the defaults to the property defaults, and update the code to use getIceProperty.

cpp/src/DataStorm/TopicI.cpp
```diff
@@ -954,27 +945,14 @@ TopicReaderI::destroy()
...
-            config.discardPolicy = DataStorm::DiscardPolicy::None;
-        }
-        else if (p->second == "SendTime")
-        {
-            config.discardPolicy = DataStorm::DiscardPolicy::SendTime;
-        }
-        else if (p->second == "SendTime")
-        {
-            config.discardPolicy = DataStorm::DiscardPolicy::Priority;
-        }
```
pepone, line 976: This was bogus, but it seems we don't have any tests that catches this error.

cpp/src/DataStorm/SessionI.cpp
externl: Could you declare self in the capture block to avoid an extra copy?

cpp/src/DataStorm/TopicI.cpp
externl: I assume `Never` is the "right" choice here.

externl (approved):
Looks good to me!

cpp/src/DataStorm/TopicI.cpp
pepone: DiscardPolicy is part of the public API. User code can still set it manually using `DiscardPolicy::None`.

Should we rename `DiscardPolicy::None` to `DiscardPolicy::Never`?

externl: I think it's preferable to just have one value.

PR #3167, opened 2024-11-20
RedHat multi-platform build fixes

pepone:


externl (approved):

externl:
Do we need this on main as well?

pepone:
> Do we need this on main as well?

Yes

config/Make.rules.Linux
```diff
@@ -35,19 +35,25 @@ else ifneq ($(and $(filter centos rhel fedora,$(linux_id)),$(filter x86_64 i%86,
...
-opt-cppflags = $(if $(CXXFLAGS),,$(filter-out -m32 -m64,$(shell setarch $1 rpm --eval %optflags)))
+opt-cppflags = $(if $(CXXFLAGS),,$(filter-out -m32 -m64 -march=%,$(shell setarch $1 rpm --eval %optflags)))
 opt-ldflags  = $(if $(LDFLAGS),,$(shell setarch $1 rpm --eval %?__global_ldflags))
 endif
 
-x64_cppflags            := $(call opt-cppflags,x86_64) -m64
+CXXFLAGS                := $(filter-out -m32 -m64 -march=%, $(CXXFLAGS))
+LDFLAGS                 := $(filter-out -m32 -m64, $(LDFLAGS))
+
+x64_cppflags            := $(call opt-cppflags,x86_64) -m64 -march=x86-64-v2
```
bernardnormier, line 51: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

Why `x86-64-v2` and not simply `x86-64`?

config/Make.rules.Linux
```diff
@@ -35,19 +35,25 @@ else ifneq ($(and $(filter centos rhel fedora,$(linux_id)),$(filter x86_64 i%86,
...
 x64_ldflags             := $(call opt-ldflags,x86_64) -m64
 x64_targetdir           = $(if $(filter %/build,$5),/x64,$(if $(filter-out $($1_target),program),64))
 x64_installdir          = $(if $(and $(filter-out $($1_target),program),$(if $5,$(filter-out %64,$5),true)),64)
 
-x86_cppflags            := $(call opt-cppflags,i386) -m32
+x86_cppflags            := $(call opt-cppflags,i386) -m32 -march=x86-64-v2
```
bernardnormier, line 56: Same here.

config/Make.rules.Linux
bernardnormier: Ok, would be nice to add a comment.

Issue #3141, opened 2024-11-12
Rename 'Comment' to 'DocComment' in the Slice Parser


InsertCreativityHere:
Currently the class we use for describing doc comments is just named `Comment`,
it would be more technically correct to rename this to `DocComment`.

InsertCreativityHere:
Fixed in #3170.

PR #3170, opened 2024-11-20
Rename `Comment` to `DocComment` in the Slice Compilers

InsertCreativityHere:
This PR implements #3141 by changing all types and 'public' functions that talk about 'Comment' to talk about 'DocComment's instead.

There were plenty of local variables named `comment`. I left these alone. It's already very clear from the context that these are doc-comments.


cpp/src/Slice/SliceUtil.cpp
```diff
@@ -321,7 +321,7 @@ Slice::filterMcppWarnings(const string& message)
 }
 
 void
-Slice::printGeneratedHeader(IceInternal::Output& out, const string& path, const string& comment)
```
InsertCreativityHere, line 324: This variable was named `commentStyle` in the header file:
```
printGeneratedHeader(IceInternal::Output& out, const std::string& path, const std::string& commentStyle = "//");
```
I fixed this to use the same name. Header and source files should use the same name IMO.

Issue #3150, opened 2024-11-13
Simplify How the Parser Splits Language-Specific Metadata


InsertCreativityHere:
When the parser encounters metadata, one of the first things it does is split the raw string into `directive:arguments`.
But unfortunately, the `directive` part can also contain colons too: `cpp:include:arguments`.

The check that we use to determine where the boundary between `directory` and `arguments` uses the following heuristic:
```
if (colonCount == 0) -> then the whole thing is a directive, and arguments = "".
if (colonCount >= 2) -> then everything before the 2nd colon is a directive, and everything after it are arguments
if (colonCount == 1) -> we check the first piece against a list of known language prefixes to see if any match.
```

We should eliminate the `(colonCount >= 2)` case. Checking for the language prefix will _always_ be sufficient for valid metadata.
And it will let us catch metadata like `foo:bar:baz` as illegal. Right now, with the current check we see `>= 2` colons and say
"ah, `foo` must be a language we don't know about. I'll skip validating it".
But this is kind of dumb. We know all of the languages that we support, so there's no reason not to keep a hard-coded list, and reject anything else.

PR #3172, opened 2024-11-20
Simplified the Parser's Logic for Splitting up Raw Metadata

InsertCreativityHere:
This PR implements #3150.

Currently, if a directive contains no colons, we know it's parser-metadata, (`["protected"]`)
and if it contains 2 colons, we just assume it's language specific metadata. (`["cpp:header-ext:hh"]`)
If there is only 1 colon, then we use a more complex check which will inspect whether the first piece is a known language prefix.

For _valid_ metadata, this is all correct.
But a downside of this approach is that for something like `foo:bar:baz`, we'll assume that `foo` is a language, and since it's a language that `slice2xxx` won't ever recognize, it will be silently skipped. This is suboptimal, since this is nonsense metadata.

----

This PR keeps the same logic for directives with no colons, but eliminated this assumption about 2 colons meaning a language.
Now, if there are _any_ colons in a piece of metadata, we will always check if it starts with a valid language.
And only if it does start with a valid language will it be treated as language metadata. If it doesn't, it is treated as parser metadata.

(Note that this works well with the metadata validator PR I also have open, which will then reject `foo:bar:baz` as invalid parser metadata).

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -105,7 +105,7 @@ Slice::Metadata::Metadata(string rawMetadata, string file, int line) : GrammarBa
     {
         // Check if the metadata starts with a language prefix.
         // NOTE: It is important that this list is kept in alphabetical order!
-        static const string languages[] = {"cpp", "cs", "java", "js", "matlab", "php", "python", "ruby", "swift"};
+        constexpr string_view languages[] = {"cpp", "cs", "java", "js", "matlab", "php", "python", "ruby", "swift"};
         string prefix = rawMetadata.substr(0, firstColonPos);
         bool hasLangPrefix = binary_search(&languages[0], &languages[sizeof(languages) / sizeof(*languages)], prefix);
         if (hasLangPrefix)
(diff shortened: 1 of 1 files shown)


cpp/src/Slice/Parser.cpp
```diff
@@ -98,46 +98,41 @@ Slice::Metadata::Metadata(string rawMetadata, string file, int line) : GrammarBa
...
+    if (firstColonPos != string::npos)
     {
-        // If the metadata contains only 1 colon, we need to check if it's for a language prefix, or for arguments.
+        // Check if the metadata starts with a language prefix.
         // NOTE: It is important that this list is kept in alphabetical order!
         static const string languages[] = {"cpp", "cs", "java", "js", "matlab", "php", "python", "ruby", "swift"};
         string prefix = rawMetadata.substr(0, firstColonPos);
-        bool isLanguage = binary_search(&languages[0], &languages[sizeof(languages) / sizeof(*languages)], prefix);
-        if (isLanguage)
+        bool hasLangPrefix = binary_search(&languages[0], &languages[sizeof(languages) / sizeof(*languages)], prefix);
```
bernardnormier, line 110: I would convert languages to a constexpr of string_view.

bernardnormier (approved):

externl (approved):

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Fixed.

Issue #2813, opened 2024-09-27
csharp/IceGrid/simple failure on Ubuntu 24.04


bernardnormier:
From https://github.com/zeroc-ice/ice/actions/runs/11075786900/job/30777485537 (on main):

```
*** [67/304] Running csharp/IceGrid/simple tests ***
[ running without deployment test - 09/27/24 18:49:33 ]
- Config: ssl,mx
[...]

PR #3177, opened 2024-11-21
Misc fixes to C# locator discovery

bernardnormier:
This PR fixes a couple bugs in the C# LocatorDiscovery and removes dead code.

See also #2813.

### DIFF
--- a/csharp/src/IceLocatorDiscovery/PluginI.cs
+++ b/csharp/src/IceLocatorDiscovery/PluginI.cs
@@ -427,7 +427,7 @@ internal class LocatorI : Ice.BlobjectAsync, Ice.Internal.TimerTask
 
                         foreach (var l in _lookups)
                         {
-                            _ = preformFindLocatorAsync(l.Key, l.Value);
+                            _ = performFindLocatorAsync(l.Key, l.Value);
                         }
                         _timer.schedule(this, _timeout);
                     }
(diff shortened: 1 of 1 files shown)


csharp/src/IceLocatorDiscovery/PluginI.cs
```diff
@@ -11,11 +11,6 @@ public Ice.Plugin
     create(Ice.Communicator communicator, string name, string[] args) => new PluginI(communicator);
 }
 
-public interface Plugin : Ice.Plugin
```
bernardnormier, line 14: We use this (obviously) internal API in C++ and Java, but not in C#, where it's internal unused code.

csharp/src/IceLocatorDiscovery/PluginI.cs
```diff
@@ -206,48 +201,6 @@ public override Task<Ice.Object_Ice_invokeResult>
...
-        {
-            Thread.Sleep(waitTime);
-        }
-        else
-        {
-            lock (_mutex)
-            {
-                while (!_locators.ContainsKey(instanceName) && _pending)
-                {
-                    Monitor.Wait(_mutex, waitTime);
```
bernardnormier, line 237: The only Monitor.Wait(_mutex, ...), which means all the Monitor.Pulse(_mutex) elsewhere are useless.

csharp/src/IceLocatorDiscovery/PluginI.cs
```diff
@@ -457,8 +409,11 @@ public void
             }
         }
 
-        async Task preformFindLocatorAsync(LookupPrx lookupPrx, LookupReplyPrx lookupReplyPrx)
+        async Task performFindLocatorAsync(LookupPrx lookupPrx, LookupReplyPrx lookupReplyPrx)
         {
+            // Exit the mutex lock before proceeding.
```
bernardnormier, line 414: Awaiting a remote call with a mutex lock (see below) is incorrect and deadlock prone. It could be the cause for #2813.

externl (approved):

bernardnormier:
The timer task (and runTimerTask) are used, e.g.
```csharp
         _timer.schedule(this, _timeout);
```

PR #3178, opened 2024-11-22
Clear local options from computed published endpoints

### LINKED ISSUE #3176: Clear local endpoint options from default PublishedEndpoints
When the PublishedEndpoints of an object adapter are computed from the endpoints of this adapter, they include all local options of these OA endpoints, which is rather confusing.

For example:
```
-- 11/18/24 16:21:47138 Slave2: Network: starting to receive udp messages
   local address = 0.0.0.0:12109
   multicast address = 239.255.0.1:12109
   local interfaces = 127.0.0.1
-- 11/18/24 16:21:47140 Slave2: Network: published endpoints for object adapter 'IceGrid.Registry.Discovery':
   udp -h MyFancyComputer -p 12109 --interface 127.0.0.1
-- 11/18/24 16:21:47141 Slave2: Network: accepting tcp connections at 127.0.0.1:65375
-- 11/18/24 16:21:47141 Slave2: Network: accepting tcp connections at 127.0.0.1:12032
-- 11/18/24 16:21:47141 Slave2: Network: accepting tcp connections at 127.0.0.1:65376
-- 11/18/24 16:21:47141 Slave2: Network: accepting tcp connections at 127.0.0.1:65377
-- 11/18/24 16:21:47169 Slave2: Network: accepted tcp connection
   local address = 127.0.0.1:12032
   remote address = 127.0.0.1:65379
```

#3175 changes the default published host in the published endpoints (keep the multicast address).

However, the local option `--interface 127.0.0.1` remains and is rather confusing. It's meaningless for remote clients that receive published proxies, and local clients shouldn't be a primary consideration here.

A related issue is the `equivalent` logic of IPEndpointI:
https://github.com/zeroc-ice/ice/blob/fad9c08d4c97fee479600536289aac91cf9c28c3/csharp/src/Ice/Internal/IPEndpointI.cs#L146

equivalent is used only for colloc optimization:
https://github.com/zeroc-ice/ice/blob/fad9c08d4c97fee479600536289aac91cf9c28c3/csharp/src/Ice/ObjectAdapter.cs#L875

The net effect is if you set the source address on your proxy, you disable colloc. This does not seem intentional and I propose we remove this comparison of source address (a local option) in the implementation of `equivalent`.

bernardnormier:
This PR clears local endpoint options (such as --interface) from published endpoints.

Fixes #3176.

### DIFF
--- a/csharp/src/Ice/Internal/TcpEndpointI.cs
+++ b/csharp/src/Ice/Internal/TcpEndpointI.cs
@@ -223,7 +223,7 @@ internal sealed class TcpEndpointI : IPEndpointI
         }
         else
         {
-           return new TcpEndpointI(instance_, publishedHost, port_, sourceAddr: null, _timeout, conId: "", _compress);
+            return new TcpEndpointI(instance_, publishedHost, port_, sourceAddr: null, _timeout, conId: "", _compress);
         }
     }
 


csharp/src/Ice/Internal/IPEndpointI.cs
```diff
@@ -152,8 +150,7 @@ public override bool equivalent(EndpointI endpoint)
         IPEndpointI ipEndpointI = (IPEndpointI)endpoint;
         return ipEndpointI.type() == type() &&
             ipEndpointI.host_.Equals(host_, StringComparison.Ordinal) &&
-            ipEndpointI.port_ == port_ &&
-            Network.addressEquals(ipEndpointI.sourceAddr_, sourceAddr_);
+            ipEndpointI.port_ == port_;
```
bernardnormier, line 153: Removed sourceAddr from this equivalent comparison in C# and Java; in C++, we didn't check sourceAddr.

cpp/src/Ice/TcpEndpointI.cpp
```diff
@@ -122,6 +122,26 @@ IceInternal::TcpEndpointI::datagram() const
     return false;
 }
 
+shared_ptr<EndpointI>
+IceInternal::TcpEndpointI::toPublishedEndpoint(string publishedHost) const
+{
+    if (publishedHost.empty() && !isAddressValid(_sourceAddr) && _connectionId.empty())
```
pepone: I read this as:

If the published host is empty and there is no local options. The published endpoint is the actual endpoint. Otherwise replace publishedHost and remove local options.

Maybe add a comment to clarify.

cpp/src/Ice/UdpEndpointI.cpp
```diff
@@ -149,6 +149,21 @@ IceInternal::UdpEndpointI::datagram() const
     return true;
 }
 
+shared_ptr<EndpointI>
+IceInternal::UdpEndpointI::toPublishedEndpoint(string publishedHost) const
+{
+    return make_shared<UdpEndpointI>(
```
pepone, line 155: Why not reuse current endpoint like in tcp, is it never possible?

pepone (approved):

cpp/src/Ice/TcpEndpointI.cpp
bernardnormier: See latest version.

An empty publishedHost means "keep existing host as is".

cpp/src/Ice/UdpEndpointI.cpp
bernardnormier: It's possible, but requires more code/complexity, and there is no benefit since we don't compute these published endpoints often.

PR #3179, opened 2024-11-22
Add C++ coverage report workflow  

externl:
This PR adds a new workflow to generate C++ coverage reports and upload them to S3. This workflow is ran on demand to avoid consuming yet another macOS runner.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,56 @@
+name: Generate API Coverage
+
+on:
+  workflow_dispatch:
+  push:
+    branches: ["coverage-ci"]
+
+jobs:
+  generate-coverage-report:
+    runs-on: macos-15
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,54 @@
...
+            ../scripts/generate-code-coverage.sh "$binary"
+          done
+
+          for library in lib/*; do
+            if [[ $library =~ lib/lib[a-zA-Z0-9]+\.dylib ]]; then
+              ../scripts/generate-code-coverage.sh "$library"
+            fi
+          done
+
+      # This will perform a full sync of the documentation to S3 every time the workflow is run since
```
pepone, line 35: We should update the comment, this is not about doxygen.

pepone (approved):

PR #3180, opened 2024-11-22
Cleanup EndpointInfo/ConnectionInfo in C++, C#, Java and JS

bernardnormier:
The EndpointInfo and ConnectionInfo classes used to be generated from local Slice and the checked-in version had lots of unused code.


cpp/include/Ice/Endpoint.h
```diff
@@ -9,41 +9,48 @@
...
 
 namespace Ice
 {
     /**
-     * Base class providing access to the endpoint details.
+     * The user-level interface to an endpoint.
      * \headerfile Ice/Ice.h
      */
-    class ICE_API EndpointInfo
+    class ICE_API Endpoint
```
bernardnormier, line 20: Change the order in this file (and other files): first Endpoint, then EndpointInfo and all other Info classes.

cpp/include/Ice/Endpoint.h
```diff
@@ -116,32 +94,6 @@ namespace Ice
...
-        /**
-         * One-shot constructor to initialize all data members.
-         * @param underlying The information of the underlying endpoint or null if there's no underlying endpoint.
-         * @param timeout The timeout for the endpoint in milliseconds.
-         * @param compress Specifies whether or not compression should be used if available when using this endpoint.
-         * @param host The host or address configured with the endpoint.
-         * @param port The port number.
-         * @param sourceAddress The source IP address.
-         */
-        IPEndpointInfo(
```
bernardnormier, line 129: The code coverage report showed that all these one-shot ctors weren't called, with the exception of the OpaqueEndpointInfo ctor.

js/src/Ice/Endpoint.js
```diff
@@ -1,62 +1,29 @@
...
-
-import { Ice as Ice_Version } from "./Version.js";
-const { EncodingVersion } = Ice_Version;
+// Copyright (c) ZeroC, Inc.
 
 /**
  *  Base class providing access to the endpoint details.
  **/
-export class EndpointInfo {
-    constructor(underlying = null, timeout = 0, compress = false) {
```
bernardnormier, line 12: Application code should not construct EndpointInfo/ConnectionInfo, so there is no reason to provide constructors in TS.

cpp/include/Ice/Connection.h
```diff
@@ -54,54 +46,6 @@ namespace Ice
 
 namespace Ice
```
pepone, line 47: The namespace Ice is declared twice in this file, we can merge the declarations.

js/src/Ice/Endpoint.js
```diff
@@ -1,62 +1,29 @@
...
-        this.underlying = underlying;
-        this.timeout = timeout;
-        this.compress = compress;
-    }
-}
+export class EndpointInfo {}
```
pepone, line 6: Isn't this changing the values used for default constructed instances?

for example with the new code `new EndpointInfo` have undefined value for `timeout` instead of `0`.

pepone (approved):

externl (approved):

cpp/include/Ice/Connection.h
bernardnormier: Fixed.

js/src/Ice/Endpoint.js
bernardnormier: Yes, except we actually always overwrite everything later, usually with `fillWithEndpointInfo`.

All this code is not very elegant, and I believe its structure is a consequence of the local Slice mapping. We should refactor in a follow-up PR: reintroduce internal constructors, simplify the structure (EndpointInfo should not hold onto Endpoint), and more.

PR #3185, opened 2024-11-24
Refactor EndpointInfo in C#

bernardnormier:
This PR simplifies EndpointInfo in C#, in particular, make all the fields readonly.

### DIFF
--- a/csharp/src/Ice/Endpoint.cs
+++ b/csharp/src/Ice/Endpoint.cs
@@ -19,94 +19,170 @@ public interface Endpoint
 /// <summary>
 /// Base class providing access to the endpoint details.
 /// </summary>
-public abstract class EndpointInfo
+public class EndpointInfo
 {
     /// <summary>
-    /// The information of the underlying endpoint or null if there's no underlying endpoint.
+    /// Gets the information for the underlying endpoint, or null if there's no underlying endpoint.
     /// </summary>
-    public EndpointInfo? underlying;
+    public readonly EndpointInfo? underlying;
 
     /// <summary>
-    /// The timeout for the endpoint in milliseconds. -1 means no timeout.
+    /// Gets the timeout of the endpoint in milliseconds. -1 means no timeout.
     /// </summary>
-    public int timeout;
+    public readonly int timeout;
 
     /// <summary>
-    /// Specifies whether or not compression should be used if available when using this endpoint.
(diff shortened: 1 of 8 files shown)


csharp/src/Ice/Endpoint.cs
```diff
@@ -19,94 +19,170 @@ public interface Endpoint
...
-
-    /// <summary>Returns true if this endpoint is a datagram endpoint.</summary>
-    /// <returns>True for a datagram endpoint.</returns>
-    public abstract bool datagram();
-
-    /// <summary>Returns true if this endpoint is secure; otherwise false.</summary>
-    /// <returns>True if the endpoint is secure.</returns>
-    public abstract bool secure();
+    /// <remarks>The type of an underlying endpoint is always the same as the type its enclosing endpoint.</remarks>
+    public virtual short type() => underlying?.type() ?? -1;
```
pepone, line 42: Should we instead assert underlying is not null?

pepone (approved):

csharp/src/Ice/Internal/WSEndpoint.cs
```diff
@@ -34,40 +34,7 @@ internal WSEndpoint(ProtocolInstance instance, EndpointI del, Ice.InputStream s)
...
-    public override Ice.EndpointInfo getInfo()
-    {
-        Ice.WSEndpointInfo info = new InfoI(this);
-        info.underlying = _delegate.getInfo();
-        info.resource = _resource;
-        info.compress = info.underlying.compress;
-        info.timeout = info.underlying.timeout;
-        return info;
-    }
+    public override EndpointInfo getInfo() => new WSEndpointInfo(_delegate.getInfo(), _resource);
```
bentoi, line 37: It's not clear to me how the secure, type, datagram, timeout, compress fields of this returned `WSEndpointInfo` are set. 

They should be set to the underlying endpoint info values.

bernardnormier: See the base class, Ice.EndpointInfo.

type(), datagram() and secure() delegate to underlying, while the timeout and compress fields are set in the constructor from underlying.

csharp/src/Ice/Endpoint.cs
bernardnormier: Even with a Debug.Assert, we would still need to return something. So I think -1 is fine here.

PR #3188, opened 2024-11-25
Address Dead Code in the Slice Compilers

InsertCreativityHere:
@externl's new coverage report was a useful navigation tool that led me to a few chunks of unused code.
Most of this code is just deleted by this PR, but there were things that _should_ of been being used by the compilers instead.

I'm just opening this PR to let CI double check that I haven't broken anything.


cpp/src/Slice/Grammar.y
```diff
@@ -562,7 +562,6 @@ optional
 {
     auto scoped = dynamic_pointer_cast<StringTok>($2);
     ContainerPtr cont = currentUnit->currentContainer();
-    assert(cont);
```
InsertCreativityHere, line 565: `currentContainer` can never return null:
```
ContainerPtr
Slice::Unit::currentContainer() const
{
    assert(!_containerStack.empty());
    return _containerStack.top();
}
```

cpp/src/Slice/Parser.cpp
```diff
@@ -405,19 +405,6 @@ Slice::Type::usesClasses() const
 // Builtin
 // ----------------------------------------------------------------------
```
InsertCreativityHere, line 407: `typeId` is only used in error messages in two places.
Both places didn't care about this `Ice::` prefix either.
I inlined the logic in the 2 places where it was needed.

cpp/src/Slice/Parser.cpp
```diff
@@ -1238,7 +1227,6 @@ Slice::Container::createClassDecl(const string& name)
         }
     }
 
-    _unit->currentContainer();
```
InsertCreativityHere, line 1241: `currentContainer` is `const`. All it does is return a reference to the current container.
So, calling this function and ignoring the return value is pointless.

cpp/src/Slice/Parser.cpp
```diff
@@ -1870,22 +1857,6 @@ Slice::Container::modules() const
     return result;
 }
```
InsertCreativityHere, line 1859: Unused code.

cpp/src/Slice/Parser.cpp
```diff
@@ -2146,11 +2102,6 @@ Slice::Container::validateConstant(
 {
     // isConstant indicates whether a constant or a data member (with a default value) is being defined.
```
InsertCreativityHere, line 2104: It's impossible for `type` to be `nullptr` here.
In the grammar file at the top of this PR, the `type` rule used to return `nullptr` if there's no currentContainer.
But there is always a container, so this `nullptr` return was impossible, meaning it's impossible to hit here.

cpp/src/Slice/Parser.cpp
```diff
@@ -2400,12 +2351,6 @@ Slice::Module::Module(const ContainerPtr& container, const string& name)
 // Constructed
 // ----------------------------------------------------------------------
```
InsertCreativityHere, line 2353: The `typeid` function is trivial and only used in 2 places, I inlined it in those 2 places.

cpp/src/Slice/Parser.cpp
```diff
@@ -2693,18 +2637,6 @@ Slice::ClassDef::canBeCyclic() const
     return false;
 }
```
InsertCreativityHere, line 2639: Unused code.

cpp/src/Slice/Parser.cpp
```diff
@@ -3042,7 +2973,6 @@ Slice::InterfaceDef::createOperation(
         }
     }
```
InsertCreativityHere, line 2975: We never checked the `_hasOperations` field,
so I removed it.

cpp/src/Slice/Parser.cpp
```diff
@@ -3790,23 +3700,6 @@ Slice::Exception::usesClasses() const
     return false;
 }
```
InsertCreativityHere, line 3702: More dead code.

cpp/src/Slice/SliceUtil.cpp
```diff
@@ -250,9 +250,9 @@ Slice::filterMcppWarnings(const string& message)
 {
     static const char* messages[] = {"Converted [CR+LF] to [LF]", "no newline, supplemented newline", 0};
```
InsertCreativityHere, line 252: No reason not to use `constexpr string_view` instead...

cpp/src/ice2slice/Gen.cpp
```diff
@@ -666,10 +666,22 @@ Gen::TypesVisitor::visitEnum(const EnumPtr& p)
 void
 Slice::Gen::TypesVisitor::visitConst(const ConstPtr& p)
 {
```
InsertCreativityHere, line 668: The place where I inlined the implementation for the `typeId` method.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -183,99 +183,7 @@ Slice::CsGenerator::fixId(const string& name, unsigned int baseTypes, bool mangl
 string
 Slice::CsGenerator::getOptionalFormat(const TypePtr& type)
 {
```
InsertCreativityHere, line 185: The code coverage report marked `Slice::Type::getOptionalFormat` as unused.
But instead of deleting it, the correct fix was to start using it.

Every single language has it's own `getOptionalFormat` which is just a re-implementation of the original one, with some prefix slapped on front. So I deleted all the redundant code, and now they all just call the `getOptionalFormat` defined in the parser...

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -372,7 +372,7 @@ namespace
                     {
                         BuiltinPtr b = dynamic_pointer_cast<Builtin>(p);
                         assert(b);
```
InsertCreativityHere, line 374: The 2nd place where we used to `typeId`. But here we already know that we're dealing with a `builtin` type, so using the virtual function `typeId` was pointless anyways. I switched it to just use the `kindAsString` method which returns `int`, `string`, `Value`, etc. and is specific to built-in types.

pepone (approved):

PR #3191, opened 2024-11-25
Refactor EndpointInfo and ConnectionInfo in C++

bernardnormier:
This is the C++ equivalent of #3185 and #3187.


cpp/src/Ice/SSL/SecureTransportTransceiverI.cpp
```diff
@@ -503,24 +503,22 @@ Ice::SSL::SecureTransport::TransceiverI::toDetailedString() const
...
         CFRetain(peerCertificate);
-        info->peerCertificate = peerCertificate;
-    }
-    else
-    {
-        info->peerCertificate = nullptr;
     }
 
-    return info;
+    return make_shared<Ice::SSL::ConnectionInfo>(
```
pepone, line 519: A minor issue with this update, and other certificates. If the constructor throws, or `_delegate->getInfo()` throws. we leak the peerCertificate.

pepone (approved):
Looks good, there was a failure with the MATLAB build.

cpp/src/Ice/SSL/SecureTransportTransceiverI.cpp
bernardnormier: It's not clear to me why the constructor or _delegate->getInfo would throw.

bernardnormier:
> Looks good, there was a failure with the MATLAB build.

Small issue in the MATLAB code, now fixed. A more interesting issue was the failure on macos release, apparently due to weak vtables. Also fixed.

PR #3193, opened 2024-11-25
Fix Slice Scanner to Only Treat '/**' as Doc Comments

### LINKED ISSUE #3192: The Slice Scanner Treats All C-Style Comments as Doc-Comments
This is a C-style comment: `/* hello */`
This is a JavaDoc-style doc -comment: `/** hello */`.

Currently the Slice compilers are treating _all_ C-style comments as doc comments.
This is not correct.

InsertCreativityHere:
This PR fixes #3192, a pretty egregious bug in the Slice scanner.
It was treating all comments of the form `/* */` as doc comments, when doc comments should require two stars on the opening slash (`/**`). So, this whole time, we've been generating doc-comments for all C-style comments.

Now, the Slice compiler will only treat a C-style comment as a doc-comment if it starts with `/**`, and the next character is anything other than `*` or `/`. It must have _exactly_ 2 stars, no more and no less.

These rules protect us against:
`/**/` which isn't a doc comment, it's really just an empty C-style comment (like `/*  */`).
and people who like to write big banner-style comments like:
```
/********************
        THING
********************/
```


cpp/src/Slice/Scanner.cpp
InsertCreativityHere, line 1: This file is generated by `flex`, and probably not worth reviewing.

cpp/src/Slice/Scanner.l
```diff
@@ -92,6 +92,7 @@ namespace
 
   /* List of start-condition states the scanner can be in. This lets the scanning be context dependent. */
```
InsertCreativityHere, line 93: At any moment the Slice scanner is in a specific state (called a "Start Condition").
This list contains all the possible states that it can be within.

Previously, we only had a single C_COMMENT state, and whenever the scanner was in this state, it would add any text it saw to the current doc-comment. This was bad.
So, now we have 2 states. C_DOC_COMMENT is triggered when the scanner sees /** and C_COMMENT is triggered when the scanner sees /*. The scanner will try to match rules in order so it will always try to match /** first.

And, the scanner will only add things to doc comments while it is in the C_DOC_COMMENT state.

cpp/src/Slice/Scanner.l
```diff
@@ -315,47 +316,59 @@ floating_literal    (({fractional_constant}{exponent_part}?)|((\+|-)?{dec}+{expo
...
 <*>"///".* {
     currentUnit->addToDocComment(yytext + 3);
 }
 
   /* Matches and consumes a C++ style comment. */
 <*>"//".* {}
 
+  /* Matches the start of a C style doc-comment, and switches the scanner to the C_DOC_COMMENT state.
+   * Specifically we match the literal "/**" when it is _not_ followed by either a '/' or '*' character. */
+<*>"/**"/[^/*] {
```
InsertCreativityHere, line 329: This is the flex-regex-style for "match `/**`, then do a positive lookahead (the `/`) for anything that is **not** (the `^`) either `/` or `*`".

cpp/src/Slice/Scanner.l
```diff
@@ -315,47 +316,59 @@ floating_literal    (({fractional_constant}{exponent_part}?)|((\+|-)?{dec}+{expo
...
 <*>"/*" {
     yy_push_state(C_COMMENT);
 }
 
   /* Matches any character except for newlines and adds them to the comments. '*' are matched one at a time to ensure
    * Flex scans '* /' correctly. Flex prioritizes longer matches over shorter ones, so '* /' will match before '*'. */
-<C_COMMENT>"*"     |
-<C_COMMENT>[^\n*]+ {
+<C_COMMENT,C_DOC_COMMENT>"*"     |
+<C_COMMENT,C_DOC_COMMENT>[^\n*]+ {
```
InsertCreativityHere, line 341: In Flex, anything in angle brackets means: "only match this rule if you are in one of these start conditions".
We re-use these rules for both the `C_COMMENT` and `C_DOC_COMMENT` conditions, since they only difference is whether the text counts towards a doc-comment or not.

cpp/src/Slice/Scanner.l
```diff
@@ -315,47 +316,59 @@ floating_literal    (({fractional_constant}{exponent_part}?)|((\+|-)?{dec}+{expo
...
+<C_COMMENT,C_DOC_COMMENT>"*/" {
     yy_pop_state();
 
     string comment(yytext);
     // The last 2 characters are the '*/' matched by this rule.
-    currentUnit->setDocComment(comment.substr(0, yyleng - 2));
+    if (YY_START == C_DOC_COMMENT)
+    {
+        currentUnit->setDocComment(comment.substr(0, yyleng - 2));
+    }
```
InsertCreativityHere, line 360: We only set the text as a doc comment if we're in the C_DOC_COMMENT start condition.

pepone (approved):

bernardnormier (approved):

PR #3196, opened 2024-11-26
Minor Comment Fixes

InsertCreativityHere:
The main point of this PR was to fix our usage of asymmetric quotations like `` `hello' `` in our Slice files.
These seem to confuse Doxygen, which thinks that they are un-closed code-blocks.
I wrote more about this on my blog, but these need to be removed before we can switch to triple-slash style doc-comments.

While reviewing the generated code, I also stumbled across a couple of things that I fixed in this PR too.
- Sometimes we'd accidentally add an extra space in between things
- Sometimes we'd forget to conjugate the word `Represent`

### DIFF
--- a/slice/Ice/Locator.ice
+++ b/slice/Ice/Locator.ice
@@ -22,7 +22,7 @@ module Ice
 {
     interface Process;
 
-    ///  This exception is raised if an adapter cannot be found.
+    /// This exception is raised if an adapter cannot be found.
     exception AdapterNotFoundException
     {
     }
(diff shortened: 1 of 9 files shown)


slice/Ice/Locator.ice
```diff
@@ -113,7 +113,7 @@ module Ice
     }
 
     /// This interface should be implemented by services implementing the <code>Ice::Locator interface</code>. It should
-    /// be advertised through an Ice object with the identity <code>`Ice/LocatorFinder'</code>. This allows clients to
+    /// be advertised through an Ice object with the identity <code>'Ice/LocatorFinder'</code>. This allows clients to
```
bernardnormier, line 116: Which syntax should be use in doc-comments to insert a string as is? `<code>text</code>` or back ticks?

bernardnormier (approved):

slice/Ice/Locator.ice
InsertCreativityHere: Doxygen seems to support both HTML `<code>` tags and Markdown backticks:
https://www.doxygen.nl/manual/htmlcmds.html
https://www.doxygen.nl/manual/markdown.html#md_codespan

I'm not sure if there's some subtle difference between them, but obviously we should prefer backticks over xml tags if there's no difference.

PR #3197, opened 2024-11-26
ConnectionInfo fixes in C++ and C#

bernardnormier:
The PR:
 - clarifies the internal Transceiver::getInfo operation
 - clears the cached connection info when a new adapter is set on a connection

### DIFF
--- a/cpp/src/Ice/ConnectionI.cpp
+++ b/cpp/src/Ice/ConnectionI.cpp
@@ -1050,6 +1050,9 @@ Ice::ConnectionI::setAdapterFromAdapter(const ObjectAdapterIPtr& adapter)
     }
     assert(adapter); // Called by ObjectAdapterI::setAdapterOnConnection
     _adapter = adapter;
+
+    // Clear cached connection info (if any) as it's no longer accurate.
+    _info = nullptr;
 }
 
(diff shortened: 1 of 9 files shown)


cpp/src/Ice/ConnectionI.cpp
```diff
@@ -3493,17 +3493,19 @@ Ice::ConnectionI::dispatchAll(
...
     }
 
+    bool incoming = !_connector;
+
+    // _adapter is set for an incoming connection until "finished"
     _info = _transceiver->getInfo(
-        _connector == nullptr,
-        _adapter ? _adapter->getName() : string{},
+        incoming,
+        incoming && _adapter ? _adapter->getName() : string{},
```
pepone, line 3506: Can we cache the adapter name in the connection, and always provide it for incoming connections?

pepone (approved):

cpp/src/Ice/ConnectionI.cpp
bernardnormier: We could, but this seems very minor and hard to test. _adapter is usually not null for an incoming connection.

pepone: Ok, the code is clear now that you remove the asserts.

Issue #1085, opened 2020-10-07
Ending '@see' Attributes with Periods


InsertCreativityHere:
Currently, when we use `@see` attributes in Slice doc comments, we don't end them in periods. For example:
```
/// The remote invocation metrics map.
/// @see RemoteMetrics
MetricsMap remotes;
```
The code we generate for Java and C# just copy the `@see ...` directly. This is fine for Java, but causes StyleCop warnings in C# (where it wants these lines to end in a period).

Changing the Slice comments to end in a period fixes the C# warnings, but breaks the JavaDoc generation. When JavaDoc sees a period in an `@see` attribute, it expects a package name, and hence for there to be a name after the period.

Should we leave the Slice files as is, and make the C# compiler auto-generate a terminating period?
Or should we add a period to the Slice comments, and make the Java compiler cut out these trailing periods?

InsertCreativityHere:
I think that once we add validation to doc comments, we should reject `@see` ending with periods, since that's the behavior used by JavaDoc: https://docs.oracle.com/javase/7/docs/technotes/tools/windows/javadoc.html#see

And then, any languages that want these lines to end in periods (`C#`), we can just generate periods at the ends.

InsertCreativityHere:
This was fixed (along with a couple other things) in #3198.

PR #3198, opened 2024-11-26
Added Slice Doc-Comment Validation for Operation Only Tags and `@see` Tags.

InsertCreativityHere:
This PR further improves the validation of doc-comments.

1) It implements #1085 - which is about how ending `@see` with periods breaks things.
Now, if a user ends `@see` with a period, we issue a warning, and remove the trailing period.

2) We enforce that `@param`, `@exception`, `@throws`, and `@return` can only be placed on operations.
If they are placed on anything else, we issue a warning, and throw the tag away (along with it's contents).

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -837,7 +837,7 @@ Slice::Contained::parseDocComment(function<string(string, string)> linkFormatter
             state = StateSee;
 
             // Remove any leading and trailing whitespace from the line.
-            // There's no concern of losing formatting for `@see` due to it's simplicity.
+            // There's no concern of losing formatting for `@see` due to its simplicity.
             lineText = IceInternal::trim(lineText);
             if (lineText.empty())
             {
@@ -846,7 +846,7 @@ Slice::Contained::parseDocComment(function<string(string, string)> linkFormatter
             else
             {
                 // '@see' tags aren't allowed to end with periods.
-                // They do not take sentences, and the trailing period will trip up some languages doc-comments.
+                // They do not take sentences, and the trailing period will trip up some language's doc-comments.
                 if (lineText.back() == '.')
                 {
                     string msg = "ignoring trailing '.' character in '" + seeTag + "' tag";
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -694,36 +694,34 @@ namespace
         if (l.find(tag) == 0)
         {
             const string ws = " \t";
+            string::size_type pos;
 
             if (namedTag)
             {
-                string::size_type n = l.find_first_not_of(ws, tag.size());
-                if (n == string::npos)
+                pos = l.find_first_not_of(ws, tag.size());
+                if (pos == string::npos)
                 {
                     return false; // Malformed line, ignore it.
                 }
-                string::size_type end = l.find_first_of(ws, n);
+                string::size_type end = l.find_first_of(ws, pos);
                 if (end == string::npos)
                 {
                     return false; // Malformed line, ignore it.
                 }
-                name = l.substr(n, end - n);
-                n = l.find_first_not_of(ws, end);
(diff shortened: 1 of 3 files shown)


cpp/src/Slice/Parser.cpp
```diff
@@ -694,36 +694,34 @@ namespace
...
             else
             {
                 name.clear();
+                pos = tag.size();
+            }
 
-                string::size_type n = l.find_first_not_of(ws, tag.size());
-                if (n == string::npos)
-                {
-                    return false; // Malformed line, ignore it.
```
InsertCreativityHere, line 724: Returning `false` if the line was empty wasn't good behavior.

It would cause the comment parser to think this was unknown tag, not a malformed tag.
And seeing `unknown doc comment tag '@see'` is very confusing.

There wasn't even a point to returning false, since all the calling code gracefully handles empty lines anyways.

bernardnormier (approved):

cpp/src/Slice/Parser.cpp
```diff
@@ -791,7 +800,14 @@ Slice::Contained::parseDocComment(function<string(string, string)> linkFormatter
...
         {
-            if (!lineText.empty())
+            if (!isOperation)
+            {
+                // If '@throws' was put on anything other than an operation, ignore it and issue a warning.
+                string msg = "the '" + throwsTag + "' tag is only valid on operations";
+                unit()->warning(file(), line(), InvalidComment, msg);
+                state = StateUnknown;
+            }
+            else if (!lineText.empty())
```
pepone, line 810: Should we emit a warning for empty throws?

cpp/src/Slice/Parser.cpp
```diff
@@ -811,15 +834,38 @@ Slice::Contained::parseDocComment(function<string(string, string)> linkFormatter
...
         {
-            if (!lineText.empty())
+            if (!isOperation)
+            {
+                // If '@return' was put on anything other than an operation, ignore it and issue a warning.
+                string msg = "the '" + returnTag + "' tag is only valid on operations";
+                unit()->warning(file(), line(), InvalidComment, msg);
+                state = StateUnknown;
+            }
+            else if (!lineText.empty())
```
pepone, line 868: I think we can have a warning for empty return tag.

cpp/src/Slice/Parser.cpp
```diff
@@ -781,7 +783,14 @@ Slice::Contained::parseDocComment(function<string(string, string)> linkFormatter
...
         {
-            if (!lineText.empty())
+            if (!isOperation)
+            {
+                // If '@param' was put on anything other than an operation, ignore it and issue a warning.
+                string msg = "the '" + paramTag + "' tag is only valid on operations";
+                unit()->warning(file(), line(), InvalidComment, msg);
+                state = StateUnknown;
+            }
+            else if (!lineText.empty())
```
pepone, line 793: I think it would make sense to emit a warning if lineText is empty. 

Do we validate that the `@param` names, match the operation param names?

pepone (approved):

cpp/src/Slice/Parser.cpp
InsertCreativityHere: The idea behind this is for users who format their comments like:
```
@throw AlreadyActivatedException
This exception is thrown if the thing is already activated
```
Just because the rest of the line is empty, doesn't mean that the section as a whole is empty.
It could just be on the next line down.

InsertCreativityHere: Nope, but I see what you were thinking now. `lineText` only includes the text that comes _after_ the name of the exception, parameter, type, etc.

The function we call is
`bool parseCommentLine(const string& l, const string& tag, bool namedTag, string& name, string& doc)`
and it has 2 'out parameters' `name` and `doc`. In this case `name` will be the name of the exception, and `doc` (which is `lineText` here) is anything that comes after that name.

PR #3199, opened 2024-11-26
Move IAP EndpointInfo and ConnectionInfo

bernardnormier:
This PR moves the IAP EndpointInfo and ConnectionInfo definitions to namespace Ice.

It also makes them available all the time - not just for iOS. These are tiny classes, there is no real benefit in ifdefing them out.


cpp/include/Ice/Connection.h
```diff
@@ -431,6 +431,68 @@ namespace Ice
         {
         }
     };
+
+    /**
+     * Provides access to the connection details of an IAP connection
+     * \headerfile Ice/Ice.h
+     */
+    class IAPConnectionInfo final : public ConnectionInfo
```
bernardnormier, line 439: I didn't update the Swift code yet even though this renaming makes an update necessary.

I am curious to see if the Swift CI will fail.

cpp/src/Ice/ios/iAPTransceiver.mm
```diff
@@ -383,9 +383,9 @@ - (void)stream:(NSStream*)stream handleEvent:(NSStreamEvent)eventCode
 IceObjC::iAPTransceiver::getInfo(bool incoming, string adapterName, string connectionId) const
 {
     assert(!incoming);
-    assert(adapterName.empty());
```
bernardnormier, line 386: See #3197.

pepone (approved):

cpp/include/Ice/Connection.h
pepone: Seems we are not building for iOS, there is some commented out code in the ci workflow:

https://github.com/zeroc-ice/ice/blob/8c76edbd3441d83394995c7a08761ce707966ed5/.github/workflows/ci.yml#L55-L62

PR #3195, opened 2024-11-26
Refactor EndpointInfo in Swift

bernardnormier:
This PR converts the Swift EndpointInfo into classes like in C++ and C#.

PR #3207, opened 2024-11-27
Refactor ConnectionInfo in Swift

bernardnormier:
This is a follow-up to #3195.


swift/src/Ice/Connection.swift
```diff
@@ -169,40 +157,176 @@ public protocol Connection: AnyObject, CustomStringConvertible {
...
+        var raw = peerCertificate
+        if raw.hasPrefix(beginPrefix) {
+            raw = String(raw.dropFirst(beginPrefix.count))
+            raw = String(raw.dropLast(endPrefix.count))
+        }
+
+        var peerCertificate: SecCertificate? = nil
+        if let data = NSData(base64Encoded: raw, options: .ignoreUnknownCharacters) {
+            if let cert = SecCertificateCreateWithData(kCFAllocatorDefault, data) {
+                peerCertificate = cert
```
pepone, line 253: Can't we just assign `self.peerCertificate` directly? the `var peerCertificate:` shadows the parameter, which is kind of confusing.

pepone (approved):

swift/src/Ice/Connection.swift
bernardnormier: We can only assign `self.peerCertificate` once, hence this helper.

An alternative would be to move this decoding to the factory.

bernardnormier: Fixed

externl (approved):

PR #3209, opened 2024-11-27
Rework EndpointInfo and ConnectionInfo in MATLAB

bernardnormier:
This PR updates EndpointInfo and ConnectionInfo in MATLAB.


matlab/src/Connection.cpp
```diff
@@ -17,8 +17,6 @@ namespace
     {
         Type = 0,
         Underlying,
-        Incoming,
```
bernardnormier, line 20: We only support outgoing connections with MATLAB. And we don't support "bidir" with MATLAB.

matlab/src/Connection.cpp
```diff
@@ -52,74 +48,74 @@ namespace
...
         // Some fields will be left unused.
-        //
 
         mwSize dims[2] = {1, 1};
         auto r = mxCreateStructArray(2, dims, Field::NumFields, infoFields);
-        mxSetFieldByNumber(r, 0, Field::Incoming, createBool(info->incoming));
-        mxSetFieldByNumber(r, 0, Field::AdapterName, createStringFromUTF8(info->adapterName));
         mxSetFieldByNumber(r, 0, Field::ConnectionId, createStringFromUTF8(info->connectionId));
 
+        string type = "other";
```
bernardnormier, line 58: I kept the same logic, just rearranged the code to be more logical.

matlab/lib/+Ice/ConnectionInfo.m
```diff
@@ -2,43 +2,30 @@
...
-            obj.incoming = incoming;
-            obj.adapterName = adapterName;
-            obj.connectionId = connectionId;
-        end
-    end
-    properties(SetAccess=private)
-        % underlying   The information of the underyling transport or null
-        %   if there's no underlying transport.
+    properties(SetAccess=immutable)
+        % underlying   The information of the underlying transport or null if there's no underlying transport.
```
pepone, line 12: For a separe PR:

You didn't update this comment but, seems that `null` isn't the correct term in MATLAB. The other comment above "or an empty array " seems more appropriate.  

Having two comments for the property seems unnecessary, the comment for the property should be placed with the properties section, see: https://www.mathworks.com/help/matlab/matlab_prog/create-help-for-classes.html

matlab/lib/+Ice/EndpointInfo.m
```diff
@@ -61,26 +47,33 @@
...
-                r = obj.secure_;
+                r = false;
             end
         end
     end
-    properties(SetAccess=private)
-        % underlying   The information of the underyling endpoint or an empty
-        %   array if there's no underlying endpoint.
+    properties(SetAccess=immutable)
+        % underlying   The information of the underlying endpoint or an empty array if there's no underlying endpoint.
```
pepone, line 55: For a follow up PR:

If you look at the example doc comments from https://www.mathworks.com/help/matlab/matlab_prog/create-help-for-classes.html 

They don't have the property name at the start.

pepone (approved):

matlab/lib/+Ice/+SSL/ConnectionInfo.m
```diff
@@ -1,28 +1,21 @@
...
-            if nargin == 0
-                underlying = [];
-                incoming = false;
-                adapterName = '';
-                connectionId = '';
-                peerCertificate = '';
-            end
-            obj@Ice.ConnectionInfo(underlying, incoming, adapterName, connectionId);
+        function obj = ConnectionInfo(underlying, peerCertificate)
+            assert(nargin == 2, 'Invalid number of arguments');
```
InsertCreativityHere, line 13: I don't think that this is correct.
MATLAB does not support multiple constructors, there can only be one constructor per class.
And that constructor must always support being called with "default constructor" semantics
(i.e. it must specifically support being called with 0 arguments).

See: https://www.mathworks.com/help/matlab/matlab_oop/class-constructor-methods.html#brd2m9e-3
> If you create a class constructor, ensure it can be called with no input arguments.

----

Note that if you do not provide a custom constructor (as we do here), MATLAB will implicitly 'generate' a default constructor for you.

InsertCreativityHere: I think what we should do instead is keep the old `if nargin == 0` logic,
but add an `else` block with the `assert nargin == N` in it.

This would ensure that the constructor can only be called with `0` arguments (default constructor) or with all `N` arguments (one-short constructor), but cannot be called with only a subset.

bernardnormier: Since all these Info classes are created only from the C++ code, I don't see how a default constructor would ever be called.

InsertCreativityHere (approved):

PR #3211, opened 2024-11-28
Small InputStream cleanup

bernardnormier:
in all languages.


cpp/src/Ice/InputStream.cpp
```diff
@@ -1294,40 +1295,25 @@ Ice::InputStream::resolveCompactId(int id) const
...
-        }
-    }
-    catch (...)
-    {
-        if (_logger)
-        {
-            Warning out(_logger);
-            out << "unknown exception raised by ice_postUnmarshal";
-        }
+        typeId = IceInternal::factoryTable->getTypeId(id);
```
pepone, line 1300: Why is this new lookup required?

pepone: I see this was previously in `readInstance`

pepone (approved):

PR #3212, opened 2024-11-29
DataStorm comments & minor fixes

pepone:
This PR includes additional fixes for DataStorm. Mostly doc comments and style fixes to make the code easier to follow.

### DIFF
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -18,14 +18,14 @@ namespace
 {
     DataSample toSample(const shared_ptr<Sample>& sample, const Ice::CommunicatorPtr& communicator, bool marshalKey)
     {
-        return {
-            sample->id,
-            marshalKey ? 0 : sample->key->getId(),
-            marshalKey ? sample->key->encode(communicator) : Ice::ByteSeq{},
-            chrono::time_point_cast<chrono::microseconds>(sample->timestamp).time_since_epoch().count(),
-            sample->tag ? sample->tag->getId() : 0,
-            sample->event,
-            sample->encode(communicator)};
+        return DataSample{
+            .id = sample->id,
+            .keyId = marshalKey ? 0 : sample->key->getId(),
+            .keyValue = marshalKey ? sample->key->encode(communicator) : Ice::ByteSeq{},
+            .timestamp = chrono::time_point_cast<chrono::microseconds>(sample->timestamp).time_since_epoch().count(),
+            .tag = sample->tag ? sample->tag->getId() : 0,
+            .event = sample->event,
+            .value = sample->encode(communicator)};
     }
 
     void cleanOldSamples(
@@ -111,8 +111,10 @@ DataElementI::attach(
         auto info = *data.config->sampleFilter;
         sampleFilter = _parent->getSampleFilterFactories()->decode(getCommunicator(), info.name, info.criteria);
     }
+
     string facet = data.config->facet ? *data.config->facet : string();
     int priority = data.config->priority ? *data.config->priority : 0;
+
     string name;
     if (data.config->name)
     {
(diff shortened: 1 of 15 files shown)


cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -18,14 +18,14 @@ namespace
...
-            sample->event,
-            sample->encode(communicator)};
+        return DataSample{
+            .id = sample->id,
+            .keyId = marshalKey ? 0 : sample->key->getId(),
+            .keyValue = marshalKey ? sample->key->encode(communicator) : Ice::ByteSeq{},
+            .timestamp = chrono::time_point_cast<chrono::microseconds>(sample->timestamp).time_since_epoch().count(),
+            .tag = sample->tag ? sample->tag->getId() : 0,
+            .event = sample->event,
+            .value = sample->encode(communicator)};
```
pepone, line 28: For large structs with many fields this style, using the filed names, seems easier to read.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -124,14 +126,20 @@ DataElementI::attach(
...
-        LongLongDict lastIds = key ? session->getLastIds(topicId, id, shared_from_this()) : LongLongDict();
+        int64_t lastId = q != data.lastIds.end() ? q->second : 0;
+        LongLongDict lastIds = key ? session->getLastIds(topicId, id, shared_from_this()) : LongLongDict{};
         DataSamples samples = getSamples(key, sampleFilter, data.config, lastId, now);
-        acks.push_back({_id, _config, lastIds, samples.samples, data.id});
+        acks.push_back(ElementDataAck{
+            .id = _id,
+            .config = _config,
+            .lastIds = std::move(lastIds),
+            .samples = std::move(samples.samples),
```
pepone, line 141: We can move them, don't need to copy these containers.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -178,7 +189,7 @@ DataElementI::attach(
         session->subscriberInitialized(topicId, id > 0 ? data.id : -data.id, data.samples, key, shared_from_this());
     if (!samplesI.empty())
     {
-        return [=, self = shared_from_this()]()
+        return [=, samplesI = std::move(samplesI), self = shared_from_this()]()
```
pepone, line 192: My understanding is that the previous code, captures `samplesI` as a copy, which is them passed by reference to initSamples.

I updated it to move capture it because it was not used anywhere else, but maybe is simpler to just capture it by reference.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -634,7 +648,7 @@ DataElementI::forward(const Ice::ByteSeq& inParams, const Ice::Current& current)
 {
     for (const auto& [_, listener] : _listeners)
     {
-        // If there's at least one subscriber interested in the update
+        // If we are forwarding a sample check if at least once of the listeners is interested in the sample.
         if (!_sample || listener.matchOne(_sample, false))
```
pepone, line 652: The `!_sample` check is there because this also forwards other operations. `_sample` is only set for `SubscriberSession::s` when sending a sample.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -216,7 +218,7 @@ SessionI::attachTags(int64_t topicId, ElementInfoSeq tags, bool initialize, cons
 
     runWithTopics(
```
pepone, line 219: We have two overloads of `runWithTopics` which are almost identical, only differing in the passed callback, one has an extra `TopicSubscribers` parameter that was never used. I got read of that overload and updated the callbacks to use the other one.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -1134,15 +1138,14 @@ SessionI::runWithTopics(
...
+        unique_lock<mutex> lock(topic->getMutex());
         if (topic->isDestroyed())
         {
             continue;
         }
-        _topicLock = &l;
+        _topicLock = &lock;
         fn(topic);
         _topicLock = nullptr;
-        l.unlock();
```
pepone, line 1145: No need for the explicit unlock here, the lock is being destroyed as it is about to go out of scope.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -1222,15 +1202,16 @@ SubscriberSessionI::getTopics(const string& name) const
 }
 
 void
-SubscriberSessionI::s(int64_t topicId, int64_t elementId, DataSample s, const Ice::Current& current)
+SubscriberSessionI::s(int64_t topicId, int64_t elementId, DataSample dataSample, const Ice::Current& current)
 {
```
pepone, line 1206: I kept the same logic, just renames parameters, and loop variables to make things more readable.

externl (approved):

PR #3218, opened 2024-12-02
Remove -c option from udp endpoints in C#

bernardnormier:
This PR from the undocumented "-c" option from udp endpoints in C#.

This was a server-endpoint option. For client option, the behavior is always "connect".

### DIFF
--- a/csharp/src/Ice/Internal/UdpTransceiver.cs
+++ b/csharp/src/Ice/Internal/UdpTransceiver.cs
@@ -270,23 +270,7 @@ internal sealed class UdpTransceiver : Transceiver
             throw new Ice.ConnectionLostException();
         }
 
-        if (_state == StateNeedConnect)
-        {
-            Debug.Assert(_incoming);
-
-            //
-            // If we must connect, then we connect to the first peer that sends us a packet.
-            //
-            bool connected = Network.doConnect(_fd, _peerAddr, null);
-            Debug.Assert(connected);
-            _state = StateConnected; // We're connected now
-
-            if (_instance.traceLevel() >= 1)
-            {
-                string s = "connected " + protocol() + " socket\n" + ToString();
-                _instance.logger().trace(_instance.traceCategory(), s);
-            }
-        }
+        Debug.Assert(_state != StateNeedConnect);
 
         buf.resize(ret, true);
         buf.b.position(ret);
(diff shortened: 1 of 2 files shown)


csharp/src/Ice/Internal/UdpTransceiver.cs
```diff
@@ -726,12 +692,11 @@ internal UdpTransceiver(
...
         string host,
         int port,
-        string mcastInterface,
-        bool connect)
+        string mcastInterface)
     {
         _endpoint = endpoint;
         _instance = instance;
-        _state = connect ? StateNeedConnect : StateNotConnected;
+        _state = StateNotConnected;
```
bernardnormier, line 699: A server transceiver (incoming = true) starts in the StateNotConnected state and remains in that state.
A client transceiver (incoming = false) starts in the StateNeedConnect state and goes up to StateConnected. It never reaches StateNotConnected.

csharp/src/Ice/Internal/UdpTransceiver.cs
```diff
@@ -554,8 +518,9 @@ public ConnectionInfo getInfo(bool incoming, string adapterName, string connecti
         {
             EndPoint localEndpoint = Network.getLocalAddress(_fd);
 
-            if (_state == StateNotConnected) // TODO: eliminate all these states
+            if (_state == StateNotConnected) // a server endpoint
             {
+                Debug.Assert(incoming);
```
bernardnormier, line 523: Question: do we create a new Connection and potentially ConnectionInfo for each incoming UDP request?

pepone: Looking at the IncomingConnectionFactory I believe there is a single udp Connection created by the factory constructor. But then caching the info object wouldn't be correct.

pepone (approved):

externl (approved):

csharp/src/Ice/Internal/UdpTransceiver.cs
bernardnormier: I believe we create a new connection for each request.  And udp bi-dir relies on these per-request connections:
https://github.com/zeroc-ice/ice/blob/41307f90a6d1ebe097f3aef5b5b5e802dd78f543/csharp/test/Ice/udp/AllTests.cs#L200

PR #3219, opened 2024-12-03
Refactor JavaScript EnpointInfo & ConnectionInfo

pepone:
Aligns the JavaScript ConnectionInfo and EndpointInfo classes, with C#, Java, ...

### DIFF
--- a/js/src/Ice/Connection.d.ts
+++ b/js/src/Ice/Connection.d.ts
@@ -133,11 +133,6 @@ declare module "ice" {
             get underlying(): Ice.ConnectionInfo | null;
 
             /**
-             * Indicates whether the connection is incoming (`true`) or outgoing (`false`).
-             */
-            get incoming(): boolean;
-
-            /**
              * The name of the adapter associated with this connection.
              */
             get adapterName(): string;
--- a/js/src/Ice/Connection.d.ts
+++ b/js/src/Ice/Connection.d.ts
@@ -130,22 +130,22 @@ declare module "ice" {
             /**
              * The information of the underlying transport, or `null` if there is no underlying transport.
              */
-            underlying: Ice.ConnectionInfo;
+            get underlying(): Ice.ConnectionInfo | null;
 
             /**
              * Indicates whether the connection is incoming (`true`) or outgoing (`false`).
              */
-            incoming: boolean;
+            get incoming(): boolean;
 
             /**
              * The name of the adapter associated with this connection.
              */
-            adapterName: string;
+            get adapterName(): string;
 
             /**
              * The connection id.
              */
-            connectionId: string;
+            get connectionId(): string;
         }
 
         /**
@@ -155,22 +155,22 @@ declare module "ice" {
             /**
              * The local address.
              */
-            localAddress: string;
+            get localAddress(): string;
 
             /**
              * The local port.
              */
-            localPort: number;
+            get localPort(): number;
 
             /**
              * The remote address.
              */
-            remoteAddress: string;
+            get remoteAddress(): string;
 
             /**
              * The remote port.
              */
-            remotePort: number;
+            get remotePort(): number;
         }
 
         /**
@@ -180,12 +180,12 @@ declare module "ice" {
             /**
              * The connection buffer receive size.
              */
-            rcvSize: number;
+            get rcvSize(): number;
 
             /**
              * The connection buffer send size.
              */
-            sndSize: number;
+            get sndSize(): number;
         }
 
         /**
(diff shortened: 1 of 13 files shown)


js/src/Ice/Connection.d.ts
```diff
@@ -222,7 +222,7 @@ declare module "ice" {
             /**
              * The headers from the HTTP upgrade request.
              */
-            headers: HeaderDict;
+            get headers(): HeaderDict;
```
bernardnormier, line 225: The HeaderDictHelper sounds like a left over from when this code was generated from local Slice.

js/src/Ice/Connection.js
```diff
@@ -71,8 +120,12 @@ export const [HeaderDict, HeaderDictHelper] = defineDictionary(
  *  Provides access to the connection details of a WebSocket connection
  **/
 export class WSConnectionInfo extends ConnectionInfo {
```
bernardnormier, line 122: I suspect we don't need the defineDictionary above.

js/src/Ice/TcpEndpointI.js
```diff
@@ -7,33 +7,36 @@ import { StringUtil } from "./StringUtil.js";
...
 
         if (this.secure()) {
-            const sslInfo = new SSLEndpointInfo();
-            sslInfo.underlying = info;
-            sslInfo.timeout = info.timeout;
-            sslInfo.compress = info.compress;
-            return sslInfo;
-        } else {
-            return info;
+            info = new SSLEndpointInfo(info);
```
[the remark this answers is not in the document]
pepone: It is tested with the browser WSS protocol. In this case is used as the underlying endpoint info for WSSEndpointInfo.

The JavaScript SSLEndpointInfo doesn't have additional fields because they are not accessible with the WebSocket API.

In JavaScript for SSLEndpointInfo, and SSLConnectionInfo we just use the constructor inherited from Ice.EndpointInfo and create them with an underlying.

js/src/Ice/WSTransceiver.js
```diff
@@ -240,29 +240,26 @@ if (typeof WebSocket !== "undefined") {
...
-                sslInfo.compress = tcpInfo.compress;
-                info.underlying = sslInfo;
-            } else {
-                info.underlying = tcpInfo;
+                info = new SSLConnectionInfo(info);
             }
 
-            info.headers = {};
-            return info;
+            return new WSConnectionInfo(info, {});
```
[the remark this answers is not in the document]
pepone: Yes, I would remove it. The JavaScript WebSocket API doesn't provide access to the headers.

js/src/Ice/Connection.js
pepone: Removed.

js/src/Ice/TcpTransceiver.js
```diff
@@ -233,16 +233,19 @@ if (typeof net.createConnection === "function") {
...
-            return info;
+            return new TCPConnectionInfo(
+                incoming,
+                adapterName,
+                connectionId,
+                this._fd.localAddress,
+                this._fd.localPort,
+                this._fd.remoteAddress,
+                this._fd.remotePort,
+                -1,
```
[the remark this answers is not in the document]
pepone: This field is not used in JavaScript, and always has this default value. Seems like we should remove it too.

js/src/Ice/Connection.js
```diff
@@ -44,35 +65,24 @@ export class IPConnectionInfo extends ConnectionInfo {
...
-        underlying,
-        incoming,
         adapterName,
         connectionId,
-        localAddress,
-        localPort,
-        remoteAddress,
-        remotePort,
-        rcvSize = 0,
+        localAddress = "",
```
bernardnormier, line 70: I would remove all the default param values. I don't think we ever use them in JS.

js/src/Ice/Connection.js
```diff
@@ -2,40 +2,61 @@
...
-        localPort = -1,
-        remoteAddress = "",
-        remotePort = -1,
-    ) {
-        super(underlying, incoming, adapterName, connectionId);
-        this.localAddress = localAddress;
-        this.localPort = localPort;
-        this.remoteAddress = remoteAddress;
-        this.remotePort = remotePort;
+    constructor(adapterName, connectionId, localAddress = "", localPort = -1, remoteAddress = "", remotePort = -1) {
```
bernardnormier, line 38: Same here: remove the default values.

bernardnormier (approved):
Looks good.

js/src/Ice/TcpEndpointI.js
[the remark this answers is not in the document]
externl: We should look into adding browsers testing with CI

externl (approved):
Looks good to me!

Issue #3216, opened 2024-12-02
Cleanup How the Slice Compiler Stores Metadata


InsertCreativityHere:
There are only 2 things in the Slice language which can be forward declared: `interface`s and `class`es.

Slice allows you to put metadata on both an interface's definition and an interface's forward declaration:
```
[okay]
interface Hello;

[also-okay]
interface Hello { ... }
```

It is also permissible to have multiple forward declarations too:
```
[fine]
interface Hello;

[also-fine]
interface Hello;
```

But there's no clear guidance on how the Slice compiler should behave in these cases.
And as a result, there's some gaps in how we handle it. But first, it's important to mention how this works at the parser level.
I'm going to talk about interfaces, but everything I say is also true for classes.

----

The two types which matter here are `InterfaceDecl` (a forward declaration) and `InterfaceDef` (a full definition).

You might expect that if you have 4 forward declarations for the same interface, we'd create 4 `InterfaceDecl`s, but this is not true.
The parser only creates an `InterfaceDecl` for the _first_ forward declaration. If you re-forward-declare that interface again, instead of making a new `InterfaceDecl`, the parser will return the already-existent one (re-using it).
But, because of how the parser sets metadata, it's metadata will be over-written each time we hit a forward declaration.
So, in practice, only the metadata you put on the _last_ forward declaration is meaningful. Any other metadata will be successively over-written by the parser.

Additionally, when you _define_ an interface, we will obviously create an `InterfaceDef`, but we'll also create a corresponding `InterfaceDecl` too (again, unless one already exists, then we re-use it).
And while it's true that there is some coupling here (`InterfaceDef` holds a pointer to it's corresponding `InterfaceDecl`), _metadata_ is not shared. So, putting metadata on a forward declaration vs the definition is semantically different.

----

This setup causes two problems:
1) From the point-of-view of the code-gen, it's not clear when/if it should be checking for metadata on the `InterfaceDef` vs the `InterfaceDecl`. Right now, the code-gen seems ignorant of this distinction, and just blindly chooses one, ignoring the other.

2) If you do put metadata on your forward declarations, it's only meaningful for the _last_ forward declaration.
This is not obvious, and even worse, depends completely on your `#include` ordering.

----

To fix this, I propose the following changes.

1) Eliminate the distinction between putting metadata on a declaration vs a definition.
Specifically, I think we should remove the metadata from the `ClassDef` and `InterfaceDef` types.
[...]

PR #3224, opened 2024-12-03
Simplified how Metadata is Stored on Types Which Support Forward Declaration

### LINKED ISSUE #3039: Handling of deprecated in Slice library
The new Slice/deprecated test (see #3038) shows two issues in the Slice library (aka Slice parser).

- `getDeprecationReason` returns an empty string and not `nullopt` for Slice metadata `["deprecated"]`
-  when we define a Slice class or interface (Def) without an existing Decl, the deprecated metadata for the Def is not copied into the Decl created by the Parser

I think these two issues are related. It could be cleaner for the Parser to consume the deprecated Metadata and provide direct support for deprecated--and not keep it as a generic metadata to be consumed by the Slice compilers.

InsertCreativityHere:
The main focus of this PR is #3216 (cleaning up how metadata is stored)
but it also fixes #3039 (which was specifically about how `deprecated` is handled between `Decl` and `Def`).

----

1) **Added a new `appendMetadata` function.**
So now we have `setMetadata` which overwrites existing metadata with a new list (_used by the validator where we want to throw away bad metadata_) and `appendMetadata` which just appends metadata to a Slice entity without touching any that's already present (_used by the parser to add metadata to entities_).
This fixes the problem where forward-declarations were constantly overwriting their own metadata.

2) **All class/interface metadata is now stored exclusively on the `Decl` types, and never the `Def` types.**
Previously, both of them had their own distinct lists of metadata. Note that the `Def` types still support the same API for metadata, but these functions just forward to the 'underlying' `Decl` type. This way the code-gen never has to worry about whether it should be using `Decl` or `Def`, both will work just fine.

I also removed all of the `visitClassDef`/`visitInterfaceDef` metadata validation functions.
It is sufficient to only visit the `Decl` types now.

### DIFF
--- a/cpp/src/Slice/MetadataValidation.cpp
+++ b/cpp/src/Slice/MetadataValidation.cpp
@@ -19,12 +19,13 @@ namespace
     public:
         MetadataVisitor(string_view language, map<string, MetadataInfo> knownMetadata);
 
+        // We don't visit `ClassDef` and `InterfaceDef` because their metadata is always stored on their corresponding
+        // `ClassDecl` and `InterfaceDecl` types. So just visiting the `Decl` types is sufficient to check everything.
+
         bool visitUnitStart(const UnitPtr&) final;
         bool visitModuleStart(const ModulePtr&) final;
         void visitClassDecl(const ClassDeclPtr&) final;
-        bool visitClassDefStart(const ClassDefPtr&) final;
         void visitInterfaceDecl(const InterfaceDeclPtr&) final;
-        bool visitInterfaceDefStart(const InterfaceDefPtr&) final;
         bool visitExceptionStart(const ExceptionPtr&) final;
         bool visitStructStart(const StructPtr&) final;
         void visitOperation(const OperationPtr&) final;
@@ -40,7 +41,7 @@ namespace
 
         string_view _language;
         map<string, MetadataInfo> _knownMetadata;
-        set<string> _seenDirectives;
+        map<string, MetadataPtr> _seenDirectives;
     };
 }
 
@@ -73,7 +74,7 @@ Slice::validateMetadata(const UnitPtr& p, string_view prefix, map<string, Metada
 
     // "amd"
     MetadataInfo amdInfo = {
-        {typeid(InterfaceDef), typeid(Operation)},
+        {typeid(InterfaceDecl), typeid(Operation)},
         MetadataArgumentKind::NoArguments,
     };
     knownMetadata.emplace("amd", std::move(amdInfo));
@@ -81,9 +82,7 @@ Slice::validateMetadata(const UnitPtr& p, string_view prefix, map<string, Metada
     // "deprecated"
     MetadataInfo deprecatedInfo = {
         {typeid(InterfaceDecl),
-         typeid(InterfaceDef),
          typeid(ClassDecl),
-         typeid(ClassDef),
          typeid(Operation),
          typeid(Exception),
          typeid(Struct),
@@ -95,11 +94,12 @@ Slice::validateMetadata(const UnitPtr& p, string_view prefix, map<string, Metada
          typeid(DataMember)},
         MetadataArgumentKind::OptionalTextArgument,
     };
+    knownMetadata.emplace("deprecate", std::move(deprecatedInfo)); // Kept as an alias for 'deprecated'.
     knownMetadata.emplace("deprecated", std::move(deprecatedInfo));
 
     // "format"
     MetadataInfo formatInfo = {
-        {typeid(InterfaceDef), typeid(Operation)},
+        {typeid(InterfaceDecl), typeid(Operation)},
         MetadataArgumentKind::SingleArgument,
         {{"compact", "sliced", "default"}},
     };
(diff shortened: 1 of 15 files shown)


cpp/src/Slice/MetadataValidation.cpp
```diff
@@ -95,26 +94,27 @@ Slice::validateMetadata(const UnitPtr& p, string_view prefix, map<string, Metada
          typeid(DataMember)},
         MetadataArgumentKind::OptionalTextArgument,
     };
+    knownMetadata.emplace("deprecate", std::move(deprecatedInfo)); // Kept as an alias for 'deprecated'.
```
InsertCreativityHere, line 97: I forgot to add this in my original metadata validation PR.
We don't actually use it in our Slice files, but we decided to keep support for the `3.7` name for backwards compatibility. We don't want to bother our users about adding a single `d` character to their metadata.

cpp/src/Slice/MetadataValidation.cpp
```diff
@@ -73,17 +74,15 @@ Slice::validateMetadata(const UnitPtr& p, string_view prefix, map<string, Metada
 
     // "amd"
     MetadataInfo amdInfo = {
-        {typeid(InterfaceDef), typeid(Operation)},
+        {typeid(InterfaceDecl), typeid(Operation)},
```
InsertCreativityHere, line 77: Now that all metadata is on the `Decl` types, these are the types that should be passed into the validator.
I added a note in the header file mentioning this fact.

cpp/src/Slice/Parser.cpp
```diff
@@ -1023,16 +1036,17 @@ Slice::Contained::isDeprecated() const
 optional<string>
 Slice::Contained::getDeprecationReason() const
 {
+    string reasonMessage;
```
InsertCreativityHere, line 1039: This is unrelated to the main goal of this PR, but is small, and fixes #3039.

If no message was supplied to `deprecated` metadata, this should return `nullopt`, not `""`.
And now it does!

cpp/src/Slice/Parser.cpp
```diff
@@ -1178,23 +1192,15 @@ Slice::Container::createClassDef(const string& name, int id, const ClassDefPtr&
     }
 
     ClassDefPtr def = make_shared<ClassDef>(shared_from_this(), name, id, base);
-    _unit->addContent(def);
```
InsertCreativityHere, line 1181: Whenever you define a class, we always make sure there's a corresponding `Decl`. If there isn't, we create one here.

Previously, we would add the `Def` to the module first, then add the `Decl` 2nd.
With my changes though, this causes metadata warnings to be emitted in a lame order.

I swapped the order, so now we add the `Decl` first, and then add the `Def` afterwards.
This fixes the order, and also... it's just simpler.

cpp/src/Slice/Parser.cpp
```diff
@@ -1330,23 +1336,15 @@ Slice::Container::createInterfaceDef(const string& name, const InterfaceList& ba
     InterfaceDecl::checkBasesAreLegal(name, bases, _unit);
 
     InterfaceDefPtr def = make_shared<InterfaceDef>(shared_from_this(), name, bases);
-    _unit->addContent(def);
```
InsertCreativityHere, line 1333: Same for interface. Now we add the `Decl` before we add the `Def`.
This is more expected from the POV of the visitors.

cpp/src/Slice/Parser.cpp
```diff
@@ -2695,6 +2693,24 @@ Slice::ClassDef::compactId() const
     return _compactId;
 }
 
+MetadataList
```
InsertCreativityHere, line 2696: The metadata API on the `Def` types just always forward to their corresponding `Decl`.

cpp/src/Slice/Parser.cpp
```diff
@@ -4713,10 +4747,7 @@ Slice::Unit::addFileMetadata(MetadataList metadata)
     }
     else
     {
```
InsertCreativityHere, line 4749: Might as well use the new `appendMetadata` API.

cpp/src/Slice/Parser.cpp
```diff
@@ -214,10 +214,17 @@ Slice::DefinitionContext::setMetadata(MetadataList metadata)
     initSuppressedWarnings();
 }
 
+void
+Slice::DefinitionContext::appendMetadata(MetadataList metadata)
+{
+    _metadata.splice(_metadata.end(), metadata);
```
bernardnormier, line 220: missing std::move

cpp/src/Slice/Parser.cpp
```diff
@@ -965,10 +972,16 @@ Slice::Contained::setMetadata(MetadataList metadata)
     _metadata = metadata;
 }
 
+void
+Slice::Contained::appendMetadata(MetadataList metadata)
+{
+    _metadata.splice(_metadata.end(), metadata);
```
bernardnormier, line 978: missing std::move

cpp/src/Slice/Parser.cpp
```diff
@@ -4713,10 +4747,7 @@ Slice::Unit::addFileMetadata(MetadataList metadata)
     }
     else
     {
-        // Append the file metadata to any existing metadata (e.g., default file metadata).
-        MetadataList l = dc->getMetadata();
-        move(metadata.begin(), metadata.end(), back_inserter(l));
-        dc->setMetadata(l);
+        dc->appendMetadata(metadata);
```
bernardnormier, line 4750: std::move

bernardnormier (approved):

externl (approved):

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Fixed.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Fixed.

PR #3226, opened 2024-12-03
Fix udp test in C#

bernardnormier:
This PR fixes the UDP test in C#. It was not running correctly and yet succeeding.

### DIFF
--- a/csharp/test/Ice/udp/AllTests.cs
+++ b/csharp/test/Ice/udp/AllTests.cs
@@ -2,222 +2,221 @@
 
 using System.Text;
 
-namespace Ice
+namespace Ice.udp;
+
+public class AllTests : global::Test.AllTests
 {
-    namespace udp
+    public class PingReplyI : Test.PingReplyDisp_
     {
-        public class AllTests : global::Test.AllTests
+        public override void reply(Ice.Current current)
         {
-            public class PingReplyI : Test.PingReplyDisp_
+            lock (this)
             {
-                public override void reply(Ice.Current current)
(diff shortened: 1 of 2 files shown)


csharp/test/Ice/udp/Client.cs
```diff
@@ -1,42 +1,37 @@
...
-        public class Client : global::Test.TestHelper
-        {
-            public override void run(string[] args)
-            {
-                var properties = createTestProperties(ref args);
-                properties.setProperty("Ice.Warn.Connections", "0");
-                properties.setProperty("Ice.UDP.SndSize", "16384");
-                using (var communicator = initialize(properties))
-                {
-                    AllTests.allTests(this);
```
bernardnormier, line 16: That was the root bug. allTests is async but we didn't await it since returns void.

csharp/test/Ice/udp/AllTests.cs
```diff
@@ -2,222 +2,221 @@
...
+            obj.ping(reply);
+            obj.ping(reply);
+            obj.ping(reply);
+            ret = replyI.waitReply(3, 2000);
+            if (ret)
+            {
+                break; // Success
             }
 
-            public static async void allTests(global::Test.TestHelper helper)
```
bernardnormier, line 54: `async void` = bad

csharp/test/Ice/udp/Client.cs
bernardnormier: Everything else is reformatting!

PR #3231, opened 2024-12-04
DataStorm comments and bug fixes

pepone:
This PR adds more comments to DataStorm implementation. It includes some updates to the port from the original datastorm implementation.

### DIFF
--- a/cpp/src/DataStorm/Contract.ice
+++ b/cpp/src/DataStorm/Contract.ice
@@ -240,7 +240,7 @@ module DataStormContract
     {
         /// Announces new and existing topics to the peer.
         ///
-        /// - During session establishment, this method announces existing topics.
+        /// - During session establishment, this operation announces existing topics.
         /// - For already established sessions, it is used to announce new topics.
         ///
         /// A publisher session announces the topics it writes, while a subscriber session announces the topics it reads.
--- a/cpp/src/DataStorm/Contract.ice
+++ b/cpp/src/DataStorm/Contract.ice
@@ -238,10 +238,14 @@ module DataStormContract
 
     interface Session
     {
-        /// Announces existing topics to the peer during session establishment.
+        /// Announces new and existing topics to the peer.
+        ///
+        /// - During session establishment, this method announces existing topics.
+        /// - For already established sessions, it is used to announce new topics.
+        ///
         /// A publisher session announces the topics it writes, while a subscriber session announces the topics it reads.
         ///
-        /// The peer receiving the announcement will invoke `attachTopic` for the topics it is interested in.
+        /// The peer receiving the announcement will invoke `attachTopic` for any topics it is interested in.
         ///
         /// @param topics The sequence of topics to announce.
         /// @param initialize Currently unused.
@@ -251,6 +255,7 @@ module DataStormContract
         /// Attaches a local topic to a remote topic when a session receives a topic announcement from a peer.
         ///
         /// This method is called if the session is interested in the announced topic, which occurs when:
+        ///
         /// - The session has a reader for a topic that the peer has a writer for, or
         /// - The session has a writer for a topic that the peer has a reader for.
         ///
--- a/cpp/src/DataStorm/NodeSessionI.cpp
+++ b/cpp/src/DataStorm/NodeSessionI.cpp
@@ -14,8 +14,13 @@ using namespace DataStormContract;
 
 namespace
 {
-    // The NodeForwarder class is used to forward calls to a Node that doesn't have a public endpoint.
-    // It implements the Slice DataContract::Node interface by forwarding calls to the target Node object.
+    // The `NodeForwarder` class forwards calls to a `Node` that lacks a public endpoint.
+    //
+    // This class implements the Slice `DataContract::Node` interface by forwarding calls to the target `Node` object
+    // using the connection established during the creation of the `NodeSession` object.
+    //
+    // The `NodeForwarder` wraps the node and session proxy parameters passed to the `DataContract::Node` operations
+    // in forwarder proxies, which handle forwarding to the corresponding target objects.
     class NodeForwarder : public Node, public enable_shared_from_this<NodeForwarder>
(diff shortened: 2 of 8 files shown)


cpp/src/DataStorm/NodeSessionI.cpp
```diff
@@ -163,8 +168,6 @@ void
 NodeSessionI::destroy()
 {
     lock_guard<mutex> lock(_mutex);
-    _destroyed = true;
```
pepone, line 166: unused

cpp/src/DataStorm/NodeSessionI.cpp
```diff
@@ -197,5 +200,5 @@ void
 NodeSessionI::addSession(SessionPrx session)
 {
     lock_guard<mutex> lock(_mutex);
-    _sessions[session->ice_getIdentity()] = std::move(session);
+    _sessions.insert_or_assign(session->ice_getIdentity(), session);
```
pepone, line 203: replaced `[]` operator to allow using non-nullable proxies in the `std::map`

cpp/src/DataStorm/NodeSessionManager.cpp
```diff
@@ -191,7 +190,7 @@ NodeSessionManager::announceTopicWriter(const string& topic, NodePrx node, const
     }
 
     auto p = _sessions.find(node->ice_getIdentity());
-    auto nodePrx = p != _sessions.end() ? p->second->getPublicNode() : node;
+    node = p != _sessions.end() ? p->second->getPublicNode() : node;
```
pepone, line 193: There was a bug here, we were not using the updated proxy.

cpp/src/DataStorm/Contract.ice
```diff
@@ -238,10 +238,14 @@ module DataStormContract
 
     interface Session
     {
-        /// Announces existing topics to the peer during session establishment.
+        /// Announces new and existing topics to the peer.
+        ///
+        /// - During session establishment, this method announces existing topics.
```
bernardnormier, line 243: We use the term 'operation' for Slice.

cpp/src/DataStorm/NodeSessionI.h
```diff
@@ -41,19 +41,27 @@ namespace DataStormI
 
     private:
         const std::shared_ptr<Instance> _instance;
+        // A proxy to the target node, representing the node that created the session.
         DataStormContract::NodePrx _node;
+
+        // The connection used to create the session.
```
bernardnormier, line 47: Never null?

cpp/src/DataStorm/NodeSessionI.h
```diff
@@ -41,19 +41,27 @@ namespace DataStormI
...
         const Ice::ConnectionPtr _connection;
 
         std::mutex _mutex;
-        bool _destroyed;
+        // A proxy to the target node.
```
bernardnormier, line 51: From this description, it's not clear when it can be nullopt.

cpp/src/DataStorm/TopicFactoryI.h
```diff
@@ -63,13 +63,18 @@ namespace DataStormI
...
         std::weak_ptr<Instance> _instance;
-        std::shared_ptr<TraceLevels> _traceLevels;
-        std::map<std::string, std::vector<std::shared_ptr<TopicI>>> _readers;
-        std::map<std::string, std::vector<std::shared_ptr<TopicI>>> _writers;
+        mutable std::mutex _mutex;
         std::int64_t _nextReaderId;
         std::int64_t _nextWriterId;
+
+        // A map of topic readers indexed by the topic name.
+        // Each key is a topic name, and the corresponding value is a vector of readers for that topic.
```
bernardnormier, line 72: It's odd that `shared_ptr<TopicI>` represents a reader (resp. writer) for a Topic.

cpp/src/DataStorm/NodeSessionManager.h
```diff
@@ -77,7 +76,12 @@ namespace DataStormI
         // the specified node and sets this member once the connection is established.
         std::optional<DataStormContract::LookupPrx> _connectedTo;
 
+        // Stores the connection currently sending an announcement. This is used to exclude the node session
```
bernardnormier, line 79: Can it be null?

cpp/src/DataStorm/TopicFactoryI.cpp
```diff
@@ -30,6 +29,7 @@ TopicFactoryI::createTopicReader(
     shared_ptr<FilterManager> sampleFilterFactories)
 {
     shared_ptr<TopicReaderI> reader;
+    auto instance = getInstance();
```
bernardnormier, line 32: Not directly from this PR.

I am not a fan of this getInstance function. It would be more readable to expand it inline, as in:
```cpp
auto instance = _instance.lock();
assert(instance); // comment explaining why it can't be null here
```

bernardnormier (approved):
Looks good.

cpp/src/DataStorm/NodeSessionManager.h
pepone: It is always non null when forwarding calls in `NodeSessionManager::forward` which is the only place where it is read.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -588,7 +589,7 @@ SessionI::disconnected(const Ice::ConnectionPtr& connection, exception_ptr ex)
     lock_guard<mutex> lock(_mutex);
     if (_destroyed || (connection && _connection != connection) || !_session)
     {
-        // Ignore we either already destroyed, or disconnected, or a new connection has already been established.
+        // Ignore we are either already destroyed, or disconnected, or a new connection has already been established.
```
externl, line 592:
```suggestion
        // Ignore since we are either already destroyed, or disconnected, or a new connection has already been established.
```

externl (approved):

PR #3233, opened 2024-12-04
Rework DataStorm getInstance() methods

pepone:


cpp/src/DataStorm/TopicI.h
```diff
@@ -27,7 +27,8 @@ namespace DataStormI
 
     public:
         TopicI(
-            std::weak_ptr<TopicFactoryI>,
```
bernardnormier, line 30: In general, we should avoid weak_ptr in constructors.

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -154,11 +152,7 @@ NodeI::createSession(
     Ice::checkNotNull(subscriberSession, __FILE__, __LINE__, current);
 
     auto instance = _instance.lock();
-    if (!instance)
-    {
-        // Ignore the Node is being destroyed.
-        return;
-    }
+    assert(instance);
```
pepone, line 155: Here we are dispatching a request, and the instance owns the OA and destroys it before shutdown, so we should always have an instance here.

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -88,7 +89,7 @@ NodeI::destroy(bool ownsCommunicator)
 {
     unique_lock<mutex> lock(_mutex);
 
-    auto instance = getInstance();
+    auto instance = _instance.lock();
```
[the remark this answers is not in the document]
pepone: Right, here we always have an instance. This method is in fact synchronously called  from the instance.

PR #3235, opened 2024-12-04
Improve Transceiver.startWrite in C#

### LINKED ISSUE #3078: Cleanup C# startWrite API
It's not clear to me why the C# startWrite has such a complicated signature compared to the C++ startWrite. In C++, we don't have this "completedSynchronously" return value. 

In any case, assuming this complicated C# API is warranted, we should:
 - rename the last parameter of Transceiver.startWrite. Having return value = completedSynchronously and out parameter = completed is too confusing.
 - consider returning a tuple (does this work with the old C# version we use on 3.7?)
 - name the "messageWritten" parameter => messageFullyWritten
 - add a real doc-comment on Transceiver.startWrite

_Originally posted by @bernardnormier in https://github.com/zeroc-ice/ice/pull/3072#discussion_r1830189093_

bernardnormier:
Fixes #3078.

### DIFF
--- a/csharp/src/Ice/Internal/Transceiver.cs
+++ b/csharp/src/Ice/Internal/Transceiver.cs
@@ -38,14 +38,17 @@ public interface Transceiver
 
     void finishRead(Buffer buf);
 
-    //
-    // Write data asynchronously.
-    //
-    // The I/O request may complete synchronously, in which case finishWrite
-    // will be invoked in the same thread as startWrite. The request
-    // will be canceled upon the termination of the thread that calls startWrite.
-    //
-    bool startWrite(Buffer buf, AsyncCallback callback, object state, out bool completed);
+    /// <summary>Writes data asynchronously. The I/O request may complete synchronously, in which case
+    /// <see cref="finishWrite" /> will be invoked in the same thread as <see cref="startWrite" />. The operation will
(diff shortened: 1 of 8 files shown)


csharp/src/Ice/Internal/Transceiver.cs
```diff
@@ -38,14 +38,17 @@ public interface Transceiver
...
-    // will be canceled upon the termination of the thread that calls startWrite.
-    //
-    bool startWrite(Buffer buf, AsyncCallback callback, object state, out bool completed);
+    /// <summary>Writes data asynchronously. The I/O request may complete synchronously, in which case
+    /// <see cref="finishWrite" /> will be invoked in the same thread as <see cref="startWrite" />. The operation will
+    /// be canceled upon the termination of the thread that calls <see cref="startWrite" />.</summary>
+    /// <param name="buf">The buffer to write.</param>
+    /// <param name="callback">The callback that this operation calls when it completes asynchronously.</param>
+    /// <param name="state">The state to pass to the callback.</param>
+    /// <param name="messageFullyWritten">When <see langword="true" />, the operation has written or is the process of
```
pepone, line 47:
```suggestion
    /// <param name="messageFullyWritten">When <see langword="true" />, the operation has written or is in the process of
```

pepone (approved):

PR #3237, opened 2024-12-05
Add C# code coverage workflow

externl:
Generated report is available here: https://code.zeroc.com/ice/csharp/main/coverage/index.html

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -72,7 +72,7 @@ jobs:
         working-directory: ./csharp
         run: |
           dotnet tool install  dotnet-coverage --global
-          dotnet-coverage collect --server-mode --background --session-id ice-coverage --output coverage.cobertura.xml --output-format cobertura
+          dotnet-coverage collect --server-mode --background --session-id ice-coverage --output coverage.cobertura.xml --output-format cobertura --settings msbuild/coverage.runsettings
           python3 allTests.py --all --workers=4 --filter Ice/ --debug --coverage-session ice-coverage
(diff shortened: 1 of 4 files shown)


.github/workflows/coverage.yml
```diff
@@ -2,9 +2,11 @@ name: Generate API Coverage
 
 on:
   workflow_dispatch:
+  push:
+    branches: ["coverage-test"]
```
pepone, line 6: Need to update this before merging

csharp/msbuild/coverage.runsettings
```diff
@@ -0,0 +1,22 @@
...
+      <DataCollector friendlyName="Code Coverage" uri="datacollector://Microsoft/CodeCoverage/2.0"
+        assemblyQualifiedName="Microsoft.VisualStudio.Coverage.DynamicCoverageDataCollector, Microsoft.VisualStudio.TraceCollector, Version=11.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a">
+        <Configuration>
+          <CodeCoverage>
+            <Sources>
+              <Include>
+                <Source>.*/src/.*</Source>
+              </Include>
+              <Exclude>
+                <Source>.*/generated/.*</Source>
```
pepone, line 14: Can we keep this to catch unused, and untested generated code?

externl: I think looking the coverage for generated code in general just messes up the numbers and doesn't generally give any meaningful information. I think generally, most generated code is unused. We filter out generated code for IceRPC coverage as well.

PR #3238, opened 2024-12-05
Fix lints in C++ code

### LINKED ISSUE #2358: C++ exceptions should have a noexcept copy constructor and assignment operator
Related to: https://github.com/zeroc-ice/ice-demos/issues/146

All C++ local exceptions should follow this rule. 

For generated user exceptions with user-specified fields, it should be true when the fields are noexcept copy-constructible and assignable. Unfortunately we can't provide this guarantee for user exceptions with fields that aren't noexcept copy-constructible  (like std::string) without breaking source compatibility. We could try to disable this diagnostics for such generated user exceptions.

bernardnormier:
This PR fixes a number of lints reported by clang-tidy.

Fixes #2358.

### DIFF
--- a/cpp/include/Ice/FactoryTable.h
+++ b/cpp/include/Ice/FactoryTable.h
@@ -21,7 +21,7 @@ namespace IceInternal
     class ICE_API FactoryTable final
     {
     public:
-        FactoryTable() = default;
+        FactoryTable() noexcept = default;
 
         FactoryTable(const FactoryTable&) = delete;
         FactoryTable& operator=(const FactoryTable&) = delete;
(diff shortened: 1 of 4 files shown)


cpp/src/slice2cpp/Gen.cpp
```diff
@@ -2025,6 +2025,12 @@ Slice::Gen::DataDefVisitor::visitExceptionStart(const ExceptionPtr& p)
             H.dec();
             H << sb;
             H << eb;
+
+            // We generate a noexcept copy constructor all the time. A C++ exception must have noexcept copy constructor
```
bernardnormier, line 2029: That's the fix for #2358: always generate a noexcept copy constructor even when the fields (e.g. string) don't have one. This will result in a crash if the field copy construction throws.

cpp/include/Ice/FactoryTable.h
```diff
@@ -88,27 +88,27 @@ namespace IceInternal
...
 
-        const std::string_view typeId;
+        const char* typeId;
     };
 
     template<class O> class DefaultValueFactoryInit
     {
     public:
-        DefaultValueFactoryInit(std::string_view tId) : typeId(tId)
+        DefaultValueFactoryInit(const char* tId) noexcept : typeId(tId)
```
bernardnormier, line 104: The `std::string_view(const char*)` constructor is not noexcept - and we always construct these objects with string literals.

externl (approved):

pepone (approved):

PR #3242, opened 2024-12-06
Fix ice_getConnectionAsync

### LINKED ISSUE #1729: getConnectionAsync throws ConnectionClosed synchronously
See `TODO` in python Ice/ami test for `getConnectionAsync`, this exception was previously throw asynchronously when python extension used the old C++98 mapping.

bernardnormier:
This PR fixes ice_getConnectionAsync to no throw synchronously NoEndpointException and other local exceptions.

Fixes #1729.

### DIFF
--- a/cpp/src/Ice/OutgoingAsync.cpp
+++ b/cpp/src/Ice/OutgoingAsync.cpp
@@ -494,10 +494,8 @@ ProxyOutgoingAsyncBase::invokeImpl(bool userThread)
     }
     catch (const Exception&)
     {
-        //
-        // If called from the user thread we re-throw, the exception
-        // will be catch by the caller and abort(ex) will be called.
-        //
+        // If called from the user thread we re-throw, the exception will be caught by the caller and handler using
+        // abort.
         if (userThread)
         {
             throw;
@@ -1064,11 +1062,7 @@ OutgoingAsync::invoke(string_view operation)
         return;
     }
 
-    //
-    // NOTE: invokeImpl doesn't throw so this can be called from the
-    // try block with the catch block calling abort(ex) in case of an
-    // exception.
-    //
+    // invokeImpl can throw. The exception should be handled by calling abort (in the caller).
     invokeImpl(true); // userThread = true
 }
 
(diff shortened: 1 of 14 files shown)


cpp/src/Ice/OutgoingAsync.cpp
```diff
@@ -1064,11 +1062,7 @@ OutgoingAsync::invoke(string_view operation)
         return;
     }
 
-    //
-    // NOTE: invokeImpl doesn't throw so this can be called from the
```
bernardnormier, line 1068: I don't know why we had this bogus comment everywhere.

csharp/src/Ice/Proxy.cs
```diff
@@ -1583,14 +1583,7 @@ public override void handleInvokeResponse(bool ok, OutgoingAsyncBase og)
     private void iceI_ice_getConnection(OutgoingAsyncCompletionCallback completed, bool synchronous)
     {
         var outgoing = new ProxyGetConnection(this, completed);
-        try
-        {
-            outgoing.invoke(_ice_getConnection_name, synchronous);
-        }
-        catch (Ice.Exception ex)
-        {
-            outgoing.abort(ex);
```
bernardnormier, line 1592: The C# code was overall correct: we were catching and handling the exception with abort.

I just move this catch/handle to the invoke method implementation, for consistency with other invoke implementations and other language mappings.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ProxyGetConnection.java
```diff
@@ -40,6 +40,10 @@ public int invokeCollocated(CollocatedRequestHandler handler) {
     }
 
     public void invoke() {
-        invokeImpl(true); // userThread = true
+        try {
```
bernardnormier, line 43: This try/catch was missing in Java, just like in C++.

pepone (approved):
Looks good.

externl (approved):

PR #3244, opened 2024-12-09
Update Swift build instructions

externl:


swift/BUILDING.md
```diff
@@ -20,36 +20,31 @@ resulting build.
...
 ### Swift Version
 
 Ice for Swift requires Swift 5 or later.
 
 ## Building Ice for Swift
 
-Then open `ice.xcodeproj` with Xcode and build the `Ice macOS` or `Ice iOS`
-targets.
+Building Ice for Swift requires having built Ice for C++ first from the `cpp` directory. This builds the required
+`XCFramework`s that are used by the Swift package.
```
bernardnormier, line 30: Meaning the Swift package requires multiple XCFrameworks?

It would be clearer to write something:
This builds the 3 C++ frameworks that are used by the Swift package.

swift/BUILDING.md
```diff
@@ -20,36 +20,31 @@ resulting build.
...
 
 ```shell
-export DEVELOPMENT_TEAM=U4TBVKNQ7F
-rake
+make PLATFORM=all
 ```
 
+### Building with Xcode
+
+`Package.swift` (Ice for Swift source code), `test/Package.swift` (Ice for Swift tests), and `test/ios/TestDriverApp.xcodeproj` (iOS test controller) can be opened and built using Xcode.
```
bernardnormier, line 46: Should it be links to the files?

For example, Package.swift is ice/Package.swift, not ice/swift/Package.swift.

swift/BUILDING.md
externl: It's actually one nested XCFramework.

pepone: I see 3 separate XCFramework:

```
lib/Ice.xcframework
lib/IceDiscovery.xcframework
lib/IceLocatorDiscovery.xcframework
```
Where is the nested XCFramework?

externl: Right, I forgot about Ice*Discovery. There's a platform dependent xcframework for each of the three platforms we support inside each of those frameworks.

bernardnormier (approved):

swift/BUILDING.md
```diff
@@ -20,36 +22,49 @@ resulting build.
...
-signing the application with a developer certificate. You need to update the Xcode projects
-to use your Apple development certificates by setting `DEVELOPMENT_TEAM` environment
-variable to the Id of your development team and regenerate the projects with `rake`.
+If targeting Swift for iOS (or iOS simulator), the XCFrameworks must also be built to include the iOS architectures:
 
 ```shell
-export DEVELOPMENT_TEAM=U4TBVKNQ7F
-rake
+cd cpp
+make PLATFORM=all
```
pepone, line 47: Or maybe just:

```suggestion
make -C ../cpp PLATFORM=all
```

pepone (approved):

PR #3245, opened 2024-12-09
Cleanup Ice.Internal.Network in C#

bernardnormier:
Remove unused methods + switched to internal methods.


externl (approved):
Looks good. I suspect this `doConnect` function might be dead code in C++ and Java as well.

pepone (approved):

PR #3251, opened 2024-12-10
Remove class-by-value support in DataStorm

bernardnormier:
This PR removes support for using Slice classes "by value" as DataStorm values. 

You can still use Slice classes as reference with DataStorm; if you're currently using Slice classes "by value", you can switch to "by reference" and maintain on-the-wire interop with earlier DataStorm/Ice releases.


cpp/include/DataStorm/Types.h
```diff
@@ -240,58 +240,6 @@ namespace DataStorm
         static T decode(const Ice::CommunicatorPtr& communicator, const Ice::ByteSeq& value) noexcept;
     };
 
-    /**
-     * The Cloner template provides a method to clone user types.
```
bernardnormier, line 244: Removed Cloner as it was used/tested only for "class by value". Its documentation (in the IceStorm manual) is also incomplete and rather confusing/incorrect.

cpp/include/DataStorm/Types.h
```diff
@@ -240,58 +240,6 @@ namespace DataStorm
...
-         * with the copy constructor.
-         *
-         * @param value The value to encode
-         * @return The cloned value
-         */
-        static T clone(const T& value) noexcept { return value; }
-    };
-
-    /**
-     * Encoder template specialization to encode Ice::Value instances.
```
bernardnormier, line 265: Only used for class-by-value.

cpp/include/DataStorm/Types.h
pepone: According to the documentation, the cloner template is also used by partial updates for cloning types. I believe it is required for class-by-reference scenarios to ensure `ice_clone` is called instead of using the shared_ptr copy constructor.

It might be worth updating `DataStorm/partial` to test class-by-reference behavior, as the current tests only cover structs. The original implementation tested a class-by-value, which is a removed by this PR.

From my understanding, without this change, the partial update will modify the previous value directly rather than computing a new value from previous.

bernardnormier: You were right, it was used by the partial feature.

I restored Cloner and updated the partial test to use class-by-reference.

pepone (approved):
Looks good.

externl (approved):

PR #3253, opened 2024-12-10
Java coverage fixes

externl:
This PR updates the coverage generation for Java (and adds a workflow schedule). I moved the coverage logic into a bash script because it's a little complicated and somebody might want to run it locally.


.github/workflows/coverage.yml
```diff
@@ -17,18 +19,18 @@ jobs:
 
       - name: Generate C++ Coverage
         working-directory: ./cpp
-        run: ../scripts/generate-code-coverage.sh
+        run: ../scripts/generate-cpp-code-coverage.sh
```
externl, line 22: I renamed this file.

scripts/generate-java-code-coverage.sh
```diff
@@ -0,0 +1,55 @@
...
+
+JACOCO_VERSION=0.8.12
+
+if [ "$(basename "$(pwd)")" != "java" ]; then
+    echo "This script must be run from the 'java' directory."
+    exit 1
+fi
+
+# make -C ../cpp srcs
+# make
```
InsertCreativityHere, line 11: We need these uncommented to actually run the build, right?

scripts/generate-java-code-coverage.sh
```diff
@@ -0,0 +1,55 @@
...
+fi
+
+if [ ! -f org.jacoco.agent-runtime.jar ]; then
+    wget https://repo1.maven.org/maven2/org/jacoco/org.jacoco.agent/$JACOCO_VERSION/org.jacoco.agent-$JACOCO_VERSION-runtime.jar -O org.jacoco.agent-runtime.jar
+fi
+
+python3 allTests.py --all --workers=4 --debug --jacoco $(pwd)/org.jacoco.agent-runtime.jar
+
+# We don't want any coverage for IceGridGUI
+find src/IceGridGUI -name '*.java' -exec rm {} \;
```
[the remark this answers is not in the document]
InsertCreativityHere: Could we remove the class files like we do for the generated code?
They should all live in `src/IceGridGUI/build/classes`.

So maybe it'd be easier to just `rm -rf src/IceGridGUI/build` instead of using `find ... -exec rm`.

scripts/generate-java-code-coverage.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# We don't want any coverage for generated code. We remove their class files and any inner classes.
+jacocoArgs=()
+for dir in src/*; do
+    [ -d "$dir" ] || continue
```
InsertCreativityHere, line 29: Actually, I think the most correct solution is to just `continue` when `dir == IceGridGui` or whatever.
This means we don't need to `find rm` anything, which would be nice.

Also, the current approach feels a little weird. `find src/IceGridGUI -name '*.java' -exec rm {} \;` removes all the source files, but not the _class_ files. So, we're still passing all the `IceGridGUI` classes to jacoco. But, I guess it doesn't display any output for them because it doesn't know about the source files?
Feels sketchy. It'd be better to just not pass the `IceGridGui` directory or anything at all IMO.

InsertCreativityHere (approved):

scripts/generate-java-code-coverage.sh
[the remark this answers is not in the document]
externl: Yes, this is a bug. I was attempting to remove the class files.

scripts/generate-java-code-coverage.sh
externl: Yes! Fixed.

PR #3254, opened 2024-12-10
Add clang-tidy config and tidy-up generated code

bernardnormier:
This PR adds a clang-tidy config file (.clang-tidy, from ice-demos), and fixes lints in the generated code.

You can run clang-tidy over all the generated code on macos as follows:
```zsh
xcrun run-clang-tidy **/generated/**/*.cpp -j12 -quiet 
```

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -13,6 +13,7 @@ WarningsAsErrors:  ''
 HeaderFilterRegex: ''
 UseColor: true
 FormatStyle: 'file'
+ExtraArgs: ['-std=c++17']
 CheckOptions:
   - { key: modernize-use-nullptr.NullMacros, value: 'NULL' }
   # std::exception_ptr is a cheap to copy, pointer-like type; we pass it by value all the time.
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -9,12 +9,12 @@ Checks:
   performance-*,
   -performance-avoid-endl
 '
-WarningsAsErrors:  ''
+# WarningsAsErrors: '*'
 HeaderFilterRegex: ''
 UseColor: true
 FormatStyle: 'file'
 ExtraArgs: ['-std=c++17']
 CheckOptions:
-  - { key: modernize-use-nullptr.NullMacros, value: 'NULL' }
+  modernize-use-nullptr.NullMacros: 'NULL'
   # std::exception_ptr is a cheap to copy, pointer-like type; we pass it by value all the time.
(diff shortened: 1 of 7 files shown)


cpp/test/Ice/custom/CustomBuffer.h
```diff
@@ -53,6 +59,19 @@ namespace Test
             return *this;
         }
 
+        CustomBuffer& operator=(CustomBuffer&& o)
+        {
+            if (_buf)
```
pepone, line 64: We can remove the if check like in `MyByteSeq::operator=(`

cpp/test/Ice/custom/CustomBuffer.h
```diff
@@ -31,9 +31,15 @@ namespace Test
...
+        CustomBuffer(CustomBuffer&& o) : _buf(o._buf), _count(o._count)
+        {
+            o._buf = nullptr;
+            o._count = 0;
+        }
+
         ~CustomBuffer()
         {
-            if (_buf != 0)
+            if (_buf)
```
pepone, line 42: We can remove the if check

pepone (approved):

cpp/test/Ice/custom/CustomBuffer.h
bernardnormier: Removed.

cpp/test/Ice/custom/CustomBuffer.h
bernardnormier: Fixed

.clang-tidy
```diff
@@ -0,0 +1,20 @@
...
+'
+WarningsAsErrors:  ''
+HeaderFilterRegex: ''
+UseColor: true
+FormatStyle: 'file'
+ExtraArgs: ['-std=c++17']
+CheckOptions:
+  - { key: modernize-use-nullptr.NullMacros, value: 'NULL' }
+  # std::exception_ptr is a cheap to copy, pointer-like type; we pass it by value all the time.
+  - { key: performance-unnecessary-value-param.AllowedTypes, value: 'std::exception_ptr' }
```
externl, line 20:
```suggestion
CheckOptions:
  modernize-use-nullptr.NullMacros: 'NULL'
  # std::exception_ptr is a cheap to copy, pointer-like type; we pass it by value all the time.
  performance-unnecessary-value-param.AllowedTypes: 'exception_ptr$;'
```

externl (approved):

.clang-tidy
bernardnormier: Fixed.

Issue #3246, opened 2024-12-09
C# Metadata Questions


InsertCreativityHere:
I've been working on #117 (improve metadata validation).
For my PR, I'm trying to keep the current behavior as much as possible (to avoid changing too much in one PR),
but I've across a few oddities that don't seem right to me. I intend to fix these in a follow-up PR, but wanted to list them here,
in case anyone had any thoughts/explanations to offer.

1) We support `SortedList` as an argument to `cs:generic` on dictionaries:
This seems nonsensical to me, why would you map a dictionary to a list, instead of using a sequence?
Also, in the [documentation](https://doc.zeroc.com/ice/3.7/the-slice-language/slice-metadata-directives#id-.SliceMetadataDirectivesv3.7-cs:generic:SortedDictionary), we only list `SortedDictionary`, not `SortedList`.
**EDIT: @pepone points out that `SortedList` _is_ actually an `IDictionary`, so we should keep support for it.**

2) We only support `cs:implements` on structs and interfaces. Why don't we support this on classes?
Note that in the [documentation](https://doc.zeroc.com/ice/3.7/the-slice-language/slice-metadata-directives#id-.SliceMetadataDirectivesv3.7-cs:implements:type), we state that it's supported on struct, class, and interface.
It's not clear if the documentation is outdated, or if this is a bug in the current validation logic.
**EDIT: after looking at it further, this metadata seems useless and will be removed instead.**

3) Should we allow `cs:attribute` on enumerators?
If you put this metadata on an enumerator, we don't reject it, but we also don't seem to generate code for it.
It's just silently ignored. Is there a reason we'd disallow this? Or is this just missing functionality.
**I intend to change `slice2cs` to generate attributes on enumerators like we do for everything else.**

4) Should we allow `cs:property` on exceptions?
The [documentation ](https://doc.zeroc.com/ice/3.7/the-slice-language/slice-metadata-directives#id-.SliceMetadataDirectivesv3.7-cs:property) says it's only supported on structs and classes, but the validation logic allowed it through: https://github.com/zeroc-ice/ice/blob/2c85ea718e31b38d27fd3728f4f70795fde91084/cpp/src/slice2cs/CsUtil.cpp#L2015-L2017 **I intend to allow this metadata on exceptions, unless there's a reason not to**

pepone:
- `cs:implements` I think we should consider dropping this, in favor of defining this in a partial.

pepone:
[...]

PR #3256, opened 2024-12-11
Use the new MetadataValidator in C#, Java, JavaScript, Ruby, PHP, and MATLAB

InsertCreativityHere:


cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -1408,8 +1026,12 @@ Slice::JavaGenerator::writeMarshalUnmarshalCode(
                     return;
                 }
             }
+            // Check if either 1) No type metadata was applied to this sequence at all or 2) 'java:type' was applied to
+            // where the sequence is used (which overrides any metadata on the definition) or 3) 'java:type' was applied
+            // to the sequence definition, and there is no metadata overriding it where the sequence is used.
```
InsertCreativityHere, line 1031: An unrelated bug I stumbled across. We weren't always checking for these metadata in the same order, which could cause us to generate uncompilable code(because the types wouldn't match). Pretty niche corner case, but still.

You'll see me messing around with `java:type`, `java:buffer`, and `java:serializable` a few times in this file.
All I'm doing is making sure we check them in a consistent (and correct) order.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -1715,49 +1343,38 @@ Slice::JavaGenerator::writeSequenceMarshalUnmarshalCode(
     string typeS = typeToString(seq, TypeModeIn, package);
     string v = param;
 
-    //
-    // If the sequence is a byte sequence, check if there's the serializable metadata to
-    // get rid of this case first.
-    //
-    BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(seq->type());
-    if (builtin && builtin->kind() == Builtin::KindByte)
```
InsertCreativityHere, line 1723: The validator already ensures that `java:serializable`  only goes on byte sequences.
No need to repeat the validation here.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -1715,49 +1343,38 @@ Slice::JavaGenerator::writeSequenceMarshalUnmarshalCode(
...
         }
+        return;
     }
 
+    BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(seq->type());
     static const char* builtinTable[] = {"Byte", "Bool", "Short", "Int", "Long", "Float", "Double", "String"};
 
-    if (builtin && (builtin->kind() == Builtin::KindByte || builtin->kind() == Builtin::KindShort ||
-                    builtin->kind() == Builtin::KindInt || builtin->kind() == Builtin::KindLong ||
-                    builtin->kind() == Builtin::KindFloat || builtin->kind() == Builtin::KindDouble))
```
InsertCreativityHere, line 1743: Same. No need to repeat the validation here.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -2192,49 +1809,16 @@ Slice::JavaGenerator::getTypeMetadata(const MetadataList& metadata, string& inst
 }
 
 bool
-Slice::JavaGenerator::hasTypeMetadata(const TypePtr& type, const MetadataList& localMetadata)
```
InsertCreativityHere, line 2195: Two simplifications:
1) This is only called for sequences, so we can eliminate alot of casting/checking of types.
2) This function included validation of the metadata, which was already being performed by the validator.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -2282,34 +1866,19 @@ Slice::JavaGenerator::getSequenceTypes(
...
-            {
-                instanceType = formalType = *meta;
-                return true;
-            }
-        }
+        instanceType = formalType = *meta;
+        return true;
+    }
 
-        if ((builtin->kind() == Builtin::KindByte || builtin->kind() == Builtin::KindShort ||
```
InsertCreativityHere, line 2297: Same, no need to repeat validation again.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -460,27 +88,17 @@ Slice::getSerialVersionUID(const ContainedPtr& p)
...
-                serialVersionUID = std::stoll(value, nullptr, 0);
-            }
-            catch (const std::exception&)
-            {
-                ostringstream os;
-                os << "ignoring invalid serialVersionUID for " << p->kindOf() << " `" << p->scoped()
-                   << "'; generating default value";
+            ostringstream os;
+            os << "failed to parse custom serialVersionUID value for " << p->kindOf() << " `" << p->scoped()
+               << "'; generating default value";
```
externl, line 99: Not related to your PR, I think this should be an error, not a warning.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -2338,6 +1905,128 @@ Slice::JavaGenerator::createOutput()
...
+        // metadata emit the conflict warnings instead of this one, since this can only go on the definition.
+        return nullopt;
+    };
+    knownMetadata.emplace("java:serializable", std::move(serializableInfo));
+
+    // "java:serialVersionUID"
+    MetadataInfo serialVersionUIDInfo;
+    serialVersionUIDInfo.validOn = {typeid(ClassDecl), typeid(Slice::Exception), typeid(Struct)};
+    serialVersionUIDInfo.acceptedArgumentKind = MetadataArgumentKind::SingleArgument;
+    knownMetadata.emplace("java:serialVersionUID", std::move(serialVersionUIDInfo));
```
externl, line 2001: should we do the validation `serialVersionUIDInfo` here in an extraValidation function?

externl (approved):

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1864,212 +1865,91 @@ Slice::CsGenerator::toArrayAlloc(const string& decl, const string& sz)
...
-void
-Slice::CsGenerator::MetadataVisitor::visitEnum(const EnumPtr& p)
-{
-    validate(p);
-}
+    // "cs:attribute"
+    MetadataInfo attributeInfo;
+    attributeInfo.validOn = {
+        typeid(Unit),
+        typeid(Module),
```
bernardnormier, line 1874: .NET does not support attributes on C# namespaces:
https://learn.microsoft.com/en-us/dotnet/api/system.attributetargets?view=net-9.0

So this looks bogus.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1864,212 +1865,91 @@ Slice::CsGenerator::toArrayAlloc(const string& decl, const string& sz)
...
-            else if (dynamic_pointer_cast<ClassDecl>(cont) || dynamic_pointer_cast<Exception>(cont))
+            if (argument == "LinkedList" || argument == "Queue" || argument == "Stack")
             {
-                if (directive == "cs:property" && arguments.empty())
-                {
-                    newLocalMetadata.push_back(metadata);
-                    continue;
-                }
+                return "'cs:generic:" + argument +
+                       "' is not supported on sequences of objects; only 'List' is supported for object sequences";
```
bernardnormier, line 1906: We should not use 'objects', as it's confusing. sequence of classes? sequence of class instances?

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1864,212 +1865,91 @@ Slice::CsGenerator::toArrayAlloc(const string& decl, const string& sz)
...
         }
-        newLocalMetadata.push_back(metadata);
-    }
+        return nullopt;
+    };
+    knownMetadata.emplace("cs:namespace", std::move(namespaceInfo));
+
+    // "cs:property"
+    MetadataInfo propertyInfo;
+    propertyInfo.validOn = {typeid(ClassDecl), typeid(Slice::Exception), typeid(Struct)};
```
bernardnormier, line 1949: For validOn, we must used ClassDecl/InterfaceDecl and not ClassDef/InterfaceDef?

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -2338,6 +1905,128 @@ Slice::JavaGenerator::createOutput()
...
+    };
+    knownMetadata.emplace("java:buffer", std::move(bufferInfo));
+
+    // "java:getset"
+    MetadataInfo getsetInfo;
+    getsetInfo.validOn = {typeid(ClassDecl), typeid(Slice::Exception), typeid(Struct), typeid(DataMember)};
+    getsetInfo.acceptedArgumentKind = MetadataArgumentKind::NoArguments;
+    knownMetadata.emplace("java:getset", std::move(getsetInfo));
+
+    // "java:implements"
```
bernardnormier, line 1951: How does this work?

Without partial (like in C#), how is the user going to implement any of the interface methods?

bernardnormier (approved):

cpp/src/slice2java/JavaUtil.cpp
pepone: From the docs:

> When used with structs, this metadata can only refer to interfaces without operations. With classes or interfaces, the generated Java interface will be marked as abstract. The code is responsible for registering a [value factory](https://doc.zeroc.com/ice/3.7/the-slice-language/classes/value-factories) if the Slice class is transferred over-the-wire and uses this metadata to implement native Java interfaces.

cpp/src/Slice/MetadataValidation.cpp
```diff
@@ -73,68 +73,62 @@ Slice::validateMetadata(const UnitPtr& p, string_view prefix, map<string, Metada
     // So, we add all the language-agnostic metadata validation into the provided list.
 
     // "amd"
```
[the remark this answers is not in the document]
InsertCreativityHere: I switched everything to use this style of initialization instead.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -859,33 +859,28 @@ Slice::Gen::validateMetadata(const UnitPtr& u)
     map<string, MetadataInfo> knownMetadata;
 
     // "cpp:array"
```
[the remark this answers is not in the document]
InsertCreativityHere: Fixed (by using brace initialization).

cpp/src/slice2java/Gen.cpp
```diff
@@ -3703,17 +3703,13 @@ void
...
     BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(p->type());
-    if (!hasTypeMetadata(p) && builtin && builtin->kind() <= Builtin::KindString)
+    bool mappedToCustomType = p->hasMetadata("java:type");
+    if (builtin && builtin->kind() < Builtin::KindObject && !mappedToCustomType)
     {
-        return; // No helpers for sequence of primitive types
-    }
-    else if (hasTypeMetadata(p) && !p->hasMetadata("java:type"))
-    {
-        return; // No helpers for custom metadata other than java:type
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes, that's correct.
For completeness (and future readers) though:

The first `if` check is completely unchanged. And the 2nd one is redundant with these new guarantees.
You can 'expand' the 'hasTypeMetadata':
```
(hasTypeMetadata(p) && !p->hasMetadata("java:type"))
    ==>
((p->hasMetadata("java:buffer") || p->hasMetadata("java:serializable")) && !p->hasMetadata("java:type"))
```
(hasTypeMetadata also checks for `java:type`, but we can ignore it because of the `&& !p->hasMetadata("java:type")`)

And we already have a variable for saying that `java:type` doesn't appear, so:
```
((p->hasMetadata("java:buffer") || p->hasMetadata("java:serializable")) && !p->hasMetadata("java:type"))
    ==>
((p->hasMetadata("java:buffer") || p->hasMetadata("java:serializable")) && !mappedToCustomType)
```
This looks very similar to the first if check of:
```
if (builtin && builtin->kind() < Builtin::KindObject && !mappedToCustomType)
```

And it is in fact, completely redundant. Because, like you point out, `java:buffer` and `java:serializable` can only appear on sequences of builtin types. so if either of these are present, _we would of already gotten into the first if block_.

cpp/src/slice2java/JavaUtil.cpp
InsertCreativityHere: Good idea!
I moved the extra `stoll` check into the metadata validation, instead of doing it on the spot.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1864,212 +1865,91 @@ Slice::CsGenerator::toArrayAlloc(const string& decl, const string& sz)
...
 
-void
-Slice::CsGenerator::MetadataVisitor::visitEnum(const EnumPtr& p)
-{
-    validate(p);
-}
+    // "cs:attribute"
+    MetadataInfo attributeInfo;
+    attributeInfo.validOn = {
+        typeid(Unit),
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes, that is exactly correct.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1864,212 +1865,91 @@ Slice::CsGenerator::toArrayAlloc(const string& decl, const string& sz)
...
+        typeid(Module),
+        typeid(InterfaceDecl),
+        typeid(ClassDecl),
+        typeid(Operation),
+        typeid(Slice::Exception),
+        typeid(Struct),
+        typeid(Enum),
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes, there's an open issue for adding it (#3246), but I didn't want to change too much in this PR.
So my validator stays as close to the original validation, including any mistakes that were in it.

After I've updated the compilers to use my `MetadataValidator`, then I'll go back and fix all the bugs that it's uncovered.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1864,212 +1865,91 @@ Slice::CsGenerator::toArrayAlloc(const string& decl, const string& sz)
...
+    MetadataInfo namespaceInfo;
+    namespaceInfo.validOn = {typeid(Module)};
+    namespaceInfo.acceptedArgumentKind = MetadataArgumentKind::SingleArgument;
+    namespaceInfo.extraValidation = [](const MetadataPtr&, const SyntaxTreeBasePtr& p) -> optional<string>
+    {
+        // 'cs:namespace' can only be applied to top-level modules
+        if (auto mod = dynamic_pointer_cast<Module>(p))
+        {
+            // Top-level modules are contained by the 'Unit'. Non-top-level modules are contained in 'Module's.
+            if (!dynamic_pointer_cast<Unit>(mod->container()))
```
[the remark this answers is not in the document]
InsertCreativityHere: Changed as suggested.

cpp/src/slice2java/Gen.cpp
```diff
@@ -3703,17 +3703,13 @@ void
 Slice::Gen::HelperVisitor::visitSequence(const SequencePtr& p)
 {
     BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(p->type());
-    if (!hasTypeMetadata(p) && builtin && builtin->kind() <= Builtin::KindString)
+    bool mappedToCustomType = p->hasMetadata("java:type");
+    if (builtin && builtin->kind() < Builtin::KindObject && !mappedToCustomType)
```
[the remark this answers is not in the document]
InsertCreativityHere: We did not, but I added a method named `mapsToJavaBuiltinType` that we call now instead.
It's the same logic, but hopefully the name makes it more digestable.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -1715,49 +1343,38 @@ Slice::JavaGenerator::writeSequenceMarshalUnmarshalCode(
...
     }
 
+    BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(seq->type());
     static const char* builtinTable[] = {"Byte", "Bool", "Short", "Int", "Long", "Float", "Double", "String"};
 
-    if (builtin && (builtin->kind() == Builtin::KindByte || builtin->kind() == Builtin::KindShort ||
-                    builtin->kind() == Builtin::KindInt || builtin->kind() == Builtin::KindLong ||
-                    builtin->kind() == Builtin::KindFloat || builtin->kind() == Builtin::KindDouble))
+    static const string bytebuffer = "java:buffer";
+    if ((seq->hasMetadata(bytebuffer) || hasMetadata(bytebuffer, metadata)) && !hasMetadata("java:type", metadata))
```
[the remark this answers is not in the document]
InsertCreativityHere: > is this to ensure java:type metadata in the parameter overrides java:buffer in the definition?

Precisely. But more generally we want to make sure that parameter metadata always overrides definition metadata.

> Do we allow both java:buffer and java:type in the same definition?

No, we explicitly check that you can only have one of `java:type`, `java:buffer` or `java:serializable` on your definition. The only way to have "two of them" is by putting one on the definition and one on the parameter.
And in this case, parameter metadata is always given priority.

So we know that if `seq->hasMetadata(bytebuffer)` is true, then the only place we need to check is the parameter's metadata (and for completeness, `java:serializable` can only go on the definition, so the only thing left to check for is whether `java:type` was put on the parameter).

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -2338,6 +1905,128 @@ Slice::JavaGenerator::createOutput()
...
+                             builtin->kind() != Builtin::KindInt && builtin->kind() != Builtin::KindLong &&
+                             builtin->kind() != Builtin::KindFloat && builtin->kind() != Builtin::KindDouble))
+            {
+                return "the 'java:buffer' metadata can only be applied to sequences of bytes, shorts, ints, longs, "
+                       "floats, or doubles";
+            }
+
+            // This check for 'hasMetadata("java:type")' looks redundant, but is necessary to ensure
+            // that 'java:type' was placed on the _definition_ and not on where the sequence was used.
+            if (seq->hasMetadata("java:type") && seq->hasMetadata("java:buffer"))
```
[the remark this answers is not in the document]
InsertCreativityHere: Good catch, the comment was supposed to refer to `java:buffer`, I got myself mixed up.
Fixed.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -2338,6 +1905,128 @@ Slice::JavaGenerator::createOutput()
...
+    MetadataInfo packageInfo;
+    packageInfo.validOn = {typeid(Unit), typeid(Module)};
+    packageInfo.acceptedArgumentKind = MetadataArgumentKind::SingleArgument;
+    packageInfo.extraValidation = [](const MetadataPtr&, const SyntaxTreeBasePtr& p) -> optional<string>
+    {
+        // If 'java:package' is applied to a module, it must be a top-level module.
+        if (auto mod = dynamic_pointer_cast<Module>(p))
+        {
+            // Top-level modules are contained by the 'Unit'. Non-top-level modules are contained in 'Module's.
+            if (!dynamic_pointer_cast<Unit>(mod->container()))
```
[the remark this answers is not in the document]
InsertCreativityHere: Also changed!

cpp/src/slice2cs/CsUtil.cpp
InsertCreativityHere: Agreed, this was the old behavior which I just tried to preserve.
I added this to the C# metadata issue I already opened: https://github.com/zeroc-ice/ice/issues/3246#issuecomment-2546350539

cpp/src/slice2cs/CsUtil.cpp
InsertCreativityHere: I went with `sequences of classes types`, I can see how "objects" would be confusing. Now it reads:
```
 "'cs:generic:" + argument + "' is not supported on sequences of classes; only 'List' is supported for such sequences"
```

cpp/src/slice2cs/CsUtil.cpp
InsertCreativityHere: Yes, after my PR for simplifying the Decl/Def situation, you must always pass in the Decls.
There's a comment saying so in `MetadataValidator.h`:
https://github.com/zeroc-ice/ice/blob/6e237b53f7d94e64ff1b5cddde6504319226dac2/cpp/src/Slice/MetadataValidation.h#L68-L70

cpp/src/slice2java/JavaUtil.cpp
bernardnormier: To be clear: even though it's documented, it's not usable and should be removed.

pepone (approved):

PR #3261, opened 2024-12-12
Update Swift build to error on warning

externl:
This PR updates the Swift build to error on warnings. I had to fix on test as a result.

### DIFF
--- a/swift/test/Ice/hold/TestI.swift
+++ b/swift/test/Ice/hold/TestI.swift
@@ -2,6 +2,7 @@
 
 import Dispatch
 import Foundation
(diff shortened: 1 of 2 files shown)


swift/test/Ice/hold/TestI.swift
```diff
@@ -2,10 +2,10 @@
 
 import Dispatch
 import Foundation
-import Ice
+@preconcurrency import Ice
```
pepone, line 5: Can you add a comment, regarding why `@preconcurrency ` is required here?

bernardnormier (approved):

PR #3262, opened 2024-12-12
Fix lints in generated C++ code

bernardnormier:
This PR fix lints in the generated C++, primarily in generated C++ header files.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -6,14 +6,16 @@ Checks:
   modernize-*,
   -modernize-avoid-c-arrays,
   -modernize-use-trailing-return-type,
+  -modernize-concat-nested-namespaces,
   performance-*,
   -performance-avoid-endl
 '
-# WarningsAsErrors: '*'
-HeaderFilterRegex: ''
+WarningsAsErrors:  '*'
+HeaderFilterRegex: '.*'
+ExcludeHeaderFilterRegex: 'include/.*'
 UseColor: true
 FormatStyle: 'file'
-ExtraArgs: ['-std=c++17']
+ExtraArgs: ['-std=c++20']
 CheckOptions:
   modernize-use-nullptr.NullMacros: 'NULL'
   # std::exception_ptr is a cheap to copy, pointer-like type; we pass it by value all the time.
--- a/cpp/src/slice2cpp/Gen.cpp
+++ b/cpp/src/slice2cpp/Gen.cpp
@@ -1227,8 +1227,9 @@ Slice::Gen::ForwardDeclVisitor::visitConst(const ConstPtr& p)
       << getDeprecatedAttribute(p) << "= ";
     writeConstantValue(H, p->type(), p->valueType(), p->value(), _useWstring, p->typeMetadata(), scope);
     H << ';';
-    if (!isConstexprType(p->type()))
+    if (!isConstexprType(p->type())) // i.e. string or wstring
     {
+        // The string/wstring constructor can throw, which produces a clang-tidy lint for const or static objects.
         H << " // NOLINT:cert-err58-cpp";
     }
     H << sp;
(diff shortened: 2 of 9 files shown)


.clang-tidy
```diff
@@ -6,14 +6,16 @@ Checks:
...
 '
-# WarningsAsErrors: '*'
-HeaderFilterRegex: ''
+WarningsAsErrors:  '*'
+HeaderFilterRegex: '.*'
+ExcludeHeaderFilterRegex: 'include/.*'
 UseColor: true
 FormatStyle: 'file'
-ExtraArgs: ['-std=c++17']
+ExtraArgs: ['-std=c++20']
```
bernardnormier, line 18: We use C++20 types (std::span) for a test in C++20 mode; clang-tidy doesn't like -std=c++17 for that test.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1111,9 +1111,14 @@ Slice::Gen::ForwardDeclVisitor::visitEnum(const EnumPtr& p)
         H << "class ";
     }
     H << getDeprecatedAttribute(p) << fixKwd(p->name());
-    if (!unscoped && p->maxValue() <= 0xFF)
+    if (!unscoped)
```
bernardnormier, line 1114: Unfortunately we have no test for unscoped: we should either test it or remove it.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1221,7 +1226,12 @@ Slice::Gen::ForwardDeclVisitor::visitConst(const ConstPtr& p)
       << typeToString(p->type(), false, scope, p->typeMetadata(), _useWstring) << " " << fixKwd(p->name()) << " "
       << getDeprecatedAttribute(p) << "= ";
     writeConstantValue(H, p->type(), p->valueType(), p->value(), _useWstring, p->typeMetadata(), scope);
-    H << ';' << sp;
+    H << ';';
+    if (!isConstexprType(p->type()))
```
pepone, line 1230: I would add a comment here, not clear what is this about.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1543,6 +1553,10 @@ Slice::Gen::ProxyVisitor::visitOperation(const OperationPtr& p)
     }
     H << nl << deprecatedAttribute << retS << ' ' << fixKwd(name) << spar << paramsDecl << contextDecl << epar
       << " const;";
+    if (ret)
+    {
+        H << " // NOLINT:modernize-use-nodiscard";
```
bernardnormier, line 1558: We don't generate a [[nodiscard]] for sync proxy functions returning something.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -2743,12 +2758,20 @@ Slice::Gen::InterfaceVisitor::visitOperation(const OperationPtr& p)
...
     if (ret)
     {
         string typeS = inputTypeToString(ret, p->returnIsOptional(), interfaceScope, p->getMetadata(), _useWstring);
         responseParams.push_back(typeS + " " + returnValueParam);
         responseParamsDecl.push_back(typeS + " ret");
         responseParamsImplDecl.push_back(typeS + " ret");
+
+        if (!amd && !isConst.empty())
+        {
+            noDiscard = "[[nodiscard]] ";
```
pepone, line 2773: What is the relation of nodiscard and const?

pepone (approved):

externl (approved):

cpp/src/slice2cpp/Gen.cpp
bernardnormier: Added comment.

cpp/src/slice2cpp/Gen.cpp
bernardnormier: Added comment.

PR #3269, opened 2024-12-16
Improve doc comment mapping in ice2slice

### LINKED ISSUE #1591: Improve the doc comment conversion in ice2slice
See @InsertCreativityHere comments in https://github.com/zeroc-ice/ice/pull/1576

pepone:
Fix #1591

### DIFF
--- a/cpp/src/ice2slice/Gen.cpp
+++ b/cpp/src/ice2slice/Gen.cpp
@@ -330,7 +330,7 @@ namespace
                 std::map<std::string, StringList>::const_iterator q = parameterDocs.find(param->name());
                 if (q != parameterDocs.end())
                 {
-                    out << nl << "/// @return " << param->name() << ": ";
+                    out << nl << "/// @returns " << param->name() << ": ";
                     for (StringList::const_iterator r = q->second.begin(); r != q->second.end();)
                     {
                         if (r != q->second.begin())
--- a/cpp/src/ice2slice/Gen.cpp
+++ b/cpp/src/ice2slice/Gen.cpp
@@ -371,7 +371,7 @@ namespace
             for (const auto& [name, docs] : comment->exceptions())
             {
                 out << nl << "/// @throws " << name << ": ";
-                for (StringList::const_iterator r =  docs.begin(); r != docs.end();)
+                for (StringList::const_iterator r = docs.begin(); r != docs.end();)
                 {
                     if (r != docs.begin())
                     {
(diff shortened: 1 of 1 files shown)


cpp/src/ice2slice/Gen.cpp
```diff
@@ -299,28 +299,114 @@ namespace
...
                         }
+                        out << (*r);
+                        r++;
                     }
                 }
             }
+
+            // Output parameters
+            parameters = operation->outParameters();
+            for (ParameterList::const_iterator it = parameters.begin(); it != parameters.end(); it++)
```
bernardnormier, line 327: It's not clear why you don't use:
```cpp
for (const auto& param: parameters)
```

cpp/src/ice2slice/Gen.cpp
```diff
@@ -299,28 +299,114 @@ namespace
...
+            {
+                ParameterPtr param = *it;
+                std::map<std::string, StringList>::const_iterator q = parameterDocs.find(param->name());
+                if (q != parameterDocs.end())
+                {
+                    out << nl << "/// @returns " << param->name() << ": ";
```
bernardnormier, line 333: This does not look correct in the situation where we have a single out and no return. It would be good to add some tests.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -299,28 +299,114 @@ namespace
...
+                    }
+                }
+            }
+
+            // Return value
+            StringList returnDocs = comment->returns();
+            if (!returnDocs.empty())
+            {
+                out << nl << "/// @returns";
+                if (operation->returnsMultipleValues())
```
bernardnormier, line 352: I am not a fan of the 'returnsMultipleValues()' function name, since an Ice-Slice operation can't do that.

The absence of doc-comments in Parser.h is also unfortunate.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -299,28 +299,114 @@ namespace
...
+                    out << (*r);
+                    r++;
+                }
+            }
+
+            // Exceptions
+            for (const auto& [name, docs] : comment->exceptions())
+            {
+                out << nl << "/// @throws " << name << ": ";
+                for (StringList::const_iterator r = docs.begin(); r != docs.end();)
```
bernardnormier, line 374: Would be nicer if we could avoid this repeating this pattern over and over. Maybe add a helper?

bernardnormier (approved):

cpp/src/ice2slice/Gen.cpp
pepone: Fixed

cpp/src/ice2slice/Gen.cpp
pepone: Fixed

PR #3273, opened 2024-12-16
Add static C++ ci testing on Linux

externl:
This PR adds ci testing for C++ static libs on Linux. We only run the `Ice/*` tests as the `CONFIGS=static` build does not include the service executables. Is there anything relevant that's missing here?

I didn't add macOS/Windows since we're already low on the number of these runners.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
           - os: ubuntu-24.04
             config: "static"
             build_flags: "CONFIGS=static"
-            test_flags: "--config=static --filter=Ice/"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -52,6 +52,13 @@ jobs:
             test_flags: "--platform=x64 --config=Debug"
             msbuild_project: "msbuild/ice.proj"
 
+          # Static builds
+          - os: ubuntu-24.04
+            config: "static"
+            build_flags: "CONFIGS=static"
+            test_flags: "--config=static --filter=Ice/"
```
pepone, line 59: We can run IceDiscovery/ too

pepone (approved):

PR #3275, opened 2024-12-16
Fix more C++ lints in cpp

bernardnormier:
These lints were fixed "by hand".

### DIFF
--- a/cpp/include/Ice/Proxy.h
+++ b/cpp/include/Ice/Proxy.h
@@ -260,7 +260,7 @@ namespace Ice
         // The copy constructor and assignment operators are to keep GCC happy.
         Proxy(const Proxy&) noexcept = default;
         Proxy& operator=(const Proxy&) noexcept { return *this; } // NOLINT
-        Proxy& operator=(Proxy&&) noexcept { return *this; } // NOLINT
+        Proxy& operator=(Proxy&&) noexcept { return *this; }      // NOLINT
 
     private:
         Prx fromReference(IceInternal::ReferencePtr&& ref) const
--- a/cpp/include/Ice/Proxy.h
+++ b/cpp/include/Ice/Proxy.h
@@ -259,8 +259,8 @@ namespace Ice
 
         // The copy constructor and assignment operators are to keep GCC happy.
         Proxy(const Proxy&) noexcept = default;
-        Proxy& operator=(const Proxy&) noexcept { return *this; }
-        Proxy& operator=(Proxy&&) noexcept { return *this; }
+        Proxy& operator=(const Proxy&) noexcept { return *this; } // NOLINT
+        Proxy& operator=(Proxy&&) noexcept { return *this; } // NOLINT
 
     private:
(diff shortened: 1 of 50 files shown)


cpp/include/Ice/Proxy.h
```diff
@@ -259,8 +259,8 @@ namespace Ice
 
         // The copy constructor and assignment operators are to keep GCC happy.
         Proxy(const Proxy&) noexcept = default;
-        Proxy& operator=(const Proxy&) noexcept { return *this; }
-        Proxy& operator=(Proxy&&) noexcept { return *this; }
+        Proxy& operator=(const Proxy&) noexcept { return *this; } // NOLINT
```
externl, line 262: Can we include the lint name like we're doing in other places.

cpp/src/Glacier2CryptPermissionsVerifier/CryptPermissionsVerifierI.cpp
```diff
@@ -107,9 +107,11 @@ namespace
     {
 #if defined(__APPLE__) || defined(_WIN32)
 
+        // NOLINTBEGIN:cert-err58-cpp
```
externl, line 110: Can we just get rid of the constants? They're only used in this function. and we make a copy all the time.

cpp/src/Ice/Exception.cpp
```diff
@@ -15,7 +15,7 @@
 // For UINTPTR_MAX on Ubuntu Precise
 //
 #ifndef __STDC_LIMIT_MACROS
-#    define __STDC_LIMIT_MACROS
+#    define __STDC_LIMIT_MACROS // NOLINT
```
externl, line 18: Ditto. What doesn't the linter like here?

externl (approved):
Looks good!

cpp/include/Ice/Proxy.h
bernardnormier: These are dummy functions to keep GCC happy. Here, clang-tidy is complaining about not testing for self-assignment, which is irrelevant.

cpp/src/Ice/Exception.cpp
bernardnormier: It doesn't like that we define a __XXX symbol.

cpp/src/Glacier2CryptPermissionsVerifier/CryptPermissionsVerifierI.cpp
bernardnormier: Fixed. In simple cases like these, we should use `const char* const`

cpp/src/DataStorm/TraceUtil.h
```diff
@@ -11,7 +11,8 @@
 #include "SessionI.h"
 #include "TopicI.h"
 
-namespace std
+// TODO: explain why we need to use namespace std here.
```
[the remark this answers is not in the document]
bernardnormier: Probably, but it's not immediately clear to me.

PR #3280, opened 2024-12-17
DataStorm minor fixes

pepone:
This PR includes minor fixes for DataStorm implementation.

### DIFF
--- a/cpp/include/DataStorm/InternalT.h
+++ b/cpp/include/DataStorm/InternalT.h
@@ -136,12 +136,13 @@ namespace DataStormI
     template<typename K, typename V>
     class AbstractFactoryT : public std::enable_shared_from_this<AbstractFactoryT<K, V>>
     {
+        /// A custom deleter to remove the element from the factory when the shared_ptr is deleted.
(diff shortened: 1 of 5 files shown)


externl (approved):

cpp/include/DataStorm/InternalT.h
```diff
@@ -168,7 +172,7 @@ namespace DataStormI
         {
             std::lock_guard<std::mutex> lock(_mutex);
             std::vector<std::shared_ptr<typename V::BaseClassType>> seq;
-            for (auto& v : values)
+            for (auto v : values)
```
pepone, line 175: I should probably revert this. The param `values` is a copy, and the previous version was moving the entries out. Which avoid the copy...

bernardnormier (approved):
> I should probably revert this. The param `values` is a copy, and the previous version was moving the entries out. Which avoid the copy... 

I agree.

cpp/include/DataStorm/InternalT.h
pepone, line 175:
```suggestion
            for (auto& v : values)
```

PR #3284, opened 2024-12-17
DataStorm connection manager fixes

pepone:


cpp/src/DataStorm/ConnectionManager.cpp
```diff
@@ -60,27 +60,20 @@ ConnectionManager::remove(const Ice::ConnectionPtr& connection)
         connection->setCloseCallback(nullptr);
         _connections.erase(p);
     }
+
     exception_ptr ex;
     try
     {
-        [[maybe_unused]] auto _ = connection->getInfo();
+        connection->throwException();
```
pepone, line 67: It is not clear why we were using `getInfo` to retrieve the connection exception. Using throwException seems the more correct way to do that.

externl: I would add an `assert(false);` after this call just in case it doesn't throw.

externl (approved):
Looks good! We use getInfo in a few other places that should be updated as well.

https://github.com/zeroc-ice/ice/blob/b6d18ff9bb7ebd3ea2d93d688d1d649b803830f4/cpp/test/Ice/timeout/AllTests.cpp#L173

cpp/src/DataStorm/ConnectionManager.cpp
bernardnormier: I would add `[[noreturn]]` to throwException.

bernardnormier (approved):

cpp/src/DataStorm/ConnectionManager.cpp
externl: I don't think `throwException` will throw if there's no exception. Like when the connection is still alive.

pepone: > I would add [[noreturn]] to throwException.

Isn't this only for functions that never return. This function can return normally if the connection is still alive.

pepone: > I would add an assert(false); after this call as we expect it to always throw.

I don't think we should. ConnectionManager::remove can also be called from SessionI::destroyImpl and then it is not warantee that the connection would throw.

PR #3287, opened 2024-12-18
Add Android CI configuration

externl:
This PR adds support for Android testing in CI. It will be disabled until we actually fix the tests.


externl:
I've added the CI but disabled it for now until we fix the tests to work with android.

.github/actions/setup-dependencies/action.yml
```diff
@@ -112,6 +112,51 @@ runs:
...
+
+        # For some reason avdmanager is placing the avd into $HOME/.config/.android/avd instead of $HOME/.android/avd
+        echo "ANDROID_AVD_HOME=$HOME/.config/.android" >> $GITHUB_ENV
+      if: matrix.config == 'android'
+
+    - name: Configure Android SDK
+      shell: bash
+      run: |
+        bash -c 'yes 2>/dev/null | sdkmanager --licenses'
+        sdkmanager "platform-tools" "build-tools;34.0.0" "platforms;android-33"
```
pepone, line 152: Is 33 here a typo, shouldn't it be 34?

.github/actions/setup-dependencies/action.yml
```diff
@@ -112,6 +112,51 @@ runs:
...
+    # See https://github.com/marketplace/actions/android-emulator-runner
+    - name: Enable KVM group perms
+      shell: bash
+      run: |
+        echo 'KERNEL=="kvm", GROUP="kvm", MODE="0666", OPTIONS+="static_node=kvm"' | sudo tee /etc/udev/rules.d/99-kvm4all.rules
+        sudo udevadm control --reload-rules
+        sudo udevadm trigger --name-match=kvm
+      if: matrix.config == 'android'
+
+    - name: Install Android SDK
```
pepone, line 129: Can't we use the ones included with the Ubuntu images?

See https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#android

externl: are these env variables already set?

pepone: Seems likely.

pepone (approved):
Looks good!

Issue #3264, opened 2024-12-12
Remove _cpp_ auto-escaping


bernardnormier:
C++ (unlike C#) does not provide a standard mechanism to escape identifiers that conflict with keywords.

slice2cpp auto-escapes identifiers that conflict with C++ keywords by adding a `_cpp_` prefix to the mapped C++ identifier:

https://doc.zeroc.com/ice/3.7/language-mappings/c++11-mapping/client-side-slice-to-c++11-mapping/c++11-mapping-for-identifiers

The resulting identifier (say `_cpp_while`) is not particularly nice.

## Proposal

Update slice2cpp to not auto-escape identifiers that conflict with C++ keywords. When this happens, the generated C++ code won't compile and users should use `cpp:identifier:xxx` to map this Slice identifier to the C++ identifier of their choice.

See #2864.

InsertCreativityHere:
This was removed by #3292 which added a different mechanism for handling identifiers that collide with keywords.

Issue #2864, opened 2024-10-07
Add xxx:identifier metadata to remap identifiers


bernardnormier:
We want to add xxx:identifier metadata (cs:identifier, cpp:identifier, java:identifier) that allows users to remap C#/C++/Java etc. to the specified identifier.

For example:
```slice
interface Greeter 
{
    ["cs:identifier:Greet"] // changes the case in C#
    string greet(string name);
}
```

See also:
https://docs.icerpc.dev/slice2/language-guide/attributes#cs::identifier-attribute

Then once implemented, we want to remove various escaping in the slice compilers. If the generated code does not compile due to the Slice identifier choice of the user, the user should use xxx:identifier to pick the replacement identifier of their choice.

PR #3292, opened 2024-12-18
Add Support for `cpp:identifier` Metadata, and the Groundwork for `xxx:identifier` Metadata in General

InsertCreativityHere:
**This description has been updated and is up-to-date**

This PR implements #3264 and is a proof of concept for #2864 by adding support for this metadata to C++.

It does this by mimicking the 3 fundamental functions in the `libSlice` Parser:
- `name`: returns the identifier of a Slice definition
- `scope`: returns the enclosing scope of a Slice definition (with a trailing `::`)
- `scoped` returns the fully-scoped identifier of a Slice definition (`scope + name`)

Currently on main, all 3 of these return the Slice identifier, and it's up the language mappings to massage them.
Now, we have a separate set of 3 functions: `mappedName`, `mappedScope`, and the less-than-ideal name of `mappedScoped`.
These functions work exactly the same as the original trio, except that they first check for any `xxx:identifier` metadata.
If it's present, we use it's argument instead of the Slice-defined name.

----

In order to make the API smooth, we need a way to inject which language we're working with into `libSlice`.
So I added a new `string_view langaugePrefix` field to `Unit`, which each compiler passes in.
When checking for mapped names, the parser substitutes this string in place of the `xxx` in `xxx:identifier`.

----

### To see exactly how this change has affected the generated code, see
https://github.com/InsertCreativityHere/compiler-comparison/commit/6d0fafb911a4bbaf2473e5fb66ebeb27e3e80254


bernardnormier:
> Now, these functions take an optional string_view langPrefix parameter, allowing callers to signal that they want the identifier for a given mapping.

I find this sub-optimal. All our compilers generate code for only one language, so it would be nicer for the compiler to tell the Parser from the get-go "my language is xxx, please lookup xxx:identifier when resolving mapped identifiers".

InsertCreativityHere:
> We should not do that, just like we don't allow cs::identifier on modules in "new Slice":

It's worth noting that modules are pretty different in IceRPC vs Ice Slice.
In IceRPC they're more of a declaration that applies to a file (like Java's `package` keyword), whereas in Ice, they're just another Slice element like anything else.

Theoretics aside, how do you propose a C++ user fixes this then?
```
module auto { ... }
```
Or even worse, this:
```
module project { module default { ... }}
```

InsertCreativityHere:
> I find this sub-optimal. All our compilers generate code for only one language, so it would be nicer for the compiler to tell the Parser from the get-go "my language is xxx, please lookup xxx:identifier when resolving mapped identifiers".

Alright, I'll see what I can do to improve it. But I'll wait until we figure out the module situation.
Because disallowing this metadata on modules will already let me simplify things a good bit.

bernardnormier:
> Theoretics aside, how do you propose a C++ user fixes this then?
> 
> ```
> module auto { ... }
> ```
> 
> Or even worse, this:
> 
> ```
> module project { module default { ... }}
> ```

It would be logical to add a new `cpp:namespace` directive that renames the mapped C++ namespace. However, this would be inconsistent with the existing `cs:namespace`, which confusingly specifies an enclosing namespace. And it's undesirable to change the semantics of 3.7 metadata in 3.8.

Now, back to new Slice: why do we have `cs::namespace` for modules and `cs::identifier` that doesn't apply to modules?

It's because the semantics are different and more specialized:
https://docs.icerpc.dev/slice2/language-guide/module#c#-mapping

We should do the same for Ice Slice. The question is which metadata do we use? Say we use `cpp:ns` (just an example), the semantics would be:
```slice
["cpp:ns:Reservation::Bundles"]
module Res::Vacation
{
  ... 
}
```

maps to C++ namespace `Reservation::Bundles`.

```slice
["cpp:ns:Foo::Reservation"]
module Res
{
    ["cpp:ns:Bundles"]
    module Vacation { ... }
}
```

maps to:
```cpp
namespace Foo::Reservation
{
    namespace Bundles
    {
    }
}
```

Keep in mind the situation is simpler with new Slice since we don't support the `module  { ... }` syntax:
https://docs.icerpc.dev/slice2/language-guide/module

InsertCreativityHere:
> We should do the same for Ice Slice. The question is which metadata do we use? Say we use cpp:ns (just an example), the > semantics would be:
> ```
> ["cpp:ns:Reservation::Bundles"]
> module Res::Vacation
> {
>  ... 
> }
> ```
> maps to C++ namespace Reservation::Bundles

Yes, but you're just describing the exact semantics of `cpp:identifer`. With my PR you can absolutely write:
```
["cpp:identifier:Foo::Reservation"]
module Res
{
    ["cpp:identifier:Bundles"]
    module Vacation { ... }
}
```
And it will be mapped to:
```
namespace Foo::Reservation {
    namespace Bundles { ... }
}
```
Since we're targeting C++17, which supports nested namespace definitions.

----

However... There might be a bug here with putting `cpp:identifer` on _nested_ modules.
I'm not sure if it will affect the last module, or the entire nested module... I need to check.

Using a name like `cpp:identifier`, it's clear that this metadata just replaces the Slice identifier with whatever you specify. This avoids the appearance of being inconsistent with `cs:namespace`, since they have totally different names.

And it's easier to explain `cpp:identifier` without a special carve-out for modules: `You can put it on anything with a Slice identifier, and the compiler will just copy/paste whatever you provided in the metadata in place of that Slice identifier.`
No need to say: `... except for modules, in which case you should use this other metadata instead`.

----

Back to new Slice though, modules are quite different between Ice and IceRPC.
So I don't think it's crazy to say "the way we handle metadata on them should be different".

And while the design of modules in IceRPC is simpler, I don't think that the way we handle and treat modules in Ice is wrong, just different. IceRPC uses the Java approach (`package Hello;`), and Ice uses the C++ approach (`namespace Hello {...}`).

InsertCreativityHere:
Also, as an aside, if we allow `cs:identifier` on modules in Slice, we could actually phase out/deprecate `cs:namespace`.
Anything you can do with `cs:namespace` you can also do with `cs:identifier`, but less confusingly, and with even more control.

If you have:
```
// module M is mapped to namespace MyWidget.M
["cs:namespace:MyWidget"]
module M {}
```
It is completely backwards compatible to update this to:
```
// module M is mapped to namespace MyWidget.M
["cs:identifier:MyWidget.M"]
module M {}
```

bernardnormier:
The argument for cs::namespace in new Slice is not an identifier but a scoped name.

You think it's ok to use lang:identifier in Ice Slice on modules where the argument of this "identifier" metadata is not an identifier?

With your example:
```slice
// module M is mapped to namespace MyWidget.M
["cs:identifier:MyWidget.M"] // MyWidget.M - the argument of cs:identifier - is not an identifier.
module M {}
```

bernardnormier:
Interestingly, cppreference names ns-name "identifier":
https://en.cppreference.com/w/cpp/language/namespace#Namespaces

InsertCreativityHere:
Your point isn't lost on me.
To be entirely honest, the fact that this works is just an accident!
I didn't think about this case, let alone intentionally design for this.

But... yes, I do think it's _okay_. Although I agree it's not _great_.

----

But think mechanistically. With your idea, we'd have two metadata: `cpp:identifier` and let's say `cpp:ns` (hypothetically).
`cpp:identifier` can go on everything except modules.
`cpp:ns` can only go on modules.

And they would have the following effect:
```
["cpp:ns:There"]
module Hello
{
    ["cpp:identifier:MyThing"]
    struct Thing { int i; }
}
```
when compiled with `slice2cpp` maps to
```
namespace There
{
    struct MyThing { ::std::int16_t i; }
}
```
These effects are indistinguishable. In both cases, instead of using the Slice-provided identifier, a different name was generated based on some metadata.
And if we implemented both of these in `slice2cpp`, those implementations would be identical.

So, these sure _feel_ like the same metadata to me. Just that it was artificially split into 2.
And now we have two metadata to document, and test, and support. Instead of just 1.
And users who know about `cpp:identifier` will be annoyed when it doesn't work on modules (for no reason they can see).

InsertCreativityHere:
My example is a little contrived though, because we're only ever dealing with identifiers.
Obviously the 'interesting' case is when we have _nested module syntax_. But, before that, let's admit that most users do not use nested module syntax. So, for most users, who are using 'normal' module syntax, there is truly no problem here.

But okay. Nested module syntax looks like this:
```
module Hello::There {...}
```
An actual question: What would you call the `Hello::There` here?

Internally, in the Slice compiler, we call it a `scoped identifier`: https://github.com/zeroc-ice/ice/blob/50fbf7e95f9387e4b0488daf695f185eef3fed47/cpp/src/Slice/Grammar.y#L382 This is definitely not the same as `identifier`, but it is still a _kind of identifier_.

And as you point out, the cppreference calls it an `ns-name`. But, `name` is really just a synonym for identifier. We use 'identifier' and 'name' interchangably in the Slice compiler (although I wish we were more consistent).

Yes, `Hello::There` isn't exactly an `identifier`, but it's in the same ballpark as one.
And it's definitely close enough that any of users will _understand_ what we mean.

Also, interestingly, cppreference has been using `ns_name` since at least 2012.
But back then they were using `name` in the explanation instead of `identifier`.
https://web.archive.org/web/20121024204606/en.cppreference.com/w/cpp/language/namespace
(Note that the page will be a little slow to load).

bernardnormier:
Regardless of which metadata we eventually select for modules, I recommend you rework this PR to focus exclusively on cpp:identifier for constructs other than modules.

Metadata for modules / namespaces / scopes deserves its own PR. For now, just rename the C++ "keyword" namespaces to something else.

cpp/src/Slice/MetadataValidation.cpp
```diff
@@ -83,9 +83,9 @@ Slice::validateMetadata(const UnitPtr& p, string_view prefix, map<string, Metada
     MetadataInfo deprecatedInfo = {
         .validOn =
             {typeid(InterfaceDecl),
```
InsertCreativityHere, line 85: I just swapped the order here to be consistent with other lists.

cpp/src/Slice/Parser.cpp
```diff
@@ -4555,19 +4551,25 @@ Slice::DataMember::DataMember(
 // ----------------------------------------------------------------------
```
InsertCreativityHere, line 4552: All the code below this is to allow the compilers to inject their `languagePrefix` into `libSlice`.

cpp/src/Slice/Parser.cpp
```diff
@@ -1050,12 +1052,6 @@ Slice::Contained::Contained(const ContainerPtr& container, const string& name)
       _container(container),
       _name(name)
 {
-    ContainedPtr cont = dynamic_pointer_cast<Contained>(_container);
-    if (cont)
-    {
-        _scoped = cont->scoped();
-    }
-    _scoped += "::" + _name;
```
InsertCreativityHere, line 1058: There isn't much point to `_scoped` now.

Most of the calls to `scoped()` and `scope()` are for mapped identifiers, and now we have to compute their values on the fly. So, there's even less point to caching the `_scoped` name up-front. So, I removed it.

cpp/src/Slice/Parser.h
```diff
@@ -363,10 +363,11 @@ namespace Slice
     {
     public:
         [[nodiscard]] ContainerPtr container() const;
-        [[nodiscard]] std::string name() const;
-        [[nodiscard]] std::string scoped() const;
-        [[nodiscard]] std::string scope() const;
-        [[nodiscard]] std::string flattenedScope() const;
```
InsertCreativityHere, line 369: `flattenedScope` was only used by `slice2cpp`, so I moved the logic into that compiler.

cpp/src/Slice/Parser.h
```diff
@@ -453,7 +453,6 @@ namespace Slice
         EnumeratorList enumerators() const;
         EnumeratorList enumerators(const std::string& identifier) const;
         ContainedList contents() const;
-        std::string thisScope() const;
```
InsertCreativityHere, line 456: This function is only used internally.
Instead of keeping it public and adding the `mappedName` flag, I just made it protected, so only things in `libSlice` (which don't care about mapped names) can access it.

cpp/src/slice2cpp/CPlusPlusUtil.cpp
```diff
@@ -176,19 +176,6 @@ namespace
...
-        const TypePtr& type,
-        bool optional,
-        const string& scope,
-        const string& fixedName,
-        const MetadataList& metadata,
-        TypeContext typeCtx)
-    {
-        string s = typeToString(type, optional, scope, metadata, typeCtx);
-        out << nl << s << ' ' << fixedName << ';';
-    }
```
InsertCreativityHere, line 190: This function is only 2 lines... but takes 7 arguments. I inlined it in the few places where it was used.

cpp/test/Slice/escape/Client.cpp
```diff
@@ -9,19 +9,15 @@
 
 using namespace std;
```
InsertCreativityHere, line 10: I changed alot of stuff in the escape test.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -905,6 +920,25 @@ Slice::Gen::validateMetadata(const UnitPtr& u)
...
+             typeid(Dictionary),
+             typeid(Enum),
+             typeid(Enumerator),
+             typeid(Const),
+             typeid(Parameter),
+             typeid(DataMember)},
+        .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
+    };
+    knownMetadata.emplace("cpp:identifier", std::move(identifierInfo));
+
```
InsertCreativityHere, line 941: This is the entry which allows the new `cpp:identifier` through the metadata validator.
Note that as-of-now, we do not support `cpp:identifier` on Modules.
But it is supported on every other relevant Slice construct.

cpp/src/Slice/Parser.h
```diff
@@ -363,10 +363,14 @@ namespace Slice
     {
     public:
         [[nodiscard]] ContainerPtr container() const;
+
         [[nodiscard]] std::string name() const;
         [[nodiscard]] std::string scoped() const;
         [[nodiscard]] std::string scope() const;
-        [[nodiscard]] std::string flattenedScope() const;
+        [[nodiscard]] std::string mappedName() const;
```
bernardnormier, line 370: It would be helpful to add doc-comments.

cpp/src/Slice/Parser.h
```diff
@@ -446,14 +449,13 @@ namespace Slice
         TypeList lookupTypeNoBuiltin(const std::string& identifier, bool emitErrors, bool ignoreUndefined = false);
         ContainedList lookupContained(const std::string& identifier, bool emitErrors);
         ExceptionPtr lookupException(const std::string& identifier, bool emitErrors);
-        [[nodiscard]] UnitPtr unit() const;
```
bernardnormier, line 449: I don't see any reason for the removal of these ``[[nodiscard]]``

cpp/src/Slice/Parser.h
```diff
@@ -483,6 +485,8 @@ namespace Slice
         }
 
     protected:
+        std::string thisScope() const;
```
bernardnormier, line 488: missing ``[nodiscard]``

cpp/src/Slice/Parser.h
```diff
@@ -1039,11 +1046,12 @@ namespace Slice
         [[nodiscard]] std::set<std::string> getTopLevelModules(const std::string& file) const;
 
     private:
-        Unit(bool all, MetadataList defaultFileMetadata);
+        Unit(std::string_view languagePrefix, bool all, MetadataList defaultFileMetadata);
```
bernardnormier, line 1049: I find this use of string_view over the top. Using strings is preferable, especially for something you adopt.

cpp/src/Slice/Parser.cpp
```diff
@@ -563,27 +564,49 @@ Slice::Contained::name() const
...
 string
-Slice::Contained::flattenedScope() const
+Slice::Contained::mappedName() const
 {
-    string s = scope();
-    string::size_type pos = 0;
-    while ((pos = s.find("::", pos)) != string::npos)
+    // First check if any 'xxx:identifier' has been applied to this element.
+    // If so, we return that instead of the element's Slice identifier.
+    static const string metadata = string(_unit->languagePrefix()) + ":identifier";
```
bernardnormier, line 586: I find this `static const` dubious. Yes, all the units in your program will most likely have the same language prefix, but at the same time, you made language prefix a per-unit field, not a global field.

cpp/src/Slice/Parser.cpp
```diff
@@ -4575,19 +4592,25 @@ Slice::DataMember::DataMember(
...
+Slice::Unit::createUnit(string_view languagePrefix, bool all, const StringList& defaultFileMetadata)
 {
     MetadataList defaultMetadata;
     for (const auto& metadataString : defaultFileMetadata)
     {
         defaultMetadata.push_back(make_shared<Metadata>(metadataString, "<command-line>", 0));
     }
 
-    UnitPtr unit{new Unit{all, std::move(defaultMetadata)}};
+    UnitPtr unit{new Unit{languagePrefix, all, std::move(defaultMetadata)}};
```
bernardnormier, line 4603: The code above is a mystery to me. Would be nice to add a comment.

cpp/src/ice2slice/Main.cpp
```diff
@@ -176,7 +176,7 @@ compile(const vector<string>& argv)
         }
         else
         {
-            UnitPtr p = Unit::createUnit(false);
+            UnitPtr p = Unit::createUnit("icerpc", false);
```
bernardnormier, line 179: How does this work since icerpc is not a mapped language in Parser.cpp?

cpp/src/Slice/Parser.cpp
```diff
@@ -5037,15 +5060,16 @@ Slice::Unit::getTopLevelModules(const string& file) const
...
     : SyntaxTreeBase(nullptr),
       Container(nullptr),
+      _languagePrefix(languagePrefix),
       _all(all),
       _defaultFileMetadata(std::move(defaultFileMetadata)),
       _errors(0),
       _currentIncludeLevel(0)
-
 {
+    assert(binary_search(&languages[0], &languages[sizeof(languages) / sizeof(*languages)], _languagePrefix));
```
bernardnormier, line 5072: How does this work for 'icerpc'?

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -242,33 +242,47 @@ namespace
...
-                break;
+                return name + "_";
             }
         }
-        return r;
+        return name;
     }
 
     /// Returns a doxygen formatted link to the provided Slice identifier.
+    /// TODO we need to add a way for the doc-comment generation to use 'cpp' identifier!
```
bernardnormier, line 275: It doesn't make sense for C++ doc-comment links to link to Slice identifiers.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -896,6 +912,25 @@ Slice::Gen::validateMetadata(const UnitPtr& u)
     };
     knownMetadata.emplace("cpp:ice_print", std::move(icePrintInfo));
 
+    // "cpp:identifier"
+    MetadataInfo identifierInfo = {
+        .validOn =
+            {typeid(InterfaceDecl),
+             typeid(Operation),
+             typeid(ClassDecl),
+             typeid(Slice::Exception),
```
bernardnormier, line 921: Why is Exception qualified?

cpp/test/Ice/servantLocator/TestAMD.ice
```diff
@@ -28,8 +28,8 @@ exception TestImpossibleException
 
     void unknownExceptionWithServantException();
 
-    string impossibleException(bool throw) throws TestImpossibleException;
-    string intfUserException(bool throw) throws TestIntfUserException, TestImpossibleException;
+    string impossibleException(["cpp:identifier:_cpp_throw"] bool throw) throws TestImpossibleException;
```
bernardnormier, line 31: I would not use leading underscores for the new C++ identifiers.

bernardnormier (commented):
For the new mapped identifiers, I would use camel case and no leading underscore, for example:

Slice param `throw` => C++ param `cppThrow`.

cpp/src/Slice/Parser.h
InsertCreativityHere: This wasn't intentional.
I must of missed this when I was resolving the merge conflicts with main.

InsertCreativityHere:
> For the new mapped identifiers, I would use camel case and no leading underscore

Fine with me. I was following our 'old' escaping convention, which I agree isn't very nice.
But if we're thinking about it, could we just do something even simpler like adding a trailing `_`?
Instead of prepending the language name.
These are our own identifiers, so there's no risk of colliding with other names. We're in full control here.

At some point, we should decide what escaping convention we want to use in other languages before I start implementing them.
Some will be pretty obvious like `@<ident>` in C#. But others will be less obvious.

cpp/src/slice2cpp/Gen.cpp
InsertCreativityHere: Just to be clear that we're dealing with the Slice type, and not `Ice::Exception` or any other base exception type.

Is is strictly necessary? No. But I think it makes the code clearer at a glance.
And it seems like we generally qualify `Ice::Exception` and `Slice::Exception` if you `rg` for `::Exception\b`.

cpp/src/slice2cpp/Gen.cpp
InsertCreativityHere: I agree, but this is going to need to be it's own PR.

To fix this, the doc-comment parsing code need to ability to look up Slice definitions by name, and check metadata on them. Right now it has no way to access any of this. And adding this functionality will affect all the languages since this comment-parsing code is shared.

This is a good thing to do anyways, so we can report broken links.
But it's going to a sub-project in-of-itself.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: It didn't, but I fixed it. And opened an issue to run `ice2slice` in CI.
Right now, we have 0 testing of this tool, and it's way too easy to break it without realizing...

cpp/src/ice2slice/Main.cpp
InsertCreativityHere: It doesn't. Good catch that I forgot to add it to the list of known languages.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: This is not my code, but as I read it, we're creating a cyclic `shared_ptr` here.
Which is just leaking memory, right?

cpp/src/Slice/Parser.cpp
InsertCreativityHere: I find it very unlikely that this would actually cause any problems.
But, I removed the `static` anyways.

So we'll re-compute this string every time the function is called.
Thankfully the compilers are already pretty fast, so it's fine I guess.

cpp/src/Slice/Parser.h
InsertCreativityHere: Fine, I switched it to use `string` instead.

cpp/src/Slice/Parser.h
InsertCreativityHere: Added.

cpp/src/Slice/Parser.h
InsertCreativityHere: I added these `[[nodiscard]]`s back. And added it to the functions above which should also be marked with it.

These functions aren't marked `const` because they can lead to errors being reported, which causes `errors++;`.
So I'm not surprised the automatic fix didn't mark them `[[nodiscard]]`. But, we should never be called `createStruct` and throwing away the value it returns. I'm sure there's other places that need this fix as well, but I didn't want to get carried away here.

cpp/src/Slice/Parser.h
InsertCreativityHere: I added some short doc-comments to each of these 6 functions.

cpp/src/Slice/Parser.cpp
pepone: SyntaxTreeBase has a _unit, and Unit is a SyntaxTreeBase so it has a unit.

We can consider replace this by a virtual `unit()` method, which for Unit can return `shared_from_this`

InsertCreativityHere: I think another option is just not having `Unit` inherit from `SyntaxTreeBase`.
It's a pretty small API surface, most of which isn't useful for `Unit`:
```
virtual void destroy();                   // Only sets `_unit = nullptr;`
UnitPtr unit() const;                     // Useless on `Unit`, it already has access to itself.
DefinitionContextPtr definitionContext() const;     // Always `nullptr` for `Unit`
virtual void visit(ParserVisitor* visitor);         // Only useful function
```
So, it'd actually be nice to remove this useless inheritance, or simplify it somehow.

Unfortunately, we use `SyntaxTreeBasePtr` as a base type for many places, so this would require a good deal of reshuffling.

cpp/src/slice2cpp/Gen.cpp
bernardnormier: ok

cpp/src/Slice/Parser.cpp
bernardnormier: I was talking about this code:
```cpp
    MetadataList defaultMetadata;
    for (const auto& metadataString : defaultFileMetadata)
    {
        defaultMetadata.push_back(make_shared<Metadata>(metadataString, "<command-line>", 0));
    }
```

cpp/src/slice2rb/Ruby.cpp
```diff
@@ -172,7 +172,7 @@ Slice::Ruby::compile(const vector<string>& argv)
                 return EXIT_FAILURE;
             }
 
-            UnitPtr u = Unit::createUnit(false);
+            UnitPtr u = Unit::createUnit("ruby", false);
```
externl, line 175: Ditto

bernardnormier:
After thinking about it, I find the new design - where we attach a single language mapping to Unit so that Unit can provided mapped names for that language - rather problematic.

Yes, it works for the slice compilers that generate a single language (per compiler). And it works for now for ice2slice, since we support only one language with (new) Slice. But it's not a clean solution. It won't work when we support Python and JS with Slice.

I think it would be cleaner to not introduce this tie-up, and just pass the language name as needed. So instead of having:
```cpp
[[nodiscard]] std::string mappedName() const; // uses Unit's configured language mapping
```

we should simply provide:
```cpp
[[nodiscard]] std::string mappedName(std::string_view languageMapping) const; 
```

And you can have a global-like variable in each Slice compiler for languageMapping.

bernardnormier:
Re: associating a language mapping with Unit

This PR introduces this association, so that Unit can provide various helper functions like:
```cpp
        /// @return The mapped identifier that this element will use in the target language.
```

(BTW, this is not a correct doc-comment. A doc-comment should be @return - only)

This works fine for Slice compilers like slice2cpp, slice2cs etc.

It's not usable by ice2slice - especially once we support multiple language mappings - but this doesn't matter as we wouldn't want to use this helper to figure out the mapped identifiers in ice2slice. 

If we could use such a helper and called it from ice2slice, we would generate xxx::identifier attributes for pretty much all converted code, which is obviously undesirable.

cpp/src/ice2slice/Main.cpp
bernardnormier: I don't think adding 'icerpc' to the list of known language mappings is a good fix. You could make this first parameter std::optional or treat "" as a special value.

InsertCreativityHere:
> It's not usable by ice2slice

My thinking was that it would be usable by ice2slice. Say you have this in your Slice definition:
```
struct uint8 { byte b; }
```
It would generate an uncompilable `.slice` file. And you could consider using `["icerpc:identifier:UInt8"]` instead.
I admit it's not the most compelling reason, since most times you'll only run `ice2slice` once, then fix edge cases like this by hand.
But I don't think this idea is broken or unusable. Just maybe not a "high value" ability.

----

TL;DR: I removed this.

cpp/src/slice2py/Python.cpp
```diff
@@ -614,7 +614,7 @@ Slice::Python::compile(const vector<string>& argv)
             }
             else
             {
-                UnitPtr u = Unit::createUnit(all);
+                UnitPtr u = Unit::createUnit("python", all);
```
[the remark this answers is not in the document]
InsertCreativityHere: Ditto.

cpp/src/slice2py/Python.cpp
```diff
@@ -563,7 +563,7 @@ Slice::Python::compile(const vector<string>& argv)
                 return EXIT_FAILURE;
             }
 
-            UnitPtr u = Unit::createUnit(false);
+            UnitPtr u = Unit::createUnit("python", false);
```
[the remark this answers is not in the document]
InsertCreativityHere: I agree, but that's just how it is. `slicepy` internally uses `python` for it's metadata.
It's unfortunate, but we can't realistically change it now.

cpp/src/Slice/Parser.h
```diff
@@ -1039,11 +1056,12 @@ namespace Slice
         [[nodiscard]] std::set<std::string> getTopLevelModules(const std::string& file) const;
 
     private:
-        Unit(bool all, MetadataList defaultFileMetadata);
+        Unit(std::string languagePrefix, bool all, MetadataList defaultFileMetadata);
```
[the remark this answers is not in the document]
InsertCreativityHere: Renamed to the more general "languageName"

cpp/test/Ice/servantLocator/TestAMD.ice
InsertCreativityHere: I renamed all the identifiers outside of the `escape` test.

If you really want, I can also do the `escape` test, but this will render the file utterly unreviewable.
And I don't think the names for that test matters, since that test is already using insane names like `break` and `return`.

cpp/src/ice2slice/Main.cpp
InsertCreativityHere: I removed `icerpc`. Now `ice2slice` just passes the empty string.
And I added an assert to `mappedName` which enforces that the language string is non-empty.
Since `ice2slice` should never be using mapped names anyways.

bernardnormier (approved):
Looks good.

externl (approved):

PR #3293, opened 2024-12-19
DataStorm linter suggested fixes

pepone:
This PR:
- Includes fixes for issues reported by clang-tidy.
- Added `[[nodiscard]]` to more DataStorm APIs where it make sense to do so.
- Added `using namespace Ice;` to DataStorm implementation.


.clang-tidy
```diff
@@ -16,8 +16,8 @@ HeaderFilterRegex: '.*'
 ExcludeHeaderFilterRegex: 'include/lmdb.h|Grammar.h'
 UseColor: true
 FormatStyle: 'file'
-ExtraArgs: ['-std=c++20']
+ExtraArgs: ['-std=c++17']
```
pepone, line 19: Otherwise we get lints for issues that require using C++ 20 and we are still using C++ 17 with GCC.

cpp/include/DataStorm/DataStorm.h
```diff
@@ -1237,22 +1237,17 @@ namespace DataStorm
...
         return MultiKeyWriter<K, V, UT>(topic, {}, std::move(name), config);
     }
 
-}
-
-//
-// Public template based API implementation
-//
-
-namespace DataStorm
```
pepone, line 1254: no point in declaring the same namespace twice.

cpp/include/DataStorm/Node.h
```diff
@@ -140,7 +140,7 @@ namespace DataStorm
          * provided the Node will use the default callback executor that executes callback in a dedicated thread.
          */
         Node(
-            Ice::CommunicatorPtr communicator,
+            const Ice::CommunicatorPtr& communicator,
```
pepone, line 143: Because we always use it as a reference.

cpp/include/DataStorm/Types.h
```diff
@@ -83,9 +83,9 @@ namespace DataStorm
             std::optional<int> sampleCount = std::nullopt,
             std::optional<int> sampleLifetime = std::nullopt,
             std::optional<ClearHistoryPolicy> clearHistory = std::nullopt) noexcept
-            : sampleCount(std::move(sampleCount)),
-              sampleLifetime(std::move(sampleLifetime)),
-              clearHistory(std::move(clearHistory))
+            : sampleCount{sampleCount},
```
pepone, line 86: see 

```
std::move of the variable 'sampleCount' of the trivially-copyable type 'std::optional<int>' has no effect; remove std::move() [performance-move-const-arg,-warnings-as-errors]
```

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -444,7 +446,6 @@ DataElementI::getConnectedElements() const
 {
     unique_lock<mutex> lock(_parent->_mutex);
     vector<string> elements;
-    elements.reserve(_listeners.size());
```
pepone, line 447: This reserve doesn't match the final size, there is an inner loop.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -465,11 +466,12 @@ DataElementI::onConnectedKeys(
...
         vector<shared_ptr<Key>> keys;
-        for (const auto& key : _connectedKeys)
+        keys.reserve(_connectedKeys.size());
+        for (const auto& [key, _] : _connectedKeys)
         {
-            keys.push_back(key.first);
+            keys.push_back(key);
         }
-        _executor->queue([self = shared_from_this(), init, keys = std::move(keys)] { init(std::move(keys)); }, true);
+        _executor->queue([init = std::move(init), keys = std::move(keys)]() mutable { init(std::move(keys)); }, true);
```
pepone, line 474: Without marking the lambda as mutable, the move keys doesn't have any effect.

cpp/src/DataStorm/ForwarderManager.h
```diff
@@ -19,25 +19,25 @@ namespace DataStormI
...
+        template<typename Prx, std::enable_if_t<std::is_base_of_v<Ice::ObjectPrx, Prx>, bool> = true>
         Prx add(std::function<void(Ice::ByteSeq, const Ice::Current&)> forwarder)
         {
             return add<Prx>(
                 [forwarder = std::move(forwarder)](
                     Ice::ByteSeq inParams,
-                    Response response,
-                    Exception exception,
+                    std::function<void(bool, const Ice::ByteSeq&)> response,
+                    std::function<void(std::exception_ptr)> exception,
```
pepone, line 40: I added `std::function` to performance-unnecessary-value-param.AllowedTypes

But this doesn't work with the alias.

cpp/src/DataStorm/Instance.h
```diff
@@ -114,16 +112,16 @@ namespace DataStormI
...
         }
 
-        void scheduleTimerTask(IceInternal::TimerTaskPtr task, const std::chrono::milliseconds& delay)
+        void scheduleTimerTask(const IceInternal::TimerTaskPtr& task, const std::chrono::milliseconds& delay)
         {
             std::unique_lock<std::mutex> lock(_mutex);
             if (_timer)
             {
-                _timer->schedule(std::move(task), delay);
+                _timer->schedule(task, delay);
```
pepone, line 120: task is a const reference

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -47,15 +48,13 @@ namespace
...
     : _instance(instance),
-      _nextPublisherSessionId{0},
-      _nextSubscriberSessionId{0},
-      _proxy{instance->getObjectAdapter()->createProxy<NodePrx>({Ice::generateUUID(), ""})},
+      _proxy{instance->getObjectAdapter()->createProxy<NodePrx>(Identity{.name = generateUUID(), .category = ""})},
       // The subscriber and publisher collocated forwarders are initalized here to avoid using a nullable proxy. These
       // objects are only used after the node is initialized and are removed in destroy implementation.
       _publisherForwarder{instance->getCollocatedForwarder()->add<PublisherSessionPrx>(
-          [this](Ice::ByteSeq inParams, const Ice::Current& current) { forwardToPublishers(inParams, current); })},
+          [this](const ByteSeq& inParams, const Current& current) { forwardToPublishers(inParams, current); })},
```
pepone, line 55: The forwarding methods take a const reference for inParams.

cpp/src/DataStorm/NodeSessionI.cpp
```diff
@@ -26,17 +27,17 @@ namespace
     public:
         NodeForwarder(
             shared_ptr<NodeSessionManager> nodeSessionManager,
-            shared_ptr<NodeSessionI> nodeSession,
+            const shared_ptr<NodeSessionI>& nodeSession,
             NodePrx node)
             : _nodeSessionManager(std::move(nodeSessionManager)),
-              _nodeSession(std::move(nodeSession)),
+              _nodeSession(nodeSession),
```
pepone, line 33: `_nodeSession` is a weak_ptr. The weak_ptr constructor takes a const reference to a shared_ptr.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -477,7 +475,7 @@ SessionI::initSamples(int64_t topicId, DataSamplesSeq samplesSeq, const Ice::Cur
 
                     vector<shared_ptr<Sample>> samplesI;
                     samplesI.reserve(samples.samples.size());
-                    auto sampleFactory = topic->getSampleFactory();
+                    const auto& sampleFactory = topic->getSampleFactory();
```
pepone, line 478: getSampleFactory returns a const reference.

cpp/src/DataStorm/TraceUtil.h
```diff
@@ -226,16 +223,7 @@ namespace DataStormI
     class Trace : public Ice::Trace
     {
     public:
-        Trace(std::shared_ptr<TraceLevels> traceLevels, const std::string& category)
-            : Ice::Trace(traceLevels->logger, category)
-        {
-        }
-    };
-
-    class Warning : public Ice::Warning
```
pepone, line 235: The Warning class was never used.

cpp/test/DataStorm/api/Writer.cpp
```diff
@@ -230,13 +230,9 @@ void ::Writer::run(int argc, char* argv[])
             reader.waitForNoWriters();
             [[maybe_unused]] auto _ = reader.getConnectedWriters();
             [[maybe_unused]] auto connectedKeys = reader.getConnectedKeys();
-            reader.getAllUnread();
+            [[maybe_unused]] auto allUnread = reader.getAllUnread();
             reader.waitForUnread(0);
             [[maybe_unused]] bool hasUnread = reader.hasUnread();
-            if (false)
```
pepone, line 236: I don't see the point of this `if (false)`, this API test is kind of weir with lots of unused params.

cpp/src/DataStorm/ForwarderManager.h
bernardnormier: > I added std::function to performance-unnecessary-value-param.AllowedTypes

I would not do that. When you copy of std::function that holds a lambda, you copy its captures.

We often pass std::function by value because we move them around. In some cases, it's ok to pass them by const& as clang-tidy suggests.

cpp/include/DataStorm/Node.h
bernardnormier: The signature is not / should not be about the implementation but what we signal to the caller.

Here, const& tells the caller: I am reading this communicator in this ctor but I don't keep a refcount on the communicator. Whereas, by value means "I keep a refcount".

I would expect the Node to keep a refcount on the communicator, possibly indirectly by holding an Ice::InstancePtr.

cpp/src/DataStorm/Node.cpp
```diff
@@ -93,13 +94,13 @@ Node::Node(optional<string_view> configFile, function<void(function<void()> call
...
-Node::Node(Ice::CommunicatorPtr communicator, function<void(function<void()> call)> customExecutor)
-    : Node(std::move(communicator), std::move(customExecutor), false)
+Node::Node(const CommunicatorPtr& communicator, function<void(function<void()> call)> customExecutor)
+    : Node(communicator, std::move(customExecutor), false)
 {
 }
 
 Node::Node(
-    Ice::CommunicatorPtr communicator,
+    const CommunicatorPtr& communicator,
```
bernardnormier, line 103: I assume _instance below keeps a refcount on communicator, so it should be by value. You may want to fix the DataStormI::Instance ctor - you actually already fixed it!

cpp/test/DataStorm/api/Writer.cpp
bernardnormier: I am surprised the compiler doesn't complain.

cpp/src/DataStorm/TopicI.h
```diff
@@ -28,7 +28,7 @@ namespace DataStormI
     public:
         TopicI(
             std::shared_ptr<Instance>,
-            std::shared_ptr<TopicFactoryI>,
+            const std::shared_ptr<TopicFactoryI>&,
```
bernardnormier, line 31: Here, switching to `const shared_ptr&` makes sense: we don't adopt a refcount but create and keep a weak_ptr from this `const shared_ptr&`.

bernardnormier (approved):
Looks good except for the addition of std::function to performance-unnecessary-value-param.AllowedTypes.

externl (approved):
Looks good!

cpp/src/DataStorm/ForwarderManager.h
pepone: reverted this, and fixed the forwarder manager to use pass by reference.

Issue #3288, opened 2024-12-18
Java Android controller build failure


externl:
From my draft Android CI PR: https://github.com/zeroc-ice/ice/actions/runs/12400418944/job/34617551843

```

> Task :compileDebugJavaWithJavac
/home/runner/work/ice/ice/java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerApp.java:289: error: cannot find symbol
                com.zeroc.Ice.SSL.Plugin plugin =
                                 ^
  symbol:   class Plugin
  location: package com.zeroc.Ice.SSL
/home/runner/work/ice/ice/java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerApp.java:290: error: cannot find symbol
                        (com.zeroc.Ice.SSL.Plugin)
                                          ^
  symbol:   class Plugin
  location: package com.zeroc.Ice.SSL
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
2 errors

> Task :compileDebugJavaWithJavac FAILED

> Task :desugarDebugFileDependencies
22 actionable tasks: 22 executed
FAILURE: Build failed with an exception.
```

This plugin no longer exists and we'll need to update the controller to use the new native Java APIs.

https://github.com/zeroc-ice/ice/blob/5f38d6218c6ba97ac55981f59866c8af6fc0cff4/java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerApp.java#L286-L299

PR #3304, opened 2024-12-23
Fix android IceSSL testing

pepone:
This PR fixes Android SSL testing, see #3288

With Ice 3.7, we relied on delaying IceSSL plug-in initialization and then manually setting the key store or trust store from the test controller helper in the communicator-initialized hook.

With Ice 3.8, this approach doesn't work because IceSSL is no longer a plug-in. It’s now a built-in transport, and its configuration must either be specified using IceSSL properties (as before) or when the communicator is created (client side) or the object adapter is created (server side).

Using the new native SSL APIs would require reviewing all object adapter creation in the test suite to correctly set the SSL configuration. For the client side, communicators are already created with a helper method, which simplifies things.

To keep it simpler, I added a new `resourceLoader` member to `InitializationData` that allows loading certificates using a custom loader; in this case, we want to load them from Android resources.

If we don’t want to take this approach of adding a new `resourceLoader`, an alternative would be to set the existing `classLoader` and wrap it with a decorator that re-implements `getResourceAsStream`.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InitializationData.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InitializationData.java
@@ -76,4 +76,14 @@ public final class InitializationData implements Cloneable {
      * the SSLEngineFactory.
      */
     public SSLEngineFactory clientSSLEngineFactory;
+
+    /**
+     * A user-supplied function used to loads resources (e.g., key stores, trust stores) given a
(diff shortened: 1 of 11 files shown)


externl (approved):
Seems good to me!

pepone:
I think it is indeed better to decorate the class loader and not add additional InitializationData members for this. Doesn't seems that this would be very useful beside this testing scenario.

bernardnormier (commented):
I agree: a custom class loader would be nicer.

PR #3307, opened 2024-12-26
Executor service fixes

### LINKED ISSUE #3300: MaxConnection tests Android failure
```
*** [22/40] Running java/Ice/maxConnections tests ***
[ running client/server test - 12/22/24 19:03:41 ]
- Config: android,Debug,x64
testing the creation of 100 connections... com.zeroc.Ice.CloseTimeoutException: Close timed out.
        at com.zeroc.Ice.QueueExecutorService.execute(QueueExecutorService.java:64)
        at com.zeroc.Ice.QueueExecutorService.executeNoThrow(QueueExecutorService.java:28)
        at com.zeroc.Ice.ConnectionI.close(ConnectionI.java:159)
        at test.Ice.maxConnections.AllTests.testCreateConnections(AllTests.java:57)
        at test.Ice.maxConnections.AllTests.allTests(AllTests.java:35)
        at test.Ice.maxConnections.Client.run(Client.java:15)
        at com.zeroc.testcontroller.ControllerApp$ControllerHelperI.run(ControllerApp.java:328)
```

pepone:
Fix #3300

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
@@ -82,15 +82,19 @@ public final class Instance implements java.util.function.Function<String, Class
         private final ThreadObserverHelper _observerHelper;
     }
 
+    // The thead pool executor uses an unbounded queue. The tasks would wait in the queue until a
+    // core pool thread is available to run it and the value of the maximumPoolSize therefore
+    // doesn't have any effect.
     private static class QueueExecutor extends java.util.concurrent.ThreadPoolExecutor {
         QueueExecutor(Properties props, String threadName) {
             super(
-                    1,
-                    1,
-                    0,
+                    10,
+                    10,
+                    1000,
                     TimeUnit.MILLISECONDS,
                     new java.util.concurrent.LinkedBlockingQueue<Runnable>(),
                     Util.createThreadFactory(props, threadName));
+            allowCoreThreadTimeOut(true);
             _observerHelper = new ThreadObserverHelper(threadName);
         }
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/QueueExecutorService.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/QueueExecutorService.java
@@ -13,14 +13,6 @@ import java.util.concurrent.RejectedExecutionException;
 final class QueueExecutorService {
     QueueExecutorService(ExecutorService executor) {
         _executor = executor;
-        _thread =
-                executeNoThrow(
-                        new Callable<Thread>() {
-                            @Override
-                            public Thread call() {
-                                return Thread.currentThread();
-                            }
-                        });
     }
 
     public <T> T executeNoThrow(Callable<T> callable) {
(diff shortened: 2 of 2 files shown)


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
```diff
@@ -82,15 +82,19 @@ protected void afterExecute(Runnable t, Throwable e) {
...
+    // doesn't have any effect.
     private static class QueueExecutor extends java.util.concurrent.ThreadPoolExecutor {
         QueueExecutor(Properties props, String threadName) {
             super(
-                    1,
-                    1,
-                    0,
+                    10,
+                    10,
+                    1000,
```
externl, line 93: what do these magic numbers refer to?

pepone: I would add a comment, but see https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/ThreadPoolExecutor.html#ThreadPoolExecutor-int-int-long-java.util.concurrent.TimeUnit-java.util.concurrent.BlockingQueue-

externl: Ok, so the `1000` for `keepAliveTime` will have  no effect here since we're setting core and max to 10.

externl (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
```diff
@@ -82,15 +82,19 @@ protected void afterExecute(Runnable t, Throwable e) {
         private final ThreadObserverHelper _observerHelper;
     }
 
+    // The thead pool executor uses an unbounded queue. The tasks would wait in the queue until a
+    // core pool thread is available to run it and the value of the maximumPoolSize therefore
+    // doesn't have any effect.
```
bernardnormier, line 87: The comment is not clear.

Presumably, maximumPoolSize has not effect because the executor uses an unbounded queue? As written, it sounds like a consequence of the wait?

It's also not clear to me what this means / add:
> The tasks would wait in the queue until a core pool thread is available to run it 

This is not obvious?

bernardnormier (commented):
I am concerned the original QueueExecutorService has a single thread _by design_, to ensure the jobs are serialized. Hard to say without comment. If it was by design, this PR is changing the design.

pepone:
If you look at when we use this executor there is two situations:

- We want to allow interrupt threads
- We are running on Android and need to avoid IO calls from UI thread

https://github.com/zeroc-ice/ice/blob/e99d2e515a3db83b11808b8b0467ca55267ffdf7/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java#L920-L936

The interrupt case is about about ensuring that Java interrupt works correctly. See https://doc.zeroc.com/ice/latest/property-reference/miscellaneous-ice-properties#id-.MiscellaneousIce.*Propertiesv3.7-Ice.ThreadInterruptSafe

We hava Ice/interrupt test in Java with seems to work fine with this PR.

In the android case this is about not calling IO APIs from the main thread, and unrelated to serialization.

Issue #3205, opened 2024-11-27
`slice2cpp` Does Not Generate `catch` Blocks in a Stable Ordering


InsertCreativityHere:
After changing `slice2cpp` to use triple slash doc-comments instead of Javadoc style comments,
I noticed that the order of `catch` blocks that `slice2cpp` generates for an operation's exception specification can randomly swap around.

This isn't critical, but it'd be nice if `slice2cpp` would always output the exact same generated code when re-compiling the same Slice.

InsertCreativityHere:
This isn't limited to C++, it seems to affect most (if not all) of the compilers:

InsertCreativityHere:
This was fixed in #3315

PR #3315, opened 2025-01-07
Fix Instabilities and Unnecessary Sorting in the Generated Code

### LINKED ISSUE #3313: User-Defined Operations are Not Generated in a Stable Ordering with `slice2cs`
`slice2cs` normally generates user-defined functions in the order that you pass them, but sometimes the order of these operations will change. Interestingly, the order of certain operations remains unchanged, but the order of groups of operations will swap around.
See this for an example:
https://github.com/InsertCreativityHere/compiler-comparison/commit/367c05c1b8f1e4a22a1e46fdf5c7b18fa56b8cc9#diff-7c8e3101e439f1f9a8ef3843193f28c141dcea274095e762b972b856a5151928

... actually, as I'm writing this out, I'm pretty sure I figured it out. I think what's happening is when you have interface inheritance,
and we iterate through the parent interfaces, the order we do that in isn't stable... that seems like a sensible explanation at least.

InsertCreativityHere:
This PR fixes #3313 and #3205.

All of the instabilities were caused by places where we were calling `sort` on `list`s of `shared_ptr`s, which meant we were sorting things by unstable memory addresses.
Thankfully, all of the sorting we were trying to do in those places was unnecessary and removed.

The two types of sorting which is removed by this PR:
- We used to sort the exception specifications of operation, so it'd be safe to call `unique` to remove duplicates.
  But this is useless. The Slice compiler already disallows the same exception to appear twice in an exception specification.
- We used to sort out operations by name, so it'd be safe to call `unique` to remove duplicates.
  But again, the Slice compiler issues errors if there's any conflicts, even from inherited names (including diamonds).

One exception:
We still sort our operations by name in C++ because we use a binary search over operation names to perform dispatch.
So we require the names to be alphabetized. We re-use the same list for generating the actual operations and the dispatch, and I just left it alone. It's fine.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -3087,7 +3087,7 @@ Slice::InterfaceDef::allOperations() const
             if (find_if(
                     result.begin(),
                     result.end(),
-                    [scoped = q->scoped()](const auto& other) { return other->scoped() == scoped; }) == result.end())
+                    [name = q->name()](const auto& other) { return other->name() == name; }) == result.end())
             {
                 result.push_back(q);
             }
@@ -3099,7 +3099,7 @@ Slice::InterfaceDef::allOperations() const
         if (find_if(
                 result.begin(),
                 result.end(),
-                [scoped = q->scoped()](const auto& other) { return other->scoped() == scoped; }) == result.end())
+                [name = q->name()](const auto& other) { return other->name() == name; }) == result.end())
         {
             result.push_back(q);
         }
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -3583,7 +3583,6 @@ Slice::Exception::createDataMember(
     checkIdentifier(name); // Don't return here -- we create the data member anyway.
 
     // Check whether any bases have defined a member with the same name already.
-    ExceptionList bl = allBases();
     for (const auto& q : allBases())
     {
         ContainedList contents;
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -3053,7 +3053,6 @@ Slice::InterfaceDef::allBases() const
 {
     InterfaceList result = _bases;
     result.sort(containedCompare);
-    result.unique(containedEqual);
     for (const auto& p : _bases)
     {
         result.merge(p->allBases(), containedCompare);
--- a/cpp/src/ice2slice/Gen.cpp
+++ b/cpp/src/ice2slice/Gen.cpp
@@ -561,8 +561,6 @@ Gen::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
         }
 
         ExceptionList throws = op->throws();
-        throws.sort();
-        throws.unique();
         if (throws.size() == 1)
         {
             out << " throws " << getUnqualified(throws.front(), scope);
--- a/cpp/src/slice2cpp/Gen.cpp
+++ b/cpp/src/slice2cpp/Gen.cpp
@@ -2651,7 +2651,6 @@ Slice::Gen::InterfaceVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
         allOpNames.push_back("ice_isA");
         allOpNames.push_back("ice_ping");
         allOpNames.sort();
-        allOpNames.unique();
 
         H << sp;
         H << nl << "/// \\cond INTERNAL";
--- a/cpp/src/slice2cpp/Gen.cpp
+++ b/cpp/src/slice2cpp/Gen.cpp
@@ -216,15 +216,10 @@ namespace
         }
         else
         {
-            throws.sort();
-            throws.unique();
-
-            //
             // Arrange exceptions into most-derived to least-derived order. If we don't
             // do this, a base exception handler can appear before a derived exception
             // handler, causing compiler warnings and resulting in the base exception
             // being marshaled instead of the derived exception.
-            //
             throws.sort(Slice::DerivedToBaseCompare());
 
             C << "[](const " << getUnqualified("::Ice::UserException&", scope) << " ex)";
@@ -233,9 +228,8 @@ namespace
             C << sb;
             C << nl << "ex.ice_throw();";
             C << eb;
-            //
+
             // Generate a catch block for each legal user exception.
-            //
             for (const auto& ex : throws)
             {
                 C << nl << "catch(const " << getUnqualified(fixKwd(ex->scoped()), scope) << "&)";
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -325,9 +325,9 @@ Slice::CsVisitor::writeInheritedOperations(const InterfaceDefPtr& p)
             // It's possible to get the same operation name through diamond inheritance.
             // But we only want one 'copy' of each operation in our list, to avoid generating duplicate methods.
             if (find_if(
-                allBaseOps.begin(),
-                allBaseOps.end(),
-                [name = baseOp->name()](const auto& other) { return other->name() == name; }) == allBaseOps.end())
+                    allBaseOps.begin(),
+                    allBaseOps.end(),
+                    [name = baseOp->name()](const auto& other) { return other->name() == name; }) == allBaseOps.end())
             {
                 allBaseOps.push_back(baseOp);
             }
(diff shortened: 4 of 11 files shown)


cpp/src/Slice/Parser.cpp
```diff
@@ -3053,7 +3053,6 @@ Slice::InterfaceDef::allBases() const
 {
     InterfaceList result = _bases;
     result.sort(containedCompare);
-    result.unique(containedEqual);
```
InsertCreativityHere, line 3056: It is illegal to list the same interface multiple times in an inheritance list, so there's no point to calling `unique` on `_bases` here.

cpp/src/Slice/Parser.cpp
```diff
@@ -3584,7 +3583,6 @@ Slice::Exception::createDataMember(
     checkIdentifier(name); // Don't return here -- we create the data member anyway.
 
     // Check whether any bases have defined a member with the same name already.
-    ExceptionList bl = allBases();
```
InsertCreativityHere, line 3587: Dead variable.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -561,8 +561,6 @@ Gen::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
         }
 
         ExceptionList throws = op->throws();
-        throws.sort();
-        throws.unique();
```
InsertCreativityHere, line 565: We no longer sort our exception specifications, because there's no point to calling `unique`.
The Slice compiler already ensures this for us.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -216,15 +216,10 @@ namespace
...
-            throws.sort();
-            throws.unique();
-
-            //
             // Arrange exceptions into most-derived to least-derived order. If we don't
             // do this, a base exception handler can appear before a derived exception
             // handler, causing compiler warnings and resulting in the base exception
             // being marshaled instead of the derived exception.
-            //
             throws.sort(Slice::DerivedToBaseCompare());
```
InsertCreativityHere, line 223: Note that in the languages, we _do_ still sort them by derived-to-base.
This sorting is stable (it won't re-arrange elements that are 'equal'). And this is necessary to prevent dead catch blocks.

We do this in all compilers except 'ice2slice' (where it's correct not to) and 'slice2php', 'slice2py', and 'slice2rb'.
It is suspicious that we aren't sorting exceptions in those languages.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -317,25 +317,20 @@ Slice::CsVisitor::writeUnmarshalDataMember(
...
-        allOps.sort();
-        allOps.unique();
-        for (OperationList::const_iterator i = allOps.begin(); i != allOps.end(); ++i)
-        {
-            string retS;
-            vector<string> params, args;
-            string ns = getNamespace(p);
-            string name = getDispatchParams(*i, retS, params, args, ns);
-            _out << sp << nl << "public abstract " << retS << " " << name << spar << params << epar << ';';
-        }
```
InsertCreativityHere, line 338: Simplified this whole block.
But the main change was removing the sorting of `OperationPtr` that it used to be doing.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -2698,18 +2693,6 @@ Slice::Gen::HelperVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
...
-        throws.unique();
-
-        //
-        // Arrange exceptions into most-derived to least-derived order. If we don't
-        // do this, a base exception handler can appear before a derived exception
-        // handler, causing compiler warnings and resulting in the base exception
-        // being marshaled instead of the derived exception.
-        //
-        throws.sort(Slice::DerivedToBaseCompare());
-
```
InsertCreativityHere, line 2712: Dead code.

cpp/src/slice2java/Gen.cpp
```diff
@@ -3956,8 +3954,6 @@ Slice::Gen::ProxyVisitor::ProxyVisitor(const string& dir) : JavaVisitor(dir) {}
 bool
 Slice::Gen::ProxyVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
 {
-    const OperationList ops = p->allOperations();
```
InsertCreativityHere, line 3959: Another dead variable...

cpp/src/slice2js/Gen.cpp
```diff
@@ -2035,7 +2036,6 @@ Slice::Gen::TypeScriptImportVisitor::visitInterfaceDefStart(const InterfaceDefPt
     }
 
     // Add imports required for operation parameters and return types.
-    const OperationList operationList = p->allOperations();
```
InsertCreativityHere, line 2038: Dead.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -1713,7 +1713,6 @@ CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     const string scoped = p->scoped();
     const string abs = getAbsolute(p);
     InterfaceList bases = p->bases();
-    const OperationList allOps = p->allOperations();
```
InsertCreativityHere, line 1716: Dead.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -317,26 +317,31 @@ Slice::CsVisitor::writeUnmarshalDataMember(
 void
 Slice::CsVisitor::writeInheritedOperations(const InterfaceDefPtr& p)
 {
```
bernardnormier, line 319: I think it would be cleaner to eliminate this helper function (called once) and just call allOperations on the interface in the calling function.

bernardnormier (approved):
I woud also remove:
> allOpNames.unique();

from line 2564 in slice2cpp/Gen.cpp.

InsertCreativityHere:
Originally I left that `unique` in because I was planning to check out the code around it, because I smelled a bug.
But I went ahead and removed this `unique` in this PR and just opened an issue for my suspicion:
#3318

cpp/src/slice2cs/Gen.cpp
InsertCreativityHere: I didn't want to make it too hard to review the changes, but you're right.
I 'inlined' this code, and actually managed to simplify a good chunk of the code around where we called this too.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -2657,7 +2651,6 @@ Slice::Gen::InterfaceVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
         allOpNames.push_back("ice_isA");
         allOpNames.push_back("ice_ping");
         allOpNames.sort();
-        allOpNames.unique();
```
InsertCreativityHere, line 2660: The Slice compiler ensures that there are no duplicate operation names, even among inherited operations.

cpp/src/slice2cs/Gen.cpp
InsertCreativityHere: After a review comment from Bernard, I just inlined the simplified version of this function.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -3224,16 +3178,30 @@ Slice::Gen::DispatcherVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
...
+            if (find_if(
+                    allOps.begin(),
+                    allOps.end(),
+                    [name = baseOp->name()](const auto& other) { return other->name() == name; }) == allOps.end())
+            {
+                allOps.push_back(baseOp);
+            }
+        }
+    }
+    for (const auto& op : allOps)
```
InsertCreativityHere, line 3197: Where `writeInheritedOperations` was inlined to.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -3223,17 +3177,14 @@ Slice::Gen::DispatcherVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     _out << fixId(name);
 
     _out << sb;
-
-    for (const auto& op : p->operations())
+    for (const auto& op : p->allOperations())
```
InsertCreativityHere, line 3180: Now we just call `allOperations` instead of `operations` followed by `writeInheritedOperations`.

A side effect of this is that now, the order that we generate operations in is reversed. It used to be Derived -> Base, but now it's Derived -> Base.

pepone (approved):

Issue #3316, opened 2025-01-07
Remove superfluous strings from Java SerialVersionUID computation


externl:
When generating the Java `SerialVersionUID` we include some (I believe) superfluous strings that I don't think are adding much.

https://github.com/zeroc-ice/ice/blob/e99d2e515a3db83b11808b8b0467ca55267ffdf7/cpp/src/slice2java/JavaUtil.cpp#L124-L141

eg. `Name: `, `Base: [`, `Members: [`, etc. 

I think we should just remove all of these and just mash all of the strings together.  We could add some kind of separator like `;` to disambiguate something like: `class Foo : Bar` and ` class FooBar`.

InsertCreativityHere:
I think that's fine.
It's not like we ever print this thing or anything.
And I think a simple separator like ';' is sufficient to prevent collisions.

InsertCreativityHere:
This was fixed in #3317.

PR #3317, opened 2025-01-07
Fix Generation of `serialVersionUID` Default Values

### LINKED ISSUE #3312: `slice2java` Does not Generate Stable Values for `serialVersionUID` Field
Every time you run `slice2java` over a Slice file, you will get a different value for this field.
This is particularly bad, since this field is supposed to represent a stable version ID for the class.
So, having it change every time you compile the file makes it completely useless, and as a result,
it is impossible to safely serialize/deserialize generated classes in Java at the moment.

This was caused by our switch to use `shared_ptr`, so it only affects main, and not any other branches.

InsertCreativityHere:
This PR fixes #3312 and #3316, both of which deal with how we generate the `serialVersionUID` field in java.
All of the 'real' changes are at the top of the first file.

But, I had to switch alot of functions defined on `JavaGenerator` from being `const` methods to `static` functions.
Since the function for computing `serialVersionUID` lives outside this `JavaGenerator` class, this change was necessary for it to be able to call `typeToString`. And changing `typeToString` to `static` had some viral effects.

### DIFF
--- a/cpp/src/slice2java/JavaUtil.cpp
+++ b/cpp/src/slice2java/JavaUtil.cpp
@@ -131,7 +131,8 @@ Slice::computeDefaultSerialVersionUID(const ContainedPtr& p)
     os << ";";
     for (const auto& member : members)
     {
-        const string typeString = JavaGenerator::typeToString(member->type(), TypeModeMember, "", member->getMetadata());
+        const MetadataList metadata = member->getMetadata();
+        const string typeString = JavaGenerator::typeToString(member->type(), TypeModeMember, "", metadata);
         os << member->name() << ":" << typeString << ",";
     }
 
--- a/cpp/src/slice2java/JavaUtil.cpp
+++ b/cpp/src/slice2java/JavaUtil.cpp
@@ -123,22 +123,17 @@ Slice::computeDefaultSerialVersionUID(const ContainedPtr& p)
 
     // Actually compute the `SerialVersionUID` value.
     ostringstream os;
-    os << "Name: " << name;
+    os << name << ":";
     if (baseName)
     {
-        os << " Base: [" << *baseName << "]";
+        os << *baseName;
     }
-    os << " Members: [";
-    for (DataMemberList::const_iterator i = members.begin(); i != members.end();)
+    os << ";";
+    for (const auto& member : members)
     {
-        os << (*i)->name() << ":" << (*i)->type();
-        i++;
-        if (i != members.end())
-        {
-            os << ", ";
-        }
+        const string typeString = JavaGenerator::typeToString(member->type(), TypeModeMember, "", member->getMetadata());
+        os << member->name() << ":" << typeString << ",";
     }
-    os << "]";
 
     // We use a custom hash instead of relying on `std::hash` to ensure cross-platform consistency.
     const string data = os.str();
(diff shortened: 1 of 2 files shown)


cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -381,7 +376,7 @@ Slice::JavaGenerator::fixKwd(const string& name)
 }
```
InsertCreativityHere, line 377: Everything below this line is just me switching from `<function> const` to `static <function>`.
There is no reason for these not to be static, but it just happens that I _needed_ them to be static in this PR, so I could call them from outside the class they're defined in.

cpp/src/slice2java/JavaUtil.h
```diff
@@ -83,78 +92,67 @@ namespace Slice
...
         //
         // Return the method call necessary to obtain the static type ID for an object type.
         //
-        [[nodiscard]] std::string getStaticId(const TypePtr&, const std::string&) const;
+        [[nodiscard]] static std::string getStaticId(const TypePtr&, const std::string&);
 
         //
         // Returns the optional type corresponding to the given Slice type.
         //
-        std::string getOptionalFormat(const TypePtr&);
```
InsertCreativityHere, line 126: I added a `[[nodiscard]]` here. This function has no side-effects and returns a string.
So, it should probably have `[[nodiscard]]` right?

cpp/src/slice2java/JavaUtil.h
```diff
@@ -237,11 +233,10 @@ namespace Slice
         // The functions return true if a custom type was defined and false to indicate
         // the default mapping was used.
         //
-        bool
-        getDictionaryTypes(const DictionaryPtr&, const std::string&, const MetadataList&, std::string&, std::string&)
-            const;
-        bool
-        getSequenceTypes(const SequencePtr&, const std::string&, const MetadataList&, std::string&, std::string&) const;
+        static bool
```
externl, line 236: I'm fine making these static but they probably don't need to be one `JavaGenerator`.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -123,22 +123,18 @@ Slice::computeDefaultSerialVersionUID(const ContainedPtr& p)
...
     {
-        os << (*i)->name() << ":" << (*i)->type();
-        i++;
-        if (i != members.end())
-        {
-            os << ", ";
-        }
+        const MetadataList metadata = member->getMetadata();
+        const string typeString = JavaGenerator::typeToString(member->type(), TypeModeMember, "", metadata);
+        os << member->name() << ":" << typeString << ",";
```
externl, line 136: This will result in an "extra" comma, which is fine.

externl (approved):

bernardnormier (approved):

cpp/src/slice2java/JavaUtil.cpp
InsertCreativityHere: Yes, I know it's a little lame, but it's just getting fed into a hash function.
So I didn't want to waste time and clock cycles checking if it's the last member.
*shrugs*

cpp/src/slice2java/JavaUtil.h
InsertCreativityHere: I agree, but I didn't want to go too crazy in what was otherwise a very small and focused PR.

PR #3329, opened 2025-01-09
Reject redefinition of Object operations

### LINKED ISSUE #3318: We Should Blacklist `ice_id`, `ice_ids`, `ice_ping`, `ice_isa` as Allowable Operation Names
If a user names any of their operations with these names, the Slice compilers will generate uncompilable code.
Because these 4 operations are super special and always generated, so we get duplicate function definitions.

This didn't used to be a problem because we blacklisted all `ice_` prefixes for Slice identifiers.
We wisely removed this, but now that it's gone, we should at least block these specific names.

Since this problem affects all mappings, it should be done in the Slice compiler.

bernardnormier:
Fixes #3318.


cpp/src/Slice/Parser.cpp
```diff
@@ -3007,35 +3007,56 @@ Slice::InterfaceDef::createOperation(
...
+        if (baseName == name)
+        {
+            ostringstream os;
+            os << "operation `" << name << "' is already defined as an operation in a base interface";
+            _unit->error(os.str());
+            return false;
+        }
+
+        string baseName2 = IceInternal::toLower(baseName);
+        string newName2 = IceInternal::toLower(name);
```
InsertCreativityHere, line 3047: Can we move this out of the `for (baseName : baseNames)` loop?
So we only have to compute it once instead of N many times.

cpp/src/Slice/Parser.h
```diff
@@ -723,6 +723,9 @@ namespace Slice
     private:
         friend class Container;
 
+        // Returns true if name does not collide with any base name; otherwise, false.
+        bool checkBaseOperationNames(const std::string& name, const std::vector<std::string>& baseNames);
```
InsertCreativityHere, line 727: I know that we're using this to check for collisions with base names, but the implementation itself isn't specific to base names.

I think a name like `checkForNameCollision` would be more apt for this function.
Along with changing `baseNames` to `otherNames` or something.

InsertCreativityHere (approved):

externl (approved):

cpp/src/Slice/Parser.h
bernardnormier: The implementation is specific to base operation names: please read the error messages.

cpp/src/Slice/Parser.cpp
bernardnormier: Moved.

PR #3332, opened 2025-01-09
Add more [[nodiscard]]

bernardnormier:
This PR adds more `[[nodiscard]]` using clang-tidy and occasionally NOLINT:modernize-use-nodiscard.

### DIFF
--- a/cpp/include/Ice/Proxy.h
+++ b/cpp/include/Ice/Proxy.h
@@ -325,6 +325,7 @@ namespace Ice
          * @param context The context map for the invocation.
          * @return A function that can be called to cancel the invocation locally.
          */
+        // NOLINTNEXTLINE:modernize-use-nodiscard
         std::function<void()> ice_isAAsync(
             std::string_view typeId,
             std::function<void(bool)> response,
@@ -360,6 +361,7 @@ namespace Ice
          * @param context The context map for the invocation.
          * @return A function that can be called to cancel the invocation locally.
          */
+        // NOLINTNEXTLINE:modernize-use-nodiscard
         std::function<void()> ice_pingAsync(
             std::function<void()> response,
             std::function<void(std::exception_ptr)> ex = nullptr,
@@ -392,6 +394,7 @@ namespace Ice
          * @param context The context map for the invocation.
          * @return A function that can be called to cancel the invocation locally.
          */
+        // NOLINTNEXTLINE:modernize-use-nodiscard
         std::function<void()> ice_idsAsync(
             std::function<void(std::vector<std::string>)> response,
             std::function<void(std::exception_ptr)> ex = nullptr,
@@ -427,6 +430,7 @@ namespace Ice
          * @param context The context map for the invocation.
          * @return A function that can be called to cancel the invocation locally.
          */
+        // NOLINTNEXTLINE:modernize-use-nodiscard
         std::function<void()> ice_idAsync(
             std::function<void(std::string)> response,
             std::function<void(std::exception_ptr)> ex = nullptr,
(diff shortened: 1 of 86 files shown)


cpp/src/IceGrid/DescriptorBuilder.h
```diff
@@ -18,11 +18,10 @@ namespace IceGrid
         XmlAttributesHelper(const XMLAttributes&, const Ice::LoggerPtr&, const std::string&, int);
 
         void checkUnknownAttributes();
-        bool contains(const std::string&) const;
-        std::map<std::string, std::string> asMap() const;
-
-        bool asBool(const std::string&) const;
-        bool asBool(const std::string&, bool) const;
+        bool contains(const std::string&) const; // NOLINT:modernize-use-nodiscard
```
bernardnormier, line 21: All these functions update the mutable _used. In some code, we call these functions only to update _used. And yes, that's not very elegant.

cpp/include/Ice/Proxy.h
```diff
@@ -325,6 +325,7 @@ namespace Ice
          * @param context The context map for the invocation.
          * @return A function that can be called to cancel the invocation locally.
          */
+        // NOLINTNEXTLINE:modernize-use-nodiscard
```
bernardnormier, line 328: It's ok to ignore the "cancel" function returned by this function.

cpp/src/Ice/Instance.h
```diff
@@ -66,34 +66,34 @@ namespace IceInternal
...
+        [[nodiscard]] ReferenceFactoryPtr referenceFactory() const;
+        [[nodiscard]] OutgoingConnectionFactoryPtr outgoingConnectionFactory() const;
+        [[nodiscard]] ObjectAdapterFactoryPtr objectAdapterFactory() const;
+        [[nodiscard]] ProtocolSupport protocolSupport() const;
+        [[nodiscard]] bool preferIPv6() const;
+        [[nodiscard]] NetworkProxyPtr networkProxy() const;
         ThreadPoolPtr clientThreadPool();
         ThreadPoolPtr serverThreadPool();
         EndpointHostResolverPtr endpointHostResolver();
         RetryQueuePtr retryQueue();
```
pepone, line 84: Why was not added here?

cpp/src/Ice/Instance.h
```diff
@@ -102,22 +102,22 @@ namespace IceInternal
         Ice::ObjectPtr findAdminFacet(std::string_view);
         Ice::FacetMap findAllAdminFacets();
```
pepone, line 106: Not clear why is not added to this find functions.

cpp/src/Ice/Instance.h
```diff
@@ -102,22 +102,22 @@ namespace IceInternal
...
 
         void setLogger(const Ice::LoggerPtr&);
         void setThreadHook(std::function<void()>, std::function<void()>);
 
-        const Ice::StringConverterPtr& getStringConverter() const { return _stringConverter; }
-        const Ice::WstringConverterPtr& getWstringConverter() const { return _wstringConverter; }
+        [[nodiscard]] const Ice::StringConverterPtr& getStringConverter() const { return _stringConverter; }
+        [[nodiscard]] const Ice::WstringConverterPtr& getWstringConverter() const { return _wstringConverter; }
 
         BufSizeWarnInfo getBufSizeWarn(std::int16_t type);
```
pepone, line 119: Here too

pepone (approved):

cpp/src/Ice/Instance.h
bernardnormier: These functions are not const.

cpp/src/Ice/Instance.h
bernardnormier: Same here: non-const.

cpp/src/Ice/Instance.h
bernardnormier: Same.

PR #3310, opened 2025-01-03
Fixes for session establishment through node forwarder

pepone:
- The confirmCreateSession request should not be forward to the peer, if there isn't a NodeSession associated with the caller's connection. Otherwise a new NodeSession would be created for the publisher, and this new NodeSession would be unaware of the subscriber session.

- The call to disconnected from a relay should not go through the session forwarder created by the relay. Using a forwarder can alter the order in which the peer dispatched the requests if there is a mix of request that go through a forwarder and direct requests. The ones that go through the forwarder have to way for the forwarder to dispatch them (collocated call) and them forward to the target session/node.

Fix #3309

PR #3333, opened 2025-01-10
DataStorm: Fixes for Session Establishment and Reconnection

### LINKED ISSUE #3309: DataStorm/reliability test hang
The tests hangs, I got the same issue a mac Book pro, and Debian.

The test hanging is ` running reader/writer as client with 2 nodes (reversed start order) test ]`

I got this testing with fixes from #3294. 

Reader:

```
Thread 1 (Thread 0x7f988966f100 (LWP 82271) "reader"):
#0  0x00007f98888a4f16 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f98888a75d8 in pthread_cond_wait () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f9889504bbd in DataStormI::DataElementI::waitForListeners (this=0x561885f8cdc0, count=1) at src/DataStorm/DataElementI.cpp:559
#3  0x00007f988950a716 in DataStormI::KeyDataWriterI::waitForReaders (this=0x561885f8cdc0, count=1) at src/DataStorm/DataElementI.cpp:1194
#4  0x0000561885abf811 in DataStorm::Writer<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::waitForReaders (this=0x7ffd50e1f680, count=1) at ../cpp/include/DataStorm/DataStorm.h:1616
#5  0x0000561885aba562 in Reader::run (this=0x7ffd50e1fec0, argc=1, argv=0x7ffd50e20078) at test/DataStorm/reliability/Reader.cpp:67
#6  0x0000561885abf989 in Test::runTest<Reader> (argc=17, argv=0x7ffd50e20078) at test/include/TestHelper.h:152
#7  0x0000561885abb1ca in main (argc=17, argv=0x7ffd50e20078) at test/DataStorm/reliability/Reader.cpp:129
```

Writer:

```
Thread 1 (Thread 0x7f3d340a2100 (LWP 82269) "writer"):
#0  0x00007f3d344a4f16 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f3d344a75d8 in pthread_cond_wait () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f3d3510cedb in std::condition_variable::wait<DataStormI::DataReaderI::getNextUnread()::<lambda()> >(std::unique_lock<std::mutex> &, struct {...}) (this=0x558c05c77120, __lock=..., __p=...) at /usr/include/c++/12/condition_variable:102
#3  0x00007f3d35106075 in DataStormI::DataReaderI::getNextUnread (this=0x558c05c36620) at src/DataStorm/DataElementI.cpp:728
#4  0x0000558c04d212ac in DataStorm::Reader<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >::getNextUnread (this=0x7ffceb1a8720) at ../cpp/include/DataStorm/DataStorm.h:1396
#5  0x0000558c04d1c387 in Writer::run (this=0x7ffceb1a89e0, argc=1, argv=0x7ffceb1a8b98) at test/DataStorm/reliability/Writer.cpp:54
#6  0x0000558c04d213a3 in Test::runTest<Writer> (argc=14, argv=0x7ffceb1a8b98) at test/include/TestHelper.h:152
#7  0x0000558c04d1cc8e in main (argc=14, argv=0x7ffceb1a8b98) at test/DataStorm/reliability/Writer.cpp:96
```
[client-122624-1336.log](https://github.com/user-attachments/files/18252823/client-122624-1336.log)
[node1-122624-1336.log](https://github.com/user-attachments/files/18252824/node1-122624-1336.log)
[node2-122624-1336.log](https://github.com/user-attachments/files/18252825/node2-122624-1336.log)
[server-122624-1336.log](https://github.com/user-attachments/files/18252826/server-122624-1336.log)

pepone:
This commit addresses two issues in the session establishment and reconnection process:

- **Session Reconnection Failure**

A bug prevented proper recovery from connection failures. Specifically, when a publisher session sent the `confirmCreateSession` request via a forwarder using a new connection (different from the one used to receive the `createSession` request), the `NodeForwarder` dispatching the request was unaware of the subscriber session. If the publisher’s connection to the forwarder was closed before the subscriber sent any additional messages, the forwarder could not notify the subscriber of the publisher’s disconnection.

As a result, if the publisher attempted to reconnect, the subscriber would ignore the request, thinking the session was still active.

- **Incorrect Use of Forwarding Proxies for Disconnect Requests**

Disconnect requests from the forwarder were sent using a forwarding proxy, which could cause unexpected behavior, such as requests being dispatched in the wrong order by the target node. Forwarding proxies are meant to be used by nodes connected to the forwarder, not by the forwarder itself.

This caused issues where `confirmCreateSession` was sent before `disconnected`, even if the forwarder invoked them in the correct reverse order. This happened because `disconnected` went through the forwarder, while `confirmCreateSession` was sent directly.

Replaces #3310 fixes #3309

### DIFF
--- a/cpp/src/DataStorm/Contract.ice
+++ b/cpp/src/DataStorm/Contract.ice
@@ -403,14 +403,28 @@ module DataStormContract
         /// @see Lookup::announceTopicReader
         void initiateCreateSession(Node* publisher);
 
-        /// Initiate the creation of a subscriber session with a node, after the target node has announced a topic
-        /// writer for which this node has a corresponding topic reader, or after the node has called
-        /// Node::initiateCreateSession.
+        /// Initiates the creation of a subscriber session with a node. The subscriber node sends this request to a
+        /// publisher node in one of the following scenarios:
         ///
-        /// @param subscriber The subscriber node initiating the session. The proxy is never null.
-        /// @param session The subscriber session being created. The proxy is never null.
-        /// @param fromRelay Indicates if the session is being created from a relay node.
-        void createSession(Node* subscriber, SubscriberSession* session, bool fromRelay);
+        /// - The subscriber has received a topic writer announcement from the publisher and has a matching topic
+        /// reader.
+        /// - The publisher node has previously send a initiateCreateSession request.
+        ///
+        /// The publisher node dispatching this request would send a confirmCreateSession request to the subscriber node
+        /// to continue session establishment. If an active session already exists with the subscriber node, the
+        /// request is ignored.
+        ///
+        /// @param subscriber The subscriber node initiating the session. This proxy is never null.
+        /// @param session The subscriber session being created. This proxy is never null.
+        /// @param fromRelay Indicates whether the session is being created from a relay node.
+        /// @param subscriberIsHostedOnRelay Specifies if the relay is hosting a forwarder for the subscriber. If the
+        /// subscriber has endpoints or an adapter ID, the relay does not host a forwarder, and the publisher is
+        /// expected to send requests directly to the subscriber node instead of going through a forwarder.
+        void createSession(
+            Node* subscriber,
+            SubscriberSession* session,
+            bool fromRelay,
+            optional(1) bool subscriberIsHostedOnRelay);
 
         /// Confirm the creation of a publisher session with a node.
         ///
--- a/cpp/src/DataStorm/Contract.ice
+++ b/cpp/src/DataStorm/Contract.ice
@@ -410,21 +410,17 @@ module DataStormContract
         /// reader.
         /// - The publisher node has previously sent a initiateCreateSession request.
         ///
-        /// The publisher node dispatching this request would send a confirmCreateSession request to the subscriber node
+        /// The publisher node dispatching this request then sends a confirmCreateSession request to the subscriber node
         /// to continue session establishment. If an active session already exists with the subscriber node, the
         /// request is ignored.
         ///
         /// @param subscriber The subscriber node initiating the session. This proxy is never null.
         /// @param session The subscriber session being created. This proxy is never null.
         /// @param fromRelay Indicates whether the session is being created from a relay node.
-        /// @param subscriberIsHostedOnRelay Specifies if the relay is hosting a forwarder for the subscriber. If the
-        /// subscriber has endpoints or an adapter ID, the relay does not host a forwarder, and the publisher is
-        /// expected to send requests directly to the subscriber node instead of going through a forwarder.
         void createSession(
             Node* subscriber,
             SubscriberSession* session,
-            bool fromRelay,
-            optional(1) bool subscriberIsHostedOnRelay);
+            bool fromRelay);
 
         /// Confirm the creation of a publisher session with a node.
         ///
--- a/cpp/src/DataStorm/Contract.ice
+++ b/cpp/src/DataStorm/Contract.ice
@@ -417,10 +417,7 @@ module DataStormContract
         /// @param subscriber The subscriber node initiating the session. This proxy is never null.
         /// @param session The subscriber session being created. This proxy is never null.
         /// @param fromRelay Indicates whether the session is being created from a relay node.
-        void createSession(
-            Node* subscriber,
-            SubscriberSession* session,
-            bool fromRelay);
+        void createSession(Node* subscriber, SubscriberSession* session, bool fromRelay);
 
         /// Confirm the creation of a publisher session with a node.
         ///
--- a/cpp/src/DataStorm/Contract.ice
+++ b/cpp/src/DataStorm/Contract.ice
@@ -408,7 +408,7 @@ module DataStormContract
         ///
         /// - The subscriber has received a topic writer announcement from the publisher and has a matching topic
         /// reader.
-        /// - The publisher node has previously send a initiateCreateSession request.
+        /// - The publisher node has previously sent a initiateCreateSession request.
         ///
         /// The publisher node dispatching this request would send a confirmCreateSession request to the subscriber node
         /// to continue session establishment. If an active session already exists with the subscriber node, the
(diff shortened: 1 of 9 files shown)


cpp/src/DataStorm/Contract.ice
```diff
@@ -403,14 +403,28 @@ module DataStormContract
...
-        /// @param subscriber The subscriber node initiating the session. The proxy is never null.
-        /// @param session The subscriber session being created. The proxy is never null.
-        /// @param fromRelay Indicates if the session is being created from a relay node.
-        void createSession(Node* subscriber, SubscriberSession* session, bool fromRelay);
+        /// Initiates the creation of a subscriber session with a node. The subscriber node sends this request to a
+        /// publisher node in one of the following scenarios:
+        ///
+        /// - The subscriber has received a topic writer announcement from the publisher and has a matching topic
+        /// reader.
+        /// - The publisher node has previously send a initiateCreateSession request.
```
externl, line 411:
```suggestion
        /// - The publisher node has previously sent a initiateCreateSession request.
```

cpp/src/DataStorm/Contract.ice
```diff
@@ -403,14 +403,28 @@ module DataStormContract
...
+        ///
+        /// The publisher node dispatching this request would send a confirmCreateSession request to the subscriber node
+        /// to continue session establishment. If an active session already exists with the subscriber node, the
+        /// request is ignored.
+        ///
+        /// @param subscriber The subscriber node initiating the session. This proxy is never null.
+        /// @param session The subscriber session being created. This proxy is never null.
+        /// @param fromRelay Indicates whether the session is being created from a relay node.
+        /// @param subscriberIsHostedOnRelay Specifies if the relay is hosting a forwarder for the subscriber. If the
+        /// subscriber has endpoints or an adapter ID, the relay does not host a forwarder, and the publisher is
```
externl, line 421: Is this really saying "is not a well known proxy"? Can we just look at the proxy to figure this out?

cpp/src/DataStorm/NodeSessionManager.cpp
```diff
@@ -86,14 +86,16 @@ NodeSessionManager::init()
...
-NodeSessionManager::createOrGet(NodePrx node, const ConnectionPtr& connection, bool forwardAnnouncements)
+NodeSessionManager::createOrGet(NodePrx node, const ConnectionPtr& newConnection, bool forwardAnnouncements)
 {
     unique_lock<mutex> lock(_mutex);
 
     auto p = _sessions.find(node->ice_getIdentity());
     if (p != _sessions.end())
     {
-        if (p->second->getConnection() != connection)
+        // If called with a newConnection we destroy the node session before creating a new one that uses the new
```
externl, line 96:
```suggestion
        // If called with a new connection we destroy the node session before creating a new one that uses the new
```

cpp/src/DataStorm/NodeSessionManager.cpp
```diff
@@ -444,11 +446,13 @@ NodeSessionManager::disconnected(const LookupPrx& lookup)
 }
 
 void
-NodeSessionManager::destroySession(const NodePrx& node)
+NodeSessionManager::destroySession(const ConnectionPtr& connection, const NodePrx& node)
 {
     unique_lock<mutex> lock(_mutex);
+    // If the session is still using the connection destroy it, otherwise the node has already
+    // replace its NodeSession and it is using a new connection.
```
externl, line 453:
```suggestion
    // Destroy the connection if the session is still using it, otherwise the node has already
    // replaced its NodeSession and it is using a new connection.
```

cpp/src/DataStorm/Contract.ice
pepone: > "Can we just look at the proxy to figure this out?"

No, the forwarder replaces the well-known proxy with a proxy that has the forwarder endpoints. And the forwarder will forward request to it over its connection.

pepone: What about something like this

subscriberIsHostedOnRelay true when fromRelay is true and the proxy passed to the relay is a well-known
proxy. In this case the relay replaces the well-known proxy with a direct proxy to the subscriber forwarder hosted by the relay.

pepone: And maybe rename it to subscriberIsWellKnown too.

bernardnormier: This operation doc-comment talks about subscriber node, publisher node, relay and then about 'forwarder' in the doc-comment for this parameter. It's not clear what this forwarder is.

> /// @param fromRelay Indicates whether the session is being created from a relay node.

Meaning, it's not a subscriber node that sends this request directly to the publisher node, but it's going through an intermediary node - a relay node?

> subscriberIsHostedOnRelay true when fromRelay is true and the proxy passed to the relay is a well-known
proxy.
> 

Not clear which proxy you're talking about here - the subscriber proxy?

> In this case the relay replaces the well-known proxy with a direct proxy to the subscriber forwarder hosted by the relay.

So if the subscriber proxy is well-known ... but then why do we need this parameter? The servant can figure out if the subscriber proxy it receives is well-known.

pepone: I removed this parameter, the servant can figure this from the session category. "sf" for subscriber forwarders, and "s" for subscribers.

bernardnormier (approved):

externl (approved):
Looks good to me!

PR #3334, opened 2025-01-10
Remove Ice.ThreadInterruptSafe

### LINKED ISSUE #3330: Remove Ice.ThreadInterruptSafe
I propose to remove this opt-in path in the Java code.

If you want to make a long invocation and interrupt the thread waiting for the response "safely" (i.e. without killing the connection if it's making IOs), just use an async invocation.

pepone:
Fixes #3330


externl (approved):

java/test/src/main/java/test/Ice/interrupt/AllTests.java
```diff
@@ -120,25 +108,6 @@ public static void allTests(test.TestHelper helper) throws InterruptedException
...
-            try {
-                test(!mainThread.isInterrupted());
-                p.sleep(2000);
-                test(false);
-            } catch (com.zeroc.Ice.OperationInterruptedException ex) {
-                // Expected
-            } catch (test.Ice.interrupt.Test.InterruptedException e) {
-                test(false);
-            }
-
```
bernardnormier, line 141: We should revisit this test in a follow-up PR. For example, the code below seems to test Java not Ice, which makes no sense.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectionI.java
```diff
@@ -130,20 +129,10 @@ public void abort() {
...
-                            new Callable<Void>() {
-                                @Override
-                                public Void call() throws LocalException {
-                                    abortImpl();
-                                    return null;
-                                }
-                            });
-        } else {
-            abortImpl();
+        synchronized (this) {
```
bernardnormier, line 132: I would remove the check above. abort() is a non-blocking operation.

bernardnormier (approved):
Looks good, much simpler.

For a follow-up PR: we should check that we want to keep `OperationInterruptedException` as our unchecked translation for InterruptedException.

PR #3338, opened 2025-01-10
Fix modernize-use-equals-default clang-tidy lints

bernardnormier:
These fixes were mostly done "by hand".


cpp/src/Ice/DynamicLibrary.cpp
```diff
@@ -19,24 +19,22 @@ using namespace std;
 
 IceInternal::DynamicLibrary::DynamicLibrary() : _hnd(nullptr) {}
 
-IceInternal::DynamicLibrary::~DynamicLibrary()
+IceInternal::DynamicLibrary::~DynamicLibrary() = default;
+/*
```
externl, line 23: We should get rid of this code block and move the comment above onto the function

externl (approved):

pepone (approved):

PR #3345, opened 2025-01-13
Some C++ header cleanup

externl:
This PR removes unused headers from some source files in cpp/src/Ice (not all of them). I used the [Include What You Use](https://include-what-you-use.org) tool.

```
iwyu_tool.py -p compile_commands.json src/Ice
```

There's still a lot more to do.


cpp/include/Ice/Config.h
```diff
@@ -5,6 +5,8 @@
 #ifndef ICE_CONFIG_H
 #define ICE_CONFIG_H
 
+// IWYU pragma: keep
```
bernardnormier, line 8: what is this about?

cpp/src/Slice/FileTracker.h
```diff
@@ -4,7 +4,8 @@
 #define FILE_TRACKER_H
 
 #include "Ice/LocalException.h"
-#include "Parser.h"
+#include <list>
```
bernardnormier, line 7: We prefer a blank line between the `""` includes and the `<>` includes.

cpp/include/Ice/Config.h
externl: I'm testing with the `include-what-you-keep` tool to find unnecessary includes.

pepone (approved):

PR #3352, opened 2025-01-13
Throw OperationInterruptedException only when catching an InterruptedException (Java)

bernardnormier:
This PR updates the Java code to remove this pattern:
```java
if (Thread.interrupted()) {
    throw new OperationInterruptedException();
}
```

This pattern is not correct. It would comparable to adding:
```cs
cancellationToken.ThrowIfCancellationRequested();
```
at the beginning of all C# methods that accept a cancellation token.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/EndpointHostResolver.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/EndpointHostResolver.java
@@ -117,14 +117,15 @@ class EndpointHostResolver {
     }
 
     synchronized void destroy() {
-        assert (!_destroyed);
-        _destroyed = true;
+        if (!_destroyed) {
+            _destroyed = true;
 
-        //
-        // Shutdown the executor. No new tasks will be accepted.
-        // Existing tasks will execute.
-        //
-        _executor.shutdown();
+            //
+            // Shutdown the executor. No new tasks will be accepted.
+            // Existing tasks will execute.
+            //
+            _executor.shutdown();
+        }
     }
 
     void joinWithThread() throws InterruptedException {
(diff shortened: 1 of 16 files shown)


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/EndpointHostResolver.java
```diff
@@ -117,14 +117,15 @@ synchronized void resolve(
     }
 
     synchronized void destroy() {
-        assert (!_destroyed);
```
bernardnormier, line 120: The interrupt test was hitting this assert when calling destroy after an interrupted communicator destroy.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/OperationInterruptedException.java
```diff
@@ -4,10 +4,6 @@
 
 /** This exception indicates a request was interrupted. */
 public final class OperationInterruptedException extends LocalException {
-    public OperationInterruptedException() {
```
bernardnormier, line 7: Removed this constructor to make sure we always provide the cause (= InterruptedException).

java/test/src/main/java/test/Ice/interrupt/AllTests.java
```diff
@@ -134,7 +134,7 @@ public static void allTests(test.TestHelper helper) throws InterruptedException
                     Thread.currentThread().interrupt();
                     InvocationFuture<Void> f = Util.getInvocationFuture(r);
                     f.waitForSent();
-                    test(false);
+                    test(Thread.interrupted());
```
bernardnormier, line 137: If sent synchronously, we don't get any exception.

pepone (approved):

externl (approved):

InsertCreativityHere (approved):
Looks good to me

Issue #3337, opened 2025-01-10
Bogus OA setup in DataStorm


pepone:
I think there is a chance here to receive a request before the `connected` callback is called and setups the OA.

https://github.com/zeroc-ice/ice/blob/8857b67804be92b40ad1f2af05e2a0907c7fba50/cpp/src/DataStorm/NodeSessionManager.cpp#L340-L354

In Node implementation we use a different pattern with first calling getConnectionAsync and setup the OA before sending the request.

https://github.com/zeroc-ice/ice/blob/8857b67804be92b40ad1f2af05e2a0907c7fba50/cpp/src/DataStorm/NodeI.cpp#L179-L199

Maybe we can simplify this by using the new `setDefaultObjectAdapter`.

pepone:
> Maybe we can simplify this by using the new setDefaultObjectAdapter.

I think we cannot because DataStorm doesn't always owns the communicator, and this could interfere with applications that pass they own communicator to DataStorm and also want to setup its own default OA.

pepone:
Fixed in #3355

PR #3355, opened 2025-01-14
Upgrade DataStorm to use a Default OA for OutgoingConnections

pepone:
This PR upgrades DataStorm implementation to use setDefaultObjectAdapter to set a default object adapter for outgoing connections. See #3337

### DIFF
--- a/cpp/include/Ice/Communicator.h
+++ b/cpp/include/Ice/Communicator.h
@@ -210,7 +210,7 @@ namespace Ice
          * @return The object adapter associated by default with new outgoing connections.
          * @see Connection::getAdapter
          */
-        [[nodiscard]] ObjectAdapterPtr getDefaultObjectAdapter() const noexcept;
+        [[nodiscard]] ObjectAdapterPtr getDefaultObjectAdapter() const;
 
         /**
          * Sets the object adapter that will be associated with new outgoing connections created by this
@@ -218,7 +218,7 @@ namespace Ice
          * @param adapter The object adapter to associate with new outgoing connections.
          * @see Connection::setAdapter
          */
-        void setDefaultObjectAdapter(ObjectAdapterPtr adapter) noexcept;
+        void setDefaultObjectAdapter(ObjectAdapterPtr adapter);
 
         /**
          * Get the implicit context associated with this communicator.
--- a/cpp/src/DataStorm/Instance.cpp
+++ b/cpp/src/DataStorm/Instance.cpp
@@ -19,6 +19,12 @@ using namespace Ice;
 Instance::Instance(CommunicatorPtr communicator, function<void(function<void()> call)> customExecutor)
     : _communicator(std::move(communicator))
 {
+    if (_communicator->getDefaultObjectAdapter() != nullptr)
+    {
+        throw invalid_argument(
+            "communicator used to initialize a DataStorm node must not have a default object adapter");
+    }
+
     PropertiesPtr properties = _communicator->getProperties();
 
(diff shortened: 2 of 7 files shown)


cpp/include/Ice/Communicator.h
```diff
@@ -210,15 +210,15 @@ namespace Ice
          * @return The object adapter associated by default with new outgoing connections.
          * @see Connection::getAdapter
          */
-        [[nodiscard]] ObjectAdapterPtr getDefaultObjectAdapter() const noexcept;
+        [[nodiscard]] ObjectAdapterPtr getDefaultObjectAdapter() const;
```
pepone, line 213: Removed because it can throw CommunicatorDestroyedExcetion

cpp/src/DataStorm/Instance.cpp
```diff
@@ -19,6 +19,12 @@ using namespace Ice;
 Instance::Instance(CommunicatorPtr communicator, function<void(function<void()> call)> customExecutor)
     : _communicator(std::move(communicator))
 {
+    if (_communicator->getDefaultObjectAdapter() != nullptr)
```
pepone, line 22: This wont catch some cases where the application set a default OA before starting the Node. For those we need to relay in the documentation.

cpp/test/DataStorm/api/Writer.cpp
```diff
@@ -36,8 +36,8 @@ void ::Writer::run(int argc, char* argv[])
             // Communicators shared with DataStorm must have a property set that can use the "DataStorm" opt-in prefix.
             Ice::InitializationData initData;
             initData.properties = make_shared<Ice::Properties>(vector<string>{"DataStorm"});
-            Node n2(Ice::initialize(initData));
-            n2.getCommunicator()->destroy();
+            Ice::CommunicatorHolder communicatorHolder(Ice::initialize(initData));
+            Node n2(communicatorHolder.communicator());
         }
```
pepone, line 41: The previous code would destroy the communicator before destroying the node, which now results in CommunicatorDestroyerException thrown from DataStormI::Instance::destroy.

cpp/test/DataStorm/config/Writer.cpp
```diff
@@ -308,10 +308,18 @@ void ::Writer::run(int argc, char* argv[])
...
+        Node node;
+        Topic<string, int> topic;
+        SingleKeyWriter<string, int> writer;
+    };
+
     cout << "testing send time discard policy... " << flush;
     {
         writers.update(false); // Not ready
-        vector<tuple<Ice::CommunicatorHolder, Node, Topic<string, int>, SingleKeyWriter<string, int>>> w;
+        vector<WriterHolder> w;
```
pepone, line 322: This is to avoid the undefined destruction order of tuples, we don't want the communicator to be destroyed before than the node.

cpp/test/DataStorm/api/Writer.cpp
```diff
@@ -36,8 +36,8 @@ void ::Writer::run(int argc, char* argv[])
             // Communicators shared with DataStorm must have a property set that can use the "DataStorm" opt-in prefix.
             Ice::InitializationData initData;
             initData.properties = make_shared<Ice::Properties>(vector<string>{"DataStorm"});
-            Node n2(Ice::initialize(initData));
-            n2.getCommunicator()->destroy();
+            Ice::CommunicatorHolder communicatorHolder(Ice::initialize(initData));
```
bernardnormier, line 39: Use {} for ctors.

bernardnormier (approved):

externl (approved):

Issue #2442, opened 2024-07-10
Ruby Ice/exceptions test failure


bernardnormier:
I am getting this crash in CI (macos release), but I can't reproduce it on my desktop.

```
[ running client/server with sliced format test - 07/09/24 20:32:03 ]
- Config: tcp
(/Users/runner/work/ice/ice/cpp/test/Ice/exceptions/build/macosx/shared/server --Ice.Default.Host=127.0.0.1 --Test.BasePort=14000 --Ice.Warn.Connections=1 --Ice.Default.Protocol=tcp --Ice.IPv6=0 --Ice.PrintStackTraces=1 --Ice.Default.SlicedFormat=1 --Ice.ThreadPool.Server.Size=1 --Ice.ThreadPool.Server.SizeMax=3 --Ice.ThreadPool.Server.SizeWarn=0 --Ice.PrintAdapterReady=1)
(ruby  /Users/runner/work/ice/ice/ruby/test/TestHelper.rb Client --Ice.Default.Host=127.0.0.1 --Test.BasePort=14000 --Ice.Warn.Connections=1 --Ice.Default.Protocol=tcp --Ice.IPv6=0 --Ice.Default.SlicedFormat=1 env={'DYLD_LIBRARY_PATH': '/Users/runner/work/ice/ice/cpp/lib', 'RUBYLIB': '/Users/runner/work/ice/ice/ruby/ruby:/Users/runner/work/ice/ice/ruby/test/Ice/exceptions'})
testing value factory registration exception... <OBJ_INFO:gc_mark_ptr@gc.c:7072> 0x0000000101440028 [0 M    ] T_NONE 
/Users/runner/work/ice/ice/ruby/test/Ice/exceptions/AllTests.rb:27: [BUG] try to mark T_NONE object
ruby 3.3.3 (2024-06-12 revision f1c7b6f435) [arm64-darwin23]

-- Crash Report log information --------------------------------------------
   See Crash Report log file in one of the following locations:
     * ~/Library/Logs/DiagnosticReports
     * /Library/Logs/DiagnosticReports
   for more details.
Don't forget to include the above Crash Report log file in bug reports.

-- Control frame information -----------------------------------------------
c:0007 p:0056 s:0050 e:000046 METHOD /Users/runner/work/ice/ice/ruby/test/Ice/exceptions/AllTests.rb:27
[...]

PR #3359, opened 2025-01-14
Swap the Order that the Parser Creates 'Def' and 'Decl' Types

InsertCreativityHere:
This PR is a long overdue follow up to my PRs which centralized the metadata validation.

There are two types which can be forward declared in Slice: `Interface`s and `Class`es.
There are separate types for definitions (`InterfaceDef` and `ClassDef`) vs forward declarations (`InterfaceDecl`, `ClassDecl`).
And, whenever the compiler creates a `Def` type, it also implicitly creates a `Decl` type to go with it.

But, the fact that the compiler creates `Def` first and then `Decl` after causes us to emit some errors/warnings in a weird order.
This was only really noticed after implementing the metadata changes. I tried fixing it in that PR, but that caused an obscure ruby bug (#2442), so it was split off into a separate PR (this one).

----

Now, the compiler will create the `Decl` types before the `Def` types.
This fixes the order of diagnostic messages, but it also lets us simplify some code in the compilers as well,
since whenever we look up a type by name, we are guaranteed to hit `Decl`s before we hit `Def`s.


InsertCreativityHere:
This no longer seems to trigger the Ice/exception Ruby bug anymore.
It's not clear what changed.

cpp/src/Slice/Parser.cpp
```diff
@@ -1263,21 +1257,12 @@ Slice::Container::createClassDecl(const string& name)
...
     }
 
     ClassDeclPtr decl = make_shared<ClassDecl>(shared_from_this(), name);
     _unit->addContent(decl);
     _contents.push_back(decl);
-
-    if (def)
-    {
-        decl->_definition = def;
-    }
```
InsertCreativityHere, line 1279: ... And a result, `def` is always unset.

cpp/src/Slice/Parser.cpp
```diff
@@ -1320,10 +1305,8 @@ Slice::Container::createInterfaceDef(const string& name, const InterfaceList& ba
         }
         else
         {
```
InsertCreativityHere, line 1307: Same idea as the code above, except for `InterfaceDecl` now.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1057,9 +1057,7 @@ Slice::Gen::ForwardDeclVisitor::visitModuleEnd(const ModulePtr&)
 void
 Slice::Gen::ForwardDeclVisitor::visitClassDecl(const ClassDeclPtr& p)
 {
-    ClassDefPtr def = p->definition();
```
InsertCreativityHere, line 1060: More dead variables.
There's got to be some linter/tool for removing these things right?

cpp/test/Slice/errorDetection/WarningInvalidMetadata.err
```diff
@@ -22,13 +22,13 @@ WarningInvalidMetadata.ice:31: warning: the 'cpp:no-default-include' metadata do
 WarningInvalidMetadata.ice:34: warning: 'amd' metadata cannot be specified as file metadata
 WarningInvalidMetadata.ice:40: warning: the 'cpp:header-ext' metadata only accepts one argument but a list was provided
 WarningInvalidMetadata.ice:40: warning: 'cpp:header-ext' metadata cannot be applied to sequences
+WarningInvalidMetadata.ice:44: warning: ignoring duplicate metadata: 'deprecated:do not use this' does not match previously applied metadata of 'deprecated'
```
InsertCreativityHere, line 25: And finally, the original reason for doing this:
now all diagnostics are in line-number ordering.

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -370,17 +370,6 @@ Slice::Ruby::CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
     _out.dec();
     _out << nl << "end"; // End of class.
 
-    //
```
InsertCreativityHere, line 373: All the exact same code is executed by `visitClassDecl` right above this.

And now that we're guaranteed to run `visitClassDecl` before `visitClassDef` (the function you're looking at right now) this code is redundant.

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -502,18 +491,6 @@ Slice::Ruby::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     _out.dec();
     _out << nl << "end"; // End of proxy class.
 
-    //
```
InsertCreativityHere, line 505: Same but for `InterfaceDecl`.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -639,13 +638,6 @@ Slice::Python::CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
 
     _out.dec();
 
-    if (_classHistory.count(scoped) == 0 && p->canBeCyclic())
-    {
-        //
-        // Emit a forward declaration for the class in case a data member refers to this type.
-        //
-        _out << sp << nl << "_M_" << type << " = IcePy.declareValue('" << scoped << "')";
-    }
```
InsertCreativityHere, line 648: Impossible to hit.

`visitClassDecl` also has this logic, and at the end, adds `scoped` to `_classHistory`.
Now that we're guaranteed to call `visitClassDecl` before `visitClassDef`,
`_classHistory.count(scoped) == 0` will always be false.

`visitInterfaceDecl` also has this logic. And since we're guaranteed to call `visitInterfaceDecl` before a call to `visitInterfaceDef`, `_classHistory` will always contain this element by this point.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -551,7 +551,6 @@ Slice::Python::CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
 {
     string scoped = p->scoped();
     string type = getAbsolute(p, "_t_");
-    string classType = getAbsolute(p, "_t_", "Disp");
```
InsertCreativityHere, line 554: So many dead variables.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1114,11 +1100,6 @@ Slice::Python::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     registerName(className);
     _out.dec();
 
-    if (_classHistory.count(scoped) == 0)
```
InsertCreativityHere, line 1117: Same but for `InterfaceDecl`.

cpp/src/slice2php/Main.cpp
```diff
@@ -131,11 +131,7 @@ CodeVisitor::visitClassDecl(const ClassDeclPtr& p)
 
         string type = getTypeVar(p);
         _out << sp << nl << "global " << type << ';';
-
-        _out << nl << "if(!isset(" << type << "))";
-        _out << sb;
         _out << nl << type << " = IcePHP_declareClass('" << scoped << "');";
-        _out << eb;
```
InsertCreativityHere, line 138: I believe that this check for `isset` is unnecessary.
This block is wrapped in an `if` statement:
```
if (_classHistory.count(scoped) == 0)
{...
_classHistory.insert(scoped); // Avoid redundant declarations.
}
```
which should ensure that this code is only generated once.

Even if I'm wrong (which I don't think I am), there's nothing wrong with setting `type = IcePHP_declareClass` multiple times, it's just redundant is all.

cpp/src/slice2php/Main.cpp
```diff
@@ -271,12 +264,6 @@ CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
 
     _out << eb; // End of class.
 
-    if (_classHistory.count(scoped) == 0 && p->canBeCyclic())
-    {
-        // Emit a forward declaration for the class in case a data member refers to this type.
-        _out << sp << nl << type << " = IcePHP_declareClass('" << scoped << "');";
-    }
```
InsertCreativityHere, line 278: Impossible to hit.
`visitClassDecl` also checks for `_classHistory.count(scoped) == 0` and inserts `scoped` if so. Since we always call `visitClassDecl` before `visitClassDef`, by the time we get here, the count will _always_ be non-zero.

externl (approved):

bernardnormier (approved):

PR #3365, opened 2025-01-15
Cleanup C++ test headers

externl:
This PR cleans up the headers in the C++ test files.

### DIFF
--- a/cpp/test/DataStorm/types/Writer.cpp
+++ b/cpp/test/DataStorm/types/Writer.cpp
@@ -4,7 +4,8 @@
 
 #include "DataStorm/DataStorm.h"
 #include "Test.h"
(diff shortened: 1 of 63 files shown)


cpp/test/DataStorm/types/Writer.cpp
```diff
@@ -4,7 +4,8 @@
 
 #include "DataStorm/DataStorm.h"
 #include "Test.h"
-#include "TestHelper.h"
+
+#include <iostream>
```
pepone, line 8: The fact that TestHelper was not used highlights a problem. The tests should use the helper runner. I would fix it in a separate PR.

pepone (approved):
Looks good

bernardnormier (approved):

PR #3366, opened 2025-01-15
Cleanup C# doc comment generation

### LINKED ISSUE #3268: Improve slice2cs doc comment conversion
The code currently strip `@link` and `@see` tags because they are supposedly not rendered by Visual Studio

https://github.com/zeroc-ice/ice/blob/910b1744b4a84294cd4f0239fbe6e55c705a91d6/cpp/src/slice2cs/Gen.cpp#L744

https://github.com/zeroc-ice/ice/blob/910b1744b4a84294cd4f0239fbe6e55c705a91d6/cpp/src/slice2cs/Gen.cpp#L765

Now that we are generating docfx API reference it would be better to preserve this and covert them as appropriate.

- `@link` should become `<see cref=...`
- `@see` should become `<seealso cref=...`

pepone:
Fix #3268

### DIFF
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -29,6 +29,66 @@ using namespace IceInternal;
 
 namespace
 {
+    void writeDocLines(Output& out, const StringList& lines)
+    {
+        for (const auto& line : lines)
+        {
+            out << nl << "///";
+            if (!line.empty())
+            {
+                out << " " << line;
+            }
+        }
+    }
+
+    /// Returns a C# formatted link to the provided Slice identifier.
+    string csLinkFormatter(string identifier, string memberComponent)
+    {
+        string result = "<see cref=\"";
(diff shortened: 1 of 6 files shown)


cpp/src/slice2cs/Gen.cpp
```diff
@@ -653,657 +713,98 @@ Slice::CsVisitor::writeDataMemberInitializers(const DataMemberList& dataMembers,
...
-    {
-        return result.erase(0, 1);
-    }
-
-    result[pos] = '.';
-    return result;
-}
-
-string
-Slice::CsVisitor::editMarkup(const string& s)
```
pepone, line 676: I need add the logic for `&lt;` and `&gt;` but would be better to move it to parseDocComment.

bernardnormier (approved):

InsertCreativityHere (approved):
Looks good.
Nicer to use the centralized logic from libSlice as well!

PR #3367, opened 2025-01-15
Allow swift:attribute metadata on structs

### LINKED ISSUE #3259: What Slice Definitions Should `swift:attribute` be Allowed on?
Swift supports a `swift:attribute` metadata, for adding attributes into the generated code. But, there's a major inconsistency in it.

The code-gen checks for this metadata on `struct`, but the validation rejects it on `structs`.
So, if you put `swift:attribute` on a struct, the validation removes it and warns you, even though the code-gen could of used it.

And the code-gen ignores this metadata on `interface`, but the validation allows it on `interface`.
So, if you put `swift:attribute` on an interface, the validation is happy, but it does absolutely nothing.

The documentation for this metadata says it's only supported on `class`, `exception`, and `enum`?
There is no mention of either `struct` or `interface`.

----

I'm not familiar with Swift, but only allowing it on `class`, `exception`, and `enum` feels like a weird list.
Should this be supported on `struct`?, on `interface`?, on anything else?

externl:
Fixes #3259

### DIFF
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -2140,8 +2140,8 @@ SwiftGenerator::MetadataVisitor::validate(const SyntaxTreeBasePtr& p, const Cont
             continue;
         }
 
-        if ((dynamic_pointer_cast<ClassDecl>(p) || dynamic_pointer_cast<InterfaceDecl>(p) ||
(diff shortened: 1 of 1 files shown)


InsertCreativityHere (approved):
It's nice to have the code-gen align with the validation now.

pepone (approved):

PR #3371, opened 2025-01-15
Cleanup 'addImport' and Surrounding Functions in `slice2swift`

InsertCreativityHere:
In `slice2swift` we have two versions of `getTopLevelModule` and `addImport`.
One takes a `ContainedPtr` and one takes a `TypePtr`.

But, the versions which took `TypePtr` were weird and not really necessary.
They basically just converted the `TypePtr` into a `ContainedPtr` to call the other one.
But most of the places where it was called, the `Type` was already a `Contained`.

Anyways, I removed the `TypePtr` versions.
And because of this, alot of unnecessary `dynamic_pointer_cast`s.

### DIFF
--- a/cpp/src/slice2swift/Gen.cpp
+++ b/cpp/src/slice2swift/Gen.cpp
@@ -132,18 +132,10 @@ Gen::ImportVisitor::visitModuleStart(const ModulePtr& p)
 bool
 Gen::ImportVisitor::visitClassDefStart(const ClassDefPtr& p)
 {
-    //
     // Add imports required for base class
-    //
-    ClassDefPtr base = p->base();
-    if (base)
-    {
-        addImport(dynamic_pointer_cast<Contained>(base), p);
-    }
+    addImport(p->base(), p);
 
-    //
     // Add imports required for data members
-    //
     const DataMemberList allDataMembers = p->allDataMembers();
     for (const auto& allDataMember : allDataMembers)
     {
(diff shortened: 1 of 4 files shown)


cpp/src/slice2swift/Gen.cpp
```diff
@@ -132,18 +132,10 @@ Gen::ImportVisitor::visitModuleStart(const ModulePtr& p)
 bool
 Gen::ImportVisitor::visitClassDefStart(const ClassDefPtr& p)
 {
-    //
     // Add imports required for base class
-    //
-    ClassDefPtr base = p->base();
-    if (base)
-    {
-        addImport(dynamic_pointer_cast<Contained>(base), p);
```
InsertCreativityHere, line 141: It's fine to pass a `nullptr` to `addImport`.
So checking `base` before calling it is unnecessary.

Also no need to cast to `Contained` now. `ClassDef` already inherits from `Contained`.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -257,16 +255,6 @@ Slice::getTopLevelModule(const ContainedPtr& cont)
     return m;
 }
 
-ModulePtr
-Slice::getTopLevelModule(const TypePtr& type)
-{
-    // TODO this check is redundant. `InterfaceDecl` _is_ a `Contained`. This whole function is weird though.
```
externl, line 263: The original code in 3.7 made sense. 

https://github.com/zeroc-ice/ice/blob/d53f0a955fd1e925fd0f6159953d781b09b9b1ba/cpp/src/slice2swift/SwiftUtil.cpp#L213-L221

cpp/src/slice2swift/Gen.cpp
```diff
@@ -156,29 +148,19 @@ Gen::ImportVisitor::visitClassDefStart(const ClassDefPtr& p)
...
-    const OperationList operationList = p->allOperations();
-    for (const auto& i : operationList)
+    for (const auto& operation : p->allOperations())
     {
-        const TypePtr ret = i->returnType();
-        if (ret && ret->definitionContext())
-        {
-            addImport(ret, p);
-        }
+        addImport(operation->returnType(), p);
```
externl, line 161: Why don't we need to check `ret->definitionContext()` anymore?

InsertCreativityHere: Because:
`dynamic_pointer_cast<Contained>(p) == true`
implies that
`p->definitionContext() != null`

Both of these are about checking the same thing: "is this a user-defined type?"

Because any user defined check will have a `definitionContext` (as opposed to something like `int`, a built-in type).
And will also be a subclass of `Contained` in the compiler.

PR #3372, opened 2025-01-15
Generate 'operator<<' Implementation for Slice-generated Enums

InsertCreativityHere:
The title pretty much sums it up.
I ended up adding this logic to `ForwardDeclVisitor`, since that's where all the logic for generating enums lives.
I didn't want to add another visitor, so I just re-used this one, even though this isn't about "forward declarations" of course.

`ForwardDeclVisitor` didn't have the ability to write into the source file (only the header file).
So I changed `ForwardDeclVisitor` to take a reference for writing into the source file.

If there's a better place you'd prefer to add this logic, let me know. I'm not married to any of them.

----

### Example

This is the code we generate for `OperationMode`:
```diff
     /// ... not important here
     enum class OperationMode : std::uint8_t
     {
         /// ... not important here
         Normal,
         /// ... not important here
         Nonmutating [[deprecated("Use Idempotent instead.")]],
         /// ... not important here
         Idempotent
     };
+
+    ::std::ostream& operator<<(::std::ostream&, OperationMode value);
```
```diff
+ ::std::ostream&
+ Ice::operator<<(::std::ostream& os, OperationMode value)
+ {
+     switch (value)
+     {
+         case OperationMode::Normal:
+             return os << "Normal";
+         case OperationMode::Nonmutating:
+             return os << "Nonmutating";
+         case OperationMode::Idempotent:
+             return os << "Idempotent";
+         default:
+             return os << static_cast<std::uint8_t>(value);
+     }
+ }
```


pepone (approved):

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1113,9 +1113,12 @@ Slice::Gen::ForwardDeclVisitor::visitInterfaceDecl(const InterfaceDeclPtr& p)
 void
 Slice::Gen::ForwardDeclVisitor::visitEnum(const EnumPtr& p)
 {
+    string mappedName = p->mappedName();
+    string underlying = string("std::") + (p->maxValue() <= numeric_limits<uint8_t>::max() ? "uint8_t" : "int32_t");
```
bernardnormier, line 1117: We should be consistent and use `::std::` everywhere.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1152,6 +1155,28 @@ Slice::Gen::ForwardDeclVisitor::visitEnum(const EnumPtr& p)
         }
     }
     H << eb << ';' << sp;
+
+    H << nl << "::std::ostream& operator<<(::std::ostream&, " << mappedName << " value);";
```
bernardnormier, line 1159: Missing declspec export.

cpp/src/slice2cpp/Gen.h
```diff
@@ -62,7 +62,7 @@ namespace Slice
         class ForwardDeclVisitor final : public ParserVisitor
         {
         public:
-            ForwardDeclVisitor(::IceInternal::Output&);
+            ForwardDeclVisitor(::IceInternal::Output& h, ::IceInternal::Output& c);
```
bernardnormier, line 65: Whenever you add .cpp support, you also need dllExport.

bernardnormier (commented):
I would rather generate this code from the StreamHelper visitor.

bernardnormier:
> I would rather generate this code from the StreamHelper visitor.

Actually no, ForwardDeclVisitor is better. I want to generate similar function for sequences and dictionaries from this visitor too. Maybe we can find a better name for this visitor.

InsertCreativityHere:
I agree that after these changes, we should probably find a better name for `ForwardDeclVisitor`.
But I'm also fine moving this code elsewhere if you decide. `ForwardDeclVisitor` was just the easiest pick because it already had code generating enums.

bernardnormier (approved):

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1152,6 +1155,28 @@ Slice::Gen::ForwardDeclVisitor::visitEnum(const EnumPtr& p)
...
+    C << nl << "switch (value)";
+    C << sb;
+    for (const auto& enumerator : enumerators)
+    {
+        const string enumeratorName = enumerator->mappedName();
+        C << nl << "case " << mappedName << "::" << enumeratorName << ":";
+        C << nl << "    return os << \"" + enumeratorName + "\";";
+    }
+    C << nl << "default:";
+    C << nl << "    return os << static_cast<" << underlying << ">(value);";
```
bernardnormier, line 1177: It's better to cast it to an int (or std::int32_t). `operator<<` for uint8_t is not what you expect.

bernardnormier:
Also:
 - remove https://github.com/zeroc-ice/ice/blob/b039c181a6161754733dc7a8a6a37a4f2d21631e/cpp/include/DataStorm/DataStorm.h#L130
 - I think it would make sense to support custom printing like for structs (see #3378).

cpp/src/slice2cpp/Gen.cpp
InsertCreativityHere: Fixed.

cpp/src/slice2cpp/Gen.cpp
InsertCreativityHere: Completely missed that, fixed!

cpp/src/slice2cpp/Gen.h
InsertCreativityHere: Also fixed!

cpp/src/slice2cpp/Gen.cpp
InsertCreativityHere: I casted it to `std::int32_t`.

cpp/include/DataStorm/DataStorm.h
```diff
@@ -118,36 +118,6 @@ namespace DataStorm
         std::shared_ptr<DataStormI::SampleT<Key, Value, UpdateTag>> _impl;
     };
 
-    /**
```
InsertCreativityHere, line 121: No longer needed, `slice2cpp` will generate this (and thankfully the implementation is the same).

cpp/test/Slice/print/Test.ice
```diff
@@ -8,7 +8,10 @@
 
 module Test
 {
-    enum FlagColor { Red, White, Blue }
+    enum FlagColor { Red, White, ["cpp:identifier:OldGloryBlue"] Blue }
```
InsertCreativityHere, line 11: `OldGloryBlue` is the specific shade of Blue used in the flag.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1158,7 +1166,31 @@ Slice::Gen::ForwardDeclVisitor::visitEnum(const EnumPtr& p)
...
+        C << sp << nl << "::std::ostream&";
+        C << nl << p->mappedScope().substr(2) << "operator<<(::std::ostream& os, " << mappedName << " value)";
+        C << sb;
+        C << nl << "switch (value)";
+        C << sb;
+        for (const auto& enumerator : enumerators)
+        {
+            const string enumeratorName = enumerator->mappedName();
+            C << nl << "case " << mappedName << "::" << enumeratorName << ":";
+            C << nl << "    return os << \"" + enumeratorName + "\";";
```
bernardnormier, line 1187: We prefer .inc() and .dec() over adding spaces "by hand".

bernardnormier (approved):

Issue #3120, opened 2024-11-11
`slice2swift` Does not Correctly Generate Links to Symbols in Nested Modules


InsertCreativityHere:
In Swift, packages are flat, meaning things can only live 1 package deep, there is no notion of "nested packages".
To fix this, any Slice constructs which are within nested modules will have the extra modules rolled into their names. Ex:
```
["swift:module:Ice:MX]
module IceMX
{
    struct MetricsFailures { ... }
}
```
will generate a struct named `MXMetricsFailures` which lives inside the `Ice` package.

This approach really screws with doc-comments.
Let's say we now do `{@link MetricsFailures}`. All we see is this string. There's no collision with reserved identifiers or keywords, so we don't change the string, and generate a Swift link to it: ` ``MetricsFailures`` `. However, nothing with that name exists. Since it _should_ be mapped to `MXMetricsFailures`.

InsertCreativityHere:
I encountered this problem in `slice2swift`, but I bet that this problem also exists anywhere where users use metadata to map module names to something else. And will become a huge headache once we add `xxx:identifier` metadata.

So the fix isn't specific to `slice2swift`, but is instead to perform some kind of type-lookup on any `@link`s, instead of just passing the raw string along. This would also have the added benefit that we could issue warnings for dead links.

PR #3374, opened 2025-01-16
Improve slice2py generated docstrings

pepone:
This PR upgrades slice2py to use the Slice builtin comment parsing, and removes the custom parsing from slice2py.

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -125,6 +125,24 @@ namespace
 
         return "???";
     }
+
+    /// Returns a DocString formatted link to the provided Slice identifier.
+    string pyLinkFormatter(string identifier, string memberComponent)
+    {
+        ostringstream os;
+        os << "`";
+        if !(identifier.empty())
+        {
+            os << Slice::Python::fixIdent(identifier);
+            if !(memberComponent.empty())
+            {
+                os << ".";
+            }
+        }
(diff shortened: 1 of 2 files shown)


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -125,6 +125,24 @@ namespace
...
+        {
+            os << Slice::Python::fixIdent(identifier);
+            if !(memberComponent.empty())
+            {
+                os << ".";
+            }
+        }
+        os << Slice::Python::fixIdent(memberComponent);
+        os << "`";
+        return os.str();
```
pepone, line 144: We need to improve this, but I don't think the current callback provides enough context.

See the expected format https://www.sphinx-doc.org/en/master/usage/domains/python.html#cross-referencing-python-objects

Here to emit `:py:meth:`, `:py:class:`, ... we need to lookup the identifier. But lookup depends on the context where the links appear, and the link formatter doesn't have this context.

bernardnormier (approved):

cpp/src/slice2py/PythonUtil.cpp
InsertCreativityHere: Yes, there's a couple other languages with this problem.
And now that we have metadata to remap identifiers, lookup is essential in all languages.
Feel free to comment on the issue: https://github.com/zeroc-ice/ice/issues/3120

InsertCreativityHere (approved):

PR #3377, opened 2025-01-16
Run ice2slice on CI

### LINKED ISSUE #3351: Run `ice2slice` in CI
It would be good to have at least some kind of testing for `ice2slice` run in CI.
Even just running it over the `./slice` directory would be good enough I think.

Because right now, where we have 0 testing of this tool, it's too east to accidentally break and not notice.
As shown from the fact that it was broken for months without anyone noticing until #3290.

pepone:
Fix #3351

### DIFF
--- a/cpp/src/ice2slice/Main.cpp
+++ b/cpp/src/ice2slice/Main.cpp
@@ -113,6 +113,10 @@ compile(const vector<string>& argv)
     bool preprocess = opts.isSet("E");
 
     string outputDir = opts.optArg("output-dir");
+    if (outputDir.empty())
+    {
+        outputDir = ".";
+    }
 
     bool debug = opts.isSet("debug");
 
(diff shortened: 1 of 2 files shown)


cpp/src/ice2slice/Main.cpp
```diff
@@ -113,6 +113,10 @@ compile(const vector<string>& argv)
     bool preprocess = opts.isSet("E");
 
     string outputDir = opts.optArg("output-dir");
+    if (outputDir.empty())
```
pepone, line 116: This was broken, not emitting any files when you don't specify the output dir. But the compiler din't give any errors, which seems odd.

externl (approved):

cpp/src/ice2slice/Main.cpp
InsertCreativityHere: Huh, that's a little weird. It just passes this into `libSlice`, so there could be something weird going on downstream.
I'll take a peek at it later.

InsertCreativityHere (approved):

PR #3378, opened 2025-01-17
Print generated class/struct in C++

### LINKED ISSUE #3324: Add Support for Printing Generated Slice Types in C++ through `operator<<`
As discussed in https://github.com/orgs/zeroc-ice/discussions/2980, we should add support for printing generated Slice types.
We should add support for printing the following:
- Builtin (most of these are already supported by virtue of being mapped to primitive types)
- Class
- Interface (really proxy)
- Struct
- Sequence
- Dictionary
- Enum

**What should we do for exceptions?**
Since exceptions are not types, they can't be held in any of the above Slice definitions.
Additionally, all the generated exceptions inherit support for `operator<<` from `Ice::Exception`:
https://github.com/zeroc-ice/ice/blob/7b37d63a397cc50b79acd2e855c94d31c89c925b/cpp/src/Ice/Exception.cpp#L503-L524
and we have the `cpp:ice_print` metadata, allowing users to provide their own `ice_print` implementations.

bernardnormier:
This PR adds printing to generated structs and classes in C++, using `operator<<(ostream&, const T&)`.

Fixes #3324.

The implementation piggy backs on the existing StreamHelpers. It adds a new `print` function on the StreamHelper template. It also adds an `Ice::print` function that prints any mapped field type (see StreamHelpers.h).

This implementation detects cycles when printing classes, to avoid stack overflow.

This implementation also supports custom printing with the new `["cpp:custom-print"]` metadata for structs and classes. For structs, the application needs to implement `operator<<(ostream&, const T&)`, while for classes, the application needs to implement `T::ice_print(ostream&) const`.


cpp/include/Ice/Proxy.h
```diff
@@ -860,6 +861,22 @@ namespace Ice
         IceInternal::ReferencePtr _reference;
         IceInternal::RequestHandlerCachePtr _requestHandlerCache;
     };
+
+    ICE_API std::ostream& operator<<(std::ostream&, const ObjectPrx&);
```
bernardnormier, line 865: Moved from ProxyFunctions.h to Proxy.h, because the generated code "sees" Proxy.h but not ProxyFunctions.h

cpp/include/Ice/StreamHelpers.h
```diff
@@ -7,18 +7,42 @@
...
 #include <iterator>
+#include <sstream>
 
 #if __has_include(<span>)
 #    include <span>
 #endif
 
+namespace std
+{
+    // TODO: temporary
```
bernardnormier, line 21: Remove when #3372 is merged.

cpp/test/Slice/print/AllTests.cpp
```diff
@@ -0,0 +1,129 @@
...
+    vector<byte> byteSeq{byte{100}, byte{150}, byte{255}};
+    vector<bool> boolSeq{true, false, true};
+    ByteBoolStruct byteBoolStruct{byteSeq, boolSeq};
+
+    testPrint(byteBoolStruct, "Test::ByteBoolStruct{myByteSeq = [100, 150, 255], myBoolSeq = [true, false, true]}");
+
+    testPrint(
+        byteBoolStruct,
+        "Test::ByteBoolStruct{myByteSeq = [0x64, 0x96, 0xff], myBoolSeq = [true, false, true]}",
+        [](ostream& os) { os << hex << showbase; });
```
bernardnormier, line 71: The user can easily change the formatting. The Ice "print" functions don't change the format flags of the stream.

cpp/test/Slice/print/AllTests.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include "Test.h"
+#include "TestHelper.h"
+
+#include <iomanip>
+
+using namespace std;
+using namespace Test;
+
+void
+Test::Neighbor::ice_print(ostream& os) const
```
bernardnormier, line 13: Example of custom class printing. For custom struct printing, see Ice::Identity, Ice::EncodingVersion.

pepone (approved):

externl (approved):

cpp/include/Ice/Value.h
```diff
@@ -96,6 +109,35 @@ namespace Ice
     private:
         SlicedDataPtr _slicedData;
     };
+
+    /// Outputs the description of a class instance to the stream. This function calls ice_print on value.
+    /// @param os The output stream.
+    /// @param value The class instance.
+    /// @return The output stream.
+    inline std::ostream& operator<<(std::ostream& os, const Value& value)
```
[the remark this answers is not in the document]
bernardnormier: We don't strictly need it, but I think it's nicer and more consistent with mapped exceptions, where (an) `operator<<` is "equivalent" to `ice_print`. See #3380.

slice/Ice/Version.ice
```diff
@@ -13,6 +13,7 @@
 [["cpp:include:Ice/Config.h"]]
 [["cpp:include:Ice/Comparable.h"]]
 [["cpp:include:cstdint"]]
+[["cpp:include:ostream"]]
```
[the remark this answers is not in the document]
bernardnormier: Because we have above:
```
[["cpp:no-default-include"]]
```

which means all C++ includes must be explicit. And since we define structs, we always get:
```cpp
void ice_printFields(std::ostream& os) const;
```

cpp/src/Ice/Value.cpp
```diff
@@ -3,15 +3,78 @@
...
+        }
+
+        bool good() const noexcept { return _currentStack; }
+
+    private:
+        ostream& _os;
+        deque<const Value*> _localStack;
+        deque<const Value*>* _currentStack{nullptr}; // may or may not point to _localStack
+
+        static const int _cycleCheckIndex;
```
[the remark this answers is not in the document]
bernardnormier: No. See https://en.cppreference.com/w/cpp/io/ios_base/xalloc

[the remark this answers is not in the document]
pepone: I see the API is thread safe. There is still an issue that this would show already printed when same class is being printed by concurrent calls.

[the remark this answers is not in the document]
bernardnormier: I am not following. 

The index (an int) is global and thread-safe. The data we store (pointer to a deque) is per `ostream` instance and not shared by all `ostream` instances.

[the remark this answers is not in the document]
pepone: Never mind you are right. I was missing that the allocation is done on a per stream with the pword call.

cpp/src/DataStorm/TraceUtil.cpp
```diff
@@ -32,3 +32,91 @@ TraceLevels::TraceLevels(const PropertiesPtr& properties, LoggerPtr logger)
 #elif defined(__GNUC__)
 #    pragma GCC diagnostic pop
 #endif
+
+// Our custom implementation of various operator<< for structs defined in Contract.ice
+
+namespace DataStormContract
```
InsertCreativityHere, line 38: It would be nice if these were in the same order as the Slice definitions, instead of a seemingly random order.
The Slice order:
```
struct DataSamples
struct ElementInfo
struct TopicInfo
struct TopicSpec
struct ElementData
struct ElementSpec
struct ElementDataAck
struct ElementSpecAck
```

InsertCreativityHere: Nevermind, I didn't realize this was just moved from `TraceUtil.h`.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -772,16 +772,18 @@ Slice::Gen::generate(const UnitPtr& p)
         }
     }
 
-    if (!dc->hasMetadata("cpp:no-default-include"))
+    // Include Ice.h since it was not included in the header.
```
InsertCreativityHere, line 775: So, as of this PR, the only effect of `cpp:no-default-includes` is to move
`#include <Ice/Ice.h>`
from the header file to the source file?

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -2317,8 +2379,7 @@ Slice::Gen::DataDefVisitor::visitClassDefEnd(const ClassDefPtr& p)
     C << nl << "return CloneEnabler<" << name << ">::clone(*this);";
     C << eb;
 
-    const string baseClass = base ? getUnqualified(base->mappedScoped(), scope) : getUnqualified("::Ice::Value", scope);
-
+    H << sp;
```
InsertCreativityHere, line 2382: You added this to fix a formatting bug?

InsertCreativityHere (approved):
Looks good.
I'll let you merge first, then I guess I'll deal with the merge conflicts in my PR,
and remove the stop-gap you put in for enums.

cpp/src/slice2cpp/Gen.cpp
bernardnormier: Yes.

cpp/src/slice2cpp/Gen.cpp
bernardnormier: Yes.

PR #3380, opened 2025-01-18
Simplify C++ exceptions

bernardnormier:
This PR simplifies exceptions in C++ and updates the base classes as follows:

### Ice::Exception

The common base class, provides what is truly common between local exceptions and user exceptions, namely:
  - `ice_print(ostream&) const`, used by the implementation of `Ice::operator<<(ostream&, const Exception&)`
  - `ice_id()`, the Slice type ID or a similar string for local exceptions

It has only 2 derived classes, LocalException and UserException (this didn't change).

### Ice::LocalException

The base class for Ice exceptions not defined in Slice. It's a true std::exception, with a custom what() message. It also holds where the exception was thrown (file + line) and can capture / provide its stack trace.

This PR moves all the custom what / stack trace code from Ice::Exception to Ice::LocalException.

### Ice::UserException

The base class for Slice-defined exceptions. User exceptions are primary class-like errors transmitted over the wire. The exception-aspect is secondary. In particular, we don't capture the stack trace nor provide a custom `what` message. 

`what()` is always the same as the Slice type ID (returned by `ice_id()`), and is typically never called.


cpp/src/Ice/LocalException.cpp
```diff
@@ -1,9 +1,513 @@
 // Copyright (c) ZeroC, Inc.
 
+#if defined(_MSC_VER)
```
bernardnormier, line 3: This code was largely moved from Exception.cpp. In particular, there was no change to the stack trace capture logic.

cpp/src/Ice/Exception.cpp
```diff
@@ -1,524 +1,5 @@
...
-        // Leaked, as libbacktrace does not provide an API to free this state.
-        bstate = backtrace_create_state(0, 1, ignoreErrorCallback, 0);
-    }
-#elif defined(ICE_BACKTRACE)
-    backTraceEnabled = true;
-#endif
-}
-
-ostream&
-Ice::operator<<(ostream& out, const Ice::Exception& ex)
```
bernardnormier, line 504: This PR eliminates the distinction `ice_print` / `operator<<`, which made this code too complicated.

As of this PR, (virtual) ice_print and `operator<<` are the same.

cpp/include/Ice/Exception.h
```diff
@@ -1,112 +1,49 @@
...
-        /**
-         * Returns the type ID of this exception. This corresponds to the Slice type ID for Slice-defined exceptions,
-         * and to a similar fully scoped name for other exceptions. For example "::Ice::CommunicatorDestroyedException".
-         * @return The type ID of this exception
-         */
+        Exception() = default;
+        Exception(const Exception&) = default;
+        Exception& operator=(const Exception&) = default;
+
+        // Need out-of-line virtual function to avoid weak vtable, which in turn requires the default constructor,
```
bernardnormier, line 22: Weak vtables break `dynamic_cast` and `catch(const XxxException&)`, and it's typically hard to find since we see these errors only at runtime.  This particular weak vtable breaks the Ice/background test when it catches `const Ice::Exception&` on macos.

cpp/src/Ice/UserException.cpp
```diff
@@ -9,6 +9,18 @@
...
+const char*
+Ice::UserException::what() const noexcept
+{
+    return ice_id();
+}
+
+void
+Ice::UserException::ice_print(ostream& os) const
+{
+    os << ice_id();
```
bernardnormier, line 21: This implementation will change in a follow-up PR. See #3378.

cpp/test/Ice/exceptions/AllTests.cpp
```diff
@@ -49,58 +49,39 @@ allTests(Test::TestHelper* helper)
...
         }
         {
-            stringstream str;
+            ostringstream str;
             customUle.ice_print(str);
-            test(str.str() == customUlePrint);
-        }
-
-        //
-        // Test operator<<().
```
bernardnormier, line 80: No need to test operator<< since it's now identical to ice_print.

externl (approved):

cpp/src/Ice/LocalException.cpp
```diff
@@ -1,9 +1,514 @@
...
+
+#endif
+        return stackTrace;
+    }
+}
+
+Ice::LocalException::LocalException(const char* file, int line, string message)
+    : _file(file),
+      _line(line),
+      _whatString(make_shared<string>(std::move(message))),
```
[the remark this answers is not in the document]
bernardnormier: No, we have to use a shared_ptr. Same for other exceptions with string and similar fields.

This is to ensure the copy ctor of the exception is noexcept.

cpp/include/Ice/LocalException.h
```diff
@@ -1,30 +1,54 @@
...
-         * @param message The error message adopted by this exception and returned by what().
-         */
-        LocalException(const char* file, int line, std::string message) : Exception(file, line, std::move(message)) {}
+        /// Constructs a local exception.
+        /// @param file The file where this exception is constructed. This C string is not copied.
+        /// @param line The line where this exception is constructed.
+        /// @param message The error message adopted by this exception and returned by what().
+        LocalException(const char* file, int line, std::string message);
+
+        [[nodiscard]] const char* what() const noexcept final;
```
InsertCreativityHere, line 25: I know that `what` is pretty standard C++ exception stuff, but should we carry over the doc-comment for it?
```
        /**
         * Returns the error message of this exception.
         * @return The error message.
         */
        [[nodiscard]] const char* what() const noexcept override;
```

cpp/src/Ice/LocalException.cpp
```diff
@@ -1,9 +1,514 @@
...
+Ice::LocalException::ice_print(ostream& os) const
+{
+    // We tolerate a null file / 0 line, even though the thrower should always set them with __FILE__, __LINE__.
+    if (_file && _line > 0)
+    {
+        os << _file << ':' << _line << ' ';
+    }
+
+    // +2 to remove the leading "::" from the type ID.
+    os << ice_id() + 2 << ' ' << what();
```
InsertCreativityHere, line 428: I didn't know you could do that.
What a crazy syntax.

InsertCreativityHere (approved):
Looks good!

PR #3382, opened 2025-01-20
Remove protected metadata

### LINKED ISSUE #3360: Remove protected Slice metadata
This metadata goes hand-in-hand with Slice operations on operations, that were removed in 3.8. So we should remove this metadata as well.

pepone:
Fix #3360


cpp/test/Ice/objects/Test.ice
```diff
@@ -61,15 +61,15 @@ class D
     bool postUnmarshalInvoked;
 }
 
-["protected"] class E
```
bernardnormier, line 64: We should remove these tests since they were apparently all about testing protected.

cpp/test/Ice/optional/Test.ice
```diff
@@ -149,7 +149,7 @@ exception RequiredException extends OptionalException
 class OptionalWithCustom
 {
     optional(1) SmallStructList l;
-    ["protected"] optional(2) SmallStructList lp;
```
bernardnormier, line 152: Same here. `lp` is the same as `l` (great name) except protected.

bernardnormier (approved):
Looks good, except I would remove the protected tests, not leave them behind without protected.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -2258,55 +2258,15 @@ Slice::Gen::DataDefVisitor::visitClassDefEnd(const ClassDefPtr& p)
 
     // Emit data members. Access visibility may be specified by metadata.
     const DataMemberList dataMembers = p->dataMembers();
-    const bool prot = p->hasMetadata("protected");
-    bool inProtected = false;
-    bool needSp = true;
```
InsertCreativityHere, line 2263: What does the formatting look like without `needSp`?
It seems like this field was for emitting a newline before the first data member, more than it was for protected stuff.

cpp/test/Slice/errorDetection/WarningInvalidMetadata.ice
```diff
@@ -100,15 +100,15 @@ class P
 // Metadata is shared between forward declarations and definitions.
 // We allow duplicate metadata, but require that that metadata must be identical.
 
-["java:nonsense", "deprecated:hello", "protected"]
+["java:nonsense", "deprecated:hello"]
 class K;
```
InsertCreativityHere, line 104: These tests weren't specifically for "protected", but metadata in general.
Since we can no longer use `protected` to check the behavior of duplicate metadata on multiple declarations,
we should do this instead:
```suggestion
["java:nonsense", "deprecated:hello", "amd"]
interface K;
```

cpp/test/Slice/errorDetection/WarningInvalidMetadata.ice
```diff
@@ -100,15 +100,15 @@ class P
...
 
-["java:nonsense", "deprecated:goodbye", "protected"]
+["java:nonsense", "deprecated:goodbye"]
 class K
```
InsertCreativityHere, line 107:
```suggestion
["java:nonsense", "deprecated:goodbye", "amd"]
interface K
```

cpp/test/Slice/errorDetection/WarningInvalidMetadata.ice
```diff
@@ -100,15 +100,15 @@ class P
...
 {
 }
 
-["java:nonsense", "deprecated", "protected"]
+["java:nonsense", "deprecated"]
 class K;
```
InsertCreativityHere, line 112:
```suggestion
["java:nonsense", "deprecated", "amd"]
interface K;
```

InsertCreativityHere (approved):
Looks good, but I agree we should get rid of the `lp` field in the Ice/optional test.

It looks like we use `E` in the `Ice/Objects` test for other purposes. So maybe that one is fine to leave.

PR #3385, opened 2025-01-20
Update copyright message

externl:
This PR updates the copyright message in our files to use our new simplified version.

```
// Copyright (c) ZeroC, Inc.
```


cpp/config/Make.xcframework.rules
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) ZeroC, Inc. All rights reserved.
+#  Copyright (c) ZeroC, Inc.
```
InsertCreativityHere, line 1:
```suggestion
# Copyright (c) ZeroC, Inc.
```

InsertCreativityHere (commented):
So far so good.

python/python/Ice/ValueFactory.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) ZeroC, Inc. All rights reserved.
+#  Copyright (c) ZeroC, Inc.
```
InsertCreativityHere, line 1: It also seems like _some_ of the Python files also got an extra space inserted in them.

Not the files which have a shebang, those python files seem fine.
But most of the other ones.

matlab/test/Ice/objects/CB8I.m
```diff
@@ -1,5 +1,5 @@
 %
-% Copyright (c) ZeroC, Inc. All rights reserved.
+% Copyright (c) ZeroC, Inc.
 %
```
InsertCreativityHere, line 3: And the MATLAB files still have a leading/trailing blank line around the copyright.
Maybe that's 'expected' in MATLAB though. Idk.

InsertCreativityHere (approved):
I only looked at 1140 files out of the 3057.
But, it seems good now.

Glad to finally have all these copyrights fixed.

PR #3391, opened 2025-01-20
Simplify operator<< for DataStorm sequence

bernardnormier:
This is a small cleanup PR.

### DIFF
--- a/cpp/include/DataStorm/DataStorm.h
+++ b/cpp/include/DataStorm/DataStorm.h
@@ -157,18 +157,9 @@ namespace DataStorm
      * @param types The sample type vector to add to the stream
      * @return The output stream
      */
-    inline std::ostream& operator<<(std::ostream& os, const std::vector<SampleEvent>& types)
+    inline std::ostream& operator<<(std::ostream& os, const SampleEventSeq& types)
     {
-        os << "[";
-        for (auto p = types.begin(); p != types.end(); ++p)
-        {
-            if (p != types.begin())
-            {
-                os << ',';
-            }
-            os << *p;
-        }
-        os << "]";
+        Ice::print(os, types);
         return os;
     }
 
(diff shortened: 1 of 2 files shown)


cpp/include/DataStorm/DataStorm.h
```diff
@@ -157,18 +157,9 @@ namespace DataStorm
...
-        for (auto p = types.begin(); p != types.end(); ++p)
-        {
-            if (p != types.begin())
-            {
-                os << ',';
-            }
-            os << *p;
-        }
-        os << "]";
+        Ice::print(os, types);
```
bernardnormier, line 162: The format is not quite the same, as Ice::print adds a space after each comma.

pepone (approved):

cpp/include/Ice/OutgoingResponse.h
```diff
@@ -99,8 +99,8 @@ namespace Ice
 
         /**
          * Get the full details of the exception marshaled into the response.
-         * @return The exception details. For Ice exceptions, it's usually produced by streaming the exception into
-         * a std::ostringstream with operator<<. It's empty when replyStatus() is ReplyStatus::Ok.
+         * @return The exception details. For Ice exceptions, it's produced by streaming the exception into a
+         * std::ostringstream with operator<<. It's empty when replyStatus() is ReplyStatus::Ok.
```
InsertCreativityHere, line 103: I know that you were just (correctly) removing `usually` here... but the `a` isn't correct here. It should be:
```suggestion
         * @return The exception details. For Ice exceptions, it's produced by streaming the exception into an
         * std::ostringstream with operator<<. It's empty when replyStatus() is ReplyStatus::Ok.
```
Because we're pronouncing the _letter_ s which starts with a vowel (it's pronounced like `es`).

InsertCreativityHere (approved):
Looks good.

PR #3393, opened 2025-01-20
Marshal and print enum by value

bernardnormier:
This is a small update to the StreamHelpers to marshal and print enumerations by value.


InsertCreativityHere (approved):
Glad to see we were on a similar page.
Looks good.

pepone (approved):

PR #3395, opened 2025-01-21
Fix for Ice/interrupt test

pepone:
#3394


bernardnormier:
Looks good. We should rework this test in a follow-up PRs, with comments explaining what we actually test.

bernardnormier (approved):

PR #3398, opened 2025-01-21
Remove ICE_BUILDING_xxx macros, streamline Config.h

bernardnormier:
This PR removes a number of ICE_BUILDING_xxx macros and streamlines the Xxx/Config.h.

Most of these Config.h are included in generated headers and used to automatically link the corresponding "stub" library on Windows.

After this PR, there are 4 remaining ICE_BUILDING_xxx macros:

- ICE_BUILDING_SLICE_COMPILERS - Windows only, used by the C++ build
- ICE_BUILDING_GENERATED_CODE - to reduce headers included in generated .cpp
- ICE_BUILDING_ICE and ICE_BUILDING_ICE_LOCATOR_DISCOVERY - used by the Python build, possibly only on Windows. Would be nice to review/remove in a follow-up PR.


pepone (approved):

cpp/include/Glacier2/Config.h
```diff
@@ -3,22 +3,12 @@
 #ifndef GLACIER2_CONFIG_H
 #define GLACIER2_CONFIG_H
 
-#include "Ice/Config.h"
+// This header is included in all Glacier generated headers.
```
externl, line 6:
```suggestion
// This header is included in all Glacier2 generated headers.
```

externl (approved):

PR #3399, opened 2025-01-21
Simplify Comment Parsing Code

InsertCreativityHere:
This PR simplifies the implementation of `parseDocComment`, which is the part of `libSlice` responsible for converting a raw `string comment` into a structured `DocComment`.

It's not meant to be a perfect cleanup.
But just to simplify some things before I add the last validation logic/fix the issues related to how `@link`s are formatted.

I used my compiler-comparison script to check, and the generated code is identical before and after my changes here.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -643,12 +643,11 @@ namespace
     StringList
     splitComment(string comment, function<string(string, string)> linkFormatter, bool stripMarkup, bool xmlEscape)
     {
-        string::size_type pos;
+        string::size_type pos = 0;
 
         if (stripMarkup)
         {
             // Strip HTML markup.
-            pos = 0;
             while ((pos = comment.find('<', pos)) != string::npos)
             {
                 string::size_type endpos = comment.find('>', pos);
@@ -836,10 +835,6 @@ Slice::Contained::parseDocComment(function<string(string, string)> linkFormatter
             {
                 comment->_parameters[name] = {};
                 currentSection = &(comment->_parameters[name]);
-                if (!lineText.empty())
-                {
-                    currentSection->push_back(lineText);
-                }
             }
         }
         else if (parseNamedCommentLine(l, throwsTag, name, lineText))
@@ -855,10 +850,6 @@ Slice::Contained::parseDocComment(function<string(string, string)> linkFormatter
             {
                 comment->_exceptions[name] = {};
                 currentSection = &(comment->_exceptions[name]);
-                if (!lineText.empty())
-                {
-                    currentSection->push_back(lineText);
-                }
             }
         }
         else if (parseNamedCommentLine(l, exceptionTag, name, lineText))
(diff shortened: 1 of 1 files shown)


cpp/src/Slice/Parser.cpp
```diff
@@ -648,48 +648,44 @@ namespace
         if (stripMarkup)
         {
             // Strip HTML markup.
```
InsertCreativityHere, line 650: Switching to a while loop lets us avoid a redundant check for `pos != string::npos`.

cpp/src/Slice/Parser.cpp
```diff
@@ -648,48 +648,44 @@ namespace
...
-                    {
-                        break;
-                    }
-                    comment.erase(pos, endpos - pos + 1);
+                    break;
                 }
-            } while (pos != string::npos);
+                comment.erase(pos, endpos - pos + 1);
+            }
+        }
```
InsertCreativityHere, line 661: This logic was nested inside `if (stripMarkup)`. This doesn't seem right to me.
XML escaping should be unrelated to markdown striping.

cpp/src/Slice/Parser.cpp
```diff
@@ -742,35 +738,27 @@ namespace
         return result;
     }
```
InsertCreativityHere, line 740: This function took a `bool namedTag` parameter which was always hard-coded.
And even worse, it took a `const string& name` which was only used some of the time based on the bool.
So I split it into 2 functions: `parseNamedCommentLine` and `parseCommentLine`. No more bool.

cpp/src/Slice/Parser.cpp
```diff
@@ -803,94 +807,69 @@ Slice::Contained::parseDocComment(function<string(string, string)> linkFormatter
 
     DocCommentPtr comment = make_shared<DocComment>();
```
InsertCreativityHere, line 808: All the remaining code below this is where we parse the actual tags (`@param` etc) into their respective fields on `DocComment`. Before, we had an entire state machine, but this was overkill.

I removed all this logic, and replaced it with a `StringList* currentSection`,
which points to whichever part of the doc-comment we're currently parsing.
At the beginning it points to the overview, but as we encounter tags, we re-point it appropiately.
I find this approach simpler than using a state-machine, and switching on the state to determine where we're writing to.

cpp/src/Slice/Parser.cpp
pepone: The code assumes anything between `<` and `>` is an html tag.

What about something like:

`a < b or x > y` , won't this strip comments like to `a  y`?

It is not a new problem, but would be nice to fix now that we are looking at this code.

We can fix in a separate PR.

pepone (approved):

InsertCreativityHere:
I agree, it's also a problem I noticed, but didn't want to fix in this PR.
At first I thought about making the check fail if there was whitespace within the `< ... >`...
but I wasn't sure enough that that was a safe assumption.

PR #3400, opened 2025-01-22
More c++ lint fixes

bernardnormier:
This PR fixes more clang-tidy -reported lints in the C++ code:
 - modernize-return-braced-init-list everywhere
 - makes all swap noexcept
 - all lints in src/Ice/*.cpp


.clang-format-ignore
```diff
@@ -1,9 +1,12 @@
 # C++ code generated from Slice files
-cpp/include/generated
-cpp/src/**/generated
-cpp/test/**/generated
-cpp/src/**/msbuild/*
-cpp/test/**/msbuild/*
+cpp/include/generated/**/*.h
```
bernardnormier, line 2: The previous version didn't work, and reformatting the generated code messed up the NOLINT comments.

cpp/include/Ice/OutgoingAsync.h
```diff
@@ -299,7 +299,7 @@ namespace IceInternal
             Ice::OperationMode mode,
             std::optional<Ice::FormatType> format,
             const Ice::Context& ctx,
-            std::function<void(Ice::OutputStream*)> write,
+            const std::function<void(Ice::OutputStream*)>& write,
```
bernardnormier, line 302: Since write / marshal is synchronous, we only need a const&

cpp/src/Ice/IPEndpointI.cpp
```diff
@@ -290,6 +290,7 @@ vector<ConnectorPtr>
 IceInternal::IPEndpointI::connectors(const vector<Address>& addresses, const NetworkProxyPtr& proxy) const
 {
     vector<ConnectorPtr> connectors;
+    connectors.reserve(addresses.size());
```
bernardnormier, line 293: vector has a ctor that takes a size_t, but it's not a "reserve" constructor (unlike C#). It adds n default elements.

cpp/src/Ice/LocalExceptions.cpp
```diff
@@ -32,9 +32,9 @@ Ice::RequestFailedException::~RequestFailedException() = default; // avoid weak
 Ice::ObjectNotExistException::ObjectNotExistException(
     const char* file,
     int line,
-    Identity id,
-    string facet,
-    string operation)
+    Identity id,      // NOLINT(performance-unnecessary-value-param)
```
bernardnormier, line 35: See comment on line 30 above.

cpp/src/Ice/ObjectAdapterFactory.cpp
```diff
@@ -114,7 +114,7 @@ IceInternal::ObjectAdapterFactory::updateObservers(void (ObjectAdapterI::*fn)())
 
 ObjectAdapterPtr
 IceInternal::ObjectAdapterFactory::createObjectAdapter(
-    string name,
+    string name, // NOLINT(performance-unnecessary-value-param)
```
pepone, line 117: Can we instead move it to OA constructor, and then use getName for inserting in _adapterNamesInUse map

cpp/src/Ice/Proxy.cpp
```diff
@@ -394,7 +394,7 @@ Ice::ObjectPrx::_endpointSelection(EndpointSelectionType newType) const
 }
 
 ReferencePtr
-Ice::ObjectPrx::_endpoints(EndpointSeq newEndpoints) const
+Ice::ObjectPrx::_endpoints(EndpointSeq newEndpoints) const // NOLINT(performance-unnecessary-value-param)
 {
     vector<EndpointIPtr> endpoints;
```
pepone, line 399: Not related to PR, but can't we do the comparison without creating this intermediary vector?

cpp/src/Ice/Proxy.cpp
```diff
@@ -447,7 +447,7 @@ Ice::ObjectPrx::_facet(string newFacet) const
 }
 
 ReferencePtr
-Ice::ObjectPrx::_fixed(ConnectionPtr connection) const
+Ice::ObjectPrx::_fixed(ConnectionPtr connection) const // NOLINT(performance-unnecessary-value-param)
```
pepone, line 450: We can fix this one when we require C++20, there is a `dynamic_pointer_cast` version that takes a rvalue reference.

pepone (approved):

.clang-format-ignore
externl: Interesting, we've been using this with CI for a while.

externl (approved):

cpp/src/Ice/ObjectAdapterFactory.cpp
bernardnormier: It doesn't work well because we use name at the end to remove it from adapterNamesInUse on exception. It's much simpler to ignore this lint.

cpp/src/Ice/LoggerAdminI.cpp
```diff
@@ -708,7 +708,7 @@ namespace
 
                 try
                 {
-                    RemoteLoggerPrx remoteLogger = *p;
+                    RemoteLoggerPrx remoteLogger{*p}; // NOLINT(performance-unnecessary-copy-initialization)
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

cpp/src/Ice/Proxy.cpp
bernardnormier: We could, but then we need the intermediary vector in the usual path where we call changeEndpoints.

PR #3405, opened 2025-01-22
Use .NET 9.x form Linux ARM64 tests

pepone:
Switch ARM Linux tests to .NET 9 to see if it is more stable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,6 @@ jobs:
             config: "release"
           - os: ubuntu-24.04-arm
             config: "release"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -27,6 +27,7 @@ jobs:
             config: "release"
           - os: ubuntu-24.04-arm
             config: "release"
+            dotnet_version: "9.x" # Use .NET 9.x on ARM https://github.com/dotnet/runtime/issues/104123
```
externl, line 30: Instead I think we should just set the version based on the os name and not bother trying to create a variable.

externl (approved):

pepone:
Seems .NET 9 doesn't work well either.

PR #3406, opened 2025-01-22
Fix more clang-tidy lints (C++)

bernardnormier:
This PR fixes clang-tidy lints in Glacier2, IceBox, IceBridge, IceStorm and DataStorm.

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,13 +1,3 @@
-# C++ code generated from Slice files
-cpp/include/generated/**/*.h
-cpp/src/**/generated/**/*.cpp
-cpp/test/**/generated/**/*.h
-cpp/test/**/generated/**/*.cpp
-cpp/src/**/msbuild/**/*.h
-cpp/src/**/msbuild/**/*.cpp
-cpp/test/**/msbuild/**/*.h
-cpp/test/**/msbuild/**/*.cpp
-
 # C++ code generated from the Bison/Flex parsers
 cpp/src/Slice/Grammar.h
 cpp/src/Slice/Grammar.cpp
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -9,6 +9,7 @@ Checks:
   -modernize-use-trailing-return-type,
   -modernize-concat-nested-namespaces,
   -modernize-use-default-member-init,
+  -modernize-macro-to-enum,
   performance-*,
   -performance-avoid-endl,
   -performance-enum-size,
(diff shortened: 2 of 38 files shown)


.clang-format-ignore
```diff
@@ -1,13 +1,3 @@
-# C++ code generated from Slice files
```
bernardnormier, line 1: I could not get this to work. The solution is to generate a clang-format off in the .h files.

.clang-tidy
```diff
@@ -9,6 +9,7 @@ Checks:
   -modernize-use-trailing-return-type,
   -modernize-concat-nested-namespaces,
   -modernize-use-default-member-init,
+  -modernize-macro-to-enum,
```
bernardnormier, line 12: That's because there are such macros in the generated Grammar.cpp _before_ we can generate the NOLINTBEGIN.

cpp/src/IceBridge/IceBridge.cpp
```diff
@@ -195,10 +195,10 @@ BridgeConnection::outgoingSuccess(ConnectionPtr outgoing)
     //
     // Flush any queued dispatches
     //
-    for (const auto& p : _queue)
+    for (auto& dispatch : _queue)
```
pepone, line 198: I assume const was removed because otherwise `std::move` would not move the members?

cpp/src/IceStorm/Service.cpp
```diff
@@ -248,7 +248,7 @@ ServiceI::start(const string& serviceName, const CommunicatorPtr& communicator,
                     throw IceBox::FailureException(__FILE__, __LINE__, "IceGrid deployment is incorrect");
                 }
 
-                int nodeid = atoi(adapterId.substr(start, end - start).c_str());
+                int nodeid = static_cast<int>(strtol(adapterId.substr(start, end - start).c_str(), nullptr, 10));
```
pepone, line 251: Shouldn't this use `std::stoi` instead of doing a static_cast to int

https://en.cppreference.com/w/cpp/string/basic_string/stol

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -394,7 +394,7 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
             auto p = rec.theQoS.find("retryCount");
             if (p != rec.theQoS.end())
             {
-                retryCount = atoi(p->second.c_str());
+                retryCount = static_cast<int>(strtol(p->second.c_str(), nullptr, 10));
```
pepone, line 397: ditto

pepone (approved):

cpp/src/IceBridge/IceBridge.cpp
bernardnormier: Yes.

cpp/src/IceStorm/Service.cpp
bernardnormier: Yes, you're right, fixed.

PR #3410, opened 2025-01-23
Makefiles cleanup

pepone:
This is an initial cleanup of our Makefiles

### DIFF
--- a/config/Make.project.rules
+++ b/config/Make.project.rules
@@ -1,20 +1,35 @@
 # Copyright (c) ZeroC, Inc.
 
+# $(call project,[$1=<project path>])
 #
-# $(call project,[$1])
+# Returns the project name by removing the lang-specific directory prefix.
+# For example, ./src/Ice becomes src/Ice, and ./test/Ice/ami becomes test/Ice/ami.
 #
-# Returns the project name, $(top_srcdir) is removed from the begining
-# of the path so for example ./src/Ice is transformed to src/Ice.
-#
+# $1 is the project path. If $1 is not provided, the current directory is used.
 project = $(patsubst $(lang_srcdir)/%,%,$(if $1,$1,$(currentdir)))
 
-#
 # The platforms, configs and languages to build.
-#
-platforms               = $(filter-out $(SKIP),$(if $(filter all,$(PLATFORMS)),$(supported-platforms),$(filter $(supported-platforms),$(PLATFORMS))))
-configs                 = $(filter-out $(SKIP),$(if $(filter all,$(CONFIGS)),$(supported-configs),$(filter $(supported-configs),$(CONFIGS))))
-languages               = $(filter-out $(SKIP),$(if $(filter all,$(or $(LANGUAGES),all)),$(supported-languages),$(filter $(supported-languages),$(LANGUAGES))))
-build-platform          := $(or $(build-platform),$(firstword $(supported-platforms)))
+
+# Define the platforms to build, excluding those listed in SKIP.
+# If PLATFORMS is set to 'all', include all supported platforms from supported-platforms.
+# Otherwise, include only the platforms in PLATFORMS that are also in supported-platforms.
+platforms = $(filter-out $(SKIP),\
+	$(if $(filter all,$(PLATFORMS)),$(supported-platforms),$(filter $(supported-platforms),$(PLATFORMS))))
+
+# Define the configurations to build, excluding those listed in SKIP.
+# If CONFIGS is set to 'all', include all supported configurations from supported-configs.
+# Otherwise, include only the configurations in CONFIGS that are also in supported-configs.
+configs = $(filter-out $(SKIP), \
+	$(if $(filter all,$(CONFIGS)),$(supported-configs),$(filter $(supported-configs),$(CONFIGS))))
+
+# Define the languages to build, excluding those listed in SKIP.
+# If LANGUAGES is set to 'all', include all supported languages from supported-languages.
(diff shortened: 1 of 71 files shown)


config/Make.project.rules
```diff
@@ -1,20 +1,35 @@
...
 project = $(patsubst $(lang_srcdir)/%,%,$(if $1,$1,$(currentdir)))
 
-#
 # The platforms, configs and languages to build.
-#
-platforms               = $(filter-out $(SKIP),$(if $(filter all,$(PLATFORMS)),$(supported-platforms),$(filter $(supported-platforms),$(PLATFORMS))))
-configs                 = $(filter-out $(SKIP),$(if $(filter all,$(CONFIGS)),$(supported-configs),$(filter $(supported-configs),$(CONFIGS))))
-languages               = $(filter-out $(SKIP),$(if $(filter all,$(or $(LANGUAGES),all)),$(supported-languages),$(filter $(supported-languages),$(LANGUAGES))))
-build-platform          := $(or $(build-platform),$(firstword $(supported-platforms)))
+
```
pepone, line 12: Added some comments to these macros to make clear what is the intend.

config/Make.rules.Darwin
```diff
@@ -55,15 +57,6 @@ endif
...
 
-clang_version = $(shell clang -dumpversion 2>&1 | cut -f1 -d\.)
-
-# We use -std=c++20 with clang 15.0 and later.
-ifeq ($(shell test $(clang_version) -ge 15 && echo yes),yes)
-    cppflags    += -std=c++20
-else
-    cppflags    += -std=c++17
-endif
-
```
pepone, line 66: Don't need to care about these older clang versions.

config/Make.rules.Linux
```diff
@@ -186,23 +166,8 @@ mkshlibname     = lib$(1).so
...
-   libbacktrace = yes
-endif
-endif
-
-IceUtil_system_libs                             = -lrt $(if $(filter yes,$(libbacktrace)),-lbacktrace)
-Ice_system_libs                                 = -ldl -lssl -lcrypto $(IceUtil_system_libs)
-ifeq ($(shell pkg-config --exists libsystemd 2> /dev/null && echo yes),yes)
-Ice_system_libs                                 += $(shell pkg-config --libs libsystemd)
-endif
+IceUtil_system_libs                             = -lrt -lbacktrace
```
pepone, line 169: lib backtrace and lib systemd should be available in all supported platforms

config/Make.tests.rules
```diff
@@ -1,12 +1,5 @@
 # Copyright (c) ZeroC, Inc.
 
-#
```
pepone, line 3: This is just the same as project macro, don't to have both.

config/Make.rules.Linux
```diff
@@ -186,35 +164,15 @@ mkshlibname     = lib$(1).so
...
+Ice_system_libs                                 = -ldl -lssl -lcrypto $(shell pkg-config --libs libsystemd)
```
bernardnormier, line 168: Missing IceUtil_system_libs

bernardnormier (approved):

config/Make.rules.Linux
```diff
@@ -1,31 +1,17 @@
...
-$(build-platform)_libtool      = $(AR)
-$(build-platform)_targetdir    = $(if $(filter %/build,$5),/$(build-platform))
-$(build-platform)_installdir   =
-$(build-platform)_objdir       =
+# Assign the linux_id to the first known distribution in the linux_ids list, or for unknown distributions to the first
+# element in the linux_ids list.
+linux_id                        ?= $(firstword $(or $(filter $(known_distributions),$(linux_ids)),$(linux_ids)))
 
-# system install directory
-system-install-dir              := $(OECORE_TARGET_SYSROOT)/usr
```
externl, line 26: Are these no longer needed?

config/Make.rules.Linux
```diff
@@ -186,35 +164,15 @@ mkshlibname     = lib$(1).so
...
-endif
-endif
-
-IceUtil_system_libs                             = -lrt $(if $(filter yes,$(libbacktrace)),-lbacktrace)
-Ice_system_libs                                 = -ldl -lssl -lcrypto $(IceUtil_system_libs)
-ifeq ($(shell pkg-config --exists libsystemd 2> /dev/null && echo yes),yes)
-Ice_system_libs                                 += $(shell pkg-config --libs libsystemd)
-endif
+IceUtil_system_libs                             = -lrt -lbacktrace
+Ice_system_libs                                 = -ldl -lssl -lcrypto $(IceUtil_system_libs) $(shell pkg-config --libs libsystemd)
```
externl, line 168: what happens if there's no systemd, does `$(shell pkg-config --libs libsystemd)` return empty?

config/Make.rules.Linux
pepone: This was for Yocto, if we ever want to add back support for it we can bring it back. But so far we are not testing/supporting this with 3.8.

config/Make.rules.Linux
pepone: Do you think we should keep the check for libsystemd installed, or just emit an error if missing?

externl: Keep the check. For instance it's not going to be available in Docker builds.

pepone: Fixed as suggested

externl: Thanks!

externl (approved):

PR #3412, opened 2025-01-23
Fix clang-tidy lints in IceGrid

bernardnormier:


externl (approved):
Preferably we wouldn't have all of these `// NOLINT(performance-unnecessary-value-param)` but it's probably more of a chore to update all of them.

pepone (approved):

PR #3414, opened 2025-01-23
Fix clang-tidy lints in Slice compilers

bernardnormier:


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2177,7 +2177,7 @@ Slice::Python::CodeVisitor::writeDocstring(const OperationPtr& op, DocstringMode
     //
     // Emit the general description.
     //
-    _out << nl << "\"\"\"";
+    _out << nl << R"(""")";
```
externl, line 2180: why not use `tripleQuotes` here and the spot below

externl (approved):

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -1185,7 +1184,7 @@ Slice::JavaGenerator::writeDictionaryMarshalUnmarshalCode(
         stream = marshal ? "ostr" : "istr";
     }
 
-    string v = param;
+    const string& v = param;
```
pepone, line 1187: Not clear why define v. can't we just use param?

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -1327,7 +1326,7 @@ Slice::JavaGenerator::writeSequenceMarshalUnmarshalCode(
     }
 
     string typeS = typeToString(seq, TypeModeIn, package);
-    string v = param;
+    const string& v = param;
```
pepone, line 1329: ditto.

pepone (approved):

cpp/src/slice2py/PythonUtil.cpp
bernardnormier: Fixes (2 spots)

PR #3419, opened 2025-01-24
Review iOS build excludes

### LINKED ISSUE #1984: Review what we exclude from iOS builds
We currently exclude all Tcp* from iOS builds of Ice C++ (direct build ands builds for Ice Swift via the Rakefile).

It's surprising that we don't exclude more. For example, we should definitely exclude Service.cpp, and CtrCHandler.cpp. It's likely we should exclude much more, like the ConsoleUtil code.

Note that none of this code is "ifdefed out" for iOS, meaning that we currently include it in the iOS builds even though it's meaningless.

pepone:
fix #1984

### DIFF
--- a/cpp/src/Ice/Makefile.mk
+++ b/cpp/src/Ice/Makefile.mk
@@ -16,8 +16,6 @@ endif
 
 Ice_excludes            = src/Ice/DLLMain.cpp
 Ice[shared]_excludes    = src/Ice/RegisterPluginsInit_min.cpp
-Ice[xcodesdk]_excludes  = src/Ice/RegisterPluginsInit_min.cpp
-Ice[static]_excludes    = src/Ice/RegisterPluginsInit_all.cpp
 
 ifeq ($(os),Linux)
 ifeq ($(shell pkg-config --exists libsystemd 2> /dev/null && echo yes),yes)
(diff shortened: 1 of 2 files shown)


externl (approved):

cpp/src/Ice/Makefile.mk
```diff
@@ -16,18 +16,27 @@ endif
...
-Ice[iphonesimulator]_extra_sources      = $(Ice[iphoneos]_extra_sources)
+ios_extrasources :=  $(wildcard $(addprefix $(currentdir)/ios/,*.cpp *.mm))
+ios_excludes := $(wildcard $(addprefix $(currentdir)/,\
+	CtrlCHandler.cpp \
+	OutputUtil.cpp \
+	RegisterPluginsInit_all.cpp \
+	Service.cpp \
+	SysLoggerI.cpp \
+	SystemdJournalI.cpp \
+	Tcp*.cpp))
```
bernardnormier, line 34: Should we also exclude UDP? WS?

bernardnormier (approved):

cpp/src/Ice/Makefile.mk
pepone: I don't think so. We support WS and UDP, we exclude TCP because there is an iOS implementation in  `ios/StreamXxx`

PR #3421, opened 2025-01-24
iOS test fixes

pepone:


.github/workflows/ci.yml
```diff
@@ -59,12 +59,11 @@ jobs:
...
-          # Swift iOS
+          # iOS
           - os: macos-15
-            config: "swift-ios"
-            working_directory: "swift"
-            build_flags: "PLATFORMS='macosx iphonesimulator'"
-            test_flags: "--platform=iphonesimulator --controller-app --rfilter=Ice/operations" # Remove rfilter once #3286 is fixed
+            config: "ios"
+            build_flags: "PLATFORMS='macosx iphonesimulator' LANGUAGES='cpp swift' CONFIGS='static'"
+            test_flags: "--platform=iphonesimulator --controller-app --rfilter=swift/Ice/operations" # Remove rfilter once #3286 is fixed
```
externl, line 66: Do we need to do `--filter=cpp` `--filter=swift`

externl (approved):

.github/workflows/ci.yml
pepone: I think we don't as there no other languages that support `iphonesimulator` platform.

pepone: Added `--languages='cpp,swift'`

externl (approved):

PR #3422, opened 2025-01-24
Fix lints in IceBT

bernardnormier:


cpp/src/IceBT/Engine.cpp
```diff
@@ -15,6 +15,8 @@ using namespace IceBT;
 
 namespace IceBT
 {
+    // foo
+
```
externl, line 19:
```suggestion
```

externl (approved):
Looks good to me. It would be nice to figure out if there's some why for us to do CI with BT.

cpp/src/IceBT/Engine.cpp
```diff
@@ -15,6 +15,8 @@ using namespace IceBT;
 
 namespace IceBT
 {
+    // foo
```
pepone: leftover?

pepone (approved):

cpp/src/IceBT/Engine.cpp
bernardnormier: removed

PR #3424, opened 2025-01-24
Remove Java Ice/translator

### LINKED ISSUE #2020: Java's `Ice/translator` Tests are not Run (and are out of date)
These tests seem to be checking the java mapping, with an emphasis on the `module`->`package` mapping.

But, these tests are clearly never run.
They aren't in the gradle file: https://github.com/zeroc-ice/ice/blob/main/java/test/slice.gradle

And the actual test files are littered with old Slice constructs: https://github.com/zeroc-ice/ice/blob/8f6f3e2c531fe078002a2d77b4d072df801d2471/java/test/Ice/translator/Underscore.ice#L12-L17 And classes implementing interfaces: https://github.com/zeroc-ice/ice/blob/8f6f3e2c531fe078002a2d77b4d072df801d2471/java/test/Ice/translator/SingleModuleWithPackage.ice#L43-L45 And other things that were removed a while ago.

----

We should either:
1) Update these tests, and start actually running them, or
2) Remove them

pepone:
I removing these build only tests.

fix #2020


externl (approved):

InsertCreativityHere (approved):
Good to see these outdated tests disappear.

But, we should probably be testing it's behavior somewhere else.
Not quite as 'thoroughly' as all of these, since I think they're overkill, but still.

PR #3425, opened 2025-01-24
Fix C++ generated code

bernardnormier:
This PR updates slice2cpp as follows:
 - the generated code for dispatch is now using a std::array instead of a C array
 - we no longer unqualify fixed Ice names such as Ice::Context
 - we no longer put :: in front of fixed Ice names such as Ice::Context and IceInternal::xxx.


pepone (approved):
Looks good. Should we also remove the global `::` for std?

bernardnormier:
> Looks good. Should we also remove the global `::` for std?

I think we should (in a separate PR). If you use Slice module "std", you should remap it in C++ to some other name.

externl (approved):

Issue #3154, opened 2024-11-15
Cannot run collocated tests with IPv6


pepone:
When using IPv6 collocated test doesn't work. The issue is that the proxy endpoint doesn't match the published endpoint

The reference uses the `0:0:0:0:0:0:0:1` from the default host set by the scripts, and the published endpoint uses `::1`

This affects C++ and C#, Java works fine with  `0:0:0:0:0:0:0:1` 
 
```
(/home/jose/Documents/3.8/ice/cpp/test/Ice/adapterDeactivation/build/x86_64-linux-gnu/shared/collocated --Ice.Default.Host=0:0:0:0:0:0:0:1 --Ice.Warn.Connections=1 --Ice.Default.Protocol=tcp --Ice.IPv6=1 --Ice.PreferIPv6Address=1 --Ice.PrintStackTraces=1 env={'LD_LIBRARY_PATH': '/home/jose/Documents/3.8/ice/cpp/lib/x86_64-linux-gnu'})
creating/destroying/recreating object adapter... ok
creating/activating/deactivating object adapter in one operation... 
error: src/Ice/ConnectionI.cpp:2387 ::Ice::ConnectTimeoutException connection establishment timed out
stack trace:
  0 Ice::Exception::Exception(char const*, int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) in /home/jose/Documents/3.8/ice/cpp/test/Ice/adapterDeactivation/build/x86_64-linux-gnu/shared/../../../../../../lib/x86_64-linux-gnu/libIce.so.38a0
  1 /home/jose/Documents/3.8/ice/cpp/test/Ice/adapterDeactivation/build/x86_64-linux-gnu/shared/../../../../../../lib/x86_64-linux-gnu/libIce.so.38a0(+0x126ab8) [0x7661c3f26ab8]
  2 /home/jose/Documents/3.8/ice/cpp/test/Ice/adapterDeactivation/build/x86_64-linux-gnu/shared/../../../../../../lib/x86_64-linux-gnu/libIce.so.38a0(+0x14100a) [0x7661c3f4100a]
  3 IceInternal::Timer::run() in /home/jose/Documents/3.8/ice/cpp/test/Ice/adapterDeactivation/build/x86_64-linux-gnu/shared/../../../../../../lib/x86_64-linux-gnu/libIce.so.38a0
[...]

PR #3428, opened 2025-01-25
Normalize IPv6 used in Endpoint.equivalent

pepone:
C# and C++, Java was already done see #3154.

I don't think we need to update JavaScript, 
- This fix is required for collocated calls which are not supported in JavaScript
- We don't do any IP address manipulation in JavaScript just pass the address directly into the WebSocket.

### DIFF
--- a/cpp/src/Ice/Network.cpp
+++ b/cpp/src/Ice/Network.cpp
@@ -1610,6 +1610,26 @@ IceInternal::getNumericAddress(const std::string& address)
     }
 }
 
+string IceInternal::normalizeIPv6Address(string_view host)
+{
+    if (host.find(':') != string::npos)
+    {
+        struct in6_addr result;
+        if (inet_pton(AF_INET6, host.data(), &result) == 1)
+        {
+            // Normalize the address
+            char buf[INET6_ADDRSTRLEN];
+            if (inet_ntop(AF_INET6, &result, buf, sizeof(buf)) != nullptr)
+            {
+                return string{buf};
+            }
+            // else conversion to string failed, keep host as is
+        }
+        // else it's not a valid IPv6 address keep host as is
+    }
+    return string{host};
+}
+
 SyscallException::ErrorCode
 IceInternal::getSocketErrno()
 {
--- a/cpp/src/Ice/Network.cpp
+++ b/cpp/src/Ice/Network.cpp
@@ -1610,7 +1610,8 @@ IceInternal::getNumericAddress(const std::string& address)
     }
 }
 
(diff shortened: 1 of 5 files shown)


csharp/src/Ice/Internal/IPEndpointI.cs
```diff
@@ -327,6 +331,23 @@ protected override bool checkOption(string option, string argument, string endpo
...
+            try
+            {
+                return IPAddress.Parse(host).ToString(); // normalized host
+            }
+            catch
+            {
+                // Ignore - don't normalize host.
+            }
+        }
+        return "";
```
pepone, line 348:
```suggestion
        return host;
```

csharp/src/Ice/Internal/IPEndpointI.cs
```diff
@@ -117,7 +120,7 @@ public override bool equivalent(EndpointI endpoint)
         }
         IPEndpointI ipEndpointI = (IPEndpointI)endpoint;
         return ipEndpointI.type() == type() &&
-            ipEndpointI.host_.Equals(host_, StringComparison.Ordinal) &&
+            ipEndpointI._normalizedHost.Equals(_normalizedHost, StringComparison.Ordinal) &&
```
bernardnormier, line 123: I would use operator== to compare these strings.

See https://learn.microsoft.com/en-us/dotnet/csharp/how-to/compare-strings.

bernardnormier (approved):

csharp/src/Ice/Internal/IPEndpointI.cs
pepone: > Make the intent of your code clear by calling an overload that explicitly specifies the type of comparison to perform.

Isn't that what we do here? Afaik there is a linter rule for this.

bernardnormier: See for example https://github.com/icerpc/icerpc-csharp/blob/fc251d6252e0135d9312dbac6b46e3dae3b23b49/src/IceRpc/ServerAddress.cs#L179

In general, we should use `==` for string comparison in C#, not Equals.

cpp/src/Ice/Network.cpp
```diff
@@ -1610,6 +1610,27 @@ IceInternal::getNumericAddress(const std::string& address)
     }
 }
 
+string
+IceInternal::normalizeIPv6Address(string_view host)
```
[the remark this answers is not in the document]
pepone: Fixed as suggested.

PR #3429, opened 2025-01-26
Add wait(CtrlCHandler&)

bernardnormier:
This PR adds a new Ice::wait(CtrlcHandler&) function and updates the doc-comments for CtrlCHandler.

With the function, a typical C++ server is as follows - we don't need `communicator->waitForShutdown`:

```cpp
int
main(int argc, char* argv[])
{
    // CtrlCHandler handles Ctrl+C and similar signals. It must be created at the beginning of the program, before
    // starting any thread.
    Ice::CtrlCHandler ctrlCHandler;

    const Ice::CommunicatorHolder communicatorHolder{argc, argv};
    const Ice::CommunicatorPtr& communicator = communicatorHolder.communicator();

    auto adapter = communicator->createObjectAdapterWithEndpoints("GreeterAdapter", "tcp -p 4061");
    adapter->add(make_shared<GreeterServer::Chatbot>(), Ice::stringToIdentity("greeter"));
    adapter->activate();

    // Wait until the application presses Ctrl+C.
    wait(ctrlCHandler);
    cout << "done" << endl;

    return 0;
}
```

### DIFF
--- a/cpp/include/Ice/CtrlCHandler.h
+++ b/cpp/include/Ice/CtrlCHandler.h
@@ -9,59 +9,49 @@
 
 namespace Ice
 {
-    /**
-     * Invoked when a signal occurs. The callback must not raise exceptions.
-     * On Linux and macOS, the callback is NOT a signal handler and can call
-     * functions that are not async-signal safe.
-     * @param sig The signal number that occurred.
-     */
+    /// The function called by CtrlCHandler when it catches a signal. This function must not throw exceptions. On Linux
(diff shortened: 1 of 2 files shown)


pepone (approved):

externl (commented):
Why not just have the wait function on `CtrlCHandler`. `handler.wait()`, like `std::condition_variable`'s `wait()`. 

I'm not sure I see the benefit of having a non-member function in this case as there's no abstraction of any kind.

externl (approved):
Looks good!

PR #3432, opened 2025-01-27
Remove local interfaces = output for TCP transport

### LINKED ISSUE #3430: local interfaces Network trace
When I start a C++ server that listens on `::0` with tcp, I get:
```
-- 01/26/25 14:51:08.853 ./server: Network: listening for tcp connections
   local address = :::4061
   local interfaces = 127.0.0.1, ::1, 192.168.1.93
-- 01/26/25 14:51:08.854 ./server: Network: published endpoints for object adapter 'GreeterAdapter':
   tcp -h studio.localdomain -p 4061 -t 60000
```

That's an Ice.Trace.Network=1 trace.

The local interfaces trace is puzzling. I suspect it's a leftover from the old default for published endpoints.

We should also print the local address (port 4061 on the IPv6 wildcard address) as `::0:0`.

bernardnormier:
and partially for the UDP transport.

Fixes #3430.

### DIFF
--- a/cpp/src/Ice/UdpTransceiver.cpp
+++ b/cpp/src/Ice/UdpTransceiver.cpp
@@ -490,19 +490,14 @@ IceInternal::UdpTransceiver::toDetailedString() const
 {
     ostringstream os;
     os << toString();
-    vector<string> intfs;
     if (isAddressValid(_mcastAddr))
     {
-        intfs = getInterfacesForMulticast(_mcastInterface, getProtocolSupport(_mcastAddr));
-    }
-    else
-    {
-        intfs = getHostsForEndpointExpand(inetAddrToString(_addr), _instance->protocolSupport(), true);
-    }
-    if (!intfs.empty())
-    {
-        os << "\nlocal interfaces = ";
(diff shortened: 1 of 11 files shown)


cpp/src/Ice/UdpTransceiver.cpp
```diff
@@ -490,19 +490,14 @@ IceInternal::UdpTransceiver::toDetailedString() const
...
-    }
-    else
-    {
-        intfs = getHostsForEndpointExpand(inetAddrToString(_addr), _instance->protocolSupport(), true);
-    }
-    if (!intfs.empty())
-    {
-        os << "\nlocal interfaces = ";
-        os << IceInternal::joinString(intfs, ", ");
+        vector<string> intfs = getInterfacesForMulticast(_mcastInterface, getProtocolSupport(_mcastAddr));
```
bernardnormier, line 495: I kept the similar UDP multicast logic/output.

getInterfacesForMulticast is called by setMcastGroup, which is itself called by UdpTransceiver::bind.

pepone (approved):

externl (approved):

PR #3434, opened 2025-01-27
Improve Doc-Comment Link Generation Machinery

InsertCreativityHere:
The main change in this PR is that the callback used to map `@link` tags has been given more context to work with.
It used to take: `(string identifier, string member)` (parsed from `@link identifier#member`).
Now it takes: `(string rawLink, ContainedPtr source, SyntaxTreeBasePtr target)`.
Having pointers to the elements lets us do things like check for `cpp:identifier`.

This PR also removes our use of `DocCommentPtr`.
There was absolutely no reason to wrap `DocComment` in a `shared_ptr`, since it was never shared between anything.
Now we use `optional<DocComment>` (`nullopt` indicates there was no doc-comment on something).

And the function for parsing a doc-comment was changed from `Contained::parseDocComment` to `DocComment::parseFrom`.
This function was setting private fields on `DocComment`, so it makes sense to have it on `DocComment` and avoid a
`friend Contained`. This was also necessary to avoid some weird `shared_from_this` hackery.


cpp/src/ice2slice/Gen.cpp
```diff
@@ -275,28 +275,27 @@ namespace
         }
     }
 
-    string slice2LinkFormatter(const string& identifier, const string& memberComponent)
+    string slice2LinkFormatter(string rawLink, const ContainedPtr&, const SyntaxTreeBasePtr&)
     {
-        // Replace links of the form `{@link Type#member}` with `{@link Type::member}`.
-        string result = "{@link ";
-        if (memberComponent.empty())
+        // The only difference with '@link' between Slice1 and Slice2 is that Slice1 uses '#' and Slice2 uses '::'.
```
bernardnormier, line 280: I assume you mean the `.ice` and `.slice` syntax, not Slice1 and Slice2. Slice1 and Slice2 are two modes of the .slice syntax.

bernardnormier (commented):
I anticipate a number of clang-tidy lints.

InsertCreativityHere:
> I anticipate a number of clang-tidy lints.

Yes, I tried to keep up with all the lint-changes that were added, but I didn't read every PR in totality.
So, there's probably some that I'm unaware of.

I expect it to complain about passing my `std::function` by `const &`.
But given that I'm passing in a function pointer (and not a lambda), I really don't see the point of passing a reference.
So, I'm hoping it won't complain about this.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -267,17 +267,45 @@ namespace
...
+
+                string parentName = getUnqualified(parent->mappedScoped(), source->mappedScope());
+                return parentName + "#" + dataMemberTarget->mappedName();
+            }
+            if (auto operationTarget = dynamic_pointer_cast<Operation>(target))
+            {
+                // Doxygen supports multiple syntaxes for operations, but none of them allow for a bare operation name.
+                // We opt for the syntax where operation names are qualified by what type they're defined on.
+                // See: https://www.doxygen.nl/manual/autolink.html#linkfunc.
+                // We also need to make sure yo include the 'Async' suffix for 'amd' operations.
```
pepone, line 289:
```suggestion
                // We also need to make sure to include the 'Async' suffix for 'amd' operations.
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1994,7 +2002,7 @@ Slice::Python::CodeVisitor::collectExceptionMembers(const ExceptionPtr& p, Membe
 }
 
 void
-Slice::Python::CodeVisitor::writeDocstring(const DocCommentPtr& comment, const string& prefix)
+Slice::Python::CodeVisitor::writeDocstring(optional<DocComment> comment, const string& prefix)
```
pepone, line 2005: I think these methods should take a reference, as they are not taking ownership of the comment.

pepone (approved):

cpp/src/ice2slice/Gen.cpp
pepone: We can also say `Ice Slice files (.ice) and IceRPC Slice files (.slice)` but is kind of verbose.

bernardnormier: This is a comment in the ice2slice implementation so you could write:
> The only difference with '@link' between the ice and slice syntaxes is that the ice syntax uses '#' ...

cpp/src/slice2py/PythonUtil.cpp
InsertCreativityHere: I changed it to use a `const&`.
I don't think it really matters though since we were always passing in xvalues to this function.

cpp/src/ice2slice/Gen.cpp
InsertCreativityHere: I went with 'Ice syntax' and 'Slice syntax' because I don't care enough to argue.
Of course, I'm still going to make my case though : vP

Because these terms feel pretty weird to me.
There is no such thing as 'Ice syntax' and 'Slice syntax' is ambiguous IMO.
Whereas 'Slice1' and 'Slice2' make it clear that we're only talking about the Slice syntax, but just two different versions of it.

It happens to be true that 'Slice1' and 'Slice2' are compilation modes in IceRPC, but that has no bearing here.
Ice has no compilation modes. And it's worth noting that the entire point of 'Slice1's existence is to interop with the 'Ice syntax'. So calling this 'Slice1' feels very natural to me. </case>

bernardnormier (approved):

PR #3445, opened 2025-01-29
Add Support for `cs:identifier` Metadata

InsertCreativityHere:
This PR adds support for `cs:identifier` in the same was as `cpp:identifier`.
And removes all the logic for escaping keywords/name collisions.


cpp/src/slice2cs/Gen.cpp
```diff
@@ -605,69 +601,13 @@ Slice::CsVisitor::emitNonBrowsableAttribute()
          << "[global::System.ComponentModel.EditorBrowsable(global::System.ComponentModel.EditorBrowsableState.Never)]";
 }
 
-string
```
InsertCreativityHere, line 608: Uncalled function.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -800,7 +739,8 @@ Slice::CsVisitor::writeParameterDocComments(const DocCommentPtr& comment, const
         auto q = commentParameters.find(param->name());
         if (q != commentParameters.end())
         {
-            _out << nl << "/// <param name=\"" << fixId(param->name()) << "\">";
+            string fixedParamName = addPrefixToIdentifier("", param->mappedName()); // Strip off any leading '@'s.
+            _out << nl << "/// <param name=\"" << fixedParamName << "\">";
```
InsertCreativityHere, line 743: I checked and doc-comment links to parameters _cannot_ start with `@`.
The link should just be to the actual identifier, without any escaping prefixes.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -228,20 +173,20 @@ Slice::CsGenerator::typeToString(const TypePtr& type, const string& package, boo
         "float",
         "double",
         "string",
-        "Ice.Object", // not used anymore
-        "Ice.ObjectPrx?",
-        "Ice.Value?"};
+        "global::Ice.Object", // not used anymore
```
pepone, line 176: I thought we decided to not use `global::` for Ice.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -298,7 +243,8 @@ string
 Slice::CsGenerator::resultStructName(const string& className, const string& opName, bool marshaledResult)
 {
     ostringstream s;
-    s << className << "_" << IceInternal::toUpper(opName.substr(0, 1)) << opName.substr(1)
+    string fixedOpName = addPrefixToIdentifier("", opName); // Strip any leading '@' from the name.
```
pepone, line 246: I don't like using empty prefix as a mean of removing `@`

cpp/src/slice2cs/Gen.cpp
```diff
@@ -1874,7 +1762,7 @@ Slice::Gen::ProxyVisitor::visitOperation(const OperationPtr& p)
             true);
         emitObsoleteAttribute(p, _out);
         _out << nl << taskResultType(p, ns);
-        _out << " " << p->name() << "Async" << spar << inParams
+        _out << " " << name << "Async" << spar << inParams
```
pepone, line 1765: Don't we nee to remove `@` if present in name?

cpp/src/slice2cs/Gen.cpp
```diff
@@ -2472,9 +2358,11 @@ void
 Slice::Gen::HelperVisitor::visitSequence(const SequencePtr& p)
 {
     string ns = getNamespace(p);
+    string name = p->mappedName();
+
     string typeS = typeToString(p, ns);
     _out << sp;
-    _out << nl << "public sealed class " << p->name() << "Helper";
+    _out << nl << "public sealed class " << name << "Helper";
```
pepone, line 2365: remove `@` from name?

cpp/src/slice2cs/CsUtil.cpp
bernardnormier: I agree!

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -142,44 +99,32 @@ Slice::CsGenerator::getUnqualified(
...
-// If the passed name is a scoped name, return the identical scoped name,
-// but with all components that are C# keywords replaced by
-// their "@"-prefixed version; otherwise, if the passed name is
-// not scoped, but a C# keyword, return the "@"-prefixed name;
-// otherwise, check if the name is one of the method names of baseTypes;
-// if so, prefix it with ice_; otherwise, return the name unchanged.
-//
 string
-Slice::CsGenerator::fixId(const string& name, unsigned int baseTypes, bool mangleCasts)
+Slice::CsGenerator::getMappedScoped(const ContainedPtr& p)
```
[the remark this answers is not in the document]
bernardnormier: I agree.

The mappedScope and mappedScoped in the Parser don't make much sense since they use `::`.

The scope is not `::` in all languages and we don't have a mechanism to pass the language scope separator from the Slice compiler to the Parser.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -142,44 +99,32 @@ Slice::CsGenerator::getUnqualified(
...
-    vector<string> newIds;
-    newIds.reserve(ids.size());
-    for (const auto& id : ids)
-    {
-        newIds.push_back(lookupKwd(id, baseTypes));
-    }
     stringstream result;
-    for (auto j = newIds.begin(); j != newIds.end(); ++j)
+
+    string name = p->mappedScoped();
```
bernardnormier, line 107: I find this code rather confusing: it operates on a `::`-separated name with the mapped C# name at the end.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -142,44 +99,32 @@ Slice::CsGenerator::getUnqualified(
...
+
     return result.str();
 }
 
+string
+Slice::CsGenerator::addPrefixToIdentifier(const string& prefix, const string& ident)
+{
+    return prefix + (ident.find('@') == 0 ? ident.substr(1) : ident);
+}
+
```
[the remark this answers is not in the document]
bernardnormier: I would add a helper function to remove the `@`, like:

```cpp
string removeAtPrefix(const string&)
```

cpp/src/slice2cs/CsUtil.cpp
bernardnormier: Use my helper function.

bernardnormier (approved):

cpp/src/slice2cs/Gen.cpp
bernardnormier: We can but it's optional. `@` creates a syntax error only when it's in the middle of an identifier. It's ignored as prefix.

InsertCreativityHere:
I changed the signature of `mappedScope` to accept a `const string& separator` argument.

This way the languages don't need to replace all the `::` with their separator of choice.
Instead they tell the parser what they want, and let it handle that.

cpp/src/slice2cs/CsUtil.cpp
InsertCreativityHere: I removed this function.
Now you can tell the parser which scope-separator you want when you call `mappedScoped`.

cpp/src/slice2cs/Gen.cpp
InsertCreativityHere: The approach that I took was to only remove the `@` when absolutely necessary and for internal code that users won't be seeing.

I think it's desirable for `cs:identifier` to be respected as much as possible. And if you put `cs:identifier:@throw` we will use `@throw` everywhere, even if the escape character isn't strictly necessary there.

The less magical `cs:identifier` is the better IMO.

cpp/src/slice2cs/CsUtil.cpp
InsertCreativityHere: I used your helper function.
But named it `removeEscapePrefix` instead of `removeAtPrefix` because the latter sounds like it's removing something at a position, which it isn't.

cpp/src/slice2cs/CsUtil.cpp
InsertCreativityHere: Oh okay, I removed the `global::` prefixes.

cpp/src/slice2cs/CsUtil.cpp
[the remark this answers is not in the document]
InsertCreativityHere: Yes, good point.

I fixed `mappedScoped` (the parser function) to accept a `separator` argument.
Now C# can just pass "." and let the parser handle it. Much cleaner.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -85,52 +52,42 @@ Slice::CsGenerator::getNamespacePrefix(const ContainedPtr& cont)
...
+        isFirst = false;
+        result << id;
     }
+
+    string fixedScope = result.str();
     string prefix = getNamespacePrefix(cont);
     if (!prefix.empty())
     {
-        if (!scope.empty())
+        if (!fixedScope.empty())
```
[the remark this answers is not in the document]
InsertCreativityHere: Theoritically it could if you passed a top-level module into this function.
But after reviewing it, we never pass a module in here.

I added an `assert(!dynamic_pointer_cast<Module>(p))` to the top of the function and simplified the code.

bernardnormier:
> I changed the signature of `mappedScope` to accept a `const string& separator` argument.
> 
> This way the languages don't need to replace all the `::` with their separator of choice. Instead they tell the parser what they want, and let it handle that.

Should this be instead a parameter of createUnit?

The implementation queries xxx:identifier based on a fixed xxx provided to createUnit, but then requires slice2xxx to provide its scope separator every time it wants a mapped scope? This doesn't sound very consistent.

InsertCreativityHere:
1) It's not something we call often, in `slice2cs` we only call it from 4 places, so it's not a huge burden.
2) It's rare, but we don't always use the same separator. There's two places in slice2cpp where we actually want `_` as our separator because we want a 'flattened' name.

So, I'm inclined to leave it this way until we're passing it many many times and it becomes a hassle to deal with.
Which might happen as I do the other language mappings, but so far, it hasn't IMO.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -85,52 +52,18 @@ Slice::CsGenerator::getNamespacePrefix(const ContainedPtr& cont)
...
-    }
-    else if (builtin)
-    {
-        return type.find('.') == string::npos ? type : "global::" + type;
-    }
-    else
-    {
-        return "global::" + type;
-    }
+    string scope = cont->mappedScope(".").substr(1);
```
pepone, line 57: So mapped scope returns `".Foo.Bar."`, and that is why we need `substr(1)` here?

I wonder if we can have a scope prefix, and C# can pass either "" or "global::" and use it for the initial occurrence.

pepone (approved):

cpp/src/slice2cs/CsUtil.cpp
InsertCreativityHere: Yes, that's correct, and also not a bad idea.

But I want to check how often we actually use this. Because the other option is just to remove this global scope piece all together and let callers add it by hand. Because I'm pretty sure we almost always just remove it except for niche things like type-ids.

pepone: It would make the intent clear, otherwise each time you see this code have to wonder why is `substr(1)` necessary.

bernardnormier: I think we should either:
 - don't start with the scope in the string returned by mappedScope (and update mappedScope doc-comment)
 or
 - add a prefix parameter

InsertCreativityHere: I fully agree with @bernardnormier's comment.

All I'm saying is that I lean towards not starting with the scope separator, but I want to wait and see how we're actually using this. I suspect the leading separator is more of a nuisance than a help, but I won't know until I look.

PR #3446, opened 2025-01-29
Reduce number of visitors in slice2java

bernardnormier:
This PR reduces the number of visitors in slice2java to just 2.

It doesn't change at all the generated code - just reshuffling of Gen.{h,cpp}

### DIFF
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -2239,8 +2239,8 @@ Slice::Gen::generate(const UnitPtr& p)
     HelperVisitor helperVisitor(_dir);
     p->visit(&helperVisitor);
 
-    ProxyVisitor proxyVisitor(_dir);
-    p->visit(&proxyVisitor);
+    ServantVisitor servantVisitor(_dir);
+    p->visit(&servantVisitor);
 }
 
 Slice::Gen::PackageVisitor::PackageVisitor(const string& dir) : JavaVisitor(dir) {}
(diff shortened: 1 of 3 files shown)


cpp/src/slice2java/Gen.cpp
```diff
@@ -4448,3 +4326,109 @@ Slice::Gen::ProxyVisitor::visitOperation(const OperationPtr& p)
...
 }
+
+void
+Slice::Gen::TypesVisitor::emitCompactIdHelper(const ClassDefPtr& p)
+{
+    string prefix = getPackagePrefix(p);
+    if (!prefix.empty())
+    {
+        prefix = prefix + ".";
+    }
```
pepone, line 4337: we can move this inside the other if clause, is not used otherwise.

pepone (approved):

cpp/src/slice2java/Gen.cpp
```diff
@@ -2262,14 +2253,14 @@ Slice::Gen::PackageVisitor::visitModuleStart(const ModulePtr& p)
 
         close();
     }
-    return false;
+    return true;
```
InsertCreativityHere, line 2256: This causes a change in the generated code. Before, when we always returned `false`, we would only ever hit top-level modules, but now we're hitting all modules.

The `if` check at the top of this function should be changed to:
```
if (!prefix.empty() && dynamic_pointer_cast<Unit>(p->container()))
```
this way we only generate `_Marker` interfaces for top-level module, like we were previously.

InsertCreativityHere (approved):
This PR introduces a change to how we generate `_Marker` interfaces.
Otherwise, my comparison tool confirms that there are no other changes to the generated code.

cpp/src/slice2java/Gen.cpp
bernardnormier: Fixed as suggested.

cpp/src/slice2java/Gen.cpp
bernardnormier: Fixed.

Issue #3358, opened 2025-01-14
Address Cyclic `Unit` references in `Parser.h`


InsertCreativityHere:
In the Slice compiler, the base type for all Slice types & definitions is `SyntaxTreeBase`. `SyntaxTreeBase` itself inherits from the _true_ base class: `GrammarBase`. This represents other grammatical objects that aren't types or definitions (metadata, parts of definitions that haven't been fully parsed yet, and tokens for holding string-literals, integer-literals, etc.)

One of the types which inherits from `SyntaxTreeBase` is `Unit` (which represents a compilation unit, a.k.a. 1 Slice file).
But, this inheritance is useless and causes problems for us because `SyntaxTreeBase` holds a `UnitPtr _unit;`, a 'cyclic' reference.

More importantly though is that this API is useless, and doesn't make sense for `Unit` to implement:
```
virtual void destroy();                   // Only sets `_unit = nullptr;`
UnitPtr unit() const;                     // Useless on `Unit`, it already has access to itself.
DefinitionContextPtr definitionContext() const;     // Always `nullptr` for `Unit`
virtual void visit(ParserVisitor* visitor);         // Only useful function.
```

We should consider either moving these useless functions down the inheritance tree, or move `Unit` up the tree, so that it isn't implementing these useless functions.

----

If that isn't possible without great effort, we should at least fix the cyclic reference issue as suggested by @pepone:
https://github.com/zeroc-ice/ice/pull/3292#discussion_r1914524492

InsertCreativityHere:
The situation could still be improved by using `weak_ptr` to avoid cycles, but I think enough simplification has taken place to say that this was fixed by #3450 and #3496.

PR #3450, opened 2025-01-29
Move `definitionContext` up from `SyntaxTreeBase` to `Contained`

InsertCreativityHere:
`DefinitionContext` provides information about the file-scope a Slice definition is contained in.
It's mostly for checking file-metadata, but there's other niche things we do with it.

Right now it's defined on `SyntaxTreeBase`, but this is the wrong place to put it.
This includes types like `Unit` and `Builtin`. Calling `definitionContext` on these always gives `nullptr`, because they aren't "Slice definitions".

So, this PR moves this method down the inheritance tree and onto `Contained`, which is what we would consider a "Slice definition".

----

This PR is a step towards #3358.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -709,24 +709,12 @@ Slice::SyntaxTreeBase::unit() const
     return _unit;
 }
 
-DefinitionContextPtr
-Slice::SyntaxTreeBase::definitionContext() const
-{
-    return _definitionContext;
-}
-
 void
 Slice::SyntaxTreeBase::visit(ParserVisitor* /*visitor*/)
 {
 }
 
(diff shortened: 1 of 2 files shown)


cpp/src/Slice/Parser.cpp
```diff
@@ -356,24 +356,12 @@ Slice::SyntaxTreeBase::unit() const
...
 Slice::SyntaxTreeBase::visit(ParserVisitor* /*visitor*/)
 {
 }
 
-Slice::SyntaxTreeBase::SyntaxTreeBase(UnitPtr unit) : _unit(std::move(unit))
-{
-    if (_unit)
-    {
-        _definitionContext = _unit->currentDefinitionContext();
-    }
```
InsertCreativityHere, line 375: Because of the round-about way we create new instances of Unit, this block is never entered.

PR #3451, opened 2025-01-30
Add Publish target for C# builds and upgrade NuGet packages

pepone:
This PR upgrades C# NuGet packages to include slice2cs compiler in all builds. The PR also adds a new MSBuild target `Publish` to publish the NuGet package to the local `global-packages` source. 

The BUILDING.md has been updated to reflect this changes.

### DIFF
--- a/csharp/BUILDING.md
+++ b/csharp/BUILDING.md
@@ -1,8 +1,7 @@
 # Building Ice for .NET
 
-This page describes how to build Ice for .NET from source and package the
-resulting binaries. As an alternative, you can download and install the
-[zeroc.ice.net][1] NuGet package.
+This page describes how to build Ice for .NET from source and package the resulting binaries. As an alternative, you
+can download and install the [zeroc.ice.net][1] NuGet package.
 
 * [Building on Windows](#building-on-windows)
   * [Windows Build Requirements](#windows-build-requirements)
@@ -45,12 +44,6 @@ msbuild msbuild\ice.proj
 
 Upon completion, the Ice assemblies for .NET 8.0 are placed in the `lib\net8.0` folder.
 
-You can skip the build of the test suite with the `BuildDist` target:
-
-```shell
-msbuild msbuild\ice.proj /t:BuildDist
-```
-
 #### Authenticode Signatures
 
 You can sign the Ice binaries with Authenticode by setting the following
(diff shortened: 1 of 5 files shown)


csharp/BUILDING.md
```diff
@@ -45,12 +44,6 @@ msbuild msbuild\ice.proj
 
 Upon completion, the Ice assemblies for .NET 8.0 are placed in the `lib\net8.0` folder.
 
-You can skip the build of the test suite with the `BuildDist` target:
-
-```shell
-msbuild msbuild\ice.proj /t:BuildDist
-```
```
pepone, line 52: BuildDist was used before when we have separate targets for sources and tests.

csharp/msbuild/zeroc.ice.net.props
```diff
@@ -8,33 +8,6 @@
         <IceSoVersion>38</IceSoVersion>
         <IceHome>$(MSBuildThisFileDirectory)..</IceHome>
         <IceAssembliesDir>$(IceHome)\lib</IceAssembliesDir>
+        <IceToolsPath>$(IceHome)\tools</IceToolsPath>
     </PropertyGroup>
-
-    <Choose>
```
pepone, line 13: The properties below were used to find slice2cs in different platforms. With the new package we will always use the Slice compiler included in the NuGet package.

externl: In the future when we publish to NuGet Gallery, will be include binaries for Linux, macOS, and Windows?

externl (approved):
Looks good. How will this work for as a multi platform package?

pepone:
> Looks good. How will this work for as a multi platform package?

I would deal with this in a follow up PR.

csharp/BUILDING.md
bernardnormier: I'd rather remove BuildDist and not keep it as an alias for Build.

csharp/BUILDING.md
```diff
@@ -131,8 +120,13 @@ To create a NuGet package, open a command prompt and run the following command:
...
 ```
 
-This creates the `zeroc.ice.net` Nuget package in the `msbuild/zeroc.ice.net`
-directory.
+This creates the `zeroc.ice.net` Nuget package in the `msbuild/zeroc.ice.net` directory.
+
+You can publish the package to your local `global-packages` source with the following command:
+
+```shell
+dotnet msbuild msbuild\ice.proj /t:Publish
```
bernardnormier, line 128: For a follow-up PR: is there anyway we can switch to "dotnet build" and not need the `msbuild/ice.proj` argument?

csharp/BUILDING.md
pepone: Ok, I would in a follow up PR, need to fix CI builds that relay in all languages to provide a BuildDist target.

csharp/BUILDING.md
pepone: We need MSBuild for the C++ dependencies. If we build then separately it would be fine to use dotnet.

PR #3455, opened 2025-01-30
Fix modernize lints in C++ tests

bernardnormier:
This is mostly an automated fix with `-fix`.


pepone (approved):

externl (approved):
Surprisingly not too bad

PR #3460, opened 2025-01-30
Reduce visitors in slice2cs

bernardnormier:
This PR consolidates the visitors in the slice2cs implementation.

### DIFF
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -835,9 +835,6 @@ Slice::Gen::generate(const UnitPtr& p)
     ResultVisitor resultVisitor(_out);
     p->visit(&resultVisitor);
 
-    HelperVisitor helperVisitor(_out);
-    p->visit(&helperVisitor);
-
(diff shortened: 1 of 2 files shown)


cpp/src/slice2cs/Gen.cpp
```diff
@@ -1056,75 +1043,126 @@ Slice::Gen::TypesVisitor::visitClassDefEnd(const ClassDefPtr& p)
...
-    }
 
-    _out << sb;
-    return true;
+        //
+        // The sequence is a custom sequence with elements of class type.
+        // Emit a dummy class that causes a compile-time error if the
+        // custom sequence type does not implement an indexer.
+        //
+        _out << sp;
```
pepone, line 1088: I assume this was already there, but seems and odd test. If a sequence doesn't implement the required methods the code should already fail to compile.

pepone (approved):

cpp/src/slice2cs/Gen.cpp
bernardnormier: We need to review this code in a follow-up PR.

PR #3463, opened 2025-01-31
Update C# build instructions and packaging

pepone:
This PR upgrades .NET NuGet package build:

- The build is now the same on all platforms, it requires that the C# Slice compiler has been previously build.
- The Slice compilers are now copied to a `<platform>-<arch>` subdir, using the same logic we have in IceRPC.
- Renamed NuGetPack to Pack
- Reworked BuildDist target to just build the src projects


csharp/BUILDING.md
```diff
@@ -3,90 +3,48 @@
...
-cd csharp
-```
-
-To build all Ice assemblies and the associated test suite, run:
+Open a command prompt and change to the `csharp` subdirectory. To build all Ice assemblies and the associated
+test suite, run:
 
 ```shell
-msbuild msbuild\ice.proj
+dotnet msbuild msbuild\ice.proj
```
bernardnormier, line 31: We should decide whether (a) building ice/csharp builds slice2cs or (b) requires slice2cs to be previously built in ice/cpp/bin/...

If (a), then my understanding is we could / should use `dotnet build` for these instructions. If (b), then we don't need "2" in the prerequisites.

csharp/BUILDING.md
```diff
@@ -3,90 +3,48 @@
...
 ```
 
-> Depending on your Visual Studio environment, you may need to specify the platform.
-> For example:
->
-> ```shell
-> msbuild msbuild\ice.proj /p:Platform=x64
-> ```
-
 Upon completion, the Ice assemblies for .NET 8.0 are placed in the `lib\net8.0` folder.
```
bernardnormier, line 34: I'd rather use `/` when the code / instruction is not Windows-specific.

csharp/BUILDING.md
```diff
@@ -3,90 +3,48 @@
...
-
-```shell
-pip install passlib
-```
-
-To run the tests, open a command window and change to the top-level directory. At the command prompt, execute:
+To run the tests, open a command prompt and change to the `csharp` subdirectory. At the command prompt, execute:
 
 ```shell
 python allTests.py
```
bernardnormier, line 50: Should this be python3?

csharp/msbuild/zeroc.ice.net.csproj
```diff
@@ -37,24 +42,72 @@
...
+            <ItemGroup>
+                <Content Include="$(SLICE2CS_STAGING_PATH)/linux-x64/slice2cs">
+                    <PackagePath>tools/linux-x64/</PackagePath>
+                    <Pack>true</Pack>
+                </Content>
+                <Content Include="$(SLICE2CS_STAGING_PATH)/linux-arm64/slice2cs">
+                    <PackagePath>tools/linux-arm64/</PackagePath>
+                    <Pack>true</Pack>
+                </Content>
+                <Content Include="$(SLICE2CS_STAGING_PATH)/macos-x64/slice2cs">
```
bernardnormier, line 89: I would remove this macos x64 support.

bernardnormier (approved):

csharp/BUILDING.md
pepone: On Windows we were using `MSBuild` command to be able to build `slice2cs`. We cannot use `dotnet msbuild` to build C++ binaries.

We are not using `dotnet build` because our NuGet package `zeroc.ice.net` doesn't match any of the projects we build it just aggregate all the other projects in a single package.

csharp/BUILDING.md
pepone: yes

PR #3464, opened 2025-01-31
Fix remaining C++ test lints

bernardnormier:
This PR fixes the remaining lints in the C++ tests and adds the tests to the CI clang-tidy check.

I also updated the CI job to exclude the generated code from the check; otherwise, it takes too long (almost 3h).

### DIFF
--- a/.github/workflows/cpp.yml
+++ b/.github/workflows/cpp.yml
@@ -62,13 +62,13 @@ jobs:
 
       - name: Select files to check
         uses: tj-actions/glob@v22
-        working-directory: cpp
         id: glob
         with:
           files: |
             **/*.cpp
           excluded-files: |
             **/generated/*.cpp
+          working-directory: cpp
 
       - name: Run Clang Tidy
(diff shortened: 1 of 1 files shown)


cpp/test/Ice/logger/Client1.cpp
```diff
@@ -36,7 +36,7 @@ Client1::run(int argc, char** argv)
     }
     test(s.find(programName) != string::npos);
     in.close();
-    remove("log.txt");
+    (void)remove("log.txt");
```
pepone, line 39: can we use `[[maybe_unused]]`?

pepone (approved):

cpp/test/Ice/logger/Client1.cpp
bernardnormier: Yes, it can be `(void)` or `[[maybe_unused]] auto _ = `.

externl (approved):

PR #3469, opened 2025-02-01
Add CtrlCHandler helper class (Swift)

bernardnormier:
This PR adds the CtrlCHandler helper class to Swift.

It's a thin wrapper over the Ice::CtrlCHandler class provided by Ice C++. It's main consumer is our Swift demos for macOS.

### DIFF
--- a/cpp/include/Ice/CtrlCHandler.h
+++ b/cpp/include/Ice/CtrlCHandler.h
@@ -37,7 +37,7 @@ namespace Ice
 
         /// Unregisters the callback function.
         /// On Linux and macOS, this destructor joins and terminates the thread created by the constructor but does not
-        /// "unmask" SIGHUP, SIGINT and SIGTERM. As a result, these signals are ignored after this destructor completes.
+        /// unmask SIGHUP, SIGINT and SIGTERM. As a result, these signals are ignored after this destructor completes.
         /// On Windows, this destructor unregisters the SetConsoleCtrlHandler handler routine, and as a result a
         /// Ctrl+C or similar signal will terminate the application after this destructor completes.
         ~CtrlCHandler();
--- a/cpp/include/Ice/CtrlCHandler.h
+++ b/cpp/include/Ice/CtrlCHandler.h
@@ -15,25 +15,31 @@ namespace Ice
     /// \headerfile Ice/Ice.h
     using CtrlCHandlerCallback = std::function<void(int sig)>;
 
-    /// Provides a portable way to handle Ctrl-C and Ctrl-C like signals. On Linux and macOS, the CtrlCHandler handles
-    /// SIGHUP, SIGINT and SIGTERM. On Windows, it is essentially a wrapper for SetConsoleCtrlHandler.
+    /// Provides a portable way to handle Ctrl+C and Ctrl+C like signals. On Linux and macOS, the CtrlCHandler handles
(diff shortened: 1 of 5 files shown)


cpp/include/Ice/CtrlCHandler.h
```diff
@@ -15,25 +15,31 @@ namespace Ice
...
         /// On Linux and macOS, this constructor masks the SIGHUP, SIGINT and SIGTERM signals and then creates a thread
         /// that waits for these signals using sigwait. On Windows, this constructor calls SetConsoleCtrlCHandler to
         /// register a handler routine that calls the supplied callback function.
-        /// Only a single CtrlCHandler object can exist in a process at a give time.
-        /// @param cb The callback function to invoke when a signal is received. The default (nullptr) means do nothing.
-        explicit CtrlCHandler(CtrlCHandlerCallback cb = nullptr);
+        /// @param cb The callback function to invoke when a signal is received.
+        explicit CtrlCHandler(CtrlCHandlerCallback cb);
+
+        CtrlCHandler(const CtrlCHandler&) = delete;
```
bernardnormier, line 35: Not having these made the class rather bug-prone.

swift/src/IceImpl/include/CtrlCHandler.h
```diff
@@ -0,0 +1,22 @@
...
+// The implementation of Ice.CtrlCHandler, which wraps the C++ class Ice::CtrlCHandler.
+ICEIMPL_API @interface ICECtrlCHandler : NSObject
+- (void)receiveSignal:(void (^)(int))callback;
+@end
+
+#ifdef __cplusplus
+
+@interface ICECtrlCHandler ()
+{
+    Ice::CtrlCHandler cppObject;
```
bernardnormier, line 16: It's essential to use a field and not a property here, since a property wants to return a copy of the C++ object (which is not copyable).

bernardnormier:
Although not tested yet, I believe this helper class can also be used when waiting for either Ctrl+C or a shutdown request, with something like:
```swift
Task {
    await ctrlCHandler.receiveSignal()
    communicator.shutdown() // initiate shutdown after receiving signal
}
```

+ the "main" thread blocks on `communicator.waitForShutdown()`

cpp/src/Ice/CtrlCHandler.cpp
```diff
@@ -67,9 +88,9 @@ handlerRoutine(DWORD dwCtrlType)
 CtrlCHandler::CtrlCHandler(CtrlCHandlerCallback callback)
 {
     unique_lock lock(globalMutex);
-    bool handler = _handler != nullptr;
+    const bool handlerAlive = _handler != nullptr;
 
-    if (handler)
+    if (handlerAlive)
```
pepone, line 93: Why don't `if (_handler)` directly?

cpp/src/Ice/CtrlCHandler.cpp
```diff
@@ -149,9 +164,9 @@ namespace
...
```
pepone, line 169: ditto use `if (_handler)`

pepone (approved):

cpp/src/Ice/CtrlCHandler.cpp
bernardnormier: Fixed.

externl (approved):
Looks good!

PR #3478, opened 2025-02-03
Allow Lists of Metadata

### LINKED ISSUE #3459: Can't apply 2 metadata to operation
I am trying to apply two metadata to the same operation.

This works:

```slice
 /// Represents a simple greeter.
    interface Greeter
    {
        /// Creates a personalized greeting.
        /// @param name: The name of the person to greet.
        /// @returns: The greeting.
        ["amd", "cs:identifier:Greet"]
        string greet(string name);
    }
```

This fails:
```slice
    /// Represents a simple greeter.
    interface Greeter
    {
        /// Creates a personalized greeting.
        /// @param name: The name of the person to greet.
        /// @returns: The greeting.
        ["amd"] <-- that's line 13
        ["cs:identifier:Greet"]
        string greet(string name);
    }
```

```
bernard@studio Greeter % dotnet build
Restore complete (0.3s)
  Server succeeded (0.0s) → Server/bin/Debug/net8.0/Server.dll
  ServerAMD failed with 1 error(s) (0.0s)
    /Users/bernard/builds/ice-demos/csharp/Ice/Greeter/slice/GreeterAMD.ice(13): error : 
      syntax error
```

InsertCreativityHere:
This PR fixes #3459. We now support listing multiple metadata in all these ways:
```
["foo", "bar"]
["foo"] ["bar"]
["foo", "bar"] ["baz"] // even this!
```
I kept the escaping/metadata tests as-is (ie. they're still using the comma syntax),
but updated everywhere else to use this 'new' syntax.
In my opinion, this 'new' syntax should be preferred going forward (as it's the syntax we use in IceRPC).


InsertCreativityHere:
Note for reviewers:
`Grammar.cpp` is generated code. Don't waste your time reviewing it.

cpp/src/IceGrid/Internal.ice
```diff
@@ -279,15 +279,15 @@ interface NodeSession
     ["cpp:const"] idempotent NodeObserver* getObserver();
 
     /// Ask the registry to load the servers on the node.
-    ["amd", "cpp:const"] idempotent void loadServers();
+    ["amd"] ["cpp:const"] idempotent void loadServers();
```
bernardnormier, line 282: I assume you kept some instances (tests) for the "," syntax?

bernardnormier (approved):

externl (approved):

cpp/src/IceGrid/Internal.ice
InsertCreativityHere: Yes, the `escape` tests all still use the old syntax, and so do the metadata-specific tests in `cpp/test/Slice`.
Everything else was updated.

PR #3480, opened 2025-02-03
Fix C++17 build with MSVC

### LINKED ISSUE #3476: Support std c++17 builds with Windows
Currently when building the demos against Ice using C++17 I get some errors.
```
[build] c:\Users\joe\Developer\zeroc-ice\ice\cpp\include\Ice\VersionFunctions.h(20,44): error C7555: use of designated initializers requires at least '/std:c++20' [C:\Users\joe\Developer\zeroc-ice\ice-demos\build\Ice\greeterAsync\ice_greeter_async_client.vcxproj]
[build]   (compiling source file '../../../cpp/Ice/greeterAsync/Client.cpp')
[build]   
[build] C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.42.34433\include\span(12): warning STL4038: The contents of <span> are available only with C++20 or later. [C:\Users\joe\Developer\zeroc-ice\ice-demos\build\Ice\greeterAsync\ice_greeter_async_client.vcxproj]
[build] c:\Users\joe\Developer\zeroc-ice\ice\cpp\include\Ice\VersionFunctions.h(14,44): error C7555: use of designated initializers requires at least '/std:c++20' [C:\Users\joe\Developer\zeroc-ice\ice-demos\build\Ice\greeter\ice_greeter_server.vcxproj]
[build]   (compiling source file '../../../cpp/Ice/greeter/Server.cpp')
```

bernardnormier:
This PR updates the Ice C++ headers to support building applications (e.g. the demos) in C++17 mode.

Fixes #3476.

Note: the Ice C++ source builds still use C++20 and the source code uses designated initializers (a C++20 feature that appears to work with GCC in C++17 mode). That's fine - we just Ice C++ users to be able to use C++17 mode for their applications.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -7,6 +7,7 @@ Checks:
   -modernize-avoid-c-arrays,
   -modernize-use-trailing-return-type,
   -modernize-macro-to-enum,
+  -modernize-use-designated-initializers,
   performance-*,
   -performance-avoid-endl,
   -performance-enum-size,
--- a/cpp/include/Ice/StreamHelpers.h
+++ b/cpp/include/Ice/StreamHelpers.h
@@ -12,7 +12,7 @@
 
 #if defined(_MSC_VER)
 // With MSVC, __has_include(<span>) evaluates to 1 in c++17 mode and the <span> header then issues a warning. This
-// is a permissible but impractical behavior. As a result, we include <span> only in c++20 mode.
+// is a permissible but impractical behavior. As a result, we include <span> only for c++20 or higher.
 #    if _MSVC_LANG >= 202002L
 #        include <span>
 #    endif
--- a/cpp/include/Ice/StreamHelpers.h
+++ b/cpp/include/Ice/StreamHelpers.h
@@ -10,7 +10,11 @@
 #include <iterator>
 #include <ostream>
 
-#if __has_include(<span>)
+#if defined(_MSC_VER) // workaround for MSVC bug
+#    if _MSVC_LANG >= 202002L
+#        include <span>
+#    endif
+#elif __has_include(<span>)
 #    include <span>
 #endif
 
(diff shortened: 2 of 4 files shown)


.clang-tidy
```diff
@@ -7,6 +7,7 @@ Checks:
   -modernize-avoid-c-arrays,
   -modernize-use-trailing-return-type,
   -modernize-macro-to-enum,
+  -modernize-use-designated-initializers,
```
bernardnormier, line 10: Otherwise clang-tidy complains about not using designated initializers in VersionFunctions.h.

Not clear why since we specify c++17 in the clang-tidy config file.

cpp/include/Ice/StreamHelpers.h
```diff
@@ -10,7 +10,11 @@
 #include <iterator>
 #include <ostream>
 
-#if __has_include(<span>)
+#if defined(_MSC_VER) // workaround for MSVC bug
```
externl, line 13: Is there an issue to link to or can you add a description or what this bug is.

externl (approved):

cpp/include/Ice/StreamHelpers.h
bernardnormier, line 13: MSVC in C++ 17 mode:
 - `__has_include(<span>)` succeeds (it should not since `<span>` is c++20-only)
 - then we include `<span>`
 - the `<span>` code complains it's C++20-only

bernardnormier: I could not find a bug report.

bernardnormier: This is actually a permissible behavior according to cppreference:
https://en.cppreference.com/w/cpp/preprocessor/include

> A __has_include result of 1 only means that a header or source file with the specified name exists. It does not mean that the header or source file, when included, would not cause an error or would contain anything useful. For example, on a C++ implementation that supports both C++14 and C++17 modes (and provides __has_include in its C++14 mode as a conforming extension), __has_include(<optional>) may be 1 in C++14 mode, but actually #include <optional> may cause an error.

Updated comment accordingly.

PR #3481, opened 2025-02-03
Add Workflow to build Windows MSI installer

pepone:


.github/workflows/build-msi-package.yml
```diff
@@ -0,0 +1,32 @@
...
+  workflow_dispatch:
+
+jobs:
+    build-msi-package:
+      runs-on: windows-latest
+      steps:
+        - name: Checkout repository
+          uses: actions/checkout@v4
+
+        - name: Build C++ NuGet Packages
```
externl, line 14: Don't we need to build c++  for the slice compilers and services?

externl, line 14: Also slice compilers and services right?

externl (approved):

.github/workflows/build-msi-package.yml
pepone: Yes. The NuGet packages include all C++ binaries.

PR #3482, opened 2025-02-03
Resolve Various TODOs in the Slice Compilers

InsertCreativityHere:


cpp/src/Slice/Parser.cpp
```diff
@@ -11,11 +11,6 @@
 #include <iterator>
 #include <limits>
 
-// TODO: fix this warning once we no longer support VS2013 and earlier
-#if defined(_MSC_VER)
-#    pragma warning(disable : 4589) // Constructor of abstract class 'Slice::Type' ignores initializer...
-#endif
-
```
InsertCreativityHere, line 18: We aren't supporting VS2013 with 3.8.

cpp/src/slice2cpp/CPlusPlusUtil.h
```diff
@@ -58,9 +58,6 @@ namespace Slice
     void
     writeAllocateCode(IceInternal::Output&, const ParameterList&, const OperationPtr&, const std::string&, TypeContext);
 
-    // TODO: remove from header file.
-    void writeMarshalUnmarshalAllInHolder(IceInternal::Output&, const std::string&, const DataMemberList&, bool, bool);
-
```
InsertCreativityHere, line 63: Not sure why, but okay. Done.

cpp/src/slice2php/Main.cpp
```diff
@@ -810,9 +810,8 @@ CodeVisitor::visitDictionary(const DictionaryPtr& p)
     }
     else if (!dynamic_pointer_cast<Enum>(keyType))
     {
-        // TODO: using 'InvalidMetadata' as our warning category for an unsupported key type feels weird.
         // See https://github.com/zeroc-ice/ice/issues/254
-        p->unit()->warning(p->file(), p->line(), InvalidMetadata, "dictionary key type not supported in PHP");
```
InsertCreativityHere, line 815: `All` means that this warning doesn't have a category, since in order to suppress it, you have to suppress `All` warnings.

This feels more correct than lumping this in with 'invalid metadata'.

PR #3495, opened 2025-02-04
Build deb packages

pepone:
This PR includes fixes for the DEB package builder action. See builds in https://github.com/zeroc-ice/ice/actions/workflows/build-deb-packages.yml


.github/actions/build-deb-package/action.yml
```diff
@@ -18,26 +18,31 @@ inputs:
...
+        mkdir -p "$HOME/build"
         docker run --rm \
-          -v $PWD/ice:/workspace/ice \
-          -v $PWD/output:/workspace/build \
+          -v "$GITHUB_WORKSPACE:/workspace/ice" \
+          -v "$HOME/build:/workspace/build" \
           -e ICE_VERSION=${{ inputs.ice_version }} \
           -e DEB_BUILD_OPTIONS="${{ inputs.deb_build_options }}" \
-          ice-debian-package-builder
+          --entrypoint /workspace/ice/packaging/dpkg/build-package.sh \
```
pepone, line 33: We execute the build script from the mounted directory.

.github/actions/build-deb-package/action.yml
```diff
@@ -18,26 +18,31 @@ inputs:
...
+          ice-deb-package-builder
```
externl, line 34: I would remove the ENTRYPOINT entirely from the build and just call the script directly when running the container.
```suggestion
          \ ice-deb-package-builder /workspace/ice/packaging/dpkg/build-package.sh
```

Docker will pass args to the endpoint commend. e.g. try
```bash
docker run --rm -it debian:12 ls
```

.github/actions/build-deb-package/action.yml
```diff
@@ -18,26 +18,31 @@ inputs:
...
-
     - name: Build Docker Image
-      run: docker build -f ${{ inputs.dockerfile_path }} -t ${{ inputs.docker_image }} .
+      run: docker build --pull -f ${{ inputs.dockerfile_path }} -t ice-deb-package-builder .
       shell: bash
 
     - name: Run Package Build with Mounted Source
       run: |
-        mkdir -p output
+        mkdir -p "$HOME/build"
```
externl, line 27: I think it would be cleaner to clone ice into `${GITHUB_WORKSPACE}/ice` like we do for the demos and create `${GITHUB_WORKSPACE}/build` for the build to use

.github/actions/build-deb-package/action.yml
```diff
@@ -18,26 +18,30 @@ inputs:
...
         docker run --rm \
-          -v $PWD/ice:/workspace/ice \
-          -v $PWD/output:/workspace/build \
+          -v "$GITHUB_WORKSPACE:/workspace/ice" \
+          -v "$HOME/build:/workspace/build" \
           -e ICE_VERSION=${{ inputs.ice_version }} \
           -e DEB_BUILD_OPTIONS="${{ inputs.deb_build_options }}" \
-          ice-debian-package-builder
+          \ ice-deb-package-builder /workspace/ice/packaging/dpkg/build-package.sh
```
externl, line 33: I don't t think this `\` is necessary.
```suggestion
          ice-deb-package-builder /workspace/ice/packaging/dpkg/build-package.sh
```

.github/actions/build-deb-package/action.yml
pepone: Fixed as suggested

externl (approved):

PR #3496, opened 2025-02-04
`Unit` Cycle Fixes & Cleanup `SyntaxTreeBase`

InsertCreativityHere:
The classes we use to represent Slice definitions in the parser rely heavily on inheritance. The base of all these types is `SyntaxTreeBase`. This PR's focus is on moving stuff out of `SyntaxTreeBase` and into more appropriate places,
further up the inheritance tree.

Specifically:
- `SyntaxTreeBase` has a default implementation of `visit` which is no-op. This is bad.
  It means that you can call `Builtin::visit`. But what does it even mean to visit a built in type?
  This `visit` function was moved up to `Contained`, which represents a Slice definition.
  These are the only sensible things to visit.
  Also, `visit` is now pure virtual. Every type should (and does) implement it's own visit logic.

- `SyntaxTreeBase` also holds a `UnitPtr` field. This was removed.
  We had to jump through hoops to get this to work for `Unit`, since this means it holds a `shared_ptr` to itself... Also, it made us unable to provide a default constructor for `SyntaxTreeBase`.
  So. This field was removed and replaced with a pure virtual `unit()` function.
  `Contained` holds a `UnitPtr` field, and returns it in it's implementation.
  Whereas `Unit` implements this function with `shared_from_this`.

- By slimming down `SyntaxTreeBase` in these ways, I was able to completely eliminate it's constructor, along with the constructors for `Type` and `Container`.

----

A result of this is that #3358 is fixed.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -11,11 +11,6 @@
 #include <iterator>
 #include <limits>
 
-// TODO: fix this warning once we no longer support VS2013 and earlier
-#if defined(_MSC_VER)
-#    pragma warning(disable : 4589) // Constructor of abstract class 'Slice::Type' ignores initializer...
-#endif
-
 using namespace std;
 using namespace Slice;
 
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -722,6 +722,13 @@ Slice::Type::usesClasses() const
 // Builtin
 // ----------------------------------------------------------------------
 
+void
+Slice::Builtin::destroy()
+{
+    _unit = nullptr;
+}
+
+
 UnitPtr
 Slice::Builtin::unit()
 {
@@ -873,6 +880,12 @@ Slice::Builtin::Builtin(const UnitPtr& unit, Kind kind) : _unit(unit), _kind(kin
 // ----------------------------------------------------------------------
 // Contained
 // ----------------------------------------------------------------------
+
+void
+Slice::Contained::destroy()
+{
+    _unit = nullptr;
+}
 
 ContainerPtr
 Slice::Contained::container() const
@@ -2352,6 +2365,13 @@ Slice::Module::visit(ParserVisitor* visitor)
     }
 }
 
+void
+Slice::Module::destroy()
+{
+    Container::destroy();
+    Contained::destroy();
+}
+
 Slice::Module::Module(const ContainerPtr& container, const string& name) : Contained(container, name) {}
 
 // ----------------------------------------------------------------------
@@ -2428,6 +2448,7 @@ Slice::ClassDef::destroy()
     _declaration = nullptr;
     _base = nullptr;
     Container::destroy();
+    Contained::destroy();
 }
 
 DataMemberPtr
@@ -2896,6 +2917,7 @@ Slice::InterfaceDef::destroy()
     _declaration = nullptr;
     _bases.clear();
     Container::destroy();
+    Contained::destroy();
 }
 
 OperationPtr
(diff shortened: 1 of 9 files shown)


cpp/src/Slice/Parser.h
```diff
@@ -356,7 +347,9 @@ namespace Slice
             KindValue
         };
 
-        Builtin(const UnitPtr& unit, Kind kind);
+        Builtin(UnitPtr unit, Kind kind);
+        void destroy();
```
bernardnormier, line 351: I don't understand the logic of destroy(). Which cycle are you breaking?

cpp/src/Slice/Parser.cpp
```diff
@@ -1067,9 +1061,9 @@ Slice::Contained::getDeprecationReason() const
 }
 
 Slice::Contained::Contained(const ContainerPtr& container, string name)
-    : SyntaxTreeBase(container->unit()),
-      _container(container),
-      _name(std::move(name))
+    : _container(container),
+      _name(std::move(name)),
+      _unit(container->unit())
```
bernardnormier, line 1066: I am puzzled by this code: instead of keeping _unit, you could just return _container->unit(), and then you don't need any destroy, since destroy() doesn't clear _container.

bernardnormier (commented):
It would be helpeful to explain which cycles the various destroy() functions break.

InsertCreativityHere:
> It would be helpeful to explain which cycles the various destroy() functions break.

These were already here, and the eventual goal is to completely remove them. So I didn't think to add comments to them.
This whole system of holding `shared_ptr` cycles and having `destroy` to break them is just flawed.

We should be using `weak_ptr`, and avoid the cycles entirely, no need for this `destroy` logic.

cpp/src/Slice/Parser.h
InsertCreativityHere: Yes, using `weak_ptr` is the obvious next step in refactoring this.
And what we should of been using from the beginning.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: I hadn't even thought of that.
That's a good idea.

cpp/src/Slice/Parser.h
InsertCreativityHere: But just for the record, that isn't what this PR was about.
I don't want this to get beyond it's scope, and this was mostly about slimming down `SyntaxTreeBase` and untangling our inheritance hierarchy.

Although I do plan to eventually switch to using `WeakPtr` instead of this current approach of ours.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: There is a downside to this: we would need to 'artificially' remove `const` from some functions.

`unit()` is a non-const function because we need the overload of `shared_from_this` which returns a non-const ptr.
Whereas accessing the `_unit` field _is_ const. And we do have const functions which use this field instead of calling `unit()`, namely `mappedScope`, `mappedName`, etc. Which in turn are used by other const functions.

I'd rather just leave this field be.
When we switch to using `weak_ptr`, all these destroy will go away anyways.

bernardnormier: The unit() accessor should be const.

> unit() is non-const because we need the overload of shared_from_this 

You can cast-away constness in this case, as in:
```cpp
const_cast<Unit*>(this)->shared_from_this();
```

InsertCreativityHere: I knew I had this ability, but wasn't sure if it was 'safe' to do.
Fixed!

cpp/src/Slice/Parser.h
```diff
@@ -385,6 +379,7 @@ namespace Slice
     class Contained : public virtual SyntaxTreeBase
     {
     public:
+        virtual void destroy() {};
```
bernardnormier, line 382: no semicolon after `{}`

bernardnormier (approved):

cpp/src/Slice/Parser.h
InsertCreativityHere: I'm surprised `clang-format` didn't complain...
Fixed!

PR #3500, opened 2025-02-05
Add support for build Debian 12 packages

pepone:
This PR add support for building the DEB packages using Debian 12 (bookworm).

I have to exclude the Python3 packages because bookworm doesn't provide Python 3.12 which is the minimum required release for Ice 3.8.


packaging/dpkg/debian/control
```diff
@@ -8,9 +8,9 @@ Build-Depends: debhelper,
                dh-exec,
                dh-php (>= 0.20),
                dh-python,
-               libbluetooth-dev [!hurd-i386 !kfreebsd-i386 !kfreebsd-amd64],
+               libbluetooth-dev <!hurd-i386 !kfreebsd-i386 !kfreebsd-amd64>,
```
pepone, line 11: this exclusion list was bogus, it should use `<` instead of `[`.

packaging/dpkg/debian/control
```diff
@@ -21,8 +21,8 @@ Build-Depends: debhelper,
                openssl,
                php-all-dev,
                php-cli,
-               python3 (>= 3.12),
-               python3-dev (>= 3.12),
+               python3 (>= 3.12) <!no-python312>,
```
pepone, line 24: This allow to build with `no-python312` build profile, without depending on python3.12+

packaging/dpkg/debian/control
```diff
@@ -21,8 +21,8 @@ Build-Depends: debhelper,
...
+               python3-dev (>= 3.12) <!no-python312>,
                python3-setuptools,
                python3-passlib
```
pepone, line 27: We still keep passlib for the tests, we can probably remove setuptools.

packaging/dpkg/docker/debian-12/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+# Install dependencies
+RUN apt-get update && apt-get install -y \
+    php-dev python3 python3-dev python3-passlib ruby-full gdb \
+    libbluetooth-dev libbz2-dev libdbus-1-dev libedit-dev libexpat1-dev liblmdb-dev \
+    libmcpp-dev libssl-dev libsystemd-dev \
+    dh-exec dh-php dh-python locales-all php-all-dev python3-setuptools \
+    build-essential \
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get clean
```
pepone, line 11: I would be best to install build dependencies dynamically when building the package, I leave that for a follow-up PR.

This will ensure that the package build dependencies are correct.

packaging/dpkg/debian/control
pepone: Debian packaging has a target to run the tests, those are run for example when we upload to the debian repositories.

packaging/dpkg/build-package.sh
```diff
@@ -9,7 +9,7 @@ fi
 
 # Generate the upstream tarball
 echo "Creating tarball for ICE_VERSION=$ICE_VERSION"
-tar -czf /workspace/zeroc-ice_${ICE_VERSION}.orig.tar.gz -C /workspace/ice .
+tar -czf /workspace/zeroc-ice_${ICE_VERSION}.orig.tar.gz --exclude=.git -C /workspace/ice .
```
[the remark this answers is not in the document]
pepone: no good reason, I'm just not familiar with it.

packaging/dpkg/debian/control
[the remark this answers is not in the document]
bernardnormier: We need python to run the tests in any case, so passlib or not doesn't change much.

bernardnormier (approved):

PR #3502, opened 2025-02-05
Python Metadata Validation

InsertCreativityHere:
This PR switches `slice2py` to use the 'new' metadata validator (I added 3 months ago).

The implementation, old validation, and docs disagree about these metadata, so insight would be appreciated:
`python:array.array`, `python:numpy.ndarray`, and `python:memoryview:X`

----

The docs state that `array.array` and `numpy.ndarray` are only valid on: `integral builtin types excluding strings`. But:
1) "strings" are not an "integral" type... So I don't see why we mention this 'exclusion' at all.
2) It says for "integral" types... but we have tests using this metadata with seqs of `bool`, `float`, and `double`.

So... I'm guessing that the documentation is just wrong. And it should say:
`... are valid on sequences of numeric built-in types and bools`

----

For `python:memoryview:X`.
The docs list no restriction on what this can be applied to.
But, the old validation for `memoryview` re-used the logic for `array` and `ndarray`, restricting what it can be applied to.

Is this correct; that `memoryview` can only go on sequences of built-in numeric types and bools?
Or did we incorrectly re-use logic, and in reality, this metadata can be used with any sequence type?

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -2544,12 +2544,50 @@ Slice::Python::validateMetadata(const UnitPtr& u)
 {
     map<string, MetadataInfo> knownMetadata;
 
-    // "python:pkgdir"
-    MetadataInfo pkgdirInfo = {
-        .validOn = {typeid(Unit)},
-        .acceptedArgumentKind = MetadataArgumentKind::RequiredTextArgument,
+    // "python:<array-type>"
+    MetadataInfo arrayTypeInfo = {
+        .validOn = {typeid(Sequence)},
+        .acceptedArgumentKind = MetadataArgumentKind::NoArguments,
+        .acceptedContext = MetadataApplicationContext::DefinitionsAndTypeReferences,
+        .extraValidation = [](const MetadataPtr& m, const SyntaxTreeBasePtr& p) -> optional<string>
+        {
+            if (auto sequence = dynamic_pointer_cast<Sequence>(p))
+            {
+                BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(sequence->type());
+                if (!builtin || !(builtin->isNumericType() || builtin->kind() == Builtin::KindBool))
+                {
+                    return "the '" + m->directive() +
+                           "' metadata can only be applied to sequences of bools, bytes, shorts, ints, longs, floats, "
+                           "or doubles";
+                }
+            }
+            return nullopt;
+        },
     };
-    knownMetadata.emplace("python:pkgdir", std::move(pkgdirInfo));
+    knownMetadata.emplace("python:array.array", arrayTypeInfo);
+    knownMetadata.emplace("python:numpy.ndarray", std::move(arrayTypeInfo));
+
+    // "python:memoryview"
+    MetadataInfo memoryViewInfo = {
+        .validOn = {typeid(Sequence)},
+        .acceptedArgumentKind = MetadataArgumentKind::RequiredTextArgument,
+        .acceptedContext = MetadataApplicationContext::DefinitionsAndTypeReferences,
+        // TODO: is this restriction correct? Or can memory-view apply to any sequence types?
+        .extraValidation = [](const MetadataPtr&, const SyntaxTreeBasePtr& p) -> optional<string>
+        {
+            if (auto sequence = dynamic_pointer_cast<Sequence>(p))
+            {
+                BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(sequence->type());
+                if (!builtin || !(builtin->isNumericType() || builtin->kind() == Builtin::KindBool))
+                {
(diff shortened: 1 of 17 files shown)


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2426,8 +2403,7 @@ Slice::Python::getImportFileName(const string& file, const UnitPtr& ut, const ve
 void
 Slice::Python::generate(const UnitPtr& un, bool all, const vector<string>& includePaths, Output& out)
 {
-    Slice::Python::MetadataVisitor visitor;
-    un->visit(&visitor);
+    validateMetadata(un);
```
bernardnormier, line 2406: we should name the UnitPtr parameter unit.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2563,223 +2539,99 @@ Slice::Python::printHeader(IceInternal::Output& out)
     out << "#\n";
 }
 
-bool
-Slice::Python::MetadataVisitor::visitUnitStart(const UnitPtr& unit)
+void
+Slice::Python::validateMetadata(const UnitPtr& u)
```
bernardnormier, line 2543: unit please

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2563,223 +2539,99 @@ Slice::Python::printHeader(IceInternal::Output& out)
...
+    // "python:package"
+    MetadataInfo packageInfo = {
+        .validOn = {typeid(Module), typeid(Unit)},
+        .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
+        .extraValidation = [](const MetadataPtr&, const SyntaxTreeBasePtr& p) -> optional<string>
         {
-            SequencePtr seq = dynamic_pointer_cast<Sequence>(type);
-            if (seq)
+            // If 'python:package' is applied to a module, it must be a top-level module.
+            // // Top-level modules are contained by the 'Unit'. Non-top-level modules are contained in 'Module's.
```
bernardnormier, line 2599: Double comment

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2563,223 +2539,99 @@ Slice::Python::printHeader(IceInternal::Output& out)
...
+    // "python:seq"
+    // We support 3 arguments to this metadata: "default", "list", and "tuple".
+    // We also allow users to omit the "seq" in the middle, ie. "python:seq:list" and "python:list" are equivalent.
+    MetadataInfo seqInfo = {
+        .validOn = {typeid(Sequence)},
+        .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
+        .validArgumentValues = {{"default", "list", "tuple"}},
+        .acceptedContext = MetadataApplicationContext::DefinitionsAndTypeReferences,
+    };
+    MetadataInfo unqualifiedSeqInfo = {
```
bernardnormier, line 2625: It would be clearer to define this variable after:
```cpp
knownMetadata.emplace("python:seq", std::move(seqInfo));
```

pepone:
> ... are valid on sequences of numeric built-in types and bools

Yes. I think this is correct.

> Is this correct; that memoryview can only go on sequences of built-in numeric types and bools?

Yes. We should have the same restrictions

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2563,223 +2539,99 @@ Slice::Python::printHeader(IceInternal::Output& out)
...
-            // Must be a top-level module.
-            if (dynamic_pointer_cast<Unit>(p->container()) && directive == "python:package")
+            return nullopt;
+        },
+    };
+    knownMetadata.emplace("python:array.array", arrayTypeInfo);
+    knownMetadata.emplace("python:numpy.ndarray", std::move(arrayTypeInfo));
+
+    // "python:memoryview"
+    MetadataInfo memoryViewInfo = {
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes we can; fixed.

cpp/src/slice2py/PythonUtil.cpp
InsertCreativityHere: Fine, I moved it.

cpp/src/slice2py/PythonUtil.cpp
InsertCreativityHere: Good catch! Fixed.

cpp/src/slice2py/PythonUtil.cpp
InsertCreativityHere: Fixed.

cpp/src/slice2py/PythonUtil.cpp
InsertCreativityHere: I fixed it in these locations, but in general we have many different (and inconsistent) names for unit.
`unt` is my personal favorite. We just have to save that `i` character...

PR #3503, opened 2025-02-05
Simplify 'swift:module' Prefix Check

InsertCreativityHere:
`slice2swift` has a special check for the `swift:module` metadata to ensure it's used consistently.

Right now, we maintain a pair of maps in a field, and each `visitModule` uses these fields.
But, since this check only cares about top-level modules, we could instead just do this in a single function:
```
visitUnit() {
    // declare the 2 maps here.
    for (module : unit->modules()) { ... }
}
```
And IMO a stand-alone, self-contained function is nicer than fields which are checked and updated by multiple calls.

### DIFF
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -1266,49 +1266,54 @@ SwiftGenerator::writeMarshalUnmarshalCode(
 }
 
 bool
-SwiftGenerator::MetadataVisitor::visitModuleStart(const ModulePtr& p)
+SwiftGenerator::MetadataVisitor::visitUnitStart(const UnitPtr& unit)
 {
-    if (dynamic_pointer_cast<Unit>(p->container()))
+    using ModuleMap = std::map<std::string, std::string>;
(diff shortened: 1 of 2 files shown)


externl (approved):

bernardnormier:
I thought the whole point of the Parser-provided metadata validation was to get rid of all Metadata visitors.

InsertCreativityHere:
> I thought the whole point of the Parser-provided metadata validation was to get rid of all Metadata visitors.

It is. But in my opinion this logic doesn't belong in the metadata validator at all.
So, I wanted to put all this logic in a stand-alone function that can be easily moved.

After this is merged, then I'll open a follow-up PR to use the new metadata visitor properly.
I wanted to do this in a separator step so it was easier to follow, and because I was under the impression that
small and focused PRs were preferable to huge PRs that change lots of stuff at once.

InsertCreativityHere:
> However, it seems your Metadata visitor is checking more than that.

Also agreed. I'm going to write a blog about it later, but I believe that the "prefix" checking is unnecessary,
and goes against our philosophy of "different module openings can have different metadata on them".

But this PR isn't about removing this logic. It's supposed to a small PR that sequesters this logic away into a single function.

PR #3509, opened 2025-02-06
Add workflow to build RPM packages

pepone:
This PR is the initial work to get RPM building in CI. There is more to do, but merging this would allow us to build the packages from CI and test things.

### DIFF
--- a/.github/actions/build-rpm-package/action.yml
+++ b/.github/actions/build-rpm-package/action.yml
@@ -24,11 +24,17 @@ runs:
 
     - name: Run Package Build with Mounted Source
       run: |
-        docker run --rm \
-          -v "$GITHUB_WORKSPACE:/workspace" \
-          -e ICE_VERSION="${{ inputs.ice_version }}" \
-          -e BUILD_ARCH="${{ inputs.build_arch }}" \
-          ice-rpm-package-builder /workspace/ice/packaging/rpm/build-package.sh
+        if [[ -n "${{ inputs.build_arch }}" ]]; then
+          docker run --rm \
+            -v "$GITHUB_WORKSPACE:/workspace" \
+            -e ICE_VERSION="${{ inputs.ice_version }}" \
+            ice-rpm-package-builder bash -c "setarch ${{ inputs.build_arch }} /workspace/ice/packaging/rpm/build-package.sh"
+        else
+          docker run --rm \
+            -v "$GITHUB_WORKSPACE:/workspace" \
+            -e ICE_VERSION="${{ inputs.ice_version }}" \
+            ice-rpm-package-builder /workspace/ice/packaging/rpm/build-package.sh
+        fi
       shell: bash
 
     - name: Upload Artifacts
--- a/.github/actions/build-rpm-package/action.yml
+++ b/.github/actions/build-rpm-package/action.yml
@@ -5,8 +5,12 @@ inputs:
     description: "The Ice version to build"
     required: true
   os:
-    description: "The target OS (e.g., ubuntu-24.04, ubuntu-24.04-arm)"
+    description: "The target OS (e.g., rhel9, rhel9-arm64, rhel9-x86)"
     required: true
+  build_arch:
+    description: "The build architecture for cross multillib builds (e.g., i686)"
+    required: false
+    default: ""
   dockerfile_path:
     description: "Path to the Dockerfile"
     required: true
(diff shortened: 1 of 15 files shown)


.github/actions/build-rpm-package/action.yml
```diff
@@ -0,0 +1,38 @@
+name: "Build RPM Package"
+description: "Builds a RPM package using a Docker image"
+inputs:
+  ice_version:
+    description: "The Ice version to build"
+    required: true
+  os:
+    description: "The target OS (e.g., rhel9, rhel9-arm64, rhel9-x86)"
```
pepone, line 8: This is just used as a prefix for uploaded packages.

.github/actions/build-rpm-package/action.yml
```diff
@@ -0,0 +1,38 @@
...
+  dockerfile_path:
+    description: "Path to the Dockerfile"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Build Docker Image
+      run: docker build --pull -f ${{ inputs.dockerfile_path }} -t ice-rpm-package-builder .
+      shell: bash
```
pepone, line 23: As for DEB packages the image build will go into a separate workflow later.

packaging/rpm/JGOODIES_LICENSE
pepone, line 1: For now I just copied this XXX_LICENSE files from ice-packaging/ice/rpm to the packagin/rpm dir. We can probably move this to a better place.

packaging/rpm/README
```diff
@@ -0,0 +1,54 @@
+======================================================================
```
pepone, line 1: This is still the README from ice-packaging/ice/rpm.

packaging/rpm/build-package.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -eux  # Exit on error, print commands
+
```
pepone, line 3: We run this script in the Docker container to build the packages.

packaging/rpm/build-package.sh
```diff
@@ -0,0 +1,23 @@
...
+mkdir -p "$RPM_BUILD_ROOT"
+cd "$RPM_BUILD_ROOT"
+
+mkdir -p BUILD BUILDROOT RPMS SOURCES SPECS SRPMS
+cp /workspace/ice/packaging/rpm/ice.spec "$RPM_BUILD_ROOT/SPECS"
+
+cd "$RPM_BUILD_ROOT/SOURCES"
+spectool -g "$RPM_BUILD_ROOT/SPECS/ice.spec"
+
+dnf builddep --spec "$RPM_BUILD_ROOT/SPECS/ice.spec" -y --setopt=multilib_policy=all
```
pepone, line 15: This command installs the build dependencies, and we should probably move this to the Dockerfile to avoid installing then with each build.

Currently we build the container in each run, but we want to move the container build to its own job.

packaging/rpm/ice.spec
```diff
@@ -0,0 +1,854 @@
+#
+# Copyright (c) ZeroC, Inc. All rights reserved.
+#
```
pepone, line 3: This is mostly a copy of the 3.7 ice.spec from the ice-packaging repository.

Removed 3.8 unsupported platforms and features but still requires more work.

packaging/rpm/docker/rhel9/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM registry.access.redhat.com/ubi9
+
+# Enable ZeroC third-party repository for mcpp-devel
+RUN curl -o /etc/yum.repos.d/zeroc-thirdparty.repo https://builds.testing.zeroc.com/public/dist/unstable/thirdparty/el9/zeroc-thirdparty.repo
```
externl, line 4: This is not going to work in CI. Maybe we should also build the third party rpms as well, or build them in and have them installed in the container already.

pepone: I'm uploading them to 3.7 public repository, they are required to build 3.7 from sources and are missing there.

bernardnormier (approved):
Should all the sample config files (for Glacier2, IceGrid etc.) be with in the packaging repo?

We also have them in ice/cpp/config which gives the impression it's the main copy.

pepone:
> Should all the sample config files (for Glacier2, IceGrid etc.) be in the packaging repo?

Note all packaging files for 3.8 go into the main ice repository

> We also have them in ice/cpp/config which gives the impression it's the main copy.

I will try to upgrade to use those in a follow up PR.

PR #3511, opened 2025-02-06
Include slice2js in ice NPM package

pepone:
This PR updates the NPM ice package to include the slice2js compiler.

It can include the slice2js from the current C++ build, or prebuilt compilers from `SLICE2JS_STAGING_PATH` environment variable.

For CI builds and release packages we will include the slice2js for all supported platforms.

### DIFF
--- a/js/BUILDING.md
+++ b/js/BUILDING.md
@@ -1,6 +1,6 @@
 # Ice for JavaScript Build Instructions
 
-This file describes how to build and install Ice for JavaScript from source code. If you prefer, you can also download 
+This file describes how to build and install Ice for JavaScript from source code. If you prefer, you can also download
 [binary distributions] for the supported platforms.
 
 <!-- TOC depthFrom:2 -->
--- a/js/BUILDING.md
+++ b/js/BUILDING.md
@@ -1,8 +1,7 @@
 # Ice for JavaScript Build Instructions
 
-This file describes how to build and install Ice for JavaScript from source
-code. If you prefer, you can also download [binary distributions] for the
-supported platforms.
+This file describes how to build and install Ice for JavaScript from source code. If you prefer, you can also download 
+[binary distributions] for the supported platforms.
 
 <!-- TOC depthFrom:2 -->
 
@@ -24,7 +23,7 @@ To build Ice for JavaScript you must have the following:
 
 - The `slice2js` compiler from Ice for C++. If you have not built Ice for C++ in this source distribution, refer to the
   [C++ build instructions](../cpp/BUILDING.md).
-- Node.js 20 or later
+- Node.js 23 or later
 
 ## Building the JavaScript libraries and NodeJS packages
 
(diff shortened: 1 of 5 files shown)


js/package.json
```diff
@@ -14,7 +14,10 @@
         "rpc"
     ],
     "type": "module",
-    "main": "src/index.js",
+    "exports": {
```
pepone, line 17: Exports is recommended for packages targeting modern NodeJS versions.

The new exported script allows to

```
import { slice2js, slicePath, compile } from "ice/slice2js";
```

`slice2js` is the full path to the slice2js for the given platform.
`slicePath` is the full path to the directory containing the Slice files, which are also included in the package.
`compile` allows to run the compiler using `child_process.spawn` we previously provided this function in the npm slice2js package.

js/package.json
```diff
@@ -35,9 +38,16 @@
...
+        "bin/",
+        "slice/",
+        "README.md"
+    ],
     "scripts": {
         "build": "gulp",
         "dist": "gulp dist",
-        "clean": "gulp clean"
+        "clean": "gulp clean",
+        "prepack": "node scripts/prepack.js"
```
pepone, line 51: The prepack script runs before "pack" and copies Slice files and Slice to JavaScript compilers to the required location.

js/BUILDING.md
```diff
@@ -24,7 +23,7 @@ To build Ice for JavaScript you must have the following:
 
 - The `slice2js` compiler from Ice for C++. If you have not built Ice for C++ in this source distribution, refer to the
   [C++ build instructions](../cpp/BUILDING.md).
-- Node.js 20 or later
+- Node.js 23 or later
```
externl, line 26: Node 22 is the current LTS. I would stick to this.

js/scripts/prepack.js
```diff
@@ -0,0 +1,73 @@
...
+
+        if (fs.lstatSync(srcFile).isDirectory()) {
+            copyFolderSync(srcFile, destFile);
+        } else {
+            fs.copyFileSync(srcFile, destFile);
+        }
+    });
+}
+
+console.log(`Copying Slice files from ${sourceDir} to ${sliceDestDir}... `);
```
externl, line 56:
```suggestion
console.log(`Copying Slice files from ${sourceDir} to ${sliceDestDir}...`);
```

js/scripts/prepack.js
```diff
@@ -0,0 +1,73 @@
...
+const stagingPath = process.env.SLICE2JS_STAGING_PATH;
+if (stagingPath) {
+    console.log(`Copying Slice compilers from ${sourceDir} to ${binDestDir}...`);
+    copyFolderSync(sourceDir, binDestDir);
+    console.log('Done.');
+} else {
+    console.log(`Compiling Slice compiler from cpp build...`);
+    let binDir = path.resolve(__dirname, '../../cpp/bin');
+    if (os.platform() === 'win32') {
+        binDir = path.resolve(binDir, process.env.PLATFORM || "x64", process.env.CONFIGURATION || "Release");
```
externl, line 69: Shouldn't we always use x64. Sounds like we could otherwise end up with a 32-bit compiler in a 64-bit arch prefix. 

I assume we're only going to ship a x64 compiler, no need to support something different here.

externl (approved):

PR #3517, opened 2025-02-07
Rework printTarget to show non-IP transport

bernardnormier:
This PR updates the LoggerMiddleware printTarget to show the transport (or colloc) for non-IP connections.

### DIFF
--- a/cpp/src/Ice/LoggerMiddleware.cpp
+++ b/cpp/src/Ice/LoggerMiddleware.cpp
@@ -145,8 +145,8 @@ LoggerMiddleware::printTarget(LoggerOutputBase& out, const Current& current) con
 
         if (auto ipConnInfo = dynamic_pointer_cast<IPConnectionInfo>(connInfo))
         {
-            out << ipConnInfo->localAddress << ':' << ipConnInfo->localPort << "<->"
-                << ipConnInfo->remoteAddress << ':' << ipConnInfo->remotePort;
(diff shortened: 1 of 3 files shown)


cpp/src/Ice/LoggerMiddleware.cpp
```diff
@@ -133,18 +133,29 @@ LoggerMiddleware::printTarget(LoggerOutputBase& out, const Current& current) con
...
     }
+    out << " over ";
 
     if (current.con)
     {
-        for (ConnectionInfoPtr connInfo = current.con->getInfo(); connInfo; connInfo = connInfo->underlying)
+        ConnectionInfoPtr connInfo = current.con->getInfo();
+        while (connInfo->underlying)
         {
-            auto ipConnInfo = dynamic_pointer_cast<IPConnectionInfo>(connInfo);
```
bernardnormier, line 141: The old logic was not good: we should not dynamic_cast for each iteration!

csharp/src/Ice/Internal/LoggerMiddleware.cs
```diff
@@ -122,29 +122,35 @@ private void printTarget(StringBuilder sb, Current current)
             sb.Append(Ice.UtilInternal.StringUtil.escapeString(current.facet, "", _toStringMode));
         }
 
+        sb.Append(" over ");
+
         if (current.con is not null)
         {
-            try
```
bernardnormier, line 127: Unclear why we had this try/catch block. Accessing fields doesn't throw.

externl (approved):

pepone (approved):

PR #3518, opened 2025-02-07
Add workflow for building XCFramework packages

pepone:


.github/workflows/build-xcframework-packages.yml
```diff
@@ -0,0 +1,25 @@
...
+        uses: ./.github/actions/setup-dependencies
+
+      - name: Build XCFramework Packages
+        run: make OPTIMIZE=yes PLATFORMS="all" CONFIGS="static" srcs
+
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: xcframework-packages
+          path: cpp/lib/XCFrameworks/*
```
externl, line 25: This will create on zip with all the frameworks. When we do a tag we'll want each xcframeworks to be in its own zip. 

https://developer.apple.com/documentation/xcode/distributing-binary-frameworks-as-swift-packages

Might be easier to separate that here for when we want to upload tag artifacts.

PR #3521, opened 2025-02-07
Fixes to build MATLAB toolboxes on CI

pepone:
This PR include fixes to build MATLAB toolboxes for Windows and Linux on CI.

### DIFF
--- a/.github/workflows/build-matlab-packages.yml
+++ b/.github/workflows/build-matlab-packages.yml
@@ -25,7 +25,9 @@ jobs:
           use_matlab: true
 
       - name: Build MATLAB ToolBox
-        run: make -C matlab toolbox
+        run: |
+          make -C cpp srcs
+          make -C matlab toolbox
         if: runner.os == 'Linux'
 
       - name: Build MATLAB ToolBox
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,8 +130,6 @@ jobs:
         with:
           working_directory: ${{ matrix.working_directory || '.' }}
           flags: ${{ matrix.test_flags }}
-        # Don't test matlab on Windows (see setup-dependencies/action.yml)
-        if: matrix.config != 'matlab' || runner.os != 'Windows'
 
       - name: Cross Test ${{ matrix.config }} on ${{ matrix.os }}
         uses: ./.github/actions/test
--- a/matlab/Makefile
+++ b/matlab/Makefile
@@ -115,9 +115,15 @@ $(icetoolbox_file):: $(icethunk_target) $(slice2matlab_path) $(lang_srcdir)/lib/
 	cp -rf $(cpp_bindir)/slice2matlab $(lang_srcdir)/toolbox/build/
 	# Doc files
 	cp -rf $(lang_srcdir)/toolbox/doc $(lang_srcdir)/toolbox/build
+ifneq (,$(MATLAB_COMMAND))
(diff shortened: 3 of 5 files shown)


.github/workflows/build-matlab-packages.yml
```diff
@@ -25,7 +25,9 @@ jobs:
           use_matlab: true
 
       - name: Build MATLAB ToolBox
-        run: make -C matlab toolbox
+        run: |
+          make -C cpp srcs
```
pepone, line 29: We were missing the C++ dependencies here

.github/workflows/ci.yml
```diff
@@ -130,8 +130,6 @@ jobs:
         with:
           working_directory: ${{ matrix.working_directory || '.' }}
           flags: ${{ matrix.test_flags }}
-        # Don't test matlab on Windows (see setup-dependencies/action.yml)
-        if: matrix.config != 'matlab' || runner.os != 'Windows'
```
pepone, line 134: MATLAB For Windows should work with the fixes to Windows builds.

matlab/Makefile
```diff
@@ -115,9 +115,15 @@ $(icetoolbox_file):: $(icethunk_target) $(slice2matlab_path) $(lang_srcdir)/lib/
 	cp -rf $(cpp_bindir)/slice2matlab $(lang_srcdir)/toolbox/build/
 	# Doc files
 	cp -rf $(lang_srcdir)/toolbox/doc $(lang_srcdir)/toolbox/build
+ifneq (,$(MATLAB_COMMAND))
```
pepone, line 118: MATLAB_COMMAND is set by CI to the executable that we need to run. We were already using this in the tests, but there were a few additional places where we still used matlab directly.

matlab/msbuild/ice.proj
```diff
@@ -6,36 +6,30 @@
         <Configuration Condition="'$(Configuration)' == ''">Release</Configuration>
         <RootDir>$([System.IO.Path]::GetFullPath( $(MSBuildThisFileDirectory)..\msbuild ))</RootDir>
         <ToolboxDir>$(MSBuildThisFileDirectory)..\toolbox\build</ToolboxDir>
```
pepone, line 8: This MSBuild project was trying to guess the version and the matlab home using the matlab executable from path. It was far too complicated.

I simplified to use a default and allow override with env variables. It was the usage of matlab here what was causing the CI issues.

matlab/msbuild/ice.proj
```diff
@@ -6,36 +6,30 @@
...
+        <MatlabVersion Condition="'$(MatlabVersion)' == ''">R2024a</MatlabVersion>
+        <MatlabHome Condition="'$(MatlabHome)' == ''">C:\Program Files\MATLAB\$(MatlabVersion)</MatlabHome>
+        <!-- Override with the env settings -->
+        <MatlabVersion Condition="'$(MATLAB_VERSION)' != ''">$(MATLAB_VERSION)</MatlabVersion>
```
bernardnormier, line 12: Should this go up on line 10?

Presumably if you set MATLAB_VERSION but not MATLAB_HOME, you want MatlabHome to be set correctly.

bernardnormier (approved):

externl (approved):

matlab/msbuild/ice.proj
pepone: Yes. Fixed

PR #3528, opened 2025-02-10
Cleanup PHP Windows build instructions

### LINKED ISSUE #3497: Remove Windows instructions from php/BUILDING.md
Ice 3.7+ does not support PHP on Windows. The Windows instructions in BUILDING.md are confusing and make the instructions more complicated.

pepone:
Fix #3497

### DIFF
--- a/php/BUILDING.md
+++ b/php/BUILDING.md
@@ -2,41 +2,31 @@
 
 This page describes how to build and install Ice for PHP from source.
 
-ZeroC provides [binary distributions] for many platforms, including
-Linux, so building Ice for PHP from source is usually unnecessary.
+ZeroC provides [binary distributions] for many platforms, including Linux, so building Ice for PHP from source is
+usually unnecessary.
 
-* [PHP Build Requirements](#php-build-requirements)
-* [Building the PHP Extension](#building-the-php-extension)
-  * [Linux or macOS](#linux-or-macos)
-* [Installing the PHP Extension](#installing-the-php-extension)
-  * [Linux or macOS](#linux-or-macos-1)
-* [PHP Dependencies](#php-dependencies)
-* [PHP Source Files](#php-source-files)
-* [Running the PHP Tests](#running-the-php-tests)
-* [Web Server Permissions](#web-server-permissions)
-* [SELinux Notes for PHP](#selinux-notes-for-php)
+- [PHP Build Requirements](#php-build-requirements)
+- [Building the PHP Extension](#building-the-php-extension)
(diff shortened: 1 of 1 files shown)


php/BUILDING.md
```diff
@@ -123,14 +103,12 @@ libIceSSL
 libbz2
 ```
 
-In general, these libraries must reside in a directory of the user's PATH.
-For Web servers, the libraries may need to reside in a system directory. For
-example, on Linux you can add the directory containing the Ice run-time
-libraries to `/etc/ld.so.conf` and run `ldconfig`.
+In general, these libraries must reside in a directory of the user's PATH. For Web servers, the libraries may need to
```
externl, line 106: > directory of the user's PATH

I think this is only true for Windows

php/BUILDING.md
```diff
@@ -139,101 +117,89 @@ phpInfo();
...
-modify their permissions to make them sufficiently accessible.
+On Linux, Apache typically runs in the `apache` account, so you will either need to change the owner or group of the
+libraries and other resources, or modify their permissions to make them sufficiently accessible.
 
 ## SELinux Notes for PHP
 
-SELinux augments the traditional Unix permissions with a number of new features.
-In particular, SELinux can prevent the httpd daemon from opening network
-connections and reading files without the proper SELinux types.
+SELinux augments the traditional Unix permissions with a number of new features. In particular, SELinux can prevent the
```
externl, line 176: Is this something we really need to worry about documenting?

php/BUILDING.md
```diff
@@ -139,101 +117,89 @@ phpInfo();
...
+The Web server normally runs in a special user account that may not necessarily have access to the Ice extension, its
+dependent libraries and PHP source files, and other resources such as Ice configuration and your application scripts.
+It is very important that you review the permissions of these files and verify that the Web server has sufficient
+access.
 
-On Linux, Apache typically runs in the `apache` account, so you will either
-need to change the owner or group of the libraries and other resources, or
-modify their permissions to make them sufficiently accessible.
+On Linux, Apache typically runs in the `apache` account, so you will either need to change the owner or group of the
+libraries and other resources, or modify their permissions to make them sufficiently accessible.
```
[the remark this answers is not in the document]
pepone: Agree, I would remove it.

externl (approved):
Looks good. What do we do for the tests. I assume we don't add to or modify all these php global config files. 

Do we pass some env variable or config file when calling `php`?

Issue #117, opened 2018-06-21
Improve metadata validation


bernardnormier:
There is currently not enough metadata validation in the Slice compilers.

pepone:
While adding metadata validation to Slice to Swift compiler, noticed that Slice::Dictionary API  doesn't allow to remove bogus metadata.

https://github.com/zeroc-ice/ice/blob/b0d4725aac41710d06078fc79f4a80d29209182c/cpp/src/Slice/Parser.h#L890-L891

This will be desirable to be able to sanitize the dictionary metadata like we do for other constructs.

I think we should come up with some kind of generic metadata validator that we can use everywhere, currently we are creating metadata visitors with each Slice compiler, but we don't have a consistent way to do the metada  validation.

InsertCreativityHere:
This was implemented in the following PRs:
- C++: https://github.com/zeroc-ice/ice/pull/3165
- C#, Java, JS, Ruby, PHP, MATLAB: https://github.com/zeroc-ice/ice/pull/3256
- Python: https://github.com/zeroc-ice/ice/pull/3502
- Swift: https://github.com/zeroc-ice/ice/pull/3530
We now have full validation of every known metadata, and reject all unknown metadata.

PR #3530, opened 2025-02-10
Use the 'New' Metadata Validator for `slice2swift`

InsertCreativityHere:
This PR switches `slice2swift` to use the new metadata validator built into `libSlice`.
It is the last language to be switched, and after this PR is merged, I will finally consider issue #117 fixed.

### DIFF
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -619,6 +619,69 @@ SwiftGenerator::validateMetadata(const UnitPtr& u)
     u->visit(&visitor);
 }
 
+void
+SwiftGenerator::validateSwiftModuleMappings(const UnitPtr& unit)
+{
+    using ModuleMap = std::map<std::string, std::string>;
+    // Each Slice unit has to map all top-level modules to a single Swift module
+    ModuleMap mappedModules;
+    // With a given Swift module a Slice module has to map to a single prefix
+    std::map<std::string, ModuleMap> mappedModulePrefixes;
+
+    // Any modules that are directly contained on the unit are (by definition) top-level modules.
+    // And since there is only one unit per compilation, this must be all the top-level modules.
+    for (const auto& module : unit->modules())
+    {
+        string swiftPrefix;
+        string swiftModule = getSwiftModule(module, swiftPrefix);
+
+        const string filename = module->definitionContext()->filename();
+        auto current = mappedModules.find(filename);
+
+        if (current == mappedModules.end())
+        {
+            mappedModules[filename] = swiftModule;
+        }
+        else if (current->second != swiftModule)
+        {
+            ostringstream os;
+            os << "invalid module mapping:\n Slice module '" << module->scoped() << "' should map to Swift module '"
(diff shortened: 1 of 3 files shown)


cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -615,8 +616,104 @@ SwiftGenerator::writeServantDocSummary(IceInternal::Output& out, const Interface
...
+        .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
+    };
+    knownMetadata.emplace("swift:module", moduleInfo);
+
+    // Pass this information off to the parser's metadata validation logic.
+    Slice::validateMetadata(u, "swift", knownMetadata);
+}
+
+void
+SwiftGenerator::validateSwiftModuleMappings(const UnitPtr& unit)
```
InsertCreativityHere, line 657: All this code was just moved out of `MetadataValidator::visitUnit` into a new static function.
I didn't change anything else about it.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -615,8 +616,104 @@ SwiftGenerator::writeServantDocSummary(IceInternal::Output& out, const Interface
...
+    knownMetadata.emplace("swift:attribute", attributeInfo);
+
+    // "swift:class-resolver-prefix"
+    MetadataInfo classResolverPrefixInfo = {
+        .validOn = {typeid(Unit)},
+        .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
+    };
+    knownMetadata.emplace("swift:class-resolver-prefix", classResolverPrefixInfo);
+
+    // "swift:inherits"
```
bernardnormier, line 636: We should drop swift:inherits.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -615,8 +616,104 @@ SwiftGenerator::writeServantDocSummary(IceInternal::Output& out, const Interface
...
+        .acceptedArgumentKind = MetadataArgumentKind::RequiredTextArgument,
+        .mustBeUnique = false,
+    };
+    knownMetadata.emplace("swift:inherits", inheritsInfo);
+
+    // "swift:module"
+    MetadataInfo moduleInfo = {
+        .validOn = {typeid(Module)},
+        // Even though it's really 'module:prefix' the validator sees this as a single argument since there's no commas.
+        .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
```
bernardnormier, line 648: We should drop the MetadataArgumentKind::MultipleArguments (?) option since we don't use it for any Slice metadata except support warnings.

bernardnormier (approved):

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere: After this is merged I'll have a follow up PR to drop `swift:inherits` and loosen the `swift:module` check.
I think that should be all the issues around Swift metadata.

PR #3532, opened 2025-02-11
Remove unused code in Ice for MATLAB

bernardnormier:
This PR removes unused code in the Ice for MATLAB C++ code.


InsertCreativityHere (approved):
Always glad to see unused code get deleted!

I assume you found this while looking into the typeID->className mapping.
But is this a duplicate of the `idToClass` you mentioned on my blog?
Or does it turn out that MATLAB doesn't have the same issue that Java did?

pepone (approved):

bernardnormier:
Ice MATLAB uses an `idToClass` implemented in MATLAB:
https://github.com/zeroc-ice/ice/blob/079552e45b7a7c23445292a9f0feca80fb0bf8ed/matlab/lib/%2BIceInternal/Util.m#L31

Issue #3529, opened 2025-02-10
Loosen the 'Prefix' Check for `swift:module` Metadata


InsertCreativityHere:
The `swift:module` is of the form: `swift:module:<Module>:<Prefix>`, so it accepts two positional arguments.
If this metadata is applied to the same module within the same file (ie. you've re-opened it in the same file), we enforce that both the `<Module>` and `<Prefix>` parts are identical.

This is overzealous.
The modules do need to match, since Swift only supports having 1 module per file, but there's no reason for the prefixes to match necessarily.

We should remove this extra checking of the prefix.

InsertCreativityHere:
This was implemented in #3533.

PR #3533, opened 2025-02-11
Cleanup Swift Metadata

InsertCreativityHere:
This PR removes the `swift:inherits` metadata, and implements #3529 (we no longer require re-openings of the same module with `swift:module` metadatas to have the same 'prefix' argument)

**EDIT**: It also now changes all the identifiers in the `scope` tests.


cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -683,36 +670,6 @@ SwiftGenerator::validateSwiftModuleMappings(const UnitPtr& unit)
                << current->second << "'" << endl;
```
bernardnormier, line 670: I would keep the message on a single line, no \n.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -656,19 +648,14 @@ SwiftGenerator::validateMetadata(const UnitPtr& u)
 void
 SwiftGenerator::validateSwiftModuleMappings(const UnitPtr& unit)
 {
-    using ModuleMap = std::map<std::string, std::string>;
-    // Each Slice unit has to map all top-level modules to a single Swift module
-    ModuleMap mappedModules;
-    // With a given Swift module a Slice module has to map to a single prefix
-    std::map<std::string, ModuleMap> mappedModulePrefixes;
+    // Each Slice unit has to map all top-level modules to a single Swift module.
```
bernardnormier, line 651: It would be helpful to describe what this map is. It seems to be a map filename -> swift module name.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -656,19 +648,14 @@ SwiftGenerator::validateMetadata(const UnitPtr& u)
...
+    map<string, string> mappedModules;
 
     // Any modules that are directly contained on the unit are (by definition) top-level modules.
     // And since there is only one unit per compilation, this must be all the top-level modules.
     for (const auto& module : unit->modules())
```
bernardnormier, line 656: I would not use module as a variable name. It's an identifier with special meaning in C++20: https://en.cppreference.com/w/cpp/keyword

swift/test/Ice/scope/AllTests.swift
```diff
@@ -201,6 +201,36 @@ func allTests(helper: TestHelper) async throws {
         try test(cmap3["a"]!!.s == s1)
     }
 
+    do {
+        let obj = try communicator.stringToProxy("dpi:\(helper.getTestEndpoint())")!
```
bernardnormier, line 205: No stringToProxy followed by checkedCast in new code, please!

If you want to test ice_isA, call ice_isA.

swift/test/Ice/scope/Test.ice
```diff
@@ -156,26 +156,54 @@ module Test
...
+            Test::CSeq opCSeq(Test::CSeq c1, out Test::CSeq c2);
+            Test::CMap opCMap(Test::CMap c1, out Test::CMap c2);
+
+            void shutdown();
+        }
+    }
+}
 
-module Inner2
+// Check that it's okay for 'swift:module' to have different 'prefix' arguments.
```
bernardnormier, line 176: I am sure we can test this "feature" with a simpler test.

swift/test/Ice/scope/Test.ice
```diff
@@ -156,26 +156,54 @@ module Test
...
-        Test::SMap opSMap(Test::SMap s1, out Test::SMap s2);
-
-        Test::C opC(Test::C c1, out Test::C c2);
-        Test::CSeq opCSeq(Test::CSeq c1, out Test::CSeq c2);
-        Test::CMap opCMap(Test::CMap c1, out Test::CMap c2);
+        long l;
+    }
 
-        void shutdown();
+    interface I
```
bernardnormier, line 185: We should definitely never name an interface 'I'.

swift/test/Ice/scope/Test.ice
```diff
@@ -156,26 +156,54 @@ module Test
...
+    class Class1
     {
-        Test::S opS(Test::S s1, out Test::S s2);
-        Test::SSeq opSSeq(Test::SSeq s1, out Test::SSeq s2);
-        Test::SMap opSMap(Test::SMap s1, out Test::SMap s2);
-
-        Test::C opC(Test::C c1, out Test::C c2);
-        Test::CSeq opCSeq(Test::CSeq c1, out Test::CSeq c2);
-        Test::CMap opCMap(Test::CMap c1, out Test::CMap c2);
+        long l;
```
bernardnormier, line 182: Trying hard to find the worse names?

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere: I didn't even notice this was split up. Fixed.

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere: I reworked this code to eliminate the need for the map.
The previous version was running this check even in included files, but this is overzealous.
Now we only run this check in the file that is actually being compiled.

And since we're only checking one file, no need for this map of filenames.

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere: Fixed.
I renamed it to `mod`.

swift/test/Ice/scope/AllTests.swift
InsertCreativityHere: I was just copy-pasting from the lines above.
I removed all the checkedCast calls from this test; all of them were superfluous.

swift/test/Ice/scope/Test.ice
InsertCreativityHere: Well, I also want to make sure that it gets mapped to the correct place as well.
Not just that "the compiler doesn't issue an error anymore".

Because we don't have a test for the swift:module anywhere, and scope felt like an adequate place to put it.
Since the point of the metadata is changing the mapped scopes.

swift/test/Ice/scope/Test.ice
InsertCreativityHere: I was just being consistent with the rest of this test.
I updated the `scope` tests to use better names.

swift/test/Ice/scope/Test.ice
InsertCreativityHere: Nope, just following our old 'conventions'!
I updated almost all of the names in this test to be less horrible.

swift/test/Ice/scope/AllTests.swift
bernardnormier: If you don't remember the preferred syntax, check the top-level CHANGELOG-3.8.md or the greeter demo:

```swift
// Swift
let greeter = try makeProxy(
  communicator: communicator,
  proxyString: "greeter:tcp -h localhost -p 4061",
  type: GreeterPrx.self)
```

bernardnormier (approved):

swift/test/Ice/scope/AllTests.swift
InsertCreativityHere: I wasn't familiar enough with Swift to remember that we supported this there.
I updated `scope` in Swift to use `makeProxy`. The other languages were all already correct.

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere: The map was deleted.
The previous check was being applied to all files, even included ones.
This seemed over-zealous to me. Now we only check the single file that is being compiler.

And since we're only checking one file, we don't need a file-name map anymore.

externl (approved):

Issue #3534, opened 2025-02-11
Allow `java:identifier` on Classes and Exceptions


InsertCreativityHere:
We are unable to support `java:identifier` on Slice classes and exceptions because of the activator mechanism we use.
It expects there to be a simple algorithm to map from a Slice type-id to a Java class name.
`java:identifier` completely decouples these two things though; and supporting it on classes and exceptions would break this activator logic.

It is obviously preferable for us to support this metadata on classes/exceptions across all languages consistently.
We should spent some effort to see if the activator situation can be improved and made more robust in a way that let's us support this.

PR #3535, opened 2025-02-11
Add Support for `java:identifier` Metadata

InsertCreativityHere:
This PR adds support for `java:identifier`, just as I've already done for C++ and C#.
It cannot be applied to modules for the same reason we haven't added it in C++ and C# yet.

But, a difference in Java is that `java:identifier` cannot be applied to classes or exceptions do to how the Java activator works.
It expects a tight coupling between a class's typeID and Java class name. `java:identifier` completely decouples these.

And since I've removed the auto-escaping logic, any users with classes or exceptions that collide with Java keywords will be unable to upgrade past 3.7. I expect this is a very rare occurrence though. See #3534 for further discussion.

### DIFF
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -3883,7 +3883,12 @@ Slice::Gen::TypesVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
     out << eb;
     close();
 
-    open(("_" + getUnqualified(p) + "PrxI"), p->file());
+    // Add a '_' prefix to the file name. We want this '_' to come after any package scopes.
+    string absolute = getUnqualified(p) + "PrxI";
+    auto scopePos = absolute.rfind('.');
+    scopePos = (scopePos == string::npos ? 0 : scopePos + 1);
+    absolute.insert(scopePos, 1, '_');
+    open(absolute, p->file());
 
     Output& outi = output();
 
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -1963,11 +1963,17 @@ Slice::JavaVisitor::writeProxyDocComment(
     //
     if (!async)
     {
-        map<string, StringList> exDocs = dc->exceptions();
-        for (const auto& exDoc : exDocs)
+        for (const auto& [name, lines] : dc->exceptions())
         {
-            out << nl << " * @throws " << fixKwd(exDoc.first) << ' ';
-            writeDocCommentLines(out, exDoc.second);
+            string scopedName = name;
+            // Try to locate the exception's definition using the name given in the comment.
+            ExceptionPtr ex = p->container()->lookupException(name, false);
+            if (ex)
+            {
+                scopedName = ex->mappedScoped(".").substr(1);
+            }
+            out << nl << " * @throws " << scopedName << ' ';
+            writeDocCommentLines(out, lines);
         }
     }
 
@@ -2099,11 +2105,17 @@ Slice::JavaVisitor::writeServantDocComment(Output& out, const OperationPtr& p, c
         out << nl << " * @return A completion stage that the servant will complete when the invocation completes.";
     }
 
-    map<string, StringList> exDocs = dc->exceptions();
-    for (const auto& exDoc : exDocs)
+    for (const auto& [name, lines] : dc->exceptions())
     {
-        out << nl << " * @throws " << fixKwd(exDoc.first) << ' ';
-        writeDocCommentLines(out, exDoc.second);
+        string scopedName = name;
+        // Try to locate the exception's definition using the name given in the comment.
+        ExceptionPtr ex = p->container()->lookupException(name, false);
+        if (ex)
+        {
+            scopedName = ex->mappedScoped(".").substr(1);
+        }
+        out << nl << " * @throws " << scopedName << ' ';
+        writeDocCommentLines(out, lines);
     }
 
     if (!dc->seeAlso().empty())
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -61,7 +61,7 @@ namespace
 
         for (const auto& param : params)
         {
-            if (param->name() == name)
+            if (param->mappedName() == name)
             {
                 return name + "_";
             }
@@ -106,13 +106,13 @@ namespace
         auto hashPos = rawLink.find('#');
         if (hashPos != string::npos)
         {
-            result += Slice::JavaGenerator::fixKwd(rawLink.substr(0, hashPos));
+            result += rawLink.substr(0, hashPos);
             result += "#";
-            result += Slice::JavaGenerator::fixKwd(rawLink.substr(hashPos + 1));
+            result += rawLink.substr(hashPos + 1);
         }
         else
         {
-            result += Slice::JavaGenerator::fixKwd(rawLink);
+            result += rawLink;
         }
         return result + "}";
     }
@@ -130,7 +130,7 @@ Slice::JavaVisitor::getResultType(const OperationPtr& op, const string& package,
         const InterfaceDefPtr interface = op->interface();
         assert(interface);
         string abs = getUnqualified(interface, package);
-        string name = op->name();
+        string name = op->mappedName();
         name[0] = static_cast<char>(toupper(static_cast<unsigned char>(name[0])));
         return abs + "." + name + "MarshaledResult";
     }
@@ -139,7 +139,7 @@ Slice::JavaVisitor::getResultType(const OperationPtr& op, const string& package,
         const ContainedPtr c = dynamic_pointer_cast<Contained>(op->container());
         assert(c);
         const string abs = getUnqualified(c, package);
-        string name = op->name();
+        string name = op->mappedName();
         name[0] = static_cast<char>(toupper(static_cast<unsigned char>(name[0])));
         return abs + "." + name + "Result";
     }
@@ -186,11 +186,11 @@ Slice::JavaVisitor::writeResultType(
     const string& package,
     const optional<DocComment>& dc)
 {
-    string opName = op->name();
+    string opName = op->mappedName();
     opName[0] = static_cast<char>(toupper(static_cast<unsigned char>(opName[0])));
 
     out << sp;
-    writeDocComment(out, "Holds the result of operation " + op->name() + ".");
+    writeDocComment(out, "Holds the result of operation " + op->mappedName() + ".");
     out << nl << "public static class " << opName << "Result";
     out << sb;
 
@@ -201,7 +201,7 @@ Slice::JavaVisitor::writeResultType(
     const ParameterList outParams = op->outParameters();
     for (const auto& outParam : outParams)
     {
-        if (outParam->name() == "returnValue")
+        if (outParam->mappedName() == "returnValue")
         {
             retval = "_returnValue";
             break;
@@ -241,7 +241,7 @@ Slice::JavaVisitor::writeResultType(
             // Emit a doc comment for the constructor if necessary.
             //
             out << nl << "/**";
-            out << nl << " * This constructor makes shallow copies of the results for operation " << opName;
+            out << nl << " * This constructor makes shallow copies of the results for operation " << op->mappedName();
             if (generateMandatoryOnly)
             {
                 out << " (overload without Optional parameters).";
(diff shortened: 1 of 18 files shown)


cpp/src/slice2java/Gen.h
```diff
@@ -49,7 +49,6 @@ namespace Slice
         //
         // Compose the argument lists for an operation.
         //
-        std::vector<std::string> getArgs(const OperationPtr&);
```
InsertCreativityHere, line 52: Dead function. Never called.

cpp/src/slice2java/Gen.h
```diff
@@ -117,7 +116,7 @@ namespace Slice
     class Gen final
     {
     public:
-        Gen(const std::string&, std::string, const std::vector<std::string>&, std::string);
```
InsertCreativityHere, line 120: For some reason we were passing in the executable name (`slice2java`) here.
We never checked or used this field though. Dead.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -343,83 +319,6 @@ Slice::JavaGenerator::output() const
...
-    stringstream result;
-    for (const auto& id : ids)
-    {
-        result << "::" + id;
-    }
-    return result.str();
-}
-
-string
-Slice::JavaGenerator::convertScopedName(const string& scoped, const string& prefix, const string& suffix)
```
InsertCreativityHere, line 375: Now that `mappedScoped` has a 'separator' argument, we can let `libSlice` give us the separator we want (`.`),
instead of needing this function to massage it after-the-fact.

cpp/src/slice2java/JavaUtil.h
```diff
@@ -114,11 +104,7 @@ namespace Slice
         // package argument matches the entity's package name, then the
         // package is removed from the result.
         //
-        [[nodiscard]] static std::string getUnqualified(
-            const ContainedPtr&,
-            const std::string& = std::string(),
-            const std::string& = std::string(),
-            const std::string& = std::string());
```
InsertCreativityHere, line 121: We no longer pass `prefix` and `suffix` in here. Instead we just add them where we need them.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -528,7 +528,7 @@ public static String typeIdToClass(String id) {
             if (buf.length() > 0) {
                 buf.append('.');
             }
-            buf.append(fixKwd(s));
```
InsertCreativityHere, line 531: A result of this change is that if a user has a class/exception that collides with a Java keyword,
they will be stuck using 3.7. This is not optimal.

java/test/src/main/java/test/Ice/adapterDeactivation/Test.ice
```diff
@@ -8,6 +8,7 @@ module Test
 
 interface TestIntf
 {
+    ["java:identifier:_transient"]
     void transient();
```
InsertCreativityHere, line 12: I kept the operation name the same because this is what we call it in every other language's `adapterDeactivation` test.

java/test/src/main/java/test/Ice/packagemd/NoPackage.ice
```diff
@@ -25,7 +25,7 @@ exception E2 extends E1
     long l;
 }
 
-exception notify /* Test keyword escape. */
```
InsertCreativityHere, line 28: `notify` is not a keyword, but it is a 'special' identifier, defined on every Object.
Here, there _should_ be no problem with it as an exception name though.

cpp/src/slice2java/Main.cpp
```diff
@@ -264,7 +264,7 @@ compile(const vector<string>& argv)
                 {
                     try
                     {
-                        Gen gen(argv[0], icecpp->getBaseName(), includePaths, output);
```
InsertCreativityHere, line 267: The executable name we were never using...

cpp/src/slice2java/Gen.cpp
```diff
@@ -1474,22 +1446,17 @@ Slice::JavaVisitor::writeDispatch(Output& out, const InterfaceDefPtr& p)
...
         {
             out << nl << "request.inputStream.skipEmptyEncapsulation();";
         }
 
-        vector<string> inArgs;
-        for (const auto& pli : inParams)
-        {
-            inArgs.push_back("iceP_" + pli->name());
-        }
-
```
InsertCreativityHere, line 1492: We already had a helper function for getting this list: `getInArgs`.

InsertCreativityHere (commented):
Also, just for the record, `slice2java` only performs case-correction for niche things we generate. Specifically:
- <OP_NAME>MarshaledResult
- <OP_NAME>Result

- set<OPTIONAL_MEMBER_NAME>
- get<OPTIONAL_MEMBER_NAME>
- has<OPTIONAL_MEMBER_NAME>
- clear<OPTIONAL_MEMBER_NAME>
- optional<OPTIONAL_MEMBER_NAME>
- is<BOOL_MEMBER_NAME>

- set<INDEXABLE_MEMBER_NAME>
- get<INDEXABLE_MEMBER_NAME>

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
bernardnormier: > if a user has a class/exception that collides with a Java keyword

Nobody has such a class or exception, so this is definitely not a concern.

cpp/src/slice2java/JavaUtil.h
```diff
@@ -114,11 +104,7 @@ namespace Slice
...
+        [[nodiscard]] static std::string getUnqualified(const ContainedPtr&, const std::string& = std::string());
```
bernardnormier, line 107: Always good to add parameter names to signatures when you update the code.

cpp/src/slice2java/Gen.h
bernardnormier: I find the comments rather confusing. Compose the parameter lists vs compose the argument lists??

java/test/src/main/java/test/Ice/adapterDeactivation/Test.ice
bernardnormier: I don't like _transient: leading underscore is for private fields. I suggest `transient_`.

java/test/src/main/java/test/Ice/packagemd/NoPackage.ice
bernardnormier: What are we testing here? If a test is redundant, it should be deleted.

java/test/src/main/java/test/Slice/escape/Key.ice
```diff
@@ -4,71 +4,71 @@
 
 [["java:package:test.Slice.escape"]]
 
-module abstract
+// TODO: figure out a better way to map module names.
+module java_abstract
 {
 
+["java:identifier:_assert"]
```
bernardnormier, line 11: I would pick somewhat recommendable remapping. identifiers with leading underscores or prefixed by java_ are not reasonable recommendations.

bernardnormier (approved):

cpp/src/slice2java/Gen.cpp
```diff
@@ -1474,22 +1446,17 @@ Slice::JavaVisitor::writeDispatch(Output& out, const InterfaceDefPtr& p)
 
             for (const auto& value : values)
             {
-                const string typeS =
+                const string ts =
```
pepone, line 1449: I would remove `ts` variable, and use typeToString directly. The variable is only used once and doesn't make the code any clear.

cpp/src/slice2java/Gen.cpp
```diff
@@ -2148,11 +2100,13 @@ Slice::JavaVisitor::writeServantDocComment(Output& out, const OperationPtr& p, c
         out << nl << " * @return A completion stage that the servant will complete when the invocation completes.";
     }
 
-    map<string, StringList> exDocs = dc->exceptions();
-    for (const auto& exDoc : exDocs)
+    for (const auto& [name, lines] : dc->exceptions())
```
pepone, line 2103: We have very similar code above, we should consider refactor to a function in a follow-up PR.

cpp/src/slice2java/Gen.cpp
```diff
@@ -2343,13 +2296,13 @@ Slice::Gen::TypesVisitor::visitClassDefStart(const ClassDefPtr& p)
                 //
                 // Generate a constructor accepting parameters for just the required members.
                 //
-                out << sp << nl << "public " << fixKwd(name) << spar;
+                out << sp << nl << "public " << name << spar;
                 vector<string> parameters;
                 for (const auto& allDataMember : allDataMembers)
```
pepone, line 2301: not related to the PR, but `allDataMember` seems an awful name for the iteration variable.

pepone (approved):

cpp/src/slice2java/Gen.h
InsertCreativityHere: While unrelated, I cleaned up much this code, and fixed the doc-comments accordingly.

cpp/src/slice2java/JavaUtil.h
InsertCreativityHere: I added parameter names to all the functions in this area of the code, including this one.

java/test/src/main/java/test/Slice/escape/Key.ice
InsertCreativityHere: Typically I agree with this, but here we're talking about the `escape` test.
There isn't a single line of this file that is "reasonable" or "recommendable".

You could argue the whole point of this test is to make sure even the most insane edge-case names still work correctly.

Hopefully the `java_` prefixes are temporary, but we have no way to remap modules, classes, or exceptions right now.
And Slice forbids leading underscores. Sure, we could pick some totally unrelated name, but this would lose the intent.
These are _supposed_ to be tested, we just lack the ability to do so right now.

java/test/src/main/java/test/Ice/adapterDeactivation/Test.ice
bernardnormier: Just pick a different name then, such as `transientOp`.

java/test/src/main/java/test/Slice/escape/Key.ice
bernardnormier: I'd suggest to add an `escaped` for readability, as in `escapedAssert`.

It's weird to have java-prefixed identifiers in Java. Your language is Java, the identifiers are obviously Java identifiers.

java/test/src/main/java/test/Ice/packagemd/NoPackage.ice
InsertCreativityHere: It's like I said, it seems like we're testing that you can throw and catch exceptions that have been re-mapped with `java:package`. If you're asking for the specific code snippet, here:
```
            try {
                initial.throwTest1Notify();
                test(false);
            } catch (notify ex) {
                // Expected
            }
```
`throwTest1Notify` throws a `notify` exception, and then we catch it.

After this we have tests that seem to be testing slicing in this context as well, with these 2 other operations:
```
    void throwTest1E2AsE1();
    void throwTest1E2AsE2();
```
But these seem like more advanced cases, whereas `notify` is testing a 'base case'.

java/test/src/main/java/test/Ice/adapterDeactivation/Test.ice
InsertCreativityHere: Fine, I changed it to `transientOp` in the places where it was absolutely necessary.
Is this our new escaping convention then? We suffix the illegal identifier with it's type?

ie. ` struct while { ... }` is mapped to `whileStruct`?

InsertCreativityHere:
In the course of adding doc-comments, parameter names, and cleaning up code,
this PR also performs significant cleanup of `slice2java` in addition to implementing `java:identifier`.

java/test/src/main/java/test/Slice/escape/Key.ice
InsertCreativityHere: I updated all the escaped named in Java, C++, and C# to use this new convention.
Our preferred escaping prefix is `escaped` now.

Obviously I'm only talking about places where we can't use a built-in escaping thing like `@` in C#.

PR #3537, opened 2025-02-13
Nightly Release Workflow Fixes

pepone:


.github/actions/setup-dependencies/action.yml
```diff
@@ -31,6 +31,7 @@ runs:
     - uses: ruby/setup-ruby@v1
       with:
         ruby-version: "3.3"
+      if: runner.os != 'Windows'
```
pepone, line 34: We don't need to install Ruby on Windows.

.github/workflows/build-msi-package.yml
```diff
@@ -25,6 +25,8 @@ jobs:
       - name: Build MSI
         run: dotnet build -c Release -p:Platform=x64
         working-directory: packaging/msi
+        env:
+          VCInstallDir: "C:\\Program Files\\Microsoft Visual Studio\\2022\\Enterprise\\VC"
```
pepone, line 29: With Visual Studio 2022 merge modules are installed in the VCInstallDir. Unfortunately this variable is not set in the GitHub actions shell.

.github/workflows/build-nightly-release-workflow.yml
```diff
@@ -32,20 +32,20 @@ jobs:
...
           echo "npm_version=${VERSION}-nightly.${DATE}.${BUILD}" >> $GITHUB_OUTPUT
-          # - NuGet: 3.8.0-nightly.YYYYMMDD.BUILD
+          # - NuGet: 3.8.0-nightly.<YYYYMMDD>.<BUILD>
           echo "nuget_version=${VERSION}-nightly.${DATE}.${BUILD}" >> $GITHUB_OUTPUT
-          # - PyPI: 3.8.0.devYYYYMMDD.BUILD
-          echo "pypi_version=${VERSION}.dev${DATE}.${BUILD}" >> $GITHUB_OUTPUT
-          # - RPM: 3.8.0-0.nightlyYYYYMMDD.BUILD
-          echo "rpm_version=${VERSION}-0.nightly${DATE}.${BUILD}" >> $GITHUB_OUTPUT
+          # - PyPI: 3.8.0.dev<YYYYMMDD><BUILD>
+          echo "pypi_version=${VERSION}.dev${DATE}${BUILD}" >> $GITHUB_OUTPUT
```
pepone, line 46: The previous version was not valid for a python package according to pep440

.github/workflows/build-nightly-release-workflow.yml
```diff
@@ -32,20 +32,20 @@ jobs:
...
+          # - RPM: 3.8.0~0.nightly<YYYYMMDD>.<BUILD>
+          echo "rpm_version=${VERSION}~0.nightly${DATE}.${BUILD}" >> $GITHUB_OUTPUT
```
pepone, line 48: `-` is not valid for the RPM version, replaced with `~`.

Another option would be to keep `3.8.0` as the version and use `nightly<YYYYMMDD>.<BUILD>` as the Release field.

.github/workflows/build-npm-packages.yml
```diff
@@ -104,8 +104,8 @@ jobs:
         working-directory: ice/js
 
       - name: Update Version
-        run: |
-          sed -i 's/"version":.*/"version": "${{ inputs.ice_version }}"/' package.json
+        working-directory: ice/js
+        run: npm version ${{ inputs.ice_version }}
```
pepone, line 108: The sed command was breaking the JSON file, better use `npm version` to update the version.

.github/workflows/build-pip-packages.yml
```diff
@@ -45,10 +45,10 @@ jobs:
         run: |
           python3 -m pip install --upgrade pip
           python3 -m pip install build
+          python3 -m pip install toml
 
       - name: Update PYPI Version
-        working-directory: python
-        run: sed -i 's/version = .*/version = "${{ inputs.ice_version }}"/' pyproject.toml
+        run: python3 packaging/pypi/update_version.py ${{ inputs.ice_version }} python/pyproject.toml
```
pepone, line 51: Added a small script that uses toml to parse and update the project version in the project file. This works better here as it runs in both Windows/macOS runners.

cpp/doxygen/examples/Ice/Makefile.mk
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) ZeroC, Inc.
 
 $(project)_libraries    = IceExamples
+$(project)_caninstall   := no
```
pepone, line 4: This is to prevent libExamples being install by make install

csharp/msbuild/ice.proj
```diff
@@ -8,6 +8,9 @@
     </PropertyGroup>
 
     <Import Project="$(MSBuildThisFileDirectory)\..\..\config\icebuilder.props" />
```
pepone, line 10: Updated this project to allow overwriting the package version we build.

packaging/dpkg/build-package.sh
```diff
@@ -15,14 +15,17 @@ else
     ICE_VERSION=$(dpkg-parsechangelog --file /workspace/build/debian/changelog --show-field Version)
 fi
 
-# Generate a tarball of the current repository state for the given ICE_VERSION
-echo "Creating tarball for ICE_VERSION=$ICE_VERSION"
+UPSTREAM_VERSION=$(echo $ICE_VERSION | cut -f1 -d'-')
```
pepone, line 18: dpkg-buildpackage expect the archive to use the upstream version (3.8.0) and not the full version in the changelog.

.github/workflows/build-msi-package.yml
externl: Should we put a comment on it? Would be nice to be able to computer this variable.

packaging/rpm/glacier2router.service
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) ZeroC, Inc.
```
pepone, line 1: missing from the old package.

packaging/rpm/ice.spec
```diff
@@ -120,14 +113,15 @@ Group: System Environment/Libraries
...
 %ifarch %{_host_cpu}
+Requires: %{?nameprefix}dsnode%{?_isa} = %{version}-%{release}
 Requires: %{?nameprefix}glacier2%{?_isa} = %{version}-%{release}
 Requires: %{?nameprefix}icegrid%{?_isa} = %{version}-%{release}
 Requires: %{?nameprefix}icebridge%{?_isa} = %{version}-%{release}
 Requires: php-%{?nameprefix}ice%{?_isa} = %{version}-%{release}
 Requires: python3-%{?nameprefix}ice%{?_isa} = %{version}-%{release}
 Requires: lib%{?nameprefix}ice3.8-c++%{?_isa} = %{version}-%{release}
 Requires: %{?nameprefix}icegridgui = %{version}-%{release}
-%endif # %{_host_cpu}
```
pepone, line 130: Removed comments after `%endif` because they result in a warning when you build the package.

externl (approved):

.github/workflows/build-msi-package.yml
pepone: Created https://github.com/zeroc-ice/ice/issues/3538

PR #3543, opened 2025-02-13
Fix handling of empty default string value

bernardnormier:
This PR updates slice2cpp to handle properly `= ""`. It also updates the Slice parser and the Slice compilers to use a clearer representation for default values.

The motivation was to get slice2cpp to generate the following for Ice::Identity:
```cpp
        /// The name of the Ice object.
        std::string name;
        /// The Ice object category.
        std::string category{};
```

This allows us to create an Ice::Identity directly without an explicit category and without initializer warning:
```cpp
Identity identity{.name = "greeter"}; // no warning with gcc because category has an explicit default
```
(I verified it actually works!)


cpp/src/Slice/Parser.cpp
```diff
@@ -2433,8 +2433,8 @@ Slice::ClassDef::createDataMember(
     const TypePtr& type,
     bool isOptional,
     int tag,
-    const SyntaxTreeBasePtr& defaultValueType,
-    const string& defaultValueString)
+    SyntaxTreeBasePtr defaultValueType,
```
bernardnormier, line 2436: All this duplicated code in the 3 createDataMember functions is unfortunate. I didn't fix it.

cpp/src/Slice/Parser.cpp
```diff
@@ -2485,17 +2485,14 @@ Slice::ClassDef::createDataMember(
...
 
-    SyntaxTreeBasePtr dlt = defaultValueType;
-    string dv = defaultValueString;
-
-    if (dlt || (dynamic_pointer_cast<Enum>(type) && !dv.empty()))
+    if (defaultValueString)
     {
         // Validate the default value.
-        if (!validateConstant(name, type, dlt, dv, false))
+        if (!validateConstant(name, type, defaultValueType, *defaultValueString, false))
```
bernardnormier, line 2491: defaultValueType is "in-out". validateConstant can replace a nullptr by an non-null EnumeratorPtr.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -1101,7 +1100,7 @@ SwiftGenerator::writeMembers(
                     out,
                     type,
                     member->defaultValueType(),
-                    defaultValue,
+                    member->defaultValue().value_or(""), // TODO: revisit this code
```
bernardnormier, line 1103: Most of the code in this function seems to be about auto-escaping to avoid collision ... and we're reworking this code already.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -282,7 +282,7 @@ Slice::CsGenerator::isValueType(const TypePtr& type)
         DataMemberList dm = s->dataMembers();
         for (const auto& i : dm)
         {
-            if (!isValueType(i->type()) || i->defaultValueType())
+            if (!isValueType(i->type()) || i->defaultValue())
```
bernardnormier, line 285: This does not change anything. I think it's just preferable to consistently check defaultValue() or defaultValueType()... and now all the code checks defaultValue() to see if there is a default value.

cpp/src/slice2php/Main.cpp
```diff
@@ -1147,9 +1147,9 @@ CodeVisitor::writeConstructorParams(const MemberInfoList& members)
         _out << '$' << p->fixedName << "=";
 
         const DataMemberPtr member = p->dataMember;
-        if (member->defaultValueType())
+        if (member->defaultValue())
```
bernardnormier, line 1150: No change, since they are either both null or both non-null.

externl (approved):

pepone (approved):

PR #3561, opened 2025-02-18
Updates to rename "ice" to "@zeroc/ice" scoped package

pepone:
This PR renames "ice" NPM module to "@zeroc/ice" scoped modules.

Using a scoped module has the advantage of identify the organization that published it in the NPM registry.

This is similar to NuGet reserved prefixes we have for `zeroc.` and `IceRpc.`.

npm also allow to configure per scope registries witch is very convenient for testing.

For uses upgrading from 3.7 it means using:

```
import { Ice } from "@zeroc/ice"
``` 

instead of 

```
import { Ice } from "ice";
```

Also update the package.json dependencies to use `@zeroc/ice` instead of `ice`.

The PR also fix slice2js to support scoped modules with `js:module` metadata, the previous version would include `@` and `/` in the import alias which are not valid characters for JavaScript variables.

### DIFF
--- a/js/src/slice2js.js
+++ b/js/src/slice2js.js
@@ -1,5 +1,3 @@
-#!/usr/bin/env node
-
 // Copyright (c) ZeroC, Inc.
 
 import { spawn } from "child_process";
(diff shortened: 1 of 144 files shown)


js/src/slice2js.js
```diff
@@ -60,17 +58,6 @@ export async function compile(args = [], options = {}) {
     });
 }
 
-// If executed directly via `npx slice2js`
```
pepone, line 63: Moved to a separate script and remove the meta.url check.

externl (approved):

bernardnormier (approved):

PR #3563, opened 2025-02-18
Add brew bottle packaging

externl:


packaging/brew/ice.rb
```diff
@@ -0,0 +1,77 @@
...
+        "MCPP_HOME=#{Formula["mcpp"].opt_prefix}",
+        "LMDB_HOME=#{Formula["lmdb"].opt_prefix}",
+        "CONFIGS=all",
+        "PLATFORMS=all",
+        "LANGUAGES=cpp",
+      ]
+      system "make", "install", *args
+
+      (libexec/"bin").mkpath
+      %w[slice2py slice2rb slice2js].each do |r|
```
pepone, line 26: Can we remove those? For these the compiler are in the respective pip/gem/npm package.

packaging/brew/build-bottle.sh
```diff
@@ -0,0 +1,28 @@
...
+
+tap_dir=$(brew --repo zeroc-ice/packaging)
+archive_path=$tap_dir/ice.tar.gz
+
+# Create archive of the current git repository
+cd "$(git rev-parse --show-toplevel)"
+git archive --format=tar.gz -o "$archive_path" HEAD
+archive_hash=$(shasum -a 256 "$archive_path" | cut -d ' ' -f 1)
+
+export ICE_VERSION=${ICE_VERSION:-3.8.0}
```
pepone, line 20: I wonder if is better to fail if not set. We should be explicit about what version to build.

pepone (approved):

packaging/brew/ice.rb
externl: I guess so.

packaging/brew/build-bottle.sh
externl: Ok, I've changed this for now, but what's is there any difference in this vs what we're doing for dpkg by reading from a file is unset.

pepone: You are right, not really different.

The caveat with debian is that we need to update the changelog. For the release version we don't want the changelog entry to be an automatic update.

PR #3566, opened 2025-02-18
Schedule a Nightly Release Build every day at 2:00 AM

pepone:


.github/workflows/nightly-release-scheduler.yml
externl, line 1: Is this additional workflow necessary?

externl (approved):

.github/workflows/nightly-release-scheduler.yml
pepone: It is for the scheduled workflow to have the default parameters.

PR #3568, opened 2025-02-18
Add Support for `swift:identifier` Metadata

InsertCreativityHere:
This PR adds support for `swift:identifier`, just like we did in the other languages so far.
However, there are some caveats to Swift.

1) Most languages will happily tolerate 'over-escaping'. But there are some contexts in which the Swift compiler will error out if you use escaping. Mostly around argument labels, and how you call functions. In these contexts we always need to remove escaping.

2) The preferred way to escape identifiers in Swift is to wrap it in backticks: https://docs.swift.org/swift-book/documentation/the-swift-programming-language/lexicalstructure/#Identifiers. However, this causes problems in the places where we want to add prefixes like `iceP_` or suffixes like `Prx`. We love adding prefixes and suffixes, so there's lots of places where we need to strip escaping from for this to work.

3) Swift does not want escaped identifiers in doc-comments, in fact, in doc-comments, wrapping something in backticks doesn't escape it, but instead tries to generate a doc-link. So, whenever we're writing a name into a doc-comment, we also need to be stripping escaping off the identifier.

So, I added a `removeEscaping` method which checks for and removes this escaping when necessary.
And we have to call it in lots of places.

### DIFF
--- a/cpp/src/IceStorm/Election.ice
+++ b/cpp/src/IceStorm/Election.ice
@@ -36,24 +36,21 @@ module IceStormElection
         /// Initialize the observer.
         /// @param llu The last log update seen by the master.
         /// @param content The topic content.
-        /// @throws ObserverInconsistencyException Raised if an
-        /// inconsistency was detected.
+        /// @throws ObserverInconsistencyException Raised if an inconsistency was detected.
         void init(LogUpdate llu, TopicContentSeq content)
             throws ObserverInconsistencyException;
 
         /// Create the topic with the given name.
         /// @param llu The log update token.
         /// @param name The topic name.
-        /// @throws ObserverInconsistencyException Raised if an
-        /// inconsistency was detected.
+        /// @throws ObserverInconsistencyException Raised if an inconsistency was detected.
         void createTopic(LogUpdate llu, string name)
             throws ObserverInconsistencyException;
 
         /// Destroy the topic with the given name.
         /// @param llu The log update token.
         /// @param name The topic name.
-        /// @throws ObserverInconsistencyException Raised if an
-        /// inconsistency was detected.
+        /// @throws ObserverInconsistencyException Raised if an inconsistency was detected.
         void destroyTopic(LogUpdate llu, string name)
             throws ObserverInconsistencyException;
 
@@ -61,8 +58,7 @@ module IceStormElection
         /// @param llu The log update token.
         /// @param topic The topic name to which to add the subscriber.
         /// @param record The subscriber information.
-        /// @throws ObserverInconsistencyException Raised if an
-        /// inconsistency was detected.
+        /// @throws ObserverInconsistencyException Raised if an inconsistency was detected.
         void addSubscriber(LogUpdate llu, string topic, IceStorm::SubscriberRecord record)
             throws ObserverInconsistencyException;
 
--- a/cpp/src/IceStorm/Election.ice
+++ b/cpp/src/IceStorm/Election.ice
@@ -37,7 +37,8 @@ module IceStormElection
         /// @param llu The last log update seen by the master.
         /// @param content The topic content.
         /// @throws ObserverInconsistencyException Raised if an
-        /// inconsisency was detected.
+        /// inconsistency was detected.
+        ["swift:identifier:initialize"]
         void init(LogUpdate llu, TopicContentSeq content)
             throws ObserverInconsistencyException;
 
@@ -45,7 +46,7 @@ module IceStormElection
         /// @param llu The log update token.
         /// @param name The topic name.
         /// @throws ObserverInconsistencyException Raised if an
-        /// inconsisency was detected.
+        /// inconsistency was detected.
         void createTopic(LogUpdate llu, string name)
             throws ObserverInconsistencyException;
 
@@ -53,7 +54,7 @@ module IceStormElection
         /// @param llu The log update token.
         /// @param name The topic name.
         /// @throws ObserverInconsistencyException Raised if an
-        /// inconsisency was detected.
+        /// inconsistency was detected.
         void destroyTopic(LogUpdate llu, string name)
             throws ObserverInconsistencyException;
 
@@ -62,7 +63,7 @@ module IceStormElection
         /// @param topic The topic name to which to add the subscriber.
         /// @param record The subscriber information.
         /// @throws ObserverInconsistencyException Raised if an
-        /// inconsisency was detected.
+        /// inconsistency was detected.
         void addSubscriber(LogUpdate llu, string topic, IceStorm::SubscriberRecord record)
             throws ObserverInconsistencyException;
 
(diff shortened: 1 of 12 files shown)


cpp/src/IceStorm/Election.ice
```diff
@@ -37,23 +37,24 @@ module IceStormElection
         /// @param llu The last log update seen by the master.
         /// @param content The topic content.
         /// @throws ObserverInconsistencyException Raised if an
-        /// inconsisency was detected.
+        /// inconsistency was detected.
+        ["swift:identifier:`init`"]
```
bernardnormier, line 41: I would remap it to initialize; I find `init` very confusing.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -2827,6 +2827,7 @@ Slice::Gen::InterfaceVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
     const string scoped = p->mappedScoped();
     const string name = p->mappedName();
 
+    // TODO this seems like a bug. We don't generate 'ice_id' etc for operation-less interfaces?
```
bernardnormier, line 2830: This looks correct to me.

Note that this code generates dispatch, not ice_id etc. ice_id etc. are implemented by `ObjectI<T>`.

cpp/src/slice2swift/Gen.cpp
```diff
@@ -267,25 +267,13 @@ Gen::TypesVisitor::visitClassDefStart(const ClassDefPtr& p)
...
-    vector<string> parts = splitScopedName(p->scoped());
-    for (auto it = parts.begin(); it != parts.end();)
-    {
-        factory << (*it);
-        if (++it != parts.end())
-        {
-            factory << "_";
-        }
-    }
+    const string factoryName = removeEscaping(prefix + p->mappedScoped("_").substr(1));
```
bernardnormier, line 271: This does not look correct.

The whole point of this "factory name" is to allow the Swift runtime to resolve a Slice type ID into a mapped class. So you can't use the mapped name in this factoryName!

scripts/Controller.ice
```diff
@@ -8,6 +8,7 @@ sequence<string> StringSeq;
 
 class Config
 {
+    ["swift:identifier:`protocol`"]
```
bernardnormier, line 11: I would only use keyword-in-ticks in actual tests.

Here, I'd remap it to `transportProtocol`.

slice/Ice/RemoteLogger.ice
```diff
@@ -64,6 +64,7 @@ module Ice
         /// init is called by attachRemoteLogger when a RemoteLogger proxy is attached.
         /// @param prefix The prefix of the associated local Logger.
         /// @param logMessages Old log messages generated before "now".
+        ["swift:identifier:`init`"]
```
bernardnormier, line 67: initialize

InsertCreativityHere: Also changed to `initialize`.

scripts/Controller.ice
InsertCreativityHere: Alright. Done.

cpp/src/slice2cpp/Gen.cpp
InsertCreativityHere: Okay yes, I forgot that we would inherit the dispatch here.
Good to know!

cpp/src/IceStorm/Election.ice
InsertCreativityHere: Sure, changed to `initialize`.

swift/test/Ice/admin/AllTests.swift
```diff
@@ -16,7 +16,7 @@ class RemoteLoggerI: Ice.RemoteLogger {
         _helper = helper
     }
 
-    func `init`(prefix: String, logMessages: [Ice.LogMessage], current _: Ice.Current) async throws {
```
InsertCreativityHere, line 19: We have a test class which implements the `RemoteLogger` generated from `Ice/RemoteLogger.ice`.

The name in that file was remapped from `init` to `initialize`, and so had to be updated here.
Any other Swift code which was using this API will also need to be updated now.

cpp/src/IceStorm/Election.ice
```diff
@@ -37,23 +37,24 @@ module IceStormElection
         /// @param llu The last log update seen by the master.
         /// @param content The topic content.
         /// @throws ObserverInconsistencyException Raised if an
-        /// inconsisency was detected.
+        /// inconsistency was detected.
+        ["swift:identifier:initialize"]
```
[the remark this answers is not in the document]
InsertCreativityHere: I honestly don't know. I just used a regex to look through for all swift keywords in any `.ice` files.
If Swift currently does not use this file and never will, I can remove these.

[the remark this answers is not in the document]
externl: It shouldn't compile any Slice files in cpp/src/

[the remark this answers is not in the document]
bernardnormier: All the Slice files in `cpp/src` are C++-only.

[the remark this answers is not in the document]
InsertCreativityHere: Alright, I removed the `swift:identifier` from this file.
This was the only file in `cpp/src` that I touched.

scripts/Controller.ice
```diff
@@ -8,6 +8,7 @@ sequence<string> StringSeq;
 
 class Config
 {
+    ["swift:identifier:transportProtocol"]
```
bernardnormier, line 11: Is all this code Python-only? I'd rather avoid confusing metadata that suggests it's used in Swift.

cpp/src/slice2swift/Gen.cpp
```diff
@@ -267,9 +265,7 @@ Gen::TypesVisitor::visitClassDefStart(const ClassDefPtr& p)
         out << eb;
     }
 
-    //
-    // For each Value class we generate an extension method in ClassResolver
-    //
+    // For each Value class we generate an extension method in ClassResolver.
```
bernardnormier, line 268: I would update the comment to stress this name is derived from the type ID, not the mapped name.

bernardnormier: If we have no test for this feature (unmarshaling a remapped class), we need to add one.

bernardnormier (approved):

scripts/Controller.ice
pepone: The controller is implemented by iOS and Android test controller applications. With iOS we have one for C++ and one for Swift. And there is a Java implementation for android.

InsertCreativityHere: Indeed, I forgot to update the Swift iOS test code. But I've updated it now.

Even though it looks like iOS just completely ignores this stuff, we do still compile the generated code, which means that we do need to remap these identifiers.

cpp/src/slice2swift/Gen.cpp
InsertCreativityHere: I added the requested comment, and opened an issue.
This isn't specific to Swift, we need to add/augment these tests in all languages.

PR #3571, opened 2025-02-18
Fix for DEB package uploads

pepone:
See https://help.sonatype.com/en/apt-repositories.html#deploying-packages-to-hosted-apt-repositories


externl (approved):
What does `@./` mean here?

packaging/dpkg/publish-packages.sh
```diff
@@ -16,7 +16,7 @@ find . -maxdepth 1 -type f \( -name "*.deb" -o -name "*.ddeb" \) | while read -r
     echo "Uploading $file to Nexus..."
     curl -u "${REPOSITORY_USERNAME}:${REPOSITORY_PASSWORD}" \
          -H "Content-Type: multipart/form-data" \
-         --data-binary "$file" \
+         --data-binary "@./$file" \
```
pepone, line 19:
```suggestion
         --data-binary "@${file}" \
```

pepone:
> What does @./ mean here?

`.` is actually not require is should be `@<filename>` to tell curl to read data from the file.

See https://curl.se/docs/manpage.html#--data-binary

PR #3572, opened 2025-02-18
Define ReplyStatus as a Slice enum

bernardnormier:
This PR (re)defines Ice::ReplyStatus as a Slice enum.

Using a Slice enum helps with consistency across languages. It's also a enum that we marshal into all reply messages.

The range of this enum is 0 (Ok) to 126, because it's encoded with the Ice 1.0 encoding, and with the 1.0 encoding, only enumerators <= 126 are encoded on a single byte.

The Swift update is temporary: I could not switch to the generated code because `swift:identifier` is not yet implemented. Furthermore, our Slice compilation system does not allow to filter-out files. So for now, Swift uses TempReplyStatus as its ReplyStatus.


cpp/src/Slice/Util.h
```diff
@@ -66,7 +66,10 @@ namespace Slice
 
     bool isProxyType(const TypePtr& type);
 
+    // Is this the first element defined in its container?
+    bool isFirstElement(const ContainedPtr& p);
+
     // Is this the first operation defined in the interface?
-    bool isFirstOperation(const OperationPtr& p);
+    inline bool isFirstOperation(const OperationPtr& p) { return isFirstElement(p); }
```
InsertCreativityHere, line 73: Should we really even keep this, instead of just calling `isFirstElement` directly?
The only things interfaces can contain are `Operation`s.

Having both, subtly implies you could have an operation which is the first 'operation' in an interface, but not the first 'element'.

CHANGELOG-3.8.md
```diff
@@ -160,7 +170,6 @@ classDiagram
     UnknownLocalException : replyStatus = UnknownLocalException
     UnknownException <|-- UnknownUserException
     UnknownUserException : replyStatus = UnknownUserException
-```
```
InsertCreativityHere, line 163: Looks like an accidental removal?

CHANGELOG-3.8.md
```diff
@@ -131,13 +131,23 @@ registered with the Ice Locator (typically the IceGrid registry).
   - Add new _AdapterName_.PublishedHost property, used to compute the default published endpoints.
   - Removed the `refreshPublishedEndpoints` operation on `ObjectAdapter`.
 
-- The local exceptions that can be marshaled are no longer limited to 6 exceptions. These exceptions derive now from
-  class `DispatchException`. DispatchException holds the reply status, which can have any value between 2 and 255, and
-  a message (string). A dispatch exception with reply status >= 5 is marshaled as its reply status (one byte) followed
-  by its message (a Slice-encoded string).
+- The local exceptions that can be marshaled have now a common base class, `DispatchException`, and are no longer
```
InsertCreativityHere, line 134:
```suggestion
- The local exceptions that can be marshaled now have a common base class (`DispatchException`), and are no longer
```

InsertCreativityHere (approved):
Looks good to me!

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Current.java
```diff
@@ -251,7 +251,7 @@ private OutgoingResponse createOutgoingResponseCore(Throwable exc) {
             replyStatus = ReplyStatus.UserException;
 
             if (requestId != 0) {
-                ostr.writeByte(replyStatus.value());
+                ostr.writeByte((byte) replyStatus.value());
```
pepone, line 254: Shouldn't we use the generated code to marshal this?

pepone (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Current.java
bernardnormier: I'll rework this code when I implement DispatchException in Java.

Then, because enums are "checked" in Java, we can't use the generated code to marshal/unmarshal any ReplyStatus in Java. We can use the generated code to marshal known reply statuses.

CHANGELOG-3.8.md
bernardnormier: Fixed

cpp/src/Slice/Util.h
bernardnormier: Fixed.

PR #3574, opened 2025-02-19
Replace Ice.Long with BigInt

### LINKED ISSUE #3564: Replace JavaScript Ice.Long with BigInt
The mapping for Slice long in JavaScript uses a custom Ice.Long class. The generated code also accepts number for convenience and it can be used when the value falls with in the valid range for JS number.

JavaScript has now BigInt https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

I think we better switch to this new standard type.

pepone:
This PR updates the JavaScript mapping for Slice long to use BigInt. For input parameters we still accept both number and BigInt. This PR removes Ice.Log that was previously used to encode Slice long as two 32 bit words.

Fix #3564


externl (approved):
Looks good to me!

bernardnormier (approved):

PR #3576, opened 2025-02-19
Update ReplyStatus range

bernardnormier:
This PR updates the ReplyStatus range to be 0..255 instead of 0..126.

The 126 limit is (was) due to a constraint with the 1.0 encoding. It assumed we could marshal/unmarshal reply statuses using the generated code... but it is actually doable only in C++ and C#, where native enums are "unchecked". In Java, Swift, Python (...), native enums are checked and the generated code is unusable for unknown enumerators.

As of this PR:
 - we use the generated code only to marshal known reply status enumerators (in C++ and C# for now, eventually in all languages)
 - we don't use the generated code to unmarshal reply status (any byte is ok)
 - in C++ and C#, where the native ReplyStatus is unchecked, we use ReplyStatus to hold the reply status in DispatchException and OutgoingResponse. We won't be able to do that in other languages where the mapped enum is checked.

### DIFF
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -132,7 +132,7 @@ registered with the Ice Locator (typically the IceGrid registry).
   - Removed the `refreshPublishedEndpoints` operation on `ObjectAdapter`.
 
 - The local exceptions that can be marshaled now have a common base class (`DispatchException`), and are no longer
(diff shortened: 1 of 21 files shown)


externl (approved):

InsertCreativityHere (approved):
Looks good, it's unfortunate that we need to do this though.

PR #3577, opened 2025-02-19
Add dispatch exception to Python

bernardnormier:
This PR adds dispatch exception to Python and adds range-check to the DispatchException constructors in C++, C# and Python.

It includes #3576.


python/python/Ice/LocalExceptions.py
```diff
@@ -1,24 +1,55 @@
...
+
+    @property
+    def replyStatus(self):
+        """
+        Gets the reply status of this exception.
+
+        Returns
+        -------
+        int
+            The reply status, as an int in the range 2..126.
```
pepone, line 41:
```suggestion
            The reply status, as an int in the range 2..255.
```

pepone (approved):

externl (approved):

PR #3582, opened 2025-02-19
Add dispatch exception to Ruby, PHP and MATLAB

bernardnormier:
Dispatch exceptions not thrown by Ruby/PHP/MATLAB code in these mappings, with just one exception: the MATLAB generated code can throw UnknownUserException.


cpp/src/slice2matlab/Main.cpp
```diff
@@ -2995,8 +2995,14 @@ CodeVisitor::visitEnum(const EnumPtr& p)
     out.inc();
     for (const auto& enumerator : enumerators)
     {
+        out << nl; // TODO: rework this code
```
bernardnormier, line 2998: All the MATLAB doc-comment generation needs refactoring.

matlab/test/Ice/exceptions/Test.ice
```diff
@@ -6,84 +6,84 @@
 
 module Test
 {
+    interface Empty
```
bernardnormier, line 9: Reformatting

pepone (approved):

externl (approved):

PR #3586, opened 2025-02-20
Fix PIP package to include slice2py script

pepone:
slice2py was missing from the new PIP package, this PR brings it back.

### DIFF
--- a/python/dist/lib/Glacier2/.gitignore
+++ b/python/dist/lib/Glacier2/.gitignore
@@ -1,1 +1,0 @@
-# Empty directory populated by sdist build
--- a/python/dist/lib/slice2py.py
+++ b/python/dist/lib/slice2py.py
@@ -13,7 +13,7 @@ def main():
     if sliceDir is not None:
         sys.argv.append("-I" + sliceDir)
     val = IcePy.compile(sys.argv)
-    sys.exit(int(val))
+    return int(val)
 
 
 if __name__ == "__main__":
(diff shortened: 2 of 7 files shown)


python/dist/lib/Glacier2/.gitignore
pepone, line 1: The dist/lib directories are required for building pip package directly from git. But only for the Python packages we ship "Ice" and "IceMX"

python/dist/lib/slice2py.py
```diff
@@ -13,7 +13,7 @@ def main():
     if sliceDir is not None:
         sys.argv.append("-I" + sliceDir)
     val = IcePy.compile(sys.argv)
-    sys.exit(int(val))
+    return int(val)
```
pepone, line 16: we already call exit below.

python/setup.py
```diff
@@ -230,7 +230,11 @@ def pre_build(self):
                     shutil.copy(source_file, target_file)
 
         # Slice sources
+        slice_excludes = ["IceDiscovery", "IceLocatorDiscovery"]
```
pepone, line 233: We don't need to include Slice definitions for these plug-ins.

python/setup.py
```diff
@@ -286,8 +290,6 @@ def run(self):
 
 # Setup configuration for the package
 setup(
-    name='zeroc-ice',
-    version='3.8.0a0',
```
pepone, line 290: We already set this in `pyproject.toml` no need to repeat it here.

externl (approved):

bernardnormier (approved):

Issue #3548, opened 2025-02-14
Proposal: Add implicit DispatchException


externl:
The current version (1.1) of the Ice protocol has no exception extensibility beyond the 6 local exceptions which can be encoded and user exceptions. This makes implementing generic middleware (think rate-limiting, authorization, etc.) painful. 

Users need to define an exception and ensure it's specified in every operation.

**Proposal:**

Add a new implicit user exception: `Ice::DispatchException` (name tbd), probably not defined in Slice so that users can't inherit from it.

Borrowing from IceRPC it could look something like https://github.com/icerpc/icerpc-slice/blob/main/IceRpc/StatusCode.slice.

E.g. (specifics to be worked out)
```slice
enum DispatchStatusCode {
     ...
     Unavailable,
     Unauthorized
     ...
}

exception DispatchException {
    DispatchStatus status;
    optional(1) string message;
}
```

This user exception would always be available in 3.8 and above and would be excluded from the exception specification checks.

bernardnormier:
What's the advantage of not defining `Ice::DispatchException` in Slice and not allowing users to derive from this exception?

> optional(1) string message;

A mandatory string field would be simpler.

bernardnormier:
The main downside I see with this proposal is it breaks the clean separation between "system exceptions" (= 6 local exceptions that can be encoded) and user exceptions.

Today, user exceptions are defined in Slice, and must be listed in the operation's exception specification. On the other hand, system exceptions are not defined in Slice and can be thrown by any invocation.

With your proposal, `Ice::DispatchException` would be a user exception (derived from `Ice::UserException` even if not defined in Slice?) that behaves like a system exception. 

Also, conceptually, user exceptions are "error values", and largely treated as such. For example, the logger middleware and its predecessor in 3.7 treats all user exceptions like successful responses, without logging any warning.

Other APIs also essentially treat Ok and UserException essentially the same way, unlike system exceptions and other local exceptions. For example, ice_invokeAsync on ObjectPrx and Blobject.

## Counterproposal: open-up ReplyStatus and allow enumerator values beyond UnknownException

See https://github.com/icerpc/icerpc-slice/blob/a12f71fff0b90fb94d9e032999754cdb6152cb41/IceRpc/Internal/IceDefinitions.slice#L90

[...]

Issue #3498, opened 2025-02-05
Add Ice.Trace.Dispatch


bernardnormier:
Most of the built-in logging provided by Ice is controlled by `Ice.Trace.Category` properties with numeric values, where the default is 0, mean "don't log anything". Then 1, 2 (...) increase logging verbosity.

See:
https://doc.zeroc.com/ice/3.7/property-reference/ice-trace

Plugins and services offer comparable `<name>.Trace.Category` properties, with the same default (0 = no logging). For example:
https://doc.zeroc.com/ice/3.7/property-reference/icessl#id-.IceSSL.*v3.7-IceSSL.Trace.Security
https://doc.zeroc.com/ice/3.7/property-reference/icegrid#id-.IceGrid.*v3.7-IceGrid.Node.Trace.Activator

This "trace" logging produces "normal" log messages, such as:
```
-- 02/05/25 09:21:42.508 icebox: Network: listening for tcp connections
   local address = 127.0.0.1:9996
```

Another form of logging is controlled by the `Ice.Warn.Category` properties:
https://doc.zeroc.com/ice/3.7/property-reference/ice-warn

The semantics are similar to the Trace properties - in particular the messages are sent to the same Ice logger - except:
 - some Ice.Warn.Category properties have a default of 1 
 - only Ice has Warn properties
 - the log message look a little bit different, for example:
```
-! 02/05/25 09:32:45.159 warning: dispatch exception: CancellationError()
   identity: 5610CD90-5021-4B90-A378-6D11FE980BF1
   facet: 
   operation: tick
   remote host: ::1 remote port: 58455
```

`--` = normal logging
`-!` = warning
`!!` = error

Ice.ThreadPool.name.SizeWarn is another logging property similar to the Ice.Warn logging:
https://doc.zeroc.com/ice/3.7/property-reference/ice-threadpool#Ice.ThreadPool.*-Ice.ThreadPool.name.SizeWarn

## Issue

The dispatch logging, controlled by Ice.Warn.Dispatch and implemented by the built-in logger middleware, is very limited since it logs only exceptions. It would be nicer if it could log all requests - including successful requests - like IceRPC's logger middleware:
https://github.com/icerpc/icerpc-csharp/blob/main/src/IceRpc.Logger/README.md
https://github.com/icerpc/icerpc-csharp/blob/fc251d6252e0135d9312dbac6b46e3dae3b23b49/src/IceRpc.Logger/LoggerMiddleware.cs#L39

[...]

PR #3588, opened 2025-02-20
Add dispatch exception and implement Ice.Trace.Dispatch in JS

bernardnormier:
This PR implements DispatchException in JS: see #3548.

It also implements Ice.Trace.Dispatch in JS: see #3498.

### DIFF
--- a/js/src/Ice/LoggerMiddleware.js
+++ b/js/src/Ice/LoggerMiddleware.js
@@ -17,7 +17,6 @@ export class LoggerMiddleware {
                 case ReplyStatus.Ok:
                 case ReplyStatus.UserException:
                     if (this._traceLevel > 0) {
-                        // TODO: fix once replyStatus is a number
                         this.logDispatch(response.replyStatus, request.current);
                     }
                     break;
@@ -69,7 +68,7 @@ export class LoggerMiddleware {
     logDispatch(replyStatus, current) {
         let output = `dispatch of ${current.operation} to `;
         this.printTarget(output, current);
-        output += ` returned a response with reply status ${replyStatus}`;
+        output += ` returned a response with reply status ${replyStatus.name}`;
         this._logger.trace(this._traceCat, output);
     }
 
--- a/js/src/Ice/LoggerMiddleware.js
+++ b/js/src/Ice/LoggerMiddleware.js
@@ -1,7 +1,8 @@
 // Copyright (c) ZeroC, Inc.
 
 import { UserException } from "./UserException.js";
-import { RequestFailedException } from "./LocalExceptions.js";
+import { DispatchException } from "./LocalExceptions.js";
+import { UnknownException } from "./LocalExceptions.js";
 import { IPConnectionInfo } from "./Connection.js";
 import { identityToString } from "./IdentityToString.js";
 import { StringUtil } from "./StringUtil.js";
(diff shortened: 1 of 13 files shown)


externl (approved):
Looks good to me. Prettier is complaining though.

js/src/Ice/LoggerMiddleware.js
```diff
@@ -15,62 +16,100 @@ export class LoggerMiddleware {
...
+        }
+        this._logger.warning(output);
     }
 
-    warning(exceptionMessage, current) {
-        let output = "dispatch exception:";
-        output += `\nidentity: ${identityToString(current.id, this._toStringMode)}`;
-        output += `\nfacet: ${StringUtil.escapeString(current.facet, "", this._toStringMode)}`;
-        output += `\noperation: ${current.operation}`;
+    printTarget(output, current) {
```
pepone, line 85: I would make printTarget a standalone function, so it remains private to this file.

js/src/Ice/TraceUtil.js
```diff
@@ -58,90 +58,30 @@ function printReply(s, inputStream) {
     const requestId = inputStream.readInt();
     s.push(`\nrequest id = ${requestId}`);
 
+    // This creates a new enumerator (e.g. "18": 18) if one doesn't exist yet.
     const replyStatus = ReplyStatus.valueOf(inputStream.readByte());
-    s.push(`\nreply status = ${replyStatus.value} `);
+    s.push(`\nreply status = ${replyStatus.name}`);
```
pepone, line 63: We provide toString that returns name

```suggestion
    s.push(`\nreply status = ${replyStatus}`);
```

pepone (approved):

pepone:
for prettier run `npx prettier --write .`

js/src/Ice/TraceUtil.js
bernardnormier: Fixed

js/src/Ice/LoggerMiddleware.js
bernardnormier: Fixed and reworked/tested the code, which was bogus.

js/src/Ice/LoggerMiddleware.js
```diff
@@ -15,62 +16,115 @@ export class LoggerMiddleware {
...
+    if (connInfo instanceof IPConnectionInfo) {
+        output.push(connInfo.localAddress);
+        output.push(":");
+        output.push(connInfo.localPort);
+        output.push("<->");
+        output.push(connInfo.remoteAddress);
+        output.push(":");
+        output.push(connInfo.remotePort);
+    } else {
+        // Connection.toString() returns a multiline string, so we just use type here for bt and similar.
```
pepone, line 127: I guess this would never happen in JavaScript we only have IP based transports.

js/src/Ice/OutgoingResponse.js
```diff
@@ -8,7 +8,7 @@ export class OutgoingResponse {
     constructor(outputStream, replyStatus, exceptionId, exceptionDetails) {
         this._outputStream = outputStream;
         this._replyStatus = replyStatus || ReplyStatus.Ok;
-        if (replyStatus !== ReplyStatus.Ok) {
+        if (replyStatus !== ReplyStatus.Ok.value) {
```
pepone, line 11: The `.value` looks suspicious. The `replayStatus` is a enumerator not a byte.

pepone (approved):

js/src/Ice/LoggerMiddleware.js
bernardnormier: current.con.getInfo() can throw (because the connection is closed), resulting in a null connInfo. This could occur if the dispatch closes its own bidir connection.

js/src/Ice/OutgoingResponse.js
bernardnormier: You're right, will fix.

PR #3592, opened 2025-02-21
Complete Dispatch exception implementation

### LINKED ISSUE #3548: Proposal: Add implicit DispatchException

bernardnormier:
This PR makes the dispatch exception implementation and test more consistent across languages and reenables cross-testing for Ice/exceptions.

Fixes #3548.

### DIFF
--- a/cpp/src/Ice/LocalExceptions.cpp
+++ b/cpp/src/Ice/LocalExceptions.cpp
@@ -21,7 +21,7 @@ namespace
     {
         if (message)
         {
-            return *message;
+            return std::move(*message);
         }
         else
         {
--- a/cpp/src/Ice/LocalExceptions.cpp
+++ b/cpp/src/Ice/LocalExceptions.cpp
@@ -17,33 +17,42 @@ using namespace IceInternal;
 
 namespace
 {
-    string createDispatchExceptionMessage(ReplyStatus replyStatus)
+    string createDispatchExceptionMessage(ReplyStatus replyStatus, optional<string> message)
     {
-        ostringstream os;
-        os << "dispatch failed with reply status " << replyStatus;
-        return os.str();
+        if (message)
+        {
+            return *message;
+        }
+        else
+        {
+            ostringstream os;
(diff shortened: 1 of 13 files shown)


cpp/src/Ice/LocalExceptions.cpp
```diff
@@ -17,33 +17,42 @@ using namespace IceInternal;
...
 {
-    string createDispatchExceptionMessage(ReplyStatus replyStatus)
+    string createDispatchExceptionMessage(ReplyStatus replyStatus, optional<string> message)
     {
-        ostringstream os;
-        os << "dispatch failed with reply status " << replyStatus;
-        return os.str();
+        if (message)
+        {
+            return *message;
```
pepone, line 24: Do we need to call std::move here to avoid a copy?

pepone (approved):

externl (approved):

cpp/src/Ice/LocalExceptions.cpp
bernardnormier: You're right, fixed. It would not be necessary without the optional unwrapping.

PR #3594, opened 2025-02-23
Add Communicator.shutdownCompleted property in C#

bernardnormier:
This PR adds the shutdownCompleted property to Communicator in C#.

### DIFF
--- a/csharp/src/Ice/Internal/Instance.cs
+++ b/csharp/src/Ice/Internal/Instance.cs
@@ -48,7 +48,8 @@ public sealed class Instance
         get
         {
             // It would be much nicer to wait asynchronously but doing so requires significant refactoring.
-            var tcs = new TaskCompletionSource();
+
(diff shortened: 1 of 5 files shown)


csharp/src/Ice/Internal/Instance.cs
```diff
@@ -43,9 +43,27 @@ public void updateThreadObservers()
         private Instance _instance;
     }
 
-    public bool destroyed()
+    internal Task shutdownCompleted
```
bernardnormier: This PR is purely about providing a nicer API since the implementation waits synchronously.

csharp/src/Ice/Internal/Instance.cs
```diff
@@ -43,9 +43,27 @@ public void updateThreadObservers()
         private Instance _instance;
     }
 
-    public bool destroyed()
```
bernardnormier, line 46: This destroyed method was unused.

pepone (approved):

externl (approved):

PR #3596, opened 2025-02-24
Add CtrlCHandler.setCallback and Communicator.shutdownCompleted in Swift

bernardnormier:
This PR adds these 2 new functions to Swift.

### DIFF
--- a/swift/src/Ice/CommunicatorI.swift
+++ b/swift/src/Ice/CommunicatorI.swift
@@ -41,8 +41,7 @@ class CommunicatorI: LocalObject<ICECommunicator>, Communicator {
 
     func shutdownCompleted() async {
         await Task {
-            // It would be much nicer to wait asynchronously but doing so requires significant refactoring of the C++
-            // code.
+            // It would be nicer to wait asynchronously but doing so requires significant refactoring of the C++ code.
             waitForShutdown()
         }.value
     }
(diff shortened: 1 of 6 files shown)


swift/src/Ice/CommunicatorI.swift
```diff
@@ -39,6 +39,15 @@ class CommunicatorI: LocalObject<ICECommunicator>, Communicator {
         handle.waitForShutdown()
     }
 
+    func shutdownCompleted() async {
+        async let task = Task {
```
bernardnormier, line 43: I verified this code works as expected. Can we write it in a simpler fashion?

swift/src/Ice/CtrlCHandler.swift
```diff
@@ -11,13 +11,20 @@ public final class CtrlCHandler {
...
+    public func setCallback(_ callback: @escaping (Int32) -> Void) {
+        self.handle.setCallback(callback)
+    }
+
     /// Waits until this handler catches a Ctrl+C or similar signal.
     /// - Returns: The signal number.
     public func catchSignal() async -> Int32 {
         return await withCheckedContinuation { continuation in
-            self.handle.catchSignal { signal in
+            let ok = self.handle.catchSignal { signal in
```
bernardnormier, line 24: (For later) We could refactor the IceStorm demo and eliminate this function.

swift/src/Ice/CommunicatorI.swift
externl: I believe you can just do `await Task { ... }.value`. No need for the async let.

externl (approved):

PR #3597, opened 2025-02-25
Remove `...Info` Structs & Minor `OutputUtil` Cleanup

InsertCreativityHere:
In some of the Slice compilers, we have a form of 'domain separation' between the parser and the code-gen.
Instead of working directly with `DataMember` or `Parameter` (parser types), we convert them into `MemberInfo` and `ParamInfo` structs.

I think the use case for these intermediate types is weak now that the parser API has been vastly improved, and logic has been centralized on the parser types. So, I removed all of these `...Info` structs and the APIs that produced/consumed them.

### DIFF
--- a/cpp/src/Ice/OutputUtil.h
+++ b/cpp/src/Ice/OutputUtil.h
@@ -89,8 +89,8 @@ namespace IceInternal
         void sb(); // Start a block.
         void eb(); // End a block.
 
-        void spar(char = '('); // Start a paramater list.
-        void epar(char = ')'); // End a paramater list.
+        void spar(std::string_view s = "("); // Start a paramater list.
+        void epar(std::string_view s = ")"); // End a paramater list.
 
     private:
         std::string _blockStart;
@@ -163,24 +163,6 @@ namespace IceInternal
     template<> inline Output& operator<<(Output& o, const EndPar&)
     {
         o.epar();
-        return o;
-    }
-
-    class ICE_API StartAbrk{};
-    extern ICE_API StartAbrk sabrk;
-
-    template<> inline Output& operator<<(Output& o, const StartAbrk&)
-    {
-        o.spar('<');
-        return o;
-    }
-
-    class ICE_API EndAbrk{};
-    extern ICE_API EndAbrk eabrk;
-
-    template<> inline Output& operator<<(Output& o, const EndAbrk&)
-    {
-        o.epar('>');
         return o;
     }
 
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -3318,45 +3318,51 @@ Slice::Operation::outParameters() const
 }
 
 ParameterList
-Slice::Operation::sortedReturnAndOutParameters(const string& returnsName)
+Slice::Operation::returnAndOutParameters(const string& returnsName)
 {
-    bool shouldEscapeReturnsName = false;
+    ParameterList outParams = outParameters();
 
-    // First sort each parameter into either 'required' or 'optional'.
-    ParameterList required, optional;
-    for (const auto& param : outParameters())
-    {
-        if (param->optional())
-        {
-            optional.push_back(param);
-        }
-        else
-        {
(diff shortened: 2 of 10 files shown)


cpp/src/Slice/Parser.cpp
```diff
@@ -3318,45 +3318,53 @@ Slice::Operation::outParameters() const
 }
 
 ParameterList
-Slice::Operation::sortedReturnAndOutParameters(const string& returnsName)
+Slice::Operation::returnAndOutParameters(const string& returnsName)
```
InsertCreativityHere, line 3321: Split this into 2 separate functions.
`returnAndOutParameters` returns a list containing... the return and out parameters.
`sortedReturnAndOutParameters` calls this and sorts the list into a specific order.

The old `sortedReturnAndOutParameters` did both of these, but there's some languages where we want to preserve the original order.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -228,36 +173,18 @@ namespace
...
-        ostringstream os;
-        os << "(";
-        for (auto q = params.begin(); q != params.end();)
-        {
-            os << paramToString(*q, scope);
-            q++;
-            if (q != params.end())
-            {
-                os << ", ";
-            }
```
InsertCreativityHere, line 258: This function was simplified and inlined.

InsertCreativityHere:
My compiler-comparison script confirms that the generated code hasn't been changed by this PR.

cpp/src/Slice/Parser.cpp
```diff
@@ -3318,45 +3318,53 @@ Slice::Operation::outParameters() const
...
+    // Check if this operation has a non-void return type.
+    if (_returnType)
     {
-        if (param->optional())
-        {
-            optional.push_back(param);
-        }
-        else
+        // Then check if any of the out parameter's mapped names collide with the return name.
+        bool shouldEscapeReturnsName = false;
```
bernardnormier, line 3329: I really want to remove all this escaping logic from the Slice compilers.

All the escaping should be performed by the user with `xxx:identifier`. Missed escaping => generated code doesn't compile.

cpp/src/Ice/OutputUtil.h
```diff
@@ -89,8 +89,8 @@ namespace IceInternal
         void sb(); // Start a block.
         void eb(); // End a block.
 
-        void spar(char = '('); // Start a paramater list.
-        void epar(char = ')'); // End a paramater list.
+        void spar(std::string_view s = "("); // Start a paramater list.
```
bernardnormier, line 92: parameter

bernardnormier (approved):

cpp/src/Slice/Parser.cpp
InsertCreativityHere: This is a little different, since it's used by _us_ to check whether _our_ parameter should be escaped, not a user-provided identifier.

I agree that user's shouldn't be using language keywords, and if they want to, they can use `xxx:identifier` for that.

But expecting them to remap a parameter because in the internal generated code, we happen to have a temporary variable named `result`... I find this less compelling. These names are implementation details that users shouldn't have to know or care about. To accomplish that, we need to escape these variables if they collide with user's identifiers.

cpp/src/Ice/OutputUtil.h
InsertCreativityHere: Oh, good catch! Usually I'm better at seeing these typos!
Fixed.

cpp/src/Slice/Parser.cpp
bernardnormier: If it's an truly internal name, we should use an invalid name like `$return`. Then no possible collision.

InsertCreativityHere: Yes, we probably should be doing this.
I'll look into removing these vestiges of "escaping" after finishing the `xxx:identifier`s.

externl (approved):

PR #3598, opened 2025-02-25
Msbuild builder

pepone:
This PR moves `zeroc.icebuilder.msbuild` builder into tools/Ice.Slice.Tools package.

Updates the C# build to use tasks from `tools/Ice.Slice.Tools` and avoid the dependency on `zeroc.icebuilder.msbuild` package.

Upgrades `zeroc.ice.net` NuGet package to include the builder. Later we will upgrade the demos to not require a separate builder package and just use `zeroc.ice.net`.

Next I will update Windows C++. NuGet package to do the same.

Opening as a Draft to get some feedback about the layout and general approach for doing this.


bernardnormier:
What is the advantage of having a single tool that can handle both C++ (with slice2cpp) and C# (with slice2cs).

As I understand it, we'll ship this tool in the C++ NuGet and the C# NuGet - in the C++ NuGet together with slice2cpp (for multiple platforms) and in the C# NuGet together with slice2cs (for multiple platforms).

Is this because there is a lot of code in common? Because it's nicer to have a single tool (Ice.Slice.Tools), just like we have a single IceRpc.Slice.Tools?

bernardnormier (approved):
We should add a README.md in tools and some doc-comments on the tools classes.

pepone:
> We should add a README.md in tools and some doc-comments on the tools classes.

Yes I will do in a follow-up PR.

externl:
Does this affect the NuGet packaging?

pepone:
> Does this affect the NuGet packaging for C++?

Yes, the C++ NuGet package includes an extra assemblies in `tools\Ice.Slice.Tools.Cpp.dll` which contains the MSBuild tasks to run the slice2cpp compiler. 

It also include the require props and targets files.

externl:
> > Does this affect the NuGet packaging for C++?
> 
> Yes, the C++ NuGet package includes an extra assemblies in `tools\Ice.Slice.Tools.Cpp.dll` which contains the MSBuild tasks to run the slice2cpp compiler.
> 
> It also include the require props and targets files.

Ok, but the `slice2cpp` compiler remains in the same place.

externl (approved):

pepone:
> but the slice2cpp compiler remains in the same place.

Yes. It adds new files, existing files remain in the usual place.

PR #3599, opened 2025-02-25
Add support for Swift nightly packaging

externl:
This PR adds support for Swift nightly packaging. We can't use the XCFramework from brew since `binaryTargets` must be relative to `Package.swift` and `systemLibrary` doesn't support XCFrameworks. 

Unfortunately this means that we can't use the main Ice repo as the packaging and we'll have to use https://github.com/zeroc-ice/ice-spm (I think should rename to `IceSwift`). 

For the nightly builds I've setup a structure similar to our nightly brew tap.

Additionally, with some help from @bernardnormier I was able to remove the dependance on Ice's private header files.


Package.swift
externl, line 1: I moved the plugin to use the expected Swift folder structure.

packaging/swift/build-swift-repo.sh
externl, line 1: This file creates the "IceSwift" repository with the updated Package.swift and all the files in the right spot.

swift/src/Ice/CommunicatorI.swift
```diff
@@ -296,11 +297,27 @@ extension Communicator {
 }
 
 struct DefaultsAndOverrides {
-    init(handle: ICECommunicator) {
-        var defaultEncoding = EncodingVersion()
-        handle.getDefaultEncoding(major: &defaultEncoding.major, minor: &defaultEncoding.minor)
-        self.defaultEncoding = defaultEncoding
-        defaultFormat = FormatType(rawValue: handle.getDefaultFormat())!
+    init(_ properties: Properties) throws {
+        let defaultEncodingValue = properties.getIceProperty("Ice.Default.EncodingVersion")
```
externl, line 301: We just read the two properties we care about manually.

swift/src/IceImpl/TraceUtil.mm
```diff
@@ -2,9 +2,13 @@
 
 #import "include/TraceUtil.h"
 
-#import "Ice/TraceUtil.h"
 #import "LoggerWrapperI.h"
 
+namespace IceInternal
+{
+    void traceSlicing(const char*, std::string_view, const char*, const Ice::LoggerPtr&);
```
bernardnormier, line 9:
```
// Redeclare internal function from the Ice C++ library.
```

swift/src/Ice/CommunicatorI.swift
```diff
@@ -296,11 +297,27 @@ extension Communicator {
...
+        default:
+            fatalError("Invalid Ice.Default.EncodingVersion value: \(defaultEncodingValue)")
+        }
+
+        let defaultFormatValue = try properties.getIcePropertyAsInt("Ice.Default.SlicedFormat")
+
+        switch defaultFormatValue {
+        case 0:
+            defaultFormat = .compactFormat
+        case 1:
```
bernardnormier, line 316: According to https://doc.zeroc.com/ice/3.7/property-reference/ice-default#id-.Ice.Default.*v3.7-Ice.Default.SlicedFormat

0 = compact
non-0 = sliced

That's not what you implemented.

bernardnormier (approved):

pepone (commented):
I wonder if we should take a different approach here.

For the 3.8 release, we could build the XCFrameworks before tagging, compute their checksum, and update `Package.swift` to reference a URL pointing to the tagged XCFramework and its computed checksum.

This would require a commit in the `3.8` branch, meaning that when building Ice for Swift from the `3.8` branch, it would use the prebuilt binary XCFrameworks.

Meanwhile, `main` would continue using the local path, always referencing XCFrameworks built from the local source.

For nightly builds, I think we should mimic the 3.8 workflow:

- Create a clone of Ice
- Push it to Nexus
- Update the binary targets in this clone to use the checksum and URL of the XCFrameworks previously uploaded to Nexus

### With this structure:

- We can commit Swift-only fixes in `3.8` while continuing to use the same XCFrameworks from the tag.
- Fixes to C++ XCFrameworks would require a new tag and an update to `binaryTargets` in `Package.swift`.

### Advantages:

- No need for a special source distribution for Swift Package Manager.
- Avoids the need for separate `Package.swift` files.

pepone (approved):

PR #3601, opened 2025-02-26
Rename dispatcher => executor in Python

bernardnormier:
This is a plain renaming, no logic change.


python/modules/IcePy/Communicator.cpp
```diff
@@ -66,7 +66,7 @@ communicatorNew(PyTypeObject* type, PyObject* /*args*/, PyObject* /*kwds*/)
     self->shutdownFuture = nullptr;
     self->shutdownException = nullptr;
     self->shutdown = false;
-    self->dispatcher = 0;
+    self->executor = 0;
```
pepone, line 69:
```suggestion
    self->executor = nullptr;
```

pepone (approved):

externl (approved):

PR #3604, opened 2025-02-26
Add Python coroutine executor

pepone:
This PR adds a new coroutine executor that can be set on the InitializationData object used to create a communicator.

The coroutine executor is used by the dispatch code to execute coroutines returned by servant dispatch methods.


python/test/Ice/asyncio/AllTests.py
```diff
@@ -25,22 +25,22 @@ async def allTestsAsync(helper, communicator):
     sys.stdout.write("testing exceptions... ")
     sys.stdout.flush()
     try:
```
pepone, line 27: Unrelated fix, we were testing the sync versions. Ice.wrap_future was not used ...

bernardnormier (approved):

python/python/Ice/Communicator.py
```diff
@@ -523,3 +524,15 @@ def findAllAdminFacets(self):
             A dictionary where the keys are facet names (str) and the values are the associated servants (Ice.Object).
         """
         return self._impl.findAllAdminFacets()
+
+    def getCoroutineExecutor(self):
```
externl, line 528: Would be nice to start using type hinting.

externl (approved):

PR #3606, opened 2025-02-26
Add Support for `js:identifier` Metadata

InsertCreativityHere:
This PR adds support for `js:identifier` just like we did in all the other languages so far.
It fixes the `escape` test to use the new metadata, and also cleans up the `slice2js` compiler a little while I was in there.


externl (approved):

InsertCreativityHere:
> Looks good! The mappedScoped method is clearly an improvement over what we have, but it falls a bit sort when you have to resort to string methods for post-processing the returned scope and deal with the language mapping details.

Yes, you're right. I was planning on waiting until after I finished the `xxx:identifier`s to fix this, but maybe I'll do it now instead.
It's really just a matter of removing leading/trailing scope pieces.

PR #3608, opened 2025-02-27
Add PythonDispatcher to handle calls to servant dispatch methods

pepone:
This PR moves the logic from Python Object._iceDispatch implementation to PythonDispatcher.dispatch. It add some comments to clarify the implementation and cleanups things.

### DIFF
--- a/python/modules/IcePy/Operation.cpp
+++ b/python/modules/IcePy/Operation.cpp
@@ -2249,16 +2249,10 @@ Upcall::dispatchImpl(PyObject* servant, const string& dispatchName, PyObject* ar
         throw Ice::UnknownException{__FILE__, __LINE__, ostr.str()};
     }
 
-    // Get the dispatch function from Ice.PythonDispatcher module.
+    // Get the dispatch function from Ice.Dispatch module.
     // lookupType() returns a borrowed reference.
-    PyObject* dispatchMethod = lookupType("Ice.PythonDispatcher.dispatch");
-    if (!dispatchMethod)
-    {
-        ostringstream ostr;
-        ostr << "dispatch method not found for identity " << communicator->identityToString(current.id)
-             << " and operation '" << dispatchName << "'";
-        throw Ice::UnknownException{__FILE__, __LINE__, ostr.str()};
-    }
+    PyObject* dispatchMethod = lookupType("Ice.Dispatch.dispatch");
+    assert(dispatchMethod);
     Py_INCREF(dispatchMethod);
     // Ensure we release the reference when this method exist.
     PyObjectHandle dispatchMethodHandle{dispatchMethod};
--- a/python/modules/IcePy/Operation.cpp
+++ b/python/modules/IcePy/Operation.cpp
@@ -1077,9 +1077,9 @@ namespace IcePy
     static PyTypeObject DispatchCallbackType = {
         /* The ob_type field must be initialized in the module init function
          * to be portable to Windows without using C++. */
-        PyVarObject_HEAD_INIT(0, 0) "IcePy.Dispatch", /* tp_name */
-        sizeof(DispatchCallbackObject),               /* tp_basicsize */
-        0,                                            /* tp_itemsize */
+        PyVarObject_HEAD_INIT(0, 0) "IcePy.DispatchCallback", /* tp_name */
+        sizeof(DispatchCallbackObject),                       /* tp_basicsize */
+        0,                                                    /* tp_itemsize */
         /* methods */
         reinterpret_cast<destructor>(dispatchCallbackDealloc), /* tp_dealloc */
         0,                                                     /* tp_print */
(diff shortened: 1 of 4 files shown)


python/modules/IcePy/Operation.cpp
```diff
@@ -2249,17 +2249,19 @@ Upcall::dispatchImpl(PyObject* servant, const string& dispatchName, PyObject* ar
...
-    PyObjectHandle dispatchMethod{getAttr(servant, "_iceDispatch", false)};
-    if (!dispatchMethod.get())
+    // Get the dispatch function from Ice.PythonDispatcher module.
+    // lookupType() returns a borrowed reference.
+    PyObject* dispatchMethod = lookupType("Ice.PythonDispatcher.dispatch");
+    if (!dispatchMethod)
     {
         ostringstream ostr;
-        ostr << "_iceDispatch method not found for identity " << communicator->identityToString(current.id)
+        ostr << "dispatch method not found for identity " << communicator->identityToString(current.id)
```
pepone, line 2258: Need a better message here.

python/python/Ice/Object.py
```diff
@@ -84,76 +81,6 @@ def ice_staticId():
...
-    def _iceDispatchCoroutine(self, cb, coro, value=None, exception=None):
-        try:
-            if exception:
-                result = coro.throw(exception)
-            else:
-                result = coro.send(value)
-
-            if result is None:
-                # The result can be None if the coroutine performs a bare yield (such as asyncio.sleep(0))
-                cb.response(None)
```
pepone, line 130: This was bogus `await asyncio.sleep(0)` would indeed return `None` but doesn't mean dispatch completed, and we should not be calling `cb.respose` in this case.

In real life this code should not be used when handling asyncio coroutines, you should have a coroutineExecutor in place.

I tested we were never exercising this branch with the test suite. Is gone from the new implementation.

python/python/Ice/PythonDispatcher.py
```diff
@@ -0,0 +1,95 @@
+# Copyright (c) ZeroC, Inc.
+
+import inspect
+import sys
+
+def is_future(obj):
```
pepone, line 6: We previously did:

```
if isinstance(result, Future) or callable(
                getattr(result, "add_done_callback", None)
            )
```

But the code only uses `add_done_callback` which is also provided by Ice.Future. Doesn't make much sense to have separate checks.

python/python/Ice/PythonDispatcher.py
```diff
@@ -0,0 +1,95 @@
...
+        The request parameters.
+    """
+
+    # Invoke the servant method with the request parameters. Let exceptions propagate to the caller.
+    result = method(*args)
+
+    # If the result is a coroutine and the communicator has a custom coroutine executor, execute the coroutine using
+    # the configured executor.
+    if inspect.iscoroutine(result):
+        assert (
```
pepone, line 43: And assert is fine here because this is only called by IcePy.

python/python/Ice/PythonDispatcher.py
```diff
@@ -0,0 +1,92 @@
...
+            result = coroutine.send(value)
+
+        if is_future(result):
+            def handle_future_result(future):
+                try:
+                    run_coroutine(cb, coroutine, value=future.result())
+                except BaseException:
+                    run_coroutine(cb, coroutine, exception=sys.exc_info()[1])
+
+            result.add_done_callback(handle_future_result)
```
pepone, line 86: Note the potential recursive call, I assuming would rarely completed synchronously.

Keep in mind `run_coroutine` is intended for dealing with AMI calls nested in AMD dispatch. If using asyncio or another event loop you have to provide a coroutine executor.

python/modules/IcePy/Operation.cpp
```diff
@@ -2249,17 +2249,19 @@ Upcall::dispatchImpl(PyObject* servant, const string& dispatchName, PyObject* ar
...
              << " and operation '" << dispatchName << "'";
         throw Ice::UnknownException{__FILE__, __LINE__, ostr.str()};
     }
+    Py_INCREF(dispatchMethod);
```
externl, line 2262: Isn't this a non-member function, why does it need the `Py_INCREF`?

python/python/Ice/PythonDispatcher.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Invoke the servant method with the request parameters. Let exceptions propagate to the caller.
+    result = method(*args)
+
+    # If the result is a coroutine and the communicator has a custom coroutine executor, execute the coroutine using
+    # the configured executor.
+    if inspect.iscoroutine(result):
+        assert (
+            len(args) > 0
+        ), "args must have at least one element representing the dispatch current parameter"
```
bernardnormier, line 45: "args must have at least one argument, current"

python/python/Ice/PythonDispatcher.py
bernardnormier: Please add a comment above "def run_coroutine"

python/python/Ice/PythonDispatcher.py
```diff
@@ -0,0 +1,92 @@
+# Copyright (c) ZeroC, Inc.
```
bernardnormier, line 1: I would rather not "dispatcher" for this Util file. And we should prefix Python code with Python.

What about simply Dispatch.py?

python/modules/IcePy/Operation.cpp
```diff
@@ -2249,17 +2249,19 @@ Upcall::dispatchImpl(PyObject* servant, const string& dispatchName, PyObject* ar
...
     }
 
-    //
-    // Get the _iceDispatch method. The _iceDispatch method will invoke the servant method and pass it the arguments.
-    //
-    PyObjectHandle dispatchMethod{getAttr(servant, "_iceDispatch", false)};
-    if (!dispatchMethod.get())
+    // Get the dispatch function from Ice.PythonDispatcher module.
+    // lookupType() returns a borrowed reference.
+    PyObject* dispatchMethod = lookupType("Ice.PythonDispatcher.dispatch");
```
bernardnormier, line 2254: I find the term dispatchMethod confusing. It can be this `dispatch` method or the `method` arg of `dispatch`.

You can't call it simply `dispatch`?

bernardnormier (approved):

PR #3610, opened 2025-02-27
Remove random shuffling from Network.getAddresses

bernardnormier:
This PR removes the random shuffling from Network.getAddresses in C++, C# and Java.

`getaddrinfo` returns addresses in the order specified by RFC 3484 and we should never reshuffle them randomly.


externl (approved):
Looks good!

pepone (commented):
Maybe I missing something here, but seems like we can now remove the EndpointSelectionType from a few APIs:

- ProtocolInstance::resolve (parameter)
- IPEndpointI::connectorsAsync (parameter)
- EndpointHostResolver::resolve (parameter)
- EndpointHostResolver::ResolveEntry (data member)
- OutgoingConnectionFactory::createAsync (parameter)
- OutgoingConnectionFactory::ConnectCallback (data member)

bernardnormier:
> Maybe I missing something here, but seems like we can now remove the EndpointSelectionType from a few APIs:

No, you're right. I've now removed them all.

In case you're wondering, we're still using the selected endpoint selection type in the `filterEndpoints` method of Reference.

pepone (approved):

PR #3611, opened 2025-02-27
Avoid using parameters in Nightly Release Workflow

pepone:
Move schedule to the release workflow


.github/workflows/build-nightly-release-workflow.yml
```diff
@@ -195,9 +172,9 @@ jobs:
     needs: set-version
     with:
       ice_version: ${{ needs.set-version.outputs.rpm_version }}
-      repository_url: ${{ inputs.repository_url }}
+      repository_url: "https://download.zeroc.com/nexus/repository"
       repository_suffix: "-nightly"
-      repository_username: ${{ inputs.repository_username }}
+      repository_username: "nightly"
```
externl, line 177: I wonder if we should change this user name to "github"

externl (approved):
Looks good. 

I would consider creating an action variable for the nightly repo and nightly username and we can just use something like ${{ vars.nightly_repo }} and ${{ vars.nightly_user }}

pepone:
I would rework this, need a few more improvments.

PR #3612, opened 2025-02-27
Fail earlier on ConnectTimeout for multi-home endpoint

bernardnormier:
This PR updates the OutgoingConnectionFactory logic to stop further "connector" connection attempt when an attempt fails with ConnectTimeoutException.

This speeds up connect failure for multi-home endpoints.

### DIFF
--- a/js/src/Ice/OutgoingConnectionFactory.js
+++ b/js/src/Ice/OutgoingConnectionFactory.js
@@ -623,8 +623,10 @@ class ConnectCallback {
 
         // We stop on ConnectTimeoutException to fail reasonably fast when the endpoint has many connectors
         // (IP addresses).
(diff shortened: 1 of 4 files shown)


js/src/Ice/OutgoingConnectionFactory.js
```diff
@@ -618,20 +619,16 @@ class ConnectCallback {
     }
 
     connectionStartFailedImpl(ex) {
-        if (ex instanceof LocalException) {
```
bernardnormier, line 621: I removed this check because I don't understand it. What other kind of exception do we get here?

Note that handleConnectionException really means "log connection exception".

pepone (approved):
Looks good!

externl (approved):

PR #3613, opened 2025-02-28
Add checkedCastAsync to C#

bernardnormier:
This PR adds a checkedCastAsync to C#.

I didn't add it in other languages; it could make sense to add it in other languages with sync & async proxy methods, such as Python.

### DIFF
--- a/csharp/src/Ice/Proxy.cs
+++ b/csharp/src/Ice/Proxy.cs
@@ -1915,6 +1915,30 @@ public class ObjectPrxHelper : ObjectPrxHelperBase
         checkedCast(proxy?.ice_facet(facet), context);
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


csharp/src/Ice/Proxy.cs
```diff
@@ -1914,6 +1914,30 @@ public static ObjectPrx createProxy(Communicator communicator, string proxyStrin
     public static ObjectPrx? checkedCast(ObjectPrx? proxy, string facet, Dictionary<string, string>? context = null) =>
         checkedCast(proxy?.ice_facet(facet), context);
 
+    /// <summary>
```
bernardnormier, line 1917: I will rework all these doc-comments in other PR.

externl (approved):

PR #3614, opened 2025-03-01
Stop printing -t 60000 for tcp, ssl (...) endpoints

bernardnormier:
This PR updates the printing of tcp/ssl/bt (...) endpoints to no longer include "-t 60000".

As a reminder: this timeout no longer has any effect in Ice 3.8 and is only kept for backwards compatibility with Ice 3.7. If you use Ice 3.8, you should never see it - and including "-t 60000" in all proxy strings is confusing.

### DIFF
--- a/cpp/src/Ice/TcpEndpointI.cpp
+++ b/cpp/src/Ice/TcpEndpointI.cpp
@@ -31,6 +31,11 @@ extern "C"
     }
 }
 
+namespace
+{
+    const int32_t defaultTimeout = 60000; // 60,000 ms.
+}
+
 IceInternal::TcpEndpointI::TcpEndpointI(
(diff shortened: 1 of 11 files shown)


pepone (approved):
I was wondering if we can make timeout optional, so that if we read `60000` from the peer we still print it. For 3.8 only environments would always remain nullopt/null and we don't print anything.

bernardnormier:
> I was wondering if we can make timeout optional, so that if we read 60000 from the peer we still print it. For 3.8 only environments would always remain nullopt/null and we don't print anything.

We can't encode "not set" for the timeout. The peer could be a 3.8 app that sends us a 60000 timeout. So I don't think we should treat a decoded 60000 differently than a locally generated 60000.

PR #3615, opened 2025-03-01
Split C# NuGet packages and rework MSBuild builders

pepone:
This PR updates C# NuGet packaging to create a package per project/assembly. The MSBuild builders are also separate packages, this aligns with what we did for IceRPC.

We produce the following packages:

- ZeroC.Glacier2 (This contains the generated code)
- ZeroC.Ice (The Ice runtime)
- ZeroC.IceBox (IceBox service)
- iceboxnet (iceboxnet executable packaged as a dotnet tool)
- ZeroC.IceDiscovery (The IceDiscovery plug-in)
- ZeroC.IceGrid  (The IceGrid generated code)
- ZeroC.IceLocatorDiscovery  (The IceLoctorDiscovery plug-in)
- ZeroC.IceStorm  (The IceStorm generated code)
- ZeroC.Ice.Slice.Tools.CSharp  (The MSBuild builder for Ice for C# projects)
- ZeroC.Ice.Slice.Tools.Cpp(The MSBuild builder for Ice for C++ projects)


cpp/msbuild/ice.nuget.targets
```diff
@@ -4,16 +4,6 @@
         <IceSrcRootDir>$(MSBuildThisFileDirectory)..\</IceSrcRootDir>
     </PropertyGroup>
 
-    <!-- Slice Tools files -->
```
pepone, line 7: Moved to `ZeroC.Ice.Slice.Tools.Cpp`

csharp/BUILDING.md
pepone, line 1: I need to improve this file in a follow up PR.

csharp/test/IceDiscovery/simple/msbuild/client/client.csproj
```diff
@@ -23,5 +23,6 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="../../../../../src/Ice/Ice.csproj" />
+    <ProjectReference Include="../../../../../src/IceDiscovery/IceDiscovery.csproj" />
```
pepone, line 26: This ensures the plug-in is available in the test output directory.

This is required because I updated the build to use the default output  path. We previously override to use the top level `lib` and `bin` directories.

tools/ZeroC.Ice.Slice.Tools.CSharp/README.md
pepone, line 1: Need to expand this README with sections about how to configure the builder.

tools/ZeroC.Ice.Slice.Tools/SliceCompilerTask.cs
```diff
@@ -53,12 +53,12 @@ protected virtual Dictionary<string, string> GetOptions()
             ["OutputDir"] = OutputDir.TrimEnd('\\')
         };
 
-        if (IncludeDirectories != null && IncludeDirectories.Length > 0)
+        if (IncludeDirectories.Length > 0)
```
pepone, line 56: This is an unrelated fix. I updated the properties to be non nullable when imported the builders into the ice repository, but the code was still doing null checks, once of this was causing the rebuilds of Slice files.

tools/ZeroC.Ice.Slice.Tools/TaskUtil.cs
```diff
@@ -5,15 +5,11 @@
 using System.IO;
 using System.Runtime.InteropServices;
 
-namespace Ice.Slice.Tools;
+namespace ZeroC.Ice.Slice.Tools;
 
 public class TaskUtil
 {
-#if NETSTANDARD2_0
```
pepone, line 12: Modern .NET Frameworks support netstandard2.0, don't need this anymore. The MSBuild used by VC++ is still based on .NET Framework.

tools/ZeroC.Ice.Slice.Tools.CSharp/README.md
```diff
@@ -0,0 +1,20 @@
+# Slice Tools for Ice for C#
```
bernardnormier, line 1: I don't like the double for. What about simply: Slice Tools for Ice C#.

Then, I think we should use the same name consistently to avoid confusion. I would use "Slice tools for Ice C#" and "Slice tools for Ice C++" and drop the builder terminology for these tools/packages.

tools/ZeroC.Ice.Slice.Tools.CSharp/README.md
```diff
@@ -0,0 +1,20 @@
...
+Linux (x64 and arm64), macOS (x64 and arm64) and Windows (x64).
+
+Once you've added the ZeroC.Ice.Slice.Tools.CSharp NuGet package to your project, the Slice files of your project are
+automatically compiled into C# files every time you build this project.
+
+> The Slice MSBuild task checks the dependencies of the Slice files it compiles and only rebuilds the Slice files that,
+> are not up to date.
+
+[Source code][source] | [Package][package] | [slice2cs options][slice2cs] | [Slice documentation][slice]
+
```
bernardnormier, line 16: Do you plan to add more information in a follow-up PR?

See https://github.com/icerpc/icerpc-csharp/tree/main/tools/IceRpc.Slice.Tools

tools/ZeroC.Ice.Slice.Tools.Cpp/README.md
```diff
@@ -0,0 +1,20 @@
+# Slice Tools for Ice for C++
```
bernardnormier, line 1: See C# comments.

bernardnormier (approved):

PR #3618, opened 2025-03-03
Add Communicator::waitForShutdownAsync in C++

bernardnormier:
This PR adds Communicator::waitForShutdownAsync in C++ and uses this function to implement Communicator.shutdownCompleted() in Swift.

### DIFF
--- a/cpp/src/Ice/ObjectAdapterFactory.cpp
+++ b/cpp/src/Ice/ObjectAdapterFactory.cpp
@@ -7,8 +7,6 @@
 #include "Ice/UUID.h"
 #include "ObjectAdapterI.h"
 
-#include <algorithm>
-
 using namespace std;
 using namespace Ice;
 using namespace IceInternal;
@@ -39,10 +37,10 @@ IceInternal::ObjectAdapterFactory::shutdown() noexcept
     }
 
     // Deactivate outside the thread synchronization, to avoid deadlocks.
-    for_each(
-        adapters.begin(),
-        adapters.end(),
-        [](const shared_ptr<ObjectAdapterI>& adapter) { adapter->deactivate(); });
+    for (const auto& adapter : adapters)
+    {
+        adapter->deactivate();
+    }
 }
 
 void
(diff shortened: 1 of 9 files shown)


pepone (approved):

cpp/src/Ice/ObjectAdapterFactory.cpp
```diff
@@ -39,59 +37,116 @@ IceInternal::ObjectAdapterFactory::shutdown()
...
+                                                            callback();
+                                                        }
+                                                    }};
+
+                // Joining this thread is challenging because a callback can destroy the communicator and indirectly
+                // this object adapter factory. So we just detach it.
+                shutdownCompletedThread.detach();
+            }
+
+            _shutdownCompletedCallbacks.emplace_back(std::move(completed));
```
externl, line 111: I think std::push_back would be more idiomatic here since we're moving an existing function, not creating a new one.

externl (approved):

cpp/src/Ice/ObjectAdapterFactory.cpp
bernardnormier: Fixed

cpp/src/Ice/ObjectAdapterFactory.cpp
```diff
@@ -39,59 +37,128 @@ IceInternal::ObjectAdapterFactory::shutdown()
...
+                                                           self->waitForShutdown();
+                                                           vector<function<void()>> callbacks;
+                                                           {
+                                                               lock_guard lg(self->_mutex);
+                                                               callbacks = std::move(self->_shutdownCompletedCallbacks);
+                                                           }
+
+                                                           for (const auto& callback : callbacks)
+                                                           {
+                                                               callback();
```
externl, line 103: We should probably try/catch around these user callbacks.

cpp/src/Ice/ObjectAdapterFactory.cpp
```diff
@@ -39,59 +37,128 @@ IceInternal::ObjectAdapterFactory::shutdown()
...
+    if (shutdownCompletedThread.joinable())
+    {
+        if (shutdownCompletedThread.get_id() == std::this_thread::get_id())
+        {
+            // destroy is called from a shutdown completed callback - this is fine, but the thread can't join itself.
+            shutdownCompletedThread.detach();
+        }
+        else
+        {
+            shutdownCompletedThread.join();
```
externl, line 152: Could we just join the tread in the ObjectAdapterFactory destructor?

externl (approved):

cpp/src/Ice/ObjectAdapterFactory.cpp
bernardnormier: The doc-comment says the callback must not throw any exception.

cpp/src/Ice/ObjectAdapterFactory.cpp
bernardnormier: It's nicer to join it in destroy (= earlier), and we would have the same join/detach issue in the destructor: the last refcount can be released by the shutdown completed thread.

pepone (approved):

PR #3619, opened 2025-03-03
Fix to correctly set NuGet PackageVersion

pepone:
This PR fixes the PackageVersion settings. The setting was not correctly pass to the pack targets and they were using default 1.0.0 version for .NET project system. Which resulted in an error with `/t:Publish` not finding the packages with the expected version `3.8.0-alpha0`

### DIFF
--- a/csharp/msbuild/ice.common.props
+++ b/csharp/msbuild/ice.common.props
@@ -22,9 +22,6 @@
     <StyleCopAnalyzersVersion>1.2.0-beta.556</StyleCopAnalyzersVersion>
   </PropertyGroup>
 
-  <!-- Import Ice version settings -->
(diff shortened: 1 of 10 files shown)


csharp/msbuild/ice.common.props
```diff
@@ -22,9 +22,6 @@
     <StyleCopAnalyzersVersion>1.2.0-beta.556</StyleCopAnalyzersVersion>
   </PropertyGroup>
 
-  <!-- Import Ice version settings -->
-  <Import Project="$(MSBuildThisFileDirectory)..\..\config\ice.version.props"/>
```
pepone, line 26: We don't need this for the tests.

bernardnormier (approved):

PR #3621, opened 2025-03-03
Remove binding test in C++-based mappings

bernardnormier:
This PR removes the Ice/binding test in all C++-based language mappings (but not C++).

The code exercised by this test is implemented in C++; there is no point in testing this behavior (client-side connection establishment) in these languages.


externl (approved):
Looks good. At least in Swift this removes all "testing" of prx.ice_endpoints()`. We should test this in the proxy test.

pepone (approved):

PR #3622, opened 2025-03-03
Replace ICE_BUILDING_ICE by ICE_SKIP_PRAGMA_COMMENT

### LINKED ISSUE #3607: Cleanup C++ build flags in python/setup.py
We should review 

https://github.com/zeroc-ice/ice/blob/6993fba35c1a966d611201786838a5cd071c430b/python/setup.py#L63

are all these macros still required?

bernardnormier:
Fixes #3607. The xxx_API = `/**/` are fine.

Question for the reviewers:
 - better name for ICE_SKIP_PRAGMA_COMMENT? It's obviously Windows-only.
 - should use it for all pragma comment in header files, not just Ice?


pepone:
> better name for ICE_SKIP_PRAGMA_COMMENT? It's obviously Windows-only.

I would go with ICE_DISABLE_PRAGMA_COMMENT_LIB or ICE_IGNORE_PRAGMA_COMMENT_LIB

> should use it for all pragma comment in header files, not just Ice?

I think we should. It is easy to understand.

pepone (approved):

externl:
> * better name for ICE_SKIP_PRAGMA_COMMENT? It's obviously Windows-only.

I think the name is fine. 

> * should use it for all pragma comment in header files, not just Ice?

Sounds reasonable.

externl (approved):

Issue #3620, opened 2025-03-03
GNU Make 4.x parallel builds broken


externl:
Parallel builds are broken when building with GNU Make 4.x. 

One cause is our use of [double colon rules](https://www.gnu.org/software/make/manual/html_node/Double_002dColon.html) do define multiple prerequisites for the same target.

eg.

```
all:: foo
all:: bar
```

With 4.x these no longer builds in parallel, and we need to update these to `all:: foo bar`. 

There are likely other issues.

pepone:
Fixed in #3623

PR #3623, opened 2025-03-04
Parallel build fixes

pepone:
See #3620

There are more improvements we can make, this is just a simple one:

The `make -j8 srcs` when from

With this branch:

```
real    4m11.484s
user    24m55.406s
sys     1m25.977s
```

With main:

```
real    5m28.230s
user    22m30.152s
sys     1m16.994s
```

### DIFF
--- a/config/Make.project.rules
+++ b/config/Make.project.rules
@@ -801,7 +801,7 @@ $1_distclean::
 #
 # Global targets
 #
-all:: $1
+all: $1
 clean:: $1_clean
 install:: $1_install
 distclean:: $1_distclean
@@ -1166,7 +1166,7 @@ projects :=
 # Global targets
 #
 .PHONY: all
-all::
+all:
 
 .PHONY: clean
 clean::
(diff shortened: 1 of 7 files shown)


pepone:
This fix doesn't work with the top-level file because we have `all::` there and we need to keep it.

You can't have both `::` and `:` for the same target.

pepone:
A second approach is to not define `all::` inside the `make-project` and similar functions and avoid this pattern

```
all:: $1
```

As show in the second commit in this branch we can instead do:

```
srcs:: $(projects)
```

After loading the projects, and do the same for tests, etc.

I just fixed cpp, and that is why the build is failing, we need to do the same for C++ based mappings as targets are not automatically added to `all::`, `srcs::`, `tets::`

With this change `make -j8` took

```
real    9m21.586s
user    65m37.426s
sys     3m30.765s
```

With main the same command takes:

externl (approved):
Looks good!

InsertCreativityHere (approved):
I'm not a make expert, but it looks reasonable to me!

bernardnormier (approved):

PR #3624, opened 2025-03-04
Cleanup C++ Windows builds

pepone:
This PR does a minor cleanup to the C++ Windows build settings

### DIFF
--- a/cpp/msbuild/ice.cpp.props
+++ b/cpp/msbuild/ice.cpp.props
@@ -16,7 +16,7 @@
   <PropertyGroup>
     <IceSrcRootDir>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\..'))</IceSrcRootDir>
     <IntDir>$(Platform)\$(Configuration)\</IntDir>
-    <LinkIncremental>false</LinkIncremental>
+    <LinkIncremental>true</LinkIncremental>
   </PropertyGroup>
 
   <Choose>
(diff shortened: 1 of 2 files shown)


cpp/msbuild/ice.cpp.props
```diff
@@ -122,14 +112,6 @@
     </Link>
   </ItemDefinitionGroup>
 
-  <ItemDefinitionGroup>
-    <Link>
-      <!-- We disable incremental linking (see above), which turns on /OPT:LBR, which in turn conflicts with -->
-      <!-- /EDITANDCONTINUE even though /OPT:LBR is an ARM-only option. So we turn off LBR. -->
-      <AdditionalOptions>/OPT:NOLBR %(AdditionalOptions)</AdditionalOptions>
```
pepone, line 129: We don't use `EDITANDCONTINUE` so not clear why we have this comment. I also enable incremental linking which seems to be the defautl.

cpp/msbuild/ice.test.props
```diff
@@ -2,23 +2,14 @@
...
 
   <PropertyGroup>
-    <IceSoVersion>38a0</IceSoVersion>
     <IceBuildingSrc>no</IceBuildingSrc>
   </PropertyGroup>
 
   <Import Project="$(MSBuildThisFileDirectory)\ice.cpp.props"/>
 
-  <!--
-      The correct value for GenerateDebugInformation depends on the compiler:
```
pepone, line 12: This is a leftover from when we have different settings from different compilers, there is a single VC compiler now.

externl (approved):

bernardnormier (approved):

PR #3629, opened 2025-03-04
Remove "--meta" option from slice2java

### LINKED ISSUE #45: Deprecate slice2java meta option
slice2java provides a "meta" option unlike other Slice compilers. It should have been deprecated in Ice 3.7 when we deprecated various slice compiler options, but was not.

ICE-8537

bernardnormier:
Fixes #45.

This PR also removes code from Parser, that was used only by this meta option.


InsertCreativityHere (approved):
Looks good.

Now that this is gone, there's more cleanup to be done for how we handle inclusion tracking though.

PR #3630, opened 2025-03-05
Add CMake support files 

externl:
This PR adds CMake support files that we will ship in our packaging.

We install the CMake files to the following locations:

- **NuGet**: `<nuget_package>/cmake/`
- **macOS**: `<install_prefix>/lib/cmake/Ice-<version>`
- **Debian**: `<install_prefix>/lib/<arch>/cmake/Ice-<version>`
- **RedHat**: ` <install_prefix>/<lib>/cmake/Ice-<version>`

### DIFF
--- a/cpp/config/IceConfig.cmake
+++ b/cpp/config/IceConfig.cmake
@@ -2,11 +2,16 @@
 
 if(WIN32)
     # On Windows Ice ships was a NuGet package, so we need to find the package prefix directory.
-    # This config file is located at <ice_nuget_package>/build/native/lib/cmake/ice/IceConfig.cmake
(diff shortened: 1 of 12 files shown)


pepone (approved):

cpp/config/IceConfig.cmake
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) ZeroC, Inc.
+
+if(WIN32)
+    # On Windows Ice ships was a NuGet package, so we need to find the package prefix directory.
```
pepone, line 4:
```suggestion
    # On Windows Ice ships as a NuGet package, so we need to find the package prefix directory.
```?

pepone (approved):

Issue #3602, opened 2025-02-26
Starting an icegridnode without an icegridregistry


bernardnormier:
This should work fine, since the node just waits for the registry to be up.

However, when I do that, I get rather scary logging:

```
 Greeter % icegridnode --Ice.Config=config.node
^C%                                                                                                                                      bernard@studio Greeter % icegridnode --Ice.Config=config.node
-! 02/26/25 11:37:07.563 icegridnode: warning: couldn't reach the IceGrid registry (this is expected if it's down, otherwise please check the value of the Ice.Default.Locator property):
   src/Ice/Network.cpp:1834 Ice::ConnectionRefusedException connection refused
   stack trace:
     0 Ice::LocalException::LocalException(char const*, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>) in libIce.38a0.dylib
     1 Ice::SyscallException::SyscallException(char const*, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>) in libIce.38a0.dylib
     2 Ice::SocketException::SocketException(char const*, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>) in libIce.38a0.dylib
     3 Ice::ConnectFailedException::ConnectFailedException(char const*, int, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>) in libIce.38a0.dylib
     4 Ice::ConnectionRefusedException::ConnectionRefusedException(char const*, int) in libIce.38a0.dylib
     5 Ice::ConnectionRefusedException::ConnectionRefusedException(char const*, int) in libIce.38a0.dylib
     6 IceInternal::doFinishConnect(int) in libIce.38a0.dylib
     7 IceInternal::StreamSocket::connect(IceInternal::Buffer&, IceInternal::Buffer&) in libIce.38a0.dylib
     8 IceInternal::TcpTransceiver::initialize(IceInternal::Buffer&, IceInternal::Buffer&) in libIce.38a0.dylib
     9 IceInternal::IdleTimeoutTransceiverDecorator::initialize(IceInternal::Buffer&, IceInternal::Buffer&) in libIce.38a0.dylib
    10 Ice::ConnectionI::initialize(IceInternal::SocketOperation) in libIce.38a0.dylib
    11 Ice::ConnectionI::message(IceInternal::ThreadPoolCurrent&) in libIce.38a0.dylib
[...]

PR #3633, opened 2025-03-05
Fix duplicate thread::join in IceGrid

bernardnormier:
See #3602.

A std::thread can only be joined once.

### DIFF
--- a/cpp/src/IceGrid/SessionManager.h
+++ b/cpp/src/IceGrid/SessionManager.h
@@ -263,7 +263,18 @@ namespace IceGrid
             _condVar.notify_all();
         }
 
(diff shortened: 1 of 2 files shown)


cpp/src/IceGrid/SessionManager.h
```diff
@@ -263,7 +263,18 @@ namespace IceGrid
             _condVar.notify_all();
         }
 
-        void join() { _thread.join(); }
+        void join()
+        {
+            std::thread thread;
```
externl, line 268: What happens on the next call after _thread is moved. Is it even usable?

externl (approved):

pepone (approved):

PR #3636, opened 2025-03-06
Slice cleaning

ayoitslilith:
Cleans up the formatting in the .ice files using an automated process combined with manual review.

- Whitespaces near brackets should now be removed.
- Indentation should be more normalized - the outermost module also now counts for indentation.
- Some nested modules have been cleaned.

Please see [this pastebin page](https://pastebin.com/2hcUq15H) for the utilized code and detailed output.


cpp/test/IceStorm/federation/Event.ice
```diff
@@ -4,10 +4,8 @@
...
-{
-    void pub(string data);
-}
-
-}
+    interface Event
+    {
+        void pub(string data);
+    }
+}
```
InsertCreativityHere, line 11: Also, we generally want our files to have a single newline (`\n`) at the end.
If you look at the `diff` in GitHub, you'll see this red-dash-in-a-circle at the end of a file.
This symbol means that that final newline is missing.

ayoitslilith: I erroneously thought that was what was causing the whitespace. I'll go back and fix that and try to snag the whitespace another way.

ayoitslilith: I was running a second pass, but I think I mistook something I was doing for adding the whitespace back rather than just indicating the newlines. It seems to be working fine now, thankfully.

cpp/src/IceStorm/DBTypes.ice
```diff
@@ -15,7 +15,7 @@ module IceStormElection
 module IceStorm
 {
     dictionary<SubscriberRecordKey, SubscriberRecord> SubscriberRecordDict;
-
+
```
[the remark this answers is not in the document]
ayoitslilith: The whitespace should be removed, and the lower issue mentioned with newlines should be fixed.

externl (changes requested):
Overall looks good! There are extra paths in the err files that should be removed. Some whitespace was lost in a few files (I commented).

cpp/test/Slice/errorDetection/DeclaratorMissing.err
```diff
@@ -1,4 +1,4 @@
-DeclaratorMissing.ice:8: missing parameter name
-DeclaratorMissing.ice:8: missing parameter name
-DeclaratorMissing.ice:9: missing parameter name
-DeclaratorMissing.ice:9: missing parameter name
+C:/Users/Lilith/Desktop/repos/ice/cpp/test/Slice/errorDetection/DeclaratorMissing.ice:7: missing parameter name
```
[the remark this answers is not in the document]
InsertCreativityHere: @externl is correct, these should only have the filenames, to avoid leaking any personal file structures :)

But, I am also very surprised the test still passes.
Looking at the code though, yes, we tolerate any path:
```
regex2 = re.compile("^.*(?=" + os.path.basename(file) + ")")
```
`^.*` matches _anything_ up to the filename. We should remove this IMO.

cpp/test/Glacier2/router/Callback.ice
```diff
@@ -6,41 +6,39 @@
...
-interface CallbackReceiver
-{
-    void callback();
+    interface CallbackReceiver
+    {
+        void callback();
 
-    void callbackEx()
+        void callbackEx()
         throws CallbackException;
```
InsertCreativityHere, line 20: Sorry in advance, because this is probably going to make your script much more difficult ._.

But this is a special case where we actually want _extra_ indentation, because this is really 1 single operation, that's been split onto 2 separate lines. And when this happens, we want to indent everything until the end of the function.

For example, if you had a super long function call, and had to split it onto multiple lines, you would do this:
```
doSomething(
    verylongarg1,
    verlylongarg2
    verylongarg3);
```
So it's clear where the function ends. Instead of this, where it's just one big blob:
```
doSomething(
verylongarg1,
verlylongarg2
verylongarg3);
```

----

TL;DR

When a single operation is split up onto multiple lines, the following lines should be indented 1 level over.
So this should look like:
```
        void callbackEx()
            throws CallbackException;
```
Tbf, this is a trivial example... because this fits on a single line, and actually should be:
```
        void callbackEx() throws CallbackException;
```
But this is a general problem your script needs to try and tolerate... Let me check for a better example.

cpp/test/Ice/optional/Test.ice
```diff
@@ -6,279 +6,277 @@
...
+        G1 gg1;
+    }
+
+    interface Initial
+    {
+        void shutdown();
+
+        ["marshaled-result"] Object pingPong(Object o);
+
+        void opOptionalException(optional(1) int a, optional(2) string b)
```
InsertCreativityHere, line 181: Here for example, this operation is very long, and so it makes more sense for it to be split up onto 2 lines.

This whole file is a good example to see how we handle super-long functions like this.
The original file's indentation was correct, except that it didn't add indentation for the outer-most module.

cpp/test/Ice/stringConverter/Test.ice
InsertCreativityHere, line 1: This file is interesting.
There used to be (and should be) a blank line between the `exception` and the `interface`, but it looks like your script ate it. Not sure what happened there.

cpp/test/Slice/deprecated/Test.ice
InsertCreativityHere, line 1: This file is an even more clear example.
It was already correctly formatted actually, but the script removed these lines around the `enum MyEnum`,
when it shouldn't have.

It seems like the weekend occurs when a Slice definition fits onto a single line:
```
enum MyEnum { A, B, C } // Script eats spaces

enum MyEnum // script leaves alone
{
    A,
    B,
    C
}
```

Both formats are fine for the record. The rule of thumb is a Slice definition should use the multiline syntax unless it's very very small. Like an empty exception: `exception {}` yeah. No reason to waste extra lines. Or very short enums like this one.

cpp/test/Slice/errorDetection/NestedModule.ice
```diff
@@ -5,7 +5,6 @@ module All::Good::Here
 {
```
InsertCreativityHere, line 5: Same here, it ate lines.
Indeed, seems to be something about single-line definitions.

ayoitslilith:
This latest commit isn't working and I know why, but we're going out to lunch - I'll amend it when I get back and make sure everything's properly in order. At the very least, I know that the script is no longer eating more lines than it should.

bernardnormier (approved):

swift/test/Slice/escape/Clash.ice
```diff
@@ -1,57 +1,55 @@
...
-    int result;
-    string istr;
-    string ostr;
-    string inS;
-    string proxy;
-}
+            void opOut(out string context, out string current, out string response, out string ex,
+                out string sent, out string cookie, out string sync, out string result, out string istr,
+                out string ostr, out optional(1) string proxy);
+            }
```
InsertCreativityHere, line 22: While the script is correctly indenting for '(' characters,
it seems like it isn't dedenting (the opposite of indenting) after it hits ')' here,

causing it permanently gain an indentation level for the rest of the file.

externl (approved):

externl:
Slice error tests need a few fixes still.

cpp/src/IceGrid/Internal.ice
```diff
@@ -13,420 +13,418 @@
...
+        void replicaAdded(InternalRegistry* replica);
+
+        /// Notification that a replica has been removed. The node should destroy the session to this replica.
+        void replicaRemoved(InternalRegistry* replica);
+    }
+
+    interface Node extends FileReader, ReplicaObserver
+    {
+        /// Load the given server. If the server resources weren't already created (database environment directories,
+            /// property files, etc), they will be created. The returned proxy is never null.
```
InsertCreativityHere, line 212: The 2nd line got indented. Possible because it contains '(' and ')'?
Not worth addressing in the script, but still should be fixed:
```suggestion
        /// Load the given server. If the server resources weren't already created (database environment directories,
        /// property files, etc), they will be created. The returned proxy is never null.
```

cpp/src/IceGrid/Internal.ice
```diff
@@ -13,420 +13,418 @@
...
+        /// Destroy the given server.
+        ["amd"] idempotent void destroyServer(string name, string uuid, int revision, string replicaName)
+            throws DeploymentException;
+
+        /// Destroy the server if it's not active.
+        ["amd"] idempotent void destroyServerWithoutRestart(string name, string uuid, int revision, string replicaName)
+            throws DeploymentException;
+
+        /// Establish a session to the given replica, this method only returns once the registration was attempted (unlike
+            /// replicaAdded below).
```
InsertCreativityHere, line 239: Same:
```suggestion
        /// Establish a session to the given replica, this method only returns once the registration was attempted (unlike
        /// replicaAdded below).
```

cpp/src/IceGrid/Internal.ice
```diff
@@ -13,420 +13,418 @@
...
+        /// Ask the registry to load the servers on the node.
+        ["amd"] ["cpp:const"] idempotent void loadServers();
+
+        /// Get the name of the servers deployed on the node.
+        ["cpp:const"] idempotent Ice::StringSeq getServers();
+
+        /// Wait for the application update to complete (the application is completely updated once all the registry
+            /// replicas have been updated). This is used by the node to ensure that before to start a server all the
+        /// replicas have the up-to-date descriptor of the server.
+        ["amd"] ["cpp:const"] void waitForApplicationUpdate(string application, int revision);
```
InsertCreativityHere, line 289: Same:
```suggestion
        /// Wait for the application update to complete (the application is completely updated once all the registry
        /// replicas have been updated). This is used by the node to ensure that before to start a server all the
        /// replicas have the up-to-date descriptor of the server.
        ["amd"] ["cpp:const"] void waitForApplicationUpdate(string application, int revision);
```

InsertCreativityHere (approved):
I reviewed all 377 files, and they look correct, except for 11 small edge-cases!
I don't think they're worth fixing at the script-level, but I added suggestions for them, so you can hopefully just click "commit suggestion".

Otherwise, it does seem like there are still some `.err` files that don't have the correct line number still.

But all in all, it's nice to finally have these Slice files consistently formatted!

PR #3650, opened 2025-03-10
Add Slice Tools Gradle plugin to build Java and Android projects

pepone:
This PR adds a new Gradle plugin `com.zeroc.ice.slice-tools` that can be used to integrate Slice to Java with gradle Java and Android projects. It is a replacement for `com.zeroc.gradle.ice-builder.slice`.

The new plugin includes the slice2java compilers, and Slice files. See BUILDING.md for details.

In contrast with the old plugin, the new plugin is tied to the given Ice version, and use the same version as Ice. This is because it bundles slice2java and the Ice Slice files. You can still use a different compiler setting toolsPath, but this would be mostly for development.

The configuration for the new plugin is a bit different and it is explained in the README.

-- Update 1 --

Moved the new slice-tools plugin to java/tools/slice-tools
Remove the separate Gradle wrapper used by the plug-in

### DIFF
--- a/.github/workflows/build-maven-packages.yml
+++ b/.github/workflows/build-maven-packages.yml
@@ -47,6 +47,7 @@ jobs:
       - name: Update Ice Version
         run: |
           sed -i "s/^iceVersion = .*/iceVersion = ${{ inputs.ice_version }}/" java/gradle.properties
+          sed -i "s/^sliceToolsVersion = .*/sliceToolsVersion = ${{ inputs.ice_version }}/" java/tools/slice-tools/gradle.properties
         if: inputs.ice_version != ''
 
       - name: Build Java Packages


tools/gradle/plugin/src/main/kotlin/com/zeroc/ice/SliceToolsResourceExtractor.kt
```diff
@@ -0,0 +1,118 @@
...
+     * Slice files are not present in the resources.
+     */
+    fun extractSliceFiles(project: Project): String? {
+        val pluginVersion = getPluginVersion(project)
+
+        val sliceSourcePath = "/slice/"
+        val sliceTargetDir = File(project.gradle.gradleUserHomeDir, "native/$pluginName-$pluginVersion/slice")
+
+        if (!sliceTargetDir.exists()) {
+            val resourceURL = this::class.java.getResource(sliceSourcePath)
```
externl, line 63: This is some sort of special resource that we can't just use without "copying"?. I.e. returning `File(resourceURL.toURI())`

tools/gradle/plugin/src/main/kotlin/com/zeroc/ice/SliceToolsResourceExtractor.kt
```diff
@@ -0,0 +1,118 @@
...
+            os.contains("nix") || os.contains("nux") -> "linux"
+            else -> os
+        }
+    }
+
+    private fun detectArch(): String {
+        val arch = System.getProperty("os.arch").lowercase()
+        return when {
+            arch.contains("64") || arch.contains("x86_64") -> "x64"
+            arch.contains("arm") || arch.contains("aarch64") -> "arm64"
```
externl, line 114: Should we just bundle the compilers in directories that match what Kotlin uses for os.arch? Or would that just be too painful.

externl (commented):
Looks good. First pass comments.

tools/gradle/plugin/src/functionalTest/kotlin/com/zeroc/ice/SliceToolsJavaPluginFunctionalTest.kt
```diff
@@ -0,0 +1,46 @@
...
+        propsFile.writeText("sliceToolsVersion=3.8.0-alpha.0")
+        buildFile.writeText(
+            """
+            plugins {
+                id('java')
+                id('com.zeroc.ice.slice-tools')
+            }
+
+            slice {
+                toolsPath = "/opt/Ice-3.8.0a/bin"
```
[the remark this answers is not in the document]
pepone: The test is mostly a place holder, doesn't require it to be installed.

tools/gradle/plugin/src/main/kotlin/com/zeroc/ice/SliceTask.kt
```diff
@@ -0,0 +1,522 @@
...
+
+            val stdoutReader = process.inputStream.bufferedReader()
+            val stderrReader = process.errorStream.bufferedReader()
+
+            val output = stdoutReader.readText()
+            val errors = stderrReader.readText()
+            val exitCode = process.waitFor()
+
+            if (exitCode != 0) {
+                logger.error("Failed to compile Slice files: $errors")
```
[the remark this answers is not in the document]
pepone: Yes I plan to improve error message in a follow up, we need to some processing to them.

tools/gradle/plugin/src/main/kotlin/com/zeroc/ice/SliceTask.kt
```diff
@@ -0,0 +1,522 @@
...
+     * - The output directory (`--output-dir`)
+     * - Default compiler arguments
+     * - Additional arguments provided by the caller
+     * - The list of Slice files to compile
+     *
+     * @param sliceFiles The list of `.ice` Slice files to compile. Defaults to an empty list.
+     * @param additionalArgs Additional command-line arguments to pass to `slice2java`. Defaults to an empty list.
+     * @return A list of strings representing the full `slice2java` command.
+     */
+    private fun getSlice2JavaCommand(sliceFiles: List<File> = emptyList(), additionalArgs: List<String> = emptyList()): List<String> {
```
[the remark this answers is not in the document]
pepone: Yes we call it with an empty list to get the used args, which are saved to compare with previous args. In this case we just want the args without slice files.

java/gradle/ice.gradle
```diff
@@ -9,18 +9,23 @@ buildscript {
...
 }
 
-apply plugin: com.zeroc.gradle.icebuilder.slice.SlicePlugin
 apply plugin: com.diffplug.gradle.spotless.SpotlessPlugin
 
 slice {
-    cppConfiguration = project.cppConfiguration
-    cppPlatform = project.cppPlatform
-    iceHome = new File(project.ext.topSrcDir).getCanonicalPath()
+    includeSearchPath = files("${project.ext.topSrcDir}/slice")
```
bernardnormier, line 19: Should we instead build the slice-tools plugin first and then use it? 
As opposed to having this separate slice logic when building Ice Java?

java/build.gradle
```diff
@@ -1,12 +1,17 @@
...
+}
+
 subprojects {
     project.ext.topSrcDir = "$rootProject.projectDir/.."
 
     project.version = iceVersion
     project.group = "com.zeroc"
 
     apply plugin: 'java'
+    apply plugin: 'com.zeroc.ice.slice-tools'
```
bernardnormier, line 14: Should this move into ice.gradle just like the spotless plugin?

tools/gradle/plugin/src/main/kotlin/com/zeroc/ice/SliceToolsResourceExtractor.kt
```diff
@@ -0,0 +1,118 @@
...
+                targetFile.mkdirs()
+            } else {
+                file.inputStream().use { input ->
+                    targetFile.outputStream().use { output -> input.copyTo(output) }
+                }
+            }
+        }
+    }
+
+    private fun detectOS(): String {
```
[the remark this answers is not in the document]
pepone: We throw when the actual path is computed if it result in a not supported combination.

java/src/com.zeroc.icegrid/build.gradle
```diff
@@ -4,9 +4,11 @@ project.ext.displayName = "IceGrid"
 project.ext.moduleName = "com.zeroc.icegrid"
 project.ext.description = "Locate, deploy, and manage Ice servers"
 
-slice {
```
bernardnormier, line 7: It's not clear to me why this syntax change is needed. Will this affect build.gradle written by users?

tools/gradle/plugin/src/main/kotlin/com/zeroc/ice/SliceToolsResourceExtractor.kt
pepone: Not sure, here I just stick which what we do for other builds, this mapping of names is simple enough and keep same layout for all tools.

tools/gradle/BUILDING.md
```diff
@@ -0,0 +1,83 @@
+# Building Ice Slice Tools Gradle Plugin
```
bernardnormier, line 1: Building the Ice Slice Tools Gradle Plugin

or maybe "the Ice Slice Tools for Gradle"?

tools/gradle/plugin/src/functionalTest/kotlin/com/zeroc/ice/SliceToolsJavaPluginFunctionalTest.kt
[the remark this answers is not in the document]
pepone: The main integration test now, is that we build Ice sources, test and android tests with the plug-in. The configuration is also much simpler than with the previous build, mostly a few Gradle properties, but there isn't that match logic for detecting things.

tools/gradle/plugin/src/main/kotlin/com/zeroc/ice/SliceExtension.kt
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) ZeroC, Inc.
+
+package com.zeroc.ice
```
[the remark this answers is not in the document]
pepone: In Gradle terms is a plugin. The plugin name is `Ice Slice Tools` and the plug-in id is `com.zeroc.ice.slice-tools`. The name is the informal way we refer to it in README and other docs.

tools/gradle/BUILDING.md
```diff
@@ -0,0 +1,83 @@
...
+available at runtime, they must be copied to the plugin’s resources directory.
+
+The plugin expects the following structure within src/main/resources/:
+
+```shell
+src/main/resources/slice         # Ice Slice files
+src/main/resources/macos-arm64   # slice2java compiler for macOS (Apple Silicon)
+src/main/resources/linux-arm64   # slice2java compiler for Linux (ARM64)
+src/main/resources/linux-x64     # slice2java compiler for Linux (x86_64)
+src/main/resources/windows-x64   # slice2java compiler for Windows (x86_64)
```
bernardnormier, line 35: I'd use `x64` for the comment. x86_64 is unusual on Windows.

tools/gradle/BUILDING.md
```diff
@@ -0,0 +1,83 @@
...
+
+```shell
+./gradlew publishToMavenLocal
+```
+
+This allows other projects on your machine to use the locally published version.
+
+### Publish to a Remote Repository (Maven Central, Artifactory, etc.)
+
+To publish the plugin to a remote repository, ensure your **Gradle properties** (`~/.gradle/gradle.properties`) contain the necessary credentials:
```
bernardnormier, line 54: Line too long

tools/gradle/plugin/src/main/kotlin/com/zeroc/ice/SliceExtension.kt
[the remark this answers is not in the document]
externl: Isn't Ice (the RPC) also `com.zeroc.Ice`. I know the casing is different, but shouldn't we this plugin use a different package?

tools/gradle/plugin/src/functionalTest/kotlin/com/zeroc/ice/SliceToolsJavaPluginFunctionalTest.kt
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) ZeroC, Inc.
+
+package com.zeroc.ice
```
bernardnormier, line 3: Should we use a subpackage name like `com.zeroc.ice.slice` or `com.zeroc.ice.slice.tools`?

Putting these classes directly com.zeroc.ice looks odd to me.

tools/gradle/BUILDING.md
```diff
@@ -0,0 +1,83 @@
...
+
+To publish the plugin to your local Maven repository (`~/.m2/repository`), run:
+
+```shell
+./gradlew publishToMavenLocal
+```
+
+This allows other projects on your machine to use the locally published version.
+
+### Publish to a Remote Repository (Maven Central, Artifactory, etc.)
```
bernardnormier, line 52: I am all for documenting the local publish. But do we need to document the "remote publish"? Presumably we're the only ones doing that.

tools/gradle/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Configuration
+
+The plugin provides the `slice` extension for configuring Slice compilation settings.
+
+### Slice Extension Properties
+
+| Property Name       | Description                                                                                                   |
+|---------------------|---------------------------------------------------------------------------------------------------------------|
+| `toolsPath`         | Path to the directory containing the `slice2java` compiler. If not set, the plugin uses the bundled compiler. |
```
bernardnormier, line 48: Do we need this toolsPath escape?

I feel we should always use the slice2java in the plugin. And we want another slice2java, just build your local plugin (which should be easy to do).

tools/gradle/plugin/src/main/kotlin/com/zeroc/ice/SliceTask.kt
```diff
@@ -0,0 +1,522 @@
...
+     * If compilation arguments change, a full recompilation is triggered.
+     */
+    @TaskAction
+    fun compileSlice() {
+        val sliceFiles = inputFiles.files.toList()
+        val outputDir = output.get().asFile
+
+        if (sliceFiles.isEmpty()) {
+            // No Slice files to compile. Ensure the output directory is clean by deleting any stale files
+            outputDir.deleteRecursively()
```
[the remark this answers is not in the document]
pepone: Need to test.

java/src/com.zeroc.icegrid/build.gradle
pepone: > It's not clear to me why this syntax change is needed.

Is not strictly need, is part of updating the plugin to use a more modern Gradle API.

> Will this affect build.gradle written by users?

Yes. The new plugin uses a Gradle source set for configuring Slice files. The old plugin uses a custom object (Java) https://github.com/zeroc-ice/ice-builder-gradle/blob/main/src/main/groovy/com/zeroc/gradle/icebuilder/slice/Java.groovy#L5-L15

java/gradle/ice.gradle
pepone: This currently rely on using

```
pluginManagement {
    includeBuild "../tools/gradle"
}
```

And there is no extra step for publishing the plug-in locally.

java/build.gradle
pepone: We can, our gradle files can use a little cleanup.

tools/gradle/plugin/src/functionalTest/kotlin/com/zeroc/ice/SliceToolsJavaPluginFunctionalTest.kt
pepone: We can but it would be part of the plugin ID, that users have to use to apply the plugin.

tools/gradle/README.md
pepone: This is similar to what we do with MSBuild, we don't publish NuGet packages and consume them.

I think the extra step to first publish a plugin for local use would add complexity to our builds.

* We have to build slice2java
* Copy it to the plugin resource folder
* Publish the plugin to a local maven repository

tools/gradle/BUILDING.md
bernardnormier: What about a simpler:  Slice Tools for Ice. The Gradle plugin part moves to the description.

tools/gradle/plugin/src/functionalTest/kotlin/com/zeroc/ice/SliceToolsJavaPluginFunctionalTest.kt
pepone: Good them, we just need to settle on what names to use.

java/gradle/ice.gradle
bernardnormier: Ok.

java/src/com.zeroc.icegrid/build.gradle
bernardnormier: So, if I use the `slice {` syntax with Ice 3.7, do I need to change this syntax when I upgrade to 3.8, or this change is recommended but optional - the old syntax still works?

pepone: The new plugin uses the new syntax. The old plugin would still work there is not a hard pressure for user to upgrade.

tools/gradle/README.md
```diff
@@ -0,0 +1,193 @@
+# Ice Slice Tools Gradle Plugin
```
bernardnormier, line 1: I think it would be clearer to have a tools subdir per language, like:

- java/tools or java/tools/gradle
- csharp/tools/ZeroC.Ice.Slice.Tools.Csharp
- cpp/tools/ZeroC.Ice.Slice.Tools.Cpp

with the common ZeroC.Ice.Slice.Tools probably in csharp/tools.

tools/gradle/settings.gradle.kts
```diff
@@ -0,0 +1,9 @@
+/*
+ * This file was generated by the Gradle 'init' task.
+ *
+ * The settings file is used to specify which projects to include in your build.
+ * For more detailed information on multi-project builds, please refer to https://docs.gradle.org/8.13/userguide/multi_project_builds.html in the Gradle documentation.
+ */
+
+rootProject.name = "slice-tools-java"
```
bernardnormier, line 8: Not immediately clear why we have `-java` in this name.

tools/gradle/README.md
pepone: Fine with me. I started with the top-level tools because that is what we did in IceRPC. It was also convenient to share code among the C++ and C# projects.

tools/gradle/BUILDING.md
pepone: removed.

tools/gradle/plugin/src/main/kotlin/com/zeroc/ice/SliceTask.kt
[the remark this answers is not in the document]
pepone: Works fine

```
jose@macbookpro kotlin-test % kotlin   
Welcome to Kotlin version 2.1.10 (JRE 23.0.2)
Warning: this REPL implementation is deprecated and will be removed soon.
Warning: REPL is not compatible with the Kotlin version 2.1.10, using '-language-version 1.9'.
Type :help for help, :quit for quit
>>> import java.io.File
>>> val foo = File("foo")
>>> foo.exists()
res3: kotlin.Boolean = true
>>> foo.deleteRecursively()
res4: kotlin.Boolean = true
>>> foo.exists()
res5: kotlin.Boolean = false
>>> foo.deleteRecursively()
res6: kotlin.Boolean = true
```

java/tools/slice-tools/README.md
```diff
@@ -0,0 +1,193 @@
+# Ice Slice Tools Gradle Plugin
+
+The **Ice Slice Tools Gradle Plugin** integrates the **Slice-to-Java compiler (`slice2java`)** into Gradle-based
```
bernardnormier, line 3: I would call this code the "Slice Tools for Ice and Java". No need for Gradle in the name. 

Likewise, the other integrations would be the "Slice Tools for Ice and C#" and the "Slice Tools for Ice and C++".

java/tools/slice-tools/README.md
```diff
@@ -0,0 +1,193 @@
...
+**Java** and **Android** projects. It automates the compilation of Slice (`.ice`) files into Java source files and
+integrates seamlessly with the Gradle build system.
```
bernardnormier, line 5: Should we add a note like: it replaces the Ice Builder Gradle plugin (use official name). While you can still use the Builder with 3.8, we recommend using the Slice tools with Ice 3.8.

bernardnormier (approved):

java/tools/slice-tools/README.md
bernardnormier: I think including Gradle / MSBuild etc. in the plugin/integration README title makes the name hard to read / understand.

In practice, we only provide a single integration per language, so there is not much value in including it in the title:
- C# => MSBuild
- C++ => MSBuild
- Java => Gradle

I prefer "Slice Tools for Ice" over "Ice Slice Tools" as the README title. "Slice Tools for Ice Gradle Plugin" has too many words.

My concern with "Slice Tools for Ice" as a README title is I don't want to give the impression that there is a single "Slice Tools for Ice" code/plugin/integration that works for Java, C# and C++. We provide programming language - specific integrations.

We can fix these README titles in a follow-up PR.

externl: This is not really true. We have CMake support as well. We also have the builders for maven and ant. We may want to update these to work with Ice 3.8.

java/tools/slice-tools/BUILDING.md
```diff
@@ -0,0 +1,50 @@
...
+## Prerequisites
+
+- JDK 17 or greater
+
+## Build
+
+The plugin can be built using the following Gradle command:
+
+```shell
+../../gradlew build
```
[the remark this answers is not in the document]
pepone: By default is build as part of Java. But you can still build it separately. Here I use `.../../gradlew build` to keep everything relative to the current dir.

You can also use Gradle `-p` option and do `./gradlew -p tools/slice-tools` from the top-level java dir to just build this part.

We already have to copies of the Gradle wrapper `java/gradlew` and `test/android/controller/gradlew`. I wanted to avoid adding a third copy.

I also want to remove android copy, but that would be a separate PR.

InsertCreativityHere (approved):
Looks good to me!

Issue #3642, opened 2025-03-07
Generate improved depends file with slice2cpp


externl:
When generating a BNF style depends file with --depend, the `slice2cpp` compiler currently (last of 3.7) generates:

```shell
> slice2cpp Foo.ice --depend
Foo.: \
  Foo.ice \
  <dependencies of foo>
```

If you specify an extension using metadata or on the command line we'll generate `Foo.<ext>:` instead.

Our current output is isn't really usable as the file `Foo.` does not exist. think it would be preferable if the target was the slice file itself:

```shell
Foo.ice: \
  <dependencies>
```

Note that in our C++ Makefiles we modify the output to make it usable.

https://github.com/zeroc-ice/ice/blob/089da7b29ffee6332727a78976f31ceb1035ea85/config/Make.project.rules#L179-L185

Fixing this would make the generated depends file usable out of the box by our new CMake support files.

>[!NOTE]
>This does not affect --depend-xml

pepone:
Compilers are not consistent with each other here:

```
jose@mac ice % ./cpp/bin/slice2cpp --depend slice/Ice/Locator.ice 
Locator.h: \
 slice/Ice/Locator.ice \
 /Users/jose/Documents/3.8/ice/slice/Ice/Identity.ice

jose@mac ice % ./cpp/bin/slice2cs --depend slice/Ice/Locator.ice 
Locator.cs: \
 slice/Ice/Locator.ice \
 /Users/jose/Documents/3.8/ice/slice/Ice/Identity.ice

jose@mac ice % ./cpp/bin/slice2java --depend slice/Ice/Locator.ice 
slice/Ice/Locator.ice: \
 /Users/jose/Documents/3.8/ice/slice/Ice/Identity.ice

jose@mac ice % ./cpp/bin/slice2js --depend slice/Ice/Locator.ice 
Locator.js: \
 slice/Ice/Locator.ice \
 /Users/jose/Documents/3.8/ice/slice/Ice/Identity.ice

jose@mac ice % ./cpp/bin/slice2matlab --depend slice/Ice/Locator.ice 
slice/Ice/Locator.ice: \
 /Users/jose/Documents/3.8/ice/slice/Ice/Identity.ice

jose@mac ice % ./cpp/bin/slice2php --depend slice/Ice/Locator.ice 
Locator.php: \
 slice/Ice/Locator.ice \
 /Users/jose/Documents/3.8/ice/slice/Ice/Identity.ice

[...]

PR #3651, opened 2025-03-10
Fix Missing Extension with `--depend` and `errorDetection` Path Checking

### LINKED ISSUE #3643: Update slice tests to not be tolerant of paths in .err files
@externl is correct, these should only have the filenames, to avoid leaking any personal file structures :)

But, I am also very surprised the test still passes.
Looking at the code though, yes, we tolerate any path:
```
regex2 = re.compile("^.*(?=" + os.path.basename(file) + ")")
```
`^.*` matches _anything_ up to the filename. We should remove this IMO.

_Originally posted by @InsertCreativityHere in https://github.com/zeroc-ice/ice/pull/3636#discussion_r1985232959_

InsertCreativityHere:
This PR fixes two issues:

1) #3642 points out that by default `--depend` is generating output with empty file extensions. This is a bug IMO.
The `--depend` code only checks for `header-ext` metadata, and falls-back to `""` is none is present.
It should be falling back to `headerExtension` like the generated code (which checks for any CLI options, and falls-back to `.h`).

2) Fixes #3643. The `errorDetection` test no longer tolerates paths in the `.err`.

### DIFF
--- a/cpp/test/Slice/errorDetection/test.py
+++ b/cpp/test/Slice/errorDetection/test.py
@@ -48,8 +48,8 @@ class SliceErrorDetectionTestCase(ClientTestCase):
                     regex2 = re.compile("^.*(?=" + os.path.basename(file) + ")")
                     i = 0
                     while i < len(lines1):
-                        line1 = regex2.sub("", lines1[i]).strip()
-                        line2 = regex2.sub("", lines2[i]).strip()
+                        line1 = regex2.sub("", lines1[i]).strip() # Actual output from slice2cpp
+                        line2 = lines2[i].strip()                 # Expected output from .err file
                         if line1 != line2:
                             raise RuntimeError(
                                 'failed! (line1 = "{0}", line2 = "{1}"'.format(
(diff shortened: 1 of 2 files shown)


cpp/test/Slice/errorDetection/test.py
```diff
@@ -48,8 +48,8 @@ def runClientSide(self, current):
                     regex2 = re.compile("^.*(?=" + os.path.basename(file) + ")")
                     i = 0
                     while i < len(lines1):
-                        line1 = regex2.sub("", lines1[i]).strip()
-                        line2 = regex2.sub("", lines2[i]).strip()
+                        line1 = regex2.sub("", lines1[i]).strip() # Actual output from slice2cpp
+                        line2 = lines2[i].strip()                 # Expected output from .err file
```
InsertCreativityHere, line 52: Now we only strip paths out of the `slice2cpp` output. Not the `err` file`.

externl (approved):

pepone (approved):

PR #3657, opened 2025-03-10
Fix IceStorm crash when database dir doesn't exist

### LINKED ISSUE #3639: Provide diagnostics when IceStorm database directory does not exist
IceStorm currently crashes without any  message:

```shell
% icebox --Ice.Config=config.icebox                      
libc++abi: terminating
zsh: abort      icebox --Ice.Config=config.icebox
```

bernardnormier:
Fixes #3639.

### DIFF
--- a/cpp/src/IceStorm/Instance.cpp
+++ b/cpp/src/IceStorm/Instance.cpp
@@ -134,7 +134,7 @@ Instance::~Instance()
     // Usually they are called before destruction, but in some cases they are not, for example when the constructor
     // of PersistentInstance fails because the database directory does not exist.
     shutdown();
-    destroy();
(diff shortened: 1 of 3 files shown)


cpp/src/IceStorm/Instance.cpp
```diff
@@ -129,7 +129,13 @@ Instance::Instance(
     }
 }
 
-Instance::~Instance() = default;
+Instance::~Instance()
+{
+    // Usually they are called before destruction, but in some cases they are not, for example when the constructor
+    // of PersistentInstance fails before the database directory does not exist.
```
externl, line 135:
```suggestion
    // of PersistentInstance fails in the case where the database directory does not exist.
```

externl (approved):

pepone (approved):

PR #3660, opened 2025-03-10
Suppress doc generation for Disp_ never implemented by the user (C#)

bernardnormier:
This PR suppresses doc generation for Disp_ classes that the user should never implement. For example, IceGrid::Admin.

Including these classes in the API reference is naturally confusing.

### DIFF
--- a/csharp/docfx/docfx.json
+++ b/csharp/docfx/docfx.json
@@ -6,7 +6,9 @@
           "files": ["src/*/*.csproj"],
           "exclude": [
             "src/Ice/Internal/**",
-            "src/Ice/UtilInternal/**"
+            "src/Ice/UtilInternal/**",
+            "src/IceDiscovery/**",
+            "src/IceLocatorDiscovery/**"
           ],
           "src": "../"
         }
--- a/slice/Glacier2/Router.ice
+++ b/slice/Glacier2/Router.ice
@@ -31,6 +31,7 @@ module Glacier2
     }
 
     /// The Glacier2 specialization of the <code>Ice::Router</code> interface.
+    ["cs:attribute:System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)"]
     interface Router extends Ice::Router
     {
         /// This category must be used in the identities of all of the client's callback objects. This is necessary in
--- a/slice/Glacier2/Session.ice
+++ b/slice/Glacier2/Session.ice
@@ -41,6 +41,7 @@ module Glacier2
     /// {@link Session}.
     /// @see Session
     /// @see SessionControl
+    ["cs:attribute:System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)"]
     interface StringSet
     {
         /// Add a sequence of strings to this set of constraints. Order is not preserved and duplicates are implicitly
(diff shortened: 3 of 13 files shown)


csharp/docfx/docfx.json
```diff
@@ -6,7 +6,9 @@
           "files": ["src/*/*.csproj"],
           "exclude": [
             "src/Ice/Internal/**",
-            "src/Ice/UtilInternal/**"
+            "src/Ice/UtilInternal/**",
+            "src/IceDiscovery/**",
```
bernardnormier, line 10: These namespaces have no public API.

slice/Glacier2/Router.ice
```diff
@@ -31,6 +31,7 @@ module Glacier2
     }
 
     /// The Glacier2 specialization of the <code>Ice::Router</code> interface.
+    ["cs:attribute:System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)"]
```
bernardnormier, line 34: cs:attribute has no effect on the generated Prx and PrxHelper. I assume this is on purpose!

csharp/docfx/docfx.json
```diff
@@ -5,13 +5,14 @@
         {
           "files": ["src/*/*.csproj"],
           "exclude": [
```
bernardnormier, line 7: As I understand it, it's about suppressing the csproj, which does not help with Ice.Internal / Ice.UtilInternal, since they don't have a csproj.

externl: For some reason I thought this could reference source files as well.

externl (approved):

InsertCreativityHere (approved):
Looks good.
Glad to see there was an easy attribute way to achieve this.

Just for the record, this PR doesn't conflict with #3636, so fine to merge whenever!

slice/Glacier2/Router.ice
InsertCreativityHere: It seems like it is. And I think the behavior makes sense.
Even if they're generated code, proxies aren't really something a user would want to modify with attributes I think.

bernardnormier:
Using cs attributes on interfaces is not the right approach:
 - it doesn't work in other languages 
 - the meaning is ambiguous (which generated constructs does it apply to?)

In new Slice, cs::attribute only apply to constructs that can't be extended using partial: https://docs.icerpc.dev/slice2/language-guide/attributes

bernardnormier:
I reworked the PR to use partial and generate partial where it was missing.

csharp/src/Glacier2/DocGen.cs
```diff
@@ -0,0 +1,50 @@
...
+// Mark some generated server-side classes as EditorBrowsableState.Never so that docfx does not generate documentation
+// pages for them.
+
+using System.ComponentModel;
+
+namespace Glacier2;
+
+#pragma warning disable SA1601 // PartialElementsMustBeDocumented
+
+[EditorBrowsable(EditorBrowsableState.Never)]
```
pepone, line 12: Isn't this attribute to avoid the classes to show up in code completion? If we just want to filter them in docfx I think it can be done in the docfx configuration.

bernardnormier: We really want both.

PR #3664, opened 2025-03-11
Improve logging of slice2java errors

pepone:
This PR improves the logging of errors and warnings from the slice-tools Gradle plugin.


java/tools/slice-tools/src/main/kotlin/com/zeroc/ice/slice/tools/SliceTask.kt
```diff
@@ -371,7 +383,7 @@ abstract class SliceTask @Inject constructor(objects: ObjectFactory) : DefaultTa
 
             // If no generated files exist, force recompilation
             if (firstGeneratedFile == null) {
-                logger.lifecycle("No generated files found for Slice file: $sliceFile (Recompiling)")
```
pepone, line 374: lifecycle is to verbose, is printed in the default output.

externl (approved):

PR #3666, opened 2025-03-11
Fix doc-comment generation in C#

bernardnormier:
This PR includes numerous fixes to slice2cs.


cpp/src/slice2cs/Gen.cpp
```diff
@@ -1110,23 +1199,27 @@ Slice::Gen::TypesVisitor::visitDictionary(const DictionaryPtr& p)
     string name = "global::System.Collections.Generic." + genericType + "<" + keyS + ", " + valueS + ">";
 
     _out << sp;
+    ostringstream summary;
+    summary << "Provides methods to marshal and unmarshal a <c>" << p->name() << "</c>.";
+    writeHelperDocComment(p, summary.str(), "helper class");
```
pepone, line 1204: Should this be "dictionary helper class", like we do for Proxy? otherwise can we just always use "helper class"?

Currently is "helper class" for all but proxy.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -1589,7 +1695,10 @@ Slice::Gen::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     string ns = getNamespace(p);
 
     _out << sp;
-    writeDocComment(p);
+    ostringstream notes;
+    notes << "Use the methods of this interface to invoke operations on a remote Ice object that implements <c>"
```
pepone, line 1699: What about using `<see cref=...` here instead of `<c>`

pepone (approved):

cpp/src/slice2cs/Gen.cpp
```diff
@@ -26,18 +26,83 @@ using namespace IceInternal;
...
+        if (!closeTag)
+        {
+            closeTag = openTag;
+        }
+
+        out << nl << "/// <" << openTag << ">" << comment << "</" << *closeTag << ">";
+    }
+
+    void
+    writeDocLines(Output& out, const string& openTag, const StringList& lines, std::optional<string> closeTag = nullopt)
```
externl, line 46:
```suggestion
    writeDocLines(Output& out, const string& openTag, const StringList& lines, optional<string> closeTag = nullopt)
```

externl (approved):

cpp/src/slice2cs/Gen.cpp
bernardnormier: Yes, we should qualify it: sequence helper class, dictionary helper class etc.

cpp/src/slice2cs/Gen.cpp
bernardnormier: `p->name();` is the Slice interface name. I don't want to reference the server-side C# interface from the proxy.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -2184,6 +2294,11 @@ Slice::Gen::ServantVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     _out << sp;
     emitAttributes(p);
 
+    ostringstream notes;
+    notes << "Your servant class implements this interface by deriving from a generated Disp_ class such as "
+          << "<see cref=\"" << p->mappedName() << "Disp_\" />.";
```
InsertCreativityHere, line 2298: Do we need "such as" here?
`p->mappedName() + "Disp_"` just _is_ the Disp right?
It's not like it's just one of many possible choices the user can implement.
```suggestion
    notes << "Your servant class implements this interface by deriving from the generated Disp_ class "
          << "<see cref=\"" << p->mappedName() << "Disp_\" />.";
```

cpp/src/slice2cs/Gen.cpp
```diff
@@ -655,23 +722,52 @@ Slice::CsVisitor::writeDataMemberInitializers(const DataMemberList& dataMembers)
...
+    const string& generatedType,
+    const string& notes)
+{
+    writeDocLine(_out, "summary", comment);
+
+    // We generate remarks only for module-level types.
+    if (dynamic_pointer_cast<Module>(p->container()))
+    {
+        _out << nl << "/// <remarks>" << "The Slice compiler generated this " << generatedType << " from Slice "
+             << p->kindOf() << " <c>" << p->scoped() << "</c>.";
```
InsertCreativityHere, line 763: This might push it over the 120 limit... but it sounds weird without it.
```suggestion
        _out << nl << "/// <remarks>" << "The Slice compiler generated this " << generatedType << " from the Slice "
             << p->kindOf() << " <c>" << p->scoped() << "</c>.";
```

cpp/src/slice2cs/Gen.cpp
```diff
@@ -655,23 +722,52 @@ Slice::CsVisitor::writeDataMemberInitializers(const DataMemberList& dataMembers)
...
+Slice::CsVisitor::writeHelperDocComment(
+    const ContainedPtr& p,
+    const string& comment,
+    const string& generatedType,
+    const string& notes)
+{
+    writeDocLine(_out, "summary", comment);
+
+    // We generate remarks only for module-level types.
+    if (dynamic_pointer_cast<Module>(p->container()))
```
InsertCreativityHere, line 760: The only types we generate helpers for are sequences, dictionaries, enums, and proxies (Slice interfaces) right?
So, I think this should be an `assert`, not an if, since we should never call this on a non-module-level-type.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -655,23 +722,52 @@ Slice::CsVisitor::writeDataMemberInitializers(const DataMemberList& dataMembers)
...
 
-    StringList overview = comment->overview();
-    if (!overview.empty())
+    // We generate remarks only for module-level types.
+    if (dynamic_pointer_cast<Module>(p->container()))
     {
-        _out << nl << "/// <summary>";
-        writeDocLines(_out, overview);
-        _out << nl << "/// </summary>";
+        _out << nl << "/// <remarks>" << "The Slice compiler generated this " << generatedType << " from Slice "
```
InsertCreativityHere, line 735: Same:
```suggestion
        _out << nl << "/// <remarks>" << "The Slice compiler generated this " << generatedType << " from the Slice "
```

cpp/src/slice2cs/Gen.cpp
```diff
@@ -655,23 +722,52 @@ Slice::CsVisitor::writeDataMemberInitializers(const DataMemberList& dataMembers)
...
     {
-        return;
+        writeDocLines(_out, "summary", comment->overview());
     }
 
-    StringList overview = comment->overview();
-    if (!overview.empty())
+    // We generate remarks only for module-level types.
+    if (dynamic_pointer_cast<Module>(p->container()))
     {
```
InsertCreativityHere, line 734: Might be nice to add an `assert(!generatedType.empty());` here, since if we have a module-level type, it _has_ to provide this field to avoid generated a poorly formatted `remarks` section.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -26,18 +26,83 @@ using namespace IceInternal;
...
+                        out << ' ';
+                    }
+                }
+                out << line;
+            }
+            out << "</" << *closeTag << ">";
         }
     }
 
+    // Standard marshal doc-comment
```
InsertCreativityHere, line 89: Why not make this a doc-comment, so we could see it on hover?
```suggestion
    /// Standard marshal doc-comment
```

cpp/src/slice2cs/Gen.cpp
```diff
@@ -26,18 +26,83 @@ using namespace IceInternal;
...
+    void writeMarshalDocComment(Output& out)
+    {
+        writeDocLine(out, "summary", "Marshals a value into an output stream.");
+        writeDocLine(out, R"(param name="ostr")", "The output stream.", "param");
+        writeDocLine(out, R"(param name="v")", "The value to marshal.", "param");
+    }
+
+    // Standard unmarshal doc-comment
```
InsertCreativityHere, line 97: Same:
```suggestion
    /// Standard unmarshal doc-comment
```

InsertCreativityHere (approved):
Looks good.

If you're up for it, the `csLinkFormatter` function in `Gen.cpp` could use some love as well.
This is what is used to generate the `see cref`s for `@link`s.

cpp/src/slice2cs/Gen.cpp
bernardnormier: > It's not like it's just one of many possible choices the user can implement.

It is actually like that. For Slice interface Greeter, you can use any Disp_ that implements this interface, not just GreeterDisp_.

cpp/src/slice2cs/Gen.cpp
bernardnormier: I find without "the" better.

cpp/src/slice2cs/Gen.cpp
bernardnormier: See above.

cpp/src/slice2cs/Gen.cpp
bernardnormier: Fixed as suggested.

PR #3667, opened 2025-03-11
Remove ice_writeMembers from C# structs

bernardnormier:
This PR removes this generated API. It's not clear to me why we added it.


pepone (approved):
Looks good. But should `ice_write` take a `ref` to avoid the copy?

bernardnormier:
> Looks good. But should `ice_write` take a `ref` to avoid the copy?

Mapped structs often record classes so it's already a reference.

For Slice structs mapped to record structs, we could pass it as `ref readonly`:

See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/method-parameters
and
https://stackoverflow.com/questions/77508998/difference-between-in-and-ref-readonly-parameters

I don't think it's critical for 3.8.

InsertCreativityHere (approved):
Looks good to me!

Issue #3668, opened 2025-03-12
Remove Platform toolset checks from C++ NuGet packages


pepone:
```
<ItemDefinitionGroup Condition="'$(PlatformToolset)' == 'v143'">
        <ClCompile>
            <AdditionalIncludeDirectories>$(MSBuildThisFileDirectory)include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
        </ClCompile>
        <Link>
            <AdditionalLibraryDirectories>$(MSBuildThisFileDirectory)lib\$(Platform)\Release;$(MSBuildThisFileDirectory)lib\$(Platform)\Debug;%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
        </Link>
    </ItemDefinitionGroup>
```

I think we should remove the PlatformToolset checks, the new VC++ runtime should be compatible with older toolsets.

PR #3672, opened 2025-03-12
Remove platform toolset checks in C++ NuGet package

pepone:
See #3668

### DIFF
--- a/cpp/msbuild/ZeroC.Ice.Cpp.targets
+++ b/cpp/msbuild/ZeroC.Ice.Cpp.targets
@@ -3,7 +3,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
     <!-- Use bin and lib directories that match the used platform toolset -->
-    <PropertyGroup Condition="'$(PlatformToolset)' == 'v143'">
+    <PropertyGroup>
         <IceConfiguration Condition="'$(UseDebugLibraries)' == 'true'">Debug</IceConfiguration>
         <IceConfiguration Condition="'$(UseDebugLibraries)' != 'true'">Release</IceConfiguration>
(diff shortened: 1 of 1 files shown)


externl (approved):

cpp/msbuild/ZeroC.Ice.Cpp.targets
```diff
@@ -14,13 +14,13 @@
     <Import Project="$(MSBuildThisFileDirectory)..\ZeroC.Ice.Slice.Tools.Cpp.targets" />
 
     <ItemGroup>
-        <_Ice_DLLs_v143 Include="$(MSBuildThisFileDirectory)bin\$(Platform)\$(IceConfiguration)\*$(IceDLLSuffix).dll" Exclude="$(MSBuildThisFileDirectory)bin\$(Platform)\$(IceConfiguration)\icedb37*.dll"/>
-        <_Ice_DLLs_v143 Include="$(MSBuildThisFileDirectory)bin\$(Platform)\$(IceConfiguration)\libssl*.dll"/>
-        <_Ice_DLLs_v143 Include="$(MSBuildThisFileDirectory)bin\$(Platform)\$(IceConfiguration)\libcrypto*.dll"/>
-        <_Ice_DLLs_v143 Include="$(MSBuildThisFileDirectory)bin\$(Platform)\$(IceConfiguration)\bzip2*.dll"/>
+        <_Ice_DLLs Include="$(MSBuildThisFileDirectory)bin\$(Platform)\$(IceConfiguration)\*$(IceDLLSuffix).dll" Exclude="$(MSBuildThisFileDirectory)bin\$(Platform)\$(IceConfiguration)\icedb38*.dll"/>
```
bernardnormier, line 17: Would be nice to add a comment describing what this ItemGroup is about.

Then, it's odd that we include in this item group essentially everything (like IceStormService?) but exclude icedb38*.dll.

bernardnormier (approved):

cpp/msbuild/ZeroC.Ice.Cpp.targets
pepone: I will review in a separate PR.

PR #3673, opened 2025-03-12
Move Slice files to their respective C# NuGet packages

### LINKED ISSUE #3665: Move Slice files to their respective C# NuGet packages
Currently, the Slice files for Ice definitions are packaged within `ZeroC.Ice.Slice.Tools`. However, it would be more logical and maintainable to distribute them within their respective packages.

Each package should include a build-transitive properties file that exposes its Slice directory, allowing dependent packages to locate and use the necessary definitions automatically.

pepone:
Fix #3665

### DIFF
--- a/csharp/src/Glacier2/ZeroC.Glacier2.props
+++ b/csharp/src/Glacier2/ZeroC.Glacier2.props
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright (c) ZeroC, Inc. -->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <ItemDefinitionGroup>
+        <SliceCompile>
+            <IncludeDirectories>%(IncludeDirectories);$(MSBuildThisFileDirectory)..\slice</IncludeDirectories>
+        </SliceCompile>
+    </ItemDefinitionGroup>
+</Project>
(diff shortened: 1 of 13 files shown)


externl (approved):

csharp/src/Glacier2/ZeroC.Glacier2.props
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright (c) ZeroC, Inc. -->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <ItemDefinitionGroup>
+        <SliceCompile>
+            <IncludeDirectories>%(IncludeDirectories);$(MSBuildThisFileDirectory)..\slice</IncludeDirectories>
```
bernardnormier, line 6: Should this be a forward `/`? These props are cross-platform.

bernardnormier (approved):

csharp/src/Glacier2/ZeroC.Glacier2.props
pepone: I think we should actually fix this to normalize the path we pass to the task. We can do it from the .targets files when we call the task.

pepone: I updated the tools package to normalize paths with `[MSBuild]::NormalizePath`

PR #3676, opened 2025-03-12
Split setup-dependencies

### LINKED ISSUE #3670: Improve setup dependencies to do fine grained installs
We are using this action to install build dependencies and test dependencies in  a number of workflows:

```
$ rg setup-dependencies
workflows\matlab.yml
19:        uses: ./.github/actions/setup-dependencies

workflows\ice2slice.yml
19:        uses: ./.github/actions/setup-dependencies

workflows\cpp.yml
40:        uses: ./.github/actions/setup-dependencies

workflows\build-xcframework-packages.yml
42:        uses: ./.github/actions/setup-dependencies

workflows\ci.yml
105:        uses: ./.github/actions/setup-dependencies

workflows\coverage.yml
16:        uses: ./.github/actions/setup-dependencies
61:        uses: ./.github/actions/setup-dependencies
101:        uses: ./.github/actions/setup-dependencies

workflows\build-swift-package.yml
28:        uses: ./.github/actions/setup-dependencies

workflows\build-pip-packages.yml
48:        uses: ./.github/actions/setup-dependencies

workflows\build-npm-packages.yml
50:        uses: ./.github/actions/setup-dependencies

workflows\build-msi-package.yml
50:        uses: ./ice/.github/actions/setup-dependencies

workflows\build-maven-packages.yml
57:        uses: ./.github/actions/setup-dependencies

workflows\build-gem-packages.yml
41:        uses: ./.github/actions/setup-dependencies

workflows\build-matlab-packages.yml
22:        uses: ./.github/actions/setup-dependencies

workflows\build-dotnet-packages.yml
50:        uses: ./.github/actions/setup-dependencies

workflows\build-brew-packages.yml
41:        uses: ./.github/actions/setup-dependencies

workflows\build-cpp-nuget-package.yml
33:        uses: ./.github/actions/setup-dependencies
```

We install too many things by default:

https://github.com/zeroc-ice/ice/blob/d84b0c15a3b3a32a9274fdac98e3e069bff72f9a/.github/actions/setup-dependencies/action.yml#L19-L45

We don't need NPM dotnet to build brew packages. We don't need passlib for the build-xxx runners and so on. The dotnet builds don't need Java.

We can either split this in several setup-xxx-dependencies, or stop using it everywhere.

See also https://github.com/zeroc-ice/ice/actions/runs/13809958094/job/38629107994

XCFramework build fails because it cannot install dotnet.

In the end all this extra install add extra time to build things, and create unnecessary dependencies.

externl:
Fixes #3670 

I've split setup-dependencies into one setup-xxx per language.


.github/actions/setup-dotnet/action.yml
```diff
@@ -0,0 +1,8 @@
+name: Setup .NET
```
pepone, line 1: Did you consider using setup-dotnet directly?

pepone (approved):
Looks good.

.github/actions/setup-dotnet/action.yml
externl: Yes, but since we want to be using the same version in all of our actions I found this be a good way to do that.

pepone: Yes. Makes sense to keep it.

PR #3678, opened 2025-03-12
Restrict cs:attribute to fewer constructs

### LINKED ISSUE #3662: Update cs:attribute to apply only to enum, enumerators and fields
For every other constructs, you should use partial.

See also https://docs.icerpc.dev/slice2/language-guide/attributes#c#-attributes

bernardnormier:
This PR restricts cs:attribute to enums, enumerators, fields and constants.

Fixes #3662.

### DIFF
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -258,6 +258,8 @@ other CPUs.
 invocation made from an Ice thread pool thread executes in a .NET thread pool thread; previously, this continuation
 was executed in a thread managed by the same Ice thread pool unless you specified `.ConfigureAwait(false)`.
 
+- The `cs:attribute` Slice metadata is now limited to enum, enumerators, fields and constants.
+
 ## Objective-C Changes
 
 - The Objective-C mapping was removed.
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -258,7 +258,7 @@ other CPUs.
 invocation made from an Ice thread pool thread executes in a .NET thread pool thread; previously, this continuation
 was executed in a thread managed by the same Ice thread pool unless you specified `.ConfigureAwait(false)`.
 
-- The `cs:attribute` Slice metadata is now limited to enum, enumerators, fields, and constants.
+- The `cs:attribute` Slice metadata is now limited to enums, enumerators, fields, and constants.
 
 ## Objective-C Changes
 
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -258,7 +258,7 @@ other CPUs.
 invocation made from an Ice thread pool thread executes in a .NET thread pool thread; previously, this continuation
 was executed in a thread managed by the same Ice thread pool unless you specified `.ConfigureAwait(false)`.
 
-- The `cs:attribute` Slice metadata is now limited to enum, enumerators, fields and constants.
+- The `cs:attribute` Slice metadata is now limited to enum, enumerators, fields, and constants.
 
 ## Objective-C Changes
 
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -640,10 +640,9 @@ Slice::CsVisitor::getDispatchParams(
 void
 Slice::CsVisitor::emitAttributes(const ContainedPtr& p)
 {
-    assert(dynamic_pointer_cast<Enum>(p->container()) ||
-        dynamic_pointer_cast<Enumerator>(p->container()) ||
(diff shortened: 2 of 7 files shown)


CHANGELOG-3.8.md
```diff
@@ -258,6 +258,8 @@ other CPUs.
 invocation made from an Ice thread pool thread executes in a .NET thread pool thread; previously, this continuation
 was executed in a thread managed by the same Ice thread pool unless you specified `.ConfigureAwait(false)`.
 
+- The `cs:attribute` Slice metadata is now limited to enum, enumerators, fields and constants.
```
InsertCreativityHere, line 261: Ever the pedant:
```suggestion
- The `cs:attribute` Slice metadata is now limited to enum, enumerators, fields, and constants.
```

cpp/src/slice2cs/Gen.cpp
```diff
@@ -1639,10 +1602,9 @@ Slice::Gen::TypesVisitor::visitDataMember(const DataMemberPtr& p)
 
     _out << sp;
 
-    emitObsoleteAttribute(p, _out);
-
     string type = typeToString(p->type(), ns, p->optional());
 
+    emitObsoleteAttribute(p, _out);
```
bernardnormier, line 1607: That's new, and we "test it" with the deprecated Nonmutating.

csharp/src/Ice/Object.cs
```diff
@@ -125,6 +125,22 @@ public abstract class ObjectImpl : Object
     /// <returns>The return value is always ::Ice::Object.</returns>
     public static string ice_staticId() => "::Ice::Object";
 
+    public static void iceCheckMode(OperationMode expected, OperationMode received)
+    {
+        Debug.Assert(expected is OperationMode.Normal or OperationMode.Idempotent); // We never expect Nonmutating
+        if (expected != received)
+        {
+            if (expected is OperationMode.Idempotent && received is not OperationMode.Normal)
```
bernardnormier, line 133: Avoid using the deprecated Nonmutating enumerator. Semantics doesn't change.

bernardnormier: We should also review this check (in all languages) and sync-up with the IceRPC logic in a follow-up PR.

received = Normal should never trigger an error.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -654,6 +640,10 @@ Slice::CsVisitor::getDispatchParams(
 void
 Slice::CsVisitor::emitAttributes(const ContainedPtr& p)
 {
+    assert(
+        dynamic_pointer_cast<Enum>(p->container()) || dynamic_pointer_cast<Enumerator>(p->container()) ||
+        dynamic_pointer_cast<DataMember>(p->container()) || dynamic_pointer_cast<Const>(p->container()));
```
InsertCreativityHere, line 645: This seems bogus, we shouldn't be calling `container` on these. We want to be checking the actual `contained` themselves.

csharp/src/Ice/Object.cs
bernardnormier: See #3679

cpp/src/slice2cs/Gen.cpp
bernardnormier: Fixed

CHANGELOG-3.8.md
```diff
@@ -258,6 +258,8 @@ other CPUs.
 invocation made from an Ice thread pool thread executes in a .NET thread pool thread; previously, this continuation
 was executed in a thread managed by the same Ice thread pool unless you specified `.ConfigureAwait(false)`.
 
+- The `cs:attribute` Slice metadata is now limited to enum, enumerators, fields, and constants.
```
InsertCreativityHere, line 261: It's my bad for not seeing this in my initial review... but we should also be consistent with plurality here.
```suggestion
- The `cs:attribute` Slice metadata is now limited to enums, enumerators, fields, and constants.
```

cpp/src/slice2cs/Gen.h
```diff
@@ -39,10 +39,10 @@ namespace Slice
             std::vector<std::string>&,
             const std::string&);
 
+        // Generates a C# attribute from the cs:attribute metadata.
```
InsertCreativityHere, line 42: Feels like a doc-comment; also it handles _all_ the attributes, it's not something you have to call one at a time.
```suggestion
        /// Generates C# attributes from any 'cs:attribute' metadata.
```

InsertCreativityHere (approved):
Looks good to me!

pepone (approved):

PR #3682, opened 2025-03-12
Java comment standardization

### LINKED ISSUE #2758: Bad java formatting
The Java formatting in some files is very very strange.

For example:
https://github.com/zeroc-ice/ice/blob/290d5292b08d83614f907dea9ad324d1687499d7/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Coordinator.java#L1983

ayoitslilith:
Squishes up all the comments and then expands them back out again using Spotless's formatting rules to do away with the strangely formatted comment blocks. Line breaks are now standardized to occur if a word would go over Spotless's character limit.

Fixes #2758.

### DIFF
--- a/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/LiveDeployment/GraphView.java
+++ b/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/LiveDeployment/GraphView.java
@@ -206,8 +206,7 @@ public class GraphView extends JFrame implements MetricsFieldContext, Coordinato
                                         _maxRefreshPeriod,
                                         1);
 
-                        // SpinnerNumberModel to set the maximum number of samples to keep in X
-                        // axis.
+                        // SpinnerNumberModel to set the maximum number of samples to keep in X axis.
                         final SpinnerNumberModel samples =
                                 new SpinnerNumberModel(_samples, _minSamples, _maxSamples, 1);
 
@@ -339,9 +338,7 @@ public class GraphView extends JFrame implements MetricsFieldContext, Coordinato
                                             if (seriesClass != null) {
                                                 _styles.remove(seriesClass);
                                             }
-                                            // Don't remove the XYChart.Series object here, to avoid
-                                            // the series style classes getting reassigned by
-                                            // JavaFX.
+                                            // Don't remove the XYChart.Series object here, to avoid the series style classes getting reassigned by JavaFX.
                                             //
                                             // TODO: _chart.getData().remove(row.series);
                                             try {
@@ -642,8 +639,7 @@ public class GraphView extends JFrame implements MetricsFieldContext, Coordinato
 
                                 columns.put(j.getField().getFieldName(), row);
                                 j.getField().setContext(GraphView.this);
-                                // When a line is clicked we select the corresponding row in the
-                                // legend table.
+                                // When a line is clicked we select the corresponding row in the legend table.
                                 javafx.scene.Node n =
                                         _chart.lookup(".chart-series-line." + styleClass);
                                 if (n != null) {
@@ -703,8 +699,7 @@ public class GraphView extends JFrame implements MetricsFieldContext, Coordinato
         // Reset the cell field necessary to avoid bogus calculations.
         row.cell.resetField();
 
-        // We need also a new click handler so click works in all segments
-        // of the line.
+        // We need also a new click handler so click works in all segments of the line.
         //
         // When a line is clicked we select the corresponding row in the legend table.
         javafx.scene.Node n = _chart.lookup(".chart-series-line." + styleClass);
@@ -749,9 +744,7 @@ public class GraphView extends JFrame implements MetricsFieldContext, Coordinato
                             metricsSeq = data.get(i.getKey());
                         }
 
-                        // Iterate over all configured values, if there isn't data for one
-                        // configured
-                        // field we need to add a gap.
+                        // Iterate over all configured values, if there isn't data for one configured field we need to add a gap.
                         for (Map.Entry<String, Map<String, MetricsRow>> j :
                                 i.getValue().entrySet()) {
                             com.zeroc.Ice.IceMX.Metrics metrics = null;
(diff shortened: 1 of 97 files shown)


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
```diff
@@ -544,16 +543,12 @@ public Class<?> apply(String typeId) {
...
-        // 1. Convert the Slice type ID into a classname (e.g., ::M::X becomes M.X).
-        // 2. Check if the application has defined any package prefixes for the top-level module
-        // (e.g.:
-        // "M").
-        //    Attempt to resolve the <package-prefix>+<class-name> in order trying each defined
-        // package
-        //    prefix.
-        // 3. If the above step fails, it checks the value of Default.Package property. If found,
-        //    prepend its value to the classname. Otherwise, attempt to use the
-        //    classname directly.
```
InsertCreativityHere, line 556: Or sometimes we have explicit lists of things, where again, the numbers really shouldn't be scattered inside the body of a monolithic paragraph.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectAdapter.java
```diff
@@ -213,8 +209,8 @@ public void waitForHold() {
      */
     public void deactivate() {
         synchronized (this) {
-            // Wait for activation or a previous deactivation to complete.
-            // This is necessary to avoid out of order locator updates.
```
InsertCreativityHere, line 217: And even though I expect this isn't a shared opinion, IMO, I think trying to maximize the length of lines in every possible case is counter-productive. I personally think comments are easier to read when the end-of-a-line corresponds to the end of a sentence or a natural break in the sentence.

I'm not saying we should fix the existing comments to do this, but as that one weirdo who goes out of his way to format comments for maximal readability, this sometimes means intentionally ending a line well-before I hit the column limit. And I'd prefer not for this intentional choice to be erased, just to maximize how many columns we can fit things into.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/RFC2253.java
```diff
@@ -247,9 +239,8 @@ else if (state.data.charAt(state.pos) != '\\') {
             }
         }
         //
-        // RFC 2253
-        // * (Stringchar | pair)
-        // Stringchar = <any character except one of special, "\" or QUOTATION >
```
InsertCreativityHere, line 252: Just one final example, in this file, it looks like we're laying out BNF grammar rules, and again, the choice of
"1 rule per line" seems both intentional and very helpful to readability.

InsertCreativityHere:
Also, just for the record, there is an open issue to try and find an even better alternative to the formatter we're using right now:
#2758

This might be beyond the scope of what you're trying to focus on, but if you could find some better alternative to what we're doing now, I would support an attempt to switch to it.

bernardnormier (approved):
Looks good.

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/LiveDeployment/GraphView.java
```diff
@@ -1157,8 +1151,8 @@ public void setValueAt(final Object value, int rowIndex, int columnIndex) {
                     row.cell.setScaleFactor(((Double) value).doubleValue());
                 } else if (_columnNames[columnIndex].equals("Color")) {
                     Color color = (Color) value;
-                    // Convert color to the CSS representation used by JavaFX style.
-                    // example: #ff00aa
+                    // Convert color to the CSS representation used by JavaFX style. example:
+                    // #ff00aa
```
externl, line 1155:
```suggestion
                    // Convert color to the CSS representation used by JavaFX style.
                    // example: #ff00aa
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
```diff
@@ -823,17 +822,16 @@ static Property findProperty(String key, PropertyArray propertyArray) {
             // If the property has a property class, check if the key is a prefix of the property.
             if (prop.propertyArray() != null) {
                 // Check if the key is a prefix of the property.
-                // The key must be:
-                // - shorter than the property pattern
-                // - the property pattern must start with the key
+                //  The key must be:
```
externl, line 825: Extra whitespace was added here.

java/src/com.zeroc.icebox/src/main/java/com/zeroc/IceBox/ServiceManagerI.java
```diff
@@ -719,8 +713,7 @@ static class StartServiceInfo {
 
             // We support the following formats:
             //
-            // <class-name> [args]
-            // <jar-file>:<class-name> [args]
+            // <class-name> [args] <jar-file>:<class-name> [args]
```
externl, line 716: previous formatting was correct

java/test/src/main/java/test/Ice/exceptions/AMDThrowerI.java
```diff
@@ -144,7 +144,7 @@ public CompletionStage<Void> throwUndeclaredCAsync(
         ex.cMem = c;
         throw ex;
         // r.completeExceptionally(ex);
-        // return r;
+        //  return r;
```
externl, line 147: Extra space

java/test/src/main/java/test/Ice/executor/AllTests.java
```diff
@@ -208,8 +208,8 @@ public static void allTests(test.TestHelper helper, final CustomExecutor executo
                 cb.check();
             }
 
-            // Hold adapter to make sure the invocations don't complete synchronously
-            // Also disable collocation optimization on p
+            // Hold adapter to make sure the invocations don't complete synchronously Also disable
+            // collocation optimization on p
```
externl, line 212:
```suggestion
            // Hold adapter to make sure the invocations don't complete synchronously. Also disable
            // collocation optimization on p.
```

externl (approved):
Looks good. A few fixes to make.

PR #3683, opened 2025-03-12
Add Support for `matlab:identifier` Metadata

InsertCreativityHere:
This PR adds the very first MATLAB metadata: `matlab:identifier`. You already know what this does...

The only difference is we don't allow it on classes or exceptions, because the `ClassResolver` needs to be able to map from a type-ID to a MATLAB type... Note that we have the same issue and limitation with `java:identifier`.

### DIFF
--- a/cpp/src/slice2matlab/Main.cpp
+++ b/cpp/src/slice2matlab/Main.cpp
@@ -1032,7 +1032,7 @@ namespace
         knownMetadata.emplace("matlab:identifier", std::move(identifierInfo));
 
         // Pass this information off to the parser's metadata validation logic.
-        Slice::validateMetadata(unit, "matlab", knownMetadata);
+        Slice::validateMetadata(unit, "matlab", std::move(knownMetadata));
     }
 }
 
--- a/cpp/src/slice2matlab/Main.cpp
+++ b/cpp/src/slice2matlab/Main.cpp
@@ -34,19 +34,6 @@ using namespace IceInternal;
 
 namespace
 {
-    string replace(const string& s, const string& patt, const string& val)
-    {
-        string r{s};
-        string::size_type pos = r.find(patt);
-        while (pos != string::npos)
-        {
-            r.replace(pos, patt.size(), val);
-            pos += val.size();
-            pos = r.find(patt, pos);
-        }
-        return r;
-    }
-
     void writeCopyright(IceInternal::Output& out, const string& file)
     {
         string f = file;
--- a/cpp/src/slice2matlab/Main.cpp
+++ b/cpp/src/slice2matlab/Main.cpp
@@ -1023,6 +1023,31 @@ namespace
             out << nl << "% Deprecated";
         }
     }
+
+    void validateMetadata(const UnitPtr& unit)
+    {
+        map<string, MetadataInfo> knownMetadata;
+
+        // "matlab:identifier"
+        MetadataInfo identifierInfo = {
+            .validOn =
+                {typeid(InterfaceDecl),
+                typeid(Operation),
+                typeid(Struct),
+                typeid(Sequence),
+                typeid(Dictionary),
+                typeid(Enum),
+                typeid(Enumerator),
+                typeid(Const),
+                typeid(Parameter),
+                typeid(DataMember)},
+            .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
+        };
+        knownMetadata.emplace("matlab:identifier", std::move(identifierInfo));
+
+        // Pass this information off to the parser's metadata validation logic.
+        Slice::validateMetadata(unit, "matlab", knownMetadata);
+    }
 }
 
 //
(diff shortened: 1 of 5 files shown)


slice/IceGrid/Descriptor.ice
```diff
@@ -42,6 +42,7 @@ module IceGrid
         Ice::StringSeq references;
 
         /// The property set properties.
+        ["matlab:identifier:propertyDescriptors"]
         PropertyDescriptorSeq properties;
```
InsertCreativityHere, line 46: `properties` is a reserved identifier in MATLAB.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -314,7 +50,20 @@ namespace
 
     void openClass(const string& abs, const string& dir, IceInternal::Output& out)
     {
```
InsertCreativityHere, line 52: This was the only place left where we called `splitAbsoluteName`, so I just inlined the logic.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -358,30 +107,13 @@ namespace
...
         //
-        const string cls = lookupKwd(v[v.size() - 1]);
+        const string cls = v[v.size() - 1];
         path += cls + ".m";
 
         out.open(path);
         FileTracker::instance()->addFile(path);
     }
 
-    //
```
InsertCreativityHere, line 368: This function was replaced by `cont->getMappedScoped(".").substr(1)`.
And just to re-iterate, I plan to cleanup all these leading/trailing separators after I finish the `xxx:identifiers`.

InsertCreativityHere:
My generated-code-comparison-script confirms that everything should be good here!

I did notice that there is bad whitespace in generated MATLAB enums. But I left it alone when I saw the code responsible:
```
        out << nl; // TODO: rework this code
        if (!isFirstElement(enumerator))
        {
            out << nl; // TODO: should be sp, but sp is broken and resets indentation
        }
```

matlab/test/Slice/escape/Test.ice
```diff
@@ -2,86 +2,74 @@
...
-    interface elseif // elseifPrx should not be escaped
+    ["matlab:identifier:MyInterface"]
+    interface operations
     {
-        void events();
-        void function();
-        void delete();
-        void checkedCast();
+        ["matlab:identifier:foobar"] void events();
+        ["matlab:identifier:func"] void function();
```
InsertCreativityHere, line 70: I should probably add tests for `matlab:identifier` on parameters.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -314,7 +50,20 @@ namespace
 
     void openClass(const string& abs, const string& dir, IceInternal::Output& out)
     {
-        vector<string> v = splitAbsoluteName(abs);
+        // Split the absolute name into individual components.
```
pepone, line 53: We have a quite similar method in slice2java, maybe we can refactor this code and add helper in Slice/Util.h, anyway that would be for another PR.

pepone (approved):

slice/IceGrid/Descriptor.ice
bernardnormier: It's not a MATLAB keyword. In particular:
```

>> matlab.lang.makeValidName('properties')

ans =

    'properties'
```

matlab/test/Slice/escape/Test.ice
```diff
@@ -2,86 +2,74 @@
 
 #pragma once
 
-module classdef::break // Should be escaped
+// TODO come up with a better way to escape modules.
+module escaped_classdef
 {
-    enum bitand // Should not be escaped
+    ["matlab:identifier:persistent_"]
```
bernardnormier, line 8: The default escaping in MATLAB is `x<Capital>`, as in:
```
>> matlab.lang.makeValidName('persistent')

ans =

    'xPersistent'
```

bernardnormier (approved):

slice/IceGrid/Descriptor.ice
InsertCreativityHere: I guess there must be some level of context-awareness here?
The docs say that it is a `"MATLAB class-definition keyword"`:
https://www.mathworks.com/help/matlab/ref/properties.html#d126e1399987

Even if it's theoretically fine, I'd still rather play it safe and remap it to something else here.
Since semantics aside, seeing `properties` in MATLAB has an understood meaning
that is different than what this sequence represents.

matlab/test/Slice/escape/Test.ice
InsertCreativityHere: I will update this test accordingly.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -1286,6 +1009,31 @@ namespace
...
+                 typeid(Enumerator),
+                 typeid(Const),
+                 typeid(Parameter),
+                 typeid(DataMember)},
+            .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
+        };
+        knownMetadata.emplace("matlab:identifier", std::move(identifierInfo));
+
+        // Pass this information off to the parser's metadata validation logic.
+        Slice::validateMetadata(unit, "matlab", knownMetadata);
```
bernardnormier, line 1035: Should be `std::move(knownMetadata)`.

PR #3685, opened 2025-03-12
Fix Broken Links in C++ and XML Stripping Order

InsertCreativityHere:
The code for parsing doc-comments has a bool to XML tags out of the comment's text.
The issue is we were doing this _after_ we fixed all the `@link`s, which in C# are mapped to XML tags.

Now the order has been swapped. We do any stripping at the very beginning of parsing, then afterwards massage the `@link`s.

----

It also fixes a few bugs in how C++ generated links. I reviewed the generated Doxygen, and 
1) Enums and Enumerators were just broken. They require special handling with Doxygen.
image
Before these types were black, indicating they weren't proper links.

2) When `@link`ing to an interface/operation type, we should always point to the proxy version, not the generated servant types.
image
Before we linked to the non-Prx type.

----

All of these issues were caught by @bernardnormier while improving the C# link generation.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -448,45 +448,46 @@ namespace
 optional<DocComment>
 Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatter, bool stripMarkup, bool escapeXml)
 {
-    string rawComment = p->docComment();
-
-    // Fix any link tags using the provided link formatter.
-    const string link = "{@link ";
-    auto pos = rawComment.find(link);
-    while (pos != string::npos)
-    {
-        auto endpos = rawComment.find('}', pos);
-        if (endpos != string::npos)
-        {
-            // Extract the linked-to identifier.
-            auto identStart = rawComment.find_first_not_of(" \t", pos + link.size());
-            auto identEnd = rawComment.find_last_not_of(" \t", endpos);
-            string linkText = rawComment.substr(identStart, identEnd - identStart);
-
-            // Then erase the entire '{@link foo}' tag from the comment.
-            rawComment.erase(pos, endpos - pos + 1);
-
-            // Attempt to resolve the link, and issue a warning if the link is invalid.
-            SyntaxTreeBasePtr linkTarget = resolveDocLink(linkText, p);
-            if (!linkTarget)
-            {
-                string msg = "no Slice element with identifier '" + linkText + "' could be found in this context";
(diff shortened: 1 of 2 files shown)


cpp/src/Slice/Parser.cpp
```diff
@@ -448,45 +448,46 @@ namespace
 optional<DocComment>
 Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatter, bool stripMarkup, bool escapeXml)
```
InsertCreativityHere, line 449: All I did was swap the order, and move the link-fixing into a `for` loop.
Since now it works over a `vector<string> lines` instead of a `string rawComment`.

Since `@link` cannot span multiple lines, this is fine.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -256,17 +256,34 @@ namespace
     /// Returns a doxygen formatted link to the provided Slice identifier.
     string cppLinkFormatter(const string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target)
     {
-        string result = "{@link ";
         if (target)
         {
-            if (auto dataMemberTarget = dynamic_pointer_cast<DataMember>(target))
+            if (dynamic_pointer_cast<DataMember>(target) || dynamic_pointer_cast<Enumerator>(target))
```
InsertCreativityHere, line 261: We should of been treating enumerators like we did fields.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -256,17 +256,34 @@ namespace
...
+            {
+                // If a link to an enum isn't qualified (ie. we're the source and target are in the same module),
+                // we have to place a '#' character in front, to explicitly tell Doxygen it's in the current scope.
+                string link = getUnqualified(enumTarget->mappedScoped(), source->mappedScope());
+                if (link.find("::") == string::npos)
+                {
+                    link.insert(0, "#");
+                }
+                return link;
+            }
```
InsertCreativityHere, line 282: Not sure why we need to do this... But, without this none of the Doxygen links to unqualified enums were working...
And alternative would be to force them to qualified, but I think this is cleaner.

InsertCreativityHere:
I've double checked these changes with my comparison script, and there are no unexpected changes introduced by this.
Only lots of links now ending with `Prx`, and one link to an enum with `#` in front now. As desired.

bernardnormier (approved):

externl (approved):
Looks good to me!

PR #3686, opened 2025-03-12
More doc-comment updates in slice2cs

bernardnormier:
This PR fixes more doc-comments generation issues in slice2cs.

### DIFF
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -640,10 +640,9 @@ Slice::CsVisitor::getDispatchParams(
 void
 Slice::CsVisitor::emitAttributes(const ContainedPtr& p)
 {
-    assert(dynamic_pointer_cast<Enum>(p->container()) ||
-        dynamic_pointer_cast<Enumerator>(p->container()) ||
-        dynamic_pointer_cast<DataMember>(p->container()) ||
-        dynamic_pointer_cast<Const>(p->container()));
+    assert(
+        dynamic_pointer_cast<Enum>(p->container()) || dynamic_pointer_cast<Enumerator>(p->container()) ||
+        dynamic_pointer_cast<DataMember>(p->container()) || dynamic_pointer_cast<Const>(p->container()));
 
     for (const auto& metadata : p->getMetadata())
     {
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -26,16 +26,80 @@ using namespace IceInternal;
 
 namespace
 {
-    void writeDocLines(Output& out, const StringList& lines)
+    void
+    writeDocLine(Output& out, const string& openTag, const string& comment, std::optional<string> closeTag = nullopt)
     {
-        for (const auto& line : lines)
+        if (comment.empty())
         {
-            out << nl << "///";
-            if (!line.empty())
-            {
-                out << " " << line;
-            }
+            return;
         }
+
+        if (!closeTag)
+        {
+            closeTag = openTag;
+        }
+
+        out << nl << "/// <" << openTag << ">" << comment << "</" << *closeTag << ">";
+    }
+
+    void writeDocLines(Output& out, const string& openTag, const StringList& lines, optional<string> closeTag = nullopt)
+    {
+        // If there is a single line, write the doc-comment as a single line. Otherwise, write the doc-comment on
+        // multiple lines.
+
+        if (lines.size() == 1)
+        {
+            writeDocLine(out, openTag, lines.front(), closeTag);
+        }
+        else
+        {
+            if (lines.empty())
+            {
+                return;
+            }
+
+            if (!closeTag)
+            {
+                closeTag = openTag;
+            }
+
+            bool firstLine = true;
+
+            for (const auto& line : lines)
+            {
+                if (firstLine)
+                {
+                    firstLine = false;
+                    out << nl << "/// <" << openTag << ">";
+                }
+                else
+                {
+                    out << nl << "///";
+                    if (!line.empty())
+                    {
+                        out << ' ';
+                    }
+                }
+                out << line;
+            }
+            out << "</" << *closeTag << ">";
+        }
+    }
+
+    /// Standard marshal doc-comment
+    void writeMarshalDocComment(Output& out)
+    {
+        writeDocLine(out, "summary", "Marshals a value into an output stream.");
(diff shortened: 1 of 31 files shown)


slice/Ice/Locator.ice
```diff
@@ -48,7 +48,7 @@ module Ice
     interface LocatorRegistry;
 
     /// The Ice locator interface. This interface is used by clients to lookup adapters and objects. It is also used by
-    /// servers to get the locator registry proxy. <p class="Note">The {@link Locator} interface is intended to be used
```
bernardnormier, line 51: The javadoc paragraph tags are problematic. Better not use them.

slice/Ice/Metrics.ice
```diff
@@ -14,7 +14,7 @@
 
 [["java:package:com.zeroc.Ice"]]
 
-/// The Ice Management eXtension facility. It provides the {@link IceMX#MetricsAdmin} interface for management clients
+/// The Ice Management eXtension facility. It provides the {@link IceMX::MetricsAdmin} interface for management clients
```
bernardnormier, line 17: Minor bug.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -103,29 +103,65 @@ namespace
...
         {
-            if (hashPos != 0)
+            if (auto builtinTarget = dynamic_pointer_cast<Builtin>(target))
             {
-                result += rawLink.substr(0, hashPos);
-                result += ".";
+                string typeS = CsGenerator::typeToString(builtinTarget, "");
+                if (typeS.find("Ice") == std::string::npos)
+                {
+                    result << "langword=\"" << typeS << "\"";
```
bernardnormier, line 118: That's for float, string and similar.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -103,29 +103,65 @@ namespace
...
+            else
+            {
+                // TODO: remove once mappedScope() is fixed.
+                string sourceScope = source->mappedScope(".").substr(1);
+                sourceScope.pop_back();
+
+                if (auto operationTarget = dynamic_pointer_cast<Operation>(target))
+                {
+                    // link to the method on the proxy interface
+                    result << "cref=\"" << CsGenerator::getUnqualified(operationTarget->interface(), sourceScope)
```
bernardnormier, line 136: This can generate fully qualified when partial qualified would work, but I couldn't find a good way to fix it.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -1013,7 +1044,11 @@ Slice::Gen::TypesVisitor::visitClassDefEnd(const ClassDefPtr& p)
                 }
             }
         }
-        _out << parameters << epar;
+
+        // Primary constructor.
+        _out << sp;
+        writeDocLine(_out, "summary", "Initializes a new instance of the " + name + " class.");
```
bernardnormier, line 1050: I didn't add doc-comment for the param constructors ... quite a bit of work, and not critical.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -103,29 +103,65 @@ namespace
...
+                }
             }
-            result += rawLink.substr(hashPos + 1);
         }
         else
         {
-            result += rawLink;
+            // Replace "::"" by "." in the raw link. This is for the situation where the user passes a Slice type
+            // reference but (a) the source Slice file does not include this type and (b) there is no cs:identifier or
+            // other identifier renaming.
```
pepone, line 155: Do we need to deal with `#` of Slice links or that is already done at this point?

cpp/src/slice2cs/Gen.cpp
```diff
@@ -2020,14 +2142,24 @@ Slice::Gen::TypesVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
...
     _out << eb << ";";
 
-    _out << sp << nl << "public static string ice_staticId() => \"" << p->scoped() << "\";";
+    ostringstream staticId;
+    staticId << '"' << p->scoped() << '"';
 
-    _out << sp << nl << "public static void write(Ice.OutputStream ostr, " << name << "Prx? v)";
+    _out << sp;
+    writeDocLine(_out, "summary", "Gets the type ID of the associated Slice interface.");
+    writeDocLine(_out, "returns", staticId.str());
```
pepone, line 2150: Should we put static id in`<c>` tag?

cpp/src/slice2cs/Gen.cpp
bernardnormier: I don't think it is or should be since it's called "rawLink".

cpp/src/slice2cs/Gen.cpp
```diff
@@ -713,7 +749,7 @@ Slice::CsVisitor::writeDataMemberInitializers(const DataMemberList& dataMembers)
 void
 Slice::CsVisitor::writeDocComment(const ContainedPtr& p, const string& generatedType, const string& notes)
 {
-    optional<DocComment> comment = DocComment::parseFrom(p, csLinkFormatter, true, true);
```
InsertCreativityHere, line 716: Now that I merged in my fix for doc-comment tags getting stripped out, you should add these bools back.

We still need them to escape comments like `returns true if a < b`, which would be picked up as invalid XML by docfx.

cpp/src/slice2cs/Gen.cpp
InsertCreativityHere: > Do we need to deal with # of Slice links or that is already done at this point?

The link will still have any `#`s in it. No processing has been done to it by this point. It's all up to this function.

> A raw link might be invalid for the cref value.

I think we should make a best effort attempt to map it.
Which should be as simple as adding `std::replace(s.begin(), s.end(), '#', '.');`

InsertCreativityHere (commented):
Haven't had the chance to fully look this over, but it looks good so far, if you don't want to wait on me!

cpp/src/slice2cs/Gen.cpp
pepone: Do we also emit a warning when target was not found?

bernardnormier: Fixed as suggested.

cpp/src/slice2cs/Gen.cpp
bernardnormier: Fixed as suggested.

PR #3688, opened 2025-03-13
Review tp_new usage in IcePy

### LINKED ISSUE #3677: Review tp_new usage in Ice for Python
This code looks suspicious passing the same arguments to tp_new and tp_init.

https://github.com/zeroc-ice/ice/blob/a266cf8a66ab88a134e8b13e565865e10c497507/python/modules/IcePy/Operation.cpp#L3026-L3035

I think we don't need to pass any args to tp_new, we should review other code calling tp_new and tp_init.

I think here it would be more correct to do:

```
    PyTuple_SET_ITEM(args.get(), 0, createString(operation));
    Py_XINCREF(asyncInvocationContext);
    PyTuple_SET_ITEM(args.get(), 1, asyncInvocationContext);
    PyTypeObject* type = reinterpret_cast<PyTypeObject*>(futureType);
    PyObject* future = type->tp_new(type, nullptr, nullptr);
    if (!future)
    {
        return nullptr;
    }
    type->tp_init(future, args.get(), nullptr); // Call the __init__ method.
```

pepone:
fix #3677

### DIFF
--- a/python/modules/IcePy/Types.cpp
+++ b/python/modules/IcePy/Types.cpp
@@ -1222,10 +1222,12 @@ IcePy::StructInfo::marshal(
     {
         if (!_nullMarshalValue.get())
         {
-            PyObjectHandle args{PyTuple_New(0)};
+            PyObjectHandle emptyArgs{PyTuple_New(0)};
             PyTypeObject* type = reinterpret_cast<PyTypeObject*>(pythonType);
-            _nullMarshalValue = type->tp_new(type, args.get(), 0);
-            type->tp_init(_nullMarshalValue.get(), args.get(), 0); // Initialize the struct members
+            // Create a new instance of the type
+            _nullMarshalValue = type->tp_new(type, emptyArgs.get(), nullptr);
+            // Call the instance __init__ method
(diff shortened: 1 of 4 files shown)


python/modules/IcePy/Types.cpp
```diff
@@ -1222,10 +1222,12 @@ IcePy::StructInfo::marshal(
     {
         if (!_nullMarshalValue.get())
         {
-            PyObjectHandle args{PyTuple_New(0)};
+            PyObjectHandle emptyArgs{PyTuple_New(0)};
             PyTypeObject* type = reinterpret_cast<PyTypeObject*>(pythonType);
-            _nullMarshalValue = type->tp_new(type, args.get(), 0);
-            type->tp_init(_nullMarshalValue.get(), args.get(), 0); // Initialize the struct members
+            // Create a new instance of the type
+            _nullMarshalValue = type->tp_new(type, emptyArgs.get(), nullptr);
```
externl, line 1228: Don't we need a different set of args for `tp_init` here?

pepone: no, we are just passing a reference to an empty list of args.

bernardnormier (approved):

PR #3690, opened 2025-03-13
Document Ice for Python future types

### LINKED ISSUE #3680: Document Python Ice.Future class
This class is part of the public Ice for Python API, yet it lacks a single doc comment.

https://github.com/zeroc-ice/ice/blob/a266cf8a66ab88a134e8b13e565865e10c497507/python/python/Ice/Future.py#L59

pepone:
This PR include doc string comments for the Future types provided by Ice for Python.

We have some documentation for this classes in the Manual:

https://doc.zeroc.com/ice/3.7/language-mappings/python-mapping/client-side-slice-to-python-mapping/asynchronous-method-invocation-ami-in-python/ami-in-python-with-futures#id-.AMIinPythonwithFuturesv3.7-FutureClassesinPython

I also prune a few methods from Ice.InvocationFuture API witch I believe there is no good use case for:

- add_done_callback_async, add_sent_callback_async: If you don't want your callback to run on the current thread when completed deal with it in the callback, doesn't seems like a common use case.
- operation - Seems we bring this from our AsyncResult style mappings

Fix #3680

### DIFF
--- a/python/python/Ice/Future.py
+++ b/python/python/Ice/Future.py
@@ -18,7 +18,36 @@ class FutureBase(object):
 
 
 def wrap_future(future, *, loop=None):
-    """Wrap Ice.Future object into an asyncio.Future."""
+    """
+    Wrap an Ice.Future object into an asyncio.Future.
+
+    This function converts an Ice.Future into an asyncio.Future to allow integration of Ice's
+    asynchronous operations with Python's asyncio framework. If the provided future is already
+    an asyncio.Future, it is returned unchanged.
+
+    If the Ice.Future is already completed, the asyncio.Future is immediately resolved. Otherwise,
+    completion callbacks are registered to ensure that the asyncio.Future reflects the state of the
+    Ice.Future, including result propagation, exception handling, and cancellation.
+
+    Parameters
+    ----------
+    future : FutureBase or asyncio.Future
+        The Ice.Future object to wrap. If an asyncio.Future is passed, it is returned as-is.
+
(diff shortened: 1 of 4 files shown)


python/python/Ice/Future.py
```diff
@@ -57,6 +86,10 @@ def forwardCompletion(sourceFuture, targetFuture):
 
 
 class Future(FutureBase):
+    """
+    A Future object representing the result of an asynchronous operation.
```
bernardnormier, line 90: What is the relationship with https://docs.python.org/3/library/concurrent.futures.html ?

bernardnormier (approved):

python/python/Ice/Future.py
pepone: We basically copied the API, so the code is compatible, then we have some additions in InvocationFuture.

I think we didn't use it directly because concurrent Futures have to be created by an executor. For example you create them using an executor.

It is not clear to me that this was the only possibility, for example couldn't IcePy implement its own executor and return concurrent.futures.Future? I don't think this is something we can rework for 3.8.

externl (approved):

python/python/Ice/Future.py
bernardnormier: I meant, should we document something like:
 - this API is similar / identical / compatible with concurrent Futures ?
 - you can convert a Future into a concurrent Future (or vice versa) using xxx

pepone: I would improve in a follow-u PR.

PR #3691, opened 2025-03-13
Add Windows workflow cache

externl:
See also:
- https://github.com/ccache/ccache/wiki/MS-Visual-Studio
- https://medium.com/@parakram.majumdar/ccache-with-msbuild-devenv-f286778e0be7

### DIFF
--- a/cpp/msbuild/ice.cpp.props
+++ b/cpp/msbuild/ice.cpp.props
@@ -6,7 +6,7 @@
   </PropertyGroup>
   <!-- Set IceHome to this source distribution. -->
   <PropertyGroup>
-      <IceHome>$(MSBuildThisFileDirectory)..\..</IceHome>
+    <IceHome>$(MSBuildThisFileDirectory)..\..</IceHome>
   </PropertyGroup>
 
   <!-- Import version settings required for library targets. -->
@@ -32,7 +32,7 @@
   </Choose>
 
   <PropertyGroup>
-      <IceToolsPath>$(IceHome)\cpp\bin\$(Platform)\$(Configuration)</IceToolsPath>
+    <IceToolsPath>$(IceHome)\cpp\bin\$(Platform)\$(Configuration)</IceToolsPath>
   </PropertyGroup>
 
   <!-- Common properties set for all builds, configurations and platforms. -->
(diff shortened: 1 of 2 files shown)


cpp/msbuild/ice.cpp.props
```diff
@@ -55,7 +55,13 @@
       <LanguageStandard>stdcpp20</LanguageStandard>
       <!-- Include directories -->
       <AdditionalIncludeDirectories>$(IceHome)\cpp\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <AdditionalIncludeDirectories>$(IceHome)\cpp\include\generated\$(Platform)\$(Configuration);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>
+        $(IceHome)\cpp\include\generated\$(Platform)\$(Configuration);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
```
pepone, line 59: I would avoid doing this, the trailing whitespaces show up in the build commands in Visual Studio, and probably in other places.

cpp/msbuild/ice.cpp.props
```diff
@@ -55,7 +55,13 @@
...
+    </ClCompile>
+    <ClCompile Condition="'$(USING_BUILD_CACHE)' == 'true'">
+      <ForcedIncludeFiles />
+      <DebugInformationFormat>OldStyle</DebugInformationFormat>
```
pepone, line 63: Is this the recommended format for the cache?

pepone (approved):

cpp/msbuild/ice.cpp.props
externl: Yep, see https://github.com/ccache/ccache/wiki/MS-Visual-Studio

cpp/msbuild/ice.cpp.props
externl: whoops, unintentional

cpp/msbuild/ice.cpp.props
pepone: I would move this settings below any other sections including ClCompile to ensure are last set and not override by others, also use a separate definition group, for example

```
<!-- Common properties required to enabled cache builds. -->
<ItemDefinitionGroup Condition="'$(USING_BUILD_CACHE)'=='true'">
  <ClCompile>
      <ForcedIncludeFiles />
      <DebugInformationFormat>OldStyle</DebugInformationFormat>
      <ObjectFileName>$(IntDir)%(FileName).obj</ObjectFileName>
    </ClCompile>
</ItemDefinitionGroup>
```

pepone (approved):

PR #3694, opened 2025-03-13
Fix doc-comments and lints in generated code (C#)

bernardnormier:
This PR adds doc-comments for constants and fixes various lints in the code generated by slice2cs.


cpp/src/Slice/SliceUtil.cpp
```diff
@@ -332,15 +332,8 @@ Slice::printGeneratedHeader(IceInternal::Output& out, const string& path, const
...
-    out << commentStyle << " <auto-generated>\n";
-    out << commentStyle << "\n";
-    out << commentStyle << " Generated from file '" << file << "'"
-        << "\n";
-    out << commentStyle << "\n";
-    out << commentStyle << " Warning: do not edit this file."
-        << "\n";
-    out << commentStyle << "\n";
-    out << commentStyle << " </auto-generated>\n";
+    out << nl << commentStyle << " <auto-generated>Generated from Slice file '" << file
```
bernardnormier, line 335: Switched to a more compact style.

slice/Ice/RemoteLogger.ice
```diff
@@ -47,7 +47,7 @@ module Ice
         LogMessageType type;
 
         /// The date and time when the {@link RemoteLogger} received this message, expressed as the number of microseconds
-        /// since the Unix Epoch (00:00:00 UTC on 1 January 1970)
+        /// since the Unix Epoch (00:00:00 UTC on 1 January 1970).
```
bernardnormier, line 50: A number of periods were missing in the doc-comments.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -910,14 +908,29 @@ Slice::Gen::Gen(const string& base, const vector<string>& includePaths, const st
...
-
     printGeneratedHeader(_out, fileBase + ".ice");
 
     _out << nl << "#nullable enable";
     _out << sp;
     _out << nl << "[assembly:Ice.Slice(\"" << fileBase << ".ice\")]";
 
     _out << sp;
+
+    /*
```
bernardnormier, line 919: We should comment out printGeneratedHeader above and reenable these pragma from time to time to make sure the generated code is lint-free.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -910,14 +908,29 @@ Slice::Gen::Gen(const string& base, const vector<string>& includePaths, const st
...
+    // Disable some warnings when auto-generated is removed from the header. See printGeneratedHeader above.
+    _out << nl << "#pragma warning disable SA1403 // File may only contain a single namespace";
+    _out << nl << "#pragma warning disable SA1611 // The documentation for parameter x is missing";
+
+    _out << nl << "#pragma warning disable CA1041 // Provide a message for the ObsoleteAttribute that marks ...";
+    _out << nl << "#pragma warning disable CA1068 // Cancellation token as last parameter";
```
bernardnormier, line 925: We should the CancellationToken lint... not urgent.

cpp/src/slice2cs/Gen.cpp
pepone: What about using DEBUG macro to conditionally enable this. No body is going to use DEBUG compilers in production, we don't ship them but we regularly do debug builds.

pepone (approved):

cpp/src/slice2cs/Gen.cpp
bernardnormier: I don't want to do this for DEBUG builds, x64 builds, macOS builds (...). All these builds should always produce the exact same code.

We could however add a slice2cs compiler flag that:
 - suppress the `<auto-generated>` header
 - generates these pragmas in the .cs code

Something like `--enable-analysis`? As an undocumented option? And we could use it in some CI builds.

pepone: Adding `--enable-analysis` sounds fine to me.

PR #3695, opened 2025-03-13
Update slice2cpp doc generation

bernardnormier:
In particular, fix the doxygen include.

### DIFF
--- a/cpp/src/slice2cpp/Gen.cpp
+++ b/cpp/src/slice2cpp/Gen.cpp
@@ -290,13 +290,10 @@ namespace
                 // Doxygen supports multiple syntaxes for operations, but none of them allow for a bare operation name.
                 // We opt for the syntax where operation names are qualified by what type they're defined on.
                 // See: https://www.doxygen.nl/manual/autolink.html#linkfunc.
-                // We also need to make sure to include the 'Async' suffix for 'amd' operations.
-                InterfaceDefPtr parent = operationTarget->interface();
-                bool amd = (parent->hasMetadata("amd") || operationTarget->hasMetadata("amd"));
 
-                string parentName = getUnqualified(parent->mappedScoped() + "Prx", source->mappedScope());
-                string opName = operationTarget->mappedName() + (amd ? "Async" : "");
-                return parentName + "::" + opName;
+                InterfaceDefPtr parent = operationTarget->interface();
+                return getUnqualified(parent->mappedScoped() + "Prx", source->mappedScope()) +
+                       "::" + operationTarget->mappedName();
             }
             if (auto builtinTarget = dynamic_pointer_cast<Builtin>(target))
             {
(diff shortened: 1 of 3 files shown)


cpp/src/slice2cpp/Gen.cpp
```diff
@@ -436,19 +433,15 @@ namespace
...
-            assert(!file.empty());
-            DefinitionContextPtr dc = p->unit()->findDefinitionContext(file);
-            assert(dc);
+        string file = p->file();
+        assert(!file.empty());
+        DefinitionContextPtr dc = p->unit()->findDefinitionContext(file);
+        assert(dc);
 
-            // TODO: why do we ignore all instances of this metadata except the first?
-            if (auto headerFile = dc->getMetadataArgs("cpp:doxygen:include"))
```
bernardnormier, line 448: This is not a reasonable API. Not surprising it didn't work.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -436,19 +433,16 @@ namespace
...
-            {
-                out << nl << "/// \\headerfile " << *headerFile;
-            }
+        // The metadata directive is cpp:doxygen, not cpp:doxygen:include. The arg of cpp:doxygen is something like
```
bernardnormier, line 441: The previous code incorrectly referenced a metadata directive with an invalid syntax: `cpp:doxygen:include` is not and can't be a metadata directive.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -940,12 +934,16 @@ Slice::Gen::validateMetadata(const UnitPtr& u)
...
 
-    // "cpp:doxygen:include"
+    // "cpp:doxygen"
+    // The metadata validation system does not support metadata names with colons. So here, for
+    // cpp:doxygen:include:Ice/Ice.h, include is either the first of two args, or part of the one arg. We consider it's
+    // a single arg that starts with 'include:'.
     MetadataInfo doxygenInfo = {
         .validOn = {typeid(Unit)},
-        .acceptedArgumentKind = MetadataArgumentKind::RequiredTextArgument,
+        .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
```
bernardnormier, line 943: commas are allowed in include statements.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1075,7 +1073,7 @@ Slice::Gen::validateMetadata(const UnitPtr& u)
     knownMetadata.emplace("cpp:type", typeInfo);
 
     // Pass this information off to the parser's metadata validation logic.
-    Slice::validateMetadata(u, "cpp", knownMetadata);
+    Slice::validateMetadata(u, "cpp", std::move(knownMetadata));
```
bernardnormier, line 1076: The one variable we want to move in this function is the big map of metadata. It's odd this std::move was missing from all compilers.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -290,13 +290,10 @@ namespace
                 // Doxygen supports multiple syntaxes for operations, but none of them allow for a bare operation name.
                 // We opt for the syntax where operation names are qualified by what type they're defined on.
                 // See: https://www.doxygen.nl/manual/autolink.html#linkfunc.
```
pepone, line 292: A few lines above we handle enums:

> // If a link to an enum isn't qualified (ie. the source and target are in the same module),
   // we have to place a '#' character in front, so Doxygen looks in the current scope.

We don't have this `#` for other types, does Doxygen resolve enums links different that links to other constructed types?

cpp/src/slice2cpp/Gen.cpp
pepone: What do you mean?

pepone (approved):

cpp/src/slice2cpp/Gen.cpp
bernardnormier: Sorry, I meant: commas are _not_ allowed in include statements. And whether or not to accept commas is the only difference between RequiredTextArgument (allows commas) and SingleArgument (does not allow commas).

cpp/src/slice2cpp/Gen.cpp
bernardnormier: Good question. What about an operation referencing another operation in the same interface, or a field referencing another field in the same struct? No `#`?

We should add a doc-comment test file that tests all these variations.

PR #3702, opened 2025-03-14
More slice2cpp doc updates

bernardnormier:
This PR continues with the slice2cpp doc updates.

The main change is to set EXTRACT_ALL to NO in the Doxyfile. This allows to detect undocumented constructs.

It also removes cond STREAM (everywhere) and cond INTERNAL (in the general code). 

`/// @private` is a better way to mark a class / function as "do not document even though it's public or protected".

### DIFF
--- a/cpp/doxygen/Doxyfile
+++ b/cpp/doxygen/Doxyfile
@@ -514,7 +514,7 @@ TIMESTAMP              = NO
 # normally produced when WARNINGS is set to YES.
 # The default value is: NO.
 
-EXTRACT_ALL            = YES
+EXTRACT_ALL            = NO
 
 # If the EXTRACT_PRIVATE tag is set to YES, all private members of a class will
 # be included in the documentation.
@@ -546,7 +546,7 @@ EXTRACT_STATIC         = NO
 # for Java sources.
 # The default value is: YES.
 
-EXTRACT_LOCAL_CLASSES  = YES
+EXTRACT_LOCAL_CLASSES  = NO
 
 # This flag is only useful for Objective-C code. If set to YES, local methods,
 # which are defined in the implementation section but not in the interface are
@@ -881,7 +881,7 @@ WARN_NO_PARAMDOC       = NO
 # will automatically be disabled.
 # The default value is: NO.
 
-WARN_IF_UNDOC_ENUM_VAL = NO
+WARN_IF_UNDOC_ENUM_VAL = YES
 
 # If the WARN_AS_ERROR tag is set to YES then doxygen will immediately stop when
 # a warning is encountered. If the WARN_AS_ERROR tag is set to FAIL_ON_WARNINGS
--- a/cpp/include/Ice/EndpointSelectionType.h
+++ b/cpp/include/Ice/EndpointSelectionType.h
@@ -9,6 +9,7 @@ namespace Ice
 {
     /**
      * Determines the order in which the Ice run time uses the endpoints in a proxy when establishing a connection.
+     * \headerfile Ice/Ice.h
      */
     enum class EndpointSelectionType : std::uint8_t
     {
--- a/cpp/include/Ice/MetricsAdminI.h
+++ b/cpp/include/Ice/MetricsAdminI.h
@@ -241,9 +241,10 @@ namespace IceInternal
                         continue; // This sub-map isn't configured.
                     }
                 }
-                _subMaps.insert(std::make_pair(
-                    p->first,
-                    std::make_pair(p->second.first, p->second.second->create(subMapPrefix, properties))));
+                _subMaps.insert(
+                    std::make_pair(
+                        p->first,
+                        std::make_pair(p->second.first, p->second.second->create(subMapPrefix, properties))));
             }
         }
 
@@ -386,9 +387,10 @@ namespace IceInternal
                 t->id = key;
 
                 p = _objects
-                        .insert(typename std::map<std::string, EntryTPtr>::value_type(
-                            key,
-                            std::make_shared<EntryT>(shared_from_this(), t, _detachedQueue.end())))
+                        .insert(
+                            typename std::map<std::string, EntryTPtr>::value_type(
+                                key,
+                                std::make_shared<EntryT>(shared_from_this(), t, _detachedQueue.end())))
                         .first;
             }
             p->second->attach(helper);
--- a/cpp/include/Ice/MetricsObserverI.h
+++ b/cpp/include/Ice/MetricsObserverI.h
@@ -113,32 +113,36 @@ namespace IceMX
 
             template<typename Y> void add(const std::string& name, Y Helper::* member)
             {
-                _attributes.insert(typename std::map<std::string, Resolver*>::value_type(
-                    name,
-                    new HelperMemberResolver<Y>(name, member)));
+                _attributes.insert(
+                    typename std::map<std::string, Resolver*>::value_type(
+                        name,
+                        new HelperMemberResolver<Y>(name, member)));
             }
 
             template<typename Y> void add(const std::string& name, Y (Helper::*memberFn)() const)
             {
-                _attributes.insert(typename std::map<std::string, Resolver*>::value_type(
-                    name,
-                    new HelperMemberFunctionResolver<Y>(name, memberFn)));
+                _attributes.insert(
+                    typename std::map<std::string, Resolver*>::value_type(
+                        name,
+                        new HelperMemberFunctionResolver<Y>(name, memberFn)));
             }
 
             template<typename I, typename O, typename Y>
             void add(const std::string& name, O (Helper::*getFn)() const, Y I::* member)
             {
-                _attributes.insert(typename std::map<std::string, Resolver*>::value_type(
-                    name,
-                    new MemberResolver<I, O, Y>(name, getFn, member)));
+                _attributes.insert(
+                    typename std::map<std::string, Resolver*>::value_type(
+                        name,
+                        new MemberResolver<I, O, Y>(name, getFn, member)));
             }
 
             template<typename I, typename O, typename Y>
             void add(const std::string& name, O (Helper::*getFn)() const, Y (I::*memberFn)() const)
             {
-                _attributes.insert(typename std::map<std::string, Resolver*>::value_type(
-                    name,
-                    new MemberFunctionResolver<I, O, Y>(name, getFn, memberFn)));
+                _attributes.insert(
+                    typename std::map<std::string, Resolver*>::value_type(
+                        name,
+                        new MemberFunctionResolver<I, O, Y>(name, getFn, memberFn)));
             }
 
             //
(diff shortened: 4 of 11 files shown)


cpp/include/Ice/EndpointSelectionType.h
```diff
@@ -9,6 +9,7 @@ namespace Ice
 {
     /**
      * Determines the order in which the Ice run time uses the endpoints in a proxy when establishing a connection.
+     * \headerfile Ice/Ice.h
```
bernardnormier, line 12: This is ok but not useful as doxygen does not generate a separate page for enum types.

cpp/include/Ice/MetricsAdminI.h
```diff
@@ -241,9 +241,10 @@ namespace IceInternal
                         continue; // This sub-map isn't configured.
                     }
                 }
-                _subMaps.insert(std::make_pair(
-                    p->first,
-                    std::make_pair(p->second.first, p->second.second->create(subMapPrefix, properties))));
+                _subMaps.insert(
```
bernardnormier, line 244: Unrelated reformatting by my newer version of clang-format.

cpp/include/Ice/MetricsObserverI.h
```diff
@@ -113,32 +113,36 @@ namespace IceMX
 
             template<typename Y> void add(const std::string& name, Y Helper::* member)
             {
-                _attributes.insert(typename std::map<std::string, Resolver*>::value_type(
-                    name,
-                    new HelperMemberResolver<Y>(name, member)));
+                _attributes.insert(
```
bernardnormier, line 116: clang-format reformatting

cpp/include/Ice/Value.h
```diff
@@ -80,13 +80,11 @@ namespace Ice
         /// @param os The output stream.
         virtual void ice_printFields(std::ostream& os) const;
 
-        /// \cond STREAM
+        /// \cond INTERNAL
```
bernardnormier, line 83: Need to migrate to `@private`.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1616,12 +1641,6 @@ Slice::Gen::ProxyVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
 
     H << eb << ';';
 
-    C << sp;
-    C << nl << "const char*" << nl << scopedPrx.substr(2) << "::ice_staticId() noexcept";
```
pepone, line 1620: Why was this removed?

pepone (approved):

cpp/src/slice2cpp/Gen.cpp
bernardnormier: I moved this code after the declaration a few lines above in the same file.

cpp/include/Ice/Value.h
bernardnormier: Actually,  we should keep `@cond INTERNAL`

cpp/doxygen/Doxyfile
```diff
@@ -2527,7 +2527,7 @@ CLASS_GRAPH            = YES
 # The default value is: YES.
 # This tag requires that the tag HAVE_DOT is set to YES.
 
-COLLABORATION_GRAPH    = YES
+COLLABORATION_GRAPH    = NO
```
bernardnormier, line 2530: I switched if off as it was always? duplicating the inheritance diagram.

cpp/include/DataStorm/DataStorm.h
```diff
@@ -1424,19 +1424,19 @@ namespace DataStorm
     }
 
     template<typename Key, typename Value, typename UpdateTag>
-    template<typename SFC>
+    template<typename SampleFilterCriteria>
```
bernardnormier, line 1427: Not using the same names as in the declaration (earlier in the same file) confuses doxygen.

cpp/include/DataStorm/Node.h
```diff
@@ -20,6 +20,8 @@ namespace DataStorm
     class DATASTORM_API NodeShutdownException final : public std::exception
```
bernardnormier, line 20: Maybe we should convert this exception into an Ice::LocalException?

cpp/include/Ice/LocalExceptions.h
```diff
@@ -74,15 +74,7 @@ namespace Ice
...
-        /**
-         * Constructs a RequestFailedException with a custom error message.
-         * @param file The file where this exception is constructed. This C string is not copied.
-         * @param line The line where this exception is constructed.
-         * @param replyStatus The reply status.
-         * @param id The identity of the Ice Object to which the request was sent.
-         * @param facet The facet to which the request was sent.
-         * @param operation The operation name of the request.
-         */
+        /// @cond INTERNAL
```
bernardnormier, line 77: Works better than `@privatesection`. See OutputStream.h

cpp/include/Ice/LoggerUtil.h
```diff
@@ -93,12 +92,14 @@ namespace Ice
     template<class L, class LPtr, void (L::*output)(const std::string&)> class LoggerOutput : public LoggerOutputBase
     {
     public:
-        inline LoggerOutput(LPtr lptr) : _logger(std::move(lptr)) {}
```
bernardnormier, line 96: The inline is superfluous.

cpp/include/Ice/OutputStream.h
```diff
@@ -406,49 +407,58 @@ namespace Ice
...
         template<typename T, typename... Te> void writeAll(const T& v, const Te&... ve)
         {
             write(v);
             writeAll(ve...);
         }
 
-        /**
-         * Writes a list of mandatory data values.
-         */
+        /// @cond INTERNAL
```
bernardnormier, line 426: Using `@privatesection / @publicsection` does not work-around this bug. `@cond INTERNAL` does.

Without this work-around, the entire page is unusable.

cpp/include/Ice/ProxyFunctions.h
```diff
@@ -201,59 +201,6 @@ namespace Ice
      */
     ICE_API bool
     proxyIdentityAndFacetEqual(const std::optional<ObjectPrx>& lhs, const std::optional<ObjectPrx>& rhs) noexcept;
-
-    /**
-     * A functor that compares the object identities of two proxies. Evaluates true if the identity in lhs
```
bernardnormier, line 206: Removed because we don't use these partially undocumented functors anywhere at all.

cpp/include/Ice/StreamHelpers.h
```diff
@@ -22,10 +22,11 @@
 
 namespace Ice
 {
-    /// \cond STREAM
-
     // StreamHelper templates used by streams to read, write and print data.
 
+    // TODO: temporary - we should document this code properly
```
bernardnormier, line 27: For a follow-up PR.

externl (approved):

cpp/include/Ice/Exception.h
```diff
@@ -15,9 +15,16 @@ namespace Ice
     class ICE_API Exception : public std::exception
     {
     public:
+        /// Default constructor.
         Exception() noexcept = default;
-        Exception(const Exception&) noexcept = default;
-        Exception& operator=(const Exception&) noexcept = default;
+
+        /// Copy constructor. Initializes this exception with a copy of other.
```
InsertCreativityHere, line 21: Is Doxygen smart enough to see that `other` has an `@param` tag underneath, or should we have:
```suggestion
        /// Copy constructor. Initializes this exception with a copy of @p other.
```

cpp/include/Ice/Exception.h
```diff
@@ -15,9 +15,16 @@ namespace Ice
...
+        /// @param other The exception to copy from.
+        Exception(const Exception& other) noexcept = default;
+
+        /// Copy assignment operator. Assigns the contents of rhs to this exception.
+        /// @param rhs The exception to copy from.
```
InsertCreativityHere, line 26: Same question:
```suggestion
        /// Copy assignment operator. Assigns the contents of @p rhs to this exception.
        /// @param rhs The exception to copy from.
```

cpp/doxygen/Doxyfile
InsertCreativityHere, line 1: Should any of these settings also be changed in the top-level `doxygen` folder?

cpp/include/Ice/StreamableTraits.h
```diff
@@ -128,9 +131,14 @@ namespace Ice
...
     template<typename T> struct StreamableTraits<T, std::enable_if_t<IsMap<T>::value || IsContainer<T>::value>>
     {
+        /// @copydoc StreamableTraits::helper
         static const StreamHelperCategory helper =
             IsMap<T>::value ? StreamHelperCategoryDictionary : StreamHelperCategorySequence;
+
+        /// The minimum number of bytes needed to marshal this type.
         static const int minWireSize = 1;
+
+        /// Is this type always encoded on a fixed number of bytes?
```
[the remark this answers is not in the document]
bernardnormier: Fixed

cpp/include/Ice/LocalExceptions.h
```diff
@@ -91,18 +83,12 @@ namespace Ice
             std::string facet,
             std::string operation);
 
-        /**
-         * Constructs a RequestFailedException without specifying the details of the current request. The details will
-         * be filled-in automatically by the Ice runtime before marshaling the exception.
-         * @param file The file where this exception is constructed. This C string is not copied.
-         * @param line The line where this exception is constructed.
-         * @param replyStatus The reply status.
-         */
```
InsertCreativityHere, line 100: Even if we don't want generate anything for this, why remove this comment and the one above?
It's still useful to people reading the code themselves. Especially for not-obvious details like
"The details will be filled-in automatically by the Ice runtime..."

cpp/include/Ice/LoggerUtil.h
```diff
@@ -129,8 +130,14 @@ namespace Ice
...
     public:
-        Trace(LoggerPtr, std::string);
+        /// Constructs a Trace object with the given logger and category.
+        /// @param logger The logger to log to.
+        /// @param category The trace category.
+        Trace(LoggerPtr logger, std::string category);
+
         ~Trace();
+
+        /// Calls Logger#trace with the collected output.
```
InsertCreativityHere, line 140: Assuming `trace` is a function here, we prefer `::`:
```suggestion
        /// Calls Logger::trace with the collected output.
```
Doxygen tolerates '#' anywhere where `::` could appear for JavaDoc compatibility, but we shouldn't abuse this.

We only use `#` for fields/enumerators, or for referencing the 'current scope', ie '#Thing' will point to something named `Thing` in the same scope as the doc-comment. Anywhere else, we should use `::`.

InsertCreativityHere (commented):
This PR seems to completely remove the generated 'namespaces' section:
After:
image
Before:
image

----

Which also means that any auto-links to namespaces are broken now.
Note how after this change, these are black, instead of blue:
After:
image
Before:
image

PR #3703, opened 2025-03-17
Modernize C++ Python extension code

pepone:
This PR modernizes C++ code in Python extension:

* Apply a number of clang-tidy fixes.
* Simplify the initialization of PyTypeObject using designated initializers.
* Remove exception handling when calling C++ noexcept methods.
* Removes workaround for old GCC versions.

### DIFF
--- a/python/modules/IcePy/BatchRequestInterceptor.cpp
+++ b/python/modules/IcePy/BatchRequestInterceptor.cpp
@@ -133,55 +133,22 @@ static PyMethodDef BatchRequestMethods[] = {
      METH_NOARGS,
      PyDoc_STR("getProxy() -> Ice.ObjectPrx")},
     {"enqueue", reinterpret_cast<PyCFunction>(batchRequestEnqueue), METH_NOARGS, PyDoc_STR("enqueue() -> None")},
-    {0, 0} /* sentinel */
+    {nullptr, nullptr} /* sentinel */
 };
 
 namespace IcePy
 {
+    // clang-format off
     PyTypeObject BatchRequestType = {
-        /* The ob_type field must be initialized in the module init function
-         * to be portable to Windows without using C++. */
-        PyVarObject_HEAD_INIT(0, 0) "IcePy.BatchRequest", /* tp_name */
-        sizeof(BatchRequestObject),                       /* tp_basicsize */
-        0,                                                /* tp_itemsize */
-        /* methods */
-        reinterpret_cast<destructor>(batchRequestDealloc), /* tp_dealloc */
-        0,                                                 /* tp_print */
-        0,                                                 /* tp_getattr */
-        0,                                                 /* tp_setattr */
-        0,                                                 /* tp_reserved */
-        0,                                                 /* tp_repr */
-        0,                                                 /* tp_as_number */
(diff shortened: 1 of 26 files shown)


python/modules/IcePy/BatchRequestInterceptor.cpp
```diff
@@ -133,55 +133,22 @@ static PyMethodDef BatchRequestMethods[] = {
      METH_NOARGS,
      PyDoc_STR("getProxy() -> Ice.ObjectPrx")},
     {"enqueue", reinterpret_cast<PyCFunction>(batchRequestEnqueue), METH_NOARGS, PyDoc_STR("enqueue() -> None")},
-    {0, 0} /* sentinel */
+    {nullptr, nullptr} /* sentinel */
 };
 
 namespace IcePy
 {
+    // clang-format off
```
pepone, line 141: I disable clang-format for PyTypeObject, to avoid extraneous indentation after the `PyVarObject_HEAD_INIT` macro

python/modules/IcePy/BatchRequestInterceptor.cpp
```diff
@@ -133,55 +133,22 @@ static PyMethodDef BatchRequestMethods[] = {
...
-        0,                                                 /* tp_dict */
-        0,                                                 /* tp_descr_get */
-        0,                                                 /* tp_descr_set */
-        0,                                                 /* tp_dictoffset */
-        0,                                                 /* tp_init */
-        0,                                                 /* tp_alloc */
-        reinterpret_cast<newfunc>(batchRequestNew),        /* tp_new */
-        0,                                                 /* tp_free */
-        0,                                                 /* tp_is_gc */
+        PyVarObject_HEAD_INIT(nullptr, 0)
```
pepone, line 143: Using designated initializers make initialization clear, and safer. The comments in the previous version refer to members that no longer exits like `tp_print` and the `0` assignment was really to a different member.

python/modules/IcePy/BatchRequestInterceptor.cpp
```diff
@@ -191,8 +158,8 @@ IcePy::initBatchRequest(PyObject* module)
     {
         return false;
     }
-    PyTypeObject* type = &BatchRequestType; // Necessary to prevent GCC's strict-alias warnings.
```
pepone, line 194: Tested with Ubuntu and Rhel G++ versions and this is no longer an issue.

python/modules/IcePy/Communicator.cpp
```diff
@@ -983,25 +983,25 @@ communicatorFindAllAdminFacets(CommunicatorObject* self, PyObject* /*args*/)
     PyTypeObject* objectType = reinterpret_cast<PyTypeObject*>(lookupType("Ice.Object"));
     PyObjectHandle plainObject{objectType->tp_alloc(objectType, 0)};
```
pepone, line 984: The usage of plainObject is puzzling, if the facet was not implementing in Python we just return this plainObject which is of no use. We should consider returning None to indicate a native implemented facet.

bernardnormier (approved):
Looks good. We should enable clang-tidy in CI for this code.

externl (approved):
Looks good!

PR #3712, opened 2025-03-17
Remove noexcept from functions that can't be noexcept

### LINKED ISSUE #3701: Review noexcept usage in C++ public API
Avoid returning a value by copy from a noexcept method if the value type does not provide a noexcept copy constructor.

See for example ObjectAdapter::getName. That is bad.

https://github.com/zeroc-ice/ice/blob/980d84df4867ff1a9810faab942d66456cde37c6/cpp/include/Ice/ObjectAdapter.h#L38

On the other hand

https://github.com/zeroc-ice/ice/blob/980d84df4867ff1a9810faab942d66456cde37c6/cpp/include/Ice/ObjectAdapter.h#L45

That is fine because shared_ptr provides a noexcept copy constructor

bernardnormier:
This PR removes 'noexcept' from various functions that can't be no except - usually because they return a value where the constructor / copy-constructor allocates memory.

It also updates a number of functions to return a `const&` and remain or become noexcept.

Fixes #3701.

### DIFF
--- a/cpp/include/Ice/ObjectAdapter.h
+++ b/cpp/include/Ice/ObjectAdapter.h
@@ -33,7 +33,7 @@ namespace Ice
 
         /// Gets the name of this object adapter.
         /// @return This object adapter's name.
-        [[nodiscard]] virtual const std::string& getName() const = 0;
+        [[nodiscard]] virtual const std::string& getName() const noexcept = 0;
 
         /// Gets the communicator this object adapter belongs to.
         /// @return This object adapter's communicator.
--- a/cpp/include/Ice/ObjectAdapter.h
+++ b/cpp/include/Ice/ObjectAdapter.h
@@ -34,8 +34,9 @@ namespace Ice
         /**
          * Get the name of this object adapter.
          * @return This object adapter's name.
+         * @remark This function remains usable after the object adapter has been deactivated.
          */
-        [[nodiscard]] virtual std::string getName() const noexcept = 0;
+        [[nodiscard]] virtual std::string getName() const = 0;
 
         /**
          * Get the communicator this object adapter belongs to.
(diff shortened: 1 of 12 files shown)


cpp/include/Ice/ObjectAdapter.h
```diff
@@ -34,8 +34,9 @@ namespace Ice
         /**
          * Get the name of this object adapter.
          * @return This object adapter's name.
+         * @remark This function remains usable after the object adapter has been deactivated.
          */
-        [[nodiscard]] virtual std::string getName() const noexcept = 0;
+        [[nodiscard]] virtual std::string getName() const = 0;
```
bernardnormier, line 39: The implementation does not simply return _name, so no noexcept.

pepone: Can't we use a constant for the unnamed OA name and always return a `cost string&`?

pepone (approved):

cpp/include/Ice/ObjectAdapter.h
bernardnormier: Yes, will fix.

externl (approved):

cpp/include/Ice/ObjectAdapter.h
```diff
@@ -31,18 +31,12 @@ namespace Ice
...
 
-        /**
-         * Get the name of this object adapter.
-         * @return This object adapter's name.
-         * @remark This function remains usable after the object adapter has been deactivated.
-         */
-        [[nodiscard]] virtual std::string getName() const = 0;
+        /// Gets the name of this object adapter.
+        /// @return This object adapter's name.
+        [[nodiscard]] virtual const std::string& getName() const = 0;
```
pepone, line 36: Should we add noexcept here too, like for getCommunicator?

bernardnormier: Yes, fixed now.

PR #3720, opened 2025-03-18
Fix operation mode check in C++, C# and Java

### LINKED ISSUE #3679: Update operation mode check during dispatch
The operation mode check performed during dispatch should always succeed when the received operation mode is Normal.

The bad situation we want to detect and report is when the client believes an operation is idempotent but the local implementation believes it's not.

See also:
https://github.com/icerpc/icerpc-csharp/blob/f109b98c1cb69755ccca7ddf3cc247982e7afb8c/src/IceRpc.Slice/IncomingRequestExtensions.cs#L13

bernardnormier:
Fixes #3679.

We should be checking this operation mode for mismatch in all languages with server-side support, but currently only check it in C++, C#, Java and Python.

We have a TODO in slice2swift:
https://github.com/zeroc-ice/ice/blob/db9df33024761cc5fb281fbb8f390b415cc041b4/cpp/src/slice2swift/SwiftUtil.cpp#L1615

I can't find where we check it in JS (if we check it).

### DIFF
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -21,7 +21,7 @@ namespace
         switch (opMode)
         {
             case Operation::Normal:
-                mode = "null"; // shorthand for most common case
+                mode += "Normal";
                 break;
             case Operation::Idempotent:
                 mode += "Idempotent";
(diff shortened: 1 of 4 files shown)


cpp/src/slice2java/Gen.cpp
```diff
@@ -21,7 +21,7 @@ namespace
         switch (opMode)
         {
             case Operation::Normal:
-                mode = "null"; // shorthand for most common case
```
bernardnormier, line 24: We should not do that anywhere.

externl (approved):

pepone (approved):

pepone:
> I can't find where we check it in JS (if we check it).

I will review JS code.

PR #3722, opened 2025-03-18
Changing doc comments to triple-slash (C++ header files)

### LINKED ISSUE #1853: Use /// doxygen comment style
We want to update the generated doc-comments as well as the existing C++ doc-comments to use the triple-slash style.

ayoitslilith:
Replaces all instances of doc comments in the C++ .h files with triple-slash comments. There are some .cpp files that tripped the script I used to do this, but considering there seems to be some intention to the usage (including but not limited to comments from outside sources concerning licensing)

Fixes #1853


cpp/test/Glacier2/dynamicFiltering/TestControllerI.h
```diff
@@ -26,9 +26,7 @@ struct SessionTuple
     SessionTuple(SessionTuple&&) = default;
 };
 
-/*
- * The test controller manipulates the router's filter tables for this session.
- */
```
InsertCreativityHere, line 31: This is a sneaky one here.

A quick primer on C++ comments; there are 4 styles:
`block comment`s look like `/* words... */`, so they have an explicit start/end (Often called C-style comment)
`block doc-comment`s look the same except they have an extra star: `/** words... */`
`line comment`s look like `// words...`, they end at the end of the line (Often called a C++-style comment)
`line doc-comment` look the same except they have an extra slash: `/// words...`

So, ideally, your script should only map block-doc-comments to line-doc-comments.
Meaning you'll want to check for `/**` at the start, not just `/*`.

----

Note that the end piece doesn't matter, it doesn't have to match the start and can have any number of stars.
So these are all normal block comments: `/* hello **/`, `/* wow */`, `/* stars *********/`
And these are all block doc-comments: `/** hello **/`, `/** wow */`, `/** stars *********/`
Only the starting piece matters.

cpp/src/Ice/UndefSysMacros.h
```diff
@@ -1,8 +1,6 @@
 // Copyright (c) ZeroC, Inc.
-
-/*! \file
-    \hideincludedbygraph
-*/
```
InsertCreativityHere, line 5: This is another special kind of block-comment that your script should leave alone.
Matching for `/**` instead of just `/*` should also avoid these though : v3

cpp/src/Glacier2/FilterManager.h
```diff
@@ -51,4 +51,4 @@ namespace Glacier2
     };
 }
 
-#endif /* FILTER_MANAGER_H */
```
InsertCreativityHere, line 54: Same idea. I'm going to stop pointing them out, I'm sure you get the idea.
Anyways, I'd just leave these normal block comments alone, and only focus on the block doc-comments for now.

cpp/test/Glacier2/dynamicFiltering/TestControllerI.h
ayoitslilith: Want to know something funny about this? The script I wrote doesn't check for those - the problem changes are only here because I thought my script had missed some edge cases. I'll go ahead and fix that.

cpp/include/Ice/Endpoint.h
```diff
@@ -20,10 +20,8 @@
 
 namespace Ice
 {
-    /**
-     * The user-level interface to an endpoint.
-     * \headerfile Ice/Ice.h
-     */
+    /// The user-level interface to an endpoint.
+    /// \headerfile Ice/Ice.h
```
ayoitslilith, line 24: Also, is this one safe to change? I'm not sure if it falls under the same category as the `! \file` ones.

bernardnormier (approved):

InsertCreativityHere (approved):
I looked at 37 of the 59 files, and those looked fine!

PR #3723, opened 2025-03-18
Add Support for `php:identifier` Metadata

InsertCreativityHere:
This PR adds support for `php:identifier` metadata, just like all the other languages.
And it updates the Slice files and tests accordingly.

The IcePHP C++ code was making decent use of the escaping logic in `slice2php`, which was removed.
So, `defineOperation`, `OperationI`, and all the code around it now need to take an extra argument (the mapped name).
And all calls to `idToClass` (which converted a typeId to a php type) were changed to `nameToClass` which takes a mapped name instead of a typeId.


php/src/Communicator.cpp
```diff
@@ -516,7 +516,7 @@ ZEND_METHOD(Ice_Communicator, identityToString)
     CommunicatorInfoIPtr _this = Wrapper<CommunicatorInfoIPtr>::value(getThis());
     assert(_this);
 
-    zend_class_entry* identityClass = idToClass("::Ice::Identity");
```
InsertCreativityHere, line 519: `idToClass` just called `nameToClass` after converting `::` to `\`.
The idea being that we were converting a type-id to a PHP class.

Now that there is no coupling between type-id and mapped named, may as well just pass the class name we know.

php/src/Communicator.cpp
```diff
@@ -835,7 +835,7 @@ ZEND_METHOD(Ice_ValueFactoryManager, add)
     ValueFactoryManagerPtr _this = Wrapper<ValueFactoryManagerPtr>::value(getThis());
     assert(_this);
 
-    zend_class_entry* factoryClass = idToClass("Ice::ValueFactory");
```
InsertCreativityHere, line 838: Is there some special PHP reason not to have a leading `\` here?
Everywhere else, we had a leading slash.

php/lib/Ice.php
```diff
@@ -40,10 +40,10 @@
 $Ice_Encoding_1_0 = new Ice\EncodingVersion(1, 0);
 $Ice_Encoding_1_1 = new Ice\EncodingVersion(1, 1);
```
InsertCreativityHere, line 42: As mentioned in the description, we have to pass both the sliceName and mappedName.
But, for these 4, obviously these names are identical.

php/test/Slice/macros/Test.ice
```diff
@@ -20,7 +20,7 @@
 
 module Test
 {
-    class Default
```
InsertCreativityHere, line 23: `default` is a PHP keyword, and class names are case-insensitive. So this had to be updated.
These Slice tests are never run cross, so instead of using `php:identifier`, I just changed the name to `YesDefault`.
To mirror the other class named `NoDefault` right under it.

cpp/src/slice2php/PHPUtil.cpp
InsertCreativityHere, line 1: The only code in the PHPUtil files were related to escaping. So, these were just outright deleted.
This escaping was also used in the IcePHP C++ code, and so some references had to be deleted from there as well.

cpp/src/slice2php/Main.cpp
```diff
@@ -262,7 +267,8 @@ CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
     }
 
     // Emit the type information.
-    _out << nl << type << " = IcePHP_defineClass('" << scoped << "', '" << escapeName(abs) << "', " << p->compactId()
+    const string abs = p->mappedScoped("\\\\");
```
pepone, line 270: We can use raw literal:

```suggestion
    const string abs = p->mappedScoped(R"(\\)");
```

php/src/Communicator.cpp
pepone: I don't think so. We should us `\\Ice`

pepone (approved):

php/src/Util.cpp
```diff
@@ -440,12 +431,19 @@ namespace
 
     zend_class_entry* createPHPException(zval* ex, const char* typeId, bool fallbackToLocalException = false)
     {
-        zend_class_entry* cls = idToClass(typeId);
+        // Convert the exception's typeId to its mapped Python type by replacing "::Ice::" with "\Ice\".
+        // This function should only ever be called on a specific list of Ice local exceptions.
```
externl, line 435:
```suggestion
        // Convert the exception's typeId to its mapped PHP type by replacing "::Ice::" with "\Ice\".
        // This function should only ever be called on a specific list of Ice local exceptions.
```

externl (approved):

php/src/Util.cpp
InsertCreativityHere: Whoops. Good catch!

php/src/Communicator.cpp
InsertCreativityHere: Yes, I think that we do overly qualify things, but for now, I'm just going to add it here to be consistent.
Instead of fixing all of our over-qualifications! : vP

cpp/src/slice2php/Main.cpp
InsertCreativityHere: I thought about it for all these backslashes, but they didn't make it any shorter, so I just left them.
I'm fine adding it here, where it at least doesn't make it any longer.

php/src/Communicator.cpp
pepone: In PHP the global scope is important in some place, depending of how types are included.

For example I remember we have issues when you dynamically require Ice files inside a PHP function, and some PHP frameworks do that.

PR #3724, opened 2025-03-18
Add operation mode check to JavaScript

pepone:


externl (approved):

bernardnormier (approved):
Looks good.

This check is about detecting a potentially problematic operation mode mismatch - a useful check in general.

With the servants implemented in JS, such problematic retries can't actually occur because all these retries involve the establishment of a new outgoing connection. And JS servants only work with bidir connections.

PR #3727, opened 2025-03-19
Add non-idempotent check in Swift

bernardnormier:
This check was missing in Swift.

### DIFF
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -1612,7 +1612,9 @@ SwiftGenerator::writeDispatchOperation(::IceInternal::Output& out, const Operati
     out << sb;
     out << nl;
 
-    // TODO: check operation mode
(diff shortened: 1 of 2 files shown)


cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -1612,7 +1612,9 @@ SwiftGenerator::writeDispatchOperation(::IceInternal::Output& out, const Operati
     out << sb;
     out << nl;
 
-    // TODO: check operation mode
+    if (op->mode() == Operation::Mode::Normal) {
```
bernardnormier, line 1615: This is actually the most correct check, and what we do already in IceRPC:
https://github.com/icerpc/icerpc-csharp/blob/f109b98c1cb69755ccca7ddf3cc247982e7afb8c/src/IceRpc.Slice/IncomingRequestExtensions.cs#L19

I'll update the other languages to perform the same check.

externl (approved):

pepone (approved):

PR #3729, opened 2025-03-19
C# StyleCop - Organize usings

ayoitslilith:
Escalates the `# Organize usings` section of .editorconfig to `warning` and uses dotnet format to reformat based on this.

### DIFF
--- a/csharp/src/Ice/Communicator.cs
+++ b/csharp/src/Ice/Communicator.cs
@@ -2,8 +2,8 @@
 
 #nullable enable
 
-using Ice.Internal;
 using System.Net.Security;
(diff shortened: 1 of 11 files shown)


csharp/src/Ice/Communicator.cs
```diff
@@ -2,8 +2,8 @@
 
 #nullable enable
 
-using Ice.Internal;
 using System.Net.Security;
+using Ice.Internal;
```
pepone, line 7: Why are system directives moved up, if we have `dotnet_sort_system_directives_first=false:warning`?

PR #3730, opened 2025-03-19
Use checkNonIdempotent in all languages

bernardnormier:
This PR updates all languages with server-side support to use a similar checkNonIdempotent check.


python/modules/IcePy/Operation.cpp
```diff
@@ -2787,12 +2781,9 @@ IcePy::TypedServantWrapper::ice_invokeAsync(
             }
         }
 
-        //
-        // See bug 4976.
```
bernardnormier, line 2791: Corresponds to ICE-4527 in Jira.

The issue is/was that 3.4.1 versions of AMI ice_xxx operations incorrectly sent Normal instead of Nonmutating in some languages. The new checkNonIdempotent logic fixes this issue too.

pepone (approved):

externl (approved):

InsertCreativityHere (approved):
Looks good to me.

PR #3733, opened 2025-03-19
StyleCop .editorconfig: .NET Coding Conventions

ayoitslilith:
Changes a whole lot of rules in the C# .editorconfig so that the notice level is elevated to warning, thus causing the formatter to change files appropriately.


pepone (commented):
There are a few lines that look too long, other than that is looking good.

csharp/src/Ice/Internal/EndpointFactoryManager.cs
```diff
@@ -46,12 +46,7 @@ public EndpointFactory get(short type)
 
     public EndpointI create(string str, bool oaEndpoint)
     {
-        string[] arr = Ice.UtilInternal.StringUtil.splitString(str, " \t\r\n");
-        if (arr == null)
-        {
-            throw new ParseException($"Failed to parse endpoint '{str}': mismatched quote");
-        }
-
+        string[] arr = Ice.UtilInternal.StringUtil.splitString(str, " \t\r\n") ?? throw new ParseException($"Failed to parse endpoint '{str}': mismatched quote");
```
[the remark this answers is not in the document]
bernardnormier: This PR was purely by editing editorconfig and running dotnet format. This suggests we don't have the correct setting for the 120 column limit?

csharp/src/Ice/Internal/CollocatedRequestHandler.cs
```diff
@@ -347,7 +347,7 @@ private void handleException(Ice.Exception ex, int requestId, bool amd)
 
     private int _requestId;
 
-    private Dictionary<OutgoingAsyncBase, int> _sendAsyncRequests = new Dictionary<OutgoingAsyncBase, int>();
-    private Dictionary<int, OutgoingAsyncBase> _asyncRequests = new Dictionary<int, OutgoingAsyncBase>();
+    private readonly Dictionary<OutgoingAsyncBase, int> _sendAsyncRequests = new Dictionary<OutgoingAsyncBase, int>();
```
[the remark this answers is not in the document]
bernardnormier: Hopefully there is a rule we can turn-on in a follow-up PR.

csharp/src/Ice/Internal/EndpointFactoryManager.cs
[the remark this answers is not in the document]
bernardnormier: As far as I can't tell, you can't configure dotnet format to wrap lines at 120 columns.

csharp/src/Ice/ConnectionI.cs
```diff
@@ -808,10 +808,7 @@ public override void message(ThreadPoolCurrent current)
                                 // The next read will read the remainder of the message.
                                 _readHeader = false;
 
-                                if (_observer is not null)
-                                {
-                                    _observer.receivedBytes(Protocol.headerSize);
-                                }
+                                _observer?.receivedBytes(Protocol.headerSize);
```
bernardnormier, line 811: I didn't know this syntax works in files without `#nullable enable`.

csharp/src/Ice/Internal/Patcher.cs
```diff
@@ -63,34 +63,19 @@ private static InvokeInfo getInvokeInfo<T>(Type t)
                 return i;
             }
 
-            MethodInfo? am = t.GetMethod("Add", [typeof(T)]);
-            if (am == null)
-            {
+            MethodInfo? am = t.GetMethod("Add", [typeof(T)]) ??
```
bernardnormier, line 66: dotnet format should also remove the `?` here, not clear why it didn't. Doesn't matter in this file.

csharp/test/Ice/assemblies/Client.cs
```diff
@@ -20,7 +20,7 @@ public override void run(string[] args)
         properties.setProperty("Ice.PreloadAssemblies", "0");
 
         string assembly =
-            String.Format("{0}/core.dll",
+            string.Format("{0}/core.dll",
 #pragma warning disable SYSLIB0012 // Type or member is obsolete
                           Path.GetFileName(Path.GetDirectoryName(Assembly.GetExecutingAssembly().CodeBase)));
```
bernardnormier, line 25: We should eliminate the few remaining "using (var ...)" in a follow-up PR.

bernardnormier (commented):
Looks good.

bernardnormier (approved):

csharp/src/Ice/ConnectionI.cs
ayoitslilith: To my understanding from trying to read up on it, `#nullable enable` only enforces a requirement for the null syntax when something is possibly null, only "enabling" an enforcement for standardization.

csharp/src/Ice/Internal/Patcher.cs
ayoitslilith: This might be a separate rule in the C# conventions - I believe one of the rules I was poking at removed this.

PR #3737, opened 2025-03-19
Remove --depend from slice2java and slice2matlab

bernardnormier:
The logic is you should never use makefile-style depend files with these languages.

See #3642.

### DIFF
--- a/man/man1/slice2java.1
+++ b/man/man1/slice2java.1
@@ -64,13 +64,6 @@ Place the generated files into directory DIR.
 Print debug information showing the operation of the Slice parser.
 
 .TP
-.BR \-\-depend\fR
-.br
-Print dependency information to standard output by default, or to the
-file specified by the --depend-file option. No code is generated when
-this option is specified.
-
-.TP
 .BR \-\-depend\-xml\fR
 .br
 Print dependency information in XML format to standard output by default,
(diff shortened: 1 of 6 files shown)


man/man1/slice2matlab.1
```diff
@@ -80,8 +73,7 @@ when this option is specified.
 .TP
 .BR \-\-depend\-file " " FILE\fR
 .br
-Directs dependency information to the specified file. The output
-format depends on whether --depend or --depend-xml is also specified.
+Directs dependency information to the specified file.
```
InsertCreativityHere, line 76: We should mention that this is useless until you also pass `--depend-xml`.

man/man1/slice2java.1
```diff
@@ -80,8 +73,7 @@ when this option is specified.
 .TP
 .BR \-\-depend\-file " " FILE\fR
 .br
-Directs dependency information to the specified file. The output
-format depends on whether --depend or --depend-xml is also specified.
+Directs dependency information to the specified file.
```
InsertCreativityHere, line 76: We should mention that this is useless until you also pass `--depend-xml`.

InsertCreativityHere (approved):

pepone (approved):
Looks good!

man/man1/slice2matlab.1
bernardnormier: Fixed

externl (approved):

PR #3738, opened 2025-03-19
Remove leftover 'implements' from Slice files and random cleanup.

InsertCreativityHere:
`implements` used to a Slice keyword, used when a `class` implemented an `interface`. We removed this, and the keyword.

`implements` is also a keyword in PHP, so I was checking our `*.ice` files for it (so I could apply `php:identifier` where necessary,
and I stumbled across some places where the old `implements` keyword wasn't properly removed.

This PR fixes those places, and some other random things I found while working on `php:identifier`.

### DIFF
--- a/cpp/test/Slice/errorDetection/IdentAsKeyword.ice
+++ b/cpp/test/Slice/errorDetection/IdentAsKeyword.ice
@@ -32,8 +32,8 @@ module Test
     interface object { void op(); }
     interface long { void op(); }
 
-    sequence<long> impLEments;
-    sequence<long> implements;
+    sequence<long> EXtends;
+    sequence<long> extends;
     sequence<long> short;
 
     sequence<module> seq1;
--- a/cpp/test/Slice/errorDetection/IllegalUseOfKeyword.ice
+++ b/cpp/test/Slice/errorDetection/IllegalUseOfKeyword.ice
@@ -1,7 +1,5 @@
 // Copyright (c) ZeroC, Inc.
 
-//
-
 module Test
 {
     exception module;
@@ -16,14 +14,13 @@ module Test
         int i;
     }
 
-    class interface;
+    class idempotent;
     class struct
     {
-        // void f();
         int i;
     }
 
-    // interface local;
+    interface interface;
     interface extends
     {
         void f();
@@ -31,7 +28,7 @@ module Test
 
     interface Foo
     {
-        void implements();
+        void module();
 
         int bar(string throws, long l, out bool void, out short s);
 
@@ -48,7 +45,7 @@ module Test
 
     sequence<int> Object;
 
-    // dictionary<int, int> LocalObject;
+    dictionary<int, int> dictionary;
 
     enum string
     {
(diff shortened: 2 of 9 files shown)


cpp/test/Slice/errorDetection/IdentAsKeyword.ice
```diff
@@ -32,8 +32,8 @@ module Test
     interface object { void op(); }
     interface long { void op(); }
 
-    sequence<long> impLEments;
-    sequence<long> implements;
```
InsertCreativityHere, line 36: This test is named `IdentAsKeyword`, and we're testing that sequence identifiers are correctly checked for keywords.
This only works if the identifier is actually a keyword. So I switched to `extends`.

cpp/test/Slice/errorDetection/IllegalUseOfKeyword.ice
```diff
@@ -1,7 +1,5 @@
 // Copyright (c) ZeroC, Inc.
 
-//
```
InsertCreativityHere, line 3: Not sure why this `//` was here... removed.

cpp/test/Slice/errorDetection/IllegalUseOfKeyword.ice
```diff
@@ -16,22 +14,21 @@ module Test
...
 
-    class interface;
+    class idempotent;
     class struct
     {
-        // void f();
         int i;
     }
 
-    // interface local;
```
InsertCreativityHere, line 26: Another place where we're supposed to be checking keywords as identifiers.
Instead of updating this file when we removed `local`, `implements`, and `LocalObject`, we just commented out 2 of them and ignored the 3rd. We should be using actual keywords in this test.

php/src/Operation.cpp
```diff
@@ -135,7 +135,7 @@ void
 IcePHP::ResultCallback::unmarshaled(zval* val, zval*, void*)
 {
     // Copy the unmarshaled value into this result callback. This increases the refcount for refcounted values and
-    // coppies non-refcound values.
```
InsertCreativityHere, line 138: A rare double-typo.

php/src/Connection.cpp
```diff
@@ -348,7 +348,7 @@ handleConnectionInfoFreeStorage(zend_object* object)
 bool
 IcePHP::connectionInit(void)
 {
-    // We register an interface and a class that implements the interface. This allows  applications to safely include
+    // We register an interface and a class that implements the interface. This allows applications to safely include
```
bernardnormier, line 351: This comment looks bogus. It was most likely referring to the Slice generated code for _local_ Slice Connection and ConnectionInfo, which no longer exists.

cpp/test/Slice/errorDetection/NotType.ice
```diff
@@ -2,29 +2,29 @@
```
InsertCreativityHere, line 2: This whole test was strange. I reworked it to better test mistaken uses of non-types.
And removed the unused `ModuleN`s and commented out classes...

bernardnormier (approved):

php/src/Connection.cpp
InsertCreativityHere: I removed the bogus comment.
Good catch.

externl (approved):

PR #3744, opened 2025-03-20
Update doc-comments in Slice files

bernardnormier:
This is first round of update of the doc-comments in Slice files. The new comments largely come from the doc-comments in IceRPC.

### DIFF
--- a/cpp/src/slice2cpp/Gen.cpp
+++ b/cpp/src/slice2cpp/Gen.cpp
@@ -2286,7 +2286,6 @@ Slice::Gen::DataDefVisitor::visitExceptionStart(const ExceptionPtr& p)
         H << nl << "/// Creates a tuple with all the fields of this exception.";
         H << nl << "/// @return A tuple with all the fields of this exception.";
         writeIceTuple(H, p->allDataMembers(), _useWstring);
-        H << sp;
 
         H << sp << nl << _dllMemberExport << "void ice_printFields(std::ostream& os) const override;";
         C << sp << nl << "void" << nl << scoped.substr(2) << "::ice_printFields(std::ostream& os) const";
@@ -2686,6 +2685,11 @@ Slice::Gen::DataDefVisitor::emitDataMember(const DataMemberPtr& p)
 
     // Use empty scope to get full qualified names in types used with future declarations.
     string scope = "";
+
+    if (!isFirstElement(p))
+    {
+        H << sp;
+    }
 
     writeDocSummary(H, p);
     H << nl << getDeprecatedAttribute(p) << typeToString(p->type(), p->optional(), scope, p->getMetadata(), _useWstring)
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -9,6 +9,8 @@
 #include <cassert>
 #include <limits>
 
+#include <iostream>
+
 using namespace std;
 using namespace Slice;
 using namespace IceInternal;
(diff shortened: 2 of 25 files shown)


cpp/src/slice2cpp/Gen.cpp
```diff
@@ -2687,6 +2686,11 @@ Slice::Gen::DataDefVisitor::emitDataMember(const DataMemberPtr& p)
     // Use empty scope to get full qualified names in types used with future declarations.
     string scope = "";
 
+    if (!isFirstElement(p))
+    {
+        H << sp;
```
bernardnormier, line 2691: Fixed spacing in generated C++ code.

pepone (commented):
Looks good.

For Python I think we should add this new type in:

https://github.com/zeroc-ice/ice/blob/114bcac1abe2851e3571eade2d11f84a729850b8/python/python/Ice/__init__.py#L76

Now it works because it is added transitively by Process.ice

And same for JavaScript it should be added to index.js.

https://github.com/zeroc-ice/ice/blob/114bcac1abe2851e3571eade2d11f84a729850b8/js/src/Ice/index.js#L38

and 

https://github.com/zeroc-ice/ice/blob/114bcac1abe2851e3571eade2d11f84a729850b8/js/src/Ice/index.js#L149

Not sure about Ruby and PHP, we import other generated files in Ice.php and Ice.rb

externl (approved):

js/src/Ice/index.js
```diff
@@ -36,6 +36,7 @@ import "./InstanceExtensions.js";
 import * as Ice_LocalException from "./LocalException.js";
 import * as Ice_LocalExceptions from "./LocalExceptions.js";
 import * as Ice_Locator from "./Locator.js";
+import * as Ice_LocatorRegistry from "./LocatorRegistry.js";
```
pepone, line 39: You should also add it below in the exports near `...Ice_Locator.Ice,`

pepone (approved):

slice/Ice/Locator.ice
```diff
@@ -5,119 +5,67 @@
...
-            throws ServerNotFoundException;
-    }
-
-    /// This interface should be implemented by services implementing the <code>Ice::Locator interface</code>. It should
-    /// be advertised through an Ice object with the identity <code>'Ice/LocatorFinder'</code>. This allows clients to
-    /// retrieve the locator proxy with just the endpoint information of the service.
+    /// Provides access to a {@link Locator} object via a fixed identity.
+    /// A LocatorFinder is always registered with identity @c Ice/LocatorFinder. This allows clients to obtain the
+    /// associated Locator proxy with just the endpoint information of the object. For example, you can use the
+    /// LocatorFinder proxy `Ice/LocatorFinder:tcp -h somehost -p 4061` to get the Locator proxy
```
bernardnormier, line 63: `<tt>...</tt>` breaks Javadoc, so I switched to markdown and enabled markdown parsing in the Slice doxyfile.

cpp/src/slice2java/Gen.cpp
```diff
@@ -96,23 +98,54 @@ namespace
...
-            result += rawLink.substr(0, hashPos);
-            result += "#";
-            result += rawLink.substr(hashPos + 1);
+            if (auto builtinTarget = dynamic_pointer_cast<Builtin>(target))
+            {
+                result << JavaGenerator::typeToObjectString(builtinTarget, TypeModeIn);
+            }
+            else
+            {
+                // TODO: Revisit this code
```
pepone, line 114: What is the issue with this code?

bernardnormier: It's fine, will remove.

slice/Ice/Locator.ice
externl: We should also enable it in the C++ Doxyfile.

externl (approved):

slice/Ice/Locator.ice
bernardnormier: Was already enabled in the C++ Doxyfile.

PR #3745, opened 2025-03-20
Dump stack when hang detected in CI

externl:
Currently Linux only, I'll add macOS (llvm) later. 

Ruff also did some formatting.

### DIFF
--- a/scripts/Expect.py
+++ b/scripts/Expect.py
@@ -430,6 +430,7 @@ class Expect(object):
         self.logfile = logfile
         self.timeout = timeout
         self.p = None
+        self.cwd = cwd
 
         if self.logfile:
             self.logfile.write('spawn: "%s"\n' % command)
@@ -736,7 +737,7 @@ class Expect(object):
     def stackDump(self):
         match platform.system():
             case "Linux":
-                cmd = [
+                dumpCmd = [
                     "gdb",
                     "-q",
                     "-n",
@@ -751,7 +752,7 @@ class Expect(object):
                     "quit",
                 ]
             case "Darwin":
-                cmd = [
+                dumpCmd = [
                     "lldb",
                     "-p",
                     f"{self.p.pid}",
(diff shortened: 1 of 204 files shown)


scripts/Util.py
```diff
@@ -1325,12 +1321,16 @@ def timedOut(self, timeout):
...
                     raise
                 except Expect.TIMEOUT:
-                    if watchDog and watchDog.timedOut(timeout):
-                        print(
-                            "process {0} is hanging - {1}".format(
-                                process, time.strftime("%x %X")
-                            )
-                        )
+                    assert watchDog
+                    if watchDog.timedOut(timeout):
```
externl, line 1325: I don't see how watchDog is never not set.

scripts/Util.py
```diff
@@ -1325,12 +1321,16 @@ def timedOut(self, timeout):
...
+                        print("process {0} is hanging - {1}".format(process, time.strftime("%x %X")))
+
+                        # If we're running in CI, dump the stack trace and exit
+                        if os.getenv("CI"):
```
externl, line 1329: When we're in CI we collect dump the stack then exit. If it's not CI then we can do it manually.

scripts/Expect.py
```diff
@@ -731,3 +732,41 @@ def test(result, expected):
...
+                    "process detach",
+                    "-o",
+                    "quit",
+                ]
+            case _:
+                print(f"stackDump not supported on {platform.system()}")
+                return
+
+        try:
+            subprocess.run(cmd)
```
pepone, line 770: It would be helpful to log a message like `Process xxx is hanging, dumping all threads...` before running gdb, just to make it more obvious what’s happening.

Also, do we run this from the main thread? If so, it might be better to redirect the output to a file and upload it as a CI artifact.

pepone (approved):
looks good

PR #3749, opened 2025-03-20
slice2cpp doc update: shows [out] parameters

bernardnormier:
This PR updates slice2cpp to document out parameters with `@param[out]`.

and misc other updates

### DIFF
--- a/cpp/include/Ice/Object.h
+++ b/cpp/include/Ice/Object.h
@@ -46,40 +46,39 @@ namespace Ice
 
         /// Tests whether this object supports a specific Slice interface.
         /// @param s The type ID of the Slice interface to test against.
-        /// @param current The Current object for the invocation.
-        /// @return True if this object has the interface
-        /// specified by s or derives from the interface
-        /// specified by s.
+        /// @param current The Current object of the incoming request.
+        /// @return @c true if this object has the interface specified by @p s or derives from the interface specified
+        /// by @p s.
         [[nodiscard]] virtual bool ice_isA(std::string s, const Current& current) const;
-        /// \cond INTERNAL
+
+        /// @private
         void _iceD_ice_isA(IncomingRequest&, std::function<void(OutgoingResponse)>);
-        /// \endcond
 
         /// Tests whether this object can be reached.
-        /// @param current The Current object for the invocation.
+        /// @param current The Current object of the incoming request.
         virtual void ice_ping(const Current& current) const;
-        /// \cond INTERNAL
-        void _iceD_ice_ping(IncomingRequest&, std::function<void(OutgoingResponse)>);
-        /// \endcond
 
-        /// Gets the Slice type IDs of the interfaces supported by this object.
-        /// @param current The Current object for the invocation.
+        /// @private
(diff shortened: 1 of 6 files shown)


cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1547,14 +1552,14 @@ Slice::Gen::ProxyVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
...
-    H << nl << "/// Obtains the Slice type ID of this interface.";
-    H << nl << "/// @return The fully-scoped type ID.";
+    H << nl << "/// Gets the type ID of the associated Slice interface.";
+    H << nl << "/// @return The string @c \"" << p->scoped() << "\".";
     H << nl << "static const char* ice_staticId() noexcept;";
 
     C << sp;
     C << nl << "const char*" << nl << scopedPrx.substr(2) << "::ice_staticId() noexcept";
     C << sb;
-    C << nl << "return \"" << scopedName << "\";";
```
bernardnormier, line 1557: This was a bug. scopedName is the "mapped" scoped name.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -2814,18 +2819,12 @@ Slice::Gen::InterfaceVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     StringList ids = p->ids();
 
     H << sp;
-    H << nl << "/// Obtains a list of the Slice type IDs representing the interfaces supported by this object.";
```
bernardnormier, line 2817: It's better to get the doc-comment via inheritance.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -2886,6 +2885,11 @@ Slice::Gen::InterfaceVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
         string_view outgoingResponseType = inIceModule ? "OutgoingResponse" : "Ice::OutgoingResponse";
 
         H << sp;
+        H << nl
+          << "/// Dispatches an incoming request to one of the member functions of this generated class, based on the "
```
bernardnormier, line 2889: except when we want to say something about the implementation.

cpp/include/Ice/Object.h
```diff
@@ -46,40 +46,39 @@ namespace Ice
 
         /// Tests whether this object supports a specific Slice interface.
         /// @param s The type ID of the Slice interface to test against.
-        /// @param current The Current object for the invocation.
-        /// @return True if this object has the interface
-        /// specified by s or derives from the interface
-        /// specified by s.
+        /// @param current The Current object of the incoming request.
+        /// @return @c true if this object has the interface specified by @p s or derives from the interface specified
```
pepone, line 50: Should we use markdown for `true` instead of @c? You started using Markdown in another PR would be more consistent to use it for all code if possible.

bernardnormier: Yes, we should.  Will fix in a follow-up PR.

The only downside of `word` vs `@c word` is the colorization with vscode-cpptools. It's minor.

Issue #3462, opened 2025-01-31
DataStorm/reliability hang (Ubuntu release CI)


bernardnormier:
From:
https://github.com/zeroc-ice/ice/actions/runs/13063024722/job/36450166492?pr=3458

```
ok
process /home/runner/work/ice/ice/cpp/test/DataStorm/reliability/build/x86_64-linux-gnu/shared/writer --Ice.Default.Host=127.0.0.1 --Test.BasePort=14200 --Ice.Warn.Connections=1 --Ice.Default.Protocol=tcp --Ice.IPv6=0 --Ice.PrintStackTraces=1 --DataStorm.Node.Multicast.Enabled=0 --DataStorm.Node.Server.Enabled=0 --DataStorm.Node.ConnectTo="tcp -p 14211" --Ice.Connection.Server.IdleTimeout=0 --Ice.Connection.Client.IdleTimeout=0 --DataStorm.Node.Name=writer-app --DataStorm.Trace.Topic=1 --DataStorm.Trace.Session=3 --DataStorm.Trace.Data=2 --Ice.Trace.Protocol=1 --Ice.LogFile=/home/runner/work/ice/ice/cpp/test/DataStorm/reliability/client-013125-0005.log pid=80273 is hanging - 01/31/25 00:06:54
process /home/runner/work/ice/ice/cpp/test/DataStorm/reliability/build/x86_64-linux-gnu/shared/writer --Ice.Default.Host=127.0.0.1 --Test.BasePort=14200 --Ice.Warn.Connections=1 --Ice.Default.Protocol=tcp --Ice.IPv6=0 --Ice.PrintStackTraces=1 --DataStorm.Node.Multicast.Enabled=0 --DataStorm.Node.Server.Enabled=0 --DataStorm.Node.ConnectTo="tcp -p 14211" --Ice.Connection.Server.IdleTimeout=0 --Ice.Connection.Client.IdleTimeout=0 --DataStorm.Node.Name=writer-app --DataStorm.Trace.Topic=1 --DataStorm.Trace.Session=3 --DataStorm.Trace.Data=2 --Ice.Trace.Protocol=1 --Ice.LogFile=/home/runner/work/ice/ice/cpp/test/DataStorm/reliability/client-013125-0005.log pid=80273 is hanging - 01/31/25 00:07:24
[...]

PR #3750, opened 2025-03-21
Additional DataStorm tracing

pepone:
This PR adds additional tracing to DataStorm session creation to help figuring out session connectivity issues, like #3462

### DIFF
--- a/cpp/src/DataStorm/NodeI.cpp
+++ b/cpp/src/DataStorm/NodeI.cpp
@@ -68,6 +68,8 @@ void
 NodeI::init()
 {
     auto self = shared_from_this();
+
+    // The instance class owns this object, so it is guaranteed to be available.
     auto instance = _instance.lock();
     assert(instance);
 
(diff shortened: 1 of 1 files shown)


externl (approved):

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -246,15 +271,32 @@ NodeI::confirmCreateSession(
     checkNotNull(publisherSession, __FILE__, __LINE__, current);
 
     unique_lock<mutex> lock(_mutex);
+    auto instance = _instance.lock();
+    assert(instance);
```
bernardnormier, line 275: Would be good to add a comment on why it's always not null.

bernardnormier (approved):

cpp/src/DataStorm/NodeI.cpp
pepone: Added a comment here, and in other similar places.

PR #3756, opened 2025-03-21
Remove dead code from Ice for JavaScript

pepone:
This PR removes some unused code from Ice for JavaScript. I using reports generated by #3754.

### DIFF
--- a/js/src/Ice/Buffer.js
+++ b/js/src/Ice/Buffer.js
@@ -52,25 +52,6 @@ export class Buffer {
         }
     }
 
-    reset() {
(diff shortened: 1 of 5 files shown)


js/src/Ice/Buffer.js
```diff
@@ -52,25 +52,6 @@ export class Buffer {
         }
     }
 
-    reset() {
```
pepone, line 55: This was probably copied from Java or C# but is not used in JavaScript.

bernardnormier (approved):

PR #3759, opened 2025-03-21
Update header in generated code

bernardnormier:
This PR updates the header (copyright etc.) generated by all Slice compilers.

For example:
```cpp
// Copyright (c) ZeroC, Inc.

// slice2cpp version 3.8.0-alpha.0
// <auto-generated>Generated from Slice file 'Locator.ice'.</auto-generated>
// clang-format off
```

It's now similar for all compilers.

(C++ only) It also replaces `\xxx` doxygen tags by `@xxx` tags.


cpp/src/slice2matlab/Main.cpp
```diff
@@ -45,7 +45,7 @@ namespace
 
         out << nl << "% Copyright (c) ZeroC, Inc.";
         out << nl << "% Generated from " << f << " by slice2matlab version " << ICE_STRING_VERSION;
```
pepone, line 47: is the missing new line between the two intentional? If so we should add a comment to clarify it.

pepone (approved):

cpp/src/slice2matlab/Main.cpp
bernardnormier: MATLAB is different from the other compilers, with the copyright notice not at the top of the file. I didn't change it in this PR. We should review it when we review the MATLAB doc-comment generation.

externl (approved):

PR #3764, opened 2025-03-24
JavaScript cleanup & lint fixes

pepone:
This PR include fixes for JavaScript to issues reported by ESLint and removes additional unused code found by looking at the coverage reports.

It also enables ESLint to run on CI.


js/gulpfile.js
```diff
@@ -116,17 +110,6 @@ gulp.task("dist", gulp.parallel(libs.map(libName => libTask(libName, "generate")
 
 gulp.task("dist:clean", gulp.parallel(libs.map(libName => libTask(libName, "clean"))));
 
-gulp.task("ice:module:package", () => gulp.src(["package.json"]).pipe(gulp.dest("node_modules/@zeroc/ice")));
```
pepone, line 119: This is not required with the current test setup. Previously we have to copy sources into node_modules to make it visible to the test suite.

js/src/Ice/AsyncResult.js
```diff
@@ -91,34 +90,6 @@ export class AsyncResult extends Promise {
         this._cancellationHandler = handler;
     }
 
-    startReadParams() {
```
pepone, line 94: The exact same methods are provided by the derived class `OutgoingAsync`. AsyncResult doesn't have an InputStream doesn't make sense to define them here.

js/src/Ice/OutgoingAsync.js
```diff
@@ -456,7 +456,7 @@ export class OutgoingAsync extends ProxyOutgoingAsyncBase {
                     }
                 }
 
-                default:
+                default: {
```
pepone, line 459: We should use `{}` when the case contains declarations. In this case we declare a const.

js/src/Ice/Reference.js
```diff
@@ -95,60 +95,6 @@ export class Reference {
         return this._communicator;
     }
 
-    getEndpoints() {
```
pepone, line 98: There is no abstract methods in JavaScript adding this isn't helpful.

js/src/Ice/TcpTransceiver.js
```diff
@@ -69,20 +75,7 @@ if (typeof net.createConnection === "function") {
                     //
                     // Socket is connected.
                     //
-                    this._desc = fdToString(this._fd, this._proxy, this._addr);
-                    this._state = StateConnected;
-                } else if (this._state === StateProxyConnectRequest) {
```
pepone, line 74: The `StateProxyConnectRequest` and `StateProxyConnectRequestPending` were never entered. And `this._proxy` was never assigned.

Seems this come from the initial port but was never implemented. Related to Socks and Http proxy support with Ice.

js/src/Ice/WSTransceiver.js
```diff
@@ -25,11 +25,26 @@ if (typeof WebSocket !== "undefined") {
     const StateClosed = 4;
 
     WSTransceiver = class {
-        constructor(instance) {
+        constructor(instance, secure, addr, resource) {
```
pepone, line 28: The transceiver construction logic was split between the constructor and a static method. That is due to how we previously include the classes in NodeJS and Browser. With the current code we can use the constructors directly witch is clear.

bernardnormier (approved):

PR #3766, opened 2025-03-24
Update Slice doc comments for IceBox and IceStorm

bernardnormier:
Also Slice doc-comments for IceDiscovery and IceLocatorDiscovery. Note that the IceDiscovery/IceLocatorDiscovery API is purely internal.


slice/IceStorm/IceStorm.ice
```diff
@@ -36,148 +36,142 @@ module IceStorm
...
     }
 
-    /// This exception indicates that an attempt was made to subscribe a proxy for which a subscription already exists.
+    /// The exception that is thrown when attempting to subscribe a proxy for which a subscription already exists.
     exception AlreadySubscribed
     {
     }
 
-    /// This exception indicates that an attempt was made to subscribe a proxy that is null.
-    exception InvalidSubscriber
```
bernardnormier, line 63: Removed because we no longer use this exception. It's also redundant with the null proxy we perform in various operations that do not throw a user exception when this check fails.

slice/IceStorm/Metrics.ice
```diff
@@ -17,26 +17,26 @@
 ["swift:module:IceStorm:MX"]
 module IceMX
 {
-    /// Provides information on IceStorm topics.
+    /// Provides information about IceStorm topics.
     class TopicMetrics extends Metrics
     {
-        /// Number of events published on the topic by publishers.
+        /// The number of events published on topics by publishers.
```
bernardnormier, line 23: Looking at the IceStorm code, it does not appear we create a TopicMetrics per-topic ... however, I could be mistaken.

externl (approved):

externl:
I wonder if we should move the IceDiscovery/IceLocatorDiscovery slice files into an internal folder that we don't include as a part of the public Slice files. No one should need to use these right?

slice/IceLocatorDiscovery/IceLocatorDiscovery.ice
```diff
@@ -11,27 +11,29 @@
...
 
-/// IceLocatorDiscovery is an Ice plug-in that enables the discovery of IceGrid and custom locators via UDP multicast.
+/// IceLocatorDiscovery is an Ice plug-in that enables the discovery of Ice locators (such as IceGrid) via UDP
+/// multicast.
 module IceLocatorDiscovery
 {
-    /// The Ice lookup reply interface must be implemented by clients which are searching for Ice locators. Ice locator
-    /// implementations invoke on this interface to provide their locator proxy.
+    /// Represents a callback object implemented by IceLocatorDiscovery clients. It allows IceDiscoIceLocatorDiscovery
+    /// very clients to receive replies to Lookup requests.
```
pepone, line 19: Seems you wanted to write something else `It allows IceDiscoIceLocatorDiscovery very clients`

slice/IceStorm/Metrics.ice
pepone: I think this depend on the configured metrics views, the per-topic is just the "id" attribute.

https://github.com/zeroc-ice/ice/blob/e67ce93b8da2b79cede68b5db77af1859723194f/cpp/src/IceStorm/InstrumentationI.cpp#L26-L27

pepone (approved):

slice/IceLocatorDiscovery/IceLocatorDiscovery.ice
bernardnormier: Fixed.

slice/IceStorm/Metrics.ice
bernardnormier: Fixed.

bernardnormier:
> I wonder if we should move the IceDiscovery/IceLocatorDiscovery slice files into an internal folder that we don't include as a part of the public Slice files. No one should need to use these right?

They are purely internal, and I don't believe we ship these files.

slice/IceStorm/IceStorm.ice
```diff
@@ -36,148 +36,142 @@ module IceStorm
...
+        /// Creates a link to another topic. All events originating on this topic will also be sent to this other topic.
         /// @param linkTo The topic to link to. This proxy is never null.
-        /// @param cost The cost to the linked topic.
-        /// @throws LinkExists Raised if a link to the same topic already exists.
+        /// @param cost The cost of the link.
+        /// @throws LinkExists Thrown when a link to @p linkTo already exists.
         void link(Topic* linkTo, int cost) throws LinkExists;
 
-        /// Destroy the link from this topic to the given topic <code>linkTo</code>.
+        /// Destroys a link from this topic.
```
InsertCreativityHere, line 111: Same thing, I think the original words really do make it clearer here.
```suggestion
        /// Destroys a link from this topic to the provided topic.
```

InsertCreativityHere (approved):
Looks good.

slice/IceBox/ServiceManager.ice
```diff
@@ -17,57 +17,57 @@
...
         void startService(string service)
             throws AlreadyStartedException, NoSuchServiceException;
 
-        /// Stop an individual service.
+        /// Stops a service.
         /// @param service The service name.
-        /// @throws AlreadyStoppedException If the service is already stopped.
-        /// @throws NoSuchServiceException If no service could be found with the given name.
+        /// @throws AlreadyStoppedException Thrown when the service is already stopped.
+        /// @throws NoSuchServiceException Thrown when IceBox does not know a service named @p service.
```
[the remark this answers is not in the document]
bernardnormier: I find `know` better. There is no find logic in IceBox.

PR #3768, opened 2025-03-24
Reject Metadata on Nested-Module-Definitions

InsertCreativityHere:
This PR:
1) Adds `bool usesNestedSyntax` to `Module`, which stores  whether a module used nested-module syntax.
2) Changes the `MetadataValidator` to reject all metadata that is applied to nested-module-definitions.
3) Adds a `getStatus` method to `Unit` which returns an exit code based on whether any errors have been emitted.

Up until now, `libSlice` only emitted errors during parsing. So the `slice2xxx` compilers only checked for errors from parsing. Now that we emit an error during the validation phase, I updated all the `slice2xxx` compilers to also check for errors at the end of everything: parsing, validation, and code-gen.


cpp/src/ice2slice/Main.cpp
```diff
@@ -222,6 +222,7 @@ compile(const vector<string>& argv)
                 }
             }
 
+            status |= p->getStatus();
```
InsertCreativityHere, line 225: After we've finished parsing, validation, and code-gen, we want to check if any errors were reported.
We use `|=` because we don't want to overwrite an `EXIT_FAILURE` with an `EXIT_SUCCESS`.

We also don't want to return because we're looping over a list of files to compile. One of them failing shouldn't stop compilation of all the others.

ruby/src/IceRuby/Slice.cpp
```diff
@@ -135,6 +135,11 @@ IceRuby_loadSlice(int argc, VALUE* argv, VALUE /*self*/)
             //
             out << "# encoding: utf-8\n";
             generate(u, all, includePaths, out);
```
InsertCreativityHere, line 137: Copied from the lines above:
```
if (!icecpp->close() || parseStatus == EXIT_FAILURE)
{
    u->destroy();
    throw RubyException(rb_eArgError, "Slice parsing failed for `%s'", cmd.c_str());
}
```

python/modules/IcePy/Slice.cpp
```diff
@@ -143,6 +143,12 @@ IcePy_loadSlice(PyObject* /*self*/, PyObject* args)
         out.setUseTab(false);
 
         generate(u, all, includePaths, out);
```
InsertCreativityHere, line 145: Copied from the lines above:
```
if (!icecpp->close() || parseStatus == EXIT_FAILURE)
{
    PyErr_Format(PyExc_RuntimeError, "Slice parsing failed for `%s'", cmd);
    u->destroy();
    return nullptr;
}
```

cpp/test/Slice/errorDetection/IllegalMetadata.err
InsertCreativityHere, line 1: I renamed the old `IllegalMetadata` test to `FileMetadataMustComeFirst`, since it was specifically focused on that.

bernardnormier:
I find validating metadata as part of the code generation sub-optimal. The logic was not introduced by this PR but is made worse by this PR since this validation can now generate errors - not just warnings.

It would be nicer for "validate metadata" to be a separate step. And if this step fails (completes with an error), we don't generate any code. With this PR, I can get into the odd situation where my Slice compilation fails with an error and yet overwrites all my generated files.

InsertCreativityHere:
Yes, @externl had a similar thought.

I started fixing this, by moving validation out of the `gen` functions, and adding a new check for errors before code-gen, which if hit, skips code-gen.

But I think I'll ask you both to ignore this issue for now, and put all those changes in an immediate follow-up PR.
Much of the `validateMetadata` implementations are inside the `Gen` classes. So fixing it moves around code which has nothing to do with the this PR's focus.

cpp/test/Slice/errorDetection/IllegalMetadata.ice
```diff
@@ -1,7 +1,11 @@
...
+["deprecated"] ["fake-metadata", "deprecated"]
+module Foo::Bar {}
 
-module Hello {}
-
-[["suppress-warning"]] // error
+// Error: metadata cannot be applied to nested module definitions.
+// Make sure we emit the error even for language-specific metadata.
+["cpp:type:string"]
+module A::B::C {}
```
InsertCreativityHere, line 11:
```suggestion
module A {
    class B {
        bool b;
        string s;
    }
}
```

cpp/test/Slice/errorDetection/IllegalMetadata.ice
```diff
@@ -1,7 +1,11 @@
 // Copyright (c) ZeroC, Inc.
```
InsertCreativityHere, line 1:
```suggestion
// Copyright (c) ZeroC, Inc. testing.
```

cpp/src/Slice/MetadataValidation.cpp
```diff
@@ -146,7 +146,18 @@ MetadataVisitor::visitUnitStart(const UnitPtr& p)
 bool
 MetadataVisitor::visitModuleStart(const ModulePtr& p)
 {
-    p->setMetadata(validate(p->getMetadata(), p));
+    const MetadataList moduleMetadata = p->getMetadata();
+    if (p->usesNestedSyntax && !moduleMetadata.empty())
+    {
```
InsertCreativityHere, line 151:
```suggestion
    MetadataList moduleMetadata = p->getMetadata();
    if (p->usesNestedSyntax && !moduleMetadata.empty())
    {
        const int hello = 17;
```

cpp/test/Slice/errorDetection/IllegalMetadata.ice
```diff
@@ -1,7 +1,11 @@
...
 
-[["suppress-warning"]] // okay
+// Error: metadata cannot be applied to nested module definitions.
+// Make sure we only emit 1 error for all the metadata; not 1 error per metadata here.
+["deprecated"] ["fake-metadata", "deprecated"]
+module Foo::Bar {}
```
InsertCreativityHere, line 6:
```slice
module A {
    class B {
        bool b;
        string s;
    }
}
```

```ice
module A {
    class B {
        bool b;
        string s;
    }
}
```

cpp/test/Slice/errorDetection/IllegalMetadata.ice
InsertCreativityHere, line 1: // Suggestion:
```suggestion
{ "hello":"there" }
```
// Ice:
```ice
{ "hello":"there" }
```
// JSON:
```json
{ "hello":"there" }
```
// Slice:
```slice
{ "hello":"there" }
```

----

// Suggestion:
```suggestion
{ "Copyright (c)":"ZeroC, Inc." }
```
// Ice:
```ice
{ "Copyright (c)":"ZeroC, Inc." }
```
// JSON:
```json
{ "Copyright (c)":"ZeroC, Inc." }
```
// Slice:
```slice
{ "Copyright (c)":"ZeroC, Inc." }
```

----

Suggestion:
```suggestion
// Copyright (c) ZeroC, Inc. hello
```
JSON:
```json
// Copyright (c) ZeroC, Inc. hello
```
Ice:
```ice
// Copyright (c) ZeroC, Inc. hello
```
Slice:
```slice
// Copyright (c) ZeroC, Inc. hello
```

externl (approved):

cpp/src/Slice/Parser.cpp
```diff
@@ -4990,6 +4994,12 @@ Slice::Unit::unit() const
     return dynamic_pointer_cast<Unit>(self);
 }
 
+int
+Slice::Unit::getStatus() const
+{
+    return (_errors ? EXIT_FAILURE : EXIT_SUCCESS);
```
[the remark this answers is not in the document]
InsertCreativityHere: Is there some obscure difference here?
Or are you saying for clarity's sake?

cpp/src/Slice/MetadataValidation.cpp
```diff
@@ -146,7 +146,18 @@ MetadataVisitor::visitUnitStart(const UnitPtr& p)
...
+        // Metadata cannot be applied to modules that used nested-module-syntax, since it's ambiguous in meaning.
+        // We issue an error, and clear the metadata, so that we're in a valid state for further validation.
+        p->unit()->error(p->file(), p->line(), "metadata cannot be applied to nested module definitions");
```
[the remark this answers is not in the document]
InsertCreativityHere: I was using the wording C++uses
```
/tmp/SAClaD0ZbY/main.cpp:13:1: error: a nested namespace definition cannot have attributes
   13 | namespace [[deprecated]] Foo::Bar {}
      | ^~~~~~~~~
```

I'll switch it back to what I had originally,
but I don't like talking about the "syntax" here, and I think calling this _thing_ a `module defined using nested module syntax` is verbose. We should come up with some better term for this _thing_.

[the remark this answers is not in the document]
bernardnormier: should be "with _the_ nested module syntax".

cpp/src/Slice/Parser.cpp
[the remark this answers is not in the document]
externl: Clarity sake.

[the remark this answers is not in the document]
InsertCreativityHere: Fixed!

cpp/src/Slice/MetadataValidation.cpp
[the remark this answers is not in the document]
InsertCreativityHere: I think `... with the nested module syntax` sounds a little awkward.
After some thought, I switched to `using` instead of `with`. I think it sounds the most natural here.

PR #3770, opened 2025-03-24
Update IceGrid Admin Slice definitions

bernardnormier:
Updates IceGrid/Admin.ice and IceGrid/Exceptions.ice.


slice/Ice/LocatorRegistry.ice
```diff
@@ -45,8 +45,8 @@ module Ice
         /// @throws AdapterNotFoundException Thrown when the locator only allows registered object adapters to register
         /// their endpoints and no object adapter with this adapter ID was registered with the locator.
         /// @throws AdapterAlreadyActiveException Thrown when an object adapter with the same adapter ID has already
-        /// registered its endpoints.
-        // Note: idempotent is not correct but kept for backwards compatibility with old implementations.
+        /// registered its endpoints. Since this operation is marked idempotent, this exception may be thrown when the
```
bernardnormier, line 48: From the Ice point of view, idempotent just means the Ice client runtime can or does retry when the connection drops while waiting for response. 

For this particular operation, it's arguably correct as far as the LocatorRegistry impl is concerned (e.g. IceGrid registry) but results in a very confusing behavior for the caller. The Ice runtime catches this exception and throws instead an ObjectAdapterIdInUse exception ... and the user scratches his head trying to figure out what happened.

slice/Ice/OperationMode.ice
```diff
@@ -30,8 +29,8 @@ module Ice
         ["swift:identifier:nonmutating"]
         Nonmutating,
 
-        /// An idempotent operation. Invoking such an operation twice in a row must result in the same object state
```
bernardnormier, line 33: It's really about retries and at-most-once semantics, not about the standard/math definition of idempotent.

slice/IceGrid/Admin.ice
```diff
@@ -625,160 +619,143 @@ module IceGrid
...
 
-        /// Set the observer proxies that receive notifications when the state of the registry or nodes changes.
+        /// Sets the observer proxies that receive notifications when the state of the registry or nodes changes.
         /// @param registryObs The registry observer.
         /// @param nodeObs The node observer.
         /// @param appObs The application observer.
         /// @param adptObs The adapter observer.
         /// @param objObs The object observer.
-        /// @throws ObserverAlreadyRegisteredException Raised if an observer is already registered with this registry.
+        /// @throws ObserverAlreadyRegisteredException Thrown when an observer is already registered with this registry.
```
bernardnormier, line 758: The idempotent is also arguably correct here - multiple calls with the same args won't change the object state. However, the client can get a very confusing ObserverAlreadyRegisteredException when such a retry occurs.

slice/IceGrid/Admin.ice
```diff
@@ -625,160 +619,143 @@ module IceGrid
...
-    /// This interface allows applications to monitor IceGrid well-known objects.
+    /// Monitors well-known objects that are added, updated or removed using {@link Admin}.
     interface ObjectObserver
     {
-        /// <code>objectInit</code> is called after the registration of an observer to indicate the state of the
-        /// registry.
-        /// @param objects The objects registered with the {@link Admin} interface (not through the deployment
-        /// mechanism).
+        /// Provides the initial list of well-known objects to the observer.
+        /// @param objects The well-known objects registered using {@link Admin} (not through the deployment mechanism).
```
pepone, line 717: I think we should move `not through the deployment mechanism` to the interface description. My understanding this is true for all methods not just init.

slice/IceGrid/Admin.ice
```diff
@@ -806,99 +783,99 @@ module IceGrid
...
+        /// Opens a server stdout file for reading.
         /// @param id The server id.
         /// @param count Specifies where to start reading the file. If negative, the file is read from the beginning.
-        /// If 0 or positive, the file is read from the last <code>count</code> lines.
-        /// @return An iterator to read the file. The returned proxy is never null.
-        /// @throws FileNotAvailableException Raised if the file can't be read.
-        /// @throws ServerNotExistException Raised if the server doesn't exist.
-        /// @throws NodeUnreachableException Raised if the node could not be reached.
-        /// @throws DeploymentException Raised if the server couldn't be deployed on the node.
+        /// If 0 or positive, the file is read from the last @p count lines.
```
pepone, line 829: Should we use code blocks for number literals? Like `0`.

slice/IceGrid/Exception.ice
```diff
@@ -17,21 +17,21 @@
...
     {
         /// The name of the application.
         string name;
     }
 
-    /// This exception is raised if a server does not exist.
+    /// The exception that is thrown when IceGrid does not know a server with the provided server ID.
     exception ServerNotExistException
     {
         /// The identifier of the server.
```
pepone, line 30: We need to review how we document "id" fields. In this file we use:

* The identifier
* The id
* The identity

pepone (approved):

externl (approved):

slice/IceGrid/Exception.ice
bernardnormier: Yes. I think it should be ID (adapter ID, server ID) or identity. Never identifier.

InsertCreativityHere (approved):

slice/IceGrid/Admin.ice
```diff
@@ -55,10 +55,10 @@ module IceGrid
         Destroyed
     }
 
-    /// A dictionary of proxies.
+    /// A dictionary of string to proxies.
```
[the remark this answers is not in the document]
bernardnormier: Feel free to add a better doc-comment.

slice/IceGrid/Admin.ice
```diff
@@ -118,20 +118,20 @@ module IceGrid
         /// The operating system name.
         string os;
 
-        /// The network name of the host running this node (as defined in uname()).
+        /// The network name of the host running this node (as returned by `uname`).
```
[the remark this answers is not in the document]
bernardnormier: Fixed

slice/IceGrid/Admin.ice
bernardnormier: Replaced by Otherwise.

slice/IceGrid/Exception.ice
bernardnormier: Fixed

PR #3772, opened 2025-03-25
Avoid registering an interceptor for publisher sessions

pepone:
This PR removes the default servant registered for publisher sessions. This default servant is only required to dispatch request sent by a publisher using a sample filter to the matching subscriber. It never dispatches a request to a publisher.

### DIFF
--- a/cpp/src/DataStorm/NodeI.cpp
+++ b/cpp/src/DataStorm/NodeI.cpp
@@ -61,7 +61,6 @@ NodeI::~NodeI()
     assert(_subscribers.empty());
     assert(_publishers.empty());
     assert(_subscriberSessions.empty());
-    assert(_publisherSessions.empty());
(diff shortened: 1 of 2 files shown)


pepone:
Replaced by #3775

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -489,21 +484,11 @@ shared_ptr<SessionI>
 NodeI::getSession(const Identity& ident) const
 {
     unique_lock<mutex> lock(_mutex);
-    if (ident.category == "s")
+    assert(ident.category == "s");
```
pepone, line 487: This is bogus because `getSession` is also used by `Node::getSessionConnection(string_view ident) const noexcept` which is part of the public API.

Issue #3773, opened 2025-03-25
Unused BatchRequestQueue::isEmpty


pepone:
This internal method is not called anywhere. It is the same for C++, Java, C#, and JavaScript. We should just remove it.

https://github.com/zeroc-ice/ice/blob/615ef9cd6bb468e37147cc4491c8318bdb9367bb/cpp/src/Ice/BatchRequestQueue.cpp#L190

PR #3774, opened 2025-03-25
JavaScript dead code removal and cleanup

pepone:


js/src/Ice/BatchRequestQueue.js
```diff
@@ -76,8 +76,4 @@ export class BatchRequestQueue {
     destroy(ex) {
         this._exception = ex;
     }
-
-    isEmpty() {
```
pepone, line 80: Dead code, we should remove this method in other languages too. See #3773

js/src/Ice/ConnectRequestHandler.js
```diff
@@ -32,16 +32,14 @@ export class ConnectRequestHandler {
         }
 
         if (!this.initialized()) {
-            for (let i = 0; i < this._requests.length; i++) {
```
pepone, line 35: This was and odd way to eliminate and element from an array.

js/src/Ice/ConnectionRequestHandler.js
pepone, line 1: Renamed to `FixedRequestHandler.js` see below

js/src/Ice/EnumBase.d.ts
```diff
@@ -14,21 +14,6 @@ declare module "@zeroc/ice" {
              */
             constructor(name: string, value: number);
 
-            /**
```
pepone, line 17: Enums should be compared by reference, not clear why we added these methods in the first place:

https://doc.zeroc.com/ice/3.7/language-mappings/javascript-mapping/client-side-slice-to-javascript-mapping/javascript-mapping-for-enumerations

I think is safe to remove them.

externl (approved):

bernardnormier (approved):

PR #3776, opened 2025-03-25
C# Line Delengthening

ayoitslilith:
Breaks up lines into more lines because they were all unsightly (and too long for readability!)
Not all lines are broken up. Some I was uncertain about - see [here](https://pastebin.com/18CGcQY0). Others, I made the decision to not break them up for readability's sake. These are <5 characters over the limit (but most often just one) , and the characters that *are* over the limit are typically of the `)` or `;` varieties. I can go back and list these, however, they will likely be listed with the ones I was uncertain regarding a fix.


csharp/test/Ice/scope/Server.cs
```diff
@@ -23,7 +23,8 @@ public override Test.MyStruct[]
         }
 
         public override Dictionary<string, Test.MyStruct>
-        opMyStructMap(Dictionary<string, Test.MyStruct> s1, out Dictionary<string, Test.MyStruct> s2, Ice.Current current)
+        opMyStructMap(Dictionary<string, Test.MyStruct> s1,
+            out Dictionary<string, Test.MyStruct> s2, Ice.Current current)
```
pepone, line 27:
```suggestion
        opMyStructMap(
            Dictionary<string, Test.MyStruct> s1,
            out Dictionary<string, Test.MyStruct> s2, Ice.Current current)
```

csharp/test/Ice/scope/Server.cs
```diff
@@ -102,8 +103,8 @@ public override Test.Inner.Inner2.MyClass[]
         }
 
         public override Dictionary<string, Test.Inner.Inner2.MyClass>
-        opMyClassMap(Dictionary<string, Test.Inner.Inner2.MyClass> c1, out Dictionary<string, Test.Inner.Inner2.MyClass> c2,
-               Ice.Current current)
+        opMyClassMap(Dictionary<string, Test.Inner.Inner2.MyClass> c1,
```
pepone, line 106: split each parameter in a separate line.

pepone (commented):
Added some comments, didn't check all tests.

csharp/src/Ice/Instrumentation.cs
```diff
@@ -275,7 +275,8 @@ public interface CommunicatorObserver
     /// <param name="ctx">The context specified by the user.
     /// </param>
     /// <returns>The invocation observer to instrument the invocation.</returns>
-    InvocationObserver getInvocationObserver(ObjectPrx prx, string operation, System.Collections.Generic.Dictionary<string, string> ctx);
+    InvocationObserver getInvocationObserver(
+        ObjectPrx prx, string operation, System.Collections.Generic.Dictionary<string, string> ctx);
```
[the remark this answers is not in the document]
bernardnormier: `System.Collections.Generic.` is used implicitly, just remove the qualification.

bernardnormier:
> Others, I made the decision to not break them up for readability's sake. These are <5 characters over the limit (but most often just one) , and the characters that _are_ over the limit are typically of the `)` or `;` varieties.

Sometimes > 120 lines is ok, but the reasons you list are not good ones. The reason to go over 120 lines are:
 - long string literals where breaking them up decreases readability
 - long list of return values in a method signature, where we want to keep the return next to the method name

csharp/test/Ice/scope/Server.cs
bernardnormier: each parameter should be on its own line in this case

csharp/src/Ice/Internal/IPEndpointI.cs
```diff
@@ -57,7 +57,8 @@ public override EndpointI connectionId(string connectionId)
         }
     }
 
-    public override void connectors_async(EndpointI_connectors callback) => instance_.resolve(host_, port_, this, callback);
+    public override void connectors_async(
+        EndpointI_connectors callback) => instance_.resolve(host_, port_, this, callback);
```
pepone, line 61:
```suggestion
    public override void connectors_async(EndpointI_connectors callback) =>
        instance_.resolve(host_, port_, this, callback);
```

csharp/src/Ice/Internal/PropertyNames.cs
```diff
@@ -13,28 +13,103 @@ internal sealed class PropertyNames
         false,
```
pepone, line 13: You must not update this file manually see the comment in the header.

csharp/src/Ice/Internal/ThreadPool.cs
```diff
@@ -365,7 +365,10 @@ public void joinWithAllThreads()
 
     public bool serialize() => _serialize;
 
-    protected sealed override void QueueTask(System.Threading.Tasks.Task task) => execute(() => TryExecuteTask(task), null);
+    protected sealed override void QueueTask(
+        System.Threading.Tasks.Task task) => execute(
+            () => TryExecuteTask(task),
+            null);
```
pepone, line 371: Can we add a using for `System.Threading.Tasks` and then write

```
protected sealed override void QueueTask(Task task) =>
    execute(() => TryExecuteTask(task), null);
```
?

csharp/src/Ice/MetricsObserverI.cs
```diff
@@ -194,20 +194,32 @@ public string resolve(MetricsHelper<T> helper, string attribute)
...
+            string name,
+            System.Reflection.MethodInfo method) => _attributes.Add(name, new MethodResolverI(name, method));
+
+        public void add(
+            string name,
+            System.Reflection.FieldInfo field) => _attributes.Add(name, new FieldResolverI(name, field));
+
+        public void add(
+            string name,
+            System.Reflection.MethodInfo method,
```
pepone, line 207: I think adding a using for `System.Reflection` would be clear to shorten these names, unless it causes a conflict.

csharp/src/Ice/OutputStream.cs
```diff
@@ -2256,7 +2256,8 @@ internal void setEncoding(EncodingVersion encoding)
     //
     private EncodingVersion _encoding;
 
-    private bool isEncoding_1_0() => _encapsStack != null ? _encapsStack.encoding_1_0 : _encoding.Equals(Util.Encoding_1_0);
+    private bool isEncoding_1_0() => _encapsStack !=
+        null ? _encapsStack.encoding_1_0 : _encoding.Equals(Util.Encoding_1_0);
```
pepone, line 2260: This still looks and odd formatting. I will avoid putting the operands of the `!=` in different lines.

csharp/src/Ice/MetricsObserverI.cs
```diff
@@ -194,20 +194,32 @@ public string resolve(MetricsHelper<T> helper, string attribute)
...
+            System.Reflection.FieldInfo field) =>
+                _attributes.Add(name, new MemberFieldResolverI(name, method, field));
+
+        public void add(
+            string name,
+            System.Reflection.MethodInfo method,
+            System.Reflection.PropertyInfo property) =>
+                _attributes.Add(name, new MemberPropertyResolverI(name, method, property));
```
pepone, line 215: No need to indent `_attributes` here.
```suggestion
      _attributes.Add(name, new MemberPropertyResolverI(name, method, property));
```

pepone (approved):
Looks good.

bernardnormier (approved):

PR #3778, opened 2025-03-25
Update Communicator, Initialize and Properties doc in C++

bernardnormier:
This PR updates the doc-comments for Communicator, Properties and Intialize.h in C++.


cpp/include/Ice/Communicator.h
```diff
@@ -20,41 +20,37 @@ namespace Ice
     class LocatorPrx;
     class RouterPrx;
 
-    /// The central object in Ice. One or more communicators can be instantiated for an Ice application.
-    /// @see Logger
-    /// @see ObjectAdapter
```
bernardnormier, line 25: I don't see the point of these random `@see`. In particular, why add a `@see` to types in function signatures?

cpp/include/Ice/Communicator.h
```diff
@@ -65,19 +61,15 @@ namespace Ice
...
-        /// @return `true` if the communicator has been shut down; `false` otherwise.
+        /// Checks whether this communicator has been shut down.
+        /// @return `true` if the communicator has been shut down, `false` otherwise.
         /// @see #shutdown
         [[nodiscard]] bool isShutdown() const noexcept;
 
-        /// Convert a stringified proxy into a proxy.
-        /// For example, <code>MyCategory/MyObject:tcp -h some_host -p 10000</code> creates a proxy that refers to the
-        /// Ice object having an identity with a name "MyObject" and a category "MyCategory", with the server running on
-        /// host "some_host", port 10000. If the stringified proxy does not parse correctly, the operation throws
```
bernardnormier, line 76: As a matter of style, the summary/description should never include "if you pass a bad param, this function throws Xyz". The correct way to document such behavior is in `@throws`.

cpp/include/Ice/Properties.h
```diff
@@ -192,6 +192,50 @@ namespace Ice
         std::vector<std::string> _optInPrefixes;
         mutable std::mutex _mutex;
     };
+
+    /// Creates an empty property set.
+    /// @return A new empty property set.
+    ICE_API PropertiesPtr createProperties();
```
bernardnormier, line 198: Moved from Initialize.h. Didn't change otherwise (except for the doc-comments).

cpp/include/Ice/Communicator.h
```diff
@@ -340,23 +320,81 @@ namespace Ice
...
         friend ICE_API CommunicatorPtr initialize(InitializationData);
+        /// @endcond
+
         friend ICE_API IceInternal::InstancePtr IceInternal::getInstance(const Ice::CommunicatorPtr&);
         friend ICE_API IceInternal::TimerPtr IceInternal::getInstanceTimer(const Ice::CommunicatorPtr&);
 
         const IceInternal::InstancePtr _instance;
     };
+
+    /// A helper class that uses Resource Acquisition Is Initialization (RAII) to hold a communicator instance, and
```
bernardnormier, line 340: Moved from Initialize.h. Didn't change except for doc-comments and a little more inlining.

cpp/include/Ice/Communicator.h
```diff
@@ -20,41 +20,37 @@ namespace Ice
...
     class RouterPrx;
 
-    /// The central object in Ice. One or more communicators can be instantiated for an Ice application.
-    /// @see Logger
-    /// @see ObjectAdapter
-    /// @see Properties
-    /// @see ValueFactory
+    /// Represents the main entry point into the Ice runtime. You create a communicator with
+    /// {@link Ice::initialize(int&, const char*[], InitializationData)} or another `initialize` overload.
+    /// @see ::initialize(int&, const char*[], InitializationData)
```
pepone, line 25: Why do you qualify the `@link Ice::` and use `::initialize` for `@see`?

cpp/include/Ice/Communicator.h
```diff
@@ -20,41 +20,37 @@ namespace Ice
...
         ~Communicator();
 
-        /// Destroy the communicator. This operation calls {@link #shutdown} implicitly. Calling {@link #destroy} cleans
-        /// up memory, and shuts down this communicator's client functionality and destroys all object adapters.
-        /// Subsequent calls to {@link #destroy} are ignored.
+        /// Destroys the communicator. This function calls #shutdown implicitly. Calling #destroy destroys all object
+        /// adapters, shuts down this communicator's client functionality, and cleans up memory. Subsequent calls to
+        /// #destroy are no-op.
+        /// @see CommunicatorHolder
         /// @see #shutdown
```
pepone, line 37: There is already a link to shutdown in the description `#shutdown implicitly` we can remove this `@see`.

cpp/include/Ice/Communicator.h
```diff
@@ -65,19 +61,15 @@ namespace Ice
...
-        /// For example, <code>MyCategory/MyObject:tcp -h some_host -p 10000</code> creates a proxy that refers to the
-        /// Ice object having an identity with a name "MyObject" and a category "MyCategory", with the server running on
-        /// host "some_host", port 10000. If the stringified proxy does not parse correctly, the operation throws
-        /// ParseException. Refer to the Ice manual for a detailed description of the syntax supported by stringified
-        /// proxies.
+        /// Converts a stringified proxy into a proxy.
         /// @param str The stringified proxy to convert into a proxy.
-        /// @return The proxy, or nullopt if <code>str</code> is an empty string.
+        /// @return The proxy, or nullopt if `str` is an empty string.
+        /// @throws ParseException Thrown when @p str is not a valid proxy string.
```
pepone, line 72: Can we document the template param?

cpp/include/Ice/Communicator.h
```diff
@@ -199,129 +187,121 @@ namespace Ice
         /// @see Connection::setAdapter
         void setDefaultObjectAdapter(ObjectAdapterPtr adapter);
 
-        /// Get the implicit context associated with this communicator.
+        /// Gets the implicit context associated with this communicator.
         /// @return The implicit context associated with this communicator; returns null when the property
```
pepone, line 191: should we use null or nullptr in the descriptions?

cpp/include/Ice/Communicator.h
```diff
@@ -199,129 +187,121 @@ namespace Ice
...
         [[nodiscard]] std::optional<RouterPrx> getDefaultRouter() const;
 
-        /// Set a default router for this communicator. All newly created proxies will use this default router. To
-        /// disable the default router, null can be used. Note that this operation has no effect on existing proxies.
-        /// You can also set a router for an individual proxy by calling the operation <code>ice_router</code> on the
-        /// proxy.
-        /// @param rtr The default router to use for this communicator.
+        /// Sets a default router for this communicator. All newly created proxies will use this default router. This
+        /// function has no effect on existing proxies.
+        /// @param rtr The default router to use for this communicator. May be nullopt.
```
pepone, line 215: Maybe mention nullopt purposes it to clean existing router.

cpp/include/Ice/Communicator.h
```diff
@@ -340,23 +320,81 @@ namespace Ice
...
+        /// @param other The holder to move from.
+        CommunicatorHolder(CommunicatorHolder&& other) = default;
+
+        CommunicatorHolder(const CommunicatorHolder&) = delete;
+
+        /// Assignment operator. Destroys the current communicator (if any) and adopts a new communicator.
+        /// @param communicator The communicator to adopt.
+        CommunicatorHolder& operator=(CommunicatorPtr communicator) noexcept;
+
+        CommunicatorHolder& operator=(const CommunicatorHolder&) noexcept = delete;
```
pepone, line 371: Do we need tp document the const variant?

cpp/include/Ice/Properties.h
```diff
@@ -43,126 +45,124 @@ namespace Ice
...
         /// @param options The command-line options.
         /// @return The command-line options that do not start with the specified prefix, in their original order.
         StringSeq parseCommandLineOptions(std::string_view prefix, const StringSeq& options);
 
-        /// Convert a sequence of command-line options into properties. All options that begin with one of the following
-        /// prefixes are converted into properties: <code>--Ice</code>, <code>--IceBox</code>, <code>--IceGrid</code>,
-        /// <code>--IceSSL</code>, <code>--IceStorm</code>, and <code>--Glacier2</code>.
+        /// Converts a sequence of command-line options into properties. All options that begin with one of the
+        /// following prefixes are converted into properties: `--Ice`, `--IceBox`, `--IceGrid`, `--IceSSL`,
+        /// `--IceStorm`, and `--Glacier2`.
```
pepone, line 147: I think we need to add `--DataStorm` here.

pepone (commented):
Just a few comments, I will finish review later.

cpp/include/Ice/Communicator.h
bernardnormier: Ice::initialize(correct params) doesn't auto-link for some reason, and @see requires an auto-link string.

bernardnormier: Updated comment.

cpp/include/Ice/Communicator.h
bernardnormier: nullptr in C++

cpp/include/Ice/Communicator.h
bernardnormier: Done

cpp/include/Ice/Communicator.h
bernardnormier: Which const variant?

In general, we don't document `=delete`  functions.

cpp/include/Ice/Properties.h
bernardnormier: See https://github.com/zeroc-ice/ice/blob/c8672febb7b70f8299999e68c1b9805a3db1634d/config/PropertyNames.xml#L403

DataStorm is currently an opt-in prefix, meaning you don't get it automatically. @extern can give more details.

bernardnormier: Also, if you look at opt-in="false" in this file, you'll see that this list is quite incomplete.

bernardnormier: Looking at the code,`isOptIn` is only used in set, so presumably we parse all these prefixes all the time.

cpp/include/Ice/Initialize.h
```diff
@@ -262,261 +118,136 @@ namespace Ice
...
-    /// @param argv Command-line arguments, possibly containing
-    /// options to set properties. If the arguments include
-    /// a <code>--Ice.Config</code> option, the corresponding configuration
-    /// files are parsed. If the same property is set in a configuration
-    /// file and in the arguments, the arguments take precedence.
-    /// Recognized options are removed from this vector upon return.
-    /// @param initData Configuration data for the new Communicator.
+    /// @name Communicator
+    /// Communicator initialization functions.
+    /// @{
```
externl, line 131: Typo?

cpp/include/Ice/Initialize.h
```diff
@@ -262,261 +118,136 @@ namespace Ice
...
+
+    /// Creates a new communicator.
+    /// @param[in,out] argc The number of arguments in @p argv. When this function parses properties from @p argv, it
+    /// reshuffles the arguments so that the remaining arguments start at the beginning of @p argv, and updates @p argc
+    /// accordingly.
+    /// @param argv The command-line arguments. This function parses arguments starting with `--` and one of the
+    /// reserved prefixes (Ice, IceSSL etc.) as properties for the new communicator. If there is an argument starting
```
externl, line 138:
```suggestion
    /// reserved prefixes (Ice, IceSSL, etc.) as properties for the new communicator. If there is an argument starting
```

cpp/include/Ice/Initialize.h
```diff
@@ -262,261 +118,136 @@ namespace Ice
...
 
-    /// Initializes a new communicator.
+    /// Creates a new communicator.
     /// @param configFile The name of an Ice configuration file.
     /// @return The new communicator.
     ICE_API CommunicatorPtr initialize(std::string_view configFile);
 
-    /// Obtains the per-process logger. This logger is used by all communicators that do not have their
-    /// own specific logger established at the time a communicator is created.
+    /// @}
```
externl, line 214: leftovers?

cpp/include/Ice/Properties.h
externl, line 147: "All options that begin with one of the" is not true as it depends on the configured opt in prefixes.

I would remove the `and` and use `, etc.` as it's not complete. Or maybe just say something like:

```
Converts a sequence of command-line options that start with an Ice property prefix (`--Ice.`, `--IceSSL.`, etc.) into properties. 
```

@pepone  is correct, `--DataStorm` could be added to this list. 

This function still ultimately calls `setProperty` which checks the opt-in prefixes.

externl (approved):

cpp/include/Ice/Communicator.h
```diff
@@ -65,19 +59,16 @@ namespace Ice
         /// @see #shutdown
         void waitForShutdownAsync(std::function<void()> completed) noexcept;
 
-        /// Check whether communicator has been shut down.
-        /// @return `true` if the communicator has been shut down; `false` otherwise.
+        /// Checks whether this communicator has been shut down.
```
pepone, line 62: Does this means, shutdown was called or shutdown has completed?

cpp/include/Ice/Communicator.h
```diff
@@ -93,17 +84,16 @@ namespace Ice
...
+        /// @return The stringified proxy, or an empty string if `obj` is nullopt.
         /// @see #stringToProxy
         std::string proxyToString(const std::optional<ObjectPrx>& obj) const;
 
-        /// Convert a set of proxy properties into a proxy. The "base" name supplied in the <code>property</code>
-        /// argument refers to a property containing a stringified proxy, such as <code>MyProxy=id:tcp -h localhost -p
-        /// 10000</code>. Additional properties configure local settings for the proxy, such as
-        /// <code>MyProxy.PreferSecure=1</code>. The "Properties" appendix in the Ice manual describes each of the
-        /// supported proxy properties.
+        /// Converts a set of proxy properties into a proxy. The "base" name supplied in the @p property argument refers
```
pepone, line 93: Should this be parameter instead of argument?

```suggestion
        /// Converts a set of proxy properties into a proxy. The "base" name supplied in the @p property parameter refers
```

cpp/include/Ice/Communicator.h
```diff
@@ -121,23 +111,22 @@ namespace Ice
...
         /// @param ident The identity to convert into a string.
         /// @return The "stringified" identity.
         [[nodiscard]] std::string identityToString(const Identity& ident) const;
 
-        /// Create a new object adapter. The endpoints for the object adapter are taken from the property
-        /// <code><em>name</em>.Endpoints</code>.
+        /// Creates a new object adapter. The endpoints for the object adapter are taken from the property
+        /// `name.Endpoints`.
         /// It is legal to create an object adapter with the empty string as its name. Such an object adapter is
         /// accessible via bidirectional connections or by collocated invocations that originate from the same
```
pepone, line 128: can we remove the `that originate from the same....` this is the definition of collocared.

cpp/include/Ice/Communicator.h
```diff
@@ -199,129 +186,121 @@ namespace Ice
         /// @see Connection::setAdapter
         void setDefaultObjectAdapter(ObjectAdapterPtr adapter);
 
-        /// Get the implicit context associated with this communicator.
-        /// @return The implicit context associated with this communicator; returns null when the property
-        /// Ice.ImplicitContext is not set or is set to None.
+        /// Gets the implicit context associated with this communicator.
+        /// @return The implicit context associated with this communicator; returns `nullptr` when the property
+        /// `Ice.ImplicitContext` is not set or is set to None.
```
pepone, line 191: Maybe quote "None" to make clear is a property value.

```suggestion
        /// `Ice.ImplicitContext` is not set or is set to "None".
```

cpp/include/Ice/Communicator.h
```diff
@@ -199,129 +186,121 @@ namespace Ice
...
         [[nodiscard]] std::optional<RouterPrx> getDefaultRouter() const;
 
-        /// Set a default router for this communicator. All newly created proxies will use this default router. To
-        /// disable the default router, null can be used. Note that this operation has no effect on existing proxies.
-        /// You can also set a router for an individual proxy by calling the operation <code>ice_router</code> on the
-        /// proxy.
-        /// @param rtr The default router to use for this communicator.
+        /// Sets the default router of this communicator. All newly created proxies will use this default router. This
+        /// function has no effect on existing proxies.
+        /// @param rtr The new default router. Use `nullopt` to remove the default router.
```
pepone, line 214: In a follow up PR we should look into improve the parameter names, it would make the documentation easier to read.

cpp/include/Ice/Communicator.h
```diff
@@ -199,129 +186,121 @@ namespace Ice
...
-        /// is set and the provided object adapter has a {@link Locator}, createAdmin registers the Admin's Process
-        /// facet with the {@link Locator}'s {@link LocatorRegistry}. createAdmin must only be called once; subsequent
-        /// calls raise InitializationException.
-        /// @param adminAdapter The object adapter used to host the Admin object; if null and Ice.Admin.Endpoints is
-        /// set, create, activate and use the Ice.Admin object adapter.
+        /// Adds the Admin object with all its facets to the provided object adapter. If `Ice.Admin.ServerId`
+        /// is set and the provided object adapter has a Locator, #createAdmin registers the Admin's Process facet with
+        /// the Locator's LocatorRegistry.
+        /// @param adminAdapter The object adapter used to host the Admin object; if null and `Ice.Admin.Endpoints` is
+        /// set, create, activate and use the `Ice.Admin` object adapter.
```
pepone, line 277: I find the description hard to understand.

Maybe clear something like:

```
The object adapter used to host the Admin object. If null and `Ice.Admin.Endpoints` is set, a new 
`Ice.Admin` object adapter will be created, activated, and used instead.
```

cpp/include/Ice/Communicator.h
pepone: missed it was deleted.

cpp/include/Ice/Properties.h
pepone: Is there a user case for application code to call this? maybe we should we mark as internal to remove it from doxygen docs.

cpp/include/Ice/Properties.h
pepone: ins't this deprecated?

cpp/include/Ice/Properties.h
```diff
@@ -192,6 +192,50 @@ namespace Ice
...
+    /// Creates a property set initialized from command-line arguments and a default property set.
+    /// @param seq The command-line arguments. This function parses arguments starting with `--` and one of the
+    /// reserved prefixes (Ice, IceSSL, etc.) as properties and removes these elements from the list. If there is an
+    /// argument starting with `--Ice.Config`, this function loads the specified configuration file. When the same
+    /// property is set in a configuration file and through a command-line argument, the command-line setting takes
+    /// precedence.
+    /// @param defaults Default values for the property set. Settings in configuration files and the arguments override
+    /// these defaults.
+    /// @return A new property set initialized with the properties that were removed from the argument vector.
+    ICE_API PropertiesPtr createProperties(StringSeq& seq, const PropertiesPtr& defaults = nullptr);
```
pepone, line 209: I think this should be deprecated, and document that it just calls the matching properties constructor.

externl: We've deprecated them in other languages but not in C++. IIRC it has something to do with the properties constructors.

cpp/include/Ice/Communicator.h
bernardnormier: Argument is correct.

cpp/include/Ice/Initialize.h
bernardnormier: Removed

cpp/include/Ice/Properties.h
bernardnormier: Update to "etc". Other concerns are for a separate PR.

cpp/include/Ice/Communicator.h
```diff
@@ -20,41 +20,35 @@ namespace Ice
...
         void shutdown() noexcept;
 
-        /// Wait until the application has called {@link #shutdown} (or {@link #destroy}). On the server side, this
-        /// operation blocks the calling thread until all currently-executing operations have completed. On the client
-        /// side, the operation simply blocks until another thread has called {@link #shutdown} or {@link #destroy}.
-        /// A typical use of this operation is to call it from the main thread, which then waits until some other thread
-        /// calls {@link #shutdown}. After shut-down is complete, the main thread returns and can do some cleanup work
-        /// before it finally calls {@link #destroy} to shut down the client functionality, and then exits the
-        /// application.
+        /// Waits until the application has called #shutdown or #destroy. On the server side, this function blocks the
```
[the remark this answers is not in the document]
bernardnormier: You're right, I reworked this comment.

cpp/include/Ice/Communicator.h
```diff
@@ -41,12 +41,9 @@ namespace Ice
...
         void shutdown() noexcept;
 
-        /// Waits until the application has called #shutdown or #destroy. On the server side, this function blocks the
-        /// calling thread until all currently-executing dispatches have completed. On the client-side, the function
-        /// simply blocks until another thread has called #shutdown or #destroy.
-        /// A typical use of this function is to call it from the main thread, which then waits until some other thread
-        /// calls #shutdown. After shutdown is complete, the main thread returns and can do some cleanup work before it
-        /// finally calls #destroy to shut down the client functionality, and then exits the application.
+        /// Waits for shutdown to complete. In particular, this function waits for all operation dispatches to complete.
+        /// In a client application, without no operation dispatch, this function returns as soon as another thread
```
pepone, line 45: without no,  doesn't sound correct here.

pepone (approved):

cpp/include/Ice/Communicator.h
```diff
@@ -199,129 +182,121 @@ namespace Ice
...
+        /// Gets the observer object of this communicator.
+        /// @return This communicator's observer object.
         [[nodiscard]] Instrumentation::CommunicatorObserverPtr getObserver() const noexcept;
 
-        /// Get the default router for this communicator.
-        /// @return The default router for this communicator.
-        /// @throws CommunicatorDestroyedException if the communicator has been destroyed.
+        /// Gets the default router of this communicator.
+        /// @return The default router of this communicator.
+        /// @throws CommunicatorDestroyedException Thrown when the communicator is destroyed.
```
InsertCreativityHere, line 204: `has been` feels more right to me. Otherwise it reads like
"when you destroy the communicator, we will throw this exception" (the present tense)
when what we want is the "perfect present passive tense":
> X will happen now if Y happened in the past.

```suggestion
        /// @throws CommunicatorDestroyedException Thrown when the communicator has been destroyed.
```

InsertCreativityHere (approved):
Looks better!
Just small Englishy things.

cpp/include/Ice/Communicator.h
```diff
@@ -199,129 +182,121 @@ namespace Ice
...
-        /// is set and the provided object adapter has a {@link Locator}, createAdmin registers the Admin's Process
-        /// facet with the {@link Locator}'s {@link LocatorRegistry}. createAdmin must only be called once; subsequent
-        /// calls raise InitializationException.
-        /// @param adminAdapter The object adapter used to host the Admin object; if null and Ice.Admin.Endpoints is
-        /// set, create, activate and use the Ice.Admin object adapter.
+        /// Adds the Admin object with all its facets to the provided object adapter. If `Ice.Admin.ServerId`
+        /// is set and the provided object adapter has a Locator, #createAdmin registers the Admin's Process facet with
+        /// the Locator's LocatorRegistry.
+        /// @param adminAdapter The object adapter used to host the Admin object; when null and `Ice.Admin.Endpoints` is
+        /// set, this function uses instead the `Ice.Admin` object adapter, after creating and activating this adapter.
```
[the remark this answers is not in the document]
bernardnormier: I removed instead completely.

cpp/include/Ice/Communicator.h
```diff
@@ -65,19 +56,16 @@ namespace Ice
...
         [[nodiscard]] bool isShutdown() const noexcept;
 
-        /// Convert a stringified proxy into a proxy.
-        /// For example, <code>MyCategory/MyObject:tcp -h some_host -p 10000</code> creates a proxy that refers to the
-        /// Ice object having an identity with a name "MyObject" and a category "MyCategory", with the server running on
-        /// host "some_host", port 10000. If the stringified proxy does not parse correctly, the operation throws
-        /// ParseException. Refer to the Ice manual for a detailed description of the syntax supported by stringified
-        /// proxies.
+        /// Converts a stringified proxy into a proxy.
+        /// @tparam Prx The type of the proxy to return.
```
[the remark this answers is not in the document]
bernardnormier: I think that's fine.  This function is provided for backward compatibility. New applications should use the proxy constructor.

PR #3780, opened 2025-03-25
Remove cpp:const from operations implemented by Ice services

bernardnormier:
This PR removes the `cpp:const` metadata from all operations implemented by Ice services (Glacier2, IceGrid, IceStorm).
I did remove the `cpp:const` metadata itself and kept it on "callback" interfaces and other interfaces that users may reasonably implement, such as Locator, Router and PermissionsVerifier.

The `cpp:const` metadata is useless since:
 - all dispatches in C++ are performed through a `shared_ptr<non-const T>`
 - dispatch functions are not supposed to be called locally for other purposes ... the only situation where they could be called on a const object.

Nevertheless, we keep it for backward compatibility. There is no need to keep it for interfaces that only these Ice services implement, hence this PR.

### DIFF
--- a/cpp/src/Glacier2/RouterI.cpp
+++ b/cpp/src/Glacier2/RouterI.cpp
@@ -111,7 +111,7 @@ Glacier2::RouterI::addProxies(ObjectProxySeq proxies, const Current& current)
 }
 
 string
-Glacier2::RouterI::getCategoryForClient(const Current&) const
+Glacier2::RouterI::getCategoryForClient(const Current&)
 {
     assert(false); // Must not be called in this router implementation.
     return {};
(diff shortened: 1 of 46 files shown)


externl:
In the spirit of const correctness I think leaving these as `cpp:const` makes more sense. Calling from const objects isn't the only use case for making a function const. It improves code clarity and helps the compiler catch any unintentional mutations of the object.

pepone:
As @externl  pointed I think `const` here was about telling this method doesn't mutate the object state.

bernardnormier:
`const` in a C++ function signature is not an implementation detail, otherwise you could override a non-const virtual function with a const virtual function. And the Slice metadata wouldn't make sense.

`const` tells the (local) client: you can call this function on a const object. Except here, the local client (Ice) which calls the abstract base class never uses const objects.

externl:
I'm not saying isn't an implementation detail. I'm saying it has other uses than just for const objects.

PR #3783, opened 2025-03-26
Add Python Ice.EventLoopAdapter for async event loop integration

pepone:
This PR introduces a new event loop adapter that enables integration of Ice with the application's chosen event loop. It eliminates the need to manually call Ice.wrap_future and provides a mechanism that can support event loops other than asyncio's default loop.


python/modules/IcePy/Communicator.cpp
```diff
@@ -70,112 +71,22 @@ communicatorNew(PyTypeObject* type, PyObject* /*args*/, PyObject* /*kwds*/)
     return self;
 }
 
+// Only called by Ice.Initialize
+// IcePy.Communicator(args: [str]|None, initData: Ice.InitializationData|None, configFile: str|None)
+
 extern "C" int
 communicatorInit(CommunicatorObject* self, PyObject* args, PyObject* /*kwds*/)
 {
-    //
```
pepone, line 76: I move the validation of the parameters to Ice.initialize, which is the only caller of IcePy.Communicator.

python/modules/IcePy/Communicator.cpp
```diff
@@ -70,112 +71,22 @@ communicatorNew(PyTypeObject* type, PyObject* /*args*/, PyObject* /*kwds*/)
...
-
-    if (arg2 == Py_None)
-    {
-        arg2 = nullptr;
-    }
-
-    PyObject* initDataType = lookupType("Ice.InitializationData");
-
-    if (arg1)
+    if (!PyArg_ParseTuple(args, "OOO", &argList, &initData, &configFile))
```
pepone, line 83: Switched this internal function to have 3 positional arguments, as mentioned in the comment above.

python/modules/IcePy/Communicator.cpp
```diff
@@ -259,19 +168,6 @@ communicatorInit(CommunicatorObject* self, PyObject* args, PyObject* /*kwds*/)
         return -1;
     }
 
-    //
-    // Remaining command line options are passed to the communicator
-    // as an argument vector in case they contain plug-in properties.
-    //
-    int argc = static_cast<int>(seq.size());
```
pepone, line 266: Unless I’m missing something, this seems completely unnecessary. We’re calling the C++ `Ice::initialize`, which can take a `StringSeq` by reference and update it.

python/modules/IcePy/Operation.cpp
```diff
@@ -381,17 +382,18 @@ operationInit(OperationObject* self, PyObject* args, PyObject* /*kwds*/)
...
-        mappedName,
-        mode,
-        amd,
-        format,
-        metadata,
-        inParams,
-        outParams,
-        returnType,
-        exceptions));
+    self->op = new OperationPtr(
```
pepone, line 385: unrelated clang-format update.

python/python/Ice/Util.py
```diff
@@ -1,51 +1,71 @@
...
     ----------
     args : list of str, optional
-        An argument list, such as sys.argv.
-    data : InitializationData, optional
-        An instance of InitializationData.
+        The command-line arguments. This function parses arguments starting with `--` and one of the
+        reserved prefixes (Ice, IceSSL etc.) as properties for the new communicator. If there is an argument starting
+        with `--Ice.Config`, this function loads the specified configuration file. When the same property is set in a
+        configuration file and through a command-line argument, the command-line setting takes precedence.
+    initData : InitializationData, optional
```
pepone, line 28: I was undecided or whether keep `data` for this keyword parameter or update it to `initData`. I decided to changed it because previously `data` was used for both the initialization data and the config file.

Keeping data would make this backwards compatible:

```
Ice.initilize(data=initData)
```

Which now should be:

```
Ice.initialize(initData=initData)
```

python/modules/IcePy/Operation.cpp
pepone: Seems we are using an older clang-format version in CI.

python/modules/IcePy/Communicator.cpp
pepone: Because this is an internal function and the only caller is Ice.initialize, I moved the validation logic there in Python code which is simpler.

I can add an assert here to ensure that we never pass both. Having 3 parameters is simpler because we don't have to do the additional type checking here.

python/modules/IcePy/Communicator.cpp
```diff
@@ -526,9 +416,7 @@ communicatorShutdownCompleted(CommunicatorObject* self, PyObject* /*args*/)
                 PyObjectHandle discard{callMethod(futureGuard.get(), "set_result", Py_None)};
             });
 
-    // Release the reference to the future object and let the caller take ownership.
-    // The callback maintains its own strong reference to the future.
-    return future.release();
+    return IcePy::wrapFuture(*self->communicator, future.get());
```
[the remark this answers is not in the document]
pepone: I will add a comment in the code:

Here future holds a strong reference to the Ice.Future object, and keeps it alive until this method return.

Ice::wrapFuture always returns a strong reference.

Here when we call wrapFuture using `future.get()` we are passing a borrow reference. The caller code still owns future.

python/python/Ice/Util.py
```diff
@@ -1,51 +1,71 @@
...
     Returns
     -------
     Communicator
-        A new communicator instance.
+        The new communicator.
+
+    .. code-block:: python
+
+    with Ice.initialize(sys.argv, eventLoop=asyncio.get_running_loop()) as communicator:
+        greeter =  VisitorCenter.GreeterPrx(communicator, "greeter:tcp -h localhost -p 4061")
```
externl, line 45:
```suggestion
        greeter = VisitorCenter.GreeterPrx(communicator, "greeter:tcp -h localhost -p 4061")
```

externl (approved):
Looks good.

python/modules/IcePy/Communicator.cpp
externl: I would keep using data here as a union: `data: Union[Ice.InitializationData,str,null]`

python/python/Ice/Util.py
```diff
@@ -1,51 +1,71 @@
...
-    communicator = IcePy.Communicator(args, data)
-    return Communicator(communicator, coroutineExecutor)
+
+    if args and not isinstance(args, list):
+        raise InitializationException("args must be a list of strings")
+
+    if initData and not isinstance(initData, InitializationData):
+        raise InitializationException("initData must be an instance of Ice.InitializationData")
+
+    if initData and configFile:
```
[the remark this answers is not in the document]
pepone: Already did, it no longer deal with multiple parameter combinations.

python/modules/IcePy/Communicator.cpp
pepone: > I would keep using data here as a type union

This would require to check the types here again, and we already do in Ice.initialize. I don't see the upside of using an union.

python/test/TestHelper.py
```diff
@@ -47,15 +47,15 @@ def createTestProperties(self, args=[]):
...
+    def initialize(self, initData=None, properties=None, args=[], eventLoop=None):
         if initData is None:
             initData = Ice.InitializationData()
             if properties:
                 initData.properties = properties
             else:
                 initData.properties = self.createTestProperties(args)
 
-        communicator = Ice.initialize(initData)
+        communicator = Ice.initialize(initData=initData, eventLoop=eventLoop)
```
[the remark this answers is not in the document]
pepone: Fixed as suggested.

python/python/Ice/Util.py
```diff
@@ -1,51 +1,74 @@
...
+    initData : InitializationData, optional
+        Options for the new communicator. This argument and the `configFile` argument are mutually exclusive.
+    configFile : str, optional
+        The path to a configuration file. This argument and the `initData` argument are mutually exclusive.
+    eventLoop : asyncio.AbstractEventLoop, optional
+        An asyncio event loop used to run coroutines and wrap futures. If provided, a new event loop adapter is created
+        and configured with the communicator. This adapter is responsible for executing coroutines returned by Ice
+        asynchronous dispatch methods and for wrapping Ice futures (from Ice Async APIs) into asyncio futures.
+        This argument and the `initData` argument are mutually exclusive. If the `initData` argument is provided, the
+        event loop adapter can be set using the InitializationData.eventLoopAdapter member.
```
bernardnormier, line 37: field or attribute instead of member?

pepone: Yes. Attribute is more correct.

PR #3786, opened 2025-03-26
Fixed Assertion Hit in `csLinkFormatter` Fallback Case

InsertCreativityHere:
This PR fixes 3 bugs in how the Slice compiler handles `@link` tags:

1) We no long support linking to parameters with `@link`. In the future we will support `@p` tags for this.
2) Using a relative link like `#myOperation` on an operation, will search inside that operation's _interface_ instead of it's parameters. For example:
```
interface Hello {
    /// Before this PR, #myOperation linked to the parameter.
    /// Now it will link to the operation underneath here.
    greet(bool myOperation);

    myOperation();
}
```

3) Removes the call to `splitScopedName` in the fallback case of `csLinkFormatter`.
This function requires it's argument to be scoped, but a user could type `{@link Hello}`; `Hello` is unscoped, and would crash the compiler. Instead of calling this function, we perform the replacement by hand.

We should also be removing any leading scope separators. Not related to the bugs here, but we should be.

----

Note that 1 and 2 match the behavior of Doxygen.
So this PR brings us closer to Doxygen.

### DIFF
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -166,7 +166,7 @@ namespace
             // Replace any '#' scope separators with '.'s.
             replace(targetS.begin(), targetS.end(), '#', '.');
             // Remove any leading scope separators.
-            if (targetS.find(".") == 0)
+            if (targetS.find('.') == 0)
             {
                 targetS.erase(0, 1);
             }
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -155,8 +155,21 @@ namespace
             // Replace "::"" by "." in the raw link. This is for the situation where the user passes a Slice type
             // reference but (a) the source Slice file does not include this type and (b) there is no cs:identifier or
             // other identifier renaming.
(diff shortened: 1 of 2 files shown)


externl:
> We no long support linking to parameters with @link. In the future we will support @p tags for this.

I assume Doxygen has this behavior?

cpp/src/slice2cs/Gen.cpp
```diff
@@ -155,8 +155,21 @@ namespace
...
+            auto pos = targetS.find("::");
+            while (pos != string::npos)
+            {
+                targetS.replace(pos, 2, ".");
+                pos = targetS.find("::", pos);
+            }
+            // Replace any '#' scope separators with '.'s.
             replace(targetS.begin(), targetS.end(), '#', '.');
+            // Remove any leading scope separators.
+            if (targetS.find(".") == 0)
```
InsertCreativityHere, line 169:
```suggestion
            if (targetS.find('.') == 0)
```

InsertCreativityHere:
Yep!

> Note that 1 and 2 match the behavior of Doxygen.
> So this PR brings us closer to Doxygen.

PR #3787, opened 2025-03-26
Simplify python reference count handling

### LINKED ISSUE #3782: Use Py_NewRef to simplify reference management
```
PyObject*
IcePy::ValueFactoryManager::getObject() const
{
    PyObject* obj = _self.get();
    Py_INCREF(obj);
    return obj;
}
```

This can be rewritten as:

```
PyObject*
IcePy::ValueFactoryManager::getObject() const
{
    return Py_NewRef(_self.geT());
}
```

This:

```
IcePy::CustomValueFactory::CustomValueFactory(PyObject* valueFactory) :
    _valueFactory(valueFactory)
{
    Py_INCREF(_valueFactory);
    assert(_valueFactory != Py_None); // This should always be present.
}
```

Should be:

```
IcePy::CustomValueFactory::CustomValueFactory(PyObject* valueFactory) :
    _valueFactory(Py_NewRef(valueFactory))
{
    assert(_valueFactory != Py_None); // This should always be present.
}
```

And so on. See https://docs.python.org/3/c-api/refcounting.html#c.Py_NewRef

pepone:
Fix #3782

### DIFF
--- a/python/modules/IcePy/Connection.cpp
+++ b/python/modules/IcePy/Connection.cpp
@@ -68,11 +68,7 @@ namespace IcePy
     class CloseCallbackWrapper final
     {
     public:
-        CloseCallbackWrapper(PyObject* cb, PyObject* con) : _cb(cb), _con(con)
-        {
-            Py_INCREF(cb);
-            Py_INCREF(con);
-        }
+        CloseCallbackWrapper(PyObject* cb, PyObject* con) : _cb(Py_NewRef(cb)), _con(Py_NewRef(con)) {}
 
         ~CloseCallbackWrapper()
         {
@@ -225,10 +221,9 @@ connectionClose(ConnectionObject* self, PyObject* /* args */)
     // Call Ice.Future.__init__
     type->tp_init(future.get(), emptyArgs.get(), nullptr);
 
-    // Create a strong reference to prevent premature release of the future object. The reference will be released by
+    // Create a new reference to prevent premature release of the future object. The reference will be released by
     // either the success or exception callback, depending on the outcome of the close operation.
-    PyObject* futureObject = future.get();
-    Py_INCREF(futureObject);
+    PyObject* futureObject = Py_NewRef(future.get());
 
     try
     {
@@ -261,9 +256,7 @@ connectionClose(ConnectionObject* self, PyObject* /* args */)
         return nullptr;
     }
 
-    // Release the reference to the future object and let the caller take ownership.
-    // The callbacks maintain their own strong reference to the future.
-    return future.release();
+    return IcePy::wrapFuture(*self->communicator, future.get());
 }
 
 extern "C" PyObject*
--- a/python/modules/IcePy/Executor.cpp
+++ b/python/modules/IcePy/Executor.cpp
@@ -74,14 +74,12 @@ IcePy::initExecutor(PyObject* module)
     return true;
 }
 
-IcePy::Executor::Executor(PyObject* executor) : _executor(executor)
+IcePy::Executor::Executor(PyObject* executor) : _executor(Py_NewRef(executor))
 {
     if (!PyCallable_Check(executor))
     {
         throw Ice::InitializationException(__FILE__, __LINE__, "executor must be a callable");
     }
-
-    Py_INCREF(executor);
 }
 
 void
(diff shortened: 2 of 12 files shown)


python/modules/IcePy/Util.cpp
```diff
@@ -271,8 +271,8 @@ IcePy::PyException::PyException() { ex = PyErr_GetRaisedException(); }
 
 IcePy::PyException::PyException(PyObject* raisedException)
 {
-    Py_XINCREF(raisedException);
```
pepone, line 274: I added the assert in case `Py_XINCREF` was used on purpose, for example we pass nullptr. By looking at the code seems we do not.

python/modules/IcePy/Util.cpp
```diff
@@ -885,10 +881,9 @@ IcePy::callMethod(PyObject* method, PyObject* arg1, PyObject* arg2)
         {
             return nullptr;
         }
-        Py_XINCREF(arg1);
-        PyTuple_SET_ITEM(args.get(), 0, arg1);
-        Py_XINCREF(arg2);
```
pepone, line 890: Here Py_XINCREF does't make sense because you cannot pass nullptr to PyTuple_SET_ITEM.

python/modules/IcePy/Util.cpp
```diff
@@ -897,21 +892,12 @@ IcePy::callMethod(PyObject* method, PyObject* arg1, PyObject* arg2)
         {
             return nullptr;
         }
-        Py_XINCREF(arg1);
-        PyTuple_SET_ITEM(args.get(), 0, arg1);
-    }
-    else if (arg2)
```
pepone, line 903: this handling of arg2 was dead code.

python/modules/IcePy/Proxy.cpp
```diff
@@ -905,19 +904,15 @@ proxyIceIsPreferSecure(ProxyObject* self, PyObject* /*args*/)
 {
     assert(self->proxy);
 
-    PyObject* b;
     try
     {
-        b = (*self->proxy)->ice_isPreferSecure() ? Py_True : Py_False;
+        return (*self->proxy)->ice_isPreferSecure() ? Py_True : Py_False;
```
pepone, line 909: Just as a remember True, False, None are immortal objects in Python 3.12, meaning we don't need to handling its reference count.

python/modules/IcePy/Connection.cpp
```diff
@@ -261,9 +256,7 @@ connectionClose(ConnectionObject* self, PyObject* /* args */)
         return nullptr;
     }
 
-    // Release the reference to the future object and let the caller take ownership.
-    // The callbacks maintain their own strong reference to the future.
-    return future.release();
+    return IcePy::wrapFuture(*self->communicator, future.get());
```
pepone, line 259: This wrapFuture call was missing in the previous PR.

python/modules/IcePy/Executor.cpp
```diff
@@ -74,14 +74,12 @@ IcePy::initExecutor(PyObject* module)
     return true;
 }
 
-IcePy::Executor::Executor(PyObject* executor) : _executor(executor)
+IcePy::Executor::Executor(PyObject* executor) : _executor(Py_NewRef(executor))
 {
     if (!PyCallable_Check(executor))
     {
         throw Ice::InitializationException(__FILE__, __LINE__, "executor must be a callable");
```
bernardnormier, line 81: Do you need to release _executor in this case? The destructor of Executor won't run if the ctor throws.

python/modules/IcePy/Operation.cpp
```diff
@@ -844,8 +843,7 @@ Operation::marshalResult(Ice::OutputStream& os, PyObject* result)
     PyObjectHandle t;
     if (numResults > 1)
     {
-        Py_XINCREF(result);
```
bernardnormier, line 847: The XINCREF was not correct - result is never null?

bernardnormier (approved):
Looks good.

python/modules/IcePy/Executor.cpp
pepone: I don't think so. `_executor` is a PyObjectHandle, its destructor should still be called and it will release the reference.

externl (approved):
Looks good!

PR #3793, opened 2025-03-27
Update doc-comments in Current, ObjectAdapter, ServantLocator

bernardnormier:


cpp/include/Ice/Current.h
```diff
@@ -16,35 +16,35 @@ namespace Ice
...
-        /// The request id unless oneway (0).
+        /// The request ID. `0` means the request is one-way.
         int requestId;
 
-        /// The encoding version used to encode the input and output parameters.
+        /// The Ice encoding version used to marshal the payload of the request.
         EncodingVersion encoding;
     };
 
+    /// @private
```
bernardnormier, line 47: Maybe we should also mark it deprecated?

We use it inside IceGrid and Glacier2 (could fix that) and we used to show it in the demos. 

Obviously, it's a big pain to code your dispatch impl defensively for the rare situation where some local code calls your function with an empty current. Better never do it and never show it as a reasonable thing to do.

cpp/include/Ice/ObjectAdapter.h
```diff
@@ -33,59 +36,49 @@ namespace Ice
...
         virtual void activate() = 0;
 
-        /// Temporarily hold receiving and dispatching requests. The object adapter can be reactivated with the
-        /// {@link #activate} operation. <p class="Note"> Holding is not immediate, i.e., after {@link #hold} returns,
-        /// the object adapter might still be active for some time. You can use {@link #waitForHold} to wait until
-        /// holding is complete.
-        /// @see #activate
-        /// @see #deactivate
-        /// @see #waitForHold
+        /// Temporarily holds receiving and dispatching requests. The object adapter can be reactivated with #activate.
```
bernardnormier, line 48: Should we mark hold/waitForHold as deprecated?

We keep them primarily for the tests that use them. We determined there is no reasonable use-cases for these operations; in particular, you can't use them to implement flow-control.

cpp/include/Ice/ObjectAdapter.h
```diff
@@ -33,59 +36,49 @@ namespace Ice
...
+        /// @remark Holding is not immediate, i.e., after #hold}returns, the object adapter might still be active for
```
pepone, line 49:
```suggestion
        /// @remark Holding is not immediate, i.e., after #hold returns, the object adapter might still be active for
```

cpp/include/Ice/ObjectAdapter.h
pepone: We should mention in the remarks that it is only provided for backwards compatibility and it should not be used in new code.

cpp/include/Ice/ObjectAdapter.h
```diff
@@ -94,288 +87,211 @@ namespace Ice
...
+        ///    the category contained in the identity.
+        ///  - If this fails, the object adapter tries to find a servant locator for the category component of the
+        ///    identity. If there is no such servant locator, the object adapter tries to find a servant locator for the
+        ///    empty category.
+        ///    - If a servant locator is found, the object adapter tries to find a servant using this servant locator.
+        ///  - If all the previous steps fail, the object adapter gives up and the caller receives an
+        ///    ObjectNotExistException or a FacetNotExistException.
+        /// @param servant The default servant to add.
+        /// @param category The category for which the default servant is registered. The empty category means it
+        /// handles all categories.
```
pepone, line 158: What about mentioning collocated transport would not consider default servants when checking if an object is local.

pepone (approved):

cpp/include/Ice/ObjectAdapter.h
externl, line 48: hold and waitForHold need new doc comments :)

cpp/include/Ice/ObjectAdapter.h
```diff
@@ -94,288 +87,211 @@ namespace Ice
...
         template<typename Prx = ObjectPrx, std::enable_if_t<std::is_base_of_v<ObjectPrx, Prx>, bool> = true>
         Prx addFacet(ObjectPtr servant, Identity id, std::string facet)
         {
             return uncheckedCast<Prx>(_addFacet(std::move(servant), std::move(id), std::move(facet)));
         }
 
-        /// Add a servant to this object adapter's Active Servant Map, using an automatically generated UUID as its
-        /// identity. Note that the generated UUID identity can be accessed using the proxy's
-        /// <code>ice_getIdentity</code> operation.
+        /// Adds a servant to this object adapter's Active Servant Map, using an automatically generated UUID as its
```
externl, line 119:
```suggestion
        /// Adds a servant to this object adapter's Active Servant Map (ASM), using an automatically generated UUID as its
```

cpp/include/Ice/ObjectAdapter.h
```diff
@@ -94,288 +87,211 @@ namespace Ice
...
-        Prx addWithUUID(ObjectPtr servant)
+        [[nodiscard]] Prx addWithUUID(ObjectPtr servant)
         {
             return uncheckedCast<Prx>(_addWithUUID(std::move(servant)));
         }
 
-        /// Like {@link #addWithUUID}, but with a facet. Calling <code>addWithUUID(servant)</code> is equivalent to
-        /// calling
-        /// {@link #addFacetWithUUID} with an empty facet.
+        /// Adds a servant to this object adapter's Active Servant Map, using an automatically generated UUID as its
```
externl, line 130:
```suggestion
        /// Adds a servant to this object adapter's Active Servant Map (ASM), using an automatically generated UUID as its
```

externl (approved):

cpp/include/Ice/ObjectAdapter.h
bernardnormier: Fixed.

cpp/include/Ice/Current.h
bernardnormier: Will fix in a follow-up PR.

cpp/include/Ice/ObjectAdapter.h
bernardnormier: It's actually more complicated and hard to explain succinctly.

See https://github.com/zeroc-ice/ice/blob/80c9e6baac79b168a25f5193a5175359cedc37de/cpp/src/Ice/ObjectAdapterI.cpp#L650

PR #3797, opened 2025-03-28
Update C++ doc-comments for Object, IncomingRequest, OutgoingResponse and more

bernardnormier:


externl (approved):

cpp/include/Ice/Format.h
```diff
@@ -7,12 +7,14 @@
 
 namespace Ice
 {
-    /// Describes the possible formats for classes and exceptions.
+    /// Specifies how to marshal classes and exceptions with the Ice 1.1 encoding.
```
pepone, line 10: What about `Specifies the format for marshaling classes...`

pepone (approved):

cpp/include/Ice/Format.h
bernardnormier: Fixed as suggested.

PR #3799, opened 2025-03-29
Remove {@link ...} syntax from C++ doc-comments

bernardnormier:
This undocumented syntax breaks the VScode doxygen support.

### DIFF
--- a/cpp/include/Ice/LocalExceptions.h
+++ b/cpp/include/Ice/LocalExceptions.h
@@ -348,8 +348,7 @@ namespace Ice
     //
 
     /// This exception is raised if a system error occurred in the server or client process. There are many possible
-    /// causes for such a system exception. For details on the cause, {@link SyscallException#error} should be
-    /// inspected.
+    /// causes for such a system exception. For details on the cause, SyscallException::error should be inspected.
     /// @headerfile Ice/Ice.h
     class ICE_API SyscallException : public LocalException
     {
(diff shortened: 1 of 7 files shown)


slice/IceBox/ServiceManager.ice
```diff
@@ -63,7 +63,7 @@ module IceBox
         void stopService(string service)
             throws AlreadyStoppedException, NoSuchServiceException;
 
-        /// Registers a new observer with this ServiceManager.
+        /// Registers a new observer with this service manager.
```
bernardnormier, line 66: ServiceManager gets autolinked to C++ Ice::ServiceManager that we no longer include in the doc since the user doesn't implement this generated interface.

And, once we fix autolink to link to proxies, it won't make sense to have a ServiceManagerPrx here.

cpp/include/Ice/LocalExceptions.h
```diff
@@ -549,7 +547,7 @@ namespace Ice
         std::shared_ptr<std::string> _id;
     };
 
-    /// This exception is raised if the {@link Communicator} has been destroyed.
+    /// This exception is raised if the Communicator has been destroyed.
```
bernardnormier, line 550: This PR is focuses on `@link`. We need more updates to these doc-comments, in separate PRs.

pepone (approved):

externl (approved):

PR #3801, opened 2025-03-31
Add checkedCastAsync to python

pepone:
This PR adds a `checkedCastAsync` implementation for Python. It also replaces the `facetOrContext` argument in checkedCast to be `facet`, there is already a optional `context` argument.


python/modules/IcePy/Proxy.cpp
```diff
@@ -1620,18 +1621,93 @@ checkedCastImpl(ProxyObject* p, const string& id, PyObject* facet, PyObject* ctx
     return Py_None;
 }
 
+static PyObject*
```
bernardnormier, line 1624: I don't understand why we implement Python's checkedCast (and now checkCastAsync) in C++.

Why not a simple wrapper (in pure Python) over the Python's ice_isAAsync?

python/python/Ice/Object.py
```diff
@@ -37,7 +37,7 @@ def ice_ping(self, current):
         """
         pass
 
-    def ice_ids(self, current=None):
+    def ice_ids(self, current):
```
bernardnormier, line 40: We should eliminate all these current=None!

python/modules/IcePy/Proxy.cpp
pepone: Yes, it might be simpler to rework this and define both in Python.

python/modules/IcePy/Proxy.cpp
```diff
@@ -1561,7 +1562,7 @@ proxyIceFlushBatchRequestsAsync(ProxyObject* self, PyObject* /*args*/, PyObject*
         return nullptr;
     }
     callback->setFuture(future.get());
-    return future.release();
+    return IcePy::wrapFuture(*self->communicator, future.get());
```
pepone, line 1565: This is an unrelated fix. Code returning an future needs to call IcePy::wrapFuture to ensure the event loop adapter is used.

python/modules/IcePy/Proxy.cpp
```diff
@@ -1576,107 +1577,6 @@ proxyIceInvokeAsync(ProxyObject* self, PyObject* args, PyObject* /*kwds*/)
     return iceInvokeAsync(reinterpret_cast<PyObject*>(self), args);
 }
 
-static PyObject*
```
pepone, line 1579: This is now implemented in python code see ObjectPrx.py

python/modules/IcePy/Proxy.cpp
pepone: Reworked to move checkedCast implementation to ObjectPrx.py

python/modules/IcePy/Proxy.cpp
```diff
@@ -2062,18 +1888,10 @@ static PyMethodDef ProxyMethods[] = {
      reinterpret_cast<PyCFunction>(proxyIceInvokeAsync),
      METH_VARARGS | METH_KEYWORDS,
      PyDoc_STR("ice_invokeAsync(op, mode, inParams[, context]) -> Ice.Future")},
-    {"ice_checkedCast",
-     reinterpret_cast<PyCFunction>(proxyIceCheckedCast),
-     METH_VARARGS | METH_CLASS,
-     PyDoc_STR("ice_checkedCast(proxy, id[, facetOrContext[, context]]) -> proxy")},
```
bernardnormier, line 2068: Why do we implement uncheckedCast in C++? (If fix needed, can wait until a follow-up PR).

python/python/Ice/ObjectPrx.py
```diff
@@ -1,11 +1,126 @@
...
+        -------
+        ObjectPrx|None
+            A proxy with the requested type, or None if the target object does not support the requested type.
+        """
+        return checkedCastAsync(ObjectPrx, proxy, facet, context)
+
+    @staticmethod
+    def ice_staticId():
+        """
+        Gets the Slice type id of the interface or class associated with this proxy class.
```
externl, line 115:
```suggestion
        Gets the Slice type ID of the interface or class associated with this proxy class.
```

python/python/Ice/ObjectPrx.py
```diff
@@ -1,11 +1,126 @@
...
+
+        Returns
+        -------
+        str
+            The type id, "::Ice::Object".
```
externl, line 120:
```suggestion
            The type ID, "::Ice::Object".
```

python/python/Ice/ObjectPrx.py
```diff
@@ -38,6 +153,27 @@ def ice_isA(self, id, context=None):
...
+        Parameters
+        ----------
+        id : str
+            The type ID of the Slice interface to test against.
+        context : dict, optional
+            The context dictionary for the invocation.
+
+        Returns
+        -------
+        bool
```
externl, line 169: Not sure how this is usually documented in Python but doesn't this really return a future that will contain a bool once it completes?

externl (approved):

bernardnormier (approved):
Looks, much simpler.

python/python/Ice/ObjectPrx.py
bernardnormier: Please remove 'class'. It's the Slice type ID of the interface.

python/modules/IcePy/Proxy.cpp
pepone: Currently we cannot create a new proxy from another proxy in Python side. ObjectPrx::createProxy is not exposed to Python.

PR #3802, opened 2025-03-31
Adding Checkstyle and OpenRewrite, removing Spotless (Java)

ayoitslilith:
This PR swaps out Spotless on Gradle for Java for Checkstyle and OpenRewrite, also on Gradle. Unlike Spotless, Checkstyle checks for both formatting and linting with `gradlew checkstyle`. Additionally, OpenRewrite will fix the detected issues with `gradlew rewriteRun`. 

As of the moment, OpenRewrite warns of several issues to be deprecated in Gradle 9.0 (but they do not cause issues right now). Additionally, it has failed to fix two instances regarding variants. To solve this, one variant option must be selected in the configuration for the issues. I did not select a specific variant as to wait and see if anyone desires a specific resolution. Please see [this file](https://drive.google.com/file/d/17tdlam-L5lqYnX2rEdWl3qdr6tEkWk90/view?usp=sharing) for more details.


bernardnormier:
> Additionally, it has failed to fix two instances regarding variants.

This does not appear to be in the Ice for Java source code. It references caffeine.

bernardnormier (commented):
It's not clear to me if CheckStyle and Rewrite are now a new target (with which name(s)?) or if they run automatically as part of a regular build (not desirable).

java/build.gradle
```diff
@@ -36,6 +45,11 @@ subprojects {
         options.encoding = "UTF-8"
         options.deprecation = true
     }
+
+    //makes checkstyle wait for rewrite. hopefully
```
pepone, line 49: Does it wait or not?

java/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,453 @@
...
+      <property name="id" value="SeparatorWrapArrayDeclarator"/>
+      <property name="tokens" value="ARRAY_DECLARATOR"/>
+      <property name="option" value="EOL"/>
+    </module>
+    <module name="SeparatorWrap">
+      <property name="id" value="SeparatorWrapMethodRef"/>
+      <property name="tokens" value="METHOD_REF"/>
+      <property name="option" value="nl"/>
+    </module>
+    <!--<module name="PackageName">
```
pepone, line 220: For modules we don't plant to fix, like PackageName, we should remove them. It is fine to keep some of these if we plan to enable them in a follow up PR.

pepone (commented):
When building with this branch there is lot of warnings reported:

```
Checkstyle files with violations: 1301
Checkstyle violations by severity: [warning:85041]
```

It would be nicer to disable the ones we are not fixing in this PR.

java/config/checkstyle/suppressions.xml
```diff
@@ -0,0 +1,26 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+<!DOCTYPE suppressions PUBLIC
+  "-//Checkstyle//DTD SuppressionFilter Configuration 1.2//EN"
+  "https://checkstyle.org/dtds/suppressions_1_2.dtd">
+ <suppressions>
```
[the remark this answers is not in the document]
ayoitslilith: I don't know why, but I had it in my head that this file was necessary. I'm removing this file in the next commit.

InsertCreativityHere (commented):
Skimming through it, I think this looks pretty nice. The indentation and style are definitely improvements, and the `import`s/`final`s are good too.


But there are definitely some types which we'd prefer to 'blacklist' from using imports.
Anytime there are 2 classes which share the same name. Explicitly writing out which package it's in makes it more readable, and eliminates the chance of bugs being introduced by rewrite rewriting them. This is not an exhaustive list, but here's the big ones off-the-top-of-my-head:
```
Object
InputStream
OutputStream
ObjectInputStream
ObjectOutputStream
Connection
SSLEngine
etc.
```


Also, and this is a very minor nitpick,
`checkStyle`+`rewrite` splits empty things on separate lines, when they used to be on a single line:
```diff
- public Ice_invokeResult() {}
+ public Ice_invokeResult() {
+ }
```
I think it was nicer (and more Java-y) when both braces were on the same line.
Feel free to ignore this though. Again. It's incredibly minor.

.github/workflows/java.yml
```diff
@@ -0,0 +1,20 @@
...
+    gradle:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Setup Java
+        uses: ./.github/actions/setup-java
+
+      - name: Run Gradle with Checkstyle and RewriteDryRun
+        timeout-minutes: 10
+        working-directory: java
+        uses: ./gradlew build
```
pepone, line 20: We already run `./gradlew build` in CI, is this workflow supposed to run a different target? We don't need to run `build` again. We also need to build slice2java before this can work.

java/build.gradle
```diff
@@ -77,3 +95,48 @@ task alljavadoc(type: Javadoc) {
...
+    exclusion(
+        '**/*.gradle',
+        'com/zeroc/Ice/PropertyNames.java',
+        '**/generated/**/*',
+        '**/*.kts',
+        'java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java',
+        'java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InputStream.java',
+        'java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/LiveDeployment/ShowIceLogDialog.java',
+        'java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/LiveDeployment/ShowLogFileDialog.java',
+        'java/test/src/main/java/test/Ice/operations/Twoways.java',
```
pepone, line 125: Why are this files excluded? You should add a comment.

ayoitslilith:
To summarize the changes made in this PR:
- Checkstyle has been implemented for reporting on desired changes for linting and formatting all non-generated files in the Java Ice project. It itself doesn't make any changes, but serves to highlight changes to be made and provide some style parameters for OpenRewrite. You can run a Checkstyle test with `gradlew checkstyle`, though this runs with a `rewriteDryRun` beforehand due to how they work in the build system. Checkstyle rules can be found in `./java/config/checkstyle/checkstyle.xml`.
- OpenRewrite has also been implemented and applied for linting and formatting with slightly less files in the Java Ice project. OpenRewrite makes changes when run with `gradlew rewriteRun` but will always log changes that it wants to make when used as `rewriteDryRun` or `rewriteRun`. `rewriteDryRun` fails by default if it detects changes that it would make if run normally, but if you want to avert this behavior, you can it with `-PdryRunFailSkip`. Please note that OpenRewrite does **not** have every rule that Checkstyle has, and additionally has some rules that Checkstyle lacks. OpenRewrite rules can be found in `./java/build.gradle` under the `rewrite` block and additional options for the rules that are not properly integrated with Checkstyle can be found in `./java/rewrite.yml`.
- Multiple rules changes have been made to the rulesets of both Checkstyle and OpenRewrite to either suit Ice's pre-existing formatting or to put off certain formatting changes for other PRs. Checkstyle rules that fall outside of the scope of this PR are commented out in `./java/config/checkstyle/checkstyle.xml`.
- The files in the Java project have been formatted by both OpenRewrite and partially myself while looking at the Checkstyle reports. The most notable manual change I have made was to reduce the use of star imports and fix compilation errors caused by OpenRewrite's modification to certain files. The overall changes have reduced the Checkstyle problem count from the tens of thousands to under one hundred, though a substantial amount of this was from disabling rules that were outside of the scope of this PR.
- ~Spotless is dead, long live Checkstyle and OpenRewrite!~ Spotless has been removed from our implementation of the Gradle build system.

java/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,443 @@
...
+                                     or preceding-sibling::*[last()][self::LCURLY]]"/>
+    </module>
+    <module name="WhitespaceAfter">
+      <property name="tokens"
+               value="COMMA, SEMI, TYPECAST, LITERAL_IF, LITERAL_ELSE, LITERAL_RETURN,
+                    LITERAL_WHILE, LITERAL_DO, LITERAL_FOR, LITERAL_FINALLY, DO_WHILE, ELLIPSIS,
+                    LITERAL_SWITCH, LITERAL_SYNCHRONIZED, LITERAL_TRY, LITERAL_CATCH, LAMBDA,
+                    LITERAL_YIELD, LITERAL_CASE, LITERAL_WHEN"/>
+    </module>
+    <!--<module name="WhitespaceAround">
```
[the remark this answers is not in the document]
ayoitslilith: Oops - I meant to Austin gave his review, but ended up forgetting. I'll make a commit for that and the comments.

.github/workflows/java.yml
ayoitslilith: This is left over from when I had a custom argument in this build command - this will be removed in a coming commit.

InsertCreativityHere: Yes, we probably don't need this file.
We already do `./gradlew build` (indirectly) with 'ci.yml'.

Also, if you look at the GitHub workflows, it says there's an error with this file as well:
(java.yml): https://github.com/zeroc-ice/ice/actions/runs/14384943695

java/config/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,454 @@
...
```
pepone, line 144: Are we planing to enable any of the rules in `WhitespaceAround`? If yes add a comment otherwise remove.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/OutgoingAsync.java
```diff
@@ -97,13 +99,13 @@ public T waitForResponseOrUserEx() throws UserException {
...
+        } catch (ExecutionException ee) {
             try {
                 throw ee.getCause().fillInStackTrace();
             } catch (RuntimeException ex) // Includes LocalException
-            {
-                throw ex;
-            } catch (UserException ex) {
+                {
+                    throw ex;
+                } catch (UserException ex) {
```
InsertCreativityHere, line 108: It added some indentation here which looks bogus to me.
Maybe the comment at the end of this line is tripping it up?

Maybe it you replaced the `catch` line with this, it will not add this suspicious indentation:
```diff
-} catch (RuntimeException ex) // Includes LocalException
+} catch (RuntimeException ex /* Includes LocalException */)
```

InsertCreativityHere (approved):
I think that this definitely constitutes an improvement. There are a couple strange formatting choices it made, but we can always tweak this in the future.

The only 'real' thing I can think of is that it would be nice to keep our qualification of 'ambiguous' types like `Object`, `InputStream`, `Process`, etc. whereas right now, `checkstyle` is just removing them (converting `com.zeroc.Ice.Object` to `Object` I mean).

It's totally possible that there's no way to prevent this replacement. But it'd be nice if there was was all.

PR #3804, opened 2025-03-31
Add Support for `ruby:identifier` Metadata

InsertCreativityHere:
This PR adds support for `ruby:identifier`. It is the last and final language in this saga.

Unlike the other languages, `slice2rb` automaticallys lower/upper-case identifiers as is necessary.

So, instead of calling `p->mappedName()` directly, I added a new `mappedName` helper function.
If it detects `ruby:identifier`, it uses the provided identifier as-is.
Otherwise, it returns `p`'s case-corrected Slice name.

Like Python and PHP, I also had to change some of the IceRuby C++ code.
`IceRuby::OperationI` was calling `fixIdent` from `slice2rb` to convert an operation's slice name to it's mapped name.
This no longer works, so we have to pass in the mapped operation name.


cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -54,16 +79,6 @@ namespace Slice::Ruby
         void visitConst(const ConstPtr&) final;
 
     private:
```
InsertCreativityHere, line 81: `getSymbol`, `registerName`, and `writeHash` were dead functions.

cpp/src/slice2rb/RubyUtil.cpp
InsertCreativityHere, line 1: I introduced some helper functions to reduce repitition and centralize the meat of the changes:
`getMappedName(p, style)` <= `fixIdent(p->name(), style)` (style defaults to `IdentNormal` ie. don't change the case).
`getMetaTypeName(p)` <= `"T_" + fixIdent(p->name(), IdentToUpper)`
`getMetaTypeReference(p)` <= `getAbsolute(p, IdentToUpper, "T_")`

`getAbsolute` was only called with the `IdentUpper` style.
So now it's no longer passed in as a parameter.

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -560,23 +477,20 @@ Slice::Ruby::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
         {
             _out << "nil";
         }
-        _out << ", [";
-        ExceptionList exceptions = op->throws();
-        for (auto u = exceptions.begin(); u != exceptions.end(); ++u)
+        _out << ", ";
+        _out.spar("[");
```
InsertCreativityHere, line 481: Between `spar` and `epar`, we automatically emit `", "` between elements.

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -875,15 +775,11 @@ Slice::Ruby::CodeVisitor::visitEnum(const EnumPtr& p)
     //
     // from_int
     //
-    {
-        _out << sp << nl << "def " << name << ".from_int(val)";
-        ostringstream sz;
-        sz << enumerators.size() - 1;
```
InsertCreativityHere, line 881: No idea what this `ostringstream` was for.
Or why this code is in it's own block.
Seemed pointless though.

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -936,25 +832,19 @@ Slice::Ruby::CodeVisitor::visitEnum(const EnumPtr& p)
     // Constant for each enumerator.
     //
     _out << sp;
-    int i = 0;
-    for (auto q = enumerators.begin(); q != enumerators.end(); ++q, ++i)
+    EnumeratorList enumerators = p->enumerators();
+    for (const auto& enumerator : enumerators)
     {
-        ostringstream idx;
-        idx << i;
```
InsertCreativityHere, line 943: Another seemingly useless `ostringstream`.

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -1101,81 +988,67 @@ Slice::Ruby::CodeVisitor::getInitializer(const DataMemberPtr& m)
...
+        return getAbsolute(st) + ".new";
     }
 
     return "nil";
 }
 
-void
-Slice::Ruby::CodeVisitor::writeHash(const string& name)
-{
-    _out << nl << "_h = 5 * _h + " << name << ".hash";
```
InsertCreativityHere, line 1123: This function was inline on line 635.

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -1253,83 +1124,70 @@ Slice::Ruby::generate(const UnitPtr& un, bool all, const vector<string>& include
     }
 
     CodeVisitor codeVisitor(out);
-    un->visit(&codeVisitor);
+    unit->visit(&codeVisitor);
 
     out << nl; // Trailing newline.
 }
 
 string
```
InsertCreativityHere, line 1132: Everything below this comment was largely rewritten.

ruby/src/IceRuby/Util.cpp
```diff
@@ -544,7 +544,11 @@ namespace
     template<size_t N>
     VALUE createRubyException(const char* typeId, std::array<VALUE, N> args, bool fallbackToLocalException = false)
     {
+        // Convert the exception's typeId to its mapped Ruby type by removing the leading "::".
+        // This function should only ever be called on a specific list of Ice local exceptions.
         string className = string{typeId}.substr(2);
+        assert(className.starts_with("Ice::"))
+
```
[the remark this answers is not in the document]
pepone: > // This function should only ever be called on a specific list of Ice local exceptions.

Maybe is clear to mention it is safe for local exception as long as they are not using "ruby:identifier"

pepone (approved):

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -1253,83 +1124,70 @@ Slice::Ruby::generate(const UnitPtr& un, bool all, const vector<string>& include
...
     {
-        //
-        // We assume all intermediate names must be upper-case (i.e., they represent
-        // the names of modules or classes).
-        //
-        result << "::" << fixIdent(ids[i], IdentToUpper);
+        case Slice::Ruby::IdentNormal:
+            break;
+        case Slice::Ruby::IdentToUpper:
+            ident[0] = (char)toupper(ident[0]);
```
externl, line 1148: We prefer static_cast over c casts.

externl (approved):
Looks good!

cpp/src/slice2rb/RubyUtil.cpp
InsertCreativityHere: Fixed.

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -723,7 +632,8 @@ Slice::Ruby::CodeVisitor::visitStructStart(const StructPtr& p)
     _out << nl << "_h = 0";
     for (const auto& member : members)
     {
-        writeHash("@" + fixIdent(member->name(), IdentNormal));
+        const string memberName = "@" + getMappedName(member);
```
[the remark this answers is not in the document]
InsertCreativityHere: \*shrugs\*. Couldn't tell you. Feel free to open an issue, but I'm definitely not the right person to ask about the intent behind Ruby mapping!

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -511,36 +432,32 @@ Slice::Ruby::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
...
-                writeType((*t)->type());
-                _out << ", " << ((*t)->optional() ? "true" : "false") << ", " << ((*t)->optional() ? (*t)->tag() : 0)
-                     << ']';
-                ++count;
+                _out << ", ";
             }
+            isFirst = false;
+            _out << '[';
+            writeType(param->type());
+            _out << ", " << (param->optional() ? "true" : "false") << ", " << (param->optional() ? param->tag() : 0)
```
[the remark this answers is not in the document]
InsertCreativityHere: We could, but here we aren't iterating over a list of unknown length.
We know exactly what we're writing `type, option, tagId`, so it's actually shorter to just write the `,`s inline ourselves.

`spar` and `epar` are most useful when we're not sure whether we're at the start of a list or not. For instance, like if we're in a for-loop.

ruby/src/IceRuby/Util.cpp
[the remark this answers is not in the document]
InsertCreativityHere: I updated the comment to:
`// This function should only ever be called on Ice local exceptions which don't use 'ruby:identifier'.`

cpp/src/Slice/Grammar.y
```diff
@@ -1259,12 +1259,8 @@ operation_preamble
         {
             currentUnit->pushContainer(op);
             currentUnit->error("keyword '" + name + "' cannot be used as operation name");
-            $$ = op; // Dummy
-        }
-        else
-        {
-            return 0;
         }
+        $$ = op; // Dummy
```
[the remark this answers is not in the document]
InsertCreativityHere: I updated the other 3 places in the grammar where we did this.
Now they're all simpler and consistent.

PR #3805, opened 2025-03-31
Update doc for exceptions (C++)

bernardnormier:
This PR updates the documentation for C++ exceptions.

### DIFF
--- a/cpp/include/DataStorm/Node.h
+++ b/cpp/include/DataStorm/Node.h
@@ -10,10 +10,7 @@ namespace DataStorm
 {
     template<typename, typename, typename> class Topic;
 
-    /// The exception NodeShutdownException. It's raised by methods which might block waiting for a condition to
-    /// occur and after the node has been shutdown. It informs the application that the condition won't occur
-    /// because the DataStorm node is being shutdown and will disconnect from other nodes.
-    ///
+    /// The exception that is thrown when a blocking operation is interrupted by the shutdown of the node.
     /// @headerfile DataStorm/DataStorm.h
     class DATASTORM_API NodeShutdownException final : public std::exception
     {
--- a/cpp/include/Ice/LocalExceptions.h
+++ b/cpp/include/Ice/LocalExceptions.h
@@ -293,7 +293,7 @@ namespace Ice
     // Timeout exceptions
     //
 
-    /// The exception that is thrown to report a timeout.
+    /// The exception that is thrown when a timeout occurs. This is the base class for all timeout exceptions.
     /// @headerfile Ice/Ice.h
     class ICE_API TimeoutException : public LocalException
     {
@@ -628,7 +628,7 @@ namespace Ice
         [[nodiscard]] const char* ice_id() const noexcept final;
     };
 
-    /// The exception that is thrown when attempting to change the connection properties of a fixed proxy.
+    /// The exception that is thrown when attempting to change a connection-related property on a fixed proxy.
     /// @headerfile Ice/Ice.h
     class ICE_API FixedProxyException final : public LocalException
     {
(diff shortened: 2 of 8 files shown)


cpp/include/DataStorm/Node.h
```diff
@@ -10,10 +10,7 @@ namespace DataStorm
 {
     template<typename, typename, typename> class Topic;
 
-    /// The exception NodeShutdownException. It's raised by methods which might block waiting for a condition to
-    /// occur and after the node has been shutdown. It informs the application that the condition won't occur
-    /// because the DataStorm node is being shutdown and will disconnect from other nodes.
-    ///
+    /// The exception that is thrown when a blocking operation is interrupted by the shutdown of the node.
```
bernardnormier, line 13: Please check this is actually correct. The previous description was rather unclear.

cpp/include/Ice/LocalExceptions.h
```diff
@@ -626,7 +628,7 @@ namespace Ice
         [[nodiscard]] const char* ice_id() const noexcept final;
     };
 
-    /// This exception indicates that an attempt has been made to change the connection properties of a fixed proxy.
+    /// The exception that is thrown when attempting to change the connection properties of a fixed proxy.
```
pepone, line 631: is this accurate? for example:

```
ReferencePtr
IceInternal::FixedReference::changeAdapterId(string /*newAdapterId*/) const
{
    throw FixedProxyException(__FILE__, __LINE__);
}
```

That is not a connection property.

pepone (approved):

cpp/include/Ice/LocalExceptions.h
```diff
@@ -516,9 +520,7 @@ namespace Ice
     // Other leaf local exceptions in alphabetical order.
     //
 
-    /// An attempt was made to register something more than once with the Ice run time. This exception is raised if an
-    /// attempt is made to register a servant, servant locator, facet, value factory, plug-in, object adapter, object,
-    /// or user exception factory more than once for the same ID.
+    /// The exception that is thrown when you attempt to register an object more than once with the Ice runtime.
```
externl, line 523:
```suggestion
    /// The exception that is thrown when you attempt to register an object more than once with an ObjectAdapter.
```

cpp/include/Ice/LocalExceptions.h
```diff
@@ -788,7 +788,7 @@ namespace Ice
         [[nodiscard]] const char* ice_id() const noexcept final;
     };
 
-    /// This exception indicates a failure in a security subsystem, such as the IceSSL plug-in.
+    /// The exception that is thrown when a failure occurs in the security subsystem; this includes IceSSL errors.
```
externl, line 791: `;` is not correct here as its not separating two independent clauses. I would write either:

> The exception that is thrown when a failure occurs in the security subsystem, including IceSSL errors.

> The exception that is thrown when a failure occurs in the security subsystem. This includes IceSSL errors.

externl (approved):

cpp/include/Ice/LocalExceptions.h
bernardnormier: It is connection-related. How would you describe it?

bernardnormier: New version:
```
/// The exception that is thrown when attempting to change a connection-related property on a fixed proxy.
```

cpp/include/Ice/LocalExceptions.h
bernardnormier: The proposed change is not accurate.

cpp/include/Ice/LocalExceptions.h
bernardnormier: Picked second option.

PR #3807, opened 2025-04-01
Rework handling of builtin operations in Python ObjectPrx

pepone:
This PR reworks the handling of the 4 builtin operations and uncheckedCat in Python proxy implementation. They are now handled in the Python code rather than in the C++ extension. Removed the special C++ code for handing this operations and treat them as other RPCs.

### DIFF
--- a/python/python/Ice/ObjectPrx.py
+++ b/python/python/Ice/ObjectPrx.py
@@ -1,5 +1,31 @@
 
 import IcePy
+from .Object import Object
+
+def uncheckedCast(type, proxy, facet=None):
+    """
+    Downcasts a proxy without confirming the target object's type via a remote invocation.
+
+    Parameters
+    ----------
+    type : type
+        The proxy target type.
+    proxy : ObjectPrx
+        The source proxy (can be None).
+
+    facet : str, optional
+        A facet name.
+
+    Returns
+    -------
+    ObjectPrx|None
+        A proxy with the requested type and facet.
+    """
+    if proxy is None:
+        return None
+    if facet is not None:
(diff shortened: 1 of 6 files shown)


python/python/Ice/ObjectPrx.py
```diff
@@ -1,5 +1,31 @@
 
 import IcePy
+from .Object import Object
+
+def uncheckedCast(type, proxy, facet=None):
+    """
+    Downcasts a proxy without confirming the target object's type via a remote invocation.
```
bernardnormier, line 7: Maybe change it to a simpler:

```
Creates a proxy from another proxy, and optionally changes the proxy's facet. This function does not invoke on the target object.
```

python/python/Ice/ObjectPrx.py
```diff
@@ -21,13 +47,13 @@ def checkedCast(type, proxy, facet=None, context=None):
     Returns
     -------
     ObjectPrx|None
-        A proxy with the requested type, or None if the target object does not support the requested type.
+        A proxy with the requested type and facet, or None if the target object does not support the requested type.
```
bernardnormier, line 50: I'd rather leave facet out the Returns description.

bernardnormier (approved):
Looks good, but we should rework the doc-comments for checkedCast and uncheckedCast in a follow-up PR.

externl (approved):

python/python/Ice/ObjectPrx.py
pepone: fixed.

pepone:
I would let the doc comment rework for a follow up PR as this affects other mappings too.

PR #3811, opened 2025-04-01
Change default ProgramName in MATLAB

bernardnormier:
It's now `matlab-client`. Set `Ice.ProgramName` to get a different value.


pepone (approved):
Looks good!

externl (approved):

PR #3813, opened 2025-04-02
Rework handling of builtin operations in Ruby ObjectPrx

pepone:
This PR reworks the handling of the 4 builtin operations in Ruby proxies, they are now treated like regular user defined operations and they use they use the same code path.

The implementation of checked and unchecked was moved to Ruby too.

### DIFF
--- a/ruby/ruby/Ice/Proxy.rb
+++ b/ruby/ruby/Ice/Proxy.rb
@@ -12,14 +12,42 @@ module Ice
                 self::ICE_ID
             end
 
-            def checkedCast(proxy, facetOrContext=nil, context=nil)
-                ice_checkedCast(proxy, self::ICE_ID, facetOrContext, context)
+            def checkedCast(proxy, facet: nil, context: nil)
+                if proxy.nil?
+                    return nil
+                end
+
+                unless facet.nil?
+                    proxy = proxy.ice_facet(facet)
+                end
+
+                if proxy.ice_isA(self::ICE_ID, context)
+                    return self::create(proxy)
+                else
+                    return nil
+                end
+            end
+
+            def uncheckedCast(proxy, facet: nil)
+                if proxy.nil?
+                    return nil
+                end
+
+                unless facet.nil?
+                    proxy = proxy.ice_facet(facet)
+                end
+
+                self::create(proxy)
             end
         end
 
       def self.included(base)
           base.extend(ClassMethods)
       end
+    end
+
+    class ObjectPrx
+        include Proxy_mixin
     end
 
     # Proxy comparison functions.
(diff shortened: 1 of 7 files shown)


ruby/ruby/Ice/ProxyFunctions.rb
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) ZeroC, Inc.
+
+require_relative 'Proxy.rb'
+require_relative 'OperationMode.rb'
+require_relative 'BuiltinSequences.rb'
+
+module ::Ice
+    module ObjectPrx_mixin
```
pepone, line 8: We define the builtin operations here instead of in Proxy.rb to avoid circular dependencies.

ruby/ruby/Ice/Proxy.rb
```diff
@@ -12,8 +12,32 @@ def ice_staticId()
                 self::ICE_ID
             end
 
-            def checkedCast(proxy, facetOrContext=nil, context=nil)
-                ice_checkedCast(proxy, self::ICE_ID, facetOrContext, context)
+            def checkedCast(proxy, facet: nil, context: nil)
```
pepone, line 15: Switched to keyword arguments, this means that if you were using the 2nd or 3rd args you will have to update the code to use a keyword arg, see the test updates.

ruby/test/Ice/facets/AllTests.rb
```diff
@@ -40,39 +40,39 @@ def allTests(helper, communicator)
     STDOUT.flush
     obj = Ice::ObjectPrx::uncheckedCast(db)
     test(obj.ice_getFacet() == "")
-    obj = Ice::ObjectPrx::uncheckedCast(db, "facetABCD")
+    obj = Ice::ObjectPrx::uncheckedCast(db, facet: "facetABCD")
```
bernardnormier, line 43: I find the new syntax clearer.

bernardnormier (approved):
Looks good.

ruby/ruby/Ice/ProxyFunctions.rb
```diff
@@ -0,0 +1,69 @@
...
+        def ice_isA(id, context=nil)
+            return ObjectPrx_mixin::OP_ice_isA.invoke(self, [id], context)
+        end
+
+        def ice_id(context=nil)
+            ObjectPrx_mixin::OP_ice_id.invoke(self, [], context)
+        end
+    end
+
+    class ObjectPrx
```
externl, line 26: Why do we also define ObjectPrx here?

externl (approved):
Looks good.

ruby/ruby/Ice/ProxyFunctions.rb
pepone: We reopen the class to add the 4 builtin methods defined in the mixin above.

PR #3814, opened 2025-04-02
Make Communicator async disposable (C#)

bernardnormier:
This PR make Ice.Communicator async disposable.

### DIFF
--- a/csharp/test/Ice/adapterDeactivation/Client.cs
+++ b/csharp/test/Ice/adapterDeactivation/Client.cs
@@ -8,7 +8,7 @@ public class Client : TestHelper
 {
     public override async Task runAsync(string[] args)
     {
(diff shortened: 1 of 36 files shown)


csharp/test/Ice/adapterDeactivation/Client.cs
```diff
@@ -8,7 +8,7 @@ public class Client : TestHelper
 {
     public override async Task runAsync(string[] args)
     {
-        using var communicator = initialize(ref args);
+        await using var communicator = initialize(ref args);
```
bernardnormier, line 11: I updated all (or at least most) of the tests with a runAsync to DisposeAsync the communicator.

externl (approved):

pepone (approved):

PR #3815, opened 2025-04-02
Add Communicator::destroyAsync in C++

bernardnormier:
Updated to join the threads in the destructor.

### DIFF
--- a/cpp/include/Ice/Communicator.h
+++ b/cpp/include/Ice/Communicator.h
@@ -37,10 +37,8 @@ namespace Ice
         /// Destroys this communicator asynchronously.
         /// @param completed The callback to call when the destruction is complete. This callback must not throw any
         /// exception.
-        /// @remark This function starts a background thread to call @p completed unless you call this function on a
-        /// communicator that has already been destroyed, in which case @p completed is called by the current thread.
-        /// You need to be careful when calling this function, and not exit your application while this background
-        /// thread is still running.
+        /// @remark This function starts a thread to call @p completed unless you call this function on a communicator
+        /// that has already been destroyed, in which case @p completed is called by the current thread.
         /// @see #destroy
         void destroyAsync(std::function<void()> completed) noexcept;
 
--- a/cpp/include/Ice/Communicator.h
+++ b/cpp/include/Ice/Communicator.h
@@ -29,11 +29,20 @@ namespace Ice
     public:
         ~Communicator();
 
-        /// Destroys the communicator. This function calls #shutdown implicitly. Calling #destroy destroys all object
-        /// adapters, shuts down this communicator's client functionality, and cleans up memory. Subsequent calls to
-        /// #destroy are no-op.
+        /// Destroys this communicator. This function calls #shutdown implicitly. Calling #destroy destroys all object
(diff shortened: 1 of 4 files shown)


cpp/include/Ice/Communicator.h
```diff
@@ -314,6 +323,8 @@ namespace Ice
         void postToClientThreadPool(std::function<void()> call);
 
     private:
+        Communicator() = default;
```
bernardnormier, line 326: (Unrelated update). We relied on a public default constructor - not good.

cpp/src/Ice/Communicator.cpp
```diff
@@ -39,12 +39,22 @@ Ice::Communicator::flushBatchRequestsAsync(CompressBatch compress)
 void
 Ice::Communicator::destroy() noexcept
 {
-    if (_instance)
+    if (_instance) // see create implementation
```
bernardnormier, line 42: Everywhere else, this const field is always non-null.

cpp/src/Ice/Instance.cpp
```diff
@@ -1734,6 +1734,36 @@ IceInternal::Instance::destroy()
...
+    bool executeCallback = false;
+    {
+        lock_guard lock(_mutex);
+        if (_state == StateDestroyed)
+        {
+            executeCallback = true; // execute outside the lock
+        }
+        else
+        {
+            // Start a background thread that calls destroy() and then executes the callback.
```
bernardnormier, line 1751: I don't see how we could use a joinable thread here. Who would join it?

pepone: Can't we join it in the destructor?

pepone (approved):

cpp/src/Ice/Instance.cpp
bernardnormier: The thread "body" keeps a refcount on Instance, and I assume it's kept while the thread exists. It's not clear to me how this join could work.

cpp/include/Ice/Communicator.h
```diff
@@ -29,12 +29,21 @@ namespace Ice
...
 
+        /// Destroys this communicator asynchronously.
+        /// @param completed The callback to call when the destruction is complete. This callback must not throw any
+        /// exception.
+        /// @remark This function starts a background thread to call @p completed unless you call this function on a
+        /// communicator that has already been destroyed, in which case @p completed is called by the current thread.
+        /// You need to be careful when calling this function, and not exit your application while this background
+        /// thread is still running.
+        /// @see #destroy
+        void destroyAsync(std::function<void()> completed) noexcept;
```
externl, line 45: can the function be marked `noexecpt`?

externl (approved):

cpp/src/Ice/Instance.cpp
pepone: > it's desirable it has a refcount on the instance.

If the instance owns the thread, the thread doesn't need to keep a refcount.

bernardnormier: You're right, I've updated the PR to join the threads in the destructor.

pepone (approved):

externl (approved):

PR #3817, opened 2025-04-03
Add async with support to Python Communicator

pepone:
This PR includes the required C++ updates from Bernard’s #3815. All C++ changes are taken directly from that PR — this one only covers the Python side.

### DIFF
--- a/python/modules/IcePy/Communicator.cpp
+++ b/python/modules/IcePy/Communicator.cpp
@@ -256,14 +256,9 @@ communicatorDestroy(CommunicatorObject* self, PyObject* /*args*/)
     auto vfm = dynamic_pointer_cast<ValueFactoryManager>((*self->communicator)->getValueFactoryManager());
     assert(vfm);
 
-    try
     {
         AllowThreads allowThreads; // Release Python's global interpreter lock to avoid a potential deadlock.
         (*self->communicator)->destroy();
-    }
-    catch (...)
-    {
-        setPythonException(current_exception());
     }
 
     vfm->destroy();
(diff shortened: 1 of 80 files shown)


python/modules/IcePy/Communicator.cpp
```diff
@@ -256,15 +256,10 @@ communicatorDestroy(CommunicatorObject* self, PyObject* /*args*/)
     auto vfm = dynamic_pointer_cast<ValueFactoryManager>((*self->communicator)->getValueFactoryManager());
     assert(vfm);
 
-    try
     {
         AllowThreads allowThreads; // Release Python's global interpreter lock to avoid a potential deadlock.
         (*self->communicator)->destroy();
```
pepone, line 261: This method is noexcept no need this catch block.

externl (approved):

python/modules/IcePy/Communicator.cpp
```diff
@@ -289,6 +282,65 @@ communicatorDestroy(CommunicatorObject* self, PyObject* /*args*/)
...
+    // Create a new reference to prevent premature release of the future object. The reference will be released by
+    // the destroy callback.
+    PyObject* futureObject = Py_NewRef(future.get());
+
+    (*self->communicator)
+        ->destroyAsync(
+            [self, vfm, futureObject]()
+            {
+                // Ensure the current thread is able to call into Python.
+                AdoptThread adoptThread;
```
bernardnormier, line 315: Just to confirm: AdoptThread is recursive - it works fine even if you have already the GIL locked?

bernardnormier (approved):

PR #3820, opened 2025-04-03
Update Communicator and OA doc comments

bernardnormier:


cpp/include/Ice/Communicator.h
```diff
@@ -29,13 +36,14 @@ namespace Ice
...
+        /// object adapters, and closes all outgoing connections. `destroy` waits for all outstanding dispatches to
+        /// complete before returning. This includes "bidirectional dispatches" that execute on outgoing connections.
         /// @see CommunicatorHolder
         void destroy() noexcept;
 
         /// Destroys this communicator asynchronously.
-        /// @param completed The callback to call when the destruction is complete. This callback must not throw any
+        /// @param completed The callback to call when the destruction is complete. This function must not throw any
         /// exception.
         /// @remark This function starts a thread to call @p completed unless you call this function on a communicator
```
pepone, line 48: Should we add to call "destroy" here?

pepone (approved):

cpp/include/Ice/Communicator.h
bernardnormier: Added.

externl (approved):

PR #3823, opened 2025-04-04
Move destructor of RegistryPluginFacade out of line

bernardnormier:
A plugin sees this destructor but (in theory) never calls it since it can't / shouldn't instantiate this class.

### DIFF
--- a/cpp/include/IceGrid/PluginFacade.h
+++ b/cpp/include/IceGrid/PluginFacade.h
@@ -60,7 +60,7 @@ namespace IceGrid
     class RegistryPluginFacade
     {
     public:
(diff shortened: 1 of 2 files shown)


cpp/include/IceGrid/PluginFacade.h
```diff
@@ -60,7 +60,7 @@ namespace IceGrid
     class RegistryPluginFacade
```
bernardnormier, line 60: There is no ICEGRID_API decoration because the implementation is provided by the icegridregistry - that loads your plugin - not by a shared library you link with.

externl (approved):
Seems fine to me

pepone (approved):

PR #3825, opened 2025-04-04
Update documentation for IceBox and Ice::Service

bernardnormier:


cpp/include/Ice/Service.h
```diff
@@ -91,120 +67,88 @@ namespace Ice
...
+        /// program processes its command-line options and calls #run. To run as a Win32 service or Unix daemon, the
+        /// program must first call `configureService` or #configureDaemon, respectively.
+        /// @param argc The number of arguments in @p argv.
         /// @param argv The command-line arguments.
-        /// @param initData Configuration data for the new Communicator.
+        /// @param initData Configuration data for the new communicator.
         /// @return The application's exit status: EXIT_FAILURE or EXIT_SUCCESS.
         int run(int argc, const char* const argv[], InitializationData initData = {});
 
 #ifdef _WIN32
```
bernardnormier, line 104: This `#ifdef _WIN32` doc doesn't work, obviously. 

We could define _WIN32 when building the doc, but then we'd lose the `#else`. A solution would be to split this class in Ice::WindowsService and Ice::DaemonService, and document both ... but quite a bit of work for this rarely used class.

Maybe we should deprecate this class for Ice 3.8?

cpp/include/Ice/Service.h
```diff
@@ -20,19 +20,18 @@ namespace Ice
...
         virtual ~Service();
 
-        /// Shutdown the service. The default implementation invokes shutdown()
-        /// on the communicator.
+        /// Shuts down the service. The default implementation calls Communicator::shutdown.
         virtual bool shutdown();
 
-        /// Notify the service about a signal interrupt. The default
-        /// implementation invokes shutdown().
+        /// Notifies the service that the program received a signal. The default implementation calls #shutdown.
```
pepone, line 31: What about documenting what signals.

cpp/include/Ice/Service.h
```diff
@@ -91,120 +67,88 @@ namespace Ice
...
-        /// command-line options. Instead of invoking main(), the
-        /// program processes its command-line options and invokes
-        /// run(). To run as a Win32 service or Unix daemon, the
-        /// program must first invoke configureService() or
-        /// configureDaemon(), respectively.
-        ///
-        /// @param argc Specifies the number of arguments in argv.
+        /// Alternative entry point for services that use their own command-line options. Instead of calling #main, the
+        /// program processes its command-line options and calls #run. To run as a Win32 service or Unix daemon, the
+        /// program must first call `configureService` or #configureDaemon, respectively.
```
pepone, line 97: Why do we use different notation for configureService and configureDaemon?

cpp/include/Ice/Service.h
```diff
@@ -91,120 +67,88 @@ namespace Ice
...
 
     protected:
-        /// Prepares a service for execution, including the creation and
-        /// activation of object adapters and servants.
-        /// @param argc Specifies the number of arguments in argv.
+        /// Prepares a service for execution, including the creation and activation of object adapters and servants.
+        /// @param argc The number of arguments in @p argv.
         /// @param argv The command-line arguments.
-        /// @param status The exit status, which is returned by main
+        /// @param status The exit status, which is returned by #main.
```
pepone, line 134: Don't we need to mark it as out?

pepone (approved):

cpp/include/Ice/Service.h
bernardnormier: Because `configureService` is not documented, it's a _WIN32-only function,

cpp/include/Ice/Service.h
bernardnormier: Yes, fixed.

cpp/include/Ice/Service.h
pepone: For `Ice/SSL/ClientAuthenticationOptions.h`

I use 

```
#if defined(ICE_USE_SCHANNEL)
endif

#if defined(ICE_USE_SECURE_TRANSPORT)
endif
```

This way doxygen would seem both, but maybe not worth here.

cpp/include/Ice/Service.h
bernardnormier: I could document it with ICE_DOXYGEN.

externl (approved):

bernardnormier:
I simplified Ice::Service based on our usage.

cpp/include/Ice/Service.h
```diff
@@ -43,168 +43,77 @@ namespace Ice
...
-        ///
-        /// If --service or --daemon are specified, the program runs as
-        /// a service, otherwise the program runs as a regular foreground
-        /// process. Any service-specific (and Ice-specific) options
-        /// are stripped from argv (just as for Ice::initialize()).
-        ///
-        /// @param args The command-line arguments.
-        /// @param initData Configuration data for the new Communicator.
-        /// @return The application's exit status: EXIT_FAILURE or EXIT_SUCCESS.
-        int main(const StringSeq& args, InitializationData initData = {});
```
bernardnormier, line 102: Removed because we never call this function.

cpp/include/Ice/Service.h
```diff
@@ -43,168 +43,77 @@ namespace Ice
...
-        /// program processes its command-line options and invokes
-        /// run(). To run as a Win32 service or Unix daemon, the
-        /// program must first invoke configureService() or
-        /// configureDaemon(), respectively.
-        ///
-        /// @param argc Specifies the number of arguments in argv.
-        /// @param argv The command-line arguments.
-        /// @param initData Configuration data for the new Communicator.
-        /// @return The application's exit status: EXIT_FAILURE or EXIT_SUCCESS.
-        int run(int argc, const wchar_t* const argv[], InitializationData initData = {});
```
bernardnormier, line 134: Removed because never called.

cpp/include/Ice/Service.h
```diff
@@ -43,168 +43,77 @@ namespace Ice
...
-        /// program processes its command-line options and invokes
-        /// run(). To run as a Win32 service or Unix daemon, the
-        /// program must first invoke configureService() or
-        /// configureDaemon(), respectively.
-        ///
-        /// @param argc Specifies the number of arguments in argv.
-        /// @param argv The command-line arguments.
-        /// @param initData Configuration data for the new Communicator.
-        /// @return The application's exit status: EXIT_FAILURE or EXIT_SUCCESS.
-        int run(int argc, const char* const argv[], InitializationData initData = {});
```
bernardnormier, line 148: Moved to `private:`

cpp/include/Ice/Service.h
externl: Surprising.

externl (approved):

cpp/include/Ice/Service.h
bernardnormier: We actually use it in the implementation of the wchar main, so I restored it.

cpp/include/Ice/Service.h
```diff
@@ -43,44 +43,22 @@ namespace Ice
...
-        /// If --service or --daemon are specified, the program runs as
-        /// a service, otherwise the program runs as a regular foreground
-        /// process. Any service-specific (and Ice-specific) options
-        /// are stripped from argv (just as for Ice::initialize()).
-        ///
-        /// @param argc Specifies the number of arguments in argv.
-        /// @param argv The command-line arguments.
-        /// @param initData Configuration data for the new Communicator.
-        /// @return The application's exit status: EXIT_FAILURE or EXIT_SUCCESS.
+#if defined(_WIN32) || defined(ICE_DOXYGEN)
```
pepone, line 55: Why is the `ICE_DOXYGEN` required:

I thought that with our setup that sets `MACRO_EXPANSION=yes` that wouldn't be required.
 
```
# If the MACRO_EXPANSION tag is set to YES, doxygen will expand all macro names
# in the source code. If set to NO, only conditional compilation will be
# performed. Macro expansion can be done in a controlled way by setting
# EXPAND_ONLY_PREDEF to YES.
# The default value is: NO.
# This tag requires that the tag ENABLE_PREPROCESSING is set to YES.

MACRO_EXPANSION        = YES
```

pepone (approved):

cpp/include/Ice/Service.h
bernardnormier: We use ICE_DOXYGEN for conditional "compilation" when generating doc with Doxygen. Our Doxyfile defines ICE_DOXYGEN. 

It sounds like this MACRO_EXPANSION is an unrelated feature.

Issue #3257, opened 2024-12-11
Should `ice2slice` Use the New `MetadataValidator`?


InsertCreativityHere:
Currently `ice2slice` performs no validation of metadata.
I think it would be preferable to still validate metadata, even if it is 'different' from the other Slice compilers.
We could also co-opt the `MetadataValidator` to issue warnings for metadata that we can't map in new Slice.

bernardnormier:
If ice2slice "translates" Ice metadata into Slice attributes, then it should validate the metadata it consumes. Currently it would be only cs metadata, in addition to non-language-specific metadata.

PR #3827, opened 2025-04-07
Minor 'ice2slice' Fixes

InsertCreativityHere:
This PR:
- fixes a bug in `ice2slice` where it wasn't registering generated files with the `FileTracker`.
- removes an `assert` which caused `ice2slice` to crash when it saw an operation.


cpp/src/Slice/FileTracker.cpp
```diff
@@ -9,12 +9,6 @@
 using namespace IceInternal;
 using namespace std;
 
-const char*
-Slice::FileException::ice_id() const noexcept
-{
-    return "::Slice::FileException";
-}
-
```
InsertCreativityHere, line 17: This exception is only used internally by the Slice compilers.
We never check or care about it's id. So why bother defining it.

cpp/src/Slice/Parser.cpp
```diff
@@ -911,12 +911,9 @@ Slice::Contained::scope() const
...
 Slice::Contained::mappedName() const
 {
-    const string languageName = unit()->languageName();
-    assert(!languageName.empty());
-
     // First check if any 'xxx:identifier' has been applied to this element.
     // If so, we return that instead of the element's Slice identifier.
-    const string metadata = languageName + ":identifier";
+    const string metadata = unit()->languageName() + ":identifier";
     if (auto customName = getMetadataArgs(metadata))
```
InsertCreativityHere, line 917: This assert was to prevent us from calling `mappedName` in `ice2slice` (ice2slice should always use the Slice name).
But, there are helper functions in `libSlice` which use this function; Indirectly causing `ice2slice` to crash.

So I removed the assert.
We can self-police ourselves to not use `mappedName` in `ice2slice`.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -610,11 +611,11 @@ Gen::TypesVisitor::visitSequence(const SequencePtr& p)
             }
             out << "<"
                 // TODO the generic argument must be the IceRPC C# mapped type
-                << typeToString(p->type(), p->scope(), false) << ">\")]";
+                << typeToString(p->type(), scope, false) << ">\")]";
```
InsertCreativityHere, line 614: We already store `p->scope()` in a variable.
No need to keep calling it over and over.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -704,6 +704,7 @@ Gen::TypesVisitor::getOutput(const ContainedPtr& contained)
     {
         string outputName = getOutputName(_fileBase, _modules.size() > 1 ? scope : "");
         unique_ptr<Output> out = make_unique<Output>(outputName.c_str());
+        FileTracker::instance()->addFile(outputName);
```
InsertCreativityHere, line 707: Whenever a Slice compiler creates a file/directory, it should let `FileTracker` know.

For `ice2slice` specifically, it calls `FileTracker::cleanup()` if something goes wrong:
https://github.com/zeroc-ice/ice/blob/ca07ff1899670987053a2d3eb34424c3ff548c2c/cpp/src/ice2slice/Main.cpp#L215-L218 But this call is useless without telling `FileTracker` which files were generated.

pepone (approved):

externl (approved):

cpp/src/Slice/Parser.cpp
```diff
@@ -911,12 +911,9 @@ Slice::Contained::scope() const
 string
 Slice::Contained::mappedName() const
 {
-    const string languageName = unit()->languageName();
-    assert(!languageName.empty());
-
     // First check if any 'xxx:identifier' has been applied to this element.
     // If so, we return that instead of the element's Slice identifier.
-    const string metadata = languageName + ":identifier";
+    const string metadata = unit()->languageName() + ":identifier";
```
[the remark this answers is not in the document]
InsertCreativityHere: That's actually a very sneaky find. I didn't see it, but you're right, a user could use `:identifier` to have an effect here.

But yes, we need to add validation which would reject bogus metadata like this. See: #3257.

cpp/src/Slice/FileTracker.cpp
bernardnormier: Whenever we define a LocalException, it needs to be a proper LocalException, with an ice_id(). We should never "half-define" a LocalException.

InsertCreativityHere: It is now a `runtime_error` instead of a `LocalException`.

It really has no business being a `LocalException` since it doesn't interact at all with the Ice runtime.
And none of the functionality offered by `LocalExecption` was ever used by it.

bernardnormier (approved):

PR #3828, opened 2025-04-07
Remove Usage of Right-Recursion in the Slice Parser

### LINKED ISSUE #3736: Type Checking of Lists Happens in an Unexpected Order
Let's say you have this Slice interface:
```
interface BarIntf extends MyModule, E { ... }
```
Where `MyModule` and `E` are a module and exception.
Of course, neither of these are interfaces, and hence, this is bogus. All good.

But, you'd expect that the error for `MyModule` would come before the error for `E`, since that's the declaration order.
However, they are emitted in reverse order:
```
NotType.ice:16: 'E' is an exception, which cannot be used as a type
NotType.ice:16: 'MyModule' is not a type
```
They also have very different messags, which is sub-optimal.

InsertCreativityHere:
Before this PR, we had a mix of left-recursion and right-recursion in the parser.
You should never use right recursion with LR-parsers (the kind that Bison generates). Taken from the [Bison manual](https://www.gnu.org/software/bison/manual/html_node/Recursion.html) itself:
> you should always use left recursion

Doing this:
- fixes a bug where the parser allowed leading commas in certain lists: `op(, int i);`
- fixes #3736 (where errors were emitted in reverse order)
- fixes the line numbers reported for some errors.

----

For example, compiling this Slice:

```slice
module Hello                   // 1
{                              // 2
    interface MyOperations :   // 3
        Hello,                 // 4
        MyOperations,          // 5
        S                      // 6
    {}                         // 7
}                              // 8
```
(on main) causes these errors:
```
D:/Code/Workspace/ice/test.ice:7: 'S' is not defined
D:/Code/Workspace/ice/test.ice:7: 'MyOperations' is not defined
D:/Code/Workspace/ice/test.ice:7: 'Hello' is not a type
```
The order is reverse of the Slice definition, and they all say line 7, despite being on different lines.

After this PR, the output is:
```
D:/Code/Workspace/ice/test.ice:4: 'Hello' is not a type
D:/Code/Workspace/ice/test.ice:5: 'MyOperations' is not defined
D:/Code/Workspace/ice/test.ice:6: 'S' is not defined
```
Much better.

----

### What the heck is 'Right-Recursion'

Recursive grammar rules are rules that call themselves; often used to handle lists.
Right recursion is when the recursive rule is on the right side, left-recursion is when it's on the left:
```
right_recurse:
    thing ',' right_recurse
vs
left_recurse:
    left_recurse ',' thing
```

Left-recursion matches our intuition that parsing happens left-to-right.
There's also algorithmic/performance reasons which make left recursion much better than right.


cpp/src/Slice/Grammar.y
```diff
@@ -1774,10 +1784,14 @@ out_qualifier
...
-}
-| out_qualifier metadata optional_type_id
+: parameter_list
+| %empty
+;
+
+// ----------------------------------------------------------------------
+parameter_list
+// ----------------------------------------------------------------------
+: out_qualifier metadata optional_type_id
```
InsertCreativityHere, line 1794: Moving `%empty` out of `parameter_list` into an intermediate rule fixes the leading comma bug.
The old semantics was allowing a _parameter_ to be empty, as opposed to a _parameter list_ to be empty.
(Empty just means the empty string).

Ex: this was valid `void op(, int i);` because it was matching the empty string as the first parameter.

cpp/src/Slice/Grammar.y
```diff
@@ -1633,44 +1633,54 @@ ICE_ENUM
...
+| enumerator_list ','
+{
+    $$ = $1;
+}
+| %empty
+{
+    $$ = make_shared<EnumeratorListTok>(); // Empty list
+}
+;
+
```
InsertCreativityHere, line 1656: There was no change to the behavior of enums.
This is to allow trailing commas in enums, after switching to left-recursion.

cpp/src/Slice/Grammar.y
```diff
@@ -1396,7 +1396,7 @@ interface_list
             else
             {
                 cont->checkIntroduced(scoped->v);
-                intfs->v.push_front(def);
+                intfs->v.push_back(def);
```
InsertCreativityHere, line 1399: Another nice effect is now we can use the more natural 'push_back' instead of 'push_front'.

externl (approved):
Looks good to me!

PR #3830, opened 2025-04-07
Rework Ice::SSL and WIN32 doc to use ICE_DOXYGEN

bernardnormier:
This PR updates the Ice::SSL and _WIN32 API doc to use ICE_DOXYGEN.

### DIFF
--- a/cpp/include/Ice/Communicator.h
+++ b/cpp/include/Ice/Communicator.h
@@ -138,7 +138,7 @@ namespace Ice
         /// It is legal to create an object adapter with the empty string as its name. Such an object adapter is
         /// accessible via bidirectional connections or by collocated invocations.
         /// @param name The object adapter name.
-        /// @param serverAuthenticationOptions The SSL configuration properties for server connections.
+        /// @param serverAuthenticationOptions The SSL options for server connections.
         /// @return The new object adapter.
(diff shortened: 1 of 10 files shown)


cpp/include/Ice/Communicator.h
```diff
@@ -138,9 +138,7 @@ namespace Ice
         /// It is legal to create an object adapter with the empty string as its name. Such an object adapter is
         /// accessible via bidirectional connections or by collocated invocations.
         /// @param name The object adapter name.
-        /// @param serverAuthenticationOptions The %SSL configuration properties for server connections.
-        /// The `SSL::ServerAuthenticationOptions` type is an alias to the platform specific %SSL server authentication
-        /// options.
+        /// @param serverAuthenticationOptions The SSL configuration properties for server connections.
```
pepone, line 141: maybe we should say options instead of properties here.

pepone (approved):

externl (approved):

cpp/include/Ice/Communicator.h
bernardnormier: Yes. I replaced all "configuration properties" by "options" when talking about SSL options.

I'd rather reserve the term configuration to Ice properties / config files etc. Not include options classes/structs.

PR #3831, opened 2025-04-07
Undefined Types No Longer Trigger Syntax Errors

### LINKED ISSUE #3735: Uncaught Syntax Error in Operation Syntax
In the 'errorDetection/NotType' test, we (presumably accidentally) used a `:` instead of a `,` to separate these parameters:
https://github.com/zeroc-ice/ice/blob/b344d2f85b7add5011b92db8b55aa6514672364c/cpp/test/Slice/errorDetection/NotType.ice#L27
This should be a syntax error, but there are no syntax errors in the corresponding `.err` file.

InsertCreativityHere:
This PR fixes #3735 (where a bogus  `;` wasn't correctly triggering a syntax error).

----

The Slice parser performs type resolution in the middle of parsing (hence the need for forward declarations).
Normally, this is fine. What isn't fine: if the parser _fails_ to resolve a type, it 'throws' a syntax error.

This is just wrong. `sequence<Foo> MySeq;` is _syntactically valid_ Slice, whether or not `Foo` is defined in scope.
And it causes real problems. Syntax errors are serious errors, they cause the parser to halt normal operation until it hits a recovery point (ie. `;` or `}`). This means it skips over other errors it should be reporting, or could end up in a bad state from which it cannot recover.

----

This PR changes the parser to more gracefully handle undefined types.
Instead of 'throw'ing a syntax error, now the parser returns `nullptr` to signal the type didn't exist, then continues like normal.
Some functions needed to be updated to handle that `TypePtr` could be `nullptr` now. But only functions internal to the parser.
We still report an error for these, so the compiler will immediately stop after the parsing stage.

### DIFF
--- a/cpp/src/Slice/Grammar.cpp
+++ b/cpp/src/Slice/Grammar.cpp
@@ -747,11 +747,11 @@ static const yytype_int16 yyrline[] =
     1646,  1647,  1651,  1660,  1672,  1689,  1695,  1713,  1725,  1729,
     1770,  1776,  1787,  1788,  1794,  1810,  1826,  1838,  1850,  1861,
     1877,  1881,  1890,  1893,  1901,  1902,  1903,  1904,  1905,  1906,
-    1907,  1908,  1909,  1910,  1915,  1919,  1924,  1949,  1979,  1985,
-    1993,  2003,  2018,  2027,  2036,  2071,  2078,  2085,  2097,  2106,
-    2120,  2121,  2122,  2123,  2124,  2125,  2126,  2127,  2128,  2129,
-    2130,  2131,  2132,  2133,  2134,  2135,  2136,  2137,  2138,  2139,
-    2140,  2141,  2142,  2143,  2144,  2145,  2146
+    1907,  1908,  1909,  1910,  1915,  1919,  1924,  1950,  1981,  1987,
+    1995,  2005,  2020,  2029,  2038,  2073,  2080,  2087,  2099,  2108,
+    2122,  2123,  2124,  2125,  2126,  2127,  2128,  2129,  2130,  2131,
+    2132,  2133,  2134,  2135,  2136,  2137,  2138,  2139,  2140,  2141,
+    2142,  2143,  2144,  2145,  2146,  2147,  2148
 };
 #endif
 
@@ -2337,7 +2337,7 @@ yyreduce:
     // It's safe to perform this check in the parser, since we already have enough information to know whether a type
     // can be optional. This is because the only types that can be forward declared (classes/interfaces) have constant
     // values for `usesClasses` (true/false respectively).
-    if (m->type->usesClasses())
+    if (m->type && m->type->usesClasses())
     {
         currentUnit->error("types that use classes cannot be marked with 'optional'");
     }
@@ -2849,7 +2849,7 @@ yyreduce:
     // It's safe to perform this check in the parser, since we already have enough information to know whether a type
     // can be optional. This is because the only types that can be forward declared (classes/interfaces) have constant
     // values for `usesClasses` (true/false respectively).
-    if (m->type->usesClasses())
+    if (m->type && m->type->usesClasses())
     {
         currentUnit->error("types that use classes cannot be marked with 'optional'");
     }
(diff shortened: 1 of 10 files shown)


cpp/test/Slice/errorDetection/WrongProxyType.ice
```diff
@@ -8,12 +8,12 @@ module Test
...
-        void f2(Seq*);
-        void f3(out Seq*);
+        void f2(Seq* s);
+        void f3(out Seq* s);
 
         Dict* f4();
-        void f5(Dict*);
-        void f6(out Dict*);
+        void f5(Dict* s);
+        void f6(out Dict* s);
```
InsertCreativityHere, line 16: All these parameters were missing an identifier.
But because of a type-resolution-error (putting '*' after a sequence is bogus),
this "missing identifier" error was being skipped over.

cpp/test/Slice/errorDetection/Undefined.ice
```diff
@@ -2,8 +2,8 @@
 
 module Test
 {
-    sequence<Foo> FooSequence;
+    sequence<Fake> FooSequence;
     dictionary<Foo, Bar> FooBarDictionary;
-    interface BarIntf extends Foo { void op(); }
-    class BarClass1 extends Foo { long l; }
+    interface BarIntf extends Ping { void op(); }
+    class BarClass1 extends Pong { long l; }
```
InsertCreativityHere, line 8: I made the names different, so it's easier to read the `.err`, and to make the parser has the right line numbers for the right errors.

cpp/test/Slice/errorDetection/Undefined.err
```diff
@@ -1,4 +1,5 @@
-Undefined.ice:5: 'Foo' is not defined
+Undefined.ice:5: 'Fake' is not defined
 Undefined.ice:6: 'Foo' is not defined
-Undefined.ice:7: 'Foo' is not defined
-Undefined.ice:8: 'Foo' is not defined
+Undefined.ice:6: 'Bar' is not defined
```
InsertCreativityHere, line 3: In this line:
`dictionary<Foo, Bar> FooBarDictionary;`
Both `Foo` and `Bar` are undefined, but only `Foo` was reported.
Since the parser skipped ahead to the `;` recovery point, ignoring everything in between.

Now we correctly emit an error for both `Foo` and `Bar` being undefined.

cpp/test/Slice/errorDetection/NotType.ice
```diff
@@ -24,7 +24,7 @@ module Test
 
     interface I
     {
-        E foo(E e1; E e2);
+        E foo(E e1, E e2);
```
InsertCreativityHere, line 27: Using a `;` here is completely bogus.
But we weren't reporting a syntax error, because `E` isn't a valid type.

Now the compiler correctly reported the `;` and I fixed it.

cpp/test/Slice/errorDetection/NotType.err
```diff
@@ -6,6 +6,6 @@ NotType.ice:16: 'E' is an exception, which cannot be used as a type
 NotType.ice:17: 'Mod' is not a type
 NotType.ice:17: 'E' is an exception, which cannot be used as a type
 NotType.ice:22: 'E' is an exception, which cannot be used as a type
-NotType.ice:23: struct 'S' must have at least one member
+NotType.ice:27: 'E' is an exception, which cannot be used as a type
```
InsertCreativityHere, line 9: Another thing this PR fixes, the `struct 'S' must have at least one member` was bogus.

If a field's type didn't exist, the syntax error caused us to skip over the field completely.
Then, when the parser recovered, `S` looked like an empty struct.
```
struct S
{
    FakeType foo;
}
```
But this is obviously bogus. `S is empty` is not the real error here.
The real error is that `FakeType` wasn't a real type.

This has now been fixed.

cpp/test/Slice/errorDetection/IdentAsKeyword.err
```diff
@@ -37,16 +37,20 @@ IdentAsKeyword.ice:37: sequences can only be defined within a module
 IdentAsKeyword.ice:37: keyword 'short' cannot be used as sequence name
 IdentAsKeyword.ice:39: syntax error
 IdentAsKeyword.ice:40: 'moDule' is not defined
+IdentAsKeyword.ice:40: sequences can only be defined within a module
```
InsertCreativityHere, line 40: Added all the errors which were being skipped due to the old 'syntax error' semantics.
You'll notice all these errors come directly after `type is not defined` or `... cannot be used as a type`.

cpp/test/Slice/errorDetection/IdentAsKeyword.ice
```diff
@@ -49,8 +49,8 @@ module Test
     dictionary<long, out> d3;
     dictionary<long, OUT> d4;
 
-    dictionary<void, void> d5;
-    dictionary<VOID, VOID> d6;
+    dictionary<void, voID> d5;
+    dictionary<VOID, VOid> d6;
```
InsertCreativityHere, line 53: I changed the names so I could tell the key-type apart from the element-type in the `.err` file.

cpp/src/Slice/Grammar.y
```diff
@@ -677,7 +677,7 @@ optional_type_id
     // It's safe to perform this check in the parser, since we already have enough information to know whether a type
     // can be optional. This is because the only types that can be forward declared (classes/interfaces) have constant
     // values for `usesClasses` (true/false respectively).
-    if (m->type->usesClasses())
+    if (m->type && m->type->usesClasses())
```
InsertCreativityHere, line 680: There were a handful of places where I had to add checks for `nullptr` before using a `TypePtr` now.

cpp/src/Slice/Grammar.y
```diff
@@ -1929,22 +1929,23 @@ type
     TypeList types = cont->lookupType(scoped->v);
     if (types.empty())
     {
-        YYERROR; // Can't continue, jump to next yyerrok
+        $$ = nullptr;
```
InsertCreativityHere, line 1932: `YYERROR` is a Bison macro which is basically 'throwing a syntax error'.
`$$ = ` is how you set a return value in Bison.

So, instead of throwing a syntax error, we just return `nullptr` now.

cpp/src/Slice/Grammar.cpp
InsertCreativityHere, line 1: `Grammar.cpp` is generated code. No need to read.

externl (approved):

bernardnormier (approved):

PR #3833, opened 2025-04-08
Deprecate and Replace `python:package` and `cs:namespace` with `python:identifier` and `cs:identifier`.

InsertCreativityHere:
The title pretty much says it all. `swift:module` and `java:package` are next.

I also simplified some `while` loops we were using to check for metadata on top-level modules.

### DIFF
--- a/cpp/src/ice2slice/Gen.cpp
+++ b/cpp/src/ice2slice/Gen.cpp
@@ -14,36 +14,20 @@ namespace
 {
     static string getCSharpNamespace(const ContainedPtr& cont, bool& hasCSharpNamespaceAttribute)
     {
-        // Traverse to the top-level module.
-        ModulePtr m;
         ContainedPtr p = cont;
         string csharpNamespace;
         while (true)
         {
             if (dynamic_pointer_cast<Module>(p))
             {
-                m = dynamic_pointer_cast<Module>(p);
-                if (csharpNamespace.empty())
-                {
-                    csharpNamespace = m->name();
-                }
-                else
-                {
-                    csharpNamespace = m->name() + "." + csharpNamespace;
-                }
-            }
-
-            if (p->isTopLevel())
-            {
+                csharpNamespace = p->mappedScoped(".").substr(1);
                 break;
             }
             p = dynamic_pointer_cast<Contained>(p->container());
             assert(p);
         }
 
-        assert(m);
-
-        if (auto metadata = m->getMetadataArgs("cs:namespace"))
+        if (auto metadata = p->getMetadataArgs("cs:namespace"))
         {
             hasCSharpNamespaceAttribute = true;
             return *metadata + "." + csharpNamespace;
--- a/cpp/src/ice2slice/Gen.cpp
+++ b/cpp/src/ice2slice/Gen.cpp
@@ -14,13 +14,32 @@ namespace
 {
     static string getCSharpNamespace(const ContainedPtr& cont, bool& hasCSharpNamespaceAttribute)
     {
+        // Traverse to the top-level module.
         ContainedPtr p = cont;
-        while (!dynamic_pointer_cast<Module>(p))
+        string csharpNamespace;
+        while (true)
         {
+            if (dynamic_pointer_cast<Module>(p))
+            {
+                // TODO we should be using the `mappedName` here, not the slice `name`.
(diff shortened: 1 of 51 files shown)


python/test/Ice/packagemd/Package.ice
```diff
@@ -2,6 +2,8 @@
```
InsertCreativityHere, line 2: The `packagemd` test is specifically about testing `python:package`, so I left it alone,
except for suppressing the deprecation warnings (I checked, and without the `suppress`, the deprecation warning was being correctly emitted).

csharp/test/Ice/namespacemd/Namespace.ice
```diff
@@ -2,6 +2,8 @@
```
InsertCreativityHere, line 2: The `namespacemd` test is specifically about testing `cs:namespace`, so I left it alone,
except for suppressing the deprecation warnings (I checked, and without the `suppress`, the deprecation warning was being correctly emitted).

cpp/src/slice2cs/Gen.cpp
```diff
@@ -874,27 +874,21 @@ Slice::CsVisitor::writeParameterDocComments(const DocComment& comment, const Par
 void
 Slice::CsVisitor::moduleStart(const ModulePtr& p)
 {
```
InsertCreativityHere, line 876: We already validate that `cs:namespace` is only allowed on top-level modules.
If it wasn't we threw the metadata out.

So, checking for `isTopLevel` here is redundant and unnecessary.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -24,29 +24,16 @@ using namespace IceInternal;
 string
 Slice::CsGenerator::getNamespacePrefix(const ContainedPtr& cont)
 {
-    //
```
InsertCreativityHere, line 27: Simplified this loop.

`m` and `p` point to the same Slice element, the only difference is their typing.
But since all we do with `m` is call `getMetadataArgs` on it, we don't need this stricter typing.
(`getMetadataArgs` is defined on both `Contained` (`p`) and `Module` (`m`)).

cpp/src/ice2slice/Gen.cpp
```diff
@@ -15,21 +15,20 @@ namespace
     static string getCSharpNamespace(const ContainedPtr& cont, bool& hasCSharpNamespaceAttribute)
     {
         // Traverse to the top-level module.
```
InsertCreativityHere, line 17: `m` and `p` pointed to the same Slice element.
The only difference is `m` was more strongly typed.
But this extra typing information wasn't necessary.

cpp/src/slice2cs/Gen.cpp
bernardnormier: I don't understand this code. Say we have no metadata whatsoever - we don't generate a C# "namespace Xxx" here?

externl: This function seems to be about generating an additional namespace declaration if I read the code correctly. 

```
void
Slice::CsVisitor::moduleStart(const ModulePtr& p)
{
    if (p->isTopLevel())
    {
        string ns = getNamespacePrefix(p);
        if (!ns.empty())
        {
            _out << sp;
            _out << nl << "namespace " << ns;
            _out << sb;
        }
    }
}
```

This function and the corresponding `moduleEnd` are called the visitors.

```
bool
Slice::Gen::TypesVisitor::visitModuleStart(const ModulePtr& p)
{
    moduleStart(p);
    _out << sp;
    _out << nl << "namespace " << p->mappedName();
    _out << sb;

    return true;
}


```

externl: They should be renamed `modulePrefixStart` and `modulePrefixEnd` as they're only about the `cs:namespace` metadata.

externl (approved):

cpp/src/slice2cs/Gen.cpp
InsertCreativityHere: Renamed them as suggested.
I agree, the previous names were confusing.

It took me some time to realize that these are just random helper functions named `moduleStart`,
not the _actual_ code-gen functions `visitModuleStart`. Probably the same as @bernardnormier.

PR #3837, opened 2025-04-09
Allow Metadata to Come Before `out` Modifier

### LINKED ISSUE #3531: The Slice Compiler Should be More Tolerant of Metadata Placement
I really want to be able to write this:
```slice
interface foo
{
    void myOp(["java:identifier:java_continue"] out int continue);
}
```
But that causes a syntax error, and this is the correct way to write it:
```slice
interface foo
{
    void myOp(out ["java:identifier:java_continue"] int continue);
}
```

I suspect there's a similar issue with other modifier keywords.
We should change the Slice compiler to either:
1) Always require the metadata to be "at the front" of a Slice 'statement'.
2) Allow arbitrary mixing of metadata/modifier keywords (ie. you can have `<metadata> out` or `out <metadata>`, or heck, even `<metadata> out <metadata>`.

InsertCreativityHere:
Fixes #3531. Currently, this is a syntax error:
```
op(["metadata"] out int i);
```
The parser requires metadata to come _after_ the `out` keyword for parameters.
This seems inconsistent with how metadata works on everything else, and IMO just looks bad.

This PR loosens the parser's rules, so that metadata can appear before and after `out`.
So, the old placement is still fully supported, but now the snippet at the top is supported too.

----

I also went through our tests and updated (almost) all of them to use the new placement.
There are still some Slice tests that use the old placement, so we can be sure that both placements are supported.


cpp/src/Slice/Grammar.cpp
InsertCreativityHere, line 1: Generated file, don't need to read.

cpp/src/Slice/Grammar.y
```diff
@@ -206,6 +206,14 @@ file_metadata
...
 ;
 
+// ----------------------------------------------------------------------
+local_metadata
+// ----------------------------------------------------------------------
+: ICE_METADATA_OPEN metadata_list ICE_METADATA_CLOSE
+{
+    $$ = $2;
+}
+
```
InsertCreativityHere, line 216: We were matching these tokens in the `metadata_directives` rule you'll see befow this one.
I just refactored them into a separate rule for grammar reasons.

cpp/src/Slice/Grammar.y
```diff
@@ -1264,13 +1272,22 @@ operation
...
+        for (const auto& param : op->parameters())
+        {
+            const bool isOutParam = param->isOutParam();
+            if (!isOutParam && seenOutParam)
+            {
+                currentUnit->error("parameter '" + param->name() + "': in-parameters cannot follow out-parameters");
+            }
+            seenOutParam |= isOutParam;
+        }
+
```
InsertCreativityHere, line 1287: We used to this check in `Parser.cpp: createParameter`, called once for each parameter.
Now we only do it once. Right here, when we've finished creating an operation.

I think this is cleaner.
But was also necessary. One of the things I changed is that we don't set `isOutParam` until _after_ we've initialized a parameter now.

cpp/src/Slice/Grammar.y
```diff
@@ -1765,112 +1782,81 @@ enumerator_initializer
...
-        op->createParameter(Ice::generateUUID(), type, isOutParam->v, false, 0); // Dummy
-        currentUnit->error("missing parameter name");
+        auto metadata = dynamic_pointer_cast<MetadataListTok>($1);
+        param->appendMetadata(std::move(metadata->v));
     }
+    $$ = $2;
 }
 ;
 
+// ----------------------------------------------------------------------
```
InsertCreativityHere, line 1846: This wasn't changed. It was just moved down.
Unfortunately git doesn't realize this.
But if you look under where `out_qualifier` used to be, you'll see it.

cpp/src/Slice/Parser.cpp
```diff
@@ -3196,7 +3196,7 @@ Slice::Operation::hasMarshaledResult() const
 }
 
 ParameterPtr
-Slice::Operation::createParameter(const string& name, const TypePtr& type, bool isOutParam, bool isOptional, int tag)
+Slice::Operation::createParameter(const string& name, const TypePtr& type, bool isOptional, int tag)
```
InsertCreativityHere, line 3199: Instead of passing `isOutParam` during construction, we now call `setIsOutParam` separately (after construction).

The old grammar rule waited until we saw _everything_ to construct a parameter.
But, the flexibility of the new grammar rules mean that we have less information up-front now.

cpp/src/Slice/Parser.cpp
```diff
@@ -4465,6 +4450,17 @@ Slice::Parameter::isOutParam() const
...
 
+void
+Slice::Parameter::setIsOutParam()
+{
+    // Emit an error if this parameter was already marked as 'out' (meaning 'out' was written multiple times in a row).
+    if (_isOutParam)
+    {
+        unit()->error("parameter '" + _name + "' was marked 'out' multiple times");
+    }
+    _isOutParam = true;
```
InsertCreativityHere, line 4461: Random side-effect of this PR:
the parser can now more-gracefully handle syntax errors like: `out out int i`.

cpp/test/Slice/errorDetection/ParamsOutOfOrder.err
InsertCreativityHere, line 1: I made the error message a little nicer (IMO).

cpp/src/Slice/Grammar.y
```diff
@@ -1264,13 +1272,22 @@ operation
...
+        for (const auto& param : op->parameters())
+        {
+            const bool isOutParam = param->isOutParam();
+            if (!isOutParam && seenOutParam)
+            {
+                currentUnit->error("parameter '" + param->name() + "': in-parameters cannot come after out-parameters");
+            }
+            seenOutParam |= isOutParam;
+        }
+
```
InsertCreativityHere, line 1287: We used to this check in Parser.cpp: createParameter, called once for each parameter.
Now we only do it once. Right here, when we've finished creating an operation.

I think this is cleaner.
But was also necessary. One of the things I changed is that we don't set isOutParam until after we've initialized a parameter now.

cpp/src/Slice/Grammar.y
```diff
@@ -1765,112 +1782,81 @@ enumerator_initializer
...
-// ----------------------------------------------------------------------
-parameters
-// ----------------------------------------------------------------------
-: parameter_list
-| %empty
-;
-
-// ----------------------------------------------------------------------
-parameter_list
-// ----------------------------------------------------------------------
```
InsertCreativityHere, line 1793: I split `parameter_list` into 2 rules: `parameter` and `paramter_list`.
All the meat is in parameter, and now `parameter_list` is really just for parsing comma-separated parameters.
The current logic was too ugly, where we duplicate each possible match, one with a leading comma and one without,

cpp/src/Slice/Grammar.y
```diff
@@ -1765,112 +1782,81 @@ enumerator_initializer
...
+    if (auto param = dynamic_pointer_cast<Parameter>($2))
     {
-        op->createParameter(Ice::generateUUID(), type, isOutParam->v, false, 0); // Dummy
-        currentUnit->error("missing parameter name");
+        auto metadata = dynamic_pointer_cast<MetadataListTok>($1);
+        param->appendMetadata(std::move(metadata->v));
     }
+    $$ = $2;
 }
 ;
```
InsertCreativityHere, line 1844: `out` and `metadata` are modifiers now that we iteratively parse in front of a parameter.
Instead of being part of the 'core' parameter rule.

This lets us handle any ordering of `out` and `metadata` without duplication.
The alternative would be matching `out metadata` and `metadata out`, but that duplicates everything twice.

cpp/src/Slice/Grammar.y
```diff
@@ -1765,112 +1782,81 @@ enumerator_initializer
...
-}
-| %empty
+: optional_type_id
 {
-    auto out = make_shared<BoolTok>();
-    out->v = false;
-    $$ = out;
-}
-;
+    auto tsp = dynamic_pointer_cast<OptionalDefTok>($1);
```
pepone, line 1789: I notice you didn't change this name, but`tsp` is quite cryptic, what does it stand for?

cpp/src/Slice/Parser.cpp
```diff
@@ -4465,6 +4450,17 @@ Slice::Parameter::isOutParam() const
     return _isOutParam;
 }
 
+void
+Slice::Parameter::setIsOutParam()
+{
+    // Emit an error if this parameter was already marked as 'out' (meaning 'out' was written multiple times in a row).
+    if (_isOutParam)
+    {
+        unit()->error("parameter '" + _name + "' was marked 'out' multiple times");
```
pepone, line 4459: Does this emit a line error? Maybe we should add this to the error detection tests.

pepone (approved):

cpp/src/Slice/Grammar.y
InsertCreativityHere: My guess has always been "type string pointer".
But no, you're right. We really love using cryptic names in the grammar.
Maybe I'll do a quick cleanup-pass next week to make these all readable.

externl (approved):
Looks good to me. These grammar rules are tricky!

cpp/src/Slice/Grammar.y
pepone: > My guess has always been "type string pointer".

Nice one!

PR #3839, opened 2025-04-11
Remove unused BatchRequestQueue::isEmpty

### LINKED ISSUE #3773: Unused BatchRequestQueue::isEmpty

pepone:
Fix #3773


externl (approved):
No JavaScript?

pepone:
> No JavaScript?

Already fixed there!

bernardnormier (approved):

Issue #3836, opened 2025-04-09
Add initData.plugins to "load" plugins


bernardnormier:
Currently, you load a plugin in a communicator by configuring:
```
Ice.Plugin.name = entry-point
```

Where the name mostly does not matter - any unique name will do.

This plugin mechanism is available in C++, C# and Java.

>[!NOTE]
> As of Ice 3.8, all plugins used "fixed" configuration properties. In Ice 3.7, the configuration of a few plugins was plugin-name dependent. See #2798.

In C++, you can alternatively load most bundled plugins by calling a Ice::registerPluginName function, see:
https://github.com/zeroc-ice/ice/blob/c229e825122c1fd171e1a400951978a2a8cddeb4/cpp/include/Ice/RegisterPlugins.h#L33

The semantics is a little different: the property-based `Ice.Plugin.name = entry-point` is totally dynamic, with the plugin library / assembly / Java class loaded at runtime, whereas with the `Ice::registerXxx` call, the plugin library must be linked with your program.

Now, most of the time, the totally dynamic aspect is not needed. Your application depends on specific plugins, and you'd prefer a compile-time or link-time error if the plugin is unavailable (because you made a typo somewhere) over a runtime error.

## Proposal: add initData.plugins with a list of plugin factories

In C++, C# and Java, initData.plugins represents a list of PluginFactory that the communicator loads/creates during initialization.

The PluginFactory remains the same as in Ice 3.7:
- C++ (function pointer): https://github.com/zeroc-ice/ice/blob/c229e825122c1fd171e1a400951978a2a8cddeb4/cpp/include/Ice/Initialize.h#L231
- C# (metaclass): https://github.com/zeroc-ice/ice/blob/c229e825122c1fd171e1a400951978a2a8cddeb4/csharp/src/Ice/PluginManagerI.cs#L14
- Java (metaclass): https://github.com/zeroc-ice/ice/blob/c229e825122c1fd171e1a400951978a2a8cddeb4/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/PluginFactory.java#L6

We would also remove all the Ice::registerXxx APIs in C++, which I assume are rarely used.

The config-based "Ice.Plugin.name = entryPoint" plugin loading remains unchanged.

TBD: what do we do for C++-based mappings like Python and Ruby.
We could keep:
```
Ice.Plugin.name = 1
```
or
```
Ice.Plugin.name = anything
```

[...]

PR #3842, opened 2025-04-11
Replace global plugin registry by initData.pluginFactories in C++ and C#

bernardnormier:
This PR replaces the global plugin registry (`Ice::registerIceDiscovery(bool)` etc.) by an initData.pluginFactories list.

If you insert a plugin factory in this list, it gets created and initialized by the communicator. It's largely the same idea as the old global registry, except not global.

In particular, `Ice::registerIceDiscovery(bool)` (and friends) did the following:
 - specify a name of the plugin ("IceDiscovery", "IceUDP", etc.)
   I replaced this name by `PluginFactory::pluginName` (C++) and `PluginFactory.pluginName` (C#)
 - provide the factory function to use (C++) / factory object to use (C#)

See:
https://github.com/zeroc-ice/ice/blob/c229e825122c1fd171e1a400951978a2a8cddeb4/cpp/include/Ice/Initialize.h#L241

and

https://github.com/zeroc-ice/ice/blob/c229e825122c1fd171e1a400951978a2a8cddeb4/csharp/src/Ice/Util.cs#L490

See #3836.

I'll update Java once #3802 is merged.


cpp/src/Ice/AddDefaultPluginFactories.h
```diff
@@ -0,0 +1,16 @@
...
+#define ICE_ADD_DEFAULT_PLUGIN_FACTORIES_H
+
+#include "Ice/PluginFactory.h"
+
+#include <vector>
+
+namespace IceInternal
+{
+    /// Adds the default plug-in factories, for IceSSL, IceTCP, plus, depending on the build, IceUDP and IceWS.
+    void addDefaultPluginFactories(std::vector<Ice::PluginFactory>&);
```
bernardnormier, line 13: Replaces `RegisterPluginsInit`.

cpp/src/Ice/Makefile.mk
```diff
@@ -15,7 +15,8 @@ Ice_extra_sources       := $(filter-out src/Ice/SSL/SecureTransport%.cpp src/Ice
 endif
 
 Ice_excludes            = src/Ice/DLLMain.cpp
-Ice[shared]_excludes    = src/Ice/RegisterPluginsInit_min.cpp
+Ice[shared]_excludes    = src/Ice/AddDefaultPluginFactories_min.cpp
+Ice[static]_excludes    = src/Ice/AddDefaultPluginFactories_all.cpp
```
bernardnormier, line 19: Before this PR, we were including both in static builds (like the XC framework lib), and apparently picking _all.

cpp/src/Ice/PluginManagerI.cpp
```diff
@@ -344,27 +305,7 @@ Ice::PluginManagerI::loadPlugin(const string& name, const string& pluginSpec, St
         cmdArgs = properties->parseCommandLineOptions(name, cmdArgs);
     }
 
-    PluginFactory factory = nullptr;
```
bernardnormier, line 347: The logic remains largely the same.

Before: factoryFunc comes from global map.
Now: factoryFunc is provided by initData

cpp/src/Ice/PluginManagerI.h
```diff
@@ -5,6 +5,7 @@
 
 #include "Ice/BuiltinSequences.h"
 #include "Ice/CommunicatorF.h"
+#include "Ice/Initialize.h"
```
bernardnormier, line 8: Should be PluginFactory.h

cpp/src/Ice/msbuild/ice/ice.vcxproj
```diff
@@ -591,7 +591,7 @@
     <ClCompile Include="..\..\Proxy.cpp" />
     <ClCompile Include="..\..\Reference.cpp" />
     <ClCompile Include="..\..\ReferenceFactory.cpp" />
-    <ClCompile Include="..\..\RegisterPluginsInit_all.cpp" />
+    <ClCompile Include="..\..\addDefaultPluginFactories_all.cpp" />
```
bernardnormier, line 594: Should be capitalized.

cpp/src/Ice/msbuild/ice/ice.vcxproj.filters
```diff
@@ -402,7 +402,7 @@
     <ClCompile Include="..\..\ReferenceFactory.cpp">
       <Filter>Source Files</Filter>
     </ClCompile>
-    <ClCompile Include="..\..\RegisterPluginsInit_all.cpp">
+    <ClCompile Include="..\..\addDefaultPluginFactories_all.cpp">
```
bernardnormier, line 405: Ditto

cpp/test/Common/TestHelper.cpp
```diff
@@ -347,3 +340,30 @@ Test::TestHelper::shutdownOnInterrupt()
     _ctrlCHandler->setCallback(shutdownOnInterruptCallback);
 }
 #endif
+
+void
+Test::TestHelper::addDefaultPluginFactories(std::vector<Ice::PluginFactory>& factories) const
+{
+    // TODO: we should not register these plug-ins all the time.
```
bernardnormier, line 347: The difficulty is we no longer have an ICE_STATIC_LIBS or similar define.

It would be much better to check a runtime value.

swift/src/IceImpl/IceUtil.mm
```diff
@@ -91,6 +70,27 @@ + (ICECommunicator*)initialize:(NSArray*)swiftArgs
         initData.logger = std::make_shared<LoggerWrapperI>(logger);
     }
 
+    // Add plug-in factories.
```
bernardnormier, line 73: The logic remains the same.

cpp/src/Ice/PluginManagerI.cpp
```diff
@@ -292,18 +242,29 @@ Ice::PluginManagerI::loadPlugins(int& argc, const char* argv[])
...
-Ice::PluginManagerI::loadPlugin(const string& name, const string& pluginSpec, StringSeq& cmdArgs)
+Ice::PluginManagerI::loadPlugin(
+    PluginFactoryFunc factoryFunc,
+    const string& name,
+    const string& pluginSpec,
+    StringSeq& cmdArgs)
 {
     assert(_communicator);
 
+    if (findPlugin(name))
```
bernardnormier, line 261: Need to add this check in C# and Java too.

cpp/src/Ice/AddDefaultPluginFactories_all.cpp
```diff
@@ -0,0 +1,40 @@
...
+    Ice::Plugin* createIceSSL(const Ice::CommunicatorPtr&, const std::string&, const Ice::StringSeq&);
+}
+
+void
+IceInternal::addDefaultPluginFactories(vector<Ice::PluginFactory>& pluginFactories)
+{
+    // These built-in plug-ins are always available and not returned by a Ice::namePluginFactory function.
+    vector<Ice::PluginFactory> defaultPluginFactories{{"IceTCP", createIceTCP}, {"IceSSL", createIceSSL}};
+
+    if (find_if(
```
pepone, line 23: What about replacing with none_of

https://en.cppreference.com/w/cpp/algorithm/all_any_none_of

cpp/test/IceDiscovery/simple/Client.cpp
```diff
@@ -1,10 +1,14 @@
 // Copyright (c) ZeroC, Inc.
 
 #include "Ice/Ice.h"
-#include "Ice/RegisterPlugins.h"
 #include "Test.h"
 #include "TestHelper.h"
 
+// Link with IceDiscovery on Windows.
+#if defined(_MSC_VER) && !defined(ICE_DISABLE_PRAGMA_COMMENT)
```
pepone, line 8: do we use ICE_DISABLE_PRAGMA_COMMENT with the tests?

python/setup.py
```diff
@@ -96,7 +96,7 @@ def filter_source(filename):
     if "ios/" in filename:
         return False
 
-    if "RegisterPluginsInit_min" in filename:
+    if "addDefaultPluginFactories_min" in filename:
```
pepone, line 99: Didn't you capitalize the file name?
```suggestion
    if "AddDefaultPluginFactories_min" in filename:
```

ruby/extconf.rb
```diff
@@ -51,7 +51,7 @@
     $srcs << f
 end
 
-$excluded = ['DLLMain.cpp', 'Main.cpp', 'RegisterPluginsInit_min.cpp']
+$excluded = ['DLLMain.cpp', 'Main.cpp', 'addDefaultPluginFactories_min.cpp']
```
pepone, line 54:
```suggestion
$excluded = ['DLLMain.cpp', 'Main.cpp', 'AddDefaultPluginFactories_min.cpp']
```

cpp/test/IceSSL/configuration/AllTests.cpp
```diff
@@ -947,6 +950,7 @@ allTests(Test::TestHelper* helper, const string& /*testDir*/, bool p12)
         //
 
         InitializationData initData;
+        helper->addDefaultPluginFactories(initData.pluginFactories);
```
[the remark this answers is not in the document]
bernardnormier: It's for ws[s].

cpp/test/IceGrid/simple/AllTests.cpp
```diff
@@ -44,6 +44,11 @@ allTests(Test::TestHelper* helper)
         // registries and make sure locator requests are forwarded.
         //
         Ice::InitializationData initData;
+        initData.pluginFactories = {
+            Ice::udpPluginFactory(),
```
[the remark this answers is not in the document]
bernardnormier: It's for when we link with a "_min" build, such as a static C++ build.

cpp/test/Ice/background/Server.cpp
```diff
@@ -112,7 +109,11 @@ Server::run(int argc, char** argv)
     string defaultProtocol = properties->getIceProperty("Ice.Default.Protocol");
     properties->setProperty("Ice.Default.Protocol", "test-" + defaultProtocol);
 
-    CommunicatorHolder communicator = initialize(argc, argv, properties);
+    Ice::InitializationData initData;
+    initData.properties = properties;
+    initData.pluginFactories = {{"Test", createTestTransport}};
```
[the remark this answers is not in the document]
bernardnormier: No, and that's an issue for many tests. The configuration is now redundant.

Should we:
a) remove the configuration, or
b) add a new flag to the TestHelper to select between the initData.pluginFactories (helper->staticLibs() == true) and the traditional configuration (helper->staticLibs() == false)?

The helper could get it from a property we set for `--config static`, like `--Test.StaticLibs`.

cpp/test/IceDiscovery/simple/Client.cpp
bernardnormier: I don't think we do, it's for consistency. Whenever we use pragma comment, we also use this macro.

cpp/test/Ice/background/Server.cpp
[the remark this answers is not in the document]
bernardnormier: Or simpler, we provide a function that tells us if Ice was built with _all or _min. Like:

```cpp
namespace IceInternal
{
    ICE_API bool isMinBuild();
}
```

cpp/src/Ice/AddDefaultPluginFactories_all.cpp
bernardnormier: Fixed as suggested. There are a few other places where we should replace find_if.

cpp/test/Ice/background/Server.cpp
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

cpp/test/Common/TestHelper.cpp
bernardnormier: Fixed.

cpp/src/Ice/PluginManagerI.cpp
bernardnormier: Fixed C#.

cpp/src/Ice/AddDefaultPluginFactories_min.cpp
```diff
@@ -0,0 +1,23 @@
...
+}
+
+void
+IceInternal::addDefaultPluginFactories(vector<Ice::PluginFactory>& pluginFactories)
+{
+    pluginFactories.insert(pluginFactories.begin(), {{"IceTCP", createIceTCP}, {"IceSSL", createIceSSL}});
+}
+
+bool
+IceInternal::isMinBuild()
```
pepone, line 20: Can we just implement this in the tests?

```
bool Test::isMinBuild()
{
   vector<Ice::PluginFactory> pluginFactories;
   addDefaultPluginFactories(pluginFactories);
   return any_of(
      pluginFactories.begin(),
      pluginFactories.end(), 
      [](const auto& factory) { 
         return factory.pluginName == "IceUDP" || factory.pluginName == "IceWS"; })
}
```

cpp/test/IceSSL/configuration/AllTests.cpp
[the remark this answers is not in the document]
pepone: Unless I mistaken this test only runs with SSL transport.

cpp/src/Ice/AddDefaultPluginFactories_min.cpp
bernardnormier: `addDefaultPluginFactories` is an internal, non-exported function. I find a separate exported function cleaner.

cpp/test/IceSSL/configuration/AllTests.cpp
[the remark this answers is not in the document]
bernardnormier: You're right: even if you run IceSSL/configuration with `--protocol ws` or wss, it doesn't use ws.

I reverted most of my changes to this test.

cpp/include/Ice/PluginFactory.h
```diff
@@ -0,0 +1,104 @@
...
+#include "Ice/BuiltinSequences.h"
+#include "Plugin.h"
+
+#include <string>
+
+namespace Ice
+{
+    class Plugin;
+
+    /// A plug-in factory function is a C function responsible for creating an Ice plug-in.
```
pepone, line 17: I'm not sure about this "C function" part. I guess this is only relevant for dynamic loaded plug-ins.

cpp/include/Ice/PluginFactory.h
```diff
@@ -0,0 +1,104 @@
...
+
+#ifndef ICE_PLUGIN_FACTORY_DECLSPEC_IMPORT
+#    define ICE_PLUGIN_FACTORY_DECLSPEC_IMPORT ICE_DECLSPEC_IMPORT
+#endif
+
+    //
+    // Checking for the API_EXPORTS macro is necessary to prevent inconsistent DLL linkage errors on Windows.
+    //
+
+#if !defined(ICE_API_EXPORTS) || defined(ICE_DOXYGEN)
```
pepone, line 45: Can you explain the logic for the export/import macros usage here?

cpp/src/Ice/AddDefaultPluginFactories_min.cpp
pepone: Ok, I didn't notice we were not exporting addDefaultPluginFactories

pepone (approved):

cpp/include/Ice/PluginFactory.h
bernardnormier: All the plug-in create function are extern "C" in C++, see:
https://doc.zeroc.com/ice/3.7/communicator-and-other-core-local-features/plug-in-facility/plug-in-api#id-.PluginAPIv3.7-C++Plug-inFactory

cpp/include/Ice/PluginFactory.h
bernardnormier: It's only relevant on Windows, so maybe we should rework these since we don't support "_min" builds on Windows, and as a result, udp/ws are always built-in.

Let's take IceDiscovery on Windows. We want applications to be able to "import" IceDiscovery (with ICE_DECLSPEC_IMPORT), however IceDiscovery itself should not import this function (it needs to export it).

Also, `ICE_DISCOVERY_API` is not defined when including Ice/Ice.h. That's why we rely on this roundabout `!defined(ICE_DISCOVERY_API_EXPORTS)`.

bernardnormier: For the built-in Ice plug-ins, it would be cleaner to just use ICE_API.

cpp/include/Ice/PluginFactory.h
```diff
@@ -0,0 +1,104 @@
...
+#    ifndef ICEBT_API_EXPORTS
+    /// Returns the factory for the Bluetooth transport plug-in, IceBT.
+    /// @return The factory for the IceBT plug-in.
+    /// @see InitializationData::plugins
+    ICE_PLUGIN_FACTORY_DECLSPEC_IMPORT PluginFactory btPluginFactory();
+#    endif
+#endif
+
+#if (defined(__APPLE__) && TARGET_OS_IPHONE != 0) || defined(ICE_DOXYGEN)
+#    ifndef ICEIAP_API_EXPORTS
```
bernardnormier, line 85: We should check this macro:

```
C:\Users\bernard\source\repos\ice>rg ICEIAP_API_EXPORTS
cpp\src\Ice\ios\iAPEndpointI.mm
27:#        if defined(ICEIAP_API_EXPORTS)

cpp\include\Ice\RegisterPlugins.h
71:#    ifndef ICEIAP_API_EXPORTS
```

cpp/src/Ice/AddDefaultPluginFactories.h
```diff
@@ -0,0 +1,16 @@
...
+#ifndef ICE_ADD_DEFAULT_PLUGIN_FACTORIES_H
+#define ICE_ADD_DEFAULT_PLUGIN_FACTORIES_H
+
+#include "Ice/PluginFactory.h"
+
+#include <vector>
+
+namespace IceInternal
+{
+    /// Adds the default plug-in factories, for IceSSL, IceTCP, plus, depending on the build, IceUDP and IceWS.
```
externl, line 12:
```suggestion
    /// Adds the default plug-in factories for IceSSL, IceTCP, and, depending on the build, IceUDP and IceWS.
```

externl (approved):
Looks good!

PR #3845, opened 2025-04-14
Python exception handling cleanup

pepone:
A small cleanup for handling exceptions from Ice APIs.

- Removed exception handler for calls to noexcept methods
- Add missing exception handlers for calls to no noexcept methods


bernardnormier (approved):
Looks good.

This noexcept vs throwing is largely determined by the return type in C++:
 - std::string, means memory allocation, means throwing
 - on the other hand, const& and shared_ptr copy ctor are noexcept

externl (approved):

Issue #3368, opened 2025-01-15
Deprecate Metadata for Generating Module Prefixes in Mapped Code


InsertCreativityHere:
We have a handful of metadata which are about re-mapping top-level modules:
`java:package`, `cs:namespace`, `swift:module`, maybe even more in other languages.

Confusingly, when these are applied to a module, it doesn't change the module's name itself.
Instead it just changed which scope the module's definition will be mapped into.
```
["cs:namespace:Something"]
module Hello {...}
```
Will cause us to generate code into the `Something.Hello` namespace.

----

Now that we've settled on introducing other metadata for remapping identifiers, these metadata are redundant, and should be deprecated.

bernardnormier:
I think we should keep these metadata as file metadata, and deprecate/remove the module metadata.

InsertCreativityHere:
This applied to `cs:namespace`, `python:package`, `java:package`, and `swift:module`.
The last one (Java) was done in #3931.

PR #3846, opened 2025-04-14
Replace `swift:module` with `swift:identifier`

InsertCreativityHere:
This PR follows in the footsteps of #3833 (which handled `cs:namespace` and `python:package`), working towards #3368.

As-of this PR, the Slice compiler still supports `swift:module`, but it's deprecated, and I have removed it's use from all of our Slice files (except the scope test), replacing it with `swift:identifier`.

Unlike the other module remapping metadata, `swift:module` supported a special 2nd argument, for adding a prefix to all a module's contents. In some tests, I manually emulated this (by using `swift:identifier` on the module's contents), in other cases (like for IceMX), I just fully dropped the prefix.

### DIFF
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -541,7 +541,7 @@ SwiftGenerator::validateMetadata(const UnitPtr& u)
                 if (contained->hasMetadata("swift:identifier"))
                 {
                     return "the 'swift:module' metadata cannot be used alongside 'swift:identifier' - both change the "
-                           "mapped name of this module";
+                           "name of the mapped module";
                 }
             }
             return nullopt;
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -540,12 +540,12 @@ SwiftGenerator::validateMetadata(const UnitPtr& u)
             {
                 if (contained->hasMetadata("swift:identifier"))
                 {
-                    return "the 'swift:module' metadata cannot be used alongside 'swift:identifier' - both change the mapped name of this module";
+                    return "the 'swift:module' metadata cannot be used alongside 'swift:identifier' - both change the "
+                           "mapped name of this module";
                 }
             }
             return nullopt;
-        }
-    };
+        }};
     knownMetadata.emplace("swift:module", moduleInfo);
 
     // Pass this information off to the parser's metadata validation logic.
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -531,6 +531,20 @@ SwiftGenerator::validateMetadata(const UnitPtr& u)
         .validOn = {typeid(Module)},
         // Even though it's really 'module:prefix' the validator sees this as a single argument since there's no commas.
         .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
+        .extraValidation = [](const MetadataPtr& metadata, const SyntaxTreeBasePtr& p) -> optional<string>
+        {
+            const string msg = "'swift:module' is deprecated; use 'swift:identifier' to remap modules instead";
+            p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
+
+            if (auto contained = dynamic_pointer_cast<Contained>(p))
+            {
+                if (contained->hasMetadata("swift:identifier"))
+                {
+                    return "the 'swift:module' metadata cannot be used alongside 'swift:identifier' - both change the mapped name of this module";
+                }
+            }
+            return nullopt;
+        }
     };
     knownMetadata.emplace("swift:module", moduleInfo);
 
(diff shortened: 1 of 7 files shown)


cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -531,7 +531,21 @@ SwiftGenerator::validateMetadata(const UnitPtr& u)
         .validOn = {typeid(Module)},
         // Even though it's really 'module:prefix' the validator sees this as a single argument since there's no commas.
         .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
-    };
+        .extraValidation = [](const MetadataPtr& metadata, const SyntaxTreeBasePtr& p) -> optional<string>
+        {
+            const string msg = "'swift:module' is deprecated; use 'swift:identifier' to remap modules instead";
+            p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
```
InsertCreativityHere, line 537: We always emit a deprecation warning, but still return `nullopt` to signal that this metadata is fine...

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -531,7 +531,21 @@ SwiftGenerator::validateMetadata(const UnitPtr& u)
...
+
+            if (auto contained = dynamic_pointer_cast<Contained>(p))
+            {
+                if (contained->hasMetadata("swift:identifier"))
+                {
+                    return "the 'swift:module' metadata cannot be used alongside 'swift:identifier' - both change the "
+                           "mapped name of this module";
+                }
```
InsertCreativityHere, line 545: ... unless a user used 'swift:module' and 'swift:identifier' together, in which case we return an error message.
The MetadataValidation machinery will handle the rest for us.

slice/Glacier2/Metrics.ice
```diff
@@ -14,7 +14,7 @@
 
 #include "Ice/Metrics.ice"
 
-["swift:module:Glacier2:MX"]
```
InsertCreativityHere, line 17: We decided to just drop the `MX` prefix from these modules's contents.

Without the 2nd argument, `swift:module:Foo` and `swift:identifier:Foo` are equivalent.

swift/test/Ice/scope/Test.ice
```diff
@@ -2,6 +2,8 @@
 
 #pragma once
 
+[["suppress-warning:deprecated"]] // for 'swift:module'
```
InsertCreativityHere, line 5: It's good to still have some tests to exercise `swift:module`.

swift/test/Ice/operations/Test.ice
```diff
@@ -415,15 +415,12 @@ module Test
...
-["swift:module:Test:Test2"]
+["swift:identifier:Test"]
 module Test2
 {
-    /**
-    *
-    * Makes sure that proxy operations are correctly generated when extending an interface from
-    * a different module (ICE-7639).
-    *
-    **/
```
InsertCreativityHere, line 426: Manually applied the prefix to the module's contents here.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -1604,8 +1618,6 @@ SwiftGenerator::writeDispatchOperation(::IceInternal::Output& out, const Operati
     const ParameterList inParams = op->inParameters();
     const bool returnsAnyValues = op->returnsAnyValues();
 
-    const string swiftModule = getSwiftModule(getTopLevelModule(op));
```
externl, line 1607: Unused I assume

externl (approved):

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere: Indeed, a totally dead local variable.
Stumbled across it while reviewing all the calls to `getSwiftModule` (the function that checked for `swift:module`)

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -531,7 +531,21 @@ SwiftGenerator::validateMetadata(const UnitPtr& u)
...
+        {
+            const string msg = "'swift:module' is deprecated; use 'swift:identifier' to remap modules instead";
+            p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
+
+            if (auto contained = dynamic_pointer_cast<Contained>(p))
+            {
+                if (contained->hasMetadata("swift:identifier"))
+                {
+                    return "the 'swift:module' metadata cannot be used alongside 'swift:identifier' - both change the "
+                           "mapped name of this module";
```
bernardnormier, line 544: "both change the name of the mapped module" sounds simpler

bernardnormier (approved):

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere, line 544:
```suggestion
                    return "the 'swift:module' metadata cannot be used alongside 'swift:identifier' - both change the "
                           "name of the mapped module";
```

InsertCreativityHere: Changed as suggested.

PR #3851, opened 2025-04-14
Remove static cppflags

bernardnormier:
This PR removes `-DICE_STATIC_LIBS` and `-DICE_HAS_BT` that are not checked anywhere.

### DIFF
--- a/cpp/config/Make.rules
+++ b/cpp/config/Make.rules
@@ -97,10 +97,6 @@ define create-cpp-test-project
 $1_srcext               := cpp
 $1_dependencies         := $$(or $$($1_dependencies),TestCommon Ice)
 
-# Also link with IceBT (Debian/Ubuntu/Yocto) when compiling the project with the static configuration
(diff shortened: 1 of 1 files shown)


cpp/config/Make.rules
```diff
@@ -97,10 +97,6 @@ define create-cpp-test-project
 $1_srcext               := cpp
 $1_dependencies         := $$(or $$($1_dependencies),TestCommon Ice)
 
-# Also link with IceBT (Debian/Ubuntu/Yocto) when compiling the project with the static configuration
-$1[static]_cppflags     += -DICE_STATIC_LIBS $(if $(IceBT_system_libs),-DICE_HAS_BT)
-$1[static]_dependencies := $(if $(IceBT_system_libs),IceBT)
```
bernardnormier, line 102: It's not clear why we link static builds with IceBT, especially since we don't call `btPluginFactory()` anywhere.

pepone (approved):

externl (approved):

Issue #2998, opened 2024-10-29
doc-comment validation for Slice operations


bernardnormier:
As far as I can tell, there is no doc-comment validation in the Slice library, which is rather unexpected.

I think:
 - it should be ok (= no warning/error) to define an operation with no doc-comment at all
 - if you add a doc-comment to an operation:
   - incorrect tag/return/exception should result in an error. For example, a typo in a param name or  an`@param` when there is no parameter
   - `@deprecated` when the operation is not marked deprecated should result in an error
   - missing doc for a param/return/exception should result in a warning

bernardnormier:
```slice
/// I'm not going to document my parameter!
void op(string s);
```

Since there is a doc-comment and your forgot to document `s`, you should get a warning: `missing @param s`.

Now, if you add `@param foo` and there is no foo parameter, you should get an error, not a warning.

>  if a user puts @param myParam with no text after it?

I would not issue a warning or error if there is no text after a `@param xxx`, assuming xxx is an actual parameter of the operation.

InsertCreativityHere:
I don't think we should complain if a user has a doc-comment, but doesn't use every single possible tag in it.
In some languages, it's totally normal not to bother documenting "obvious" parameters, to avoid pointless doc comments that you see all the time in C# and Java:
```
/// This operation sorts a dictionary.
/// @param d: the dictionary to sort.
sortDictionary(d: MyDictionary);
```

Arguably, I think that ommitting the `@param` here would actually _improve_ readability.

----

I agree that we should report inaccuracies though:
1) A user puts `@return` on a void operation.
2) A user put `@param xxx`, but there isn't a corresponding parameter for it.
3) A user put `@throws type` but the types isn't in their exception list.

----

And I'm on the fence about:

> @deprecated when the operation is not marked deprecated should result in an error

But I'm going to open an issue to discuss this, because I think there's a bigger 'problem' here.

PR #3852, opened 2025-04-14
Finish Operation Doc Comment Validation

InsertCreativityHere:
This PR finishes implementing #2998.

Now the parser will:
- reject `@param foo` if `foo` isn't the name of an operation parameter
- reject `@return` if it's applied to a `void `operation
- reject `@throws Foo` if `Foo` isn't in it's exception specific (or derived from an exception that is)

We also now reject duplicate tags.
So a comment can't have more than one `@deprecated`, or more than one `@return`.
And for 'named' tags, you can't use the same name more than once either.
```
@param foo this is totally fine (if you have a parameter named `foo`)
@param Foo this will be rejected, because you already did `@param foo`.
@param bar totally fine because it's a different name.
```

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -586,7 +586,7 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
                     { return ex->isBaseOf(exceptionTarget) || ex->scoped() == exceptionTarget->scoped(); };
                     if (std::find_if(exceptionSpec.begin(), exceptionSpec.end(), exceptionCheck) == exceptionSpec.end())
                     {
-                        const string msg = "'" + actualTag + " " + name + "': this exception is not listed in" +
+                        const string msg = "'" + actualTag + " " + name + "': this exception is not listed in " +
                                            "(or a sub-exception of) this operation's exception specification";
                         p->unit()->warning(p->file(), p->line(), InvalidComment, msg);
                     }
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -536,7 +536,8 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
                 const auto paramNameCheck = [&name](const ParameterPtr& p) { return p->name() == name; };
                 if (std::find_if(params.begin(), params.end(), paramNameCheck) == params.end())
                 {
-                    const string msg = "'" + paramTag + " " + name + "' does not correspond to any parameter in operation '" + p->name() + "'";
+                    const string msg = "'" + paramTag + " " + name +
+                                       "' does not correspond to any parameter in operation '" + p->name() + "'";
                     p->unit()->warning(p->file(), p->line(), InvalidComment, msg);
                 }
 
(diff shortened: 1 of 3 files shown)


cpp/src/Slice/Parser.cpp
```diff
@@ -522,47 +522,88 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
...
+                {
+                    comment._parameters[name] = {};
+                    currentSection = &(comment._parameters[name]);
+                }
             }
         }
-        else if (parseNamedCommentLine(l, exceptionTag, name, lineText))
+        else if (
+            parseNamedCommentLine(l, throwsTag, name, lineText) ||
+            parseNamedCommentLine(l, exceptionTag, name, lineText))
```
InsertCreativityHere, line 560: We support `@exception` as an alias for `@throws`, and used to parse/validate them separately.
I combined them both into a single block to avoid code-duplication.

cpp/src/Slice/Parser.cpp
```diff
@@ -522,47 +522,88 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
...
-                currentSection = nullptr;
-            }
-            else
-            {
-                comment._exceptions[name] = {};
-                currentSection = &(comment._exceptions[name]);
+                // Check that the '@param <name>' corresponds to an actual parameter in the operation.
+                const ParameterList params = operationTarget->parameters();
+                const auto paramNameCheck = [&name](const ParameterPtr& param) { return param->name() == name; };
+                if (std::find_if(params.begin(), params.end(), paramNameCheck) == params.end())
```
bernardnormier, line 537: You should use std::none_of. I would also put the lambda inline, no need to give it a name.
https://en.cppreference.com/w/cpp/algorithm/all_any_none_of.

cpp/src/Slice/Parser.cpp
```diff
@@ -522,47 +522,88 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
 
         if (parseNamedCommentLine(l, paramTag, name, lineText))
```
bernardnormier, line 523: It would be really nice to avoid naming anying `l`!

cpp/src/Slice/Parser.cpp
```diff
@@ -522,47 +522,88 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
...
+                else
+                {
+                    // ... and matches one of the exceptions in the operation's specification.
+                    const ExceptionList exceptionSpec = operationTarget->throws();
+                    const auto exceptionCheck = [&exceptionTarget](const ExceptionPtr& ex)
+                    { return ex->isBaseOf(exceptionTarget) || ex->scoped() == exceptionTarget->scoped(); };
+                    if (std::find_if(exceptionSpec.begin(), exceptionSpec.end(), exceptionCheck) == exceptionSpec.end())
+                    {
+                        const string msg = "'" + actualTag + " " + name + "': this exception is not listed in " +
+                                           "(or a sub-exception of) this operation's exception specification";
```
bernardnormier, line 590: The sentence is not grammatically correct.

cpp/src/Slice/Parser.cpp
```diff
@@ -522,47 +522,88 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
...
+                {
+                    const string msg = "'" + paramTag + " " + name +
+                                       "' does not correspond to any parameter in operation '" + p->name() + "'";
+                    p->unit()->warning(p->file(), p->line(), InvalidComment, msg);
+                }
+
+                // Check if this is a duplicate tag. If it is, ignore it and issue a warning.
+                if (comment._parameters.count(name) != 0)
+                {
+                    const string msg = "ignoring duplicate doc-comment tag: '" + paramTag + " " + name + "'";
```
externl, line 547:
```suggestion
                    const string msg = "duplicate doc-comment tag: '" + paramTag + " " + name + "'";
```

cpp/src/Slice/Parser.cpp
externl, line 590: Maybe something like:

const string msg = "'" + actualTag + " " + name +
    "': this exception is neither listed in nor derived from any exception in the operation's exception specification";

externl (approved):

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Very nice!
I updated my `find_if`s accordingly.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Fixed. This was just the (unchanged) original name.
I renamed it to `line`.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Fixed, using externl's suggestion.

InsertCreativityHere: I have taken this.

PR #3854, opened 2025-04-14
Add Missing BOM Back to the `Slice/utf8BOM` Test

### LINKED ISSUE #3387: UTF8 BOM Seems to be Missing from UTF8BOM Test
We have a test to ensure that the Slice compilers can correctly tolerate the presence of Byte-Order-Markers in Slice files,
https://github.com/zeroc-ice/ice/blob/main/cpp/test/Slice/utf8BOM/Test.ice

But, opening this file up in a hex-editor, there is no BOM at the beginning of the file.
I guess it was deleted by an auto-formatter or one of the times we changed the copyright headers.

We should add the BOM back, and change the test so it fails if the BOM was deleted somehow.

InsertCreativityHere:
This PR fixes #3387.
The Slice parser has a special grammar rule for matching and handling UTF BOMs in Slice files. And we even have a test which exists to exercise this part of the parser. Except that the Slice file we use for testing no longer has a BOM in it...

My guess is it was accidentally removed by one of our automated copyright header updates.

So, this PR adds the BOM back to the file, and adds a Python test script to the directory which:
1) Checks that the BOM is present (in case it gets deleted by accident again)
2) Runs `slice2cpp` over the file and makes sure it succeeds with no warnings/errors

----

Previously, we were just letting msbuild compile this Slice file during the test build.
Now that we call `slice2cpp` directly from the test script, having `msbuild` build it is redundant.
So I deleted all the `msbuild` files for this test directory. Which saves quite a couple lines.


cpp/test/Slice/utf8BOM/Test.ice
```diff
@@ -1,10 +1,9 @@
-// Copyright (c) ZeroC, Inc.
+﻿// Copyright (c) ZeroC, Inc.
```
InsertCreativityHere, line 1: You can't see it, but there is a BOM at the start of this file.

cpp/test/Slice/utf8BOM/test.py
```diff
@@ -0,0 +1,42 @@
...
+        if os.path.exists(outdir):
+            shutil.rmtree(outdir)
+        os.mkdir(outdir)
+
+        try:
+            # Make sure the BOM marker wasn't accidentally deleted by over-zealous reformatting.
+            current.write("Checking for UTF8-BOM marker... ")
+            with open(test_file_name, 'rb') as test_file:
+                BOM = test_file.read(3)
+                if (BOM[0] != 0xEF or BOM[1] != 0xBB or BOM[2] != 0xBF):
```
externl, line 25: Another option is something like:
```suggestion
                if BOM[:3] != b'\xEF\xBB\xBF':
```

externl (approved):

pepone (approved):

cpp/test/Slice/utf8BOM/test.py
InsertCreativityHere: Fixed!

InsertCreativityHere:
After learning about this bug with MCPP on Ubuntu (see #3940), I updated the BOM test.
So now it has 2 BOMs, a valid one at the beginning, and an invalid one in the middle of the file.
On Windows/MacOS we expect only 1 error (for the middle one), and on linux we tolerate a 2nd error (for the beginning one).

I also added specific BOM error handling to the Slice scanner. So now it will say:
```
encountered unexpected UTF-8 BOM in input; BOMs can only appear at the beginning of files
```
instead of:
```
illegal input character: '\357'
illegal input character: '\273'
illegal input character: '\277'
```

Lastly... we were apparently testing BOMs in the `IllegalIdentifier` test too.
I removed all this because BOMs are unrelated to 'identifier's, illegal or otherwise.

cpp/src/Slice/Scanner.cpp
InsertCreativityHere, line 1: Generated file. Should probably ignore.

InsertCreativityHere, line 1: Generated file, should ignore.

cpp/test/Slice/errorDetection/IllegalIdentifier.ice
```diff
@@ -1,20 +1,8 @@
-
-// NOTE: Keep the initial blank line. This is necessary to avoid a problem
-// with system-provided mcpp versions that don't include ZeroC's BOM patch.
```
InsertCreativityHere, line 3: This comment is from 7 years ago. So I guess this has been a known issue for a while... And we even knew that having a comment on the first line triggered the re-ordering bug.

Adding this blank line at the top feels like a lame fix though.

cpp/test/Slice/utf8BOM/test.py
```diff
@@ -0,0 +1,52 @@
...
+
+from Util import ClientTestCase, SliceTranslator, TestSuite, platform, Linux
+
+
+class SliceUtf8BomTestCase(ClientTestCase):
+    def runClientSide(self, current):
+        test_file_name = "Test.ice"
+        slice2cpp = SliceTranslator("slice2cpp", quiet=True)
+
+        # Temporary output directory for 'slice2cpp to use.
```
pepone, line 14: Why do we need to create the output directory? I was thinking slice2cpp shouldn't generate any code here as the Slice file has errors.

pepone (approved):

cpp/src/Slice/Scanner.l
```diff
@@ -529,8 +529,18 @@ floating_literal    (({fractional_constant}{exponent_part}?)|((\+|-)?{dec}+{expo
 
   /* ========== Others ========== */
 
-  /* Matches and consumes a BOM, but only when the scanner has just started scanning a new file. */
-<INITIAL>{bom} {}
+  /* Handles UTF-8 BOMs.
+   * The scanner has a special 'INITIAL' state, which it enters whenever it hits the start of a file.
+   * This occurrs when parsing initially begins, but also when we see the start of an '#include'd file.
```
bernardnormier, line 534: occurs

bernardnormier (approved):

cpp/src/Slice/Scanner.l
InsertCreativityHere: Good catch.
Fixed!

cpp/test/Slice/utf8BOM/test.py
InsertCreativityHere: I added it just to make sure that the tests will clean up after themselves no matter what.
Although you're right that under normal circumstances, nothing will be generated.

We also create this tmp folder in the `errorDetection` tests, which is in the same situation: https://github.com/zeroc-ice/ice/blob/86dfedab1473ae2d03d65309ac78b1400229d86a/cpp/test/Slice/errorDetection/test.py#L16-L19

InsertCreativityHere, line 14:
```suggestion
        # Temporary output directory for 'slice2cpp' to use.
```

PR #3856, opened 2025-04-15
Changing rewriteDryRun to *not* fail by default. 

ayoitslilith:
This PR inverts the behavior of `rewriteDryRun` in regards to failing. Now, `gradlew rewriteDryRun` will **not** cause a build failure. Instead, to cause it to fail while building, one must invoke `-PenableDryRunFail`. CI builds should still fail when OpenRewrite would make changes.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -14,8 +14,14 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Setup C++
+        uses: ./.github/actions/setup-cpp
+
       - name: Setup Java
         uses: ./.github/actions/setup-java
+
+      - name: Build slice2java
+        run: make -C cpp slice2java
 
       - name: Run rewriteDryRun with build failure enabled
         run: ./java/gradlew rewriteDryRun -PenableDryRunFail --project-dir ./java/
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -18,4 +18,4 @@ jobs:
         uses: ./.github/actions/setup-java
 
       - name: Run rewriteDryRun with build failure enabled
-        uses: ./java/gradlew rewriteDryRun -PenableDryRunFail
+        run: ./java/gradlew rewriteDryRun -PenableDryRunFail
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -24,4 +24,5 @@ jobs:
         run: make -C cpp slice2java
 
       - name: Run rewriteDryRun with build failure enabled
-        run: ./java/gradlew rewriteDryRun -PenableDryRunFail -PenableCops --project-dir ./java/
+        working-directory: java
+        run: ./java/gradlew rewriteDryRun -PenableCops -PfailOnDryRunResults
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-      
+
       - name: Setup Java
         uses: ./.github/actions/setup-java
 
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -11,6 +11,9 @@ jobs:
   rewriteCatch:
     runs-on: ubuntu-24.04
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      
       - name: Setup Java
         uses: ./.github/actions/setup-java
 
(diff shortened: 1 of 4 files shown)


ayoitslilith:
`-PenableCops` has been added as a possible argument. This argument **must** be passed in the terminal in order to use `checkstyle`, `rewriteRun`, or `rewriteDryRun`. This has been added as a means of preventing them from using extra time during a standard `gradlew build`, but `checkstyle` and `rewriteDryRun` are still invoked in the CI along with the new property.

pepone:
> -PenableCops has been added as a possible argument...

Is the build time an issue here? Otherwise I would not add extra arguments

.github/workflows/java.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: ./.github/actions/setup-cpp
+
+      - name: Setup Java
+        uses: ./.github/actions/setup-java
+
+      - name: Build slice2java
+        run: make -C cpp slice2java
+
+      - name: Run rewriteDryRun with build failure enabled
+        run: ./java/gradlew rewriteDryRun -PenableDryRunFail -PenableCops --project-dir ./java/
```
pepone, line 27: I would set `working-directory: java` and simplify the command to `./gradlew rewriteDryRun -PenableDryRunFail`

java/build.gradle
```diff
@@ -144,10 +150,22 @@ dependencies {
...
         rewrite.failOnDryRunResults = true
     }
-    finalizedBy rewriteDryRun
+}
+
+tasks.named('rewriteRun') {
+    task ->
+        if (!project.hasProperty("enableCops")) {
+            task.enabled = false
+        }
```
pepone, line 170: You can simplify it as:

```suggestion
        task.enabled = project.hasProperty("enableCops")
```

But I would prefer to get rid of this option, unless it makes the build too slow.

java/build.gradle
```diff
@@ -144,10 +150,22 @@ dependencies {
...
-    if (!project.hasProperty("dryRunFailSkip")) {
+// Use -PenableDryRunFail in the command line to enable failing the project during any operation with rewriteDryRun.
+// You need to use -PenableCops to be able to invoke the checkstyle, rewriteRun, and rewriteDryRun tasks.
+// The tasks still need to be manually called if nothing else invokes them.
+tasks.named('rewriteDryRun') {
+    task ->
+        if (!project.hasProperty("enableCops")) {
+            task.enabled = false
+        }
+    if (project.hasProperty("enableDryRunFail")) {
```
pepone, line 161: I would use the same property as open rewrite.

And them do something like:

```
rewrite.failOnDryRunResults = project.hasProperty("failOnDryRunResults") && project.property("failOnDryRunResults").toString().toBoolean()
```

ayoitslilith:
> Is the build time an issue here? Otherwise I would not add extra arguments

I was asked to do this by Bernard. Ultimately, it adds roughly ~20 seconds to the build time on the machine I am using, but I am unsure how much it would add on other computers.

bernardnormier:
There are two issues with the current situation on main:

- a default build and especially rebuild takes more time because we perform these checks all the time
- a default build will fail when the formatting is now perfect

The solution is to NOT perform either during a default build. Checking and reformatting should be other targets. Naturally CI must check and not reformat.

.github/workflows/java.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - name: Setup Java
+        uses: ./.github/actions/setup-java
+
+      - name: Build slice2java
+        run: make -C cpp slice2java
+
+      - name: Run rewriteDryRun with build failure enabled
+        working-directory: java
+        run: ./gradlew rewriteDryRun -PenableCops -PfailOnDryRunResults
```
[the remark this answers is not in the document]
ayoitslilith: I went ahead and changed this as requested. Admittedly, it had slipped my mind that enableCops/enableCodeAnalysis would also serve to prevent builds from failing.

java/build.gradle
```diff
@@ -144,10 +151,17 @@ dependencies {
...
-        rewrite.failOnDryRunResults = true
-    }
-    finalizedBy rewriteDryRun
+// You need to use -PenableCodeAnalysis to be able to invoke the checkstyle, rewriteRun, and rewriteDryRun tasks.
+// The tasks still need to be manually called if nothing else invokes them.
+tasks.named('rewriteDryRun') {
+    task ->
+        task.enabled = enableCodeAnalysis
+    project.logger.lifecycle(enableCodeAnalysis.toString())
+    project.logger.lifecycle(task.enabled.toString())
```
pepone, line 160: remove this logging before merge

pepone (approved):
Looks good!

PR #3857, opened 2025-04-15
Add pluginFactories to initData in Java

### LINKED ISSUE #3836: Add initData.plugins to "load" plugins

bernardnormier:
This PR adds pluginFactories to initData in Java.

See also #3842.
Fixes #3836.

### DIFF
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -252,6 +252,10 @@ Slice sequences, enabled by the `cpp:array` metadata directive, and a compatible
 earlier releases, this unaligned unmarshaling was turned on automatically on x86 and x64 CPUs, and turned off on all
 other CPUs.
 
+- Replaced the `Ice::registerPluginFactory` mechanism (with `Ice::registerIceUDP`, `Ice::registerIceDiscovery`, etc.)
+by plug-in factories on InitializationData. The corresponding plug-ins are created during communicator initialization.
+See `InitializationData::pluginFactories`.
+
 - Removed StringConverterPlugin and ThreadHookPlugin.
 
 ## C# Changes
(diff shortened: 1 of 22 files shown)


externl (approved):
Looks good to me!

pepone (approved):

PR #3858, opened 2025-04-15
Enforce fixed names for Ice plug-ins

### LINKED ISSUE #3844: Plug-in name
In Ice, the name of a plug-in is traditionally the name after `Ice.Plugin.` in configuration. This name can be anything.

For example, all the following entries load the IceDiscovery plug-in:
```
Ice.Plugin.Discovery = ...discovery plug-in spec
Ice.Plugin.IceDiscovery = ...discovery plug-in spec
Ice.Plugin.FooDiscovery = ...discovery plug-in spec
```

This flexibility allows us to load multiple instances of the same plug-in a communicator.

However, no plug-in provided with Ice takes advantage of this feature. All the plug-ins are singletons, and it's not meaningful to install multiple copies in the same communicator (with different names).  Worse, the plug-in implementations typically don't handle well such accidental installation, made possible by the flexible Ice.Plugin.name feature.

Related issue: #2798.

Even though plug-in names are flexible and user-specified, each plug-in has a default/preferred name. That's the name registered by the Ice::registerXxx function in Ice 3.7. With #3842, that's the `pluginName` provided by each proxy factory.

## Issue

Because of the plug-in name flexibility, you can fairly easily get into the situation where you load multiple instances of the same plug-in. For example, you specify:
```cpp
initData.pluginFactories = {Ice::discoveryPluginFactory()};
```
which uses the plug-in name "IceDiscovery", and you also configure:

```
Ice.Plugin.Discovery = ...discovery plug-in spec
```

In some cases, it may be hard to detect as we currently "reject" duplicate transport registrations through asserts.

https://github.com/zeroc-ice/ice/blob/52b0dad35395079ac9f39664bdd32f1cb843680f/cpp/src/Ice/EndpointFactoryManager.cpp#L38

## Proposal

For all singleton plug-ins (all Ice-provided plug-ins so far), the plug-in name as provided by the plug-in factory is the only allowed name. 

If you give the proxy factory function another name, you get a `PluginInitializationException`.

This way, each singleton plug-in can easily ensure only one instance is ever created in a communicator.

bernardnormier:
Fixes #3844

### DIFF
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -206,6 +206,9 @@ classDiagram
   New local exceptions:\
   ConnectionAbortedException, ConnectionClosedException, ParseException
 
+- The plug-ins provided by Ice have now fixed names: IceIAP, IceBT, IceUDP, IceWS, IceDiscovery, IceLocatorDiscovery.
+This fixed name is the only name you can use when loading/configuring such a plug-in with the Ice.Plugin.name property.
+
 ## Slice Language Changes
 
 - Removed local Slice. `local` is no longer a Slice keyword.
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -206,7 +206,7 @@ classDiagram
   New local exceptions:\
   ConnectionAbortedException, ConnectionClosedException, ParseException
 
-- The plug-ins provided by Ice have now fixed names: IceIAP, IceBT, IceUDP, IceWS, IceDiscovery, IceLocatorDiscovery.
+- The plug-ins provided by Ice now have fixed names: IceIAP, IceBT, IceUDP, IceWS, IceDiscovery, IceLocatorDiscovery.
 This fixed name is the only name you can use when loading/configuring such a plug-in with the Ice.Plugin.name property.
 
 ## Slice Language Changes
--- a/cpp/src/Ice/SSL/SSLEndpointI.cpp
+++ b/cpp/src/Ice/SSL/SSLEndpointI.cpp
@@ -25,10 +25,17 @@ extern "C"
 
         if (name != pluginName)
         {
+#ifdef _MSC_VER
+#    pragma warning(push)
+#    pragma warning(disable : 4297) // function assumed not to throw an exception but does
+#endif
             throw Ice::PluginInitializationException{
                 __FILE__,
                 __LINE__,
                 "the SSL plug-in must be named '" + pluginName + "'"};
+#ifdef _MSC_VER
+#    pragma warning(pop)
+#endif
         }
 
         IceInternal::InstancePtr instance = IceInternal::getInstance(communicator);
--- a/cpp/src/Ice/SSL/SSLEndpointI.cpp
+++ b/cpp/src/Ice/SSL/SSLEndpointI.cpp
@@ -12,10 +12,25 @@ using namespace std;
 using namespace Ice;
 using namespace Ice::SSL;
 
+namespace
+{
+    const char* const sslPluginName = "IceSSL";
+}
+
 extern "C"
 {
-    Plugin* createIceSSL(const CommunicatorPtr& communicator, const string&, const StringSeq&)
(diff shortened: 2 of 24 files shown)


cpp/src/Ice/SSL/SSLEndpointI.cpp
```diff
@@ -12,10 +12,32 @@ using namespace std;
 using namespace Ice;
 using namespace Ice::SSL;
 
+namespace
+{
+    const char* const sslPluginName = "IceSSL";
+}
+
 extern "C"
```
pepone, line 20: I wonder if there is any point in keeping `extern "C"`for SSL, which is always loaded from Ice library. The only use case I see is if you were loading IceSSL with a different name (not clear this worked) you will now get this error. Otherwise the plug-in is already registered and no need for extern C.

cpp/src/Ice/UdpEndpointI.cpp
```diff
@@ -15,10 +15,32 @@ using namespace std;
 using namespace Ice;
 using namespace IceInternal;
 
+namespace
+{
+    const char* const udpPluginName = "IceUDP";
+}
+
 extern "C"
```
pepone, line 23: Here this is only for static libraries, extern C seems not relevant. You either use a dynamic Ice build and this is loaded, or you use a static library and you have to use the factory to register it. I assuming extern "C" here is only relevant for dlsym and similar mechanism, where we want to avoid name mangling. That applies to TPC/WS too.

cpp/src/IceDiscovery/PluginI.cpp
```diff
@@ -9,19 +9,41 @@
...
-createIceDiscovery(const Ice::CommunicatorPtr& communicator, const string&, const Ice::StringSeq&)
+createIceDiscovery(const Ice::CommunicatorPtr& communicator, const string& name, const Ice::StringSeq&)
 {
+    string pluginName{discoveryPluginName};
+
+    if (name != pluginName)
+    {
+#ifdef _MSC_VER
+#    pragma warning(push)
+#    pragma warning(disable : 4297) // function assumed not to throw an exception but does
```
pepone, line 29: Maybe mention `VC++ assumes extern "C" function ....`

pepone (approved):

CHANGELOG-3.8.md
```diff
@@ -206,6 +206,9 @@ classDiagram
   New local exceptions:\
   ConnectionAbortedException, ConnectionClosedException, ParseException
 
+- The plug-ins provided by Ice have now fixed names: IceIAP, IceBT, IceUDP, IceWS, IceDiscovery, IceLocatorDiscovery.
```
externl, line 209:
```suggestion
- The plug-ins provided by Ice now have fixed names: IceIAP, IceBT, IceUDP, IceWS, IceDiscovery, IceLocatorDiscovery.
```

externl (approved):

cpp/src/Ice/SSL/SSLEndpointI.cpp
bernardnormier: Fixed as suggested: the built-in plug-in functions for tcp, ssl, udp, ws and iap are no longer extern "C".

cpp/src/Ice/TcpEndpointI.cpp
```diff
@@ -21,16 +21,34 @@ using namespace std;
 using namespace Ice;
 using namespace IceInternal;
 
-extern "C"
+namespace
 {
-    Plugin* createIceTCP(const CommunicatorPtr& c, const string&, const StringSeq&)
+    const char* const tcpPluginName = "IceTCP";
+
+    Plugin* createIceTCP(const CommunicatorPtr& c, const string& name, const StringSeq&)
```
bernardnormier, line 28: It's now a regular C++ function in the anonymous namespace.

externl (approved):
Looks good!

pepone (approved):

PR #3860, opened 2025-04-16
PIP and GEM build fixes

pepone:
This PR fixes the build of PIP and GEM packages. The build was failing due to not include the new IceDiscovery and IceLocatorDiscovery headers.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -34,7 +34,9 @@ ice_cpp_sources = [
     "../cpp/src/IceLocatorDiscovery",
     "../cpp/src/slice2py",
     "../cpp/src/Slice",
-    "../cpp/include/Ice"]
+    "../cpp/include/Ice",
+    "../cpp/include/IceDiscovery",
+    "../cpp/include/IceLocatorDiscovery",]
 
 # Include directories for the build process
 include_dirs = [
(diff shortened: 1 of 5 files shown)


python/setup.py
```diff
@@ -34,7 +34,9 @@
     "../cpp/src/IceLocatorDiscovery",
     "../cpp/src/slice2py",
     "../cpp/src/Slice",
-    "../cpp/include/Ice"]
+    "../cpp/include/Ice",
+    "../cpp/include/IceDiscovery",
+    "../cpp/include/IceLocatorDiscovery",]
```
pepone, line 39: We need to copy the new headers to the PIP distribution.

python/setup.py
```diff
@@ -48,15 +50,10 @@
...
-        f"dist/ice/cpp/src/IceDiscovery/msbuild/icediscovery/{platform}/{configuration}",
-        f"dist/ice/cpp/src/IceLocatorDiscovery/msbuild/icelocatordiscovery/{platform}/{configuration}",
         f"dist/bzip2-{bzip2_version}"])
 else:
     ice_cpp_sources.append("../cpp/include/generated/Ice")
-    include_dirs.extend([
-        "dist/ice/cpp/include/generated",
-        "dist/ice/cpp/src/IceDiscovery/generated",
-        "dist/ice/cpp/src/IceLocatorDiscovery/generated"])
+    include_dirs.extend(["dist/ice/cpp/include/generated"])
```
pepone, line 57: We no longer need these paths in the header search path.

pepone:
`./generated/Lookup.h` doesn't work on Windows because we have per configuration generated files. I think the simpler is to set the output-dir to IceDiscovery, IceLocatorDiscovery to avoid ambiguity when include the Lookup.h in PIP and GEM builds.

PR #3861, opened 2025-04-16
Fixes for GEM and PIP build failures

pepone:
This PR fixes PIP and GEM build issues related to Lookup.h being available under both IceDiscovery and IceLocatorDiscovery. This is problematic because we build both in the same shared library and we end up with ambiguity of what header should be included.

### DIFF
--- a/cpp/src/IceDiscovery/Makefile.mk
+++ b/cpp/src/IceDiscovery/Makefile.mk
@@ -4,6 +4,8 @@ $(project)_libraries := IceDiscovery
 
 # Tell the slice2cpp rule to ignore the top-level IceDiscovery include directory.
 IceDiscovery_includedir                 := $(project)/generated/IceDiscovery
+$(project)_generated_srcdir             := $(project)/generated/IceDiscovery
+$(project)_generated_includedir         := $(project)/generated/IceDiscovery
 
 IceDiscovery_targetdir                  := $(libdir)
 IceDiscovery_dependencies               := Ice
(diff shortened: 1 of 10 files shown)


externl (approved):

cpp/src/IceDiscovery/Makefile.mk
```diff
@@ -4,6 +4,8 @@ $(project)_libraries := IceDiscovery
 
 # Tell the slice2cpp rule to ignore the top-level IceDiscovery include directory.
 IceDiscovery_includedir                 := $(project)/generated/IceDiscovery
+$(project)_generated_srcdir             := $(project)/generated/IceDiscovery
```
bernardnormier, line 7: Should this be IceDiscovery for consistency with the _includedir setting above?

cpp/src/IceLocatorDiscovery/Makefile.mk
```diff
@@ -4,6 +4,8 @@ $(project)_libraries := IceLocatorDiscovery
 
 # Tell the slice2cpp rule to ignore the top-level IceLocatorDiscovery include directory.
 IceLocatorDiscovery_includedir                  := $(project)/generated/IceLocatorDiscovery
+$(project)_generated_srcdir                     := $(project)/generated/IceLocatorDiscovery
```
bernardnormier, line 7: Ditto

bernardnormier (approved):
Looks good.

An alternative - possibly simpler - would be to rename these internal .ice files to for example DiscoveryLookup.ice and LocatorDiscoveryLookup.ice.

PR #3865, opened 2025-04-16
Simplify ProtocolPluginFacade in C++, C# and Java

bernardnormier:
The ProtocolPluginFacade is what transport plug-ins use to register their endpoint factory(ies) with a communicator.


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ProtocolPluginFacade.java
```diff
@@ -5,25 +5,25 @@
...
-    //
-    Class<?> findClass(String className);
+    // Gets an EndpointFactory.
+    public EndpointFactory getEndpointFactory(short type) {
+        return _instance.endpointFactoryManager().get(type);
+    }
+
+    // Looks up a Java class by name.
+    public Class<?> findClass(String className) {
+        return _instance.findClass(className);
```
externl, line 27: Can we also get rid of this findClass method. It's just a wrapper around the instance one which is a wrapper around the static `Util.findClass` which is "public"

externl (approved):

externl (approved):

PR #3869, opened 2025-04-17
Fix PyPi build

externl:
This PR fixes the PyPi build. With the changes to IceDiscovery and IceLocatorDiscovery. Namely 

- We need to include cpp/include/{IceDiscovery,IceLocatorDiscovery}
- On Linux/macOS the generated internal headers are now in cpp/include/generated.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -34,7 +34,9 @@ ice_cpp_sources = [
     "../cpp/src/IceLocatorDiscovery",
     "../cpp/src/slice2py",
     "../cpp/src/Slice",
-    "../cpp/include/Ice"]
+    "../cpp/include/Ice",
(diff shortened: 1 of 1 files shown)


python/setup.py
```diff
@@ -45,18 +47,19 @@
 
 # Platform-specific adjustments
 if sys.platform == 'win32':
-    ice_cpp_sources.append(f"../cpp/include/generated/{platform}/{configuration}/Ice")
+    ice_cpp_sources.extend([f"../cpp/include/generated/{platform}/{configuration}/Ice",
```
bernardnormier, line 50: This does not look correct.  I believe all the header files go into a local `{platform}/{configuration}/IceDiscovery`.

bernardnormier (approved):

PR #3870, opened 2025-04-17
Add SliceTypeId annotations in Java

bernardnormier:
This PR adds new annotation classes for Slice type ID and compact ID, updates slice2java to generate them as appropriate.

It also updates the implementation of the Object methods (ice_isA etc.) to use these annotations.


externl (approved):
Looks good!

Issue #3835, opened 2025-04-08
Replace Ice.Package by initData field


bernardnormier:
In Java, we use `Ice.Package` and `Ice.Default.Package` to tell an Ice communicator how to resolve class/exception Slice type IDs into classes when the corresponding Slice definitions bear the `java:package:` metadata.

See:
https://doc.zeroc.com/ice/3.7/property-reference/miscellaneous-ice-properties#id-.MiscellaneousIce.*Propertiesv3.7-Ice.Package.module

and
https://doc.zeroc.com/ice/3.7/language-mappings/java-mapping/client-side-slice-to-java-mapping/customizing-the-java-mapping#id-.CustomizingtheJavaMappingv3.7-JavaPackageConfigurationProperties

There are two issues with this setup:
1. It does not work with the new java:identifier metadata on modules
2. The "mapping" is performed at runtime - and can be reconfigured at runtime - when it's really a compile time concern. The corresponding Java package must be available at compile-time.

## Proposal

Replace the Ice.Package and Ice.Default.Package properties by a new initData field:
```java
java.util.List<java.lang.Class> slicePackageInfo;
```

Each element is the metaclass of the "package info/marker class" generated by the Slice compiler. We would rename this class SlicePackageInfo.

The rule is: whenever you use `java:identifier:xxx` on a module in Slice, AND this module defines a class or exception that you may unmarshal in Java, you need to add an entry to `initData.slicePackageInfo`.
Naturally, this entry is optional for the Ice and IceMX modules. It's required for all other modules, including Glacier2, IceGrid, IceStorm, etc.

If you use the deprecated `java:package:xxx` metadata directive:
- it conflicts with java:identifier on modules (=> Slice compilation fails)
- you need to insert entries in `slicePackageInfo` as if you were using `java:identifier` on these modules, per the rule above

The new SlicePackageInfo class is simply:
```java
public class SlicePackageInfo {
    public final static TYPE_ID = "::VisitorCenter";
}
```

### Compact ID sub-proposal

We currently generate a complicated class hierarchy for compact ID support in Java.

We could replace this hierarchy by a final static map in the SlicePackageInfo class, like:
```java
public static final Map<Integer, String> COMPACT_ID_MAP = Map.ofEntries(
    Map.entry(123, "::Foo::Bar"), // compact ID 123 maps to type ID ::Foo::Bar
    Map.entry(124, "::Foo::Baz")
);
``` 

[...]

PR #3872, opened 2025-04-18
Refactor class/exception unmarshaling in Java

bernardnormier:
This PR refactors the class and exception unmarshaling in Java, while maintaining a large level of backwards compatibility.

Prior to this PR, the class/exception unmarshaling in Java was driven by the Ice.Package.module and Ice.Default.Package properties. And the user had some control over the class unmarshaling with the ValueFactoryManager.

This PR introduces a new interface, SliceLoader, and a new initData.sliceLoader field that allows the user to set his own "Slice loader". A Slice loader is responsible to create an instance of a class/exception from a type ID. This type ID can be either a regular type ID string (`::ClearSky::AtmosphericConditions`) or a compact ID string ("123").

This PR also provides a number of implementations of the SliceLoader interface.

The default behavior is to emulate the 3.7 behavior - see updates in Instance.java.

See also #3835.

### DIFF
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -1760,8 +1760,6 @@ Slice::Gen::TypesVisitor::visitModuleStart(const ModulePtr& p)
 bool
 Slice::Gen::TypesVisitor::visitClassDefStart(const ClassDefPtr& p)
 {
-    emitCompactIdHelper(p);
-
     string name = p->mappedName();
     ClassDefPtr baseClass = p->base();
     string package = getPackage(p);
@@ -3567,32 +3565,6 @@ Slice::Gen::TypesVisitor::visitOperation(const OperationPtr& p)
             out << nl << getUnqualified(*t, package) << ".class";
         }
         out << eb << ';';
-    }
-}
-
-void
-Slice::Gen::TypesVisitor::emitCompactIdHelper(const ClassDefPtr& p)
-{
-    if (p->compactId() >= 0)
-    {
-        string prefix = getPackagePrefix(p);
-        if (!prefix.empty())
-        {
-            prefix = prefix + ".";
-        }
-
-        ostringstream os;
-        os << prefix << "com.zeroc.IceCompactId.TypeId_" << p->compactId();
-        open(os.str(), p->file());
-        Output& out = output();
-
-        out << sp;
-        writeHiddenDocComment(out);
-        out << nl << "public class TypeId_" << p->compactId();
-        out << sb;
-        out << nl << "public final static String typeId = \"" << p->scoped() << "\";";
-        out << eb;
-        close();
     }
 }
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ClassSliceLoader.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ClassSliceLoader.java
@@ -1,0 +1,48 @@
+// Copyright (c) ZeroC, Inc.
+
+package com.zeroc.Ice;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Implements SliceLoader using an array of annotated classes.
+ */
+public final class ClassSliceLoader implements SliceLoader {
+    private final Map<String, Class<?>> _typeIdToClass = new HashMap<>();
+
+    /**
+     * Creates a ClassSliceLoader.
+     *
+     * @param classes An array of classes with the {@link SliceTypeId} and/or {@link CompactSliceTypeId} annotation.
+     */
+    public ClassSliceLoader(Class<?>... classes) {
+        for (Class<?> c : classes) {
+            SliceTypeId typeId = c.getAnnotation(SliceTypeId.class);
+            if (typeId != null) {
+                _typeIdToClass.put(typeId.value(), c);
+            }
+            CompactSliceTypeId compactTypeId = c.getAnnotation(CompactSliceTypeId.class);
+            if (compactTypeId != null) {
+                _typeIdToClass.put(String.valueOf(compactTypeId.value()), c);
+            }
+        }
+    }
+
+    @Override
+    public java.lang.Object newInstance(String typeId) {
+        Class<?> c = _typeIdToClass.get(typeId);
+        if (c != null) {
+            try {
+                return c.getDeclaredConstructor().newInstance();
+            } catch (Exception ex) {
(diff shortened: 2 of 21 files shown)


cpp/src/slice2java/Gen.cpp
```diff
@@ -3570,32 +3568,6 @@ Slice::Gen::TypesVisitor::visitOperation(const OperationPtr& p)
     }
 }
 
-void
-Slice::Gen::TypesVisitor::emitCompactIdHelper(const ClassDefPtr& p)
```
bernardnormier, line 3574: Removed these generated helper classes.

As of this PR, if your Java app receives classes with Compact ID,  you need to create a ClassSliceLoader with these classes.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ClassSliceLoader.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Implements SliceLoader using an array of annotated classes.
+ */
+public final class ClassSliceLoader implements SliceLoader {
+    private final Map<String, Class<?>> _typeIdToClass = new HashMap<>();
+
+    /**
+     * Creates a ClassSliceLoader.
+     *
+     * @param classes An array of classes with the {@link SliceTypeId} and/or {@link CompactSliceTypeId} annotation.
```
bernardnormier, line 17: You can also pass an actual array if you prefer.

java/test/src/main/java/test/Ice/packagemd/AllTests.java
```diff
@@ -98,56 +100,49 @@ public static InitialPrx allTests(TestHelper helper) {
             }
 
             {
-                //
-                // Define Ice.Package.Test2=testpkg and try again.
```
bernardnormier, line 102: In 3.7, you could define your Ice.Package.module property after communicator initialization and the implied "Slice loader" would use this property.

I find this feature unreasonable so I dropped it.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InputStream.java
```diff
@@ -1620,29 +1611,6 @@ protected String readTypeId(boolean isIndex) {
             }
         }
 
-        protected Class<?> resolveClass(String typeId) {
-            Class<?> cls = null;
-            if (_typeIdCache == null) {
-                _typeIdCache = new HashMap<>(); // Lazy initialization.
```
bernardnormier, line 1626: In the existing code, there was a per-InputStream (!) cache for typeId to Class and compactId to Class. I removed them both. The caching, if any, is now performed by the communicator-wide SliceLoader implementations.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InputStream.java
```diff
@@ -1620,29 +1611,6 @@ protected String readTypeId(boolean isIndex) {
...
-            } else {
-                cls = _typeIdCache.get(typeId);
-            }
-
-            if (cls == EncapsDecoder.class) {// Marker for non-existent class.
```
bernardnormier, line 1631: I did not (and could not) keep this "not found" caching. See comments in DefaultPackageSliceLoader.

I guess it could be meaningful (performance wise) if you get of unknown  derived classes that you then slice to a known base. In this unlikely situation, you could install your own SliceLoader that somehow optimizes "not found".

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
```diff
@@ -546,145 +544,13 @@ public void setThreadHooks(Runnable threadStart, Runnable threadStop) {
...
-        //
-        // See if the application defined any package prefixes with Ice.Package.MODULE property.
-        //
-        int pos = typeId.indexOf(':', 2);
-        if (pos != -1) {
-            String topLevelModule = typeId.substring(2, pos);
-            String[] packagePrefixes = _builtInModulePackagePrefixes.get(topLevelModule);
-            if (packagePrefixes == null) {
-                packagePrefixes =
-                    _initData.properties.getIcePropertyAsList("Ice.Package." + topLevelModule);
```
bernardnormier, line 605: Even though we document a single value, the mapping for a module was actually a list of package-prefixes. Do we want to keep this behavior?

It's mostly useful for IceMX. See #3873.

pepone (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
bernardnormier: See #3873.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
```diff
@@ -814,6 +679,49 @@ public void initialize(Communicator communicator, InitializationData initData) {
...
 
+            // Update _initData.sliceLoader.
+
+            var sliceLoader = new CompositeSliceLoader();
+
+            if (_initData.sliceLoader != null) {
+                sliceLoader.add(_initData.sliceLoader);
+            }
+
+            _initData.sliceLoader = sliceLoader;
```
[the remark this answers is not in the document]
bernardnormier: That's what I did initially but then switched back to initData.sliceLoader for consistency.

We keep around a clone of _initData and set value in this close, like _initData.logger, _initData.properties and more. While Instance._sliceLoader would be nicer, it would be inconsistent with these other fields.

externl (approved):
Looks good!

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ClassSliceLoader.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Creates a ClassSliceLoader.
+     *
+     * @param classes An array of classes with the {@link SliceTypeId} annotation. Each class may also have the
+     * {@link CompactSliceTypeId} annotation.
+     */
+    public ClassSliceLoader(Class<?>... classes) {
+        for (Class<?> c : classes) {
+            SliceTypeId typeId = c.getAnnotation(SliceTypeId.class);
+            if (typeId != null) {
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

PR #3880, opened 2025-04-21
Add test for JavaScript default servants

### LINKED ISSUE #3755: Default servants not tested in JavaScript
We need to add tests for default servant to JavaScript. This feature is implemented but we have no single test for it.

pepone:
Fix #3755


js/src/Ice/ConnectRequestHandler.js
```diff
@@ -118,9 +118,6 @@ export class ConnectRequestHandler {
             } catch (ex) {
                 if (ex instanceof RetryException) {
                     exception = ex.inner;
-
-                    // Remove the request handler before retrying.
-                    this._reference.getInstance().requestHandlerFactory().removeRequestHandler(this._reference, this);
```
pepone, line 123: The test revealed this bug.

externl (approved):

js/test/Ice/defaultServant/Client.ts
```diff
@@ -0,0 +1,107 @@
...
+
+import { Ice } from "@zeroc/ice";
+import { TestHelper } from "../../Common/TestHelper.js";
+import { Test } from "./Test.js";
+const test = TestHelper.test;
+
+export class Client extends TestHelper {
+    async allTests(args: string[]) {
+        const communicator = this.communicator();
+        const oa = new Test.RemoteObjectAdapterPrx(communicator, `remote-object-adapter:${this.getTestEndpoint()}`);
```
bernardnormier, line 11: I don't understand why we need this complicated setup. We can't simply host an OA locally and call it directly?

pepone: With JavaScript we use the C++ Echo Server from `cpp/test/Ice/echo` to workaround the fact that the JavaScript mapping doesn't support incoming connections. The test scripts takes care of automatically starting the EchoServer for JavaScript ClientServerTests.

If we want to use a client test only, we can add a custom test.py script that ensures echo server is also started.

bernardnormier: I meant: we can keep this client -> c++ server -> back to same JS client.

But when the client need to call its own local OA, it can just call it directly. It does not need to make remote invocations via the C++ server.

pepone: Simplified as suggested.

bernardnormier (approved):
Looks good.

PR #3881, opened 2025-04-21
Adjusting how tasks are integrated into Java build system

ayoitslilith:
Instead of setting flags to enable and disable tasks in the Gradle build file, we now use command line arguments to change the build behavior when building off of the Makefile or ice.proj when building in Java.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -25,4 +25,4 @@ jobs:
 
       - name: Run rewriteDryRun with build failure enabled
         working-directory: java
-        run: ./gradlew rewriteDryRun -PenableCodeAnalysis=true
+        run: ./gradlew rewriteDryRun
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -12,9 +12,6 @@ allprojects {
     }
 }
 
-// Defines and initializes enableCodeAnalysis to the corresponding property in gradle.properties.
-def enableCodeAnalysis = project.property('enableCodeAnalysis').toBoolean()
-
 subprojects {
     project.ext.topSrcDir = "$rootProject.projectDir/.."
 
@@ -50,11 +47,7 @@ subprojects {
     }
 
     // We want Checkstyle to generally run after Rewrite and to exclude files generated from Slice.
-    // You need to use -PenableCodeAnalysis to be able to invoke the checkstyle, rewriteRun, and rewriteDryRun tasks
-    // The tasks still need to be manually called if nothing else invokes them.
     tasks.withType(Checkstyle) {
-        task ->
-            task.enabled = enableCodeAnalysis
         dependsOn(rewriteDryRun)
         exclude(
             '**/generated/**'
@@ -151,15 +144,7 @@ dependencies {
     )
 }
 
-// You need to use -PenableCodeAnalysis to be able to invoke the checkstyle, rewriteRun, and rewriteDryRun tasks.
-// The tasks still need to be manually called if nothing else invokes them.
+// Causes build failure when rewriteRun would make changes.
 tasks.named('rewriteDryRun') {
-    task ->
-        task.enabled = enableCodeAnalysis
     rewrite.failOnDryRunResults = true
-}
-
-tasks.named('rewriteRun') {
-    task ->
-        task.enabled = enableCodeAnalysis
 }
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -48,7 +48,9 @@ subprojects {
 
     // We want Checkstyle to generally run after Rewrite and to exclude files generated from Slice.
     tasks.withType(Checkstyle) {
-        dependsOn(rewriteDryRun)
+        task ->
+            rewriteDryRun.dependsOn(task)
+            rewriteRun.dependsOn(task)
         exclude(
             '**/generated/**'
         )
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -46,7 +46,7 @@ subprojects {
         options.deprecation = true
     }
 
-    // We want Checkstyle to generally run after Rewrite and to exclude files generated from Slice.
+    // We want Checkstyle to generally run before Rewrite and to exclude files generated from Slice.
     tasks.withType(Checkstyle) {
         task ->
             rewriteDryRun.dependsOn(task)
--- a/java/msbuild/ice.proj
+++ b/java/msbuild/ice.proj
@@ -26,14 +26,14 @@
   </Target>
 
   <Target Name="BuildDist" DependsOnTargets="slice2java">
-    <Exec Command="$(Gradle) $(GRADLEARGS) dist -x rewriteDryRun -x checkstyle -PcppPlatform=$(Platform) -PcppConfiguration=$(Configuration)"
+    <Exec Command="$(Gradle) $(GRADLEARGS) dist -PcppPlatform=$(Platform) -PcppConfiguration=$(Configuration)"
           WorkingDirectory="$(MSBuildThisFileDirectory).."
           StdOutEncoding="utf-8"
           StdErrEncoding="utf-8" />
   </Target>
 
   <Target Name="Build" DependsOnTargets="slice2java">
-    <Exec Command="$(Gradle) $(GRADLEARGS) build -x rewriteDryRun -x checkstyle -PcppPlatform=$(Platform) -PcppConfiguration=$(Configuration)"
+    <Exec Command="$(Gradle) $(GRADLEARGS) build -PcppPlatform=$(Platform) -PcppConfiguration=$(Configuration)"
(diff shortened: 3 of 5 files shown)


java/msbuild/ice.proj
```diff
@@ -26,14 +26,14 @@
   </Target>
 
   <Target Name="BuildDist" DependsOnTargets="slice2java">
-    <Exec Command="$(Gradle) $(GRADLEARGS) dist -PcppPlatform=$(Platform) -PcppConfiguration=$(Configuration)"
+    <Exec Command="$(Gradle) $(GRADLEARGS) dist -x rewriteDryRun -x checkstyle -PcppPlatform=$(Platform) -PcppConfiguration=$(Configuration)"
```
externl, line 29: I know I recommend to put these on dist but now I'm actually not sure that it's required.

.github/workflows/java.yml
```diff
@@ -25,4 +25,4 @@ jobs:
 
       - name: Run rewriteDryRun with build failure enabled
         working-directory: java
-        run: ./gradlew rewriteDryRun -PenableCodeAnalysis=true
+        run: ./gradlew rewriteDryRun
```
externl, line 28: We should also run checkstyle in a separate task.

pepone:
I still see:

```
[ant:checkstyle] [WARN] /Users/jose/Documents/3.8/dryrun-toggle-2/java/src/com.zeroc.ice/build/generated/source/slice/main/com/zeroc/Ice/LoggerAdmin.java:28: Javadoc tag '@param' should be preceded with an empty line. [RequireEmptyLineBeforeBlockTagGroup]
[ant:checkstyle] [WARN] /Users/jose/Documents/3.8/dryrun-toggle-2/java/src/com.zeroc.ice/build/generated/source/slice/main/com/zeroc/Ice/LoggerAdmin.java:60: Line is longer than 120 characters (found 133). [LineLength]
[ant:checkstyle] [WARN] /Users/jose/Documents/3.8/dryrun-toggle-2/java/src/com.zeroc.ice/build/generated/source/slice/main/com/zeroc/Ice/LoggerAdmin.java:61: Javadoc tag '@param' should be preceded with an empty line. [RequireEmptyLineBeforeBlockTagGroup]
```

I'm not sure why we have:

```
    // We want Checkstyle to generally run after Rewrite and to exclude files generated from Slice.
    tasks.withType(Checkstyle) {
        dependsOn(rewriteDryRun)
        exclude(
            '**/generated/**'
        )
    }
```

This introduces a dependency from Checkstyle to Rewrite, which means Rewrite now runs as part of the default build task — since Checkstyle is wired into check, and check is part of build.

As a result, we end up having to explicitly disable Rewrite in builds where it’s not needed. Is there a specific reason Rewrite has to run before Checkstyle? Could we avoid this dependency unless it’s strictly required?

The default build with `checkStyle` is quite fast I'm not convinced that we need to disable it by default:

```
BUILD SUCCESSFUL in 27s
93 actionable tasks: 93 executed
jose@mac java % ./gradlew build

BUILD SUCCESSFUL in 624ms
93 actionable tasks: 30 executed, 63 up-to-date
jose@mac java % 
```

This is after removing the `rewriteDryRun` dependency in check. I think we should:

- Remove dependency from Checkstyle to Rewrite.
- No need for Makefile and ice.proj updates `-x rewriteDryRun -x checkstyle`
- Keep `./gradlew rewriteDryRun` for CI

externl:
@pepone Fine by me, however we should also make sure that checkstyle doesn't cause the build to fail but does fail in a CI check.

ayoitslilith:
> I still see:
> 
> ```
> [ant:checkstyle] [WARN] /Users/jose/Documents/3.8/dryrun-toggle-2/java/src/com.zeroc.ice/build/generated/source/slice/main/com/zeroc/Ice/LoggerAdmin.java:28: Javadoc tag '@param' should be preceded with an empty line. [RequireEmptyLineBeforeBlockTagGroup]
> [ant:checkstyle] [WARN] /Users/jose/Documents/3.8/dryrun-toggle-2/java/src/com.zeroc.ice/build/generated/source/slice/main/com/zeroc/Ice/LoggerAdmin.java:60: Line is longer than 120 characters (found 133). [LineLength]
> [ant:checkstyle] [WARN] /Users/jose/Documents/3.8/dryrun-toggle-2/java/src/com.zeroc.ice/build/generated/source/slice/main/com/zeroc/Ice/LoggerAdmin.java:61: Javadoc tag '@param' should be preceded with an empty line. [RequireEmptyLineBeforeBlockTagGroup]
> ```

Are you running it through `make`? The expected behavior is that `./gradlew build` still requires the use of exemption arguments `-x checkstyle -x rewriteDryRun'. Regardless, the expected behavior will be changing with this next PR - `checkstyle` now should run alongside the default build sequence.

> 
> I'm not sure why we have:
> 
> ```
>     // We want Checkstyle to generally run after Rewrite and to exclude files generated from Slice.
>     tasks.withType(Checkstyle) {
>         dependsOn(rewriteDryRun)
>         exclude(
>             '**/generated/**'
>         )
>     }
> ```
> 
> This introduces a dependency from Checkstyle to Rewrite, which means Rewrite now runs as part of the default build task — since Checkstyle is wired into check, and check is part of build.
> 
> As a result, we end up having to explicitly disable Rewrite in builds where it’s not needed. Is there a specific reason Rewrite has to run before Checkstyle? Could we avoid this dependency unless it’s strictly required?
> 

The intention is for Rewrite to use Checkstyle as part of its formula - however, looking now, I'm realizing I made an obvious mistake in the order of dependencies. While this doesn't seem to have caused any issues related to Rewrite's use of Checkstyle's styles, they certainly do not need to depend on each other in this direction.

> The default build with `checkStyle` is quite fast I'm not convinced that we need to disable it by default:
> 
> ```
> BUILD SUCCESSFUL in 27s
> 93 actionable tasks: 93 executed
> jose@mac java % ./gradlew build
> 
> BUILD SUCCESSFUL in 624ms
> 93 actionable tasks: 30 executed, 63 up-to-date
> jose@mac java % 
> ```
> 
> This is after removing the `CheckStye` -> `rewriteDryRun` dependency. I think we should:
> 
>     * Remove dependency from Checkstyle to Rewrite.
> 
>     * No need for Makefile and ice.proj updates `-x rewriteDryRun -x checkstyle`
> 
>     * Keep `./gradlew rewriteDryRun` for CI

With this next PR, `checkstyle` will still run in the default build sequence, but `rewriteDryRun` should not. `rewriteDryRun` and `rewriteRun` have now been properly set to depend on `checkstyle` tasks, `checkstyle` tasks will run in the default build sequence, and `rewriteDryRun` is no longer automatically called when `checkstyle` tasks are called.

java/msbuild/ice.proj
pepone: We should remove `-x checkstyle`

java/msbuild/ice.proj
```diff
@@ -26,14 +26,14 @@
...
           WorkingDirectory="$(MSBuildThisFileDirectory).."
           StdOutEncoding="utf-8"
           StdErrEncoding="utf-8" />
   </Target>
 
   <Target Name="Build" DependsOnTargets="slice2java">
-    <Exec Command="$(Gradle) $(GRADLEARGS) build -PcppPlatform=$(Platform) -PcppConfiguration=$(Configuration)"
+    <Exec Command="$(Gradle) $(GRADLEARGS) build -x rewriteDryRun -x checkstyle -PcppPlatform=$(Platform) -PcppConfiguration=$(Configuration)"
```
pepone, line 36: Here too, remove `-x checkstyle`

pepone (approved):

java/build.gradle
```diff
@@ -50,12 +47,10 @@ subprojects {
...
 
     // We want Checkstyle to generally run after Rewrite and to exclude files generated from Slice.
-    // You need to use -PenableCodeAnalysis to be able to invoke the checkstyle, rewriteRun, and rewriteDryRun tasks
-    // The tasks still need to be manually called if nothing else invokes them.
     tasks.withType(Checkstyle) {
         task ->
-            task.enabled = enableCodeAnalysis
-        dependsOn(rewriteDryRun)
+            rewriteDryRun.dependsOn(task)
+            rewriteRun.dependsOn(task)
```
externl, line 53: are these necessary?

ayoitslilith: Without these, OpenRewrite has inconsistent behavior regarding following Checkstyle rules.

java/build.gradle
```diff
@@ -50,12 +47,10 @@ subprojects {
     }
 
     // We want Checkstyle to generally run after Rewrite and to exclude files generated from Slice.
```
externl, line 49: I think this comment needs updating.

PR #3885, opened 2025-04-22
Revert session recovery fixes from d6737c2

pepone:
This PR reverts the fixes from d6737c2 which turn out to not be correct. We still have to fix the root issue with the session recovery starving but I let this for a separate PR.

### DIFF
--- a/cpp/src/DataStorm/NodeI.cpp
+++ b/cpp/src/DataStorm/NodeI.cpp
@@ -150,8 +150,6 @@ NodeI::createSession(
     checkNotNull(subscriber, __FILE__, __LINE__, current);
     checkNotNull(subscriberSession, __FILE__, __LINE__, current);
 
-    bool isWellKnown = subscriber->ice_getEndpoints().empty() && subscriber->ice_getAdapterId().empty();
-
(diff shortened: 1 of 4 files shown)


cpp/src/DataStorm/NodeI.cpp
```diff
@@ -199,32 +197,12 @@ NodeI::createSession(
...
-                        if (traceLevels->session > 2)
-                        {
-                            Trace out(traceLevels->logger, traceLevels->sessionCat);
-                            out << "node '" << current.id << "' is discarding current session '" << session->getId()
-                                << "' because subscriber '" << subscriber
-                                << "' (using a well-known proxy) requested creating a session over a new connection";
-                        }
-                        session->disconnect();
-                        session->reconnect(*subscriber, current.con);
-                    }
```
pepone, line 217: Here `disconnected()` marked the local side as disconnected, and then the call to reconnect starts the recovery. The assumption was that at this point the peer was disconnected and that is why `createSession` request is received over a new connection. The assumption is bogus for example a concurrent session attempt might have just success over a different connection.

bernardnormier (approved):

externl (approved):

PR #3888, opened 2025-04-22
Replace FactoryTable by SliceLoader in C++

bernardnormier:
This PR adds SliceLoader to C++ and replaces the internal factory table class by a new IceInternal::DefaultSliceLoader that provides similar functionality.

This PR does not remove ValueFactory/ValueFactoryManager in C++. The next step is to replace ValueFactoryManager in Python, Ruby and PHP, before removing it from C++.


cpp/include/Ice/DefaultSliceLoader.h
```diff
@@ -0,0 +1,165 @@
...
+            }
+        }
+
+        void removeClass(std::string_view typeId, int compactId)
+        {
+            std::lock_guard lock{_mutex};
+            auto p = _classFactories.find(typeId);
+            if (p != _classFactories.end())
+            {
+                if (--p->second.second == 0)
```
bernardnormier, line 65: That's the same logic as in the previous FactoryTable. It's presumably to support unloading of libraries that register class/exceptions.

cpp/include/Ice/DefaultSliceLoader.h
```diff
@@ -0,0 +1,165 @@
...
+            if (p != _exceptionFactories.end())
+            {
+                if (--p->second.second == 0)
+                {
+                    _exceptionFactories.erase(p);
+                }
+            }
+        }
+
+        // temporary
```
bernardnormier, line 112: This is needed until we can eliminate the compact ID resolver used by Python etc.

cpp/test/Ice/objects/Client.cpp
```diff
@@ -33,14 +26,12 @@ class Client : public Test::TestHelper
...
-    properties->setProperty("Ice.AcceptClassCycles", "1");
-
-    Ice::CommunicatorHolder communicator = initialize(argc, argv, properties);
-    communicator->getValueFactoryManager()->add(makeFactory<BI>(), "::Test::B");
-    communicator->getValueFactoryManager()->add(makeFactory<CI>(), "::Test::C");
-    communicator->getValueFactoryManager()->add(makeFactory<DI>(), "::Test::D");
+    Ice::InitializationData initData;
+    initData.properties = createTestProperties(argc, argv);
+    initData.properties->setProperty("Ice.AcceptClassCycles", "1");
+    initData.sliceLoader = make_shared<CustomSliceLoader>();
```
bernardnormier, line 32: Use/test the new initData.sliceLoader mechanism.

cpp/src/Ice/InputStream.cpp
```diff
@@ -1568,7 +1557,12 @@ Ice::InputStream::EncapsDecoder10::throwException(UserExceptionFactory exception
         //
         if (!exceptionFactory)
```
bernardnormier, line 1558: Need to update Python etc. that use this exceptionFactory parameter.

pepone (approved):

cpp/include/Ice/SliceLoader.h
```diff
@@ -0,0 +1,67 @@
...
+    {
+    public:
+        virtual ~SliceLoader();
+
+        /// Creates an instance of a class mapped from a Slice class based on a Slice type ID.
+        /// @param typeId The Slice type ID or compact type ID.
+        /// @return A new instance of the class identified by @p typeId, or nullptr if the implementation cannot find
+        /// the corresponding class.
+        /// @throws MarshalException Thrown when the corresponding class was found but its instantiation failed.
+        virtual ValuePtr newClassInstance(std::string_view typeId) const;
```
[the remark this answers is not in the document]
bernardnormier: I initially made them pure virtual, and then changed them to return nullptr.

The logic is when you create a derived class, you typically only want to override only one of these functions, and "return nullptr" is the correct no-op behavior.

externl (approved):
Looks good!

PR #3891, opened 2025-04-23
Retry session recovery if it doesn't success with a configured timeout

pepone:


cpp/src/DataStorm/SessionI.cpp
```diff
@@ -743,8 +755,14 @@ SessionI::retry(NodePrx node, exception_ptr exception)
...
         }
 
-        _retryTask = make_shared<IceInternal::InlineTimerTask>([node = std::move(node), self = shared_from_this()]
-                                                               { self->reconnect(node); });
+        _retryTask = make_shared<IceInternal::InlineTimerTask>(
+            [node = std::move(node), self = shared_from_this()]
+            {
+                self->_nextRetryAttemptTask =
+                    make_shared<IceInternal::InlineTimerTask>([self, node] { self->retry(node, nullptr); });
+                self->_instance->scheduleTimerTask(self->_nextRetryAttemptTask, 10s);
```
pepone, line 763: I think we should add a separate property for this timeout. Not clear we will want exponential backoff strategy like we have for retries.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -743,8 +755,14 @@ SessionI::retry(NodePrx node, exception_ptr exception)
...
+                self->reconnect(node);
+            });
         _instance->scheduleTimerTask(_retryTask, delay);
```
pepone, line 766: One gotcha here is we accumulate both delays, for example if this retry corresponds to nextRetryAttemptTask firing, we have already waited 10s, and now we will "delay" until reconnect is scheduled. I think will be good to discount the time we have already waited.

config/PropertyNames.xml
```diff
@@ -411,6 +411,7 @@
         <property name="Node.RetryCount" default="6" languages="cpp" />
         <property name="Node.RetryDelay" default="500" languages="cpp" />
         <property name="Node.RetryMultiplier" default="2" languages="cpp" />
+        <property name="Node.RetryTimeout" default="10000" languages="cpp" />
```
pepone, line 414: A new property that controls how much time we wait for a recovery attempt to success before we schedule a new one. Defaults to 10s.

PR #3893, opened 2025-04-23
Bringing back 'requires' files for packages in Ruby

ayoitslilith:
This PR brings back the files to connect Ice for Ruby with some of the various packages that were included in the past. Someone using an installation of Ice for Ruby will still have to use `requires 'PackageName'` in order to use the packages that are being brought back through this PR.

### DIFF
--- a/ruby/ruby/Glacier2.rb
+++ b/ruby/ruby/Glacier2.rb
@@ -1,0 +1,7 @@
+#
+# Copyright (c) ZeroC, Inc. All rights reserved.
+#
+
+require 'Glacier2/Router'
+require 'Glacier2/PermissionsVerifier'
+require 'Glacier2/Metrics'
--- a/ruby/ruby/Glacier2.rb
+++ b/ruby/ruby/Glacier2.rb
@@ -1,4 +1,4 @@
-# Copyright (c) ZeroC, Inc. All rights reserved.
+# Copyright (c) ZeroC, Inc.
 
 require 'Glacier2/Router'
 require 'Glacier2/PermissionsVerifier'
--- a/ruby/ruby/Glacier2.rb
+++ b/ruby/ruby/Glacier2.rb
@@ -1,6 +1,4 @@
-#
 # Copyright (c) ZeroC, Inc. All rights reserved.
-#
 
 require 'Glacier2/Router'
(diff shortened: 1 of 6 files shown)


ruby/ruby/Glacier2.rb
```diff
@@ -0,0 +1,7 @@
+#
```
pepone, line 1: Copyright should be just

```
# Copyright (c) ZeroC, Inc. All rights reserved.
```

ruby/ruby/IceBox.rb
```diff
@@ -0,0 +1,5 @@
+#
+# Copyright (c) ZeroC, Inc. All rights reserved.
+#
+
+require 'IceBox/IceBox'
```
pepone, line 5: Here we want to include `IceBox/ServiceManager.rb` this file was renamed in 3.8.

ruby/ruby/IceGrid.rb
```diff
@@ -0,0 +1,9 @@
+#
+# Copyright (c) ZeroC, Inc. All rights reserved.
+#
+
+require 'IceGrid/Admin.rb'
+require 'IceGrid/Descriptor.rb'
+require 'IceGrid/FileParser.rb'
+require 'IceGrid/Registry.rb'
+require 'IceGrid/UserAccountMapper.rb'
```
pepone, line 9: We should include all generated files. `Exception.rb` and `Session.rb` are missing.

pepone (commented):
We should also update ruby/Rakefile to copy the corresponding Slice files into the gem

```
diff --git a/ruby/Rakefile b/ruby/Rakefile
index 0bcee9ba67..76ccd72650 100644
--- a/ruby/Rakefile
+++ b/ruby/Rakefile
@@ -84,7 +84,11 @@ task :generate_sources do
 
     # Define source Slice directories included in the gem.
     ice_slice_sources = [
+        "../slice/Glacier2",
         "../slice/Ice",
+        "../slice/IceBox",
+        "../slice/IceGrid",
+        "../slice/IceStorm",
     ]
 ```

pepone (approved):
Looks good!

I forgot to mention that we have also removed the "All rights reserved." from copyrights.

PR #3895, opened 2025-04-23
Update IcePy to use SliceLoader

bernardnormier:
This PR updates the InputStream constructors in C++ to accept an optional SliceLoaderPtr parameter.

This Slice loader parameter is used by IcePy to inject its Python-specific Slice loader, that creates Python class instances and the C++ ExceptionReader for exceptions (like before).

### DIFF
--- a/cpp/include/Ice/DefaultSliceLoader.h
+++ b/cpp/include/Ice/DefaultSliceLoader.h
@@ -121,7 +121,7 @@ namespace IceInternal
             {
                 return p->second;
             }
-            return {};
+            return std::to_string(compactId);
         }
 
     private:
--- a/cpp/src/Ice/InputStream.cpp
+++ b/cpp/src/Ice/InputStream.cpp
@@ -121,7 +121,8 @@ Ice::InputStream::InputStream(
 }
 
 Ice::InputStream::InputStream(InputStream&& other) noexcept
-    : InputStream{other._instance, other._encoding, std::move(other), nullptr} // only moves (and resets) the base class
+    // only moves (and resets) the base class
+    : InputStream{other._instance, other._encoding, std::move(other), other._sliceLoader}
 {
     _closure = other._closure;
     _startSeq = other._startSeq;
--- a/cpp/src/Ice/InputStream.cpp
+++ b/cpp/src/Ice/InputStream.cpp
@@ -80,7 +80,10 @@ Ice::InputStream::InputStream(const CommunicatorPtr& communicator, const vector<
 {
 }
 
-Ice::InputStream::InputStream(const CommunicatorPtr& communicator, pair<const byte*, const byte*> p, SliceLoaderPtr sliceLoader)
+Ice::InputStream::InputStream(
+    const CommunicatorPtr& communicator,
+    pair<const byte*, const byte*> p,
+    SliceLoaderPtr sliceLoader)
     : InputStream{
           getInstance(communicator).get(),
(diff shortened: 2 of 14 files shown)


cpp/include/Ice/DefaultSliceLoader.h
```diff
@@ -121,7 +121,7 @@ namespace IceInternal
             {
                 return p->second;
             }
-            return {};
+            return std::to_string(compactId);
```
bernardnormier, line 124: The ultimate goal is to remove this compact ID resolver.

python/modules/IcePy/Communicator.cpp
```diff
@@ -152,6 +153,8 @@ communicatorInit(CommunicatorObject* self, PyObject* args, PyObject* /*kwds*/)
         // We always supply our own implementation of ValueFactoryManager.
         data.valueFactoryManager = ValueFactoryManager::create();
 
+        // data.sliceLoader remains null as we don't want to change the Slice loader for the Ice C++ runtime.
```
bernardnormier, line 156: It actually breaks for Ice::ObjectNotFoundException, which is created/caught in the location-caching code (all in C++).

externl (approved):

python/modules/IcePy/Communicator.cpp
pepone: Why is `Ice::ObjectNotFoundException` loaded using the SliceLoader? This is a dispatch exception not an Slice excpetion.

python/modules/IcePy/DefaultSliceLoader.cpp
```diff
@@ -0,0 +1,56 @@
...
+    if (!info)
+    {
+        return nullptr;
+    }
+
+    // Instantiate the object.
+    auto* type = reinterpret_cast<PyTypeObject*>(info->pythonType);
+    PyObjectHandle emptyArgs{PyTuple_New(0)};
+    PyObjectHandle obj{type->tp_new(type, emptyArgs.get(), nullptr)};
+    if (!obj.get())
```
pepone, line 38: We provide an operator bool for `PyObjectHandle`, so it makes sense to use it here.

```suggestion
    if (!obj)
```

pepone (approved):

cpp/src/Ice/InputStream.cpp
```diff
@@ -66,46 +66,62 @@ IceInternal::Ex::throwMarshalException(const char* file, int line, string reason
...
+    Buffer& buf,
+    bool adopt,
+    SliceLoaderPtr sliceLoader)
+    : InputStream{instance, encoding, Buffer{buf, adopt}, std::move(sliceLoader)}
 {
 }
 
 Ice::InputStream::InputStream(InputStream&& other) noexcept
-    : InputStream{other._instance, other._encoding, std::move(other)} // only moves (and resets) the base class
+    : InputStream{other._instance, other._encoding, std::move(other), nullptr} // only moves (and resets) the base class
```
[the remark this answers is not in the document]
bernardnormier: Fixed

python/modules/IcePy/Communicator.cpp
bernardnormier: ObjectNotFound is a Slice exception defined here:
https://github.com/zeroc-ice/ice/blob/f5995683fe2184a0c44661dfe9d9868c4234a2d8/slice/Ice/Locator.ice#L27

Not to be confused with ObjectNotExistException.

PR #3898, opened 2025-04-24
Use SliceLoader in Ruby and PHP

bernardnormier:
This PR is the counterpart of #3895 for Ruby and PHP.

### DIFF
--- a/cpp/src/Ice/InputStream.cpp
+++ b/cpp/src/Ice/InputStream.cpp
@@ -1057,10 +1057,14 @@ Ice::InputStream::InputStream(Instance* instance, EncodingVersion encoding, Buff
       _encoding(encoding),
       _classGraphDepthMax(instance->classGraphDepthMax()),
       _valueFactoryManager(instance->initializationData().valueFactoryManager),
-      _sliceLoader{sliceLoader ? std::move(sliceLoader) : instance->sliceLoader()}
+      _sliceLoader{std::move(sliceLoader)}
 {
+    if (!_sliceLoader)
+    {
+        const_cast<SliceLoaderPtr&>(_sliceLoader) = _instance->sliceLoader();
+    }
+
     assert(_valueFactoryManager);
-    assert(_sliceLoader);
 }
 
 ReferencePtr
--- a/php/src/DefaultSliceLoader.h
+++ b/php/src/DefaultSliceLoader.h
@@ -1,0 +1,26 @@
+// Copyright (c) ZeroC, Inc.
+
+#ifndef ICEPHP_DEFAULT_SLICE_LOADER_H
+#define ICEPHP_DEFAULT_SLICE_LOADER_H
+
+#include "Communicator.h"
+#include "Config.h"
+#include "Ice/Ice.h"
+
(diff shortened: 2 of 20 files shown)


php/src/DefaultSliceLoader.h
```diff
@@ -0,0 +1,26 @@
...
+#ifndef ICEPHP_DEFAULT_SLICE_LOADER_H
+#define ICEPHP_DEFAULT_SLICE_LOADER_H
+
+#include "Communicator.h"
+#include "Config.h"
+#include "Ice/Ice.h"
+
+namespace IcePHP
+{
+    /// Instantiates generated Python classes/exceptions based on the Slice type ID.
```
pepone, line 12:
```suggestion
    /// Instantiates generated PHP classes/exceptions based on the Slice type ID.
```

ruby/src/IceRuby/DefaultSliceLoader.cpp
```diff
@@ -0,0 +1,45 @@
...
+    // opportunity to preserve the object.
+    ClassInfoPtr info =
+        typeId == Ice::Value::ice_staticId() ? lookupClassInfo("::Ice::UnknownSlicedValue") : lookupClassInfo(typeId);
+
+    if (!info)
+    {
+        return nullptr;
+    }
+
+    volatile VALUE obj = callRuby(rb_class_new_instance, 0, reinterpret_cast<VALUE*>(0), info->rubyClass);
```
pepone, line 30: volatile is not necessary here. The VALUE is passed to ValueReader which register its address to prevent GC.

The volatile keyword is required in some situations to prevent GC, when variables can be optimized away. See "`volatile` to protect from GC" in https://ruby-hacking-guide.github.io/gc.html

```suggestion
    VALUE obj = callRuby(rb_class_new_instance, 0, reinterpret_cast<VALUE*>(nullptr), info->rubyClass);
```

pepone (approved):

ruby/src/IceRuby/DefaultSliceLoader.cpp
bernardnormier: The compiler does not like the nullptr in `reinterpret_cast<VALUE*>(nullptr)`, so I changed it back to 0.

externl (approved):

cpp/src/Ice/InputStream.cpp
```diff
@@ -1057,10 +1057,14 @@ Ice::InputStream::InputStream(Instance* instance, EncodingVersion encoding, Buff
...
-      _sliceLoader{sliceLoader ? std::move(sliceLoader) : instance->sliceLoader()}
+      _sliceLoader{std::move(sliceLoader)}
 {
+    if (!_sliceLoader)
+    {
+        const_cast<SliceLoaderPtr&>(_sliceLoader) = _instance->sliceLoader();
+    }
+
     assert(_valueFactoryManager);
-    assert(_sliceLoader);
```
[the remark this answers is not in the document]
bernardnormier: It can't. The line above assigns a non-null value to _sliceLoader if it's null.

PR #3899, opened 2025-04-24
Fix IcePy build with clang

bernardnormier:


externl (approved):

externl:
Looks like we need a version/feature check:

```
modules/IcePy/Communicator.cpp:38:38: error: unknown warning group '-Wcast-function-type-mismatch', ignored [-Werror,-Wunknown-warning-option]
   38 | #    pragma clang diagnostic ignored "-Wcast-function-type-mismatch"
      |          
```

PR #3902, opened 2025-04-24
Report exceptions from DataStorm session creation.

pepone:
This PR updates DataStorm publisher and subscriber session creation to report errors. This allows to retry the session creation when the state of the session gets out of sync, meaning on peer detects the closure before the other and is trying to migrate the session to a new connection.

### DIFF
--- a/cpp/src/DataStorm/Contract.ice
+++ b/cpp/src/DataStorm/Contract.ice
@@ -394,20 +394,17 @@ module DataStormContract
     /// The reason for the session creation failure.
     enum SessionError
     {
-        // The session is already connected.
+        /// The session is already connected.
         AlreadyConnected,
 
-        // Node is shutting down.
+        /// Node is shutting down.
         NodeShutdown,
 
-        // The connection failed
-        ConnectionFailure,
-
-        // A confirmation was received for a session that doesn't exist.
+        /// A confirmation was received for a session that doesn't exist.
         SessionNotFound,
 
-        // The session failed for an unknown reason.
-        UnknownError,
+        /// The session creation failed due to an internal error.
+        Internal,
     }
 
     /// Throws when the session cannot be created.
--- a/cpp/src/DataStorm/Contract.ice
+++ b/cpp/src/DataStorm/Contract.ice
@@ -391,6 +391,31 @@ module DataStormContract
         void s(long topicId, long elementId, DataSample sample);
     }
 
+    /// The reason for the session creation failure.
+    enum SessionError
+    {
+        // The session is already connected.
+        AlreadyConnected,
+
+        // Node is shutting down.
+        NodeShutdown,
+
+        // The connection failed
+        ConnectionFailure,
+
+        // A confirmation was received for a session that doesn't exist.
+        SessionNotFound,
+
+        // The session failed for an unknown reason.
+        UnknownError,
+    }
+
+    /// Throws when the session cannot be created.
+    exception SessionCreationException
+    {
+        SessionError error;
+    }
+
     /// The Node interface allows DataStorm nodes to create publisher and subscriber sessions with each other.
     ///
     /// When a node has a writer for a topic that another node is reading, the node initiates the creation of a
@@ -406,8 +431,9 @@ module DataStormContract
         /// reader for which this node has a corresponding topic writer.
         ///
         /// @param publisher The publisher node initiating the session. The proxy is never null.
+        /// @throws SessionCreationException if the session cannot be created.
         /// @see Lookup::announceTopicReader
-        void initiateCreateSession(Node* publisher);
+        ["amd"] void initiateCreateSession(Node* publisher) throws SessionCreationException;
 
         /// Initiates the creation of a subscriber session with a node. The subscriber node sends this request to a
         /// publisher node in one of the following scenarios:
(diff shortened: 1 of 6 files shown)


cpp/src/DataStorm/Contract.ice
```diff
@@ -391,6 +391,31 @@ module DataStormContract
...
+        NodeShutdown,
+
+        // The connection failed
+        ConnectionFailure,
+
+        // A confirmation was received for a session that doesn't exist.
+        SessionNotFound,
+
+        // The session failed for an unknown reason.
+        UnknownError,
```
externl, line 410: Maybe just `Unknown`. We don't prepend Error to anything else.

cpp/src/DataStorm/Contract.ice
```diff
@@ -391,6 +391,31 @@ module DataStormContract
...
+    enum SessionError
+    {
+        // The session is already connected.
+        AlreadyConnected,
+
+        // Node is shutting down.
+        NodeShutdown,
+
+        // The connection failed
+        ConnectionFailure,
```
externl, line 404: Not obvious what this means. Is it more like a HTTP 502 error that a backend server can't be connected to or that an existing connection  has failed.

pepone: Maybe we should have InternalError to group both, this and unknown

cpp/src/DataStorm/Contract.ice
```diff
@@ -391,6 +391,31 @@ module DataStormContract
         void s(long topicId, long elementId, DataSample sample);
     }
 
+    /// The reason for the session creation failure.
+    enum SessionError
```
bernardnormier, line 395: I would call it SessionCreationError.

cpp/src/DataStorm/Contract.ice
```diff
@@ -391,6 +391,31 @@ module DataStormContract
...
+    {
+        // The session is already connected.
```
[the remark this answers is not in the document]
bernardnormier: Yes, although it doesn't matter much since it's internal.

cpp/src/DataStorm/Contract.ice
bernardnormier: Or Internal[Error], like StatusCode::InternalError in IceRPC. Means "some other error without a specific code for the caller".

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -667,6 +667,13 @@ SessionI::retry(NodePrx node, exception_ptr exception)
         {
             rethrow_exception(exception);
         }
+        catch (const DataStormContract::SessionCreationException& ex)
+        {
+            if (ex.error == SessionError::NodeShutdown)
+            {
+                return false;
+            }
```
bernardnormier, line 675: else we ignore this exception. You should add a comment explaining why, to make it clear it's on purpose.

bernardnormier (commented):
It's not clear to me where we retry on SessionCreationException.

pepone:
> It's not clear to me where we retry on SessionCreationException.

In NodeI when there is an exception creation a session we call removePublisherSession or removeSubscriberSession

Both these methods call `Session::retry` and only remove the session if no retry can be attempted.

I will add a comment in the exception handlers/callbacks where we call them.

cpp/src/DataStorm/Contract.ice
```diff
@@ -391,6 +391,31 @@ module DataStormContract
...
+        SessionNotFound,
+
+        // The session failed for an unknown reason.
+        UnknownError,
+    }
+
+    /// Throws when the session cannot be created.
+    exception SessionCreationException
+    {
+        SessionError error;
```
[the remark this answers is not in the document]
pepone: Not sure, I renamed `Unknown` to `Internal` I don't think it is a good idea to pass internal error details as user exceptions.

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -373,50 +403,73 @@ NodeI::createPublisherSession(
...
+            retrySubscriberSessionCreation(publisher, session, current_exception());
+        }
+        // Else node is shutting down.
+        throw SessionCreationException{SessionError::Internal};
     }
 }
 
 void
-NodeI::removeSubscriberSession(const NodePrx& node, const shared_ptr<SubscriberSessionI>& session, exception_ptr ex)
+NodeI::retrySubscriberSessionCreation(
```
pepone, line 420: I split removeSubscriberSession into two methods: retrySubscriberSessionCreation and removeSubscriberSession. The removeSubscriberSession method is now only called when a retry isn’t possible. I think this separation makes the intent clearer.

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -181,20 +209,10 @@ NodeI::createSession(
 
         unique_lock<mutex> lock(_mutex);
         session = createPublisherSessionServant(*subscriber);
```
pepone, line 211: createPublisherSessionServant no longer handles  ObjectAdapterDestroyedException and return null, I think is simpler to let the callers of this method to handle it.

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -223,29 +244,54 @@ NodeI::createSession(
                         uncheckedCast<PublisherSessionPrx>(session->getProxy()),
                         nullptr,
                         [self, subscriber, session](auto ex)
-                        { self->removePublisherSession(*subscriber, session, ex); });
+                        { self->retryPublisherSessionCreation(*subscriber, session, ex); });
```
pepone, line 247: The logic is the same but the intention is now clear. We attempt a retry, and only remove the session if a retry is not possible.

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -315,11 +366,29 @@ NodeI::createSubscriberSession(
             _proxy,
             nullptr,
             [self = shared_from_this(), session, subscriber](exception_ptr ex)
-            { self->removePublisherSession(subscriber, session, ex); });
+            {
+                if (session)
+                {
+                    self->retryPublisherSessionCreation(subscriber, session, ex);
+                }
+                // Else node is shutting down, or the session was established by a concurrent call.
```
pepone, line 374: The logic here remains unchanged. The `if (session)` check corresponds to the previous `if (session && ...)` that was inside removePublisherSession.

This refactor makes it clearer that we do not retry if the initial call to `initiateCreateSessionAsync` fails, as the retry logic is handled within the Session servant class, and at this point, a Session servant has not yet been created.

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -343,17 +412,7 @@ NodeI::createPublisherSession(
         if (!session)
         {
             session = createSubscriberSessionServant(publisher);
```
pepone, line 414: We no longer return nullptr, and just handle the exceptions.

externl (approved):
Looks good to me!

cpp/src/DataStorm/Contract.ice
```diff
@@ -406,8 +428,9 @@ module DataStormContract
         /// reader for which this node has a corresponding topic writer.
         ///
         /// @param publisher The publisher node initiating the session. The proxy is never null.
+        /// @throws SessionCreationException if the session cannot be created.
```
bernardnormier, line 431: For @throws in Slice, I've used the C# style: Thrown when xxx

bernardnormier (approved):
Looks good.

PR #3905, opened 2025-04-24
Simplify compact IDs in C++

bernardnormier:
This PR simplifies the handling of compact IDs in C++. 

It also:
 - updates C# and Java to be in sync in C++
 - updates all languages to compare compactId with `-1` (meaning not set), and not `>= 0`.

### DIFF
--- a/cpp/include/Ice/Initialize.h
+++ b/cpp/include/Ice/Initialize.h
@@ -89,9 +89,6 @@ namespace Ice
         /// @param con The connection associated with this call. May be null.
         std::function<void(std::function<void()> call, const Ice::ConnectionPtr& con)> executor{};
 
-        /// @private
-        std::function<std::string(int id)> compactIdResolver{};
-
         /// The batch request interceptor, which is called by the Ice runtime to enqueue a batch request.
         /// @param req An object representing the batch request.
         /// @param count The number of requests currently in the queue.
--- a/cpp/include/Ice/InputStream.h
+++ b/cpp/include/Ice/InputStream.h
@@ -222,11 +222,10 @@ namespace Ice
         EncodingVersion skipEncapsulation();
 
         /// Reads the start of a value or exception slice.
-        /// @return The Slice type ID for this slice.
-        std::string startSlice()
+        void startSlice()
         {
             assert(_currentEncaps && _currentEncaps->decoder);
-            return _currentEncaps->decoder->startSlice();
+            _currentEncaps->decoder->startSlice();
         }
 
         /// Indicates that the end of a value or exception slice has been reached.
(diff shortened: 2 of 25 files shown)


cpp/include/Ice/Initialize.h
```diff
@@ -89,9 +89,6 @@ namespace Ice
         /// @param con The connection associated with this call. May be null.
         std::function<void(std::function<void()> call, const Ice::ConnectionPtr& con)> executor{};
 
-        /// @private
-        std::function<std::string(int id)> compactIdResolver{};
```
bernardnormier, line 93: It was used previously by Python/Ruby/PHP.

cpp/include/Ice/InputStream.h
```diff
@@ -222,11 +222,10 @@ namespace Ice
         EncodingVersion skipEncapsulation();
 
         /// Reads the start of a value or exception slice.
-        /// @return The Slice type ID for this slice.
-        std::string startSlice()
```
bernardnormier, line 226: The returned value was unused and it made this PR more complicated.

csharp/src/Ice/SlicedData.cs
```diff
@@ -13,12 +13,14 @@ public sealed record class SlicedData(SliceInfo[] slices);
 /// <summary>
 /// Encapsulates the details of a slice with an unknown type.
 /// </summary>
-/// <param name="typeId">The Slice type ID or compact type ID for this slice.</param>
+/// <param name="typeId">The Slice type ID for this slice.</param>
```
bernardnormier, line 16: Restored the previous version.

cpp/include/Ice/SlicedData.h
```diff
@@ -59,6 +60,14 @@ namespace Ice
...
               isLastSlice(isLastSlice)
         {
+            if (compactId == -1)
+            {
+                assert(typeId.empty());
+            }
+            else
+            {
+                assert(!typeId.empty());
+            }
```
externl, line 70:
```suggestion
            assert((compactId == -1) == typeId.empty());
```

Single assert and the compiler no if/else blocks for the compiler to feel tempted to leave in.

externl (approved):

cpp/include/Ice/SlicedData.h
pepone, line 70: I find a bit odd to have if/else that just contain an assert, what about:

```suggestion
            assert((compactId == -1 && typeId.empty()) || (compactId != -1 && !typeId.empty()));
```

pepone (approved):

cpp/include/Ice/SlicedData.h
bernardnormier: I applied Joe's suggestion.

bernardnormier: On second thought, I reverted to an `if`: I prefer two asserts.

Also, in this particular code, we need to use `this->typeId` because `typeId` is the moved-out argument.

PR #3906, opened 2025-04-24
Replace ValueFactoryManager by SliceLoader in Python

bernardnormier:
This PR replaces ValueFactory and ValueFactoryManager by a SliceLoader in Python.


python/modules/IcePy/PySliceLoader.cpp
```diff
@@ -0,0 +1,52 @@
...
+{
+    assert(_sliceLoader != Py_None);
+
+    if (!PyCallable_Check(_sliceLoader))
+    {
+        throw Ice::InitializationException{__FILE__, __LINE__, "sliceLoader must be a callable"};
+    }
+}
+
+IcePy::PySliceLoader::~PySliceLoader() { Py_DECREF(_sliceLoader); }
```
pepone, line 20: My understanding is this runs with the GIL locked, when communicator dealloc erases the map entry

python/modules/IcePy/PySliceLoader.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+Ice::ValuePtr
+IcePy::PySliceLoader::newClassInstance(string_view typeId) const
+{
+    // The unmarshaling always runs with the GIL locked.
+    assert(PyGILState_Check());
+
+    // Get the type information.
+    ValueInfoPtr info =
```
pepone, line 29: We can do this later, after calling the sliceLoader, this wont be used if the loader returns None.

pepone (approved):

python/modules/IcePy/PySliceLoader.cpp
```diff
@@ -0,0 +1,64 @@
...
+    {
+        throw Ice::InitializationException{__FILE__, __LINE__, "sliceLoader must be a callable"};
+    }
+}
+
+IcePy::PySliceLoader::~PySliceLoader()
+{
+    // Often called when the Python interpreter is finalizing.
+#if PY_VERSION_HEX >= 0x030D0000 // Python 3.13 and later
+    if (!Py_IsFinalizing())
```
bernardnormier, line 25: Ice/objects and Ice/slicing/objects fail with Python 3.13 Debug on Windows without this check.

python/modules/IcePy/PySliceLoader.cpp
```diff
@@ -0,0 +1,64 @@
...
+    {
+        Py_DECREF(_sliceLoader);
+    }
+#else
+    Py_DECREF(_sliceLoader);
```
bernardnormier, line 30: Not sure what to do here: we don't want to leak and we don't want to crash during Python shutdown.

python/modules/IcePy/Communicator.cpp
```diff
@@ -1606,6 +1594,22 @@ IcePy::getCommunicatorWrapper(const Ice::CommunicatorPtr& communicator)
...
+    auto p = sliceLoaderMap.find(communicator);
+    assert(p != sliceLoaderMap.end());
+    return p->second;
+}
+
+void
+IcePy::removeSliceLoader(const Ice::CommunicatorPtr& communicator)
+{
+    // Called by destroy[Async], which can be called multiple times.
+    sliceLoaderMap.erase(communicator);
```
pepone, line 1610: This is only safe if the deallocation of the slice loader doesn't release the GIL. It is a bit unpredictable when the GIL is released and reacquired. I think it would be clear to use a regular C++ lock for sliceLoaderMap, specially because the Slice loader can be provided by the user.

externl: Sorry, I'm not quite following what you mean here.

python/modules/IcePy/Communicator.cpp
```diff
@@ -1606,6 +1594,22 @@ IcePy::getCommunicatorWrapper(const Ice::CommunicatorPtr& communicator)
...
+    // It's in the map until the communicator is destroyed.
+    auto p = sliceLoaderMap.find(communicator);
+    assert(p != sliceLoaderMap.end());
+    return p->second;
+}
+
+void
+IcePy::removeSliceLoader(const Ice::CommunicatorPtr& communicator)
+{
+    // Called by destroy[Async], which can be called multiple times.
```
externl, line 1609: Would we assert here?

```
assert(PyGILState_Check());
```

python/modules/IcePy/Communicator.cpp
pepone: The PythonSlice loader can hold a user object, when this object is released it can do:

```
// Pseudo code
{
  AllowThreads allow; // <- Now another thread can reenter removeSliceLoader
  ....
}
// Continue with the GIL locked
```

python/test/Ice/objects/TestI.py
```diff
@@ -39,6 +39,16 @@ def ice_preMarshal(self):
     def ice_postUnmarshal(self):
         self.postUnmarshalInvoked = True
 
+def customSliceLoader(typeId):
+    if typeId == "::Test::B":
```
externl, line 43: Could use a switch statement here.

externl (approved):

python/modules/IcePy/Communicator.cpp
bernardnormier: Yes, the Python SliceLoader, when deallocated, can temporarily release the GIL.

I don't see why this is a concern here. I believe all this code, part of the "Python communicator destroy", always run prior to the finalization of the Python interpreter.

pepone: > I don't see why this is a concern here

My concern is that `sliceLoaderMap.erase` is only protected by the GIL, if it can be released while erase deallocates the SliceLoader, another thread can reenter erase.

python/test/Ice/objects/TestI.py
bernardnormier: Fixed as suggested

externl (approved):

pepone (approved):

PR #3909, opened 2025-04-25
Move IceMX code back to com.zeroc.IceMX package.

### LINKED ISSUE #3873: Mapping IceMX to Java
IceMX suffers from a split-package issue in Ice 3.7, which was fixed by #1963 and #1973.

The #1973 solution is to map a single module (IceMX) to a list of package-prefixes: `com.zeroc.Ice[.IceMX]`, `com.zeroc.IceStorm[.IceMX]`, etc.

Note that in Ice 3.7, an Ice.Package.module's value is a single package-prefix, not a list. And IceMX maps to `com.zeroc.IceMX`.

Furthermore, the ModuleToPackageSliceLoader introduced in #3872 maps a module (including Ice.Package.module) to a single package, not a list.

## Proposal

Revise the split-package module fix by mapping  module IceMX to `com.zeroc.IceMX` (like in Ice 3.7).  And include all the Metrics generated code (from Ice, IceStorm, Glacier2) in Ice.jar.

pepone:
Fix #3873


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/IceMX/MetricsHelper.java
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) ZeroC, Inc.
 
-package com.zeroc.Ice.IceMX;
+package com.zeroc.IceMX;
```
bernardnormier, line 3: Should we move this file to java/src/com/zeroc/ice/src/main/com/zeroc/IceMX/MetricsHelper.java ?

java/src/com.zeroc.ice/build.gradle
```diff
@@ -8,7 +8,11 @@ project.ext.description = "Ice is a comprehensive RPC framework that helps you b
 sourceSets {
     main {
         slice {
-            srcDir "$project.ext.topSrcDir/slice/Ice"
+            srcDir "$project.ext.topSrcDir/slice/"
+
+            includes = ["**/Ice/*.ice",
```
bernardnormier, line 13: Are the leading `**` necessary?

bernardnormier (approved):

externl (approved):

java/src/com.zeroc.ice/build.gradle
pepone: They are not, removed.

PR #3910, opened 2025-04-25
Replace ValueFactory(Manager) by SliceLoader in Ruby

bernardnormier:
This is the Ruby version of #3906.


ruby/test/Ice/objects/Client.rb
```diff
@@ -6,10 +6,14 @@
...
 require './AllTests'
+require './TestI.rb'
 
 class Client < ::TestHelper
     def run(args)
-        self.init(args:args) do |communicator|
+        initData = Ice::InitializationData.new
+        initData.properties = self.createTestProperties(args:args)
+        initData.sliceLoader = CustomSliceLoader.new
+        self.init(initData:initData, args:args) do |communicator|
```
externl, line 16: No need to pass args a second time.

ruby/test/Ice/objects/TestI.rb
```diff
@@ -47,3 +47,15 @@ def ice_postUnmarshal
         @postUnmarshalInvoked = true
     end
 end
+
+class CustomSliceLoader
+    def newInstance(typeId)
+        if typeId == '::Test::B'
```
externl, line 53: I would use a case statement.

externl (approved):

ruby/src/IceRuby/Communicator.cpp
```diff
@@ -8,32 +8,22 @@
...
-    {
-        auto vfm = dynamic_pointer_cast<ValueFactoryManager>((*communicator)->getValueFactoryManager());
-        assert(vfm);
-        vfm->markSelf();
-    }
-    catch (const Ice::CommunicatorDestroyedException&)
-    {
-        // Ignore. This is expected.
-    }
+    using SliceLoaderMap = map<Ice::CommunicatorPtr, Ice::SliceLoaderPtr>;
```
pepone, line 25: Do you prefer to have the `using` even if it is only used in one place, like in this case?

pepone (approved):

ruby/src/IceRuby/Communicator.cpp
bernardnormier: Removed.

PR #3911, opened 2025-04-25
Replace ValueFactory(Manager) by SliceLoader in PHP

bernardnormier:
This is the PHP counterpart of #3906 and #3910.


php/lib/Ice/InitializationData.php
```diff
@@ -3,16 +3,26 @@
...
     {
         $this->properties = $properties;
         $this->logger = $logger;
+        $this->sliceLoader = $sliceLoader;
     }
 
-    public $properties;
-    public $logger;
+    public ?\IcePHP_Properties $properties;
+    public ?\IcePHP_Logger $logger;
```
bernardnormier, line 24: There is no test for this field so the use-case is unclear.

php/src/Communicator.cpp
```diff
@@ -52,21 +53,10 @@ namespace IcePHP
...
-
-    class CustomValueFactory;
-    using CustomValueFactoryPtr = shared_ptr<CustomValueFactory>;
-
     // CommunicatorInfoI encapsulates communicator-related information that
     // is specific to a PHP "request". In other words, multiple PHP requests
     // might share the same communicator instance but still need separate
-    // workspaces. For example, we don't want the value factories installed
-    // by one request to influence the behavior of another request.
+    // workspaces.
```
bernardnormier, line 59: This logic is not clear to me.

php/src/Communicator.cpp
```diff
@@ -1155,9 +943,25 @@ ZEND_FUNCTION(Ice_initialize)
...
+        member = "sliceLoader";
+        {
+            if ((data = zend_hash_str_find(Z_OBJPROP_P(zvinit), member.c_str(), member.size())) != 0)
+            {
+                assert(Z_TYPE_P(data) == IS_INDIRECT);
+                initDataSliceLoader = Z_INDIRECT_P(data);
+
+                if (!ZVAL_IS_NULL(initDataSliceLoader))
+                {
+                    if (Z_TYPE_P(initDataSliceLoader) != IS_OBJECT)
```
bernardnormier, line 956: I could not find a way to check the type implements interface SliceLoader. But it's not need now that the type enforcement is performed at the PHP level.

pepone: > But it's not need now that the type enforcement is performed at the PHP level.

Why do we need to check it is an object then?

php/src/PHPSliceLoader.cpp
```diff
@@ -0,0 +1,77 @@
...
+            const_cast<char*>("newInstance"),
+            sizeof("newInstance") - 1,
+            &obj,
+            1,
+            &arg,
+            0);
+    }
+    zend_catch
+    {
+        // obj;
```
pepone, line 45: What is this comment about?

pepone (approved):

php/src/PHPSliceLoader.cpp
externl, line 45: Missing error handling?

php/src/PHPSliceLoader.h
```diff
@@ -0,0 +1,29 @@
...
+    /// Implements SliceLoader using a PHP SliceLoader.
+    class PHPSliceLoader final : public Ice::SliceLoader
+    {
+    public:
+        PHPSliceLoader(zval* sliceLoader, const CommunicatorInfoPtr& communicatorInfo);
+        ~PHPSliceLoader() final;
+
+        [[nodiscard]] Ice::ValuePtr newClassInstance(std::string_view typeId) const final;
+
+        // TODO: we currently don't support loading of custom Slice exceptions from Python.
```
externl, line 21:
```suggestion
        // TODO: we currently don't support loading of custom Slice exceptions from PHP.
```

externl (approved):

php/src/Communicator.cpp
bernardnormier: I'll remove it.

PR #3913, opened 2025-04-27
Replace ValueFactory(Manager) by SliceLoader in Swift

bernardnormier:
This PR replaces ValueFactory and ValueFactory by SliceLoader in Swift.

Since the loading is "dynamic" in Swift (resolves method names at runtime), the SliceLoader implementations perform caching like in Java.


cpp/src/slice2swift/Gen.cpp
```diff
@@ -241,52 +241,37 @@ Gen::TypesVisitor::visitClassDefStart(const ClassDefPtr& p)
 
     ClassDefPtr base = p->base();
 
-    out << sp;
-    out << nl << "@_documentation(visibility: internal)";
-    out << nl << "public class " << removeEscaping(name)
```
bernardnormier, line 246: It's not clear to me why we generated this extra class, with a single method that returns the type object.

This PR simplifies the logic by generating an extension method that returns the type object (`AnyObject.Type?`).

cpp/src/slice2swift/Gen.cpp
```diff
@@ -241,52 +241,37 @@ Gen::TypesVisitor::visitClassDefStart(const ClassDefPtr& p)
...
         if (++it != parts.end())
         {
-            factory << "_";
+            methodName << "_";
         }
     }
 
     out << sp;
-    out << nl << "public extension " << getUnqualified("Ice.ClassResolver", swiftModule);
+    out << nl << "public extension " << getUnqualified("Ice.DefaultSliceLoader", swiftModule);
```
bernardnormier, line 259: It's now an extension for DefaultSliceLoader and the ClassResolver class is gone.

swift/src/Ice/DefaultSliceLoader.swift
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) ZeroC, Inc.
+
+import Foundation
+
+/// Implements SliceLoader using the generated code.
+public final class DefaultSliceLoader: NSObject, SliceLoader {
```
bernardnormier, line 6: It's public for the tests that rely of the prefix.

swift/test/TestCommon/TestCommon.swift
```diff
@@ -38,7 +38,7 @@ public enum TestFailed: Error {
     case testFailed
 }
 
-public protocol TestHelper {
+public protocol TestHelper: AnyObject {
```
bernardnormier, line 41: AnyObject is required for the `unowned` property in the Ice/optional test.

swift/src/Ice/DefaultSliceLoader.swift
```diff
@@ -0,0 +1,61 @@
...
+            // Regular type IDs start with "::", while compact IDs don't.
+            let flatName =
+                typeId.hasPrefix("::") ? typeId.dropFirst(2).replacingOccurrences(of: "::", with: "_") : typeId
+
+            let selector = Selector("resolveTypeId\(prefix)_\(flatName)")
+
+            guard DefaultSliceLoader.responds(to: selector) else {
+                return nil
+            }
+            return DefaultSliceLoader.perform(selector).takeUnretainedValue() as? AnyObject.Type
```
bernardnormier, line 58: The autoreleasepool and `takeUnretainedValue()` comes from the previous ClassResolver code. I assume they are correct.

swift/src/Ice/NotFoundSliceLoaderDecorator.swift
```diff
@@ -0,0 +1,49 @@
...
+                } else {
+                    if let logger = logger {
+                        self.logger = nil  // we log once
+                        logger.warning(
+                            """
+                            SliceLoader: Type ID '\(typeId)' not found and the not found cache is full. \
+                            The cache size is set to \(cacheSize). You can increase the cache size by setting property \
+                            Ice.SliceLoader.NotFoundCacheSize.
+                            """
+                        )
```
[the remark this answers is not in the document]
bernardnormier: There is no concurrency since this code is within a `mutex.sync` block.

However, it's not correct to log within `mutex.sync` so I should indeed rework this code to log outside the lock.

swift/src/Ice/Initialize.swift
```diff
@@ -162,6 +162,22 @@ private func initializeImpl(
 
         precondition(initData.logger != nil && initData.properties != nil)
 
+        if let sliceLoader = initData.sliceLoader {
+            initData.sliceLoader = CompositeSliceLoader(sliceLoader, DefaultSliceLoader())
+        } else {
+            initData.sliceLoader = DefaultSliceLoader()
+        }
```
pepone, line 169: Maybe using map and nil coalescing is more "Swifty"

```
initData.sliceLoader = initData.sliceLoader.map {
    CompositeSliceLoader($0, DefaultSliceLoader()) } ?? DefaultSliceLoader()
```

pepone (approved):

swift/src/Ice/DefaultSliceLoader.swift
externl: IIRC `perform` returns an `Unmanaged` type. It's just a Swift object we'r getting back so no special memory management is required.

externl (approved):

swift/src/Ice/NotFoundSliceLoaderDecorator.swift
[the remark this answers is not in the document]
bernardnormier: Fixed

swift/src/Ice/Initialize.swift
bernardnormier: I find my version more readable, if not as Swifty.

PR #3918, opened 2025-04-29
Minor Debugging Improvements to the Slice Scanner & the Test's `SliceTranslator`

InsertCreativityHere:
This PR:

1) When a test wants to run a Slice compiler, it does so through a `SliceTranslator`. This class passes `quiet=True` when it runs,
eating any command line output. I changed this default to `False`, and added an argument that can be used to override it.
Setting `quiet=True` makes it very difficult to debug test failures (since you can't see the error messages).
But, being able to turn it on is important for tests like `errorDetection`, where we expect lots of failures.
The metric is: Do we expect this test to fail? It should run quietly. Do we expect this test to pass? It should be run normally, so any unexpected errors are visible in the test output.

2) I added debugging functionality to the Slice scanner, so it now prints it's state-stack when a Slice compiler is run with `--debug` mode. This is only really useful for me. But it offers valuable insight into if/when the scanner switches states and what states it's switching into. All information that is not reported by the scanner right now.
Instead of calling `yy_push_state` directly, there's an intermediate function `pushState` which calls this function and afterwards emits debugging information.


cpp/src/Slice/Scanner.cpp
InsertCreativityHere, line 1: Generated from `Scanner.l`.

cpp/src/Slice/Scanner.l
```diff
@@ -693,6 +696,36 @@ namespace
         }
     }
 
+    void printStateStack()
```
InsertCreativityHere, line 699: The debugging output looks like:
```
...
--pushed: current state stack [ 3 (0) ]
--pushed: current state stack [ 3 0 (4) ]
--accepting rule at line 419 ("                 // Not new.
")                                              // Not new.
--popped: current state stack [ 3 (0) ]
...
```

cpp/src/Slice/Scanner.l
```diff
@@ -693,6 +696,36 @@ namespace
...
+        {
+            fprintf(stderr, "%d ", yy_start_stack[i]);
+        }
+        fprintf(stderr, "(%d) ]\n", YY_START);
+    }
+
+    void pushState(int newState)
+    {
+        yy_push_state(newState);
+        if (yy_flex_debug)
```
InsertCreativityHere, line 712: This `yy_flex_debug` bool is controlled by passing `--debug` on the command line.
Ex: `slice2cpp --debug Test.ice`.

externl (approved):

pepone:
I don't see how `--debug` is handled — this PR doesn't include any code for it. The description also mentions changes related to "quiet" mode, but I don't see those changes here either.

Also, I'm not a big fan of including code solely for internal debugging purposes. We usually avoid this kind of thing.

externl:
> I don't see how `--debug` is handled — this PR doesn't include any code for it. 
This is an already existing flag for `slice2cpp`

> The description also mentions changes related to "quiet" mode, but I don't see those changes here either.
It's the python changes. 

> Also, I'm not a big fan of including code solely for internal debugging purposes. We usually avoid this kind of thing.
This is an already existing feature.

InsertCreativityHere:
So, indeed, `--debug` already exists, and is used to set a special scanner flag (`yy_flex_debug`) which is built in to Flex (which generates our scanner for us). The `--debug` flag also does other things, but they aren't relevant here.
And `yy_flex_debug` tells the scanner to output internal debugging information while running.

All this is already true.
All I did was add a little more debugging to the scanner, so that now, it will also report when it switches between states (comment, preprocessor, string, etc. things which require different parsing than normal slice)

InsertCreativityHere:
> Also, I'm not a big fan of including code solely for internal debugging purposes. We usually avoid this kind of thing.

Fair enough. We could delete the extra debugging I added. But, everytime something goes wrong with the scanner, I hack some version of this back into it, so I can see what's going on. This time I figured I might as well just commit it.

If you feel strongly that we shouldn't be doing this, I'll remove it from the PR and relegate it to a `txt` file on my system instead
: vP

pepone:
I'm fine with keeping this extra debug if you thing is worth it.

PR #3919, opened 2025-04-29
Update MATLAB toolbox logo

### LINKED ISSUE #3806: Incorrect icon in MATLAB toolbox (Windows)
The icon is not the ice icon as it should be. It's also distorted.

pepone:
Fix #3806

matlab-ice-toolbox

MATLAB was resizing our icon to 160x120 an adding a grey background. I exported the new Ice icon centered in a 160x120 white box.


externl (approved):

matlab/msbuild/ice.proj
```diff
@@ -20,7 +20,7 @@
         </When>
         <Otherwise>
             <PropertyGroup>
-                <MatlabCommand>&quot;$(MatlabHome)\bin\matlab.exe&quot; -nodesktop -nosplash -wait -log -minimize</MatlabCommand>
+                <MatlabCommand>&quot;$(MatlabHome)\bin\matlab.exe&quot; -nodesktop -nosplash -wait -log -minimize -r</MatlabCommand>
```
pepone, line 23: `-r` is required when building the toolbox locally.  It should not affect CI builds because we use the When above by setting `MATLAB_COMMAND`

bernardnormier (approved):

PR #3925, opened 2025-04-30
Remove Ruby InitializationData logger

pepone:
This PR removes the logger data member from Ruby InitializationData, as there is no point in providing a data member that cannot be set.

### DIFF
--- a/ruby/.gitignore
+++ b/ruby/.gitignore
@@ -1,2 +1,6 @@
 dist
 *.gem
+ruby/Glacier2/*
+ruby/IceBox/*
+ruby/IceGrid/*
+ruby/IceStorm/*
--- a/ruby/src/IceRuby/Communicator.cpp
+++ b/ruby/src/IceRuby/Communicator.cpp
@@ -155,20 +155,13 @@ IceRuby_initialize(int argc, VALUE* argv, VALUE /*self*/)
 
         if (!NIL_P(initData))
         {
-            volatile VALUE properties = callRuby(rb_iv_get, initData, "@properties");
-            volatile VALUE logger = callRuby(rb_iv_get, initData, "@logger");
-            volatile VALUE initDataSliceLoader = callRuby(rb_iv_get, initData, "@sliceLoader");
-
+            VALUE properties = callRuby(rb_iv_get, initData, "@properties");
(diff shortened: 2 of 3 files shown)


ruby/.gitignore
```diff
@@ -1,2 +1,6 @@
 dist
 *.gem
+ruby/Glacier2/*
+ruby/IceBox/*
+ruby/IceGrid/*
+ruby/IceStorm/*
```
pepone, line 6: We added back this modules but .gitignore was not updated.

ruby/src/IceRuby/Communicator.cpp
```diff
@@ -155,20 +155,13 @@ IceRuby_initialize(int argc, VALUE* argv, VALUE /*self*/)
 
         if (!NIL_P(initData))
         {
-            volatile VALUE properties = callRuby(rb_iv_get, initData, "@properties");
-            volatile VALUE logger = callRuby(rb_iv_get, initData, "@logger");
-            volatile VALUE initDataSliceLoader = callRuby(rb_iv_get, initData, "@sliceLoader");
-
+            VALUE properties = callRuby(rb_iv_get, initData, "@properties");
```
pepone, line 158: `volatile` here is not required. This code just gets a reference to the properties ruby object pass by the client initData, and unwraps it to get the underlying C++ properties.

ruby/src/IceRuby/Communicator.cpp
```diff
@@ -155,20 +155,13 @@ IceRuby_initialize(int argc, VALUE* argv, VALUE /*self*/)
...
             {
                 data.properties = getProperties(properties);
             }
 
-            if (!NIL_P(logger))
-            {
-                throw RubyException(rb_eArgError, "custom logger is not supported");
-            }
-
+            VALUE initDataSliceLoader = callRuby(rb_iv_get, initData, "@sliceLoader");
```
pepone, line 164: volatile isn't required here either. We get a reference to the Slice Loader in the user initData. Later the call to `make_shared<RubySliceLoader>(initDataSliceLoader)` register the address to keep it around for the life time of the communicator.

externl (approved):

bernardnormier (approved):

PR #3926, opened 2025-04-30
Remove ClassInfo::interface, which was always false.

### LINKED ISSUE #3897: Remove interface field from PHP ClassInfo
It's always set to false.

See https://github.com/zeroc-ice/ice/blob/1392e3890d5f2b86d4e473b859ff20c3064e73ea/cpp/src/slice2php/Main.cpp#L273

pepone:
Fix #3897

I will rename ClassInfo to ValueInfo in a follow-up PR.


externl (approved):

bernardnormier (approved):
I would keep ClassInfo, defineClass, etc.. Using `class` is fine.

PR #3927, opened 2025-04-30
Compute VCInstallDir when building MSI packages

### LINKED ISSUE #3538: Compute VCInstallDir in GitHub actions
The build-msi-package needs to know the VC install location, to be able to locate the VC redistributable merge modules.

When using VC command prompt you can use VCInstallDir but this variable doesn't exits in the GitHub actions shell.

Instead of using a hardcoded value we can compute it from the output of:

```
"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" --latest
```

pepone:
Fix #3538

### DIFF
--- a/.github/workflows/build-cpp-windows-packages.yml
+++ b/.github/workflows/build-cpp-windows-packages.yml
@@ -82,13 +82,18 @@ jobs:
           ./gradlew.bat dist -PcppPlatform=x64 -PcppConfiguration=Release
         working-directory: ice/java
 
+      # MSI packaging depends on VCInstallDir to locate The C++ merge modules
+      - name: Add VCInstallDir to Environment
+        run: |
+          $vsPath = & "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath
+          echo "VCInstallDir=$vsPath\VC" | Out-File -FilePath $env:GITHUB_ENV -Append
+        shell: powershell
+
       - name: Build MSI
         run: dotnet build -c Release -p:Platform=x64
         working-directory: ice/packaging/msi
         env:
           StagingDir: "${{ github.workspace }}\\staging"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-cpp-windows-packages.yml
```diff
@@ -82,13 +82,18 @@ jobs:
           ./gradlew.bat dist -PcppPlatform=x64 -PcppConfiguration=Release
         working-directory: ice/java
 
+      # MSI packaging depends on VCInstallDir to locate The C++ merge modules
+      - name: Add VCInstallDir to Environment
+        run: |
+          $vsPath = & "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath
+          echo "VCInstallDir=$vsPath\VC" | Out-File -FilePath $env:GITHUB_ENV -Append
```
externl, line 89: https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions?tool=powershell#setting-an-environment-variable

according to this to latest format to use with PowerShell 6 is

```
"{environment_variable_name}={value}" >> $env:GITHUB_ENV
```

externl (approved):

.github/workflows/build-cpp-windows-packages.yml
pepone: Do we know if PowerShell 6 is the default with this runner?

externl: In other words actions we do 
echo "USING_BUILD_CACHE=true" >> $env:GITHUB_ENV
        echo "CLToolExe=C:\cl.exe" >> $env:GITHUB_ENV
        echo "CLToolPath=C:\" >> $env:GITHUB_ENV

PR #3929, opened 2025-04-30
Fix macos ld warning

### LINKED ISSUE #3900: macOS Linker warning
We're getting this with the recent version of Xcode when building C++/Ruby/Python. 

```
ld: warning: reducing alignment of section __DATA,__common from 0x8000 to 0x4000 because it exceeds segment maximum alignment
```

Previously we were seeing a similar warning during the Swift build of slice2swift. I thought I opened an issue for it but I can't find it now.

bernardnormier:
Fixes #3900.

We still get this warning when linking slice2swift as part of the Swift build. We could suppress it the same way by updating Package.swift.


bernardnormier:
This issue is apparently specific to mcpp-based code, and the new linker flag is only available with the latest `ld` on macos.

PR #3930, opened 2025-04-30
Allow `java:identifier` on Classes & Exceptions

### LINKED ISSUE #3534: Allow `java:identifier` on Classes and Exceptions

InsertCreativityHere:
Fixes #3534 now that we have the machinery to handle remapped type-ids (ie. `SliceLoader`).
This PR also disallows mixing `java:package` with `java:identifier`, and updates other languages accordingly.

----

Note: this doesn't actually update any Slice files. this was split off into a separate PR (see #3931).
Switching from `java:package` to `java:identifier` has to be done at the same time as the switch from `setProperty("Ice.Package...")` to `SliceLoader`s.
Needless to say, this changes many lines of code in the tests.

### DIFF
--- a/cpp/src/slice2cs/CsUtil.cpp
+++ b/cpp/src/slice2cs/CsUtil.cpp
@@ -1798,6 +1798,12 @@ Slice::CsGenerator::validateMetadata(const UnitPtr& u)
             const string msg = "'cs:namespace' is deprecated; use 'cs:identifier' to remap modules instead";
             p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
 
+            if (auto cont = dynamic_pointer_cast<Contained>(p); cont && cont->hasMetadata("cs:identifier"))
+            {
+                return "the 'cs:namespace' metadata cannot be used alongside 'cs:identifier' - both change the name of "
+                       "the mapped module";
+            }
+
             // 'cs:namespace' can only be applied to top-level modules
             // Top-level modules are contained by the 'Unit'. Non-top-level modules are contained in 'Module's.
             if (auto mod = dynamic_pointer_cast<Module>(p); mod && !mod->isTopLevel())
--- a/cpp/src/slice2cs/CsUtil.cpp
+++ b/cpp/src/slice2cs/CsUtil.cpp
@@ -1800,8 +1800,7 @@ Slice::CsGenerator::validateMetadata(const UnitPtr& u)
 
             if (auto cont = dynamic_pointer_cast<Contained>(p); cont && cont->hasMetadata("cs:identifier"))
             {
-                return "the 'cs:namespace' metadata cannot be used alongside 'cs:identifier' - both change the name of "
-                       "the mapped module";
+                return "A Slice element can only have one of 'cs:namespace' and 'cs:identifier' applied to it";
             }
 
             // 'cs:namespace' can only be applied to top-level modules
--- a/cpp/src/slice2java/JavaUtil.cpp
+++ b/cpp/src/slice2java/JavaUtil.cpp
@@ -1722,8 +1722,9 @@ Slice::JavaGenerator::validateMetadata(const UnitPtr& u)
         .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
         .extraValidation = [](const MetadataPtr& metadata, const SyntaxTreeBasePtr& p) -> optional<string>
         {
-            const string msg = "'java:package' is deprecated; use 'java:identifier' to remap modules instead";
-            p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
+            // TODO uncomment after replacing 'java:package' with 'java:identifier'.
+            // const string msg = "'java:package' is deprecated; use 'java:identifier' to remap modules instead";
+            // p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
 
             if (auto cont = dynamic_pointer_cast<Contained>(p); cont && cont->hasMetadata("java:identifier"))
             {
--- a/cpp/src/slice2java/JavaUtil.cpp
+++ b/cpp/src/slice2java/JavaUtil.cpp
@@ -1702,6 +1702,8 @@ Slice::JavaGenerator::validateMetadata(const UnitPtr& u)
             {typeid(Module),
              typeid(InterfaceDecl),
              typeid(Operation),
+             typeid(ClassDecl),
+             typeid(Slice::Exception),
              typeid(Struct),
              typeid(Sequence),
              typeid(Dictionary),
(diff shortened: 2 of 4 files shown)


cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -1720,6 +1722,16 @@ Slice::JavaGenerator::validateMetadata(const UnitPtr& u)
         .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
         .extraValidation = [](const MetadataPtr&, const SyntaxTreeBasePtr& p) -> optional<string>
         {
+            // TODO uncomment after replacing 'java:package' with 'java:identifier'.
+            // const string msg = "'java:package' is deprecated; use 'java:identifier' to remap modules instead";
+            // p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
+
+            if (auto cont = dynamic_pointer_cast<Contained>(p); cont && cont->hasMetadata("java:identifier"))
+            {
+                return "the 'java:package' metadata cannot be used alongside 'java:identifier' - both change the name "
```
externl, line 1731: Isn't java:package more of a prefix?

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1798,6 +1798,12 @@ Slice::CsGenerator::validateMetadata(const UnitPtr& u)
             const string msg = "'cs:namespace' is deprecated; use 'cs:identifier' to remap modules instead";
             p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
 
+            if (auto cont = dynamic_pointer_cast<Contained>(p); cont && cont->hasMetadata("cs:identifier"))
+            {
+                return "the 'cs:namespace' metadata cannot be used alongside 'cs:identifier' - both change the name of "
+                       "the mapped module";
```
bernardnormier, line 1804: should be "mapped namespace"

bernardnormier: I also find the message inaccurate. We could support `java:package` on an outer module (or at file level) and `csharp:identifier` on an inner module, but we chose not to. What about simply:

```
'cs:namespace' and 'cs:identifier' cannot be used together
```

cpp/src/slice2java/JavaUtil.cpp
bernardnormier: See C# comment. And in Java, it's mapped package.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -1720,6 +1722,16 @@ Slice::JavaGenerator::validateMetadata(const UnitPtr& u)
         .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
         .extraValidation = [](const MetadataPtr&, const SyntaxTreeBasePtr& p) -> optional<string>
         {
+            // TODO uncomment after replacing 'java:package' with 'java:identifier'.
+            // const string msg = "'java:package' is deprecated; use 'java:identifier' to remap modules instead";
+            // p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
+
+            if (auto cont = dynamic_pointer_cast<Contained>(p); cont && cont->hasMetadata("java:identifier"))
```
bernardnormier, line 1729: Please name this variable `contained`.

bernardnormier (commented):
What about JavaScript/TypeScript?

https://doc.zeroc.com/ice/3.7/language-mappings/javascript-mapping/client-side-slice-to-javascript-mapping/javascript-mapping-for-modules

InsertCreativityHere:
> What about JavaScript/TypeScript?

The `js:module` metadata operates in a fundamentally different way than `cs:namespace`, `java:package`, etc.
For Typescript, Slice modules are mapped to _namespaces_, which are different than _modules_.

Looking through the generated code, `js:module` only affects the importing/exporting, not the scoping/names of things.
So there's no 'collision' risk of using both `js:module` and `js:identifier`. They accomplish different things.

pepone:
Just to clarify, "js:module" controls how the generated code is imported in JavaScript. Multiple Slice modules can be packaged together into a single JavaScript module.

For example, all built-in Slice modules are packaged in the `@zeroc/ice` JavaScript module, and can later be imported with:

```
import {Ice, Glacier2 } from "@zeroc/ice";
```

The slice2js compiler uses the `js:module` information to determine which JavaScript module is required for a given Slice module. It conditionally imports the generated code directly when the modules belong to the same JavaScript module, or imports the module itself when they are packaged separately.

cpp/src/slice2cs/CsUtil.cpp
InsertCreativityHere: I went with: `"A Slice element can only have one of 'cs:namespace' and 'cs:identifier' applied to it"`

It was pointed out that "cannot be used alongside" is vague. And I think "cannot be used together" is too.
This new message is more specific about the problem and how to fix it, but still simpler than my original message.

cpp/src/slice2java/JavaUtil.cpp
InsertCreativityHere: This was just copy-pasted from elsewhere. Instead of `cont` I went with `element`.

So, still no short-hand naming! But `element` is semantically closer to what we're getting here.
Even if the actual type just happens to be `Contained`.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1798,6 +1798,11 @@ Slice::CsGenerator::validateMetadata(const UnitPtr& u)
             const string msg = "'cs:namespace' is deprecated; use 'cs:identifier' to remap modules instead";
             p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
 
+            if (auto cont = dynamic_pointer_cast<Contained>(p); cont && cont->hasMetadata("cs:identifier"))
+            {
+                return "A Slice element can only have one of 'cs:namespace' and 'cs:identifier' applied to it";
```
externl, line 1803: Is it better to say module here as it's the only thing it can be?
```suggestion
                return "A Slice module can only have one of 'cs:namespace' and 'cs:identifier' applied to it";
```

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -1720,6 +1722,15 @@ Slice::JavaGenerator::validateMetadata(const UnitPtr& u)
         .acceptedArgumentKind = MetadataArgumentKind::SingleArgument,
         .extraValidation = [](const MetadataPtr&, const SyntaxTreeBasePtr& p) -> optional<string>
         {
+            // TODO uncomment after replacing 'java:package' with 'java:identifier'.
+            // const string msg = "'java:package' is deprecated; use 'java:identifier' to remap modules instead";
+            // p->unit()->warning(metadata->file(), metadata->line(), Deprecated, msg);
+
+            if (auto element = dynamic_pointer_cast<Contained>(p); element && element->hasMetadata("java:identifier"))
+            {
+                return "A Slice element can only have one of 'java:package' and 'java:identifier' applied to it";
```
externl, line 1731: Ditto?

externl (approved):
Looks good. I prefer "A Slice module ..." as this is all about modules.

cpp/src/slice2cs/CsUtil.cpp
InsertCreativityHere: The issue is that at this stage of validation, we don't actually know what the type is. It might not be a module.
It could be anything. Maybe the user thinks they need to put `java:package` on everything and incorrectly writes:
```slice
["java:package:com.austininc"]
module HelloThere
{
    ["java:package:com.austininc"]
    ["java:identifier:FailureException"]
    exception Failure {} 
}
```
It would be weird and misleading for the error to talk about modules, when the real problem is with the exception!

bernardnormier (approved):

PR #3931, opened 2025-04-30
Replace `'java:package' -> 'java:identifier'` and `setProperty("Ice.Package.xxx") -> initData.sliceLoader`

InsertCreativityHere:
This PR:
- Deprecates the old `java:package` metadata.
- Replaces the `java:package` metadata with the `java:identifier` metadata.
- Updates all the java tests to use `initData.sliceLoader = ...` instead of `properties.setProperty("Ice.Package.XXX", "YYY")`.

In 64% of the tests, I just deleted `"Ice.Package.XXX"` without adding `sliceLoader`, because they didn't actually need it
(ie. they never unmarshalled a class/exception). I'm sure there's still more tests where it could of been deleted, but where it wasn't as obvious. But, I don't want to waste too much time going after these, when there's more important stuff to focus on.

### DIFF
--- a/java/test/src/main/java/test/Ice/ami/Server.java
+++ b/java/test/src/main/java/test/Ice/ami/Server.java
@@ -13,7 +13,7 @@ import test.TestHelper;
 public class Server extends TestHelper {
     public void run(String[] args) {
         var initData = new InitializationData();
-        initData.sliceLoader = new ModuleToPackageSliceLoader("::Test", "test.Ice.adapterDeactivation.Test");
+        initData.sliceLoader = new ModuleToPackageSliceLoader("::Test", "test.Ice.ami.Test");
         initData.properties = createTestProperties(args);
         // This test kills connections, so we don't want warnings.
         initData.properties.setProperty("Ice.Warn.Connections", "0");
--- a/java/test/src/main/java/test/Ice/ami/Server.java
+++ b/java/test/src/main/java/test/Ice/ami/Server.java
@@ -3,28 +3,25 @@
 package test.Ice.ami;
 
 import com.zeroc.Ice.Communicator;
+import com.zeroc.Ice.InitializationData;
+import com.zeroc.Ice.ModuleToPackageSliceLoader;
 import com.zeroc.Ice.ObjectAdapter;
-import com.zeroc.Ice.Properties;
 import com.zeroc.Ice.Util;
 
 import test.TestHelper;
 
 public class Server extends TestHelper {
     public void run(String[] args) {
-        Properties properties = createTestProperties(args);
-        properties.setProperty("Ice.Package.Test", "test.Ice.ami");
-        //
+        var initData = new InitializationData();
+        initData.sliceLoader = new ModuleToPackageSliceLoader("::Test", "test.Ice.adapterDeactivation.Test");
+        initData.properties = createTestProperties(args);
         // This test kills connections, so we don't want warnings.
-        //
-        properties.setProperty("Ice.Warn.Connections", "0");
-
-        //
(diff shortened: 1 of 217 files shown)


java/test/src/main/java/test/Slice/generation/list-generated.out
```diff
@@ -1,15 +1,12 @@
...
   <source name="File1.ice">
-    <file name="classes/test/Slice/generation/Test/_Marker.java"/>
     <file name="classes/test/Slice/generation/Test/Interface1Prx.java"/>
     <file name="classes/test/Slice/generation/Test/_Interface1PrxI.java"/>
-    <file name="classes/test/Slice/generation/modpkg/Test2/_Marker.java"/>
     <file name="classes/test/Slice/generation/modpkg/Test2/Class1.java"/>
     <file name="classes/test/Slice/generation/Test/Interface1.java"/>
   </source>
   <source name="File2.ice">
-    <file name="classes/test/Slice/generation/Test/_Marker.java"/>
```
InsertCreativityHere, line 12: The `_Marker` files are only generated when you use `java:package`.
After switching to `java:identifier` they are no longer generated in this test.

scripts/tests/Slice/generation.py
```diff
@@ -8,7 +8,7 @@ class SliceGenerationTestCase(ClientTestCase):
     def runClientSide(self, current):
         current.write("testing list-generated... ")
 
-        slice2java = SliceTranslator("slice2java")
+        slice2java = SliceTranslator("slice2java", quiet=True)
```
InsertCreativityHere, line 11: Unrelated to the PR, but passing `--list-generated` causes the `depends` content to be written to `stdout`.
Passing `true` here prevents it from getting dumped into the testing output.

java/test/src/main/java/test/IceBox/configuration/TestServiceI.java
```diff
@@ -10,8 +10,6 @@
 public class TestServiceI implements Service {
     @Override
     public void start(String name, Communicator communicator, String[] args) {
-        communicator.getProperties().setProperty("Ice.Package.Test", "test.IceBox.configuration");
```
InsertCreativityHere, line 13: This was the reason I discovered #3943.

Thankfully though, this test doesn't actually unmarshal any classes or exceptions.
So the 'correct fix' was to just delete this property being set. As I also did in many other tests.

java/test/src/main/java/test/Ice/packagemd/NoPackage.ice
```diff
@@ -2,7 +2,9 @@
```
InsertCreativityHere, line 2: The `packagemd` test is explicitly about testing `java:package` metadata.
It was left alone, except for adding some warning suppression, and making sure we test applying it to a module as well as at file level.

externl (approved):

java/test/src/main/java/test/Ice/ami/Server.java
```diff
@@ -3,28 +3,25 @@
...
 
 import test.TestHelper;
 
 public class Server extends TestHelper {
     public void run(String[] args) {
-        Properties properties = createTestProperties(args);
-        properties.setProperty("Ice.Package.Test", "test.Ice.ami");
-        //
+        var initData = new InitializationData();
+        initData.sliceLoader = new ModuleToPackageSliceLoader("::Test", "test.Ice.adapterDeactivation.Test");
```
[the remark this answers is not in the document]
InsertCreativityHere: A very good catch.

pepone (approved):

PR #3937, opened 2025-05-02
Add JavaScript FileLogger

### LINKED ISSUE #3591: Add test for LoggerMiddleware in JS
I tested it manually, but it would be nice to exercise this code in the test suite to make sure we don't introduce bugs.

pepone:
This PR also enables the logger middleware for Ice/operation to ensure the code is exercise. Fixes #3591

### DIFF
--- a/config/makeprops.py
+++ b/config/makeprops.py
@@ -421,7 +421,7 @@ final class PropertyNames
         "{name}",
         {prefixOnly},
         {isOptIn},
-        new Property[]{{{properties}
+        new Property[] {{{properties}
         }});
 
 """
--- a/js/src/Ice/InstanceExtensions.js
+++ b/js/src/Ice/InstanceExtensions.js
@@ -15,7 +15,7 @@ import { Timer } from "./Timer.js";
 import { TraceLevels } from "./TraceLevels.js";
 import { LocalException } from "./LocalException.js";
 import { CommunicatorDestroyedException, InitializationException } from "./LocalExceptions.js";
-import { getProcessLogger, setProcessLogger } from "./ProcessLogger.js";
+import { setProcessLogger } from "./ProcessLogger.js";
 import { ToStringMode } from "./ToStringMode.js";
 import { ProtocolInstance } from "./ProtocolInstance.js";
 import { TcpEndpointFactory } from "./TcpEndpointFactory.js";
--- a/js/src/Ice/InstanceExtensions.js
+++ b/js/src/Ice/InstanceExtensions.js
@@ -23,8 +23,7 @@ import { WSEndpointFactory } from "./WSEndpointFactory.js";
 import { Promise } from "./Promise.js";
 import { ConnectionOptions } from "./ConnectionOptions.js";
 import { StringUtil } from "./StringUtil.js";
-import { FileLogger  } from "./Logger.js";
-
+import { FileLogger } from "./Logger.js";
 
 import { Ice as Ice_Router } from "./Router.js";
 const { RouterPrx } = Ice_Router;
@@ -182,7 +181,6 @@ Instance.prototype.finishSetup = function (communicator) {
             this._initData.properties.getIcePropertyAsInt("Ice.Connection.Client.EnableIdleCheck") > 0,
             this._initData.properties.getIcePropertyAsInt("Ice.Connection.Client.InactivityTimeout"),
         );
-
 
         if (FileLogger !== null && this._initData.logger === null) {
             const logfile = this._initData.properties.getIceProperty("Ice.LogFile");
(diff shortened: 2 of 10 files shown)


scripts/Util.py
```diff
@@ -2346,21 +2349,20 @@ def start(self, process, current, args, props, envs, watchDog):
         }
 
         traceFile = ""
-        if not isinstance(process.getMapping(current), JavaScriptMixin):
```
bernardnormier, line 2349: Does this mean we're now running all JS tests with logging-to-file enabled?

js/test/Ice/operations/Client.ts
```diff
@@ -38,6 +38,8 @@ export class Client extends TestHelper {
         try {
             const [properties] = this.createTestProperties(args);
             properties.setProperty("Ice.BatchAutoFlushSize", "100");
+            properties.setProperty("Ice.Trace.Protocol", "1");
```
bernardnormier, line 41: Protocol=1 and Network=3 sounds excessive.

js/src/Ice/Logger.js
```diff
@@ -74,3 +76,27 @@ export class Logger {
...
+let FileLogger = null;
+
+if (typeof fs.open === "function") {
+    FileLogger = class extends Logger {
+        constructor(prefix, filename) {
+            super(prefix);
+            this._filename = filename;
+        }
+
+        write(message, indent) {
```
bernardnormier, line 89: It would be nice to have a plain `write(message)` like in C#, and leave the formatting to the base class.

js/src/Ice/InstanceExtensions.js
```diff
@@ -181,6 +182,12 @@ Instance.prototype.finishSetup = function (communicator) {
             this._initData.properties.getIcePropertyAsInt("Ice.Connection.Client.InactivityTimeout"),
         );
 
+        if (FileLogger !== null && this._initData.logger === null) {
+            const logfile = this._initData.properties.getIceProperty("Ice.LogFile");
+            if (logfile.length > 0) {
+                this._initData.logger = new FileLogger("", logfile);
```
bernardnormier, line 188: Why is the prefix always empty?

scripts/Util.py
pepone: No. You still have to enable it in the test. This code was here to ignore Ice.LogFile property with the JavaScript mapping, as it was previously not supported.

js/src/Ice/InstanceExtensions.js
```diff
@@ -181,6 +182,12 @@ Instance.prototype.finishSetup = function (communicator) {
             this._initData.properties.getIcePropertyAsInt("Ice.Connection.Client.InactivityTimeout"),
         );
 
+        if (FileLogger !== null && this._initData.logger === null) {
```
externl, line 185: Maybe we should read the property first and show an exception if we're in the browser and the user is trying to use the file logger.

js/src/Ice/InstanceExtensions.js
pepone: I guess we can use Ice.ProgramName as the default?

We have to fix process logger too:

https://github.com/zeroc-ice/ice/blob/0cca8df118385badf630932c306fc35e59814f85/js/src/Ice/ProcessLogger.js#L12

js/test/Ice/operations/Client.ts
pepone: This is to ensure that we exercise this code paths.

config/makeprops.py
```diff
@@ -421,7 +421,7 @@ def writePropertyArray(self, propertyArray):
         "{name}",
         {prefixOnly},
         {isOptIn},
-        new Property[]{{{properties}
+        new Property[] {{{properties}
```
pepone, line 424: Unrelated formatting fix for generated Java props.

js/src/Ice/InstanceExtensions.js
```diff
@@ -181,9 +183,19 @@ Instance.prototype.finishSetup = function (communicator) {
...
+            if (logFile.length > 0) {
+                if (FileLogger === null) {
+                    throw new InitializationException("Ice.LogFile property is not supported in Web Browsers");
+                }
+                this._initData.logger = new FileLogger(programName, logFile);
+            } else {
+                this._initData.logger = new Logger(programName);
+            }
         }
+        setProcessLogger(this._initData.logger);
```
bernardnormier, line 198: I don't understand this change. The process logger should not be the logger of the most recently created communicator.

pepone: I’d like to avoid lazy initialization in ProcessLogger, since it complicates setting the prefix for the default logger. What about:

```
if (getProcessLogger() === null) {
    setProcessLogger(this._initData.logger);
}
```

externl (approved):

js/src/Ice/InstanceExtensions.js
bernardnormier: The process logger and initData.logger are/should be totally independent.

bernardnormier (approved):

Issue #3936, opened 2025-05-02
Improve slice2cpp handling of invalid metadata


externl:
This slice file (note the bad metadata on `AtmosphericConditions`
```slice
// Copyright (c) ZeroC, Inc.

#pragma once

module ClearSky
{
    /// Represents the atmospheric conditions measured by a sensor.
    [cpp:custom-print]
    class AtmosphericConditions
    {
        /// The temperature in degrees Celsius.
        double temperature;

        /// The humidity in percent.
        double humidity;
    }

    /// A weather station collects readings from sensors.
    interface WeatherStation
    {
        /// Reports a new reading to the weather station(s) via the IceStorm service.
        /// @param sensorId The unique identifier of the sensor that took the reading.
        /// @param timeStamp The time the reading was taken.
        /// @param reading The atmospheric conditions measured by the sensor.
        void report(string sensorId, string timeStamp, AtmosphericConditions reading);
    }
}
```

causes this output

```
❯ slice2cpp WeatherStation.ice 
/Users/joe/Developer/zeroc-ice/ice-demos/cpp/IceStorm/weather/WeatherStation.ice:8: invalid character between metadata
/Users/joe/Developer/zeroc-ice/ice-demos/cpp/IceStorm/weather/WeatherStation.ice:8: invalid character between metadata
/Users/joe/Developer/zeroc-ice/ice-demos/cpp/IceStorm/weather/WeatherStation.ice:8: invalid character between metadata
/Users/joe/Developer/zeroc-ice/ice-demos/cpp/IceStorm/weather/WeatherStation.ice:8: invalid character between metadata
/Users/joe/Developer/zeroc-ice/ice-demos/cpp/IceStorm/weather/WeatherStation.ice:8: invalid character between metadata
/Users/joe/Developer/zeroc-ice/ice-demos/cpp/IceStorm/weather/WeatherStation.ice:8: invalid character between metadata
/Users/joe/Developer/zeroc-ice/ice-demos/cpp/IceStorm/weather/WeatherStation.ice:8: invalid character between metadata
/Users/joe/Developer/zeroc-ice/ice-demos/cpp/IceStorm/weather/WeatherStation.ice:8: invalid character between metadata
/Users/joe/Developer/zeroc-ice/ice-demos/cpp/IceStorm/weather/WeatherStation.ice:8: invalid character between metadata
[...]

PR #3939, opened 2025-05-03
Improve Reporting of Various Metadata Syntax Errors

InsertCreativityHere:
This PR improves reporting/recovery of syntax errors users may make in metadata, fixing #3936.

1) I added an "error recovery" point inside metadata. So any errors that occur between `[` and `]` or `[[` and `]]` won't escape the brackets, and will be recovered from upon hitting the closing bracket.

2) If the Slice scanner hits a 'bad character' in metadata, it will report an error for _every single character_. This is overkill. Now the scanner will report a single error per longest contiguous run of bad chars. I also renamed the `BAD_CHAR` token to `BAD_TOKEN` to signal that it can contain multiple chars now.

3) Bullet-point-1 is a fallback to catch and contain _any_ unexpected syntax errors. But we have even better handling of bad chars (bullet-point-2), allowing us to recover at the next `,` character.

4) Now, the Slice parser tolerates 'empty' metadata, and warns about it. Note that this can occur in 2 ways: either literally typing an empty `[]` (bad user!) or they forgot to wrap it metadata quotes, causing it to be thrown away as `BAD_TOKEN`.


pepone (approved):
Looks good, do we need any special handling for the case that a user forgets to close the metadata?

externl (approved):

InsertCreativityHere:
> Looks good, do we need any special handling for the case that a user forgets to close the metadata?

In this case, everything is kind of screwed anyways. Trying to handle unmatched brackets, braces, parenthesis, etc. is a very tricky thing to do well; right now we don't have any logic to handle such cases.

PR #3945, opened 2025-05-07
Update `Forward.ice` Test Section in `cpp/Slice/errorDetection`.

### LINKED ISSUE #3252: Cleanup `cpp/test/Slice/errorDetection` Testing
I was checking the C++ Code Coverage report, and to my surprise, parts of `slice2java` were being hit! Same for the other compilers.

After some searching, it turns out that the `cpp/Slice/errorDetection` test actually runs _all_ the compilers.
We have this little bit slapped on the end of our test script in that folder:
```
            current.write("Forward.ice... ")
            for language in [
                "cpp",
                "cs",
                "html",
                "java",
                "js",
                "matlab",
                "php",
                "py",
                "rb",
                "swift",
            ]:
                compiler = SliceTranslator("slice2%s" % language)
                if not os.path.isfile(compiler.getCommandLine(current)):
                    continue
                compiler.run(
                    current, args=["forward/Forward.ice", "--output-dir", "tmp"]
                )
```

1) We should remove `html` from this list. `slice2html` is long dead. We should also remove this logic where we just silently skip any compilers that don't exist. If something in this list is bogus, we should obviously see an error for it.

2) We're only testing the compilers over `Forward.ice`, a _very_ small file: `module test { class F; }`.
Either we should just completely remove this testing (because it is unexpected that a `cpp` test is running the other compilers).
Or we should at least expand this testing and run the compilers on a less useless Slice file.

InsertCreativityHere:
Fixes #3252.

At the bottom of the `Slice/errorDetection` test, we run each of the Slice compilers over `Forward.ice` to make sure they work without errors. But the test file is very small just a single class forward declaration in a module. Also, the list of compilers is also out of date, including `slice2html` (the test silently skips non-existent compilers (not great)).

This PR updates the list of compilers (removes `slice2html` and adds `ice2slice`), will now fail if a compiler doesn't exist, and instead of running them over this small Slice file, runs them against all `*.ice` files in the `./slice` directory.
On my local machine, this added 29 seconds to testing. So about 3s per compiler.

----

While grepping for `ice2slice`, I learned that we _do_ actually have an action (`Run ice2slice`) testing it (added back in January).
But there's something fishy. `iceslice` was broken for a few months until #3827 in April, and yet the action never failed...

Regardless, properly testing `ice2slice` was one of the main reasons behind this PR.
Now that I know the action exists, we should either:
1) close this PR, delete this part of `errorDetection`, and fix the action (possibly to run on more platforms/configurations)
2) or we should delete the action in favor of this PR.

I honestly don't care one way or another. As long as `ice2slice` is being tested, it's fine with me! So I leave it up to you!

### DIFF
--- a/cpp/test/Slice/errorDetection/forward/Forward.ice
+++ b/cpp/test/Slice/errorDetection/forward/Forward.ice
@@ -1,4 +1,0 @@
-module test
-{
-    class F;
-}
--- a/cpp/test/Slice/errorDetection/test.py
+++ b/cpp/test/Slice/errorDetection/test.py
@@ -94,7 +94,7 @@ class SliceErrorDetectionTestCase(ClientTestCase):
             for file in files:
                 try:
                     compiler.run(current, args=[file, "--output-dir", "tmp", "-I" + sliceDir])
-                except:
+                except RuntimeError:
                     failures.append(file)
 
             current.writeln("failed!" if failures else "ok")
--- a/cpp/test/Slice/errorDetection/test.py
+++ b/cpp/test/Slice/errorDetection/test.py
@@ -11,78 +11,105 @@ from Util import ClientTestCase, SliceTranslator, TestSuite
 class SliceErrorDetectionTestCase(ClientTestCase):
     def runClientSide(self, current):
         testdir = current.testsuite.getPath()
-        slice2cpp = SliceTranslator("slice2cpp", quiet=True)
 
+        # Create a temporary directory for all the Slice compilers to write their output to.
         outdir = "{0}/tmp".format(testdir)
         if os.path.exists(outdir):
             shutil.rmtree(outdir)
         os.mkdir(outdir)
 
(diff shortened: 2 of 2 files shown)


externl:
Not sure how I feel about adding a 30s test

pepone:
I agree with @extern, about this is redundant with Ice builds, for ice2slice we have also a separate CI job that does the same.

I think this Forward.ice was a not very good test related to forward declarations, we need to test forward declarations with the language mappings.

InsertCreativityHere:
Okay, I'll close this and open a PR that just deletes this `Forward.ice` test then.
We're already testing forward declarations in the language mappings, so now sure what the point of this was.

But, then we need to update the `ice2slice` workflow to run on more than just `ubuntu-24.04`.
Because `ice2slice` was broken for a few months (at least on windows) until #3827,
and yet the action was still green, failing to alert us to the problems in `ice2slice`.

InsertCreativityHere:
Closed in favor of #3955.

PR #3947, opened 2025-05-07
Remove view-type from dictionary, improve Ice/span test

bernardnormier:
This PR removes the cpp:view-type metadata from dictionary because:
 - we don't provide any type (such as span) to use it with
   - it was only useful in the event to added your own view-type handling with a StreamHelper
 - we didn't have any test for dictionary view-types

This PR also improves the existing Ice/span test for cpp:view-type on sequences to map StringSeq to `std::span<const std::string_view>` (a true view type) and not `std::span<std::string>` (which worked but was lame).

### DIFF
--- a/cpp/test/Ice/span/Test.ice
+++ b/cpp/test/Ice/span/Test.ice
@@ -12,31 +12,31 @@ module Test
 
     interface TestIntf
     {
-        ["cpp:view-type:std::span<const std::byte>"] ByteSeq opByteSpan(
+        ByteSeq opByteSpan(
             ["cpp:view-type:std::span<const std::byte>"] ByteSeq dataIn,
-            ["cpp:view-type:std::span<const std::byte>"] out ByteSeq dataOut);
+            out ByteSeq dataOut);
 
-        ["cpp:view-type:std::span<const std::int16_t>"] ShortSeq opShortSpan(
+        ShortSeq opShortSpan(
             ["cpp:view-type:std::span<const std::int16_t>"] ShortSeq dataIn,
-            ["cpp:view-type:std::span<const std::int16_t>"] out ShortSeq dataOut);
+            out ShortSeq dataOut);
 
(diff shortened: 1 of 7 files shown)


externl (approved):

cpp/test/Ice/span/TestAMD.ice
```diff
@@ -10,34 +10,36 @@ module Test
     sequence<short> ShortSeq;
     sequence<string> StringSeq;
 
+    // No metadata on in-parameters are we don't use this generated code for the client.
```
pepone, line 13:
```suggestion
    // No metadata on in-parameters and we don't use this generated code for the client.
```

pepone (approved):

cpp/test/Ice/span/Test.ice
```diff
@@ -12,31 +12,31 @@ module Test
 
     interface TestIntf
```
InsertCreativityHere, line 13: Why remove the metadata from the `out` parameters in this file, but not `TestAMD.ice`?
This file is used by both the client side and server side (non-amd).

cpp/test/Ice/span/TestAMD.ice
InsertCreativityHere, line 13:
```suggestion
    // No metadata on in-parameters as we don't use this generated code for the client.
```
This sentence needs a tweak.

But I also think we should just keep this metadata here.
There is value to testing that applying this metadata to in-parameters has no effect on the server side.

And we usually keep `Test.ice` and `TestAMD.ice` in close sync with one another (except for `["amd"]` of course).
introducing inconsistencies will make them harder to maintain IMO.

InsertCreativityHere (approved):
Fine with me.
I doubt anyone was `view-type`ing their dictionaries anyways.

cpp/test/Ice/span/Test.ice
bernardnormier: view-type has no effect on out parameters, except with AMD.

PR #3951, opened 2025-05-09
Remove Ice.Override.Secure from JS

bernardnormier:
Small Property cleanup.


bernardnormier:
I think it's incorrect: JS should use this property.

PR #3953, opened 2025-05-12
Handle cs:identifier in ice2slice

### LINKED ISSUE #3832: Update `ice2slice` to map `cs:identifier` Metadata
`ice2slice` handles `cs:namespace`, so we should probably update it to also map `cs:identifier` as well.

pepone:
Fix #3832

### DIFF
--- a/cpp/src/ice2slice/Gen.cpp
+++ b/cpp/src/ice2slice/Gen.cpp
@@ -43,7 +43,7 @@ namespace
     static string getCsMappedScope(const ContainedPtr& cont, const string& separator)
     {
         string scoped;
-        auto container = std::dynamic_pointer_cast<Contained>(cont->container());
+        auto container = dynamic_pointer_cast<Contained>(cont->container());
         if (container)
         {
             scoped = getCsMappedScoped(container, separator);
--- a/cpp/src/ice2slice/Gen.cpp
+++ b/cpp/src/ice2slice/Gen.cpp
@@ -27,35 +27,25 @@ namespace
     // We cannot use mappedName because it depends on the language name used to create the Unit.
     static string getCsMappedName(const ContainedPtr& cont)
     {
-        if (auto metadata = cont->getMetadataArgs("cs:identifier"))
-        {
-            return *metadata;
-        }
-        else
-        {
-            return cont->name();
-        }
+        return cont->getMetadataArgs("cs:identifier").value_or(cont->name());
     }
 
-    static string getCsMappedScoped(const ContainedPtr& cont, const string& separator);
+    static string getCsMappedScoped(const ContainedPtr& cont);
 
     // We cannot use mappedScope because it depends on the language name used to create the Unit.
-    static string getCsMappedScope(const ContainedPtr& cont, const string& separator)
+    static string getCsMappedScope(const ContainedPtr& cont)
     {
         string scoped;
         auto container = dynamic_pointer_cast<Contained>(cont->container());
         if (container)
         {
-            scoped = getCsMappedScoped(container, separator);
+            scoped = getCsMappedScoped(container);
         }
-        return scoped + separator;
+        return scoped + '.';
     }
 
     // We cannot use mappedScoped because it depends on the language name used to create the Unit.
-    static string getCsMappedScoped(const ContainedPtr& cont, const string& separator)
-    {
-        return getCsMappedScope(cont, separator) + getCsMappedName(cont);
-    }
+    static string getCsMappedScoped(const ContainedPtr& cont) { return getCsMappedScope(cont) + getCsMappedName(cont); }
 
     static string getCsNamespacePrefix(const ContainedPtr& cont)
     {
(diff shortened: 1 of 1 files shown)


bernardnormier (commented):
I find the terminology and logic very confusing. It would be better to refactor both.

This `getCSharpNamespace(..., bool& hasCSharpNamespaceAttribute)` is called twice:

1) apparently to map fully-qualified references, where  hasCSharpNamespaceAttribute is not used. It should also have nothing to do with namespaces.

2) to check if we need to generated a `cs:namespace` is the .slice file. Here, `hasCSharpNamespaceAttribute` is definitely not the correct term. It suggests the .ice file has a "csharp namespace attribute", which may or may not be the case. Something like `hasModuleRemapping` would be much better.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -17,18 +17,24 @@ namespace
...
                 else
                 {
-                    csharpNamespace = p->name() + "." + csharpNamespace;
+                    csharpNamespace = p->mappedName() + "." + csharpNamespace;
+                }
+
+                if (auto metadata = p->getMetadataArgs("cs:identifier"))
+                {
+                    hasCSharpNamespaceAttribute = true;
+                    return *metadata;
```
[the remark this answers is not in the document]
pepone: Yes, this was bogus, reworked.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -12,26 +12,54 @@ using namespace IceInternal;
...
+        {
+            return *metadata;
+        }
+        else
+        {
+            return cont->name();
+        }
+    }
+
+    static string getCSharpMappedScoped(const ContainedPtr& cont, const string& separator);
```
pepone, line 39: We cannot use mappedName, mappedScope, .. because they depend on the language-mapping set when you create the Unit.

bernardnormier: Would be nice to add this comment in the code.

Also, what about calling these function getCsXxx, etc. Having both cs and csharp looks odd.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -43,13 +71,30 @@ namespace
...
+            return "";
+        }
+    }
+
+    static string getCSharpNamespace(const ContainedPtr& cont, bool& hasModuleRemapping)
+    {
+        // Module remapping is required if the module has a cs:namespace attribute or any of the modules has
+        // a cs:identifier
+        string csNamespacePrefix = getCSharpNamespacePrefix(cont);
+        hasModuleRemapping = !csNamespacePrefix.empty() || cont->scope() != getCSharpMappedScope(cont, "::");
```
bernardnormier, line 87: `|| cont->scope() != getCSharpMappedScope(cont, "::");`

I find the logic too round-about. We remap the module when the Slice scope of cont is different from the mapped scope of cont using `::` separators?

What about a simpler comparison: take the default mapping and compare it with the actual mapping?
And to do even simpler, this comparison could be performed by the unique caller: we just need one function that returns the actual mapping and another that returns the default mapping. Then no hasModuleRemapping, no separator parameter ... so much simpler.

cpp/src/ice2slice/Gen.cpp
InsertCreativityHere: Or just call it `getMappedScoped` and have it take a language prefix parameter.
In the future, when we add Python support for icerpc, we'd want to re-use these for python metadata as well.

pepone: We can rename them when we add support for new mappings, all this code is internal to ice2slice.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -12,26 +12,57 @@ using namespace IceInternal;
...
+        {
+            return *metadata;
+        }
+        else
+        {
+            return cont->name();
+        }
+    }
+
+    static string getCsMappedScoped(const ContainedPtr& cont, const string& separator);
```
InsertCreativityHere, line 40: `separator` will always be `"."` in the context that these functions are being called.
I'd remove this extra parameter and just hardcode `"."` in `getCsMappedScope`.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -153,10 +197,13 @@ namespace
...
-            string csharpNamespace = getCSharpNamespace(contained, hasCSharpNamespaceAttribute);
-
-            os << csharpNamespace << "." << contained->name();
+            string csNamespacePrefix = getCsNamespacePrefix(contained);
+            string scoped = getCsMappedScoped(contained, ".").substr(1);
+            if (!csNamespacePrefix.empty())
+            {
+                os << csNamespacePrefix << ".";
+            }
+            os << scoped;
```
InsertCreativityHere, line 206: I think these should be equivalent:
```suggestion
            string csNamespace = getCsMappedNamespace(contained);
            os << csNamespace << "." << getCsMappedName();
```

cpp/src/ice2slice/Gen.cpp
```diff
@@ -43,13 +74,26 @@ namespace
...
         if (auto metadata = p->getMetadataArgs("cs:namespace"))
         {
-            hasCSharpNamespaceAttribute = true;
-            return *metadata + "." + csharpNamespace;
+            return *metadata;
+        }
+        else
+        {
+            return "";
+        }
```
InsertCreativityHere, line 82: There's a nice convenience function for this:
```
return getMetadataArgs("cs:namespace").value_or("");
```
Unfortunately, GitHub won't let me suggest it, since the `if` wasn't changed in this PR...

InsertCreativityHere (approved):
Looks good!
Just some minor simplification suggestions.

bernardnormier (approved):
Looks good.

PR #3954, opened 2025-05-12
Remove --nowarn option

### LINKED ISSUE #3635: Review usage of --nowarn command line option in icegrid
We should review the usage of the `--nowarn` command line option in `icegridnode` and `icegridregistry` and possibly remove it. 

- In my option its a bit to broad and mostly unhelpful.

- Its description is not quite correct: "Don't print any security warnings." We do more than this.

- https://github.com/zeroc-ice/ice/blob/8a7cb7f262c5c463e019d19ac50d013a9eedc9ab/cpp/src/IceGrid/IceGridNode.cpp#L265-L270
  This should just be an error.

- https://github.com/zeroc-ice/ice/blob/8a7cb7f262c5c463e019d19ac50d013a9eedc9ab/cpp/src/IceGrid/IceGridNode.cpp#L355-L359
  Seems fine to just print the message.

- https://github.com/zeroc-ice/ice/blob/8a7cb7f262c5c463e019d19ac50d013a9eedc9ab/cpp/src/IceGrid/IceGridNode.cpp#L458-L471
  Not sure what the idea is with this one. I would just always print the a message on startup in the registry is unreachable. 

- https://github.com/zeroc-ice/ice/blob/8a7cb7f262c5c463e019d19ac50d013a9eedc9ab/cpp/src/IceGrid/RegistryI.cpp#L207-L213
   Are these the security warnings? The message reads more like information than a warning and a warning makes it sound like something you should never do. 

- etc.

pepone:
Fix #3635

### DIFF
--- a/cpp/src/Glacier2/Glacier2Router.cpp
+++ b/cpp/src/Glacier2/Glacier2Router.cpp
@@ -443,13 +443,12 @@ RouterService::usage(const string& appName)
                      "-h, --help           Show this message.\n"
                      "-v, --version        Display the Ice version.\n";
 #ifndef _WIN32
-    options.append(
-        "\n"
-        "\n"
-        "--daemon             Run as a daemon.\n"
-        "--pidfile FILE       Write process ID into FILE.\n"
-        "--noclose            Do not close open file descriptors.\n"
-        "--nochdir            Do not change the current working directory.\n");
+    options.append("\n"
+                   "\n"
+                   "--daemon             Run as a daemon.\n"
+                   "--pidfile FILE       Write process ID into FILE.\n"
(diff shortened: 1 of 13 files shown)


bernardnormier (approved):
You need to fix the man pages:

```
man1/icegridnode.1
56:.BR \-\-nowarn\fR
58:Do not print security warnings.

man1/glacier2router.1
30:.BR \-\-nowarn\fR
32:Suppress warnings.

man1/icegridregistry.1
54:.BR \-\-nowarn\fR
56:Do not print security warnings.

man1/icebox.1
29:.BR \-\-nowarn\fR
31:Suppress warnings.
```

Issue #3853, opened 2025-04-14
Move 'Comment-Parsing' Into a Discrete Step of `libSlice`'s Parsing Pipeline


InsertCreativityHere:
Back in the day, each compiler had it's own comment-parsing logic, which was run as-needed by that compiler.
#3030 Centralized all the comment-parsing into `libSlice`, eliminating all the duplicated code.
But there is still room for improvement, since all the comment-parsing is done "on-demand".

Meaning that comment parsing happens in the middle of code-gen, whenever the code-gen wants to generate the mapped comment. I think it'd be better to do all the comment-parsing up-front, and _before_ the code-gen phase.

### Problems with the current approach:
1) Some doc-comment warnings will be emitted multiple times.
For example, `slice2cpp` has separate visitors for generating proxies vs generating servants.
Each of these visitors will separately parse interface/operation doc-comments,
Not only is this inefficient, but it will cause warnings to be emitted twice. Since the comment-parser runs twice.

2) Doc-comments are validated in code-gen order, not definition order. For example `slice2cpp` generates code for interfaces before it generates code for exceptions. This means that the line-numbers for doc-comment diagnostics will appear in a 'random' order, depending on what order you defined things in.

3) It's just cleaner for doc-comment parsing/validation to happen before code-gen. Just like we do for metadata.

----

It's also worth noting that doc-comments in general are getting an overhaul.
We will eventually need to add a way for each language to specify how to format `` `foo` `` and `@p`.
We already have a mechanism in place for formatting `@link`, but it could probably be improved.

PR #3958, opened 2025-05-12
Bundle Doc-Comment Mapping Information into a New `DocCommentFormatter` Class

InsertCreativityHere:
Different languages have different syntaxes for doc-links. So, right now, when the Slice compiler parses a DocComment, it takes a `DocLinkFormatter` (an alias for a function pointer with a specific signature).

This works, but passing a function pointer isn't super elegant, especially since we're soon going to need formatters for `@p`, code-fences, and possibly more. Which would mean having 3 free functions floating around, and passing 3 pointers to the parser.

In anticipation of this, this PR introduces a new `DocCommentFormatter` class, which contains virtual functions that each compiler can override. So now, instead of passing bools and function pointers into `DocComment::parseFrom`, each compiler provides it's own implementation of `DocCommentFormatter` and passes this in instead.

----

In a follow up PR, each language compiler will provide a single instance of this class to `libSlice` (DI style).
And `libSlice` will handle parsing all the comments for us. Whereas right now, each compiler parses comments "on-demand". This means some comments are parsed multiple times or not at all. And that this parsing happens in the middle of code-gen.
But this requires some more re-shuffling of code. See #3853.


cpp/src/Slice/Parser.h
```diff
@@ -244,36 +244,43 @@ namespace Slice
...
-    /// The first argument (`rawLink`) is the raw link text, taken verbatim from the doc-comment.
-    /// The second argument (`source`) is a pointer to the Slice element that the doc comment (and link) are written on.
-    /// The third argument (`target`) is a pointer to the Slice element that is being linked to.
-    /// If the parser could not resolve the link, this will be `nullptr`.
-    ///
-    /// This function should return the fully formatted link.
-    /// `DocComment::parseFrom` replaces the entire '{@link <rawLink>}' by the string this function returns.
-    using DocLinkFormatter =
-        std::string (*)(const std::string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target);
+    class DocCommentFormatter
```
pepone, line 247: What about renaming the class to `DocCommentParserOptions`.

pepone (approved):

cpp/src/Slice/Parser.h
externl, line 247: Can we just use a struct that has two bools and a function for fields. I don't really see a benefit for using inheritance here.

InsertCreativityHere: Just a matter of preference I guess, I don't see a functional difference in these approaches.
If everyone prefers we keep using the current function-pointer approach, and just wrap things into an OptionBag struct,
that's fine with me too. Still an improvement over the current approach IMO.

But I did consider doing this, and ended up choosing the interface-like structure because:
- pure virtual functions that subclasses implement just felt cleaner than passing pointers to functions in the anonymous namespace.
- it's easier to document functions than it is to document function pointers
- all the relevant code is better organized; Just search for `DocCommentFormatter` and bam, everything is in the classes,
  instead of being free-floating functions somewhere in the anonymous namespace.

InsertCreativityHere: Sure, I'm not opposed to changing the name.
Going to wait for other reviewers though, to avoid changing it multiple times.

bernardnormier: Assuming the shouldXxx options are sensible and that we want to keep them, I agree with Joe: it should be a concrete options bag class, named `DocCommentParserOptions`.

InsertCreativityHere: I'm really more concerned with the `linkFormatter` (and future `codeFormatter`, etc.) and whether they should be function pointers stored in a concrete struct or virtual functions implemented by classes.

Because I'm not sure we should keep `shouldStripMarkup`, and `shouldUseXmlEscaping` is only used by C#.
We removed all the `<tag>`s from our doc-comments, and we don't recommend them to users, since we don't map them. So it's kind of weird to keep functionality for removing them (since they shouldn't be present anyways).

Perhaps we should talk about these bools before I do this refactoring.

PR #3960, opened 2025-05-13
Add `getTopLevelModule` Method to `Contained`

InsertCreativityHere:
This PR adds a convenience function to `libSlice`: `getTopLevelModule`.
It does what it says on the tin. You invoke it on a Slice element, and it returns the top-level module which contains that element.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -972,13 +972,14 @@ Slice::Contained::getTopLevelModule() const
     while (!p->isTopLevel())
     {
         p = dynamic_pointer_cast<Contained>(p->container());
-        assert(p);
     }
 
     // By the time we reach here, 'p' must be a top-level element; we cast it to a module.
     // Note that this cast can fail for elements erroneously defined outside of a module!
     // However, these are guaranteed not to exist after the parsing stage has completed.
-    return dynamic_pointer_cast<Module>(p);
+    ModulePtr topLevelModule = dynamic_pointer_cast<Module>(p);
+    assert(topLevelModule);
+    return topLevelModule;
 }
 
 string
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -953,6 +953,34 @@ Slice::Contained::isTopLevel() const
     return dynamic_pointer_cast<Unit>(container()) != nullptr;
 }
 
+ModulePtr
+Slice::Contained::getTopLevelModule() const
+{
+    ContainedPtr p;
+
+    // Round-about way to get `ContainedPtr` for this element, since we can't `enable_shared_from_this`.
+    for (const auto& sibling : container()->contents())
+    {
+        if (sibling.get() == this)
+        {
+            p = sibling;
+        }
+    }
+    assert(p);
+
+    // Navigate up through this element's parents until we hit a top-level element.
+    while (!p->isTopLevel())
+    {
+        p = dynamic_pointer_cast<Contained>(p->container());
+        assert(p);
+    }
+
+    // By the time we reach here, 'p' must be a top-level element; we cast it to a module.
+    // Note that this cast can fail for elements erroneously defined outside of a module!
+    // However, these are guaranteed not to exist after the parsing stage has completed.
+    return dynamic_pointer_cast<Module>(p);
+}
+
 string
 Slice::Contained::name() const
 {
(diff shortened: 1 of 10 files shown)


cpp/src/Slice/Parser.cpp
```diff
@@ -953,6 +953,35 @@ Slice::Contained::isTopLevel() const
...
+
+    // Round-about way to get `ContainedPtr` for this element, since we can't `enable_shared_from_this`.
+    for (const auto& sibling : container()->contents())
+    {
+        if (sibling.get() == this)
+        {
+            p = sibling;
+        }
+    }
+    assert(p);
```
InsertCreativityHere, line 969: Is there a better way to obtain a `ContainedPtr` of `this` here?

`Contained` doesn't `enable_shared_from_this` and cannot, so that it's subclasses can support `shared_from_this` with more specific typing.

pepone: Did you consider defining this as a separate function `Slice::getTopLevelModule(ContainedPtr) const`?

InsertCreativityHere: Yes, and I am still considering it! If no one has a better idea, I'll probably switch a free function.
But I figured I'd ask before falling back to it, since we tend to prefer member functions over free functions.

bernardnormier: I don't find this convoluted code acceptable.

If you want to use a member function, make it virtual and override it at least in Module. Then use recursion.

bernardnormier (changes requested):

cpp/src/Slice/Parser.cpp
InsertCreativityHere: I also don't find it acceptable, I was just showing what I wanted in hopes there was a 5* C++ way to accomplish it.
I'll simplify and switch to a free function, as suggested.

bernardnormier: Virtual member function and recursion seems pretty straightforward to me. Am I missing something?

InsertCreativityHere: No, you're not.

Just that it would of been even more straightforward if I could get a `ContainedPtr` of this.
Then, I'd only need one implementation on `Contained`, instead of 2 (one for `Contained` and one for `Module`).

Either way, I changed the implementation as suggested.

cpp/src/Slice/Parser.cpp
```diff
@@ -2422,6 +2435,17 @@ Slice::Module::kindOf() const
     return "module";
 }
 
+ModulePtr
+Slice::Module::getTopLevelModule() const
+{
+    if (auto parent = dynamic_pointer_cast<Contained>(container()))
+    {
+        parent->getTopLevelModule();
```
bernardnormier, line 2443: missing return

cpp/src/Slice/Parser.cpp
```diff
@@ -953,6 +953,19 @@ Slice::Contained::isTopLevel() const
     return dynamic_pointer_cast<Unit>(container()) != nullptr;
 }
 
+ModulePtr
+Slice::Contained::getTopLevelModule() const
+{
+    if (auto parent = dynamic_pointer_cast<Contained>(container()))
+    {
+        parent->getTopLevelModule();
```
bernardnormier, line 961: missing return

bernardnormier (approved):

cpp/src/Slice/Parser.cpp
```diff
@@ -953,6 +953,19 @@ Slice::Contained::isTopLevel() const
...
+Slice::Contained::getTopLevelModule() const
+{
+    if (auto parent = dynamic_pointer_cast<Contained>(container()))
+    {
+        return parent->getTopLevelModule();
+    }
+    // `Module` has its own implementation of this function. So reaching here means we hit an
+    // element which is a top-level non-module. The parser will report an error for it, but
+    // until we exit (at the end of parsing) this element will exist, and needs to be handled.
+    return nullptr;
```
pepone, line 966: Can this really happen?

InsertCreativityHere: Yeah for sure. You could compile this Slice file:
```
#pragma once
class MyClass {} // a top-level class!
```
Of course, the compiler will report an error `"classes can only be defined within a module"` during the parsing phase, and we won't generate any code (since an error occurred), but we still happily continue parsing until we reach the end of the file.

Only the code-gen (and some validation) can assume that this is non-null. Because the error will of prevented us from hitting those phases otherwise. But the scanning and parsing phases can make no such assumption.

pepone (approved):

PR #3962, opened 2025-05-13
Add Support for new `@remarks` Doc-Comment Tag

InsertCreativityHere:
This PR adds parser support for `@remark` (or `@remarks`) blocks in Slice doc-comments, and code-gen support for `slice2cpp` (where it's mapped to `@remarks`) and `slice2cs` where it's mapped to `<remarks>`.

Once this PR is merged, we'll need to add support to all the other languages which generate doc-comments as well.
Although the mapping for this tag may be less straight-forward in some of them.

----

So now in C++, we generate this additional text:
```diff
     /// A sequence of metrics.
+    /// @remarks We use a sequence here instead of a map because the ID of the metrics is already included in the Metrics
+    /// class and using sequences of metrics objects is more efficient than using dictionaries since lookups are not
+    /// necessary.
     using MetricsMap = std::vector<MetricsPtr>;
```
```diff
         /// Logs a LogMessage.
         /// @param message The message to log.
         /// @param context The request context.
         /// @return A future that becomes available when the invocation completes.
+        /// @remarks ::Ice::RemoteLoggerPrx::log may be called by ::Ice::LoggerAdminPrx before ::Ice::RemoteLoggerPrx::init.
         [[nodiscard]] std::future<void> logAsync(const LogMessage& message, const Ice::Context& context = Ice::noExplicitContext) const;
```
```diff
     /// Looks for a locator using UDP multicast.
+    /// @remarks This interface is implemented by Ice locator implementations and can be used by clients to find
+    /// available Ice locators on the network.
+    /// Ice locator implementations provide a well-known 'Ice/LocatorLookup' object accessible through UDP multicast.
+    /// Clients typically make a multicast findLocator request to find the locator proxy.
     /// @see LookupReply
     class LookupPrx : public Ice::Proxy<LookupPrx, Ice::ObjectPrx>
```
etc... you get the idea.

This is what it looks like in C#. Note that for most generated types, we already generated a `remarks` section.
In cases where there's a user-provided remarks, and one about the generated type, we concatenate them with a `<para>` tag.
```diff
     /// <summary>Represents a version of the Ice encoding. Ice supports version 1.0 and 1.1 of this encoding.</summary>
-    /// <remarks>The Slice compiler generated this record struct from Slice struct <c>::Ice::EncodingVersion</c>.</remarks>
+    /// <remarks>The Ice encoding is also known as the Slice encoding.
+    /// <para />
+    /// The Slice compiler generated this record struct from Slice struct <c>::Ice::EncodingVersion</c>.</remarks>
     public partial record struct EncodingVersion
```
```diff
        /// <summary>Stops a service.</summary>
        /// <param name="service">The service name.</param>
        /// <param name="context">The request context.</param>
        /// <exception cref="IceBox.AlreadyStoppedException">Thrown when the service is already stopped.</exception>
        /// <exception cref="IceBox.NoSuchServiceException">Thrown when IceBox does not know a service named @p service.</exception>
+       /// <remarks>This is a fake remark just to show how it looks for operations.</remarks>
        void stopService(string service, global::System.Collections.Generic.Dictionary<string, string>? context = null);
```

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -553,7 +553,7 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
                 else
                 {
                     comment._parameters[name] = {};
-                    currentSection = &(comment._parameters[name]);
+                    currentSection = &comment._parameters[name];
                 }
             }
         }
@@ -603,18 +603,18 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
                     else
                     {
                         comment._exceptions[name] = {};
-                        currentSection = &(comment._exceptions[name]);
+                        currentSection = &comment._exceptions[name];
                     }
                 }
             }
         }
         else if (parseCommentLine(line, remarkTag, lineText) || parseCommentLine(line, remarksTag, lineText))
         {
-            currentSection = &(comment._remarks);
+            currentSection = &comment._remarks;
         }
         else if (parseCommentLine(line, seeTag, lineText))
         {
-            currentSection = &(comment._seeAlso);
+            currentSection = &comment._seeAlso;
 
             // Remove any leading and trailing whitespace from the line.
             // There's no concern of losing formatting for `@see` due to its simplicity.
@@ -660,7 +660,7 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
                 }
                 else
                 {
-                    currentSection = &(comment._returns);
+                    currentSection = &comment._returns;
                 }
             }
         }
@@ -676,7 +676,7 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
             else
             {
                 comment._isDeprecated = true;
-                currentSection = &(comment._deprecated);
+                currentSection = &comment._deprecated;
             }
         }
         else // This line didn't introduce a new tag. Either we're in the overview or a tag whose content is multi-line.
@@ -693,7 +693,7 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
                 }
 
                 // '@see' tags are not allowed to span multiple lines.
-                if (currentSection == &(comment._seeAlso))
+                if (currentSection == &comment._seeAlso)
                 {
                     string msg = "'@see' tags cannot span multiple lines and must be of the form: '@see identifier'";
                     p->unit()->warning(p->file(), p->line(), InvalidComment, msg);
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -507,6 +507,7 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
     const string throwsTag = "@throws";
     const string exceptionTag = "@exception";
     const string remarkTag = "@remark";
+    const string remarksTag = "@remarks";
     const string seeTag = "@see";
     const string returnTag = "@return";
     const string deprecatedTag = "@deprecated";
@@ -607,7 +608,7 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
                 }
             }
         }
-        else if (parseCommentLine(line, remarkTag, lineText))
+        else if (parseCommentLine(line, remarkTag, lineText) || parseCommentLine(line, remarksTag, lineText))
         {
             currentSection = &(comment._remarks);
         }
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -506,6 +506,7 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
     const string paramTag = "@param";
     const string throwsTag = "@throws";
     const string exceptionTag = "@exception";
+    const string remarksTag = "@remarks";
     const string seeTag = "@see";
     const string returnTag = "@return";
     const string deprecatedTag = "@deprecated";
@@ -605,6 +606,10 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
                     }
                 }
             }
+        }
+        else if (parseCommentLine(line, remarksTag, lineText))
+        {
+            currentSection = &(comment._remarks);
         }
         else if (parseCommentLine(line, seeTag, lineText))
         {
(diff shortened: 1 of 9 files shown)


cpp/src/Slice/Parser.h
```diff
@@ -280,6 +280,8 @@ namespace Slice
 
         /// Contains all introductory lines up to the first tag.
         [[nodiscard]] StringList overview() const;
+        /// Returns the contents of any '@remarks' tags.
```
externl, line 283:
```suggestion
        /// Returns the contents of any '@remarks' or '@remark' tags.
```

externl (approved):

cpp/src/Slice/Parser.h
InsertCreativityHere: Fixed, but I swapped the order.
If we list both, I think `remark` should come before `remarks`.

cpp/src/Slice/Parser.cpp
```diff
@@ -606,6 +608,10 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
                 }
             }
         }
+        else if (parseCommentLine(line, remarkTag, lineText) || parseCommentLine(line, remarksTag, lineText))
+        {
+            currentSection = &(comment._remarks);
```
bernardnormier, line 613: Since we have earlier:
```cpp
StringList* currentSection = &comment._overview;
```

it's likely all the parenthesis are superfluous.

cpp/src/Slice/Parser.h
```diff
@@ -280,6 +280,8 @@ namespace Slice
 
         /// Contains all introductory lines up to the first tag.
         [[nodiscard]] StringList overview() const;
+        /// Returns the contents of any '@remark' or '@remarks' tags.
```
bernardnormier, line 283: I understand the performance of this code does not matter, however this API looks gratuitously bad.

I have a bunch of `std::list<string>` fields, and I provide accessors for them. What do they look like?

1) I return copies each time: 
```cpp
[[nodiscard]] StringList remarks() const;
```

2) I return a const& to the field
```cpp
[[nodiscard]] const StringList& remarks() const noexcept { return _remarks; }
```

Why pick (1)?

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -388,6 +388,13 @@ namespace
             writeDocLines(out, doc->overview(), true);
         }
 
+        const auto& remarks = doc->remarks();
```
bernardnormier, line 391: I don't understand how this code compiles without warning. You're taking a const reference to a temporary object.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -558,6 +565,13 @@ namespace
             writeOpDocExceptions(out, op, doc);
         }
 
+        const auto& remarks = doc.remarks();
```
bernardnormier, line 568: Same here.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -772,15 +772,28 @@ Slice::CsVisitor::writeDocComment(const ContainedPtr& p, const string& generated
...
+    {
+        remarks = comment->remarks();
+    }
     if (!generatedType.empty())
     {
-        _out << nl << "/// <remarks>" << "The Slice compiler generated this " << generatedType << " from Slice "
-             << p->kindOf() << " <c>" << p->scoped() << "</c>.";
+        if (!remarks.empty())
+        {
+            remarks.emplace_back("");
```
bernardnormier, line 784: I don't think blank lines are expected/correct in C# doc-comments. We should look into using `<para>`.

InsertCreativityHere:
In the case where there's both user-supplied remarks, and compiler-generated remarks,
I removed the blank line and replaced it with a paragraph break (ie. a self-closing `<para />` tag).
Here is what the rendered output for `@remarks` looks like for C# DocFX (note the line-break):
image
And here is the geneated code:
```cs
    /// <summary>Observes the status of services in an IceBox server.</summary>
    /// <remarks>This interface is implemented by admin tools that monitor the IceBox server.
    /// <para />
    /// The Slice compiler generated this client-side interface from Slice interface <c>::IceBox::ServiceObserver</c>.
    /// Use the methods of this interface to invoke operations on a remote Ice object that implements <c>ServiceObserver</c>.</remarks>
    /// <seealso cref="ServiceManager.addObserver" />
```

cpp/src/slice2cs/Gen.cpp
InsertCreativityHere: I removed the blank line and replace it with a `para` tag.
See https://github.com/zeroc-ice/ice/pull/3962#issuecomment-2881104196

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Indeed, I removed all the parenthesis.

cpp/src/slice2cpp/Gen.cpp
InsertCreativityHere: I'm also surprised that this wasn't given a warning, or even lint violation...
Regardless, I fixed it.

csharp/docfx/docfx.json
```diff
@@ -58,7 +58,7 @@
       "_appFaviconPath": "images/favicon.ico",
       "_appLogoPath": "images/logo.png",
       "_appLogoUrl": "https://doc.zeroc.com/api/ice/main/csharp",
-      "_appFooter": "© 2024 ZeroC",
```
InsertCreativityHere, line 61: I generated the docs to see how it would be rendered, and noticed that the copyright years as old.
So I just went ahead and updated it.

cpp/test/Slice/errorDetection/WarningInvalidComment.ice
```diff
@@ -18,7 +18,7 @@ module Test
     }
 
     /// This is a test overview.
-    /// @remark: This is an unknown comment tag which spans 1 line.
```
InsertCreativityHere, line 21: This test is about testing invalid tags.
Now that `@remark` is valid, we needed a new made-up tag instead!

cpp/src/slice2cs/Gen.cpp
```diff
@@ -767,20 +767,33 @@ void
...
     }
 
     if (!generatedType.empty())
     {
-        _out << nl << "/// <remarks>" << "The Slice compiler generated this " << generatedType << " from Slice "
-             << p->kindOf() << " <c>" << p->scoped() << "</c>.";
+        // If there's user-provided remarks, and a generated-type message, we introduce a paragraph between them.
+        if (!remarks.empty())
+        {
+            remarks.push_back("<para />");
```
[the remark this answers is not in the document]
InsertCreativityHere: So, `<br/>` is supported, but seems like a relatively recent addition (at least officially).
But I think that `<para>` is more appropriate, and looks better.
`br` just starts a new-line, whereas `para` will also add vertical space between the 2 sections.

Whatever remarks the user adds are going to be completely unrelated to our "this was generated from XXX" message.
Having this extra space makes it clear these are unrelated. whereas `br` gives the impression that these sentences go together, but for some reason we just wanted to control where the line started.

bernardnormier (approved):

Issue #2274, opened 2024-06-06
Update slice2swift doc-comment generation


bernardnormier:
See https://www.swift.org/documentation/docc/writing-symbol-documentation-in-your-source-files

The doc-comment we currently generate are not quite correct:
 - parameter/returns should be capitalized
 - parameter should not list the parameter type
 - there are no many blank lines
 - we should use the Parameters style when there are more than 2 parameters

InsertCreativityHere:
Even after I fix the doc-comment generation, we still need to go through our own Swift code and make all the fixes mentioned in this comment. Because there's plenty of mistakes throughout those comments too, and not even just from the checked-in generated code.

InsertCreativityHere:
This is fixed in the doc-comment generation.
But I'm leaving this issue open because we still need to fix all these mistakes in the hand-written API comments.
I'm also editing the title to reflect this fact.

externl:
I made fixes to the source in https://github.com/zeroc-ice/ice/pull/4503

PR #3963, opened 2025-05-13
Fix Swift docc warnings

externl:
This PR fixes various warnings I noticed when reviewing the Swift doc action.  Some were just missing parameters, others were caused by documentation the "external value" of a parameter. For example:

```
warning: External name 'value' used to document parameter
- .build/plugins/outputs/ice/Ice/destination/CompileSlice/BuiltinSequences.swift:71:21-71:26
///
69
70
71 +
/// - Parameter ostr: The stream to write to.
/// - Parameter value: The sequence value to write to the stream. suggestion: Replace 'value' with 'v'
72
73
public static func write(to ostr: OutputStream, value v: ObjectSeq)
ostr.write(size: v.count)
warning: External name 'value' used to document parameter
```

I also updated the `swift-docc-plugin` plugin version to the latest.

### DIFF
--- a/swift/src/Ice/ToStringMode.swift
+++ b/swift/src/Ice/ToStringMode.swift
@@ -54,16 +54,16 @@ extension InputStream {
 extension OutputStream {
     /// Writes an enumerated value to the stream.
     ///
-    /// parameter v: `ToStringMode` - The enumerator to write.
-    public func write(_ v: ToStringMode) {
-        write(enum: v.rawValue, maxValue: 2)
+    /// - parameter value: `ToStringMode` - The enumerator to write.
+    public func write(_ value: ToStringMode) {
+        write(enum: value.rawValue, maxValue: 2)
     }
 
     /// Writes an optional enumerated value to the stream.
     ///
-    /// parameter tag: `Int32` - The numeric tag associated with the value.
+    /// - parameter tag: `Int32` - The numeric tag associated with the value.
     ///
-    /// parameter v: `ToStringMode` - The enumerator to write.
+    /// - parameter value: `ToStringMode` - The enumerator to write.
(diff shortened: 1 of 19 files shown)


swift/src/Ice/ToStringMode.swift
```diff
@@ -63,7 +63,7 @@ extension OutputStream {
     ///
     /// parameter tag: `Int32` - The numeric tag associated with the value.
     ///
-    /// parameter _: `ToStringMode` - The enumerator to write.
+    /// parameter v: `ToStringMode` - The enumerator to write.
```
InsertCreativityHere, line 66:
```suggestion
    /// parameter value: `ToStringMode` - The enumerator to write.
```

pepone (approved):
Looks good.

Looking at swift docs they capitalize Parameter, Returns, etc

https://developer.apple.com/documentation/xcode/writing-symbol-documentation-in-your-source-files

InsertCreativityHere (approved):
Looks good to me!

InsertCreativityHere:
> Looking at swift docs they capitalize Parameter, Returns, etc

There's an open issue to fix this: #2274
Would be good to fix, but feels like it belongs in a different PR than this one.

bernardnormier (approved):
We need a much more thorough fix on these Swift doc-comments, that uses best practices.

See for example:
https://developer.apple.com/documentation/xcode/writing-symbol-documentation-in-your-source-files

https://www.swift.org/documentation/docc/writing-symbol-documentation-in-your-source-files

In particular:
 - Why are we inserting all these types? This makes no sense.
 - Parameter(s) and Returns should be capitalized
 - Extra `-` in Returns
 - No need for all these blank lines

PR #3964, opened 2025-05-14
Python API doc improvements

pepone:
This PR contains several fixes and improvements to the generated Python API doc.


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -111,15 +111,15 @@ namespace
         SequencePtr seq = dynamic_pointer_cast<Sequence>(type);
         if (seq)
         {
-            return typeToDocstring(seq->type(), false) + "[]";
+            return "list[" + typeToDocstring(seq->type(), false) + "]";
```
pepone, line 114: This format has the advantage of working with sphinx auto-linking

python/docs/index.rst
```diff
@@ -5,5 +5,9 @@ Ice documentation
    :maxdepth: 2
    :caption: Contents:
```
pepone, line 7: Added the missing modules.

python/python/Ice/BatchRequestInterceptor.py
```diff
@@ -1,6 +1,6 @@
 # Copyright (c) ZeroC, Inc.
 
-class BatchRequestInterceptor(object):
+class BatchRequestInterceptor:
```
pepone, line 3: This is not required with Python 3.x, all classes are implicitly derived from object.

python/python/Ice/Communicator.py
```diff
@@ -15,22 +15,21 @@ class Communicator:
 
     Example
     -------
-    The following example shows how to create a communicator and use the `with` statement to ensure that the
+    The following example shows how to create a communicator and use the ``async with`` statement to ensure that the
```
pepone, line 18: The double back-tick is the restructure text format for inline code blocks.

python/python/Ice/Future.py
```diff
@@ -19,7 +19,7 @@ def __await__(self):
 
 def wrap_future(future, *, loop=None):
     """
-    Wrap an Ice.Future object into an asyncio.Future.
+    Wrap an :class:`Ice.Future` object into an ``asyncio.Future``.
```
pepone, line 22: this `:class:` shows how to cross-reference see also https://www.sphinx-doc.org/en/master/usage/domains/python.html

python/python/Ice/ObjectPrx.py
```diff
@@ -946,6 +946,3 @@ def __repr__(self):
 
     def __str__(self):
         return repr(self)
-
-
-    __module__ = 'Ice'
```
pepone, line 951: __module__ is really not required, replaced with `__all__` in `Ice/__init__.py`

python/python/Ice/__init__.py
```diff
@@ -113,3 +113,108 @@
 # Native PropertiesAdmin admin facet.
 #
 NativePropertiesAdmin = IcePy.NativePropertiesAdmin
+
+__all__ = [
```
pepone, line 117: __all__ defines the the public API, used by `import *` and by sphinx to figure out what types are documented with automodule.

Lot of types where missing in the previous setup.

slice/IceGrid/Admin.ice
```diff
@@ -591,7 +591,7 @@ module IceGrid
         /// message doesn't exceed the given size.
         /// @param lines The lines read from the file. If there was nothing to read from the file since the last call to
         /// read, an empty sequence is returned. The last line of the sequence is always incomplete (and therefore no
-        /// '\n' should be added when writing the last line to the to the output device).
+        /// new-line character should be added when writing the last line to the to the output device).
```
pepone, line 594: Mapping the `\n` correctly to sphinx requires double escaping when it appears in regular text, or escaping when in a code block.

externl (approved):

python/python/Ice/__init__.py
bernardnormier: Would be nice to add a comment in the code.

slice/IceGrid/Admin.ice
bernardnormier: newline in one word

bernardnormier (approved):

python/python/Ice/Exception.py
```diff
@@ -1,13 +1,20 @@
...
 
     def ice_id(self):
-        """Returns the type id of this exception."""
-        return self._ice_id
+        """
+        Returns the type ID of this exception. This corresponds to the Slice type ID for Slice-defined exceptions,
+        and to a similar fully scoped name for other exceptions. For example "::Ice::CommunicatorDestroyedException".
 
-    __module__ = "Ice"
+        Returns
```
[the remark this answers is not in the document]
pepone: Reworked a bit.

python/python/Ice/Future.py
pepone: I think it works both ways, at least when using in a Python file.

Issue #3457, opened 2025-01-30
Review Doc-Comment Link Generation in Each Language


InsertCreativityHere:
Now that the link-formatter in the parser has more context to work with, and a pointer to what's being linked to,
we should review our link generation in each of the languages to make sure it's idiomatic, well-formatted, and functional.

These should also be updated to take the new `xxx:identifier` metadata into account as it is added to each language.

- [x] C++
- [ ] C#
- [ ] Java
- [ ] JavaScript
- [ ] MATLAB
- [ ] PHP
- [ ] Python
- [ ] Ruby
- [ ] Swift

pepone:
C# was fixed in d062993f3684ab08f87b842f94c4470e2069d7c4

PR #3966, opened 2025-05-14
Improve Python link formatter

pepone:
Python side of #3457

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -220,7 +220,7 @@ namespace
             else
             {
                 string targetScoped = dynamic_pointer_cast<Contained>(target)->mappedScoped(".").substr(1);
-                result << ":class:`" result << targetScoped;
+                result << ":class:`" << targetScoped;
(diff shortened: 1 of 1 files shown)


externl (approved):

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -186,29 +186,77 @@ namespace
...
+            else
+            {
+                if (auto operationTarget = dynamic_pointer_cast<Operation>(target))
+                {
+                    string targetScoped = operationTarget->interface()->mappedScoped(".").substr(1);
+
+                    // link to the method on the proxy interface
+                    result << ":meth:`" << targetScoped << "Prx." << operationTarget->mappedName() << "`";
+                }
+                else
```
InsertCreativityHere, line 222: I think this would be better as:
```
if (isBuiltin) {}
elseif (isOperation) {}
else {}
```
instead of the current:
```
if (isBuiltin) {}
else {
    if (isOperation) {}
    else {}
}
```
It's not like we're doing anything else in that `else` block.

InsertCreativityHere (approved):
Looks good to me!
Definitely better than what we have now!

PR #3968, opened 2025-05-14
Simplify implementation of PropertiesAdmin facet in C++

bernardnormier:
Before this PR, NativePropertiesAdmin was an interface (abstract base class), and PropertiesAdminI was the servant class that implemented everything.

This extra abstract base class was confusing; it looked as if the user was supposed to implement it.

This PR converts  NativePropertiesAdmin into the (final) servant class. The public API remains the same.


cpp/src/Ice/NativePropertiesAdmin.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+Ice::NativePropertiesAdmin::~NativePropertiesAdmin() = default; // Out of line to avoid weak vtable
+
+Ice::NativePropertiesAdmin::NativePropertiesAdmin(PropertiesPtr properties, LoggerPtr logger)
+    : _properties(std::move(properties)),
+      _logger(std::move(logger))
+{
+}
+
+string
```
bernardnormier, line 24: All the code below comes from the previous PropertiesAdminI.cpp.

externl (approved):

pepone (approved):

PR #3971, opened 2025-05-15
Add default values to generated memberwise initializer in Swift

bernardnormier:
This PR add default values to the generated memberwise initializer in Swift (for structs, classes, and maybe exceptions).

This matches the behavior of the default memberwise initializer for structs, that we can't use because it's internal:
https://docs.swift.org/swift-book/documentation/the-swift-programming-language/initialization/#Default-Initializers


pepone (approved):

externl (approved):
Looks good!

PR #3973, opened 2025-05-15
No need to cancel invocations after the future is done

pepone:
This is a minor fix for Python Invocation futures to avoid calling the cancel closure via the async invocation context if the future is already done.

### DIFF
--- a/python/python/Ice/InvocationFuture.py
+++ b/python/python/Ice/InvocationFuture.py
@@ -40,8 +40,10 @@ class InvocationFuture(Future):
         After cancellation, `done()` returns `True`, and attempting to retrieve the result will raise
         an `Ice.InvocationCanceledException`.
         """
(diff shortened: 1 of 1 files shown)


externl (approved):

python/python/Ice/InvocationFuture.py
```diff
@@ -40,8 +40,10 @@ def cancel(self):
         After cancellation, `done()` returns `True`, and attempting to retrieve the result will raise
         an `Ice.InvocationCanceledException`.
         """
-        self._asyncInvocationContext.cancel()
-        return Future.cancel(self)
+        canceled = Future.cancel(self)
```
bernardnormier, line 43: We should use the Python spelling, `cancelled`.

bernardnormier (approved):

Issue #3341, opened 2025-01-10
Android CI failure


externl:
```
Ran 40 tests in 6.0 minutes 9.94 seconds
39 succeeded and 1 failed:
- Ice/background
  [ running client/server test - 01/10/25 20:19:33 ]
  - Config: ssl,android
  (executing `Ice/background/Server' on `Android' args = ['--Ice.Default.Host=127.0.0.1', '--Ice.Warn.Connections=1', '--Ice.Default.Protocol=ssl', '--Ice.IPv6=0', '--IceSSL.***', '--IceSSL.DefaultDir=/home/runner/work/ice/ice/certs', '--IceSSL.KeystoreType=BKS', '--IceSSL.TruststoreType=BKS', '--IceSSL.Keystore=server.bks', '--Ice.ThreadPool.Server.Size=1', '--Ice.ThreadPool.Server.SizeMax=3', '--Ice.ThreadPool.Server.SizeWarn=0', '--Ice.PrintAdapterReady=1'])
  (executing `Ice/background/Client' on `Android' args = ['--Ice.Default.Host=127.0.0.1', '--Ice.Warn.Connections=1', '--Ice.Default.Protocol=ssl', '--Ice.IPv6=0', '--IceSSL.***', '--IceSSL.DefaultDir=/home/runner/work/ice/ice/certs', '--IceSSL.KeystoreType=BKS', '--IceSSL.TruststoreType=BKS', '--IceSSL.Keystore=client.bks'])
  [...]
    File "/home/runner/work/ice/ice/java/../scripts/Util.py", line 1502, in stop
      process.terminate()
      ~~~~~~~~~~~~~~~~~^^
    File "/home/runner/work/ice/ice/java/../scripts/Util.py", line 2495, in terminate
      self.output = self.proxy.terminate().strip()
                    ~~~~~~~~~~~~~~~~~~~~^^
    File "/home/runner/work/ice/ice/scripts/Controller.ice", line 385, in terminate
  Ice.ConnectionLostException: connection lost: recv() returned zero
Waiting for the emulator to shutdown... ok
```

pepone:
It seems like the emulator crash:

```
Storing crashdata in: /tmp/android-runner/emu-crash-35.3.11.db, detection is enabled for process: 15592
2025-01-14T11:20:55.4530949Z WARNING      | Your GPU drivers may have a bug. Switching to software rendering.
[...]

PR #3974, opened 2025-05-15
Enable Android testing

pepone:
Enabled back Android CI to see if we can reproduce #3341

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,13 +88,12 @@ jobs:
             test_flags: "--platform=x64"
             build_cpp_and_python: true
 
-          # Android (disabled until we fix https://github.com/zeroc-ice/ice/issues/3341)
-          # - os: ubuntu-24.04
-          #   config: "android"
-          #   working_directory: "java"
-          #   test_flags: "--android --controller-app"
-          #   build_cpp_and_python: true
-          #   build_android_controller: true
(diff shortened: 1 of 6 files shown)


java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerApp.java
```diff
@@ -15,6 +15,10 @@
 import com.zeroc.Ice.Logger;
 import com.zeroc.Ice.Time;
 
+import java.util.List;
+import java.util.Locale;
+import java.io.Writer;
```
pepone, line 20: We used to have a `*` import, seems it was removed by checkstyle changes, and we didn't notice because we are not building Android on CI.

.github/workflows/ci.yml
```diff
@@ -88,13 +88,12 @@ jobs:
...
-          #   working_directory: "java"
-          #   test_flags: "--android --controller-app"
-          #   build_cpp_and_python: true
-          #   build_android_controller: true
+          - os: ubuntu-24.04
+            config: "android"
+            working_directory: "java"
+            test_flags: "--android --controller-app"
+            build_cpp_and_python: true
+            build_android_controller: true
```
pepone, line 97: Adding this back to see if it is more stable, and ensure we don't accidentally break things.

bernardnormier (approved):

PR #3978, opened 2025-05-16
Fix bogus sent callback in C++ echo server used by JavaScript tests

### LINKED ISSUE #3957: js Ice/operations test failure
Echo server crash running this test on Windows

https://github.com/zeroc-ice/ice/actions/runs/14976264260/job/42069284174

[crash-dumps-release-windows-2022.zip](https://github.com/user-attachments/files/20173658/crash-dumps-release-windows-2022.zip)
[test-logs-release-windows-2022.zip](https://github.com/user-attachments/files/20173659/test-logs-release-windows-2022.zip)

pepone:
I think this is likely the cause why Echo server was crashing. Fix #3957

### DIFF
--- a/cpp/test/Ice/echo/BlobjectI.cpp
+++ b/cpp/test/Ice/echo/BlobjectI.cpp
@@ -69,9 +69,9 @@ BlobjectI::ice_invokeAsync(
                 current.operation,
                 current.mode,
                 inEncaps,
-                [](bool, const std::vector<byte>&) { assert(0); },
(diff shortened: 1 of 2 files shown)


externl (approved):

bernardnormier (approved):
So the crash was occuring for oneway requests that could not get forwarded synchronously?

pepone:
> So the crash was occuring for oneway requests that could not get forwarded synchronously?

I didn't get a stack from the crash, but test failures where in this code path.

PR #3986, opened 2025-05-16
Fix queue marshaling in slice2cs

bernardnormier:


externl (approved):

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1239,15 +1239,7 @@ Slice::CsGenerator::writeSequenceMarshalUnmarshalCode(
 
             if (isGeneric && !isList && !isStack)
             {
-                if (isValueType(type))
-                {
-                    out << "e.Current";
-                }
-                else
-                {
-                    out << "(e.Current is null ? new ";
```
bernardnormier, line 1248: We should not "fix" null in sequences of non-nullable elements.

pepone (approved):

PR #3988, opened 2025-05-17
Test PKCS12 certificates with Java IceSSL/configuration test

### LINKED ISSUE #3826: Try using PKCS12 certificates with Java
When generating p12 certificates with Java 21 keytool I got this warning:

```
(venv) jose@mac certs % keytool -noprompt -importkeystore -srcalias server  -srckeystore server.p12 -srcstorepass:file password.txt -srcstoretype PKCS12 -destkeystore server.jks -deststorepass:file password.txt -destkeypass:file password.txt  -destalias server -deststoretype JKS 
Importing keystore server.p12 to server.jks...

Warning:
The JKS keystore uses a proprietary format. It is recommended to migrate to PKCS12 which is an industry standard format using "keytool -importkeystore -srckeystore server.jks -destkeystore server.jks -deststoretype pkcs12".
```

If we do so we should update IceGrid GUI to allow importing this certificates.

pepone:
Fix #3826

PKCS12 is already supported by Java out of the box. The `IceSSL.KeystoreType` and `IceSSL.TruststoreType` property can be used to change the store type to use, this is already documented in 3.7.

This PR updates the IceSSL/configuration to run tests with both JKS and PKCS12.


bernardnormier (approved):
Using an enum for the key store type (as suggested by Copilot) would be indeed better.

Should we also add a README.md in the certs directly describing what these certs are and how to regenerate them. When these certs expired, we should be able to regenerate them without too much research.

pepone:
> Should we also add a README.md in the certs directly describing what these certs are and how to regenerate them. When these certs expired, we should be able to regenerate them without too much research.

I want to rework the scripts we use to generate the certificates. We currently rely on the custom CA utils provided by `zeroc-icecertutils`. Moving to a plain OpenSSL based scripts, and keytool for the JKS/BKS side. I will open a separate PR for this.

PR #3990, opened 2025-05-18
Update certificates used to test ssl transport

pepone:
This PR updates the certificates used for test the ssl transport. The ones in the top-level "certs" directory.

The `makecerts.sh` bash script can be used to regenerate the certificates. It depends on the `openssl` and `keytool` commands.

There is only PEM and PKCS12 certificates.

I have updated Java and Android tests for ssl transport to use PKCS12 instead of JKS and BKS.

I plan to update the certificates and scripts for `IceSSL/configuration` in a follow up PR.


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/SSLEngine.java
```diff
@@ -131,17 +131,13 @@ public void initialize() {
                 //
                 KeyManager[] keyManagers = null;
                 KeyStore keys = null;
-                if (_keystoreStream != null || !keystorePath.isEmpty()) {
```
pepone, line 134: With 3.7 we have setters in the Plugin API for keystore and truststore, but that was remove and this code was not properly cleanup.

java/test/android/controller/build.gradle
```diff
@@ -79,16 +79,18 @@ android {
     }
 }
 
-task copyBksTask(type: Copy) {
-    from "${rootProject.projectDir}/../../../../certs/client.bks"
-    from "${rootProject.projectDir}/../../../../certs/server.bks"
+task copyCertificatesTask(type: Copy) {
```
pepone, line 82: Modern Android versions come with PKCS12 support, and generating BKS certificates is always a pain because the required cryptographic provider is not installed by default with the JDK.

scripts/Util.py
```diff
@@ -3738,24 +3738,16 @@ def getJavaArgs(self, process, current):
...
-            )
+        props.update(
+            {
+                "IceSSL.Keystore": "server.p12" if isinstance(process, Server) else "client.p12",
+                "IceSSL.KeystorePassword": "password",
+                "IceSSL.KeystoreType": "PKCS12",
+                "IceSSL.Truststore": "ca.p12",
+                "IceSSL.TruststorePassword": "password",
+                "IceSSL.TruststoreType": "PKCS12",
+            }
```
pepone, line 3749: We now use the same configuration for both Android and Java. Additionally, the configuration has been updated to explicitly use a truststore, rather than relying on the keystore being used as the truststore when the latter is not set.

bernardnormier (approved):

certs/server.cnf
```diff
@@ -0,0 +1,26 @@
...
+O  = ZeroC,\ Inc.
+OU = Ice
+CN = 127.0.0.1
+emailAddress = info@zeroc.com
+
+[ v3_extensions ]
+subjectKeyIdentifier = hash
+authorityKeyIdentifier = keyid:always,issuer
+keyUsage = critical, digitalSignature, nonRepudiation, keyEncipherment
+extendedKeyUsage = serverAuth, clientAuth
```
externl, line 21: Is `clientAuth` correct here?

externl (approved):

certs/makecerts.sh
```diff
@@ -0,0 +1,46 @@
...
+
+    # Remove CSR after signing
+    rm -f ${name}.csr
+
+    # Export to PKCS#12
+    rm -f ${name}.p12
+    openssl pkcs12 -export -out ${name}.p12 \
+        -inkey ${name}_key.pem -in ${name}_cert.pem -certfile ca_cert.pem \
+        -name "${name}" -passout pass:"${PASSWORD}" \
+        -legacy # For compatibility with macOS Keychain
```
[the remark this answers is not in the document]
pepone: Basically the p12 support in macOS is limited to old algorithms 3DES, and such that are now marked as legacy in OpenSSL 3. The new defaults are not supported.

We already have this issue with the old certs: https://github.com/zeroc-ice/icecertutils/commit/e2038a5c686e47355a1d2c3a025300effebb28de#diff-f7a5e846d1dee97d1402efc0b0dcebf3f42c04946fea608af3aceb060999503dR122-R135

certs/server.cnf
pepone: We have it before but not sure if it is strictly required.

certs/makecerts.sh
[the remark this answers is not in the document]
externl: Have we considered moving away from p12 and just using separate certificate and key files? In my experience this is the more common approach for deploying web servers and getting certs from Let's Encrypt and other CAs?

certs/server.cnf
externl: I'm pretty sure it's incorrect, can we try remming it.

certs/makecerts.sh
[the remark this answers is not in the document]
pepone: Actually I was mistaken here, we don't support providing the key in a separate file.

certs/server.cnf
pepone: Seems this is required for C#, when the server certificate is used as the "client".

This is a call to self:

```
    public override async Task<int>
    opWithResultAsyncDispatchAsync(Ice.Current current)
    {
        test(Thread.CurrentThread.Name.Contains("Ice.ThreadPool.Server"));
        await Task.Yield();
        test(!Thread.CurrentThread.Name.Contains("Ice.ThreadPool.Server"));
        return await self(current).opWithResultAsync();
    }
```

I going to keep this for now and add a comment in the file.

PR #3992, opened 2025-05-19
Add PHP API generation to GitHub actions

externl:
This PRs adds PHP API reference generation to our GitHub actions. 

Something things are missing and there's more work to be done.


php/phpdoc.dist.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpdocumentor
+    configVersion="3"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="https://www.phpdoc.org"
+>
+    <title>Ice 3.8 Slice API Reference</title>
```
pepone, line 7: is Slice correct here? or are we also documenting the non Slice APIs.

pepone (approved):

PR #3996, opened 2025-05-19
Don't use IceSSL.Keystore as fallback for IceSSL.Truststore

pepone:
This PR updates Java IceSSL configuration to not use IceSSL.Truststore as a fallback for IceSSL.Keystore.

There is some additional cleanup, but nothing else should change behaviors.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/SSLEngine.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/SSLEngine.java
@@ -278,11 +278,9 @@ public class SSLEngine {
                     }
                 }
 
-                //
-                // Collect the trust managers. Use IceSSL.Truststore if specified, otherwise use the
-                // Java root CAs if Ice.Use.PlatformCAs is enabled. If none of these are enabled,
-                // use the keystore or a dummy trust manager which rejects any certificate.
-                //
+                // Collect the trust managers. Use IceSSL.Truststore if specified, otherwise use the Java root CAs if
+                // Ice.Use.PlatformCAs is enabled. If none of these are enabled use a dummy trust manager which
+                // rejects any certificate.
                 javax.net.ssl.TrustManager[] trustManagers = null;
                 {
                     String algorithm = TrustManagerFactory.getDefaultAlgorithm();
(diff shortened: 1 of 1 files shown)


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/SSLEngine.java
```diff
@@ -182,37 +151,29 @@ public void initialize() {
...
-                    }
-
-                    if (!alias.isEmpty()) {
-                        //
-                        // If the user selected a specific alias, we need to wrap the key managers
-                        // in order to return the desired alias.
-                        //
+                    } else {
+                        // If the user selected a specific alias, ensure it correspond with a key entry.
                         if (!keys.isKeyEntry(alias)) {
```
pepone, line 169: Previously we check this below in ` if (!alias.isEmpty()) {` but is only useful here, when the "alias" was provided by the user. Don't need to do that when we just retrieve it from the keystore doing the same check.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/SSLEngine.java
```diff
@@ -278,85 +232,62 @@ public void initialize() {
...
+                                    }
 
-                    //
-                    // Attempting to establish an outgoing connection with an empty truststore can
-                    // cause hangs that eventually result in an exception such as:
-                    //
-                    // java.security.InvalidAlgorithmParameterException: the trustAnchors parameter
-                    // must be non-empty
-                    //
-                    if (trustStore != null && trustStore.size() == 0) {
```
pepone, line 333: move this up, to the only branch with a non null trustStore.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/SSLEngine.java
```diff
@@ -278,85 +232,62 @@ public void initialize() {
...
             } catch (GeneralSecurityException ex) {
                 throw new InitializationException(
                     "SSL transport: unable to initialize context", ex);
             }
         }
-
-        //
-        // Clear cached input streams.
-        //
-        _keystoreStream = null;
```
pepone, line 358: never used.

bernardnormier (approved):

externl (approved):

externl (approved):
Does this warrant a changelog entry?

pepone:
> Does this warrant a changelog entry?

Sure, I will go over all IceSSL changes later and update the changelog

PR #3998, opened 2025-05-19
Update READMEs

bernardnormier:
This PR updates the READMEs for 7 out of 9 languages (not updated: JS and Swift).

The goal is to keep the README very simple, with a few links and short sample code. The sample code is a simplified version of the Greeter demo.

### DIFF
--- a/cpp/README.md
+++ b/cpp/README.md
@@ -1,24 +1,28 @@
 # Ice for C++
 
-[Getting started] | [Examples] | [NuGet packages] | [Documentation] | [Building from source]
+[Examples] | [Ice Manual] | [Building from source]
 
 The [Ice framework] provides everything you need to build networked applications,
 including RPC, pub/sub, server deployment, and more.
 
-Ice for C++ is the C++ implementation of Ice.
+Ice for C++ is the C++ implementation of the Ice framework.
 
 ## Sample Code
 
 ```slice
-// Slice definitions (Hello.ice)
+// Slice definitions (Greeter.ice)
 
 #pragma once
 
-module Demo
+module VisitorCenter
 {
-    interface Hello
+    /// Represents a simple greeter.
+    interface Greeter
     {
-        void sayHello();
+        /// Creates a personalized greeting.
+        /// @param name The name of the person to greet.
+        /// @return The greeting.
+        string greet(string name);
     }
 }
 ```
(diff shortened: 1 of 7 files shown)


cpp/README.md
```diff
@@ -1,118 +1,119 @@
...
-    {
-        cerr << ex.what() << endl;
-        return 1;
-    }
+    Ice::CommunicatorPtr communicator = Ice::initialize(argc, argv);
+    Ice::CommunicatorHolder communicatorHolder{communicator};
+
+    VisitorCenter::GreeterPrx greeter{communicator, "greeter:tcp -h localhost -p 4061"};
+
+    string greeting = greeter.greet("alice");
```
pepone, line 47: Should we prefer `->` like in the demos?

java/README.md
```diff
@@ -49,44 +53,42 @@ public class Client
 
 import com.zeroc.Ice.*;
```
pepone, line 54: Remove this * import

python/README.md
```diff
@@ -10,62 +10,71 @@ Ice for Python is the Python implementation of the Ice framework.
...
-# once it goes out of scope.
-with Ice.initialize(sys.argv) as communicator:
-    hello = Demo.HelloPrx(communicator, "hello:default -h localhost -p 10000")
-    hello.sayHello()
+async def main():
+    async with Ice.initialize(sys.argv, eventLoop=asyncio.get_running_loop()) as communicator:
+        greeter = VisitorCenter.GreeterPrx(communicator, "greeter:tcp -h localhost -p 4061")
+        greeting = await greeter.greetAsync(getpass.getuser())
+        print(greeting)
 ```
```
pepone, line 44: We should probably include the call to main, in the client and server snippets

```
if __name__ == "__main__":
    asyncio.run(main())
```

pepone (approved):

cpp/README.md
bernardnormier: Yes, even though I find this syntax a bit odd without a comment.

What's the advantage of `->`?
 - it's the traditional invocation syntax in C++ (so highly desirable to keep it)
 - it allows us to use the same invocation syntax with non-nullable proxies and nullable proxies (thanks to `operator->`  on std::optional).

Will fix.

cpp/README.md
```diff
@@ -1,118 +1,119 @@
 # Ice for C++
 
-[Getting started] | [Examples] | [NuGet packages] | [Documentation] | [Building from source]
+[Examples] | [Ice Manual] | [Building from source]
```
externl, line 3: I think we should say documentation. It's fine that we call it that but I think when linking to it we should avoid the term manual.

java/README.md
```diff
@@ -49,44 +53,42 @@ public class Client
...
-    public void sayHello(com.zeroc.Ice.Current current)
-    {
-        System.out.println("Hello World!");
+    public String greet(String name, Current current) {
+        System.out.println("Dispatching greet request { name = '" + name + "' }");
+        return "Hello, " + name + "!";
     }
-}
-```
+}``
```
externl, line 89:
```suggestion
}
```
```

externl (approved):
Looks good.

PR #3999, opened 2025-05-20
Remove XML Stripping from Doc-Comment Parsing

InsertCreativityHere:
Currently, the doc-comment parser can strip out XML tags from doc-comments, controlled by a bool `stripMarkup`. However, Slice doc-comments have no special logic for XML tags, and we don't recommend using them.

This PR removes this logic. So now any XML tags in comments will be preserved.
This only affects `ice2slice`, `slice2cs`, `slice2matlab`, `slice2py`, and `slice2swift`; the only languages where `true` was passed for this bool.

### DIFF
--- a/cpp/src/IceGrid/Internal.ice
+++ b/cpp/src/IceGrid/Internal.ice
@@ -389,7 +389,7 @@ module IceGrid
     interface InternalRegistry extends FileReader
     {
         /// Register a node with the registry. If a node with the same name is already registered,
-        /// <code>registerNode</code> overrides the existing registration only when the previously
+        /// registerNode overrides the existing registration only when the previously
         /// registered node is not active.
         /// @param info Some information on the node.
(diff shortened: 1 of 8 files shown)


externl (approved):

cpp/src/IceGrid/Internal.ice
```diff
@@ -389,7 +389,7 @@ module IceGrid
     interface InternalRegistry extends FileReader
     {
         /// Register a node with the registry. If a node with the same name is already registered,
-        /// <code>registerNode</code> overrides the existing registration only when the previously
+        /// registerNode overrides the existing registration only when the previously
```
bernardnormier, line 392: I would prefer "this operation" when referencing the operation we're describing. Same for registerReplica below.

bernardnormier (approved):

PR #4001, opened 2025-05-20
Remove Ice.BlobjectAsync python class

### LINKED ISSUE #3994: Python doc warning
```
/Users/runner/work/ice/ice/python/python/Ice/__init__.py:docstring of Ice.BlobjectAsync.BlobjectAsync.ice_invoke:16: ERROR: Unexpected indentation. [docutils]
/Users/runner/work/ice/ice/python/python/Ice/__init__.py:docstring of Ice.BlobjectAsync.BlobjectAsync.ice_invoke:17: WARNING: Block quote ends without a blank line; unexpected unindent. [docutils]
```

pepone:
Fix #3994

In Python mapping the `Ice.BlobjectAsync` class is identical to the `Ice.Blobject`. We don't need both. This is because in Python you don't nee different classes for sync and async dispatches, is all dynamic base on the return type.


python/test/Ice/blobject/RouterI.py
```diff
@@ -70,7 +70,7 @@ def done(self, future):
             self._future.set_exception(ex)
 
 
-class BlobjectAsyncI(Ice.BlobjectAsync):
+class BlobjectAsyncI(Ice.Blobject):
```
bernardnormier, line 73: It would be nice to refactor (simplify) this test and use asyncio for the async portion. For a follow-up PR?

PR #4003, opened 2025-05-20
Update C# BUILDING

bernardnormier:


pepone (approved):
Looks good, in a follow up PR we should update to mention how Slice compilers are included in the the tools package.

For example the default only includes the slice2cs compiler for the current platform.

externl (approved):

PR #4005, opened 2025-05-20
Update doc comments generated by slice2java

externl:
This PR is phase 1 of updating the generated API reference for Ice for Java.  Its goal is to reduce the number of javadoc warnings coming from the generated code. 

This means generating doc comments for constructors, various marshaling helper methods, and more.

Note that while `@hidden` can be used to exclude something from being included in the Javadoc output, it's unfortunately not excluded from javadoc parsing, so we need doc comments for these as well.


cpp/src/Slice/StringLiteralUtil.cpp
externl, line 1: I moved this here from C++.

cpp/src/slice2java/Gen.cpp
```diff
@@ -1801,25 +1830,11 @@ Slice::Gen::TypesVisitor::visitClassDefStart(const ClassDefPtr& p)
...
-            {
-                hasRequiredMembers = true;
-            }
-        }
-
         //
         // Default constructor.
         //
         out << sp;
+        writeDocComment(out, "Constructs a new " + name + ".");
```
bernardnormier, line 1837: The Java standard is "Creates a new ...".

InsertCreativityHere:
For anyone curious, here's the _actual_ changes caused by this PR:
https://github.com/InsertCreativityHere/compiler-comparison/commit/442e06aed6f90386baf2b596f369b95b3b8ab330
It shows a diff of the generated code before and after this PR via my little compiler-comparison tool.

cpp/src/slice2java/Gen.cpp
```diff
@@ -1840,22 +1855,24 @@ Slice::Gen::TypesVisitor::visitClassDefStart(const ClassDefPtr& p)
...
-            if (hasRequiredMembers && hasOptionalMembers)
+            if (!requiredMembers.empty() && !optionalMembers.empty())
             {
                 //
                 // Generate a constructor accepting parameters for just the required members.
                 //
-                out << sp << nl << "public " << name << spar;
+                out << sp;
+                out << nl << "/**";
+                out << nl << " * Constructs a new {@code " << name << "} with required fields initialized.";
```
bernardnormier, line 1865: I'd rather avoid the term `required`. What about:
```text
Creates a new NAME with values for all fields not marked optional in the Slice definition for NAME. 
```

Note that this constructor is unique to Java. I wish we did not generate it, however, dropping it would most likely break some existing code for no good reason.

cpp/src/slice2java/Gen.cpp
```diff
@@ -1555,18 +1555,24 @@ Slice::JavaVisitor::writeHiddenProxyDocComment(Output& out, const OperationPtr&
     // the _iceI_ methods are all async
 
     out << nl << "/**";
-    out << nl << " * @hidden";
+
+    out << nl << " * Invokes the " << p->name()
```
bernardnormier, line 1559: Did you try to generate instead `@SuppressWarnings("doclint:missing")` ?

It's Java 18, but maybe we could run Javadoc 18+ to generate the API reference?

https://bugs.openjdk.org/browse/JDK-8275351

cpp/src/slice2java/Gen.cpp
```diff
@@ -1900,7 +1917,12 @@ Slice::Gen::TypesVisitor::visitClassDefStart(const ClassDefPtr& p)
             //
             // Generate a constructor accepting parameters for all members.
             //
-            out << sp << nl << "public " << name << spar;
+            out << sp;
+            out << nl << "/**";
+            out << nl << " * Constructs a new {@code " << name << "} with all fields initialized.";
```
bernardnormier, line 1922: What about:
Creates a new NAME with values for all its fields.

cpp/src/slice2java/Gen.cpp
```diff
@@ -2112,22 +2145,23 @@ Slice::Gen::TypesVisitor::visitExceptionStart(const ExceptionPtr& p)
                     }
                 }
 
-                DataMemberList optionalMembers = p->orderedOptionalDataMembers();
-
                 // Generate a constructor accepting parameters for just the required members.
-                out << sp << nl << "public " << name << spar;
+                out << sp;
+                out << nl << "/**";
+                out << nl << " * Constructs a new {@code " << name << "} with required fields initialized.";
```
bernardnormier, line 2151: See earlier comment.

cpp/src/slice2java/Gen.cpp
```diff
@@ -2352,7 +2404,12 @@ Slice::Gen::TypesVisitor::visitStructEnd(const StructPtr& p)
             paramNames.push_back(memberName);
         }
 
-        out << sp << nl << "public " << name << spar << parameters << epar;
+        out << sp;
+        out << nl << "/**";
+        out << nl << " * Constructs a new {@code " << name << "} with required fiel.";
```
bernardnormier, line 2409: Typo

cpp/src/slice2java/Gen.cpp
```diff
@@ -2492,7 +2558,13 @@ Slice::Gen::TypesVisitor::visitStructEnd(const StructPtr& p)
     out << nl << "return c;";
     out << eb;
 
-    out << sp << nl << "public void ice_writeMembers(com.zeroc.Ice.OutputStream ostr)";
+    out << sp;
+    out << nl << "/**";
+    out << nl << " * Marshals this object's data members into an output stream.";
```
bernardnormier, line 2563: We should use 'field' for Java.

cpp/src/slice2java/Gen.cpp
```diff
@@ -1960,22 +1982,34 @@ Slice::Gen::TypesVisitor::visitClassDefEnd(const ClassDefPtr& p)
...
     out << eb;
 
     // Implementation of '_iceWriteImpl'.
     int iter = 0;
+
     out << sp;
-    writeHiddenDocComment(out);
+    writeDocComment(
+        out,
+        "{@inheritDoc}\n"
```
bernardnormier, line 2011: We need `@inheritdoc` + `@hidden` because just `@hidden` overrides the parent's doc-comment?

cpp/src/slice2java/Gen.cpp
externl: I didn't try it. While it might work we support Java 17 so it won't work for our users.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectionI.java
```diff
@@ -20,8 +20,9 @@
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
+// Public because it's used by the 'Ice/metrics' test.
 /**
- * @hidden Public because it's used by the 'Ice/metrics' test.
+ * @hidden
  */
```
InsertCreativityHere, line 26: For what it's worth, there's lots of these kinds of `@hidden` tags in our source code.

cpp/src/slice2java/Gen.cpp
```diff
@@ -1854,22 +1870,27 @@ Slice::Gen::TypesVisitor::visitClassDefStart(const ClassDefPtr& p)
...
                 // Generate a constructor accepting parameters for just the required members.
                 //
-                out << sp << nl << "public " << name << spar;
+                out << sp;
+                out << nl << "/**";
+                out << nl << " * Creates a new {@code " << name
+                    << "} with values for all fields not marked optional in the Slice definition for {@code "
+                    << p->scoped() << "}.";
+                out << nl << " *";
+                writeParamDocComments(out, requiredMembers);
```
InsertCreativityHere, line 1884: There's a couple things here:
1) If there are no parameters, we have an extra newline for no reason from the `nl << *`.
We should only add this spacing if there's going to be more tags after.

2) Maybe this is intentional, but `writeParamDocComments` only writes user-provided doc comments. So you can get things like this. So if a user doesn't document their members or something, you get this:
```
    /**
     * Creates a new {@code C} with values for all fields not marked optional in the Slice definition for {@code ::Test::C}.
     *
     */
    public C(int requiredA, int requiredB, String ss)
```
Maybe this is fine, but won't it complain about missing @param tags?

externl: Fixed.

externl:
> We're using a mix of `**/` and `*/` to end our generated doc-comments. We can wait to fix the ones we already have, but even in the new additions we're mixing them.
> 
> I personally prefer the `**/` for symmetry, but we should pick one.

yes, this is on my list. I prefer ` */`

cpp/src/slice2java/Gen.cpp
```diff
@@ -3676,10 +3844,24 @@ Slice::Gen::ServantVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
...
 
-        writeHiddenDocComment(out);
+        bool throwsUserException = !op->throws().empty() || op->hasMetadata("java:UserException");
+
+        out << nl << "/**";
+        out << nl << " * Dispatches the operation " << opName << ".";
+        out << nl << " *";
+        out << nl << " * @param obj the servant object";
+        out << nl << " * @param request the incoming request";
+        out << nl << " * @return a {@code CompletionStage} that will complete when the operation is done";
```
bernardnormier, line 3854: when the dispatch is done

bernardnormier (approved):

InsertCreativityHere (approved):
Much better!
I guess on to phase 2 now? : v)

Issue #3829, opened 2025-04-07
Use Helper Functions in `Slice/Grammar.y` Parser


InsertCreativityHere:
We should define and use helper functions to reduce code-duplication in the Bison grammar.

InsertCreativityHere:
This was implemented in #4009.

PR #4009, opened 2025-05-21
Cleanup Slice grammar and Improve Error Messages for Out-of-Bounds Constants

InsertCreativityHere:
This PR implements #3829, the final parser cleanup for 3.8. It eliminates lots of code duplication by adding some intermediate grammar rules, and using helper functions in the Bison code, instead of having the exact same code copied in multiple places.

Much of this code duplication was for resolving and checking integer constants.
While working on this, I also improved the error messages we emit for integers that are out of bounds; for tags, compact ids, and enumerator values (these must be between `0` and `int32::MAX` (inclusive)).
The fact I didn't need to update any compact-id tests afterwards  indicates that our testing of compact IDs is very lacking.


cpp/src/Slice/GrammarUtil.h
```diff
@@ -51,6 +51,8 @@ namespace Slice
 
     struct BoolTok final : public GrammarBase
     {
+        BoolTok(bool value) : v(value) {}
```
InsertCreativityHere, line 54: Added a constructor to use with `make_shared`.

cpp/src/Slice/Grammar.cpp
InsertCreativityHere, line 1: This file is generated by Bison.
Just ignore it, only `Grammar.y` needs to be reviewed.

cpp/test/Slice/errorDetection/EnumeratorValues.err
```diff
@@ -1,5 +1,5 @@
 EnumeratorValues.ice:9: value for enumerator 'C' is out of range
```
InsertCreativityHere, line 1: Now instead of just saying "out of range", we say what the value was, and specifically state why it's bad, so the user knows what the "allowed range" is from the message.

cpp/test/Slice/errorDetection/OptionalMembers.err
```diff
@@ -1,30 +1,26 @@
...
 OptionalMembers.ice:35: 'abc' is not defined
-OptionalMembers.ice:36: tag is out of range
-OptionalMembers.ice:37: tag is out of range
-OptionalMembers.ice:38: tag is out of range
+OptionalMembers.ice:36: invalid tag '0x80000000': tags must be less than or equal to '2147483647'
+OptionalMembers.ice:37: invalid tag '-0x80000001': tags cannot be negative
+OptionalMembers.ice:38: invalid tag '-1': tags cannot be negative
 OptionalMembers.ice:40: tag for optional data member 'm8' is already in use
-OptionalMembers.ice:41: tag is out of range
-OptionalMembers.ice:41: invalid tag 'C3'
```
InsertCreativityHere, line 9: The old setup also caused us to emit 2 redundant errors in some situations (see how they're both the same line).
Now we only emit 1 error for each bad value.

cpp/src/Slice/Grammar.y
```diff
@@ -584,94 +605,14 @@ type_id
 // ----------------------------------------------------------------------
 optional
 // ----------------------------------------------------------------------
-: ICE_OPTIONAL_OPEN ICE_INTEGER_LITERAL ')'
+: ICE_OPTIONAL_OPEN integer_constant ')'
```
InsertCreativityHere, line 608: We used to have 2 separate grammar rules for matching `ICE_INTEGER_LITERAL` and `scoped_name`.
We did this in multiple places, so I created an intermediate grammar rule `integer_constant` which handles both.

cpp/src/Slice/Grammar.y
```diff
@@ -584,94 +605,14 @@ type_id
...
-            }
-            tag = static_cast<int>(l);
-        }
-    }
-    else if (enumerator)
-    {
-        tag = enumerator->value();
-    }
 
-    if (tag < 0)
```
InsertCreativityHere, line 672: We also had duplicate code checking the bounds of integers, that I centralized in a helper function `checkIntegerBounds`.

cpp/src/Slice/Grammar.y
```diff
@@ -1189,30 +1032,24 @@ return_type
 ;
 
 // ----------------------------------------------------------------------
-operation_preamble
```
InsertCreativityHere, line 1192: We used to have duplicates of every rule, one normal, and one with `idempotent` in front.
I added an intermediate rule that matches an optional `idempotent` keyword to eliminate this duplication.

cpp/src/Slice/Grammar.y
```diff
@@ -116,6 +117,26 @@ slice_error(const char* s)
...
+    /// @param expectInterfaceType `true` is the type ends with '*' (indicated a proxy), false otherwise.
+    ///        This function will automatically emit errors for interfaces missing '*', or non-interfaces using '*'.
+    [[nodiscard]] TypePtr lookupTypeByName(const string& name, bool expectInterfaceType);
+
+    /// Resolves an `InterfaceDef` named `name` from the current scope and returns it.
+    /// If no such interface could be found, this returns `nullptr` instead.
+    /// @param name The (possibly scoped) name to resolve.
+    [[nodiscard]] InterfaceDefPtr lookupInterfaceByName(const string& name);
+
+    /// Checks if the provided integer is between '0' and 'int32_t::max' (inclusive).
```
externl, line 134: I would use interval notation. From your comment is it `[0, int32_t::max]` or `(0, int32_t::max]`

externl (approved):
Looks good to me. This stuff is hard to follow.

cpp/src/Slice/Grammar.y
InsertCreativityHere: Fair point, I guess this was ambiguous. I was meaning `[0, int32_t::max]` (fully inclusive).
I used the notation, and updated the comment to talk about the token, instead of the integer directly.

cpp/src/Slice/Grammar.y
```diff
@@ -2164,3 +1934,96 @@ keyword
 %%
 
 // NOLINTEND
+
+namespace
```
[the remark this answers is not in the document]
InsertCreativityHere: Why specifically? So I have some idea of where to draw the line.

I moved the interface lookup function into the parser, but the original helper function calls `checkIntroduced` which shouldn't be called after parsing. I'm not sure if I should add a doc-comment saying `don't call this function unless you're the parser!`, or if I should keep 2 helper functions. One in the parser with most of the meat, that is fine for anyone to call, and a (now much smaller) one in `Grammar.y` with the parser specific logic.

Also I don't think `checkIntegerBounds` can live anywhere else.
It's signature takes an `IntegerTokPtr`, and only the grammar/scanner use these `Tok` types.

cpp/src/Slice/Grammar.y
```diff
@@ -116,6 +117,26 @@ slice_error(const char* s)
     }
 }
 
+namespace
+{
+    /// Resolves a `Type` named `name` from the current scope and returns it.
+    /// If no such type could be found, this returns `nullptr` instead.
+    /// @param name The (possibly scoped) name to resolve.
+    /// @param expectInterfaceType `true` is the type ends with '*' (indicated a proxy), false otherwise.
```
pepone, line 125:
```suggestion
    /// @param expectInterfaceType `true` if the type ends with '*' (indicated a proxy), false otherwise.
```

cpp/src/Slice/Grammar.y
```diff
@@ -116,6 +117,26 @@ slice_error(const char* s)
...
+
+    /// Resolves an `InterfaceDef` named `name` from the current scope and returns it.
+    /// If no such interface could be found, this returns `nullptr` instead.
+    /// @param name The (possibly scoped) name to resolve.
+    [[nodiscard]] InterfaceDefPtr lookupInterfaceByName(const string& name);
+
+    /// Checks if the provided integer token's value is within the range ['0' ... 'int32_t::max'] (inclusive).
+    /// If it is within this range, this function will return `true`.
+    /// Otherwise thi will return `false`, and automatically emit an error stating the issue.
+    bool checkIntegerBounds(const IntegerTokPtr& token, string_view kindString);
```
pepone, line 137: Would be worth mention what kindString is for in the doc comment.

cpp/src/Slice/Grammar.y
```diff
@@ -584,94 +605,14 @@ type_id
...
-    else if (enumerator)
-    {
-        tag = enumerator->value();
-    }
 
-    if (tag < 0)
+    if (integer && checkIntegerBounds(integer, "tag"))
     {
-        currentUnit->error("invalid tag '" + scoped->v + "'");
+        tag = static_cast<int>(integer->v);
```
pepone, line 615: Shouldn't we use a fixed size type here instead of int?

cpp/src/Slice/Grammar.y
```diff
@@ -2164,3 +1934,77 @@ keyword
...
+        return firstType;
+    }
+
+    InterfaceDefPtr lookupInterfaceByName(const string& name)
+    {
+        ContainerPtr cont = currentUnit->currentContainer();
+        InterfaceDefPtr interfaceDef = cont->lookupInterfaceDef(name, true);
+        if (interfaceDef)
+        {
+            cont->checkIntroduced(name);
```
pepone, line 1972: Why is checkIntroduced required here? Seems of for a lookup method to do this.

pepone (approved):

PR #4014, opened 2025-05-22
Test implementation reuse in inheritance test (Swift)

bernardnormier:

PR #4015, opened 2025-05-23
Remove Disp structs in Swift, part 1

bernardnormier:
This is part 1 of the removal of the generated Disp structs in Swift.

## Background

Currently, for each Slice interface Name, we generate on the server-side:
 - protocol Name
 - extension Name
 - struct NameDisp that conforms to protocol Dispatcher

struct NameDisp holds onto a servant of type Name. NameDisp uses servant to implement Dispatcher.dispatch.

The reason for this complicated setup is our desire to support implementation reuse, as demonstrated by the filesystem demo. And, more importantly, this well-known Swift bug:
https://github.com/swiftlang/swift/issues/42725

prevents us from "overriding" protocol extension automatically.

## Proposed fix

The proposed fix, implemented by this PR, boils down to:

- while supporting implementation reuse is nice feature, it's a very rare use-case. So it's desirable to make the standard use-case (no impl reuse) simpler, while requiring extra work from the user in the reuse case.
- in 3.8, the dispatch method is now public / explained / documented, so we can user to override it in the advanced impl-reuse use-case

And that's exactly what this PR does:
- in the standard use-case, you just use the non-overridable dispatch implementation provided by the generated protocol extension
- in the impl-reuse use-case, you need to define an overridable dispatch instance method (or override this method from a base class), and provide a trivial implementation, namely:

```swift
func dispatch(_ request: Ice.IncomingRequest) async throws -> Ice.OutgoingResponse {
        // The implementation forwards to the static dispatch method defined on the generated protocol extension.
        try await Self.dispatch(self, request: request)
    }
```

See the Ice/inheritance test for an example.

Note:
- This PR includes #4014.
- This PR does not remove NameDisp, just stops using it in 3 tests. A follow-up PR will actually remove NameDisp.

### DIFF
--- a/cpp/src/slice2swift/Gen.cpp
+++ b/cpp/src/slice2swift/Gen.cpp
@@ -1325,18 +1325,20 @@ Gen::ServantVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
 
     out << sp;
     writeDocSummary(out, p);
-    out << nl << "public protocol " << servant;
-    if (!baseNames.empty())
+    out << nl << "public protocol " << servant << ":";
+    if (baseNames.empty())
     {
-        out << ":";
+        out << " " << getUnqualified("Ice.Dispatcher", swiftModule);
     }
-
-    for (auto i = baseNames.begin(); i != baseNames.end();)
+    else
     {
-        out << " " << (*i);
-        if (++i != baseNames.end())
+        for (auto i = baseNames.begin(); i != baseNames.end();)
         {
-            out << ",";
+            out << " " << (*i);
+            if (++i != baseNames.end())
+            {
+                out << ",";
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


swift/test/Ice/operations/Server.swift
```diff
@@ -29,7 +29,7 @@ class Server: TestHelperI {
             key: "TestAdapter.Endpoints", value: getTestEndpoint(num: 0))
         let adapter = try communicator.createObjectAdapter("TestAdapter")
         try adapter.add(
-            servant: MyDerivedClassDisp(MyDerivedClassI(self)), id: Ice.Identity(name: "test"))
+            servant: MyDerivedClassI(self), id: Ice.Identity(name: "test"))
```
bernardnormier, line 32: This test checks you can override Object operations by conforming to protocol Ice.Object, or overriding method(s) from a class that conforms to Ice.Object.

The same logic still works after this update.

cpp/src/slice2swift/Gen.cpp
```diff
@@ -1381,12 +1383,70 @@ bool
...
+
+    // TODO: doc-comment
+    out << sp;
+    out << nl << "public func dispatch(_ request: Ice.IncomingRequest) async throws -> Ice.OutgoingResponse" << sb;
+    out << nl << "try await Self.dispatch(self, request: request)";
+    out << eb;
+
+    // TODO: doc-comment
+    out << sp;
+    out << nl << "public static func dispatch(_ servant: " << servant
```
bernardnormier, line 1421: This is the existing dispatch logic, from NameDisp.

cpp/src/slice2swift/Gen.cpp
```diff
@@ -1381,12 +1383,70 @@ bool
...
+    const string unescapedName = removeEscaping(servant);
+    const string disp = unescapedName + "Disp";
+    const string traits = unescapedName + "Traits";
 
     out << sp;
     writeServantDocSummary(out, p, swiftModule);
-    out << nl << "extension " << getRelativeTypeString(p, swiftModule);
+    out << nl << "extension " << servant;
+    out << sb;
+    out << nl << "private static var defaultObject: " << getUnqualified("Ice.Object", swiftModule) << sb;
```
bernardnormier, line 1395: We do create a new `Ice.ObjectI<NameTraits>` for each dispatch. This is indeed a little wasteful... but if you call over and over the `ice_` operations, this wastefulness is the least of your problems! 

Note that a protocol extension cannot define a stored property.

pepone (approved):

externl (approved):

PR #4016, opened 2025-05-23
Remove Disp structs in Swift, part 2

bernardnormier:
This PR completes the removal of the Disp structs in Swift.


swift/src/Ice/Object.swift
```diff
@@ -107,27 +108,3 @@ open class ObjectI<T: SliceTraits>: Object {
         // Do nothing
     }
 }
-
-/// Request dispatcher for plain Object servants.
```
bernardnormier, line 111: With this PR, you can no longer easily instantiate a plain `Ice::Object` servant. I think that's ok.

externl (approved):

externl:
Looks good! Much nicer.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -675,7 +675,7 @@ SwiftGenerator::typeToString(const TypePtr& type, const ContainedPtr& usedBy, bo
         "Swift.Float",
         "Swift.Double",
         "Swift.String",
-        "Ice.Disp",      // Object
+        "Ice.Value",     // Object
```
pepone, line 678:
```suggestion
        "Ice.Value",     // Object not used anymore
```

pepone (approved):

cpp/src/slice2swift/SwiftUtil.cpp
bernardnormier: `Object` actually remains a valid syntax in Slice (it's an alias for `Value`).

But we apparently don't test it in Swift.

pepone: In slice2cs it has this "not used anymore" comment.

https://github.com/zeroc-ice/ice/blob/14159b77b284a769bc513ef792dbfd9798f3d410/cpp/src/slice2cs/CsUtil.cpp#L127

bernardnormier: You're right, it's no longer used because of the logic in this function. Fixed comment.
```cpp
    if (builtin)
    {
        if (builtin->kind() == Builtin::KindObject)
        {
            t = getUnqualified(builtinTable[Builtin::KindValue], currentModule);
        }
        else
        {
            t = getUnqualified(builtinTable[builtin->kind()], currentModule);
        }
    }
```

PR #4021, opened 2025-05-27
Make InvocationCanceledException package private.

bernardnormier:
The user can never catch this exception; however, we use it internally.


pepone (approved):
Should we also make `cancelable` method package internal?

```
public synchronized void cancelable(final CancellationHandler handler) {
        if (_cancellationException != null) {
            try {
                throw _cancellationException;
            } finally {
                _cancellationException = null;
            }
        }
        _cancellationHandler = handler;
    }
```

Not clear why this has to be public.

InsertCreativityHere (approved):

externl (approved):

bernardnormier:
Yes, we should make many of these methods (maybe all) package private.

PR #4032, opened 2025-05-29
Update JavaScript build instructions

pepone:
This PR:

* Updates the JavaScript build instructions.
* Improves the HTML page used for browser testing.
* Updates the JavaScript build dependencies.

### DIFF
--- a/js/BUILDING.md
+++ b/js/BUILDING.md
@@ -1,184 +1,121 @@
-# Ice for JavaScript Build Instructions
+# Building Ice for JavaScript from source
 
-This file describes how to build and install Ice for JavaScript from source code. If you prefer, you can also download
-[binary distributions] for the supported platforms.
+## Table of contents
 
-<!-- TOC depthFrom:2 -->
+- [Building Ice for JavaScript from source](#building-ice-for-javascript-from-source)
+  - [Table of contents](#table-of-contents)
+  - [Prerequisites](#prerequisites)
+  - [Build roadmap](#build-roadmap)
+  - [Building Ice for JavaScript](#building-ice-for-javascript)
+  - [Running the tests](#running-the-tests)
+  - [Creating the NPM package](#creating-the-npm-package)
+    - [Slice compilers](#slice-compilers)
+  - [Generating the API reference](#generating-the-api-reference)
 
-- [JavaScript Build Requirements](#javascript-build-requirements)
-- [Building the JavaScript libraries and NodeJS packages](#building-the-javascript-libraries-and-nodejs-packages)
-- [Running the JavaScript Tests](#running-the-javascript-tests)
-  - [Browser Information](#browser-information)
-    - [Self-Signed Certificate](#self-signed-certificate)
-    - [Secure WebSockets on iOS and Android](#secure-websockets-on-ios-and-android)
-      - [Installing Certificates on iOS](#installing-certificates-on-ios)
-      - [Installing Certificates on Android](#installing-certificates-on-android)
-- [Installing a Source Build](#installing-a-source-build)
+## Prerequisites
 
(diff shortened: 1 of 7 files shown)


js/BUILDING.md
```diff
@@ -1,184 +1,121 @@
...
-<!-- /TOC -->
+1. Node.js 22 or later.
 
-## JavaScript Build Requirements
+2. The Slice to JavaScript compiler (slice2js) from the C++ source distribution. \
+   Refer to the [build instructions](../cpp/BUILDING.md) in the `cpp` folder for details on how to build the C++ source
+   distribution.
 
-To build Ice for JavaScript you must have the following:
+3. Python 3.12 is required for running the tests. The Glacier2 test also require the `passlib` Python package.
```
bernardnormier, line 23: requires the

js/BUILDING.md
```diff
@@ -1,184 +1,121 @@
...
 
-- The `slice2js` compiler from Ice for C++. If you have not built Ice for C++ in this source distribution, refer to the
-  [C++ build instructions](../cpp/BUILDING.md).
-- Node.js 22 or later
+4. Ice for Python is required for running the JavaScript in a web browser.\
```
bernardnormier, line 25: the JavaScript tests?

js/BUILDING.md
```diff
@@ -1,184 +1,121 @@
...
+   Refer to the [build instructions](../python/BUILDING.md) in the `python` folder for details on how to build the
```
bernardnormier, line 26: Probably not a good advice as you'd typically use the Python wheel (possibly nightly wheel).

js/BUILDING.md
```diff
@@ -1,184 +1,121 @@
...
+using the following command:
 
 ```shell
 npm install
 npm run build
 ```
 
-On Windows, you need to set the platform and configuration in order to locate `slice2js`. For example, if you have
-built C++ with the x64 platform and the Release configuration, you can use the following command to build JavaScript:
+> On Windows, if you’re not using the default C++ build configuration (e.g., x64/Release), you need to ensure that the
```
bernardnormier, line 49: e.g. should be i.e. here?

bernardnormier (approved):

js/BUILDING.md
pepone: Should we mention how to install it, or is enough to say it is required?

js/BUILDING.md
externl, line 25:
```suggestion
4. Ice for Python is required for running the JavaScript tests in a web browser.\
```

js/BUILDING.md
externl: I think if a user is doing a source build they should just do a source build for everything.

js/BUILDING.md
```diff
@@ -1,184 +1,121 @@
...
-Next you must install the certificate authority on your device. The simplest way
-is to email the CA certificate (`certs/cacert.pem`) to yourself and then follow
-the instructions below to install the certificate on your device.
-
-Once installed, you connect to the server using the same IP address used to
-create the certificates.
+### Slice compilers
 
-##### Installing Certificates on iOS
+By default, the NPM package `@zeroc\ice` includes only the `slice2js` compiler created by the local C++ build.
```
externl, line 96:
```suggestion
By default, the NPM package `@zeroc/ice` includes only the `slice2js` compiler created by the local C++ build.
```

externl (approved):

js/BUILDING.md
pepone: Seems there is a problem with GitHub, and it is not displaying the latest commits in the PR.

externl: Weird!

PR #4037, opened 2025-05-30
Building cpp

pepone:
This PR updates the C++ build instructions for 3.8

### DIFF
--- a/cpp/BUILDING.md
+++ b/cpp/BUILDING.md
@@ -48,16 +48,24 @@ flowchart LR
 
 #### On Ubuntu and other Debian-based systems
 
+Install the required third-party libraries:
+
 ```shell
-sudo apt-get install libedit-dev libexpat1-dev liblmdb-dev libmcpp-dev \
-    libssl-dev libsystemd-dev
+sudo apt-get install libedit-dev libexpat1-dev liblmdb-dev libmcpp-dev libssl-dev libsystemd-dev
 ```
 
 #### On RHEL 9
 
+Add the ZeroC repository:
+
 ```shell
 dnf install https://zeroc.com/download/ice/3.7/el9/ice-repo-3.7.el9.noarch.rpm
-dnf install mcpp-devel bzip2-devel openssl-devel expat-devel libedit-devel lmdb-devel libsystemd-devel
+```
+
+Install the required third-party libraries:
+
+```shell
+dnf install bzip2-devel expat-devel libedit-devel libsystemd-devel lmdb-devel mcpp-devel openssl-devel
 ```
 
 ### Building


cpp/BUILDING.md
```diff
@@ -1,345 +1,277 @@
...
+- A C++ compiler with support for the C++17 standard.
+  - GCC on Linux
+  - Clang on macOS
+  - Visual Studio 2022 on Windows
+- The following third-party libraries:
+  - bzip2 1.0, used for protocol compression
+  - expat 2.1 or higher, used by IceGrid
+  - libedit, required for the command-line utilities on Linux and macOS
+  - LMDB 0.9, used for the IceStorm and IceGrid databases
+  - mcpp 2.7.2 with ZeroC patches, used by the Slice compilers
```
bernardnormier: I find this a bit confusing, since we often use "system-provided" mcpp.

Maybe we don't need to/shouldn't list these dependencies as "prerequisites", since we explain how to install these dependencies below?

cpp/BUILDING.md
```diff
@@ -1,345 +1,277 @@
...
+  - D-Bus on Linux, required for the IceBT Bluetooth transport
+  - BlueZ on Linux, required for the IceBT Bluetooth transport
+
+## Building Ice for C++ on Linux
+
+### Installing Build Dependencies
+
+#### On Ubuntu and other Debian-based systems
+
+You can install the required third-party libraries with:
```
bernardnormier: I would shorten this, and show just the command.

cpp/BUILDING.md
```diff
@@ -1,345 +1,277 @@
...
-sudo dnf install lmdb-devel mcpp-devel
+sudo apt-get install libbluetooth-dev libbz2-dev libdbus-1-dev libedit-dev libexpat1-dev liblmdb-dev libmcpp-dev libssl-dev libsystemd-dev
 ```
 
-##### Ice for Bluetooth
+#### On RHEL 9
 
-In addition, on Ubuntu and Debian distributions where the Ice for Bluetooth plug-in is supported, you need to install
-the following packages in order to build the IceBT transport plug-in:
+First, install the ZeroC RHEL 9 repository (which provides the patched `mcpp` package):
```
bernardnormier: Same here, just show the two commands.

cpp/BUILDING.md
```diff
@@ -1,345 +1,277 @@
...
 
 ```shell
-sudo apt-get install pkg-config libdbus-1-dev libbluetooth-dev
+dnf install mcpp-devel bzip2-devel openssl-devel expat-devel libedit-devel lmdb-devel libsystemd-devel
 ```
 
-> _We have experienced problems with BlueZ versions up to and including 5.39, as
-> well as 5.44 and 5.45. At this time we recommend using the daemon (`bluetoothd`)
-> from BlueZ 5.43._
+#### On Other Linux Distributions
```
bernardnormier: I would remove that. Doesn't provide any useful info.

cpp/BUILDING.md
```diff
@@ -1,345 +1,277 @@
...
 
-bzip, expat and libedit are included with your system.
+### Building
 
-You can install LMDB and mcpp using Homebrew:
+Once you have installed the required libraries, you can build Ice for C++ by running:
 
 ```shell
-brew install lmdb mcpp
+make
```
bernardnormier: Maybe show `make -j10`?

Can't show just `-j` as it's infinite and causes the build to hang.

cpp/BUILDING.md
```diff
@@ -1,345 +1,277 @@
...
+### Testing
 
-The C++ source tree supports multiple build configurations and platforms. To see the supported configurations and
-platforms:
+You can run the tests with:
 
 ```shell
-make print V=supported-configs
-make print V=supported-platforms
+python3 allTests.py --all
```
bernardnormier: We should standardize on plain `python`.

cpp/BUILDING.md
```diff
@@ -1,345 +1,204 @@
...
 
-This builds the Ice for C++ SDK and the Ice for C++ test suite, with release binaries for the default platform. The
-default platform depends on the Visual Studio Command Prompt you use.
-
-Set the MSBuild `Configuration` property to `Debug` to build debug binaries instead:
+By default, Ice for C++ is installed to `/opt/Ice-3.8a0`. To change the installation location, set the `prefix` Makefile variable:
 
 ```shell
-msbuild /m msbuild\ice.proj /p:Configuration=Debug
+make install prefix=~/Ice
```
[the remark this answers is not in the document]
bernardnormier: Maybe use `$HOME/ice` instead of `~`? I would also use lowercase for the Ice homedir.

cpp/BUILDING.md
```diff
@@ -1,345 +1,277 @@
...
 
-```shell
-msbuild /m msbuild\ice.proj
-```
+On macOS, you can build Ice for C++ for macOS, iOS devices, and iOS simulators using the `macosx`, `iphoneos`, and
+`iphonesimulator` platforms, respectively.
 
-This builds the Ice for C++ SDK and the Ice for C++ test suite, with release binaries for the default platform. The
-default platform depends on the Visual Studio Command Prompt you use.
+By default, the build produces binaries for the `macosx` platform. You can build for multiple platforms by listing them
```
bernardnormier: It's called `macos`, isn't it?

cpp/BUILDING.md
```diff
@@ -1,345 +1,277 @@
...
+in the `PLATFORMS` Make variable, or use `all` to build all supported platforms at once.
 
-Set the MSBuild `Configuration` property to `Debug` to build debug binaries instead:
+There are two build configurations: `shared` and `static`, which produce shared and static libraries, respectively.
```
bernardnormier: My understanding is this is no longer true. `static` makes a difference only for the tests.

pepone: I didn't knew that.

cpp/BUILDING.md
pepone: Removed

scripts/Util.py
```diff
@@ -3550,6 +3550,8 @@ def __init__(self, options=[]):
             if self.buildConfig == platform.getDefaultBuildConfig():
                 if isinstance(platform, Windows):
                     self.buildConfig = "Release"
+                elif self.buildPlatform in ["iphoneos", "iphonesimulator"]:
+                    self.buildConfig = "release"
```
pepone, line 3554: I updated this to match the default build configuration.

cpp/BUILDING.md
```diff
@@ -1,345 +1,273 @@
...
-install these packages as shown below:
+#### On Ubuntu and other Debian-based systems
 
-##### RHEL 9
+Install the required third-party libraries:
 
 ```shell
-sudo dnf install https://zeroc.com/download/ice/3.7/el9/ice-repo-3.7.el9.noarch.rpm
-sudo dnf install lmdb-devel mcpp-devel
+sudo apt-get install libedit-dev libexpat1-dev liblmdb-dev libmcpp-dev libssl-dev libsystemd-dev
```
externl, line 54: Should we include bluetooth?

cpp/BUILDING.md
```diff
@@ -1,345 +1,273 @@
...
-- [pkg-config][9] 0.29 or later
-- [D-Bus][10] 1.10 or later
-- [BlueZ][11] 5.37 or later
-
-These packages are provided with the system and can be installed with:
+Add the ZeroC repository:
 
 ```shell
-sudo apt-get install pkg-config libdbus-1-dev libbluetooth-dev
+dnf install https://zeroc.com/download/ice/3.7/el9/ice-repo-3.7.el9.noarch.rpm
```
externl, line 62: What do we need for from this repo?

cpp/BUILDING.md
```diff
@@ -1,345 +1,273 @@
...
 ```
 
-Run `make` to build the Ice C++ libraries, services and test suite. Set `V=1` to get a more detailed build output. You
-can build only the libraries and services with the `srcs` target, or only the tests with the `tests` target. For
-example:
+Or build both shared and static libraries by running:
 
 ```shell
-make V=1 -j8 srcs
+make -j10 CONFIGS=all
```
externl, line 90: I think we should remove -j10, it's very arbitrary. Anyone who was used make knows they have do `-jXXX`.

cpp/BUILDING.md
```diff
@@ -1,345 +1,273 @@
...
-Using the first Command Prompt produces `Win32` binaries by default, while the second Command Prompt produces `x64`
-binaries by default.
+### Installing Build Dependencies
 
-In the Command Prompt, change to the `cpp` subdirectory:
+You can install the required third-party libraries using brew
 
 ```shell
-cd cpp
+brew install mcpp lmdb
```
externl, line 126: We should consider adding a `Brewfile` to the root of the ice repo.

cpp/BUILDING.md
```diff
@@ -1,345 +1,273 @@
...
-## Cleaning the source build on Linux or macOS
+### Installation
 
-Running `make clean` will remove the binaries created for the default configuration and platform.
+You can install Ice for C++ by running:
 
-To clean the binaries produced for a specific configuration or platform, you need to specify the `CONFIGS` or
-`PLATFORMS` variable. For example, `make CONFIGS=static clean` will clean the static configuration build.
+```shell
+make -j10 install
```
bernardnormier, line 210: Should we show `-j10` for the install target? I assume the installation doesn't actually run in parallel.

cpp/BUILDING.md
```diff
@@ -1,345 +1,273 @@
...
+```
 
-To clean the build for all the supported configurations and platforms, run `make CONFIGS=all PLATFORMS=all clean`.
+By default, Ice for C++ is installed to `/opt/Ice-3.8a0`. To change the installation location, set the `prefix` Makefile
+variable:
 
-Running `make distclean` will also clean the build for all the configurations and platforms. In addition, it will also
-remove the generated files created by the Slice compilers.
+```shell
+make -j10 install prefix=$HOME/ice
```
bernardnormier, line 217: Same here

bernardnormier (approved):

cpp/BUILDING.md
externl: I don't think we should show it at all.

cpp/BUILDING.md
bernardnormier: I think developers just copy the instructions as-is and will end up with very long builds if we don't include a `-j`.

externl: I think it's okay to put some onous the user.

externl: For example see https://github.com/git/git/blob/master/INSTALL

cpp/BUILDING.md
pepone: MCPP if I'm not mistaken.

externl: That's unfortunate. IIRC it's the only platform that requires our own mcpp.

externl (approved):

cpp/BUILDING.md
bernardnormier: If we need this repo, we should upgrade it to be "3.8".

pepone: I will check if it is still required

cpp/BUILDING.md
pepone: We only ever include it with Ubuntu/Debian, we should test if we can include it here too.

cpp/BUILDING.md
pepone: It is required for both lmdb-devel mcpp-devel

pepone: I will upgrade the URL once we have the new repository.

PR #4038, opened 2025-06-01
Upgrade Swift code base to Swift 6

bernardnormier:
This is an initial, incomplete upgrade of the code base to Swift 6 with strict concurrency.

It's based on externl/swift-structured-concurrency.

### DIFF
--- a/swift/Plugins/CompileSlice/CompileSlicePlugin.swift
+++ b/swift/Plugins/CompileSlice/CompileSlicePlugin.swift
@@ -1,7 +1,6 @@
 // Copyright (c) ZeroC, Inc.
 
 import Foundation
-
 @preconcurrency import PackagePlugin
 
 enum PluginError: Error {
--- a/swift/Plugins/CompileSlice/CompileSlicePlugin.swift
+++ b/swift/Plugins/CompileSlice/CompileSlicePlugin.swift
@@ -1,7 +1,8 @@
 // Copyright (c) ZeroC, Inc.
 
 import Foundation
-import PackagePlugin
+
+@preconcurrency import PackagePlugin
 
 enum PluginError: Error {
     case invalidTarget(Target)
--- a/swift/src/Ice/Communicator.swift
+++ b/swift/src/Ice/Communicator.swift
@@ -4,7 +4,7 @@ import Foundation
 
 /// The central object in Ice. One or more communicators can be instantiated for an Ice application. Communicator
 /// instantiation is language-specific, and not specified in Slice code.
-public protocol Communicator: AnyObject {
+public protocol Communicator: AnyObject, Sendable {
     /// Destroy the communicator. This operation calls shutdown implicitly. Calling destroy cleans up
     /// memory, and shuts down this communicator's client functionality and destroys all object adapters. Subsequent
     /// calls to destroy are ignored.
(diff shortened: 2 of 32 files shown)


swift/Plugins/CompileSlice/CompileSlicePlugin.swift
```diff
@@ -1,7 +1,8 @@
 // Copyright (c) ZeroC, Inc.
 
 import Foundation
-import PackagePlugin
+
+@preconcurrency import PackagePlugin
```
bernardnormier, line 4: This code produces many deprecated warnings that I have not fixed.

swift/src/Ice/Communicator.swift
```diff
@@ -210,7 +210,7 @@ public protocol Communicator: AnyObject {
     /// - parameter servant: `Dispatcher` The servant that implements the new Admin facet.
     ///
     /// - parameter facet: `String` The name of the new Admin facet.
-    func addAdminFacet(servant: Dispatcher, facet: String) throws
+    func addAdminFacet(servant: Dispatcher & Sendable, facet: String) throws
```
bernardnormier, line 213: We should actually make Dispatcher sendable.

In Ice, any Dispatcher registered with an object adapter can be called concurrently and therefore should be sendable.

swift/src/Ice/Dispatcher.swift
```diff
@@ -5,7 +5,7 @@ public protocol Dispatcher {
     /// Dispatches an incoming request and returns the corresponding outgoing response.
     /// - Parameter request: The incoming request.
     /// - Returns: The outgoing response.
-    func dispatch(_ request: IncomingRequest) async throws -> OutgoingResponse
+    func dispatch(_ request: sending IncomingRequest) async throws -> OutgoingResponse
```
bernardnormier, line 8: Should probably be `-> sending OutgoingResponse`.

swift/src/Ice/EncodingVersionExtensions.swift
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) ZeroC, Inc.
+
+import Foundation
+
+extension EncodingVersion {
+    /// Returns the 1.0 encoding version.
+    static public var Encoding_1_0 : EncodingVersion {
```
bernardnormier: This "work-around" is not source-compatible as far as I can tell.

swift/src/Ice/Initialize.swift
```diff
@@ -267,10 +267,6 @@ public let intVersion: Int = 30850
 /// B indicates the minor version, and C indicates the patch level.
 public let stringVersion: String = "3.8.0-alpha.0"
 
-public let Encoding_1_0 = EncodingVersion(major: 1, minor: 0)
```
bernardnormier, line 270: I think we should just keep these rarely used global "constants" and mark then as `nonisolated(unsafe)`

swift/src/Ice/Proxy.swift
```diff
@@ -1024,7 +1024,7 @@ open class ObjectPrxI: ObjectPrx {
         mode: OperationMode,
         format: FormatType? = nil,
         write: ((OutputStream) -> Void)? = nil,
-        read: @escaping (InputStream) throws -> T,
+        read: @escaping (InputStream) throws -> sending T,
```
bernardnormier, line 1027: The change that makes `continuation.resume(returning: returnValue)` work.

swift/src/IceImpl/include/Config.h
```diff
@@ -10,3 +10,5 @@
 #endif
 
 #define ICE_SWIFT_UNAVAILABLE(msg) __attribute__((unavailable(msg)))
+
+#define SWIFT_SENDING __attribute__((swift_attr("sending")))
```
bernardnormier, line 14: Requires Swift 6. Should this be ICE_SWIFT_SENDING?

PR #4049, opened 2025-06-03
Support Swift 6 Structured Concurrency

externl:
This PR updates Ice for Swift to be compatible with Swift 6's Structured Concurrency requirements.


swift/Plugins/CompileSlice/CompileSlicePlugin.swift
```diff
@@ -55,39 +55,56 @@ struct CompileSlicePlugin: BuildToolPlugin {
 
         let fm = FileManager.default
 
+        // target.directory is deprecated and has been replaced by target.directoryURL. Unfortunately, it seems that the developers forgot to add this
```
bernardnormier: line too long

bernardnormier: I would add a TODO. Presumably, they will fix it one day.

swift/src/Ice/Communicator.swift
```diff
@@ -210,7 +210,7 @@ public protocol Communicator: AnyObject {
     /// - parameter servant: `Dispatcher` The servant that implements the new Admin facet.
     ///
     /// - parameter facet: `String` The name of the new Admin facet.
-    func addAdminFacet(servant: Dispatcher, facet: String) throws
+    func addAdminFacet(servant: Dispatcher & Sendable, facet: String) throws
```
bernardnormier, line 213: I would add a "TODO: make Dispatcher sendable"

swift/src/Ice/CurrentExtensions.swift
```diff
@@ -169,7 +169,7 @@ extension Current {
     }
 }
 
-private let currentProtocolEncoding = EncodingVersion(major: 1, minor: 0)
+nonisolated(unsafe) private let currentProtocolEncoding = EncodingVersion(major: 1, minor: 0)
```
bernardnormier, line 172: Is this really necessary for a private let constant?

swift/src/Ice/AdminFacetFactory.swift
```diff
@@ -4,9 +4,9 @@ import IceImpl
 
 class AdminFacetFacade: ICEDispatchAdapter {
     private let communicator: Communicator
-    var servant: Dispatcher
+    var servant: Dispatcher & Sendable
```
bernardnormier: Why is servant an internal var?

swift/src/Ice/InputStream.swift
```diff
@@ -795,18 +795,18 @@ extension InputStream {
...
     ///
     /// - returns: `ObjectPrx?` - The proxy read from the stream.
-    public func read(tag: Int32, type _: ObjectPrx.Protocol) throws -> ObjectPrx? {
+    public func read(tag: Int32, type _: ObjectPrx.Protocol) throws -> sending ObjectPrx? {
         return try read(tag: tag) as ObjectPrxI?
     }
 
     /// Reads a value from the stream.
-    public func read(cb: @escaping (Value?) throws -> Void) throws {
+    public func read(cb: @Sendable @escaping (Value?) throws -> Void) throws {
```
bernardnormier, line 808: We should add a comment on the two "read value" function, something like:
```swift
// - Parameter cb: The closure that sets the value in the caller. This closure always executed in the same thread as the caller.
// TODO: We marked cb as @Sendable to work-around Swift 6 concurrency check errors.
```

swift/test/Ice/location/AllTests.swift
```diff
@@ -6,6 +6,7 @@ import TestCommon
 
 func allTests(_ helper: TestHelper) async throws {
     // This function must be sendable as it can be concurrently executed by a TaskGroup.
+    nonisolated(unsafe) let helper = helper
```
bernardnormier: It sounds like TestHelper should be sendable.

swift/test/Ice/location/AllTests.swift
```diff
@@ -268,7 +269,7 @@ func allTests(_ helper: TestHelper) async throws {
 
     try await withThrowingTaskGroup(of: Void.self) { [hello] taskGroup in
         for _ in 0..<1000 {
-            taskGroup.addTask {
+            taskGroup.addTask { @Sendable in
```
bernardnormier: Not clear why you need this `@Sendable`.

bernardnormier (approved):

swift/src/Ice/CurrentExtensions.swift
externl: Yep. The compiler complains.

swift/src/Ice/AdminFacetFactory.swift
externl: It should be `let`. It's internal as we access it from outside the class in CommunicatorI.

swift/test/Ice/location/AllTests.swift
externl: Yes

swift/test/Ice/location/AllTests.swift
externl: It's because the closure here is sending. 

```
277 |             taskGroup.addTask {
    |                               `- error: passing closure as a 'sending' parameter risks causing data races between code in the current task and concurrent execution of the closure
278 |                 do {
279 |                     try await hello.sayHello()
280 |                     try test(false)
    |                         `- note: closure captures 'test(_:file:line:)' which is accessible to code in the current task
281 |                 } catch is Ice.NotRegisteredException {}
282 |             }
```

externl: I could also fix by making a copy of test, but I think making the closure sendable is simpler here.

bernardnormier: Making TestHelper sendable doesn't fix it?

externl: It does! Fixed!

PR #4050, opened 2025-06-03
Improve Java Linting

InsertCreativityHere:
This PR improves our checkstyle/rewrite setup to match the changes I'm making to the ice-demos.
I added comments to each of the changes. They're all too different to talk about here.

### DIFF
--- a/.github/workflows/java.yml
+++ b/.github/workflows/java.yml
@@ -8,7 +8,7 @@ on:
     branches: ["main"]
 
 jobs:
-  rewriteCatch:
+  checkstyle:
     runs-on: ubuntu-24.04
     steps:
       - name: Checkout repository
@@ -23,6 +23,25 @@ jobs:
       - name: Build slice2java
         run: make -C cpp slice2java
 
-      - name: Run rewriteDryRun with build failure enabled
+      - name: Run checkstyle formatter
+        working-directory: java
+        run: ./gradlew check
+
+  rewrite:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup C++
+        uses: ./.github/actions/setup-cpp
+
+      - name: Setup Java
+        uses: ./.github/actions/setup-java
+
+      - name: Build slice2java
+        run: make -C cpp slice2java
+
+      - name: Run rewrite linter
         working-directory: java
         run: ./gradlew rewriteDryRun
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -124,6 +124,7 @@ dependencies {
 }
 
 // Causes build failure when rewriteRun would make changes.
+def runningInCi = providers.environmentVariable("CI").isPresent()
 tasks.named('rewriteDryRun') {
-    rewrite.failOnDryRunResults = true
+    rewrite.failOnDryRunResults = runningInCi
 }
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -3,7 +3,7 @@
 plugins {
     id 'com.zeroc.ice.slice-tools' apply false
     id 'checkstyle'
-    id 'org.openrewrite.rewrite' version '7.3.0'
+    id 'org.openrewrite.rewrite' version '7.7.0'
 }
 
 allprojects {
--- a/java/build.gradle
+++ b/java/build.gradle
@@ -45,14 +45,6 @@ subprojects {
         options.encoding = "UTF-8"
         options.deprecation = true
     }
-
-    // We want Checkstyle to generally run before Rewrite and to exclude files generated from Slice.
-    tasks.withType(Checkstyle) {
-        task ->
-            rewriteDryRun.dependsOn(task)
-            rewriteRun.dependsOn(task)
-    }
-
 }
 
 task clean(type: Delete) {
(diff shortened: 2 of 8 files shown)


java/build.gradle
```diff
@@ -45,14 +45,6 @@ subprojects {
...
         options.deprecation = true
     }
-
-    // We want Checkstyle to generally run before Rewrite and to exclude files generated from Slice.
-    tasks.withType(Checkstyle) {
-        task ->
-            rewriteDryRun.dependsOn(task)
-            rewriteRun.dependsOn(task)
-    }
-
```
InsertCreativityHere, line 55: 1) This is only necessary when you rewrite the code automatically in CI. Something we never do.
2) Even then, this order is wrong. You'd want `checkstyle` to run _after_ `rewrite` so it checks the rewritten code.

java/build.gradle
```diff
@@ -102,21 +94,9 @@ checkstyle {
...
-        'org.openrewrite.java.RemoveUnusedImports',
-        'org.openrewrite.java.format.NormalizeFormat',
-        'org.openrewrite.java.format.BlankLines',
-        'org.openrewrite.java.format.Spaces',
-        'org.openrewrite.java.format.NormalizeTabsOrSpaces',
-        'org.openrewrite.java.format.TabsAndIndents',
-        'org.openrewrite.java.format.NormalizeLineBreaks',
-        'org.openrewrite.java.format.RemoveTrailingWhitespace',
-        'org.openrewrite.staticanalysis.CodeCleanup'
-    )
```
InsertCreativityHere, line 117: These were moved into the `rewrite.yml` file.
A) It's better to declaratively add them there, then programmatically add them here.
B) It lets us share this list with `slice-tools`, instead of needing to repeat it in every project's config.

java/build.gradle
```diff
@@ -102,21 +94,9 @@ checkstyle {
...
-    checkstyleConfigFile = file("config/checkstyle/checkstyle.xml")
```
InsertCreativityHere, line 118: `checkstyleConfigFile` isn't necessary if the `checkstyle` plugin is active.
`rewrite` can coordinate with that plugin to get the config.

java/build.gradle
```diff
@@ -144,6 +124,7 @@ dependencies {
 }
 
 // Causes build failure when rewriteRun would make changes.
+def runningInCi = providers.environmentVariable("CI").isPresent()
 tasks.named('rewriteDryRun') {
-    rewrite.failOnDryRunResults = true
+    rewrite.failOnDryRunResults = runningInCi
```
InsertCreativityHere, line 129: A) I doubt anyone ever runs `./gradlew rewriteDryRun` manually
B) I made this change in the demos. Without it I would only see the warnings for one project at a time. Which is pretty slow. By not having it fail, it'll show you all the warnings at once. Much easier to fix.

Note that `false` is the default. So logically, we're just making it more strict for CI, and using the default otherwise.

java/gradle/ice.gradle
```diff
@@ -28,22 +30,11 @@ slice {
...
-                } else if (path.endsWith("/com/zeroc/Ice/PropertyNames.java")) {
-                    return true
-                }
-                return false
-            }
-        }
+    // If we're running in CI, we want the build to fail if any warnings are emitted.
+    // This doesn't affect what checkstyle emits, just whether it 'fails' or 'succeeds' from Gradle's perspective.
+    if (runningInCi) {
+        maxWarnings = 0
```
InsertCreativityHere, line 37: Without this, `checkstyle` will always show as "succeeded", even if there were formatting errors.
This setting ensures CI will fail if there are formatting errors.

java/gradle/ice.gradle
```diff
@@ -28,22 +30,11 @@ slice {
...
-            exclude {
-                def path = it.file.toPath().normalize().toString().replace(File.separator, '/')
-                if (path.contains("/build/generated")) {
-                    return true
-                } else if (path.endsWith("/com/zeroc/Ice/PropertyNames.java")) {
-                    return true
-                }
-                return false
-            }
-        }
```
InsertCreativityHere, line 46: This is now handled by the `config/checkstyle/suppressions.xml` file.

java/tools/slice-tools/build.gradle.kts
```diff
@@ -92,8 +94,22 @@ tasks.named<Test>("test") {
     useJUnitPlatform()
 }
```
InsertCreativityHere, line 96: I just updated `slice-tools` to be on-par with the `Ice` and `ice-demos` build setups.

java/rewrite.yml
```diff
@@ -1,5 +1,43 @@
...
+  - org.openrewrite.staticanalysis.ForLoopControlVariablePostfixOperators
+  - org.openrewrite.staticanalysis.HideUtilityClassConstructor
+  - org.openrewrite.staticanalysis.NeedBraces
+  - org.openrewrite.staticanalysis.OperatorWrap
+  - org.openrewrite.staticanalysis.RemoveCallsToObjectFinalize
+  - org.openrewrite.staticanalysis.ReplaceStringBuilderWithString
+  - org.openrewrite.staticanalysis.ReplaceThreadRunWithThreadStart
+  - org.openrewrite.staticanalysis.TypecastParenPad
+  - org.openrewrite.staticanalysis.UnnecessaryParentheses
+  # - org.openrewrite.java.OrderImports TODO enable this in a separate PR
```
InsertCreativityHere, line 21: We only had `OrderImports` set for the `slice-tools`, but not for Ice.
Seems like a useful thing to have, but it changed lots of lines.
So, after this PR I'll enable this option, and make the necessary changes.

.github/workflows/java.yml
```diff
@@ -8,7 +8,7 @@ on:
     branches: ["main"]
 
 jobs:
-  rewriteCatch:
+  checkstyle:
```
InsertCreativityHere, line 11: `checkstyle` and `rewrite` are totally unrelated tools. `rewrite` has some awareness of `checkstyle`, but only some.
Currently, our `java` workflow only runs `rewriteDryRun`. This PR updates it to also run `checkstyle`.

PR #4052, opened 2025-06-04
Update 3.8 Changelog to Mention Important Slice Changes

InsertCreativityHere:
I believe this covers all the important changes to Slice. I still plan to go through the commit history and see if anything else pops out at me. But that sounds pretty tedious, so I haven't done it yet.

### DIFF
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -274,8 +274,8 @@ classDiagram
   Slice loader implemented by Ice, and you don't need to do anything to help Ice locate these generated classes.
   However, in Java and MATLAB, there is no such registration at startup, and you need to help Ice locate these generated
   classes when:
-  - you remap either the class name or an enclosing module using the java:identifier, java:package, or
-    matlab:identifier metadata; or
+  - you remap either the class name or an enclosing module using the `["java:identifier"]`, `["java:package"]`, or
+    `["matlab:identifier"]` metadata; or
   - you assign a compact ID to your class
 
   You help Ice locate these classes by installing a Slice loader in `InitializationData`, just like when you provide a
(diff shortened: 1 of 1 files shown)


CHANGELOG-3.8.md
```diff
@@ -274,8 +274,8 @@ classDiagram
   Slice loader implemented by Ice, and you don't need to do anything to help Ice locate these generated classes.
   However, in Java and MATLAB, there is no such registration at startup, and you need to help Ice locate these generated
   classes when:
-  - you remap either the class name or an enclosing module using the java:identifier, java:package, or
-    matlab:identifier metadata; or
+  - you remap either the class name or an enclosing module using the `["java:identifier"]`, `["java:package"]`, or
```
bernardnormier: `["java:identifier"]` is not a valid metadata directive. I prefer just `java:identifier`.

CHANGELOG-3.8.md
```diff
@@ -311,26 +311,92 @@ classDiagram
 
 - Removed local Slice. `local` is no longer a Slice keyword.
 
+- Added new metadata for customizing the mapped names of Slice definitions in each language.
+  This metadata is of the form: `["<lang>:<identifier>"]`, where `<lang>` can be any of the standard language prefixes,
```
bernardnormier: You meant `["<lang>:identifier:<identifier>"]`

CHANGELOG-3.8.md
```diff
@@ -311,26 +311,92 @@ classDiagram
...
+  and that definition's identifier will be `<identifier>` in the specified language, with no additional processing.
+
+  For example:
+
+  ```slice
+  ["cs:identifier:MyNamespace"]
+  ["java:identifier:com.example.my_package"]
```
bernardnormier: I would avoid `_` in sample Java package names.

CHANGELOG-3.8.md
```diff
@@ -311,26 +311,92 @@ classDiagram
...
+
+  The argument is used as a drop-in replacement for the Slice identifier, with no additional processing.
+  With the above example, `slice2cs` will generate `namespace MyNamespace {}` and `slice2java` will generate
+  `package com.example.my_package;`. All other compilers will map the module using its Slice-provided identifier,
+  as usual.
+
+  This metadata can be applied to any Slice definitions with an identifier, and is available for all languages.
+
+  Note that this only affects the _mapped_ name of Slice definitions.
+  It has no effect on type-ids, or a definition's on-the-wire representation.
```
bernardnormier: type-ids => Slice type IDs

CHANGELOG-3.8.md
```diff
@@ -311,26 +311,92 @@ classDiagram
...
+
+- Deprecated the `["cs:namespace"]`, `["java:package"]`, `["python:package"]`, and `["swift:module"]` metadata.
```
bernardnormier: I don't like these syntactically incorrect reference. `cs:namespace` etc. would be clearer in my view.

CHANGELOG-3.8.md
```diff
@@ -311,26 +311,92 @@ classDiagram
...
+  ```
+
+- Metadata can now be applied to Slice enumerators.
+
+- Removed the `["protected"]` metadata. This was primarily for classes with operations, which are no longer allowed.
+
 - Removed the `["preserve-slice"]` metadata. Slice classes marshaled in the sliced format are now always preserved when
   unmarshaled.
 
+- Added `["deprecated"]` as an alias for the `["deprecate"]` metadata.
```
bernardnormier: the class and exception should be kept together

InsertCreativityHere: I see what you mean. I moved `deprecated` out of the way of the sliced format discussions.

CHANGELOG-3.8.md
InsertCreativityHere: Yes, I removed the `["` metadata delimiters, like I did for the other incomplete metadata directives.

CHANGELOG-3.8.md
InsertCreativityHere: Fixed.

CHANGELOG-3.8.md
InsertCreativityHere: I removed the underscore.
They're just concatenated together, which seems fine.

CHANGELOG-3.8.md
InsertCreativityHere: Yes... I did...
Good catch!

CHANGELOG-3.8.md
InsertCreativityHere: Fair enough. I got carried away here.
They're no longer metadata, but I kept the backticks so they're formatted nicely.

PR #4053, opened 2025-06-04
Update Slice-to-C# struct mapping

bernardnormier:
This is a small update to the Slice to C# mapping for struct.

Before this PR, a struct with only "value type" field was mapped to a struct only if none of its fields had default values.

This PR removes the default value condition.  A modern C# struct can have default values for its fields / properties.

### DIFF
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -1715,7 +1715,6 @@ Slice::Gen::TypesVisitor::visitDataMember(const DataMemberPtr& p)
         addSemicolon = false;
     }
 
-    // Generate the default value for this field unless the enclosing type is a struct.
     if (p->defaultValue())
     {
         string defaultValue = *p->defaultValue();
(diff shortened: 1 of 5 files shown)


cpp/src/slice2cs/Gen.cpp
```diff
@@ -1564,6 +1564,27 @@ Slice::Gen::TypesVisitor::visitStructEnd(const StructPtr& p)
...
+    {
+        bool hasDefaultValue = false;
+        for (const auto& q : dataMembers)
+        {
+            if (q->defaultValue())
+            {
+                hasDefaultValue = true;
+                break;
+            }
+        }
```
InsertCreativityHere, line 1579: Probably could be:
```suggestion
        const bool hasDefaultValue = std::any_of(dataMembers.begin(), dataMembers.end(), [](const DataMemberPtr& q) { return q->defaultValue(); });
```
But split onto multiple lines I'd bet.

InsertCreativityHere (approved):
Looks good to me!

InsertCreativityHere (approved):

pepone (approved):

externl (approved):

PR #4054, opened 2025-06-05
Use package.json browser field for browser-friendly builds

pepone:
This PR updates how we handle Node.js-specific code by using the browser field in package.json. This allows bundlers like Vite, Webpack, and Rollup to automatically replace Node-only modules with browser versions.

## Added browser-specific files:

- FileLogger.browser.js: avoids importing `fs` by providing a no-op logger.
- TcpTransceiver.browser.js: avoids importing `net` by providing a no-op transceiver.
- TimerUtil.browser.js: replaces the old combined browser/Node logic. This avoids relying on typeof process, which could fail if process is polyfilled.

## Other changes:

- Enables ws and wss transports for Node.js. Modern Node.js versions support the WebSocket API, so we now run WebSocket tests in CI too.


js/gulpfile.js
```diff
@@ -146,23 +147,9 @@ gulp.task("test:common:generate", cb => {
     pump([gulp.src(["../scripts/Controller.ice"]), slice2js(), gulp.dest("test/Common")], cb);
 });
 
-// A rollup plug-in to resolve "net" module as a mockup file.
-
-const nodeMockups = ["net", "fs", "path"];
-function NodeMockupResolver() {
```
pepone, line 152: Replaced this custom plug-in by `@rollup/plugin-node-resolve` which allow to use "browser" package field.

js/gulpfile.js
```diff
@@ -252,11 +239,16 @@ for (const name of tests) {
...
             input: input,
-            plugins: [IceResolver(), NodeMockupResolver()],
+            plugins: [IceResolver()],
             onwarn: (warning, next) => {
                 // Ignore the "this is undefined" warning, let rollup silently rewrite it.
                 // This avoids warnings from the TypeScript polyfills for async disposable resources.
                 if (warning.code === "THIS_IS_UNDEFINED") return;
+
+                // Ignore unresolved fs/path dynamic imports (Node-only usage)
+                if (warning.code === "UNRESOLVED_IMPORT" && ["fs", "path"].includes(warning.source)) {
```
pepone, line 249: This is to avoid a warning in the properties test, which dynamically imports "fs" and "path"

js/src/Ice/Logger.js
```diff
@@ -1,7 +1,5 @@
 // Copyright (c) ZeroC, Inc.
 
-import fs from "fs";
```
pepone, line 3: We no longer import Node.js specific modules in files that are used by both Node/Browser. FileLogger used "fs" for Node and was split in FileLoger.js (Node implementation), FileLoger.browser.js (mockup for browser)

js/src/Ice/TcpTransceiver.js
```diff
@@ -9,354 +9,352 @@ import {
     ConnectFailedException,
     SocketException,
 } from "./LocalExceptions.js";
-import net from "net";
-
-let TcpTransceiver = null;
-
-if (typeof net.createConnection === "function") {
```
pepone, line 16: Same as before, just removed this big `if` as this is now only used in Node.js

js/src/Ice/TimerUtil.browser.js
```diff
@@ -0,0 +1,131 @@
+// Copyright (c) ZeroC, Inc.
+
+let Timer = {};
+
```
pepone, line 4: This is the browser code, which was previously in TimerUtil.js, with and if/else for Node.js/Browser versions.

js/src/Ice/WSEndpoint.js
```diff
@@ -151,6 +151,6 @@ export class WSEndpoint extends EndpointI {
     }
 
     connectable() {
-        return typeof WebSocket !== "undefined";
+        return true;
```
pepone, line 154: WebSockets are now enabled in both Browser/Node.js

js/src/Ice/WSTransceiver.js
```diff
@@ -6,333 +6,329 @@ import { ConnectionInfo as SSLConnectionInfo } from "./SSL/ConnectionInfo.js";
...
+        return Math.min(t, 25);
+    }
+
+    setCallbacks(connectedCallback, bytesAvailableCallback, bytesWrittenCallback) {
+        this._connectedCallback = connectedCallback;
+        this._bytesAvailableCallback = bytesAvailableCallback;
+        this._bytesWrittenCallback = bytesWrittenCallback;
+    }
 
-if (typeof WebSocket !== "undefined") {
```
pepone, line 11: Just removed this `if`, WebSocket is supported in modern Node.js, we don't support the older versions anymore.

js/test/Ice/binding/Client.ts
```diff
@@ -7,9 +7,9 @@ import { TestHelper } from "../../Common/TestHelper.js";
 const ArrayUtil = Ice.ArrayUtil;
 const test = TestHelper.test;
 
-const isConnectionFailed = (ex: Ice.Exception) =>
-    (!TestHelper.isBrowser() && ex instanceof Ice.ConnectionRefusedException) ||
-    (TestHelper.isBrowser() && ex instanceof Ice.ConnectFailedException) ||
+const isConnectionFailed = (defaultProtocol: string, ex: Ice.Exception) =>
```
pepone, line 10: The exception is different base on the protocol, previously using isBrowser == false was equivalent to TCP transport, but that is no longer the case.

js/test/Ice/info/Client.ts
```diff
@@ -125,12 +110,19 @@ export class Client extends TestHelper {
             test(parseInt(ctx.get("localPort")!) === ipConnectionInfo!.remotePort);
         }
 
+        function getHeader(ctx: Map<string, string>, key: string): string | undefined {
```
pepone, line 113: Node.js is using lower case for the keys.

externl (approved):
Looks good!

PR #4060, opened 2025-06-06
Replace ERR_get_error_line_data usage

### LINKED ISSUE #4022: Replace usage of deprecated ERR_get_error_line_data
The OpenSSL implementation of the ssl transport uses `ERR_get_error_line_data` which is deprecated in OpenSSL 3.0, we should check if we can replace its usage.

https://docs.openssl.org/3.5/man3/ERR_get_error/#synopsis

pepone:
fix #4022

### DIFF
--- a/cpp/include/Ice/SSL/Config.h
+++ b/cpp/include/Ice/SSL/Config.h
@@ -45,6 +45,9 @@
 #    include <Security/SecureTransport.h>
 #    include <Security/Security.h>
 #else
+#    ifndef OPENSSL_API_COMPAT
+#        define OPENSSL_API_COMPAT 0x30000000L // Target OpenSSL 3.0 or later
+#    endif
 #    define ICE_USE_OPENSSL
 #    include <openssl/err.h>
 #    include <openssl/ssl.h>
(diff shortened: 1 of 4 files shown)


cpp/include/Ice/SSL/Config.h
```diff
@@ -45,6 +45,9 @@
 #    include <Security/SecureTransport.h>
 #    include <Security/Security.h>
 #else
+#    ifndef OPENSSL_API_COMPAT
+#        define OPENSSL_API_COMPAT 0x30000000L // Target OpenSSL 3.0 or later
```
externl: What is this about?

externl: Do we only supports the OpenSSL >= 3 now?

pepone: this is to not include older APIs deprecated in 3.0

bernardnormier: Should update the comment to make it clearer.

cpp/src/Ice/SSL/SSLUtil.cpp
```diff
@@ -576,57 +576,26 @@ Ice::SSL::ScopedCertificate::~ScopedCertificate()
...
+Ice::SSL::getErrors()
 {
     ostringstream ostr;
 
-    const char* file;
-    const char* data;
-    int line;
-    int flags;
     unsigned long err;
     int count = 0;
```
bernardnormier: You should replace this count by a bool, like `firstError`.

bernardnormier (approved):

bernardnormier (approved):

PR #4063, opened 2025-06-06
Remove debian zeroc-ice-all-rutime zeroc-ice-all-dev metapackage

pepone:


packaging/dpkg/debian/control
```diff
@@ -284,3 +247,21 @@ Description: Bridge service
  with minimal effort. Ice takes care of all interactions with low-level
  network programming interfaces and allows you to focus your efforts on
  your application logic.
+
+Package: zeroc-ice-all-dev
```
bernardnormier, line 251: When you install or upgrade to this package, it installs nothing?

bernardnormier (approved):

externl (approved):

packaging/dpkg/debian/control
pepone: It still install the README/LICENSE but because it is in `oldlibs` it can be autoremoved. oldlibs is a special section for transitional packages that can be safely removed.

It is only required for the 3.7 upgrade.

PR #4068, opened 2025-06-06
Convert CtrlCHandler into a struct (Swift)

bernardnormier:


externl (approved):

swift/src/Ice/CtrlCHandler.swift
```diff
@@ -4,10 +4,10 @@ import IceImpl
 
 /// Helps applications handle Ctrl+C (SIGINT) and similar signals (SIGHUP and SIGTERM). Only available on macOS.
 @available(macOS 14, *)
```
externl: should this be 15 now?

externl (approved):

swift/src/Ice/CtrlCHandler.swift
bernardnormier: Yes

PR #4069, opened 2025-06-06
Use more structs in Swift

bernardnormier:
This PR replaces several classes by structs in Swift, + misc cleanup.

### DIFF
--- a/cpp/src/slice2swift/Gen.cpp
+++ b/cpp/src/slice2swift/Gen.cpp
@@ -993,7 +993,7 @@ Gen::TypesVisitor::visitEnum(const EnumPtr& p)
     out << nl << "/// Read an enumerated value.";
     out << nl << "///";
     out << nl << "/// - Returns:  The enumerated value.";
-    out << nl << "func read() throws -> sending " << name;
+    out << nl << "func read() throws -> " << name;
     out << sb;
     out << nl << "let rawValue: " << enumType << " = try read(enumMaxValue: " << p->maxValue() << ")";
     out << nl << "guard let val = " << name << "(rawValue: rawValue) else";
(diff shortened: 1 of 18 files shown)


externl (approved):

cpp/src/slice2swift/Gen.cpp
```diff
@@ -993,7 +993,7 @@ Gen::TypesVisitor::visitEnum(const EnumPtr& p)
     out << nl << "/// Read an enumerated value.";
     out << nl << "///";
     out << nl << "/// - Returns:  The enumerated value.";
-    out << nl << "func read() throws -> sending " << name;
+    out << nl << "func read() throws -> " << name;
```
pepone, line 996: Was this sending bogus?

swift/src/Ice/IncomingRequest.swift
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) ZeroC, Inc.
 /// Represents a request received by a connection. It's the argument given to `Dispatcher.dispatch`.
-public final class IncomingRequest {
+public struct IncomingRequest {
```
pepone, line 3: Why is a struct preferred here?

swift/src/Ice/ProcessI.swift
```diff
@@ -2,7 +2,7 @@
 
 import IceImpl
 
-class ProcessI: Process, @unchecked Sendable {
+struct ProcessI: Process {
```
pepone, line 5: No clear why using a struct is preferred here.

pepone (approved):

cpp/src/slice2swift/Gen.cpp
bernardnormier: It was not bogus, just unnecessary. We only need sending for NonSendable types, and all generated enums are now Sendable.

swift/src/Ice/IncomingRequest.swift
bernardnormier: structs are in general preferred over classes in Swift, especially for simple objects.

Most types in the Swift library are structs: String, Array, Dictionary, Mutex...

PR #4071, opened 2025-06-08
Update Swift code to use Synchronization.Mutex

bernardnormier:
This PR updates the Swift source code to use Mutex instead of our own Mutex, when doable.

It also renames the existing internal Ice.Mutex struct to UncheckedMutex.


swift/src/Ice/ServantManager.swift
```diff
@@ -12,7 +12,7 @@ final class ServantManager: Dispatcher {
     // when a servant is not found on a Swift Admin OA.
     private var adminId: Identity?
 
-    private var mutex = Mutex()
+    private var mutex = UncheckedMutex()
```
bernardnormier, line 15: We can't use Synchronization.Mutex to protect ServantManager's state because Dispatcher is not Sendable.

swift/test/Ice/udp/AllTests.swift
```diff
@@ -6,37 +6,25 @@ import Ice
 import TestCommon
 
 class PingReplyI: PingReply, @unchecked Sendable {
-    var _replies: Int32 = 0
-    var _lock = os_unfair_lock()
-    var _semaphore = DispatchSemaphore(value: 0)
+    private let _semaphore = DispatchSemaphore(value: 0)
```
bernardnormier, line 9: All the updates in this PR are straightforward except the updates to this file.

The main fix is to get rid of the _replies property and just use the DispatchSemaphore, including its count.

swift/test/Ice/udp/AllTests.swift
```diff
@@ -6,37 +6,25 @@ import Ice
...
+        try uncheckedCast(prx: adapter.addWithUUID(self), type: PingReplyPrx.self).ice_datagram()
     }
 
     func waitReply(expectedReplies: Int, timeout: Int) -> Bool {
+        precondition(timeout > 0, "Timeout must be greater than 0")
         let end = DispatchTime.now() + .milliseconds(timeout)
-        while _replies < expectedReplies {
-            let begin = DispatchTime.now()
-            let delay = end.uptimeNanoseconds - begin.uptimeNanoseconds
-            if delay > 0 {
```
bernardnormier, line 31: Since the timeout is absolute (a deadline), not clear we included this logic.

swift/test/Ice/udp/AllTests.swift
```diff
@@ -48,11 +36,6 @@ public func allTests(_ helper: TestHelper) async throws {
     let communicator = helper.communicator()
     communicator.getProperties().setProperty(key: "ReplyAdapter.Endpoints", value: "udp")
     let adapter = try communicator.createObjectAdapter("ReplyAdapter")
-    var replyI = PingReplyI()
```
bernardnormier, line 51: All the create then later reset logic doesn't make sense.

swift/test/Ice/udp/AllTests.swift
```diff
@@ -91,11 +71,12 @@ public func allTests(_ helper: TestHelper) async throws {
...
         do {
             seq = ByteSeq(repeating: 0, count: 1024)
+            let replyI = PingReplyI()
+            let reply = try replyI.getProxy(adapter)
+
             while true {
                 seq = ByteSeq(repeating: 0, count: seq.count * 2 + 10)
-                replyI.reset()
                 try await obj.sendByteSeq(seq: seq, reply: reply)
-                _ = replyI.waitReply(expectedReplies: 1, timeout: 10000)
```
bernardnormier, line 98: We're looping until an invocation throws DatagramLimitException. What's the point of this "waitReply" call that doesn't throw?

swift/test/Ice/udp/AllTests.swift
```diff
@@ -108,7 +89,9 @@ public func allTests(_ helper: TestHelper) async throws {
         communicator.getProperties().setProperty(key: "Ice.UDP.SndSize", value: "64000")
         seq = ByteSeq(repeating: 0, count: 50000)
         do {
-            replyI.reset()
+            let replyI = PingReplyI()
+            let reply = try replyI.getProxy(adapter)
+
```
bernardnormier, line 94: I find the `catch is Ice.DatagramLimitException {}` below suspicious: either the `setProperty(key: "Ice.UDP.SndSize", value: "64000")` above worked and the invocation should not throw this exception. Otherwise, this test is pointless.

swift/test/Ice/udp/AllTests.swift
```diff
@@ -48,11 +36,6 @@ public func allTests(_ helper: TestHelper) async throws {
     let communicator = helper.communicator()
     communicator.getProperties().setProperty(key: "ReplyAdapter.Endpoints", value: "udp")
     let adapter = try communicator.createObjectAdapter("ReplyAdapter")
```
pepone, line 38: How do we know this test will not use collocation? The reply object and proxy share the same communicator.

swift/test/Ice/udp/AllTests.swift
pepone: I agree seems unnecessary

pepone (approved):

swift/src/Ice/ServantManager.swift
externl: Should we have a comment on this field?

swift/test/Ice/udp/AllTests.swift
externl: We should review this test in other languages.

externl (approved):
Looks good!

swift/test/Ice/udp/AllTests.swift
bernardnormier: In these UDP tests, we pass the reply proxy as an argument to `ping`, and it's the ping implementation in the server  (TestI.swift) that calls this proxy. There is no collocation concern.

PR #4075, opened 2025-06-09
Make Swift ServantLocator Sendable

externl:


swift/test/Ice/servantLocator/ServantLocatorI.swift
```diff
@@ -5,33 +5,36 @@ import Ice
 import Synchronization
 import TestCommon
 
-class Cookie {
+private class Cookie {
```
bernardnormier: I would nest both Cookie and State inside ServantLocatorI.

swift/test/Ice/invoke/TestI.swift
```diff
@@ -16,7 +16,7 @@ final class ServantLocatorI: Ice.ServantLocator {
     func deactivate(_: String) {}
 }
 
-final class DispatcherI: Ice.Dispatcher {
+struct DispatcherI: Ice.Dispatcher {
```
bernardnormier, line 19: I don't think this dispatch should be async.

bernardnormier (approved):

PR #4077, opened 2025-06-09
Update Swift ServantManager to use Synchronization.Mutex

bernardnormier:
This PR also removes UncheckedMutex.


swift/src/Ice/ObjectAdapterI.swift
```diff
@@ -24,6 +22,7 @@ final class ObjectAdapterI: LocalObject<ICEObjectAdapter>, ObjectAdapter, ICEDis
     }
 
     private let communicator: Communicator
+    private let servantManager: ServantManager
```
bernardnormier, line 25: With the addition of setAdminId below, we no longer need to expose ServantManager to any other class - it's cleaner but doesn't change much otherwise.

swift/src/Ice/ServantManager.swift
```diff
@@ -1,28 +1,28 @@
...
-// TODO: use Mutex and remove @unchecked Sendable
-final class ServantManager: Dispatcher, @unchecked Sendable {
-    private let adapterName: String
-    private let communicator: Communicator
+import Synchronization
 
-    private var servantMapMap = [Identity: [String: Dispatcher]]()
-    private var defaultServantMap = [String: Dispatcher]()
-    private var locatorMap = [String: ServantLocator]()
+final class ServantManager: Dispatcher {
```
bernardnormier, line 5: ServantManager can't be a struct because it has a Mutex stored property.

externl (approved):
Looks good!

swift/src/Ice/ServantManager.swift
```diff
@@ -1,28 +1,28 @@
...
 
-    // This is used to distinguish between ObjectNotExistException and FacetNotExistException
-    // when a servant is not found on a Swift Admin OA.
-    private var adminId: Identity?
+    private struct State {
```
bernardnormier, line 7: Per the Swift Standard Library Programmer's manual "Crucial type aliases and nested types, not exceeding a handful of lines in length" appear first in a type declaration.

PR #4081, opened 2025-06-10
Add Xcode support to SliceCompile SPM plug-in

pepone:
This PR replaces #4062

- Add support to use the SliceCompile with Xcode projects.
- Update the plug-in to compile all `.ice` files in the target inputs by default.
- The configuration is now optional, for additional sources or for configuring the search_path.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -41,28 +41,24 @@ let package = Package(
             name: "Ice",
             dependencies: ["IceImpl"],
             path: "swift/src/Ice",
-            exclude: ["slice-plugin.json"],
             plugins: [.plugin(name: "CompileSlice")]
         ),
         .target(
             name: "Glacier2",
             dependencies: ["Ice"],
             path: "swift/src/Glacier2",
-            exclude: ["slice-plugin.json"],
             plugins: [.plugin(name: "CompileSlice")]
         ),
         .target(
             name: "IceGrid",
             dependencies: ["Ice", "Glacier2"],
             path: "swift/src/IceGrid",
-            exclude: ["slice-plugin.json"],
             plugins: [.plugin(name: "CompileSlice")]
         ),
         .target(
             name: "IceStorm",
             dependencies: ["Ice"],
             path: "swift/src/IceStorm",
-            exclude: ["slice-plugin.json"],
             plugins: [.plugin(name: "CompileSlice")]
         ),
         .target(
--- a/swift/Plugins/CompileSlice/CompileSlicePlugin.swift
+++ b/swift/Plugins/CompileSlice/CompileSlicePlugin.swift
@@ -3,14 +3,48 @@
 import Foundation
 import PackagePlugin
 
-enum PluginError: Error {
+// The entry point for command line builds with SPM.
+extension CompileSlicePlugin: BuildToolPlugin {
+    func createBuildCommands(context: PluginContext, target: Target) async throws -> [Command] {
+        guard let swiftTarget = target as? SwiftSourceModuleTarget else {
+            throw PluginError.invalidTarget("\(type(of: target))")
+        }
 
+        return try createBuildCommands(
+            outputDir: context.pluginWorkDirectoryURL,
+            inputFiles: swiftTarget.sourceFiles,
+            slice2swift: try context.tool(named: "slice2swift").url
+        )
+    }
+}
+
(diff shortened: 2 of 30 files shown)


Package.swift
```diff
@@ -41,28 +41,24 @@ let package = Package(
             name: "Ice",
             dependencies: ["IceImpl"],
             path: "swift/src/Ice",
-            exclude: ["slice-plugin.json"],
```
pepone, line 44: The plug-in adds the config file to the build command inputs. This avoid the need to exclude, and ensures compilation run when the configuration changes.

swift/Plugins/CompileSlice/CompileSlicePlugin.swift
```diff
@@ -22,108 +56,104 @@ enum PluginError: Error {
...
 ///     "search_paths": ["Slice", "../OtherModule/Slice"]
 /// }
 /// ```
 struct Config: Codable {
-    /// List of Slice files or directories containing Slice files
-    var sources: [String]
-    /// List of directories to add as search paths when calling slice2swift.
+    /// Optional list of Slice files or directories containing Slice files.
+    /// Paths are relative to the `slice-plugin.json` file. Slice files in the target's sources are always included.
+    var sources: [String]?
```
pepone, line 78: sources is now optional, to add Slice files that are not in the target inputs. Maybe would be clear to have `additional_sources`.

swift/Plugins/CompileSlice/CompileSlicePlugin.swift
```diff
@@ -22,108 +56,104 @@ enum PluginError: Error {
...
             return .buildCommand(
-                displayName: displayName,
+                displayName: "Compile Slice \(sliceSource.lastPathComponent)",
                 executable: slice2swift,
-                arguments: arguments,
-                outputFiles: [outputFile]
+                arguments: searchPaths + ["--output-dir", outputDir.path, sliceSource.path],
+                // It's important to declare both the Slice file and the config file as input files so the build system can
+                // detect changes. This also avoids warnings about unused files.
+                // TODO: We should also include imported Slice dependencies as inputs.
```
pepone, line 154: Would be nice to fix this, not clear if we can run `slice2swift --depend` at this point.

swift/test/Ice/adapterDeactivation/slice-plugin.json
```diff
@@ -1,3 +0,0 @@
-{
-  "sources": ["Test.ice"]
-}
```
pepone, line 3: For this simple case no need to include a config anymore, we just build all `.ice` files in the target inputs.

swift/Plugins/CompileSlice/CompileSlicePlugin.swift
externl: I think we need to restructure this config in general. Maybe do in a follow-up PR?

pepone: I prefer to do it in a separate PR, as it is not clear what the new structure would be. The main issue fixed here is to allow using the plug-in in Xcode projects.

swift/test/Package.swift
externl, line 1: I think we can also get rid of the list of slice files.

externl:
Looks good!

swift/Plugins/CompileSlice/CompileSlicePlugin.swift
```diff
@@ -3,14 +3,48 @@
 import Foundation
 import PackagePlugin
 
-enum PluginError: Error {
+// The entry point for command line builds with SPM.
+extension CompileSlicePlugin: BuildToolPlugin {
+    func createBuildCommands(context: PluginContext, target: Target) async throws -> [Command] {
```
bernardnormier, line 8: Is the `async` necessary?

pepone: That is the definition in `BuildToolPlugin` other plug-in seems to do the same even for synchronous code. Not sure how this work in Swift

externl: In Swift you don't have do declare throws and async if your function implementing a protocol doesn't need it. We've been removing them, not sure how good of an idea it is.

PR #4087, opened 2025-06-10
Mention Triple-Slash Doc-Comments in the CHANGELOG

InsertCreativityHere:
I couldn't find a commit with this in it's message or description, but this is something else we added in 3.8,
so it should be mentioned in the CHANGELOG.

---

After adding it to the changelog, I copied my example into a test Slice file (to make sure it was valid Slice!), but I noticed that the old `/** ... */` comments didn't seem to be generating anything. Sure enough, there was actually a bug in the scanner, causing us to completely ignore JavaDoc style comments in some cases.
I fixed this bug.

### DIFF
--- a/cpp/src/Slice/Parser.h
+++ b/cpp/src/Slice/Parser.h
@@ -1060,7 +1060,7 @@ namespace Slice
         /// @param comment The raw comment string. It shouldn't contain newlines, nor any comment formatting characters.
         ///     Unlike `setDocComment`, this function doesn't process the comment string in any way.
         // Note that this is only called for triple-slash style comments, where each line is added separately.
-        void addToDocComment(const std::string& comment);
+        void appendToDocComment(const std::string& comment);
         /// Returns the current doc-comment string. This function uses move semantics. After calling it, the currently
         /// stored doc-comment will be cleared. This string can span multiple lines, and won't include formatting like
         /// leading '///' or opening '/**' etc.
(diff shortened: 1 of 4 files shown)


cpp/src/Slice/Scanner.cpp
```diff
@@ -1970,28 +1970,28 @@ case 28:
...
-    popState();
-
     string comment(yytext);
     // The last 2 characters are the '*/' matched by this rule.
     if (YY_START == C_DOC_COMMENT)
     {
         currentUnit->setDocComment(comment.substr(0, yyleng - 2));
     }
+
+    popState();
```
InsertCreativityHere, line 1980: The issue was that we were popping off the current state before we checked it.

When this function is called, the scanner is in `C_DOC_COMMENT` mode.
But calling `popState` changes the scanner to it's previous state.
Meaning `YY_START` (the current state) would not be what we expect here.

pepone (approved):

cpp/src/Slice/Parser.h
```diff
@@ -1051,9 +1051,10 @@ namespace Slice
 
         [[nodiscard]] std::string languageName() const;
 
-        void setDocComment(const std::string& comment);
-        void addToDocComment(const std::string& comment);
+        void setDocComment(const std::string& comment); // Only used for JavaDoc style comments.
```
bernardnormier: Would be nice to pick clearer names and actually add doc-comments for these functions.

InsertCreativityHere: I think `setDocComment` is clear, but I did change `addToDocComment`, and added proper doc-comments for all 3 of these functions.

PR #4088, opened 2025-06-10
Use authentication token to access ice-swift-nightly repository

pepone:


externl (commented):
I think we should use the built-in GITHUB_TOKEN.

packaging/swift/update-ice-swift-nightly.sh
```diff
@@ -51,4 +54,4 @@ git config user.name "ZeroC"
 git config user.email "git@zeroc.com"
 git commit -m "ice: $ice_version Nightly build"
 git tag -a "$ice_version" -m "ice: $ice_version"
-git push origin main
+git push "https://x-access-token:${ICE_SWIFT_NIGHTLY_TOKEN}@github.com/zeroc-ice/ice-swift-nightly.git" main --tags
```
externl: Is this necessary? Does the remote not retain the token?

.github/workflows/build-nightly-release-workflow.yml
```diff
@@ -230,6 +230,5 @@ jobs:
     with:
       ice_version: ${{ needs.set-version.outputs.semver_version }}
       repository_url: "${{ inputs.repository_url }}/nightly/"
-      repository_username: ${{ inputs.repository_username }}
     secrets:
-      REPOSITORY_PASSWORD: ${{ secrets.NEXUS_NIGHTLY_PASSWORD }}
+      ICE_SWIFT_NIGHTLY_TOKEN: ${{ secrets.ICE_SWIFT_NIGHTLY_TOKEN }}
```
externl: Maybe we should have a shared token that we can use for the brew tap as well: `secrets. GITHUB_NIGHTLY_PAT `, and use `GITHUB_NIGHTLY_PAT` for the env name.

pepone: What does PAT stand for?

packaging/swift/update-ice-swift-nightly.sh
pepone: I guess no.

.github/workflows/build-nightly-release-workflow.yml
externl: Personal access token.

pepone: Created a shared token ICE_NIGHTLY_PUBLISH_TOKEN and updated the workflow to use it.

PR #4089, opened 2025-06-10
iOS controller fixes

externl:


swift/test/ios/TestDriverApp/ControllerI.swift
```diff
@@ -28,29 +29,18 @@ class ProcessI: CommonProcess {
...
         _view = view
         _ipv4 = ipv4
         _ipv6 = ipv6
-
-        //TODO: remove the queues
-        _serverDispatchQueue = DispatchQueue(label: "Server", qos: .background)
-        _clientDispatchQueue = DispatchQueue(label: "Client", qos: .background)
     }
 
+    @MainActor
```
externl, line 43: I forget the exact error but we're running on the main actor as the compiler is not happy about _view.println being potentially called concurrently.

bernardnormier (approved):

Issue #210, opened 2018-09-22
"deprecate" metadata


bernardnormier:
Affects Ice 3.7

The [Slice deprecate metadata](https://doc.zeroc.com/ice/3.7/the-slice-language/deprecating-slice-definitions) applies to various Slice definitions and currently affects only the generated code. If you use a Slice definition marked "deprecate" in another Slice definition, you don't get a warning when compiling your Slice file. For example, Slice-compiling the following Slice file does not emit any warning:
```
module DepTest
{
    ["deprecated:foobar"] interface A
    {
        void op();
        A* get();
    }

    ["deprecate"] struct S
    {
        int x;
    }

    interface B
    {
        A* get();
        void op(S s);
    }
}
```

We should:
(a) consider renaming "deprecate" to "deprecated" to match the corresponding C++ attribute and Java annotation

(b) emit a warning when Slice-compiling a definition that uses a deprecated type, except when this definition is itself marked deprecated.

For example:
```
    ["deprecate"] struct S
    {
        int x;
    }

    interface B
    {
        void op(S s);
    }
```
should emit a warning unless `op` is itself marked deprecated.

(operations on a deprecated interface can use this deprecated type without being marked themselves deprecated)

PR #4093, opened 2025-06-11
Add Slice Deprecation Warnings

InsertCreativityHere:
This PR implements the long-awaited #210.

Now, the Slice compiler will issue a warning if you use a deprecated thing (unless the referencer is also marked 'deprecated').

----

Our own Slice files only generated one warning after this change, in `slice/IceGrid/Descriptor.ice`.
We have this field: `BoxedDistributionDescriptor distrib;`, where the type is marked deprecated.
There are 3 other fields named `distrib` in this file (but with different types), all 3 of them are marked deprecated,
so I figure this other one I came across should probably be `deprecated` as well.


cpp/test/Slice/deprecated/Test.ice
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) ZeroC, Inc.
```
InsertCreativityHere, line 2: This test is about how deprecated affects the generated code.
So, we don't want `slice2cpp` emitting deprecation warnings in it.

slice/IceGrid/Descriptor.ice
```diff
@@ -450,6 +450,7 @@ module IceGrid
         BoxedString description;
 
         /// The updated distribution application descriptor.
+        ["deprecated"]
```
InsertCreativityHere, line 453: The only deprecation warning in our Slice files.
`BoxedDistributionDescriptor` is a deprecated type.

So, I assume this field should just be marked deprecated.
Any `IceGrid` experts should feel free to correct me!

cpp/src/Slice/Parser.cpp
```diff
@@ -5078,6 +5079,8 @@ Slice::Unit::parse(const string& filename, FILE* file, bool debugMode)
         popContainer();
         assert(_definitionContextStack.size() == 1);
         popDefinitionContext();
+
+        emitDeprecationWarningsFor(unit());
```
InsertCreativityHere, line 5083: We emit any deprecation warnings after parsing finishes (only if parsing was successful), but before we hand control over to the language-specific code.

cpp/src/Slice/DeprecationReporter.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include "DeprecationReporter.h"
+
+#include <sstream>
+
+using namespace std;
+using namespace Slice;
+
+namespace
+{
+    class DeprecationVisitor final : public ParserVisitor
```
InsertCreativityHere, line 12: Unfortunately, a visitor is the only way to check this, because of how compilation works in Ice. Bison parses from the inside out, so we don't know what metadata will be applied to something until we've already fully parsed it.

Issue #3456, opened 2025-01-30
Should We Fix `scope` and `scoped` in the Slice Parser to no Longer Emit a Leading Scope?


InsertCreativityHere:
Currently, when you call any of `scope`, `scoped`, `mappedScope` or `mappedScoped`, they will always start with a leading scope separator.
Often this is a pain, and we have to `substr` this leading separator away. So it might make more sense to just not emit it.

----

If we do decide to keep it, we should add a `prefix` parameter to these functions that allows you to opt-out of having it, or customize it.
Since while `::` is globally scoped, in C# it's `global::`, so it's different among different languages.

See the original discussion here:
https://github.com/zeroc-ice/ice/pull/3445#discussion_r1936008685

InsertCreativityHere:
This was implemented in #4094.

PR #4094, opened 2025-06-11
Add Bool to Remove Leading Scope Separator from `mappedScoped`

InsertCreativityHere:
This PR implements #3456.
Currently, the `mappedScope` and `mappedScoped` functions in the Slice parser will always return a leading scope separator:
`::Foo::Struct`, `.Foo.Struct`, `_Foo_Struct`.
Sometimes this is what we want, sometimes it isn't. But the current 'solution' of `mappedScope().substr(1)` isn't elegant.

So this PR adds a bool to `mappedScope` and `mappedScoped`, which tells the compiler whether the leading separator is desired.

----

I checked the generated code with my comparison script: this PR doesn't change any generated code for any compilers.
Which is good.


cpp/src/Slice/Parser.cpp
```diff
@@ -995,20 +995,24 @@ Slice::Contained::mappedName() const
...
     string scoped;
     if (auto container = dynamic_pointer_cast<Contained>(_container))
     {
-        scoped = container->mappedScoped(separator);
+        scoped = container->mappedScoped(separator, leading) + separator;
     }
-    return scoped + separator;
+    else if (leading)
+    {
+        scoped = separator;
```
InsertCreativityHere, line 1014: This method emits both a leading and trailing separator (like : `::Foo::`).
The behavior of the trailing separator is unchanged,
But, now the leading separator is only emitted if `leading` is true.

cpp/src/Slice/SliceUtil.cpp
```diff
@@ -509,8 +509,7 @@ Slice::isProxyType(const TypePtr& type)
 bool
 Slice::isFirstElement(const ContainedPtr& p)
 {
-    ContainerPtr container = dynamic_pointer_cast<Container>(p->container());
-    return p == container->contents().front();
```
InsertCreativityHere, line 513: Unrelated to this PR. But `container()` returns a `ContainerPtr`... so... no idea why we're dynamic casting here.

cpp/src/slice2swift/SwiftUtil.h
```diff
@@ -37,7 +37,6 @@ namespace Slice
             const StringList&,
             bool commentFirst = true,
             const std::string& space = " ");
-        void writeSeeAlso(IceInternal::Output&, const StringList&, const ContainerPtr&);
```
InsertCreativityHere, line 40: This function isn't implemented anywhere. Nor called anywhere.
100% dead.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -32,29 +32,23 @@ namespace
         return cont->getMetadataArgs("cs:identifier").value_or(cont->name());
     }
```
InsertCreativityHere, line 34: This is only tangentially related to my PR. I merged `getCsMappedScoped` into `getCsMappedScope`, since that was the only place it was being called from, and the function was a one-liner.

Also, we _never_ want a leading `.` in this context, so I made the same change as I did in `libSlice`, but assuming that `leading` is always `false`.

cpp/src/Slice/Parser.h
```diff
@@ -395,10 +395,14 @@ namespace Slice
         /// Returns the mapped identifier that this element will use in the target language.
         [[nodiscard]] std::string mappedName() const;
         /// Returns the mapped scope that this element will be generated in in the target language.
-        /// (equivalent to `mappedScope(separator) + mappedName()`).
-        [[nodiscard]] std::string mappedScoped(const std::string& separator = "::") const;
+        /// @param separator this string will be used to separate scope segments (defaults to "::").
+        /// @param leading if true, the returned string will have a leading `separator` (defaults to true).
+        /// @remark equivalent to `mappedScope(separator, leading) + mappedName()`.
+        [[nodiscard]] std::string mappedScoped(const std::string& separator = "::", bool leading = true) const;
```
bernardnormier: You should insert a blank like before the doc-comment block of the next function.

cpp/src/Slice/Parser.h
```diff
@@ -395,10 +395,14 @@ namespace Slice
         /// Returns the mapped identifier that this element will use in the target language.
         [[nodiscard]] std::string mappedName() const;
         /// Returns the mapped scope that this element will be generated in in the target language.
-        /// (equivalent to `mappedScope(separator) + mappedName()`).
-        [[nodiscard]] std::string mappedScoped(const std::string& separator = "::") const;
+        /// @param separator this string will be used to separate scope segments (defaults to "::").
```
bernardnormier: We capitalize the description in C++ like in C#: This string [...]. 

Java and possibly other languages use different conventions.

cpp/src/Slice/Parser.h
externl: I think we should remove the default separator value. It really only applies to C++. Since most (at least half I think) of the class now explicitly need to pass a separator so they can pass false. I think we should remove the default bool for leading as well.

InsertCreativityHere: Fair enough about the separator. I removed the default, since like you said, it wasn't used very much.

For the bool parameter, how would you feel about switching the default to `false` instead of `true`?
There's 110 places where we pass `false`, and only 61 places where we pass `true`.
And I think "not having extra stuff at the front" is a reasonable and expected behavior to get by default.
Or are you just against the idea of a default bool here in general?

externl: Up to you! I'm fine with making false the default.

InsertCreativityHere: Fixed!

cpp/src/Slice/Parser.h
InsertCreativityHere: I capitalized the lines. I was too used to writing JavaDoc!

PR #4095, opened 2025-06-11
Simplify Default Constructor Argument Generation in `slice2js`

InsertCreativityHere:
Just a small PR to move some triplicated code into a helper function.
We copy/pasted the exact same code for classes, exceptions, and structs.
Might as well move it into a helper function I figured.

### DIFF
--- a/cpp/src/slice2js/Gen.h
+++ b/cpp/src/slice2js/Gen.h
@@ -21,7 +21,7 @@ namespace Slice
     protected:
         void writeMarshalDataMembers(const DataMemberList&, const DataMemberList&);
         void writeUnmarshalDataMembers(const DataMemberList&, const DataMemberList&);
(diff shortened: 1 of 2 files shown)


cpp/src/slice2js/Gen.h
```diff
@@ -21,7 +21,7 @@ namespace Slice
     protected:
         void writeMarshalDataMembers(const DataMemberList&, const DataMemberList&);
         void writeUnmarshalDataMembers(const DataMemberList&, const DataMemberList&);
-        void writeInitDataMembers(const DataMemberList&);
+        void writeOneShotConstructorArguments(const DataMemberList& members);
```
InsertCreativityHere, line 24: `writeInitDataMembers` was 100% dead, not even implemented anywhere.
So. That's just free real estate there!

Issue #4097, opened 2025-06-12
Validate Ice.MessageSizeMax and report errors


pepone:
With Ice 3.7 if the value of `Ice.MessageSizeMax` doesn't fit with the expected range we silently adjust it to max int32 value.

The maximum size for an Ice message is 2,147,483,647 bytes (~2.14 GB), this is because the protocol uses an int32 for the message size.

And because `Ice.MessageSizeMax` is specified in kilobytes the maximum accepted value for the property is 2,147,483,647/1,024 (~2097151 kilobytes)

https://github.com/zeroc-ice/ice/blob/83580f61c84d70a2bba5f523f0acddb1c213f601/cpp/src/Ice/Instance.cpp#L1075-L1084

Instead of silently adjusting we should report an initialization error. We need to review all language mappings.

PR #4098, opened 2025-06-12
Validate Ice.MessageSizeMax and Ice.BatchAutoFlushSize properties

pepone:
This is the C++ side of #4097

### DIFF
--- a/cpp/src/Ice/Instance.cpp
+++ b/cpp/src/Ice/Instance.cpp
@@ -1071,54 +1071,45 @@ IceInternal::Instance::initialize(const Ice::CommunicatorPtr& communicator)
         const_cast<ConnectionOptions&>(_clientConnectionOptions) = readConnectionOptions("Ice.Connection.Client");
         const_cast<ConnectionOptions&>(_serverConnectionOptions) = readConnectionOptions("Ice.Connection.Server");
 
-        // The max size of an Ice protocol message in bytes. The maximum value is 0x7fffffff, which is the maximum
-        // value for an int32_t.
-        const int32_t size_max = 0x7fffffff;
+        // The maximum size of an Ice protocol message in bytes. This is limited to 0x7fffffff, which corresponds to
+        // the maximum value of a 32-bit signed integer (int32_t).
+        const int32_t messageSizeMaxUpperLimit = numeric_limits<int32_t>::max();
         {
             int32_t num = _initData.properties->getIcePropertyAsInt("Ice.MessageSizeMax");
-            if (static_cast<size_t>(num) > static_cast<size_t>(size_max / 1024))
+            if (static_cast<size_t>(num) > static_cast<size_t>(messageSizeMaxUpperLimit / 1024))
             {
                 ostringstream os;
                 os << "Ice.MessageSizeMax '" << num << "' is too large, it must be less than or equal to '"
-                   << (size_max / 1024) << "'";
+                   << (messageSizeMaxUpperLimit / 1024) << "' KiB";
                 throw InitializationException(__FILE__, __LINE__, os.str());
             }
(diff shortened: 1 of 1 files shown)


cpp/src/Ice/Instance.cpp
```diff
@@ -1071,11 +1071,21 @@ IceInternal::Instance::initialize(const Ice::CommunicatorPtr& communicator)
         const_cast<ConnectionOptions&>(_clientConnectionOptions) = readConnectionOptions("Ice.Connection.Client");
         const_cast<ConnectionOptions&>(_serverConnectionOptions) = readConnectionOptions("Ice.Connection.Server");
 
+        // The max size of an Ice protocol message in bytes. The maximum value is 0x7fffffff, which is the maximum
+        // value for an int32_t.
+        const int32_t size_max = 0x7fffffff;
```
bernardnormier: We use camelCase for variable names, including constants.

Also, given the property name (MessageSizeMax), I would pick a different name, like:
```cpp
const int32_t messageSizeMaxUpperLimit = std::numeric_limits<int32_t>::max(); // in bytes
```

cpp/src/Ice/Instance.cpp
```diff
@@ -1071,11 +1071,21 @@ IceInternal::Instance::initialize(const Ice::CommunicatorPtr& communicator)
...
         {
             int32_t num = _initData.properties->getIcePropertyAsInt("Ice.MessageSizeMax");
-            if (num < 1 || static_cast<size_t>(num) > static_cast<size_t>(0x7fffffff / 1024))
+            if (static_cast<size_t>(num) > static_cast<size_t>(size_max / 1024))
             {
-                const_cast<size_t&>(_messageSizeMax) = static_cast<size_t>(0x7fffffff);
+                ostringstream os;
+                os << "Ice.MessageSizeMax '" << num << "' is too large, it must be less than or equal to '"
+                   << (size_max / 1024) << "'";
```
bernardnormier: I would include "KiB" in the message.

See https://en.wikipedia.org/wiki/Kilobyte

cpp/src/Ice/Instance.cpp
```diff
@@ -1095,13 +1105,16 @@ IceInternal::Instance::initialize(const Ice::CommunicatorPtr& communicator)
         else
         {
             int32_t num = _initData.properties->getIcePropertyAsInt("Ice.BatchAutoFlushSize"); // 1MB default
-            if (num < 1)
+            if (static_cast<size_t>(num) > static_cast<size_t>(size_max / 1024))
             {
-                const_cast<size_t&>(_batchAutoFlushSize) = static_cast<size_t>(num);
+                ostringstream os;
+                os << "Ice.BatchAutoFlushSize '" << num << "' is too large, it must be less than or equal to '"
```
bernardnormier, line 1102: KiB in the message, see above

cpp/src/Ice/Instance.cpp
```diff
@@ -1112,9 +1125,9 @@ IceInternal::Instance::initialize(const Ice::CommunicatorPtr& communicator)
 
         {
             int32_t num = _initData.properties->getIcePropertyAsInt("Ice.ClassGraphDepthMax");
-            if (num < 1 || static_cast<size_t>(num) > static_cast<size_t>(0x7fffffff))
+            if (num < 1)
             {
-                const_cast<size_t&>(_classGraphDepthMax) = static_cast<size_t>(0x7fffffff);
+                const_cast<size_t&>(_classGraphDepthMax) = static_cast<size_t>(size_max);
```
bernardnormier: This is unrelated to messageSizeMaxUpperLimit. Just use `std::numeric_limits<int32_t>::max()`.

Issue #4096, opened 2025-06-12
Update C++ mapping for dictionary<string, T>


bernardnormier:
I propose we change the default mapping to use `std::less<>` (a transparent comparator).

We already use this mapping for Ice::Context using a metadata directive, but we should do it by default for all `dictionary<string, T>` types.

PR #4099, opened 2025-06-12
Update MATLAB mapping for Slice dictionaries

### LINKED ISSUE #4079: Update MATLAB mapping for Slice dictionary
We should use dictionary instead of containers.Map:

https://www.mathworks.com/help/matlab/dictionary.html

bernardnormier:
They now map to MATLAB dictionaries. Fixes #4079.

There are still a few `containers.Map` in the matlab directory that I'll fix in separate PRs:
 - some are internal used for marshaling/unmarshaling
 - others are for the mapped PropertyDict type that lacks proper testing. Related issue: #4096.


matlab/src/Util.cpp
```diff
@@ -202,24 +226,29 @@ IceMatlab::getContext(mxArray* p, Ice::Context& m)
...
     {
         mxArray* params[1];
         params[0] = p;
-        mxArray* keys;
-        mexCallMATLAB(1, &keys, 1, params, "keys");
-        mxArray* values;
-        mexCallMATLAB(1, &values, 1, params, "values");
-        assert(mxGetM(keys) == 1 && mxGetM(values) == 1);
-        assert(mxGetN(keys) == mxGetN(values));
-        const size_t n = mxGetN(keys);
```
bernardnormier, line 188: We were getting a 1 x N cell array with `keys` and `values` on the containers.Map.
Now it's a M x 1 cell array returned by `stringToCellArray`.

matlab/src/Util.cpp
```diff
@@ -172,25 +192,29 @@ IceMatlab::createContext(const Ice::Context& m)
     mxArray* r;
```
bernardnormier, line 161: All this code is about converting a C++ context into a MATLAB context.

matlab/src/Util.cpp
```diff
@@ -202,24 +226,29 @@ IceMatlab::getContext(mxArray* p, Ice::Context& m)
     {
```
bernardnormier, line 199: Now it's about converting a MATLAB context into a C++ context.

matlab/test/Ice/operations/TwowaysAMI.m
```diff
@@ -414,149 +414,135 @@ function twowaysAMI(helper, p)
             assert(strcmp(r{3}{1}(2), ''));
             assert(strcmp(r{3}{2}(1), 'abcd'));
 
-            di1 = containers.Map('KeyType', 'int32', 'ValueType', 'logical');
+            di1 = configureDictionary('uint8', 'logical');
```
bernardnormier, line 417: It's basically all the same as in Twoways.m, except we make async calls.

CHANGELOG-3.8.md
```diff
@@ -694,6 +694,12 @@ initialization. See `InitializationData.pluginFactories`.
 
 ## MATLAB Changes
 
+- Changed the mapping for Slice dictionaries to MATLAB. A Slice dictionary now always maps to a MATLAB dictionary; the
```
externl: This first sentence is confusing. I reads like Slice dictionaries are not mapped to a type called "MATLAB". I would just remove the "to MATLAB"

externl (approved):

matlab/src/Util.cpp
```diff
@@ -202,24 +226,29 @@ IceMatlab::getContext(mxArray* p, Ice::Context& m)
...
         params[0] = p;
-        mxArray* keys;
-        mexCallMATLAB(1, &keys, 1, params, "keys");
-        mxArray* values;
-        mexCallMATLAB(1, &values, 1, params, "values");
-        assert(mxGetM(keys) == 1 && mxGetM(values) == 1);
-        assert(mxGetN(keys) == mxGetN(values));
-        const size_t n = mxGetN(keys);
+        mxArray* keysString;
+        mexCallMATLAB(1, &keysString, 1, params, "keys");
```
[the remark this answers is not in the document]
bernardnormier: already done

cpp/src/slice2matlab/Main.cpp
```diff
@@ -2446,18 +2410,23 @@ CodeVisitor::visitDictionary(const DictionaryPtr& p)
...
     {
-        //
-        // We use a struct array when the key is a structure type because we can't use containers.Map.
-        //
-        out << nl << "r = struct('key', {}, 'value', {});";
+        valueType = typeToString(value);
+    }
+    else if (cls)
+    {
+        valueType = "IceInternal.ValueHolder";
```
pepone: Looking at the code, we first create a dictionary where the value type is ValueHolder, and later, once the classes have been fully unmarshalled, we call convert to build a new dictionary where the value type is a cell array containing the resolved values.

I wonder if we could instead use a single dictionary from the beginning, with cell arrays as the value type. Initially, each cell would contain the corresponding ValueHolder, and once unmarshalling is complete, we would replace the ValueHolder in each cell with the final value. In theory, this would avoid allocating a second dictionary and copying all the keys.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -2558,47 +2516,39 @@ CodeVisitor::visitDictionary(const DictionaryPtr& p)
...
-        if (st)
+        out << nl << "r = configureDictionary('" << typeToString(key) << "', '"
+            << (scalarValue ? typeToString(value) : "cell") << "');";
+        out << nl << "keys = d.keys;";
+        out << nl << "values = d.values;";
+        out << nl << "for i = 1:d.numEntries";
+        out.inc();
+        out << nl << "k = keys(i);";
+
+        if (cls || scalarValue) // the temporary ValueHolder is a scalar too
```
pepone: Doesn't scalar refers to value classes? ValueHolder is not a value class `classdef ValueHolder < handle`

matlab/src/Util.cpp
```diff
@@ -42,6 +42,30 @@ namespace
...
+
+    // This function consumes the cell array input argument and returns a MATLAB string array.
+    mxArray* cellArrayToString(mxArray* cellArray)
+    {
+        mxArray* stringArray;
+        mxArray* params[1];
+        params[0] = cellArray;
+        mexCallMATLAB(1, &stringArray, 1, params, "string");
+
+        mxDestroyArray(cellArray);
```
pepone: I think is clear if the caller take care of the destroy, which would be near the creation.

matlab/src/Util.cpp
```diff
@@ -42,6 +42,30 @@ namespace
...
+
+    // This function consumes the string array input argument and returns a cell array of char arrays.
+    mxArray* stringToCellArray(mxArray* stringArray)
+    {
+        mxArray* cellArray;
+        mxArray* params[1];
+        params[0] = stringArray;
+        mexCallMATLAB(1, &cellArray, 1, params, "cellstr");
+
+        mxDestroyArray(stringArray);
```
pepone: Is the destroy call correct here? Isn't this a reference to the context keys/values pass by the application?

matlab/src/Util.cpp
```diff
@@ -172,25 +196,29 @@ IceMatlab::createContext(const Ice::Context& m)
...
     else
     {
         mwSize dims[2] = {1, 0};
         dims[1] = static_cast<int>(m.size());
-        auto keys = mxCreateCellArray(2, dims);
-        auto values = mxCreateCellArray(2, dims);
+        auto keysCell = mxCreateCellArray(2, dims);
+        auto valuesCell = mxCreateCellArray(2, dims);
         int idx = 0;
         for (auto p : m)
```
pepone: `const auto& p`

cpp/src/slice2matlab/Main.cpp
bernardnormier: See:
https://www.mathworks.com/help/matlab/ref/double.isscalar.html

and

https://www.mathworks.com/help/matlab/matlab_prog/getting-started-with-dictionaries.html
"Store Different Data Types in Dictionary"

When this helper function returns true, we store the mapped value as a scalar (1x1 array) in the dictionary. When this helper function returns false, we wrap the mapped value in a cell, because the type is "complex" and not homogeneous ... can't use a scalar.

dictionary provides different syntaxes for scalar and cell values:
```
dict(key) % get/set scalar value
dict{key} % get/set cell value with wrap/unwrap
```

matlab/src/Util.cpp
bernardnormier: All these calls return a `mxArray*`. It would be clearer to use `mxArray*`.

pepone: My comment was about the loop, not changed in your PR.

matlab/src/Util.cpp
bernardnormier: Moved the destroy to the caller.

matlab/src/Util.cpp
bernardnormier: The application transmit the dictionary, and we retrieve the keys and values arrays with a `mexCallMATLAB` which presumably allocates a new array.

I moved the destroy to the caller too.

matlab/src/Util.cpp
bernardnormier: Fixed for loop.

matlab/src/Util.cpp
```diff
@@ -172,25 +196,29 @@ IceMatlab::createContext(const Ice::Context& m)
     mxArray* r;
     if (m.empty())
     {
-        mexCallMATLAB(1, &r, 0, 0, "containers.Map");
+        mxArray* params[2];
+        params[0] = mxCreateString("char");
+        params[1] = params[0];
+        mexCallMATLAB(1, &r, 2, params, "configureDictionary");
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested, except with a single destroy.

cpp/src/slice2matlab/Main.cpp
bernardnormier: I updated the code as suggested: convert now rewrites the input dictionary (instead of creating a new dictionary) and ValueHolder is put in a cell.

CHANGELOG-3.8.md
bernardnormier: Fixed as suggested.

matlab/src/Util.cpp
```diff
@@ -141,25 +161,33 @@ IceMatlab::createStringMap(const map<string, string, std::less<>>& m)
...
         mxArray* params[2];
-        params[0] = keys;
-        params[1] = values;
-        mexCallMATLAB(1, &r, 2, params, "containers.Map");
+        params[0] = cellArrayToString(keysCell);
+        params[1] = cellArrayToString(valuesCell);
+        mxDestroyArray(keysCell);
+        mxDestroyArray(valuesCell);
+
+        mexCallMATLAB(1, &r, 2, params, "dictionary");
```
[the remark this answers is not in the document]
bernardnormier: The code is fine as is.

matlab/src/Util.cpp
```diff
@@ -171,24 +199,32 @@ IceMatlab::getContext(mxArray* p, Ice::Context& m)
...
+        mexCallMATLAB(1, &valuesString, 1, params, "values");
+        mxArray* values = stringToCellArray(valuesString);
+        mxDestroyArray(valuesString);
+
+        assert(mxGetM(keys) == mxGetM(values));
+        assert(mxGetN(keys) == 1 && mxGetN(values) == 1);
+
+        const size_t size = mxGetM(keys);
         try
         {
```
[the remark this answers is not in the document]
bernardnormier: keys and values are freed (destroyed) on lines 233, 234, 238 and 239.

pepone (approved):
Looks good!

cpp/src/slice2matlab/Main.cpp
bernardnormier: Actually, it doesn't rewrite the input dictionary as far as I can tell. The input dictionary remains unchanged. `r = d` makes a copy.  Maybe we could improve this convert to convert in-place (for a separate PR).

PR #4100, opened 2025-06-12
Replace containers.Map by dictionary in MATLAB marshaling code

bernardnormier:
This PR replaces all containers.Map by dictionary in the internal marshaling/unmarshaling code, except in the EncapsEncoder10 class.

The EncapsEncoder10 code is complicated, rarely used, and I'd rather not break it by accident.

### DIFF
--- a/matlab/lib/+IceInternal/EncapsDecoder11_InstanceData.m
+++ b/matlab/lib/+IceInternal/EncapsDecoder11_InstanceData.m
@@ -10,7 +10,6 @@ classdef EncapsDecoder11_InstanceData < handle
             obj.slices = {};
             obj.indirectionTables = {};
             obj.sliceFlags = uint8(0);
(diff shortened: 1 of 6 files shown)


matlab/lib/+IceInternal/EncapsDecoder11_InstanceData.m
```diff
@@ -23,7 +22,7 @@
         typeId
         typeIdIndex
         compactId
-        indirectPatchList
+        indirectPatchList dictionary = dictionary % unconfigured dictionary
```
bernardnormier, line 25: The property `unmarshaledMap` is an array, while the property `indirectPatchList` is a dictionary.

MATLAB dictionaries have a built-in "not valid" value, the unconfigured dictionary.

pepone (approved):

externl (approved):

PR #4101, opened 2025-06-12
Add transparent comparator (std::less<>) to C++ maps

bernardnormier:
This PR adds `std::less<>` to a few additional `std::map<std::string, ...>`.

This indirectly removes duplicated code in PHP and MATLAB.

### DIFF
--- a/cpp/src/Ice/HttpParser.cpp
+++ b/cpp/src/Ice/HttpParser.cpp
@@ -672,10 +672,10 @@ IceInternal::HttpParser::getHeader(const string& name, string& value, bool toLow
     return false;
 }
 
-map<string, string>
+map<string, string, std::less<>>
 IceInternal::HttpParser::getHeaders() const
 {
-    map<string, string> headers;
+    map<string, string, std::less<>> headers;
     for (const auto& header : _headers)
     {
         headers.insert(make_pair(header.second.first, IceInternal::trim(header.second.second)));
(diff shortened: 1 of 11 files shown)


cpp/src/Ice/HttpParser.h
```diff
@@ -50,7 +50,7 @@ namespace IceInternal
 
         bool getHeader(const std::string&, std::string&, bool) const;
 
-        [[nodiscard]] std::map<std::string, std::string> getHeaders() const;
+        [[nodiscard]] std::map<std::string, std::string, std::less<>> getHeaders() const;
```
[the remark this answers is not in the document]
bernardnormier: Doesn't work well since we don't include Ice headers in the file.

cpp/src/Ice/HttpParser.cpp
```diff
@@ -672,10 +672,10 @@ IceInternal::HttpParser::getHeader(const string& name, string& value, bool toLow
     return false;
 }
 
-map<string, string>
+map<string, string, std::less<>>
```
pepone, line 675: Should we use `HeaderDict`  from Connection.h?

pepone (approved):

externl (approved):

cpp/src/Ice/HttpParser.cpp
bernardnormier: I didn't because `HttpParser.{h,cpp}` barely includes any Ice code, only Ice/Config.h and Ice/StringUtil.h in the .cpp file.

PR #4111, opened 2025-06-13
Change MATLAB mapping for sequence<string>

bernardnormier:
to string array, for consistency with the dictionary mapping.

This is highly compatible with the old mapping. I could have left all the tests unchanged.

I also updated the various Ice APIs (Ice.initialize, Ice.createProperties, ...) that accept or return a `sequence<string>`: they now accept either a string array or a cell array of char; they return a string array.


matlab/lib/+Ice/InputStream.m
```diff
@@ -298,11 +298,11 @@ function reset(obj, data)
                 obj.pos = pos + sz;
             end
         end
-        function r = readStringSeq(obj)
+        function r = readStringSeq(obj) % returns a string array
             sz = obj.readAndCheckSeqSize(1);
-            r = cell(1, sz);
+            r = strings(sz); % preallocate array
```
bernardnormier: Should be 1, sz

bernardnormier: Fixed.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -2154,7 +2154,7 @@ CodeVisitor::visitSequence(const SequencePtr& p)
         out.dec();
         out << nl << "end";
     }
-    else if ((b && b->kind() == Builtin::KindString) || dictContent || seqContent || proxy)
```
bernardnormier, line 2157: This was unused confusing code. We return a number of lines above when `b->kind() == Builtin::KindString`

matlab/lib/+Ice/OutputStream.m
```diff
@@ -257,7 +257,7 @@ function writeStringSeq(obj, v)
             sz = length(v);
             obj.writeSize(sz);
             for i = 1:sz
-                obj.writeString(v{i});
+                obj.writeString(char(v(i)));
```
bernardnormier, line 260: When v is a string array, `char` extracts the char array.
When v is a cell array of char, `v(i)` returned a 1x1 cell and `char` gets the char array out of the cell.

matlab/lib/+Ice/createProperties.m
```diff
@@ -21,19 +21,19 @@
...
     if length(varargin) >= 1 && ~isempty(varargin{1})
         args = varargin{1};
     else
-        args = {};
+        args = [];
     end
     if length(varargin) >= 2 && ~isempty(varargin{2})
         if ~isa(varargin{2}, 'Ice.Properties')
-            throw(LocalException('Ice:ArgumentException', 'expecting Ice.Properties object'));
+            throw(Ice.LocalException('Ice:ArgumentException', 'expecting Ice.Properties object'));
```
bernardnormier, line 36: All the `throw(LocalException` are bogus. Will fix in a separate PR.

matlab/src/Util.cpp
```diff
@@ -47,19 +47,15 @@ namespace
...
+        mexCallMATLAB(1, &stringArray, 1, &cellArray, "string");
         return stringArray;
     }
 
-    // This function converts the string array input argument and returns a cell array of char arrays.
+    // This function converts the string array input argument and returns a cell array of char.
     mxArray* stringToCellArray(mxArray* stringArray)
     {
         mxArray* cellArray;
-        mxArray* params[1];
```
bernardnormier, line 60: We don't need a params with a single argument: we can just take the reference of the single param.

matlab/src/Util.cpp
```diff
@@ -169,10 +165,9 @@ IceMatlab::createStringMap(const map<string, string, std::less<>>& m)
     }
     else
     {
-        mwSize dims[2] = {1, 0};
-        dims[1] = static_cast<int>(m.size());
-        auto keysCell = mxCreateCellArray(2, dims);
-        auto valuesCell = mxCreateCellArray(2, dims);
+        mwSize size = static_cast<int>(m.size());
+        auto keysCell = mxCreateCellMatrix(1, size);
```
bernardnormier, line 169: Simpler syntax for 2-dimensional cell arrays

matlab/lib/+Ice/InputStream.m
```diff
@@ -298,11 +298,11 @@ function reset(obj, data)
...
             end
         end
-        function r = readStringSeq(obj)
+        function r = readStringSeq(obj) % returns a string array
             sz = obj.readAndCheckSeqSize(1);
-            r = cell(1, sz);
+            r = strings(1, sz); % preallocate array
             for i = 1:sz
-                r{i} = obj.readString();
+                r(i) = obj.readString();
```
pepone, line 305: is there an implicit conversion here?

pepone (approved):

matlab/lib/+Ice/InputStream.m
bernardnormier: Yes, `strings` creates a string array and when you assign a char array into a string array "slot", it's converted to a string.

PR #4112, opened 2025-06-14
Add type to more mapped properties (MATLAB)

bernardnormier:
This PR updates mapped properties and related code in MATLAB:
 - all properties are now typed except properties mapped from optional fields and fields that use classes
 - updated default value for struct properties - it's now an empty array instead of a new instance
 - all properties now have default values, and ctor no longer assign the default values
 - removed the "NoInit" path in struct/class constructors


matlab/lib/+Ice/InputStream.m
```diff
@@ -647,14 +647,21 @@ function skipOptional(obj, format)
...
 
-            if isNull(impl) % decoded a null proxy
-                r = [];
-            elseif nargin == 2
-                % Instantiate a proxy of the requested type.
-                constructor = str2func(cls);
-                r = constructor(obj.communicator, '', impl);
+            if nargin == 2
+                if isNull(impl)
+                    emptyFunc = str2func(strcat(cls, '.empty'));
```
bernardnormier, line 652: Returns an empty array of the proxy type.

matlab/lib/+Ice/InputStream.m
```diff
@@ -687,7 +697,11 @@ function readValue(obj, cb, formalType)
                 obj.initEncaps();
             end
             function check(v)
-                if isempty(v) || isa(v, formalType)
+                if isempty(v)
+                    % Pass empty array of the formalType.
+                    emptyFunc = str2func(strcat(formalType, '.empty'));
```
bernardnormier, line 702: It's nice a thing to do (return the correct type). In practice, it doesn't really matter since we can't give a type to "class" properties (= mapped for field with a Slice type) because we always convert this property in place for struct and exception properties.

The logic is a bit different for classes, and I didn't want to refactor this complicated "convert" code.

matlab/test/Ice/defaultValue/AllTests.m
```diff
@@ -107,14 +107,18 @@ function allTests(helper)
             assert(v.c1 == Color.red);
             assert(isempty(v.bs));
             assert(isempty(v.is));
-            assert(~isempty(v.st));
+            assert(isempty(v.st));
             assert(v.dict.numEntries == 0);
 
+            % Check we can assign empty into bs and is.
```
bernardnormier, line 113: It's actually not completely obvious it works: the mapped sequence is constrained by `(1,:)`, and yet accepts this 0x0 empty array.

matlab/test/Slice/escape/AllTests.m
```diff
@@ -36,11 +36,11 @@ function allTests(helper)
...
 
             c = classdef_.logical_();
+            defaultFor = classdef_.global_();
             assert(c.else_ == classdef_.persistent_.break_);
-            assert(c.for_.case_ == classdef_.persistent_.catch_);
-            assert(c.for_.continue_ == 1);
-            assert(c.for_.eq_ == 2);
+            assert(isempty(c.for_));
             assert(c.int64 == true);
+            % Even though c.for_ is empty, we can still marshal it.
```
bernardnormier, line 43: I kept the existing logic: an empty mapped-struct array is marshaled as a default instance of the mapped struct.

matlab/test/Slice/escape/AllTests.m
```diff
@@ -98,15 +97,15 @@ function allTests(helper)
             fprintf('testing exception... ');
 
             e = classdef_.bitand_();
-            assert(isempty(e.identifier_));
-            assert(isempty(e.message_));
+            assert(strcmp(e.identifier_, '1'));
```
bernardnormier, line 100: I mapped default values for all Slice fields, including exception fields. Previously, these default values were ignored for exceptions fields (all very minor since MATLAB is client-only).

cpp/src/slice2matlab/Main.cpp
```diff
@@ -332,32 +323,7 @@ namespace
 
     bool needsConversion(const TypePtr& type)
     {
-        SequencePtr seq = dynamic_pointer_cast<Sequence>(type);
```
bernardnormier, line 335: This was a complicated way to write "uses classes but is not a class".

CHANGELOG-3.8.md
```diff
@@ -709,6 +709,17 @@ initialization. See `InitializationData.pluginFactories`.
 
   (1) A single Slice string maps to a MATLAB char array, not to a MATLAB string.
 
+- All fields are not mapped to typed MATLAB properties except optional fields and fields whose type is a class or uses
```
pepone:
```suggestion
- All fields are now mapped to typed MATLAB properties except optional fields and fields whose type is a class or uses
```

matlab/test/Slice/escape/AllTests.m
```diff
@@ -36,11 +36,11 @@ function allTests(helper)
             fprintf('testing class... ');
 
             c = classdef_.logical_();
+            defaultFor = classdef_.global_();
```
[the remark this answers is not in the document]
bernardnormier: It's just a variable.

matlab/lib/+Ice/ConnectionClosedException.m
```diff
@@ -10,20 +10,20 @@
 classdef ConnectionClosedException < Ice.LocalException
     properties
         % ConnectionClosedException - True if the connection was closed by the application, false if it was closed by the Ice runtime.
-        closedByApplication logical
+        closedByApplication logical = false
```
bernardnormier: Synchronized with the generated code for classes.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -288,10 +268,10 @@ namespace
                     case Builtin::KindFloat:
                     case Builtin::KindDouble:
                         return "0";
-                    case Builtin::KindObject:
                     case Builtin::KindObjectProxy:
-                    case Builtin::KindValue:
-                        return "[]";
+                        return "Ice.ObjectPrx.empty";
+                    default: // Object, Value
+                        return "Ice.UnknownSlicedValue.empty";
```
pepone, line 274: Why not `Ice.Value.empty`?

matlab/lib/+Ice/RequestFailedException.m
```diff
@@ -14,28 +14,26 @@
 classdef RequestFailedException < Ice.DispatchException
     properties
         % id - The identity of the Ice Object to which the request was sent.
-        id Ice.Identity
+        id (1, :) Ice.Identity = Ice.Identity.empty
```
pepone, line 17: Why do we we use `(1, :)` for this property? Doesn't this mean we accept a list?

What about:

```
id Ice.Identity {mustBeScalarOrEmpty} = Ice.Identity.empty
```

pepone (approved):

cpp/src/slice2matlab/Main.cpp
```diff
@@ -288,10 +268,10 @@ namespace
                     case Builtin::KindFloat:
                     case Builtin::KindDouble:
                         return "0";
-                    case Builtin::KindObject:
                     case Builtin::KindObjectProxy:
-                    case Builtin::KindValue:
-                        return "[]";
+                        return "Ice.ObjectPrx.empty";
+                    default: // Object, Value
```
externl, line 273: I know this works but I find it clearer if we case statements instead of relying on default.

externl (approved):

InsertCreativityHere (approved):
Looks good to me.

cpp/src/slice2matlab/Main.cpp
bernardnormier: Because `Ice.Value` is abstract: empty only works with concrete types.

matlab/lib/+Ice/RequestFailedException.m
bernardnormier: Fixed as suggested.

cpp/src/slice2matlab/Main.cpp
bernardnormier: Fixed as suggested.

PR #4115, opened 2025-06-16
Use implicit default for MATLAB properties

bernardnormier:
This PR updates the Slice compiler for MATLAB to use implicit default value for properties other than:
 - properties mapped from fields with an explicit default value
 - untyped properties (class and "use class" properties, and properties mapped from optional fields)
 - dictionary properties
 - struct properties (not strictly necessary but clear since they have no dimension)


externl (approved):

pepone (approved):

InsertCreativityHere (approved):
Looks good to me again!

PR #4118, opened 2025-06-17
Add Support for '@remarks' to the Java Mapping

InsertCreativityHere:
Unfortunately, Java has no support for `@remarks`, or anything even resembling it.
The best we can do is append it to the end of the summary.

Here's what the new generated code looks like for `@remarks`, and what it looks like in the generated API reference:

----

This is what `EncodingVersion` looks like now:

```diff
 /**
  * Represents a version of the Ice encoding. Ice supports version 1.0 and 1.1 of this encoding.
+ * <p><b>Remarks:</b>
+ * The Ice encoding is also known as the Slice encoding.</p>
  **/
 public final class EncodingVersion implements java.lang.Cloneable,
```
image

----

----

And this is what `RemoteLoggerPrx.log` looks like now:

```diff
     /**
      * Logs a LogMessage.
+     * <p><b>Remarks:</b>
+     * {@link com.zeroc.Ice.RemoteLoggerPrx#log} may be called by {@link com.zeroc.Ice.LoggerAdminPrx} before {@link com.zeroc.Ice.RemoteLoggerPrx#init}.</p>
      * @param message The message to log.
      * @param context The Context map to send with the invocation.
      **/
     default void log(LogMessage message, java.util.Map<String, String> context)
```
image
Compared to before this PR:
image

### DIFF
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -1315,7 +1315,7 @@ Slice::JavaVisitor::writeRemarksDocComment(Output& out, const DocComment& commen
     }
 
     out << nl << " * <p><b>Remarks:</b>";
-    out << " * ";
+    out << nl << " * ";
     writeDocCommentLines(out, remarks);
     out << "</p>";
 }
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -699,7 +699,7 @@ Slice::JavaVisitor::writeSyncIceInvokeMethods(
 
     // Generate a synchronous version of this operation which doesn't takes a context parameter.
     out << sp;
-    writeProxyDocComment(out, p, package, dc, false, "");
+    writeProxyOpDocComment(out, p, package, dc, false, "");
     if (dc && dc->isDeprecated())
     {
         out << nl << "@Deprecated";
@@ -717,7 +717,7 @@ Slice::JavaVisitor::writeSyncIceInvokeMethods(
 
     // Generate a synchronous version of this operation which takes a context parameter.
     out << sp;
-    writeProxyDocComment(out, p, package, dc, false, contextDoc);
+    writeProxyOpDocComment(out, p, package, dc, false, contextDoc);
     if (dc && dc->isDeprecated())
     {
         out << nl << "@Deprecated";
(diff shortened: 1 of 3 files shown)


cpp/src/slice2java/Gen.cpp
```diff
@@ -699,7 +699,7 @@ Slice::JavaVisitor::writeSyncIceInvokeMethods(
 
     // Generate a synchronous version of this operation which doesn't takes a context parameter.
     out << sp;
-    writeProxyDocComment(out, p, package, dc, false, "");
```
InsertCreativityHere, line 702: Not essential to the PR, but this name was just bad.
It writes the doc-comment for a proxy operation, not the proxy itself.

cpp/src/slice2java/Gen.cpp
```diff
@@ -1592,7 +1609,7 @@ Slice::JavaVisitor::writeHiddenProxyDocComment(Output& out, const OperationPtr&
 }
 
 void
-Slice::JavaVisitor::writeServantDocComment(Output& out, const OperationPtr& p, const string& package, bool async)
```
InsertCreativityHere, line 1595: Same, this writes the comment for a servant method, not a servant. Bad name.

java/build.gradle
```diff
@@ -80,7 +80,10 @@ task alljavadoc(type: Javadoc) {
     destinationDir = file("$buildDir/docs/javadoc")
     options.encoding = 'UTF-8'
     // Where to find source files for the different modules
-    options.addStringOption('-module-source-path', './src/*/src/main/java/:./src/*/build/generated/source/slice/main')
```
InsertCreativityHere, line 83: Using a `:` separator only works on Unix platforms. It's better to use `File.pathSeparator`, which will use the correct separator for the current platform.

Without this fix, Javadoc generation was broken on Windows.

externl (approved):

cpp/src/slice2java/Gen.cpp
```diff
@@ -1305,6 +1305,21 @@ Slice::JavaVisitor::writeExceptionDocComment(Output& out, const OperationPtr& op
...
+    const StringList& remarks = comment.remarks();
+    if (remarks.empty())
+    {
+        return;
+    }
+
+    out << nl << " * <p><b>Remarks:</b>";
+    out << nl << " * ";
+    writeDocCommentLines(out, remarks);
+    out << "</p>";
```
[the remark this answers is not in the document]
InsertCreativityHere: We always write these tags in-line, and I'm pretty sure the formatter complained once when I didn't.

pepone (approved):

bernardnormier (approved):

PR #4119, opened 2025-06-17
Fix Random Minor Issues in `slice2js` and `NOLINT`s

InsertCreativityHere:
While planning out how to add `@remarks` to (Java/Type)Script, I came across a few inconsistencies, that I'd rather fix in a PR separate to the `@remarks` one.

There were also 2 formatting mistakes in our NOLINTs that I've been aware of a little while.
Figured I might as well fix them now too.

### DIFF
--- a/cpp/src/Ice/StringUtil.cpp
+++ b/cpp/src/Ice/StringUtil.cpp
@@ -4,7 +4,7 @@
 #include "Ice/StringConverter.h"
 #include <cassert>
 #include <cstring>
-#include <string.h> // NOLINT(modernize-deprecated-headers):for strerror_r
+#include <string.h> // NOLINT(modernize-deprecated-headers): for strerror_r
 
 #include <iomanip>
 #include <sstream>
--- a/cpp/src/slice2cpp/Gen.cpp
+++ b/cpp/src/slice2cpp/Gen.cpp
@@ -1371,7 +1371,7 @@ Slice::Gen::ForwardDeclVisitor::visitConst(const ConstPtr& p)
     if (!isConstexprType(p->type())) // i.e. string or wstring
     {
         // The string/wstring constructor can throw, which produces a clang-tidy lint for const or static objects.
-        H << " // NOLINT(cert-err58-cpp,modernize-raw-string-literal)";
+        H << " // NOLINT(cert-err58-cpp, modernize-raw-string-literal)";
     }
 }
 
--- a/cpp/src/slice2js/Gen.cpp
+++ b/cpp/src/slice2js/Gen.cpp
@@ -245,12 +245,6 @@ Slice::JsVisitor::JsVisitor(Output& out) : _out(out) {}
 
 Slice::JsVisitor::~JsVisitor() = default;
 
-vector<pair<string, string>>
-Slice::JsVisitor::imports() const
-{
-    return _imports;
-}
-
 void
 Slice::JsVisitor::writeMarshalDataMembers(const DataMemberList& dataMembers, const DataMemberList& optionalMembers)
 {
--- a/cpp/src/slice2js/Gen.cpp
+++ b/cpp/src/slice2js/Gen.cpp
@@ -2198,7 +2198,9 @@ Slice::Gen::TypeScriptVisitor::visitClassDefStart(const ClassDefPtr& p)
     {
         _out << sp;
         writeDocCommentFor(dataMember);
-        _out << nl << dataMember->mappedName() << ": " << typeToTsString(dataMember->type(), true) << ";";
+        const string optionalModifier = dataMember->optional() ? "?" : "";
+        _out << nl << dataMember->mappedName() << optionalModifier << ": " << typeToTsString(dataMember->type(), true)
+             << ";";
     }
     _out << eb;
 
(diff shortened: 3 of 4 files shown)


cpp/src/slice2js/Gen.cpp
```diff
@@ -241,7 +241,7 @@ namespace
     }
 }
 
-Slice::JsVisitor::JsVisitor(Output& out, const vector<pair<string, string>>& imports) : _out(out), _imports(imports) {}
```
InsertCreativityHere, line 244: We never passed `imports` to the constructor, it was always default-initialized.

cpp/src/slice2js/Gen.cpp
```diff
@@ -2558,13 +2558,17 @@ Slice::Gen::TypeScriptVisitor::visitExceptionStart(const ExceptionPtr& p)
         _out << epar << ";";
     }
 
-    for (const auto& dataMember : allDataMembers)
```
InsertCreativityHere, line 2561: I'm pretty sure this is supposed to be `dataMembers`.
Using `allDataMembers` here, means that derived exceptions will always shadow their parent's fields.
For example, this Slice:
```
    exception BaseException { int myInt; string myString; }
    exception DerivedException : BaseException { bool myBool; string myOtherString; }
```
currently emits this code on main:
```
    export class BaseException extends __module__zeroc_ice.Ice.UserException
    {
        /**
         * One-shot constructor to initialize all data members.
         */
        constructor(myInt?: number, myString?: string);
        myInt: number;
        myString: string;
    }

    export class DerivedException extends __global_Foo.BaseException
    {
        /**
         * One-shot constructor to initialize all data members.
         */
        constructor(myInt?: number, myString?: string, myBool?: boolean, myOtherString?: string);
        myInt: number;
        myString: string;
        myBool: boolean;
        myOtherString: string;
    }
```

cpp/src/slice2js/Gen.cpp
```diff
@@ -2558,13 +2558,17 @@ Slice::Gen::TypeScriptVisitor::visitExceptionStart(const ExceptionPtr& p)
...
+    for (const auto& dataMember : dataMembers)
     {
+        _out << sp;
+        writeDocCommentFor(dataMember);
```
InsertCreativityHere, line 2560: We weren't emitting doc-comments for exception members.
I see no reason for this, since we're emitting them just fine for structs and classes.

cpp/src/slice2js/Gen.cpp
```diff
@@ -2636,10 +2640,10 @@ Slice::Gen::TypeScriptVisitor::visitStructStart(const StructPtr& p)
     {
         _out << sp;
         writeDocCommentFor(dataMember);
-        _out << nl << dataMember->mappedName() << ":" << typeToTsString(dataMember->type(), true) << ";";
```
InsertCreativityHere, line 2639: Missing space after the colon here.
On main, we're generating this (which is incorrect):
```
    export class MyStruct
    {
        constructor(MyInt?: number, myString?: string);
        // ...

        MyInt:number;

        myString:string;
    }
```

cpp/src/slice2js/Gen.h
```diff
@@ -31,7 +28,7 @@ namespace Slice
 
         ::IceInternal::Output& _out;
 
-        std::vector<std::pair<std::string, std::string>> _imports;
```
InsertCreativityHere, line 34: ~~Since we never pass `_imports` into the constructor, we may as well just default initialize it like this.~~

After a closer look, `_imports` was totally unused. So I just deleted it instead.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1371,7 +1371,7 @@ Slice::Gen::ForwardDeclVisitor::visitConst(const ConstPtr& p)
     if (!isConstexprType(p->type())) // i.e. string or wstring
     {
         // The string/wstring constructor can throw, which produces a clang-tidy lint for const or static objects.
-        H << " // NOLINT(cert-err58-cpp,modernize-raw-string-literal)";
```
InsertCreativityHere, line 1374: There should be a comma between lints.
We do this correctly everywhere else I could find.

cpp/src/Ice/StringUtil.cpp
```diff
@@ -4,7 +4,7 @@
 #include "Ice/StringConverter.h"
 #include <cassert>
 #include <cstring>
-#include <string.h> // NOLINT(modernize-deprecated-headers):for strerror_r
```
InsertCreativityHere, line 7: There should be a space after the colon here.
We do this correctly everywhere else I could find.

cpp/src/slice2js/Gen.cpp
InsertCreativityHere: Nevermind, it turns out that this `_imports` thing was completely dead code.
So I just completely removed it instead...

cpp/src/slice2js/Gen.cpp
```diff
@@ -2558,13 +2552,17 @@ Slice::Gen::TypeScriptVisitor::visitExceptionStart(const ExceptionPtr& p)
         _out << epar << ";";
     }
 
-    for (const auto& dataMember : allDataMembers)
+    for (const auto& dataMember : dataMembers)
     {
+        _out << sp;
+        writeDocCommentFor(dataMember);
+        // TODO why do we only have this optional check for exception members, but not for class members?
```
pepone: We should add this `?:` to classes optional members too.

cpp/src/slice2js/Gen.cpp
pepone: The space is not mandatory, but is the most common.

cpp/src/slice2js/Gen.cpp
InsertCreativityHere: I copied this logic from exception to class.
So now both of them should be correctly marking their optional members.

PR #4122, opened 2025-06-18
Add Support for `@remarks` to `slice2py` and Fix Enum Doc-Comments

InsertCreativityHere:
This PR does what it says on the tin!

# Support for `@remarks`

We use a tool named `sphinx` for generating Python documentation, and sphinx supports a `Notes` section.
This is what we map `@remarks` to in Python, because it fits the bill pretty well.
(https://the-ultimate-sphinx-tutorial.readthedocs.io/en/latest/_guide/_styleguides/docstrings-guidelines.html#notes)
The documentation only explicitly mentions it being used to document methods and modules, but it seems to render just fine everywhere where I tried it.

We always generate the notes at the end of the doc-comment, which matches our convention, and sphinx's.

Here's what the rendered notes sections look like:

----

for a function:
Screenshot 2025-06-18 120828

----

for anything else:
Screenshot 2025-06-18 120843

----

Definitely doesn't look... amazing, but that's an issue with the HTML output, not the generated-code.

# Enum Doc-Comments.
On main, this is what the rendered output for a Slice-generated enum looks like:

----

Screenshot 2025-06-18 120925

----

This is what my PR makes it look like:

----

Screenshot 2025-06-18 120943

----

I'd argue it looks much better now.
For the code we document by hand, we don't seem to have a consistent style, I fixed 2 of ours, just while testing, but we should really look through all of our Python doc-comments and fix them. I'm sure these issues aren't exclusive to enums...

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -371,6 +371,9 @@ namespace Slice::Python
         /// Write constructor parameters with default values.
         void writeConstructorParams(const DataMemberList& members);
 
+        /// Writes the provided @p remarks in its own subheading in the current comment (if @p remarks is non-empty).
+        void writeRemarksDocComment(const StringList& remarks, bool needsNewline);
+
         void writeDocstring(const optional<DocComment>&, const string& = "");
         void writeDocstring(const optional<DocComment>&, const DataMemberList&);
         void writeDocstring(const optional<DocComment>&, const EnumeratorList&);
@@ -1975,20 +1978,44 @@ Slice::Python::CodeVisitor::getOperationMode(Slice::Operation::Mode mode)
 }
 
 void
+Slice::Python::CodeVisitor::writeRemarksDocComment(const StringList& remarks, bool needsNewline)
+{
+    if (!remarks.empty())
+    {
+        if (needsNewline)
+        {
+            _out << nl;
+        }
+        _out << nl << "Notes";
+        _out << nl << "-----";
+        for (const auto& line : remarks)
+        {
+            _out << nl << "    " << line;
+        }
+    }
+}
+
+void
 Slice::Python::CodeVisitor::writeDocstring(const optional<DocComment>& comment, const string& prefix)
 {
     if (comment)
     {
         const StringList& overview = comment->overview();
-        if (!overview.empty())
+        const StringList& remarks = comment->remarks();
+        if (overview.empty() && remarks.empty())
         {
-            _out << nl << prefix << tripleQuotes;
-            for (const auto& line : overview)
-            {
-                _out << nl << line;
-            }
-            _out << nl << tripleQuotes;
+            return;
         }
+
+        _out << nl << prefix << tripleQuotes;
+        for (const auto& line : overview)
+        {
+            _out << nl << line;
+        }
+
+        writeRemarksDocComment(remarks, !overview.empty());
+
+        _out << nl << tripleQuotes;
     }
 }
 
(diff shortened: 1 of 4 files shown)


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2098,22 +2129,21 @@ Slice::Python::CodeVisitor::writeDocstring(const optional<DocComment>& comment,
...
-                for (auto q = p->second.begin(); q != p->second.end(); ++q)
+                _out << ":"; // Only emit a trailing ':' if there's documentation to emit for it.
+                for (const auto& line : p->second)
                 {
-                    if (q != p->second.begin())
-                    {
-                        _out << nl;
-                    }
-                    _out << *q;
+                    _out << nl << "    " << line;
```
InsertCreativityHere, line 2139: Unrelated fix so enumerator documentation is output correctly.

python/docs/conf.py
```diff
@@ -6,6 +6,8 @@
 import os
 import sys
 sys.path.insert(0, os.path.abspath('../python'))
+if sys.platform == 'win32':
+    sys.path.insert(0, os.path.abspath('../python/x64/Release'))
```
InsertCreativityHere, line 10: This was necessary to get the doc-build (`make html` with `sphinx`) to work on Windows.
Because Windows outputs the `IcePy` module into these nested folders you know.

python/python/Ice/CompressBatch.py
```diff
@@ -6,13 +6,15 @@ class CompressBatch(EnumBase):
     """
     The batch compression option when flushing queued batch requests.
```
InsertCreativityHere, line 8: Fixed this to be consistent with the generated code.

`Enumerators` isn't a 'standard' heading, and as far as I can tell, there is no convention or standard for writing the type (`CompressBatch`) after each enumerator. This seems to be something we invented (or copied from another language...)

python/python/Ice/ToStringMode.py
```diff
@@ -10,16 +10,19 @@ class ToStringMode(EnumBase):
 
     Enumerators:
```
InsertCreativityHere, line 11: Made it consistent with the generated code. There's probably other hand-written enums that need fixing too.
I stopped after 2 because I was using these as test dummies while figuring out how to make enums render better.

InsertCreativityHere:
In case anyone cares, these 4 documentation functions cover all the generated code.
Here specifically, is what each function handles doc-comments for:
```
- writeDocString<string> (
    CodeVisitor::visitModuleStart # module
)
- writeDocString<DataMemberList> (
    CodeVisitor::visitClassDefStart # class & data-member
    CodeVisitor::visitExceptionStart # exception & data-member
    CodeVisitor::visitStructStart # struct & data-member
)
- writeDocString<EnumeratorList> (
    CodeVisitor::visitEnum # enum and enumerators
)
- writeDocString<Operation, Mode> (
    CodeVisitor::visitInterfaceDefStart # servant operation
    CodeVisitor::visitInterfaceDefStart # sync proxy operation
    CodeVisitor::visitInterfaceDefStart # async proxy operation
)
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2098,22 +2129,21 @@ Slice::Python::CodeVisitor::writeDocstring(const optional<DocComment>& comment,
         _out << nl << "Enumerators:";
         for (const auto& enumerator : enumerators)
         {
-            _out << nl << enumerator->mappedName() << " -- ";
+            _out << nl << nl << "- " << enumerator->mappedName();
```
externl, line 2132: I assume the extra nl is intentional?

externl (approved):

cpp/src/slice2py/PythonUtil.cpp
InsertCreativityHere: Yep, the first `nl` moves to the next line, and the second one adds a blank line between 2 enumerators.

bernardnormier (approved):

InsertCreativityHere:
I agree a state variable would be better here.
This is a problem in most of the language compilers actually, since the code was largely copy-pasted.
But I'm going to leave this for a future PR.

PR #4124, opened 2025-06-18
Add workflow for building docker images

pepone:
This PR introduces a new workflow for building the Docker container images used for RPM builds.

It builds the following images:

- rpm-repo-builder (A image used to sign RPMs and create RPM repositories)
- ice-rpm-builder-rhel9 (Ice 3.8 RPM builder for rhel9)
- ice-rpm-builder-rhel10 (Ice 3.8 RPM builder for rhel10)
- packaging/rpm/docker/amzn2023 (Ice 3.8 RPM builder for amzn2023)

There is a few minor fixes to ice.spec for compatibility with Red Hat 10, but more cleanup to remove i686 is left for a follow up PR.

Added a new rpm-sign.sh script that is used by the rpm-repo-builder container for signing RPMs. The idea is we would upload the signed RPMs to the job run, or tag. Which allow us to regenerate the repositories from the GitHub artifacts.

### DIFF
--- a/.github/workflows/build-container-images.yml
+++ b/.github/workflows/build-container-images.yml
@@ -1,0 +1,110 @@
+name: "Build Docker CI Images"
+
+on:
+  workflow_dispatch
+
+jobs:
+  build-images:
+    name: "Build ${{ matrix.image }} (${{ matrix.platform }})"
+    runs-on: ${{ matrix.platform == 'arm64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
+    strategy:
+      matrix:
+        include:
+          - platform: amd64
+            image: rpm-repo-builder
+            context: packaging/rpm/docker/rpm-repo-builder
+          - platform: arm64
+            image: rpm-repo-builder
+            context: packaging/rpm/docker/rpm-repo-builder
+          - platform: amd64
+            image: ice-rpm-builder-rhel9
+            context: packaging/rpm/docker/rhel9
+          - platform: arm64
+            image: ice-rpm-builder-rhel10
+            context: packaging/rpm/docker/rhel10
+          - platform: amd64
+            image: ice-rpm-builder-rhel10
+            context: packaging/rpm/docker/rhel10
+          - platform: arm64
+            image: ice-rpm-builder-rhel10
+            context: packaging/rpm/docker/rhel10
+          - platform: amd64
+            image: ice-rpm-builder-amzn2023
+            context: packaging/rpm/docker/amzn2023
+          - platform: arm64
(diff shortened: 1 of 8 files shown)


.github/workflows/build-container-images.yml
```diff
@@ -0,0 +1,110 @@
...
+      - name: Log in to GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: zeroc-ice
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Create and push manifest
+        run: |
+          docker manifest create ghcr.io/zeroc-ice/${{ matrix.image }}:latest \
```
pepone, line 110: The manifest allow us to use `ice-rpm-builder-amzn2023` and it gets redirected to the corresponding arch image

packaging/rpm/docker/amzn2023/Dockerfile
```diff
@@ -13,16 +17,5 @@ RUN dnf install --setopt=multilib_policy=all -y \
     redhat-rpm-config \
     glibc-static \
     libstdc++-static && \
-    dnf clean all
-
-# Copy Ice spec file into the container
-COPY packaging/rpm/ice.spec /usr/src/ice.spec
-
-# Install build dependencies. Do not use dnf builddep as it doesn't obey multilib_policy=all.
```
pepone, line 21: Simplified this because we are not building i686 anymore, don't need for multilib_policy and we can use builddep

packaging/rpm/docker/rpm-repo-builder/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM rockylinux:9
```
pepone, line 1: We can use any RPM based distribution for this, using Rocky Linux is simpler as we avoid the registration steps.

packaging/rpm/ice.spec
```diff
@@ -7,6 +7,16 @@
...
+%define javaversion 17-amazon-corretto
+%endif
+
+%if "%{dist}" == ".el10"
+%define javaversion 21-openjdk
+%endif
+
 %define shadow shadow-utils
 %define javapackagestools javapackages-tools
 # Unfortunately bzip2-devel does not provide pkgconfig(bzip2) as of EL7
```
bernardnormier, line 22: Is this fixed? If not, let's update this comment.

bernardnormier (approved):

packaging/rpm/ice.spec
pepone: I will rework in a follow up PR when I get rid of i686 in ice.spec.

PR #4125, opened 2025-06-19
Use Properties and Communicator constructors (MATLAB)

bernardnormier:
This PR provides public & preferred constructors for Properties and Communicator in MATLAB.

createProperties and initialize are still provided for backwards compatiblity.


matlab/lib/+Ice/Communicator.m
```diff
@@ -30,18 +30,43 @@
...
+            %   options - Name=Value options provided by the Ice.InitializationData class.
+            %
+            % Returns:
+            %   communicator (Ice.Communicator) - The new communicator.
+            %   remArgs (string array) - Contains the remaining command-line arguments that were not used to set
+            %     properties.
+            arguments
+                args (1, :) = {}
+                options.?Ice.InitializationData
+                options.Properties (1, 1) Ice.Properties = Ice.Properties()
```
bernardnormier, line 55: See https://www.mathworks.com/help/matlab/matlab_prog/validate-name-value-arguments.html#mw_f9a1e6a7-464c-46e8-85c8-0937093c595b

As far as I can tell, the default values from `Ice.InitializationData` are not transferred - we need to repeat them here.

matlab/lib/+Ice/createProperties.m
```diff
@@ -1,36 +1,9 @@
...
-    %     Settings in configuration files and the argument vector override these defaults.
-    %
-    % Returns:
-    %   properties (Ice.Properties) - A new property set initialized with the property settings that were removed from
-    %      the argument vector and the default property set.
-    %   remArgs (string array) - Contains the remaining command-line arguments that were not used to set properties.
-
-    % Copyright (c) ZeroC, Inc.
-
+    % createProperties - Deprecated function to create an Ice.Properties object. Use the Ice.Properties constructor
```
bernardnormier, line 2: We do the same in other languages, e.g.
https://github.com/zeroc-ice/ice/blob/43b41cebeae2c99588de2179d1f8be8dad17d58a/csharp/src/Ice/Util.cs#L99

matlab/test/lib/TestHelper.m
```diff
@@ -92,25 +92,20 @@
...
-                    else
-                        initData.Properties = obj.createTestProperties(varargin{1});
-                    end
-                end
+        function communicator = initialize(obj, args, options)
+            arguments
+                obj (1, 1) TestHelper
+                args (1, :) = {}
+                options.?Ice.InitializationData
+                options.Properties (1, 1) Ice.Properties = obj.createTestProperties(args)
```
bernardnormier, line 104: I verified this works.

matlab/lib/+Ice/Communicator.m
```diff
@@ -30,18 +30,43 @@
...
+            % Parameters:
+            %   args (cell array of char or string array) - An optional argument vector. Any Ice-related options in this
+            %     vector are used to initialize the communicator properties.
+            %   options - Name=Value options provided by the Ice.InitializationData class.
+            %
+            % Returns:
+            %   communicator (Ice.Communicator) - The new communicator.
+            %   remArgs (string array) - Contains the remaining command-line arguments that were not used to set
+            %     properties.
+            arguments
```
bernardnormier, line 52: Note that no argument is an InitializationData. InitializationData is only used to validate the Name=Value args.

pepone (approved):

externl (approved):

PR #4126, opened 2025-06-19
Add test to test/Ice/properties (C++)

bernardnormier:
This PR adds a small test to the Properties test. It shows how Ice.ProgramName gets set.

### DIFF
--- a/cpp/test/Ice/properties/Client.cpp
+++ b/cpp/test/Ice/properties/Client.cpp
@@ -228,6 +228,17 @@ Client::run(int, char**)
         }
         cout << "ok" << endl;
     }
+
+    {
+        cout << "testing createProperties with args... " << flush;
+        Ice::StringSeq args{"--Foo=Bar", "--Ice.Trace.Network=3"};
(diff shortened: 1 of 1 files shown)


cpp/test/Ice/properties/Client.cpp
```diff
@@ -228,6 +228,17 @@ Client::run(int, char**)
...
         cout << "ok" << endl;
     }
+
+    {
+        cout << "testing createProperties with args... " << flush;
+        Ice::StringSeq args{"--Foo=Bar", "--Ice.Trace.Network=3"};
+        Ice::PropertiesPtr properties = Ice::createProperties(args);
+        test(args.size() == 1);
+        test(args[0] == "--Foo=Bar");
+        test(properties->getIceProperty("Ice.ProgramName") == "--Foo=Bar");
```
bernardnormier, line 238: This behavior is problematic when `args[0]` is not the program name, like in this test and in MATLAB.

externl (approved):

externl (approved):

pepone (approved):

PR #4128, opened 2025-06-19
Fix NoEndpointException constructor calls

externl:
In C++, Java, and JavaScript, many calls to the NoEndpointException
constructor were mistakenly calling the wrong constructor.


.github/workflows/cpp.yml
externl, line 1: A little update to have clang format print a diff of the format changes.

PR #4130, opened 2025-06-19
Add workflow for build ice-repo RPM packages for 3.8 and nightly

pepone:
This PR adds a GitHub Actions workflow to build the ice-repo RPM packages.

We now build a separate repository package for each channel, currently 3.8 and nightly, which installs the corresponding `.repo` file.

The following packages are generated:

- ice-repo-3.8
- ice-repo-nightly


.github/workflows/build-rpm-ice-repo-packages.yml
```diff
@@ -0,0 +1,106 @@
...
+            -e CHANNEL="${{ matrix.channel }}" \
+            ghcr.io/zeroc-ice/ice-rpm-builder-${{ matrix.distribution }} \
+            /workspace/ice/packaging/rpm/build-repo-package.sh
+
+      - name: Sign RPM packages and create repository
+        run: |
+          docker_args=(
+            --rm
+            -v "$GITHUB_WORKSPACE:/workspace"
+            -e GPG_KEY="${{ secrets.ICE_3_8_CI_SIGNER_KEY }}"
```
externl, line 61: Is this not the same signing key we use for our usual release? I'd prefer to limit the number of secrets we need to manage. We should only need one signing key I think.

.github/workflows/build-rpm-ice-repo-packages.yml
```diff
@@ -0,0 +1,106 @@
...
+        with:
+          registry: ghcr.io
+          username: zeroc-ice
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build RPM packages
+        run: |
+          docker run --rm \
+            -v "$GITHUB_WORKSPACE:/workspace" \
+            -e PLATFORM="${{ matrix.distribution }}" \
```
externl, line 51: We seem to have a few terms for the same thing: distribution/platform. I think distribution is the correct term to use.

.github/workflows/build-rpm-ice-repo-packages.yml
```diff
@@ -0,0 +1,106 @@
...
+            --rm
+            -v "$GITHUB_WORKSPACE:/workspace"
+            -e GPG_KEY="${{ secrets.ICE_3_8_CI_SIGNER_KEY }}"
+            -e GPG_KEY_ID="${{ secrets.ICE_3_8_CI_SIGNER_KEY_ID }}"
+            ghcr.io/zeroc-ice/rpm-repo-builder
+          )
+
+          docker run "${docker_args[@]}" /usr/local/bin/rpm-sign /workspace/build/
+
+      - name: Sync RPM packages to S3
```
externl, line 68: If I read this correctly this will only work temporarily. Once the distribution already has some packages and we release a new version this would destroy those old packages. Presumably we should sync locally first.

.github/workflows/build-rpm-ice-repo-packages.yml
```diff
@@ -0,0 +1,106 @@
...
+on:
+  workflow_dispatch
+
+jobs:
+  build:
+    name: "Build Ice Repo RPM packages for ${{ matrix.distribution }}"
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        include:
```
externl, line 12: Why build 3.8? In general we'll only want to "build" 3.8 packages when we tag?

pepone: This workflow would run manually when we want to create the repo package, we only need to do so once per channel. 3.8 3.9 nightly

.github/workflows/build-rpm-ice-repo-packages.yml
pepone: Yes we need to unify terms in a follow up PR, as there many workflows to update.

.github/workflows/build-rpm-ice-repo-packages.yml
pepone: We already do, and only copy new RPMs, I could add some comments.

```
# Download the Channel directory from s3
aws s3 sync s3://zeroc-downloads/ice/$CHANNEL/$DISTRIBUTION $DISTRIBUTION

# Use cp -n to avoid override existing files
rpm="ice-repo-${CHANNEL}-1.0.0-1.${DISTRIBUTION}.noarch.rpm"
cp -n "build/RPMS/noarch/$rpm" "${DISTRIBUTION}/$rpm"

# Upload to S3
aws s3 sync "$DISTRIBUTION" s3://zeroc-downloads/ice/$CHANNEL/$DISTRIBUTION --delete
```

.github/workflows/build-rpm-ice-repo-packages.yml
pepone: I didn't want to upload the old release key, but we can of course do that. This relies on a password-less key.

externl (approved):

pepone:
Merging as is, I'm preparing a new PR with further updates to the release workflow.

PR #4131, opened 2025-06-20
Improve doc-comments generated by slice2matlab

bernardnormier:
This PR reworks the doc-comment generation by slice2matlab, in particular:

 - switch to the correct terminology (input arguments, output arguments)
 - always generate doc-comment, even when there is no doc-comment in the Slice files

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -3441,41 +3441,48 @@ Slice::Operation::outParameters() const
     return result;
 }
 
-ParameterList
-Slice::Operation::returnAndOutParameters(const string& returnsName)
+ParameterPtr
+Slice::Operation::returnParameter()
 {
-    ParameterList outParams = outParameters();
-
-    // Check if this operation has a non-void return type.
     if (_returnType)
     {
-        // Then check if any of the out parameter's mapped names collide with the return name.
-        bool shouldEscapeReturnsName = false;
-        for (const auto& param : outParams)
+        // MATLAB allows input and output arguments to share the same name. Likewise, the .slice syntax allows in and
+        // return parameters to share the same name. So we only check out-parameters here.
+        string returnName = "returnValue";
+        for (const auto& param : outParameters())
         {
-            if (param->mappedName() == returnsName)
+            if (param->mappedName() == returnName)
             {
-                shouldEscapeReturnsName = true;
+                // If the return parameter name collides with an existing out parameter name, we escape it.
+                // The returnName is used as both Slice identifier and mapped identifier.
+                returnName += "_";
+                break;
             }
         }
 
-        // Create the dummy return-value parameter.
-        const string fixedName = returnsName + (shouldEscapeReturnsName ? "_" : "");
-        ParameterPtr returnParam =
-            make_shared<Parameter>(shared_from_this(), fixedName, _returnType, _returnIsOptional, _returnTag);
+        ParameterPtr returnParam = make_shared<Parameter>(
+            shared_from_this(), returnName, _returnType, _returnIsOptional, _returnTag);
+
+        // TODO: presumably, this yields a bad mapped name when the operation name is remapped with xxx:identifier.
         returnParam->setMetadata(getMetadata());
-
-        outParams.push_back(returnParam);
+        return returnParam;
     }
-
-    return outParams;
+    else
+    {
+        return nullptr;
+    }
 }
 
 ParameterList
-Slice::Operation::sortedReturnAndOutParameters(const string& returnsName)
+Slice::Operation::sortedReturnAndOutParameters()
 {
     // We get the return and out parameters, and filter out any optional parameters into a separate 'optional' list.
-    ParameterList required = returnAndOutParameters(returnsName);
+    ParameterList required = outParameters();
+    if (_returnType)
+    {
+        required.push_back(returnParameter());
+    }
+
     ParameterList optional;
 
     // First sort each parameter into either 'required' or 'optional'.
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -3460,8 +3460,8 @@ Slice::Operation::returnParameter()
             }
         }
 
-        ParameterPtr returnParam = make_shared<Parameter>(
-            shared_from_this(), returnName, _returnType, _returnIsOptional, _returnTag);
+        ParameterPtr returnParam =
+            make_shared<Parameter>(shared_from_this(), returnName, _returnType, _returnIsOptional, _returnTag);
 
         // TODO: presumably, this yields a bad mapped name when the operation name is remapped with xxx:identifier.
         returnParam->setMetadata(getMetadata());
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -3442,29 +3442,25 @@ Slice::Operation::outParameters() const
 }
 
 ParameterPtr
-Slice::Operation::returnParameter()
+Slice::Operation::returnParameter(const string& name)
 {
     if (_returnType)
     {
-        // MATLAB allows input and output arguments to share the same name. Likewise, the .slice syntax allows in and
-        // return parameters to share the same name. So we only check out-parameters here.
-        string returnName = "returnValue";
-        for (const auto& param : outParameters())
-        {
-            if (param->mappedName() == returnName)
-            {
-                // If the return parameter name collides with an existing out parameter name, we escape it.
-                // The returnName is used as both Slice identifier and mapped identifier.
-                returnName += "_";
-                break;
-            }
-        }
-
         ParameterPtr returnParam =
-            make_shared<Parameter>(shared_from_this(), returnName, _returnType, _returnIsOptional, _returnTag);
+            make_shared<Parameter>(shared_from_this(), name, _returnType, _returnIsOptional, _returnTag);
 
-        // TODO: presumably, this yields a bad mapped name when the operation name is remapped with xxx:identifier.
-        returnParam->setMetadata(getMetadata());
+        // Remove xxx:identifier metadata: we don't want to rename the return parameter to be the remapped operation
+        // name.
+        MetadataList metadata = getMetadata();
+        const string xxxIdentifier = unit()->languageName() + ":identifier";
+        metadata.erase(
+            remove_if(
+                metadata.begin(),
+                metadata.end(),
+                [&xxxIdentifier](const MetadataPtr& m) { return m->directive() == xxxIdentifier; }),
+            metadata.end());
+
+        returnParam->setMetadata(metadata);
         return returnParam;
     }
     else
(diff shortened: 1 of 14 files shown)


cpp/src/Slice/Parser.cpp
```diff
@@ -3441,41 +3441,48 @@ Slice::Operation::outParameters() const
...
-    ParameterList outParams = outParameters();
-
-    // Check if this operation has a non-void return type.
     if (_returnType)
     {
-        // Then check if any of the out parameter's mapped names collide with the return name.
-        bool shouldEscapeReturnsName = false;
-        for (const auto& param : outParams)
+        // MATLAB allows input and output arguments to share the same name. Likewise, the .slice syntax allows in and
+        // return parameters to share the same name. So we only check out-parameters here.
```
InsertCreativityHere: I'm absolutely fine with having a comment here, but I'd rather not have it specific to MATLAB or what decisions we choose for IceRPC. `libSlice` should be conceptually decoupled from any specific language.

cpp/src/Slice/Parser.cpp
```diff
@@ -3441,41 +3441,48 @@ Slice::Operation::outParameters() const
...
 
-        // Create the dummy return-value parameter.
-        const string fixedName = returnsName + (shouldEscapeReturnsName ? "_" : "");
         ParameterPtr returnParam =
-            make_shared<Parameter>(shared_from_this(), fixedName, _returnType, _returnIsOptional, _returnTag);
-        returnParam->setMetadata(getMetadata());
+            make_shared<Parameter>(shared_from_this(), returnName, _returnType, _returnIsOptional, _returnTag);
 
-        outParams.push_back(returnParam);
+        // TODO: presumably, this yields a bad mapped name when the operation name is remapped with xxx:identifier.
```
InsertCreativityHere: Yes, this is correct. It's part of a larger issue caused by metadata being ambiguous between operations and return types. We have all the knowledge necessary to fix it in our declarative metadata validator, but it's not exposed in a way that's useful here.

cpp/src/Slice/Parser.h
```diff
@@ -698,31 +698,32 @@ namespace Slice
...
         /// Returns a list of all this operation's out-parameters (all parameters marked with 'out').
         [[nodiscard]] ParameterList outParameters() const;
-        /// Returns this operation's out parameters and return type (if the operation is non-void), in that order.
-        /// @param returnsName The name that should be returned by `returnValueParam->name()`.
-        //
-        // Creating this temporary Parameter doesn't introduce cycles, since nothing from the AST points to it,
-        // even if it points back into the AST. So it will be destroyed when the returned list goes out of scope.
-        [[nodiscard]] ParameterList returnAndOutParameters(const std::string& returnsName);
+
+        /// Creates a temporary Parameter that represents this operation's return value.
```
InsertCreativityHere, line 713: I'd mention that this dummy parameter's name is `returnValue`, instead of the empty string or something.

cpp/src/Slice/Parser.h
```diff
@@ -698,31 +698,32 @@ namespace Slice
...
+        /// void.
+        [[nodiscard]] ParameterPtr returnParameter();
+
         /// Returns this operation's out parameters and return type sorted in this order: '(required..., optional...)'.
         /// If the this operation's return type is non-void and non-optional, it is at the end of the 'required' list.
         /// Otherwise, required parameters are kept in definition order and optional parameters are sorted by tag.
         ///
         /// For convenience, non-void return types are represented by a dummy `Parameter` in this list.
         /// However it's important to note that it is not _actually_ a parameter.
         ///
```
InsertCreativityHere, line 729:
```suggestion
        //
```
Or better yet, now that better understand what tags doxygen supports, maybe this non-doc-comment under here should become an `@remarks`.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -46,15 +50,14 @@ namespace
...
-    void writeCopyright(IceInternal::Output& out, string_view file)
+    void writeGeneratedFrom(IceInternal::Output& out, string_view file)
     {
         string::size_type pos = file.find_last_of('/');
         if (pos != string::npos)
         {
             file = file.substr(pos + 1);
         }
-
-        out << nl << "% Copyright (c) ZeroC, Inc.";
```
InsertCreativityHere, line 57: We emit a copyright line in every other language's compiler.
Should we not be doing this?

InsertCreativityHere (commented):
Looks good so far, haven't finished reviewing the MATLAB logic so far though.

cpp/src/Slice/Parser.cpp
bernardnormier: So what do you suggest?

Add a `bool checkInParamsToo` that would always be false because we only call this function from slice2matlab and ice2slice where a new "return name" can match an existing in-param name?

cpp/src/Slice/Parser.cpp
bernardnormier: I think the cleanest would be to filter out xxx:identifier from the setMetadata in this function.

cpp/src/slice2matlab/Main.cpp
bernardnormier: It doesn't work well in MATLAB because it's included in the help. And we should consider removing this copyright generation from other Slice compilers.

cpp/src/Slice/Parser.cpp
```diff
@@ -3441,41 +3441,48 @@ Slice::Operation::outParameters() const
...
+        for (const auto& param : outParameters())
         {
-            if (param->mappedName() == returnsName)
+            if (param->mappedName() == returnName)
             {
-                shouldEscapeReturnsName = true;
+                // If the return parameter name collides with an existing out parameter name, we escape it.
+                // The returnName is used as both Slice identifier and mapped identifier.
+                returnName += "_";
+                break;
```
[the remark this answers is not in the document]
bernardnormier: If a user provides parameters named `returnValue` and `returnValue_`, he's definitely trying to break the code generation. No need to worry about this artificial case.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -1978,23 +2028,29 @@ CodeVisitor::visitStructStart(const StructPtr& p)
     out.inc();
     vector<string> memberNames;
     DataMemberList convertMembers;
-    for (const auto& member : members)
+    for (const auto& q : members)
```
externl, line 2061: I think member was a better name.

externl (approved):

cpp/src/slice2matlab/Main.cpp
bernardnormier: I changed this name for consistency with other duplicates of the "same" loop.

cpp/src/Slice/Parser.cpp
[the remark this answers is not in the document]
bernardnormier: Resolved TODO

cpp/src/Slice/Parser.cpp
bernardnormier: Reworked this code. The only reasonable solution is for the caller to provide the name, and to _not_ perform any escaping here. Any escaping performed in this Parser function may not match the escaping performed by the caller.

cpp/src/Slice/Parser.cpp
bernardnormier: Filtered-out identifier as proposed.

cpp/src/Slice/Parser.h
bernardnormier: See updated code.

cpp/src/Slice/Util.h
```diff
@@ -70,5 +70,12 @@ namespace Slice
 
     /// Returns the first sentence in the provided list of lines.
     std::string getFirstSentence(const StringList& lines);
+
+    /// Checks if @p param collides with any mapped name in @p params. If yes, return param followed by an underscore;
+    /// otherwise, return param as is.
+    /// @param params The parameter list to check against.
+    /// @param param The parameter name to check.
+    /// @return The parameter name, possibly modified to avoid collisions.
+    std::string getEscapedParamName(const ParameterList& params, std::string_view param);
```
bernardnormier, line 79: We are doing this escaping in pretty much all compilers, so may as well have a common Util function.

cpp/src/ice2slice/Gen.cpp
```diff
@@ -509,7 +509,24 @@ Gen::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
         out << epar;
 
         // Write the operation's return type.
-        const ParameterList returnAndOutParams = op->returnAndOutParameters("returnValue");
+        string returnValueName = "returnValue";
+        // Make sure returnValueName does not collide with any out parameter name:
+        for (const auto& param : op->outParameters())
+        {
+            if (param->name() == returnValueName)
```
bernardnormier, line 516: We can't use getEscapedParamName because we compare with _name_ here. getEscapedParamName compares mapped name.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -679,7 +667,7 @@ Slice::CsVisitor::getDispatchParams(
         retS = typeToString(op->returnType(), ns, op->returnIsOptional());
     }
 
-    string currentParamName = getEscapedParamName(op, "current");
+    string currentParamName = getEscapedParamName(op->parameters(), "current");
```
bernardnormier, line 670: `->parameters()` means all parameters. I kept the old logic; it's likely we could do better.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -1318,11 +1310,15 @@ SwiftGenerator::writeMarshalAsyncOutParams(::IceInternal::Output& out, const Ope
     // Marshal parameters in this order '(required..., optional...)'.
 
     out << nl << "let " << operationReturnDeclaration(op) << " = value";
-    for (const auto& param : op->sortedReturnAndOutParameters("returnValue"))
+    for (const auto& param : op->sortedReturnAndOutParameters(getEscapedParamName(op->outParameters(), "returnValue")))
     {
-        // 'isOutParam' fails for return types, and for return types, we don't want the 'mappedName'.
```
bernardnormier, line 1323: I guess this was about the xxx:identifier metadata on the operation?

This is now fixed, so we can use mappedName all we want.

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -43,7 +43,7 @@ namespace
         Slice::validateMetadata(unit, "ruby", std::move(knownMetadata));
     }
 
-    string getEscapedParamName(const OperationPtr& p, const string& name)
+    string getEscapedRubyParamName(const OperationPtr& p, const string& name)
```
bernardnormier, line 46: Can't use the Util function because of the special Slice::Ruby::getMappedName logic.

cpp/src/Slice/Parser.cpp
```diff
@@ -3441,41 +3441,44 @@ Slice::Operation::outParameters() const
...
-        }
-
-        // Create the dummy return-value parameter.
-        const string fixedName = returnsName + (shouldEscapeReturnsName ? "_" : "");
         ParameterPtr returnParam =
-            make_shared<Parameter>(shared_from_this(), fixedName, _returnType, _returnIsOptional, _returnTag);
-        returnParam->setMetadata(getMetadata());
-
-        outParams.push_back(returnParam);
+            make_shared<Parameter>(shared_from_this(), name, _returnType, _returnIsOptional, _returnTag);
```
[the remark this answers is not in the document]
bernardnormier: It's assigned to returnParam, and returnParam is subsequently returned.

pepone (approved):

PR #4133, opened 2025-06-22
Add support for MATLAB 2025a

bernardnormier:
This PR updates the source code to support MATLAB 2025a.

Previously, we were returning a 0/nullptr `mxArray*` to mean "success", but this no longer works. So I replaced all these `return 0` by `return createEmptyArray()`.

This PR also removes a few function / constants that should be never used.

### DIFF
--- a/matlab/lib/+Ice/Communicator.m
+++ b/matlab/lib/+Ice/Communicator.m
@@ -84,7 +84,7 @@ classdef Communicator < IceInternal.WrapperObject
 
             enc = obj.getProperties().getProperty('Ice.Default.EncodingVersion');
             if isempty(enc)
-                obj.encoding = Ice.currentEncoding();
+                obj.encoding = IceInternal.Protocol.CurrentEncoding;
             else
                 arr = sscanf(enc, '%u.%u');
                 if length(arr) ~= 2
--- a/matlab/lib/+Ice/identityToString.m
+++ b/matlab/lib/+Ice/identityToString.m
@@ -14,5 +14,5 @@ function r = identityToString(id, mode)
         id (1, 1) Ice.Identity
         mode (1, 1) Ice.ToStringMode = Ice.ToStringMode.Unicode
     end
-    r = IceInternal.Util.callWithResult('Ice_identityToString', id, int32(mode));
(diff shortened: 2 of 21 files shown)


matlab/lib/+Ice/Communicator.m
```diff
@@ -84,7 +84,7 @@
 
             enc = obj.getProperties().getProperty('Ice.Default.EncodingVersion');
             if isempty(enc)
-                obj.encoding = Ice.currentEncoding();
+                obj.encoding = IceInternal.Protocol.CurrentEncoding;
```
bernardnormier, line 87: It's now a constant.

matlab/lib/+Ice/identityToString.m
```diff
@@ -14,5 +14,5 @@
         id (1, 1) Ice.Identity
         mode (1, 1) Ice.ToStringMode = Ice.ToStringMode.Unicode
     end
-    r = IceInternal.Util.callWithResult('Ice_identityToString', id, int32(mode));
+    r = IceInternal.Util.callWithResult('Ice_identityToString', id, mode);
```
bernardnormier, line 17: The cast was bogus. We expect an enumerator.

matlab/test/Ice/objects/Client.m
```diff
@@ -10,7 +10,7 @@ function client(args)
 
     % We need to use the ClassSliceLoader for the classes with compact IDs. Naturally, it also works for classes
     % without a compact ID.
-    properties = Ice.Properties(args)
+    properties = Ice.Properties(args);
```
bernardnormier, line 13: added semicolon

matlab/test/Ice/proxy/AllTests.m
```diff
@@ -367,33 +367,31 @@
 
             assert(strcmp(communicator.identityToString(base.ice_identity(Ice.stringToIdentity('other')).ice_getIdentity()), 'other'));
 
-            %{
```
bernardnormier, line 370: Enabled / fix this test.

pepone (approved):

externl (approved):

PR #4136, opened 2025-06-23
Rework release workflow to support nightly and 3.8 channels

pepone:
This PR reworks the Nightly release workflow to support 3.8 and Nightly channels.


.github/workflows/release/publish-release.yml
```diff
@@ -0,0 +1,97 @@
...
+        required: true
+        type: string
+      run_id:
+        required: true
+        type: string
+jobs:
+  publish-release:
+    runs-on: 'ubuntu-24.04'
+
+    steps:
```
externl: I think these should be separate jobs. In case one fails, they would all run independently.

externl (approved):
Looks good!

PR #4138, opened 2025-06-23
Fix Over-Tolerant Parsing of Doc-Comment Tags

### LINKED ISSUE #4129: @returns doc-comment tag in Slice
It appears to be treated like `@return` followed by a `s` character.

InsertCreativityHere:
This PR fixes #4129, which was caused by the parser not checking the _end_ of a doc-comment tag.
All we checked was basically `text.find("@param") == 0`, which will match `@param`, but also `@paramdghkjdgf`.
Not great.

This PR updates the parser to enforce that the character after a tag must be whitespace or a line-break.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -345,46 +345,62 @@ namespace
 
     bool parseNamedCommentLine(string_view l, string_view tag, string& name, string& doc)
     {
+        const string ws = " \t";
+        const auto tagLength = tag.size();
+
+        // Check if the line starts with the provided tag (leading whitespace has already been stripped off).
         if (l.find(tag) == 0)
         {
-            const string ws = " \t";
-
-            auto nameStart = l.find_first_not_of(ws, tag.size());
-            if (nameStart == string::npos)
+            // The tag must be immediately followed by whitespace, or be the entire line (for multiline tags).
+            const auto tagEnd = l.find_first_of(ws, tagLength);
+            if (tagEnd == tagLength || l.length() == tagLength)
             {
-                return false; // Malformed line, ignore it.
-            }
+                auto nameStart = l.find_first_not_of(ws, tagLength);
+                if (nameStart == string::npos)
+                {
+                    return false; // Malformed line, ignore it.
+                }
 
-            auto nameEnd = l.find_first_of(ws, nameStart);
-            if (nameEnd == string::npos)
-            {
-                return false; // Malformed line, ignore it.
-            }
-            name = l.substr(nameStart, nameEnd - nameStart);
+                auto nameEnd = l.find_first_of(ws, nameStart);
+                if (nameEnd == string::npos)
+                {
+                    return false; // Malformed line, ignore it.
+                }
+                name = l.substr(nameStart, nameEnd - nameStart);
 
-            // Store whatever remains of the doc comment in the `doc` string.
-            auto docSplitPos = l.find_first_not_of(ws, nameEnd);
-            if (docSplitPos != string::npos)
-            {
-                doc = l.substr(docSplitPos);
-            }
+                // Store whatever remains of the doc comment in the `doc` string.
+                auto docSplitPos = l.find_first_not_of(ws, nameEnd);
+                if (docSplitPos != string::npos)
+                {
+                    doc = l.substr(docSplitPos);
+                }
 
-            return true;
+                return true;
+            }
         }
         return false;
     }
 
     bool parseCommentLine(string_view l, string_view tag, string& doc)
     {
+        const string ws = " \t";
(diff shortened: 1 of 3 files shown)


cpp/test/Slice/errorDetection/WarningInvalidComment.err
```diff
@@ -4,25 +4,28 @@ WarningInvalidComment.ice:11: warning: the '@exception' tag is only valid on ope
...
+WarningInvalidComment.ice:43: warning: '@throws CommentDummy': this exception is not listed in nor derived from any exception in this operation's specification
+WarningInvalidComment.ice:54: warning: ignoring duplicate doc-comment tag: '@param value'
+WarningInvalidComment.ice:54: warning: '@param namess' does not correspond to any parameter in operation 'stringOp'
+WarningInvalidComment.ice:54: warning: ignoring duplicate doc-comment tag: '@return'
+WarningInvalidComment.ice:54: warning: '@throws SomeOtherException': this exception is not listed in nor derived from any exception in this operation's specification
+WarningInvalidComment.ice:54: warning: ignoring duplicate doc-comment tag: '@throws CommentDummy'
+WarningInvalidComment.ice:29: warning: ignoring unknown doc tag '@something' in comment
+WarningInvalidComment.ice:29: warning: ignoring unknown doc tag '@params' in comment
+WarningInvalidComment.ice:29: warning: ignoring unknown doc tag '@throwss' in comment
+WarningInvalidComment.ice:29: warning: ignoring unknown doc tag '@deprecated:' in comment
```
InsertCreativityHere, line 19: GitHub is struggling to show the diff here.
Only the 3 lines above this comment are new.
But adding them shifted the line numbers around.
All that changed on the other lines are their line-numbers.

cpp/src/Slice/Parser.cpp
```diff
@@ -665,7 +681,7 @@ Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatt
                 // We've encountered an unknown doc tag.
                 if (line[0] == '@')
                 {
-                    auto unknownTag = line.substr(0, line.find_first_of(" \t:"));
+                    auto unknownTag = line.substr(0, line.find_first_of(" \t"));
```
InsertCreativityHere, line 683: Colons aren't supported after doc-comment tags.
Must of gotten some wires crossed with IceRPC.

cpp/src/Slice/Parser.cpp
```diff
@@ -345,46 +345,62 @@ namespace
 
     bool parseNamedCommentLine(string_view l, string_view tag, string& name, string& doc)
     {
+        const string ws = " \t";
+        const auto tagLength = tag.size();
```
externl, line 349: I think we should prefer to return early in the case of an error. 

if (l.find(tag) != 0)
{
    return false;
}

...

cpp/src/Slice/Parser.cpp
```diff
@@ -345,46 +345,62 @@ namespace
...
         return false;
     }
 
     bool parseCommentLine(string_view l, string_view tag, string& doc)
     {
+        const string ws = " \t";
+        const auto tagLength = tag.size();
+
+        // Check if the line starts with the provided tag (leading whitespace has already been stripped off).
         if (l.find(tag) == 0)
```
externl: Ditto.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: I refactored the code to remove some duplication, and to fail-fast in all of the `if`s except for one which used an out-parameter.

cpp/src/Slice/Parser.cpp
```diff
@@ -374,21 +404,6 @@ namespace
...
-            // Find the first whitespace that appears after the tag. Everything after it is part of the `doc` string.
-            auto docSplitPos = l.find_first_not_of(" \t", tag.size());
-            if (docSplitPos != string::npos)
-            {
-                doc = l.substr(docSplitPos);
-            }
-            return true;
-        }
-        return false;
-    }
```
InsertCreativityHere, line 390: `parseNamedComment` re-uses this function now, so I had to move it above `parseNamedComment`.

cpp/src/Slice/Parser.cpp
```diff
@@ -343,11 +343,41 @@ namespace
         return result;
     }
 
+    bool parseCommentLine(string_view l, string_view tag, string& doc)
```
bernardnormier: Please don't use `l` as a parameter name.

InsertCreativityHere: Renamed it to `line`.
`l` was just the original name.

PR #4140, opened 2025-06-23
Add Support for `@remarks` to `slice2matlab`

InsertCreativityHere:
This PR just does what it says.
MATLAB has no special documentation thing for remarks or notes, so we just generate a normal section named `Remarks`.
We always generate this at the end, after the summary, properties, see-also-s etc.

Here is the changes my PR causes to the generated code:

----

For the `EncodingVersion` struct:
```diff
@@ -12,6 +12,9 @@ classdef (Sealed) EncodingVersion
     %     major - The major version of the Ice encoding.
     %     minor - The minor version of the Ice encoding.
     %
+    %% Remarks
+    %     The Ice encoding is also known as the Slice encoding.
+    %
     %   Generated from Version.ice by slice2matlab version 3.8.0-alpha.0
```

and for the `LookupPrx` proxy:
```diff
@@ -21,6 +21,12 @@ classdef LookupPrx < Ice.ObjectPrx
     %
     %   See also IceLocatorDiscovery.LookupReply
     %
+    %% Remarks
+    %     This interface is implemented by Ice locator implementations and can be used by clients to find
+    %     available Ice locators on the network.
+    %     Ice locator implementations provide a well-known 'Ice/LocatorLookup' object accessible through UDP multicast.
+    %     Clients typically make a multicast findLocator request to find the locator proxy.
+    %
     %   Generated from Lookup.ice by slice2matlab version 3.8.0-alpha.0

     methods
```

----

And here is what the `help` output looks like within MATLAB (for an enumerator and an operation):

----

Screenshot 2025-06-23 154919

----

Screenshot 2025-06-23 154928

### DIFF
--- a/cpp/src/slice2matlab/Main.cpp
+++ b/cpp/src/slice2matlab/Main.cpp
@@ -388,13 +388,9 @@ namespace
         }
     }
 
-    void writeSeeAlso(IceInternal::Output& out, const optional<DocComment>& doc, const ContainerPtr& container)
+    void writeSeeAlso(IceInternal::Output& out, const DocComment& doc, const ContainerPtr& container)
     {
-        if (!doc)
-        {
-            return;
-        }
-        const StringList& seeAlso = doc->seeAlso();
+        const StringList& seeAlso = doc.seeAlso();
         if (seeAlso.empty())
         {
             return;
@@ -476,6 +472,19 @@ namespace
                 out << ": " << *deprecationReason;
             }
         }
+    }
+
+    void writeRemarks(IceInternal::Output& out, const DocComment& doc)
+    {
+        const StringList& remarks = doc.remarks();
+        if (remarks.empty())
+        {
+            return;
+        }
+
+        out << nl << "%";
+        out << nl << "%% Remarks";
+        writeDocLines(out, remarks, 4);
     }
 
     void writeConstructorDoc(IceInternal::Output& out, const string& name, const DataMemberList& fields)
@@ -568,8 +577,12 @@ namespace
             writePropertiesSummary(out, name, ex->dataMembers());
         }
 
-        writeSeeAlso(out, doc, p->container());
         writeDeprecated(out, doc, p);
+        if (doc)
+        {
+            writeSeeAlso(out, *doc, p->container());
+            writeRemarks(out, *doc);
+        }
     }
 
     void writeOpDocSummary(IceInternal::Output& out, const OperationPtr& p, bool async)
@@ -686,7 +699,12 @@ namespace
             }
         }
 
-        writeSeeAlso(out, doc, p->container());
+        if (doc)
+        {
+            writeSeeAlso(out, *doc, p->container());
+            writeRemarks(out, *doc);
+        }
+
         out << nl;
     }
 
@@ -760,8 +778,12 @@ namespace
             << p->scoped() << ".";
         out << nl << "%     uncheckedCast - Creates a " << name << " from another proxy without any validation.";
 
-        writeSeeAlso(out, doc, p->container());
         writeDeprecated(out, doc, p);
+        if (doc)
+        {
+            writeSeeAlso(out, *doc, p->container());
+            writeRemarks(out, *doc);
+        }
     }
 
     void declareProperty(IceInternal::Output& out, const DataMemberPtr& field)
@@ -1054,8 +1076,12 @@ namespace
             doc ? doc->overview() : StringList{},
             false);
 
-        writeSeeAlso(out, doc, field->container());
         writeDeprecated(out, doc, field);
+        if (doc)
+        {
+            writeSeeAlso(out, *doc, field->container());
+            writeRemarks(out, *doc);
+        }
     }
 
     void writeArguments(
--- a/cpp/src/slice2matlab/Main.cpp
+++ b/cpp/src/slice2matlab/Main.cpp
@@ -388,9 +388,13 @@ namespace
         }
     }
 
-    void writeSeeAlso(IceInternal::Output& out, const DocComment& doc, const ContainerPtr& container)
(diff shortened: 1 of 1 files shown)


cpp/src/slice2matlab/Main.cpp
```diff
@@ -568,8 +577,12 @@ namespace
             writePropertiesSummary(out, name, ex->dataMembers());
         }
 
-        writeSeeAlso(out, doc, p->container());
         writeDeprecated(out, doc, p);
```
InsertCreativityHere, line 588: Is there a standard ordering for deprecated vs see-also?

I swapped these purely so that I could move `seeAlso` into a block where we know `doc` is non-null.
If there's an ordering convention for these in MATLAB, I'll undo this though.

bernardnormier: There is no MATLAB convention, we choose what we want to do.

I prefer see also at the end, and to output deprecated fairly early. For example, in operations, I output deprecated before the arguments. If a method is deprecated, may as well learn about it early.

> where we know doc is non-null.

Don't do that. The writeSeeAlso helper I wrote checks if doc is null.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -388,13 +388,9 @@ namespace
         }
     }
 
-    void writeSeeAlso(IceInternal::Output& out, const optional<DocComment>& doc, const ContainerPtr& container)
+    void writeSeeAlso(IceInternal::Output& out, const DocComment& doc, const ContainerPtr& container)
```
bernardnormier: Please keep the old signature and behavior.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -478,6 +474,19 @@ namespace
         }
     }
 
+    void writeRemarks(IceInternal::Output& out, const DocComment& doc)
```
bernardnormier: Please use the same logic as the original writeSeeAlso.

cpp/src/slice2matlab/Main.cpp
```diff
@@ -478,6 +474,19 @@ namespace
...
+    {
+        const StringList& remarks = doc.remarks();
+        if (remarks.empty())
+        {
+            return;
+        }
+
+        out << nl << "%";
+        out << nl << "%% Remarks";
```
bernardnormier: The formatting should be consistent other sub-sections in the generated doc-comment, like Creation, the method summary, Input Argument etc.:
 - single `%`
 - aligned with the other sub-sections, i.e. `%` followed by 3 spaces, then ident with 2 additional spaces.

```matlab
%NAME Beginning of description
%   More description
%
%   Remarks
%     Something more to say.
%     On more than one line.
````

File: cpp/src/slice2matlab/Main.cpp [hunk 5]
```diff
@@ -568,8 +577,12 @@ namespace
             writePropertiesSummary(out, name, ex->dataMembers());
         }
 
-        writeSeeAlso(out, doc, p->container());
         writeDeprecated(out, doc, p);
+        if (doc)
```
bernardnormier: I changed the code in a previous PR to avoid this pattern and instead perform the check in writeXxx. I certainly prefer the pattern from my change.

cpp/src/slice2matlab/Main.cpp
InsertCreativityHere: Fixed as suggested!

cpp/src/slice2matlab/Main.cpp
InsertCreativityHere: > I prefer see also at the end, and to output deprecated fairly early.

I agree with your preference, which is what I modified the code to do.

cpp/src/slice2matlab/Main.cpp
InsertCreativityHere: I switched to using the optional-check logic.
Although I still think it's repetitive to check if the comment is non-null over and over.

InsertCreativityHere:
@bernardnormier here's what the generated code looks like now:

```diff
@@ -20,6 +20,10 @@ classdef OperationMode < uint8

         %IDEMPOTENT An idempotent operation. The Ice client runtime does not guarantee at-most-once semantics for such an
         %   operation.
+        %
+        %   Remarks
+        %     When an operation is idempotent, the Ice runtime will attempt to transparently recover from certain
+        %     runtime errors by re-issuing a failed request transparently.
         Idempotent (2)
     end
     methods(Static)
```

```diff
@@ -112,6 +115,9 @@ classdef RemoteLoggerPrx < Ice.ObjectPrx
             %       Ice.Future scalar
             %
             %   See also log, Ice.Future.
+            %
+            %   Remarks
+            %     log may be called by LoggerAdmin before init.

             arguments
                 obj (1, 1) Ice.RemoteLoggerPrx
```

```diff
@@ -23,6 +23,9 @@ classdef ServiceObserverPrx < Ice.ObjectPrx
     %
     %   See also IceBox.ServiceManager.addObserver
     %
+    %   Remarks
+    %     This interface is implemented by admin tools that monitor the IceBox server.
+    %
     %   Generated from ServiceManager.ice by slice2matlab version 3.8.0-alpha.0

     methods
```

```diff
@@ -21,6 +21,12 @@ classdef LookupPrx < Ice.ObjectPrx
     %
     %   See also IceLocatorDiscovery.LookupReply
     %
+    %   Remarks
+    %     This interface is implemented by Ice locator implementations and can be used by clients to find
+    %     available Ice locators on the network.
+    %     Ice locator implementations provide a well-known 'Ice/LocatorLookup' object accessible through UDP multicast.
+    %     Clients typically make a multicast findLocator request to find the locator proxy.
+    %
     %   Generated from Lookup.ice by slice2matlab version 3.8.0-alpha.0
```

bernardnormier (approved):
Looks good.

PR #4141, opened 2025-06-23
Update MATLAB doc-comments

bernardnormier:
This PR updates all the non-generated doc-comments in Ice for MATLAB.


matlab/lib/+Ice/PropertyException.m
```diff
@@ -1,9 +1,6 @@
-% PropertyException   Summary of PropertyException
-%
-% This exception is raised when there is an error while getting or setting a property. For example, when
-% trying to set an unknown Ice property.
-
-% Copyright (c) ZeroC, Inc.
-
 classdef (Sealed) PropertyException < Ice.LocalException
+    %PROPERTYEXCEPTION The exception that is thrown when a property cannot be set or retrieved. For example, this
+    %   exception is thrown when attempting to set an unknown Ice property.
```
externl, line 3: PropertyException. We can also throw this for when a property has an invalid value.

matlab/lib/+Ice/stringVersion.m
```diff
@@ -1,9 +1,10 @@
...
-    % stringVersion  Returns the Ice version in the form A.B.C, where A
-    %   indicates the major version, B indicates the minor version, and
-    %   C indicates the patch level.
+    %STRINGVERSION Returns the Ice version in the form A.B.C, where A indicates the major version, B indicates the
+    %   minor version, and C indicates the patch level.
     %
-    % Returns (char) - The Ice version.
+    %   Output Arguments
+    %     r - The Ice version.
+    %       character vector
```
externl, line 7: Should all of these `character vector` be `character vector | string scalar`?

externl (approved):

matlab/lib/+Ice/stringVersion.m
bernardnormier: No, an output argument has a single type. It's only an input argument that can have multiple types.

Further more, for "string" input arguments, we only accept character vector (never string). It's because a Slice string maps to a character vector. On the other hand, for `sequence<string>` input arguments, we accept both cell array of character [vector] (= the Ice 3.7 mapping) and string array (= the Ice 3.8 mapping). If you spot any hand-written API where we document only one or the other, it's most likely a bug.

matlab/lib/+Ice/ConnectionIdleException.m
```diff
@@ -1,8 +0,0 @@
-% ConnectionIdleException   Summary of ConnectionIdleException
```
InsertCreativityHere, line 1: Not for this PR, but we should also delete this exception from Swift.

matlab/lib/+Ice/Future.m
```diff
@@ -63,31 +60,31 @@
             nextId = nextId + 1;
         end
     end
-    methods
-        function delete(obj)
+    methods (Hidden)
+         function delete(obj)
```
InsertCreativityHere: Extra space added here?
```suggestion
    methods (Hidden)
        function delete(obj)
```

matlab/lib/+Ice/PropertyException.m
bernardnormier: This description was copied from the C++ description.

matlab/lib/+Ice/OptionalFormat.m
```diff
@@ -1,23 +1,44 @@
...
+        %SIZE Variable-length format with size.
         Size = uint8(4)
+
+        %VSIZE Variable-length format with variable size.
         VSize = uint8(5)
+
+        %FSIZE Fixed-length format with size.
         FSize = uint8(6)
+
+        %CLASS Class format.
```
InsertCreativityHere: We should really mention that this is no longer supported.

matlab/lib/+Ice/OptionalFormat.m
```diff
@@ -1,23 +1,44 @@
...
+        %F8 Fixed-length format (8 bytes).
         F8 = uint8(3)
+
+        %SIZE Variable-length format with size.
         Size = uint8(4)
+
+        %VSIZE Variable-length format with variable size.
         VSize = uint8(5)
+
+        %FSIZE Fixed-length format with size.
```
InsertCreativityHere, line 38: I think these explanations are lackluster compared to the C++ comments.
We should at least mention the range of bytes that these can take up, and be more descriptive than saying:
`Variable-length format with size.`

matlab/lib/+Ice/OptionalFormat.m
```diff
@@ -1,23 +1,44 @@
...
-    % unmarshaling code if the optional is not known to the receiver.
+    %   OptionalFormat Properties:
+    %     F1 - Fixed-length format (1 byte).
+    %     F2 - Fixed-length format (2 bytes).
+    %     F4 - Fixed-length format (4 bytes).
+    %     F8 - Fixed-length format (8 bytes).
+    %     Size - Variable-length format with size.
+    %     VSize - Variable-length format with variable size.
+    %     FSize - Fixed-length format with size.
+    %     Class - Class format.
```
InsertCreativityHere, line 14: Same here too I guess.

matlab/lib/+Ice/OptionalFormat.m
```diff
@@ -1,23 +1,44 @@
 classdef OptionalFormat < uint8
-    % OptionalFormat   Summary of OptionalFormat
+    %OPTIONALFORMAT The optional format type.
+    %   An optional value is encoded with a specific optional format. This optional format describes how the data is
+    %   encoded and how it can be skipped by the unmarshaling code if the optional is not known to the receiver.
```
InsertCreativityHere: We prefer 'marshaled' for Ice, and 'encoded' for IceRPC I believe.
```suggestion
    %   marshaled and how it can be skipped by the unmarshaling code if the optional is not known to the receiver.
```

Just to be sure, the C++ doc-comment for this uses the term 'marshaled'.
There are other small differences in this comment from the C++ version too, like this one doesn't mention the tag.

matlab/lib/+Ice/Value.m
```diff
@@ -30,31 +29,36 @@
             %assert(index > 0); % Check for rollover
             obj.iceInternal_ = index;
         end
+
         function ice_preMarshal(~)
-            % ice_preMarshal - The Ice run time invokes this method prior to
-            %   marshaling an object's data members. This allows a subclass
-            %   to override this method in order to validate its data members.
+            %ICE_PREMARSHAL - Ice invokes this method prior to marshaling the fields of this object. This allows a
```
InsertCreativityHere: I'm sure it's just keeping the previous text here, but I've seen a mix of `<FUNCTIONNAME> - ...` and `<FUNCTIONNAME> ...`, favoring the version without the dash.

We should pick one and use it everywhere IMO.

matlab/lib/+Ice/ToStringMode.m
```diff
@@ -1,29 +1,28 @@
...
-% instantiation is language-specific, and not specified in Slice code.
-%
-% ToStringMode Properties:
-%   Unicode - Characters with ordinal values greater than 127 are kept as-is in the resulting string.
-%   ASCII - Characters with ordinal values greater than 127 are encoded as universal character names in the resulting string: \\unnnn for BMP characters and \\Unnnnnnnn for non-BMP characters.
-%   Compat - Characters with ordinal values greater than 127 are encoded as a sequence of UTF-8 bytes using octal escapes.
-%
-% See also Ice.Logger, Ice.ObjectAdapter, Ice.Properties, The output mode for xxxToString method such as identityToString and proxyToString. The actual encoding format for, the string is the same for all modes: you don't need to specify an encoding format or mode when reading such a, string.
+classdef ToStringMode < uint8
+    %TOSTRINGMODE The output mode for xxxToString method such as identityToString and proxyToString.
```
InsertCreativityHere: Missing an `s` here. To be fair, it's also missing in the C++ comment. We should fix both.
```suggestion
    %TOSTRINGMODE The output mode for xxxToString methods such as identityToString and proxyToString.
```

matlab/lib/+Ice/PropertyException.m
externl: It should be fixed there as well!

matlab/lib/+Ice/EndpointInfo.m
```diff
@@ -1,26 +1,25 @@
 classdef EndpointInfo < handle
-    % EndpointInfo   Summary of EndpointInfo
+    %ENDPOINTINFO Base class for the endpoint info classes.
     %
-    % Base class providing access to the endpoint details.
+    %   EndpointInfo Properties:
+    %     underlying - The EndpointInfo of the underlying endpoint, if any.
+    %     timeout - The timeout for the endpoint in milliseconds. 0 means non-blocking, -1 means no timeout.
```
pepone, line 6: What is this `non-blocking` about?

matlab/lib/+Ice/Future.m
```diff
@@ -1,44 +1,41 @@
...
+        %   logical scalar
+        Read (1, 1) logical = false
 
-        % State - The current state of the future. Its initial value is
-        %   'running' and its final value is 'finished'. A remote invocation
-        %   transitions from 'running' to 'sent' to 'finished'.
-        State char = 'running'
+        %STATE The current state of the future. Its initial value is 'running' and its final value is 'finished'.
+        %   'running' | 'sent' | 'finished'
+        State (1, :) char = 'running'
```
pepone, line 38: Wonder why we choose a char array for State.

matlab/lib/+Ice/ImplicitContext.m
```diff
@@ -113,14 +100,17 @@ function setContext(obj, newContext)
             end
             r = obj.iceCallWithResult('put', key, value);
         end
+
         function r = remove(obj, key)
-            % remove - Remove the entry for the given key in the underlying
-            %   context.
+            %REMOVE Removes the entry for the given key in the underlying context.
```
pepone: Will be good to document what happens if you attempt to remove a non existing key.

matlab/lib/+Ice/InitializationData.m
```diff
@@ -1,9 +1,11 @@
...
+    %INITIALIZATIONDATA Represents a set of options that you can specify when initializing a communicator.
     %
-    % InitializationData Properties:
-    %   Properties - The properties of the communicator.
-    %   SliceLoader - The Slice loader, used to unmarshal Slice classes and exceptions.
+    %   InitializationData Properties:
+    %     Properties - The properties of the communicator.
+    %     SliceLoader - The Slice loader, used to unmarshal Slice classes and exceptions.
+    %
+    %   See also Ice.initialize, Ice.Communicator.
```
pepone, line 9: missing copy right?

matlab/lib/+Ice/Properties.m
```diff
@@ -72,56 +71,73 @@
...
                 obj (1, 1) Ice.Properties
                 key (1, :) char
             end
             r = obj.iceCallWithResult('getProperty', key);
         end
+
         function r = getIceProperty(obj, key)
-            % getIceProperty - Get an Ice property by key. If the property is not set,
-            %   its default value is returned.
+            %GETICEPROPERTY Gets an Ice property by key. If the property is not set, its default value is returned.
```
pepone, line 92: This method can throw PropertyException if the property is unknown

The C++ version bogusly document invalid_argument, but it actually throws PropertyException:

```
        /// Gets an Ice property by key.
        /// @param key The property key.
        /// @return The property value, or the default value for this property if the property is not set.
        /// @throws std::invalid_argument Thrown when the property is not a known Ice property.
        /// @see #setProperty
        std::string getIceProperty(std::string_view key);
```

pepone (approved):

matlab/lib/+Ice/SliceLoader.m
```diff
@@ -1,17 +1,20 @@
...
+        %   Input Arguments
+        %     typeId - The Slice type ID or compact ID.
+        %       character vector
         %
-        % Returns (handle or []) - The new class instance or exception instance, or an empty array if the implementation
-        %   cannot find a class or exception for typeId.
+        %    Output Arguments
+        %      r - The new class instance or exception instance, or an empty array if the implementation cannot find a
+        %        class or exception for typeId.
+        %        Ice.Value | Ice.UserException | empty array
```
InsertCreativityHere, line 17: As a general statement, I find our new style for documenting the type of an argument to be less readable than previously. I'm not saying that the previous comments were correct or great, just the style of having it as a separate line at the end instead of parenthesized:

For example here, this is the new doc vs the old doc:
Screenshot 2025-06-24 114649
Screenshot 2025-06-24 114555

InsertCreativityHere (approved):
Only reviewed a small handful of files, due to the size of this PR.

There's a bunch of minor changes between the C++ doc-comments and the new MATLAB comments. I imagine these aren't intentional, but most of the discrepancies I saw didn't meaningfully impact the clarity of the comment.

matlab/lib/+Ice/OptionalFormat.m
InsertCreativityHere: Also `class format` is actually a totally different thing.

matlab/lib/+Ice/SliceLoader.m
bernardnormier: The new style is the preferred MATLAB style. Try for example help dictionary/insert:
```
dictionary/insert - Add entries to a dictionary
    This MATLAB function assigns the value to key in dictionary, d.

    Syntax
      d2 = insert(d1,key,value)
      d2 = insert(d1,key,value,Overwrite=tf)

    Input Arguments
      d1 - Input dictionary
        dictionary object
      key - Key set
        scalar | array
      value - Value set
        scalar | array
      tf - Option to overwrite existing entries
        true or 1 (default) | false or 0
     ...
```

(Mathworks doesn't use  `-` after the function name consistently. I didn't include it in our style)

matlab/lib/+Ice/ToStringMode.m
bernardnormier: Fixed everywhere.

matlab/lib/+Ice/Value.m
bernardnormier: Yes, fixed.

matlab/lib/+Ice/OptionalFormat.m
bernardnormier: Fixed

matlab/lib/+Ice/OptionalFormat.m
bernardnormier: This is ultra minor since we don't document InputStream and OutputStream.

matlab/lib/+Ice/OptionalFormat.m
bernardnormier: Fixed

matlab/lib/+Ice/Future.m
bernardnormier: fixed

matlab/lib/+Ice/Properties.m
bernardnormier: We should fix this doc-comment in all languages in a separate PR.

matlab/lib/+Ice/InitializationData.m
bernardnormier: Fixed

matlab/lib/+Ice/EndpointInfo.m
bernardnormier: Not sure. We have the same comment in C++. We should fix them everywhere now that this timeout is no longer used for anything.

matlab/lib/+Ice/ImplicitContext.m
bernardnormier: documented.

PR #4146, opened 2025-06-24
Require `@throws` to Exactly Match Exception Specifications

### LINKED ISSUE #4132: @throws/@exception should be an exact match
When documenting a Slice exception with `@throws` or `@exception`, you can currently document either an exception in the operation's exception specification _or_ a derived exception. See:

https://github.com/zeroc-ice/ice/blob/f29ac59850f2a3c4ac9692b017a701e84df610a1/cpp/src/Slice/Parser.cpp#L567

We should simplify the logic and require an exact match.

An exact match is simpler for code generation, because we can iterate over the exception specification to document all exceptions - or possibly document only exceptions with `@throws`/`@exception`. I.e., use the same logic as for parameters.

This requirement has a very minor impact on users. In particular, adding derived exceptions in the exception specification doesn't affect the contract.

See also #4131, which relies on this simplification.

InsertCreativityHere:
Fixes #4132.

### DIFF
--- a/cpp/test/Slice/errorDetection/WarningInvalidComment.err
+++ b/cpp/test/Slice/errorDetection/WarningInvalidComment.err
@@ -7,11 +7,12 @@ WarningInvalidComment.ice:17: warning: ignoring trailing '.' character in '@see'
 WarningInvalidComment.ice:43: warning: '@param NoParam' does not correspond to any parameter in operation 'voidOp'
 WarningInvalidComment.ice:43: warning: '@return' is only valid on operations with non-void return types
 WarningInvalidComment.ice:43: warning: '@throws FakeException': no exception with this name could be found from the current scope
-WarningInvalidComment.ice:43: warning: '@throws CommentDummy': this exception is not listed in nor derived from any exception in this operation's specification
+WarningInvalidComment.ice:43: warning: '@throws CommentDummy': this exception is not listed in this operation's exception specification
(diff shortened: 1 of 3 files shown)


cpp/test/Slice/errorDetection/WarningInvalidComment.err
```diff
@@ -7,11 +7,12 @@ WarningInvalidComment.ice:17: warning: ignoring trailing '.' character in '@see'
 WarningInvalidComment.ice:43: warning: '@param NoParam' does not correspond to any parameter in operation 'voidOp'
 WarningInvalidComment.ice:43: warning: '@return' is only valid on operations with non-void return types
 WarningInvalidComment.ice:43: warning: '@throws FakeException': no exception with this name could be found from the current scope
-WarningInvalidComment.ice:43: warning: '@throws CommentDummy': this exception is not listed in nor derived from any exception in this operation's specification
+WarningInvalidComment.ice:43: warning: '@throws CommentDummy': this exception is not listed in this operation's exception specification
```
bernardnormier: I think there is too much "this" in the message. Could be:
```
this exception is not listed in the exception specification
exception not listed in the exception specification of <opname>
```

bernardnormier (approved):

cpp/test/Slice/errorDetection/WarningInvalidComment.err
InsertCreativityHere: After reading it to myself, you're completely right.
I went with your first (keeping the first `this`).

PR #4148, opened 2025-06-25
Add package to install DEB repository

pepone:
This PR adds a new workflow for creating `ice-repo-3.8` and `ice-repo-nightly` debian packages, this packages can be used to install the corresponding debian repository.

This would simplify the setup instructions, and is consistent with what we already do for RPM distributions.

### DIFF
--- a/.github/workflows/build-deb-ice-repo-packages.yml
+++ b/.github/workflows/build-deb-ice-repo-packages.yml
@@ -42,12 +42,12 @@ jobs:
         run: |
           docker run --rm \
             -v "$GITHUB_WORKSPACE:/workspace" \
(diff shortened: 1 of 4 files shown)


externl (approved):
Looks good. I know this type of package is common for yum repos. Is it common for apt repos, I can't recall ever using one?

pepone:
.NET has one

https://learn.microsoft.com/en-gb/dotnet/core/install/linux-debian?tabs=dotnet9

A similar approach is also mentioned in debian wiki

https://wiki.debian.org/DebianRepository/SetupWithReprepro#Help_people_add_your_repository

PR #4151, opened 2025-06-25
Properties doc comment updates

externl:


pepone (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
```diff
@@ -177,6 +177,7 @@ public synchronized String getProperty(String key) {
      *
      * @param key The property key.
      * @return The property value or the default value.
+     * @throws PropertyException Raised if the property is not a known Ice property.
```
bernardnormier: Do we prefer Raised over Thrown in Java?

bernardnormier (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
```diff
@@ -290,6 +292,7 @@ public String[] getPropertyAsList(String key) {
      *
      * @param key The property key.
      * @return The property value interpreted as list of strings, or the default value.
+     * @throws PropertyException Raised if the property is not a known Ice property.
```
InsertCreativityHere: Some Python leaked in here:
```suggestion
     * @throws PropertyException If the property is not a known Ice property.
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
InsertCreativityHere:
```suggestion
     * @throws PropertyException If the property is not a known Ice property.
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
InsertCreativityHere: Wording taken from the JavaDoc reference guide thing:
https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#@exception
If we prefer to use `Thrown if ...`, that's fine too.

externl: This was actually the terminology we're using in the rest of the file. I'll update them all to use your suggestion.

PR #4153, opened 2025-06-25
Fix doc-comment for Communicator destroy, shutdown and related methods

### LINKED ISSUE #2561: Improve Communicator destroy doc comments
(Not from this PR). I find the description misleading. We should never use "shut down the communicator's client functionality". Shut down is only about server-side functionality.

_Originally posted by @bernardnormier in https://github.com/zeroc-ice/ice/pull/2554#discussion_r1689812343_

bernardnormier:
Fixed #2561.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
@@ -5,7 +5,6 @@ package com.zeroc.Ice;
 import com.zeroc.Ice.Instrumentation.CommunicatorObserver;
 import com.zeroc.Ice.SSL.SSLEngineFactory;
 
-import java.security.Identity;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
@@ -5,6 +5,7 @@ package com.zeroc.Ice;
 import com.zeroc.Ice.Instrumentation.CommunicatorObserver;
 import com.zeroc.Ice.SSL.SSLEngineFactory;
 
+import java.security.Identity;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


matlab/lib/+Ice/Communicator.m
```diff
@@ -113,8 +113,8 @@ function destroy(obj)
         end
 
         function f = destroyAsync(obj)
-            %DESTROYASYNC Asynchronously destroys the communicator. Calling destroyAsync cleans up memory, and shuts
-            %   down this communicator's client functionality. Subsequent calls to destroyAsync are ignored.
+            %DESTROYASYNC Asynchronously destroys the communicator.
```
bernardnormier, line 116: Not clear why we provide this destroyAsync in MATLAB. It should always complete synchronously. Maybe we should remove it?

externl (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -5,6 +5,7 @@
 import com.zeroc.Ice.Instrumentation.CommunicatorObserver;
 import com.zeroc.Ice.SSL.SSLEngineFactory;
 
+import java.security.Identity;
```
pepone: This import looks bogus, probably added by Visual Studio Code.

js/src/Ice/Communicator.d.ts
```diff
@@ -11,8 +11,10 @@ declare module "@zeroc/ice" {
          */
         interface Communicator {
             /**
-             * Destroy the communicator. This operation calls {@link Communicator#shutdown} implicitly. Calling {@link Communicator#destroy} cleans up
-             * memory, and shuts down this communicator's client functionality and destroys all object adapters.
+             * Destroys this communicator. This method calls {@link Communicator#shutdown} implicitly. Calling
+             * {@link Communicator#destroy} destroys all object adapters, and closes all outgoing connections. `destroy`
+             * waits for all outstanding dispatches to complete before returning. This includes "bidirectional
```
pepone: In JavaScript the method is async, doesn't wait before returning.

pepone (approved):

js/src/Ice/Communicator.d.ts
bernardnormier: Fixed.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
bernardnormier: Fixed

PR #4154, opened 2025-06-25
Sync Java doc comments with C++

externl:
This PR syncs Java doc comments with their C++ counterparts. 

Namely:

- Communicator
- ObjectAdater
- Connection
- ObjectPrx
- *Exception
- Util/Initialize

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
@@ -17,13 +17,10 @@ public interface Connection {
     void close();
 
     /**
-     * Create a special proxy that always uses this connection. This can be used for callbacks from
-     * a server to a client if the server cannot directly establish a connection to the client, for
-     * example because of firewalls. In this case, the server would create a proxy using an already
-     * established connection from the client.
+     * Creates a special proxy (a "fixed proxy") that always uses this connection.
      *
-     * @param id The identity for which a proxy is to be created.
-     * @return A proxy that matches the given identity and uses this connection.
+     * @param id The identity of the target object.
+     * @return A fixed proxy with the provided identity.
      * @see #setAdapter
      */
     ObjectPrx createProxy(Identity id);
@@ -53,15 +50,15 @@ public interface Connection {
     ObjectAdapter getAdapter();
 
     /**
-     * Get the endpoint from which the connection was created.
+     * Gets the endpoint from which the connection was created.
      *
      * @return The endpoint from which the connection was created.
      */
     Endpoint getEndpoint();
 
     /**
-     * Flush any pending batch requests for this connection. This means all batch requests invoked
-     * on fixed proxies associated with the connection.
+     * Flushes any pending batch requests for this connection. This corresponds to all batch requests invoked on
+     * fixed proxies associated with the connection.
      *
      * @param compress Specifies whether or not the queued batch requests should be compressed
      *     before being sent over the wire.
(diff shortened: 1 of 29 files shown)


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectionRefusedException.java
```diff
@@ -3,8 +3,7 @@
 package com.zeroc.Ice;
 
 /**
- * This exception indicates a connection failure for which the server host actively refuses a
- * connection.
+ * The exception that is thrown when the server host actively refuses a connection.
  */
```
InsertCreativityHere: I suspect this is a one-liner now that it's shorter:
`/** The exception that is thrown when the server host actively refuses a connection. */`

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/UnknownUserException.java
```diff
@@ -3,10 +3,8 @@
 package com.zeroc.Ice;
 
 /**
- * The dispatch returned a {@link UserException} that was not declared in the operation's exception
- * specification (the <code>throws</code> clause). This is necessary in order to not violate the
- * contract established by an operation's signature: Only local exceptions and user exceptions
- * declared in the <code>throws</code> clause can be raised.
+ * The exception that is thrown when a client receives a UserException that was not declared in the operation's
+ * exception specification.
```
InsertCreativityHere: In C++ we can rely on auto-linking, here I think it'd be nice to keep the link:
```suggestion
 * The exception that is thrown when a client receives a {@link UserException} that was not declared in the 
 * operation's exception specification.
```
since we don't have any other @see or anything for it otherwise.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/UnknownLocalException.java
```diff
@@ -3,11 +3,7 @@
 package com.zeroc.Ice;
 
 /**
- * This exception is raised if an operation call on a server raises a local exception. Because local
- * exceptions are not transmitted by the Ice protocol, the client receives all local exceptions
- * raised by the server as {@link UnknownLocalException}. The only exception to this rule are all
- * exceptions derived from {@link DispatchException}, which are transmitted by the Ice protocol even
- * though they are declared <code>local</code>.
+ * The exception that is thrown when a dispatch failed with a LocalException that is not a DispatchException.
```
InsertCreativityHere: Same here, since JavaDoc doesn't autolink:
```suggestion
 * The exception that is thrown when a dispatch failed with a {@link LocalException} that is not a DispatchException.
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/UnknownException.java
```diff
@@ -3,8 +3,8 @@
 package com.zeroc.Ice;
 
 /**
- * This exception is raised if an operation call on a server raises an unknown exception, that is,
- * any exception which does not derive from {@link LocalException} or {@link UserException}.
+ * The exception that is thrown when a dispatch failed with an exception that is not a LocalException or a
+ * UserException.
```
InsertCreativityHere: Same, old links were nice to have:
```suggestion
 * The exception that is thrown when a dispatch failed with an exception that is not a {@link LocalException} or a
 * {@link UserException}.
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -102,16 +100,16 @@ public static Properties createProperties(
     }
 
     /**
-     * Creates a communicator using a default configuration.
+     * Creates a new communicator.
      *
-     * @return A new communicator instance.
```
InsertCreativityHere, line 107: This should probably be lower-case, but can be fixed in a future PR.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -285,11 +283,10 @@ public static Identity stringToIdentity(String s) {
...
 
     /**
-     * Converts an object identity to a string.
+     * Converts an Identity into a string using the specified mode.
      *
      * @param ident The object identity to convert.
-     * @param toStringMode Specifies if and how non-printable ASCII characters are escaped in the
-     *     result.
+     * @param toStringMode Specifies how to handle non-ASCII characters and non-printable ASCII characters.
      * @return The string representation of the object identity.
```
InsertCreativityHere: Should probably be:
`@return The stringified identifier`
to be consistent with the changes made to `stringToIdentifier`, and the C++ doc-comment for `identityToString`.

(Again, ignoring the casing)

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -424,8 +423,8 @@ public static String stringVersion() {
     }
 
     /**
-     * Returns the Ice version as an integer in the form <code>A.BB.CC</code>, where <code>A</code>
-     * indicates the major version, <code>BB</code> indicates the minor version, and <code>CC</code>
+     * Returns the Ice version as an integer in the form {@code A.BB.CC}, where {@code A}
+     * indicates the major version, {@code BB} indicates the minor version, and {@code CC}
      * indicates the patch level. For example, for Ice 3.3.1, the returned value is 30301.
      *
      * @return The Ice version.
```
InsertCreativityHere: This comment is completely incorrect. It returns versions in the form `AABBCC`, not `A.BB.CC`.
To be fair, the comment was already incorrect before this PR.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
```diff
@@ -211,108 +210,108 @@ CompletableFuture<Object.Ice_invokeResult> ice_invokeAsync(
...
-     * Returns the identity embedded in this proxy.
+     * Gets the identity embedded in this proxy.
      *
      * @return The identity of the target object.
      */
     Identity ice_getIdentity();
 
     /**
-     * Returns a proxy that is identical to this proxy, except for the identity.
+     * Creates a proxy that is identical to this proxy, except for the identity.
```
InsertCreativityHere: Every other file was updated to say `Creates a new`, instead of just `Creates`.
```suggestion
     * Creates a new proxy that is identical to this proxy, except for the identity.
```
But for some reason, this file's comments omit the `new` in many places.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
```diff
@@ -211,108 +210,108 @@ CompletableFuture<Object.Ice_invokeResult> ice_invokeAsync(
...
      */
     ObjectPrx ice_identity(Identity newIdentity);
 
     /**
-     * Returns the per-proxy context for this proxy.
+     * Gets the per-proxy context for this proxy.
      *
      * @return The per-proxy context. If the proxy does not have a per-proxy (implicit) context, the
-     *     return value is <code>null</code>.
+     *     return value is null.
```
InsertCreativityHere: It's interesting. Every other keyword wrapped in `<code>` tags was mapped to use the new `{@code }` tag instead... except for `null`. I'm not aware of any reason why `null` is less deserving of `{@code }` than `true`.
But all of the `null`s in this file had their code formatting stripped.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
```diff
@@ -53,77 +50,75 @@ public interface Connection {
...
      */
     String type();
 
     /**
-     * Return a description of the connection as human readable text, suitable for logging or error
+     * Gets a description of the connection as human readable text, suitable for logging or error
      * messages.
      *
      * @return The description of the connection as human readable text.
+     * @remark This method remains usable after the connection is closed or aborted.
```
InsertCreativityHere: `@remark` is a special Doxygen thing, not supported by JavaDoc.

InsertCreativityHere (approved):
Reviewed everything except for Communicator, and looks pretty good!
Just some minor things, as usual.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
bernardnormier: We should also update the example to 3.8.1

PR #4158, opened 2025-06-25
Remove Dead Stuff from `slice2swift`

InsertCreativityHere:


cpp/src/slice2swift/Gen.cpp
```diff
@@ -1267,7 +1290,80 @@ Gen::TypesVisitor::visitInterfaceDefEnd(const InterfaceDefPtr&)
 void
 Gen::TypesVisitor::visitOperation(const OperationPtr& op)
 {
-    writeProxyOperation(out, op);
```
InsertCreativityHere, line 1270: This is the only place `writeProxyOperation` was called from, so I inlined it.
The code was copy-pasted from `SwiftUtil.cpp` with no changes.

cpp/src/slice2swift/Gen.cpp
```diff
@@ -1434,5 +1530,63 @@ Gen::ServantExtVisitor::visitInterfaceDefEnd(const InterfaceDefPtr&)
 void
 Gen::ServantExtVisitor::visitOperation(const OperationPtr& op)
 {
-    writeDispatchOperation(out, op);
```
InsertCreativityHere, line 1437: This is the only place `writeDispatchOperation` was called from, so I inlined it.
The code was copy-pasted from `SwiftUtil.cpp` with no changes.

cpp/src/slice2swift/Gen.cpp
```diff
@@ -25,6 +26,28 @@ namespace
         assert(dc);
         return dc->getMetadataArgs("swift:class-resolver-prefix").value_or("");
     }
+
```
InsertCreativityHere, line 29: This function was copy-pasted from `SwiftUtil.cpp`, since it is no longer used in that file.

cpp/src/slice2swift/Gen.h
```diff
@@ -28,7 +28,7 @@ namespace Slice
         std::vector<std::string> _includePaths;
         std::string _fileBase;
```
InsertCreativityHere, line 30: `SwiftGenerator` was removed in this PR.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -900,15 +878,9 @@ SwiftGenerator::writeMemberwiseInitializer(
 }
 
 void
-SwiftGenerator::writeMembers(
```
InsertCreativityHere, line 903: This function has a 4th parameter: `int typeCtx` which had a default value of `0`.
None of the callers ever passed it, so `typeCtx` was always `0`.

cpp/src/slice2swift/SwiftUtil.h
```diff
@@ -8,94 +8,78 @@
 
 using StringPairList = std::list<std::pair<std::string, std::string>>;
 
-namespace Slice
+namespace Slice::Swift
 {
-    const int TypeContextInParam = 1;
-    const int TypeContextProtocol = 2;
```
InsertCreativityHere, line 14: These were only used alongside the `typeCtx` parameter of `writeMembers`.
That parameter was removed, so these constants are dead now.

cpp/src/slice2swift/SwiftUtil.h
```diff
@@ -8,94 +8,78 @@
...
-
     std::string getSwiftModule(const ModulePtr&, std::string&);
     std::string getSwiftModule(const ModulePtr&);
```
InsertCreativityHere, line 15: Every single function in this class was effectively static, and the class held no data.
So I turned it into what it logically was: a nested namespace; there was no reason for it to be a class.

cpp/src/slice2swift/SwiftUtil.h
```diff
@@ -8,94 +8,78 @@
...
 
-    class SwiftGenerator
-    {
-    public:
-        SwiftGenerator() = default;
-        SwiftGenerator(const SwiftGenerator&) = delete;
-        virtual ~SwiftGenerator() = default;
-
-        SwiftGenerator& operator=(const SwiftGenerator&) = delete;
```
InsertCreativityHere, line 26: The diff is really struggling here. All I changed was removing these constructors, destructors, and operators.

For the other methods in this file, I removed `static` if it was present, and dedented them.
I didn't change anything else about them.

cpp/src/slice2swift/SwiftUtil.h
```diff
@@ -8,94 +8,78 @@
...
-
-        bool usesMarshalHelper(const TypePtr&);
-        void writeMarshalInParams(::IceInternal::Output&, const OperationPtr&);
-        void writeMarshalAsyncOutParams(::IceInternal::Output&, const OperationPtr&);
-        void writeUnmarshalInParams(::IceInternal::Output&, const OperationPtr&);
-        void writeUnmarshalOutParams(::IceInternal::Output&, const OperationPtr&);
-        void writeUnmarshalUserException(::IceInternal::Output& out, const OperationPtr&);
-        void writeSwiftAttributes(::IceInternal::Output&, const MetadataList&);
-        void writeProxyOperation(::IceInternal::Output&, const OperationPtr&);
-        void writeDispatchOperation(::IceInternal::Output&, const OperationPtr&);
```
InsertCreativityHere, line 97: Well, except for these 2 functions, which were inlined and deleted from here.
Other than them, there were no other changes here.

bernardnormier (approved):
Looks good.

PR #4161, opened 2025-06-26
Output parameters on separate lines and add missing params to checkedCastAsync (C#)

bernardnormier:
This PR updates spar in IceInternal::Output to give the option to output parameters on separate lines (one param per line), as opposed to all parameters on the same line.

Then, applied this option to slice2cs for generated proxy methods, and server-side AMD methods.

For example:
```csharp
        /// <summary>Finds an object by identity and returns a dummy proxy with the endpoint(s) that can be used to reach this
        /// object. This dummy proxy may be an indirect proxy that requires further resolution using
        /// <see cref="global::Ice.LocatorPrx.findAdapterById" />.</summary>
        /// <param name="id">The identity.</param>
        /// <returns>A dummy proxy, or null if an object with the requested identity was not found.</returns>
        /// <exception cref="Ice.ObjectNotFoundException">Thrown when an object with the requested identity was not found. The caller
        /// should treat this exception like a null return value.</exception>
        Ice.ObjectPrx? findObjectById(
            Identity id,
            global::System.Collections.Generic.Dictionary<string, string>? context = null);

        /// <summary>Finds an object by identity and returns a dummy proxy with the endpoint(s) that can be used to reach this
        /// object. This dummy proxy may be an indirect proxy that requires further resolution using
        /// <see cref="global::Ice.LocatorPrx.findAdapterById" />.</summary>
        /// <param name="id">The identity.</param>
        /// <param name="progress">The sent progress provider.</param>
        /// <param name="cancel">A cancellation token that receives the cancellation requests.</param>
        /// <returns>A task that represents the asynchronous operation.</returns>
        /// <exception cref="Ice.ObjectNotFoundException">Thrown when an object with the requested identity was not found. The caller
        /// should treat this exception like a null return value.</exception>
        global::System.Threading.Tasks.Task<Ice.ObjectPrx?> findObjectByIdAsync(
            Identity id,
            global::System.Collections.Generic.Dictionary<string, string>? context = null,
            global::System.IProgress<bool>? progress = null,
            global::System.Threading.CancellationToken cancel = default);
```

Also added missing parameters (progress and cancel) to checkedCastAsync.

### DIFF
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -2067,14 +2067,14 @@ Slice::Gen::TypesVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
 
     _out << sp;
     writeDocLine(_out, "summary", checkedCastSummary.str());
-    writeDocLine(_out, R"(param name="b")", "The source proxy.", "param");
-    writeDocLine(_out, R"(param name="ctx")", "The request context.", "param");
+    writeDocLine(_out, R"(param name="proxy")", "The source proxy.", "param");
+    writeDocLine(_out, R"(param name="context")", "The request context.", "param");
     writeDocLine(_out, "returns", checkedCastReturns.str());
     _out << nl << "public static " << name
-         << "Prx? checkedCast(Ice.ObjectPrx? b, global::System.Collections.Generic.Dictionary<string, string>? ctx = "
+         << "Prx? checkedCast(Ice.ObjectPrx? proxy, global::System.Collections.Generic.Dictionary<string, string>? context = "
             "null) =>";
     _out.inc();
-    _out << nl << "b is not null && b.ice_isA(ice_staticId(), ctx) ? new " << name << "PrxHelper(b) : null;";
+    _out << nl << "proxy is not null && proxy.ice_isA(ice_staticId(), context) ? new " << name << "PrxHelper(proxy) : null;";
     _out.dec();
 
     ostringstream checkedCastWithFacetSummary;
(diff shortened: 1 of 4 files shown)


pepone (approved):

cpp/src/slice2cs/Gen.cpp
```diff
@@ -2217,9 +2238,10 @@ Slice::Gen::TypesVisitor::visitOperation(const OperationPtr& p)
         _out << sp;
         writeOpDocComment(p, {"<param name=\"" + context + "\">The request context.</param>"}, false);
         emitObsoleteAttribute(p, _out);
-        _out << nl << retS << " " << name << spar << getParams(p, ns)
-             << ("global::System.Collections.Generic.Dictionary<string, string>? " + context + " = null") << epar
-             << ';';
+        _out << nl << retS << " " << name;
+        _out.spar("(", true);
+        _out << getParams(p, ns);
+        _out << "global::System.Collections.Generic.Dictionary<string, string>? " + context + " = null" << epar << ';';
```
InsertCreativityHere: Every other place where we do this little string concatenation for `context`, we wrap it in parenthesis for clarity:
```suggestion
        _out << ("global::System.Collections.Generic.Dictionary<string, string>? " + context + " = null") << epar << ';';
```

csharp/src/Ice/Proxy.cs
```diff
@@ -1794,80 +1794,79 @@ public static ObjectPrx createProxy(Communicator communicator, string proxyStrin
...
-    /// <param name="facet">The facet for the new proxy.</param>
-    /// <param name="context">The Context map for the invocation.</param>
-    /// <returns>The new proxy with the specified facet.</returns>
+    /// <param name="proxy">The source proxy.</param>
+    /// <param name="facet">A facet name.</param>
+    /// <param name="context">The request context.</param>
+    /// <param name="progress">Sent progress provider.</param>
+    /// <param name="cancel">A cancellation token that receives the cancellation requests.</param>
+    /// <returns>A proxy with the requested type and facet, or null if the source proxy is null or if the target facet
+    /// does not support the requested type.</returns>
```
InsertCreativityHere: (Jumping off Copilot's comment):
We also talk about null source proxies here, when the signature is non-null.

InsertCreativityHere (approved):
Looks good!

externl (approved):

csharp/src/Ice/Proxy.cs
```diff
@@ -1794,80 +1794,79 @@ public static ObjectPrx createProxy(Communicator communicator, string proxyStrin
...
+    /// <returns>A proxy with the requested type, or null if the source proxy is null or if the target object does not
+    /// support the requested type.</returns>
     public static async Task<ObjectPrx?> checkedCastAsync(
         ObjectPrx proxy,
-        Dictionary<string, string>? context = null) =>
-       await proxy.ice_isAAsync("::Ice::Object", context).ConfigureAwait(false) ? proxy : null;
+        Dictionary<string, string>? context = null,
+        IProgress<bool>? progress = null,
+        CancellationToken cancel = default) =>
+       await proxy.ice_isAAsync("::Ice::Object", context, progress, cancel).ConfigureAwait(false) ? proxy : null;
```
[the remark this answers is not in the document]
bernardnormier: Fixed.

cpp/src/slice2cs/Gen.cpp
bernardnormier: Fixed as suggested.

csharp/src/Ice/Proxy.cs
bernardnormier: Fixed

PR #4163, opened 2025-06-27
Remove public from generated interface method in Java

bernardnormier:
In Java, we don't want to add "public" on interface methods.

### DIFF
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -3624,8 +3624,7 @@ Slice::Gen::TypesVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
             "@param communicator The communicator of the new proxy.\n"
             "@param proxyString The string representation of the proxy.\n"
             "@return The new proxy.");
-    out << nl << "public static " << prxName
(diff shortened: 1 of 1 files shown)


cpp/src/slice2java/Gen.cpp
```diff
@@ -3624,8 +3624,7 @@ Slice::Gen::TypesVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
             "@param communicator The communicator of the new proxy.\n"
             "@param proxyString The string representation of the proxy.\n"
             "@return The new proxy.");
-    out << nl << "public static " << prxName
-        << " createProxy(com.zeroc.Ice.Communicator communicator, String proxyString)";
+    out << nl << "static " << prxName << " createProxy(com.zeroc.Ice.Communicator communicator, String proxyString)";
```
pepone, line 3627: is this because they are implicitly public?

pepone (approved):

cpp/src/slice2java/Gen.cpp
bernardnormier: Yes, and the preferred style in Java is to omit public for interface methods.

PR #4166, opened 2025-06-28
Matlab code cleanup

bernardnormier:
Small updates to the MATLAB code base, mostly reformatting.

### DIFF
--- a/cpp/src/Slice/SliceUtil.cpp
+++ b/cpp/src/Slice/SliceUtil.cpp
@@ -577,3 +577,19 @@ Slice::getEscapedParamName(const ParameterList& params, std::string_view param)
     }
     return string{param};
 }
+
+string
+Slice::getTypeName(const TypePtr& type)
+{
+    if (auto builtin = std::dynamic_pointer_cast<Builtin>(type))
+    {
+        return builtin->kindAsString();
+    }
+    else
+    {
+        // a type is either built-in or contained
+        auto contained = std::dynamic_pointer_cast<Contained>(type);
+        assert(contained);
+        return contained->scoped();
+    }
+}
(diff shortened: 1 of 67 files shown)


matlab/lib/+Ice/EndpointInfo.m
```diff
@@ -1,4 +1,4 @@
-classdef EndpointInfo < handle
```
bernardnormier, line 1: EndpointInfo and ConnectionInfo are immutable classes that don't need to be handles ... this was inherited from local Slice.

pepone (approved):

matlab/lib/+Ice/InputStream.m
```diff
@@ -828,7 +828,7 @@ function initEncaps(obj)
             end
         end
     end
-    methods (Static, Access = private)
+    methods (Static,  Access = private)
```
InsertCreativityHere: Extra space here:
```suggestion
    methods (Static, Access = private)
```

matlab/lib/+Ice/UserException.m
```diff
@@ -26,7 +26,7 @@
             %
         end
     end
-    methods(Abstract,Access=protected)
+    methods (Abstract,Access = protected)
```
InsertCreativityHere: Missing a space:
```suggestion
    methods (Abstract, Access = protected)
```

cpp/src/Slice/SliceUtil.cpp
```diff
@@ -577,3 +577,19 @@ Slice::getEscapedParamName(const ParameterList& params, std::string_view param)
     }
     return string{param};
 }
+
+string
+Slice::getTypeName(const TypePtr& type)
```
InsertCreativityHere: I'd prefer `getSliceTypeString` here.
By convention we use`name` to refer to unscoped/unqualified names.

InsertCreativityHere (approved):
Looks good to me!
This is a nice function to have.

PR #4167, opened 2025-06-30
Cleanup `JsUtil` in `slice2js`

InsertCreativityHere:
The `JsGenerator` class has no members, and only static functions.
This PR replaces this class with what it logically is: a nested namespace.

I also added `[[nodiscard]]`s and parameter names where appropriate.

### DIFF
--- a/cpp/src/slice2js/Gen.h
+++ b/cpp/src/slice2js/Gen.h
@@ -7,7 +7,7 @@
 
 namespace Slice
 {
-    class JsVisitor : public JsGenerator, public ParserVisitor
+    class JsVisitor : public ParserVisitor
     {
     public:
         JsVisitor(::IceInternal::Output& output);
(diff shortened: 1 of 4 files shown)


cpp/src/slice2js/Gen.h
```diff
@@ -7,7 +7,7 @@
 
 namespace Slice
 {
```
InsertCreativityHere, line 9: `JsGenerator` is deleted by this PR.

cpp/src/slice2js/JsUtil.h
```diff
@@ -6,30 +6,26 @@
 #include "../Ice/OutputUtil.h"
 #include "../Slice/Parser.h"
 
-namespace Slice
+namespace Slice::JavaScript
 {
```
InsertCreativityHere, line 10: I added `[[nodiscard]]`, removed `static`, and added names to any parameters which were missing them.
These are the only changes I made to these methods.

PR #4168, opened 2025-06-30
Cleanup `CsUtil` in `slice2cs`

InsertCreativityHere:
The `CsGenerator` class has no members, and all its functions are effectively static.
This PR replaces this class with what it logically is: a nested namespace.

I also added `[[nodiscard]]`s and parameter names where appropriate.

### DIFF
--- a/cpp/src/slice2cs/CsUtil.cpp
+++ b/cpp/src/slice2cs/CsUtil.cpp
@@ -22,33 +22,28 @@ using namespace Slice;
 using namespace IceInternal;
 
 string
-Slice::CsGenerator::getNamespacePrefix(const ContainedPtr& cont)
+Slice::Csharp::getNamespacePrefix(const ContainedPtr& p)
 {
-    // Traverse to the top-level module.
-    ContainedPtr p = cont;
-    while (!p->isTopLevel())
+    if (auto topLevel = p->getTopLevelModule())
     {
-        p = dynamic_pointer_cast<Contained>(p->container());
-        assert(p);
+        return topLevel->getMetadataArgs("cs:namespace").value_or("");
     }
-    assert(dynamic_pointer_cast<Module>(p));
-
-    return p->getMetadataArgs("cs:namespace").value_or("");
+    return "";
 }
 
 string
-Slice::CsGenerator::getNamespace(const ContainedPtr& cont)
+Slice::Csharp::getNamespace(const ContainedPtr& p)
 {
-    assert(!dynamic_pointer_cast<Module>(cont));
+    assert(!dynamic_pointer_cast<Module>(p));
 
-    string scope = cont->mappedScope(".");
+    string scope = p->mappedScope(".");
     scope.pop_back(); // Remove the trailing '.' on the scope.
-    string prefix = getNamespacePrefix(cont);
+    string prefix = getNamespacePrefix(p);
     return (prefix.empty() ? scope : prefix + "." + scope);
 }
 
 string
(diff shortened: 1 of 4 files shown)


cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -22,33 +22,28 @@ using namespace Slice;
...
+    if (auto topLevel = p->getTopLevelModule())
     {
-        p = dynamic_pointer_cast<Contained>(p->container());
-        assert(p);
+        return topLevel->getMetadataArgs("cs:namespace").value_or("");
     }
-    assert(dynamic_pointer_cast<Module>(p));
-
-    return p->getMetadataArgs("cs:namespace").value_or("");
+    return "";
```
InsertCreativityHere, line 31: We have a dedicated `getTopLevelModule` function now to replace this logic.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -22,33 +22,28 @@ using namespace Slice;
...
 }
 
 string
-Slice::CsGenerator::getNamespace(const ContainedPtr& cont)
+Slice::Csharp::getNamespace(const ContainedPtr& p)
```
InsertCreativityHere, line 35: `cont` is a bad name. Is it a `Contained`? a `Container`?

cpp/src/slice2cs/CsUtil.h
```diff
@@ -6,88 +6,78 @@
...
+namespace Slice::Csharp
 {
-    class CsGenerator
-    {
-    public:
-        CsGenerator() = default;
-        CsGenerator(const CsGenerator&) = delete;
-        virtual ~CsGenerator() = default;
-
-        CsGenerator& operator=(const CsGenerator&) = delete;
```
InsertCreativityHere, line 18: `CsGenerator` was deleted, and so were these useless methods.

cpp/src/slice2cs/CsUtil.h
```diff
@@ -6,88 +6,78 @@
 #include "../Ice/OutputUtil.h"
 #include "../Slice/Parser.h"
 
-namespace Slice
+namespace Slice::Csharp
```
InsertCreativityHere, line 9: Unfortunately, the git diff is struggling here.
All I did was add `[[nodiscard]]`s, remove `static`, and added parameter names where they were missing.
Other, the function signatures and doc-comments are exactly the same.

cpp/src/slice2cs/CsUtil.h
```diff
@@ -6,88 +6,78 @@
...
+    [[nodiscard]] std::string resultType(const OperationPtr& op, const std::string& package, bool dispatch = false);
+    [[nodiscard]] std::string taskResultType(const OperationPtr& op, const std::string& scope, bool dispatch = false);
+    [[nodiscard]] std::string getOptionalFormat(const TypePtr& type);
+    [[nodiscard]] std::string getStaticId(const TypePtr& type);
+
+    /// Is this Slice type mapped to a C# value type?
+    [[nodiscard]] bool isValueType(const TypePtr& type);
+
+    /// Is this Slice struct mapped to a C# class?
+    [[nodiscard]] inline bool isMappedToClass(const StructPtr& p) { return !isValueType(p); }
```
InsertCreativityHere, line 38: The one exception is that I `inline`d this function, since it's very short and no longer a member function.

bernardnormier: You would most likely get link errors without this inline.

PR #4172, opened 2025-06-30
Minor Cleanup in `slice2java`

InsertCreativityHere:
Unlike the other compilers, the `JavaGenerator` class does hold some state, so it's not completely useless like in the other compilers.

But I still went ahead and removed some dead functions from it, and added `[[nodiscard]]` and parameter names where they should be.

### DIFF
--- a/cpp/src/slice2java/JavaUtil.cpp
+++ b/cpp/src/slice2java/JavaUtil.cpp
@@ -284,7 +284,7 @@ Slice::JavaGenerator::open(const string& qualifiedEntity, const string& sliceFil
 {
     assert(_out == nullptr);
 
-    JavaOutput* out = new JavaOutput;
+    auto* out = new JavaOutput;
     try
     {
         out->openClass(qualifiedEntity, _dir, sliceFile);
--- a/cpp/src/slice2java/JavaUtil.cpp
+++ b/cpp/src/slice2java/JavaUtil.cpp
@@ -156,10 +156,6 @@ Slice::mapsToJavaBuiltinType(const TypePtr& p)
 
 Slice::JavaOutput::JavaOutput() = default;
 
-Slice::JavaOutput::JavaOutput(ostream& os) : Output(os) {}
-
-Slice::JavaOutput::JavaOutput(const char* s) : Output(s) {}
-
 void
 Slice::JavaOutput::openClass(const string& cls, const string& prefix, const string& sliceFile)
 {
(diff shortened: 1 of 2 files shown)


cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -156,10 +156,6 @@ Slice::mapsToJavaBuiltinType(const TypePtr& p)
 
 Slice::JavaOutput::JavaOutput() = default;
 
-Slice::JavaOutput::JavaOutput(ostream& os) : Output(os) {}
-
-Slice::JavaOutput::JavaOutput(const char* s) : Output(s) {}
```
InsertCreativityHere, line 161: These constructors were never called.

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -288,7 +284,7 @@ Slice::JavaGenerator::open(const string& qualifiedEntity, const string& sliceFil
 {
     assert(_out == nullptr);
 
-    JavaOutput* out = createOutput();
```
InsertCreativityHere, line 291: `createOuput` was inlined here.
It's a uselessly small function that's only called in this one spot.

cpp/src/slice2java/JavaUtil.h
```diff
@@ -141,81 +139,86 @@ namespace Slice
             OptionalMember
         };
 
-        std::string getWriteFunction(const std::string&, const TypePtr&);
-        std::string getReadFunction(const std::string&, const TypePtr&);
```
InsertCreativityHere, line 145: These functions aren't defined anywhere. Not sure what they're from.

cpp/src/slice2java/JavaUtil.h
```diff
@@ -141,81 +139,86 @@ namespace Slice
...
         //
-        static bool hasTypeMetadata(const SequencePtr&, const MetadataList& = MetadataList());
+        [[nodiscard]] static bool
+        hasTypeMetadata(const SequencePtr& seq, const MetadataList& localMetadata = MetadataList());
 
         //
         // Obtain the concrete and abstract types for a dictionary or sequence type.
         // The functions return true if a custom type was defined and false to indicate
         // the default mapping was used.
         //
```
InsertCreativityHere, line 209: The bool these functions return is only useful sometimes.
So, I didn't add `nodiscard`, since we do actually discard it's return value sometimes.

externl (approved):
Looks good. Would be good to move these functions outside of the class. No need for them to be static member functions.

PR #4173, opened 2025-06-30
Refactor `slice2java` Function Placement

InsertCreativityHere:
This PR only moves existing code, it does not alter any implementations, nor functionality.

Functions which `write` to an `Output` should be in `Gen`, since they generate code.
Any other functions which do not, belong in `JavaUtil`.
This PR moves all the functions of `slice2java` around according to this rule.

Due to the huge amount of code moving around, git diff is completely useless.
Due to the reviewability of this PR, I'm only going to rely on CoPilot and my own pedantic self-review. Feel free to though.

### DIFF
--- a/cpp/src/slice2java/Gen.h
+++ b/cpp/src/slice2java/Gen.h
@@ -106,19 +106,26 @@ namespace Slice
 
         static void writeMarshalProxyParams(IceInternal::Output&, const std::string&, const OperationPtr&, bool);
         static void writeUnmarshalProxyResults(IceInternal::Output&, const std::string&, const OperationPtr&);
-        static void writeMarshalServantResults(IceInternal::Output&, const std::string&, const OperationPtr&, const std::string&);
+        static void
+        writeMarshalServantResults(IceInternal::Output&, const std::string&, const OperationPtr&, const std::string&);
 
         //
         // Generate a throws clause containing only checked exceptions.
         // op is provided only when we want to check for the java:UserException metadata
         //
-        static void writeThrowsClause(IceInternal::Output&, const std::string&, const ExceptionList&, const OperationPtr& op = nullptr);
+        static void writeThrowsClause(
+            IceInternal::Output&,
+            const std::string&,
+            const ExceptionList&,
+            const OperationPtr& op = nullptr);
 
         //
         // Marshal/unmarshal a data member.
         //
-        static void writeMarshalDataMember(IceInternal::Output&, const std::string&, const DataMemberPtr&, int&, bool = false);
-        static void writeUnmarshalDataMember(IceInternal::Output&, const std::string&, const DataMemberPtr&, int&, bool = false);
+        static void
+        writeMarshalDataMember(IceInternal::Output&, const std::string&, const DataMemberPtr&, int&, bool = false);
+        static void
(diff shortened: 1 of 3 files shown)


cpp/src/slice2java/Gen.h
```diff
@@ -8,89 +8,129 @@
...
+        OptionalMember
+    };
+
     class JavaVisitor : public JavaGenerator, public ParserVisitor
     {
     public:
         ~JavaVisitor() override;
 
-    protected:
-        JavaVisitor(const std::string&);
```
InsertCreativityHere, line 17: This `protected` section was moved to the bottom of the class.
There is no reason for all the helper functions to be protected instead of plain 'ol public here.

cpp/src/slice2java/Gen.h
```diff
@@ -8,89 +8,129 @@
...
-
-        enum ParamDir
-        {
-            InParam,
-            OutParam
-        };
```
InsertCreativityHere, line 23: This enum was never used. So I deleted it.

cpp/src/slice2java/Gen.h
```diff
@@ -8,89 +8,129 @@
...
+            bool optionalMapping,
+            std::int32_t tag,
+            const std::string& param,
+            bool marshal,
+            int& iter,
+            const std::string& customStream = "",
+            const MetadataList& metadata = MetadataList(),
+            const std::string& patchParams = "");
 
-        std::string getResultType(const OperationPtr&, const std::string&, bool, bool);
```
InsertCreativityHere, line 25: Otherwise, the rest of these functions are preserved exactly as they were, except I added `static` to the ones which didn't access any member variables. That's the only change though. I didn't even fix the comments, or add parameter names in this PR. Baby steps.

cpp/src/slice2java/Gen.h
```diff
@@ -8,89 +8,129 @@
...
-        std::string getPatcher(const TypePtr&, const std::string&, const std::string&);
-
-        /// Returns a vector of this operation's parameters with each of them formatted as 'paramType paramName'.
-        /// If 'internal' is true, the names will be prefixed with "iceP_".
-        std::vector<std::string>
-        getParamsProxy(const OperationPtr& op, const std::string& package, bool optionalMapping, bool internal = false);
-
-        /// Returns a vector of this operation's parameter's names in order.
-        /// If 'internal' is true, the names will be prefixed with "iceP_".
-        std::vector<std::string> getInArgs(const OperationPtr& op, bool internal = false);
```
InsertCreativityHere, line 56: These 3 functions were helper functions, not code-gen.
They were moved to `JavaUtil`.

cpp/src/slice2java/JavaUtil.h
```diff
@@ -8,106 +8,72 @@
...
-        JavaGenerator(const JavaGenerator&) = delete;
-        virtual ~JavaGenerator();
-
-        JavaGenerator& operator=(const JavaGenerator&) = delete;
-
-        static void validateMetadata(const UnitPtr&);
-
-        void close();
-
-        JavaGenerator(std::string);
```
InsertCreativityHere, line 74: Most of the methods in this class were static, and this PR moves them outside the class, converting them into free functions.
Git diff interprets this as if I cut-and-pasted this section at the bottom of the file.

PR #4175, opened 2025-07-01
Add test for proxy hierarchy

pepone:
This PR adds a new test for Slice interface inheritance, specially for the Proxy mapping.

It also fixes the generated TypeScript definition to correctly support proxy hierarchies.

### DIFF
--- a/ruby/test/Ice/proxy/AllTests.rb
+++ b/ruby/test/Ice/proxy/AllTests.rb
@@ -173,7 +173,7 @@ def allTests(helper, communicator)
         test(false)
     rescue Ice::ParseException
     end
-    # This is an unknown endpoint warning, not a parse rescueion.
+    # This is an unknown endpoint warning, not a parse exception.
     #
     #begin
     #   b1 = communicator.stringToProxy("test -f the:facet:tcp")
@@ -816,15 +816,17 @@ def allTests(helper, communicator)
     cPrx = Test::CPrx.new(communicator, "c:#{helper.getTestEndpoint()}")
 
     aPrx = cPrx.opA(cPrx)
-    test(Test::CPrx.checkedCast(aPrx) != nil)
+    test(aPrx == cPrx)
 
     bPrx = cPrx.opB(cPrx)
-    test(Test::CPrx.checkedCast(bPrx) != nil)
+    test(bPrx == cPrx)
 
     cPrx = cPrx.opC(cPrx)
 
     s = Test::S.new(cPrx, cPrx)
     s = cPrx.opS(s)
+    test(s.a == cPrx)
+    test(s.b == cPrx)
     puts "ok"
 
     print "testing communicator shutdown/destroy... "


cpp/src/slice2js/Gen.cpp
```diff
@@ -2347,9 +2347,28 @@ Slice::Gen::TypeScriptVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
...
     //
     const string prxName = p->mappedName() + "Prx";
+    InterfaceList bases = p->bases();
     _out << sp;
     writeDocCommentFor(p);
     _out << nl << "export class " << prxName << " extends " << _iceImportPrefix << "Ice.ObjectPrx";
+    if (!bases.empty())
+    {
+        _out << " implements ";
+        bool first = true;
```
bernardnormier: I find `_out.spar("")` followed by `_out.epar("")` more elegant.

cpp/src/slice2js/Gen.cpp
```diff
@@ -2466,6 +2476,24 @@ Slice::Gen::TypeScriptVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     _out << sp;
     writeDocCommentFor(p, true, false);
     _out << nl << "export abstract class " << p->mappedName() << " extends " << _iceImportPrefix << "Ice.Object";
+    if (!bases.empty())
+    {
+        _out << " implements ";
+
+        bool first = true;
+        for (const auto& base : bases)
```
bernardnormier, line 2476: Same here.

bernardnormier (commented):
As part of this PR, you removed:

- the AMD portion of the Ice/proxy test in all (several) languages

This makes sense since the Ice/proxy test is focused on client-side proxy operations. The server-side implementation is / should be irrelevant to this test

- The Collocated test

It's not clear to me that it's a useless test. In particular, we have various branches in the test that checks if the connection is null ... and now are dead code. Maybe you should undo this change?

cpp/test/Ice/proxy/AllTests.cpp
```diff
@@ -1226,6 +1226,26 @@ allTests(TestHelper* helper)
     cout << "ok" << endl;
 #endif
 
+    cout << "testing proxy hierarchy... " << flush;
+    {
+        optional<CPrx> cPrx = Test::CPrx(communicator, "c:" + helper->getTestEndpoint());
+
+        optional<Test::APrx> aPrx = cPrx->opA(cPrx);
+        test(Ice::checkedCast<Test::CPrx>(aPrx) != std::nullopt);
```
bernardnormier: I don't find the checkedCast meaningful here, especially since checkedCast should be very rare in C++.
Can we instead test:
```cpp
test(aPtr == cPrx);
```
?

cpp/test/Ice/proxy/AllTests.cpp
```diff
@@ -1226,6 +1226,26 @@ allTests(TestHelper* helper)
...
+
+        optional<Test::BPrx> bPrx = cPrx->opB(cPrx);
+        test(Ice::checkedCast<Test::CPrx>(bPrx) != std::nullopt);
```
bernardnormier: Ditto

cpp/test/Ice/proxy/AllTests.cpp
```diff
@@ -1226,6 +1226,26 @@ allTests(TestHelper* helper)
...
+
+        cPrx = cPrx->opC(cPrx);
+
+        Test::S s{
+            .a = cPrx,
+            .b = cPrx,
+        };
+        s = cPrx->opS(s);
```
bernardnormier, line 1245: Should we actually verify the returned struct is equal to s?

csharp/test/Ice/proxy/AllTests.cs
```diff
@@ -1024,6 +1024,21 @@ public class AllTests : global::Test.AllTests
 
         output.WriteLine("ok");
 
+        output.Write("testing proxy hierarchy... ");
+        var cPrx = Test.CPrxHelper.createProxy(communicator, "c:" + helper.getTestEndpoint());
+
+        Test.APrx aPrx = cPrx.opA(cPrx);
```
bernardnormier, line 1030: Same comment as C++.

externl (approved):

InsertCreativityHere (approved):
Looks good to me!

PR #4176, opened 2025-07-01
Move ice-compilers to the language package

pepone:
This PR moves the Slice compilers to the corresponding language packages

slice2cpp -> libice-c++-devel
slice2py -> python3-ice
slice2php -> php-ice
ice2slice -> ice2slice (new package)

We keep ice-compilers as a transitional package for clean upgrade from 3.7

I didn't add php/python -devel packages because this seems uncommon to do.

Slice compilers that doesn't belong to one of these packages are no longer installed.

```
slice2cs
slice2java
slice2js
slice2matlab
slice2rb
slice2swift
```


packaging/rpm/ice.spec
```diff
@@ -203,13 +209,14 @@ network programming interfaces and allows you to focus your efforts on
...
 #
-%package -n %{?nameprefix}ice-compilers
-Summary: Slice compilers for developing Ice applications
+%package -n %{?nameprefix}ice2slice
+Summary: The ice2slice compiler
 Requires: %{?nameprefix}ice-slice = %{version}-%{release}
-%description -n %{?nameprefix}ice-compilers
-This package contains Slice compilers for developing Ice applications.
+%description -n %{?nameprefix}ice2slice
+This package contains the ice2slice compiler to transforms Slice definitions in .ice files into
```
bernardnormier: This package contains the Ice-to-Slice compiler (ice2slice). The Ice-to-Slice compiler converts Slice definitions in .ice files into Slice definitions in .slice files. 
.ice is the Slice syntax and format understood by the Slice compilers provided by Ice;  .slice is the Slice syntax and format understood by slicec, the Slice compiler provided by IceRPC.

packaging/rpm/ice.spec
```diff
@@ -376,6 +383,13 @@ make -C java DESTDIR=%{buildroot} install_javadir=%{_javadir} prefix=%{_prefix}
 rm -f %{buildroot}%{_datadir}/ice/LICENSE
 rm -f %{buildroot}%{_datadir}/ice/ICE_LICENSE
 
+rm -f %{buildroot}%{_bindir}/slice2cs
```
bernardnormier: Maybe we should do some `make install` work to avoid these rm?

bernardnormier (approved):

packaging/rpm/ice.spec
```diff
@@ -203,13 +209,14 @@ network programming interfaces and allows you to focus your efforts on
...
-%package -n %{?nameprefix}ice-compilers
-Summary: Slice compilers for developing Ice applications
+%package -n %{?nameprefix}ice2slice
+Summary: The ice2slice compiler
 Requires: %{?nameprefix}ice-slice = %{version}-%{release}
-%description -n %{?nameprefix}ice-compilers
-This package contains Slice compilers for developing Ice applications.
+%description -n %{?nameprefix}ice2slice
+This package contains the ice2slice compiler to transforms Slice definitions in .ice files into
+Slice definitions in .slice files.
```
externl:
```suggestion
This package includes the ice2slice compiler, which transforms Slice definitions stored in .ice files into Slice definitions stored in .slice files.
```

packaging/rpm/ice.spec
```diff
@@ -376,6 +383,13 @@ make -C java DESTDIR=%{buildroot} install_javadir=%{_javadir} prefix=%{_prefix}
 rm -f %{buildroot}%{_datadir}/ice/LICENSE
 rm -f %{buildroot}%{_datadir}/ice/ICE_LICENSE
 
+rm -f %{buildroot}%{_bindir}/slice2cs
+rm -f %{buildroot}%{_bindir}/slice2java
+rm -f %{buildroot}%{_bindir}/slice2js
+rm -f %{buildroot}%{_bindir}/slice2matlab
+rm -f %{buildroot}%{_bindir}/slice2rb
+rm -f %{buildroot}%{_bindir}/slice2swift
```
externl: Maybe we should update `make install` to no install these binaries.

packaging/rpm/ice.spec
externl: Why is this necessary do we include `${_bindir}/*`?

externl (approved):

packaging/rpm/ice.spec
pepone: The RPM packaging would complain if there are binaries in `%{buildroot}%{_bindir}` which are not installed with any package.

pepone: I will try, and update this if it doesn't turn to be too complicated with the Makefiles.

PR #4177, opened 2025-07-01
Fix Swift Doc-Link Generation

InsertCreativityHere:
This PR updates the `swiftLinkFormatter` to actually work.
To link a symbol in DocC, you wrap it in double backticks (not what we're doing right now).
There's also some small cleanup included in this PR as well.

I tested linking to data members, operations, enumerators, basically everything except for modules and parameters.
Note that linking to parameters is already disallowed by the parser.
I did stumble across a small bug with linking to constants, but the bug is with how we generate constants, not the link generation logic. See my blog post for more explanation.

----
----

Here's some pictures of the generated docs in a before/after format:
Screenshot 2025-07-01 at 09 57 27
Screenshot 2025-07-01 at 09 57 32

----
----
Screenshot 2025-07-01 at 09 56 35
Screenshot 2025-07-01 at 09 56 45

----
----
Screenshot 2025-07-01 at 09 54 00
Screenshot 2025-07-01 at 09 54 52

Note that when linking to interfaces, we always link to the proxy type.

### DIFF
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -52,7 +52,7 @@ namespace
                 assert(contained);
             }
 
-            // If the link involves an interface definition, we need to switch it to the corresponding declaration.
+            // If the link involves an interface definition, we need to switch to its declaration.
             // The code-gen considers `Def` the servant type, and `Decl` the proxy type. We want to link to the proxy.
             if (auto interfaceDef = dynamic_pointer_cast<InterfaceDef>(contained))
             {
@@ -67,9 +67,9 @@ namespace
             string mappedLink = Slice::Swift::getRelativeTypeString(contained, sourceModule) + nameSuffix;
             std::replace(mappedLink.begin(), mappedLink.end(), '.', '/');
 
-            // DocC only supports linking to symbols that are in the same module.
-            // So if the source and target elements are in the same module, we can generate a DocC link
-            // (using double back-ticks). Otherwise, we emit the mapped name in monospace (using single back-ticks).
+            // DocC does not support cross-module linking.
+            // So we can only generate a DocC link (using double back-ticks) if the source and target are in the same
+            // module. Otherwise we emit the mapped name in monospace formatting (using single back-ticks).
             const string ticks = (sourceModule == targetModule ? "``" : "`");
             return ticks + mappedLink + ticks;
         }


cpp/src/slice2swift/Gen.cpp
```diff
@@ -1080,10 +1080,10 @@ Gen::TypesVisitor::visitConst(const ConstPtr& p)
     const TypePtr type = p->type();
     const string swiftModule = getSwiftModule(p->getTopLevelModule());
 
+    out << sp;
     writeDocSummary(out, p);
     out << nl << "public let " << p->mappedName() << ": " << typeToString(type, p) << " = ";
     writeConstantValue(out, type, p->valueType(), p->value(), swiftModule);
```
InsertCreativityHere, line 1086: Unrelated formatting fix.
We were accidentally generating this:
```
import Foundation
import Ice
public let Foo: Swift.Int32 = 45


public extension ... blah blah blah
```
So I moved the space before the constant, instead of after it.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -655,7 +695,7 @@ Slice::Swift::typeToString(const TypePtr& type, const ContainedPtr& usedBy, bool
         "Swift.Float",
```
InsertCreativityHere, line 702: Everything below this comment is unrelated cleanup in `typeToString`.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -665,38 +705,19 @@ Slice::Swift::typeToString(const TypePtr& type, const ContainedPtr& usedBy, bool
...
-        }
-        else
-        {
-            t = getUnqualified(builtinTable[builtin->kind()], currentModule);
-        }
-    }
-
-    ClassDeclPtr cl = dynamic_pointer_cast<ClassDecl>(type);
-    InterfaceDeclPtr prx = dynamic_pointer_cast<InterfaceDecl>(type);
-    ContainedPtr cont = dynamic_pointer_cast<Contained>(type);
```
InsertCreativityHere, line 688: `ClassDecl` and `InterfaceDecl` _are_ `Contained`s.
So, there's no reason to have separate branches for them where we just call the exact same function...

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -665,38 +705,19 @@ Slice::Swift::typeToString(const TypePtr& type, const ContainedPtr& usedBy, bool
...
-    if (builtin)
-    {
-        if (builtin->kind() == Builtin::KindObject)
-        {
-            t = getUnqualified(builtinTable[Builtin::KindValue], currentModule);
-        }
-        else
-        {
-            t = getUnqualified(builtinTable[builtin->kind()], currentModule);
-        }
```
InsertCreativityHere, line 683: This extra if/else here isn't really necessary, since we already modified the corresponding `builtinTable` entry.
You could argue that it's less clear now...
But we already have a comment about it where we define the table, and eliminating the branch is simpler.

externl:
I think we shouldn't generate warnings and should just not include a link if we don't know it.

bernardnormier:
> I think we shouldn't generate warnings and should just not include a link if we don't know it.

If you have doc-comment with `{@link Symbol}` and the parser can't resolve Symbol, it should definitely generate a warning.

Then, a secondary question is what happens next during code generation after such a warning: the code generator should not warn again. And it's ok for the code generator to not generate a proper link in this case.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -15,28 +15,68 @@ using namespace IceInternal;
...
+                nameSuffix = "/" + contained->mappedName();
+
+                // Move up a level to the container type.
+                contained = dynamic_pointer_cast<Contained>(contained->container());
+                assert(contained);
             }
-            result += rawLink.substr(hashPos + 1);
+
+            // If the link is to an operation, DocC requires you to list the operation's labels in the link signature.
+            if (auto operation = dynamic_pointer_cast<Operation>(contained))
```
pepone: Seems this should `else if`

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -15,28 +15,68 @@ using namespace IceInternal;
...
+                nameSuffix += "context:)";
+
+                // Move up a level to the container type.
+                contained = dynamic_pointer_cast<Contained>(contained->container());
+                assert(contained);
+            }
+
+            // If the link was to an interface definition, we need to switch to the corresponding declaration.
+            // The code-gen considers `Def` the servant type, and `Decl` the proxy type. We want to link to the proxy.
+            if (auto interfaceDef = dynamic_pointer_cast<InterfaceDef>(contained))
```
pepone, line 57: is this block also used for operations, after they move up? If it is not an `else if` would make the intention clear, if it is the comment should make it clear.

InsertCreativityHere: Yes, this block is also used by operations after they move up.
I added a comment saying so.

InsertCreativityHere:
I rewrote the link logic.
Now we only emit a link if the source and target are in the same module.
Since DocC can't handle cross-module linking (in general).

Also updated some comments.

pepone (approved):
Looks good!

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -8,35 +8,82 @@
...
-                result += "/";
+                contained = interfaceDef->declaration();
             }
-            result += rawLink.substr(hashPos + 1);
+
+            const string sourceModule = Slice::Swift::getSwiftModule(source->getTopLevelModule());
+            const string targetModule = Slice::Swift::getSwiftModule(contained->getTopLevelModule());
+
+            // Get the mapped-and-qualified name of the thing we're linking to, but replace any '.' with '/'.
+            // DocC uses forward slashes to separate symbols instead of periods.
```
externl, line 66: Because they represent URLs.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -8,35 +8,82 @@
...
-        else
+        else if (auto builtin = dynamic_pointer_cast<Builtin>(target))
         {
-            result += rawLink;
+            // We have no way to link to the builtin types, so we just emit the mapped type in a monospaced format.
+            return "`" + Slice::Swift::typeToString(builtin, source, false) + "`";
+        }
+        else // We couldn't resolve the link target and make a best-effort attempt to map the raw link.
+        {
+            // For Swift all we can do is replace any doxygen separators with the DocC separator ('/').
```
externl, line 83: The compiler issues a warning in such case right?

externl (approved):

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere: Yep, 100%!
https://github.com/zeroc-ice/ice/blob/7fbfb6f2988bc1ede37d3c06c4b61c4061700ae7/cpp/src/Slice/Parser.cpp#L470-L484

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere: Oh, that makes sense actually.

PR #4183, opened 2025-07-02
Use dataclass for mapped structs in Python

bernardnormier:
This PR is an attempt to use `@dataclass` for mapped structs in Python.

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -1261,212 +1261,43 @@ Slice::Python::CodeVisitor::visitStructStart(const StructPtr& p)
     writeModuleHasDefinitionCheck(_out, p, name);
     _out.inc();
     _out << nl << abs << " = None";
-    _out << nl << "class " << name << "(object):";
+    _out << nl << "@dataclass(order=True";
+    if (Dictionary::isLegalKeyType(p))
+    {
+        _out << ", unsafe_hash=True";
+    }
+    _out << ')';
+
+    _out << nl << "class " << name << ":";
     _out.inc();
 
     writeDocstring(DocComment::parseFrom(p, pyLinkFormatter), members);
 
-    _out << nl << "def __init__(self";
-    writeConstructorParams(p->dataMembers());
-    _out << "):";
-    _out.inc();
-    for (const auto& member : members)
+    for (const auto& field : members)
     {
-        writeAssign(member);
-    }
-    _out.dec();
+        _out << nl << field->mappedName() << ": ";
+        writeType(field->type());
 
-    //
-    // Only generate __hash__ and the comparison operators if this structure type is a legal dictionary key type.
-    //
-    if (Dictionary::isLegalKeyType(p))
-    {
-        _out << sp << nl << "def __hash__(self):";
-        _out.inc();
-        _out << nl << "_h = 0";
-        int iter = 0;
-        for (const auto& member : members)
+        if (field->defaultValue())
         {
-            string s = "self." + member->mappedName();
-            writeHash(s, member->type(), iter);
+            _out << " = ";
+            writeConstantValue(field->type(), field->defaultValueType(), *field->defaultValue());
         }
-        _out << nl << "return _h % 0x7fffffff";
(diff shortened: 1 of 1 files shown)


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1261,213 +1261,44 @@ Slice::Python::CodeVisitor::visitStructStart(const StructPtr& p)
     writeModuleHasDefinitionCheck(_out, p, name);
     _out.inc();
     _out << nl << abs << " = None";
-    _out << nl << "class " << name << "(object):";
+    _out << nl << "@dataclass(order=True";
+    if (Dictionary::isLegalKeyType(p))
+    {
+        _out << ", unsafe_hash=True";
```
bernardnormier: This is for compatibility with the existing generated code (as I understand it), but I don't find it correct. Marking the datataclass as frozen and therefore hashable would be more correct but less backwards compatible.

See hash discussion on https://docs.python.org/3/library/dataclasses.html.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1261,213 +1261,44 @@ Slice::Python::CodeVisitor::visitStructStart(const StructPtr& p)
...
-    _out << "):";
-    _out.inc();
-    for (const auto& member : members)
+    for (const auto& field : members)
     {
-        writeAssign(member);
-    }
-    _out.dec();
+        _out << nl << field->mappedName() << ": ";
+        writeType(field->type());
```
bernardnormier, line 1278: The type printed by `writeType` is pretty bad, like IcePy._t_string.

It would be nice to fix that. I don't think there is a slice2py helper function that outputs the correct type.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1261,213 +1261,44 @@ Slice::Python::CodeVisitor::visitStructStart(const StructPtr& p)
...
-
-            //
-            // The None value is not orderable in Python 3.
-            //
-            _out << nl << "if self." << memberName << " != other." << memberName << ':';
-            _out.inc();
-            _out << nl << "return False";
-            _out.dec();
+            // See writeAssign.
+            _out << " = " << getTypeReference(st) << "()";
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

cpp/src/slice2py/PythonUtil.cpp
pepone: > I don't find it correct.

Because mutable types should not be used as dictionary keys? or do you have a different concern?

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1261,213 +1261,43 @@ Slice::Python::CodeVisitor::visitStructStart(const StructPtr& p)
...
+    _out << nl << "@dataclass";
+    if (Dictionary::isLegalKeyType(p))
+    {
+        _out << "(order=True, unsafe_hash=True)";
+    }
+
+    _out << nl << "class " << name << ":";
     _out.inc();
 
     writeDocstring(DocComment::parseFrom(p, pyLinkFormatter), members);
```
bernardnormier, line 1273: Also need to fix docstrings (not done in this PR).

cpp/src/slice2py/PythonUtil.cpp
pepone: I think the type printed here is intended to pass to the C extension. maybe we should rename the method to `WriteIcePyType`

cpp/src/slice2py/PythonUtil.cpp
bernardnormier: Yes, according to the reference page:
> Having a __hash__() implies that instances of the class are immutable.

cpp/src/slice2py/PythonUtil.cpp
pepone: I added `typeToTypeHintString` in #4191

pepone (approved):

pepone:
I merge this changes in #4213

PR #4185, opened 2025-07-02
Add Ice.Current param type in Python

bernardnormier:


InsertCreativityHere (approved):
Looks good to me!

python/python/Ice/Object.py
```diff
@@ -9,7 +9,7 @@ class Object:
     The base class for servants.
     """
 
-    def ice_isA(self, id, current):
```
pepone, line 12: I think is clear to have `from .Current import Current` and use `Current`.

pepone (approved):

PR #4186, opened 2025-07-02
Fix '@link' Mapping in MATLAB

InsertCreativityHere:
This PR updates MATLAB to property map `@link` for all Slice types.
The generated links work well in the 'Command Window' (via the `help` command).
And also show up well in the browser (via the `doc` command).

I did not check the generated HTML you get with publish, because I couldn't get `publish` working...

Included below are the generated code, and how it renders with `help` and `doc` respectively:

----

----

**This Slice definition:**
```
    /// The Glacier2 specialization of the {@link Ice::Router} interface.
    interface Router extends Ice::Router { ... }
```
**will produce this generated code:**
```
classdef RouterPrx < Ice.RouterPrx
    %ROUTERPRX The Glacier2 specialization of the <a href="matlab:help Ice.RouterPrx -displayBanner">RouterPrx</a> interface.
    % ... more stuff
```
**This is what it looks like in the command window:**
image
**And what it looks in the rendered doc page (with `doc Glacier2.RouterPrx`):**
image

----

----

**And then one other example, here's the Slice definition for `RemoteLogger::log`:**
```
        /// Logs a LogMessage.
        /// @param message The message to log.
        /// @remark {@link log} may be called by {@link LoggerAdmin} before {@link init}.
        void log(LogMessage message);
```
**This is the generated code for it:**
```
            %LOG Logs a LogMessage.
            %
            %   Input Arguments
            %     message - The message to log.
            %       Ice.LogMessage scalar
            %     context - The request context.
            %       unconfigured dictionary (default) | dictionary(string, string) scalar
            %
            %   Remarks
            %     <a href="matlab:help Ice.RemoteLoggerPrx/log -displayBanner">log</a> may be called by <a href="matlab:help Ice.LoggerAdminPrx -displayBanner">LoggerAdminPrx</a> before <a href="matlab:help Ice.RemoteLoggerPrx/init -displayBanner">init</a>.
```
**Here's the output of `help Ice/RemoteLoggerPrx/log`, including what happens when you click the `init` link:**
image
**And here's the rendered doc-page (`doc Ice/RemoteLoggerPrx/log`):**
image

### DIFF
--- a/cpp/src/slice2matlab/Gen.cpp
+++ b/cpp/src/slice2matlab/Gen.cpp
@@ -69,11 +69,6 @@ namespace
             return builtinTable[builtin->kind()];
         }
 
-        if (auto cl = dynamic_pointer_cast<ClassDecl>(type))
-        {
-            return cl->mappedScoped(".");
-        }
-
         if (auto proxy = dynamic_pointer_cast<InterfaceDecl>(type))
         {
             return proxy->mappedScoped(".") + "Prx";
@@ -84,12 +79,9 @@ namespace
             return "dictionary";
         }
 
-        if (auto contained = dynamic_pointer_cast<Contained>(type))
-        {
-            return contained->mappedScoped(".");
-        }
-
-        return "???";
+        auto contained = dynamic_pointer_cast<Contained>(type);
+        assert(contained); // every 'Type' is either 'Builtin', or 'Contained'
+        return contained->mappedScoped(".");
     }
 
     // Type represents a dictionary value type or the type of a sequence element.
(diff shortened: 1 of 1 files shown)


cpp/src/slice2matlab/Gen.cpp
```diff
@@ -69,11 +70,6 @@ namespace
             return builtinTable[builtin->kind()];
         }
 
-        if (auto cl = dynamic_pointer_cast<ClassDecl>(type))
-        {
-            return cl->mappedScoped(".");
-        }
```
InsertCreativityHere, line 75: `ClassDecl` inherits from `Contained`, so it's already being handled by the case at the bottom.
Probably leftover redundancy from when `Class` and `Interface` were split.

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -281,26 +274,54 @@ namespace
...
+        else if (auto builtin = dynamic_pointer_cast<Builtin>(target))
+        {
+            displayText = linkText = typeToString(builtin);
+        }
+        else // we couldn't resolve the link target and make a best-effort attempt to map the link.
         {
-            return rawLink;
+            // All we can do is replace any doxygen separators with the MATLAB separator ('.').
+            // Note that '.' can always be used in place of '/', it's just convention to use '/' in certain cases.
+            std::regex separatorRegex{"::|#"};
```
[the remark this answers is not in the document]
InsertCreativityHere: The idea was valid, but the suggested change was bogus.

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -281,26 +274,54 @@ namespace
...
+        string linkText;    // The fully-scoped name of a MATLAB element, which the link resolves to when clicked.
+
+        if (auto contained = dynamic_pointer_cast<Contained>(target))
         {
-            string result;
-            if (hashPos != 0)
+            if (dynamic_pointer_cast<DataMember>(contained) || dynamic_pointer_cast<Enumerator>(contained))
+            {
+                // Data member & enumerator links should be of the form "<ScopedType>/<propertyName>".
+                auto container = dynamic_pointer_cast<Contained>(contained->container());
```
bernardnormier, line 288: Calling this ContainedPtr `container` is rather confusing.

bernardnormier (approved):

cpp/src/slice2matlab/Gen.cpp
InsertCreativityHere: I'm open to alternative names, but this field does indeed hold the 'container'. Even if the type of the field isn't a `ContainerPtr`. In an ideal world we could do like Rust and have a `Contained + Container`, but such is C++.

PR #4188, opened 2025-07-02
Update Current and InitializationData to use dataclass

bernardnormier:


python/python/Ice/InitializationData.py
```diff
@@ -33,12 +40,11 @@ class InitializationData:
...
-        self.logger = None
-        self.threadStart = None
-        self.threadStop = None
-        self.executor = None
-        self.batchRequestInterceptor = None
-        self.eventLoopAdapter = None
-        self.sliceLoader = None
+    properties: Properties | None = None
+    logger: Logger | None = None
+    threadStart: callable = None
```
pepone: Not sure `callable` is the correct type hint here, see https://docs.python.org/3/library/typing.html#annotating-callable-objects

python/python/Ice/InitializationData.py
```diff
@@ -1,44 +1,55 @@
...
-    threadStop : callable
+    threadStop : Callable[[], None] | None
         A callable that is invoked when an Ice thread is stopped.
-    executor : callable
+    executor : Callable[[Callable[[], None], Connection], None] | None
         A callable that is invoked when Ice needs to execute an activity. The callable
         receives two arguments: a callable and an Ice.Connection object. The executor must
         eventually invoke the callable with no arguments.
-    batchRequestInterceptor : callable
+    batchRequestInterceptor : BatchRequestInterceptor | None
```
bernardnormier, line 33: Currently this batchRequestInterceptor can be either:

```
"batch request interceptor must either be a callable or an object with an 'enqueue' method"
```

I don't think we should use this "or" pattern. Just pick one. In this case, I would get rid of the  ABC and use a Callable.

pepone (approved):

externl (approved):

PR #4191, opened 2025-07-04
Initial support for Python type hints

pepone:
This PR is an initial step in adding support for type-hints to Ice for Python:

- Add type hints to proxy classes, and servant-skeleton classes.
- Update Python tests to use slice2py generated code.

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -19,6 +19,132 @@ namespace
 {
     const char* const tripleQuotes = R"(""")";
 
+    string typeToTypeHintString(const TypePtr& type, bool optional)
+    {
+        assert(type);
+
+        if (optional)
+        {
+            if (isProxyType(type))
+            {
+                // We map optional proxies like regular proxies, as XxxPrx or None.
+                return typeToTypeHintString(type, false);
+            }
+            else
+            {
+                return typeToTypeHintString(type, false) + " | None";
+            }
+        }
+
+        static constexpr string_view builtinTable[] = {
+            "int",
+            "bool",
+            "int",
+            "int",
+            "int",
+            "float",
+            "float",
+            "str",
+            "Ice.Value",
+            "Ice.ObjectPrx | None",
+            "Ice.Value"};
+
+        if (auto builtin = dynamic_pointer_cast<Builtin>(type))
+        {
+            if (builtin->kind() == Builtin::KindObject)
+            {
+                return string{builtinTable[Builtin::KindValue]};
+            }
+            else
+            {
+                return string{builtinTable[builtin->kind()]};
+            }
+        }
+        else if (auto cl = dynamic_pointer_cast<ClassDecl>(type))
+        {
+            return cl->mappedScoped(".");
+        }
+        else if (auto st = dynamic_pointer_cast<Struct>(type))
+        {
+            return st->mappedScoped(".");
+        }
+        else if (auto proxy = dynamic_pointer_cast<InterfaceDecl>(type))
+        {
+            return proxy->mappedScoped(".") + "Prx | None";
+        }
+        else if (auto en = dynamic_pointer_cast<Enum>(type))
+        {
+            return en->mappedScoped(".");
+        }
+        else if (auto seq = dynamic_pointer_cast<Sequence>(type))
+        {
+            return "list[" + typeToTypeHintString(seq->type(), false) + "]";
+        }
+        else if (auto dict = dynamic_pointer_cast<Dictionary>(type))
+        {
+            ostringstream os;
+            os << "dict[" << typeToTypeHintString(dict->keyType(), false) << ", "
+               << typeToTypeHintString(dict->valueType(), false) << "]";
+            return os.str();
+        }
+
+        return "???";
+    }
+
+    string operationReturnTypeHint(const OperationPtr& operation, bool forDispatch, bool async)
+    {
+        assert(operation);
+        string returnTypeHint;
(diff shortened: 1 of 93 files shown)


python/msbuild/ice.proj
```diff
@@ -46,6 +46,15 @@
         <PythonGenerated Include="$(MSBuildThisFileDirectory)\..\python\**\*_ice.py"/>
     </ItemGroup>
 
+    <ItemGroup>
+        <ClientSliceCompile Include="$(MSBuildThisFileDirectory)..\test\**\*.ice" Exclude="**\Server*.ice">
```
bernardnormier, line 50: The exclude is because some tests define client-private or server-private slice files in Client*.ice or Server*.ice files?

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -65,110 +67,94 @@ namespace
                 return string{builtinTable[builtin->kind()]};
             }
         }
-
-        ClassDeclPtr cl = dynamic_pointer_cast<ClassDecl>(type);
-        if (cl)
+        else if (auto cl = dynamic_pointer_cast<ClassDecl>(type))
```
bernardnormier, line 70: (minor) I would move these "misc contained types" (ClassDecl, Struct, Enum) to the last branch of the if. Something like:

```cpp
else
{
    auto contained = dynamic_pointer_cast<Contained>(type);
    assert(contained);
    return contained->mappedScoped(".");
}
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -17,9 +17,16 @@ using namespace IceInternal;
 
 namespace
 {
+    enum OperationMode
```
bernardnormier, line 20: The name is confusing since we already use OperationMode for something else in Ice. We should also use `enum class` in new C++ code.

What about:
```cpp
enum class MethodKind { SyncInvocation, AsyncInvocation, Dispatch };
```

?

bernardnormier (approved):
looks good!

python/msbuild/ice.proj
pepone: Yes, we have a few test that rely on having different definitions. Previously each test configure what Slice files to load with loadSlice in the Client and Server entry points.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -65,110 +67,94 @@ namespace
...
-            }
-            else if (op->returnType())
-            {
-                os << typeToDocstring(op->returnType(), op->returnIsOptional());
-            }
+            returnTypeHint = "None";
         }
 
-        return os.str();
+        if (mode == OpAsync)
```
externl, line 141: Maybe use a switch statement here?

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -552,7 +538,7 @@ Slice::Python::CodeVisitor::writeOperations(const InterfaceDefPtr& p)
             capName[0] = static_cast<char>(toupper(static_cast<unsigned char>(capName[0])));
             _out << sp;
             _out << nl << "@staticmethod";
-            _out << nl << "def " << capName << "MarshaledResult(result, current):";
+            _out << nl << "def " << capName << "MarshaledResult(result, current: Ice.Current):";
```
externl, line 541: Can we have a type hint for `result` as well?

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -892,29 +887,41 @@ Slice::Python::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
...
+    _out << nl << "context: dict[str, str] | None = None";
+    _out.dec();
+    _out << nl << ") -> Awaitable[" << prxTypeHint << "]:";
     _out.inc();
     _out << nl << "return Ice.checkedCastAsync(" << prxAbs << ", proxy, facet, context)";
     _out.dec();
 
     _out << sp << nl << "@staticmethod";
-    _out << nl << "def uncheckedCast(proxy, facet=None):";
+    _out << nl << "def uncheckedCast(proxy: Ice.ObjectPrx | None, facet: str|None=None) -> " << prxTypeHint << ":";
```
externl, line 917:
```suggestion
    _out << nl << "def uncheckedCast(proxy: Ice.ObjectPrx | None, facet: str|None = None) -> " << prxTypeHint << ":";
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1460,7 +1467,14 @@ Slice::Python::CodeVisitor::visitStructStart(const StructPtr& p)
...
     _out << sp << nl << "def __repr__(self):";
     _out.inc();
-    _out << nl << "return f\"" << getAbsolute(p) << "(" << formatFields(members) << ")\"";
+    if (members.empty())
+    {
+        _out << nl << "return \"" << getAbsolute(p) << "()\"";
+    }
+    else
+    {
+        _out << nl << "return f\"" << getAbsolute(p) << "(" << formatFields(members) << ")\"";
```
externl, line 1476: We do this in a few places, might be worth a helper function

bernardnormier (approved):

InsertCreativityHere (approved):
Looks good to me, although I agree with most of the comments that have already been made.

cpp/src/slice2py/PythonUtil.cpp
pepone: Fixing in a follow-up PR.

PR #4193, opened 2025-07-07
Add type hints to Communicator and pyright fixes

externl:
This PR adds type hints to Communicator.py and fixes various issues with our python configuration.

Moves `pyproject.toml` to the root folder. This is where tools expect (especially vscode) it to be found. We may need to update the build to account for this.

### DIFF
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -166,8 +166,6 @@ Slice::JavaVisitor::writeMarshalUnmarshalCode(
 
     string typeS = typeToString(type, TypeModeIn, package, metadata);
 
-    assert(!marshal || mode != OptionalMember); // Only support OptionalMember for un-marshaling
-
     const BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(type);
     if (builtin)
     {
@@ -232,11 +230,6 @@ Slice::JavaVisitor::writeMarshalUnmarshalCode(
                     {
                         out << nl << param << " = " << stream << ".readProxy(" << tag << ");";
                     }
-                    else if (mode == OptionalMember)
-                    {
-                        out << nl << stream << ".skip(4);";
-                        out << nl << param << " = " << stream << ".readProxy();";
-                    }
                     else
                     {
                         out << nl << param << " = " << stream << ".readProxy();";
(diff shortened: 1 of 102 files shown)


python/python/Ice/Communicator.py
```diff
@@ -156,7 +172,7 @@ def stringToProxy(self, str):
         """
         return self._impl.stringToProxy(str)
 
-    def proxyToString(self, proxy):
+    def proxyToString(self, proxy: ObjectPrx | None) -> str:
```
InsertCreativityHere, line 175: Your type hint looks correct, but the comment for this function is mis-typed.
proxy can be `Ice.ObjectPrx | None`, but the comment doesn't mention `None`.

python/python/Ice/Communicator.py
```diff
@@ -130,7 +146,7 @@ def isShutdown(self):
         """
         return self._impl.isShutdown()
 
-    def stringToProxy(self, str):
+    def stringToProxy(self, str: str) -> ObjectPrx | None:
```
InsertCreativityHere, line 149: This one also doesn't mention `None` in the comment.
So maybe this is a bigger problem to be fixed in a separate PR.

InsertCreativityHere (approved):
Looks good to me!

python/python/Ice/Communicator.py
```diff
@@ -193,7 +209,7 @@ def propertyToProxy(self, property):
         """
         return self._impl.propertyToProxy(property)
 
-    def proxyToProperty(self, proxy, property):
+    def proxyToProperty(self, proxy: ObjectPrx, property: str) -> dict:
```
pepone: Would be good to specify the dict type, is this `dict[str, str]`?

python/python/Ice/Communicator.py
```diff
@@ -106,7 +122,7 @@ def waitForShutdown(self):
         while not self._impl.waitForShutdown(500):
             pass
 
-    def shutdownCompleted(self):
+    def shutdownCompleted(self) -> Future:
```
pepone: `Awaitable[None]`?

pepone (approved):

python/python/Ice/Communicator.py
externl: Fixed.

python/python/Ice/Communicator.py
externl: Fixed (I think).

PR #4196, opened 2025-07-08
Add type hinting to Future

externl:


python/python/Ice/EventLoopAdapter.py
```diff
@@ -41,7 +41,7 @@ def wrapFuture(self, future: Future) -> Awaitable:
 
         Returns
         -------
-        Future-like
-            A Future-like object that can be awaited in the application event loop.
+        Awaitable
```
externl, line 44: Future-like does not imply awaitable.

python/python/Ice/Future.py
```diff
@@ -1,24 +1,193 @@
...
+        Set the result of the future and mark it as completed.
+
+        This method stores the provided `result` and transitions the future's state to "done". Any registered
+        callbacks are executed after the state update.
+
+        If the future is not in a running state, this method has no effect.
+
+        Parameters
+        ----------
+        result : _T
```
pepone, line 86: is `_T` the convention for generic parameters?

python/python/Ice/Future.py
```diff
@@ -1,24 +1,193 @@
...
+        If the future is already complete, `fn` is called immediately from the calling thread.
+
+        Parameters
+        ----------
+        fn : callable
+            The function to execute upon completion. Takes the future as its argument.
+        *args : Any
+            Additional positional arguments (for compatibility with different Future implementations).
+        **kwargs : Any
+            Additional keyword arguments (for compatibility with different Future implementations).
```
pepone: Are `*args` and `**kwargs` parameters for `add_done_callback`?

python/python/Ice/Future.py
```diff
@@ -1,24 +1,193 @@
...
+# Type variable for the result type of the Future
+_T = TypeVar("_T")
+
 
 #
 # This class defines an __await__ method so that coroutines can call 'await <future>'.
 #
-class FutureBase:
-    def __await__(self):
+class FutureBase(Generic[_T], Awaitable[_T], ABC):
```
externl, line 21: I plan to remove `FutureBase` in a followup PR (if possible).

python/python/Ice/Future.py
```diff
@@ -346,4 +513,8 @@ def _warn(self, msg):
     StateDone = "done"
 
 
-__all__ = ["Future", "wrap_future", "FutureBase"]
+# FutureLike is a type alias for an asyncio.Future, concurrent.futures.Future, or any awaitable object.
+FutureLike: TypeAlias = asyncio.Future[_T] | concurrent.futures.Future[_T] | FutureBase[_T]
```
externl, line 523: We can update this in the future to include other Future types, or a protocol (FutureProtocol) that can match some hopefully general future types.

python/python/Ice/Future.py
```diff
@@ -55,12 +220,12 @@ def wrap_future(future, *, loop=None):
...
     assert isinstance(future, FutureBase), "Ice.Future is expected, got {!r}".format(future)
 
-    def forwardCompletion(sourceFuture, targetFuture):
+    def forwardCompletion(sourceFuture: FutureBase | asyncio.Future, targetFuture: FutureBase | asyncio.Future):
         if not targetFuture.done():
             if sourceFuture.cancelled():
                 targetFuture.cancel()
-            elif sourceFuture.exception():
-                targetFuture.set_exception(sourceFuture.exception())
+            elif (sourceException := sourceFuture.exception()) is not None:
```
pepone, line 233: Can we use 2 overloads here?

```
@overload
def forwardCompletion(sourceFuture: FutureBase, targetFuture: asyncio.Future): ...

@overload
def forwardCompletion(sourceFuture: asyncio.Future, targetFuture: FutureBase): ...

def forwardCompletion(sourceFuture: FutureBase | asyncio.Future, targetFuture: FutureBase | asyncio.Future):
```

See https://docs.python.org/3/library/typing.html#overload

python/python/Ice/asyncio/__init__.py
```diff
@@ -1,5 +1,5 @@
 # Copyright (c) ZeroC, Inc.
 
-# ruff: noqa: F403
+from .EventLoopAdapter import EventLoopAdapter
 
-from .EventLoopAdapter import *
```
externl, line 5: import * is considered not great. We should really try to avoid it.

python/python/Ice/Future.py
```diff
@@ -170,15 +335,15 @@ def add_done_callback(self, fn):
...
+            The function to execute upon completion. Takes the future as its argument.
         """
         with self._condition:
             if self._state == Future.StateRunning:
                 self._doneCallbacks.append(fn)
                 return
         fn(self)
 
-    def result(self, timeout=None):
+    def result(self, timeout: int | float | None = None) -> _T:
```
pepone, line 352: I don't understand why these are required here.

python/python/Ice/asyncio/EventLoopAdapter.py
```diff
@@ -12,11 +14,12 @@ class EventLoopAdapter(Ice.EventLoopAdapter):
     the communicator.
     """
 
-    def __init__(self, eventLoop):
+    def __init__(self, eventLoop: asyncio.AbstractEventLoop):
         self._eventLoop = eventLoop
 
-    def runCoroutine(self, coroutine):
+    def runCoroutine(self, coroutine: Coroutine) -> FutureLike:
```
pepone, line 20: Will be better if we limit FutureLike, to what is required here. `add_done_callback` and `result`. We can refine it in a follow-up PR. I think this is the only method that would need to return a `FutureLike` object.

python/python/Ice/Future.py
externl: It's what asyncio does.

python/python/Ice/Future.py
externl: No, I was trying stuff earlier. I caught that earlier and removed.

python/python/Ice/Future.py
externl: The issue with this change is that `sourceFuture.exception()` the first time does not guarentee the next call will not be None. 

Overload won't help it.

python/python/Ice/Future.py
externl: why what are required?

python/python/Ice/Future.py
pepone: I don't think we need to have such a broad Future definition. If you look at `Dispatch.py` the only place we use it, the requirements for the interface are much more narrow. Just need the `add done_callback` and `result` methods.

python/python/Ice/asyncio/EventLoopAdapter.py
externl: Sure, I spent some time trying to find a common protocol but didn't get anything that worked. It's also not obvious to me that it would work for other random "Future like" types. We can try later though.

python/python/Ice/Future.py
pepone: This end up in the wrong line, I think you updated the code while I was commenting.

python/python/Ice/Future.py
pepone: I was commenting about the type declarations but GitHub move the comment down.

Regarding `exception()` the future should be immutable once settle, isn't this part of the contract?

externl: I don't think so. There's no way for the linter to know this.

pepone (approved):

python/python/Ice/Future.py
```diff
@@ -294,7 +467,7 @@ def set_exception(self, ex):
 
         Parameters
         ----------
-        ex : Exception
```
InsertCreativityHere, line 297: This doesn't match the function specification.

python/python/Ice/Future.py
```diff
@@ -309,12 +482,12 @@ def set_exception(self, ex):
         self._callCallbacks(callbacks)
 
     @staticmethod
-    def completed(result):
+    def completed(result: _T) -> "Future[_T]":
         f = Future()
         f.set_result(result)
         return f
 
-    def _wait(self, timeout, testFn=None):
```
InsertCreativityHere, line 317: Good catch.

InsertCreativityHere (approved):
Looks good to me!

PR #4198, opened 2025-07-08
'slice2cs' and 'slice2java' No Longer Try to Link to equences or Dictionaries

InsertCreativityHere:
`slice2cs` and `slice2java` do not generate actual types or type-defs for sequences or dictionaries.
Instead we just generate the mapped language type, and a `Helper` type.

### DIFF
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -105,52 +105,55 @@ namespace
     }
 
     /// Returns a C# formatted link to the provided Slice identifier.
-    // TODO this fails for sequences and dictionaries.
     string csLinkFormatter(const string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target)
     {
         ostringstream result;
         result << "<see ";
 
-        if (target)
+        if (auto builtinTarget = dynamic_pointer_cast<Builtin>(target))
         {
-            if (auto builtinTarget = dynamic_pointer_cast<Builtin>(target))
+            string typeS = typeToString(builtinTarget, "");
+            if (builtinTarget->kind() == Builtin::KindObjectProxy || builtinTarget->kind() == Builtin::KindValue)
             {
-                string typeS = typeToString(builtinTarget, "");
-                if (builtinTarget->kind() == Builtin::KindObjectProxy || builtinTarget->kind() == Builtin::KindValue)
-                {
-                    // Remove trailing '?':
-                    typeS.pop_back();
-                    result << "cref=\"" << typeS << "\"";
-                }
-                else
-                {
-                    // All other builtin types correspond to C# language keywords.
-                    result << "langword=\"" << typeS << "\"";
-                }
+                // Remove trailing '?':
+                typeS.pop_back();
+                result << "cref=\"" << typeS << "\"";
             }
             else
             {
-                string sourceScope = source->mappedScope(".");
-                sourceScope.pop_back(); // Remove the trailing '.' scope separator.
-
-                result << "cref=\"";
-                if (auto operationTarget = dynamic_pointer_cast<Operation>(target))
-                {
-                    // link to the method on the proxy interface
-                    result << getUnqualified(operationTarget->interface(), sourceScope) << "Prx."
-                           << operationTarget->mappedName();
-                }
-                else if (auto interfaceTarget = dynamic_pointer_cast<InterfaceDecl>(target))
-                {
-                    // link to the proxy interface
-                    result << getUnqualified(interfaceTarget, sourceScope) << "Prx";
-                }
-                else
-                {
-                    result << getUnqualified(dynamic_pointer_cast<Contained>(target), sourceScope);
-                }
-                result << "\"";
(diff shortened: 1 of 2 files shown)


cpp/src/slice2cs/Gen.cpp
```diff
@@ -105,52 +105,55 @@ namespace
...
-                {
-                    result << getUnqualified(dynamic_pointer_cast<Contained>(target), sourceScope);
-                }
-                result << "\"";
+            if (dynamic_pointer_cast<Sequence>(contained) || dynamic_pointer_cast<Dictionary>(contained))
+            {
+                // slice2cs doesn't generate type-defs for sequences or dictionaries, so there's nothing to link to.
+                // So instead of generating a link, we just output the sequence/dictionary name in code formatting.
+                return "<c>" + getUnqualified(contained, sourceScope) + "</c>";
+            }
```
InsertCreativityHere: This is the only code that was added to this function.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -105,52 +105,55 @@ namespace
     }
 
     /// Returns a C# formatted link to the provided Slice identifier.
-    // TODO this fails for sequences and dictionaries.
     string csLinkFormatter(const string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target)
     {
         ostringstream result;
         result << "<see ";
 
-        if (target)
```
InsertCreativityHere, line 114: I also flattened this if/else tree a little bit.

cpp/src/slice2java/Gen.cpp
```diff
@@ -91,43 +91,42 @@ namespace
...
+            result << getUnqualified(parent, sourceScope) << "#" << fieldTarget->mappedName();
+        }
+        else if (auto interfaceTarget = dynamic_pointer_cast<InterfaceDecl>(target))
+        {
+            // Link to the proxy interface.
+            result << getUnqualified(interfaceTarget, sourceScope) << "Prx";
+        }
+        else if (auto contained = dynamic_pointer_cast<Contained>(target))
+        {
+            if (dynamic_pointer_cast<Sequence>(contained) || dynamic_pointer_cast<Dictionary>(contained))
```
InsertCreativityHere, line 122: This if block is the only new code added to this function.

cpp/src/slice2java/Gen.cpp
```diff
@@ -91,43 +91,42 @@ namespace
     /// Returns a javadoc formatted link to the provided Slice identifier.
     string javaLinkFormatter(const string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target)
     {
+        string sourceScope = getPackage(source);
+
         ostringstream result;
         result << "{@link ";
 
-        if (target)
```
InsertCreativityHere, line 97: I also flattened out this `if` tree.
Unfortunately, the diff is struggling with it.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -105,52 +105,55 @@ namespace
...
-                }
-                else
-                {
-                    result << getUnqualified(dynamic_pointer_cast<Contained>(target), sourceScope);
-                }
-                result << "\"";
+            if (dynamic_pointer_cast<Sequence>(contained) || dynamic_pointer_cast<Dictionary>(contained))
+            {
+                // slice2cs doesn't generate type-defs for sequences or dictionaries, so there's nothing to link to.
+                // So instead of generating a link, we just output the sequence/dictionary name in code formatting.
```
pepone:
```suggestion
                // slice2cs doesn't generate C# types for sequences or dictionaries, so there's nothing to link to.
                // Instead, we just output the sequence or dictionary name in code formatting.
```

Replaced type-def here, because it is not a C# thing. Removed the second `So` to avoid so...So`

cpp/src/slice2java/Gen.cpp
```diff
@@ -91,43 +91,42 @@ namespace
...
+            // Link to the proxy interface.
+            result << getUnqualified(interfaceTarget, sourceScope) << "Prx";
+        }
+        else if (auto contained = dynamic_pointer_cast<Contained>(target))
+        {
+            if (dynamic_pointer_cast<Sequence>(contained) || dynamic_pointer_cast<Dictionary>(contained))
             {
-                result << typeToObjectString(builtinTarget, TypeModeIn);
+                // slice2java doesn't generate type-defs for sequences or dictionaries, so there's nothing to link to.
+                // So instead of generating a link, we just output the sequence/dictionary name in code formatting.
```
pepone:
```suggestion
                // slice2java doesn't generate Java types for sequences or dictionaries, so there's nothing to link to.
                // Instead, we just output the sequence or dictionary name in code formatting.
```

same as above.

pepone (approved):

cpp/src/slice2cs/Gen.cpp
InsertCreativityHere: Good suggestion, your version is definitely more clear than mine!

externl (approved):

PR #4200, opened 2025-07-09
Sort python imports

externl:
This PR sorts Python imports with Ruff.


python/python/Ice/__init__.py
```diff
@@ -1,6 +1,6 @@
 # Copyright (c) ZeroC, Inc.
 
-# ruff: noqa: F401, F821, E402, F403, F405
+# ruff: noqa: F401, F821, E402, F403, F405, I001
```
pepone: Can we split this list and have comments for the things we disable?

externl: Will do, would be good to actually fix most of these at some point.

PR #4202, opened 2025-07-09
Move `validateMetadata` and `linkFormatter` Functions

InsertCreativityHere:
This PR doesn't change any code, it only moves existing logic.

I'm doing this for 2 reasons:
1) Right now, we perform metadata validation in the middle of the code-gen phase. We really should be doing this validation _before_ we start generating any code. Especially since it can produce errors.

2) Right now, we pass in a `linkFormatter` every single time we want to generate a comment. It would be better if we could use injection, and provide this up front when we call the `createUnit` entry function, instead of passing it over and over.

----

So, I moved the `linkFormatters`. They are defined in the proper `Slice::XXX` namespace, instead of an anonymous namespace. In my next PR, this will let me pass them into `createUnit`. I also made the `validateMetadata` functions a free function in `XXXUtil`, instead of being a static function on our various `Gen` classes.

Some of the compilers already had their metadata validation in the right place. So I only had to update some of them.


cpp/src/ice2slice/Gen.cpp
```diff
@@ -219,25 +219,6 @@ namespace
         }
     }
```
InsertCreativityHere, line 221: This function was defined in an anonymous namespace, now it's defined in the `Slice` namespace.
The code wasn't changed, I just copy-pasted this function outside of the anonymous namespace, and added the necessary `Slice::` qualification to it.

cpp/src/ice2slice/Gen.h
```diff
@@ -75,6 +75,8 @@ namespace Slice
             std::map<std::string, std::unique_ptr<IceInternal::Output>> _outputs;
         };
     };
+
```
InsertCreativityHere, line 78: It also has an entry in the header file.
This will let me next PR reference it from `Main.cpp` where we call `createUnit`.

cpp/src/slice2cpp/CPlusPlusUtil.h
```diff
@@ -69,6 +69,11 @@ namespace Slice
 
     std::string findMetadata(const MetadataList&, TypeContext = TypeContext::None);
     bool inWstringModule(const SequencePtr&);
+
+    /// Returns a doxygen formatted link to the provided Slice identifier.
+    std::string cppLinkFormatter(const std::string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target);
+
+    void validateCppMetadata(const UnitPtr&);
```
InsertCreativityHere, line 77: `validateCppMetadata` is now a free function.

Originally it was named `validateMetadata`, but I added the `Cpp` part in the middle.
Half of the compilers had this name, half of them didn't. I think including the language name is cleaner,
since `validateMetadata` is also the name of the 'main' validation function in `libSlice`.

cpp/src/slice2cpp/Gen.h
```diff
@@ -201,8 +201,6 @@ namespace Slice
             IceInternal::Output& H;
             bool _firstElement{true};
         };
-
-        static void validateMetadata(const UnitPtr&);
```
InsertCreativityHere, line 205: It used to be a static function on the code-gen.
There's absolutely no reason to tie validation to the code-gen though.

PR #4203, opened 2025-07-09
Inject 'linkFormatter' Functions Up-Front via `createUnit`

InsertCreativityHere:
Currently, whenever one of the compiler wants to generate a doc-comment, it has to call `parseFrom` and provide a linkFormatter.
Having to repeatedly provide this `linkFormatter` is unnecessary though, since it never changes.

This PR changes things so now we inject a `linkFormatter` into `libSlice` once, when we first call `createUnit`.
And the doc-comment parsing code simply retrieves the injected `linkFormatter` when necessary.

This doesn't make a huge difference right now, but will be very important for supporting `@p` tags and code-snippets, which will need their own `xxxFormatter` functions as well.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -5203,7 +5203,10 @@ Slice::Unit::getTopLevelModules(const string& file) const
     }
 }
 
-Slice::Unit::Unit(string languageName, optional<DocLinkFormatter> linkFormatter, bool all) : _languageName(std::move(languageName)), _linkFormatter(std::move(linkFormatter)), _all(all)
+Slice::Unit::Unit(string languageName, optional<DocLinkFormatter> linkFormatter, bool all)
+    : _languageName(std::move(languageName)),
+      _linkFormatter(std::move(linkFormatter)),
+      _all(all)
 {
     if (!languageName.empty())
     {
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -440,8 +440,13 @@ namespace
 }
 
 optional<DocComment>
-Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatter, bool escapeXml)
+Slice::DocComment::parseFrom(const ContainedPtr& p, bool escapeXml)
 {
+    const optional<DocLinkFormatter>& linkFormatter = p->unit()->linkFormatter();
+    // Some compilers don't generate doc-comments, and so don't provide a link formatter.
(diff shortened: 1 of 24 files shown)


ruby/src/IceRuby/Slice.cpp
```diff
@@ -114,7 +114,7 @@ IceRuby_loadSlice(int argc, VALUE* argv, VALUE /*self*/)
                 throw RubyException(rb_eArgError, "Slice preprocessing failed for `%s'", cmd.c_str());
             }
 
-            UnitPtr u = Unit::createUnit("ruby", all);
+            UnitPtr u = Unit::createUnit("ruby", nullopt, all);
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes, but this is actually desired behavior.
Ruby should never be calling `parseFrom`, and if it accidentally does, we want to hit that assert to find out.

cpp/src/Slice/Parser.cpp
```diff
@@ -440,8 +440,13 @@ namespace
 }
 
 optional<DocComment>
-Slice::DocComment::parseFrom(const ContainedPtr& p, DocLinkFormatter linkFormatter, bool escapeXml)
+Slice::DocComment::parseFrom(const ContainedPtr& p, bool escapeXml)
 {
+    const optional<DocLinkFormatter>& linkFormatter = p->unit()->linkFormatter();
```
pepone, line 445: The previous design was clear regarding to dependencies. Now there is a hidden dependency in a formatter that is optional, and it comes from Unit that doesn't really use it.

Is the next step to move the parsing of the comments to Unit? If this is just about removing the parameter I prefer not to, as it makes the dependencies less obvious.

InsertCreativityHere: My end goal is to have all the comment parsing done up-front within `libSlice`.
Instead of the current logic where the compilers parse comments every single time they need one.
And `Unit` is simply the entry point into `libSlice` for the compilers.

So instead of each Slice element holding `string comment`,
they could instead hold a pre-parsed `optional<DocComment> comment`.

I can re-open this PR with the full change if you'd rather review when everything is done.

pepone: > My end goal is to have all the comment parsing done up-front within libSlice.

That sounds fine.

I assume later we can pass the link formatter from unit to parse rather that have an assert here that should never be hit.

InsertCreativityHere: Yes, the next step is to remove all the `parseFrom`s from the languages, and do the parsing here in `libSlice` now.
Already half-way done with moving everything over!

externl (approved):

PR #4206, opened 2025-07-10
Repalce Ice.EnumBase with enum.Enum

externl:
This PR replaces our custom `Ice.EnumBase` with the builtin `enum.Enum`

### DIFF
--- a/python/test/Ice/exceptions/TestI.py
+++ b/python/test/Ice/exceptions/TestI.py
@@ -1,6 +1,5 @@
 # Copyright (c) ZeroC, Inc.
 
-import sys
 import Test
 
 import Ice


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1572,66 +1573,32 @@ Slice::Python::CodeVisitor::visitEnum(const EnumPtr& p)
...
 
     //
     // Emit the type information.
     //
     _out << sp << nl << getMetaTypeReference(p) << " = IcePy.defineEnum(\"" << scoped << "\", " << name << ", ";
     writeMetadata(p->getMetadata());
-    _out << ", " << name << "._enumerators)";
+    _out << ", ";
+    _out.spar("{");
+    for (const auto& enumerator : enumerators)
```
externl, line 1596: one thing we could do is to remove this and access `__members__` in EnumInfo to read the enumerators.

python/python/Ice/LocalExceptions.py
```diff
@@ -20,16 +20,21 @@ class DispatchException(LocalException):
     wire".
     """
 
-    def __init__(self, replyStatus, msg=""):
+    def __init__(self, replyStatus: int, msg: str = ""):
```
pepone:
```suggestion
    def __init__(self, replyStatus: int | None, msg: str = ""):
```

python/test/Ice/exceptions/TestI.py
```diff
@@ -1,5 +1,6 @@
 # Copyright (c) ZeroC, Inc.
 
+import sys
```
pepone: is this required?

externl: No, just leftovers.

PR #4209, opened 2025-07-10
More Python type hinting

externl:


python/python/Ice/InvocationFuture.py
```diff
@@ -77,7 +79,7 @@ def is_sent_synchronously(self):
         with self._condition:
             return self._sentSynchronously
 
-    def add_sent_callback(self, fn):
+    def add_sent_callback(self, fn: Callable):
```
pepone: Should `void` methods use `-> None` ?

externl: Yes, I think we would prefer them.

PR #4211, opened 2025-07-14
Updates to Python Future and Dispatch

externl:
- Remove FutureBase
- Simplify FutureLike
- Add type hints to dispatch

### DIFF
--- a/python/python/Ice/Dispatch.py
+++ b/python/python/Ice/Dispatch.py
@@ -56,7 +56,7 @@ def dispatch(cb: IcePy.DispatchCallback, method: Callable, args: list[Any]):
 
     # If the result is a future, attach a done callback to handle dispatch completion.
     if is_future(result):
-        # Use a more precise type annotation while maintaining runtime behavior
+        # Use a more precise type annotation while maintaining runtime behavior.
         result = cast(FutureLike, result)
 
         def handle_future_result(future: FutureLike):


python/python/Ice/Dispatch.py
```diff
@@ -50,8 +56,10 @@ def dispatch(cb, method, args):
 
     # If the result is a future, attach a done callback to handle dispatch completion.
     if is_future(result):
+        # Use a more precise type annotation while maintaining runtime behavior
+        result = cast(FutureLike, result)
```
externl: Would be nice to try and merge these in the future.

python/python/Ice/Future.py
```diff
@@ -519,8 +288,92 @@ def _warn(self, msg: str):
...
+
+    Returns
+    -------
+    asyncio.Future
+        A future that mirrors the state of the input Ice.Future.
+
+    Raises
+    ------
+    AssertionError
+        If `future` is not an instance of Ice.Future.
```
pepone:
```suggestion
        If `future` is not an instance of Ice.Future or asyncio.Future.
```

pepone (approved):

python/python/Ice/Dispatch.py
```diff
@@ -50,8 +56,10 @@ def dispatch(cb, method, args):
 
     # If the result is a future, attach a done callback to handle dispatch completion.
     if is_future(result):
+        # Use a more precise type annotation while maintaining runtime behavior
```
InsertCreativityHere: Definitely a nitpick, but:
```suggestion
        # Use a more precise type annotation while maintaining runtime behavior.
```

InsertCreativityHere (approved):
Looks good to me!

Issue #4212, opened 2025-07-15
Rework depend files generation


pepone:
We currently have printMakefileDependencies that generates XML, JSON, and Makefile depend files.

The implementation parses the output of calling the preprocessor with `-M` and then rewrites it as a XML, JSON, or Makefile style with replaced extensions.

I think it would be clear if we have `set<string> getDepedencies(string file, StringList includePaths)`. And then separate functions to emit the dependencies in the desired formats.

With the refactor of slice2py for generating a Python module for each Slice definition, the `printMakeFileDependencies` model doesn't work because it implies there is a 1 to 1 relationship between Slice and the generated code.

```
jose@studio python % ../cpp/bin/slice2cpp --depend -Islice ../slice/Ice/Locator.ice 
Locator.h: \
 ../slice/Ice/Locator.ice \
 /Users/jose/Documents/3.8/ice/slice/Ice/Identity.ice
jose@studio python % ../cpp/bin/slice2rb --depend -Islice ../slice/Ice/Locator.ice
Locator.rb: \
 ../slice/Ice/Locator.ice \
 /Users/jose/Documents/3.8/ice/slice/Ice/Identity.ice
jose@studio python % ../cpp/bin/slice2java --depend -Islice ../slice/Ice/Locator.ice
```

For example The Slice for Python generates:

```
jose@studio python % ../cpp/bin/slice2py --list-generated modules -Islice ../slice/Ice/Locator.ice
Ice/AdapterNotFoundException.py
Ice/Locator.py
Ice/LocatorF.py
Ice/LocatorFinder.py
Ice/LocatorFinderF.py
Ice/ObjectNotFoundException.py
```

PR #4213, opened 2025-07-15
Refactor Python code generation

pepone:
This PR refactors the Python code generation to create Python packages and Python modules statically, instead of relying on the code dynamically add them to the sys.modules.

Each Slice module is still mapped to a package with the same name. 

- The `python:package` metadata has been removed, `python:identifier` can be used to remap a module to a different package.
- The `python:pkgdir` metadata has been removed, the generated files are always placed in a directory matching the mapped package. The `--output-dir` slice2py argument can still be used to change the base directory for generated code.
- The `--prefix` slice2py option was removed, `python:identifier` is the only supported option for remap a definition.
- The `--package` and `--no-package` slice2py options have been replaced by `--build=all|modules|index`.
- There is a new `--list-generated=all|modules|index` to list what files will be generated.
- The generation of package index files (`__init__.py`) has been reworked, `__init__.py` files are no longer edited in place, you have to pass all the definitions that provide symbols to the package.
- The generated files now import only the required definitions
- The generated code contain separated runtime and typing imports, for runtime we import just what we need. For type hints we import the required packages as this provides better output in the generated html.
- The `loadSlice` functionality has bee temporarily disabled, I will bring it back in a follow up PR.

## Generated code structure

For each Slice definition we create a Python module with the same name, for example the generated code for `class Foo in module Bar` goes to `Bar/Foo.py`. This Python file contains all definitions generated from the given Slice definition, typically the type and the meta-type. Buf for interfaces there is also the ProxyType, ServantType, and the Proxy meta-type.

This PR also includes the changes from  #4183


cpp/src/Slice/Preprocessor.cpp
```diff
@@ -595,13 +590,9 @@ Slice::Preprocessor::printMakefileDependencies(
...
-            {
-                result = pyPrefix + result;
-            }
             if ((pos = result.find(suffix)) != string::npos)
             {
-                result.replace(pos, suffix.size() - 1, "_ice.py");
+                result.replace(pos, suffix.size() - 1, ".py");
             }
             break;
         }
```
pepone, line 598: We should improve this in a follow up PR, see #4212

cpp/src/slice2py/Python.cpp
```diff
@@ -42,388 +43,67 @@ namespace
...
-        if (pos != string::npos)
-        {
-            string parent = abs.substr(0, pos);
-            _modules.remove(parent);
-        }
-    }
+        // Create a new output file for this package.
+        string fileName = packageName;
+        replace(fileName.begin(), fileName.end(), '.', '/');
+        fileName += "/__init__.py";
```
externl, line 51: Should we be using `IceInternal::separator`? Or does `/` work fine?

cpp/src/slice2py/PythonUtil.h
```diff
@@ -6,43 +6,78 @@
...
+#include <string>
+
 namespace Slice::Python
 {
-    /// Get the package directory from metadata (if any).
-    std::string getPackageDirectory(const std::string&, const Slice::UnitPtr&);
+    /// Creates the package directory structure for a given Python module name.
+    ///
+    /// For example, if the module name is "Foo.Bar.Baz", this function creates
+    /// the directories "Foo/Bar" under the specified output path.
```
externl, line 18: Should this be `Foo/Bar/Baz`?

cpp/src/slice2py/PythonUtil.h
```diff
@@ -6,43 +6,78 @@
...
-    std::string getTypeReference(const Slice::ContainedPtr& p);
-
-    /// Get the fully-qualified name of the given definition, including any package defined via metadata,
-    /// but "_M_" is prepended to the first name segment, and "_t_" is prepended to the final name segment,
-    /// indicating that this is a an explicit reference to a type.
-    std::string getMetaTypeReference(const Slice::ContainedPtr& p);
-
-    /// Emit a comment header.
-    void printHeader(IceInternal::Output&);
+    /// @p unit is the Slice unit to generate code for.
```
externl: Shouldn't these be `@param`?

cpp/src/slice2py/PythonUtil.h
```diff
@@ -6,43 +6,78 @@
...
+    /// @p outputDir The base-directory to write the generated Python files to.
+    void generate(const Slice::UnitPtr& unit, const std::string& outputDir);
 
     int compile(const std::vector<std::string>&);
 
     /// Returns a DocString formatted link to the provided Slice identifier.
     std::string
     pyLinkFormatter(const std::string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target);
 
+    /// Validates the Slice python metadata for the given unit.
```
externl:
```suggestion
    /// Validates the Slice Python metadata for the given unit.
```

cpp/src/slice2py/PythonUtil.h
```diff
@@ -6,43 +6,78 @@
...
     void validatePythonMetadata(const UnitPtr&);
+
+    // Collects Python definitions for each generated Python package.
+    // Each package corresponds to a Slice module and includes an `__init__.py` file
+    // that re-exports selected definitions from the package's internal modules.
+    class PackageVisitor final : public ParserVisitor
+    {
+    public:
+        bool visitModuleStart(const ModulePtr&);
```
externl:
```suggestion
        bool visitModuleStart(const ModulePtr&) final;
```

python/python/Ice/LocalExceptions.py
```diff
@@ -57,16 +55,14 @@ class RequestFailedException(DispatchException):
     The base exception for the 3 NotExist exceptions.
     """
 
-    def __init__(
-        self, replyStatus: int, id: Ice.Identity | None = None, facet: str = "", operation: str = "", msg: str = ""
-    ):
+    def __init__(self, replyStatus, id=None, facet="", operation="", msg=""):
```
externl: The typing got lost here.

python/python/Ice/LocalExceptions.py
```diff
@@ -108,8 +104,8 @@ class ObjectNotExistException(RequestFailedException):
     The dispatch could not find a servant for the identity carried by the request.
     """
 
-    def __init__(self, id: Ice.Identity | None = None, facet: str = "", operation: str = "", msg: str = ""):
-        RequestFailedException.__init__(self, Ice.ReplyStatus.ObjectNotExist.value, id, facet, operation, msg)
+    def __init__(self, id=None, facet="", operation="", msg=""):
```
externl: This whole file actually.

python/python/Ice/Object.py
```diff
@@ -109,7 +64,7 @@ def ice_ids(self, current: Current) -> list[str] | Awaitable[list[str]]:
         list of str
             A list of type IDs.
         """
-        return ["::Ice::Object"]
+        return [self.ice_id(current)]
```
externl: This file also seems to have merge issues.

python/python/Ice/ObjectAdapter.py
```diff
@@ -21,7 +25,7 @@ class ObjectAdapter:
     identities, and proxies.
     """
 
-    def __init__(self, impl: IcePy.ObjectAdapter):
+    def __init__(self, impl):
```
externl: Ditto.

python/python/Ice/ObjectAdapter.py
```diff
@@ -119,7 +123,7 @@ def isDeactivated(self) -> bool:
         bool
             True if `deactivate` has been called on this object adapter; otherwise, False.
         """
-        return self._impl.isDeactivated()
+        self._impl.isDeactivated()
```
externl: The return got lost.

python/python/Ice/Util.py
```diff
@@ -98,7 +101,7 @@ def initialize(args=None, initData=None, configFile=None, eventLoop=None):
     return Communicator(communicator, eventLoopAdapter)
 
 
-def identityToString(identity: Ice.Identity, toStringMode: ToStringMode | None = None) -> str:
+def identityToString(identity: Identity, toStringMode: ToStringMode = None) -> str:
```
externl: Typing was lost in this file as well.

externl (commented):
So far so good. Looks like some typing was lost merging in main.

cpp/src/slice2py/PythonUtil.h
pepone: No. In python the module name `Foo.Bar.Baz` corresponds to `Foo/Bar/Baz.py`.

cpp/src/slice2py/Python.cpp
pepone: Seems to work fine, that being say using `IceInternal::separator` seems the right thing to do. I will update the code.

cpp/src/slice2py/PythonUtil.h
pepone: I think you are right.

python/python/Ice/LocalExceptions.py
pepone: I will bring them back.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1,33 +1,208 @@
 // Copyright (c) ZeroC, Inc.
 
 #include "PythonUtil.h"
+#include "../Ice/FileUtil.h"
```
externl, line 4: I think it would be nice to split this into multiple files (after this is merged).

python/python/Ice/EventLoopAdapter.py
```diff
@@ -1,9 +1,13 @@
...
 
+from __future__ import annotations
+
 from abc import ABC, abstractmethod
 from collections.abc import Awaitable, Coroutine
+from typing import TYPE_CHECKING
 
-from .Future import Future, FutureLike
+if TYPE_CHECKING:
+    import Ice
```
externl, line 10: Why can't we use `from .Future import Future, FutureLike`?

python/python/Ice/ObjectAdapter.py
```diff
@@ -5,11 +5,10 @@
 
 from typing import TYPE_CHECKING, final
 
-import Ice
 import IcePy
 
 if TYPE_CHECKING:
-    from .Communicator import Communicator
+    import Ice
```
externl, line 11: Is this necessary? Why can't we use the `from .Communicator`?

python/python/Ice/ObjectPrx.py
```diff
@@ -4,13 +4,15 @@
...
-from typing import Self
+from typing import TYPE_CHECKING, Self
 
-import Ice
 import IcePy
 
 from .Object import Object
 
+if TYPE_CHECKING:
+    import Ice
```
externl, line 14: Ditto.

python/python/Ice/Util.py
```diff
@@ -1,45 +1,51 @@
...
+
+if TYPE_CHECKING:
+    import Ice
 
 __name__ = "Ice"
 
 
 @overload
-def initialize() -> Communicator: ...
+def initialize() -> Ice.Communicator: ...
```
externl, line 26: We're already importing Communicator, do we need to get from Ice?

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2212,220 +2466,178 @@ Slice::Python::CodeVisitor::writeDocstring(const OperationPtr& op, OperationMode
...
+        Output& out = getModuleOutputFile(moduleName, outputFiles, outputDir);
+        out << sp;
+        out << fragment.code;
+        out << nl;
+    }
 }
 
 string
-Slice::Python::pyLinkFormatter(const string& rawLink, const ContainedPtr&, const SyntaxTreeBasePtr& target)
+Slice::Python::pyLinkFormatter(const string& rawLink, const ContainedPtr& p, const SyntaxTreeBasePtr& target)
```
InsertCreativityHere: Very minor, but in the other formatters we use `source` for clarity.

```suggestion
Slice::Python::pyLinkFormatter(const string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target)
```
Which also nicely lines up with the name for `typeToHintString`.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2446,7 +2658,7 @@ Slice::Python::pyLinkFormatter(const string& rawLink, const ContainedPtr&, const
             }
             else
             {
-                result << "``" << typeToTypeHintString(builtinTarget, false) << "``";
+                result << "``" << typeToTypeHintString(builtinTarget, false, p) << "``";
```
```suggestion
                result << "``" << typeToTypeHintString(builtinTarget, false, source) << "``";
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1999,50 +2257,52 @@ Slice::Python::CodeVisitor::writeDocstring(const optional<DocComment>& comment,
...
-Slice::Python::CodeVisitor::writeDocstring(const OperationPtr& op, OperationMode mode)
+Slice::Python::CodeVisitor::writeDocstring(const OperationPtr& op, MethodKind methodKind, Output& out)
 {
     optional<DocComment> comment = DocComment::parseFrom(op);
     if (!comment)
     {
         return;
     }
 
+    auto p = dynamic_pointer_cast<Contained>(op->container());
```
InsertCreativityHere: Little nicer way to avoid the cast, since we have a concrete `OperationPtr`:
```suggestion
    auto p = op->interface();
```

python/python/Ice/Util.py
pepone: Not required. But I updated this to get consistent names in the generated docs.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -228,237 +398,631 @@ namespace Slice::Python
...
 
         /// Writes the provided @p remarks in its own subheading in the current comment (if @p remarks is non-empty).
-        void writeRemarksDocComment(const StringList& remarks, bool needsNewline);
+        void writeRemarksDocComment(const StringList& remarks, bool needsNewline, Output& out);
+
+        void writeDocstring(const optional<DocComment>&, const string&, Output&);
+        void writeDocstring(const optional<DocComment>&, const DataMemberList&, Output&);
+        void writeDocstring(const optional<DocComment>&, const EnumPtr&, Output&);
+
+        void writeDocstring(const OperationPtr&, MethodKind, Output&);
```
InsertCreativityHere, line 432: I'm pretty sure most (maybe even all) of these methods can be marked `static` now, since the `Output` is passed in as a parameter now.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1935,50 +2187,56 @@ Slice::Python::CodeVisitor::writeDocstring(const optional<DocComment>& comment,
...
         }
-        _out << nl << "Attributes";
-        _out << nl << "----------";
+        out << nl << "Attributes";
+        out << nl << "----------";
         for (const auto& member : members)
         {
-            _out << nl << member->mappedName() << " : " << typeToTypeHintString(member->type(), member->optional());
+            ;
+            out << nl << member->mappedName() << " : "
```
InsertCreativityHere: A semicolon snuck in here:
```suggestion
            out << nl << member->mappedName() << " : "
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1503,411 +1896,270 @@ Slice::Python::CodeVisitor::visitEnum(const EnumPtr& p)
...
-            _out << getTypeReference(lte);
-        }
-        else
-        {
-            assert(false); // Unknown const type.
-        }
+    }
+    else if (auto enumeration = dynamic_pointer_cast<Slice::Enum>(type))
+    {
+        EnumeratorPtr enumerator = dynamic_pointer_cast<Enumerator>(valueType);
```
InsertCreativityHere, line 2076: Just as a general comment, this PR sometimes qualify things with `Slice::`, other times it doesn't.
Personally, I think it's fine to omit the `Slice::`, since there really shouldn't be any collisions and it keeps things looking tidier.

InsertCreativityHere (approved):
Looks good to me.
Just some small stuff, as usual!

cpp/src/slice2py/PythonUtil.cpp
pepone: for a follow up PR

Issue #4006, opened 2025-05-20
Override dispatch doc-comment in generated code


bernardnormier:
The generated code for an interface should always override the dispatch doc-comment, and not inherit the base Object.dispatch doc-comment.

Already implemented in C++:
https://github.com/zeroc-ice/ice/blob/1b7f68f822a5ff474ccc8d3a0b2d52db141b1a35/cpp/src/slice2cpp/Gen.cpp#L2860

Need to fix:
- [] C#
- [] Java
- [] JS/TS
- [] Swift

InsertCreativityHere:
`slice2swift` already generates a new doc-comment for the generated `dispatch` method:
https://github.com/zeroc-ice/ice/blob/eee7201c957fe4fc2c8c2d4a05c7d357a5bd1e5a/cpp/src/slice2swift/Gen.cpp#L1469-L1473

InsertCreativityHere:
C# and Java were done in #4216.

PR #4216, opened 2025-07-15
Override `dispatch` Doc-Comments in C# and Java

InsertCreativityHere:
This PR implements #4006 for C# and Java.
Instead of the generated `dispatch` methods inheriting an incorrect doc-comment from `Object.dispatch`,
now `slice2cs` and `slice2java` will generate an overriding doc-comment on `dispatch`.

The new comments are based off of the relevant comment in C++:
https://github.com/zeroc-ice/ice/blob/1b7f68f822a5ff474ccc8d3a0b2d52db141b1a35/cpp/src/slice2cpp/Gen.cpp#L2860-L2865
Including some language-specific things taken from each language's `Object.dispatch`.

### DIFF
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -4788,7 +4788,8 @@ Slice::Gen::ServantVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
         out << nl << " * @param request the incoming request";
         out << nl << " * @return the outgoing response";
         out << nl
-            << " * @throws com.zeroc.Ice.UserException if a {@code UserException} is thrown, Ice will marshal it as the response payload.";
+            << " * @throws com.zeroc.Ice.UserException if a {@code UserException} is thrown, Ice will marshal it as "
+               "the response payload.";
         out << nl << " */";
         out << nl << "@Override";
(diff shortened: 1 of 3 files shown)


cpp/src/slice2java/Gen.cpp
```diff
@@ -4780,8 +4780,18 @@ Slice::Gen::ServantVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
...
-            << "default java.util.concurrent.CompletionStage<com.zeroc.Ice.OutgoingResponse> dispatch("
+        out << nl << "/**";
+        out << nl
+            << " * Dispatches an incoming request to one of the methods of this generated interface, based on the "
+               "operation name carried by the request.";
+        out << nl << " *";
+        out << nl << " * @param request the incoming request";
+        out << nl << " * @return the outgoing response";
+        out << nl
+            << " * @throws com.zeroc.Ice.UserException when the response payload contains a {@code UserException}.";
```
pepone: I think it should be more like, when a user exception is thrown Ice marshals it as the response payload. The The current sentence sounds like the response payload already contains a exception and we throw the exception in this case.

pepone (approved):

InsertCreativityHere:
I updated the comment (in `Object` and the generated code) to:
```
If a {@code UserException} is thrown, Ice will marshal it as the response payload.
```

PR #4217, opened 2025-07-17
Add back loadSlice

pepone:
This PR adds back support for IcePy.loadSlice, the internal IcePy.compile has been removed.

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -2690,7 +2690,7 @@ Slice::Python::dynamicCompile(const vector<string>& files, const vector<string>&
     while (!fragments.empty())
     {
         size_t fragmentsSize = fragments.size();
-        for (vector<PythonCodeFragment>::iterator it = fragments.begin(); it != fragments.end();)
+        for (auto it = fragments.begin(); it != fragments.end();)
         {
             PythonCodeFragment fragment = *it;
             const auto& moduleImports = runtimeImports[fragment.moduleName];


cpp/src/slice2py/Python.cpp
```diff
@@ -336,37 +336,7 @@ Slice::Python::compile(const vector<string>& argv)
...
-                    if (++it != allDefinitions.end())
-                    {
-                        out << ",";
-                    }
-                }
-                out.dec();
-                out << nl << "]";
-                out << nl;
-            }
+            writePackageIndex(imports, out);
```
pepone, line 339: Moved package index creation to a function so that it can be used by both dynamic and static code generation.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -605,8 +642,6 @@ Slice::Python::ImportVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     addRuntimeImport("Ice.ObjectPrx", {"checkedCastAsync", "Ice_checkedCastAsync"}, p);
     addRuntimeImport("Ice.ObjectPrx", {"uncheckedCast", "Ice_uncheckedCast"}, p);
 
-    addRuntimeImport("typing", {"TYPE_CHECKING", ""}, p);
```
pepone, line 608: This was redundant the `addTypingImport` ensures this.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -902,8 +936,25 @@ Slice::Python::ImportVisitor::addRuntimeImportForMetaType(
 bool
 Slice::Python::PackageVisitor::visitModuleStart(const ModulePtr& p)
 {
-    // Add the __init__.py file to the list of generated modules.
-    _packageIndexFiles.insert(getMappedPackage(p) + "/__init__.py");
+    string packageName = getMappedPackage(p) + p->mappedName();
+
+    // Ensure all parent packages exits, that is necessary to account for modules
```
pepone, line 941: moving this here ensures outer packages are created before the inner packages. which is important for dynamic compilation and avoid the need to sort them later.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2543,92 +2576,220 @@ namespace
...
+        const Python::ModuleImportsMap& typingImports,
+        Output& out)
+    {
         out << sp;
         out << nl << "from __future__ import annotations";
         out << nl << "import IcePy";
 
-        outputFiles[moduleName] = std::move(output);
-        return out;
+        // Write the runtime imports.
```
pepone, line 2592: Would be good to improve this to write the imports in a format compatible with Ruff checks.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2543,92 +2576,220 @@ namespace
...
+            }
+            out.dec();
+        }
     }
 }
 
-void
-Slice::Python::generate(const UnitPtr& unit, const std::string& outputDir)
+vector<Slice::Python::PythonCodeFragment>
+Slice::Python::dynamicCompile(const vector<string>& files, const vector<string>& cppArgs, bool debug)
```
pepone, line 2632: The `IcePy.loadSlice` implementation calls this to get all the code fragments it will evaluate. The fragments are returned in the order they must be evaluated.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2543,92 +2576,220 @@ namespace
...
-        }
-        else
+        // If we didn't remove any module from the runtime imports, it means that we have a circular dependency between
+        // Slice files which is not allowed.
+        if (fragments.size() == fragmentsSize)
         {
-            moduleName = getPythonModuleForDefinition(fragment.contained);
+            assert(false);
+            throw runtime_error(
+                "Circular dependency detected in Slice files. Please check the imports and package definitions.");
```
pepone, line 2743: If I'm not mistaken the Slice parser already ensures this, and we should probably remove this extra check.

cpp/src/slice2py/PythonUtil.h
```diff
@@ -12,6 +12,33 @@
 
 namespace Slice::Python
 {
+    using ModuleImportsMap = std::map<std::string, std::set<std::pair<std::string, std::string>>>;
```
pepone, line 15: I will add more comments in a follow up PR that further refactors this code.

python/modules/IcePy/Init.cpp
```diff
@@ -96,7 +96,6 @@ static PyMethodDef methods[] = {
      METH_VARARGS,
      PyDoc_STR("internal function")},
     {"loadSlice", reinterpret_cast<PyCFunction>(IcePy_loadSlice), METH_VARARGS, PyDoc_STR("loadSlice(cmd) -> None")},
-    {"compile", reinterpret_cast<PyCFunction>(IcePy_compile), METH_VARARGS, PyDoc_STR("internal function")},
```
pepone, line 99: This internal function was no longer used, I removed it.

python/modules/IcePy/Slice.cpp
```diff
@@ -72,153 +65,76 @@ IcePy_loadSlice(PyObject* /*self*/, PyObject* args)
...
         return nullptr;
     }
 
     vector<string> cppArgs;
-    Ice::StringSeq includePaths;
     bool debug = false;
-    bool all = false;
-    if (opts.isSet("D"))
+
+    for (const auto& arg : opts.argVec("D"))
```
pepone, line 77: No need for the extra `if (opts.isSet` the `argVec` call should return an empty vector if not set.

python/modules/IcePy/Slice.cpp
```diff
@@ -72,153 +65,76 @@ IcePy_loadSlice(PyObject* /*self*/, PyObject* args)
...
+        PyErr_Format(PyExc_RuntimeError, "error during Slice compilation: %s", ex.what());
+        return nullptr;
+    }
 
-        // TODO rework loadSlice
-        // generate(u, all, includePaths, out);
-        if (u->getStatus() == EXIT_FAILURE)
+    for (const auto& fragment : fragments)
+    {
+        PyObject* moduleRef = PyImport_AddModule(fragment.moduleName.c_str());
```
pepone, line 107: This creates or adds the module and returns a borrow reference to it.

python/modules/IcePy/Slice.cpp
```diff
@@ -72,153 +65,76 @@ IcePy_loadSlice(PyObject* /*self*/, PyObject* args)
...
-        // We need to invoke Ice.updateModules() so that all of the types we've just generated
-        // are made "public".
-        //
-        code += "\nIce.updateModules()\n";
 
-        PyObjectHandle src{
-            Py_CompileString(const_cast<char*>(code.c_str()), const_cast<char*>(file.c_str()), Py_file_input)};
-        if (!src.get())
+        // The file name is set to the Slice file the code was generated from, or to the package index file name for
+        // package index fragments.
```
pepone, line 115: The package index is created from multiple Slice files, that is why we don't use a Slice file as the Py_file_input.

python/modules/IcePy/Slice.cpp
```diff
@@ -72,153 +65,76 @@ IcePy_loadSlice(PyObject* /*self*/, PyObject* args)
...
             return nullptr;
         }
 
-        PyObjectHandle globals{PyDict_New()};
-        if (!globals.get())
+        PyObject* moduleDict = PyModule_GetDict(moduleRef); // Borrowed reference
+        if (fragment.isPackageIndex)
         {
-            return nullptr;
+            // If this is a package index, we need to set the `__path__` attribute.
```
pepone, line 126: `__path__` is required for Python to interpret it as a package, an empty value is fine.

python/modules/IcePy/Slice.cpp
```diff
@@ -72,153 +65,76 @@ IcePy_loadSlice(PyObject* /*self*/, PyObject* args)
...
+            PyObject* path = PyList_New(1);
+            PyList_SetItem(path, 0, PyUnicode_FromString(fragment.fileName.c_str()));
+            PyDict_SetItemString(moduleDict, "__path__", path);
         }
 
-        PyDict_SetItemString(globals.get(), "__builtins__", PyEval_GetBuiltins());
-        PyObjectHandle val{PyEval_EvalCode(src.get(), globals.get(), nullptr)};
-        if (!val.get())
+        PyObjectHandle result{PyEval_EvalCode(code.get(), moduleDict, moduleDict)};
```
pepone, line 150: This evaluate the code in the context of the current Python module, represented by `moduleDict`

externl (approved):
Looks good!

PR #4218, opened 2025-07-17
Minor fix for better sequence typing imports

pepone:


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -702,7 +702,21 @@ Slice::Python::ImportVisitor::visitDataMembers(const ContainedPtr& parent, const
...
+            {
+                addTypingImport(sequence->type(), parent);
+                // TODO We should use list[T] for sequence data members.
+                addTypingImport("collections.abc", {"Sequence", ""}, parent);
+            }
+            else if (auto dictionary = dynamic_pointer_cast<Dictionary>(type))
+            {
+                addTypingImport(dictionary->keyType(), parent);
+                addTypingImport(dictionary->valueType(), parent);
+            }
```
pepone, line 715: For containers we need to import the container types not the container itself.

Additionally we need to import Sequence depending of the context. For now typeToTypeHintString is using Sequence everywhere but we need to provide additional context to decide if we should use sequece, list , a user type, etc.

PR #4219, opened 2025-07-17
More Python type hinting fixes

externl:
Various updates to Python mapping for type hinting. 

In Ice and Glacier2, just a few warnings left (including generated code):

```
❯ pyright python/Glacier2 python/Ice
/Users/joe/Developer/zeroc-ice/ice/python/python/Glacier2/SSLInfo.py
  /Users/joe/Developer/zeroc-ice/ice/python/python/Glacier2/SSLInfo.py:46:12 - error: "Sequence" is not defined (reportUndefinedVariable)
/Users/joe/Developer/zeroc-ice/ice/python/python/Ice/__init__.py
  /Users/joe/Developer/zeroc-ice/ice/python/python/Ice/__init__.py:181:16 - error: Module is not callable (reportCallIssue)
2 errors, 0 warnings, 0 informations 
```

Still need to fix these and issues in IceGrid/IceMx

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -1210,8 +1210,10 @@ Slice::Python::CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
     out << "))";
     out.dec();
 
+    // Use setattr to set the _ice_type attribute on the class. Linting tools will complain about this if we
+    // don't use setattr, as the _ice_type attribute is not defined in the class body.
     out << sp;
-    out << nl << valueName << "._ice_type = " << metaType;
+    out << nl << "setattr(" << valueName << ", '_ice_type', " << metaType << ")";
 
     out << sp;
     out << nl << "__all__ = [\"" << valueName << "\", \"" << metaType << "\"]";
@@ -1228,6 +1230,7 @@ Slice::Python::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     string prxName = className + "Prx";
     string metaType = getMetaType(p);
     InterfaceList bases = p->bases();
+    OperationList operations = p->operations();
 
     // Emit a forward declarations for the proxy meta type.
     BufferedOutput outF;
@@ -1289,7 +1292,6 @@ Slice::Python::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     out << nl << "super().__init__(communicator, proxyString)";
     out.dec();
 
-    OperationList operations = p->operations();
     for (const auto& operation : operations)
     {
         const string opName = operation->name();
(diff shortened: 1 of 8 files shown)


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1210,8 +1210,10 @@ Slice::Python::CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
     out << "))";
     out.dec();
 
+    // Use setattr to set the _ice_type attribute on the class. Linting tools will complain about this if we
+    // don't use setattr, as the _ice_type attribute is not defined in the class body.
     out << sp;
-    out << nl << valueName << "._ice_type = " << metaType;
+    out << nl << "setattr(" << valueName << ", '_ice_type', " << metaType << ")";
```
externl, line 1216: `_ice_type` is tricky as there's a circular dependency.

pepone: Can we do `_ice_type: IcePy.TypeInfo | None ` in the class declaration?

python/python/Ice/ObjectPrx.py
```diff
@@ -11,10 +11,19 @@
...
+    from .EndpointSelectionType import EndpointSelectionType
+    from .Identity import Identity
+    from .Locator import LocatorPrx
+    from .Router import RouterPrx
 
-def uncheckedCast(type, proxy, facet=None):
+    T = TypeVar("T", bound=ObjectPrx)
+
+
+def uncheckedCast(type: Type[T], proxy: ObjectPrx, facet: str | None = None) -> T | None:
```
pepone: proxy can be None according to the implementation. 

Can we use overloads? either pass None and returns None, or pass a proxy and returns a proxy of type T

python/python/Ice/ObjectPrx.py
```diff
@@ -69,7 +80,9 @@ def checkedCast(type, proxy, facet=None, context: dict[str, str] | None = None):
     return IcePy.ObjectPrx.newProxy(type, proxy) if proxy.ice_isA(type.ice_staticId(), context=context) else None
 
 
-async def checkedCastAsync(type, proxy, facet=None, context: dict[str, str] | None = None):
+async def checkedCastAsync(
+    type: Type[T], proxy: ObjectPrx, facet: str | None = None, context: dict[str, str] | None = None
```
pepone: proxy can be none here too.

pepone (approved):

cpp/src/slice2py/PythonUtil.cpp
externl: We could, but it's not really correct as it's never None.

python/python/Ice/ObjectPrx.py
```diff
@@ -40,16 +58,31 @@ def uncheckedCast(type, proxy, facet=None):
...
     Downcasts a proxy after confirming the target object's type via a remote invocation.
 
     Parameters
     ----------
     type : type
         The proxy target type.
-    proxy : ObjectPrx
-        The source proxy (can be None).
+
+    proxy : ObjectPrx or None
```
pepone, line 72: I think we should switch to `ObjectPrx | None` for handwritten code, generated code would use the same for type hint and docs. For a follow up PR.

python/python/Ice/ObjectPrx.py
```diff
@@ -4,26 +4,44 @@
...
     Downcasts a proxy without confirming the target object's type via a remote invocation.
 
     Parameters
     ----------
     type : type
         The proxy target type.
-    proxy : ObjectPrx
-        The source proxy (can be None).
+
+    proxy : ObjectPrx  or None
```
pepone:
```suggestion
    proxy : ObjectPrx or None
```

pepone (approved):

PR #4220, opened 2025-07-18
Improve Python sequence type-hints

pepone:
This PR improves the type hints generated for Sequences to account for metadata and also for the context where the type is used.

I have also done some cleanup of PythonUtil, by moving most declarations to the header files.

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -23,497 +23,345 @@ using namespace IceInternal;
 
 namespace
 {
-    /// Determines the mapped package for a given Slice definition.
-    /// @param p The Slice definition to get the mapped package for.
-    /// @param packageSeparator Use this character as the separator between package segments.
-    /// @return The mapped package name, with the specified separator.
-    string getMappedPackage(const SyntaxTreeBasePtr& p, char packageSeparator = '.')
-    {
-        if (dynamic_pointer_cast<Builtin>(p))
-        {
-            return string{"Ice"} + packageSeparator;
-        }
-        else
-        {
-            auto contained = dynamic_pointer_cast<Contained>(p);
-            assert(contained);
-            string package = contained->mappedScope(string{packageSeparator});
-            if (packageSeparator != '.')
-            {
-                // Replace "." with the specified separator.
-                replace(package.begin(), package.end(), '.', packageSeparator);
-            }
-            return package;
-        }
-    }
-
-    /// Returns the fully qualified name of the Python module that corresponds to the given Slice definition.
-    ///
-    /// Each Slice module is mapped to a Python package with the same name, but with "::" replaced by ".".
-    /// Within that package, each Slice definition is mapped to a Python module with the same name as the definition.
-    ///
-    /// For example:
-    /// - A Slice definition named `Baz` in the module `::Bar::Foo` is mapped to the Python module `"Bar.Foo.Baz"`.
-    ///
-    /// @param p The Slice definition to map to a Python module.
-    /// @return The fully qualified Python module name corresponding to the Slice definition.
-    string getPythonModuleForDefinition(const SyntaxTreeBasePtr& p)
-    {
-        if (auto builtin = dynamic_pointer_cast<Builtin>(p))
-        {
-            static const char* builtinTable[] =
-                {"", "", "", "", "", "", "", "", "Ice.Value", "Ice.ObjectPrx", "Ice.Value"};
-
-            return builtinTable[builtin->kind()];
-        }
-        else
-        {
-            auto contained = dynamic_pointer_cast<Contained>(p);
-            assert(contained);
-            return getMappedPackage(contained) + contained->mappedName();
-        }
-    }
-
-    /// Returns the fully qualified name of the Python module where the given Slice definition is forward-declared.
-    ///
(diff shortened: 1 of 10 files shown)


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -23,497 +23,345 @@ using namespace IceInternal;
...
+        static const char* builtinTable[] = {
+            "IcePy._t_byte",
+            "IcePy._t_bool",
+            "IcePy._t_short",
+            "IcePy._t_int",
+            "IcePy._t_long",
+            "IcePy._t_float",
+            "IcePy._t_double",
+            "IcePy._t_string",
+            "_Ice_Value_t",
```
pepone, line 108: We where using `__` double underscore prefix, but that is problematic when you try to use such variable inside a class name. Python mangle the names of double underscore names in classes.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -23,497 +23,345 @@ using namespace IceInternal;
...
-            "int",
-            "float",
-            "float",
-            "str",
-            "Ice.Object | None", // Not used anymore
-            "Ice.ObjectPrx | None",
-            "Ice.Value | None"};
-
-        if (auto builtin = dynamic_pointer_cast<Builtin>(type))
+        else if (auto seq = dynamic_pointer_cast<Sequence>(type))
```
pepone, line 188: This is the main update in this PR.

python/modules/IcePy/Types.cpp
```diff
@@ -2233,34 +2233,39 @@ IcePy::SequenceInfo::unmarshalPrimitiveSequence(
 bool
 IcePy::SequenceInfo::SequenceMapping::getType(const Ice::StringSeq& metadata, Type& t)
 {
-    for (auto p = metadata.begin(); p != metadata.end(); ++p)
+    for (const auto& name : metadata)
     {
-        if ((*p) == "python:seq:default" || (*p) == "python:default")
```
pepone, line 2238: Removed "default"

python/modules/IcePy/Types.cpp
```diff
@@ -2233,34 +2233,39 @@ IcePy::SequenceInfo::unmarshalPrimitiveSequence(
...
+            t = SEQ_LIST;
             return true;
         }
-        else if ((*p) == "python:seq:tuple" || (*p) == "python:tuple")
+        else if (name == "python:tuple")
         {
             t = SEQ_TUPLE;
             return true;
         }
-        else if ((*p) == "python:seq:list" || (*p) == "python:list")
```
pepone, line 2248: My understanding is that using `||` was bogus. `"python:seq:list` is supposed to overwrite `python:tuple`

python/python/Ice/Object.py
```diff
@@ -9,7 +9,7 @@
 import IcePy
 
 from .OperationMode import OperationMode
-from .StringSeq import __Ice_StringSeq_t as Ice_StringSeq_t
```
pepone, line 12: We needed the alias because `__` is mangled when we use it in Ice.Object implementation.

python/test/Ice/custom/AllTests.py
```diff
@@ -53,9 +53,9 @@ def allTests(helper, communicator):
         test(b2[i] == byteList[i])
 
     (r, b2) = custom.opByteList2(byteList)
```
pepone, line 55: Updated this test to account for the removal of "python:default" and "python:seq:default"

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -23,497 +23,345 @@ using namespace IceInternal;
...
+                    os << " | bytes";
+                }
+                if (seq->hasMetadata("python:numpy.ndarray"))
+                {
+                    assert(elementType);
+                    os << " | numpy.typing.NDArray[" << numpyBuiltinTable[elementType->kind()] << "]";
+                }
             }
-            else
+            else if (seq->hasMetadata("python:list"))
```
externl, line 223: Why do we have `list`? Is list not the default?

python/modules/IcePy/Types.cpp
externl: What is the logic for these? The comments in the validation seem to indicate that that they're just aliases.  I fee like we should not allow this overwrite behavior.

externl (approved):

python/modules/IcePy/Types.cpp
pepone: They are not alias, we have some documentation for this in:

https://doc.zeroc.com/ice/3.7/language-mappings/python-mapping/client-side-slice-to-python-mapping/python-mapping-for-sequences#id-.PythonMappingforSequencesv3.7-CustomizingtheSequenceMappinginPython

https://doc.zeroc.com/ice/3.7/the-slice-language/slice-metadata-directives#id-.SliceMetadataDirectivesv3.7-python:seq:default,python:seq:listandpython:seq:tuple

cpp/src/slice2py/PythonUtil.cpp
pepone: It is the default for most types, but not for bytes.

```
sequence<byte> ByteSeq; // -> bytes
["python:list"]sequence<byte> ByteList // list[int]
```

PR #4223, opened 2025-07-21
Improve the Slice Compiler's Handling of Bad Identifiers

InsertCreativityHere:
This PR changes how the Slice parser handles 'bad' or missing identifiers.
(Only for sequences, dictionaries, enums, and constants. I'll do the remaining types in a 2nd PR).

----

Currently, when we run across a bad/missing identifier, the Slice parser will do one of two things:
1) Generate a random UUID as a stand-in identifier.
2) Fail to create the type, and return `nullptr` instead of a `SequencePtr`, `EnumPtr`, etc.

Option 1 is bad because we will sometimes leak the UUID. If you write: `const byte = 300;`,
you'll get: `constant '38E0C286-AF9A-40C9-ABE2-37A7D4E9FDC5' has illegal...`

Option 2 prevents us from further validating inside the type, and can lead to bogus errors later on.
For example if you write `module { class C {} }`. You'll get an error for missing a module name,
but you'll also get a bogus error that `classes can only be defined within a module`.

----

So, this PR adds a new grammar rule `definition_name` to gracefully handle these situations.
1) **The Slice parser will _always_ create the type**, even if it has a bad identifier. No more returning `nullptr`.
2) **We always use the identifier the user provided**, no more random UUIDs. If they use a keyword, we use that as the identifier, if they forget an identifier, we use the empty string. 

Since the compiler will immediately exist after parsing (since it failed), the code-gen and validation is shielded from worrying about this change.


cpp/src/Slice/Grammar.cpp
InsertCreativityHere, line 1: Generated file. No need to review.

cpp/src/Slice/Grammar.y
```diff
@@ -376,30 +376,22 @@ opt_semicolon
 opt_semicolon
 | sequence_def
 {
-    assert($1 == nullptr || dynamic_pointer_cast<Sequence>($1));
```
InsertCreativityHere, line 379: No need to check `nullptr`; we _always_ create the type now.

cpp/src/Slice/Grammar.y
```diff
@@ -376,30 +376,22 @@ opt_semicolon
...
+    assert(dynamic_pointer_cast<Sequence>($1));
 }
 ';'
-| sequence_def
-{
-    currentUnit->error("';' missing after sequence definition");
-}
```
InsertCreativityHere, line 385: The Slice compiler already emits an error for syntax error for missing semicolons.
No need to have a separate rule for this.

cpp/src/Slice/Grammar.y
```diff
@@ -1339,29 +1331,20 @@ exception
...
-: ICE_SEQUENCE '<' metadata type '>' ICE_IDENTIFIER
+: ICE_SEQUENCE '<' metadata type '>' definition_name
 {
     auto ident = dynamic_pointer_cast<StringTok>($6);
     auto metadata = dynamic_pointer_cast<MetadataListTok>($3);
     auto type = dynamic_pointer_cast<Type>($4);
     ContainerPtr cont = currentUnit->currentContainer();
     $$ = cont->createSequence(ident->v, type, std::move(metadata->v));
 }
-| ICE_SEQUENCE '<' metadata type '>' keyword
```
InsertCreativityHere, line 1350: `definition_name` already emits an error for 'bad' identifiers.
So we can remove all these duplicated error-condition rules.

cpp/src/Slice/Grammar.y
```diff
@@ -1371,80 +1354,30 @@ dictionary_def
...
-    auto valueMetadata = dynamic_pointer_cast<MetadataListTok>($6);
-    auto valueType = dynamic_pointer_cast<Type>($7);
-    ContainerPtr cont = currentUnit->currentContainer();
-    $$ = cont->createDictionary(ident->v, keyType, std::move(keyMetadata->v), valueType, std::move(valueMetadata->v)); // Dummy
-    currentUnit->error("keyword '" + ident->v + "' cannot be used as dictionary name");
-}
-;
-
-// ----------------------------------------------------------------------
-enum_id
```
InsertCreativityHere, line 1388: This whole `enum_id` is just a poor-man's `definition_name`.

cpp/src/Slice/Grammar.y
```diff
@@ -1371,80 +1354,30 @@ dictionary_def
...
 // ----------------------------------------------------------------------
 enum_def
 // ----------------------------------------------------------------------
-: enum_id
+: ICE_ENUM definition_name
 {
-    auto ident = dynamic_pointer_cast<StringTok>($1);
+    auto ident = dynamic_pointer_cast<StringTok>($2);
     ContainerPtr cont = currentUnit->currentContainer();
     EnumPtr en = cont->createEnum(ident->v);
```
InsertCreativityHere, line 1366: We always create the enum now.
No need to repeatedly check if `en` is `nullptr`.

cpp/test/Slice/errorDetection/MissingIdentifier.ice
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) ZeroC, Inc.
```
InsertCreativityHere, line 1: We don't have any real testing of missing definition names, so I added a test here.

cpp/test/Slice/errorDetection/CaseInsensitive.err
```diff
@@ -6,7 +6,6 @@ CaseInsensitive.ice:26: module 'M1' is capitalized inconsistently with its previ
 CaseInsensitive.ice:30: module 'C1' is capitalized inconsistently with its previous name: 'c1'
 CaseInsensitive.ice:35: redefinition of parameter 'aa'
 CaseInsensitive.ice:36: parameter 'BB' differs only in capitalization from parameter 'bb'
-CaseInsensitive.ice:36: 'BB' has changed meaning
```
InsertCreativityHere: The real error is that `BB` and `bb` differ only in capitalization.
These `changed meaning` errors were bogus, and now they're gone.

cpp/src/Slice/Parser.cpp
```diff
@@ -2144,7 +2151,7 @@ Slice::Container::visitContents(ParserVisitor* visitor)
 bool
 Slice::Container::checkIntroduced(const string& scopedName, ContainedPtr namedThing)
 {
-    if (scopedName[0] == ':') // Only unscoped names introduce anything.
```
InsertCreativityHere, line 2147: If the user forgot an identifier, skip this check.
It's all about checking that identifiers are unambiguous.

cpp/src/Slice/Parser.cpp
```diff
@@ -2827,7 +2834,7 @@ Slice::ClassDef::ClassDef(const ContainerPtr& container, const string& name, int
       _base(std::move(base)),
       _compactId(id)
 {
-    if (_compactId != -1)
```
InsertCreativityHere, line 2830: Don't add an entry to the compact-id map if this has no identifier.
The code in it, assumes the identifier is non-empty.

InsertCreativityHere:
Ran everything through my compiler-comparison script, and this PR does not change the generated code.

cpp/src/Slice/Parser.cpp
```diff
@@ -1582,37 +1582,42 @@ Slice::Container::createStruct(const string& name, NodeType nodeType)
 }
 
 SequencePtr
-Slice::Container::createSequence(const string& name, const TypePtr& type, MetadataList metadata, NodeType nodeType)
```
InsertCreativityHere, line 1585: We used to pass a `NodeType` parameter to indicate if this was a 'real' element or a 'dummy' one.
Now we can just check if `name.empty()` instead. So this parameter isn't needed anymore.

cpp/src/Slice/Parser.cpp
```diff
@@ -1582,37 +1582,42 @@ Slice::Container::createStruct(const string& name, NodeType nodeType)
...
+Slice::Container::createSequence(const string& name, const TypePtr& type, MetadataList metadata)
 {
-    ContainedList matches = unit()->findContents(thisScope() + name);
-    if (!matches.empty())
+    SequencePtr p = make_shared<Sequence>(shared_from_this(), name, type, std::move(metadata));
+    _contents.push_back(p);
```
InsertCreativityHere, line 1588: We always create the element up-front now.

And we only run all the name related checks if `!name.empty()`.
There's no reason to check the names of malformed elements.

cpp/src/Slice/Parser.cpp
```diff
@@ -5016,6 +5023,7 @@ Slice::Unit::findDefinitionContext(string_view file) const
 void
 Slice::Unit::addContent(const ContainedPtr& contained)
 {
+    assert(!contained->name().empty());
```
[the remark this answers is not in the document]
InsertCreativityHere: This is actually a good thing.
We should never be calling `addContent` for Slice elements that are missing names.
This whole function is about storing elements in a <name, ptr> map so we can look them up later.
And it's impossible to look up an element with no name.

cpp/test/Slice/errorDetection/CaseInsensitive.err
```diff
@@ -28,8 +28,10 @@ CaseInsensitive.ice:146: dictionary 'D' differs only in capitalization from dict
 CaseInsensitive.ice:147: redefinition of module 'm1' as dictionary
 CaseInsensitive.ice:148: dictionary 'M1' differs only in capitalization from module 'm1'
 CaseInsensitive.ice:151: enumeration 'eN1' differs only in capitalization from enumeration 'en1'
+CaseInsensitive.ice:151: 'eN1' has changed meaning
```
InsertCreativityHere, line 31: We get some extra error messages now that validation can proceed a little further than before.
Some of these feel repetitive, but that's for a separate PR.

cpp/src/Slice/Grammar.y
```diff
@@ -376,30 +376,22 @@ opt_semicolon
...
 | enum_def
 {
-    assert($1 == nullptr || dynamic_pointer_cast<Enum>($1));
+    assert(dynamic_pointer_cast<Enum>($1));
 }
 opt_semicolon
 | const_def
 {
-    assert($1 == nullptr || dynamic_pointer_cast<Const>($1));
+    assert(dynamic_pointer_cast<Const>($1));
```
pepone, line 394: Is `"';' missing after const definition"` below still required?

cpp/src/Slice/Parser.cpp
```diff
@@ -2144,7 +2152,7 @@ Slice::Container::visitContents(ParserVisitor* visitor)
 bool
 Slice::Container::checkIntroduced(const string& scopedName, ContainedPtr namedThing)
 {
-    if (scopedName[0] == ':') // Only unscoped names introduce anything.
+    if (scopedName.empty() || scopedName[0] == ':') // Only unscoped names introduce anything.
```
pepone, line 2155: not changed by the PR, but using scopedName for this parameter, and then check if it is unscoped look awkward, would be better to use just name or a more appropriate param.

cpp/src/Slice/Parser.cpp
```diff
@@ -1622,87 +1627,92 @@ Slice::Container::createDictionary(
...
 EnumPtr
-Slice::Container::createEnum(const string& name, NodeType nodeType)
+Slice::Container::createEnum(const string& name)
 {
-    ContainedList matches = unit()->findContents(thisScope() + name);
-    if (!matches.empty())
+    EnumPtr p = make_shared<Enum>(shared_from_this(), name);
+    _contents.push_back(p);
+
+    checkForGlobalDefinition("enums");
```
externl, line 1685: What does this function do?

cpp/test/Slice/errorDetection/ConstDef.err
```diff
@@ -8,7 +8,7 @@ ConstDef.ice:74: initializer of type 'string' is incompatible with the type 'byt
 ConstDef.ice:75: initializer of type 'double' is incompatible with the type 'short' of constant 'ic6'
 ConstDef.ice:76: type of initializer is incompatible with the type 'int' of constant 'ic7'
 ConstDef.ice:77: initializer of type 'bool' is incompatible with the type 'long' of constant 'ic8'
-ConstDef.ice:79: missing constant name
+ConstDef.ice:79: missing name
```
externl, line 11: Should we use "identifier" instead of "name"

externl (approved):

cpp/src/Slice/Parser.cpp
InsertCreativityHere: This function reports an error if something is defined outside of a module.
The argument is just slapped into the error message, like: `enums can only be defined within a module`.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Agreed, I will fix right after this PR.
Since fixing it now gives me merge conflicts with my other PR!

cpp/src/Slice/Grammar.y
InsertCreativityHere: It's not required, but the 'default' error message the compiler emits for constants missing the trailing semicolon is very uninformative, so I left this in. Hopefully I can improve in a future PR.

cpp/test/Slice/errorDetection/ConstDef.err
InsertCreativityHere: We could, but I don't have any issue with `name`, and it's what we're already using in the error message.

Issue #4199, opened 2025-07-09
Remove `-E` Slice compiler options


pepone:
I think this is rarely used if at all.

`-E Print the preprocessor output on stdout.` We have deemphasized the use of the preprocessor for a while, and I think we can remove this.

externl:
Did you mean `-E` in the title?

InsertCreativityHere:
This option was removed in #4224.

PR #4224, opened 2025-07-21
Remove Compiler Support for '-E' (print preprocessed output) Option

InsertCreativityHere:
This PR implements #4199.


cpp/src/ice2slice/Main.cpp
```diff
@@ -162,78 +158,60 @@ compile(const vector<string>& argv)
             return EXIT_FAILURE;
         }
```
InsertCreativityHere, line 160: All I did was remove the `if (preprocess)` block, and dedent the `else` block.

pepone (approved):

externl (approved):
Looks good. Some man pages were just missing -E documentation?

PR #4226, opened 2025-07-22
Fix type hints and default values for Python constructors and dataclasses

externl:
This PR updates type hints and subsequently default values for Python constructors and data classes.


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1054,15 +1061,15 @@ Slice::Python::CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
 
     // ice_id
     out << sp;
-    out << nl << "def ice_id(self):";
+    out << nl << "def ice_id(self) -> str:";
```
externl, line 1209: Fix missing type hints

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1788,10 +1790,19 @@ Slice::Python::CodeVisitor::getTypeInitializer(const DataMemberPtr& field)
     {
         return getImportAlias(enumeration) + "." + enumeration->enumerators().front()->mappedName();
     }
-    else
+    else if (!constructor && isMutableTypeNonClassType(field->type()))
```
externl: `field(...)` can only be used on data members, not in constructors.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1890,12 +1901,19 @@ Slice::Python::CodeVisitor::writeAssign(const DataMemberPtr& member, Output& out
 {
     const string memberName = member->mappedName();
 
-    // Structures are treated differently (see bug 3676).
-    StructPtr st = dynamic_pointer_cast<Struct>(member->type());
-    if (st && !member->optional())
+    // Mutable types cannot be used as default values in Python as they are shared across instances.
+    if (isMutableNonOptionalType(member->type()) && !member->optional())
```
InsertCreativityHere: Maybe pull `member->type()` into a variable at the top, since we call this function in 4 separate places.

cpp/src/slice2py/PythonUtil.h
```diff
@@ -147,6 +147,10 @@ namespace Slice::Python
     /// Helper method to emit the generated code that format the fields of a type in __repr__ implementation.
     std::string formatFields(const DataMemberList& members);
 
+    /// Checks if the given Slice type corresponds to a mutable Python which is not always optional.
+    /// Namely if the type corresponds to a Python struct, sequence, or dictionary type.
```
InsertCreativityHere: Little pedantic, but Python doesn't really have structs. Maybe:
```suggestion
    /// Namely if the type corresponds to a Python dataclass, sequence, or dictionary type.
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1889,13 +1900,21 @@ void
...
-            << getImportAlias(st) << "()";
+        out << nl << "self." << memberName << " = " << memberName << " if " << memberName << " is not None else ";
+
+        if (dynamic_pointer_cast<Struct>(memberType))
+        {
+            out << getImportAlias(memberType) << "()";
+        }
+        else
+        {
+            out << typeToTypeHintString(memberType, false, member, false) << "()";
```
pepone: The type hint is not necessarily a constructor, for example `dict[str, str]()` would not be correct I think

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1788,10 +1790,19 @@ Slice::Python::CodeVisitor::getTypeInitializer(const DataMemberPtr& field)
...
+    else if (!constructor && isMutableNonOptionalType(field->type()))
     {
-        return "None";
+        if (auto st = dynamic_pointer_cast<Struct>(field->type()))
+        {
+            return "field(default_factory=" + getImportAlias(st) + ")";
+        }
+        else
+        {
+            return "field(default_factory=" + typeToTypeHintString(field->type(), false, field, false) + ")";
```
pepone: I don't think we can use a type hint as the default.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1965,9 +1984,12 @@ Slice::Python::CodeVisitor::writeConstructorParams(const DataMemberList& members
     out << "self";
     for (const auto& member : members)
     {
+        // Using mutable types in Python as default values is bad as they are shared across instances.
+        // The standard way to handle this is to use an optional with a default value of None, and then
+        // initialize the member in the constructor if it is None.
         const string typeHint = typeToTypeHintString(
             member->type(),
-            member->optional(),
+            member->optional() || isMutableNonOptionalType(member->type()),
```
pepone: Why do we need this?

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1889,13 +1900,21 @@ void
...
 {
     const string memberName = member->mappedName();
+    const TypePtr memberType = member->type();
 
-    // Structures are treated differently (see bug 3676).
-    StructPtr st = dynamic_pointer_cast<Struct>(member->type());
-    if (st && !member->optional())
+    // Mutable types cannot be used as default values in Python as they are shared across instances.
+    if (isMutableNonOptionalType(memberType) && !member->optional())
     {
```
pepone, line 2066: Maybe this method should be named `canBeUsedAsDefaultValue`

cpp/src/slice2py/PythonUtil.cpp
externl: Since we can't use a default value for these types: dataclasses, sequences, and dictionaries. They need to be optional in the constructor, then initialized in the body.

cpp/src/slice2py/PythonUtil.cpp
externl: Yes, just realized this. Am fixing.

cpp/src/slice2py/PythonUtil.cpp
externl: Yes, you're right. I've made some fixes. Will need to improve to account for metadata in a follow-up PR.

cpp/src/slice2py/PythonUtil.h
```diff
@@ -341,7 +346,7 @@ namespace Slice::Python
         void writeOperations(const InterfaceDefPtr&, IceInternal::Output&);
 
         // Get the default value for initializing a given type.
-        std::string getTypeInitializer(const DataMemberPtr&);
+        std::string getTypeInitializer(const DataMemberPtr& dataMember, bool constructor);
```
InsertCreativityHere: This parameter is named `field` in the source, but `dataMember` in the header.
We should pick one of these to use in both places.

cpp/src/slice2py/PythonUtil.h
```diff
@@ -147,6 +147,11 @@ namespace Slice::Python
     /// Helper method to emit the generated code that format the fields of a type in __repr__ implementation.
     std::string formatFields(const DataMemberList& members);
 
+    /// Checks if the given Slice type corresponds to non-optional type which can be used as default value in Python.
```
InsertCreativityHere, line 137:
```suggestion
    /// Checks if the given Slice type corresponds to a non-optional type which can be used as a default value in Python.
```
Or to keep it under 120 columns:
```suggestion
    /// Checks if the given Slice type corresponds to a non-optional type which is usable as a default value in Python.
```

pepone (approved):

PR #4227, opened 2025-07-22
Type hinting for test/Glacier2/router

externl:
I also generated some overloads to help with uncheckedCast.

### DIFF
--- a/python/test/Glacier2/router/Client.py
+++ b/python/test/Glacier2/router/Client.py
@@ -7,7 +7,6 @@ import threading
 from typing import Any
 
 from generated.test.Glacier2.router import Test
-
 from TestHelper import TestHelper
 
 import Glacier2


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1286,6 +1287,21 @@ Slice::Python::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     out << nl << "return Ice_checkedCastAsync(" << prxName << ", proxy, facet, context)";
     out.dec();
 
+    out << sp << nl << "@overload";
+    out << nl << "@staticmethod";
+    out << nl << "def uncheckedCast(proxy: Ice.ObjectPrx, facet: str | None = None) -> "
+        << prxTypeHint.substr(0, prxTypeHint.find_first_of(" | None")) << ":";
```
pepone, line 1428: I think we can use getImportAlias, after merging #4228

python/test/Glacier2/router/Client.py
```diff
@@ -5,14 +5,14 @@
...
-import Test
+from generated.client import Test
 from TestHelper import TestHelper
 
 import Glacier2
 import Ice
 
 
-def test(b):
+def test(b: object | None):
```
pepone: should we use "Any"?

cpp/src/slice2py/PythonUtil.cpp
externl: I'll make a note to do this.

InsertCreativityHere (approved):
Looks good to me!

PR #4228, opened 2025-07-22
Rework generated Python import statements to avoid alias

pepone:


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -70,24 +70,89 @@ Slice::Python::getPythonModuleForForwardDeclaration(const SyntaxTreeBasePtr& p)
...
+        }
     }
     return declarationModule;
 }
 
 string
-Slice::Python::getImportAlias(const SyntaxTreeBasePtr& p)
+Slice::Python::getImportAlias(
+    const ContainedPtr& source,
+    const std::map<std::string, std::string>& allImports,
```
InsertCreativityHere: Usually we don't qualify things with `std` in the source files, just the header files.
```suggestion
    const map<string, string>& allImports,
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -70,24 +70,95 @@ Slice::Python::getPythonModuleForForwardDeclaration(const SyntaxTreeBasePtr& p)
...
+
+    if (moduleName == source->mappedScoped("."))
+    {
+        // If the source module is the same as the module name being imported. We are using a
+        // definition from the current module and we don't need an alias.
+        useAlias = false;
+    }
+    else if (find(all.begin(), all.end(), name) != all.end())
+    {
+        // The name being bind comes from a different module and conflicts with one of the names
```
```suggestion
        // The name being bound comes from a different module and conflicts with one of the names
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -70,24 +70,95 @@ Slice::Python::getPythonModuleForForwardDeclaration(const SyntaxTreeBasePtr& p)
...
+        // exported by the source module. We need to use an alias.
+        useAlias = true;
+    }
+    else
+    {
+        // If the name being bind is already imported from a different module, we need to use an alias.
```
```suggestion
        // If the name being bound is already imported from a different module, we need to use an alias.
```

cpp/src/slice2py/PythonUtil.h
```diff
@@ -111,44 +111,34 @@ namespace Slice::Python
     /// If the definition represents an interface, the suffix `"Prx"` is appended to the alias to refer to the proxy
     /// type.
     ///
+    /// @param source The Slice definition that is importing the given definition.
```
externl, line 114: allImports is missing from doc comment.

cpp/src/slice2py/PythonUtil.h
```diff
@@ -111,44 +111,34 @@ namespace Slice::Python
...
     /// @param p The Slice definition to get the Python import alias for.
     /// @return The alias to use when importing the given definition in generated Python code.
-    std::string getImportAlias(const SyntaxTreeBasePtr& p);
+    std::string getImportAlias(
+        const ContainedPtr& source,
+        const std::map<std::string, std::string>& allImports,
+        const SyntaxTreeBasePtr& p);
+
+    std::string getImportAlias(
```
externl, line 122: Would be good to have a doc comment here.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -70,24 +70,95 @@ Slice::Python::getPythonModuleForForwardDeclaration(const SyntaxTreeBasePtr& p)
...
-        return builtinTable[builtin->kind()];
+        if (builtin->kind() == Builtin::KindObjectProxy)
+        {
+            return getImportAlias(source, allImports, "Ice.ObjectPrx", "ObjectPrx");
+        }
+        else if (builtin->kind() == Builtin::KindValue || builtin->kind() == Builtin::KindObject)
+        {
+            return getImportAlias(source, allImports, "Ice.Value", "Value");
+        }
+        return "";
```
externl, line 108: Is this an expected case?

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -70,24 +70,95 @@ Slice::Python::getPythonModuleForForwardDeclaration(const SyntaxTreeBasePtr& p)
...
+
+    // Whether we need to use an alias for the import.
+    bool useAlias = false;
+
+    if (moduleName == source->mappedScoped("."))
+    {
+        // If the source module is the same as the module name being imported. We are using a
+        // definition from the current module and we don't need an alias.
+        useAlias = false;
+    }
```
InsertCreativityHere, line 136: I guess it's nice for 'documentation' purposes, but there's no point to setting `useAlias = false` here,
it's already guaranteed to be false. Maybe there's a better way to structure this `if {} else if {} else {}`?

cpp/src/slice2py/PythonUtil.cpp
pepone: Yes for Python built-ins, when there is nothing to import.

cpp/src/slice2py/PythonUtil.h
pepone: I will go over this comment in a follow-up PR need to update a few things to reflect latest updates.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -156,15 +227,19 @@ Slice::Python::typeToTypeHintString(
         "float",
         "float",
         "str",
-        "Ice.Object | None", // Not used anymore
+        "Ice.Value | None", // Builtin::KindObject not used anymore
         "Ice.ObjectPrx | None",
         "Ice.Value | None"};
```
InsertCreativityHere, line 232: With the new code-paths none of these last 3 are used anymore right?
Since they're handled explicitly with our `if (builtin->kind() ...)` checks.

InsertCreativityHere (approved):
Looks good to me!

cpp/src/slice2py/PythonUtil.cpp
pepone: I will review this.

PR #4229, opened 2025-07-22
Make Doc-Comment Parsing a Discrete Compilation Phase

InsertCreativityHere:
This PR implements #3853.

Currently, we parse doc-comments on-demand, scattered throughout the code-gen.

Now, we perform all the parsing in a single, discrete phase.
When the Slice parser runs, it stores any doc-comments it sees as a raw `StringList`.
Then later on, each compiler will call `parseDocCommentsWithin` to parse those raw doc-comments in place.

Compilers that don't care about doc-comments (php and ruby) simply don't call this function.

### DIFF
--- a/cpp/src/Slice/DocCommentParser.cpp
+++ b/cpp/src/Slice/DocCommentParser.cpp
@@ -26,6 +26,8 @@ namespace Slice
         void visitEnum(const EnumPtr& p) final;
         void visitConst(const ConstPtr& p) final;
 
+        bool shouldVisitIncludedDefinitions() const final { return true; }
+
     private:
         void parseDocCommentFor(const ContainedPtr& p);
 
--- a/cpp/src/Slice/DocCommentParser.cpp
+++ b/cpp/src/Slice/DocCommentParser.cpp
@@ -1,0 +1,143 @@
+// Copyright (c) ZeroC, Inc.
+
+#include "DocCommentParser.h"
+
+using namespace std;
+using namespace Slice;
+
+namespace Slice
+{
+    class DocCommentParser final : public ParserVisitor
+    {
+    public:
+        DocCommentParser(DocLinkFormatter linkFormatter) : _linkFormatter(std::move(linkFormatter)) {};
+
+        bool visitModuleStart(const ModulePtr& p) final;
+        void visitClassDecl(const ClassDeclPtr& p) final;
+        bool visitClassDefStart(const ClassDefPtr& p) final;
+        void visitInterfaceDecl(const InterfaceDeclPtr& p) final;
+        bool visitInterfaceDefStart(const InterfaceDefPtr& p) final;
+        bool visitExceptionStart(const ExceptionPtr& p) final;
+        bool visitStructStart(const StructPtr& p) final;
+        void visitOperation(const OperationPtr& p) final;
+        void visitParameter(const ParameterPtr& p) final;
+        void visitDataMember(const DataMemberPtr& p) final;
+        void visitSequence(const SequencePtr& p) final;
+        void visitDictionary(const DictionaryPtr& p) final;
+        void visitEnum(const EnumPtr& p) final;
+        void visitConst(const ConstPtr& p) final;
+
+    private:
+        void parseDocCommentFor(const ContainedPtr& p);
+
+        DocLinkFormatter _linkFormatter;
+    };
+}
+
+void
+Slice::parseAllDocCommentsWithin(const UnitPtr& unit)
+{
+    if (auto linkFormatter = unit->linkFormatter())
+    {
+        DocCommentParser visitor{*linkFormatter};
+        unit->visit(&visitor);
+    }
+}
+
+bool
+DocCommentParser::visitModuleStart(const ModulePtr& p)
+{
+    parseDocCommentFor(p);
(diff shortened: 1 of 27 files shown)


cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -0,0 +1,143 @@
+// Copyright (c) ZeroC, Inc.
+
```
InsertCreativityHere, line 2: I'm going to move all the other doc-comment parsing code (scattered around Parser.cpp) into this file.
But for now, I left it alone in Parser.cpp to keep this PR reviewable.

cpp/src/Slice/Parser.cpp
```diff
@@ -407,19 +408,31 @@ namespace
...
+        return comment;
+    }
+}
+
+void
+Slice::DocComment::parse(const ContainedPtr& p, const DocLinkFormatter& linkFormatter)
+{
+    // TODO: this is a temporary hack since only "csharp" happens to set 'escapeXml'.
+    // If true, escapes all XML special characters in the parsed comment. Defaults to false.
+    const bool escapeXml = (p->unit()->languageName() == "cs");
```
InsertCreativityHere, line 432: I'm going to make this nicer in a follow-up PR.

cpp/src/Slice/Parser.cpp
```diff
@@ -4719,9 +4729,9 @@ Slice::DataMember::DataMember(
 // ----------------------------------------------------------------------
 
 UnitPtr
```
InsertCreativityHere, line 4731: We no longer inject the linkFormatter into the unit.
This gives each compiler more flexibility about when & how it handles doc-comments.

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -1935,11 +1935,11 @@ Slice::Gen::DataDefVisitor::visitExceptionStart(const ExceptionPtr& p)
             H << nl << "/// One-shot constructor to initialize all data members.";
             for (const auto& dataMember : allDataMembers)
             {
```
InsertCreativityHere, line 1937: I slightly simplified this.
We were storing the entire `DocComment`, when all we actually used was the overview.
So, now we just store the overview instead of copying the entire comment.

cpp/test/Slice/errorDetection/WarningInvalidComment.err
```diff
@@ -4,16 +4,6 @@ WarningInvalidComment.ice:11: warning: the '@exception' tag is only valid on ope
 WarningInvalidComment.ice:11: warning: the '@return' tag is only valid on operations
 WarningInvalidComment.ice:15: warning: missing link target after '@see' tag
 WarningInvalidComment.ice:17: warning: ignoring trailing '.' character in '@see' tag
-WarningInvalidComment.ice:43: warning: '@param NoParam' does not correspond to any parameter in operation 'voidOp'
```
pepone, line 7: Why are we removing these warnings?

pepone (approved):

cpp/test/Slice/errorDetection/WarningInvalidComment.err
InsertCreativityHere: Meant to comment this in the file... but these warnings were duplicates.
If you look at the bottom of this file, you'll see they appear a 2nd time.

This was because slice2cpp parsed some comments twice (in different visitors).

With the new setup we guarantee doc-comments are parsed exactly once.
Eliminating this duplication.

cpp/src/Slice/Parser.cpp
```diff
@@ -407,19 +408,31 @@ namespace
...
-    StringList lines = splitComment(p->docComment());
+    StringList lines = splitComment(rawDocComment);
     if (lines.empty())
     {
         return nullopt;
     }
+    else
+    {
+        DocComment comment;
+        comment._rawDocCommentLines = std::move(lines);
```
externl, line 422: Do we not have a constructor that takes lines?

cpp/src/Slice/DocCommentParser.h
```diff
@@ -0,0 +1,13 @@
...
+
+#ifndef DOC_COMMENT_PARSER_H
+#define DOC_COMMENT_PARSER_H
+
+#include "Parser.h"
+
+namespace Slice
+{
+    /// Parses all doc-comments within the provided unit (in-place).
+    void parseAllDocCommentsWithin(const UnitPtr& unit, DocLinkFormatter linkFormatter);
```
externl: Maybe just `parseDocComments`?

cpp/src/Slice/Parser.cpp
InsertCreativityHere: No we don't right now.
We could add one I guess, but it's also just this one spot, so I don't think it really matters.

cpp/src/Slice/DocCommentParser.h
InsertCreativityHere: I changed it to `parseAllDocComments`, to make sure it communicates that this is recursive.
Whereas `parseDocComments` sounds like it only parses doc-comments on the _current_ thing you pass in, not necessarily drilling into its contents.

PR #4234, opened 2025-07-23
Syntax Errors Now Say Which Tokens were Expected/Unexpected

InsertCreativityHere:
While looking into Bison's error recovery mechanisms, I came across the [parse.error option](https://www.gnu.org/software/bison/manual/html_node/_0025define-Summary.html#index-_0025define-parse_002eerror).
This PR defines `parse.error detailed`, which causes Bison to emit more detailed syntax errors.

Right now on main, this is all that Bison emits for syntax errors: `syntax error`. Very unhelpful.
Now it emits errors of the form: `syntax error, unexpected <BAD_TOKEN>, expecting <GOOD TOKENS>`

For example, if you forget a semicolon after your sequence:
```
module Foo {
    sequence<int> MySeq

    interface Hello {...}
}
```
the syntax error you'll get now is:
`syntax error, unexpected interface keyword, expecting ';'`
Much better.


cpp/src/Slice/Grammar.cpp
InsertCreativityHere, line 1: Generated file.
Don't need to review.

cpp/src/Slice/Grammar.h
InsertCreativityHere, line 1: Generated file.
Don't need to review.

cpp/src/Slice/Grammar.y
```diff
@@ -104,17 +104,8 @@ using namespace Slice;
 void
 slice_error(const char* s)
 {
-    // yacc and recent versions of Bison use "syntax error" instead
-    // of "parse error".
-
```
InsertCreativityHere, line 109: The version of Bison that we're using only uses 'syntax error'.
This `if` has probably been dead for years.

cpp/src/Slice/Grammar.y
```diff
@@ -151,54 +142,57 @@ namespace
 // Specify a custom location type for storing the location & filename of tokens.
 %define api.location.type {Slice::TokenContext}
 
+// Tell Bison to emit additional information with it's syntax error messages (like which tokens it expected).
+%define parse.error detailed
+
 // All keyword tokens. Make sure to modify the "keyword" rule in this
 // file if the list of keywords is changed. Also make sure to add the
 // keyword to the keyword table in Scanner.l.
```
InsertCreativityHere, line 150: These strings are the "human readable" token names Bison will use when reporting errors.

cpp/src/Slice/Grammar.y
```diff
@@ -313,91 +307,58 @@ definitions
...
-| class_decl
+| class_decl ';'
 {
     assert($1 == nullptr || dynamic_pointer_cast<ClassDecl>($1));
 }
-';'
-| class_decl
-{
-    currentUnit->error("';' missing after class forward declaration");
-}
```
InsertCreativityHere, line 329: I removed all these custom rules for detecting missing semicolons.
This made sense before, when all you'd see was `syntax error`.

But now that Bison tells you it expected `';'`, I don't think these extra rules add anything.

cpp/src/Slice/Grammar.y
```diff
@@ -313,91 +307,58 @@ definitions
 // ----------------------------------------------------------------------
 definition
 // ----------------------------------------------------------------------
-: module_def
+: module_def opt_semicolon
 {
     assert($1 == nullptr || dynamic_pointer_cast<Module>($1));
 }
-opt_semicolon
```
InsertCreativityHere, line 320: What we're doing here is called a 'midrule action'.
Where user code runs in the middle of an actions.

There is no benefit to checking between `module Foo` and the optional semicolon.
So I simplified this into a normal rule...

cpp/src/Slice/Grammar.y
```diff
@@ -1722,14 +1683,6 @@ metadata_list
...
     $$ = metadataList;
 }
-| metadata_list ',' BAD_TOKEN
-{
-    $$ = $1;
-}
-| BAD_TOKEN
-{
-    $$ = make_shared<MetadataListTok>();
-}
```
InsertCreativityHere, line 1732: I removed this rule, because otherwise Bison will 'expect' to see a BAD_TOKEN and say this to users:
`syntax error: unexpected ..., expected string literal or invalid character`
And we should never advise users to put invalid characters in their slice.

After I removed this rule it just says:
`syntax error: unexpected ..., expected string literal`

----

For the record, this rule was about detecting unquoted strings in metadata (ZeroC Slice requires all metadata be quoted). But now the parser just treats them as-if they were quoted (but we still emit an error of course).

cpp/src/Slice/Parser.cpp
```diff
@@ -2151,16 +2151,16 @@ Slice::Container::visitContents(ParserVisitor* visitor)
 }
 
 void
-Slice::Container::checkIntroduced(const string& scopedName, ContainedPtr namedThing)
+Slice::Container::checkIntroduced(const string& name, ContainedPtr namedThing)
```
InsertCreativityHere, line 2154: Unrelated cleanup from my last PR.

It feels weird to have a parameter named `scopedName`, when the first thing the function does is check if it's actually unscoped...

cpp/src/Slice/Scanner.cpp
InsertCreativityHere, line 1: Generated file, no need to review.

cpp/src/Slice/Scanner.l
```diff
@@ -472,7 +472,7 @@ floating_literal    (({fractional_constant}{exponent_part}?)|((\+|-)?{dec}+{expo
     StringTokPtr str = make_shared<StringTok>();
     str->literal = yytext;
     *yylval = str;
-    return BAD_TOKEN;
```
InsertCreativityHere, line 475: This is where we now treat unquoted string literals as-if they were quoted.

If you look at the lines just above you this you'll see it only affects `METADATA`, and we still emit an error message.
It just simplifies the grammar rules if we return a normal string-literal instead of a bad-token.

cpp/src/Slice/Grammar.y
```diff
@@ -104,17 +104,8 @@ using namespace Slice;
...
-
-    if (strcmp(s, "parse error") == 0)
-    {
-        currentUnit->error("syntax error");
-    }
-    else
-    {
-        currentUnit->error(s);
-    }
+    // We want 'unit' to emit the error so it's error count is incremented.
```
bernardnormier: its error count

cpp/src/Slice/Grammar.y
```diff
@@ -151,54 +142,57 @@ namespace
 // Specify a custom location type for storing the location & filename of tokens.
 %define api.location.type {Slice::TokenContext}
 
+// Tell Bison to emit additional information with it's syntax error messages (like which tokens it expected).
```
bernardnormier: its

bernardnormier (approved):

cpp/src/Slice/Grammar.y
InsertCreativityHere: Fixed!

cpp/src/Slice/Grammar.y
InsertCreativityHere: Also fixed! Good catches.

pepone (approved):

PR #4237, opened 2025-07-23
Suppress extra generated __init__

bernardnormier:
This PR updates slice2py to not generate `__init__` for proxies (we use the base class `__init__`. Same for classes and exceptions without their own fields.


pepone (approved):
Looks good, the constructors for classes and exception look identical, maybe we should have writeConstructor and use it for both.

InsertCreativityHere (approved):
Looks good to me!

PR #4238, opened 2025-07-23
Replace getMappedPackage with existing mappedScope and mappedScoped

pepone:


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -27,27 +27,6 @@ namespace
     const char* const tripleQuotes = R"(""")";
 }
 
-string
-Slice::Python::getMappedPackage(const SyntaxTreeBasePtr& p, char packageSeparator)
-{
-    if (dynamic_pointer_cast<Builtin>(p))
-    {
-        return string{"Ice"} + packageSeparator;
```
pepone, line 35: this path was never used.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -61,7 +40,7 @@ Slice::Python::getPythonModuleForDefinition(const SyntaxTreeBasePtr& p)
     {
         auto contained = dynamic_pointer_cast<Contained>(p);
         assert(contained);
-        return getMappedPackage(contained) + contained->mappedName();
+        return contained->mappedScoped(".");
```
pepone, line 43: The python generated module for a Slice definition is always the mapped scoped using '.' as the separator.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -187,7 +166,9 @@ Slice::Python::getMetaType(const SyntaxTreeBasePtr& p)
     {
         auto contained = dynamic_pointer_cast<Contained>(p);
         assert(contained);
-        string s = "_" + getMappedPackage(contained, '_') + contained->mappedName();
+        string s = "_" + contained->mappedScoped("_");
+        // Replace here is required in case any module is remap with python:identifier into a nested package.
+        replace(s.begin(), s.end(), '.', '_');
```
pepone, line 171: Here we need that '.' in `python:identifier` is also replaced, which is not done by `mappedScoped`.

PR #4240, opened 2025-07-24
Add support to run Python tests with dynamic generated code

pepone:
This PR add support to run the Python test suite with the dynamic generated code. Pass `--load-slice` to allTests.py.

I updated CI to also run tests with this flag, and removed the generated code to be sure things work as expected.

There are some updates for `loadSlice` implementation for tests that were not working correctly with the new implementation.

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -425,7 +425,7 @@ Slice::Python::createCodeFragmentForPythonModule(const ContainedPtr& contained, 
     fragment.sliceFileName = contained->file();
     string fileName = fragment.moduleName;
     replace(fileName.begin(), fileName.end(), '.', '/');
-    fragment.fileName = fileName + (isForwardDeclaration ? "_iceF.py" : ".py");
+    fragment.fileName = fileName + ".py";
     fragment.isPackageIndex = false;
     return fragment;
 }
@@ -2773,6 +2773,13 @@ Slice::Python::dynamicCompile(const vector<string>& files, const vector<string>&
         unit->destroy();
     }
 
+    vector<string> generatedModules;
+    for (const auto& fragment : fragments)
+    {
+        // Collect the names of all generated modules.
+        generatedModules.push_back(fragment.moduleName);
+    }
+
     // The list of non generated modules that are imported by the generated modules.
     vector<string> builtinModules{
         "Ice.FormatType",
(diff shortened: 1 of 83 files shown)


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -425,7 +425,7 @@ Slice::Python::createCodeFragmentForPythonModule(const ContainedPtr& contained,
     fragment.sliceFileName = contained->file();
     string fileName = fragment.moduleName;
     replace(fileName.begin(), fileName.end(), '.', '/');
-    fragment.fileName = fileName + (isForwardDeclaration ? "_iceF.py" : ".py");
```
pepone, line 428: we were adding `_iceF.py` twice , I plan to replace with `_forward.py` in a follow up PR.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2803,12 +2810,13 @@ Slice::Python::dynamicCompile(const vector<string>& files, const vector<string>&
         {
             PythonCodeFragment fragment = *it;
             const auto& moduleImports = runtimeImports[fragment.moduleName];
+
             bool unseenDependencies = false;
             for (const auto& [moduleName, _] : moduleImports)
             {
                 // If the current module depends on a module that is being generated but not yet seen we postpone its
                 // compilation.
-                if (runtimeImports.find(moduleName) != runtimeImports.end() &&
```
pepone, line 2811: Here we check if the module we depend on is part of this compilation, but "runtimeImports" doesn't include things like `_iceF` because they doesn't have any dependencies. This result in the "_iceF" modules not correctly sorted for compilation.

python/modules/IcePy/Slice.cpp
```diff
@@ -147,6 +148,9 @@ IcePy_loadSlice(PyObject* /*self*/, PyObject* args)
             }
         }
 
+        PyObject* builtins = PyEval_GetBuiltins();
+        PyDict_SetItemString(moduleDict, "__builtins__", builtins);
```
pepone, line 152: Without this importing modules can result in errors. When importing from outside a package.

python/test/Ice/objects/Objects.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) ZeroC, Inc.
+
```
pepone, line 2: Moved this from TestI to a separate file because the client depends on this definition, and don't want to include the server_private module used by TestI

PR #4246, opened 2025-07-25
Simplify getInvocationFuture

bernardnormier:
A tiny update to Ice.getInvocationFuture.

getInvocationFuture is just a convenience method:

```java
var f = Util.getInvocationFuture(cf);
```

is equivalent to:
```java
var f = (InvocationFuture<String>)cf;
```

except you don't need to specify the type.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
@@ -496,7 +496,8 @@ public final class Util {
      *
      * @param f The CompletableFuture returned by an asynchronous invocation.
      * @param <T> The result type.
-     * @return The InvocationFuture object.
+     * @return f casted to an {@code InvocationFuture<T>}.
+     * @throws ClassCastException if f is not an {@code InvocationFuture<T>}.
      */
     public static <T> InvocationFuture<T> getInvocationFuture(CompletableFuture<T> f) {
         return (InvocationFuture<T>) f;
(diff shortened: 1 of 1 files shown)


InsertCreativityHere (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -492,18 +492,14 @@ public static EncodingVersion currentEncoding() {
...
+     * @return f casted to an {@code InvocationFuture<T>}.
+     * @throws ClassCastException if f is not an {@code InvocationFuture<T>}.
      */
-    public static <T> InvocationFuture<T> getInvocationFuture(
-            CompletableFuture<T> f) {
-        if (!(f instanceof InvocationFuture)) {
-            throw new IllegalArgumentException(
-                "future did not originate from an asynchronous proxy invocation");
-        }
+    public static <T> InvocationFuture<T> getInvocationFuture(CompletableFuture<T> f) {
```
pepone, line 502: Should we mark this as deprecated? Users can just do the cast doesn't seems that useful.

pepone (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
bernardnormier: That's what I wanted to do first but there is actually a small utility to this method: no need to write the type. And we use it quite a bit in the tests / it's not trivial to fix, so may as well keep it non-deprecated.

For example:

```java
Util.getInvocationFuture(someAsyncCall).invocationFutureMethod(...);
```
vs
```java
((InvocationFuture<SomeComplicatedType>)someAsyncCall).invocationFutureMethod(...);
```

What's odd is that we don't return an `InvocationFuture<T>`, like we do in other languages.

PR #4251, opened 2025-07-28
Update generated _iceI_ methods in Java

bernardnormier:
This PR updates the generated `_iceI_` proxy methods in Java:
 - they are now `private` instead of `default` (a Java9 feature)
 - they are now generated after the 4 or 8 mapped public methods


InsertCreativityHere (approved):
Looks good to me!

PR #4255, opened 2025-07-29
Remove KindObject from builtin table

bernardnormier:
This cleanup PR removes KindObject from the Parser's builtin table.

`Object` without `*` is an alias for `Value`, and it was odd to keep this separate entry, especially since the "natural" mapping, Ice.Object, is never a mapped type.

The only slight downside is the error message when using `Object` incorrectly: it now says `Value`.

### DIFF
--- a/cpp/src/slice2java/JavaUtil.cpp
+++ b/cpp/src/slice2java/JavaUtil.cpp
@@ -233,7 +233,7 @@ Slice::Java::mapsToJavaBuiltinType(const TypePtr& p)
 {
     if (auto builtin = dynamic_pointer_cast<Builtin>(p))
     {
-        return builtin->kind() < Builtin::KindObject;
+        return builtin->kind() <= Builtin::KindString;
     }
     return false;
 }
@@ -303,11 +303,7 @@ Slice::Java::getStaticId(const TypePtr& type, const string& package)
 
     assert((b && b->usesClasses()) || cl);
 
-    if (b && b->kind() == Builtin::KindObject)
-    {
-        return "com.zeroc.Ice.Object.ice_staticId()";
-    }
-    else if (b && b->kind() == Builtin::KindValue)
+    if (b && b->kind() == Builtin::KindValue)
     {
         return "com.zeroc.Ice.Value.ice_staticId()";
     }
@@ -341,7 +337,6 @@ Slice::Java::typeToString(
         "float",
         "double",
         "String",
-        "com.zeroc.Ice.Object",
         "com.zeroc.Ice.ObjectPrx",
         "com.zeroc.Ice.Value"};
 
@@ -353,7 +348,6 @@ Slice::Java::typeToString(
         "java.util.OptionalLong",
         "java.util.Optional<java.lang.Float>",
         "java.util.OptionalDouble",
-        "???",
         "???",
         "???",
         "???"};
@@ -382,7 +376,6 @@ Slice::Java::typeToString(
                     return getUnqualified(builtinOptionalTable[builtin->kind()], package);
                 }
                 case Builtin::KindString:
-                case Builtin::KindObject:
                 case Builtin::KindObjectProxy:
                 case Builtin::KindValue:
                 {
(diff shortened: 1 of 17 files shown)


cpp/src/slice2matlab/Gen.cpp
```diff
@@ -3043,11 +3036,10 @@ CodeVisitor::marshal(
         const TypePtr content = seq->type();
         const BuiltinPtr b = dynamic_pointer_cast<Builtin>(content);
 
-        if (b && b->kind() != Builtin::KindObject && b->kind() != Builtin::KindObjectProxy &&
-            b->kind() != Builtin::KindValue)
+        if (b && b->kind() != Builtin::KindObjectProxy && b->kind() != Builtin::KindValue)
```
pepone: maybe ` <= Builtin::KindString` like we do in other places.

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -3043,11 +3036,10 @@ CodeVisitor::marshal(
...
         {
             static const char* builtinTable[] =
-                {"Byte", "Bool", "Short", "Int", "Long", "Float", "Double", "String", "???", "???", "???", "???"};
+                {"Byte", "Bool", "Short", "Int", "Long", "Float", "Double", "String", "???", "???"};
```
pepone: can we remove the "???" entries, we check in the if that they are not in this range.

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -3286,11 +3277,10 @@ CodeVisitor::unmarshal(
...
```
pepone: ditto

pepone (approved):

js/src/Ice/Operation.js
```diff
@@ -28,7 +28,6 @@ const builtinHelpers = [
     FloatHelper,
     DoubleHelper,
     StringHelper,
-    Value,
     ObjectPrx,
     Value,
```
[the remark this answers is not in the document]
bernardnormier: This PR removes this entry everywhere. There is no mismatch.

cpp/src/slice2matlab/Gen.cpp
bernardnormier: Fixed

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -233,7 +233,7 @@ Slice::Java::mapsToJavaBuiltinType(const TypePtr& p)
 {
     if (auto builtin = dynamic_pointer_cast<Builtin>(p))
```
InsertCreativityHere, line 233: At the top of this file we have a `builtInBufferTable` that was missed.
Of course, we don't actually use the entry for Object, but that array has 11 elements when it should now only have 10.

It's also just not a very well written method, but that's a separate thing.

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -3286,11 +3276,9 @@ CodeVisitor::unmarshal(
...
         const BuiltinPtr b = dynamic_pointer_cast<Builtin>(content);
 
-        if (b && b->kind() != Builtin::KindObject && b->kind() != Builtin::KindObjectProxy &&
-            b->kind() != Builtin::KindValue)
+        if (b && b->kind() <= Builtin::KindString)
         {
-            static const char* builtinTable[] =
-                {"Byte", "Bool", "Short", "Int", "Long", "Float", "Double", "String", "???", "???", "???", "???"};
+            static const char* builtinTable[] = {"Byte", "Bool", "Short", "Int", "Long", "Float", "Double", "String"};
             string bs = builtinTable[b->kind()];
```
InsertCreativityHere: Not your changes, but bs is a dead variable. Looks like we meant to use it on the next line... but forgot to??
```suggestion
```

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -3043,11 +3036,9 @@ CodeVisitor::marshal(
...
             out << nl << stream << ".write" << builtinTable[b->kind()] << "Seq";
```
InsertCreativityHere: Not your changes, but `bs` is a dead variable. Looks like we meant to use it in the 2nd line... but forgot to??
```suggestion
            out << nl << stream << ".write" << builtinTable[b->kind()] << "Seq";
```

InsertCreativityHere (approved):
Looks good to me!

cpp/src/slice2java/JavaUtil.cpp
bernardnormier: Fixed

PR #4257, opened 2025-07-29
Update Python mapping for class and exception to use dataclass

bernardnormier:
This PR also refactors a little slice2py to reduce code duplication.


python/python/Ice/__init__.py
```diff
@@ -340,7 +339,6 @@
     "currentProtocolEncoding",
     "dispatch",
     "encodingVersionToString",
-    "format_fields",
```
bernardnormier, line 343: As I understand it, format_fields and StringUtil.py were used to generate `__repr__` in struct/class/exception; it's not generated by dataclass.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2100,39 +2014,6 @@ Slice::Python::CodeVisitor::writeConstantValue(
     }
 }
 
-void
-Slice::Python::CodeVisitor::writeConstructorParams(
```
bernardnormier, line 2104: Replaced by dataclass.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2004,46 +1958,6 @@ Slice::Python::CodeVisitor::writeMetadata(const MetadataList& metadata, Output&
     out << ')';
 }
 
-void
-Slice::Python::CodeVisitor::writeAssign(const ContainedPtr& source, const DataMemberPtr& member, Output& out)
```
bernardnormier, line 2008: Replaced by dataclass.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1913,23 +1884,6 @@ Slice::Python::CodeVisitor::getTypeInitializer(const ContainedPtr& source, const
     return "None";
 }
 
-void
-Slice::Python::CodeVisitor::writeRepr(const ContainedPtr& p, const DataMemberList& members, Output& out)
```
bernardnormier, line 1917: Now generated by dataclass

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1619,150 +1734,6 @@ Slice::Python::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
...
-
-    out << sp;
-    out << nl << "__all__ = [\"" << name << "\", \"" << metaType << "\"]";
-
-    _codeFragments.push_back(createCodeFragmentForPythonModule(p, out.str()));
-    return false;
-}
-
-bool
-Slice::Python::CodeVisitor::visitStructStart(const StructPtr& p)
```
bernardnormier, line 1705: Moved up.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1619,150 +1734,6 @@ Slice::Python::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     return false;
 }
 
-bool
-Slice::Python::CodeVisitor::visitExceptionStart(const ExceptionPtr& p)
```
bernardnormier, line 1623: Moved.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1208,7 +1232,98 @@ Slice::Python::CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
...
+
+    out << sp;
+    out << nl << "__all__ = [\"" << name << "\", \"" << metaType << "\"]";
+
+    _codeFragments.push_back(createCodeFragmentForPythonModule(p, out.str()));
+    _out.reset();
+}
+
+void
+Slice::Python::CodeVisitor::visitDataMember(const DataMemberPtr& p)
```
bernardnormier, line 1305: Shared method for struct/class/exception fields.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -587,83 +644,6 @@ Slice::Python::ImportVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
...
-        // If the exception has no base, we import the Ice.UserException type.
-        addRuntimeImport("Ice.UserException", "UserException", p);
-    }
-    // Add imports required for the data members.
-    visitDataMembers(p, allDataMembers);
-    return false;
-}
-
-void
-Slice::Python::ImportVisitor::visitDataMembers(const ContainedPtr& parent, const list<DataMemberPtr>& members)
```
pepone, line 637: We previously has this method because we want to visit all-data members, for the generated `__init__`. I guess we no longer need this with the dataclass update.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -503,9 +506,63 @@ Slice::Python::ImportVisitor::visitClassDefStart(const ClassDefPtr& p)
...
+
+    // Import field if we have at least one data member that cannot be used as a default value.
+    // This is required to use the `dataclasses.field` function to initialize the field.
+    if (!canBeUsedAsDefaultValue(p->type()))
+    {
+        addRuntimeImport("dataclasses", "field", parent);
+    }
+
+    // Add imports required for data member types.
+    auto type = p->type();
```
pepone: move this up as we also call it in `!canBeUsedAsDefaultValue(p->type())`

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1128,38 +1165,29 @@ Slice::Python::CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
     _codeFragments.push_back(createCodeFragmentForPythonModule(p->declaration(), outF.str()));
 
     // Emit the class definition.
-    BufferedOutput out;
+    _out = std::make_unique<BufferedOutput>();
+    auto& out = *_out;
+
+    // Equality for Slice classes is reference equality, so we disable the dataclass eq.
+    out << nl << "@dataclass(eq=False)";
```
pepone, line 1171: Should we do the same for exceptions?

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1128,38 +1165,29 @@ Slice::Python::CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
     _codeFragments.push_back(createCodeFragmentForPythonModule(p->declaration(), outF.str()));
 
     // Emit the class definition.
-    BufferedOutput out;
+    _out = std::make_unique<BufferedOutput>();
+    auto& out = *_out;
```
pepone, line 1168: Would be good to find a better pattern, maybe keep the output in the fragment. And have createFragment/currentFragment like:

```
auto out = createFragment(p);

out = currentFragment(p);
```

python/python/Ice/__init__.py
pepone: Yes format_fields was a helper to  use in our previous `__repr__`

pepone (approved):

cpp/src/slice2py/PythonUtil.cpp
bernardnormier: It was mostly replaced by the standard visitDataMember (singular). We rely on the dataclass-generated `__init__`.

cpp/src/slice2py/PythonUtil.cpp
bernardnormier: I used a similar pattern for slice2matlab:
https://github.com/zeroc-ice/ice/blob/4baf1edb5ee9fea1a577b91a5866d51d1224e5e6/cpp/src/slice2matlab/Gen.h#L61

cpp/src/slice2py/PythonUtil.cpp
bernardnormier: No, we can keep the generated eq. I first tried to keep eq for classes but this causes a number of tests to fail.

For exceptions, it's ok to compare them "by value".

cpp/src/slice2py/PythonUtil.cpp
bernardnormier: Fixed

PR #4259, opened 2025-07-30
Simplify and unify generation of dependency files

pepone:
This PR reworks how Slice compiler generate dependency files. Slice compiler support three different formats for dependency files:

- Makefile style dependencies (--depend).
- Dependencies encoded as XML (--depend-xml).
- Dependencies encoded as JSON (--depend-json) slice2js only.

Before this PR this work but parsing the output of `mcpp -M`, which emits Makefile style dependencies, and transform it to the desired style.

The new implementation gets the dependencies directly from the parser `Unit::getAllFiles` returns all files that were parsed by the unit, in other words its dependencies, and then a regular visitor that collects and emit the dependencies in the desired format.

There is some additional cleanup for Slice compilers main.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -1,6 +1,7 @@
 // Copyright (c) ZeroC, Inc.
 
 #include "Parser.h"
+#include "../Ice/FileUtil.h"
 #include "DeprecationReporter.h"
 #include "DocCommentParser.h"
 #include "Ice/StringUtil.h"
@@ -162,6 +163,12 @@ Slice::DefinitionContext::filename() const
     return _filename;
 }
 
+const string&
+Slice::DefinitionContext::resolvedFilename() const
+{
+    return _resolvedFilename;
+}
+
 int
 Slice::DefinitionContext::includeLevel() const
 {
@@ -175,9 +182,10 @@ Slice::DefinitionContext::seenDefinition() const
 }
 
 void
-Slice::DefinitionContext::setFilename(string filename)
+Slice::DefinitionContext::setFilename(string filename, string resolvedFilename)
 {
     _filename = std::move(filename);
+    _resolvedFilename = std::move(resolvedFilename);
 }
 
 void
@@ -4835,9 +4843,59 @@ Slice::Unit::currentLine() const
 }
 
 int
-Slice::Unit::setCurrentFile(const std::string& currentFile, int lineNumber)
+Slice::Unit::setCurrentFile(std::string currentFile, int lineNumber)
 {
     assert(!currentFile.empty());
+
+    string absolutePath;
+    if (IceInternal::isAbsolutePath(currentFile))
+    {
+        absolutePath = currentFile;
+        if (!IceInternal::fileExists(absolutePath))
+        {
+            // MCPP can report bogus absolute paths, for files using relative includes.
+            // see: https://github.com/zeroc-ice/ice/issues/4253
+
+            // If the file is an absolute path, and it does not exists, it must be relative to the current definition
+            // context.
+            DefinitionContextPtr dc = currentDefinitionContext();
+            assert(dc);
+            string previousDir = dirName(dc->resolvedFilename());
+
+            vector<string> tokens1;
+            vector<string> tokens2;
+
+            IceInternal::splitString(previousDir, "/\\", tokens1);
+            IceInternal::splitString(absolutePath, "/\\", tokens2);
+
+            auto i1 = tokens1.begin();
+            auto i2 = tokens2.begin();
+
+            // Skip the common path elements.
+            while (i1 != tokens1.end() && i2 != tokens2.end() && *i1 == *i2)
+            {
+                i1++;
+                i2++;
+            }
+
+            // The remaining elements represent the path relative the current definition context.
+            absolutePath = "";
+            for (; i2 != tokens2.end(); ++i2)
+            {
+                absolutePath += "/" + *i2;
+            }
+
+            absolutePath = normalizePath(previousDir + absolutePath);
+            currentFile = absolutePath;
+        }
+    }
+    else
+    {
+        // Relative paths are relative to the current definition context, which represents the file being parsed.
+        DefinitionContextPtr dc = currentDefinitionContext();
+        assert(dc);
+        absolutePath = normalizePath(Slice::dirName(dc->resolvedFilename()) + "/" + currentFile);
+    }
 
     enum LineType
     {
@@ -4892,10 +4950,20 @@ Slice::Unit::setCurrentFile(const std::string& currentFile, int lineNumber)
         }
     }
 
-    DefinitionContextPtr dc = currentDefinitionContext();
-    assert(dc);
-    dc->setFilename(currentFile);
-    _definitionContextMap.insert(make_pair(currentFile, dc));
+    if ((type == File && _allFiles.empty()) || type == Push)
+    {
+        // Either this is the first file being parsed, or we are pushing a new file onto the stack.
+
+        DefinitionContextPtr dc = currentDefinitionContext();
+        assert(dc);
+        dc->setFilename(currentFile, absolutePath);
+        _definitionContextMap.insert(make_pair(currentFile, dc));
+
+        if (find(_allFiles.begin(), _allFiles.end(), absolutePath) == _allFiles.end())
+        {
+            _allFiles.push_back(absolutePath);
+        }
+    }
 
     return static_cast<int>(type);
 }
@@ -5075,12 +5143,7 @@ Slice::Unit::includeFiles() const
 StringList
 Slice::Unit::allFiles() const
 {
-    StringList result;
-    for (const auto& [key, value] : _definitionContextMap)
-    {
-        result.push_back(key);
-    }
-    return result;
+    return _allFiles;
 }
 
 int
--- a/cpp/src/Slice/Parser.h
+++ b/cpp/src/Slice/Parser.h
@@ -217,10 +217,12 @@ namespace Slice
         DefinitionContext(int includeLevel);
 
         [[nodiscard]] const std::string& filename() const;
+        [[nodiscard]] const std::string& resolvedFilename() const;
+        [[nodiscard]] int line() const;
         [[nodiscard]] int includeLevel() const;
         [[nodiscard]] bool seenDefinition() const;
 
-        void setFilename(std::string filename);
+        void setFilename(std::string filename, std::string resolvedFilename);
         void setSeenDefinition();
 
         [[nodiscard]] const MetadataList& getMetadata() const;
@@ -237,6 +239,7 @@ namespace Slice
         int _includeLevel;
         MetadataList _metadata;
         std::string _filename;
+        std::string _resolvedFilename;
         bool _seenDefinition{false};
         std::set<WarningCategory> _suppressedWarnings;
     };
@@ -1095,7 +1098,7 @@ namespace Slice
         [[nodiscard]] const std::string& topLevelFile() const;
         [[nodiscard]] int currentLine() const;
 
-        int setCurrentFile(const std::string& currentFile, int lineNumber);
+        int setCurrentFile(std::string currentFile, int lineNumber);
         [[nodiscard]] int currentIncludeLevel() const;
 
         void addFileMetadata(MetadataList metadata);
@@ -1159,6 +1162,7 @@ namespace Slice
         std::map<std::string, DefinitionContextPtr, std::less<>> _definitionContextMap;
         std::map<std::int32_t, std::string> _typeIds;
         std::map<std::string, std::set<std::string>> _fileTopLevelModules;
+        StringList _allFiles;
     };
 
     extern Unit* currentUnit; // The current parser for bison/flex
--- a/cpp/src/Slice/Preprocessor.cpp
+++ b/cpp/src/Slice/Preprocessor.cpp
@@ -126,19 +126,17 @@ Slice::Preprocessor::normalizeIncludePath(const string& path)
 
 namespace
 {
-    vector<string> baseArgs(vector<string> args, bool keepComments, const string& languageArg, const string& fileName)
+    vector<string> baseArgs(vector<string> args, const string& languageArg, const string& fileName)
     {
-        if (keepComments)
-        {
-            args.emplace_back("-C");
-        }
+        // -C keep comments
+        args.emplace_back("-C");
+
+        // UTF-8 encoding
         args.emplace_back("-e");
         args.emplace_back("en_us.utf8");
 
-        //
-        // Define version macros __ICE_VERSION__ is preferred. We keep
-        // ICE_VERSION for backward compatibility with 3.5.0.
-        //
+        // Define version macros __ICE_VERSION__ is preferred. We keep ICE_VERSION for backward compatibility
+        // with 3.5.0.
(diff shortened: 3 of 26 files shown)


cpp/src/ice2slice/Main.cpp
```diff
@@ -88,24 +88,24 @@ compile(const vector<string>& argv)
         return EXIT_SUCCESS;
     }
 
-    vector<string> cppArgs;
+    vector<string> preprocessorArgs;
```
pepone, line 91: I renamed `cppArgs`, `cppHandle` and `icecpp` in all compilers for "better" names. That is `preprocessorArgs`, `preprocessedHandle` and `preprocessor` respectively.

cpp/src/ice2slice/Main.cpp
```diff
@@ -138,76 +138,56 @@ compile(const vector<string>& argv)
...
-        UnitPtr p = Unit::createUnit("", false);
-        int parseStatus = p->parse(*i, cppHandle, debug);
-
-        if (!icecpp->close())
-        {
-            p->destroy();
-            return EXIT_FAILURE;
-        }
+            PreprocessorPtr preprocessor = Preprocessor::create(argv[0], fileName, preprocessorArgs);
+            FILE* preprocessedHandle = preprocessor->preprocess("-D__ICE2SLICE__");
```
pepone, line 147: Updated `preprocess` to throw an std::exception rather than returning null, to simplify error handling.

cpp/src/ice2slice/Main.cpp
```diff
@@ -138,76 +138,56 @@ compile(const vector<string>& argv)
...
-        }
-        else
-        {
-            parseAllDocComments(p, Slice::slice2LinkFormatter);
+            unit = Unit::createUnit("", false);
+            int parseStatus = unit->parse(fileName, preprocessedHandle, debug);
 
-            DefinitionContextPtr dc = p->findDefinitionContext(p->topLevelFile());
-            assert(dc);
+            preprocessor->close();
```
pepone, line 153: Also updated `Preprocessor::close` to trow.

cpp/src/Slice/Parser.cpp
```diff
@@ -162,6 +163,12 @@ Slice::DefinitionContext::filename() const
     return _filename;
 }
 
+const string&
+Slice::DefinitionContext::resolvedFilename() const
+{
+    return _resolvedFilename;
```
pepone, line 169: The absolute path for the Slice file.

cpp/src/Slice/Parser.cpp
```diff
@@ -4892,10 +4950,20 @@ Slice::Unit::setCurrentFile(const std::string& currentFile, int lineNumber)
         }
     }
 
-    DefinitionContextPtr dc = currentDefinitionContext();
-    assert(dc);
-    dc->setFilename(currentFile);
-    _definitionContextMap.insert(make_pair(currentFile, dc));
+    if ((type == File && _allFiles.empty()) || type == Push)
```
pepone, line 4958: There was no point in keep inserting entries in the definition context map, when we pop a definition context, of when this method is not pushing a new file.

cpp/src/Slice/Parser.cpp
```diff
@@ -5075,12 +5143,7 @@ Slice::Unit::includeFiles() const
 StringList
 Slice::Unit::allFiles() const
 {
-    StringList result;
-    for (const auto& [key, value] : _definitionContextMap)
-    {
-        result.push_back(key);
-    }
-    return result;
+    return _allFiles;
```
pepone, line 5151: Added `_allFiles` to preserve the order in which we visited them, first entry is the top-level file itself.

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -126,19 +126,17 @@ Slice::Preprocessor::normalizeIncludePath(const string& path)
 
 namespace
 {
-    vector<string> baseArgs(vector<string> args, bool keepComments, const string& languageArg, const string& fileName)
```
pepone, line 129: Removed keepComments, and with this update we were always passing "true".

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -154,34 +152,25 @@ namespace
...
 
 FILE*
-Slice::Preprocessor::preprocess(bool keepComments, const string& languageArg)
+Slice::Preprocessor::preprocess(const string& languageArg)
 {
-    if (!checkInputFile())
-    {
-        return nullptr;
-    }
+    checkInputFile();
```
pepone, line 157: Updated it to trow.

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -252,398 +239,19 @@ Slice::Preprocessor::preprocess(bool keepComments, const string& languageArg)
...
-    //
     // Calling this again causes the memory buffers to be freed.
-    //
     mcpp_use_mem_buffers(1);
 
     return _cppHandle;
 }
 
-bool
-Slice::Preprocessor::printMakefileDependencies(
```
pepone, line 268: That is the old generation of dependencies, despite its name it also emit xml, json fragments.

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -252,398 +239,19 @@ Slice::Preprocessor::preprocess(bool keepComments, const string& languageArg)
...
-                        {
-                            string s = includePaths[static_cast<size_t>(p - fullIncludePaths.begin())] +
-                                       file.substr(p->length());
-                            if (IceInternal::isAbsolutePath(newFile) || s.size() < newFile.size())
-                            {
-                                newFile = s;
-                            }
-                        }
-                    }
-                    file = newFile;
```
pepone, line 445: I removed this logic in the new implementation because I think it is bogus.

Here we shorten file names, by resolving them with the given include paths and settle for the one that produces the shortest path.

But when Make is processing the dependency files, it doesn't know about this include paths. I think this is also true for other tools.

cpp/src/Slice/SliceUtil.cpp
```diff
@@ -18,63 +18,61 @@ using namespace std;
 using namespace Slice;
 using namespace IceInternal;
 
-namespace
+string
```
pepone, line 21: moved to Slice namespace to use in the setCurrentFile impl.

cpp/src/Slice/SliceUtil.cpp
```diff
@@ -605,3 +624,137 @@ Slice::getTypeScopedName(const TypePtr& type)
         return contained->scoped();
     }
 }
+
+string
+Slice::relativePath(const string& path1, const string& path2)
```
pepone, line 629: I moved it from JsUtil, but this end-up being used only slice2js, so I should probably move it back to slice2js.

cpp/src/Slice/SliceUtil.cpp
```diff
@@ -605,3 +624,137 @@ Slice::getTypeScopedName(const TypePtr& type)
...
+    }
+    newPath += f1;
+
+    return newPath;
+}
+
+void
+Slice::DependencyVisitor::visitUnitEnd(const UnitPtr& unit)
+{
+    _dependencyMap[unit->topLevelFile()] = unit->allFiles();
```
pepone, line 687: Collect all dependencies for the unit.

cpp/src/Slice/SliceUtil.cpp
```diff
@@ -605,3 +624,137 @@ Slice::getTypeScopedName(const TypePtr& type)
...
+    StringList dependencies = _dependencyMap[source];
+    // Write the target
+    if (dependencies.empty())
+    {
+        os << target << ":" << std::endl;
+    }
+    else
+    {
+        os << target << ": \\" << std::endl;
+        for (const auto& dependency : dependencies)
```
pepone, line 706: We emit absolute paths, we can use `relativePath` and emit dependencies relative to the working directory, either would be fine.

cpp/src/Slice/Parser.h
```diff
@@ -217,10 +217,12 @@ namespace Slice
         DefinitionContext(int includeLevel);
 
         [[nodiscard]] const std::string& filename() const;
+        [[nodiscard]] const std::string& resolvedFilename() const;
```
bernardnormier, line 228: Would be nice to add a comment that describes the difference between filename and resolvedFilename.

cpp/src/Slice/Parser.cpp
```diff
@@ -4835,10 +4843,60 @@ Slice::Unit::currentLine() const
...
+    string absolutePath;
+    if (IceInternal::isAbsolutePath(currentFile))
+    {
+        absolutePath = currentFile;
+        if (!IceInternal::fileExists(absolutePath))
+        {
+            // MCPP can report bogus absolute paths, for files using relative includes.
+            // see: https://github.com/zeroc-ice/ice/issues/4253
+
+            // If the file is an absolute path, and it does not exists, it must be relative to the current definition
```
bernardnormier: does not exist

csharp/tools/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
```diff
@@ -114,9 +114,8 @@ protected override int ExecuteTool(string pathToTool, string responseFileCommand
                 foreach (ITaskItem source in Sources)
                 {
                     var inputs = new List<string>();
-                    var depends = dependsDoc.DocumentElement.SelectNodes(
-                        string.Format("/dependencies/source[@name='{0}']/dependsOn",
-                                        source.GetMetadata("Identity")));
```
pepone, line 119: This use of XPath was brittle as it relies on string match, the new helper FindSourceNode uses FileInfo to compare file paths.

cpp/src/slice2rb/Ruby.cpp
```diff
@@ -89,48 +90,48 @@ Slice::Ruby::compile(const vector<string>& argv)
...
         consoleErr << argv[0] << ": error: no input file" << endl;
         usage(argv[0]);
         return EXIT_FAILURE;
     }
 
-    if (depend && dependxml)
+    if (depend && dependXML)
     {
-        consoleErr << argv[0] << ": error: cannot specify both --depend and --dependxml" << endl;
+        consoleErr << argv[0] << ": error: cannot specify both --depend and --depend-xml" << endl;
```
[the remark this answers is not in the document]
pepone: `--dependxml` was bogus, the message should match the option name.

cpp/src/Slice/Util.h
```diff
@@ -83,5 +85,42 @@ namespace Slice
...
+    [[nodiscard]] std::string relativePath(const std::string& path1, const std::string& path2);
+
+    /// The DependencyVisitor class is used to collect dependencies of Slice units.
+    class DependencyVisitor : public ParserVisitor
+    {
+    public:
+        void visitUnitEnd(const UnitPtr& unit) final;
+
+        /// Writes the dependencies in Makefile format to the specified file. If the file is empty, it writes the
+        /// dependencies to the standard output.
```
InsertCreativityHere: To make it clear that we're talking about `if (file == "")` and not `if the file is empty (on your filesystem)`
```suggestion
        /// Writes the dependencies in Makefile format to the specified file. If 'dependFile' is empty, it writes the
        /// dependencies to standard output instead.
```

cpp/src/Slice/Util.h
```diff
@@ -83,5 +85,42 @@ namespace Slice
...
+        ///
+        /// @param dependFile The file to write the dependencies to or empty to write to standard output.
+        /// @param source The source file for which dependencies are being written.
+        /// @param target The target file that is generated from the source. This is used as the Makefile target.
+        void
+        writeMakefileDependencies(const std::string& dependFile, const std::string& source, const std::string& target);
+
+        /// Writes the dependencies in XML format to the specified file. If the file is empty, it writes the
+        /// dependencies to the standard output.
```
InsertCreativityHere: Same:
```suggestion
        /// Writes the dependencies in XML format to the specified file. If 'dependFile' is empty, it writes the
        /// dependencies to standard output instead.
```

cpp/src/Slice/Util.h
```diff
@@ -83,5 +85,42 @@ namespace Slice
...
+        ///
+        /// This method write the dependencies for all visited units in a single XML document.
+        ///
+        /// @param dependFile The file to write the dependencies to or empty to write to standard output.
+        void writeXMLDependencies(const std::string& dependFile);
+
+        /// Writes the dependencies in JSON format to the specified file. If the file is empty, it writes the
+        /// dependencies to the standard output.
```
InsertCreativityHere: Same:
```suggestion
        /// Writes the dependencies in JSON format to the specified file. If 'dependFile' is empty, it writes the
        /// dependencies to standard output instead.
```

cpp/src/Slice/Parser.h
```diff
@@ -216,11 +216,21 @@ namespace Slice
...
+        /// This may be an absolute or relative path.
+        ///
+        /// @return A reference to the file name.
         [[nodiscard]] const std::string& filename() const;
+
+        /// Returns the resolved file name of the definition context. This is always an absolute path.
+        ///
+        /// @return A reference to the resolved file name.
+        [[nodiscard]] const std::string& resolvedFilename() const;
+        [[nodiscard]] int line() const;
```
InsertCreativityHere: What is this `line` function?

cpp/src/Slice/Util.h
```diff
@@ -83,5 +85,42 @@ namespace Slice
...
     /// @return The fully qualified name of the type.
     std::string getTypeScopedName(const TypePtr& type);
+
+    [[nodiscard]] std::string relativePath(const std::string& path1, const std::string& path2);
+
+    /// The DependencyVisitor class is used to collect dependencies of Slice units.
+    class DependencyVisitor : public ParserVisitor
+    {
+    public:
+        void visitUnitEnd(const UnitPtr& unit) final;
```
InsertCreativityHere: I don't think it's meaningful for this to be a `Visitor`.
`Visitor`s are really all about traveling down the Syntax tree, which this class doesn't care about.
I think it'd be simpler for this to just be a stand-alone class and replace `visitUnitEnd` with `addDependenciesFor`.

Also, should probably be a `final` class.

cpp/src/Slice/SliceUtil.cpp
```diff
@@ -605,3 +624,136 @@ Slice::getTypeScopedName(const TypePtr& type)
...
+Slice::DependencyVisitor::visitUnitEnd(const UnitPtr& unit)
+{
+    _dependencyMap[unit->topLevelFile()] = unit->allFiles();
+}
+
+void
+Slice::DependencyVisitor::writeMakefileDependencies(
+    const string& dependFile,
+    const string& source,
+    const std::string& target)
```
InsertCreativityHere: No need to qualify string here:
```suggestion
    const string& target)
```

InsertCreativityHere (commented):
Looks good so far! Still a couple files to finish looking at though.

cpp/src/Slice/Parser.h
pepone: ~~Not sure it was already there.~~ Actually I think copilot might have added it because it resonates with filename.  Removed.

cpp/src/Slice/Util.h
pepone: Reworked as suggested

InsertCreativityHere (approved):
Looks good to me!

PR #4263, opened 2025-07-30
Small __str__ fixes in Python

bernardnormier:


python/python/Ice/UserException.py
```diff
@@ -4,10 +4,9 @@
 
 
 class UserException(IceException):
-    """The base class for all user-defined exceptions."""
-
-    def __str__(self) -> str:
```
bernardnormier, line 9: Removed because useless.

python/python/Ice/Value.py
```diff
@@ -51,8 +55,5 @@ def ice_getSlicedData(self) -> SlicedData | None:
         """
         return getattr(self, "_ice_slicedData", None)
 
-    def __str__(self) -> str:
```
bernardnormier, line 54: Ditto

python/python/Ice/ObjectPrx.py
```diff
@@ -989,4 +989,4 @@ def __repr__(self) -> str:
         )
 
     def __str__(self) -> str:
-        return repr(self)
+        return self.ice_toString()
```
bernardnormier, line 992: Fixed to return the stringified proxy.

pepone (approved):
Looks good

PR #4268, opened 2025-07-31
Python: improve type-hint for sequence types

pepone:
This PR improve the Python generated type hint for sequence types to account for the metadata. It also update the initialization of sequence fields to use a factory appropriate for the mapped type.

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -186,9 +186,6 @@ Slice::Python::CodeVisitor::typeToTypeHintString(
         }
     }
 
-    static constexpr string_view builtinTable[] =
-        {"int", "bool", "int", "int", "int", "float", "float", "str", "Ice.ObjectPrx | None", "Ice.Value | None"};
-
     if (auto builtin = dynamic_pointer_cast<Builtin>(type))
     {
         if (builtin->kind() == Builtin::KindObjectProxy)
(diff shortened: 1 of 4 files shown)


cpp/src/slice2py/PythonUtil.h
```diff
@@ -273,7 +268,10 @@ namespace Slice::Python
         }
 
     private:
-        void visitDataMembers(const ContainedPtr&, const std::list<DataMemberPtr>&);
+        /// Add the runtime imports for the given Sequence definition.
+        /// @param definition is the Sequence definition being imported.
```
bernardnormier: This should be: "The Sequence definition ...". Same for the next param.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -546,7 +529,11 @@ Slice::Python::ImportVisitor::visitDataMember(const DataMemberPtr& p)
     // For fields with a type that is a Struct, we need to import it as a RuntimeImport, to
     // initialize the field in the constructor. For other contained types, we only need the
     // import for type hints.
-    if (dynamic_pointer_cast<Struct>(type) || dynamic_pointer_cast<Enum>(type))
+    if (auto sequence = dynamic_pointer_cast<Sequence>(type))
```
bernardnormier, line 532: It's not clear to me why we don't have a special import-handling for dictionaries as well. How does it work?

pepone: This custom handling is for handling sequence metadata. For dictionary we always map it to dict.

The key/value types are handled separately like the sequence element type.

See:

https://github.com/pepone/ice/blob/1e8ce8409b8f224be611c0c3fa9a5ac7279cc16f/cpp/src/slice2py/PythonUtil.cpp#L852-L860

InsertCreativityHere (approved):
Looks good to me!

PR #4269, opened 2025-07-31
Refactor Slice for Python compiler to increase code sharing

pepone:


cpp/src/slice2py/Main.cpp
```diff
@@ -220,70 +196,39 @@ main(int argc, char* argv[])
         Ice::CtrlCHandler ctrlCHandler;
         ctrlCHandler.setCallback(interruptedCallback);
 
-        DependencyGenerator dependencyGenerator;
+        auto dependencyGenerator = make_unique<DependencyGenerator>();
```
pepone, line 199: The dependency generator is only used by slice2py executable.

cpp/src/slice2py/Main.cpp
```diff
@@ -220,70 +196,39 @@ main(int argc, char* argv[])
...
+        {
+            compilationKind = PythonCompilationKind::Index;
+        }
+        else
+        {
+            compilationKind = PythonCompilationKind::All;
         }
 
-        if (status == EXIT_FAILURE)
+        PythonCompilationResult compilationResult = Slice::Python::compile(
```
pepone: Maybe these should be just `CompilationResult` and `CompilationKind`.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2552,53 +2521,134 @@ namespace
...
 
-        unit->visit(&packageVisitor);
-        unit->visit(&importVisitor);
+        for (auto& fragment : fragments)
+        {
+            Python::BufferedOutput out;
+            writeHeader(out);
+            writeImports(runtimeImports[fragment.moduleName], typingImports[fragment.moduleName], out);
+            out << sp;
+            out << fragment.code;
```
pepone, line 2646: I think would be clear to write the imports, and the header in the code visitor when we start the fragment.

There is no good reason to delay this.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2552,53 +2521,134 @@ namespace
...
-            importVisitor.getTypingImports(),
-            importVisitor.getAllImportNames()};
-        unit->visit(&codeVisitor);
+            fragment.code = out.str();
+        }
 
-        const vector<PythonCodeFragment>& newFragments = codeVisitor.codeFragments();
-        fragments.insert(fragments.end(), newFragments.begin(), newFragments.end());
-        unit->destroy();
+        fragments = sortCodeFragments(fragments, importVisitor.getRuntimeImports());
```
pepone: sortCodeFragments is only required when call by loadSlice, I will add a param to let the method know if sorting is required.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2607,29 +2657,11 @@ Slice::Python::dynamicCompile(const vector<string>& files, const vector<string>&
...
-        "Ice.ObjectPrx_forward",
-        "Ice.OperationMode",
-        "Ice.StringUtil",
-        "Ice.UserException",
-        "Ice.Value",
-        "Ice.Value_forward",
-        "abc",
-        "dataclasses",
-        "enum",
-        "typing"};
```
pepone, line 2672: This list is not required, we only check dependencies that are part of the current compilation and builtin modules are by definition not part of the Slice compilation.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2844,26 +2812,26 @@ Slice::Python::validatePythonMetadata(const UnitPtr& unit)
     // "python:identifier"
     MetadataInfo identifierInfo = {
         .validOn =
-            {typeid(Module),
-             typeid(InterfaceDecl),
-             typeid(Operation),
-             typeid(ClassDecl),
+            {typeid(Slice::Module),
```
pepone, line 2841: Validation metadata started to emit warnings, after I add `PythonCompilationKind::Module` that is the second time I got bitten by a similar error.

Would be clear to use an enum, is not like you can apply metadata to any typeid.

PR #4273, opened 2025-08-01
Bring back IcePy.compile for slice2py script

pepone:
I earlier removed `IcePy.compile` when refactoring the Python compilation because appear to be unused. But it is used by the slice2py script ship with the PIP package.


cpp/src/slice2py/Main.cpp
```diff
@@ -1,78 +1,11 @@
...
-#include <iostream>
-#include <mutex>
-#include <string>
-
 using namespace std;
-using namespace Slice;
-using namespace Slice::Python;
 using namespace IceInternal;
 
-namespace
```
pepone, line 25: Moved all this code to PythonUtil so that it can be used by IcePy.

python/dist/lib/slice2py.py
```diff
@@ -12,7 +12,7 @@ def main():
     # Automatically add the slice dir.
     if sliceDir is not None:
         sys.argv.append("-I" + sliceDir)
-    val = IcePy.compile(sys.argv)
+    val = IcePy.compileSlice(sys.argv)
```
pepone, line 15: Renamed it to compileSlice, which seems more consistent with loadSlice. This is a purely internal function.

PR #4276, opened 2025-08-04
Clean Up Redefinition/Changed Meaning Checks in the Slice Compiler

InsertCreativityHere:
This PR:
1) demystifies the `checkIntroduced` function in the Slice compiler. This function is really about detecting when things changed meaning, so I updated it's name and added a doc-comment.
2) adds a helper function for checking when a redefinition has occurred, instead of having this check copy-pasted in 5 separate places. There was also a small bug in the old logic that was fixed.


cpp/src/Slice/Parser.cpp
```diff
@@ -4646,7 +4590,7 @@ Slice::Parameter::tag() const
 string
 Slice::Parameter::kindOf() const
 {
-    return "parameter declaration";
```
InsertCreativityHere, line 4649: I saw in some error messages that we were calling these "parameter declarations".
I don't think we ever call them that anywhere else. So I updated the name to just 'parameter'.

I checked the entire repo for this string and the only place that needed updating was in the tests.

cpp/src/Slice/Parser.cpp
```diff
@@ -2237,17 +2272,11 @@ Slice::Container::checkIntroduced(const string& name, ContainedPtr namedThing)
...
-            // We don't want to emit an error if one of them is a parameter/member, but the other isn't.
-            // For example, if one is a parameter and the other is a class, they're logically in different scopes.
-            // But, if both of them _are_ a parameter, or both of them _are not_ a parameter, then they're in the same
-            // scope, and we should report an error. The same applies for data members.
-            if ((isFirstAParameter == isSecondAParameter) && (isFirstADataMember == isSecondADataMember))
+            auto isInSelfContainedScope = [](const ContainedPtr& p) {
+                return dynamic_pointer_cast<DataMember>(p) || dynamic_pointer_cast<Parameter>(p);
+            };
+
+            if (!isInSelfContainedScope(it->second) && !isInSelfContainedScope(namedThing))
```
InsertCreativityHere: This is slightly different than the old logic. With the old logic, we would still emit `has changed meaning` errors if `it->second` and `namedThing` were both parameters or both data-members. Now we never emit `has changed meaning` errors for parameters or data-members.

This is because parameters and data-members are in their own scope, and can't have child elements.
So when they "change meaning", it's really just a 'redefinition error'. No need to emit an extra error on top of the redefinition error about it `changing meaning`... You'll see what I mean in the test updates.

cpp/test/Slice/errorDetection/CaseInsensitive.err
```diff
@@ -1,26 +1,22 @@
-CaseInsensitive.ice:8: redefinition of operation 'op' as operation 'op'
+CaseInsensitive.ice:8: redefinition of operation 'op'
 CaseInsensitive.ice:14: operation 'oP' differs only in capitalization from operation 'op'
-CaseInsensitive.ice:14: redefinition of operation 'op' as operation 'oP'
```
InsertCreativityHere, line 3: There was a bug in the operation check that caused us to emit 2 errors, when only 1 was really necessary.
The new helper function fixes this bug.

cpp/test/Slice/errorDetection/CaseInsensitive.err
```diff
@@ -1,26 +1,22 @@
...
 CaseInsensitive.ice:19: operation 'ice_isa' differs only in capitalization from operation 'ice_isA', which is defined in a base interface
 CaseInsensitive.ice:26: module 'M1' is capitalized inconsistently with its previous name: 'm1'
 CaseInsensitive.ice:30: module 'C1' is capitalized inconsistently with its previous name: 'c1'
 CaseInsensitive.ice:35: redefinition of parameter 'aa'
 CaseInsensitive.ice:36: parameter 'BB' differs only in capitalization from parameter 'bb'
-CaseInsensitive.ice:36: 'BB' has changed meaning
```
InsertCreativityHere, line 9: We no longer emit these spurious `changed meaning` errors for parameters and data-members.
The real error is that you redefined them (or they only differed in capitalization).

parameters and data members cannot change meaning since they're in their in their own scope.

cpp/src/Slice/Parser.cpp
```diff
@@ -76,6 +76,41 @@ namespace
 
         return false;
     }
+
+    /// Reports any naming conflicts between @p name and @p definitions.
+    /// This should only be called for Slice elements that are _not_ defined at module scope.
+    /// For example, this is fine to use for operations, enumerators, data members, and parameters.
+    /// For elements that are defined at global scope, `findContents` must be used instead.
```
externl: This comment seems incorrect :)

cpp/src/Slice/Parser.cpp
```diff
@@ -76,6 +76,41 @@ namespace
...
+    /// @return @c false if there are no name conflicts, @c true otherwise.
+    bool doesNameConflict(string_view name, string_view kind, const ContainedList& definitions)
+    {
+        const string lowerName = IceInternal::toLower(name);
+        const auto match = std::find_if(
+            definitions.begin(),
+            definitions.end(),
+            [&](const auto& p) { return lowerName == IceInternal::toLower(p->name()); });
+        if (match != definitions.end())
+        {
```
externl: I think we should refer to fail early. 

```
 if (match == definitions.end())
{
    return false;
}

File: cpp/src/Slice/Parser.cpp [hunk 7]
```diff
@@ -2237,17 +2273,10 @@ Slice::Container::checkIntroduced(const string& name, ContainedPtr namedThing)
...
-
-            // We don't want to emit an error if one of them is a parameter/member, but the other isn't.
-            // For example, if one is a parameter and the other is a class, they're logically in different scopes.
-            // But, if both of them _are_ a parameter, or both of them _are not_ a parameter, then they're in the same
-            // scope, and we should report an error. The same applies for data members.
-            if ((isFirstAParameter == isSecondAParameter) && (isFirstADataMember == isSecondADataMember))
+            auto isInSelfContainedScope = [](const ContainedPtr& p)
+            { return dynamic_pointer_cast<DataMember>(p) || dynamic_pointer_cast<Parameter>(p); };
+
+            if (!isInSelfContainedScope(it->second) && !isInSelfContainedScope(namedThing))
```
externl, line 2283: A comment would still be useful here.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Good catch. Totally wrong on my part here.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Fine, changed as suggested.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: I added a comment back.

PR #4279, opened 2025-08-04
Python type-hint minor fixes

pepone:


python/python/Ice/Util.py
```diff
@@ -45,7 +45,7 @@ def initialize(  # type: ignore
     args: list[str] | None = None,
     initData: Ice.InitializationData | None = None,
     configFile: str | None = None,
-    eventLoop: Ice.EventLoopAdapter | None = None,
```
externl, line 48: Note that type hint check are currently disabled for this function entirely. 

```
@overload
def initialize(
    args: list[str], configFile: str | None = None, eventLoop: Ice.EventLoopAdapter | None = None
) -> Ice.Communicator: ...
```

This overload will need to be updated. Would be best to fix the type hint for this function.

python/python/Ice/Util.py
```diff
@@ -17,36 +16,43 @@
...
-def initialize(args: list[str]) -> Ice.Communicator: ...
+def initialize(args: list[str]) -> Communicator: ...
+
 
+@overload
+def initialize(args: list[str], initData: InitializationData) -> Communicator: ...
 
 @overload
-def initialize(args: list[str], initData: Ice.InitializationData) -> Ice.Communicator: ...
+def initialize(args: None = None,  initData: InitializationData = InitializationData()) -> Communicator: ...
```
externl: `= InitializationData()` Is this safe?

pepone: I  think so, because `@overloads` are only use for type-hints. 

Another alternative would be `initData: InitializationData | None = None`

externl: If they're only used for type hints how does it work?

externl (approved):
Looks good. There are some ruff failures.

PR #4284, opened 2025-08-05
Python type hint updates

externl:
This PR starts adding type hints to the Ice tests, first with Ice/adapterDeactivation. It also adds `IcePy.pyi` which has the beginnings for the type hints of IcePy (there's more to do in this file).


python/python/.gitignore
```diff
@@ -4,6 +4,7 @@ IceStorm/*
 IceBox/*
 Glacier2/*
 IcePy.*
+!IcePy.pyi
```
InsertCreativityHere, line 11: Would be nice to have a comment explaining what this file is.

python/python/Ice/ObjectAdapter.py
```diff
@@ -582,7 +582,7 @@ def getPublishedEndpoints(self) -> tuple[IcePy.Endpoint, ...]:
         """
         return self._impl.getPublishedEndpoints()
 
-    def setPublishedEndpoints(self, newEndpoints: tuple[IcePy.Endpoint, ...]) -> None:
+    def setPublishedEndpoints(self, newEndpoints: tuple[IcePy.Endpoint, ...] | list[IcePy.Endpoint]) -> None:
```
pepone, line 585: We should update the doc comment type too.

pepone (approved):

InsertCreativityHere (approved):
Looks good to me!

PR #4285, opened 2025-08-05
Fix python TODO comments

pepone:
A few small fixes for Python TODOs


python/modules/IcePy/Communicator.cpp
```diff
@@ -442,7 +440,7 @@ communicatorWaitForShutdown(CommunicatorObject* self, PyObject* args)
     // Do not call waitForShutdown from the main thread, because it prevents
     // signals (such as keyboard interrupts) from being delivered to Python.
     //
-    if (PyThread_get_thread_ident() == mainThreadId)
+    if (isMainThread())
```
pepone, line 443: I added this helper to check if running in the main thread.

python/python/Ice/ObjectAdapter.py
```diff
@@ -73,16 +73,12 @@ def waitForHold(self) -> None:
         """
         Wait until the object adapter holds requests.
 
-        Calling `hold` initiates holding of requests, and `waitForHold` only returns when holding of requests has been completed.
-        """
-        #
-        # TODO should be part of the documented behavior above. Should we add a timeout parameter with a default value?
```
pepone, line 79: This TODO was bogus, the comment is about the IcePy implementation. We call it in a loop to avoid blocking the main thread.

python/test/Ice/operations/ServerAMD.py
```diff
@@ -9,12 +9,7 @@
 if "--load-slice" in sys.argv:
     TestHelper.loadSlice("Test.ice")
 
-# TODO: This is no longer relevant. All supported Python versions provide async/await keywords.
-# We want to test coroutines but older versions of Python cannot
-# load a source file that uses the async/await keywords, so we
-# have two versions of MyDerivedClassI.
-#
-from TestAMDCoroI import MyDerivedClassI
+from TestAMDI import MyDerivedClassI
```
pepone, line 12: Removed the comment, and renamed TestAMDCoroI to TestAMDI, removed the old TestAMDI which was no longer used.

bernardnormier (approved):
Looks good.

PR #4286, opened 2025-08-05
Improve Bad-Identifier Handling for Modules, Structs, and Operations

InsertCreativityHere:
This PR follows #4223 and #4276, and continues improving the way the Slice parser handles keyword or missing identifiers.

It improves identifier handling for modules, structs, and operations. All that's left after this are classes, interfaces, and exceptions.

It also slightly tweaks how the parser checks for 'changed meaning' errors. Often the logic was double-counting 'redefinition' errors as 'changed meaning' errors, which isn't helpful.


cpp/src/Slice/Grammar.y
```diff
@@ -309,7 +309,7 @@ definition
 // ----------------------------------------------------------------------
 : module_def opt_semicolon
 {
-    assert($1 == nullptr || dynamic_pointer_cast<Module>($1));
```
InsertCreativityHere, line 312: We always create the element now, even if it has a 'bad' identifier.
So the returned element should never be `nullptr`.

cpp/src/Slice/Grammar.y
```diff
@@ -368,35 +368,22 @@ definition
...
 module_def
 // ----------------------------------------------------------------------
-: ICE_MODULE ICE_IDENTIFIER
+: ICE_MODULE definition_name
 {
     currentUnit->setSeenDefinition();
 
     auto ident = dynamic_pointer_cast<StringTok>($2);
     ContainerPtr cont = currentUnit->currentContainer();
     ModulePtr module = cont->createModule(ident->v, false);
```
InsertCreativityHere, line 377: Again, we always create the element, so there's no risk of `module` being nullptr.
This helps simplify the rest of the logic here.

cpp/src/Slice/Grammar.y
```diff
@@ -618,25 +579,10 @@ optional_type_id
 }
 ;
 
-// ----------------------------------------------------------------------
```
InsertCreativityHere, line 621: The new `definition_name` rule handles keywords and identifiers.
It replaces these little side-rules we used to have before.

cpp/src/Slice/Grammar.y
```diff
@@ -700,7 +634,7 @@ class_name
 // ----------------------------------------------------------------------
 class_id
 // ----------------------------------------------------------------------
```
InsertCreativityHere, line 643: `ICE_IDENT_OPEN` is just `<identifier>(`, same for `definition_name_open`, it just means 'something' followed by an opening parenthesis.

cpp/src/Slice/Grammar.y
```diff
@@ -1018,45 +952,15 @@ operation_preamble
...
-        }
+        interface->checkHasChangedMeaning(name, op);
+        currentUnit->pushContainer(op);
         $$ = op;
     }
     else
     {
         $$ = nullptr;
     }
 }
```
InsertCreativityHere, line 970: No longer needed. `definition_name` handles the keyword case.

cpp/src/Slice/Parser.cpp
```diff
@@ -1249,54 +1249,60 @@ Slice::Container::destroyContents()
 ModulePtr
 Slice::Container::createModule(const string& name, bool nestedSyntax)
 {
-    ContainedList matches = unit()->findContents(thisScope() + name);
-    matches.sort(containedCompare); // Modules can occur many times...
-    matches.unique(containedEqual); // ... but we only want one instance of each.
+    ModulePtr q = make_shared<Module>(shared_from_this(), name, nestedSyntax);
```
InsertCreativityHere, line 1252: All of these 'createXXX' methods have been updated to:
1) Only perform name-related checks if a name was provided by the parser.
2) Always create the 'XXX', even if the name is 'bad' for some reason.
3) But, if the identifier was bad, we don't want to add the element into any of `unit`'s lookup tables.

cpp/src/Slice/Parser.cpp
```diff
@@ -1642,7 +1653,6 @@ SequencePtr
 Slice::Container::createSequence(const string& name, const TypePtr& type, MetadataList metadata)
 {
     SequencePtr p = make_shared<Sequence>(shared_from_this(), name, type, std::move(metadata));
-    _contents.push_back(p);
```
InsertCreativityHere, line 1645: I moved these to the bottoms of their respective `createXXX` methods for consistency of placement.
It doesn't meaningfully change anything.

cpp/src/Slice/Parser.cpp
```diff
@@ -2271,19 +2282,31 @@ Slice::Container::checkHasChangedMeaning(const string& name, ContainedPtr namedT
...
-            {
-                unit()->error("'" + firstComponent + "' has changed meaning");
-            }
+        // If the two identifiers resolve to different elements, but those elements are in the same container,
+        // then the 'real' error is that the user redefined an element, not that an element changed meaning.
+        // This error is reported elsewhere in the code, so nothing to do here.
+        if (it->second->container() == namedThing->container())
+        {
+            return;
+        }
```
InsertCreativityHere, line 2300: This is the new addition to this logic.
You'll see that adding it reduced the number of unnecessary `changed meaning` errors in the `errorDetection` tests.

cpp/test/Slice/errorDetection/CaseInsensitive.err
```diff
@@ -24,10 +24,8 @@ CaseInsensitive.ice:146: dictionary 'D' differs only in capitalization from dict
 CaseInsensitive.ice:147: redefinition of module 'm1' as dictionary
 CaseInsensitive.ice:148: dictionary 'M1' differs only in capitalization from module 'm1'
 CaseInsensitive.ice:151: enumeration 'eN1' differs only in capitalization from enumeration 'en1'
-CaseInsensitive.ice:151: 'eN1' has changed meaning
```
InsertCreativityHere, line 27: The real error was that `eN1` and `en1` conflict with one another.
Not that `eN1` changed meaning...

cpp/test/Slice/errorDetection/ChangedMeaning.err
```diff
@@ -8,7 +8,6 @@ ChangedMeaning.ice:79: 'c2' has changed meaning
 ChangedMeaning.ice:96: 'CounterSeq' has changed meaning
 ChangedMeaning.ice:105: redefinition of parameter 'param'
 ChangedMeaning.ice:123: 'E' is not an exception
-ChangedMeaning.ice:123: 'E' has changed meaning
```
InsertCreativityHere, line 11: The real error is that `E` was used as an exception when it is not, not that it changed meaning.

cpp/test/Slice/errorDetection/IdentAsKeyword.err
```diff
@@ -56,14 +56,15 @@ IdentAsKeyword.ice:54: keyword 'idempotent' cannot be used as a name
 IdentAsKeyword.ice:54: enums can only be defined within a module
 IdentAsKeyword.ice:55: enums can only be defined within a module
 IdentAsKeyword.ice:55: enumeration 'IDEMPOTENT' differs only in capitalization from enumeration 'idempotent'
-IdentAsKeyword.ice:55: 'IDEMPOTENT' has changed meaning
 IdentAsKeyword.ice:57: enums can only be defined within a module
 IdentAsKeyword.ice:57: keyword 'long' cannot be used as enumerator
 IdentAsKeyword.ice:57: keyword 'byte' cannot be used as enumerator
 IdentAsKeyword.ice:58: enums can only be defined within a module
 IdentAsKeyword.ice:60: interfaces can only be defined within a module
+IdentAsKeyword.ice:60: keyword 'module' cannot be used as a name
```
InsertCreativityHere, line 64: Here it's actually good to have 2 errors, because they're both true and unrelated.
We didn't see the 2nd error before, because the `createInterface` logic exited early with `nullptr`.
Now that it doesn't exit early, we report all the errors we see.

cpp/src/Slice/Parser.cpp
```diff
@@ -1249,54 +1249,60 @@ Slice::Container::destroyContents()
...
+            bool differsOnlyInCase = matches.front()->name() != name;
+            if (ModulePtr module = dynamic_pointer_cast<Module>(p))
+            {
+                if (differsOnlyInCase) // Modules can be reopened only if they are capitalized correctly.
+                {
+                    ostringstream os;
+                    os << "module '" << name << "' is capitalized inconsistently with its previous name: '"
+                       << module->name() << "'";
+                    unit()->error(os.str());
+                    hasValidIdentifier = false;
```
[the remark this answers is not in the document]
InsertCreativityHere: Indeed, the point of this PR is that we _want_ to create the module, even if it has a bad identifier.

cpp/test/Slice/errorDetection/DerivedRedefinition.err
```diff
@@ -1,5 +1,6 @@
 DerivedRedefinition.ice:9: operation 'ice_ping' is already defined as an operation in a base interface
 DerivedRedefinition.ice:14: operation 'op' is already defined as an operation in a base interface
+DerivedRedefinition.ice:15: redefinition of operation 'op'
 DerivedRedefinition.ice:15: operation 'op' is already defined as an operation in a base interface
```
InsertCreativityHere, line 4: Maybe we could improve it, but getting 2 errors for line 15 is correct, because `op` is doubly-redefined.
One in the same interface and once in a parent interface.

cpp/src/Slice/Grammar.cpp
InsertCreativityHere, line 1: Generated code, probably not worth reading.

cpp/src/Slice/Parser.cpp
```diff
@@ -1249,54 +1249,60 @@ Slice::Container::destroyContents()
...
+        {
+            unit()->addTopLevelModule(unit()->currentFile(), name);
+        }
 
-    for (const auto& p : matches)
-    {
-        bool differsOnlyInCase = matches.front()->name() != name;
-        ModulePtr module = dynamic_pointer_cast<Module>(p);
-        if (module)
+        bool hasValidIdentifier = true;
```
bernardnormier: Should be something like hasConflictingIdentifier

cpp/src/Slice/Parser.cpp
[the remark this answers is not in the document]
bernardnormier: I would add a `break` after the (first) error.

cpp/src/Slice/Parser.cpp
```diff
@@ -1249,54 +1249,60 @@ Slice::Container::destroyContents()
...
+            // If this module has a valid identifier and doesn't conflict with another definition,
+            // add it to the unit's contentMap so it can be looked up later.
+            unit()->addContent(q);
         }
-    }
 
-    if (!checkIdentifier(name))
-    {
-        return nullptr;
+        checkIdentifier(name);
```
bernardnormier: Add a comment - why do we check the identifier so late?

cpp/src/Slice/Parser.cpp
```diff
@@ -3106,53 +3127,57 @@ Slice::InterfaceDef::createOperation(
...
-    // Check whether enclosing interface has the same name.
-    if (name == this->name())
+    if (!name.empty())
     {
-        ostringstream os;
-        os << "interface name '" << name << "' cannot be used as operation name";
-        unit()->error(os.str());
-        return nullptr;
-    }
+        bool hasValidIdentifier = !doesNameConflict(name, "operation", _contents);
```
bernardnormier: Need to rename variable: it's not about valid identifier (as checked by checkIdentifier, that this function doesn't call as far as I can tell).
It's only about a name that is not empty and non-conflicting.

cpp/src/Slice/Parser.cpp
```diff
@@ -1249,54 +1249,60 @@ Slice::Container::destroyContents()
...
+        if (hasValidIdentifier)
         {
-            ostringstream os;
-            os << "module '" << name << "' differs only in capitalization from " << matches.front()->kindOf()
-               << " name '" << matches.front()->name() << "'";
-            unit()->error(os.str());
-            return nullptr;
+            // If this module has a valid identifier and doesn't conflict with another definition,
+            // add it to the unit's contentMap so it can be looked up later.
+            unit()->addContent(q);
```
externl, line 1303: Should we be doing this after `checkIdentifier` like before?

InsertCreativityHere: The order really doesn't matter.
I updated the name of this function to `reportIllegalSuffixOrUnderscore`, which is much clearer than `checkIdentifier`.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: I renamed the variable to `hasConflictingIdentifier` per your other suggestion.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: I updated the name of this function to `reportIllegalSuffixOrUnderscore`.
1) This is much more clear than `checkIdentifier`
2) It makes it clear that (going forward) we don't care about the return value.
After I finish updating classes, exceptions, and interfaces like I did in this PR, I'll remove the return value all together.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: Yes, that's a more clear name.
I updated them all as suggested.

cpp/src/Slice/Parser.cpp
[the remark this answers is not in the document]
InsertCreativityHere: I added a break to avoid emitting duplicate errors here.

bernardnormier (approved):

PR #4288, opened 2025-08-05
Move custom class loader doc (Java)

bernardnormier:
Moved from
https://doc.zeroc.com/ice/3.7/language-mappings/java-mapping/custom-class-loaders

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InitializationData.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InitializationData.java
@@ -41,7 +41,7 @@ public final class InitializationData implements Cloneable {
     /** The logger for the communicator. */
     public Logger logger;
 
-    /** The communicator observer used by the Ice run-time. */
+    /** The communicator observer used by the communicator. */
     public CommunicatorObserver observer;
 
     /** threadStart is called whenever the communicator starts a new thread. */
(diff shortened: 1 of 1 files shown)


InsertCreativityHere (approved):
Looks good.
But I still think that we should mention the necessity of class-loaders in the real docs too.
Otherwise it will be too easy for users to miss.

Maybe we could link to this paragraph under the `java:identifier`, etc. metadata?

ie. If using X means a user will needs application class loaders, we should mention it in the documentation for X.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InitializationData.java
```diff
@@ -52,7 +52,22 @@ public InitializationData clone() {
      */
     public Runnable threadStop;
 
-    /** The custom class loader for the communicator. */
+    /** The custom class loader for the communicator.
+     * An application can supply a custom class loader that the Ice communicator will use when it:
+     * <ul>
+     *  <li>unmarshals class instances and user exceptions</li>
+     *  <li>loads Ice plug-ins</li>
+     *  <li>loads IceSSL certificate verifiers and password callbacks</li>
```
pepone: We removed IceSSL certificate verifiers and password callbacks.

pepone (approved):

bernardnormier:
> aragraph under the `java:identifier`, etc. metadata?

We should indeed document Slice loaders somewhere in the Ice Manual. But not these more advanced class loaders.

PR #4289, opened 2025-08-06
Enable strict mode for TypeScript tests

pepone:


externl (approved):

cpp/src/slice2js/Gen.cpp
```diff
@@ -2414,6 +2414,17 @@ Slice::Gen::TypeScriptVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
          << ", "
          << "facet?: string): " << prxName << ";";
 
+    _out << sp;
```
bernardnormier, line 2406: I don't understand why we need this third overload.

js/src/Ice/ImplicitContext.d.ts
```diff
@@ -29,7 +29,7 @@ declare module "@zeroc/ice" {
              *
              * @param newContext - The new context.
              */
-            setContext(newContext: Context): void;
+            setContext(newContext: Context | null): void;
```
bernardnormier: I would not accept null for newContext in TS.

js/test/Ice/binding/Client.ts
```diff
@@ -53,25 +54,28 @@ export class Client extends TestHelper {
         out.write("testing binding with single endpoint... ");
         {
             const adapter = await com.createObjectAdapter("Adapter", "default");
+            const test1 = await adapter!.getTestIntf();
```
bernardnormier: For my understanding: when I get a proxy (like adapter) in TS, and I know it's not null, I have to use `proxy!` every time I use it? This sounds sub-optimal.

There is no way to assert once that it's not null and not use `!` after, like in C#?

js/test/Ice/facets/Client.ts
```diff
@@ -33,26 +33,26 @@ export class Client extends TestHelper {
         out.writeLine("ok");
 
         out.write("testing checked cast... ");
-        prx = await Ice.ObjectPrx.checkedCast(db);
+        prx = (await Ice.ObjectPrx.checkedCast(db)) as Ice.ObjectPrx;
```
bernardnormier: The checkedCast followed by `as` does not look like a good model.

js/test/Ice/inheritance/Client.ts
```diff
@@ -14,10 +14,10 @@ export class Client extends TestHelper {
         const initial = new Test.InitialPrx(communicator, `initial:${this.getTestEndpoint()}`);
 
         out.write("getting proxies for interface hierarchy... ");
-        const ia = await initial.iaop();
-        const ib1 = await initial.ib1op();
-        const ib2 = await initial.ib2op();
-        const ic = await initial.icop();
+        const ia = (await initial.iaop()) as Test.MA.IAPrx;
```
bernardnormier: Would `!` have the same effect as `as xxx` here?

pepone: yes, both are not-op in the resulting JavaScript.

js/test/Ice/binding/Client.ts
pepone: With `as FooPrx` the type is understood as non nullable and you don't need to use `!` in all versions.

Without strict mode, the compiler only complain on the first occurrence I belive.

cpp/src/slice2js/Gen.cpp
pepone: The compiler doesn't allow use the other two when It doesn't know if the type is null or not. If it nows the proxy is null it picks the overload accepting null and it knows it will return null, if it knows it is not null it will use the one using a null proxy. But when it doesn't know it will refuse to use any of them, the third overload works in this case.

bernardnormier: Maybe we should drop the `null` overload them - what's the point of keeping it?

js/test/Ice/binding/Client.ts
bernardnormier: As commented below, I don't find the style `[un]checkedCast` followed by `as XPrx` good. It appears as two casts in a row. 

I'd rather use `!` to state "I know this nullable XPrx is never null", assuming it's doable, as in:
```js
const adapter = (await com.createObjectAdapter("Adapter", "default"))!;
```

pepone: I think we can add a helper here:

```
function testNonNull<T>(value: T): asserts value is NonNullable<T> {
    if (value == null) throw new Error("Value is null or undefined");
}
```

then rather that our `test(adapter != null)` use `testNonNull(adapter)` and the compiler would understand it is non null in subsequent calls.

cpp/src/slice2js/Gen.cpp
pepone: Removed the null overload.

js/src/Ice/AsyncResult.js
```diff
@@ -40,7 +40,7 @@ export class AsyncResult extends Promise {
     }
 
     markSent(done) {
-        DEV: console.assert((this._state & AsyncResult.Done) === 0);
+        console.assert((this._state & AsyncResult.Done) === 0);
```
pepone, line 43: I removed this `DEV: ` labels, and use console.assert directly. Bundlers can be configure to ignore them, no need for a custom label.

js/src/Ice/ObjectPrxExtensions.js
```diff
@@ -405,7 +405,7 @@ ObjectPrx._invoke = function (p, name, mode, fmt, ctx, marshalFn, unmarshalFn, u
     });
 
     try {
-        r.prepare(name, mode, ctx);
+        r.prepare(mode, ctx);
```
pepone, line 408: Removed `name` because the implementation never uses it. This is the only call to `prepare`.

js/src/Ice/OutgoingConnectionFactory.js
```diff
@@ -471,8 +471,8 @@ class ConnectCallback {
         this._factory.finishGetConnection(this._endpoints, this._current, connection, this);
     }
 
-    connectionStartFailed(connection, ex) {
-        DEV: console.assert(this._current !== null);
+    connectionStartFailed(ex) {
```
pepone, line 474: `connection` was never used.

js/src/Ice/TcpTransceiver.js
```diff
@@ -235,7 +235,7 @@ class TcpTransceiver {
         );
     }
 
-    setBufferSize(rcvSize, sndSize) {
+    setBufferSize(_, sndSize) {
```
pepone, line 238: None of the JavaScript transport use the `rcvSize` maybe we should rename this method to setSendBufferSize and have a single parameter.

js/test/Common/TestHelper.d.ts
```diff
@@ -24,9 +24,10 @@ export class TestHelper {
...
     setControllerHelper(controllerHelper: ControllerHelper): void;
     serverReady(): void;
-    static test(value: boolean, ex?: Error): void;
 
     static isSafari(): boolean;
     static isBrowser(): boolean;
     static isWorker(): boolean;
 }
+
+export function test(value: boolean, ex?: Error): asserts value;
```
pepone, line 33: With this `asserts value` the TypeScript compiler understand value is true.

For example
```
let foo = FooPrx.checkedCast(prx);
test(foo !== null);
await foo.op(); // No need for .!
```

js/test/Glacier2/router/Client.ts
```diff
@@ -2,9 +2,7 @@
 
 import { Ice, Glacier2 } from "@zeroc/ice";
 import { Test } from "./Callback.js";
-import { TestHelper } from "../../Common/TestHelper.js";
-
-const test = TestHelper.test;
+import { TestHelper, test } from "../../Common/TestHelper.js";
```
pepone, line 5: It is clear to import the function directly.

js/test/Glacier2/router/Client.ts
```diff
@@ -17,7 +15,7 @@ class CallbackReceiverI extends Test.CallbackReceiver {
         this._p = new Ice.Promise();
     }
 
-    callback(current: Ice.Current) {
+    callback(_: Ice.Current) {
```
pepone, line 18: linter and compiler warn that `current` is unused.

We can:
- use `_` prefix to indicate that is intentional.
- Remove the parameter, even if is part of the paren signature: `callback()`.
- Disable the warnings for unused parameters.

bernardnormier: I prefer the `_` solution.

PR #4293, opened 2025-08-06
Make javabean field private

bernardnormier:
This PR updates the JavaBean-style mapping field to be private, just like for optional fields.


InsertCreativityHere (approved):
Looks good to me!

PR #4295, opened 2025-08-07
Add more type hints to Python tests

externl:
This PR adds more type hints to the Python tests. Doing in batches so the PRs aren't too large.


.github/workflows/python.yml
```diff
@@ -54,4 +54,16 @@ jobs:
           pyright python/python
           pyright python/test/Glacier2
           pyright python/test/Ice/adapterDeactivation
+          pyright python/test/Ice/admin
           pyright python/test/Ice/ami
+          pyright python/test/Ice/asyncio
+          pyright python/test/Ice/blobject
+          pyright python/test/Ice/current
+          # pyright python/test/Ice/custom
```
externl, line 62: I started the commented out ones but they require additional fixes in the compiler and IcePy.pyi.

python/python/Ice/Proxy.py
```diff
@@ -27,6 +27,7 @@ def proxyIdentityCompare(lhs: IcePy.ObjectPrx, rhs: IcePy.ObjectPrx) -> int:
     elif lhs and not rhs:
         return 1
     else:
+        assert lhs is not None and rhs is not None
```
externl, line 30: the None cases are handled by the above `if` and `elif`s

python/python/Ice/Util.py
```diff
@@ -36,6 +36,15 @@ def initialize(
 ) -> Communicator: ...
 
 
+@overload
+def initialize(
```
pepone: Seems we can remove the overload above this now.

python/python/IcePy.pyi
```diff
@@ -266,8 +266,12 @@ class ObjectPrx:
     def ice_getRouter(self) -> Ice.RouterPrx | None: ...
     def ice_identity(self, newIdentity: Ice.Identity) -> Self: ...
     def ice_invocationTimeout(self, timeout: int) -> Self: ...
-    # def ice_invoke(self, *args, **kwargs): ...
-    # def ice_invokeAsync(self, *args, **kwargs): ...
+    def ice_invoke(
+        self, operation: str, mode: Ice.OperationMode, inParams: bytes, ctx: dict[str, str]
```
pepone: Shouldn't context be optional here?

python/test/Ice/blobject/RouterI.py
```diff
@@ -124,21 +129,24 @@ def destroy(self):
...
 
 class ServantLocatorI(Ice.ServantLocator):
-    def __init__(self, blobject):
+    def __init__(self, blobject: Ice.Blobject):
         self._blobject = blobject
 
-    def locate(self, current: Ice.Current):
-        return self._blobject  # and the cookie
+    @override
+    def locate(self, current: Ice.Current) -> tuple[Ice.Object | None, object]:
```
pepone: Doesn't this need `object | None` for the cookie?

python/test/Ice/blobject/RouterI.py
```diff
@@ -124,21 +129,24 @@ def destroy(self):
...
+        return self._blobject, None  # and the cookie
 
-    def finished(self, current, object, cookie):
+    @override
+    def finished(self, current: Ice.Current, servant: Ice.Object, cookie: object):
```
pepone: Cookie can also be None.

python/test/Ice/blobject/RouterI.py
```diff
@@ -150,18 +158,20 @@ def __init__(self, communicator, sync):
         communicator.setDefaultRouter(proxy)
         self._adapter.activate()
 
-    def useSync(self, sync):
-        self._locator.useSync(sync)
-
+    @override
     def getClientProxy(self, current: Ice.Current):
```
pepone: missing return type.

python/test/Ice/exceptions/AllTests.py
```diff
@@ -2,219 +2,34 @@
...
-
 class EmptyI(Test.Empty):
     pass
 
 
 class ServantLocatorI(Ice.ServantLocator):
-    def locate(self, current: Ice.Current):
-        return None
+    @override
+    def locate(self, current: Ice.Current) -> tuple[Ice.Object | None, object]:
```
pepone: object | None for the cookie

python/test/Ice/exceptions/AllTests.py
```diff
@@ -2,219 +2,34 @@
...
+    def finished(self, current: Ice.Current, servant: Ice.Object, cookie: object):
         pass
 
-    def deactivate(self, category):
+    @override
+    def deactivate(self, category: str):
         pass
 
 
-class CallbackBase:
```
pepone, line 32: Why was this removed?

python/test/Ice/exceptions/ServerAMD.py
```diff
@@ -178,7 +29,7 @@ def run(self, args: list[str]):
             adapter = communicator.createObjectAdapter("TestAdapter")
             adapter2 = communicator.createObjectAdapter("TestAdapter2")
             adapter3 = communicator.createObjectAdapter("TestAdapter3")
-            object = ThrowerI()
+            object = TestI.ThrowerI()
```
pepone: We were testing both AMD and non AMD, how does this work now?

python/test/Ice/location/AllTests.py
```diff
@@ -13,15 +15,11 @@ def sayHello(self, current: Ice.Current):
...
-    if not b:
-        raise RuntimeError("test assertion failed")
-
-
-def allTests(helper, communicator):
+def allTests(helper: TestHelper, communicator: Ice.Communicator):
     manager = Test.ServerManagerPrx(communicator, f"ServerManager:{helper.getTestEndpoint()}")
     locator = communicator.getDefaultLocator()
-    test(manager)
+    assert manager is not None
```
pepone: This assert is not required. We just create manager with its constructor.

python/test/Ice/location/AllTests.py
```diff
@@ -170,7 +177,9 @@ def allTests(helper, communicator):
         test(False)
     except Ice.NotRegisteredException as ex:
         test(ex.kindOfObject == "object")
-        test(ex.id == "unknown/unknown")
+        assert ex.id is not None
+        test(str(ex.id) == "unknown/unknown")
```
pepone: Why are the `None` check, and the `str` call required here? I think we are missing the type-hint in `NotRegisteredException`

This should be just `test(ex.id == "unknown/unknown")` as before.

python/test/Ice/location/AllTests.py
```diff
@@ -181,12 +190,13 @@ def allTests(helper, communicator):
         test(False)
     except Ice.NotRegisteredException as ex:
         test(ex.kindOfObject == "object adapter")
-        test(ex.id == "TestAdapterUnknown")
+        test(str(ex.id) == "TestAdapterUnknown")
```
pepone: ditto

python/test/Ice/location/Server.py
```diff
@@ -19,53 +20,61 @@ def __init__(self):
...
+    def addObject(self, obj: Ice.ObjectPrx | None, current: Ice.Current):
+        assert obj is not None
         self._addObject(obj)
 
-    def _addObject(self, obj):
+    def _addObject(self, obj: Ice.ObjectPrx):
         self._objects[obj.ice_getIdentity()] = obj
 
-    def getAdapter(self, adapter):
+    def getAdapter(self, adapter: str):
```
pepone: missing return type

python/test/Ice/exceptions/AllTests.py
externl, line 32: All these classes appear to be unused.

python/test/Ice/exceptions/ServerAMD.py
```diff
@@ -4,162 +4,13 @@
...
 
-if "--load-slice" in sys.argv:
-    TestHelper.loadSlice("Test.ice")
-
-from generated.test.Ice.exceptions import Test
-
 import Ice
 
-
-class ThrowerI(Test.Thrower):
```
externl, line 17: I just reused the non-and ThrowerI, it was the exact same impl. 

EDIT: I was mistaken. Will fix.

python/test/Ice/exceptions/ServerAMD.py
externl: I got mixed up when I updated this and thought that they had the same impl. Will fix.

python/test/Ice/exceptions/AllTests.py
externl: was unused from what I could tell. I removed it and nothing broke.

python/test/Ice/location/AllTests.py
externl: `ex.id` is `Identity | None`.

externl: Fixed to be `str`

python/test/Ice/blobject/test.py
```diff
@@ -3,6 +3,6 @@
 # This test doesn't support running with IceSSL, the Router object in the client process uses
 # the client certificate and fails with "unsupported certificate purpose"
 
-from Util import TestSuite
+from Util import TestSuite  # type: ignore
```
pepone, line 6: Should this have a TODO?

python/test/Ice/exceptions/TestAMDI.py
```diff
@@ -0,0 +1,183 @@
...
+    @override
+    def throwAfterException(self, current: Ice.Current):
+        # Cannot be implemented with Futures
+        f = Ice.Future()
+        f.set_exception(Test.A())
+        return f
+
+    @override
+    def throwMarshalException(self, current: Ice.Current) -> Awaitable[tuple[int, int]]:
+        # We return the wrong type on purpose to test that we throw MarshalException.
```
pepone, line 167: Don't we need to disable type checks here to be able to return the wrong type?

pepone (approved):

python/test/Ice/exceptions/TestAMDI.py
externl: Pyright doesn't care. I think because of the signature of the Ice.Futre func. Something we should look into improving.

python/test/Ice/blobject/test.py
externl: Probably not, I doubt we'll want to add type hinting to the Util code.

bernardnormier (approved):

PR #4296, opened 2025-08-07
Cleanup NotRegisteredException doc comments

pepone:


cpp/include/Ice/LocalExceptions.h
```diff
@@ -683,25 +683,23 @@ namespace Ice
         [[nodiscard]] const char* ice_id() const noexcept final;
     };
 
-    /// The exception that is thrown when you attempt to find or deregister something that is not registered with the
-    /// Ice runtime or Ice locator.
+    /// An attempt was made to find or deregister something that is not registered with the Ice run time or Ice
```
bernardnormier: The latest style for runtime is runtime without space or dash.

bernardnormier: I would actually simplify here and write:
```
An attempt was made to find or deregister something that is not registered with Ice.
```

Where "with Ice" can mean an Ice communicator, an Ice locator implementation, an object adapter, etc. Various Ice-provided containers.

python/python/Ice/LocalExceptions.py
```diff
@@ -419,49 +419,34 @@ class NoEndpointException(LocalException):
...
-    This exception is raised if an attempt is made to remove a servant, servant locator, facet, plug-in, or object
-    adapter that is not currently registered. It's also raised if the Ice locator can't find an object or object adapter
-    when resolving an indirect proxy or when an object adapter is activated.
     """
 
-    def __init__(self, kindOfObject: str, id: Identity | None, msg: str):
+    def __init__(self, kindOfObject: str, id: str, msg: str):
         LocalException.__init__(self, msg)
         self.__kindOfObject = kindOfObject
         self.__id = id
```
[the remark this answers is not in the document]
bernardnormier: The identity was a bug. It should be a string.

bernardnormier (approved):

PR #4303, opened 2025-08-11
Add type hinting to Python Ice/operations

externl:
Additionally this PR updates the type using for marshaling dictionaries to be `Mapping`. This allows for more flexibility in the types of values we can accept through type hinting.

Note that Marshaled Result is not yet done.  I plan to do this after updating the remaining tests.


python/test/Ice/operations/BatchOneways.py
```diff
@@ -39,27 +37,27 @@ def enqueue(self, request, count, size):
...
+def batchOneways(p: Test.MyClassPrx) -> None:
+    bs1 = bytes([0 for _ in range(0, 10 * 1024)])
 
     p.opByteSOneway(bs1)
     batch = Test.MyClassPrx.uncheckedCast(p.ice_batchOneway())
 
     batch.ice_flushBatchRequests()  # Empty flush
     if batch.ice_getConnection():
-        batch.ice_getConnection().flushBatchRequests(Ice.CompressBatch.BasedOnProxy)
+        cast(Ice.Connection, batch.ice_getConnection()).flushBatchRequests(Ice.CompressBatch.BasedOnProxy)
```
pepone: Why do we need a cast here? can't we just assert `ice_getConnection` return a connection ? for example `is not None`?

python/test/Ice/operations/BatchOnewaysFuture.py
```diff
@@ -55,28 +53,34 @@ def batchOneways(p):
...
+        cast(Ice.Connection, batch1.ice_getConnection()).close().result()
         batch1.ice_pingAsync()
         batch2.ice_pingAsync()
 
         batch1.ice_getConnection()
         batch2.ice_getConnection()
 
         batch1.ice_pingAsync()
-        batch1.ice_getConnection().close().result()
+        cast(Ice.Connection, batch1.ice_getConnection()).close().result()
```
pepone, line 65: Not related to the changes in this PR:

Any idea what are we testing here? I looked at other versions of this but there is no comments. My guess this is related to the close connection not resulting in exceptions in subsequent calls.

Looking at the sync version is a bit different:

```
        batch1.ice_ping()
        batch1.ice_getConnection().close().result()

        batch1.ice_ping()
        batch2.ice_ping()
```

There is one ice_ping on each proxy, instead of two. With the async version here we don't now if first ice_pingAsync failed or not. We should probably just call `.result` like above.

pepone (approved):

python/test/Ice/operations/BatchOneways.py
externl: We would need to assign the result of the operation to a variable to do that. Calling the function twice doesn't guarantee the same result.

python/test/Ice/operations/BatchOnewaysFuture.py
externl: Best to open an issue so we can fix everywhere.

python/test/Ice/operations/BatchOneways.py
bernardnormier: I am not following. @pepone was asking why:
```py
batch.ice_getConnection().flushBatchRequests(Ice.CompressBatch.BasedOnProxy)
```

doesn't work.

externl: Calling `batch.ice_getConnection()` which returns `Ice.Connection | None` doesn't guarantee the next time you call it it won't be `None` (from Python's viewpoint). So we have to either assign the return value and assert or do the cast.

pepone: We can do:

```
connection = batch.ice_getConnection()
assert connection is not None
connection.flushBatchRequests(Ice.CompressBatch.BasedOnProxy)
```

externl: Yes

PR #4304, opened 2025-08-11
More Python test type hinting

externl:


python/test/Ice/timeout/AllTests.py
```diff
@@ -25,10 +21,10 @@ def connect(prx):
         except Ice.ConnectTimeoutException:
             # Can sporadically occur with slow machines
             pass
-    return prx.ice_getConnection()  # Establish connection
+    assert prx.ice_getConnection()  # Establish connection
```
pepone: Does Python always execute code in asserts?

pepone (approved):

python/test/Ice/timeout/AllTests.py
externl: Ah, good catch. They're not executed if you use `-O`. Will fix.

PR #4309, opened 2025-08-12
Fix for DataStorm session retries

### LINKED ISSUE #4231: DataStorm/reliability failure
https://github.com/zeroc-ice/ice/actions/runs/16464928429/job/46539916125

[test-logs-cpp-win32-release-windows-2022.zip](https://github.com/user-attachments/files/21685276/test-logs-cpp-win32-release-windows-2022.zip)

pepone:
Fix #4231

### DIFF
--- a/cpp/src/DataStorm/NodeI.cpp
+++ b/cpp/src/DataStorm/NodeI.cpp
@@ -219,7 +219,8 @@ NodeI::createSessionAsync(
                     {
                         Trace out(traceLevels->logger, traceLevels->sessionCat);
                         out << "node '" << current.id << "' is ignoring '" << current.operation << "' request from '"
-                            << subscriber << "' because session '" << session->getId() << "' is already connected to: '" << session->getSession() << "'";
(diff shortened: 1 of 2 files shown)


cpp/src/DataStorm/NodeI.cpp
```diff
@@ -458,6 +458,24 @@ NodeI::retrySubscriberSessionCreation(
...
 {
+    try
+    {
+        rethrow_exception(ex);
+    }
+    catch (const SessionCreationException& sessionCreationException)
+    {
+        if (sessionCreationException.error == SessionCreationError::AlreadyConnected && session->checkSession())
+        {
+            // A concurrent session attempt from the peer succeeded, no need to retry.
```
pepone, line 471: In this case both peers are seeing the session as connected.

- This node Session::checkSession returns true, meaning it is already connected.
- The peer node just throw `SessionCreationException` with `AlreadyConnected` error code.

There is no need to keep retrying, and if we keep the retries will keep failing with already connected and the session will be destroyed when the retry limit is hit.

bernardnormier (approved):

PR #4313, opened 2025-08-13
Add type hints to Ice/slicing

externl:
The only thing I didn't do was SlicedData and SliceInfo as there were already lots of updates. I'll do that in a separate PR. 

Pright complains about the function complexity of the huge test function in AllTests.py so I broke it up into smaller functions. Unfortunately this makes it harder to review.

### DIFF
--- a/python/test/Ice/slicing/objects/Server.py
+++ b/python/test/Ice/slicing/objects/Server.py
@@ -2,8 +2,8 @@
 
 # Copyright (c) ZeroC, Inc.
 
-from collections.abc import Awaitable, Mapping
 import sys
+from collections.abc import Awaitable, Mapping
 from typing import override
 
 from TestHelper import TestHelper, test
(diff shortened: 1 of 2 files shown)


python/test/Ice/slicing/objects/AllTests.py
```diff
@@ -1808,15 +1866,27 @@ def allTests(helper, communicator):
         #
         test(PreservedI.counter == 0)
         p = t.PBSUnknown2AsPreservedWithGraph()
-        test(p is not None)
+        assert p is not None
         test(PreservedI.counter == 1)
         t.checkPBSUnknown2WithGraph(p)
-        p._ice_slicedData = None  # Break the cycle.
+        p._ice_slicedData = None  # Break the cycle. # pyright: ignore
```
pepone, line 1872: why do we need to manually break cycles in Python?

pepone (approved):

python/test/Ice/slicing/objects/AllTests.py
externl: I was wondering this myself. Probably just a copy/paste from C++.

PR #4321, opened 2025-08-14
Add type hints to Python Ice/info

externl:
This PR adds type hints to Python Ice/info and the required types in IcePy.pyi. It also removes some untested code in the various TestI implementations.


pepone (approved):
Looks good!

PR #4324, opened 2025-08-14
Python type hinting for Ice/objects and Ice/optional

externl:
This PR also removes support for the marshaled-result metadata directive.

### DIFF
--- a/python/test/Ice/operations/TestI.py
+++ b/python/test/Ice/operations/TestI.py
@@ -510,25 +510,25 @@ class MyDerivedClassI(Test.MyDerivedClass):
     def opWStringLiterals(self, current: Ice.Current) -> Sequence[str]:
         return self.opStringLiterals(current)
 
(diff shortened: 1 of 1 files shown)


python/test/Ice/operations/TestI.py
```diff
@@ -510,26 +510,25 @@ def opStringLiterals(self, current: Ice.Current) -> Sequence[str]:
     def opWStringLiterals(self, current: Ice.Current) -> Sequence[str]:
         return self.opStringLiterals(current)
 
-    # TODO: Marshaled result
     def opMStruct1(self, current: Ice.Current):  # pyright: ignore
-        return Test.MyClass.OpMStruct1MarshaledResult(Test.Structure(), current)
+        return Test.Structure()
```
bernardnormier, line 514: You plan to fix the pyright ignore in a follow-up PR?

externl: I should remove that now!

PR #4327, opened 2025-08-18
Fixes for PHP registered communicators

### LINKED ISSUE #4326: Mutex mismatch in Ice for PHP registered communicator code
Ice for PHP has a `_registeredCommunicatorsMutex` that protects the `_registeredCommunicators` map from concurrent access - not by the single threaded PHP code but from:
 - the reaper thread
 - the PHP code

since they can execute concurrently. 

Naturally the reaper thread should lock this mutex but does not. It locks instead an unrelated mutex, `_reapFinishedMutex`.

pepone:
Fix #4326

### DIFF
--- a/php/src/Communicator.cpp
+++ b/php/src/Communicator.cpp
@@ -226,7 +226,10 @@ ZEND_METHOD(Ice_Communicator, destroy)
             _this->ac->ids.clear();
 
             // Cancel the reap task if it is still scheduled.
-            _timer->cancel(_this->ac->reapTask);
+            if (_this->ac->reapTask)
+            {
+                _timer->cancel(_this->ac->reapTask);
+            }
         }
 
         c->destroy();


php/src/Communicator.cpp
```diff
@@ -118,7 +110,7 @@ namespace
         RegisteredCommunicatorMap::iterator p = _registeredCommunicators.begin();
         while (p != _registeredCommunicators.end())
         {
-            if (p->second->lastAccess + std::chrono::hours(p->second->expires) <= now)
+            if (p->second->expires > 0 && p->second->lastAccess + std::chrono::hours(p->second->expires) <= now)
```
pepone: expires 0 means the communicator doesn't expire as per the docs. The lastAccess field is only initialized when expires > 0, I think the intent is clear this way.

php/src/Communicator.cpp
```diff
@@ -1042,28 +1032,27 @@ ZEND_FUNCTION(Ice_register)
...
+        // Start the reap thread if necessary, the reap thread reaps communicator every 5 minutes.
         if (!_reapThread.joinable())
         {
             _reapThread = std::thread(
                 [&]
                 {
-                    std::unique_lock lock(_reapFinishedMutex);
-                    _reapCond.wait_for(lock, std::chrono::minutes(5), [&] { return _reapingFinished; });
-                    if (_reapingFinished)
+                    while (true)
```
pepone: We need the loop to reap every 5 minutes until `_reapingFinished` is set to true.

php/src/Communicator.cpp
```diff
@@ -1510,22 +1496,6 @@ IcePHP::communicatorInit(void)
         {
             return false;
         }
-
-        if (INI_BOOL(const_cast<char*>("ice.hide_profiles")))
```
pepone, line 1514: I don't see the point of keeping support for this, seems a misfeature.

php/src/Communicator.cpp
```diff
@@ -1042,28 +1032,27 @@ ZEND_FUNCTION(Ice_register)
...
                 [&]
                 {
-                    std::unique_lock lock(_reapFinishedMutex);
-                    _reapCond.wait_for(lock, std::chrono::minutes(5), [&] { return _reapingFinished; });
-                    if (_reapingFinished)
+                    while (true)
                     {
-                        return;
+                        std::unique_lock lock(_registeredCommunicatorsMutex);
+                        _reapCond.wait_for(lock, std::chrono::minutes(5), [&] { return _reapingFinished; });
```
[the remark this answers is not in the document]
pepone: `_reapingFinished` is set to true when the Ice for PHP module is being shutdown, if there is a spurious wakeup at this point we don't want to keep waiting. Otherwise we always until the wait_for times out, then reap and loop over.

php/src/Communicator.cpp
```diff
@@ -92,23 +89,18 @@ namespace
...
-    // std::mutex constructor is constexpr so it is statically initialized
-    // _registeredCommunicatorsMutex protects _registeredCommunicators and _reapThread
-    // _reapFinishedMutex protects _reapingFinished
+    // Protects _registeredCommunicators, _reapThread, and _reapingFinished
     std::mutex _registeredCommunicatorsMutex;
-    std::mutex _reapFinishedMutex;
     std::condition_variable _reapCond;
     std::thread _reapThread;
-    bool _reapingFinished = true;
+    bool _reapingFinished{false};
```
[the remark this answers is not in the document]
pepone: _reapingFinished is used in the spurious wakeup condition to decide if we should keep waiting or not. Starts as false because we have not finished reaping.

php/src/Communicator.cpp
bernardnormier: Should be minutes (not from this PR, but still a bug).

php/src/Communicator.cpp
```diff
@@ -1042,28 +1032,27 @@ ZEND_FUNCTION(Ice_register)
...
+        // Update the expiration time. If a communicator is registered with multiple IDs, we always use the most
+        // recent expiration setting.
         info->ac->expires = static_cast<int>(expires);
         info->ac->lastAccess = std::chrono::steady_clock::now();
 
-        //
-        // Start the timer if necessary. Reap expired communicators every five minutes.
-        //
+        // Start the reap thread if necessary, the reap thread reaps communicator every 5 minutes.
         if (!_reapThread.joinable())
```
bernardnormier: It's not clear to me why we don't use IceInternal::Timer here. Then no need for this condition variable logic.

php/src/Communicator.cpp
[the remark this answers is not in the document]
bernardnormier: While correct, I find this code hard to follow (like Copilot). If we don't simply replace all this code by a Timer, please add a comment.

externl (approved):

php/src/Communicator.cpp
bernardnormier: The 5 minutes loop is pretty odd, since it means we only get 5 minutes granularity. Makes testing pretty tough.

If we just used a Timer, we could simply use the provided "expires" as the period.

pepone: The 3.7 code uses a repeated timer task with a fixed 5 minutes period too, what should we use the timeout of the first registered communicator?

pepone:
Reworked this to use a Timer, and to accept expires that are a fraction of a minute using a double instead of an int.

php/src/Communicator.cpp
```diff
@@ -92,43 +95,57 @@ namespace
...
+            if (activeCommunicator)
             {
-                ++p;
+                activeCommunicator->communicator->destroy();
             }
         }
-    }
+
+    private:
+        ActiveCommunicatorPtr _activeCommunicator;
```
bernardnormier: I would make this field `const`.

php/src/Communicator.cpp
```diff
@@ -200,13 +217,21 @@ ZEND_METHOD(Ice_Communicator, destroy)
...
         {
             lock_guard lock(_registeredCommunicatorsMutex);
-            for (vector<string>::iterator p = _this->ac->ids.begin(); p != _this->ac->ids.end(); ++p)
+            for (const auto& id : _this->ac->ids)
             {
-                _registeredCommunicators.erase(*p);
+                _registeredCommunicators.erase(id);
             }
             _this->ac->ids.clear();
+            reapTask = _this->ac->reapTask;
```
bernardnormier: You could cancel within the mutex lock. The `cancel` is fast and doesn't execute any callback.

php/src/Communicator.cpp
```diff
@@ -1040,32 +1063,21 @@ ZEND_FUNCTION(Ice_register)
...
-                    _reapCond.wait_for(lock, std::chrono::minutes(5), [&] { return _reapingFinished; });
-                    if (_reapingFinished)
-                    {
-                        return;
-                    }
-                    reapRegisteredCommunicators();
-                });
-        }
+        info->ac->reapTask = make_shared<ReapCommunicatorTimerTask>(info->ac);
+        _timer->scheduleRepeated(info->ac->reapTask, info->ac->expires);
```
bernardnormier: I think the period should be "expires / 2", so that the actual expiration occurs between 1 and 1.5 expires, not between 1 and 2 expires.

php/src/Communicator.cpp
```diff
@@ -1128,7 +1140,7 @@ ZEND_FUNCTION(Ice_find)
         RETURN_NULL();
     }
 
-    if (p->second->expires > 0)
+    if (p->second->expires > std::chrono::milliseconds::zero())
```
bernardnormier: I would find something like:
```cpp
// Set last access for the reap task.
if (p->second->reapTask)
{
    ...
}
```

clearer.

PR #4331, opened 2025-08-19
Remove custom Slice loader support in PHP

bernardnormier:
This PR removes the custom Slice loader support in PHP.

The existing code worked (and was tested) for simple PHP apps but was not correctly implementation for registered communicators. 

With registered communicators, it is essential that the application registers a new PHP custom Slice loader per PHP request, with a "setSliceLoader" or similar call on communicator. Rather than implement and test this rarely used API, this PR removes custom Slice loader support.

This PR also removes a related feature: preMarshal and postUnmarshal hooks. If you can't register a custom class for unmarshaling, postUnmarshal is not useful. + removed preMarshal for consistency.


php/src/Communicator.cpp
```diff
@@ -69,13 +66,11 @@ namespace IcePHP
         Ice::CommunicatorPtr getCommunicator() const final;
         Ice::SliceLoaderPtr getSliceLoader() const final;
```
bernardnormier, line 67: Note that we still have a per PHP request SliceLoader, now created lazily. It's now always a DefaultSliceLoader.

The whole unmarshaling relies on the per PHP-request CommunicatorInfo. I believe it's because unmarshaled proxies must know which CommunicatorInfo they belong to.

pepone (approved):

externl (approved):
Looks good. Would be good to include the removal of pre and post marshal functions in the changelog.

PR #4332, opened 2025-08-19
Improve Bad-Identifier Handling for DataMembers, Parameters, and Exceptions.

InsertCreativityHere:


cpp/src/Slice/Grammar.cpp
InsertCreativityHere, line 1: Generated code from Bison, no need to review.

cpp/src/Slice/Grammar.h
InsertCreativityHere, line 1: Generated code from Bison, no need to review.

cpp/src/Slice/Grammar.y
```diff
@@ -475,26 +469,21 @@ exception_decl
...
 // ----------------------------------------------------------------------
-: exception_id exception_extends
+: ICE_EXCEPTION definition_name exception_extends
 {
-    auto ident = dynamic_pointer_cast<StringTok>($1);
-    auto base = dynamic_pointer_cast<Exception>($2);
+    auto ident = dynamic_pointer_cast<StringTok>($2);
+    auto base = dynamic_pointer_cast<Exception>($3);
     ContainerPtr cont = currentUnit->currentContainer();
     ExceptionPtr ex = cont->createException(ident->v, base);
```
InsertCreativityHere, line 477: We always create the exception now, even if the identifier is 'bad'. So there's no need to handle the case where `ex = nullptr` anymore.

cpp/src/Slice/Grammar.y
```diff
@@ -794,106 +783,59 @@ data_member
...
+    else if (auto st = dynamic_pointer_cast<Struct>(currentUnit->currentContainer()))
     {
         if (def->isOptional)
         {
             currentUnit->error("optional data members are not supported in structs");
-            dm = st->createDataMember(def->name, def->type, false, 0, nullptr, std::nullopt); // Dummy
-        }
-        else
-        {
-            dm = st->createDataMember(def->name, def->type, false, -1, nullptr, std::nullopt);
```
InsertCreativityHere, line 813: For some reason `Struct::createDataMember` had 2 fields for storing optional tag info... we don't support optional struct fields. I removed these useless parameters from `createDataMember`, and simplified.

cpp/src/Slice/Grammar.y
```diff
@@ -794,106 +783,59 @@ data_member
...
     }
-    auto ex = dynamic_pointer_cast<Exception>(currentUnit->currentContainer());
-    if (ex)
+    else if (auto ex = dynamic_pointer_cast<Exception>(currentUnit->currentContainer()))
     {
         dm = ex->createDataMember(def->name, def->type, def->isOptional, def->tag, value->v, value->valueAsString);
     }
-    currentUnit->currentContainer()->checkHasChangedMeaning(def->name, dm);
-    $$ = dm;
-}
```
InsertCreativityHere, line 854: The logic below this comment was if you used a keyword as a DataMember identifier. This is already handled by the parser now, no need for this extra handling anymore.

cpp/src/Slice/Grammar.y
```diff
@@ -794,106 +783,59 @@ data_member
...
-        $$ = st->createDataMember(name, type, false, 0, nullptr, std::nullopt); // Dummy
-    }
-    auto ex = dynamic_pointer_cast<Exception>(currentUnit->currentContainer());
-    if (ex)
-    {
-        $$ = ex->createDataMember(name, type, false, 0, nullptr, std::nullopt); // Dummy
-    }
-    assert($$);
-    currentUnit->error("keyword '" + name + "' cannot be used as data member name");
-}
```
InsertCreativityHere, line 876: The logic below this comment was for if you forget to give your DataMember an identifier. This is already handled by the parser now, no need for this extra handling anymore.

cpp/src/Slice/Grammar.y
```diff
@@ -1165,14 +1107,20 @@ exception_list
 {
     auto exceptionList = dynamic_pointer_cast<ExceptionListTok>($1);
     auto exception = dynamic_pointer_cast<Exception>($3);
```
InsertCreativityHere, line 1109: This code is for exception specifications. The old logic had this crazy idea:
"if you put a bogus exception in your specification, we'll create an 'ephemeral' exception to put in the list"
Just no.

Now, if you use a bogus exception, the lookup returns `nullptr`, and we just don't add it to your exception specification.

cpp/src/Slice/Grammar.y
```diff
@@ -1185,18 +1133,23 @@ exception
     auto scoped = dynamic_pointer_cast<StringTok>($1);
     ContainerPtr cont = currentUnit->currentContainer();
     ExceptionPtr exception = cont->lookupException(scoped->v, true);
-    if (!exception)
+    if (exception)
     {
-        exception = cont->createException(Ice::generateUUID(), 0, Dummy); // Dummy
```
InsertCreativityHere, line 1190: The crazy part where we would generate ephemeral exceptions in case you listed a bogus exception.
Deleted.

cpp/src/Slice/Grammar.y
```diff
@@ -1346,33 +1299,6 @@ parameter
     }
     $$ = param;
 }
```
InsertCreativityHere, line 1301: The code beneath this handled keyword parameters `void op(int float);`.
Now this is already handled by the parser, no need for this extra logic here.

cpp/src/Slice/Grammar.y
```diff
@@ -1346,33 +1299,6 @@ parameter
...
-    ParameterPtr param;
-
-    auto op = dynamic_pointer_cast<Operation>(currentUnit->currentContainer());
-    if (op)
-    {
-        param = op->createParameter(ident->v, type, false, 0); // Dummy
-        currentUnit->error("keyword '" + ident->v + "' cannot be used as parameter name");
-    }
-    $$ = param;
-}
```
InsertCreativityHere, line 1362: The code beneath this handled missing parameter names `void op(int);`.
Now this is already handled by the parser, no need for this extra logic here.

cpp/src/Slice/Parser.h
```diff
@@ -19,12 +19,6 @@
 
 namespace Slice
 {
-    enum NodeType
-    {
-        Dummy,
-        Real
-    };
```
InsertCreativityHere, line 26: Now that we don't create crazy ephemeral Slice definitions anymore, this enum is unused and was deleted.

cpp/src/Slice/Parser.h
```diff
@@ -867,11 +860,9 @@ namespace Slice
     {
     public:
         Struct(const ContainerPtr& container, const std::string& name);
-        DataMemberPtr createDataMember(
+        [[nodiscard]] DataMemberPtr createDataMember(
             const std::string& name,
             const TypePtr& type,
-            bool isOptional,
-            std::int32_t tag,
```
InsertCreativityHere, line 874: This function creates data members in structs.
There is absolutely 0 reason to ever pass in whether it's optional or it's tag value.
Structs do no allow optional members.

cpp/test/Slice/errorDetection/DeclaratorMissing.err
```diff
@@ -1,4 +1,4 @@
-DeclaratorMissing.ice:7: missing parameter name
```
InsertCreativityHere, line 1: Before, bad identifier handling was done by each rule with it's own specific error message.
I updated the error messages now that they're using the more generic identifier checking that the parser does.

cpp/test/Slice/errorDetection/IdentAsKeyword.err
```diff
@@ -70,8 +70,10 @@ IdentAsKeyword.ice:66: interfaces can only be defined within a module
 IdentAsKeyword.ice:66: syntax error, unexpected out keyword
 IdentAsKeyword.ice:67: interfaces can only be defined within a module
 IdentAsKeyword.ice:69: interfaces can only be defined within a module
```
InsertCreativityHere, line 72: These extra 2 error messages are correct and good.
We only see them now because parsing can go a little further than before-hand.

cpp/src/Slice/Parser.cpp
```diff
@@ -1157,36 +1157,41 @@ Slice::Container::createInterfaceDecl(const string& name)
...
-Slice::Container::createException(const string& name, const ExceptionPtr& base, NodeType nodeType)
+Slice::Container::createException(const string& name, const ExceptionPtr& base)
 {
-    ContainedList matches = unit()->findContents(thisScope() + name);
-    if (!matches.empty())
+    ExceptionPtr p = make_shared<Exception>(shared_from_this(), name, base);
+
+    checkForGlobalDefinition("exceptions");
+
+    if (!name.empty())
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes, this is the point of the PR.
It's fine if an exception has an empty name during parsing, and we will exit immediately after parsing.

cpp/src/Slice/Grammar.y
```diff
@@ -1165,14 +1107,20 @@ exception_list
 {
     auto exceptionList = dynamic_pointer_cast<ExceptionListTok>($1);
     auto exception = dynamic_pointer_cast<Exception>($3);
-    exceptionList->v.push_back(exception);
+    if (exception)
```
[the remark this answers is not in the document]
InsertCreativityHere: Yep, this is the intended behavior.

bernardnormier (approved):

PR #4333, opened 2025-08-19
Add Full Support for `@p` and Code-Snippets to Slice Doc-Comments

InsertCreativityHere:
This PR adds full support (parsing _and_ code-gen in all languages) for `@p` (paramref) tags and markdown code-snippets.
Note: these are different than code-blocks which we do not support.

Instead of each language providing a single `DocLinkFormatter`, each language now provides a `DocCommentFormatter`, which provides functions for formatting links, code-snippets, and param-refs.

To see the exact changes this PR makes to the generated code, see this commit:
https://github.com/InsertCreativityHere/compiler-comparison/commit/fd75d6482836e3ce7e1e0cd281163b74f78e341e#diff-1df6277f7efe2c45afd0c3267f483bb86b73dfb6db309a3e4fdd1f9da629ad4b

### DIFF
--- a/cpp/src/Slice/DocCommentParser.cpp
+++ b/cpp/src/Slice/DocCommentParser.cpp
@@ -14,7 +14,7 @@ namespace Slice
     class DocCommentParser final : public ParserVisitor
     {
     public:
-        DocCommentParser(DocLinkFormatter linkFormatter) : _linkFormatter(linkFormatter) {}
+        DocCommentParser(DocCommentFormatter& formatter) : _formatter(formatter) {}
 
         bool visitModuleStart(const ModulePtr& p) final;
         void visitClassDecl(const ClassDeclPtr& p) final;
@@ -37,14 +37,14 @@ namespace Slice
         /// @remark This function must be called on a doc-comment before it's usable for code-gen purposes.
         void parseDocCommentFor(const ContainedPtr& p);
 
-        DocLinkFormatter _linkFormatter;
+        DocCommentFormatter& _formatter;
     };
 }
 
 void
-Slice::parseAllDocComments(const UnitPtr& unit, DocLinkFormatter linkFormatter)
+Slice::parseAllDocComments(const UnitPtr& unit, DocCommentFormatter& formatter)
 {
-    DocCommentParser visitor{linkFormatter};
+    DocCommentParser visitor{formatter};
     unit->visit(&visitor);
 }
 
@@ -256,44 +256,50 @@ DocCommentParser::parseDocCommentFor(const ContainedPtr& p)
         return;
     }
 
-    // TODO: this is a temporary hack since only "csharp" happens to set 'escapeXml'.
-    // If true, escapes all XML special characters in the parsed comment. Defaults to false.
-    const bool escapeXml = (p->unit()->languageName() == "cs");
-
-    // Split the comment's raw text up into lines.
+    // Get the comment's raw lines, and let the formatter perform any preprocessing it wants to.
     StringList lines = docComment->_rawDocCommentLines;
+    _formatter.preprocess(lines);
 
-    // Escape any XML entities if necessary.
-    if (escapeXml)
+    const string ws = " \t";
+
+    // Format any code-snippets. Slice fully supports Markdown's backtick handling.
+    // Normal text backticks can be escaped with a backslash: '\`',
+    // And we allow multiple backticks to be used as a single delimiter: '`` wow ` is cool``'.
+    for (auto& line : lines)
     {
-        const string amp = "&amp;";
-        const string lt = "&lt;";
-        const string gt = "&gt;";
-
-        for (auto& line : lines)
+        size_t pos = 0;
+        while ((pos = line.find('`', pos)) != string::npos)
         {
-            string::size_type pos = 0;
-            while ((pos = line.find_first_of("&<>", pos)) != string::npos)
+            // If this backtick is escaped with a backslash, skip it.
+            if (pos > 0 && line[pos-1] == '\'')
             {
-                switch (line[pos])
-                {
-                    case '&':
-                        line.replace(pos, 1, amp);
-                        pos += amp.size();
-                        break;
-                    case '<':
-                        line.replace(pos, 1, lt);
-                        pos += lt.size();
-                        break;
-                    case '>':
-                        line.replace(pos, 1, gt);
-                        pos += gt.size();
-                        break;
-                    default:
-                        assert(false);
-                        break;
-                }
+                pos += 1;
+                continue;
             }
+
+            // Count the number of successive backticks. The end of the code-snippet must have the same number.
+            auto openingEnd = line.find_first_not_of('`', pos);
+            openingEnd = (openingEnd == string::npos ? line.size() : openingEnd);
+            auto count = openingEnd - pos;
+
+            // Find the closing delimeter, starting for the end of the opening delimeter.
+            auto delimiter = string(count, '`');
+            auto closingStart = line.find(delimiter, openingEnd);
+            if (closingStart == string::npos)
+            {
+                // No matching delimeter was found.
+                p->unit()->warning(p->file(), p->line(), InvalidComment, "missing parameter name after '@p' tag");
+                break; // Skip to the next line, this line is broken.
+            }
+
+            // Format the snippet's text, and replace the entire raw code-snippet with the returned string.
+            const string snippetText = line.substr(openingEnd, closingStart - openingEnd);
+            const string formattedSnippet = _formatter.formatCode(snippetText);
+            line.erase(pos, closingStart + count - pos);
+            line.insert(pos, formattedSnippet);
+
+            // Go on to check for the next code-snippet, skipping past the one we just formatted.
+            pos += formattedSnippet.size();
         }
     }
 
@@ -301,15 +307,15 @@ DocCommentParser::parseDocCommentFor(const ContainedPtr& p)
     const string link = "{@link ";
     for (auto& line : lines)
     {
-        auto pos = line.find(link);
-        while (pos != string::npos)
+        size_t pos = 0;
+        while ((pos = line.find(link, pos)) != string::npos)
         {
             auto endpos = line.find('}', pos);
             if (endpos != string::npos)
             {
                 // Extract the linked-to identifier.
-                auto identStart = line.find_first_not_of(" \t", pos + link.size());
-                auto identEnd = line.find_last_not_of(" \t", endpos);
+                auto identStart = line.find_first_not_of(ws, pos + link.size());
+                auto identEnd = line.find_last_not_of(ws, endpos);
                 string linkText = line.substr(identStart, identEnd - identStart);
 
                 // Then erase the entire '{@link foo}' tag from the comment.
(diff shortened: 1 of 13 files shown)


cpp/src/Slice/Parser.h
```diff
@@ -256,19 +256,6 @@ namespace Slice
...
-    ///
-    /// The first argument (`rawLink`) is the raw link text, taken verbatim from the doc-comment.
-    /// The second argument (`source`) is a pointer to the Slice element that the doc comment (and link) are written on.
-    /// The third argument (`target`) is a pointer to the Slice element that is being linked to.
-    /// If the parser could not resolve the link, this will be `nullptr`.
-    ///
-    /// This function should return the fully formatted link, which will replace the entire '{@link <rawLink>}'
-    /// in a raw doc-comment.
-    using DocLinkFormatter =
-        std::string (*)(const std::string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target);
```
InsertCreativityHere, line 270: No longer needed. instead of a function-callback type, now it's just a member function of `DocCommentFormatter`.

cpp/src/Slice/DocCommentParser.h
```diff
@@ -5,9 +5,60 @@
...
+#    pragma warning(disable : 4100)
+#endif
+#ifdef __GNUC__
+// unused parameter ‘rawComment’
+#    pragma GCC diagnostic ignored "-Wunused-parameter"
+#endif
+#if defined(__clang__)
+// unused parameter ‘rawComment’
+#    pragma clang diagnostic ignored "-Wunused-parameter"
+#endif
```
InsertCreativityHere: We provide default implementations which are no-op, so they don't use the parameters.
But we still want to give the parameters name for documentation's sake.

cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -256,60 +256,67 @@ DocCommentParser::parseDocCommentFor(const ContainedPtr& p)
         return;
     }
 
-    // TODO: this is a temporary hack since only "csharp" happens to set 'escapeXml'.
-    // If true, escapes all XML special characters in the parsed comment. Defaults to false.
-    const bool escapeXml = (p->unit()->languageName() == "cs");
-
-    // Split the comment's raw text up into lines.
```
InsertCreativityHere, line 263: Instead of having this hard-coded check for escaping XML in `cs`, now each language can 'preprocess' doc-comments to do whatever they need to before the 'libSlice' doc-comment parser runs.

cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -256,60 +256,67 @@ DocCommentParser::parseDocCommentFor(const ContainedPtr& p)
...
+    // Get the comment's raw lines, and let the formatter perform any preprocessing it wants to.
     StringList lines = docComment->_rawDocCommentLines;
+    _formatter.preprocess(lines);
 
-    // Escape any XML entities if necessary.
-    if (escapeXml)
```
InsertCreativityHere, line 267: All this escaping code was C# specific. It was simplified and moved to `slice2cs`.

cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -256,60 +256,67 @@ DocCommentParser::parseDocCommentFor(const ContainedPtr& p)
...
-        const string amp = "&amp;";
-        const string lt = "&lt;";
-        const string gt = "&gt;";
+    const string ws = " \t";
 
-        for (auto& line : lines)
+    // Format any code-snippets. Slice fully supports Markdown's backtick handling.
+    // Normal text backticks can be escaped with a backslash: '\`',
+    // And we allow multiple backticks to be used as a single delimiter: '`` wow ` is cool``'.
+    for (auto& line : lines)
```
InsertCreativityHere, line 285: This code is all new and handles the parsing of code-snippets.

cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -332,19 +339,87 @@ DocCommentParser::parseDocCommentFor(const ContainedPtr& p)
...
-            pos = line.find(link, pos);
         }
     }
 
     // Some tags are only valid if they're applied to an operation.
     // And we need a reference to the operation to make sure any names used in the tag match the names in the operation.
     OperationPtr operationTarget = dynamic_pointer_cast<Operation>(p);
 
-    const string ws = " \t";
+    // Fix any '@p' tags using the provided param-ref formatter.
```
InsertCreativityHere, line 370: This code is all new and handles the parsing of parameter references.

cpp/src/slice2cpp/Main.cpp
```diff
@@ -23,6 +23,18 @@ namespace
     bool interrupted = false;
 }
 
+class CppDocCommentFormatter final : public DocCommentFormatter
+{
+    string formatCode(string rawText) final { return "<tt>" + rawText + "</tt>"; }
```
InsertCreativityHere: This is how Doxygen recommends formatting code-snippets:
https://www.doxygen.nl/manual/commands.html#cmdc
There is also `@c` but this only formats the next word, not an entire snippet.

cpp/src/slice2cs/Main.cpp
```diff
@@ -23,6 +23,44 @@ namespace
     bool interrupted = false;
 }
 
+class CsharpDocCommentFormatter final : public DocCommentFormatter
+{
+    void preprocess(StringList& rawComment) final
```
InsertCreativityHere, line 28: C# uses XML tags in it's doc-comments, so we need to escape any XML special characters before parsing the doc-comment.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -31,6 +31,20 @@ namespace
     const char* const tripleQuotes = R"(""")";
 }
 
+class PythonDocCommentFormatter final : public DocCommentFormatter
+{
+    string formatCode(string rawText) final
+    {
+        // We target Sphinx (RST) for Python doc-comments, which use double backticks for code formatting.
```
```suggestion
        // We target Sphinx (RST) for Python doc-comments, which uses double backticks for code formatting.
```

cpp/src/ice2slice/Gen.cpp
```diff
@@ -768,23 +768,3 @@ Gen::TypesVisitor::getOutput(const ContainedPtr& contained)
         return *(inserted.first->second);
     }
 }
-
-string
-Slice::slice2LinkFormatter(const string& rawLink, const ContainedPtr&, const SyntaxTreeBasePtr&)
```
InsertCreativityHere, line 773: This was moved to `Main.cpp`, the one place it's now used.

cpp/src/slice2cs/Main.cpp
```diff
@@ -23,6 +23,44 @@ namespace
...
+                    case '<':
+                        line.replace(pos, 1, "&lt;");
+                        break;
+                    case '>':
+                        line.replace(pos, 1, "&gt;");
+                        break;
+                }
+                // Skip over the leading '&' character to avoid 'find'ing it again.
+                pos += 1;
+            }
```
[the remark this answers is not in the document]
InsertCreativityHere: This isn't a concern. We only check for `&`, and if it's followed by 'a', the code will just skip it.
This is not optimally efficient, but the extra simplicity is worth checking 2 more letters than is necessary IMO.

cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -256,60 +256,67 @@ DocCommentParser::parseDocCommentFor(const ContainedPtr& p)
...
+            openingEnd = (openingEnd == string::npos ? line.size() : openingEnd);
+            auto count = openingEnd - pos;
+
+            // Find the closing delimeter, starting for the end of the opening delimeter.
+            auto delimiter = string(count, '`');
+            auto closingStart = line.find(delimiter, openingEnd);
+            if (closingStart == string::npos)
+            {
+                // No matching delimeter was found.
+                const string msg = "unterminated code snippet: missing closing backtick delimiter";
```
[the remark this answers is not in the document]
InsertCreativityHere: This is more detail than is worth mentioning.

cpp/src/Slice/DocCommentParser.cpp
pepone: Do all compilers have to provide a `formatCode`, for example if the target supports Markdown style code blocks we can avoid doing all this work.

cpp/src/Slice/DocCommentParser.h
pepone: I think is clear to provide the default implementation as a non inline, and keep the parameter name just for the declaration in the header.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -31,6 +31,20 @@ namespace
     const char* const tripleQuotes = R"(""")";
 }
 
+class PythonDocCommentFormatter final : public DocCommentFormatter
+{
+    string formatCode(const string& rawText) final
+    {
+        // We target Sphinx (RST) for Python doc-comments, which uses double backticks for code formatting.
+        return "``" + rawText + "``";
```
pepone, line 39: There is a possibility for rawText to contain a double back-tick, if the Slice block was it self delimited with a triple back-tick, I think we can probably ignore this.

cpp/src/Slice/DocCommentParser.h
```diff
@@ -5,9 +5,60 @@
 
 #include "Parser.h"
 
+#ifdef _MSC_VER
```
bernardnormier: Just use `[[maybe_unused]]`

cpp/src/Slice/DocCommentParser.h
bernardnormier: That's fine too. In particular, it's preferable to avoid defining polymorphic classes with only inline functions; that's why we occasionally put a default dtor out of line.

cpp/src/slice2cpp/Main.cpp
```diff
@@ -23,6 +23,18 @@ namespace
     bool interrupted = false;
 }
 
+class CppDocCommentFormatter final : public DocCommentFormatter
+{
+    string formatCode(const string& rawText) final { return "<tt>" + rawText + "</tt>"; }
```
bernardnormier: I would use backticks and not `<tt>`. That's why we do for code blocks in our own C++ doc-comments. That's even what you did in this PR.

cpp/src/Slice/DocCommentParser.cpp
InsertCreativityHere: No, they only need to provide a `formatCode` if they do something different than use normal Markdown.

InsertCreativityHere: `ice2slice`, `slice2js`, `slice2swift` use the default implementation which just wraps the text in backticks.

cpp/src/slice2cpp/Main.cpp
InsertCreativityHere: I am absolutely fine switching back to backticks, but my reason for using `<tt>` is that we don't need to worry about escaping the inside. For example, if you write this in your Slice (which is totally valid):
```cpp
/// Here is code: `` b = '`'; ``
```
We would generate a broken doc-comment:
```cpp
/// Here is code: ` b = '`'; `           // Unterminated snippet!
```

Yes, this is a minor concern, but using `<tt>` costs us nothing, and completely avoids this problem.

cpp/src/Slice/DocCommentParser.cpp
pepone: I mean we still have to do all this parsing, even if the generated code is identical to the original code. Then wrapping in backticks doesn't account for backtick escaping does it?

InsertCreativityHere: No it doesn't account for this, but that's somewhat intentional. Even though backtick escaping is part of Markdown, there's actually many implementations that don't handle it correctly.

`slicec` (consumer of `ice2slice`) and JSDoc/TSDoc (consumer of `slice2js`) only support single backticks, and get confused if you use multiple backticks as a single delimiter, for example.

pepone: So we map

```
``Foo`Bar``
```

To:

```
`Foo`Bar`
```

InsertCreativityHere: Yes, that's the best we can do unfortunately. Even if we preserved the number of backticks from Slice, it would be treated like this by those languages
` `Foo`Bar` `

Which isn't what we'd want. Obviously, we can update `slicec` later on to handle it, but right now it doesn't.
And I couldn't find a way to escape them at all for JSDoc/TSDoc. All that happens is you get this error:
`A code span must contain at least one character between the backticks` because it treats the opening ` `` ` as an empty code-span unfortunately.

If you can find a way to handle it, let me know! But, this also isn't the most critical feature.

bernardnormier: Also obviously we can only support features that Doxygen itself supports, with the same syntax or a subset of what Doxygen supports.

See https://www.doxygen.nl/manual/markdown.html#md_codespan

> To show a literal backtick or single quote inside a code span use double backticks, i.e.

PR #4334, opened 2025-08-20
Python Ice/custom type hinting

### LINKED ISSUE #4305: Remove the `python:seq:list` and `python:seq:tuple` Metadata
These metadata are equivalent to the newer (and shorter) `python:list` and `python:tuple` metadata.
I think we should just remove the old `python:seq:list` and `python:seq:tuple` metadata instead of keeping these around as 'aliases'.

It's true that this will break existing metadata, but it should be trivial for users to just remove the `seq:` in the middle.

externl:
This PR adds type hinting to the Python Ice/custom test. 

Closes #4305


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -247,6 +250,10 @@ Slice::Python::CodeVisitor::typeToTypeHintString(
 
             auto elementType = dynamic_pointer_cast<Builtin>(seq->type());
             bool isByteSequence = elementType && elementType->kind() == Builtin::KindByte;
+            bool isBoolSequence = elementType && elementType->kind() == Builtin::KindBool;
+            auto sequenceMetadata = getSequenceMetadata(seq, localMetadata);
```
externl, line 255: We were not previously checking "local metadata"

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -688,21 +753,45 @@ Slice::Python::ImportVisitor::visitConst(const ConstPtr& p)
...
+        addRuntimeImport("array", "array", source);
     }
-    else if (sequence->hasMetadata("python:memoryview"))
+    else if (directive == "python:memoryview")
     {
-        // TODO
+        auto arguments = metadata ? metadata->arguments() : "";
+        auto [factory, typeHint] = splitMemoryviewArguments(arguments);
+
+        if (dynamic_pointer_cast<ClassDef>(source) || dynamic_pointer_cast<Struct>(source))
```
externl: Is this the correct place for this source type check?

cpp/src/slice2py/PythonUtil.h
```diff
@@ -184,8 +184,21 @@ namespace Slice::Python
...
+    /// @return A pair containing the module and name components.
+    std::pair<std::string, std::string> splitFQDN(const std::string& fqdn);
+
+    /// Splits the arguments the python:memoryview metadata into the factory function and optional type hint.
+    /// @param arguments The arguments of python:memoryview
+    /// @return A pair containing the factory function and optional type hint
+    std::pair<std::string, std::optional<std::string>> splitMemoryviewArguments(const std::string& arguments);
+
+    // A helper class to initialize the _outBuffer member of BufferedOutput before is
+    // passed to the Output constructor.
```
InsertCreativityHere: Not your change, but this should be:
```suggestion
    // A helper class to initialize the _outBuffer member of BufferedOutput before it's
    // passed to the Output constructor.
```

python/test/Ice/custom/AllTests.py
```diff
@@ -10,26 +10,11 @@
 if "--load-slice" in sys.argv:
     TestHelper.loadSlice("Test.ice")
 
-try:
```
externl, line 13: Pyright did not like this structure. It had no way to know if `numpy` was available or not.

cpp/src/slice2py/PythonUtil.h
```diff
@@ -184,8 +184,21 @@ namespace Slice::Python
...
 
-    // A helper class to initialize the _outBuffer member of BufferedOutput before is passed to the Output
-    // constructor.
+    /// Get sequence metadata associated with the given sequence and any local metadata.
+    Slice::MetadataPtr getSequenceMetadata(const SequencePtr& seq, const MetadataList& localMetadata);
+
+    /// Splits a FQDN into its Module and Name components.
+    /// @param fqdn The fully qualified domain name to split.
+    /// @return A pair containing the module and name components.
+    std::pair<std::string, std::string> splitFQDN(const std::string& fqdn);
```
InsertCreativityHere: Definitely not critical, but is this really called an FQDN?
I've really only heard that used to talk about networking (domain names and all).

Maybe just fully-qualified type or something? `splitQualifiedType`?

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -258,34 +265,78 @@ Slice::Python::CodeVisitor::typeToTypeHintString(
...
                 }
+                else if (sequenceMetadata && metadataDirective == "python:memoryview")
+                {
+                    auto [_, typeHint] = splitMemoryviewArguments(sequenceMetadata->arguments());
+                    if (typeHint)
+                    {
+                        auto [package, name] = splitFQDN(*typeHint);
+                        auto memoryViewType = getImportAlias(source, package, name);
+                        os << " | " << memoryViewType;
+                    }
```
InsertCreativityHere: Is it possible that we could fail to import the type if it doesn't exist or something?

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2872,6 +3047,36 @@ Slice::Python::validatePythonMetadata(const UnitPtr& unit)
                        "' metadata can only be applied to sequences of bools, bytes, shorts, ints, longs, floats, "
                        "or doubles";
             }
+
+            if (m->directive() == "python:memoryview")
+            {
+                // Argument can't be empty as it sets MetadataArgumentKind::RequiredTextArgument
+                auto arguments = m->arguments();
+                assert(!arguments.empty());
```
InsertCreativityHere: The code-gen will never see bad identifiers, but this is run in the validator, before we can make any assumptions about whether or not the metadata is valid.

So it's possible that the arguments will be empty at this point.
Another part of the validator will emit an error for it of course, but that check could happen _after_ this one.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2079,6 +2210,50 @@ Slice::Python::getAll(const ContainedPtr& definition)
...
+                return metadata;
+            }
+        }
+    }
+
+    return nullptr;
+}
+
+pair<string, string>
+Slice::Python::splitFQDN(const std::string& fqdn)
```
InsertCreativityHere: No need to qualify this.
```suggestion
Slice::Python::splitFQDN(const string& fqdn)
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2079,6 +2210,50 @@ Slice::Python::getAll(const ContainedPtr& definition)
...
+{
+    size_t lastDot = fqdn.rfind('.');
+    if (lastDot != std::string::npos)
```
InsertCreativityHere: Nor this
```suggestion
    if (lastDot != string::npos)
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2079,6 +2210,50 @@ Slice::Python::getAll(const ContainedPtr& definition)
...
+        return {fqdn.substr(0, lastDot), fqdn.substr(lastDot + 1)};
+    }
+    return {"", fqdn};
+}
+
+pair<string, optional<string>>
+Slice::Python::splitMemoryviewArguments(const string& arguments)
+{
+    size_t pos = arguments.find(':');
+    if (pos != std::string::npos)
```
InsertCreativityHere: Nor this. Obviously nothing _wrong_ with qualifying it, but we should be consistent.
```suggestion
    if (pos != string::npos)
```

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -258,34 +265,78 @@ Slice::Python::CodeVisitor::typeToTypeHintString(
                 {
                     os << " | bytes";
                 }
-                if (seq->hasMetadata("python:numpy.ndarray"))
+
+                if (metadataDirective == "python:array.array" && isBoolSequence)
+                {
+                    // array.array does not support boolean values. The common convention is to use a
+                    // char array, array.array('b'), whose Python type is int.
```
pepone: I will write something more like: For boolean sequences "python:array.array" is mapped to `array.array('b')`, whose type-hint is `array.array[int]`.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -258,34 +265,78 @@ Slice::Python::CodeVisitor::typeToTypeHintString(
...
                     const string numpyAlias = getImportAlias(source, "numpy");
                     os << " | " << numpyAlias << ".typing.NDArray[" << numpyAlias << "."
                        << numpyBuiltinTable[elementType->kind()] << "]";
                 }
+                else if (sequenceMetadata && metadataDirective == "python:memoryview")
+                {
+                    auto [_, typeHint] = splitMemoryviewArguments(sequenceMetadata->arguments());
+                    if (typeHint)
+                    {
+                        auto [package, name] = splitFQDN(*typeHint);
```
pepone: Not sure about naming this splitFQDN. The term FQND doesn't seems a common outside domain names.

Maybe `splitFullyQualifiedName` the term "fully qualified name" is used in https://peps.python.org/pep-0737/

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -688,21 +749,45 @@ Slice::Python::ImportVisitor::visitConst(const ConstPtr& p)
...
+
+        if (typeHint)
+        {
+            auto [typeHintPackage, typeHintName] = splitFQDN(*typeHint);
+            addRuntimeImport(typeHintPackage, typeHintName, source);
+        }
+        else
+        {
+            // Otherwise, we have no idea what the type is so we just Any
+            addRuntimeImport("typing", "Any", source);
```
pepone: These two should be `addTypingImport` they are not used at runtime.

cpp/src/slice2py/PythonUtil.cpp
pepone: I think so, the other option is to do this in the `ImportVisitor::visitDataMember` implementation from where we call this.

What about Exceptions, can't they contain sequences? I think we might need a similar check for numpy/array. The runtime import is only required for data-members, for operations the typing import should be enough.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1383,30 +1471,73 @@ Slice::Python::CodeVisitor::visitDataMember(const DataMemberPtr& p)
...
+            auto arguments = sequenceMetadata->arguments();
+            auto [factory, _] = splitMemoryviewArguments(arguments);
+
+            auto [factoryPackage, factoryFunction] = splitFQDN(factory);
+
+            auto factoryFunctionAlias = getImportAlias(parent, factoryPackage, factoryFunction);
+
+            // This builtin type integer corresponds to the values of Ice.BuiltinBool, Ice.BuiltinByte, etc.
+            int builtinType;
+            switch (elementType->kind())
```
pepone: I will use a static array like we do for other things:

```
static const char* builtinTable[] = {
    "1", // Builtin::KindByte
    "0", // Builtin::KindBool
    "2", // Builtin::KindShort
    "3", // Builtin::KindInt
    "4", // Builtin::KindLong
    "5", // Builtin::KindFloat
    "6", // Builtin::KindDouble
};
assert(elementType->kind() <= Builtin::KindDouble);
...
<< builtinTable[elementType->kind()]

File: CHANGELOG-3.8.md [hunk 15]
```diff
@@ -838,6 +838,14 @@ initialization. See `InitializationData.pluginFactories`.
...
 - Removed support for the 'marshaled-result' metadata directive.
 
+- Removed support for the `python:seq` metadata. Use `python:list` and `python:tuple` instead.
+
+- Added type hints to the Ice package and generated code.
+
+- Added additional (optional) argument to `python:memoryview` metadata directive which is used to specify a type hint.
+  For example, `["python:memoryview:CustomFactory.myByteSeq:array.array"]` indicates that the factory function
+  will return an `array.array`.
+
```
pepone, line 848: One minor issue is we don't support generic type hints because of how we generate the required imports.

For example if you write:

```
["python:memoryview:CustomFactory.myByteSeq:array.array[int]"]
```

This would generate a bogus import like:

```
from array import array[int]
```

externl: Yes, this is something we could improve.

cpp/src/slice2py/PythonUtil.cpp
externl: I changed to FQN.

cpp/src/slice2py/PythonUtil.cpp
externl: Aren't they needed at runtime for dataclasses

cpp/src/slice2py/PythonUtil.cpp
externl: Yeah, but it's a user problem.

cpp/src/slice2py/PythonUtil.h
externl: Renamed to `splitFQN`

cpp/src/slice2py/PythonUtil.cpp
externl: I removed the assert.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -258,34 +266,77 @@ Slice::Python::CodeVisitor::typeToTypeHintString(
...
                    << "]";
             }
-            else if (seq->hasMetadata("python:array.array"))
+            else if (metadataDirective == "python:array.array")
             {
                 assert(elementType && elementType->kind() <= Builtin::KindDouble);
-                const string arrayAlias = getImportAlias(source, "array");
-                os << arrayAlias << ".array[" << typeToTypeHintString(seq->type(), false, source, forMarshaling) << "]";
+                const string arrayAlias = getImportAlias(source, "array", "array");
+                // array.array does not support boolean values. The common convention is to use a
```
pepone: I would update this comment to match the one above regarding array and bool seq.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -688,21 +749,63 @@ Slice::Python::ImportVisitor::visitConst(const ConstPtr& p)
...
     {
         // Import numpy for using it in the field factory.
-        addRuntimeImport("numpy", "", source);
+        if (needsRunTimeImport)
+        {
+            addRuntimeImport("numpy", "", source);
+        }
+        else
+        {
+            addRuntimeImport("numpy", "", source);
```
pepone, line 772: The else branch should call `addTypingImport`

PR #4335, opened 2025-08-20
Simplify PHP communicator destroy implementation

pepone:


php/src/Communicator.cpp
```diff
@@ -209,28 +209,25 @@ ZEND_METHOD(Ice_Communicator, destroy)
     assert(c);
     CommunicatorMap* m = reinterpret_cast<CommunicatorMap*>(ICE_G(communicatorMap));
     assert(m);
-    if (m->find(c) != m->end())
```
pepone, line 212: This lookup doesn't make sense, the communicator must be in the communicatorMap for the request if destroy is being called.

bernardnormier: What if you call destroy twice on the same communicator in the same PHP request?

Doesn't matter much since you can call destroy twice on the C++ communicator.

pepone: Yes calling destroy twice should be fine, I have improved the new registeredCommunicator test to check that calling destroy removes the registrations, and that a second call has no effect.

bernardnormier: I assume that with your test, `m->erase(c)` returns 0 for the second destroy.

externl (approved):

php/src/Communicator.cpp
pepone: Yes, second call should return 0, we are removing by key.

PR #4341, opened 2025-08-21
Do not Parse the Inside of Code-Snippets and CHANGELOG Updates

InsertCreativityHere:
This PR updates the 3.8 changelog to include all the new doc-comment changes, and the removal of the `-E` compiler option.

It also updates the doc-comment parsing in 2 ways:
1) The logic for fixing `@link`, `@p`, and code-snippets was moved out into helper functions, instead of being in one giant monolith function. The logic was kept identical, it was just moved. That's all.
2) Previously we would handle each tag in separate passes, but this had bad side effects like letting this: `` `@p param` `` get picked up as _both_ a code-snippet and a param-ref, which is wrong. Now we handle all the inline tags in a single pass, preventing any parsing 'overlap'.

### DIFF
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -374,6 +374,47 @@ classDiagram
   int sendRequest(string message);
   ```
 
+- Added support for 2 new doc-comment tags: `@remark` and `@p`
+  `@remark` can be used to provide additional (often non-critical) information in a doc-comment:
+
+  ```slice
+  /// Provides access to a radar scanner.
+  /// @remark for sonar, use {@link SonarScan} instead.
+  class RadarScan {}
+  ```
+
+  `@p` can only be used on operations, and provides a way to reference it's parameters in the doc-comment.
+  This is different than `@param` which is used to describe a parameter, not reference one. For example:
+
+  ```slice
+  /// Returns how many seconds passed between start and end.
+  /// @param start The starting time.
+  /// @param end The ending time. Must be greater than @p start.
+  int getDuration(int start, int end);
+  ```
+
+- Added support for Markdown code-snippets in doc-comments:
+
+  ```slice
+  /// This is `true`, and can be used as a default value: `bool b = ConstBool;`
+  const bool ConstBool = true;
+  ```
+
+  You can use any number of backticks to mark the beginning/end of the code-snippet, but they must match in number.
+  This is useful if the code-snippet itself will contain backticks:
+
+  ```slice
+  /// This snippet ``contains a ` character!`` just fine.
+  ```
+
+  Outside of code-snippets, you can escape a backtick with a backslash to get a normal textual backtick character:
+
+  ```slice
+  /// This character \` doesn't start a code-snippet and will appear like a normal backtick character.
+  ```
+
+- `@link` tags are now mapped to proper documentation links in Swift and MATLAB.
+
 - Lists of metadata can be split into separate brackets now, allowing for longer metadata to be placed on separate lines
   or for metadata to be grouped by functionality. For example, you can now write:
 
@@ -430,6 +471,8 @@ classDiagram
   should be used to generate Slice API documentation.
 
 - Removed the `--impl` argument from the Slice compilers.
+
+- Removed the `-E` flag from the Slice compilers.
 
 - You can now use identifiers with underscores or with the Ice prefix without any special compiler option.
 
--- a/cpp/src/Slice/DocCommentParser.cpp
+++ b/cpp/src/Slice/DocCommentParser.cpp
@@ -261,6 +261,168 @@ namespace
         ContainedList results = linkSourceScope->lookupContained(linkText, false);
         return (results.empty() ? nullptr : results.front());
     }
+
+    /// Formats a code-snippet starting at \p pos.
+    /// Slice fully supports Markdown's backtick handling, so normal text backticks can be escaped with a backslash
+    /// '\`', and we allow multiple backticks to be used as a single delimiter: '`` wow ` is cool``'.
+    /// @param line The line containing the code-snippet. This line will be edited in place to remove the original
+    ///             Slice-style code-snippet and replace it with a properly formatted code-snippet.
+    /// @param pos The position where the '`' appears in the line. This value should be updated to account for
+    ///            any characters this function inserts to avoid double-parsing things.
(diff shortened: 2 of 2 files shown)


CHANGELOG-3.8.md
```diff
@@ -374,6 +374,46 @@ classDiagram
   int sendRequest(string message);
   ```
 
+- Added support for 2 new doc-comment tags: `@remark` and `@p`
+  `@remark` can be used to provide additional (often non-critical) information in a doc-comment:
+
+  ```slice
+  /// Provides access to a radar scanner.
+  /// @remark for sonar, use {@link SonarScan} instead.
```
bernardnormier: Should be "For sonar"

CHANGELOG-3.8.md
```diff
@@ -374,6 +374,46 @@ classDiagram
...
+  `@p` can only be used on operations, and provides a way to reference parameters in its doc-comment.
+  This is different than `@param`. `@param` is used to describe a parameter, not reference one. For example:
+
+  ```slice
+  /// @param start The start.
+  /// @param end The end. Must be greater than @p start.
+  int getDuration(int start, int end);
+  ```
+
+- Added support for Markdown style code-snippets in doc-comments:
```
bernardnormier: I would call these "Mardown-style code spans" and include a link to https://www.doxygen.nl/manual/markdown.html#md_codespan.

Naturally, for doc-comments, we should use the same terminology as Doxygen.

CHANGELOG-3.8.md
```diff
@@ -374,6 +374,46 @@ classDiagram
...
+
+  ```slice
+  /// This is `true`, and can be used as a default value: `bool b = ConstBool;`
```
bernardnormier: I'd rather name this const `Yes`.

CHANGELOG-3.8.md
```diff
@@ -431,6 +471,8 @@ classDiagram
 
 - Removed the `--impl` argument from the Slice compilers.
 
+- Removed the `-E` flag from the Slice compilers.
```
bernardnormier: We should use a consistent terminology: flag, argument, option?

In the Ice Manual, we use "command-line options".

CHANGELOG-3.8.md
pepone: I find this confusing, we only support inline code-blocks, when I see code-snippets,  I will assume we a snippet can span multiple-lines, maybe Markdown use a different terminology for multi-line code-blocks?

bernardnormier: In the context of doc-comments, we should definitely use the same terminology as Doxygen. See: https://www.doxygen.nl/manual/markdown.html

So inline and single-line enclosed in backtick: code spans.
Multi-line with triple backtick: fenced code blocks

It's not clear to me if we support anything other than code spans.

InsertCreativityHere: > It's not clear to me if we support anything other than code spans.

No, we only support code-spans.
In the future we could support code-blocks, but these are more involved to parse and map correctly.

----

I'll update the terminology to match Doxygen.

CHANGELOG-3.8.md
InsertCreativityHere: Everywhere else in the CHANGELOG we also refer to these as options. I updated both places to use this term instead of `argument` or `flag`.

CHANGELOG-3.8.md
InsertCreativityHere: Updated to use 'code-span' everywhere.

CHANGELOG-3.8.md
InsertCreativityHere: Fixed.

cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -321,122 +308,182 @@ DocCommentParser::parseDocCommentFor(const ContainedPtr& p)
...
-                    p->unit()->warning(p->file(), p->line(), InvalidComment, msg);
-                }
-            }
-            else
+            // If the next element to handle is a param-ref.
+            if (nextParamrefPos != string::npos && (nextParamrefPos < nextBacktickPos && nextParamrefPos < nextLinkPos))
             {
-                p->unit()->warning(p->file(), p->line(), InvalidComment, "'@p' tags can only be used on operations");
+                pos = nextParamrefPos;
+                formatParamrefElement(p, _formatter, line, pos);
```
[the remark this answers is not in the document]
InsertCreativityHere: It's logic is flawed here.

> When nextLinkPos ... is string::npos, the comparison nextBacktickPos < nextLinkPos will be false since string::npos is the maximum value.

This is a factually incorrect statement.

cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -620,9 +665,12 @@ DocCommentParser::parseDocCommentFor(const ContainedPtr& p)
                 if (line[0] == '@')
                 {
                     auto unknownTag = line.substr(0, line.find_first_of(" \t"));
-                    const string msg = "ignoring unknown doc tag '" + unknownTag + "' in comment";
-                    p->unit()->warning(p->file(), p->line(), InvalidComment, msg);
-                    currentSection = nullptr;
+                    if (unknownTag != "@p")
```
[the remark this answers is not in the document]
InsertCreativityHere: That is an understandable conclusion to reach, but is not correct here.
A `@p` tag will be _mapped_ to another `@p` tag in C++, which could be incorrectly flagged by this logic.
Once we've finished inline-parsing, any `@p` tags can only be from this, and should not be treated as 'incorrect'.

pepone (approved):

cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -261,58 +261,45 @@ namespace
...
-    StringList lines = docComment->_rawDocCommentLines;
-    _formatter.preprocess(lines);
-
-    const string ws = " \t";
 
-    // Format any code-snippets. Slice fully supports Markdown's backtick handling.
-    // Normal text backticks can be escaped with a backslash: '\`',
-    // And we allow multiple backticks to be used as a single delimiter: '`` wow ` is cool``'.
-    for (auto& line : lines)
+    /// Formats a code-snippet starting at \p pos.
```
bernardnormier: Please replace all "snippet" by code span(s).

InsertCreativityHere: Fixed.

bernardnormier (approved):

PR #4342, opened 2025-08-21
Add IcePy stubs and py.typed files to zeroc-ice package

externl:
This PR adds more typing information to the zeroc-ice python package

- py.typed for Ice, etc. to indicate packages are typed
- stub file for IcePy.

I tested on macOS (using local build and the demos), we should also test Windows and Linux. 

See https://peps.python.org/pep-0561/#partial-stub-packages


python/setup.py
```diff
@@ -321,7 +321,7 @@ def run(self):
 
 # Setup configuration for the package
 setup(
-    packages=packages,
+    packages=packages + ["IcePy-stubs"],
```
pepone: Why don't add "IcePy-stubs" to the packages array, which is only used here, or move the array here if you prefer.

externl: I'll move the array here

PR #4347, opened 2025-08-22
Fix for typing import of bare modules

pepone:
The typing import didn't handle correctly importing a bare module like "import numpy".  This PR fixes the issue and refactors the code that handle runtime and typing imports to use a helper function and avoid duplication.

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -249,7 +249,6 @@ Slice::Python::CodeVisitor::typeToTypeHintString(
     }
     else
     {
-        string definitionModule = getPythonModuleForDefinition(type);
         string sourceModule = getPythonModuleForDefinition(source);
 
         auto contained = dynamic_pointer_cast<Contained>(type);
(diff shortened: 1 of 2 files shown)


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1028,24 +1001,6 @@ Slice::Python::ImportVisitor::addTypingImport(
     }
 }
 
-void
-Slice::Python::ImportVisitor::addTypingImport(const string& packageName, const ContainedPtr& source)
```
pepone, line 1032: This was unused.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -753,7 +752,6 @@ Slice::Python::ImportVisitor::visitDictionary(const DictionaryPtr& p)
 void
 Slice::Python::ImportVisitor::visitEnum(const EnumPtr& p)
 {
-    // TODO if a value is initialized with a constant, we need to import the type of the constant.
```
pepone, line 756: The TODO was bogus, the generated code uses the constant value not the constant, there is no need to import the constant.

externl (approved):

PR #4356, opened 2025-08-27
Fix C# multicast bug on Windows 

### LINKED ISSUE #3882: C# Ice.PluginInitializationException for IceDiscovery when IP address is not specified
### Version

- Ice 3.8 Nightly
- Windows 11

### Describe the bug

When attempting to initialize a communicator for a server in C# using IceDiscovery to listen for clients on Windows 11, you will get an Ice.PluginInitializationException unless you specify an IP. In my case, specifying the IP was done through a command argument `dotnet run --IceDiscovery.Interface=127.0.0.1`.

### Expected behavior

Instead of throwing an exception, the server is expected to finish initializing and begin listening for clients.

### Additional context

_No response_

bernardnormier:
This PR fixes #3882.

The root issue was that getLocalAddresses was returning sometimes on Windows 11 169.x addresses (C# only).

This PR also simplifies the corresponding C++ code.

### DIFF
--- a/cpp/src/Ice/Network.cpp
+++ b/cpp/src/Ice/Network.cpp
@@ -113,7 +113,8 @@ namespace
         return fd;
     }
 
-    vector<Address> getLocalAddresses(ProtocolSupport protocol, bool includeLoopback, bool singleAddressPerInterface)
+    // Only used for multicast.
+    vector<Address> getLocalAddresses(ProtocolSupport protocol)
     {
         vector<Address> result;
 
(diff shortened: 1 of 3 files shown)


cpp/src/Ice/Network.cpp
```diff
@@ -113,7 +113,8 @@ namespace
         return fd;
     }
 
-    vector<Address> getLocalAddresses(ProtocolSupport protocol, bool includeLoopback, bool singleAddressPerInterface)
```
bernardnormier, line 116: We only call this function once with includeLoopback=true and singleAddressPerInterface=true

cpp/src/Ice/Network.cpp
```diff
@@ -191,15 +184,15 @@ namespace
...
-            if (curr->ifa_addr && (includeLoopback || !(curr->ifa_flags & IFF_LOOPBACK)))
+            if (curr->ifa_addr)
             {
                 if (curr->ifa_addr->sa_family == AF_INET && protocol != EnableIPv6)
                 {
                     Address addr;
                     memcpy(&addr.saStorage, curr->ifa_addr, sizeof(sockaddr_in));
                     if (addr.saIn.sin_addr.s_addr != 0)
                     {
-                        if (!singleAddressPerInterface || interfaces.find(curr->ifa_name) == interfaces.end())
```
bernardnormier, line 202: becomes `!true || ...`

cpp/src/Ice/Network.cpp
```diff
@@ -291,10 +284,9 @@ namespace
                     {
                         Address addr;
                         memcpy(&addr.saStorage, &ifr[j].ifr_addr, sizeof(sockaddr_in));
-                        if (addr.saIn.sin_addr.s_addr != 0 &&
```
bernardnormier, line 294: All this code is apparently for Solaris and AIX. I think we should remove it.

csharp/src/Ice/Internal/Network.cs
```diff
@@ -578,7 +579,9 @@ internal static IPAddress[] getLocalAddresses(int protocol, bool singleAddressPe
         try
         {
             addresses = new List<IPAddress>();
-            NetworkInterface[] nics = NetworkInterface.GetAllNetworkInterfaces();
+            IEnumerable<NetworkInterface> nics = NetworkInterface.GetAllNetworkInterfaces().Where(
+                nic => nic.SupportsMulticast && nic.OperationalStatus == OperationalStatus.Up);
```
bernardnormier, line 583: The actual fix.

cpp/src/Ice/Network.cpp
InsertCreativityHere: I agree we should just remove all this.
Definitely seems like it's just leftover from Solaris support...

cpp/src/Ice/Network.cpp
InsertCreativityHere: Very nice simplification.
Crazy we still had these 2 bools sitting around...

InsertCreativityHere (approved):
Looks good to me!

cpp/src/Ice/Network.cpp
externl: We should definitely remove this.

externl (approved):

PR #4358, opened 2025-08-27
Update doc-comments for checkedCast and uncheckedCast

bernardnormier:
This PR updates the doc-comments for checkedCast and uncheckedCast, and avoids using the term "downcast" and "cast".

Both checkedCast and uncheckedCast are factory methods. They don't perform any casting.

### DIFF
--- a/cpp/include/Ice/ProxyFunctions.h
+++ b/cpp/include/Ice/ProxyFunctions.h
@@ -30,22 +30,20 @@ namespace Ice
         }
     }
 
-    /// Downcasts a proxy without confirming the target object's type via a remote invocation.
+    /// Creates a new proxy from an existing proxy.
     /// @param proxy The source proxy.
-    /// @return A proxy with the requested type.
-    /// @remark The preferred syntax is to construct the proxy from another proxy using the explicit proxy
-    /// constructor.
+    /// @return A new proxy with the requested type.
+    /// @remark This is a purely local operation.
     template<typename Prx, std::enable_if_t<std::is_base_of_v<ObjectPrx, Prx>, bool> = true>
     Prx uncheckedCast(const ObjectPrx& proxy)
     {
         return Prx::_fromReference(proxy._getReference());
     }
 
-    /// Downcasts a proxy without confirming the target object's type via a remote invocation.
+    /// Creates a new proxy from an existing proxy.
     /// @param proxy The source proxy (can be nullopt).
-    /// @return A proxy with the requested type.
-    /// @remark The preferred syntax is to construct the proxy from another proxy using the explicit ObjectPrx
-    /// constructor.
+    /// @return A new proxy with the requested type, or nullopt if the source proxy is nullopt.
+    /// @remark This is a purely local operation.
     template<typename Prx, std::enable_if_t<std::is_base_of_v<ObjectPrx, Prx>, bool> = true>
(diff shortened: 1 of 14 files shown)


cpp/include/Ice/ProxyFunctions.h
```diff
@@ -103,22 +104,22 @@ namespace Ice
         }
     }
 
-    /// Downcasts a proxy after confirming the target object's type via a remote invocation.
+    /// Creates a new proxy from an existing proxy after confirming the target object's type via a remote invocation.
     /// @param proxy The source proxy (can be nullopt).
     /// @param context The request context.
-    /// @return A proxy with the requested type, or nullopt if the source proxy is nullopt or if the target object does
```
bernardnormier, line 109: I removed the "if the source proxy is nullopt" for several reasons:
- arguably, if the source proxy is nullopt, the target doesn't implement the requested type!
- the return sentence is much more readable without this "if"
- in general, you should not make remote invocations on null proxy. We only keep this option for backwards compatibility with the old API.

swift/src/Ice/Proxy.swift
```diff
@@ -315,21 +315,20 @@ public func makeProxy(communicator: Ice.Communicator, proxyString: String, type:
     try communicator.makeProxyImpl(proxyString) as ObjectPrxI
 }
 
-/// Casts a proxy to `Ice.ObjectPrx`. This call contacts the server and will throw an Ice run-time exception
-/// if the target object does not exist or the server cannot be reached.
+/// Creates a new proxy from an existing proxy after confirming the target object's type via a remote invocation.
 ///
-/// - parameter prx: `Ice.ObjectPrx` - The proxy to cast to `Ice.ObjectPrx`.
+/// - parameter prx: `Ice.ObjectPrx` - The source proxy.
```
bernardnormier, line 320: Note that the Swift doc-comments format still needs to be updated. They are currently using an odd mix of Python/MATLAB doc-comment style with Swift flavoring.

python/python/Ice/ObjectPrx.py
```diff
@@ -129,9 +131,9 @@ class ObjectPrx(IcePy.ObjectPrx):
     """
 
     @staticmethod
-    def uncheckedCast(proxy: ObjectPrx, facet: str | None = None) -> ObjectPrx | None:
+    def uncheckedCast(proxy: ObjectPrx | None, facet: str | None = None) -> ObjectPrx | None:
```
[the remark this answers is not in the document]
bernardnormier: This matches the 3.7 behavior and the documentation.

externl (approved):

swift/src/Ice/Proxy.swift
InsertCreativityHere: There's an open issue tracking this: #2274.
It's fixed in the generated code, but the hand-written code still needs fixing.

cpp/include/Ice/ProxyFunctions.h
```diff
@@ -30,22 +30,20 @@ namespace Ice
         }
     }
 
-    /// Downcasts a proxy without confirming the target object's type via a remote invocation.
```
InsertCreativityHere, line 33: I think we should keep this comment (but with the better wording). Otherwise users need to read `checkedCast` first for the context to understand `uncheckedCast`. It's obvious to us, but we're not the target for these doc-comments.

InsertCreativityHere (approved):

PR #4362, opened 2025-08-28
Cleanup "database environment"

bernardnormier:
This PR removes most mentions of database environment + misc other cleanups, mostly in IceGrid.

### DIFF
--- a/cpp/src/IceStorm/Service.cpp
+++ b/cpp/src/IceStorm/Service.cpp
@@ -46,8 +46,6 @@ namespace
         void stop() final;
 
     private:
-        void createDbEnv(const Ice::CommunicatorPtr&);
-
         std::shared_ptr<IceStorm::TopicManagerImpl> _manager;
         std::shared_ptr<IceStorm::TransientTopicManagerImpl> _transientManager;
         std::optional<IceStorm::TopicManagerPrx> _managerProxy;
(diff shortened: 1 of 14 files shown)


cpp/src/IceStorm/Service.cpp
```diff
@@ -46,8 +46,6 @@ namespace
         void stop() final;
 
     private:
-        void createDbEnv(const Ice::CommunicatorPtr&);
```
bernardnormier, line 49: This function was not implemented.

cpp/src/IceStorm/Service.cpp
```diff
@@ -103,9 +101,6 @@ ServiceI::start([[maybe_unused]] const string& serviceName, const CommunicatorPt
     auto topicAdapter = communicator->createObjectAdapter("IceStorm.TopicManager");
     auto publishAdapter = communicator->createObjectAdapter("IceStorm.Publish");
 
-    //
-    // We use the name of the service for the name of the database environment.
```
bernardnormier, line 107: This is not true, and LMDB environment don't have a "name", just a path. Also, unrelated to the code below.

csharp/src/iceboxnet/ServiceManagerI.cs
```diff
@@ -692,8 +692,8 @@ private void stopAll()
             }
 
             //
-            // For each service, we call stop on the service and flush its database environment to
```
bernardnormier, line 695: Odd comment about flushing database environments. An IceBox service may or may not use a database.  And IceBox itself doesn't maintain any database.

InsertCreativityHere (approved):
Looks good to me.

PR #4368, opened 2025-09-01
Remove secure proxy option and PreferSecure proxy property

bernardnormier:
This PR removes the `secure` proxy option (`-s` in stringified proxies) and the `PreferSecure` proxy property.

Also removes all associated properties and proxy methods.

This option and property were about dealing with proxy with both tcp and ssl endpoints (or ws and wss endpoints).
We don't recommend creating such proxies and as a result we shouldn't provide APIs, properties (etc.) to fine-tune the handling of such proxies.


cpp/src/Glacier2/RoutingTable.cpp
```diff
@@ -65,7 +65,7 @@ Glacier2::RoutingTable::add(
             throw ObjectNotExistException{__FILE__, __LINE__};
         }
 
-        ObjectPrx proxy = prx->ice_twoway()->ice_secure(false)->ice_facet(""); // We add proxies in default form.
+        ObjectPrx proxy = prx->ice_twoway()->ice_facet(""); // We add proxies in default form.
```
pepone: I will write it as:

```
proxies.push_back(prx->ice_twoway()->ice_facet(""));
```

pepone (approved):
Looks good!

js/test/Ice/info/Client.ts
```diff
@@ -47,11 +47,6 @@ export class Client extends TestHelper {
         test(ipEndpoint.sourceAddress == "10.10.10.10");
         test(ipEndpoint.compress);
         test(!ipEndpoint.datagram());
-        test(
```
externl: As far as I can tell this section of test still needs to be removed from C#, Java, and Python. 

Although do we really want to remove this tests though? We still have this API on Endpoint.

bernardnormier: You're right, I should keep this test.

cpp/src/Glacier2/RoutingTable.cpp
bernardnormier: Fixed as suggested.

externl (approved):

PR #4372, opened 2025-09-03
Turn on SEPARATE_MEMBER_PAGES in Doxygen

bernardnormier:
This generates a separate page for each enum and other module/namespace members.

### DIFF
--- a/doxygen/Doxyfile
+++ b/doxygen/Doxyfile
@@ -266,7 +266,7 @@ INHERIT_DOCS           = YES
 # of the file/class/namespace that contains it.
 # The default value is: NO.
 
-SEPARATE_MEMBER_PAGES  = NO
+SEPARATE_MEMBER_PAGES  = YES
 
 # The TAB_SIZE tag can be used to set the number of spaces in a tab. Doxygen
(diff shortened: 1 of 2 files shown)


externl (approved):

doxygen/Doxyfile
```diff
@@ -893,7 +893,7 @@ WARN_NO_PARAMDOC       = NO
 # will automatically be disabled.
 # The default value is: NO.
 
-WARN_IF_UNDOC_ENUM_VAL = NO
+WARN_IF_UNDOC_ENUM_VAL = YES
```
InsertCreativityHere, line 896: From the explanation, it seems meaningless to set this to `YES`, because we set `EXTRACT_ALL = YES` for Slice, and
> If EXTRACT_ALL is set to YES then this flag will automatically be disabled.

InsertCreativityHere: Or maybe the most correct fix is to keep this setting as `YES` and set `EXTRACT_ALL = NO`.
Not completely clear what effect this may have on the Doxygen Slice code though,

InsertCreativityHere (approved):

Issue #4374, opened 2025-09-03
Bogus JavaScript generated code for `js:identifier`


pepone:
When you remap a module to a nested module with `js:identifier` it generates bogus code like:

```
export const Remote.Clock = {};
```

This should generate code similar to what we generate for a nested module:

```
export const Remote = {};
Remote.Clock = {};
...
```

Complete example:

```
jose@studio greeter % cat M1.ice 
// module Time becomes namespace remote.clock in JavaScript.
["js:identifier:Remote.Clock"]
module Time
{
    interface AlarmClock {
        void ring();
    }
}
jose@studio greeter % 
jose@studio greeter % 
jose@studio greeter % npx slice2js M1.ice
jose@studio greeter % cat M1.js 
// Copyright (c) ZeroC, Inc.

// slice2js version 3.8.0-alpha.0
// <auto-generated>Generated from Slice file 'M1.ice'.</auto-generated>

/* eslint-disable */
/* jshint ignore: start */

import { Ice } from "@zeroc/ice";

export const Remote.Clock = {};

const iceC_Remote.Clock_AlarmClock_ids = [
    "::Ice::Object",
    "::Time::AlarmClock"
];

Remote.Clock.AlarmClock = class extends Ice.Object {};

Remote.Clock.AlarmClockPrx = class extends Ice.ObjectPrx {};

Ice.TypeRegistry.declareProxyType("Remote.Clock.AlarmClockPrx", Remote.Clock.AlarmClockPrx);

Ice.defineOperations(
    Remote.Clock.AlarmClock,
    Remote.Clock.AlarmClockPrx,
    iceC_Remote.Clock_AlarmClock_ids,
    "::Time::AlarmClock", {
        "ring": [, 0, , , , , , , ]
    });
```

pepone:
Fixed in #4376

PR #4376, opened 2025-09-04
Fix JavaScript generated code for nested modules

pepone:
#4374


bernardnormier (approved):
We should add a test as well.

externl (approved):

PR #4377, opened 2025-09-04
Update Ice.loadSlice to take arguments as a list of strings

pepone:
This PR updates Ice.loadSlice to take arguments as a list of strings. Previously it took to arguments and string "cmd" and the args, but both were used for the same purpose as show in the docs:

Previously all these were equivalent:

```
Ice.loadSlice("-I/opt/IcePy/slice Color.ice")
Ice.loadSlice("-I/opt/IcePy/slice", ["Color.ice"])
Ice.loadSlice("", ["-I/opt/IcePy/slice", "Color.ice"])
```

With this PR you should write:

```
Ice.loadSlice(["-I/opt/IcePy/slice", "Color.ice"])
```

### DIFF
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -962,6 +962,20 @@ initialization. See `InitializationData.pluginFactories`.
   For example, `["python:memoryview:CustomFactory.myByteSeq:array.array"]` indicates that the factory function
   will return an `array.array`.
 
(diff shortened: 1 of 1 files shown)


bernardnormier (approved):
You should add an entry to the Changelog.

Do you need to make a similar update to Ruby in a follow-up PR?

PR #4385, opened 2025-09-05
Js dict mapping

pepone:
This PR updates the mapping for JavaScript dictionaries with a long key type. 
Since Slice long is now mapped to BigInt, we can use the built-in JavaScript Map type.
Previously, Ice.HashMap was required because Ice.Long objects could not be used as keys in a Map.

### DIFF
--- a/js/test/Ice/operations/Twoways.ts
+++ b/js/test/Ice/operations/Twoways.ts
@@ -7,6 +7,16 @@ import { TestHelper, test } from "../../Common/TestHelper.js";
 const int64MaxValue = 9223372036854775807n;
 const int64MinValue = -9223372036854775808n;
 
+function floatEquals(a: number | undefined, b: number | undefined): boolean {
+    if (a === undefined && b === undefined) {
+        return true;
+    } else if (a === undefined || b === undefined) {
+        return false;
+    } else {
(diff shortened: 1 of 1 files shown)


cpp/src/slice2js/Gen.cpp
```diff
@@ -2046,7 +2046,7 @@ Slice::Gen::TypeScriptVisitor::typeToTsString(const TypePtr& type, bool nullable
         const TypePtr keyType = d->keyType();
         builtin = dynamic_pointer_cast<Builtin>(keyType);
         ostringstream os;
-        if ((builtin && builtin->kind() == Builtin::KindLong) || dynamic_pointer_cast<Struct>(keyType))
+        if (dynamic_pointer_cast<Struct>(keyType))
```
bernardnormier, line 2049: Do we still need Ice.HashMap? Why not use Map all the time?

Map supports object keys.

pepone: The difference is the key equality, Ice.HashMap uses object.equals, and Map uses `===` operator.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#key_equality

We still need this for using struct as keys.

externl (approved):

PR #4386, opened 2025-09-05
Fix bugs in ice2slice

bernardnormier:
This PR fixes a few small bugs in ice2slice.

### DIFF
--- a/cpp/src/ice2slice/Gen.cpp
+++ b/cpp/src/ice2slice/Gen.cpp
@@ -412,17 +412,13 @@ Gen::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     out << nl << "interface " << p->name();
     if (bases.size() > 0)
     {
-        out << " :";
-        for (auto q = bases.begin(); q != bases.end();)
+        out << " : ";
+        out.spar("");
+        for (const auto& base : bases)
         {
-            InterfaceDefPtr base = *q;
-            out << " " << getUnqualified(base, scope);
-            q++;
-            if (q != bases.end())
-            {
-                out << ",";
-            }
+            out << getUnqualified(base, scope);
         }
+        out.epar("");
     }
     out << " {";
     out.inc();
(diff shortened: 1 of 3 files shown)


InsertCreativityHere (approved):
All looks good to me!

cpp/src/ice2slice/Gen.cpp
```diff
@@ -448,113 +412,147 @@ Gen::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
...
-        out << spar;
-        for (const auto& param : op->inParameters())
-        {
-            out << paramToString(param, scope);
-        }
-        out << epar;
+    writeDocComment(op, out);
+    // We don't write cs::identifier for operations because Original Slice often uses cs:identifier to get
+    // proper casing for the mapped C# method. New Slice already generates the correct casing.
+    // TODO: should we issue a warning for this omission? See writeCsIdentifier.
```
pepone, line 444: What about checking if the Ice Slice identifier matches the default IceRPC identifier and omit the metadata if it is the same. We can do it in a follow up PR.

pepone (approved):

cpp/src/ice2slice/Gen.cpp
bernardnormier: I would rather not do this for several reasons:

1) It's complicated, and the user would most likely not understand what's going on

2) We already have two different "ToPascalCase" algorithm in IceRPC, one in Rust and another one in C#.

https://github.com/icerpc/icerpc-csharp/blob/f109b98c1cb69755ccca7ddf3cc247982e7afb8c/tools/Common/IceRpc.CaseConverter.Internal/Converter.cs#L9

https://github.com/icerpc/icerpc-csharp/blob/f109b98c1cb69755ccca7ddf3cc247982e7afb8c/tools/slicec-cs/src/cs_util.rs#L117

A third version in C++ isn't appealing.

3) Other than proper casing, the main other reason to use `cs:identifier` on a field/method name is to escape keywords / avoid collisions with other names. Collisions are less likely in IceRPC C# since it's using Pascal case and adding Async to method names.

PR #4388, opened 2025-09-05
Fix Slice Compiler Hang when Preprocessor Returns Non-Zero Status

### LINKED ISSUE #4387: Slice compiler hang when missing -I
I can reproduce it with both slice2cpp and ice2slice:

```shell
slice2cpp IceStorm.ice 
/Users/bernard/builds/ice/slice/IceStorm/IceStorm.ice:13: error: Can't open include file "Ice/Identity.ice"
    #include "Ice/Identity.ice"
/Users/bernard/builds/ice/slice/IceStorm/Metrics.ice:13: error: Can't open include file "Ice/Metrics.ice"
    #include "Ice/Metrics.ice"
    from /Users/bernard/builds/ice/slice/IceStorm/IceStorm.ice: 14:    #include "Metrics.ice"
2 errors in preprocessor. <= hangs forever
```

InsertCreativityHere:
This PR fixes #4387.

Our entry point into MCPP is `preprocessor->preprocess` which either returns a handle if everything worked, or `nullptr` if it failed to preprocess. In the past we correctly checked for `nullptr` and exited early, but this was removed during the MCPP refactoring 2 months ago. Instead this check was replaced with an `assert`, which as far as I can tell is incorrect here.

If you replicate #4387 with a debug build, you'll see the assert hit on this line.

### DIFF
--- a/cpp/src/ice2slice/Main.cpp
+++ b/cpp/src/ice2slice/Main.cpp
@@ -167,7 +167,10 @@ compile(const vector<string>& argv)
         {
             PreprocessorPtr preprocessor = Preprocessor::create(argv[0], fileName, preprocessorArgs);
             FILE* preprocessedHandle = preprocessor->preprocess("-D__ICE2SLICE__");
-            assert(preprocessedHandle);
+            if (preprocessedHandle == nullptr)
+            {
+                return EXIT_FAILURE;
+            }
 
             unit = Unit::createUnit("", false);
             int parseStatus = unit->parse(fileName, preprocessedHandle, debug);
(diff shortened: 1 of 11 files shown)


cpp/src/ice2slice/Main.cpp
```diff
@@ -167,7 +167,10 @@ compile(const vector<string>& argv)
         {
             PreprocessorPtr preprocessor = Preprocessor::create(argv[0], fileName, preprocessorArgs);
             FILE* preprocessedHandle = preprocessor->preprocess("-D__ICE2SLICE__");
-            assert(preprocessedHandle);
+            if (preprocessedHandle == nullptr)
```
pepone, line 170: I think it would be nicer to fix `preprocess` to throw instead of returning null.

InsertCreativityHere: I'd rather keep it the way it is, since `preprocess` already throws for other reasons.
In the compilers, there's always 2 failure states:
1) A user made a mistake (like forgetting `-I`). Here, the Slice compilers are working correctly, there's no exceptional state. So we return `nullptr` to signal: "everything is working correctly, but the user messed up"
2) There is a problem within the preprocessor itself (i.e. a bug). This is when `preprocess` will throw, since it's no longer working how we expect it to.

PR #4390, opened 2025-09-09
Update build instructions for Linux RPM based distributions

pepone:


cpp/BUILDING.md
```diff
@@ -68,6 +68,28 @@ Install the required third-party libraries:
 dnf install bzip2-devel expat-devel libedit-devel libsystemd-devel lmdb-devel mcpp-devel openssl-devel
 ```
 
+#### On RHEL 10
```
bernardnormier: I would combine RHEL9 and RHEL 10 in a single section.

cpp/BUILDING.md
```diff
@@ -68,6 +68,28 @@ Install the required third-party libraries:
...
+
+Add the ZeroC repository:
```
bernardnormier: It's not clear to me why we need this repo, especially since we apparently don't need it for Amazon.
Would be nice to clarify, e.g.:

Add the ZeroC repository (for mcpp-devel)

pepone: RHEL doesn't provide the mcpp and lmdb devel packages

PR #4393, opened 2025-09-09
Simplify sample service config files

bernardnormier:
This PR simplifies the sample service config files in cpp/config. These are config files for Windows; we should probably move them somewhere else.

### DIFF
--- a/cpp/config/icegridnode.cfg
+++ b/cpp/config/icegridnode.cfg
@@ -1,67 +1,38 @@
 #
-# Sample configuration file for the IceGrid node service
+# Sample configuration file for the IceGrid node (node1).
 #
 
-#
-# Proxy to the IceGrid registry - TCP is used by default.
-#
-Ice.Default.Locator=DemoIceGrid/Locator:tcp -h localhost -p 4061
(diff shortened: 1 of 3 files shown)


pepone (approved):

cpp/config/icegridnode.cfg
```diff
@@ -1,67 +1,38 @@
...
-# The directory where the IceGrid node maintains its data (such as server
-# configuration files). This directory must already exist when the node
-# starts.
-#
-# The path name below is merely a suggestion, based on the assumption
-# that the node is running under the LocalService account.
-#
+# The path name below is merely a suggestion, based on the assumption that the node is running under the
+# LocalService account.
 IceGrid.Node.Data=C:\Windows\ServiceProfiles\LocalService\AppData\Local\ZeroC\icegrid\node1
```
externl, line 21: I wonder if we should use something more generic for the folder name than copying the node's name.

Similar to the default registry folder name (`registry`):  `C:\Windows\ServiceProfiles\LocalService\AppData\Local\ZeroC\icegrid\node`

externl (approved):

PR #4394, opened 2025-09-09
IceGrid interop/upgrade fixes

bernardnormier:
This PR makes 3 related fixes:
 - adds back the DbEnvDescriptor for database schema compatibility with 3.7
   - this also provides better interop with the 3.7 admin tools, and user-written admin tools, but we don't really care about these.
 - remove the InternalDbEnvDescriptor (and more) from Internal.ice to break even more the registry<-> node interop when mixing 3.7 and 3.8
   - we don't want this interop, and it didn't interop before this PR either
- update the CHANGELOG to explain all this.

### DIFF
--- a/slice/IceGrid/Descriptor.ice
+++ b/slice/IceGrid/Descriptor.ice
@@ -116,6 +116,7 @@ module IceGrid
         string dbHome;
 
         /// The configuration properties of the database environment.
+        ["matlab:identifier:propertyDescriptors"]
         PropertyDescriptorSeq properties;
     }
 
--- a/slice/IceGrid/Descriptor.ice
+++ b/slice/IceGrid/Descriptor.ice
@@ -102,6 +102,27 @@ module IceGrid
     ["java:type:java.util.LinkedList<AdapterDescriptor>"]
     sequence<AdapterDescriptor> AdapterDescriptorSeq;
 
+    /// A Freeze database environment descriptor (deprecated, no longer used).
+    ["deprecated:This descriptor is provided for schema compatibility. It is no longer used as of Ice 3.8."]
+    struct DbEnvDescriptor
+    {
+        /// The name of the database environment.
+        string name;
+
+        /// The description of this database environment.
+        string description;
+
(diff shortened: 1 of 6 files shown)


bernardnormier:
> > we want this interop, and it didn't interop before this PR either
> 
> Do you mean **we don't want**?

Yes, fixed.

slice/IceGrid/Descriptor.ice
```diff
@@ -102,6 +102,27 @@ module IceGrid
...
+        /// The home of the database environment.
+        string dbHome;
+
+        /// The configuration properties of the database environment.
+        PropertyDescriptorSeq properties;
+    }
+
+    /// A sequence of {@link DbEnvDescriptor}.
+    ["deprecated:This descriptor is provided for schema compatibility. It is no longer used as of Ice 3.8."]
+    sequence<DbEnvDescriptor> DbEnvDescriptorSeq;
```
InsertCreativityHere, line 125: Why not keep the `["java:type:java.util.LinkedList<DbEnvDescriptor>"]` metadata that used to be on this?

InsertCreativityHere (approved):

pepone (approved):

slice/IceGrid/Descriptor.ice
bernardnormier: Because this type is only kept for schema compatibility, and the database schema code in pure C++.

externl (approved):

slice/IceGrid/Descriptor.ice
```diff
@@ -102,6 +102,27 @@ module IceGrid
...
+        string description;
+
+        /// The home of the database environment.
+        string dbHome;
+
+        /// The configuration properties of the database environment.
+        PropertyDescriptorSeq properties;
+    }
+
+    /// A sequence of {@link DbEnvDescriptor}.
```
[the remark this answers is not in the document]
bernardnormier: This is nonsense. It's a Slice file and the only link syntax is the one I used.

PR #4404, opened 2025-09-10
Turn-off connection abort warnings in C++ tests that abort connections

bernardnormier:


cpp/test/IceStorm/stress/Subscriber.cpp
```diff
@@ -235,7 +235,14 @@ class Subscriber final : public Test::TestHelper
 void
 Subscriber::run(int argc, char** argv)
 {
-    Ice::CommunicatorHolder ich = initialize(argc, argv, make_shared<Ice::Properties>(vector<string>{"IceStormAdmin"}));
+    InitializationData initData;
+    initData.properties = make_shared<Ice::Properties>(vector<string>{"IceStormAdmin"});
+    initData.properties = createProperties(argc, argv, initData.properties);
```
bernardnormier, line 240: I could not figure out a simpler way to do this. The API isn't great.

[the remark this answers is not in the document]
bernardnormier: I am calling `Ice::createProperties` here. And yes, it's confusing.

externl (approved):

cpp/test/IceStorm/stress/Subscriber.cpp
pepone: Maybe another `createTestProperties` overload that takes a defaults (Properties)

pepone (approved):

cpp/test/IceStorm/stress/Subscriber.cpp
bernardnormier: So far we only need it once for this test.

PR #4406, opened 2025-09-11
Fix propertyToProxy return type in TypeScript and Python declarations

pepone:
Additionally fix the doc comment to be consitent accross the language mappings.

### DIFF
--- a/csharp/src/Ice/Communicator.cs
+++ b/csharp/src/Ice/Communicator.cs
@@ -139,8 +139,7 @@ public sealed class Communicator : IDisposable, IAsyncDisposable
 
     /// <summary>
     /// Converts a set of proxy properties into a proxy. The "base" name supplied in the <paramref name="property" />
-    /// argument refers to a property containing a stringified proxy, such as
-    /// <code>MyProxy=id:tcp -h localhost -p 10000</code>.
+    /// argument refers to a property containing a stringified proxy, such as `MyProxy=id:tcp -h localhost -p 10000`.
     /// Additional properties configure local settings for the proxy.
     /// </summary>
     ///


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -127,17 +127,16 @@ public String proxyToString(ObjectPrx proxy) {
...
-     * MyProxy=id:tcp -h localhost -p 10000}. Additional properties configure local settings
-     * for the proxy.
+     * Converts a set of proxy properties into a proxy. The "base" name supplied in the {@code property} argument
+     * refers to a property containing a stringified proxy, such as {@code MyProxy=id:tcp -h localhost -p 10000}.
+     * Additional properties configure local settings for the proxy.
      *
-     * @param prefix The base property name.
-     * @return The proxy, or null if the property is not set.
+     * @param property The base property name.
+     * @returns The proxy, or null if the property is not set.
```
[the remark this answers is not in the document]
externl: I believe Copilot is correct here.

python/python/Ice/Communicator.py
```diff
@@ -191,12 +191,11 @@ def proxyToString(self, proxy: ObjectPrx | None) -> str:
         """
         return self._impl.proxyToString(proxy)
 
-    def propertyToProxy(self, property: str) -> ObjectPrx:
+    def propertyToProxy(self, property: str) -> ObjectPrx | None:
```
externl, line 194: Do we not test this anywhere in Python?

pepone: Ice/proxy does, not sure why the linter didn't complain with this new signature. For TypeScript I have to fix the test.

PR #4416, opened 2025-09-12
Fix IceGridGUI bug and Java Properties._clone bug

bernardnormier:
Move the undocumented "AdminRouter" config to IceGridAdmin.Router.

### DIFF
--- a/config/PropertyNames.xml
+++ b/config/PropertyNames.xml
@@ -227,6 +227,8 @@
         <property name="Password" languages="cpp" />
         <property name="Port" languages="cpp" />
         <property name="Replica" languages="cpp" />
+        <!-- An undocumented internal object adapter used by IceGridGUI. -->
+        <property name="Router" class="ObjectAdapter" languages="java" />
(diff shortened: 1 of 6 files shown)


config/PropertyNames.xml
```diff
@@ -227,14 +227,15 @@
         <property name="Password" languages="cpp" />
         <property name="Port" languages="cpp" />
         <property name="Replica" languages="cpp" />
+        <!-- An undocumented internal object adapter used by IceGridGUI. -->
+        <property name="Router" class="ObjectAdapter" languages="java" />
```
externl: "Router" is a bit overloaded as we generally associate it with a Proxy property class.

```
<property name="Router" class="Proxy" languages="all" />
```

Maybe we call it something like `Server`. Like the `--server` command line option we use.

externl (approved):

config/PropertyNames.xml
bernardnormier: Renamed to Internal. It's fairly unrelated to the icegridadmin `--server` option.

PR #4419, opened 2025-09-15
Rename MSI installer to IceServices-$(Version).msi

### LINKED ISSUE #4411: Ice MSI installs Ice services
Now that the MSI installs only the services and admin tools for these services, we should rework the project to make this clear, e.g.:

- rename the MSI file to IceServices.msi or similar
- in the dialogs, make clear you're about to install the Ice services on your computer, not "Ice"
- same for the folder you select - it's where you install the Ice services

The project name and other internal components can remain "Ice". This issue is about the public presentation.

pepone:
Fix #4411

### DIFF
--- a/.github/workflows/publish-misc-packages.yml
+++ b/.github/workflows/publish-misc-packages.yml
@@ -39,7 +39,7 @@ jobs:
           set -euo pipefail
           if [ "${CHANNEL}" = "nightly" ]; then
             # Rename the MSI file to Ice-nightly.msi if it's the nightly channel
-            aws s3 cp staging/windows-msi/*.msi "s3://zeroc-downloads/ice/${CHANNEL}/Ice-nightly.msi"
+            aws s3 cp staging/windows-msi/*.msi "s3://zeroc-downloads/ice/${CHANNEL}/Ice-Services-nightly.msi"
           else
             aws s3 cp staging/windows-msi/*.msi "s3://zeroc-downloads/ice/${CHANNEL}/"
           fi
--- a/packaging/msi/IceServices.wixproj
+++ b/packaging/msi/IceServices.wixproj
@@ -15,8 +15,8 @@
     <DefineConstants>$(DefineConstants);IceManualUrl=https://docs.zeroc.com/ice/latest/cpp/</DefineConstants>
     <DefineConstants>$(DefineConstants);IceSourceUrl=https://github.com/zeroc-ice/ice</DefineConstants>
     <DefineConstants>$(DefineConstants);IceReleaseNotesUrl=https://docs.zeroc.com/ice/latest/cpp/ice-3-8-0-preview</DefineConstants>
-    <OutputName>Ice-$(IcePackageVersion)</OutputName>
-    <ProductCode>Ice-$(Version)</ProductCode>
+    <OutputName>Ice-Services-$(IcePackageVersion)</OutputName>
+    <ProductCode>Ice-Services-$(Version)</ProductCode>
   </PropertyGroup>
   <ItemGroup>
     <BindPath Include="$(MSBuildThisFileDirectory)..\..\" BindName="SOURCE_ROOT" />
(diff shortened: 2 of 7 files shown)


packaging/msi/IceServices.wixproj
```diff
@@ -40,8 +39,6 @@
       Text="The IceGridGUI must be build before creating the MSI package." />
     <RemoveDir Directories="$(StagingDir)" />
     <MakeDir Directories="$(StagingDir)bin" />
-    <MakeDir Directories="$(StagingDir)slice" />
-    <Copy SourceFiles="@(Slices)" DestinationFolder="$(StagingDir)slice\%(Slices.RecursiveDir)" />
```
pepone, line 44: We don't need to include Slice files with the MSI anymore.

packaging/msi/StartMenu.wxs
```diff
@@ -14,24 +14,14 @@
...
-                <util:InternetShortcut
-                    Id="ManualLink"
-                    Name="Ice Manual"
-                    Type="url"
-                    Target="$(IceManualUrl)" />
-                <util:InternetShortcut
-                    Id="SourceCodeLink"
-                    Name="Source Code on Github"
-                    Type="url"
-                    Target="$(IceSourceUrl)" />
```
pepone, line 26: We Just keep the IceGridGUI shortcut in start menu.

.github/workflows/publish-misc-packages.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           set -euo pipefail
           if [ "${CHANNEL}" = "nightly" ]; then
             # Rename the MSI file to Ice-nightly.msi if it's the nightly channel
-            aws s3 cp staging/windows-msi/*.msi "s3://zeroc-downloads/ice/${CHANNEL}/Ice-nightly.msi"
+            aws s3 cp staging/windows-msi/*.msi "s3://zeroc-downloads/ice/${CHANNEL}/Ice-Services-nightly.msi"
```
externl, line 42: Should we have the full version that includes the date?

packaging/msi/StartMenu.wxs
bernardnormier: I am wondering if it should say:
"IceGrid GUI 3.8 Preview"
to distinguish it from
"IceGrid GUI 3.8.0"
"IceGrid GUI 3.8.4"
"IceGrid GUI 3.9.0"

I assume you can install all these versions at the same time on the same Windows system?

Or maybe the installer can / should prevent the installation of several 3.8 versions of the Ice services?

packaging/msi/Package.wxs
```diff
@@ -4,16 +4,16 @@
...
         InstallerVersion="500"
         Compressed="yes"
-        Id="Ice_$(Version)">
+        Id="Ice_Services_$(Version)">
 
         <SummaryInformation
-            Description="Ice $(Version)"
-            Comments="Ice $(Version) MSI Installer"
+            Description="Ice Services $(Version)"
+            Comments="Ice Services $(Version) MSI Installer"
```
bernardnormier, line 16: I would include the service names (IceGrid, IceStorm, ...) in the keywords.

Should all the keywords be lowercase?

packaging/msi/StartMenu.wxs
externl: I think it would be nice if we actually included the full nightly version

.github/workflows/publish-misc-packages.yml
pepone: You mean keep both in s3, we need this one without versions for the links, what is the point of the versioned one if we don't have any links?

externl: Hmm.. fair enough point about the links, they would be impossible to link to. 

I think  showing the specific nightly version in the installer would be useful. We don't do this for anything else at the moment either (e.g. in `slice2* --version`) so it's fine, but maybe worth considering in the future.

packaging/msi/Package.wxs
pepone: https://docs.firegiant.com/wix/api/wixtoolsetdtfwindowsinstaller/summaryinfo/#keywords

`Installer` is included by default, doesn't mention anything about case.

packaging/msi/StartMenu.wxs
pepone: I adding "IceGrid GUI 3.8.0 Preview" for now.

Regarding multiple install I opened an issue in Wix repository a while ago but didn't get much help: https://github.com/orgs/wixtoolset/discussions/9063

PR #4422, opened 2025-09-16
Support for Xcode 26

externl:
This PR adds "support" for macOS 26, and more importantly, Xcode 26. The only issue I ran into was with the Swift tests. 

```
| actor PreservedCounter {
 7 |     static var counter: Int32 = 0
   |                |- error: static property 'counter' is not concurrency-safe because it is nonisolated global shared mutable state [#MutableGlobalVariable]
   |                |- note: convert 'counter' to a 'let' constant to make 'Sendable' shared state immutable
   |                |- note: add '@MainActor' to make static property 'counter' part of global actor 'MainActor'
   |                `- note: disable concurrency-safety checks if accesses are protected by an external synchronization mechanism
 8 | }
```

Turns out we were not even testing this as there's no GC in Swift.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
             python_tests: true
 
           # iOS
-          - os: macos-15
+          - os: macos-26
             config: "ios"
             build_flags: "OPTIMIZE=no PLATFORMS='macosx iphonesimulator' CONFIGS='static shared' LANGUAGES='cpp swift'"
             # TODO swift/Ice/udp and swift/Ice/location are failing regularly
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       matrix:
         include:
           # Release builds
-          - os: macos-15
+          - os: macos-26
```
bernardnormier, line 24: Should we also test on macos-15?

bernardnormier (approved):

.github/workflows/ci.yml
externl: I don't think we need to bother. And we already have a limited number of macOS runners.

PR #4430, opened 2025-09-16
Return nil for unsupported admin facets in Swift

### LINKED ISSUE #4429: Return nil for unsupported admin facets in Swift
We should update Swift to be consistent with Python.

See #4423.

externl:
Fixes #4429

### DIFF
--- a/swift/src/IceImpl/Communicator.mm
+++ b/swift/src/IceImpl/Communicator.mm
@@ -341,8 +341,12 @@
             return [NSNull null];
         }
 
-        id dispatchAdapter = [self facetToDispatchAdapter:servant];
-        return dispatchAdapter ? dispatchAdapter : [NSNull null];
+        if (id dispatchAdapter = [self facetToDispatchAdapter:servant])
+        {
+            return dispatchAdapter;
+        }
+
+        return [NSNull null];
     }
     catch (...)
     {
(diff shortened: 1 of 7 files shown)


swift/src/IceImpl/Communicator.mm
```diff
@@ -407,7 +409,7 @@ - (ICEProperties*)getProperties
         return [factory createProperties:self handle:propertiesAdminHandle];
     }
 
-    return [factory createUnsupported:self];
+    return [NSNull null];
```
bernardnormier: This function is private to this file, it's not a bridging function.

We should prefer nil over NSNull in such ObjC code, and return NSNull only in bridging functions that return a nullable id where nil means "the function failed, see error".

So here, I would return nil and review all the callers (bridging functions) and convert in those functions nil to NSNull when appropriate.

swift/src/IceImpl/Communicator.mm
```diff
@@ -376,12 +381,9 @@ - (ICEProperties*)getProperties
     return [ICEProperties getHandle:props];
 }
 
-- (id<ICEDispatchAdapter>)facetToDispatchAdapter:(const Ice::ObjectPtr&)servant
+- (id)facetToDispatchAdapter:(const Ice::ObjectPtr&)servant
```
bernardnormier: I would keep `id<ICEDispatchAdapter>` and return an ICEDispatchAdapter or nil.

PR #4435, opened 2025-09-18
Fix Slice tools to correctly compute include search path

pepone:
This PR fixes a bug in the Slice tools tasks that results in bogus include path used when there are multiple IncludeDirectories.

### DIFF
--- a/csharp/tools/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
+++ b/csharp/tools/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
@@ -41,7 +41,7 @@ public abstract class SliceCompilerTask : ToolTask
 
     protected abstract string GeneratedExtensions { get; }
 
-    protected virtual Dictionary<string, string> GetOptions()
+    protected virtual Dictionary<string, string> GetOptions(ITaskItem item)
     {
         var options = new Dictionary<string, string>
         {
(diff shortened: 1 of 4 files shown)


csharp/tools/ZeroC.Ice.Slice.Tools/ZeroC.Ice.Slice.Tools.targets
```diff
@@ -79,7 +79,6 @@
         <ItemGroup>
             <_SliceCompileNormalized Include="@(SliceCompile)">
                 <OutputDir>$([MSBuild]::NormalizePath('%(SliceCompile.OutputDir)'))</OutputDir>
-                <IncludeDirectories>$([MSBuild]::NormalizePath('%(SliceCompile.IncludeDirectories)'))</IncludeDirectories>
```
pepone, line 82: I added this NormalizePath in a previous PR, which was causing this bug. It returns a string instead of a list and we end-up with `-Ipath1;path2` instead of `-Ipath1 -Ipath2`.

csharp/tools/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
```diff
@@ -41,17 +41,18 @@ protected virtual string GetGeneratedPath(ITaskItem item, string outputDir, stri
...
-    protected virtual Dictionary<string, string> GetOptions()
+    protected virtual Dictionary<string, string> GetOptions(ITaskItem item)
     {
         var options = new Dictionary<string, string>
         {
             ["IceToolsPath"] = IceToolsPath,
             ["OutputDir"] = OutputDir.TrimEnd('\\')
         };
 
-        if (IncludeDirectories.Length > 0)
```
pepone, line 52: `IncludeDirectories` task property is normalized by MSBuild, and we cannot use it for dependency comparison, the DependTaks uses `item.GetMetadata("IncludeDirectories")` and we need to do the same here for GetOptions to ensure that the options saved to the `.d` file are identical to those computed by the depend task, otherwise we end-up with unnecessary rebuilds when comparing a normalized included path with a non normalized path and they do not match.

PR #4437, opened 2025-09-18
Update IceGrid GUI's icegridadmin execution

### LINKED ISSUE #4413: Add homebrew path to IceGrid GUI on macos
The IceGrid GUI needs icegridadmin to parse XML files.

On macos, the default brew bin directory (`/usr/local/bin`) used to be in the system PATH. But that's no longer true on Apple Silicon. As a result, the IceGrid GUI currently does not find `icegridadmin` when started by a double-click.

externl:
The previous behavior was to rely on the default executable behavior on all platforms. This was fine on Windows (and generally Linux), but broke on Apple Silicon brew installs, as the default brew install is not in the default system PATH.

Unfortunately you can't just modify the PATH env variable of the process builder as it is only inherited by the executable. 

Now we just search the PATH ourselves looking for `icegridadmin`, additionally checking the brew prefix on macOS. If we can't find it we just fallback to calling "icegridadmin".  

Fixes #4413


java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Coordinator.java
```diff
@@ -2188,7 +2188,9 @@ public ApplicationDescriptor parseFile(File file) {
             // Start icegridadmin server
             //
             try {
-                _icegridadminProcess = Runtime.getRuntime().exec("icegridadmin --server");
+                String icegridadmin = Utils.getIceGridAdmin();
+                ProcessBuilder processBuilder = new ProcessBuilder(icegridadmin, "--server");
+                _icegridadminProcess = processBuilder.start();
```
InsertCreativityHere: Could do this instead:
```suggestion
                _icegridadminProcess =
                    Runtime.getRuntime().exec(new String[] {icegridadmin, "--server"});
```
`exec` uses `ProcessBuilder` under the hood so no difference. And it's not like we're actually using any of the `ProcessBuilder` builder methods here.

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Utils.java
```diff
@@ -501,5 +503,47 @@ private static void addSet(
         }
     }
 
-    private Utils() {}
+    /**
+     * Get the path to the icegridadmin executable. Searches the PATH environment variable on Unix-like systems.
+     * On Windows, it simply returns "icegridadmin" since the working directory is searched first.
```
InsertCreativityHere, line 509: > ... and we know `IceGridGUI` is in the same folder as `icegridadmin`.`

Think it's worth saying something like that ^ since searching the working directory first doesn't tell you much.

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Utils.java
```diff
@@ -501,5 +503,47 @@ private static void addSet(
...
+        // Windows checks the working directory first when running an executable. The IceGrid GUI is always started
+        // from the same directory as icegridadmin so don't need to search for anything.
+        if (System.getProperty("os.name").startsWith("Windows")) {
+            return exe;
+        }
+
+        String path = System.getenv("PATH");
+
+        List<String> searchDirs;
+        if (path == null || path.isEmpty()) {
```
InsertCreativityHere: Super minor, but no real reason to check `isEmpty` here, `split` will correctly handle the empty string by returning an empty array.
```suggestion
        if (path == null) {
```

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Utils.java
```diff
@@ -501,5 +503,45 @@ private static void addSet(
...
 
-    private Utils() {}
+    /**
+     * Get the path to the icegridadmin executable. Searches the PATH environment variable on Unix-like systems.
+     * On Windows, it simply returns "icegridadmin" since the working directory is searched first.
+     *
+     * @return the path to the icegridadmin executable
+     */
+    public static String getIceGridAdmin() {
+        String exe = "icegridadmin";
```
InsertCreativityHere: Since this is a constant:
```suggestion
        final String exe = "icegridadmin";
```

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Utils.java
```diff
@@ -501,5 +503,45 @@ private static void addSet(
...
+        // Add the default brew bin directory to the search list on macOS.
+        if (System.getProperty("os.name").startsWith("Mac")) {
+            String brewPath = "/opt/homebrew/bin";
+            if (!searchDirs.contains(brewPath)) {
+                searchDirs.add(brewPath);
+            }
+        }
+
+        for (String dir : searchDirs) {
+            java.io.File f = new java.io.File(dir, exe);
```
InsertCreativityHere: No need to qualify, since you `import java.io.File`:
```suggestion
            File f = new File(dir, exe);
```

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Utils.java
```diff
@@ -501,5 +503,45 @@ private static void addSet(
...
+            if (f.exists() && f.canExecute()) {
```
InsertCreativityHere: `canExecute` can only return true if the file `exists`, so no need to double-check exists here:
```suggestion
            if (f.canExecute()) {
```
(https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/File.html#canExecute())

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Utils.java
```diff
@@ -501,5 +503,45 @@ private static void addSet(
...
+        List<String> searchDirs;
+        if (path == null || path.isEmpty()) {
+            searchDirs = new ArrayList<>();
+        } else {
+            searchDirs = new ArrayList<>(Arrays.asList(path.split(File.pathSeparator)));
+        }
+
+        // Add the default brew bin directory to the search list on macOS.
+        if (System.getProperty("os.name").startsWith("Mac")) {
+            String brewPath = "/opt/homebrew/bin";
```
InsertCreativityHere: Also a constant:
```suggestion
            final String brewPath = "/opt/homebrew/bin";
```

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Utils.java
```diff
@@ -501,5 +503,48 @@ private static void addSet(
...
+
+        // Windows checks the working directory first when running an executable. The IceGrid GUI is always started
+        // from the same directory as icegridadmin so don't need to search for anything.
+        if (System.getProperty("os.name").startsWith("Windows")) {
+            return exe;
+        }
+
+        String path = System.getenv("PATH");
+
+        List<String> searchDirs;
```
InsertCreativityHere: Okay, I think this is the cleanest:
```
List<String> searchDirs = new ArrayList<>();
if (path != null) {
    Collections.addAll(searchDirs, path.split(File.pathSeparator));
}
```
You probably need to `import java.util.Collections;`

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Coordinator.java
```diff
@@ -2188,7 +2188,9 @@ public ApplicationDescriptor parseFile(File file) {
             // Start icegridadmin server
             //
             try {
-                _icegridadminProcess = Runtime.getRuntime().exec("icegridadmin --server");
+                String icegridadmin = Utils.getIceGridAdmin();
+                _icegridadminProcess =
```
bernardnormier: why is this code on 2 lines?

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Utils.java
```diff
@@ -501,5 +504,46 @@ private static void addSet(
...
     }
 
-    private Utils() {}
+    /**
+     * Get the path to the icegridadmin executable. Searches the PATH environment variable on Unix-like systems.
+     * On Windows, it simply returns "icegridadmin" since the working directory is searched first.
+     *
+     * @return the path to the icegridadmin executable
+     */
+    public static String getIceGridAdmin() {
```
bernardnormier: Not clear why this method is public.

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Coordinator.java
bernardnormier: I am referring to lines 2192-3.

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Utils.java
externl: Fixed.

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Coordinator.java
externl: Fixed.

PR #4438, opened 2025-09-18
Reject `#include` Directives that Come After Slice Definitions

InsertCreativityHere:
Our documentation has always stated that this is disallowed, although we have apparently never enforced this.
This PR adds enforcement for this restriction. If we decide **not** to adopt this PR, we should drop our 'restriction' from the docs.

----

This enforcement is done at the scanning phase, because we have little control over what MCPP does.

Note that compiler runs in the this order:
1) preprocessing (Preprocessor.cpp)
2) scanning (Scanner.cpp)
3) parsing (Grammar.cpp)
...


cpp/src/Slice/Parser.cpp
```diff
@@ -4497,6 +4497,12 @@ Slice::Unit::setCurrentFile(std::string currentFile, int lineNumber)
     {
         case Push:
         {
+            DefinitionContextPtr dc = currentDefinitionContext();
+            if (dc && dc->seenDefinition())
+            {
+                error("'#include' directives must come before any Slice definitions in a file");
+            }
+
```
InsertCreativityHere, line 4505: This is the enforcement, in the `setCurrentFile` function.
This function is called by the scanner whenever it is about to enter or exit an included file.
The `Push` case means we're just at the start of a newly included file.

So we check if we've seen any Slice definitions in the current file, and if so, report an error,

cpp/src/Slice/Scanner.cpp
InsertCreativityHere, line 1: Generated Flex code, don't need to read.

cpp/src/Slice/Scanner.l
```diff
@@ -639,9 +639,10 @@ namespace
                     line = line.substr(1);
                 }
             }
-            lineTypeCode = currentUnit->setCurrentFile(line, yylineno);
+            lineTypeCode = currentUnit->setCurrentFile(line, newLineNumber);
```
InsertCreativityHere, line 642: These changes aren't important.
We now delay updating the current line number (`yylineno`) until after `setCurrentFile`.
Otherwise error messages that function emits have incorrect line numbers.

cpp/src/Slice/Parser.cpp
```diff
@@ -4497,6 +4497,12 @@ Slice::Unit::setCurrentFile(std::string currentFile, int lineNumber)
     {
         case Push:
         {
+            DefinitionContextPtr dc = currentDefinitionContext();
+            if (dc && dc->seenDefinition())
+            {
+                error("'#include' directives must come before any Slice definitions in a file");
+            }
```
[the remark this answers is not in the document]
InsertCreativityHere: It's just a restriction that we have. I don't think it needs a justification in the code.
I'm the only person who will ever read this code anyways.

bernardnormier:
> I'd rather not add this new "enforcement".

I didn't understand that the current effect of inclusion in the middle of a Slice file is largely random. So it's unlikely anybody is using this feature.

InsertCreativityHere:
Just to add some context, this is how the code-gen skips included definitions:
```cpp
void
Slice::Container::visitContents(ParserVisitor* visitor)
{
    for (const auto& p : _contents)
    {
        if (visitor->shouldVisitIncludedDefinitions() || p->includeLevel() == 0)
        {
            p->visit(visitor);
        }
    }
}
```
Whenever we visit a container (Module, Class, etc.), we don't visit any deeper into it if we're in an included file (`includeLevel > 0`).

So, if the code-gen relies on `visitDataMember`, `visitOperation`, etc. these are skipped and the generated code broken.
But if the code-gen only uses `visitClass`, etc. and never drills down into the members, the generated code will work.

Which approach we take in the code-gen is kind of random.
Even within a single compiler, some of it's visitors will drill down to the leafs, and others won't.

externl (approved):

PR #4439, opened 2025-09-18
Improve InitializationData initializer in Swift

bernardnormier:
and update some tests to use the new init.

### DIFF
--- a/swift/test/Ice/scope/Client.swift
+++ b/swift/test/Ice/scope/Client.swift
@@ -5,9 +5,8 @@ import TestCommon
 
 public class Client: TestHelperI, @unchecked Sendable {
     override public func run(args: [String]) async throws {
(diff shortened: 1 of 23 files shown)


swift/test/Ice/scope/Client.swift
```diff
@@ -5,10 +5,7 @@ import TestCommon
 
 public class Client: TestHelperI, @unchecked Sendable {
     override public func run(args: [String]) async throws {
-        var initData = Ice.InitializationData()
-        initData.properties = try createTestProperties(args)
-        initData.sliceLoader = DefaultSliceLoader("IceScope")
```
bernardnormier, line 10: Unused since there is no corresponding metadata in the Slice file.

externl (approved):

pepone (approved):

Issue #4441, opened 2025-09-19
Windows segmentation fault in Selector::getNextHandler


pepone:
A user report getting segmentation faults on Windows. When the crash happens the "key" is null in

https://github.com/zeroc-ice/ice/blob/b6c3109b7de5ee3f8f56c561865c007d3f6c056f/cpp/src/Ice/Selector.cpp#L171

Seems the root cause is a missing throw statement in the code above, which constructs a SocketException but never actually throws it.

https://github.com/zeroc-ice/ice/blob/b6c3109b7de5ee3f8f56c561865c007d3f6c056f/cpp/src/Ice/Selector.cpp#L156-L162

In this block the key is expected to be null, and by let the code continue we end up using a null pointer.

pepone:
> [!NOTE]
> AI-assisted note from a static-review audit; verified on `main` (`56f9ec0dd6`). Informational — no action implied.

For the record: an AI audit flagged `Selector::getNextHandler` for "reading the wrong error source" (using `WSAGetLastError()` after `GetQueuedCompletionStatus` rather than `GetLastError()`). On re-verification this does **not** stand as a functional bug — on Windows `WSAGetLastError()` reads the same thread-local last-error slot as `GetLastError()` (Ice itself uses `WSAGetLastError()` for Win32 error retrieval in `Network.cpp`), so timeouts are classified correctly (→ `SelectorTimeoutException`) and no stale value is propagated.

The `std::abort()` on an internal completion-port error is intentional (introduced by #4551, "Abort on fatal error") and reached only on a genuine fatal `GQCS` failure — not on timeouts or normal shutdown (threads are woken via `PostQueuedCompletionStatus` before the port is closed). So the finding is refuted as a new defect; the only open question is the pre-existing design choice of abort-on-fatal-error (#4551) vs retry-on-transient-error, which is a product decision rather than an error-source bug.

PR #4442, opened 2025-09-19
Fix Windows segmentation fault in Selector::getNextHandler

pepone:
See  #4441


externl (approved):
Too bad we can't test this. Especially with the 5s magic value.

PR #4446, opened 2025-09-22
Add exe targets for binaries on Windows to CMake

externl:
This PR adds CMake targets for the binaries shipped in the NuGet: `icebox` and `iceboxadmin`

### DIFF
--- a/cpp/config/IceTargets.cmake
+++ b/cpp/config/IceTargets.cmake
@@ -85,25 +85,8 @@ if(WIN32)
     endif()
   endfunction()
 
-  # Find all executables in the bin directories
-  file(GLOB ICE_EXECUTABLES
-       "${Ice_PREFIX}/build/native/bin/${Ice_WIN32_PLATFORM}/Release/*.exe"
-       "${Ice_PREFIX}/build/native/bin/${Ice_WIN32_PLATFORM}/Debug/*.exe")
-
-  # Get unique executable names without path and extension
-  set(ICE_EXE_NAMES)
-  foreach(exe_path ${ICE_EXECUTABLES})
-    get_filename_component(exe_name ${exe_path} NAME_WE)
-    list(APPEND ICE_EXE_NAMES ${exe_name})
-  endforeach()
-
-  # Remove duplicates (debug/release variants)
-  list(REMOVE_DUPLICATES ICE_EXE_NAMES)
-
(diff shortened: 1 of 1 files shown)


cpp/config/IceTargets.cmake
```diff
@@ -52,6 +52,60 @@ find_path(Ice_SLICE_DIR
...
+
+    if(Ice_${name}_EXE_DEBUG)
+      set_property(TARGET Ice::${name}_EXE PROPERTY
+        IMPORTED_LOCATION_DEBUG "${Ice_${name}_EXE_DEBUG}")
+    endif()
+
+    # Set the appropriate location based on build type
+    if(Ice_${name}_EXE_RELEASE OR Ice_${name}_EXE_DEBUG)
+      set_property(TARGET Ice::${name}_EXE PROPERTY
+        IMPORTED_LOCATION "$<IF:$<CONFIG:Debug>,${Ice_${name}_EXE_DEBUG},${Ice_${name}_EXE_RELEASE}>")
```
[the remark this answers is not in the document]
externl: Using the correct DEBUG/RELEASE is important. The case where only one would exist and not the other could only happen with a source build, not the distributed NuGet. In this case it's fine to fail.

cpp/config/IceTargets.cmake
```diff
@@ -52,6 +52,43 @@ find_path(Ice_SLICE_DIR
...
+    endif()
+  endfunction()
+
+  add_ice_executable(icebox)
+  add_ice_executable(iceboxadmin)
```
pepone: I think iceboxadmin is included by accident in NuGet. There is no need to ship 4 versions of iceboxadmin

cpp/config/IceTargets.cmake
```diff
@@ -52,6 +52,42 @@ find_path(Ice_SLICE_DIR
   NO_DEFAULT_PATH
   REQUIRED)
 
+# Find all executables in the native/bin directory and create imported targets
+if(WIN32)
```
bernardnormier, line 56: For my understanding: WIN32 means Windows for cmake, not the Win32 platform (aka x86)?

bernardnormier (approved):

cpp/config/IceTargets.cmake
externl: Yes!

PR #4447, opened 2025-09-22
Upgrade MSI Path settings

pepone:
A minor update for the MSI:

- Add to path is now enabled by default.
- Removed text about reboot, it is not required.

### DIFF
--- a/packaging/msi/UI.wxs
+++ b/packaging/msi/UI.wxs
@@ -15,7 +15,6 @@
 
         <!-- Customize Exit Dialog -->
         <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="View Release Notes" />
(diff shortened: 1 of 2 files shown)


packaging/msi/UI.wxs
```diff
@@ -15,7 +15,6 @@
 
         <!-- Customize Exit Dialog -->
         <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="View Release Notes" />
-        <Property Id="WIXUI_EXITDIALOGOPTIONALTEXT" Value="You need to reboot your computer to get the updated system path." />
```
bernardnormier, line 18: For a separare PR:

It would be helpful to include text + link to https://docs.zeroc.com/ice/latest/cpp/using-the-ice-service-installer.

PR #4456, opened 2025-09-23
Fix IceBridge description

bernardnormier:
We should never use the term "Ice bridge".

### DIFF
--- a/packaging/deb/debian/control
+++ b/packaging/deb/debian/control
@@ -240,8 +240,8 @@ Section: net
 Depends: libzeroc-ice3.8 (= ${binary:Version}),
          ${misc:Depends},
          ${shlibs:Depends}
-Description: Bridge service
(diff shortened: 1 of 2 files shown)


pepone (approved):

packaging/deb/debian/control
```diff
@@ -240,8 +240,8 @@ Section: net
 Depends: libzeroc-ice3.8 (= ${binary:Version}),
          ${misc:Depends},
          ${shlibs:Depends}
-Description: Bridge service
- This package contains the Ice bridge. The Ice bridge allows you to bridge
```
bernardnormier, line 244: We use "This package contains xxx" for several packages. Is this a standard RPM/DEB description?

If not, we should remove it, as it's pretty obvious package icebridge contains IceBridge.

Issue #4428, opened 2025-09-16
Reduce number of Ice::initialize overloads


bernardnormier:
We provide many - too many - initialize overloads to create Communicator in C++, C#, Java, etc.

For example in C++, we have 12 initialize overloads. This large number of overloads make the API reference hard to read. They also provide duplicate functionality: we shouldn't provide several ways to do the same thing.

> The same applies to the DataStorm::Node constructor overloads, and the Ice::Service::main overloads, and probably more.

## Proposal

I propose to reduce the number of `initialize` overloads in C++ to only 5 (see below), and make similar initialize reductions in other languages.

In C++, the main initialize is / should be:
```cpp
CommunicatorPtr initialize(InitializationData initData = {});
```

Then we provide 2 "convenience" overloads that create this InitializationData for the caller:
```cpp
CommunicatorPtr initialize(int argc, char* argv[]); // same args as main
CommunicatorPtr initialize(int argc, wchar_t* argv[]); // Windows only, same args as wmain
```

and two extra convenience factory functions for the two initialize above, for non-standard "const" argv:
```cpp
CommunicatorPtr initialize(int argc, const char* argv[]);
CommunicatorPtr initialize(int argc, const wchar_t* argv[]); // Windows only
```

And we remove all the other initialize, namely:
 - argc, argv, initData 
 - argc, argv, configFile
 - configFile
 - args, initiData (args = StringSeq)
 - args, configFile

If you need these "advanced" initialize, you can and should create your InitializationData and fill it with Properties created with a configFile, a StringSeq args, etc.

PR #4459, opened 2025-09-23
Simplify loading of plug-ins

bernardnormier:
This PR is a small simplification of the PluginManager, and a preliminary step for #4428.

Prior to this PR, `Ice::initialize` transmitted the command-line arguments to the plug-in manager (through an internal API). This allowed the plug-in manager to "parse-out" these command-line args into properties based on the plug-in name, before loading the plug-in.

This was obviously over the top, and also not documented.

### DIFF
--- a/cpp/src/Ice/PluginManagerI.cpp
+++ b/cpp/src/Ice/PluginManagerI.cpp
@@ -174,13 +174,10 @@ Ice::PluginManagerI::destroy() noexcept
 Ice::PluginManagerI::PluginManagerI(CommunicatorPtr communicator) : _communicator(std::move(communicator)) {}
 
 bool
-Ice::PluginManagerI::loadPlugins(int& argc, const char* argv[])
+Ice::PluginManagerI::loadPlugins()
 {
     assert(_communicator);
-
     bool libraryLoaded = false;
-
-    StringSeq cmdArgs = argsToStringSeq(argc, argv);
 
     const string prefix = "Ice.Plugin.";
     PropertiesPtr properties = _communicator->getProperties();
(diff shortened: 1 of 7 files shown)


cpp/src/Ice/PluginManagerI.cpp
```diff
@@ -295,14 +287,9 @@ Ice::PluginManagerI::loadPlugin(
...
         args.erase(args.begin());
 
-        //
-        // Convert command-line options into properties. First we
-        // convert the options from the plug-in configuration, then
-        // we convert the options from the application command-line.
-        //
+        // Convert command-line options into properties.
         PropertiesPtr properties = _communicator->getProperties();
         args = properties->parseCommandLineOptions(name, args);
```
bernardnormier, line 292: args corresponds to the args at the end of the Ice.Plugin.name value. The name of the plug-in is used as key when parsing these args into properties. I doubt this feature is used much!

externl (approved):

cpp/src/Ice/PluginManagerI.cpp
pepone: So before this PR we passed both args from the command line and args in the plugin name, and now we just pass the args in the plugin name?

pepone (approved):

cpp/src/Ice/PluginManagerI.cpp
bernardnormier: Before this PR, we parsed into properties the args from the Ice.Plugin.name value AND command-line args  (using name as key for the parsing). This PR removes the command-line args parsing but keeps the args parsing. 

Note that the args minus the parsed-out properties are them given to the plug-in factory (no change in this PR).

PR #4462, opened 2025-09-24
Reduce number of Ice::initialize overloads in C++

bernardnormier:
This PR reduces the number of Ice::initialize overloads in C++. See #4428.

It also:
- updates the initialize API in MATAB and Swift (fully done)
- updates the DataStorm::Node constructors, but more work is need in a follow-up PR
- updates the implementation of initialize in Python/Ruby/PHP, but not the Python/Ruby/PHP API (for a follow-up PR)

The changelog entry is also for a follow-up PR.


cpp/include/DataStorm/Node.h
```diff
@@ -27,86 +27,54 @@ namespace DataStorm
     {
     public:
         /// Constructs a DataStorm node.
-        /// A node is the main DataStorm object. It is required to construct topics.
-        /// @param argc The number of arguments in argv.
-        /// @param argv The configuration arguments.
-        /// @param configFile The path to an optional Ice configuration file.
+        /// A node is the main DataStorm object. It is required to construct topics. The node uses the given Ice
```
bernardnormier, line 30: This needs more refactoring. I think the correct solution is to introduce a NodeOptions struct used by the primary constructor. And all other constructors are convenience constructors.

matlab/lib/+Ice/Communicator.m
```diff
@@ -25,14 +25,14 @@
...
+        function obj = Communicator(args, options)
             %COMMUNICATOR Constructs a new communicator.
             %
             %   Examples
             %     communicator = Ice.Communicator();
-            %     [communicator, remArgs] = Ice.Communicator(args);
+            %     communicator = Ice.Communicator(args);
             %     communicator = Ice.Communicator(Properties = props, SliceLoader = sliceLoader);
-            %     [communicator, remArgs] = Ice.Communicator(args, SliceLoader = sliceLoader);
+            %     communicator = Ice.Communicator(args, Properties = props, SliceLoader = sliceLoader);
```
bernardnormier, line 35: I kept args + Properties because I could not find a way to use an arguments block (see below) and disallow this combination. I think that's fine.

ruby/src/IceRuby/Communicator.cpp
```diff
@@ -253,18 +219,13 @@ IceRuby_initialize(int argc, VALUE* argv, VALUE /*self*/)
             //
             // We start at index 1 in order to skip the element that we inserted earlier.
             //
-            for (i = 1; i < ac; ++i)
+            for (size_t i = 1; i < seq.size(); ++i)
```
bernardnormier, line 222: We should not do all this args manipulation in the C++ initialize, but move it into the Ruby code. For a follow-up PR.

cpp/src/Ice/Initialize.cpp
```diff
@@ -139,83 +139,35 @@ Ice::stringSeqToArgs(const StringSeq& args, int& argc, const wchar_t* argv[])
...
-Ice::initialize(StringSeq& args, string_view configFile)
-{
-    InitializationData initData;
-    initData.properties = createProperties();
-    initData.properties->load(configFile);
-    return initialize(args, std::move(initData));
-}
-
-Ice::CommunicatorPtr
-Ice::initialize(InitializationData initData)
```
bernardnormier, line 199: With this initialize, we were NOT cloning the initData.properties. 

However, I think this behavior is unexpected, that's why I changed it in the new initialize(InitializationData).

cpp/test/Ice/executor/Client.cpp
```diff
@@ -28,7 +28,7 @@ Client::run(int argc, char** argv)
     { executor->execute(make_shared<ExecutorCall>(call), conn); };
     // The communicator must be destroyed before the executor is terminated.
     {
-        Ice::CommunicatorHolder communicator = initialize(argc, argv, initData);
+        Ice::CommunicatorHolder communicator = initialize(initData);
```
bernardnormier, line 31: This syntax works because the `CommunicatorHolder(CommunicatorPtr)` constructor is _not_ explicit. Making it explicit would be cleaner but would also break some code, in particular several tests.

cpp/src/DataStorm/Node.cpp
```diff
@@ -12,53 +12,28 @@ using namespace Ice;
...
-        initData.properties = properties;
-
-        return initialize(argc, argv, std::move(initData));
+        InitializationData initData{.properties = properties};
+        return initialize(std::move(initData));
     }
 
-#ifdef _WIN32
-    CommunicatorPtr argsToCommunicator(int& argc, const wchar_t* argv[], optional<string_view> configFile)
+    CommunicatorPtr argsToCommunicator(int& argc, const char* argv[])
```
pepone, line 22: Not sure why we settle on this name, to me `createCommunicatorWithArgs` seems like a better fit.

cpp/src/Ice/Initialize.cpp
```diff
@@ -139,83 +139,35 @@ Ice::stringSeqToArgs(const StringSeq& args, int& argc, const wchar_t* argv[])
...
 Ice::CommunicatorPtr
-Ice::initialize(int& argc, const char* argv[], InitializationData initData)
+Ice::initialize(InitializationData initData)
 {
-    initData.properties = createProperties(argc, argv, initData.properties);
+    // We clone the properties: if two communicators share the same initData, we don't want them to share the same
+    // properties object.
+    if (initData.properties)
+    {
+        initData.properties = initData.properties->clone();
```
externl: I think we should prefer the copy constructor.

ruby/src/IceRuby/Communicator.cpp
pepone: I guess we need to create the Properties in Ruby and not pass args.

cpp/src/Ice/Initialize.cpp
pepone: Isn't the cloning a problem for Scripting languages, for example the Properties wrapper end up pointing to the original, while we are using a clone?

pepone (approved):

ruby/src/IceRuby/Communicator.cpp
bernardnormier: Yes. For a follow-up PR.

cpp/src/Ice/Initialize.cpp
bernardnormier: Why is it a problem?

If you look at existing code, we never assume the initData.properties object becomes the properties object of the communicator. We always use "getProperties" on the communicator after initialization. And getProperties in the scripting languages creates a new wrapper.

cpp/src/DataStorm/Node.cpp
bernardnormier: Yes, for the follow-up PR.

externl (commented):
I think I'd prefer if we didn't do the copy. I would expect that I pass a properties shared object into something I would expect that I could keep using it.

If a user is sharing init data between two communicators, they should be responsible for making a clone if they want to. This goes for any type of reference counted object. 

Users otherwise need to be very careful not to keep using the properties object they set. This copying just makes things more confusing.

cpp/src/Ice/Initialize.cpp
bernardnormier: I updated initialize to no longer clone the properties.

cpp/include/Ice/Initialize.h
```diff
@@ -64,6 +64,7 @@ namespace Ice
     struct InitializationData
     {
         /// The properties for the communicator.
+        /// When not-null, this corresponds to the object returned by the communicator's #getProperties function.
```
externl:
```suggestion
        /// When not-null, this corresponds to the object returned by the Communicator::getProperties function.
```

externl (approved):
Looks good!

pepone (approved):

cpp/test/Ice/location/Server.cpp
```diff
@@ -16,17 +16,17 @@ class Server : public Test::TestHelper
...
 
-    Ice::CommunicatorHolder communicator = initialize(argc, argv, initData);
+    Ice::CommunicatorHolder communicator = initialize(initData);
     //
     // Register the server manager. The server manager creates a new
     // 'server' (a server isn't a different process, it's just a new
     // communicator and object adapter).
     //
-    Ice::PropertiesPtr properties = communicator->getProperties();
+    test(properties == communicator->getProperties());
```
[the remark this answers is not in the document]
bernardnormier: ???

PR #4464, opened 2025-09-24
Reduce number of initialize in Java

bernardnormier:
This is the Java counterpart of #4462 and #4463.


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -102,121 +102,64 @@ public static Properties createProperties(
...
-            String[] args, String configFile, List<String> remainingArgs) {
-        InitializationData initData = null;
-        if (configFile != null) {
-            initData = new InitializationData();
-            initData.properties = new Properties();
-            initData.properties.load(configFile);
-        }
-
-        return initialize(args, initData, remainingArgs);
+    public static Communicator initialize(String[] args, List<String> remainingArgs) {
```
bernardnormier: This overload is not essential. However, we provide this functionality in C++, C# and Swift, and also Ice 3.7, so may as well provide it in Java too.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -102,121 +101,61 @@ public static Properties createProperties(
...
-            List<String> rArgs = new ArrayList<>();
-            initData.properties = createProperties(args, initData.properties, rArgs);
-            args = rArgs.toArray(new String[rArgs.size()]);
-        }
-
-        var communicator = new Communicator(initData);
-        communicator.finishSetup(args != null ? args : new String[0], remainingArgs);
-        return communicator;
+    public static Communicator initialize(String[] args) {
+        var initData = new InitializationData();
```
pepone, line 141: Can we just call the other overload with null remaining args?

pepone (approved):

externl (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
bernardnormier: Fixed as suggested.

PR #4465, opened 2025-09-24
Simplify initialize in Python

bernardnormier:
The Python follow-up for #4462.


python/modules/IcePy/Communicator.cpp
```diff
@@ -82,44 +82,33 @@ communicatorNew(PyTypeObject* type, PyObject* /*args*/, PyObject* /*kwds*/)
...
 
 extern "C" int
 communicatorInit(CommunicatorObject* self, PyObject* args, PyObject* /*kwds*/)
 {
-    PyObject* argList{nullptr};
-    PyObject* initData{nullptr};
-    PyObject* configFile{nullptr};
-    if (!PyArg_ParseTuple(args, "OOO", &argList, &initData, &configFile))
+    PyObject* pyInitData{nullptr};
+    if (!PyArg_ParseTuple(args, "O", &pyInitData))
```
pepone, line 91: I think we should switch this call to use "O!" for typed object, we use "O" because with 3.7 the argument type was unknown, it can be either args, initData, configFile. This should have been refactor when we updated this but I didn't notice it at the momemt.

We use "O!" in a number of place, for example:

```
modules/IcePy/Connection.cpp
267:    if (!PyArg_ParseTuple(args, "O!", identityType, &id))
359:    if (!PyArg_ParseTuple(args, "O!", compressBatchType, &compressBatch))
388:    if (!PyArg_ParseTuple(args, "O!", compressBatchType, &compressBatch))
```

We can do this in a follow up PR.

pepone (approved):

python/python/Ice/Util.py
```diff
@@ -24,23 +24,17 @@
...
 
 
 @overload
-def initialize(
-    args: list[str] | None = None,
-    initData: None = None,
-    configFile: str | None = None,
-    eventLoop: asyncio.AbstractEventLoop | None = None,
-) -> Communicator: ...
+def initialize(*, initData: InitializationData | None = None) -> Communicator: ...
```
externl, line 31: Not ideal, but we can fix in a followup PR.

externl (approved):

PR #4466, opened 2025-09-24
Simplify Ice.initialize in Ruby

bernardnormier:
This is the Ruby follow-up for #4462.


ruby/test/Ice/properties/Client.rb
```diff
@@ -109,5 +109,20 @@ def run(args)
         end
         puts "ok"
 
+        print "testing Ice.initialize with args... "
```
bernardnormier, line 112: This code was not tested by the test suite.

ruby/test/Ice/properties/Client.rb
```diff
@@ -109,5 +109,20 @@ def run(args)
...
+        args = ["--Foo.Bar=1", "--Foo.Bar=2", "--Ice.Trace.Network=3"]
+        Ice.initialize(args) do |communicator|
+            test(communicator.getProperties().getIceProperty("Ice.Trace.Network") == "3")
+            test(args.length == 2)
+        end
+
+        args = ["--Foo.Bar=1", "--Foo.Bar=2", "--Ice.Trace.Network=3"]
+        Ice.initialize(args) do |communicator, remArgs|
```
bernardnormier: I don't understand why we provide this remArgs syntax since we already modify args in place. Should we remove it?

ruby/src/IceRuby/Communicator.cpp
```diff
@@ -40,128 +40,57 @@ static const rb_data_type_t IceRuby_CommunicatorType = {
...
-
-    private:
-        Ice::CommunicatorPtr _communicator;
-    };
-}
-
 extern "C" VALUE
 IceRuby_initialize(int argc, VALUE* argv, VALUE /*self*/)
 {
+    // This is the implementation of the Ice.initialize module method. There is no Ruby wrapper.
```
bernardnormier, line 46: With a Ruby wrapper, we would pass only initData, and the wrapper would create Properties from args when needed.

ruby/src/IceRuby/Communicator.cpp
```diff
@@ -170,61 +99,53 @@ IceRuby_initialize(int argc, VALUE* argv, VALUE /*self*/)
...
-        {
-            data.properties = Ice::createProperties();
+            // Replace the contents of the given argument list with the filtered arguments.
+            callRuby(rb_ary_clear, args);
+
+            for (const auto& arg: seq)
+            {
+                volatile VALUE str = createString(arg);
+                callRuby(rb_ary_push, args, str);
+            }
```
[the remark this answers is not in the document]
bernardnormier: This is all about writing back the remaining arguments into args.

ruby/src/IceRuby/Communicator.cpp
```diff
@@ -40,128 +40,57 @@ static const rb_data_type_t IceRuby_CommunicatorType = {
...
 
-        if (argc > 2)
+        if (argc > 1)
         {
             throw RubyException(rb_eArgError, "invalid number of arguments to Ice::initialize");
         }
 
         volatile VALUE initDataCls = callRuby(rb_path2class, "Ice::InitializationData");
-        volatile VALUE args = Qnil, initData = Qnil, configFile = Qnil;
+        volatile VALUE args = Qnil, rbInitData = Qnil;
```
pepone, line 64: It will be clear to declare `args` and `rbInitData` in separate lines.

ruby/test/Ice/properties/Client.rb
pepone: I agree, doesn't make sense to do both. I will prefer to remove remArgs too.

pepone (approved):

externl (approved):

Issue #4467, opened 2025-09-25
Missing ice files in libice-c++-devel rpm


iivdam:
### Version

- Ice 3.8 Preview
- CentOS Stream release 10

### Describe the bug

CMake error after installing libice-c++-devel:

`sudo dnf install https://download.zeroc.com/ice/nightly/el10/ice-repo-nightly-1.0.0-1.el10.noarch.rpm`
`sudo dnf install libice-c++-devel`

icenius@localhost:~/build/ice-demos/cpp/DataStorm/clock$ cmake3 -B build -S . -G Ninja
CMake Error at /usr/lib64/cmake/Ice-3.8/IceTargets.cmake:47 (find_path):
  Could not find Ice_SLICE_DIR using the following files: Ice/Identity.ice
Call Stack (most recent call first):
  /usr/lib64/cmake/Ice-3.8/IceConfig.cmake:17 (include)
  /home/icenius/build/ice-demos/cpp/cmake/common.cmake:67 (find_package)
  CMakeLists.txt:5 (include)

-- Configuring incomplete, errors occurred!

### Expected behavior

CMake on demo's succeeds after install of libice-c++-devel.

### Additional context

After installing ice-slice cmake succeeds 

`sudo dnf install ice-slice`

icenius@localhost:~/build/ice-demos/cpp/DataStorm/clock$ cmake3 -B build -S . -G Ninja
-- Found Ice: /usr/lib64/cmake/Ice-3.8/IceTargets.cmake (found version "3.8.0-alpha.0")
-- Configuring done (0.1s)
-- Generating done (0.0s)
-- Build files have been written to: /home/icenius/build/ice-demos/cpp/DataStorm/clock/build

PR #4468, opened 2025-09-25
Add missing dependency in libice-c++-devel

pepone:
see #4467


externl (approved):

packaging/rpm/ice.spec
```diff
@@ -199,6 +199,7 @@ your application logic.
 %package -n lib%{?nameprefix}ice-c++-devel
 Summary: Libraries and headers for developing Ice applications in C++.
 Requires: lib%{?nameprefix}ice3.8-c++ = %{version}-%{release}
+Requires: %{?nameprefix}ice-slice = %{version}-%{release}
```
bernardnormier, line 202: We need to do the same for all other "dev kit" packages even if they are not named devel, like the Python and PHP dev kits.

pepone: Updated, and fixed DEB packages too.

PR #4471, opened 2025-09-25
Refactor DataStorm::Node constructor with new NodeOptions

bernardnormier:


cpp/include/DataStorm/Node.h
```diff
@@ -20,62 +22,70 @@ namespace DataStorm
...
-        /// Constructs a DataStorm node.
-        /// A node is the main DataStorm object. It is required to construct topics.
-        /// @param customExecutor An optional executor used to execute user callbacks, if no callback executor is
-        /// provided the Node will use the default callback executor that executes callback in a dedicated thread.
-        Node(std::function<void(std::function<void()> call)> customExecutor = nullptr);
+        Node(int& argc, const wchar_t* argv[]);
 
+        /// @copydoc initialize(int&, const char*[])
+        /// @remark Windows only.
+        Node(int& argc, wchar_t* argv[])
```
bernardnormier: I added this constructor for consistency. It's odd to provide the non-standard "const" argv but omit the standard non-const argv.

cpp/include/DataStorm/Node.h
```diff
@@ -20,62 +22,70 @@ namespace DataStorm
...
-            Ice::CommunicatorPtr communicator,
-            std::function<void(std::function<void()> call)> customExecutor = nullptr);
+        /// A node is the main DataStorm object. It is required to construct topics.
+        /// @param options The options to configure the node.
+        /// @remark This is the main Node constructor. All other constructors are convenience constructors that call
+        /// this constructor.
+        explicit Node(NodeOptions options = {});
 
-        /// Constructs a DataStorm node.
+        /// Constructs a DataStorm node with the specified communicator.
```
bernardnormier, line 56: We use this constructor in many tests.

cpp/include/DataStorm/Node.h
```diff
@@ -20,62 +22,70 @@ namespace DataStorm
...
 
+    /// Options to configure a DataStorm node.
+    /// @headerfile DataStorm/DataStorm.h
+    struct NodeOptions
+    {
+        /// The Ice communicator used by the node. If nullptr, the node creates its own communicator.
+        Ice::CommunicatorPtr communicator{nullptr};
+
+        /// Whether or not the node owns the communicator.
+        /// This option is only used the communicator field is not nullptr. If true, the node destroys the communicator
```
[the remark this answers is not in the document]
bernardnormier: Fixed

cpp/include/DataStorm/Node.h
```diff
@@ -20,62 +22,64 @@ namespace DataStorm
...
         /// A node is the main DataStorm object. It is required to construct topics.
-        /// @param argc The number of arguments in argv.
-        /// @param argv The configuration arguments.
-        /// @param customExecutor An optional executor used to execute user callbacks, if no callback executor is
-        /// provided the Node will use the default callback executor that executes callback in a dedicated thread.
-        Node(int& argc, char* argv[], std::function<void(std::function<void()> call)> customExecutor = nullptr)
-            : Node(argc, const_cast<const char**>(argv), std::move(customExecutor))
+        /// @param communicator The communicator used by the node. If nullptr, the node creates its own communicator.
+        /// @remark This constructor sets the nodeOwnsCommunicator option to false.
+        explicit Node(Ice::CommunicatorPtr communicator)
```
pepone, line 60: I guess we keep this to ensure that `.nodeOwnsComunicator` is set to false?

pepone (approved):

externl (approved):

cpp/include/DataStorm/Node.h
bernardnormier: It's primarily because this constructor is "common" - used in many tests.

You could write instead:
```cpp
NodeOptions options{.communicator = communicator};
Node node{std::move(options)};
```

but that's less convenient. It's also C++20 syntax with Visual C++.

PR #4477, opened 2025-09-25
Exclude keychain files in Swift test build

externl:


swift/test/Package.swift
```diff
@@ -103,10 +103,25 @@ let testTargets = testDirectories.map { (testPath, testConfig) in
...
+                exclude.append(relativePath)
+            }
+        }
+    }
+
     return Target.target(
         name: name,
         dependencies: dependencies,
         path: testPath,
+        exclude: exclude,
```
pepone, line 124: another option would be to upgrade tests to create keychains in a dedicated dir, "keychains" and then just need to exclude that.

externl: We could do that too. For another PR though.

PR #4480, opened 2025-09-26
Fix Swift package to update Slice files

pepone:
This PR fixes two issues with the Swift package:

- Files removed from ice were not being removed from ice-swift-nightly. The script always kept all sources in ice-swift-nightly and only added new ones.
- Slice changes in ice were not committed to ice-swift-nightly.

### DIFF
--- a/packaging/swift/update-ice-swift-nightly.sh
+++ b/packaging/swift/update-ice-swift-nightly.sh
@@ -24,10 +24,10 @@ cd "$root_dir/packaging/swift"
 git clone "https://x-access-token:${ICE_NIGHTLY_PUBLISH_TOKEN}@github.com/zeroc-ice/ice-swift-nightly.git" -b main
 cd ice-swift-nightly
 
-# Clean up the repository to ensure files removed from ice are not kept in ice-swift-nightly
-rm -rfv *
+# Remove existing directories to avoid keeping stale files that may have been removed from the Ice repository.
+rm -rfv cpp slice swift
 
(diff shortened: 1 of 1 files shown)


packaging/swift/update-ice-swift-nightly.sh
```diff
@@ -24,6 +24,10 @@ cd "$root_dir/packaging/swift"
 git clone "https://x-access-token:${ICE_NIGHTLY_PUBLISH_TOKEN}@github.com/zeroc-ice/ice-swift-nightly.git" -b main
 cd ice-swift-nightly
 
+# Clean up the repository to ensure files removed from ice are not kept in ice-swift-nightly
+rm -rfv *
```
externl: Should we also include `./.*`? 

EDIT: on second though I think that's probably considered unsafe. I think `..` is included in `.*` and would destroy git history. And we don't check in any hidden files.

pepone: It would also remove ".git" which we need. Maybe a more correct approach is to list them first with `git ls-tree`

bernardnormier: I think it would be clearer to delete only what we copy below, and add a comment.
```shell
rm -rfv cpp slice swift
```

externl: I don't think we need to worry about it. We don't track any hidden files.

bernardnormier (approved):

PR #4483, opened 2025-09-26
Add devcontainer.yml workflow

externl:


.github/workflows/devcontainer.yml
```diff
@@ -0,0 +1,24 @@
+name: Dev Containers
+
+on:
+  workflow_dispatch:
+  push:
+    branches: ["main"]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: ["main"]
```
pepone, line 9: Do we need to run this on every pull-request/push?

.github/workflows/devcontainer.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Build dev containers
```
pepone, line 18: What are we doing here, building a docker container for each dev-container env? What is the advantage?

PR #4486, opened 2025-09-29
Fix for retry on selector internal errors

pepone:
See #4441

Two issues fixed in this PR:

- The getNextHandler IOCP implementation was returning a null handler if GetQueuedCompletionStatus failed without returning an overlapped structure. We now loop and retry after a small 1 ms delay. This indicates an internal error; the assumption is the system is under high load and cannot allocate resources.
- The Linux/macOS implementation that calls epoll and kevent respectively was missing a continue, which prevented the selector from retrying in this condition. We also reduced the sleep_for timeout to 1 ms (from 5 s).

### DIFF
--- a/cpp/src/Ice/Selector.cpp
+++ b/cpp/src/Ice/Selector.cpp
@@ -14,6 +14,7 @@
 #include <chrono>
 #include <thread>
 
+#ifndef ICE_USE_CFSTREAM
 // Constants shared across Windows and Unix selector implementations.
 
 // Warn every 100 iterations if there is an unexpected selector error. To avoid flooding the log.
@@ -21,6 +22,7 @@ static constexpr int selectorRetryWarnEvery = 100;
 
 // Backoff duration for transient errors. This is the time we wait before retrying to avoid overloading the CPU.
 static constexpr auto transientErrorBackoff = std::chrono::milliseconds(1);
+#endif
 
 using namespace std;
 using namespace IceInternal;


cpp/src/Ice/Selector.cpp
```diff
@@ -116,31 +116,40 @@ Selector::getNextHandler(SocketOperation& status, DWORD& count, int& error, int
...
-                throw SelectorTimeoutException();
+                if (error == WAIT_TIMEOUT)
+                {
+                    throw SelectorTimeoutException();
+                }
+                else
+                {
+                    if (++getQueuedPackageFailed > 100)
+                    {
+                        getQueuedPackageFailed = 0;
```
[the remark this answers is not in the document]
bernardnormier: At least add a comment.

cpp/src/Ice/Selector.cpp
```diff
@@ -116,31 +116,40 @@ Selector::getNextHandler(SocketOperation& status, DWORD& count, int& error, int
...
+                        Ice::SocketException ex(__FILE__, __LINE__, error);
+                        Ice::Warning out(_instance->initializationData().logger);
+                        out << "couldn't dequeue packet from completion port:\n" << ex;
+                    }
+                    std::this_thread::sleep_for(1ms); // Sleep 1ms to avoid looping
```
bernardnormier: I would rework this comment to something like:

> Sleep 1ms to avoid retrying immediately.

cpp/src/Ice/Selector.cpp
```diff
@@ -116,31 +116,40 @@ Selector::getNextHandler(SocketOperation& status, DWORD& count, int& error, int
     LPOVERLAPPED ol;
     error = ERROR_SUCCESS;
 
-    if (!GetQueuedCompletionStatus(_handle, &count, &key, &ol, timeout > 0 ? timeout * 1000 : INFINITE))
+    int getQueuedPackageFailed = 0;
```
bernardnormier: I don't understand this variable name. Why Package?

cpp/src/Ice/Selector.cpp
```diff
@@ -645,10 +655,15 @@ Selector::select(int timeout)
...
-            out << "selector failed:\n" << ex;
-            std::this_thread::sleep_for(5s); // Sleep 5s to avoid looping
+            if (++selectorFailed > 100)
+            {
+                selectorFailed = 0;
+                Ice::SocketException ex(__FILE__, __LINE__, IceInternal::getSocketErrno());
+                Ice::Warning out(_instance->initializationData().logger);
+                out << "selector failed:\n" << ex;
+            }
+            std::this_thread::sleep_for(1ms); // Sleep 1ms to avoid looping
```
bernardnormier: See earlier sleep comment.

cpp/src/Ice/Selector.cpp
```diff
@@ -116,31 +116,40 @@ Selector::getNextHandler(SocketOperation& status, DWORD& count, int& error, int
...
         }
-        AsyncInfo* info = static_cast<AsyncInfo*>(ol);
-        if (info)
-        {
-            status = info->status;
-        }
-        count = 0;
-        error = WSAGetLastError();
-        return reinterpret_cast<EventHandler*>(key);
+        break;
```
bernardnormier: I don't find this while(true) ... break style elegant.

cpp/src/Ice/Selector.cpp
externl: packet?

cpp/src/Ice/Selector.cpp
```diff
@@ -645,10 +655,15 @@ Selector::select(int timeout)
                 continue;
             }
 
-            Ice::SocketException ex(__FILE__, __LINE__, IceInternal::getSocketErrno());
-            Ice::Error out(_instance->initializationData().logger);
-            out << "selector failed:\n" << ex;
-            std::this_thread::sleep_for(5s); // Sleep 5s to avoid looping
+            if (++selectorFailed > 100)
+            {
+                selectorFailed = 0;
```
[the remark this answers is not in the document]
externl: Using a constant makes sense to me.

cpp/src/Ice/Selector.cpp
pepone: What do you suggest?

pepone: I guess we can do:

```
while (!GetQueuedCompletionStatus(_handle, &count, &key, &ol, timeout > 0 ? timeout * 1000 : INFINITE))
    {
....
```

externl: I'm also not a big fan of while(true) loops. I like `while (!GetQueuedCompletionStatus(` better!

cpp/src/Ice/Selector.cpp
pepone: Yes, GetQueuedCompletionStatus deques packets

externl (commented):
In the case where we keep failing over and over (for whatever reason), will this while loop exit if we attempt to destroy the thread pool (say on communicator shutdown)?

My concern is that that we will just hang in these loops instead of going back to the thread pool loop.

cpp/src/Ice/Selector.cpp
```diff
@@ -139,9 +157,9 @@ Selector::getNextHandler(SocketOperation& status, DWORD& count, int& error, int
             status = info->status;
         }
         count = 0;
-        error = WSAGetLastError();
         return reinterpret_cast<EventHandler*>(key);
     }
+    error = 0;
```
bernardnormier, line 162: I would _move_ the `error = ...` from line 127 here.

bernardnormier (approved):

pepone:
> My concern is that that we will just hang in these loops instead of going back to the thread pool loop.

That might be indeed an issue with this update.

Issue #4495, opened 2025-10-01
Remove "chunking" when sending/receiving byte buffers


bernardnormier:
We currently send and receive buffers, because of connection timeouts. For example:

https://github.com/zeroc-ice/ice/blob/dd78cfa3928afe32d94ba878264b7b0bb5d74d7b/csharp/src/Ice/Internal/StreamSocket.cs#L447

We removed / refactored connection timeouts in 3.8 and this chunking is no longer meaningful.

bernardnormier:
Fixed by a number of "chunking" PRs.

PR #4496, opened 2025-10-01
Remove chunking in C# Transceiver

bernardnormier:
The C# portion of #4495.

### DIFF
--- a/csharp/src/Ice/Internal/StreamSocket.cs
+++ b/csharp/src/Ice/Internal/StreamSocket.cs
@@ -90,10 +90,6 @@ internal sealed class StreamSocket
 
     public Socket fd() => _fd;
 
-    public int getSendPacketSize(int length) => _maxSendPacketSize > 0 ? Math.Min(length, _maxSendPacketSize) : length;
-
-    public int getRecvPacketSize(int length) => _maxRecvPacketSize > 0 ? Math.Min(length, _maxRecvPacketSize) : length;
-
     public void setBufferSize(int rcvSize, int sndSize) => Network.setTcpBufSize(_fd, rcvSize, sndSize, _instance);
 
     public int read(Buffer buf)
@@ -145,12 +141,11 @@ internal sealed class StreamSocket
     {
         Debug.Assert(_fd != null && _readEventArgs != null);
 
-        int packetSize = getRecvPacketSize(buf.b.remaining());
         try
         {
             _readCallback = callback;
             _readEventArgs.UserToken = state;
-            _readEventArgs.SetBuffer(buf.b.rawBytes(), buf.b.position(), packetSize);
+            _readEventArgs.SetBuffer(buf.b.rawBytes(), buf.b.position(), buf.b.remaining());
             return !_fd.ReceiveAsync(_readEventArgs);
         }
         catch (System.Net.Sockets.SocketException ex)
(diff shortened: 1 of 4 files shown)


csharp/src/Ice/Internal/Transceiver.cs
```diff
@@ -44,8 +44,8 @@ public interface Transceiver
     /// <param name="buf">The buffer to write.</param>
     /// <param name="callback">The callback that this operation calls when it completes asynchronously.</param>
     /// <param name="state">The state to pass to the callback.</param>
-    /// <param name="messageFullyWritten">When <see langword="true" />, the operation has written or is in the process of
-    /// writing the remainder of the message; otherwise, <see langword="false"/>.</param>
+    /// <param name="messageFullyWritten">When <see langword="true" />, the operation has written or is in the process
```
bernardnormier, line 47: Given the revised logic, we should rename messageFullyWritten => messageWritten. For a follow-up PR.

csharp/src/Ice/Internal/StreamSocket.cs
```diff
@@ -259,10 +253,7 @@ public void finishWrite(Buffer buf)
     {
         if (_fd == null) // Transceiver was closed
         {
-            if (buf.size() - buf.b.position() < _maxSendPacketSize)
-            {
-                buf.b.position(buf.b.limit()); // Assume all the data was sent for at-most-once semantics.
-            }
+            buf.b.position(buf.b.limit()); // Assume all the data was sent for at-most-once semantics.
```
bernardnormier, line 256: See SSL version.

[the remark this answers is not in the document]
bernardnormier: Assuming all the data was sent is the pessimistic option for at-most-once semantics.

csharp/src/Ice/SSL/TransceiverI.cs
```diff
@@ -283,7 +270,7 @@ public void finishWrite(Ice.Internal.Buffer buf)
             {
                 throw ex.InnerException;
             }
-            buf.b.position(buf.b.position() + sent);
+            buf.b.position(buf.b.position() + buf.b.remaining());
```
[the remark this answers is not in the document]
bernardnormier: This is madness. _writeResult is a plain Task, it doesn't contain the bytes written.

pepone (approved):
The code looks good, not clear if the back pressure windows failures in CI are related to the change, seems they could be.

externl (approved):

csharp/src/Ice/Internal/StreamSocket.cs
```diff
@@ -231,14 +226,13 @@ public bool startWrite(Buffer buf, AsyncCallback callback, object state, out boo
...
-        int packetSize = getSendPacketSize(buf.b.remaining());
         try
         {
             _writeCallback = callback;
             _writeEventArgs.UserToken = state;
-            _writeEventArgs.SetBuffer(buf.b.rawBytes(), buf.b.position(), packetSize);
+            _writeEventArgs.SetBuffer(buf.b.rawBytes(), buf.b.position(), buf.b.remaining());
             bool completedSynchronously = !_fd.SendAsync(_writeEventArgs);
-            messageFullyWritten = packetSize == buf.b.remaining();
+            messageFullyWritten = true;
```
externl, line 235: I think this is worth a comment to say that while isn't not actually fully written, it's good enough for what we use this for.

externl (approved):

PR #4497, opened 2025-10-02
Remove support for using poll as a selector mechanism

pepone:


cpp/src/Ice/Selector.h
```diff
@@ -103,7 +99,9 @@ namespace IceInternal
     private:
         void wakeup();
         void checkReady(EventHandler*);
+#    if defined(ICE_USE_KQUEUE)
         void updateSelector();
+#    endif
```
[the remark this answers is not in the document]
pepone: If this was right the linux builds will be failing but they are not!

cpp/src/Ice/Selector.cpp
```diff
@@ -687,10 +632,10 @@ Selector::checkReady(EventHandler* handler)
     }
 }
 
+#    if defined(ICE_USE_KQUEUE)
 void
 Selector::updateSelector()
 {
```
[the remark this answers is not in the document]
pepone: same here, if this was right the linux build will be failing.

cpp/src/Ice/Network.h
```diff
@@ -33,7 +33,7 @@ typedef int ssize_t;
 #    include <unistd.h>
 #endif
 
-#if defined(__linux__) && !defined(ICE_NO_EPOLL)
+#if defined(__linux__)
 #    define ICE_USE_EPOLL 1
 #elif (defined(__APPLE__) || defined(__FreeBSD__) || defined(__FreeBSD_kernel__)) && TARGET_OS_IPHONE == 0
```
bernardnormier: (Unrelated to this PR)
I don't see how this compile on FreeBSD: TARGET_OS_IPHONE is defined on Apple, not FreeBSD.

cpp/src/Ice/Selector.h
```diff
@@ -15,10 +15,6 @@
 #    include <sys/epoll.h>
 #elif defined(ICE_USE_KQUEUE)
 #    include <sys/event.h>
-#elif defined(ICE_USE_IOCP)
-// Nothing to include
-#elif defined(ICE_USE_POLL)
-#    include <sys/poll.h>
 #endif
 
 #include <condition_variable>
```
bernardnormier: I would move condition_variable next to map and add ICE_USE_CFSTREAM in an elif.

cpp/src/Ice/Selector.cpp
```diff
@@ -406,11 +388,6 @@ Selector::finish(EventHandler* handler, bool closeNow)
         //
         updateSelector();
     }
-#    elif !defined(ICE_USE_EPOLL)
```
bernardnormier, line 409: Super confusing conditional.

cpp/src/Ice/Network.h
externl, line 39: Does `#elif defined(__APPLE__)` imply `TARGET_OS_IPHONE > 0` maybe we should fix to be something like:

```
#elif defined(__APPLE__) && TARGET_OS_IPHONE > 0
#    define ICE_USE_CFSTREAM 1
#elif defined(__APPLE__) || defined(__FreeBSD__) || defined(__FreeBSD_kernel__)
#    define ICE_USE_KQUEUE 1

**Review by bernardnormier — APPROVED** [review]
Looks good.

File: cpp/src/Ice/Network.h
Comment (externl) [reply cpp/src/Ice/Network.h:39]: Or maybe we can just use feature detection since we support C++17

```
#  if __has_include(<sys/epoll.h>)
#        define ICE_USE_EPOLL 1
#    elif __has_include(<sys/event.h>)
#        define ICE_USE_KQUEUE 1
#    endif
```

externl (approved):

cpp/src/Ice/Network.h
pepone: Maybe was falling back to poll, and we didn't notice.

cpp/src/Ice/Network.h
pepone: I applied the first suggestion.

PR #4503, opened 2025-10-03
Update Swift doc comments to use standard conventions

externl:
This changes were made with the help of Copilot. 

In a follow-up PR I plan to review the generated code doc comments.

### DIFF
--- a/swift/src/Ice/Communicator.swift
+++ b/swift/src/Ice/Communicator.swift
@@ -25,7 +25,7 @@ public protocol Communicator: AnyObject, Sendable {
 
     /// Checks whether or not `shutdown` was called on this communicator.
     ///
-    /// - returns: `Bool` - True if shutdown was called on this communicator; false otherwise.
+    /// - Returns: True if shutdown was called on this communicator; false otherwise.
     func isShutdown() -> Bool
 
     /// Convert a stringified proxy into a proxy.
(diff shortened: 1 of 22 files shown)


swift/src/Ice/Communicator.swift
```diff
@@ -79,79 +73,76 @@ public protocol Communicator: AnyObject, Sendable {
...
     ///
-    /// - parameter adapter: The object adapter to associate with new outgoing connections.<
+    /// - Parameter adapter: The object adapter to associate with new outgoing connections.
     func setDefaultObjectAdapter(_ adapter: ObjectAdapter?)
 
     /// Get the implicit context associated with this communicator.
     ///
-    /// - returns: `ImplicitContext` - The implicit context associated with this communicator; returns null when the
-    ///                                 property Ice.ImplicitContext is not set or is set to None.
+    /// - Returns: The implicit context associated with this communicator; returns null when the
```
pepone:
```suggestion
    /// - Returns: The implicit context associated with this communicator; returns nil when the
```

swift/src/Ice/Communicator.swift
```diff
@@ -190,38 +179,36 @@ public protocol Communicator: AnyObject, Sendable {
     /// Ice.Admin.InstanceName is not set. If Ice.Admin.DelayCreation is 0 or not set, getAdmin is called
     /// by the communicator initialization, after initialization of all plugins.
     ///
-    /// - returns: `ObjectPrx?` - A proxy to the main ("") facet of the Admin object, or a null proxy if no Admin
-    /// object is configured.
+    /// - Returns: A proxy to the main ("") facet of the Admin object, or a null proxy if no Admin
```
pepone:
```suggestion
    /// - Returns: A proxy to the main ("") facet of the Admin object, or a nil proxy if no Admin
```

We use both null and nil for the comments, we have settle on one, I think nil is the more appropriate for Swift.

swift/src/Ice/Communicator.swift
```diff
@@ -190,38 +179,36 @@ public protocol Communicator: AnyObject, Sendable {
...
     func removeAdminFacet(_ facet: String) throws -> Dispatcher
 
     /// Returns a facet of the Admin object.
     ///
-    /// - parameter facet: `String` The name of the Admin facet.
-    ///
-    /// - returns: `Dispatcher?` - The servant associated with this Admin facet, or null if no facet is registered with the
-    /// given name.
+    /// - Parameter facet: The name of the Admin facet.
+    /// - Returns: The servant associated with this Admin facet, or null if no facet is registered with the
```
pepone, line 205:
```suggestion
    /// - Returns: The servant associated with this Admin facet, or nil if no facet is registered with the
```

swift/src/Ice/OutputStream.swift
```diff
@@ -159,13 +161,11 @@ public final class OutputStream {
...
-    /// - parameter typeId: `String` - The Slice type ID corresponding to this slice.
-    ///
-    /// - parameter compactId: `Int32` - The Slice compact type ID corresponding to this
-    ///   slice or -1 if no compact ID is defined for the type ID.
-    ///
-    /// - parameter last: `Bool` - True if this is the last slice, false otherwise.
+    /// - Parameters:
+    ///   - typeId: The Slice type ID corresponding to this slice.
+    ///   - compactId: The Slice compact type ID corresponding to this
+    ///     slice or -1 if no compact ID is defined for the type ID.
```
pepone, line 167: Can't we use 120 line length like in the rest of the code?

swift/src/Ice/Properties.swift
```diff
@@ -7,143 +7,115 @@ import Foundation
...
+    /// - Returns: The property value interpreted as list of strings, or the default value.
     func getPropertyAsListWithDefault(key: String, value: StringSeq) -> StringSeq
 
     /// Get all properties whose keys begins with prefix. If prefix is an empty string, then all
     /// properties are returned.
-    ///
-    /// - parameter prefix: `String` The prefix to search for (empty string if none).
-    ///
-    /// - returns: `PropertyDict` - The matching property set.
+    /// - Parameter prefix: The prefix to search for (empty string if none).
```
pepone:
```suggestion
    /// - Parameter prefix: The prefix to search for, or an empty string to return all properties.
```

pepone (approved):

swift/src/Ice/InputStream.swift
```diff
@@ -764,16 +754,15 @@ extension InputStream {
...
     public func read<ProxyImpl>() throws -> ProxyImpl? where ProxyImpl: ObjectPrxI {
         return try ProxyImpl.ice_read(from: self)
     }
 
     /// Reads an optional proxy from the stream (internal helper).
     ///
-    /// - parameter tag: `Int32` - The tag of the optional data member or parameter.
-    ///
-    /// - returns: `ProxyImpl?` - The proxy read from the stream.
+    /// - Parameter tag: The tag of the optional data member or parameter.
```
bernardnormier, line 764: We should replace "data member" by field when talking about Slice.

bernardnormier (approved):

swift/src/Ice/Communicator.swift
externl: Will fix in follow-up

swift/src/Ice/OutputStream.swift
externl: Yes, we should.

swift/src/Ice/InputStream.swift
externl: Will do in follow-up PR.

PR #4506, opened 2025-10-05
Remove chunking from SecureTransport

bernardnormier:
This PR removes the chunking in the C++ SecureTransport transceiver (C++, macOS).

### DIFF
--- a/cpp/src/Ice/SSL/SecureTransportTransceiverI.cpp
+++ b/cpp/src/Ice/SSL/SecureTransportTransceiverI.cpp
@@ -82,23 +82,6 @@ Ice::SSL::SecureTransport::TransceiverI::initialize(IceInternal::Buffer& readBuf
         _connected = true;
     }
 
-    //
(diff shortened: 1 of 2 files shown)


cpp/src/Ice/SSL/SecureTransportTransceiverI.cpp
```diff
@@ -82,23 +82,6 @@ Ice::SSL::SecureTransport::TransceiverI::initialize(IceInternal::Buffer& readBuf
         _connected = true;
     }
 
-    //
-    // Limit the size of packets passed to SSLWrite/SSLRead to avoid
```
bernardnormier, line 86: My understanding is all the transports we provide on macOS and iOS are non-blocking. As a result, there is no reason to limit the amount of data given to SSLWrite/SSLRead.

pepone (approved):

externl (approved):

PR #4507, opened 2025-10-06
Remove membio support in OpenSSL transceiver

pepone:
This PR removes membio support in OpenSSL transceiver, that can be used by non socket based transport. As far as I see we don't have any non-socket based transports.

### DIFF
--- a/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
+++ b/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
@@ -593,9 +593,6 @@ OpenSSL::TransceiverI::TransceiverI(
       _peerCertificate(nullptr),
       _ssl(nullptr),
       _sslCtx(nullptr),
-      _sentBytes(0),
-      _maxSendPacketSize(0),
-      _maxRecvPacketSize(0),
       _localSSLContextSelectionCallback(
           serverAuthenticationOptions.serverSSLContextSelectionCallback
               ? serverAuthenticationOptions.serverSSLContextSelectionCallback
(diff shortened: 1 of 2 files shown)


cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
```diff
@@ -154,47 +125,31 @@ OpenSSL::TransceiverI::initialize(IceInternal::Buffer& readBuffer, IceInternal::
...
-                            else if (SSL_want_write(_ssl))
-                            {
-                                return IceInternal::SocketOperationWrite;
-                            }
-
-                            break;
+                            return IceInternal::SocketOperationRead;
                         }
-
-                        if (IceInternal::connectionLost() || IceInternal::getSocketErrno() == 0)
```
bernardnormier, line 194: Why did you remove this if block? It's not a mechanical "memBio" removal.

cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
```diff
@@ -394,32 +309,25 @@ OpenSSL::TransceiverI::write(IceInternal::Buffer& buf)
...
-                        if (IceInternal::wouldBlock())
-                        {
-                            assert(SSL_want_write(_ssl));
-                            return IceInternal::SocketOperationWrite;
-                        }
+                    if (IceInternal::wouldBlock())
+                    {
+                        assert(SSL_want_write(_ssl));
+                        return IceInternal::SocketOperationWrite;
                     }
```
bernardnormier, line 339: Here you kept the connectionLost if block.

cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
pepone: An oversight, we need to keep it. Fixed.

bernardnormier:
What about removing _maxSendPacketSize and _maxRecvPacketSize?

PR #4510, opened 2025-10-06
Remove chunking from JavaScript TCP transport

pepone:
This PR:
- Removes chunking from JavaScript TCP transport.
- Updates comments in WS transport to explain why it still uses chunking
- Removes a workaround for old Safari version bug.

### DIFF
--- a/config/PropertyNames.xml
+++ b/config/PropertyNames.xml
@@ -117,7 +117,7 @@
         <property name="SyslogPort" languages="java" default="514"/>
         <property name="TCP.Backlog" languages="cpp,csharp,java" default="511" />
         <property name="TCP.RcvSize" languages="cpp,csharp,java" />
-        <property name="TCP.SndSize" languages="cpp,csharp,java" />
+        <property name="TCP.SndSize" languages="cpp,csharp,java,js" />
         <property name="ThreadPool.Client" class="ThreadPool" languages="cpp,csharp,java" />
         <property name="ThreadPool.Server" class="ThreadPool" languages="cpp,csharp,java" />
         <property name="ThreadPriority" languages="csharp,java" />
--- a/js/src/Ice/WSTransceiver.js
+++ b/js/src/Ice/WSTransceiver.js
@@ -158,6 +158,11 @@ class WSTransceiver {
         }
         console.assert(this._fd);
 
+        // If the write buffer is full, we wait until some data has been sent before sending more data.
+        // The bufferedAmount represent the number of bytes that have been queued using send() but not yet transmitted to the network.
+        // we don't want to keep calling send() or the socket might be closed by the browser.
+        // The writeReadyTimeout uses an exponential backoff to avoid busy loop.
+
         const cb = () => {
             if (this._fd) {
                 const packetSize =
@@ -190,13 +195,6 @@ class WSTransceiver {
 
             this._fd.send(slice);
             byteBuffer.position += packetSize;
-
-            // WORKAROUND for Safari issue. The websocket accepts all the data (bufferedAmount is always 0). We
-            // relinquish the control here to ensure timeouts work properly.
-            if (IsSafari && byteBuffer.remaining > 0) {
-                Timer.setTimeout(cb, this.writeReadyTimeout());
-                return false;
-            }
         }
         return true;
     }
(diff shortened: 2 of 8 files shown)


config/PropertyNames.xml
```diff
@@ -118,6 +118,7 @@
         <property name="TCP.Backlog" languages="cpp,csharp,java" default="511" />
         <property name="TCP.RcvSize" languages="cpp,csharp,java" />
         <property name="TCP.SndSize" languages="cpp,csharp,java" />
+        <property name="WS.MaxBufferedAmount" languages="js" />
```
bernardnormier: We should keep the properties in alphabetical order.

externl: We maybe want to keep this list sorted alphabetically.

js/src/Ice/WSTransceiver.js
```diff
@@ -25,7 +25,7 @@ class WSTransceiver {
     constructor(instance, secure, addr, resource) {
         this._readBuffers = [];
         this._readPosition = 0;
-        this._maxSendPacketSize = instance.properties().getPropertyAsIntWithDefault("Ice.TCP.SndSize", 512 * 1024);
+        this._maxBufferedAmount = instance.properties().getPropertyAsIntWithDefault("Ice.WS.MaxBufferedAmount", 512 * 1024);
```
bernardnormier: Use getIceProperty and set the default in PropertyNames.xml.

externl: Should we set the default in the xml, and use getIceProperty? Is this default just a random number we picked or does it have significance?

js/src/Ice/WSTransceiver.js
```diff
@@ -158,13 +158,18 @@ class WSTransceiver {
         }
         console.assert(this._fd);
 
+        // If the write buffer is full, we wait until some data has been sent before sending more data.
+        // The bufferedAmount represents the number of bytes that have been queued using send() but not yet transmitted to the network.
+        // we don't want to keep calling send() or the socket might be closed by the browser.
```
externl:
```suggestion
        // We don't want to keep calling send() or the socket might be closed by the browser.
```

js/src/Ice/WSTransceiver.js
```diff
@@ -158,13 +158,18 @@ class WSTransceiver {
...
         const cb = () => {
             if (this._fd) {
                 const packetSize =
-                    this._maxSendPacketSize > 0 && byteBuffer.remaining > this._maxSendPacketSize
-                        ? this._maxSendPacketSize
+                    this._maxBufferedAmount > 0 && byteBuffer.remaining > this._maxBufferedAmount
+                        ? this._maxBufferedAmount
                         : byteBuffer.remaining;
-                if (this._fd.bufferedAmount + packetSize <= this._maxSendPacketSize) {
+                if (this._fd.bufferedAmount + packetSize <= this._maxBufferedAmount) {
```
externl, line 172: What happens here if _maxBufferedAmount is set to 0?

js/src/Ice/WSTransceiver.js
```diff
@@ -158,13 +158,18 @@ class WSTransceiver {
...
+        // The bufferedAmount represents the number of bytes that have been queued using send() but not yet transmitted to the network.
+        // we don't want to keep calling send() or the socket might be closed by the browser.
+        // The writeReadyTimeout uses an exponential backoff to avoid busy loop.
+
         const cb = () => {
             if (this._fd) {
                 const packetSize =
-                    this._maxSendPacketSize > 0 && byteBuffer.remaining > this._maxSendPacketSize
-                        ? this._maxSendPacketSize
+                    this._maxBufferedAmount > 0 && byteBuffer.remaining > this._maxBufferedAmount
```
externl, line 169: Why would a user want to set `_maxBufferedAmount` value to `0`? Or for that matter, why isn't is 0 by default?

pepone: 0 means no Max, we will keep buffering data and if the WebSocket cannot keep-up the browser will close it. Definitely not a good default.

js/src/Ice/WSTransceiver.js
pepone: We don't enter this branch, which is for stop sending more data. And continue to send the whole buffer.

PR #4517, opened 2025-10-07
android controller logging

externl:
- **Add more Android controller logging**
- **only run android ci**

### DIFF
--- a/scripts/Util.py
+++ b/scripts/Util.py
@@ -2284,7 +2284,7 @@ class RemoteProcessController(ProcessController):
         def teardown(self, current, success):
             pass
 
-    def __init__(self, current, endpoints="tcp"):
+    def __init__(self, current, endpoints):
         self.processControllerProxies = {}
         self.controllerApps = []
         self.driver = current.driver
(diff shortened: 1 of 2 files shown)


scripts/Util.py
```diff
@@ -2511,30 +2522,29 @@ def startEmulator(self, avd):
...
 
         if self.emulator.poll():
             raise RuntimeError("failed to start the Android emulator `{}' on port {}".format(avd, port))
 
         self.avd = avd
 
-        #
+        print("waiting for the emulator to respond to adb")
+        subprocess.run([self.adb(), "wait-for-device"], timeout=60, check=True)
+
```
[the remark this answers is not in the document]
externl: I think CalledProcessError is fine.

pepone (approved):

scripts/Util.py
```diff
@@ -2511,30 +2520,29 @@ def startEmulator(self, avd):
...
-            except RuntimeError:
-                pass  # expected if device is offline
-            #
-            # If the emulator doesn't complete boot in 300 seconds give up
-            #
-            if (time.time() - t) > 300:
-                raise RuntimeError("couldn't start the Android emulator `{}'".format(avd))
+        # Wait for up to 5 minutes (300 seconds)
+        while (time.time() - t) <= 300:
+            if run(f"{self.adb()} shell getprop sys.boot_completed").strip() == "1":
```
pepone, line 2540: Is this correct? The previous code expected run to return an array, does it return a string?

pepone (approved):

scripts/Util.py
externl: The previous code was a little confusing. `run` returns a string, and we were actually really only checking the first character.

PR #4518, opened 2025-10-08
Swift documentation updates

### LINKED ISSUE #2274: Update slice2swift doc-comment generation

externl:
This PR fixes issues with doc comments in generated Swift code and merges the various doc sites into a single one.

Closes #2274

Screenshot 2025-10-08 at 3 44 27 PM

### DIFF
--- a/.github/actions/documentation/action.yml
+++ b/.github/actions/documentation/action.yml
@@ -68,7 +68,7 @@ runs:
 
         swift package --allow-writing-to-directory ./swift/docs generate-documentation \
           --enable-experimental-combined-documentation \
-          --fallback-display-name "ZeroC Ice for Swift" \
+          --fallback-display-name "Ice for Swift" \
           --target Ice --target IceStorm --target Glacier2 --target IceGrid \
           --output-path ./swift/docs --transform-for-static-hosting --hosting-base-path /ice/main/api/swift 2>&1 | tee -a doc_output.log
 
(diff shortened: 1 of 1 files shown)


cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -95,7 +86,7 @@ Slice::Swift::writeDocSummary(IceInternal::Output& out, const ContainedPtr& p)
         return;
     }
 
-    bool hasStarted = false;
+    bool hasStarted{false};
```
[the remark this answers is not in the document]
bernardnormier: We should definitely use `bool hasStarted{false};`

cpp/src/slice2swift/Gen.cpp
```diff
@@ -1265,9 +1258,9 @@ Gen::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
 
     out << nl << "/// Extracts a proxy from the stream. The stream must have been initialized with a communicator.";
```
bernardnormier, line 1259: I find the Extract doc-comment very odd. Obviously, we're unmarshaling or decoding or maybe even reading a proxy from the stream; we're not "extracting" anything.

externl: "Extracts a" is the same wording we use in C# and Java as well. Best to open a separate issue to changes these if you want.

PR #4533, opened 2025-10-10
Improve the Formatting of 'Network.java'

InsertCreativityHere:
Was trying to read this file earlier, and the formatting in it is just too bad to ignore.
This file is already complicated enough, but round after round of automatic formatting has left it even harder to read.

So this PR just cleans up the formatting, with a couple small simplifications to the code.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Network.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Network.java
@@ -24,13 +24,13 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
+import java.util.stream.Collectors;
 
 /**
  * @hidden Public because it's used by IceDiscovery and IceLocatorDiscovery.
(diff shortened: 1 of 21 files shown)


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Network.java
```diff
@@ -45,23 +45,27 @@ public final class Network {
...
         "(([01]?\\d\\d?|2[0-4]\\d|25[0-5])\\.){3}([01]?\\d\\d?|2[0-4]\\d|25[0-5])";
-    private static final String ipv6Pattern =
-        "(([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,7}:|([0-9a-fA-F]{1,4}:){1,6}:[0-9a-"
-            + "fA-F]{1,4}|([0-9a-fA-F]{1,4}:){1,5}(:[0-9a-fA-F]{1,4}){1,2}|([0-9a-fA-F]{1,4}:){1,4}(:[0-9a-fA-F]{1"
-            + ",4}){1,3}|([0-9a-fA-F]{1,4}:){1,3}(:[0-9a-fA-F]{1,4}){1,4}|([0-9a-fA-F]{1,4}:){1,2}(:[0-9a-fA-F]{1,"
-            + "4}){1,5}|[0-9a-fA-F]{1,4}:((:[0-9a-fA-F]{1,4}){1,6})|:((:[0-9a-fA-F]{1,4}){1,7}|:)|fe80:(:[0-9a-fA-"
-            + "F]{0,4}){0,4}%[0-9a-zA-Z]{1,}|::(ffff(:0{1,4}){0,1}:){0,1}((25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])"
-            + "\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9])|([0-9a-fA-F]{1,4}:){1,4}:((25[0-5]|(2[0-4]|1{0,1}"
-            + "[0-9]){0,1}[0-9])\\.){3,3}(25[0-5]|(2[0-4]|1{0,1}[0-9]){0,1}[0-9]))";
+    private static final String ipv6Pattern = "("
```
InsertCreativityHere, line 48: This regex can match lots of different patterns, so I split it up as `one pattern` = `one line` so it's actually readable.
Well... more readable...

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Network.java
```diff
@@ -412,15 +394,13 @@ public static void setSendBufferSize(SocketChannel fd, int size) {
     }
 
     public static int getSendBufferSize(SocketChannel fd) {
```
InsertCreativityHere, line 396: Don't see the point to this temporary variable.
We either return the size immediately, or catch an exception and throw.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Network.java
```diff
@@ -617,23 +582,18 @@ public static List<InetSocketAddress> getAddresses(
...
                         break;
                     }
                 }
             }
-        } catch (java.net.SocketException ex) {
-            throw new SocketException(ex);
-        } catch (java.lang.SecurityException ex) {
+        } catch (java.net.SocketException | java.lang.SecurityException ex) {
             throw new SocketException(ex);
         }
```
InsertCreativityHere, line 610: You can catch multiple exceptions at once with `|` since Java 7...

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Network.java
```diff
@@ -795,30 +715,7 @@ public static String fdToString(
     }
 
     public static String fdToString(SelectableChannel fd) {
```
InsertCreativityHere, line 729: This implementation was identical to the underloaded `fdToString`.
Might as well re-use that implementation instead of copy/pasting it.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Network.java
```diff
@@ -861,14 +758,6 @@ public static String addressesToString(
         return s.toString();
     }
 
-    public static String addressesToString(
-            InetAddress localAddr,
-            int localPort,
-            InetAddress remoteAddr,
-            int remotePort) {
-        return addressesToString(localAddr, localPort, remoteAddr, remotePort, null, null);
-    }
```
InsertCreativityHere, line 877: This function isn't called anywhere in the `ice` repo.
And even if it's public, no user should ever touch `Network.java`...

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Network.java
pepone: We can get rid of this once we move to JDK 22

https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/net/InetAddress.html#ofLiteral(java.lang.String)

PR #4525, opened 2025-10-09
Fix Java udp test failure

bernardnormier:
This PR fixes an java/Ice/udp test failure on Ubuntu with IPv6.

The root cause was in the code/logic for eliminating duplicates in the server-side addresses.

Fixes #4474 
Fixes #3389

PR #4534, opened 2025-10-11
Update UDP multicast code in C++ and C#

bernardnormier:
This PR synchronizes the C++ and C# code with the Java code. See #4525.

### DIFF
--- a/cpp/src/Ice/Network.cpp
+++ b/cpp/src/Ice/Network.cpp
@@ -175,8 +175,7 @@ namespace
             // We always include the loopback interface (provided it's up), even if it doesn't support multicast.
             // This is required for the DataStorm tests on Ubuntu 24.04.
             unsigned int flags = curr->ifa_flags;
-            if ((flags & IFF_UP) != 0 &&
-                ((flags & IFF_MULTICAST) != 0 || (flags & IFF_LOOPBACK) != 0) &&
+            if ((flags & IFF_UP) != 0 && ((flags & IFF_MULTICAST) != 0 || (flags & IFF_LOOPBACK) != 0) &&
                 curr->ifa_addr)
             {
                 if (curr->ifa_addr->sa_family == AF_INET && protocol != EnableIPv6)
--- a/cpp/src/Ice/Network.cpp
+++ b/cpp/src/Ice/Network.cpp
@@ -145,6 +145,7 @@ namespace
                                 if (addr.saIn.sin_addr.s_addr != 0)
                                 {
                                     result.push_back(addr);
+                                    break; // We return at most one address per interface.
                                 }
                             }
                             else if (addr.saStorage.ss_family == AF_INET6 && protocol != EnableIPv4)
@@ -152,6 +153,7 @@ namespace
                                 if (!IN6_IS_ADDR_UNSPECIFIED(&addr.saIn6.sin6_addr))
                                 {
                                     result.push_back(addr);
+                                    break; // We return at most one address per interface.
                                 }
                             }
                         }
@@ -202,7 +204,6 @@ namespace
                     }
                 }
             }
-
             curr = curr->ifa_next;
         }
 
--- a/cpp/src/Ice/Network.cpp
+++ b/cpp/src/Ice/Network.cpp
@@ -36,12 +36,6 @@ using namespace IceInternal;
 
 namespace
 {
-    struct AddressCompare
-    {
-    public:
-        bool operator()(const Address& lhs, const Address& rhs) const { return compareAddress(lhs, rhs) < 0; }
-    };
-
     void sortAddresses(vector<Address>& addrs, ProtocolSupport protocol, bool preferIPv6)
     {
         if (protocol == EnableBoth)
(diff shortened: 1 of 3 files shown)


externl (approved):
Looks good!

cpp/src/Ice/Network.cpp
```diff
@@ -999,21 +967,27 @@ IceInternal::isAddressValid(const Address& addr)
...
     if (isWildcard(intf, protocolSupport, ipv4Wildcard))
     {
+        vector<string> interfaces;
         vector<Address> addrs = getLocalAddresses(ipv4Wildcard ? EnableIPv4 : protocolSupport);
         for (const auto& addr : addrs)
         {
-            interfaces.push_back(inetAddrToString(addr)); // We keep link local addresses for multicast
+            string addrStr = inetAddrToString(addr);
+            // Suppress duplicates.
+            if (find(interfaces.begin(), interfaces.end(), addrStr) == interfaces.end())
```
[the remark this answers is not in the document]
bernardnormier: The number of interfaces is small and this code is not called often.

cpp/src/Ice/Network.cpp
```diff
@@ -140,28 +134,25 @@ namespace
             {
                 for (PIP_ADAPTER_ADDRESSES aa = adapter_addresses; aa != nullptr; aa = aa->Next)
                 {
```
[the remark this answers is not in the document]
bernardnormier: I think the code is clear enough.

cpp/src/Ice/Network.cpp
```diff
@@ -192,7 +186,7 @@ namespace
                         if (interfaces.find(curr->ifa_name) == interfaces.end())
                         {
                             result.push_back(addr);
-                            interfaces.insert(curr->ifa_name);
+                            interfaces.insert(curr->ifa_name); // one address per interface name
```
bernardnormier: Without this check, we (can) get "address already in use" because we may end up listening multiple times on the same interface.

scripts/DataStormUtil.py
```diff
@@ -38,9 +31,13 @@ def getEffectiveProps(self, current, props):
         if ("DataStorm.Node.Multicast.Enabled", 1) in props.items():
             port = current.driver.getTestPort(20)
             props["DataStorm.Node.Multicast.Endpoints"] = f"udp -h 239.255.0.1 -p {port}"
-            props["DataStorm.Node.Multicast.Proxy"] = (
-                f"DataStorm/Lookup -d:udp -h 239.255.0.1 --sourceAddress 127.0.0.1 -p {port}"
```
bernardnormier, line 42: I believe this `--sourceAddress 127.0.0.1` was really a bug. On Ubuntu, where loopback is not configured for multicast with the GitHub runners, this would bypass the UDP multicast logic by sending directly to loopback ... and only worked when we listened for UDP on loopback, which we shouldn't do for an interface without multicast.

cpp/src/Ice/Network.cpp
```diff
@@ -140,28 +134,27 @@ namespace
...
+                        for (PIP_ADAPTER_UNICAST_ADDRESS ua = aa->FirstUnicastAddress; ua != nullptr; ua = ua->Next)
                         {
-                            if (addr.saIn.sin_addr.s_addr != 0)
+                            Address addr;
+                            memcpy(&addr.saStorage, ua->Address.lpSockaddr, ua->Address.iSockaddrLength);
+                            if (addr.saStorage.ss_family == AF_INET && protocol != EnableIPv6)
                             {
-                                result.push_back(addr);
-                                break; // a single address per interface is sufficient
+                                if (addr.saIn.sin_addr.s_addr != 0)
```
pepone: Can we use `!= INADDR_ANY` to make the check clear?

cpp/src/Ice/Network.cpp
pepone: Can we check the return of insert instead of having a separate call to find? interfaces is a set, so duplicates are not accepted, something like:

```
// one address per interface name
if (auto [_, inserted] = interfaces.insert(curr->ifa_name); inserted)
{
    result.push_back(addr);
}
```

cpp/src/Ice/Network.cpp
```diff
@@ -999,21 +977,27 @@ IceInternal::isAddressValid(const Address& addr)
 vector<string>
 IceInternal::getInterfacesForMulticast(const string& intf, ProtocolSupport protocolSupport)
 {
-    vector<string> interfaces;
     bool ipv4Wildcard = false;
     if (isWildcard(intf, protocolSupport, ipv4Wildcard))
     {
+        vector<string> interfaces;
         vector<Address> addrs = getLocalAddresses(ipv4Wildcard ? EnableIPv4 : protocolSupport);
```
pepone, line 978: Not changed by this PR, but what is the logic for the ipv4Wildcard check here, why don't we always use the protocolSupport parameter?

pepone (approved):

scripts/DataStormUtil.py
```diff
@@ -38,9 +31,13 @@ def getEffectiveProps(self, current, props):
...
-            )
+            # We can't use --interface 127.0.0.1 on Linux: the loopback interface may not support multicast.
+            if isinstance(platform, Linux):
```
externl, line 35: If possible it would be nice to use the same endpoints on all platforms.

externl (approved):

cpp/src/Ice/Network.cpp
bernardnormier: Fixed as suggested.

cpp/src/Ice/Network.cpp
bernardnormier: I didn't look into it.

cpp/src/Ice/Network.cpp
bernardnormier: Fixed.

scripts/DataStormUtil.py
bernardnormier: Will fix in a follow-up PR.

PR #4537, opened 2025-10-13
collect android simulator logs

externl:
We now collect the logs from the adb process controller logs.


scripts/Util.py
```diff
@@ -2545,10 +2548,6 @@ def startEmulator(self, avd):
             raise RuntimeError(f"emulator '{avd}' not booted after 300s")
 
     def startControllerApp(self, current, ident):
-        # Stop previous controller app before starting new one
-        for ident in self.controllerApps:
```
externl, line 2549: This was triggering a restart every time we started as we always add to the list before calling start.

PR #4541, opened 2025-10-14
Links Should Target the Async Versions of Operations in C# and Python

InsertCreativityHere:
Are there any other languages that we should change as well?


cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1722,7 +1722,7 @@ Slice::Csharp::csLinkFormatter(const string& rawLink, const ContainedPtr& source
         {
             // link to the method on the proxy interface
             result << getUnqualified(operationTarget->interface(), sourceScope) << "Prx."
-                   << operationTarget->mappedName();
+                   << operationTarget->mappedName() + "Async";
```
externl:
```suggestion
                   << operationTarget->mappedName() << "Async";
```

Issue #4528, opened 2025-10-10
Always generate AMD skeletons


bernardnormier:
Currently, when you define an interface, you decide which operation is going to be implemented synchronously (the default) or asynchronously (with the amd metadata directive).

However, you are often in the situation where you don't know: you specify the interface from the client point-of-view, and you don't know (and should not worry about) server-side implementation details.

Ideally, there would be no "amd" and all skeletons would be AMD-only. However, making everything AMD-only would break source compatibility and require significant refactoring of existing code, so that's not a viable solution.

This issue affects C++, C#, and Java. In JS/TS, Python, and Swift, you can implement any skeleton method synchronously or asynchronously.

## Proposal

In C++, C#, and Java, always generate an Async skeleton in addition to the regular skeleton.

For example:
Slice interface Greeter => Greeter + AsyncGreeter skeletons in C++.

In C#, it would be Greeter, GreeterDisp_ (default) + AsyncGreeter, AsyncGreeterDisp_ (all async).

If you put "amd" on your interface, the default and the Async skeletons would be identical, except for their names.

The base class (Ice.Object), which we also use as the dispatcher interface, would remain unique and sync-only for ice_ping and friends. So no AsyncObject.

PR #4542, opened 2025-10-14
Generate async skeletons in C#

bernardnormier:
This PR updates slice2cs to always generate async skeletons (see #4528).

It also removes the various TestAMD.ice from the tests as they are no longer needed.


cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -43,24 +43,30 @@ Slice::Csharp::getNamespace(const ContainedPtr& p)
 }
 
 string
-Slice::Csharp::getUnqualified(const ContainedPtr& p, const string& package)
+Slice::Csharp::getUnqualified(const ContainedPtr& p, const string& package, const string& prefix)
 {
+    string name = p->mappedName();
+    if (!prefix.empty())
+    {
+        name = prefix + removeEscapePrefix(name);
```
bernardnormier, line 51: The difficulty here is escaped keywords such as `@case`. `Async@case` is not a valid identifier.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -123,14 +124,16 @@ namespace
         result[pos] = '.';
         return result;
     }
+    */
 
-    void writeSeeAlso(Output& out, const StringList& lines)
+    void writeSeeAlso([[maybe_unused]] Output& out, const StringList& lines)
```
bernardnormier, line 129: The seeAlso C# in mapping doesn't quite work at present.

cpp/src/slice2cs/Gen.h
```diff
@@ -128,10 +122,10 @@ namespace Slice
             void visitOperation(const OperationPtr&) final;
         };
 
-        class ServantVisitor final : public CsVisitor
+        class SkeletonVisitor final : public CsVisitor
```
bernardnormier, line 125: Renamed Servant => Skeleton.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -559,47 +562,6 @@ Slice::CsVisitor::getInArgs(const OperationPtr& op, bool internal)
     return args;
 }
 
-string
-Slice::CsVisitor::getDispatchParams(
```
bernardnormier, line 563: Moved to the SkeletonVisitor class.

csharp/test/Ice/dictMapping/MyClassAMDI.cs
```diff
@@ -1,48 +1,44 @@
 // Copyright (c) ZeroC, Inc.
 
-namespace Ice.dictMapping
+namespace Ice.dictMapping.AMD;
+
+public sealed class MyClassI : Test.AsyncMyClassDisp_
 {
-    namespace AMD
```
bernardnormier, line 5: Changed the namespace syntax and reformat document. No other changes.

csharp/test/Ice/metrics/ServerAMD.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System.Reflection;
+
+[assembly: CLSCompliant(true)]
+
+[assembly: AssemblyTitle("IceTest")]
+[assembly: AssemblyDescription("Ice test")]
+[assembly: AssemblyCompany("ZeroC, Inc.")]
+
+public class Server : Test.TestHelper
```
bernardnormier, line 11: This file is identical to Server.cs. Without it, the test suite doesn't execute the client/amd server test.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -2618,3 +2596,56 @@ Slice::Gen::ServantVisitor::writeDispatch(const InterfaceDefPtr& p)
...
+Slice::Gen::SkeletonVisitor::getDispatchParams(
+    const OperationPtr& op,
+    string& retS,
+    vector<string>& params,
+    vector<string>& args,
+    const string& ns)
+{
+    string name = op->mappedName();
+    InterfaceDefPtr interface = op->interface();
+    ParameterList parameterss;
```
pepone: What is the point of the "parameterss" variable, seems we never read it.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -2618,3 +2596,56 @@ Slice::Gen::ServantVisitor::writeDispatch(const InterfaceDefPtr& p)
...
+    }
+
+    string currentParamName = getEscapedParamName(op->parameters(), "current");
+    params.push_back("Ice.Current " + currentParamName);
+    args.push_back(currentParamName);
+    return name;
+}
+
+string
+Slice::Gen::SkeletonVisitor::prependAsyncPrefix(const string& name) const
```
pepone: maybe is clear to rename these methods `prependSkeletonPrefix` and `skeletonPrefix` the skeleton prefix is either empty string or Async, implementation remains the same.

csharp/test/Ice/metrics/ServerAMD.cs
pepone: We can fix this in a separate PR.

pepone (approved):

cpp/src/slice2cs/Gen.cpp
```diff
@@ -2618,3 +2596,56 @@ Slice::Gen::ServantVisitor::writeDispatch(const InterfaceDefPtr& p)
...
+    }
+    else
+    {
+        params = getParams(op, ns);
+        args = getArgs(op);
+        parameterss = op->parameters();
+        retS = typeToString(op->returnType(), ns, op->returnIsOptional());
+    }
+
+    string currentParamName = getEscapedParamName(op->parameters(), "current");
```
externl: Did you mean to use the `parameters` variable here?

externl (approved):

csharp/test/Ice/metrics/ServerAMD.cs
bernardnormier: Not sure what's the best fix is.

cpp/src/slice2cs/Gen.cpp
bernardnormier: See @externl comment below.

cpp/src/slice2cs/Gen.cpp
bernardnormier: Fixed.

cpp/src/slice2cs/Gen.cpp
bernardnormier: Fixed. Renamed parameterList. This bug was in the previous code.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -2618,3 +2596,56 @@ Slice::Gen::ServantVisitor::writeDispatch(const InterfaceDefPtr& p)
...
+Slice::Gen::SkeletonVisitor::prependSkeletonPrefix(const string& name) const
+{
+    return _async ? "Async" + removeEscapePrefix(name) : name;
+}
+
+string
+Slice::Gen::SkeletonVisitor::skeletonPrefix() const
+{
+    return _async ? "Async" : "";
+}
```
InsertCreativityHere, line 2651: These functions are in the opposite order in the header file, would be nice for the order to match.

InsertCreativityHere (approved):
Slice compiler changes look good, I trust the mechanistic file changes are fine.

cpp/src/slice2cs/Gen.cpp
bernardnormier: Will fix in follow-up PR.

PR #4549, opened 2025-10-15
Always generate async skeletons in Java

bernardnormier:
This PR updates slice2java to always generate async skeletons.

See #4528.


externl (approved):

cpp/src/slice2java/Gen.cpp
```diff
@@ -4783,15 +4790,43 @@ Slice::Gen::ServantVisitor::visitOperation(const OperationPtr& p)
...
+Slice::Gen::SkeletonVisitor::skeletonPrefix() const
+{
+    return _async ? "Async" : "";
+}
+
+string
+Slice::Gen::SkeletonVisitor::prependSkeletonPrefix(const string& name) const
+{
+    return skeletonPrefix() + name;
+}
```
InsertCreativityHere, line 4823: This function is a one-liner, that's only called in 2 places; can we just inline it for `slice2java`?

java/test/src/main/java/test/Ice/slicing/objects/AMDTestI.java
```diff
@@ -6,37 +6,38 @@
...
+import test.Ice.slicing.objects.server.Test.PSUnknownException;
+import test.Ice.slicing.objects.server.Test.Preserved;
+import test.Ice.slicing.objects.server.Test.SBSKnownDerived;
+import test.Ice.slicing.objects.server.Test.SBSUnknownDerived;
+import test.Ice.slicing.objects.server.Test.SBase;
+import test.Ice.slicing.objects.server.Test.SS1;
+import test.Ice.slicing.objects.server.Test.SS2;
+import test.Ice.slicing.objects.server.Test.SS3;
+import test.Ice.slicing.objects.server.Test.SUnknown;
+import test.Ice.slicing.objects.server.Test.TestIntf;
```
InsertCreativityHere, line 32: Originally, I was curious if we still need this import, since we now implement `AsyncTestIntf`.
But, this import is still used in the code, for things like: `TestIntf.ParamTest1Result`, etc.

Obviously, this works just fine, but it feels weird to use the types from `TestIntf` when implementing `AsyncTestIntf`...

InsertCreativityHere (approved):
Looks good to me.

pepone (approved):

java/test/src/main/java/test/Ice/slicing/objects/AMDTestI.java
bernardnormier: It's because these nested result types are used by the skeletons/servants _and_ by the proxy code. It would not make sense to duplicate these result type classes in the default and async skeletons.

cpp/src/slice2java/Gen.cpp
bernardnormier: As a general rule, we don't inline code because it's short.

InsertCreativityHere: Sure, of course we don't. I'm saying that generally code is easier to follow without extra functions between the caller and the outcome, it adds indirection.

For the other languages, this is offset, because the function body is more complex.
But when the body is just `skeletonPrefix() + name`, putting this behind a function named `prependSkeletonPrefix` is redundant, no different than the function `addOne(int i) { return i + 1; }`.

I appreciate that it's more consistent with the other compilers, and I love consistency, but I learned long ago that trying to keep the compilers consistent just makes everything more complex and harder to unravel in the long run.

Issue #4543, opened 2025-10-14
Fix doc-comment generation for @see


bernardnormier:
The Slice Parser does not handle `@see` properly for doc-comment generation.

It should map the string argument to a Parser symbol, not to a string.

InsertCreativityHere:
Fixed in #4550.

PR #4550, opened 2025-10-15
Fix '@see' Tag Mapping and Some Broken References

InsertCreativityHere:
This PR fixes the generation of '@see' tags in each language by adding a new function (`formatSeeAlso`) to the `DocCommentFormatter` class, which each compiler implements.

Note that currently, we don't map '@see' at all in Python, PHP, Ruby, or Swift. So these languages weren't touched by this PR.

The parser handles looking up (and validating) the referenced Slice element, then passing it to this function.

See #4543.

----

I ran this PR through my comparison script to make sure everything looked good. Here's some diffs in the generated code to show the changes:

## C++

```diff
@@ -1218,7 +1218,7 @@ namespace DataStormContract
     std::ostream& operator<<(std::ostream& os, const ElementInfo& value);

     /// Contains metadata about a topic, including its name and associated reader/writer IDs.
-    /// @see Session#announceTopics
+    /// @see SessionPrx::announceTopics
     struct TopicInfo
```

## C#

```diff
@@ -1234,8 +1234,8 @@ namespace DataStormContract
     /// This interface specifies the operations for communication between publisher and subscriber sessions.</summary>
     /// <remarks>The Slice compiler generated this client-side interface from Slice interface <c>::DataStormContract::Session</c>.
     /// Use the methods of this interface to invoke operations on a remote Ice object that implements <c>Session</c>.</remarks>
-    /// <seealso cref="PublisherSession" />
-    /// <seealso cref="SubscriberSession" />
+    /// <seealso cref="PublisherSessionPrx" />
+    /// <seealso cref="SubscriberSessionPrx" />
     public partial interface SessionPrx : Ice.ObjectPrx
```
```diff
@@ -1249,7 +1249,7 @@ namespace DataStormContract
         /// <param name="topics">The sequence of topics to announce.</param>
         /// <param name="initialize">Currently unused.</param>
         /// <param name="context">The request context.</param>
-        /// <seealso cref="attachTopic" />
+        /// <seealso cref="global::DataStormContract.SessionPrx.attachTopic" />
         void announceTopics(
```

## Java

```diff
@@ -26,7 +26,7 @@ public interface Node extends com.zeroc.Ice.Object {
      * @return A completion stage that the servant will complete when the invocation completes.
      * @throws DataStormContract.SessionCreationException Thrown when the session cannot be created.
      *
-     * @see Lookup::announceTopicReader
+     * @see DataStormContract.LookupPrx#announceTopicReader
      */
     java.util.concurrent.CompletionStage<Void> initiateCreateSessionAsync(
```
```diff
@@ -8,7 +8,7 @@ package IceStorm;
 /**
  * Internal operations for a topic.
  *
- * @see Topic
+ * @see com.zeroc.IceStorm.TopicPrx
  */
 @com.zeroc.Ice.SliceTypeId(value = "::IceStorm::TopicInternal")
```

## JavaScript

```diff
@@ -932,8 +932,8 @@ export namespace DataStormContract {
      * The base interface for publisher and subscriber sessions.
      *
      * This interface specifies the operations for communication between publisher and subscriber sessions.
-     * @see PublisherSession
-     * @see SubscriberSession
+     * @see {@link DataStormContract.PublisherSessionPrx}
+     * @see {@link DataStormContract.SubscriberSessionPrx}
      */
     export class SessionPrx extends __module__zeroc_ice.Ice.ObjectPrx {
```
```diff
@@ -9,7 +9,7 @@ declare module "@zeroc/ice" {
         /**
          * The exception that is thrown when a client tries to destroy a session with a router, but no session exists for
          * this client.
-         * @see Router#destroySession
+         * @see {@link Glacier2.RouterPrx.destroySession}
          */
         export class SessionNotExistException extends Ice.UserException {}
```
```diff
+++ b/slice2js/slice/IceBox/ServiceManager.d.ts
@@ -23,7 +23,7 @@ declare module "@zeroc/ice" {

         /**
          * Observes the status of services in an IceBox server.
-         * @see ServiceManager#addObserver
+         * @see {@link IceBox.ServiceManagerPrx.addObserver}
          * @remarks
          * This interface is implemented by admin tools that monitor the IceBox server.
          */
```

## MATLAB

```diff
@@ -184,7 +184,7 @@ classdef SessionPrx < Glacier2.SessionPrx
             %
             %   See also allocateObjectByType, Ice.Future.
             %
-            %   See also setAllocationTimeout, releaseObject
+            %   See also IceGrid.SessionPrx/setAllocationTimeout, IceGrid.SessionPrx/releaseObject
```
```diff
@@ -16,7 +16,7 @@ classdef (Sealed) SSLInfo
     %     cipher - The negotiated cipher suite.
     %     certs - The certificate chain.
     %
-    %   See also SSLPermissionsVerifier
+    %   See also Glacier2.SSLPermissionsVerifierPrx
```

### DIFF
--- a/cpp/src/ice2slice/Gen.cpp
+++ b/cpp/src/ice2slice/Gen.cpp
@@ -287,9 +287,9 @@ namespace
         }
 
         // 'seeAlso' links have already been converted to Slice2 syntax by `formatSeeAlso`.
-        for (const string& correctedLink : comment->seeAlso())
+        for (const string& mappedLink : comment->seeAlso())
         {
-            out << nl << "/// @see " << correctedLink;
+            out << nl << "/// @see " << mappedLink;
         }
     }
 }


cpp/src/Ice/ThreadPool.cpp
```diff
@@ -186,11 +186,6 @@ IceInternal::ThreadPool::initialize()
     }
 #endif
 
-    //
-    // We use just one thread as the default. This is the fastest
-    // possible setting, still allows one level of nesting, and
-    // doesn't require to make the servants thread safe.
-    //
```
InsertCreativityHere, line 193: After stumbling upon this in Java, we decided that this comment is bogus and best deleted from all languages.

cpp/src/IceGrid/FileCache.cpp
```diff
@@ -4,7 +4,7 @@
 #include "../Ice/FileUtil.h"
 #include "Ice/Communicator.h"
 #include "Ice/Properties.h"
-#include "IceGrid/Exception.h"
```
InsertCreativityHere, line 7: I had to move some exceptions from `Exception.ice` to `Admin.ice` to fix some circular `@see` references.
You'll see the swap further down in this PR.

cpp/src/IceGrid/Parser.cpp
```diff
@@ -2662,10 +2662,6 @@ Parser::exception(std::exception_ptr pex)
     {
         error("registry '" + ex.name + "' couldn't be reached:\n" + ex.reason);
     }
-    catch (const ServerUnreachableException& ex)
-    {
-        error("server '" + ex.name + "' couldn't be reached:\n" + ex.reason);
-    }
```
InsertCreativityHere, line 2668: This exception is defined in Slice, but is never thrown or used anywhere except for this one place.
Removed the exception for being dead.

cpp/src/Slice/Parser.h
```diff
@@ -390,13 +390,13 @@ namespace Slice
         /// Returns the mapped identifier that this element will use in the target language.
         [[nodiscard]] std::string mappedName() const;
```
InsertCreativityHere, line 392: Bernard realized these doc-comments were actually flipped!
Unrelated to these changes, but good to fix.

slice/Ice/PropertyDict.ice
```diff
@@ -18,7 +18,6 @@ module Ice
 {
     /// A simple collection of properties, represented as a dictionary of key/value pairs. Both key and value are
     /// strings.
-    /// @see PropertiesAdmin#getPropertiesForPrefix
```
InsertCreativityHere, line 21: There is no way for this `@see` to work unfortunately.
It references an operation (can't be forward declared) and `#include`ing the file where `PropertiesAdmin` is defined would be circular.

slice/IceGrid/Admin.ice
```diff
@@ -208,6 +208,76 @@ module IceGrid
         ApplicationUpdateDescriptor descriptor;
     }
```
InsertCreativityHere, line 210: These exceptions were copy-pasted (with no changes) from `Exception.ice`.
`Admin.ice` is the only Slice file where they're used, and moving them was necessary for some `@see` in `Admin.ice` to link properly.

cpp/src/Ice/ThreadPool.cpp
bernardnormier: It's preferable to not include random fixes in this Slice doc-comment PR.

cpp/src/slice2cs/CsUtil.h
```diff
@@ -80,9 +80,17 @@ namespace Slice::Csharp
...
         const std::string& customStream = "");
 
-    /// Returns a C# formatted link to the provided Slice identifier.
-    std::string
-    csLinkFormatter(const std::string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target);
+    /// Converts a Slice-formatted link into a C# formatted link.
+    /// @param rawLink The references's raw text, taken verbatim from the doc-comment.
+    /// @param source A pointer to the Slice element that the doc-comment (and reference) are written on.
+    /// @param target A pointer to the Slice element that is being referenced, or `nullptr` if it doesn't exist.
+    /// @param mappedLink Output parameter where the formatted link is written to.
```
bernardnormier: I am not a fan of out parameters in new code. Just return a pair.

And if you really want an out parameter, use the doxygen out syntax to document it properly.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -104,9 +104,16 @@ namespace
...
     }
 
-    void writeSeeAlso([[maybe_unused]] Output& out, [[maybe_unused]] const StringList& lines)
+    void writeSeeAlso(Output& out, const StringList& seeAlso)
     {
-        // See #4543.
+        for (const string& line : seeAlso)
+        {
+            out << nl << "///";
+            if (!line.empty())
```
bernardnormier, line 113: If an empty line is possible, add a comment describing when.

cpp/src/slice2java/JavaUtil.h
```diff
@@ -116,9 +116,17 @@ namespace Slice
...
-        /// Returns a javadoc formatted link to the provided Slice identifier.
-        std::string
-        javaLinkFormatter(const std::string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target);
+        /// Converts a Slice-formatted link into a JavaDoc formatted link.
+        /// @param rawLink The references's raw text, taken verbatim from the doc-comment.
+        /// @param source A pointer to the Slice element that the doc-comment (and reference) are written on.
+        /// @param target A pointer to the Slice element that is being referenced, or `nullptr` if it doesn't exist.
+        /// @param mappedLink Output parameter where the formatted link is written to.
+        /// @returns @c false if the link was to a Slice element which isn't mapped in Java, @c true otherwise.
+        bool javaLinkFormatter(
```
bernardnormier: See C# comment.

bernardnormier (approved):

cpp/src/slice2cs/CsUtil.h
InsertCreativityHere: Updated to return a pair.

cpp/src/slice2java/JavaUtil.h
InsertCreativityHere: Updated to return a pair.

cpp/src/slice2cs/Gen.cpp
InsertCreativityHere: Fixed.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1719,8 +1718,8 @@ Slice::Csharp::csLinkFormatter(const string& rawLink, const ContainedPtr& source
         if (dynamic_pointer_cast<Sequence>(contained) || dynamic_pointer_cast<Dictionary>(contained))
         {
             // slice2cs doesn't generate C# types for sequences or dictionaries, so there's nothing to link to.
-            // Instead, we just output the sequence or dictionary name in code formatting.
-            return "<c>" + getUnqualified(contained, sourceScope) + "</c>";
+            // So, we return 'false' to signal this, and just output the sequence or dictionary name.
+            return make_pair(false, getUnqualified(contained, sourceScope));
```
externl: I believe you can now do:

```suggestion
            return {false, getUnqualified(contained, sourceScope)};
```

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -1764,6 +1763,5 @@ Slice::Csharp::csLinkFormatter(const string& rawLink, const ContainedPtr& source
         result << "cref=\"" << targetS << "\"";
     }
 
-    result << " />";
-    return result.str();
+    return make_pair(true, result.str());
```
externl:
```suggestion
    return {true, result.str()};
```

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -598,52 +598,49 @@ Slice::Java::getSequenceTypes(
...
+        mappedLink = getUnqualified(contained, sourceScope);
+
         if (dynamic_pointer_cast<Sequence>(contained) || dynamic_pointer_cast<Dictionary>(contained))
         {
-            // slice2java doesn't generate Java types for sequences or dictionaries, so there's nothing to link to.
-            // Instead, we just output the sequence or dictionary name in code formatting.
-            return "{@code " + getUnqualified(contained, sourceScope) + "}";
+            // slice2java doesn't generate types for sequences or dictionaries, so there's nothing to link to.
+            // We return 'false' to signal this.
+            return make_pair(false, mappedLink);
```
externl:
```suggestion
            return {false, mappedLink};
```

cpp/src/slice2java/JavaUtil.cpp
```diff
@@ -598,52 +598,49 @@ Slice::Java::getSequenceTypes(
...
     }
     else
     {
-        result << rawLink;
+        mappedLink = rawLink;
     }
 
-    result << '}';
-    return result.str();
+    return make_pair(true, mappedLink);
```
externl:
```suggestion
    return {true, mappedLink};
```

cpp/src/slice2cs/CsUtil.h
InsertCreativityHere: Just for the record, this is what the rendered Doxygen looks like:
image

externl (approved):

cpp/src/ice2slice/Gen.cpp
```diff
@@ -286,33 +286,10 @@ namespace
             }
         }
 
-        for (string ident : comment->seeAlso())
+        // 'seeAlso' links have already been converted to Slice2 syntax by `formatSeeAlso`.
+        for (const string& correctedLink : comment->seeAlso())
```
pepone: niptick: I would use resolvedLink.

cpp/src/IceGrid/FileCache.cpp
pepone: Why don't forward declare them?

cpp/src/slice2cs/Gen.cpp
pepone: What is the point of adding the empty lines to seeAlso list?

cpp/src/slice2js/Gen.cpp
```diff
@@ -2292,10 +2279,9 @@ Slice::Gen::TypeScriptVisitor::writeOpDocSummary(Output& out, const OperationPtr
 
         writeOpDocExceptions(out, op, *comment);
 
-        const StringList& seeAlso = comment->seeAlso();
-        if (!seeAlso.empty())
+        for (const string& line : comment->seeAlso())
```
pepone, line 2282: No empty line checks here?

cpp/src/IceGrid/FileCache.cpp
InsertCreativityHere: We don't allow forward declaring exceptions, only classes & exceptions can be forward declared in Slice.

pepone: Forgot exceptions cannot be forward declared, just classes and interfaces.

cpp/src/slice2js/Gen.cpp
InsertCreativityHere: Yes, we only need the check for C#, because of how strict it's XML `seealso` is.
The other languages aren't as strict, with what can come after an `@see`, so we can always map `@see` to something, even if it's just code-formatted text.

cpp/src/IceGrid/FileCache.cpp
InsertCreativityHere: I wish that we did! Would make this easier!
But the real fix is to do what we did for Slice2, and not care about declaration order
: vP

cpp/src/slice2cs/Gen.cpp
InsertCreativityHere: That was leftover from a bad-merge, since Bernard replaced all this code yesterday.
Removed.

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -283,64 +283,25 @@ namespace
...
-            if (p != seeAlso.begin())
-            {
-                out << ", ";
-            }
-
-            string sym = *p;
-            string rest;
-            string::size_type pos = sym.find('#');
-            if (pos != string::npos)
+            // All references must be on one line.
```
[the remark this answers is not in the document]
InsertCreativityHere: Updated as suggested, but also added `This is the conventional way to format 'See also' in MATLAB.` to explain _why_ we do it differently than every other compiler.

cpp/src/ice2slice/Gen.cpp
InsertCreativityHere: How about `mappedLink`? Because it isn't necessarily a 'resolvedLink'.
Even if we can't resolve the type you're pointing to, we'll still make a best-effort attempt to map it from Ice to Slice2.

pepone: mappedLink sounds fine too.

InsertCreativityHere: Changed it to `mappedLink`!

PR #4548, opened 2025-10-15
Java ThreadPool format cleanup

pepone:
This PR cleanups Java ThreadPool code format, and removes some unused code.

PR #4551, opened 2025-10-15
Abort on fatal error

### LINKED ISSUE #4441: Windows segmentation fault in Selector::getNextHandler

pepone:
This updates C++ and Java selector code to abort/halt on fatal errors. Retrying this failures is of no help. Fixes #4441

It also includes the Java Thread Pool updates from https://github.com/zeroc-ice/ice/pull/4548

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Selector.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Selector.java
@@ -26,10 +26,8 @@ final class Selector {
             throw new SyscallException(ex);
         }
 
-        //
-        // The Selector holds a Set representing the selected keys. The
-        // Set reference doesn't change, so we obtain it once here.
-        //
+        // The Selector holds a set representing the selected keys. The set reference doesn't change, so we obtain it
+        // once here.
         _keys = _selector.selectedKeys();
     }
 
@@ -182,6 +180,8 @@ final class Selector {
             if (_selectNow) {
                 _selector.selectNow();
             } else if (timeout > 0) {
+                // If interrupted is set, it means that wakeup was called, and select should return immediately
+                // without changes, allow the caller to go to finish select and process the ready handlers.
                 if (_selector.select(timeout * 1000) == 0 && !_interrupted) {
                     throw new TimeoutException();
                 }
@@ -277,7 +277,7 @@ final class Selector {
     private final HashSet<EventHandler> _readyHandlers = new HashSet<>();
     private boolean _selecting;
     private boolean _selectNow;
-    // Marked as volatile to ensure that when the selector is wakeup, after calling wakeup(), the _interrupted
+    // Marked as volatile to ensure that when the selector is woken up, after calling wakeup(), the _interrupted
     // change is visible to the thread blocked in select().
     private volatile boolean _interrupted;
     private int _spuriousWakeUp;


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ThreadPool.java
```diff
@@ -59,16 +58,7 @@ public void execute(ThreadPoolCurrent current) {
...
 
-    static final class InterruptWorkItem implements ThreadPoolWorkItem {
-        @Override
-        public void execute(ThreadPoolCurrent current) {
-            // Nothing to do, this is just used to interrupt the thread pool selector.
-        }
-    }
-
-    //
     // Exception raised by the thread pool work queue when the thread pool is destroyed.
```
InsertCreativityHere, line 61: Even if this isn't public, it's clearly meant to be a doc-comment, not a 'normal' comment.
```suggestion
    /** Exception raised by the thread pool work queue when the thread pool is destroyed. */
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ThreadPool.java
```diff
@@ -590,9 +493,7 @@ private synchronized boolean followerWait(ThreadPoolCurrent current) {
                 try {
                     wait();
                 } catch (InterruptedException e) {
-                    //
-                    // Eat the InterruptedException.
-                    //
+                    // Ignore
```
InsertCreativityHere, line 496: There's actually a convention in Java to name the exception `ignored`, instead of `e` and adding a comment.
Many IDEs (and even checkstyle) are aware of it.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Selector.java
```diff
@@ -318,6 +277,8 @@ private int fromJavaOps(int o) {
     private final HashSet<EventHandler> _readyHandlers = new HashSet<>();
     private boolean _selecting;
     private boolean _selectNow;
-    private boolean _interrupted;
+    // Marked as volatile to ensure that when the selector is wakeup, after calling wakeup(), the _interrupted
```
bernardnormier: is woken up

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Selector.java
```diff
@@ -179,57 +178,19 @@ void finishSelect(List<EventHandlerOpPair> handlers) {
...
-                    } catch (Throwable ex1) {
-                        System.out.println(s);
-                    }
-                } catch (Throwable ex2) {
-                    // Ignore
+        try {
+            if (_selectNow) {
+                _selector.selectNow();
+            } else if (timeout > 0) {
+                if (_selector.select(timeout * 1000) == 0 && !_interrupted) {
```
bernardnormier, line 185: If _interrupted is true, we return successfully from select. Is that the desired behavior? Maybe add a comment?

pepone: If `_interrupted` is true, means we have call `_selector.wakeup()`, which causes select to unblock.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Selector.java
```diff
@@ -260,9 +221,7 @@ private void updateSelector() {
     private void checkReady(EventHandler handler) {
         if ((handler._ready & ~handler._disabled & handler._registered) != 0) {
             _readyHandlers.add(handler);
-            if (_selecting) {
-                wakeup();
-            }
+            wakeup();
```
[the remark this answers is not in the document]
pepone: wakeup has the same check.

PR #4559, opened 2025-10-16
Better iOS simulator boot

externl:
For reasons unknown `xcrun simctl bootstatus` seems to hang quite often on the GitHub runner. When I tried printing the simulator state after a timeout the simulator seems to be booted. 

Instead of using this command, this PR polls the boot status (similar to what we do for Android).

It also adds a log file for the Python  driver communicator.

### DIFF
--- a/scripts/Util.py
+++ b/scripts/Util.py
@@ -2701,7 +2701,7 @@ class iOSSimulatorDevice:
             return device["state"] == "Booted"
         return False
 
-    def get() -> dict[str, str] | None:
+    def get(self) -> dict[str, str] | None:
         output = run("xcrun simctl list devices --json")
         data = json.loads(output)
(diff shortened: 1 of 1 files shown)


scripts/Util.py
```diff
@@ -2691,6 +2682,37 @@ def stopControllerApp(self, ident):
             pass
 
 
+@dataclass
+class iOSSimulatorDevice:
+    """Represents a single iOS Simulator device output by `xcrun simctl list devices --json`. Only
+    the fields we care about are included here.
+    """
```
pepone:
```suggestion
    """
    Represents a single iOS Simulator device output by `xcrun simctl list devices --json`. Only
    the fields we care about are included here.
    """
```

pepone (approved):

PR #4560, opened 2025-10-16
Rework Properties in C#

bernardnormier:
This PR simplifies the Properties constructors in C#, and un-deprecates the args-parsing Util.createProperties methods.

### DIFF
--- a/csharp/src/Ice/Properties.cs
+++ b/csharp/src/Ice/Properties.cs
@@ -11,25 +11,6 @@ using System.Text.RegularExpressions;
 namespace Ice;
 
 /// <summary>
-/// Represents a set of options that you can specify when creating a <see cref="Properties"/> object.
-/// </summary>
-public sealed record class PropertiesOptions
-{
-    /// <summary>
-    ///  Gets or sets the initial properties of the new Properties object.
-    /// </summary>
-    /// <value>The initial properties.</value>
-    public Properties? defaults { get; set; }
-
-    /// <summary>
-    /// Gets or sets a list of prefixes to allow in the new Properties object. Setting a property with a property
-    /// prefix that is opt-in (e.g. IceGrid, IceStorm, Glacier2, etc.) but not in this list is considered an error.
-    /// </summary>
-    /// <value>The list of opt-in prefixes.</value>
-    public ImmutableList<string> optInPrefixes { get; set; } = ImmutableList<string>.Empty;
-}
-
-/// <summary>
 /// Represents a set of properties used to configure Ice and Ice-based applications. A property is a key/value pair,
 /// where both the key and the value are strings. By convention, property keys should have the form
 /// <c>application-name>[.category[.sub-category]].name</c>.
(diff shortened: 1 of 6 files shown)


csharp/src/Ice/Properties.cs
```diff
@@ -3,32 +3,44 @@
...
 
 /// <summary>
-/// A property set used to configure Ice and Ice applications. Properties are key/value pairs, with both keys and
-/// values being strings.
+/// Represents a set of options that you can specify when creating a <see cref="Properties"/> object.
 /// </summary>
-public sealed class Properties
+public sealed record class PropertiesOptions
 {
-    private class PropertyValue
```
bernardnormier, line 18: Moved to the bottom of the inner class.

csharp/src/Ice/Properties.cs
```diff
@@ -3,32 +3,44 @@
...
-        {
-            this.value = value;
-            this.used = used;
-        }
+    /// <summary>
+    /// Gets or sets a list of prefixes to allow in the new Properties object. Setting a property with a property
+    /// prefix that is opt-in (e.g. IceGrid, IceStorm, Glacier2, etc.) but not in this list is considered an error.
+    /// </summary>
+    /// <value>The list of opt-in prefixes.</value>
+    public ImmutableList<string> optInPrefixes { get; set; } = ImmutableList<string>.Empty;
```
bernardnormier: The previous `List<string>` was not quite correct.

csharp/src/Ice/Properties.cs
```diff
@@ -3,32 +3,44 @@
...
+/// <summary>
+/// Represents a set of properties used to configure Ice and Ice-based applications. A property is a key/value pair,
+/// where both the key and the value are strings. By convention, property keys should have the form
+/// <c>application-name>[.category[.sub-category]].name</c>.
+/// </summary>
+/// <remarks>This class is thread-safe: multiple threads can safely read and write the properties without their own
+/// synchronization.</remarks>
+public sealed class Properties
+{
+    private readonly Dictionary<string, PropertyValue> _propertySet = [];
```
bernardnormier, line 22: Renamed _properties to _propertySet as I found `_properties` confusing.

csharp/src/Ice/Properties.cs
```diff
@@ -518,6 +473,67 @@ public List<string> getUnusedProperties()
...
+    /// <summary>
+    /// Creates a new Properties object, loads the configuration files specified by the Ice.Config property or the
+    /// ICE_CONFIG environment variable, and parses the Ice command-line options.
+    /// </summary>
+    /// <param name="args">Command lines options.</param>
+    /// <param name="options">Options to configure the new Properties objects.</param>
+    /// <returns>A new Properties object.</returns>
+    internal static Properties loadConfigAndParseIceOptions(ref string[] args, PropertiesOptions options)
+    {
+        var properties = new Properties(options);
```
bernardnormier: All the logic from the previous "args parsing constructor" moved here without change.

csharp/src/iceboxnet/Server.cs
```diff
@@ -56,7 +58,9 @@ public static int Main(string[] args)
 
         var initData = new Ice.InitializationData
         {
-            properties = new Ice.Properties(ref args, new Ice.Properties(["IceBox"]))
+            properties = Ice.Util.createProperties(
```
bernardnormier: A more elegant way to create properties without default properties.

csharp/src/Ice/Properties.cs
```diff
@@ -38,80 +50,23 @@ public Properties()
...
-            //
-            // If Ice.Config is not set, load from ICE_CONFIG (if set)
-            //
-            loadConfigFiles = !_properties.ContainsKey("Ice.Config");
-        }
-
-        if (loadConfigFiles)
-        {
-            loadConfig();
+            _optInPrefixes.AddRange(defaults._optInPrefixes);
```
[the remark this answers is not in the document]
bernardnormier: Fixed.

csharp/src/Ice/Util.cs
```diff
@@ -7,46 +7,49 @@
...
+/// Utility methods for the Ice runtime.
 /// </summary>
 public sealed class Util
 {
     /// <summary>
     /// Creates a new empty property set.
     /// </summary>
     /// <returns>A new empty property set.</returns>
-    [Obsolete("Use Ice.Properties() constructor instead.")]
+    [Obsolete("Use the Ice.Properties() constructor instead.")]
```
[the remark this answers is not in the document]
bernardnormier: It un-deprecates the createProperties-that-parse args. Maybe it's inconsistent and I should un-deprecate `createProperties()` too?

csharp/src/Ice/Util.cs
```diff
@@ -7,46 +7,49 @@
...
+    /// <param name="args">The command-line arguments. This method parses arguments starting with `--` and one of the
+    /// reserved prefixes (Ice, IceSSL, etc.) as properties and removes these elements from the list. If there is an
+    /// argument starting with `--Ice.Config`, this method loads the specified configuration file. When the same
+    /// property is set in a configuration file and through a command-line argument, the command-line setting takes
+    /// precedence.</param>
+    /// <param name="defaults">Default values for the new property set. Settings in configuration files and the
+    /// arguments override these defaults.</param>
+    /// <returns>A new property set.</returns>
+    /// <remarks>This method loads properties from files specified by the <c>ICE_CONFIG</c> environment variable when
+    /// there is no <c>--Ice.Config</c> command-line argument. It also gives <c>Ice.ProgramName</c> a default value.
```
pepone: Don't we also check for `Ice.Config` set in the defaults?

```
if (!loadConfigFiles)
        {
            // If Ice.Config is not set, load from ICE_CONFIG (if set).
            loadConfigFiles = !propertySet.ContainsKey("Ice.Config");
        }
```

csharp/src/iceboxnet/Server.cs
pepone: What if we can write something like:

```
(properties, args) = Properties.fromArgs(
    args,
    new Ice.PropertiesOptions { optInPrefixes = ImmutableList.Create("IceBox") })
);
```

csharp/src/Ice/Properties.cs
```diff
@@ -518,6 +473,67 @@ public List<string> getUnusedProperties()
...
     }
 
+    /// <summary>
+    /// Creates a new Properties object, loads the configuration files specified by the Ice.Config property or the
+    /// ICE_CONFIG environment variable, and parses the Ice command-line options.
+    /// </summary>
+    /// <param name="args">Command-line options.</param>
+    /// <param name="options">Options to configure the new Properties object.</param>
+    /// <returns>A new Properties object.</returns>
+    internal static Properties loadConfigAndParseIceOptions(ref string[] args, PropertiesOptions options)
```
externl: Why not keep createProperties deprecated and make this public. 

I find `loadConfigAndParseIceOptions` a bit wordy.

externl (commented):
What is the recommend pattern for creating and using a property set with opt in prefixes set. As far as I can tell this doesn't fix the issue of needing two property objects to use opt-in prefixes.

To me the API seems more confusing now. I'd have to call the properties constructor followed by the util create method.

csharp/src/iceboxnet/Server.cs
bernardnormier: I prefer the `ref args` over the factory method that returns a pair.

csharp/src/Ice/Properties.cs
```diff
@@ -3,32 +3,25 @@
...
 namespace Ice;
 
 /// <summary>
-/// A property set used to configure Ice and Ice applications. Properties are key/value pairs, with both keys and
-/// values being strings.
+/// Represents a set of properties used to configure Ice and Ice-based applications. A property is a key/value pair,
+/// where both the key and the value are strings. By convention, property keys should have the form
+/// <c>application-name>[.category[.sub-category]].name</c>.
 /// </summary>
+/// <remarks>This class is thread-safe: multiple threads can safely read and write the properties without their own
```
externl: 👍

csharp/src/Ice/Util.cs
bernardnormier: Say Ice.Config is set in the defaults and you call createProperties(args) where args contains another `--Ice.Config`. The new `--Ice.Config` would "win" since we don't check Ice.Config until after reparsing this `--Ice.Config` arg; I don't think it's worth mentioning here.

externl (approved):

csharp/src/Ice/Properties.cs
```diff
@@ -3,75 +3,82 @@
...
+/// synchronization.</remarks>
```
pepone: I will remove the without synchronization, that is imply when you say is thread-safe

pepone (approved):

externl (approved):

InsertCreativityHere (approved):

PR #4561, opened 2025-10-17
Fix extraction of Slice files in the Gradle user directory

pepone:
A small for an issue reported by @InsertCreativityHere when trying to include the built-in Slice files.

The files where extracted to `slice-tools-3.8.0-nightly-20251016.1-SNAPSHOT\slice\resources\slice` instead of `slice-tools-3.8.0-nightly-20251016.1-SNAPSHOT\slice\`

### DIFF
--- a/java/tools/slice-tools/src/main/kotlin/com/zeroc/slice/tools/SliceToolsResourceExtractor.kt
+++ b/java/tools/slice-tools/src/main/kotlin/com/zeroc/slice/tools/SliceToolsResourceExtractor.kt
@@ -79,7 +79,9 @@ object SliceToolsResourceExtractor {
                 .toList()
 
             for (resource in resourcePaths) {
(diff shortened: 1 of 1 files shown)


java/tools/slice-tools/src/main/kotlin/com/zeroc/slice/tools/SliceToolsResourceExtractor.kt
```diff
@@ -79,7 +79,8 @@ object SliceToolsResourceExtractor {
                 .toList()
 
             for (resource in resourcePaths) {
-                val relativePath = resource.removePrefix(resourcePath)
+                val relativePath = resource.removePrefix("resources/slice/")
```
pepone: The resource names returned by the jar entries don't start with `/` so this `removePrefix` was doing nothing at the files end-up in the wrong directory.

PR #4563, opened 2025-10-17
Fix DataStorm MessageSizeMax settings

### LINKED ISSUE #4558: DataStorm: set MessageSizeMax to 0 for internal data exchange
### Version

- Ice3.8 preview

### Describe the bug

For IceGrid and IceStorm the MessageSizeMax for internal adapters is set to '0', 'No limit on internal data exchanged'.
It should also be set to '0' for the internal adapters of DataStorm. Testing the case with >ten thousand writers/readers result in a loop.

### Expected behavior

A warning if the max writer/reader objects or the max message size is reached.

### Additional context

_No response_

pepone:
Fix #4558

### DIFF
--- a/config/PropertyNames.xml
+++ b/config/PropertyNames.xml
@@ -408,7 +408,6 @@
         <property name="Node.Server" class="ObjectAdapter" languages="cpp" />
         <property name="Node.Server.Enabled" default="1" languages="cpp" />
         <property name="Node.Server.ForwardDiscoveryToMulticast" default="0" languages="cpp" />
-        <property name="Node.Client.MessageSizeMax" languages="cpp" />
         <property name="Topic.ClearHistory" default="OnAll" languages="cpp" />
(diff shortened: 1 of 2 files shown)


cpp/src/DataStorm/Instance.cpp
```diff
@@ -86,8 +86,10 @@ Instance::Instance(CommunicatorPtr communicator, function<void(function<void()>
     // reference's AdapterId to determine if a collocated call can be used.
     //
     // A named adapter is required because we cannot assign an AdapterId to a nameless adapter.
-    auto collocated = generateUUID();
+    string collocated = generateUUID();
```
bernardnormier, line 89: `generateUUID` is obsolete, so we should not use it in new code.

Then, are you sure this code does something useful?

My understanding is MessageSizeMax is a Connection property, and these is no connection with colloc.

pepone: You are right not needed for collocation.

PR #4566, opened 2025-10-18
Update Properties in Java

bernardnormier:
This is the Java counterpart of #4560.

PR #4567, opened 2025-10-18
Update C++ properties

bernardnormier:
This PR updates Properties in C++ (see #4560 and #4566).

It also simplifies several "argv" functions to use a template instead of 4 overloads.


cpp/include/DataStorm/Node.h
```diff
@@ -112,6 +101,17 @@ namespace DataStorm
         [[nodiscard]] Ice::ConnectionPtr getSessionConnection(std::string_view ident) const noexcept;
 
     private:
+        template<typename ArgvT> NodeOptions createNodeOptions(int& argc, ArgvT argv)
+        {
+            Ice::InitializationData initData;
+            initData.properties = std::make_shared<Ice::Properties>(argc, argv, "DataStorm");
```
bernardnormier, line 107: Can't use the `{.properties =` in a header file since it must be C++17 compatible.

cpp/include/Ice/Properties.h
```diff
@@ -16,34 +16,131 @@
...
+        /// when there is no `--Ice.Config` command-line argument. It also gives `Ice.ProgramName` a default value.
+        template<typename ArgvT>
+        Properties(int& argc, ArgvT argv, const PropertiesPtr& defaults = nullptr) : Properties{defaults}
+        {
+            StringSeq args = argsToStringSeq(argc, argv);
+            loadArgs(args);
+            stringSeqToArgs(args, argc, argv);
+        }
+
+        /// @private
```
bernardnormier, line 104: If we keep DataStorm opt-in, should not be `@private`.

cpp/src/Ice/Initialize.cpp
```diff
@@ -25,119 +25,6 @@ namespace
     Ice::LoggerPtr processLogger;
 }
 
-StringSeq
```
bernardnormier, line 28: Moved to Properties.cpp without change.

php/src/Communicator.cpp
```diff
@@ -476,7 +476,7 @@ ZEND_METHOD(Ice_Communicator, getProperties)
     try
     {
         Ice::PropertiesPtr props = _this->getCommunicator()->getProperties();
-        if (!createProperties(return_value, props))
+        if (!IcePHP::createProperties(return_value, props))
```
bernardnormier, line 479: Without the qualification, the new Ice::createProperties template function is found first via ADL.

cpp/src/Ice/Properties.cpp
```diff
@@ -784,7 +825,88 @@ Ice::Properties::parseLine(string_view line, const StringConverterPtr& converter
...
+
+Ice::Properties::Properties(const PropertiesPtr& defaults)
+{
+    if (defaults)
+    {
+        Properties properties{*defaults}; // temporary copy
+        _propertySet = std::move(properties._propertySet);
+        _optInPrefixes = std::move(properties._optInPrefixes);
+    }
+}
```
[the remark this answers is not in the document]
bernardnormier: It's slightly less efficient than the suggested alternative, however:
 - efficiency doesn't matter for this code
 - it's more elegant to not duplicate the copying logic with the other/defaults mutex locked.

cpp/src/Ice/Properties.cpp
```diff
@@ -784,7 +825,88 @@ Ice::Properties::parseLine(string_view line, const StringConverterPtr& converter
...
+    {
+        if (q != args.end())
+        {
+            //
+            // Use the first argument as the value for Ice.ProgramName. Replace
+            // any backslashes in this value with forward slashes, in case this
+            // value is used by the event logger.
+            //
+            string name = *q;
+            replace(name.begin(), name.end(), '\\', '/');
```
pepone, line 857: If the replace is about the event logger it should be Windows only.

pepone (approved):

cpp/src/Ice/Properties.cpp
```diff
@@ -76,96 +76,137 @@ namespace
...
+        }
+        else
+        {
+            ++i;
         }
     }
-    else
+
+    //
+    // Make sure that argv[argc] == 0, the ISO C++ standard requires this.
```
externl:
```suggestion
    // Make sure that argv[argc] == nullptr, the ISO C++ standard requires this.
```

cpp/src/Ice/Properties.cpp
```diff
@@ -76,96 +76,137 @@ namespace
...
-            tmp.push_back(s);
+            ++i;
         }
-        ++q;
     }
-    args = tmp;
 
-    if (!loadConfigFiles)
+    //
+    // Make sure that argv[argc] == 0, the ISO C++ standard requires this.
```
externl:
```suggestion
    // Make sure that argv[argc] == nullptr, the ISO C++ standard requires this.
```

cpp/src/Ice/Properties.cpp
```diff
@@ -76,96 +76,137 @@ namespace
...
+    // We can only do this if we've shifted the array, otherwise argv[argc]
+    // may point to an invalid address.
+    //
+    if (argv && argcOrig != argc)
     {
-        //
-        // If Ice.Config is not set, load from ICE_CONFIG (if set)
-        //
-        loadConfigFiles = (_properties.find("Ice.Config") == _properties.end());
+        argv[argc] = 0;
```
externl:
```suggestion
        argv[argc] = nullptr;
```

cpp/src/Ice/Properties.cpp
externl: What's the issue here with the event logger? Shouldn't this be something that the event logger does instead of being here?

externl (approved):

cpp/src/Ice/Properties.cpp
bernardnormier: I will review/update all ProgramName-related code in a follow-up PR.

PR #4571, opened 2025-10-20
Enable DataStorm property prefix all the time

bernardnormier:
This PR enables the DataStorm property prefix all the time, because it's legitimate for user applications to create their own communicators with DataStorm properties.

### DIFF
--- a/cpp/test/DataStorm/api/Writer.cpp
+++ b/cpp/test/DataStorm/api/Writer.cpp
@@ -31,32 +31,10 @@ void ::Writer::run(int argc, char* argv[])
         }
 
         {
-            // Communicators shared with DataStorm must have a property set that can use the "DataStorm" opt-in prefix.
-            Ice::InitializationData initData;
-            initData.properties = make_shared<Ice::Properties>("DataStorm");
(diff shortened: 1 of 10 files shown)


cpp/test/DataStorm/api/Writer.cpp
```diff
@@ -31,33 +31,11 @@ void ::Writer::run(int argc, char* argv[])
...
-            Ice::InitializationData initData;
-            initData.properties = make_shared<Ice::Properties>("DataStorm");
-            Ice::CommunicatorHolder communicatorHolder{Ice::initialize(initData)};
-            Node n2{communicatorHolder.communicator()};
+            Ice::CommunicatorPtr communicator = Ice::initialize();
+            Ice::CommunicatorHolder communicatorHolder{communicator};
+            Node n2{communicator};
         }
 
-        Ice::InitializationData initData;
```
bernardnormier, line 41: Remove these "n2x" tests as I don't see what they were testing.

pepone (approved):

externl (approved):

PR #4572, opened 2025-10-20
Un-deprecate createProperties in C#, Java, and MATLAB

bernardnormier:
This PR undeprecates createProperties in  C#, Java, and MATLAB, and adds comments like:

```
This method is provided for backwards compatibility. New code should call the {@link Properties#Properties(String[], java.util.List)} constructor directly.
```

It was one of the very newly deprecated functions in 3.8. Furthermore, in many languages (Swift, Python, PHP, Ruby), createProperties is actually the only API to create properties.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
@@ -6,7 +6,6 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.security.Identity;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ThreadFactory;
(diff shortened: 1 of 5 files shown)


matlab/lib/+Ice/createProperties.m
```diff
@@ -15,6 +16,9 @@
     %       string array
     %
     %   See also Ice.Properties/Properties.
+
+    % Copyright (c) ZeroC, Inc.
+
```
externl, line 21: Was this intentional?

externl (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -6,82 +6,74 @@
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.security.Identity;
```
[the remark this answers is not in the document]
pepone: Need to remove this, probably added by VSCode.

pepone (approved):

matlab/lib/+Ice/createProperties.m
InsertCreativityHere: Yep, this is how we write Copyrights in MATLAB.

InsertCreativityHere (approved):

PR #4573, opened 2025-10-20
DiscardPolicy test fixes

pepone:


cpp/test/DataStorm/config/Reader.cpp
```diff
@@ -264,6 +264,8 @@ void ::Reader::run(int argc, char* argv[])
         test(reader.getNextUnread().getValue() == "value2");
         test(reader.getNextUnread().getValue() == "value21");
         test(reader.getNextUnread().getEvent() == DataStorm::SampleEvent::Remove);
+        // All other samples should have been discarded and there should not be any unread samples.
+        test(reader.hasUnread() == false);
```
pepone, line 268: We were only testing that the first 3 samples match the expected value, hasUnread is to ensure the others are not in the unread queue but discarded.

cpp/test/DataStorm/config/Writer.cpp
```diff
@@ -269,14 +269,17 @@ void ::Writer::run(int argc, char* argv[])
         config.priority = 1;
         auto writer2 = makeSingleKeyWriter(topic, "elemdp1", "", config);
 
-        writer1.add("value1");
-        writer1.update("value2");
-        writer1.partialUpdate<string>("concat")("1");
-        writer1.remove();
+        // Wait for the reader to connect to both writers, then send the data.
+        writer1.waitForReaders();
+        writer2.waitForReaders();
```
pepone, line 274: It is clear to wait for readers to connect, even if here everything is serialized over a single connection. A better test would use multiple nodes for the publishers...

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -763,16 +763,18 @@ DataReaderI::initSamples(
...
-            _discardPolicy == DataStorm::DiscardPolicy::Priority &&
-            priority < _connectedKeys[sample->key].back()->priority)
+
+        // Apply discard policies:
+        // - SendTime: discard samples older than the last received sample.
+        // - Priority: discard samples from publisher with lower priority than the highest priority among the connected
+        //   publishers for the same key. The subscriber list is sorted by priority in addConnectedKey.
+        if ((_discardPolicy == DataStorm::DiscardPolicy::SendTime && sample->timestamp <= _lastSendTime) ||
+            (_discardPolicy == DataStorm::DiscardPolicy::Priority &&
+             priority < _connectedKeys[sample->key].back()->priority))
```
externl, line 773: Maybe we should create function for this common logic? `shouldDiscard`?

bernardnormier (approved):

Issue #4565, opened 2025-10-17
Proposal: Set Ice.Program fallback value during Communicator initialization.


externl:
Ice::initialize should give `Ice.ProgramName` a fallback value (in case it's not set by the user).

Currently, if not set by the user we set `Ice.Program` name during the construction of Properties, but only in one of the constructors, and only in some languages.

What is the current behavior:

- **C++:** We set from args argument in the Properties-with-args constructor. Users need to be careful not to pass a random list.
- **C#:**  We get it from `AppDomain.CurrentDomain.FriendlyName`
- **Java:** Nothing. Maybe I just don't see it?
- **Ruby:** We do`callRuby(rb_gv_get, "$0");`(We actually do this in two places: properties construction and communicator initialization. 
- **Python:** Nothing. We rely on C++ args behavior.
- **MATLAB:** We set to `"matlab-client"`
- **PHP:**  Nothing. We rely on C++ args behavior. this only works if calling php script.php, if you're calling from something like php-fpm this will not bet set I think.
- **JS:** Nothing.
- **Swift:** Nothing. We rely on C++ args behavior.

Proposal
 - Stop setting Ice.Program name in Properties-with-args in C++.
 - In general, Properties creation doesn't set a default ProgramName.
 - In all language, Ice::initialize (Communicator constructor) gives Ice.ProgramName a fallback value.
 - C++ Ice::initialize(argc, argv) would set Ice.Program name if not set in properties and argc is not empty.
 - All C++ wrapped languages should set Ice.ProgrameName before calling C++ Ice::initialize.

PR #4574, opened 2025-10-20
Set Ice.ProgramName in initialize

bernardnormier:
This PR updates the C++ and C# code to set Ice.ProgramName in initialize instead of the Properties code itself.

This is a partial fix for #4565.

### DIFF
--- a/cpp/include/Ice/Initialize.h
+++ b/cpp/include/Ice/Initialize.h
@@ -18,6 +18,8 @@
 #include "StringUtil.h"
 #include "TimerTask.h"
 
+#include <algorithm>
+
 namespace Ice
 {
     /// Represents a set of options that you can specify when initializing a communicator.
@@ -94,8 +96,19 @@ namespace Ice
     /// @return The new communicator.
     template<typename ArgvT> inline CommunicatorPtr initialize(int& argc, ArgvT argv)
     {
+        auto properties = std::make_shared<Properties>(argc, argv);
+        if (properties->getProperty("Ice.ProgramName").empty() && argc > 0)
+        {
+            StringSeq args = argsToStringSeq(argc, argv);
+            std::string programName = args[0];
+            // Replace any backslashes in this value with forward slashes, in case this value is used by the event
+            // logger.
+            std::replace(programName.begin(), programName.end(), '\\', '/');
+            properties->setProperty("Ice.ProgramName", std::move(programName));
+        }
+
         InitializationData initData;
-        initData.properties = std::make_shared<Properties>(argc, argv);
+        initData.properties = properties;
         return initialize(std::move(initData));
     }
 
--- a/cpp/src/Ice/Instance.cpp
+++ b/cpp/src/Ice/Instance.cpp
@@ -928,6 +928,12 @@ IceInternal::Instance::initialize(const Ice::CommunicatorPtr& communicator)
 {
     try
     {
+        string programName;
+        if (_initData.properties)
+        {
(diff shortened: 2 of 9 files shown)


cpp/include/Ice/Initialize.h
```diff
@@ -94,8 +96,19 @@ namespace Ice
     /// @return The new communicator.
     template<typename ArgvT> inline CommunicatorPtr initialize(int& argc, ArgvT argv)
     {
+        auto properties = std::make_shared<Properties>(argc, argv);
+        if (properties->getProperty("Ice.ProgramName").empty() && argc > 0)
+        {
+            StringSeq args = argsToStringSeq(argc, argv);
+            std::string programName = args[0];
+            // Replace any backslashes in this value with forward slashes, in case this value is used by the event
```
bernardnormier, line 104: I kept the replace logic here, as it's the most backwards compatible.

cpp/src/Ice/Instance.cpp
```diff
@@ -928,6 +928,12 @@ IceInternal::Instance::initialize(const Ice::CommunicatorPtr& communicator)
 {
     try
     {
+        string programName;
+        if (_initData.properties)
+        {
+            programName = _initData.properties->getIceProperty("Ice.ProgramName");
```
bernardnormier, line 934: It's cleaner to read this property once in this function instead of multiple times. And reading it all the time has the advantage of marking this property as "used".

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
```diff
@@ -564,6 +564,9 @@ void initialize(Communicator communicator, InitializationData initData) {
 
             Properties properties = _initData.properties;
 
+            // TODO: set a fallback value like in C#.
```
bernardnormier, line 567: Currently, there is no fallback or default in Java.

csharp/test/Ice/admin/AllTests.cs
```diff
@@ -211,9 +211,10 @@ public static void allTests(global::Test.TestHelper helper)
             // Test: PropertiesAdmin::getProperties()
             //
             Dictionary<string, string> pd = pa.getPropertiesForPrefix("");
-            test(pd.Count == 5);
+            test(pd.Count == 6);
             test(pd["Ice.Admin.Endpoints"] == "tcp -h 127.0.0.1");
             test(pd["Ice.Admin.InstanceName"] == "Test");
+            test(pd["Ice.ProgramName"] == "server");
```
bernardnormier, line 217: There was a small bug in C#: `Ice.ProgramName` was not always set.

csharp/src/Ice/Internal/Instance.cs
```diff
@@ -580,7 +580,15 @@ internal void initialize(Ice.Communicator communicator, Ice.InitializationData i
 
         try
         {
-            _initData.properties ??= new Ice.Properties();
+            _initData.properties ??= new Properties();
```
pepone, line 583: In C++ version this assignment is done with the mutex locked, should we lock the mutex here too?

pepone (approved):

externl (approved):

csharp/src/Ice/Internal/Instance.cs
bernardnormier: It doesn't matter. This initData is not supposed to be shared by multiple communicators, and initData itself is not thread-safe.

PR #4577, opened 2025-10-20
Don't set connection properties in colloc tests

bernardnormier:
This PR removes the setting of connection properties in various colloc tests. These properties have no effect on colloc.


swift/test/Ice/ami/Collocated.swift
```diff
@@ -6,24 +6,6 @@ import TestCommon
 
 class Collocated: TestHelperI, @unchecked Sendable {
     override public func run(args: [String]) async throws {
-        let properties = try createTestProperties(args)
```
bernardnormier, line 9: properties is not used below - dead code.

pepone (approved):

PR #4579, opened 2025-10-21
Update ObjectAdapter thread pool creation condition

pepone:
This PR updates the creation of the per-OA thread pool to ensure that if any OA thread pool property is set, we create the per-OA thread pool. Previously the code only checked `ThreadPool.Size` and `ThreadPool.SizeMax` properties.

### DIFF
--- a/cpp/src/Ice/ObjectAdapterI.cpp
+++ b/cpp/src/Ice/ObjectAdapterI.cpp
@@ -929,13 +929,7 @@ Ice::ObjectAdapterI::initialize(optional<RouterPrx> router)
         // Otherwise the OA will use the default server thread pool.
         // This is done before the creation of the incoming connection factory as the thread pool is needed during
         // creation for the call to incFdsInUse.
-        bool createThreadPool = !properties->getProperty(_name + ".ThreadPool.Serialize").empty() ||
-                                !properties->getProperty(_name + ".ThreadPool.Size").empty() ||
-                                !properties->getProperty(_name + ".ThreadPool.SizeMax").empty() ||
-                                !properties->getProperty(_name + ".ThreadPool.SizeWarn").empty() ||
-                                !properties->getProperty(_name + ".ThreadPool.ThreadIdleTime").empty();
-
-        if (createThreadPool)
+        if (properties->getPropertiesForPrefix(_name + ".ThreadPool.").size() > 0)
         {
             _threadPool = ThreadPool::create(_instance, _name + ".ThreadPool", 0);
         }
--- a/csharp/src/Ice/ObjectAdapter.cs
+++ b/csharp/src/Ice/ObjectAdapter.cs
@@ -1088,15 +1088,7 @@ public sealed class ObjectAdapter
         {
             // If the user configured any of the ObjectAdapter.ThreadPool properties, create a per-adapter thread pool.
             // Otherwise the OA will use the default server thread pool.
-            bool createThreadPool =
-                properties.getProperty(_name + ".ThreadPool.Serialize").Length > 0 ||
-                properties.getProperty(_name + ".ThreadPool.Size").Length > 0 ||
-                properties.getProperty(_name + ".ThreadPool.SizeMax").Length > 0 ||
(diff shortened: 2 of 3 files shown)


cpp/src/Ice/ObjectAdapterI.cpp
```diff
@@ -925,14 +925,17 @@ Ice::ObjectAdapterI::initialize(optional<RouterPrx> router)
...
-        if (threadPoolSize > 0 || threadPoolSizeMax > 0)
+        // If the user configured any of the ObjectAdapter.ThreadPool properties, create a per-adapter thread pool.
+        // Otherwise the OA will use the default server thread pool.
+        // This is done before the creation of the incoming connection factory as the thread pool is needed during
+        // creation for the call to incFdsInUse.
+        bool createThreadPool = !properties->getProperty(_name + ".ThreadPool.Serialize").empty() ||
+                                !properties->getProperty(_name + ".ThreadPool.Size").empty() ||
+                                !properties->getProperty(_name + ".ThreadPool.SizeMax").empty() ||
+                                !properties->getProperty(_name + ".ThreadPool.SizeWarn").empty() ||
+                                !properties->getProperty(_name + ".ThreadPool.ThreadIdleTime").empty();
```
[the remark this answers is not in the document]
pepone: These properties are not supported in C++

csharp/src/Ice/ObjectAdapter.cs
```diff
@@ -1086,9 +1086,17 @@ internal ObjectAdapter(
 
         try
         {
-            int threadPoolSize = properties.getPropertyAsInt(_name + ".ThreadPool.Size");
-            int threadPoolSizeMax = properties.getPropertyAsInt(_name + ".ThreadPool.SizeMax");
-            if (threadPoolSize > 0 || threadPoolSizeMax > 0)
+            // If the user configured any of the ObjectAdapter.ThreadPool properties, create a per-adapter thread pool.
+            // Otherwise the OA will use the default server thread pool.
+            bool createThreadPool =
```
bernardnormier: I don't like this check:
 - we should always check Ice properties with getIceProperty to make sure there is no typo
 - this long list makes it easy to miss a property or forget to add/remove a property if we ever change the ThreadPool properties

What about calling instead `properties.getPropertiesForPrefix`, and then create the ThreadPool if any property has this prefix?

csharp/src/Ice/ObjectAdapter.cs
```diff
@@ -1086,9 +1086,17 @@ internal ObjectAdapter(
...
+                properties.getProperty(_name + ".ThreadPool.Serialize").Length > 0 ||
+                properties.getProperty(_name + ".ThreadPool.Size").Length > 0 ||
+                properties.getProperty(_name + ".ThreadPool.SizeMax").Length > 0 ||
+                properties.getProperty(_name + ".ThreadPool.SizeWarn").Length > 0 ||
+                properties.getProperty(_name + ".ThreadPool.StackSize").Length > 0 ||
+                properties.getProperty(_name + ".ThreadPool.ThreadIdleTime").Length > 0 ||
+                properties.getProperty(_name + ".ThreadPool.ThreadPriority").Length > 0;
+            if (createThreadPool)
             {
                 _threadPool = new Ice.Internal.ThreadPool(_instance, _name + ".ThreadPool", 0);
```
bernardnormier, line 1093: Not added here - I am not a fan of this trailing `0` for the server idle timeout.

pepone: Not clear what you mean

csharp/src/Ice/ObjectAdapter.cs
pepone: Fixed as suggested

csharp/src/Ice/ObjectAdapter.cs
bernardnormier: I am talking about the `int` trailing parameter in the ThreadPool constructor.

PR #4582, opened 2025-10-21
Update tests to use one log file per communicator

bernardnormier:
This PR updates the idleTimeout and inactivityTimeout tests to use a log file for each communicator.

### DIFF
--- a/cpp/test/Common/TestHelper.cpp
+++ b/cpp/test/Common/TestHelper.cpp
@@ -255,6 +255,18 @@ Test::TestHelper::getTestPort(const Ice::PropertiesPtr& properties, int num)
     return properties->getPropertyAsIntWithDefault("Test.BasePort", 12010) + num;
 }
 
+void
+Test::TestHelper::updateLogFileProperty(const Ice::PropertiesPtr& properties, const string& suffix)
+{
+    string logFile = properties->getIceProperty("Ice.LogFile");
+    if (!logFile.empty())
+    {
+        string newLogFile = logFile.substr(0, logFile.find_last_of('.')) + suffix +
+            logFile.substr(logFile.find_last_of('.'));
+        properties->setProperty("Ice.LogFile", newLogFile);
+    }
+}
+
 Ice::PropertiesPtr
 Test::TestHelper::createTestProperties(int& argc, char* argv[], const Ice::PropertiesPtr& defaults)
 {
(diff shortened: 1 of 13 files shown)


csharp/test/Ice/idleTimeout/AllTests.cs
```diff
@@ -96,7 +97,8 @@ private static async Task testEnableDisableIdleCheck(
         properties.setProperty("Ice.Connection.Client.IdleTimeout", "1");
         properties.setProperty("Ice.Connection.Client.EnableIdleCheck", enabled ? "1" : "0");
         properties.setProperty("Ice.Warn.Connections", "0");
-        Communicator communicator = Util.initialize(new InitializationData { properties = properties });
+        global::Test.TestHelper.updateLogFileProperty(properties, "-idleTimeout=1s");
+        await using Communicator communicator = Util.initialize(new InitializationData { properties = properties });
```
bernardnormier, line 101: The `await using` was missing too.

js/test/Ice/idleTimeout/Client.ts
```diff
@@ -31,24 +31,21 @@ async function testConnectionNotAbortedByIdleCheck(properties: Ice.Properties, h
...
     properties.setProperty("Ice.Connection.Client.EnableIdleCheck", "1");
     properties.setProperty("Ice.Warn.Connections", "0");
+    helper.updateLogFileProperty(properties, "-idleTimeout=1s");
     const initData = new Ice.InitializationData();
     initData.properties = properties;
-    const communicator = Ice.initialize(initData);
-    try {
-        const adapter = await communicator.createObjectAdapter("");
-        const callback = Test.TestIntfPrx.uncheckedCast(adapter.add(new TestI(), Ice.stringToIdentity("test")));
+    await using communicator = Ice.initialize(initData);
```
bernardnormier, line 37: Not clear why we didn't use await using for these tests.

externl (approved):

cpp/test/Common/TestHelper.cpp
```diff
@@ -255,6 +255,26 @@ Test::TestHelper::getTestPort(const Ice::PropertiesPtr& properties, int num)
...
 
+void
+Test::TestHelper::updateLogFileProperty(const Ice::PropertiesPtr& properties, const string& suffix)
+{
+    string logFile = properties->getIceProperty("Ice.LogFile");
+    if (!logFile.empty())
+    {
+        size_t pos = logFile.find_last_of('.');
+        string newLogFile;
+        if (pos != string::npos)
```
pepone: niptick: I think is a bit clear to do like in Java `pos == string::npos` and reverse the if/else blocks.

pepone (approved):

PR #4584, opened 2025-10-22
Remove Java finalizers

### LINKED ISSUE #4581: Remove Java finalizers
> Java finalizers, specifically the Object.finalize() method, have been deprecated since Java 9. Furthermore, in Java 18, the method was marked for removal, indicating its eventual elimination in a future Java release.

These plus the Assert.FinalizerAssert should be removed.

pepone:
Fix #4581


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectAdapter.java
```diff
@@ -1262,32 +1262,6 @@ public SSLEngineFactory getSSLEngineFactory() {
         return _sslEngineFactory;
     }
```
InsertCreativityHere, line 1256: Higher up in this file, we have 3 of these:
```
            // Prevent finalizer from complaining about the adapter not being destroyed.
            _state = StateDestroyed;
```
Can we delete these now that `finalize` is gone?

bernardnormier (approved):

PR #4585, opened 2025-10-22
Fix android controller proxy

externl:
This PR fixes the `RemoteProcessController` to work correctly when using a direct proxy. It worked previously since there was a bug in the `AndroidProcessController` that caused the controller to mistakenly indicate that it used IceDiscovery.

### DIFF
--- a/scripts/Util.py
+++ b/scripts/Util.py
@@ -2327,6 +2327,9 @@ class RemoteProcessController(ProcessController):
     def __str__(self):
         return "remote controller"
 
+    def setup(self, current):
+        pass
+
     def getHost(self, current):
         return self.getController(current).getHost(current.config.protocol, current.config.ipv6)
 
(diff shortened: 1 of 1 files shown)


scripts/Util.py
```diff
@@ -2540,7 +2542,6 @@ def startEmulator(self, avd):
         if port == -1:
             raise RuntimeError("cannot find free port in range 5554-5584, to run android emulator")
 
-        self.device = "emulator-{}".format(port)
```
externl, line 2543: This was the android bug

scripts/Util.py
```diff
@@ -2395,7 +2394,9 @@ def callback(future):
         while nRetry < 120:
             nRetry += 1
 
-            if self.supportsDiscovery():
+            # If the process controller supports discovery or if we have a direct endpoint to it, try to
+            # ping it. This can take a few tries as the controller app might still be starting.
+            if self.supportsDiscovery() or controllerEndpoints is not None:
                 proxy.ice_pingAsync().add_done_callback(callback)
```
pepone, line 2406: Should we just remove this `if`, in what case are we not entering this branch and just waiting?

externl: Here it would be waiting for a controller to register itself. Lets keep until we fix https://github.com/zeroc-ice/ice/issues/4587

pepone (approved):
Looks good!

PR #4588, opened 2025-10-22
Make Logger disposable in C#

bernardnormier:
This PR makes the Logger interface disposable in C#, and clarifies who is responsible to dispose a logger provided through initData.


csharp/src/Ice/LoggerPlugin.cs
```diff
@@ -1,54 +1,36 @@
...
+/// A special plug-in that installs a logger during a communicator's initialization.
+/// Both initialize and destroy are no-op. See Ice::InitializationData.
 /// </summary>
 public class LoggerPlugin : Plugin
 {
     /// <summary>
-    /// Initializes a new instance of the <see cref="LoggerPlugin" /> class. This constructor installs a custom
-    /// logger to the given communicator.
+    /// Initializes a new instance of the <see cref="LoggerPlugin" /> class. This constructor installs a custom logger
+    /// in <paramref name="communicator" />. The communicator takes ownership of <paramref name="logger" /> and is
```
bernardnormier, line 15: This ownership transfer seems the most practical.

csharp/src/Ice/LoggerI.cs
```diff
@@ -195,6 +200,14 @@ public override void error(string message)
 
     public override Logger cloneWithPrefix(string prefix) => new TraceLoggerI(prefix, _console);
 
+    public override void Dispose()
+    {
+        if (_console && Trace.Listeners.Contains(_consoleListener))
+        {
+            Trace.Listeners.Remove(_consoleListener);
```
bernardnormier, line 207: This is new behavior, which seems fine.

csharp/test/Ice/exceptions/Server.cs
```diff
@@ -4,29 +4,6 @@
 
 namespace Ice.exceptions;
 
-public sealed class DummyLogger : Ice.Logger
```
bernardnormier, line 7: This DummyLogger wasn't used anywhere.

csharp/src/iceboxnet/ServiceManagerI.cs
```diff
@@ -539,15 +539,6 @@ private void startService(string service, string entryPoint, string[] args)
                     info.args = initData.properties.parseCommandLineOptions(service, info.args);
                 }
 
-                //
-                // Clone the logger to assign a new prefix. If one of the built-in loggers is configured
-                // don't set any logger.
-                //
-                if (initData.properties.getIceProperty("Ice.LogFile").Length == 0)
```
bernardnormier, line 546: Removed this special case.

csharp/src/Ice/Internal/LoggerAdminLoggerI.cs
```diff
@@ -8,46 +8,69 @@ internal interface LoggerAdminLogger : Logger
...
 }
 
+// Decorates a logger.
 internal sealed class LoggerAdminLoggerI : LoggerAdminLogger
 {
+    private static readonly DateTime _unixEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+    private readonly Logger _localLogger;
+    private readonly LoggerAdminI _loggerAdmin;
+    private Thread _sendLogThread;
+    private volatile bool _detached;
```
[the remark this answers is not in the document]
bernardnormier: What about lines 30, 40, 50, 60?

csharp/src/Ice/Logger.cs
```diff
@@ -4,36 +4,35 @@
 
 namespace Ice;
 
-public interface Logger
+/// <summary>Represents Ice's abstraction for logging and tracing. Applications can provide their own logger by
+/// implementing this abstraction and setting a logger on the communicator.</summary>
+public interface Logger : IDisposable
```
pepone, line 9: Is this a good idea? Can we instead check if the implementation is IDisposable and owned.

Doesn't seems ideal that all loggers need to be disposable.

bernardnormier: I believe that's the correct design in C#. 

If the caller/owner must call Dispose, the interface should be disposable.

Having code like:
```csharp
if (logger is IDisposable disposable)
{
    disposable.Dispose();
}
```

is completely unmaintainable.

In C++, the equivalent is shared pointers. The sharing is part of the API: we don't discover at runtime if a configured logger is shared. 

In IceRPC, IProtocolConnection is IAsyncDisposable: 
https://github.com/icerpc/icerpc-csharp/blob/2c83083b3bf7d65aa95c8315ca8c2a994d22cb1d/src/IceRpc/IProtocolConnection.cs#L17

It's part of the interface: the caller doesn't cast the protocol connection to figure this out.

Same for https://github.com/icerpc/icerpc-csharp/blob/2c83083b3bf7d65aa95c8315ca8c2a994d22cb1d/src/IceRpc/Transports/IDuplexConnection.cs#L29 and more.

externl: I guess https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.iloggerfactory?view=net-9.0-pp is probably more apt comparison.

csharp/src/Ice/Internal/Instance.cs
```diff
@@ -551,12 +551,16 @@ public void
...
     }
 
-    public void
-    setLogger(Ice.Logger logger) =>
-        //
-        // No locking, as it can only be called during plug-in loading
-        //
+    internal void setLogger(Logger logger)
+    {
+        // No locking, as it can only be called during plug-in loading.
```
pepone: I would mention the LoggerPlugin in the comment:


```suggestion
        // Used by LoggerPlugin only, no locking, as it can only be called during plug-in loading.
```

csharp/src/Ice/Internal/Instance.cs
```diff
@@ -910,7 +918,8 @@ internal void finishSetup(Communicator communicator)
             if (_adminFacetFilter.Count == 0 || _adminFacetFilter.Contains(loggerFacetName))
             {
                 LoggerAdminLogger logger = new LoggerAdminLoggerI(_initData.properties, _initData.logger);
-                setLogger(logger);
+                // Decorate the existing logger, and don't change _ownLogger.
+                _initData.logger = logger;
```
pepone, line 922: So if we own the local loger it will be disposed when we dispose LoggerAdminLogger?

csharp/src/Ice/Internal/LoggerAdminLoggerI.cs
```diff
@@ -8,46 +8,69 @@ internal interface LoggerAdminLogger : Logger
...
+        {
+            var logMessage = new LogMessage(LogMessageType.ErrorMessage, now(), "", message);
+            log(logMessage);
+        }
     }
 
     public string getPrefix() => _localLogger.getPrefix();
 
-    public Ice.Logger cloneWithPrefix(string prefix) => _localLogger.cloneWithPrefix(prefix);
+    public Logger cloneWithPrefix(string prefix) => _localLogger.cloneWithPrefix(prefix);
```
pepone, line 69: Not change in this PR but this implementation doesn't make much sense to me, you clone LoggerAdminLoggerI and you get but the underlying local logger?

csharp/src/Ice/Internal/LoggerAdminLoggerI.cs
```diff
@@ -129,12 +153,12 @@ private void run()
...
-                while (!_destroyed && _jobQueue.Count == 0)
+                while (!_detached && _jobQueue.Count == 0)
                 {
                     Monitor.Wait(_mutex);
                 }
 
-                if (_destroyed)
+                if (_detached)
                 {
                     break; // for(;;)
```
pepone: Not related to the PR:

```suggestion
                    break; // while(true)
```

csharp/src/Ice/Util.cs
```diff
@@ -277,7 +277,11 @@ public static Logger getProcessLogger()
     {
         lock (_processLoggerMutex)
         {
-            _processLogger ??= new ConsoleLoggerI(AppDomain.CurrentDomain.FriendlyName);
+            if (_processLogger is null)
+            {
+                _processLogger = new ConsoleLoggerI(AppDomain.CurrentDomain.FriendlyName);
```
pepone, line 282: If setProcessLogger is not called this is never disposed, I guess there isn't a good place to dispose of statics like this.

pepone (approved):

csharp/src/Ice/Internal/Instance.cs
bernardnormier: Yes, the LoggerAdminLogger is a decorator that doesn't change ownership.

csharp/src/Ice/Internal/LoggerAdminLoggerI.cs
externl: I would just remove this comment. It's just a break statement.

externl (approved):

csharp/src/Ice/Internal/Instance.cs
bernardnormier: Fixed

csharp/src/Ice/Internal/LoggerAdminLoggerI.cs
bernardnormier: We return a clone of the decoratee, which can make sense. 

The real question is what's the use-case for cloneWithPrefix?

csharp/src/Ice/Internal/LoggerAdminLoggerI.cs
bernardnormier: Fixed.

csharp/src/Ice/Util.cs
bernardnormier: Yes, static fields are not disposed.

PR #4589, opened 2025-10-22
Enable CA2000 warning for C# tests

pepone:


csharp/msbuild/CodeAnalysis.Tests.globalconfig
```diff
@@ -138,9 +138,6 @@ dotnet_diagnostic.CA1840.severity = none
 # CA1854: Prefer a 'TryGetValue' call
 dotnet_diagnostic.CA1854.severity = none
 
-# Call System.IDisposable.Dispose on object
```
bernardnormier, line 141: (For a separate PR). It would be good to reduce the number of rule-overriding in this file.

PR #4594, opened 2025-10-23
Update logger in C++

bernardnormier:
This PR update the C++ logger, LoggerAdmin, and IceBox to be consistent with the C# version. See #4588.

### DIFF
--- a/cpp/src/Ice/LoggerAdminI.cpp
+++ b/cpp/src/Ice/LoggerAdminI.cpp
@@ -137,6 +137,23 @@ namespace
     // Helper functions
     //
 
+    LoggerPtr unwrapLocalLogger(const LoggerPtr& localLogger)
+    {
+        // There is currently no way to have a null local logger
+        assert(localLogger);
+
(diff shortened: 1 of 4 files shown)


cpp/src/Ice/LoggerAdminI.cpp
```diff
@@ -136,6 +137,23 @@ namespace
     // Helper functions
     //
 
+    LoggerPtr unwrapLocalLogger(const LoggerPtr& localLogger)
+    {
+        // There is currently no way to have a null local logger
+        assert(localLogger);
+
+        auto wrapper = dynamic_pointer_cast<LoggerAdminLoggerI>(localLogger);
```
pepone, line 145: Do we ever pass a `LoggerAdminLoggerI` logger to this method?

pepone (approved):

externl (approved):

cpp/src/Ice/LoggerAdminI.cpp
bernardnormier: I doubt we do that in the tests. However, it's in theory possible since Communicator::getLogger returns the wrapped logger.

PR #4595, opened 2025-10-23
Upgrade Open Rewrite version

pepone:
This PR upgrades the openrewrite version to the latest stable.
It updates the CI java workflow to pass `--no-parallel` to avoid warnings with Gradle 9.0 See https://github.com/openrewrite/rewrite/issues/6175


InsertCreativityHere:
Looks good!

I also saw the `no-parallel`, but was hoping to find some better solution. I couldn't find one though :/

I can fix the rewrite issues after you merge if you want. Going to open a follow-up PR to change rewrite to actually print the errors to the console, instead of just giving you the filename and nothing else...

PR #4598, opened 2025-10-23
Set destination directory for source and doc JARs

pepone:
With the previous change to use `withSourcesJar()` and `withJavadocJar()` the JARs where created in the default output directory, instead of the expected lib. This make the last maven publish to fail.

### DIFF
--- a/java/gradle/library.gradle
+++ b/java/gradle/library.gradle
@@ -4,14 +4,15 @@ jar {
     destinationDirectory = new File("${libDir}")
 }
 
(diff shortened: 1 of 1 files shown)


java/gradle/library.gradle
```diff
@@ -4,14 +4,15 @@ jar {
     destinationDirectory = new File("${libDir}")
 }
 
-task jarSources(type:Jar, dependsOn: jar){
```
pepone, line 7: That was duplicate task, we were creating two source JARs in different directories.

PR #4600, opened 2025-10-23
Make Logger Autocloseable in Java

bernardnormier:
This PR updates the Logger interface to be Autocloseable + other updates to bring Java in sync with C#.

### DIFF
--- a/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Logger.java
+++ b/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Logger.java
@@ -60,4 +60,9 @@ class Logger implements com.zeroc.Ice.Logger {
         throw new UnsupportedOperationException(
             "cloneWithPrefix is not supported on IceGridGUI.Logger");
     }
+
+    @Override
+    public void close() throws Exception {
+        // No resources to close in this logger.
+    }
 }
--- a/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Logger.java
+++ b/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Logger.java
@@ -62,7 +62,7 @@ class Logger implements com.zeroc.Ice.Logger {
     }
 
     @Override
-    public void close() throws Exception {
+    public void close() {
         // No resources to close in this logger.
     }
 }
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
@@ -503,9 +503,17 @@ public final class Instance {
         _referenceFactory = _referenceFactory.setDefaultRouter(router);
     }
 
-    public void setLogger(Logger logger) {
-        // No locking, as it can only be called during plug-in loading
+    void setLogger(Logger logger) {
+        // Only called by the LoggerPlugin constructor, so no need to lock.
+        if (_ownLogger) {
+            try {
+                _initData.logger.close();
+            } catch (Exception ex) {
+                // Ignore exceptions on close
+            }
+        }
         _initData.logger = logger;
+        _ownLogger = true;
     }
 
     public void setThreadHooks(Runnable threadStart, Runnable threadStop) {
@@ -836,7 +844,8 @@ public final class Instance {
             String loggerFacetName = "Logger";
             if (_adminFacetFilter.isEmpty() || _adminFacetFilter.contains(loggerFacetName)) {
                 LoggerAdminLogger logger = new LoggerAdminLoggerI(properties, _initData.logger);
-                setLogger(logger);
+                // Decorate the existing logger, and don't change _ownLogger.
+                _initData.logger = logger;
                 _adminFacets.put(loggerFacetName, logger.getFacet());
             }
 
@@ -970,8 +979,9 @@ public final class Instance {
             }
 
             if (_initData.logger instanceof LoggerAdminLogger) {
-                // This only disables the remote logging; we don't set or reset _initData.logger
-                ((LoggerAdminLogger) _initData.logger).destroy();
+                // This only disables the remote logging; we don't set or reset _initData.logger, and we keep logging
+                // to the underlying logger.
+                ((LoggerAdminLogger) _initData.logger).detach();
             }
 
             // Now, destroy the thread pools. This must be done *only* after all the connections are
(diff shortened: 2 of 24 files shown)


java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Logger.java
```diff
@@ -60,4 +60,9 @@ public Logger cloneWithPrefix(String prefix) {
         throw new UnsupportedOperationException(
             "cloneWithPrefix is not supported on IceGridGUI.Logger");
     }
+
+    @Override
+    public void close() throws Exception {
```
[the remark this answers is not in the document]
InsertCreativityHere: Copilot's suggestion is right, although it's reasoning is a little odd.
Java supports 'covariant' exception specifications.
i.e. when you override a method, you can `throws` a subset of the exceptions.
This includes not throwing anything.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
```diff
@@ -503,9 +503,17 @@ public synchronized void setDefaultRouter(RouterPrx router) {
...
+        // Only called by the LoggerPlugin constructor, so no need to lock.
+        if (_ownLogger) {
+            try {
+                _initData.logger.close();
+            } catch (Exception ex) {
+                // Ignore exceptions on close
+            }
+        }
         _initData.logger = logger;
+        _ownLogger = true;
```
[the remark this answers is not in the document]
bernardnormier: Here, we want to take ownership.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerI.java
```diff
@@ -95,6 +95,11 @@ public Logger cloneWithPrefix(String prefix) {
         return new LoggerI(prefix);
     }
 
+    @Override
+    public void close() throws Exception {
```
InsertCreativityHere, line 99: No need to have the `throws` just because the parent method did:
```suggestion
    public void close() {
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -316,8 +318,17 @@ public static Logger getProcessLogger() {
...
      */
     public static void setProcessLogger(Logger logger) {
+        Objects.requireNonNull(logger);
         synchronized (_processLoggerMutex) {
+            if (_ownProcessLogger) {
+                try {
+                    _processLogger.close();
+                } catch (Exception e) {
+                    // Ignore exceptions on close
+                }
```
InsertCreativityHere: A general convention in Java would be to do this:
```suggestion
                } catch (Exception ignored) {
                }
```
Many IDEs and even our linter are aware of this convention.
But, we have this lint turned off, since we never follow it...
Hopefully one day I can re-enable.

java/test/src/main/java/test/Ice/binding/Server.java
```diff
@@ -36,7 +36,12 @@ public String getPrefix() {
 
                 @Override
                 public Logger cloneWithPrefix(String prefix) {
-                    return this;
+                    return this; // not a correct implementation
+                }
+
+                @Override
+                public void close() throws Exception {
```
InsertCreativityHere:
```suggestion
                public void close() {
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
```diff
@@ -836,7 +848,8 @@ void finishSetup(Communicator communicator) {
             String loggerFacetName = "Logger";
             if (_adminFacetFilter.isEmpty() || _adminFacetFilter.contains(loggerFacetName)) {
                 LoggerAdminLogger logger = new LoggerAdminLoggerI(properties, _initData.logger);
-                setLogger(logger);
+                // Decorate the existing logger, and don't change _ownLogger.
```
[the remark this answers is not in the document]
bernardnormier: Please read the comment.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instance.java
```diff
@@ -503,9 +503,17 @@ public synchronized void setDefaultRouter(RouterPrx router) {
...
 
-    public void setLogger(Logger logger) {
-        // No locking, as it can only be called during plug-in loading
+    void setLogger(Logger logger) {
+        // Only called by the LoggerPlugin constructor, so no need to lock.
+        if (_ownLogger) {
+            try {
+                _initData.logger.close();
+            } catch (Exception ex) {
+                // Ignore exceptions on close
```
InsertCreativityHere:
```suggestion
            } catch (Exception ignored) {
```

InsertCreativityHere (approved):
Looks good to me!

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerI.java
bernardnormier: This was incorrect. We need the exception specification because it can be overridden.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerAdminLoggerI.java
```diff
@@ -86,16 +99,16 @@ public void run() {
         for (; ; ) {
```
pepone: update to`while(true)`

pepone (approved):

externl (approved):
Looks good!

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerI.java
InsertCreativityHere: I missed that we were deriving from this class. My wrong.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
bernardnormier: Fixed as suggested

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerAdminLoggerI.java
bernardnormier: done

PR #4601, opened 2025-10-23
Remove unused code from TestHelper in C#

bernardnormier:


csharp/test/TestCommon/TestHelper.cs
```diff
@@ -169,13 +141,11 @@ internal static void fail(string message, string detailMessage)
         Environment.Exit(1);
     }
 
-    public void setControllerHelper(ControllerHelper controllerHelper) => _controllerHelper = controllerHelper;
-
-    public void serverReady() => _controllerHelper?.serverReady();
+    public void serverReady()
+    {
+    }
```
InsertCreativityHere, line 146: Is there a reason to keep this no-op function?
It's true we call it in many places, but it's no-op.

InsertCreativityHere (approved):

csharp/test/TestCommon/TestHelper.cs
pepone: This was required to run C# tests in the Android emulator but we are no longer doing that.

pepone (approved):

externl (approved):

csharp/test/TestCommon/TestHelper.cs
externl: It would be a gigantic pain to add them back if we remove them. I think it would be nice to eventually stop relying on  `Ice.PrintAdapterReady` when running with the local driver and use serverReady like we do for the remote controller.

Issue #4590, opened 2025-10-22
Prefer Communicator constructor over initialize


bernardnormier:
Like for createProperties vs the Properties constructor, we should offer and prefer Communicator construction over Ice.Util.initialize in some languages, starting with:

-  C# 
The syntax is more natural in C# and this enables the CA2000 code analyzer check (make sure you dispose your communicator).

- Java
Similar to C#, minus the code analysis.

- MATLAB (already done)

PR #4602, opened 2025-10-23
Prefer Communicator constructor over Util.initialize in Java

bernardnormier:
This PR is the Java portion of #4590.

Note that I didn't add more try-with-resources in the tests: we still rely on destroy instead of try-with-resources or close in many tests.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
@@ -48,7 +48,7 @@ public final class Communicator implements AutoCloseable {
     }
 
     /**
-     * Constructs a communicator with defaults options.
+     * Constructs a communicator with default options.
      *
      */
     public Communicator() {
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
@@ -48,7 +48,6 @@ public final class Communicator implements AutoCloseable {
 
     /**
      * Constructs a communicator with default options.
-     *
      */
     public Communicator() {
         this(new InitializationData());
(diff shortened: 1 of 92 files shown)


java/test/src/main/java/test/Glacier2/router/Client.java
```diff
@@ -19,7 +19,6 @@
 import com.zeroc.Ice.ProcessPrx;
 import com.zeroc.Ice.RouterFinderPrx;
 import com.zeroc.Ice.SocketException;
-import com.zeroc.Ice.Util;
```
bernardnormier, line 22: Also replaced many calls to Util.stringToIdentity by new Identity(...) and removed unused Util imports.

pepone (approved):

java/test/src/main/java/test/Ice/echo/BlobjectI.java
```diff
@@ -4,6 +4,7 @@
 
 import com.zeroc.Ice.BlobjectAsync;
 import com.zeroc.Ice.Current;
+import com.zeroc.Ice.Identity;
```
externl: Is this import necessary?

externl (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -29,8 +30,54 @@ public final class Communicator implements AutoCloseable {
     private final Instance _instance;
 
     /**
-     * Destroy the communicator.
-     * This Java-only method overrides close in java.lang.AutoCloseable and does not throw any exception.
+     * Constructs a communicator with the specified options.
+     *
+     * @param initData the options for the new communicator
+     *
+     */
```
InsertCreativityHere, line 36: Extra blank line here:
```suggestion
     * @param initData the options for the new communicator
     */
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -29,8 +30,54 @@ public final class Communicator implements AutoCloseable {
...
+        } catch (RuntimeException ex) {
+            _instance.destroy(false);
+            throw ex;
+        }
+    }
+
+    /**
+     * Constructs a communicator with default options.
+     *
+     */
```
InsertCreativityHere, line 51: And here too, no need for trailing blank lines:
```suggestion
    /**
     * Constructs a communicator with default options.
     */
```

InsertCreativityHere (approved):
Looks good to me!

PR #4608, opened 2025-10-26
Do not register an OA without published endpoints

### LINKED ISSUE #4592: Do not register OA without endpoints with the locator
We currently register an Object Adapter with the locator if the AdapterId property is set.

We should also check that the Object Adapter has endpoints, registering an adapter without endpoints is not useful.

It’s valid to set the AdapterId for an Object Adapter without endpoints to enable collocated calls for default servants.

This is exactly what we do in DataStorm:

https://github.com/zeroc-ice/ice/blob/1877de043cf78963dd70dd88b72a780f5ef9e8f7/cpp/src/DataStorm/Instance.cpp#L89-L99

bernardnormier:
This PR updates the ObjectAdapter code in C++, C#, and Java, to ensure an object adapter without published endpoints does not register these endpoints with the locator.

Fixes #4592.

### DIFF
--- a/cpp/include/Ice/ObjectAdapter.h
+++ b/cpp/include/Ice/ObjectAdapter.h
@@ -298,7 +298,8 @@ namespace Ice
         [[nodiscard]] virtual EndpointSeq getPublishedEndpoints() const = 0;
 
         /// Sets the endpoints that proxies created by this object adapter will contain.
-        /// @param newEndpoints The new set of endpoints that the object adapter will embed in proxies.
+        /// @param newEndpoints The new set of endpoints that the object adapter will embed in proxies. Must contain
+        /// at least one endpoint.
         virtual void setPublishedEndpoints(EndpointSeq newEndpoints) = 0;
 
     protected:
--- a/cpp/src/Ice/ObjectAdapterI.cpp
+++ b/cpp/src/Ice/ObjectAdapterI.cpp
@@ -114,9 +114,7 @@ Ice::ObjectAdapterI::activate()
     {
         try
         {
-            Ice::Identity dummy;
-            dummy.name = "dummy";
-            updateLocatorRegistry(locatorInfo, createDirectProxy(dummy));
+            updateLocatorRegistry(locatorInfo, createDirectProxy(Identity{.name = "dummy", .category = ""}));
         }
         catch (const Ice::LocalException&)
         {
--- a/cpp/src/Ice/ObjectAdapterI.cpp
+++ b/cpp/src/Ice/ObjectAdapterI.cpp
@@ -69,6 +69,7 @@ Ice::ObjectAdapterI::activate()
 {
     LocatorInfoPtr locatorInfo;
     bool printAdapterReady = false;
+    bool hasPublishedEndpoints = false;
 
     {
         lock_guard lock(_mutex);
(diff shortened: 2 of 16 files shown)


cpp/include/Ice/ObjectAdapter.h
```diff
@@ -298,7 +298,8 @@ namespace Ice
         [[nodiscard]] virtual EndpointSeq getPublishedEndpoints() const = 0;
 
         /// Sets the endpoints that proxies created by this object adapter will contain.
-        /// @param newEndpoints The new set of endpoints that the object adapter will embed in proxies.
+        /// @param newEndpoints The new set of endpoints that the object adapter will embed in proxies. Must contain
```
bernardnormier, line 301: That's a new requirement. Registering empty published endpoints doesn't make sense / isn't useful.

I also want to allow `setPublishedEndpoints` only on an activated OA, but most of our uses of `setPublishedEndpoints` in the tests are on non-activated OAs.

cpp/test/Ice/location/AllTests.cpp
```diff
@@ -626,6 +626,16 @@ allTests(Test::TestHelper* helper, const string& ref)
...
     }
 
+    cout << "testing indirect object adapter without published endpoints... " << flush;
+    int32_t setRequestCount = registry.getSetRequestCount();
+    communicator->getProperties()->setProperty("CollocAdapter.AdapterId", "CollocId");
+    ObjectAdapterPtr collocAdapter = communicator->createObjectAdapter("CollocAdapter");
+    collocAdapter->activate();                               // not necessary, but allowed
+    test(setRequestCount == registry->getSetRequestCount()); // no set call on registry
+    collocAdapter->deactivate();
+    test(setRequestCount == registry->getSetRequestCount()); // no set call on registry
```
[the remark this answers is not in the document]
bernardnormier: getSetRequestCount in a generated proxy function that returns a `std::int32_t`.

cpp/src/Ice/ObjectAdapterI.cpp
```diff
@@ -106,28 +107,32 @@ Ice::ObjectAdapterI::activate()
...
-        // later.
-        //
+        try
         {
-            lock_guard lock(_mutex);
-            _state = StateUninitialized;
-            _conditionVariable.notify_all();
+            Ice::Identity dummy;
+            dummy.name = "dummy";
+            updateLocatorRegistry(locatorInfo, createDirectProxy(dummy));
```
pepone:
```suggestion
            updateLocatorRegistry(locatorInfo, createDirectProxy(Identity{.name = "dummy", .category = ""}));
```

csharp/msbuild/CodeAnalysis.Src.globalconfig
```diff
@@ -1,6 +1,9 @@
 global_level = 110
 is_global = true
 
+# CA1515: Because an application's API isn't typically referenced from outside the assembly, types can be made internal
+dotnet_diagnostic.CA1515.severity = none
```
pepone, line 5: Where do we get this warning?

pepone (approved):

csharp/msbuild/CodeAnalysis.Src.globalconfig
externl: Doesn't this only apply to exes? Also, why not follow this recommendation.

externl (approved):

csharp/msbuild/CodeAnalysis.Src.globalconfig
bernardnormier: With this setting, I get CA1515 on many or all local exceptions in LocalExceptions.cs. Try for yourself.

bernardnormier: > Doesn't this only apply to exes? Also, why not follow this recommendation.

We can't follow this recommendation for types that are part of the public API.

externl: I don't follow. There is no public API for an exe.

PR #4609, opened 2025-10-27
Add Communicator constructor to JavaScript

pepone:
The JS side of #4590

### DIFF
--- a/js/src/Ice/Communicator.d.ts
+++ b/js/src/Ice/Communicator.d.ts
@@ -9,7 +9,28 @@ declare module "@zeroc/ice" {
          * @see {@link ObjectAdapter}
          * @see {@link Properties}
          */
-        interface Communicator {
+        class Communicator {
+            /**
+             * Creates a communicator.
+             *
+             * @param initData Additional initialization data.
+             * @returns The initialized communicator.
+             * @throws {@link InitializationException} If an error occurs during initialization.
+             */
+            constructor(initData?: InitializationData);
+
+            /**
+             * Creates a communicator.
+             *
+             * @param args A command-line argument vector. Any Ice-related options in this vector are used to initialize
+             * the communicator. This method modifies the argument vector by removing any Ice-related options.
+             * @param initData Additional initialization data. Property settings in args override property settings in
+             * initData.
+             * @returns The initialized communicator.
+             * @throws {@link InitializationException} If an error occurs during initialization.
+             */
+            constructor(args: string[], initData?: InitializationData);
+
             /**
              * Destroys this communicator. This method calls {@link shutdown} implicitly. Calling {@link destroy}
              * destroys all object adapters, and closes all outgoing connections. {@code destroy} waits for all
--- a/js/src/Ice/Communicator.js
+++ b/js/src/Ice/Communicator.js
@@ -1,17 +1,55 @@
 // Copyright (c) ZeroC, Inc.
 
+import { CompositeSliceLoader } from "./CompositeSliceLoader.js";
+import { InitializationData } from "./InitializationData.js";
+import { InitializationException } from "./LocalExceptions.js";
+import { ObjectPrx } from "./ObjectPrx.js";
+import { Promise } from "./Promise.js";
+import { Properties } from "./Properties.js";
+import { defaultSliceLoaderInstance } from "./DefaultSliceLoader.js";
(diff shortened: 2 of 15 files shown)


js/src/Ice/Communicator.js
```diff
@@ -1,17 +1,55 @@
...
 import { identityToString } from "./IdentityToString.js";
-import { Promise } from "./Promise.js";
-import { ObjectPrx } from "./ObjectPrx.js";
 
 //
 // Ice.Communicator
 //
 export class Communicator {
-    constructor(initData) {
+    constructor(arg1, arg2) {
```
externl, line 17: I'm not really a big fan of this pattern and would prefer to have a factory method that calls the constructor.

js/src/Ice/Communicator.js
```diff
@@ -1,17 +1,55 @@
...
+                initData = arg2;
+            } else {
+                throw new InitializationException("invalid argument to initialize");
+            }
+        }
+
+        if (initData === null) {
+            initData = new InitializationData();
+        } else {
+            initData = initData.clone();
```
externl, line 43: I'm not sure there's any benefit to making a clone here.  In Java we don't appear to make a clone.  

All the objects it holds are classes. Our "deep copy" clone just copies the references. I'm not really convinced we should have this clone operation. Us making a clone here is not really helping a user who's sharing InitData between communicators. 

Maybe I'm missing something?

js/src/Ice/Communicator.js
```diff
@@ -1,17 +1,55 @@
...
         this._shutdownPromise = new Promise();
+
+        let args = null;
+        let initData = null;
+
+        if (arg1 instanceof Array) {
+            args = arg1;
+        } else if (arg1 instanceof InitializationData) {
+            initData = arg1;
+        } else if (arg1 !== undefined && arg1 !== null) {
```
externl, line 28: I thought the convention in JS was to use `!= null` as it checks both `null` and `undefined` and is short.

js/src/Ice/Communicator.js
pepone: We assign initData.properties bellow, sliceLoader, ... this avoid changing the provided object references.

js/src/Ice/Communicator.js
pepone: We use this style all over the place.

js/src/Ice/Communicator.js
bernardnormier: This PR is about _not_ using a factory method (initialize).

js/src/Ice/Communicator.d.ts
```diff
@@ -9,7 +9,28 @@ declare module "@zeroc/ice" {
...
+             * Creates a communicator.
+             *
+             * @param args A command-line argument vector. Any Ice-related options in this vector are used to initialize
+             * the communicator. This method modifies the argument vector by removing any Ice-related options.
+             * @param initData Additional initialization data. Property settings in args override property settings in
+             * initData.
+             * @returns The initialized communicator.
+             * @throws {@link InitializationException} If an error occurs during initialization.
+             */
+            constructor(args: string[], initData?: InitializationData);
```
bernardnormier, line 32: While it's a direct translation of initialize, I don't think it's correct. We should provide 3 overloads like in C++, C#, Java:
1. Parameterless
2. InitData
3. Args

In C# and C++, 1+2 are combined thanks to default value.

js/src/Ice/Communicator.js
bernardnormier: We clone initData in C# and C++ (where it's passed by value), but indeed not in Java:

```cs
initData = initData is null ? new InitializationData() : initData with { };
```

pepone: C# uses a `record class` this is still a reference type, which we copy.
C++  pass by value no additional copy

In Java we have this comment:

```
    // Only for use by com.zeroc.Ice.Communicator.
    // If the application provides an initData, the initData argument is a clone.
    void initialize(Communicator communicator, InitializationData initData) {
```

Seems clone was removed in https://github.com/zeroc-ice/ice/commit/8354d3da9f4cd39e7cb39b9f3f3488ea42bd9195

But it was not added to the Communicator constructor.

bernardnormier: It was an accidental removal.

PR #4611, opened 2025-10-27
Make Python Ice.Properties constructor public

### LINKED ISSUE #4607: Prefer Properties init over createProperties in Python
We want to fix the Properties initializer method in Python and prefer this method over createProperties  like we already do in other languages (C++, C#, Java, MATLAB).

See also #4603.

pepone:
Fix #4607

### DIFF
--- a/python/python/Ice/Properties.py
+++ b/python/python/Ice/Properties.py
@@ -19,7 +19,9 @@ class Properties:
     @overload
     def __init__(self, *, properties: IcePy.Properties): ...
 
-    def __init__(self, args: list[str] | None = None, defaults: Self | None = None, properties: IcePy.Properties | None = None):
+    def __init__(
+        self, args: list[str] | None = None, defaults: Self | None = None, properties: IcePy.Properties | None = None
+    ):
         """
         Initialize a new instance of the Properties class.
 


python/python/Ice/Properties.py
```diff
@@ -13,8 +13,53 @@ class Properties:
     application-name[.category[.sub-category]].name.
     """
 
-    def __init__(self, impl: IcePy.Properties):
-        self._impl = impl
+    def __init__(self, args: list[str] | IcePy.Properties | None = None, defaults: Self | None = None):
```
externl: What if we move `IcePy.Properties | None` to the end, then add an `@overload`  that's just args and defaults, and one that's just IcePy.Properties. 

Maybe we only want the first overload.

python/python/Ice/Util.py
```diff
@@ -108,13 +108,17 @@ def createProperties(args: list[str] | None = None, defaults: Properties | None
     If an argument list is supplied, this function parses arguments starting with ``--`` and a known Ice prefix
     (e.g., ``Ice``, ``IceSSL``), and removes recognized arguments from the list.
 
+    This function is provided for backwards compatibility. New code should call the :class:`Properties` constructor
```
bernardnormier, line 105: I would simplify the doc-comment above, like in C++, C#, Java.

python/python/Ice/Util.py
```diff
@@ -124,23 +128,23 @@ def createProperties(args: list[str] | None = None, defaults: Properties | None
     Examples
     --------
     .. code-block:: python
-
         # Create a new empty property set.
```
bernardnormier: I would move the examples to the Properties init. No need to provide examples for backwards-compatibility functions.

python/python/Ice/Properties.py
```diff
@@ -13,8 +13,53 @@ class Properties:
...
+        """
+        Creates a new property set.
```
bernardnormier: Is Creates... a standard description for init methods?

python/python/Ice/Properties.py
bernardnormier: @externl suggestion also means updating the C++ code, which this PR so far does not.

python/python/Ice/Properties.py
pepone: Initializes seems slightly more common.

python/python/Ice/Properties.py
pepone: > Maybe we only want the first overload.

You need at-least two overloads.

> means updating the C++ code, which this PR so far does not.

Why do we need to update C++ for this?

python/python/Ice/Properties.py
```diff
@@ -13,8 +13,59 @@ class Properties:
     application-name[.category[.sub-category]].name.
     """
 
-    def __init__(self, impl: IcePy.Properties):
-        self._impl = impl
+    @overload
+    def __init__(self, args: list[str] | None = None, defaults: Self | None = None): ...
+
+    @overload
+    def __init__(self, *, properties: IcePy.Properties): ...
```
externl, line 20: Do we need this overload? I assume we only call this constructor from the C++ code.

externl (approved):

python/python/Ice/Properties.py
pepone: Communicator.py uses it:

https://github.com/zeroc-ice/ice/blob/0ca219f622e726b512f6201fe1a85f383c1be64e/python/python/Ice/Communicator.py#L396-L406

externl: Should we just have the Python Communicator hold a Ice.Properties instead?

pepone: I will try in a separate PR. With the current code we some times create the properties in the C++ side, and we later wrap them with this constructor when they are accessed from Python.

We can instead create an empty Python property set in the Communicator constructor if they are not set in initData and there are not args.

PR #4614, opened 2025-10-28
JavaScript Communicator constructor cleanup

pepone:


js/src/Ice/Initialize.d.ts
```diff
@@ -2,6 +2,17 @@
...
+        /**
+         * Creates a communicator.
+         *
+         * This method is provided for backwards compatibility. New code should call the {@link Communicator}
+         * constructor directly.
+         *
+         * @returns The initialized communicator.
+         * @throws {@link InitializationException} If an error occurs during initialization.
+         */
+        function initialize(): Communicator;
```
bernardnormier, line 14: Should we still show the 2-3 overloads? 

With the current syntax, it's not clear which arguments you can pass to initialize.

js/src/Ice/Communicator.js
```diff
@@ -14,35 +14,26 @@ import { identityToString } from "./IdentityToString.js";
...
-            throw new InitializationException("invalid argument to initialize");
-        }
-
-        if (arg2 !== undefined && arg2 !== null) {
-            if (arg2 instanceof InitializationData && initData === null) {
-                initData = arg2;
+        if (arguments.length === 0) {
+            initData = new InitializationData();
+        } else if (arguments.length === 1) {
+            const arg1 = arguments[0];
```
bernardnormier: arg0 sounds like a better name.

js/src/Ice/Communicator.js
```diff
@@ -14,35 +14,26 @@ import { identityToString } from "./IdentityToString.js";
...
+            if (arg1 instanceof Array) {
+                initData = new InitializationData();
+                initData.properties = new Properties(arg1);
+            } else if (arg1 instanceof InitializationData) {
+                initData = arg1.clone();
             } else {
-                throw new InitializationException("invalid argument to initialize");
+                throw new InitializationException("invalid argument pass to Communicator constructor");
             }
```
bernardnormier, line 33: I would something like write: 

invalid argument provided to the Communicator constructor

js/src/Ice/Communicator.d.ts
```diff
@@ -10,6 +10,14 @@ declare module "@zeroc/ice" {
          * @see {@link Properties}
          */
         class Communicator {
+            /**
+             * Creates a communicator.
```
bernardnormier: with default options?

js/src/Ice/Communicator.d.ts
```diff
@@ -10,6 +10,14 @@ declare module "@zeroc/ice" {
...
+             *
+             * @returns The initialized communicator.
+             * @throws {@link InitializationException} If an error occurs during initialization.
+             */
+            constructor();
+
```
bernardnormier, line 20: For initData: see doc-comment in other languages

js/src/Ice/Initialize.d.ts
pepone: The doc will should the 3 overloads, and for TypeScript the typechecker will enforce them

```
initialize();
initialize(args);
initialize(initData);
```

Matching the communicator constructor.

PR #4617, opened 2025-10-29
Add Slice to Csharp --enable-analysis

### LINKED ISSUE #3700: slice2cs --enable-analysis
We could however add a slice2cs compiler flag that:
 - suppress the `<auto-generated>` header
 - generates these pragmas in the .cs code

Something like `--enable-analysis`? As an undocumented option? And we could use it in some CI builds.

_Originally posted by @bernardnormier in https://github.com/zeroc-ice/ice/pull/3694#discussion_r1995687192_

pepone:
Fix #3700

### DIFF
--- a/csharp/test/Ice/ami/AllTests.cs
+++ b/csharp/test/Ice/ami/AllTests.cs
@@ -57,7 +57,7 @@ public class AllTests : global::Test.AllTests
 
         public void Report(bool value)
         {
-            SentSynchronously = sentSynchronously;
+            SentSynchronously = value;
             Sent = true;
         }
 


cpp/src/slice2cs/Main.cpp
```diff
@@ -105,6 +105,8 @@ usage(const string& n)
                   "--depend                 Generate Makefile dependencies.\n"
                   "--depend-xml             Generate dependencies in XML format.\n"
                   "--depend-file FILE       Write dependencies to FILE instead of standard output.\n"
+                  "--enable-analysis        Enable static analysis for generated code by not including\n"
+                  "                         <auto-generated> tag which suppresses analyzer warnings.\n"
```
InsertCreativityHere: Since there are potentially multiple tags involved (one per file).
```suggestion
                  "                         <auto-generated> tags which suppresses analyzer warnings.\n"
```

csharp/test/Ice/ami/AllTests.cs
```diff
@@ -55,10 +55,10 @@ public bool SentSynchronously
             }
         }
 
-        public void Report(bool sentSynchronously)
+        public void Report(bool value)
         {
-            SentSynchronously = sentSynchronously;
-            Sent = true;
+            _sentSynchronously = value;
+            _sent = true;
```
InsertCreativityHere: Not familiar with this specific test, but it does seem a little suspicious not using the mutex here.

pepone: This was an unintended change, it was mean to just rename `sentSynchronously` to `value` to match the declaration and avoid the warning.

externl (approved):
If we plan to have a public flag, we should also update the man page. Is there a reason a user would want to use this flag?

pepone:
> If we plan to have a public flag, we should also update the man page. Is there a reason a user would want to use this flag?

I think we should documented it, I don't like the idea of including "non supported/hidden" flags.

PR #4619, opened 2025-10-31
Disable inactivity check for DataStorm session connections

### LINKED ISSUE #4618: Disable InactivityTimeout for DataStorm session connections
DataStorm maintains sessions between connected nodes. The session is associated with a connection and is automatically established when the connection is closed. We want to keep this connections open even if they are inactive, it also avoid recreating sessions when the inactivity timeout closes the connections.

pepone:
Fix #4618

### DIFF
--- a/cpp/src/DataStorm/ConnectionManager.cpp
+++ b/cpp/src/DataStorm/ConnectionManager.cpp
@@ -17,11 +17,10 @@ ConnectionManager::add(
     shared_ptr<void> object,
     function<void(const ConnectionPtr&, exception_ptr)> callback)
 {
-    lock_guard<mutex> lock(_mutex);
-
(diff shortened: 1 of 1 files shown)


cpp/src/DataStorm/ConnectionManager.cpp
```diff
@@ -18,6 +18,10 @@ ConnectionManager::add(
     function<void(const ConnectionPtr&, exception_ptr)> callback)
 {
     lock_guard<mutex> lock(_mutex);
+
+    // Disable the inactivity timeout on the connection.
```
bernardnormier, line 20: I would do this before locking the mutex, because it's cleaner.

As written, you wonder if this locking is required or if there is a potential lock-acquisition issue between this mutex and the connection mutex.

PR #4621, opened 2025-10-31
Enable iOS swift/Ice/udp

externl:
This PR fixes the swift/Ice/udp test on iOS. The main issue was lack of threads (really blocking the few threads we have). We were blocking the Swift threads in `communicator.waitForShutdown()` and a in a part of the test implementation.

### DIFF
--- a/swift/test/Ice/udp/AllTests.swift
+++ b/swift/test/Ice/udp/AllTests.swift
@@ -42,10 +42,7 @@ class PingReplyI: PingReply, @unchecked Sendable {
                 return false
             }
 
-            guard let result = await group.next() else {
-                fatalError("empty task group")
-            }
-
+            let result = await group.next()!
             group.cancelAll()
             return result
         }


swift/test/Ice/udp/AllTests.swift
```diff
@@ -1,30 +1,54 @@
...
+                }
+                return true
+            }
+
+            group.addTask {
+                try? await Task.sleep(for: .milliseconds(timeout))
+                return false
+            }
+
+            guard let result = await group.next() else {
```
pepone: Can't we just do `let result = await group.next()!` we know the group is not empty.

pepone (approved):

swift/test/Ice/udp/AllTests.swift
externl: Sure

InsertCreativityHere (approved):
Looks fine to me!

PR #4622, opened 2025-11-03
Format see also in Python & Swift

### LINKED ISSUE #4570: Map `@see` Tags in Python and Swift
Currently, `@see` tags are properly mapped in C++, C#, Java, JavaScript, and MATLAB.
We don't map any doc-comments at all for PHP and Ruby.

This leaves Python and Swift, which should handle this tag instead of ignoring it (what they currently do).

pepone:
Fix #4570

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -46,11 +46,6 @@ class PythonDocCommentFormatter final : public DocCommentFormatter
     {
         return Slice::Python::pyLinkFormatter(rawLink, source, target);
     }
-
-    string formatSeeAlso(const string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target) final
-    {
-        return Slice::Python::pyLinkFormatter(rawLink, source, target);
-    }
 };
 
 string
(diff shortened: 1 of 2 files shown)


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -46,6 +46,11 @@ class PythonDocCommentFormatter final : public DocCommentFormatter
     {
         return Slice::Python::pyLinkFormatter(rawLink, source, target);
     }
+
+    string formatSeeAlso(const string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target) final
+    {
+        return Slice::Python::pyLinkFormatter(rawLink, source, target);
+    }
```
InsertCreativityHere: This isn't necessary, `formatSeeAlso` is default-implemented to call `formatLink`:
https://github.com/zeroc-ice/ice/blob/daaa6e39bde259b7f8089c60db81d18eb29c7eff/cpp/src/Slice/DocCommentParser.cpp#L68-L73
and indeed, this has the same implementation as `formatLink`.

cpp/src/slice2swift/Main.cpp
```diff
@@ -32,6 +32,11 @@ class SwiftDocCommentFormatter final : public DocCommentFormatter
     {
         return Slice::Swift::swiftLinkFormatter(rawLink, source, target);
     }
+
+    string formatSeeAlso(const string& rawLink, const ContainedPtr& source, const SyntaxTreeBasePtr& target) final
+    {
+        return Slice::Swift::swiftLinkFormatter(rawLink, source, target);
+    }
```
InsertCreativityHere: Also not necessary, will already re-use `formatLink` automatically.

PR #4625, opened 2025-11-03
Fix bogus server session released check in IceGrid node.

### LINKED ISSUE #4624: Bogus serverSessionReleased check in IceGrid
This check looks bogus `_desc` and `descriptor` point to the same `InternalServerDescriptor`.

https://github.com/zeroc-ice/ice/blob/daaa6e39bde259b7f8089c60db81d18eb29c7eff/cpp/src/IceGrid/ServerI.cpp#L1953-L1961

Seems like the intention was to this check before than `_desc` is assigned, but looking back this is present in the initial implementation of IceGrid.

pepone:
Fix #4624

### DIFF
--- a/cpp/src/IceGrid/ServerI.cpp
+++ b/cpp/src/IceGrid/ServerI.cpp
@@ -1952,14 +1952,14 @@ ServerI::updateImpl(const shared_ptr<InternalServerDescriptor>& descriptor)
 {
     assert(_load && descriptor);
 
-    _desc = descriptor;
-    _waitForReplication = true;
-
     // Remember if the server was just released by a session, this will be used later to not update the configuration
     // on the disk (as an optimization and to allow users to review the configuration file after allocating a server --
     // that's useful if the server configuration is bogus and the session server can't start).
     bool serverSessionReleased = _desc && _desc->activation == "session" && _desc->revision == descriptor->revision &&
                                  !_desc->sessionId.empty() && descriptor->sessionId.empty();
+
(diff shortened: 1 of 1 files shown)


cpp/src/IceGrid/ServerI.cpp
```diff
@@ -1952,15 +1952,15 @@ ServerI::updateImpl(const shared_ptr<InternalServerDescriptor>& descriptor)
 {
     assert(_load && descriptor);
 
-    _desc = descriptor;
-    _waitForReplication = true;
-
     // Remember if the server was just released by a session, this will be used later to not update the configuration
     // on the disk (as an optimization and to allow users to review the configuration file after allocating a server --
     // that's useful if the server configuration is bogus and the session server can't start).
     bool serverSessionReleased = _desc && _desc->activation == "session" && _desc->revision == descriptor->revision &&
```
pepone, line 1958: This check only makes sense if we do it before updating `_desc`.

externl (approved):

cpp/src/IceGrid/ServerI.cpp
bernardnormier: Previously, serverSessionReleased was always false since `!_desc->sessionId.empty() && descriptor->sessionId.empty()` is necessarily false when both _desc and descriptor are the same.

Do we have any test that exercises the new optimized code path (the return on line 2105 below)? Would be interesting to check the coverage report.

PR #4626, opened 2025-11-03
Enable and Fix Java Lints that Come from the Compiler

InsertCreativityHere:
This PR enables all the built in code & doc lints built into the Java compiler,
and then excludes some of the unimportant ones which are tedious to fix.

I also cleaned up the `TestHelper::getTestEndpoint` method in Java; we had 6 overloads and 3 of them were barely used.

### DIFF
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -3946,6 +3946,9 @@ Slice::Gen::TypesVisitor::visitSequence(const SequencePtr& p)
     out << eb;
     out << eb;
 
+    out << sp;
+    out << nl << "private " << name << "Helper() {}";
+
     out << eb;
     close();
 }
@@ -4070,6 +4073,9 @@ Slice::Gen::TypesVisitor::visitDictionary(const DictionaryPtr& p)
     out << nl << "return java.util.Optional.empty();";
     out << eb;
     out << eb;
+
+    out << sp;
+    out << nl << "private " << name << "Helper() {}";
 
     out << eb;
     close();


cpp/src/slice2java/Gen.cpp
```diff
@@ -2928,7 +2928,7 @@ Slice::Gen::TypesVisitor::visitExceptionEnd(const ExceptionPtr& p)
         out << sp;
         writeDocComment(
             out,
-            "@{inheritDoc}\n"
```
InsertCreativityHere, line 2931: This '@' sign has always been in the wrong place, and we just never noticed apparently.

cpp/src/slice2java/Gen.cpp
```diff
@@ -3946,6 +3946,9 @@ Slice::Gen::TypesVisitor::visitSequence(const SequencePtr& p)
     out << eb;
     out << eb;
 
+    out << sp;
+    out << nl << "private " << name << "Helper() {}";
```
InsertCreativityHere, line 3950: This is in the generated sequence helper class.
These helpers classes are fully static, so the compiler wants them to have private constructors:
i.e. make it impossible to construct an instance of these classes.

cpp/src/slice2java/Gen.cpp
```diff
@@ -4071,6 +4074,9 @@ Slice::Gen::TypesVisitor::visitDictionary(const DictionaryPtr& p)
...
+
```
InsertCreativityHere, line 4079: This is for the dictionary helper class.

Same thing, make it impossible to construct an instance of this fully static helper class.

java/build.gradle
```diff
@@ -43,7 +43,8 @@ subprojects {
     }
 
     tasks.withType(JavaCompile) {
-        options.compilerArgs << "-Xlint:unchecked" << "-Xlint:-options"
+        options.compilerArgs << "-Xdoclint:all,-missing"
+                             << "-Xlint:all,-rawtypes,-exports,-serial,-try,-missing-explicit-ctor,-deprecation"
```
InsertCreativityHere, line 47: Enable all lints, except for the ones with minus signs in front.
Maybe we can come back and clean these all up too, but they're not that important, and affect hundreds of lines of code.

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/LiveDeployment/MetricsViewEditor.java
```diff
@@ -429,7 +429,7 @@ public boolean isDataFlavorSupported(DataFlavor flavor) {
         private MetricsViewTransferableData _data;
     }
 
-    /** Class used to transfer selected metric fields with drag & drop. */
+    /** Class used to transfer selected metric fields with drag {@literal &} drop. */
```
InsertCreativityHere, line 432: JavaDoc was treating this '&' as a broken HTML escape.
We want a literal '&' in our comment.

java/src/com.zeroc.icediscovery/src/main/java/com/zeroc/IceDiscovery/LookupI.java
```diff
@@ -221,7 +221,7 @@ public LookupI(LocatorRegistryI registry, LookupPrx lookup, Properties propertie
         Endpoint[] single = new Endpoint[1];
         for (Endpoint endpt : lookup.ice_getEndpoints()) {
             single[0] = endpt;
```
InsertCreativityHere, line 223: Lots of unnecessary casts leftover from before we changed the `ice_xxx` methods to return a properly downcasted type.

java/test/src/main/java/test/Ice/executor/AllTests.java
```diff
@@ -74,7 +74,7 @@ public static void allTests(TestHelper helper, final CustomExecutor executor) {
 
         int mult = 1;
         if (!"tcp".equals(communicator.getProperties().getIceProperty("Ice.Default.Protocol"))
-            || helper.isAndroid()) {
```
InsertCreativityHere, line 77: Some of the `TestHelper` functions are static, and we were calling them on an instance.

java/test/src/main/java/test/Ice/optional/SerializableClass.java
```diff
@@ -18,6 +18,11 @@ public boolean equals(java.lang.Object obj) {
         return false;
     }
 
+    @Override
+    public int hashCode() {
+        return _v;
+    }
+
```
InsertCreativityHere, line 25: Java requires that if two things are `equal`, that their `hashCode`s must also be equal.
Right above this, we override `equals`, but weren't overriding `hashCode`, breaking this contract.

PR #4628, opened 2025-11-04
Test renamed classes

### LINKED ISSUE #3583: Add Tests To Ensure Remapped Classes and Exceptions can be Successfully Unmarshaled
We should add tests to ensure that we can successfully decode classes and exceptions which have been remapped by either: `xxx:identifier` or each language's version of `cpp:namespace`.

pepone:
Fix #3583

### DIFF
--- a/swift/test/Ice/objects/AllTests.swift
+++ b/swift/test/Ice/objects/AllTests.swift
@@ -109,6 +109,17 @@ func allTests(_ helper: TestHelper) async throws -> InitialPrx {
     try test(dout !== nil)
     output.writeLine("ok")
 
+    output.write("testing renamed classes... ")
+    var renamed = SwiftClass(data: "renamed")
+    renamed = try await initial.opRenamedClass(renamed)!
+    try test(renamed.data == "renamed")
+
+    var derivedRenamed = SwiftDerivedClass(data: "renamed", moreData: "derived")
+    derivedRenamed = try await initial.opRenamedClass(derivedRenamed)! as! SwiftDerivedClass
+    try test(derivedRenamed.data == "renamed")
+    try test(derivedRenamed.moreData == "derived")
+    output.writeLine("ok")
+
     output.write("checking consistency... ")
     try test(b1out !== b2out)
     try test(b1out!.theA === b2out)


pyrightconfig.json
```diff
@@ -13,4 +13,5 @@
     "reportUnusedClass": "error",
     "reportUnusedFunction": "error",
     "reportUnusedVariable": "error",
+    "pythonVersion": "3.12"
```
InsertCreativityHere, line 16: Is this intentional?

csharp/test/Ice/objects/AllTests.cs
```diff
@@ -130,6 +132,20 @@ public static InitialPrx allTests(global::Test.TestHelper helper)
                 test(dout != null);
                 output.WriteLine("ok");
 
+                output.Write("testing renamed classes... ");
+                output.Flush();
+                var renamed = new Test.CsClass("renamed");
+                renamed = initial.opRenamedClass(renamed);
+                Debug.Assert(renamed is not null);
```
InsertCreativityHere, line 139: We should probably be using `test` instead of asserting, so our test system can gracefully handle a failure, instead of dying.
```suggestion
                test(renamed is not null);
```

csharp/test/Ice/objects/AllTests.cs
```diff
@@ -130,6 +132,20 @@ public static InitialPrx allTests(global::Test.TestHelper helper)
...
+                test(renamed.data == "renamed");
+
+                var derivedRenamed = new Test.CsDerivedClass("renamed", "derived");
+                derivedRenamed = initial.opRenamedClass(derivedRenamed) as Test.CsDerivedClass;
+                Debug.Assert(derivedRenamed is not null);
```
InsertCreativityHere, line 144: Same here
```suggestion
                test(derivedRenamed is not null);
```

csharp/test/Ice/objects/AllTests.cs
```diff
@@ -1,5 +1,7 @@
 // Copyright (c) ZeroC, Inc.
 
+using System.Diagnostics;
+
```
InsertCreativityHere, line 4: Also then this new import wouldn't be necessary:
```suggestion

```

swift/test/Ice/objects/Test.ice
```diff
@@ -171,6 +171,19 @@ module Test
         F2* f2;
     }
```
InsertCreativityHere, line 172: The Slice file was changed in Swift, but the actual test is missing for Swift.

pyrightconfig.json
pepone: Yes, this avoids the language-server making recommendations for compatibility with older versions.

PR #4634, opened 2025-11-05
Use async/await in Swift iOS controller

externl:


swift/test/ios/TestDriverApp/ControllerI.swift
```diff
@@ -140,23 +140,44 @@ class ControllerHelperI: ControllerHelper, TextWriter, @unchecked Sendable {
...
+
+    private let _state = Mutex(State())
 
     public init(view: ViewController, testName: String, args: [String], exe: String) {
         _view = view
         _testName = testName
         _args = args
         _exe = exe
+
+        let (readyStream, readyStreamContinuation) = AsyncStream<Void>.makeStream()
```
pepone: Can we use:

```
(_readyStream, _readyStreamContinuation) = ...
```

swift/test/ios/TestDriverApp/ControllerI.swift
```diff
@@ -140,23 +140,44 @@ class ControllerHelperI: ControllerHelper, TextWriter, @unchecked Sendable {
...
+
+        _readyStream = readyStream
+        _readyStreamContinuation = readyStreamContinuation
+
+        _completedStream = completedStream
+        _completedStreamContinuation = completedStreamContinuation
+    }
+
+    deinit {
+        _communicator?.getLogger().print("testName: \(_testName) exe: \(_exe) ControllerHelperI deinit called")
```
pepone: What is the point of this tracing?

swift/test/ios/TestDriverApp/ControllerI.swift
```diff
@@ -207,45 +230,104 @@ class ControllerHelperI: ControllerHelper, TextWriter, @unchecked Sendable {
...
-                if status == 0 {
-                    return
-                } else {
-                    throw CommonProcessFailedException(reason: self._out)
+            group.addTask {
+                // The task cancellation from another thread that's also reading the stream
+                // can cause this task stream's iterator to return. We just loop until
+                // we get the completed status or the task is cancelled.
+                var completedStatus: Int32?
+                while completedStatus == nil && !Task.isCancelled {
```
pepone, line 247: Why is the loop required? What about something like:

```swift
var iterator = self._completedStream.makeAsyncIterator()
_ = await iterator.next()

if Task.isCancelled {
    return .timeout
} else {
    let status = self._state.withLock { $0.completedStatus! }
    return .completed(status)
}
```

swift/test/ios/TestDriverApp/ControllerI.swift
```diff
@@ -207,45 +230,104 @@ class ControllerHelperI: ControllerHelper, TextWriter, @unchecked Sendable {
...
+                try? await Task.sleep(for: .seconds(Int(timeout)))
+                return .timeout
+            }
+
+            group.addTask {
+                // The task cancellation from another thread that's also reading the stream
+                // can cause this task stream's iterator to return. We just loop until
+                // we get the completed status or the task is cancelled.
+                var completedStatus: Int32?
+                while completedStatus == nil && !Task.isCancelled {
```
pepone, line 294: ditto for this loop.

swift/test/ios/TestDriverApp/ControllerI.swift
externl: Leftovers.

swift/test/ios/TestDriverApp/ControllerI.swift
externl: The comment explains it, maybe poorly. It's because we're reading from the stream in two Tasks. If one of those tasks gets cancelled, it cancels the stream reading in both. Not to be confused with the task in the task group though.

pepone: I don't follow where this cancelation happens, here the only visible cancelation is per group, and there is a group per call.

externl: From what I can figure out is that we have two tasks running, both reading from this stream with the for loop. This creates two iterators under the hood. It seems that when one task group is cancelled, it somehow is causes the stream read to return in the other task.

externl: Another approach is to create a third stream.

pepone: Seems a rather odd behavior.

externl: Yeah. It's either this or create another stream.

pepone (approved):

swift/test/ios/TestDriverApp/ControllerI.swift
pepone: What seems to happen here is that different iterators return the same next task, when we cancel this task if affects all callers event if using different iterators because there is a single shared task.

Issue #4635, opened 2025-11-06
Add Support for .NET10


pepone:
There is a number of build failures when upgrading to .NET 10

```
  Ice net10.0.0 failed with 13 error(s) (6,8s)
    /Users/jose/Documents/3.8/ice/csharp/src/Ice/SSL/SSLEngine.cs(87,28): error SYSLIB0057: 'X509Certificate2.X509Certificate2(string, SecureString?, X509KeyStorageFlags)' is obsolete: 'Loading certificate data through the constructor or Import is obsolete. Use X509CertificateLoader instead to load certificates.' (https://aka.ms/dotnet-warnings/SYSLIB0057)
    /Users/jose/Documents/3.8/ice/csharp/src/Ice/SSL/SSLEngine.cs(91,28): error SYSLIB0057: 'X509Certificate2.X509Certificate2(string, string?, X509KeyStorageFlags)' is obsolete: 'Loading certificate data through the constructor or Import is obsolete. Use X509CertificateLoader instead to load certificates.' (https://aka.ms/dotnet-warnings/SYSLIB0057)
    /Users/jose/Documents/3.8/ice/csharp/src/Ice/SSL/SSLEngine.cs(141,21): error SYSLIB0057: 'X509Certificate2Collection.Import(string)' is obsolete: 'Loading certificate data through the constructor or Import is obsolete. Use X509CertificateLoader instead to load certificates.' (https://aka.ms/dotnet-warnings/SYSLIB0057)
    /Users/jose/Documents/3.8/ice/csharp/src/Ice/SSL/SSLEngine.cs(178,42): error SYSLIB0058: 'SslStream.HashAlgorithm' is obsolete: 'KeyExchangeAlgorithm, KeyExchangeStrength, CipherAlgorithm, CipherStrength, HashAlgorithm and HashStrength properties of SslStream are obsolete. Use NegotiatedCipherSuite instead.' (https://aka.ms/dotnet-warnings/SYSLIB0058)
    /Users/jose/Documents/3.8/ice/csharp/src/Ice/SSL/SSLEngine.cs(178,71): error SYSLIB0058: 'SslStream.HashStrength' is obsolete: 'KeyExchangeAlgorithm, KeyExchangeStrength, CipherAlgorithm, CipherStrength, HashAlgorithm and HashStrength properties of SslStream are obsolete. Use NegotiatedCipherSuite instead.' (https://aka.ms/dotnet-warnings/SYSLIB0058)
[...]

PR #4636, opened 2025-11-06
.NET 10 build fixes

pepone:
This PR updates C# sources to compile with .NET 10 (tested with RC2).

See #4635

### DIFF
--- a/.github/actions/setup-dotnet/action.yml
+++ b/.github/actions/setup-dotnet/action.yml
@@ -9,12 +9,12 @@ runs:
   using: "composite"
   steps:
     - name: Setup .NET 8
-      uses: actions/setup-dotnet@v4
+      uses: actions/setup-dotnet@v5
       with:
-        dotnet-version: 8.x
+        dotnet-version: 8.0.x
 
     - name: Setup .NET 10
       if: inputs.include_net10 == 'true'
-      uses: actions/setup-dotnet@v4
+      uses: actions/setup-dotnet@v5
       with:
-        dotnet-version: 10.x
+        dotnet-version: 10.0.x


.github/workflows/ci.yml
```diff
@@ -134,6 +158,12 @@ jobs:
       - name: Setup .NET
         uses: ./.github/actions/setup-dotnet
 
+      - uses: actions/setup-dotnet@v4
```
externl: We already have setup-botnet above, we should integrate theses settings into it

pepone: Not sure we only need this for .NET 10 tests, the other we run in a few more places.

What do you suggest add an extra parameter to setup-dotnet?

.github/actions/setup-dotnet/action.yml
externl, line 1: Do we need both .NET 8 and .NET 10 for these builds?

pepone: Yes, Ice assemblies always target .NET 8, only the tests are built with .NET 10 target framework.

csharp/msbuild/CodeAnalysis.Tests.globalconfig
```diff
@@ -32,3 +32,6 @@ dotnet_diagnostic.CA1016.severity = none
 
 # CA2008: Do not create tasks without passing a TaskScheduler
 dotnet_diagnostic.CA2008.severity = none
+
+# CA1515: Because an application's API isn't typically referenced from outside the assembly, types can be made internal
```
bernardnormier, line 36: You moved this rule because you were getting this warning for tests too?

csharp/test/IceSSL/configuration/PlatformTests.cs
```diff
@@ -43,8 +43,8 @@ private static void clientValidatesServerUsingValidationCallback(TestHelper help
     {
         Console.Out.Write("client validates server certificate using validation callback... ");
         Console.Out.Flush();
-        using var serverCertificate =
-            new X509Certificate2(Path.Combine(certificatesPath, "ca1/server.p12"), "password");
+        using X509Certificate2 serverCertificate =
+            X509CertificateLoader.LoadPkcs12FromFile(Path.Combine(certificatesPath, "ca1/server.p12"), "password");
```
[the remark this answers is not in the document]
bernardnormier: Sounds correct.

csharp/msbuild/CodeAnalysis.Tests.globalconfig
pepone: Yes with .NET 10 this is enabled by default, I want to fix this in a separate PR.

PR #4639, opened 2025-11-07
Return undefined for non existing enumerators

pepone:
This PR upgrades the implementation of JavaScript EnumBase to match the documented behavior.

The `valueOf` method must return undefined for unknown enumerators.

See https://docs.zeroc.com/ice/latest/javascript/enumerations

### DIFF
--- a/cpp/src/slice2js/Gen.cpp
+++ b/cpp/src/slice2js/Gen.cpp
@@ -2759,11 +2759,10 @@ Slice::Gen::TypeScriptVisitor::visitEnum(const EnumPtr& p)
     }
     _out << nl;
     _out << nl << "/**";
-    _out << nl << " * Returns the enumerator for the given value.";
+    _out << nl << " * Returns the enumerator corresponding to the specified numeric value.";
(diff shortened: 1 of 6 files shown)


cpp/src/slice2js/Gen.cpp
```diff
@@ -2762,7 +2762,8 @@ Slice::Gen::TypeScriptVisitor::visitEnum(const EnumPtr& p)
     _out << nl << " * Returns the enumerator for the given value.";
     _out << nl << " *";
     _out << nl << " * @param value The enumerator value.";
-    _out << nl << " * @returns The enumerator for the given value.";
+    _out << nl << " * @returns The enumerator for the given value or undefined if the value does not correspond to any";
+    _out << nl << " * enumerator.";
     _out << nl << " */";
     _out << nl << "static valueOf(value: number): " << name << ";";
```
InsertCreativityHere: The return type of this signature should probably be updated now right?
```suggestion
    _out << nl << "static valueOf(value: number): " << name << " | undefined;";
```

PR #4641, opened 2025-11-10
Add doc comments for Ice.HashMap

pepone:
This PR documents Ice.HashMap TypeScript definition, and remove some non standard methods from its API.

The goal of HashMap is to provide a Map like type that can use custom comparison for the keys, the standard JavaScript Map type always use `===` identity comparison. The Ice mapping uses `equals` for key that are mapped to a struct type.

The JavaScript map type doesn't provide `equals` or `merge` I removed those.
The `[Symbol.iterator]` method was provided by the implementation but missing in the TypeScript definitions.

### DIFF
--- a/js/src/Ice/HashMap.d.ts
+++ b/js/src/Ice/HashMap.d.ts
@@ -33,7 +33,7 @@ declare module "@zeroc/ice" {
             constructor(keyComparator: (k1: Key, k2: Key) => boolean);
 
             /**
-             * Adds a new entry with a specified key and value to this Map, or updates an existing entry if the key
+             * Adds a new entry with a specified key and value to this HashMap, or updates an existing entry if the key
              * already exists.
              *
              * @param key The key of the entry.


js/src/Ice/HashMap.d.ts
```diff
@@ -13,27 +13,101 @@ declare module "@zeroc/ice" {
...
             delete(key: Key): Value | undefined;
+
+            /**
+             * Removes all entries from the HashMap.
+             */
             clear(): void;
+
+            /**
+             * Executes a provided function once for each key-value pair in the HashMap.
+             * @param fn The function to execute for each entry.
```
externl, line 76:
```suggestion
             * Executes a provided function once for each key-value pair in the HashMap.
             *
             * @param fn The function to execute for each entry.
```

js/src/Ice/HashMap.d.ts
```diff
@@ -13,27 +13,101 @@ declare module "@zeroc/ice" {
               };
 
         class HashMap<Key extends HashMapKey, Value> {
+            /**
+             * Creates an empty HashMap that uses the operator === for key and value comparisons.
+             */
             constructor();
+
+            /**
+             * Creates a copy of the specified HashMap.
```
InsertCreativityHere:
```suggestion
             * Creates a copy from the specified HashMap.
```

js/src/Ice/HashMap.d.ts
```diff
@@ -13,27 +13,101 @@ declare module "@zeroc/ice" {
...
+            /**
+             * Creates an empty HashMap that uses the specified key comparator for key comparisons.
+             *
+             * @param keyComparator - A function that compares two keys for equality.
+             */
+            constructor(keyComparator: (k1: Key, k2: Key) => boolean);
+
+            /**
+             * Adds a new entry with a specified key and value to this Map, or updates an existing entry if the key
+             * already exists.
```
InsertCreativityHere: This isn't wrong, but everywhere we go out of our way to call this a "HashMap" and not just a "Map".

```suggestion
             * Adds a new entry with a specified key and value to this HashMap, or updates an existing entry if the key
             * already exists.
```

PR #4642, opened 2025-11-10
Install Windows Debug tools on windows-2025 images

pepone:
The Windows Debug Tools are missing in Windows 2025 images. This PR installs them using a custom action.

### DIFF
--- a/.github/actions/install-windows-debug-tools/action.yml
+++ b/.github/actions/install-windows-debug-tools/action.yml
@@ -1,8 +1,9 @@
 name: Windows Debug Tools Installer
+description: "Installs Debugging Tools for Windows using the Windows SDK installer."
 runs:
   using: "composite"
   steps:
-    - name: Install Debugging Tools via Visual Studio Installer
+    - name: Install Debugging Tools for Windows
       shell: pwsh
       run: |
         $sdkPath = "$env:TEMP\sdksetup.exe"


.github/actions/install-windows-debug-tools/action.yml
```diff
@@ -0,0 +1,22 @@
...
+    - name: Install Debugging Tools via Visual Studio Installer
+      shell: pwsh
+      run: |
+        $sdkPath = "$env:TEMP\sdksetup.exe"
+
+        Write-Host "Downloading Windows SDK installer..."
+        Invoke-WebRequest -Uri "https://go.microsoft.com/fwlink/?linkid=2338977" -OutFile $sdkPath
+
+        Write-Host "Running Windows SDK installer (Debugging Tools only)..."
+        $args = '/quiet /norestart /features OptionId.WindowsDesktopDebuggers /log "{0}"' -f sdk-install.log
```
externl: I would write the log file to tmp. Otherwise this log file will get picked up with the test of our test log files

.github/workflows/ci.yml
```diff
@@ -152,6 +152,10 @@ jobs:
         if: matrix.config == 'android'
         uses: ./.github/actions/setup-android
 
+      - name: Install Windows Debug Tools
+        if: matrix.os == 'windows-2025'
```
externl: Can we do for all Windows?
```suggestion
        if: runner.os == 'Windows'
```

pepone: Updated to run on all Windows versions.

PR #4645, opened 2025-11-12
Fixes for JavaScript browser testing

### LINKED ISSUE #4644: Ice JS Browser Test Failure: `TypeError: can't access property "apply", method is undefined`
Came across this error while trying to run the browser tests locally on my Windows 11 computer.
Sounds like we're trying to apply a function which is undefined?

Here's the output I see in the browser window:
```
-! 11/11/2025, 16:45:10.717 warning: failed to dispatch createLogDirectory to 'Browser/ProcessController' over :-1<->127.0.0.1:15002
   dispatchImpl@http://127.0.0.1:8080/ice.js:8981:21
   method@http://127.0.0.1:8080/ice.js:9083:48
   defineOperations/classType.prototype.dispatch@http://127.0.0.1:8080/ice.js:9200:23
   dispatch@http://127.0.0.1:8080/ice.js:10068:34
   dispatch@http://127.0.0.1:8080/ice.js:10306:47
   dispatchAsync@http://127.0.0.1:8080/ice.js:13338:49
   dispatchAll@http://127.0.0.1:8080/ice.js:13309:34
   upcall@http://127.0.0.1:8080/ice.js:12524:22
   message@http://127.0.0.1:8080/ice.js:12493:14
   start/<@http://127.0.0.1:8080/ice.js:12078:28
   socketBytesAvailable@http://127.0.0.1:8080/ice.js:17106:22
   initialize/this._fd.onmessage@http://127.0.0.1:8080/ice.js:16889:48
   
connection with process controller registry closed
```

----

Pretty obvious problem: we forgot to implement the `createLogDirectory` operation on `ProcessControllerI`:
https://github.com/zeroc-ice/ice/blob/b0365d4b70cf528f6c8a7d64310a8526a5eab3c9/js/test/Common/ControllerI.js#L54 This type implements Slice interfaces `BrowserProcessController` and by inheritance `ProcessController`:
```slice
    interface ProcessController
    {
        ...

        idempotent string createLogDirectory(string path);
    }

    interface BrowserProcessController extends ProcessController { ... }
```
But we don't actually provide an implementation for `createLogDirectory`. So, when the test controller dispatches, it looks up `createLogDirectory`, gets back `undefined`, then tries to call `apply` on `undefined`, giving the error message I actually saw.

----

There's some larger problem here that I'm not familiar enough to see.

We should of caught this issue up-front in `defineOperations` and thrown an `OperationNotExist` exception:
https://github.com/zeroc-ice/ice/blob/b0365d4b70cf528f6c8a7d64310a8526a5eab3c9/js/src/Ice/Operation.js#L509-L511 but for some reason we don't hit this.

Instead we get all the way to the lookup, where we only have a `console.assert` which does fail, but doesn't stop execution.
https://github.com/zeroc-ice/ice/blob/b0365d4b70cf528f6c8a7d64310a8526a5eab3c9/js/src/Ice/Operation.js#L246-L247

pepone:
Fix #4644

### DIFF
--- a/js/src/Ice/Operation.js
+++ b/js/src/Ice/Operation.js
@@ -244,7 +244,9 @@ function dispatchImpl(servant, op, request) {
     // Check to make sure the servant implements the operation.
     //
     const method = servant[op.servantMethod];
-    console.assert(typeof method === "function");
+    if (typeof method !== "function") {
+        throw new OperationNotExistException();
+    }
 
     if (op.mode === OperationMode.Normal) {
         checkNonIdempotent(request.current);
--- a/js/test/Common/ControllerI.js
+++ b/js/test/Common/ControllerI.js
@@ -112,6 +112,10 @@ class ProcessControllerI extends Test.Common.BrowserProcessController {
         return "127.0.0.1";
     }
 
+    createLogDirectory() {
+        // No-op in browser environment.
+    }
+
     redirect(url, current) {
         current.con.close();
         window.location.href = url;
--- a/js/test/Common/TestHelper.js
+++ b/js/test/Common/TestHelper.js
@@ -76,7 +76,7 @@ export class TestHelper {
 
     updateLogFileProperty(properties, suffix) {
         const logFile = properties.getIceProperty("Ice.LogFile");
-        if (logFile.length > 0) {
+        if (!TestHelper.isBrowser() && logFile.length > 0) {
             const dotIndex = logFile.lastIndexOf(".");
             const newLogFile =
                 dotIndex == -1
--- a/scripts/Util.py
+++ b/scripts/Util.py
@@ -2463,7 +2463,7 @@ class RemoteProcessController(ProcessController):
         # TODO: support envs?
 
         traceProps = process.getEffectiveTraceProps(current)
-        if traceProps:
+        if traceProps and not current.config.browser:
             if "Ice.ProgramName" in props:
                 programName = props["Ice.ProgramName"]
             else:


js/src/Ice/Operation.js
```diff
@@ -244,7 +244,9 @@ function dispatchImpl(servant, op, request) {
     // Check to make sure the servant implements the operation.
     //
     const method = servant[op.servantMethod];
-    console.assert(typeof method === "function");
+    if (typeof method !== "function") {
+        throw new OperationNotExistException();
```
pepone, line 248: This catches the case the application didn't implemented an operation, the other check we already have just catches operations missing from the Slice definition, for example client/server using different Slice definitions.

js/test/Common/ControllerI.js
```diff
@@ -112,6 +112,10 @@ class ProcessControllerI extends Test.Common.BrowserProcessController {
         return "127.0.0.1";
     }
 
+    createLogDirectory() {
+        // No-op in browser environment.
```
pepone, line 116: This was missing, it will be good to port ControllerI to TypeScript to catch such errors at compile time.

js/test/Common/TestHelper.js
```diff
@@ -76,7 +76,7 @@ export class TestHelper {
 
     updateLogFileProperty(properties, suffix) {
         const logFile = properties.getIceProperty("Ice.LogFile");
-        if (logFile.length > 0) {
+        if (!TestHelper.isBrowser() && logFile.length > 0) {
```
pepone, line 79: Ice.LogFile is not supported in the browser.

scripts/Util.py
```diff
@@ -2463,7 +2463,7 @@ def start(self, process, current, args, props, envs, watchDog):
         # TODO: support envs?
 
         traceProps = process.getEffectiveTraceProps(current)
-        if traceProps:
+        if traceProps and not current.config.browser:
```
pepone, line 2466: ditto Ice.LogFile is not supported in JavaScript browser environment.

InsertCreativityHere (approved):
Looks good to me!

externl (approved):

PR #4650, opened 2025-11-12
Add support for build with Visual Studio 2026

pepone:
This PR updates Windows builds to work with Visual Studio 2026

The main issue is that the dependency packages didn't link correctly when using Visual Studio 2026 which has a new default platform toolset.

I uploaded updated version of the required dependencies to NuGet.org for this to work correctly.

I rename the packages to use `ZeroC.` reserved prefix, and removed the `.v143` suffix, as they don't require a specific toolset.

### DIFF
--- a/cpp/msbuild/ice.test.props
+++ b/cpp/msbuild/ice.test.props
@@ -9,7 +9,6 @@
 
   <ItemDefinitionGroup>
     <Link>
-      <GenerateDebugInformation>Yes</GenerateDebugInformation>
       <FullProgramDatabaseFile>false</FullProgramDatabaseFile>
       <GenerateDebugInformation>DebugFull</GenerateDebugInformation>
     </Link>


cpp/msbuild/ice.test.props
```diff
@@ -11,7 +11,7 @@
     <Link>
       <GenerateDebugInformation>Yes</GenerateDebugInformation>
       <FullProgramDatabaseFile>false</FullProgramDatabaseFile>
-      <GenerateDebugInformation>DebugFastLink</GenerateDebugInformation>
+      <GenerateDebugInformation>DebugFull</GenerateDebugInformation>
```
pepone, line 13: DebugFastLink was removed in Visual Studio 2026

externl (approved):

InsertCreativityHere (approved):
Looks good to me!

PR #4655, opened 2025-11-19
Improve Hand-Written Doc-Comments in Java (Part 1)

InsertCreativityHere:
This PR improves the doc-comments on our Java source code, drawing heavy inspiration from the current doc-comments in C++.
But it wasn't just a blind copy-paste, I fixed the tags and the terminology wherever I saw them.

Also made a few changes to the C++ doc-comments which I think were grammatically necessary, or helped clarity.

To avoid being huge and unreviewable, this PR only fixes half the Java files (those starting with the letter 'O' to the letter 'Z').

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SliceTypeId.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SliceTypeId.java
@@ -11,15 +11,15 @@ import java.lang.annotation.Target;
  * Assigns a Slice type ID to a class or to an interface.
  *
  * <p>The Slice compiler assigns Slice type IDs to the classes and interfaces it generates from Slice classes,
- * exceptions, and interfaces.
+ * exceptions, and interfaces.</p>
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.TYPE})
 public @interface SliceTypeId {
     /**
-     * Returns the Slice type ID assigned to this class or interface.
+     * The Slice type ID of this class or interface.
      *
-     * @return the Slice type ID 
+     * @return the Slice type ID of this class or interface
     */
     String value();
 }


cpp/include/Ice/Properties.h
```diff
@@ -199,8 +199,9 @@ namespace Ice
         /// mismatched, the default list is returned.  Within single quotes or double quotes, you can escape the quote
         /// in question with a backslash, e.g. O'Reilly can be written as O'Reilly, "O'Reilly" or 'O\'Reilly'.
         /// @param key The property key.
-        /// @return The property value interpreted as list of strings, or the default value if the property is not set.
-        /// @throws PropertyException If the property is not a known Ice property.
+        /// @return The property value interpreted as a list of strings, or the default value if the property is not
+        ///     set.
+        /// @throws PropertyException Thrown when the property is not a known Ice property.
```
InsertCreativityHere, line 204: We go to great effort to consistently start `@throws` comments with `Thrown when ...` in C++.
I updated the few `@throws` I found which were not following this apparent convention.

cpp/include/Ice/Properties.h
```diff
@@ -199,8 +199,9 @@ namespace Ice
         /// mismatched, the default list is returned.  Within single quotes or double quotes, you can escape the quote
         /// in question with a backslash, e.g. O'Reilly can be written as O'Reilly, "O'Reilly" or 'O\'Reilly'.
         /// @param key The property key.
-        /// @return The property value interpreted as list of strings, or the default value if the property is not set.
-        /// @throws PropertyException If the property is not a known Ice property.
+        /// @return The property value interpreted as a list of strings, or the default value if the property is not
```
InsertCreativityHere, line 202: Changed `interpreted as list` to `interpreted as a list` for grammar.

cpp/include/Ice/Properties.h
```diff
@@ -210,13 +211,14 @@ namespace Ice
...
-        /// @return The property value interpreted as list of strings, or the default value if the property is not set.
+        /// @return The property value interpreted as a list of strings, or the default value if the property is not
+        ///     set.
         /// @see #setProperty
         StringSeq getPropertyAsListWithDefault(std::string_view key, const StringSeq& value);
 
-        /// Gets all properties whose keys begins with @p prefix. If @p prefix is the empty string, then all properties
+        /// Gets all properties whose keys begin with @p prefix. If @p prefix is the empty string, then all properties
         /// are returned.
-        /// @param prefix The prefix to search for (empty string if none).
```
InsertCreativityHere, line 219: I find `(empty string if none)` vague and not helpful here.
We already spell it out much more clearly in the summary.

cpp/include/Ice/ServantLocator.h
```diff
@@ -38,7 +38,7 @@ namespace Ice
         /// @param cookie The cookie that was returned by #locate.
         virtual void finished(const Current& curr, const ObjectPtr& servant, const std::shared_ptr<void>& cookie) = 0;
 
-        /// Notifies this servant locator that the object adapter is being deactivated.
+        /// Notifies this servant locator that the object adapter in which it's installed is being deactivated.
```
InsertCreativityHere, line 41: In Java we explained _which_ object adapter we're talking about.
I think this is more clear, so I added it to the C++ comment to match.

cpp/include/Ice/UserException.h
```diff
@@ -10,7 +10,7 @@ namespace Ice
     class InputStream;
     class OutputStream;
 
-    /// Abstract base class for all Ice exceptions defined in Slice.
```
InsertCreativityHere, line 13: No need to qualify this; `UserException` is just the base for all Slice-defined exceptions, not specific to `Ice` exceptions.

cpp/src/Ice/Properties.cpp
```diff
@@ -20,7 +20,7 @@ using namespace IceInternal;
 
 namespace
 {
-    /// Find the Ice property array for a given property name.
```
InsertCreativityHere, line 23: We prefer the "simple present tense" for our doc-comments.
Which is a fancy way of saying verbs should end with `s` most of the time.

cpp/include/Ice/Value.h
```diff
@@ -33,16 +33,16 @@ namespace Ice
         Value& operator=(const Value&) = delete;
         Value& operator=(Value&&) = delete;
```
InsertCreativityHere, line 35: These are the only doc-comments I changed substantially; I found the Java comments much better.

On their own, these functions don't do anything, they're no-op.
They certainly don't: `Validates or updates the fields of this object before marshaling.`

These functions exist to be overridden by subclasses, so I think we should state that clearly in our documentation.

cpp/src/Ice/Properties.cpp
```diff
@@ -49,10 +49,10 @@ namespace
         return nullptr;
     }
 
-    /// Find the default value for an Ice property. If there is no default value, return an empty string.
```
InsertCreativityHere, line 52: For all the other property functions (and most files), we mention these 'special cases' in the `@returns`, not the summary.

For example:
https://github.com/zeroc-ice/ice/blob/3de7c3508ec261328e0f2064d8bf03c5f3ef8fd1/cpp/include/Ice/Properties.h#L145-L156

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/ConnectionInfo.java
```diff
@@ -15,7 +15,6 @@ public final class ConnectionInfo extends com.zeroc.Ice.ConnectionInfo {
     /** The certificate chain verification status. */
     public final boolean verified;
 
-    // internal constructor
```
InsertCreativityHere, line 18: I think these are leftover from when we used to have `Ice` and `IceInternal` in separate packages.
Now that they're all in the same package, it's obvious that these constructors are 'internal' (they're not marked `public`). I searched for all these comments and removed them all.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InputStream.java
```diff
@@ -1271,7 +1271,7 @@ public int readEnum(int maxValue) {
      *
      * @param <T> The value type.
      * @param cb The consumer to notify when the extracted instance is available. The stream
-     *     extracts Slice values in stages. The Ice run time calls accept on the consumer when the
```
InsertCreativityHere, line 1274: Also searched for everywhere we used `Ice run time` and replaced them with `Ice runtime`, which is more correct. Note: we were already using `runtime` (one word) in nearly everywhere else.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/PluginInitializationException.java
```diff
@@ -2,7 +2,7 @@
 
 package com.zeroc.Ice;
 
-/** This exception indicates that a failure occurred while initializing a plug-in. */
+/** The exception that is thrown when the initialization of an Ice plugin fails. */
```
InsertCreativityHere, line 5: By convention we start 90% of our exception doc-comments with:
`The exception that is thrown when...`.

I fixed the handful of exceptions that were not following this pattern.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/UnknownException.java
```diff
@@ -40,7 +40,6 @@ public String ice_id() {
         return "::Ice::UnknownException";
     }
 
-    // Logically protected too.
```
InsertCreativityHere, line 43: Also came across a couple of these comments and don't understand why they're here.
These constructors are package-private, I don't know what "logically protected" means.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SliceInfo.java
```diff
@@ -23,19 +23,25 @@ public final class SliceInfo {
...
         this.compactId = compactId;
         this.bytes = bytes;
         this.hasOptionalMembers = hasOptionalMembers;
         this.isLastSlice = isLastSlice;
+
+        if (compactId == -1) {
+            assert (!typeId.isEmpty());
+        } else {
+            assert (typeId.isEmpty());
+        }
```
InsertCreativityHere, line 45: This is one place that's unrelated to the rest of the PR.
We had this check in C++ and there's no reason to also add an `assert` check here in Java.

cpp/src/slice2java/Gen.cpp
```diff
@@ -914,7 +914,7 @@ Slice::JavaVisitor::writeResultTypeMarshalUnmarshalCode(
     const string& package,
     const string& streamName,
     const string& paramPrefix,
```
InsertCreativityHere, line 916: Marshaling was spelt wrong here.
It should have 1 `l`, not 2.
I `rg`ed the entire repo and fixed all the instances.

java/src/com.zeroc.glacier2/src/main/java/com/zeroc/Glacier2/package-info.java
```diff
@@ -1,4 +1,4 @@
 // Copyright (c) ZeroC, Inc.
 
-/** APIs for clients of the Glacier2 firewall traversal service. */
+/** Client APIs to communicate through firewalls and across NATs with Glacier2. */
```
bernardnormier: Is describing a package as "APIs" a Java convention?

Also, the emphasis should be on Glacier2, like "APIs to communicate with the Glacier2 service". And these are not only client APIs.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/OutputStreamWrapper.java
```diff
@@ -5,19 +5,18 @@
...
-// big the sequence that is eventually written will be. To avoid excessive
-// data copying, this class mantains a private _bytes array of 254 bytes and,
-// initially, writes data into that array. If more than 254 bytes end up being
-// written, we write a dummy sequence size of 255 (which occupies five bytes
-// on the wire) into the stream and, once this stream is closed, patch
-// that size to match the actual size. Otherwise, if the _bytes buffer contains
-// fewer than 255 bytes when this stream is closed, we write the sequence size
-// as a single byte, followed by the contents of the _bytes buffer.
+/**
+ * Wraps one of our {@link com.zeroc.Ice.OutputStream}s in a way that provides the API of {@link java.io.OutputStream},
```
bernardnormier: I would rewrite this description as follows:

Decorates OuputStream to marshal an arbitrary Java class into a sequence of bytes.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SliceInfo.java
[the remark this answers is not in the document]
InsertCreativityHere: It's not that critical. `assert` is fine here.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Plugin.java
```diff
@@ -3,14 +3,14 @@
 package com.zeroc.Ice;
 
 /**
- * A communicator plug-in. A plug-in generally adds a feature to a communicator, such as support for
- * a protocol. The communicator loads its plug-ins in two stages: the first stage creates the plug-ins,
- * and the second stage invokes {@link Plugin#initialize} on each one.
+ * Represents a communicator plug-in. A plug-in generally adds a feature to a communicator,
```
pepone, line 6: we use plugin one word below. I think plugin is the preferred but we have a mix usage in all languages. I guess is better to fix in a separate PR.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
```diff
@@ -83,48 +86,62 @@ public Properties(String[] args, Properties defaults) {
...
     public Properties(String[] args, Properties defaults, List<String> remainingArgs) {
         this(defaults);
         loadArgs(args, remainingArgs);
     }
 
     /**
+     * Constructs a property set with additional opt-in prefixes.
+     *
+     * @param optInPrefixes -
+     *
```
pepone, line 111: do we need both `@param` and `@hidden`?

cpp/src/Ice/PropertyUtil.h
```diff
@@ -21,7 +21,7 @@ namespace IceInternal
     /// @param prefix The prefix to validate.
     /// @param properties The properties to consider.
     /// @param propertyArray The property array to search against.
-    /// @throws PropertyException if unknown properties are found.
+    /// @throws PropertyException Thrown when an unknown property is found.
```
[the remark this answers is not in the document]
InsertCreativityHere: You are comparing Java code to C++ code fallaciously.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Plugin.java
InsertCreativityHere: Yes, we use a random mix of spellings... And it wasn't clear which we prefer like it was with `run-time` for example.
Will fix in a follow-up PR.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
InsertCreativityHere: Unfortunately yes. I just had `hidden` originally, but then it complained.
Hence why I added these 'placeholder' `@param` tags... A little lame, but that's just how it is.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/OutputStreamWrapper.java
InsertCreativityHere: Replaced `arbitrary Java class` with `serializable Java class`, but otherwise rewrote as suggested.

java/src/com.zeroc.glacier2/src/main/java/com/zeroc/Glacier2/package-info.java
InsertCreativityHere: > Is describing a package as "APIs" a Java convention?

I don't think there is any real convention for how to document packages in Java, but talking about it in terms of `API` doesn't strike me as uncommon either. I do think it's weird that we say `APIs` plural instead of `API`.
For example, gRPC uses `API ` singular: https://grpc.github.io/grpc-java/javadoc/

Issue #4654, opened 2025-11-18
`java/Ice/udp` test failure on Android with IPv6


InsertCreativityHere:
Get this failure when running the Android tests on-device with `--all` enabled:
```
*** [40/40] Running java/Ice/udp tests ***
[ running client/server test - 11/18/25 14:56:45 ]
- Config: ssl,android,usb
testing udp... Expected 1 replies, got 0
ok
testing udp multicast... ok
testing udp bi-dir connection... ok
[ running client/server test - 11/18/25 14:56:46 ]
- Config: ws,compress,ipv6,serialize,mx,android,usb
testing udp... ok
testing udp multicast... Expected 1 replies, got 0
Expected 1 replies, got 0
Expected 1 replies, got 0
Expected 1 replies, got 0
Expected 1 replies, got 0
java.lang.RuntimeException
        at test.Ice.udp.AllTests.test(AllTests.java:23)
        at test.Ice.udp.AllTests.allTests(AllTests.java:189)
        at test.Ice.udp.Client.run(Client.java:22)
        at com.zeroc.testcontroller.ControllerApp$ControllerHelperI.run(ControllerApp.java:244)
unexpected exit status: expected: 0, got -1


test in Ice/udp failed:
unexpected exit status: expected: 0, got -1


Ran 40 tests in 5.0 minutes 48.57 seconds
39 succeeded and 1 failed:
- Ice/udp
  [ running client/server test - 11/18/25 14:56:46 ]
  - Config: ws,compress,ipv6,serialize,mx,android,usb
  testing udp... ok
  testing udp multicast... Expected 1 replies, got 0
  [...]
  java.lang.RuntimeException
        at test.Ice.udp.AllTests.test(AllTests.java:23)
        at test.Ice.udp.AllTests.allTests(AllTests.java:189)
        at test.Ice.udp.Client.run(Client.java:22)
        at com.zeroc.testcontroller.ControllerApp$ControllerHelperI.run(ControllerApp.java:244)

  test in Ice/udp failed:
  unexpected exit status: expected: 0, got -1
allTests.py: unexpected exception raised:
Traceback (most recent call last):
  File "D:\Code\Workspace\ice\java\..\scripts\Util.py", line 4323, in runTests
    sys.exit(driver.run(mappings, args))
SystemExit: 1
```
Works just fine with IPv4, but not IPv6.

----

This is apparently a previously known issue, last touched in this PR: https://github.com/zeroc-ice/ice/pull/4568
[...]

PR #4656, opened 2025-11-19
Use Port-Forarding to Work Around Broken Android Test Controller

### LINKED ISSUE #4637: Android Test Suite Doesn't work On-Device
After fixing #4587 locally, I ran the Android tests in the emulator, and tried to run them on my Samsung Galaxy A25, but the test-controller for android returns unusable proxies, and the tests fail.

The problem is the published host for proxies returned by `ControllerApp`.
If we aren't running the emulator, we don't set `PublishedHost` via properties, meaning we compute one with this call:
```
publishedHost = InetAddress.getLocalHost().getHostName();
```
On MacOS and Windows, this returns a proper DNS hostname, but on android this always returns `localhost`.
This happens both on device and with the emulator, but this doesn't affect the emulator, since we set an explicit `PublishedHost` in `initData`.

----

Dumping my phone's network interfaces suggests that `getLocalHost` on Android always uses the `lo` interface, unlike other platforms:
```
INTERFACE 0: [dummy0] isLoopback=NO isVirtual=NO isUp=YES
    ADDR 0: fe80::e09b:bfff:fe80:44ed%dummy0
        isLoopback:  NO
        isSiteLocal: NO
        isLinkLocal: YES

INTERFACE 1: [wlan0] isLoopback=NO isVirtual=NO isUp=YES
    ADDR 0: [fe80::d0ef:ccff:fe8a:9db0%wlan0]
        isLoopback:  NO
        isSiteLocal: NO
        isLinkLocal: YES
    ADDR 1: [fd9d:1fff:250b:c846:d0ef:ccff:fe8a:9db0]
        isLoopback:  NO
        isSiteLocal: NO
        isLinkLocal: NO
    ADDR 2: [fd9d:1fff:250b:c846:2481:a77:c863:e8a3]
        isLoopback:  NO
        isSiteLocal: NO
        isLinkLocal: NO
    ADDR 3: [Austins-Galaxy-A25.jupiter.zeroc.com]    << What we want
        isLoopback:  NO
        isSiteLocal: YES
        isLinkLocal: NO

INTERFACE 2: [lo] isLoopback=YES isVirtual=NO isUp=YES
    ADDR 0: [ip6-localhost]
        isLoopback:  YES
        isSiteLocal: NO
        isLinkLocal: NO
    ADDR 1: [localhost]                 << What is actually published
        isLoopback:  YES
        isSiteLocal: NO
        isLinkLocal: NO
```
Sanity check: adding `initData.properties.setProperty("ControllerAdapter.PublishedHost", "Austins-Galaxy-A25")` to the on-device test-controller does fix this problem, and the tests all work (except Ice/udp, but whatever).

This isn't a quirk of Java, checking the device's hostname via `abd shell hostname` also returns `localhost`.
Looking online, some people said this is expected, but other people got reasonable DNS names running the same command. So maybe this is manufacturer dependent?

----

It's possible to fix this in general by looking through all the available INetAddresses and choosing the sane one.
But it'd be nice to have a larger sample size than `1` before making this change:
```java
// ControllerApp.java

                // Find the first non-loopback site-local INetAddress available for this device.
                //
                // We can't use `.networkInterfaces` or `.inetAddresses` because they are not available on Android.
                InetAddress hostAddress = null;
                try {
interfaceLoop:      for (var interfaces = NetworkInterface.getNetworkInterfaces(); interfaces.hasMoreElements();) {
                        var intf = interfaces.nextElement();
                        if (intf.isUp() && !intf.isLoopback()) {
                            for (var addresses = intf.getInetAddresses(); addresses.hasMoreElements();) {
                                var addr = addresses.nextElement();
                                if (addr.isSiteLocalAddress()) {
                                    hostAddress = addr;
                                    break interfaceLoop;
                                }
                            }
                        }
                    }
                } catch (SocketException ignored) {}
                if (hostAddress != null) {
                    initData.properties.setProperty("ControllerAdapter.PublishedHost", hostAddress.getHostName());
                }
```

InsertCreativityHere:
This PR fixes #4637 for our test suite, but not in general.
For our purposes we can just use port-forwarding for on-device testing (same as we do for the emulator),

For users in the real-world, running servers on Android devices, they will need to set `PublishedHost` in order for their server to return functioning proxies.

----

I tested this PR on-device and with the emulator and everything worked fine (except for IPv6 with the `Ice/udp` test, see #4654).

### DIFF
--- a/java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerApp.java
+++ b/java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerApp.java
@@ -15,6 +15,7 @@ import com.zeroc.Ice.Time;
 
 import java.io.File;
 import java.io.Writer;
+import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 
@@ -178,7 +179,7 @@ public class ControllerApp extends Application {
             initData.properties.setProperty("ControllerAdapter.PublishedHost", "127.0.0.1");
 
             if (bluetooth) {
-                initData.properties.setProperty("Ice.Plugin.IceBT", "com.zeroc.IceBT.PluginFactory");
+                initData.pluginFactories = Collections.singletonList(new com.zeroc.IceBT.PluginFactory());
             }
 
             _communicator = new com.zeroc.Ice.Communicator(initData);
--- a/java/test/android/controller/build.gradle
+++ b/java/test/android/controller/build.gradle
@@ -99,8 +99,7 @@ clean {
 dependencies {
     implementation files("${rootProject.projectDir}/../../../lib/glacier2-${project.iceVersion}.jar")
     implementation files("${rootProject.projectDir}/../../../lib/ice-${project.iceVersion}.jar")
-    runtimeOnly files("${rootProject.projectDir}/../../../lib/icediscovery-${project.iceVersion}.jar")
-    runtimeOnly files("${rootProject.projectDir}/../../../lib/icebt-${project.iceVersion}.jar")
+    implementation files("${rootProject.projectDir}/../../../lib/icebt-${project.iceVersion}.jar")
 
     implementation files("${rootProject.projectDir}/../../../lib/test.jar")
     runtimeOnly "org.apache.commons:commons-compress:1.20"


java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerApp.java
```diff
@@ -173,16 +173,13 @@ public ControllerI(boolean bluetooth) {
...
-                // We use a hardcoded port that's forwarded to the host machine by the emulator.
-                initData.properties.setProperty("ControllerAdapter.Endpoints", "tcp -p 15001");
-                initData.properties.setProperty("ControllerAdapter.PublishedHost", "127.0.0.1");
-            }
-            else
+            // We use a hardcoded port that is forwarded to the device / emulator with `adb forward`.
+            initData.properties.setProperty("ControllerAdapter.Endpoints", "tcp -p 15001");
+            initData.properties.setProperty("ControllerAdapter.PublishedHost", "127.0.0.1");
+
+            if (!isEmulator())
```
externl: I don't think there's any reason to have discovery or AdapterID  configured now.

java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerApp.java
```diff
@@ -173,21 +173,13 @@ public ControllerI(boolean bluetooth) {
...
-                initData.properties.setProperty("ControllerAdapter.Endpoints", "tcp -p 15001");
-                initData.properties.setProperty("ControllerAdapter.PublishedHost", "127.0.0.1");
-            }
-            else
+            // We use a hardcoded port that is forwarded to the device / emulator with `adb forward`.
+            initData.properties.setProperty("ControllerAdapter.Endpoints", "tcp -p 15001");
+            initData.properties.setProperty("ControllerAdapter.PublishedHost", "127.0.0.1");
+
+            if (bluetooth)
             {
```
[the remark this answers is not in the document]
externl: We should probably be running rewrite on the android controller too!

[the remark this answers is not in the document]
InsertCreativityHere: Yep, looking into it now!

java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerApp.java
```diff
@@ -173,21 +173,12 @@ public ControllerI(boolean bluetooth) {
...
-                    initData.properties.setProperty("Ice.Plugin.IceBT", "com.zeroc.IceBT.PluginFactory");
-                }
-                initData.properties.setProperty("Ice.Plugin.IceDiscovery", "com.zeroc.IceDiscovery.PluginFactory");
-                initData.properties.setProperty("IceDiscovery.DomainId", "TestController");
+            // We use a hardcoded port that is forwarded to the device / emulator with `adb forward`.
+            initData.properties.setProperty("ControllerAdapter.Endpoints", "tcp -p 15001");
+            initData.properties.setProperty("ControllerAdapter.PublishedHost", "127.0.0.1");
+
+            if (bluetooth) {
+                initData.properties.setProperty("Ice.Plugin.IceBT", "com.zeroc.IceBT.PluginFactory");
```
pepone: We should use the new plugin factory on initData.

pepone (approved):

java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerApp.java
InsertCreativityHere: Fixed!

PR #4659, opened 2025-11-21
Add default for Ice.ProgramName in C++ and C++ based languages

externl:
Part of #4565

I originally used the term fallback. However, I feel that _default_ is more correct as it's what we say about other properties.

In the case where we can't determine the program name, like in MATLAB, or we're running from the Python REPL, then we use `IceLANG`. Please let me know if you prefer something different. 

However, I wonder in general if this behavior is desirable, since this default can be quite long and could pollute the log file. Maybe we should leave the property empty unless explicitly set. Right now there's no way to disable or opt-out of this new behavior.

### DIFF
--- a/cpp/src/Ice/Instance.cpp
+++ b/cpp/src/Ice/Instance.cpp
@@ -124,7 +124,7 @@ namespace
 #elif defined(_WIN32)
         vector<char> buffer(MAX_PATH);
         DWORD len = GetModuleFileNameA(NULL, buffer.data(), static_cast<DWORD>(buffer.size()));
-        if (len > 0 && len < buffer.size())
+        if (len > 0)
         {
             programName = string{buffer.data(), static_cast<size_t>(len)};
         }


CHANGELOG-3.8.md
```diff
@@ -727,6 +727,14 @@ See `InitializationData::pluginFactories`.
   There is no clean way to recover from these errors, and keeping malfunctioning systems running is of no help;
   aborting the application allows the replica management system to take over in a replicated setup.
 
+- Added a default value for the `Ice.ProgramName` property. The property is now set to:
+  - The first argument of the argv argument passed to `Ice::initialize`.
```
bernardnormier: This is not new behavior.

Also, Otherwise below is unclear: there is no "if" in the first point.

CHANGELOG-3.8.md
```diff
@@ -1042,6 +1054,10 @@ initialization. See `InitializationData.pluginFactories`.
 - Removed the self-reference parameter from `InvocationFuture.add_sent_callback`. Registered callbacks no longer
   receive the `Ice.InvocationFuture` instance as their first parameter.
 
+- Added a default value for the `Ice.ProgramName` property. The first value in `sys.argv` will be used if available;
+  otherwise, the default value is `IcePy`.  It used to be the first element in the args cell given
```
bernardnormier: Since space after period.

CHANGELOG-3.8.md
```diff
@@ -1085,6 +1101,10 @@ initialization. See `InitializationData.pluginFactories`.
 
 - Added a SwiftPM plugin, `CompileSlice`, that lets you compile Slice files as part of SwiftPM and Xcode builds.
 
+- Added a default value for the `Ice.ProgramName` property. The first value in `CommandLine.arguments` will be used
```
bernardnormier: I find the "if available" unclear. You mean if CommandLine.arguments is not empty?
And that's all the time or only when you call a specific Ice.initialize?

CHANGELOG-3.8.md
```diff
@@ -923,6 +931,10 @@ initialization. See `InitializationData.pluginFactories`.
   `ice_postunmarshal` only makes sense if the application can register a custom Slice loader and Ice for PHP does not
   provide custom Slice loaders.
 
+- Added a default value for the `Ice.ProgramName` property. The `SCRIPT_FILENAME` server variable will be used if
+  available; otherwise, the default value is `IcePHP`. It used to be the first element in the args cell given
```
bernardnormier: "args cell" is MATLAB terminology.

cpp/src/Ice/Instance.cpp
```diff
@@ -103,6 +104,36 @@ namespace
...
+        vector<char> buffer(PATH_MAX);
+        ssize_t len = readlink(procExePath.c_str(), buffer.data(), buffer.size());
+        if (len > 0)
+        {
+            programName = string{buffer.data(), static_cast<size_t>(len)};
+        }
+#elif defined(_WIN32)
+        vector<char> buffer(MAX_PATH);
+        DWORD len = GetModuleFileNameA(NULL, buffer.data(), static_cast<DWORD>(buffer.size()));
+        if (len > 0 && len < buffer.size())
```
bernardnormier: From the API reference:
https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamea

I don't see how the returned len can be > MAX_PATH. So what's special about len = MAX_PATH that you want to silently fail?

cpp/src/Ice/Instance.cpp
```diff
@@ -103,6 +104,36 @@ namespace
...
+        {
+            programName = string{buffer.data(), len};
+        }
+#endif
+        return programName;
```
bernardnormier, line 137: I thought you wanted to keep only the string after the last "/" or "\"?

cpp/test/Ice/properties/Client.cpp
```diff
@@ -238,6 +239,27 @@ Client::run(int, char**)
         test(properties->getIceProperty("Ice.Trace.Network") == "3");
         cout << "ok" << endl;
     }
+
+    {
+        cout << "testing Ice.ProgramName fallback... " << flush;
```
bernardnormier: We should use a consistent terminology.

matlab/test/Ice/properties/Client.m
```diff
@@ -72,4 +72,10 @@ function client(args)
         assert(isa(ex, 'Ice.PropertyException'));
     end
     fprintf('ok\n');
+
+    fprintf('testing Ice.ProgramName fallback... ');
```
bernardnormier: default

php/test/Ice/properties/Client.php
```diff
@@ -85,5 +85,10 @@ function run($args)
         test($communicator->getProperties()->getIceProperty("Ice.Trace.Network") == "3");
         test(count($args) == 2);
         echo "ok\n";
+
+        echo "testing Ice.ProgramName fallback...";
```
bernardnormier: default

python/test/Ice/properties/Client.py
```diff
@@ -121,3 +121,12 @@ def run(self, args: list[str]) -> None:
         except Ice.PropertyException:
             pass
         print("ok")
+
+        sys.stdout.write("testing Ice.ProgramName fallback... ")
```
bernardnormier: default

ruby/src/IceRuby/Properties.cpp
```diff
@@ -46,13 +46,6 @@ IceRuby_createProperties(int argc, VALUE* argv, VALUE /*self*/)
             defaults = getProperties(argv[1]);
         }
 
-        //
-        // Insert the program name (stored in the Ruby global variable $0) as the first
```
bernardnormier, line 50: How does this work now in Ruby?

bernardnormier (commented):
The CHANGELOG entries need more work.

CHANGELOG-3.8.md
externl: > You mean if CommandLine.arguments is not empty?
Yes
> And that's all the time or only when you call a specific Ice.initialize?
All of the time

cpp/src/Ice/Instance.cpp
externl: Yes, you're right. I think I just misread the documentation.

cpp/test/Ice/properties/Client.cpp
externl: I switched form fallback to default and forgot to update the tests.

cpp/src/Ice/Instance.cpp
```diff
@@ -103,6 +104,36 @@ namespace
...
+        string procExePath{"/proc/self/exe"};
+        vector<char> buffer(PATH_MAX);
+        ssize_t len = readlink(procExePath.c_str(), buffer.data(), buffer.size());
+        if (len > 0)
+        {
+            programName = string{buffer.data(), static_cast<size_t>(len)};
+        }
+#elif defined(_WIN32)
+        vector<char> buffer(MAX_PATH);
+        DWORD len = GetModuleFileNameA(NULL, buffer.data(), static_cast<DWORD>(buffer.size()));
```
pepone, line 126: Is it correct to use `GetModuleFileNameA` which returns a string in the application native encoding, not necessarily UTF-8.

ruby/src/IceRuby/Properties.cpp
externl: We were already setting Ice.ProgramName in ruby. I just the setting from properties construction.

In `ruby/src/IceRuby/Communicator.cpp`
```
 // Insert the program name (stored in the Ruby global variable $0) unless already set.
        if (initData.properties->getProperty("Ice.ProgramName").empty())
        {
            volatile VALUE progName = callRuby(rb_gv_get, "$0");
            initData.properties->setProperty("Ice.ProgramName", getString(progName));
        }
```

cpp/src/Ice/Instance.cpp
```diff
@@ -103,6 +104,36 @@ namespace
         }
     }
 
+    string getProgramName()
+    {
+        string programName{"IceC++"};
+#if defined(__APPLE__) && TARGET_OS_IPHONE == 0
+        vector<char> buffer(PATH_MAX);
+        int len = proc_pidpath(getpid(), buffer.data(), static_cast<uint32_t>(buffer.size()));
```
pepone, line 112: This not a macOS documented API.

cpp/src/Ice/Instance.cpp
externl: I'm on the fence for this. Any preference?

cpp/src/Ice/Instance.cpp
```diff
@@ -103,6 +104,36 @@ namespace
...
+        if (len > 0)
+        {
+            programName = string{buffer.data(), static_cast<size_t>(len)};
+        }
+#elif defined(__linux__)
+        // Read the command line from proc
+        string procExePath{"/proc/self/exe"};
```
pepone: Why don't declare this as a `const char*`?

cpp/src/Ice/Instance.cpp
pepone: using the full path sounds too much.

cpp/src/Ice/Instance.cpp
externl: Maybe we should use the unicode variant `GetModuleFileNameW`, then convert to ute-8?

php/src/Communicator.cpp
```diff
@@ -911,6 +911,28 @@ ZEND_FUNCTION(Ice_initialize)
...
+        // Try to get the script filename from PHP's global variables
+        zval* scriptName = zend_hash_str_find(&EG(symbol_table), "_SERVER", sizeof("_SERVER") - 1);
+        string programName{"IcePHP"};
+        if (scriptName && Z_TYPE_P(scriptName) == IS_ARRAY)
+        {
+            zval* phpSelf =
+                zend_hash_str_find(Z_ARRVAL_P(scriptName), "SCRIPT_FILENAME", sizeof("SCRIPT_FILENAME") - 1);
+            if (phpSelf && Z_TYPE_P(phpSelf) == IS_STRING)
+            {
+                string name = string{Z_STRVAL_P(phpSelf), Z_STRLEN_P(phpSelf)};
```
pepone:
```suggestion
                string name{Z_STRVAL_P(phpSelf), Z_STRLEN_P(phpSelf)};
```

php/src/Communicator.cpp
```diff
@@ -911,6 +911,28 @@ ZEND_FUNCTION(Ice_initialize)
         initData.properties = std::make_shared<Ice::Properties>();
     }
 
+    if (initData.properties->getIceProperty("Ice.ProgramName").empty())
```
pepone, line 914: I find this to complicated and the script name is probably not specially good anyway. main.php or client.php, I think just use "IcePHP" as the default should be simpler and sufficient. Same for Python.

cpp/src/Ice/Instance.cpp
externl: It is not, but it seems to be what everyone uses.

php/src/Communicator.cpp
externl: That can be said for C++, `client`, `server`, etc. It seems to work just fine.

cpp/src/Ice/Instance.cpp
externl: Actually, I think `GetModuleFileNameA` is correct. We want the native encoding.

php/src/Communicator.cpp
pepone: For PHP what would this typically be `index.php`?  

I think we are going out of the way to compute the default value. Just using IceXXX seems a pretty good default without making things more complicated than required, and users can easily set this to something meaningful for the application using Ice.ProgramName.

cpp/src/Ice/Instance.cpp
pepone: Aren't property files UTF-8 encoded? We pass this value to setProperty which doesn't do any encoding conversions.

php/src/Communicator.cpp
externl: I don't think the logger prefix is very useful to be setting automatically in general. I would prefer to not set anything by default. 

However we want to maintain some  backwards compatibility. We used to pass the command line args to get the program name. This essentially keeps the same effect. I find the script name better than IceXXX.

cpp/src/Ice/Instance.cpp
bernardnormier: Property files are (are supposed to be) UTF-8 encoded, but properties in memory are not. Here the programName is a just a property in memory, we don't write it to a file.

CHANGELOG-3.8.md
```diff
@@ -380,6 +380,10 @@ classDiagram
 - Added `getIceProperty`, `getIcePropertyAsInt`, and `getIcePropertyAsList` methods to `Properties`. These methods
   return the value of an Ice property, or the default value if the property is not set.
 
+- The `Ice.ProgramName` property is now always set during communicator initialization, either to the executable/script
+  name when available, or `Ice<LANGUAGE>` otherwise. Previously, it was set to the first argument the args array
```
bernardnormier: to the first argument _of_ the args array

cpp/src/Ice/Instance.cpp
pepone: You are right, parseLine does the conversion to native.

CHANGELOG-3.8.md
bernardnormier: I find this entry confusing. In particular, we still use `argv[0]` in C++ when initializing a communicator with `initialize(argc, argv)`.  Reading your description, it sounds like we don't.

python/modules/IcePy/Communicator.cpp
```diff
@@ -171,6 +171,45 @@ communicatorInit(CommunicatorObject* self, PyObject* args, PyObject* /*kwds*/)
...
 
+    // Set Ice.ProgramName if not already set.
+    if (initData.properties->getIceProperty("Ice.ProgramName").empty())
+    {
+        string programName{"IcePy"};
+        // Get python program name using python c api
+        PyObject* sysModule = PyImport_ImportModule("sys");
+        if (sysModule)
+        {
+            PyObject* argv = PyObject_GetAttrString(sysModule, "argv");
```
pepone: use `PyObjectHandle argv{PyObject_GetAttrString(sysModule, "argv")}`

and no need for `Py_XDECREF(argv);`

and `argv.get()` when you want the `PyObject*` that is what we do in most places.

python/modules/IcePy/Communicator.cpp
```diff
@@ -171,6 +171,45 @@ communicatorInit(CommunicatorObject* self, PyObject* args, PyObject* /*kwds*/)
         return -1;
     }
 
+    // Set Ice.ProgramName if not already set.
+    if (initData.properties->getIceProperty("Ice.ProgramName").empty())
+    {
+        string programName{"IcePy"};
+        // Get python program name using python c api
+        PyObject* sysModule = PyImport_ImportModule("sys");
```
pepone: Here too we should use "PyObjectHandle"

php/src/Communicator.cpp
```diff
@@ -911,6 +911,33 @@ ZEND_FUNCTION(Ice_initialize)
...
+    {
+        // Try to get the script filename from PHP's global variables
+        zval* scriptName = zend_hash_str_find(&EG(symbol_table), "_SERVER", sizeof("_SERVER") - 1);
```
pepone: `scriptName` doesn't seems a good name for this variable, this is the array containing all entries with info about the execution environment. What about `scriptInfo`

php/src/Communicator.cpp
```diff
@@ -911,6 +911,33 @@ ZEND_FUNCTION(Ice_initialize)
...
+        string programName{"IcePHP"};
+        if (scriptName && Z_TYPE_P(scriptName) == IS_ARRAY)
+        {
+            zval* phpSelf =
```
pepone: I will call this `phpScriptName`

CHANGELOG-3.8.md
externl: Since this is in the general section I was trying to not be too specific with the behavior. It's only in C++ where we take the first argument of the args array unmodified. In the other languages we strip the path. I can add a clause describing the C++ behavior.

python/modules/IcePy/Communicator.cpp
```diff
@@ -171,6 +171,43 @@ communicatorInit(CommunicatorObject* self, PyObject* args, PyObject* /*kwds*/)
...
+    // Set Ice.ProgramName if not already set.
+    if (initData.properties->getIceProperty("Ice.ProgramName").empty())
+    {
+        string programName{"IcePy"};
+        // Get python program name using python c api
+        PyObjectHandle sysModule{PyImport_ImportModule("sys")};
+        if (sysModule.get())
+        {
+            PyObjectHandle argv{PyObject_GetAttrString(sysModule.get(), "argv")};
+            if (argv.get() && PyList_Check(argv.get()) && PyList_Size(argv.get()) > 0)
```
pepone: `get` is not required here for testing is set, the handler provides `operator bool`

```suggestion
            if (argv && PyList_Check(argv.get()) && PyList_Size(argv.get()) > 0)
```

CHANGELOG-3.8.md
```diff
@@ -380,6 +380,12 @@ classDiagram
 - Added `getIceProperty`, `getIcePropertyAsInt`, and `getIcePropertyAsList` methods to `Properties`. These methods
   return the value of an Ice property, or the default value if the property is not set.
 
+- The `Ice.ProgramName` property is now always set during communicator initialization, either to the executable/script
+  name when available, or `Ice<LANGUAGE>` otherwise. Previously, `Ice.ProgramName` was only set when the args array
+  was passed to `createProperties`, and was set to the first element of that array.
+
+  In C++, the first element of the `argv` array is still used if passed to `initialize`.
```
pepone, line 387: Isn't this also true for C#, Java?

externl: Currently, it is not.

cpp/src/Ice/Instance.cpp
bernardnormier: What's the point for this `len < buffer.size()` since it's always true?

bernardnormier (approved):

PR #4662, opened 2025-11-26
Upgrade C# solution files to SLNX format

pepone:


csharp/msbuild/ice.dist.slnx
```diff
@@ -0,0 +1,17 @@
+<Solution>
+  <Configurations>
+    <Platform Name="Any CPU" />
+    <Platform Name="x64" />
```
bernardnormier: I understand the Any CPU platform but not x64 and x86. If they are needed, can you add a comment?

csharp/msbuild/ice.slnx
```diff
@@ -0,0 +1,263 @@
+<Solution>
+  <Configurations>
+    <Platform Name="Any CPU" />
+    <Platform Name="x64" />
```
bernardnormier: Same here

csharp/msbuild/ice.dist.slnx
pepone: We can probably remove them, they were added by the converter. I will check.

PR #4663, opened 2025-11-26
Update the Remaining Java Doc-Comments

InsertCreativityHere:
This PR follows #4655 and finishes updating the remaining Java doc-comments we have to:
A) be consistent with the comments we have in C++
B) be more clear and correctly formatted with regard to JavaDoc

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
@@ -69,7 +69,7 @@ public interface ObjectPrx {
     /**
      * Tests whether the target object of this proxy can be reached.
      *
-     * @return a future that completes when a response is received from the target object
+     * @return a future that completes when the invocation completes
      */
     CompletableFuture<Void> ice_pingAsync();
 
@@ -77,7 +77,7 @@ public interface ObjectPrx {
      * Tests whether the target object of this proxy can be reached.
      *
      * @param context the request context
-     * @return a future that completes when a response is received from the target object
+     * @return a future that completes when the invocation completes
      */
     CompletableFuture<Void> ice_pingAsync(Map<String, String> context);
 


cpp/include/Ice/LocalExceptions.h
```diff
@@ -682,7 +681,7 @@ namespace Ice
         [[nodiscard]] const char* ice_id() const noexcept final;
     };
 
-    /// An attempt was made to find or deregister something that is not registered with Ice.
+    /// The exception that is thrown when attempting to find or deregister something that is not registered with Ice.
```
InsertCreativityHere, line 684: This is our standard wording for documenting exceptions.

cpp/include/Ice/Object.h
```diff
@@ -63,7 +63,7 @@ namespace Ice
         /// @private
         void _iceD_ice_ping(IncomingRequest&, std::function<void(OutgoingResponse)>);
 
-        /// Gets the Slice interfaces supported by this object as a list of type IDs.
+        /// Gets the Slice interfaces supported by this object as a list of Slice type IDs.
```
InsertCreativityHere, line 66: Elsewhere in this file, we specifically say `Slice type IDs`, not just `type IDs`.
Even in the `@return` here we say Slice.

cpp/include/Ice/NativePropertiesAdmin.h
```diff
@@ -13,7 +13,7 @@
 
 namespace Ice
 {
-    /// The default implementation for the "Properties" admin facet.
+    /// The default implementation of the "Properties" admin facet.
```
InsertCreativityHere, line 16: Ultra minor, but I think `of` is more grammatically correct here.

cpp/include/Ice/Proxy.h
```diff
@@ -319,12 +319,12 @@ namespace Ice
         /// @private
         void _iceI_ping(const std::shared_ptr<IceInternal::OutgoingAsyncT<void>>&, const Ice::Context&) const;
 
-        /// Gets the Slice interfaces supported by this object as a list of type IDs.
+        /// Gets the Slice interfaces supported by this object as a list of Slice type IDs.
```
InsertCreativityHere, line 322: Again, we usually qualify `type IDs` with `Slice`. Just updated these 3 functions to follow that.

cpp/include/Ice/Proxy.h
```diff
@@ -527,7 +527,7 @@ namespace Ice
             std::function<void(std::exception_ptr)> ex,
             std::function<void(bool)> sent = nullptr) const;
 
-        /// Flushes asynchronously any pending batched requests for this communicator.
+        /// Flushes any pending batched requests for this proxy asynchronously.
```
InsertCreativityHere, line 530: Grammar is a little off, and this is a proxy, not a communicator.

cpp/include/Ice/SlicedData.h
```diff
@@ -27,7 +27,7 @@ namespace Ice
         /// The Slice type ID for this slice. It's empty when the compact ID is set (not `-1`).
         const std::string typeId;
 
-        /// The Slice compact type ID for this slice. No compact ID is encoded as `-1`.
+        /// The Slice compact type ID for this slice. If no compact ID is set, it's encoded as `-1`.
```
InsertCreativityHere: The old sentence was confusingly worded.
It sounds like it's saying "compact IDs are never encoded as `-1`." which is not what we want to get across.

cpp/src/slice2java/Gen.cpp
InsertCreativityHere, line 1971: There is no need for closing `p` tags in java doc-comments; in fact they're discouraged.
Look for `<p>` on https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

InsertCreativityHere, line 1971: There is no need to put closing `p` tags in JavaDoc comments. In fact they're discouraged.
Search for `<p>` on https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/CompositeSliceLoader.java
```diff
@@ -6,12 +6,12 @@
...
 import java.util.List;
 
-/** Implements SliceLoader by combining multiple SliceLoaders. */
+/** Implements {@link SliceLoader} by combining multiple SliceLoader instances. */
 public final class CompositeSliceLoader implements SliceLoader {
     private final List<SliceLoader> _loaders;
 
     /**
-     * Creates a CompositeSliceLoader that combines the given SliceLoaders.
+     * Constructs a CompositeSliceLoader that combines the given SliceLoaders.
```
InsertCreativityHere, line 14: We prefer `constructs` in Java instead of `creates`.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Current.java
```diff
@@ -85,7 +85,7 @@ public Current(
      * this method to ensure that when an operation's mode is not idempotent (locally), the incoming
      * request's operation mode is not idempotent.
      *
-     * @throws MarshalException Thrown when the request's operation mode is {@link OperationMode#Idempotent}
+     * @throws MarshalException if the request's operation mode is {@link OperationMode#Idempotent}
```
InsertCreativityHere, line 88: `Thrown when` is the C++ wording, Java should just use `if ...`.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -471,8 +471,7 @@ public Thread newThread(Runnable r) {
 
     /**
      * Given a path name, first try to open it as a class path resource (the path is treated as
-     * absolute). If that fails, fall back to the file system. Returns null if the file does not
-     * exist and raises IOException if an error occurs.
```
InsertCreativityHere, line 475: This is already stated in the `@return`. No need to repeat it.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/AlreadyRegisteredException.java
```diff
@@ -7,9 +7,9 @@
  */
 public final class AlreadyRegisteredException extends LocalException {
     /**
-     * Constructs an AlreadyRegisteredException with the specified kind of object and ID.
+     * Constructs an AlreadyRegisteredException.
      *
-     * @param kindOfObject the kind of object that could not be removed
+     * @param kindOfObject the kind of object that is already registered
```
InsertCreativityHere, line 12: `is already registered` seems more correct than `could not be removed`.

cpp/include/Ice/Communicator.h
```diff
@@ -157,7 +157,6 @@ namespace Ice
         /// @param serverAuthenticationOptions The SSL options for server connections.
         /// @return The new object adapter.
         /// @see #createObjectAdapter
-        /// @see ObjectAdapter
```
InsertCreativityHere, line 160: No need to have these `@see`, Doxygen will already link the return type which is an `ObjectAdapterPtr`.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Object.java
```diff
@@ -101,7 +99,7 @@ default String ice_id(Current current) {
     /**
      * Returns the type ID of the associated Slice interface.
      *
-     * @return The return value is always ::Ice::Object.
+     * @return The return value is always {@code "::Ice::Object"}.
```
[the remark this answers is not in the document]
InsertCreativityHere: Copilot is missing the important distinction between whether the comment is a full sentence, or just a sentence fragment. Here, this is a full sentence, and so it should be capitalized as one.

cpp/include/Ice/SlicedData.h
bernardnormier: I would definitely avoid the term "encoded" or "marshaled" here. Should be simply something like:
`-1` means there is no compact ID.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/DispatchException.java
```diff
@@ -3,42 +3,44 @@
...
- * transmitted "over the wire".
+ * The exception that is thrown when a dispatch failed. This is the base class for local exceptions that can be
+ * marshaled and transmitted "over the wire".
+ * You can throw this exception in the implementation of an operation, or in a middleware. The Ice runtime then
+ * logically rethrows this exception to the client.
  */
 public class DispatchException extends LocalException {
     /**
-     * Constructs a DispatchException with the specified reply status, message, and cause.
+     * Constructs a DispatchException with the specified reply-status, message, and cause.
```
bernardnormier: I would remove all these "-": reply status is fine.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ModuleToPackageSliceLoader.java
```diff
@@ -7,32 +7,34 @@
...
- * Implements SliceLoader using a map Slice module to Java package.
+ * Implements SliceLoader using a map of Slice-module names to Java-package names.
  */
 public final class ModuleToPackageSliceLoader implements SliceLoader {
     private final Map<String, String> _moduleToPackageMap;
     private final ClassLoader _classLoader;
 
-    // We cache successful resolutions. The size of this cache is bounded by the number of Slice classes and exceptions
-    // in the program. We can't cache unsuccessful resolutions because it would create an unbounded cache.
+    /**
```
bernardnormier, line 16: Does it make sense to add doc-comments on private fields?

bernardnormier (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ModuleToPackageSliceLoader.java
InsertCreativityHere: Yes it does. Even if these won't show up in the public API reference of course, making these doc-comments means:
1) The Java tooling can use these comments (when you hover over a type for instance)
2) The JavaDoc tool will parse and validate these comments
3) Lets us use richer syntax like `@link`

There is no benefit to keeping non-doc comments like we are here, and nothing but upsides to switching.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/DispatchException.java
InsertCreativityHere: Fixed, I put it back to how it was (no dash).

cpp/include/Ice/SlicedData.h
InsertCreativityHere: Fixed in both Java and C++.
Now we say: ``The Slice compact type ID for this slice. `-1` means the slice has no compact ID.``

cpp/include/Ice/ObjectAdapter.h
```diff
@@ -193,15 +195,17 @@ namespace Ice
 
         /// Looks up a servant.
         /// @param id The identity of an Ice object.
```
pepone, line 197: We should improve the "find" description and "id" param. More like:

```
/// Looks up a servant by Identity.
/// @param id The identity.
```

For a separate PR.

cpp/include/Ice/ObjectAdapter.h
```diff
@@ -298,8 +302,8 @@ namespace Ice
         [[nodiscard]] virtual EndpointSeq getPublishedEndpoints() const = 0;
 
         /// Sets the endpoints that proxies created by this object adapter will contain.
-        /// @param newEndpoints The new set of endpoints that the object adapter will embed in proxies. Must contain
-        /// at least one endpoint.
+        /// @param newEndpoints The new set of endpoints that the object adapter will embed in proxies.
+        /// Must contain at least one endpoint.
```
pepone: Should we document `std::invalid_argument` raised from this method?

cpp/include/Ice/SlicedData.h
```diff
@@ -27,7 +27,7 @@ namespace Ice
         /// The Slice type ID for this slice. It's empty when the compact ID is set (not `-1`).
         const std::string typeId;
 
-        /// The Slice compact type ID for this slice. No compact ID is encoded as `-1`.
+        /// The Slice compact type ID for this slice. `-1` means the slice has no compact ID.
```
pepone: What about "Or `-1` if the slice ...`

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/CompactSliceTypeId.java
```diff
@@ -11,7 +11,7 @@
  * Assigns a compact Slice type ID to a class.
  *
  * <p>The Slice compiler assigns both a Slice type ID and a compact Slice type ID to the mapped class of a Slice class
- * that specifies a compact type ID.</p>
+ * that specifies a compact type ID.
```
pepone, line 14: Did you remove the end `</p>` on purpose?

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Endpoint.java
```diff
@@ -2,19 +2,22 @@
...
+ */
 public interface Endpoint {
     /**
-     * Return a string representation of the endpoint.
+     * Returns a string representation of this endpoint.
      *
-     * @return The string representation of the endpoint.
+     * @return the string representation of this endpoint
      */
     String _toString();
```
pepone, line 15: Can we get rid of this `_toString`, and directly use `Object.toString`? seems like a leftover of local Slice.

We already have:

```
    @Override
    public String toString() {
        return _toString();
    }
```

for a separate PR.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InitializationData.java
```diff
@@ -83,20 +84,18 @@ public InitializationData clone() {
...
-     * The plug-in factories. The corresponding plug-ins are created during communicator initialization,
+     * A list of plug-in factories. The corresponding plug-ins are created during communicator initialization,
      * in order, before all other plug-ins.
      */
     public List<PluginFactory> pluginFactories = Collections.emptyList();
 
-    /**
-     * The Slice loader, used to unmarshal Slice classes and exceptions.
-     */
+    /** The Slice loader, used to unmarshal Slice classes and exceptions. */
```
pepone: We should clarify the relation with the class loader above, for example the class loader description mentions is used to load class instances and exceptions. My understanding is the Slice loader takes precedence and the class loader will not be used when you provide the Slice loader. I think it would be helpful to clarify this for users.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Logger.java
```diff
@@ -3,54 +3,54 @@
 package com.zeroc.Ice;
 
 /**
- * Represents Ice's abstraction for logging and tracing. Applications can provide their own logger by implementing this
- * abstraction and setting a logger on the communicator.
+ * Represents Ice's abstraction for logging and tracing.
+ * Applications can provide their own logger by implementing this abstraction and setting a logger on the communicator.
```
pepone:
```suggestion
 * Applications can provide their own logger by implementing this interface and setting a logger on the communicator.
```
?

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerPlugin.java
```diff
@@ -8,7 +8,7 @@
  */
 public class LoggerPlugin implements Plugin {
     /**
-     * Constructs a logger plugin. This constructor installs a custom logger in the communicator. The communicator takes
+     * Constructs a logger plugin and installs it in the provided communicator. The communicator takes
```
pepone: "it" sounds like we are installing the "logger plugin" instead of the "logger".

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Object.java
```diff
@@ -160,8 +156,7 @@ static CompletionStage<OutgoingResponse> _iceD_ice_ping(Object obj, IncomingRequ
     }
 
     /**
-     * Dispatches an incoming request and returns the corresponding outgoing response
-     * for the {@link ice_ids} operation.
+     * Dispatches an incoming request and returns a corresponding outgoing response for the {@link ice_ids} operation.
```
pepone, line 159: Just curious why did you switch "the corresponding"  to "a corresponding", the former sounded better to me?

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
```diff
@@ -60,113 +62,107 @@ public interface ObjectPrx {
...
-     * @param context The context map for the invocation.
+     * @param context the request context
      */
     void ice_ping(Map<String, String> context);
 
     /**
      * Tests whether the target object of this proxy can be reached.
      *
-     * @return A future for the completion of the request.
+     * @return a future that completes when the ping is successfully dispatched
```
pepone: I will not talk about dispatched here. For example if you use a oneway proxy or datagram proxy to sent the ping the future can complete before the ping is dispatched. We should talk about invocation for oneway the invocation completes when the request is sent, for twoways when we receive the response, ...

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
```diff
@@ -206,369 +200,368 @@ CompletableFuture<Object.Ice_invokeResult> ice_invokeAsync(
...
+     * connection and ignore the return value.
      *
      * @return The {@link Connection} for this proxy.
-     * @see Connection
      */
     Connection ice_getConnection();
 
     /**
-     * Gets an executor object that uses the Ice thread pool.
+     * Returns an executor object that uses the Ice thread pool.
```
pepone, line 543: We should clarify which thread pool, I assume is the Ice client thread pool.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
```diff
@@ -636,9 +629,9 @@ static ObjectPrx checkedCast(ObjectPrx obj) {
     /**
      * Creates a new proxy from an existing proxy after confirming the target object's type via a remote invocation.
      *
-     * @param obj The source proxy.
-     * @param context The {@code Context} map for the invocation.
-     * @return A new proxy or {@code null} if the target object does not support the specified type.
+     * @param obj the source proxy
+     * @param context the request context
+     * @return a new proxy or {@code null} if the target object does not support the specified type
```
pepone: "the specified type" sounds like we pass the type as a parameter. Maybe " if the target object does not support this proxy's Slice type ID"?

pepone (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/CompactSliceTypeId.java
InsertCreativityHere: Yes, there is no need for closing p tags in java doc-comments; in fact they're discouraged.
Look for <p> on https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html

cpp/include/Ice/ObjectAdapter.h
InsertCreativityHere: Documented this exception in both C++ and Java (and removed `Must contain at least one endpoint` from the `@param` since it was just saying the same thing).

cpp/include/Ice/SlicedData.h
InsertCreativityHere: Fine with me.
Changed to `The Slice compact type ID for this slice, or `-1` if the slice has no compact ID.`
In both C++ and Java

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerPlugin.java
InsertCreativityHere: You're correct. Changed it to:
`Constructs a logger plugin and installs a custom logger in the provided communicator`

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Object.java
InsertCreativityHere: Well, the boring answer is that `a` sounded better to me :)
I think the reason why is: the difference between `a` vs `the` is that `the` is more specific than `a` is.
`the chair` talks about a single chair, which is well-known from the context, `a chair` could be any chair in the world.

Here, it's not like we have a specific `OutgoingResponse` type or singleton for the operation being dispatched.
We're going to create a new instance of `OutgoingResponse` (which could be used for any operation), and return it.

And this is the first time we mention outgoing response in the sentence. At the bottom of the comment, we have
`@return a {@link CompletionStage} that will complete with the outgoing response`
Here, `the` is perfect, since we're talking about a single, specific, pre-established outgoing response now: the outgoing response we mentioned in the summary.

----

A simpler example is:
`This returns an empty string` makes it sound like each time you call this, you're get a freshly allocated empty string.
`This returns the empty string` makes the empty string sound like a singleton. You aren't allocating a new empty string every time you call this.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
InsertCreativityHere: For the case of `ice_ping` specifically then, how about:
`a future that completes when a response is received from the target object`
I don't think `oneway` is meaningful here, kind of defeats the purpose of pinging something.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
InsertCreativityHere: Fixed proxies can use an object adapter's thread pool, so it isn't always guaranteed to use the client thread pool.
So it could be the Ice client thread pool or the Ice server thread pool depending on the context.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
InsertCreativityHere: Agreed, looks like this comment was previously copied from C++, where it does use a type parameter!

I went with
`@return a new proxy or {@code null} if the target object does not support this proxy's type`
Because I don't think it makes sense to "support a Slice type ID", since that's just a string.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Endpoint.java
InsertCreativityHere: Sure, we could probably remove this from the API for `Endpoint`, and we also have it on `Connection` as well.
Indeed, outside the scope of this PR though.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InitializationData.java
InsertCreativityHere: I think the clearest way was to update the doc for `classLoader` saying that:
```
     * <p>If a custom {@link #sliceLoader} is provided, then that will be used to unmarshal class instances and
     * user exceptions, instead of this class loader.
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
pepone: I will just says when the "invocation completes"

InsertCreativityHere: Changed as suggested.

PR #4665, opened 2025-12-01
Reject duplicated options when parsing an endpoint string

### LINKED ISSUE #487: Reject duplicated options when parsing stringified endpoints
Ice accepts stringified endpoints with duplicated options, such as:
```
Hello.Proxy=hello:tcp -p 10002 -h localhost -p 10000 -h 127.0.0.1:udp -p 10000:ssl -p 10001
```
which is surprising and error-prone. The last option appears to win.

This bug affects Ice 3.7 and earlier releases. We can't fix it in a patch release as such a fix could break existing applications.

bernardnormier:
Fixes #487 in all languages + add tests.

### DIFF
--- a/cpp/src/Ice/EndpointI.cpp
+++ b/cpp/src/Ice/EndpointI.cpp
@@ -101,10 +101,7 @@ IceInternal::EndpointI::initWithOptions(vector<string>& args)
         {
             if (!knownOptions.insert(option).second)
             {
-                throw Ice::ParseException{
-                    __FILE__,
-                    __LINE__,
-                    "duplicate option '" + option + "' in endpoint " + str};
+                throw Ice::ParseException{__FILE__, __LINE__, "duplicate option '" + option + "' in endpoint " + str};
             }
         }
         else
--- a/cpp/src/Ice/EndpointI.cpp
+++ b/cpp/src/Ice/EndpointI.cpp
@@ -1,8 +1,10 @@
 // Copyright (c) ZeroC, Inc.
 
 #include "EndpointI.h"
+#include "Ice/LocalExceptions.h"
 #include "Ice/OutputStream.h"
 
+#include <set>
 #include <sstream>
 
 using namespace std;
@@ -63,7 +65,7 @@ IceInternal::EndpointI::initWithOptions(vector<string>& args)
     vector<string> unknown;
 
     ostringstream ostr;
-    ostr << '`' << protocol() << " ";
+    ostr << "'" << protocol() << " "; // protocol means transport here
     for (const auto& arg : args)
     {
         if (arg.find_first_of(" \t\n\r") != string::npos)
@@ -77,6 +79,8 @@ IceInternal::EndpointI::initWithOptions(vector<string>& args)
     }
     ostr << "'";
     const string str = ostr.str();
+
+    set<string> knownOptions;
 
     for (vector<string>::size_type n = 0; n < args.size(); ++n)
     {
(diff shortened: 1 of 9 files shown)


js/test/Ice/proxy/Client.ts
```diff
@@ -416,6 +416,38 @@ export class Client extends TestHelper {
...
+
+        objPrx = communicator.stringToProxy("hello:tcp -h host1 -p 10000 --sourceAddress 127.0.0.1");
+        test(objPrx !== null);
+        try {
+            communicator.stringToProxy(
+                "hello:tcp -h host1 -p 10000 --sourceAddress 127.0.0.1 --sourceAddress 10.0.0.1");
+            test(false);
+        } catch (ex) {
+            test(ex instanceof Ice.ParseException, ex as Error);
+        }
```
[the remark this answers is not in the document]
bernardnormier: JS does not support udp.

cpp/src/Ice/EndpointI.cpp
```diff
@@ -93,7 +97,17 @@ IceInternal::EndpointI::initWithOptions(vector<string>& args)
...
 
-        if (!checkOption(option, argument, str))
+        if (checkOption(option, argument, str))
+        {
+            if (!knownOptions.insert(option).second)
+            {
+                throw Ice::ParseException{
+                    __FILE__,
+                    __LINE__,
+                    "duplicate option '" + option + "' in endpoint " + str};
```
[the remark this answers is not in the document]
bernardnormier: That's the C++ convention: lowercase and no trailing period.

externl (approved):

cpp/test/Ice/proxy/AllTests.cpp
```diff
@@ -413,6 +413,46 @@ allTests(TestHelper* helper)
     id2 = Ice::stringToIdentity(idStr);
     test(id == id2);
 
+    //
+    // Test proxies with duplicated endpoint option.
+    //
+
+    std::optional<Ice::ObjectPrx> objPrx = communicator->stringToProxy("hello:tcp -h host1 -p 4061");
```
pepone: Will be a bit clear to have just an array with the 3 string that retuns in a ParseException and test in a loop, no need to test pasrsing a "good endpoint" here.

js/test/Ice/proxy/Client.ts
```diff
@@ -416,6 +416,39 @@ export class Client extends TestHelper {
         id2 = Ice.stringToIdentity(idStr);
         test(id.equals(id2));
 
+        //
+        // Test proxies with duplicated endpoint option.
+        //
+
+        let objPrx = communicator.stringToProxy("hello:tcp -h host1 -p 4061");
```
pepone: We should use `defaultProtocol` defined at the begining, for browser compatibility. I think you cannot parse TCP endpoints in a browser can you?

cpp/test/Ice/proxy/AllTests.cpp
bernardnormier: Without a good endpoint "base", you don't know if the ParseException comes from the duplicated option or some other error in the endpoint / stringified proxy.

But yes, I could restructure this test to use a loop.

js/test/Ice/proxy/Client.ts
bernardnormier: We're running this test in a browser?

js/src/Ice/EndpointI.js
```diff
@@ -37,7 +41,12 @@ export class EndpointI {
                 argument = args[i++];
             }
 
-            if (!this.checkOption(option, argument, str)) {
+            if (this.checkOption(option, argument, str)) {
+                if (knownOptions.has(option)) {
+                    throw new ParseException(`duplicate option '${option}' in endpoint ${str}`);
```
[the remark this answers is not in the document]
bernardnormier: same style as C++

js/test/Ice/proxy/Client.ts
pepone: Only manually, not in CI. There is `--browser Manual` option for js/allTests.py

bernardnormier: Fixed

cpp/test/Ice/proxy/AllTests.cpp
bernardnormier: Updated tests to use the `default` transport + loop.

pepone (approved):
Looks good!

PR #4666, opened 2025-12-01
Prevent Checkstyle from Caching 'up-to-date' So it Always Re-Runs

InsertCreativityHere:
The latest gradle has better caching, which helps speed up builds.
Normally, this is a good thing, but it's causing gradle to skip running `checkstyle` more than once without cleaning in the middle.

Because the first time that `checkstyle` is run, gradle marks it as up-to-date. And if you try to re-run it, it will happily skip it.

----

This PR marks all checkstyle tasks as never up to date, and also disables gradle's built in state tracking.
This way `checkstyle` will always be run, and yield the exact same output. No matter how many times it's been run before.

### DIFF
--- a/java/test/android/controller/build.gradle
+++ b/java/test/android/controller/build.gradle
@@ -114,6 +114,11 @@ task checkstyle(type: Checkstyle) {
 
 tasks.named('checkstyle').configure {
     dependsOn(copyCertificates)
+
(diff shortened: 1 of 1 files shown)


bernardnormier:
The issue is when the build has checkstyle errors. You want to get these errors again and again - not just the first time. Maybe there is a way to improve this PR to cache only when there is no error.

Issue #4025, opened 2025-05-27
Remove Support for Writing `null` Enum Values in Java


InsertCreativityHere:
The static implementation of `ice_write` that `slice2java` generates for enums has a special case to handle `null`.
If you try to marshal a `null` enumerator, we'll instead marshal the first declared enumerator.
This seems very unexpected in my opinion.

My only guess is that we're trying to save users from possibly getting a NPE (null-pointer exception).
But... Users really shouldn't be passing `null` enums around, and we shouldn't be trying to save them from the consequences of doing so.

----

The code-gen in question:
```
out << sp << nl << "public static void ice_write(com.zeroc.Ice.OutputStream ostr, " << name << " v)";
out << sb;
out << nl << "if(v == null)";
out << sb;
string firstEnum = enumerators.front()->mappedName();
out << nl << "ostr.writeEnum(" << absolute << '.' << firstEnum << ".value(), " << p->maxValue() << ");";
out << eb;
out << nl << "else";
out << sb;
out << nl << "ostr.writeEnum(v.value(), " << p->maxValue() << ");";
out << eb;
out << eb;
```

InsertCreativityHere:
> I believe we also have a similar behavior for strings somewhere.

Yes, we have a special handling of null sequences, where if you marshal a `null` sequence, we will marshal an empty sequence of the corresponding type. Same for string, if you marshal a `null` string, we will marshal the empty string for you.

Personally, these don't feel that crazy to me. Not like replacing a `null` enum with the first enumerator (which is crazy).
Also, this behavior also exists in C# too. So, I'd rather just fix the weird Java-specific behavior and leave the rest alone.

----

Here are all the special cases: 
[...]

PR #4672, opened 2025-12-02
Remove Support for Marshaling `null` Enums and Structs in Java

InsertCreativityHere:
This PR implements #4025, and goes one step further.

In Java, if you marshal a null enum, we will instead marshal the first enumerator you declared in your Slice enum.
And, if you marshal a null struct (mapped to a class in Java), we will instead marshal a default instantiated class instance.

Both of these behaviors are unique to Java and in my opinion, unexpected.
This PR removes both of these special cases, now if you try to marshal a null enum or struct, you will get a `NullPointerException`.

----

After this PR, the only special handling for `null` is that if you pass a `null` array/buffer/etc., we will instead marshal an empty sequence. I don't think this is nearly as crazy, and it's not specific to Java either.

### DIFF
--- a/CHANGELOG-3.8.md
+++ b/CHANGELOG-3.8.md
@@ -825,6 +825,12 @@ initialization. See `InitializationData.pluginFactories`.
   by a WebSocket connection. Once the WebSocket `bufferedAmount` reaches this limit, sending additional data
   is delayed until it drops below the threshold.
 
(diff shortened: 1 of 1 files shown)


bernardnormier (commented):
I have two concerns with this PR:
 - it looks like we have the same null encoded as "default" in JS, Python and PHP. What's the plan for these languages?
 - there is no CHANGELOG entry

PR #4673, opened 2025-12-03
IcePy doc comment updates

externl:
This PR improves the doc comment situation in IcePy, for both the generated docs and editors/linters.

1. Sphinx reads the doc comments from the native extension directly. Its support of stub files doesn't work with our setup.
2. VScode / pylance reads the doc comments from our stub file and does not "fallback" to the native extension. I think it would be able to read the doc comments from the extension, but we need the stub file for the type hints. 

To make both cases happy, I've added doc comments for the re-exported IcePy types to the native extension and the stub file (with the help of Copilot). You can generate types from the extension (using stubgen), but right now they are incomplete, and it breaks the linting. It would be nice to improve this in the future. 

I've additionally added support for linking to the official python docs for python types.

### DIFF
--- a/python/test/Ice/adapterDeactivation/AllTests.py
+++ b/python/test/Ice/adapterDeactivation/AllTests.py
@@ -1,7 +1,7 @@
 # Copyright (c) ZeroC, Inc.
 
 import sys
-from typing import Any
+from typing import Any, cast
 
 from generated.test.Ice.adapterDeactivation import Test
 from TestHelper import TestHelper
@@ -97,7 +97,7 @@ def allTests(helper: TestHelper, communicator: Ice.Communicator) -> Test.TestInt
         # create new connection
         cachedConnection = obj.ice_getCachedConnection()
         assert cachedConnection is not None
-        cachedConnection.close().result()
+        cast(Ice.Future[None], cachedConnection.close()).result()
         obj.ice_ping()
 
         # TODO: compare underlying object adapter objects
@@ -109,7 +109,7 @@ def allTests(helper: TestHelper, communicator: Ice.Communicator) -> Test.TestInt
         # create new connection
         cachedConnection = obj.ice_getCachedConnection()
         assert cachedConnection is not None
-        cachedConnection.close().result()
+        cast(Ice.Future[None], cachedConnection.close()).result()
         obj.ice_ping()
 
         cachedConnection = obj.ice_getCachedConnection()
--- a/python/test/Ice/ami/AllTests.py
+++ b/python/test/Ice/ami/AllTests.py
@@ -158,7 +158,7 @@ def allTests(helper: TestHelper, communicator: Ice.Communicator, collocated: boo
         f = p.startDispatchAsync()
         assert isinstance(f, Ice.InvocationFuture)
         f.sent()  # Ensure the request was sent before we close the connection.
-        closeConnectionFuture = con.close()
+        closeConnectionFuture = cast(Ice.Future[None], con.close())
 
         # give time for startDispatch to start in the server before we call finishDispatch
         time.sleep(0.1)  # 100ms
@@ -181,7 +181,7 @@ def allTests(helper: TestHelper, communicator: Ice.Communicator, collocated: boo
         # Wait until the connection is gracefully closed.
         con = p.ice_getCachedConnection()
         assert con is not None
-        con.close().result()
+        cast(Ice.Future[None], con.close()).result()
         test(closeConnectionFuture.done())
 
         print("ok")
@@ -476,7 +476,7 @@ def allTestsFuture(helper: TestHelper, communicator: Ice.Communicator, collocate
         # Wait until the connection is gracefully closed.
         con = b1.ice_getConnection()
         assert con is not None
-        con.close().result()
+        cast(Ice.Future[None], con.close()).result()
         cb = FutureFlushCallback()
         f = b1.ice_flushBatchRequestsAsync()
         assert isinstance(f, Ice.InvocationFuture)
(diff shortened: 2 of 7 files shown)


python/python/IcePy-stubs/__init__.pyi
```diff
@@ -198,26 +533,15 @@ class ObjectPrx:
     def ice_endpointSelection(self, newType: Ice.EndpointSelectionType) -> Self: ...
     def ice_endpoints(self, newEndpoints: tuple[Endpoint, ...] | list[Endpoint]) -> Self: ...
     def ice_facet(self, new_facet: str) -> Self: ...
-    def ice_fixed(self, connection: Connection) -> Self: ...
+    def ice_fixed(self, connection: Ice.Connection) -> Self: ...
```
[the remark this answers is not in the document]
externl: No, we want to be explicit and use the Ice one, not IcePy.

python/modules/IcePy/Connection.cpp
```diff
@@ -591,54 +591,159 @@ connectionThrowException(ConnectionObject* self, PyObject* /*args*/)
...
+     PyDoc_STR("abort() -> None\n\n"
+               "Aborts this connection.")},
+    {"close",
+     reinterpret_cast<PyCFunction>(connectionClose),
+     METH_NOARGS,
+     PyDoc_STR("close() -> Ice.Future[None]\n\n"
+               "Starts a graceful closure of this connection once all outstanding invocations have completed.\n\n"
+               "Returns\n"
+               "-------\n"
+               "Ice.Future[None]\n"
```
pepone: The method doesn't returns always return an `Ice.Future`, it calls `wrapFuture` which can return another future type. Can we use `Awaitable[None]` like for Python code?

This is true for other methods that return futures, we always call `wrapFuture`.

python/modules/IcePy/Connection.cpp
```diff
@@ -591,54 +591,159 @@ connectionThrowException(ConnectionObject* self, PyObject* /*args*/)
...
      METH_VARARGS,
-     PyDoc_STR("setAdapter(Ice.ObjectAdapter) -> None")},
+     PyDoc_STR("setAdapter(adapter: Ice.ObjectAdapter | None) -> None\n\n"
+               "Associates an object adapter with this connection.\n\n"
+               "When a connection receives a request, it dispatches this request using\n"
+               "its associated object adapter. If the associated object adapter is None,\n"
+               "the connection rejects any incoming request with an ObjectNotExistException.\n\n"
+               "Parameters\n"
+               "----------\n"
+               "adapter : Ice.ObjectAdapter or None\n"
```
pepone, line 638:
```suggestion
               "adapter : Ice.ObjectAdapter | None\n"
```

Don't we use the same syntax in both places, the type hint and the doc comment?

python/python/Ice/IcePyTypes.py
```diff
@@ -0,0 +1,61 @@
...
+    "OpaqueEndpointInfo",
+    "SSLConnectionInfo",
+    "SSLEndpointInfo",
+    "TCPConnectionInfo",
+    "TCPEndpointInfo",
+    "UDPConnectionInfo",
+    "UDPEndpointInfo",
+    "WSConnectionInfo",
+    "WSEndpointInfo",
+    "currentEncoding",
```
bernardnormier, line 51: Starting with currentEncoding: I don't see the point of reexporting or documenting any of these symbols in Python, with the exception of `loadSlice, `intVersion` and `stringVersion`.

In particular, all the Protocol symbols are totally useless: the protocol is always "1.0".

python/python/Ice/ObjectPrx.py
```diff
@@ -870,14 +871,14 @@ def ice_getConnectionId(self) -> str:
         """
         return super().ice_getConnectionId()
 
-    def ice_fixed(self, connection: IcePy.Connection) -> Self:
+    def ice_fixed(self, connection: Connection) -> Self:
```
bernardnormier, line 874: This update is because IcePy.Connection is reexported as Ice.Connection?

python/modules/IcePy/Init.cpp
```diff
@@ -222,29 +236,30 @@ static PyMethodDef methods[] = {
...
-         "loadSlice(args: list[str]) -> None\\n\\n"
-         "Dynamically loads Slice definitions from one or more files and generates the corresponding Python code.\\n\\n"
-         "Parameters\\n"
-         "----------\\n"
-         "args : list[str]\\n"
+         "loadSlice(args: list[str]) -> None\n\n"
+         "Dynamically loads Slice definitions from one or more files and generates the corresponding Python code.\n\n"
+         "Parameters\n"
+         "----------\n"
+         "args : list[str]\n"
```
pepone, line 229: We should describe args

python/python/IcePy-stubs/__init__.pyi
```diff
@@ -49,30 +98,204 @@ class Communicator:
...
+
+        Parameters
+        ----------
+        compress : Ice.CompressBatch
+            Specifies whether or not the queued batch requests should be
+            compressed before being sent over the wire.
+
+        Returns
+        -------
+        Ice.Future[None]
```
pepone: use Awaitable like in the return type.

python/python/IcePy-stubs/__init__.pyi
```diff
@@ -101,33 +358,90 @@ class Endpoint:
     def __ne__(self, other: object) -> bool: ...
 
 class EndpointInfo:
+    """
+    Base class for all endpoint info classes.
+
+    Provides access to the endpoint details. Endpoint info classes are
+    used to get information about the endpoints that a connection or
+    proxy uses.
```
pepone: Any reason to not use 120 line length here?

externl: nope.

python/python/Ice/__init__.py
```diff
@@ -69,6 +40,34 @@
...
+    SSLEndpointInfo,
+    TCPConnectionInfo,
+    TCPEndpointInfo,
+    UDPConnectionInfo,
+    UDPEndpointInfo,
+    WSConnectionInfo,
+    WSEndpointInfo,
+    currentEncoding,
+    currentProtocol,
+    currentProtocolEncoding,
```
InsertCreativityHere, line 62: Also, just because we were talking about it, do these things need to be in this list?

Maybe I'm misunderstanding what this is, but it looks like we're re-exporting IcePy types into the Ice module?

externl: Yes, we are!

PR #4675, opened 2025-12-03
Fix check for binary distribution in C++ NuGet package

pepone:
This PR contains a minor fix for the C++ NuGet package to correctly detect the binary distribution.


cpp/tools/ZeroC.Ice.Slice.Tools.Cpp/ZeroC.Ice.Slice.Tools.Cpp.props
```diff
@@ -3,7 +3,7 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <!-- default values -->
     <Choose>
-        <When Condition="Exists('$(MSBuildThisFileDirectory)..\ZeroC.Ice.Cpp.nuspec')">
```
pepone, line 6: This file is not installed when you install the C++ package, .NET installs the `.nuspec` but C++ excludes them.

PR #4678, opened 2025-12-03
Add "service status" command to icegridadmin

### LINKED ISSUE #43: Add icegridadmin service status
You can start and stop a service with icegridadmin, however you currently can't figure out it if it's running or not.

ICE-8268

bernardnormier:
Fixes #43.


cpp/src/IceGrid/Grammar.y
```diff
@@ -99,6 +99,7 @@ yyerror(const char* s)
 %token ICEGRID_STRING
 %token ICEGRID_START
 %token ICEGRID_STOP
+%token ICEGRID_STATUS
```
bernardnormier, line 102: The existing command for server "status" is called state.

We should consider renaming state -> status (as in server status), and possibly keeping state as an alias for status.

externl: To me sate != status. Status can include state, but they'r not the same thing.

externl (approved):
Looks good!

java/src/com.zeroc.icebox/src/main/java/com/zeroc/IceBox/Admin.java
```diff
@@ -79,6 +80,23 @@ private static int run(
             String command = commands.get(i);
             if ("shutdown".equals(command)) {
                 manager.shutdown();
+            } else if ("status".equals(command)) {
+                if (++i >= commands.size()) {
+                    System.err.println("IceBox.Admin: no service name specified.");
```
InsertCreativityHere: Unsurprisingly, we aren't consistent in this file, but I think it's better for error messages to not end with punctuation
```suggestion
                    System.err.println("IceBox.Admin: no service name specified");
```
And I'm definitely not a fan of having a mix like we currently do in this file.

java/src/com.zeroc.icebox/src/main/java/com/zeroc/IceBox/Admin.java
```diff
@@ -79,6 +80,23 @@ private static int run(
...
+                String service = commands.get(i);
+                try {
+                    boolean running = manager.isServiceRunning(service);
+                    if (running) {
+                        System.out.println("running");
+                    } else {
+                        System.out.println("stopped");
+                    }
+                } catch (NoSuchServiceException ex) {
+                    System.err.println("IceBox.Admin: unknown service '" + service + "'");;
```
InsertCreativityHere: An extra semi-colon snuck in here. But I wouldn't be opposed to keeping it.
```suggestion
                    System.err.println("IceBox.Admin: unknown service '" + service + "'");
```

java/src/com.zeroc.icebox/src/main/java/com/zeroc/IceBox/Admin.java
```diff
@@ -79,6 +80,23 @@ private static int run(
...
+                    return 1;
+                }
+                String service = commands.get(i);
+                try {
+                    boolean running = manager.isServiceRunning(service);
+                    if (running) {
+                        System.out.println("running");
+                    } else {
+                        System.out.println("stopped");
+                    }
```
InsertCreativityHere, line 95: There's absolutely nothing wrong with this, but I think this would be simpler:
```suggestion
                    String state = (manager.isServiceRunning(service) ? "running" : "stopped");
                    System.out.println(state);
```

slice/IceBox/ServiceManager.ice
```diff
@@ -62,6 +62,13 @@ module IceBox
         void stopService(string service)
             throws AlreadyStoppedException, NoSuchServiceException;
 
+        /// Returns whether a service is running.
+        /// @param service The service name.
```
InsertCreativityHere: I think this could be a little more descriptive:
```suggestion
        /// @param service The name of the service to check.
```

InsertCreativityHere (approved):

slice/IceBox/ServiceManager.ice
```diff
@@ -62,6 +62,13 @@ module IceBox
...
+        /// @return @c true if the service was started and not stopped; otherwise @c false.
```
InsertCreativityHere: We decided against supported `@c` and opted to support only backticks for our code formatting.
```suggestion
        /// @return `true` if the service was started and not stopped; otherwise `false`.
```

InsertCreativityHere (approved):

pepone (approved):

csharp/src/iceboxnet/ServiceManagerI.cs
```diff
@@ -166,7 +166,7 @@ public override void stopService(string name, Ice.Current current)
             for (i = 0; i < _services.Count; ++i)
             {
                 info = _services[i];
-                if (info.name.Equals(name, StringComparison.Ordinal))
+                if (info.name == name)
```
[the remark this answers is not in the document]
bernardnormier: @copilot, can you replace string Equals in this file by `==` comparisons?

csharp/src/iceboxnet/ServiceManagerI.cs
```diff
@@ -191,6 +191,21 @@ public override void stopService(string name, Ice.Current current)
...
+    {
+        lock (_mutex)
+        {
+            foreach (ServiceInfo info in _services)
+            {
+                if (info.name == name)
+                {
+                    return info.status == ServiceStatus.Started;
+                }
+            }
```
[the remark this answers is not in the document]
bernardnormier: It would be inconsistent to use LINQ only for this new method.

csharp/src/iceboxnet/ServiceManagerI.cs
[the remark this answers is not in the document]
bernardnormier: Fixed

PR #4679, opened 2025-12-03
Remove Special Null Marshaling Logic from Javascript, PHP, and Python

InsertCreativityHere:
This PR follows #4672, and removes the special null handling from Javascript, PHP, and Python, which I believe is all of it.

Note that we still have special null handling for sequences in all languages (where passing in `null` for a sequence is treated the same as passing in an empty sequence). I don't find this to be as egregious, and at least we're consistent about it in all languages.

### DIFF
--- a/python/modules/IcePy/Types.cpp
+++ b/python/modules/IcePy/Types.cpp
@@ -1127,7 +1127,7 @@ IcePy::StructInfo::getId() const
 bool
 IcePy::StructInfo::validate(PyObject* val)
 {
-    return val == Py_None || PyObject_IsInstance(val, pythonType) == 1;
+    return PyObject_IsInstance(val, pythonType) == 1;
 }
 
 bool


python/modules/IcePy/Types.cpp
```diff
@@ -1172,21 +1170,7 @@ IcePy::StructInfo::marshal(
...
-            PyObjectHandle emptyArgs{PyTuple_New(0)};
-            auto* type = reinterpret_cast<PyTypeObject*>(pythonType);
-            // Create a new instance of the type
-            _nullMarshalValue = type->tp_new(type, emptyArgs.get(), nullptr);
-            // Call the instance __init__ method
-            type->tp_init(_nullMarshalValue.get(), emptyArgs.get(), nullptr);
-        }
-        p = _nullMarshalValue.get();
-    }
+    assert(PyObject_IsInstance(p, pythonType) == 1); // validate() should have caught this.
```
bernardnormier, line 1173: No need to update validate?

InsertCreativityHere: I could of sworn I fixed the validate, but I don't know where it went.
Either way, fixed now.

pepone (approved):

Issue #389, opened 2019-04-04
Add address information to socket exception or a new exception


bentoi:
We have a number of network failures that occur when dealing with IP addresses (connect and  bind in particular). The IP address is not included in the exception raised for the failure making it difficult in some situation for the user application to figure out the origin of the exception. It would be good to stringify the address and somehow ensure its included with the exception. 

We could add an address member to SocketException or add it to ConnectFailedException and add BindException.

bernardnormier:
We should include the address in the message. Not add a separate field.

PR #4680, opened 2025-12-03
Add optional remote address to ConnectFailedException and other local…

bernardnormier:
… exceptions

This PR adds the remote address in the error messages for some SocketException (but not SocketException itself) in C++.

See #389.


cpp/src/Ice/LocalExceptions.cpp
```diff
@@ -348,10 +348,22 @@ namespace
...
+            return "no error code";
         }
         return IceInternal::errorToString(error);
     }
+
+    string makePrefixWithAddress(string messagePrefix, optional<string> address)
+    {
+        if (address.has_value())
+        {
+            return std::move(messagePrefix) + " (remote address = " + std::move(*address) + ")";
```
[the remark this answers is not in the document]
bernardnormier: I don't understand this comment. `address` (`*address`) is only moved once in this code.

cpp/src/Ice/ios/iAPConnector.mm
```diff
@@ -19,7 +19,7 @@
     EASession* session = [[EASession alloc] initWithAccessory:_accessory forProtocol:_protocol];
     if (!session)
     {
-        throw Ice::ConnectFailedException(__FILE__, __LINE__, 0);
```
bernardnormier, line 22: We should avoid the "0" error code, as it doesn't provide any useful info. Better supply a string with some info about the error.

cpp/src/Ice/LocalExceptions.cpp
```diff
@@ -376,7 +398,7 @@ namespace
     {
         if (error == 0)
         {
-            return "recv() returned zero";
+            return "no error code";
```
bernardnormier, line 401: We often pass "0" when we don't have an error code. Better state that in the error message than some confusing message about recv.

cpp/src/Ice/Network.cpp
```diff
@@ -811,9 +811,16 @@ IceInternal::closeSocket(SOCKET fd)
...
-    return s.str();
+    if (isAddressValid(addr))
+    {
+        ostringstream s;
+        s << inetAddrToString(addr) << ':' << getPort(addr);
+        return s.str();
+    }
+    else
+    {
+        return "";
```
bernardnormier, line 822: This allows me to call `addrToString(addr)` blindly when throwing exceptions. If it's invalid, it's like I didn't provide an address.

bernardnormier (commented):
You'll notice that I didn't change any SSL code, even though SSL transceivers throw ConnectFailedException etc.

It's because it's not easy to get the remote address in this code ... for a follow-up PR.

cpp/src/Ice/LocalExceptions.cpp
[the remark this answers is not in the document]
pepone: I don't understand it either, but what is the point of `std::move(*address)` if we are just creating a new string.

pepone (approved):
> You'll notice that I didn't change any SSL code, even though SSL transceivers throw ConnectFailedException etc.


We can add a method in the Transceiver to do this, so you can ask the delegate to throw the exception.

cpp/include/Ice/LocalExceptions.h
externl, line 1: I'm not really a big fan of having so many overloads. For the 11 times we throw this exception we should just fill out all the parameters.

cpp/src/Ice/LocalExceptions.cpp
externl: I think we should rework this. All of this confusion comes from the overloads. recv() returning zero is still relevant in the places it was intended.

cpp/src/Ice/LocalExceptions.cpp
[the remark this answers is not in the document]
bernardnormier: Probably not that useful. My idea was to use operator+ overload (8):
https://en.cppreference.com/w/cpp/string/basic_string/operator%2B

externl (approved):

cpp/src/Ice/LocalExceptions.cpp
bernardnormier: No, the confusion comes from the use of "0" to mean "I don't know and I am too lazy to provide a proper message". We should indeed fix all these instances.

cpp/src/Ice/StreamSocket.cpp
```diff
@@ -190,7 +190,7 @@ StreamSocket::read(char* buf, size_t length)
 #endif
         if (ret == 0)
         {
-            throw Ice::ConnectionLostException(__FILE__, __LINE__, 0);
+            throw Ice::ConnectionLostException(__FILE__, __LINE__, 0, addrToString(_addr));
```
InsertCreativityHere: Is this also somewhere where we should avoid supplying a `0` error code?

cpp/src/Ice/StreamSocket.cpp
```diff
@@ -248,7 +248,7 @@ StreamSocket::write(const char* buf, size_t length)
...
```
InsertCreativityHere: Same

cpp/src/Ice/LocalExceptions.cpp
```diff
@@ -348,10 +348,22 @@ namespace
...
         }
         return IceInternal::errorToString(error);
     }
+
+    string makePrefixWithAddress(string messagePrefix, optional<string> address)
+    {
+        if (address.has_value() && !address->empty())
+        {
+            return std::move(messagePrefix) + " (remote address = " + *address + ")";
```
InsertCreativityHere, line 360: Is this move here really necessary?
Return-Value-Optimization doesn't work because we concatenate onto it?

InsertCreativityHere (approved):

cpp/src/Ice/LocalExceptions.cpp
bernardnormier: Yes, that's correct.

cpp/src/Ice/StreamSocket.cpp
bernardnormier: Yes, fixed.

PR #4681, opened 2025-12-04
Rename server state to server status (icegridadmin)

bernardnormier:


cpp/src/IceGrid/Scanner.l
```diff
@@ -213,7 +213,7 @@ namespace IceGrid
         keywordMap["describe"] = ICEGRID_DESCRIBE;
         keywordMap["properties"] = ICEGRID_PROPERTIES;
         keywordMap["property"] = ICEGRID_PROPERTY;
-        keywordMap["state"] = ICEGRID_STATE;
+        keywordMap["status"] = ICEGRID_STATUS;
```
bernardnormier, line 216: Missing from my previous PR.

cpp/src/IceGrid/Parser.h
```diff
@@ -54,15 +54,15 @@ namespace IceGrid
...
+        void statusServer(const std::list<std::string>&);
         void enableServer(const std::list<std::string>&, bool);
         void pidServer(const std::list<std::string>&);
         void propertiesServer(const std::list<std::string>&, bool);
         void listAllServers(const std::list<std::string>&);
 
         void startService(const std::list<std::string>&);
         void stopService(const std::list<std::string>&);
-        void serviceStatus(const std::list<std::string>&);
+        void statusService(const std::list<std::string>&);
```
bernardnormier, line 65: Renamed for consistency.

cpp/src/IceGrid/Scanner.l
InsertCreativityHere, line 216: Should we still support the old `state` keyword for backwards compatibility?
If we wanted to, it would be as simple as adding this one line:
```suggestion
        keywordMap["status"] = ICEGRID_STATUS;
        keywordMap["state"] = ICEGRID_STATUS; // Kept for backwards compatibility
```
No need to muck around with the grammar or parser.

InsertCreativityHere (approved):
Looks good. I think `status` is more expected than `state`.

pepone (approved):

cpp/src/IceGrid/Scanner.l
externl: No.

externl (approved):

PR #4683, opened 2025-12-04
Add peerAddress/serverAddress to ConnectFailedException etc.

bernardnormier:
This is the C# counterpart of #4680.

Like #4680, the SSL transceiver doesn't specify the address is these modified exceptions. That's for a follow-up PR.


csharp/src/Ice/Internal/UdpTransceiver.cs
```diff
@@ -356,7 +356,7 @@ public void finishRead(Buffer buf)
 
         if (ret == 0)
         {
-            throw new Ice.ConnectionLostException();
+            throw new Ice.ConnectionLostException(peerAddress: null);
```
pepone: Why null here?

pepone (approved):

externl (approved):

csharp/src/Ice/Internal/UdpTransceiver.cs
bernardnormier: Just a bug, fixed.

PR #4684, opened 2025-12-04
Add peerAddress/serverAddress to ConnectFailedException etc.

bernardnormier:
This is the Java counterpart of #4680 and #4683.

### DIFF
--- a/java/src/com.zeroc.icebt/src/main/java/com/zeroc/IceBT/TransceiverI.java
+++ b/java/src/com.zeroc.icebt/src/main/java/com/zeroc/IceBT/TransceiverI.java
@@ -312,7 +312,7 @@ final class TransceiverI implements Transceiver {
                 startReadWriteThreads();
             }
         } catch (IOException ex) {
-            exception(new ConnectFailedException(ex));
+            exception(new ConnectFailedException(null, ex));
         } catch (Exception ex) {
             exception(new SocketException(ex));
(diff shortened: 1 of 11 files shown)


InsertCreativityHere (approved):
Looks good to me!

java/src/com.zeroc.icebt/src/main/java/com/zeroc/IceBT/TransceiverI.java
```diff
@@ -312,7 +312,7 @@ private void runConnectThread() {
                 startReadWriteThreads();
             }
         } catch (IOException ex) {
-            exception(new ConnectFailedException(ex));
+            exception(new ConnectFailedException(null, ex));
```
pepone, line 315: If we use a string instead of a InetSocketAddress, we might add the info for this transport too.

pepone (approved):

externl (approved):

java/src/com.zeroc.icebt/src/main/java/com/zeroc/IceBT/TransceiverI.java
bernardnormier: I don't think it's critical for Bluetooth. This change (extra address) is primarily for the situation where you need to distinguish between several servers / peers, which would be very uncommon with BT.

PR #4685, opened 2025-12-05
Print icegridnode ready to standard out instead of using the logger

### LINKED ISSUE #528: Print IceGrid usage and server ready message to stdout
IceGrid uses Ice::Service::print to print the registry/node usage and the ready message for nodes (enabled with IceGrid.Node.PrintServersReady). It would be better to output these on the console output instead like for the object adapter ready message (it's using consoleOut from IceUtilInternal).

pepone:
Fix #528 

One curios thing not fixed that by this PR is that we still print the reading message when there is a deployment error. See handing of DeploymentException above.

### DIFF
--- a/cpp/src/IceGrid/IceGridNode.cpp
+++ b/cpp/src/IceGrid/IceGridNode.cpp
@@ -496,7 +496,7 @@ NodeService::startImpl(int argc, char* argv[], int& status)
         {
             //
             // Create was interrupted, return true as if the service was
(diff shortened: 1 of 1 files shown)


bernardnormier (approved):
With respect to the DeploymentException, it's unrelated to this PR, and I guess it's fine: you use this unusual deployment option, the extra deployment fails and you get a warning but the node still starts.

PR #4686, opened 2025-12-05
Do not export dispatch method

### LINKED ISSUE #3617: Improve IcePy::lookupType
IcePy uses lookupType to import the dispatch helper:

```
PyObject* dispatchMethod = lookupType("Ice.Dispatch.dispatch")
```

This produces unreliable results, for example it returns 0 for cross tests.

Adding `from .Dispatch import *` to `Ice/__init__.py` fixes it, but seems like this shouldn't be necessary. Particularly because this is a private helper function and adding it to init exposes it with the public Ice module.

I think will be better to review lookupType and avoid the need for this extra import.

pepone:
Fix #3617 

This PR removes the workaround for #3617 , avoid importing the internal "from .Dispatch import dispatch" in Ice module.

Additionally updated CI to test 3.12, 3.13, and 3.14 Python versions.

Update default to be 3.14, and added separate jobs to test 3.12 and 3.13.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,18 +143,21 @@ jobs:
             msbuild_project: "msbuild/ice.proj"
             build_cpp_and_python: true
             python_version: "3.12"
+            python_tests: true
 
           - os: macos-26
             config: "Python312"
             working_directory: "python"
             build_cpp_and_python: true
             python_version: "3.12"
+            python_tests: true
 
           - os: ubuntu-24.04
             config: "Python312"
             working_directory: "python"
             build_cpp_and_python: true
             python_version: "3.12"
+            python_tests: true
 
           # Python 3.13 testing
           - os: windows-2025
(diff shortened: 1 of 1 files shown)


python/python/Ice/__init__.py
```diff
@@ -360,7 +359,6 @@
     "currentEncoding",
     "currentProtocol",
     "currentProtocolEncoding",
-    "dispatch",
```
bernardnormier, line 363: There is no other change to the Python code. So IcePy.lookupType was already fixed?

pepone: I removed the workaround and no longer get the crash.

CI Python 3.14 crash on Ubuntu not sure if it is related, as we were not testing this version before. I will split the CI setup to a separate PR to be sure.

pepone:
This change triggers a crash with Python 3.14 on Ubuntu, need to debug.

PR #4687, opened 2025-12-05
Update CI to test additional Python versions

pepone:
This PR updates the default Python version used in CI to 3.14, and add jobs to also test Python 3.12 and Python 3.13.

This are the same changes that I have in #4686, but without the Python code updates.


externl (approved):
I'm not sure that this is really worth it for CI. Adding too many variants is going to consume many runners. I guess we can do this for now and walk it back if it becomes an issue.

PR #4688, opened 2025-12-05
Remove currentEncoding, currentProtocol and currentProtocolEncoding f…

bernardnormier:
…rom public API

(currentEncoding is removed entirely).

All these "current" meant latest supported version, namely:
currentProtocol = 1.0 since Ice 1.0
currentProtocolEncoding = 1.0 for protocol 1.0 (the only protocol ever supported by Ice)
currentProtocolEncoding = 1.1 since Ice 3.5


js/src/Ice/Initialize.js
pepone: It can be removed as reported by the linter

js/src/Ice/OutputStream.js
```diff
@@ -511,7 +511,7 @@ export class OutputStream {
             this._format = instance.defaultsAndOverrides().defaultFormat;
         } else {
             if (this._encoding === null) {
-                this._encoding = Protocol.currentEncoding;
+                this._encoding = Protocol.Encoding_1_1;
```
pepone: Copilot is right, you need to import the constant in:

```
import { Encoding_1_0, Protocol } from "./Protocol.js";
```

update to 

```
import { Encoding_1_0, Encoding_1_1, Protocol } from "./Protocol.js";
```

and then use `Encoding_1_1` instead of `Protocol.Encoding_1_1`

csharp/src/Ice/Internal/Protocol.cs
```diff
@@ -87,38 +89,13 @@ public sealed class Protocol
 
     internal static void checkSupportedEncoding(Ice.EncodingVersion v)
     {
-        if (v.major != encodingMajor || v.minor > encodingMinor)
+        if (v.major != Util.Encoding_1_1.major || v.minor > Util.Encoding_1_1.minor)
```
pepone: This is referencing the wrong `Util` class `Ice.Internal.Util` instead of `Ice.Util`

externl (approved):

cpp/test/Ice/proxy/AllTests.cpp
```diff
@@ -1010,7 +1010,15 @@ allTests(TestHelper* helper)
 
     ref13 = "test -p 1.3:" + endp;
     cl13 = MyClassPrx(communicator, ref13);
-    cl13->ice_ping(); // use protocol 1.0
+    try
```
pepone, line 1013: Should we mention this in CHANGELOG? I guess nobody should care

pepone (approved):

cpp/test/Ice/proxy/AllTests.cpp
bernardnormier: Nobody should be using these functions / constants.

PR #4689, opened 2025-12-05
Add favicon to Doxygen

### LINKED ISSUE #1868: Add Ice icon to Doxygen builds
Should we add our simplified Ice icon here?

_Originally posted by @InsertCreativityHere in https://github.com/zeroc-ice/ice/pull/1865#discussion_r1504590953_

pepone:
Fix #1868


js/assets/favicon.ico
pepone, line 1: JS was using an outdated icon.

bernardnormier:
Did you forget to add favicon.ico to assets?

externl (commented):
I think we should just use our logo.svg. .ico files are generally a fallback these days.

PR #4698, opened 2025-12-06
3.8.0 Version update

pepone:


cpp/msbuild/ZeroC.Ice.Cpp.nuspec
```diff
@@ -3,15 +3,15 @@
...
+    <version>3.8.0</version>
     <authors>ZeroC</authors>
     <copyright>Copyright (c) ZeroC, Inc.</copyright>
     <license type="expression">GPL-2.0-only</license>
     <projectUrl>https://github.com/zeroc-ice/ice</projectUrl>
     <icon>logo-128x128.png</icon>
     <requireLicenseAcceptance>false</requireLicenseAcceptance>
     <description>The Ice framework provides everything you need to build networked applications, including RPC, pub/sub, server deployment, and more.</description>
-    <releaseNotes>https://docs.zeroc.com/ice/latest/cpp/ice-3-8-0-preview</releaseNotes>
+    <releaseNotes>https://docs.zeroc.com/ice/3.8/cpp/ice-3-8-0</releaseNotes>
```
externl, line 14: I think we should try to keep links on main to `latest` and only update to 3.8 release links when we cut the branch. Otherwise we have to go back and change these again after we release.

pepone: Updated the PR to merge to 3.8

PR #4701, opened 2025-12-08
Simplify Python type lookup

### LINKED ISSUE #3617: Improve IcePy::lookupType

pepone:
Fix #3617

### DIFF
--- a/python/modules/IcePy/Util.cpp
+++ b/python/modules/IcePy/Util.cpp
@@ -555,31 +555,14 @@ IcePy::lookupType(const string& typeName)
     string moduleName = typeName.substr(0, dot);
     string name = typeName.substr(dot + 1);
 
-    //
-    // First search for the module in sys.modules.
-    //
-    PyObject* sysModules{PyImport_GetModuleDict()};
-    assert(sysModules);
-
-    PyObject* module = PyDict_GetItemString(sysModules, const_cast<char*>(moduleName.c_str()));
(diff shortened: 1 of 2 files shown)


python/modules/IcePy/Util.cpp
```diff
@@ -555,31 +555,14 @@ IcePy::lookupType(const string& typeName)
     string moduleName = typeName.substr(0, dot);
     string name = typeName.substr(dot + 1);
 
-    //
-    // First search for the module in sys.modules.
-    //
-    PyObject* sysModules{PyImport_GetModuleDict()};
-    assert(sysModules);
-
-    PyObject* module = PyDict_GetItemString(sysModules, const_cast<char*>(moduleName.c_str()));
```
bernardnormier, line 564: So the fix is to no longer look for symbols in sysModules?

pepone: Yes, we just call PyImport_ImportModule all the time.

PR #4711, opened 2025-12-09
Make internal version functions internal

bernardnormier:
This PR removes a number of "version" functions from the public API, but making them internal and removing them entirely.

This includes stringToProtocolVersion, stringToEncodingVersion, protocolVersionToString, encodingVersionToString, Protocol_1_0.

Updated all languages except JS.


cpp/src/Ice/Protocol.cpp
```diff
@@ -6,6 +6,75 @@
 using namespace std;
 using namespace Ice;
 
+string
```
bernardnormier, line 28: Moved from (deleted) VersionFunctions.cpp

cpp/src/Ice/Protocol.cpp
```diff
@@ -3,9 +3,97 @@
...
+{
+    string str{s};
+    std::string::size_type pos = str.find_first_of('.');
+    if (pos == std::string::npos)
+    {
+        throw ParseException(__FILE__, __LINE__, "malformed Ice version value '" + str + "'");
+    }
+
+    std::istringstream majStr(str.substr(0, pos));
+    int32_t majVersion;
```
pepone, line 75: did you consider using std:: stoi?

cpp/src/Ice/Protocol.cpp
```diff
@@ -3,9 +3,97 @@
...
+ostream&
+Ice::operator<<(ostream& out, const EncodingVersion& version)
+{
+    return out << static_cast<int>(version.major) << "." << static_cast<int>(version.minor);
+}
+
+void
+IceInternal::stringToMajorMinor(string_view s, uint8_t& major, uint8_t& minor)
+{
+    string str{s};
```
pepone, line 67: We can consider validating versions only contains decimal digits and one dot. I think the integer conversion here will allow a sign, and non decimal base numbers.

But it might be overkill as this is now internal and we should not be passing such version strings.

csharp/src/Ice/Internal/Protocol.cs
```diff
@@ -100,6 +140,39 @@ internal static bool
...
+            throw new ParseException($"malformed version value in '{str}'");
+        }
+
+        string majStr = str[..pos];
+        string minStr = str[(pos + 1)..];
+        int majVersion;
+        int minVersion;
+        try
+        {
+            majVersion = int.Parse(majStr, CultureInfo.InvariantCulture);
```
pepone, line 157: We can use NumberStyles.None to only accept decimal digits

```suggestion
            majVersion = int.Parse(majStr, NumberStyles.None);
```

See:
 https://learn.microsoft.com/en-us/dotnet/api/system.globalization.numberstyles?view=net-10.0
 https://learn.microsoft.com/en-us/dotnet/api/system.int32.parse?view=net-10.0#system-int32-parse(system-string-system-globalization-numberstyles)

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Protocol.java
```diff
@@ -135,5 +178,55 @@ public static boolean isSupported(EncodingVersion version, EncodingVersion suppo
...
+    private static byte stringToMajor(String str) {
+        int pos = str.indexOf('.');
+        if (pos == -1) {
+            throw new ParseException("malformed version value in '" + str + "'");
+        }
+
+        String majStr = str.substring(0, pos);
+        int majVersion;
+        try {
+            majVersion = Integer.parseInt(majStr);
```
pepone, line 190: We can pass radix 10, to no allow other bases. My understanding is all this versions should be base 10.

pepone (approved):

cpp/src/Ice/Protocol.cpp
bernardnormier: I only moved this code. The focus of this PR was to make these APIs internal or delete them.

cpp/src/Ice/Protocol.cpp
externl: I'm not sure that any of this code is new. We could also use std::regex.

externl (approved):
Looks good. Maybe worthy of a changelog entry.

InsertCreativityHere (approved):
Looks good to me!

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Protocol.java
bernardnormier: The plain parseInt we use is base10 already.

PR #4712, opened 2025-12-09
Remove ICE_API from internal classes/functions

bernardnormier:
It's easier to add ICE_API in a patch release then remove it.

### DIFF
--- a/cpp/src/Ice/SSL/RFC2253.h
+++ b/cpp/src/Ice/SSL/RFC2253.h
@@ -25,7 +25,7 @@ namespace Ice::SSL::RFC2253
 {
     using RDNSeq = std::list<std::pair<std::string, std::string>>;
 
-    struct ICE_API RDNEntry
+    struct RDNEntry
     {
         RDNSeq rdn;
         bool negate;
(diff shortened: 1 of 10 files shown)


cpp/src/Ice/SSL/RFC2253.h
```diff
@@ -25,7 +25,7 @@ namespace Ice::SSL::RFC2253
 {
     using RDNSeq = std::list<std::pair<std::string, std::string>>;
 
-    struct ICE_API RDNEntry
```
bernardnormier, line 28: I don't think ICE_API is ever necessary for a data-only struct.

cpp/src/Ice/SSL/SSLUtil.cpp
```diff
@@ -633,7 +633,8 @@ Ice::SSL::encodeCertificate(X509* certificate)
     BIO_free(out);
     return result;
 }
-ICE_API X509*
```
bernardnormier, line 636: It's already specified in the header.

externl (approved):

pepone (approved):

PR #4713, opened 2025-12-09
Skip duplicated endpoint option test with bluetooth

bernardnormier:
in C++ and Java

### DIFF
--- a/cpp/test/Ice/proxy/AllTests.cpp
+++ b/cpp/test/Ice/proxy/AllTests.cpp
@@ -417,21 +417,24 @@ allTests(TestHelper* helper)
     //
     // Test proxies with duplicated endpoint option.
     //
-    string goodBase = "hello:default -h host1 -p 4061 --sourceAddress 127.0.0.1";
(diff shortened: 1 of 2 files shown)


cpp/test/Ice/proxy/AllTests.cpp
```diff
@@ -417,21 +417,24 @@ allTests(TestHelper* helper)
     //
     // Test proxies with duplicated endpoint option.
     //
-    string goodBase = "hello:default -h host1 -p 4061 --sourceAddress 127.0.0.1";
-    test(communicator->stringToProxy(goodBase));
+    if (protocol != "bt")
```
bernardnormier, line 420: It looks like we only ever test with `bt` in C++, never `bts`.

pepone, line 420: In Java we use indexOf("bt") which also match "bts". In theory you can set "--protocol=bts" in the command line

pepone (approved):

PR #4715, opened 2025-12-09
Python controller fixes

pepone:
This PR updates the test scripts to not use loadSlice and compile the Slice definitions using IcePy.compileSlice instead.

Using loadSlice is problematic when you call it for multiple places, as we can end up with multiple definitions of the same classes.

### DIFF
--- a/scripts/Controller.py
+++ b/scripts/Controller.py
@@ -12,12 +12,12 @@ from Util import (
     IceProcess,
     Mapping,
     Result,
+    compileControllerDefinitions,
     parseOptions,
     platform,
     runTests,
     toplevel,
     traceback,
-    compileControllerDefinitions,
 )
 
 


scripts/Util.py
```diff
@@ -27,6 +27,19 @@
 toplevel = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 
 
+def compileControllerDefinitions():
```
bernardnormier, line 30: Should we recommend this pattern as an alternative to loadSlice?

If yes, would be good for compileSlice to be in Ice not IcePy.

bernardnormier (approved):

scripts/Util.py
pepone: Yes I think we should make it public, I will do in a separate PR.

externl (approved):

PR #4716, opened 2025-12-09
Update Bluetooth setup for Android test controller

pepone:
This PR reworks the Bluetooth setup in Android test controller to work with recent Android versions.

### DIFF
--- a/java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerActivity.java
+++ b/java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerActivity.java
@@ -46,6 +46,9 @@ public class ControllerActivity extends Activity {
             throw new IllegalStateException("Layout must include a View with android:id=\"@+id/outputList\"");
         }
 
+        _outputAdapter = new ArrayAdapter<>(this, android.R.layout.simple_list_item_1, _output);
+        _outputListView.setAdapter(_outputAdapter);
+
         WifiManager wifiManager = (WifiManager) getApplicationContext().getSystemService(Context.WIFI_SERVICE);
         _multicastLock = wifiManager.createMulticastLock("com.zeroc.testcontroller");
         _multicastLock.acquire();
@@ -65,6 +68,13 @@ public class ControllerActivity extends Activity {
         super.onRequestPermissionsResult(requestCode, permissions, grantResults);
 
         if (requestCode == REQ_CODE_BT_PERMISSIONS) {
+
+            if (grantResults.length == 0) {
+                showBluetoothError(R.string.no_bluetooth);
+                completeSetup(false);
+                return;
+            }
+
             boolean allGranted = true;
             for (int result : grantResults) {
                 if (result != PackageManager.PERMISSION_GRANTED) {
@@ -76,7 +86,7 @@ public class ControllerActivity extends Activity {
             if (allGranted) {
                 enableBluetoothIfNeeded();
             } else {
-                showBluetoothError(R.string.no_bluetooth);
+                showBluetoothError(R.string.bluetooth_permission_denied);
                 completeSetup(false);
             }
         }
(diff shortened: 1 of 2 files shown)


java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerActivity.java
```diff
@@ -46,23 +54,50 @@ public void onCreate(Bundle savedInstanceState) {
...
+                }
+            }
+            
+            if (allGranted) {
+                enableBluetoothIfNeeded();
+            } else {
+                showBluetoothError(R.string.no_bluetooth);
+                completeSetup(false);
+            }
+        }
```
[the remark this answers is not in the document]
pepone: This seems correct, as per https://developer.android.com/reference/androidx/core/app/ActivityCompat.OnRequestPermissionsResultCallback

java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerActivity.java
```diff
@@ -143,8 +171,61 @@ public void onNothingSelected(AdapterView<?> arg0) {}
...
+        if (!neededPermissions.isEmpty()) {
+            ActivityCompat.requestPermissions(
+                    this,
+                    neededPermissions.toArray(new String[0]),
+                    REQ_CODE_BT_PERMISSIONS
+            );
+        } else {
+            enableBluetoothIfNeeded();
+        }
+    }
```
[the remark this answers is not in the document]
pepone: the min sdk is set to 34, no  need for this check.

java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerActivity.java
```diff
@@ -46,23 +57,59 @@ public void onCreate(Bundle savedInstanceState) {
...
+                showBluetoothError(R.string.no_bluetooth);
+                completeSetup(false);
+                return;
+            }
+
+            boolean allGranted = true;
+            for (int result : grantResults) {
+                if (result != PackageManager.PERMISSION_GRANTED) {
+                    allGranted = false;
+                    break;
```
externl, line 82: This logic seems confusing to me. We only ask for one permission? Why not set allGranted to false, then check for `PackageManager.PERMISSION_GRANTED`. Does java have some kind of `has/filter/find` function for lists?

java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerActivity.java
```diff
@@ -16,17 +16,25 @@
...
+import androidx.core.app.ActivityCompat;
+import androidx.core.content.ContextCompat;
+
 import java.util.LinkedList;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
 public class ControllerActivity extends Activity {
+    private static final int REQ_CODE_BT_PERMISSIONS = 1001;
+    private static final int REQ_CODE_BT_ENABLE = 1002;
```
InsertCreativityHere, line 31: Are these arbitrary numbers we picked, or do they have some specific meaning?

java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerActivity.java
InsertCreativityHere: We can end up asking for either 1 or 2 permissions, depending on whether they're already allowed or not.

But if _any_ of the permissions we're granted, than we consider the whole thing failed.

----

For better or worse, the Android API really prefers to use native arrays, not `List` types. And the API for Java Arrays is pretty minimal. If this was a list, then we probably could write this simpler, but it's not.

pepone: For 3.8.1 we should port the controller to kotlin

PR #4717, opened 2025-12-09
Fix prune artifact to not use a subshell

pepone:
This is an attempt to fix https://github.com/zeroc-ice/ice/actions/runs/20062708096/job/57548655148

### DIFF
--- a/.github/workflows/prune-nightly-artifacts.yml
+++ b/.github/workflows/prune-nightly-artifacts.yml
@@ -20,5 +20,3 @@ jobs:
             run: |
               chmod +x ./packaging/release/prune-nightly-artifacts.sh
               ./packaging/release/prune-nightly-artifacts.sh
-            # Allow the release to proceed even if pruning fails
-            continue-on-error: true


.github/workflows/prune-nightly-artifacts.yml
```diff
@@ -20,3 +20,5 @@ jobs:
             run: |
               chmod +x ./packaging/release/prune-nightly-artifacts.sh
               ./packaging/release/prune-nightly-artifacts.sh
+            # Allow the release to proceed even if pruning fails
```
externl: Can this commonly fail? Would be nice to add a notification of some kind in the event of a failure so that we don't this failing. 🤷‍♂️ 

Since the script already handles `aws rm` failures, why not let this fail?

pepone: fair enough, removed.

pepone: I still get the same error after merging this PR.

The issues was the use of post-increment "((kept++))" in combination with "set -e", the first time the arithmetic expression is evaluated it returns 0, and exist status is 1. 

First time I got bite this in bash. Using pre-increment fixes the problem.

Trap the error helped debug the real cause:

```
trap 'echo "🔥 ERROR: command \"$BASH_COMMAND\" exited with code $?" >&2' ERR
```

PR #4722, opened 2025-12-09
Remove IceBuilder from test project files

bernardnormier:
This PR removes all references to IceBuilder in the C++ test project files.


pepone (approved):
Looks good!

PR #4726, opened 2025-12-10
Add compileSlice public function to Ice for Python

### LINKED ISSUE #4718: Add Ice.compileSlice public method to Ice for Python
This is useful to dynamically compile Slice to Python, instead of loadSlice

https://github.com/zeroc-ice/ice/blob/54e6c6d0b9e9d5208b600da21c061b7aa05d2ed9/scripts/Util.py#L30-L40

loadSlice is problematic when you load the same file multiple times, as classes get redefined and it can result in marshall exceptions, because the `isinstance` checks fail. For example a instance create for the previous definition.

pepone:
Fix #4718

### DIFF
--- a/java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerActivity.java
+++ b/java/test/android/controller/src/main/java/com/zeroc/testcontroller/ControllerActivity.java
@@ -2,11 +2,17 @@
 
 package com.zeroc.testcontroller;
 
+import java.util.LinkedList;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import android.Manifest;
 import android.app.Activity;
 import android.bluetooth.BluetoothAdapter;
 import android.bluetooth.BluetoothManager;
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.PackageManager;
 import android.net.wifi.WifiManager;
 import android.os.Bundle;
 import android.view.View;
@@ -16,15 +22,8 @@ import android.widget.ListView;
 import android.widget.Spinner;
 import android.widget.Toast;
 
-import android.Manifest;
-import android.content.pm.PackageManager;
-
 import androidx.core.app.ActivityCompat;
 import androidx.core.content.ContextCompat;
-
-import java.util.LinkedList;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 
 public class ControllerActivity extends Activity {
     private static final int REQ_CODE_BT_PERMISSIONS = 1001;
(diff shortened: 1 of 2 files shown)


python/python/Ice/Slice.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Notes
+    -----
+    Calling :func:`loadSlice` multiple times replaces the previously generated type definitions.
+    Any existing objects, proxies, or unmarshaled instances created from the earlier definitions
+    become incompatible with the newly loaded types. As a result, attempts to marshal or unmarshal
+    values using these stale instances may fail or raise exceptions, because their types no longer
+    match the currently loaded definitions.
+    """
+    args.insert(0, "loadSlice")
```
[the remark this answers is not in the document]
pepone: Removed here, IcePy.loadSlice already adds it.

python/python/Ice/Slice.py
```diff
@@ -0,0 +1,61 @@
...
+    args : list[str]
+        The list of arguments to pass to the Slice for Python compiler.
+        The accepted arguments are identical to those supported by the `slice2py` command-line tool.
+
+    Returns
+    -------
+    int
+        The exit code returned by the Slice for Python compiler.
+    """
+    return IcePy.compileSlice(["Ice.compileSlice", *args])
```
externl, line 28: What is this first entry in the list?

externl (commented):
Right now we have a mix of plain re-exported and wrapped types. Why not just re-export these? Aside from cases where re-export isn't possible, do you see one way being preferable than the other?

python/python/Ice/Slice.py
pepone: The program name `argv[0]`

pepone:
> Right now we have a mix of plain re-exported and wrapped types. Why not just re-export these? Aside from cases where re-export isn't possible, do you see one way being preferable than the other?

Documenting this in Python code seems simpler, for compileSlice we also add argv[0] which is used as the program name.

python/python/Ice/Slice.py
externl: Having this seems very unnecessary to me. Even weirder to me is we eventually call

https://github.com/zeroc-ice/ice/blob/440c0349d17a54cefa0008bd753f706afbed9952/cpp/src/slice2py/PythonUtil.cpp#L3209-L3236

and dump errors to stderr along with the usage instructions. 

We should just remove all this stuff about programName and return the error back to he user in Python. Am I missing something?

pepone: Currently this is called by both slice2py native compiler, the slice2py Python script, and IcePy.compileSlice.

slice2py pip script just call this and so stderr output and help makes sense there.
slice2py is impleted by calling IcePy.compileSlice

We can add a separate version for use directly in Python script.

pepone:
We decided to improve loadSlice and to not add compileSlice public function.

PR #4729, opened 2025-12-10
IceBT C++ fixes

pepone:


cpp/test/Ice/facets/AllTests.cpp
```diff
@@ -63,6 +52,18 @@ allTests(Test::TestHelper* helper)
         localOAEndpoint = ostr.str();
     }
 
+    cout << "testing add facet with uuid... " << flush;
+    {
+        ObjectAdapterPtr adapter = communicator->createObjectAdapterWithEndpoints("TestAdapter2", localOAEndpoint);
```
pepone, line 57: The localOAEndpoint string has the expected format  for bt and non bt protocols.

cpp/test/Ice/proxy/AllTests.cpp
```diff
@@ -718,8 +718,8 @@ allTests(TestHelper* helper)
     test(compObj->ice_compress(true)->ice_getCompress() == optional<bool>(true));
     test(compObj->ice_compress(false)->ice_getCompress() == optional<bool>(false));
 
-    Ice::LocatorPrx loc1(communicator, "loc1:default -p 10000");
```
pepone, line 721: We can avoid default here, as this proxies are not valid when default == bt

cpp/src/IceBT/Engine.cpp
```diff
@@ -1209,6 +1209,12 @@ namespace IceBT
 
                 if (p != _connectThreads.end()) // May be missing if destroy() was called.
                 {
+                    // We are about to remove the current thread from the list, so detach it.
+                    // This prevents std::terminate() from being called when the thread object is destroyed.
+                    if (p->joinable())
+                    {
+                        p->detach();
+                    }
```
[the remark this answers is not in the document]
bernardnormier: The only caller of this function passes this_thread::get_id() as the threadId.

We should remove this threadId parameter and simply compute the currenThreadId in this function to make the code clearer.

externl (approved):

PR #4733, opened 2025-12-10
Load slice improvements

pepone:
This PR improves loadSlice to avoid redefining types, when there are multiple loadSlice calls.

We always run Slice compiler for all the given Slice files, this return a list of code fragments each correspond to a Python module or package. We only compile the new code fragments.

### DIFF
--- a/scripts/Controller.py
+++ b/scripts/Controller.py
@@ -85,6 +85,7 @@ class ControllerDriver(Driver):
 
         self.initCommunicator()
         import Ice
+
         Ice.loadSlice([os.path.join(toplevel, "scripts", "Controller.ice")])
 
         from Test import Common as Test_Common


python/modules/IcePy/Slice.cpp
```diff
@@ -108,6 +115,12 @@ IcePy_loadSlice(PyObject* /*self*/, PyObject* args)
 
     for (const auto& fragment : compilationResult.fragments)
     {
+        // Skip if the Python module for this fragment was already loaded
+        if (loadedFragments.find(fragment.moduleName) != loadedFragments.end())
```
bernardnormier, line 119: A consequence of this update (+ insert on line 178) is if two or more fragments in the compilationResult have the same moduleName, we now skip the fragments number 2 and up.

If each moduleName is expected to be unique within a compilationResult, it would be nice to add a comment on moduleName to that effect.

pepone: We have

```
/// The Python module for the generated code.
std::string moduleName;
```

As per python definition of module

>A module is a file containing Python definitions and statements. The file name is the module name with the suffix .py appended

Because we generate a Python module for each Slice type in a given Slice file, the Python module names are unique. And the python module name is derived from the Slice scoped name.

I will upgrade the `moduleName` comment to make things clear.

Note this moduleName corresponds to the actual .py files we generate when using compileSlice instead of loadSlice.

bernardnormier: For example, the module name for Slice interface `VisitorCenter::Greeter` is `VisitorCenter/Greeter`?

pepone: `VisitorCenter.Greeter`

externl (approved):

InsertCreativityHere (approved):

PR #4744, opened 2025-12-11
Fix setting of SliceTools assembly path

pepone:


bernardnormier (approved):
Is the inconsistency between C++ and C# (IceSliceToolsAssembly vs SliceToolsAssembly) on purpose?

pepone:
> Is the inconsistency between C++ and C# (IceSliceToolsAssembly vs SliceToolsAssembly) on purpose?

No it is not. I will fix in a separate PR.

PR #4746, opened 2025-12-11
Improve Ice.loadSlice documentation comments

pepone:


python/modules/IcePy/Init.cpp
```diff
@@ -122,31 +122,48 @@ logger : Ice.Logger
...
+
+This function does not generate any Python source files. Instead, the generated Python code is loaded directly into the
+running interpreter.
+
+This function does not generate any code for Slice files included by the Slice files being loaded. It is the caller's
+responsibility to load all necessary Slice definitions. This can be done in a single call to :func:`loadSlice` by
+providing all Slice files (including included files) in the `args` parameter, or by making multiple calls to
+:func:`loadSlice`.
+
+When :func:`loadSlice` is called multiple times, the definitions that are already loaded are reused and only new
```
bernardnormier: What about:
```
When :func:`loadSlice` is called multiple times with the same Slice file, the corresponding Python code is not regenerated nor reloaded. 
```

and overlook the situation where you add new definitions to that Slice file between reload ... ultra unusual.

python/modules/IcePy/Init.cpp
```diff
@@ -122,31 +122,48 @@ logger : Ice.Logger
 
     constexpr const char* IcePy_loadSlice_doc = R"(loadSlice(args: list[str]) -> None
 
-Dynamically loads Slice definitions from one or more files and generates the corresponding Python code.
+Compiles Slice definitions and loads the generated code directly into the current Python environment.
+
+This function does not generate any Python source files. Instead, the generated Python code is loaded directly into the
+running interpreter.
```
externl, line 128:
```suggestion
Compiles Slice definitions and loads the generated code directly into the current Python environment. This function does not generate any Python source files.
```

python/modules/IcePy/Init.cpp
```diff
@@ -122,31 +122,48 @@ logger : Ice.Logger
...
+
+This function does not generate any code for Slice files included by the Slice files being loaded. It is the caller's
+responsibility to load all necessary Slice definitions. This can be done in a single call to :func:`loadSlice` by
+providing all Slice files (including included files) in the `args` parameter, or by making multiple calls to
+:func:`loadSlice`.
+
+When :func:`loadSlice` is called multiple times, the definitions that are already loaded are reused and only new
+definitions are compiled.
```
externl: Does order matter?

python/modules/IcePy/Init.cpp
```diff
@@ -122,31 +122,48 @@ logger : Ice.Logger
 
     constexpr const char* IcePy_loadSlice_doc = R"(loadSlice(args: list[str]) -> None
 
-Dynamically loads Slice definitions from one or more files and generates the corresponding Python code.
+Compiles Slice definitions and loads the generated code directly into the current Python environment.
+
+This function does not generate any Python source files. Instead, the generated Python code is loaded directly into the
+running interpreter.
+
+This function does not generate any code for Slice files included by the Slice files being loaded. It is the caller's
```
bernardnormier, line 130: I'd prefer something like:
```
This function does not generate any code for Slice files included by the Slice files listed in args.
```

python/modules/IcePy/Init.cpp
```diff
@@ -122,31 +122,48 @@ logger : Ice.Logger
...
+definitions are compiled.
 
 Parameters
 ----------
 args : list[str]
-    List of command-line arguments for Slice compilation, following
-    the same syntax as the slice2py compiler.)";
+    The list of command-line arguments to pass to the Slice loader.
+    The accepted arguments are the same as those supported by the `slice2py` tool, except that
+    the `--output-dir`, `--depend`, `--depend-xml`, and `--depend-file` options are ignored.
```
externl: Why should throw an exception if these are passed.

bernardnormier (approved):

python/modules/IcePy/Init.cpp
pepone: No it does not. You only need to ensure they are loaded before importing types that require them

python/modules/IcePy/Init.cpp
pepone: We can do in a separate PR, this PR was just about improving the documentation.

python/modules/IcePy/Init.cpp
pepone: Fine but I will just say `not reloaded`

python/modules/IcePy/Init.cpp
InsertCreativityHere: Honestly, I kind of prefer having the extra sentence. Otherwise, it sounds kind of contradictory to someone less familiar with all of this.

> Compiles Slice definitions and loads the generated code ...

> This function dos not generate any Python source files.

Well, does it generate code or not?
The extra sentence saying it's generated and directly loaded into the interpreter clears this up.

InsertCreativityHere (approved):

python/modules/IcePy/Init.cpp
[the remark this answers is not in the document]
pepone: Actually copilot is right.

pepone: We already do, fixed the doc comment.

externl (approved):

PR #4751, opened 2025-12-12
Improve Python API reference

pepone:
This PR updates Sphinx configuration to generate a page per type, only for types exported in the packages. For example the public types listed in `Ice/__init__.py` `__all__`.

It excludes internal names staring with `_`. The description for top-level packages need improvement.

### DIFF
--- a/python/.gitignore
+++ b/python/.gitignore
@@ -2,3 +2,10 @@ dist
 docs/_build
 docs/.venv
 docs/api
+
+docs/Glacier2.*.rst
+docs/Ice.*.rst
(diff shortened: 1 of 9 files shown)


bernardnormier:
> In general I'd prefer to just do the default thing the documentation does. These changes generally are just more work to maintain as sphinx changes over time.
> 
> * Does this break the ability to link to our site with `intersphinx`?
> * Are we still generating to `_buid/html`?
> * Does the documentation workflow need to be udpated?

The current presentation, with one huge page for each Slice module, is certainly not good.

Issue #4661, opened 2025-11-25
Update C# doc-comments


bernardnormier:
We need to synchronize the C# doc-comments with C++ and Java.

PR #4752, opened 2025-12-12
Improve C# Doc-Comments to Match C++ and Java

InsertCreativityHere:
This PR implements #4661 by updating and improving the C# doc-comments,
and to bring them inline with the C++ and Java doc-comments.

It also makes some smaller fixes to C++ and Java where there were typos, or where the C# doc-comment had something I felt should be 'back-ported' to the other language's comments.

### DIFF
--- a/csharp/src/Ice/ObjectAdapter.cs
+++ b/csharp/src/Ice/ObjectAdapter.cs
@@ -519,22 +519,22 @@ public sealed class ObjectAdapter
     }
 
     /// <summary>
-    /// Remove a servant (that is, the default facet) from the object adapter's Active Servant Map.
+    /// Removes a servant from the object adapter's Active Servant Map.
     /// </summary>
-    /// <param name="id">The identity of the Ice object that is implemented by the servant. If the servant implements multiple
-    /// Ice objects, remove has to be called for all those Ice objects. Removing an identity that is not in
-    /// the map throws NotRegisteredException.</param>
+    /// <param name="id">The identity of the Ice object that is implemented by the servant.</param>
     /// <returns>The removed servant.</returns>
+    /// <exception cref="NotRegisteredException">Thrown when no servant with the given identity is registered.
+    /// </exception>
     public Object remove(Identity id) => removeFacet(id, "");
 
     /// <summary>
-    /// Like remove, but with a facet.
-    /// Calling remove(id) is equivalent to calling
-    /// removeFacet with an empty facet.
+    /// Removes a servant from the object adapter's Active Servant Map, while specifying a facet.
     /// </summary>
     /// <param name="id">The identity of the Ice object that is implemented by the servant.</param>
     /// <param name="facet">The facet. An empty facet means the default facet.</param>
     /// <returns>The removed servant.</returns>
+    /// <exception cref="NotRegisteredException">Thrown when no servant with the given identity and facet is
+    /// registered.</exception>
     public Object removeFacet(Identity id, string facet)
     {
         lock (_mutex)
@@ -547,13 +547,13 @@ public sealed class ObjectAdapter
     }
 
     /// <summary>
-    /// Remove all facets with the given identity from the Active Servant Map.
-    /// The operation completely removes the Ice
-    /// object, including its default facet. Removing an identity that is not in the map throws
-    /// NotRegisteredException.
+    /// Removes all facets with the given identity from the Active Servant Map. This method completely removes the
+    /// Ice object, including its default facet.
     /// </summary>
     /// <param name="id">The identity of the Ice object to be removed.</param>
     /// <returns>A collection containing all the facet names and servants of the removed Ice object.</returns>
+    /// <exception cref="NotRegisteredException">Thrown when no servant with the given identity is registered.
+    /// </exception>
     public Dictionary<string, Object> removeAllFacets(Identity id)
     {
         lock (_mutex)
@@ -566,12 +566,12 @@ public sealed class ObjectAdapter
     }
 
     /// <summary>
-    /// Remove the default servant for a specific category.
-    /// Attempting to remove a default servant for a category that
-    /// is not registered throws NotRegisteredException.
+    /// Removes the default servant for a specific category.
     /// </summary>
     /// <param name="category">The category of the default servant to remove.</param>
     /// <returns>The default servant.</returns>
+    /// <exception cref="NotRegisteredException">Thrown when no default servant is registered for the given category.
+    /// </exception>
     public Object removeDefaultServant(string category)
     {
         lock (_mutex)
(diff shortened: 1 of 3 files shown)


csharp/src/Ice/Communicator.cs
```diff
@@ -90,16 +106,17 @@ public ValueTask DisposeAsync()
     }
 
     /// <summary>
-    /// Destroys the communicator. This method calls <see cref="shutdown" /> implicitly. Calling destroy destroys all
+    /// Destroys this communicator. This method calls <see cref="shutdown" /> implicitly. Calling destroy destroys all
```
bernardnormier: I think it should be documented as "An alias for Dispose". And then all the details (like it calls shutdown) should go to the Dispose description. Same for Java Communicator destroy/close if that's not already done.

The vast majority of Ice C# applications should not call `destroy` or `Dispose` explicitly but through using, or better, await using.

I would also avoid linking to destroy in other doc-comments. Link to Dispose or DisposeAsync instead.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -75,9 +75,7 @@ public Communicator(String[] args) {
     }
 
     /**
-     * Destroys the communicator.
-     *
-     * @see #destroy
+     * Closes this communicator. It's an alias for {@link #destroy}.
      */
```
bernardnormier, line 83: See C# comment

csharp/src/Ice/ObjectAdapter.cs
```diff
@@ -595,7 +631,8 @@ public Dictionary<string, Object> findAllFacets(Identity id)
     /// <summary>
     /// Look up a servant in this object adapter's Active Servant Map, given a proxy.
     /// This operation only tries to lookup a servant in the Active Servant Map. It does not attempt to find a servant
-    /// by using any installed ServantLocator.</summary>
+    /// by using any installed ServantLocator.
```
bernardnormier: Lots of "s" missing for in the summaries (Removes, Looks up, Finds...)

csharp/src/Ice/Connection.cs
```diff
@@ -119,34 +127,33 @@ System.Threading.Tasks.Task flushBatchRequestsAsync(
...
 {
     /// <summary>
-    /// The underlying connection information.
+    /// The information of the underlying transport or null if there's no underlying transport.
     /// </summary>
     public readonly ConnectionInfo? underlying;
 
     /// <summary>
-    /// Whether the connection is an incoming connection (<c>true</c>) or an outgoing connection (<c>false</c>).
+    /// <see langword="true"/> if this an incoming connection, <see langword="false"/> otherwise.
```
pepone: Missing `is`

```suggestion
    /// <see langword="true"/> if this is an incoming connection, <see langword="false"/> otherwise.
```

csharp/src/Ice/Object.cs
```diff
@@ -156,8 +152,19 @@ public ValueTask<OutgoingResponse> dispatchAsync(IncomingRequest request)
...
+/// <remarks>This class is provided for backward compatibility. You should consider deriving directly from
+/// <see cref="Object"/> and overriding the <see cref="Object.dispatchAsync"/> method.</remarks>
 public abstract class BlobjectAsync : Object
 {
+    /// <summary>
+    /// Dispatches an incoming request.
+    /// </summary>
+    /// <param name="inEncaps">An encapsulation containing the encoded in-parameters for the operation.</param>
+    /// <param name="current">The Current object of the incoming request.</param>
+    /// <returns>A task that will complete an instance of <see cref="Object_Ice_invokeResult"/>.</returns>
```
pepone: this sounds odd, should it be `complete with an instance ...`?

csharp/src/Ice/OutgoingResponse.cs
```diff
@@ -15,7 +15,7 @@ public sealed class OutgoingResponse
     /// Gets the exception ID of the response.
     /// </summary>
     /// <value>The exception ID of the response. It's null when replyStatus is <see cref="ReplyStatus.Ok" /> or
-    /// <see cref="ReplyStatus.UserException"/> . Otherwise, this ID is the value returned by
+    /// <see cref="ReplyStatus.UserException" /> . Otherwise, this ID is the value returned by
```
pepone:
```suggestion
    /// <see cref="ReplyStatus.UserException" />. Otherwise, this ID is the value returned by
```

csharp/src/Ice/Proxy.cs
```diff
@@ -499,7 +510,7 @@ public abstract class ObjectPrxHelperBase : ObjectPrx
     /// respects, that is, if their object identity, endpoints timeout settings, and so on are all equal.
     /// </summary>
     /// <param name="other">The proxy to compare this proxy with.</param>
-    /// <returns>True if this proxy is equal to r; false, otherwise.</returns>
+    /// <returns><see langword="true"/> if this proxy is equal to r; <see langword="false"/> otherwise.</returns>
```
pepone:
```suggestion
    /// <returns><see langword="true"/> if this proxy is equal to <paramref name="other"/>; <see langword="false"/> otherwise.</returns>
```

csharp/src/Ice/Proxy.cs
```diff
@@ -1310,7 +1330,7 @@ public ObjectPrx ice_compress(bool compress)
     /// Obtains the compression override setting of this proxy.
     /// </summary>
     /// <returns>The compression override setting. If no optional value is present, no override is
```
pepone, line 1333: I find this method comment hard to understand

bernardnormier: It would probably written before we updated the C# mapping to use `?`. And not-set is now simply null.

csharp/src/Ice/ObjectAdapter.cs
```diff
@@ -432,45 +459,54 @@ public ObjectPrx addFacet(Object servant, Identity identity, string facet)
...
+    /// <list type="bullet">
+    /// <item>The object adapter tries to find a servant for the identity and facet in the Active Servant Map.</item>
+    /// <item>If this fails, the object adapter tries to find a default servant for the category component of the
+    /// identity.</item>
+    /// <item>If this fails, the object adapter tries to find a default servant for the empty category, regardless of
+    /// the category contained in the identity.</item>
+    /// <item>If this fails, the object adapter tries to find a servant locator for the category component of the
+    /// identity. If there is no such servant locator, the object adapter tries to find a servant locator for the
+    /// empty category.
+    /// <list type="bullet">
```
externl, line 496: Is this nested list intentional?

csharp/src/Ice/Object.cs
InsertCreativityHere: Yes, fixed!

csharp/src/Ice/Connection.cs
InsertCreativityHere: Fixed.

csharp/src/Ice/Communicator.cs
InsertCreativityHere: Updated both and Java and C# like this.
Linked to Dispose instead of DisposeAsync, just because it's a simpler doc-comment.

csharp/src/Ice/ObjectAdapter.cs
InsertCreativityHere: The list is intentional, it shouldn't be nested though, just one level.
Am I missing something about C# here?

csharp/src/Ice/Proxy.cs
InsertCreativityHere: Updated to `The compression override setting. If null is returned, no override is set. Otherwise,`...

csharp/src/Ice/ObjectAdapter.cs
externl: Is this  list not nested inside the outer one?
```
/// <list type="bullet">
    /// <item>The object adapter tries to find a servant for the identity and facet in the Active Servant Map.</item>
    /// <item>If this fails, the object adapter tries to find a default servant for the category component of the
    /// identity.</item>
    /// <item>If this fails, the object adapter tries to find a default servant for the empty category, regardless of
    /// the category contained in the identity.</item>
    /// <item>If this fails, the object adapter tries to find a servant locator for the category component of the
    /// identity. If there is no such servant locator, the object adapter tries to find a servant locator for the
    /// empty category.
    /// <list type="bullet">Collapse comment
```

InsertCreativityHere: Ah, I was looking at the list above this one.
This is intentional, to match the list we have in C++:
```
...
        ///  - If this fails, the object adapter tries to find a servant locator for the category component of the
        ///    identity. If there is no such servant locator, the object adapter tries to find a servant locator for the
        ///    empty category.
        ///    - If a servant locator is found, the object adapter tries to find a servant using this servant locator.
        ///  - If all the previous steps fail, the object adapter gives up and the caller receives an
        ///    ObjectNotExistException or a FacetNotExistException.
...
```
Could probably just dedent the tick-mark and not bother with a nested list though?

PR #4783, opened 2025-12-18
Enable C# deterministic builds

### LINKED ISSUE #4772: Generate deterministic DLLs in C#
See https://nuget.info/packages/ZeroC.IceStorm/3.8.0

and
https://github.com/icerpc/icerpc-csharp/blob/3f804147f98f20266e4a2ad456be64caa99cfa1c/src/Directory.Build.props#L9

https://github.com/icerpc/icerpc-csharp/blob/3f804147f98f20266e4a2ad456be64caa99cfa1c/build/IceRpc.Src.props#L9

pepone:
Fix #4772

### DIFF
--- a/csharp/src/Directory.Build.props
+++ b/csharp/src/Directory.Build.props
@@ -18,7 +18,7 @@
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
     </PropertyGroup>
 
-    <PropertyGroup Condition="'$(GITHUB_ACTIONS)' == 'true'">
+    <PropertyGroup Condition="'$(CI)' == 'true'">
         <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
     </PropertyGroup>
     <Import Project="$(MSBuildThisFileDirectory)..\tools\ZeroC.Ice.Slice.Tools\ZeroC.Ice.Slice.Tools.props"/>


pepone:
The whitespace check should ignore `packages.lock.json` files; I think it’s better to keep them as generated by .NET.

csharp/src/Directory.Build.props
```diff
@@ -10,10 +10,17 @@
...
         <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+        <!-- Enable package lock file to ensure consistent package versions across builds -->
+        <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
+        <RestoreLockedMode>true</RestoreLockedMode>
         <!-- TODO: move to common.props and fix analyzer warnings in tests -->
         <AnalysisMode>All</AnalysisMode>
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
     </PropertyGroup>
+
+    <PropertyGroup Condition="'$(GITHUB_ACTIONS)' == 'true'">
```
externl: Should check for the CI variable instead? IIRC it's the common variable set by CI runners. Which is useful if anyone wants to build their own packages outside of GitHub.

pepone:
```suggestion
    <PropertyGroup Condition="'$(CI)' == 'true'">
```

PR #4786, opened 2025-12-18
Fix NuGet packaging props

### LINKED ISSUE #4773: Add description to NuGet packages (C#)
See for example:
https://github.com/icerpc/icerpc-csharp/blob/3f804147f98f20266e4a2ad456be64caa99cfa1c/src/IceRpc.Transports.Quic/IceRpc.Transports.Quic.csproj#L4

bernardnormier:
This PR adds a description to all NuGet packages, and more.

Fixes #4773.

### DIFF
--- a/csharp/msbuild/ice.common.props
+++ b/csharp/msbuild/ice.common.props
@@ -11,8 +11,6 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <ImplicitUsings>true</ImplicitUsings>
     <LangVersion>12.0</LangVersion>
-    <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <RepositoryUrl>https://github.com/zeroc-ice/ice</RepositoryUrl>
     <RollForward>Major</RollForward>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningLevel>4</WarningLevel>
@@ -33,6 +31,7 @@
     <Product>Ice</Product>
     <RepositoryType>git</RepositoryType>
     <RepositoryUrl>https://github.com/zeroc-ice/ice</RepositoryUrl>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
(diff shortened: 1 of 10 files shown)


csharp/src/Glacier2/Glacier2.csproj
```diff
@@ -3,13 +3,12 @@
   <Import Project="../../msbuild/ice.common.props" />
   <PropertyGroup>
     <AssemblyName>Glacier2</AssemblyName>
+    <Description>Client library for the Glacier2 service</Description>
```
bernardnormier, line 6: Since we have separate logos for Glacier2, IceGrid etc., we should use them for these packages (for a follow-up PR).

See for example: https://www.nuget.org/packages/ZeroC.Glacier2

bernardnormier, line 6: I used "Client library" even though it's more than a client library.

After researching how to call this assembly, "client library" appears to be to most common / understandable description.

csharp/msbuild/ice.common.props
```diff
@@ -21,6 +21,20 @@
     <StyleCopAnalyzersVersion>1.2.0-beta.556</StyleCopAnalyzersVersion>
   </PropertyGroup>
 
+  <!-- Common assembly and packaging properties -->
+  <PropertyGroup>
+    <Authors>ZeroC, Inc.</Authors>
```
bernardnormier, line 24: Essentially the same as in icerpc-csharp.

csharp/msbuild/ice.common.props
```diff
@@ -21,6 +21,20 @@
...
+    <Copyright>Copyright (c) ZeroC, Inc.</Copyright>
+    <PackageIcon>logo-128x128.png</PackageIcon>
+    <PackageLicenseExpression>GPL-2.0-only</PackageLicenseExpression>
+    <PackageProjectUrl>https://zeroc.com/zeroc-ice</PackageProjectUrl>
+    <PackageReadmeFile>README.md</PackageReadmeFile>
+    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
+    <Product>Ice</Product>
+    <RepositoryType>git</RepositoryType>
+    <RepositoryUrl>https://github.com/zeroc-ice/ice</RepositoryUrl>
```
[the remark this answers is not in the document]
bernardnormier: Fixed

externl (approved):

pepone (approved):

PR #4792, opened 2025-12-19
Ice 3.9 Linux package updagtes

pepone:
This PR contains a few fixes for the Linux packages, related to the 3.9 version update on main.


packaging/deb/debian/control
```diff
@@ -244,33 +244,3 @@ Description: IceBridge service
  with minimal effort. Ice takes care of all interactions with low-level
  network programming interfaces and allows you to focus your efforts on
  your application logic.
-
```
pepone, line 247: This were just for updating from 3.7, and don't need to keep them around for 3.9.

PR #4799, opened 2025-12-23
Fix macOS C++ build failures

pepone:
This PR fixes C++ macOS build failures on 3.7 branch

### DIFF
--- a/cpp/include/Ice/Proxy.h
+++ b/cpp/include/Ice/Proxy.h
@@ -1472,12 +1472,16 @@ ICE_API bool proxyIdentityAndFacetLess(const ::std::shared_ptr<ObjectPrx>& lhs,
 ICE_API bool proxyIdentityAndFacetEqual(const ::std::shared_ptr<ObjectPrx>& lhs,
                                         const ::std::shared_ptr<ObjectPrx>& rhs);
 
+#ifdef __clang__
+#   pragma clang diagnostic push
+#   pragma clang diagnostic ignored "-Wdeprecated-declarations" // for std::binary_function below
+#endif
+
 /**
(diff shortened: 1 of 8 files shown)


cpp/include/Ice/Proxy.h
```diff
@@ -1479,7 +1479,7 @@ ICE_API bool proxyIdentityAndFacetEqual(const ::std::shared_ptr<ObjectPrx>& lhs,
  */
 
 struct ProxyIdentityLess
-#if (ICE_CPLUSPLUS < 201703L)
+#if (ICE_CPLUSPLUS < 201103L)
```
bernardnormier: I think it would be cleaner from an source compatibility point of view to suppress the warning, instead of changing the API. See #4800.

cpp/src/IceDB/IceDB.cpp
```diff
@@ -322,10 +322,6 @@ DbiBase::DbiBase() :
 {
 }
 
-DbiBase::~DbiBase()
```
bernardnormier, line 325: That's fine because it's all internal, even though IceDB APIs are "exported" to other services.

cpp/src/IceGrid/DescriptorHelper.h
```diff
@@ -144,6 +143,7 @@ class ServerHelper : public CommunicatorHelper, public IceUtil::SimpleShared
 
     ServerHelper(const ServerDescriptorPtr&, bool = false);
     ServerHelper() { }
+    virtual ~ServerHelper() { }
```
bernardnormier: Same here, that's fine because it's all internal.

cpp/include/Ice/Proxy.h
pepone: I will rebase after you merge #4800

PR #4809, opened 2025-12-26
Build additional languages with CI

pepone:
This PR updates CI workflow on the 3.7 branch to test additional language mappings, and fixes some build issues.

The job is based on what we have in main, but there is no cache support. I was initially thinking of adding it in a follow up PR. We are already hitting the cache limit with our main and 3.8 builds so there is no point in adding more entries to the cache.

### DIFF
--- a/.github/actions/setup-dotnet-framework/action.yml
+++ b/.github/actions/setup-dotnet-framework/action.yml
@@ -1,0 +1,21 @@
+name: Setup .NET Framework
+description: "Install .NET Framework Developer/Targeting Packs for building .NET Framework projects"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install .NET Framework 4.5.1 Developer Pack
+      if: runner.os == 'Windows'
+      shell: powershell
+      run: |
+        # Install .NET Framework 4.5.1 Developer Pack using Chocolatey (pre-installed on GitHub runners)
+        Write-Host "Installing .NET Framework 4.5.1 Developer Pack via Chocolatey..."
+        choco install netfx-4.5.1-devpack -y --no-progress
+
+        # Verify the targeting pack is installed
+        $targetingPackPath = "${env:ProgramFiles(x86)}\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5.1"
+        if (Test-Path $targetingPackPath) {
+            Write-Host "Targeting pack verified at: $targetingPackPath"
+        } else {
+            throw "Targeting pack not found at expected location: $targetingPackPath"
+        }
--- a/csharp/BUILDING.md
+++ b/csharp/BUILDING.md
@@ -28,8 +28,8 @@ A source build of Ice for .NET on Windows produces two sets of assemblies:
 In order to build Ice for .NET from source, you need:
 
 * A [supported version][3] of Visual Studio when building .NET Framework 4.5 Assemblies.
-* Visual Studio 2022 with [.NET 6.0 SDK][4] or [.NET 7.0 SDK][5] when building the .NET Standard 2.0 Assemblies.
-* Visual Studio 2022 with [.NET 6.0 SDK][4] and [.NET 7.0 SDK][5] when building the NuGet packages.
+* Visual Studio 2022 with [.NET 8.0 SDK][4] or [.NET 7.0 SDK][5] when building the .NET Standard 2.0 Assemblies.
+* Visual Studio 2022 with [.NET 8.0 SDK][4] and [.NET 7.0 SDK][5] when building the NuGet packages.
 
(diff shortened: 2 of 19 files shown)


.github/actions/setup-dotnet-framework/action.yml
```diff
@@ -0,0 +1,21 @@
+name: Setup .NET Framework
+description: "Install .NET Framework Developer/Targeting Packs for building .NET Framework projects"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install .NET Framework 4.5.1 Developer Pack
```
pepone, line 7: This is required to build .NET on Windows

csharp/msbuild/ice.common.props
```diff
@@ -25,6 +25,9 @@
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <RepositoryUrl>https://github.com/zeroc-ice/ice</RepositoryUrl>
     <ProduceReferenceAssembly>false</ProduceReferenceAssembly>
+
+    <!-- Avoid warnings about end-of-line target framework -->
```
pepone, line 29: I initially added this to avoid warnings when targeting .NET 6 with the test suite. I think is best to target .NET 8/.NET 10.

So we can revert this.

csharp/BUILDING.md
```diff
@@ -28,8 +28,8 @@ A source build of Ice for .NET on Windows produces two sets of assemblies:
 In order to build Ice for .NET from source, you need:
 
 * A [supported version][3] of Visual Studio when building .NET Framework 4.5 Assemblies.
-* Visual Studio 2022 with [.NET 6.0 SDK][4] or [.NET 7.0 SDK][5] when building the .NET Standard 2.0 Assemblies.
-* Visual Studio 2022 with [.NET 6.0 SDK][4] and [.NET 7.0 SDK][5] when building the NuGet packages.
+* Visual Studio 2022 with [.NET 8.0 SDK][4] or [.NET 7.0 SDK][5] when building the .NET Standard 2.0 Assemblies.
+* Visual Studio 2022 with [.NET 8.0 SDK][4] and [.NET 7.0 SDK][5] when building the NuGet packages.
```
pepone, line 32: I think we should just support .NET 8 and .NET 10 SDKs, the target frameworks are still netstandard 2.0 and .net45. Just build using a modern SDK

matlab/src/msbuild/ice/ice.vcxproj
```diff
@@ -13,7 +13,6 @@
   </ItemGroup>
   <PropertyGroup Label="Globals">
     <ProjectGuid>{89C40F1A-1761-46C1-B326-5B20BE6F8173}</ProjectGuid>
-    <WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>
```
pepone, line 16: using `WindowsTargetPlatformVersion` requires install an older SDK. Better to use the default.

scripts/Util.py
```diff
@@ -417,7 +417,7 @@ def parseBuildVariables(self, component, variables):
         pass # Nothing to do, we don't support the make build system on Windows
 
     def getDefaultBuildPlatform(self):
-        return "x64" if "X64" in os.environ.get("PLATFORM", "") else "Win32"
+        return "Win32" if "Win32" in os.environ.get("PLATFORM", "") else "x64"
```
pepone, line 420: I switched to use x64 by default.

ice.proj
```diff
@@ -29,31 +29,6 @@
       <Properties>Platform=$(Platform);Configuration=$(Configuration)</Properties>
     </Projects>
 
-    <!-- For PHP build both thread safe and non thread safe configurations -->
```
pepone, line 32: We are not going to build PHP with the top-level project `ice.proj` on Windows, PHP can still be build from the php directory. We are not shipping PHP binaries for 3.7.11 on Windows.

csharp/BUILDING.md
```diff
@@ -126,7 +126,7 @@ necessary.
 
 ### Linux and macOS Build Requirements
 
-You need the [.NET 6.0 SDK][4] or [.NET 7.0 SDK][5] to build Ice for .NET from source.
+You need the [.NET 8.0 SDK][4] or [.NET 7.0 SDK][5] to build Ice for .NET from source.
```
externl, line 129: I thought we only wanted to support the 8 and 10 SDKs. We'll do this in a follow-up?

csharp/BUILDING.md
externl, line 159: Ditto for this net7

csharp/BUILDING.md
pepone: Yes I will go over the building docs in a follow up PR

PR #4816, opened 2025-12-29
Add PromiseKit as a SPM dependency

### LINKED ISSUE #4815: Ice 3.7 swift build failures with Xcode 26
The macOS Swift build is failing because the PromiseKit dependency is not available, seems like a Carthage bug.

```
jose@studio ice % carthage bootstrap --cache-builds --use-xcframeworks --platform macOS
*** Checking out PromiseKit at "6.22.1"
*** No cache found for PromiseKit, building with all downstream dependencies
*** xcodebuild output can be found in /var/folders/06/p2qxm8jn5m7bls0lc3hmtfq00000gn/T/carthage-xcodebuild.H2Yhhx.log
*** Building scheme "PromiseKit" in PromiseKit.xcodeproj
jose@studio ice % ls Carthage/Build/PromiseKit.xcframework 
Info.plist	ios-arm64
```

https://github.com/Carthage/Carthage/issues/3399

pepone:
This PR updates the Swift build to add PromiseKit as a SPM dependency. Fix #4815.

Tested macOS and iphonesimulator builds


swift/ice.xcodeproj/project.xcworkspace/contents.xcworkspacedata
externl, line 1: This file should not be committed?

swift/test/Ice/ami/AllTests.swift
```diff
@@ -2,6 +2,7 @@
 // Copyright (c) ZeroC, Inc. All rights reserved.
 //
 
+import Foundation
```
bernardnormier, line 5: Why do we need to import Foundation in all these tests?

bernardnormier (approved):

swift/test/Ice/ami/AllTests.swift
pepone: Without this

```
swift/test/Ice/ami/AllTests.swift:177:13: error: cannot find 'Thread' in scope
            Thread.sleep(forTimeInterval: 0.01)
```

Seems this was imported transitively when using the carthage package.

Issue #4709, opened 2025-12-08
slice2py: no docstring for constants


bernardnormier:
slice2py does not emit any docstring for constants.

PR #4817, opened 2025-12-30
Improve 'slice2py' to Generate Doc-comments for Constants

InsertCreativityHere:
This PR implements most of #4709.

We now generate doc-strings on constants.
Doc-comments take the form of `#:` and come before the thing being documented.
Doc-strings take the form of `""" ... """` and come after the thing being documented.

While Sphinx has full support for both of these, the official VsCode "Python" extension only supports doc-strings.

----

I say "most of", because these comments don't appear in our generated API reference.
That is an issue with how we re-export things though, not with the code-gen.
The generated code will still have these comments, and the extension detects and displays them in VsCode.

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -2033,9 +2033,9 @@ Slice::Python::CodeVisitor::visitConst(const ConstPtr& p)
     out << sp;
 
     string name = p->mappedName();
-    writeConstDocstring(p, out);
     out << nl << name << " = ";
     writeConstantValue(p, p->type(), p->valueType(), p->value(), out);
+    writeDocstring(p, out, "constant");
 
     out << sp;
     out << nl << "__all__ = [\"" << name << "\"]";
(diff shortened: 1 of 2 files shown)


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1211,7 +1211,7 @@ Slice::Python::CodeVisitor::writeOperations(const InterfaceDefPtr& p, Output& ou
         out << epar << operationReturnTypeHint(operation, MethodKind::Dispatch) << ":";
         out.inc();
 
-        writeDocstring(operation, MethodKind::Dispatch, out);
+        writeOpDocstring(operation, MethodKind::Dispatch, out);
```
InsertCreativityHere, line 1214: We had 3 overloads all named `writeDocString`, I gave them all different names.
Doing this actually revealed a useless `optional` parameter!

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2276,7 +2278,7 @@ Slice::Python::CodeVisitor::writeSeeAlso(const StringList& seeAlso, bool needsNe
 }
 
 void
-Slice::Python::CodeVisitor::writeDocstring(const ContainedPtr& p, Output& out, const optional<string>& generatedType)
+Slice::Python::CodeVisitor::writeDocstring(const ContainedPtr& p, Output& out, string_view generatedType)
```
InsertCreativityHere: For this overload of `writeDocString`, `generated_type` was _always_ passed.
So making it a defaulted `optional` parameter is useless.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2442,7 +2436,59 @@ Slice::Python::CodeVisitor::writeDocstring(const EnumPtr& p, Output& out)
 }
 
 void
-Slice::Python::CodeVisitor::writeDocstring(const OperationPtr& op, MethodKind methodKind, Output& out)
+Slice::Python::CodeVisitor::writeConstDocstring(const ConstPtr& p, IceInternal::Output& out)
```
InsertCreativityHere: The meat of this PR.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2276,7 +2278,7 @@ Slice::Python::CodeVisitor::writeSeeAlso(const StringList& seeAlso, bool needsNe
 }
 
 void
-Slice::Python::CodeVisitor::writeDocstring(const ContainedPtr& p, Output& out, const optional<string>& generatedType)
+Slice::Python::CodeVisitor::writeDocstring(const ContainedPtr& p, Output& out, const string& generatedType)
```
InsertCreativityHere, line 2281: For this overload of `writeDocString`, `generated_type` was _always_ passed.
So having it be `optional` was pointless.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2442,7 +2436,56 @@ Slice::Python::CodeVisitor::writeDocstring(const EnumPtr& p, Output& out)
 }
 
 void
-Slice::Python::CodeVisitor::writeDocstring(const OperationPtr& op, MethodKind methodKind, Output& out)
+Slice::Python::CodeVisitor::writeConstDocstring(const ConstPtr& p, IceInternal::Output& out)
+{
+    const optional<DocComment>& comment = p->docComment();
```
externl: Do both pylance (whatever vscode uses) and sphinx play nice with the `#:` format?

InsertCreativityHere: No, apparently pylance can't handle the doc-comment format unfortunately.
Thankfully, Sphinx handles both, so I switched to use the common denominator: `"""..."""` doc-strings.

PR #4826, opened 2025-12-31
Update C# build to support require .NET 4.7.2, and .NET Standard 2.0

pepone:


.github/workflows/build-dotnet-packages.yml
pepone, line 1: A new workflow to build .NET NuGet packages base on what we do from 3.8.0

csharp/test/IceUtil/inputUtil/msbuild/client/client.csproj
```diff
@@ -0,0 +1,31 @@
...
+    <Compile Include="../../../../TestCommon/TestHelper.cs" />
+    <Compile Include="../../Client.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="zeroc.icebuilder.msbuild" Version="5.0.9" />
+  </ItemGroup>
+  <Choose>
+    <When Condition="'$(ICE_BIN_DIST)' == 'all'">
+      <ItemGroup>
+        <PackageReference Include="zeroc.ice.net" Version="3.7.10" />
```
bernardnormier, line 22: I guess we need to replace 3.7.10 by 3.7.11 everywhere, in a follow-up PR?

csharp/src/IceBox/msbuild/icebox/icebox.csproj
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="../../../../msbuild/ice.common.props" />
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <AssemblyName>iceboxnet</AssemblyName>
+    <Version>$(IceVersion)</Version>
+    <OutputPath>../../../../bin</OutputPath>
```
bernardnormier, line 8: Can we use the same iceboxnet for .NET and .NET Framework?

I thought we would need separate builds.

pepone: we build two versions just using the same project

PR #4830, opened 2025-12-31
Reenable icegrid/replication in CI

bernardnormier:


cpp/test/IceGrid/replication/Client.cpp
```diff
@@ -20,7 +20,7 @@ Client::run(int argc, char** argv)
 {
     Ice::PropertiesPtr properties = createTestProperties(argc, argv);
     properties->setProperty("Ice.Warn.Connections", "0");
-    properties->setProperty("Ice.Default.Timeout", "100");
```
bernardnormier, line 23: This timeout (100 ms) is certainly too low. Even though there was no comment here, we typically set the low timeouts to speed up tests on Windows where connection establishment failures take a long time to fail ... so we timeout instead.

cpp/test/IceGrid/replication/Client.cpp
```diff
@@ -20,7 +20,7 @@ Client::run(int argc, char** argv)
...
+    properties->setProperty("Ice.Override.ConnectTimeout", "1000");
```
bernardnormier, line 23: The 1000ms (1s) connect timeout is the new minimum in Ice 3.8.

pepone (approved):

externl (approved):

PR #4832, opened 2026-01-02
Run C++11 mapping tests with CI

pepone:


.github/workflows/ci.yml
```diff
@@ -58,3 +63,12 @@ jobs:
...
           flags: "--all"
+
+      - name: Test C++11 on ${{ matrix.os }}
+        uses: ./.github/actions/test
+        timeout-minutes: 45
+        with:
+          working_directory: cpp
+          flags: >-
+            --all
+            --config ${{ runner.os == 'Windows' && 'Cpp11-Release' || 'cpp11-shared' }}
```
bernardnormier, line 74: What does this mean?

My understanding is we want to pass `--config cpp11-shared` to run the C++11 tests on Linux and macOS. It works differently on Windows?

pepone: On Windows we have Cpp11-Release/Cpp11-Debug configurations.

We pass `--config Cpp11-Release` on Windows, `--config cpp11-shared` non windows

bernardnormier: so the above expands to `--config Cpp11-Release` on Windows and `--config cpp11-shared` elsewhere?

pepone: Yes, similar to JavaScript:

```
> osname='Windows'
'Windows'
> console.log(osname == 'Windows' && 'Cpp11-Release' || 'cpp11-shared');
Cpp11-Release
undefined
> osname='Linux'
'Linux'
> console.log(osname == 'Windows' && 'Cpp11-Release' || 'cpp11-shared');
cpp11-shared
undefined
```

PR #4834, opened 2026-01-02
Improve Enum Doc-Comment Formatting in Python

InsertCreativityHere:
In Python, our doc comments (both generated and hand-written) are inconsistent and sub-optimal.

### Comparison of the API reference pages before and after:
CompressBatch Before:
Screenshot 2026-01-02 at 14 54 56
CompressBatch After:
Screenshot 2026-01-02 at 14 55 11

----

EndpointSelectionType Before:
Screenshot 2026-01-02 at 14 55 31
EndpointSelectionType After:
Screenshot 2026-01-02 at 14 55 44

### Current Behavior

Right now, we never generate doc-comments on the enumerators themselves, instead we generate a list of enumerators on the enclosing enum. There are 2 different styles we use at random:
```
Enumerators:

    - Yes:
        Compress the batch requests.

    - No:
        Don't compress the batch requests.
```
and
```
    Enumerators
    -----------
    Random : EndpointSelectionType
        Random causes the endpoints to be arranged in a random order.
    Ordered : EndpointSelectionType
        Ordered forces the Ice runtime to use the endpoints in the order they appeared in the proxy.
```

### What this PR does

We no longer write this list of enumerators in the enum's doc-string.
Enumerator doc-comments are now placed directly on the enumerator itself:
```
class CompressBatch(Enum):
    """
    The batch compression option when flushing queued batch requests.
    """

    Yes = 0
    """
    Compress the batch requests.
    """

    No = 1
    """
    Don't compress the batch requests.
    """
```
This seems more logical, works better with the Python extension in VSCode, and looks better in the generated API reference.

### DIFF
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -2376,7 +2376,7 @@ Slice::Python::CodeVisitor::writeEnumeratorDocstring(const EnumeratorPtr& p, Out
 
     out << nl << tripleQuotes;
 
-    StringList overview = comment->overview();;
+    StringList overview = comment->overview();
     StringList remarks = comment->remarks();
     StringList seeAlso = comment->seeAlso();
 


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1989,12 +1989,13 @@ Slice::Python::CodeVisitor::visitEnum(const EnumPtr& p)
     out << nl << "class " << name << "(" << getImportAlias(p, "enum", "Enum") << "):";
     out.inc();
 
-    writeEnumDocstring(p, out);
+    writeDocstring(p, out, "enum class");
```
InsertCreativityHere, line 1992: Now that we don't generate this list of enumerators in the enum doc-comment, we don't need a special `writeDocString` method for enums.

It can just use the normal `writeDocString` that every other type uses.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1989,12 +1989,13 @@ Slice::Python::CodeVisitor::visitEnum(const EnumPtr& p)
...
 
-    out << nl;
     for (const auto& enumerator : enumerators)
     {
+        out << sp;
```
InsertCreativityHere, line 1996: We weren't generating spaces between enumerators before.
Now we are.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2334,6 +2329,13 @@ Slice::Python::CodeVisitor::writeDocstring(const ContainedPtr& p, Output& out, c
         "The Slice compiler generated this " + generatedType + " from Slice " + p->kindOf() + " ``" + p->scoped() +
         "``.");
```
InsertCreativityHere, line 2331: I just moved these lines down to where we were doing the other writes to `out`. Instead of having these mixed in with where we're gathering doc-strings.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2364,74 +2366,26 @@ Slice::Python::CodeVisitor::writeDocstring(const ContainedPtr& p, Output& out, c
...
-        seeAlso = comment->seeAlso();
-    }
-    if (!remarks.empty())
+    if (!comment)
     {
-        remarks.emplace_back(""); // empty line
+        return;
     }
-    remarks.push_back(
-        "The Slice compiler generated this enum class from Slice " + p->kindOf() + " ``" + p->scoped() + "``.");
```
InsertCreativityHere, line 2401: I decided to not generate this remark for each enumerator, otherwise it looked really noisy seeing this 4 times back-to-back.

It's still on the enum though.

bernardnormier (approved):

PR #4835, opened 2026-01-02
Sort Python API class/enum/etc Members by Source, not Alphabetically

InsertCreativityHere:
When Sphinx generates the Python reference for a class, exception, enum, etc., it's members are sorted alphabetically.
This PR changes this so that now members retain the order they have in the source file (i.e. they aren't sorted anymore).

----

With the current behavior things which are logically grouped in our source end up randomly placed in the API reference. This looks especially bad for enums where enumerators are not sorted by value, how I'd expect.

Like look at `CompressBatch`, the enumerators are completely reversed from how they are in the source:
Screenshot 2026-01-02 at 15 44 46
Or worse, `ReplyStatus`, which looks completely random:
Screenshot 2026-01-02 at 15 47 25

I haven't consumed alot of Python API references, so maybe alphabetized members is what everyone expects, and enumerator values are inconsequential. If so, feel free to reject this PR.

### DIFF
--- a/python/docs/conf.py
+++ b/python/docs/conf.py
@@ -69,6 +69,7 @@ autodoc_typehints = "both"
 autodoc_typehints_description_target = "all"
 add_module_names = True
 python_use_unqualified_type_names = False
+autodoc_member_order = "bysource"
 
 # =============================================================================
 # Intersphinx Configuration


externl (approved):

bernardnormier:
Another point of reference is the API reference for:

IceRPC C#:
https://code.icerpc.dev/csharp/0.5.x/api/api/IceRpc.StatusCode.html

Ice C#:
https://code.zeroc.com/ice/main/api/csharp/api/Ice.ReplyStatus.html

Ice C++:
https://code.zeroc.com/ice/main/api/cpp/namespaceIce_ae6f82aeb1d3c31c73700d252f091ea10.html#ae6f82aeb1d3c31c73700d252f091ea10

None of them use alphabetical sorting.

pepone (approved):

InsertCreativityHere:
> None of them use alphabetical sorting.

Yes, I just wanted to make sure that there wasn't some expectation of alphabetized docs in Python.
Different languages have their own different conventions is all.

PR #4845, opened 2026-01-07
Remove Incorrect 'virtual' Modifiers

### LINKED ISSUE #4846: Incorrect `virtual` Modifiers Applied to Logically `final` Functions
There are at least 2 classes which are final, but declare `virtual` functions.
This is at best useless, since nothing can ever override these functions.

We should remove/replace these incorrect modifiers where present.

InsertCreativityHere:
This PR fixes #4846.

### DIFF
--- a/cpp/src/IceBT/AcceptorI.cpp
+++ b/cpp/src/IceBT/AcceptorI.cpp
@@ -19,7 +19,7 @@ using namespace IceBT;
 
 namespace
 {
-    class ProfileCallbackI : public ProfileCallback
+    class ProfileCallbackI final : public ProfileCallback
     {
     public:
         ProfileCallbackI(AcceptorIPtr acceptor) : _acceptor(std::move(acceptor)) {}


cpp/src/Ice/ConnectionFactory.h
```diff
@@ -88,11 +88,11 @@ namespace IceInternal
                 std::function<void(Ice::ConnectionIPtr, bool)>,
                 std::function<void(std::exception_ptr)>);
```
InsertCreativityHere, line 90: This class is `final` and doesn't meaningfully inherit from anything:
```
class ConnectCallback final : public std::enable_shared_from_this<ConnectCallback>
```

cpp/test/Ice/operations/BatchOneways.cpp
```diff
@@ -18,7 +18,7 @@ namespace
     public:
         BatchRequestInterceptorI() = default;
```
InsertCreativityHere, line 20: This class is `final` and has no superclasses:
```
class BatchRequestInterceptorI final
```

cpp/src/Ice/ConnectionFactory.h
```diff
@@ -193,8 +193,8 @@ namespace IceInternal
         //
 
 #if defined(ICE_USE_IOCP)
```
InsertCreativityHere, line 195: This class is `final`, but inherits from `EventHandler`:
```
class IncomingConnectionFactory final : public EventHandler
```

cpp/src/IceBT/AcceptorI.cpp
```diff
@@ -19,7 +19,7 @@ using namespace IceBT;
 
 namespace
 {
-    class ProfileCallbackI : public ProfileCallback
+    class ProfileCallbackI final : public ProfileCallback
```
InsertCreativityHere, line 22: This is somewhat unrelated to the other changes, but should fix this warning:
```
/usr/lib/llvm-21/bin/../include/c++/v1/__memory/construct_at.h:61:3: error:
      destructor called on non-final '(anonymous namespace)::ProfileCallbackI' that has virtual functions
      but non-virtual destructor [-Werror,-Wdelete-non-abstract-non-virtual-dtor]
   61 |   __loc->~_Tp();
      |   ^
```

cpp/src/Ice/ConnectionFactory.h
bernardnormier: There is a high likelihood that these virtual functions should be marked final (and not virtual). Not nothing.

cpp/test/Ice/operations/BatchOneways.cpp
bernardnormier: should be final or override

PR #4852, opened 2026-01-08
Schedule Nightly builds for main, 3.8, and 3.7 branches

pepone:
This PR adds support for scheduling nightly builds for main, 3.8, and 3.7 branches. The GitHub schedule events are only dispatched to the default branch, we then dispatch a nightly build on the branch we want to build the nightlies.

### DIFF
--- a/.github/workflows/dispatch-nightly-release.yml
+++ b/.github/workflows/dispatch-nightly-release.yml
@@ -17,6 +17,9 @@ jobs:
     if: github.repository == 'zeroc-ice/ice'
     runs-on: ubuntu-24.04
     steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
       - name: Dispatch nightly build for main (2:00 UTC)
         if: github.event.schedule == '0 2 * * *'
         uses: ./.github/actions/dispatch-workflow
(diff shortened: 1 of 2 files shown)


.github/workflows/dispatch-nightly-release.yml
```diff
@@ -0,0 +1,42 @@
+name: Dispatch Nightly Release
+
+on:
+  # Central scheduler (runs from the default branch)
+  schedule:
```
externl, line 5: Schedules only work on the main branch?

pepone: Yes, they only work on the default branch.

See https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule

externl: > In a public repository, scheduled workflows are automatically disabled when no repository activity has occurred in 60 days. For information on re-enabling a disabled workflow, see [Disabling and enabling a workflow](https://docs.github.com/en/enterprise-server/actions/using-workflows/disabling-and-enabling-a-workflow#enabling-a-workflow).

This is interesting.

PR #4860, opened 2026-01-09
Fix JavaScript property validation

### LINKED ISSUE #2102: Ice property validation is broken in JavaScript
Ice property validation is likely broken in JavaScript. I haven't written a test to verify and the only way we would notice is if a warning when setting an Ice property was not emitted.

I noticed this when working on https://github.com/zeroc-ice/ice/pull/2100

Issues:
https://github.com/zeroc-ice/ice/blob/711bb2939294bb1dbb940e3a4a6966e89c5ed735/config/makeprops.py#L589
* JS regex strings don't require surrounded `/`

https://github.com/zeroc-ice/ice/blob/711bb2939294bb1dbb940e3a4a6966e89c5ed735/js/src/Ice/PropertiesI.js#L173
* This substring check is wrong

https://github.com/zeroc-ice/ice/blob/711bb2939294bb1dbb940e3a4a6966e89c5ed735/js/src/Ice/PropertiesI.js#L181
* The inner loop check should not be using the length of the element at `j`.

pepone:
Fixes #2102

### DIFF
--- a/CHANGELOG-3.7.md
+++ b/CHANGELOG-3.7.md
@@ -109,6 +109,11 @@ particular aspect of Ice.
   associated thread-pool has no thread available to process these messages. As a result, back-pressure now works in
   Ice for C# just like in other language mappings.
 
+## JavaScript Changes
+
+- Fixed a bug in the Ice for JavaScript property validation code, which resulted in incorrect property validation
+  for Ice properties. (https://github.com/zeroc-ice/ice/issues/2102)
+
 # Changes in Ice 3.7.10
 
 These are the major changes since Ice 3.7.9.


config/makeprops.py
```diff
@@ -576,17 +576,17 @@ def fix(self, propertyName):
 
     def deprecatedImpl(self, propertyName):
         if self.currentSection in self.validSections:
-            self.srcFile.write("    new Property(\"/^%s\.%s/\", true, null),\n" % (self.currentSection, \
+            self.srcFile.write("    new Property(/^%s\\.%s/, true, null),\n" % (self.currentSection, \
```
pepone: Updated to generate RegExp literals

js/src/Ice/PropertiesI.js
```diff
@@ -157,39 +157,28 @@ class Properties
             throw new InitializationException("Attempt to set property with empty key");
         }
 
-        let dotPos = key.indexOf(".");
+        const dotPos = key.indexOf(".");
         if(dotPos !== -1)
         {
             const prefix = key.substr(0, dotPos);
-            for(let i = 0; i < PropertyNames.validProps.length; ++i)
+            const i = PropertyNames.clPropNames.indexOf(prefix);
```
pepone, line 164: Replaced the outer loop with a search of prefix in clPropNames.

js/src/Ice/PropertiesI.js
```diff
@@ -199,14 +188,11 @@ class Properties
                     }
                     else
                     {
-                        pComp = new RegExp(pattern.toUpperCase());
-                        found = pComp.test(key.toUpperCase());
+                        found = new RegExp(pattern.source, "i").test(key);
```
pepone: We only have to create a regexp in the else branch, for the case insensitive comparison

pepone:
> As well, properties like "Ice.Trace.Networkkk" aren't detected as unknown.

Fixed the regexp it was missing the `$` to match the end of the string.

config/makeprops.py
```diff
@@ -512,15 +512,15 @@ def fix(self, propertyName):
         return propertyName.replace(".", "\\.").replace("[any]", "[^\\s]+")
 
     def deprecatedImpl(self, propertyName):
-        self.srcFile.write("             new Property(@\"^%s\.%s$\", true, null),\n" % (self.currentSection, \
+        self.srcFile.write("             new Property(@\"^%s\\.%s$\", true, null),\n" % (self.currentSection, \
```
externl, line 515: Is there an issue here with C#?

pepone: Python3 complains `\.` is an invalid scape sequence

PR #3072, opened 2024-11-05
Fix C# connection to correctly mark message as sent

pepone:
This PR fixes C# Connection to correctly mark messages as sent, once that startWrite handles the complete buffer.

This bug was introduced in  #2270, probably and oversight because of the subtle difference between C# and C++ startWrite calls.

Here to warrantee at-most once semantics we have to mark the request as sent, once that startWrite is has consumed the complete buffer.

For C++ this is the return value of `startWrite`, for C# the return value is whether or not the write call completed synchronously and there is a separate out parameter that indicates whether or not the call consumed the complete buffer.

This must be back ported to 3.7, but the issue doesn't affect 3.7.10.

Fixes #2722

PR #4861, opened 2026-01-09
Fix C# connection to correctly mark messages as sent

### LINKED ISSUE #3079: Backport #3072 to 3.7
We need to back port the fix from #3072 to 3.7 branch

pepone:
Fix #3079, see also #3072

### DIFF
--- a/csharp/src/Ice/ConnectionI.cs
+++ b/csharp/src/Ice/ConnectionI.cs
@@ -1071,16 +1071,16 @@ namespace Ice
                                 this,
                                 out messageWritten);
 
-                            // If the startWrite call consumed the complete buffer, we assume the message
-                            // is sent now for at-most-once semantics.
+                            // If the startWrite call wrote the message, we assume the message is sent now for at-most-once
+                            // semantics in the event the connection is closed while the message is still in _sendStreams.
                             if (messageWritten && _sendStreams.Count > 0)
                             {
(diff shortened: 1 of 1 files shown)


csharp/src/Ice/ConnectionI.cs
```diff
@@ -1064,15 +1064,23 @@ public override bool startAsync(int operation, IceInternal.AsyncCallback complet
...
-                            bool completed;
-                            if (_transceiver.startWrite(_writeStream.getBuffer(), completedCallback, this, out completed))
+                            bool messageWritten = false;
+                            bool completedSynchronously = _transceiver.startWrite(
+                                _writeStream.getBuffer(),
+                                completedCallback,
+                                this,
+                                out messageWritten);
+
+                            // If the startWrite call consumed the complete buffer, we assume the message
```
bernardnormier: We should use the comments currently on main, not the comments from the old PR:

```cs
                        // If the startWrite call wrote the message, we assume the message is sent now for at-most-once
                        // semantics in the event the connection is closed while the message is still in _sendStreams.
                        if (messageWritten && _sendStreams.Count > 0)
                        {
                            // See finish() code.
                            _sendStreams.First.Value.isSent = true;
                        }


```

pepone: Fixed the comments to match what we have in main.

PR #4776, opened 2025-12-17
Update Swift Doc Comments to Match C++ et. al.

InsertCreativityHere:
This PR implements #4747.

PR #4862, opened 2026-01-09
Update Python Doc-Comments

InsertCreativityHere:
This updates the doc-comments we have in Python to match the other languages, just like:
https://github.com/zeroc-ice/ice/pull/4776, https://github.com/zeroc-ice/ice/pull/4752, etc. did.

This PR does not include the doc-comment updates I made to the actual IcePy C++ code. I'll open a separate PR with those changes after this one, to keep this PR at least possible to review.

### DIFF
--- a/python/test/Ice/ami/AllTests.py
+++ b/python/test/Ice/ami/AllTests.py
@@ -40,9 +40,6 @@ class CallbackBase:
             self._called = True
             self._cond.notify()
 
-    def exception(self, ex: BaseException):
-        test(False)
-
 
 class FutureDoneCallback(CallbackBase):
     def isA(self, f: Ice.Future[bool]):
@@ -126,8 +123,8 @@ class FutureFlushExCallback(CallbackBase):
         CallbackBase.__init__(self)
         self._cookie = cookie
 
-    @override
-    def exception(self, ex: BaseException):
+    def ex(self, f: Ice.Future):
+        test(f.exception() is not None)
         self.called()
 
     def sent(self, sentSynchronously: bool):
@@ -521,7 +518,7 @@ def allTestsFuture(helper: TestHelper, communicator: Ice.Communicator, collocate
         cb = FutureFlushExCallback()
         f = con.flushBatchRequestsAsync(Ice.CompressBatch.BasedOnProxy)
         assert isinstance(f, Ice.InvocationFuture)
-        f.add_done_callback(cb.exception)
+        f.add_done_callback(cb.ex)
         f.add_sent_callback(cb.sent)
         cb.check()
         test(not f.is_sent())
--- a/python/python/Ice/Future.py
+++ b/python/python/Ice/Future.py
@@ -96,7 +96,7 @@ class Future(Awaitable[_T]):
         with self._condition:
             return self._state in [Future.StateCancelled, Future.StateDone]
 
-    def add_done_callback(self, fn: Callable[[Self], None]) -> None:
+    def add_done_callback(self, fn: Callable[["Future"], Any]) -> None:
         """
         Attaches a callback function which will be called when this future completes or is cancelled.
         If this future is already complete, ``fn`` is called immediately from the calling thread.


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -479,26 +479,27 @@ private CommunicatorFlushBatch _iceI_flushBatchRequestsAsync(CompressBatch compr
...
         return _instance.createAdmin(adminAdapter, adminId);
     }
 
     /**
-     * Gets a proxy to the main facet of the Admin object. getAdmin also creates the Admin object and creates and
-     * activates the {@code Ice.Admin} object adapter to host this Admin object if {@code Ice.Admin.Endpoints} is set.
-     * The identity of the Admin object created by getAdmin is {@code {value of Ice.Admin.InstanceName}/admin}, or
-     * {@code {UUID}/admin} when {@code Ice.Admin.InstanceName} is not set. If {@code Ice.Admin.DelayCreation} is
-     * {@code 0} or not set, getAdmin is called by the communicator initialization, after initialization of all plugins.
+     * Gets a proxy to the main facet of the Admin object. {@code getAdmin} also creates the Admin object and creates
```
InsertCreativityHere, line 497: I added `{@code ...` formatting around `getAdmin`.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -54,7 +54,7 @@ public static Properties createProperties(String[] args, List<String> remainingA
      * constructor directly.
      *
      * @param args The command-line arguments.
-     * @param defaults Default values for the property set. Settings in configuration files and
+     * @param defaults Default values for the new property set. Settings in configuration files and
```
InsertCreativityHere, line 57: This is the wording used in the other languages. Updated to include `new`.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -128,8 +128,9 @@ public static Communicator initialize(String[] args) {
     /**
      * Converts a stringified identity into an Identity.
      *
```
InsertCreativityHere, line 130: Accidentally skipped over this function while updating Java apparently.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectAdapter.java
```diff
@@ -622,6 +621,9 @@ public synchronized Map<String, Object> findAllFacets(Identity identity) {
     /**
      * Looks up a servant with an identity and a facet. It's equivalent to calling {@link #findFacet}.
      *
+     * <p>This method only tries to find the servant in the ASM and among the default servants.
+     * It does not attempt to locate a servant using servant locators.
```
InsertCreativityHere, line 625: We have this note in the other big languages. Added it to Java for consistency.

cpp/include/Ice/NativePropertiesAdmin.h
```diff
@@ -38,6 +38,8 @@ namespace Ice
         std::function<void()> addUpdateCallback(std::function<void(const PropertyDict&)> cb);
 
     private:
```
InsertCreativityHere, line 40: I know it's private in C++, but it has a doc-comment in the other languages.
Doesn't hurt to have it here as well.

cpp/include/Ice/Logger.h
```diff
@@ -54,7 +54,7 @@ namespace Ice
 
         /// Returns a clone of the logger with a new prefix.
         /// @param prefix The new prefix for the logger.
-        /// @return A logger instance.
+        /// @return A new logger instance with the specified prefix.
```
InsertCreativityHere, line 57: I preferred the comment we had in Python for this.
Updated all the other languages to use it.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -479,26 +479,27 @@ private CommunicatorFlushBatch _iceI_flushBatchRequestsAsync(CompressBatch compr
...
+     * and activates the {@code Ice.Admin} object adapter to host this Admin object if {@code Ice.Admin.Endpoints} is
+     * set. The identity of the Admin object created by {@code getAdmin} is
+     * {@code {value of Ice.Admin.InstanceName}/admin}, or {@code {UUID}/admin} when {@code Ice.Admin.InstanceName}
```
pepone, line 500: Does Java doc automatically deal with the nested `{}`?

python/python/Ice/Communicator.py
```diff
@@ -174,9 +174,13 @@ def waitForShutdown(self) -> None:
...
         """
-        Return an awaitable that completes when the communicator's shutdown completes.
+        Returns an :class:`Awaitable` that completes when the communicator's shutdown completes.
+        This task always completes successfully.
 
-        The returned Future always completes successfully.
+        Notes
+        -----
+        The shutdown of a communicator completes when all its incoming connections are closed.
+        Awaiting this task is equivalent to calling :func:`waitForShutdown`.
```
pepone: saying that is equivalent to waitForShutdown is a bit confusing, one is synchronous and blocks the other is asynchronous and doesn't block.

python/python/Ice/CompressBatch.py
```diff
@@ -6,7 +6,7 @@
 
 class CompressBatch(Enum):
     """
-    The batch compression option when flushing queued batch requests.
+    Represents batch compression options when flushing queued batch requests.
```
pepone: Wouldn't `for flushing` be more clear here?

cpp/src/Ice/NativePropertiesAdmin.cpp
```diff
@@ -178,8 +178,8 @@ Ice::NativePropertiesAdmin::addUpdateCallback(std::function<void(const PropertyD
 }
 
 void
-Ice::NativePropertiesAdmin::removeUpdateCallback(std::list<std::function<void(const PropertyDict&)>>::iterator p)
+Ice::NativePropertiesAdmin::removeUpdateCallback(std::list<std::function<void(const PropertyDict&)>>::iterator cb)
```
externl: p is more correct here imo.

python/python/Ice/LocalExceptions.py
```diff
@@ -40,19 +42,19 @@ def __init__(self, replyStatus: int | None, msg: str = ""):
     @property
     def replyStatus(self):
```
externl: Missing return type hint here I think.

python/python/Ice/LocalExceptions.py
```diff
@@ -66,44 +68,44 @@ def __init__(
...
         Returns
         -------
         Identity
-            The identity of the Ice Object to which the request was sent.
+            The identity.
         """
         return self.__id
 
     @property
     def facet(self):
```
externl: Ditto.

python/python/Ice/LocalExceptions.py
```diff
@@ -66,44 +68,44 @@ def __init__(
...
         Returns
         -------
         str
-            The facet to which the request was sent.
+            The facet.
         """
         return self.__facet
 
     @property
     def operation(self):
```
externl: Ditto.

python/python/Ice/LocalExceptions.py
externl: ditto.

externl, line 1: This whole file should be double checked for type missing type hints.

python/python/Ice/Object.py
```diff
@@ -72,74 +72,75 @@ class Object:
...
         id : str
-            The Slice type ID.
+            The type ID of the Slice interface to test against.
         current : Current
-            The current context.
+            The Current object of the incoming request.
 
         Returns
         -------
         bool
```
externl: Return type is incorrect.

python/python/Ice/Object.py
```diff
@@ -72,74 +72,75 @@ class Object:
...
     def ice_ping(self, current: Current) -> None | Awaitable[None]:
         """
-        A reachability test for the target object.
+        Tests whether this object can be reached.
 
         Parameters
         ----------
         current : Current
-            The current context.
+            The Current object of the incoming request.
```
externl, line 99: Missing returns. This is a weird one though.

python/python/Ice/Object.py
```diff
@@ -72,74 +72,75 @@ class Object:
...
 
         Returns
         -------
         Sequence[str]
```
externl: Incorrect return type.

python/python/Ice/Object.py
```diff
@@ -72,74 +72,75 @@ class Object:
...
 
         Parameters
         ----------
         current : Current
-            The current context.
+            The Current object of the incoming request.
 
         Returns
         -------
         str
```
externl: Ditto.

python/python/Ice/Util.py
```diff
@@ -36,18 +36,20 @@ def initialize(
...
-    :class:`Communicator` constructor directly.
+    Creates a new communicator.
+
+    This function is provided for backwards compatibility.
+    New code should call the :class:`Communicator` constructor directly.
 
     Parameters
     ----------
-    args : list of str, optional
+    args : list[str], optional
```
externl: These `, optional` should be `| None`

```suggestion
    args : list[str] | None
```

python/python/Ice/Util.py
```diff
@@ -36,18 +36,20 @@ def initialize(
...
         The command-line arguments.
     eventLoop : asyncio.AbstractEventLoop, optional
-        An asyncio event loop used to run coroutines and wrap futures. This argument and the `initData` argument are
-        mutually exclusive.
+        An asyncio event loop used to run coroutines and wrap futures.
     initData : InitializationData, optional
```
externl: Ditto.

python/python/Ice/Util.py
```diff
@@ -59,36 +61,36 @@ def initialize(
...
     """
-    Convert an object identity to a string.
+    Converts an Identity into a string using the specified mode.
 
     Parameters
     ----------
     identity : Ice.Identity
-        The object identity to convert.
+        The identity.
     toStringMode : Ice.ToStringMode, optional
```
externl: Ditto.

python/python/Ice/Proxy.py
```diff
@@ -34,12 +64,42 @@ def proxyIdentityCompare(lhs: IcePy.ObjectPrx | None, rhs: IcePy.ObjectPrx | Non
 
 
 def proxyIdentityAndFacetEqual(lhs: IcePy.ObjectPrx | None, rhs: IcePy.ObjectPrx | None) -> bool:
-    """Determines whether the identities and facets of two proxies are equal."""
+    """
+    Checks if the object identifiers and facets of two proxies are equal.
```
pepone: using `identities` instead of `identifiers` would be more consistent

python/python/IcePy-stubs/__init__.pyi
```diff
@@ -307,36 +301,37 @@ class Connection:
...
+    ConnectionInfo | None: The information of the underlying transport or ``None`` if there's no underlying transport.
+    """
 
     adapterName: str
-    """str: adapter associated with the connection"""
+    """str: The name of the adapter associated with the connection."""
 
     incoming: bool
-    """bool: whether connection is incoming"""
+    """bool: ``True`` if this an incoming connection, ``False`` otherwise."""
```
pepone:
```suggestion
    """bool: ``True`` if this is an incoming connection, ``False`` otherwise."""
```

cpp/include/Ice/ObjectAdapter.h
```diff
@@ -23,7 +23,7 @@ namespace Ice
     ///   #activate); and
     /// - maintain a dispatch pipeline and servants that handle the requests (see #add, #addDefaultServant, and #use).
     ///
-    /// An object adapter can dispatch "bidirectional requests"--requests it receives over an outgoing connection
+    /// An object adapter can dispatch "bidirectional requests" -- requests it receives over an outgoing connection
```
bernardnormier: No space before or after Em dash is (was) correct:
https://www.merriam-webster.com/grammar/em-dash-en-dash-how-to-use

cpp/src/Ice/NativePropertiesAdmin.cpp
InsertCreativityHere: Fine : v/ back to `p`.

python/python/Ice/LocalExceptions.py
InsertCreativityHere: Yes, fixed!

python/python/Ice/LocalExceptions.py
InsertCreativityHere: Fixed.

python/python/Ice/LocalExceptions.py
InsertCreativityHere: Fixed.

python/python/Ice/LocalExceptions.py
InsertCreativityHere: Fixed.

InsertCreativityHere: I fixed all the missing return type hints in this file.
I must of started on this file before I decided to start fixing them.

python/python/Ice/Object.py
InsertCreativityHere: I added the `| Awaitable[bool]`.

python/python/Ice/Object.py
InsertCreativityHere: Yeah, that's why I skipped it entirely (it has return type `None | Awaitable[None]`).
Not really clear what to write, since conceptually this doesn't return any data.

And it only sometimes returns an `Awaitable`, sometimes not. So... also would be weird to only focus on the awaitable and ignore the real possibility this just returns `None`.

python/python/Ice/Object.py
InsertCreativityHere: Fixed to include the `Awaitable` part.

python/python/Ice/Object.py
InsertCreativityHere: Fixed to include the `Awaitable` part.

python/python/Ice/Util.py
InsertCreativityHere: After talking it over, we settled on `list[str] | None, optional`.
All the `, optional`s in Python were updated accordingly.

python/python/Ice/Util.py
InsertCreativityHere: Fixed.

python/python/Ice/Util.py
InsertCreativityHere: Fixed.

python/python/Ice/Communicator.py
bernardnormier: Should be ... equivalent to awaiting... like in C#.

bernardnormier (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
InsertCreativityHere: Yep, I was wondering this too. It seems smart enough to handle nested pair braces:
https://code.zeroc.com/ice/main/api/java/com.zeroc.ice/com/zeroc/Ice/Communicator.html#getAdmin()

cpp/include/Ice/ObjectAdapter.h
InsertCreativityHere: I updated it back to the old spacing (closed em-dashes). But neither is more correct than the other to my knowledge.
This is at the bottom of the page you linked to:

> Spacing around an em dash varies. Most newspapers insert a space before and after the dash, and many popular magazines do the same, but most books and journals omit spacing, closing whatever comes before and after the em dash right up next to it. This website prefers the latter, its style requiring the closely held em dash in running text.

I learned AP-style writing in school, which uses open em-dashes, is all.

python/python/Ice/Communicator.py
InsertCreativityHere: Fixed.

python/python/Ice/CompressBatch.py
InsertCreativityHere: Fine with me. I updated this comment in all the languages (except MATLAB and JS, which still need comment work).

python/python/Ice/Proxy.py
InsertCreativityHere: Yes, a typo on my part. Fixed!

python/python/IcePy-stubs/__init__.pyi
InsertCreativityHere: Fixed. Was also wrong in Java.

python/test/Ice/ami/AllTests.py
```diff
@@ -521,7 +518,7 @@ def allTestsFuture(helper: TestHelper, communicator: Ice.Communicator, collocate
         cb = FutureFlushExCallback()
         f = con.flushBatchRequestsAsync(Ice.CompressBatch.BasedOnProxy)
         assert isinstance(f, Ice.InvocationFuture)
-        f.add_done_callback(cb.exception)
```
InsertCreativityHere, line 524: Updating the type-hints of `add_done_callback` seems to of caught a bug in this test.
`add_done_callback` takes a `Callable` which accepts an `Ice.Future` as an argument.

Here we were passing a function which took a `BaseException`. The function didn't use this argument... so no harm I guess, but it seemed incorrect.
Also weird that the function is named `exception`, but didn't actually check that the `Future` completed with an exception? Idk. I updated the test to what I think it was trying to do.

cpp/include/Ice/ObjectAdapter.h
externl: I too prefer open em-dashes.

PR #4868, opened 2026-01-10
Add 3.7 Swift Package to build release workflow

pepone:


.github/workflows/build-swift-packages.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: macos-15
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Setup C++
+        uses: ./.github/actions/setup-cpp
+
+      - name: Build SPM sources
+        run: ./build-spm-sources.py
```
pepone, line 27: With 3.7 we rely on the SPM package to include the generated sources.

.github/workflows/publish-brew-packages.yml
```diff
@@ -59,15 +59,15 @@ jobs:
 
       - name: Clone Repository
         run: |
-          git clone "https://x-access-token:${{ secrets.ICE_NIGHTLY_PUBLISH_TOKEN }}@github.com/zeroc-ice/homebrew-nightly.git" -b main
+          git clone "https://x-access-token:${{ secrets.ICE_NIGHTLY_PUBLISH_TOKEN }}@github.com/zeroc-ice/homebrew-nightly.git" -b ${{ inputs.channel }}
```
pepone: and unrelated fix for brew already in main/3.8

I should probably cherry-pick this from main in its own commit.

.github/workflows/publish-npm-packages.yml
```diff
@@ -56,7 +56,9 @@ jobs:
           echo "always-auth=true" >> ~/.npmrc
 
           # Publish both slice2js and ice packages
-          for pkg in staging/js-npm-packages/*.tgz; do
```
pepone: I will fix this in main first, and cherry pick. It avoids relying on specific package names for NPM

.github/workflows/publish-swift-packages.yml
```diff
@@ -0,0 +1,63 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_ID: ${{ inputs.run_id }}
+        run: |
+          gh run download "$RUN_ID" --repo zeroc-ice/ice --pattern "spm-sources" --dir staging
+
+      - name: Extract SPM sources
+        id: extract-spm-sources
+        run: |
+          # Find the spm-sources tar.gz file and extract the version from the filename
+          tarball=(staging/spm-sources/spm-sources-*.tar.gz)
```
pepone: In may we extract the version from the XCFramework packages, here we extract it from the pre-built SPM sources.

I think in moving this logic to update-ice-swift-nightly.sh like we do for XCFrameworks in main.

packaging/swift/build-spm-sources.py
```diff
@@ -0,0 +1,293 @@
+#!/usr/bin/env python3
```
pepone, line 1: This is the logic we previously have in dist-utils to build SPM sources, but as a standalone script.

packaging/swift/Package.swift
```diff
@@ -0,0 +1,43 @@
+// swift-tools-version:5.0
```
pepone, line 1: This come from https://github.com/zeroc-ice/ice-spm

packaging/swift/update-ice-swift-nightly.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
pepone, line 1: This come from main, but removed the XCFramework logic, as we don't have XCFrameworks in 3.7

.github/workflows/publish-brew-packages.yml
pepone: In the end is simpler to have all formulas in main branch for Homebrew. the formulas already include the channel in the name `ice@39`, `ice@38`, ...

swift/src/IceImpl/LocalObject.mm
```diff
@@ -70,7 +70,8 @@ - (void)dealloc
 
         // Find the object in the cache. If it's not there, it's likely that another thread already replaced
         // the object with a new one and then that new one was quickly deallocated.
-        if (auto p = cachedObjects->find(_cppObject.get()); p != cachedObjects->end())
+        auto p = cachedObjects->find(_cppObject.get());
+        if (p != cachedObjects->end())
```
pepone, line 74: This `if (auto` caused a build failure with C++14 used by the Swift package.

packaging/swift/Package.swift
```diff
@@ -0,0 +1,43 @@
...
+    .iOS(.v12),
+  ],
+  products: [
+    .library(name: "Ice", targets: ["Ice"]),
+    .library(name: "Glacier2", targets: ["Glacier2"]),
+    .library(name: "IceStorm", targets: ["IceStorm"]),
+    .library(name: "IceGrid", targets: ["IceGrid"])
+  ],
+  dependencies: [
+    .package(url: "https://github.com/mxcl/PromiseKit.git", from: "6.8.10")
```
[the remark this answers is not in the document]
bernardnormier: Is 6.8.10 an actual release? I can't find it.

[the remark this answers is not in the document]
pepone: It is not.  That is what we have in the released package https://github.com/zeroc-ice/ice-spm/blob/973dc700d98a2b56875d078df67e90eeb2d10847/Package.swift#L25

We are using a different version in 3.7 branch now https://github.com/zeroc-ice/ice/blob/bd9fa670bc2d6c15700d8d81ca3f75c2e66ce571/swift/Rakefile#L853

I will upgrade it to latest 6.x

PR #4874, opened 2026-01-12
Simplify 'lookupType' Methods to Return `TypePtr` instead of `TypeList`

InsertCreativityHere:
In the parser, the two `lookupType` methods take a string, and return a list of possible matches (`TypeList`).
But, we only ever use the first entry from the list, and conceptually, it's weird to return a list anyways.
It should be up to the method to determine the correct type, not let the caller sift through a list to figure it out.

These functions were updated to return the first match, instead of an entire list of possibilities.

### DIFF
--- a/cpp/test/Slice/errorDetection/IdentAsKeyword.err
+++ b/cpp/test/Slice/errorDetection/IdentAsKeyword.err
@@ -49,7 +49,9 @@ IdentAsKeyword.ice:46: dictionaries can only be defined within a module
 IdentAsKeyword.ice:48: syntax error, unexpected out keyword
 IdentAsKeyword.ice:49: dictionaries can only be defined within a module
 IdentAsKeyword.ice:51: syntax error, unexpected void keyword
+IdentAsKeyword.ice:52: exception name 'VOID' is capitalized inconsistently with its previous name: '::Void'
 IdentAsKeyword.ice:52: 'VOID' is an exception, which cannot be used as a type
+IdentAsKeyword.ice:52: exception name 'VOid' is capitalized inconsistently with its previous name: '::Void'
 IdentAsKeyword.ice:52: 'VOid' is an exception, which cannot be used as a type
 IdentAsKeyword.ice:52: dictionaries can only be defined within a module
 IdentAsKeyword.ice:54: keyword 'idempotent' cannot be used as a name


cpp/src/Slice/Grammar.cpp
InsertCreativityHere, line 1: Bison generated, don't review unless very bored.

cpp/src/Slice/Grammar.y
```diff
@@ -712,28 +712,26 @@ class_extends
 {
     auto scoped = dynamic_pointer_cast<StringTok>($2);
     ContainerPtr cont = currentUnit->currentContainer();
-    TypeList types = cont->lookupType(scoped->v);
+    TypePtr resolvedType = cont->lookupType(scoped->v);
     $$ = nullptr;
-    if (!types.empty())
+    if (resolvedType)
     {
-        auto cl = dynamic_pointer_cast<ClassDecl>(types.front());
```
InsertCreativityHere, line 719: Also swapped the branches here. Better to have `if (cl)` instead of `if (!cl)` IMO.

cpp/src/Slice/Parser.cpp
```diff
@@ -1470,70 +1470,66 @@ Slice::Container::lookupTypeNoBuiltin(const string& identifier, bool emitErrors,
             continue; // Ignore interface and class definitions.
         }
 
-        if (emitErrors && matches.front()->scoped() != (thisScope() + sc))
```
InsertCreativityHere: We only output the errors if `emitErrors` is `true`.
So, checking it here makes no difference.

cpp/test/Slice/errorDetection/IdentAsKeyword.err
```diff
@@ -49,7 +49,9 @@ IdentAsKeyword.ice:46: dictionaries can only be defined within a module
 IdentAsKeyword.ice:48: syntax error, unexpected out keyword
 IdentAsKeyword.ice:49: dictionaries can only be defined within a module
 IdentAsKeyword.ice:51: syntax error, unexpected void keyword
```
InsertCreativityHere, line 51: These errors are correct, but were previously hidden by our buggy logic.

cpp/src/Slice/Parser.cpp
```diff
@@ -1470,70 +1470,66 @@ Slice::Container::lookupTypeNoBuiltin(const string& identifier, bool emitErrors,
...
+        if (p->scoped() != (thisScope() + sc))
```
InsertCreativityHere, line 1473: The old logic was actually broken. This code takes place in a loop: 
`for (const auto& p : matches)`
But, instead of checking `p`, we would keep checking `matches.front()` every single loop. Not good.

PR #4879, opened 2026-01-13
Make authenticode signatures for Nightly optional

pepone:
This PR updates the logic for signing Windows binaries. For stable channel we always sign binaries, for nightly channel the default is to not sign, authenticode_sign input parameter allow to override the default.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -15,7 +15,7 @@ on:
         type: boolean
         default: false
       authenticode_sign:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-release.yml
```diff
@@ -14,6 +14,11 @@ on:
         required: false
         type: boolean
         default: false
+      authenticode_sign:
+        description: "Whether to Authenticode sign Windows binaries (always true for stable channel)"
```
bernardnormier: Should be "always true when quality is stable". There is no stable channel.

PR #4880, opened 2026-01-13
Stop treating warnings as errors by default

### LINKED ISSUE #4856: Only build with warnings as errors in CI
Our current builds (C++, C#, etc.) are configured to treat warnings as errors when possible. This inevitably breaks when building with a new compiler, os version, etc., and is not very helpful for users who just want a build that works. 

The Homebrew project often has to make various fixes to get Ice to build on various platforms/compiler combos due to this setting.

I think we should take the CMake philosophy, where it is bad practice for a dependency to treat warnings as errors, as it breaks the build for no good reason. 

 

We should, however, have this enabled for CI builds.

externl:
The `WARNINGS_AS_ERRORS` env variable can be set to 'yes' to enable.  This is the only truthy value. 

Fixes #4856


cpp/msbuild/ice.cpp.props
```diff
@@ -42,7 +42,7 @@
       <PreprocessorDefinitions>_CONSOLE;_WIN32_WINNT=0x0A00;WIN32_LEAN_AND_MEAN</PreprocessorDefinitions>
       <DisableSpecificWarnings>4121;4250;4251;4275;4324</DisableSpecificWarnings>
       <AdditionalOptions>/utf-8 /bigobj %(AdditionalOptions)</AdditionalOptions>
-      <TreatWarningAsError>true</TreatWarningAsError>
+      <TreatWarningsAsErrors Condition="'$(ICE_WARNINGS_AS_ERRORS)' != ''">$(ICE_WARNINGS_AS_ERRORS)</TreatWarningsAsErrors>
```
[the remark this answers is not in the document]
pepone: ~~The new value seems correct. The previous code was using a bogus property name.~~

Actually copilot is right here in C++ the metadata is sigular.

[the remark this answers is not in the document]
bernardnormier: You should add a comment about this confusing name in `ice.cpp.props`.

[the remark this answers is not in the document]
pepone: This is a standard VC++ thing

https://learn.microsoft.com/en-us/cpp/build/reference/vcxproj-file-structure?view=msvc-170#itemgroup-elements

cpp/msbuild/ice.cpp.props
```diff
@@ -55,7 +55,8 @@
       <LanguageStandard>stdcpp20</LanguageStandard>
       <!-- Include directories -->
       <AdditionalIncludeDirectories>$(IceHome)\cpp\include;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
-      <AdditionalIncludeDirectories>$(IceHome)\cpp\include\generated\$(Platform)\$(Configuration);%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
+      <AdditionalIncludeDirectories>
```
pepone: I think this white-spaces will show up in Visual Studio/MSBuild. Better keep the previous version.

config/Make.rules.Linux
```diff
@@ -143,7 +143,9 @@ ifeq ($(filter amzn,$(linux_id)),)
 fully_static_ldflags = -static
 endif
 
-cppflags        = -fvisibility=hidden -Wall -Wextra -Wredundant-decls -Wshadow -Wdeprecated -Werror -pthread \
+cppflags        = -fvisibility=hidden -Wall -Wextra -Wredundant-decls -Wshadow -Wdeprecated \
+                  $(if $(ICE_WARNINGS_AS_ERRORS),-Werror,) \
```
bernardnormier: We should be consistent with how we use/set these variables.

With this PR:
OPTIMIZE=yes => we optimize
OPTIMIZE set to any other value => we don't optimize
OPTIMIZE not set => we set it to yes

ICE_WARNINGS_AS_ERRORS set to any value => we treat warnings as errors.

cpp/msbuild/ice.cpp.props
```diff
@@ -42,7 +42,7 @@
       <PreprocessorDefinitions>_CONSOLE;_WIN32_WINNT=0x0A00;WIN32_LEAN_AND_MEAN</PreprocessorDefinitions>
       <DisableSpecificWarnings>4121;4250;4251;4275;4324</DisableSpecificWarnings>
       <AdditionalOptions>/utf-8 /bigobj %(AdditionalOptions)</AdditionalOptions>
-      <TreatWarningAsError>true</TreatWarningAsError>
+      <TreatWarningAsError Condition="'$(ICE_WARNINGS_AS_ERRORS)' != ''">$(ICE_WARNINGS_AS_ERRORS)</TreatWarningAsError>
```
bernardnormier: It's rather unusual. We rarely use env variables in the project files. And in the rare spot we do, we use a different logic:

```
  <!-- See https://github.com/ccache/ccache/wiki/MS-Visual-Studio -->
  <PropertyGroup Condition="'$(USING_BUILD_CACHE)' == 'true'">
    <UseMultiToolTask>true</UseMultiToolTask>
  </PropertyGroup>
  <ItemDefinitionGroup Condition="'$(USING_BUILD_CACHE)' == 'true'">
    <ClCompile>
      <ForcedIncludeFiles />
      <DebugInformationFormat>OldStyle</DebugInformationFormat>
      <ObjectFileName>$(IntDir)%(FileName).obj</ObjectFileName>
    </ClCompile>
  </ItemDefinitionGroup>
</Project>
```

externl: It's okay to have different logic. 

If we do it a different way we would likely need both `==` and `!=` conditionals which is messy.

config/Make.rules.Linux
externl: `ICE_WARNINGS_AS_ERRORS` is consistent with the usage of the `V` variable. The setup for `OPTIMIZE` is arguably bad as it's much more error prone.

cpp/msbuild/ice.cpp.props
pepone: MSBuild expose the environment variables as MSBuild properties, the distinction is a bit blurry.

Maybes is a bit clear to put this in is own ItemDefinitionGroup:

```
<ItemDefinitionGroup Condition="'$(ICE_WARNINGS_AS_ERRORS)' != ''">
  <ClCompile>
    <TreatWarningAsError>$(ICE_WARNINGS_AS_ERRORS)</TreatWarningAsError>
  </ClCompile
</ItemDefinitionGroup>
```

But they are equivalent...

externl: preferably I would change the `USING_BUILD_CAHCE` check to also use `!= ''`

config/Make.rules.Linux
bernardnormier: OPTIMIZE is used to augment cppflags on the next line. It's more relevant in the context of C++ flags.

externl: Fine, I'll change it, optimize is actually even worse. In C# and Swift we actually check for `no`.

```
 $(if $(filter $(OPTIMIZE),no),/p:Configuration=Debug)
 ```

PR #4883, opened 2026-01-13
Update `IcePy` Doc-Comment Strings to Match Python & Other Languages

InsertCreativityHere:
This PR does 3 things:
- Fixes some types in our `*.py` files to use `Ice.xxx` instead of `IcePy.xxx`.
- Updates the `ObjectPrx.py` file, which I forgot to save in my last PR...
- Updates the doc-comments in our C++ `IcePy` code to match our other comments.

It also re-exports `IcePy.Endpoint` as `Ice.Endpoint`. It really seems like we meant to, but just didn't?
Let me know if there was a reason why we kept `Endpoint` as a purely `IcePy` type...

### DIFF
--- a/python/python/Ice/Proxy.py
+++ b/python/python/Ice/Proxy.py
@@ -8,9 +8,11 @@ IcePy._t_ObjectPrx = IcePy.defineProxy("::Ice::Object", IcePy.ObjectPrx)
 
 __name__ = "Ice"
 
+
 #
 # Proxy comparison functions.
 #
+
 
 def proxyIdentityEqual(lhs: ObjectPrx | None, rhs: ObjectPrx | None) -> bool:
     """


python/modules/IcePy/ImplicitContext.cpp
```diff
@@ -248,24 +248,27 @@ static PyMethodDef ImplicitContextMethods[] = {
     {"getContext",
      reinterpret_cast<PyCFunction>(implicitContextGetContext),
```
InsertCreativityHere, line 249: These return types are just wrong as far as I can tell.
Seems they copy the C++ code that underlies them, instead of the Python API they're implementing.

python/modules/IcePy/Init.cpp
```diff
@@ -28,97 +28,98 @@ namespace
...
 
 Returns
 -------
 int
-    The Ice version as an integer for version comparison.)";
+    The Ice version.)";
 
     constexpr const char* IcePy_createProperties_doc =
-        R"(createProperties(args: list[str] | None = None) -> Ice.Properties
+        R"(createProperties(args: list[str] | None = None, defaults: Ice.Properties | None = None) -> Ice.Properties
```
InsertCreativityHere, line 51: This doc-comment was just completely missing the `defaults` argument.
I double checked the implementation, and the IcePy parsing code: `defaults` should of been here.

python/modules/IcePy/Init.cpp
```diff
@@ -28,97 +28,98 @@ namespace
...
 
 Returns
 -------
 Ice.Identity
-    The identity.
+    An Identity created from the provided string.
 
 Raises
 ------
-IdentityParseException
```
InsertCreativityHere, line 80: `IdentityParseException` was removed in our exception-hierarchy refactoring awhile ago.
It's now just `ParseException`.

python/modules/IcePy/Init.cpp
```diff
@@ -28,97 +28,98 @@ namespace
...
-    If the string is not a valid identity.)";
+ParseException
+    If the string cannot be converted to an object identity.)";
 
     constexpr const char* IcePy_identityToString_doc =
-        R"(identityToString(ident: Ice.Identity, mode: Ice.ToStringMode = Ice.ToStringMode.Unicode) -> str
+        R"(identityToString(identity: Ice.Identity, toStringMode: Ice.ToStringMode | None = None) -> str
```
InsertCreativityHere, line 87: The default value of `Ice.ToStringMode.Unicode` was incorrect and a misunderstanding.
If you don't pass anything into this function, we default to `None`, and it just so happens that if we see `None`, we internally map it to `Unicode`.

But the API level, this should of been `Ice.ToStringMode | None = None`.

python/modules/IcePy/Properties.cpp
```diff
@@ -695,7 +695,7 @@ static PyMethodDef PropertyMethods[] = {
     {"getPropertyWithDefault",
      reinterpret_cast<PyCFunction>(propertiesGetPropertyWithDefault),
      METH_VARARGS,
```
InsertCreativityHere, line 697: In the other languages, we call this parameter `value`, not `default`.

python/python/Ice/IcePyTypes.py
```diff
@@ -6,6 +6,7 @@
     BatchRequest,
     Connection,
     ConnectionInfo,
+    Endpoint,
```
InsertCreativityHere, line 9: So, double-check me on this, but we weren't re-exporting `IcePy.Endpoint` as `Ice.Endpoint` as part of the public API.

And judging by the reference to this type, and the doc-comments we had on it, it really seems like this should of been re-exported. Let me know if this should remain private though.

python/python/IcePy-stubs/__init__.pyi
```diff
@@ -600,27 +601,27 @@ class Operation:
...
     def invoke(self, proxy: ObjectPrx, args: tuple) -> Any: ...
     def invokeAsync(self, proxy: ObjectPrx, args: tuple) -> Awaitable[Any]: ...
+    def deprecate(self, reason: str): ...
 
 class Properties:
-    def clone(self) -> Properties: ...
-    def getCommandLineOptions(self) -> list[str]: ...
+    def getProperty(self, key: str) -> str: ...
     def getIceProperty(self, key: str) -> str: ...
+    def getPropertyWithDefault(self, key: str, value: str) -> str: ...
```
InsertCreativityHere, line 611: All of the `...WithDefault` functions had their second parameter renamed from `default` to `value`.

python/python/IcePy-stubs/__init__.pyi
```diff
@@ -683,15 +684,16 @@ class WSEndpointInfo(EndpointInfo):
     resource: str
     """str: The URI configured with the endpoint."""
 
-def compileSlice(args: list[str]) -> int: ...
-def createProperties(args: list[str] | None, defaults: Ice.Properties | None): ...
 def stringVersion() -> str: ...
 def intVersion() -> int: ...
+def createProperties(args: list[str] | None = None, defaults: Ice.Properties | None = None) -> Ice.Properties: ...
```
InsertCreativityHere: I updated the signature of this function. It was missing the ` = None` defaults, and the return type.

python/python/IcePy-stubs/__init__.pyi
```diff
@@ -163,7 +192,7 @@ class Connection:
         compress: Ice.CompressBatch,
     ) -> Awaitable[None]:
         """
-        Flushes any pending batch requests for this connection.
+        Flushes any pending batch requests for this connection asynchronously.
```
InsertCreativityHere, line 195: Added `asynchronously` to distinguish between `flushBatchRequests` and `flushBatchRequestsAsync`.

python/python/IcePy-stubs/__init__.pyi
```diff
@@ -145,6 +145,35 @@ class Connection:
         """
```
InsertCreativityHere, line 145: The functions in this stub file were in a random psuedo-alphabetical order.
Whereas the functions in the `IcePy` C++, `Ice` Python code and other languages were in a consistent order.

I swapped the order of functions in this file to match up with the others. In the few places where I actually changed the text, I have comments in this PR pointing them out.

python/modules/IcePy/Operation.cpp
```diff
@@ -850,23 +850,38 @@ IcePy::Operation::convertParam(PyObject* p, Py_ssize_t pos)
 }
 
 static PyMethodDef OperationMethods[] = {
```
InsertCreativityHere, line 852: I updated these to have the function signature for it's doc-comment (like we do for other unimportant/hidden IcePy functions) instead of just saying `internal function`.

python/modules/IcePy/ConnectionInfo.cpp
```diff
@@ -178,17 +169,17 @@ static PyGetSetDef ConnectionInfoGetters[] = {
...
      reinterpret_cast<getter>(connectionInfoGetUnderlying),
      nullptr,
-     PyDoc_STR("Ice.ConnectionInfo | None: underlying connection information"),
+     PyDoc_STR("ConnectionInfo | None: The information of the underlying transport or ``None`` if there's no underlying transport."),
      nullptr},
     {"incoming",
      reinterpret_cast<getter>(connectionInfoGetIncoming),
      nullptr,
-     PyDoc_STR("bool: whether connection is incoming"),
+     PyDoc_STR("bool: ``True`` if this is an incoming connection, ``False`` otherwise."),
```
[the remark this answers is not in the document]
InsertCreativityHere: I don't think it does.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
```diff
@@ -230,7 +230,7 @@ CompletableFuture<Object.Ice_invokeResult> ice_invokeAsync(
     /**
      * Gets the facet for this proxy.
      *
-     * @return the facet for this proxy. If the proxy uses the default facet, the return value is the empty string
```
InsertCreativityHere, line 233: Full sentences in Java should end with periods.

python/python/Ice/__init__.py
```diff
@@ -43,6 +43,7 @@
     BatchRequest,
     Connection,
     ConnectionInfo,
+    Endpoint,
     EndpointInfo,
```
InsertCreativityHere, line 47: It seems like we meant to re-export `IcePy.Endpoint` as `Ice.Endpoint`, but just didn't?

python/python/Ice/ObjectPrx.py
```diff
@@ -228,110 +215,111 @@ def ice_getCommunicator(self) -> Communicator:
...
+        Tests whether the target object of this proxy can be reached.
 
         Parameters
         ----------
         context : dict[str, str] | None, optional
-            The context dictionary for the invocation.
-
-        Examples
-        --------
-        >>> obj.ice_ping(context={"key": "value"})
```
InsertCreativityHere, line 276: We only have 2 examples, both for `ice_ping`.
I removed them, because nowhere else do we have examples, and these are kind of weird examples anyways IMO.

python/python/Ice/ObjectPrx.py
```diff
@@ -513,68 +502,68 @@ def ice_locatorCacheTimeout(self, timeout: int) -> Self:
...
     def ice_getInvocationTimeout(self) -> int:
         """
-        Returns the invocation timeout of this proxy.
+        Gets the invocation timeout of this proxy.
 
         Returns
         -------
         int
-            The invocation timeout value (in seconds).
+            The invocation timeout value (in milliseconds).
```
InsertCreativityHere, line 533: This is a big doc change.

We used to say the returned value was in _seconds_, but looking through the code, I'm pretty sure it's in _milliseconds_. Please double-check that I'm not wrong here.

python/python/Ice/ObjectPrx.py
```diff
@@ -369,7 +358,7 @@ def ice_getIdentity(self) -> Identity:
 
     def ice_identity(self, newIdentity: Identity) -> Self:
         """
-        Create a new proxy that is identical to this proxy, except for the per-proxy context.
```
InsertCreativityHere, line 372: The old comment was a bogus copy-paste from `ice_context` looks like.

python/modules/IcePy/Init.cpp
```diff
@@ -28,97 +28,98 @@ namespace
...
-Returns the Ice version as a string.
+Returns the Ice version in the form ``A.B.C``, where ``A`` indicates the major version,
+``B`` indicates the minor version, and ``C`` indicates the patch level.
 
 Returns
 -------
 str
-    The Ice version in the format 'A.B.C' where A is the major version,
-    B is the minor version, and C is the patch version.)";
+    The Ice version.)";
```
externl:
```suggestion
    The Ice version.";
```

python/python/Ice/ObjectPrx.py
```diff
@@ -38,19 +38,17 @@ def uncheckedCast(type: Type[T], proxy: ObjectPrx | None, facet: str | None = No
...
         The source proxy.
-
     facet : str | None, optional
         A facet name.
 
     Returns
     -------
-    ObjectPrx | None
-        A new proxy with the requested type, or None if the source proxy is None.
+    T | None
```
externl, line 50: We should double check this works right with the type pyright. Mainly so that it indexed knows that the return type is an ObjectPrx.

python/python/Ice/ObjectPrx.py
externl: Yes, it should be milliseconds. We should check the other languages.

externl (approved):

python/python/Ice/ObjectPrx.py
InsertCreativityHere: Agreed, see: https://github.com/zeroc-ice/ice/issues/4892

python/modules/IcePy/Init.cpp
pepone: One minor issue, is we don't parse keyword arguments in the implementation.

Seems that Python allows to indicate this in the signature https://peps.python.org/pep-0570/


```suggestion
        R"(createProperties(args: list[str] | None = None, defaults: Ice.Properties | None = None, /) -> Ice.Properties
```

pepone (approved):

python/python/Ice/ObjectPrx.py
InsertCreativityHere: So, it's interesting. Sphinx seems to completely ignore the `T` and just replaces it with `ObjectPrx` in the docs. Probably because that's what it's bound to: `T = TypeVar("T", bound=ObjectPrx)`.

But looking at our other generic (in `Future.py`), it just slaps in the `_T` as-is. No link, no formatting, no extra meaning. Just `_T`.

Not sure there's anything to do about this though.

python/modules/IcePy/Init.cpp
InsertCreativityHere: Isn't this true of all of our functions in `IcePy`?
We use `PyArg_ParseTuple` basically everywhere to do the parsing of `args` into actual types, and from the docs:
> Parse the parameters of a function that takes only positional parameters into local variables.

https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTuple

There is a separate function for also parsing keyworded args (`PyArg_ParseTupleAndKeywords`), which 'rg' says we never call.

pepone: Yes, this is probably an issue with more IcePy functions.

It is probably not very important, at lest for the ones that are not re-exported.

No need to fix in this PR. Just saw the code and notice it didn't agree with the actual implementation.

InsertCreativityHere: I think it's worth fixing, even if indeed, it's very minor!
Will fix in a follow-up PR though.

PR #4893, opened 2026-01-15
Stop building ES5 JavaScript libraries with Babel

pepone:
This PR updates JavaScript builds to not generate ES5 libraries with babel.

There is no reason to support ES5 browser in 2026. It is also better to do babel compilation for the whole application if required, so that you target the specific platforms you want to support.

I also updated the gulp build to a more recent version, and updated a few build/test script to use ES modules. Ice 3.7 is still using cjs, and we are not changing this in a patch release.


externl (approved):
Looks good!

PR #4896, opened 2026-01-15
Fix linter warnings in GitHub composite actions

pepone:


.github/actions/build/action.yml
```diff
@@ -1,41 +1,30 @@
 name: Build Ice
+description: Builds Ice for the current platform using Make (macOS/Linux) or MSBuild (Windows).
 
 inputs:
   working_directory:
     description: "The working directory to run the build in"
     default: "."
-    type: string
```
pepone, line 7: type is not valid for actions only for workflows.

.github/actions/build/action.yml
```diff
@@ -1,41 +1,30 @@
...
   msbuild_command:
     description: "The msbuild command to use"
     default: "msbuild /m"
-    type: string
 
   build_cpp_and_python:
     description: "Build C++ and Python"
-    type: choice
     default: "false"
-    options:
```
pepone, line 28: ditto

.github/actions/documentation/action.yml
```diff
@@ -14,9 +14,6 @@ inputs:
     description: "GitHub token used by phive to access the GitHub API (to get the versions of the packages)"
     required: true
 
-# Disable the default permissions for the action.
-permissions: {}
```
pepone, line 18: You cannot change permissions in an action.

.github/actions/setup-cpp/action.yml
```diff
@@ -1,4 +1,5 @@
 name: Setup C++
+description: Installs C++ build dependencies and configures the build environment for the current platform.
```
pepone, line 2: Description is a required field.

PR #4898, opened 2026-01-15
Update Ice for C++ build instructions

pepone:
Update Ice for C++ 3.7 build instructions.

### DIFF
--- a/cpp/BUILDING.md
+++ b/cpp/BUILDING.md
@@ -2,13 +2,12 @@
 
 This document describes how to build and install Ice for C++ from source.
 
-ZeroC provides [Ice binary distributions] for many platforms and compilers, including Windows and Visual Studio, so
-building Ice from source is usually unnecessary.
+ZeroC provides [Ice binary distributions] for many platforms, including Linux, macOS and Windows, so building Ice from
+source is usually unnecessary.
 
 ## Table of Contents
 
 - [Build roadmap](#build-roadmap)
-- [Prerequisites](#prerequisites)
 - [Building Ice for C++ on Linux](#building-ice-for-c-on-linux)
   - [Installing build dependencies](#installing-build-dependencies)
   - [Building](#building)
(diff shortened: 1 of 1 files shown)


cpp/BUILDING.md
```diff
@@ -34,97 +31,58 @@ unnecessary.
...
 ### Operating Systems and Compilers
 
-Ice was extensively tested using the operating systems and compiler versions
-listed on [supported platforms][2].
+Ice was extensively tested using the operating systems and compiler versions listed on [supported platforms].
 
-On Windows, the build requires a recent version of Visual Studio. When building
-with Visual Studio 2017 you have to install Desktop development with C++ workload
-and select the optional component Windows 8.1 SDK and UCRT SDK.
+On Windows, the build requires Visual Studio 2019 or Visual Studio 2022. The built binaries are compatible with later
```
externl: No Visual Studio 2026?

pepone: You can use the resulting packages with Visual Studio 2026, but the build requires one of this versions.

The issue is the 3.7 NuGet dependencies doesn't support Visual Studio 2026, and this will require re-package all of them.

I think is fine for now, as we are the ones that build the packages most of the time.

cpp/BUILDING.md
```diff
@@ -1,31 +1,28 @@
...
   - [Operating Systems and Compilers](#operating-systems-and-compilers)
   - [Third-Party Libraries](#third-party-libraries)
-    - [AIX](#aix)
     - [Linux](#linux)
     - [macOS](#macos)
     - [Windows](#windows)
-- [Building Ice for AIX, Linux or macOS](#building-ice-for-aix-linux-or-macos)
-  - [Build configurations and platforms](#build-configurations-and-platforms)
-  - [C++11 mapping](#c11-mapping)
+- [Building Ice for C++ on Linux or macOS](#building-ice-for-c-on-linux-or-macos)
```
bernardnormier: We should keep the AIX instructions. We still support AIX on 3.7.

cpp/BUILDING.md
```diff
@@ -1,31 +1,28 @@
-# Building Ice for C++
+# Ice for C++ Build Instructions
```
bernardnormier: We should use the current (main/3.8) title:
Building Ice for C++ from Source

cpp/BUILDING.md
bernardnormier: I would say "the resulting binaries" or "the binaries produced by the build".

cpp/BUILDING.md
```diff
@@ -34,97 +31,58 @@ unnecessary.
...
-```
-
-In addition, on Ubuntu and Debian distributions where the Ice for Bluetooth
-plug-in is supported, you need to install the following packages in order to
-build the IceBT transport plug-in:
-
-- [pkg-config][9] 0.29 or later
-- [D-Bus][10] 1.10 or later
-- [BlueZ][11] 5.37 or later
+- [pkg-config] 0.29 or later
```
bernardnormier: We should remove these versions.

cpp/BUILDING.md
pepone: Reworked the document to align with 3.8/main structure.

bernardnormier (approved):

PR #4904, opened 2026-01-16
Simplify and modernize Python build instructions

pepone:


python/BUILDING.md
```diff
@@ -1,165 +1,165 @@
-# Ice for Python Build Instructions
+# Building Ice for Python from Source
```
externl, line 1: Our `BUILDING.md` files seem to have a mix of `Source` and `source`. I prefer `Source`.

python/BUILDING.md
```diff
@@ -1,165 +1,165 @@
...
+## Building Ice for Python from source
 
-* VS2015 x86 Native Tools Command Prompt
+This build compiles Ice for Python directly from the source tree and requires a prior build of Ice for C++.
 
-or
+### Prerequisites
 
-* VS2015 x64 Native Tools Command Prompt
+1. **Python 3.12 or later**
```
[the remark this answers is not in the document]
externl: What is the current min version for 3.7.11? Have we dropped all Python 2 OS builds?

python/BUILDING.md
```diff
@@ -1,165 +1,165 @@
...
-You can also download and install a [binary distribution].
+## Table of Contents
 
-* [Building with Pip](#building-with-pip)
-* [Building with Visual Studio 2015 and MSBuild (Python 3\.11 for Windows)](#building-with-visual-studio-2015-and-msbuild-python-311-for-windows)
-* [Building on Linux or macOS](#building-on-linux-or-macos)
-* [Configuring your Environment for Python](#configuring-your-environment-for-python)
-* [Running the Python Tests](#running-the-python-tests)
+* [Build roadmap](#build-roadmap)
+* [Building Ice for Python from source](#building-ice-for-python-from-source)
```
bernardnormier: I would prefer a different subtitle, which should match the description below, i.e.

- Source Build
 ...
- Python Package Build

python/BUILDING.md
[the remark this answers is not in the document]
pepone: Code is still there but we are not building Python 2x in any supported platforms.

bernardnormier (approved):

python/BUILDING.md
pepone: I will normalize this once all the updates are merged

python/BUILDING.md
pepone: Fixed as suggested

PR #4908, opened 2026-01-17
Simplify and modernize PHP build instructions

pepone:


php/BUILDING.md
```diff
@@ -1,130 +1,59 @@
...
+    c++(Ice for C++) --> php(Ice for PHP)
+    php --> tests(Tests)
 ```
 
-Ensure that `php` and `php-config` for the version of PHP you wish to
-build against are first in your PATH.
-
-Run `make` to build the extension.
-
-### Windows
```
pepone, line 55: Removed all the Windows setup as we no longer build it

php/BUILDING.md
```diff
@@ -136,227 +65,96 @@ Review the output for a line like this:
...
 
 ```shell
-set ICE_BIN_DIST=cpp
 python allTests.py
 ```
 
-If everything worked out, you should see lots of `ok` messages. In case of a
-failure, the tests abort with `failed`.
-
-## Web Server Permissions
```
pepone, line 316: These sections seems outdated and out of scoped for BUILDING.

php/BUILDING.md
```diff
@@ -1,130 +1,59 @@
...
 
-This page describes how to build and install Ice for PHP from source.
+This file describes how to build and install Ice for PHP from source.
+ZeroC provides [binary distributions] for many platforms, including Linux, so building Ice for PHP from source is
+usually unnecessary.
 
-ZeroC provides [binary distributions] for many platforms, including
-Linux and Windows, so building Ice for PHP from source is usually
-unnecessary.
+## Table of Contents
```
bernardnormier: Should not be a title. You should use the Markdown All in One extension to generate and maintain this TOC.

php/BUILDING.md
```diff
@@ -136,227 +65,96 @@ Review the output for a line like this:
...
-install them in a common directory, you may need to modify PHP's `include_path`
-directive so that the PHP interpreter is able to locate these files. For
-example, you could append to `php.ini`:
+The Ice extension makes no assumptions about the location of these files, so you can install them anywhere you like.
+For example, you can simply include them in the same directory as your application scripts. Alternatively, if you
+prefer to install them in a common directory, you may need to modify PHP's `include_path` directive so that the PHP
+interpreter is able to locate these files. For example, you could append to `php.ini`:
 
 ```ini
 include_path=${include_path}";C\ice\php\lib"
```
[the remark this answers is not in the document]
bernardnormier: Windows leftover?

php/BUILDING.md
externl: We also have and should fix this on main. I think in a few of these BUILDING files.

php/BUILDING.md
```diff
@@ -1,130 +1,59 @@
...
 
-This builds the extension with `Release` binaries for the default platform. The
-extension will be placed in `lib\x64\Release\php_ice.dll` for the `x64` platform
-and `lib\Win32\Release\php_ice.dll` for the `Win32` platform.
+### Prerequisites
 
-The default configuration builds the extension against the thread-safe PHP run
-time. You can build with the non-thread-safe run time using the `NTS-Release` or
-`NTS-Debug` configuration:
+1. **PHP 7.1 or later**
```
externl, line 25: PHP 5 missing?

pepone: I think that is fine, we keep the code, but it is probably not use anywhere.

PR #4911, opened 2026-01-20
Use MCPP 2.7.2.20

pepone:


.github/workflows/publish-rpm-packages.yml
```diff
@@ -56,14 +56,21 @@ jobs:
           if [[ "$DISTRIBUTION" == "rhel9" || "$DISTRIBUTION" == "rhel10" ]]; then
             echo "Also downloading MCPP RPMs for $DISTRIBUTION..."
 
+            # MCPP packages
+            if [[ "$CHANNEL" == "3.7" ]]; then
+              MCPP_PACKAGES="rpm-packages-3.7-${DISTRIBUTION}-*"
+            else
+              MCPP_PACKAGES="rpm-packages-3.8-${DISTRIBUTION}-*"
```
externl: I don't think we should keep 3.7 branch specific things on main.

pepone: Fixed

PR #4912, opened 2026-01-20
Use MCPP 2.7.2.20

pepone:


.github/workflows/publish-rpm-packages.yml
```diff
@@ -54,18 +54,23 @@ jobs:
...
         run: |
-
           # If building for RHEL 9 or 10, also download MCPP RPMs.
           echo "Also downloading MCPP RPMs for $DISTRIBUTION..."
 
+          # MCPP packages
+          if [[ "$CHANNEL" == "3.7" ]]; then
+            MCPP_PACKAGES="rpm-packages-3.7-${DISTRIBUTION}-*"
+          else
+            MCPP_PACKAGES="rpm-packages-3.8-${DISTRIBUTION}-*"
```
externl: I would remove this if statement is if it's not needed on this branch.

PR #4918, opened 2026-01-20
Fix incorrect invocation timeout doc-comments

### LINKED ISSUE #4892: Review our Documentation for `ice_invocationTimeout` and `ice_getInvocationTimeout`
While working on my Python doc-comment PR I found a doc-bug where we documented the value as being in seconds, when it was actually in milliseconds.

Looking at `main` and `3.7`, our doc-comments have a mix of reporting the value is in seconds and milliseconds, and at least some of these are probably wrong.

We should look through this function in each language and each branch and update them to be correct.

bernardnormier:
Fixes #4892.


externl (approved):

InsertCreativityHere (approved):
I assume this is all of them.
Looks good!

PR #4923, opened 2026-01-22
Fix sn.exe not found in PATH during strong name signing

pepone:
## Summary
- Fix strong name signing failure when building .NET packages with `dotnet msbuild`
- The `sn.exe` tool is not in PATH by default on GitHub Actions Windows runners
- Added logic to locate `sn.exe` in Windows SDK directories (NETFX 4.8.2, 4.8.1, 4.8)

## Test plan
- [x] Tested with [build-dotnet-packages workflow](https://github.com/zeroc-ice/ice/actions/runs/21243357704) on this branch
- [ ] Verify nightly builds succeed after merge

### DIFF
--- a/.github/workflows/build-dotnet-packages.yml
+++ b/.github/workflows/build-dotnet-packages.yml
@@ -181,5 +181,5 @@ jobs:
         with:
           name: dotnet-nuget-packages
           path: |
-            csharp/msbuild/zeroc.ice.net/*.nupkg
-            csharp/msbuild/zeroc.ice.net/*.snupkg
+            csharp/msbuild/*.nupkg
+            csharp/msbuild/*.snupkg


csharp/msbuild/ice.sign.targets
```diff
@@ -11,13 +11,23 @@
         <AssemblyOriginatorKeyFile Condition="'$(PUBLIC_KEYFILE)' == ''">$(KEYFILE)</AssemblyOriginatorKeyFile>
     </PropertyGroup>
 
+  <!-- Locate sn.exe in Windows SDK. NetFxToolsDirectory is not set when using dotnet msbuild. -->
```
externl: Should we just ensure this is set in CI instead? Hardcoding various paths here seems brittle.

pepone: Maybe use it from PATH, and update CI to add the correct directory to PATH?

PR #4924, opened 2026-01-22
Add job-level timeout to CI workflow

### LINKED ISSUE #4873: Review CI job timeouts
This took 5h until it was canceled

https://github.com/zeroc-ice/ice/actions/runs/20866364167/job/60004932749?pr=4865

pepone:
## Summary
- Add a 180-minute (3 hour) job-level timeout to the CI workflow
- Prevents CI jobs from running indefinitely when individual steps hang
- Individual step timeouts exist but the job itself had no timeout, allowing it to run for GitHub's default 6-hour limit

See https://github.com/zeroc-ice/ice/actions/runs/20866364167/job/60004932749 where a job ran for ~6 hours before being cancelled.

Closes #4873

## Test plan
- [ ] CI workflow runs successfully with the new timeout

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,6 +140,7 @@ jobs:
             build_cpp_and_python: true
 
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 180
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4


.github/workflows/ci.yml
```diff
@@ -140,6 +140,7 @@ jobs:
             build_cpp_and_python: true
 
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 180
```
externl, line 143: Doing this kills means we skip the scripts that run on failure, such as collecting logs and digs. It's better to set timeouts on individual steps.

pepone: Note that is an upper limit 3hours. We can stick to the default 6hours. Individual steps should already use lower timeouts.

pepone:
Let's keep the default

PR #4930, opened 2026-01-22
Add CI workflow for Linux distribution testing

pepone:
Build and run C++98/C++11 tests using Docker images from the package build workflows.

### DIFF
--- a/.github/workflows/ci-linux.yml
+++ b/.github/workflows/ci-linux.yml
@@ -169,7 +169,14 @@ jobs:
               set -ex
               make CONFIGS=${{ matrix.config }} -j\$(nproc) srcs
               make CONFIGS=${{ matrix.config }} -j\$(nproc) tests
-              python3 allTests.py --config=${{ matrix.config }} --workers=4 --debug --continue
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-linux.yml
```diff
@@ -0,0 +1,181 @@
+name: CI Linux
```
externl, line 1: I wonder if we should phrase it more as CI Docker? Or CI All Platforms. Since we do actually do linux in ci.yml right?

.github/workflows/ci-linux.yml
```diff
@@ -0,0 +1,181 @@
...
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: zeroc-ice
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and test C++ (${{ matrix.config }})
```
externl, line 162: Maybe we should add a timeout to this step.

PR #4976, opened 2026-01-29
Standardize Platform and Configuration defaults to x64/Release

pepone:
Update all MSBuild ice.proj files and Gradle properties to consistently default to x64 platform and Release configuration, matching the top-level ice.proj defaults.

### DIFF
--- a/cpp/msbuild/ice.proj
+++ b/cpp/msbuild/ice.proj
@@ -1,15 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-    <PropertyGroup Condition="'$(Configuration)' == ''">
-        <Configuration>Release</Configuration>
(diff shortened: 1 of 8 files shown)


externl (approved):
Looks good. Does allTests.py use the same defaults as well?

cpp/msbuild/ice.proj
```diff
@@ -1,15 +1,9 @@
...
-        <Configuration>Release</Configuration>
-    </PropertyGroup>
-
-    <PropertyGroup Condition="'$(Platform)' == ''">
-        <Platform>Win32</Platform>
-    </PropertyGroup>
-
     <PropertyGroup>
+        <Configuration Condition="'$(Configuration)' == ''">Release</Configuration>
+        <Platform Condition="'$(Platform)' == ''">x64</Platform>
```
InsertCreativityHere, line 6: Glad to see `x64` as the default now!

InsertCreativityHere (approved):
Also looks good to me!

PR #4977, opened 2026-01-29
Document libstdc++ compatibility workaround for MATLAB

pepone:


externl (approved):
Looks good. I agree with copilot. we should update the ToC.

PR #4979, opened 2026-01-29
Update Android test emulator to use android-36.1 system image

pepone:
Also added --force flag to avdmanager create to handle cases where the AVD directory already exists from a previous run.

### DIFF
--- a/java-compat/BUILDING.md
+++ b/java-compat/BUILDING.md
@@ -174,6 +174,14 @@ export PATH=~/Library/Android/sdk/platform-tools:$PATH
 export PATH=~/Library/Android/sdk/emulator:$PATH
 ```
 
+On Linux:
+
+```shell
+export PATH=~/Android/Sdk/cmdline-tools/latest/bin:$PATH
+export PATH=~/Android/Sdk/platform-tools:$PATH
+export PATH=~/Android/Sdk/emulator:$PATH
+```
+
 On Windows:
 
 ```shell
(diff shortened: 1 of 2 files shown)


scripts/Util.py
```diff
@@ -3304,7 +3304,7 @@ def _getDefaultSource(self, processType):
         }[processType]
 
     def getSDKPackage(self):
-        return "system-images;android-33;google_apis;{}".format(
+        return "system-images;android-36.1;google_apis_playstore;{}".format(
```
InsertCreativityHere, line 3307: I'm not sure what the difference is, but both of these packages are available, and not the same:
```
system-images;android-36.1;google_apis;
system-images;android-36.1;google_apis_playstore;
```
And right now, for `3.7.11` and `3.8.01, we're using the non-playstore version.

scripts/Util.py
```diff
@@ -2474,7 +2474,7 @@ def startControllerApp(self, current, ident):
                 pass
             # The SDK is downloaded by test VMs instead of here.
             #run("sdkmanager \"{0}\"".format(sdk), stdout=True, stdin="yes", stdinRepeat=True) # yes to accept licenses
-            run("avdmanager -v create avd -k \"{0}\" -d \"Nexus 6\" -n IceTests".format(sdk))
+            run("avdmanager -v create avd -k \"{0}\" -d \"Nexus 6\" -n IceTests --force".format(sdk))
```
InsertCreativityHere, line 2477: Didn't know this flag existed! I just manually deleted the avd directory the couple times it failed to remove properly. Cool!

scripts/Util.py
pepone: Seems the one with `_playstore` gets installed by default with a fresh Android Studio install.

With a fresh Android Studio install I see:

```
jose@casiopea:~/Documents/ZeroC/ice/java$ sdkmanager --list_installed | grep system-images
  system-images;android-36.1;google_apis_playstore;x86_64 | 4       | Google Play Intel x86_64 Atom System Image | system-images/android-36.1/google_apis_playstore/x86_64
```

PR #4987, opened 2026-01-30
Remove bzip2 dependency from zeroc.ice.net NuGet package

pepone:
iceboxnet doesn't use compression, so bzip2 is not needed.

### DIFF
--- a/csharp/msbuild/THIRD_PARTY_LICENSE.txt
+++ b/csharp/msbuild/THIRD_PARTY_LICENSE.txt
@@ -1,6 +1,3 @@
-License agreement for bzip2/libbzip2:
---------------------------------------
-
 License agreement for MCPP:
 ----------------------------
 


csharp/msbuild/THIRD_PARTY_LICENSE.txt
```diff
@@ -1,48 +1,6 @@
 License agreement for bzip2/libbzip2:
 --------------------------------------
 
---------------------------------------------------------------------------
```
bernardnormier, line 4: You should also remove the title above.

csharp/msbuild/ice.proj
```diff
@@ -80,11 +80,6 @@
         >$([System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture.ToString().ToLower())</IceOSArch>
         <IceSliceCompilerExe Condition="'$(OS)' == 'Windows_NT'">slice2cs.exe</IceSliceCompilerExe>
         <IceSliceCompilerExe Condition="'$(OS)' != 'Windows_NT'">slice2cs</IceSliceCompilerExe>
-        <!-- bzip2 package properties (Windows only, restored via NuGet) -->
```
bernardnormier, line 83: I believe we still need to install bzip2.dll, when we test compression on Windows.

pepone: This only run with the packaging target.

The tests seems to rely on the one from C++ builds

https://github.com/zeroc-ice/ice/blob/eb23cd5f8da9aa945bbad79a66666fc6afc6224d/scripts/Util.py#L3427-L3429

This is fine because we always need to run nuget restore on C++ to build slice2cpp.

PR #4992, opened 2026-01-31
Remove Python type info validation

### LINKED ISSUE #4734: Remove Python type info validation
The IcePy extensions validate Types, to ensure dispatch and invocation are called with the expected types. This type validation is redundant with the validation done by the type checker now that we support type hints.

It would be better to remove the validate methods, and then throw if marshaling got an unexpected type.

see:

https://github.com/zeroc-ice/ice/blob/c53c9239a8c938b75eb1ece377270c8d3d792ce0/python/modules/IcePy/Operation.cpp#L704

pepone:
## Summary

- Remove redundant `validate()` methods from TypeInfo classes in IcePy
- Type validation is now handled directly in `marshal()` methods, which throw appropriate Python exceptions (ValueError) when marshaling encounters unexpected types
- This eliminates duplicate validation since Python type hints provide static type checking at development time

Fixes #4734

## Test plan

- [x] Build C++ library with `make -C cpp -j20`
- [x] Build Python bindings with `make -C python -j20`
- [x] Run Python tests with `python3 python/allTests.py --all` - all 33 tests pass
- [x] Verify ValueError exceptions are still raised for invalid values (out-of-range byte, short, int, float)

### DIFF
--- a/python/modules/IcePy/Types.cpp
+++ b/python/modules/IcePy/Types.cpp
@@ -47,7 +47,7 @@ namespace
     // Wraps the current Python exception in a ValueError with the original exception as the cause.
     // TODO: In the next major release, we can remove this helper and avoid wrapping the original exceptions.
     //
-    void wrapRaisedExceptionWithValueError(const char* message)
+    void wrapRaisedExceptionInValueError(const char* message)
     {
         assert(PyErr_Occurred());
         PyObject* originalException = PyErr_GetRaisedException();
@@ -620,7 +620,7 @@ IcePy::PrimitiveInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap*, bo
             int isTrue = PyObject_IsTrue(p);
             if (isTrue < 0)
             {
-                wrapRaisedExceptionWithValueError("invalid value for bool type");
+                wrapRaisedExceptionInValueError("invalid value for bool type");
                 throw AbortMarshaling();
             }
             os->write(isTrue ? true : false);
@@ -631,7 +631,7 @@ IcePy::PrimitiveInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap*, bo
             long val = PyLong_AsLong(p);
             if (PyErr_Occurred())
             {
-                wrapRaisedExceptionWithValueError("invalid value for byte type");
+                wrapRaisedExceptionInValueError("invalid value for byte type");
                 throw AbortMarshaling();
             }
             if (val < 0 || val > 255)
@@ -647,7 +647,7 @@ IcePy::PrimitiveInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap*, bo
             long val = PyLong_AsLong(p);
             if (PyErr_Occurred())
             {
-                wrapRaisedExceptionWithValueError("invalid value for short type");
+                wrapRaisedExceptionInValueError("invalid value for short type");
                 throw AbortMarshaling();
             }
             if (val < SHRT_MIN || val > SHRT_MAX)
@@ -663,7 +663,7 @@ IcePy::PrimitiveInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap*, bo
             long val = PyLong_AsLong(p);
             if (PyErr_Occurred())
             {
-                wrapRaisedExceptionWithValueError("invalid value for int type");
+                wrapRaisedExceptionInValueError("invalid value for int type");
                 throw AbortMarshaling();
             }
             if (val < INT_MIN || val > INT_MAX)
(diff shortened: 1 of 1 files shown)


python/modules/IcePy/Operation.cpp
```diff
@@ -693,87 +693,58 @@ Operation::marshalResult(Ice::OutputStream& os, PyObject* result)
...
-        PyObject* arg = PyTuple_GET_ITEM(resultTuple.get(), info->pos);
-        if ((!info->optional || arg != Py_None) && !info->type->validate(arg))
+        // Marshal the required out parameters.
+        for (const auto& info : outParams)
         {
-            // TODO: Provide the parameter name instead?
-            ostringstream ostr;
-            ostr << "cannot marshal result: invalid value for out argument " << (info->pos + 1) << " in operation '"
-                 << mappedName << "'";
-            if (PyErr_Occurred())
```
pepone, line 710: This code was here to prevent returning a type that didn't match the declared type `info->type`.

python/modules/IcePy/Operation.cpp
```diff
@@ -693,87 +693,58 @@ Operation::marshalResult(Ice::OutputStream& os, PyObject* result)
...
-                try
-                {
-                    throwPythonException();
-                }
-                catch (const Ice::LocalException& ex)
-                {
-                    ostr << "':\n" << ex.what();
-                }
+                PyObject* arg = PyTuple_GET_ITEM(resultTuple.get(), info->pos);
+                info->type->marshal(arg, &os, &objectMap, false, &info->metadata);
```
pepone, line 705: We still check for errors during marshaling, were we need to convert types from python to C++ for writing to the underlying C++ OutputStream.

python/modules/IcePy/Types.cpp
```diff
@@ -42,6 +42,19 @@ namespace
     class InvalidSequenceFactoryException
     {
     };
+
+    //
+    // Wraps the current Python exception in a ValueError with the original exception as the cause.
+    // TODO: In the next major release, we can remove this helper and avoid wrapping the original exceptions.
+    //
+    void wrapRaisedExceptionWithValueError(const char* message)
```
bernardnormier: Should this be `wrapRaisedExceptionInValueError`?

python/modules/IcePy/Types.cpp
```diff
@@ -721,58 +620,96 @@ IcePy::PrimitiveInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap*, bo
             int isTrue = PyObject_IsTrue(p);
             if (isTrue < 0)
             {
-                assert(false); // validate() should have caught this.
+                wrapRaisedExceptionWithValueError("invalid value for bool type");
```
bernardnormier: wrapRaisedException... is all about wrapping the current Python exception. When `isTrue < 0`, we have a current Python exception?

python/modules/IcePy/Types.cpp
```diff
@@ -42,6 +42,19 @@ namespace
     class InvalidSequenceFactoryException
     {
     };
+
+    //
+    // Wraps the current Python exception in a ValueError with the original exception as the cause.
+    // TODO: In the next major release, we can remove this helper and avoid wrapping the original exceptions.
```
bernardnormier, line 48: I don't understand the compatibility issue here.

externl, line 48: I would open an issue to track this.

python/modules/IcePy/Types.cpp
```diff
@@ -1892,7 +1803,13 @@ IcePy::SequenceInfo::marshalPrimitiveSequence(const PrimitiveInfoPtr& pi, PyObje
 
                 if (!writeString(item, os))
                 {
-                    assert(PyErr_Occurred());
+                    if (!PyErr_Occurred())
```
bernardnormier, line 1797: I find the `if (item != Py_None && !checkString(item))` block above redundant. writeString calls checkString too.

python/modules/IcePy/Types.cpp
pepone: Yes `PyObject_IsTrue` returns `-1` on failure. https://docs.python.org/3/c-api/object.html

python/modules/IcePy/Types.cpp
pepone: We were doing this checks in validate and throwing ValueError, the exceptions set by Python might be different. For example `PyLong_AsLong` can now set `OverflowError` for conditions that we previously handle in validate. 

Rather than changing the exceptions we throw, I keep the python exception as the cause.

python/modules/IcePy/Types.cpp
pepone: You are right, writeString already handles this.

PR #4994, opened 2026-01-31
Add CI workflow to test PIP package

### LINKED ISSUE #4946: Run Python tests with PIP package
We can run the Python test suite with the wheel by:

- Build using `python -m build`
- Install the package in an venv and run the tests from there

pepone:
## Summary

- Add a new GitHub Actions workflow (`test-pip-package.yml`) that tests the Python wheel package
- Builds the wheel using `python -m build`
- Installs the wheel in a virtual environment
- Runs a subset of pure Python tests using dynamic Slice loading (`--load-slice`)

Fixes #4946

### DIFF
--- a/.github/workflows/test-pip-package.yml
+++ b/.github/workflows/test-pip-package.yml
@@ -7,11 +7,13 @@ on:
     paths:
       - "python/**"
       - "slice/**"
+      - ".github/actions/setup-python/**"
   pull_request:
     branches: ["main"]
     paths:
       - "python/**"
       - "slice/**"
+      - ".github/actions/setup-python/**"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-pip-package.yml
```diff
@@ -0,0 +1,70 @@
+name: "Test PIP Package"
+
+on:
+  workflow_dispatch:
+  push:
+    branches: ["main"]
+    paths:
+      - "python/**"
```
externl, line 8: We may also want to include the path to the setup-python action: `./.github/actions/setup-python`

PR #4995, opened 2026-01-31
Port MATLAB build changes from 3.7

### LINKED ISSUE #4983: Update MATLAB building prerequisites
https://github.com/zeroc-ice/ice/blob/main/matlab/BUILDING.md#building-ice-for-matlab-on-linux

The Linux section is outdated, we should document what we actually use in CI. See 3.7 updates.

### LINKED ISSUE #4984: Port MATLAB build changes from 3.7
See https://github.com/zeroc-ice/ice/pull/4982

We want to do the same to avoid the need for LD_PRELOAD libstdc++ and to use the MATLAB bundled SSL libraries.

pepone:
## Summary

Port MATLAB build fixes from PR #4982 (3.7 branch) to main:

- Remove `LD_PRELOAD` workaround for libstdc++ compatibility
- Build with g++-11 for compatible libstdc++ with MATLAB's bundled library
- Add `USE_MATLAB_SSL` option to link with MATLAB's bundled OpenSSL libraries (`libmwssl`, `libmwcrypto`)
- Update CI workflows to use the new build configuration
- Update MATLAB building prerequisites documentation for Linux (#4983)

Fixes #4984
Fixes #4983

## Files Changed

| File | Change |
|------|--------|
| `.github/actions/setup-matlab/action.yml` | Remove LD_PRELOAD workaround |
| `config/Make.rules.Linux` | Add USE_MATLAB_SSL conditional for MATLAB OpenSSL |
| `.github/workflows/build-matlab-packages.yml` | Install g++-11, set CXX and USE_MATLAB_SSL |
| `.github/workflows/ci.yml` | Add MATLAB build environment setup step |
| `matlab/BUILDING.md` | Update Linux prerequisites to reflect CI requirements |

## Test plan

- [ ] MATLAB toolbox builds successfully on Linux
- [ ] MATLAB tests pass without LD_PRELOAD workaround

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,17 +194,6 @@ jobs:
         if: matrix.config == 'matlab'
         uses: ./.github/actions/setup-matlab
 
-      - name: Setup MATLAB Build Environment
-        if: matrix.config == 'matlab' && runner.os == 'Linux'
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y g++-11
-          echo "CXX=g++-11" >> $GITHUB_ENV
-          echo "USE_MATLAB_SSL=yes" >> $GITHUB_ENV
-          # g++-11 headers only support _FORTIFY_SOURCE up to 2, but Ubuntu 24.04's dpkg-buildflags sets it to 3
-          echo "CXXFLAGS=-D_FORTIFY_SOURCE=2" >> $GITHUB_ENV
-        shell: bash
-
       - name: Setup Android
         if: matrix.config == 'android'
         uses: ./.github/actions/setup-android
--- a/matlab/BUILDING.md
+++ b/matlab/BUILDING.md
@@ -33,7 +33,7 @@ flowchart LR
 
 ### Prerequisites
 
-1. **MATLAB R2025b or later**
+1. **A recent version of MATLAB**
 
 2. **Visual Studio 2022**
 
@@ -87,9 +87,9 @@ You can install the toolbox from within MATLAB by double-clicking on the file.
 
 ### Prerequisites
 
-1. **MATLAB R2025b or later**
+1. **A recent version of MATLAB**
 
-2. **Ubuntu 24.04** with the default C++ compiler
+2. **GCC 11 or greater**
 
 3. **Ice for C++ source build** (using the `shared` configuration)
 
--- a/.github/actions/setup-matlab/action.yml
+++ b/.github/actions/setup-matlab/action.yml
@@ -35,6 +35,17 @@ runs:
       shell: bash
       if: runner.os == 'Linux'
 
+    - name: Setup MATLAB Build Environment
+      if: runner.os == 'Linux'
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y g++-11
+        echo "CXX=g++-11" >> $GITHUB_ENV
+        echo "USE_MATLAB_SSL=yes" >> $GITHUB_ENV
+        # g++-11 headers only support _FORTIFY_SOURCE up to 2, but Ubuntu 24.04's dpkg-buildflags sets it to 3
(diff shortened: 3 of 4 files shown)


matlab/BUILDING.md
```diff
@@ -74,72 +87,62 @@ You can install the toolbox from within MATLAB by double-clicking on the file.
 
 ### Prerequisites
 
-The build system requires MATLAB 2017b or MATLAB 2019b for Linux, with a supported C++ compiler and a Perl installation.
+1. **MATLAB R2025b or later**
 
-- Use GCC 4.9 for 2017b, see [MATLAB 2017b supported compilers][2].
-- Use GCC 6.3 for 2019b, see [MATLAB 2019b supported compilers][3].
+2. **Ubuntu 24.04** with the default C++ compiler
```
InsertCreativityHere: If users are using the stock gcc which ships with Ubuntu, won't they still need to set `LD_PRELOAD`?
I saw that we fixed the OpenSSL issue in our make.rules, but that doesn't fix the `GLIBCXX` symbols being too new for MATLAB, right?

Or alternatively we could tell them to install `g++-11` like we do.
I imagine this is a bigger pain than just setting an environment variable though.

pepone: To avoid the LD_PRELOAD workaround you need to build with g++11, which uses a stdc++ lib compatible with the one included by MATLAB.

.github/workflows/ci.yml
```diff
@@ -194,6 +194,17 @@ jobs:
         if: matrix.config == 'matlab'
         uses: ./.github/actions/setup-matlab
 
+      - name: Setup MATLAB Build Environment
```
externl: should we move this into setup-matlab?

matlab/BUILDING.md
InsertCreativityHere: Sure, but then we should mention this here, instead of just saying "the default compiler".

matlab/BUILDING.md
```diff
@@ -1,35 +1,48 @@
...
+    matlab --> toolbox(Toolbox Package)
+```
 
 ## Building Ice for MATLAB on Windows
 
 ### Prerequisites
 
-The build system requires MATLAB 2016a or later for Windows, Visual Studio 2022
-and a Perl installation.
+1. **MATLAB R2025b or later**
```
bernardnormier: I am not fond of all these very specific requirements in the BUILDING.md files. I would prefer something less specific like "a recent version of MATLAB", a recent version of Visual Studio.

In practice, there is some leeway for these versions. You should be able to make it work for R2025a and VS2026. If you want to know the _supported_ versions, check the supported platforms page.

matlab/BUILDING.md
```diff
@@ -74,72 +87,62 @@ You can install the toolbox from within MATLAB by double-clicking on the file.
 
 ### Prerequisites
 
-The build system requires MATLAB 2017b or MATLAB 2019b for Linux, with a supported C++ compiler and a Perl installation.
+1. **MATLAB R2025b or later**
```
bernardnormier: Same here, except it could be something like:
 - a recent version of MATLAB for Linux
 - GCC 11 or greater

and not mention a specific Linux distribution

matlab/BUILDING.md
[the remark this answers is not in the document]
bernardnormier: I would not get into these details. Maybe MATLAB 2026b will work "out of the box" with the default GCC on Ubuntu 24.04.

If you want to see how we do it, check the action scripts.

pepone: I removed the details as per Bernard suggestion, and just say GCC 11 or greater

matlab/BUILDING.md
pepone: Fixed

.github/workflows/ci.yml
pepone: Fixed as suggested.

PR #4997, opened 2026-01-31
Add address information to JavaScript socket exceptions

### LINKED ISSUE #389: Add address information to socket exception or a new exception

pepone:
## Summary

Include the peer address in exception messages for `ConnectFailedException`, `ConnectionLostException`, and `ConnectionRefusedException` to help users identify the origin of network failures.

Fixes #389

This completes the JavaScript implementation requested in the issue. The same fix was already implemented for:
- C++ (#4680)
- C# (#4683)
- Java (#4684)

## Changes

- **LocalExceptions.js/d.ts**: Added constructors to socket exception classes that accept an optional address parameter and format it into the message
- **TcpTransceiver.js**: Updated `translateError()` to include address in exceptions
- **WSTransceiver.js**: Updated exception creation to include the WebSocket URL

## Example Output

Before:
```
Ice.ConnectFailedException: connect failed
```

After:
```
Ice.ConnectFailedException: Failed to connect to 127.0.0.1:9999.
```


js/src/Ice/TcpTransceiver.js
```diff
@@ -264,19 +264,20 @@ function fdToString(fd, targetAddr) {
     return addressesToString(fd.localAddress, fd.localPort, fd.remoteAddress, fd.remotePort, targetAddr);
 }
 
-function translateError(state, err) {
+function translateError(state, err, addr) {
+    const address = addr ? `${addr.host}:${addr.port}` : null;
     if (!err) {
```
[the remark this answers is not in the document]
pepone: I'm going to refactor this translate error in a separate PR, this issue is not related to the addition of the address.

js/src/Ice/LocalExceptions.d.ts
```diff
@@ -163,17 +170,23 @@ declare module "@zeroc/ice" {
         /**
          * The exception that is thrown when a connection establishment fails.
          */
-        class ConnectFailedException extends SocketException {}
+        class ConnectFailedException extends SocketException {
+            constructor(peerAddress?: string, options?: ErrorOptions);
```
bernardnormier, line 182: Why not a simpler:
```js
constructor(peerAddress?: string, cause?: unknown);
```

I don't see the benefit of introducing this ErrorOptions interface.

externl: I agree, we only have one error option.

pepone: We can pass cause directly, but this is the standard Error argument in JavaScript

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

externl: Okay, let's keep it then!

bernardnormier: It would be nice to add a comment that references this Error constructor.

PR #4999, opened 2026-02-01
Auto-detect Ice slice directory in CompileSlice plugin

### LINKED ISSUE #4882: Improve the Swift Slice plugin to include ice slice in search path
Currently to use Slice files from Ice package you have to manually add them to the `search_paths` in the plug-in configuration.

see for example:

https://github.com/zeroc-ice/ice-demos/blob/64ec7d12ea1ad2595a9a76408ea16b7a589e369f/swift/Glacier2/Session/Sources/Client/slice-plugin.json#L3

It would be nice to always include the slice directory from the Ice package in the search path. I think this is consistent with what we do in other languages, where there is no extra setup for use them.

pepone:
## Summary

- CompileSlice plugin now automatically includes the Ice slice directory in the search path
- Uses `#filePath` to derive the slice directory from the plugin's source location
- Validates by checking for `Ice/Identity.ice`
- Removes need for manual `search_paths` configuration for Ice includes

## Changes

- Updated `CompileSlicePlugin.swift` with automatic slice directory detection
- Removed redundant `search_paths` from Glacier2, IceGrid, IceStorm configs
- Deleted 10 test config files that only had `search_paths`

Fixes #4882

## Test plan

- [x] Build test package with `swift build --target Ice_ami`
- [x] Verify correct search path in verbose output (`-I.../ice/slice`)
- [x] Run swift-format and lint

### DIFF
--- a/swift/Plugins/CompileSlice/CompileSlicePlugin.swift
+++ b/swift/Plugins/CompileSlice/CompileSlicePlugin.swift
@@ -139,7 +139,7 @@ struct CompileSlicePlugin {
             // Add additional slice files from config.sources. The files are relative to the config file location.
             let baseDirectory = configFileURL.deletingLastPathComponent()
             for source in config.sources ?? [] {
-                let sourceFileOrDirectory = baseDirectory.appendingPathComponent(source)
+                let sourceFileOrDirectory = baseDirectory.appending(path: source)
                 if sourceFileOrDirectory.pathExtension == "ice" {
                     sliceSources.append(sourceFileOrDirectory)
                 } else {
@@ -154,7 +154,7 @@ struct CompileSlicePlugin {
             // Add additional search paths from config.search_paths.
             // These paths are relative to the config file location.
             for path in config.search_paths ?? [] {
-                searchPaths.append("-I\(baseDirectory.appendingPathComponent(path).path)")
+                searchPaths.append("-I\(baseDirectory.appending(path: path).path)")
             }
         }
 
@@ -165,7 +165,7 @@ struct CompileSlicePlugin {
 
         // Create the build commands for each slice file.
         return sliceSources.map { sliceSource in
-            let outputFile = outputDir.appendingPathComponent(sliceSource.lastPathComponent)
+            let outputFile = outputDir.appending(path: sliceSource.lastPathComponent)
                 .deletingPathExtension()
                 .appendingPathExtension("swift")
 


swift/Plugins/CompileSlice/CompileSlicePlugin.swift
```diff
@@ -13,7 +13,8 @@ extension CompileSlicePlugin: BuildToolPlugin {
         return try createBuildCommands(
             outputDir: context.pluginWorkDirectoryURL,
             inputFiles: swiftTarget.sourceFiles,
-            slice2swift: try context.tool(named: "slice2swift").url
+            slice2swift: try context.tool(named: "slice2swift").url,
+            iceSliceDir: CompileSlicePlugin.iceSliceDir
```
bernardnormier: I don't see the logic of this iceSliceDir parameter. As far as I cant tell, the createBuildCommands function can just retrieve isSliceDir. It doesn't depend on the args to createBuildCommands.

pepone: You are right, removed the arg.

swift/Plugins/CompileSlice/CompileSlicePlugin.swift
```diff
@@ -109,8 +130,11 @@ struct CompileSlicePlugin {
             .first(where: { $0.url.lastPathComponent == Self.configFileName })?
             .url
 
-        // By default we use an empty search path for Slice compilation.
+        // Start with the Ice slice directory in the search path (if found).
```
bernardnormier: I think the Ice-Slice directory should be last in the include path, not first. This way, you could for example overwrite some built-in definitions, like build your own extended `IceStorm/IceStorm.ice` file. That's much harder if the built-in definitions are included first.

Do we include these definitions first in other Slice tools?

swift/Plugins/CompileSlice/CompileSlicePlugin.swift
```diff
@@ -79,18 +80,38 @@ struct Config: Codable {
...
 
+    /// The Ice slice directory, derived from this plugin's source file location.
+    /// Path: CompileSlicePlugin.swift -> CompileSlice -> Plugins -> swift -> (ice root) -> slice
+    private static let iceSliceDir: URL? = {
+        var url = URL(fileURLWithPath: #filePath)
+        for _ in 0..<4 {
+            url.deleteLastPathComponent()
+        }
+        url.append(path: "slice")
+        let identityIce = url.appending(path: "Ice/Identity.ice")
```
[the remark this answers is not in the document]
externl: We should update everything to use `append`. The other one is deprecated. https://developer.apple.com/documentation/foundation/url/appendingpathcomponent(_:)

swift/Plugins/CompileSlice/CompileSlicePlugin.swift
pepone: In C# (3.8) we append them. Each package does:

```
<!-- Copyright (c) ZeroC, Inc. -->
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
    <ItemDefinitionGroup>
        <SliceCompile>
            <IncludeDirectories>%(IncludeDirectories);$(MSBuildThisFileDirectory)..\slice</IncludeDirectories>
        </SliceCompile>
    </ItemDefinitionGroup>
</Project>
```

C++ MSBuild:

```
            <ItemDefinitionGroup>
                <SliceCompile>
                    <IncludeDirectories>$(MSBuildThisFileDirectory)..\slice</IncludeDirectories>
                </SliceCompile>
            </ItemDefinitionGroup>
```

Users can overwrite, append or prepend.

bernardnormier: Ok, since we provide less flexibility in Swift, I would put the built-in Ice-Slice files last in the include path.

PR #5006, opened 2026-02-03
Make ProtocolInstance field final.

pepone:
This merges the 3.7.2-1 patch from https://github.com/zeroc-ice/ice/compare/v3.7.2...v3.7.2.1-java-npe-fix into 3.7 branch.


java/src/Ice/src/main/java/com/zeroc/IceInternal/UnderlyingEndpointFactory.java
```diff
@@ -69,15 +70,14 @@ public void destroy()
         {
             _factory.destroy();
         }
-        _instance = null;
```
bernardnormier, line 72: The main "fix" here (and other file) to leave _instance alone in destroy.

bernardnormier (approved):

PR #5007, opened 2026-02-03
Use prebuilt slice2swift artifact bundle for SwiftPM

### LINKED ISSUE #4996: Prebuilt slice2swift artifact bundle for Ice Swift (remove MCPP source build)
Ice for Swift currently builds slice2swift from source, which pulls in MCPP. We should ship slice2swift as a SwiftPM artifact bundle so consumers get a prebuilt binary and don’t need to build MCPP. This will simplify SwiftPM/Xcode usage and align with how we distribute other binaries.

pepone:
Replace the source-built slice2swift executable target with a prebuilt artifact bundle. This removes the mcpp dependency from Package.swift and improves build times for SwiftPM users.

Changes:
- Add Make.artifactbundle.rules to build slice2swift.artifactbundle.zip
- Update Package.swift to use binaryTarget for slice2swift
- Rename workflows: build/publish-xcframework-packages → build/publish-swift-artifacts
- Add CI steps to build, upload, and publish the artifact bundle

Fix #4996

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -146,9 +146,6 @@ compile_commands.json
 .build
 .swiftpm
 
-# slice2swift artifact bundle (built on-demand for SwiftPM)
-cpp/bin/slice2swift.artifactbundle
-
 # LLVM Code Coverage maps
 *.profraw
 *.profdata


bernardnormier:
> As commented previously, I don't think we need an artifact bundle for slice2swift. My understanding is that artifact bundles are for cross-platform projects, e.g. if we wanted to support both macOS and Linux. Until we add cross platform support (if we ever do), I would use a simple zip of slice2swift. Not an artifact bundle.

I missed the reply on the previous PR.

.github/workflows/build-swift-artifacts.yml
```diff
@@ -1,4 +1,4 @@
-name: "Build Brew Packages"
+name: "Build Swift Artifacts"
```
bernardnormier: As I understand it, we're building C++ code (XCFramework packages and the slice2swift artifact bundle).

Describing these packages as "Swift artifacts" is a bit confusing - it's really C++ packages later used by Ice for Swift. So maybe change that to: "Build C++ Artifacts for Swift" or "Build C++ Packages for Swift" or "Build C++ Dependencies for Swift".

.github/workflows/build-swift-artifacts.yml
```diff
@@ -22,7 +22,7 @@ permissions:
   contents: write
 
 jobs:
-  build-xcfranework-packages:
+  build-swift-artifacts:
```
bernardnormier: Same for build-swift-artifacts, publish-swift-artifacts, etc.

.github/workflows/build-swift-artifacts.yml
pepone: Renamed the workflows.

.gitignore
```diff
@@ -146,6 +146,9 @@ compile_commands.json
 .build
 .swiftpm
 
+# slice2swift artifact bundle (built on-demand for SwiftPM)
+cpp/bin/slice2swift.artifactbundle
```
externl: Pretty sure we already exclude all files in cpp/bin from `cpp/.gitignore`.

pepone: Yes, removed this.

PR #5011, opened 2026-02-04
Use TextEncoder.encodeInto for JavaScript string encoding (#5008)

### LINKED ISSUE #5008: Improve string encoding to use TextEncoder
https://github.com/zeroc-ice/ice/blob/82037ee96195981074c512a9d0d332b0447f1f59/js/src/Ice/Buffer.js#L148-L167

we can use TextEncoder.encodeInto to encode directly in the target buffer:

https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encodeInto

We need to patch the size after the string is written.

We can probably port this to 3.7.12 too.

pepone:
Replace the deprecated unescape(encodeURIComponent()) pattern with TextEncoder.encodeInto() to encode strings directly into the target buffer, avoiding intermediate allocations.

Key changes:
- Use TextEncoder.encodeInto() to write UTF-8 directly to buffer
- Pre-calculate max UTF-8 size (string.length * 3) to determine size encoding (1-byte vs 5-byte) upfront
- Optimize reserve() to only copy written data (up to _limit)

Fix #5008

### DIFF
--- a/js/src/Ice/Buffer.js
+++ b/js/src/Ice/Buffer.js
@@ -7,16 +7,6 @@ const indexOutOfBoundsExceptionMsg = "IndexOutOfBoundsException";
 // Singleton TextEncoder for UTF-8 string encoding
 const textEncoder = new TextEncoder();
 
-// Write size at position using 1-byte or 5-byte encoding
-function writeSizeAt(v, position, size, sizeLength) {
-    if (sizeLength === 1) {
-        v.setUint8(position, size);
-    } else {
-        v.setUint8(position, 255);
-        v.setInt32(position + 1, size, true);
-    }
-}
-
 export class Buffer {
     constructor(buffer) {
         if (buffer !== undefined) {
@@ -159,33 +149,11 @@ export class Buffer {
     }
 
     writeString(stream, v) {
-        if (v.length === 0) {
-            stream.writeSize(0);
-            return;
-        }
-
-        // JavaScript string.length returns the number of UTF-16 code units. The maximum UTF-8 bytes per
-        // UTF-16 code unit is 3, which occurs for BMP characters in the range U+0800–U+FFFF (e.g., CJK characters).
-        // Surrogate pairs (2 UTF-16 code units) encode to 4 UTF-8 bytes, which is only 2 bytes per code unit.
-        const maxUtf8Size = v.length * 3;
-
-        // Decide size encoding based on max possible encoded length. Slice 1.x size encoding allows using
-        // more bytes than strictly required, so it's valid to encode a size as 5 bytes even when 1 would suffice.
-        const sizeLength = maxUtf8Size <= 254 ? 1 : 5;
-
-        // Over-allocating buffer capacity avoids an extra allocation and copy that TextEncoder.encode() would require.
-        stream.expand(sizeLength + maxUtf8Size);
-
-        const sizePos = this._position;
-        this._position += sizeLength;
-
-        const targetView = new Uint8Array(this.b, this._position, maxUtf8Size);
-        const { written } = textEncoder.encodeInto(v, targetView);
-
-        writeSizeAt(this.v, sizePos, written, sizeLength);
-        this._position += written;
-
-        // Adjust limit to actual bytes written (we expanded for max, but may have written less)
+        const encoded = textEncoder.encode(v);
+        stream.writeSize(encoded.length);
+        this.expand(encoded.length);
+        new Uint8Array(this.b, this._position, encoded.length).set(encoded);
+        this._position += encoded.length;
         this._limit = this._position;
     }
 


js/src/Ice/Buffer.js
```diff
@@ -59,7 +72,7 @@ export class Buffer {
                 this.b = new ArrayBuffer(capacity);
             } else {
                 const b = new Uint8Array(capacity);
-                b.set(new Uint8Array(this.b));
+                b.set(new Uint8Array(this.b, 0, this._limit));
```
pepone, line 65: This avoid copy the unused part.

js/src/Ice/Buffer.js
```diff
@@ -93,7 +106,7 @@ export class Buffer {
             if (this._position + v.length > this._limit) {
                 throw new RangeError(bufferOverflowExceptionMsg);
             }
-            new Uint8Array(this.b, 0, this.b.byteLength).set(v, this._position);
+            new Uint8Array(this.b, this._position, v.byteLength).set(v);
```
pepone, line 99: This is a cosmetic change we create the buffer view at the right offset, and just call set without offset.

js/src/Ice/Buffer.js
```diff
@@ -146,24 +159,30 @@ export class Buffer {
     }
 
     writeString(stream, v) {
-        //
-        // Encode the string as utf8
-        //
-        const encoded = unescape(encodeURIComponent(v));
```
pepone, line 152: Getting rid of this deprecated API, avoid the copy, and removing the write loop is the core of this PR.

js/src/Ice/Buffer.js
```diff
@@ -146,24 +159,30 @@ export class Buffer {
...
-        if (this._position + sz > this._limit) {
-            throw new RangeError(bufferOverflowExceptionMsg);
-        }
-        for (let i = 0; i < sz; ++i) {
-            this.v.setUint8(this._position, v.charCodeAt(i));
-            this._position++;
-        }
+        // Decide size encoding based on max possible encoded length
+        const sizeLength = maxUtf8Size <= 254 ? 1 : 5;
+
```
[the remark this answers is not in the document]
pepone: I think the 4 extra bytes in 85 bytes would be negligible, code is simpler and avoids copy.

js/src/Ice/Buffer.js
```diff
@@ -146,24 +159,30 @@ export class Buffer {
...
+            stream.writeSize(0);
+            return;
+        }
 
-        stream.writeSize(encoded.length);
-        stream.expand(encoded.length);
-        this.putString(encoded, encoded.length);
-    }
+        // Maximum UTF-8 byte length: string.length * 3 (safe upper bound)
+        const maxUtf8Size = v.length * 3;
```
bernardnormier: This `* 3` requires more explanation / comment.

In C#, GetMaxByteCount is implemented roughly as `count * 3` because count represents UTF-16 "characters". Is it the same in JS? If not, it should be `* 4` since some Unicode characters are encoded on 4 UTF-8 bytes.

js/src/Ice/Buffer.js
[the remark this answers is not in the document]
bernardnormier: Yes, this is fine.

js/src/Ice/Buffer.js
```diff
@@ -146,24 +159,30 @@ export class Buffer {
...
-            throw new RangeError(bufferOverflowExceptionMsg);
-        }
-        for (let i = 0; i < sz; ++i) {
-            this.v.setUint8(this._position, v.charCodeAt(i));
-            this._position++;
-        }
+        // Decide size encoding based on max possible encoded length
+        const sizeLength = maxUtf8Size <= 254 ? 1 : 5;
+
+        stream.expand(sizeLength + maxUtf8Size);
```
bernardnormier: This pre-expansion is more debatable. We could over-allocate a lot for a long ASCII string.

The question is what is preferable:
 - preallocate too much
 - use an encode API that allocates the encoded string in a separate buffer that we then copy into the marshaling buffer

The Ice C# code uses the second option: no preallocation, no concern about rewriting the size / writing the size later, overall simpler logic.

The IceRPC C# code uses a better piecemeal writing that may not be available in JS.

js/src/Ice/Buffer.js
```diff
@@ -4,6 +4,19 @@ const bufferOverflowExceptionMsg = "BufferOverflowException";
...
+// Singleton TextEncoder for UTF-8 string encoding
+const textEncoder = new TextEncoder();
+
+// Write size at position using 1-byte or 5-byte encoding
+function writeSizeAt(v, position, size, sizeLength) {
+    if (sizeLength === 1) {
+        v.setUint8(position, size);
+    } else {
+        v.setUint8(position, 255);
+        v.setInt32(position + 1, size, true);
```
externl: what does true mean here?

js/src/Ice/Buffer.js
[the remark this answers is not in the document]
externl: Sounds fine but we should note this in a comment.

js/src/Ice/Buffer.js
pepone: Is the same in JavaScript https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length

I will update the comment.

js/src/Ice/Buffer.js
pepone: little-endian

js/src/Ice/Buffer.js
pepone: We can use `TextEncoder.encode` instead of `TextEncoder.encodeInto` and then copy the Uint8Array.set 

https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder/encode
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/set

But not sure, if you are writing many such string the net effect is the first call allocates extra memory and the subsequent call might avoid the allocation because it was previously allocated. Note that our expand strategy is to double the buffer already.

bernardnormier: I would rather use this `encode`, and not change too much the allocation behavior in a patch release.

pepone: Reworked to use `encode` which make the code much simpler.

js/src/Ice/Buffer.js
```diff
@@ -146,24 +149,12 @@ export class Buffer {
...
-        //
-        const encoded = unescape(encodeURIComponent(v));
-
+        const encoded = textEncoder.encode(v);
         stream.writeSize(encoded.length);
-        stream.expand(encoded.length);
-        this.putString(encoded, encoded.length);
-    }
-
-    putString(v, sz) {
```
bernardnormier, line 159: Is it ok to remove this function in a patch release?

js/src/Ice/Buffer.js
```diff
@@ -146,24 +149,12 @@ export class Buffer {
     }
 
     writeString(stream, v) {
```
bernardnormier, line 151: There is only one writeXxx in this file, writeString. All other encoding functions are putXxx.

Here v is the string (similar to the putXxx v parameter). If I follow this code, `stream` is the OutputStream that holds this buffer and we need stream because Buffer itself doesn't have a writeSize. Pretty odd API.

js/src/Ice/Buffer.js
pepone: The only caller was `writeString`, this is supposed to be an internal Buffer, users are not supposed to create/use Buffer instance directly.

js/src/Ice/Buffer.js
pepone: Yes it is, maybe we should add a write size here, and rename this as putString without stream parameter. Then have OutputStream call writeSize or maybe putSize?

Anyway that is for a follow-up PR.

PR #5020, opened 2026-02-05
Add module aggregation support to slice2js

### LINKED ISSUE #4914: slice2js does not merge nested modules correctly
### Version

- Ice version 3.8.0
- installed with `npm i @zeroc/ice`

### Describe the bug

When using a nested module (in this case, 2 deep) the modules are not correctly merged when imported in slice2js
Example:
`first.ice`:
```
module outer {
  module inner {

    interface First {
        int getVersion();
    };

  };
};
```
`second.ice`:
```
module outer {
  module inner {

    interface Second {
        bool isAlive();
    };

  };
};
```
then, `combined.ice`:
```
#include <first.ice>
#include <second.ice>

module outer {
  module inner {

    interface Combined {
        string getMessage();
    };

  };
};
```

The compiled output:
```javascript
// Copyright (c) ZeroC, Inc.

// slice2js version 3.8.0
// <auto-generated>Generated from Slice file 'combined.ice'.</auto-generated>

/* eslint-disable */
/* jshint ignore: start */

import { Ice } from "@zeroc/ice";

import { 
    outer as outer___first, } from "./first.js"
import { 
    outer as outer___second, } from "./second.js"

const outer = {
    ...outer___first,
    ...outer___second,
};
export { outer };
outer.inner = outer.inner || {};

const iceC_outer_inner_Combined_ids = [
    "::Ice::Object",
    "::outer::inner::Combined"
];

outer.inner.Combined = class extends Ice.Object {};

outer.inner.CombinedPrx = class extends Ice.ObjectPrx {};

Ice.TypeRegistry.declareProxyType("outer.inner.CombinedPrx", outer.inner.CombinedPrx);

Ice.defineOperations(
    outer.inner.Combined,
    outer.inner.CombinedPrx,
    iceC_outer_inner_Combined_ids,
    "::outer::inner::Combined", {
        "getMessage": [, 0, , [7], , , , , ]
    });
```

As you can see, the inner module definitions of `first.js` are overwritten when the spread operator is applied to combine the outer namespace:
```javascript
const outer = {
    ...outer___first,
    ...outer___second,
};
```
I verified this by importing the output and inspecting... the combined module in javascript was missing the `First` definitions.

### Expected behavior

A recursive merge has to be done on any module namespaces in imports, instead of just merging (shallowly) the outer module namespace. 

### Additional context

_No response_

pepone:
This PR improves slice2js to aggregate nested submodules, from its direct and transitive includes.

Added a new test Ice/moduleAggregation to test the new feature.

Fixes #4914

### DIFF
--- a/cpp/src/slice2js/Gen.cpp
+++ b/cpp/src/slice2js/Gen.cpp
@@ -1041,7 +1041,8 @@ Slice::Gen::ImportVisitor::writeImports(
         _out.inc();
         for (const auto& topLevelModule : topLevelModules)
         {
-            _out << nl << topLevelModule << " as " << importPathToIdentifier(jsImportedModule) << "_" << topLevelModule << ", ";
+            _out << nl << topLevelModule << " as " << importPathToIdentifier(jsImportedModule) << "_" << topLevelModule
+                 << ", ";
             if (isLocal)
             {
                 aggregatedModules.insert(topLevelModule);


pepone:
> It's not immediately clear why you deleted js/bin/slice2js as part of this PR.

Restored, was unintentionally

pepone:
> Seems good to me! Quite complex.

Module aggregation is inherently painful here because there is a real impedance mismatch between Slice modules and JavaScript modules. In Slice, you can have multiple top-level modules with arbitrarily nested sub-modules, while in JavaScript a module is fundamentally a file.

For this patch release, we also cannot change the overall structure too much. We were already aggregating top-level modules, and there was an existing TODO to extend this to sub-modules, which this change addresses.

Finally, part of the complexity is historical: when Ice for JavaScript was first released, JavaScript modules did not exist yet, and the original design predates the current ES module model.

cpp/src/slice2js/Gen.h
```diff
@@ -42,6 +42,68 @@ namespace Slice
...
+        [[nodiscard]] bool shouldVisitIncludedDefinitions() const final;
+        [[nodiscard]] const std::map<std::string, std::map<std::string, std::set<std::string>>>&
+        nestedModulesByTopLevel() const;
+        [[nodiscard]] const std::map<std::string, std::map<std::string, std::set<std::string>>>&
+        importedTypesByTopLevel() const;
+
+    private:
+        void addImportedType(const ContainedPtr& definition);
+
+        // Returns true if the definition context has js:module metadata.
```
InsertCreativityHere: I know these are private, but why not use `///` so the tooling can display hover text?

cpp/src/slice2js/Gen.cpp
```diff
@@ -185,6 +203,239 @@ namespace
...
+        // Transitive include - use the owning direct include's key.
+        _importedTypesByTopLevel[_topLevelFile][*includeKey].insert(typeScope);
+    }
+}
+
+bool
+Slice::IncludeAggregationVisitor::hasJsModuleMetadata(const DefinitionContextPtr& dc) const
+{
+    // Used by TypeScript generation - files with js:module don't need TypeScript imports.
+    return !getJavaScriptModule(dc).empty();
```
InsertCreativityHere: This works but feels very indirect. I think it would be better to 'inline' this function, and replace it with the already existing:
```suggestion
    dc->hasMetadata("js:module");
```

cpp/src/slice2js/Gen.h
```diff
@@ -42,6 +42,68 @@ namespace Slice
...
+        // Used by TypeScript generation - files with js:module don't need TypeScript imports.
+        [[nodiscard]] bool hasJsModuleMetadata(const DefinitionContextPtr& dc) const;
```
InsertCreativityHere: The compiler already has a `hasMetadata` that could be used instead I feel.
Also, this function is only called from 1 place anyways. Probably best to inline.

cpp/src/slice2js/Gen.cpp
```diff
@@ -185,6 +203,239 @@ namespace
...
+    }
+    return std::nullopt;
+}
+
+bool
+Slice::IncludeAggregationVisitor::visitModuleStart(const ModulePtr&)
+{
+    return true;
+}
+
```
InsertCreativityHere: This is unnecessary. All the `visitStart` methods are default implemented with `return true;`.
Can just delete function this from the source and header.

cpp/src/slice2js/Gen.cpp
```diff
@@ -185,6 +203,239 @@ namespace
...
+    return _nestedModulesByTopLevel;
+}
+
+const std::map<std::string, std::map<std::string, std::set<std::string>>>&
+Slice::IncludeAggregationVisitor::importedTypesByTopLevel() const
+{
+    return _importedTypesByTopLevel;
+}
+
+[[nodiscard]] bool
```
InsertCreativityHere: Why have `[[nodiscard]]` here? We already have it in the header file.
Usually we don't put `[[nodiscard]]`s in the source.
```suggestion
bool
```

cpp/src/slice2js/Gen.cpp
```diff
@@ -543,11 +797,20 @@ Slice::Gen::generate(const UnitPtr& p)
...
 
-        TypeScriptImportVisitor importVisitor(_typeScriptOutput);
+        IncludeAggregationVisitor moduleVisitor;
+        p->visit(&moduleVisitor);
+
+        // Get the imported types for the top-level file.
+        const auto& importedTypesByTopLevel = moduleVisitor.importedTypesByTopLevel();
+        auto it = importedTypesByTopLevel.find(p->topLevelFile());
+        map<string, set<string>> importedTypesByInclude =
+            it != importedTypesByTopLevel.end() ? it->second : map<string, set<string>>{};
```
InsertCreativityHere: This is the only place we call `importedTypesByTopLevel` from, why not just move this extra `find` logic into it instead of having it in (the already pretty large and important) `generate` function?

Also not clear what it's doing. It's called `importedTypesByTopLevel`, but then it gives us extra stuff and we have to search for the 'topLevel' anyways?

cpp/src/slice2js/Gen.cpp
```diff
@@ -185,6 +203,239 @@ namespace
...
+{
+    addImportedType(p);
+    return true;
+}
+
+void
+Slice::IncludeAggregationVisitor::visitOperation(const OperationPtr&)
+{
+    // Nothing to do
+}
```
InsertCreativityHere: I guess there's nothing wrong with this, but usually we don't implement `no-op` methods just for the sake of it.
This is already default implemented as `no-op`, so there's no need to implement it here, except for "documentation" I guess?

js/package.json
```diff
@@ -34,6 +34,7 @@
     },
     "browserslist": "> 0.25%, not dead",
     "devDependencies": {
+        "test_nested_modules": "file:test/Ice/moduleAggregation/module",
```
InsertCreativityHere, line 37: Why do we need this in the top-level `js` package?
Why don't other tests need to be here?

pepone: We use this to test "js:module" which is used to import the generated code from an external pacakge.

PR #5024, opened 2026-02-06
Cleanup JavaScript Buffer API

### LINKED ISSUE #5018: Improve JavaScript buffer API
There is only one writeXxx in this file, writeString. All other encoding functions are putXxx.

Here v is the string (similar to the putXxx v parameter). If I follow this code, `stream` is the OutputStream that holds this buffer and we need stream because Buffer itself doesn't have a writeSize. Pretty odd API.

_Originally posted by @bernardnormier in https://github.com/zeroc-ice/ice/pull/5011#discussion_r2765960091_

pepone:
Fix #5018

### DIFF
--- a/js/src/Ice/Buffer.js
+++ b/js/src/Ice/Buffer.js
@@ -4,9 +4,6 @@ const bufferOverflowExceptionMsg = "BufferOverflowException";
 const bufferUnderflowExceptionMsg = "BufferUnderflowException";
 const indexOutOfBoundsExceptionMsg = "IndexOutOfBoundsException";
 
-// Singleton TextEncoder for UTF-8 string encoding
-const textEncoder = new TextEncoder();
-
 export class Buffer {
     constructor(buffer) {
         if (buffer !== undefined) {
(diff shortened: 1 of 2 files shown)


externl (approved):

js/src/Ice/Buffer.js
```diff
@@ -148,15 +148,26 @@ export class Buffer {
...
         const encoded = textEncoder.encode(v);
-        stream.writeSize(encoded.length);
+        this.putSize(encoded.length);
         this.expand(encoded.length);
         new Uint8Array(this.b, this._position, encoded.length).set(encoded);
         this._position += encoded.length;
         this._limit = this._position;
     }
 
+    putSize(v) {
```
bernardnormier: The new putString and putSize are inconsistent with the existing writeXxx and putXxx.

In the existing code:
 - OutputStream.writeXxx expands then calls Buffer.putXxx
 - Buffer.putXxx NEVER expands and just write the bytes, assuming the buffer is large enough

We should keep this pattern.

pepone: Then , what about moving `const encoded = textEncoder.encode(v);` to writeString, otherwise it doesn't know how much to expand the buffer.

Then we remove putString, and just call putArray with the encoded data. We call `writeSize` from OuputStream.writeString and no need for `putSize`.

PR #5026, opened 2026-02-06
Add IceBT library to RPM distributions

### LINKED ISSUE #4749: Add IceBT library to RPM distributions
The IceBT library is only built with Debian and Ubuntu distributions, we should add this to RPM distributions too.

pepone:
## Summary
- Add IceBT (Bluetooth transport) library to the RPM packages for el9 and el10 distributions
- Add conditional `BuildRequires` for `bluez` and `dbus-1` (excluded on amzn2023)
- Add `libIceBT.so` to `libice3.9-c++` runtime and `libice-c++-devel` packages
- No Makefile or Dockerfile changes needed — the build system already conditionally compiles IceBT when bluez/dbus are detected

Closes #4749

RPM build with the changes https://github.com/zeroc-ice/ice/actions/runs/21822246350

### DIFF
--- a/packaging/rpm/ice.spec
+++ b/packaging/rpm/ice.spec
@@ -60,12 +60,8 @@ BuildRequires: pkgconfig(openssl)
 BuildRequires: pkgconfig(bzip2)
 BuildRequires: pkgconfig(lmdb)
 BuildRequires: pkgconfig(libsystemd)
-
-# Bluetooth transport is only built for RHEL distributions
-%if "%{dist}" != ".amzn2023"
 BuildRequires: pkgconfig(bluez)
 BuildRequires: pkgconfig(dbus-1)
-%endif
 
 # Amazon Linux 2023 does not provide pkgconfig(mcpp)
 %if "%{dist}" == ".amzn2023"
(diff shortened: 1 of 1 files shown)


packaging/rpm/ice.spec
```diff
@@ -61,6 +61,12 @@ BuildRequires: pkgconfig(bzip2)
 BuildRequires: pkgconfig(lmdb)
 BuildRequires: pkgconfig(libsystemd)
 
+# Bluetooth transport is only built for RHEL distributions
+%if "%{dist}" != ".amzn2023"
+BuildRequires: pkgconfig(bluez)
```
externl, line 63: Is bluez not available on amzn2023?

pepone: I didn't check, I was under the impression that there is no point into adding this to a cloud based distribution. BT is about local point communications.

packaging/rpm/ice.spec
```diff
@@ -419,6 +425,9 @@ cp -p java/lib/icegridgui.jar %{buildroot}%{_javadir}/icegridgui.jar
 %{_libdir}/libDataStorm.so.*
 %{_libdir}/libGlacier2.so.*
 %{_libdir}/libIce.so.*
+%if "%{dist}" != ".amzn2023"
```
bernardnormier: Could we do something cleaner than all these `%if "%{dist}" != ".amzn2023"`?

Imagine when we want to support another distribution without BT.

packaging/rpm/ice.spec
externl: Filtering it out “just because” feels like more effort than it’s worth. Amazon Linux 2023 ships bluez packages and is explicitly intended to be usable as a bare-metal distribution, not just a cloud image.

Filtering this out seems artificial.

pepone: Update the file to not exclude amzn2023

PR #5029, opened 2026-02-06
Fix slice2js generating invalid JS identifiers from hyphenated module names

### LINKED ISSUE #5019: slice2js generates invalid TypeScript identifiers for hyphenated module names
## Summary

When a Slice file uses `js:module` metadata with a module name containing hyphens (e.g., `test-nested-modules`), the generated TypeScript declaration file (`.d.ts`) contains invalid identifiers.

## Steps to Reproduce

1. Create a Slice file with hyphenated module name:
   ```slice
   [["js:module:test-nested-modules"]]

   module Outer {
       struct First {
           int value;
       }
   }
   ```

2. Create another Slice file that includes the first without `js:module`:
   ```slice
   #include "First.ice"

   module External {
       struct UsesFirst {
           Outer::First first;
       }
   }
   ```

3. Run `slice2js --typescript` on both files.

## Expected Behavior

The generated `.d.ts` file should have valid TypeScript identifiers, e.g.:
```typescript
import * as __module_test_nested_modules from "test-nested-modules";
```

## Actual Behavior

The generated `.d.ts` file has invalid identifiers with hyphens:
```typescript
import * as __module_test-nested-modules from "test-nested-modules";
```

This causes TypeScript compilation errors:
```
error TS1005: 'from' expected.
error TS1005: ';' expected.
```

## Impact

- TypeScript compilation fails for any code that includes Slice files from external modules with hyphenated names
- The JavaScript code is generated correctly (uses string import specifier)
- Only the TypeScript declarations are affected

## Suggested Fix

In the slice2js code generator, when creating the import alias identifier (e.g., `__module_<name>`), sanitize the module name by replacing hyphens with underscores:

```
"test-nested-modules" → "__module_test_nested_modules"
```

The string import specifier should remain unchanged (`"test-nested-modules"`), only the variable/identifier needs sanitization.

## Workaround

Use module names with underscores instead of hyphens:
- `test_nested_modules` instead of `test-nested-modules`

## Files Involved

- `cpp/src/slice2js/Gen.cpp` - TypeScript declaration generation
- `cpp/src/slice2js/Gen.h`

pepone:
## Summary
- Fixes #5019 — `pathToModule()` in `slice2js` generated invalid JavaScript/TypeScript identifiers when module names or file paths contained hyphens (e.g., `__module_test-nested-modules`).
- The function had two bugs: character replacements were conditional on path separators being present, and hyphens were never replaced even when the conditional was entered.
- Replaced the piecemeal character replacements with a single loop that substitutes any non-identifier character (`[^a-zA-Z0-9_$]`) with `_`.
- Added a `Slice/moduleName` test that includes a hyphenated `.ice` filename to exercise the fix.

### DIFF
--- a/cpp/src/slice2js/Gen.cpp
+++ b/cpp/src/slice2js/Gen.cpp
@@ -12,6 +12,7 @@
 #include <cassert>
 #include <iostream>
 #include <iterator>
+#include <regex>
 
 using namespace std;
 using namespace Slice;
@@ -25,9 +26,10 @@ namespace
     // e.g., "../foo/bar/baz.ice" -> "__foo_bar_baz", "@zeroc/ice" -> "_zeroc_ice"
     string importPathToIdentifier(const string& path)
     {
-        // Only strip known file extensions; a bare module name like "my.module" must not be truncated.
+        // Strip common extensions from the end of the path, so that the generated identifier is cleaner.
+        // If the code imports ./Foo/Foo.ice  we want the generated identifier to be _Foo_Foo, not _Foo_Foo_ice.
         string identifier = path;
-        static constexpr string_view extensions[] = {".d.ts", ".ice", ".js"};
+        static constexpr string_view extensions[] = {".ice", ".js"};
         for (string_view ext : extensions)
         {
             if (identifier.size() > ext.size() &&
@@ -39,15 +41,8 @@ namespace
         }
 
         // Replace any character that is not valid in a JavaScript identifier with '_'.
-        for (char& c : identifier)
-        {
-            if (!((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '$'))
-            {
-                c = '_';
-            }
-        }
-
-        return identifier;
+        static const regex disallowedChars("[^a-zA-Z0-9_$]");
+        return regex_replace(identifier, disallowedChars, "_");
     }
 
     bool ends_with(string_view s, string_view suffix)
@@ -794,7 +789,8 @@ Slice::Gen::ImportVisitor::writeImports(const UnitPtr& p)
         {
             if (modules.find("Ice") != modules.end())
             {
-                _out << nl << "import { Ice as Ice_" << importPathToIdentifier(imported) << " } from \"" << imported << "\"";
+                _out << nl << "import { Ice as Ice_" << importPathToIdentifier(imported) << " } from \"" << imported
+                     << "\"";
             }
         }
 
@@ -853,8 +849,8 @@ Slice::Gen::ImportVisitor::writeImports(const UnitPtr& p)
             _out.inc();
             for (const auto& topLevelModule : topLevelModules)
             {
-                _out << nl << topLevelModule << " as " << topLevelModule << "_" << importPathToIdentifier(jsImportedModule)
-                     << ", ";
+                _out << nl << topLevelModule << " as " << topLevelModule << "_"
+                     << importPathToIdentifier(jsImportedModule) << ", ";
                 aggregatedModules.insert(topLevelModule);
             }
             _out.dec();
@@ -1915,7 +1911,8 @@ Slice::Gen::TypeScriptImportVisitor::writeImports()
     _out << sp;
     for (const auto& moduleName : _importedModules)
     {
-        _out << nl << "import * as __module_" << importPathToIdentifier(moduleName) << " from \"" << moduleName << "\";";
+        _out << nl << "import * as __module_" << importPathToIdentifier(moduleName) << " from \"" << moduleName
+             << "\";";
     }
     return _importedTypes;
 }


cpp/src/slice2js/Gen.cpp
```diff
@@ -21,25 +21,18 @@ using namespace IceInternal;
 
 namespace
 {
-    // Convert a path to a module name, e.g., "../foo/bar/baz.ice" -> "__foo_bar_baz"
+    // Convert a path or module name to a valid JavaScript identifier, e.g., "../foo/bar/baz.ice" -> "__foo_bar_baz"
     string pathToModule(const string& path)
```
externl: The function names doesn't seem to match the description any longer.

cpp/src/slice2js/Gen.cpp
```diff
@@ -21,25 +21,18 @@ using namespace IceInternal;
...
     {
         string moduleName = removeExtension(path);
 
-        size_t pos = moduleName.find('/');
```
externl, line 29: was all of this wrong?

cpp/src/slice2js/Gen.cpp
```diff
@@ -21,25 +21,18 @@ using namespace IceInternal;
...
-
-        if (pos != string::npos)
-        {
-            // Replace remaining path separators ('/', '\') and ('.') with '_'
-            replace(moduleName.begin(), moduleName.end(), '/', '_');
-            replace(moduleName.begin(), moduleName.end(), '\\', '_');
-            replace(moduleName.begin(), moduleName.end(), '.', '_');
-            // Replace @ in scoped moduleName names with _
-            replace(moduleName.begin(), moduleName.end(), '@', '_');
+            if (!isalnum(static_cast<unsigned char>(c)) && c != '_' && c != '$')
```
externl: when do we encounter `$`?

pepone: `$` is a valid character in both file/module names, and in JavaScript identifiers, we don't escape it.

InsertCreativityHere: Minor, but I think this is easier to understand:
```suggestion
            if (!(isalnum(static_cast<unsigned char>(c)) || c == '_' || c == '$'))
```
Guaranteed to be equivalent by De Morgan's Laws

js/test/Slice/moduleName/Client.ts
```diff
@@ -0,0 +1,19 @@
...
+import { Test } from "./test-dep.js";
+import { MyTest } from "./Test.js";
+
+export class Client extends TestHelper {
+    async run() {
+        const out = this.getWriter();
+
+        out.write("testing module names with special characters... ");
+        const foo = new Test.Foo(42);
+        test(foo.x === 42);
```
InsertCreativityHere, line 13: Do these `test` statements actually matter here?
Isn't just constructing the object enough of a test?

Or (and I'm not a JS expert by any means) if there was a broken identifier, would just importing the file be enough to catch it?

js/test/Slice/moduleName/Test.ice
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) ZeroC, Inc.
```
InsertCreativityHere, line 1: It looks like the new code was designed to handle `js:module` metadata too right? (Because we allow '$')
Should at least one of these tests also use the metadata?

Or I missing what this change is accomplishing.

js/test/Slice/moduleName/Client.ts
pepone: Not 100% sure, TypeScript might be tempted to optimize it away.

js/test/Slice/moduleName/Test.ice
pepone: When we import a symbol in JavaScript , we can import from a Node.js package (js:module), the metadata name is confusing. Or from a JavaScript file which is by definition a module.

And we generate something like:

```
import { 
    Test as Test___test_dep, } from "./test-dep.js"
```

This apply if you are importing from "test-deep" because the Slice file contains `["js:module:test-deep"]` or importing from `./test-deep.js` because the file is named `test-deep.ice` and doesn't have `js:module` metadata.

Previously the identifier would contain `-` or other non valid characters.

cpp/src/slice2js/Gen.cpp
pepone: I just replaced this part with a similar check but also using ASCII range.

cpp/src/slice2js/Gen.cpp
pepone: renamed `importPathToIdentifier`, the name didn't reflect what the previous implementation did either.

cpp/src/slice2js/Gen.cpp
pepone: It was incomplete, for example we only handle `/`, `\\`, and `@` which are valid in paths/package-names, but not in JavaScript identifiers. For other special character we just produced invalid identifiers. For example `-` in a Slice file name would result in generate code that cannot be imported in other generated files.

js/test/Slice/moduleName/Test.ice
pepone: So to bring the argument home, the important part is not whether the argument is a package path, or a Slice include path. The important part is that we return a valid JavaScript identifier.

cpp/src/slice2js/Gen.cpp
```diff
@@ -21,28 +21,33 @@ using namespace IceInternal;
...
-            // Replace remaining path separators ('/', '\') and ('.') with '_'
-            replace(moduleName.begin(), moduleName.end(), '/', '_');
-            replace(moduleName.begin(), moduleName.end(), '\\', '_');
-            replace(moduleName.begin(), moduleName.end(), '.', '_');
-            // Replace @ in scoped moduleName names with _
-            replace(moduleName.begin(), moduleName.end(), '@', '_');
+            if (!((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') || c == '_' || c == '$'))
+            {
+                c = '_';
+            }
```
InsertCreativityHere: At this point, could we just use:
```
std::regex disallowedChars("^[a-zA-Z0-9_$]");
goodIdentifier = std::regex_replace(identifier, disallowedChars, "_");
```
Regex are definitely a little heavy for this, but it seems simpler than what we're doing, if we want to avoid the locale issue of `isalnum`.

js/test/Slice/moduleName/Test.ice
InsertCreativityHere: Okay, but this would only be visible in the generated code then, I think I see.
My confusion was that the code was all about converting 'illegal' chars to underscores, but I didn't see a single underscore anywhere in the test!

cpp/src/slice2js/Gen.cpp
```diff
@@ -21,28 +21,33 @@ using namespace IceInternal;
...
+    // Create a valid JavaScript identifier from an import path, used as a prefix for imported symbols.
+    // e.g., "../foo/bar/baz.ice" -> "__foo_bar_baz", "@zeroc/ice" -> "_zeroc_ice"
+    string importPathToIdentifier(const string& path)
     {
-        string moduleName = removeExtension(path);
-
-        size_t pos = moduleName.find('/');
-        if (pos == string::npos)
+        // Only strip known file extensions; a bare module name like "my.module" must not be truncated.
+        string identifier = path;
```
externl: I we just have to assume no one is using `.ice` as a module name like this?

cpp/src/slice2js/Gen.cpp
```diff
@@ -21,28 +21,33 @@ using namespace IceInternal;
...
+        static constexpr string_view extensions[] = {".d.ts", ".ice", ".js"};
+        for (string_view ext : extensions)
```
externl, line 33: This is probably have a comment as to what it's doing. Or maybe Move the above comment down a few lines.

cpp/src/slice2js/Gen.cpp
```diff
@@ -21,28 +21,33 @@ using namespace IceInternal;
...
+    // e.g., "../foo/bar/baz.ice" -> "__foo_bar_baz", "@zeroc/ice" -> "_zeroc_ice"
+    string importPathToIdentifier(const string& path)
     {
-        string moduleName = removeExtension(path);
-
-        size_t pos = moduleName.find('/');
-        if (pos == string::npos)
+        // Only strip known file extensions; a bare module name like "my.module" must not be truncated.
+        string identifier = path;
+        static constexpr string_view extensions[] = {".d.ts", ".ice", ".js"};
```
externl: Are these user files? Are `mis` and `cjs` extensions we need to worry about?

externl (approved):

js/test/Slice/moduleName/Test.ice
pepone: Right, this are local names in the generated file. The local name you give to the imported thing.

cpp/src/slice2js/Gen.cpp
pepone: We can probably distiguish the two cases:

```
["js:module:foo.ice"]
```

and

```
include "Foo.ice"
```

I will create a separate PR for this issue, I want to merge #5020 before doing more changes to slice2js.

cpp/src/slice2js/Gen.cpp
pepone: `.d.ts` is actually not required, this match Slice includes or import for generated files so either `.ice` or `.js`

PR #5030, opened 2026-02-06
Build & ship IceBox client library in Swift, MATLAB, and JavaScript

### LINKED ISSUE #4828: Build & ship IceBox client library in Swift, MATLAB
We currently build & ship the IceBox client library in most languages, but not MATLAB and Swift. This is inconsistent.

pepone:
## Summary
- Add IceBox client library to **Swift**: new `IceBox` product/target in `Package.swift`, source files in `swift/src/IceBox/`, and test dependency in `swift/test/Package.swift`
- Add IceBox client library to **MATLAB**: `make-matlab-package` rule in `Makefile`, MSBuild `SliceCompile` entry in `matlab/lib/msbuild/ice.proj`, and toolbox slice copy
- Add IceBox client library to **JavaScript**: `IceBox` in gulpfile `libs` array, new `js/src/IceBox/` module, and exports in `js/src/index.js` / `index.d.ts`

The IceBox client library contains only code generated from the `slice/IceBox/ServiceManager.ice` Slice definitions, consistent with how it's done in Java, Python, and C#.

Fixes #4828

## Test plan
- [ ] Verify Swift `IceBox` target builds via `swift build`
- [ ] Verify MATLAB IceBox package generates via `make` (Linux) and MSBuild (Windows)
- [ ] Verify JavaScript IceBox module generates via `gulp dist`

### DIFF
--- a/js/src/IceBox/.npmignore
+++ b/js/src/IceBox/.npmignore
@@ -1,1 +1,1 @@
-# Ensure we include generated files ignored by the .gitignore file
+# Override .gitignore for npm publish, ensure the generated code is included in the package
--- a/js/src/Glacier2/.npmignore
+++ b/js/src/Glacier2/.npmignore
@@ -1,1 +1,1 @@
-# Ensure we include generated files ignored by the .gitignore file
+# Override .gitignore for npm publish, ensure the generated code is included in the package
--- a/js/src/Ice/.npmignore
+++ b/js/src/Ice/.npmignore
@@ -1,1 +1,1 @@
-# Ensure we include generated files ignored by the .gitignore file
+# Override .gitignore for npm publish, ensure the generated code is included in the package
--- a/js/src/IceGrid/.npmignore
+++ b/js/src/IceGrid/.npmignore
@@ -1,1 +1,1 @@
-# Ensure we include generated files ignored by the .gitignore file
+# Override .gitignore for npm publish, ensure the generated code is included in the package
--- a/js/src/IceStorm/.npmignore
+++ b/js/src/IceStorm/.npmignore
@@ -1,1 +1,1 @@
-# Ensure we include generated files ignored by the .gitignore file
+# Override .gitignore for npm publish, ensure the generated code is included in the package


js/src/IceBox/.npmignore
```diff
@@ -0,0 +1 @@
+# Ensure we include generated files ignored by the .gitignore file
```
externl: Do the above generated files need to be here as per the comment?

pepone: We use this pattern for other modules

See https://github.com/zeroc-ice/ice/blob/main/js/src/IceGrid/.npmignore

externl: I'm not following the logic though. What's the benefit of an empty `.npmignore` file? 

Don't we want to be "including" the generated files ignored by `.gitignore`. I.e. have a `!Name.js` entry in these files?

pepone: https://docs.npmjs.com/cli/v9/using-npm/developers#keeping-files-out-of-your-package

By adding an empty .npmignore we are ignoring the .gitignore rules. Don't need to have `!Name.js`

swift/src/IceBox/IceBox.docc/IceBox.md
```diff
@@ -0,0 +1,3 @@
+# ``IceBox``
+
+IceBox client library. Host multiple independent services in the same Ice server.
```
bernardnormier: I find this rather confusing. Obviously the client library doesn't host anything.

What about something like:
The IceBox client library allows you to administer an IceBox server remotely.

We should also review & fix the Swift Glacier2.md (etc.) in a follow-up PR.

js/src/IceBox/.npmignore
externl: Very confusing documentation.

> If you want to include something that is excluded by your .gitignore file, you can create an empty .npmignore file to override it.

So by "something", they mean everything?

pepone: I updated the comments and create a separate issue to stop using the `.npmignore` files

See https://github.com/zeroc-ice/ice/issues/5036

PR #5031, opened 2026-02-06
Create @zeroc/slice2js package and add unplugin integration tests

### LINKED ISSUE #4728: Replace Gulp build system
Gulp (tasked based runners in general) is no longer cool. We should look into replacing it with whatever the latest NPM tooling is. 

Replacements ideas:

- Webpack
- Rollup
- Parcel
- ... (please update this list)

We should also really make sure that we have a slice compiler plugin for these popular bundlers.

pepone:
This PR creates a new package `@zeroc/slice2js` the package contains:

- slice2js binaries for all supported platforms
- unplugin plugin that is compatible with vite/rollup/webpack/esbuild (maybe other but not tested)
- Adds a test for the mentioned bundleres/build systems.

We keep rollup as the task coordinator for our packages, but it uses the new plug-in to run the Slice compiler.

For now I kept slice2js in the main package for compatibility with 3.8.0, but maybe we should drop it. Thoughts?

The reason to add a separate package for the plugin is to avoid to have additional dependencies in `@zeroc/ice` package, the main package is kept with 0 runtime dependencies, we only have dev dependencies for the build and the tests.

Fixes #4728

### DIFF
--- a/js/packages/slice2js/scripts/prepack.js
+++ b/js/packages/slice2js/scripts/prepack.js
@@ -1,5 +1,9 @@
 // Copyright (c) ZeroC, Inc.
 
+// Prepack script for @zeroc/slice2js. Bundles the Slice definitions and platform-specific slice2js compiler
+// binaries into the package directory before packing. When SLICE2JS_STAGING_PATH is set, all supported
+// platform binaries are copied from the staging directory; otherwise, the locally built binary is used.
+
 import fs from "fs";
 import os from "os";
 import path from "path";
--- a/js/packages/ice/scripts/prepack.js
+++ b/js/packages/ice/scripts/prepack.js
@@ -1,5 +1,10 @@
 // Copyright (c) ZeroC, Inc.
 
+// Prepack script for @zeroc/ice. Bundles the README, Slice definitions, and platform-specific slice2js
+// compiler binaries into the package directory before packing. When SLICE2JS_STAGING_PATH is set, all
+// supported platform binaries are copied from the staging directory; otherwise, the locally built binary
+// is used.
+
 import fs from "fs";
 import os from "os";
 import path from "path";


js/packages/slice2js/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+const exitCode = await compile(["--typescript", "--output-dir", "src/generated", "slice/Hello.ice"]);
+```
+
+## CLI
+
+The package provides the `slice2js` command:
+
+```bash
+npx slice2js --typescript --output-dir src/generated slice/Hello.ice
```
pepone: Moving this here is a breaking change, is the more logical place but we cannot have this in both as it would cause a conflict.

js/packages/slice2js/bin/slice2js
```diff
@@ -7,7 +7,7 @@ try {
     // - argv[0]: The JavaScript interpreter (e.g., npx or node)
     // - argv[1]: The script's file path
     // - Remaining arguments are passed to slice2js via the compile method
-    const exitCode = await compile(process.argv.slice(2), { stdio: "inherit" });
```
pepone, line 10: inherit stdin was causing a hang when running with unplugin, and we are not reading it.

externl: You mean `stdio`?

js/packages/slice2js/README.md
externl: I would be inclined to just not expose `slice2js` as a binary for the 3.8.1 release for this package. When we release 3.9 we can move it back in.

js/packages/slice2js/scripts/prepack.js
```diff
@@ -0,0 +1,106 @@
...
+            copyFolderSync(srcFile, destFile);
+        } else {
+            fs.copyFileSync(srcFile, destFile);
+        }
+    });
+}
+
+// Ensure we are running from Ice source distribution
+if (!fs.existsSync(sliceSourceDir)) {
+    console.error("The pack command must be run from the Ice source distribution.");
```
externl: Shouldn't we just not ship this file as part log any NPM package?

js/packages/slice2js/src/unplugin-slice2js.ts
```diff
@@ -0,0 +1,82 @@
...
+ * @param {string[] | string} [options.args] - Additional slice2js CLI arguments.
+ * @param {string} [options.cwd] - Base directory for resolving inputs.
+ * @param {string} [options.toolsPath] - Directory containing the slice2js binary (for source builds).
+ * @param {string} [options.slicePath] - Directory containing Slice definitions (for source builds).
+ */
+export async function runSlice2js(options) {
+    const cwd = options.cwd ? path.resolve(options.cwd) : process.cwd();
+    const outputDir = path.resolve(cwd, options.outputDir);
+    const inputs = fg.sync(toArray(options.inputs), { cwd, absolute: true });
+
```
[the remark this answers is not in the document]
externl: a legit concern?

js/packages/slice2js/src/unplugin-slice2js.ts
externl, line 1: Should / can we write this new code in TypeScript?

js/packages/slice2js/package.json
```diff
@@ -0,0 +1,46 @@
+{
+    "name": "@zeroc/slice2js",
+    "version": "3.9.0-alpha0",
```
externl: IIRC we should be using `.0` so that it's correctly semver comparable against `.1`, etc. 
```suggestion
    "version": "3.9.0-alpha.0",
```

js/packages/test/Slice/unplugin/vite/vite.test.ts
externl, line 1: Can these test files be in typescript?

js/package.json
```diff
@@ -16,6 +16,9 @@
         "Ice",
         "rpc"
     ],
+    "workspaces": [
```
externl, line 7: do we need list "ice" as a workspace?

externl (commented):
First pass! So far so good!

js/packages/slice2js/bin/slice2js
pepone: setting this is equivalent to set inherit to all three stdin/stdout/stderr.

For what I tested stdin was causing the hang 

See https://nodejs.org/api/child_process.html#optionsstdio

js/packages/slice2js/src/unplugin-slice2js.ts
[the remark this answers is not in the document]
pepone: Need to review

js/packages/slice2js/scripts/prepack.js
pepone: Yes that seems the most reasonable.

js/packages/slice2js/src/unplugin-slice2js.ts
pepone: most probably yes,  I will check.

js/package.json
pepone: I think we might want to restructure things a bit to get a proper workspace, with 3 packages. `@zeroc/slice2js`, `@zeroc/ice`, `tests`. this way we won't have ice as a workspace, and packages can reference each other through the workspace.

Currently the way tests load ice is a bit of a hack.

js/packages/slice2js/README.md
bernardnormier: "Moving this here is a breaking change"

Can you explain what you mean?

bernardnormier (approved):

js/packages/slice2js/README.md
pepone: For example our 3.8 demos expect this to work:

```
jose@mac test-js % npm install @zeroc/ice --save

added 1 package, and audited 2 packages in 1s

found 0 vulnerabilities
jose@mac test-js % npx slice2js --version
3.8.0
jose@mac test-js % 
```

If we move it to the new plugin package, and you update to `3.8.1` this stop working until you install the new `@zeroc/slice2js` package.

js/packages/slice2js/scripts/prepack.js
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) ZeroC, Inc.
+
```
externl, line 2: Would we add a comment here explaining what this script does?

js/packages/slice2js/src/slice2js.ts
```diff
@@ -0,0 +1,78 @@
...
+        archPrefix = "arm64";
+        exe = "slice2js";
+        break;
+    case "linux":
+        platformPrefix = "linux";
+        exe = "slice2js";
+        break;
+    default:
+        console.error(`Unsupported platform: ${os.platform()}`);
+        process.exit(1);
```
externl: Throwing does seem better than exiting the process. If you end up calling this as part of a build script, like web pack or something, I think this would kill everything.  Or maybe this is the expected way to fail on import?

js/packages/slice2js/src/unplugin-slice2js.ts
```diff
@@ -0,0 +1,87 @@
...
+    return Array.isArray(value) ? value : [value];
+}
+
+export interface Slice2jsOptions extends CompileOptions {
+    /** Slice files or glob patterns (e.g. "*.ice", "**\/*.ice"). */
+    inputs: string | string[];
+    /** Output directory for generated files. */
+    outputDir: string;
+    /** Additional include directories. */
+    include?: string | string[];
```
externl, line 22: Why not just accept only a list?

js/packages/slice2js/src/unplugin-slice2js.ts
```diff
@@ -0,0 +1,87 @@
...
+function toArray(value: string | string[] | undefined): string[] {
+    if (!value) {
+        return [];
+    }
+    return Array.isArray(value) ? value : [value];
+}
+
+export interface Slice2jsOptions extends CompileOptions {
+    /** Slice files or glob patterns (e.g. "*.ice", "**\/*.ice"). */
+    inputs: string | string[];
```
externl, line 18: Why not just accept only a list?

js/packages/slice2js/src/unplugin-slice2js.ts
```diff
@@ -0,0 +1,87 @@
...
+    /** Output directory for generated files. */
+    outputDir: string;
+    /** Additional include directories. */
+    include?: string | string[];
+    /** Additional slice2js CLI arguments. */
+    args?: string | string[];
```
externl, line 24: Ditto.

js/packages/slice2js/README.md
```diff
@@ -0,0 +1,131 @@
+# @zeroc/slice2js
+
+The Slice-to-JavaScript compiler and build plugin for [Ice].
+
+This package provides:
+
+- The `slice2js` compiler, which compiles Slice definitions (`.ice` files) into JavaScript and TypeScript
+- Build plugins for [Vite], [Rollup], [webpack], and [esbuild] via [unplugin]
+- A programmatic API for running the compiler
```
externl:
```suggestion
- The `slice2js` compiler, which compiles Slice definitions (`.ice` files) into JavaScript and TypeScript.
- Build plugins for [Vite], [Rollup], [webpack], and [esbuild] via [unplugin].
- A programmatic API for running the compiler.
```

js/packages/slice2js/README.md
```diff
@@ -0,0 +1,131 @@
...
+## Installation
+
+```bash
+npm install --save-dev @zeroc/slice2js
+```
+
+## Build Plugin
+
+The build plugin compiles `.ice` files automatically during your build. It supports glob patterns and watches `.ice`
+files for changes in development mode.
```
externl, line 20: What is this development mode?

js/packages/slice2js/README.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+### Plugin Options
+
+| Option      | Type                 | Description                                                      |
+|-------------|----------------------|------------------------------------------------------------------|
+| `inputs`    | `string \| string[]` | Slice files or glob patterns (e.g. `"**/*.ice"`)                 |
+| `outputDir` | `string`             | Output directory for generated `.js` and `.d.ts` files           |
+| `include`   | `string \| string[]` | Additional Slice include directories (`-I` flags)                |
+| `args`      | `string \| string[]` | Additional slice2js CLI arguments (e.g. `--typescript`)          |
```
externl, line 97: I think we should only accept lists

js/package.json
```diff
@@ -1,74 +1,44 @@
...
-        "test_nested_modules": "file:test/Ice/moduleAggregation/module",
         "@eslint/js": "^9.32.0",
         "@rollup/plugin-node-resolve": "^16.0.1",
         "@rollup/plugin-strip": "^3.0.4",
-        "@types/node": "^22.15.24",
+        "@rollup/plugin-typescript": "^12.1.0",
+        "@types/node": "^25.2.3",
         "c8": "^10.1.3",
         "del": "^8.0.0",
+        "esbuild": "^0.27.3",
```
externl, line 19: These dev deps are used by our build system?

externl (approved):
Looks good!

js/packages/slice2js/src/slice2js.ts
pepone: Agree, already fixed.

js/packages/slice2js/src/unplugin-slice2js.ts
pepone: This seems the convention with other plugins. For example https://www.npmjs.com/package/unplugin-vue-components does the same for dirs.

js/packages/slice2js/README.md
pepone: I'm fine either way but seems common among other plugins

https://www.npmjs.com/package/unplugin-vue-components

js/package.json
pepone: For the unplugin test.

js/packages/slice2js/README.md
pepone: This is for builders that support dev mode that watch for changes, and recompile automatically.

The typical `npm run dev` where you get your application rebuild, this will also recompile Slice files.

PR #5034, opened 2026-02-09
Standardize OPTIMIZE flag checks in Makefiles

### LINKED ISSUE #4889: Review usage of OPTIMIZE=no in Makefiles
Generally we only support OPTIMIZE=yes. Everything else should mean no. However in C# and Swift we explicitly check for `no`. Is this the behavior we want? Seems confusing.

C#
```
 $(if $(filter $(OPTIMIZE),no),/p:Configuration=Debug)
 ```
Swift:
```
CONFIG ?= $(if $(filter $(OPTIMIZE),no),Debug,Release)
```

_Originally posted by @externl in https://github.com/zeroc-ice/ice/pull/4880#discussion_r2691110682_

### PR DESCRIPTION
**Pull request by pepone**
## Summary
- Standardize all `OPTIMIZE` flag checks to use `$(filter yes,$(OPTIMIZE))` pattern consistently across C#, Swift, and C++ iOS Makefiles
- Previously C# and Swift explicitly checked for `no` while C++ checked for `yes`, causing inconsistent build configurations when `OPTIMIZE` was set to an unexpected value (e.g. `OPTIMIZE=foo` would produce Debug C++ builds but Release C#/Swift builds)

Fixes #4889

### DIFF
--- a/csharp/Makefile
+++ b/csharp/Makefile
@@ -1,17 +1,22 @@
 # Copyright (c) ZeroC, Inc.
 
+top_srcdir := ..
+
+include $(top_srcdir)/config/Make.rules
+
 DOTNETBUILDARGS = $(if $(filter yes,$(OPTIMIZE)),,/p:Configuration=Debug)
 
-all:
+all::
 	dotnet msbuild msbuild/ice.proj $(DOTNETBUILDARGS)
 
-tests:
+tests::
 	dotnet msbuild msbuild/ice.proj $(DOTNETBUILDARGS)
 
-srcs:
+srcs::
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: csharp/Makefile [hunk 1]
```diff
@@ -1,17 +1,22 @@
 # Copyright (c) ZeroC, Inc.
 
-DOTNETBUILDARGS = $(if $(filter $(OPTIMIZE),no),/p:Configuration=Debug)
+top_srcdir := ..
 
-all:
+include $(top_srcdir)/config/Make.rules
```
pepone, line 5: We need to include this for the default `OPTIMIZE=yes` setting.

externl (approved):

Issue #5015, opened 2026-02-04
Simplify OutputStream::writeConverted implementation (C++)


bernardnormier:
The implementation of writeConverted (to convert a narrow string and write it as UTF-8 bytes) is more complicated than it needs to be.

The Slice 1.x size encoding does not require using the minimum number of bytes: you can encode a small size on 5 bytes.

PR #5039, opened 2026-02-10
Simplify OutputStream::writeConverted size encoding

pepone:
## Summary
- Replace the guess-and-fixup pattern in `OutputStream::writeConverted` with a simpler threshold-based approach: strings <= 84 bytes use 1-byte size encoding (worst-case 3x expansion gives 252 <= 254), larger strings use 5-byte size encoding unconditionally. This eliminates the `memmove` fixups when the guessed size encoding didn't match the actual UTF-8 size.
- Add `echoString` operation and string converter test cases exercising both code paths with 1x, 2x, and 3x UTF-8 expansion for small and large strings.

Fixes https://github.com/zeroc-ice/ice/issues/5015

### DIFF
--- a/cpp/src/Ice/OutputStream.cpp
+++ b/cpp/src/Ice/OutputStream.cpp
@@ -703,8 +703,24 @@ Ice::OutputStream::write(const char*)
 void
 Ice::OutputStream::writeConverted(const char* vdata, size_t vsize)
 {
+    StringConverterPtr stringConverter = _stringConverter;
+    if (!stringConverter)
+    {
+        stringConverter = getProcessStringConverter();
+    }
+
+    if (!stringConverter)
+    {
+        // No converter installed; write the string as-is (assumed to be UTF-8 already).
+        writeSize(static_cast<int32_t>(vsize));
+        Container::size_type position = b.size();
+        resize(position + vsize);
+        memcpy(&b[position], vdata, vsize);
+        return;
+    }
+
     //
-    // Convert the narrow string to UTF-8 using the installed string converter and write the result to the stream.
+    // Convert the narrow string to UTF-8 using the string converter and write the result to the stream.
     //
     // The worst-case expansion for converting a narrow string to UTF-8 is 3x (e.g., a single byte in Shift_JIS can
     // expand to 3 bytes in UTF-8). We use this upper bound to decide the size encoding:
@@ -713,27 +729,10 @@ Ice::OutputStream::writeConverted(const char* vdata, size_t vsize)
     //
     try
     {
-        StringConverterPtr stringConverter = _stringConverter;
-        if (!stringConverter)
-        {
-            stringConverter = getProcessStringConverter();
-        }
-
-        if (!stringConverter)
-        {
-            // No converter installed; write the string as-is (assumed to be UTF-8 already).
-            writeSize(static_cast<int32_t>(vsize));
-            Container::size_type position = b.size();
-            resize(position + vsize);
-            memcpy(&b[position], vdata, vsize);
-            return;
-        }
-
         if (vsize <= 254 / 3)
         {
             // The maximum UTF-8 size is vsize * 3 <= 252, which fits in a 1-byte size encoding.
-            write(uint8_t(0)); // 1-byte size placeholder
-            size_t firstIndex = b.size();
+            auto sizePos = startOneByteSize();
 
             StreamUTF8BufferI buffer(*this);
             byte* lastByte = stringConverter->toUTF8(vdata, vdata + vsize, buffer);
@@ -742,13 +741,12 @@ Ice::OutputStream::writeConverted(const char* vdata, size_t vsize)
                 resize(static_cast<size_t>(lastByte - b.begin()));
             }
 
-            assert(b.size() - firstIndex <= 254);
-            rewriteSize(static_cast<int32_t>(b.size() - firstIndex), b.begin() + firstIndex - 1);
+            endOneByteSize(sizePos);
         }
         else
         {
-            // Use the 5-byte size encoding unconditionally.
-            write(uint8_t(255)); // size marker
+            // Write the first byte of the 5-byte size encoding, followed by a 4-byte size placeholder.
+            write(uint8_t(255));
             auto sizePos = startSize();
 
             StreamUTF8BufferI buffer(*this);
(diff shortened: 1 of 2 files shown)


cpp/src/Ice/OutputStream.cpp
```diff
@@ -704,79 +704,61 @@ void
...
+                resize(static_cast<size_t>(lastByte - b.begin()));
             }
 
-            if (guessedSize <= 254)
-            {
-                rewriteSize(actualSize, b.begin() + firstIndex - 1);
-            }
-            else
+            assert(b.size() - firstIndex <= 254);
+            rewriteSize(static_cast<int32_t>(b.size() - firstIndex), b.begin() + firstIndex - 1);
```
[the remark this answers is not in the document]
pepone: The 3x upper bound is safe for all standard narrow string encodings — no single-byte encoding maps to characters beyond the BMP (which would require 4 UTF-8 bytes), and multi-byte encodings like Shift_JIS expand at most 1.5x (2 input bytes → 3 UTF-8 bytes). A custom converter exceeding 3x would be broken. The `assert` catches this during development; in release builds the precondition is simply assumed correct, same as other asserts throughout the codebase.

cpp/test/Ice/stringConverter/Client.cpp
```diff
@@ -128,6 +128,42 @@ Client::run(int argc, char** argv)
...
+        //
+        cout << "testing writeConverted with small and large strings... " << flush;
+
+        // Small strings (vsize <= 84): uses 1-byte size encoding
+        test(proxy->echoString("") == "");
+        test(proxy->echoString(string(50, 'a')) == string(50, 'a'));
+        test(proxy->echoString(string(50, char(0xE9))) == string(50, char(0xE9)));
+        test(proxy->echoString(string(50, char(0xA4))) == string(50, char(0xA4)));
+        test(proxy->echoString(string(84, char(0xE9))) == string(84, char(0xE9)));
+
```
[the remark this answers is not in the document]
pepone: Good catch — added `string(84, char(0xA4))` in f37ebc1. This tests the worst-case boundary: 84 euro signs → 252 UTF-8 bytes, which is the maximum that fits in 1-byte size encoding.

externl (approved):
Looks right to me. Might be nice to add a few more comments to make things easier to follow.

cpp/src/Ice/OutputStream.cpp
```diff
@@ -704,79 +704,61 @@ void
...
             stringConverter = getProcessStringConverter();
         }
-        if (stringConverter)
-        {
-            lastByte = stringConverter->toUTF8(vdata, vdata + vsize, buffer);
-            converted = true;
-        }
 
-        if (!converted)
+        if (!stringConverter)
```
bernardnormier: I think it would be clearer to move all this "no string converter" code at the very beginning of the function, outside the try/catch block and before the comment.

cpp/src/Ice/OutputStream.cpp
```diff
@@ -704,79 +704,61 @@ void
...
+        if (vsize <= 254 / 3)
         {
-            resize(static_cast<size_t>(lastByte - b.begin()));
-        }
-        size_t lastIndex = b.size();
-
-        auto actualSize = static_cast<int32_t>(lastIndex - firstIndex);
+            // The maximum UTF-8 size is vsize * 3 <= 252, which fits in a 1-byte size encoding.
+            write(uint8_t(0)); // 1-byte size placeholder
+            size_t firstIndex = b.size();
```
bernardnormier: I would add a comment here, like: firstIndex is the index of the first byte after the size.

cpp/src/Ice/OutputStream.cpp
```diff
@@ -704,79 +704,61 @@ void
...
-                rewriteSize(actualSize, b.begin() + firstIndex - 1);
-            }
-            else
+            assert(b.size() - firstIndex <= 254);
+            rewriteSize(static_cast<int32_t>(b.size() - firstIndex), b.begin() + firstIndex - 1);
+        }
+        else
+        {
+            // Use the 5-byte size encoding unconditionally.
+            write(uint8_t(255)); // size marker
```
bernardnormier: "size marker" is rather confusing here. It's just the first byte of a size encoding on 5 bytes.

cpp/src/Ice/OutputStream.cpp
```diff
@@ -704,79 +704,61 @@ void
...
+
+            StreamUTF8BufferI buffer(*this);
+            byte* lastByte = stringConverter->toUTF8(vdata, vdata + vsize, buffer);
+            if (lastByte != b.end())
             {
-                rewriteSize(actualSize, b.begin() + firstIndex - 1 - 4);
+                resize(static_cast<size_t>(lastByte - b.begin()));
             }
+
+            endSize(sizePos);
```
bernardnormier, line 759: This startSize/endSize is obviously much nicer than the 1-byte conditional. I am wondering if we could add a startOneByteSize/endOneByteSize helpers to get similar code for 1-byte?

cpp/src/Ice/OutputStream.cpp
pepone: Done in 02b045f — moved the converter lookup and early-return path to the top of `writeConverted`, before the try/catch block and the algorithm comment.

cpp/src/Ice/OutputStream.cpp
pepone: Addressed in 02b045f — the `firstIndex` variable is now gone, replaced by the `startOneByteSize()`/`endOneByteSize()` helpers which make the pattern self-documenting.

cpp/src/Ice/OutputStream.cpp
pepone: Fixed in 02b045f — changed to: "Write the first byte of the 5-byte size encoding, followed by a 4-byte size placeholder."

cpp/src/Ice/OutputStream.cpp
pepone: Done in 02b045f — added `startOneByteSize()`/`endOneByteSize()` private helpers to `OutputStream.h`. The two branches are now symmetric:

```cpp
if (vsize <= 254 / 3)
{
    auto sizePos = startOneByteSize();
    // ... convert ...
    endOneByteSize(sizePos);
}
else
{
    write(uint8_t(255));
    auto sizePos = startSize();
    // ... convert ...
    endSize(sizePos);
}
```

PR #5043, opened 2026-02-10
Fix Incorrect bool Passed to `createUnit` and `parse`

InsertCreativityHere:
It looks like these two bools accidentally got swapped when this code was moved from `slice2py/Main.cpp` to `slice2py/PythonUtil.cpp`. That's a little scary to think about.

`createUnit`'s 2nd parameter is `bool all`, which tells the Slice compiler whether or not to generate code for included files. Most languages obviously don't want that, but Ruby, PHP, and MATLAB sometimes do.

`parse`'s 3rd parameter is `bool debugMode`, which tells the Slice compiler whether or not to print debugging information to stdout while running.


bernardnormier:
I included the same fix in #5049.

PR #5049, opened 2026-02-11
Add case normalization to Slice library

bernardnormier:
This PR adds opt-in case normalization to the Slice library. It's currently implemented only for C#.


externl (approved):

cpp/src/slice2matlab/Main.cpp
```diff
@@ -218,7 +218,7 @@ namespace
                     return EXIT_FAILURE;
                 }
 
-                unit = Unit::createUnit("matlab", all);
+                unit = Unit::createUnit("matlab", {.all = all});
```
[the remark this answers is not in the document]
bernardnormier: It actually works fine.

cpp/src/slice2cs/Main.cpp
```diff
@@ -232,7 +232,7 @@ compile(const vector<string>& argv)
                 return EXIT_FAILURE;
             }
 
-            unit = Unit::createUnit("cs", false);
+            unit = Unit::createUnit("cs");
             int parseStatus = unit->parse(fileName, preprocessedHandle, debug);
```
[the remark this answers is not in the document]
bernardnormier: I will add a command-line option in a follow-up PR.

pepone (approved):

cpp/src/Slice/Parser.cpp
```diff
@@ -807,6 +816,19 @@ Slice::Contained::Contained(const ContainerPtr& container, string name) : _conta
...
+Slice::Contained::getCustomMappedName() const
+{
+    if (auto customName = getMetadataArgs(unit()->languageName() + ":identifier"))
+    {
+        return *customName;
+    }
+    else
+    {
+        return nullopt;
+    }
```
InsertCreativityHere: `getMetadataArgs` already returns the exact same `optional<string>`. So, could this just be:
```suggestion
    return getMetadataArgs(unit()->languageName() + ":identifier");
```

cpp/src/Slice/Parser.h
```diff
@@ -1154,6 +1177,7 @@ namespace Slice
 
         const std::string _languageName;
         bool _all;
+        bool _normalizeCase;
```
InsertCreativityHere: This doesn't feel like a command-line flag. It feels like file metadata (Unless you pass in `--icerpc`, then you get case-normalization no matter what of course).

Our command line options are things like `output-dir`, `depend-xml`, `debug`; nothing that affects the language mapping. Whereas metadata is _all about_ altering the language mapping you get out of the box.

cpp/src/Slice/Util.h
```diff
@@ -14,6 +14,15 @@ namespace Slice
...
     std::string baseName(const std::string& path);
     std::string dirName(const std::string& path);
+
+    /// Converts a string to Pascal case. The input string must be a valid Slice identifier. For example, "myClass"
+    /// becomes "MyClass".
+    std::string toPascalCase(const std::string& s);
+
+    /// Converts a string to camel case. The input string must be a valid Slice identifier. For example, "MyClass"
+    /// becomes "myClass".
+    std::string toCamelCase(const std::string& s);
```
InsertCreativityHere: Why add this extra condition `The input string must be a valid Slice identifier`. The implementations don't seem to care, and these functions would actually be very useful outside of just your use case.

If you `rg -i toupper` in `cpp`, there's plenty of places that could be updated to use your new functions (which I'm happy to do in a separate PR after yours, since it's not the most exciting work :))

```suggestion
    /// Converts a string to Pascal case. For example, "myClass" becomes "MyClass".
    std::string toPascalCase(const std::string& s);

    /// Converts a string to camel case. For example, "MyClass" becomes "myClass".
    std::string toCamelCase(const std::string& s);
```

cpp/src/Slice/Parser.cpp
```diff
@@ -641,13 +641,22 @@ string
...
     {
         return *customName;
     }
 
+    if (unit()->normalizeCase())
+    {
+        if (unit()->languageName() == "cs")
+        {
+            // Most C# identifiers use PascalCase. We override mappedName for camel case identifiers.
+            return toPascalCase(_name);
```
InsertCreativityHere: We used to have language-specific code mixed and tangled all through `libSlice`, and making it truly language-agnostic was a big undertaking in the name of cleanliness.

Adding a C# specific code-path to the language-independent `libSlice` feels like a regression. Especially if the end goal is to support this in all languages, which I think would be an awesome thing to do for 3.9, but would cause the number of overloads and branches to become large.

### A Proposal:

Let compilers inject the case normalization into `libSlice` by adding a `function<string(Contained)> normalizeName` parameter to `createUnit`. It would be defaulted for now (until we add case normalization in all the languages).

For example, C# would pass in:
```
string
Slice::Csharp::normalizeName(ContainedPtr& p)
{
    if (auto parameter = dynamic_pointer_cast<Parameter>(p))
    {
        return toCamelCase(p->name());
    }
    else
    {
        return toPascalCase(p->name());
    }
}
```
1) It's centralized. If you want to see the case-normalization logic, it's all in one nice place, no need to hunt around for overrides in the massive `Parser.cpp` where you'll probably miss one anyways.
2) It's keeps language-specific implementation details in the language-specific compilers, which is cleaner and will help me sleep at night. 
3) It's more declarative than overrides and if-language-equals branches. Declarative is nice.
4) No need for to add overrides. Here we only needed one, but if we added normalization for Java let's say, we'd need an override for modules and operations too (lowercase in Java). Swift would force us to add an override for enumerators. etc. It's just opening a flood-gate.

InsertCreativityHere (commented):
I understand that case-normalization is critical to have for `icerpc` support, but are we planning on supporting it in non-icerpc languages too? It's something we talked about doing in Ice many times before, and I think it would be good to make this a goal for 3.9!

cpp/src/Slice/Parser.cpp
bernardnormier: Reworked PR only those lines.

cpp/src/Slice/Parser.cpp
```diff
@@ -4842,9 +4853,12 @@ Slice::Unit::getTopLevelModules(const string& file) const
     }
 }
 
-Slice::Unit::Unit(string languageName, bool all) : _languageName(std::move(languageName)), _all(all)
+Slice::Unit::Unit(string languageName, UnitOptions options)
+    : _languageName(std::move(languageName)),
+      _all(options.all),
+      _defaultMappedName(std::move(options.defaultMappedName))
 {
-    if (!languageName.empty())
```
bernardnormier, line 4847: Unrelated bug fix

cpp/src/Slice/Parser.h
```diff
@@ -1063,14 +1069,26 @@ namespace Slice
...
 
+    /// Options for creating a Unit.
+    struct UnitOptions
+    {
+        /// When true, generate code for Slice definitions in included files.
+        bool all{false};
+
+        /// Provides a function that returns the default mapped name for a Contained element. nullptr, the default,
+        /// is equivalent to using the element's Slice name as its mapped name.
+        std::function<std::string(const Contained&)> defaultMappedName{nullptr};
```
bernardnormier, line 1080: Can't easily use a ContainedPtr as Contained is not "enable shared from this".

externl (approved):

InsertCreativityHere (approved):
Looks good to me!

pepone (approved):

PR #5054, opened 2026-02-12
Fix release workflow bugs that could cause publish failures

pepone:
- Replace hardcoded --repo zeroc-ice/ice with ${{ github.repository }} to enable fork testing
- Remove unreachable stable code path in brew publish (job only runs for nightly)
- Fix gem wildcard upload by using a loop instead of curl --upload-file with glob
- Add NODE_AUTH_TOKEN env to NPM stable publish step
- Rename job IDs from 'build' to 'publish' for consistency in nuget, pip, and maven workflows
- Add build_number to nightly release to differentiate retry attempts

### DIFF
--- a/.github/workflows/publish-npm-packages.yml
+++ b/.github/workflows/publish-npm-packages.yml
@@ -73,5 +73,3 @@ jobs:
             echo "Publishing package: $pkg"
             npm publish "$pkg" --provenance
           done
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}


.github/workflows/publish-npm-packages.yml
```diff
@@ -73,3 +73,5 @@ jobs:
             echo "Publishing package: $pkg"
             npm publish "$pkg" --provenance
           done
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
```
pepone: I think this is not required when using trusted publisher and actions/setup-node

See example in: https://docs.npmjs.com/trusted-publishers

This was already tested in 3.7.11 publishing, and worked without the extra env.

PR #5058, opened 2026-02-12
Modernize RPM spec files

pepone:
- Fix README version: libice3.8-c++ -> libice3.9-c++
- Fix spec version: 3.8.0~alpha0 -> 3.9.0~alpha0
- Fix grammar: "packages doesn't" -> "packages do not"
- Add %{mmversion} macro to centralize major.minor version
- Replace all hardcoded "3.9" in package names with %{mmversion}
- Fix ice2slice description indentation (remove leading spaces)
- Remove unnecessary ldconfig calls from packages without libraries: icebox, ice-utils, dsnode, icebridge, icegrid, glacier2
- Extract %create_ice_user macro for DRY user creation
- Fix icegridgui launcher: quote "$@" and use direct JAR path
- Remove deprecated Group: tag from repo specs

### DIFF
--- a/packaging/rpm/ice.spec
+++ b/packaging/rpm/ice.spec
@@ -7,7 +7,7 @@
 %define create_ice_user() \
 getent group ice > /dev/null || groupadd -r ice \
 getent passwd ice > /dev/null || \\\
-  useradd -r -g ice -d %{_localstatedir}/lib/ice -s /sbin/nologin -c "Ice Service account" ice
+  useradd -r -g ice -d %{_localstatedir}/lib/ice -s /sbin/nologin -c "Ice service account" ice
 
 # git_tag, when defined, is typically a branch, for example 3.7
 %if 0%{?git_tag:1}


packaging/rpm/ice.spec
```diff
@@ -1,13 +1,22 @@
 # Copyright (c) ZeroC, Inc.
 
+# Major.minor version for library sonames and package names
+%define mmversion 3.9
+
+# Macro for creating the ice user and group
+%define create_ice_user() \
+getent group ice > /dev/null || groupadd -r ice \
+getent passwd ice > /dev/null || \\\
+  useradd -r -g ice -d %{_localstatedir}/lib/ice -s /sbin/nologin -c "Ice Service account" ice
```
externl: I wonder if it should really be "Ice service account"
```suggestion
  useradd -r -g ice -d %{_localstatedir}/lib/ice -s /sbin/nologin -c "Ice service account" ice
```

packaging/rpm/ice.spec
```diff
@@ -442,10 +451,6 @@ exit 0
 %doc packaging/rpm/README
 %{_bindir}/icebox
 %{_mandir}/man1/icebox.1*
-%post -n %{?nameprefix}icebox -p /sbin/ldconfig
-%postun -n %{?nameprefix}icebox
```
externl, line 446: what was all of this about?

pepone: You only need this when the package includes shared libraries.

PR #5065, opened 2026-02-12
Add Slice dependency tracking to CMake slice2cpp_generate

### LINKED ISSUE #3642: Generate improved depends file with slice2cpp

pepone:
## Summary
- Add transitive Slice dependency tracking using `slice2cpp --depend` and CMake's `DEPFILE` support, so changes to included `.ice` files trigger recompilation automatically
- Add `slice2cpp` executable as an explicit dependency so upgrading Ice triggers recompilation of all Slice files
- New helper script `slice2cpp_depend.cmake` that runs `--depend` at build time and transforms the output for CMake's depfile format

Fixes #3642

## Test plan
- [x] Build an ice-demo project (e.g., `cpp/Ice/greeter`) with this change
- [x] Modify an included `.ice` file from the Ice SDK and rebuild — verify recompilation is triggered
- [x] Verify clean builds work correctly
- [x] Test with both Ninja and Make generators

### DIFF
--- a/cpp/config/slice2cpp.cmake
+++ b/cpp/config/slice2cpp.cmake
@@ -1,8 +1,5 @@
 # Copyright (c) ZeroC, Inc.
 
-# Capture the directory containing this script so we can locate the helper script at build time.
-set(_slice2cpp_cmake_dir "${CMAKE_CURRENT_LIST_DIR}")
-
 # Function to generate C++ source files from Slice (.ice) files for a target using slice2cpp
 # The target must have the Slice files in its sources
 # The generated files are added to the target sources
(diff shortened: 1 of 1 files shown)


cpp/config/slice2cpp.cmake
```diff
@@ -1,5 +1,8 @@
 # Copyright (c) ZeroC, Inc.
 
+# Capture the directory containing this script so we can locate the helper script at build time.
+set(_slice2cpp_cmake_dir "${CMAKE_CURRENT_LIST_DIR}")
```
externl: I don't think we really need this extra var do we?

pepone: Need to test with our demos. I haven't tested it yet.

externl: Yes, we should definitely test it! Pretty sure we can also remove this var.

PR #5068, opened 2026-02-13
Fix slice2js unplugin type declarations leaking all bundler dependencies

### LINKED ISSUE #5067: slice2js unplugin: type declarations leak all bundler dependencies
## Problem

The `@zeroc/slice2js` package uses [unplugin](https://github.com/unjs/unplugin) to provide bundler-agnostic plugins. When a consumer imports a specific bundler adapter, e.g.:

```ts
import slice2js from "@zeroc/slice2js/unplugin/esbuild";
```

TypeScript resolves the type chain through `unplugin`'s main declaration file (`unplugin/dist/index.d.mts`), which imports types from **every** supported bundler:

```ts
// unplugin/dist/index.d.mts
import { CompilationContext, JsPlugin } from "@farmfe/core";
import { Compilation, Compiler as RspackCompiler, ... } from "@rspack/core";
import { BunPlugin, PluginBuilder } from "bun";
import { BuildOptions, Loader, Plugin as EsbuildPlugin, ... } from "esbuild";
import { Plugin as RolldownPlugin } from "rolldown";
import { EmittedAsset, Plugin as RollupPlugin, ... } from "rollup";
import { Plugin as UnloaderPlugin } from "unloader";
import { Plugin as VitePlugin } from "vite";
import { Compilation as Compilation$1, Compiler as WebpackCompiler, ... } from "webpack";
```

Since the consumer only installs their chosen bundler (e.g., `esbuild`), `tsc` fails with errors for all the missing bundler modules:

```
node_modules/unplugin/dist/index.d.ts(2,46): error TS2307: Cannot find module '@farmfe/core' or its corresponding type declarations.
node_modules/unplugin/dist/index.d.ts(3,94): error TS2307: Cannot find module '@rspack/core' or its corresponding type declarations.
node_modules/unplugin/dist/index.d.ts(5,42): error TS2307: Cannot find module 'rolldown' or its corresponding type declarations.
node_modules/unplugin/dist/index.d.ts(6,98): error TS2307: Cannot find module 'rollup' or its corresponding type declarations.
node_modules/unplugin/dist/index.d.ts(7,42): error TS2307: Cannot find module 'unloader' or its corresponding type declarations.
node_modules/unplugin/dist/index.d.ts(8,38): error TS2307: Cannot find module 'vite' or its corresponding type declarations.
node_modules/unplugin/dist/index.d.ts(9,132): error TS2307: Cannot find module 'webpack' or its corresponding type declarations.
```

This forces every consumer to set `"skipLibCheck": true` in their `tsconfig.json`.

This issue exists in both `unplugin@2.3.11` and the latest `unplugin@3.0.0`.

See https://github.com/unjs/unplugin/issues/349

## Suggested solution

Ship hand-written type declarations for each bundler entry point in `@zeroc/slice2js`, so that `unplugin`'s barrel type file is never resolved by the consumer's `tsc`.

### 1. Create per-bundler type declarations

Each entry point gets its own `.d.ts` that only imports from the relevant bundler:

```ts
// unplugin/esbuild.d.ts
import type { Plugin } from "esbuild";

interface Slice2JsOptions {
    inputs: string[];
    args?: string[];
    outputDir?: string;
}

export default function slice2js(options: Slice2JsOptions): Plugin;
```

```ts
// unplugin/vite.d.ts
import type { Plugin } from "vite";

interface Slice2JsOptions {
    inputs: string[];
    args?: string[];
    outputDir?: string;
}

export default function slice2js(options: Slice2JsOptions): Plugin;
```

And similarly for `rollup`, `webpack`, etc.

### 2. Wire up the types via `exports` in `package.json`

Use the `exports` field to map each subpath to its own type declaration:

```json
{
    "exports": {
        "./unplugin/esbuild": {
            "types": "./unplugin/esbuild.d.ts",
            "import": "./unplugin/esbuild.mjs"
        },
        "./unplugin/vite": {
            "types": "./unplugin/vite.d.ts",
            "import": "./unplugin/vite.mjs"
        },
        "./unplugin/rollup": {
            "types": "./unplugin/rollup.d.ts",
            "import": "./unplugin/rollup.mjs"
        },
        "./unplugin/webpack": {
            "types": "./unplugin/webpack.d.ts",
            "import": "./unplugin/webpack.mjs"
        }
    }
}
```

### Why this works

- The **runtime code** still uses `unplugin` internally -- no runtime changes needed.
- The **type declarations** are isolated per bundler, so `tsc` only resolves types for the bundler the consumer actually uses.
- Consumers no longer need `skipLibCheck: true` to work around missing bundler type declarations.
- A shared `Slice2JsOptions` interface can be extracted into a common `.d.ts` file to avoid duplication.

### Impact on consumers

No consumer-side changes are required. The `exports` field with `"types"` takes priority in TypeScript's module resolution, so existing import statements like `import slice2js from "@zeroc/slice2js/unplugin/esbuild"` automatically resolve to the new narrow declarations. Consumers can then remove `skipLibCheck: true` from their `tsconfig.json` once the fix is published.

pepone:
## Summary

- Add hand-written type declarations for each bundler entry point (esbuild, vite, rollup, webpack)
- These declarations only import from the specific bundler, not from `unplugin`
- Update prepack script to copy hand-written types over auto-generated ones

Fixes #5067

## Test plan

- [x] Tested with `Ice/greeter` demo (esbuild) - `tsc --noEmit --skipLibCheck false` passes
- [x] Tested with `Ice/react-greeter` demo (vite) - `tsc --noEmit --skipLibCheck false` passes

### DIFF
--- a/js/packages/slice2js/scripts/prepack.js
+++ b/js/packages/slice2js/scripts/prepack.js
@@ -82,15 +82,13 @@ let binDestDir = path.resolve(__dirname, "../bin");
 const typesSourceDir = path.resolve(__dirname, "../types/unplugin");
 const typesDestDir = path.resolve(__dirname, "../dist/unplugin");
 
-if (fs.existsSync(typesSourceDir)) {
-    console.log(`Copying unplugin type declarations from ${typesSourceDir} to ${typesDestDir}...`);
-    fs.readdirSync(typesSourceDir).forEach(file => {
-        if (file.endsWith(".d.ts")) {
-            fs.copyFileSync(path.join(typesSourceDir, file), path.join(typesDestDir, file));
-        }
-    });
-    console.log("Done.");
-}
+console.log(`Copying unplugin type declarations from ${typesSourceDir} to ${typesDestDir}...`);
+fs.readdirSync(typesSourceDir).forEach(file => {
+    if (file.endsWith(".d.ts")) {
+        fs.copyFileSync(path.join(typesSourceDir, file), path.join(typesDestDir, file));
+    }
+});
+console.log("Done.");
 
 if (stagingPath) {
     // Check that all compilers are available in the staging path


js/packages/slice2js/scripts/prepack.js
```diff
@@ -76,6 +76,22 @@ console.log("Done.");
 const stagingPath = process.env.SLICE2JS_STAGING_PATH;
 let binDestDir = path.resolve(__dirname, "../bin");
 
+// Copy hand-written type declarations for unplugin adapters.
+// These replace the auto-generated ones to avoid leaking all bundler dependencies via unplugin's barrel types.
+// See https://github.com/zeroc-ice/ice/issues/5067
+const typesSourceDir = path.resolve(__dirname, "../types/unplugin");
+const typesDestDir = path.resolve(__dirname, "../dist/unplugin");
+
+if (fs.existsSync(typesSourceDir)) {
```
externl: Does this ever not exist?

pepone: You are right, this check is not required!

PR #5086, opened 2026-02-15
Fix WebSocket URL construction for IPv6 and wss default port

### LINKED ISSUE #5083: Fix WebSocket URL port handling for wss in WSTransceiver.js
## Description

`WSTransceiver.js` only omits the port from the WebSocket URL when `port === 80`, regardless of whether the scheme is `ws` or `wss`. The default port for `wss` is 443, not 80.

### Affected line

- [WSTransceiver.js line 33](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/WSTransceiver.js#L33)

### Current code

```js
let url = secure ? "wss" : "ws";
url += "://" + addr.host;
if (addr.port !== 80) {
    url += ":" + addr.port;
}
```

### Expected code

```js
let url = secure ? "wss" : "ws";
url += "://" + addr.host;
if (addr.port !== (secure ? 443 : 80)) {
    url += ":" + addr.port;
}
```

### Impact

- **`wss` on port 443:** Port is unnecessarily included in the URL (`wss://host:443/`). This is typically harmless since browsers handle it correctly, but it's not standard.
- **`wss` on port 80:** Port is incorrectly omitted (`wss://host/`), which would cause the browser to connect to port 443 (the default for `wss`) instead of the intended port 80.

Note: The C++ implementation doesn't have this issue because it uses raw sockets with HTTP upgrade requests rather than constructing WebSocket URLs.

### LINKED ISSUE #5084: Fix WebSocket URL construction for IPv6 literal addresses in WSTransceiver.js
## Description

`WSTransceiver.js` concatenates `addr.host` directly into the WebSocket URL without wrapping IPv6 literal addresses in brackets. Per RFC 3986 and RFC 6455, IPv6 literals in URLs must be enclosed in brackets (`[::1]`), otherwise the colons are misinterpreted as port separators, producing an invalid URL.

### Affected line

- [WSTransceiver.js line 32](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/WSTransceiver.js#L32)

### Current code

```js
let url = secure ? "wss" : "ws";
url += "://" + addr.host;
if (addr.port !== 80) {
    url += ":" + addr.port;
}
```

For an IPv6 host like `::1` on port 4061, this produces:
```
ws://::1:4061/
```
which is an invalid URL — the colons in the address are ambiguous with the port separator.

### Expected behavior

IPv6 addresses should be wrapped in brackets:
```
ws://[::1]:4061/
```

The codebase already detects IPv6 addresses by checking for `:` in the host string (see [IPEndpointI.js line 89](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/IPEndpointI.js#L89)):

```js
const addQuote = this._host.indexOf(":") != -1;
```

A similar check should be applied when constructing the WebSocket URL.

pepone:
## Summary
- Wrap IPv6 literal addresses in brackets when constructing WebSocket URLs, per RFC 3986/6455 (#5084)
- Use 443 as the default port for `wss` scheme instead of always using 80 (#5083)

Fixes #5083
Fixes #5084

### DIFF
--- a/js/packages/ice/src/Ice/WSTransceiver.js
+++ b/js/packages/ice/src/Ice/WSTransceiver.js
@@ -29,8 +29,9 @@ class WSTransceiver {
         this._writeReadyTimeout = 0;
 
         let url = secure ? "wss" : "ws";
-        url += "://" + addr.host;
-        if (addr.port !== 80) {
+        const isIPv6 = addr.host.indexOf(":") !== -1;
+        url += "://" + (isIPv6 ? `[${addr.host}]` : addr.host);
+        if (addr.port !== (secure ? 443 : 80)) {
             url += ":" + addr.port;
         }
         url += resource ? resource : "/";


js/packages/ice/src/Ice/WSTransceiver.js
```diff
@@ -29,8 +29,9 @@ class WSTransceiver {
         this._writeReadyTimeout = 0;
 
         let url = secure ? "wss" : "ws";
-        url += "://" + addr.host;
-        if (addr.port !== 80) {
+        const isIPv6 = addr.host.indexOf(":") !== -1;
```
externl, line 32: why not use `address.host.includes(":")`?

pepone: No good reason. We use `indexOf` for strings all over the place, we should probably updates its usage everywhere in a follow up PR.

PR #5087, opened 2026-02-15
Fix exception handling in CurrentExtensions.js

### LINKED ISSUE #5080: Fix exception path in createEmptyOutgoingResponse calling non-existent method
## Description

In `CurrentExtensions.js`, the catch block in `createEmptyOutgoingResponse` calls `this.createOutgoingResponse(ex)` which does not exist. The correct method name is `createOutgoingResponseWithException`. This results in a `TypeError` at runtime if `writeEmptyEncapsulation` throws.

### Affected line

- [CurrentExtensions.js line 41](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/CurrentExtensions.js#L41)

### Current code

```js
Current.prototype.createEmptyOutgoingResponse = function () {
    const ostr = startReplyStream(this);
    if (this.requestId != 0) {
        try {
            ostr.writeEmptyEncapsulation(this.encoding);
        } catch (ex) {
            return this.createOutgoingResponse(ex); // non-existent method
        }
    }
    return new OutgoingResponse(ostr);
};
```

### Expected code

```js
        } catch (ex) {
            return this.createOutgoingResponseWithException(ex);
        }
```

### Impact

If `writeEmptyEncapsulation` throws an exception, the catch block produces a `TypeError: this.createOutgoingResponse is not a function` instead of properly marshaling the error into an outgoing response. Note that `createOutgoingResponseWithResult` (line 27) already uses the correct method name.

The C++ reference implementation correctly calls `makeOutgoingResponse` (the exception variant) in the same path:

https://github.com/zeroc-ice/ice/blob/main/cpp/src/Ice/OutgoingResponse.cpp#L228

### LINKED ISSUE #5081: Fix exception fallback in createOutgoingResponseWithException retrying with same exception
## Description

In `CurrentExtensions.js`, the catch block in `createOutgoingResponseWithException` retries `createOutgoingResponseCore` with the same original `exception` instead of the newly caught marshal failure. This makes the fallback retry pointless — it will fail the same way again.

### Affected lines

- [CurrentExtensions.js lines 47-54](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/CurrentExtensions.js#L47-L54)

### Current code

```js
Current.prototype.createOutgoingResponseWithException = function (exception) {
    try {
        return createOutgoingResponseCore(this, exception);
    } catch {
        // Try a second time with the marshal exception. This should not fail.
        return createOutgoingResponseCore(this, exception); // same exception — will fail the same way
    }
};
```

### Expected code

```js
Current.prototype.createOutgoingResponseWithException = function (exception) {
    try {
        return createOutgoingResponseCore(this, exception);
    } catch (ex) {
        // Try a second time with the marshal exception. This should not fail.
        return createOutgoingResponseCore(this, ex);
    }
};
```

### Reference

The C++ implementation correctly passes the *new* exception (`current_exception()`) on retry:

```cpp
OutgoingResponse makeOutgoingResponse(exception_ptr exc, const Current& current) noexcept
{
    try {
        return makeOutgoingResponseCore(exc, current);
    } catch (...) {
        return makeOutgoingResponseCore(current_exception(), current);
    }
}
```

https://github.com/zeroc-ice/ice/blob/main/cpp/src/Ice/OutgoingResponse.cpp#L265-L276

The intent (confirmed by the comment) is: if marshaling a `UserException` in sliced format fails, retry by marshaling the marshal failure itself as an `UnknownException`, which requires no complex encoding and should always succeed.

pepone:
## Summary
- Fix `createEmptyOutgoingResponse` to call the correct method `createOutgoingResponseWithException` instead of non-existent `createOutgoingResponse` (#5080)
- Fix `createOutgoingResponseWithException` to retry with the caught marshal exception (`ex`) instead of the original exception (#5081)

Fixes #5080
Fixes #5081

### DIFF
--- a/js/packages/ice/src/Ice/CurrentExtensions.js
+++ b/js/packages/ice/src/Ice/CurrentExtensions.js
@@ -38,7 +38,7 @@ Current.prototype.createEmptyOutgoingResponse = function () {
         try {
             ostr.writeEmptyEncapsulation(this.encoding);
         } catch (ex) {
-            return this.createOutgoingResponse(ex);
+            return this.createOutgoingResponseWithException(ex);
         }
     }
     return new OutgoingResponse(ostr);
@@ -47,9 +47,9 @@ Current.prototype.createEmptyOutgoingResponse = function () {
 Current.prototype.createOutgoingResponseWithException = function (exception) {
     try {
         return createOutgoingResponseCore(this, exception);
-    } catch {
+    } catch (ex) {
         // Try a second time with the marshal exception. This should not fail.
-        return createOutgoingResponseCore(this, exception);
+        return createOutgoingResponseCore(this, ex);
     }
 };
 


js/packages/ice/src/Ice/CurrentExtensions.js
```diff
@@ -47,9 +47,9 @@ Current.prototype.createEmptyOutgoingResponse = function () {
 Current.prototype.createOutgoingResponseWithException = function (exception) {
     try {
         return createOutgoingResponseCore(this, exception);
-    } catch {
+    } catch (ex) {
         // Try a second time with the marshal exception. This should not fail.
-        return createOutgoingResponseCore(this, exception);
+        return createOutgoingResponseCore(this, ex);
```
pepone, line 52: The doc comment was correct, but we were retrying with the original exception not with the marshal exception.

js/packages/ice/src/Ice/CurrentExtensions.js
```diff
@@ -38,7 +38,7 @@ Current.prototype.createEmptyOutgoingResponse = function () {
         try {
             ostr.writeEmptyEncapsulation(this.encoding);
         } catch (ex) {
-            return this.createOutgoingResponse(ex);
+            return this.createOutgoingResponseWithException(ex);
```
pepone, line 41: Seems like a copy/paste error, in JavaScript the method that handles exceptions has a different name. There is no method overloading.

One thing not clear is why we handle exceptions from `writeEmptyEncapsulation` and not from `startReplyStream`. The C# version is identical.

The C++ version is different:

https://github.com/zeroc-ice/ice/blob/478533e0746c7ce4c9f2a9e64e737ee310541726/cpp/src/Ice/OutgoingResponse.cpp#L214-L232

InsertCreativityHere: Java is identical to C# and JS here, and looking through it's implementation, `startReplyStream` can only throw an `OutOfMemoryError`, which there's no real point in catching. But `writeEmptyEncapsulation` can throw a `MarshalException` if there's an encoding version issue.

Seems the same in JS too. So I guess there just isn't a meaningful exception to catch from `startReplyStream` in these languages. Now, why is C++ different and everything gets wrapped in the `try/catch`? I have no idea. The first 3 `write` calls can only throw `bad_alloc`... not sure what the point in handling that is.

PR #5088, opened 2026-02-15
Fix Properties.js getPropertyAsList default and parse crash

### LINKED ISSUE #5079: Fix getPropertyAsList returning 0 instead of an empty array in Properties.js
## Description

`getPropertyAsList` passes `0` as the default value to `getPropertyAsListWithDefault`, instead of `[]`. When the property is unset, callers receive the number `0` instead of an empty array, which breaks array semantics.

### Affected line

- [Properties.js line 94](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/Properties.js#L94)

### Current code

```js
getPropertyAsList(key) {
    return this.getPropertyAsListWithDefault(key, 0);
}
```

### Expected code

```js
getPropertyAsList(key) {
    return this.getPropertyAsListWithDefault(key, []);
}
```

### Why it matters

`getPropertyAsListWithDefault` only normalizes `undefined` and `null` defaults to `[]` (line 103). The `0` value passes through that guard and is returned directly to callers when the property is not set. Any caller using array methods (`.length`, `.map`, `.forEach`, etc.) on the result would fail at runtime.

The C++ reference implementation correctly passes an empty `StringSeq`:

```cpp
StringSeq Properties::getPropertyAsList(string_view key)
{
    return getPropertyAsListWithDefault(key, StringSeq());
}
```

https://github.com/zeroc-ice/ice/blob/main/cpp/src/Ice/Properties.cpp#L289-L291

### LINKED ISSUE #5082: Fix Properties.parse crash on empty or whitespace-only config content
## Description

`Properties.parse()` crashes with a `TypeError` when called with an empty string or whitespace/newline-only content because `String.match()` returns `null` when there are no matches.

### Affected line

- [Properties.js line 219](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/Properties.js#L219)

### Current code

```js
parse(data) {
    data.match(/[^\r\n]+/g).forEach(line => this.parseLine(line));
}
```

When `data` is `""`, `"\n"`, `"  \r\n  "`, etc., `match()` returns `null`, and calling `.forEach()` on `null` throws:
```
TypeError: Cannot read properties of null (reading 'forEach')
```

### Expected code

```js
parse(data) {
    (data.match(/[^\r\n]+/g) ?? []).forEach(line => this.parseLine(line));
}
```

### Reference

The C++ `Properties::load` uses `getline` in a while loop, which naturally handles empty files by simply not iterating:

```cpp
while (getline(in, line))
{
    // ...
    parseLine(line, stringConverter);
}
```

https://github.com/zeroc-ice/ice/blob/main/cpp/src/Ice/Properties.cpp#L577

pepone:
## Summary
- Fix `getPropertyAsList` to pass `[]` instead of `0` as the default value (#5079)
- Fix `parse()` to handle `null` return from `String.match()` on empty/whitespace-only content (#5082)

## Test plan
- [ ] Verify `getPropertyAsList` returns an empty array when the property is not set
- [ ] Verify `parse("")` and `parse("\n")` do not throw

Fixes #5079
Fixes #5082

### DIFF
--- a/js/packages/ice/src/Ice/Properties.js
+++ b/js/packages/ice/src/Ice/Properties.js
@@ -91,7 +91,7 @@ export class Properties {
     }
 
     getPropertyAsList(key) {
-        return this.getPropertyAsListWithDefault(key, 0);
+        return this.getPropertyAsListWithDefault(key, []);
     }
 
     getIcePropertyAsList(key) {
@@ -216,7 +216,7 @@ export class Properties {
     }
 
     parse(data) {
-        data.match(/[^\r\n]+/g).forEach(line => this.parseLine(line));
+        (data.match(/[^\r\n]+/g) ?? []).forEach(line => this.parseLine(line));
     }
 
     parseLine(line) {


InsertCreativityHere (approved):
Looks good to me!

PR #5089, opened 2026-02-15
Fix encoding version validation in ConnectionI.js

### LINKED ISSUE #5078: Fix encoding version validation in ConnectionI.js
## Description

The encoding version validation in `ConnectionI.js` compares `protocolVersion.minor` instead of `encodingVersion.minor` in two locations, causing the minor version check to validate the wrong variable.

### Affected lines

- [Line 494](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/ConnectionI.js#L494)
- [Line 1071](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/ConnectionI.js#L1071)

### Current code (both locations)

```js
if (
    encodingVersion.major != Protocol.currentProtocolEncoding.major ||
    protocolVersion.minor != Protocol.currentProtocolEncoding.minor // wrong variable
) {
```

### Expected code

```js
if (
    encodingVersion.major != Protocol.currentProtocolEncoding.major ||
    encodingVersion.minor != Protocol.currentProtocolEncoding.minor
) {
```

### Impact

Currently silent because both the protocol version and encoding version are `1.0`, so `protocolVersion.minor` and `encodingVersion.minor` have the same value (`0`). The bug would cause incorrect validation if they ever diverge.

The C++ reference implementation correctly compares the encoding version in both fields:
- https://github.com/zeroc-ice/ice/blob/main/cpp/src/Ice/ConnectionI.cpp#L1295
- https://github.com/zeroc-ice/ice/blob/main/cpp/src/Ice/ConnectionI.cpp#L2623

pepone:
## Summary
- Fix two locations that incorrectly compare `protocolVersion.minor` instead of `encodingVersion.minor` when validating the encoding version in message headers (#5078)

Fixes #5078

### DIFF
--- a/js/packages/ice/src/Ice/ConnectionI.js
+++ b/js/packages/ice/src/Ice/ConnectionI.js
@@ -491,7 +491,7 @@ export class ConnectionI {
                     encodingVersion._read(this._readStream);
                     if (
                         encodingVersion.major != Protocol.currentProtocolEncoding.major ||
-                        protocolVersion.minor != Protocol.currentProtocolEncoding.minor
+                        encodingVersion.minor != Protocol.currentProtocolEncoding.minor
                     ) {
                         throw new MarshalException(
                             `Invalid protocol encoding version in message header: ${encodingVersion.major}.${encodingVersion.minor}`,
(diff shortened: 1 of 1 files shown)


js/packages/ice/src/Ice/ConnectionI.js
pepone, line 1: Seems we don't have any tests that catch this.

PR #5090, opened 2026-02-15
Fix exception chaining to use { cause: ex } consistently

### LINKED ISSUE #5085: Fix exception chaining to use { cause: ex } consistently
## Description

Several call sites pass the original exception directly as the second argument to exception constructors instead of wrapping it in `{ cause: ex }`. Since `Exception extends Error` and `Error(message, options)` expects `options.cause`, the original exception is silently lost.

The codebase already uses the correct `{ cause: ex }` pattern in many places (WSTransceiver.js, Properties.js, TcpTransceiver.js, OpaqueEndpoint.js, ObjectAdapter.js, ConnectionI.js, InputStream.js), but 6 call sites use the incorrect form.

### Affected lines

**Protocol.js:**
- [Line 189](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/Protocol.js#L189): `throw new ParseException(..., ex)`
- [Line 206](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/Protocol.js#L206): `throw new ParseException(..., ex)`

**ReferenceFactory.js:**
- [Line 213](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/ReferenceFactory.js#L213): `throw new ParseException(..., ex)`
- [Line 275](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/ReferenceFactory.js#L275): `throw new ParseException(..., e)`
- [Line 290](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/ReferenceFactory.js#L290): `throw new ParseException(..., e)`
- [Line 400](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/ReferenceFactory.js#L400): `throw new ParseException(..., ex)`

### Example

```js
// Current (cause is silently lost)
throw new ParseException(`invalid version value in '${str}'`, ex);

// Expected (cause is properly chained)
throw new ParseException(`invalid version value in '${str}'`, { cause: ex });
```

pepone:
## Summary
- Fix 6 call sites in `Protocol.js` and `ReferenceFactory.js` that pass the original exception directly as the second argument to `ParseException` constructors instead of wrapping it in `{ cause: ex }` (#5085)

## Test plan
- [ ] Verify that exceptions thrown from version parsing preserve the original cause
- [ ] Verify that exceptions thrown from proxy string parsing preserve the original cause

Fixes #5085

### DIFF
--- a/js/packages/ice/src/Ice/Protocol.js
+++ b/js/packages/ice/src/Ice/Protocol.js
@@ -186,7 +186,7 @@ function stringToMajor(str) {
         }
         return majVersion;
     } catch (ex) {
-        throw new ParseException(`invalid version value in '${str}'`, ex);
+        throw new ParseException(`invalid version value in '${str}'`, { cause: ex });
     }
 }
 
@@ -203,7 +203,7 @@ function stringToMinor(str) {
         }
         return minVersion;
     } catch (ex) {
-        throw new ParseException(`invalid version value in '${str}'`, ex);
+        throw new ParseException(`invalid version value in '${str}'`, { cause: ex });
     }
 }
 
--- a/js/packages/ice/src/Ice/ReferenceFactory.js
+++ b/js/packages/ice/src/Ice/ReferenceFactory.js
@@ -210,7 +210,7 @@ export class ReferenceFactory {
                     try {
                         facet = StringUtil.unescapeString(argument, 0, argument.length);
                     } catch (ex) {
-                        throw new ParseException(`invalid facet in ${s}': `, ex);
+                        throw new ParseException(`invalid facet in ${s}': `, { cause: ex });
                     }
 
                     break;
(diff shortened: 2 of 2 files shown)


js/packages/ice/src/Ice/Protocol.js
```diff
@@ -186,7 +186,7 @@ function stringToMajor(str) {
         }
         return majVersion;
     } catch (ex) {
-        throw new ParseException(`invalid version value in '${str}'`, ex);
+        throw new ParseException(`invalid version value in '${str}'`, { cause: ex });
```
pepone, line 189: This constructor is inherited from the base Error class:

`Ice.ParseException -> Ice.LocalException -> Ice.Exception -> Errror`

PR #5091, opened 2026-02-16
Use TextDecoder for string decoding in InputStream

pepone:
## Summary
- Replace manual UTF-8 decoding (`String.fromCharCode` + `decodeURIComponent(escape(...))`) with `TextDecoder` in `InputStream.readString`
- Rename `Buffer.getString` to `Buffer.getView`, returning a zero-copy `Uint8Array` view instead of a decoded string
- Add proper error handling: `RangeError` maps to `UnmarshalOutOfBoundsException`, invalid UTF-8 throws `MarshalException`

Backport of #5069 to 3.7.

### DIFF
--- a/js/src/Ice/Stream.js
+++ b/js/src/Ice/Stream.js
@@ -24,6 +24,11 @@ require("../Ice/Version");
 // Singleton TextEncoder for UTF-8 string encoding
 const textEncoder = new TextEncoder();
 
+// Reusable TextDecoder instance for UTF-8 string decoding.
+// - fatal: true throws on invalid UTF-8 sequences instead of replacing with U+FFFD
+// - ignoreBOM: true does not interpret a leading BOM as a special character
+const textDecoder = new TextDecoder("utf-8", { fatal: true, ignoreBOM: true });
+
 const ArrayUtil = Ice.ArrayUtil;
 const Debug = Ice.Debug;
 const ExUtil = Ice.ExUtil;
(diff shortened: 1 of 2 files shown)


js/src/Ice/Stream.js
```diff
@@ -1744,21 +1749,17 @@ class InputStream
...
+            return textDecoder.decode(this._buf.getView(len));
         }
         catch(ex)
         {
-            throw new Ice.UnmarshalOutOfBoundsException();
+            if(ex instanceof RangeError)
+            {
+                throw new Ice.UnmarshalOutOfBoundsException();
+            }
+            throw new Ice.MarshalException("invalid UTF-8 string");
```
InsertCreativityHere, line 1762: With the 3.7 API, is there a way to pass along an "underlying cause" (`ex` here) to the `MarshalException` constructor?

js/src/Ice/Stream.js
```diff
@@ -1744,21 +1749,17 @@ class InputStream
...
-
         try
         {
-            return this._buf.getString(len);
+            return textDecoder.decode(this._buf.getView(len));
         }
         catch(ex)
         {
-            throw new Ice.UnmarshalOutOfBoundsException();
+            if(ex instanceof RangeError)
```
InsertCreativityHere, line 1758: I realize this is how the old code was formatted,
but our style in JS has a space between the keyword and the parenthesis right?
```
        catch (ex)
        {
            if (ex instanceof RangeError)
```
Maybe better to fix separately, since `Buffer` is also missing the space it looks like...

js/src/Ice/Stream.js
pepone: I will check

js/src/Ice/Stream.js
pepone: I think it does not in 3.7, we changed this in 3.8.

InsertCreativityHere: Oh yeah, I forgot this was all 3.7. Not important to fix it's formatting anyways then!

PR #5102, opened 2026-02-16
Fix off-by-one in OpenSSL password callback

### LINKED ISSUE #5096: SSL: Fix critical and high-severity security bugs across backends
## Summary

Security review of `cpp/src/Ice/SSL/` identified several confirmed bugs with security impact across all three TLS backends and shared code. All fixes are surgical (1–2 lines each) and suitable for a patch release.

### C1. Schannel: Dangling pointer in hostname verification
**File:** `SchannelEngine.cpp:1411`

```cpp
extraPolicyPara.pwszServerName = const_cast<wchar_t*>(Ice::stringToWstring(host).c_str());
```

`Ice::stringToWstring(host)` returns a temporary `wstring` destroyed at the semicolon. `CertVerifyCertificateChainPolicy` (line 1422) reads a dangling pointer. Hostname verification may silently pass for the wrong certificate.

**Fix:** Store the `wstring` in a local variable:
```cpp
wstring hostW = Ice::stringToWstring(host);
extraPolicyPara.pwszServerName = const_cast<wchar_t*>(hostW.c_str());
```

### C2. Schannel: Root store handle overwritten after duplication
**File:** `SchannelTransceiverI.cpp:129,148`

Line 129 conditionally duplicates `_rootStore` into `_credentials.hRootStore`. Line 148 unconditionally overwrites it with the raw `_rootStore`. The duplicated handle leaks, and `close()` later calls `CertCloseStore` on a store it doesn't own — a potential double-close or use-after-close.

**Fix:** Remove line 148 (`_credentials.hRootStore = _rootStore;`).

### C3. OpenSSL: Off-by-one in password callback buffer write
**File:** `OpenSSLEngine.cpp:48-53`

```cpp
int sz = static_cast<int>(passwd.size());
if (sz > size)          // should be >= size
{
    sz = size - 1;
}
strncpy(buf, passwd.c_str(), sz);
buf[sz] = '\0';         // OOB write when sz == size
```

When password length equals buffer size, `buf[sz]` writes one byte past the end.

**Fix:** Change `if (sz > size)` to `if (sz >= size)`.

### H1/H2. Shared: Tautological condition makes hex pair parsing dead code
**File:** `RFC2253.cpp:149` and `RFC2253.cpp:434`

```cpp
if (special.find(data[pos]) != string::npos || data[pos] != '\\' || data[pos] != '"')
```

`data[pos] != '\\' || data[pos] != '"'` is always true — a character cannot be both `\` and `"`. The `else` branch (`unescapeHex()` / `parseHexPair()`) never executes. This breaks DN matching when Distinguished Names contain hex-escaped values, affecting TrustManager accept/reject decisions.

**Fix:** Change `!=` to `==` in both locations:
```cpp
if (special.find(data[pos]) != string::npos || data[pos] == '\\' || data[pos] == '"')
```

### H3. SecureTransport: Private key loading loop always reads index 0
**File:** `SecureTransportUtil.cpp:484`

```cpp
for (CFIndex i = 0; i < count; ++i)
{
    auto itemRef = ... CFArrayGetValueAtIndex(items.get(), 0);  // should be 'i'
```

The loop iterates with `i` but always accesses index `0`. If the private key isn't at index 0, it will never be found.

**Fix:** Change `0` to `i`.

pepone:
## Summary
- Fix buffer overflow in `Ice_SSL_opensslPasswordCallback` in `OpenSSLEngine.cpp:48`
- When password length equals buffer size, `buf[sz]` writes one byte past the end
- Change guard from `sz > size` to `sz >= size`

Fixes #5096

## Test plan
- [ ] Run OpenSSL SSL tests
- [ ] Test with password exactly matching buffer size

### DIFF
--- a/cpp/src/Ice/SSL/OpenSSLEngine.cpp
+++ b/cpp/src/Ice/SSL/OpenSSLEngine.cpp
@@ -44,20 +44,18 @@ extern "C"
         auto* p = reinterpret_cast<OpenSSL::SSLEngine*>(userData);
         assert(p);
         string passwd = p->password();
-        int sz = static_cast<int>(passwd.size());
-        if (sz > size)
-        {
-            // Password too long for the buffer.
-            return -1;
-        }
-        memcpy(buf, passwd.c_str(), sz);
+
+        // Follow the OpenSSL documentation example: copy the password into the buffer, truncating if necessary, and
+        // null-terminate. See https://docs.openssl.org/3.0/man3/SSL_CTX_set_default_passwd_cb/#examples
+        strncpy(buf, passwd.c_str(), static_cast<size_t>(size));
+        buf[size - 1] = '\0';
 
         for (auto& character : passwd)
         {
             character = '\0';
         }
 
-        return sz;
+        return static_cast<int>(strlen(buf));
     }
 }
 


cpp/src/Ice/SSL/OpenSSLEngine.cpp
```diff
@@ -47,10 +47,10 @@ extern "C"
         int sz = static_cast<int>(passwd.size());
         if (sz > size)
         {
-            sz = size - 1;
+            // Password too long for the buffer.
+            return -1;
         }
-        strncpy(buf, passwd.c_str(), sz);
-        buf[sz] = '\0';
```
externl, line 53: Why don't we need the null termination?

pepone: We should probably keep it, this is not required in newer versions:

https://docs.openssl.org/3.0/man3/PEM_read_bio_PrivateKey/#examples

But it is required and document in older versions:

https://docs.openssl.org/1.1.1/man3/SSL_CTX_set_default_passwd_cb/#examples

cpp/src/Ice/SSL/OpenSSLEngine.cpp
```diff
@@ -47,10 +47,10 @@ extern "C"
         int sz = static_cast<int>(passwd.size());
         if (sz > size)
         {
-            sz = size - 1;
+            // Password too long for the buffer.
+            return -1;
```
InsertCreativityHere: We should probably document this somewhere, since before this chance, this function would always 'succeed' with a positive return value. Now it can return negative numbers to indicate failure.

cpp/src/Ice/SSL/OpenSSLEngine.cpp
InsertCreativityHere: It's not clear which of the two callbacks you posted links to, that this is supposed to be implementing. It seems to me, they have different expectations from one another, both in version 3.0 and 1.1.1.

Comparing the docs of each (as of version 3.0):

`read_bio_PrivateKey` states: `The callback must return the number of characters in the passphrase or -1 if an error occurred. The passphrase can be arbitrary data; in the case where it is a string, it is not NUL terminated.` Seems like this callback is only for supplying to `PEM_read_bio_PrivateKey(...)`. This matches the new implementation logic.

But `set_default_passwd_cb` states: `The password is considered to be a '\0' terminated string. If the password does not fit into the buffer, the password is truncated.` and `The actual length of the password must be returned to the calling function`. This matches our old implementation logic.

pepone: I think the null termination is expected here, reworked this.

externl (approved):

cpp/src/Ice/SSL/OpenSSLEngine.cpp
```diff
@@ -44,20 +44,18 @@ extern "C"
...
-        {
-            sz = size - 1;
-        }
-        strncpy(buf, passwd.c_str(), sz);
-        buf[sz] = '\0';
+
+        // Follow the OpenSSL documentation example: copy the password into the buffer, truncating if necessary, and
+        // null-terminate. See https://docs.openssl.org/3.0/man3/SSL_CTX_set_default_passwd_cb/#examples
+        strncpy(buf, passwd.c_str(), static_cast<size_t>(size));
+        buf[size - 1] = '\0';
```
InsertCreativityHere, line 51: So the idea is to now always write to the entirety of the buffer, padding with `\0` if necessary.

I guess there's no concern about `password()` returning a string with other `\0`' chars in it? We weren't handling this before, so fine to ignore I guess.

pepone: `\0` is not going to appear as part of a password in practice, this is just the C null terminator for strings, in case OpenSSL uses C APIs that expect a null terminated string, instead of relying just on the returned size.

PR #5108, opened 2026-02-16
Clear SSL password from memory after initialization

### LINKED ISSUE #5098: SSL: TLS hardening improvements across backends
## Summary

Security review of `cpp/src/Ice/SSL/` identified several opportunities to strengthen the TLS posture across all three backends. Some of these may be more appropriate for a minor release depending on compatibility considerations.

### OpenSSL: Password never scrubbed from `_password` member
**File:** `OpenSSLEngine.h:35`, `OpenSSLEngine.cpp:99`

The local copy in the password callback is scrubbed (lines 55-58), but `_password` persists for the engine's entire lifetime. The password is only needed during certificate/key loading in `initialize()`.

**Fix:** Clear at end of `initialize()`:
```cpp
_password.clear();
_password.shrink_to_fit();
```

### Schannel: No wildcard certificate support
**File:** `SchannelEngine.cpp:706-709`

DNS SAN matching uses exact `std::find()` comparison. Wildcard certificates (e.g., `*.example.com`) fail validation, potentially forcing administrators to disable hostname checking entirely.

**Fix:** Implement wildcard matching that checks if the SAN starts with `*.` and compares the remaining suffix.

### Schannel: CN matching uses substring search without verifying field start boundary
**File:** `SchannelEngine.cpp:696-703`

`dn.find(cn)` could match within another RDN value. The check verifies the match is followed by end-of-string or `,`, but does not verify it starts at a field boundary.

**Fix:** Also verify `pos == 0 || dn[pos - 1] == ','`.

### All backends: No minimum TLS version enforcement

None of the three backends explicitly set a TLS 1.2 floor:

- **OpenSSL:** Uses `TLS_method()` without calling `SSL_CTX_set_min_proto_version`
- **SecureTransport:** Does not call `SSLSetProtocolVersionMin`
- **Schannel:** Uses `SCH_USE_STRONG_CRYPTO` (disables SSL 3.0) but does not set `grbitEnabledProtocols`

All rely on OS/library defaults, which may permit TLS 1.0/1.1 on older systems.

### SecureTransport: `errno` set to OSStatus value
**File:** `SecureTransportTransceiverI.cpp:359,432,455`

```cpp
errno = err;
if (IceInternal::connectionLost())
```

OSStatus values occupy a different numeric space than POSIX errno. An OSStatus could coincidentally match `ECONNRESET` or `EPIPE`, causing misclassified errors.

**Fix:** Map OSStatus errors to appropriate POSIX equivalents, or use a separate error handling path.

### Schannel: SSPI flags equality check instead of bitwise AND
**File:** `SchannelTransceiverI.cpp:367`

```cpp
if (flags != _ctxFlags)
```

SSPI documentation states returned context flags should be verified using bitwise AND (to check that requested features are present), not strict equality. Returned flags may include additional bits.

**Fix:** Change to bitwise AND checks for each required flag.

pepone:
## Summary

Clear `_password` using `OPENSSL_cleanse` at the end of `OpenSSLEngine::initialize()` since it is only
needed during certificate/key loading. The local copy in the password callback was already scrubbed, but
the `_password` member persisted for the engine's entire lifetime.

Also moved the `_password` field from the base `SSLEngine` class to `OpenSSLEngine` since it is
OpenSSL-specific.

Fixes #5098

### DIFF
--- a/cpp/include/Ice/SSL/ClientAuthenticationOptions.h
+++ b/cpp/include/Ice/SSL/ClientAuthenticationOptions.h
@@ -32,8 +32,10 @@ namespace Ice::SSL
         ///
         /// This callback is invoked by the SSL transport for each new outgoing connection before starting the SSL
         /// handshake to determine the appropriate client credentials. The callback must return a `SCH_CREDENTIALS` that
-        /// represents the client's credentials. The SSL transport takes ownership of the credentials' `paCred` and
-        /// `hRootStore` members and releases them when the connection is closed.
+        /// represents the client's credentials. The SSL transport takes ownership of the credentials' `paCred`
+        /// member and releases it when the connection is closed. The `hRootStore` from the returned credentials is not
+        /// used for certificate validation; use `trustedRootCertificates` or `serverCertificateValidationCallback`
+        /// instead.
         ///
         /// @param host The target host name.
         /// @return The client SSL credentials.
--- a/cpp/include/Ice/SSL/ServerAuthenticationOptions.h
+++ b/cpp/include/Ice/SSL/ServerAuthenticationOptions.h
@@ -33,8 +33,10 @@ namespace Ice::SSL
         ///
         /// This callback is invoked by the SSL transport for each new incoming connection before starting the SSL
         /// handshake to determine the appropriate server credentials. The callback must return a `SCH_CREDENTIALS` that
-        /// represents the server's credentials. The SSL transport takes ownership of the credentials' `paCred` and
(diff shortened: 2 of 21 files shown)


externl (approved):

InsertCreativityHere (approved):
Sure, looks good to me.

We also have a `_password` field in `Ice::SSL::SSLEngine` which seems dead, but doesn't get this `OPENSSL_cleanse` treatment. (this PR only touches `Ice:SSL::OpenSSL::SSLEngine`)

PR #5109, opened 2026-02-16
Improve error handling in SSL utility functions

### LINKED ISSUE #5097: SSL: Add missing null checks, return value checks, and variable initialization
## Summary

Security review of `cpp/src/Ice/SSL/` identified several locations where missing null checks, unchecked return values, or uninitialized variables could lead to crashes or silent misbehavior. These are defensive coding improvements suitable for a patch release.

### SecureTransport: `_tflags` not initialized
**File:** `SecureTransportTransceiverI.h:75`

```cpp
mutable std::uint8_t _tflags;
```

Never initialized in either constructor. Controls SSL I/O read/write state flags (`SSLWantRead`, `SSLWantWrite`). Reading an uninitialized value is undefined behavior.

**Fix:** Add in-class initializer: `mutable std::uint8_t _tflags{0};`

### Shared: SSLEngine member variables not initialized
**File:** `SSLEngine.h:63-66`

```cpp
bool _checkCertName;
int _verifyPeer;
int _securityTraceLevel;
```

Only set in `initialize()`. If read before initialization, values are undefined. Compare with `_revocationCheckCacheOnly{false}` which does have an in-class initializer.

**Fix:** Add in-class initializers: `bool _checkCertName{false}; int _verifyPeer{0}; int _securityTraceLevel{0};`

### OpenSSL: `SSL_set_ex_data` return not checked
**File:** `OpenSSLTransceiverI.cpp:93`

```cpp
SSL_set_ex_data(_ssl, 0, this);
```

If this fails, the verify callback retrieves null via `SSL_get_ex_data` and dereferences it.

### OpenSSL: `SSL_CTX_set_default_verify_paths` return not checked
**File:** `OpenSSLEngine.cpp:157`

When `IceSSL.UsePlatformCAs` is enabled but `IceSSL.CAs` is not set, failure here means no trusted CAs are loaded. Peer verification silently breaks with unclear error messages.

### OpenSSL: `defaultSSLContextSelectionCallback` missing null check on `SSL_CTX_new`
**File:** `OpenSSLTransceiverI.cpp:44-49`

`SSL_CTX_new(TLS_method())` can return null. The returned pointer is used immediately without checking.

### OpenSSL: Missing `ERR_clear_error()` before handshake
**File:** `OpenSSLTransceiverI.cpp:111`

The `read()` and `write()` methods correctly call `ERR_clear_error()`, but the handshake loop does not. Stale errors from context setup cause misleading error messages.

### OpenSSL: `BIO_new()` return not null-checked
**File:** `SSLUtil.cpp:473,621`

`BIO_new(BIO_s_mem())` can return null. Passing null to `X509_NAME_print_ex` or `PEM_write_bio_X509` would crash.

### SecureTransport: `assert(trust)` for security-critical null check
**File:** `SecureTransportEngine.cpp:753`

`assert` is compiled out in release builds. If `trust` is null, `SecTrustSetNetworkFetchAllowed(trust, false)` is undefined behavior.

**Fix:** Replace with a runtime check that throws `SecurityException`.

### Shared: `readFile()` doesn't check `tellg()` for -1
**File:** `SSLUtil.cpp:80-100`

```cpp
buffer.resize(static_cast<size_t>(is.tellg()));
```

`tellg()` returns -1 on error. Casting to `size_t` produces an enormous value, causing `bad_alloc` rather than a clean error.

pepone:
## Summary
- Check `tellg()` return for -1 in `readFile()` to produce a clean error instead of a huge allocation attempt
- Reject odd-length hex strings in `parseBytes()` instead of silently dropping the last nibble

Fixes #5097

## Test plan
- [ ] Run SSL tests across all platforms
- [ ] Verify certificate loading still works correctly

### DIFF
--- a/cpp/src/Ice/SSL/SSLUtil.h
+++ b/cpp/src/Ice/SSL/SSLUtil.h
@@ -46,8 +46,8 @@ namespace Ice::SSL
     // Determine if a directory exists, with an optional parent directory.
     std::optional<std::string> resolveDirPath(const std::string& path, const std::string& parentDir = "");
 
-    // Parse a hex string (e.g., "AB:CD:EF" or "ABCDEF") into a byte buffer. Spaces and colons are ignored.
-    // Returns false if the string contains invalid characters or has an odd number of hex digits.
+    /// Parse a hex string (e.g., "AB:CD:EF" or "ABCDEF") into a byte buffer. Spaces and colons are ignored.
+    /// @return `false` if the string contains invalid characters or has an odd number of hex digits.
     bool parseBytes(const std::string&, std::vector<unsigned char>&);
 
 #if defined(ICE_USE_SCHANNEL)


cpp/src/Ice/SSL/SSLUtil.cpp
```diff
@@ -86,7 +92,12 @@ Ice::SSL::readFile(const string& file, vector<char>& buffer)
     }
 
     is.seekg(0, is.end);
-    buffer.resize(static_cast<size_t>(is.tellg()));
+    streampos size = is.tellg();
+    if (size == streampos(-1))
+    {
+        throw CertificateReadException(__FILE__, __LINE__, "error determining file size " + file);
```
externl:
```suggestion
        throw CertificateReadException(__FILE__, __LINE__, "error determining file size: " + file);
```

cpp/src/Ice/SSL/SSLUtil.cpp
```diff
@@ -86,7 +92,12 @@ Ice::SSL::readFile(const string& file, vector<char>& buffer)
     }
 
     is.seekg(0, is.end);
-    buffer.resize(static_cast<size_t>(is.tellg()));
+    streampos size = is.tellg();
+    if (size == streampos(-1))
```
externl: Should this be?

```suggestion
    if (size == streampos{-1})
```

externl (approved):

cpp/src/Ice/SSL/SSLUtil.h
```diff
@@ -46,6 +46,8 @@ namespace Ice::SSL
     // Determine if a directory exists, with an optional parent directory.
     std::optional<std::string> resolveDirPath(const std::string& path, const std::string& parentDir = "");
 
+    // Parse a hex string (e.g., "AB:CD:EF" or "ABCDEF") into a byte buffer. Spaces and colons are ignored.
+    // Returns false if the string contains invalid characters or has an odd number of hex digits.
```
InsertCreativityHere: I say might as well make it a doc-comment.
```suggestion
    /// Parse a hex string (e.g., "AB:CD:EF" or "ABCDEF") into a byte buffer. Spaces and colons are ignored.
    /// @return `false` if the string contains invalid characters or has an odd number of hex digits.
```

PR #5110, opened 2026-02-16
Fix typos and remove unused includes in SSL code

### LINKED ISSUE #5099: SSL: Code quality cleanups (typos, unused includes, minor fixes)
## Summary

Security review of `cpp/src/Ice/SSL/` identified several low-severity code quality issues. These are safe, low-risk cleanups.

### `AltNAmeIP` should be `AltNameIP`
**File:** `SSLUtil.h:37`

```cpp
const int AltNAmeIP = 7;
```

Inconsistent with `AltNameDNS`, `AltNameEmail`, `AltNameDirectory`, `AltNameURL`. Also referenced in `SSLUtil.cpp:348` and `SchannelEngine.cpp:671`.

### `"unknow"` should be `"unknown"`
**File:** `SchannelEngine.cpp:972`

```cpp
os << "SSL transport: error unknow key algorithm: '" << keyInfo->Algorithm.pszObjId << "'";
```

### OpenSSL: Unused `#include <iostream>`
**File:** `OpenSSLTransceiverI.cpp:23`

### Schannel: Unused `#include <iostream>`
**File:** `SchannelEngine.cpp:17`

### SecureTransport: Missing `[[fallthrough]]` in `escapeX509Name`
**File:** `SecureTransportUtil.cpp:70-95`

Switch cases for special characters intentionally fall through to `default`. Without `[[fallthrough]]`, a future maintainer could "fix" the apparent bug by adding a `break`, which would break DN escaping used by TrustManager.

### Shared: `ScopedCertificate` classes lack deleted copy constructors
**File:** `SSLUtil.h:56-99`

All three backend-specific `ScopedCertificate` RAII wrappers allow copying, which would result in double-free of the underlying certificate resource.

**Fix:** Add `= delete` for copy constructor and copy assignment operator.

### Shared: `const_cast` to set const members in `SSLEngine::initialize()`
**File:** `SSLEngine.cpp:62-64`

```cpp
const_cast<bool&>(_revocationCheckCacheOnly) = ...;
const_cast<int&>(_revocationCheck) = ...;
```

Technically undefined behavior. Either remove `const` from these members or initialize them in the constructor.

### Shared: `parseBytes()` silently drops trailing nibble on odd-length hex
**File:** `SSLUtil.cpp:46-77`

An odd-length hex string silently loses its last character. Used for certificate fingerprint parsing.

**Fix:** Return `false` if the cleaned hex string has an odd length.

### SecureTransport: Naive `.pem` extension check
**File:** `SecureTransportUtil.cpp:576`

```cpp
if (file.find(".pem") != string::npos)
```

Matches `file.pem.bak`, `my.pemthing`, etc. Should check that the filename ends with `.pem`.

### Schannel: `CryptReleaseContext` never called for acquired `cryptProv`
**File:** `SchannelEngine.cpp:1015-1034`

After acquiring the cryptographic context and importing the key, the `cryptProv` handle is never released. Resource leak.

### Schannel: `ISC_RET_EXTENDED_ERROR` used in request flags
**File:** `SchannelTransceiverI.cpp:119`

`ISC_RET_EXTENDED_ERROR` (a return flag) is used where `ISC_REQ_EXTENDED_ERROR` (a request flag) is expected. Same numeric value, but semantically incorrect.

pepone:
## Summary
- Rename `AltNAmeIP` to `AltNameIP` to match naming convention (`AltNameDNS`, `AltNameEmail`, etc.)
- Fix `"unknow"` typo to `"unknown"` in `SchannelEngine.cpp` error message
- Remove unused `#include <iostream>` from `OpenSSLTransceiverI.cpp` and `SchannelEngine.cpp`

Fixes #5099

## Test plan
- [ ] Build on all platforms to verify no compilation errors from the rename

### DIFF
--- a/cpp/src/Ice/SSL/SchannelEngine.cpp
+++ b/cpp/src/Ice/SSL/SchannelEngine.cpp
@@ -968,7 +968,7 @@ Schannel::SSLEngine::initialize()
                     if (strcmp(keyInfo->Algorithm.pszObjId, szOID_RSA_RSA))
                     {
                         ostringstream os;
-                        os << "SSL transport: error unknown key algorithm: '" << keyInfo->Algorithm.pszObjId << "'";
+                        os << "SSL transport: unknown key algorithm: '" << keyInfo->Algorithm.pszObjId << "'";
                         throw InitializationException(__FILE__, __LINE__, os.str());
                     }
 


cpp/src/Ice/SSL/SchannelEngine.cpp
```diff
@@ -969,7 +968,7 @@ Schannel::SSLEngine::initialize()
                     if (strcmp(keyInfo->Algorithm.pszObjId, szOID_RSA_RSA))
                     {
                         ostringstream os;
-                        os << "SSL transport: error unknow key algorithm: '" << keyInfo->Algorithm.pszObjId << "'";
+                        os << "SSL transport: error unknown key algorithm: '" << keyInfo->Algorithm.pszObjId << "'";
```
InsertCreativityHere: Judging from the other error messages in this file, it probably shouldn't have `error` in front.
Also, I'm pretty sure that the message as written is grammatically incorrect. So:
```suggestion
                        os << "SSL transport: unknown key algorithm: '" << keyInfo->Algorithm.pszObjId << "'";
```

PR #5111, opened 2026-02-16
Delete copy and move operations on ScopedCertificate wrappers

### LINKED ISSUE #5099: SSL: Code quality cleanups (typos, unused includes, minor fixes)

pepone:
## Summary
- Delete copy constructor, copy assignment, move constructor, and move assignment on all three `ScopedCertificate` RAII wrappers to prevent potential double-free

Fixes #5099

## Test plan
- [ ] Build on all platforms
- [ ] Run SSL tests to verify no behavioral changes

### DIFF
--- a/cpp/src/Ice/SSL/SSLEngine.cpp
+++ b/cpp/src/Ice/SSL/SSLEngine.cpp
@@ -59,8 +59,9 @@ Ice::SSL::SSLEngine::initialize()
     _securityTraceLevel = properties->getIcePropertyAsInt("IceSSL.Trace.Security");
     _securityTraceCategory = "Security";
 
-    _revocationCheckCacheOnly = properties->getIcePropertyAsInt("IceSSL.RevocationCheckCacheOnly") > 0;
-    _revocationCheck = properties->getIcePropertyAsInt("IceSSL.RevocationCheck");
+    const_cast<bool&>(_revocationCheckCacheOnly) =
+        properties->getIcePropertyAsInt("IceSSL.RevocationCheckCacheOnly") > 0;
+    const_cast<int&>(_revocationCheck) = properties->getIcePropertyAsInt("IceSSL.RevocationCheck");
 }
 
 void
--- a/cpp/src/Ice/SSL/SSLEngine.h
+++ b/cpp/src/Ice/SSL/SSLEngine.h
@@ -64,8 +64,8 @@ namespace Ice::SSL
         int _verifyPeer;
         int _securityTraceLevel;
         std::string _securityTraceCategory;
-        bool _revocationCheckCacheOnly{false};
-        int _revocationCheck{0};
+        const bool _revocationCheckCacheOnly{false};
+        const int _revocationCheck{0};
     };
 }
 


cpp/src/Ice/SSL/SSLEngine.cpp
```diff
@@ -59,9 +59,8 @@ Ice::SSL::SSLEngine::initialize()
     _securityTraceLevel = properties->getIcePropertyAsInt("IceSSL.Trace.Security");
     _securityTraceCategory = "Security";
 
-    const_cast<bool&>(_revocationCheckCacheOnly) =
```
externl, line 62: This is a pattern that we use all over the place. I get that it's not a great practice, but the const data member is what we want.

pepone: Reverted this part.

PR #5117, opened 2026-02-17
Fix C# and Java TrustManager to reject connections on DN parse failure

pepone:
## Summary

Fixed the TrustManager in C# and Java to properly handle the rare situation where Ice's strict RFC 2253
parser cannot parse a certificate's subject DN string produced by the platform's SSL stack
(`X500Principal.getName(RFC2253)` in Java, `X500DistinguishedName.Name` in C#).

Previously, the `ParseException` was caught in a block that wrapped the entire accept/reject evaluation
logic. After catching, execution fell through to `return accept.isEmpty()`. If only reject rules were
configured (no accept rules), the connection was silently accepted — effectively bypassing the
`IceSSL.TrustOnly` reject rules.

Note that the TrustManager runs after standard SSL/TLS certificate verification has already succeeded, so
the certificate is already trusted by the platform. This bug could only be triggered by a discrepancy
between the platform's DN serialization and Ice's strict RFC 2253 parser, not by a malformed certificate.

Fix: catch the `ParseException` immediately after parsing the DN and return `false` (reject), rather than
letting execution fall through.

### DIFF
--- a/csharp/src/Ice/SSL/TrustManager.cs
+++ b/csharp/src/Ice/SSL/TrustManager.cs
@@ -118,69 +118,70 @@ internal sealed class TrustManager
             X500DistinguishedName subjectDN = info.certs[0].SubjectName;
             string subjectName = subjectDN.Name;
             Debug.Assert(subjectName != null);
+            // Decompose the subject DN into the RDNs.
+            if (_traceLevel > 0)
+            {
+                if (info.incoming)
+                {
+                    _communicator.getLogger().trace("Security", "trust manager evaluating client:\n" +
+                        "subject = " + subjectName + "\n" + "adapter = " + info.adapterName + "\n" + description);
+                }
+                else
+                {
+                    _communicator.getLogger().trace("Security", "trust manager evaluating server:\n" +
+                        "subject = " + subjectName + "\n" + description);
+                }
+            }
+
+            List<RFC2253.RDNPair> dn;
             try
             {
-                // Decompose the subject DN into the RDNs.
-                if (_traceLevel > 0)
-                {
-                    if (info.incoming)
-                    {
-                        _communicator.getLogger().trace("Security", "trust manager evaluating client:\n" +
-                            "subject = " + subjectName + "\n" + "adapter = " + info.adapterName + "\n" + description);
-                    }
-                    else
-                    {
-                        _communicator.getLogger().trace("Security", "trust manager evaluating server:\n" +
-                            "subject = " + subjectName + "\n" + description);
-                    }
-                }
-
-                List<RFC2253.RDNPair> dn = RFC2253.parseStrict(subjectName);
-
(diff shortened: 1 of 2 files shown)


csharp/src/Ice/SSL/TrustManager.cs
externl, line 141: We should move this check to be just around `RFC2253.parseStrict`.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/TrustManager.java
```diff
@@ -157,6 +157,7 @@ boolean verify(ConnectionInfo info, String desc) {
             } catch (ParseException ex) {
                 String m = "Ice.SSL: unable to parse certificate DN `" + subjectName + "'\nreason: " + ex.getMessage();
                 _communicator.getLogger().warning(m);
+                return false;
```
externl: Same here.

csharp/src/Ice/SSL/TrustManager.cs
pepone: do you mean the try/catch?

externl: Yep!

pepone: Fixed

externl:
> > Don't we want to catch this exception in C++ and return false like in C# and Java? Otherwise should we just let the exception propagate everywhere?
> 
> Yes the C++ bug is that the connection is terminated with the ParseException. Java/C# bug is that we accepted the connection if there was a exception. We can include the C++ fix in this PR even if it is kind of a different bug.

A separate PR is fine. Just wanted to make sure I understood the issue correctly.

pepone:
See https://github.com/zeroc-ice/ice/issues/5134

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/TrustManager.java
```diff
@@ -101,62 +101,70 @@ boolean verify(ConnectionInfo info, String desc) {
...
+            try {
+                dn = RFC2253.parseStrict(subjectName);
+            } catch (ParseException ex) {
+                _communicator
+                    .getLogger()
+                    .warning(
+                        "Ice.SSL: unable to parse certificate DN `"
+                            + subjectName
+                            + "'\nreason: "
+                            + ex.getMessage());
```
InsertCreativityHere, line 142: Is clang-format okay with this? Seems to fit in under 120 columns.
```suggestion
                    .warning(
                        "Ice.SSL: unable to parse certificate DN `" + subjectName + "'\nreason: " + ex.getMessage());
```

InsertCreativityHere (approved):
Looks good to me.

Issue #5114, opened 2026-02-16
Java/C# SSL: Defensive coding and code quality improvements


pepone:
## Summary

Security review of the Java and C# SSL implementations found several medium and low severity issues suitable for a patch release.

### M1. C#: `unescapeHex` off-by-one bounds check
**File:** `csharp/.../SSL/RFC2253.cs:170`

`pos + 2 >= data.Length` rejects valid hex pairs at end of string. Should be `pos + 2 > data.Length`. Same bug as C++.

### M2. Java: No minimum TLS version enforcement
**File:** `java/.../SSL/SSLEngine.java:277`

`SSLContext.getInstance("TLS")` supports all TLS versions. Using `"TLSv1.2"` would be more explicit.

### M3. C#: `isAbsolutePath` potential `IndexOutOfRangeException`
**File:** `csharp/.../SSL/SSLEngine.cs:323`

No length guard on non-Windows before accessing `path[1]`.

### M4. C#: `_cipher` possibly null before authentication
**File:** `csharp/.../SSL/TransceiverI.cs:510`

### M5. Java: `BUFFER_OVERFLOW` handled with assert
**File:** `java/.../SSL/TransceiverI.java:190`

`assert` is disabled at runtime with `-da`. Should throw a meaningful exception.

### L6. Java: `SSLEngineFactory` Javadoc incorrectly documents `peerPort` (`SSLEngineFactory.java:14`)

PR #5119, opened 2026-02-17
Add IceSSL.VerifyPeer range validation in C# and Java

pepone:
## Summary

Add `IceSSL.VerifyPeer` range validation (0-2) in C# and Java, throwing an `InitializationException`
for out-of-range values. This matches the existing C++ behavior.

Also includes minor defensive fixes:
- C# `SSLEngine.isAbsolutePath()`: add length guard to prevent `IndexOutOfRangeException` on empty paths
  on non-Windows platforms.
- C# `TransceiverI`: initialize `_cipher` to empty string to prevent `NullReferenceException` if
  `getInfo()` is called before authentication completes.
- Java `TransceiverI`: replace `assert` with runtime `SecurityException` for `BUFFER_OVERFLOW` status,
  since asserts may be disabled with `-da`.

Relates to #5114

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/SSLEngine.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/SSLEngine.java
@@ -62,6 +62,10 @@ public class SSLEngine {
 
         // VerifyPeer determines whether certificate validation failures abort a connection.
         _verifyPeer = properties.getIcePropertyAsInt("IceSSL.VerifyPeer");
+        if (_verifyPeer < 0 || _verifyPeer > 2) {
+            throw new com.zeroc.Ice.InitializationException(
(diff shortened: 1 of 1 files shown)


csharp/src/Ice/SSL/SSLEngine.cs
```diff
@@ -29,6 +29,10 @@ internal void initialize()
         _defaultDir = properties.getIceProperty("IceSSL.DefaultDir");
 
         _verifyPeer = properties.getIcePropertyAsInt("IceSSL.VerifyPeer");
+        if (_verifyPeer < 0 || _verifyPeer > 2)
```
externl, line 32: We should make sure to have this check in other languages.

pepone: Added the same validation in Java. C++ already had it.

PR #5120, opened 2026-02-17
Java/C# SSL code quality improvements

pepone:
## Summary
- **Java SSLEngine**: Use consistent `!= null` check for key password cleanup (was `length > 0`)
- **Java EndpointI**: Use `.equals()` instead of `==` for String comparison in `connectionId()`
- **Java ConnectionInfo**: Defensive copy of `certs` array to prevent external mutation
- **Java SSLEngineFactory**: Fix incorrect Javadoc claiming `peerPort` is always `-1` for outgoing connections
- **C# TransceiverI**: Use `&= ~` instead of `^=` for flag clearing to avoid toggling if bit is not set

Relates to #5114

## Test plan
- [ ] Verify Java and C# SSL tests pass
- [ ] Verify no behavioral regressions in endpoint connection ID comparison

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/EndpointI.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/EndpointI.java
@@ -62,7 +62,6 @@ final class EndpointI extends com.zeroc.Ice.EndpointI {
 
     @Override
     public com.zeroc.Ice.EndpointI connectionId(String connectionId) {
-        assert connectionId != null;
         if (connectionId.equals(_delegate.connectionId())) {
             return this;
         } else {


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SSL/EndpointI.java
```diff
@@ -62,7 +62,8 @@ public String connectionId() {
 
     @Override
     public com.zeroc.Ice.EndpointI connectionId(String connectionId) {
-        if (connectionId == _delegate.connectionId()) {
+        assert connectionId != null;
```
InsertCreativityHere: Honestly I wouldn't even bother with this assert, we don't have it in any of the other `connectionId` methods, and asserts are disabled by default in Java anyways. If a user tries passing in a `null` string to any of the `connectionId`s, they won't be surprised when they get an exception from doing so.

pepone: Removed.

Issue #5126, opened 2026-02-17
`RegSetValueExW` size computed from narrow string length — buffer overread for non-ASCII paths


pepone:
## `RegSetValueExW` Size Computed from Narrow String Length

**File:** `cpp/src/iceserviceinstall/ServiceInstaller.cpp` lines 807–813
**Severity:** Medium — buffer overread for non-ASCII paths

### Description

The data passed to `RegSetValueExW` is a wide string (`stringToWstring(resourceFile).c_str()`), but the size parameter uses the narrow string's `length()`:

```cpp
res = RegSetValueExW(
    key,
    L"EventMessageFile",
    0,
    REG_EXPAND_SZ,
    reinterpret_cast<const BYTE*>(stringToWstring(resourceFile).c_str()),
    static_cast<DWORD>(resourceFile.length() + 1) * sizeof(wchar_t));
```

For ASCII-only paths this works by coincidence since the byte count equals the character count. But for multi-byte UTF-8 characters (e.g., accented characters, CJK in paths), `resourceFile.length()` returns the UTF-8 byte count which is larger than the wide character count. This causes `RegSetValueExW` to read past the end of the wide string buffer.

### Fix

Store the wide string and use its length for the size parameter:

```cpp
wstring wResourceFile = stringToWstring(resourceFile);
res = RegSetValueExW(
    key,
    L"EventMessageFile",
    0,
    REG_EXPAND_SZ,
    reinterpret_cast<const BYTE*>(wResourceFile.c_str()),
    static_cast<DWORD>((wResourceFile.length() + 1) * sizeof(wchar_t)));
```

Issue #5125, opened 2026-02-17
Incorrect `assert` in `getServiceInstallerPath` checks truthiness instead of `!= npos`


pepone:
## Incorrect `assert` Checks Truthiness Instead of `!= npos`

**File:** `cpp/src/iceserviceinstall/ServiceInstaller.cpp` line 450
**Severity:** Low — the assert is logically wrong but cannot misfire in practice

### Description

`assert(path.find_last_of("/\\"))` tests whether the return value is non-zero (truthy). This doesn't express the intended check:

1. Path `"\foo"` → `find_last_of` returns `0` → falsy → assert fires incorrectly.
2. No separator found → `string::npos` → truthy → assert passes when it should fail.

```cpp
string path = wstringToString(buffer);
assert(path.find_last_of("/\\"));
return path.substr(0, path.find_last_of("/\\"));
```

### Analysis

In practice, the impact is negligible. The `path` comes from `GetModuleFileNameW`, which always returns a fully-qualified Windows path like `C:\Program Files\...\something.exe`. This means:

- There will always be at least one separator, so the `npos` case cannot happen.
- The last separator will never be at position 0 since Windows paths start with a drive letter (`C:\`), so the false-trigger-at-zero case cannot happen either.

Still, the assert doesn't express its actual intent and would behave incorrectly on unexpected input.

### Fix

```cpp
assert(path.find_last_of("/\\") != string::npos);
```

Issue #5123, opened 2026-02-17
Operator precedence bug: `!_configFile.find("HKLM\\") == 0` inverts the starts-with test


pepone:
## Operator Precedence Bug — `!x.find(...) == 0` Inverts the "Starts With" Test

**File:** `cpp/src/iceserviceinstall/ServiceInstaller.cpp` lines 193 and 248
**Severity:** High — logic error

### Description

Due to C++ operator precedence, `!_configFile.find("HKLM\\") == 0` is parsed as `(!_configFile.find("HKLM\\")) == 0`. Unary `!` binds more tightly than `==`.

- When `_configFile` **starts with** `"HKLM\"`: `find` returns `0`, `!0` is `true` (1), `1 == 0` is `false`. Body is **skipped**.
- When it does **not** start with `"HKLM\"`: `find` returns non-zero, `!nonzero` is `false` (0), `0 == 0` is `true`. Body **executes**.

The condition is always the **opposite** of intended.

### Impact

**Line 193:** `grantPermissions` is called on `HKLM\` registry paths as if they were file paths, passing registry paths to Windows file security APIs. Conversely, actual file paths that *should* get permissions granted are skipped.

```cpp
if (!_configFile.find("HKLM\\") == 0)   // BUG: always inverted
{
    grantPermissions(_configFile);
}
```

**Line 248:** `GetFullPathName` is called on `HKLM\` registry paths, and actual file paths fall through to the `else` branch that treats them as literal strings without resolving to a full path.

```cpp
if (!_configFile.find("HKLM\\") == 0)   // BUG: always inverted
{
    char fullPath[MAX_PATH];
    if (GetFullPathName(_configFile.c_str(), MAX_PATH, fullPath, 0) > MAX_PATH)
    {
        throw runtime_error("Could not compute the full path of " + _configFile);
    }
    command += string(fullPath) + "\"";
}
else
{
    command += _configFile + "\"";
}
```

### Fix

Replace `!x.find(...) == 0` with `x.find(...) != 0` at both locations:

```cpp
if (_configFile.find("HKLM\\") != 0)
```

PR #5127, opened 2026-02-17
Fix multiple bugs in iceserviceinstall ServiceInstaller

pepone:
## Summary

- Fix operator precedence bug where `!_configFile.find("HKLM\\") == 0` was parsed as `(!_configFile.find("HKLM\\")) == 0`, inverting the starts-with test at two locations (#5123)
- Fix `assert` in `getServiceInstallerPath` to check `!= string::npos` instead of truthiness (#5125)
- Fix `RegSetValueExW` buffer overread for non-ASCII paths by computing size from the wide string length instead of the narrow string length (#5126)
- Remove dead `fileExists` declaration from `ServiceInstaller.h` (never defined)

## Test plan

- [ ] Verify iceserviceinstall builds on Windows
- [ ] Test service installation with file-based config (non-HKLM path) — `grantPermissions` and `GetFullPathName` should now be called correctly
- [ ] Test service installation with HKLM registry config path — should skip file operations
- [ ] Test with non-ASCII characters in resource file path to verify `RegSetValueExW` uses correct buffer size

### DIFF
--- a/cpp/src/iceserviceinstall/ServiceInstaller.cpp
+++ b/cpp/src/iceserviceinstall/ServiceInstaller.cpp
@@ -190,7 +190,7 @@ IceServiceInstaller::install(const PropertiesPtr& properties)
         }
     }
 
-    if (!_configFile.find("HKLM\\") == 0)
+    if (_configFile.find("HKLM\\") != 0)
     {
         grantPermissions(_configFile);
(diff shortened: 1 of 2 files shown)


InsertCreativityHere (approved):
Looks good to me.

PR #5139, opened 2026-02-18
Fix ConnectorI ordering and EndpointI hash in IceBT

### LINKED ISSUE #5130: Fix IceBT ConnectorI::operator< and EndpointI::hash() correctness bugs
### 1. `ConnectorI::operator<` has broken strict weak ordering — undefined behavior

**File:** `cpp/src/IceBT/ConnectorI.cpp` lines 82–85

The `_addr` comparison is missing the reverse-direction `else if` branch. After checking `_addr < p->_addr`, the code falls through to `_uuid` even when `_addr > p->_addr`, which can cause both `a < b` and `b < a` to be true simultaneously, violating strict weak ordering.

```cpp
if (_addr < p->_addr)
{
    return true;
}
// BUG: Missing reverse check. Falls through to _uuid when _addr > p->_addr.

if (_uuid < p->_uuid)
{
    return true;
}
```

Example: connector A has `_addr="XX"`, `_uuid="aa"` and connector B has `_addr="AA"`, `_uuid="zz"`. Then `A < B` is false (falls through to uuid, `"aa" < "zz"` = true), but `B < A` is also false (falls through to uuid, `"zz" < "aa"` = false)... actually wait, let me re-trace: `A.operator<(B)`: `"XX" < "AA"` = false, fall through, `"aa" < "zz"` = true → returns true. `B.operator<(A)`: `"AA" < "XX"` = true → returns true. Both `A < B` and `B < A` are true — undefined behavior in STL ordered containers.

**Fix:** Add the missing `else if` branch:

```cpp
if (_addr < p->_addr)
{
    return true;
}
else if (p->_addr < _addr)
{
    return false;
}
```

### 2. `_channel` missing from `EndpointI::hash()` but present in `operator==`

**File:** `cpp/src/IceBT/EndpointI.cpp` lines 490–500 vs. 253–257

`hash()` omits `_channel` but `operator==` compares it. This violates the invariant that equal objects must produce equal hashes. Two endpoints differing only in `_channel` will hash the same but compare as not-equal.

**Fix:** Add `IceInternal::hashAdd(h, _channel);` to the hash computation.

pepone:
## Summary
- Fix broken strict weak ordering in `ConnectorI::operator<` — the `_addr` comparison was missing the reverse-direction `else if` branch, causing it to fall through to `_uuid` when `_addr > p->_addr`. This could result in both `A < B` and `B < A` being true, which is undefined behavior in STL ordered containers.
- Add missing `_channel` to `EndpointI::hash()` to match `operator==`, fixing the invariant that equal objects must produce equal hashes.

Fixes #5130

## Test plan
- [ ] CI passes (Linux, since IceBT is Linux-only)

### DIFF
--- a/cpp/src/IceBT/EndpointI.cpp
+++ b/cpp/src/IceBT/EndpointI.cpp
@@ -493,7 +493,6 @@ IceBT::EndpointI::hash() const noexcept
     size_t h = 5381;
     IceInternal::hashAdd(h, _addr);
     IceInternal::hashAdd(h, _uuid);
-    IceInternal::hashAdd(h, _channel);
     IceInternal::hashAdd(h, _timeout);
     IceInternal::hashAdd(h, _connectionId);
     IceInternal::hashAdd(h, _compress);


cpp/src/IceBT/EndpointI.cpp
```diff
@@ -493,6 +493,7 @@ IceBT::EndpointI::hash() const noexcept
     size_t h = 5381;
     IceInternal::hashAdd(h, _addr);
     IceInternal::hashAdd(h, _uuid);
+    IceInternal::hashAdd(h, _channel);
```
InsertCreativityHere: We also aren't hashing the `_name` field here, which is interesting, since the two fields we're missing (`_channel` and `_name`) are also the two fields we don't marshal. Maybe this is intentional and we don't want to hash fields which aren't marshaled? What do you think?

Also, just for the record, there is no bug here actually. The invariant that `x1 == x2` => `x1.hash() == x2.hash()` is totally upheld. It's okay for `==` to be more strict than the hash is; the hashes will still be equal.

pepone: You are right that this is not a bug, but seems an oversight. No comment as for why is used in `operator==` and `operator<`.

The `_name` is missing on all three which seems on purpose. That say adding channel here wont make big difference either way.

bernardnormier: Not including all fields in the hash is definitely fine. We typically do it to keep the hash function reasonably fast.

bernardnormier: After looking at the code more closely, -c (_channel) is an OA endpoint -only option that we should not include in this hash.

pepone: is it correct to include it in equals?

bernardnormier: Yes, everything should be included in `operator==`.

Note that for coloc-optimization detection, we use `equivalent`, not `operator==`. `equivalent` doesn't use channel.

externl (approved):

PR #5140, opened 2026-02-18
Fix multiple bugs in IceBT DBus layer

### LINKED ISSUE #5129: Fix multiple bugs in IceBT DBus layer
### 1. `ExceptionI` shadows base class `reason` member — error messages lost

**File:** `cpp/src/IceBT/DBus.cpp` lines 29–45

`ExceptionI` inherits from `IceBT::DBus::Exception` (which has `std::string reason` at `DBus.h:21`) and declares its own `string reason` at line 38, shadowing the base class member.

- `ExceptionI(const string& s)` calls `Exception(s)`, setting the **base** `reason`.
- `ExceptionI(const ErrorWrapper& w)` and `ExceptionI(const DBusError& err)` call `init()`, which sets the **derived** `reason`. The base remains empty.

All callers catch `const DBus::Exception& ex` and read `ex.reason` (the base member). Any exception from a `DBusError` has an empty `reason` from the caller's perspective.

**Fix:** Remove the shadowing member and set the base `reason` directly in `init()`.

### 2. DBus boolean read uses `bool` (1 byte) instead of `dbus_bool_t` (4 bytes) — stack corruption

**File:** `cpp/src/IceBT/DBus.cpp` lines 593–598

`dbus_message_iter_get_basic()` writes a `dbus_bool_t` (4 bytes) but the code passes the address of a C++ `bool` (1 byte), overwriting 3 bytes beyond the variable on the stack. Note the write path correctly uses `dbus_bool_t`.

```cpp
case Type::KindBoolean:
{
    bool v;  // 1 byte
    ::dbus_message_iter_get_basic(_iter, &v);  // writes 4 bytes
    return make_shared<BooleanValue>(v);
}
```

**Fix:** Use `dbus_bool_t v;` and convert with `v != 0`.

### 3. `VariantValue::clone()` returns alias instead of deep copy

**File:** `cpp/src/IceBT/DBus.h` line 216

Every other `Value` subclass (`DictEntryValue`, `ArrayValue`, `StructValue`) creates a new independent copy in `clone()`. `VariantValue::clone()` returns `shared_from_this()`, meaning "cloned" variants are aliased.

In practice, impact is limited since DBus values are typically read-only after parsing, but this violates the contract established by all other `Value` subclasses.

**Fix:** `return make_shared<VariantValue>(v ? v->clone() : nullptr);`

### 4. `using namespace std;` at file scope in header

**File:** `cpp/src/IceBT/DBus.h` line 14

Pollutes the namespace of every translation unit that includes `DBus.h`. Impact is limited since this is a private header within `src/IceBT/`, not a public API header.

**Fix:** Remove and explicitly qualify `std::` where needed.

pepone:
## Summary
- Fix `ExceptionI` shadowing base class `reason` member — error messages from `DBusError` were lost because `init()` set the derived member while callers read the base member.
- Fix stack corruption in DBus boolean read — `dbus_message_iter_get_basic` writes a `dbus_bool_t` (4 bytes) but the code passed the address of a `bool` (1 byte). Benign on little-endian x86-64 due to alignment padding, but technically undefined behavior.
- Fix `VariantValue::clone()` returning an alias (`shared_from_this()`) instead of a deep copy, violating the contract established by all other `Value` subclasses.

The `using namespace std` cleanup (issue item 4) is deferred to a separate change.

Fixes #5129

### DIFF
--- a/cpp/src/IceBT/DBus.h
+++ b/cpp/src/IceBT/DBus.h
@@ -211,11 +211,11 @@ namespace IceBT::DBus
 
         VariantValue(ValuePtr val) : v(std::move(val)), _type(make_shared<VariantType>()) {}
 
-        TypePtr getType() const override { return _type; }
+        [[nodiscard]] TypePtr getType() const override { return _type; }
 
-        ValuePtr clone() const override { return make_shared<VariantValue>(v ? v->clone() : nullptr); }
+        [[nodiscard]] ValuePtr clone() const override { return make_shared<VariantValue>(v ? v->clone() : nullptr); }
 
-        std::string toString() const override { return v ? v->toString() : "nil"; }
+        [[nodiscard]] std::string toString() const override { return v ? v->toString() : "nil"; }
 
         ValuePtr v;
 


cpp/src/IceBT/DBus.cpp
```diff
@@ -592,9 +590,9 @@ namespace
                     return nullptr;
                 case Type::KindBoolean:
                 {
-                    bool v;
+                    dbus_bool_t v;
                     ::dbus_message_iter_get_basic(_iter, &v);
-                    return make_shared<BooleanValue>(v);
+                    return make_shared<BooleanValue>(v != 0);
```
InsertCreativityHere: Elsewhere in this file, we use these macros defined in `dbus.h` when working with D-BUS bools.
Also just feels a little more readable than comparing to `0`.
```suggestion
                    return make_shared<BooleanValue>(v != FALSE);
```

PR #5141, opened 2026-02-18
Fix DataStorm correctness bugs

### LINKED ISSUE #5132: Fix DataStorm correctness bugs: data race, move-assignment leak, double-negation, uninitialized member
### 1. Data race in `NodeI::forwardToPublishers` — missing mutex lock

**File:** `cpp/src/DataStorm/NodeI.cpp` lines 667–679
**Severity:** High — data race / undefined behavior

`forwardToPublishers` iterates `_publishers` without acquiring `_mutex`. Its counterpart `forwardToSubscribers` (line 653) correctly acquires the lock. Since `_publishers` is concurrently modified by `createPublisherSessionServant`, `removePublisherSession`, and `destroy`, iterating without the lock is a data race.

**Fix:** Add `lock_guard<mutex> lock(_mutex);` at the start of `forwardToPublishers`.

### 2. `Node::operator=(Node&&)` leaks the previous instance

**File:** `cpp/src/DataStorm/Node.cpp` lines 93–99
**Severity:** High — resource leak

The move-assignment operator overwrites `_instance` without first destroying the previous instance. The destructor (lines 66–72) correctly calls `_instance->destroy(_ownsCommunicator)`. If a live `Node` is move-assigned, its original communicator, object adapters, timer, callback executor, and connection manager are all leaked.

**Fix:** Destroy the old instance before overwriting, and handle self-assignment:

```cpp
Node& Node::operator=(Node&& node) noexcept
{
    if (this != &node)
    {
        if (_instance)
        {
            _instance->destroy(_ownsCommunicator);
        }
        _instance = std::move(node._instance);
        _factory = std::move(node._factory);
        _ownsCommunicator = node._ownsCommunicator;
    }
    return *this;
}
```

### 3. Double-negation of filter element ID in `DataElementI::disconnect()`

**File:** `cpp/src/DataStorm/DataElementI.cpp` lines 641–654
**Severity:** Medium — filter subscriptions not cleaned up on element destruction

When a data element with filter subscriptions is destroyed, `disconnect()` passes the already-negated `k.second` to `disconnectFromFilter`. This value is forwarded to `unsubscribeFromFilter` (SessionI.cpp line 1076), which negates it again: `subscriber.get(-elementId)`. The double-negation produces a positive key, but filter entries are stored under negative keys — the lookup fails silently and the filter subscription is orphaned.

Contrast with `detachFilter` (line 392), which correctly calls `p->second.get(topicId, -elementId)` with a positive `elementId` that gets negated once.

**Fix:** Negate `k.second` in `disconnect()`:

```cpp
listener.first.session->disconnectFromFilter(k.first, -k.second, shared_from_this(), ks.second->id);
```

### 4. `DataReaderI::_instanceCount` is never initialized or updated

**File:** `cpp/src/DataStorm/DataElementI.h` line 380, `DataElementI.cpp` lines 692–696
**Severity:** Medium — undefined behavior

`int _instanceCount` is not initialized in any constructor and never assigned anywhere. `getInstanceCount()` returns an indeterminate value. Reading an uninitialized variable is UB.

**Fix:** Initialize to zero: `int _instanceCount{0};`

If instance counting is intended to be functional, it should also be incremented/decremented in the appropriate connection callbacks.

pepone:
## Summary
- **Data race**: Add missing `lock_guard<mutex>` in `NodeI::forwardToPublishers` which iterates `_publishers` without synchronization, unlike its counterpart `forwardToSubscribers`.
- **Resource leak**: Fix `Node::operator=(Node&&)` to destroy the previous instance before overwriting, preventing leaks of the communicator, adapters, timer, and connection manager.
- **Double-negation**: Fix `DataElementI::disconnect()` passing an already-negated filter element ID to `disconnectFromFilter`, which negates it again in `unsubscribeFromFilter` — the positive key fails to match filter entries stored under negative keys, silently orphaning the subscription.
- **Uninitialized member**: Initialize `DataReaderI::_instanceCount` to zero to avoid undefined behavior from reading an indeterminate value.

Fixes #5132

## Test plan
- [ ] CI passes
- [ ] DataStorm test suite passes

### DIFF
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -645,7 +645,7 @@ DataElementI::disconnect()
             const auto& k = ks.first;
             if (k.second < 0)
             {
-                listener.first.session->disconnectFromFilter(k.first, k.second, shared_from_this(), ks.second->id);
+                listener.first.session->disconnectFromFilter(k.first, -k.second, shared_from_this(), ks.second->id);
             }
             else
             {
--- a/cpp/src/DataStorm/Node.cpp
+++ b/cpp/src/DataStorm/Node.cpp
@@ -92,9 +92,16 @@ Node::waitForShutdown() const noexcept
 Node&
 Node::operator=(Node&& node) noexcept
 {
-    _instance = std::move(node._instance);
-    _factory = std::move(node._factory);
-    _ownsCommunicator = node._ownsCommunicator;
+    if (this != &node)
+    {
+        if (_instance)
+        {
+            _instance->destroy(_ownsCommunicator);
+        }
+        _instance = std::move(node._instance);
+        _factory = std::move(node._factory);
+        _ownsCommunicator = node._ownsCommunicator;
+    }
     return *this;
 }
 
--- a/cpp/src/DataStorm/DataElementI.h
+++ b/cpp/src/DataStorm/DataElementI.h
@@ -377,7 +377,7 @@ namespace DataStormI
 
         std::deque<std::shared_ptr<Sample>> _samples;
         std::shared_ptr<Sample> _last;
-        int _instanceCount;
+        int _instanceCount{0};
         DataStorm::DiscardPolicy _discardPolicy;
         std::chrono::time_point<std::chrono::system_clock> _lastSendTime;
         std::function<void(const std::shared_ptr<Sample>&)> _onSamples;
--- a/cpp/src/DataStorm/NodeI.cpp
+++ b/cpp/src/DataStorm/NodeI.cpp
@@ -668,6 +668,7 @@ void
 NodeI::forwardToPublishers(const ByteSeq& inParams, const Current& current) const
 {
     // Forward the invocation to all publishers with an active session, don't need to wait for the result.
+    lock_guard<mutex> lock(_mutex);
     assert(current.id == _publisherForwarder->ice_getIdentity());
     for (const auto& [_, publisher] : _publishers)
     {


cpp/src/DataStorm/Node.cpp
```diff
@@ -92,9 +92,16 @@ Node::waitForShutdown() const noexcept
 Node&
 Node::operator=(Node&& node) noexcept
 {
-    _instance = std::move(node._instance);
-    _factory = std::move(node._factory);
-    _ownsCommunicator = node._ownsCommunicator;
+    if (this != &node)
```
InsertCreativityHere, line 95: This is to avoid doing anything when you `std::move` something into itself?...
Something like: `myNode = std::move(myNode)`?

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -645,7 +645,7 @@ DataElementI::disconnect()
             const auto& k = ks.first;
             if (k.second < 0)
             {
-                listener.first.session->disconnectFromFilter(k.first, k.second, shared_from_this(), ks.second->id);
+                listener.first.session->disconnectFromFilter(k.first, -k.second, shared_from_this(), ks.second->id);
```
externl, line 648: Can we add some comments here? `k.second` refers to an `elementId`? The negative implies something special?

externl: This eventually calls `SessionI::unsubscribeFromFilter`

https://github.com/zeroc-ice/ice/blob/adcf402e09b576676b680256f638a1af06e11625/cpp/src/DataStorm/SessionI.cpp#L1076

So we want to double negate this?

InsertCreativityHere: I'm pretty sure negative values indicate a key-filter instead of a key.
But, indeed, switching back and forth from negative to positive internally is confusing, and a comment would be nice.

cpp/src/DataStorm/Node.cpp
pepone: This prevents bugs with self assignment. I think is standard practice.

See for example:

https://github.com/zeroc-ice/ice/blob/1befe175522daba05ad742090ac354eeeab277c1/cpp/src/Ice/OutputStream.cpp#L129-L133

https://github.com/zeroc-ice/ice/blob/1befe175522daba05ad742090ac354eeeab277c1/cpp/src/Ice/Buffer.cpp#L88-L92

InsertCreativityHere: Yes, seems like a good practice. It's just such a corner case (moving into yourself) that I really never thought about it before is all!

I imagine there are other `operator=\(.*&&.*\)` that also need this fix in our code base...

pepone: The move is not always as obvious as `myNode = std::move(myNode)`, for example two references to the same object. And it is just and extra pointer check.

cpp/src/DataStorm/DataElementI.cpp
pepone: DataStorm use negative IDs for elements that represent filters. But some of the APIs for filter expect the positive ID and negate it inside.

pepone: We can add a comment here, but this probably deserves a follow up, to make the usage of `elementId` vs `-elementId` clear through the DataStorm code base. Here I was trying to fix the bogus usage following the existing pattern.

externl, line 648: Are we missing a test for this?

pepone: I added a separate PR for the cleanup, see https://github.com/zeroc-ice/ice/pull/5161

pepone: I think this just results in leaking the subscription, but doesn't affect the behaivor.

PR #5142, opened 2026-02-18
Fix data races in test code found by ThreadSanitizer

pepone:
## Summary

Fixes data races in test infrastructure code detected by ThreadSanitizer (TSan):

- **Ice/location**: `ServerLocator::_requestCount` and `ServerLocatorRegistry::_setRequestCount` were plain integers incremented from multiple dispatch threads via `const_cast`. Changed to `std::atomic` (with `mutable` for the const method case), removing the `const_cast` hack.
- **IceBridge/simple**: `MyClassI::_datagramCount` was a plain `int` written by `datagram()` and read by `getDatagramCount()` on different dispatch threads. Changed to `std::atomic<int>`.
- **IceGrid/session**: `PermissionsVerifier.cpp` called `shutdownOnInterrupt()` before `initialize()`, creating a race where the sigwait thread could read `_communicator` while the main thread was still assigning it. Moved `shutdownOnInterrupt()` after `initialize()`.

All three are test-only bugs — no changes to the Ice runtime.

### DIFF
--- a/cpp/test/Ice/location/ServerLocator.cpp
+++ b/cpp/test/Ice/location/ServerLocator.cpp
@@ -15,6 +15,7 @@ ServerLocatorRegistry::ServerLocatorRegistry() = default;
 void
 ServerLocatorRegistry::setAdapterDirectProxy(string adapter, optional<ObjectPrx> object, const Current&)
 {
+    lock_guard lock(_mutex);
     ++_setRequestCount;
     if (!object)
     {
@@ -33,6 +34,7 @@ ServerLocatorRegistry::setReplicatedAdapterDirectProxy(
     optional<ObjectPrx> object,
     const Current&)
 {
+    lock_guard lock(_mutex);
     ++_setRequestCount;
     if (!object)
     {
@@ -49,6 +51,7 @@ ServerLocatorRegistry::setReplicatedAdapterDirectProxy(
 void
 ServerLocatorRegistry::setServerProcessProxy(string, optional<ProcessPrx>, const Current&)
 {
+    lock_guard lock(_mutex);
     ++_setRequestCount;
 }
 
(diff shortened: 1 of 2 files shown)


cpp/test/Ice/location/ServerLocator.h
```diff
@@ -32,7 +34,7 @@ class ServerLocatorRegistry final : public Test::TestLocatorRegistry
 private:
     std::map<std::string, std::optional<Ice::ObjectPrx>> _adapters;
```
externl, line 37: What about _adapters. We also mutate modify it in the same placeless we modify _setRequestCount.

cpp/test/Ice/location/ServerLocator.cpp
```diff
@@ -112,7 +112,7 @@ ServerLocator::findObjectById(Identity id, const Current&) const
 std::optional<ObjectPrx>
 ServerLocator::findAdapterById(string id, const Current& current) const
 {
-    ++const_cast<int&>(_requestCount);
```
externl, line 115: Why were doing the const_cast before?

pepone: Because this is a const method, we now declare the field as mutable.

externl: I completely missed that.

cpp/test/Ice/location/ServerLocator.h
externl: Since the thread pool has two threads (if I read the test correctly). I get that maybe we don't make simultaneous calls, but don't we need at a minimum some memory barrier?

pepone: I adding a mutex to be safe, I only fixed `_setRequestCount` initially because that is the issue reported by the sanitizer

Issue #5133, opened 2026-02-17
DataStorm: broken notifyListenerWaiters synchronization and minor cleanup


pepone:
### 1. `notifyListenerWaiters` wait predicate is always true (no-op wait)

**Files:** `cpp/src/DataStorm/TopicI.cpp` line 792, `DataElementI.cpp` line 626

Both `TopicI::notifyListenerWaiters` and `DataElementI::notifyListenerWaiters` use:

```cpp
_notified = 0;
_cond.notify_all();
_cond.wait(lock, [&]() { return _notified < _waiters; });
```

Since `_notified` was just set to 0 and `_waiters > 0`, the predicate `0 < N` is immediately true. The wait always returns without blocking. The intended condition is likely `_notified >= _waiters` (wait until all waiters have acknowledged).

The `_notified` / `_waiters` bookkeeping is effectively dead code. The system works because `_cond.notify_all()` is sufficient for the waiters, but the intended guarantee — that the notifier waits until all waiters have observed the state change — is not enforced.

### 2. Missing `notify_all` in `DataElementI::waitForListeners`

**File:** `cpp/src/DataStorm/DataElementI.cpp` lines 560–565

`TopicI::waitForListeners` (line 774) calls `_cond.notify_all()` after `++_notified`. The `DataElementI` version does not. Currently masked by issue #1 above (the wait never blocks anyway).

### 3. Unconditional trace output on sample delivery hot path

**File:** `cpp/src/DataStorm/SessionI.cpp` lines 1272–1285

`Trace` object is always constructed and emitted when `current.con != _connection`, regardless of configured trace level. Should be gated with a level check.

### 4. Unused `#include <iostream>` in ConnectionManager.cpp

**File:** `cpp/src/DataStorm/ConnectionManager.cpp` line 6

### 5. Unused variable `cancel` in `KeyDataWriterI::forward`

**File:** `cpp/src/DataStorm/DataElementI.cpp` line 1361

Return value of `ice_invokeAsync` is captured but never used.

pepone:
Fixed in https://github.com/zeroc-ice/ice/pull/5143

PR #5143, opened 2026-02-18
Fix broken notifyListenerWaiters synchronization and minor cleanup

pepone:
## Summary

Fixes several issues in DataStorm (#5133):

- **Simplify `notifyListenerWaiters`** in `TopicI` and `DataElementI`: the `_notified`/`_waiters` handshake was broken (wait predicate `_notified < _waiters` was always true after resetting `_notified = 0`, so the wait never blocked). Since `notify_all()` alone is sufficient to wake waiters — and no caller depends on waiters having acknowledged before proceeding — the fix removes the dead handshake and keeps only `notify_all()`. The `_waiters` and `_notified` members are removed.
- **Gate trace output** in `SubscriberSessionI::s` on `_traceLevels->session > 0` to avoid unconditional `Trace` construction on the sample delivery hot path.
- **Remove unused `#include <iostream>`** in `ConnectionManager.cpp`.
- **Remove unused `cancel` variable** in `KeyDataWriterI::forward`.

## Test plan

- [x] Run `python allTests.py --filter DataStorm/` — all 7 tests pass

### DIFF
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -325,7 +325,7 @@ DataElementI::attachFilter(
     const shared_ptr<SessionI>& session,
     SessionPrx prx,
     const string& facet,
-    int64_t filterId,
+    int64_t subscriberId,
     const shared_ptr<Filter>& filter,
     const string& name,
     int priority)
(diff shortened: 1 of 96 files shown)


cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -617,14 +613,9 @@ DataElementI::removeConnectedKey(const shared_ptr<Key>& key, const shared_ptr<Su
 }
 
 void
-DataElementI::notifyListenerWaiters(unique_lock<mutex>& lock) const
+DataElementI::notifyListenerWaiters(unique_lock<mutex>&) const
```
externl: Can we remove this parameter from the two `notifyListenerWaiters` functions?

pepone: Removed the unused parameters.

cpp/src/DataStorm/TopicI.cpp
```diff
@@ -694,17 +694,17 @@ TopicI::getUpdaters() const
...
 }
 
 void
-TopicI::decListenerCount(const shared_ptr<SessionI>& session)
+TopicI::decListenerCount()
 {
     --_listenerCount;
-    notifyListenerWaiters(session->getTopicLock());
+    notifyListenerWaiters();
 }
```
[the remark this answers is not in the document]
pepone: The lock was only passed to wait on the condition, which was not required. Acquiring the lock was already the parent responsibility, this PR doesn't changed that.

Issue #4755, opened 2025-12-12
Update Python doc comments


bernardnormier:
Some Python doc-comments are out of sync with the C++/Java doc-comments.

externl:
Is this completed @InsertCreativityHere?

PR #5145, opened 2026-02-18
Finish Updating Python Doc-Comments

### LINKED ISSUE #4881: Improve Formatting of Doc-Comments for Free Python Functions Defined in IcePy
We define a handful of free functions in the IcePy C++ code:
https://github.com/zeroc-ice/ice/blob/main/python/modules/IcePy/Init.cpp
https://github.com/zeroc-ice/ice/blob/f83c33b82803072c53e9072852e3282408cb005f/python/python/IcePy-stubs/__init__.pyi#L686-L694

The rendered API reference for these functions is messed up.
For some of these, it's not a problem, because we 'shadow' them with real Python functions in `Ice`, and Sphinx uses those functions instead. But, for free functions which aren't 'shadow'ed by definitions in `Ice`, there's problems.

For example, take `loadSlice`: https://code.zeroc.com/ice/main/api/python/Ice.loadSlice.html
compared to `createProperties`: https://code.zeroc.com/ice/main/api/python/Ice.createProperties.html

It looks like `loadSlice` is documented as an attribute, and not a function?
At the very top, it `list[str]) -> None` instead of the nice function rendering `createProperties` does.
And at the very bottom, it says: `Type: loadSlice(args`, which is nonsense. Functions don't have types.

For something without any parameters, the formatting is better, but still messed up.
Take `stringVersion()`: https://code.zeroc.com/ice/main/api/python/Ice.stringVersion.html#Ice.stringVersion

InsertCreativityHere:
This PR implements #4755 (finish updating doc-comments in Python) and fixes #4881 (add wrapper functions to `Ice` that wrap `IcePy` functions so that the generated docs for these functions are more reasonable).

### DIFF
--- a/python/modules/IcePy/Util.h
+++ b/python/modules/IcePy/Util.h
@@ -121,9 +121,7 @@ namespace IcePy
     /// Extracts the members of an encoding version.
     bool getEncodingVersion(PyObject*, Ice::EncodingVersion&);
 
-    /// Call a Python method.
     PyObject* callMethod(PyObject*, const std::string&, PyObject* = nullptr, PyObject* = nullptr);
-    /// Call a Python method.
     PyObject* callMethod(PyObject*, PyObject* = nullptr, PyObject* = nullptr);
 
     /// Returns true if the current thread is the main thread.
--- a/python/python/Ice/Util.py
+++ b/python/python/Ice/Util.py
@@ -184,7 +184,7 @@ def loadSlice(args: list[str]) -> None:
     RuntimeError
         If an error occurs during Slice parsing or compilation.
     """
-    return IcePy.loadSlice(args)
+    IcePy.loadSlice(args)
 
 
 def getSliceDir() -> str | None:
--- a/python/python/Ice/IcePyTypes.py
+++ b/python/python/Ice/IcePyTypes.py
@@ -20,9 +20,6 @@ from IcePy import (
     UDPEndpointInfo,
     WSConnectionInfo,
     WSEndpointInfo,
-    intVersion,
-    loadSlice,
-    stringVersion,
 )
 
 __all__ = [
@@ -43,7 +40,4 @@ __all__ = [
     "UDPEndpointInfo",
     "WSConnectionInfo",
     "WSEndpointInfo",
-    "intVersion",
-    "loadSlice",
-    "stringVersion",
 ]


python/modules/IcePy/Init.cpp
```diff
@@ -143,11 +143,11 @@ args : list[str]
...
-        - `-DNAME` Define NAME as 1.
-        - `-DNAME=DEF` Define NAME as DEF.
-        - `-UNAME` Remove any definition for NAME.
-        - `-IDIR` Put DIR in the include file search path.
-        - `-d`, `--debug`  Print debug messages.
+        - ``-DNAME``:  Define NAME as 1.
+        - ``-DNAME=DEF``:  Define NAME as DEF.
+        - ``-UNAME``:  Remove any definition for NAME.
+        - ``-IDIR``:  Put DIR in the include file search path.
+        - ``-d``, ``--debug``:  Print debug messages.
```
InsertCreativityHere, line 150: Sphinx uses double back-ticks for code formatting, not single.
Also, I think adding a colon separator makes this list more readable.

python/modules/IcePy/Logger.h
```diff
@@ -36,9 +34,7 @@ namespace IcePy
 
     void cleanupLogger();
 
-    //
-    // Create a Python object that delegates to a C++ implementation.
-    //
```
InsertCreativityHere, line 41: I went through all the `//` style comments and either updated them to a real doc-comment (`///`), or just trimmed the comment down if it was more of an implementation detail comment than a doc-comment (like the comment for `LoggerWrapper` above this).

python/python/Ice/Communicator.py
```diff
@@ -32,11 +32,13 @@
...
     Communicator is the central object in Ice. Its responsibilities include:
+
     - creating and managing outgoing connections
     - executing callbacks in its client thread pool
     - creating and destroying object adapters
     - loading plug-ins
     - managing properties (configuration), retries, logging, instrumentation, and more.
+
     A communicator is usually the first object you create when programming with Ice.
     You can create multiple communicators in a single program, but this is not common.
```
InsertCreativityHere, line 43: Sphinx requires empty lines around markdown lists in certain situations. Here is what the API reference currently looks like (without the extra lines):
image

----

The list is all collapsed together. Adding these blank lines fixes it:

----

image

python/python/Ice/ObjectAdapter.py
```diff
@@ -228,17 +229,23 @@ def addDefaultServant(self, servant: Object, category: str) -> None:
         Adds a default servant to handle requests for a specific category. When an object adapter dispatches an
         incoming request, it tries to find a servant for the identity and facet carried by the request in the
         following order:
```
InsertCreativityHere, line 231: Here there was also an extra column of whitespace in front of the list elements, which caused an error while building the docs:
```
D:\Code\Workspace\ice\python\python\Ice\ObjectAdapter.py:docstring of Ice.ObjectAdapter.ObjectAdapter.addDefaultServant:4: ERROR: Unexpected indentation. [docutils]
```
This extra space in front was removed from the list.

python/python/Ice/Util.py
```diff
@@ -57,6 +57,33 @@ def initialize(
         The new communicator.
     """
     return Communicator(initData=initData) if initData is not None else Communicator(args, eventLoop)
+
```
InsertCreativityHere: The new wrapper functions for `stringVersion`, `intVersion`, and `loadSlice` (these were the only free functions from IcePy that didn't already have wrapper functions for them).

Without these wrappers the API reference for free functions is messed up. See #4881.
The doc-comments for these functions match C# and Java. and also the existing doc-comments in IcePy.

python/modules/IcePy/Init.cpp
```diff
@@ -158,7 +158,7 @@ RuntimeError
 
 Compiles Slice definitions. The behavior is identical to that of the `slice2py` compiler.
 
-Slice compilation errors and warnings are printed to standard error (stderr).
+Any errors or warnings raised during compilation are printed to 'stderr'.
```
pepone: is the `'` single quote intentional here?

python/modules/IcePy/Logger.h
```diff
@@ -36,9 +34,7 @@ namespace IcePy
...
+    /// Create a Python object that delegates to a C++ implementation.
```
pepone, line 37: This comment isn't very useful, maybe we should just remove.

python/modules/IcePy/Util.h
```diff
@@ -68,123 +52,78 @@ namespace IcePy
...
-    // Extracts the members of an encoding version.
-    //
+    /// Extracts the members of an encoding version.
     bool getEncodingVersion(PyObject*, Ice::EncodingVersion&);
 
-    //
-    // Call a Python method.
-    //
+    /// Call a Python method.
     PyObject* callMethod(PyObject*, const std::string&, PyObject* = nullptr, PyObject* = nullptr);
```
pepone, line 124: I guess this PR is just about the formatting, but the comment here doesn't have any value.

python/python/Ice/Util.py
```diff
@@ -123,6 +150,43 @@ def createProperties(args: list[str] | None = None, defaults: Properties | None
...
+            - ``-UNAME``:  Remove any definition for NAME.
+            - ``-IDIR``:  Put DIR in the include file search path.
+            - ``-d``, ``--debug``:  Print debug messages.
+
+    Raises
+    ------
+    RuntimeError
+        If an error occurs during Slice parsing or compilation.
+    """
+    return IcePy.loadSlice(args)
```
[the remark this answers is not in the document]
pepone: I think copilot is right, we don't need the return herre.

python/python/Ice/__init__.py
```diff
@@ -165,7 +165,16 @@
...
-from .Util import createProperties, getSliceDir, identityToString, initialize, stringToIdentity
+from .Util import (
+    createProperties,
+    getSliceDir,
+    identityToString,
+    initialize,
+    intVersion,
+    loadSlice,
+    stringToIdentity,
+    stringVersion,
```
pepone, line 173: `intVersion`, `loadSlice`, and `stringVersion` are imported twice, the second import shadows the first one. It works but it is not required, we should remove the `IcePyTypes` import for those.

InsertCreativityHere: Sure, I removed the old imports from the `IcePyTypes` import list. Now we only import the ones from 'Ice'.

python/python/Ice/Util.py
externl: Agree. The return doesn't seem necessary.

python/python/Ice/Util.py
externl, line 1: I believe you also need to remove the re-export from `IcePyTypes.py`

python/modules/IcePy/Logger.h
```diff
@@ -11,9 +11,7 @@
 
 namespace IcePy
 {
-    //
     // LoggerWrapper delegates to a Python implementation.
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes, this comment just isn't helpful enough to be a doc-comment.
To me, it reads more like an implementation detail about the thing, than being about the thing itself.

python/modules/IcePy/Init.cpp
InsertCreativityHere: Yeah, it was intentional on my part. I thought about using backticks, but usually I use backticks for programmatic constructs like class or variable names, and here, stdout is more of a concept than the name of a program element.

But if you have a preference of delimiter, I can change. None of the comments in these files are very consistent with these minor details anyways!

python/modules/IcePy/Util.h
InsertCreativityHere: I just deleted the comments. I don't think they had 0 value, but it was pretty close to 0!

python/python/Ice/Util.py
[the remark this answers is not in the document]
InsertCreativityHere: Yes, I removed the return. We just call the function normally now.

python/modules/IcePy/Util.h
```diff
@@ -68,123 +52,78 @@ namespace IcePy
...
+    /// Call a Python method.
```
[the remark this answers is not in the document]
InsertCreativityHere: I just removed these comments.

python/python/Ice/Util.py
InsertCreativityHere: Okay, I also removed these old re-exports from `IcePyTypes`.
I though these served a greater purpose than just re-exporting into Ice for __init__ to re-re-export them again.

python/python/Ice/Util.py
InsertCreativityHere: I removed the `return`.

python/python/Ice/Util.py
externl: They're required when we want to reference the re-exported types as `Ice.` types from inside Ice itself.

PR #5146, opened 2026-02-18
Handle ParseException in C++ TrustManager

### LINKED ISSUE #5134: Handle ParseException in C++ TrustManager
> > Don't we want to catch this exception in C++ and return false like in C# and Java? Otherwise should we just let the exception propagate everywhere?
> 
> Yes the C++ bug is that the connection is terminated with the ParseException. Java/C# bug is that we accepted the connection if there was a exception. We can include the C++ fix in this PR even if it is kind of a different bug.

A separate PR is fine. Just wanted to make sure I understood the issue correctly.

_Originally posted by @externl in https://github.com/zeroc-ice/ice/issues/5117#issuecomment-3917029630_

externl:
Fixes #5134

### DIFF
--- a/cpp/src/Ice/SSL/TrustManager.cpp
+++ b/cpp/src/Ice/SSL/TrustManager.cpp
@@ -119,7 +119,18 @@ TrustManager::verify(const ConnectionInfoPtr& info) const
     // If there is no certificate then we match false.
     if (info->peerCertificate)
     {
-        auto subject = DistinguishedName(getSubjectName(info->peerCertificate));
+        DistinguishedName subject{list<pair<string, string>>{}};
+        try
+        {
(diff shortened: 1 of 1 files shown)


cpp/src/Ice/SSL/TrustManager.cpp
```diff
@@ -119,7 +119,18 @@ TrustManager::verify(const ConnectionInfoPtr& info) const
     // If there is no certificate then we match false.
     if (info->peerCertificate)
     {
-        auto subject = DistinguishedName(getSubjectName(info->peerCertificate));
+        DistinguishedName subject{list<pair<string, string>>{}};
```
externl, line 122: Another option is to try/catch the entire scope where `subject` is used to avoid calling this list constructor.

PR #5152, opened 2026-02-19
Fix adapter ID filter polluted with category filter entries

pepone:
## Summary

Fixed a bug in `FilterManager::create` where the adapter ID filter was polluted with category filter
entries because `stringToSeq` appended to a shared `vector` that was never cleared between uses. As a
result, when both `Glacier2.Filter.Category.Accept` and `Glacier2.Filter.AdapterId.Accept` were
configured, the adapter ID filter would also accept adapter IDs matching category names.

Fix: refactored the two `stringToSeq` overloads to return by value (`stringToStringSeq`,
`stringToIdentitySeq`), eliminating the reuse-without-clear bug class entirely.

### DIFF
--- a/cpp/src/Glacier2/FilterManager.cpp
+++ b/cpp/src/Glacier2/FilterManager.cpp
@@ -15,19 +15,22 @@ using namespace Ice;
 // Parse a space delimited string into a sequence of strings.
 //
 
-static void
-stringToSeq(const string& str, vector<string>& seq)
+static vector<string>
+stringToStringSeq(const string& str)
 {
+    vector<string> seq;
     IceInternal::splitString(str, " \t", seq);
 
     //
     // TODO: do something about unmatched quotes
     //
+    return seq;
 }
 
-static void
-stringToSeq(const string& str, vector<Identity>& seq)
+static vector<Identity>
+stringToIdentitySeq(const string& str)
 {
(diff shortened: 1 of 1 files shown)


cpp/src/Glacier2/FilterManager.cpp
```diff
@@ -190,20 +192,11 @@ Glacier2::FilterManager::create(shared_ptr<Instance> instance, const string& use
...
-    // TODO: refactor initialization of filters.
-    //
-    allow = props->getIceProperty("Glacier2.Filter.AdapterId.Accept");
-    stringToSeq(allow, allowSeq);
-    auto adapterIdFilter = make_shared<Glacier2::StringSetI>(allowSeq);
+    auto adapterIdFilter =
+        make_shared<Glacier2::StringSetI>(stringToStringSeq(props->getIceProperty("Glacier2.Filter.AdapterId.Accept")));
 
-    //
-    // TODO: Object id's from configurations?
```
InsertCreativityHere, line 201: Definitely not the most descriptive `TODO`... but does this PR actually accomplish this TODO? Seems unrelated to changing the functions to use return values instead of 'out' parameters.

pepone: I removed it because seems non actionable, it has been there for 20 years!

PR #5153, opened 2026-02-19
Simplify Glacier2 session exception handling

pepone:
## Summary
- Inline the `unexpectedAuthorizeException` logic directly into `checkPermissionsException` and `authorizeException`, and remove the now-unused method.
- Narrow catch clauses from `catch (...)` to `catch (const Ice::Exception&)` since these handlers only receive exceptions from Ice async invocation callbacks.
- Remove the extra `std::exception` and `catch (...)` handlers from `unexpectedCreateSessionException` and `sessionDestroyException`.

## Test plan
- [ ] Run Glacier2 tests with session trace level >= 1

### DIFF
--- a/cpp/src/Glacier2/SessionRouterI.cpp
+++ b/cpp/src/Glacier2/SessionRouterI.cpp
@@ -129,9 +129,14 @@ namespace Glacier2
             {
                 exception(ex);
             }
-            catch (...)
+            catch (const Ice::Exception& e)
             {
-                unexpectedAuthorizeException(ex);
+                if (_sessionRouter->sessionTraceLevel() >= 1)
+                {
+                    Warning out(_instance->logger());
+                    out << "exception while verifying permissions:\n" << e;
+                }
+                exception(make_exception_ptr(PermissionDeniedException("internal server error")));
             }
         }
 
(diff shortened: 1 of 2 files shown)


cpp/src/Glacier2/SessionRouterI.cpp
```diff
@@ -395,6 +395,16 @@ CreateSession::unexpectedAuthorizeException(exception_ptr ex)
             Warning out(_instance->logger());
             out << "exception while verifying permissions:\n" << e;
         }
+        catch (const std::exception& e)
+        {
+            Warning out(_instance->logger());
+            out << "exception while verifying permissions:\n" << e.what();
+        }
+        catch (...)
```
externl: I don't think we should worry about `...`. Isn't ` catch (const std::exception& e)` sufficient for all of this. 

This is called from an exception callback from a request. All of our exceptions derive from std::exception.

I think we should only catch `std::exception` in all of these. It's likely that `Ice::Exception` is also fully adequate. Do we every throw a non-ice exception from an invocation?

pepone: You are right the extra catch didn't make much sense, I removed the `unexpectedAuthorizeException` helper, and updated the callers to handle `Ice::Exception` directly.  Having a function catch `...` to pass to the helper that only deals with `Ice::Exception` made the expectation unclear with the previous code.

It is mostly a cosmetic fix but I think makes the intent clear.

PR #5154, opened 2026-02-19
Fix Glacier2 proxy address filter matching bugs

pepone:
## Summary

Fixed bugs in the Glacier2 proxy address filter matching (`ProxyVerifier.cpp`):

- **`MatchesString::match` ignored `pos` parameter**: always compared from position 0 instead of the
  current position, causing incorrect matching when a literal string followed a numeric range in an
  address filter (e.g., `12[7].0.0.*` would fail to match `127.0.0.1`). The practical impact is overly
  restrictive filtering — valid addresses would be incorrectly rejected.
- **`ContainsString::match` off-by-one**: advanced `pos` one character past the match end, which could
  cause subsequent matchers to skip a character.

Also includes minor cleanups:
- Removed dead `EndsWithNumber` class (unreachable code).
- Removed `std::ends` artifact that inserted an embedded null into trace output.
- Added bounds check in `MatchesString` using overflow-safe arithmetic.
- Added test case with numeric range in hostname (`12[7].0.0.*`).

### DIFF
--- a/cpp/src/Glacier2/ProxyVerifier.cpp
+++ b/cpp/src/Glacier2/ProxyVerifier.cpp
@@ -181,7 +181,8 @@ namespace Glacier2
 
         bool match(const string& space, string::size_type& pos) override
         {
-            if (strncmp(space.c_str(), _criteria.c_str(), _criteria.size()) == 0)
+            if (pos + _criteria.size() <= space.size() &&
+                strncmp(space.c_str() + pos, _criteria.c_str(), _criteria.size()) == 0)
             {
                 pos += _criteria.size();
                 return true;
@@ -212,7 +213,7 @@ namespace Glacier2
             {
                 return false;
             }
-            pos = offset + _criteria.size() + 1;
+            pos = offset + _criteria.size();
             return true;
         }
 
@@ -279,7 +280,6 @@ namespace Glacier2
                     ostr << range.start << " up to " << range.end;
                 }
             }
-            ostr << ends;
             _description = ostr.str();
         }
 
@@ -351,26 +351,6 @@ namespace Glacier2
         }
     };
 
-    class EndsWithNumber final : public MatchesNumber
-    {
-    public:
-        EndsWithNumber(const vector<int>& values, const vector<Range>& ranges)
-            : MatchesNumber(values, ranges, "ends with ")
-        {
-        }
-
-        bool match(const string& space, string::size_type& pos) override
-        {
-            pos = space.find_last_not_of("0123456789", pos);
-            if (pos == space.size() - 1)
-            {
-                return false;
-            }
-
-            return MatchesNumber::match(space, pos);
-        }
-    };
-
     //
     // AddressMatcher factories abstract away the logic of which matching
     // objects need to be created depending on the state of the filter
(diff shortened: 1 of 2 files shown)


cpp/src/Glacier2/ProxyVerifier.cpp
```diff
@@ -181,7 +181,8 @@ namespace Glacier2
 
         bool match(const string& space, string::size_type& pos) override
         {
-            if (strncmp(space.c_str(), _criteria.c_str(), _criteria.size()) == 0)
+            if (space.size() - pos >= _criteria.size() &&
+                strncmp(space.c_str() + pos, _criteria.c_str(), _criteria.size()) == 0)
```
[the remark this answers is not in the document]
pepone: Good catch — fixed. Changed to `pos + _criteria.size() <= space.size()` to avoid unsigned underflow.

cpp/test/Glacier2/staticFiltering/test.py
```diff
@@ -292,6 +292,21 @@ def setup(self, current):
...
+        testcases.extend(
+            [
+                (
+                    "testing address filter with numeric range in hostname",
+                    ("12[7].0.0.*", "", "", "", "", ""),
+                    [
+                        (True, "hello:tcp -h 127.0.0.1 -p 12010"),
+                        (False, "hello:tcp -h 128.0.0.1 -p 12010"),
+                        (False, "hello:tcp -h 126.0.0.1 -p 12010"),
+                    ],
```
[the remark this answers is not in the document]
pepone: The address filter operates on hostnames and IP addresses alike — the test uses an IP address because it works reliably in CI without DNS. The description reflects the feature being tested (numeric ranges in the host portion of an address filter) rather than the specific test data.

cpp/src/Glacier2/ProxyVerifier.cpp
```diff
@@ -351,26 +352,6 @@ namespace Glacier2
         }
     };
 
-    class EndsWithNumber final : public MatchesNumber
```
pepone, line 354: This was dead code. The implementation only match numbers inside `[..]`. The `EndsWithFactory` factory is only active after all `*`, `[`, `]` groups have been processed.

cpp/src/Glacier2/ProxyVerifier.cpp
```diff
@@ -181,7 +181,8 @@ namespace Glacier2
 
         bool match(const string& space, string::size_type& pos) override
```
pepone, line 182: The match was ignoring `pos` here. Seems this is only activated after `[]` in an address, which the implementation supports but is not documented in https://docs.zeroc.com/ice/3.8/cpp/glacier2-1#Glacier2.*-Glacier2.Filter.Category.Accept

cpp/test/Glacier2/staticFiltering/test.py
```diff
@@ -292,6 +292,21 @@ def setup(self, current):
             ),
         ]
 
+        testcases.extend(
+            [
+                (
+                    "testing address filter with numeric range in hostname",
+                    ("12[7].0.0.*", "", "", "", "", ""),
```
externl, line 299: Do we have test for larger ranges, like `[7,8]`, `[7-9]`?

pepone: I don't think so. Note that what was broken here was matching the string after the range, not the range match.

PR #5160, opened 2026-02-19
Random Slice Compiler Fixes

InsertCreativityHere:
This PR fixes all the random little things Claude pointed out in the Slice compilers and `libSlice`.
Each individual thing is pretty small, so I just added PR comments to explain each of them.

### DIFF
--- a/cpp/src/Slice/MetadataValidation.cpp
+++ b/cpp/src/Slice/MetadataValidation.cpp
@@ -99,16 +99,16 @@ Slice::validateMetadata(const UnitPtr& p, string_view prefix, map<string, Metada
 
     // We keep support for 'deprecate' as an alias for 'deprecated', but disallow using both on the same element.
     deprecatedInfo.extraValidation = [](const MetadataPtr&, const SyntaxTreeBasePtr& q) -> optional<string>
+    {
+        if (auto contained = dynamic_pointer_cast<Contained>(q))
         {
-            if (auto contained = dynamic_pointer_cast<Contained>(q))
+            if (contained->hasMetadata("deprecated"))
             {
-                if (contained->hasMetadata("deprecated"))
-                {
-                    return "you cannot apply both 'deprecated' and 'deprecate' metadata to the same element";
-                }
+                return "you cannot apply both 'deprecated' and 'deprecate' metadata to the same element";
             }
-            return nullopt;
-        },
+        }
+        return nullopt;
+    },
     knownMetadata.emplace("deprecate", std::move(deprecatedInfo));
 
     // "format"
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -4047,9 +4047,7 @@ Slice::Enum::destroy()
     destroyContents();
 }
 
-Slice::Enum::Enum(const ContainerPtr& container, const string& name) : Contained(container, name)
-{
-}
+Slice::Enum::Enum(const ContainerPtr& container, const string& name) : Contained(container, name) {}
 
 // ----------------------------------------------------------------------
 // Enumerator


cpp/src/Slice/MetadataValidation.cpp
```diff
@@ -95,8 +95,21 @@ Slice::validateMetadata(const UnitPtr& p, string_view prefix, map<string, Metada
              typeid(DataMember)},
         .acceptedArgumentKind = MetadataArgumentKind::OptionalTextArgument,
     };
```
InsertCreativityHere, line 97: Right now, the compiler was allowing you to place both `deprecated` and `deprecate` on the same element when this is a bogus thing to do. So I added an extra validation function to `deprecate` which will reject itself if there is already `deprecated` metadata on the same thing.

We already have logic to prevent duplicates (so putting `deprecated` twice is already handled), but since `deprecated` and `deprecate` are technically different metadata, this duplicate check let it through.

cpp/src/Slice/Parser.cpp
```diff
@@ -11,7 +11,6 @@
 #include <cassert>
 #include <cstring>
 #include <iterator>
-#include <limits>
```
InsertCreativityHere, line 14: This include moved to the `Parser.h` header file.
You'll see.

cpp/src/Slice/Parser.cpp
```diff
@@ -2197,7 +2198,7 @@ Slice::ClassDecl::isClassType() const
 size_t
 Slice::ClassDecl::minWireSize() const
 {
-    return 1; // At least four bytes for an instance, if the instance is marshaled as an index.
```
InsertCreativityHere, line 2200: Bogus comment. It says the min is 4, despite us clearly returning 1. 
Fixed the comment to be correct.

cpp/src/Slice/Parser.cpp
```diff
@@ -2955,10 +2956,9 @@ Slice::Operation::hasMarshaledResult() const
             return true;
         }
 
-        for (const auto& p : _contents)
```
InsertCreativityHere, line 2958: Not a bug, but might as well just call `outParameters()` instead of checking `_contents`, casting to a parameter, and checking `isOutParam`.

cpp/src/Slice/Parser.cpp
```diff
@@ -4048,9 +4047,7 @@ Slice::Enum::destroy()
     destroyContents();
 }
 
-Slice::Enum::Enum(const ContainerPtr& container, const string& name)
-    : Contained(container, name),
-      _minValue(numeric_limits<int32_t>::max())
```
InsertCreativityHere, line 4053: We set `_minValue`'s value in the field declaration now, instead of the constructor.

cpp/src/Slice/Parser.h
```diff
@@ -947,7 +948,7 @@ namespace Slice
 
     private:
         bool _hasExplicitValues{false};
-        std::int32_t _minValue;
+        std::int32_t _minValue{std::numeric_limits<std::int32_t>::max()};
```
InsertCreativityHere, line 951: Right here

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -86,7 +86,7 @@ Slice::Preprocessor::normalizeIncludePath(const string& path)
     //
     // MCPP does not handle "-IC:/" well as an include path.
     //
```
InsertCreativityHere, line 88: Old logic seemed bogus. The new logic is that we want to replace all '\' with '/' unless we are specifically dealing with a path like 'C:\'. So we check if the path is exactly 3 characters long, the first character is a letter, the 2nd character is ':' and the 3rd character is '\'.

If all of those are true, then we skip the if and don't replace slashes.

cpp/src/Slice/StringLiteralUtil.cpp
```diff
@@ -48,21 +48,14 @@ namespace
...
         unsigned char cutOff)
         : _nonPrintableEscaped(std::move(nonPrintableEscaped)),
-          _printableEscaped(printableEscaped + "\\"),
+          // Double quotes don't need to be escaped in Matlab because the string delimiter is a single quote.
+          _printableEscaped(printableEscaped + "\\" + (escapeMode != Matlab ? "\"" : "")),
           _escapeMode(escapeMode),
           _cutOff(cutOff),
           _shortUCNPrefix(escapeMode == Matlab ? "\\x" : "\\u"),
           _octalChars("01234567"),
-          _hexChars("01234567890ABCDEFabcdef")
```
InsertCreativityHere, line 56: There was an extra 0 here. Doesn't really break anything, but good to remove.

cpp/src/Slice/StringLiteralUtil.cpp
```diff
@@ -48,21 +48,14 @@ namespace
...
-
+          _hexChars("0123456789ABCDEFabcdef")
     {
-        //
-        // Double quotes don't need to be escaped in Matlab because the string delimiter is a single quote.
-        //
-        if (_escapeMode != Matlab)
-        {
-            const_cast<string&>(_printableEscaped) += '"';
-        }
```
InsertCreativityHere, line 65: We add this in the initialization list, instead of in the constructor now.

cpp/src/slice2js/Gen.cpp
```diff
@@ -32,7 +31,7 @@ namespace
         {
             f = relativePath(f, Slice::dirName(relativeTo));
         }
-        else if (f[0] != '.')
+        else if (!f.empty() && f[0] != '.')
```
pepone, line 34: Something is very wrong if we pass an empty path here, I think we can either throw at the beginning or just assert.

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -819,13 +819,15 @@ namespace
         }
         else if (auto dict = dynamic_pointer_cast<Dictionary>(type))
         {
+            out << "dictionary(";
```
pepone, line 822: How was this not catch in the tests?

pepone (approved):

cpp/src/slice2cpp/Gen.cpp
```diff
@@ -2197,7 +2197,7 @@ Slice::Gen::DataDefVisitor::visitClassDefStart(const ClassDefPtr& p)
         H << sp;
         H << nl << "/// Creates a tuple with all the fields of this class.";
         H << nl << "/// @return A tuple with all the fields of this class.";
-        writeIceTuple(H, p->allDataMembers(), _useWstring);
```
InsertCreativityHere, line 2200: We already had this stored in another variable, no need to re-call the function.

cpp/src/slice2java/Gen.cpp
```diff
@@ -3,13 +3,9 @@
 #include "Gen.h"
 #include "../Slice/Util.h"
 #include "Ice/StringUtil.h"
```
InsertCreativityHere, line 5: Unused includes apparently.

cpp/src/slice2java/Gen.cpp
```diff
@@ -18,7 +14,7 @@ using namespace IceInternal;
 
 namespace
 {
-    static const char* builtinNameTable[] = {"Byte", "Bool", "Short", "Int", "Long", "Float", "Double", "String"};
+    const char* builtinNameTable[] = {"Byte", "Bool", "Short", "Int", "Long", "Float", "Double", "String"};
```
InsertCreativityHere, line 17: This isn't a member, it's just in an anonymous namespace, so there's no point marking it static.

cpp/src/slice2js/Gen.cpp
```diff
@@ -32,7 +31,7 @@ namespace
         {
             f = relativePath(f, Slice::dirName(relativeTo));
         }
-        else if (f[0] != '.')
```
InsertCreativityHere, line 35: If this function was passed the empty string, this wouldn't work.

cpp/src/slice2js/Gen.cpp
```diff
@@ -2075,13 +2074,11 @@ bool
...
 {
     // Add imports required for base exception types.
-    ExceptionPtr base = p->base();
-    if (base)
+    if (ExceptionPtr base = p->base())
     {
         addImport(dynamic_pointer_cast<Contained>(base));
     }
 
-    const DataMemberList allDataMembers = p->allDataMembers();
```
InsertCreativityHere, line 2084: This variable was unused.

cpp/src/slice2js/JsUtil.cpp
```diff
@@ -524,7 +524,7 @@ Slice::JavaScript::jsLinkFormatter(const string& rawLink, const ContainedPtr&, c
             else
             {
                 string targetScoped = dynamic_pointer_cast<Contained>(target)->mappedScoped(".");
-                if (auto interfaceTarget = dynamic_pointer_cast<InterfaceDecl>(target))
```
InsertCreativityHere, line 527: `interfaceTarget` was unused. All we care about is "is this an interface"? We don't actually call any of `InterfaceDecl`'s methods though.

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -819,13 +819,15 @@ namespace
...
+
```
InsertCreativityHere, line 823: An actual bug, without this, we were generating doc-comments like:
```
        % MS
        %   string vector
        ms (1, :) char
        
        % MD
        %   string, string) scalar
        md (1, 1) dictionary = configureDictionary('char', 'char')
```
Where we're just missing the 'dictionary(' in front of the key/value types of the field. So, now we generate:
```diff
-string, string) scalar
+dictionary(string, string) scalar
```
Or whatever the correct key/value type obviously.

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -3316,7 +3318,7 @@ Slice::matlabLinkFormatter(const string& rawLink, const ContainedPtr&, const Syn
             displayText = opTarget->mappedName();
             linkText = interfaceDef->mappedScoped(".") + "Prx" + "/" + displayText;
         }
-        else if (auto interfaceTarget = dynamic_pointer_cast<InterfaceDecl>(target))
```
InsertCreativityHere, line 3319: Again, we never actually used `interfaceTarget`. Just knowing that this type is an interface is enough. No need to store the casted value.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -1879,7 +1879,7 @@ Slice::Python::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
             out << ')';
         }
 
-        // A trailing command to ensure that the outut is interpreted as a Python tuple.
```
InsertCreativityHere, line 1882: Just typos here.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2708,9 +2708,9 @@ namespace
                 {
                     for (const auto& [name, alias] : moduleImports.definitions)
                     {
-                        bool allreadyImported = !allImports.insert(alias.empty() ? name : alias).second;
```
InsertCreativityHere, line 2711: Another typo.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2963,7 +2963,7 @@ Slice::Python::pyLinkFormatter(const string& rawLink, const ContainedPtr&, const
         {
             string targetScoped = dynamic_pointer_cast<Contained>(target)->mappedScoped(".");
             result << ":class:`" << targetScoped;
-            if (auto interfaceTarget = dynamic_pointer_cast<InterfaceDecl>(target))
```
InsertCreativityHere, line 2966: Another unused 'interfaceTarget'.

cpp/src/slice2swift/Gen.h
```diff
@@ -5,9 +5,6 @@
 
 #include "SwiftUtil.h"
 
-using namespace std;
-using namespace Slice;
-
```
InsertCreativityHere, line 10: These `using namespace`s were unused.
Also, we never have `using namespace`s in the header files, only source files.
So I removed these.

cpp/test/Slice/errorDetection/WarningInvalidMetadata.ice
```diff
@@ -76,6 +76,7 @@ module Test
     }
 
     ["cpp:const"] ["cpp:ice_print"]
+    ["deprecate"] ["deprecated"] // You cannot apply both 'deprecate' and 'deprecated' to the same element.
```
InsertCreativityHere, line 79: I added a new test case to make sure you can't mix `deprecated` and `deprecate` together.

cpp/src/slice2matlab/Gen.cpp
InsertCreativityHere: This is in the generated doc-comments, and I don't think we do any testing of the MATLAB doc-comments.

cpp/src/slice2js/Gen.cpp
InsertCreativityHere: Totally fine with me. It seemed unlikely we'd ever get `""` here, but I wasn't confident enough to add an assert.

PR #5161, opened 2026-02-20
Fix DataStorm filter element ID handling and correctness bugs

### LINKED ISSUE #5132: Fix DataStorm correctness bugs: data race, move-assignment leak, double-negation, uninitialized member

pepone:
## Summary

DataStorm uses negative IDs to represent subscriptions to filters, and positive IDs to represent subscriptions to keys. This PR centralizes the sign conversion in `attachFilter` and updates methods that take a filter ID to accept the already-negated (internal) value, instead of performing sign conversions in multiple places.

This PR also includes the following correctness fixes (closes #5141):

- **Data race**: Add missing `lock_guard<mutex>` in `NodeI::forwardToPublishers` which iterates `_publishers` without synchronization, unlike its counterpart `forwardToSubscribers`.
- **Resource leak**: Fix `Node::operator=(Node&&)` to destroy the previous instance before overwriting, preventing leaks of the communicator, adapters, timer, and connection manager.
- **Double-negation**: Fix `DataElementI::disconnect()` passing an already-negated filter element ID to `disconnectFromFilter`, which negates it again — the positive key fails to match filter entries stored under negative keys, silently orphaning the subscription.
- **Uninitialized member**: Initialize `DataReaderI::_instanceCount` to zero to avoid undefined behavior from reading an indeterminate value.

Fixes #5132

## Test plan
- [ ] CI passes
- [ ] DataStorm test suite passes

### DIFF
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -341,7 +341,7 @@ DataElementI::attachFilter(
     // Negate the element ID for internal storage — filter subscriptions use negative IDs to distinguish them from
     // key subscriptions. All subsequent internal functions receive and use this negated ID directly.
     assert(elementId > 0);
-    int64_t filterId = -elementId;
+    const int64_t filterId = -elementId;
 
     bool added = false;
     auto subscriber = p->second.addOrGet(topicId, filterId, subscriberId, filter, sampleFilter, name, priority, added);


cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -338,8 +338,12 @@ DataElementI::attachFilter(
         p = _listeners.emplace(std::move(listenerKey), Listener{std::move(prx), facet}).first;
     }
 
+    // Negate the element ID for internal storage — filter subscriptions use negative IDs to distinguish them from
+    // key subscriptions. All subsequent internal functions receive and use this negated ID directly.
+    int64_t filterId = -elementId;
```
externl: Can we also add a `> 0` assert.

bernardnormier: Can we make this filterId const?

PR #5168, opened 2026-02-23
Fix minor issues in IceDiscovery and IceLocatorDiscovery

pepone:
## Summary

- Add missing `break` after endpoint match in the lookup reply endpoint matching loop in IceDiscovery and IceLocatorDiscovery (C++, C#, Java). The C++ IceDiscovery `LookupI.cpp` already had the correct `break`.
- Remove unused `endpts` variable in IceLocatorDiscovery `PluginI.cpp` that redundantly called `ice_getEndpoints()`.
- Add null check for the `optional` `reply` parameter in IceDiscovery `LookupI`'s `findObjectById` and `findAdapterById`. These are datagram/multicast operations so a null reply is silently ignored.
- Remove unused `#include "Ice/Connection.h"` from IceDiscovery `LookupI.cpp`.
- Remove unused `_name` member from IceLocatorDiscovery `PluginI`.
- Add missing `final` specifier to IceDiscovery `ObjectRequest` class for consistency with `AdapterRequest`.

### DIFF
--- a/.github/workflows/build-brew-packages.yml
+++ b/.github/workflows/build-brew-packages.yml
@@ -40,9 +40,18 @@ jobs:
       - name: Setup C++
         uses: ./.github/actions/setup-cpp
 
+      - name: Download IceGrid GUI JAR
+        id: icegridgui
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh run download --repo ${{ github.repository }} --name icegridgui-macos-app --dir staging
+          icegridgui_dmg_sha256=$(shasum -a 256 "staging/IceGridGUI.dmg" | cut -d ' ' -f 1)
+          echo "icegridgui_dmg_sha256=$icegridgui_dmg_sha256" >> $GITHUB_OUTPUT
+
       - name: Update Nightly Tap
         run: |
-          ./packaging/brew/update-nightly-tap.sh "${{ inputs.channel }}" "${{ inputs.quality }}" "${{ inputs.ice_version }}"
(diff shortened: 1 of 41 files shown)


cpp/src/IceDiscovery/LookupI.cpp
```diff
@@ -228,7 +227,7 @@ LookupI::setLookupReply(const LookupReplyPrx& lookupReply)
 void
 LookupI::findObjectById(string domainId, Ice::Identity id, optional<LookupReplyPrx> reply, const Ice::Current&)
 {
-    if (domainId != _domainId)
+    if (domainId != _domainId || !reply)
```
bernardnormier: I don't think a null reply should be expected and ignored. I would prefer a call to checkNotNull:

https://code.zeroc.com/ice/3.8/api/cpp/namespaceIce_a58c4fc54affa02b4c8203a41d0c18feb.html#a58c4fc54affa02b4c8203a41d0c18feb

Other than logging, it doesn't make a big difference since this is a datagram call.

bernardnormier: We should add the same check in C# and Java.

pepone: do you prefer throwing because this will show up as a dispatch exception?

bernardnormier: Yes. And also because when you read the code, you see that a null proxy is not an expected path.

pepone: Added null checks

PR #5169, opened 2026-02-23
Nightly tap update fixes

externl:
- Calculate DMG SHA256 and pass to update-nightly-tap.sh script
- Remove SHA256 calculation and DMG download from update script
- Fix workflow_dispatch inputs structure in build-icegridgui-macos-app.yml


.github/workflows/build-brew-packages.yml
```diff
@@ -40,9 +40,18 @@ jobs:
       - name: Setup C++
         uses: ./.github/actions/setup-cpp
 
+      - name: Download IceGrid GUI JAR
+        id: icegridgui
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh run download --repo ${{ github.repository }} --name icegridgui-jar-macOS --dir staging
```
pepone: Is this the correct artifact, shouldn't we download the DMG? `icegridgui-macos-app`

PR #5171, opened 2026-02-23
Remove non-current changelogs (main)

bernardnormier:


externl (approved):

externl:
I think we should consider renaming the change log to `CHANGELOG.md`. Instead of including the version in the name.

pepone (approved):

PR #5174, opened 2026-02-23
Fix IceGrid bugs found during code review

pepone:
- Fix LMDB lookup in Database::addOrUpdateObject using wrong variable
- Fix canRemoveServerDirectory always returning false for IceBox servers
- Fix sort no-op in ServerI::updateImpl (begin, begin -> begin, end)
- Fix data race in hasReplicaGroupFilters/hasTypeFilters (missing mutex)
- Add default initializers for uninitialized members in ServerI and DescriptorHandler

### DIFF
--- a/cpp/src/IceGrid/ServerI.h
+++ b/cpp/src/IceGrid/ServerI.h
@@ -156,7 +156,7 @@ namespace IceGrid
         std::optional<Ice::ProcessPrx> _process;
         std::set<std::string> _activatedAdapters;
         std::optional<std::chrono::steady_clock::time_point> _failureTime;
-        ServerActivation _previousActivation{Disabled};
+        ServerActivation _previousActivation{ServerI::Disabled};
         std::shared_ptr<IceInternal::TimerTask> _timerTask;
         bool _waitForReplication{false};
         std::string _stdErrFile;


cpp/src/IceGrid/Database.cpp
```diff
@@ -1556,9 +1556,8 @@ Database::addOrUpdateObject(const ObjectInfo& info, int64_t dbSerial)
         {
             IceDB::ReadWriteTxn txn(_env);
 
-            Ice::Identity k;
             ObjectInfo v;
-            update = _objects.get(txn, k, v);
+            update = _objects.get(txn, id, v);
```
externl, line 1560: What's the consequence of this bug? Seems like it should be more serious.

cpp/src/IceGrid/NodeI.cpp
```diff
@@ -715,10 +715,6 @@ NodeI::canRemoveServerDirectory(const string& name)
         }
         serviceDataDirs.push_back(content);
     }
-    if (!contents.empty())
```
externl, line 718: why is this wrong?

cpp/src/IceGrid/Database.cpp
bernardnormier: Looking at the code, it seems update is never false and addObject throws ObjectExistException. Not clear how this worked.

cpp/src/IceGrid/ServerI.h
```diff
@@ -143,22 +143,22 @@ namespace IceGrid
 
         std::shared_ptr<InternalServerDescriptor> _desc;
 #ifndef _WIN32
-        uid_t _uid;
-        gid_t _gid;
+        uid_t _uid = 0;
+        gid_t _gid = 0;
```
externl: can be use brace initialization?

cpp/src/IceGrid/NodeI.cpp
bernardnormier, line 718: The reason for this deletion is not clear.

cpp/src/IceGrid/Database.cpp
pepone: I will try to add a test. I think you can hit this if a object is dynamically register with a type, and then updated with a different type. The replica would keep both types. With the previous code "update" is always false and the object is never deleted from the _objectsByType. So we keep both types.

Queries to the master are not affected this bug is in the replica database synchronization.

pepone: With the bug, update is always false when a replica receives an object update. The `_objectCache` only contains objects registered via application descriptors — objects added dynamically through the IceGrid admin interface are only stored in the LMDB `_objects` map. So the `_objectCache.has(id)` check on line 1549 passes, and the LMDB lookup (which used a default-constructed identity instead of id) never finds a match.

This means the following block is never reached:

```
if (update)
{
    _objectsByType.del(txn, v.type, v.proxy->ice_getIdentity());
}
```

In practice this doesn't cause data corruption because the public IceGrid admin API doesn't allow changing an object's type — updateObject only updates the proxy. So the _objectsByType index stays consistent even without the del.

The observable effect of the bug is that the trace log (line 1588) reports "added" instead of "updated", and the observer is notified with objectAdded instead of objectUpdated (line 1582 vs 1578).

cpp/src/IceGrid/NodeI.cpp
pepone: At this point, `contents` can only contain `data_` entries (service data directories). Any entry that doesn't start with `data_` causes an early return on line 714.

The removed `if (!contents.empty())` check was preventing the code from ever reaching the `serviceDataDirs` loop on line 736, which explicitly validates each service data directory is empty before allowing removal. This loop was effectively dead code — it was never reached because the `!contents.empty()` guard returned false whenever there were any `data_` entries.

cpp/src/IceGrid/PluginFacadeI.cpp
```diff
@@ -263,6 +263,7 @@ RegistryPluginFacadeI::getReplicaGroupFilters(const string& id) const
 bool
 RegistryPluginFacadeI::hasReplicaGroupFilters() const
 {
+    lock_guard lock(_mutex);
```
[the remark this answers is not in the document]
bernardnormier: We use C++17 or greater.

cpp/src/IceGrid/Database.cpp
bernardnormier: Ok, and addObject (2805) calls `_objects.put` which actually "updates" the entry.

I first thought we call addObject (Database.cpp:1494) which would throw ObjectExistsException for a duplicate entry.

So, as per your comment, the observable effect of this bug is very small.

pepone: Yes. is a minor issue which explain why we didn't notice.

cpp/src/IceGrid/ServerI.h
```diff
@@ -143,22 +143,22 @@ namespace IceGrid
 
         std::shared_ptr<InternalServerDescriptor> _desc;
 #ifndef _WIN32
-        uid_t _uid;
-        gid_t _gid;
+        uid_t _uid{0};
```
bernardnormier: I would rather use `-1` as the default. `0` = super-user, which should not be selected by accident.

Same for _gid.

pepone: Fixed as suggested.

PR #5177, opened 2026-02-24
Fix IceStorm bugs found during code review

pepone:
- Fix deadlock in Observers::wait() by locking _reapedMutex instead of _mutex
- Remove stale outer loop causing N-squared database writes in observerInit()
- Fix wrong wait condition (&&  should be ||) in Subscriber::shutdown()
- Initialize uninitialized _coord member in NodeI
- Hold mutex during _topics iteration in getContent() to prevent data race
- Fix QoS trace output to include key-value pairs in subscribeAndGetPublisher
- Gate unconditional trace behind replication trace level check in Observers::wait()

[icestorm-code-review.md](https://github.com/user-attachments/files/25527543/icestorm-code-review.md)

### DIFF
--- a/cpp/src/IceStorm/Observers.cpp
+++ b/cpp/src/IceStorm/Observers.cpp
@@ -196,8 +196,7 @@ Observers::wait(const string& op)
         if (_traceLevels->replication > 0)
         {
             Ice::Trace out(_traceLevels->logger, _traceLevels->replicationCat);
-            out << "number of observers '" << _observers.size() << "' is less than the majority '" << _majority
-                << "'";
+            out << "number of observers '" << _observers.size() << "' is less than the majority '" << _majority << "'";
         }
         throw Ice::UnknownException(__FILE__, __LINE__, "too few observers");
     }


cpp/src/IceStorm/Observers.cpp
```diff
@@ -183,7 +183,7 @@ Observers::wait(const string& op)
             int id = p->id;
             p = _observers.erase(p);
 
-            lock_guard<mutex> reapedLock(_mutex);
+            lock_guard<mutex> reapedLock(_reapedMutex);
```
pepone, line 186: We lock the wrong mutex, see other uses of `_reaped`.

cpp/src/IceStorm/Observers.cpp
```diff
@@ -193,8 +193,12 @@ Observers::wait(const string& op)
     // If we now no longer have the majority of observers we raise.
     if (_observers.size() < _majority)
     {
-        Ice::Trace out(_traceLevels->logger, _traceLevels->replicationCat);
-        out << "number of observers '" << _observers.size() << "' is less than the majority '" << _majority << "'";
+        if (_traceLevels->replication > 0)
```
pepone, line 196: Unconditional trace without checking replication tracing was enabled.

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -794,10 +794,7 @@ Subscriber::shutdown()
     unique_lock lock(_mutex);
 
     _shutdown = true;
-    while (_outstanding > 0 && !_events.empty())
```
pepone, line 797: This previous condition, only wait if both are true. I think this is supposed to wait for both no outstanding and no events.

cpp/src/IceStorm/TopicI.cpp
```diff
@@ -510,6 +510,7 @@ TopicImpl::subscribeAndGetPublisher(QoS qos, Ice::ObjectPrx obj)
                 {
                     out << ',';
                 }
+                out << '[' << p->first << "," << p->second << ']';
```
pepone: We only output the `,`

cpp/src/IceStorm/TopicManagerI.cpp
```diff
@@ -408,26 +408,25 @@ TopicManagerImpl::observerInit(const LogUpdate& llu, const TopicContentSeq& cont
 
         _subscriberMap.clear(txn);
 
-        for (auto p = content.begin(); p != content.end(); ++p)
-            for (const auto& c : content)
```
pepone, line 412: We have two nested loops, probably introduced during the C++11 port.

cpp/src/IceStorm/TopicManagerI.cpp
```diff
@@ -578,10 +577,8 @@ TopicManagerImpl::observerRemoveSubscriber(const LogUpdate& llu, const string& n
 void
 TopicManagerImpl::getContent(LogUpdate& llu, TopicContentSeq& content)
 {
-    {
-        lock_guard lock(_mutex);
-        reap();
-    }
+    lock_guard lock(_mutex);
+    reap();
```
pepone: We were iterating _topics without holding the mutex, everywhere else we do.

cpp/src/IceStorm/Observers.cpp
bernardnormier: As a result of this bug, this exception handler would always deadlock. Would be nice to add a test that exercises this code path.

For now, can you add a comment in the header file? We must always lock _mutex and _reapedMutex in this order.

cpp/src/IceStorm/TopicI.cpp
bernardnormier: Should this " = ", e.g. `[reliability = ordered]`, `[retryCount = -1]`.

See https://docs.zeroc.com/ice/3.8/cpp/icestorm-quality-of-service

cpp/src/IceStorm/TopicManagerI.cpp
bernardnormier: Should we release `lock` before we create the transaction? As far as I can tell, we no longer need it at that point.

cpp/src/IceStorm/TransientTopicI.cpp
```diff
@@ -150,6 +150,7 @@ TransientTopicImpl::subscribeAndGetPublisher(QoS qos, optional<Ice::ObjectPrx> o
                 {
                     out << ',';
                 }
+                out << '[' << p->first << "," << p->second << ']';
```
bernardnormier: I would use " = " here as well.

cpp/src/IceStorm/TopicI.cpp
```diff
@@ -510,6 +510,7 @@ TopicImpl::subscribeAndGetPublisher(QoS qos, Ice::ObjectPrx obj)
                 {
                     out << ',';
```
bernardnormier: Should be ", "?

cpp/src/IceStorm/TopicManagerI.cpp
externl: why not just lock again when we access topics. Is there a need to hold during this whole function?

PR #5178, opened 2026-02-25
Fix IceBox bugs found during code review

pepone:
## Summary

- Handle `CommunicatorDestroyedException` in observer notifications (`servicesStarted`,
  `servicesStopped`, `addObserver`) in C++, C#, and Java. During shutdown, the observer's communicator
  may already be destroyed when these notifications are sent, causing an unhandled exception. The fix
  wraps the async calls in try/catch blocks.
- Initialize `ServiceInfo::status` to `Stopped` in C++ to avoid undefined behavior from an
  uninitialized enum field.
- Mark C++ `servicesStarted`/`servicesStopped` as `noexcept`.
- Remove stale "Must be called with 'this' unlocked" comments in C#.
- Remove dead `run()` declaration, unused `envName` member, and unused includes in C++.

### DIFF
--- a/cpp/src/IceBox/ServiceManagerI.cpp
+++ b/cpp/src/IceBox/ServiceManagerI.cpp
@@ -828,6 +828,7 @@ IceBox::ServiceManagerI::servicesStarted(
             catch (const CommunicatorDestroyedException&)
             {
                 // Expected during shutdown if the observer's communicator is destroyed.
+                break;
             }
         }
     }
@@ -849,6 +850,7 @@ IceBox::ServiceManagerI::servicesStopped(
             catch (const CommunicatorDestroyedException&)
             {
                 // Expected during shutdown if the observer's communicator is destroyed.
+                break;
             }
         }
     }
--- a/java/src/com.zeroc.icebox/src/main/java/com/zeroc/IceBox/ServiceManagerI.java
+++ b/java/src/com.zeroc.icebox/src/main/java/com/zeroc/IceBox/ServiceManagerI.java
@@ -613,7 +613,7 @@ final class ServiceManagerI implements ServiceManager {
                                 return null;
                             });
                 } catch (CommunicatorDestroyedException ex) {
-                    // Expected during shutdown if the observer's communicator is destroyed.
+                    observerFailed(observer, ex);
                 }
             }
         }
@@ -632,7 +632,7 @@ final class ServiceManagerI implements ServiceManager {
                                 return null;
                             });
                 } catch (CommunicatorDestroyedException ex) {
-                    // Expected during shutdown if the observer's communicator is destroyed.
+                    observerFailed(observer, ex);
                 }
             }
         }


cpp/src/IceBox/ServiceManagerI.cpp
```diff
@@ -798,7 +809,11 @@ IceBox::ServiceManagerI::stopAll()
 
     _services.clear();
 
-    servicesStopped(stoppedServices, _observers);
+    set<ServiceObserverPrx> observers;
+    observers.swap(_observers);
```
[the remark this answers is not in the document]
pepone: fixed

cpp/src/IceBox/ServiceManagerI.cpp
```diff
@@ -165,10 +164,7 @@ IceBox::ServiceManagerI::startService(string name, const Current&)
                 if (started)
                 {
                     service.status = Started;
-
-                    vector<string> services;
-                    services.push_back(name);
-                    servicesStarted(services, _observers);
```
bernardnormier, line 171: I don't understand why we're changing this code. servicesStarted makes async calls (no awaiting) ... it's ok to make such call within a mutex lock.

cpp/src/IceBox/ServiceManagerI.cpp
```diff
@@ -241,10 +246,7 @@ IceBox::ServiceManagerI::stopService(string name, const Current&)
                 if (stopped)
                 {
                     service.status = Stopped;
-
-                    vector<string> services;
-                    services.push_back(name);
-                    servicesStopped(services, _observers);
+                    observers = _observers;
```
bernardnormier: Same here.

cpp/src/IceBox/ServiceManagerI.cpp
```diff
@@ -798,7 +812,11 @@ IceBox::ServiceManagerI::stopAll()
 
     _services.clear();
 
-    servicesStopped(stoppedServices, _observers);
+    // Notify observers without holding _mutex.
```
bernardnormier: Not necessary.

cpp/src/IceBox/ServiceManagerI.cpp
pepone: Can't we deadlock if this throws synchronously?

bernardnormier: servicesStarted throws synchronously if the observers' communicator is destroyed. Not sure this can happen. And then you're concerned that `_conditionalVariable.notify_all()` is not called?

I guess we could fix servicesStarted and friends to ignore CommunicatorDestroyedException? Make them noexcept?

pepone: Updated the changes to just handle  CommunicatorDestroyedException and make the methods noexcept

cpp/src/IceBox/ServiceManagerI.cpp
```diff
@@ -815,25 +813,39 @@ IceBox::ServiceManagerI::makeObserverCompletedCallback(ServiceObserverPrx observ
...
         for (const auto& observer : observers)
         {
-            observer->servicesStartedAsync(services, nullptr, makeObserverCompletedCallback(observer));
+            try
+            {
+                observer->servicesStartedAsync(services, nullptr, makeObserverCompletedCallback(observer));
+            }
+            catch (const CommunicatorDestroyedException&)
+            {
+                // Expected during shutdown if the observer's communicator is destroyed.
```
bernardnormier, line 830: Should we add a break then?

It's not like the other proxies are going to work if the communicator is destroyed.

cpp/src/IceBox/ServiceManagerI.cpp
```diff
@@ -815,25 +813,39 @@ IceBox::ServiceManagerI::makeObserverCompletedCallback(ServiceObserverPrx observ
...
         for (const auto& observer : observers)
         {
-            observer->servicesStoppedAsync(services, nullptr, makeObserverCompletedCallback(observer));
+            try
+            {
+                observer->servicesStoppedAsync(services, nullptr, makeObserverCompletedCallback(observer));
+            }
+            catch (const CommunicatorDestroyedException&)
+            {
+                // Expected during shutdown if the observer's communicator is destroyed.
```
bernardnormier, line 852: Ditto

java/src/com.zeroc.icebox/src/main/java/com/zeroc/IceBox/ServiceManagerI.java
```diff
@@ -206,12 +206,16 @@ public void addObserver(final ServiceObserverPrx observer, Current current) {
...
-                    });
+            try {
+                observer.servicesStartedAsync(activeServices.toArray(new String[0]))
+                    .exceptionally(
+                        ex -> {
+                            observerFailed(observer, ex);
+                            return null;
+                        });
+            } catch (CommunicatorDestroyedException ex) {
+                observerFailed(observer, ex);
```
bernardnormier, line 217: Looking at observerFailed, it sounds like we could/should always call it when `observer.servicesStartedAsync` throws CommunicatorDestroyedException synchronously.

We currently only do it here.

externl (approved):

PR #5179, opened 2026-02-25
Fix IceBridge bugs found during code review

pepone:
## Summary

- Clean up `BridgeConnection` from `_connections` map and notify it of the
  exception when `ice_getConnectionAsync` throws synchronously, preventing
  subsequent dispatches on the same connection from queuing indefinitely
- Replace `cerr` with `consoleErr` for consistent error reporting
- Rename `bc` to `bridgeConnection` for clarity

### DIFF
--- a/cpp/src/IceBridge/IceBridge.cpp
+++ b/cpp/src/IceBridge/IceBridge.cpp
@@ -1,5 +1,6 @@
 // Copyright (c) ZeroC, Inc.
 
+#include "../Ice/ConsoleUtil.h"
 #include "../Ice/Options.h"
 #include "Ice/Connection.h"
 #include "Ice/ObjectAdapter.h"
@@ -11,6 +12,7 @@
 
 using namespace std;
 using namespace Ice;
+using namespace IceInternal;
 
 namespace
 {
@@ -371,7 +373,7 @@ BridgeI::ice_invokeAsync(
     function<void(exception_ptr)> error,
     const Current& current)
 {
-    shared_ptr<BridgeConnection> bc;
+    shared_ptr<BridgeConnection> bridgeConnection;
     {
         lock_guard<mutex> lg(_lock);
 
@@ -397,8 +399,8 @@ BridgeI::ice_invokeAsync(
             //
             target = target->ice_connectionId(Ice::generateUUID());
 
-            bc = make_shared<BridgeConnection>(_adapter, target, current.con);
-            _connections.emplace(current.con, bc);
+            bridgeConnection = make_shared<BridgeConnection>(_adapter, target, current.con);
+            _connections.emplace(current.con, bridgeConnection);
 
             auto self = shared_from_this();
             current.con->setCloseCallback([self](const auto& con) { self->closed(con); });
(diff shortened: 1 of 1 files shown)


cpp/src/IceBridge/IceBridge.cpp
```diff
@@ -413,25 +415,28 @@ BridgeI::ice_invokeAsync(
...
-                    [self, bc](auto ex) { self->outgoingException(bc, ex); });
+                    [self, bridgeConnection](auto outgoing)
+                    { self->outgoingSuccess(bridgeConnection, std::move(outgoing)); },
+                    [self, bridgeConnection](auto ex) { self->outgoingException(bridgeConnection, ex); });
             }
             catch (const std::exception&)
             {
+                _connections.erase(current.con);
+                bridgeConnection->outgoingException(current_exception());
                 error(current_exception());
```
externl, line 426: We should also consider renaming `error` to `logError`

cpp/src/IceBridge/IceBridge.cpp
```diff
@@ -413,25 +415,28 @@ BridgeI::ice_invokeAsync(
...
                 //
                 target->ice_getConnectionAsync(
-                    [self, bc](auto outgoing) { self->outgoingSuccess(bc, std::move(outgoing)); },
-                    [self, bc](auto ex) { self->outgoingException(bc, ex); });
+                    [self, bridgeConnection](auto outgoing)
+                    { self->outgoingSuccess(bridgeConnection, std::move(outgoing)); },
+                    [self, bridgeConnection](auto ex) { self->outgoingException(bridgeConnection, ex); });
             }
             catch (const std::exception&)
             {
```
externl, line 423: I guess we can get `CommunicatorDestroyedException` here?

cpp/src/IceBridge/IceBridge.cpp
```diff
@@ -397,8 +399,8 @@ BridgeI::ice_invokeAsync(
             //
             target = target->ice_connectionId(Ice::generateUUID());
 
-            bc = make_shared<BridgeConnection>(_adapter, target, current.con);
-            _connections.emplace(current.con, bc);
+            bridgeConnection = make_shared<BridgeConnection>(_adapter, target, current.con);
+            _connections.emplace(current.con, bridgeConnection);
```
externl, line 403: We also call `_connections.emplace` in `outgoingSuccess`. Is this correct to be adding it here? 

If we don't have this emplace then we don't need the extra exception handling.

cpp/src/IceBridge/IceBridge.cpp
```diff
@@ -413,25 +415,28 @@ BridgeI::ice_invokeAsync(
...
                 target->ice_getConnectionAsync(
-                    [self, bc](auto outgoing) { self->outgoingSuccess(bc, std::move(outgoing)); },
-                    [self, bc](auto ex) { self->outgoingException(bc, ex); });
+                    [self, bridgeConnection](auto outgoing)
+                    { self->outgoingSuccess(bridgeConnection, std::move(outgoing)); },
+                    [self, bridgeConnection](auto ex) { self->outgoingException(bridgeConnection, ex); });
             }
             catch (const std::exception&)
             {
+                _connections.erase(current.con);
```
bernardnormier, line 424: This looks fine. However, keep in mind the only exception that ice_getConnectionAsync can throw synchronously in CommunicatorDestroyedException.

When the communicator is destroyed, the whole service is going down so this cleanup is not critical.

cpp/src/IceBridge/IceBridge.cpp
pepone: That is a different connection. We keep both incoming and outgoing connections per bridgeConnection in this map.

cpp/src/IceBridge/IceBridge.cpp
pepone: This is the `error` method from the `Ice::Service` base class.

PR #5180, opened 2026-02-25
Fix minor bugs found during code review in IceDB, dsnode, and iceserviceinstall

pepone:
## Summary

- Fix LMDB env handle leak in `IceDB::Env` constructor when `mdb_env_set_maxdbs`, `mdb_env_set_mapsize`, `mdb_env_set_maxreaders`, or `mdb_env_open` fails after `mdb_env_create` succeeds. The destructor doesn't run since the constructor didn't complete, so `mdb_env_close` must be called explicitly before throwing.
- Initialize `Env::_menv` to `nullptr`.
- Use `consoleErr` instead of `cerr` in dsnode, consistent with other Ice services.
- Use `nullptr` instead of `0` for `SID*` pointer initialization in iceserviceinstall.

### DIFF
--- a/cpp/src/dsnode/Node.cpp
+++ b/cpp/src/dsnode/Node.cpp
@@ -13,8 +13,8 @@ usage(const string& n)
 {
     consoleErr << "Usage: " << n << " [options]\n";
     consoleErr << "Options:\n"
-            "-h, --help               Show this message.\n"
-            "-v, --version            Display the DataStorm version.\n";
+                  "-h, --help               Show this message.\n"
+                  "-v, --version            Display the DataStorm version.\n";
 }
 
 int


cpp/src/dsnode/Node.cpp
```diff
@@ -1,18 +1,20 @@
...
+using namespace IceInternal;
 
 static void
 usage(const string& n)
 {
-    cerr << "Usage: " << n << " [options]\n";
-    cerr << "Options:\n"
-            "-h, --help               Show this message.\n"
-            "-v, --version            Display the DataStorm version.\n";
+    consoleErr << "Usage: " << n << " [options]\n";
```
pepone, line 14: We use `consoleErr` in other tools, let's do the same here for consistency. It deals with the windows console encoding.

cpp/src/IceDB/IceDB.cpp
```diff
@@ -103,6 +104,7 @@ Env::Env(const string& path, MDB_dbi maxDbs, size_t mapSize, unsigned int maxRea
         rc = mdb_env_set_mapsize(_menv, mapSize);
         if (rc != MDB_SUCCESS)
         {
+            mdb_env_close(_menv);
             throw LMDBException(__FILE__, __LINE__, rc);
```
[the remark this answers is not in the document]
pepone: I don't think a helper would be clear here.

cpp/src/IceDB/IceDB.cpp
```diff
@@ -79,6 +79,7 @@ Env::Env(const string& path, MDB_dbi maxDbs, size_t mapSize, unsigned int maxRea
         rc = mdb_env_set_maxdbs(_menv, maxDbs);
         if (rc != MDB_SUCCESS)
         {
+            mdb_env_close(_menv);
```
bernardnormier, line 82: These "leaks" are so minor I would not mention them in the changelog.

PR #5188, opened 2026-02-26
Rename internal MSBuild properties with _ prefix in Slice Tools

pepone:
Prefix internal MSBuild properties with `_` to distinguish them from user-configurable properties. Also rename the C# task parameter `IceToolsPath` to `IceSliceToolsPath` for consistency with the C++ targets, and fix the `IceSlIceSliceToolsPath` typo in the C# targets.

### DIFF
--- a/csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
+++ b/csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
@@ -41,6 +41,15 @@ public abstract class SliceCompilerTask : ToolTask
 
     protected abstract string GeneratedExtensions { get; }
 
+    /// <summary>
+    /// Gets the compilation options for a source item. The options correspond to the public task properties that
+    /// affect the compilation command. They are persisted in the generated dependency file (<c>.d</c>) by
+    /// <see cref="ExecuteTool"/>. On subsequent builds, <see cref="SliceDependTask.GetOptions"/> computes the
+    /// current options and compares them with the persisted values to determine whether recompilation is needed.
+    /// Both methods must return the same set of keys for the comparison to work correctly.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
```diff
@@ -45,7 +45,7 @@ protected virtual Dictionary<string, string> GetOptions(ITaskItem item)
     {
         var options = new Dictionary<string, string>
         {
-            ["IceToolsPath"] = IceToolsPath,
+            ["IceSliceToolsPath"] = IceSliceToolsPath,
```
bernardnormier, line 57: A bit confusing here. This option corresponds to a task parameter? Property? 

Would deserve a comment!

csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceDependTask.cs
```diff
@@ -44,7 +44,7 @@ public virtual Dictionary<string, string> GetOptions(ITaskItem item)
     {
         var options = new Dictionary<string, string>
         {
-            ["IceToolsPath"] = IceToolsPath,
+            ["IceSliceToolsPath"] = IceSliceToolsPath,
```
bernardnormier, line 55: Ditto

bernardnormier (approved):

csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
pepone: The MSBuild property `_ IceSliceToolsPath` is used to set the `IceSliceToolsPath` property of the `SliceCompilerTask`. I will add a comment.

Issue #4947, opened 2026-01-26
Generate IcePy Stubfile Directly with `stubgen`


InsertCreativityHere:
Currently, we have a checked in stub-file for the IcePy definitions:
https://github.com/zeroc-ice/ice/tree/main/python/python/IcePy-stubs

My understanding is that once all of the public doc-comments on the IcePy types have been fixed, we can (and should) delete this checked in stub-file, and instead let stubgen regenerate it as needed from the IcePy types, whenever we build the Python docs.

In my recent PRs working on the Python doc-comments, I updated many of these comments, but there are still a couple files that need some work. Once these last couple files have had their comments updated, we should look into how to fix our doc-build pipeline.

externl:
Looked at this while investigating the duplicated doc comments, and it is more involved than fixing the remaining ones.

Running `stubgen -m IcePy --include-docstrings` against a built IcePy and swapping the result in gives **920 pyright errors**. Adding `--include-private` drops that to **106** — the difference is the `_t_*` Slice type markers, which stubgen skips as private and which 818 of those errors depend on.

The remaining 106 break down as:

| Count | Cause |
|---:|---|
| 56 | `*args/**kwargs` on 29 methods stubgen cannot infer |
| 42 | `typing` imports it does not emit (`Awaitable`, `Self`, `Type`) |
| 3 | `_t_ObjectPrx`, `_t_Value`, `_t_UnknownSlicedValue`, still missed |
| 5 | an `ice_identity` override mismatch and a `Properties` type mismatch |

Of the 29 it cannot infer, 27 are `__init__` — but only four are real constructors. The stub deliberately omits `__init__` for `Connection`, `Endpoint`, `Logger` and the rest, and stubgen invents one for each carrying Python's own `"Initialize self. See help(type(self)) for accurate signature."`.

The signatures themselves are not the problem: stubgen recovers 200 of 211 and all 26 docstrings, because its C extension mode parses the `close() -> Awaitable[None]` first line we already write. Giving a type's `tp_doc` a signature line recovers its constructor too — confirmed on `Properties`.

So adopting this needs a post-processing step that injects the missing imports, deletes the invented `__init__`s and their boilerplate, lifts signature lines back out of the class docstrings, and hand-declares three markers. That is a script whose purpose is undoing stubgen's output choices, and it will need revisiting whenever mypy changes them.

[...]

PR #5190, opened 2026-02-27
Fixes to IcePy Doc Signatures

InsertCreativityHere:
Seems like implementing #4947 is actually going to be a pain, so I'm shelving it.
But, I noticed some inconsistencies while working on it, so this PR fixes those mistakes.

Most of these are actual bugs which were being hidden by our hand-altered stub-file, and I only found once I tried to use a freshly generated stub-file.

### DIFF
--- a/python/modules/IcePy/Proxy.cpp
+++ b/python/modules/IcePy/Proxy.cpp
@@ -1430,7 +1430,7 @@ static PyMethodDef ProxyMethods[] = {
     {"newProxy",
      reinterpret_cast<PyCFunction>(proxyNewProxy),
      METH_VARARGS | METH_STATIC,
-     PyDoc_STR("newProxy(type: Type[T], proxy: Ice.ObjectPrx) -> proxy")},
+     PyDoc_STR("newProxy(type: Type[T], proxy: Ice.ObjectPrx) -> T")},
     {} /* sentinel */
 };
 


python/modules/IcePy/BatchRequestInterceptor.cpp
```diff
@@ -41,8 +41,8 @@ Ice.ObjectPrx
 Queues this request.)";
 
     constexpr const char* BatchRequestType_doc =
-        R"(Represents a batch request. A batch request is created by invoking an operation on a batch-oneway or
-batch-datagram proxy.)";
+        R"(Represents a batch request.
+A batch request is created by invoking an operation on a batch-oneway or batch-datagram proxy.)";
```
InsertCreativityHere, line 45: Updated to match the formatting in our `__init__.pyi` file.
Also just looks better IMO.

python/modules/IcePy/Communicator.cpp
```diff
@@ -1500,8 +1500,11 @@ static PyMethodDef CommunicatorMethods[] = {
     {"_setWrapper",
      reinterpret_cast<PyCFunction>(communicatorSetWrapper),
      METH_VARARGS,
```
InsertCreativityHere, line 1502: Updated the doc comments from `internal function` to function signatures `stubgen` can use.

python/modules/IcePy/Communicator.cpp
```diff
@@ -1392,7 +1392,7 @@ static PyMethodDef CommunicatorMethods[] = {
     {"shutdownCompleted",
      reinterpret_cast<PyCFunction>(communicatorShutdownCompleted),
      METH_NOARGS,
```
InsertCreativityHere, line 1394: We always prefer `Awaitable[T]` over our own `Future` type.
We were already using `Awaitable[T]` in our hand-fixed `__init__.pyi` file.

python/modules/IcePy/Init.cpp
```diff
@@ -201,24 +201,42 @@ static PyMethodDef methods[] = {
      reinterpret_cast<PyCFunction>(IcePy_setProcessLogger),
      METH_VARARGS,
      PyDoc_STR(IcePy_setProcessLogger_doc)},
```
InsertCreativityHere, line 203: Also updated all of these doc-comments from `internal function` to real type-signatures.
As of this PR, all the `internal function` comments are gone, and _everything_ in IcePy has a doc-comment which works with `stubgen`.

python/modules/IcePy/Proxy.cpp
```diff
@@ -1256,36 +1256,36 @@ static PyMethodDef ProxyMethods[] = {
     {"ice_identity",
      reinterpret_cast<PyCFunction>(proxyIceIdentity),
      METH_VARARGS,
```
InsertCreativityHere, line 1258: We use `Self` in `__init__.pyi` which feels more correct to me.
Also updated some parameter names to match the `__init__.pyi`.

python/modules/IcePy/ObjectAdapter.cpp
```diff
@@ -1504,35 +1504,35 @@ static PyMethodDef AdapterMethods[] = {
     {"findDefaultServant",
      reinterpret_cast<PyCFunction>(adapterFindDefaultServant),
      METH_VARARGS,
```
InsertCreativityHere, line 1506: Some of these signatures have a type, but no parameter name. Others have a parameter name, but no type.
Not sure why, but all of these parameter names were updated to have both a name and type.

python/modules/IcePy/ObjectAdapter.cpp
```diff
@@ -1544,11 +1544,11 @@ static PyMethodDef AdapterMethods[] = {
     {"getPublishedEndpoints",
      reinterpret_cast<PyCFunction>(adapterGetPublishedEndpoints),
      METH_NOARGS,
-     PyDoc_STR("getPublishedEndpoints() -> None")},
```
InsertCreativityHere, line 1547: This one was just flat-out wrong. Who knows why.

python/modules/IcePy/Proxy.cpp
```diff
@@ -1439,7 +1430,7 @@ static PyMethodDef ProxyMethods[] = {
     {"newProxy",
      reinterpret_cast<PyCFunction>(proxyNewProxy),
      METH_VARARGS | METH_STATIC,
-     PyDoc_STR("newProxy(proxyType, proxy) -> proxy")},
+     PyDoc_STR("newProxy(type: Type[T], proxy: Ice.ObjectPrx) -> proxy")},
```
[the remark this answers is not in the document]
externl: Return type still needs fixing.

Issue #5074, opened 2026-02-14
Duplicate error handling in BT StreamSocket


pepone:
https://github.com/zeroc-ice/ice/blob/c6a76c9f8ef34e2fd7f7b2957fa1e0e418400d0a/cpp/src/IceBT/StreamSocket.cpp#L128-L135

Seems the else should throw `SocketException`.

Issue #5073, opened 2026-02-14
Bogus access to DBus services


pepone:
https://github.com/zeroc-ice/ice/blob/c6a76c9f8ef34e2fd7f7b2957fa1e0e418400d0a/cpp/src/IceBT/DBus.cpp#L1084-L1085

**Current Implementation:**

```cpp
DBusHandlerResult handleMessage(DBusMessage* m)
{
    vector<FilterPtr> filters;
    map<string, ServicePtr> services;
    {
        lock_guard lock(_mutex);
        filters = _filters;
        services = _services;  // Copy for thread safety
    }

    // ...filter handling...

    if (msg->isMethodCall())
    {
        auto p = services.find(msg->getPath());
        if (p != _services.end())  // BUG: Should be 'services', not '_services'!
        {
            // ...
        }
    }
}
```

**Issue:**

- The code copies `_services` to local `services` for thread safety
- But then accesses `_services.end()` instead of `services.end()`
- This is a data race - accessing `_services` without the lock

PR #5198, opened 2026-03-02
Fix IceBT bugs backported from 3.8

pepone:
## Summary

- Fix data race in DBus `handleMessage` where `_services` (shared member) was used instead of the local `services` copy outside the lock (#5073)
- Fix multiple bugs in IceBT DBus layer (#5140): shadowing `reason` member losing error messages, `bool` vs `dbus_bool_t` stack corruption in `dbus_message_iter_get_basic`, and `VariantValue::clone()` returning an alias instead of a deep copy
- Fix `ConnectorI::operator<` missing early returns causing incorrect ordering when `_addr` matches but subsequent fields differ (#5139)
- Fix duplicate error handling in BT `StreamSocket` read (#5074)

### DIFF
--- a/cpp/src/IceBT/DBus.h
+++ b/cpp/src/IceBT/DBus.h
@@ -268,7 +268,7 @@ public:
 
     virtual ValuePtr clone() const
     {
-        return new VariantValue(v ? v->clone() : 0);
+        return new VariantValue(v ? v->clone() : nullptr);
     }
 
     virtual std::string toString() const


cpp/src/IceBT/DBus.h
```diff
@@ -268,7 +268,7 @@ class VariantValue : public Value
 
     virtual ValuePtr clone() const
     {
-        return const_cast<VariantValue*>(this);
```
bernardnormier, line 271: Why is this code incorrect?

pepone: We do this deep cloning for other types, struct, arrays, etc. It seems more logically to also do the same for VariantValue.

Cloning a VariantValue should be equivalent to cloning the contained type.

PR #5203, opened 2026-03-02
Initialize SSL member variables to prevent undefined behavior (#5105)

pepone:
Backport from 972b0c9e1b6f55f79fa495ca6fb411f7168a4c27


externl (approved):

cpp/src/IceSSL/OpenSSLTransceiverI.h
```diff
@@ -70,8 +70,8 @@ class TransceiverI : public IceInternal::Transceiver
     bool _connected;
     std::string _cipher;
     std::vector<IceSSL::CertificatePtr> _certs;
-    bool _verified;
-    TrustError _trustError;
+    bool _verified{false};
```
bernardnormier: I don't think we can use this syntax with old C++ compilers.

cpp/src/IceSSL/SChannelTransceiverI.h
```diff
@@ -119,8 +119,8 @@ class TransceiverI : public IceInternal::Transceiver
     SecPkgContext_StreamSizes _sizes;
     std::string _cipher;
     std::vector<IceSSL::CertificatePtr> _certs;
-    bool _verified;
-    TrustError _trustError;
+    bool _verified{false};
```
bernardnormier: Ditto

bernardnormier (commented):
I don't think this fix is compatible with older C++ compilers supported by Ice 3.7.

cpp/src/IceSSL/OpenSSLTransceiverI.h
pepone: Updated the code for C++98 compatibility.

PR #5206, opened 2026-03-02
Add --rpc none|ice|icerpc option to slice2cs

bernardnormier:
This PR updates slice2cs to generate code for either Ice (the default) or IceRPC.

The new option is as follows:
   --rpc none: generate code IceRPC's ZeroC.Slice (no IceRpc dependency; used to build the IceRpc core assembly)
   --rpc ice: generate code for Ice (the default)
   --rpc icerpc: generate code for the IceRPC+Slice integration (depends on IceRpc.Slice)

When you specify `--rpc icerpc`, slice2cs generates two files for each Filename.ice source file: Filename.cs and Filename.IceRpc.cs; otherwise, it only generates Filename.cs.

This PR also updates the C# ZeroC.Ice.Slice.Tools with a new `Rpc` option for the SliceCompiler and SliceDepend tasks.

You would typically enable it with:
```
  <ItemDefinitionGroup>
    <SliceCompile>
      <Rpc>icerpc</Rpc>
    </SliceCompile>
  </ItemDefinitionGroup>
```

The new generated code is largely the same as the code generated by slicec-cs in Slice1 mode.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -2827,6 +2827,24 @@ Slice::InterfaceDef::operations() const
 }
 
 OperationList
+Slice::InterfaceDef::allOperations() const
+{
+    OperationList result = allInheritedOperations();
+
+    for (const auto& q : operations())
+    {
+        if (none_of(
+                result.begin(),
+                result.end(),
+                [name = q->name()](const auto& other) { return other->name() == name; }))
+        {
+            result.push_back(q);
+        }
+    }
+    return result;
+}
+
+OperationList
 Slice::InterfaceDef::allInheritedOperations() const
 {
     OperationList result;
@@ -2841,24 +2859,6 @@ Slice::InterfaceDef::allInheritedOperations() const
             {
                 result.push_back(q);
             }
-        }
-    }
-    return result;
-}
-
-OperationList
-Slice::InterfaceDef::allOperations() const
-{
-    OperationList result = allInheritedOperations();
-
-    for (const auto& q : operations())
-    {
-        if (none_of(
-                result.begin(),
-                result.end(),
-                [name = q->name()](const auto& other) { return other->name() == name; }))
-        {
-            result.push_back(q);
         }
     }
     return result;
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -2738,8 +2738,9 @@ Slice::InterfaceDef::createOperation(
                 break;
             }
         }
-        // Check the operations of the Object pseudo-interface.
-        if (!checkBaseOperationNames(name, {"ice_id", "ice_ids", "ice_ping", "ice_isA"}))
+
+        // Check the operations of the Object pseudo-interface unless we're generating code for Object itself.
+        if (scoped() != "::Ice::Object" && !checkBaseOperationNames(name, {"ice_id", "ice_ids", "ice_ping", "ice_isA"}))
         {
             hasConflictingIdentifier = true;
         }
@@ -2826,7 +2827,7 @@ Slice::InterfaceDef::operations() const
 }
 
 OperationList
-Slice::InterfaceDef::allOperations() const
+Slice::InterfaceDef::allInheritedOperations() const
 {
     OperationList result;
     for (const auto& p : _bases)
@@ -2842,6 +2843,13 @@ Slice::InterfaceDef::allOperations() const
             }
         }
     }
+    return result;
+}
+
+OperationList
+Slice::InterfaceDef::allOperations() const
+{
+    OperationList result = allInheritedOperations();
 
     for (const auto& q : operations())
     {
--- a/cpp/src/slice2cs/CsUtil.cpp
+++ b/cpp/src/slice2cs/CsUtil.cpp
@@ -34,12 +34,12 @@ Slice::Csharp::getNamespace(const ContainedPtr& p)
 }
 
 string
-Slice::Csharp::getUnqualified(const ContainedPtr& p, const string& ns, const string& prefix)
+Slice::Csharp::getUnqualified(const ContainedPtr& p, const string& ns, const string& prefix, const string& suffix)
 {
     string name = p->mappedName();
-    if (!prefix.empty())
+    if (!prefix.empty() || !suffix.empty())
     {
-        name = prefix + removeEscapePrefix(name);
+        name = prefix + removeEscapePrefix(name) + suffix;
     }
 
     // If contained is an operation, a field, or an enumerator, we use the enclosing type.
@@ -119,5 +119,84 @@ Slice::Csharp::writeDocLines(
             out << line;
         }
         out << "</" << closeTag.value_or(openTag) << ">";
+    }
+}
+
+Slice::Csharp::CsharpDocCommentFormatter::CsharpDocCommentFormatter(
+    function<pair<bool, string>(const string&, const ContainedPtr&, const SyntaxTreeBasePtr&)> linkFormatter)
+    : _linkFormatter(std::move(linkFormatter))
+{
+}
+
+void
+Slice::Csharp::CsharpDocCommentFormatter::preprocess(StringList& rawComment)
+{
+    for (auto& line : rawComment)
+    {
+        // Escape any XML special characters in the comment.
+        string::size_type pos = 0;
+        while ((pos = line.find_first_of("&<>", pos)) != string::npos)
+        {
+            switch (line[pos])
+            {
+                case '&':
+                    line.replace(pos, 1, "&amp;");
+                    break;
+                case '<':
+                    line.replace(pos, 1, "&lt;");
+                    break;
+                case '>':
+                    line.replace(pos, 1, "&gt;");
+                    break;
+            }
+            // Skip over the leading '&' character to avoid 'find'ing it again.
+            pos += 1;
+        }
+    }
+}
+
+string
+Slice::Csharp::CsharpDocCommentFormatter::formatCode(const string& rawText)
+{
+    return "<c>" + rawText + "</c>";
+}
+
+string
+Slice::Csharp::CsharpDocCommentFormatter::formatParamRef(const string& param)
+{
+    return "<paramref name=\"" + param + "\" />";
+}
+
+string
+Slice::Csharp::CsharpDocCommentFormatter::formatLink(
+    const string& rawLink,
+    const ContainedPtr& source,
+    const SyntaxTreeBasePtr& target)
+{
+    auto [mapToLink, qualifiedName] = _linkFormatter(rawLink, source, target);
+    if (mapToLink)
+    {
+        return "<see " + qualifiedName + " />";
+    }
+    else
+    {
+        return "<c>" + qualifiedName + "</c>";
+    }
+}
+
+string
+Slice::Csharp::CsharpDocCommentFormatter::formatSeeAlso(
+    const string& rawLink,
+    const ContainedPtr& source,
+    const SyntaxTreeBasePtr& target)
+{
+    auto [mapToLink, qualifiedName] = _linkFormatter(rawLink, source, target);
+    if (mapToLink)
+    {
+        return "<seealso " + qualifiedName + " />";
+    }
+    else
+    {
+        return "";
     }
 }
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -63,6 +63,7 @@ Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool ena
 
     if (_genMode == GenMode::Ice)
     {
+        _out << sp;
         _out << nl << "[assembly:Ice.Slice(\"" << fileBase << ".ice\")]";
     }
     else
@@ -78,7 +79,6 @@ Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool ena
             _iceRpcOut << nl << "using ZeroC.Slice;";
             _iceRpcOut << nl << "using IceRpc.Slice;";
             _iceRpcOut << sp;
-            _iceRpcOut << nl << "[assembly:Slice(\"" << fileBase << ".ice\")]";
         }
     }
 }
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -7,6 +7,7 @@
 #include "CsMetadataValidator.h"
 #include "CsUtil.h"
 #include "Ice/StringUtil.h"
+#include "IceRpcVisitors.h"
 #include "IceVisitors.h"
 
 using namespace std;
@@ -14,7 +15,9 @@ using namespace Slice;
 using namespace Slice::Csharp;
 using namespace IceInternal;
 
-Slice::Gen::Gen(const string& base, const string& dir, bool enableAnalysis) : _enableAnalysis(enableAnalysis)
+Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool enableAnalysis)
+    : _genMode(genMode),
+      _enableAnalysis(enableAnalysis)
 {
     string fileBase = base;
     string::size_type pos = base.find_last_of("/\\");
@@ -23,9 +26,12 @@ Slice::Gen::Gen(const string& base, const string& dir, bool enableAnalysis) : _e
         fileBase = base.substr(pos + 1);
     }
     string file = fileBase + ".cs";
+    string iceRpcFile = fileBase + ".IceRpc.cs";
+
     if (!dir.empty())
     {
         file = dir + '/' + file;
+        iceRpcFile = dir + '/' + iceRpcFile;
     }
 
     _out.open(file.c_str());
@@ -36,49 +42,58 @@ Slice::Gen::Gen(const string& base, const string& dir, bool enableAnalysis) : _e
         throw FileException(os.str());
     }
     FileTracker::instance()->addFile(file);
-    printHeader();
 
-    if (!_enableAnalysis)
+    if (_genMode == GenMode::IceRpc)
     {
-        printGeneratedHeader(_out, fileBase + ".ice");
+        _iceRpcOut.open(iceRpcFile.c_str());
+        if (!_iceRpcOut)
+        {
+            ostringstream os;
+            os << "cannot open '" << iceRpcFile << "': " << IceInternal::errorToString(errno);
+            throw FileException(os.str());
+        }
+        FileTracker::instance()->addFile(iceRpcFile);
     }
 
-    _out << sp;
-    _out << nl << "#nullable enable";
-    _out << sp;
-    _out << nl << "[assembly:Ice.Slice(\"" << fileBase << ".ice\")]";
-
-    _out << sp;
-
-    if (_enableAnalysis)
+    printHeader(_out, fileBase + ".ice", _enableAnalysis);
+    if (_genMode == GenMode::IceRpc)
     {
-        // Disable some warnings when auto-generated is removed from the header. See printGeneratedHeader above.
-        _out << nl << "#pragma warning disable SA1403 // File may only contain a single namespace";
-        _out << nl << "#pragma warning disable SA1611 // The documentation for parameter x is missing";
-
-        _out << nl << "#pragma warning disable CA1041 // Provide a message for the ObsoleteAttribute that marks ...";
-        _out << nl << "#pragma warning disable CA1068 // Cancellation token as last parameter";
-        _out << nl << "#pragma warning disable CA1725 // Change parameter name istr_ to istr in order to match ...";
-
-        // Missing doc - only necessary for the tests.
-        _out << nl << "#pragma warning disable SA1602";
-        _out << nl << "#pragma warning disable SA1604";
-        _out << nl << "#pragma warning disable SA1605";
+        printHeader(_iceRpcOut, fileBase + ".ice", _enableAnalysis);
     }
 
-    _out << nl << "#pragma warning disable CS1591 // Missing XML Comment";
-    _out << nl << "#pragma warning disable CS1573 // Parameter has no matching param tag in the XML comment";
-    _out << nl << "#pragma warning disable CS0612 // Type or member is obsolete";
-    _out << nl << "#pragma warning disable CS0618 // Type or member is obsolete";
-    _out << nl << "#pragma warning disable CS0619 // Type or member is obsolete";
+    if (_genMode == GenMode::Ice)
+    {
+        _out << nl << "[assembly:Ice.Slice(\"" << fileBase << ".ice\")]";
+    }
+    else
+    {
+        _out << sp;
+        _out << nl << "using ZeroC.Slice;";
+        _out << sp;
+        _out << nl << "[assembly:Slice(\"" << fileBase << ".ice\")]";
+
+        if (_genMode == GenMode::IceRpc)
+        {
+            _iceRpcOut << sp;
+            _iceRpcOut << nl << "using ZeroC.Slice;";
+            _iceRpcOut << nl << "using IceRpc.Slice;";
+            _iceRpcOut << sp;
+            _iceRpcOut << nl << "[assembly:Slice(\"" << fileBase << ".ice\")]";
+        }
+    }
 }
 
 Slice::Gen::~Gen()
 {
     if (_out.isOpen())
     {
-        _out << '\n';
+        _out << nl;
         _out.close();
+    }
+    if (_iceRpcOut.isOpen())
+    {
+        _iceRpcOut << nl;
+        _iceRpcOut.close();
     }
 }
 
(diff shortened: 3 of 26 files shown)


cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -0,0 +1,671 @@
...
+    SequencePtr seq = dynamic_pointer_cast<Sequence>(type);
+    bool readOnlyMemory = seq && hasFixedSizeBuiltinElements(seq) && !seq->hasMetadata("cs:generic") &&
+                          context == TypeContext::OutgoingParam;
+
+    string tagFormat = getTagFormat(type);
+
+    if (readOnlyMemory)
+    {
+        out << nl << "if (" << fieldName << ".Span != null)";
+    }
```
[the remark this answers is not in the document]
bernardnormier: This is what we currently generate in slicec-cs. We should make sure we have a test for this mapping.

csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
```diff
@@ -89,6 +94,11 @@ protected override string GenerateCommandLineCommands()
             builder.AppendSwitchIfNotNull("-I", Path.GetFullPath(path));
         }
 
+        if (Rpc != "ice")
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested

csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
```diff
@@ -27,6 +27,11 @@ public abstract class SliceCompilerTask : ToolTask
 
     public string[] IncludeDirectories { get; set; } = Array.Empty<string>();
 
+    /// <summary>
+    /// The RPC provider to generate code for, corresponds to the <c>--rpc</c> compiler option.
+    /// </summary>
+    public string Rpc { get; set; } = "ice";
+
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

csharp/src/ZeroC.Ice.Slice.Tools/Slice2CSharpTask.cs
```diff
@@ -22,8 +22,14 @@ protected override void TraceGenerated()
...
+        Rpc == "icerpc" ?
+            new ITaskItem[]
+            {
+                new TaskItem(GetGeneratedPath(source, OutputDir, ".cs")),
+                new TaskItem(GetGeneratedPath(source, OutputDir, ".IceRpc.cs"))
+            } :
+            new ITaskItem[]
+            {
+                new TaskItem(GetGeneratedPath(source, OutputDir, ".cs"))
+            };
```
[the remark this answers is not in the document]
bernardnormier: icerpc is (will be) tested in icerpc-csharp.

csharp/src/ZeroC.Ice.Slice.Tools/SliceCompile.CSharp.xaml
```diff
@@ -49,6 +49,12 @@
         </StringListProperty.DataSource>
     </StringListProperty>
 
+    <EnumProperty Name="Rpc" DisplayName="RPC Provider" Description="Generate code for the specified RPC framework.">
+        <EnumValue Name="none" DisplayName="None" />
+         <EnumValue Name="ice" DisplayName="Ice" />
+        <EnumValue Name="icerpc" DisplayName="IceRPC" />
+    </EnumProperty>
```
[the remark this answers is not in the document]
bernardnormier: Fixed as suggested.

cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -0,0 +1,1486 @@
...
+        {
+            _out << nl << "payload: Request.Encode" << operationName;
+            _out << spar;
+            for (const auto& param : operation->inParameters())
+            {
+                _out << param->mappedName();
+            }
+            // TODO: escape encodeOptions
+            _out << "encodeOptions: EncodeOptions";
+            _out << epar;
```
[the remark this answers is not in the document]
bernardnormier: The description of the issue is correct. Should be fixed in a follow-up PR.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -34,12 +34,12 @@ Slice::Csharp::getNamespace(const ContainedPtr& p)
 }
 
 string
-Slice::Csharp::getUnqualified(const ContainedPtr& p, const string& ns, const string& prefix)
+Slice::Csharp::getUnqualified(const ContainedPtr& p, const string& ns, const string& prefix, const string& suffix)
 {
     string name = p->mappedName();
-    if (!prefix.empty())
+    if (!prefix.empty() || !suffix.empty())
```
bernardnormier, line 40: Whenever we have a prefix or suffix, we need to remove the escape prefix (`@`).

cpp/src/slice2cs/Gen.cpp
```diff
@@ -36,76 +42,134 @@ Slice::Gen::Gen(const string& base, const string& dir, bool enableAnalysis) : _e
...
+    {
+        Slice::IceRpc::TypesVisitor typesVisitor(_out);
+        p->visit(&typesVisitor);
 
-    // Default skeleton.
-    Slice::Ice::SkeletonVisitor skeletonVisitor(_out, false);
-    p->visit(&skeletonVisitor);
+        if (_genMode == GenMode::IceRpc)
+        {
+            Slice::IceRpc::ProxyVisitor proxyVisitor(_iceRpcOut);
```
bernardnormier, line 128: This generated code depends on IceRpc and IceRpc.Slice.

cpp/src/slice2cs/IceCsUtil.cpp
```diff
@@ -83,7 +83,7 @@ Slice::Csharp::typeToString(const TypePtr& type, const string& ns, bool optional
     InterfaceDeclPtr proxy = dynamic_pointer_cast<InterfaceDecl>(type);
     if (proxy)
     {
-        return getUnqualified(proxy, ns) + "Prx?";
+        return getUnqualified(proxy, ns, "", "Prx") + "?";
```
bernardnormier, line 86: We want to include the Prx in the getUnqualified code to properly remove the "@" escape.

csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
```diff
@@ -27,6 +27,11 @@ public abstract class SliceCompilerTask : ToolTask
 
     public string[] IncludeDirectories { get; set; } = Array.Empty<string>();
 
+    /// <summary>
+    /// The RPC provider to generate code for, corresponds to the <c>--rpc</c> compiler option.
+    /// </summary>
+    public string Rpc { get; set; } = "ice";
```
bernardnormier, line 33: We can't just rely on AdditionalOptions: the tasks need to know if the generated code is .cs or .cs + .IceRpc.cs.

cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -0,0 +1,1486 @@
...
+    _out << sp;
+    _out << nl << "/// <inheritdoc/>";
+    _out << nl << "public SliceEncodeOptions? EncodeOptions { get; init; }";
+    _out << sp;
+    _out << nl << "/// <inheritdoc/>";
+    _out << nl << "public required IceRpc.IInvoker Invoker { get; init; }";
+    _out << sp;
+    _out << nl << "/// <inheritdoc/>";
+    _out << nl << "public IceRpc.ServiceAddress ServiceAddress { get; init; } = _defaultServiceAddress;";
+    _out << sp;
```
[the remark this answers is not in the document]
bernardnormier: This is the intended behavior / API. If you use this parameterless constructor, you must set Invoker immediately.

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -0,0 +1,671 @@
...
+        // Proxy types are mapped the same way for optional and non-optional types.
+        return csFieldType(type, ns) + "?";
+    }
+    // else, just use the regular mapping.
+
+    static const char* builtinTable[] =
+        {"byte", "bool", "short", "int", "long", "float", "double", "string", "IceRpc.ServiceAddress?", "SliceClass?"};
+
+    BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(type);
+    if (builtin)
```
pepone: Can we  use?

```suggestion
    if ( BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(type);)
```

and same for others below.

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -0,0 +1,671 @@
...
+        return csFieldType(seq->type(), ns) + "[]";
+    }
+
+    DictionaryPtr d = dynamic_pointer_cast<Dictionary>(type);
+    if (d)
+    {
+        string typeName = d->getMetadataArgs("cs:generic").value_or("Dictionary");
+        return "global::System.Collections.Generic." + typeName + "<" + csFieldType(d->keyType(), ns) + ", " +
+               csFieldType(d->valueType(), ns) + ">";
+    }
```
pepone: I think is clear to do
```
if (SequencePtr seq = dynamic_pointer_cast<Sequence>(type))
{
    ...
}
else if (DictionaryPtr d = dynamic_pointer_cast<Dictionary>(type))
{
    ...
}
```

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -0,0 +1,671 @@
...
+            return "global::System.Collections.Generic.IEnumerable<" + elementTypeStr + ">" + (optional ? "?" : "");
+        }
+    }
+
+    DictionaryPtr d = dynamic_pointer_cast<Dictionary>(type);
+    if (d)
+    {
+        return "global::System.Collections.Generic.IEnumerable<global::System.Collections.Generic.KeyValuePair<" +
+               csFieldType(d->keyType(), ns) + ", " + csFieldType(d->valueType(), ns) + ">>" + (optional ? "?" : "");
+    }
```
pepone: same here

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -0,0 +1,671 @@
...
+    if (field->optional())
+    {
+        return false;
+    }
+
+    BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(field->type());
+    if (builtin)
+    {
+        return builtin->kind() == Builtin::KindString;
+    }
```
pepone: use the `isString` helper, defined above.

cpp/src/slice2cs/IceRpcCsUtil.cpp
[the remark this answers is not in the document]
pepone: Seems this just check for empty, and relies on implicit conversion

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2265

`OperationTests.IceRpc.cs` uses this null checks

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -0,0 +1,671 @@
...
+    out << nl << fieldName << (isCsValueType(type) ? ".Value" : "") << ",";
+    out << nl << "(ref SliceEncoder encoder, " << csType(type, ns, context) << " value) => ";
+    out.inc();
+    encodeField(out, "value", type, ns, context, "encoder");
+    out.dec();
+    out << ");";
+    out.dec();
+
+    out << eb;
+    // else, don't encode anything for null.
```
pepone, line 418: Should this `// else` go into the generated code, it matches the `if` that we generate for check the value is present.

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -0,0 +1,671 @@
...
+            result << getUnqualified(interfaceTarget, sourceScope, "", "Proxy");
+        }
+        else
+        {
+            result << getUnqualified(contained, sourceScope);
+        }
+        result << "\"";
+    }
+    else
+    {
```
pepone, line 595: Not clear what are we dealing with in here.

csharp/src/ZeroC.Ice.Slice.Tools/SliceCompile.CSharp.xaml
```diff
@@ -49,6 +49,20 @@
         </StringListProperty.DataSource>
     </StringListProperty>
 
+    <EnumProperty Name="Rpc" DisplayName="RPC Provider" Description="Generate code for the specified RPC framework.">
+        <EnumValue Name="none" DisplayName="None" />
```
pepone, line 53: EnumValue supports Description property, Visual Studio will display this in the editor for the field.

csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
```diff
@@ -64,6 +69,12 @@ protected virtual Dictionary<string, string> GetOptions(ITaskItem item)
             options["IncludeDirectories"] = value;
         }
 
+        value = item.GetMetadata("Rpc");
```
pepone, line 72: Did you consider putting this and the `Rpc` property in the derived `SliceCSharpTask` that handles C#? The C++ tasks derive from this common class but don't support the `Rpc` setting.

cpp/src/slice2cs/Main.cpp
```diff
@@ -169,12 +201,30 @@ compile(const vector<string>& argv)
...
             FILE* preprocessedHandle = preprocessor->preprocess("-D__SLICE2CS__");
+
             if (preprocessedHandle == nullptr)
             {
                 return EXIT_FAILURE;
             }
 
-            unit = Unit::createUnit("cs");
+            UnitOptions unitOptions{};
+            if (genMode != Slice::GenMode::Ice)
```
pepone: Above we use `if (genMode == Slice::GenMode::None || genMode == Slice::GenMode::IceRpc)` which is equivalent to this. We should be consistent

cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -0,0 +1,1487 @@
...
+                        << (csIncomingParamType(param->type(), ns, param->optional()) + " " +
+                            toPascalCase(param->mappedName()));
+                }
+                out << epar;
+            }
+            out << '>';
+        }
+    }
+
+    void writeMethod(IceInternal::Output& out, const OperationPtr& operation, const std::string& ns, bool dispatch)
```
pepone: What about naming this `writeMethodDeclaration` and adding a comment.

cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -0,0 +1,1487 @@
...
+    }
+
+    // My operations
+    for (const auto& operation : p->operations())
+    {
+        _out << sp;
+        _out << nl << "/// <inheritdoc/>";
+        _out << nl << "public ";
+
+        string operationName = removeEscapePrefix(operation->mappedName());
```
pepone, line 862: Why do we need to remove escape prefix here and not above for inherit operations?

cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -0,0 +1,1487 @@
...
+        string featureParam = escapeParamName("features", operation->inParameters());
+        string cancellationTokenParam = escapeParamName("cancellationToken", operation->inParameters());
+
+        writeMethod(_out, operation, ns, false);
+
+        _out << " =>";
+        _out.inc();
+
+        // There is no equivalent for [compress] in Ice-Slice.
```
pepone: I find this comment unclear.

cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -0,0 +1,1487 @@
...
+    }
+    _out << eb;
+
+    // Generate extension classes.
+    _out << sp;
+    writeDocLine(
+        _out,
+        "summary",
+        R"(Provides an extension method for <see cref="SliceEncoder" /> to encode a <see cref=")" + name +
+            R"(Proxy" />."))");
```
pepone: Is there an extra `)` here?
```suggestion
            R"(Proxy" />.")");
```

cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -0,0 +1,1487 @@
...
+    _out << nl << "public static class " << name << "ProxySliceEncoderExtensions";
+    _out << sb;
+    writeDocLine(
+        _out,
+        "summary",
+        R"(Encodes a nullable <see cref=")" + name +
+            R"(Proxy" /> as a nullable <see cref="IceRpc.ServiceAddress" />.)");
+    writeDocLine(_out, "param name=\"encoder\"", "The Slice encoder.", "param");
```
pepone: We use `R"(` above to escape quotes, and `\"` here, seems a bit inconsistent.

pepone (approved):

cpp/src/slice2cs/Gen.h
```diff
@@ -8,20 +8,29 @@
 
 namespace Slice
 {
+    enum class GenMode
+    {
+        None,
+        Ice,
+        IceRpc
+    };
```
externl, line 16: I'd rather not have the `--rpc` flag and instead hav `--icerpc` and have the build tooling filter the various .cs files.

Not critical for this PR.

externl (approved):
Looks good!

cpp/src/Slice/Parser.cpp
```diff
@@ -2738,8 +2738,9 @@ Slice::InterfaceDef::createOperation(
                 break;
             }
         }
-        // Check the operations of the Object pseudo-interface.
-        if (!checkBaseOperationNames(name, {"ice_id", "ice_ids", "ice_ping", "ice_isA"}))
+
+        // Check the operations of the Object pseudo-interface unless we're generating code for Object itself.
+        if (scoped() != "::Ice::Object" && !checkBaseOperationNames(name, {"ice_id", "ice_ids", "ice_ping", "ice_isA"}))
```
InsertCreativityHere, line 2743: I'm really not a fan of having special sentinel type-ids baked into the parser like this.
This gives me more reason to suspect the new `Ice/Object.ice` is the wrong way to go here.

cpp/src/slice2cs/Main.cpp
```diff
@@ -43,6 +44,7 @@ usage(const string& n)
                   "-UNAME                   Remove any definition for NAME.\n"
                   "-IDIR                    Put DIR in the include file search path.\n"
                   "--output-dir DIR         Create files in the directory DIR.\n"
+                  "--rpc [none|ice|icerpc]  Generate code for the specified RPC framework. Default is 'ice'.\n"
```
InsertCreativityHere, line 47: We talked about this yesterday, but I think in a follow-up it would be nice to simplify this down to:
`--icerpc` instead of `--rpc XXX`, by removing the need for the `none` case. Simpler API, no argument validation, and I think it's surprising that passing `--rpc None` still causes `-D__ICERPC__`.

`--rpc none` is needed by the icerpc-csharp build, but an alternative would be for it to use `--icerpc`, then filter out any files ending with `.IceRpc.cs`.

cpp/src/slice2cs/Main.cpp
```diff
@@ -169,12 +201,30 @@ compile(const vector<string>& argv)
...
             }
 
-            unit = Unit::createUnit("cs");
+            UnitOptions unitOptions{};
+            if (genMode != Slice::GenMode::Ice)
+            {
+                unitOptions.defaultMappedName = [](const Contained& contained)
+                {
+                    // Convert all names to pascal case except for parameters, which are converted to camel case.
+                    if (dynamic_cast<const Parameter*>(&contained))
```
InsertCreativityHere, line 216: We always prefer this in the parser:
```suggestion
                    if (dynamic_pointer_cast<Parameter>(contained))
```

cpp/src/slice2cs/IceVisitors.h
```diff
@@ -52,7 +50,7 @@ namespace Slice::Ice
         ResultVisitor(IceInternal::Output&);
 
         bool visitModuleStart(const ModulePtr&) final;
-        void visitModuleEnd(const ModulePtr&) final;
+
```
InsertCreativityHere, line 53: This is a pretty small class, I don't think this blank line should be left here.
```suggestion
```

cpp/src/slice2cs/IceVisitors.h
```diff
@@ -63,7 +61,7 @@ namespace Slice::Ice
         SkeletonVisitor(IceInternal::Output& output, bool async);
 
         bool visitModuleStart(const ModulePtr&) final;
-        void visitModuleEnd(const ModulePtr&) final;
+
```
InsertCreativityHere, line 64: Ditto.
```suggestion
```

InsertCreativityHere (commented):
Still reviewing the IceRpc specific code, but a couple comments on the other code.

cpp/src/slice2cs/Main.cpp
pepone: We have the same argument in slicec-cs don't we. I think the filtering is going to be a "hack", generate this files  but don't build them. When the core function of SliceCompile task is to ensure files are generated and build. And I fear we will end-up with a special case just to support this.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -36,76 +42,134 @@ Slice::Gen::Gen(const string& base, const string& dir, bool enableAnalysis) : _e
...
     }
 
-    _out << nl << "#pragma warning disable CS1591 // Missing XML Comment";
-    _out << nl << "#pragma warning disable CS1573 // Parameter has no matching param tag in the XML comment";
-    _out << nl << "#pragma warning disable CS0612 // Type or member is obsolete";
-    _out << nl << "#pragma warning disable CS0618 // Type or member is obsolete";
-    _out << nl << "#pragma warning disable CS0619 // Type or member is obsolete";
+    if (_genMode == GenMode::Ice)
+    {
+        _out << nl << "[assembly:Ice.Slice(\"" << fileBase << ".ice\")]";
```
InsertCreativityHere, line 67: There should be a space here to separate the assemble from the `#pragma`s I feel:
```suggestion
        _out << sp;
        _out << nl << "[assembly:Ice.Slice(\"" << fileBase << ".ice\")]";
```

For reference, this is what this PR generates:
```
#pragma warning disable CS1591 // Missing XML Comment
#pragma warning disable CS1573 // Parameter has no matching param tag in the XML comment
#pragma warning disable CS0612 // Type or member is obsolete
#pragma warning disable CS0618 // Type or member is obsolete
#pragma warning disable CS0619 // Type or member is obsolete
[assembly:Ice.Slice("DBTypes.ice")]
```
And right now on main, we have a space between then (also the order swapped, but that's fine).

cpp/src/slice2cs/Gen.cpp
```diff
@@ -36,76 +42,134 @@ Slice::Gen::Gen(const string& base, const string& dir, bool enableAnalysis) : _e
...
+        _out << sp;
+        _out << nl << "[assembly:Slice(\"" << fileBase << ".ice\")]";
+
+        if (_genMode == GenMode::IceRpc)
+        {
+            _iceRpcOut << sp;
+            _iceRpcOut << nl << "using ZeroC.Slice;";
+            _iceRpcOut << nl << "using IceRpc.Slice;";
+            _iceRpcOut << sp;
+            _iceRpcOut << nl << "[assembly:Slice(\"" << fileBase << ".ice\")]";
```
InsertCreativityHere: I'm not exactly sure what the consequences are of this,
but we aren't currently generating an `assembly` attribute for the `.IceRpc.cs` files with `slicec-cs`.
`slicec-cs` only generates them for the rpc agnostic `.cs` files.

cpp/src/slice2cs/IceRpcCsUtil.cpp
bernardnormier: No, I'd rather not generate such comments.

cpp/src/slice2cs/IceRpcCsUtil.cpp
bernardnormier: It's the same link resolution logic/code as for Ice.

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -0,0 +1,671 @@
...
+        BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(type);
+        return builtin && builtin->kind() == Builtin::KindBool;
+    }
+
+    [[nodiscard]] bool isString(const TypePtr& type)
+    {
+        BuiltinPtr builtin = dynamic_pointer_cast<Builtin>(type);
+        return builtin && builtin->kind() == Builtin::KindString;
+    }
```
InsertCreativityHere, line 32: I find these functions a little much. They're each only called from one location, and are not so complicated. I'd rather just inline them.

csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
bernardnormier: I put this Rpc in Common because the command-line building is in Common. It would be nice to clean this up in a follow-up PR.

cpp/src/slice2cs/IceRpcVisitors.cpp
bernardnormier: Renamed writeMethodSignature and added doc-comment.

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -0,0 +1,671 @@
...
+    out.dec();
+    out << ");";
+    out.dec();
+
+    out << eb;
+    // else, don't encode anything for null.
+}
+
+void
+Slice::Csharp::decodeField(Output& out, const TypePtr& type, const string& ns, TypeContext)
```
InsertCreativityHere: We have a trailing `TypeContext` parameter that is dead? And this isn't an API we inherit or anything.
Should we remove this parameter, or is it supposed to be getting used somewhere in the future?

----

Looking at `decoding.rs`, it never relies on `TypeContext`. So I'm skeptical if it's necessary here.

InsertCreativityHere (approved):
So far looks good to me.

I didn't completely review all the IceRpc code, since there's alot.
Will keep reviewing in the background, hopefully as more tests are added.

cpp/src/slice2cs/IceRpcVisitors.cpp
bernardnormier: It's a bug. We need to remove the escape prefix for both. Async was also missing for the inherited operations.

cpp/src/slice2cs/IceRpcVisitors.cpp
bernardnormier: Reworked comment.

cpp/src/slice2cs/Main.cpp
InsertCreativityHere: I think it makes more sense to have this in icerpc, compared to ice.
With icerpc, there is a clean separation between rpc and non-rpc generated code, which doesn't exist in ice.
And IceRpc needs the ability to generate non-rpc code due to how we structured it (which is fine). Ice does not have this need.

Also, we are probably the only people who need to pass `--rpc none`. Users will either pass `--rpc ice` (99% by default), or `--rpc icerpc` (for interop). Having this 3rd `none` option suggests it's a reasonable choice for a user to pick, when it just isn't (in the context of interop-ing ice with icerpc).

> And I fear we will end-up with a special case just to support this.

Yes, but I think that passing `--rpc none` is also a special case, because of how absolutely niche it is. To me, the question is just where should the 'specialness' live.
And I'd rather have it live in icerpc (since it's specific to icerpc, not ice), instead of it being part of the command-line API for slice2cs (a tool primarily for Ice).

And if we get rid of `--rpc none`, then we're just left with `--rpc icerpc` and `--rpc ice`, and that point, It just feels cleaner to only have `--icerpc` (for `--rpc icerpc) and if you don't pass it, you get `--rpc ice` by default.

cpp/src/slice2cs/IceRpcCsUtil.cpp
bernardnormier: Removed.

cpp/src/slice2cs/Gen.cpp
bernardnormier: Fixed

cpp/src/slice2cs/Gen.cpp
bernardnormier: Removed from IceRpc.cs.

cpp/src/slice2cs/IceRpcVisitors.cpp
bernardnormier: Yes, we could replace various `\"` by `R"(` strings. Not critical.

csharp/src/ZeroC.Ice.Slice.Tools/SliceCompile.CSharp.xaml
bernardnormier: What do you suggest? We can fix it in a follow-up PR.

cpp/src/slice2cs/Main.cpp
bernardnormier: There is a difference between Ice and IceRPC here:

- for slice2cs, this none option is only to build the IceRpc core. Using Ice-Slice on its own (no RPC) is not an important or valid use-case.
- for slicec, the none option is more generally useful as we allow users to use Slice independently of any RPC

pepone: I think this is going to add friction with the compilation task where we always add all the generated code to the build. 

Maybe is simpler to get the generated code commit in IceRPC core, is not like this is going to change.

PR #5220, opened 2026-03-03
Fix exception message when batch count is negative (#5159)

pepone:
backport from 578beb7513


InsertCreativityHere (approved):
Looks good to me!

Here, this isn't 'a bug' since we don't print out the bad value like we did on 3.8.
But, it's still useless to set, since we never touch this out parameter after throwing.

externl (approved):
There's no message here in 3.7. So the title is not quite accurate. This is really just some cleanup.

PR #5222, opened 2026-03-03
Fix MatchesString::match ignoring pos parameter (#5154)

pepone:
Backport from 53967012fe

### DIFF
--- a/cpp/src/Glacier2/ProxyVerifier.cpp
+++ b/cpp/src/Glacier2/ProxyVerifier.cpp
@@ -489,7 +489,7 @@ public:
     create(const vector<int>&, const vector<Range>&)
     {
         assert(false); // unreachable -- groups are always processed inside the parser loop
-        return 0;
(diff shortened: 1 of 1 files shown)


cpp/src/Glacier2/ProxyVerifier.cpp
```diff
@@ -507,9 +486,10 @@ class EndsWithFactory : public AddressMatcherFactory
     }
 
     AddressMatcher*
-    create(const vector<int>& ports, const vector<Range>& ranges)
+    create(const vector<int>&, const vector<Range>&)
     {
-        return new EndsWithNumber(ports, ranges);
+        assert(false); // unreachable -- groups are always processed inside the parser loop
+        return 0;
```
InsertCreativityHere: On `3.7`, is it better to keep `return 0;` because that's what we use everywhere, or should it be:

```suggestion
        return ICE_NULLPTR;
```

PR #5226, opened 2026-03-04
Fix IceBox ServiceManagerI observer notification bugs

pepone:
- C++: Catch CommunicatorDestroyedException in addObserver to prevent unhandled exception propagating to the client during shutdown
- Java: Use break instead of observerFailed in servicesStarted and servicesStopped catch blocks to avoid ConcurrentModificationException from removing observers during iteration

### DIFF
--- a/csharp/src/iceboxnet/ServiceManagerI.cs
+++ b/csharp/src/iceboxnet/ServiceManagerI.cs
@@ -762,6 +762,10 @@ internal class ServiceManagerI : ServiceManagerDisp_, IDisposable
             {
                 await observer.servicesStartedAsync(services).ConfigureAwait(false);
             }
+            catch (Ice.CommunicatorDestroyedException)
+            {
+                // Expected during shutdown if the observer's communicator is destroyed.
+            }
             catch (System.Exception ex)
             {
                 removeObserver(observer, ex);
@@ -786,6 +790,10 @@ internal class ServiceManagerI : ServiceManagerDisp_, IDisposable
             try
             {
                 await observer.servicesStoppedAsync(services).ConfigureAwait(false);
+            }
+            catch (Ice.CommunicatorDestroyedException)
+            {
+                // Expected during shutdown if the observer's communicator is destroyed.
             }
             catch (System.Exception ex)
             {


cpp/src/IceBox/ServiceManagerI.cpp
```diff
@@ -306,7 +306,14 @@ IceBox::ServiceManagerI::addObserver(ServiceObserverPrx observer) // NOLINT(perf
...
         if (activeServices.size() > 0)
         {
-            observer->servicesStartedAsync(activeServices, nullptr, makeObserverCompletedCallback(observer));
+            try
+            {
+                observer->servicesStartedAsync(activeServices, nullptr, makeObserverCompletedCallback(observer));
+            }
+            catch (const CommunicatorDestroyedException&)
+            {
+                _observers.erase(observer);
```
pepone, line 315: We do this in the C# and Java versions. Not very important but I think is clear to have the same behavior in all 3 implementations.

java/src/com.zeroc.icebox/src/main/java/com/zeroc/IceBox/ServiceManagerI.java
```diff
@@ -613,7 +613,8 @@ private void servicesStarted(List<String> services, Set<ServiceObserverPrx> obse
                                 return null;
                             });
                 } catch (CommunicatorDestroyedException ex) {
-                    observerFailed(observer, ex);
+                    // Expected during shutdown if the observer's communicator is destroyed.
+                    break;
```
pepone, line 617: We don't want to modify `_observers` when CommunicatorDestroyedException is thrown synchronously while we are iterating it.

bernardnormier: If you look at the corresponding C++ and C# code:
 - in C++, we catch and ignore CommunicatorDestroyedException
 - in C#, we handled like other exceptions and call observerFailed that removes the observer from the dictionary

It's correct in C# because iterators support entry removal; however, I think we should make the code consistent across languages and also ignore CommunicatorDestroyedException in C#.

PR #5227, opened 2026-03-04
Fix IceBridge handling for CommunicatorDestroyedException

pepone:


cpp/src/IceBridge/IceBridge.cpp
```diff
@@ -419,11 +419,10 @@ BridgeI::ice_invokeAsync(
                     { self->outgoingSuccess(bridgeConnection, std::move(outgoing)); },
                     [self, bridgeConnection](auto ex) { self->outgoingException(bridgeConnection, ex); });
             }
-            catch (const std::exception&)
+            catch (const Ice::CommunicatorDestroyedException&)
             {
-                _connections.erase(current.con);
-                bridgeConnection->outgoingException(current_exception());
-                error(current_exception());
+                // The only exception thrown synchronously should be CommunicatorDestroyedException,
```
pepone, line 424: No point in doing this cleanup if the services is being shutdown.

Removing the connection here is problematic if `BridgeI::closed` is later call (from the connection close callback) as it expect the connection to be present.

PR #5232, opened 2026-03-05
Update Slice files for compatibility with IceRPC

pepone:


slice/Ice/Identity.ice
```diff
@@ -16,6 +16,9 @@
 [["js:module:@zeroc/ice"]]
 
 /// The Ice RPC framework.
```
bernardnormier, line 18: This comment is a bit confusing, but correct. It's the doc-comment for the Doxygen Ice module.

PR #5248, opened 2026-03-16
Update slice2cs for IceRPC to use IceRpc.Ice.Codec

bernardnormier:
This is the companion PR for https://github.com/icerpc/icerpc-csharp/pull/4310

### DIFF
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -69,16 +69,17 @@ Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool ena
     else
     {
         _out << sp;
-        _out << nl << "using ZeroC.Slice.Codec;";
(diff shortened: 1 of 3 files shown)


cpp/src/slice2cs/Gen.cpp
```diff
@@ -69,16 +69,17 @@ Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool ena
     else
     {
         _out << sp;
-        _out << nl << "using ZeroC.Slice.Codec;";
+        _out << nl << "using IceRpc.Ice;";
```
bernardnormier, line 72: Slice.Codec is replaced by Ice and Ice.Codec because IceClass and IceException are now in IceRpc.Ice.

pepone (approved):

externl (approved):

InsertCreativityHere (approved):

PR #5251, opened 2026-03-18
Replace '--rpc xxx' flag in slice2cs with just '--icerpc'.

InsertCreativityHere:
This PR simplifies the `--rpc` option in `slice2cs` (which takes an argument of either `none`, `ice`, or `icerpc`),
down to just `--icerpc` (with no arguments).

It effectively removes the `none` option, and maps `--rpc ice` -> `<empty>` and `--rpc icerpc` -> `--icerpc`.

----

I kept the GenMode enum because we like enums more than bools, even though we basically only ever use it as a bool.

### DIFF
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -59,28 +59,22 @@ Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool ena
     if (_genMode == GenMode::IceRpc)
     {
         printHeader(_iceRpcOut, fileBase + ".ice", _enableAnalysis);
-    }
 
-    if (_genMode == GenMode::Ice)
-    {
-        _out << sp;
-        _out << nl << "[assembly:Ice.Slice(\"" << fileBase << ".ice\")]";
-    }
-    else
-    {
         _out << sp;
         _out << nl << "using IceRpc.Ice;";
         _out << nl << "using IceRpc.Ice.Codec;";
         _out << sp;
         _out << nl << "[assembly:Ice(\"" << fileBase << ".ice\")]";
 
-        if (_genMode == GenMode::IceRpc)
-        {
-            _iceRpcOut << sp;
-            _iceRpcOut << nl << "using IceRpc.Ice;";
-            _iceRpcOut << nl << "using IceRpc.Ice.Codec;";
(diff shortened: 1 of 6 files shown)


cpp/src/slice2cs/Gen.cpp
```diff
@@ -59,28 +59,22 @@ Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool ena
...
-    }
-    else
-    {
         _out << sp;
         _out << nl << "using IceRpc.Ice;";
         _out << nl << "using IceRpc.Ice.Codec;";
         _out << sp;
         _out << nl << "[assembly:Ice(\"" << fileBase << ".ice\")]";
 
-        if (_genMode == GenMode::IceRpc)
```
InsertCreativityHere, line 77: Note that `GenMode::None` still meant we were generating code for IceRpc, just that we didn't want to generate `.IceRpc.cs` files.

Now that we added the manual filtering logic in https://github.com/icerpc/icerpc-csharp/pull/4272, there is no difference between `None` and `IceRpc` blocks.

csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
```diff
@@ -102,7 +102,7 @@ protected override string GenerateCommandLineCommands()
 
         if (Rpc.Length > 0 && Rpc != "ice")
```
pepone, line 103: Maybe just `Rpc == "icerpc"`

bernardnormier: Unless I am missing something, we should add validation for the `<Rpc>` argument.

With this code, if you specify `<Rpc>FooBar</Rpc>`, you'll get ice ... not really what you'd expect.

Would it be easier to use a bool property, like `<IceRpc>true</IceRpc>` instead of `<Rpc>`? This would be more consistent with the new `--icerpc` option.

InsertCreativityHere: I had assumed that the ToolTask would automatically verify you entered a valid input, since we list them in the xaml file. Perhaps not.

Either way, I agree a bool is better, I just didn't think anyone else would agree.
I'll change it in an immediate follow-up PR.

PR #5254, opened 2026-03-19
Replace 'Rpc' String Option with 'IceRpc' Bool Option in the Slice Tool

InsertCreativityHere:
**Addendum**
The name of the new option was changed to just `IceRpc`, instead of `GenerateIceRpc`.

**Original**

This PR follows up on https://github.com/zeroc-ice/ice/pull/5251#discussion_r2956070194 and switches the `Rpc` option (a string) in the `SliceCompile` tool to `GenerateIceRpc` (a bool).

This is:
1) More in-line with how it's used in the tool
2) More in-line with the `slice2cs` CLI, where `icerpc` is a boolean flag.

### DIFF
--- a/csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
+++ b/csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
@@ -30,7 +30,7 @@ public abstract class SliceCompilerTask : ToolTask
     /// <summary>
     /// Specifies whether to pass '--icerpc' to 'slice2cs'.
     /// </summary>
-    public bool GenerateIceRpc { get; set; } = false;
+    public bool IceRpc { get; set; } = false;
 
     public string[] AdditionalOptions { get; set; } = Array.Empty<string>();
 
@@ -69,10 +69,10 @@ public abstract class SliceCompilerTask : ToolTask
             options["IncludeDirectories"] = value;
         }
 
-        value = item.GetMetadata("GenerateIceRpc");
+        value = item.GetMetadata("IceRpc");
         if (!string.IsNullOrEmpty(value))
         {
-            options["GenerateIceRpc"] = value;
+            options["IceRpc"] = value;
         }
 
         if (AdditionalOptions.Length > 0)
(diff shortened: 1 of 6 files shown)


csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
```diff
@@ -28,9 +28,9 @@ public abstract class SliceCompilerTask : ToolTask
     public string[] IncludeDirectories { get; set; } = Array.Empty<string>();
 
     /// <summary>
-    /// The RPC provider to generate code for.
+    /// Specifies whether to pass '--icerpc' to 'slice2cs'.
     /// </summary>
-    public string Rpc { get; set; } = "ice";
+    public bool GenerateIceRpc { get; set; } = false;
```
bernardnormier: I think just IceRpc and `<IceRpc>true</IceRpc>` would be nicer. This option is specific to SliceCompile.

 It's slice2cs's icerpc option, not the generate-icerpc option.

csharp/src/ZeroC.Ice.Slice.Tools/SliceCompile.CSharp.xaml
```diff
@@ -49,18 +49,19 @@
...
     </StringListProperty>
 
-    <EnumProperty Name="Rpc" DisplayName="RPC Provider" Description="Generate code for the specified RPC framework.">
-        <EnumValue Name="ice" DisplayName="Ice" />
-        <EnumValue Name="icerpc" DisplayName="IceRPC" />
-        <EnumProperty.DataSource>
+    <BoolProperty
+        Name="GenerateIceRpc"
+        DisplayName="Generate IceRpc"
+        Description="Generate code for use with IceRpc projects.">
```
bernardnormier: I would capitalize IceRPC here: IceRPC projects.

csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
InsertCreativityHere: I updated the name to just `IceRpc`.

I thought it would be a vague name, but I see the logic in having it exactly match the name of the flag in slice2cs.

csharp/src/ZeroC.Ice.Slice.Tools/SliceCompile.CSharp.xaml
InsertCreativityHere: Fixed!

bernardnormier (approved):

PR #5256, opened 2026-03-20
Simplify code generated by slice2cs --icerpc

bernardnormier:
With this PR, slice2cs --icerpc generates a single file per .ice source file, fileBase.IceRpc.cs.

It also generates the `using IceRpc.Ice.Operations` only when the source Ice Slice file contains one or more interfaces, which simplifies the logic in icerpc-csharp (no need to ever delete .IceRpc.cs files).

### DIFF
--- a/cpp/src/slice2cs/Gen.cpp
+++ b/cpp/src/slice2cs/Gen.cpp
@@ -19,19 +19,17 @@ Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool ena
     : _genMode(genMode),
       _enableAnalysis(enableAnalysis)
 {
-    string fileBase = base;
+    _fileBase = base;
     string::size_type pos = base.find_last_of("/\\");
     if (pos != string::npos)
     {
-        fileBase = base.substr(pos + 1);
+        _fileBase = base.substr(pos + 1);
     }
-    string file = fileBase + ".cs";
-    string iceRpcFile = fileBase + ".IceRpc.cs";
+    string file = _genMode == GenMode::IceRpc ? _fileBase + ".IceRpc.cs" : _fileBase + ".cs";
 
     if (!dir.empty())
     {
         file = dir + '/' + file;
-        iceRpcFile = dir + '/' + iceRpcFile;
     }
 
     _out.open(file.c_str());
@@ -43,38 +41,17 @@ Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool ena
     }
     FileTracker::instance()->addFile(file);
 
+    printHeader(_fileBase + ".ice");
     if (_genMode == GenMode::IceRpc)
     {
-        _iceRpcOut.open(iceRpcFile.c_str());
-        if (!_iceRpcOut)
-        {
-            ostringstream os;
-            os << "cannot open '" << iceRpcFile << "': " << IceInternal::errorToString(errno);
-            throw FileException(os.str());
-        }
-        FileTracker::instance()->addFile(iceRpcFile);
-    }
-
-    printHeader(_out, fileBase + ".ice", _enableAnalysis);
-    if (_genMode == GenMode::IceRpc)
-    {
-        printHeader(_iceRpcOut, fileBase + ".ice", _enableAnalysis);
-
         _out << sp;
         _out << nl << "using IceRpc.Ice;";
         _out << nl << "using IceRpc.Ice.Codec;";
-        _out << sp;
-        _out << nl << "[assembly:Ice(\"" << fileBase << ".ice\")]";
-
-        _iceRpcOut << sp;
-        _iceRpcOut << nl << "using IceRpc.Ice;";
-        _iceRpcOut << nl << "using IceRpc.Ice.Codec;";
-        _iceRpcOut << nl << "using IceRpc.Ice.Operations;";
     }
     else
     {
         _out << sp;
-        _out << nl << "[assembly:Ice.Slice(\"" << fileBase << ".ice\")]";
+        _out << nl << "[assembly:Ice.Slice(\"" << _fileBase << ".ice\")]";
     }
 }
 
@@ -84,11 +61,6 @@ Slice::Gen::~Gen()
     {
         _out << nl;
         _out.close();
-    }
-    if (_iceRpcOut.isOpen())
-    {
-        _iceRpcOut << nl;
-        _iceRpcOut.close();
     }
 }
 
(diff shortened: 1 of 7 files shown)


cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -182,7 +182,26 @@ namespace
     string accessModifier(const ContainedPtr& p) { return p->hasMetadata("cs:internal") ? "internal" : "public"; }
 }
 
-Slice::IceRpc::TypesVisitor::TypesVisitor(IceInternal::Output& out) : CsVisitor(out) {}
+Slice::IceRpc::TypesVisitor::TypesVisitor(IceInternal::Output& out, std::string fileBase)
```
bernardnormier: Constructor + visitUnitStart is the only new code in the file. 

Everything else is moving existing functions around without other changes.

cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -182,7 +182,26 @@ namespace
...
+bool
+Slice::IceRpc::TypesVisitor::visitUnitStart(const UnitPtr& unit)
+{
+    if (unit->contains<InterfaceDef>())
+    {
+        // The proxy and skeleton code depends on this using directive.
+        _out << nl << "using IceRpc.Ice.Operations;";
+    }
+
+    _out << nl << "[assembly:IceGeneratedCode(\"" << _fileBase << ".ice\")]";
```
bernardnormier: Renamed IceAttribute (in Codec) => IceGeneratedCodeAttribute.

cpp/src/slice2cs/IceRpcVisitors.h
```diff
@@ -28,6 +30,10 @@ namespace Slice::IceRpc
 
         void visitEnum(const EnumPtr&) final;
 
+        bool visitInterfaceDefStart(const InterfaceDefPtr&) final;
```
bernardnormier, line 31: For proxies; they are now types as well.

cpp/src/slice2cs/IceRpcVisitors.h
bernardnormier, line 12: Oddly enough, the comment was already correct. (Really it was buggy prior to this PR).

pepone (approved):

cpp/src/slice2cs/Gen.cpp
```diff
@@ -19,19 +19,17 @@ Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool ena
...
+    _fileBase = base;
     string::size_type pos = base.find_last_of("/\\");
     if (pos != string::npos)
     {
-        fileBase = base.substr(pos + 1);
+        _fileBase = base.substr(pos + 1);
     }
-    string file = fileBase + ".cs";
-    string iceRpcFile = fileBase + ".IceRpc.cs";
+    string file = _genMode == GenMode::IceRpc ? _fileBase + ".IceRpc.cs" : _fileBase + ".cs";
```
bernardnormier, line 28: Ice => .cs
IceRpc => .IceRpc.cs

I think it's clearer than use .cs for both. And .IceRpc.cs does indeed always depend on IceRpc.dll.

csharp/src/ZeroC.Ice.Slice.Tools/Slice2CSharpDependTask.cs
```diff
@@ -8,16 +8,10 @@ namespace ZeroC.Ice.Slice.Tools;
...
-                new TaskItem(GetGeneratedPath(source, source.GetMetadata("OutputDir"), ".IceRpc.cs"))
-            } :
-            new ITaskItem[]
-            {
-                new TaskItem(GetGeneratedPath(source, source.GetMetadata("OutputDir"), ".cs"))
-            };
+        new ITaskItem[]
+        {
+            new TaskItem(GetGeneratedPath(source, source.GetMetadata("OutputDir"), IceRpc ? ".IceRpc.cs" : ".cs"))
+        };
```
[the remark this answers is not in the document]
bernardnormier: It's not a concern as the slice2cs with IceRPC support has not shipped yet.

cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -182,7 +182,26 @@ namespace
...
+
+bool
+Slice::IceRpc::TypesVisitor::visitUnitStart(const UnitPtr& unit)
+{
+    if (unit->contains<InterfaceDef>())
+    {
+        // The proxy and skeleton code depends on this using directive.
+        _out << nl << "using IceRpc.Ice.Operations;";
+    }
+
```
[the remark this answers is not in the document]
bernardnormier: Not a concern - it's all new code.

cpp/src/slice2cs/IceRpcVisitors.h
```diff
@@ -41,23 +47,11 @@ namespace Slice::IceRpc
             bool hasBase,
             const DataMemberList& fields,
             const DataMemberList& allBaseFields);
```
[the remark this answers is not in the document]
bernardnormier: Fixed name in function declaration.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -19,19 +19,17 @@ Slice::Gen::Gen(const string& base, const string& dir, GenMode genMode, bool ena
...
 
     if (!dir.empty())
     {
         file = dir + '/' + file;
-        iceRpcFile = dir + '/' + iceRpcFile;
     }
```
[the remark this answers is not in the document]
bernardnormier: Should be fixed.

cpp/src/slice2cs/Gen.cpp
```diff
@@ -115,53 +87,52 @@ Slice::Gen::generate(const UnitPtr& p)
...
     else
     {
-        Slice::IceRpc::TypesVisitor typesVisitor(_out);
+        // The types visitor must be the first visitor since it optionally generates a using directive for all
+        // visitors to use.
+        Slice::IceRpc::TypesVisitor typesVisitor(_out, _fileBase);
```
InsertCreativityHere: I think it'd be a little cleaner to just have the logic here instead of shoehorned in the `TypeVisitor`.
All the code here also only runs once per file, so there's no difference vs using `visitUnitStart`.
```suggestion
        if (unit->contains<InterfaceDef>())
        {
            // The proxy and skeleton code depends on this using directive.
            _out << nl << "using IceRpc.Ice.Operations;";
        }
        _out << nl << "[assembly:IceGeneratedCode(\"" << _fileBase << ".ice\")]";Expand commentComment
        _out << sp;

        Slice::IceRpc::TypesVisitor typesVisitor(_out);
```
This also gets rid of the new `_fileBase` field on TypesVisitor, which it only uses just to write the assembly attribute.

InsertCreativityHere (approved):

cpp/src/slice2cs/Gen.cpp
bernardnormier: Fixed as suggested.

PR #5261, opened 2026-03-24
Remove the 'ice2slice' Tool

InsertCreativityHere:
It is my current understanding that we want to remove `ice2slice`, and users will now continue using their `.ice` files instead of converting them to `.slice` if they're interested in interop with Ice.

So, with a heavy heart, I `rg`ed for `ice.?slice` and removed everywhere that was related to `ice2slice`.

### DIFF
--- a/packaging/deb/debian/control
+++ b/packaging/deb/debian/control
@@ -137,6 +137,14 @@ Description: Glacier2 router
  network programming interfaces and allows you to focus your efforts on
  your application logic.
 
+Package: zeroc-ice-ice2slice
+Architecture: any
+Section: oldlibs
+Depends: ${misc:Depends}
+Breaks: zeroc-ice-slice (<< ${binary:Version})
+Description: Ice to Slice compiler
+ Transitional package - safe to remove
+
 Package: zeroc-ice-slice
 Architecture: all
 Section: devel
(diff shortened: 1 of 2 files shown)


packaging/deb/debian/control
```diff
@@ -137,24 +137,6 @@ Description: Glacier2 router
  network programming interfaces and allows you to focus your efforts on
  your application logic.
 
-Package: zeroc-ice-ice2slice
```
pepone, line 140: I think we might need a transitional package for the clean removal. My concern is this will break `apt-get upgrade`.  If you have this package installed it requires zeroc-ice-slice 3.8.0, but updated packages will pull zeroc-ice-slice 3.8.1 and since you cannot install both it will break. I think the same is true for RPMs.

externl: I don't think we'll merge this commit onto 3.8.

pepone: We still need a clean upgrade from 3.8 to 3.9. But what is the point of keeping it in 3.8 if we are not going to support it?

pepone: My understanding is that main is for now basically 3.8, we should not start yet to do 3.9 work. Maybe is better to sort this out in a issue that documents the removal.

bernardnormier: We should remove ice2slice from 3.8(.2) as well.

PR #5262, opened 2026-03-24
Add Support for `*` in Links to Control Whether it Links to a Proxy or a Servant Type

InsertCreativityHere:
This PR just adds support for the syntax, it doesn't go through and update all the `@see` and `@link` we have.
(Except for `PermissionsVerifier` so it's at least being a little tested). Will open a follow-up PR to update the rest.
```slice
module MyModule {
    /// This will link to a `Foo`: {@link Foo}
    /// This will link to a `FooPrx`: {@link Foo*}.
    interface Foo {}
}
```
The same logic applies to `@see` tags as well.

With regards to client-only languages: we don't generate doc-comments at all for PHP and Ruby, which just leaves MATLAB.
In MATLAB, the `*` is effectively ignored. You will always get the proxy type, since there's no servant type to link to.

### DIFF
--- a/cpp/src/Slice/DocCommentParser.cpp
+++ b/cpp/src/Slice/DocCommentParser.cpp
@@ -237,8 +237,19 @@ namespace
 
     /// Returns a pointer to the Slice element referenced by `linkText`, relative to the scope of `source`.
     /// If the link cannot be resolved, `nullptr` is returned instead.
+    ///
+    /// @remark For classes and interfaces, the resolved type will always be an `xxxDecl` and never an `xxxDef`.
     SyntaxTreeBasePtr resolveDocLink(string linkText, const ContainedPtr& source)
     {
+        assert(!linkText.empty());
+
+        // If the link ends with '*' (i.e. the syntax for linking to a proxy), remove it so we can lookup the type.
+        bool useProxyType = linkText.back() == '*';
+        if (useProxyType)
+        {
+            linkText.pop_back();
+        }
+
         // First we check if the link is to a builtin type.
         if (auto kind = Builtin::kindFromString(linkText))
         {
@@ -253,20 +264,33 @@ namespace
         }
 
         // Perform the actual lookup.
+        ContainedList results;
         auto separatorPos = linkText.find('#');
         if (separatorPos == 0)
         {
             // If the link starts with '#', it is explicitly relative to the `linkSourceScope` container.
-            ContainedList results = source->unit()->findContents(linkSourceScope->thisScope() + linkText.substr(1));
-            return (results.empty() ? nullptr : results.front());
+            results = source->unit()->findContents(linkSourceScope->thisScope() + linkText.substr(1));
         }
-        else if (separatorPos != string::npos)
+        else
         {
-            // If the link has a '#' anywhere else, convert it to '::' so we can look it up.
-            linkText.replace(separatorPos, 1, "::");
+            if (separatorPos != string::npos)
+            {
+                // If the link has a '#' anywhere else, convert it to '::' so we can look it up.
+                linkText.replace(separatorPos, 1, "::");
+            }
+            results = linkSourceScope->lookupContained(linkText, false);
         }
-        ContainedList results = linkSourceScope->lookupContained(linkText, false);
-        return (results.empty() ? nullptr : results.front());
+        // We want to use the first match, if there were any.
+        ContainedPtr result = (results.empty() ? nullptr : results.front());
+
+        // If the link ended with '*', emit a warning if the link was to a non-interface.
+        if (useProxyType && result && !dynamic_pointer_cast<InterfaceDecl>(result))
+        {
+            string msg = "'*' cannot be used with '" + result->name() + "' as it is not an interface";
+            source->unit()->warning(source->file(), source->line(), InvalidComment, msg);
+        }
+
+        return result;
     }
 
     /// Formats a code-span starting at \p pos.
@@ -334,6 +358,14 @@ namespace
 
             // Then erase the entire '{@link foo}' tag from the comment.
             line.erase(pos, endpos - pos + 1);
+
+            // If the user is missing a target after '@link', emit a warning and don't attempt a lookup.
+            if (linkText.empty())
+            {
+                const string msg = "missing link target after '" + linkTag + "' tag";
+                p->unit()->warning(p->file(), p->line(), InvalidComment, msg);
+                return;
+            }
 
             // Attempt to resolve the link, and issue a warning if the link is invalid.
             SyntaxTreeBasePtr linkTarget = resolveDocLink(linkText, p);
@@ -452,7 +484,7 @@ DocCommentParser::parseDocCommentFor(const ContainedPtr& p)
         while (true)
         {
             // We check for all inline tags all at the same time, to prevent overlap.
-            // For example: `@p hello` should _only_ be a code-span, not also a param-ref.
+            // For example: 'this `@p hello` is good' should _only_ have a code-span, not also a param-ref.
             size_t nextBacktickPos = line.find('`', pos);
             size_t nextLinkPos = line.find("{@link ", pos);
             size_t nextParamrefPos = line.find("@p", pos);
--- a/cpp/src/Slice/Grammar.y
+++ b/cpp/src/Slice/Grammar.y
@@ -1711,7 +1711,7 @@ namespace
         }
         if (!interface && expectInterfaceType)
         {
-            string msg = "'" + name + "' must be an interface";
+            string msg = "'*' cannot be used with '" + name + "' as it is not an interface";
             currentUnit->error(msg);
         }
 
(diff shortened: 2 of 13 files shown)


cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -348,6 +344,14 @@ Slice::Swift::swiftLinkFormatter(const string& rawLink, const ContainedPtr& sour
         string mappedLink = Slice::Swift::getRelativeTypeString(contained, sourceModule) + nameSuffix;
         std::replace(mappedLink.begin(), mappedLink.end(), '.', '/');
 
+        // Unfortunately, we always have to get the mapped name of the proxy type, then fix it afterwards.
+        // The code-gen takes `InterfaceDecl` to mean "a proxy" and `InterfaceDef` to mean "a servant". But it's
+        // possible a user only declared an interface, and didn't define it, so we can't assume the `Def` exists.
+        if (dynamic_pointer_cast<InterfaceDecl>(contained) && !useProxyType)
+        {
+            mappedLink.erase(mappedLink.length() - 3);
```
externl: Can we just "rstrip" `Prx` from the end instead of using a magic number?

externl (approved):

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere: Hah, `rstrip` in the C++ standard library? If only : vP There's no function I know of (nor Claude) like that. `string_view` has a `remove_suffix`, but A) we need a `string` here, and B) it still just takes a count anyways.

I guess we could do this:
```suggestion
            mappedLink.erase(mappedLink.length() - (sizeof("Prx") - 1));
```
or actually even this is better:
```suggestion
            mappedLink.erase(mappedLink.length() - strlen("Prx"));
```
Do you think this is an improvement? Not clear to me.

cpp/src/Slice/Grammar.y
```diff
@@ -1711,7 +1711,7 @@ namespace
         }
         if (!interface && expectInterfaceType)
         {
-            string msg = "'" + name + "' must be an interface";
```
bernardnormier, line 1714: We should consider moving these functions to GrammarUtil.h (for a separate PR)

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -572,14 +572,14 @@ Slice::Csharp::icerpcLinkFormatter(const string& rawLink, const ContainedPtr& so
         result << "cref=\"";
         if (auto operationTarget = dynamic_pointer_cast<Operation>(target))
         {
-            // link to the method on the proxy interface
+            // link to the method on the proxy interface.
```
bernardnormier, line 575: proxy struct

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -572,14 +572,14 @@ Slice::Csharp::icerpcLinkFormatter(const string& rawLink, const ContainedPtr& so
...
             result << getUnqualified(operationTarget->interface(), sourceScope, "", "Proxy") << "."
                    << removeEscapePrefix(operationTarget->mappedName()) << "Async";
         }
-        else if (auto interfaceTarget = dynamic_pointer_cast<InterfaceDecl>(target))
+        else if (dynamic_pointer_cast<InterfaceDecl>(target) && rawLink.back() == '*')
         {
             // link to the proxy interface
```
bernardnormier, line 581: proxy struct

cpp/src/slice2cs/IceRpcCsUtil.cpp
bernardnormier: Should also check for `*` to distinguish between interface and service?

cpp/src/slice2cs/IceRpcCsUtil.cpp
```diff
@@ -572,14 +572,14 @@ Slice::Csharp::icerpcLinkFormatter(const string& rawLink, const ContainedPtr& so
...
-            result << getUnqualified(interfaceTarget, sourceScope, "", "Proxy");
+            result << getUnqualified(contained, sourceScope, "", "Proxy");
```
bernardnormier, line 582: The service interface is `getUnqualified(contained, sourceScope, "I", "Service");`

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -3318,6 +3318,7 @@ Slice::matlabLinkFormatter(const string& rawLink, const ContainedPtr&, const Syn
             displayText = opTarget->mappedName();
             linkText = interfaceDef->mappedScoped(".") + "Prx" + "/" + displayText;
         }
+        // Don't check for trailing '*', since MATLAB _only_ generates proxy types, not servant types.
```
bernardnormier, line 3321: I think it's odd to generate a link for the proxy when the author of the Slice doc-comment explicitly asked for a doc-comment to the server-side interface/skeleton.

I'd rather not generate any link to this case.

cpp/src/slice2swift/SwiftUtil.cpp
```diff
@@ -348,6 +344,14 @@ Slice::Swift::swiftLinkFormatter(const string& rawLink, const ContainedPtr& sour
         string mappedLink = Slice::Swift::getRelativeTypeString(contained, sourceModule) + nameSuffix;
         std::replace(mappedLink.begin(), mappedLink.end(), '.', '/');
 
+        // Unfortunately, we always have to get the mapped name of the proxy type, then fix it afterwards.
+        // The code-gen takes `InterfaceDecl` to mean "a proxy" and `InterfaceDef` to mean "a servant". But it's
```
bernardnormier, line 349: I find this comment confusing. AFAIK, `target` here is never an `InterfaceDef`, and this comment suggests it could be.

cpp/src/slice2swift/SwiftUtil.cpp
InsertCreativityHere: I changed it to `strlen("Prx")`. More clear what we're doing.

cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -253,20 +264,33 @@ namespace
...
         }
-        ContainedList results = linkSourceScope->lookupContained(linkText, false);
-        return (results.empty() ? nullptr : results.front());
+        // We want to use the first match, if there were any.
+        ContainedPtr result = (results.empty() ? nullptr : results.front());
+
+        // If the link ended with '*', emit a warning if the link was to a non-interface.
+        if (useProxyType && result && !dynamic_pointer_cast<InterfaceDecl>(result))
+        {
+            string msg = "'*' cannot be used with '" + result->name() + "' as it is not an interface";
```
pepone, line 289: Doesn't the Grammar has a similar check, why are both required?

cpp/src/slice2js/JsUtil.cpp
```diff
@@ -516,22 +516,19 @@ Slice::JavaScript::jsLinkFormatter(const string& rawLink, const ContainedPtr&, c
         {
             if (auto operationTarget = dynamic_pointer_cast<Operation>(target))
             {
+                // link to the method on the proxy interface.
```
pepone, line 519: Do we plan to support linking to Service operations? this doesn't seem very consistent:

Foo -> links to service
Foo* -> links to proxy
Foo::bar -> links to the proxy operation

cpp/src/slice2swift/SwiftUtil.cpp
pepone: I agree I would rather generate an inline code block with the Slice type name, similar to what we do for links to other unresolved types.

cpp/src/slice2js/JsUtil.cpp
bernardnormier: I think we should support for the `Foo*` for operation as well, i.e.:

```
Foo:bar -> link to skeleton operation
Foo*::bar -> links to proxy operation
```

InsertCreativityHere: I'd rather not have 2 syntaxes to express the same thing (`Foo*::bar` and `Foo::bar*`).
I definitely see the consistency angle you're coming from (that you're marking the interface type),
but I think we can still have a consistent syntax by framing it as `{@link (target)*}`
where that target could be an interface or an operation. But the star is always trailing.

----

Also, on a technical level, checking if the last character is `*` is MUCH simpler, than needing to parse the identifier and check each segment for `*`s. Both for us, and for our Doxygen changes.

InsertCreativityHere:
Closing this for now, can revisit in the future.

**It's not clear what the ideal and most consistent syntax is for this at the Slice level.**
Always attaching star to interfaces (`Foo*::myOp`) is the most valid, but makes it asymmetrically harder to reference proxy methods (you couldn't write `mpOp*`, you have to qualify if with the proxy type). Always attaching the star to the end (`Foo::myOp*`) is more ergonomic and symmetric, but on-the-face of it is nonsensical.

**Modifying Doxygen's code to properly support this `*` in it's links is not a small change either**,
and even after much research, the side-effects of doing this is not clear to me. Especially for `@see` in Doxygen, which relies on auto-linking to work. Touching anything in the auto-linking code is a big can of worms.

Yes, there is a use-case for being able to link to proxies vs servants, but it's a small one, mostly just for us.
I don't think it's worth the time investment at the moment.

PR #5264, opened 2026-03-24
Add Copilot PR review instructions

externl:
First pass at a Copilot PR review instructions file. Claude helped me write. Please tell me if you wanted something changed. Currently just focuses on the changelog and binary compatibility. 

This is a file we'll need to tweak over time as we observe it in action.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -9,7 +9,7 @@ Binary compatibility includes changes such as:
 - Removing or renaming public classes, methods, or functions
 - Changing method signatures (parameters, return types)
 - Changing class hierarchy or virtual table layout
-- Removing or reordering public data members
+- Removing or reordering data members of public types
 - Changing enum values or constants
 
 ### Rules by branch:


.github/copilot-instructions.md
```diff
@@ -0,0 +1,55 @@
+# Copilot PR Review Instructions
+
+## Binary Compatibility Check
+
+**When reviewing PRs, check if changes break public API binary compatibility.**
```
pepone, line 5: There was a good C++ centric guide for this in https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B

But not sure how specific this review instructions need to be

For 3.8 we have a dedicated CI job to check this: https://github.com/zeroc-ice/ice/actions/workflows/compat-nightly.yml

.github/copilot-instructions.md
```diff
@@ -0,0 +1,55 @@
...
+This PR targets a release branch and includes changes that break public API binary compatibility.
+Binary compatibility must be preserved on release branches.
+
+Breaking change(s):
+- [Description of the breaking change]
+```
+
+## Changelog Validation
+
+**When reviewing PRs, check if code changes have a CHANGELOG.md entry.**
```
pepone, line 32: Hopefully this doesn't add much noise and verbosity to the reviews

.github/copilot-instructions.md
```diff
@@ -0,0 +1,55 @@
...
+- Removing or renaming public classes, methods, or functions
+- Changing method signatures (parameters, return types)
+- Changing class hierarchy or virtual table layout
+- Removing or reordering public data members
+- Changing enum values or constants
+
+### Rules by branch:
+
+- **`main` branch**: Breaking binary compatibility is **allowed**.
+- **Release branches** (e.g., `3.7`, `3.8`): Breaking binary compatibility is **not allowed**.
```
InsertCreativityHere, line 18: Should we just say that they're allowed on main, but not on any other branch?
We don't actually define what a "release branch" is. And Asking Copilot what it thinks a release branch is, it answered:

> A release branch is a separate branch in version control that is created to prepare for a software release. It allows for final testing and bug fixes before the code is merged back into the main branch for deployment.

Which is not what we want here.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Binary Compatibility Check
+
+**When reviewing PRs, check if changes break public API binary compatibility.**
+
+Binary compatibility includes changes such as:
+
+- Removing or renaming public classes, methods, or functions
+- Changing method signatures (parameters, return types)
+- Changing class hierarchy or virtual table layout
```
InsertCreativityHere, line 11: I'm not a vtable expert, but isn't this just saying the same thing twice?
There's no way to manually change the layout of a vtable right?
Or is this to handle weird C++ specific stuff like the virtual destructor trick.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,55 @@
...
+- Removing or reordering public data members
```
InsertCreativityHere: Pretty minor, but I think this is more accurate.
Moving private members in a public type also breaks binary compatibility.
```suggestion
- Removing or reordering data members of public types
```

.github/copilot-instructions.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Changelog Validation
+
+**When reviewing PRs, check if code changes have a CHANGELOG.md entry.**
+
+### Requires changelog entry:
+
+- Non-trivial bug fixes, features, API changes, security fixes, performance improvements
+- Breaking changes, deprecations
+- Multi-language changes
```
InsertCreativityHere, line 38: Does "multi-language change" belong in this list?
If I make a small change to both C# and Java, I don't think that necessarily means it should be CHANGELOGged.

.github/copilot-instructions.md
externl: We can tweak as we go if it's too noisy. Not not noisy enough.

.github/copilot-instructions.md
externl: We can update this list as necessary.

.github/copilot-instructions.md
externl: Hard to say how Copilot will really interpret this. I'd prefer to commit this and just see how it handles it.

.github/copilot-instructions.md
externl: I think it's fine. We can see if it works or doesn't.

bernardnormier: >  but isn't this just saying the same thing twice

It's not.

For example, you change the vtable layout by adding or removing a virtual function. This has no effect on the class hierarchy. Adding a non-virtual function is fine wrt compatibility.

PR #5268, opened 2026-03-26
Fix dotnet format lints

bernardnormier:
All about throwing when some variable / expression is null.

### DIFF
--- a/csharp/src/Ice/Internal/DefaultsAndOverrides.cs
+++ b/csharp/src/Ice/Internal/DefaultsAndOverrides.cs
@@ -25,12 +25,9 @@ public sealed class DefaultsAndOverrides
         val = properties.getIceProperty("Ice.Default.SourceAddress");
         if (val.Length > 0)
         {
-            defaultSourceAddress = Network.getNumericAddress(val);
-            if (defaultSourceAddress == null)
-            {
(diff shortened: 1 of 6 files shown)


InsertCreativityHere (approved):

csharp/src/Ice/Internal/DefaultsAndOverrides.cs
```diff
@@ -25,12 +25,9 @@ internal DefaultsAndOverrides(Ice.Properties properties, Ice.Logger logger)
...
         {
-            defaultSourceAddress = Network.getNumericAddress(val);
-            if (defaultSourceAddress == null)
-            {
-                throw new Ice.InitializationException("invalid IP address set for Ice.Default.SourceAddress: `" +
-                                                      val + "'");
-            }
+            defaultSourceAddress = Network.getNumericAddress(val) ??
+                throw new Ice.InitializationException(
+                    "invalid IP address set for Ice.Default.SourceAddress: '" + val + "'");
```
pepone, line 30: we should use string interpolation for val

pepone (approved):

csharp/src/Ice/Internal/DefaultsAndOverrides.cs
bernardnormier: For a follow-up PR.

PR #5278, opened 2026-04-01
Rpm fix 3.8.1 1

pepone:


packaging/rpm/ice.spec
externl, line 4: Good catch!

externl (approved):
Looks good! We should rg though the branch for 3.9

PR #5287, opened 2026-04-07
Add ["oneway"] metadata

### LINKED ISSUE #5224: Add ["oneway"] metadata directive
We want to add a `["oneway"]` metadata directive to Slice operations.

The semantics would be as follows: if you call a `["oneway"]` operation with a two-way proxy, you get `OnewayOnlyException`. You don't get this exception for any other proxy invocation modes, since all other invocation modes (Oneway, Datagram, etc.) have oneway semantics (no response).

`OnewayOnlyException` is a new local exception, the counterpart of the existing `TwowayOnlyException`.

This `["oneway"]` metadata directive will be more useful when targeting IceRPC (see #5206). With IceRPC, a Slice operation marked oneway is sent as a one-way request; see https://docs.icerpc.dev/slice2/language-guide/operation#oneway-attribute.

externl:
Fixes #5224


cpp/include/Ice/LocalExceptions.h
```diff
@@ -856,6 +856,27 @@ namespace Ice
         [[nodiscard]] const char* ice_id() const noexcept final;
     };
 
+    /// The exception that is thrown when attempting to invoke a oneway-only operation (an operation with the
```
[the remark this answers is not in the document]
externl: oneway does not have a vowel sound here.

cpp/src/slice2java/Gen.cpp
```diff
@@ -1473,6 +1473,13 @@ Slice::JavaVisitor::writeIceIHelperMethods(
     out << nl << futureImplType << " f = new com.zeroc.Ice.OutgoingAsync<>(this, \"" << p->name() << "\", "
         << sliceModeToIceMode(p->mode()) << ", sync, " << (hasExceptionSpecification ? "_iceE_" + name : "null")
         << ");";
+    if (p->hasMetadata("oneway"))
+    {
+        out << nl << "if (this.ice_isTwoway())";
```
pepone: Why didn't you add a helper in Java?

js/packages/ice/src/Ice/ObjectPrxExtensions.js
```diff
@@ -347,10 +353,13 @@ ObjectPrx.prototype.ice_instanceof = function (T) {
 //
 // Generic invocation for operations that have input parameters.
 //
-ObjectPrx._invoke = function (p, name, mode, fmt, ctx, marshalFn, unmarshalFn, userEx, args) {
+ObjectPrx._invoke = function (p, name, mode, fmt, ctx, marshalFn, unmarshalFn, userEx, args, onewayOnly) {
     if (unmarshalFn !== null || userEx.length > 0) {
         p._checkAsyncTwowayOnly(name);
```
pepone: Not related to the PR, but we should rename  `_checkAsyncTwowayOnly` to `_checkTwowayOnly`

pepone (approved):

CHANGELOG.md
```diff
@@ -21,4 +21,8 @@ These are the changes since the [Ice 3.8.1] release.
 
 ### Slice Language Changes
 
+- Added the `["oneway"]` metadata directive for Slice operations. When applied to an operation that does not return
+  data (no return type, out parameters, or exception specification), the generated proxy code throws
+  `OnewayOnlyException` if the operation is invoked on a twoway proxy.
```
bernardnormier: Is it thrown synchronously, like TwowayOnlyException?

https://docs.zeroc.com/ice/3.8/cpp/asynchronous-method-invocation-ami-in-c#AsynchronousMethodInvocation(AMI)inC++-AsynchronousExceptionSemantics

cpp/test/Ice/operations/Oneways.cpp
```diff
@@ -56,6 +56,18 @@ oneways(const Ice::CommunicatorPtr&, const Test::MyClassPrx& proxy)
         p->opIdempotent();
     }
 
+    {
+        // Calling a ["oneway"] operation on a twoway proxy throws OnewayOnlyException.
```
bernardnormier, line 65: We should also verify the success path.

csharp/test/Ice/operations/Oneways.cs
```diff
@@ -14,6 +14,16 @@ internal static void oneways(global::Test.TestHelper helper, Test.MyClassPrx p)
         p.opVoid();
         p.opIdempotent();
 
+        // Calling a ["oneway"] operation on a twoway proxy throws OnewayOnlyException.
```
bernardnormier, line 20: Add test for success path.

matlab/lib/+Ice/OnewayOnlyException.m
```diff
@@ -0,0 +1,19 @@
...
+    % Copyright (c) ZeroC, Inc.
+
+    methods (Hidden)
+        function obj = OnewayOnlyException(operation)
+            if nargin == 0
+                superArgs = {};
+            else
+                assert(nargin == 1, 'Invalid number of arguments');
+                superArgs = {'Ice:OnewayOnlyException',...
+                    sprintf('cannot invoke oneway operation ''%s'' with a twoway proxy', operation)};
```
InsertCreativityHere, line 14: Of course, we aren't fully consistent, but it seems like we prefer capitalized error messages in MATLAB:
```suggestion
                    sprintf('Cannot invoke oneway operation ''%s'' with a twoway proxy', operation)};
```
It's possible that that's not correct MATLAB though, and just something we're doing.

python/test/Ice/operations/Oneways.py
```diff
@@ -24,6 +24,15 @@ def oneways(helper: TestHelper, p: Test.MyClassPrx) -> None:
     #
     p.opIdempotent()
 
+    #
+    # opOneway - calling a ["oneway"] operation on a twoway proxy throws OnewayOnlyException
```
InsertCreativityHere: Maybe this is a sentence?
```suggestion
    # opOneway - calling a ["oneway"] operation on a twoway proxy throws OnewayOnlyException.
```

cpp/src/slice2java/Gen.cpp
```diff
@@ -1473,6 +1473,13 @@ Slice::JavaVisitor::writeIceIHelperMethods(
     out << nl << futureImplType << " f = new com.zeroc.Ice.OutgoingAsync<>(this, \"" << p->name() << "\", "
         << sliceModeToIceMode(p->mode()) << ", sync, " << (hasExceptionSpecification ? "_iceE_" + name : "null")
         << ");";
+    if (p->hasMetadata("oneway"))
```
bernardnormier, line 1473: When binary compatibility is not a concern, we should clean-up this code and check for both oneway-only and twoway-only in the same spot (in the generated code, not this boolean parameter).

js/packages/ice/src/Ice/Operation.js
```diff
@@ -128,6 +128,7 @@ function parseOperation(name, arr) {
 
     r.sendsClasses = arr[7] === true;
     r.returnsClasses = arr[8] === true;
+    r.onewayOnly = arr[9] === true;
```
bernardnormier, line 132: You need to update the comment as well.

php/src/Operation.cpp
```diff
@@ -707,10 +722,11 @@ ZEND_FUNCTION(IcePHP_defineOperation)
     zval* outParams;
     zval* returnType;
     zval* exceptions;
+    zend_bool oneway;
 
     if (zend_parse_parameters(
             ZEND_NUM_ARGS(),
-            const_cast<char*>("osslla!a!a!a!"),
```
InsertCreativityHere, line 713: You should `rg` for `Ice_defineOperation_arginfo` and update it.
Need to bump the argument count, and add add the `oneway` argument info.

js/packages/ice/src/Ice/ObjectPrxExtensions.js
```diff
@@ -347,10 +353,13 @@ ObjectPrx.prototype.ice_instanceof = function (T) {
 //
 // Generic invocation for operations that have input parameters.
```
bernardnormier: Is this comment not accurate? We need such check for operations with no input parameters as well.

matlab/lib/+Ice/ObjectPrx.m
```diff
@@ -1131,7 +1131,7 @@ function iceEndWriteParams(~, os)
             os.endEncapsulation();
         end
 
-        function is = iceInvoke(obj, op, mode, twowayOnly, os, hasOutParams, exceptions, ctx)
+        function is = iceInvoke(obj, op, mode, twowayOnly, onewayOnly, os, hasOutParams, exceptions, ctx)
```
bernardnormier: I am not a fan of this API. twowayOnly obviously implies !onewayOnly.

I think it would be cleaner to put these checks directly in the generated code, and not in iceInvoke via parameters. Maybe I am missing something?

ruby/test/Ice/operations/Oneways.rb
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) ZeroC, Inc.
+
+require 'Ice'
+
+def oneways(helper, p)
+    p = Test::MyClassPrx.uncheckedCast(p.ice_oneway())
```
InsertCreativityHere, line 6: Why do we need this line? `p` is already ` MyClassPrx` and we're about to `ice_twoway()` it anyways.

cpp/src/slice2swift/Gen.cpp
```diff
@@ -1314,6 +1314,15 @@ Gen::TypesVisitor::visitOperation(const OperationPtr& op)
     // Invoke
     //
     out << sp;
+
+    if (op->hasMetadata("oneway"))
```
bernardnormier, line 1317: The twoway-only check is performed in _invoke below based on the read and userException arguments?

.vscode/settings.json
```diff
@@ -1,7 +1,8 @@
 {
     "C_Cpp.autoAddFileAssociations": false,
     "files.associations": {
-        "Make.*": "makefile"
+        "Make.*": "makefile",
+        "**/matlab/**/*.m": "matlab",
```
InsertCreativityHere: This really doesn't matter but:
```suggestion
        "**/matlab/**/*.m": "matlab"
```

bernardnormier (commented):
As I understand it, we have two separate logics for checking "two way only" depending on the language:

(a) check in the generated code
(b) check "from" the generated code by passing a bool argument to a helper function, or sometimes using other arguments for this bool (like the presence/absence of unmarshaling callbacks for return/exception)


And unfortunately this PR mimics this logic for the new oneway-only check, with sometimes checks in the generated code, and sometimes new parameters injected into helper functions. The former is binary compatible while the latter is not.

On main (3.9), I think we should actually perform both the twoway-only and oneway-only checks directly in the generated code (not in a helper function).
And on 3.8, we should perform the new oneway-only check in the generated code, and leave the existing helper/twoway check logic as-is to maintain binary compatibility.

csharp/src/Ice/LocalExceptions.cs
```diff
@@ -955,6 +955,24 @@ public TwowayOnlyException(string operation)
...
+/// <summary>
+/// The exception that is thrown when attempting to invoke a oneway-only operation (an operation with the
+/// ["oneway"] metadata directive) on a twoway proxy.
+/// </summary>
+public sealed class OnewayOnlyException : LocalException
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="OnewayOnlyException" /> class.
+    /// </summary>
+    /// <param name="operation">The operation name.</param>
```
InsertCreativityHere, line 967: I'd prefer a message a little closer to what the other languages write:
```suggestion
    /// <param name="operation">The oneway-only operation name.</param>
```

pepone:
> The former is binary compatible while the latter is not.

One note here is for some languages there is no per operation generated code. For example JavaScript, Python, PHP.

matlab/lib/+Ice/ObjectPrx.m
externl: I've updated the PR to place the checks in the generated code when possible.

matlab/lib/+Ice/OnewayOnlyException.m
externl: Seems to be a split down the middle.

ruby/test/Ice/operations/Oneways.rb
externl: The proxy used in this function should generally be oneway. I'm also going to use it shortly for another test.

.vscode/settings.json
externl: I find it preferable to have the trailing commas.

csharp/src/Ice/LocalExceptions.cs
externl: This is the same language we use for `TwoWayOnlyException` in C++.

cpp/src/slice2swift/Gen.cpp
externl: It's based on an  `isTwoway` bool, if there's a `userException` callback. We actually have an overload of invoke for with a generic for the return type. This unfortunately also has the `isTwoway` bool, which seems wrong.

externl:
I've updated this PR:

- No more helper public functions in C++, C#, and Java. The check is just inline in the generated code. We still have a helper in the PHP, Python, and Ruby, since we call it in several places and there's no generated code.
- Do the check inline in MATLAB and Swift as well. 
- For Python, JS, Ruby, and PHP, we add an optional trailing bool parameter to the "define operation" function, which is backwards compatible with previously generated code. I tested with 3.8. Once we backport to 3.8, we can remove this optionality from the `main` branch.

CHANGELOG.md
```diff
@@ -21,4 +21,9 @@ These are the changes since the [Ice 3.8.1] release.
 
 ### Slice Language Changes
 
+- Added the `["oneway"]` metadata directive for Slice operations. When applied to an operation that does not return
```
bernardnormier: The entry should say that this metadata directive can only be applied to operation that ...

The way it's written, it sounds like you can apply it to any operation.

cpp/include/Ice/LocalExceptions.h
```diff
@@ -856,6 +856,27 @@ namespace Ice
...
+    /// ["oneway"] metadata directive) on a twoway proxy.
```
bernardnormier: Should this be "using" or "with" a twoway proxy?

You really invoke an operation on a remote Ice object, not "on" the proxy.

cpp/src/slice2java/Gen.cpp
bernardnormier, line 1473: Should this be the first statement in the method? I don't see the reason to create the OutgoingAsync first.

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -1292,6 +1293,15 @@ CodeVisitor::visitOperation(const OperationPtr& op)
         out << nl << self << ".iceEndWriteParams(os_);";
     }
 
+    if (onewayOnly)
```
bernardnormier: I would generate this check earlier, before we marshal the parameters.

cpp/src/slice2swift/Gen.cpp
```diff
@@ -1314,6 +1314,15 @@ Gen::TypesVisitor::visitOperation(const OperationPtr& op)
     // Invoke
     //
     out << sp;
```
bernardnormier: It's not clear why we have this sp right after the sb.

cpp/test/Ice/operations/TestAMDI.cpp
```diff
@@ -729,6 +729,12 @@ MyDerivedClassI::opIdempotentAsync(function<void()> response, function<void(exce
     response();
 }
 
+void
+MyDerivedClassI::opOnewayAsync(function<void()> response, function<void(exception_ptr)>, const Current&)
+{
+    response();
```
bernardnormier, line 736: We should test that we received a oneway request. `test(current.requestId == 0);`

cpp/test/Ice/operations/TestI.cpp
```diff
@@ -567,6 +567,11 @@ MyDerivedClassI::opIdempotent(const Ice::Current& current)
     test(current.mode == OperationMode::Idempotent);
 }
 
+void
+MyDerivedClassI::opOneway(const Ice::Current&)
+{
+}
```
bernardnormier, line 574: Ditto

bernardnormier (approved):

pepone (approved):

php/src/Operation.cpp
InsertCreativityHere: You should still `rg` for `Ice_defineOperation_arginfo`. You don't need to bump the argument count (since it's only the number of required arguments). But you still need to add an entry for the new argument.

TL;DR:
Add `ZEND_ARG_INFO(0, oneway)` to line 88 of `php/src/Init.cpp`.
I would put an automatic suggestion if I could : v)

InsertCreativityHere (approved):

PR #5294, opened 2026-04-08
Generate cs:attribute on the constant field, not the wrapper class 

### LINKED ISSUE #5288: cs:attribute on Slice constants
We currently generate the C# attribute on the mapped C# class, when we should generate it on the mapped C# constant.

pepone:
Fix #5288


InsertCreativityHere (approved):
Looks good to me!

bernardnormier (approved):

Issue #5281, opened 2026-04-03
Add back @p to out parameters


bernardnormier:
Issue #5276 (fixed by #5280) was not correct, and we should revert the commit.

`@p` to an out parameter is semantically correct in Slice and when this Slice out parameter is mapped to a language out parameter. 

The actual issue is the mapping, and the context or lack of context while mapping the comment (summary, parameter comment, return comment) in C# and other languages. We should rework the comment-mapping to provide sufficient context when mapping `@p` references. If the referenced parameter is not there, we can map it like a `@c`.

InsertCreativityHere:
Yes that is the other option I mentioned in my PR. I’ll revert and update the language mappings to handle it.

It’s worth noting this actually doesn’t affect most languages anyways, since most lack any special “paramref” syntax. So they get mapped to code-formatting anyways.

InsertCreativityHere:
I reverted my PR (see fb69a9a6a3f71e03edd2c2e030d20dc1e2721796), and opened a new PR with the option 2 (advocated by this issue): #5300

InsertCreativityHere:
`@p` are always allowed now (since they're semantically correct even for `out` parameters).
But the language mapping was changed so that out-parameters always map to code-formatting.
This was done by #5300.

We can improve this for the 2 languages that have `@p` (C++) and `<paramref ... />` (C#), but not sure if it's worth it.
`out` parameters are not mapped as out parameters for async or marshaled-result operations. Additionally `@p` and `@c` are equivalent in C++ anyways. Can always revisit improving the mapping in the future if we want.

PR #5300, opened 2026-04-13
Better Handle `@p` Tags

InsertCreativityHere:
This PR is the alternative to #5280 (which was reverted) and implements #5281.

Instead of disallowing `@p` to link to out-parameters, or emitting bogus links (what we do now),
we map such links to use code-formatting, but also leave it up to each language to decide how to map it, if we want to try and be fancier in the future.

Because in C++ and C#, it's possible to map out-parameter links, but only in certain cases (sync skeleton operations that don't return a result). But, odds are, no one is linking to out-parameters, and checking these conditions during doc-comment formatting is painful. So, for now, out-parameters just always use code-formatting. Can always revisit later.


CHANGELOG.md
```diff
@@ -23,6 +24,9 @@ These are the changes since the [Ice 3.8.1] release.
   do not return data (no return type, out parameters, or exception specification). `OnewayOnlyException` is thrown
   if a `["oneway"]` operation is invoked using a twoway proxy. In languages where exceptions can be thrown synchronously or asynchronously, this exception is always thrown synchronously.
 
-- Removed support for using `@p` tags in doc-comments to reference out parameters.
```
InsertCreativityHere, line 26: This was reintroduced by a bad merge from last week, after being reverted beforehand.

CHANGELOG.md
```diff
@@ -25,4 +26,9 @@ These are the changes since the [Ice 3.8.1] release.
 
 - Removed support for using `@p` tags in doc-comments to reference out parameters.
 
+### C# Changes
+
+- Changed the mapping of `@p` tags which reference out parameters in Slice.
+These now generate a `<c>...</c>` instead of a `<paramref name=... />`.
```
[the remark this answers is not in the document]
InsertCreativityHere: Good catch. A bad merge or something.
This has been fixed now.

cpp/src/Slice/DocCommentParser.h
```diff
@@ -27,12 +27,13 @@ namespace Slice
...
         /// This function is called by the doc-comment parser to map '@p' tags into each language's syntax.
-        /// @param param The mapped name of the parameter that is being referenced.
+        /// @param paramName The mapped name of the parameter that is being referenced.
+        /// @param paramPtr A pointer to the parameter object that is being referenced.
         /// @return A properly formatted parameters reference in the target language. The doc-comment parser will
         /// replace the entire "@p <rawParamName>" string with the returned value.
         //
         // By default we just emit the parameter's name in code formatting.
-        [[nodiscard]] virtual std::string formatParamRef(const std::string& param);
+        [[nodiscard]] virtual std::string formatParamRef(const std::string& paramName, const ParameterPtr& paramPtr);
```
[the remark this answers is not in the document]
InsertCreativityHere: We always handle the null case to be safe.

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -238,9 +238,16 @@ Slice::Csharp::CsharpDocCommentFormatter::formatCode(const string& rawText)
 }
 
 string
-Slice::Csharp::CsharpDocCommentFormatter::formatParamRef(const string& param)
+Slice::Csharp::CsharpDocCommentFormatter::formatParamRef(const string& paramName, const ParameterPtr& paramPtr)
 {
-    return "<paramref name=\"" + param + "\" />";
+    if (paramPtr && paramPtr->isOutParam())
```
pepone, line 243: C# synchronous methods do support regular out parameters

https://docs.zeroc.com/ice/3.8/csharp/csharp-mapping-for-parameters-and-return-values

> An out parameter is mapped to an out parameter with the same name in synchronous proxy and skeleton methods; the type of the parameter is the mapped C# type.

InsertCreativityHere: This becomes tricky though, since we generate both a sync and async skeleton version, so they can't share a doc-comment (if we want paramref in one but not the other).

Guess we also need to check for marshaledResult as well. If you use it, we also don't use out parameters.

InsertCreativityHere:
This also doesn't work for C++ either, where we also sometimes return a result, instead of having the out parameters in the current function... Will require reworking of the comment generation system...

cpp/src/slice2cs/CsUtil.cpp
bernardnormier: I think it's ok to _not_ link to out parameters in C# as well.

CHANGELOG.md
```diff
@@ -23,6 +25,14 @@ These are the changes since the [Ice 3.8.1] release.
...
-- Removed support for using `@p` tags in doc-comments to reference out parameters.
+### C++ Changes
+
+- Changed the mapping of `@p [NAME]` tags which reference out parameters in Slice.
+These now generate `` `[NAME]` `` instead of `@p [NAME]`.
+
+### C# Changes
+
+- Changed the mapping of `@p [NAME]` tags which reference out parameters in Slice.
+These now generate `<c>[NAME]</c>` instead of `<paramref name=[NAME] />`.
```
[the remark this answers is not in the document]
InsertCreativityHere: I fixed the indentation. But we do actually want literal backticks in our Markdown, since we generate backticks in the generated code. So, that is correct as written, and Copilot is misunderstanding what I want.

cpp/src/Slice/DocCommentParser.h
```diff
@@ -27,12 +27,13 @@ namespace Slice
         [[nodiscard]] virtual std::string formatCode(const std::string& rawText);
 
         /// This function is called by the doc-comment parser to map '@p' tags into each language's syntax.
-        /// @param param The mapped name of the parameter that is being referenced.
+        /// @param paramName The mapped name of the parameter that is being referenced.
+        /// @param paramPtr A pointer to the parameter object that is being referenced.
```
pepone: Will be good to clarify when is null expected.

pepone (approved):

cpp/src/Slice/DocCommentParser.h
InsertCreativityHere: Good point, and fixed!

Should of written it anyways to be consistent with `formatLink` and `formatSeeAlso`...

cpp/src/Slice/DocCommentParser.h
```diff
@@ -27,12 +27,13 @@ namespace Slice
         [[nodiscard]] virtual std::string formatCode(const std::string& rawText);
 
         /// This function is called by the doc-comment parser to map '@p' tags into each language's syntax.
-        /// @param param The mapped name of the parameter that is being referenced.
+        /// @param paramName The mapped name of the parameter that is being referenced.
+        /// @param paramPtr A pointer to the parameter that is being referenced, or `nullptr` if it doesn't exist.
         /// @return A properly formatted parameters reference in the target language. The doc-comment parser will
```
externl: Maybe we can expand the comment to note that when `paramPtr ` is not `nullptr`, then `paramName` == `paramPtr->mappedName()`?

externl (approved):

cpp/src/Slice/DocCommentParser.h
InsertCreativityHere: Sure, I updated the comment to say:
```
        /// @param paramName If @p paramPtr is non-null, this is the mapped name of the parameter being referenced.
        /// Otherwise, this is the identifier after '@p', taken verbatim from the doc-comment.
```

cpp/src/slice2cs/CsUtil.cpp
```diff
@@ -238,9 +238,16 @@ Slice::Csharp::CsharpDocCommentFormatter::formatCode(const string& rawText)
...
+    {
+        return formatCode(paramName);
+    }
+    else
+    {
+        return "<paramref name=\"" + paramName + "\" />";
```
bernardnormier, line 249: This doesn't look correct: when paramPtr is null, we want to use formatCode.

cpp/src/Slice/DocCommentParser.h
```diff
@@ -27,12 +27,14 @@ namespace Slice
         [[nodiscard]] virtual std::string formatCode(const std::string& rawText);
 
         /// This function is called by the doc-comment parser to map '@p' tags into each language's syntax.
-        /// @param param The mapped name of the parameter that is being referenced.
+        /// @param paramName If @p paramPtr is non-null, this is the mapped name of the parameter being referenced.
+        /// Otherwise, this is the identifier after '@p', taken verbatim from the doc-comment.
+        /// @param paramPtr A pointer to the parameter that is being referenced, or `nullptr` if it doesn't exist.
```
bernardnormier, line 32: I find this confusing. When paramPtr is null, the Parser issues a warning, right? And we've just generating code despite the warning - because it's a warning, not an error?

I think it would be much cleaner for the caller to NEVER pass a null paramPtr. In this situation, it should simply call formatCode directly.

InsertCreativityHere: > When paramPtr is null, the Parser issues a warning, right?

Yes, we issue a warning about this. The logic behind this decision is if `paramPtr` is null, the user made a typo, and they should fix it. `slicec` shouldn't try to "cover-up" their mistake by using a different mapping for the parameter.

We use what the user gave us, and if it's bogus, `slicec` and the generated code will issue them warnings.
Instead of `slicec` emitting a warning (which could get lost), and outputting generated code hiding their mistake.

cpp/src/slice2cs/CsUtil.cpp
InsertCreativityHere: See https://github.com/zeroc-ice/ice/pull/5300#discussion_r3108968711

I think we should only use code-formatting for things we don't want to (or can't) map (out parameters).
I don't think we should use it when the user has made a typo/mistake.

`nullptr` here means we couldn't find the parameter, i.e. the user has made a typo, or fundamentally misunderstands what `@p` does. Either way, that's their problem, and we shouldn't try and "fix" it for them. Not to mention this "fix" is half-baked. They clearly wanted a `paramref`, not code-formatting.

pepone: I think it would be clear to not call this callback with nullptr, using param with an non-existing params doesn't make much sense, the Slice warning should be enough.

InsertCreativityHere: To be honest, I don't see the big deal about `nullptr`. The rule should be `@p` is mapped to `paramref`.
Why should we should change the mapping in the case where a user made a typo?

This is different than the case where they reference an out-parameter, where there's no parameter to link to, due to our operation mapping (sometime).

----

Note that the behavior with respect to "bad parameters" is unchanged by this PR. RIght now, if you type `@p foo`, we will always generate `<paramref name="foo">`, no questions. We still emit a warning if `foo` isn't a parameter in your operation, but the mapping doesn't care. You asked for a `paramref`, so we give you one.

pepone: >  user made a typo?

Usually compilers fail when user make a typo, so maybe a Slice error is fine here. We are otherwise just generating code for a bogus input.

InsertCreativityHere: Sure, if this was in any other context, 100%, I'd agree. But this is in a doc-comment, and I don't think any compilers will straight up fail due to a bad doc-comment. Rust certainly doesn't:
```
warning: unresolved link to `ErrorKinhhd`                                                                                                                                                                                                                                             
  --> slice-codec\src\error.rs:56:37
   |
56 |     /// Returns the corresponding [`ErrorKinhhd`] that describes this error.
   |                                     ^^^^^^^^^^^ no item named `ErrorKinhhd` in scope
```

Doxygen also just emits a warning but otherwise succeeds. No hard errors.

----

> We are otherwise just generating code for a bogus input.

I don't see that as a problem.
If the user types `@p foobar`, we'll generate `<paramref name="foobar"`, as requested.
If a user types `[cs:identifier:true] class A {}`, well, we're going to generate a class named `true`.
There's a difference between "we're generating bogus code" vs "the user gave us bogus input".

But I think the mapping should be the mapping, not switch to code format some of the time.
And I also don't think it's correct (or expected) for a small doc-comment typo to fail your build.
Are there other compilers which are so strict about doc-comments out of the box?

pepone: > Are there other compilers which are so strict about doc-comments out of the box?

That is not fair because most compile plainly ignore comments, which are processed later by a separate tool. For example C++ compiler don't see code comment at all. But here Slice compilers operate on the comments. This is more like a C++ compiler generating assembly for a variable that doesn't exists...

InsertCreativityHere: Yes, I probably should of written "tools" instead of "compilers" there.
The 2 examples I gave were indeed tools; using `cargo doc` and `doxygen`.

But, wording aside, I think my point still stands.
No compilers I know would issue errors for this (because they ignore it), and even tools that do process doc-comments (`cargo doc` and `doxygen` at least), also only emit warnings, not errors.

----

Clearly still more discussion to have (and I'll create a blog about it), but I'm going to merge this PR as-is.
This PR is just about fixing the behavior for out-parameters.
How we handle bad `@p` is a separate thing, which isn't changed by this PR.
`paramPtr` could always be null, I just added a function which called attention to this fact : v)

And if we change `@p` typos to errors, we should probably do the same for `@link` and `@see` too,
making this a larger issue than what this PR is about.

PR #5302, opened 2026-04-13
Reject Tags Greater than 30 Which are Encoded on a Single Byte

### LINKED ISSUE #5259: Explicitly Handle Optionals Tags of the Form `0bxxx11111`
The current structure for optional tags in Ice is as follows:
- The first 3 bits represent the type of the tag (`F4`, `VSize`, etc.)
- The last 5 bits represent the value of the tag. With 5 bits, this value must be in the range `[0, 31]`
- If the tag value doesn't fit in this range, we encode a value of `30`, followed by the actual value encoded as a `size`.

There are 2 special cases that our decoding handles:
- If the full byte is `255 (0b11111111)`: this is the `OPTIONAL_END_MARKER`, which means there's no more tags.
- Like we already mentioned, if the value is `30 (0bxxx11110`), this means the tag value follows as a `size`.

But this raises the natural question. If the range is `[0, 31]`, and `30` is a special value... what about `31`?
I guess we avoided `31` to avoid confusion with `OPTIONAL_END_MARKER`. But this is actually not possible now that we disallow optional classes.

----

Regardless, if the other side sends a tag byte of the form `0bxxx11111` (other than `0b11111111`), our decoding will fail in unpredictable ways, usually a `MarshalException`, but not always. And there's at least 1 way to achieve a silent failure state.

If the other side sends a tag of `0b10011111`, we only decode 1 byte (since it doesn't end with `30 (0b11110)`).
If the decoder is looking for a tag value smaller than `30` (likely), it sees this tag is too large, and will then rewind.
Instead of rewinding by the `1` byte it read though, it rewinds by `2` bytes (since it always assumes `31` is written on 2 bytes).
This means the decoder is positioned 1 byte earlier than it should be, possibly outside the memory it should be reading from.

It also just so happens that we tolerate an extra byte in the encapsulation. So, this off-by-one shift could go entirely unnoticed:
https://github.com/zeroc-ice/ice/blob/cff87dcb7158f0d9eded8390d715b1ec55016c25/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InputStream.java#L260-L265

InsertCreativityHere:
This PR fixes #5259.

This is how we encode tags in all languages:
 - If a tag is less than 30, we encode it on a single byte,
 - Else we write `0xF0` ('30'), then encode the tag as a size.

But on the decoding side, we have a gap in our logic.
We check if the tag is `30`, and if so, we decode a size. Otherwise we decode the tag as a single byte.
The tag data is sent on 5 bits, so it's possible values are between `0` and `31` (inclusive).
Hence: the decoder is ignoring the case when it receives `31`.

Receiving a tag value of `31` encoded on a single byte is a malformed tag, and we should reject it (what this PR does).
Right now, decoding such a value puts the decoder into various invalid states depending on what tags it was expecting to find.

### DIFF
--- a/cpp/src/Ice/InputStream.cpp
+++ b/cpp/src/Ice/InputStream.cpp
@@ -1144,6 +1144,7 @@ Ice::InputStream::readOptImpl(int32_t readTag, OptionalFormat expectedFormat)
         auto tag = static_cast<int32_t>(v >> 3);
         if (tag > 30)
         {
+            // We check for '> 30' instead of '> 29' because 30 is special sentinel tag, handled by the next block.
             ostringstream os;
             os << "invalid tag '" << tag << "': tags larger than 29 must be encoded as a size";
             throw MarshalException(__FILE__, __LINE__, os.str());
--- a/cpp/src/Ice/InputStream.cpp
+++ b/cpp/src/Ice/InputStream.cpp
@@ -1142,6 +1142,12 @@ Ice::InputStream::readOptImpl(int32_t readTag, OptionalFormat expectedFormat)
 
         auto format = static_cast<OptionalFormat>(v & 0x07); // First 3 bits.
         auto tag = static_cast<int32_t>(v >> 3);
+        if (tag > 30)
+        {
+            ostringstream os;
+            os << "invalid tag '" << tag << "': tags larger than 29 must be encoded as a size";
+            throw MarshalException(__FILE__, __LINE__, os.str());
+        }
         if (tag == 30)
         {
             tag = readSize();
--- a/csharp/src/Ice/InputStream.cs
+++ b/csharp/src/Ice/InputStream.cs
@@ -1555,7 +1555,7 @@ public sealed class InputStream
         }
         catch (ArgumentException ex)
         {
-            throw new MarshalException("Invalid UTF8 string", ex);
+            throw new MarshalException("Invalid UTF8 string.", ex);
         }
     }
 
@@ -1859,7 +1859,8 @@ public sealed class InputStream
             int tag = v >> 3;
             if (tag > 30)
             {
-                throw new MarshalException($"invalid tag '{tag}': tags larger than 29 must be encoded as a size");
+                // We check for '> 30' instead of '> 29' because 30 is special sentinel tag, handled by the next block.
+                throw new MarshalException($"Invalid tag '{tag}': tags larger than 29 must be encoded as a size.");
             }
             if (tag == 30)
             {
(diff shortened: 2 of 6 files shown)


externl (approved):
Looks good to me!

csharp/src/Ice/InputStream.cs
```diff
@@ -1857,6 +1857,10 @@ private bool readOptImpl(int readTag, OptionalFormat expectedFormat)
 
             var format = (OptionalFormat)(v & 0x07); // First 3 bits.
             int tag = v >> 3;
+            if (tag > 30)
+            {
+                throw new MarshalException($"invalid tag '{tag}': tags larger than 29 must be encoded as a size");
```
bernardnormier: The normal style for C# error message is capitalized sentences.

cpp/src/Ice/InputStream.cpp
```diff
@@ -1142,6 +1142,12 @@ Ice::InputStream::readOptImpl(int32_t readTag, OptionalFormat expectedFormat)
 
         auto format = static_cast<OptionalFormat>(v & 0x07); // First 3 bits.
         auto tag = static_cast<int32_t>(v >> 3);
+        if (tag > 30)
```
bernardnormier, line 1145: The disconnect between the check (> 30) and the message (larger than 29) sounds like a typo.

Assuming it's intentional, you should add a comment explaining this disconnect.

InsertCreativityHere: It is indeed intentional. I'll add a comment!
Any value `> 29` needs to be encoded as a `size`. And `30` is a sentinel value indicating: "you need to read a size" to the decoder.

csharp/src/Ice/InputStream.cs
InsertCreativityHere: Many of the exceptions we throw in this file are not following this...
I'll also update them to be proper sentences.

csharp/src/Ice/InputStream.cs
```diff
@@ -1555,7 +1555,7 @@ public string readString()
         }
         catch (ArgumentException ex)
         {
-            throw new MarshalException("Invalid UTF8 string", ex);
+            throw new MarshalException("Invalid UTF8 string.", ex);
```
InsertCreativityHere, line 1558: Many of the error messages in this file were not proper sentences.
And if we don't fix them when we see them, they will never be fixed.

csharp/src/Ice/InputStream.cs
```diff
@@ -1876,7 +1881,7 @@ private bool readOptImpl(int readTag, OptionalFormat expectedFormat)
             {
                 if (format != expectedFormat)
                 {
-                    throw new MarshalException("invalid optional data member `" + tag + "': unexpected format");
```
InsertCreativityHere, line 1879: I also removed some lingering ``` `...' ``` quotes, replacing the tick with a single-quote.

bernardnormier (approved):

PR #5304, opened 2026-04-13
swift: Use getUnqualified for OnewayOnlyException in slice2swift

externl:


InsertCreativityHere (approved):
Looks good.
I'm not sure how it works right now without any qualification at all...

externl:
> Looks good. I'm not sure how it works right now without any qualification at all...

Pretty sure its because we `import Ice`

pepone (approved):

PR #5310, opened 2026-04-16
Backport ZeroC.Ice.Slice.Tools from main

externl:
Backport 5efc47c975e from `main` to `3.8`. With this `ZeroC.Ice.Slice.Tools` and `slice2cs` are in sync with `main`.

### DIFF
--- a/csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
+++ b/csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
@@ -152,7 +152,8 @@ public abstract class SliceCompilerTask : ToolTask
 
                     // Save the dependencies for each source to a dependency file
                     //
-                    // Foo.ice -> $(OutputDir)/SliceCompile.Foo.d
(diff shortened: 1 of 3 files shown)


csharp/src/ZeroC.Ice.Slice.Tools/Common/SliceCompilerTask.cs
```diff
@@ -161,8 +162,9 @@ protected override int ExecuteTool(string pathToTool, string responseFileCommand
                                 new XElement("options",
                                     GetOptions(source).Select(e => new XElement(e.Key, e.Value))))));
```
InsertCreativityHere, line 163: Fun fact, this is the 2nd longest run of contiguous `)` (6) in all of our repos!
1st place goes to our Python Makefile:
https://github.com/zeroc-ice/ice/blob/d546f7cbd18e2caa9feaa118eef055408c562b02/python/python/Makefile#L4

InsertCreativityHere (approved):
Looks good to me!

pepone (approved):

PR #5311, opened 2026-04-17
Make oneway param required in operations

externl:
- PHP: Increase required arg count to 10 and remove optional '|b' separator in format string
- Python: Remove optional '|p' separator in format string and drop default initializer
- Ruby: Require exactly 9 arguments instead of 8..9, remove fallback default for onewayOnly


ruby/src/IceRuby/Operation.cpp
```diff
@@ -86,20 +86,17 @@ static const rb_data_type_t IceRuby_OperationType = {
     .flags = RUBY_TYPED_FREE_IMMEDIATELY,
 };
 
-// Accepts 8 or 9 arguments. The 9th argument (onewayOnly) is optional and defaults to false for backward compatibility
-// with code generated by older versions of slice2rb.
 extern "C" VALUE
 IceRuby_defineOperation(int argc, VALUE* argv, VALUE /*self*/)
```
InsertCreativityHere: This should probably go back to being a normal function which directly takes arguments, instead of the more unchecked and ultra-flexible signature it's using now. So:

1) Changing this:
`rb_define_module_function(iceModule, "__defineOperation", CAST_METHOD(IceRuby_defineOperation), -1);`
to
`rb_define_module_function(iceModule, "__defineOperation", CAST_METHOD(IceRuby_defineOperation), 9);`
To tell Ruby "hey pass this 9 arguments" instead of "I want to use argc, argv".

2) Changing the signature to actually match that. Changing this:
`IceRuby_defineOperation(int argc, VALUE* argv, VALUE /*self*/)`
to
`IceRuby_defineOperation(VALUE /*self*/, VALUE sliceName, VALUE mappedName, VALUE mode, VALUE format, VALUE inParams, VALUE outParams, VALUE returnType, VALUE exceptions, VALUE onewayOnly)`
but with better formatting.

This is more readable, more safe, and lets you delete your `argc != 9` check.

InsertCreativityHere (approved):
Looks good to me!

pepone (approved):

PR #5312, opened 2026-04-17
Spring Cleaning of the Slice Compilers

InsertCreativityHere:
This PR is an exercise in consistency: a cleaning-up of small things that I've been wanting to do a little while:
- Changed the name of `UnitPtr` parameters to `unit` everywhere (instead of a mix of `u`, `un`, `ut`, `p`, and `unit`)
- Renamed `optional()` to `isOptional()` for correctness and to avoid ambiguity with `std::optional`
- Removes unnecessary overqualification (`Slice::` and `std::` in source files which are `using Slice` or `using std`)
  - Except for function calls like `std::move`, `std::stoll`, etc.
  - And except for `Slice::Exception`, which we always qualify to avoid any confusion with real exceptions
- Removed any leading `::` global scope markers (like `::IceInternal::Output` -> `IceInternal::Output`)
- Adds missing parameter names to the header files
  - Except for deleted functions, copy/move constructors, and the visitXXX functions

### DIFF
--- a/cpp/src/slice2cs/IceRpcVisitors.h
+++ b/cpp/src/slice2cs/IceRpcVisitors.h
@@ -56,7 +56,7 @@ namespace Slice::IceRpc
     class SkeletonVisitor final : public CsVisitor
     {
     public:
-        SkeletonVisitor(IceInternal::Output&);
+        SkeletonVisitor(IceInternal::Output& output);
 
         bool visitModuleStart(const ModulePtr&) final;
 
--- a/cpp/src/slice2java/Gen.cpp
+++ b/cpp/src/slice2java/Gen.cpp
@@ -892,7 +892,7 @@ Slice::JavaVisitor::writeResultTypeMarshalUnmarshalCode(
 
     for (const auto& param : op->sortedReturnAndOutParameters(retval))
     {
-        const bool isOptional = param->optional();
+        const bool isOptional = param->isOptional();
         const string name = paramPrefix + param->mappedName();
         const string patchParams = isMarshaling ? getPatcher(param->type(), package, name) : "";
 
@@ -970,7 +970,7 @@ Slice::JavaVisitor::writeResultType(
             }
         }
         out << nl << "public "
-            << typeToString(outParam->type(), TypeModeIn, package, outParam->getMetadata(), true, outParam->optional())
+            << typeToString(outParam->type(), TypeModeIn, package, outParam->getMetadata(), true, outParam->isOptional())
             << ' ' << outParam->mappedName() << ';';
         out << sp;
     }
@@ -1058,7 +1058,7 @@ Slice::JavaVisitor::writeResultType(
                         package,
                         outParam->getMetadata(),
                         true,
-                        !generateMandatoryOnly && outParam->optional()) +
+                        !generateMandatoryOnly && outParam->isOptional()) +
                     " " + outParam->mappedName());
         }
         out << epar;
@@ -1079,7 +1079,7 @@ Slice::JavaVisitor::writeResultType(
         {
             const string name = outParam->mappedName();
             out << nl << "this." << name << " = ";
-            if (outParam->optional() && generateMandatoryOnly)
+            if (outParam->isOptional() && generateMandatoryOnly)
             {
                 out << ofFactory(outParam->type()) << "(" << name << ");";
             }
@@ -1181,10 +1181,10 @@ Slice::JavaVisitor::writeMarshaledResultType(
                     package,
                     outParam->getMetadata(),
                     true,
-                    outParam->optional()) +
+                    outParam->isOptional()) +
                 " " + outParam->mappedName());
 
-        hasOpt = hasOpt || outParam->optional();
+        hasOpt = hasOpt || outParam->isOptional();
     }
     out << currentParam << epar;
     out << sb;
@@ -1263,7 +1263,7 @@ Slice::JavaVisitor::writeMarshaledResultType(
         }
         for (const auto& outParam : outParams)
         {
-            if (outParam->optional())
+            if (outParam->isOptional())
             {
                 out << ofFactory(outParam->type()) + "(" + outParam->mappedName() + ")";
             }
@@ -1528,8 +1528,8 @@ Slice::JavaVisitor::writeMarshalProxyParams(
             out,
             package,
             param->type(),
-            param->optional(),
-            param->optional() && optionalMapping,
+            param->isOptional(),
+            param->isOptional() && optionalMapping,
             param->tag(),
             "iceP_" + param->mappedName(),
             true,
@@ -1580,7 +1580,7 @@ Slice::JavaVisitor::writeUnmarshalProxyResults(Output& out, const string& packag
         else
         {
             assert(outParams.size() == 1);
-            isOptional = outParams.front()->optional();
+            isOptional = outParams.front()->isOptional();
             type = outParams.front()->type();
             tag = outParams.front()->tag();
             metadata = outParams.front()->getMetadata();
@@ -1659,7 +1659,7 @@ Slice::JavaVisitor::writeMarshalServantResults(
         else
         {
             assert(params.size() == 1);
-            isOptional = params.front()->optional();
+            isOptional = params.front()->isOptional();
             type = params.front()->type();
             tag = params.front()->tag();
             metadata = params.front()->getMetadata();
@@ -1705,7 +1705,7 @@ Slice::JavaVisitor::writeThrowsClause(
 void
 Slice::JavaVisitor::writeMarshalDataMember(Output& out, const string& package, const DataMemberPtr& member, int& iter)
 {
-    const bool isOptional = member->optional();
+    const bool isOptional = member->isOptional();
     const bool forStruct = dynamic_pointer_cast<Struct>(member->container()) != nullptr;
     const string memberName = (forStruct ? "this." : "") + member->mappedName();
 
@@ -1742,7 +1742,7 @@ void
 Slice::JavaVisitor::writeUnmarshalDataMember(Output& out, const string& package, const DataMemberPtr& member, int& iter)
 {
     const TypePtr& type = member->type();
-    const bool isOptional = member->optional();
+    const bool isOptional = member->isOptional();
     const bool forStruct = (bool)dynamic_pointer_cast<Struct>(member->container());
     const string stream = forStruct ? "istr" : "istr_";
     assert(!isOptional || !forStruct);           // optional members aren't allowed in structs.
@@ -1879,7 +1879,7 @@ Slice::JavaVisitor::writeDataMemberInitializers(Output& out, const DataMemberLis
         TypePtr t = member->type();
         if (member->defaultValue())
         {
-            if (member->optional())
+            if (member->isOptional())
             {
                 string capName = member->mappedName();
                 capName[0] = static_cast<char>(toupper(static_cast<unsigned char>(capName[0])));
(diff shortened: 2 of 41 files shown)


cpp/src/slice2swift/SwiftUtil.h
```diff
@@ -10,81 +10,81 @@ using StringPairList = std::list<std::pair<std::string, std::string>>;
...
-
-    bool usesMarshalHelper(const TypePtr&);
-    void writeMarshalInParams(::IceInternal::Output&, const OperationPtr&);
-    void writeMarshalAsyncOutParams(::IceInternal::Output&, const OperationPtr&);
-    void writeUnmarshalInParams(::IceInternal::Output&, const OperationPtr&);
-    void writeUnmarshalOutParams(::IceInternal::Output&, const OperationPtr&);
-    void writeUnmarshalUserException(::IceInternal::Output& out, const OperationPtr&);
-    void writeSwiftAttributes(::IceInternal::Output&, const MetadataList&);
+        ::IceInternal::Output& out,
+        const TypePtr& type,
```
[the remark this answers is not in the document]
InsertCreativityHere: My regex for catching these wasn't perfect I guess.
Fixed.

cpp/src/slice2cs/IceRpcVisitors.h
```diff
@@ -56,7 +56,7 @@ namespace Slice::IceRpc
     class SkeletonVisitor final : public CsVisitor
     {
     public:
-        SkeletonVisitor(IceInternal::Output&);
+        SkeletonVisitor(IceInternal::Output& output);
```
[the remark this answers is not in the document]
InsertCreativityHere: This is because the Swift visitors have a private field named `out`, which should be named `_out`.
Will fix in another PR.

cpp/src/slice2swift/SwiftUtil.h
```diff
@@ -10,81 +10,81 @@ using StringPairList = std::list<std::pair<std::string, std::string>>;
...
+        bool marshal,
+        std::int32_t tag = -1);
+
+    bool usesMarshalHelper(const TypePtr& type);
+    void writeMarshalInParams(::IceInternal::Output& out, const OperationPtr& op);
+    void writeMarshalAsyncOutParams(::IceInternal::Output& out, const OperationPtr& op);
+    void writeUnmarshalInParams(::IceInternal::Output& out, const OperationPtr& op);
+    void writeUnmarshalOutParams(::IceInternal::Output& out, const OperationPtr& op);
+    void writeUnmarshalUserException(::IceInternal::Output& out, const OperationPtr& op);
+    void writeSwiftAttributes(::IceInternal::Output& out, const MetadataList& metadata);
```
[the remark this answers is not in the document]
InsertCreativityHere: My regex missed `::` that came right after a `(`. Fixed.

cpp/src/slice2java/Gen.cpp
```diff
@@ -969,9 +969,11 @@ Slice::JavaVisitor::writeResultType(
                 out << nl << " */";
             }
         }
-        out << nl << "public "
-            << typeToString(outParam->type(), TypeModeIn, package, outParam->getMetadata(), true, outParam->optional())
-            << ' ' << outParam->mappedName() << ';';
+
+        string typeString =
+            typeToString(outParam->type(), TypeModeIn, package, outParam->getMetadata(), true, outParam->isOptional());
```
InsertCreativityHere, line 974: Moved into a variable to avoid needing to split the function call onto 7 lines after `option()` changed to `isOptional()` (pushing it over the 120 limit).

cpp/src/slice2java/Gen.cpp
```diff
@@ -4560,9 +4562,11 @@ Slice::Gen::SkeletonVisitor::visitInterfaceDefEnd(const InterfaceDefPtr& p)
 
             for (const auto& value : values)
             {
+                string typeString =
+                    typeToString(value->type(), TypeModeIn, package, value->getMetadata(), true, value->isOptional());
```
InsertCreativityHere, line 4566: Same. Pulled into a separate variable to keep within 120 columns and avoid this call taking 7 lines instead of 2.

cpp/src/slice2java/JavaUtil.h
```diff
@@ -124,26 +124,20 @@ namespace Slice::Java
...
-    void validateJavaMetadata(const UnitPtr&);
+    void validateJavaMetadata(const UnitPtr& unit);
 
-    class JavaOutput final : public ::IceInternal::Output
+    class JavaOutput final : public IceInternal::Output
     {
     public:
         JavaOutput();
 
-        //
```
InsertCreativityHere, line 134: Turned this into a real doc-comment. Technically unrelated to this PR though.

cpp/src/slice2js/Gen.cpp
```diff
@@ -648,7 +648,7 @@ Slice::Gen::Gen(const string& base, const string& dir, bool typeScript)
     }
 }
 
-Slice::Gen::Gen(const string& base, const string& /*dir*/, bool typeScript, ostream& out)
```
InsertCreativityHere, line 651: This is an actual code change, this constructor had a dead parameter, which I removed.
`dir` isn't meaningful when we're writing to `stdout` (what this constructor does).

Since the `ostream` feels more important than the `typeScript` bool, I also changed the order a little.

cpp/src/slice2php/Main.cpp
```diff
@@ -313,15 +313,15 @@ CodeVisitor::visitClassDefStart(const ClassDefPtr& p)
         _out << "array(";
         for (auto q = members.begin(); q != members.end(); ++q)
         {
+            const bool isOptional = (*q)->isOptional();
```
InsertCreativityHere, line 316: Stored this in a variable instead of needing to call it twice.

cpp/src/slice2php/Main.cpp
```diff
@@ -632,15 +632,15 @@ CodeVisitor::visitExceptionStart(const ExceptionPtr& p)
         _out << "array(";
         for (auto dmli = members.begin(); dmli != members.end(); ++dmli)
         {
+            const bool isOptional = (*dmli)->isOptional();
```
InsertCreativityHere, line 635: Stored this in a variable instead of needing to call it twice.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -2574,7 +2575,8 @@ Slice::Python::CodeVisitor::writeOpDocstring(const OperationPtr& op, MethodKind
         {
             assert(outParams.size() == 1);
             const auto& param = outParams.front();
-            out << nl << typeToTypeHintString(param->type(), param->optional(), p, methodKind == MethodKind::Dispatch);
+            out << nl;
+            out << typeToTypeHintString(param->type(), param->isOptional(), p, methodKind == MethodKind::Dispatch);
```
InsertCreativityHere, line 2579: The extra 2 characters from `isOptional` pushed this over the 120 limit.

cpp/src/slice2rb/RubyUtil.h
```diff
@@ -9,13 +9,10 @@
...
     /// Generates Ruby code for the provided translation unit.
-    void generate(
-        const Slice::UnitPtr& unit,
-        bool all,
-        const std::vector<std::string>& includePaths,
-        IceInternal::Output& out);
+    void
+    generate(const UnitPtr& unit, bool all, const std::vector<std::string>& includePaths, IceInternal::Output& out);
 
-    /// This enum is used by ::getMappedName to control the casing of the identifier it returns.
```
InsertCreativityHere, line 18: This file even used leading `::` in the doc-comments.
I removed these and put the function names in backticks.

cpp/src/slice2js/Gen.cpp
```diff
@@ -2139,9 +2139,9 @@ Slice::Gen::TypeScriptImportVisitor::writeImports()
 }
 
 string
-Slice::Gen::TypeScriptVisitor::importPrefix(const string& s) const
+Slice::Gen::TypeScriptVisitor::importPrefix(const string& scopedName) const
```
InsertCreativityHere, line 2142: Changed the parameter name from `s` to `scopedName`, since `s` is not... the greatest name.

cpp/src/slice2js/Gen.h
```diff
@@ -22,32 +22,33 @@ namespace Slice
...
-        ::IceInternal::Output& _out;
+        IceInternal::Output& _out;
         std::string _jsModule;
     };
 
     class IncludeAggregationVisitor : public ParserVisitor
     {
     public:
-        bool visitUnitStart(const UnitPtr& unit) final;
-        void visitModuleEnd(const ModulePtr& module) final;
```
InsertCreativityHere, line 50: This is the only place where we include a parameter name in `visitXXX` functions. So. For consistency with the 99% of other overrides which don't have a name, I removed them here.

Maybe we should add parameter names to all the `visitXXX` functions though. On the fence.

cpp/src/slice2js/Gen.h
```diff
@@ -98,13 +99,13 @@ namespace Slice
     class Gen final
     {
     public:
-        Gen(const std::string&, const std::string&, bool);
+        Gen(const std::string& base, const std::string& dir, bool typeScript);
 
-        Gen(const std::string&, const std::string&, bool, std::ostream&);
```
InsertCreativityHere, line 103: The 2nd `string` parameter was dead.

cpp/src/slice2js/Gen.h
```diff
@@ -150,23 +152,21 @@ namespace Slice
...
         public:
-            ExportsVisitor(::IceInternal::Output&, std::set<std::string>);
+            ExportsVisitor(IceInternal::Output& out, std::set<std::string> importedModules);
 
             bool visitModuleStart(const ModulePtr&) final;
 
             [[nodiscard]] std::set<std::string> exportedModules() const;
 
         private:
-            std::string encodeTypeForOperation(const TypePtr&);
```
InsertCreativityHere, line 160: This function is never called, and never even implemented!
Deleted.

cpp/src/slice2java/JavaUtil.h
```diff
@@ -124,26 +124,20 @@ namespace Slice::Java
...
-        // necessary.
-        //
-        void openClass(const std::string&, const std::string&, const std::string& = std::string());
+        /// Open a file to hold the source for a Java class. After successfully opening the file,
+        /// the function invokes `printHeader()` and then emits a "package" statement if necessary.
+        /// Intermediate directories will be created as necessary to open the file in the package.
+        /// @param cls The class name (including an optional leading package).
+        /// @param prefix Specifies a directory prefix in which to generate the class.
+        /// @param sliceFile The Slice file that the class is being generated from.
+        void openClass(const std::string& cls, const std::string& prefix, const std::string& sliceFile = std::string());
```
pepone: No directly related to the PR, but why is `sliceFile` optional, seems we only called once and pass the parameter:

```
jose@mac slice2java % rg openClass
JavaUtil.h
146:        void openClass(const std::string&, const std::string&, const std::string& = std::string());

JavaUtil.cpp
807:Slice::Java::JavaOutput::openClass(const string& cls, const string& prefix, const string& sliceFile)
874:        out->openClass(qualifiedEntity, _dir, sliceFile);
```

pepone (approved):

externl (approved):

cpp/src/slice2java/JavaUtil.h
InsertCreativityHere: No idea, I removed the useless default value.
Maybe leftover from before we added a separate constructor for `stdout`? Just a guess.

PR #5314, opened 2026-04-21
Move Code-Gen Visitors out of 'Gen' Classes

InsertCreativityHere:
The code-gen visitors for `slice2cpp`, `slice2java`, `slice2js`, and `slice2swift` all have this structure where we have a single top-level class named `Gen`, and all the visitors are nested classes inside `Gen`.

I don't see any reason for this structure though, none of them use any instance fields or methods on `Gen`...
So this PR moves the visitor classes out of `Gen` and makes them top-level, normal classes in the `Slice` namespace.
Since GitHub captures this poorly, all this means is:

1) Dedented the classes in their header files, after moving them out of `Gen`.
2) Removed the extra `Gen::` qualifier from the function implementations.

### DIFF
--- a/cpp/src/slice2cpp/Gen.cpp
+++ b/cpp/src/slice2cpp/Gen.cpp
@@ -1183,10 +1183,7 @@ Slice::SliceLoaderVisitor::visitExceptionStart(const ExceptionPtr& p)
     return false;
 }
 
-Slice::ProxyVisitor::ProxyVisitor(Output& h, Output& c, string dllExport)
-    : H(h),
-      C(c),
-      _dllExport(std::move(dllExport))
+Slice::ProxyVisitor::ProxyVisitor(Output& h, Output& c, string dllExport) : H(h), C(c), _dllExport(std::move(dllExport))
 {
 }
 
@@ -2481,11 +2478,7 @@ Slice::DataDefVisitor::printFields(const DataMemberList& fields, bool firstField
     }
 }
 
-Slice::InterfaceVisitor::InterfaceVisitor(
-    IceInternal::Output& h,
-    IceInternal::Output& c,
-    string dllExport,
-    bool async)
+Slice::InterfaceVisitor::InterfaceVisitor(IceInternal::Output& h, IceInternal::Output& c, string dllExport, bool async)
     : H(h),
       C(c),
       _dllExport(std::move(dllExport)),
--- a/cpp/src/slice2cpp/Gen.cpp
+++ b/cpp/src/slice2cpp/Gen.cpp
@@ -859,7 +859,7 @@ Slice::Gen::getSourceExt(string_view file, const UnitPtr& unit)
     return dc->getMetadataArgs("cpp:source-ext").value_or("");
 }
 
-Slice::Gen::ForwardDeclVisitor::ForwardDeclVisitor(Output& h, Output& c, string dllExport)
+Slice::ForwardDeclVisitor::ForwardDeclVisitor(Output& h, Output& c, string dllExport)
     : H(h),
       C(c),
       _dllExport(std::move(dllExport))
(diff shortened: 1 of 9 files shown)


cpp/src/slice2cpp/Gen.cpp
```diff
@@ -859,17 +859,17 @@ Slice::Gen::getSourceExt(string_view file, const UnitPtr& unit)
...
       _dllExport(std::move(dllExport))
 {
 }
 
 bool
-Slice::Gen::ForwardDeclVisitor::visitModuleStart(const ModulePtr& p)
+Slice::ForwardDeclVisitor::visitModuleStart(const ModulePtr& p)
 {
-    _useWstring = setUseWstring(p, _useWstringHist, _useWstring);
+    _useWstring = Gen::setUseWstring(p, _useWstringHist, _useWstring);
```
InsertCreativityHere, line 872: The C++ code does actually call these `setUseWString` and `resetUseWString` functions on `Gen`.
They're static functions though, not instance.

So totally fine to call, but now that these classes aren't in `Gen`, we pick up the `Gen` scope here.

cpp/src/slice2cpp/Gen.h
```diff
@@ -53,159 +53,159 @@ namespace Slice
         std::vector<std::string> _includePaths;
         std::string _dllExport;
         std::string _dir;
+    };
```
InsertCreativityHere, line 56: The diff is really struggling here.
None of the code changed. I only dedented it, and moved the closing `};` for `class Gen` up from the bottom of the file, to right above this comment.

cpp/src/slice2java/Gen.cpp
```diff
@@ -2329,7 +2329,7 @@ Slice::JavaVisitor::writeParamDocComments(IceInternal::Output& out, const DataMe
     }
 }
 
-Slice::Gen::Gen(string base, string dir) : _base(std::move(base)), _dir(std::move(dir)) {}
```
InsertCreativityHere, line 2332: The only code-change this PR makes is right here.
`_base` was a dead field on `Gen` in `slice2java`. We set it in the constructor, then never touch it again.

I removed it.

cpp/src/slice2java/Gen.h
```diff
@@ -192,53 +192,52 @@ namespace Slice
         void generate(const UnitPtr& unit);
 
     private:
-        std::string _base;
         std::string _dir;
+    };
+
+    class TypesVisitor final : public JavaVisitor
```
InsertCreativityHere, line 198: Once again, no changes other than dedenting these lines.

cpp/src/slice2js/Gen.h
```diff
@@ -114,148 +114,148 @@ namespace Slice
         std::string _fileBase;
         bool _useStdout;
         bool _typeScript;
+    };
```
InsertCreativityHere, line 117: You get the idea. Dedent only.

cpp/src/slice2swift/Gen.cpp
```diff
@@ -83,7 +83,7 @@ Gen::Gen(const string& base, const string& dir)
     _out << nl << "import Foundation";
 }
```
InsertCreativityHere, line 85: Every other compiler includes the `Slice::` top-level-namespace in function implementations, to distinguish between class functions and anonymous-namespace functions.

Don't know why `slice2swift` didn't, but I changed it to include the `Slice::` for consistency with the other compilers.

cpp/src/slice2cpp/Gen.cpp
pepone: Did you consider moving this to Slice namespace as regular functions, in CPlusPlusUtil file?

pepone (approved):

cpp/src/slice2cpp/Gen.cpp
InsertCreativityHere: Yes, after this I'm going to open another PR to do exactly this, and also a couple other small fixes to finish off my cleaning. Probably could of moved it in this PR, but was worried it'd be harder to review given how many lines this already touches!

externl (approved):

PR #5315, opened 2026-04-22
More minor cleanup in the Slice compilers

InsertCreativityHere:
This PR:
- Moves some functions which are only used in one file/scope to the place where they're used.
- Renames `Operation::interface()` to `Operation::parentInterface()`
  More on this later
- Moves the contents of `CPlusPlusUtil` to the `Slice::Cpp` namespace, instead of just the `Slice` namespace.
  This is what we do for all the other `xxxUtil` files. C++ was the only exception.
- Updates some constructors to use the `Slice::baseName` util function which was added, but only used for python.
- Fixes a variable name in `slice2swift`. We had a private variable named `out`, which I updated to `_out` (since it's private).
   We omitted the underscore to avoid conflicting with the _other_ `_out` on `Gen`, from things were nested in it...

----

Why change `interface()` to `parentInterface()`?

Windows platforms actually defines `interface` to be `struct`. No seriously. If you type "interface", right click it, and "Go To Definition" on Windows, it takes you to these two lines of code in the system include files:
```
#define __STRUCT__ struct
#define interface __STRUCT__
```

Because of this, my VsCode C++ extension has always hated our use of the identifier `interface`, and will show any files using it as having errors, even though they compile fine. But it's annoying that I can't tell if a file really has errors, or just this `interface` issue.
By changing the name, for the first time in a while, my VsCode shows no errors in our C++ code.

### DIFF
--- a/cpp/src/Slice/Parser.cpp
+++ b/cpp/src/Slice/Parser.cpp
@@ -111,6 +111,58 @@ namespace
         }
         return true;
     }
+
+    /// Checks an identifier for illegal syntax and reports any errors that are present.
+    bool reportIllegalSuffixOrUnderscore(const string& identifier)
+    {
+        // check whether the identifier is scoped
+        size_t scopeIndex = identifier.rfind("::");
+        bool isScoped = scopeIndex != string::npos;
+        string name;
+        if (isScoped)
+        {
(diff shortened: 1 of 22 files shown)


pepone (approved):
Looks good!

externl (approved):

PR #5317, opened 2026-04-30
Fix disabling of always-activation servers on failure

### LINKED ISSUE #5306: activation="always" IceBox server is not retried after a transient startup failure and remains Inactive after registry/dependency recovery
### Version

- Ice Version : 3.7.5
- Os：linux

### Describe the bug

Reproduction steps

Stop icegridregistry
Kill the icebox process for PsServerps1
icegridnode reconnects and attempts to reactivate PsServerps1
During startup, service PsIce throws Ice::ConnectionRefusedException
The server process exits with code 1
PsServerps1 returns to Inactive
Even after the registry/dependency is fully available again, icegridnode does not try to activate the server again automatically

### Expected behavior

Since the server is configured with activation="always", we expected IceGrid to retry activation after a transient startup failure once the registry/dependency becomes available again, or at least provide a supported mechanism for this recovery path.

### Additional context

_No response_

externl:
Fixes #5306

Since its inception, IceGrid's `always` activation mode has had an undocumented (as far as I can tell) edge case in how it interacts with the [IceGrid.Node.DisableOnFailure](https://docs.zeroc.com/ice/3.8/csharp/icegrid-1#IceGrid.*-IceGrid.Node.DisableOnFailure) property.

If an "always" server failed during activation we disabled it. 

https://github.com/zeroc-ice/ice/blob/9311976c46dab00e462e70ca4e723823adb990e3/cpp/src/IceGrid/ServerI.cpp#L1397

As stated in the issue linked issue 

>  The current behavior contradicts the documented semantics of DisableOnFailure=0 ("the node does not disable servers") and is surprising — there's no good reason to treat a startup failure differently from a post-startup crash for always servers.

Simplify disableOnFailure() to only check the DisableOnFailure setting, remove logic that incorrectly disabled servers with activation="always" on startup failure

If approved, we should also decide if we want to backport this fix to 3.8.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ These are the changes since the [Ice 3.8.1] release.
 ### General Changes
 
 - Removed the 'ice2slice' compiler
+- Fixed IceGrid so servers with activation="always" are no longer disabled after a startup failure when
+  IceGrid.Node.DisableOnFailure is set to 0 (the default).
 
 ### Slice Language Changes
 
--- a/cpp/src/IceGrid/ServerI.cpp
+++ b/cpp/src/IceGrid/ServerI.cpp
@@ -1390,11 +1390,9 @@ ServerI::disableOnFailure()
     }
 
     //
-    // If disable on failure is configured or if the activation mode
-    // is always and the server wasn't active at the time of the
-    // failure we disable the server.
+    // Disable the server only when disable-on-failure is configured.
     //
-    if (_disableOnFailure != 0s || (_activation == Always && (_state == Activating || _state == WaitForActivation)))
+    if (_disableOnFailure != 0s)
     {
         _previousActivation = _activation;
         _activation = Disabled;
--- a/cpp/test/IceGrid/activation/AllTests.cpp
+++ b/cpp/test/IceGrid/activation/AllTests.cpp
@@ -521,6 +521,24 @@ allTests(Test::TestHelper* helper)
         }
         pingers.resize(0);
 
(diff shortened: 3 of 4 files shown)


pepone (approved):
Looks good, I think we should include this in 3.8.2

externl:
> Looks good, I think we should include this in 3.8.2

The only "risk" is if someone was relying on this behavior for some reason. 🤷‍♂️

InsertCreativityHere (approved):
I think this is truer to the intent of `always` and is closer to the behavior I'd expect.

The downside is that if a server has some issue, it will continually fail to start in a loop forever. But, I think that's fine. The user can intervene in such cases.

bernardnormier:
I don't like this fix. As pointed out by @InsertCreativityHere, this will result in the Node trying over and over to restart the server (with 500ms pause between attempts).

For example, you've misconfigured the database path for your Always server, and IceGrid attempts to start it continuously. Not great. 

I think the root issue for #5306 is the Node is attempting to start an Always server when the (only) Registry is down. This won't work, and we should not attempt this bound-to-fail start. With other activation modes, the Registry always drives the activation so this is not an issue.

A better fix would be to somehow wait for the registry to be available (some NodeSession to be established) when we restart an activation = Always server.

See: 
https://github.com/zeroc-ice/ice/blob/9311976c46dab00e462e70ca4e723823adb990e3/cpp/src/IceGrid/ServerI.cpp#L2716

Naturally, we should also document the current behavior: an Always server is disabled when it fails during startup.

externl:
Ok, I'll close this PR.

PR #5321, opened 2026-05-11
js: bump dependencies and fix test tsconfig for TS 6

externl:


pepone:
I was wondering if there is any tool that can help audit for malicious updates.

I come across https://socket.dev/

There is also ` npm-check-updates --cooldown 7`

Even if we only have build dependencies, and malicious update my try to inject code, execute commands, ...

externl:
> I was wondering if there is any tool that can help audit for malicious updates.
> 
> I come across https://socket.dev/
> 
> There is also ` npm-check-updates --cooldown 7`
> 
> Even if we only have build dependencies, and malicious update my try to inject code, execute commands, ...

Yeah I've seen this type of thing mentioned a bit recently. Also saw this from this morning: https://safedep.io/mass-npm-supply-chain-attack-tanstack-mistral/.

I'll look into setting a min release age with npm-check-updates (which is what I typically used to update the deps).

Issue #5316, opened 2026-04-23
C#: SSL transceiver adds ~64 µs per-call overhead vs TCP on ping benchmark


pepone:
## Summary

While updating the [icerpc-csharp-performance](https://github.com/icerpc/icerpc-csharp-performance) benchmarks to the Ice 3.8 and IceRPC nightly builds, I noticed the SSL path in Ice for C# adds an unexpectedly large **per-invocation** latency compared to plain TCP — around **~64 µs per call** on a simple `ping()` benchmark, on top of the (expected) one-time handshake cost.

The same cert, same loopback, and same small payload running through the IceRPC library over the ice protocol shows essentially **0 µs** of per-call TLS overhead, so this isn't TLS crypto cost — it looks like an adapter-layer issue in `src/Ice/SSL/TransceiverI.cs`. Opening this for the 3.8.2 release to investigate.

## Measurement

`Comparative.Benchmarks` PingLatency, BenchmarkDotNet, .NET 10, Apple M1 / macOS, loopback (`::1`), long-lived connection (handshake amortized in `GlobalSetup`).

| Library                       | Plain TCP | TLS       | Δ (TLS − TCP) |
|-------------------------------|----------:|----------:|--------------:|
| **Ice 3.8 nightly (C#)**      |  69.8 µs  | 133.9 µs  | **+64 µs**    |
| **IceRPC nightly (ice protocol, C#)** | 65.1 µs | 65.4 µs | +0.3 µs |
| **IceRPC nightly (icerpc protocol)**  | 80.3 µs | 81.0 µs | +0.7 µs |
| gRPC (reference)              |  90.3 µs  |           |               |

The Ice TLS column uses the SSL transport (`ssl -h 127.0.0.1 -p 4061`); the IceRPC row uses the same PKCS#12 cert and same `SslStream` under the hood. Both handshakes are confirmed (`TransportConnectionInformation.RemoteCertificate` is non-null for all TLS variants).

## Likely cause

The SSL transceiver at [`csharp/src/Ice/SSL/TransceiverI.cs`](https://github.com/zeroc-ice/ice/blob/main/csharp/src/Ice/SSL/TransceiverI.cs) adapts `SslStream`'s Task-based API to Ice's older `SocketAsyncEventArgs`-based connection core, and that adapter adds a few costs on every read and every write:

**1. Unconditional `ContinueWith(..., TaskScheduler.Default)` on every I/O**

```csharp
// TransceiverI.cs — startRead
[...]

PR #5329, opened 2026-05-13
Avoid thread-pool hop when SSL read/write completes synchronously

pepone:
Picks up the small, self-contained patch from https://github.com/zeroc-ice/ice/issues/5316#issuecomment-4433468645. The parent issue was closed as not-planned because the larger TLS regression lives elsewhere, but this slice is worth landing on its own:

- In `startRead` / `startWrite`, invoke the callback inline when `SslStream.ReadAsync` / `WriteAsync` completes synchronously, skipping the `ContinueWith(..., TaskScheduler.Default)` thread-pool hop.
- In `finishAuthenticate`, null `_writeResult` after `Wait()` so any future defensive check doesn't see a stale handshake task.

### DIFF
--- a/csharp/src/Ice/SSL/TransceiverI.cs
+++ b/csharp/src/Ice/SSL/TransceiverI.cs
@@ -111,6 +111,10 @@ internal sealed class TransceiverI : Ice.Internal.Transceiver
         try
         {
             _readResult = _sslStream.ReadAsync(buf.b.rawBytes(), buf.b.position(), buf.b.remaining());
+            if (_readResult.IsCompletedSuccessfully)
+            {
+                return true;
+            }
             _readResult.ContinueWith(
                 task => callback(state),
                 TaskScheduler.Default);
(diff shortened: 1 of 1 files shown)


csharp/src/Ice/SSL/TransceiverI.cs
```diff
@@ -111,6 +111,10 @@ public bool startRead(Ice.Internal.Buffer buf, Ice.Internal.AsyncCallback callba
         try
         {
             _readResult = _sslStream.ReadAsync(buf.b.rawBytes(), buf.b.position(), buf.b.remaining());
+            if (_readResult.IsCompletedSuccessfully)
+            {
+                return true;
```
pepone, line 116: The startX / finishX pair is mandatory regardless of whether startX reported sync or async completion — the ThreadPool dispatches finishX either way (sync via the inline completedCallback → queueReadyForIOHandler → startMessage → finishAsync,
  async via the ContinueWith callback running the same path on a thread-pool thread).

bernardnormier: While this is probably correct, it adds complexity without much of a speed-up, so I would not include this update

csharp/src/Ice/SSL/TransceiverI.cs
```diff
@@ -428,6 +436,10 @@ private void finishAuthenticate()
             {
                 throw ex.InnerException;
             }
+            finally
+            {
+                _writeResult = null;
```
bernardnormier, line 441: We could keep this _writeResult clearing.

PR #2360, opened 2024-06-27
Back pressure fix from 3.7

bentoi:
This PR merges the back pressure fix from 3.7 to main.

PR #5330, opened 2026-05-13
Skip Task.Run hop in ConnectionI.startAsync when not needed

pepone:
Follow-up to https://github.com/zeroc-ice/ice/issues/5316 and #5329. Captures the TLS-path latency win that issue closed without addressing.

## Background

`ConnectionI.startAsync` currently wraps every I/O dispatch in `Task.Run`, hopping the call onto a .NET ThreadPool thread. The wrap was added in #2360 (\"Back pressure fix from 3.7\") to defend against a real Windows-specific failure: async socket I/O is initiated on a thread and the OS implicitly cancels it with `SocketError.OperationAborted` if that thread later terminates. Ice ThreadPool workers can be reaped when idle past `Ice.ThreadPool.*.ThreadIdleTime`, so an Ice worker that dispatches I/O and then sits idle long enough can have its in-flight I/O killed.

## Observation

The defense is only required when **both** of these hold:

1. **Windows.** On Linux and macOS, .NET's async socket I/O uses `epoll` / `kqueue` with a dedicated `SocketAsyncEngine` thread that owns the readiness loop; the initiating thread's lifetime is irrelevant.
2. **The Ice thread pool can shrink.** Worker reaping requires `_threadIdleTime != Timeout.InfiniteTimeSpan` AND `_sizeMax > 1` (the shrink check is `_inUse < _threads.Count - 1`, which is unreachable when the floor of 1 worker is also the ceiling).

Both inputs are known at `ConnectionI` construction time. The default Ice thread pool (`Size = 1`, `SizeMax = Size = 1`) can never shrink, so default-config users on every platform have been paying the hop for nothing. Non-Windows users with multi-threaded pools pay it for nothing too.

## Change

- Expose `bool canShrink` on `ThreadPool` (`_threadIdleTime != Timeout.InfiniteTimeSpan && _sizeMax > 1`).
- In `ConnectionI`, cache `_threadHopRequired = _isWindows && _threadPool.canShrink` at construction.
- Refactor `startAsync` to hold its body in a local function, then dispatch via `Task.Run` only when `_threadHopRequired`; otherwise invoke inline. The local function is non-escaping on the inline path, so no closure allocation.

## Behavior matrix

| Platform | Pool config | Before | After |
|---|---|---|---|
| Windows | default (`SizeMax=1`) | `Task.Run` hop | inline |
| Windows | `SizeMax>1` | `Task.Run` hop | `Task.Run` hop (unchanged) |
| Linux/macOS | default | `Task.Run` hop | inline |
| Linux/macOS | `SizeMax>1` | `Task.Run` hop | inline |

## Test plan

- [ ] CI green on all platforms.
- [ ] `Ice/ami` test on Linux/macOS (default + multi-threaded pool).
- [ ] `Ice/ami` test on Windows with `Ice.ThreadPool.Client.SizeMax=4` (exercises the preserved `Task.Run` path).
- [ ] TLS `PingLatency` benchmark to confirm the ~22 µs recovery on default Linux/macOS.

### DIFF
--- a/csharp/src/Ice/ConnectionI.cs
+++ b/csharp/src/Ice/ConnectionI.cs
@@ -593,6 +593,19 @@ public sealed class ConnectionI : Internal.EventHandler, CancellationHandler, Co
             return false;
         }
 
+        if (_threadHopRequired)
+        {
+            // Run the I/O on a .NET ThreadPool thread so it survives the initiating Ice worker exiting.
+            // See ctor for when this is required.
+            Task.Run(doIO);
+        }
+        else
+        {
+            doIO();
+        }
+
+        return true;
+
         void doIO()
         {
             lock (_mutex)
@@ -653,19 +666,6 @@ public sealed class ConnectionI : Internal.EventHandler, CancellationHandler, Co
                 }
             }
         }
-
-        if (_threadHopRequired)
-        {
-            // Run the I/O on a .NET ThreadPool thread so it survives the initiating Ice worker exiting.
-            // See ctor for when this is required.
-            Task.Run(doIO);
-        }
-        else
-        {
-            doIO();
-        }
-
-        return true;
     }
 
     public override bool finishAsync(int operation)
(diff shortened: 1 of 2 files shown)


csharp/src/Ice/ConnectionI.cs
```diff
@@ -593,10 +593,7 @@ public override bool startAsync(int operation, Ice.Internal.AsyncCallback comple
             return false;
         }
 
-        // Run the IO operation on a .NET thread pool thread to ensure the IO operation won't be interrupted if the
-        // Ice thread pool thread is terminated (.NET Socket read/write fail with a SocketError.OperationAborted
-        // error if started from a thread which is later terminated).
-        Task.Run(() =>
+        void doIO()
```
bernardnormier, line 609: While the current style is easier to review, I'd rather put this local function at the end of the method.

bernardnormier (approved):

PR #5334, opened 2026-05-20
Fix integer overflow in readAndCheckSeqSize sequence-size guard

pepone:
readAndCheckSeqSize computed 'size * minWireSize' in 32-bit, so a large peer-supplied sequence size overflowed and bypassed the bounds check that guards sequence unmarshaling, leading to excessive allocation (and, in C++, a potential out-of-bounds read).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ These are the changes since the [Ice 3.8.1] release.
   is now performed using 64-bit integers. This affects the C++, C#, Java, and Swift mappings, as well
   as the Python, Ruby, and PHP mappings, which build on the C++ core.
 
+- Fixed `InputStream.readSize` in the Swift mapping to reject a negative size, consistent with the
+  other mappings. A peer could otherwise send a negative size and crash the receiver.
+
 ### Slice Language Changes
 
 - Added the `["oneway"]` metadata directive for Slice operations. This directive can only be applied to operations that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,14 +23,9 @@ These are the changes since the [Ice 3.8.1] release.
 
 - Fixed an unbounded memory allocation when unmarshaling a proxy with a large endpoint count.
 
-- Fixed an integer overflow in the sequence-size validation performed while unmarshaling. A peer
-  could send a message with an oversized sequence size whose `size * minWireSize` product overflowed
-  a 32-bit integer, defeating the bounds check that guards sequence unmarshaling. The size arithmetic
-  is now performed using 64-bit integers. This affects the C++, C#, Java, and Swift mappings, as well
-  as the Python, Ruby, and PHP mappings, which build on the C++ core.
+- Fixed an integer overflow in the sequence-size validation performed while unmarshaling.
 
-- Fixed `InputStream.readSize` in the Swift mapping to reject a negative size, consistent with the
-  other mappings. A peer could otherwise send a negative size and crash the receiver.
+- Fixed `InputStream.readSize` in the Swift mapping to reject a negative size.
 
 ### Slice Language Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@ These are the changes since the [Ice 3.8.1] release.
 
 - Fixed an unbounded memory allocation when unmarshaling a proxy with a large endpoint count.
 
+- Fixed an integer overflow in the sequence-size validation performed while unmarshaling. A peer
+  could send a message with an oversized sequence size whose `size * minWireSize` product overflowed
+  a 32-bit integer, defeating the bounds check that guards sequence unmarshaling. The size arithmetic
+  is now performed using 64-bit integers. This affects the C++, C#, Java, and Swift mappings, as well
+  as the Python, Ruby, and PHP mappings, which build on the C++ core.
+
 ### Slice Language Changes
 
 - Added the `["oneway"]` metadata directive for Slice operations. This directive can only be applied to operations that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ might need to be aware of.
   - [Slice Language Changes](#slice-language-changes)
   - [C++ Changes](#c-changes)
   - [C# Changes](#c-changes-1)
+  - [Swift Changes](#swift-changes)
 
 ## Changes in Ice 3.9.0
 
@@ -24,8 +25,6 @@ These are the changes since the [Ice 3.8.1] release.
 - Fixed an unbounded memory allocation when unmarshaling a proxy with a large endpoint count.
 
 - Fixed an integer overflow in the sequence-size validation performed while unmarshaling.
-
-- Fixed `InputStream.readSize` in the Swift mapping to reject a negative size.
 
 ### Slice Language Changes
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,15 @@ These are the changes since the [Ice 3.8.1] release.
 
 - Removed the 'ice2slice' compiler
 
+- Fixed an integer overflow in the sequence-size validation performed while unmarshaling. A peer
```
bernardnormier: I find this entry too detailed. We should stop after "while unmarshaling.".

CHANGELOG.md
```diff
@@ -19,6 +19,15 @@ These are the changes since the [Ice 3.8.1] release.
...
+  could send a message with an oversized sequence size whose `size * minWireSize` product overflowed
+  a 32-bit integer, defeating the bounds check that guards sequence unmarshaling. The size arithmetic
+  is now performed using 64-bit integers. This affects the C++, C#, Java, and Swift mappings, as well
+  as the Python, Ruby, and PHP mappings, which build on the C++ core.
+
+- Fixed `InputStream.readSize` in the Swift mapping to reject a negative size, consistent with the
```
bernardnormier: All the entries should be formatted using 120 columns like the rest of this file.

cpp/include/Ice/InputStream.h
```diff
@@ -929,8 +929,10 @@ namespace Ice
 
         void* _closure{nullptr};
 
-        int _startSeq{-1};
-        int _minSeqSize{0};
+        // The sequence-size bookkeeping is kept in 64-bit so that 'size * minWireSize' arithmetic
+        // cannot overflow and defeat the bounds check in readAndCheckSeqSize.
+        std::int64_t _startSeq{-1};
```
bernardnormier: This does not appear binary compatible. Also, I am pretty sure we don't need int64 here.

bernardnormier (changes requested):

cpp/src/Ice/InputStream.cpp
```diff
@@ -428,22 +428,29 @@ Ice::InputStream::readAndCheckSeqSize(int minSize)
...
     // possibly enclosed sequences), something is wrong with the marshaled
     // data: it's claiming having more data that what is possible to read.
     //
-    if (_startSeq + _minSeqSize > static_cast<int>(b.size()))
+    // We also reject any sequence whose minimum size exceeds INT32_MAX. The Ice
+    // protocol encodes a message size as a 32-bit integer, so no sequence can
+    // legitimately require more; this also keeps 'sz * minSize' within a 32-bit
+    // int for the callers that use the returned size to advance the stream.
+    //
+    if (_startSeq + _minSeqSize > static_cast<int64_t>(b.size()) || _minSeqSize > INT32_MAX)
```
bernardnormier: This update ignores basic guarantees here:
- `_startSeq` is less than INT32_MAX
- `_startSeq + _minSeqSize` is less than INT32_MAX
- `b.size()` is less than INT32_MAX

swift/src/Ice/InputStream.swift
```diff
@@ -552,7 +554,12 @@ extension InputStream {
     public func readSize() throws -> Int32 {
         let byteVal: UInt8 = try read()
         if byteVal == 255 {
-            return try read()
+            let v: Int32 = try read()
+            // A size is a non-negative value; reject a negative one encoded in the 5-byte form.
+            if v < 0 {
+                throw MarshalException(endOfBufferMessage)
```
bernardnormier: The error message is not correct.

cpp/src/Ice/InputStream.cpp
```diff
@@ -428,26 +428,30 @@ Ice::InputStream::readAndCheckSeqSize(int minSize)
     // the estimated remaining buffer size. This estimation is based on
     // the minimum size of the enclosing sequences, it's _minSeqSize.
     //
+    // 'sz' is peer-controlled (up to INT32_MAX), so we compute the minimum size of this sequence in
+    // 64-bit: 'sz * minSize' would otherwise overflow a 32-bit int and bypass the bounds check below.
+    int64_t minSeqSize = static_cast<int64_t>(sz) * minSize;
```
bernardnormier, line 433: How do we know `b.begin() + _startSeq + _minSeqSize` below does not overflow, in particular on a 32-bit system?

I think that's the case but it deserves a comment. It could make sense to switch to `(i - b.begin())` like you did previously.

bernardnormier: Also, we know `_startSeq + _minSeqSize` doesn't overflow thanks to the check below, from the previous call.

swift/src/Ice/InputStream.swift
```diff
@@ -16,8 +16,10 @@ public final class InputStream {
 
     private var encaps: Encaps!
 
-    private var startSeq: Int32 = -1
-    private var minSeqSize: Int32 = 0
+    // Uses Int rather than Int32 so the 'size * minWireSize' arithmetic in readAndCheckSeqSize
```
bernardnormier: We should keep In32 and use the same logic as in the other languages.

cpp/src/Ice/InputStream.cpp
pepone: Reworked

swift/src/Ice/InputStream.swift
```diff
@@ -587,22 +592,27 @@ extension InputStream {
         // the estimated remaining buffer size. This estimation is based on
         // the minimum size of the enclosing sequences, it's minSeqSize.
         //
-        if startSeq == -1 || pos > (startSeq + minSeqSize) {
+        // 'sz' is peer-controlled (up to Int32.max), so we compute the minimum size of this
+        // sequence in 64-bit (Int): 'sz * minSize' would overflow a 32-bit value and bypass the
+        // bounds check.
+        var newMinSeqSize = sz * minSize
+        if startSeq == -1 || pos > Int(startSeq) + Int(minSeqSize) {
```
bernardnormier: We should not cast startSeq/minSeqSize to Int: we know their addition doesn't overflow Int32.

swift/src/Ice/InputStream.swift
```diff
@@ -587,22 +592,27 @@ extension InputStream {
         // the estimated remaining buffer size. This estimation is based on
         // the minimum size of the enclosing sequences, it's minSeqSize.
         //
-        if startSeq == -1 || pos > (startSeq + minSeqSize) {
+        // 'sz' is peer-controlled (up to Int32.max), so we compute the minimum size of this
+        // sequence in 64-bit (Int): 'sz * minSize' would overflow a 32-bit value and bypass the
```
bernardnormier: We should specify the type (Int64, not Int) for clarity here.

swift/src/Ice/InputStream.swift
```diff
@@ -587,22 +592,27 @@ extension InputStream {
...
+        // bounds check.
+        var newMinSeqSize = sz * minSize
+        if startSeq == -1 || pos > Int(startSeq) + Int(minSeqSize) {
             startSeq = Int32(pos)
-            minSeqSize = Int32(sz * minSize)
         } else {
-            minSeqSize += Int32(sz * minSize)
+            newMinSeqSize += Int(minSeqSize)
```
bernardnormier: Same here, cast to Int64, not Int. This is all about avoiding overflows, better be specific.

Same for the next cast.

bernardnormier (approved):

swift/src/Ice/InputStream.swift
externl: in 64-bit (Int):

this is only true on a 64-bit os (which is all that we support). Can we use Int64 for more clarity here?

externl (approved):

PR #5335, opened 2026-05-20
Enforce RFC 6455 control-frame limits in the WebSocket transport

pepone:
A WebSocket control frame (close, ping, pong) was length-decoded like a data frame, so a peer could send a tiny control frame declaring a payload of up to 2 GB. The receiver buffered the full declared payload before applying any size limit, allowing a remote, pre-authentication denial of service through memory exhaustion.

The transceivers now reject, before allocating any payload buffer, a control frame whose payload exceeds 125 bytes or that is fragmented (FIN bit clear), per RFC 6455 section 5.5. They also reject any frame with a non-zero RSV bit, since Ice negotiates no extension. Fixed in the C++, C#, and Java transceivers; the C++ fix also covers the Python, Ruby, PHP, MATLAB, and Swift mappings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,12 +19,10 @@ These are the changes since the [Ice 3.8.1] release.
 
 - Removed the 'ice2slice' compiler
 
-- Fixed the WebSocket transport to enforce RFC 6455 section 5.5 on control frames. A peer could
-  previously send a `ping`, `pong`, or `close` control frame declaring a payload of up to 2 GB,
-  which the receiver buffered before applying any size limit, leading to excessive memory
-  allocation. Control frames whose payload exceeds 125 bytes or that are fragmented are now
-  rejected with a `ProtocolException`. This affects the C++, C#, and Java mappings, as well as the
-  Python, Ruby, PHP, MATLAB, and Swift mappings, which build on the C++ core.
+- Fixed the WebSocket transport to enforce the RFC 6455 limits on control frames. A peer could previously send a
+  tiny `ping`, `pong`, or `close` frame declaring a multi-gigabyte payload that the receiver buffered before applying
+  any size limit. Control frames larger than 125 bytes or that are fragmented are now rejected with a
+  `ProtocolException`.
 
 ### Slice Language Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,10 +19,7 @@ These are the changes since the [Ice 3.8.1] release.
 
 - Removed the 'ice2slice' compiler
 
-- Fixed the WebSocket transport to enforce the RFC 6455 limits on control frames. A peer could previously send a
-  tiny `ping`, `pong`, or `close` frame declaring a multi-gigabyte payload that the receiver buffered before applying
-  any size limit. Control frames larger than 125 bytes or that are fragmented are now rejected with a
-  `ProtocolException`.
+- Fixed the WebSocket transport to enforce the RFC 6455 limits on control frames.
 
 ### Slice Language Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,13 @@ These are the changes since the [Ice 3.8.1] release.
 
 - Removed the 'ice2slice' compiler
 
+- Fixed the WebSocket transport to enforce RFC 6455 section 5.5 on control frames. A peer could
+  previously send a `ping`, `pong`, or `close` control frame declaring a payload of up to 2 GB,
+  which the receiver buffered before applying any size limit, leading to excessive memory
+  allocation. Control frames whose payload exceeds 125 bytes or that are fragmented are now
+  rejected with a `ProtocolException`. This affects the C++, C#, and Java mappings, as well as the
+  Python, Ruby, PHP, MATLAB, and Swift mappings, which build on the C++ core.
+
 ### Slice Language Changes
 
 - Added the `["oneway"]` metadata directive for Slice operations. This directive can only be applied to operations that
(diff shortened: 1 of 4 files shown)


cpp/src/Ice/WSTransceiver.cpp
```diff
@@ -1162,6 +1172,30 @@ IceInternal::WSTransceiver::preRead(Buffer& buf)
             // 127:   The subsequent eight bytes contain the payload length
             //
             _readPayloadLength = static_cast<size_t>((ch & byte{0x7f}));
+
+            //
+            // RFC 6455 section 5.5: control frames (close, ping and pong) must not be fragmented
```
InsertCreativityHere:
```suggestion
            // RFC 6455 section 5.5: control frames (close, ping, and pong) must not be fragmented
```
This isn't an exact quote of the spec (like the colon implies) so adding this comma is fine.

cpp/src/Ice/WSTransceiver.cpp
```diff
@@ -1120,6 +1120,16 @@ IceInternal::WSTransceiver::preRead(Buffer& buf)
             byte ch = *_readI++;
             _readOpCode = static_cast<int>(ch & byte{0xf});
 
+            //
+            // No extension is negotiated, so the RSV1, RSV2 and RSV3 bits must all be 0.
```
InsertCreativityHere: Ever the Oxford comma pedant:
```suggestion
            // No extension is negotiated, so the RSV1, RSV2, and RSV3 bits must all be 0.
```

cpp/src/Ice/WSTransceiver.cpp
```diff
@@ -1162,6 +1172,30 @@ IceInternal::WSTransceiver::preRead(Buffer& buf)
             // 127:   The subsequent eight bytes contain the payload length
             //
             _readPayloadLength = static_cast<size_t>((ch & byte{0x7f}));
+
+            //
+            // RFC 6455 section 5.5: control frames (close, ping and pong) must not be fragmented
+            // and must have a payload length of 125 bytes or less - they cannot use the 16-bit or
+            // 64-bit extended length encoding. Enforce this before allocating any payload buffer.
```
InsertCreativityHere, line 1179: Should we also mention the requirement that the `FIN` bit must be set? Since we check both `FIN` and the payload size, but we only mention the payload size in this comment.

Not critical though, maybe quickly mentioning non-fragmented is good enough.

csharp/src/Ice/Internal/WSTransceiver.cs
```diff
@@ -963,6 +963,16 @@ private bool preRead(Buffer buf)
...
 
+                //
+                // No extension is negotiated, so the RSV1, RSV2 and RSV3 bits must all be 0.
+                //
+                if ((ch & 0x70) != 0)
+                {
+                    throw new Ice.ProtocolException("invalid WebSocket frame: RSV bits must be 0");
+                }
+
+                bool finalFrame = (ch & FLAG_FINAL) == FLAG_FINAL;
```
InsertCreativityHere, line 974: **EDIT**: nevermind, there doesn't seem to be a way to do this in C#. Ignore my comment.

----

We mark this `const` in C++, and `final` in Java, might as well be consistent here
```
                readonly bool finalFrame = (ch & FLAG_FINAL) == FLAG_FINAL;
```
I know that we aren't very good at using `readonly` for C# variables though...

csharp/src/Ice/Internal/WSTransceiver.cs
```diff
@@ -963,6 +963,16 @@ private bool preRead(Buffer buf)
                 int ch = _readBuffer.b.get(_readBufferPos++);
                 _readOpCode = ch & 0xf;
 
+                //
+                // No extension is negotiated, so the RSV1, RSV2 and RSV3 bits must all be 0.
```
InsertCreativityHere:
```suggestion
                // No extension is negotiated, so the RSV1, RSV2, and RSV3 bits must all be 0.
```

csharp/src/Ice/Internal/WSTransceiver.cs
```diff
@@ -1005,6 +1015,26 @@ private bool preRead(Buffer buf)
                 // 127:   The subsequent eight bytes contain the payload length
                 //
                 _readPayloadLength = ch & 0x7f;
+
+                //
+                // RFC 6455 section 5.5: control frames (close, ping and pong) must not be fragmented
```
InsertCreativityHere:
```suggestion
                // RFC 6455 section 5.5: control frames (close, ping, and pong) must not be fragmented
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java
```diff
@@ -627,18 +627,25 @@ private boolean preRead(Buffer buf) {
                 }
                 _readOpCode = ch & 0xf;
 
+                // No extension is negotiated, so the RSV1, RSV2 and RSV3 bits must all be 0.
```
InsertCreativityHere:
```suggestion
                // No extension is negotiated, so the RSV1, RSV2, and RSV3 bits must all be 0.
```

InsertCreativityHere (approved):
Looks good to me!
JavaScript already handles this correctly I assume?

CHANGELOG.md
```diff
@@ -19,6 +19,13 @@ These are the changes since the [Ice 3.8.1] release.
 
 - Removed the 'ice2slice' compiler
 
+- Fixed the WebSocket transport to enforce RFC 6455 section 5.5 on control frames. A peer could
```
bernardnormier: Format on 120 columns and clean-up this entry.

bernardnormier (approved):

pepone:
Yes. The JavaScript mapping uses the platform `WebSocket` (browser-native or Node's built-in), which does the framing itself and enforces RFC 6455 §5.5. `WSTransceiver.js` only receives deframed payloads via `onmessage` — it never parses opcodes or control frames, so there's nothing on the Ice side to harden.

PR #5336, opened 2026-05-20
debian: install and use sysusers.d/tmpfiles.d config files

bluca:
sysusers.d/tmpfiles.d config files allow a package to use declarative configuration instead of manually written maintainer scripts. This also allows image-based systems to be created with /usr/ only, and also allows for factory resetting a system and recreating /etc/ on boot.

https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html
https://www.freedesktop.org/software/systemd/man/latest/tmpfiles.d.html


packaging/deb/debian/control
```diff
@@ -8,6 +8,7 @@ Build-Depends: debhelper,
                dh-exec,
                dh-php (>= 0.20),
                dh-python,
+               dh-sequence-installsysusers,
```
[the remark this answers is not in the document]
bluca: There's no such sequence, it's implicit in compat 13

pepone:
Hi @bluca

Thanks for this PR, I want to test our packaging workflow to check everything works. One question do you think is safe to back-port this to the 3.8 branch later, which is the branch we will use to publish the 3.8.2 Debian packages, or better keep this for the next major release 3.9 packaging?

packaging/deb/debian/zeroc-icegrid.tmpfiles
```diff
@@ -0,0 +1,3 @@
+d /var/lib/ice/icegrid 0770 ice adm
+d /var/lib/ice/icegrid/node1 0770 ice adm
+d /var/lib/ice/icegrid/registry 0770 ice adm
```
pepone, line 3: are `libzeroc-ice3.9.dirs` and `zeroc-icegrid.dirs` still required?

bluca:
> Hi @bluca
> 
> Thanks for this PR, I want to test our packaging workflow to check everything works. One question do you think is safe to back-port this to the 3.8 branch later, which is the branch we will use to publish the 3.8.2 Debian packages, or better keep this for the next major release 3.9 packaging?

Hi, the content of the PR as-is is safe to upload not just to debian unstable/ubuntu dev, but also to trixie/bookworm and resolute/noble, so if you wish to backport it to other branches that should be ok, provided the user/dirs are the same

packaging/deb/debian/zeroc-icegrid.tmpfiles
bluca: They are not, good catch, dropped

packaging/deb/debian/libzeroc-ice3.9.sysusers
```diff
@@ -0,0 +1 @@
+u ice - "Ice Service account" /var/lib/ice /usr/sbin/nologin
```
externl, line 1: We were previously also creating an `ice` group. Is this still the case?

bluca: yes, the default is to create a matching group unless manually overridden

pepone (approved):

externl (approved):
Looks good to me!

PR #5339, opened 2026-05-21
Reject a slice header too small for its optional members

pepone:
A sliced-format slice that sets the optional-members flag carries at least a 1-byte end marker, so its size must be >= 5. The 1.1 decoder now enforces this in C++, C#, Java, JavaScript, MATLAB, and Swift.

### DIFF
--- a/cpp/src/Ice/InputStream.cpp
+++ b/cpp/src/Ice/InputStream.cpp
@@ -1930,7 +1930,10 @@ Ice::InputStream::EncapsDecoder11::startSlice()
     if (_current->sliceFlags & FLAG_HAS_SLICE_SIZE)
     {
         _stream->read(_current->sliceSize);
-        if (_current->sliceSize < 4)
+        // A slice with optional members carries at least the 1-byte end marker in its body, so its
+        // size (which includes the 4-byte size field) must be >= 5.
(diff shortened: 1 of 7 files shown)


cpp/src/Ice/InputStream.cpp
```diff
@@ -1923,7 +1923,10 @@ Ice::InputStream::EncapsDecoder11::startSlice()
     if (_current->sliceFlags & FLAG_HAS_SLICE_SIZE)
     {
         _stream->read(_current->sliceSize);
-        if (_current->sliceSize < 4)
+        // A slice with optional members carries at least the 1-byte end marker in its body, so its
+        // size (which includes the 4-byte size field) must be >= 5.
+        int32_t minSliceSize = (_current->sliceFlags & FLAG_HAS_OPTIONAL_MEMBERS) ? 5 : 4;
```
bernardnormier, line 1936: I don't understand this update.

Is it because later on (where?), we read this byte end marker without checking the position relative to `_current->sliceSize`?

bernardnormier (approved):

cpp/src/Ice/InputStream.cpp
bernardnormier: I would add a comment like:

> We rely on this check in the slice-preservation logic.

In the slice preservation code, we indeed rely on this specific check.

externl (approved):

PR #5340, opened 2026-05-22
Reject a batch request count larger than the message can contain

pepone:
Validate the batch request count against the remaining message data before accumulating it into the dispatch counters.


bernardnormier (approved):
It's indeed unclear why we set `info.requestCount` to `0` in Java (unrelated to this PR).

externl (approved):
Looks good!

PR #5344, opened 2026-05-22
Warn against DES passwords at startup

bernardnormier:
This PR updates the Glacier2CryptPermissionsVerifier plug-in to issue a warning when the configured password file contains old DES passwords.

DES passwords are not secure and should not be used in production.

### DIFF
--- a/cpp/src/Glacier2/CryptPermissionsVerifier.cpp
+++ b/cpp/src/Glacier2/CryptPermissionsVerifier.cpp
@@ -66,7 +66,7 @@ namespace
         CommunicatorPtr _communicator;
     };
 
-    map<string, string> retrievePasswordMap(const string& file, LoggerPtr logger)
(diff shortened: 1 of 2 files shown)


cpp/src/Glacier2/CryptPermissionsVerifier.cpp
```diff
@@ -94,8 +96,20 @@ namespace
...
+            hasDESStylePassword =
+                hasDESStylePassword || (password.find('$') == string::npos && password.size() == 13);
+
             passwords.insert(make_pair(userId, password));
         }
+
+        if (hasDESStylePassword)
+        {
+            Warning out(logger);
+            out << "The password file '" << file << "' contains one or more DES-style passwords. DES is a weak "
```
externl, line 108: Should we tell the user what alternative to use?

externl (approved):
Looks good!

pepone (approved):

PR #5350, opened 2026-05-25
Reject peer-initiated TLS renegotiation

pepone:
Disable peer-initiated TLS renegotiation in the OpenSSL and Schannel SSL transports:

- **OpenSSL**: set `SSL_OP_NO_RENEGOTIATION` on the `SSL_CTX`.
- **Schannel**: `SEC_I_RENEGOTIATE` now throws `ProtocolException` instead of entering the renegotiation state machine. The now-unreachable renegotiation-handling code (`_sslConnectionRenegotiating`, `_extraBuffer`, the two handshake re-entry paths) is removed in the same change.

The SecureTransport (macOS) transport doesn't expose renegotiation to Ice — `SSLRead` returns the request as an SSL error, which already throws `ProtocolException`. No change needed there.

Resolves zeroc-ice/ice-audit#20.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,9 @@ These are the changes since the [Ice 3.8.1] release.
   `IceSSL.CertFile` is imported into a temporary keychain (removed when the communicator is destroyed) instead of
   the user's login keychain.
 
+- Rejected peer-initiated TLS renegotiation in the OpenSSL SSL engine and on the server side of Schannel-based SSL
+  connections. This applies to all SSL-based transports (`ssl`, `wss`, `bts`).
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -41,6 +41,9 @@ These are the changes since the [Ice 3.8.1] release.
 
 - Changed the macOS SSL transport to require TLS 1.2 or later.
 
+- Rejected peer-initiated TLS renegotiation in the OpenSSL SSL transport and on the server side of the Schannel
```
bernardnormier: Is it in the SSL transport (meaning ssl over tcp), or in the SSL engine (meaning ssl over tcp, bt, iap, etc.)?

bernardnormier (approved):

externl (approved):

PR #5352, opened 2026-05-25
Reject an IP endpoint port outside the 0..65535 range

pepone:
## Summary

- C++, C#, Java, and JavaScript: validate the `port` field while unmarshaling an `IPEndpointI` and throw `MarshalException` if it falls outside 0..65535.
- Add a CHANGELOG entry under General Changes.

## Test plan

### DIFF
--- a/cpp/src/Ice/IPEndpointI.cpp
+++ b/cpp/src/Ice/IPEndpointI.cpp
@@ -419,6 +419,10 @@ IceInternal::IPEndpointI::IPEndpointI(ProtocolInstancePtr instance, InputStream*
     s->read(const_cast<string&>(_host), false);
     const_cast<string&>(_normalizedHost) = normalizeIPv6Address(_host);
     s->read(const_cast<int32_t&>(_port));
+    if (_port < 0 || _port > 65535)
+    {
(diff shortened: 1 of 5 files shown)


cpp/src/Ice/IPEndpointI.cpp
```diff
@@ -419,6 +419,10 @@ IceInternal::IPEndpointI::IPEndpointI(ProtocolInstancePtr instance, InputStream*
     s->read(const_cast<string&>(_host), false);
     const_cast<string&>(_normalizedHost) = normalizeIPv6Address(_host);
     s->read(const_cast<int32_t&>(_port));
+    if (_port < 0 || _port > 65535)
+    {
+        throw MarshalException{__FILE__, __LINE__, "port value '" + to_string(_port) + "' out of range in endpoint"};
```
bernardnormier: I would simplify the message to simply:
port value XXX is out of range

In the other message (for parsing), we actually include the endpoint in the mssage.

bernardnormier (approved):

externl (approved):
Looks good. I wonder if we should also reject `0`.

pepone:
> Looks good. I wonder if we should also reject 0.

I don't think so, 0 is valid for and endpoint port.

PR #5353, opened 2026-05-25
Reject a zero entry in a class indirection table

pepone:
## Summary

- In `EncapsDecoder11::endSlice` and `EncapsDecoder11::skipSlice` (C++, C#, Java), check the value returned by `readSize()` for each indirection-table entry and throw `MarshalException` if it is `<= 0`. Mirrors the validation that the public `read(...)` entry point already performs, so the private `readInstance` helper's `index > 0` precondition holds for every caller.
- Without this guard a wire-encoded zero indirection-table index reaches `assert(index > 0)` in `readInstance` — which aborts debug builds (remote DoS in debug mode) and silently synthesizes a new instance ID in release builds.

## Test plan

### DIFF
--- a/cpp/src/Ice/InputStream.cpp
+++ b/cpp/src/Ice/InputStream.cpp
@@ -1961,7 +1961,12 @@ Ice::InputStream::EncapsDecoder11::endSlice()
         IndexList indirectionTable(static_cast<size_t>(_stream->readAndCheckSeqSize(1)));
         for (auto& p : indirectionTable)
         {
-            p = readInstance(_stream->readSize(), nullptr, nullptr);
+            int32_t i = _stream->readSize();
+            if (i <= 0)
+            {
+                throw MarshalException(__FILE__, __LINE__, "invalid indirection-table index");
+            }
+            p = readInstance(i, nullptr, nullptr);
         }
 
         //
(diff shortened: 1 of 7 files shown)


cpp/src/Ice/InputStream.cpp
```diff
@@ -1961,7 +1961,12 @@ Ice::InputStream::EncapsDecoder11::endSlice()
         IndexList indirectionTable(static_cast<size_t>(_stream->readAndCheckSeqSize(1)));
         for (auto& p : indirectionTable)
         {
-            p = readInstance(_stream->readSize(), nullptr, nullptr);
+            int32_t index = _stream->readSize();
+            if (index <= 0)
```
bernardnormier: I would rather replace the assert in readInstance by this check.

I find the check here odd because readSize already guarantees that it return a size (>= 0). What's special with readInstance is value `0` is not valid.

csharp/src/Ice/InputStream.cs
```diff
@@ -2629,7 +2629,12 @@ internal override void endSlice()
                 int[] indirectionTable = new int[_stream.readAndCheckSeqSize(1)];
                 for (int i = 0; i < indirectionTable.Length; ++i)
                 {
-                    indirectionTable[i] = readInstance(_stream.readSize(), null);
+                    int index = _stream.readSize();
+                    if (index <= 0)
+                    {
+                        throw new MarshalException("invalid indirection-table index");
+                    }
+                    indirectionTable[i] = readInstance(index, null);
```
bernardnormier: Same here: I prefer the check in readInstance.

bernardnormier (approved):

externl (approved):

PR #5354, opened 2026-05-26
Add the AllowedOrigins object adapter property

pepone:
## Summary

Adds a new object adapter property `<AdapterName>.AllowedOrigins`. When set on an adapter with a WebSocket endpoint, the server validates the upgrade request's `Origin` header against a comma-separated allowlist and rejects requests from origins not in the list. Defaults to empty (no enforcement); `*` is also permissive.

- Implemented in C++, C#, and Java (JavaScript Ice servers are bidir-only and don't accept inbound WS handshakes).
- Browsers always send `Origin`; non-browser clients do not — so the check only filters browser-originated connections, preserving full backward compatibility for every existing non-browser Ice peer.
- The check runs once per connection at the WebSocket handshake; subsequent Ice traffic is binary WS frames with no HTTP layer.
- Property registered in `config/PropertyNames.xml`; per-language tables regenerated via `makeprops.py`.
- Origin canonicalization is lowercase scheme/host with default ports (80/443) omitted, per RFC 6454.

Mitigates Cross-Site WebSocket Hijacking against Ice WS deployments that rely on network topology for trust, or `wss://` deployments using only TLS client certificates (where the browser silently attaches the victim's cert to a cross-origin handshake). See zeroc-ice/ice-audit#19 for the threat model and design proposal.

The default-from-published-endpoints behavior is planned for 3.9.0 as a separate change.

## Test plan


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java
```diff
@@ -1246,4 +1263,51 @@ private void prepareWriteHeader(byte opCode, int payloadLength) {
     private static final String _wsUUID = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
 
     static final Charset _ascii = Charset.forName("US-ASCII");
+
+    // Parse the value of the ObjectAdapter property "AllowedOrigins" into a canonicalized set of origins.
+    // Each entry is "scheme://host[:port]", lowercased, with the default port for the scheme (80/443) omitted.
+    // The literal "*" passes through unchanged and signals "allow any origin".
+    static Set<String> parseAllowedOrigins(String value) {
```
externl: Why can't we use `getPropertyAsList`?

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java
```diff
@@ -1246,4 +1263,51 @@ private void prepareWriteHeader(byte opCode, int payloadLength) {
...
+        }
+        return result;
+    }
+
+    private static String canonicalizeOrigin(String origin) {
+        if (origin.equals("*")) {
+            return origin;
+        }
+        int sep = origin.indexOf("://");
+        if (sep == -1) {
```
externl: I think we should throw an exception if the user has provided a malformed origin.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java
```diff
@@ -1246,4 +1263,51 @@ private void prepareWriteHeader(byte opCode, int payloadLength) {
...
+            return ""; // malformed
+        }
+        String scheme = origin.substring(0, sep).toLowerCase(java.util.Locale.ROOT);
+        String rest = origin.substring(sep + 3);
+        int pathStart = rest.indexOf('/');
+        if (pathStart >= 0) {
```
externl: Ditto, this is not a valid origin. I think we should fail here with an exception.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java
```diff
@@ -1246,4 +1263,51 @@ private void prepareWriteHeader(byte opCode, int payloadLength) {
...
+    private static String canonicalizeOrigin(String origin) {
+        if (origin.equals("*")) {
+            return origin;
+        }
+        int sep = origin.indexOf("://");
+        if (sep == -1) {
+            return ""; // malformed
+        }
+        String scheme = origin.substring(0, sep).toLowerCase(java.util.Locale.ROOT);
+        String rest = origin.substring(sep + 3);
```
externl: Should we use `authority` instead of `rest`?

externl (commented):
In C# and Java can we use the building URI parsing?

csharp/src/Ice/Internal/WSTransceiver.cs
```diff
@@ -801,6 +802,32 @@ private void handleRequest(Buffer responseBuffer)
             throw new WebSocketException("invalid value `" + key + "' for WebSocket key");
         }
 
+        //
+        // Optionally validate the Origin header against the adapter's allowed-origins list.
```
bernardnormier, line 806: I assume we'll document this limitation (non-browser clients don't send Origin; this check always succeeds for such clients).

bernardnormier (approved):

cpp/src/Ice/WSTransceiver.cpp
```diff
@@ -51,6 +51,45 @@ namespace
...
+            [](unsigned char c) { return std::tolower(c); });
+        if (auto colon = hostPort.rfind(':'); colon != string::npos)
+        {
+            string_view port = string_view{hostPort}.substr(colon + 1);
+            if ((scheme == "http" && port == "80") || (scheme == "https" && port == "443"))
+            {
+                hostPort = hostPort.substr(0, colon);
+            }
+        }
+        return scheme + "://" + hostPort;
```
externl: `hostPort` means host and port?

cpp/src/Ice/WSTransceiver.cpp
```diff
@@ -961,6 +1015,31 @@ IceInternal::WSTransceiver::handleRequest(Buffer& responseBuffer)
         throw WebSocketException("invalid value '" + key + "' for WebSocket key");
     }
 
+    //
+    // Optionally validate the Origin header against the adapter's allowed-origins list.
+    // Browsers always send Origin; non-browser clients do not, so an absent header bypasses the check.
+    //
+    if (!_allowedOrigins.empty() && _allowedOrigins.count("*") == 0)
```
externl: Maybe we can just compute `!_allowedOrigins.empty() && _allowedOrigins.count("*") == 0` up front instead of having to do `_allowedOrigins.count("*")` for every connection.

cpp/src/Ice/WSTransceiver.cpp
```diff
@@ -51,6 +51,45 @@ namespace
...
+        if (sep == string_view::npos || sep == 0)
+        {
+            throw PropertyException(__FILE__, __LINE__, "malformed origin '" + string{origin} + "'");
+        }
+        string scheme{origin.substr(0, sep)};
+        transform(scheme.begin(), scheme.end(), scheme.begin(), [](unsigned char c) { return std::tolower(c); });
+        string_view authority = origin.substr(sep + 3);
+        if (authority.empty() || authority.find_first_of("/?#@") != string_view::npos)
+        {
+            throw PropertyException(__FILE__, __LINE__, "malformed origin '" + string{origin} + "'");
```
externl: PropertyException seems like the wrong thing to throw since we use this to verify browser Origin.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java
pepone: Fixed to use `getPropertyAsList`

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java
pepone: Fixed as suggested.

cpp/src/Ice/WSTransceiver.h
```diff
@@ -10,16 +10,26 @@
...
+    // The literal "*" passes through unchanged and signals "allow any origin".
+    // Throws PropertyException if any entry is not a syntactically valid origin.
+    std::set<std::string> parseAllowedOrigins(const std::vector<std::string>& entries);
+
     class WSTransceiver final : public Transceiver
     {
     public:
         WSTransceiver(ProtocolInstancePtr, TransceiverPtr, std::string, std::string);
-        WSTransceiver(ProtocolInstancePtr, TransceiverPtr);
+        WSTransceiver(ProtocolInstancePtr, TransceiverPtr, std::set<std::string> allowedOrigins);
```
externl, line 32: Why do we need two constructors? Why would we ever want the other one that doesn't include allowedOrigins.

csharp/src/Ice/Internal/WSTransceiver.cs
```diff
@@ -1710,5 +1738,46 @@ private void prepareWriteHeader(byte opCode, int payloadLength)
...
+        if (origin == "*")
+        {
+            return origin;
+        }
+        // RFC 6454: a serialized origin is exactly "scheme://host[:port]", with no path, query, fragment, or
+        // userinfo. We reject these before letting System.Uri canonicalize the rest.
+        int sep = origin.IndexOf("://", StringComparison.Ordinal);
+        if (sep <= 0 || origin.IndexOfAny(_invalidOriginChars, sep + 3) >= 0)
+        {
+            throw new PropertyException($"malformed origin '{origin}'");
```
externl: Ditto for PropertyException.

csharp/src/Ice/Internal/WSTransceiver.cs
```diff
@@ -1710,5 +1738,46 @@ private void prepareWriteHeader(byte opCode, int payloadLength)
...
+
+    internal static string canonicalizeOrigin(string origin)
+    {
+        if (origin == "*")
+        {
+            return origin;
+        }
+        // RFC 6454: a serialized origin is exactly "scheme://host[:port]", with no path, query, fragment, or
+        // userinfo. We reject these before letting System.Uri canonicalize the rest.
+        int sep = origin.IndexOf("://", StringComparison.Ordinal);
```
externl: Is this part necessary. We can just parse the URI like we do below then check for a scheme. No need to do our own check.

csharp/src/Ice/Internal/WSTransceiver.cs
```diff
@@ -1710,5 +1738,46 @@ private void prepareWriteHeader(byte opCode, int payloadLength)
...
+        if (sep <= 0 || origin.IndexOfAny(_invalidOriginChars, sep + 3) >= 0)
```
externl: Name for this we can just check that path and params are empty.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java
```diff
@@ -1246,4 +1270,59 @@ private void prepareWriteHeader(byte opCode, int payloadLength) {
...
+            || containsAny(origin, sep + 3, "/?#@")) {
+            throw new PropertyException("malformed origin '" + origin + "'");
+        }
+        URI uri;
+        try {
+            uri = new URI(origin);
+        } catch (URISyntaxException ex) {
+            throw new PropertyException("malformed origin '" + origin + "'");
+        }
+        String scheme = uri.getScheme();
```
externl, line 1282: Ditto for Java.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java
```diff
@@ -1246,4 +1270,59 @@ private void prepareWriteHeader(byte opCode, int payloadLength) {
...
+        }
+        return result;
+    }
+
+    static String canonicalizeOrigin(String origin) {
+        if (origin.equals("*")) {
+            return origin;
+        }
+        // RFC 6454: a serialized origin is exactly "scheme://host[:port]", with no path, query, fragment, or
+        // userinfo. We reject these before letting java.net.URI canonicalize the rest.
```
externl: I think we should let java parse it, then reject it.

cpp/src/Ice/WSTransceiver.h
pepone: This only applies for incoming connections the allowed origins correspond to the OA configured origins.

cpp/src/Ice/WSTransceiver.cpp
pepone: Yes, updated to throw the appropriate argument exception

csharp/src/Ice/Internal/WSTransceiver.cs
pepone: Ok, reworked to use the parsed URI

csharp/src/Ice/Internal/WSTransceiver.cs
pepone: Reworked to use the parsed URI

cpp/src/Ice/WSTransceiver.cpp
pepone: Reworked now `*` is equivalent to not defined property, empty set which disables the check.

cpp/src/Ice/WSTransceiver.cpp
```diff
@@ -124,6 +125,43 @@ namespace
     }
 }
 
+string
+IceInternal::canonicalizeOrigin(string_view origin)
+{
+    // Throws std::invalid_argument for any input that is not a serialized origin per RFC 6454. Callers catch and
+    // rethrow as the appropriate exception (PropertyException for property entries, WebSocketException for inbound
+    // Origin headers).
```
externl:
```suggestion
    // Throws std::invalid_argument for any input that is not a serialized origin per RFC 6454.
```

csharp/src/Ice/Internal/WSTransceiver.cs
```diff
@@ -1711,4 +1740,30 @@ private void prepareWriteHeader(byte opCode, int payloadLength)
     private const string _wsUUID = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
 
     private static readonly UTF8Encoding _utf8 = new UTF8Encoding(false, true);
+
+    // Throws ArgumentException for any input that is not a serialized origin per RFC 6454. Callers catch and rethrow
+    // as the appropriate exception (PropertyException for property entries, WebSocketException for inbound Origin
+    // headers).
```
externl:
```suggestion
    // Throws ArgumentException for any input that is not a serialized origin per RFC 6454.
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java
```diff
@@ -492,6 +497,25 @@ private void handleRequest(Buffer responseBuffer) {
...
+            if (origin != null) {
+                String canonical;
+                try {
+                    canonical = canonicalizeOrigin(origin.trim());
+                } catch (IllegalArgumentException ex) {
+                    throw new WebSocketException("invalid Origin header '" + origin + "'");
+                }
+                if (!_allowedOrigins.contains(canonical)) {
+                    throw new WebSocketException(
+                        "origin '" + origin + "' is not in the adapter's AllowedOrigins list");
```
externl: Since this is really a message for the client. Maybe we should say something like:


```suggestion
                        "origin '" + origin + "' is not allowed`");
```

externl (approved):
Looks good!

PR #5355, opened 2026-05-26
Remove SSLEngine::destroy and rely on the destructor for cleanup

pepone:
## Summary

`Ice::SSL::SSLEngine::destroy()` is never called: `IceInternal::Instance` creates the engine, calls `initialize()`, and only releases the `shared_ptr` in `Instance::destroy()` (no explicit `_sslEngine->destroy()`). As a result, the cleanup in each subclass `destroy()` body never ran on shutdown, and never ran when `initialize()` itself threw.

- Move the OpenSSL `SSL_CTX_free(_ctx)` cleanup from `destroy()` into `~SSLEngine()`.
- Move the Schannel cleanup (chain engine, root store, imported cert keysets, cert contexts, stores) from `destroy()` into `~SSLEngine()`. Schannel previously had no destructor declared; one is now added explicitly.
- Drop `virtual void destroy() = 0;` from the base `Ice::SSL::SSLEngine`.
- Drop the now-empty `destroy()` override on `SecureTransport::SSLEngine`.

No call-site changes — nothing in the tree calls `destroy()` on an `SSLEngine`.

## Test plan

- [x] Local cpp build passes on macOS
- [ ] CI green on Windows (Schannel) and Linux (OpenSSL)

### DIFF
--- a/cpp/src/Ice/SSL/OpenSSLEngine.h
+++ b/cpp/src/Ice/SSL/OpenSSLEngine.h
@@ -17,10 +17,9 @@ namespace Ice::SSL::OpenSSL
     {
     public:
         SSLEngine(const IceInternal::InstancePtr&);
-        ~SSLEngine();
+        ~SSLEngine() override;
 
         void initialize() final;
-        void destroy() final;
         [[nodiscard]] std::string sslErrors() const;
         [[nodiscard]] std::string password() const { return _password; }
         [[nodiscard]] Ice::SSL::ClientAuthenticationOptions
--- a/cpp/src/Ice/SSL/SSLEngine.h
+++ b/cpp/src/Ice/SSL/SSLEngine.h
@@ -23,7 +23,7 @@ namespace Ice::SSL
     {
     public:
         SSLEngine(const IceInternal::InstancePtr&);
-        ~SSLEngine();
+        virtual ~SSLEngine();
 
         [[nodiscard]] Ice::LoggerPtr getLogger() const;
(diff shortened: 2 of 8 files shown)


cpp/src/Ice/SSL/SSLEngine.h
```diff
@@ -34,9 +34,6 @@ namespace Ice::SSL
         // Setup the engine.
         virtual void initialize() = 0;
```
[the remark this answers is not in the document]
bernardnormier: We only ever construct the leaf classes with make_shared, so that's not a concern.

Nevertheless, in C++, the destructor of a base class such as SSLEngine should be virtual. That's the recommended pattern.

cpp/src/Ice/SSL/OpenSSLEngine.h
```diff
@@ -20,7 +20,6 @@ namespace Ice::SSL::OpenSSL
         ~SSLEngine();
 
         void initialize() final;
```
externl, line 22: Would be nice to also get rid of initialize at some point as well. Not sure this temporal initialization is necessary.

externl (approved):
Looks good.

bernardnormier (approved):

cpp/src/Ice/SSL/OpenSSLEngine.h
pepone: I would check, I think this was original required when this was a plugin but it is probable it is no longer required.

PR #5358, opened 2026-05-26
Dispose Ice.SSL.SSLEngine from Internal.Instance.destroy()

pepone:
## Summary

Mirrors the C++ fix in #5355 for the .NET mapping. Before this change, the `X509Certificate2` instances stored in `Ice.SSL.SSLEngine._certs` and `_caCerts` were never disposed — the unmanaged certificate key handles (CryptoAPI / OpenSSL, depending on platform) lingered on the engine until the GC eventually finalized them. Apps that churn communicators or hold many certificates accumulated unmanaged handles between GC passes.

- `Ice.SSL.SSLEngine` now implements `IDisposable`. `Dispose()` walks `_certs` and `_caCerts` and disposes each certificate, then nulls the collections.
- `Internal.Instance.destroy()` calls `_sslEngine?.Dispose()` after the plugin manager is destroyed, and nulls the `_sslEngine` field alongside the other component fields under the mutex.
- Suppressed CA1001 on `Internal.Instance` (Ice uses its own `destroy()` lifecycle, not `IDisposable`) and CA2213 on `SSLEngine._logger` (borrowed from the communicator, not owned).

## Test plan

- [x] `csharp` solution builds with 0 warnings, 0 errors
- [x] Local run of `csharp` IceSSL/configuration suite: 1 succeeded
- [ ] CI green

### DIFF
--- a/csharp/src/Ice/Internal/Instance.cs
+++ b/csharp/src/Ice/Internal/Instance.cs
@@ -22,7 +22,7 @@ public sealed class BufSizeWarnInfo
 }
 
 // CA1001: Instance holds an IDisposable _sslEngine but follows Ice's destroy()-based
-// lifecycle instead of IDisposable. The engine is disposed from destroy().
+// lifecycle instead of IDisposable.
 #pragma warning disable CA1001
 public sealed class Instance
 #pragma warning restore CA1001
@@ -1139,10 +1139,6 @@ public sealed class Instance
         //
         _pluginManager?.destroy();
 
-        // Release the X509Certificate2 instances held by the SSL engine before the Instance goes
-        // away; the certificate handles would otherwise linger until the GC finalizes them.
-        _sslEngine?.Dispose();
-
         lock (_mutex)
         {
             _objectAdapterFactory = null;
@@ -1160,6 +1156,7 @@ public sealed class Instance
             _endpointFactoryManager = null;
             _pluginManager = null;
 
+            _sslEngine?.Dispose();
             _sslEngine = null;
 
             _adminAdapter = null;
--- a/csharp/src/Ice/Internal/Instance.cs
+++ b/csharp/src/Ice/Internal/Instance.cs
@@ -21,7 +21,11 @@ public sealed class BufSizeWarnInfo
     public int rcvSize;
 }
 
+// CA1001: Instance holds an IDisposable _sslEngine but follows Ice's destroy()-based
+// lifecycle instead of IDisposable. The engine is disposed from destroy().
+#pragma warning disable CA1001
 public sealed class Instance
+#pragma warning restore CA1001
 {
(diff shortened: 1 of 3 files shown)


csharp/src/Ice/Internal/Instance.cs
```diff
@@ -21,7 +21,11 @@ public sealed class BufSizeWarnInfo
     public int rcvSize;
 }
 
+// CA1001: Instance holds an IDisposable _sslEngine but follows Ice's destroy()-based
+// lifecycle instead of IDisposable. The engine is disposed from destroy().
+#pragma warning disable CA1001
 public sealed class Instance
+#pragma warning restore CA1001
```
pepone, line 28: An alternative would be to add a `destroy` method in the SslEngine like we have for other classes.

csharp/src/Ice/Internal/Instance.cs
```diff
@@ -1135,6 +1139,10 @@ public void destroy()
         //
         _pluginManager?.destroy();
 
+        // Release the X509Certificate2 instances held by the SSL engine before the Instance goes
```
bernardnormier: I don't find such comment useful: you hold a disposable object and you dispose it in Dispose(Async) or destroy.

csharp/src/Ice/Internal/Instance.cs
```diff
@@ -1135,6 +1139,10 @@ public void destroy()
...
+        // away; the certificate handles would otherwise linger until the GC finalizes them.
+        _sslEngine?.Dispose();
```
bernardnormier: Since `_sslEngine.Dispose` is not thread-safe, should we call it within the mutex lock?

bernardnormier (approved):

csharp/src/Ice/Internal/Instance.cs
```diff
@@ -21,7 +21,11 @@ public sealed class BufSizeWarnInfo
     public int rcvSize;
 }
 
+// CA1001: Instance holds an IDisposable _sslEngine but follows Ice's destroy()-based
+// lifecycle instead of IDisposable. The engine is disposed from destroy().
```
externl: I would remove the part about engine. This comment is only relevant in context to the current PR.
```suggestion
// lifecycle instead of IDisposable.
```

csharp/src/Ice/Internal/Instance.cs
```diff
@@ -1135,6 +1139,10 @@ public void destroy()
         //
         _pluginManager?.destroy();
 
+        // Release the X509Certificate2 instances held by the SSL engine before the Instance goes
+        // away; the certificate handles would otherwise linger until the GC finalizes them.
```
externl: Not sure we need this comment.

externl (approved):

PR #5360, opened 2026-05-27
Open the Slice preprocessor temp file atomically with O_EXCL/O_NOFOLLOW

pepone:
## Summary

- Adds an `openExclusive(path)` helper in `cpp/src/Slice/Preprocessor.cpp` that atomically creates and opens the preprocessor's temporary output file (`O_CREAT|O_EXCL|O_NOFOLLOW` on POSIX, `_O_CREAT|_O_EXCL` on Windows).
- Replaces the Windows `_wtempnam + fopen` sequence and the cross-platform CWD-fallback `fopen("w+")` with this helper.
- The POSIX primary path (`tmpfile()`) was already atomic; left unchanged.

Mitigates a narrow TOCTOU race where a local attacker could plant a symlink or pre-create a file at the generated path between name selection and open. See zeroc-ice/ice-audit#30 (graded Low).

## Test plan

- [ ] CI green across platforms
- [ ] Manual: `slice2cpp` smoke test produces expected `.cpp`/`.h` (verified locally on macOS).

### DIFF
--- a/cpp/src/Slice/Preprocessor.cpp
+++ b/cpp/src/Slice/Preprocessor.cpp
@@ -10,19 +10,59 @@
 #include <algorithm>
 #include <cassert>
 #include <cstring>
+#include <fcntl.h>
 #include <fstream>
 #include <iterator>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <vector>
 
-#ifndef _WIN32
+#ifdef _WIN32
+#    include <io.h>
+#else
 #    include <sys/wait.h>
+#    include <unistd.h>
 #endif
 
 using namespace std;
 using namespace Slice;
 using namespace IceInternal;
+
+namespace
+{
+    // Atomically create and open a new file for read/write, failing if the file already exists or, on POSIX, if the
+    // path resolves to a symlink. This avoids the TOCTOU race that a name-then-open sequence would otherwise expose
+    // to a local attacker who could plant a symlink at the generated path.
+    FILE* openExclusive(const string& path)
+    {
+#ifdef _WIN32
+        const wstring wpath = Ice::stringToWstring(path);
+        int fd = ::_wopen(wpath.c_str(), _O_RDWR | _O_CREAT | _O_EXCL | _O_BINARY, _S_IREAD | _S_IWRITE);
+        if (fd == -1)
+        {
+            return nullptr;
+        }
+        FILE* fp = ::_fdopen(fd, "w+");
+        if (fp == nullptr)
+        {
+            ::_close(fd);
+        }
+        return fp;
+#else
+        int fd = ::open(path.c_str(), O_RDWR | O_CREAT | O_EXCL | O_NOFOLLOW, S_IRUSR | S_IWUSR);
+        if (fd == -1)
+        {
+            return nullptr;
+        }
+        FILE* fp = ::fdopen(fd, "w+");
+        if (fp == nullptr)
+        {
+            ::close(fd);
+        }
+        return fp;
+#endif
+    }
+}
 
 //
 // mcpp defines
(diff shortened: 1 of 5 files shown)


cpp/src/Slice/Preprocessor.cpp
```diff
@@ -9,21 +9,82 @@
...
+                &fd,
+                path,
+                _O_RDWR | _O_CREAT | _O_EXCL | _O_BINARY | _O_TEMPORARY,
+                _SH_DENYRW,
+                _S_IREAD | _S_IWRITE) != 0)
+        {
+            return nullptr;
+        }
+        FILE* fp = ::_fdopen(fd, "w+");
+        if (fp == nullptr)
```
bernardnormier: Our usual style is `if (!fp)`

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -9,21 +9,82 @@
...
+        // If unlink fails for any reason other than ENOENT (which means another process already removed the entry,
+        // which is the state we wanted anyway) we treat it as an open failure: the caller would otherwise have no way
+        // to clean up the stray file once close() relies solely on the descriptor.
+        if (::unlink(path) != 0 && errno != ENOENT)
+        {
+            ::close(fd);
+            return nullptr;
+        }
+        FILE* fp = ::fdopen(fd, "w+");
+        if (fp == nullptr)
```
bernardnormier: Same here

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -201,32 +262,32 @@ Slice::Preprocessor::preprocess(const string& languageArg)
...
         if (_cppHandle == nullptr)
         {
 #ifdef _WIN32
-            _cppFile = "slice-" + Ice::generateUUID();
+            _cppHandle = openExclusive(Ice::stringToWstring("slice-" + Ice::generateUUID()).c_str());
 #else
-            _cppFile = ".slice-" + Ice::generateUUID();
+            _cppHandle = openExclusive((".slice-" + Ice::generateUUID()).c_str());
 #endif
-            _cppHandle = IceInternal::fopen(_cppFile, "w+");
```
bernardnormier, line 229: Should we delete IceInternal::fopen as well? It's no longer called after this PR.

bernardnormier (approved):

cpp/src/Slice/Preprocessor.cpp
pepone: Yes, removed.

externl (approved):

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -256,14 +317,11 @@ Slice::Preprocessor::close()
 {
     if (_cppHandle != nullptr)
     {
+        // The temp file deletes itself when the descriptor is closed: tmpfile() returns an already-unlinked file on
+        // POSIX primary; the POSIX CWD fallback unlinks the path immediately after open; and Windows uses
```
InsertCreativityHere: What is "POSIX primary"? Is primary supposed to be here or just a leftover word?

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -9,21 +9,82 @@
...
+        return fp;
+    }
+#else
+    FILE* openExclusive(const char* path)
+    {
+        int fd = ::open(path, O_RDWR | O_CREAT | O_EXCL | O_NOFOLLOW, S_IRUSR | S_IWUSR);
+        if (fd == -1)
+        {
+            return nullptr;
+        }
```
InsertCreativityHere, line 67: Parts of this comment are weird, it reads like "the code isn't actually fixed yet" and "here's what the code would look like when we fix it".
Instead of describing how the code actually is.

Like the whole last sentence about `once 'close()' relies solely on the descriptor`. It sounds like we haven't actually implemented that, when that's what this PR does. Same with `and no TOCTOU window remains` which is saying what _would_ be the case _after_ we fix the problem.

InsertCreativityHere (approved):
Looks good to me.

cpp/src/Slice/Preprocessor.cpp
pepone: reworked the comment

cpp/src/Slice/Preprocessor.cpp
pepone: reworked the comment, that was about primary path but was indeed unclear

PR #5364, opened 2026-05-28
Check for null IP connection info in IceGrid and Glacier2

bernardnormier:
This is tiny cleanup PR.


pepone (approved):
Looks good, not sure how realistic is to run Glacier2 router over BT.

bernardnormier:
> Looks good, not sure how realistic is to run Glacier2 router over BT.

I agree. But if you do, it would be strange that the only API that fails is the creation of SSL-based sessions.

externl (approved):

PR #5392, opened 2026-05-29
Fix string converter usage in OutputStream

### LINKED ISSUE #5387: Performance: getProcessStringConverter() takes a global lock on every string marshal
> [!NOTE]
> **AI-assisted finding — needs human triage.** This issue was produced by an automated performance audit of the Ice C++ runtime and verified by re-reading the cited code against source at commit `c617a356`. It was **not** benchmarked, profiled, or proven at runtime — the impact rating is a static-review estimate. Please confirm the finding and measure the impact before acting on it.

## Summary

The default string-marshaling path acquires a process-wide mutex on every converted string write, even in the common case where no string converter is installed. This is genuine cross-thread lock contention on the marshaling hot path.

The default `OutputStream::write(string_view, convert = true)` routes to `writeConverted()`:

**File:** `cpp/include/Ice/OutputStream.h:543`
```cpp
void write(std::string_view v, bool convert = true)
{
    auto sz = static_cast<std::int32_t>(v.size());
    if (convert && sz > 0)
    {
        writeConverted(v.data(), static_cast<size_t>(sz));
    }
    ...
```

`writeConverted()` falls back to `getProcessStringConverter()` whenever no per-stream converter is set (the default):

**File:** `cpp/src/Ice/OutputStream.cpp:706`
```cpp
StringConverterPtr stringConverter = _stringConverter;
if (!stringConverter)
{
    stringConverter = getProcessStringConverter();
}
```

and `getProcessStringConverter()` takes a `lock_guard` on a global mutex — even when the converter is null:

**File:** `cpp/src/Ice/StringConverter.cpp:227`
```cpp
StringConverterPtr
Ice::getProcessStringConverter()
{
    lock_guard lock(processStringConverterMutex);
    return processStringConverter;
}
```

So every string marshaled by a default-configured application serializes through one global mutex.

**Severity:** High (top priority — this removes a *global lock*, not just a copy).
**Binary-compatible:** yes — the process-converter globals live in `StringConverter.cpp`; no public type or signature changes.

## Fix

Store the process string-converter globals (narrow and wide get/set) in `std::atomic<std::shared_ptr<...>>` and replace the locked getters/setters with atomic load/store. Also drop the redundant `shared_ptr` copy of `_stringConverter` at `OutputStream.cpp:706` (read it once).

This change has a concurrency-correctness story, so it is worth isolating in its own commit.

bernardnormier:
This PR fixes the string converter and wstring converter usage in the OutputStream implementation, to avoid retrieving the process-wide stringConverter (resp. wstringConverter) each time we marshal a string.

See #5387.

I didn't fix #5387 as suggested because `std::atomic<std::shared_ptr<T>>` is only available as of c++20, and Xcode clang on macos - that's supposed to support c++20 - doesn't actually provide this library feature.

### DIFF
--- a/cpp/src/Ice/OutputStream.cpp
+++ b/cpp/src/Ice/OutputStream.cpp
@@ -68,6 +68,10 @@ Ice::OutputStream::OutputStream(
       _format(format),
       _currentEncaps(nullptr)
 {
+    if (!_wstringConverter)
+    {
+        _wstringConverter = getProcessWstringConverter();
+    }
 }
 
 Ice::OutputStream::OutputStream(
@@ -85,6 +89,11 @@ Ice::OutputStream::OutputStream(
       _currentEncaps(nullptr)
 {
     b.reset();
+
+    if (!_wstringConverter)
+    {
+        _wstringConverter = getProcessWstringConverter();
+    }
 }
 
 Ice::OutputStream::OutputStream(const CommunicatorPtr& communicator, EncodingVersion encoding)
@@ -104,6 +113,7 @@ Ice::OutputStream::OutputStream(Instance* instance, EncodingVersion encoding)
           instance->getStringConverter(),
           instance->getWstringConverter())
 {
+    assert(_wstringConverter);
 }
 
 Ice::OutputStream::OutputStream(OutputStream&& other) noexcept
@@ -116,6 +126,7 @@ Ice::OutputStream::OutputStream(OutputStream&& other) noexcept
       _currentEncaps(other._currentEncaps)
 {
     // Reset other to its default state.
+    other._wstringConverter = getProcessWstringConverter();
     other._closure = nullptr;
     other._encoding = Encoding_1_1;
     other._format = FormatType::CompactFormat;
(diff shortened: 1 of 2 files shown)


cpp/src/Ice/OutputStream.cpp
```diff
@@ -104,6 +113,7 @@ Ice::OutputStream::OutputStream(Instance* instance, EncodingVersion encoding)
           instance->getStringConverter(),
```
bernardnormier, line 113: This is the common path. `instance->getStringConverter()` gets the cached value in Instance without locking any mutex.

cpp/src/Ice/OutputStream.cpp
```diff
@@ -703,13 +715,7 @@ Ice::OutputStream::write(const char*)
 void
 Ice::OutputStream::writeConverted(const char* vdata, size_t vsize)
 {
-    StringConverterPtr stringConverter = _stringConverter;
-    if (!stringConverter)
-    {
-        stringConverter = getProcessStringConverter();
```
bernardnormier, line 709: Removed the "get that locks a global mutex" that was the focus of #5387.

externl (approved):
Looks good!

pepone (approved):

cpp/src/Ice/OutputStream.cpp
```diff
@@ -116,6 +126,7 @@ Ice::OutputStream::OutputStream(OutputStream&& other) noexcept
       _currentEncaps(other._currentEncaps)
 {
     // Reset other to its default state.
+    other._wstringConverter = getProcessWstringConverter();
     other._closure = nullptr;
```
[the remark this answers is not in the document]
bernardnormier: This is not a real concern.

PR #5397, opened 2026-05-30
Add AGENTS.md with AI security-audit guidelines

pepone:
Adds `AGENTS.md`, with guidelines doc for AI coding agents (Codex, Claude Code, Copilot, etc.) running **security audits** of this repository. It documents:

- the two-repo model (`zeroc-ice/ice` public, `zeroc-ice/ice-audit` private);
- the mandatory two-independent-pass workflow (initial report + verification) and the verdict vocabulary;
- issue routing by severity — security-sensitive Medium-and-above go private, everything else public;
- issue conventions (labels, milestone, AI-generated banner, commit-pinned citations);
- the catalogue of already-dismissed audit patterns so agents don't re-file findings closed as "Not planned".

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,209 @@
+# Repository Guidelines for AI Coding Agents — Security Audits
+
+These conventions apply to all AI coding assistants (Claude Code, Codex, Copilot, etc.) performing
+**security audits** of this repository. They cover how an audit is run, how findings are verified,
+where issues are filed, and which findings have already been considered and dismissed.
+
+## The two repositories
+
+| Repo | Visibility | Role |
+|------|-----------|------|
+| [`zeroc-ice/ice`](https://github.com/zeroc-ice/ice) | **public** | The Ice source. Fixes land here as normal PRs. |
+| [`zeroc-ice/ice-audit`](https://github.com/zeroc-ice/ice-audit) | **private** | Audit reports and the finding tracker. |
+
+Prior audit reports live at the root of `ice-audit`, e.g.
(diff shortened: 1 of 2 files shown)


bernardnormier (approved):

bernardnormier:
Since an agent can only run such an audit when it has access to both repos (ice and ice-audit), should we put this AGENTS.md in the ice-audit repo?

AGENTS.md
```diff
@@ -0,0 +1,209 @@
+# Repository Guidelines for AI Coding Agents — Security Audits
```
externl, line 1: AGENTS.md is not just about security audits. I would just go with the typical `# AGENTS.md`

See https://agents.md


```suggestion
# AGENTS.md
```

AGENTS.md
```diff
@@ -0,0 +1,209 @@
...
+
+These conventions apply to all AI coding assistants (Claude Code, Codex, Copilot, etc.) performing
+**security audits** of this repository. They cover how an audit is run, how findings are verified,
+where issues are filed, and which findings have already been considered and dismissed.
```
externl, line 5: Is this necessary? Seems like wasted tokens/context.

externl (commented):
**_If_** I understand how AGENTS.md is supposed to work. See also https://agents.md

- I think this is too bloated. AGENTS.md is meant to be read by agentic agents per session. it should be relatively lean to limit token usage. 
- It's not just about security audits.

We should have something more like Homebrew's:
- https://github.com/Homebrew/brew/blob/main/AGENTS.md
- https://github.com/Homebrew/homebrew-core/blob/main/AGENTS.md

if we want something an agent reads when doing audit analytics I think belongs elsewhere.

pepone:
I will move this to ai-audit, and keep as a security audit recipe instead of as the general agents file

PR #5399, opened 2026-06-01
Fix DataStorm relay topic-announcement loss 

### LINKED ISSUE #5359: DataStorm/reliability C++ test hangs in 2-relay-node case across node-restart
The C++ `DataStorm/reliability` test hung in CI for ~5 minutes and was killed by the test runner. The failing iteration is one of the 2-relay-node cases in `cpp/test/DataStorm/reliability/test.py` (writer and reader connect to *different* relay nodes).

## Environment

- Platform: Windows (GitHub Actions, `D:\a\ice\ice\cpp\...`), Win32 / Release.
- Test: `cpp/test/DataStorm/reliability` (driver runs all cases in `test.py`).
- The case that hangs is one of the entries at indices 10–15 in `props` (the 2-node variants), with the reader connected to `relay-node-2` (port 14212) and the writer connected to `relay-node-1` (port 14211). Logs match the `reversedStart=True` 2-node variant.
- The hang occurs on the **first subtest** of `Writer.cpp`/`Reader.cpp` — `"testing writer connection closure"` (the `string` topic block).

## Symptom

- `writer.exe`: connects to `relay-node-1`, finishes session setup (`s/1`, `p/1`), attaches topic `0:string:[k2]`/`[k3]`, then only sends/receives idle validate-connection heartbeats for ~5 minutes (waiting in `barrier.getNextUnread()` in `Writer.cpp:31`).
- `reader.exe`: connects to `relay-node-2`, announces `topic writer 'string'` to `reader-app`, establishes its node session, then sits idle for ~5 minutes (waiting in `barrier.waitForReaders()` in `Reader.cpp:29`).
- `node1 ↔ node2` bridge connection is established but only exchanges heartbeats.
- Test driver eventually kills the processes; node1 logs `ConnectionLostException` from the killed writer.

## Timeline (run timestamps `19:35:*`)

| Time | Event |
|------|-------|
| 19:35:09.596 | Previous 2-node iteration ends; `node2` disconnects from `node1`. |
| 19:35:10.146 | `node1` stops accepting on 14211 (relay restart between cases). |
| 19:35:11.663 | `node1` listens again. |
| 19:35:11.691 | `reader.exe` starts → connects to `node2` (14212), announces topic. |
| 19:35:11.695 | `writer.exe` starts → connects to `node1` (14211). |
| **19:35:11.701** | `node2` records `topic writer(s) 'string' announced (peer = 'reader-app')`. |
| **19:35:11.712** | `node2 → node1` bridge connection (re)created (~10 ms *after* reader-app's announce on `node2`). |
| 19:35:11+ → 19:40:11 | Only validate-connection heartbeats on every leg. `node1` never receives the reader-app announcement forwarded across the bridge. |
| 19:40:13.419 | Test runner kills the processes; `ConnectionLostException` from the killed writer.exe. |

## Diagnosis

Compare against the prior 2-node iteration at 19:35:09 where both apps happened to land on `node1` directly — full topic announces propagate and samples flow.

In the failing iteration, the topology is split: `writer-app → node1`, `reader-app → node2`. `node2` *does* record reader-app's topic announce, but the `node2 → node1` bridge was created ~10 ms *after* that announce arrived. The bridge then carries no topic-announce traffic — only heartbeats — so writer-app never learns about the remote barrier writer and the test deadlocks.

Most likely cause: when the node session between the two relays re-establishes after a restart, topic announcements that arrived *before* the bridge came up are not flushed across the new bridge. Area to investigate: the relay/forwarding path in `cpp/src/DataStorm` that re-syncs topic announcements when a node session (`NodeSessionI`) transitions disconnected → connected.

## Logs

[datastorm-reliability-052626-1935.zip](https://github.com/user-attachments/files/28301397/datastorm-reliability-052626-1935.zip)

pepone:
Fix #5359

### DIFF
--- a/cpp/src/DataStorm/NodeSessionManager.cpp
+++ b/cpp/src/DataStorm/NodeSessionManager.cpp
@@ -95,6 +95,8 @@ NodeSessionManager::createOrGet(NodePrx node, const ConnectionPtr& newConnection
         // connection
         if (p->second->getConnection() != newConnection)
         {
+            // The node reconnected over a new connection; drop the announcements received over the old one.
+            removeAnnouncements(p->second->getConnection());
             p->second->destroy();
             _sessions.erase(p);
         }
@@ -119,6 +121,14 @@ NodeSessionManager::createOrGet(NodePrx node, const ConnectionPtr& newConnection
         make_shared<NodePrx>(node),
         [self = shared_from_this(), node = std::move(node)](const ConnectionPtr& connection, exception_ptr) mutable
         { self->destroySession(connection, node); });
+
+    // When this session forwards announcements, replay to the new peer the announcements we received earlier from
+    // other peers. Otherwise an announcement received before this session existed would never reach the new peer, and
+    // applications on opposite sides of a relay chain could wait forever for discovery.
+    if (auto lookup = session->getLookup())
+    {
+        announceKnownTopics(*lookup, newConnection);
+    }
 
     return session;
 }
@@ -146,7 +156,13 @@ NodeSessionManager::announceTopicReader(const string& topic, NodePrx node, const
     }
 
     auto p = _sessions.find(node->ice_getIdentity());
-    node = p != _sessions.end() ? p->second->getPublicNode() : node;
+    if (p != _sessions.end())
+    {
+        node = p->second->getPublicNode();
+    }
+
+    // Remember the announcement so it can be replayed when a node session is (re)established later.
+    recordAnnouncement(node, {topic}, {}, connection);
 
     // Set the exclude connection to prevent forwarding the announcement back to the sender.
     _exclude = connection;
@@ -193,7 +209,13 @@ NodeSessionManager::announceTopicWriter(const string& topic, NodePrx node, const
     }
 
     auto p = _sessions.find(node->ice_getIdentity());
-    node = p != _sessions.end() ? p->second->getPublicNode() : node;
+    if (p != _sessions.end())
+    {
+        node = p->second->getPublicNode();
+    }
+
+    // Remember the announcement so it can be replayed when a node session is (re)established later.
+    recordAnnouncement(node, {}, {topic}, connection);
 
     // Set the exclude connection to prevent forwarding the announcement back to the sender.
     _exclude = connection;
@@ -258,7 +280,13 @@ NodeSessionManager::announceTopics(
     }
 
     auto p = _sessions.find(node->ice_getIdentity());
-    node = p != _sessions.end() ? p->second->getPublicNode() : node;
+    if (p != _sessions.end())
+    {
+        node = p->second->getPublicNode();
+    }
+
+    // Remember the announcement so it can be replayed when a node session is (re)established later.
+    recordAnnouncement(node, readers, writers, connection);
 
     // Set the exclude connection to prevent forwarding the announcement back to the sender.
     _exclude = connection;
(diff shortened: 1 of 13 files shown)


cpp/src/DataStorm/NodeSessionManager.cpp
```diff
@@ -258,7 +286,16 @@ NodeSessionManager::announceTopics(
...
+    if (p != _sessions.end())
+    {
+        node = p->second->getPublicNode();
+    }
+
+    // Remember the announcement so it can be replayed when a node session is (re)established later.
+    if (connection && hasNodeSession(connection))
+    {
+        recordAnnouncement(node, readers, writers, connection);
+    }
```
externl: We do this a few times. Is it worth a utility function?

cpp/src/DataStorm/NodeSessionManager.h
```diff
@@ -40,11 +42,50 @@ namespace DataStormI
...
+        // deliberately not cached. Must be called with `_mutex` held.
+        [[nodiscard]] bool hasNodeSession(const Ice::ConnectionPtr&) const;
+
         void connect(const DataStormContract::LookupPrx&, const DataStormContract::NodePrx&);
 
         void connected(const DataStormContract::NodePrx&, const DataStormContract::LookupPrx&);
 
-        void disconnected(const DataStormContract::NodePrx&, const DataStormContract::LookupPrx&);
+        void
+        disconnected(const Ice::ConnectionPtr&, const DataStormContract::NodePrx&, const DataStormContract::LookupPrx&);
```
bernardnormier, line 88: missing blank line

bernardnormier (approved):

cpp/test/DataStorm/relayReconnect/Reader.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+class Reader : public Test::TestHelper
+{
+public:
+    Reader() : Test::TestHelper(false) {}
+
+    void run(int, char**) override;
+};
+
+void ::Reader::run(int argc, char* argv[])
```
externl, line 17: Is the global scope necessary?

```suggestion
void Reader::run(int argc, char* argv[])
```

cpp/test/DataStorm/relayReconnect/Reader.cpp
```diff
@@ -0,0 +1,38 @@
...
+    // Block until the writer's sample is received across the relay chain. This only completes once
+    // discovery propagates across the relay path that is completed after both announcements were
+    // received - the failure mode this test reproduces.
+    auto sample = reader.getNextUnread();
+    test(sample.getValue() == 42);
+
+    cout << "reader completed" << endl;
+}
+
+DEFINE_TEST(::Reader)
```
externl, line 38:
```suggestion
DEFINE_TEST(Reader)
```

scripts/DataStormUtil.py
```diff
@@ -1,13 +1,31 @@
...
 
+def waitForLogMessage(path, pattern, timeout=60):
+    """Wait for a regular-expression match in a log file.
+
+    Used by DataStorm tests to synchronize on trace messages written to an Ice.LogFile."""
+    expr = re.compile(pattern)
+    deadline = time.time() + timeout
+    while time.time() < deadline:
+        try:
+            with open(path, "r") as file:
```
externl: Instead of reopening the file over and over, I think we can do something like:

```python
 with open(path, 'r', encoding='utf-8') as f:
        while True:
            line = f.readline()
            if not line:
                time.sleep(0.1)
                continue
            if expr.search(line):
                return
  ...
```

externl (approved):

cpp/src/DataStorm/NodeSessionManager.cpp
pepone: These are two separate concerns, I think is fine to keep this inline.

cpp/test/DataStorm/relayReconnect/Reader.cpp
pepone: Yes it is, because we have `using DataStorm` which defines `Reader`.

PR #5400, opened 2026-06-01
Use string converter when marshaling exceptions

bernardnormier:
This PR fixes the marshaling of exceptions to correctly use the configured string/wstring converters (if any).

This is a follow-up to #5392.

### DIFF
--- a/cpp/src/Ice/ConnectionI.cpp
+++ b/cpp/src/Ice/ConnectionI.cpp
@@ -12,7 +12,7 @@
 #include "Ice/Properties.h"
 #include "IdleTimeoutTransceiverDecorator.h"
 #include "Instance.h"
-#include "ObjectAdapterI.h"   // For getThreadPool()
+#include "ObjectAdapterI.h" // For getThreadPool()
 #include "OutgoingResponseInternal.h"
 #include "ReferenceFactory.h" // For createProxy().
 #include "RequestHandler.h"   // For RetryException
--- a/cpp/src/Ice/ConnectionI.cpp
+++ b/cpp/src/Ice/ConnectionI.cpp
@@ -13,6 +13,7 @@
 #include "IdleTimeoutTransceiverDecorator.h"
 #include "Instance.h"
 #include "ObjectAdapterI.h"   // For getThreadPool()
+#include "OutgoingResponseInternal.h"
 #include "ReferenceFactory.h" // For createProxy().
 #include "RequestHandler.h"   // For RetryException
 #include "ThreadPool.h"
@@ -3453,9 +3454,10 @@ Ice::ConnectionI::dispatchAll(
             {
                 // Received request on a connection without an object adapter.
                 sendResponse(
-                    makeOutgoingResponse(
+                    makeOutgoingResponseCore(
                         make_exception_ptr(ObjectNotExistException{__FILE__, __LINE__}),
-                        request.current()),
+                        request.current(),
+                        _instance.get()),
                     0);
             }
 
--- a/cpp/src/Ice/OutgoingResponse.cpp
+++ b/cpp/src/Ice/OutgoingResponse.cpp
@@ -2,9 +2,11 @@
 
 #include "Ice/OutgoingResponse.h"
 #include "Ice/Demangle.h"
+#include "Ice/Initialize.h"
 #include "Ice/LocalExceptions.h"
 #include "Ice/ObjectAdapter.h"
 #include "Ice/UserException.h"
+#include "OutgoingResponseInternal.h"
 #include "Protocol.h"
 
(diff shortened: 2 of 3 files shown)


cpp/src/Ice/OutgoingResponse.cpp
```diff
@@ -30,131 +32,6 @@ namespace
         os << "dispatch failed with " << typeId << ": " << what;
         return os.str();
     }
-
-    // The "core" implementation of makeOutgoingResponse for exceptions. Note that it can throw an exception.
```
bernardnormier, line 34: Moved as-is (except for the ostr construction) to a function with the same name in namespace Ice. See below.

cpp/src/Ice/OutgoingResponse.cpp
```diff
@@ -264,13 +141,136 @@ Ice::makeOutgoingResponse(bool ok, pair<const byte*, const byte*> encapsulation,
 OutgoingResponse
 Ice::makeOutgoingResponse(std::exception_ptr exc, const Current& current) noexcept
 {
+    IceInternal::Instance* instance = getInstance(current.adapter->getCommunicator()).get();
```
bernardnormier, line 144: In this public/exported function, current.adapter is guaranteed to be non-null.

cpp/src/Ice/ConnectionI.cpp
```diff
@@ -3453,9 +3454,10 @@ Ice::ConnectionI::dispatchAll(
             {
                 // Received request on a connection without an object adapter.
                 sendResponse(
-                    makeOutgoingResponse(
+                    makeOutgoingResponseCore(
```
bernardnormier, line 3457: That's the one situation where we "export" a current with a null current.adapter.

cpp/src/Ice/OutgoingResponseInternal.h
```diff
@@ -0,0 +1,23 @@
...
+#include "Ice/OutgoingResponse.h"
+
+namespace Ice
+{
+    /// @private
+    /// The "core" implementation of makeOutgoingResponse for exceptions. Note that it can throw an exception.
+    /// @param exception The exception to marshal into the response.
+    /// @param current A reference to the Current object of the request. @c current.adapter can be null here; the
+    /// implementation uses @p instance instead.
+    /// @param instance The IceInternal::Instance object. Not null.
```
externl, line 15: Missing `@throws`. We mention this in the summary and in the comment on the implementation. But glancing at the code its not obvious what we're referring to.

cpp/src/Ice/OutgoingResponse.cpp
```diff
@@ -264,13 +141,136 @@ Ice::makeOutgoingResponse(bool ok, pair<const byte*, const byte*> encapsulation,
...
+        return makeOutgoingResponseCore(current_exception(), current, instance);
     }
 }
+
+// The "core" implementation of makeOutgoingResponse for exceptions. Note that it can throw an exception.
+OutgoingResponse
+Ice::makeOutgoingResponseCore(std::exception_ptr exc, const Current& current, Instance* instance)
+{
+    assert(exc);
+    OutputStream ostr{instance, currentProtocolEncoding};
```
bernardnormier, line 162: It's important to construct OutputStream with an instance here to get the communciator's string converter / wstring converter _in case_ we marshal an error message or a user exception with string/wstring fields.

The previous version neglected to do that, which was incorrect.

cpp/src/Ice/OutgoingResponse.cpp
```diff
@@ -264,13 +141,136 @@ Ice::makeOutgoingResponse(bool ok, pair<const byte*, const byte*> encapsulation,
...
+
+    if (current.requestId != 0)
```
bernardnormier, line 164: None of the code below changed. Just moved.

cpp/src/Ice/ConnectionI.cpp
externl, line 3457: Not obvious to me that the behavior is the same regarding exceptions, maybe that's fine.

externl (approved):

cpp/src/Ice/ConnectionI.cpp
bernardnormier: makeOutgoingResponseCore is not expected to throw for a simple exception like ObjectNotExistException.

pepone (approved):

cpp/src/Ice/OutgoingResponseInternal.h
bernardnormier: It can throw any exception. Keep in mind this is an internal, non-documented and non-exported function.

PR #5405, opened 2026-06-02
Avoid copying the key in the generic dictionary read

### LINKED ISSUE #5404: Performance: generic dictionary read copies the key instead of moving it
The generic Slice dictionary read helper deep-copies each key into the map instead of moving it.

**File:** `cpp/include/Ice/StreamHelpers.h:333`
```cpp
typename T::value_type p;                                  // pair<const K, V>
stream->read(const_cast<typename T::key_type&>(p.first));
auto i = v.insert(v.end(), p);                             // copies the key
stream->read(i->second);
```

Because `value_type` is `pair<const K, V>`, the key is already `const` in `p`, so `insert(value_type)` can only copy it into the node — one extra heap allocation per entry for every dictionary read.

## Fix
```cpp
typename T::key_type key;
stream->read(key);
auto i = v.emplace_hint(v.end(), std::move(key), typename T::mapped_type());
stream->read(i->second);
```
`emplace_hint` constructs the node's `const` key directly from a movable local (no copy), while keeping the in-place value read and the end-hint.

- **Strictly better:** measured faster for both sorted and unsorted input, all sizes; no new type requirements (the old code already default-constructs key and value).
- **Correctness:** identical results for `std::map`/`std::unordered_map`, including duplicate-key last-wins.
- **Binary-compatible:** yes (header-only inline).

pepone:
Fixes #5404

### DIFF
--- a/cpp/include/Ice/StreamHelpers.h
+++ b/cpp/include/Ice/StreamHelpers.h
@@ -330,9 +330,9 @@ namespace Ice
             v.clear();
             while (sz--)
             {
-                typename T::value_type p;
-                stream->read(const_cast<typename T::key_type&>(p.first));
-                auto i = v.insert(v.end(), p);
+                typename T::key_type key;
+                stream->read(key);
+                auto i = v.emplace_hint(v.end(), std::move(key), typename T::mapped_type());
                 stream->read(i->second);
             }
         }


externl (approved):

cpp/include/Ice/StreamHelpers.h
```diff
@@ -330,9 +330,9 @@ namespace Ice
             v.clear();
             while (sz--)
             {
-                typename T::value_type p;
-                stream->read(const_cast<typename T::key_type&>(p.first));
-                auto i = v.insert(v.end(), p);
+                typename T::key_type key;
+                stream->read(key);
+                auto i = v.emplace_hint(v.end(), std::move(key), typename T::mapped_type());
```
bernardnormier, line 335: We can't do that, especially in a patch release. See:
https://docs.zeroc.com/ice/3.8/cpp/dictionaries#Dictionaries-CustomizingtheDictionaryMappingwithcpp:type

We document the requirements for a custom dictionary, and here you're changing them.

pepone: Ok, let's delay to 3.9, part of the problem is we forget to update this docs when we move to C++17

> Less formally, this means you can use any class or template class that looks like a standard map or unordered_map as your custom dictionary type.

The emplace_hint is part of this "less formally"...

PR #5349, opened 2026-05-25
Enable only forward-secret cipher suites in the macOS SSL transport

pepone:
Restrict the SecureTransport cipher suites to forward-secret ECDHE AES-GCM.

PR #5431, opened 2026-06-08
Harden the macOS SecureTransport defaults: ciphers, minimum TLS version, temporary keychain (3.7.12)

pepone:
Backport of #5349 (C2), #5342 (C1), and #5343 (C3) to the 3.7 branch, part of the 3.7.12 security release. Combined since all three modify the macOS IceSSL SecureTransport engine.

In 3.7 the engine only called `SSLSetEnabledCiphers`/`SSLSetProtocolVersionMin` when the corresponding property was set, so the **default** configuration was vulnerable:

- **#5349** — with no ciphers configured, SecureTransport used its default set (static-RSA, no forward secrecy, plus 3DES/SWEET32). The engine now enables only forward-secret ECDHE+AES-GCM suites by default.
- **#5342** — with no minimum protocol version configured, SecureTransport could negotiate down to TLS 1.0. The engine now requires TLS 1.2 by default.

An explicit `IceSSL.Ciphers` / `IceSSL.ProtocolVersionMin` setting still takes precedence.

- **#5343** — with the forward-secret default ciphers, a private key in the user's **login keychain** cannot complete an ECDHE handshake on the server side. So when `IceSSL.Keychain` is unset, the certificate is now imported into a private **temporary keychain** (created with `mkdtemp` under `_CS_DARWIN_USER_TEMP_DIR`, removed by the engine destructor) instead of the login keychain. C3 is what makes the C2 default actually usable on macOS.

**Behavior changes:** peers offering only static-RSA/3DES ciphers, or limited to TLS 1.0/1.1, can no longer connect to a macOS peer using the default configuration. The obsolete SSLv3 cases were removed from the `IceSSL/configuration` test (with forward-secret default ciphers an SSLv3 handshake has no compatible cipher suite), and the test no longer pre-creates per-test keychains.

Verified the full `IceSSL/configuration` suite passes on macOS. macOS C++ IceSSL only.

### DIFF
--- a/CHANGELOG-3.7.md
+++ b/CHANGELOG-3.7.md
@@ -145,6 +145,17 @@ These are the changes since Ice 3.7.11.
 - Fixed a null pointer dereference in the IceGrid registry and Glacier2 router when creating an
   SSL-based session over a non-IP transport (such as Bluetooth).
 
+- Changed the macOS SSL transport to enable only forward-secret (ECDHE) cipher suites when no
+  ciphers are explicitly configured. An `IceSSL.Ciphers` setting still takes precedence.
+
+- Changed the macOS SSL transport to require TLS 1.2 or later, by changing the default value of
+  `IceSSL.ProtocolVersionMin` to `tls1_2`. An explicit `IceSSL.ProtocolVersionMin` setting still
+  takes precedence.
+
+- Changed the macOS SSL transport so that, when `IceSSL.Keychain` is not set, the certificate
+  configured with `IceSSL.CertFile` is imported into a temporary keychain (removed when the
+  communicator is destroyed) instead of the user's login keychain.
+
 ## Swift Changes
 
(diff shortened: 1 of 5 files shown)


CHANGELOG-3.7.md
```diff
@@ -134,6 +134,23 @@ These are the changes since Ice 3.7.11.
 - The Glacier2 `crypt` permissions verifier now logs a warning at startup when its password file
   contains DES-style password hashes, which rely on a weak algorithm.
 
+- On macOS, the IceSSL SecureTransport engine now enables only forward-secret cipher suites (ECDHE
```
bernardnormier: We should use a simpler entry derived from the CHANGELOG-3.8 entry. It's not exactly the same since we still use `IceSSL.ProtocolVersionMin` in 3.7.

bernardnormier (approved):

PR #5433, opened 2026-06-09
Update version to 3.7.12

pepone:
Bumps the version from 3.7.11 to 3.7.12 across all version-carrying files (canonical `config/version.env` + `config/Make.rules`, `ICE_STRING_VERSION`/`ICE_INT_VERSION`, NPM base version, msbuild nuspec/props/targets, language manifests, SDK plists, MATLAB, Swift xcodeproj, BUILDING.md, and the dash-format release-notes URLs).

Changelog entries are **added**, not rewritten: a new `(3.7.12-1)` Debian changelog entry and a new RPM `%changelog` entry (with `Release` reset to 1); the released 3.7.11 entries are preserved.

Intentionally untouched: `CHANGELOG-3.7.md` (so this doesn't conflict with the open 3.7.12 security PRs), the MSI/RPM `UpgradeCode` (must stay constant for major upgrades), and `NIGHTLY.md` / `build-release.yml` dated examples.

**Open item:** the MSI `ProductCode` in `packaging/msi/Ice.aip` should be regenerated to a new GUID for the major upgrade — to be done in Advanced Installer.


externl (approved):
Wow, that's a lot of files!

cpp/include/IceUtil/ResourceConfig.h
```diff
@@ -8,7 +8,7 @@
 #include "winver.h"
 
 #define ICE_VERSION 3,7,11,0
-#define ICE_STRING_VERSION "3.7.11\0"
+#define ICE_STRING_VERSION "3.7.12\0"
```
bernardnormier, line 11: Missing update above

bernardnormier (approved):

InsertCreativityHere (approved):
Looks good to me!
I checked every instance of `3.?7` in the repo, and the only 'serious' one leftover was the macro definition in `ResourceConfig.h`.

But we also have:
- 3 instances of `3.7.11` left in `NIGHTLY.md` under the [section for Ruby](https://github.com/zeroc-ice/ice/blob/3.7/NIGHTLY.md#ice-for-ruby), and a less important one in an example for how to install the [nightly Windows C++](https://github.com/zeroc-ice/ice/blob/3.7/NIGHTLY.md#windows-)
- 3 instances of `3.7.10` left in `index-ice-pdb-sources.py` [here](https://github.com/zeroc-ice/ice/blob/3.7/packaging/release/index-ice-pdb-sources.py). But, they're only in examples, not actual code. So, not important to update.
- 1 instance of `3.7.9` in `bug_report.yml` [here](https://github.com/zeroc-ice/ice/blob/3.7/.github/ISSUE_TEMPLATE/bug_report.yml). Again, it's just an example, but might be worth updating to the latest since that's what users will see when filing bugs by default.

PR #5438, opened 2026-06-10
Fix sample code in README.md

### LINKED ISSUE #5001: Fix code sample in C# README.md (3.7)
The C# code is not quite correct in the code samples.

bernardnormier:
Fixes #5001.

### DIFF
--- a/csharp/README.md
+++ b/csharp/README.md
@@ -23,30 +23,33 @@ module Demo
 ```csharp
 // Client application
 using(var communicator = Ice.Util.initialize(ref args))
-var hello = HelloPrxHelper.checkedCast(
-    communicator.stringToProxy("hello:default -h localhost -p 10000"));
-hello.sayHello();
+{
+    var hello = Demo.HelloPrxHelper.checkedCast(
+        communicator.stringToProxy("hello:default -h localhost -p 10000"));
+    hello.sayHello();
+}
 ```
 
 ```csharp
 // Server application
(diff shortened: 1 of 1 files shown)


csharp/README.md
```diff
@@ -23,30 +23,33 @@ module Demo
 ```csharp
 // Client application
 using(var communicator = Ice.Util.initialize(ref args))
```
bernardnormier, line 25: This is really a simplified version of the minimal demo.

csharp/README.md
```diff
@@ -23,30 +23,33 @@ module Demo
...
+}
 ```
 
 ```csharp
 // Server application
 using(var communicator = Ice.Util.initialize(ref args))
-
-// Shut down the communicator on Ctrl+C or Ctrl+Break.
-Console.CancelKeyPress += (sender, eventArgs) =>
 {
```
bernardnormier, line 36: The main fix are the missing scopes.

csharp/README.md
pepone, line 25: Can we switch to the new `using` style without `() {}`?

pepone (approved):

csharp/README.md
bernardnormier: using statements were introduced in C# 8.0.

The idea is to keep parity with the 3.7 demos, that still supports old .NET and old C#.  This code sample is derived from the minimal demo.

PR #5493, opened 2026-06-12
Fix infinite loop in Slice compilers

pepone:
on unterminated {@link doc-comment tag

formatLinkElement did nothing when a '{@link' tag had no closing '}' on the same line, leaving 'pos' unchanged. The caller loop then re-found the same tag at the same position forever, hanging every slice2* compiler at 100% CPU on such input.

Mirror formatCodeElement: emit an 'unterminated link tag' warning and skip the rest of the line. Also trim whitespace around the link identifier; the previous find_last_not_of started at the closing brace itself and never trimmed anything.

### DIFF
--- a/cpp/src/Slice/DocCommentParser.cpp
+++ b/cpp/src/Slice/DocCommentParser.cpp
@@ -325,12 +325,24 @@ namespace
         const string linkTag = "{@link ";
 
         auto endpos = line.find('}', pos);
-        if (endpos != string::npos)
+        if (endpos == string::npos)
         {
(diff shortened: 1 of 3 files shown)


InsertCreativityHere (approved):
Looks good.
Checked the logic for code-formatting and `@param`, and that already seems fine.

Also never realized that that `find_last_not_of`'s end bound was _inclusive_... Time to check all of our usage of that too! :)

bernardnormier (approved):

PR #5494, opened 2026-06-12
Fix Schannel certificate-context reference counting (IceSSL)

### LINKED ISSUE #5520: Schannel SSL engine mismanages certificate-context reference counts (per-connection leak; FindCert stores unowned references)
## Summary

While investigating an internal audit finding that claimed a use-after-free on the `IceSSL.FindCert` Schannel path, we determined there is **no use-after-free** — the certificate contexts handed to Schannel as credentials remain valid because the certificate store keeps them alive for the engine's lifetime. However, the investigation uncovered two genuine `CERT_CONTEXT` reference-count defects in the Schannel SSL engine.

## Defect 1 — credential-selection callbacks leak one certificate reference per connection

`createClientAuthenticationOptions` / `createServerAuthenticationOptions` bumped the reference count of every configured certificate and **discarded the result**:

```cpp
for (const auto& cert : _allCerts)
{
    CertDuplicateCertificateContext(cert); // return value discarded
}
```

Nothing ever frees that reference. The transceiver already takes its own reference (it duplicates each `paCred` context into its own list and frees those in `close()`), so the engine's bump is a pure leak: **one leaked `CERT_CONTEXT` reference per certificate per connection**. This also prevents the certificate store from ever closing cleanly.

## Defect 2 — `findCertificates` (IceSSL.FindCert) stored references it did not own

The enumeration loop retained each `PCCERT_CONTEXT` returned by `CertFindCertificateInStore`, but that function frees the context passed as `pPrevCertContext` on the next call (and frees the last one when it returns null). The engine therefore stored references it had already released. The memory stays valid because the store keeps the contexts alive, but the destructor then called `CertFreeCertificateContext` on them — a teardown over-release. This is the part the audit mischaracterized as an active use-after-free; in practice the credentials handed to Schannel are always live.

## Verification

Closing each store in the engine destructor with `CERT_CLOSE_STORE_CHECK_FLAG` returned `CRYPT_E_PENDING_CLOSE` (`0x8009200f`, outstanding contexts) on essentially every SSL connection in the `IceSSL/configuration` test — despite the test passing. After the fix, every store closes clean and all tests still pass.

## Fix (see #5494)

- Remove the discarded `CertDuplicateCertificateContext` loops in both credential callbacks.
- Duplicate each context in `findCertificates` so the engine owns the references it stores (balancing the destructor's free, and removing the teardown over-release).
- Add a debug-only `CERT_CLOSE_STORE_CHECK_FLAG` + `assert` in the engine destructor to catch certificate-reference leaks early.

## Platform

Windows (Schannel) only.

pepone:
Fixes #5520.

This PR corrects certificate-context reference-count handling in the Schannel SSL engine. Note: this is **not** a use-after-free — the certificates handed to Schannel as credentials stay valid because the certificate store keeps them alive for the engine's lifetime. The real defects are reference-count errors:

1. **Credential-callback leak.** `createClientAuthenticationOptions` / `createServerAuthenticationOptions` bumped each certificate's reference count via `CertDuplicateCertificateContext` and discarded the result. The transceiver already takes its own reference (freed in `close()`), so the engine's bump leaked one `CERT_CONTEXT` reference per certificate per connection and kept the certificate store from closing cleanly. Removed the discarded duplicate loops.

2. **`findCertificates` ownership.** The `IceSSL.FindCert` enumeration loop retained each `PCCERT_CONTEXT`, but `CertFindCertificateInStore` frees the context passed as `pPrevCertContext` on the next call, so the engine stored references it had already released — leading to a teardown over-release in the destructor. Duplicate each context so the engine owns the references it stores, mirroring the file-based certificate path.

3. **Regression guard.** Added a debug-only `CERT_CLOSE_STORE_CHECK_FLAG` + `assert` in the engine destructor to catch certificate-reference leaks early.

Verified with the `IceSSL/configuration` test: before the fix, closing the store with `CERT_CLOSE_STORE_CHECK_FLAG` reported `CRYPT_E_PENDING_CLOSE` (outstanding contexts) on essentially every connection; after the fix every store closes clean and all tests pass.

### DIFF
--- a/cpp/src/Ice/SSL/SchannelEngine.cpp
+++ b/cpp/src/Ice/SSL/SchannelEngine.cpp
@@ -788,7 +788,15 @@ Schannel::SSLEngine::~SSLEngine()
 
         for (vector<HCERTSTORE>::const_iterator i = _stores.begin(); i != _stores.end(); ++i)
         {
+#ifdef NDEBUG
             CertCloseStore(*i, 0);
+#else
+            // In debug builds, close the store with CERT_CLOSE_STORE_CHECK_FLAG to assert that every certificate
+            // context obtained from it was freed. The call returns false (last error CRYPT_E_PENDING_CLOSE) if any
+            // context is still outstanding, which catches certificate-reference leaks early.
+            const BOOL allContextsFreed = CertCloseStore(*i, CERT_CLOSE_STORE_CHECK_FLAG);
+            assert(allContextsFreed && "SSL transport: certificate context leaked (store closed with live contexts)");
+#endif
         }
     }
     catch (...)
@@ -1270,11 +1278,9 @@ Schannel::SSLEngine::createClientAuthenticationOptions(const string& host) const
         .clientCredentialsSelectionCallback =
             [this](const string&)
         {
-            for (const auto& cert : _allCerts)
-            {
-                CertDuplicateCertificateContext(cert);
-            }
-
+            // The caller (TransceiverI) duplicates each paCred context into its own list and frees those in close(),
+            // so the credentials we return must not bump the reference count here — doing so leaks one reference per
+            // certificate per connection and prevents the certificate store from ever closing cleanly.
             return SCH_CREDENTIALS{
                 .dwVersion = SCH_CREDENTIALS_VERSION,
                 .cCreds = static_cast<DWORD>(_allCerts.size()),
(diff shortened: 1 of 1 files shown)


cpp/src/Ice/SSL/SchannelEngine.cpp
```diff
@@ -570,7 +570,10 @@ namespace
                          0,
                          next)) != 0)
                 {
```
bernardnormier, line 572: The check above should be `!= nullptr` or nothing, not `!= 0`.

bernardnormier (approved):

cpp/src/Ice/SSL/SchannelEngine.cpp
pepone: Fixed in 49f6199 — changed the check to `!= nullptr`.

PR #5558, opened 2026-06-12
Fix slice2cs --icerpc bugs: marshal vs declaration order for parameters

bernardnormier:
See new CHANGELOG entry.


cpp/src/slice2cs/IceRpcVisitors.cpp
```diff
@@ -1430,9 +1429,9 @@ Slice::IceRpc::SkeletonVisitor::writeRequestClass(const InterfaceDefPtr& interfa
         _out << nl << "IceRpc.IncomingRequest request,";
         _out << nl << "global::System.Threading.CancellationToken cancellationToken) =>";
 
-        ParameterList inParameters = operation->inParameters();
+        ParameterList sortedInParameters = operation->sortedInParameters();
```
pepone, line 1433: Maybe add a comment this is sorted in marshaling order?

pepone (approved):

cpp/src/slice2cs/IceRpcVisitors.cpp
bernardnormier: Done

PR #5560, opened 2026-06-12
Reject invalid negative sizes in InputStream

bernardnormier:
This PR adds extra validation to InputStream in Swift.

I think the fix is very minor and not worthy of a CHANGELOG entry.

### DIFF
--- a/swift/src/Ice/InputStream.swift
+++ b/swift/src/Ice/InputStream.swift
@@ -231,7 +231,11 @@ public final class InputStream {
         case .VSize:
             try skip(readSize())
         case .FSize:
-            try skip(read())
+            let sz: Int32 = try read()
+            if sz < 0 {
+                throw MarshalException("invalid negative size for optional field")
+            }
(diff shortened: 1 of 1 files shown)


swift/src/Ice/InputStream.swift
```diff
@@ -291,6 +295,7 @@ public final class InputStream {
     ///
     /// - Parameter count: The number of bytes to skip.
     public func skip(_ count: Int32) throws {
+        precondition(count >= 0, "skip count is negative")
```
pepone, line 298: I feel that an exception would be more appropriate here, the current callers are safe but if we miss one this tears the process down.

pepone (approved):

swift/src/Ice/InputStream.swift
bernardnormier: We use preconditions pretty much everywhere in this class, including the other skip overload. If we switch to exceptions, we should do it everywhere.

PR #5371, opened 2026-05-29
Remove cherry-picked entries from the Ice 3.9.0 changelog

pepone:
## Description

Every entry currently in the **Ice 3.9.0** changelog describes a change that was cherry-picked into the `3.8` branch. Those entries are now recorded in the **Ice 3.8.2** section of `CHANGELOG-3.8.md` (companion PR #5370 on the `3.8` branch).

Following the policy of keeping in `main` only entries that were *not* cherry-picked, this PR removes all of those entries and strips the 3.9.0 section back to its heading. The section will be repopulated as 3.9-only changes land.

Companion PR (move into 3.8.2): #5370

PR #5561, opened 2026-06-12
Fix inverted null check in IcePy adapterFind

### LINKED ISSUE #5521: [Audit-High] IcePy adapterFind has an inverted null check — adapter.find() crashes on an unregistered identity and returns None for a registered one

## Summary

`adapterFind` in the IcePy extension has its null check **inverted** relative to every sibling method in the same file. As written, `adapter.find(identity)` returns `None` when a servant **was** found, and for an unregistered identity it `dynamic_pointer_cast`s a null `shared_ptr` — an assertion failure in debug builds and a null-pointer call (`wrapper->getObject()`) in release builds.

## The bug

```cpp
// python/modules/IcePy/ObjectAdapter.cpp:912-921
obj = (*self->adapter)->find(ident);
...
if (!obj)                               // inverted
{
    auto wrapper = dynamic_pointer_cast<ServantWrapper>(obj);   // obj is null here
    assert(wrapper);                    // asserts (debug) / null deref (release)
    return wrapper->getObject();
}
else
{
    return Py_None;                     // returned when a servant WAS found
}
```

The siblings `adapterFindFacet`, `adapterFindByProxy`, and `adapterFindDefaultServant` all use the correct `if (obj)`.

## Impact

`ObjectAdapter.find()` is entirely broken in the Python mapping: it can never return the servant, and calling it with an unregistered identity crashes the process in release builds.

## Source references

- **(inverted check)** — [`ObjectAdapter.cpp#L912-L921`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/ObjectAdapter.cpp#L912-L921)
- **(correct sibling)** — [`ObjectAdapter.cpp#L959-L967`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/ObjectAdapter.cpp#L959-L967)

## Recommendation

`if (obj)`, mirroring the siblings.

## Severity

**High** — a public API method that both never works and crashes on the common miss case; one-token fix. Verified **CONFIRMED** by the second-model pass; ranked first in both priority lists.

pepone:
`adapterFind` in the IcePy extension had its null check inverted relative to its siblings (`adapterFindFacet`, `adapterFindByProxy`, `adapterFindDefaultServant`): `ObjectAdapter.find()` returned `None` when a servant was registered, and asserted (debug) / dereferenced a null pointer (release) for an unregistered identity.

This PR flips the check to `if (obj)`, matching the siblings, and adds a regression test to the `Ice/defaultServant` suite covering `find`/`findFacet` for both registered and unregistered identities. The test runs before any default servant is registered, since `findServant` falls back to default servants.

Fixes #5521

### DIFF
--- a/python/test/Ice/defaultServant/AllTests.py
+++ b/python/test/Ice/defaultServant/AllTests.py
@@ -15,6 +15,31 @@ def allTests(helper: TestHelper, communicator: Ice.Communicator):
 
     servant = MyObjectI.MyObjectI()
 
+    # Test find/findFacet before any default servant is registered, since findServant
+    # falls back to default servants for identities not in the active servant map.
+    sys.stdout.write("testing find and findFacet... ")
+    sys.stdout.flush()
+
+    identity = Ice.Identity()
+    identity.name = "registered"
+
+    oa.add(servant, identity)
+    test(oa.find(identity) == servant)
+
+    missing = Ice.Identity()
+    missing.name = "unregistered"
+    test(oa.find(missing) is None)
+
+    oa.addFacet(servant, identity, "theFacet")
+    test(oa.findFacet(identity, "theFacet") == servant)
+    test(oa.findFacet(identity, "missingFacet") is None)
+
+    oa.removeFacet(identity, "theFacet")
+    oa.remove(identity)
+    test(oa.find(identity) is None)
+
+    print("ok")
+
     # Register default servant with category "foo"
     oa.addDefaultServant(servant, "foo")
 
--- a/python/test/Ice/defaultServant/AllTests.py
+++ b/python/test/Ice/defaultServant/AllTests.py
@@ -15,8 +15,6 @@ def allTests(helper: TestHelper, communicator: Ice.Communicator):
 
     servant = MyObjectI.MyObjectI()
 
-    # Test find/findFacet before any default servant is registered, since findServant
-    # falls back to default servants for identities not in the active servant map.
     sys.stdout.write("testing find and findFacet... ")
     sys.stdout.flush()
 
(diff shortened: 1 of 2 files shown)


pepone:
Thanks for the review. Addressing both comments:

**1. `findFacet` on an unregistered identity (Copilot @ AllTests.py:35) — applied.** Added an assertion exercising `findFacet` against the existing unregistered `missing` identity, so the find/findFacet block now covers a missing identity for facets too (it previously only covered a missing *facet* on a registered identity):

```python
test(oa.findFacet(missing, "theFacet") is None)
```

**2. CHANGELOG.md entry (Copilot @ ObjectAdapter.cpp:916) — not applicable here.** This is a 3.8.3-milestone, backport-bound fix. Per commit `2b0f57c447` (#5371, "Remove cherry-picked entries from the Ice 3.9.0 changelog"), main's `CHANGELOG.md` holds only 3.9.0-exclusive entries; backported fixes get their changelog entry in `CHANGELOG-3.8.md` on the `3.8` branch at backport time, not in main's 3.9.0 changelog. So no entry is added here.

Verification: `ruff check`/`ruff format --check` clean, `pyright` reports 0 errors, and `Ice/defaultServant` passes. An independent review (codex) found no further issues with the fix or test coverage.

python/test/Ice/defaultServant/AllTests.py
```diff
@@ -15,6 +15,32 @@ def allTests(helper: TestHelper, communicator: Ice.Communicator):
 
     servant = MyObjectI.MyObjectI()
 
+    # Test find/findFacet before any default servant is registered, since findServant
```
bernardnormier: I would remove the comment about default servants - more confusing than anything else.

bernardnormier (approved):

python/test/Ice/defaultServant/AllTests.py
pepone: Done in 9cf7a43 — removed the comment.

PR #5562, opened 2026-06-12
Fix Ice.Future wait bugs: timeout=0 and sent() on never-sent invocations

### LINKED ISSUE #5523: [Audit-Medium] Python Future._wait treats timeout=0 as wait-forever — result(0)/exception(0) block indefinitely

## Summary

`Ice.Future._wait` distinguishes bounded from unbounded waiting with `if timeout:` — which is false for `0`/`0.0`. A caller using the standard "poll, return immediately" idiom `fut.result(0)` or `fut.exception(0)` therefore falls into the `else` branch and blocks **indefinitely** on the condition variable instead of raising `TimeoutException` at once.

## The bug

```python
# python/python/Ice/Future.py:244-258
def _wait(self, timeout: float | None, testFn: Callable[[], bool]) -> bool:
    while testFn():
        if timeout:                       # 0 / 0.0 is falsy → wrong branch
            start = time.time()
            self._condition.wait(timeout)
            timeout -= time.time() - start
            if timeout <= 0:
                return False
        else:
            self._condition.wait()        # blocks indefinitely
    return True
```

`result()`, `exception()`, and `InvocationFuture.sent()` forward their documented maximum-wait `timeout` straight into `_wait`.

## Impact

`fut.result(0)` / `fut.exception(0)` on a still-running future hangs forever, contradicting the documented maximum-wait contract.

## Source references

- **(falsy-zero branch)** — [`Future.py#L244-L258`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/python/Ice/Future.py#L244-L258)
- **(public callers forwarding the timeout)** — [`Future.py#L115-L180`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/python/Ice/Future.py#L115-L180)

## Recommendation

`if timeout is not None:`.

## Severity

**Medium** — a documented, common timeout value silently becomes an unbounded block. Verified **CONFIRMED** by the second-model pass. The same audit found a second, independent wait bug in `InvocationFuture.sent()` (filed separately).

### LINKED ISSUE #5524: [Audit-Medium] Python InvocationFuture.sent() hangs when the invocation is cancelled or fails before being sent — documented InvocationCanceledException path is dead

## Summary

`InvocationFuture.sent()` waits with the predicate `lambda: not self._sent`. When an invocation is **cancelled or fails before it is ever sent**, `Future.cancel()` / `set_exception()` flip the state and `notify_all()` — but never set `_sent` — so the predicate stays true and `_wait` simply waits again. With `timeout=None` the call blocks forever; with a finite timeout it raises `TimeoutException`. The documented `InvocationCanceledException` / propagated-exception branches just below the wait are unreachable for the never-sent case.

## The bug

```python
# python/python/Ice/InvocationFuture.py:125-134
def sent(self, timeout=None) -> bool:
    with self._condition:
        if not self._wait(timeout, lambda: not self._sent):   # only checks _sent
            raise TimeoutException()
        if self._state == Future.StateCancelled:
            raise InvocationCanceledException()                # dead for never-sent futures
        elif self._exception:
            raise self._exception                              # dead for never-sent futures
        else:
            return self._sentSynchronously
```

## Impact

Issue an AMI call, `cancel()` it before it reaches the wire (or have the connection fail before send), then call `.sent()`: the docstring promises `InvocationCanceledException`; instead the call hangs. Per the verification pass, the existing AMI test demonstrates the reachable terminal state `not is_sent() and done()` after cancellation ([`python/test/Ice/ami/AllTests.py#L774-L797`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/test/Ice/ami/AllTests.py#L774-L797)).

## Source references

- **(`sent` predicate and dead branches)** — [`InvocationFuture.py#L101-L134`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/python/Ice/InvocationFuture.py#L101-L134)
- **(cancel/set_exception don't set `_sent`)** — [`Future.py#L34-L58`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/python/Ice/Future.py#L34-L58), [`Future.py#L213-L236`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/python/Ice/Future.py#L213-L236)

## Recommendation

Terminate the predicate once the future is done, e.g. `lambda: not self._sent and self._state == Future.StateRunning`.

## Severity

**Medium** — a documented exception path is dead and the call hangs instead. Verified **CONFIRMED** by the second-model pass. The same audit found a second, independent wait bug in `Future._wait(0)` (filed separately).

pepone:
Two independent wait bugs in the Python `Future` classes:

- `Future._wait` distinguished bounded from unbounded waiting with `if timeout:`, which is false for `0`/`0.0`. The standard polling idiom `future.result(0)` / `future.exception(0)` therefore blocked indefinitely instead of raising `TimeoutException` immediately. Fixed with `if timeout is not None:`.

- `InvocationFuture.sent()` waited on the predicate `lambda: not self._sent`. When an invocation is cancelled (or fails) before ever being sent, `cancel()`/`set_exception()` flip the state and notify, but never set `_sent`, so `sent()` kept waiting forever and the documented `InvocationCanceledException` path was unreachable. The predicate now also terminates when the future leaves the running state.

Regression tests added to the `Ice/ami` suite: a serverless "future timeout" section covering `result(0)`/`exception(0)`/expired timeouts, and `sent()` checks on the existing cancel-before-send futures (which previously hung).

Fixes #5523
Fixes #5524

### DIFF
--- a/python/python/Ice/InvocationFuture.py
+++ b/python/python/Ice/InvocationFuture.py
@@ -123,7 +123,9 @@ class InvocationFuture(Future):
             If the invocation was cancelled before it was sent.
         """
         with self._condition:
-            if not self._wait(timeout, lambda: not self._sent):
+            # Stop waiting once the invocation is sent, or once it completes without
+            # ever being sent (cancelled or failed before being sent).
+            if not self._wait(timeout, lambda: not self._sent and self._state == Future.StateRunning):
                 raise TimeoutException()
 
             if self._state == Future.StateCancelled:
--- a/python/python/Ice/InvocationFuture.py
+++ b/python/python/Ice/InvocationFuture.py
@@ -109,6 +109,8 @@ class InvocationFuture(Future):
         timeout : int | float | None, optional
             Maximum time (in seconds) to wait for the invocation to be sent.
             If ``None`` (the default), this function waits indefinitely.
+            A timeout of ``0`` returns immediately (a non-blocking poll), raising :class:`Ice.TimeoutException`
+            if the invocation has not yet been sent.
 
         Returns
         -------
--- a/python/test/Ice/ami/AllTests.py
+++ b/python/test/Ice/ami/AllTests.py
@@ -262,6 +262,34 @@ def allTestsFuture(helper: TestHelper, communicator: Ice.Communicator, collocate
 
     print("ok")
 
+    sys.stdout.write("testing future timeout... ")
+    sys.stdout.flush()
+
+    # result/exception with a zero or expired timeout must raise TimeoutException
+    # instead of waiting indefinitely.
+    runningFuture = Ice.Future()
+    try:
+        runningFuture.result(0)
+        test(False)
+    except Ice.TimeoutException:
+        pass
+    try:
+        runningFuture.exception(0)
+        test(False)
+    except Ice.TimeoutException:
+        pass
+    try:
+        runningFuture.result(0.01)
+        test(False)
+    except Ice.TimeoutException:
+        pass
+
+    runningFuture.set_result(15)
+    test(runningFuture.result(0) == 15)
+    test(runningFuture.exception(0) is None)
+
+    print("ok")
+
     sys.stdout.write("testing done callback... ")
     sys.stdout.flush()
 
@@ -795,6 +823,18 @@ def allTestsFuture(helper: TestHelper, communicator: Ice.Communicator, collocate
         p.ice_ping()
         test(not f1.is_sent() and f1.done())
         test(not f2.is_sent() and f2.done())
+
+        # sent() must not block once the invocation is cancelled before being sent.
+        try:
+            f1.sent(10)
+            test(False)
+        except Ice.InvocationCanceledException:
+            pass
+        try:
+            f2.sent(10)
+            test(False)
+        except Ice.InvocationCanceledException:
+            pass
 
         testController.holdAdapter()
 
--- a/python/test/Ice/ami/AllTests.py
+++ b/python/test/Ice/ami/AllTests.py
@@ -272,16 +272,19 @@ def allTestsFuture(helper: TestHelper, communicator: Ice.Communicator, collocate
         runningFuture.result(0)
         test(False)
     except Ice.TimeoutException:
+        # Expected: the future is still running, so the bounded wait must time out.
         pass
     try:
         runningFuture.exception(0)
         test(False)
     except Ice.TimeoutException:
+        # Expected: the future is still running, so the bounded wait must time out.
         pass
     try:
         runningFuture.result(0.01)
         test(False)
     except Ice.TimeoutException:
+        # Expected: the future is still running, so the bounded wait must time out.
         pass
 
     runningFuture.set_result(15)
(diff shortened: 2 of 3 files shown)


python/python/Ice/InvocationFuture.py
```diff
@@ -123,7 +123,9 @@ def sent(self, timeout: int | float | None = None) -> bool:
             If the invocation was cancelled before it was sent.
         """
         with self._condition:
-            if not self._wait(timeout, lambda: not self._sent):
+            # Stop waiting once the invocation is sent, or once it completes without
```
pepone, line 128: We should consider updating the doc comments in this file to mention `0` as it is kind of an special case.

pepone:

**Applied**
- *Empty `except` blocks (lines 274/279/284/831/836):* added a brief `# Expected: …` comment to each, matching the convention used elsewhere in the test suite. No control-flow change.
- *`sent(10)` is slow to fail if `sent()` regresses:* lowered both to `sent(2)`. The future is already cancelled at that point, so with the fix `sent()` returns immediately regardless; the smaller bound just makes a hypothetical regression fail faster.

- *Replace `Ice.Future()` with `concurrent.futures.Future()` (github-code-quality "non-callable"):* this is a false positive — `Ice.Future` is a real, callable class (`python/python/Ice/Future.py`); the analyzer just can't resolve the Ice extension module. The whole point of this block is to verify *Ice.Future's* timeout semantics (issue #5523), which differ from the stdlib future, so switching types would test the wrong thing.
- *Run `result(0)`/`exception(0)` in daemon threads so a regression fails fast:* declined. The fix is in place and asserted; a regression would still be caught (as a job timeout). Wrapping these polls in watchdog threads adds disproportionate complexity and would be inconsistent with the rest of this integration-style suite, which would also hang on regressions.
- *Add a CHANGELOG.md entry:* this is a 3.8.3-milestone, backport-bound fix. Per the project convention (commit `2b0f57c447` / #5371 removed cherry-picked entries from the 3.9.0 changelog, and main's CHANGELOG.md currently holds only 3.9-exclusive entries), backported fixes get their changelog entry in `CHANGELOG-3.8.md` on the `3.8` branch at backport time, not in main's 3.9.0 changelog.

python/python/Ice/InvocationFuture.py
pepone: Good call. The `timeout` param docstrings for `sent`, `result`, and `exception` now document that `timeout=0` is a non-blocking poll: it returns immediately, raising `TimeoutException` if the result/sent-state is not yet available. The existing `None` = wait-indefinitely wording is preserved.

python/test/Ice/ami/AllTests.py
```diff
@@ -262,6 +262,34 @@ def allTestsFuture(helper: TestHelper, communicator: Ice.Communicator, collocate
 
     print("ok")
 
+    sys.stdout.write("testing future timeout... ")
+    sys.stdout.flush()
+
+    # result/exception with a zero or expired timeout must raise TimeoutException
+    # instead of waiting indefinitely.
+    runningFuture = Ice.Future()
```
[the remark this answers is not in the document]
bernardnormier: I don't understand this comment. This code is about testing Ice.Future.

bernardnormier (approved):

PR #5563, opened 2026-06-12
Fix crash in IcePy ice_invokeAsync when the context argument is omitted

### LINKED ISSUE #5525: [Audit-Medium] IcePy async Blobject converts an omitted context with dictionaryToContext(nullptr) — assert/null deref on ice_invokeAsync without a context

## Summary

In IcePy's **async** Blobject invocation, the optional context argument is parsed with `"|O"` and initialized to `nullptr`, so it is null when the caller omits it. The conversion guard `if (!ctx || ctx != Py_None)` is true in exactly that case and calls `dictionaryToContext(nullptr, context)` — which begins with `assert(PyDict_Check(dict))` and dereferences the pointer: assertion failure in debug, null deref in release. The **synchronous** Blobject path has the correct guard, and even the *later* use in the same async function is correct — only the conversion guard is wrong, confirming a typo.

## The bug

```cpp
// python/modules/IcePy/Operation.cpp:1823-1830 (async path)
Ice::Context context;
if (!ctx || ctx != Py_None)        // true when ctx omitted (nullptr)
{
    if (!dictionaryToContext(ctx, context)) { return nullptr; }   // dictionaryToContext asserts PyDict_Check(dict)
}

// :1740-1747 (sync path — correct)
if (ctx != nullptr && ctx != Py_None)
{
    if (!dictionaryToContext(ctx, context)) { return nullptr; }
}
```

## Impact

`proxy.ice_invokeAsync(operation, mode, inParams)` with the optional 4th context argument omitted crashes (reachable via the low-level blobject API; typed generated stubs always pass a context, which is why the common path doesn't hit it).

## Source references

- **(wrong async guard; optional `"|O"` parse)** — [`Operation.cpp#L1797-L1830`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Operation.cpp#L1797-L1830)
- **(correct sync guard)** — [`Operation.cpp#L1740-L1747`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Operation.cpp#L1740-L1747)
- **(`dictionaryToContext` asserts a dict)** — [`Util.cpp#L438-L446`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Util.cpp#L438-L446)

## Recommendation

`if (ctx && ctx != Py_None)`, matching the sync path.

## Severity

**Medium** — crash on a legal public-API call shape; one-token fix. Verified **CONFIRMED** by the second-model pass.

pepone:
In the async Blobject invocation path (`AsyncBlobjectInvocation::handleInvoke`), the optional context argument is parsed with `"|O"` and is null when the caller omits it. The conversion guard `if (!ctx || ctx != Py_None)` was true in exactly that case and called `dictionaryToContext(nullptr, ...)` — an assertion failure in debug builds and a segfault in release builds. The fix uses the same guard as the synchronous path: `if (ctx && ctx != Py_None)`.

Regression test: the `Ice/blobject` client now calls `ice_invoke`/`ice_invokeAsync` with the context argument omitted (the async call previously crashed the client with SIGSEGV).

Fixes #5525


pepone:
## Review responses

### Pyright CI failure — fixed
The `Pyright` check was failing with `reportCallIssue` on `test/Ice/blobject/Client.py` because the type stub `python/python/IcePy-stubs/__init__.pyi` declared `ice_invoke`/`ice_invokeAsync` with `ctx: dict[str, str] | None` and **no default**, so omitting `ctx` was a type error — even though the runtime treats it as optional, which is precisely the behavior this PR fixes. I gave `ctx` a default of `None` in both signatures so the stub now matches the runtime. I also narrowed the `ice_invokeAsync` result to `Ice.Future` in the regression test (via `assert isinstance(f, Ice.Future)`, matching the existing pattern in `RouterI.py`) so Pyright accepts `.result()`. Pyright now reports 0 errors and the blobject suite passes.

### Copilot @ Operation.cpp:1826 — add a CHANGELOG.md entry → not applicable
This is a 3.8.3-milestone, backport-bound fix. Per project convention — see commit `2b0f57c447` ("Remove cherry-picked entries from the Ice 3.9.0 changelog", #5371) and the fact that main's `CHANGELOG.md` currently contains only 3.9-exclusive entries — backported fixes get their changelog entry in `CHANGELOG-3.8.md` on the `3.8` branch at backport time, not in main's 3.9.0 changelog. No changelog entry is added here.

bernardnormier (approved):

PR #5564, opened 2026-06-12
Fix IcePy marshal error paths that continued with a pending Python exception

### LINKED ISSUE #5522: [Audit-Medium] IcePy sequence-marshal error paths return instead of throwing AbortMarshaling — corrupt/truncated message with a pending Python exception

## Summary

Three error paths in IcePy's sequence marshaling `return;` with a Python exception pending instead of throwing `AbortMarshaling()`. The marshaling driver in `Operation.cpp` detects per-parameter marshal failure **only** by catching `AbortMarshaling`, so a plain return lets the caller treat the parameter as marshaled and keep writing — producing a structurally corrupt or truncated encapsulation, plus a stale Python exception that surfaces later. For an optional sequence the `writeOptional` tag was already emitted, so the stream is definitively inconsistent.

## The bug

```cpp
// python/modules/IcePy/Types.cpp

// SequenceInfo::marshal, size-determination branch (:1288-1292)
if (!fs.get()) { assert(PyErr_Occurred()); return; }

// SequenceInfo::marshal, generic element branch (:1312-1316)
PyObjectHandle fastSeq{PySequence_Fast(p, "expected a sequence value")};
if (!fastSeq.get()) { return; }

// SequenceInfo::marshalPrimitiveSequence (:1568-1573)
PyObjectHandle fs{getSequence(pi, p)};
if (!fs.get()) { assert(PyErr_Occurred()); return; }
```

The null-**item** path in the same function does it correctly (`:1322-1327`): `throw AbortMarshaling();`.

## Impact

Passing a non-sequence (an `int`, or an object whose `__iter__` raises) for a `sequence<T>` parameter or return value silently corrupts the outgoing message instead of aborting the invocation. Per the verification pass: `prepareRequest` handles Python marshal failure only in its `AbortMarshaling` catch, so the encapsulation can be completed and — for async invocation — **submitted to the wire** before the later pending-error check fires.

## Source references

- **(the three `return` sites)** — [`Types.cpp#L1288-L1292`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L1288-L1292), [`#L1312-L1316`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L1312-L1316), [`#L1568-L1573`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L1568-L1573)
- **(correct sibling in the same function)** — [`Types.cpp#L1322-L1327`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L1322-L1327)
- **(driver catches only `AbortMarshaling`)** — [`Operation.cpp#L978-L1023`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Operation.cpp#L978-L1023)

## Recommendation

`throw AbortMarshaling();` (with the Python exception left set) at the three sites, matching the null-item path.

## Severity

**Medium** — a user type error produces a corrupt wire message and a deferred, misattributed exception instead of a clean failure. Verified **CONFIRMED** by the second-model pass.

### LINKED ISSUE #5527: [Audit-Low, grouped] IcePy "!PyObject_IsInstance" swallows the -1 error result at three sites in Types.cpp
> **AI-generated audit finding** — these are low-severity items from an automated **correctness** audit of the Python source (IcePy C++ extension + Python runtime) performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). They have **not been triaged by a human yet**; verify each before acting on it. Grouped: the same error-sentinel idiom bug at three sites in `Types.cpp` (the third was found by the verification pass). Links pinned to commit `e248644d788c7e76e14af8676529201c54c1735d`.

`PyObject_IsInstance` returns `1`, `0`, or **`-1` on error** (e.g. a metaclass `__instancecheck__` that raises). `!(-1)` is false, so at these three sites an error result skips the failure branch and the code **proceeds with a Python exception already set**. The correct `!= 1` idiom appears in the same file (`StructInfo::marshal` [`Types.cpp#L1104-L1107`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L1104-L1107)).

### 1. `ValueInfo::marshal`

[`Types.cpp#L2544-L2548`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L2544-L2548) — on `-1`, marshaling of the value proceeds with a pending exception instead of aborting.

### 2. `ReadValueCallback::invoke`

[`Types.cpp#L2994-L3005`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L2994-L3005) — on `-1`, the type-mismatch `MarshalException` is skipped and `_cb->unmarshaled(obj, ...)` runs with a pending exception.

### 3. `ExceptionInfo::marshal` (new — found by the verification pass)

[`Types.cpp#L3015-L3022`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L3015-L3022) — same form; a `-1` result proceeds into exception marshaling with a pending Python exception.

**Trigger:** a user class with a custom `__instancecheck__`/metaclass that raises during the check (rare). **Fix:** `if (PyObject_IsInstance(...) != 1)` at all three sites.

Severity: **Low** for all three (rare trigger; the consequence is continuing with a pending exception, which surfaces later as a misattributed error). Sites 1–2 verified **CONFIRMED**; site 3 contributed by the verification pass.

pepone:
Two groups of error-path fixes in `python/modules/IcePy/Types.cpp`:

- Three sequence-marshal error paths (`SequenceInfo::marshal` size-determination branch, its generic-element branch, and `SequenceInfo::marshalPrimitiveSequence`) returned with a Python exception pending instead of throwing `AbortMarshaling`. The marshaling driver detects per-parameter failure only by catching `AbortMarshaling`, so passing a non-sequence for a `sequence<T>` parameter produced a corrupt/truncated message that could reach the wire, with the original `TypeError` surfacing later misattributed to an unrelated call. They now throw `AbortMarshaling`, matching the null-item path in the same function.

- Three `if (!PyObject_IsInstance(...))` checks (`ValueInfo::marshal`, `ReadValueCallback::invoke`, `ExceptionInfo::marshal`) swallowed the `-1` error result and proceeded with a pending exception when a custom `__instancecheck__` raises. They now use the `!= 1` idiom already used by `StructInfo::marshal`.

Regression tests in the `Ice/operations` suite pass non-sequence values for sequence parameters (sync and async) and verify a clean `TypeError` plus a still-usable connection. Pre-fix, the corrupt message reached the server and the stale exception broke subsequent exception construction (`unable to create Python exception class for type ID ::Ice::UnknownLocalException`).

Fixes #5522
Fixes #5527

### DIFF
--- a/python/modules/IcePy/Types.cpp
+++ b/python/modules/IcePy/Types.cpp
@@ -921,9 +921,9 @@ int32_t
 IcePy::EnumInfo::valueForEnumerator(PyObject* p) const
 {
     int isInstance = PyObject_IsInstance(p, pythonType);
-    if (isInstance < 0)
+    if (isInstance == -1)
     {
-        assert(PyErr_Occurred()); // A custom __instancecheck__ raised; preserve the pending exception.
+        assert(PyErr_Occurred());
         return -1;
     }
     else if (isInstance == 0)
@@ -1113,7 +1113,7 @@ IcePy::StructInfo::marshal(
     const Ice::StringSeq*)
 {
     int isInstance = PyObject_IsInstance(p, pythonType);
-    if (isInstance < 0)
+    if (isInstance == -1)
     {
         assert(PyErr_Occurred());
         throw AbortMarshaling();
@@ -2560,9 +2560,8 @@ IcePy::ValueInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap* objectM
     }
 
     int isInstance = PyObject_IsInstance(p, pythonType);
-    if (isInstance < 0)
+    if (isInstance == -1)
     {
-        // A custom __instancecheck__ raised; propagate the pending Python exception.
         assert(PyErr_Occurred());
         throw AbortMarshaling();
     }
@@ -3022,9 +3021,8 @@ IcePy::ReadValueCallback::invoke(const std::shared_ptr<Ice::Value>& p)
         //
         PyObject* obj = reader->getObject(); // Borrowed reference.
         int isInstance = PyObject_IsInstance(obj, _info->pythonType);
-        if (isInstance < 0)
+        if (isInstance == -1)
         {
-            // A custom __instancecheck__ raised; propagate the pending Python exception.
             assert(PyErr_Occurred());
             throw AbortMarshaling();
         }
@@ -3052,9 +3050,8 @@ void
 IcePy::ExceptionInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap* objectMap)
 {
     int isInstance = PyObject_IsInstance(p, pythonType);
-    if (isInstance < 0)
+    if (isInstance == -1)
     {
-        // A custom __instancecheck__ raised; propagate the pending Python exception.
         assert(PyErr_Occurred());
         throw AbortMarshaling();
     }
--- a/python/modules/IcePy/Types.cpp
+++ b/python/modules/IcePy/Types.cpp
@@ -1293,7 +1293,7 @@ IcePy::SequenceInfo::marshal(
                     if (!fs.get())
                     {
                         assert(PyErr_Occurred());
-                        return;
+                        throw AbortMarshaling();
                     }
                     sz = PySequence_Fast_GET_SIZE(fs.get());
                 }
@@ -1317,7 +1317,8 @@ IcePy::SequenceInfo::marshal(
         PyObjectHandle fastSeq{PySequence_Fast(p, "expected a sequence value")};
         if (!fastSeq.get())
         {
-            return;
+            assert(PyErr_Occurred());
+            throw AbortMarshaling();
         }
 
         Py_ssize_t sz = PySequence_Fast_GET_SIZE(fastSeq.get());
@@ -1574,7 +1575,7 @@ IcePy::SequenceInfo::marshalPrimitiveSequence(const PrimitiveInfoPtr& pi, PyObje
     if (!fs.get())
     {
         assert(PyErr_Occurred());
-        return;
+        throw AbortMarshaling();
     }
 
     Py_ssize_t sz = 0;
@@ -2546,7 +2547,7 @@ IcePy::ValueInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap* objectM
         return;
     }
 
-    if (!PyObject_IsInstance(p, pythonType))
+    if (PyObject_IsInstance(p, pythonType) != 1)
     {
         PyErr_Format(PyExc_ValueError, "expected value of type %s", id.c_str());
         throw AbortMarshaling();
@@ -3001,7 +3002,7 @@ IcePy::ReadValueCallback::invoke(const std::shared_ptr<Ice::Value>& p)
         // Verify that the value's type is compatible with the formal type.
         //
         PyObject* obj = reader->getObject(); // Borrowed reference.
-        if (!PyObject_IsInstance(obj, _info->pythonType))
+        if (PyObject_IsInstance(obj, _info->pythonType) != 1)
         {
             throw MarshalException{
                 __FILE__,
@@ -3024,7 +3025,7 @@ IcePy::ReadValueCallback::invoke(const std::shared_ptr<Ice::Value>& p)
 void
 IcePy::ExceptionInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap* objectMap)
 {
-    if (!PyObject_IsInstance(p, pythonType))
+    if (PyObject_IsInstance(p, pythonType) != 1)
     {
         PyErr_Format(PyExc_ValueError, "expected exception %s", id.c_str());
         throw AbortMarshaling();
--- a/python/modules/IcePy/Types.cpp
+++ b/python/modules/IcePy/Types.cpp
@@ -2547,7 +2547,14 @@ IcePy::ValueInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap* objectM
         return;
     }
 
-    if (PyObject_IsInstance(p, pythonType) != 1)
+    int isInstance = PyObject_IsInstance(p, pythonType);
+    if (isInstance < 0)
+    {
+        // A custom __instancecheck__ raised; propagate the pending Python exception.
+        assert(PyErr_Occurred());
+        throw AbortMarshaling();
+    }
+    else if (isInstance == 0)
     {
         PyErr_Format(PyExc_ValueError, "expected value of type %s", id.c_str());
         throw AbortMarshaling();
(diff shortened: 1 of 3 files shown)


pepone:

**Applied**
- *`PyObject_IsInstance` can return `-1` (exception already set); the `!= 1` check then overwrote it with `PyErr_Format` (flagged at `ValueInfo::marshal`, `ExceptionInfo::marshal`, and `ReadValueCallback::invoke`):* good catch — fixed all three sites to distinguish `-1` from `0`. On `-1` (a custom `__instancecheck__` raised) we now `assert(PyErr_Occurred())` and throw `AbortMarshaling`, propagating the original Python exception intact; only the `0` "not an instance" case formats the type-mismatch `ValueError`/`MarshalException`. (commit `87f2f1d8ca`)
- *Empty `except` blocks (lines 248/254/260/266):* added a brief `# Expected: …` comment to each.

- *Description says sync+async but only `opByteSAsync` is tested:* the async path goes through the same marshaling code as the sync path, so a single async case is sufficient; added a short comment noting this rather than adding redundant permutations.
- *Add a CHANGELOG.md entry:* this is a 3.8.3-milestone, backport-bound fix. Per the project convention (commit `2b0f57c447` / #5371; main's CHANGELOG.md holds only 3.9-exclusive entries), backported fixes get their changelog entry in `CHANGELOG-3.8.md` on the `3.8` branch at backport time, not main's 3.9.0 changelog.

pepone:
Follow-up: codex's review pass found two more `PyObject_IsInstance` checks that overwrote the `-1` error result with a generic `ValueError`:

- `EnumInfo::valueForEnumerator` (returns `-1` on error)
- `StructInfo::marshal` (throws `AbortMarshaling`)

Both now use the same three-way handling as the earlier fixes: `-1` preserves the pending Python exception (so a raising `__instancecheck__` propagates intact), and `0` formats the `expected value of type ...` `ValueError`. Build is clean and the Ice/operations, Ice/objects, Ice/enums, and Ice/exceptions suites pass.

python/modules/IcePy/Types.cpp
```diff
@@ -915,7 +915,13 @@ IcePy::EnumInfo::destroy()
 int32_t
 IcePy::EnumInfo::valueForEnumerator(PyObject* p) const
 {
-    if (PyObject_IsInstance(p, pythonType) != 1)
+    int isInstance = PyObject_IsInstance(p, pythonType);
+    if (isInstance < 0)
+    {
+        assert(PyErr_Occurred()); // A custom __instancecheck__ raised; preserve the pending exception.
```
pepone: I don't see the point of the comment

```suggestion
        assert(PyErr_Occurred());
```

python/modules/IcePy/Types.cpp
```diff
@@ -1101,7 +1107,13 @@ IcePy::StructInfo::marshal(
     bool optional,
     const Ice::StringSeq*)
 {
-    if (PyObject_IsInstance(p, pythonType) != 1)
+    int isInstance = PyObject_IsInstance(p, pythonType);
+    if (isInstance < 0)
```
pepone: I think is clear to check `== -1` which is the documented value for errors.

> On error, returns -1 and sets an exception.

Same for other ocurrences.

See https://docs.python.org/3/c-api/object.html

python/modules/IcePy/Types.cpp
```diff
@@ -2541,7 +2554,14 @@ IcePy::ValueInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap* objectM
         return;
     }
 
-    if (!PyObject_IsInstance(p, pythonType))
+    int isInstance = PyObject_IsInstance(p, pythonType);
+    if (isInstance < 0)
+    {
+        // A custom __instancecheck__ raised; propagate the pending Python exception.
```
pepone: The code is obvious don't need this comment.

python/modules/IcePy/Types.cpp
```diff
@@ -3015,7 +3042,14 @@ IcePy::ReadValueCallback::invoke(const std::shared_ptr<Ice::Value>& p)
 void
 IcePy::ExceptionInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap* objectMap)
 {
-    if (!PyObject_IsInstance(p, pythonType))
+    int isInstance = PyObject_IsInstance(p, pythonType);
+    if (isInstance < 0)
+    {
+        // A custom __instancecheck__ raised; propagate the pending Python exception.
```
pepone: ditto

python/modules/IcePy/Types.cpp
pepone: Done in 068de9f — removed the comment; the branch is now just `assert(PyErr_Occurred()); return -1;`.

python/modules/IcePy/Types.cpp
pepone: Done in 068de9f — switched all five `PyObject_IsInstance` error checks from `< 0` to `== -1`.

python/modules/IcePy/Types.cpp
pepone: Done in 068de9f.

python/modules/IcePy/Types.cpp
pepone: Done in 068de9f.

python/modules/IcePy/Types.cpp
```diff
@@ -1288,7 +1288,7 @@ IcePy::SequenceInfo::marshal(
                     if (!fs.get())
                     {
                         assert(PyErr_Occurred());
-                        return;
+                        throw AbortMarshaling();
                     }
```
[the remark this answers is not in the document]
pepone: Added in 068de9f — a "Python Changes" changelog entry for the marshaling fix.

python/modules/IcePy/Types.cpp
```diff
@@ -3015,7 +3016,7 @@ IcePy::ReadValueCallback::invoke(const std::shared_ptr<Ice::Value>& p)
 void
 IcePy::ExceptionInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap* objectMap)
 {
-    if (!PyObject_IsInstance(p, pythonType))
+    if (PyObject_IsInstance(p, pythonType) != 1)
     {
         PyErr_Format(PyExc_ValueError, "expected exception %s", id.c_str());
         throw AbortMarshaling();
```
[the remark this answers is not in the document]
pepone: This is already handled in the current code: the `== -1` branch asserts the pending exception and returns/aborts without constructing anything, and only the `== 0` mismatch case builds the `ValueError` — so the original Python exception is preserved.

python/test/Ice/operations/Twoways.py
```diff
@@ -238,6 +238,37 @@ def twoways(helper: TestHelper, p: Test.MyClassPrx) -> None:
...
+        p.opStringSS([42], [])  # pyright: ignore
+        test(False)
+    except TypeError:
+        pass
+
+    try:
+        p.opByteSAsync(123, 456)  # pyright: ignore
+        test(False)
+    except TypeError:
+        pass
```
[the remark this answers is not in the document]
pepone: The async path runs through the same marshaling code as the sync path, so the single `opByteSAsync` case is enough to cover it — there's a comment in the test noting this. The description's "sync and async" means both paths are exercised, not every permutation in each.

bernardnormier (approved):

PR #5565, opened 2026-06-12
Use numpy.bool_ instead of the removed numpy.bool alias

### LINKED ISSUE #5526: [Audit-Medium] Python _ArrayUtil uses numpy.bool — AttributeError on NumPy 1.24–1.26 breaks the NumPy array mapping

## Summary

`Ice._ArrayUtil` references `numpy.bool` — the deprecated alias that was **removed in NumPy 1.24** (re-added in 2.0). On NumPy 1.24.x–1.26.x the attribute access raises `AttributeError`, which the surrounding `try/except ImportError` does **not** catch, so the module fails to import. NumPy 1.26 is the first affected release that supports Python 3.12 (IcePy's minimum), and the dependency is unconstrained in `pyproject.toml`, so the broken range is reachable.

## The bug

```python
# python/python/Ice/_ArrayUtil.py:29-50
try:
    import numpy
    BuiltinNumpyTypes = [
        numpy.bool,        # removed in NumPy 1.24; re-added in 2.0 — AttributeError, not ImportError
        numpy.int8,
        ...
    ]
except ImportError:
    pass
```

## Impact

With NumPy 1.24–1.26 installed, unmarshaling a sequence mapped to NumPy arrays makes the C extension lazily import `Ice._ArrayUtil.createNumPyArray`; the import fails with `AttributeError: module 'numpy' has no attribute 'bool'`, breaking the NumPy array mapping at first use (not at `import Ice`).

## Source references

- **(the alias use, `ImportError`-only catch)** — [`_ArrayUtil.py#L29-L50`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/python/Ice/_ArrayUtil.py#L29-L50)
- **(lazy import from the extension)** — [`Types.cpp#L2180-L2187`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L2180-L2187)
- **(unconstrained NumPy)** — [`pyproject.toml#L36-L37`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/pyproject.toml#L36-L37)
- **(NumPy 1.24 expired-deprecations notes)** — https://numpy.org/doc/1.24/release/1.24.0-notes.html#expired-deprecations

## Recommendation

Use the stable spelling `numpy.bool_`.

## Severity

**Medium** — the NumPy array mapping is broken on three supported NumPy minor versions, including the only 1.x line that supports Python 3.12. Verified **CONFIRMED** by the second-model pass against the NumPy release notes.

pepone:
`Ice._ArrayUtil` referenced `numpy.bool`, a deprecated alias removed in NumPy 1.24 and only reintroduced in NumPy 2.0. On NumPy 1.24–1.26 (the only 1.x line supporting Python 3.12), the attribute access raises `AttributeError` — which the surrounding `try/except ImportError` does not catch — breaking the NumPy array mapping at first use. The stable spelling `numpy.bool_` works on all supported NumPy versions.

No new test: the breakage only manifests on NumPy 1.24–1.26, and the existing `Ice/custom` suite covers the NumPy array mapping (verified passing with this change).

Fixes #5526

### DIFF
--- a/python/python/Ice/_ArrayUtil.py
+++ b/python/python/Ice/_ArrayUtil.py
@@ -30,7 +30,9 @@ try:
     import numpy
 
     BuiltinNumpyTypes = [
-        numpy.bool,
+        # Use numpy.bool_, not the numpy.bool alias: the alias was removed in NumPy 1.24
+        # and only reintroduced in NumPy 2.0.
+        numpy.bool_,
         numpy.int8,
         numpy.int16,
         numpy.int32,


pepone:
I don't think is worth changing this

- 1.x is out of support already see https://endoflife.date/numpy
- While the change is compatible with 2.x, I think `bool_` alias is intended to avoid conflicts when you import it, which we don't do here. We use `Numpy.bool`.
- Fixing this will give the impression that we still want to support this older releases we don't, we should not introduced fixes for a version that is already out of support.

pepone:
## Follow-up: the NumPy compatibility test had the same defect

A code-review (codex) pass found that the runtime fix alone was incomplete: the NumPy custom test (`Ice/custom`) still referenced the removed `numpy.bool` alias, so the very test meant to exercise NumPy compatibility would itself have raised `AttributeError` on NumPy 1.24–1.26 (where the alias is gone; it was re-added in NumPy 2.0). Notably `Server.py` has no `from __future__ import annotations`, so its `numpy.typing.NDArray[numpy.bool]` annotation was evaluated eagerly at class-definition time.

Root cause: the `slice2py` code generator mapped Slice `bool` sequences to `numpy.bool` in two places in `cpp/src/slice2py/PythonUtil.cpp` (the type-hint table and the dataclass `default_factory` table). Fixed both to emit `numpy.bool_`, and updated the one hand-written override in `python/test/Ice/custom/Server.py`. The `generated/` test files are gitignored build artifacts and now regenerate as `numpy.bool_`.

Verification:
- `grep -rn 'numpy\.bool\b' python/ | grep -v 'numpy\.bool_'` → no matches.
- `ruff check python/test/Ice/custom/ --ignore E402` and `ruff format --check` → clean; pyright clean on changed files.
- `python3 allTests.py --filter=Ice/custom` → passes (incl. the `python:numpy.ndarray` cases). (Local NumPy is 2.x, so the broken 1.24–1.26 range isn't exercised locally; the fix removes the alias dependency.)

## Re: CHANGELOG.md entry (Copilot @ `_ArrayUtil.py:35`)

Respectfully rebutting. This is a 3.8.3-milestone, backport-bound fix. Per commit `2b0f57c447` (#5371, "Remove cherry-picked entries from the Ice 3.9.0 changelog") and the fact that main's `CHANGELOG.md` holds only 3.9.0-exclusive entries, backported fixes get their changelog entry in `CHANGELOG-3.8.md` on the `3.8` branch at backport time — not in main's 3.9.0 changelog. So no CHANGELOG.md change belongs in this PR.

pepone:
Agreed — closing this per your assessment.

Your reasoning holds: the only NumPy versions where `numpy.bool` is missing are 1.24–1.26, which are EOL (NumPy 2.0 re-added the alias), so the change only matters for already-unsupported releases, and we shouldn't introduce fixes for them.

For the record, a review pass found that slice2py also *generates* `numpy.bool` in user code (the `numpy.typing.NDArray[...]` hints and the `numpy.empty(0, numpy.bool)` dataclass factories in `PythonUtil.cpp`), not just the runtime `_ArrayUtil.py` — so the surface was a little broader than the original one-liner. It doesn't change the conclusion, though: every affected NumPy version is out of support, and on supported NumPy (2.x) `numpy.bool` works. Closing; trivially reopenable if you'd rather keep the generator change for tidiness.

PR #5566, opened 2026-06-12
Fix Incorrect Codegen for Java Custom Dictionaries

### LINKED ISSUE #5479: slice2java: hardcoded 'ostr' instead of the custom stream in the in-line dictionary marshal path — non-compiling Java

## Summary

In `slice2java`, `writeDictionaryMarshalUnmarshalCode` computes the stream name from `customStream` and uses it everywhere except the non-optional marshal branch, which hardcodes `ostr`. With a custom stream name (e.g. a class/exception data member's `ostr_`, or a marshaled-result member's `_ostr`), the generated `writeSize` calls reference an undefined `ostr`, so the generated Java does not compile.

## The bug

```cpp
// cpp/src/slice2java/Gen.cpp:504-509  (hardcoded ostr; 'stream' computed at :396-400)
out << nl << "if (" << param << " == null)";
out << sb;
out << nl << "ostr.writeSize(0);";                 // should be << stream <<
out << eb << " else";
out << sb;
out << nl << "ostr.writeSize(" << param << ".size());";   // should be << stream <<
```

The unmarshal branch (`:524`) correctly uses `stream`.

## Impact

A dictionary data member of a class/exception (custom stream `ostr_`) or a marshaled-result member (custom stream `_ostr`) with local `java:type:Instance:Formal` metadata whose formal type differs from the definition's mapping (so the helper check at `:412` is bypassed and the in-line path is taken) generates `_iceWriteImpl(OutputStream ostr_)` that references undefined `ostr` → does not compile.

## Source references

- **(slice2java / hardcoded ostr in dict marshal branch)** — [`cpp/src/slice2java/Gen.cpp#L504-L509`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2java/Gen.cpp#L504-L509)

## Recommendation

Use the computed `stream` variable in the non-optional marshal branch, matching the unmarshal branch.

## Severity

**Medium** — generated Java references an undefined stream symbol and does not compile for custom-stream dictionary members.

InsertCreativityHere:
This fixes #5479, cleans up some adjacent code, and adds tests.

Right now on main, we have 0 tests for using `java:type` on dictionaries.
So instead of just adding a test to cover this specific bug, I had to add full testing for this feature.
I wrote the custom dictionary testing to be at parity with custom sequence testing:
- ~~I added custom dictionaries to the `operations` test, just like we do for sequences~~
- I renamed the current `custom` test to `customSequence` (since that's all it tests).
- I added a dedicated `customDictionary` test which mirrors `customSequence`.


cpp/src/slice2java/Gen.cpp
```diff
@@ -503,10 +503,10 @@ Slice::JavaVisitor::writeDictionaryMarshalUnmarshalCode(
 
             out << nl << "if (" << param << " == null)";
             out << sb;
-            out << nl << "ostr.writeSize(0);";
+            out << nl << stream << ".writeSize(0);";
```
InsertCreativityHere, line 506: This is the actual fix, where we used to hardcode `ostr` instead of using the `stream` variable.

cpp/src/slice2java/Gen.cpp
```diff
@@ -894,7 +894,7 @@ Slice::JavaVisitor::writeResultTypeMarshalUnmarshalCode(
     {
         const bool isOptional = param->isOptional();
         const string name = paramPrefix + param->mappedName();
-        const string patchParams = isMarshaling ? getPatcher(param->type(), package, name) : "";
+        const string patchParams = isMarshaling ? "" : getPatcher(param->type(), package, name);
```
InsertCreativityHere, line 897: Confusingly, we set this `isMarshaling` to `false` when we're `marshaling`.
I swapped every call-site and use of this parameter so now the name is actually correct.

cpp/src/slice2java/Gen.cpp
```diff
@@ -1746,7 +1746,7 @@ Slice::JavaVisitor::writeUnmarshalDataMember(Output& out, const string& package,
     const TypePtr& type = member->type();
     const bool isOptional = member->isOptional();
     const bool forStruct = (bool)dynamic_pointer_cast<Struct>(member->container());
-    const string stream = forStruct ? "istr" : "istr_";
+    const string stream = forStruct ? "" : "istr_";
```
InsertCreativityHere, line 1749: We have code which sets `istr` for the empty string.
Everywhere else we let the function decide on `istr` using this default, except for this one call site.
I just changed this call-site to be consistent with the others.

java/test/src/main/java/test/IceBox/admin/TestServiceI.java
```diff
@@ -14,8 +14,7 @@ public TestServiceI(Communicator serviceManagerCommunicator) {
         // Install a custom admin facet.
         serviceManagerCommunicator.addAdminFacet(facet, "TestFacet");
 
-        // The TestFacetI servant also implements java.util.function.Consumer<java.util.Map<String,
-        // String>>.
+        // The TestFacetI servant also implements java.util.function.Consumer<java.util.Map<String, String>>.
```
InsertCreativityHere, line 17: This easily fits on a single line.
Probably leftover from our previously java formatting passes.

java/test/src/main/java/test/Ice/operations/Twoways.java
```diff
@@ -769,6 +770,141 @@ static void twoways(TestHelper helper, MyClassPrx p) {
...
+            di1.put((byte) 100, Boolean.FALSE);
+            Map<Byte, Boolean> di2 = new LinkedHashMap<>();
+            di2.put((byte) 10, Boolean.TRUE);
+            di2.put((byte) 11, Boolean.FALSE);
+            di2.put((byte) 101, Boolean.TRUE);
+
+            MyClass.OpByteBoolLinkedDResult r = p.opByteBoolLinkedD(di1, di2);
+
+            test(r.p3.equals(di1));
+            test(r.returnValue.size() == 4);
```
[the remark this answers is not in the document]
InsertCreativityHere: I did even better and check the class directly with `getClass()`.
`instanceof` would still compare true since `LinkedHashMap` is a subclass of `HashMap`.

cpp/src/slice2java/Gen.cpp
```diff
@@ -1746,7 +1746,7 @@ Slice::JavaVisitor::writeUnmarshalDataMember(Output& out, const string& package,
...
     assert(!isOptional || !forStruct);           // optional members aren't allowed in structs.
```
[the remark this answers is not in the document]
InsertCreativityHere: No, this works just fine.

java/test/src/main/java/test/Ice/operations/Twoways.java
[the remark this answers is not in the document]
InsertCreativityHere: This is already tested sufficiently by the `customDictionary` test.

Note that we also don't enforce the decoded type for sequences in the `operations` test.
And I think it's correct for testing of sequences & dictionaries to remain at parity.

bernardnormier (approved):
Approving — this is a solid fix.

**Codegen:** The core change (routing the dictionary size-prefix through `stream` instead of the hardcoded `ostr` in `writeDictionaryMarshalUnmarshalCode`) is correct and complete — those were the only two hardcoded stream references in that function. It misfired exactly where the resolved stream isn't `ostr`: a custom `java:type` dictionary used as a class/exception data member (`ostr_`) or in a `["marshaled-result"]` operation (`_ostr`), where the generated Java referenced an undefined `ostr` and failed to compile. The accompanying `isMarshaling` inversion and the struct `"istr"` → `""` change are both behavior-preserving (all three result-type callers flip consistently; an empty stream name resolves to `istr`/`ostr` via the default), so they're pure readability/consistency cleanups.

**Tests:** Genuinely protective rather than just broad. `opMLinkedDict1/2` (marshaled-result ops returning a custom dict) exercise the `_ostr` path, so pre-fix the generated code wouldn't even compile — CI compilation is the regression guard. The new `customDictionary` suite asserts both round-trip equality and the concrete runtime class (`getClass() == LinkedHashMap.class`), which is what actually validates the `java:type` mapping. Rename is clean and the suites are auto-discovered by the directory-walk runner, so both run in CI.

**One request before merge — add a CHANGELOG entry.** This is a user-facing `slice2java` bug (generated Java didn't compile for an affected dictionary), so it warrants an entry. There's no Java section under 3.9.0 yet, so it needs the section plus a TOC line:

In the table of contents:
```markdown
  - [Java Changes](#java-changes)
```

Under `## Changes in Ice 3.9.0`:
```markdown
### Java Changes

- Fixed `slice2java` to generate compilable code for a dictionary with a custom `java:type` when it is used
  as a data member of a class or exception, or as the result of a `["marshaled-result"]` operation. The generated
  marshaling code previously referred to an undefined output stream variable, so the affected Java source failed
  to compile.
```

pepone (approved):

java/test/src/main/java/test/Ice/customSequence/Test.ice
```diff
@@ -2,9 +2,7 @@
 
 #pragma once
 
-#include "Ice/BuiltinSequences.ice"
```
InsertCreativityHere, line 5: This include was unused.

java/test/src/main/java/test/Ice/operations/Test.ice
```diff
@@ -66,13 +66,13 @@ module Test
     dictionary<MyEnum, string> MyEnumStringD;
     dictionary<MyStruct, MyEnum> MyStructMyEnumD;
```
InsertCreativityHere, line 68: We were missing whitespace in some places here. No idea why, but I fixed it.

PR #5567, opened 2026-06-12
Harden IcePy error handling: getVersion, getString, compileSlice, createFuture

### LINKED ISSUE #5528: [Audit-Low] IcePy getVersion minor branch operates on the wrong handle — incomplete copy of the major branch
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Python source (IcePy C++ extension + Python runtime) performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it. Originally graded Medium; **regraded to Low** by the verification pass (valid integer versions still parse correctly).

## Summary

The minor-version branch of IcePy's `getVersion` is an incomplete copy of the (correct) major-version branch: the `PyNumber_Long` coercion result is stored into **`major`**, the null check tests the unchanged **`minor`** handle (always non-null there), and `PyLong_AsLong` reads the original, un-coerced `minor` object.

## The bug

```cpp
// python/modules/IcePy/Util.cpp:57-72
if (minor.get())
{
    major = PyNumber_Long(minor.get());   // coerces minor, stores into `major`
    if (!minor.get())                      // checks the wrong handle — always false here
    {
        PyErr_Format(PyExc_ValueError, "version minor must be a numeric value");
        return false;
    }
    long m = PyLong_AsLong(minor.get());   // reads the ORIGINAL minor, not the coerced value
    ...
    v.minor = static_cast<uint8_t>(m);
}
```

Two consequences: (a) a failed `PyNumber_Long` (NULL + exception set) is not caught and execution continues with a pending exception; (b) the intended numeric coercion is bypassed, so a `minor` that needs coercion reaches `PyLong_AsLong` un-coerced and raises a `TypeError` (a different error than the major field would report).

## Impact

Plain integer `major`/`minor` fields — the overwhelmingly common case — parse correctly, which is why this is Low. The defect affects non-integer numeric-like `minor` values (inconsistent coercion vs. the `major` field, wrong/misattributed error) on paths through `getEncodingVersion`/`getProtocolVersion` (e.g. `proxy.ice_encodingVersion(v)`).

## Source references

- **(the copied branch)** — [`Util.cpp#L57-L72`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Util.cpp#L57-L72)

## Recommendation

```cpp
minor = PyNumber_Long(minor.get());
if (!minor.get()) { PyErr_Format(...); return false; }
long m = PyLong_AsLong(minor.get());
```

## Severity

**Low** (regraded from Medium by the verification pass) — the handle mix-up is real but integer versions are unaffected.

### LINKED ISSUE #5529: [Audit-Low] IcePy getString ignores UTF-8 encoding failure — returns "" with a pending UnicodeEncodeError

## Summary

`IcePy::getString` ignores a `PyUnicode_AsUTF8String` failure: it returns `""` and leaves the `UnicodeEncodeError` pending. Callers (`getStringArg`, `getIdentity`, `dictionaryToContext`, `listToStringSeq`, …) treat the empty string as a valid value and report success, so the operation proceeds with silently wrong data while a stale Python exception surfaces later at an unrelated point.

## The bug

```cpp
// python/modules/IcePy/Util.cpp:101-119
string IcePy::getString(PyObject* p)
{
    assert(p == Py_None || checkString(p));
    string str;
    if (p != Py_None)
    {
        PyObjectHandle bytes{PyUnicode_AsUTF8String(p)};
        if (bytes.get()) { ... str.assign(s, sz); }
        // else: failure silently ignored — returns "" with UnicodeEncodeError pending
    }
    return str;
}
```

## Impact

Passing a string containing a lone surrogate (e.g. `'\ud800'`) as an identity name/category, adapter id, facet, or context key/value is not rejected at the conversion boundary — the value becomes `""` and the invocation proceeds with wrong data plus a deferred, misattributed exception.

## Source references

- **(swallowed failure)** — [`Util.cpp#L101-L119`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Util.cpp#L101-L119)
- **(`getStringArg` returns success regardless)** — [`Util.cpp#L121-L134`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Util.cpp#L121-L134)

## Recommendation

Propagate the failure (return an error indicator from `getString` or check `PyErr_Occurred()` at the call sites) so non-UTF-8-encodable input is rejected at the boundary.

## Severity

**Low** — silent wrong data plus a deferred exception for malformed (lone-surrogate) input. Verified **CONFIRMED** by the second-model pass.

### LINKED ISSUE #5530: [Audit-Low] IcePy_compileSlice accepts an empty argument list — out-of-bounds argSeq[0]/args[0] indexing (UB)
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Python source (IcePy C++ extension + Python runtime) performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it. Verdict **PARTIAL**: the empty-input hazard is real, but the verification pass corrected the primary fault locus (see below).

## Summary

`IcePy_compileSlice` accepts an empty argument list — unlike `IcePy_loadSlice`, it inserts no dummy `argv[0]` and does not validate non-emptiness. With empty input, out-of-bounds `std::vector` indexing (UB) is reachable. Per the verification pass, the **first** out-of-bounds access is `const string programName = args[0]` inside `Slice::Python::compile` itself ([`cpp/src/slice2py/PythonUtil.cpp#L3157-L3160`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2py/PythonUtil.cpp#L3157-L3160)) — reached before the wrapper's catch handlers, which repeat the same unsafe `argSeq[0]` indexing.

## The bug

```cpp
// python/modules/IcePy/Slice.cpp:199-213
try
{
    rc = Slice::Python::compile(argSeq);                      // argSeq may be empty → args[0] OOB inside compile
}
catch (const std::exception& ex)
{
    consoleErr << argSeq[0] << ": error:" << ex.what() << endl;   // also OOB for empty input
    rc = EXIT_FAILURE;
}
catch (...)
{
    consoleErr << argSeq[0] << ": error:unknown exception" << endl;  // also OOB
    rc = EXIT_FAILURE;
}
```

## Impact

`IcePy.compileSlice([])` performs out-of-bounds vector indexing — undefined behavior — instead of failing with a clean Python error. Local API misuse only.

## Source references

- **(no emptiness validation; unsafe catch handlers)** — [`Slice.cpp#L185-L213`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Slice.cpp#L185-L213)
- **(primary OOB locus, corrected by verification)** — [`PythonUtil.cpp#L3157-L3160`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2py/PythonUtil.cpp#L3157-L3160)

## Recommendation

Reject an empty list in `IcePy_compileSlice` with a Python `ValueError` (or insert a dummy program name like `IcePy_loadSlice` does) before calling `compile`.

## Severity

**Low** — UB on a local, degenerate API call. Verified **PARTIAL** by the second-model pass (hazard real; fault locus corrected).

### LINKED ISSUE #5531: [Audit-Low] IcePy createFuture ignores tp_init failure — returns a partially-initialized future with an error set

## Summary

IcePy's `createFuture` calls `type->tp_init(...)` without checking its return value. If `Ice.InvocationFuture.__init__` raises, `createFuture` returns a **non-null but partially-initialized** future with a Python error still set. The caller (`AsyncInvocation::invoke`) only checks for a null pointer, so it proceeds to call `set_sent`/`set_result`/`set_exception` on the half-constructed object while the unrelated `__init__` exception remains pending.

## The bug

```cpp
// python/modules/IcePy/Operation.cpp:2604-2608
// Call the Ice.InvocationFuture.__init__ method.
type->tp_init(future.get(), initArgs.get(), nullptr);   // negative return not checked

// Release the reference to the future object and let the caller take ownership.
return future.release();
```

## Impact

An `__init__` failure (e.g. resource exhaustion; low likelihood) is silently converted into use of an invalid future object, with the original error surfacing later at an unrelated `PyErr_Occurred()` check.

## Source references

- **(unchecked `tp_init`)** — [`Operation.cpp#L2583-L2608`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Operation.cpp#L2583-L2608)
- **(caller checks only for null)** — [`Operation.cpp#L1394-L1398`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Operation.cpp#L1394-L1398)

## Recommendation

Check `tp_init`'s return value and return `nullptr` (with the exception set) on failure.

## Severity

**Low** — rare trigger; the failure mode is use of a partially-initialized object with a deferred error. Verified **CONFIRMED** by the second-model pass.

pepone:
Groups four low-severity error-handling fixes in the IcePy extension, all from the correctness audit:

- **getVersion** (`Util.cpp`): the minor-version branch was an incomplete copy of the major branch — it stored the `PyNumber_Long` coercion into `major`, null-checked the wrong handle, and read the un-coerced value. It now coerces, checks, and reads `minor`. (#5528)

- **getString** (`Util.cpp`, `Types.cpp`): a `PyUnicode_AsUTF8String` failure (e.g. a lone surrogate) silently produced `""` with a pending `UnicodeEncodeError` that surfaced later misattributed (typically as `SystemError: ... returned a result with an exception set`). The boundary callers that take user input — `getStringArg`, `getIdentity`, `listToStringSeq`, `tupleToStringSeq`, `dictionaryToContext`, and the string marshal in `writeString` — now detect the failure and propagate it, so the bad string is rejected at the conversion boundary. (#5529)

- **compileSlice** (`Slice.cpp`): an empty argument list reached `Slice::Python::compile`, which indexes `args[0]` (out-of-bounds, UB), as did the catch handlers. An empty list is now rejected with `ValueError`. (#5530)

- **createFuture** (`Operation.cpp`): the `tp_init` return value was ignored, so an `Ice.InvocationFuture.__init__` failure returned a partially-initialized future with an error set. It now returns `nullptr` on failure. (#5531)

Regression tests: `Ice/operations` rejects lone-surrogate string parameters, context keys, and identity names with `UnicodeEncodeError` (and verifies the connection stays usable); `Slice/import` checks `IcePy.compileSlice([])` raises `ValueError`. No tests for the getVersion/createFuture fixes (contrived triggers).

Fixes #5528
Fixes #5529
Fixes #5530
Fixes #5531

### DIFF
--- a/python/modules/IcePy/Slice.cpp
+++ b/python/modules/IcePy/Slice.cpp
@@ -196,6 +196,13 @@ IcePy_compileSlice(PyObject* /*self*/, PyObject* args)
         return nullptr;
     }
 
+    // Slice::Python::compile and the catch handlers below expect args[0] to be the program name.
+    if (argSeq.empty())
+    {
+        PyErr_Format(PyExc_ValueError, "arguments cannot be empty");
+        return nullptr;
+    }
+
     int rc;
     try
     {
--- a/python/modules/IcePy/Slice.cpp
+++ b/python/modules/IcePy/Slice.cpp
@@ -199,7 +199,9 @@ IcePy_compileSlice(PyObject* /*self*/, PyObject* args)
     // Slice::Python::compile and the catch handlers below expect args[0] to be the program name.
     if (argSeq.empty())
     {
-        PyErr_Format(PyExc_ValueError, "arguments cannot be empty");
+        PyErr_Format(
+            PyExc_ValueError,
+            "compileSlice requires a non-empty argument list whose first element is the program name");
         return nullptr;
     }
 
--- a/python/modules/IcePy/Types.cpp
+++ b/python/modules/IcePy/Types.cpp
@@ -85,7 +85,12 @@ namespace IcePy
         }
         else if (checkString(p))
         {
-            os->write(getString(p), false); // Bypass string conversion.
+            string str = getString(p);
+            if (PyErr_Occurred())
+            {
+                return false; // The string is not UTF-8 encodable.
+            }
+            os->write(str, false); // Bypass string conversion.
         }
         else
         {
--- a/python/modules/IcePy/Types.cpp
+++ b/python/modules/IcePy/Types.cpp
@@ -88,7 +88,7 @@ namespace IcePy
             string str = getString(p);
             if (PyErr_Occurred())
             {
-                return false; // The string is not UTF-8 encodable.
+                return false; // String conversion failed; a Python exception is set.
             }
             os->write(str, false); // Bypass string conversion.
         }
--- a/python/modules/IcePy/Types.cpp
+++ b/python/modules/IcePy/Types.cpp
@@ -2782,6 +2782,10 @@ IcePy::ValueWriter::_iceWrite(Ice::OutputStream* os) const
             throw AbortMarshaling();
         }
         string id = getString(ret.get());
+        if (PyErr_Occurred())
+        {
+            throw AbortMarshaling(); // String conversion failed; a Python exception is set.
+        }
         os->startSlice(id, -1, true);
         os->endSlice();
     }
--- a/python/modules/IcePy/Communicator.cpp
+++ b/python/modules/IcePy/Communicator.cpp
@@ -500,7 +500,10 @@ communicatorShutdownCompleted(CommunicatorObject* self, PyObject* /*args*/)
     }
 
     // Call Ice.Future.__init__
-    type->tp_init(future.get(), emptyArgs.get(), nullptr);
+    if (type->tp_init(future.get(), emptyArgs.get(), nullptr) < 0)
+    {
+        return nullptr;
+    }
 
     // Create a new reference to prevent premature release of the future object. The reference will be released by
     // the callback.
--- a/python/modules/IcePy/Connection.cpp
+++ b/python/modules/IcePy/Connection.cpp
@@ -375,7 +375,10 @@ connectionClose(ConnectionObject* self, PyObject* /* args */)
     }
 
     // Call Ice.Future.__init__
-    type->tp_init(future.get(), emptyArgs.get(), nullptr);
+    if (type->tp_init(future.get(), emptyArgs.get(), nullptr) < 0)
+    {
+        return nullptr;
+    }
 
     // Create a new reference to prevent premature release of the future object. The reference will be released by
     // either the success or exception callback, depending on the outcome of the close operation.
--- a/python/modules/IcePy/Logger.cpp
+++ b/python/modules/IcePy/Logger.cpp
@@ -77,7 +77,12 @@ IcePy::LoggerWrapper::getPrefix()
     {
         throwPythonException();
     }
-    return getString(tmp.get());
+    string result = getString(tmp.get());
+    if (PyErr_Occurred())
+    {
+        throwPythonException();
+    }
+    return result;
 }
 
 Ice::LoggerPtr
--- a/python/modules/IcePy/Operation.cpp
+++ b/python/modules/IcePy/Operation.cpp
@@ -2602,7 +2602,10 @@ IcePy::createFuture(PyObject* asyncInvocationContext)
     PyTuple_SET_ITEM(initArgs.get(), 0, Py_NewRef(asyncInvocationContext));
 
     // Call the Ice.InvocationFuture.__init__ method.
-    type->tp_init(future.get(), initArgs.get(), nullptr);
+    if (type->tp_init(future.get(), initArgs.get(), nullptr) < 0)
+    {
+        return nullptr;
+    }
 
     // Release the reference to the future object and let the caller take ownership.
     return future.release();
--- a/python/modules/IcePy/Util.cpp
+++ b/python/modules/IcePy/Util.cpp
@@ -126,7 +126,7 @@ IcePy::getStringArg(PyObject* p, const string& arg, string& val)
         val = getString(p);
         if (PyErr_Occurred())
         {
-            return false; // The string is not UTF-8 encodable.
+            return false; // String conversion failed; a Python exception is set.
         }
     }
     else if (p != Py_None)
@@ -377,7 +377,7 @@ IcePy::listToStringSeq(PyObject* l, Ice::StringSeq& seq)
             str = getString(item);
             if (PyErr_Occurred())
             {
-                return false; // The string is not UTF-8 encodable.
+                return false; // String conversion failed; a Python exception is set.
             }
         }
         else if (item != Py_None)
@@ -433,7 +433,7 @@ IcePy::tupleToStringSeq(PyObject* t, Ice::StringSeq& seq)
             str = getString(item);
             if (PyErr_Occurred())
             {
-                return false; // The string is not UTF-8 encodable.
+                return false; // String conversion failed; a Python exception is set.
             }
         }
         else if (item != Py_None)
@@ -463,7 +463,7 @@ IcePy::dictionaryToContext(PyObject* dict, Ice::Context& context)
             keystr = getString(key);
             if (PyErr_Occurred())
             {
-                return false; // The string is not UTF-8 encodable.
+                return false; // String conversion failed; a Python exception is set.
             }
         }
         else if (key != Py_None)
@@ -478,7 +478,7 @@ IcePy::dictionaryToContext(PyObject* dict, Ice::Context& context)
             valuestr = getString(value);
             if (PyErr_Occurred())
             {
-                return false; // The string is not UTF-8 encodable.
+                return false; // String conversion failed; a Python exception is set.
             }
         }
         else if (value != Py_None)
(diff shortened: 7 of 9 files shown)


python/modules/IcePy/Slice.cpp
```diff
@@ -196,6 +196,13 @@ IcePy_compileSlice(PyObject* /*self*/, PyObject* args)
         return nullptr;
     }
 
+    // Slice::Python::compile and the catch handlers below expect args[0] to be the program name.
+    if (argSeq.empty())
+    {
+        PyErr_Format(PyExc_ValueError, "arguments cannot be empty");
```
pepone: We can make the error message clear, mention the require first arg is the program name, like we do in the comment.

python/modules/IcePy/Types.cpp
```diff
@@ -85,7 +85,12 @@ namespace IcePy
         }
         else if (checkString(p))
```
pepone, line 86: Is this update correct, we already use checkString above.

pepone:

**Applied**
- *`compileSlice` `ValueError` message too generic:* made it actionable — "compileSlice requires a non-empty argument list whose first element is the program name".
- *`// The string is not UTF-8 encodable` comment is not always accurate:* reworded to the exception-agnostic `// String conversion failed; a Python exception is set.` at every new site.
- *Empty `except` blocks (Slice/import `Client.py`; Twoways UnicodeEncodeError blocks):* added brief `# Expected: …` comments.

- *Introduce a `getStringOrSetError` helper to reduce duplication:* declined for this PR — it's a 3.8.3-milestone backport fix and we're keeping it minimal/surgical. Introducing a new internal helper and rewiring the ~6 boundary sites broadens the patch and the backport risk; it's a reasonable follow-up cleanup on main.
- *Add a CHANGELOG.md entry:* 3.8.3-milestone, backport-bound fix. Per the project convention (commit `2b0f57c447` / #5371; main's CHANGELOG.md holds only 3.9-exclusive entries), backported fixes get their changelog entry in `CHANGELOG-3.8.md` on the `3.8` branch at backport time, not main's 3.9.0 changelog.

**Additional fix from a code-review pass:** an independent review (codex) found that one more `getString` boundary — `ValueWriter::_iceWrite`, which converts the value's `ice_id()` type ID during class marshaling — still ignored a conversion failure. It now checks `PyErr_Occurred()` and aborts marshaling, consistent with the other boundaries this PR hardens. (commit `0bc87688e2`)

pepone:
A code-review (codex) pass found additional instances of the same bug classes addressed in this PR.

**Applied fixes:**
- Two more unchecked `tp_init` Future-initialization sites, now checked exactly like the `createFuture` fix (`if (type->tp_init(...) < 0) { return nullptr; }`):
  - `Communicator.cpp` — `communicatorShutdownCompleted` (the `waitForShutdownAsync` path)
  - `Connection.cpp` — `connectionClose` (the close path)
  Previously, if `Ice.Future.__init__` raised, these proceeded with an uninitialized future.
- One more reachable unchecked `getString` boundary, now propagating conversion failures via `throwPythonException()` when `PyErr_Occurred()`:
  - `Logger.cpp` — `LoggerWrapper::getPrefix` (a logger prefix containing a lone surrogate would otherwise yield "" with a pending exception → deferred `SystemError`).

**Intentionally left unchanged:** the remaining `getString` calls the review flagged (Slice type IDs, Python class module/name, Slice member names, and the `RequestFailedException` facet/operation/identity and unknown-exception message text in `Util.cpp`) consume identifiers from generated code/type metadata or run in the exception-reporting path, rather than arbitrary external input. A lone surrogate is not a reachable injection point there, and adding checks would put dead branches in the hot unmarshaling and exception-reporting paths. They were intentionally left unchanged.

python/modules/IcePy/Slice.cpp
pepone: Done — the message now reads `compileSlice requires a non-empty argument list whose first element is the program name`, calling out the program-name requirement as the comment describes.

python/modules/IcePy/Types.cpp
pepone: `checkString(p)` only confirms `p` is a Python `str` object; it doesn't guarantee the string is UTF-8-encodable. `getString` calls `PyUnicode_AsUTF8String`, which fails for a `str` containing a lone surrogate (e.g. `"\ud800"`) — it returns `""` and leaves a `UnicodeEncodeError` set. That's the boundary from #5529: without this check the marshaled string is silently empty and the stale exception surfaces later, misattributed. So the `PyErr_Occurred()` check is needed in addition to `checkString`.

bernardnormier (approved):

PR #5568, opened 2026-06-12
Fix slice2py writing typing-only whole-module imports as runtime imports

### LINKED ISSUE #5483: slice2py: typing-only whole-module imports emitted as unconditional runtime imports

## Summary

In `slice2py`'s `writeImports`, the loop that builds the `if TYPE_CHECKING:` block writes whole-module imports to the runtime buffer `out` instead of the typing-only buffer `outT` (a copy-paste from the runtime-imports loop). A typing-only dependency (e.g. `numpy` via `["python:numpy.ndarray"]` sequence metadata) therefore becomes a hard runtime import. Additionally `hasTypingImports` is set without writing into `outT`, so a module whose typing imports are solely whole-module imports would emit `if TYPE_CHECKING:` with an empty suite (`IndentationError`, currently latent).

## The bug

```cpp
// cpp/src/slice2py/PythonUtil.cpp:2692-2707  (writeImports — whole-module imports written to `out`, not `outT`)
```

## Impact

A typing-only metadata dependency becomes an unconditional runtime import of the generated module, pulling in (and requiring) packages like `numpy` at import time even when only used for type hints. The empty `TYPE_CHECKING` suite is a latent `IndentationError`.

## Source references

- **(slice2py / typing-only whole-module imports written to runtime buffer)** — [`cpp/src/slice2py/PythonUtil.cpp#L2692-L2707`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2py/PythonUtil.cpp#L2692-L2707)

## Recommendation

Write the typing-only whole-module imports to `outT` (the `TYPE_CHECKING` buffer), not `out`, and only set `hasTypingImports` when content is actually emitted into `outT`.

## Severity

**Medium** — typing-only dependencies become mandatory runtime imports in generated Python.

pepone:
In `writeImports`, the loop that builds the `if TYPE_CHECKING:` block wrote whole-module imports to the runtime buffer `out` instead of the typing-only buffer `outT` (a copy-paste from the runtime-imports loop). A typing-only dependency such as `numpy` (via `["python:numpy.ndarray"]` sequence metadata used only in operation signatures) therefore became an unconditional runtime import of the generated module. This also made the latent empty-`TYPE_CHECKING`-suite `IndentationError` possible, since `hasTypingImports` was set without writing anything into `outT`; with the fix, the flag always reflects content actually emitted into the buffer.

For example, the generated `Custom.py` of the `Ice/custom` test (interface with numpy-mapped sequence parameters) previously contained a top-level `import numpy`; it is now emitted inside the `if TYPE_CHECKING:` block.

Regression test: new `Slice/typingImports` suite compiles a Slice file whose numpy-mapped sequence is used only in interface operations, then asserts (via `ast`) that `import numpy` appears in the `TYPE_CHECKING` block and not at module top level, and that all generated files byte-compile.

Fixes #5483

### DIFF
--- a/python/test/Slice/typingImports/Test.ice
+++ b/python/test/Slice/typingImports/Test.ice
@@ -1,0 +1,16 @@
+// Copyright (c) ZeroC, Inc.
+
+#pragma once
+
+["python:identifier:generated.test.Slice.typingImports.Test"]
+module Test
+{
+    ["python:numpy.ndarray"] sequence<int> IntSeq;
+
+    // IntSeq is only used in operation signatures, so the generated code for this interface
+    // only needs numpy for type hints: the numpy import must be typing-only.
+    interface TypingImports
+    {
+        IntSeq op(IntSeq values);
+    }
+}
--- a/cpp/src/slice2py/PythonUtil.cpp
+++ b/cpp/src/slice2py/PythonUtil.cpp
@@ -2693,14 +2693,14 @@ namespace
             {
                 if (moduleImports.imported)
                 {
-                    out << nl << "import " << moduleName;
+                    outT << nl << "import " << moduleName;
                     if (moduleImports.moduleAlias.empty())
                     {
                         allImports.insert(moduleName);
                     }
                     else
                     {
-                        out << " as " << moduleImports.moduleAlias;
+                        outT << " as " << moduleImports.moduleAlias;
                         allImports.insert(moduleImports.moduleAlias);
                     }
                     hasTypingImports = true;
--- a/python/test/Slice/typingImports/Client.py
+++ b/python/test/Slice/typingImports/Client.py
@@ -16,7 +16,9 @@ import IcePy
 
 def importsModule(nodes: list[ast.stmt], module: str) -> bool:
     return any(
-        isinstance(node, ast.Import) and any(alias.name == module for alias in node.names) for node in nodes
+        (isinstance(node, ast.Import) and any(alias.name == module for alias in node.names))
+        or (isinstance(node, ast.ImportFrom) and node.module == module)
+        for node in nodes
     )
 
 
--- a/python/test/Slice/typingImports/Client.py
+++ b/python/test/Slice/typingImports/Client.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+
+# Copyright (c) ZeroC, Inc.
+
+import ast
+import os
+import py_compile
+import sys
+import tempfile
+
+from TestHelper import TestHelper, test
+
+import Ice
+import IcePy
+
+
+def importsModule(nodes: list[ast.stmt], module: str) -> bool:
+    return any(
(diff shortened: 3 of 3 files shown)


pepone:
## Review responses

**Ruff CI:** The failing "Ruff" check (`ruff format . --check --diff`) is fixed. I ran `ruff format` on `python/test/Slice/typingImports/Client.py`, which collapsed the multi-line `any(...)` expression in `importsModule` to a single line. Both `ruff check --ignore E402` and `ruff format --check` now pass clean for the suite.

**Copilot @ Client.py:20 — `importsModule` only detects `ast.Import`, not `ast.ImportFrom` — APPLIED.** Broadened the helper so it also returns true for an `ast.ImportFrom` whose `node.module == module`. It still returns true for `import numpy` and now also for `from numpy import ndarray`. The suite assertions still hold (the runtime section has neither form; the `TYPE_CHECKING` suite has the whole-module import), and the test still passes.

**Copilot @ Client.py:49 — test looks for `TypingImports.py` but the module is `Test` — REBUTTED (incorrect).** The new slice2py generates one module file per interface. The file is generated at `generated/test/Slice/typingImports/Test/TypingImports.py` (confirmed via `find ... -name 'TypingImports.py'`), and the test passes, so the lookup does find it.

**Copilot @ PythonUtil.cpp:2700 — add a CHANGELOG.md entry — REBUTTED (not applicable here).** This is a 3.8.3-milestone, backport-bound fix. Per project convention (see #5371 "Remove cherry-picked entries from the Ice 3.9.0 changelog", commit `2b0f57c447`), `main`'s CHANGELOG.md holds only 3.9-exclusive entries; backported fixes get their changelog entry in `CHANGELOG-3.8.md` on the `3.8` branch at backport time, not in main's 3.9.0 changelog.

python/test/Slice/typingImports/Test.ice
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) ZeroC, Inc.
+
+#pragma once
+
+["python:identifier:generated.test.Slice.typingImports.Test"]
+module Test
+{
+    ["python:numpy.ndarray"] sequence<int> IntSeq;
```
pepone, line 8: Is this the only case for typing only imports?

pepone: Effectively yes, for this bug. The defect was specifically about *whole-module* typing-only imports (the `import X` form) being written to the runtime buffer. The only metadata that produces one is `["python:numpy.ndarray"]` on a sequence used solely in operation signatures — it emits `import numpy` (`addTypingImport("numpy", "", …)`, the one empty-name typing import in `PythonUtil.cpp`). Every other typing-only dependency — forward-declared types, `Ice.Current`, `from numpy import ndarray`, `array.array`, memoryview type hints — is a *from-import* (`from X import Y`), which was always written to the correct `outT` buffer. So this is the representative (and currently only) trigger for the fixed path.

bernardnormier (approved):

PR #5570, opened 2026-06-12
Fix slice2js emitting an invalid '?' suffix when the first parameter is optional

### LINKED ISSUE #5480: slice2js: areRemainingParamsOptional never checks the first parameter — invalid TypeScript .d.ts
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Ice C++ source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source (including against generated `slice2js` output) by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; please verify the reasoning, reproducibility, and severity before acting on it.

## Summary

`areRemainingParamsOptional` increments the iterator **before** comparing, so the operation's first parameter is never compared against `name`. When the optional parameter being checked is the first parameter, the loop runs off the end without matching, the required-parameter check is skipped, and the function returns `true` — emitting the TypeScript `?` suffix where it is not allowed (a required parameter cannot follow an optional one).

## The bug

```cpp
// cpp/src/slice2js/Gen.cpp:2419-2435
bool areRemainingParamsOptional(const ParameterList& params, const string& name)
{
    auto it = params.begin();
    do { it++; } while (it != params.end() && (*it)->name() != name);   // skips position 0
    ...
}
```

Used at `:2694-2695` to decide whether the `?` suffix is valid in the generated `.d.ts`.

## Impact

`void op(optional(1) int a, int b);` generates `op(a?: number, b: number, context?: ...)` — TS1016 (a required parameter cannot follow an optional parameter); the generated `.d.ts` fails to type-check.

## Source references

- **(slice2js / areRemainingParamsOptional skips the first parameter)** — [`cpp/src/slice2js/Gen.cpp#L2419-L2435`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2js/Gen.cpp#L2419-L2435)

## Recommendation

Compare before incrementing (or use a `for`/`find` that examines `params.front()`), so the first parameter is included in the required-after-optional check.

## Severity

**Medium** — generated TypeScript `.d.ts` fails to type-check when an operation's first parameter is optional.

pepone:
`areRemainingParamsOptional` incremented its iterator before comparing, so the operation's first parameter was never compared against `name`. When the optional parameter being checked is the first parameter, the loop ran off the end, the required-parameter check was skipped, and the function returned `true` — emitting the TypeScript `?` suffix even when a required parameter follows. The generated `.d.ts` then failed to type-check with TS1016 (a required parameter cannot follow an optional parameter).

The loop now finds the named parameter starting from `params.begin()`, skips it, and checks the parameters that follow.

Regression test: the JS `Ice/optional` suite gains `opOptionalFirstParam(optional(2) int p1, int p2)` — an optional first parameter followed by a required one — exercised by the client against both servants. Pre-fix, the js build itself fails because gulp type-checks the generated `Test.d.ts` (`error TS1016` at the new operation); post-fix the signature is `p1: number | undefined, p2: number` and the suite passes.

Fixes #5480

### DIFF
--- a/cpp/src/slice2js/Gen.cpp
+++ b/cpp/src/slice2js/Gen.cpp
@@ -2698,8 +2698,10 @@ Slice::TypeScriptVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
         for (const auto& param : inParams)
         {
             // TypeScript doesn't allow optional parameters with '?' prefix before required parameters.
+            // Only the input parameters appear in this signature, so 'out' parameters must not be
+            // considered when deciding whether the remaining parameters are all optional.
             const string optionalPrefix =
-                param->isOptional() && areRemainingParamsOptional(paramList, param->name()) ? "?" : "";
+                param->isOptional() && areRemainingParamsOptional(inParams, param->name()) ? "?" : "";
             _out
                 << (param->mappedName() + optionalPrefix + ": " +
                     typeToTsString(param->type(), true, true, param->isOptional()));
--- a/cpp/src/slice2js/Gen.cpp
+++ b/cpp/src/slice2js/Gen.cpp
@@ -2419,10 +2419,16 @@ namespace
     bool areRemainingParamsOptional(const ParameterList& params, const string& name)
     {
         auto it = params.begin();
-        do
+        while (it != params.end() && (*it)->name() != name)
         {
-            it++;
-        } while (it != params.end() && (*it)->name() != name);
+            ++it;
+        }
+
+        // Skip the parameter itself: only the parameters that follow it matter.
+        if (it != params.end())
+        {
+            ++it;
+        }
 
         for (; it != params.end(); ++it)
         {
--- a/cpp/src/slice2js/Gen.cpp
+++ b/cpp/src/slice2js/Gen.cpp
@@ -2418,7 +2418,8 @@ namespace
 {
     bool areRemainingParamsOptional(const ParameterList& params, const string& name)
     {
-        auto it = std::find_if(params.begin(), params.end(), [&name](const auto& param) { return param->name() == name; });
+        auto it =
+            std::find_if(params.begin(), params.end(), [&name](const auto& param) { return param->name() == name; });
         assert(it != params.end()); // The caller always passes a parameter that exists in params.
         return std::all_of(std::next(it), params.end(), [](const auto& param) { return param->isOptional(); });
     }
--- a/cpp/src/slice2js/Gen.cpp
+++ b/cpp/src/slice2js/Gen.cpp
@@ -2418,27 +2418,9 @@ namespace
 {
     bool areRemainingParamsOptional(const ParameterList& params, const string& name)
     {
-        auto it = params.begin();
-        while (it != params.end() && (*it)->name() != name)
-        {
-            ++it;
-        }
-        assert(it != params.end());
-
-        // Skip the parameter itself: only the parameters that follow it matter.
-        if (it != params.end())
-        {
-            ++it;
-        }
-
-        for (; it != params.end(); ++it)
-        {
-            if (!(*it)->isOptional())
-            {
-                return false;
-            }
-        }
-        return true;
+        auto it = std::find_if(params.begin(), params.end(), [&name](const auto& param) { return param->name() == name; });
+        assert(it != params.end()); // The caller always passes a parameter that exists in params.
+        return std::all_of(std::next(it), params.end(), [](const auto& param) { return param->isOptional(); });
     }
 }
 
--- a/cpp/src/slice2js/Gen.cpp
+++ b/cpp/src/slice2js/Gen.cpp
@@ -2423,6 +2423,7 @@ namespace
         {
             ++it;
         }
+        assert(it != params.end());
 
         // Skip the parameter itself: only the parameters that follow it matter.
         if (it != params.end())
--- a/js/packages/test/Ice/optional/ClientPrivate.ice
+++ b/js/packages/test/Ice/optional/ClientPrivate.ice
@@ -17,4 +17,13 @@ module Test
 
         void opVoid(optional(1) int a, optional(2) string v);
     }
+
+    // Regression test for issue #5480: slice2js must not emit a '?' suffix for an optional
+    // parameter that is followed by a required parameter (this would generate invalid
+    // TypeScript that fails to compile, TS1016). This interface is compiled only for the JS
+    // client; its generated .d.ts is type-checked by the build, which is what exercises the bug.
+    interface OptionalParamOrder
+    {
+        int opOptionalFirstParam(optional(1) int p1, int p2);
+    }
 }
--- a/js/packages/test/Ice/optional/ClientPrivate.ice
+++ b/js/packages/test/Ice/optional/ClientPrivate.ice
@@ -17,19 +17,4 @@ module Test
 
         void opVoid(optional(1) int a, optional(2) string v);
     }
-
-    // Regression test for issue #5480: slice2js must not emit a '?' suffix for an optional
-    // parameter that is followed by a required parameter (this would generate invalid
-    // TypeScript that fails to compile, TS1016). This interface is compiled only for the JS
-    // client; its generated .d.ts is type-checked by the build, which is what exercises the bug.
-    interface OptionalParamOrder
-    {
-        int opOptionalFirstParam(optional(1) int p1, int p2);
-
-        // A trailing optional *input* parameter must still get the '?' suffix in the proxy
-        // signature even when a required *out* parameter follows it, because 'out' parameters
-        // are not part of the input-parameter list. slice2js must not consider 'out' parameters
-        // when deciding whether the remaining input parameters are all optional.
-        void opOptionalLastInParam(optional(1) int p1, out int p2);
-    }
 }
--- a/js/packages/test/Ice/optional/ClientPrivate.ice
+++ b/js/packages/test/Ice/optional/ClientPrivate.ice
@@ -25,5 +25,11 @@ module Test
     interface OptionalParamOrder
     {
         int opOptionalFirstParam(optional(1) int p1, int p2);
+
+        // A trailing optional *input* parameter must still get the '?' suffix in the proxy
+        // signature even when a required *out* parameter follows it, because 'out' parameters
+        // are not part of the input-parameter list. slice2js must not consider 'out' parameters
+        // when deciding whether the remaining input parameters are all optional.
+        void opOptionalLastInParam(optional(1) int p1, out int p2);
     }
 }
--- a/js/packages/test/Ice/optional/AMDInitialI.ts
+++ b/js/packages/test/Ice/optional/AMDInitialI.ts
@@ -61,10 +61,6 @@ export class AMDInitialI extends Test.Initial {
         return [p1, p1];
     }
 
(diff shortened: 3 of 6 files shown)


pepone:
## Fixed the `release on ubuntu-24.04` CI failure

That check failed because the `Ice/optional` test runs **cross-language** (e.g. a JS client against a C# server). The new `opOptionalFirstParam` regression operation had been added only to the JS side of the **shared** `Ice/optional/Test.ice` (plus the JS servants and a client-side runtime call). The C# server therefore did not implement it, so the JS client's invocation raised `OperationNotExistException` and the suite failed.

### Fix
Issue #5480 is purely about the **generated `.d.ts` failing to type-check** (TS1016: a required parameter cannot follow an optional one). It is caught at **build time** by gulp's `tsc` (`noEmitOnError`) — no server round-trip is required. So I relocated the regression to the **client-only** `js/packages/test/Ice/optional/ClientPrivate.ice` (compiled only for the JS client, never shared with servers) as a standalone `OptionalParamOrder` interface, and reverted the shared `Test.ice`, both servants, and the client-side runtime call. The build's `tsc` type-check of the generated `ClientPrivate.d.ts` is what now exercises #5480.

Verified end-to-end:
- Generated signature has **no** `?` suffix: `opOptionalFirstParam(p1: number | undefined, p2: number, ...)`.
- **Red-check:** reverting only the `Gen.cpp` fix makes the build fail with `ClientPrivate.d.ts(121,55): error TS1016: A required parameter cannot follow an optional parameter.` — proving the relocated test still catches the bug.
- `Ice/optional` suite passes (no longer references a cross-language op).

### Copilot review comments
- **`Gen.cpp:2434` (find could return `true` if `name` not in `params`) — APPLIED.** Added `assert(it != params.end());` right after the find-loop. The function is only ever called with a valid parameter name, so this is a defensive guard.
- **`InitialI.ts` / `AMDInitialI.ts` (`_current` → `_`) — MOOT.** Those servant changes were reverted as part of the relocation, so there is nothing left to rename.
- **`Gen.cpp:2423` CHANGELOG entry — REBUTTED.** This is a 3.8.3 backport-bound fix. Per commit `2b0f57c447` (#5371, "Remove cherry-picked entries from the Ice 3.9.0 changelog") and the fact that main's `CHANGELOG.md` holds only 3.9-exclusive entries, backported fixes get their changelog entry in `CHANGELOG-3.8.md` on the `3.8` branch at backport time — not in main's 3.9.0 changelog.

pepone:
**Follow-up: a second, related slice2js bug found during review.**

While reviewing this fix, a code-review pass (codex) surfaced a related defect in the same helper. `areRemainingParamsOptional` was called with `paramList` (which includes `out` parameters), but the generated TypeScript proxy signature contains only the **input** parameters. So for an operation whose last *input* parameter is optional but which also has a required `out` parameter — e.g. `void op(optional(1) int p1, out int p2)` — the required `out` parameter caused the helper to return `false`, and `p1` did **not** get its `?` suffix. The generated signature became `op(p1: number | undefined, ...)`, forcing callers to pass `undefined` explicitly instead of omitting the argument.

Fix (commit `b8f91553df`): pass `inParams` rather than `paramList` at the single call site (proxy signatures). The dispatch/servant side never emits `?`, so it is unaffected. Added a second regression operation `opOptionalLastInParam(optional(1) int p1, out int p2)` to the client-only `ClientPrivate.ice`; its generated `.d.ts` now correctly shows `p1?: number | undefined`, while `opOptionalFirstParam` (optional-first-then-required) correctly has no `?`. JS build and the `Ice/optional` suite pass.

js/packages/test/Ice/optional/ClientPrivate.ice
```diff
@@ -17,4 +17,19 @@ module Test
 
         void opVoid(optional(1) int a, optional(2) string v);
     }
+
+    // Regression test for issue #5480: slice2js must not emit a '?' suffix for an optional
```
pepone: We should check if this bugs affects other mappings.

cpp/src/slice2js/Gen.cpp
```diff
@@ -2419,10 +2419,17 @@ namespace
...
+        while (it != params.end() && (*it)->name() != name)
         {
-            it++;
-        } while (it != params.end() && (*it)->name() != name);
+            ++it;
+        }
+        assert(it != params.end());
+
+        // Skip the parameter itself: only the parameters that follow it matter.
+        if (it != params.end())
```
pepone: What is the point of this `if`, are we guarding against a bogus call, which provide a name not in parmas?

cpp/src/slice2js/Gen.cpp
```diff
@@ -2419,10 +2419,17 @@ namespace
     bool areRemainingParamsOptional(const ParameterList& params, const string& name)
     {
         auto it = params.begin();
-        do
+        while (it != params.end() && (*it)->name() != name)
```
pepone: Wondering if we can replace the loops with `find_if` and `find_if_not`

cpp/src/slice2js/Gen.cpp
pepone: Right — there is no runtime case where `name` is absent. `name` is an invariant from the single call site, which passes `param->name()` for a `param` it draws from the same `inParams` list, so the parameter is always present. There is nothing meaningful to do when it is missing, so I removed the defensive `if` and let the `assert` document the invariant instead.

cpp/src/slice2js/Gen.cpp
pepone: Done — the find loop is now `std::find_if`, and the trailing check is `std::all_of(std::next(it), params.end(), ...)`. I went with `all_of` rather than `find_if_not` because it reads directly as "all remaining params are optional"; it is equivalent to the suggested `find_if_not(...) == end()`.

js/packages/test/Ice/optional/ClientPrivate.ice
pepone: Checked the other Slice compilers — this bug is specific to slice2js:

- **slice2py**: not affected. It computes the last required in-parameter and emits the `= None` default only for parameters that *follow* it (`PythonUtil.cpp`, the `lastRequiredParameter` logic), so `op(optional(1) int p1, int p2)` correctly produces no default on either param.
- **slice2cs / slice2java**: not affected. Optional parameters map to nullable / wrapper *types* (`T?`, `Optional<T>`), not language-level default/optional parameters, so there's no required-after-optional ordering constraint and no default-argument emission in operation signatures.
- **slice2php / slice2rb / slice2matlab**: not affected. `isOptional` there only drives marshaling metadata (tags), not method-signature ordering.

Only slice2js emits `param?: T` TypeScript signatures, which carry the required-after-optional (TS1016) constraint, and `areRemainingParamsOptional` was the single place that decision is made.

bernardnormier (approved):
While the fix looks correct, the overall feature (special mapping for trailing params where they are all optional) is not.

We should fix that for Ice 3.9. See https://github.com/icerpc/icerpc-csharp/issues/4121.

js/packages/test/Ice/optional/ClientPrivate.ice
```diff
@@ -17,4 +17,19 @@ module Test
 
         void opVoid(optional(1) int a, optional(2) string v);
     }
+
```
InsertCreativityHere: Reading through the PR, it sounds like this new operation used to live in `Test.ice` and also was called at runtime. This broke cross-testing and so the test was moved here, and the runtime test deleted.

I don't think that test's placement is correct though, it should stay in `Test.ice`. `ClientPrivate` is just for our special hack to test unknown optionals. And just having the operation there can't break anything. It was only _calling_ the operation that broke cross-testing.

js/packages/test/Ice/optional/ClientPrivate.ice
```diff
@@ -17,4 +17,19 @@ module Test
...
+    // Regression test for issue #5480: slice2js must not emit a '?' suffix for an optional
+    // parameter that is followed by a required parameter (this would generate invalid
+    // TypeScript that fails to compile, TS1016). This interface is compiled only for the JS
+    // client; its generated .d.ts is type-checked by the build, which is what exercises the bug.
```
InsertCreativityHere: I also think that these comments are too wordy. At some point the extra details just make it harder to digest IMO.

```suggestion
    // Test for issue #5480: slice2js must not emit a '?' suffix for an optional parameter that is followed by a
    // required parameter. This interface's generated .d.ts is type-checked by the build and exercises the bug.
```

js/packages/test/Ice/optional/ClientPrivate.ice
```diff
@@ -17,4 +17,19 @@ module Test
...
+    interface OptionalParamOrder
+    {
+        int opOptionalFirstParam(optional(1) int p1, int p2);
+
+        // A trailing optional *input* parameter must still get the '?' suffix in the proxy
+        // signature even when a required *out* parameter follows it, because 'out' parameters
+        // are not part of the input-parameter list. slice2js must not consider 'out' parameters
+        // when deciding whether the remaining input parameters are all optional.
```
InsertCreativityHere: Same here. Feels like it's just restating the same fact in slightly different ways.

```suggestion
        // slice2js must not consider 'out' parameters when deciding if the remaining 'in' parameters are all optional.
        // Here, the trailing optional 'in' parameter must still get a '?' suffix in the proxy signature.
```

InsertCreativityHere (approved):
The fix looks good, I think this test was correct being in `Test.ice` though, and it shouldn't of been moved to `ClientPrivate.ice`.

js/packages/test/Ice/optional/ClientPrivate.ice
pepone: Done — moved both regression ops (`opOptionalFirstParam` and `opOptionalLastInParam`) back into the shared `Test.ice`, in the `Initial` interface where `opOptionalFirstParam` originally lived, and reverted `ClientPrivate.ice` to its pre-PR state.

The servants (`InitialI`/`AMDInitialI`) implement both ops, but the client never *calls* them at runtime — so cross-language testing is unaffected (the C# server doesn't need them; only the runtime call broke it before). #5480 is still caught at build time by `tsc` type-checking the generated `Test.d.ts`.

Verified: JS build + `Ice/optional` suite pass, and injecting the buggy signature into `Test.d.ts` reproduces `error TS1016` (red-check), confirming the relocated test is still covered by the build's type-check.

js/packages/test/Ice/optional/ClientPrivate.ice
pepone: Applied (adapted to the new location). The comment moved with the op into `Test.ice` and now reads:

```
// Regression test for issue #5480: slice2js must not emit a '?' suffix for an optional parameter
// that is followed by a required parameter (this would generate invalid TypeScript, TS1016).
```

I dropped the "this interface's generated .d.ts" phrasing since the op now lives in the existing `Initial` interface rather than a dedicated one.

js/packages/test/Ice/optional/ClientPrivate.ice
pepone: Applied (adapted). Moved into `Test.ice` and shortened to:

```
// slice2js must not consider 'out' parameters when deciding if the remaining 'in' parameters are
// all optional; this trailing optional 'in' parameter must still get a '?' suffix in the proxy signature.
```

Issue #5512, opened 2026-06-12
[Audit-Low] Java PerThreadImplicitContext.remove returns null where SharedImplicitContext.remove returns ""


pepone:
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Java source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

`PerThreadImplicitContext.remove` returns `null` when the calling thread has never set any context, while `SharedImplicitContext.remove` normalizes the same logical situation (no previous value) to `""` — and every other `PerThread`/`Shared` accessor in the file also normalizes missing values to `""`. The two implementations of the same `ImplicitContext` interface disagree, and callers can receive a surprise `null`.

## The bug

```java
// java/src/com.zeroc.ice/.../Ice/ImplicitContextI.java:205-216 (PerThreadImplicitContext)
Map<String, String> threadContext = _map.get(Thread.currentThread());
if (threadContext == null) {
    return null;                   // Shared.remove returns "" here
}
String val = threadContext.remove(key);
if (val == null) {
    val = "";                      // the missing-key case IS normalized
}
return val;
```

The C# per-thread port returns `""` in the equivalent case, confirming a Java-only inconsistency.

## Impact

`communicator.getImplicitContext().remove("key")` on a thread that never touched the context returns `null` under `Ice.ImplicitContext=PerThread` but `""` under `Shared` — an inconsistency that can surface as an NPE in caller code that trusts the documented "previous value" return.

## Source references

Pinned to commit `e248644d788c7e76e14af8676529201c54c1735d`:

- **(per-thread `remove` returns null)** — [`ImplicitContextI.java#L199-L216`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ImplicitContextI.java#L199-L216)
- **(shared `remove` normalizes to `""`)** — [`ImplicitContextI.java#L80-L92`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ImplicitContextI.java#L80-L92)
[...]

Issue #5499, opened 2026-06-12
[Audit-Medium, grouped] C# ImplicitContext: get/remove/setContext throw where the contract promises ""/replace; per-thread getContext returns the live dictionary


pepone:
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the C# source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it. Grouped: five defects, all in `ImplicitContextI.cs`, all the same fix theme.

## Summary

Five contract violations in `csharp/src/Ice/ImplicitContextI.cs`. Four are the same C# idiom slip: a `Dictionary<…>` indexer `[key]` (or `.Add`) is used where the method's own **dead guard** (`?? ""`, `if (val == null)`) shows it intended a default/replace — so the method throws where the `ImplicitContext` contract promises "return the empty string" or "replace". The fifth returns the live per-thread dictionary where the contract promises a copy. In every case a sibling method in the same file does it correctly (`put` uses `TryGetValue`; `SharedImplicitContext.setContext` overwrites; `SharedImplicitContext.getContext` copies) — the signature of an incomplete port off a `Hashtable`-style collection whose indexer returned `null`.

### 1. `SharedImplicitContext.get` throws `KeyNotFoundException` on a missing key

```csharp
// ImplicitContextI.cs:86
string val = _context[key] ?? "";   // indexer throws if key absent; `?? ""` is dead
return val;
```

`ImplicitContext.get` is documented to return "the empty string if no value is associated with the key".

### 2. `SharedImplicitContext.remove` throws `KeyNotFoundException` on a missing key

```csharp
// ImplicitContextI.cs:112
string val = _context[key];         // throws if key absent
if (val == null) { val = ""; } else { _context.Remove(key); }   // dead null branch
```

### 3. `PerThreadImplicitContext.get` throws `KeyNotFoundException` on a missing key

```csharp
// ImplicitContextI.cs:238
string val = threadContext[key] ?? "";   // throws if key absent in this thread's context
```

The missing-*map* case (thread has no context at all) correctly returns `""` just above; the missing-*key* case throws.

### 4. `PerThreadImplicitContext.setContext` uses `Dictionary.Add` — throws on replace

```csharp
// ImplicitContextI.cs:204
[...]

PR #5572, opened 2026-06-12
Fix JS ImplicitContext returning undefined where the contract says empty string

### LINKED ISSUE #5518: [Audit-Low] JS ImplicitContext compares native Map results with === null — get/put/remove return undefined where the contract says ""
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the JavaScript source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

`ImplicitContext.get`/`put`/`remove` normalize missing keys with `if (val === null) { val = ""; }` — but the backing `Ice.Context` resolves to a **native `Map`** (`defineDictionary` with `keysEqual === undefined` selects `Map` in `HashMap.js`), and `Map.prototype.get` returns `undefined` for an absent key, never `null`. Every `=== null` normalization is dead: `get` and `remove` return `undefined` for a missing key, and `put` returns `undefined` as the "old value" for a previously-unset key — where the contract (and the C++, C#, and Java implementations) return `""`.

This is the JavaScript member of the cross-language ImplicitContext family (C#: #5499, Java: #5512).

## The bug

```js
// js/packages/ice/src/Ice/ImplicitContext.js:41-46 (get; same pattern in put/remove)
let val = this._context.get(key);
if (val === null) {     // Map.get returns undefined, never null — dead branch
    val = "";
}
return val;
```

## Impact

`communicator.getImplicitContext().get("missing")` returns `undefined` instead of `""`; same for the first `put` and a missing-key `remove`. This conflicts with the declared `string` return type and the other language mappings.

Note from the verification pass: the operations test **explicitly expects the buggy behavior** — [`js/packages/test/Ice/operations/Twoways.ts#L1323-L1326`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/test/Ice/operations/Twoways.ts#L1323-L1326) asserts that a first `put` returns `undefined`, masking the defect; that assertion needs updating along with the fix.

## Source references

- **(dead `=== null` normalizations)** — [`ImplicitContext.js#L36-L78`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/ImplicitContext.js#L36-L78)
- **(`Context` resolves to native `Map`)** — [`Context.js#L19-L21`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/Context.js#L19-L21), [`HashMap.js#L343-L355`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/HashMap.js#L343-L355)

## Recommendation

Compare against `undefined` (`if (val === undefined)`) in all three methods, and fix the masking test assertion.

## Severity

**Low** — public API returns `undefined` where the contract says `""`. Verified **CONFIRMED** by the second-model pass.

pepone:
`ImplicitContext.get`/`put`/`remove` normalized missing keys with `=== null`, but the backing `Ice.Context` is a native `Map`, whose `get` returns `undefined` for an absent key — never `null`. The dead normalizations meant `get`/`remove` returned `undefined` for a missing key and `put` returned `undefined` as the old value of a previously-unset key, where the declared return type (and the C++, C#, and Java implementations) say `""`. The checks now compare against `undefined`; the `null` normalizations of input arguments are unchanged.

The `Ice/operations` test explicitly asserted the buggy behavior (`test(r === undefined)` for a first `put`); it now asserts `""` and additionally covers `get` and `remove` of missing keys and the old-value returns of `put`.

This is the JavaScript member of the cross-language ImplicitContext family (C#: #5499, Java: #5512).

Fixes #5518

### DIFF
--- a/js/packages/ice/src/Ice/ImplicitContext.js
+++ b/js/packages/ice/src/Ice/ImplicitContext.js
@@ -39,7 +39,7 @@ export class ImplicitContext {
         }
 
         let val = this._context.get(key);
-        if (val === null) {
+        if (val === undefined) {
             val = "";
         }
 
@@ -55,7 +55,7 @@ export class ImplicitContext {
         }
 
         let oldVal = this._context.get(key);
-        if (oldVal === null) {
+        if (oldVal === undefined) {
             oldVal = "";
         }
 
@@ -72,7 +72,7 @@ export class ImplicitContext {
         let val = this._context.get(key);
         this._context.delete(key);
 
-        if (val === null) {
+        if (val === undefined) {
             val = "";
         }
         return val;
--- a/js/packages/ice/src/Ice/ImplicitContext.js
+++ b/js/packages/ice/src/Ice/ImplicitContext.js
@@ -34,7 +34,7 @@ export class ImplicitContext {
     }
 
     get(key) {
-        if (key === null) {
+        if (key === null || key === undefined) {
             key = "";
         }
 
@@ -47,10 +47,10 @@ export class ImplicitContext {
     }
 
     put(key, value) {
-        if (key === null) {
+        if (key === null || key === undefined) {
             key = "";
         }
-        if (value === null) {
+        if (value === null || value === undefined) {
             value = "";
         }
 
@@ -65,7 +65,7 @@ export class ImplicitContext {
     }
 
     remove(key) {
-        if (key === null) {
+        if (key === null || key === undefined) {
             key = "";
         }
 
--- a/js/packages/ice/src/Ice/ImplicitContext.js
+++ b/js/packages/ice/src/Ice/ImplicitContext.js
@@ -26,7 +26,7 @@ export class ImplicitContext {
     }
 
     containsKey(key) {
-        if (key === null) {
+        if (key === null || key === undefined) {
             key = "";
         }
 
--- a/js/packages/ice/src/Ice/ImplicitContext.d.ts
+++ b/js/packages/ice/src/Ice/ImplicitContext.d.ts
@@ -35,7 +35,7 @@ declare module "@zeroc/ice" {
              * Gets the value associated with the specified key in the request context.
              *
              * @param key - The key.
-             * @returns The value associated with the key, or the empty string if no value is associated with the key.
+             * @returns The value associated with the key, or the empty string if the key is not set.
              */
             get(key: string): string;
 
@@ -44,7 +44,7 @@ declare module "@zeroc/ice" {
              *
              * @param key - The key.
              * @param value - The value.
-             * @returns The previous value associated with the key, if any.
+             * @returns The previous value associated with the key, or the empty string if the key was not set.
              */
             put(key: string, value: string): string;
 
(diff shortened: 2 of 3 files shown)


js/packages/ice/src/Ice/ImplicitContext.js
```diff
@@ -39,7 +39,7 @@ export class ImplicitContext {
         }
 
         let val = this._context.get(key);
-        if (val === null) {
+        if (val === undefined) {
             val = "";
```
pepone, line 43: We should review the doc comments in `ImplicitContext.d.ts`. It is not clear that put/remove return `""` when not match is found by reading the doc comment.

pepone:
Thanks for the review. Both comments addressed:

**1. ImplicitContext.js:60 (Copilot) — applied.** With `undefined` as the missing-key sentinel for value reads, an `undefined` stored as a value would be indistinguishable from a missing key. I now normalize `undefined` inputs to `""` so `undefined` is never stored: `put` normalizes both an `undefined` key and an `undefined` value, and `get`/`put`/`remove` normalize an `undefined` key (alongside the existing `null` handling). The value-read normalizations stay `=== undefined` as intended.

**2. Twoways.ts:1331 (Copilot) — applied.** Introduced `const implicitContext = ic.getImplicitContext();` once and reused it for all the `containsKey`/`get`/`put`/`remove`/`getContext`/`setContext` calls in that block. Behavior is identical.

**Follow-up from a local review pass:** `containsKey` was not normalizing `undefined`/`null` like `get`/`put`/`remove` now do, so after `put(undefined, "x")` (key normalized to `""`), `get(undefined)` returned the value but `containsKey(undefined)` returned `false`. I normalized the key in `containsKey` for consistency and added test assertions covering null/undefined-key normalization across `put`/`get`/`containsKey`/`remove`.

Verified: `npm run build` (gulp type-check) passes, prettier clean on both files, and `allTests.py --filter="Ice/operations"` passes.

js/packages/ice/src/Ice/ImplicitContext.js
pepone: Updated the doc comments in `ImplicitContext.d.ts` so the `""`-when-not-found behavior is explicit:

- `get(key)` → "The value associated with the key, or the empty string if the key is not set."
- `put(key, value)` → "The previous value associated with the key, or the empty string if the key was not set."
- `remove(key)` → "The value the key had, or the empty string if the key was not set."

This is doc-comment only; runtime behavior is unchanged. Prettier and `npm run build` both pass.

bernardnormier (approved):

InsertCreativityHere (approved):
Looks good to me!

PR #5573, opened 2026-06-12
Fix JS struct equality so different struct types never compare equal

### LINKED ISSUE #5517: [Audit-Low] JS Struct/ArrayUtil equality: dead "this.prototype !== other.prototype" guard — different struct types with matching members compare equal
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the JavaScript source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

Both `Struct.equals` and `ArrayUtil.eq` guard with `x.prototype !== y.prototype` — but only constructors carry a `.prototype` property; **instances** do not. On instances both sides are `undefined`, so the type guard never fires, and equality falls through to a field-by-field comparison. Two structs of *different* Slice types whose members happen to coincide compare equal.

## The bug

```js
// js/packages/ice/src/Ice/Struct.js:21 (equals)
if (this.prototype !== other.prototype) {   // undefined !== undefined → never true
    return false;
}

// js/packages/ice/src/Ice/ArrayUtil.js:8 (eq)
} else if (e1.prototype !== e2.prototype) {  // same problem
    return false;
```

The intent was a same-type check: `Object.getPrototypeOf(this) !== Object.getPrototypeOf(other)` (or comparing `constructor`).

## Impact

This `equals` is installed on `Ice.Identity` (`IdentityExtensions.js`) and backs every `HashMap(HashMap.compareEquals)` keyed on `Identity` (ServantManager, LocatorTable, RouterInfo). `Struct.hashCode` hashes field values without a type discriminator, so two cross-type instances with matching field shapes pass both the hash pre-filter and the equality check. Requires distinct Slice types with coinciding members, so real-world impact is narrow.

## Source references

- **(dead guard, struct)** — [`Struct.js#L12-L34`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/Struct.js#L12-L34)
- **(dead guard, arrays)** — [`ArrayUtil.js#L3-L15`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/ArrayUtil.js#L3-L15)
- **(installed on `Ice.Identity`)** — [`IdentityExtensions.js#L3-L11`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/IdentityExtensions.js#L3-L11)
- **(type-blind field hashing)** — [`Struct.js#L73-L82`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/Struct.js#L73-L82)

## Recommendation

`Object.getPrototypeOf(this) !== Object.getPrototypeOf(other)` in both sites.

## Severity

**Low** — clear mechanism, narrow real-world trigger. Verified **CONFIRMED** by the second-model pass.

pepone:
`Struct.equals` guarded with `this.prototype !== other.prototype`, but instances don't carry a `.prototype` property, so the guard compared `undefined !== undefined` and never fired: two structs of different Slice types with coinciding members compared equal. The guard now uses `Object.getPrototypeOf`, which compares the instances' actual prototypes. This `equals` is installed on `Ice.Identity` and backs every Identity-keyed `HashMap` (ServantManager, LocatorTable, RouterInfo).

`ArrayUtil.eq` had the same dead `e1.prototype !== e2.prototype` branch. There it is **removed** rather than made live: `eq` delegates to the element's own `equals` when present, and making the prototype guard live would have wrongly rejected comparisons that are legitimately cross-class — most notably struct fields holding proxies, where a `MyInterfacePrx` written to a stream is read back as a base proxy class and equality is by reference (this actually broke the `Ice/stream` suite when tried). With the dead branch removed, `eq`'s behavior is unchanged, and the type check lives where it belongs, in `Struct.equals`.

Regression test: the `Ice/operations` suite gains `StillAnotherStruct`, member-identical to `AnotherStruct`, and asserts cross-type instances with equal fields do not compare equal (pre-fix they did), while same-type equality still holds. Verified with `Ice/operations`, `Ice/stream`, `Ice/optional`, and `Ice/location` (Identity-keyed locator table).

Fixes #5517

### DIFF
--- a/js/packages/ice/src/Ice/ArrayUtil.js
+++ b/js/packages/ice/src/Ice/ArrayUtil.js
@@ -5,9 +5,9 @@ const eq = function (e1, e2) {
         return true; // If identity compare equals members are equal.
     } else if (e1 === null || e1 === undefined || e2 === null || e2 === undefined) {
         return false;
-    } else if (e1.prototype !== e2.prototype) {
-        return false;
     } else if (typeof e1.equals == "function") {
+        // The type's own equals decides; for example, struct equality requires the same type
+        // while proxy equality compares the references regardless of the proxy class.
         return e1.equals(e2);
     } else if (e1 instanceof Array || e1 instanceof Uint8Array) {
         return ArrayUtil.equals(e1, e2, eq);
--- a/js/packages/ice/src/Ice/Struct.js
+++ b/js/packages/ice/src/Ice/Struct.js
@@ -18,6 +18,11 @@ function equals(other) {
         return false;
     }
 
+    if (typeof other !== "object") {
+        // A struct can never equal a primitive (number/string/boolean/symbol/bigint) or a function.
+        return false;
+    }
+
     if (Object.getPrototypeOf(this) !== Object.getPrototypeOf(other)) {
         return false;
     }
--- a/js/packages/ice/src/Ice/Struct.js
+++ b/js/packages/ice/src/Ice/Struct.js
@@ -18,7 +18,7 @@ function equals(other) {
         return false;
     }
 
-    if (this.prototype !== other.prototype) {
+    if (Object.getPrototypeOf(this) !== Object.getPrototypeOf(other)) {
         return false;
     }
 
--- a/js/packages/test/Ice/operations/Test.ice
+++ b/js/packages/test/Ice/operations/Test.ice
@@ -20,6 +20,12 @@ module Test
         string s;
     }
 
+    // Same members as AnotherStruct: instances of the two types must not compare equal.
+    struct StillAnotherStruct
+    {
+        string s;
+    }
+
     struct Structure
     {
         MyClass* p;
--- a/js/packages/test/Ice/operations/Twoways.ts
+++ b/js/packages/test/Ice/operations/Twoways.ts
@@ -260,6 +260,16 @@ export async function twoways(
     }
 
     {
+        // Struct equality requires the same Slice type: structs of different types with
+        // identical members must not compare equal.
+        const as1 = new Test.AnotherStruct("abc");
+        test(as1.equals(new Test.AnotherStruct("abc")));
+        test(!as1.equals(new Test.AnotherStruct("def")));
+        test(!as1.equals(new Test.StillAnotherStruct("abc")));
+        test(!new Test.StillAnotherStruct("abc").equals(as1));
+    }
+
+    {
         const bsi1 = new Uint8Array([0x01, 0x11, 0x12, 0x22]);
         const bsi2 = new Uint8Array([0xf1, 0xf2, 0xf3, 0xf4]);
 
--- a/js/packages/test/Ice/operations/Twoways.ts
+++ b/js/packages/test/Ice/operations/Twoways.ts
@@ -267,6 +267,14 @@ export async function twoways(
         test(!as1.equals(new Test.AnotherStruct("def")));
         test(!as1.equals(new Test.StillAnotherStruct("abc")));
         test(!new Test.StillAnotherStruct("abc").equals(as1));
+
+        // Comparing a struct to a primitive must return false without throwing. The symbol and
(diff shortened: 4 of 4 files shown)


js/packages/ice/src/Ice/ArrayUtil.js
```diff
@@ -5,9 +5,9 @@ const eq = function (e1, e2) {
         return true; // If identity compare equals members are equal.
     } else if (e1 === null || e1 === undefined || e2 === null || e2 === undefined) {
         return false;
-    } else if (e1.prototype !== e2.prototype) {
-        return false;
     } else if (typeof e1.equals == "function") {
+        // The type's own equals decides; for example, struct equality requires the same type
+        // while proxy equality compares the references regardless of the proxy class.
```
pepone, line 10: Is this correct? Shouldn't proxy equality require both same type, and same underlying reference equals

pepone:
## Review responses

**1. `Struct.js` — guard against primitive `other` (Copilot) — APPLIED.**
`Object.getPrototypeOf` is the right comparison, but it should never be reached with a primitive. I added an explicit guard right after the existing null/undefined check:

```js
if (typeof other !== "object") {
    // A struct can never equal a primitive (number/string/boolean/symbol/bigint) or a function.
    return false;
}
```

This keeps `equals` total and non-throwing. Worth a precise note: on current V8 (Node 22) `Object.getPrototypeOf` coerces primitives via `ToObject` (per ES2015) and returns the wrapper prototype, so the unguarded code already returned `false` for primitives rather than throwing — the ES5 `TypeError` behavior the comment describes does not reproduce on modern engines. The guard is still the correct, intent-revealing fix: it short-circuits all primitives (including `symbol`/`bigint`) before `getPrototypeOf`, so behavior is well-defined regardless of engine. `this` is always the struct instance, so only `other` needs guarding.

I also added regression assertions to the struct-equality block in `Twoways.ts` proving a struct never equals a primitive and never throws — specifically including the `symbol` and `bigint` cases:

```js
test(!as1.equals(5));
test(!as1.equals("abc"));
test(!as1.equals(true));
test(!as1.equals(Symbol()));
test(!as1.equals(10n));
```

**2. CHANGELOG entry (Copilot) — REBUTTED.**
This is a 3.8.3-milestone, backport-bound fix. Per the established convention (cf. #5371 / commit `2b0f57c447`), `main`'s `CHANGELOG.md` holds only 3.9-exclusive entries; backported fixes get their changelog entry in `CHANGELOG-3.8.md` on the `3.8` branch at backport time, not in main's 3.9.0 changelog. No CHANGELOG change here.

Verification: `npm run build` (gulp type-check) succeeds; `prettier --check` clean; `Ice/operations`, `Ice/stream`, and `Ice/optional` suites all pass.

js/packages/ice/src/Ice/ArrayUtil.js
pepone: Ice proxy equality is by value — it compares the underlying reference (identity, endpoints, facet, and all proxy settings) and is independent of the static proxy class. Two proxies of different static types referring to the same target with the same settings are equal, so `proxy.equals(other)` must not require the same JS class. That's why `eq` delegates to the element's own `equals` with no prototype guard: a struct field declared as a proxy is read back from a stream as a base proxy class and must still compare equal to the original — making the prototype guard live here is exactly what broke `Ice/stream`. So proxy equality requires the same underlying reference (which `equals` checks) but **not** the same type. The struct-type discrimination lives in `Struct.equals`, where two different Slice struct types must not be equal.

bernardnormier (approved):

PR #5574, opened 2026-06-12
Fix JS ConnectRequestHandler completing queued requests through a nonexistent property

### LINKED ISSUE #5514: [Audit-Medium] JS ConnectRequestHandler.flushRequests calls request.out.completedEx — a flush failure surfaces as a raw TypeError, losing the LocalException and skipping queued requests
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the JavaScript source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it. Originally graded High ("queue hangs forever"); **regraded to Medium** by the verification pass (see Severity).

## Summary

`ConnectRequestHandler.flushRequests`'s failure path calls `request.out.completedEx(ex)`, but the queued items **are** the outgoing-async objects themselves — they're pushed as `this._requests.push(out)`, and the sibling `setException` path correctly calls `request.completedEx(...)`. `request.out` is `undefined`, so the `LocalException` branch throws a raw `TypeError` instead of completing the invocation.

## The bug

```js
// js/packages/ice/src/Ice/ConnectRequestHandler.js:118-128 (flushRequests failure path)
} catch (ex) {
    if (ex instanceof RetryException) {
        ...
    } else {
        console.assert(ex instanceof LocalException, ex);
        exception = ex;
        request.out.completedEx(ex);   // request.out is undefined → TypeError
    }
}
```

The push site (`:23`) stores the async object directly, and `setException` (`:82-90`) completes the same queue entries with `request.completedEx(this._exception)`.

## Impact

The `TypeError` is not a `LocalException`, so it escapes `flushRequests` and rejects the async `setConnection` call. Per the verification pass, the originally-claimed permanent hang does **not** occur: `Reference.getConnection` catches the rejection and calls `handler.setException(ex)`, which drains the queue. The real impact is still significant:

- the original `LocalException` is **lost** — every affected invocation settles with an opaque `TypeError` instead of the real cause;
- the in-progress flush is aborted, so later-queued requests are never sent on this handler;
- requests already flushed earlier in the same loop can be re-failed from the still-undrained queue.

Trigger: any invocation queued while the connection is being established, where the subsequent send fails with a `LocalException` (e.g. the connection closes during flush).

## Source references

- **(the bad dereference)** — [`ConnectRequestHandler.js#L120-L128`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/ConnectRequestHandler.js#L120-L128)
- **(push site / correct sibling path)** — [`ConnectRequestHandler.js#L17-L24`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/ConnectRequestHandler.js#L17-L24), [`#L82-L90`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/ConnectRequestHandler.js#L82-L90)
- **(outer recovery that prevents the hang)** — [`Reference.js#L817-L841`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/Reference.js#L817-L841), [`OutgoingAsync.js#L74-L80`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/OutgoingAsync.js#L74-L80)

## Recommendation

`request.completedEx(ex);`, mirroring the `setException` path.

## Severity

**Medium** (regraded from High by the verification pass, which traced the recovery path with Node.js probes) — no hang, but a lost failure cause, an aborted flush, and possible double-fail of earlier requests on the connection-establishment path.

pepone:
The failure path of `ConnectRequestHandler.flushRequests` called `request.out.completedEx(ex)`, but the queued entries are the outgoing-async objects themselves (`this._requests.push(out)`), so `request.out` is `undefined` and the call threw a raw `TypeError`. The original `LocalException` was lost (affected invocations settled with an opaque `TypeError` via the outer recovery in `Reference.getConnection`), the in-progress flush was aborted so later-queued requests were never sent on the handler, and already-flushed requests could be re-failed from the still-undrained queue.

The fix calls `request.completedEx(ex)`, mirroring the sibling `setException` path that completes the same queue entries.

No regression test: triggering this path deterministically requires a connection that fails with a `LocalException` in the middle of `flushRequests` during connection establishment — a race that can't be reliably staged from the public API. Verified with the connection-establishment-heavy `Ice/ami`, `Ice/exceptions`, and `Ice/binding` suites.

Fixes #5514

### DIFF
--- a/js/packages/ice/src/Ice/ConnectRequestHandler.js
+++ b/js/packages/ice/src/Ice/ConnectRequestHandler.js
@@ -122,7 +122,7 @@ export class ConnectRequestHandler {
                 } else {
                     console.assert(ex instanceof LocalException, ex);
                     exception = ex;
-                    request.out.completedEx(ex);
+                    request.completedEx(ex);
                 }
             }
         }


pepone:
Thanks for the suggestion @copilot. I'm going to **decline** adding a dedicated unit test here, for a few reasons:

1. **The failure path isn't deterministically reproducible from the public API.** Triggering the `flushRequests` failure branch requires a connection that fails with a `LocalException` *mid-flush, during connection establishment* — a timing-dependent race. Forcing it deterministically would mean reaching into and stubbing `ConnectRequestHandler` and/or the transceiver internals.

2. **There's no unit/mocking harness for this layer in the JS test suite.** The JS tests here are integration-style (real client/server over real transports); there is no existing fixture for mocking `ConnectRequestHandler` or the transceivers. Standing one up just for this would be disproportionate to a one-line fix.

3. **The fix mirrors an already-proven path.** `flushRequests` now completes queued entries with `request.completedEx(ex)`, identical to the sibling `setException` path (which iterates the same `this._requests` queue and calls `request.completedEx(...)`). The queue holds the outgoing-async objects themselves (pushed via `this._requests.push(out)`), so `request.out` was always `undefined` — the old code could never have worked on this branch.

The change was verified against the connection-establishment-heavy `Ice/ami`, `Ice/exceptions`, and `Ice/binding` suites, which exercise queued-request flushing on connection setup. CI is green.

bernardnormier (approved):

PR #5577, opened 2026-06-13
Add bounds check in InputStream.skipEncapsulation (JS)

pepone:


bernardnormier (approved):

js/packages/test/Ice/stream/Client.ts
```diff
@@ -544,6 +544,37 @@ export class Client extends TestHelper {
             test(dict2.get("key2")!.s.e == Test.MyEnum.enum3);
         }
         out.writeLine("ok");
+
+        out.write("testing skipEncapsulation... ");
```
InsertCreativityHere, line 548: I don't know how worth it adding this test is. Especially since we don't have any corresponding tests for `skipEmptyEncapsulation`, `readEncapsulation`, or `startEncapsulation` which all have the exact same check that we neglected to do for `skipEncapsulation`.

InsertCreativityHere (approved):
Looks good to me.

js/packages/test/Ice/stream/Client.ts
pepone: I think is worth keeping, we can add additional tests for other untested methods in a follow-up PR.

PR #5578, opened 2026-06-13
Fix Ice for Ruby marshaling of an empty sequence<byte> passed as a string

### LINKED ISSUE #5538: [Audit-High] IceRuby: empty sequence<byte> given as a Ruby String writes a raw 4-byte int instead of writeSize(0) — desyncs the rest of the message
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Ruby source (IceRuby C++ extension + Ruby runtime) performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

In IceRuby's `SequenceInfo::marshalPrimitiveSequence`, an **empty** Ruby `String` supplied for a `sequence<byte>` writes `os->write(int32_t(0))` — the raw 4-byte little-endian int overload — instead of `writeSize(0)` (a single `0x00` size byte). The reader consumes a size-prefixed sequence: it reads size 0 and leaves **three stray `0x00` bytes** in the stream, desynchronizing every field that follows in the message.

## The bug

```cpp
// ruby/src/IceRuby/Types.cpp:1348-1361 (KindByte)
if (!NIL_P(str))
{
    const char* s = RSTRING_PTR(str);
    const long len = RSTRING_LEN(str);
    if (s == 0 || len == 0)
    {
        os->write(int32_t(0));          // BUG — raw 4-byte int, not a size
    }
    else
    {
        os->write(reinterpret_cast<const byte*>(s), reinterpret_cast<const byte*>(s + len));  // size-prefixed
    }
}
```

The non-empty sibling uses the byte-range overload, which writes `writeSize(sz)` followed by the bytes — so the correct empty encoding is `writeSize(0)`. The Python port always uses the byte-range overload and PHP writes `writeSize(0)`, confirming the intent.

## Impact

Marshaling any `sequence<byte>` whose Ruby value is `""` (operation parameter, struct/class member, dictionary value) deterministically emits three extra bytes; anything following it in the message is decoded from the wrong offset — silent corruption or a misleading `MarshalException` at the peer.

## Source references

- **(the wrong overload)** — [`Types.cpp#L1348-L1361`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/ruby/src/IceRuby/Types.cpp#L1348-L1361)
- **(byte-range overload writes a size first)** — [`cpp/src/Ice/OutputStream.cpp#L359-L368`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/OutputStream.cpp#L359-L368)
- **(reader consumes a size-prefixed sequence)** — [`Types.cpp#L1505-L1511`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/ruby/src/IceRuby/Types.cpp#L1505-L1511), [`cpp/src/Ice/InputStream.cpp#L488-L500`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/InputStream.cpp#L488-L500)
- **(Python/PHP siblings)** — [`python/modules/IcePy/Types.cpp#L1604-L1612`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L1604-L1612), [`php/src/Types.cpp#L1474-L1477`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/php/src/Types.cpp#L1474-L1477)

## Recommendation

`os->writeSize(0);`

## Severity

**High** — deterministic wire corruption for a perfectly ordinary value (`""`); one-token fix. Verified **CONFIRMED** by the second-model pass; ranked first in both priority lists.

pepone:
Fixes #5538.

In `SequenceInfo::marshalPrimitiveSequence`, an empty Ruby `String` supplied for a `sequence<byte>` was marshaled with `os->write(int32_t(0))` (a raw 4-byte little-endian integer) instead of `os->writeSize(0)` (a single size byte). The three extra bytes desynchronized the rest of the message, producing a `MarshalException` (or silent corruption) for any value that follows it.

The fix matches the non-empty path (the byte-range `write` overload writes the size first) and the Python/PHP siblings.

### Test

Added a regression test to `ruby/test/Ice/operations` (`Twoways.rb`): `opByteS("", bsi2)` — an empty `p1` supplied as a string followed by a non-empty `p2`. It fails before the fix (the stray bytes desync `p2`) and passes after.

Milestone: 3.8.3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,9 @@ might need to be aware of.
   asymmetric, and the class defined `eql?` without a matching `hash`; equal endpoints now compare consistently and
   can be used as `Hash`/`Set` keys.
 
+- Fixed Ice for Ruby to correctly marshal an empty `sequence<byte>` supplied as a string: it now writes a single size
+  byte instead of a raw 4-byte integer, which previously desynchronized the rest of the message.
+
 These are the changes since the [Ice 3.8.2] release.
 
 [Ice 3.8.2]: https://github.com/zeroc-ice/ice/blob/3.8/CHANGELOG-3.8.md
--- a/ruby/test/Ice/operations/Twoways.rb
+++ b/ruby/test/Ice/operations/Twoways.rb
@@ -325,6 +325,13 @@ def twoways(helper, communicator, p)
     test(rso.unpack("C*") == arr)
 
     #
+    # opByteS with an empty sequence supplied as an empty string.
+    #
+    rso, bso = p.opByteS("", bsi2)
+    test(bso.length == 0)
+    test(rso.unpack("C*") == bsi2)
+
+    #
     # opBoolS
     #
     bsi1 = [true, true, false]
--- a/ruby/src/IceRuby/Types.cpp
+++ b/ruby/src/IceRuby/Types.cpp
@@ -1352,7 +1352,7 @@ IceRuby::SequenceInfo::marshalPrimitiveSequence(const PrimitiveInfoPtr& pi, VALU
                 const long len = RSTRING_LEN(str);
                 if (s == 0 || len == 0)
                 {
-                    os->write(int32_t(0));
+                    os->writeSize(0);
                 }
                 else
                 {


CHANGELOG.md
```diff
@@ -6,9 +6,15 @@ comprehensive report of every change we made in a release, it does provide detai
...
+  - [General Changes](#general-changes)
 
 ## Changes in Ice 3.9.0
 
 These are the changes since the [Ice 3.8.2] release.
 
+### General Changes
+
+- Fixed Ice for Ruby to correctly marshal an empty `sequence<byte>` supplied as a string: it now writes a single size
+  byte instead of a raw 4-byte integer, which previously desynchronized the rest of the message.
```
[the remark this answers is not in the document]
pepone: This is fine, until we cherry pick fixes in 3.8 changelog entries stay in 3.9 main changelog.

ruby/test/Ice/operations/Twoways.rb
```diff
@@ -324,6 +324,16 @@ def twoways(helper, communicator, p)
     test(rso.length == 8)
     test(rso.unpack("C*") == arr)
 
+    #
+    # opByteS with an empty sequence supplied as an empty string. Regression test: an empty
+    # Ruby String marshaled as a sequence<byte> must emit writeSize(0) (a single size byte),
+    # not a raw 4-byte int. Passing an empty p1 followed by a non-empty p2 ensures the bug's
+    # 3 stray bytes would desync and corrupt p2, so the round-trip must return p2 intact.
```
bernardnormier: I would remove the Regression test: ... part of the comment. It was just a bug.

bernardnormier (approved):

ruby/test/Ice/operations/Twoways.rb
InsertCreativityHere: Agreed, describing the old behavior is more confusing than anything IMO.

InsertCreativityHere (approved):

ruby/test/Ice/operations/Twoways.rb
pepone: 🤖 **Claude:** Done — trimmed the comment to just `# opByteS with an empty sequence supplied as an empty string.`, dropping the old-behavior description. Also rebased onto `origin/main` (was 43 behind): `Types.cpp` auto-merged cleanly with the `%d` enum fix, and the changelog entry now lives in the existing **Ruby Changes** section. Verified IceRuby `Types.cpp` still compiles. (b5c5de8a01)

PR #5583, opened 2026-06-13
Fix slice2rb duplicate context parameter for a Context-named parameter

### LINKED ISSUE #5482: slice2rb: duplicate 'context' parameter when an in-parameter is named Context — Ruby SyntaxError
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Ice C++ source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source (including against generated `slice2rb` output) by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; please verify the reasoning, reproducibility, and severity before acting on it.

## Summary

In `slice2rb`, the parameter-name collision check compares the *unmodified* identifier against `"context"`, but the method signature emits the *lower-cased* identifier. A Slice parameter named `Context` is emitted as `context`, the clash with the synthesized `context` parameter goes undetected, and the generator produces `def op(context, context=nil)` — a Ruby `SyntaxError`, making the generated file unloadable.

## The bug

```cpp
// cpp/src/slice2rb/RubyUtil.cpp:18-28  (collision check uses IdentNormal)
// signature emits IdentToLower (cpp/src/slice2rb/RubyUtil.cpp:308; used at :317-326)
```

The check sees `Context`, which doesn't equal `"context"`; the emitted name is `context`, which does.

## Impact

`void op(string Context);` generates `def op(context, context=nil)` — a Ruby `SyntaxError`; the generated file cannot be loaded.

## Source references

- **(slice2rb / collision check vs lowercased emission)** — [`cpp/src/slice2rb/RubyUtil.cpp#L18-L28`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2rb/RubyUtil.cpp#L18-L28)

## Recommendation

Perform the `"context"` collision check against the lower-cased identifier that is actually emitted (or escape the emitted name on collision).

## Severity

**Medium** — a Slice parameter named `Context` generates a Ruby `SyntaxError` (unloadable file).

pepone:
Fixes #5482.

In `slice2rb`, `getEscapedRubyParamName` decides whether to escape the synthesized trailing `context` parameter of a generated proxy method. The check compared the *unmodified* identifier against `"context"` over *all* parameters, but in-parameters are emitted in the signature with their **lower-cased** mapped name. So a Slice parameter named `Context` was emitted as `context`, the clash with the synthesized `context` parameter went undetected, and the generator produced:

```ruby
def op(context, context=nil)   # Ruby SyntaxError: duplicated argument name
```

making the generated file unloadable.

The fix limits the check to **in-parameters** (the only ones in the signature) and compares against the same **lower-cased** name they are emitted with — i.e. it mirrors the emission loop exactly. This also removes a pre-existing spurious escape: an operation whose only `context`/`Context` parameter is an *out*-parameter no longer escapes the synthesized name unnecessarily.

Verified with Ruby reflection on the generated proxy:

| operation | generated params |
| --- | --- |
| `op(string context, …)` (in-param) | `[:context, …, :context_]` (unchanged) |
| `opOut(out string context, …)` (out-only) | `[:context]` (was spuriously `:context_`) |
| `opClashContext(string Context)` (in-param) | `[:context, :context_]` (**fixed**) |

### Test

Added `void opClashContext(string Context)` to `ruby/test/Slice/escape/Clash.ice`. The escape test compiles it in-process via `Ice::loadSlice`, so the duplicated-argument `SyntaxError` fails the test at load before the fix and passes after — no new assertion code needed.

Milestone: 3.8.3.

### DIFF
--- a/cpp/src/slice2rb/RubyUtil.cpp
+++ b/cpp/src/slice2rb/RubyUtil.cpp
@@ -15,16 +15,19 @@ using namespace IceInternal;
 
 namespace
 {
-    string getEscapedRubyParamName(const OperationPtr& p, const string& name)
+    // Returns the name of the synthesized trailing 'context' parameter for a generated proxy method. It is escaped
+    // to 'context_' when an in-parameter's lower-cased mapped name is already 'context' (e.g. a parameter named
+    // 'Context'), to avoid emitting a duplicate parameter name in the method signature.
+    string getEscapedContextParam(const OperationPtr& p)
     {
-        for (const auto& param : p->parameters())
+        for (const auto& param : p->inParameters())
         {
-            if (Slice::Ruby::getMappedName(param) == name)
+            if (Slice::Ruby::getMappedName(param, Slice::Ruby::IdentToLower) == "context")
             {
-                return name + "_";
+                return "context_";
             }
         }
-        return name;
+        return "context";
     }
 }
 
@@ -319,7 +322,7 @@ Slice::Ruby::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
         {
             _out << inParams << ", ";
         }
-        const string contextParamName = getEscapedRubyParamName(op, "context");
+        const string contextParamName = getEscapedContextParam(op);
         _out << contextParamName << "=nil)";
         _out.inc();
         _out << nl << proxyName << "_mixin::OP_" << op->name() << ".invoke(self, [" << inParams;
--- a/ruby/test/Slice/escape/Clash.ice
+++ b/ruby/test/Slice/escape/Clash.ice
@@ -20,6 +20,10 @@ module Clash
         void opOut(out string context, out string current, out string response, out string ex,
             out string sent, out string cookie, out string sync, out string result, out string istr,
             out string ostr, out optional(1) string proxy);
+
+        // A parameter whose mapped name lower-cases to "context" must not clash with the synthesized
+        // trailing "context" parameter of the generated proxy method.
+        void opClashContext(string Context);
     }
 
     class Cls


cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -19,7 +19,10 @@ namespace
     {
         for (const auto& param : p->parameters())
```
pepone: We should just loop over the in-params, we already have a method to get them.

CHANGELOG.md
```diff
@@ -6,9 +6,16 @@ comprehensive report of every change we made in a release, it does provide detai
...
 - [Changes in Ice 3.9.0](#changes-in-ice-390)
+  - [General Changes](#general-changes)
 
 ## Changes in Ice 3.9.0
 
 These are the changes since the [Ice 3.8.2] release.
 
+### General Changes
+
+- Fixed `slice2rb` to escape an operation parameter whose mapped name lower-cases to `context` (such as a parameter
```
pepone: This should go into a "Ruby Changes" section see the changelog in 3.8 branch for reference.

cpp/src/slice2rb/RubyUtil.cpp
pepone: Done in 505a4a2 — the loop now iterates `p->inParameters()` directly instead of walking all parameters with a `!isOutParam()` filter.

CHANGELOG.md
pepone: Done in 505a4a2 — moved to a new "Ruby Changes" section (with a TOC entry), matching the per-language sections in CHANGELOG-3.8.md.

CHANGELOG.md
```diff
@@ -6,9 +6,16 @@ comprehensive report of every change we made in a release, it does provide detai
...
+  named `Context`). It previously clashed with the synthesized trailing `context` parameter of the generated proxy
+  method, producing a Ruby `SyntaxError` that made the generated file unloadable.
```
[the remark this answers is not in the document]
pepone: Reworded in 505a4a2: it now describes escaping the synthesized trailing `context` parameter (renaming it to `context_`) when an operation has an in-parameter whose mapped name lower-cases to `context`, rather than 'an operation parameter'.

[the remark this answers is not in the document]
pepone: Left under "Changes in Ice 3.9.0": main is 3.9a0, so changes landing here are documented in the 3.9 changelog. The fix is also milestoned 3.8.3, and that entry goes into CHANGELOG-3.8.md when it is cherry-picked to the 3.8 branch — so the milestone and the main changelog section don't conflict.

CHANGELOG.md
```diff
@@ -6,9 +6,17 @@ comprehensive report of every change we made in a release, it does provide detai
...
 - [Changes in Ice 3.9.0](#changes-in-ice-390)
+  - [Ruby Changes](#ruby-changes)
 
 ## Changes in Ice 3.9.0
 
 These are the changes since the [Ice 3.8.2] release.
 
+### Ruby Changes
+
+- Fixed `slice2rb` to escape the synthesized trailing `context` parameter of a generated proxy method (renaming it
```
bernardnormier: I would not include any changelog entry for this bug fix. Pascal case is not a common naming convention for Slice parameters.

bernardnormier (approved):

cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -17,9 +17,12 @@ namespace
 {
     string getEscapedRubyParamName(const OperationPtr& p, const string& name)
```
InsertCreativityHere: The logic looks good, but I think there's an issue with the _design_ of this change.

In theory, `getEscapedRubyParamName` is a general purpose escaping method, and is not specific to `context` as the comment suggests nor should be specific to `inParameters`.

In reality, after removing most of our old escaping logic, yes, this function is only called for `context` now. I'd be fine renaming this function to `getEscapedContextParam`, or maybe even just inlining it.
But I don't think it should be kept as is. Otherwise it gives the impression it works for any param you want to escape, when this is no longer the case.

pepone: 🤖 **Claude:** Renamed `getEscapedRubyParamName` → `getEscapedContextParam` and dropped the unused `name` parameter — it was only ever called with "context", so the general-purpose name/signature no longer apply. Rebuilt slice2rb and confirmed the escaping still works (`opClashContext(context, context_=nil)`, generated file passes `ruby -c`). (a9d59e4)

CHANGELOG.md
pepone: 🤖 **Claude:** Done — removed the changelog entry (and the now-empty `### Ruby Changes` section + its TOC line). Also rebased the branch onto `origin/main` to clear the conflict; the PR now carries only the slice2rb fix + the Slice test. (a9d59e4)

PR #5584, opened 2026-06-13
Fix IcePHP marshaling of non-empty optional string sequences

### LINKED ISSUE #5532: [Audit-High] IcePHP: non-empty optional seq<string> skips endSize() — unpatched zero FSize corrupts the encapsulation
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the PHP source (IcePHP C++ extension + PHP runtime) performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

In IcePHP's `SequenceInfo::marshal`, a **non-empty optional/tagged string sequence** writes its 4-byte FSize placeholder via `startSize()` and then takes the primitive fast path, which `return`s before the `endSize()` tail that patches the placeholder. The emitted FSize stays zero, producing a malformed encapsulation. `KindString` is the **only** primitive whose `variableLength()` returns true, so optional `seq<string>` is exactly the affected case.

## The bug

```cpp
// php/src/Types.cpp:1461-1507 (SequenceInfo::marshal)
Ice::OutputStream::size_type sizePos = 0;
if (optional)
{
    if (elementType->variableLength())
    {
        sizePos = os->startSize();           // writes a zero 4-byte FSize placeholder
    }
    ...
}
...
else
{
    PrimitiveInfoPtr pi = dynamic_pointer_cast<PrimitiveInfo>(elementType);
    if (pi)
    {
        marshalPrimitiveSequence(pi, zv, os);
        return;                              // BUG — skips endSize()
    }
    ...
}

if (optional && elementType->variableLength())
{
    os->endSize(sizePos);                    // never reached for a non-empty string sequence
}
```

The IcePy port has the same structure but **falls through** to `endSize` ([`python/modules/IcePy/Types.cpp#L1302-L1335`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/python/modules/IcePy/Types.cpp#L1302-L1335)), confirming the intended control flow. The empty case (`sz == 0`) doesn't take the early return and is unaffected.

## Impact

Marshaling a non-empty optional/tagged `seq<string>` parameter or data member (e.g. `optional(1) Ice::StringSeq` set to `["a"]`) emits an invalid zero FSize. Per the verification pass: an in-sync reader that knows the field can still skip the 4 FSize bytes and decode the body, but an older reader that must **skip the unknown optional** uses the zero FSize and loses framing — breaking Slice version tolerance and corrupting the decode of everything after the field.

## Source references

- **(the marshal path)** — [`Types.cpp#L1449-L1507`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/php/src/Types.cpp#L1449-L1507)
- **(`KindString` is the only variable-length primitive)** — [`Types.cpp#L716-L720`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/php/src/Types.cpp#L716-L720)
- **(`startSize`/`endSize` placeholder contract)** — [`cpp/include/Ice/OutputStream.h#L238-L250`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/include/Ice/OutputStream.h#L238-L250)

## Recommendation

Remove the early `return` (restructure so the primitive fast path falls through to the `endSize` tail), mirroring the IcePy structure.

## Severity

**High** — a wire-format corruption on a documented optional-parameter shape. Verified **CONFIRMED** by the second-model pass (with the "every peer fails" trigger narrowed to unknown-optional skipping, which doesn't change the priority).

pepone:
Fixes #5532.

`IcePHP::SequenceInfo::marshal` took a primitive fast path that returned before the trailing `endSize()`, so a non-empty optional/tagged `seq<string>` left the 4-byte FSize placeholder written by `startSize()` as zero. A peer that must skip the field as an _unknown_ optional then uses that zero size and loses framing, corrupting the rest of the encapsulation. The primitive path now falls through to `endSize()`, mirroring the IcePy implementation. `seq<string>` is the only affected case, since `KindString` is the only variable-length primitive.

### Testing

Verified by structural parity with the IcePy reference (identical control flow apart from the removed early `return`) and an independent codex audit; the existing `optional` test (which round-trips an optional `StringSeq` between matched peers) still passes.

The zero FSize is only observable on the unknown-optional **skip** path: a peer that doesn't know the tag uses the FSize to skip the field, and the mis-skip only breaks framing when there is data after it. In the optional test that path is reachable solely through the class-parameter `opClassAndUnknownOptional`/`Initial2` hook (a non-empty optional `seq<string>` there does reproduce it), so a dedicated check would either overload that class-versioning operation or require a new cross-language operation. For a one-line, IcePHP-only wire fix that exactly mirrors the correct IcePy code, we rely on that parity plus the codex audit instead.

Addresses the AI-generated audit finding in #5532, re-verified against the source.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ might need to be aware of.
   - [C# Changes](#c-changes-1)
   - [JavaScript Changes](#javascript-changes)
   - [MATLAB Changes](#matlab-changes)
+  - [PHP Changes](#php-changes)
   - [Python Changes](#python-changes)
   - [Ruby Changes](#ruby-changes)
   - [Ice Service Changes](#ice-service-changes)
@@ -75,6 +76,10 @@ might need to be aware of.
 
 - Fixed `Ice.Future.wait('sent')`, which could block indefinitely or time out even after the request had been
   sent. The wait now completes as soon as the invocation reaches or passes the requested state.
+
+### PHP Changes
+
+- Fixed Ice for PHP mis-marshaling a non-empty optional `sequence<string>`.
 
 ### Python Changes
 
--- a/php/src/Types.cpp
+++ b/php/src/Types.cpp
@@ -1458,32 +1458,31 @@ IcePHP::SequenceInfo::marshal(zval* zv, Ice::OutputStream* os, ObjectMap* object
         sz = static_cast<int32_t>(zend_hash_num_elements(arr));
     }
 
+    // An optional sequence whose elements are variable-length is written as an FSize-framed field:
+    // startSize() reserves the 4-byte size and endSize() patches it. The two calls are driven by a
+    // single flag so they can never get out of sync, regardless of how the body below is structured.
+    const bool fSize = optional && elementType->variableLength();
+
     Ice::OutputStream::size_type sizePos = 0;
-    if (optional)
+    if (fSize)
     {
-        if (elementType->variableLength())
-        {
-            sizePos = os->startSize();
-        }
-        else if (elementType->wireSize() > 1)
-        {
-            os->writeSize(sz == 0 ? 1 : sz * elementType->wireSize() + (sz > 254 ? 5 : 1));
-        }
+        sizePos = os->startSize();
+    }
+    else if (optional && elementType->wireSize() > 1)
+    {
+        os->writeSize(sz == 0 ? 1 : sz * elementType->wireSize() + (sz > 254 ? 5 : 1));
     }
 
     if (sz == 0)
     {
         os->writeSize(0);
     }
+    else if (PrimitiveInfoPtr pi = dynamic_pointer_cast<PrimitiveInfo>(elementType))
+    {
+        marshalPrimitiveSequence(pi, zv, os);
+    }
     else
     {
-        PrimitiveInfoPtr pi = dynamic_pointer_cast<PrimitiveInfo>(elementType);
-        if (pi)
-        {
-            marshalPrimitiveSequence(pi, zv, os);
-            return;
-        }
-
         os->writeSize(sz);
 
         zval* val;
@@ -1491,9 +1490,9 @@ IcePHP::SequenceInfo::marshal(zval* zv, Ice::OutputStream* os, ObjectMap* object
         {
             if (!elementType->validate(val, false))
             {
-                ostringstream os;
-                os << "invalid value for sequence element '" << id << "'";
-                invalidArgument(os.str());
+                ostringstream ostr;
+                ostr << "invalid value for sequence element '" << id << "'";
+                invalidArgument(ostr.str());
                 throw AbortMarshaling();
             }
             elementType->marshal(val, os, objectMap, false);
@@ -1501,7 +1500,7 @@ IcePHP::SequenceInfo::marshal(zval* zv, Ice::OutputStream* os, ObjectMap* object
         ZEND_HASH_FOREACH_END();
     }
 
-    if (optional && elementType->variableLength())
+    if (fSize)
     {
         os->endSize(sizePos);
     }


bernardnormier:
The fix is correct, but I'd go further and simplify this function. The root issue is that `startSize()` and `endSize()` must be paired, yet the same `optional && elementType->variableLength()` condition is evaluated in two far-apart places with a multi-branch body in between — which is exactly what let an early `return` slip past `endSize()` in the first place. Removing the `return` fixes the symptom, but the structure that allowed it is still there.

Suggested rewrite — drive both calls from a single flag and flatten the body so no early exit can sit between the open and close:

```cpp
void
IcePHP::SequenceInfo::marshal(zval* zv, Ice::OutputStream* os, ObjectMap* objectMap, bool optional)
{
    int32_t sz = 0;
    HashTable* arr = 0;
    if (Z_TYPE_P(zv) != IS_NULL)
    {
        assert(Z_TYPE_P(zv) == IS_ARRAY); // validate() should have caught this.
        arr = Z_ARRVAL_P(zv);
        sz = static_cast<int32_t>(zend_hash_num_elements(arr));
    }

    // An optional sequence whose elements are variable-length is written as an FSize-framed field:
    // startSize() reserves the 4-byte size and endSize() patches it. The two calls are driven by a
    // single flag so they can never get out of sync, regardless of how the body below is structured.
    const bool fSize = optional && elementType->variableLength();

    Ice::OutputStream::size_type sizePos = 0;
    if (fSize)
    {
        sizePos = os->startSize();
    }
    else if (optional && elementType->wireSize() > 1)
    {
        os->writeSize(sz == 0 ? 1 : sz * elementType->wireSize() + (sz > 254 ? 5 : 1));
    }

    if (sz == 0)
    {
        os->writeSize(0);
    }
    else if (PrimitiveInfoPtr pi = dynamic_pointer_cast<PrimitiveInfo>(elementType))
    {
        marshalPrimitiveSequence(pi, zv, os);
    }
    else
    {
        os->writeSize(sz);

        zval* val;
        ZEND_HASH_FOREACH_VAL(arr, val)
        {
            if (!elementType->validate(val, false))
            {
                ostringstream ostr;
                ostr << "invalid value for sequence element '" << id << "'";
                invalidArgument(ostr.str());
                throw AbortMarshaling();
            }
            elementType->marshal(val, os, objectMap, false);
        }
        ZEND_HASH_FOREACH_END();
    }

    if (fSize)
    {
        os->endSize(sizePos);
    }
}
```

This is wire-identical to the current fix across every path (empty / primitive / generic, optional/non-optional, fixed vs. variable element). Changes:

- A single `fSize` flag drives both `startSize` and `endSize`, so the open/close condition can't drift.
- Flattened `if / else if / else` for the payload — no nested `if (pi)` inside `else`, and no `return` in the middle that can bypass the close.
- Renamed the inner `ostringstream os` to `ostr` so it no longer shadows the `OutputStream* os` parameter (harmless today, but a trap if anyone later adds an `os->...` call in that block).

One tradeoff: this diverges cosmetically from the IcePy reference the PR mirrors (still wire-identical). If strict parity matters, the `fSize` hoist is the part worth keeping; the rename is optional.

(I also considered an RAII guard for the start/end pair — the textbook answer — but `endSize()` would then run during stack unwinding on `AbortMarshaling`, and a throwing destructor risks `std::terminate`; it'd also be inconsistent with the sibling mappings. The single-flag form gets the safety without that hazard.)

bernardnormier (commented):
We should add a CHANGELOG entry for this PR.

pepone:
🤖 **Claude:** @bernardnormier added a **PHP Changes** CHANGELOG entry for this fix (user-facing description of the optional `sequence<string>` marshaling bug), and rebased the branch onto `origin/main` (was 43 behind). The PR now carries the IcePHP marshaling fix + the changelog entry. (502fef5a)

pepone:
🤖 **Claude:** Applied your rewrite verbatim — single `fSize` flag pairing `startSize()`/`endSize()`, flattened `if/else if/else` payload (no nested `if (pi)`, no early return between open and close), and `ostringstream os`→`ostr` (which also clears Copilot's shadowing note). Confirmed IcePHP `Types.cpp` compiles, and double-checked wire-equivalence across every path (empty / primitive / generic, optional fixed vs variable) with codex — confirmed identical. Agreed on skipping the RAII guard given the throwing-destructor-during-`AbortMarshaling`-unwind hazard. (6d3d36bb89)

CHANGELOG.md
```diff
@@ -54,6 +55,11 @@ might need to be aware of.
 - Fixed `ice_getCachedConnection` to return an empty `Ice.Connection` array, rather than an empty
   `double` array, when the proxy has no cached connection.
 
+### PHP Changes
+
+- Fixed Ice for PHP mis-marshaling a non-empty optional sequence of variable-length elements (such as an optional
+  `sequence<string>`): the optional value's size was left unpatched, producing a request the peer could not decode.
```
bernardnormier: Tightening to the exact affected case: only `seq<string>` is actually mis-marshaled (it's the sole variable-length primitive — `PrimitiveInfo::variableLength()` is `kind == KindString`). Variable-length *non-primitive* elements never take the buggy early return, so the broader "variable-length elements" phrasing is imprecise.

```suggestion
- Fixed Ice for PHP mis-marshaling a non-empty optional `sequence<string>`.
```

bernardnormier (approved):
LGTM. The `fSize` flag pairs `startSize()`/`endSize()` so they can't desync, and dropping the early `return` on the primitive path means a non-empty optional `seq<string>` now patches its FSize. All other branches are unchanged, and it mirrors IcePy. One CHANGELOG simplification inline.

PR #5586, opened 2026-06-13
Fix IcePHP setBufferSize validation and UDP ConnectionInfo properties

### LINKED ISSUE #5534: [Audit-Medium] IcePHP Connection::setBufferSize reads Z_LVAL_P on unvalidated "zz" arguments — non-int input applies a garbage buffer size
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the PHP source (IcePHP C++ extension + PHP runtime) performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

`Ice_Connection::setBufferSize` parses its two arguments with the `"zz"` spec (any zval, no coercion) and then reads both unconditionally with `Z_LVAL_P`. For a non-int argument (string, array, object, double, bool), `Z_LVAL_P` reads the wrong zval union member — for a string, the `zend_string*` pointer reinterpreted as a `zend_long` — and the garbage value is passed to the C++ `Connection::setBufferSize`.

## The bug

```cpp
// php/src/Connection.cpp:222-236
zval* r;
zval* s;
if (zend_parse_parameters(ZEND_NUM_ARGS(), const_cast<char*>("zz"), &r, &s) != SUCCESS)
{
    RETURN_NULL();
}

int rcvSize = static_cast<int>(Z_LVAL_P(r));   // r/s may be any type
int sndSize = static_cast<int>(Z_LVAL_P(s));
```

The sibling directly above does it right: `flushBatchRequests` parses `"z"` and guards `if (Z_TYPE_P(compress) != IS_LONG) { invalidArgument(...); RETURN_NULL(); }` before reading `Z_LVAL_P` ([`Connection.cpp#L137-L168`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/php/src/Connection.cpp#L137-L168)). No `php/lib` wrapper validates the arguments first.

## Impact

`$connection->setBufferSize("foo", [])` applies an arbitrary buffer size to the connection instead of raising `InvalidArgumentException`. No out-of-bounds access — the read stays within the zval union — but the resulting value is garbage.

## Source references

- **(unchecked `Z_LVAL_P`)** — [`Connection.cpp#L217-L245`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/php/src/Connection.cpp#L217-L245)
- **(guarded sibling)** — [`Connection.cpp#L137-L168`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/php/src/Connection.cpp#L137-L168)

## Recommendation

Parse with `"ll"` into `zend_long` locals, or add `Z_TYPE_P(...) != IS_LONG` guards for both arguments, mirroring `flushBatchRequests`.

## Severity

**Medium** — a public API silently accepts any argument type and applies a garbage value. Verified **CONFIRMED** by the second-model pass.

### LINKED ISSUE #5536: [Audit-Low] IcePHP UDP ConnectionInfo sets rcvSize/sndSize as undeclared dynamic properties (deprecated in PHP 8.2+)
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the PHP source (IcePHP C++ extension + PHP runtime) performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it.

## Summary

`rcvSize` and `sndSize` are declared as class properties only on `Ice\TCPConnectionInfo`. The UDP branch of `connectionInfoToZval` nevertheless sets both on `Ice\UDPConnectionInfo` instances, creating **dynamic properties on an internal class** — a deprecated path since PHP 8.2 (deprecation notices; incompatible with `#[\AllowDynamicProperties]`-less classes going forward).

## The bug

```cpp
// php/src/Connection.cpp:533-542 (UDP branch)
add_property_string(zv, "mcastAddress", ...);
add_property_long(zv, "mcastPort", ...);
add_property_long(zv, "rcvSize", static_cast<long>(info->rcvSize));   // not declared on the UDP class
add_property_long(zv, "sndSize", static_cast<long>(info->sndSize));   // not declared on the UDP class
```

The declarations exist only for TCP ([`Connection.cpp#L417-L434`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/php/src/Connection.cpp#L417-L434)); `udpConnectionInfoClassEntry` derives from `ipConnectionInfoClassEntry`, which doesn't carry them. The C++ `Ice::UDPConnectionInfo` does have both fields, so the intent is right — only the property declarations are missing.

## Impact

Reading `$info->rcvSize` on a UDP connection info works but goes through the deprecated dynamic-property path on PHP ≥ 8.2, and the published class schema is inconsistent with the actual instance shape.

## Source references

- **(UDP branch sets undeclared properties)** — [`Connection.cpp#L533-L542`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/php/src/Connection.cpp#L533-L542)
- **(declarations TCP-only)** — [`Connection.cpp#L417-L434`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/php/src/Connection.cpp#L417-L434)

## Recommendation

Declare `rcvSize`/`sndSize` on `udpConnectionInfoClassEntry` in `connectionInit`, mirroring the TCP entry.

## Severity

**Low** — schema inconsistency / deprecation-path usage, values are correct at runtime. Verified **CONFIRMED** by the second-model pass.

pepone:
Fixes #5534. Fixes #5536.

Two correctness fixes in the IcePHP `Connection` code:

- **#5534** — `Connection::setBufferSize` parsed its arguments with `"zz"` and then read `Z_LVAL_P` unconditionally, so a non-integer argument applied a garbage buffer size instead of raising an error. It now parses `"ll"` into `zend_long`, so the Zend engine rejects non-numeric input (raising `TypeError`) instead of reading a garbage value; numeric strings and floats are still coerced as an int-typed parameter would be, so legitimate callers are unaffected. This brings it in line with IcePy and the sibling `flushBatchRequests`, which also reject non-integer input.
- **#5536** — the UDP branch of `createConnectionInfo` sets `rcvSize`/`sndSize`, but those properties were declared only on `TCPConnectionInfo`, so on UDP they were created as dynamic properties (deprecated on PHP 8.2+). They are now declared on `UDPConnectionInfo`, mirroring the TCP entry.

### Testing

Adds `info`-test coverage for both: a negative `setBufferSize` call that must raise `TypeError`, and assertions that `rcvSize`/`sndSize` are declared properties on `Ice\UDPConnectionInfo`. Confirmed red→green locally (the red run showed the garbage size being applied). Independently re-checked with codex.

Addresses the AI-generated audit findings in #5534 and #5536, re-verified against the source.

### DIFF
--- a/php/src/Connection.cpp
+++ b/php/src/Connection.cpp
@@ -224,15 +224,23 @@ ZEND_METHOD(Ice_Connection, setBufferSize)
     Ice::ConnectionPtr _this = Wrapper<Ice::ConnectionPtr>::value(getThis());
     assert(_this);
 
-    zval* r;
-    zval* s;
-    if (zend_parse_parameters(ZEND_NUM_ARGS(), const_cast<char*>("zz"), &r, &s) != SUCCESS)
+    zend_long r;
+    zend_long s;
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), const_cast<char*>("ll"), &r, &s) != SUCCESS)
     {
         RETURN_NULL();
     }
 
-    int rcvSize = static_cast<int>(Z_LVAL_P(r));
-    int sndSize = static_cast<int>(Z_LVAL_P(s));
+    // Connection::setBufferSize takes int arguments in C++; reject values outside that range instead of silently
+    // truncating them.
+    if (r < INT_MIN || r > INT_MAX || s < INT_MIN || s > INT_MAX)
+    {
+        invalidArgument("buffer size must be in the range of a 32-bit signed integer");
+        RETURN_NULL();
+    }
+
+    int rcvSize = static_cast<int>(r);
+    int sndSize = static_cast<int>(s);
 
     try
     {
@@ -432,6 +440,8 @@ IcePHP::connectionInit(void)
         "",
         ZEND_ACC_PUBLIC);
     zend_declare_property_long(udpConnectionInfoClassEntry, "mcastPort", sizeof("mcastPort") - 1, 0, ZEND_ACC_PUBLIC);
+    zend_declare_property_long(udpConnectionInfoClassEntry, "rcvSize", sizeof("rcvSize") - 1, 0, ZEND_ACC_PUBLIC);
+    zend_declare_property_long(udpConnectionInfoClassEntry, "sndSize", sizeof("sndSize") - 1, 0, ZEND_ACC_PUBLIC);
 
     // Register the WSConnectionInfo class.
     INIT_NS_CLASS_ENTRY(ce, "Ice", "WSConnectionInfo", nullptr);
--- a/php/src/Connection.cpp
+++ b/php/src/Connection.cpp
@@ -235,7 +235,7 @@ ZEND_METHOD(Ice_Connection, setBufferSize)
     // truncating them.
     if (r < INT_MIN || r > INT_MAX || s < INT_MIN || s > INT_MAX)
     {
-        invalidArgument("buffer size must be in the range of a 32-bit signed integer");
+        invalidArgument("buffer size must be in the range of a C++ int");
         RETURN_NULL();
     }
 
--- a/php/test/Ice/info/Client.php
+++ b/php/test/Ice/info/Client.php
@@ -104,6 +104,20 @@ function allTests($helper)
         $connection = $testIntf->ice_getConnection();
         $connection->setBufferSize(1024, 2048);
 
+        // setBufferSize must reject non-integer arguments instead of applying a garbage size (see #5534).
+        try {
+            $connection->setBufferSize("not an int", 2048);
+            test(false);
+        } catch (TypeError $ex) {
+        }
+
+        // It must also reject values outside the C++ int range instead of silently truncating them (see #5534).
+        try {
+            $connection->setBufferSize(2 ** 40, 2048);
+            test(false);
+        } catch (InvalidArgumentException $ex) {
+        }
+
         $info = $connection->getInfo();
         $tcpinfo = getTCPConnectionInfo($info);
         test($tcpinfo instanceof Ice\TCPConnectionInfo);
@@ -139,6 +153,16 @@ function allTests($helper)
             test($ctx["ws.Sec-WebSocket-Version"] == "13");
             test(isset($ctx["ws.Sec-WebSocket-Key"]));
         }
+
+        // rcvSize and sndSize must be declared properties on Ice\UDPConnectionInfo, not deprecated dynamic
+        // properties created only at runtime (see #5536).
+        $udpInfo = $testIntf->ice_datagram()->ice_getConnection()->getInfo();
+        test($udpInfo instanceof Ice\UDPConnectionInfo);
+        $udpInfoClass = new ReflectionClass(Ice\UDPConnectionInfo::class);
+        test($udpInfoClass->hasProperty("rcvSize"));
+        test($udpInfoClass->hasProperty("sndSize"));
+        test($udpInfo->rcvSize >= 0);
+        test($udpInfo->sndSize >= 0);
     }
     echo "ok\n";
 


php/src/Connection.cpp
```diff
@@ -224,15 +224,15 @@ ZEND_METHOD(Ice_Connection, setBufferSize)
...
+    zend_long s;
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), const_cast<char*>("ll"), &r, &s) != SUCCESS)
     {
         RETURN_NULL();
     }
 
-    int rcvSize = static_cast<int>(Z_LVAL_P(r));
-    int sndSize = static_cast<int>(Z_LVAL_P(s));
+    int rcvSize = static_cast<int>(r);
+    int sndSize = static_cast<int>(s);
```
pepone, line 243: We need to validate the inputs against the accepted range by the C++ API.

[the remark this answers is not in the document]
pepone: Done. `setBufferSize` now validates both arguments are within the C++ `int` range before the cast and raises `InvalidArgumentException` (via the existing `invalidArgument()` helper, mirroring the int-range check in `Types.cpp`) instead of silently truncating. I used the full int32 range to match the C++ `setBufferSize(int, int)` signature (same as IcePy's `"ii"`). The info test now covers both a non-integer argument (`TypeError`) and an out-of-range one (`InvalidArgumentException`).

php/src/Connection.cpp
```diff
@@ -224,15 +224,15 @@ ZEND_METHOD(Ice_Connection, setBufferSize)
...
 
-    zval* r;
-    zval* s;
-    if (zend_parse_parameters(ZEND_NUM_ARGS(), const_cast<char*>("zz"), &r, &s) != SUCCESS)
+    zend_long r;
+    zend_long s;
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), const_cast<char*>("ll"), &r, &s) != SUCCESS)
     {
         RETURN_NULL();
     }
```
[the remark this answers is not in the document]
pepone: I kept this as-is: `if (zend_parse_parameters(...) != SUCCESS) { RETURN_NULL(); }` is the established idiom throughout IcePHP (flushBatchRequests, the Proxy/Communicator methods, etc.). On PHP 8 the TypeError/ArgumentCountError raised by zend_parse_parameters is already pending, so it propagates and the trailing RETURN_NULL is effectively a no-op. I'd rather keep this consistent with the rest of the extension than special-case one method; a sweep to RETURN_THROWS everywhere would be a reasonable separate cleanup if preferred.

php/src/Connection.cpp
```diff
@@ -224,15 +224,23 @@ ZEND_METHOD(Ice_Connection, setBufferSize)
...
 
-    int rcvSize = static_cast<int>(Z_LVAL_P(r));
-    int sndSize = static_cast<int>(Z_LVAL_P(s));
+    // Connection::setBufferSize takes int arguments in C++; reject values outside that range instead of silently
+    // truncating them.
+    if (r < INT_MIN || r > INT_MAX || s < INT_MIN || s > INT_MAX)
+    {
+        invalidArgument("buffer size must be in the range of a 32-bit signed integer");
```
bernardnormier: Given the checks, I would remove "32-bit" from the message. You're really checking the int range which happens to match int32 on most computers today.

bernardnormier (approved):

php/src/Connection.cpp
pepone: Done — reworded to "buffer size must be in the range of a C++ int" in 82a22237d7.

PR #5592, opened 2026-06-15
Always run peer verification on the OpenSSL and SecureTransport transports

pepone:
verifyPeer is now also run when the peer presents no certificate, so the OpenSSL and SecureTransport transports behave consistently with the Schannel transport. Adds an IceSSL configuration test covering this case.

### DIFF
--- a/cpp/test/IceSSL/configuration/AllTests.cpp
+++ b/cpp/test/IceSSL/configuration/AllTests.cpp
@@ -405,8 +405,18 @@ testCertificateVerification(
         server->ice_ping();
         test(false);
     }
-    catch (const LocalException&)
+    catch (const ProtocolException&)
     {
+        // Expected, if reported as an SSL alert by the server.
+    }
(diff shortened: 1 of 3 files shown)


cpp/test/IceSSL/configuration/AllTests.cpp
```diff
@@ -393,6 +393,23 @@ testCertificateVerification(
...
+    d["IceSSL.VerifyPeer"] = "1";
+    d["IceSSL.TrustOnly"] = "C=US, ST=Florida, O=ZeroC,"
+                            "OU=Ice test infrastructure, emailAddress=info@zeroc.com, CN=ca1.client";
+    server = fact->createServer(d);
+    try
+    {
+        server->ice_ping();
+        test(false);
+    }
+    catch (const LocalException&)
```
bernardnormier: Can we catch something more specific than LocalException?

bernardnormier (approved):

PR #5599, opened 2026-06-15
Detach listeners when a topic is destroyed while elements are attached

### LINKED ISSUE #5469: DataStorm: topic-destroy disconnect callback is dead code (suppressed by isDestroyed) — listener leak + debug assert

## Summary

`SessionI::disconnect(topicId, topic)` is called only from `TopicI::disconnect()`, which is called only from `TopicI::destroy()` — *after* `_destroyed` is set. The callback it schedules, `runWithTopic(..., [&]{ unsubscribe(topicId, topic); })`, is always skipped because `runWithTopic` bails out on `topic->isDestroyed()`. So `unsubscribe` → `element->detachKey/detachFilter` is dead code on this path, leaving `TopicI::_listenerCount` and the elements' `_listeners` stale when a topic is destroyed while readers/writers are still attached to peers.

## The bug

```cpp
// cpp/src/DataStorm/SessionI.cpp:931-944  (disconnect schedules unsubscribe)
runWithTopic(topicId, topic, [&](TopicSubscriber&) { unsubscribe(topicId, topic); });
// cpp/src/DataStorm/SessionI.cpp:1226-1237  (runWithTopic bails on isDestroyed)
// cpp/src/DataStorm/TopicI.cpp:127-147  (_destroyed set before disconnect)
```

## Impact

When a `Topic` is destroyed while its readers/writers are still attached to peers (the API permits this), `TopicI::_listenerCount` is never decremented → `assert(_listenerCount == 0)` fires in debug builds; in release, each surviving `DataElementI` keeps its `_listeners` entries (holding `shared_ptr<SessionI>`) and stale `hasWriters()/hasReaders()/getConnectedWriters()` results until the element itself is destroyed. The session side is removed, so the two sides diverge.

## Source references

- **(DataStorm / disconnect callback suppressed by isDestroyed)** — [`cpp/src/DataStorm/SessionI.cpp#L931-L944`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/DataStorm/SessionI.cpp#L931-L944)
- **(DataStorm / _destroyed set before disconnect)** — [`cpp/src/DataStorm/TopicI.cpp#L127-L147`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/DataStorm/TopicI.cpp#L127-L147)

## Recommendation

Perform the listener detachment on the destroy path without going through `runWithTopic`'s `isDestroyed` gate (or detach before setting `_destroyed`), so `_listenerCount`/`_listeners` are consistently torn down.

## Severity

**Medium** — listener bookkeeping leak and debug assert when a topic is destroyed while elements remain attached.

pepone:
SessionI::disconnect (called from TopicI::destroy) routed through runWithTopic,
whose isDestroyed gate is always true on this path, so the listener detachment
(unsubscribe -> detachKey/detachFilter) was skipped and TopicI::_listenerCount
was left stale, tripping a debug assert when a topic is destroyed while
readers/writers remain attached to a peer. Add an ignoreDestroyed flag to
runWithTopic, used only by the destroy path, so the listeners are detached.

Fixes #5469

### DIFF
--- a/changelog.d/datastorm/5469-datastorm-topic-destroy-detach.md
+++ b/changelog.d/datastorm/5469-datastorm-topic-destroy-detach.md
@@ -1,2 +1,3 @@
-- Fixed a bug in DataStorm where destroying a topic while a peer was still attached to one of its elements did not
-  detach the listeners, leaking them and (in debug builds) tripping an assertion.
+- Fixed a bug in DataStorm where destroying a topic before the readers and writers created from it, while a peer was
+  still attached to one of them, leaked the listeners and (in debug builds) tripped an assertion. This is uncommon,
+  since readers and writers are normally destroyed before their topic.
--- a/changelog.d/datastorm/5469-datastorm-topic-destroy-detach.md
+++ b/changelog.d/datastorm/5469-datastorm-topic-destroy-detach.md
@@ -1,0 +1,2 @@
+- Fixed a bug in DataStorm where destroying a topic while a peer was still attached to one of its elements did not
+  detach the listeners, leaking them and (in debug builds) tripping an assertion.
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -941,7 +941,10 @@ SessionI::disconnect(int64_t topicId, TopicI* topic)
         return; // Peer topic detached first.
     }
 
-    runWithTopic(topicId, topic, [&](TopicSubscriber&) { unsubscribe(topicId, topic); });
+    // disconnect() is called from TopicI::destroy() after the topic is marked destroyed, so pass ignoreDestroyed to
+    // detach the listeners anyway. Otherwise unsubscribe (and the element detachKey/detachFilter it drives) is
+    // skipped by the isDestroyed gate and TopicI::_listenerCount is left stale, tripping a debug assert.
+    runWithTopic(topicId, topic, [&](TopicSubscriber&) { unsubscribe(topicId, topic); }, true);
 
     auto& subscriber = _topics.at(topicId);
     subscriber.removeSubscriber(topic);
@@ -1223,7 +1226,11 @@ SessionI::runWithTopics(int64_t topicId, const function<void(TopicI*, TopicSubsc
 }
 
 void
-SessionI::runWithTopic(int64_t topicId, TopicI* topic, const function<void(TopicSubscriber&)>& callback)
+SessionI::runWithTopic(
+    int64_t topicId,
+    TopicI* topic,
+    const function<void(TopicSubscriber&)>& callback,
+    bool ignoreDestroyed)
 {
     auto t = _topics.find(topicId);
     if (t != _topics.end())
@@ -1232,7 +1239,7 @@ SessionI::runWithTopic(int64_t topicId, TopicI* topic, const function<void(Topic
         if (p != t->second.getSubscribers().end())
         {
             unique_lock<mutex> lock(topic->getMutex());
-            if (topic->isDestroyed())
+            if (!ignoreDestroyed && topic->isDestroyed())
             {
                 return;
             }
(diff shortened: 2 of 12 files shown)


pepone:
🤖 **Claude:** CI failures triaged + fixed in `8d04ce5938`.

**All the failures were the same thing** — the new `DataStorm/topicDestroy` test hanging:
- `debug on macos-26`, all Windows C++ jobs → the test's writer hung (harness killed it after the timeout).
- `debian13-arm64` / `amzn2023-x86_64` (6h00m) → the same hang running to the job's 6-hour cap.
- It only reproduces under load/timing — it passed on `macos-release`/`ubuntu` and 400/400 in an isolated loop here.

**Root cause — a test setup race, not a deadlock.** I pulled the writer coredump from the `debug-macos-26` artifact and reproduced locally under CPU stress to capture both stacks:
- **Reader** is stuck in `reader.waitForWriters()` (`Reader.cpp:27`).
- **Writer** is parked in `control.waitForNoReaders()`.
- Every other thread idle; **no mutex held**.

`waitForWriters()` is level-triggered (counts *current* writers). The writer attaches, its own `waitForReaders()` returns, and it destroys the topic so quickly that it detaches the reader **before the reader's `waitForWriters()` observes the writer** → the reader waits forever, never reaches `control.getNextUnread()`, never tears down, never detaches `control` → the writer hangs in `waitForNoReaders()`. (The core `_listenerCount`/assert fix in this PR is fine.)

**Fix (test-only):** added a reverse `sync` topic — the reader signals `ready` once its `waitForWriters()` returns, and the writer waits for it before destroying. The destroy can no longer outrun the reader's attach.

**Verified:** the unpatched test hangs within ~70 iterations under CPU stress; after the fix it ran **1000 stressed iterations clean**, and the standard harness passes.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -941,7 +941,10 @@ SessionI::disconnect(int64_t topicId, TopicI* topic)
         return; // Peer topic detached first.
     }
 
-    runWithTopic(topicId, topic, [&](TopicSubscriber&) { unsubscribe(topicId, topic); });
+    // disconnect() is called from TopicI::destroy() after the topic is marked destroyed, so pass ignoreDestroyed to
+    // detach the listeners anyway. Otherwise unsubscribe (and the element detachKey/detachFilter it drives) is
+    // skipped by the isDestroyed gate and TopicI::_listenerCount is left stale, tripping a debug assert.
+    runWithTopic(topicId, topic, [&](TopicSubscriber&) { unsubscribe(topicId, topic); }, /*ignoreDestroyed=*/true);
```
bernardnormier: I think it would be cleaner to "inline" runWithTopic-that-ignores-destroy right here, compared to adding this extra param with a confusing name.

bernardnormier (commented):
Can you also comment on "why no changelog fragment" for this bug fix?

changelog.d/datastorm/5469-datastorm-topic-destroy-detach.md
```diff
@@ -0,0 +1,2 @@
+- Fixed a bug in DataStorm where destroying a topic while a peer was still attached to one of its elements did not
+  detach the listeners, leaking them and (in debug builds) tripping an assertion.
```
bernardnormier: The current text reads as though this happens routinely. Since the trigger is specifically destroying a topic *before* the readers/writers created from it (uncommon under normal RAII order), maybe make that explicit:

```suggestion
- Fixed a bug in DataStorm where destroying a topic before the readers and writers created from it, while a peer was
  still attached to one of them, leaked the listeners and (in debug builds) tripped an assertion. This is uncommon,
  since readers and writers are normally destroyed before their topic.
```

bernardnormier (approved):
LGTM. Verified the fix is consistent with the `runWithTopic` locking protocol (topic `_mutex` is released in `TopicI::disconnect` before `session->disconnect`, so re-locking is safe), and that the trigger is genuinely uncommon: normal RAII teardown destroys readers/writers before their topic, and `DataElementI::disconnect` already zeroes the topic's listener count on that path. One changelog wording suggestion below to make the rarity explicit.

PR #5600, opened 2026-06-15
Fix data race on DataStorm user-callback members

### LINKED ISSUE #5470: DataStorm: data race on user-callback std::function members when callbacks are replaced at runtime

## Summary

The queued executor lambdas in `DataElementI` read the member `std::function`s (`_onSamples`, `_onConnectedKeys`, `_onConnectedElements`) on the `CallbackExecutor` thread **without any lock**, while `onSamples`/`onConnectedKeys`/`onConnectedElements` replace those members under `_parent->_mutex`. The public API documents that callbacks may be replaced at runtime, so concurrent replacement while samples flow is supported usage — and produces a torn read of a `std::function` (UB).

## The bug

```cpp
// cpp/src/DataStorm/DataElementI.cpp  (queued lambdas read members unlocked, e.g. :231-232,799-808,933-936)
_executor->queue([self = ..., sample] { self->_onSamples(sample); });   // reads _onSamples on executor thread
// setters under _parent->_mutex, e.g. :469-508
```

## Impact

An application that sets a reader/writer callback again while samples are being delivered triggers a data race on the `std::function` member (UB). The documented "callbacks will be replaced" usage is exactly this scenario.

## Source references

- **(DataStorm / callback members read unlocked on executor thread)** — [`cpp/src/DataStorm/DataElementI.cpp#L799-L808`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/DataStorm/DataElementI.cpp#L799-L808)
- **(DataStorm / callback setters under _parent->_mutex)** — [`cpp/src/DataStorm/DataElementI.cpp#L469-L508`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/DataStorm/DataElementI.cpp#L469-L508)

## Recommendation

Capture the callback by value at queue time (under the lock) instead of re-reading the member on the executor thread, so the executor never reads a `std::function` being concurrently replaced.

## Severity

**Medium** — data race / UB on user-callback members when callbacks are replaced at runtime while samples are flowing.

pepone:
The queued executor lambdas read the _onSamples/_onConnectedKeys/
_onConnectedElements std::function members on the callback-executor thread
without holding _parent->_mutex, while the onSamples/onConnectedKeys/
onConnectedElements setters replace those members under that mutex. Capture the
callback by value at queue time (under the lock) so the executor runs its own
copy and never reads a std::function being concurrently replaced. This mirrors
how the init callbacks are already captured.

Fixes #5470

### DIFF
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -228,8 +228,8 @@ DataElementI::attachKey(
 
     if (_onConnectedElements && added)
     {
-        _executor->queue([self = shared_from_this(), name]
-                         { self->_onConnectedElements(DataStorm::CallbackReason::Connect, name); });
+        _executor->queue([callback = _onConnectedElements, name]
+                         { callback(DataStorm::CallbackReason::Connect, name); });
     }
 
     if (addConnectedKey(key, subscriber))
@@ -286,9 +286,8 @@ DataElementI::detachKey(
         {
             if (_onConnectedElements)
             {
-                _executor->queue(
-                    [self = shared_from_this(), subscriber]
-                    { self->_onConnectedElements(DataStorm::CallbackReason::Disconnect, subscriber->name); });
+                _executor->queue([callback = _onConnectedElements, subscriber]
+                                 { callback(DataStorm::CallbackReason::Disconnect, subscriber->name); });
             }
             if (p->second.remove(topicId, elementId))
             {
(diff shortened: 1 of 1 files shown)


bernardnormier (approved):
LGTM. Verified the capture-by-value happens under `_parent->_mutex` at every queue site (the `initSamples`/`queue` paths run inside `runWithTopics`, which holds `topic->getMutex()` == the element's `_parent->_mutex`, the same mutex the setters use), so the executor never reads a `std::function` being concurrently replaced. Consistent with the existing init-callback capture, and no remaining cross-thread member reads.

PR #5602, opened 2026-06-15
Resolve DataStorm partial updates per key

### LINKED ISSUE #5473: DataStorm: multi-key partial updates computed against _last of a different key — silent data corruption

## Summary

`DataWriterI` and `DataReaderI` keep a single per-element `_last` sample, not a per-key one, when applying partial updates. `MultiKeyWriter::partialUpdate(key, value)` is public API; when a multi-key writer (or a reader subscribed to several keys) interleaves partial updates on different keys, the "previous" value passed to the user updater is the latest sample of whatever key was written last — producing wrong values for the key being updated.

## The bug

```cpp
// cpp/src/DataStorm/DataElementI.cpp:1023-1031  (writer publish)
if (sample->event == DataStorm::SampleEvent::PartialUpdate)
{
    _parent->getUpdater(sample->tag)(_last, sample, ...);   // _last is last sample of ANY key
}
// reader side: :780-789, :922-925 have the same single-slot _last
```

## Impact

A multi-key writer (or multi-key reader) interleaving partial updates across keys feeds the user updater the previous value of a *different* key, silently corrupting the updated key's value. Limited to multi-key + partial-update usage.

## Source references

- **(DataStorm / partial update against per-element _last)** — [`cpp/src/DataStorm/DataElementI.cpp#L1023-L1031`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/DataStorm/DataElementI.cpp#L1023-L1031)

## Recommendation

Track the previous sample per key (keyed `_last`) for partial-update resolution on both the writer and reader sides, or document and enforce a single-key restriction for partial updates.

## Severity

**Medium** — silent data corruption for multi-key partial updates (the "previous" value comes from a different key).

pepone:
DataWriterI and DataReaderI kept a single per-element _last sample, so a partial
update on one key was resolved against the most recent sample of a different
key, corrupting the fields the update does not set. Track the last sample per
key (_lastByKey) for partial-update resolution on both the writer (publish) and
reader (queue, initSamples) sides. Adds a multi-key case to the DataStorm/partial
test.

Fixes #5473


cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -753,7 +753,9 @@ DataReaderI::initSamples(
     }
 
     vector<shared_ptr<Sample>> valid;
-    shared_ptr<Sample> previous = _last;
+    // Resolve partial updates against the previous sample of the SAME key. A single `previous` would resolve a
+    // partial update for one key against the most recent sample of a different key.
```
bernardnormier: Nit: drop the second sentence — it describes what the old (buggy) single-slot code would do. The first line says what's needed.

```suggestion
    // Resolve partial updates against the previous sample of the SAME key.
```

cpp/src/DataStorm/DataElementI.h
```diff
@@ -377,7 +377,9 @@ namespace DataStormI
         TopicReaderI* _parent;
 
         std::deque<std::shared_ptr<Sample>> _samples;
-        std::shared_ptr<Sample> _last;
+        // The last sample received for each key, used to resolve partial updates per key. A single shared slot would
+        // resolve a partial update for one key against the most recent sample of a different key.
```
bernardnormier: Same nit here — trim the redundant "a single shared slot would..." sentence.

```suggestion
        // The last sample received for each key, used to resolve partial updates per key.
```

bernardnormier (approved):
LGTM — correct per-key fix. Verified the key identity holds up (`shared_ptr<Key>` is interned by `AbstractFactoryT`, the same basis `_keyElements`/`_connectedKeys` already rely on), all three sites (writer `publish`, reader `queue`/`initSamples`) resolve and update consistently, and the `initSamples` commit point matches where the old `_last` was set. Nice targeted multi-key test.

Left two trivial comment nits inline.

One non-blocking follow-up to consider (not for this PR): `_lastByKey` is the only one of these collections that's unbounded — `_connectedKeys` is erased on disconnect in `removeConnectedKey`, and `_samples` is trimmed by sampleCount/lifetime/clearHistory, but `_lastByKey` keeps one strong `Key`+`Sample` ref per distinct key ever seen. For the usual bounded key set it's a non-issue; only high-cardinality/ephemeral keys with partial updates would grow it. If we ever want to bound it: the **reader** side could piggyback on `removeConnectedKey` (a key with no connected publisher re-inits with a full sample on reconnect, so dropping its entry is safe), but the **writer** side can't reuse that hook — its `_lastByKey` is the last *published* value and is independent of connected readers. Worth a follow-up issue at most.

PR #5604, opened 2026-06-15
Track last ids for DataStorm filter subscriptions

### LINKED ISSUE #5471: DataStorm: filter subscriptions have no lastId tracking — duplicate samples on reconnect + debug assert

## Summary

Filter↔filter attachments (e.g. a filtered/any-key reader attached to an any-key writer, where the element id is negative) pass that negative id to `getLastIds`, but `subscribeToFilter` never populates `subscriber.keys[negativeId]`. The resulting `lastIds` dict is always empty, so after a session reconnect the peer writer's `getSamples` is called with `lastId = 0` and re-sends its entire retained queue. On the receiving side `subscriberInitialized` keeps the old `ElementSubscriber` (its `lastId` survives reconnects), tripping a debug assert and (in release) rewinding `lastId` and re-queueing already-processed samples — duplicate delivery.

## The bug

```cpp
// cpp/src/DataStorm/DataElementI.cpp:132-134  (attach passes id to getLastIds even when id < 0)
LongLongDict lastIds = key ? session->getLastIds(topicId, id, shared_from_this()) : LongLongDict{};
// cpp/src/DataStorm/SessionI.cpp:1106-1120,1152-1159  (keys[negativeId] never populated by subscribeToFilter)
assert(samples.front().id > elementSubscriber->lastId);   // fires in debug
elementSubscriber->lastId = samples.back().id;             // release: lastId rewound
```

## Impact

A filtered/any-key reader attached to an any-key writer that drops and reconnects while the writer retains history re-receives the entire retained queue: debug assert, or (release) duplicate delivery modulo discard policy.

## Source references

- **(DataStorm / negative filter id passed to getLastIds)** — [`cpp/src/DataStorm/DataElementI.cpp#L132-L134`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/DataStorm/DataElementI.cpp#L132-L134)
- **(DataStorm / subscriberInitialized assert/rewind)** — [`cpp/src/DataStorm/SessionI.cpp#L1152-L1159`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/DataStorm/SessionI.cpp#L1152-L1159)

## Recommendation

Track last ids for filter subscriptions (populate the relevant map in `subscribeToFilter`), or skip `getLastIds` for filter attachments and carry the per-element `lastId` across reconnects so retained history is not re-sent.

## Severity

**Medium** — duplicate sample delivery (and a debug assert) for filtered/any-key readers after a reconnect.

pepone:
`getLastIds` returned an empty dict for a subscription to a filter (any-key/filtered) writer, because only
subscriptions to key writers are recorded in `subscriber.keys`. As a result, after a session reconnect the writer
was asked for samples from `lastId` 0 and re-sent its entire retained queue, delivering duplicate samples to the
connected reader (and tripping a debug assert).

The trigger is the writer being any-key/filtered: any reader connected to such a writer is affected, regardless of
whether the reader itself is a key or filtered reader (the reader's subscription to a filter writer element is not
indexed in `subscriber.keys`).

The fix reports the `lastId` for each subscription to a filter writer element directly from the element
subscribers, keyed by the writer's element id (the id the writer matches in `DataElementI::attach`). It iterates
all such subscriptions rather than looking one up because filters have no per-id index (unlike keys), and a single
filter can match multiple writer elements with distinct element ids.

Adds a reconnect regression test to the DataStorm/reliability test.

Fixes #5471

### DIFF
--- a/changelog.d/datastorm/5471-datastorm-filter-lastid.md
+++ b/changelog.d/datastorm/5471-datastorm-filter-lastid.md
@@ -1,0 +1,3 @@
+- Fixed a bug in DataStorm where a filtered or any-key reader could receive duplicate samples after a session
+  reconnect: the writer re-sent its entire retained history instead of resuming after the samples the reader had
+  already received.
--- a/changelog.d/datastorm/5471-datastorm-filter-lastid.md
+++ b/changelog.d/datastorm/5471-datastorm-filter-lastid.md
@@ -1,3 +1,3 @@
-- Fixed a bug in DataStorm where a filtered or any-key reader could receive duplicate samples after a session
-  reconnect: the writer re-sent its entire retained history instead of resuming after the samples the reader had
-  already received.
+- Fixed a bug in DataStorm where a reader connected to an any-key (filtered) writer could receive duplicate
+  samples after a session reconnect: the writer re-sent its entire retained history instead of resuming after
+  the samples the reader had already received.
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -1120,13 +1120,13 @@ SessionI::getLastIds(int64_t topicId, int64_t keyOrFilterId, const std::shared_p
             // after a reconnect, delivering duplicate samples. We iterate all filter subscriptions rather than the
             // single keyOrFilterId entry because multiple any-key writers share a filter id but have distinct
             // element ids.
-            for (auto& [eid, elementSubscribers] : subscriber.getAll())
+            for (auto& [elementId, elementSubscribers] : subscriber.getAll())
             {
-                if (eid < 0)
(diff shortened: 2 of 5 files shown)


bernardnormier (approved):
LGTM. Verified the id round-trip: `attachFilter` stores the subscription under `-elementId` (the writer's element id), and the new branch emits it back as the positive id the writer matches via `data.lastIds.find(_id)`. The scan over negative-id entries is the right call since `keyOrFilterId` is the peer subscriber id, not the `_elements` storage key. The assert relaxation is also correct (and incidentally avoids the spurious `keys[0]` insertion in debug). Nice test.

PR #5606, opened 2026-06-15
Remove dead code in MATLAB and Swift

bernardnormier:
More dead code removal.

### DIFF
--- a/matlab/src/ice.h
+++ b/matlab/src/ice.h
@@ -33,7 +33,6 @@ extern "C"
     ICE_MATLAB_API mxArray* Ice_Communicator_unref(void*);
     ICE_MATLAB_API mxArray* Ice_Communicator_destroy(void*);
     ICE_MATLAB_API mxArray* Ice_Communicator_stringToProxy(void*, const char*, void**);
-    ICE_MATLAB_API mxArray* Ice_Communicator_proxyToString(void*, void*);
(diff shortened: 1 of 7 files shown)


swift/src/Ice/InputStream.swift
```diff
@@ -156,21 +156,6 @@ public final class InputStream {
         return encoding
     }
 
-    /// Skips over an encapsulation.
-    ///
-    /// - Returns: The encoding version of the skipped encapsulation.
-    func skipEncapsulation() throws -> EncodingVersion {
```
bernardnormier, line 162: Safe to remove since it's internal.

matlab/src/ice.h
```diff
@@ -33,7 +33,6 @@ extern "C"
     ICE_MATLAB_API mxArray* Ice_Communicator_unref(void*);
     ICE_MATLAB_API mxArray* Ice_Communicator_destroy(void*);
     ICE_MATLAB_API mxArray* Ice_Communicator_stringToProxy(void*, const char*, void**);
-    ICE_MATLAB_API mxArray* Ice_Communicator_proxyToString(void*, void*);
```
pepone, line 36: Isn't this part of the public API?

pepone (approved):

matlab/src/ice.h
bernardnormier: No, Communicator.proxyToString is implemented in MATLAB.

PR #5607, opened 2026-06-15
Remove dead code in Python, Ruby and PHP

bernardnormier:


php/src/Util.h
```diff
@@ -14,7 +14,6 @@ extern "C"
 
 namespace IcePHP
 {
-    void* createWrapper(zend_class_entry*, size_t);
```
InsertCreativityHere, line 17: Leftovers from when we switched from PHP5 to PHP7 apparently [.](https://github.com/zeroc-ice/ice/commit/a124372175471b4b536f9cacaee27d1e65af1a32)

InsertCreativityHere (approved):
Looks good to me!

pepone (approved):

PR #5608, opened 2026-06-16
Fix MATLAB ice_getConnectionAsync to pass the communicator

### LINKED ISSUE #5548: [Audit-High] MATLAB ice_getConnectionAsync builds Ice.Connection without the communicator — fetchOutputs errors on every successful call

## Summary

The `fetch` closure inside `ObjectPrx.ice_getConnectionAsync` constructs the result with `Ice.Connection(con)` — one argument. The `Connection` constructor is `Connection(impl, communicator)` with no single-argument path: its body unconditionally reads `communicator`. Every successful `fetchOutputs()` therefore raises a "Not enough input arguments" error. Both synchronous siblings pass `obj.communicator` correctly.

## The bug

```matlab
% matlab/lib/+Ice/ObjectPrx.m:1060-1066 (ice_getConnectionAsync)
function varargout = fetch(f)
    con = libpointer('voidPtr', 0); % Output param
    f.iceCall('fetch', con);
    assert(~isNull(con));
    varargout{1} = Ice.Connection(con);          % BUG — communicator missing
end

% matlab/lib/+Ice/Connection.m:20-25
function obj = Connection(impl, communicator)
    assert(isa(impl, 'lib.pointer'));
    obj@IceInternal.WrapperObject(impl);
    obj.communicator = communicator;             % unconditional — no nargin guard
end
```

`ice_getConnection` (`ObjectPrx.m:1042`) and `ice_getCachedConnection` (`:1086`) both call `Ice.Connection(v, obj.communicator)`; `obj` is already captured in the nested `fetch` closure.

## Impact

`f = proxy.ice_getConnectionAsync(); con = f.fetchOutputs();` deterministically errors on every successful fetch — the async variant of `ice_getConnection` is entirely unusable. (Even with a `nargin` guard it would produce a communicator-less `Connection`, breaking `Connection.createProxy`.) No MATLAB test exercises this path, consistent with the defect surviving.

## Source references

- **(single-arg construction)** — [`ObjectPrx.m#L1046-L1066`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIce/ObjectPrx.m#L1046-L1066)
- **(two-arg constructor, no fallback)** — [`Connection.m#L20-L25`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIce/Connection.m#L20-L25)
- **(correct sync siblings)** — [`ObjectPrx.m#L1026-L1043`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIce/ObjectPrx.m#L1026-L1043), [`#L1069-L1087`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIce/ObjectPrx.m#L1069-L1087)

## Recommendation

`varargout{1} = Ice.Connection(con, obj.communicator);`

## Severity

**High** — a public async API that deterministically fails on every successful call. Verified **CONFIRMED** by the second-model pass; ranked first in both priority lists.

pepone:
Fixes #5548.

### Problem
The `fetch` closure in `ObjectPrx.ice_getConnectionAsync` built its result with `Ice.Connection(con)` (one argument), but `Ice.Connection`'s constructor is `Connection(impl, communicator)` and reads `communicator` unconditionally. Every successful `fetchOutputs()` therefore failed with *"Not enough input arguments"*, making the asynchronous variant of `ice_getConnection` unusable. (Even with a `nargin` guard the result would be a communicator-less `Connection`, which breaks `Connection.createProxy`.)

### Fix
Pass `obj.communicator`, matching the synchronous siblings `ice_getConnection` and `ice_getCachedConnection`.

### Testing
Added a regression test to the `Ice/ami` suite that calls `ice_getConnectionAsync().fetchOutputs()` and then `con.createProxy(...)` (which needs the communicator). Verified locally with MATLAB R2025b against a Python server:
- before the fix: the `ami` suite fails at this test
- after the fix: the full `ami` suite passes

Fix also reviewed with codex locally.

### DIFF
--- a/matlab/test/Ice/ami/AllTests.m
+++ b/matlab/test/Ice/ami/AllTests.m
@@ -154,6 +154,13 @@ classdef AllTests
             fprintf('ok\n');
 
             if ~isempty(p.ice_getConnection())
+                fprintf('testing ice_getConnectionAsync... ');
+                con = p.ice_getConnectionAsync().fetchOutputs();
+                assert(isa(con, 'Ice.Connection'));
+                proxy = con.createProxy(p.ice_getIdentity());
+                assert(strcmp(proxy.ice_getIdentity().name, p.ice_getIdentity().name));
+                fprintf('ok\n');
+
                 fprintf('testing connection close... ');
 
                 %
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,9 @@ might need to be aware of.
 - Fixed the unmarshaling of unknown optional values with tags greater than or equal to 30. These no
   longer desynchronize the input stream causing spurious `MarshalException`.
 
+- Fixed the `ice_getConnectionAsync` proxy method. Retrieving the result of the returned future failed on every
+  successful call because the `Ice.Connection` was constructed without its communicator.
+
 ### Python Changes
 
 - Fixed Ice for Python to reliably abort request marshaling when an invalid value is supplied for a sequence
--- a/matlab/lib/+Ice/ObjectPrx.m
+++ b/matlab/lib/+Ice/ObjectPrx.m
@@ -1061,7 +1061,7 @@ classdef ObjectPrx < IceInternal.WrapperObject
                 con = libpointer('voidPtr', 0); % Output param
                 f.iceCall('fetch', con);
                 assert(~isNull(con));
-                varargout{1} = Ice.Connection(con);
+                varargout{1} = Ice.Connection(con, obj.communicator);
             end
             r = Ice.Future(future, 'ice_getConnection', 1, 'Ice_GetConnectionFuture', @fetch);
         end


matlab/test/Ice/ami/AllTests.m
```diff
@@ -154,6 +154,17 @@ function allTests(helper)
             fprintf('ok\n');
 
             if ~isempty(p.ice_getConnection())
+                fprintf('testing ice_getConnectionAsync... ');
+                % Regression test for issue #5548: the Connection built by the
```
bernardnormier: I would delete this comment.

bernardnormier (approved):

PR #5609, opened 2026-06-16
Fix MATLAB InputStream.skipOptionals tag-30 sentinel detection

### LINKED ISSUE #5549: [Audit-Medium] MATLAB InputStream.skipOptionals uses a left shift for the tag-30 sentinel — unread optionals with tag >= 30 desync the stream

## Summary

`InputStream.skipOptionals` detects the tag-30 sentinel with `bitshift(v, 3)` — a **left** shift. The optional framing byte encodes `(tag << 3) | format`, so recovering the tag needs a **right** shift (`bitshift(v, -3)`), as the sibling `readOptionalImpl` in the same file correctly does. A left-shifted `uint8` has its low three bits clear (and saturates), so it can never equal 30 — the branch is dead, and the size byte(s) that follow a tag-30 sentinel are never skipped.

## The bug

```matlab
% matlab/lib/+Ice/InputStream.m:613-619 (skipOptionals) — WRONG
format = bitand(v, uint8(7)); % Read first 3 bits.
if bitshift(v, 3) == uint8(30)     % LEFT shift — can never equal 30
    obj.skipSize();
end
obj.skipOptional(format);

% :566-575 (readOptionalImpl) — correct sibling
tag = uint32(bitshift(v, -3));     % RIGHT shift
...
if tag == 30
    tag = obj.readSize();
end
```

The JS port uses `v >> 3` here.

## Impact

`skipOptionals` runs at the end of an encapsulation and at the end of a 1.1 slice, to skip optionals the receiver did not read. When an **unread** optional has a tag ≥ 30 (encoded as the `30` sentinel followed by a size), the size bytes are left in the stream — the decode desynchronizes, producing a spurious `MarshalException` ("buffer size does not match...") or a misparse of the following data. This breaks interop on perfectly valid messages from peers whose Slice definitions use high optional tags (or more optionals than the receiver knows about).

## Source references

- **(left shift)** — [`InputStream.m#L608-L619`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIce/InputStream.m#L608-L619)
- **(correct sibling)** — [`InputStream.m#L551-L575`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIce/InputStream.m#L551-L575)
- **(tag-30 encoding on the write side)** — [`OutputStream.m#L432-L446`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIce/OutputStream.m#L432-L446)

## Recommendation

`if bitshift(v, -3) == uint8(30)`.

## Severity

**Medium** — a deterministic stream desync, gated on skipping an unread optional with tag ≥ 30. Verified **CONFIRMED** by the second-model pass.

pepone:
Fixes #5549.

### Problem
`InputStream.skipOptionals` detected the tag-30 sentinel with a **left** shift (`bitshift(v, 3)`). For a `uint8`, a left shift can never equal 30, so the branch was dead and the size byte(s) following an unread optional with tag ≥ 30 were never skipped. `skipOptionals` runs at the end of an encapsulation (and at the end of a 1.1 slice) to skip optionals the receiver did not read, so any valid message carrying an unread optional with a tag ≥ 30 would desynchronize the stream and raise a spurious `MarshalException`.

### Fix
Use a **right** shift (`bitshift(v, -3)`) to recover the tag, matching the sibling `readOptionalImpl` in the same file (and `OutputStream.writeOptionalImpl` on the write side).

### Testing
Added a self-contained regression test (`skipUnreadOptionals`) to the `Ice/optional` suite. It marshals an encapsulation with a low-tag optional plus several optionals with tag ≥ 30 (tag 30, tag 50, and tag 300 which needs a multi-byte size), reads back only the low-tag optional, and relies on `endEncapsulation()` → `skipOptionals()` to skip the rest. Verified locally with MATLAB R2025b:
- before the fix: the test fails with a buffer-size `MarshalException`
- after the fix: the full `Ice/optional` suite passes (compact and sliced formats)

Fix also reviewed with codex locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,9 @@ might need to be aware of.
 - Fixed `ice_getCachedConnection` to return an empty `Ice.Connection` array, rather than an empty
   `double` array, when the proxy has no cached connection.
 
+- Fixed the unmarshaling of optional values. An unread optional value with a tag greater than or equal to 30 no
+  longer desynchronizes the input stream and produces a spurious `MarshalException`.
+
 ### Python Changes
 
 - Fixed Ice for Python to reliably abort request marshaling when an invalid value is supplied for a sequence
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,8 +54,8 @@ might need to be aware of.
 - Fixed `ice_getCachedConnection` to return an empty `Ice.Connection` array, rather than an empty
   `double` array, when the proxy has no cached connection.
 
-- Fixed the unmarshaling of optional values. An unread optional value with a tag greater than or equal to 30 no
-  longer desynchronizes the input stream and produces a spurious `MarshalException`.
+- Fixed the unmarshaling of unknown optional values with tags greater than or equal to 30. These no
+  longer desynchronize the input stream causing spurious `MarshalException`.
 
 ### Python Changes
 
--- a/matlab/test/Ice/optional/AllTests.m
+++ b/matlab/test/Ice/optional/AllTests.m
@@ -773,13 +773,9 @@ classdef AllTests
         end
 
         function skipUnreadOptionals(communicator)
-            % Regression test for issue #5549: InputStream.skipOptionals must recognize the
-            % tag-30 sentinel with a right shift. We marshal an encapsulation that holds one
-            % low-tag optional plus several optionals with tag >= 30 (each encoded as the
-            % tag-30 sentinel followed by a size), read back only the low-tag optional, and
-            % rely on endEncapsulation() -> skipOptionals() to skip the rest. With the former
-            % left-shift bug the sentinel was never detected, the size byte(s) were left in the
-            % stream, and endEncapsulation() failed with a buffer-size MarshalException.
+            % Marshal an encapsulation with one optional whose tag fits on a single byte,
+            % followed by several optionals with tags >= 30. Read back only the first optional, and let 
+            % `endEncapsulation() -> skipOptionals()` skip the rest to exercise detection of the tag-30 sentinel.
             encoding = Ice.EncodingVersion(1, 1);
 
             os = Ice.OutputStream(encoding);
--- a/matlab/test/Ice/optional/AllTests.m
+++ b/matlab/test/Ice/optional/AllTests.m
@@ -7,6 +7,10 @@ classdef AllTests
 
             communicator = helper.communicator();
 
+            fprintf('testing skipping of unread optionals with tag >= 30... ');
+            AllTests.skipUnreadOptionals(communicator);
+            fprintf('ok\n');
+
             ref = ['initial:', helper.getTestEndpoint()];
             initial = InitialPrx(communicator, ref);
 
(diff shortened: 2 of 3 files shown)


matlab/test/Ice/optional/AllTests.m
```diff
@@ -767,5 +771,36 @@
...
         end
+
+        function skipUnreadOptionals(communicator)
+            % Regression test for issue #5549: InputStream.skipOptionals must recognize the
+            % tag-30 sentinel with a right shift. We marshal an encapsulation that holds one
+            % low-tag optional plus several optionals with tag >= 30 (each encoded as the
+            % tag-30 sentinel followed by a size), read back only the low-tag optional, and
+            % rely on endEncapsulation() -> skipOptionals() to skip the rest. With the former
+            % left-shift bug the sentinel was never detected, the size byte(s) were left in the
+            % stream, and endEncapsulation() failed with a buffer-size MarshalException.
```
bernardnormier: This comment is quite verbose. Suggesting a tighter version that keeps the intent (what the test does and that it exercises sentinel detection):

```suggestion
            % Marshal an encapsulation with one low-tag optional plus several with tag >= 30,
            % read back only the low-tag one, and let endEncapsulation() -> skipOptionals()
            % skip the rest. This exercises detection of the tag-30 sentinel.
```

bernardnormier (approved):

CHANGELOG.md
```diff
@@ -7,25 +7,20 @@ might need to be aware of.
...
-  instead of the marshal order. As a result, an operation with an optional parameter declared before a required
-  parameter (or optional parameters declared out of tag order) could fail to decode or decode incorrect values.
+### MATLAB Changes
 
-- Fixed a bug in `slice2cs --icerpc` where the generated proxy built the response tuple in marshal order while
-  declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
-  different from their marshal order could return values in the wrong tuple slots or fail to compile.
-
-These are the changes since the [Ice 3.8.2] release.
+- Fixed the unmarshaling of optional values. An unread optional value with a tag greater than or equal to 30 no
```
bernardnormier: It should be "unknown" or "unknown and skipped", not unread.

matlab/test/Ice/optional/AllTests.m
```diff
@@ -7,6 +7,10 @@
 
             communicator = helper.communicator();
 
+            fprintf('testing skipping of unread optionals with tag >= 30... ');
```
InsertCreativityHere: The terminology we use everywhere else is `unknown` not `unread`.
```suggestion
            fprintf('testing skipping of unknown optionals with tag >= 30... ');
```
Although it is technically true that unknown tags will go unread... but this isn't important.

CHANGELOG.md
```diff
@@ -7,25 +7,20 @@ might need to be aware of.
...
+  longer desynchronizes the input stream and produces a spurious `MarshalException`.
```
InsertCreativityHere: Agreed about `unknown`, but I also think the first sentence sounds repetitive. How about?:
```suggestion
- Fixed the unmarshaling of unknown optional values with tags greater than or equal to 30. These no
  longer desynchronize the input stream causing spurious `MarshalException`.
```

matlab/test/Ice/optional/AllTests.m
InsertCreativityHere: I think the comment could be even a little clearer. We don't have any idea of `low-tag optional`s, and what's really important is tags fitting on 1 byte or not:
```suggestion
            % Marshal an encapsulation with one optional whose tag fits on a single byte,
            % followed by several optionals with tags >= 30. Read back only the first optional, and let 
            % `endEncapsulation() -> skipOptionals()` skip the rest to exercise detection of the tag-30 sentinel.
```

InsertCreativityHere (approved):
Looks good to me, some comments could use slight tweaking though! : v)

PR #5611, opened 2026-06-16
Fix MATLAB EncapsDecoder patch-map accounting on sparse slots

### LINKED ISSUE #5550: [Audit-Medium] MATLAB EncapsDecoder miscounts patchMapLength across sparse cell-array slots — premature ice_postUnmarshal and suppressed missing-instance checks

## Summary

`EncapsDecoder` tracks outstanding value-graph patches in `patchMap` (a cell array) with a separate counter `patchMapLength`. MATLAB cell arrays auto-extend with `[]` filler slots when a high index is assigned, and the two accounting sites treat those fillers inconsistently:

- `addPatchEntry` increments the counter **only** when `index > length(patchMap)` — appending a real entry to a pre-existing *filler* slot adds an outstanding patch without incrementing (undercount);
- `unmarshal` decrements for **any** `index <= length(patchMap)`, including filler slots that never held an entry (spurious decrement).

`patchMapLength == 0` gates when `ice_postUnmarshal` runs and the "index for class received but no instance" integrity checks in both decoders — so the gate can hit zero while patches are still outstanding.

## The bug

```matlab
% matlab/lib/+IceInternal/EncapsDecoder.m (addPatchEntry)
if index <= length(obj.patchMap)
    pl = obj.patchMap{index};                       % may be [] (auto-extend filler)
else
    pl = {};
    obj.patchMapLength = obj.patchMapLength + 1;     % counted ONLY here
end
...
obj.patchMap{index} = pl;                            % high index extends with [] fillers

% (unmarshal)
if index <= length(obj.patchMap)
    l = obj.patchMap{index};                         % may be []
    for i = 1:length(l) ... end
    obj.patchMap{index} = [];
    obj.patchMapLength = obj.patchMapLength - 1;      % decremented for ANY in-range index
end
```

The JS reference uses a real `Map` and only counts/deletes entries that actually exist.

## Impact

On forward/circular value graphs (a slice referencing an instance with a higher wire index than instances read before it — common in object graphs), the counter can reach zero early: `ice_postUnmarshal` runs before all members are patched, and the missing-instance `MarshalException` can be suppressed — silently delivering a value graph with an unpatched (unset) member. Contiguous graphs are unaffected.

Worked example: `addPatchEntry(5)` extends `patchMap` to length 5 (`[]` at 1–4, counter = 1); a later `unmarshal(2)` enters the in-range branch on the empty slot and decrements the counter to 0 while the patch for index 5 is still outstanding.

## Source references

- **(asymmetric accounting)** — [`EncapsDecoder.m#L71-L170`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIceInternal/EncapsDecoder.m#L71-L170)
- **(gated integrity checks)** — [`EncapsDecoder10.m#L10-L25`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIceInternal/EncapsDecoder10.m#L10-L25), [`EncapsDecoder11.m#L364-L374`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIceInternal/EncapsDecoder11.m#L364-L374)
- **(JS reference: entry-based `Map` accounting)** — [`js/packages/ice/src/Ice/InputStream.js#L84-L179`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/InputStream.js#L84-L179)

## Recommendation

Count/decrement only on actual presence, mirroring the JS port: increment whenever a genuinely new patch list is started (including into a filler slot), and decrement/enter the patch branch only when `~isempty(obj.patchMap{index})`.

## Severity

**Medium** — premature post-unmarshal callbacks and suppressed missing-instance detection on non-contiguous value graphs. Verified **CONFIRMED** by the second-model pass.

pepone:
Fixes #5550.

### Problem
`EncapsDecoder` tracks outstanding value-graph patches in `patchMap` (a cell array) with the counter `patchMapLength`. MATLAB cell arrays auto-extend with empty `[]` filler slots when a high index is assigned, and the two accounting sites treated those fillers inconsistently:
- `addPatchEntry` incremented the counter only when the index was beyond the array, so appending into a pre-existing filler slot added a patch without incrementing (undercount);
- `unmarshal` decremented for any in-range index, including filler slots that never held a patch (spurious decrement).

On value graphs with non-contiguous instance indices, `patchMapLength` could reach zero while patches were still outstanding — running `ice_postUnmarshal` before all members were patched and suppressing the missing-instance check.

### Fix
Count and decrement only on actual presence (`~isempty(patchMap{index})`), mirroring the entry-based `Map` accounting in the JavaScript mapping.

### Testing
The full `Ice/objects` suite (which exercises circular/forward value graphs in both compact and sliced formats) passes. A dedicated reproduction was not added: deterministically forcing the non-contiguous filler scenario requires precise control of internal wire-index assignment, which is disproportionate to a two-line fix. Reviewed with codex.

### DIFF
--- a/matlab/lib/+IceInternal/EncapsDecoder.m
+++ b/matlab/lib/+IceInternal/EncapsDecoder.m
@@ -68,6 +68,13 @@ classdef (Hidden, Abstract) EncapsDecoder < handle
             r = obj.is.getCommunicator().getSliceLoader().newInstance(typeId);
         end
 
+        function r = hasPatchList(obj, index)
+            % True if patchMap holds an actual patch list at index. Cell arrays auto-extend with
+            % empty filler slots when a higher index is assigned, so an in-range index is not
+            % enough -- the slot must be non-empty.
+            r = index <= length(obj.patchMap) && ~isempty(obj.patchMap{index});
+        end
+
         function addPatchEntry(obj, index, cb)
             %assert(index > 0);
             %
@@ -87,12 +94,12 @@ classdef (Hidden, Abstract) EncapsDecoder < handle
             % the callback will be called when the instance is
             % unmarshaled.
             %
-            if index <= length(obj.patchMap)
+            if obj.hasPatchList(index)
                 pl = obj.patchMap{index};
             else
                 %
-                % We have no outstanding instances to be patched for this
-                % index, so make a new entry in the patch map.
+                % We have no outstanding instances to be patched for this index, so make a new
+                % entry in the patch map.
                 %
                 pl = {};
                 obj.patchMapLength = obj.patchMapLength + 1;
@@ -120,7 +127,7 @@ classdef (Hidden, Abstract) EncapsDecoder < handle
             %
             v.iceRead(obj.is);
 
-            if index <= length(obj.patchMap)
+            if obj.hasPatchList(index)
                 %
                 % Patch all instances now that the instance is unmarshaled.
                 %


CHANGELOG.md
```diff
@@ -54,6 +54,10 @@ might need to be aware of.
 - Fixed `ice_getCachedConnection` to return an empty `Ice.Connection` array, rather than an empty
   `double` array, when the proxy has no cached connection.
 
+- Fixed the unmarshaling of value graphs whose instances have non-contiguous indices. The decoder could
+  miscount the outstanding patches, invoking `ice_postUnmarshal` before all members were patched and suppressing
+  the missing-instance check.
+
```
bernardnormier: Let's drop this CHANGELOG entry. The bug isn't reachable by a conformant sender: instance indices are minted densely and in increasing order at first reference (`++_valueIdIndex` in `OutputStream`), so the decoder always populates `patchMap` contiguously and never creates the `[]` filler gaps this fixes. Only a malformed/malicious message (which reads an arbitrary index straight off the wire) can produce the gap, so this is a decoding-hardening fix with no observable effect on valid traffic — too minor for a CHANGELOG entry.

```suggestion
```

bernardnormier (approved):
Reviewed against the C# reference (`InputStream.cs`) and the encoder. The fix is correct: `hasPatchList` restores true key-presence semantics, matching C#'s `Dictionary` / JS's `Map` (count and remove only on real entries, not on cell-array `[]` fillers).

One important point on reachability — this is **not reachable by a conformant sender**, so it's a decoding-hardening fix rather than a value-graph correctness bug:

- Instance indices are minted densely and in increasing order at first reference (`++_valueIdIndex` — `OutputStream` 1.0 `registerValue`, 1.1 `writeInstance`). The decoder reads references in stream order, so `patchMap` is always populated contiguously and the `[]` filler gaps this fixes never form for valid traffic.
- In 1.1 it's stronger still: instances are registered in `unmarshaledMap` before their members are read, so references resolve immediately and patches don't survive at all.
- The only way to create a gap is a malformed/malicious message, where the index is read straight off the wire (`readInt` in 1.0, `readSize` in 1.1). There the old code could miss the "index for class received, but no instance" `MarshalException` (silently accepting a truncated graph) or run `ice_postUnmarshal` early.

Correct and worth merging as hardening. I suggested dropping the CHANGELOG entry separately (too minor — no effect on conformant traffic). Approving.

PR #5612, opened 2026-06-16
Fix slice2matlab long constants/defaults above 2^53 losing precision

### LINKED ISSUE #5484: slice2matlab: long constants/defaults above 2^53 silently lose precision

## Summary

`slice2matlab` emits `long` constants and defaults as bare numeric literals. MATLAB parses a bare numeric literal as `double` before converting to `int64`, so values with magnitude > 2^53 are rounded — the generated `int64` property holds the wrong value. slice2php handles this case by emitting out-of-range longs as strings; slice2matlab has no equivalent.

## The bug

```cpp
// cpp/src/slice2matlab/Gen.cpp:131-142  (constantValue, KindLong -> return value;), used at :167,2720
```

Contrast `slice2php/Main.cpp:991-1003`, which emits out-of-range longs as strings.

## Impact

`const long L = 9007199254740993;` yields `…992` in the generated MATLAB (precision loss for any `long` constant/default above 2^53).

## Source references

- **(slice2matlab / bare long literal loses precision > 2^53)** — [`cpp/src/slice2matlab/Gen.cpp#L131-L142`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2matlab/Gen.cpp#L131-L142)

## Recommendation

Emit `long` constants/defaults whose magnitude exceeds 2^53 in a form MATLAB parses exactly into `int64` (e.g. via `int64(...)` of a string literal), as slice2php does.

## Severity

**Medium** — `long` constants/defaults above 2^53 are silently rounded in generated MATLAB.

pepone:
Fixes #5484.

### Problem
`slice2matlab` emitted `long` constants and defaults as bare numeric literals. MATLAB parses a bare literal as a `double` before converting to `int64`, so values with magnitude above 2^53 were silently rounded (e.g. `const long L = 9007199254740993` generated `…992`).

### Fix
In `constantValue`, emit `long` values with magnitude above 2^53 as `int64(sscanf('<value>', '%ld'))`, which parses exactly to `int64`. Values that are exactly representable as a `double` are still emitted as plain literals.

### Testing
Added a `const long` above 2^53 to the `Ice/defaultValue` suite and assert it is exact. Verified with MATLAB R2025b: the generated code parses to the exact `int64` and the suite passes; regenerating with the previous compiler produces the rounded literal and the suite fails. Reviewed with codex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,10 @@ might need to be aware of.
 - Fixed `Ice.Future.wait('sent')`, which could block indefinitely or time out even after the request had been
   sent. The wait now completes as soon as the invocation reaches or passes the requested state.
 
+- Fixed `slice2matlab` to map all `long` constants and default values to MATLAB `int64`. They were previously
+  emitted as bare numeric literals, which MATLAB interprets as `double`, silently losing precision for values
+  with magnitude greater than 2^53.
+
 ### PHP Changes
 
 - Fixed Ice for PHP mis-marshaling a non-empty optional `sequence<string>`.
--- a/cpp/src/slice2matlab/Gen.cpp
+++ b/cpp/src/slice2matlab/Gen.cpp
@@ -128,11 +128,17 @@ namespace
                     {
                         return "sprintf('" + toStringLiteral(value, "\a\b\f\n\r\t\v", "", Matlab, 255) + "')";
                     }
+                    case Builtin::KindLong:
+                    {
+                        // A Slice long maps to MATLAB int64. A bare numeric literal is parsed by MATLAB as a
+                        // double, which loses precision above 2^53, so wrap the value in int64(...), which
+                        // MATLAB parses directly as an integer.
+                        return "int64(" + value + ")";
+                    }
                     case Builtin::KindBool:
                     case Builtin::KindByte:
                     case Builtin::KindShort:
                     case Builtin::KindInt:
-                    case Builtin::KindLong:
                     case Builtin::KindFloat:
                     case Builtin::KindDouble:
                     case Builtin::KindObjectProxy:


cpp/src/slice2matlab/Gen.cpp
```diff
@@ -128,11 +129,22 @@ namespace
...
+                    case Builtin::KindLong:
+                    {
+                        // MATLAB parses a bare numeric literal as a double before converting
+                        // it to int64, which silently loses precision for magnitudes above
+                        // 2^53. Emit such values via sscanf so they parse exactly into int64.
+                        int64_t l = std::stoll(value, nullptr, 0);
+                        if (l > 9007199254740992LL || l < -9007199254740992LL)
+                        {
+                            return "int64(sscanf('" + std::to_string(l) + "', '%ld'))";
+                        }
```
pepone: Can't we just use `"int64(" + value + ")";`?

See "Loss of Precision Due to Conversion" in https://www.mathworks.com/help/matlab/matlab_prog/integers.html

CHANGELOG.md
```diff
@@ -54,6 +54,9 @@ might need to be aware of.
 - Fixed `ice_getCachedConnection` to return an empty `Ice.Connection` array, rather than an empty
   `double` array, when the proxy has no cached connection.
 
+- Fixed `slice2matlab` to emit `long` constants and default values whose magnitude exceeds 2^53 in a form that
+  MATLAB parses to the exact `int64` value, instead of a bare numeric literal that silently loses precision.
```
bernardnormier: The CHANGELOG should describe the actual fix: all `long` constants now map to `int64`, not only those above 2^53.

```suggestion
- Fixed `slice2matlab` to map all `long` constants and default values to MATLAB `int64`. They were previously
  emitted as bare numeric literals, which MATLAB interprets as `double`, silently losing precision for values
  with magnitude greater than 2^53.
```

matlab/test/Ice/defaultValue/Test.ice
```diff
@@ -44,6 +44,7 @@ module Test
     const short ConstShort = 16000;
     const int ConstInt = 3;
     const long ConstLong = 4;
+    const long ConstLongAbove2Pow53 = 9007199254740993; // 2^53 + 1; see issue #5484
```
bernardnormier: We don't need a dedicated constant to verify the `int64` mapping; the existing `ConstLong` works. Drop this line.

```suggestion
```

matlab/test/Ice/defaultValue/AllTests.m
```diff
@@ -33,6 +33,12 @@ function allTests(helper)
             assert(v.zeroD == 0);
             assert(v.zeroDotD == 0);
 
+            % Regression for issue #5484: a long constant above 2^53 must be represented
+            % exactly, not silently rounded by MATLAB's double literal parsing.
+            assert(isa(ConstLongAbove2Pow53.value, 'int64'));
+            assert(ConstLongAbove2Pow53.value == int64(9007199254740993));
+            assert(ConstLongAbove2Pow53.value - int64(9007199254740992) == 1);
+
```
bernardnormier: This whole block can go: we don't need a new constant, the issue-number comment, or the asserts that just exercise MATLAB's own `int64` arithmetic.

```suggestion
```

Instead, add a single bare assertion right after the existing `assert(v.l == ConstLong.value);` further down, so the test stands on its own and reuses the existing constant:

```matlab
assert(isa(ConstLong.value, 'int64'));
```

cpp/src/slice2matlab/Gen.cpp
```diff
@@ -128,11 +128,17 @@ namespace
                     {
                         return "sprintf('" + toStringLiteral(value, "\a\b\f\n\r\t\v", "", Matlab, 255) + "')";
                     }
+                    case Builtin::KindLong:
+                    {
+                        // A Slice long maps to MATLAB int64. A bare numeric literal is parsed by MATLAB as a
+                        // double, which loses precision above 2^53, so wrap the value in int64(...), which
+                        // MATLAB parses directly as an integer.
+                        return "int64(" + value + ")";
```
InsertCreativityHere, line 136: This works just fine in modern MATLAB. But I was surprised to see this works!

When I used MATLAB in the past, I remember being caught up by not storing these kinds of large constants as strings. Thankfully we only support 2025 edition, so I guess no concern for older editions here.

matlab/test/Ice/defaultValue/AllTests.m
```diff
@@ -33,6 +33,12 @@ function allTests(helper)
             assert(v.zeroD == 0);
             assert(v.zeroDotD == 0);
 
+            % Regression for issue #5484: a long constant above 2^53 must be represented
+            % exactly, not silently rounded by MATLAB's double literal parsing.
+            assert(isa(ConstLongAbove2Pow53.value, 'int64'));
+            assert(ConstLongAbove2Pow53.value == int64(9007199254740993));
+            assert(ConstLongAbove2Pow53.value - int64(9007199254740992) == 1);
```
InsertCreativityHere: MATLAB has a constant for this that we should consider using for clarity:
```suggestion
            boundary = int64(flintmax)
            assert(isa(ConstLongAbove2Pow53.value, 'int64'));
            assert(ConstLongAbove2Pow53.value == (boundary + 1));
            assert(ConstLongAbove2Pow53.value - boundary == 1);
```
`flintmax` is the maximum value for which a floating point and an integer are equivalent.
Which is exactly what we care about here.

InsertCreativityHere (approved):
Looks good to me!

matlab/test/Ice/defaultValue/AllTests.m
InsertCreativityHere: I think verifying the value is actually meaningful and more important than just verifying the type.
Because the type is actually irrelevant to this bug, all that matters is the value.

Trying to store numbers bigger than `flintmax` was a known gotcha back in the day, and for older MATLAB versions, writing `int64(BIGNUMBER)`  would still fail, because `BIGNUMBER` is parsed as a double before conversion, even thought the type was `int64`. So just checking the type doesn't really say anything about the value being correct.

matlab/test/Ice/defaultValue/AllTests.m
```diff
@@ -33,6 +33,13 @@ function allTests(helper)
             assert(v.zeroD == 0);
             assert(v.zeroDotD == 0);
 
+            % A long constant above flintmax (the largest value for which a double and an integer are
+            % equivalent) must be represented exactly, not silently rounded by MATLAB's double literal parsing.
+            boundary = int64(flintmax);
+            assert(isa(ConstLongAbove2Pow53.value, 'int64'));
+            assert(ConstLongAbove2Pow53.value == (boundary + 1));
```
bernardnormier: This is just testing MATLAB, I don't see the point.

bernardnormier (approved):

matlab/test/Ice/defaultValue/AllTests.m
pepone: Removed

PR #5613, opened 2026-06-16
Fix out-of-bounds access in MATLAB MEX endpoint accessors

### LINKED ISSUE #5552: [Audit-Medium, grouped] MATLAB MEX endpoint accessors off-by-one (idx > size instead of >=) — OOB read in ice_getEndpoint, OOB write in ice_setEndpoint
> **AI-generated audit finding** — these are findings from an automated **correctness** audit of the MATLAB source (pure-MATLAB runtime + C++ MEX extension) performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). They have **not been triaged by a human yet**; verify each before acting on it. Grouped: the identical off-by-one in the two adjacent endpoint accessors — one read side, one write side. Links pinned to commit `e248644d788c7e76e14af8676529201c54c1735d`.

The MEX endpoint accessors take **0-based** indices (the MATLAB wrappers pass `i - 1`, [`ObjectPrx.m#L492-L509`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIce/ObjectPrx.m#L492-L509), [`#L528-L540`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/lib/%2BIce/ObjectPrx.m#L528-L540)), so the valid range is `[0, size-1]` — but both bounds checks use `idx > size` instead of `idx >= size`, letting `idx == size` through to unchecked `std::vector::operator[]`.

### 1. `Ice_ObjectPrx_ice_getEndpoint` — out-of-bounds read

```cpp
// matlab/src/ObjectPrx.cpp:415-425
auto endpoints = restoreProxy(self)->ice_getEndpoints();
if (idx > endpoints.size())            // BUG — should be >=
{
    throw std::invalid_argument("index outside range");
}
*r = createShared<Ice::Endpoint>(endpoints[idx]);   // OOB read of a shared_ptr for idx == size
```

[`ObjectPrx.cpp#L415-L425`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/src/ObjectPrx.cpp#L415-L425)

### 2. `Ice_ObjectPrx_ice_setEndpoint` — out-of-bounds write

```cpp
// matlab/src/ObjectPrx.cpp:439-455
vector<shared_ptr<Ice::Endpoint>>* v = reinterpret_cast<...>(arr);   // pre-sized to num entries
if (idx > v->size())                   // BUG — should be >=
{
    throw std::invalid_argument("index outside range");
}
(*v)[idx] = deref<Ice::Endpoint>(e);   // OOB shared_ptr write for idx == size — heap corruption
```

[`ObjectPrx.cpp#L433-L455`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/src/ObjectPrx.cpp#L433-L455)

## Impact

Undefined behavior (OOB `shared_ptr` read; OOB `shared_ptr` assignment — heap corruption) for `idx == size`. The stock MATLAB wrappers stay in range, so both are latent — reachable by any direct/custom caller of the MEX entry points or a future wrapper change. Local-trigger only.

## Recommendation

`if (idx >= ...size())` in both functions.

## Severity

**Medium** (both) — unchecked one-past-end vector accesses behind an off-by-one guard; two-character fixes. Verified **CONFIRMED** by the second-model pass.

pepone:
Fixes #5552.

### Problem
The MEX endpoint accessors `Ice_ObjectPrx_ice_getEndpoint` and `Ice_ObjectPrx_ice_setEndpoint` guarded their `std::vector` access with `idx > size` instead of `idx >= size`. The indices are 0-based (the MATLAB wrappers pass `i - 1`), so `idx == size` slipped past the check into an out-of-bounds element — an out-of-bounds `shared_ptr` read in the getter and an out-of-bounds `shared_ptr` write (heap corruption) in the setter.

### Fix
Use `idx >= size` in both accessors.

### Testing
The MEX extension builds cleanly. No runtime test was added: the stock MATLAB wrappers never pass `idx == size`, so the defect is latent (reachable only by a direct/custom MEX caller), and a meaningful test would be disproportionate to a two-character fix. Reviewed with codex.

### DIFF
--- a/matlab/src/ObjectPrx.cpp
+++ b/matlab/src/ObjectPrx.cpp
@@ -417,7 +417,7 @@ extern "C"
         try
         {
             auto endpoints = restoreProxy(self)->ice_getEndpoints();
-            if (idx > endpoints.size())
+            if (idx >= endpoints.size())
             {
                 throw std::invalid_argument("index outside range");
             }
@@ -445,7 +445,7 @@ extern "C"
         try
         {
             vector<shared_ptr<Ice::Endpoint>>* v = reinterpret_cast<vector<shared_ptr<Ice::Endpoint>>*>(arr);
-            if (idx > v->size())
+            if (idx >= v->size())
             {
                 throw std::invalid_argument("index outside range");
             }


CHANGELOG.md
```diff
@@ -33,6 +34,11 @@ might need to be aware of.
   `RangeError` instead of being silently ignored, matching the buffer-position behavior of the other language
   mappings.
 
+### MATLAB Changes
+
+- Fixed an out-of-bounds access in the MEX extension's endpoint accessors that occurred when an index equal to
+  the number of endpoints was passed.
```
bernardnormier: The current wording describes the internal trigger condition (an index equal to the endpoint count), but a normal user never passes indices — the stock wrappers (`ice_getEndpoints` / `ice_endpoints`) never pass `idx == size`, so the defect is latent. Suggesting wording that conveys that:

```suggestion
- Fixed a latent out-of-bounds access in the MEX extension's endpoint accessors. The defect
  was not reachable through the standard proxy API.
```

CHANGELOG.md
```diff
@@ -54,6 +54,9 @@ might need to be aware of.
 - Fixed `ice_getCachedConnection` to return an empty `Ice.Connection` array, rather than an empty
   `double` array, when the proxy has no cached connection.
 
+- Fixed a latent out-of-bounds access in the MEX extension's endpoint accessors. The defect
```
bernardnormier: We should actually not include any CHANGELOG entry for such latent (= non user reachable) bug.

bernardnormier (approved):

InsertCreativityHere (approved):
Looks good to me.

Agreed on not writing the CHANGELOG entry here. This is really more "code-hardening" than a "bug-fix" it seems like. `idx == endpoints.size()` is probably never passed.

PR #5614, opened 2026-06-16
Fix memory leak in MATLAB MEX getMajorMinor

### LINKED ISSUE #5553: [Audit-Low] MATLAB MEX getMajorMinor leaks the mxGetProperty copies of major/minor on every version conversion

## Summary

`getMajorMinor` reads the `major` and `minor` properties with `mxGetProperty`, which per MathWorks documentation returns a **freshly allocated copy the caller owns** — and never calls `mxDestroyArray` on either. Both arrays leak on every successful conversion. The correct ownership pattern is in `getIdentity` in the same file, which destroys both property copies after reading them.

## The bug

```cpp
// matlab/src/Util.cpp:28-44
void getMajorMinor(mxArray* p, uint8_t& major, uint8_t& minor)
{
    auto maj = mxGetProperty(p, 0, "major");   // owned copy
    ...
    major = static_cast<uint8_t>(mxGetScalar(maj));
    auto min = mxGetProperty(p, 0, "minor");   // owned copy
    ...
    minor = static_cast<uint8_t>(mxGetScalar(min));
    // neither maj nor min is mxDestroyArray'd
}
```

## Impact

A steady leak of two MATLAB arrays on every encoding/protocol-version conversion — `ice_encodingVersion`, opaque endpoint info, and every other path through `getMajorMinor` — accumulating over a long-lived MATLAB session.

## Source references

- **(the leak)** — [`Util.cpp#L28-L44`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/src/Util.cpp#L28-L44)
- **(correct sibling, `getIdentity`)** — [`Util.cpp#L141-L157`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/matlab/src/Util.cpp#L141-L157)

## Recommendation

`mxDestroyArray(maj);` / `mxDestroyArray(min);` after reading each scalar (mind the early-throw paths).

## Severity

**Low** — a steady, small leak. Verified **CONFIRMED** by the second-model pass against the documented `mxGetProperty` ownership contract. (The IcePy `getVersion` copy-paste bug does **not** recur here — both fields are read from the correct handles.)

pepone:
Fixes #5553.

### Problem
`getMajorMinor` reads the `major` and `minor` properties with `mxGetProperty`, which returns a copy the caller owns, but never called `mxDestroyArray`. Both arrays leaked on every encoding/protocol-version conversion (`ice_encodingVersion`, opaque endpoint info, and every other path through `getMajorMinor`), accumulating over a long-lived MATLAB session.

### Fix
Destroy both copies on all paths, including the early-throw (non-scalar) paths.

### Testing
The MEX extension builds cleanly. No runtime test was added — a steady, small leak is not practical to assert in the test suite. Reviewed with codex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,10 @@ might need to be aware of.
   emitted as bare numeric literals, which MATLAB interprets as `double`, silently losing precision for values
   with magnitude greater than 2^53.
 
+- Fixed a memory leak that occurred each time a proxy was marshaled, unmarshaled, or had its
+  encoding version set with `ice_encodingVersion`. Each of these operations leaked a small amount
+  of memory that accumulated over the lifetime of a MATLAB session.
+
 ### PHP Changes
 
 - Fixed Ice for PHP mis-marshaling a non-empty optional `sequence<string>`.
--- a/matlab/src/Util.cpp
+++ b/matlab/src/Util.cpp
@@ -27,20 +27,25 @@ namespace
 
     void getMajorMinor(mxArray* p, uint8_t& major, uint8_t& minor)
     {
+        // mxGetProperty returns a copy that the caller owns and must destroy.
         auto maj = mxGetProperty(p, 0, "major");
         assert(maj);
         if (!mxIsScalar(maj))
         {
+            mxDestroyArray(maj);
             throw std::invalid_argument("major is not a scalar");
         }
         major = static_cast<uint8_t>(mxGetScalar(maj));
+        mxDestroyArray(maj);
         auto min = mxGetProperty(p, 0, "minor");
         assert(min);
         if (!mxIsScalar(min))
         {
+            mxDestroyArray(min);
             throw std::invalid_argument("minor is not a scalar");
         }
         minor = static_cast<uint8_t>(mxGetScalar(min));
+        mxDestroyArray(min);
     }
 
     // This function converts the cell array input argument and returns a MATLAB string array.


CHANGELOG.md
```diff
@@ -33,6 +34,11 @@ might need to be aware of.
   `RangeError` instead of being silently ignored, matching the buffer-position behavior of the other language
   mappings.
 
+### MATLAB Changes
+
+- Fixed a memory leak in the MEX extension. The `major` and `minor` property copies read during encoding and
+  protocol version conversion are now released.
```
bernardnormier: The entry describes an implementation detail (`major`/`minor` property copies) rather than the impact. I traced the call chain: `getMajorMinor` is only reached via `getEncodingVersion`, whose only three callers are `Ice_ObjectPrx_read` (unmarshaling a proxy), `Ice_ObjectPrx_write` (marshaling a proxy), and `Ice_ObjectPrx_ice_encodingVersion` (setting a proxy's encoding). It's specific to `EncodingVersion` — not protocol-version or endpoint handling. Suggesting wording that captures when the leak actually occurred:

```suggestion
- Fixed a memory leak that occurred each time a proxy was marshaled, unmarshaled, or had its
  encoding version set with `ice_encodingVersion`. Each of these operations leaked a small amount
  of memory that accumulated over the lifetime of a MATLAB session.
```

bernardnormier (approved):

InsertCreativityHere (approved):
Looks good to me!

PR #5618, opened 2026-06-16
Reject certificates without a Subject Key Identifier on SecureTransport

### LINKED ISSUE #5445: SecureTransport: crash (CFRetain(NULL)) when the configured certificate lacks a Subject Key Identifier

## Summary

On macOS (SecureTransport), `findCertificateChain`/`loadCertificate` passes a possibly-null `hash` to `CFDictionarySetValue(query, kSecAttrSubjectKeyID, hash.get())`. With `kCFTypeDictionaryValueCallBacks`, a null value triggers `CFRetain(NULL)` — a documented crash. `hash` is only populated when the certificate exposes a Subject Key Identifier (an optional X.509 extension per RFC 5280), so a cert without an SKI crashes communicator initialization.

## The bug

```cpp
// cpp/src/Ice/SSL/SecureTransportUtil.cpp:433
CFDictionarySetValue(query.get(), kSecAttrSubjectKeyID, hash.get());   // hash may be null
```

`hash` is populated only when `getCertificateProperty(cert, kSecOIDSubjectKeyIdentifier)` returns a property containing a "Key Identifier" label (≈ lines 408-422). A later use of `hash` (≈ line 513) is guarded `if (hash)`; this one is not.

## Impact

On macOS, `IceSSL.CertFile` (PEM/DER) + `IceSSL.KeyFile` referencing a certificate that lacks the optional SKI extension crashes the process during communicator initialization (`CFRetain(NULL)`).

## Source references

- **(Ice / unguarded hash passed to CFDictionarySetValue)** — [`cpp/src/Ice/SSL/SecureTransportUtil.cpp#L408-L433`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/SSL/SecureTransportUtil.cpp#L408-L433)

## Recommendation

Only set `kSecAttrSubjectKeyID` in the query when `hash` is non-null (guard the `CFDictionarySetValue` with `if (hash)`, mirroring the guarded use at line ~513).

## Severity

**Medium** — communicator initialization crashes on macOS when the configured certificate has no Subject Key Identifier extension.

pepone:
Closes #5445.

On macOS (SecureTransport), `loadPrivateKey` passed a possibly-null Subject Key Identifier (`hash`) as the value to `CFDictionarySetValue` on a `kCFTypeDictionaryValueCallBacks` dictionary. When `IceSSL.CertFile` + `IceSSL.KeyFile` reference a certificate that has no SKI extension, `hash` is null and the call aborts the process during communicator initialization. (The actual mechanism on current macOS is an uncaught `NSInvalidArgumentException` via the toll-free `__NSDictionaryM` bridge — not the `CFRetain(NULL)` originally hypothesized in the issue — but the outcome is the same hard abort.)

The SKI is required to import the certificate into the keychain: it is both the keychain lookup key and the key label that pairs the private key with the certificate. A certificate without one therefore can't be imported on macOS, so we now reject it with a clear `CertificateReadException` instead of crashing.

### Notes
- The rejection set is exactly the set that aborts today, so no working macOS configuration is affected.
- This makes macOS stricter than the OpenSSL and Schannel transports, which accept no-SKI certificates. Acceptable since the macOS path crashes today; the error message explains why.
- A separate, same-class null-guard issue on the iOS `findCertificateChain` path (`kSecAttrLabel` from `CFDictionaryGetValue`, ~line 910) will be addressed in a separate PR.
- No regression test: the `IceSSL/configuration` cert fixtures are all generated with an SKI, and regenerating requires rebuilding the entire committed cert tree; a no-SKI fixture wasn't worth it for this one-line guard.

### DIFF
--- a/cpp/src/Ice/SSL/SecureTransportUtil.cpp
+++ b/cpp/src/Ice/SSL/SecureTransportUtil.cpp
@@ -405,7 +405,8 @@ namespace
         // A certificate must expose a Subject Key Identifier to be imported into the keychain: the SKI is used
         // both to locate an already-imported certificate (the query below) and as the key label that pairs the
         // private key with the certificate (further down). Reject a certificate without one with a clear error
-        // instead of passing a null value to CFDictionarySetValue, which aborts the process.
+        // instead of passing a null value to CFDictionarySetValue, which aborts the process. Certificates without
+        // a Subject Key Identifier are uncommon in practice, so rejecting them is acceptable.
         if (!hash)
         {
             throw CertificateReadException(
@@ -504,8 +505,8 @@ namespace
         // kSecKeyLabel attribute should match the subject key identifier.
         //
         vector<SecKeychainAttribute> attributes;
-        if (hash)
         {
+            // hash is guaranteed non-null here: loadPrivateKey rejects certificates without a Subject Key Identifier.
             SecKeychainAttribute attr;
             attr.tag = kSecKeyLabel;
             attr.data = const_cast<UInt8*>(CFDataGetBytePtr(hash.get()));
--- a/cpp/src/Ice/SSL/SecureTransportUtil.cpp
+++ b/cpp/src/Ice/SSL/SecureTransportUtil.cpp
@@ -402,6 +402,19 @@ namespace
             }
         }
 
+        // A certificate must expose a Subject Key Identifier to be imported into the keychain: the SKI is used
+        // both to locate an already-imported certificate (the query below) and as the key label that pairs the
+        // private key with the certificate (further down). Reject a certificate without one with a clear error
+        // instead of passing a null value to CFDictionarySetValue, which aborts the process.
+        if (!hash)
+        {
+            throw CertificateReadException(
+                __FILE__,
+                __LINE__,
+                "SSL transport: the certificate for key file '" + file +
+                    "' has no Subject Key Identifier extension and cannot be imported into the macOS keychain");
+        }
+
         const void* values[] = {keychain};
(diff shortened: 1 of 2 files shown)


cpp/src/Ice/SSL/SecureTransportUtil.cpp
pepone: This `if (hash)` is no longer required.

cpp/src/Ice/SSL/SecureTransportUtil.cpp
```diff
@@ -402,6 +402,19 @@ namespace
             }
         }
 
+        // A certificate must expose a Subject Key Identifier to be imported into the keychain: the SKI is used
+        // both to locate an already-imported certificate (the query below) and as the key label that pairs the
+        // private key with the certificate (further down). Reject a certificate without one with a clear error
+        // instead of passing a null value to CFDictionarySetValue, which aborts the process.
+        if (!hash)
```
pepone, line 410: SKI less certificates are now uncommon.

bernardnormier (approved):
LGTM. The early `if (!hash)` guard correctly prevents the sole abort site — `CFDictionarySetValue(query, kSecAttrSubjectKeyID, hash.get())` passing a NULL value into a `kCFTypeDictionaryValueCallBacks` dictionary — and it covers every way `hash` can end up null (no SKI extension, no "Key Identifier" entry, or a null value). Rejecting is the right call: the SKI is required as both the keychain lookup key and the key-label, and the rejection set is exactly what aborts today, so no working macOS config is affected.

+1 on your own note that the later `if (hash)` guard is now redundant — fine to fold in here or as a quick follow-up.

PR #5622, opened 2026-06-17
Fix C# ObjectAdapter.findAllFacets for unknown identity (#5501)

### LINKED ISSUE #5501: [Audit-Medium] C# ServantManager.findAllFacets throws KeyNotFoundException for an unknown identity — "Can be empty" contract violated

## Summary

`ServantManager.findAllFacets` looks up `_servantMapMap[ident]` with the `Dictionary` indexer, which throws `KeyNotFoundException` for an unknown identity and never returns `null` — so the `is not null` guard and the empty-dictionary fallback below it are dead code. `ObjectAdapter.findAllFacets` documents its return as "Can be empty".

## The bug

```csharp
// csharp/src/Ice/Internal/ServantManager.cs:210-221
internal Dictionary<string, Object> findAllFacets(Identity ident)
{
    lock (_mutex)
    {
        Dictionary<string, Object>? m = _servantMapMap[ident];   // throws KeyNotFoundException
        if (m is not null)                                       // dead
        {
            return new Dictionary<string, Object>(m);
        }
        return new Dictionary<string, Object>();                 // dead
    }
}
```

Every other method in the file (`addServant`, `removeServant`, `findServant`, `removeAllFacets`, `hasServant`) uses `TryGetValue` correctly.

## Impact

`adapter.findAllFacets(id)` for an identity with no registered servants throws `KeyNotFoundException` instead of returning an empty dictionary, contrary to the documented contract. `Instance.findAllAdminFacets` relies on the same path.

## Source references

- **(throwing indexer, dead guards)** — [`csharp/src/Ice/Internal/ServantManager.cs#L210-L221`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/csharp/src/Ice/Internal/ServantManager.cs#L210-L221)
- **("Can be empty" contract)** — [`csharp/src/Ice/ObjectAdapter.cs#L614-L626`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/csharp/src/Ice/ObjectAdapter.cs#L614-L626)

## Recommendation

```csharp
if (_servantMapMap.TryGetValue(ident, out var m))
{
    return new Dictionary<string, Object>(m);
}
return new Dictionary<string, Object>();
```

## Severity

**Medium** — public API throws where its contract promises an empty collection. Verified **CONFIRMED** by the second-model pass.

pepone:
Fixes #5501

`ServantManager.findAllFacets` looked up `_servantMapMap[ident]` with the `Dictionary` indexer, which throws `KeyNotFoundException` for an unknown identity — so the `is not null` guard and empty-dictionary fallback below it were dead code, while `ObjectAdapter.findAllFacets` documents its return as "Can be empty". Switched to `TryGetValue`, matching the file's existing idiom.

Adds a `findAllFacets` regression test in the facets suite (known identity → facets; unknown identity → empty, not an exception). Verified RED→GREEN; independently confirmed by codex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,9 +29,6 @@ might need to be aware of.
   declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
   different from their marshal order could return values in the wrong tuple slots or fail to compile.
 
-- Fixed `ObjectAdapter.findAllFacets` in C# throwing `KeyNotFoundException` for an unknown identity instead of
-  returning an empty dictionary as documented.
-
 ### JavaScript Changes
 
 - Assigning an out-of-range or non-integer value to an `InputStream` or `OutputStream` position now throws a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ might need to be aware of.
   declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
   different from their marshal order could return values in the wrong tuple slots or fail to compile.
 
+- Fixed `ObjectAdapter.findAllFacets` in C# throwing `KeyNotFoundException` for an unknown identity instead of
+  returning an empty dictionary as documented.
+
 ### JavaScript Changes
 
 - Assigning an out-of-range or non-integer value to an `InputStream` or `OutputStream` position now throws a
--- a/csharp/test/Ice/facets/AllTests.cs
+++ b/csharp/test/Ice/facets/AllTests.cs
@@ -96,7 +96,7 @@ public class AllTests : global::Test.AllTests
             // findAllFacets returns all registered facets for a known identity.
             Dictionary<string, Ice.Object> all = adapter.findAllFacets(Ice.Util.stringToIdentity("d"));
             test(all.ContainsKey(""));
-            // For an unknown identity it must return an empty dictionary, not throw (see #5501).
+            // For an unknown identity it must return an empty dictionary, not throw.
             test(adapter.findAllFacets(Ice.Util.stringToIdentity("nonexistent")).Count == 0);
         }
         output.WriteLine("ok");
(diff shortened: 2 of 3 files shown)


csharp/test/Ice/facets/AllTests.cs
```diff
@@ -91,6 +91,16 @@ public class AllTests : global::Test.AllTests
         test(fm[""] == obj3);
         output.WriteLine("ok");
 
+        output.Write("testing findAllFacets... ");
+        {
+            // findAllFacets returns all registered facets for a known identity.
+            Dictionary<string, Ice.Object> all = adapter.findAllFacets(Ice.Util.stringToIdentity("d"));
+            test(all.ContainsKey(""));
+            // For an unknown identity it must return an empty dictionary, not throw (see #5501).
```
pepone:
```suggestion
            // For an unknown identity it must return an empty dictionary, not throw.
```

CHANGELOG.md
```diff
@@ -29,6 +29,9 @@ might need to be aware of.
   declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
   different from their marshal order could return values in the wrong tuple slots or fail to compile.
 
+- Fixed `ObjectAdapter.findAllFacets` in C# throwing `KeyNotFoundException` for an unknown identity instead of
+  returning an empty dictionary as documented.
+
```
bernardnormier: `findAllFacets` is effectively never called in practice, so a user would never observe this bug. Per our CHANGELOG policy that's noise rather than a notable change — suggest dropping the entry entirely:

```suggestion
```

bernardnormier (approved):
LGTM. Fix verified: `_servantMapMap[ident]` threw `KeyNotFoundException` for an unknown identity, contradicting the documented "Can be empty" return of the public `ObjectAdapter.findAllFacets`; `TryGetValue` is correct and the test covers it. One suggestion: drop the CHANGELOG entry, since `findAllFacets` is effectively never called and the change isn't user-observable.

PR #5623, opened 2026-06-17
Fix C# stringToThreadPriority for the "ThreadPriority." prefix form (#5500)

### LINKED ISSUE #5500: [Audit-Medium] C# stringToThreadPriority misuses Substring(start, length) — any "ThreadPriority.*" property value throws

## Summary

`Ice.Internal.Util.stringToThreadPriority` passes the original string length as the **length** (count) argument of `Substring(int startIndex, int length)`, treating it as an end index. For any value beginning with `ThreadPriority.`, `startIndex + length` exceeds the string length and `Substring` throws `ArgumentOutOfRangeException` — unconditionally.

## The bug

```csharp
// csharp/src/Ice/Internal/Util.cs:15-18
if (s.StartsWith("ThreadPriority.", StringComparison.Ordinal))
{
    s = s.Substring("ThreadPriority.".Length, s.Length);   // start=15, length=s.Length → throws
}
```

Intended: the single-argument overload `s.Substring("ThreadPriority.".Length)`.

## Impact

Setting a thread-pool priority in the documented prefixed form — e.g. `Ice.ThreadPool.Server.ThreadPriority=ThreadPriority.Highest` — crashes instead of returning `Highest`. The unprefixed form (`=Highest`) skips the branch and works, which is why the common path doesn't catch this.

## Source references

- **(Substring misuse)** — [`csharp/src/Ice/Internal/Util.cs#L9-L39`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/csharp/src/Ice/Internal/Util.cs#L9-L39)

## Recommendation

Use the single-argument `Substring("ThreadPriority.".Length)` (or `s[15..]`).

## Severity

**Medium** — config-parsing crash in a supported property form; not network-facing. Verified **CONFIRMED** by the second-model pass.

pepone:
Fixes #5500

`Ice.Internal.Util.stringToThreadPriority` passed the original string length as the *length* argument of `Substring(int startIndex, int length)`, treating it as an end index. For any value beginning with `ThreadPriority.` (e.g. a `Ice.ThreadPriority=ThreadPriority.AboveNormal` property), `startIndex + length` exceeded the string length and `Substring` threw `ArgumentOutOfRangeException` unconditionally. Switched to the single-argument `Substring`.

Adds a parsing regression test in the `threadPoolPriority` suite. Verified RED→GREEN; confirmed by codex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ might need to be aware of.
   declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
   different from their marshal order could return values in the wrong tuple slots or fail to compile.
 
+- Fixed the C# parsing of a `ThreadPriority` property (such as `Ice.ThreadPriority`) throwing
+  `ArgumentOutOfRangeException` for any value written with the `ThreadPriority.` prefix.
+
 ### JavaScript Changes
 
 - Assigning an out-of-range or non-integer value to an `InputStream` or `OutputStream` position now throws a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,9 +29,6 @@ might need to be aware of.
   declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
   different from their marshal order could return values in the wrong tuple slots or fail to compile.
 
-- Fixed the C# parsing of a `ThreadPriority` property (such as `Ice.ThreadPriority`) throwing
-  `ArgumentOutOfRangeException` for any value written with the `ThreadPriority.` prefix.
-
 ### JavaScript Changes
 
 - Assigning an out-of-range or non-integer value to an `InputStream` or `OutputStream` position now throws a
--- a/csharp/test/Ice/threadPoolPriority/Client.cs
+++ b/csharp/test/Ice/threadPoolPriority/Client.cs
@@ -8,6 +8,16 @@ public class Client : global::Test.TestHelper
     {
         using Communicator communicator = initialize(ref args);
         TextWriter output = getWriter();
+
+        output.Write("testing stringToThreadPriority parsing... ");
+        // The "ThreadPriority." prefix form must parse (regression test for #5500).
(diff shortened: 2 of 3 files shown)


csharp/test/Ice/threadPoolPriority/Client.cs
```diff
@@ -8,6 +8,16 @@ public override void run(string[] args)
     {
         using Communicator communicator = initialize(ref args);
         TextWriter output = getWriter();
+
+        output.Write("testing stringToThreadPriority parsing... ");
+        // The "ThreadPriority." prefix form must parse (regression test for #5500).
```
pepone:
```suggestion
```

CHANGELOG.md
```diff
@@ -29,6 +29,9 @@ might need to be aware of.
   declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
   different from their marshal order could return values in the wrong tuple slots or fail to compile.
 
+- Fixed the C# parsing of a `ThreadPriority` property (such as `Ice.ThreadPriority`) throwing
+  `ArgumentOutOfRangeException` for any value written with the `ThreadPriority.` prefix.
+
```
bernardnormier: The `Ice.ThreadPriority` property values aren't documented, so in practice nobody configures these thread priorities — the `ThreadPriority.`-prefixed form a user would have to write to hit this is effectively unreachable. Per our CHANGELOG policy this isn't a user-observable change, so suggest dropping the entry:

```suggestion
```

bernardnormier (approved):
LGTM. Fix verified: the two-arg `Substring(startIndex, length)` was passed `s.Length` as the length, so `startIndex + length` always overran and any `ThreadPriority.`-prefixed value threw `ArgumentOutOfRangeException`; single-arg `Substring` is correct and the test covers it. One suggestion: drop the CHANGELOG entry, since the `Ice.ThreadPriority` values are undocumented and effectively unused in practice.

PR #5626, opened 2026-06-17
Fix slice2cs doubled cs:namespace prefix for nested modules (#5478)

### LINKED ISSUE #5478: slice2cs: cs:namespace prefix wrapped around nested modules too — non-compiling C#
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Ice C++ source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source (including against generated `slice2cs` output) by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; please verify the reasoning, reproducibility, and severity before acting on it.

## Summary

`slice2cs` opens the `cs:namespace` prefix block for **every** module, including nested ones. `getNamespacePrefix` walks to the *top-level* module's `cs:namespace` metadata, so it returns the prefix for nested modules too — and nested modules are visited inside the enclosing namespace block. The result is a doubled prefix (`Foo.A.Foo.B`) that does not match the type references computed by `getNamespace()` (`Foo.A.B`), so the generated C# does not compile.

## The bug

```cpp
// cpp/src/slice2cs/CsVisitor.cpp:16-31  (visitModuleStart -> namespacePrefixStart for every module)
// getNamespacePrefix walks to the top-level module's cs:namespace (cpp/src/slice2cs/CsUtil.cpp:16)
```

For `["cs:namespace:Foo"] module A { module B { struct S; } }` the generator emits
`namespace Foo { namespace A { namespace Foo { namespace B { ... }}}}` — `S` lands in `Foo.A.Foo.B`, while references use `Foo.A.B`. Affects both the Ice and IceRPC visitors (shared `CsVisitor::visitModuleStart`).

## Impact

Any Slice file combining `cs:namespace` metadata with nested modules generates non-compiling C#.

## Source references

- **(slice2cs / namespace prefix opened for every module)** — [`cpp/src/slice2cs/CsVisitor.cpp#L16-L31`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2cs/CsVisitor.cpp#L16-L31)
- **(slice2cs / getNamespacePrefix walks to top-level module)** — [`cpp/src/slice2cs/CsUtil.cpp#L16`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2cs/CsUtil.cpp#L16)

## Recommendation

Emit the `cs:namespace` prefix block only for the top-level module (where the metadata is validated to live), not for nested modules.

## Severity

**Medium** — `cs:namespace` combined with nested modules generates non-compiling C# (`cs:namespace` is deprecated but still supported).

pepone:
Fixes #5478

`slice2cs` opened the `cs:namespace` prefix block in `visitModuleStart` for every module, including nested ones. `getNamespacePrefix` walks to the top-level module's `cs:namespace` metadata, so a nested module received the prefix again, producing a doubled namespace (e.g. `Foo.A.Foo.B`) that did not match the references computed by `getNamespace()` (`Foo.A.B`) — the generated C# did not compile. Guarded the prefix block with `Module::isTopLevel()` so it is emitted only at the top-level module.

Adds a nested-module regression test to the `namespacemd` suite (`module Inner { struct S }` under the `cs:namespace`-tagged `WithNamespace`, referenced from `AllTests.cs`). Verified RED→GREEN: without the fix the generated `Namespace.cs` fails to compile (CS0234 — doubled prefix); with it the suite compiles and `WithNamespace.Inner.S` resolves at `Ice.namespacemd.WithNamespace.Inner`. Confirmed by codex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@ might need to be aware of.
   internal dictionary (matching the `Shared` implementation). Code that mutated the returned dictionary to
   update the implicit context must now use `put` or `setContext`.
 
+- Fixed `slice2cs` emitting a doubled `cs:namespace` prefix for modules nested inside a module carrying
+  `cs:namespace` metadata, which produced C# that did not compile.
+
 ### JavaScript Changes
 
 - Assigning an out-of-range or non-integer value to an `InputStream` or `OutputStream` position now throws a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,8 +47,8 @@ might need to be aware of.
   internal dictionary (matching the `Shared` implementation). Code that mutated the returned dictionary to
   update the implicit context must now use `put` or `setContext`.
 
-- Fixed `slice2cs` emitting a doubled `cs:namespace` prefix for modules nested inside a module carrying
-  `cs:namespace` metadata, which produced C# that did not compile.
+- Fixed a bug in `slice2cs` handling of `cs:namespace`: a nested module received the namespace prefix twice
+  (e.g. `Foo.A.Foo.B` instead of `Foo.A.B`), producing C# that did not compile.
 
 ### JavaScript Changes
 
--- a/csharp/test/Ice/namespacemd/AllTests.cs
+++ b/csharp/test/Ice/namespacemd/AllTests.cs
@@ -64,6 +64,10 @@ public class AllTests : global::Test.AllTests
             output.Write("testing types with package... ");
             output.Flush();
             {
+                // A type in a module nested under a cs:namespace-tagged module must resolve at
+                // WithNamespace.Inner.S (single prefix), not a doubled prefix (regression test for #5478).
+                test(typeof(WithNamespace.Inner.S).Namespace == "Ice.namespacemd.WithNamespace.Inner");
+
                 WithNamespace.C1 c1 = initial.getWithNamespaceC2AsC1();
                 test(c1 != null);
                 test(c1 is WithNamespace.C2);
--- a/csharp/test/Ice/namespacemd/AllTests.cs
+++ b/csharp/test/Ice/namespacemd/AllTests.cs
@@ -67,6 +67,9 @@ public class AllTests : global::Test.AllTests
                 // A type in a module nested under a cs:namespace-tagged module must resolve at
                 // WithNamespace.Inner.S (single prefix), not a doubled prefix (regression test for #5478).
                 test(typeof(WithNamespace.Inner.S).Namespace == "Ice.namespacemd.WithNamespace.Inner");
+                test(typeof(global::ZeroC.Foo.Bar.S).Namespace == "ZeroC.Foo.Bar");
+                test(typeof(global::ZeroC.Other.Bar.S).Namespace == "ZeroC.Other.Bar");
+                test(typeof(global::ZeroC.Other.Baz.S).Namespace == "ZeroC.Other.Baz");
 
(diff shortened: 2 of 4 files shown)


csharp/test/Ice/namespacemd/Namespace.ice
```diff
@@ -26,4 +26,14 @@ module WithNamespace
...
+    // A module nested under a cs:namespace-tagged module: its types must map to
+    // WithNamespace.Inner.*, not a doubled-prefix namespace (regression test for #5478).
+    module Inner
+    {
+        struct S
+        {
+            int i;
+        }
+    }
 }
```
pepone, line 39: Would be good to add some additional tests here, something like:

```
["cs:namespace:ZeroC.Foo"]
module Foo
{
    module Bar {
      struct S { string str; }
    }
}


["cs:namespace:ZeroC.Other"]
module Other
{
  module Bar {
    struct S { string str; }
  }
  module Baz {
    struct S { string str; }
  }
}
```

csharp/test/Ice/namespacemd/AllTests.cs
```diff
@@ -64,6 +64,10 @@ public static Test.InitialPrx allTests(TestHelper helper)
             output.Write("testing types with package... ");
             output.Flush();
             {
+                // A type in a module nested under a cs:namespace-tagged module must resolve at
+                // WithNamespace.Inner.S (single prefix), not a doubled prefix (regression test for #5478).
+                test(typeof(WithNamespace.Inner.S).Namespace == "Ice.namespacemd.WithNamespace.Inner");
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Declining this one — the bug being guarded against is a *doubled namespace prefix*, so deriving the expected value from another generated type's namespace could shift with the bug and mask the regression. The explicit literal is intentional here. I did add your additional `ZeroC.Foo`/`ZeroC.Other` cases below. codex agreed.

csharp/test/Ice/namespacemd/Namespace.ice
pepone: 🤖 **Claude:** Added the `["cs:namespace:ZeroC.Foo"]` (nested `Bar`) and `["cs:namespace:ZeroC.Other"]` (nested `Bar` + `Baz`) cases. Verified the generated C# resolves them at single-prefix namespaces (`ZeroC.Foo.Foo.Bar`, `ZeroC.Other.Other.Bar`/`.Baz`) with no doubling, and the suite compiles. (d57a7b3)

CHANGELOG.md
```diff
@@ -29,6 +29,9 @@ might need to be aware of.
   declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
   different from their marshal order could return values in the wrong tuple slots or fail to compile.
 
+- Fixed `slice2cs` emitting a doubled `cs:namespace` prefix for modules nested inside a module carrying
+  `cs:namespace` metadata, which produced C# that did not compile.
```
bernardnormier:
```suggestion
- Fixed a bug in `slice2cs` handling of `cs:namespace`: a nested module received the namespace prefix twice
  (e.g. `Foo.A.Foo.B` instead of `Foo.A.B`), producing C# that did not compile.
```

bernardnormier (approved):
LGTM. Fix and regression test verified. One optional inline suggestion to make the CHANGELOG entry clearer.

PR #5628, opened 2026-06-17
Fix connection leak in OutgoingConnectionFactory

### LINKED ISSUE #5504: [Audit-High] Java OutgoingConnectionFactory.removeConnection uses two-arg Map.remove on a List-valued map — closed connections are never evicted (unbounded leak)

## Summary

`OutgoingConnectionFactory.removeConnection` calls `remove(key, connection)` on `MultiHashMap`, which extends `HashMap<K, List<V>>` and defines only `putOne` — no `removeOne`. The calls therefore bind to the JDK default `Map.remove(Object key, Object value)`, which removes the mapping only if the currently mapped value `.equals(value)`. The mapped value is a `List<ConnectionI>` and the argument is a single `ConnectionI`, so the comparison is always false and **nothing is ever removed**. Closed client connections stay in `_connections` and `_connectionsByEndpoint` until communicator destruction.

## The bug

```java
// java/src/com.zeroc.ice/.../Ice/OutgoingConnectionFactory.java:20-31
private static class MultiHashMap<K, V> extends HashMap<K, List<V>> {
    public void putOne(K key, V value) {
        List<V> list = this.get(key);
        if (list == null) { list = new LinkedList<>(); this.put(key, list); }
        list.add(value);
    }
    // only putOne is defined; no removeOne
}

// :524-531
private synchronized void removeConnection(ConnectionI connection) {
    if (_destroyed) { return; }
    _connections.remove(connection.connector(), connection);                       // (A)
    _connectionsByEndpoint.remove(connection.endpoint(), connection);              // (B)
    _connectionsByEndpoint.remove(connection.endpoint().compress(true), connection); // (C)
}
```

(A)/(B)/(C) are the conditional `HashMap.remove(Object, Object)` overload; `List.equals(ConnectionI)` is never true. The clear intent (mirroring `putOne`) was to remove the single `connection` from the per-key list.

## Impact

`removeConnection` is the `removeFromFactory` callback registered with every outgoing `ConnectionI` at construction, invoked from `ConnectionI.finish` after reaching `StateFinished` — i.e. on **every** normal client-side connection closure. Every closed connection (and its connector/endpoint keys) is retained for the communicator's lifetime — an unbounded memory leak for clients with connection churn (short-lived servers, idle timeouts). `findConnection`/`findConnectionByEndpoint` filter on `isActiveOrHolding()`, so no dead connection is ever handed out — which masks the functional symptom and leaves only the leak.

## Source references

- **(`MultiHashMap` — only `putOne`)** — [`OutgoingConnectionFactory.java#L20-L31`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/OutgoingConnectionFactory.java#L20-L31)
- **(the three no-op removes)** — [`OutgoingConnectionFactory.java#L524-L531`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/OutgoingConnectionFactory.java#L524-L531)
- **(callback registration)** — [`OutgoingConnectionFactory.java#L337-L358`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/OutgoingConnectionFactory.java#L337-L358)
- **(invoked from `ConnectionI.finish`)** — [`ConnectionI.java#L916-L932`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectionI.java#L916-L932)

## Recommendation

Add a `removeOne(K, V)` to `MultiHashMap` that fetches the list, removes the single value, and drops the key when the list becomes empty; call it from `removeConnection`. (The C++/C# audit cross-check note: this `Map.remove(k,v)` overload trap is Java-specific; a codebase grep for other multimap patterns with an add-one helper but no remove-one is cheap insurance.)

## Severity

**High** — unbounded `ConnectionI` leak over the communicator lifetime on the standard client connection-close path. Verified **CONFIRMED** by the second-model pass; ranked first in both priority lists.

pepone:
Fixes #5504.

### Problem
`MultiHashMap<K,V>` (a private nested class in `OutgoingConnectionFactory`) extends `HashMap<K, List<V>>` and defined only `putOne` — no `removeOne`. `removeConnection` therefore called the JDK default `Map.remove(Object key, Object value)`, which removes a mapping only if the currently-mapped value `.equals(value)`. The mapped value is a `List<ConnectionI>` and the argument is a single `ConnectionI`, so the comparison is always false and **nothing was ever removed**. `removeConnection` runs on every normal client connection close, so every closed connection (and its connector/endpoint keys) was retained for the communicator's lifetime — an unbounded leak.

### Fix
Add `MultiHashMap.removeOne(K, V)` (mirroring `putOne`) that removes the single value from the per-key list and drops the key when the list becomes empty, and call it from `removeConnection`.

### Testing
The leak lives in private factory state with no public observability, so a dedicated test would be disproportionate; verified by inspection and codex review. `./gradlew assemble check rewriteDryRun` passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ might need to be aware of.
 - [Changes in Ice 3.9.0](#changes-in-ice-390)
   - [General Changes](#general-changes)
   - [C# Changes](#c-changes)
+  - [Java Changes](#java-changes)
   - [JavaScript Changes](#javascript-changes)
   - [Python Changes](#python-changes)
   - [Ruby Changes](#ruby-changes)
@@ -28,6 +29,11 @@ might need to be aware of.
 - Fixed a bug in `slice2cs --icerpc` where the generated proxy built the response tuple in marshal order while
   declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
   different from their marshal order could return values in the wrong tuple slots or fail to compile.
+
+### Java Changes
+
+- Fixed a connection leak in Ice for Java. Closed outgoing connections were never removed from the connection
+  factory's internal maps, so they accumulated for the lifetime of the communicator.
 
 ### JavaScript Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@ might need to be aware of.
 - [Changes in Ice 3.9.0](#changes-in-ice-390)
   - [General Changes](#general-changes)
   - [C# Changes](#c-changes)
-  - [Java Changes](#java-changes)
   - [JavaScript Changes](#javascript-changes)
   - [Python Changes](#python-changes)
   - [Ruby Changes](#ruby-changes)
@@ -29,11 +28,6 @@ might need to be aware of.
 - Fixed a bug in `slice2cs --icerpc` where the generated proxy built the response tuple in marshal order while
   declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
   different from their marshal order could return values in the wrong tuple slots or fail to compile.
-
-### Java Changes
-
-- Fixed a connection leak in Ice for Java. Closed outgoing connections were never removed from the connection
-  factory's internal maps, so they accumulated for the lifetime of the communicator.
 
 ### JavaScript Changes
 
--- a/changelog.d/java/5504.md
+++ b/changelog.d/java/5504.md
@@ -1,0 +1,2 @@
+- Fixed a memory leak in Ice for Java. Outgoing connections were not released after they closed, so the
+  memory used by a long-running program grew over the lifetime of the communicator.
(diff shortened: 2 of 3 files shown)


InsertCreativityHere (approved):
Looks good to me.

bernardnormier (approved):
LGTM. Verified C++ and C# already evict correctly, so this was Java-only.

PR #5629, opened 2026-06-17
Fix unbounded _replicaGroups growth in IceDiscovery LocatorRegistryI

### LINKED ISSUE #5507: [Audit-Medium] Java IceDiscovery LocatorRegistryI removes the empty replica-group entry by adapterId instead of replicaGroupId — _replicaGroups grows unboundedly

## Summary

In Java IceDiscovery's `LocatorRegistryI.setReplicatedAdapterDirectProxy`, when the last adapter is removed from a replica group, the now-empty `Set` is removed from `_replicaGroups` by **`adapterId`** — but the map is keyed by **`replicaGroupId`** (both the insert branch and the `get` two lines above use `replicaGroupId`). Since the two ids differ in any real deployment, the `remove` is a no-op and the empty entry stays forever. The C++ and C# ports do this correctly — a Java-only divergence.

## The bug

```java
// java/src/com.zeroc.icediscovery/.../IceDiscovery/LocatorRegistryI.java:54-60
_adapters.remove(adapterId);
Set<String> s = _replicaGroups.get(replicaGroupId);   // map is keyed by replicaGroupId
if (s != null) {
    s.remove(adapterId);
    if (s.isEmpty()) {
        _replicaGroups.remove(adapterId);             // BUG: removes by adapterId
    }
}
```

## Impact

Query results stay correct (`findAdapter` enters the replica-group branch with an empty set and returns null), but the entry is never cleaned up — unbounded growth of `_replicaGroups` as replica groups are registered/deregistered with distinct group ids over a long-running discovery locator, and `findObject` iterates the dead empty keys.

## Source references

- **(wrong key)** — [`LocatorRegistryI.java#L39-L60`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.icediscovery/src/main/java/com/zeroc/IceDiscovery/LocatorRegistryI.java#L39-L60)
- **(C++ does it correctly)** — [`cpp/src/IceDiscovery/LocatorI.cpp#L37-L66`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceDiscovery/LocatorI.cpp#L37-L66)
- **(C# does it correctly)** — [`csharp/src/IceDiscovery/LocatorI.cs#L27-L56`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/csharp/src/IceDiscovery/LocatorI.cs#L27-L56)

## Recommendation

`_replicaGroups.remove(replicaGroupId);`

## Severity

**Medium** — unbounded map growth plus dead-key scans in `findObject`; one-token fix. Verified **CONFIRMED** by the second-model pass, cross-checked against the correct C++ and C# implementations.

pepone:
Fixes #5507.

### Problem
In `LocatorRegistryI.setReplicatedAdapterDirectProxy`, when the last adapter was removed from a replica group, the now-empty `Set` was removed from `_replicaGroups` by `adapterId` — but the map is keyed by `replicaGroupId` (the insert branch and the `get` two lines above both use `replicaGroupId`). Since the two ids differ in any real deployment, the `remove` was a no-op and the empty entry stayed forever, growing `_replicaGroups` unbounded (and `findObject` then iterates the dead empty keys).

### Fix
`_replicaGroups.remove(replicaGroupId)`, matching the C++ and C# implementations.

### Testing
A one-token key fix on private discovery state; a dedicated test is disproportionate. Verified by inspection and codex review (cross-checked against the correct C++/C# ports). `./gradlew assemble check rewriteDryRun` passes.

### DIFF
--- a/java/src/com.zeroc.icediscovery/src/main/java/com/zeroc/IceDiscovery/LocatorRegistryI.java
+++ b/java/src/com.zeroc.icediscovery/src/main/java/com/zeroc/IceDiscovery/LocatorRegistryI.java
@@ -55,7 +55,7 @@ class LocatorRegistryI implements LocatorRegistry {
             if (s != null) {
                 s.remove(adapterId);
                 if (s.isEmpty()) {
-                    _replicaGroups.remove(adapterId);
+                    _replicaGroups.remove(replicaGroupId);
                 }
             }
         }


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ might need to be aware of.
   - [General Changes](#general-changes)
   - [C++ Changes](#c-changes)
   - [C# Changes](#c-changes-1)
+  - [Java Changes](#java-changes)
```
bernardnormier, line 12: Drop this TOC line too, since we're removing the only Java entry this PR adds (otherwise it points at an empty section).

```suggestion
```

CHANGELOG.md
```diff
@@ -52,6 +53,11 @@ might need to be aware of.
 - Fixed a bug in `slice2cs` handling of `cs:namespace`: a nested module received the namespace prefix twice
   (e.g. `Foo.A.Foo.B` instead of `Foo.A.B`), producing C# that did not compile.
 
+### Java Changes
+
+- Fixed unbounded map growth in Ice for Java's IceDiscovery. An emptied replica-group entry was removed using the
+  adapter id instead of the replica-group id, so the empty entry was never cleaned up.
+
```
bernardnormier, line 67: Let's drop the CHANGELOG entry. In IceDiscovery, `setReplicatedAdapterDirectProxy` is a collocated self-registration of the local communicator's own adapters, so the leaked keys are the local adapters' (typically static) replica-group ids — a bounded handful of stale empty entries, not unbounded growth, and only under adapter churn that's already rare (rarer still with discovery + replica groups). Below the changelog bar. Removing the now-empty section heading along with it.

```suggestion
```

bernardnormier (approved):
LGTM. `_replicaGroups` is keyed by `replicaGroupId`, so the old `remove(adapterId)` was a no-op that left an empty set behind; `remove(replicaGroupId)` matches C++ ([`LocatorI.cpp`] erases the iterator from `find(replicaGroupId)`) and C# (`Remove(replicaGroupId)`). Java-only port bug. Two inline notes on the CHANGELOG.

PR #5630, opened 2026-06-17
Fix NPE in MetricsMap.getMatching when reflective construction fails

### LINKED ISSUE #5510: [Audit-Low] Java MetricsMap.getMatching NPEs when reflective newInstance() fails — only "assert false" handles the failure

## Summary

In `MetricsMap.getMatching`, the only handling for a reflective `newInstance()` failure is `assert false` — a no-op when assertions are disabled (the production default). If construction throws, `e` stays null and `e.attach(helper)` raises a raw `NullPointerException` out of the observer hot path instead of degrading gracefully.

## The bug

```java
// java/src/com.zeroc.ice/.../Ice/MetricsMap.java:385-397
Entry e = _objects.get(key);
if (e == null) {
    try {
        T t = _class.getDeclaredConstructor().newInstance();
        t.id = key;
        e = new Entry(t);
        _objects.put(key, e);
    } catch (Exception ex) {
        assert false;     // no-op with assertions disabled; e stays null
    }
}
e.attach(helper);          // NPE if newInstance() threw
```

The same `assert false`-only pattern recurs in `IceMX.Observer.getObserver` and `ObserverFactory.getObserver`.

## Impact

Latent: the built-in Slice `Metrics` subclasses all have public no-arg constructors, so today's classes don't throw. But a metrics class whose no-arg constructor is inaccessible to reflection (e.g. under module access restrictions) turns a recoverable construction failure into an NPE in the instrumentation path.

## Source references

- **(assert-only catch, then dereference)** — [`MetricsMap.java#L378-L398`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsMap.java#L378-L398)

## Recommendation

Return null (and have callers skip attaching) or log and degrade gracefully when reflective construction fails, instead of falling through to the dereference.

## Severity

**Low** — incorrect failure path, latent with the built-in metrics classes. Verified **CONFIRMED** by the second-model pass.

pepone:
Fixes #5510.

### Problem
In `MetricsMap.getMatching`, the only handling for a reflective `newInstance()` failure was `assert false` — a no-op when assertions are disabled (the production default). If construction threw, `e` stayed null and the following `e.attach(helper)` raised a `NullPointerException` out of the observer path.

### Fix
Return `null` on the failure so callers skip attaching. The `IceMX.Observer` and `IceMX.ObserverFactory` callers already null-check the returned entry, and the intermediate `SubMap`/typed `getMatching` wrappers simply propagate it.

### Testing
Latent with the built-in metrics classes (all have accessible no-arg constructors), so a dedicated test would need an artificial reflection-hostile metrics class — disproportionate for a Low-severity fix. Verified by inspection (caller null-safety) and codex review. `./gradlew assemble check rewriteDryRun` passes.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsMap.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsMap.java
@@ -391,6 +391,7 @@ public class MetricsMap<T extends Metrics> {
                     _objects.put(key, e);
                 } catch (Exception ex) {
                     assert false;
+                    return null;
                 }
             }
             e.attach(helper);


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsMap.java
```diff
@@ -390,7 +390,10 @@ public Entry getMatching(MetricsHelper<T> helper, Entry previous) {
                     e = new Entry(t);
                     _objects.put(key, e);
                 } catch (Exception ex) {
+                    // Reflective construction of the metrics object failed (e.g. the no-arg
+                    // constructor is inaccessible). Degrade gracefully: callers skip a null entry.
```
bernardnormier: Drop the comment — per our usual convention the rationale belongs in the commit/PR, not the source; `assert false; return null;` is self-explanatory.

```suggestion
```

bernardnormier (approved):
LGTM. With assertions disabled (the production default), a throwing `newInstance()` left `e == null`, so `e.attach(helper)` NPE'd. Returning `null` is safe — the terminal callers (`IceMX.Observer`, `IceMX.ObserverFactory`) null-check the entry and the `SubMap`/typed wrappers propagate it. One nit inline.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsMap.java
InsertCreativityHere: Is it really correct to simply return `null` here?

The fact we had an `assert false` here makes it seem like reaching this block is a serious error.
So changing the behavior to be "we silently ignore when this happens" feels weird to me.
Why not instead rethrow the exception as a `RuntimeException`?

InsertCreativityHere: Or if this behavior is correct and we want to tolerate the case when this happens, we should keep the `return null;` but instead drop the `assert false;` all together I feel like.

PR #5631, opened 2026-06-17
Fix LoggerAdminLoggerI.detach leaving _detached false

### LINKED ISSUE #5511: [Audit-Low] Java LoggerAdminLoggerI.detach() leaves _detached false when no send-log thread was ever started

## Summary

`LoggerAdminLoggerI.detach()` sets `_detached = true` only inside the `if (_sendLogThread != null)` branch, then calls `_loggerAdmin.destroy()` unconditionally. If no remote logger ever triggered the send thread, `detach()` destroys the logger admin but leaves `_detached == false`, so later log calls still enter `LoggerAdminI.log` against the destroyed admin; if a remote logger was attached but no prior message started the thread, logging can even create a new send thread against the destroyed send communicator and wait indefinitely.

This is the Java analogue of the C++ `LoggerAdminLoggerI::detach` finding (more severe in C++); tracked separately because the consequences differ.

## The bug

```java
// java/src/com.zeroc.ice/.../Ice/LoggerAdminLoggerI.java:62-86
public void detach() {
    Thread thread = null;
    synchronized (this) {
        if (_sendLogThread != null) {   // _detached only set when a thread exists
            thread = _sendLogThread;
            _sendLogThread = null;
            _detached = true;
            notifyAll();
        }
    }
    ...
    _loggerAdmin.destroy();             // unconditional
}
```

## Impact

Inconsistent post-detach state: `print`/`trace`/`warning`/`error` keep building and enqueueing `LogMessage`s for a destroyed admin (a minor queued-message leak), and in the attached-but-idle case a `SendLogThread` can be started against the destroyed send communicator and wait forever because `_detached` is still false. Mostly benign in Java (no `std::terminate` analogue), but wrong lifecycle state.

## Source references

- **(conditional `_detached`, unconditional destroy)** — [`LoggerAdminLoggerI.java#L63-L86`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerAdminLoggerI.java#L63-L86)
- **(post-detach logging path)** — [`LoggerAdminLoggerI.java#L168-L183`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerAdminLoggerI.java#L168-L183), [`LoggerAdminI.java#L150-L167`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerAdminI.java#L150-L167)

## Recommendation

Set `_detached = true` unconditionally in `detach()`.

## Severity

**Low** — inconsistent lifecycle state and a minor leak; no crash path in Java. Verified **CONFIRMED** by the second-model pass.

pepone:
Fixes #5511.

### Problem
`LoggerAdminLoggerI.detach()` set `_detached = true` only inside the `if (_sendLogThread != null)` branch, but called `_loggerAdmin.destroy()` unconditionally. If no send-log thread was ever started, `detach()` destroyed the logger admin yet left `_detached == false`, so later `print`/`trace`/`warning`/`error` calls still ran against the destroyed admin.

### Fix
Set `_detached = true` unconditionally at the top of the `synchronized(this)` block; the `notifyAll()` and thread-join handling stay inside the `if`.

### Testing
Internal lifecycle state with no clean public probe; a dedicated test is disproportionate for this Low-severity fix. Verified by inspection and codex review. `./gradlew assemble check rewriteDryRun` passes.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerAdminLoggerI.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerAdminLoggerI.java
@@ -63,10 +63,10 @@ final class LoggerAdminLoggerI implements LoggerAdminLogger, Runnable {
     public void detach() {
         Thread thread = null;
         synchronized (this) {
+            _detached = true;
             if (_sendLogThread != null) {
                 thread = _sendLogThread;
                 _sendLogThread = null;
-                _detached = true;
                 notifyAll();
             }
         }


bernardnormier (approved):
LGTM. `detach()` destroys the logger admin unconditionally, so `_detached` must be set unconditionally too — otherwise, when no remote logger was ever attached (so `_sendLogThread` was never lazily created), `detach()` left `_detached == false` and a later `print`/`trace`/`warning`/`error` would still call `_loggerAdmin.log(...)` on the destroyed admin. Moving `_detached = true` to the top of the `synchronized` block fixes it; the `notifyAll()` correctly stays in the `if` since the only possible waiter in `run()` is `_sendLogThread` itself. Minor/low-severity, no CHANGELOG entry needed.

PR #5632, opened 2026-06-17
Fix StringUtil.match to match the C++ implementation

### LINKED ISSUE #5505: [Audit-Medium] Java StringUtil.match misuses substring(begin, end) as C++ substr(pos, len) — wildcard patterns with a non-empty tail throw or mismatch

## Summary

`StringUtil.match` is a mechanical port of C++ `IceInternal::match`, where `std::string::substr(pos, len)` takes a **length** as its second argument. Java's `String.substring(beginIndex, endIndex)` takes an **end index**. The port passes lengths as end indexes, so any wildcard pattern with a non-empty tail (`foo*bar`) either throws `StringIndexOutOfBoundsException` or compares the wrong characters.

## The bug

```java
// java/src/com.zeroc.ice/.../Ice/StringUtil.java:533-534
// Make sure end of the strings match
if (!s.substring(endIndex, s.length() - endIndex)
        .equals(pat.substring(beginIndex + 1, pat.length() - beginIndex - 1))) {
    return false;
}
```

Intended: the suffix of `s` from `endIndex` and the part of `pat` after the `*` — i.e. `s.substring(endIndex)` and `pat.substring(beginIndex + 1)` (cf. [`cpp/src/Ice/StringUtil.cpp#L764-L809`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/StringUtil.cpp#L764-L809)).

## Impact

`StringUtil.match("fooXXXbar", "foo*bar", false)` evaluates `substring(6, 3)` (`beginIndex > endIndex`) and throws `StringIndexOutOfBoundsException`; in other shapes it silently mismatches. The bug is latent for the sole in-tree caller — `MetricsAdminI.validateProperties` uses only trailing-`*` patterns (`"Accept.*"`), which return early at the prefix check — but `match` is a public method.

## Source references

- **(lengths passed as end indexes)** — [`StringUtil.java#L503-L539`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/StringUtil.java#L503-L539)
- **(sole in-tree caller, early-returns on trailing-`*`)** — [`MetricsAdminI.java#L23-L36`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsAdminI.java#L23-L36)

## Recommendation

`s.substring(endIndex)` and `pat.substring(beginIndex + 1)`.

## Severity

**Medium** — public utility throws or mismatches for wildcard patterns with a tail. Verified **CONFIRMED** by the second-model pass.

pepone:
Fixes #5505.

### Problem
`StringUtil.match` is a port of C++ `IceInternal::match` with two divergences from it:
1. The suffix comparison passed lengths to `String.substring`, which takes an end index (C++ `substr` takes a length). A wildcard pattern with a non-empty tail (e.g. `foo*bar`) threw `StringIndexOutOfBoundsException` or compared the wrong characters. (This is the reported bug.)
2. A Java-only `endLength == 0` fast path returned `true` for any trailing wildcard, so `match("foo", "foo*", false)` returned `true` whereas C++ returns `false` — a wildcard must consume at least one character when `emptyMatch` is false.

### Fix
- Use `s.substring(endIndex)` and `pat.substring(beginIndex + 1)`.
- Remove the `endLength == 0` fast path so the `emptyMatch` check applies to trailing wildcards too. `match()` now behaves exactly like the C++ implementation.

The only in-tree caller, `MetricsAdminI.validateProperties`, uses `emptyMatch=false` with trailing-`*` patterns; the second change only affects a degenerate property whose key is exactly the wildcard prefix (e.g. `…Accept.` with nothing after the dot), which is now correctly rejected.

### Testing
Added a `checking string matching...` section to the `IceUtil/inputUtil` test covering exact, leading, trailing, and mid-pattern wildcards plus `emptyMatch` semantics. Verified with JDK 21: the test fails against the previous code and passes after the fix. `./gradlew assemble check rewriteDryRun` passes.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/StringUtil.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/StringUtil.java
@@ -518,9 +518,6 @@ public final class StringUtil {
         // Make sure there is something present in the middle to match the wildcard.
         // If emptyMatch is true, allow a match of "".
         int endLength = pat.length() - beginIndex - 1;
-        if (endLength == 0) {
-            return true;
-        }
         if (endLength > s.length()) {
             return false;
         }
@@ -530,8 +527,7 @@ public final class StringUtil {
         }
 
         // Make sure end of the strings match
-        if (!s.substring(endIndex, s.length() - endIndex)
-            .equals(pat.substring(beginIndex + 1, pat.length() - beginIndex - 1))) {
+        if (!s.substring(endIndex).equals(pat.substring(beginIndex + 1))) {
             return false;
         }
 
--- a/java/test/src/main/java/test/IceUtil/inputUtil/Client.java
+++ b/java/test/src/main/java/test/IceUtil/inputUtil/Client.java
@@ -181,5 +181,44 @@ public class Client extends TestHelper {
             test(StringUtil.splitString("a\"b", ":") == null);
         }
         System.out.println("ok");
+
+        System.out.print("checking string matching... ");
+        System.out.flush();
+        {
+            // No wildcard: exact match.
+            test(StringUtil.match("foo", "foo", false));
+            test(!StringUtil.match("foo", "bar", false));
+
+            // Trailing wildcard.
+            test(StringUtil.match("foobar", "foo*", false));
+            test(!StringUtil.match("foobar", "baz*", false));
+
+            // A trailing wildcard must match at least one character unless emptyMatch is true.
+            test(!StringUtil.match("foo", "foo*", false));
+            test(StringUtil.match("foo", "foo*", true));
+
+            // Leading wildcard.
+            test(StringUtil.match("foobar", "*bar", false));
+            test(!StringUtil.match("foobar", "*baz", false));
+
+            // A leading wildcard must also match at least one character unless emptyMatch is true.
+            test(!StringUtil.match("bar", "*bar", false));
+            test(StringUtil.match("bar", "*bar", true));
+
+            // Wildcard with a non-empty tail (regression test for #5505).
+            test(StringUtil.match("fooXXXbar", "foo*bar", false));
+            test(!StringUtil.match("fooXXXbaz", "foo*bar", false));
+
+            // The wildcard must match at least one character unless emptyMatch is true.
+            test(!StringUtil.match("foobar", "foo*bar", false));
+            test(StringUtil.match("foobar", "foo*bar", true));
+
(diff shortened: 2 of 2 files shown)


InsertCreativityHere:
Tangentially related to this PR, the C++ code is actually bogus too?
Actually, it seems like we got the behavior for C++ and Java flipped around??
https://github.com/zeroc-ice/ice/blob/c6c2686611dd18779976e0d53310c09264c0b73a/cpp/src/Ice/StringUtil.cpp#L802-L807 The 2nd parameter is a `count`, which should logically be `s.length() - endIndex`, but we're passing `s.length()`, which is just wrong. I mean, C++ tolerates it and caps it, so it's not a "real bug", but it's still wrong. Realistically, we should just be calling `substr(endIndex)` anyways... There's no point passing the 2nd parameter, it just defaults to `npos`.

Is there an open issue corresponding to the C++ side you know of? Or should I fix it?

pepone:
Good catch. You're right that it's wrong but harmless on the C++ side: `std::string::substr(pos, count)` caps `count` to the remaining length, so `s.substr(endIndex, s.length())` still returns the suffix from `endIndex` — the result is correct, the second argument is just redundant. The clean form is `s.substr(endIndex)` and `pat.substr(beginIndex + 1)`.

No open issue for the C++ side — #5505 only covers the Java port. Go ahead and fix it; a one-line cleanup like that doesn't need its own issue.

bernardnormier (approved):
LGTM. Both fixes are correct and bring `match()` in line with C++:

- The suffix comparison now uses single-arg `substring` (Java `substring(begin,end)` takes an end index, not a length like C++ `substr`); the old code computed `s.substring(endIndex, s.length()-endIndex)`, which for a mid-pattern wildcard like `foo*bar` is `substring(begin>end)` and threw `StringIndexOutOfBoundsException` (the reported bug).
- Dropping the `endLength == 0` fast path lets the existing `(!emptyMatch && endIndex == beginIndex)` guard apply to trailing wildcards, so `match("foo","foo*",false)` is now false.

The added test covers exact/leading/trailing/mid-wildcard and `emptyMatch` semantics. (Copilot's suggested `match("","*",false)` case isn't applicable — `match()` asserts `!s.isEmpty()`.)

InsertCreativityHere (approved):
Looks good to me!

Issue #5503, opened 2026-06-12
[Audit-Medium] C# MetricsAdminI.getMaps<T> reads a reused view's _maps without _mutex — unsafe concurrent Dictionary read/write


pepone:
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the C# source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it. Verified **PARTIAL**: the central unsafe access is confirmed, but the originally-claimed failure mechanism was narrowed (see Severity).

## Summary

`MetricsAdminI.getMaps<T>` enumerates `_views` and calls `v.getMap<T>` without holding `_mutex`, while `updateViews()` (under `_mutex`) mutates the `_maps` dictionary of **reused** `MetricsViewI` instances via `addOrUpdateMap`/`removeMap`. An ordinary `Dictionary` read (`TryGetValue` inside `getMap`) concurrent with a write is undefined behavior in .NET.

## The bug

```csharp
// csharp/src/Ice/Internal/MetricsAdminI.cs:895-907
public List<MetricsMap<T>> getMaps<T>(string mapName) where T : IceMX.Metrics, new()
{
    var maps = new List<MetricsMap<T>>();
    foreach (MetricsViewI v in _views.Values)        // no lock (_mutex)
    {
        MetricsMap<T> map = v.getMap<T>(mapName);    // TryGetValue on a Dictionary that updateViews mutates
        if (map != null)
        {
            maps.Add(map);
        }
    }
    return maps;
}
```

Every other access to `_views` and to `MetricsViewI._maps` is under `lock (_mutex)`. `getMaps<T>` is invoked from `ObserverFactory.update()` (property-/map-update driven) while an admin `setProperties`/`enableMetricsView` call can be running `updateViews()` on another thread.

Note from the verification pass: `updateViews` **replaces** `_views` rather than mutating the enumerated dictionary, so the originally-claimed collection-modified exception on `_views` itself is not established. The confirmed race is on the reused `MetricsViewI` objects: their ordinary `_maps` dictionary is mutated at `addOrUpdateMap` while an unlocked `getMaps` reads it through `getMap`/`TryGetValue`.

## Impact

[...]

PR #5633, opened 2026-06-17
Fix unsynchronized access to the metrics views in MetricsAdminI.getMaps

### LINKED ISSUE #5506: [Audit-Medium] Java MetricsAdminI.getMaps reads a reused view's _maps without the admin monitor — unsafe concurrent HashMap read/write
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Java source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it. Verified **PARTIAL**: the central unsafe access is confirmed; the originally-claimed `ConcurrentModificationException` mechanism was narrowed (see Severity).

## Summary

`MetricsAdminI.getMaps` enumerates `_views` and calls `MetricsViewI.getMap` without holding the admin monitor, while `updateViews()` (under `synchronized(this)`) **reuses** `MetricsViewI` instances and mutates their plain `HashMap` `_maps` via `addOrUpdateMap`/`removeMap`. `getMaps` is invoked from `IceMX.ObserverFactory.update()`, which deliberately runs *after* the admin lock is released — an unsafe concurrent read/write on a non-thread-safe `HashMap`.

This is the Java twin of the C# finding zeroc-ice/ice#5503 — the same defect exists in both ports.

## The bug

```java
// java/src/com.zeroc.ice/.../Ice/MetricsAdminI.java:244-253
public <T extends Metrics> List<MetricsMap<T>> getMaps(String mapName, Class<T> cl) {
    List<MetricsMap<T>> maps = new ArrayList<>();
    for (MetricsViewI v : _views.values()) {        // _views read without lock (not volatile)
        MetricsMap<T> map = v.getMap(mapName, cl);  // reads v._maps (plain HashMap) without lock
        if (map != null) {
            maps.add(map);
        }
    }
    return maps;
}
```

Every other access to `_views` and to each `MetricsViewI._maps` runs under `synchronized(this)` (`updateViews`, `addOrUpdateMap`, `removeMap`).

Note from the verification pass: `_views` is **replaced**, not structurally mutated, and `MetricsViewI.getMap` performs a `HashMap.get` rather than iteration — so the originally-claimed `ConcurrentModificationException` path is not present. The verified defect is the unsafe/stale concurrent lookup on the reused views' `_maps` (and the unlocked, non-`volatile` read of `_views` itself, which has no happens-before edge to its replacement).

## Impact

A metrics configuration update (admin `setProperties` → `updateViews()`) racing `ObserverFactory.update()` on another thread performs unsynchronized `HashMap` read/write — stale or torn lookups in the observer update path.

## Source references

- **(unlocked `getMaps`)** — [`MetricsAdminI.java#L244-L253`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsAdminI.java#L244-L253)
- **(`updateViews` reuses views, mutates `_maps` under the monitor)** — [`MetricsAdminI.java#L80-L135`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsAdminI.java#L80-L135), [`MetricsViewI.java#L17-L61`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsViewI.java#L17-L61)
- **(caller outside the lock)** — [`ObserverFactory.java#L119-L128`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/java/src/com.zeroc.ice/src/main/java/com/zeroc/IceMX/ObserverFactory.java#L119-L128)

## Recommendation

Take `synchronized(this)` around the `_views` enumeration in `getMaps` (snapshot the maps under the lock) — same fix shape as the C# counterpart.

## Severity

**Medium** — confirmed unsafe concurrent `HashMap` read/write between config updates and observer updates; the CME-on-`_views` portion of the original claim was retired by the verification pass.

pepone:
Fixes #5506.

### Problem
`MetricsAdminI.getMaps` enumerated `_views` and read each reused `MetricsViewI`'s plain `HashMap` `_maps` (via `getMap`) without holding the admin monitor, while `updateViews`/`addOrUpdateMap`/`removeMap` mutate them under `synchronized(this)`. `getMaps` is called from `IceMX.ObserverFactory.update()` after the admin lock is released, so a metrics configuration update racing an observer update was an unsafe concurrent `HashMap` read/write.

### Fix
- Snapshot the maps under `synchronized(this)` in `getMaps` (same fix shape as the C# counterpart #5503).
- Drop the redundant method-level `synchronized` on `registerSubMap` — its critical section is already guarded by an inner `synchronized(this)` block — so its `factory.update()` runs **outside** the admin monitor, matching `registerMap` and `unregisterMap`. Without this, the new `getMaps` lock (acquired observer-factory → admin from `ObserverFactory.update()`) would invert with `registerSubMap`'s admin → observer-factory order and could deadlock.

### Testing
A deterministic test for a data race / lock inversion is disproportionate and would be flaky; verified by inspection and two codex review passes — the first caught the lock inversion, which this revision resolves. `./gradlew assemble check rewriteDryRun` passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,8 +60,9 @@ might need to be aware of.
 
 ### Java Changes
 
-- Fixed an unsynchronized access to the metrics views in Ice for Java. `MetricsAdminI.getMaps` now snapshots the
-  metrics maps under the admin monitor, avoiding a data race with concurrent metrics configuration updates.
+- Fixed a data race in the Ice for Java metrics (IceMX) implementation. Reconfiguring the metrics views at
+  runtime while metrics were being collected could corrupt the internal metrics maps or throw a
+  `ConcurrentModificationException`.
 
 ### JavaScript Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ might need to be aware of.
   - [General Changes](#general-changes)
   - [C++ Changes](#c-changes)
   - [C# Changes](#c-changes-1)
+  - [Java Changes](#java-changes)
   - [JavaScript Changes](#javascript-changes)
   - [MATLAB Changes](#matlab-changes)
   - [PHP Changes](#php-changes)
@@ -56,6 +57,11 @@ might need to be aware of.
 
 - Fixed a bug in `slice2cs` handling of `cs:namespace`: a nested module received the namespace prefix twice
   (e.g. `Foo.A.Foo.B` instead of `Foo.A.B`), producing C# that did not compile.
+
+### Java Changes
+
+- Fixed an unsynchronized access to the metrics views in Ice for Java. `MetricsAdminI.getMaps` now snapshots the
+  metrics maps under the admin monitor, avoiding a data race with concurrent metrics configuration updates.
 
 ### JavaScript Changes
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsAdminI.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/MetricsAdminI.java
@@ -208,7 +208,7 @@ public class MetricsAdminI
         }
     }
 
-    public synchronized <S extends Metrics> void registerSubMap(
+    public <S extends Metrics> void registerSubMap(
             String map, String subMap, Class<S> cl, java.lang.reflect.Field field) {
         boolean updated;
         MetricsMapFactory<?> factory;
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -29,6 +30,11 @@ might need to be aware of.
   declaring it in declaration order. As a result, an operation whose out-parameters were declared in an order
   different from their marshal order could return values in the wrong tuple slots or fail to compile.
 
+### Java Changes
+
+- Fixed an unsynchronized access to the metrics views in Ice for Java. `MetricsAdminI.getMaps` now snapshots the
+  metrics maps under the admin monitor, avoiding a data race with concurrent metrics configuration updates.
```
bernardnormier: Suggest leading with the observable symptom rather than the internals (`getMaps` / "admin monitor"):

```suggestion
- Fixed a data race in the Ice for Java metrics (IceMX) implementation. Reconfiguring the metrics views at
  runtime while metrics were being collected could corrupt the internal metrics maps or throw a
  `ConcurrentModificationException`.
```

bernardnormier (approved):
LGTM. The data-race fix is correct (the C# `getMaps` already takes `lock (_mutex)`, so this was a missed lock in the Java port), and removing the redundant method-level `synchronized` on `registerSubMap` correctly keeps `factory.update()` outside the admin monitor — avoiding the OF-monitor → admin-monitor (observer path) vs admin → OF (registerSubMap) inversion that the new `getMaps` lock would otherwise enable. One CHANGELOG wording tweak inline.

PR #5635, opened 2026-06-17
Fix Glacier2 session-creation strand on synchronous verifier failure

bernardnormier:
## Problem

`Glacier2::CreateSession::create()` catches a synchronous `Ice::Exception` from `authorize()` and calls `finished(current_exception())`, which only invokes the caller's exception callback. It does **not** erase the connection from `_pending` or replay the queued `_pendingCallbacks` — that cleanup only happens in `CreateSession::exception()` (the asynchronous failure path).

Because `startCreateSession()` inserts the connection into `_pending` *before* `authorize()` runs, a synchronous throw from `authorize()` (e.g. a collocated/erroring permissions verifier, or a communicator shutting down) permanently strands:

- the connection's `_pending` entry — with no session created, no close callback is registered to remove it, so the connection can never create a session again; and
- any other session-creation requests queued on that connection as `_pendingCallbacks` — they are never re-driven, so those clients hang with no response.

## Fix

Split the `create()` try/catch so the two failure sources are handled distinctly:

- A failure of `startCreateSession()` (which happens *before* the request is added to `_pending`) still routes through `finished()` — there is nothing to clean up.
- A synchronous failure of `authorize()` (after the request was added to `_pending`) now routes through `exception()`, exactly like the asynchronous failure path. `exception()` erases the `_pending` entry via `finishCreateSession(_current.con, nullptr)` and replays the queued `_pendingCallbacks`.

Funnelling *all* exceptions through `exception()` unconditionally would be wrong, since a `startCreateSession()` failure never owned a `_pending` entry; the split keeps each path correct.

## Notes

No behavior change on the success path or the existing asynchronous failure path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ might need to be aware of.
   - [MATLAB Changes](#matlab-changes)
   - [Python Changes](#python-changes)
   - [Ruby Changes](#ruby-changes)
+  - [Ice Service Changes](#ice-service-changes)
 
 ## Changes in Ice 3.9.0
 
@@ -72,6 +73,14 @@ might need to be aware of.
 - Fixed `Ice::Endpoint` comparison in Ice for Ruby. `<=>` compared an endpoint with itself, making `==`/`<=>`
   asymmetric, and the class defined `eql?` without a matching `hash`; equal endpoints now compare consistently and
   can be used as `Hash`/`Set` keys.
+
+### Ice Service Changes
+
+#### Glacier2
+
+- Fixed a Glacier2 router bug where a synchronous failure of the permissions verifier during session creation could
+  permanently strand the client connection: the client could no longer create a session, and any other
+  session-creation requests queued on the same connection hung forever.
 
 These are the changes since the [Ice 3.8.2] release.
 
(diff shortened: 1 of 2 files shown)


bernardnormier:
Closing — on review this isn't a real bug, so there's nothing to fix.

`CreateSession::authorize()` only makes a lambda-based async call on the verifier proxy (`checkPermissionsAsync`/`authorizeAsync`). Ice's async proxy invocations don't raise synchronous exceptions to the caller for invocation errors: `OutgoingAsync::invoke` (`cpp/src/Ice/OutgoingAsync.cpp`) catches any synchronous exception from prepare/marshalling/send and routes it to `abort()`, which — with a lambda exception callback present — delivers it via `invokeExceptionAsync()` to the exception callback (`checkPermissionsException`/`authorizeException` → `CreateSession::exception()`). That path already erases the `_pending` entry and replays the queued callbacks.

So `authorize()` can't throw `Ice::Exception` synchronously. The `catch` in `create()` only ever catches a throw from `startCreateSession()`, which happens before the connection is added to `_pending`; for that path `finished()` is the correct handling, and the original code is correct.

PR #5637, opened 2026-06-17
Fix several IceGrid registry and node bugs

bernardnormier:
This PR bundles four independent, single-file IceGrid fixes.

### 1. Application updates wedge forever on a database error (`Database::finishApplicationUpdate`)

`finishApplicationUpdate()` removed the application from the in-progress `_updating` set only on the `DeploymentException` path. On an `IceDB::LMDBException` (e.g. `MDB_MAP_FULL` when the LMDB map is exhausted) it only logged and rethrew, and a `ServerNotExistException` from `checkUpdate()` matched no handler at all. The orphaned `_updating` entry made `waitForUpdate()` block forever, so every subsequent add/update/remove of that application hung until the registry was restarted. All failure paths now run `finishUpdating()` before rethrowing.

(No CHANGELOG entry: an `LMDBException` such as `MDB_MAP_FULL` is effectively fatal to the registry in practice, so clearing the orphaned entry is correctness hygiene rather than an observable behavior change.)

### 2. Null-deref crash of a starting slave registry (`ReplicaSessionManager::create`)

`create(replica)` waited on `_condVar` for `_master`, intending to block until the manager was initialized — but `_master` is set in the `SessionManager` **constructor**, so the wait returned immediately while `_thread`/`_database` were still null. Since the `InternalRegistryI` servant is live on the already-activated registry adapter before the replica session manager is initialized, a master calling `registerWithReplica` into the slave during its startup window dereferenced null `_thread`/`_database` and crashed the slave. The wait now blocks on `_thread`; the pointers are captured under the lock, and `destroy()` notifies so a parked waiter bails out instead of hanging or crashing during shutdown.

### 3. Registry-wide session-establishment stall (`ReapThread::connectionClosed`)

`connectionClosed()` destroyed sessions while holding the reaper `_mutex`, unlike `run()`/`terminate()` which deliberately destroy outside the lock. `SessionI::destroyImpl` releases the session's allocations; releasing a `session`-allocation-mode server runs `ServerEntry::releasedNoSync` (`sync()` + `waitForSyncNoThrow()`, which waits indefinitely for the node) plus synchronous Glacier2 filter calls. Holding `_mutex` across that blocked the reaper loop and every `ReapThread::add()` — the call every new client/admin/node/replica session registers through — so a client disconnecting while holding such an allocation could stall establishment of all new sessions registry-wide. The reapables are now swapped out under the lock and destroyed after releasing it.

### 4. Spurious empty identity in the Glacier2 session filter (`AllocatableObjectCache`)

`allocated()`/`released()` built the identity-filter update with `Ice::IdentitySeq seq(1)` followed by `push_back`, producing a 2-element sequence `["", realId]` — i.e. they added/removed an **empty** identity to/from the session's Glacier2 `IdentitySet` on every allocate/release. Constructed the sequence with exactly the real identity instead. (No CHANGELOG entry: the empty identity matches no real object and `IdentitySet` add/remove is idempotent, so there is no observable user impact.)

### DIFF
--- a/cpp/src/IceGrid/ReapThread.cpp
+++ b/cpp/src/IceGrid/ReapThread.cpp
@@ -173,15 +173,22 @@ ReapThread::add(const shared_ptr<Reapable>& reapable, chrono::seconds timeout, c
 void
 ReapThread::connectionClosed(const Ice::ConnectionPtr& con)
 {
-    lock_guard lock(_mutex);
-    auto p = _connections.find(con);
-    if (p != _connections.end())
+    set<shared_ptr<Reapable>> reap;
     {
-        for (const auto& reapable : p->second)
+        lock_guard lock(_mutex);
+        auto p = _connections.find(con);
+        if (p == _connections.end())
         {
-            reapable->destroy(false);
+            return;
         }
+        reap.swap(p->second);
         _connections.erase(p);
+    }
+
+    // Destroy the reapables outside the lock, like run() and terminate().
+    for (const auto& reapable : reap)
+    {
+        reapable->destroy(false);
     }
 }
 
--- a/cpp/src/IceGrid/ReapThread.cpp
+++ b/cpp/src/IceGrid/ReapThread.cpp
@@ -173,7 +173,7 @@ ReapThread::add(const shared_ptr<Reapable>& reapable, chrono::seconds timeout, c
 void
 ReapThread::connectionClosed(const Ice::ConnectionPtr& con)
 {
-    set<shared_ptr<Reapable>> reap;
+    set<shared_ptr<Reapable>> reapables;
     {
         lock_guard lock(_mutex);
         auto p = _connections.find(con);
@@ -181,12 +181,12 @@ ReapThread::connectionClosed(const Ice::ConnectionPtr& con)
         {
             return;
         }
-        reap.swap(p->second);
+        reapables.swap(p->second);
         _connections.erase(p);
     }
 
     // Destroy the reapables outside the lock, like run() and terminate().
-    for (const auto& reapable : reap)
+    for (const auto& reapable : reapables)
     {
         reapable->destroy(false);
     }
--- a/cpp/src/IceGrid/ReplicaSessionManager.cpp
+++ b/cpp/src/IceGrid/ReplicaSessionManager.cpp
@@ -295,12 +295,11 @@ void
 ReplicaSessionManager::create(const InternalRegistryPrx& replica)
 {
     shared_ptr<Thread> thread;
-    shared_ptr<Database> database;
     {
         unique_lock lock(_mutex);
 
-        // Wait until the session manager is fully initialized (or destroyed). _thread and _database are set by the
-        // other create() overload once the slave registry is ready.
+        // Wait until the session manager is fully initialized (or destroyed). _thread is set by the other create()
+        // overload once the slave registry is ready.
         _condVar.wait(lock, [this] { return _thread || !_communicator; });
 
         if (!_thread)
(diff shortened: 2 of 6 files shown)


cpp/src/IceGrid/ReplicaSessionManager.cpp
```diff
@@ -294,22 +294,35 @@ ReplicaSessionManager::create(
...
+        thread = _thread;
+        database = _database;
     }
 
     if (replica->ice_getIdentity() != _master->ice_getIdentity())
     {
-        _database->getTraceLevels()->logger->error("can only create sessions with the master replica");
+        database->getTraceLevels()->logger->error("can only create sessions with the master replica");
         return;
     }
```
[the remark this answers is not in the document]
bernardnormier: Good catch — you're right, and it's a regression from this change specifically.

The previous `wait(return _master)` guaranteed `_master` was engaged before the dereference. My new predicate waits on `_thread` instead, and the initializing `create()` overload sets `_thread` unconditionally — it doesn't depend on `_master`. Since `_master` is set only in the `SessionManager` constructor, and only when the communicator has a default locator, a slave without a default locator that received `registerWithReplica` would now dereference a disengaged `_master`. In normal operation a slave always has a default locator (that's how it reaches the master), so this is a misconfiguration edge case — but the old code degraded to a hang there, not a crash.

Fixed by adding a `!_master` guard (log + early return) before the identity comparison.

cpp/src/IceGrid/ReapThread.cpp
```diff
@@ -173,16 +173,23 @@ ReapThread::add(const shared_ptr<Reapable>& reapable, chrono::seconds timeout, c
 void
 ReapThread::connectionClosed(const Ice::ConnectionPtr& con)
 {
-    lock_guard lock(_mutex);
-    auto p = _connections.find(con);
-    if (p != _connections.end())
+    set<shared_ptr<Reapable>> reap;
```
pepone: I think is clear to rename `reap` -> `reapables` like in the comment.

cpp/src/IceGrid/ReplicaSessionManager.cpp
```diff
@@ -294,22 +294,42 @@ ReplicaSessionManager::create(
...
+
+        // Capture under the lock: destroy() clears _thread/_database, so we keep them alive for the calls below.
+        thread = _thread;
+        database = _database;
+    }
+
+    if (!_master)
+    {
+        // No master registry configured (the communicator has no default locator), so there is nothing to do.
+        database->getTraceLevels()->logger->error("cannot create a replica session: no master registry configured");
```
pepone:
```suggestion
         _traceLevels->logger->error("cannot create a replica session: no master registry configured");
```

cpp/src/IceGrid/ReplicaSessionManager.cpp
```diff
@@ -294,22 +294,42 @@ ReplicaSessionManager::create(
...
+        return;
     }
 
     if (replica->ice_getIdentity() != _master->ice_getIdentity())
     {
-        _database->getTraceLevels()->logger->error("can only create sessions with the master replica");
+        database->getTraceLevels()->logger->error("can only create sessions with the master replica");
```
pepone:
```suggestion
         _traceLevels->logger->error("can only create sessions with the master replica");
```

cpp/src/IceGrid/ReplicaSessionManager.cpp
```diff
@@ -294,22 +294,42 @@ ReplicaSessionManager::create(
...
+            // Destroyed before initialization completed.
+            return;
+        }
+
+        // Capture under the lock: destroy() clears _thread/_database, so we keep them alive for the calls below.
+        thread = _thread;
+        database = _database;
+    }
+
+    if (!_master)
```
pepone: Can we just `assert(_master)` here? This should only be called for slaves, and `RegistryI` ensures the master is set in this case (`startImpl` fails for a slave without a default locator).

pepone (approved):

cpp/src/IceGrid/ReapThread.cpp
bernardnormier: renamed

cpp/src/IceGrid/ReplicaSessionManager.cpp
bernardnormier: Fixed as suggested

cpp/src/IceGrid/ReplicaSessionManager.cpp
bernardnormier: Agreed — switched to `assert(_master)`. The invariant holds: a slave can't start without a default locator (`RegistryI::startImpl` returns false otherwise), so `_master` is always engaged when this runs; and a master never sets `_thread`, so it returns at the earlier `!_thread` check before reaching this line. Also dropped the now-unused captured `database` (your `_traceLevels` change was its last user).

PR #5639, opened 2026-06-17
Fix IceGrid node deadlock between server and adapter operations

bernardnormier:
### Lock-order inversion (deadlock)

`ServerI` owns its object adapters and the one unavoidable cross-lock nesting runs that way: `ServerI::update()` holds the `ServerI` `_mutex` and, via `updateImpl()`, calls `ServerAdapterI::destroy()`, which takes the adapter `_mutex`. So the intended order is **`ServerI::_mutex` before `ServerAdapterI::_mutex`**.

Several `ServerAdapterI` methods violated it by calling back into `ServerI` (which acquires the `ServerI` mutex) **while holding the adapter mutex**:

| Method | Call under the adapter `_mutex` |
|---|---|
| `setDirectProxy` | `getState()`, `adapterActivated()` / `adapterDeactivated()` |
| `activateAsync` | `isAdapterActivatable()`, `getState()` |
| `getDirectProxy` | `isAdapterActivatable()` |
| `updateEnabled` | `isEnabled()` |

Each is the inverse order (adapter mutex, then `ServerI` mutex) and forms an AB-BA cycle with `update() -> updateImpl() -> destroy()`, deadlocking the node and hanging all further operations on it.

The originally reported instance is `setDirectProxy` vs `activate`: `setDirectProxy` held the adapter mutex across `adapterDeactivated()`, which blocks until the server leaves the `Activating` state — and the only code that advances that state, `activate()`, needs the adapter mutex (via `clear()`):

```cpp
_condVar.wait(lock, [this] { return _state != InternalServerState::Activating; }); // adapterDeactivated, holds nothing of the adapter
for (const auto& adpt : adpts) { adpt.second->clear(); }                           // activate(), needs the adapter _mutex
```

### Companion lost wakeup

`ServerI::setStateNoSync` only notified `_condVar` on the `Destroyed` transition. The successful activation path notifies explicitly in `activate()`, but the activation-**failure** path (`Activating -> Deactivating -> Inactive`) never notified, so a thread parked in `adapterDeactivated()`/`terminated()` waiting for `_state != Activating` was never woken when activation failed.

## Fix

Establish a single lock order — **`ServerI::_mutex` before `ServerAdapterI::_mutex`, never the reverse** — so no `ServerAdapterI` method calls into `ServerI` while holding the adapter mutex:

- **`setDirectProxy`** — snapshot `getState()` before taking the adapter mutex; call `adapterActivated()`/`adapterDeactivated()` after releasing it.
- **`activateAsync` / `getDirectProxy` / `updateEnabled`** — query the needed `ServerI` state (`getState()` / `isAdapterActivatable()` / `isEnabled()`) before taking the adapter mutex, then use the snapshot under the lock.
- **`ServerI::setStateNoSync`** — notify `_condVar` whenever the server leaves the `Activating` state (not only on `Destroyed`), so the activation-failure path wakes parked waiters.
- Documented the lock order above the `_mutex` field in `ServerI.h` and `ServerAdapterI.h`.

## Notes

Snapshotting the server state before the lock means it is read slightly earlier than before; these reads were already best-effort (the original `setDirectProxy` called `getState()` three times non-atomically). If the server is `Destroyed`, `getState()` throws `ObjectNotExistException` — at least as safe as the prior behavior.

### DIFF
--- a/cpp/src/IceGrid/ServerAdapterI.cpp
+++ b/cpp/src/IceGrid/ServerAdapterI.cpp
@@ -68,8 +68,7 @@ ServerAdapterI::activateAsync(
         {
             return;
         }
-        _activateAfterDeactivating =
-            serverState >= ServerState::Deactivating && serverState < ServerState::Destroying;
+        _activateAfterDeactivating = serverState >= ServerState::Deactivating && serverState < ServerState::Destroying;
     }
 
     //
--- a/cpp/src/IceGrid/ServerAdapterI.cpp
+++ b/cpp/src/IceGrid/ServerAdapterI.cpp
@@ -33,6 +33,9 @@ ServerAdapterI::activateAsync(
     function<void(exception_ptr)>,
     const Ice::Current&)
 {
+    ServerState serverState = _server->getState();
+    bool activatable = _server->isAdapterActivatable(_id);
+
     {
         lock_guard lock(_mutex);
         if (_enabled && _proxy)
@@ -49,7 +52,7 @@ ServerAdapterI::activateAsync(
             // Nothing else waits for this adapter so we must make sure that this
             // adapter if still activatable.
             //
-            if (!_enabled || !_server->isAdapterActivatable(_id))
+            if (!_enabled || !activatable)
             {
                 response(nullopt);
                 return;
@@ -68,7 +71,7 @@ ServerAdapterI::activateAsync(
             return;
         }
         _activateAfterDeactivating =
-            _server->getState() >= ServerState::Deactivating && _server->getState() < ServerState::Destroying;
+            serverState >= ServerState::Deactivating && serverState < ServerState::Destroying;
     }
 
     //
@@ -104,6 +107,8 @@ ServerAdapterI::activateAsync(
 optional<Ice::ObjectPrx>
 ServerAdapterI::getDirectProxy(const Ice::Current&) const
 {
+    bool activatable = _server->isAdapterActivatable(_id);
+
     lock_guard lock(_mutex);
 
     //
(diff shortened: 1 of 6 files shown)


cpp/src/IceGrid/ServerAdapterI.cpp
```diff
@@ -33,6 +33,9 @@ ServerAdapterI::activateAsync(
     function<void(exception_ptr)>,
     const Ice::Current&)
 {
+    ServerState serverState = _server->getState();
+    bool activatable = _server->isAdapterActivatable(_id);
```
pepone, line 38: I would add a brief comment about why we read this outside the mutex.

cpp/src/IceGrid/ServerAdapterI.cpp
```diff
@@ -104,6 +107,8 @@ ServerAdapterI::activateAsync(
 optional<Ice::ObjectPrx>
 ServerAdapterI::getDirectProxy(const Ice::Current&) const
 {
+    bool activatable = _server->isAdapterActivatable(_id);
```
pepone, line 108: ditto, brief comment about reading without the mutex purporse.

cpp/src/IceGrid/ServerAdapterI.cpp
```diff
@@ -116,72 +121,73 @@ ServerAdapterI::getDirectProxy(const Ice::Current&) const
...
-    if (!_node->allowEndpointsOverride())
     {
-        if (proxy && _proxy)
+        lock_guard lock(_mutex);
+
+        //
+        // We don't allow to override an existing proxy by another non
+        // null proxy if the server is not inactive.
+        //
+        if (!_node->allowEndpointsOverride() && proxy && _proxy && serverState == ServerState::Active)
```
pepone:
```suggestion
        if (!_node->allowEndpointsOverride() && activated && _proxy && serverState == ServerState::Active)
```

cpp/src/IceGrid/ServerAdapterI.cpp
```diff
@@ -116,72 +121,73 @@ ServerAdapterI::getDirectProxy(const Ice::Current&) const
...
-    {
-        _node->observerUpdateAdapter({_id, _proxy});
+        if (_node->getTraceLevels()->adapter > 1)
+        {
+            Ice::Trace out(_node->getTraceLevels()->logger, _node->getTraceLevels()->adapterCat);
+            out << "server '" + _serverId + "' adapter '" << _id << "' " << (_proxy ? "activated" : "deactivated");
+            if (_proxy)
+            {
+                out << ": " << _proxy;
+            }
```
pepone: I find clear to do:


```suggestion
            out << "server '" + _serverId + "' adapter '" << _id << "' ";
            if (activated)
            {
                out << "activated: " << _proxy;
            }
            else
            {
                out << "deactivated";
            }
```

pepone (approved):

cpp/src/IceGrid/ServerAdapterI.cpp
```diff
@@ -116,72 +121,73 @@ ServerAdapterI::getDirectProxy(const Ice::Current&) const
...
-    //
-    if (!_node->allowEndpointsOverride())
     {
-        if (proxy && _proxy)
+        lock_guard lock(_mutex);
+
+        //
+        // We don't allow to override an existing proxy by another non
+        // null proxy if the server is not inactive.
+        //
```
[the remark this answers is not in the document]
bernardnormier: Fixed

cpp/src/IceGrid/ServerAdapterI.cpp
bernardnormier: Fixed as suggested

cpp/src/IceGrid/ServerAdapterI.cpp
bernardnormier: Fixed as suggested

PR #5642, opened 2026-06-17
Fix IceGrid node crash on property update of a process-less server

### LINKED ISSUE #5460: IceGrid ServerI: *_process dereferenced on property update for servers with processRegistered == false

## Summary

`ServerI` dereferences `*_process` (after `assert(_process)`) on the property-update path while the server is `Active`, but a server can reach `Active` without a `_process` when `processRegistered == false`. A no-restart, property-only update on such a running server takes the `_state == Active` branch and dereferences the disengaged `optional<ProcessPrx>` — UB in release builds.

## The bug

```cpp
// cpp/src/IceGrid/ServerI.cpp:1204-1209  (and :2760-2765 setStateNoSync)
_load->addCallback(response, exception);
updateRevision(desc->uuid, desc->revision);
assert(_process);
_load->startRuntimePropertiesUpdate(*_process);   // UB if _process == nullopt
```

`checkActivation()` (`:2525`) explicitly allows reaching `Active` without `_process`, and `NodeCache.cpp:755-776` sets `processRegistered = false` for `Ice.Admin.Enabled=0`, empty `Ice.Admin.Endpoints`, or `ice-version < 3.3`.

## Impact

Deploy a server with `Ice.Admin.Enabled=0`, start it, then apply a no-restart application update changing only a property → `*_process` dereference on a disengaged optional (assert in debug, UB in release).

## Source references

- **(IceGrid / *_process deref on Active property update)** — [`cpp/src/IceGrid/ServerI.cpp#L1204-L1209`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceGrid/ServerI.cpp#L1204-L1209)
- **(IceGrid / processRegistered == false reaches Active)** — [`cpp/src/IceGrid/NodeCache.cpp#L755-L776`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceGrid/NodeCache.cpp#L755-L776)

## Recommendation

Guard the runtime-properties update on `_process`: skip (or handle gracefully) when `_process == nullopt`, rather than asserting and dereferencing.

## Severity

**Medium** — crash / UB on a property-only update of a running server that has no registered process.

pepone:
Guards the runtime-properties update on `_process` so a no-restart, property-only update of a running server with `Ice.Admin.Enabled=0` (which reaches the Active state without a registered process) no longer dereferences a disengaged optional and crashes the node. Includes a `noRestartUpdate` regression test.

Fixes #5460

### DIFF
--- a/changelog.d/service-icegrid/5460.md
+++ b/changelog.d/service-icegrid/5460.md
@@ -1,0 +1,2 @@
+- Fixed a crash in the IceGrid node triggered by a no-restart application update that changed only the
+  properties of a running server whose Ice.Admin object adapter is disabled.
--- a/changelog.d/service-icegrid/5460.md
+++ b/changelog.d/service-icegrid/5460.md
@@ -1,2 +1,0 @@
-- Fixed a crash in the IceGrid node triggered by a no-restart application update that changed only the
-  properties of a running server whose Ice.Admin object adapter is disabled.
--- a/cpp/src/IceGrid/ServerI.cpp
+++ b/cpp/src/IceGrid/ServerI.cpp
@@ -2497,6 +2497,7 @@ ServerI::updateRuntimePropertiesCallback(exception_ptr ex, const shared_ptr<Inte
 AdapterPrxDict
 ServerI::getAdapterProxies() const
 {
+    // assert(locked());
     AdapterPrxDict adapters;
     for (const auto& [id, servant] : _adapters)
     {
@@ -2508,6 +2509,7 @@ ServerI::getAdapterProxies() const
 void
 ServerI::finishLoad()
 {
+    // assert(locked());
     // Completes the pending load command, handing the current server and adapter proxies back to the caller.
     assert(_load);
     assert(_this);
--- a/cpp/src/IceGrid/ServerI.cpp
+++ b/cpp/src/IceGrid/ServerI.cpp
@@ -2497,7 +2497,7 @@ ServerI::updateRuntimePropertiesCallback(exception_ptr ex, const shared_ptr<Inte
 AdapterPrxDict
 ServerI::getAdapterProxies() const
 {
-    // assert(locked());
+    // Must be called with _mutex held.
     AdapterPrxDict adapters;
     for (const auto& [id, servant] : _adapters)
     {
@@ -2509,7 +2509,7 @@ ServerI::getAdapterProxies() const
 void
 ServerI::finishLoad()
 {
-    // assert(locked());
+    // Must be called with _mutex held.
     // Completes the pending load command, handing the current server and adapter proxies back to the caller.
     assert(_load);
     assert(_this);
@@ -2519,7 +2519,7 @@ ServerI::finishLoad()
 bool
 ServerI::checkActivation()
 {
-    // assert(locked());
+    // Must be called with _mutex held.
     if (_state == InternalServerState::WaitForActivation || _state == InternalServerState::ActivationTimeout)
     {
         //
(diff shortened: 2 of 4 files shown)


bernardnormier:
A few things on this one — partly about scope/framing, partly about the fix shape.

**1. Severity / framing.** A null `_process` on a running server really only happens for interop with very old Ice servers that have no `Process` admin facet — and `Ice.Admin.Enabled=0` reaching `Active` without a process is a misconfiguration, not something a user sets out to do. So I'd treat this as a low-severity defensive guard rather than a scenario users hit:
- Drop the `changelog.d/service-icegrid/5460.md` entry — the trigger is a config bug / ancient interop, nothing a user would notice in normal operation.
- The test comment ("A server with `Ice.Admin.Enabled=0` reaches the Active state...") reads as if this is a supported setup; reword to make clear it's an edge/misconfig guard.

**2. The `load()` completion path is inconsistent.** In the new no-process branch we do `addCallback(response, exception)` and then immediately `finishLoad()`, i.e. register a load callback and synchronously drain it in the same breath. The `Activating` branch a few lines up already has the right idiom for "respond now": it calls `response(...)` directly, without touching `_load`. The no-process case should do the same:

```cpp
else if (_state == InternalServerState::Active)
{
    updateRevision(desc->uuid, desc->revision);
    if (_process)
    {
        _load->addCallback(response, exception); // Must be called before startRuntimePropertiesUpdate!
        _load->startRuntimePropertiesUpdate(*_process);
    }
    else
    {
        // No process to push runtime properties to — nothing to update; respond directly.
        assert(_this);
        response(*_this, getAdapterProxies(), secondsToInt(_activationTimeout), secondsToInt(_deactivationTimeout));
    }
}
```

The `setStateNoSync` no-process branch is a genuinely different case (the callback was registered earlier and is legitimately drained on reaching `Active`), so it can keep `finishLoad()`.

**3. Dedup the adapter-proxy dict (Copilot also flagged this).** The `AdapterPrxDict` build is now in three places: the `Activating` branch, the new no-process response, and `finishLoad()`. Extract one small private helper and use it in all three:

```cpp
AdapterPrxDict
ServerI::getAdapterProxies() const
{
    AdapterPrxDict adapters;
    for (const auto& [id, servant] : _adapters)
    {
        adapters.insert({id, servant->getProxy()});
    }
    return adapters;
}
```

Then `finishLoad()` becomes `_load->finished(*_this, getAdapterProxies(), _activationTimeout, _deactivationTimeout);` and the `Activating` branch loses its inline loop too.

bernardnormier (dismissed):
See my detailed comment above — three items: (1) drop the CHANGELOG entry and reframe this as a low-severity guard (config-bug / ancient no-Process-facet interop, not a user scenario), (2) the no-process `load()` branch should respond directly like the Activating branch instead of `addCallback` + synchronous `finishLoad()`, and (3) extract a `getAdapterProxies()` helper to dedupe the three adapter-dict sites.

bernardnormier (commented):
Submitting as a comment to clear the review request — see my detailed notes above (CHANGELOG framing, the no-process `load()` completion path, and the `getAdapterProxies()` dedup). None of it is blocking.

bernardnormier (approved):
Thanks — all three points addressed: CHANGELOG dropped, the no-process `load()` branch now responds directly (with `addCallback` correctly scoped to the `_process` case), and `getAdapterProxies()` dedupes the three sites. The `setStateNoSync` path keeping `finishLoad()` is right.

One tiny nit: the `// assert(locked());` comments on `getAdapterProxies()`/`finishLoad()` reference a `locked()` helper that doesn't exist — a plain `// Must be called with _mutex held.` would read clearer. Non-blocking.

PR #5643, opened 2026-06-17
Detect cyclic variables in IceGrid descriptors

### LINKED ISSUE #5464: IceGrid Resolver::substitute() has no cycle detection — stack overflow on self-referential descriptor variables

## Summary

`Resolver::substitute()` resolves `${var}` references by recursively calling itself with no depth limit or visited-set, so a self-referential or cyclic variable definition recurses until stack overflow. (Property-set references get an explicit circular-dependency check; variable substitution does not.)

## The bug

```cpp
// cpp/src/IceGrid/DescriptorHelper.cpp:827-897  (substitute)
bool param;
string val = getVariable(name, useParams, param);
if (!param)
{
    val = substitute(val, false, useIgnored);   // unbounded recursion, no cycle guard
}
```

Contrast the property-set circular-dependency check at `:936-959`.

## Impact

A descriptor with `<variable name="x" value="${x}"/>` (or `x → y → x`) makes substitution recurse until stack overflow, crashing the registry (or `icegridadmin`) during deployment validation. Triggered by admin-supplied descriptor input.

## Source references

- **(IceGrid / variable substitution lacks cycle detection)** — [`cpp/src/IceGrid/DescriptorHelper.cpp#L827-L897`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceGrid/DescriptorHelper.cpp#L827-L897)

## Recommendation

Track the set of variables currently being expanded (or a recursion depth bound) and raise a descriptive `DeploymentException` on a cycle, mirroring the property-set circular-dependency check.

## Severity

**Medium** — stack-overflow crash from a self-referential/cyclic descriptor variable during deployment.

pepone:
Detects self-referential or cyclic descriptor variables and raises a `DeploymentException`, instead of recursing until the registry runs out of stack.

The check is path-based — a variable is marked in-progress before its value is resolved and unmarked afterward — so it flags only genuine cycles (`${a}` → `${a}`, or mutual A→B→A) and never the legitimate case of a variable referenced more than once (e.g. `${a}${a}`). It is similar in spirit to the existing circular-dependency check for property-set references, but more precise: that one only tracks visited references (without unmarking), so it can over-reject a property set referenced twice via independent parents.

Includes an `update` regression test.

Fixes #5464

### DIFF
--- a/changelog.d/service-icegrid/5464.md
+++ b/changelog.d/service-icegrid/5464.md
@@ -1,0 +1,2 @@
+- Fixed the IceGrid registry to reject a descriptor with a self-referential or cyclic variable definition by
+  raising a deployment error, instead of crashing from unbounded recursion.
--- a/cpp/src/IceGrid/DescriptorHelper.cpp
+++ b/cpp/src/IceGrid/DescriptorHelper.cpp
@@ -822,6 +822,13 @@ Resolver::hasReplicaGroup(const string& id) const
 string
 Resolver::substitute(const string& v, bool useParams, bool useIgnored) const
 {
+    set<string> inProgress;
+    return substitute(v, useParams, useIgnored, inProgress);
+}
+
+string
+Resolver::substitute(const string& v, bool useParams, bool useIgnored, set<string>& inProgress) const
+{
     string value(v);
     string::size_type beg = 0;
     string::size_type end = 0;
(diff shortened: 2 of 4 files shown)


bernardnormier (commented):
LGTM. I traced the full path and it's correct:

- The cycle detection is path-based (insert before recursing, `erase` after), so it catches self-reference and mutual cycles (A→B→A) while correctly *not* false-positiving on legitimate repeated references like `${a}${a}`.
- `substitute()` throwing `std::invalid_argument` is the right local idiom — the other error paths in that function (malformed/undefined variable) already do, and `operator()`'s `catch (const std::exception&)` translates it into the `DeploymentException` the registry returns. So the crash becomes a clean deployment error returned to the client, thrown registry-side in `Database::addApplication`.
- The test triggers the real path by referencing `${cycle}` in `server->exe` (a never-referenced cyclic variable would never recurse).

One small thing on the description: it says this "mirrors the existing circular-dependency check for property-set references," but it's actually *more precise* than that one — `getProperties()` inserts into its `resolved` set without ever erasing (visited-only), so it would over-reject a property set referenced twice via independent parents. The new variable check uses proper on-stack semantics and avoids that. Nothing to change here (the property check is out of scope and may be intentional) — just noting that 'mirrors' undersells it.

bernardnormier (approved):
Approving — correct fix, good test. See my comment above for the detail (and a minor note that the description undersells the change vs the property-set check).

PR #5645, opened 2026-06-17
Fix IceGrid registry crashes in allocatable object cleanup

### LINKED ISSUE #5458: IceGrid AllocatableObjectCache::remove() missing return after 'unknown object' — null deref / assert

## Summary

`AllocatableObjectCache::remove()` logs "can't remove unknown object" when the entry is not found but does **not** return; it falls through into `removeImpl(id)` (which asserts the entry exists) and `entry->getType()` (a null dereference). The analogous `ObjectCache::remove()` returns after logging.

## The bug

```cpp
// cpp/src/IceGrid/AllocatableObjectCache.cpp:160-175
entry = getImpl(id);
if (!entry)
{
    Ice::Error out(_communicator->getLogger());
    out << "can't remove unknown object '" ...;   // no return
}
removeImpl(id);                       // asserts entry exists (Cache.h:80-99); UB in release builds
auto p = _types.find(entry->getType()); // entry == nullptr → null deref
```

## Impact

`ServerCache::removeCommunicator` removing an allocatable id that was never added — e.g. two allocatables deployed with the same identity (the second `add()` logs "duplicate" and is skipped), then the application is removed: the second `remove()` hits this path → assertion failure (debug) or end-iterator dereference / null deref (release). Registry crash.

## Source references

- **(IceGrid / missing return after unknown-object log)** — [`cpp/src/IceGrid/AllocatableObjectCache.cpp#L160-L175`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceGrid/AllocatableObjectCache.cpp#L160-L175)
- **(IceGrid / ObjectCache::remove returns after logging — reference)** — [`cpp/src/IceGrid/ObjectCache.cpp#L97-L102`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceGrid/ObjectCache.cpp#L97-L102)

## Recommendation

`return` after logging the unknown-object error, mirroring `ObjectCache::remove()`.

## Severity

**Medium** — registry crash (assert / null deref) when removing an allocatable id that was never added.

### LINKED ISSUE #5459: IceGrid Allocatable release drain calls request->cancel() on a null request — registry crash

## Summary

In `Allocatable`'s release drain loop, the outer `catch (const Ice::UserException&)` calls `request->cancel(...)` — but entries queued via `tryAllocate` are stored with a **null** request. When such an entry's `canTryAllocate()` throws an `Ice::UserException` that is not `AllocationException` (e.g. `ObjectNotRegisteredException` from a destroyed `AllocatableObjectEntry`), the handler dereferences the null `request`.

## The bug

```cpp
// cpp/src/IceGrid/Allocatable.cpp:256-309
if ((request && allocatable->allocate(request, true)) ||
    (!request && allocatable->canTryAllocate()))   // request may be null here
{ ... }
...
catch (const Ice::UserException&)
{
    request->cancel(current_exception());   // request may be nullptr
}
```

Null-request entries are queued by `queueAllocationAttempt` (`Allocatable.cpp:348`). For `AllocatableObjectEntry`, `canTryAllocate` → `checkAllocatable()` throws `ObjectNotRegisteredException` (a `UserException`, not `AllocationException`, so `tryAllocate`'s internal catch doesn't swallow it) when the entry was destroyed.

## Impact

A server entry released while a `tryAllocate` attempt (null request) is queued on it, and the allocatable object examined by `canTryAllocate` was destroyed by a concurrent application update → null `request->cancel()` → registry crash.

## Source references

- **(IceGrid / null request->cancel in release drain)** — [`cpp/src/IceGrid/Allocatable.cpp#L256-L309`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceGrid/Allocatable.cpp#L256-L309)
- **(IceGrid / null-request entries queued)** — [`cpp/src/IceGrid/Allocatable.cpp#L348`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceGrid/Allocatable.cpp#L348)

## Recommendation

Guard the cancel: only call `request->cancel(...)` when `request` is non-null; for null-request (`tryAllocate`) entries, handle the user exception without dereferencing `request`.

## Severity

**Medium** — registry crash (null dereference) when a queued `tryAllocate` attempt's target is destroyed during release.

pepone:
Two registry null-dereference crash fixes in the allocatable path:
- Only call `request->cancel()` when the request is non-null (`tryAllocate` attempts are queued with a null request).
- Return after logging an unknown-object error in `AllocatableObjectCache::remove()`, mirroring `ObjectCache::remove()`.

Fixes #5459
Fixes #5458

### DIFF
--- a/changelog.d/service-icegrid/icegrid-alloc-nullderef.md
+++ b/changelog.d/service-icegrid/icegrid-alloc-nullderef.md
@@ -1,0 +1,2 @@
+- Fixed a crash in the IceGrid registry that could occur when an allocatable object was removed by an
+  application update while an allocation request for that object was still pending.
--- a/changelog.d/service-icegrid/icegrid-alloc-nullderef.md
+++ b/changelog.d/service-icegrid/icegrid-alloc-nullderef.md
@@ -1,2 +1,3 @@
-- Fixed a crash in the IceGrid registry that could occur when an allocatable object was removed by an
-  application update while an allocation request for that object was still pending.
+- Fixed a crash in the IceGrid registry that could occur when an application declared the same
+  allocatable object identity more than once (which IceGrid tolerates with a logged error) and that
+  application was later updated or removed.
--- a/cpp/src/IceGrid/Allocatable.cpp
+++ b/cpp/src/IceGrid/Allocatable.cpp
@@ -305,7 +305,10 @@ Allocatable::release(const shared_ptr<SessionI>& session, bool fromRelease)
                 }
                 catch (const Ice::UserException&)
                 {
-                    request->cancel(current_exception());
+                    if (request)
+                    {
+                        request->cancel(current_exception());
+                    }
                 }
             }
         }
--- a/cpp/src/IceGrid/Allocatable.cpp
+++ b/cpp/src/IceGrid/Allocatable.cpp
@@ -305,10 +305,10 @@ Allocatable::release(const shared_ptr<SessionI>& session, bool fromRelease)
                 }
                 catch (const Ice::UserException&)
                 {
-                    if (request)
-                    {
-                        request->cancel(current_exception());
-                    }
+                    // Reachable only via the allocate(request, true) branch above; canTryAllocate()
+                    // never throws a UserException, so request is non-null here.
+                    assert(request);
+                    request->cancel(current_exception());
                 }
             }
         }
--- a/cpp/src/IceGrid/AllocatableObjectCache.cpp
+++ b/cpp/src/IceGrid/AllocatableObjectCache.cpp
@@ -168,6 +168,7 @@ AllocatableObjectCache::remove(const Ice::Identity& id)
         {
             Ice::Error out(_communicator->getLogger());
             out << "can't remove unknown object '" << _communicator->identityToString(id) << "'";
+            return;
         }
         removeImpl(id);
 


cpp/src/IceGrid/Allocatable.cpp
```diff
@@ -305,7 +305,10 @@ Allocatable::release(const shared_ptr<SessionI>& session, bool fromRelease)
                 }
                 catch (const Ice::UserException&)
                 {
-                    request->cancel(current_exception());
+                    if (request)
+                    {
+                        request->cancel(current_exception());
+                    }
```
bernardnormier: `if (request)` reads as if the null case can happen, but this outer catch is only reachable from the `allocate(request, true)` branch (which throws `AllocationException`); the `canTryAllocate()` branch never throws a `UserException`, so `request` is non-null here. An `assert` documents the invariant instead of silently tolerating a state that can't occur.

```suggestion
                    // Reachable only via the allocate(request, true) branch above; canTryAllocate()
                    // never throws a UserException, so request is non-null here.
                    assert(request);
                    request->cancel(current_exception());
```

changelog.d/service-icegrid/icegrid-alloc-nullderef.md
```diff
@@ -0,0 +1,2 @@
+- Fixed a crash in the IceGrid registry that could occur when an allocatable object was removed by an
+  application update while an allocation request for that object was still pending.
```
bernardnormier: The reachable crash this PR fixes is the `AllocatableObjectCache::remove` null-deref, which fires when a duplicate allocatable object identity is removed more than once — not when an allocation request is pending (that scenario describes the `Allocatable.cpp` guard, which isn't a demonstrably reachable crash). Suggest describing the actual trigger:

```suggestion
- Fixed a crash in the IceGrid registry that could occur when an application declared the same
  allocatable object identity more than once (which IceGrid tolerates with a logged error) and that
  application was later updated or removed.
```

bernardnormier (approved):
Both fixes are correct. Two inline suggestions: tighten the `Allocatable.cpp` guard to an `assert` (the null case can't actually occur), and rewrite the changelog fragment to describe the reachable trigger.

PR #5649, opened 2026-06-17
Fix out-of-bounds FD_SET in the IceGrid node termination listener

bernardnormier:
On POSIX, the IceGrid node's `Activator` termination listener built an `fd_set` with `FD_SET` over every active server's status-pipe file descriptor, with no `FD_SETSIZE` guard:

```cpp
fd_set fdSet;
FD_ZERO(&fdSet);
FD_SET(_fdIntrRead, &fdSet);
for (const auto& p : _processes)
{
    FD_SET(p.second.pipeFd, &fdSet); // no check against FD_SETSIZE
    ...
}
```

`fd_set` is a fixed-size bitmap of `FD_SETSIZE` (1024) bits. Once any monitored fd reaches 1024, `FD_SET` writes past the end of the structure — undefined behavior / stack corruption — and that server's termination is never detected (its bit can't be represented). A node running a few hundred servers crosses fd 1024 easily, since each server holds several descriptors.

This replaces `select()` with `poll()`, whose `pollfd` array has no `FD_SETSIZE` limit:

- Build a `vector<pollfd>` (interrupt pipe + each server's status pipe) under `_mutex`.
- Poll, retrying on `EINTR`.
- Treat any non-zero `revents` (`POLLIN`/`POLLHUP`/`POLLERR`/`POLLNVAL`) as activity; collect the ready fds into a `set<int>` and process terminations as before.

Behavior is otherwise unchanged; only the readiness mechanism differs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ might need to be aware of.
   - [MATLAB Changes](#matlab-changes)
   - [Python Changes](#python-changes)
   - [Ruby Changes](#ruby-changes)
+  - [Ice Service Changes](#ice-service-changes)
 
 ## Changes in Ice 3.9.0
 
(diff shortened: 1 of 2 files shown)


bernardnormier:
Closing for now.

This converts the IceGrid node's termination listener from `select()` to `poll()` to fix an out-of-bounds `FD_SET` when a monitored status-pipe descriptor is `>= FD_SETSIZE` (1024).

It's a non-trivial change for a rare limitation: it only affects a node running enough servers **concurrently** to exceed 1024 open file descriptors, and only when the process fd soft limit has been raised above 1024 (otherwise the node hits `EMFILE` first). Rather than ship this in a patch release, we'll defer it to the Ice 3.9 development cycle. The branch is preserved, so this PR can be reopened then.

PR #1, opened 2015-06-26
Force correct overload selection to fix compatibility with Intel C++

alexmnazarenko:
Hi. We've managed to build a working version of Ice with intel compiler (Composer XE 2015, Win x64). It works with our (quite complex) app, so hopefully build is correct.

However, it required patch to Ice code which took a while to figure out. Everything just froze on connection establishment. Patch seems to be correct and harmless, so we suggest it to developers to apply in some later version to save trouble to other desperate Intel C++ users.

The problem is in subtle difference in overload selection. In a code:

```C++
#include <iostream>

struct T {}; 
struct PtrT { PtrT(T*) {} };

void f(bool&) { std::cout << "FAIL - bool" << std::endl; }
void f(PtrT) { std::cout << "OK - PtrT" << std::endl; }

int main() {
  f(new T);
}
```

Intel selects "bool&" overload without a warning. Ice source contains few overloads of this type (e.g. IceInternal::OutgoingConnectionFactory::create, Ice::Handle is similar to PtrT in example). The result is awful - pointer to callback gets casted to bool, callback is lost, application freezes.

Hopefully, we got all such places.

Issue #5653, opened 2026-06-18
Proposal: adopt changelog fragments to eliminate CHANGELOG.md merge conflicts


bernardnormier:
## Problem

`CHANGELOG.md` is a single shared file that nearly every PR edits. With many concurrent PRs in flight we hit two recurring failure modes:

1. **Textual merge conflicts** — PRs insert entries at the same place in the same file.
2. **Duplicate section headings / TOC entries** — a PR branched from a `main` that lacked a section re-creates `### <Language> Changes` and its TOC line; when two such PRs target the same section, the second to merge produces a duplicate heading (or a conflict).

Both stem from the same root cause: *PRs edit one shared file*. Better merge discipline doesn't fix it; not editing the shared file does.

We also want to keep the good part of the current workflow: **the changelog entry (or the decision that there is no entry) is written while the PR is fresh**, not reconstructed weeks later at release time.

## Proposal: changelog fragments

Each PR that warrants an entry adds a small file under `changelog.d/`; a script collates the fragments into `CHANGELOG.md` at release time. This is the well-trodden "news fragments" pattern (towncrier, changesets, CPython, pip, Twisted, …).

```
changelog.d/
  cpp/5445.md                    ← issue number (PR fixes a single zeroc-ice/ice issue); one bullet of entry text
  java/5507.md
  java/streamutil-cleanup.md     ← slug (no single public issue)
```

- **No conflicts**: two PRs add different files ⇒ never conflict.
- **No duplicate headings**: section headings + TOC are *generated* from the fragment locations, never authored per-PR.
- **3.8 backports almost free**: the fragment is a file in git, so cherry-picking a fix to the `3.8` branch carries its fragment along, and a collation run on that branch builds `CHANGELOG-3.8.md` — no more hand-copying entries between `CHANGELOG.md` and `CHANGELOG-3.8.md`.
- **Decoupling**: *authoring/deciding the entry* stays at PR time (where the judgment is — symptom-first wording, or a deliberate "no entry"); *assembling the changelog* becomes a mechanical release step. "No entry" = no fragment; a `no-changelog` label lets CI require *either* a fragment *or* the label, so the decision is recorded rather than forgotten.

### The fragment identifier

The filename's leading token only needs to be **unique**:

- **A PR that fixes a single `zeroc-ice/ice` issue** → use that **issue number** (e.g. `changelog.d/java/5507.md`). It exists before the PR and matches the usual `fix<issue>` branch name — no need to invent anything in this common case.
[...]

PR #5654, opened 2026-06-18
Fix IceStorm subscriber creation and event-forwarding robustness

bernardnormier:
Three independent IceStorm robustness fixes, all in `Subscriber.cpp`.

### 1. Per-subscriber publisher servant dispatchable before its subscriber is set

`Subscriber::create` added the per-subscriber publisher servant to the publish adapter (`add()`) **before** assigning its `_subscriber`. The servant identity is deterministic (`topic.<T>.publish.<id>`), so a publisher still holding a publisher proxy from a previous subscription of the same subscriber could dispatch `ice_invoke` in the window between `add()` and `setSubscriber()` and dereference a null `shared_ptr`. The unsynchronized write/read of `_subscriber` was also a data race.

Fix: build the proxy with `createProxy()` and only `add()` the servant to the adapter once its subscriber has been set. This also removes the need for the old cleanup `catch`, and the adapter's own synchronization now establishes the happens-before for `_subscriber`.

### 2. Malformed `retryCount` QoS leaks the per-subscriber servant

`retryCount` was parsed with `stoi()`, which throws `std::invalid_argument` / `std::out_of_range`. The old cleanup handler caught only `Ice::Exception`, so a non-numeric value escaped *after* the servant was registered — leaking it and permanently poisoning that subscriber identity (later valid subscribes fail with `AlreadyRegisteredException` until restart). `retryCount` is now parsed like the other numeric QoS values and a bad value is rejected with `BadQoS` (and, per fix #1, no servant is registered yet).

### 3. `SubscriberLink::flush` misses `std::out_of_range` on the `cost` context

The publisher-supplied `"cost"` request-context value was parsed with `stoi()` catching only `std::invalid_argument`. An out-of-range value (e.g. `"99999999999"`) threw `std::out_of_range`, which discarded the already-swapped-out link events and aborted the publish loop (surfacing as `UnknownException` to the publisher). `std::out_of_range` is now caught and defaults `cost = 0`.

### DIFF
--- a/changelog.d/service-icestorm/icestorm-subscriber-robustness.md
+++ b/changelog.d/service-icestorm/icestorm-subscriber-robustness.md
@@ -1,0 +1,3 @@
+- Fixed an IceStorm bug where subscribing to a topic with an invalid (non-numeric) `retryCount` QoS value left that
+  subscriber unable to be subscribed to the topic until IceStorm was restarted. The value is now rejected with
+  `BadQoS`.
--- a/cpp/src/IceStorm/Subscriber.cpp
+++ b/cpp/src/IceStorm/Subscriber.cpp
@@ -64,6 +64,25 @@ namespace
                 return IceStorm::Instrumentation::SubscriberState::SubscriberStateError;
         }
     }
+
+    // Parses str (ignoring surrounding whitespace) as a base-10 int consumed in full; nullopt otherwise.
+    optional<int> toInt(const string& str)
+    {
+        const string s = IceInternal::trim(str);
+        try
+        {
+            size_t pos = 0;
+            int value = stoi(s, &pos);
+            if (pos == s.size())
+            {
+                return value;
+            }
+        }
+        catch (const std::exception&)
+        {
+        }
+        return nullopt;
+    }
 }
 
 // Each of the various Subscriber types.
@@ -320,14 +339,7 @@ namespace
                 auto q = p->context.find("cost");
                 if (q != p->context.end())
                 {
-                    try
-                    {
-                        cost = stoi(q->second);
-                    }
-                    catch (const std::invalid_argument&)
-                    {
-                        cost = 0;
-                    }
+                    cost = toInt(q->second).value_or(0);
                 }
                 if (cost > _rec.cost)
                 {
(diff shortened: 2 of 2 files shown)


cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -383,103 +387,107 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
...
         }
+        else // if(newObj->ice_isTwoway())
+        {
+            assert(newObj->ice_isTwoway());
+            subscriber = make_shared<SubscriberTwoway>(instance, rec, proxy, retryCount, 5, *newObj);
+        }
+        per->setSubscriber(subscriber);
+
+        // The subscriber is set, so it's now safe to make the servant dispatchable.
+        instance->publishAdapter()->add(per, perId);
```
bernardnormier, line 495: No changelog entry for this fix — it's effectively latent for a normal deployment.

The per-subscriber publisher servant lives on the externally-reachable `IceStorm.Publish` adapter with a *deterministic* identity (`<instanceName>/topic.<T>.publish.<subscriberId>`), so it's reachable in principle. But nothing reaches it in normal operation: publishers use `getPublisher()` (which calls `subscriber->queue()` directly), and the proxy returned by `subscribeAndGetPublisher` is never used as a publish target — not by any test/demo, nor by any internal IceStorm path.

To hit the null `_subscriber`, a client would have to (a) actually publish to that returned per-subscriber proxy, and (b) do so in the `createProxy()`→`add()` window during a concurrent unsubscribe/resubscribe of the same subscriber — the only moment a retained proxy's deterministic identity lines up with a half-initialized servant (on a first subscribe nobody holds the proxy yet). A real but narrow race a normal deployment won't hit, so no entry. The reorder is still worth making: the old ordering, and the unsynchronized `_subscriber` write/read, were genuinely wrong.

changelog.d/service-icestorm/icestorm-subscriber-robustness.md
```diff
@@ -0,0 +1,3 @@
+- Fixed an IceStorm bug where subscribing to a topic with an invalid (non-numeric) `retryCount` QoS value left that
```
bernardnormier, line 1: This is using the new changelog fragment format I proposed.

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -383,103 +387,107 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
...
+        {
+            // Use Oneway in case of Batch Oneway
+            newObj = newObj->ice_oneway();
+        }
+        else if (newObj->ice_isBatchDatagram())
+        {
+            // Use Datagram in case of Batch Datagram
+            newObj = newObj->ice_datagram();
+        }
+        assert(newObj);
```
pepone: What is the point of this assert `newObj` is clearly not null. Is used above and both `ice_oneway/ice_datagram` return non-null proxy.

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -383,103 +387,107 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
...
+            newObj = rec.obj;
+        }
 
-            p = rec.theQoS.find("connectionCached");
-            if (p != rec.theQoS.end())
+        p = rec.theQoS.find("locatorCacheTimeout");
+        if (p != rec.theQoS.end())
+        {
+            istringstream is(IceInternal::trim(p->second));
+            int locatorCacheTimeout;
```
pepone: `stoi` is looser than this — it parses prefixes (`10abc` → 10), so it won't reject trailing junk the way this `eof()` check does. To unify on `stoi` without losing that, use `stoi(value, &pos)` and reject when `pos` didn't consume the whole string (same for `retryCount`).

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -383,103 +387,107 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
...
-                // Use Oneway in case of Batch Oneway
-                newObj = newObj->ice_oneway();
-            }
-            else if (newObj->ice_isBatchDatagram())
+        p = rec.theQoS.find("connectionCached");
+        if (p != rec.theQoS.end())
+        {
+            istringstream is(IceInternal::trim(p->second));
+            int connectionCached;
+            if (!(is >> connectionCached) || !is.eof())
```
pepone: ditto — `stoi(value, &pos)` and reject if `pos` didn't consume the whole string.

pepone (approved):

cpp/src/IceStorm/Subscriber.cpp
bernardnormier: Removed

cpp/src/IceStorm/Subscriber.cpp
bernardnormier: Fixed as suggested - added a `toInt` local helper that uses `stoi(value, &pos)`

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -328,6 +328,10 @@ namespace
                     {
                         cost = 0;
                     }
+                    catch (const std::out_of_range&)
+                    {
+                        cost = 0;
+                    }
```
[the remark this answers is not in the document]
bernardnormier: Fixed indirectly.

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -383,103 +395,104 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
...
-                reliability = p->second;
+                retryCount = *value;
             }
-            if (!reliability.empty() && reliability != "ordered")
+            else
             {
-                throw BadQoS("invalid reliability: " + reliability);
+                throw BadQoS("invalid retry count (numeric value required): " + p->second);
             }
+        }
```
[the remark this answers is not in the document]
bernardnormier: Very minor

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -320,14 +339,7 @@ namespace
...
-                        cost = stoi(q->second);
-                    }
-                    catch (const std::invalid_argument&)
-                    {
-                        cost = 0;
-                    }
+                    cost = toInt(q->second).value_or(0);
                 }
                 if (cost > _rec.cost)
                 {
```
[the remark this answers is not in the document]
bernardnormier: That's fine

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -64,6 +64,25 @@ namespace
...
+            if (pos == s.size())
+            {
+                return value;
+            }
+        }
+        catch (const std::exception&)
+        {
+        }
+        return nullopt;
+    }
```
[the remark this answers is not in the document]
bernardnormier: Not a concern.

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -383,103 +395,104 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
...
+        try
+        {
+            newObj = rec.obj->ice_invocationTimeout(instance->sendTimeout());
+        }
+        catch (const Ice::FixedProxyException&)
+        {
+            // In the event IceStorm is collocated this could be a fixed proxy in which case its not possible to
+            // set the timeout.
+            newObj = rec.obj;
+        }
```
pepone, line 439: Unrelated to the PR fix, maybe is simper to use `ice_isFixed`  and avoid `std::optional` too:

```
Ice::ObjectPrx newObj = rec.obj->ice_isFixed() ?
    rec.obj->ice_invocationTimeout(instance->sendTimeout()) :
    rec.obj;
```

pepone (approved):

PR #5655, opened 2026-06-18
Fix IceStorm topic-operation guards and a shutdown hang

bernardnormier:
Three independent IceStorm fixes.

### 1. Shutdown hang via a lost condition-variable wakeup (`Subscriber.cpp`)

`Subscriber::error()` has an early-return path — taken when an additional error reply arrives while the subscriber is already offline and inside its discard interval — that decrements `_outstanding` but, unlike every other exit path of `error()`/`completed()`, did **not** notify `_condVar` when `_shutdown` is set. If the decrement that brings `_outstanding` to zero takes this path, `Subscriber::shutdown()` waits forever — and it runs while holding `TopicImpl::_subscribersMutex` and `TopicManagerImpl::_mutex`, so `ServiceI::stop()` hangs the whole IceStorm process. The path now notifies the waiter, matching the others.

### 2. `subscribeAndGetPublisher` / `link` miss the `_destroyed` check (`TopicI.cpp`)

`TopicImpl::subscribeAndGetPublisher` and `TopicImpl::link` committed subscriber/link records to LMDB without first checking `_destroyed`, unlike `unlink()` and `destroy()`. A subscribe/link dispatch that serializes after `destroy()` on `_subscribersMutex` commits a subscriber record keyed by the destroyed topic with no placeholder record. On restart `TopicManagerImpl` asserts the first record per topic is a placeholder — so debug builds abort and release builds resurrect the destroyed topic with an orphan subscriber. Both operations now throw `ObjectNotExistException` when the topic is destroyed.

### 3. `TransientTopicImpl::unlink` dereferences a null optional (`TransientTopicI.cpp`)

`unlink` dereferenced its `optional<TopicPrx>` parameter without the `checkNotNull` that `subscribeAndGetPublisher`, `unsubscribe`, and `link` all perform, so a null `Topic*` proxy on the wire was undefined behavior. It now calls `checkNotNull` first.

### DIFF
--- a/changelog.d/service-icestorm/icestorm-topic-guards-shutdown.md
+++ b/changelog.d/service-icestorm/icestorm-topic-guards-shutdown.md
@@ -1,0 +1,4 @@
+- Fixed a hang that could prevent IceStorm from shutting down cleanly when a subscriber still had requests outstanding
+  to an unreachable endpoint.
+- Fixed a race in IceStorm where destroying a topic concurrently with a subscribe or link to the same topic could leave
+  an orphaned record in the database, causing the destroyed topic to reappear the next time IceStorm was restarted.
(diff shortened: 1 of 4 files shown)


pepone (approved):

InsertCreativityHere (approved):
Looks good to me!

PR #5656, opened 2026-06-18
Fix IceStorm election deadlock in NodeI::recovery

bernardnormier:
### The bug

`IceStormElection::NodeI::recovery()` acquires the node's `recursive_mutex` and then waits on `_condVar` while the mutex is held **twice** by the same thread — `recovery()` is called from several operations that already hold `_mutex`. `condition_variable_any::wait` releases the lock exactly once, so the mutex stays held (recursion count drops 2→1) and the wait never returns: no other thread can acquire `_mutex` to drive `_updateCounter` to zero via `finishUpdate()`.

A master replica that loses the majority of its replicas while an update is in flight (`_updateCounter > 0`) therefore deadlocks permanently — all node dispatches block on `_mutex`, and on the `check()` path the election timer thread is the one that parks, killing the election protocol.

`merge()` and `invitation()` already carry the comment *"The recursive mutex (_mutex) must only be locked once by this thread"*; `recovery()` violated it.

### The fix

`recovery()` is split into:

- a **public** `recovery(int64_t)` that acquires `_mutex` once and delegates — used by callers that do **not** already hold the lock (`timeout`, the post-lock paths in `mergeContinue`, `invitation`, and the external `TopicI`/`TopicManagerI` dispatch paths);
- a **private** `recovery(unique_lock<recursive_mutex>&, int64_t)` that runs the recovery logic against an already-held lock — so the condition-variable wait fully releases the singly-held recursive mutex.

The five internal callers that already hold `_mutex` (`start`, `check`, `mergeContinue`'s pre-checks, `startUpdate`, `updateMaster`) now pass their `unique_lock` (four of them converted from `lock_guard`). This matches the documented "lock only once" contract.

### DIFF
--- a/cpp/src/IceStorm/NodeI.h
+++ b/cpp/src/IceStorm/NodeI.h
@@ -10,6 +10,7 @@
 #include "Replica.h"
 
 #include <condition_variable>
+#include <mutex>
 #include <set>
 
 namespace IceStormElection
@@ -58,6 +59,10 @@ namespace IceStormElection
     private:
         void setState(NodeState);
 
+        // Runs the recovery logic with _mutex already held by the caller, so the condition-variable wait can release
+        // it. Callers that don't already hold _mutex use the public recovery(), which acquires it and delegates here.
+        void recovery(std::unique_lock<std::mutex>& lock, std::int64_t generation = -1);
+
         const IceInternal::TimerPtr _timer;
         const std::shared_ptr<IceStorm::TraceLevels> _traceLevels;
         const std::shared_ptr<IceStormElection::Observers> _observers;
@@ -93,8 +98,8 @@ namespace IceStormElection
         IceInternal::TimerTaskPtr _checkTask;
         IceInternal::TimerTaskPtr _mergeContinueTask;
 
-        mutable std::recursive_mutex _mutex;
-        std::condition_variable_any _condVar;
+        mutable std::mutex _mutex;
+        std::condition_variable _condVar;
     };
 
     class FinishUpdateHelper
--- a/changelog.d/service-icestorm/icestorm-recovery-deadlock.md
+++ b/changelog.d/service-icestorm/icestorm-recovery-deadlock.md
@@ -1,0 +1,3 @@
+- Fixed a deadlock in a replicated IceStorm deployment. A master replica that lost the majority of its replicas (for
+  example during a network partition) while a topic update — such as a subscribe or unsubscribe — was in progress
+  could hang, halting all further operations on that node until it was restarted.
--- a/cpp/src/IceStorm/NodeI.cpp
+++ b/cpp/src/IceStorm/NodeI.cpp
@@ -152,18 +152,18 @@ NodeI::start()
     // We use this lock to ensure that recovery is called before CheckTask
     // is scheduled, even if timeout is 0
     //
-    lock_guard lock(_mutex);
+    unique_lock<mutex> lock(_mutex);
 
     _checkTask = make_shared<CheckTask>(shared_from_this());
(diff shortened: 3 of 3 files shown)


pepone (approved):

pepone (approved):
**LGTM — verified correct and complete.**

The bug: `recovery()` reacquired the recursive `_mutex` while the calling op already held it, then `_condVar.wait()`. `condition_variable_any::wait` releases the lock exactly once, so on the doubly-held mutex the recursion count only drops 2→1 — never fully released — so no other thread can take `_mutex` to run `finishUpdate()` and drive `_updateCounter` to 0. Permanent deadlock when a master loses replica majority with an update in flight. Confirmed against `main`.

The split fixes it cleanly: the public `recovery()` acquires once and delegates; the private `recovery(unique_lock&, …)` waits on the caller's singly-held lock (`NodeI.cpp:980`), so the wait fully releases the mutex.

The part I checked hardest is the call-site classification, since any lock-holder still routed through the public overload would just re-introduce the deadlock. It's complete:
- private overload (lock held): `start`, `check`, `mergeContinue` pre-lock, `startUpdate`, `updateMaster`;
- public overload (no lock): `timeout` (342), `mergeContinue` post-lock (567/609/629), `invitation` (782), and the external `node()->recovery(generation)` in `TopicI`/`TopicManagerI`.

I also traced the transitive paths: `ObserverUpdateHelper` and the `CachedRead`/`FinishUpdate` helpers don't hold `_mutex` across the `recovery(generation)` call (`startObserverUpdate`/`finishUpdate` are separate scoped locks, and `startUpdate` releases before the external caller recovers), so nothing reaches the waiting overload doubly-locked.

One minor nit, non-blocking: `assert(lock.owns_lock())` (`NodeI.cpp:968`) is the only guard on the contract and is compiled out under `NDEBUG`; `owns_lock()` also stays `true` if the mutex is locked more than once, so it can't catch a future re-introduced double-lock. Fine to lean on the "lock only once" comments (370/720) + the new header comment — just worth keeping in mind.

*(Cross-checked with codex and an independent verification pass — both agree.)*

bernardnormier:
Note that I replaced the recursive mutex by a regular mutex in a post-initial PR review.

cpp/src/IceStorm/NodeI.h
```diff
@@ -58,6 +59,10 @@ namespace IceStormElection
     private:
         void setState(NodeState);
 
+        // Runs the recovery logic with _mutex already held by the caller, so the condition-variable wait can release
+        // it. Callers that don't already hold _mutex use the public recovery(), which acquires it and delegates here.
+        void recovery(std::unique_lock<std::mutex>& lock, std::int64_t generation = -1);
```
InsertCreativityHere, line 64: Minor, but making this a doc-comment lets it integrate better in C++, whereas regular comments are completely ignored. Since this is private, it makes no different to the API reference.
```suggestion
        /// Runs the recovery logic with _mutex already held by the caller, so the condition-variable wait can release
        /// it. Callers that don't already hold _mutex use the public recovery(), which acquires it and delegates here.
        void recovery(std::unique_lock<std::mutex>& lock, std::int64_t generation = -1);
```

cpp/src/IceStorm/NodeI.h
```diff
@@ -10,6 +10,7 @@
 #include "Replica.h"
 
 #include <condition_variable>
+#include <mutex>
```
InsertCreativityHere, line 13: This is just to make a transitive include, explicit now?

InsertCreativityHere (approved):
Looks good to me.

cpp/src/IceStorm/NodeI.h
bernardnormier: yes

pepone (approved):

PR #5659, opened 2026-06-19
Fix flushBatchRequests hang when a connection closes during flush

bernardnormier:
## Problem

`ConnectionI::sendAsyncRequest` throws the internal `RetryException` (not an `Ice::LocalException`) when the connection is already closed. `CommunicatorFlushBatchAsync::flushConnection` only caught `LocalException`, so a `RetryException`:

1. **leaked the in-flight `_useCount` increment** — the communicator-wide flush future/sent-callback never completed (a permanent hang), even for the healthy connections, and
2. **let an internal exception escape** the public `flushBatchRequests` / `flushBatchRequestsAsync` API.

This is a TOCTOU race: a connection passes the `isActiveOrHolding()` check when the connection list is snapshotted, then is closed (e.g. the peer gracefully closes it) before `flushConnection` runs the flush outside the factory lock.

## Fix

- **C++** and **C#**: catch `RetryException` in `flushConnection`, balance `_useCount` via `check(false)`, and rethrow the wrapped local exception — which `OutgoingConnectionFactory::flushAsyncBatchRequests` already catches and ignores. This mirrors the adjacent `LocalException` catch.
- **JavaScript**: there is no `_useCount` (the communicator flush is a `Promise.all` over per-connection flushes), but `ConnectionFlushBatch` passed the raw `RetryException` to `markFinishedEx`, leaking it out of **both** `Connection.flushBatchRequests` and `Communicator.flushBatchRequests`. It now completes the invocation with the underlying local exception.
- **Java** already handled this correctly (`CommunicatorFlushBatch` catches `RetryException`); no change.

## Notes

- This is the first PR to use the `changelog.d/` fragment layout proposed in #5653 (Option B). Three fragments added, one per affected language.

### DIFF
--- a/changelog.d/cpp/flush-batch-retryexception-hang.md
+++ b/changelog.d/cpp/flush-batch-retryexception-hang.md
@@ -1,0 +1,3 @@
+- Fixed a hang in `Communicator::flushBatchRequests` and `flushBatchRequestsAsync`. When a connection was closed while
+  the communicator was flushing its batch requests, the returned future or completion callback could remain pending
+  forever.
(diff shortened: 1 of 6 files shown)


pepone (approved):

InsertCreativityHere (approved):
Looks good to me.

PR #5666, opened 2026-06-22
Fixed Escaping of `returnValue` Out Parameters in 'slice2swift'

### LINKED ISSUE #5481: slice2swift: inconsistent returnValue escaping between declaration and marshal loop — non-compiling Swift
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Ice C++ source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source (including against generated `slice2swift` output) by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; please verify the reasoning, reproducibility, and severity before acting on it.

## Summary

`slice2swift` uses two different collision-escaping conventions for the synthetic `returnValue` name. `operationReturnDeclaration` escapes via `paramLabel`, which **prepends** an underscore on collision (`_returnValue`); the marshal loop uses the shared `getEscapedParamName`, which **appends** one (`returnValue_`). With an out-parameter named exactly `returnValue`, the servant dispatch destructures `let (iceP__returnValue, iceP_returnValue) = value` but then marshals the undeclared `iceP_returnValue_` — so the generated Swift does not compile.

## The bug

```cpp
// cpp/src/slice2swift/SwiftUtil.cpp:1241-1251  (marshal loop uses getEscapedParamName -> appends _)
// operationReturnDeclaration (:1195) uses paramLabel (:1123-1134) -> prepends _
// shared getEscapedParamName: cpp/src/Slice/Util.h:86
```

The proxy-side unmarshal path is internally consistent; only the servant marshal path breaks.

## Impact

`string op(out string returnValue);` generates servant code declaring `iceP__returnValue` / `iceP_returnValue` but marshaling the undeclared `iceP_returnValue_` → generated Swift does not compile.

## Source references

- **(slice2swift / inconsistent returnValue escaping)** — [`cpp/src/slice2swift/SwiftUtil.cpp#L1241-L1251`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2swift/SwiftUtil.cpp#L1241-L1251)

## Recommendation

Use one consistent escaping convention for the synthetic `returnValue` across `operationReturnDeclaration` and the marshal loop (both prepend or both via `getEscapedParamName`).

## Severity

**Medium** — generated Swift servant code does not compile when an out-parameter is named `returnValue`.

InsertCreativityHere:
This PR fixes #5481 and adds a test.

If an operation had an out parameter named `returnValue`, `slice2swift` would generate uncompilable code.
I don't think this is worth a changelog entry.

The problem is we had 2 escaping functions, which we did not use consistently:
- `paramLabel` which was in `slice2swift` which would return `_returnValue`
- `getEscapedParamName` from `libSlice` which would return `returnValue_`

I deleted `paramLabel` and updated the call sites to use `getEscapedParamName`. This removes some barely used code, and fixes the mismatch in generated parameter names.

### DIFF
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -1121,19 +1121,6 @@ Slice::Swift::writeMarshalUnmarshalCode(
 }
 
 string
-Slice::Swift::paramLabel(const string& param, const ParameterList& params)
-{
-    for (const auto& q : params)
-    {
-        if (q->mappedName() == param)
-        {
-            return "_" + removeEscaping(param);
-        }
-    }
-    return param;
-}
-
-string
 Slice::Swift::operationReturnType(const OperationPtr& op)
 {
     ostringstream os;
@@ -1149,7 +1136,7 @@ Slice::Swift::operationReturnType(const OperationPtr& op)
     {
         if (returnIsTuple)
         {
-            os << paramLabel("returnValue", outParams) << ": ";
+            os << getEscapedParamName(outParams, "returnValue") << ": ";
         }
         os << typeToString(returnType, op, op->returnIsOptional());
     }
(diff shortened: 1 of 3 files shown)


bernardnormier (approved):
LGTM — consolidating on `getEscapedParamName` is the right fix, and agree no changelog entry is needed.

PR #5670, opened 2026-06-22
Fixed `@param` and `@throws` Being Labeled as 'Unknown' Tags

### LINKED ISSUE #5475: Slice: @param/@throws with the description on the next line is rejected as an unknown doc tag (description lost)

## Summary

`parseNamedCommentLine` rejects a `@param`/`@throws`/`@exception` line whose name runs to the end of the line (no trailing whitespace) by returning `false`. Because `DocComment::createUnparsed` right-trims every line, a comment written with the description on the *next* line (`/// @param id` then `/// The identifier.`) has no whitespace after `id`, so the line is treated as an unknown tag — producing a bogus "ignoring unknown doc tag '@param'" warning and silently discarding the following description.

## The bug

```cpp
// cpp/src/Slice/DocCommentParser.cpp:204-233  (parseNamedCommentLine)
auto nameEnd = line.find_first_of(ws, nameStart);
if (nameEnd == string::npos)
{
    return false;   // name runs to end of line -> rejected
}
```

Lines are right-trimmed in `DocComment::createUnparsed` (`Parser.cpp:390`); the rejected line then falls into the unknown-tag branch (`:690-703`). `formatParamrefElement` (`:390-395`) already handles "name runs to end of line" correctly, so the convention is supported elsewhere.

## Impact

Doc comments that put the parameter/exception description on the following line emit a misleading "unknown doc tag" warning and the description is dropped from the generated documentation.

## Source references

- **(Slice / @param name-to-EOL rejected)** — [`cpp/src/Slice/DocCommentParser.cpp#L204-L233`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Slice/DocCommentParser.cpp#L204-L233)

## Recommendation

Accept a name that runs to end-of-line (treat the rest of the line as empty description and continue), as `formatParamrefElement` does.

## Severity

**Medium** — valid doc comments produce spurious warnings and silently lose `@param`/`@throws` descriptions.

InsertCreativityHere:
This PR fixes #5475.

There was a bug in how we parsed doc-comment tags with names before their description (ex: `@param <name> <desc>`)
that caused us to report the tag as "unknown" if the description didn't start until the following line.

### DIFF
--- a/cpp/src/Slice/DocCommentParser.cpp
+++ b/cpp/src/Slice/DocCommentParser.cpp
@@ -210,25 +210,19 @@ namespace
         {
             // Then we perform additional parsing to extract the name...
 
-            auto nameStart = line.find_first_not_of(ws, tag.size());
+            auto nameStart = doc.find_first_not_of(ws);
             if (nameStart == string::npos)
             {
-                return false; // Malformed line, ignore it.
+                return false; // Malformed line, missing the name part after the tag. Ignore this line.
             }
 
-            auto nameEnd = line.find_first_of(ws, nameStart);
-            if (nameEnd == string::npos)
-            {
-                return false; // Malformed line, ignore it.
-            }
-            name = line.substr(nameStart, nameEnd - nameStart);
+            // If there's no whitespace after the name, that means the name runs to the end of the line.
+            auto nameEnd = doc.find_first_of(ws, nameStart);
+            name = (nameEnd == string::npos) ? doc.substr(nameStart) : doc.substr(nameStart, nameEnd - nameStart);
 
-            // Store whatever remains of the doc comment in the `doc` string.
-            auto docSplitPos = line.find_first_not_of(ws, nameEnd);
-            if (docSplitPos != string::npos)
-            {
-                doc = line.substr(docSplitPos);
-            }
+            // If there's any non-whitespace characters after the name, store them in the `doc` string.
+            auto docStart = doc.find_first_not_of(ws, nameEnd);
+            doc = (docStart == string::npos) ? "" : doc.substr(docStart);
 
             return true;
         }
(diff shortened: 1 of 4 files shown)


cpp/test/Slice/errorDetection/WarningInvalidComment.ice
```diff
@@ -57,6 +57,20 @@ module Test
         /// @throws SomeOtherException should give a not-thrown-by-this-operation warning.
         /// @throws CommentDummy should give a duplicate-tag warning.
         string stringOp(string name, int value, out byte myOut) throws CommentDummy;
+
+        /// This tests that doc-comment tags with named parameters can start on the line following the tag.
+        /// @param param
+        ///     It should be okay for the description of a parameter to start on the next line.
+        /// @throws CommentDummy
+        ///     It should be okay for the description of an exception to start on the next line.
+        bool fullMultilineCommentTest(int param) throws CommentDummy;
```
InsertCreativityHere, line 66: Without the changes in this PR, this used to generate the following warnings:
```
.../WarningInvalidComment.ice:66: warning: ignoring unknown doc tag '@param' in comment
.../WarningInvalidComment.ice:66: warning: ignoring unknown doc tag '@throws' in comment
```
I also manually checked the generated code for these 2 operations, and it looks like the multiline comment is handled correctly in both cases (at least with `slice2cpp` and `slice2cs`).

cpp/src/Slice/DocCommentParser.cpp
```diff
@@ -210,25 +210,19 @@ namespace
...
             {
-                return false; // Malformed line, ignore it.
+                return false; // Malformed line, missing the name part after the tag. Ignore this line.
             }
 
-            auto nameEnd = line.find_first_of(ws, nameStart);
-            if (nameEnd == string::npos)
-            {
-                return false; // Malformed line, ignore it.
-            }
```
InsertCreativityHere, line 223: The main fix was removing this `nameEnd == string::npos` (i.e. the name takes up the rest of the line).
But I also switched from using `line` to `doc`.

`line` was the entire line, `doc`, is just the part of the comment that comes after then name. Not critical for the fix, but feels cleaner. For example:
- line = `@param foo I like foo!`
- doc = `I like foo!`

pepone (approved):

bernardnormier (approved):
LGTM — good fix and the multi-line tests cover the regression well.

PR #5671, opened 2026-06-22
Fix permanent shutdown hang in ConnectionI::sendResponse (#5452)

### LINKED ISSUE #5452: ConnectionI::sendResponse double-decrements _upcallCount on a non-LocalException — permanent shutdown hang

## Summary

In `ConnectionI::sendResponse`, `--_upcallCount` is the first statement of the inner `try`. Any exception that is **not** an `Ice::LocalException` (e.g. the rethrown `_exception` when it originated from a user exception recorded via `dispatchAll`'s `catch (...)`, or `std::bad_alloc` from `sendMessage`) bypasses the inner `catch (const LocalException&)` and reaches the outer `catch (...)`, where `dispatchException(..., 1)` decrements `_upcallCount` a **second** time. The counter goes negative, the `== 0` notifications never fire, and `waitUntilFinished()` (hence communicator destroy) hangs forever.

## The bug

```cpp
// cpp/src/Ice/ConnectionI.cpp:2441-2513  (sendResponse)
try
{
    if (--_upcallCount == 0) { ... }     // (a) first decrement
    ...
    rethrow_exception(exception);         // may be a non-LocalException
}
catch (const LocalException&)
{
    setState(StateClosed, current_exception());   // (b) may run after an early unlock
}
// outer:
catch (...) { dispatchException(current_exception(), 1); }   // (a) decrements _upcallCount AGAIN
```

There is also a secondary issue (b): on the path where a dispatch completes after the connection reached `StateFinished`, the code unlocks, calls `_removeFromFactory`, then rethrows; the inner catch then runs `setState(StateClosed, …)` with the mutex not held (harmless today because `StateFinished` is terminal, but a formal data race).

## Impact

An AMD dispatch whose response machinery records a non-Ice exception on the connection, followed by another in-flight dispatch completing, drives `_upcallCount` negative; the `== 0` notification never fires and `waitUntilFinished()`/communicator destroy hangs permanently.

## Source references

- **(Ice / sendResponse double-decrement + unlocked setState)** — [`cpp/src/Ice/ConnectionI.cpp#L2441-L2513`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/ConnectionI.cpp#L2441-L2513)

## Recommendation

Ensure `_upcallCount` is decremented exactly once on every exit path — e.g. catch `...` (not only `LocalException`) inside `sendResponse` and route through a single decrement, or use a scope guard. Take `_mutex` before the inner `setState(StateClosed, …)`.

## Severity

**Medium** — permanent shutdown hang (`waitUntilFinished` never returns) on a non-Ice exception during response sending; exotic preconditions.

bernardnormier:
Fixes #5452.

## The bug

`ConnectionI::sendResponse` decremented `_upcallCount` once at the start of its inner `try`, then — when a **non-`LocalException`** propagated out of the inner `catch (const LocalException&)` — decremented it a **second** time in the outer `catch (...) { dispatchException(current_exception(), 1); }`. The counter went negative, the `_upcallCount == 0` notification never fired, and `waitUntilFinished()` (hence communicator destruction) hung **forever**.

A non-`LocalException` can reach that path via the rethrown connection `_exception` when it was set from a non-Ice dispatch exception, or `std::bad_alloc` from `sendMessage`.

## The fix

Rework `sendResponse` so `_upcallCount` is decremented **exactly once**:

- Hold `_mutex` for the whole function — `lock` now outlives the `try`, so the single `catch` runs with the mutex held (the previous nesting existed only to keep the lock for the inner `setState`).
- When the connection is already closed (`_state >= StateClosed`), **`return`** instead of rethrowing the stored exception — the response can't be sent and there's nothing else to do. This also removes the audit's secondary issue: the inner `setState(StateClosed, …)` could previously run after `lock.unlock()` on the connection-finished path.
- Handle a send failure with a single `catch (...) { setState(StateClosed, current_exception()); }`. `catch (...)` (rather than `LocalException`) keeps `sendResponse` non-throwing, matching the old behaviour and the neighbouring catch-all handlers.

`dispatchException` is no longer called from `sendResponse` (still used by the fatal-dispatch path).

The reworked method now lines up with the C#/Java/JS mappings, which each decrement once and never had this bug (verified — the double-decrement was C++-only; Swift and the script mappings route responses back through the C++ runtime and inherit the fix).

## Cleanup

Removed the confusingly named one-line `ConnectionI::exception()` helper (used in a single place) by inlining it into `startAsync`.

## Testing

Builds clean. `Ice/operations`, `Ice/ami`, `Ice/exceptions`, and `Ice/hold` pass (twoway/oneway/AMI/batch/collocated/AMD, plus exception and connection hold/close paths).

No CHANGELOG entry (exotic preconditions; internal robustness).

### DIFF
--- a/cpp/src/Ice/ConnectionI.cpp
+++ b/cpp/src/Ice/ConnectionI.cpp
@@ -401,7 +401,11 @@ Ice::ConnectionI::startAsync(
     }
     catch (const Ice::LocalException&)
     {
-        exception(current_exception());
+        {
+            std::lock_guard lock(_mutex);
+            setState(StateClosed, current_exception());
+        }
+
         if (connectionStartFailed)
         {
             connectionStartFailed(shared_from_this(), current_exception());
@@ -1850,13 +1854,6 @@ Ice::ConnectionI::setBufferSize(int32_t rcvSize, int32_t sndSize)
     _info = nullptr; // Invalidate the cached connection info
 }
 
-void
-Ice::ConnectionI::exception(std::exception_ptr ex)
-{
-    std::lock_guard lock(_mutex);
-    setState(StateClosed, ex);
-}
-
 Ice::ConnectionI::ConnectionI(
     CommunicatorPtr communicator,
     const InstancePtr& instance,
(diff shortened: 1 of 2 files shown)


cpp/src/Ice/ConnectionI.cpp
```diff
@@ -401,7 +401,11 @@ Ice::ConnectionI::startAsync(
     }
     catch (const Ice::LocalException&)
     {
-        exception(current_exception());
+        {
+            std::lock_guard lock(_mutex);
```
bernardnormier, line 405: Inlined the now-deleted exception(std::exception_ptr) function. This was the only call site.

cpp/src/Ice/ConnectionI.cpp
```diff
@@ -2418,74 +2415,71 @@ Ice::ConnectionI::sendResponse(OutgoingResponse response, uint8_t compress)
...
+    std::unique_lock lock(_mutex);
+    assert(_state > StateNotValidated);
+
     try
     {
-        std::unique_lock lock(_mutex);
-        assert(_state > StateNotValidated);
+        bool finished = false;
 
-        try
```
bernardnormier, line 2427: Removed this nested try block.

cpp/src/Ice/ConnectionI.cpp
```diff
@@ -2418,74 +2415,71 @@ Ice::ConnectionI::sendResponse(OutgoingResponse response, uint8_t compress)
...
-            {
-                // Resume reading if the connection is active and the dispatch count is about to be less than
-                // _maxDispatches.
-                _threadPool->update(shared_from_this(), SocketOperationNone, SocketOperationRead);
-                if (_idleTimeoutTransceiver)
-                {
-                    _idleTimeoutTransceiver->enableIdleCheck();
-                }
-            }
+            return;
```
bernardnormier, line 2452: This missing `return` was another bug,

cpp/src/Ice/ConnectionI.cpp
```diff
@@ -2418,74 +2415,71 @@ Ice::ConnectionI::sendResponse(OutgoingResponse response, uint8_t compress)
...
+
+        if (_state == StateClosing && _upcallCount == 0)
         {
-            setState(StateClosed, current_exception());
+            initiateShutdown();
         }
     }
     catch (...)
     {
-        dispatchException(current_exception(), 1);
```
bernardnormier, line 2488: `dispatchException(current_exception(), 0);` is the same as  lock + `setState(StateClosed, current_exception());`

cpp/src/Ice/ConnectionI.cpp
```diff
@@ -2418,74 +2415,71 @@ Ice::ConnectionI::sendResponse(OutgoingResponse response, uint8_t compress)
...
+        setState(StateClosed, current_exception());
     }
```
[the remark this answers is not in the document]
bernardnormier: _removeFromFactory doesn't throw

cpp/src/Ice/ConnectionI.cpp
```diff
@@ -2418,74 +2415,71 @@ Ice::ConnectionI::sendResponse(OutgoingResponse response, uint8_t compress)
 {
     bool isTwoWay = !_endpoint->datagram() && response.current().requestId != 0;
 
-    bool finished = false;
+    // We hold _mutex for the duration of this function, including the catch handler below; hence the lock is not
+    // declared inside the try block.
+    std::unique_lock lock(_mutex);
+    assert(_state > StateNotValidated);
+
```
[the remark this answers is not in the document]
bernardnormier: Very unlikely hang.

pepone (approved):

PR #5674, opened 2026-06-23
Delete copy/move on IceRuby classes that register a GC root

### LINKED ISSUE #5591: Delete copy/move on IceRuby VALUE-wrapper classes that register a GC root
Follow-up to a review comment on #5580.

Several IceRuby C++ wrapper classes register a Ruby GC root tied to the **address** of a `VALUE` member — `rb_gc_register_address(&_member)` in the constructor and `rb_gc_unregister_address(&_member)` in the destructor:

- `RubySliceLoader` — `_sliceLoader` (`ruby/src/IceRuby/RubySliceLoader.cpp`)
- `ValueWriter` — `_object` (`ruby/src/IceRuby/Types.cpp`)
- `ValueReader` — `_object` (`ruby/src/IceRuby/Types.cpp`)
- `ExceptionReader` — `_ex` (`ruby/src/IceRuby/Types.cpp`)

Because the registered root is the member's *address*, these classes are logically **non-copyable and non-movable**. A default copy/move would not re-register the new instance's address, and the moved-from / copied instance's destructor would `rb_gc_unregister_address` an address that was never registered — potentially dropping the only valid root for a still-referenced `VALUE` (use-after-free).

This is **not** an active bug today: each of these is created with `make_shared` and only ever handled through a `shared_ptr`, so no copy or move ever occurs, and the user-declared destructors already suppress the implicit *move* operations. But nothing prevents an accidental copy, and the invariant is fragile — so it's worth making explicit.

### Proposed change

Explicitly `= delete` the copy/move constructors and assignment operators on each of these classes, so the non-copyable/non-movable intent is enforced at compile time. No behavior change — this is hardening only, a good fit for 3.9.0 rather than a 3.8.x patch.

pepone:
`RubySliceLoader`, `ValueWriter` and `ValueReader` register a Ruby GC root tied to the **address** of a `VALUE` member (`rb_gc_register_address(&_member)`), so they are logically non-copyable and non-movable: a default copy/move would not re-register the new instance's address, and the moved-from/copied instance's destructor would `rb_gc_unregister_address` an address that was never registered. All four special members are now explicitly `= delete`d.

`ExceptionReader` must stay copyable — `ice_throw` does `throw *this` and `make_exception_ptr` copies it — and its copy constructor already re-registers a fresh root, so only its move constructor and assignment operators are deleted.

This is not an active bug today (these are only ever created with `make_shared` and handled through a `shared_ptr`), so it's hardening only with no behavior change — a good fit for 3.9.0.

Closes #5591

### DIFF
--- a/ruby/src/IceRuby/Types.h
+++ b/ruby/src/IceRuby/Types.h
@@ -451,6 +451,12 @@ namespace IceRuby
         ValueWriter(VALUE, ValueMap*, const ClassInfoPtr&);
         ~ValueWriter();
 
+        // Registers a GC root tied to the address of _object, so it's neither copyable nor movable.
+        ValueWriter(const ValueWriter&) = delete;
+        ValueWriter(ValueWriter&&) = delete;
+        ValueWriter& operator=(const ValueWriter&) = delete;
+        ValueWriter& operator=(ValueWriter&&) = delete;
+
         void ice_preMarshal() final;
 
         void _iceWrite(Ice::OutputStream*) const final;
@@ -474,6 +480,12 @@ namespace IceRuby
         ValueReader(VALUE, const ClassInfoPtr&);
         ~ValueReader();
 
+        // Registers a GC root tied to the address of _object, so it's neither copyable nor movable.
+        ValueReader(const ValueReader&) = delete;
+        ValueReader(ValueReader&&) = delete;
+        ValueReader& operator=(const ValueReader&) = delete;
+        ValueReader& operator=(ValueReader&&) = delete;
+
         void ice_postUnmarshal() final;
 
         void _iceWrite(Ice::OutputStream*) const final;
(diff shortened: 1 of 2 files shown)


ruby/src/IceRuby/Types.h
```diff
@@ -496,7 +508,12 @@ namespace IceRuby
     {
     public:
         ExceptionReader(const ExceptionInfoPtr&);
+        // Copyable (ice_throw throws *this, and make_exception_ptr copies); the copy constructor re-registers a fresh
+        // GC root for _ex. Registering a root tied to the address of _ex makes it non-movable and non-assignable.
         ExceptionReader(const ExceptionReader&);
```
bernardnormier, line 511: We already have a copy constructor and a destructor, no need to delete the move constructor. I would undo this update to ExceptionReader.

bernardnormier (approved):

ruby/src/IceRuby/Types.h
pepone: 🤖 **Claude:** Done in 7cea509 — reverted the ExceptionReader update. The copy constructor + destructor already suppress the implicit move constructor, so the deletes were redundant.

Issue #5490, opened 2026-06-12
[low-severity, grouped] IceStorm / DataStorm low-severity findings (2 items)


pepone:
> **AI-generated audit finding** — these are low-severity items from an automated **correctness** audit of the Ice C++ source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). They have **not been triaged by a human yet**; please verify each before acting on it. Grouped to reduce noise. Links pinned to commit `e248644d788c7e76e14af8676529201c54c1735d`.

This issue collects **low-severity** correctness findings in IceStorm and DataStorm.

### 1. (IceStorm, admin CLI) `replica` command: missing `return` after error → deref of a disengaged optional
[`cpp/src/IceStorm/Parser.cpp#L246-L251`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceStorm/Parser.cpp#L246-L251) — `error("This topic is not replicated")` does not return; the following `node->nodes()` runs on a `nullopt` node → UB in `icestormadmin` against a non-replicated/transient IceStorm. Add a `return`.

### 2. (DataStorm) `FilterManagerT` mutated without synchronization; `set` silently fails to replace
[`cpp/include/DataStorm/InternalT.h#L518-L577`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/include/DataStorm/InternalT.h#L518-L577) — `set()` uses non-replacing `map::emplace` (a second `setKeyFilter`/`setSampleFilter` with the same name keeps the old factory), and `_factories` has no internal lock while `decode`/`get`/`create` run on Ice dispatch threads. Use insert-or-assign and synchronize the map.

Severity: **Low** for both.

PR #5675, opened 2026-06-23
Fix icestormadmin replica command crash on a non-replicated IceStorm

pepone:
Addresses the **IceStorm** finding in #5490 (item 1).

`Parser::replica` (the `icestormadmin` `replica` command) was missing a `return` after its "This topic is not replicated" error:

```cpp
auto node = manager->getReplicaNode();   // std::optional<NodePrx>
if (!node)
{
    error("This topic is not replicated");   // only prints + increments the error count; no return
}
auto nodes = node->nodes();                  // deref of a disengaged std::optional -> UB
```

`getReplicaNode()` returns `nullopt` for a non-replicated IceStorm — including the common non-replicated **persistent** deployment (the `id == -1` path constructs the instance with a default `nullopt` node proxy), not just the transient case — so this fired in normal use, crashing `icestormadmin`. The fix adds the missing `return`, matching the sibling error path a few lines above.

`icestormadmin` is C++-only, so no other language mapping needs the same change.

The DataStorm finding (item 2) in #5490 is still under review and will be handled separately, so this PR does not close the issue.

### DIFF
--- a/changelog.d/service-icestorm/icestormadmin-replica-crash.md
+++ b/changelog.d/service-icestorm/icestormadmin-replica-crash.md
@@ -1,0 +1,2 @@
+- Fixed a crash in `icestormadmin`: running the `replica` command against a non-replicated IceStorm instance
+  dereferenced an empty node proxy. The command now reports that the instance is not replicated instead of crashing.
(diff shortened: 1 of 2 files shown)


changelog.d/service-icestorm/icestormadmin-replica-crash.md
```diff
@@ -0,0 +1,2 @@
+- Fixed a crash in `icestormadmin`: running the `replica` command against a non-replicated IceStorm instance
+  dereferenced an empty node proxy. The command now reports that the instance is not replicated instead of crashing.
```
bernardnormier: The changelog generally should not provide insights into implementation details (e.g. deferenced a null pointer and crashes).  Keep it user-focused.

bernardnormier (approved):

changelog.d/service-icestorm/icestormadmin-replica-crash.md
pepone: 🤖 **Claude:** Done in bddeb20 — reworded the entry to drop the implementation detail and keep it user-focused.

PR #5678, opened 2026-06-23
Fix WebSocket ping/pong control-frame masking in C++, C#, and Java (RFC 6455 §5.5.3)

### LINKED ISSUE #5447: WebSocket: ping payload not unmasked on receipt and pong payload not masked on send (RFC 6455 §5.5.3)

## Summary

The WebSocket transceiver violates RFC 6455 §5.5.3 in both directions for control frames. On receipt, a masked client **ping** payload is copied raw (the unmask loop only runs for `ReadStatePayload`), so the echoed pong differs from the ping. On send, `prepareWriteHeader(OP_PONG, …)` advertises `FLAG_MASKED` with a random key but copies the payload **without** applying the mask. A conforming peer therefore receives a corrupted pong.

## The bug

```cpp
// cpp/src/Ice/WSTransceiver.cpp:1429-1434  (ping payload memcpy'd raw, never unmasked)
// cpp/src/Ice/WSTransceiver.cpp:1557-1566  (pong advertises FLAG_MASKED but payload not XORed)
```

The adjacent OP_CLOSE path (≈ lines 1583-1584) *does* XOR its payload, confirming the omission is a bug rather than a design choice. Ice↔Ice is unaffected only because Ice never initiates pings (`StatePingPending` is never assigned) and ignores pong payloads.

## Impact

Any conforming non-Ice peer or WS-aware intermediary that sends payload-bearing pings and validates the returned pong payload (a common liveness probe) sees a mismatch and may drop the connection.

## Source references

- **(Ice / ping payload not unmasked on receipt)** — [`cpp/src/Ice/WSTransceiver.cpp#L1429-L1434`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/WSTransceiver.cpp#L1429-L1434)
- **(Ice / pong payload not masked on send)** — [`cpp/src/Ice/WSTransceiver.cpp#L1557-L1566`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/WSTransceiver.cpp#L1557-L1566)

## Recommendation

Unmask the ping payload on receipt before storing it as the pong payload, and apply the mask to the pong payload on send (as the OP_CLOSE path already does) when `FLAG_MASKED` is set.

## Severity

**Medium** — RFC 6455 control-frame violation breaks interop with conforming WebSocket peers/intermediaries that verify pong payloads.

pepone:
The C++, C#, and Java WebSocket transceivers mishandled the payload of control frames in both directions:

- **On receipt**, a masked client **ping** payload was copied into `_pingPayload` raw — the unmask loop only ran for data frames — so the echoed **pong** carried the still-masked bytes instead of the ping payload.
- **On send**, the `OP_PONG` path copied `_pingPayload` without applying `_writeMask`, even though `prepareWriteHeader` had advertised `FLAG_MASKED` with a random key, so a conforming peer received a corrupted pong.

Each transceiver now unmasks the ping payload on receipt (when the connection is incoming/server, like the data path) and masks the pong payload on send (when outgoing/client, like the `OP_DATA`/`OP_CLOSE` paths). Ice↔Ice was unaffected because Ice never initiates pings and ignores pong payloads, but a conforming non-Ice peer that sends payload-bearing pings and validates the returned pong (a common liveness probe) saw a mismatch and could drop the connection.

The C# (`WSTransceiver.cs`) and Java (`WSTransceiver.java`) transceivers are line-for-line ports of the C++ one and carried the identical bug. JavaScript wraps the host's native `WebSocket` (no hand-rolled framing) and is not affected; Python/Ruby/PHP/Swift/MATLAB use the C++ core.

### Test

Extends the existing `wsAllowedOrigins` ping/pong control-frame test (the raw-WS probe from #5593) with a payload-bearing ping that asserts the pong echoes the payload verbatim. Because `scripts/tests/Ice/wsAllowedOrigins.py` is mapping-agnostic and C++, C#, and Java all have that test server, the one test covers all three mappings. It fails without the receive-side fix (the server echoes the masked bytes) and passes with it — verified green for all three.

Closes #5447

### DIFF
--- a/changelog.d/cpp/ws-control-frame-masking.md
+++ b/changelog.d/cpp/ws-control-frame-masking.md
@@ -1,0 +1,3 @@
+- Fixed a WebSocket bug where Ice did not unmask the payload of a received ping before echoing it in the pong, and
+  did not mask the pong payload when sending as a client. A conforming WebSocket peer that sends payload-bearing pings
+  and validates the returned pong (a common liveness probe) now receives the correct payload (RFC 6455 §5.5.3).
--- a/changelog.d/cpp/ws-control-frame-masking.md
+++ b/changelog.d/cpp/ws-control-frame-masking.md
@@ -1,3 +1,4 @@
-- Fixed a WebSocket bug where Ice did not unmask the payload of a received ping before echoing it in the pong, and
-  did not mask the pong payload when sending as a client. A conforming WebSocket peer that sends payload-bearing pings
-  and validates the returned pong (a common liveness probe) now receives the correct payload (RFC 6455 §5.5.3).
+- Fixed a WebSocket bug where Ice, acting as a server, echoed a received ping's payload in the pong
+  without unmasking it (and, as a client, sent the pong payload unmasked). A non-Ice WebSocket peer
+  that sends payload-bearing pings and validates the echoed pong could therefore drop the connection
+  (RFC 6455 §5.5.3). The common cases — browser/JS clients and empty keepalive pings — were unaffected.
--- a/changelog.d/csharp/ws-control-frame-masking.md
+++ b/changelog.d/csharp/ws-control-frame-masking.md
@@ -1,0 +1,3 @@
+- Fixed a WebSocket bug where Ice did not unmask the payload of a received ping before echoing it in the pong, and
+  did not mask the pong payload when sending as a client. A conforming WebSocket peer that sends payload-bearing pings
+  and validates the returned pong (a common liveness probe) now receives the correct payload (RFC 6455 §5.5.3).
--- a/changelog.d/csharp/ws-control-frame-masking.md
+++ b/changelog.d/csharp/ws-control-frame-masking.md
@@ -1,3 +1,4 @@
-- Fixed a WebSocket bug where Ice did not unmask the payload of a received ping before echoing it in the pong, and
-  did not mask the pong payload when sending as a client. A conforming WebSocket peer that sends payload-bearing pings
-  and validates the returned pong (a common liveness probe) now receives the correct payload (RFC 6455 §5.5.3).
+- Fixed a WebSocket bug where Ice, acting as a server, echoed a received ping's payload in the pong
+  without unmasking it (and, as a client, sent the pong payload unmasked). A non-Ice WebSocket peer
+  that sends payload-bearing pings and validates the echoed pong could therefore drop the connection
+  (RFC 6455 §5.5.3). The common cases — browser/JS clients and empty keepalive pings — were unaffected.
--- a/changelog.d/java/ws-control-frame-masking.md
+++ b/changelog.d/java/ws-control-frame-masking.md
@@ -1,0 +1,3 @@
+- Fixed a WebSocket bug where Ice did not unmask the payload of a received ping before echoing it in the pong, and
+  did not mask the pong payload when sending as a client. A conforming WebSocket peer that sends payload-bearing pings
+  and validates the returned pong (a common liveness probe) now receives the correct payload (RFC 6455 §5.5.3).
--- a/changelog.d/java/ws-control-frame-masking.md
+++ b/changelog.d/java/ws-control-frame-masking.md
@@ -1,3 +1,4 @@
-- Fixed a WebSocket bug where Ice did not unmask the payload of a received ping before echoing it in the pong, and
-  did not mask the pong payload when sending as a client. A conforming WebSocket peer that sends payload-bearing pings
-  and validates the returned pong (a common liveness probe) now receives the correct payload (RFC 6455 §5.5.3).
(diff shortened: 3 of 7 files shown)


changelog.d/cpp/ws-control-frame-masking.md
```diff
@@ -0,0 +1,3 @@
+- Fixed a WebSocket bug where Ice did not unmask the payload of a received ping before echoing it in the pong, and
+  did not mask the pong payload when sending as a client. A conforming WebSocket peer that sends payload-bearing pings
+  and validates the returned pong (a common liveness probe) now receives the correct payload (RFC 6455 §5.5.3).
```
bernardnormier:
```suggestion
- Fixed a WebSocket bug where an Ice server echoed a received ping's payload in the pong without
  unmasking it, so a non-Ice WebSocket peer that sends payload-bearing pings and validates the echoed
  pong (e.g. an L7 load balancer or gateway health check) could drop the connection (RFC 6455 §5.5.3).
  The common cases — browser/JS clients and empty keepalive pings — were unaffected.
```

Updated from my earlier suggestion: dropped the "and, as a client, sent the pong payload unmasked" clause. The client-side masking fix is correct and worth keeping in the code, but it isn't reachable in a pure-Ice deployment — Ice servers never originate a ping, so an Ice client never receives a payload-bearing ping to echo. The only real-world trigger is the server-side case described above.

changelog.d/csharp/ws-control-frame-masking.md
```diff
@@ -0,0 +1,3 @@
+- Fixed a WebSocket bug where Ice did not unmask the payload of a received ping before echoing it in the pong, and
+  did not mask the pong payload when sending as a client. A conforming WebSocket peer that sends payload-bearing pings
+  and validates the returned pong (a common liveness probe) now receives the correct payload (RFC 6455 §5.5.3).
```
bernardnormier:
```suggestion
- Fixed a WebSocket bug where an Ice server echoed a received ping's payload in the pong without
  unmasking it, so a non-Ice WebSocket peer that sends payload-bearing pings and validates the echoed
  pong (e.g. an L7 load balancer or gateway health check) could drop the connection (RFC 6455 §5.5.3).
  The common cases — browser/JS clients and empty keepalive pings — were unaffected.
```

Updated from my earlier suggestion: dropped the "and, as a client, sent the pong payload unmasked" clause. The client-side masking fix is correct and worth keeping in the code, but it isn't reachable in a pure-Ice deployment — Ice servers never originate a ping, so an Ice client never receives a payload-bearing ping to echo. The only real-world trigger is the server-side case described above.

changelog.d/java/ws-control-frame-masking.md
```diff
@@ -0,0 +1,3 @@
+- Fixed a WebSocket bug where Ice did not unmask the payload of a received ping before echoing it in the pong, and
+  did not mask the pong payload when sending as a client. A conforming WebSocket peer that sends payload-bearing pings
+  and validates the returned pong (a common liveness probe) now receives the correct payload (RFC 6455 §5.5.3).
```
bernardnormier:
```suggestion
- Fixed a WebSocket bug where an Ice server echoed a received ping's payload in the pong without
  unmasking it, so a non-Ice WebSocket peer that sends payload-bearing pings and validates the echoed
  pong (e.g. an L7 load balancer or gateway health check) could drop the connection (RFC 6455 §5.5.3).
  The common cases — browser/JS clients and empty keepalive pings — were unaffected.
```

Updated from my earlier suggestion: dropped the "and, as a client, sent the pong payload unmasked" clause. The client-side masking fix is correct and worth keeping in the code, but it isn't reachable in a pure-Ice deployment — Ice servers never originate a ping, so an Ice client never receives a payload-bearing ping to echo. The only real-world trigger is the server-side case described above.

bernardnormier (approved):
LGTM — verified the C++ side end-to-end (direction logic, the `std::byte`/`uint8_t` casts, and that `prepareWriteHeader` populates `_writeMask` only for `!_incoming`); the masking idiom matches the existing OP_DATA path. The C#/Java diffs are structurally identical ports.

One wording nit on the changelog fragments: the bug is real (an Ice WS/WSS server behind an L7 load balancer or gateway that sends payload-bearing pings and validates the echoed pong could see its connection dropped), but "a common liveness probe" overstates it — the common cases (browser/JS clients, empty keepalive pings) were all unaffected, as the masking is a no-op on an empty payload. Suggested rewrite below (same text for all three fragments) — leads with the observable symptom and drops the "common" claim.

Also fine to ship as-is if you'd rather not churn; the fix itself is correct.

Issue #5486, opened 2026-06-12
[low-severity, grouped] Ice transport-layer low-severity correctness findings (12 items)


pepone:
> **AI-generated audit finding** — these are low-severity items from an automated **correctness** audit of the Ice C++ source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). They have **not been triaged by a human yet**; please verify each before acting on it. Grouped into one issue to reduce noise — split out any you want tracked separately. Links are pinned to commit `e248644d788c7e76e14af8676529201c54c1735d`.

This issue collects the **low-severity** correctness findings in the Ice transport layer.

### 1. SOCKS4 rejection reported as an "HTTP proxy error" (copy-paste)
[`cpp/src/Ice/NetworkProxy.cpp#L140-L146`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/NetworkProxy.cpp#L140-L146) — a SOCKS4 connect rejection raises a diagnostic naming the HTTP proxy. Wrong message only.

### 2. Handshake `switch` on `SSL_get_error` has no `default` — busy-spin
[`cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp#L118-L196`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp#L118-L196) — `SSL_ERROR_WANT_X509_LOOKUP`/`WANT_RETRY_VERIFY`/`WANT_ASYNC` (reachable with a user-supplied `SSL_CTX`) fall through and the outer loop retries immediately, spinning.

### 3. `SSL_get_verify_result` read inside the verify callback (stale/generic diagnostic)
[`cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp#L559-L580`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp#L559-L580) — the real failure reason should come from `X509_STORE_CTX_get_error(ctx)`; as written the reported error is stale/generic. Diagnostics only. *(Verified PARTIAL: the "always X509_V_OK" phrasing is too absolute; the diagnostic-quality defect stands.)*

[...]

PR #5679, opened 2026-06-23
Fix IPv4 loopback range and IPv6 scope-id address comparison (#5486)

pepone:
Addresses #5486 (items 7, 8). One of a set of PRs splitting the grouped transport audit.

- `compareAddress` now compares `sin6_scope_id`, so two IPv6 addresses that differ only by their zone (e.g. `fe80::1%eth0` vs `fe80::1%eth1`) no longer compare equal — this previously allowed connection reuse to the wrong scoped interface.
- IPv4 loopback recognition now covers the whole `127.0.0.0/8` range instead of only `127.0.0.1`, fixing misclassification of hosts such as `127.0.0.2`.

Added deterministic unit assertions in `Ice/info`, verified red→green with each item isolated. Codex-audited.

### DIFF
--- a/cpp/src/Ice/Network.cpp
+++ b/cpp/src/Ice/Network.cpp
@@ -1914,7 +1914,8 @@ IceInternal::isLoopbackOrMulticastAddress(const string& name)
         if (inet_pton(AF_INET, name.c_str(), &addr) > 0)
         {
             // It's an IPv4 address. The whole 127.0.0.0/8 range is loopback, not just 127.0.0.1.
-            return (ntohl(addr.s_addr) & 0xFF000000u) == 0x7F000000u || IN_MULTICAST(ntohl(addr.s_addr));
+            const uint32_t host = ntohl(addr.s_addr);
+            return (host & 0xFF000000u) == 0x7F000000u || IN_MULTICAST(host);
         }
         else if (inet_pton(AF_INET6, name.c_str(), &addr6) > 0)
         {
--- a/cpp/src/Ice/Network.cpp
+++ b/cpp/src/Ice/Network.cpp
@@ -720,6 +720,16 @@ IceInternal::compareAddress(const Address& addr1, const Address& addr2)
         {
             return 1;
         }
+
+        // Distinguish scoped link-local addresses such as fe80::1%eth0 and fe80::1%eth1.
+        if (addr1.saIn6.sin6_scope_id < addr2.saIn6.sin6_scope_id)
+        {
+            return -1;
+        }
+        else if (addr2.saIn6.sin6_scope_id < addr1.saIn6.sin6_scope_id)
+        {
+            return 1;
+        }
     }
 
     return 0;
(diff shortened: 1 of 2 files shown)


cpp/src/Ice/Network.cpp
```diff
@@ -720,6 +720,16 @@ IceInternal::compareAddress(const Address& addr1, const Address& addr2)
         {
             return 1;
         }
+
+        // Distinguish scoped link-local addresses such as fe80::1%eth0 and fe80::1%eth1.
+        if (addr1.saIn6.sin6_scope_id < addr2.saIn6.sin6_scope_id)
```
pepone, line 725: Is it safe to check this for non link-local addresses?

bernardnormier (commented):
Both fixes look correct. Two small things:

- The test comments carry `(issue #5486 item N)` references — we prefer tests to stand on their own without issue-number tags (the substantive explanation is great as-is). Suggestions below to drop them.
- No changelog fragment needed here: binding to a `127.x` address other than `127.0.0.1`, or to a zoned IPv6 link-local, is uncommon enough that this doesn't warrant a user-facing entry.

FYI I checked the other mappings for the same two bugs: C# and Java both classify the full `127.0.0.0/8` as loopback via the standard library (`IPAddress.IsLoopback` / `InetAddress.isLoopbackAddress()`), and their connection-reuse equality compares the IPv6 scope id (`IPAddress.Equals` / `Inet6Address.equals`). JS compares endpoints by host string, so neither bug applies. So this C++ fix is the only one needed (one minor Java `compareAddress` ordering nit aside).

cpp/test/Ice/info/AllTests.cpp
```diff
@@ -55,6 +56,35 @@ void
 allTests(Test::TestHelper* helper)
 {
     Ice::CommunicatorPtr communicator = helper->communicator();
+
+    cout << "testing internal network address classification... " << flush;
+    {
+        // The whole 127.0.0.0/8 range is IPv4 loopback, not just 127.0.0.1 (issue #5486 item 8).
```
bernardnormier:
```suggestion
        // The whole 127.0.0.0/8 range is IPv4 loopback, not just 127.0.0.1.
```

cpp/test/Ice/info/AllTests.cpp
```diff
@@ -55,6 +56,35 @@ void
...
+        test(IceInternal::isLoopbackOrMulticastAddress("127.0.0.1"));
+        test(IceInternal::isLoopbackOrMulticastAddress("127.0.0.2"));
+        test(IceInternal::isLoopbackOrMulticastAddress("127.255.255.254"));
+        test(IceInternal::isLoopbackOrMulticastAddress("239.1.2.3")); // multicast
+        test(!IceInternal::isLoopbackOrMulticastAddress("10.0.0.1"));
+        test(!IceInternal::isLoopbackOrMulticastAddress("126.0.0.1"));
+        test(!IceInternal::isLoopbackOrMulticastAddress("128.0.0.1"));
+
+        // compareAddress must distinguish IPv6 addresses that differ only by their scope id, e.g.
+        // fe80::1%eth0 vs fe80::1%eth1 (issue #5486 item 7).
```
bernardnormier:
```suggestion
        // compareAddress must distinguish IPv6 addresses that differ only by their scope id, e.g.
        // fe80::1%eth0 vs fe80::1%eth1.
```

bernardnormier (approved):
Approving. The C++ fixes are correct, and I confirmed C#/Java handle both cases via the standard library and JS compares by host string, so no parallel fixes are needed. Filed a follow-up issue for the minor Java `compareAddress` scope-id ordering nit. The test-comment suggestions above are optional cleanup.

Issue #5362, opened 2026-05-28
Remove server-side bidirectional support on datagram connections


pepone:
Ice's bidirectional connections feature lets a server invoke proxies in the reverse direction over the originating connection. That model is well-defined for connection-oriented transports (`tcp`, `ssl`, `ws`, `wss`, `bt`, `bts`): each accepted connection corresponds to a kernel-level four-tuple, the peer identity is stable for the connection's lifetime, and the reverse-direction invocation is genuinely "the same connection going back."

For UDP, this doesn't hold. A UDP server has a single bound socket multiplexing every client; there is no per-peer kernel state. Ice represents the whole thing with one `ConnectionI` on the server side whose notion of "the peer" is just the source address of the most recently received datagram. The bidir hook (`Connection::setAdapter` on a datagram connection + `current.con` in the servant) routes its callback through whichever client that single slot currently reflects.

This is acknowledged in our own test:

https://github.com/zeroc-ice/ice/blob/f40f194a8065ad6e8658707b2415e0378148f4fc/cpp/test/Ice/udp/AllTests.cpp#L170-L175

> // This feature is only half-implemented. In particular, we maintain a single Connection object on the server side
> // that gets updated each time we receive a new request.

So the feature ships with a `half-implemented` warning in its own test, and even sequential test scenarios need a retry loop because the "single slot" race surfaces under normal load.

## Proposal

Remove server-side bidirectional support on datagram connections in 3.9.0.

## What goes

- `Ice::Connection::setAdapter(...)` throws when called on a datagram connection (or, equivalently, the dispatch from a server-side UDP connection into a client-side adapter goes away).
- The "udp bi-dir connection" block in `cpp/test/Ice/udp/AllTests.cpp` and the equivalent test code in other mappings.
- Any documentation that claims UDP supports bidirectional invocations.

## What stays

- Outgoing UDP — `ice_datagram()` proxy invokes one-way calls to a server. Unchanged.
- Incoming UDP — server adapter receives one-way calls. Unchanged.
- Client-side UDP callback proxies — a client creating its own UDP object adapter, registering a servant, and sending its proxy to the server so the server can invoke back. That's *not* bidir; it's two independent UDP endpoints. The "send `reply` proxy" pattern at https://github.com/zeroc-ice/ice/blob/f40f194a8065ad6e8658707b2415e0378148f4fc/cpp/test/Ice/udp/AllTests.cpp#L42-L70 stays.

## Cross-language

[...]

PR #5680, opened 2026-06-23
Fix UDP ENOBUFS handling and multicast interface quoting (#5486)

pepone:
Addresses #5486 (items 9, 10). One of a set of PRs splitting the grouped transport audit.

- **Item 9 (C++/Unix):** a transient `ENOBUFS` on a UDP send is routine on macOS/BSD during bursts — a transient local condition, not a connection failure. It was treated as fatal and closed the connection; since UDP is best-effort, `UdpTransceiver::write` now drops the datagram instead of throwing. Limited to non-Windows, because `noBuffers()` also matches `WSAEFAULT` there, which must not be silently dropped.
- **Item 10 (C++, C#, Java):** a multicast `--interface` value was quoted only when it contained `:`, so a Windows adapter friendly name with a space (e.g. `"Ethernet 2"`) was emitted unquoted and failed to re-parse. It is now quoted whenever it contains a separator the parser splits on (whitespace or `:`). The same bug existed in all three mappings; JS has no `UdpEndpointI` (N/A).

Item 10 has a red→green round-trip test in `Ice/proxy` for C++, C#, and Java; `Ice/udp` passes with no regression.

No changelog: with the catastrophic server-adapter case (bidirectional UDP, slated for removal in #5362) out of scope, item 9's remaining impact — a transient `ENOBUFS` burst on UDP sends no longer closing the connection — is minor, consistent with how the #5679 fixes were treated. Codex-audited.

### DIFF
--- a/cpp/src/Ice/UdpTransceiver.cpp
+++ b/cpp/src/Ice/UdpTransceiver.cpp
@@ -175,9 +175,10 @@ repeat:
 
 #ifndef _WIN32
         // ENOBUFS means the socket send buffer is momentarily full; this is routine on macOS/BSD during
-        // bursts. UDP is best-effort, so drop this datagram instead of throwing, which would close the
-        // connection (fatal for the never-recreated shared incoming-datagram connection). Limited to
-        // non-Windows: on Windows noBuffers() also matches WSAEFAULT, which must not be silently dropped.
+        // bursts. UDP is best-effort, so drop this datagram instead of throwing. Throwing would close the
+        // connection, and a server's UDP adapter uses a single datagram socket that Ice never re-creates, so
+        // closing it would permanently stop the adapter from receiving datagrams. Limited to non-Windows: on
+        // Windows noBuffers() also matches WSAEFAULT, which must not be silently dropped.
         if (noBuffers())
         {
             buf.i = buf.b.end();
--- a/cpp/src/Ice/UdpTransceiver.cpp
+++ b/cpp/src/Ice/UdpTransceiver.cpp
@@ -174,11 +174,10 @@ repeat:
         }
 
 #ifndef _WIN32
-        // ENOBUFS means the socket send buffer is momentarily full; this is routine on macOS/BSD during
-        // bursts. UDP is best-effort, so drop this datagram instead of throwing. Throwing would close the
-        // connection, and a server's UDP adapter uses a single datagram socket that Ice never re-creates, so
-        // closing it would permanently stop the adapter from receiving datagrams. Limited to non-Windows: on
-        // Windows noBuffers() also matches WSAEFAULT, which must not be silently dropped.
+        // ENOBUFS means the local send buffer is momentarily full. This is routine on macOS/BSD during bursts
+        // and is a transient local condition rather than a connection failure. UDP is best-effort, so drop this
+        // datagram instead of throwing (throwing would needlessly close the connection). Limited to non-Windows:
+        // there noBuffers() also matches WSAEFAULT, which must not be silently dropped.
         if (noBuffers())
         {
             buf.i = buf.b.end();
--- a/cpp/src/Ice/UdpTransceiver.cpp
+++ b/cpp/src/Ice/UdpTransceiver.cpp
@@ -173,6 +173,18 @@ repeat:
             return SocketOperationWrite;
         }
 
+#ifndef _WIN32
+        // ENOBUFS means the socket send buffer is momentarily full; this is routine on macOS/BSD during
+        // bursts. UDP is best-effort, so drop this datagram instead of throwing, which would close the
+        // connection (fatal for the never-recreated shared incoming-datagram connection). Limited to
+        // non-Windows: on Windows noBuffers() also matches WSAEFAULT, which must not be silently dropped.
+        if (noBuffers())
+        {
+            buf.i = buf.b.end();
+            return SocketOperationNone;
+        }
+#endif
+
         throw SocketException(__FILE__, __LINE__, getSocketErrno());
     }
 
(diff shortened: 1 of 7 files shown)


cpp/src/Ice/UdpTransceiver.cpp
```diff
@@ -173,6 +173,19 @@ IceInternal::UdpTransceiver::write(Buffer& buf)
             return SocketOperationWrite;
         }
 
+#ifndef _WIN32
+        // ENOBUFS means the socket send buffer is momentarily full; this is routine on macOS/BSD during
+        // bursts. UDP is best-effort, so drop this datagram instead of throwing. Throwing would close the
+        // connection, and a server's UDP adapter uses a single datagram socket that Ice never re-creates, so
+        // closing it would permanently stop the adapter from receiving datagrams. Limited to non-Windows: on
+        // Windows noBuffers() also matches WSAEFAULT, which must not be silently dropped.
+        if (noBuffers())
```
pepone, line 181: It is not clear why Windows include WSAEFAULT in the `noBuffers`, seems unrelated to the use we have in StreamSocket for it.,

pepone: `recv` doesn't return `WSAENOBUFS` — only `WSAEFAULT` (send returns both). So in `noBuffers()` (which drives the StreamSocket read/write halve-and-retry), `WSAEFAULT` is the recv-side half of the pair. But retrying it isn't actually required: halving the length can't validate a bad pointer, and we always pass a correctly-sized buffer. Looks defensive/legacy (no rationale in git, predates 2012), and it's harmless only because it degrades to a throw. Unrelated to this UDP send path, which is why the drop stays non-Windows for now.

bernardnormier (commented):
The two fixes are correct. A few requests before approval:

**1. The `ENOBUFS` comment is confusing.** It's in `write()` (sending), but justifies the change via a server's UDP adapter receiving datagrams. The only way a server reaches `write()` is bidirectional UDP (the `sendto(_peerAddr)` reply path) over the socket it also receives on — and that feature is undocumented, half-implemented, and slated for removal (#5362). So the rationale shouldn't lean on it. The best-effort argument stands on its own (and covers the common client-send case too). Suggestion inline.

**2. PR description** — same issue: please drop the "destroying the never-recreated shared incoming-datagram connection" framing (that's the bidir/server-adapter story being removed in #5362) and justify item 9 on the client best-effort send path instead.

**3. No changelog needed.** With the catastrophic server-adapter case going away, item 9's remaining impact (a transient `ENOBUFS` burst on UDP sends no longer closing the connection) is minor, consistent with how we treated the #5679 fixes.

**4. Cross-mapping: item 10 is duplicated in C# and Java** — both quote the multicast interface on `:` only and have the identical round-trip bug for a name with a space:
- `csharp/src/Ice/Internal/UdpEndpointI.cs`: `_mcastInterface.Contains(':', ...)`
- `java/.../com/zeroc/Ice/UdpEndpointI.java`: `_mcastInterface.indexOf(':') != -1`

Please fix both in this same PR (JS has no UdpEndpointI, so it's N/A). Item 9 is C++/Unix-specific and doesn't need a parallel change here.

**5. Test nit** — drop the `(issue #5486 item 10)` tag; tests should stand on their own. Suggestion inline.

cpp/src/Ice/UdpTransceiver.cpp
```diff
@@ -173,6 +173,19 @@ IceInternal::UdpTransceiver::write(Buffer& buf)
             return SocketOperationWrite;
         }
 
+#ifndef _WIN32
+        // ENOBUFS means the socket send buffer is momentarily full; this is routine on macOS/BSD during
+        // bursts. UDP is best-effort, so drop this datagram instead of throwing. Throwing would close the
+        // connection, and a server's UDP adapter uses a single datagram socket that Ice never re-creates, so
+        // closing it would permanently stop the adapter from receiving datagrams. Limited to non-Windows: on
+        // Windows noBuffers() also matches WSAEFAULT, which must not be silently dropped.
```
bernardnormier: Justify on UDP best-effort semantics alone, without the bidir/server-adapter narrative (moot per #5362):

```suggestion
        // ENOBUFS means the local send buffer is momentarily full. This is routine on macOS/BSD during bursts
        // and is a transient local condition rather than a connection failure. UDP is best-effort, so drop this
        // datagram instead of throwing (throwing would needlessly close the connection). Limited to non-Windows:
        // there noBuffers() also matches WSAEFAULT, which must not be silently dropped.
```

cpp/test/Ice/proxy/AllTests.cpp
```diff
@@ -262,6 +262,11 @@ allTests(TestHelper* helper)
     b1 = communicator->stringToProxy(R"(test:udp --sourceAddress "::1" --interface "0:0:0:0:0:0:0:1%lo")");
     test(b1 == communicator->stringToProxy(b1->ice_toString()));
 
+    // A multicast interface containing a space (e.g. a Windows adapter friendly name) must round-trip
+    // through ice_toString(), i.e. it must be quoted (issue #5486 item 10).
```
bernardnormier:
```suggestion
    // A multicast interface containing a space (e.g. a Windows adapter friendly name) must round-trip
    // through ice_toString(), i.e. it must be quoted.
```

bernardnormier (approved):
All review points addressed: the ENOBUFS comment is now justified on best-effort UDP grounds (no bidir/server-adapter narrative), the description is reworked, item 10 is fixed in C# and Java alongside C++ (matching separator sets, with round-trip tests in all three), the issue-number test tag is gone, and the no-changelog rationale is sound. Verified the C#/Java quoting matches the C++  set. LGTM.

PR #5681, opened 2026-06-23
Fix OpenSSL transport handshake correctness and error-path leaks (#5486)

pepone:
Addresses #5486 (items 2, 3, 4, 6). One of a set of PRs splitting the grouped transport audit.

- Add a `default:` to the handshake `SSL_get_error` switch so results such as `SSL_ERROR_WANT_X509_LOOKUP`/`WANT_ASYNC` (reachable with a user-supplied `SSL_CTX`) throw instead of busy-spinning.
- Report the verify failure reason from `X509_STORE_CTX_get_error(ctx)` rather than the stale/generic `SSL_get_verify_result(_ssl)` inside the per-certificate verify callback.
- Reject an empty `IceSSL.CertFile` before dereferencing `&buffer[0]` (undefined behavior on a zero-byte file).
- Free leaked OpenSSL objects on error paths: the `BIO` when the context selection callback returns null, the popped extra-chain `X509` when `SSL_CTX_add_extra_chain_cert` fails, and the `X509` in `decodeCertificate` when `X509_check_purpose` fails.

The OpenSSL backend is not built on macOS; verified by review + codex (OpenSSL ownership semantics checked against the docs). CI exercises the Linux/OpenSSL build.

### DIFF
--- a/changelog.d/cpp/openssl-cert-verification-reason.md
+++ b/changelog.d/cpp/openssl-cert-verification-reason.md
@@ -1,0 +1,3 @@
+- Fixed certificate verification error reporting in the OpenSSL-based IceSSL transport. A rejected peer
+  certificate now reports the specific reason (such as "certificate has expired") instead of a generic
+  "rejected by the certificate validation callback" message.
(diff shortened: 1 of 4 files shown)


bernardnormier (approved):
Verified each fix against OpenSSL ownership semantics (popped `c` so no double-free, `X509_STORE_CTX_get_error` is the right in-callback reason source, local `x` leak). All correct. Changelog scoping looks right.

PR #5682, opened 2026-06-23
Fix SSPI buffer leak on Schannel handshake failure (#5486)

pepone:
Addresses #5486 (item 5). One of a set of PRs splitting the grouped transport audit.

A failed Schannel handshake threw before freeing the output buffers Schannel allocated via `ASC_REQ`/`ISC_REQ_ALLOCATE_MEMORY`, leaking the security token and the `SECBUFFER_ALERT` buffer on every failed handshake (the existing token free runs only on the success path). Both the handshake-loop branch and the initial `InitializeSecurityContext` branch now free their allocated buffers with `FreeContextBuffer` before throwing.

Windows-only; verified by review + codex (SSPI buffer-ownership checked against the Microsoft docs). CI exercises the Windows/Schannel build.

### DIFF
--- a/cpp/src/Ice/SSL/SchannelTransceiverI.cpp
+++ b/cpp/src/Ice/SSL/SchannelTransceiverI.cpp
@@ -299,15 +299,13 @@ Schannel::TransceiverI::sslHandshake(SecBuffer* initialBuffer)
             }
             else if (err != SEC_I_CONTINUE_NEEDED && err != SEC_E_OK)
             {
-                SecBuffer* outToken = getSecBufferWithType(outBufferDesc, SECBUFFER_TOKEN);
-                if (outToken && outToken->pvBuffer)
+                if (outBuffers[0].pvBuffer) // token
                 {
-                    FreeContextBuffer(outToken->pvBuffer);
+                    FreeContextBuffer(outBuffers[0].pvBuffer);
                 }
-                SecBuffer* outAlert = getSecBufferWithType(outBufferDesc, SECBUFFER_ALERT);
-                if (outAlert && outAlert->pvBuffer)
+                if (outBuffers[1].pvBuffer) // alert
                 {
-                    FreeContextBuffer(outAlert->pvBuffer);
+                    FreeContextBuffer(outBuffers[1].pvBuffer);
                 }
 
                 ostringstream os;
--- a/cpp/src/Ice/SSL/SchannelTransceiverI.cpp
+++ b/cpp/src/Ice/SSL/SchannelTransceiverI.cpp
@@ -197,8 +197,6 @@ Schannel::TransceiverI::sslHandshake(SecBuffer* initialBuffer)
                 0);
             if (err != SEC_E_OK && err != SEC_I_CONTINUE_NEEDED)
             {
-                // Free the token Schannel allocated via ISC_REQ_ALLOCATE_MEMORY before throwing; the
-                // FreeContextBuffer below this block runs only on the success path.
                 if (outBuffer.pvBuffer)
                 {
                     FreeContextBuffer(outBuffer.pvBuffer);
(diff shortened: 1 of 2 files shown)


cpp/src/Ice/SSL/SchannelTransceiverI.cpp
```diff
@@ -197,6 +197,12 @@ Schannel::TransceiverI::sslHandshake(SecBuffer* initialBuffer)
                 0);
             if (err != SEC_E_OK && err != SEC_I_CONTINUE_NEEDED)
             {
+                // Free the token Schannel allocated via ISC_REQ_ALLOCATE_MEMORY before throwing; the
+                // FreeContextBuffer below this block runs only on the success path.
```
pepone:
```suggestion
```

cpp/src/Ice/SSL/SchannelTransceiverI.cpp
```diff
@@ -295,6 +301,20 @@ Schannel::TransceiverI::sslHandshake(SecBuffer* initialBuffer)
             }
             else if (err != SEC_I_CONTINUE_NEEDED && err != SEC_E_OK)
             {
+                // Free the output buffers Schannel allocated via ASC_REQ/ISC_REQ_ALLOCATE_MEMORY before
+                // throwing; otherwise the security token and the TLS alert buffer leak on every failed
+                // handshake (the token free below this branch is never reached on failure).
+                SecBuffer* outToken = getSecBufferWithType(outBufferDesc, SECBUFFER_TOKEN);
```
pepone:
```suggestion
                SecBuffer* outToken = getSecBufferWithType(outBufferDesc, SECBUFFER_TOKEN);
```

cpp/src/Ice/SSL/SchannelTransceiverI.cpp
```diff
@@ -295,6 +299,17 @@ Schannel::TransceiverI::sslHandshake(SecBuffer* initialBuffer)
...
+                if (outToken && outToken->pvBuffer)
+                {
+                    FreeContextBuffer(outToken->pvBuffer);
+                }
+                SecBuffer* outAlert = getSecBufferWithType(outBufferDesc, SECBUFFER_ALERT);
+                if (outAlert && outAlert->pvBuffer)
+                {
+                    FreeContextBuffer(outAlert->pvBuffer);
+                }
```
bernardnormier:
```suggestion
                if (outBuffers[0].pvBuffer) // token
                {
                    FreeContextBuffer(outBuffers[0].pvBuffer);
                }
                if (outBuffers[1].pvBuffer) // alert
                {
                    FreeContextBuffer(outBuffers[1].pvBuffer);
                }
```

bernardnormier (approved):
Both frees are correct — verified against the success-path idiom (success frees the token at line 311; on a successful handshake Schannel doesn't populate an alert, so the alert slot stays null) and there's no double-free (stack `SecBuffer` descriptors, nothing else frees them on the throw paths). Cross-mapping is N/A: manual SSPI buffer management is Windows/C++-specific (C# uses SslStream, Java uses JSSE).

Approving. Two non-blocking follow-ups:

**1. Simplify the loop-branch cleanup** (suggestion inline). `outBuffers` is a fixed 2-slot array we initialize ourselves (`[0]` = token, `[1]` = alert) and Schannel fills in place, so there's no need to search by type with `getSecBufferWithType` — direct indexing is simpler and matches the first hunk's direct `outBuffer.pvBuffer` free.

**2. Please add a changelog fragment.** The second fix's failure path is shared between incoming and outgoing (the `while` loop does `AcceptSecurityContext`/`InitializeSecurityContext` then a common failure check), so on a server a peer that repeatedly fails handshakes leaks the token + alert buffers per attempt — remotely triggerable. (The first/initial-ISC fix is outgoing-only and minor, so the entry should focus on the server-side leak.) Suggested `changelog.d/cpp/schannel-handshake-leak.md`:

> Fixed a memory leak in the Schannel-based IceSSL transport (Windows) where each failed TLS handshake leaked the security-token and alert buffers allocated by Schannel. On a server, a peer repeatedly failing handshakes could leak memory over time.

PR #5683, opened 2026-06-23
Fix idle-check race aborting a connection that just received data (#5486)

pepone:
Addresses #5486 (item 11). One of a set of PRs splitting the grouped transport audit.

`ConnectionI::idleCheck` aborted the connection whenever the idle timer task fired, even when the IO thread had rescheduled that one-shot task (because fresh bytes arrived) while the fired task was waiting for the connection mutex. `idleCheck` now skips the abort when the idle check task is still scheduled, mirroring the guard already used by `inactivityCheck`. Adds `IdleTimeoutTransceiverDecorator::idleCheckScheduled()`.

The race sub-window isn't deterministically testable; `Ice/idleTimeout` passes with no regression. Codex-audited (no deadlock / lock-order inversion).

### DIFF
--- a/cpp/src/Ice/ConnectionI.cpp
+++ b/cpp/src/Ice/ConnectionI.cpp
@@ -2290,7 +2290,11 @@ void
 Ice::ConnectionI::idleCheck(const chrono::seconds& idleTimeout) noexcept
 {
     std::lock_guard lock(_mutex);
-    if (_state == StateActive && _idleTimeoutTransceiver->idleCheckEnabled())
+    // Skip the abort if the idle check timer task was rescheduled (because we received bytes) while this
+    // fired task was waiting for _mutex. This mirrors the guard in inactivityCheck and closes the race where
+    // Timer::run dequeues the one-shot task before invoking it.
+    if (_state == StateActive && _idleTimeoutTransceiver->idleCheckEnabled() &&
+        !_idleTimeoutTransceiver->idleCheckScheduled())
     {
         setState(
             StateClosed,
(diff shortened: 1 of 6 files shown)


pepone:
🤖 **Claude:** I'm not convinced this race is worth fixing in practice — looking for other opinions before we merge or close.

`idleCheck` only aborts a connection after it receives nothing — not even a heartbeat — for a full idle timeout. A healthy connection sends `ValidateConnection` frames every `idleTimeout / 2`, so the peer's read timer is rescheduled long before it ever nears the boundary. To hit the race the connection has to be silent for ~`idleTimeout` (effectively dead) **and then** receive a byte in the ~millisecond window between the timer task being dequeued and `idleCheck` acquiring the connection mutex. The result is a spurious `ConnectionAbortedException` → reconnect, not a hang or crash.

So it reads as real but practically unreachable. The one counter-argument: `inactivityCheck` already guards the same race class, so leaving `idleCheck` unguarded is an asymmetry future audits will keep re-flagging.

Keep it (C++ + C#/Java) or close as won't-fix?

bernardnormier (commented):
I don't think we should fix it. Make the logic more complicated with no real upside. I would close as won't-fix.

pepone:
🤖 **Claude:** Closing as won't-fix per discussion. The race is real but practically unreachable — heartbeats keep healthy connections far from the idle boundary, so `idleCheck` only fires on an effectively-dead connection (full rationale in #5700). Split item 11 of #5486 out into #5700 and moved it to the *Not planned* milestone to document the design choice.

Issue #5491, opened 2026-06-12
[low-severity, grouped] Slice compiler & code-generator low-severity findings (8 items)


pepone:
> **AI-generated audit finding** — these are low-severity items from an automated **correctness** audit of the Ice C++ source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). They have **not been triaged by a human yet**; please verify each before acting on it. Grouped to reduce noise. Links pinned to commit `e248644d788c7e76e14af8676529201c54c1735d`.

This issue collects **low-severity** findings in the Slice compiler front-end and the code generators.

### 1. (Slice) Signed overflow computing implicit enumerator values at `INT32_MAX`
[`cpp/src/Slice/Parser.cpp#L3991-L3999`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Slice/Parser.cpp#L3991-L3999) — the `_lastValue == INT32_MAX` error is reported but execution continues to `_lastValue + 1` (signed overflow, UB). *(Regraded Low by the verification pass: the input has already been rejected, so this is invalid-input robustness, not output corruption.)* Trigger: `enum E { A = 2147483647, B }`.

### 2. (Slice) Metadata language-prefix match not anchored on `':'`
[`cpp/src/Slice/MetadataValidation.cpp#L283-L297`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Slice/MetadataValidation.cpp#L283-L297) — `directive.find(_language) == 0` treats any directive merely starting with the language name as that language's (e.g. `cstring:` for `cs`), producing a warning aimed at the wrong language. Compare against `_language + ":"`.

### 3. (Slice grammar) Unset mid-rule `$$` read as `$4` in the operation error production
[`cpp/src/Slice/Grammar.y#L947-L964`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Slice/Grammar.y#L947-L964) — the mid-rule action never assigns `$$`, so the throws clause reads an undefined Bison stack slot. Gated by an already-reported syntax error (codegen never sees it), but a dangling semantic-value read. Assign `$$ = $1;`.

### 4. (IceDB) `Env::Env` leaks the LMDB env on the `BadEnvException` path
[...]

PR #5690, opened 2026-06-23
Miscellaneous Fixes to `libSlice`

InsertCreativityHere:
This PR fixes points 1, 2, and 3 of #5491. There are comments explaining each fix inline in the PR.

No CHANGELOG entry since I don't think anyone has ever triggered any of these bugs, nor ever will.

### DIFF
--- a/cpp/src/Slice/Grammar.cpp
+++ b/cpp/src/Slice/Grammar.cpp
@@ -741,17 +741,17 @@ static const yytype_int16 yyrline[] =
      501,   510,   521,   534,   540,   551,   568,   581,   592,   591,
      618,   622,   633,   653,   665,   678,   677,   711,   737,   746,
      749,   757,   766,   769,   773,   781,   809,   843,   858,   864,
-     874,   878,   887,   917,   916,   948,   947,   970,   974,   985,
-     999,   998,  1032,  1042,  1052,  1057,  1067,  1071,  1080,  1089,
-    1092,  1096,  1104,  1114,  1129,  1140,  1157,  1170,  1186,  1185,
-    1210,  1211,  1215,  1224,  1236,  1253,  1259,  1275,  1287,  1300,
-    1308,  1322,  1323,  1329,  1330,  1336,  1340,  1349,  1352,  1360,
-    1361,  1362,  1363,  1364,  1365,  1366,  1367,  1368,  1369,  1374,
-    1378,  1383,  1388,  1398,  1402,  1487,  1493,  1501,  1511,  1526,
-    1535,  1544,  1579,  1586,  1593,  1605,  1619,  1624,  1632,  1644,
-    1649,  1662,  1663,  1664,  1665,  1666,  1667,  1668,  1669,  1670,
-    1671,  1672,  1673,  1674,  1675,  1676,  1677,  1678,  1679,  1680,
-    1681,  1682,  1683,  1684,  1685,  1686,  1687,  1688
+     874,   878,   887,   917,   916,   948,   947,   971,   975,   986,
+    1000,   999,  1033,  1043,  1053,  1058,  1068,  1072,  1081,  1090,
+    1093,  1097,  1105,  1115,  1130,  1141,  1158,  1171,  1187,  1186,
+    1211,  1212,  1216,  1225,  1237,  1254,  1260,  1276,  1288,  1301,
+    1309,  1323,  1324,  1330,  1331,  1337,  1341,  1350,  1353,  1361,
+    1362,  1363,  1364,  1365,  1366,  1367,  1368,  1369,  1370,  1375,
+    1379,  1384,  1389,  1399,  1403,  1488,  1494,  1502,  1512,  1527,
+    1536,  1545,  1580,  1587,  1594,  1606,  1620,  1625,  1633,  1645,
+    1650,  1663,  1664,  1665,  1666,  1667,  1668,  1669,  1670,  1671,
+    1672,  1673,  1674,  1675,  1676,  1677,  1678,  1679,  1680,  1681,
+    1682,  1683,  1684,  1685,  1686,  1687,  1688,  1689
 };
 #endif
 
@@ -2798,13 +2798,14 @@ yyreduce:
     {
         currentUnit->popContainer();
     }
-    yyerrok;
+    yyval = yyvsp[-2];
+    yyerrok; // Signal that the error has been recovered from.
 }
-#line 2804 "src/Slice/Grammar.cpp"
+#line 2805 "src/Slice/Grammar.cpp"
     break;
 
   case 76: /* operation: operation_preamble error ')' @7 throws  */
-#line 956 "src/Slice/Grammar.y"
+#line 957 "src/Slice/Grammar.y"
 {
     auto op = dynamic_pointer_cast<Operation>(yyvsp[-1]);
     auto el = dynamic_pointer_cast<ExceptionListTok>(yyvsp[0]);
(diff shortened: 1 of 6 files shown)


cpp/src/Slice/Grammar.cpp
InsertCreativityHere, line 1: Generated Bison code, don't need to read.

cpp/src/Slice/Grammar.y
```diff
@@ -950,7 +950,8 @@ throws
     {
         currentUnit->popContainer();
     }
-    yyerrok;
+    $$ = $1;
```
InsertCreativityHere, line 953: We were forgetting to set the "return value" of this grammar rule which handles syntax errors in operations. Since it could only be hit by a syntax error, we'd never get any further than the parsing phase. But it's possible this could of caused a null-deref somewhere in the parser? Not clear.

cpp/src/Slice/MetadataValidation.cpp
```diff
@@ -283,18 +283,18 @@ MetadataVisitor::validate(MetadataList metadata, const SyntaxTreeBasePtr& p, boo
     for (auto i = metadata.begin(); i != metadata.end();)
     {
         const string& directive = (*i)->directive();
```
InsertCreativityHere, line 285: Rewrote this check to only keep metadata that either:
1) Doesn't have any ':' characters at all (language agnostic metadata)
2) Starts with a known language prefix, followed by `:`

Previously, we weren't enforcing that the prefix be followed by `:` accidentally letting things like:
`cpps:type` through (since all we checked was that it _starts with_ `cpp`).

cpp/src/Slice/Parser.cpp
```diff
@@ -3993,10 +3993,14 @@ Slice::Enum::createEnumerator(const string& name, optional<int32_t> explicitValu
             ostringstream os;
             os << "value for enumerator '" << name << "' is out of range";
             unit()->error(os.str());
```
InsertCreativityHere, line 3995: Apparently, signed integer overflow is UB in C++. I never knew this and still struggle to believe it.
Like, are there really any implementations out there that don't use 2's complement wrapping??

Anyways, to avoid this UB, now we "manually wrap" in the case where a user has an enumerator value equal to `2,147,483,647`. We already had a test for this, that was working just fine. So I don't think this actually changes any behavior. It's just a more strict adherence to the spec.

cpp/src/Slice/Parser.cpp
```diff
@@ -3993,10 +3993,14 @@ Slice::Enum::createEnumerator(const string& name, optional<int32_t> explicitValu
...
+            nextValue = numeric_limits<int32_t>::min(); // Wrap around to the minimum value.
```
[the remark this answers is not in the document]
InsertCreativityHere: I disagree. I doubt any of these fixes affect literally anyone.
- `enum implicit value overflow handling`: no actual change except for platforms not using 2's complement.
- `metadata validation` only handles pretty bad typos in metadata, which we'd still be emitting errors for anyways
- `grammar semantic-value fix` not clear there is any user-facing effect. But it requires syntax errors to trigger.

cpp/test/Slice/errorDetection/WarningInvalidMetadata.ice
```diff
@@ -26,6 +26,8 @@
 [["bad:unknown"]]
 // Ignore metadata that has a valid language prefix, but that doesn't match the current compiler.
 [["php:unknown"]]
+// Ignore metadata that just happens to start with a valid language prefix, but that isn't an exact match.
+[["cpp_fake:unknown"]]
```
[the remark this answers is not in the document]
InsertCreativityHere: > The new [["cpp_fake:unknown"]] entry will not be treated as language-prefixed metadata by the parser

Yes. That is the entire point of my fix. Previously is was being treated as if a `cpp:` metadata. Now it won't be treated as any metadata for the known languages. This is a good thing.

pepone (approved):

cpp/src/Slice/Parser.cpp
bernardnormier, line 3996: Clamp rather than wrap, and skip the duplicate check on this error path (needs `bool checkValue = true;` declared next to `nextValue`, and `bool checkForDuplicates = checkValue;` at the duplicate-check section below — see top-level comment):
```suggestion
            checkValue = false;     // We already reported an error; skip the duplicate-value check below.
            nextValue = _lastValue; // Clamp rather than overflow; enumerator values must be >= 0.
```

cpp/test/Slice/errorDetection/WarningInvalidMetadata.ice
```diff
@@ -27,6 +27,10 @@
 // Ignore metadata that has a valid language prefix, but that doesn't match the current compiler.
 [["php:unknown"]]
 
+// Ensure metadata that happens to start with a valid language prefix is rejected as unknown.
+// Otherwise, this would be reported as "duplicate metadata" due to the previous 'cpp:header-ext' metadata.
```
bernardnormier, line 31: Heads-up (non-blocking): this case doesn't actually exercise the prefix-anchoring fix. `cpp_fake` isn't a known language, so the metadata parser splits at the first `:` and the directive becomes just `"cpp_fake"` (no colon) — which takes the `colonIndex == npos` -> validate path and reports "unknown metadata" in **both** the old and new code. And I don't see how it could ever be reported as "duplicate metadata" (the unknown-metadata check returns before any uniqueness check), so the second line of the comment is misleading. The test is a fine guard, but I'd drop or correct that rationale.

bernardnormier (commented):
All three fixes are correct in intent — verified against #5491. Two requests.

**Point 1 (enum overflow):** `INT32_MIN` works but has two downsides: it pollutes the enum's `_minValue` via the `< _minValue` branch (a negative value the model assumes can't occur), and for `enum E { A = 2147483647, B, C }` it makes `_lastValue` negative so `C` silently gets `INT32_MIN + 1` with *no* error — only `B` is flagged. Clamping to `_lastValue` (staying at `INT32_MAX`) keeps the value `>= 0`, leaves `_minValue` alone, and flags every out-of-range trailing enumerator. The only reason to reach for `INT32_MIN` is to dodge the secondary "duplicate value" diagnostic — so let's suppress that explicitly with a `checkValue` flag instead. Concretely:

1. Declare it next to `nextValue`:
```cpp
    int32_t nextValue;
    bool checkValue = true;
```
2. In the out-of-range branch, set `checkValue = false;` and clamp `nextValue = _lastValue;` (inline suggestion below).
3. Change `bool checkForDuplicates = true;` (just below, at the duplicate-check section) to `bool checkForDuplicates = checkValue;` so the clamped value doesn't trigger a spurious "same value as" diagnostic.

**Point 2 (metadata test):** non-blocking — see the inline note; the new test doesn't actually exercise the anchoring change and its rationale comment looks incorrect.

No CHANGELOG is right — these are all invalid-input robustness on already-rejected inputs. Holding approval just for the Point 1 tweak.

cpp/src/Slice/Parser.cpp
InsertCreativityHere: **TL;DR:** Claude views negative enumerators as "bad values", but this is actually the entire point.
This is like claiming `string::npos` is bad because you can't have a negative string index.
Negative enumerator values are sentinels, which lets downstream code tell "hey this is a 'placeholder' for error recovery purposes". Also the changes cause more verbose error output, compared to the current logic which emits exactly one "out of range" error, which I think is better.

----

**Not TL;DR:**

I don't think "model correctness" is fixable here. In general, I've tried, and there's just no way to keep the AST 100% correct in the face of arbitrary syntax errors. Here, we'd just be trading off:
- With my code we would have negative enumerator values, which violates the model
- With the suggested changes we'd have multiple enumerators equal to `int32::max`. also violating the model.

In practice, none of this really matters, since we die after the parsing phase, and during parsing, we're careful to not make many assumptions about "model correctness".

Anyways, downsides to this change:

1) **No way to detect bogus values downstream:** with negative values, code can detect "hey this is a bad value, let me skip validation". Remember, we allow using enumerator values for all kinds of purposes in Slice!

2) **Noisier error output:** With the current logic, we only emit a single error that your enumerator's implicit values are overflowing. With these changes, we would emit `N` many errors, one for each enumerator including and after the overflowing one. We only emit an error when `_lastValue == numeric_limits<int32_t>::max()`. With the overflow logic, we wrap to -2billion, and no longer hit that branch. If we stay at `int32::max` instead of wrapping, we'll keep hitting that error-reporting branch over and over. You could argue this doesn't matter because no one is getting this error... but I think issuing 1 error is better than multiple all stating the same thing.

cpp/test/Slice/errorDetection/WarningInvalidMetadata.ice
InsertCreativityHere: Yes, the bug outlined in #5491 wasn't actually a bug. My fix to the code was purely a correctness sanity check.

We already check for valid language prefixes before we hit the code in question here, so there's no way for something like `cpp_fake` to make it through. I just thought to add this test case to guard against future possible regressions. And if this regression _did_ ever occur, I think it would then be flagged as a duplicate.

If you don't like the comment, fine, I can remove it. But I added it make sure we don't blindly swap the error message in the case this regression does occur in the future. This _should_ be an `unknown` error, not a `duplicate`.

InsertCreativityHere:
It doesn't actually matter, since `_minValue` was only used by the code-gen, which we never reach.
But I added an `if` statement. We only update `_minValue` for valid enumerator values now.

bernardnormier (approved):

PR #5691, opened 2026-06-24
Treat ENOBUFS as a connection error on the OpenSSL transport

### LINKED ISSUE #5444: OpenSSL: halving packetSize on ENOBUFS violates SSL_write retry contract — transient buffer pressure drops the connection

## Summary

On `ENOBUFS` (`SSL_ERROR_SYSCALL`), the OpenSSL transceiver halves `packetSize` and retries `SSL_write`. But at that point the record is already encrypted and pending inside OpenSSL, so retrying with a *smaller* length fails OpenSSL's same-arguments retry contract (`ssl3_write_bytes`'s `len >= wnum + wpend_tot` check → `SSL_R_BAD_LENGTH`), which the code then turns into a fatal `ProtocolException`. The halving heuristic was copied from the raw TCP transceiver, where it is valid.

## The bug

```cpp
// cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp:340-344
if (IceInternal::noBuffers() && packetSize > 1024)
{
    packetSize /= 2;
    continue;   // re-enters SSL_write with a smaller length than the pending record
}
```

Unlike a plain socket write, `SSL_write` must be retried with the *same* buffer and length after a would-block; `SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER` is not set.

## Impact

On Linux, when `send()` returns `ENOBUFS` while ≥ ~17 KB remains pending in OpenSSL, a transient buffer-pressure condition is converted into a fatal `ProtocolException` that tears down the connection.

## Source references

- **(Ice / ENOBUFS halves packetSize before SSL_write retry)** — [`cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp#L333-L344`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp#L333-L344)

## Recommendation

Do not change `packetSize` between `SSL_write` retries. On `noBuffers()`, treat it as a would-block (return `SocketOperationWrite` and retry the same length later), as the SSL write contract requires, rather than shrinking the request.

## Severity

**Medium** — a transient `ENOBUFS` terminates an otherwise-healthy SSL connection on Linux.

pepone:
Fixes #5444.

## Summary

Removes an incorrect, copied-from-TCP retry heuristic from the OpenSSL transceiver. On a `send()`/`recv()` `ENOBUFS` (surfaced as `SSL_ERROR_SYSCALL`), it halved `packetSize` and retried `SSL_write`/`SSL_read` with a **smaller** length:

```cpp
if (IceInternal::noBuffers() && packetSize > 1024) { packetSize /= 2; continue; }
```

This is invalid for `SSL_write`: a non-completed write must be retried with the **same** length while a record is pending (`SSL_write(3)`: *"it must be repeated with the same arguments … the data and length should still be the same"*). OpenSSL enforces this in `ssl3_write_bytes` (`len < wnum + wpend_tot` → `SSL_R_BAD_LENGTH`), so the smaller retry would turn the condition into a bogus *"SSL protocol error during write."*

The fix removes the halving in both directions and lets `ENOBUFS` fall through to a `SocketException` (a connection error) — matching curl and nginx, which do not retry SSL writes on `ENOBUFS`; none of curl, nginx, Boost.Asio, stunnel, CPython, or Node.js shrink an `SSL_write` length on retry.

## Reachability — this is cleanup, not a hot-path fix

OpenSSL is the Linux SSL backend, and on Linux this condition is effectively never hit:
- `send(2)`: *"ENOBUFS … Normally, this does not occur in Linux. Packets are just silently dropped when a device queue overflows."*
- `recv(2)`: `ENOBUFS` is **not a documented error at all** (a recv resource failure is `ENOMEM`), so the read-side block was strictly dead code.

So this removes incorrect/dead code rather than fixing a frequently-hit bug. Surviving a genuine transient `ENOBUFS` would need a bounded backoff (a naive would-block retry can busy-spin, since `ENOBUFS` isn't cleared by socket writability like `EWOULDBLOCK`); that complexity isn't warranted for a condition this rare, so we err on the simple, contract-correct side.

### DIFF
--- a/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
+++ b/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
@@ -343,12 +343,6 @@ OpenSSL::TransceiverI::write(IceInternal::Buffer& buf)
                         continue;
                     }
 
-                    if (IceInternal::noBuffers() && packetSize > 1024)
-                    {
-                        packetSize /= 2;
-                        continue;
-                    }
-
                     if (IceInternal::wouldBlock())
                     {
                         assert(SSL_want_write(_ssl));
@@ -433,12 +427,6 @@ OpenSSL::TransceiverI::read(IceInternal::Buffer& buf)
                 {
                     if (IceInternal::interrupted())
                     {
-                        continue;
-                    }
-
-                    if (IceInternal::noBuffers() && packetSize > 1024)
-                    {
-                        packetSize /= 2;
                         continue;
                     }
 


bernardnormier (approved):
Correct — the shrink-and-retry was copied from StreamSocket where it's valid for byte-stream TCP, but it violates SSL_write's same-length-retry contract. Both removed blocks now fall through to SocketException, ENOBUFS is effectively unreachable on Linux anyway, and the OpenSSL transceiver is C++-only so there's no cross-mapping. No changelog needed.

externl (approved):

PR #5692, opened 2026-06-24
Fix CFStream selector ready() data race on _ready (iOS)

### LINKED ISSUE #5453: CFStream Selector::ready mutates handler->_ready before acquiring the selector mutex (iOS data race)

## Summary

In the CFStream (iOS) `Selector` implementation, `Selector::ready(EventHandler*, SocketOperation, bool)` modifies `handler->_ready` **before** acquiring `_mutex`. The dedicated run-loop thread reads `handler->_ready` (in `EventHandlerWrapper::ready()`/`checkReady()`) while holding only the selector mutex, so there is a read/write race on `_ready` between thread-pool threads and the run-loop thread. A stale read can miss a ready-state transition (handler never added to `_readyHandlers` → stalled connection) or cause a spurious wake.

## The bug

```cpp
// cpp/src/Ice/Selector.cpp:1121-1140  (CFStream Selector::ready)
if (((handler->_ready & status) != 0) == value) { return; }
if (value)  { handler->_ready = (SocketOperation)(handler->_ready | status); }   // unlocked
else        { handler->_ready = (SocketOperation)(handler->_ready & ~status); }  // unlocked
lock_guard lock(_mutex);   // lock taken only here
...
```

`_ready` is read under `_mutex` by the run-loop thread (`Selector.cpp:926-947,962`), but written here before the lock is taken. The epoll/kqueue builds are safe because everything is serialized by the single thread-pool mutex; only the CFStream variant has its own lock plus a foreign thread.

## Impact

On iOS, a transceiver `ReadyCallback` update racing CFSocket callbacks on the run-loop thread can lose a ready transition (stalled connection) or spuriously wake — a data race on `_ready`.

## Source references

- **(Ice / CFStream ready mutates _ready before locking)** — [`cpp/src/Ice/Selector.cpp#L1121-L1140`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/Selector.cpp#L1121-L1140)

## Recommendation

Move the `handler->_ready` read-modify-write inside `lock_guard lock(_mutex)` so the run-loop thread and pool threads serialize on the same lock.

## Severity

**Medium** — data race on the CFStream selector's `_ready` (iOS only); can stall a connection or spuriously wake.

pepone:
Fixes #5453.

On iOS (CFStream selector), `Selector::ready(EventHandler*, SocketOperation, bool)` modified `handler->_ready` **before** acquiring `_mutex`:

```cpp
if (((handler->_ready & status) != 0) == value) { return; }   // read, unlocked
if (value) handler->_ready |= status; else handler->_ready &= ~status;   // write, unlocked
lock_guard lock(_mutex);   // lock taken only here
...
```

The dedicated run-loop thread reads `handler->_ready` **under** `_mutex` (in `startSelect`/`finishSelect`, via `EventHandlerWrapper::checkReady()`/`readyOp()`). So a `ready()` update from a thread-pool thread races with the run-loop reads: the transition can be lost (the handler is never added to the ready set → stalled connection) or read torn.

Take `_mutex` at the top of `ready()` so the read-modify-write of `_ready` and the run-loop reads serialize on the same lock. The epoll/kqueue and IOCP selectors don't have this race — they have no separate run-loop thread, so all access is serialized by the thread-pool mutex; only the CFStream variant has its own lock plus a foreign thread.

Verified with clang-format and a `-fsyntax-only -DICE_USE_CFSTREAM` compile (CoreFoundation is available on macOS); the iOS CI is the full build/test.

### DIFF
--- a/changelog.d/cpp/cfstream-selector-ready-race.md
+++ b/changelog.d/cpp/cfstream-selector-ready-race.md
@@ -1,0 +1,1 @@
+- Fixed a bug in the iOS (CFStream) transport where a connection could intermittently stall.
--- a/cpp/src/Ice/Selector.cpp
+++ b/cpp/src/Ice/Selector.cpp
@@ -1120,6 +1120,9 @@ Selector::finish(EventHandler* handler, bool closeNow)
 void
 Selector::ready(EventHandler* handler, SocketOperation status, bool value)
 {
+    // Hold _mutex across the read-modify-write of handler->_ready. The run-loop thread reads _ready under _mutex
+    // (in startSelect/finishSelect, via EventHandlerWrapper), so updating it unlocked here would be a data race.
+    lock_guard lock(_mutex);
     if (((handler->_ready & status) != 0) == value)
     {
         return; // Nothing to do if ready state already correctly set.
@@ -1134,7 +1137,6 @@ Selector::ready(EventHandler* handler, SocketOperation status, bool value)
         handler->_ready = static_cast<SocketOperation>(handler->_ready & ~status);
     }
 
-    lock_guard lock(_mutex);
     std::map<EventHandler*, EventHandlerWrapperPtr>::iterator p = _wrappers.find(handler);
     assert(p != _wrappers.end());
     p->second->checkReady();


bernardnormier (approved):
Correct and minimal — moving the lock to the top serializes the read-modify-write of `_ready` with the run-loop reads in `EventHandlerWrapper::checkReady()`. No new re-entrancy (the later `checkReady` portion was already under the lock; the hold is just extended over a comparison and bit ops), and the fix is rightly scoped to CFStream since the epoll/kqueue/IOCP `ready()` has no separate run-loop thread.

One suggestion: consider adding a changelog fragment, since this is a user-visible iOS stall rather than dead code. Something like:

> Fixed a bug in the iOS (CFStream) transport where a connection could intermittently stall.

Up to you whether it clears the bar.

PR #5693, opened 2026-06-24
Drain transient IceStorm topics on shutdown

### LINKED ISSUE #5468: IceStorm: _transientManager is never assigned — transient topics are not drained on shutdown

## Summary

In `IceStorm::Service`, both transient start paths create the `TransientTopicManagerImpl` in a local variable and never assign it to the `_transientManager` member. `stop()`'s `if (_transientManager) { _transientManager->shutdown(); }` is therefore always false, so transient topics and their subscribers are never drained on shutdown — events accepted before shutdown are silently dropped instead of flushed.

## The bug

```cpp
// cpp/src/IceStorm/Service.cpp:107-126 (IceBox transient path) and :349-375 (IceGrid-embedded path)
auto manager = make_shared<TransientTopicManagerImpl>(_instance);   // local only
_managerProxy = topicAdapter->add<TopicManagerPrx>(manager, topicManagerId);
...
// cpp/src/IceStorm/Service.cpp:385-400 (stop)
if (_transientManager) { _transientManager->shutdown(); }   // member never set → always false
```

The persistent path drains correctly via `_manager->shutdown()`; the transient path does not.

## Impact

On a transient IceStorm deployment (`IceStorm.Transient=1` or the IceGrid-embedded variant), `TransientTopicImpl::shutdown()`/`Subscriber::shutdown()` never run, so queued/outstanding events are not flushed before teardown and are silently lost.

## Source references

- **(IceStorm / _transientManager never assigned)** — [`cpp/src/IceStorm/Service.cpp#L107-L126`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceStorm/Service.cpp#L107-L126)
- **(IceStorm / stop checks the unset member)** — [`cpp/src/IceStorm/Service.cpp#L385-L400`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceStorm/Service.cpp#L385-L400)

## Recommendation

Assign the created manager to `_transientManager` on both transient paths so `stop()` drains it.

## Severity

**Medium** — transient IceStorm topics are not drained on shutdown; accepted events are silently dropped.

pepone:
Fixes #5468.

`IceStorm::Service` created the `TransientTopicManagerImpl` in a **local variable** on both transient start paths (the IceBox transient path and the IceGrid-embedded path) and never assigned it to the `_transientManager` member:

```cpp
auto manager = make_shared<TransientTopicManagerImpl>(_instance);   // local only
_managerProxy = topicAdapter->add<TopicManagerPrx>(manager, topicManagerId);
```

So `stop()`'s drain was dead code:

```cpp
if (_transientManager) { _transientManager->shutdown(); }   // member never set → always false
```

`TransientTopicManagerImpl::shutdown()` cascades `topic->shutdown()` → `Subscriber::shutdown()`, so on a transient IceStorm deployment (`IceStorm.Transient=1` or the IceGrid-embedded variant) transient topics and their subscribers were never drained on shutdown — unlike the persistent path, which drains via `_manager->shutdown()`. Outstanding/queued events were dropped instead of flushed.

The fix assigns the created manager to `_transientManager` on both paths.

Verified locally: `IceStorm/single` (which runs the `transient=True` configuration) passes, so the transient start→stop path now exercises the drain without regression.

### DIFF
--- a/changelog.d/service-icestorm/icestorm-transient-drain.md
+++ b/changelog.d/service-icestorm/icestorm-transient-drain.md
@@ -1,0 +1,1 @@
+- Fixed a bug where a transient IceStorm service could drop queued events on shutdown instead of delivering them.
--- a/cpp/src/IceStorm/Service.cpp
+++ b/cpp/src/IceStorm/Service.cpp
@@ -109,8 +109,8 @@ ServiceI::start([[maybe_unused]] const string& serviceName, const CommunicatorPt
         _instance = make_shared<Instance>(instanceName, communicator, publishAdapter, topicAdapter, nullptr);
         try
         {
-            auto manager = make_shared<TransientTopicManagerImpl>(_instance);
-            _managerProxy = topicAdapter->add<TopicManagerPrx>(manager, topicManagerId);
+            _transientManager = make_shared<TransientTopicManagerImpl>(_instance);
+            _managerProxy = topicAdapter->add<TopicManagerPrx>(_transientManager, topicManagerId);
         }
         catch (const Ice::LocalException& ex)
         {
@@ -362,8 +362,8 @@ ServiceI::start(
 
     try
     {
-        auto manager = make_shared<TransientTopicManagerImpl>(_instance);
-        _managerProxy = topicAdapter->add<TopicManagerPrx>(manager, id);
+        _transientManager = make_shared<TransientTopicManagerImpl>(_instance);
+        _managerProxy = topicAdapter->add<TopicManagerPrx>(_transientManager, id);
     }
     catch (const Ice::LocalException& ex)
     {


bernardnormier (approved):
Correct — both transient start paths assigned to a local `manager` instead of the `_transientManager` member, so `stop()`'s `if (_transientManager)` drain was dead code. The fix makes them symmetric with the persistent path. I confirmed the impact is real: `Subscriber::shutdown()` waits for the event queue to drain (`_outstanding == 0 && _events.empty()`), so previously a transient service abandoned in-flight events on shutdown. Cross-mapping N/A (the IceStorm service is C++-only).

One suggestion: consider a changelog fragment, since this is user-visible. Something like:

> Fixed a bug where a transient IceStorm service could drop queued events on shutdown instead of delivering them.

Up to you whether it clears the bar.

externl (approved):

PR #5694, opened 2026-06-24
Simplify IceStorm Subscriber proxy setup

### LINKED ISSUE #5663: Simplify IceStorm Subscriber code
Unrelated to the PR fix, maybe is simper to use `ice_isFixed`  and avoid `std::optional` too:

```
Ice::ObjectPrx newObj = rec.obj->ice_isFixed() ?
    rec.obj->ice_invocationTimeout(instance->sendTimeout()) :
    rec.obj;
```

_Originally posted by @pepone in https://github.com/zeroc-ice/ice/pull/5654#discussion_r3444347574_

pepone:
Fixes #5663.

Cleanup of the subscriber proxy setup in `Subscriber::create`:

- Use `ice_isFixed()` to decide whether to override the invocation timeout, instead of catching `FixedProxyException` for control flow.
- Make `newObj` a plain `Ice::ObjectPrx` instead of `std::optional<Ice::ObjectPrx>` (downstream `newObj->...` calls keep working via `ObjectPrx::operator->`).

```cpp
Ice::ObjectPrx newObj =
    rec.obj->ice_isFixed() ? *rec.obj : rec.obj->ice_invocationTimeout(instance->sendTimeout());
```

No behavior change: a fixed proxy (possible when IceStorm is collocated) is used as-is since its invocation timeout can't be changed, otherwise the timeout is overridden — same as the previous try/catch.

Verified locally: builds, and `IceStorm/single` passes (covers oneway/twoway/batch/datagram/ordered subscribers and the transient configuration).

### DIFF
--- a/cpp/src/IceStorm/Subscriber.cpp
+++ b/cpp/src/IceStorm/Subscriber.cpp
@@ -425,8 +425,7 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
             throw BadQoS("invalid reliability: " + reliability);
         }
 
-        // Override the invocation timeout with the send timeout. ice_invocationTimeout is a local operation that
-        // works on any proxy, including a fixed proxy (which rec.obj can be when IceStorm is collocated).
+        // Override the invocation timeout with the send timeout. 
         Ice::ObjectPrx newObj = rec.obj->ice_invocationTimeout(instance->sendTimeout());
 
         p = rec.theQoS.find("locatorCacheTimeout");
--- a/cpp/src/IceStorm/Subscriber.cpp
+++ b/cpp/src/IceStorm/Subscriber.cpp
@@ -425,7 +425,7 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
             throw BadQoS("invalid reliability: " + reliability);
         }
 
-        // Override the invocation timeout with the send timeout. 
+        // Override the invocation timeout with the send timeout.
         Ice::ObjectPrx newObj = rec.obj->ice_invocationTimeout(instance->sendTimeout());
 
         p = rec.theQoS.find("locatorCacheTimeout");
--- a/cpp/src/IceStorm/Subscriber.cpp
+++ b/cpp/src/IceStorm/Subscriber.cpp
@@ -425,18 +425,10 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
             throw BadQoS("invalid reliability: " + reliability);
         }
 
-        // Override the invocation timeout.
-        optional<Ice::ObjectPrx> newObj;
-        try
-        {
-            newObj = rec.obj->ice_invocationTimeout(instance->sendTimeout());
-        }
-        catch (const Ice::FixedProxyException&)
-        {
-            // In the event IceStorm is collocated this could be a fixed proxy in which case its not possible to
-            // set the timeout.
-            newObj = rec.obj;
-        }
+        // Override the invocation timeout, unless rec.obj is a fixed proxy. When IceStorm is collocated, rec.obj
+        // may be a fixed proxy, whose invocation timeout cannot be changed.
+        Ice::ObjectPrx newObj =
+            rec.obj->ice_isFixed() ? *rec.obj : rec.obj->ice_invocationTimeout(instance->sendTimeout());
 
         p = rec.theQoS.find("locatorCacheTimeout");
         if (p != rec.theQoS.end())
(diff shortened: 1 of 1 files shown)


cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -425,18 +425,10 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
...
-        catch (const Ice::FixedProxyException&)
-        {
-            // In the event IceStorm is collocated this could be a fixed proxy in which case its not possible to
-            // set the timeout.
-            newObj = rec.obj;
-        }
+        // Override the invocation timeout, unless rec.obj is a fixed proxy. When IceStorm is collocated, rec.obj
+        // may be a fixed proxy, whose invocation timeout cannot be changed.
+        Ice::ObjectPrx newObj =
+            rec.obj->ice_isFixed() ? *rec.obj : rec.obj->ice_invocationTimeout(instance->sendTimeout());
```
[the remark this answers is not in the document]
pepone: Copilot is right here, `ice_invocationTimeout` doesn't throw.

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -425,18 +425,10 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
...
-            newObj = rec.obj->ice_invocationTimeout(instance->sendTimeout());
-        }
-        catch (const Ice::FixedProxyException&)
-        {
-            // In the event IceStorm is collocated this could be a fixed proxy in which case its not possible to
-            // set the timeout.
-            newObj = rec.obj;
-        }
+        // Override the invocation timeout, unless rec.obj is a fixed proxy. When IceStorm is collocated, rec.obj
+        // may be a fixed proxy, whose invocation timeout cannot be changed.
```
bernardnormier:
```suggestion
        // A fixed proxy can't be marshaled, so rec.obj is fixed only when subscribe() was collocated and passed
        // one (e.g. a bidirectional-connection proxy). Its invocation timeout can't be changed, so use it as-is;
        // otherwise override the invocation timeout.
```

bernardnormier (approved):
Behavior-identical cleanup — `ice_isFixed()` is exactly the inverse of the `FixedProxyException` path, the `optional`->plain `ObjectPrx` change is safe (`operator->` exists and `newObj` is always assigned), and no new null assumption (the old code already deref'd `rec.obj`). No changelog needed (internal refactor).

One suggestion on the comment: it's terse enough to be confusing, since "fixed proxy" (connection-bound) and "collocated" (no connection) sound contradictory. The key fact is that fixed proxies can't be marshaled (`FixedReference::streamWrite` throws), so a fixed proxy can only reach `rec.obj` when the `subscribe()` call was collocated and passed one. Suggest spelling that out:

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -425,18 +425,9 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
...
-        }
-        catch (const Ice::FixedProxyException&)
-        {
-            // In the event IceStorm is collocated this could be a fixed proxy in which case its not possible to
-            // set the timeout.
-            newObj = rec.obj;
-        }
+        // Override the invocation timeout with the send timeout. ice_invocationTimeout is a local operation that
+        // works on any proxy, including a fixed proxy (which rec.obj can be when IceStorm is collocated).
+        Ice::ObjectPrx newObj = rec.obj->ice_invocationTimeout(instance->sendTimeout());
```
pepone:
```suggestion
        // Override the invocation timeout with the send timeout. 
        Ice::ObjectPrx newObj = rec.obj->ice_invocationTimeout(instance->sendTimeout());
```

pepone:
Reworked the PR because ice_invocationTimeout doesn't throw for fixed proxies, this catch for `FixedProxyException` survived the `ice_timeout -> ice_invocationTimeout` updates but was effectively dead code.

bernardnormier (approved):

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -425,18 +425,8 @@ Subscriber::create(const shared_ptr<Instance>& instance, const SubscriberRecord&
...
-            newObj = rec.obj->ice_invocationTimeout(instance->sendTimeout());
-        }
-        catch (const Ice::FixedProxyException&)
-        {
-            // In the event IceStorm is collocated this could be a fixed proxy in which case its not possible to
-            // set the timeout.
-            newObj = rec.obj;
-        }
+        // Override the invocation timeout with the send timeout. 
+        Ice::ObjectPrx newObj = rec.obj->ice_invocationTimeout(instance->sendTimeout());
```
pepone:
```suggestion
        // Override the invocation timeout with the send timeout.
        Ice::ObjectPrx newObj = rec.obj->ice_invocationTimeout(instance->sendTimeout());
```

PR #5695, opened 2026-06-24
Fix Java IceLocatorDiscovery request never completing on same-proxy retry

### LINKED ISSUE #5497: [Audit-Medium] C# IceLocatorDiscovery: same-proxy retry throws into a discarded task — request's TaskCompletionSource never completes, caller hangs

## Summary

In C# IceLocatorDiscovery, when an invocation on a discovered locator fails with an exception that has no specific mapping (e.g. `ConnectTimeoutException`), the generic `catch` in `Request.performInvokeAsync` re-enters `LocatorI.invoke` to retry. If rediscovery resolves to a proxy **equal to the one that just failed**, `Request.invoke` takes the `else` branch and `throw`s `_exception` — but that throw happens inside the `catch` block of the **discarded** fire-and-forget `performInvokeAsync` task (`_ = performInvokeAsync(l)`), so nothing routes it to `SetException`. The request's `TaskCompletionSource` never completes and the `Task` returned by `ice_invokeAsync` hangs.

## The bug

```csharp
// csharp/src/IceLocatorDiscovery/PluginI.cs:41-93
public void invoke(Ice.LocatorPrx l)
{
    if (_locatorPrx == null || !_locatorPrx.Equals(l))
    {
        _locatorPrx = l;
        _ = performInvokeAsync(l);
    }
    else
    {
        Debug.Assert(_exception != null);
        throw _exception;                       // (A) thrown inside the discarded task's catch
    }

    async Task performInvokeAsync(Ice.LocatorPrx locator)
    {
        try { ... SetResult(result); }
        catch (Ice.RequestFailedException exc) { SetException(exc); }
        ...
        catch (Exception exc)
        {
            _exception = exc;
            _locator.invoke(_locatorPrx, this); // (B) re-enters LocatorI.invoke -> may call invoke(l) again
        }
    }
}
```

When the retry path (B) leads back to `Request.invoke` with a locator equal to `_locatorPrx`, branch (A) throws instead of completing the `TaskCompletionSource`. There is no sibling catch around that call inside `performInvokeAsync`'s `catch (Exception)` body, and the task itself was discarded, so the exception is observed by no one.

## Impact

A client invocation awaiting locator resolution hangs until communicator shutdown instead of receiving the failure. Trigger: a found locator becomes unreachable so `ice_invokeAsync` throws a non-mapped exception, and rediscovery returns the same proxy.

## Source references

- **(same-proxy branch throws / generic retry catch)** — [`csharp/src/IceLocatorDiscovery/PluginI.cs#L41-L93`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/csharp/src/IceLocatorDiscovery/PluginI.cs#L41-L93)
- **(rediscovery queues and re-invokes the request)** — [`PluginI.cs#L306-L325`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/csharp/src/IceLocatorDiscovery/PluginI.cs#L306-L325), [`PluginI.cs#L330-L350`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/csharp/src/IceLocatorDiscovery/PluginI.cs#L330-L350)

## Recommendation

On the same-proxy path, complete the request's `TaskCompletionSource` with `_exception` (`SetException`) instead of throwing into the discarded continuation.

## Severity

**Medium** — a caller of `ice_invokeAsync` can hang indefinitely on a same-proxy retry. Verified **CONFIRMED** by the second-model pass; ranked first in the corrected priority list.

### LINKED ISSUE #5660: Java IceLocatorDiscovery: request never completes on same-proxy retry (counterpart of #5497)
> **AI-generated audit finding** — opened from a correctness analysis of the Java source by Claude (Anthropic), model **Claude Opus 4.8** (`claude-opus-4-8`), while reviewing #5627 / #5497, and independently cross-checked against the source by **OpenAI Codex (GPT-5)**. The cross-check confirmed the core defect and the fix and corrected the failure mechanism (see Notes). **Not yet triaged by a human** — verify the reasoning, reproducibility, and severity before acting.

## Summary

The Java `IceLocatorDiscovery` plug-in has the same root defect that #5627 fixes in C#. When an invocation on a discovered locator fails with an exception that has no specific mapping (e.g. `ConnectTimeoutException`), `Request.exception` re-enters `LocatorI.invoke` to retry. If discovery resolves to a locator proxy **equal to the one that just failed**, `Request.invoke` takes the same-proxy `else` branch — and instead of completing the request, it calls `exception(_exception)`, which re-enters retry again. The request's `CompletableFuture` is **never completed**, so the `CompletionStage` returned by `ice_invokeAsync` never resolves.

## The bug

`java/src/com.zeroc.icelocatordiscovery/src/main/java/com/zeroc/IceLocatorDiscovery/PluginI.java`

```java
// Request.invoke — lines 62-83
void invoke(LocatorPrx l) {
    if (_locatorPrx == null || !_locatorPrx.equals(l)) {   // structural (value) equality
        _locatorPrx = l;
        ... // async invoke; on failure -> exception(...)
    } else {
        assert (_exception != null);
        exception(_exception);   // (!) re-enters retry instead of completing _future
    }
}

// Request.exception — lines 85-104
private void exception(LocalException ex) {
    try { throw ex; }
    catch (RequestFailedException exc)            { _future.completeExceptionally(ex); }
    catch (UnknownException exc)                  { _future.completeExceptionally(ex); }
    catch (NoEndpointException exc)               { _future.completeExceptionally(new ObjectNotExistException()); }
    catch (ObjectAdapterDeactivatedException exc) { _future.completeExceptionally(new ObjectNotExistException()); }
    catch (ObjectAdapterDestroyedException exc)   { _future.completeExceptionally(new ObjectNotExistException()); }
    catch (CommunicatorDestroyedException exc)    { _future.completeExceptionally(new ObjectNotExistException()); }
    catch (LocalException exc) {                   // _exception is ALWAYS caught here
        _exception = exc;
        _locator.invoke(_locatorPrx, Request.this); // retry
    }
}
```

`_exception` is assigned in exactly one place — that final catch-all (line 101). So it always holds a default-category exception, and feeding it back through `exception()` re-hits the catch-all → `_locator.invoke(...)` (retry), never `_future.completeExceptionally(...)`. The request is never completed.

## Why the retry never terminates

`LocatorI.invoke` compares the locator by **reference identity** (`_locator != locator`, line 325), whereas `Request.invoke` compares by **value** (`.equals`, line 63). After a successful `foundLocator`, `LocatorI._locator` holds a freshly-decoded proxy that is `.equals()` to — but not the same object as — the request's cached `_locatorPrx`. The two checks then disagree:

- `LocatorI.invoke` sees `_locator != _locatorPrx` (true) → line 326 calls `request.invoke(_locator)` **synchronously**;
- `Request.invoke` sees `_locatorPrx.equals(_locator)` (true) → same-proxy `else` → `exception(_exception)` → `_locator.invoke(...)` → ...

a synchronous cycle with no completion and no async yield, which grows the stack until **`StackOverflowError`**. The `_nextRetry` throttle (checked at line 327) can't break it: `_nextRetry` is assigned only in `_retryTask` on lookup **timeout** (line 485); a locator that keeps answering discovery cancels that timer in `foundLocator` (lines 267-273) and never advances `_nextRetry`.

(When `LocatorI._locator` and `Request._locatorPrx` are the *same* object reference, the flow instead falls through to a fresh multicast lookup and the request still never completes on the same-proxy path until discovery eventually times out and flushes pending requests through the void locator.)

## Expected vs actual

- **Expected:** like C++ (`_exceptionCallback(_exception)` in `cpp/src/IceLocatorDiscovery/PluginI.cpp`) and the C# fix in #5627, the same-proxy branch should complete the request with the last failure.
- **Actual:** it re-enters retry; the future returned by `ice_invokeAsync` never completes — worst case `StackOverflowError`.

## Suggested fix

Complete the request directly in the same-proxy branch, mirroring C++/C#:

```java
} else {
    assert (_exception != null);
    _future.completeExceptionally(_exception); // complete directly; matches C++ / C# (#5627)
}
```

`_exception` is the original unmapped failure, so completing with it loses no mapping.

## Notes

- Same family as #5497 (C#), fixed by #5627. The mappings differ in *how* they break: C# silently dropped the exception (`throw` into a discarded fire-and-forget `async` task); Java re-enters retry.
- The codex cross-check **confirmed** the never-completing-future defect and the fix, and **corrected** an initial "multicast-paced livelock" description to the synchronous-recursion / `StackOverflowError` mechanism above.
- No automated test proposed: there is no Java `IceLocatorDiscovery` test harness, and a deterministic repro of the multicast-rediscovery race is flaky.

pepone:
When a locator invocation fails with an unmapped exception and locator rediscovery resolves to a proxy equal to the one that just failed, `Request.invoke` (same-proxy `else` branch) called `exception(_exception)`, which re-entered the retry path instead of completing the request. As a result the `CompletableFuture` returned to the caller never completed (worst case, a `StackOverflowError` from the synchronous recursion).

This completes the future with the last failure in that branch, matching C++ and the C# plug-in.

Java counterpart of the C# fix #5497 / #5627.

Closes #5660

### DIFF
--- a/java/src/com.zeroc.icelocatordiscovery/src/main/java/com/zeroc/IceLocatorDiscovery/PluginI.java
+++ b/java/src/com.zeroc.icelocatordiscovery/src/main/java/com/zeroc/IceLocatorDiscovery/PluginI.java
@@ -77,8 +77,10 @@ class PluginI implements Plugin {
                     exception(ex);
                 }
             } else {
-                assert (_exception != null); // Don't retry if the proxy didn't change
-                exception(_exception);
+                // The proxy didn't change, so retrying would loop. Complete the request with the last failure,
+                // like C++ and the C# plug-in.
+                assert (_exception != null);
+                _future.completeExceptionally(_exception);
             }
         }
 
--- a/java/src/com.zeroc.icelocatordiscovery/src/main/java/com/zeroc/IceLocatorDiscovery/PluginI.java
+++ b/java/src/com.zeroc.icelocatordiscovery/src/main/java/com/zeroc/IceLocatorDiscovery/PluginI.java
@@ -77,8 +77,7 @@ class PluginI implements Plugin {
                     exception(ex);
                 }
             } else {
-                // The proxy didn't change, so retrying would loop. Complete the request with the last failure,
-                // like C++ and the C# plug-in.
+                // The proxy didn't change, so retrying wont help.
                 assert (_exception != null);
                 _future.completeExceptionally(_exception);
             }
--- a/changelog.d/java/5660.md
+++ b/changelog.d/java/5660.md
@@ -1,0 +1,2 @@
+- Fixed a bug in Java IceLocatorDiscovery where a request could fail to complete (worst case, overflow the
+  stack) when a failed locator invocation was retried and locator rediscovery returned the same locator proxy.


java/src/com.zeroc.icelocatordiscovery/src/main/java/com/zeroc/IceLocatorDiscovery/PluginI.java
```diff
@@ -77,8 +77,10 @@ void invoke(LocatorPrx l) {
                     exception(ex);
                 }
             } else {
-                assert (_exception != null); // Don't retry if the proxy didn't change
-                exception(_exception);
+                // The proxy didn't change, so retrying would loop. Complete the request with the last failure,
+                // like C++ and the C# plug-in.
+                assert (_exception != null);
+                _future.completeExceptionally(_exception);
```
pepone:
```suggestion
                // The proxy didn't change, so retrying wont help.
                assert (_exception != null);
                _future.completeExceptionally(_exception);
```

bernardnormier (approved):
Correct, and faithful to the C# plug-in: C# splits map-and-retry (`handleException`) from completion (`SetException`), and its same-proxy branch calls `SetException(_exception)`. Java conflates both into `exception()`, so the old `exception(_exception)` re-entered the retry loop (recursing until StackOverflowError); completing `_future` directly is the equivalent of C#'s `SetException`. This aligns Java with C++/C#, completing the cross-mapping. Changelog reads well.

PR #5697, opened 2026-06-24
Fix RetryQueue cancel-vs-running-task race (C++, C#, Java)

### LINKED ISSUE #5488: [low-severity] RetryQueue::cancel erases an executing task — debug-build assertion failure
> **AI-generated audit finding** — low-severity item from an automated **correctness** audit of the Ice C++ source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; please verify before acting on it. Link pinned to commit `e248644d788c7e76e14af8676529201c54c1735d`.

### `RetryQueue::cancel` erases a task that is already executing — debug-build assertion failure
[`cpp/src/Ice/RetryQueue.cpp#L121-L150`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/RetryQueue.cpp#L121-L150)

`cancel()` erases the task from `_requests` unconditionally, then learns from `timer()->cancel()` that the task already fired (returns false). When the still-running `RetryTask::runTimerTask()` later calls `RetryQueue::remove(task)`, the `assert(_requests.find(task) != _requests.end())` fails — aborting debug builds. Release builds self-heal (the `erase` is a no-op). Trigger: cancel an invocation from a user thread at the instant its scheduled retry fires.

**Fix:** only erase from `_requests` when `timer()->cancel()` succeeds.

Severity: **Low** (debug-only abort; release behavior correct).

pepone:
`RetryQueue::cancel` removed the task from the request set *before* asking the timer to cancel it. If a user cancels an invocation at the instant its scheduled retry fires, the task is already running, so the timer cancel returns false — but the task has already been removed.

**C++** (`RetryQueue.cpp`) — the still-running `runTimerTask()` then calls `remove()`, whose `assert(_requests.find(task) != end())` fails, aborting debug builds (release builds self-heal). The premature removal can also let `destroy()`'s wait observe an empty queue while a retry is still in flight, tearing down the client thread pool underneath it. Fixed by only erasing from `_requests` when `timer()->cancel()` succeeds (which guarantees the task will not run, so `remove()` won't be called for it). When the queue is being destroyed every remaining task is running and is removed by `remove()`, so `cancel()` defers to it.

**C#** (`RetryQueue.cs`) — same shape. No assert (its `remove` is guarded), but the same premature-`destroy()` hazard. Same fix: only remove after a successful `timer().cancel()`.

**Java** (`RetryTask.java`) — different mechanism, same hazard: `asyncRequestCanceled` evaluated `_queue.remove(this)` *before* claiming the task via the one-shot `cancel()` flag, so a `run()` that had already claimed the task could be untracked while still in `retry()`. Reordered to `cancel() && _queue.remove(this)` so the losing canceler leaves the running task in the queue for `run()` to remove.

**JavaScript** — single-threaded (run-to-completion); `run()` and `cancel()` cannot interleave, so the race is structurally impossible. No change.

The cross-mapping analysis was independently cross-checked. Verified: C++ compiles, Java compile + checkstyle + OpenRewrite clean, C# builds with 0 warnings/errors.

Closes #5488

### DIFF
--- a/cpp/src/Ice/RetryQueue.cpp
+++ b/cpp/src/Ice/RetryQueue.cpp
@@ -141,8 +141,7 @@ IceInternal::RetryQueue::cancel(const RetryTaskPtr& task)
     // the timer is gone and every remaining task is running, so it is likewise removed by remove().
     if (_instance && _instance->timer()->cancel(task))
     {
-        _requests.erase(task);
-        return true;
+        return _requests.erase(task) != 0;
     }
     return false;
 }
--- a/cpp/src/Ice/RetryQueue.cpp
+++ b/cpp/src/Ice/RetryQueue.cpp
@@ -135,17 +135,14 @@ bool
 IceInternal::RetryQueue::cancel(const RetryTaskPtr& task)
 {
     lock_guard<mutex> lock(_mutex);
-    if (_requests.erase(task) > 0)
+    // Only remove the task if we cancel it in the timer before it runs. If timer()->cancel returns false, the
+    // task is already executing and runTimerTask will call remove() to erase it from _requests; erasing it
+    // here would make that remove() fail its assertion. When the queue is being destroyed (_instance is null)
+    // the timer is gone and every remaining task is running, so it is likewise removed by remove().
+    if (_instance && _instance->timer()->cancel(task))
     {
-        if (_instance)
-        {
-            return _instance->timer()->cancel(task);
-        }
-        else if (_requests.empty())
-        {
-            // If we are destroying the queue, destroy is probably waiting on the queue to be empty.
-            _conditionVariable.notify_one();
-        }
+        _requests.erase(task);
+        return true;
     }
     return false;
 }
--- a/csharp/src/Ice/Internal/RetryQueue.cs
+++ b/csharp/src/Ice/Internal/RetryQueue.cs
@@ -131,8 +131,7 @@ public class RetryQueue
             // removed by remove(), which wakes destroy().
             if (_instance != null && _instance.timer().cancel(task))
             {
-                _requests.Remove(task);
-                return true;
+                return _requests.Remove(task);
             }
             return false;
         }
--- a/csharp/src/Ice/Internal/RetryQueue.cs
+++ b/csharp/src/Ice/Internal/RetryQueue.cs
@@ -124,17 +124,15 @@ public class RetryQueue
     {
         lock (_mutex)
         {
-            if (_requests.Remove(task))
+            // Only remove the task if we cancel it in the timer before it runs. If timer().cancel returns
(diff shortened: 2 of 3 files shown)


cpp/src/Ice/RetryQueue.cpp
```diff
@@ -135,17 +135,14 @@ bool
...
-        {
-            return _instance->timer()->cancel(task);
-        }
-        else if (_requests.empty())
-        {
-            // If we are destroying the queue, destroy is probably waiting on the queue to be empty.
-            _conditionVariable.notify_one();
-        }
+        _requests.erase(task);
+        return true;
```
bernardnormier: Optional: return the erase result rather than an unconditional `true`. Equivalent today (when `timer()->cancel` succeeds the task is guaranteed in `_requests`), but more honest if that invariant ever broke.
```suggestion
        return _requests.erase(task) != 0;
```

csharp/src/Ice/Internal/RetryQueue.cs
```diff
@@ -124,17 +124,15 @@ public bool cancel(RetryTask task)
...
-                {
-                    return _instance.timer().cancel(task);
-                }
-                else if (_requests.Count == 0)
-                {
-                    // If we are destroying the queue, destroy is probably waiting on the queue to be empty.
-                    System.Threading.Monitor.Pulse(_mutex);
-                }
+                _requests.Remove(task);
+                return true;
```
bernardnormier: Same as the C++ side — return the actual removal result:
```suggestion
                return _requests.Remove(task);
```

bernardnormier (approved):
Verified all three fixes and they're correct and faithful to each other:

- **C++/C#**: erasing only after a successful `timer()->cancel()` means a task that's already running is left for `runTimerTask`/`remove()` to erase, fixing the assert (C++ debug) and the premature-empty hazard in `destroy()`; during destroy (`_instance` null) `cancel()` short-circuits and defers to `remove()`, which still wakes the waiter, so dropping the old notify/Pulse branch is safe.
- **Java**: reordering to `cancel() && _queue.remove(this)` makes only the claim-winner remove the task, so a losing canceler no longer untracks a task that `run()` is still processing.
- **JS N/A** (run-to-completion) is correct.

No changelog: in release the everyday cancel-vs-fire race self-heals (the assert is debug-only); the only release crash needs a triple coincidence (cancel landing exactly as a retry fires, *and* during communicator destruction), which is too narrow to be worth an entry.

Approving. Two optional defensive tweaks below — return the actual remove result instead of an unconditional `true`, so the return is honest if the "timer-cancellable ⇔ in the set" invariant ever broke.

PR #5698, opened 2026-06-24
Fix Ice::Service --daemon treating EOF as a readiness signal

### LINKED ISSUE #5456: Service --daemon: parent reports success when the child dies before signaling readiness (EOF treated as ready)

## Summary

In `--daemon` mode, the parent process reads a readiness byte from the child over a pipe but only handles `read() == -1`. A return of `0` (EOF — the daemon crashed/exited before writing the readiness byte) is treated identically to a successful "ready" read: the parent prints nothing and `_exit(EXIT_SUCCESS)`. Init systems and scripts then believe the daemon started even though it is gone.

## The bug

```cpp
// cpp/src/Ice/Service.cpp:1469-1524
if (read(fds[0], &c, 1) == -1) { ... }   // only -1 handled
break;
...
if (c != 0) { /* failure path */ }
_exit(EXIT_SUCCESS);                      // reached on EOF (read == 0) too
```

## Impact

If the grandchild dies during communicator init / `start()` without reaching the error-reporting `write` (killed by a signal, `abort()` in a plugin), the parent exits with success. Supervisors relying on `--daemon`'s exit status conclude the service is running when it is not.

## Source references

- **(Ice / daemon parent treats EOF as readiness)** — [`cpp/src/Ice/Service.cpp#L1469-L1524`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/Service.cpp#L1469-L1524)

## Recommendation

Treat `read() == 0` (EOF before the readiness byte) as a startup failure: report it and `_exit` with a non-zero status, distinguishing it from a successful readiness signal.

## Severity

**Medium** — `--daemon` reports success when the daemon died during startup; wrong exit status misleads supervisors.

pepone:
Fixes #5456.

In `--daemon` mode the parent reads a readiness byte from the daemon over a pipe but only handled `read() == -1`. A return of `0` (EOF — the daemon exited during startup before writing the byte, e.g. killed by a signal or `abort()` in a plug-in, bypassing the C++ exception handlers) left the readiness byte at its initial `0`, so the parent reported success and `_exit(EXIT_SUCCESS)`, misleading service supervisors.

The parent now treats EOF before the readiness byte as a startup failure: it reports `daemon exited during startup before signaling readiness` and exits with a non-zero status. The normal ready (byte `0`) and failure (byte `1` + message) paths are unchanged — `read()` returns 1 there.

Verified against the source (the byte protocol + the pipe write-end lifecycle across the double fork), an AddressSanitizer standalone repro of the parent read loop (red→green, normal paths unregressed), and a codex cross-check. Not a regression — 3.7 has the identical gap. No in-tree test: `--daemon` double-forks/detaches and the trigger is a mid-startup kill, disproportionate to reproduce reliably in CI.

### DIFF
--- a/changelog.d/cpp/service-daemon-eof-readiness.md
+++ b/changelog.d/cpp/service-daemon-eof-readiness.md
@@ -1,0 +1,3 @@
+- Fixed `Ice::Service` in `--daemon` mode reporting success when the daemon exited during startup before
+  signaling readiness (for example after being killed by a signal or aborting in a plug-in). The parent process
+  now exits with a non-zero status so service supervisors don't conclude the service started.
--- a/changelog.d/cpp/service-daemon-eof-readiness.md
+++ b/changelog.d/cpp/service-daemon-eof-readiness.md
@@ -1,3 +1,0 @@
-- Fixed `Ice::Service` in `--daemon` mode reporting success when the daemon exited during startup before
-  signaling readiness (for example after being killed by a signal or aborting in a plug-in). The parent process
-  now exits with a non-zero status so service supervisors don't conclude the service started.
--- a/cpp/src/Ice/Service.cpp
+++ b/cpp/src/Ice/Service.cpp
@@ -1473,7 +1473,8 @@ Ice::Service::runDaemon(int argc, char* argv[], InitializationData initData)
         char c = 0;
         while (true)
         {
-            if (read(fds[0], &c, 1) == -1)
+            ssize_t rc = read(fds[0], &c, 1);
+            if (rc == -1)
             {
                 if (IceInternal::interrupted())
                 {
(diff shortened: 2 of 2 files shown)


bernardnormier (approved):
The fix is correct: EOF on the readiness pipe (`read() == 0`) previously fell through to `break` with `c == 0`, so the parent reported success and exited 0; the new `rc == 0` branch reports failure and exits non-zero, while the ready/failure (`read() == 1`) paths are unchanged. Verified the EINTR retry and the post-loop `c != 0` message path still hold.

One request: please drop the changelog fragment (`changelog.d/cpp/service-daemon-eof-readiness.md`). `Ice::Service` is public in name only — it's really internal scaffolding for the Ice services (Glacier2/IceGrid/IceBox), so an entry naming it doesn't map to anything a user recognizes, and this mid-startup-crash edge case is low severity. With the fragment removed, this is good to merge.

pepone:
🤖 **Claude:** Dropped the changelog fragment in 9e25bdd — agreed, `Ice::Service` is internal scaffolding and this is a low-severity edge case. The PR is now just the `Service.cpp` fix.

PR #5701, opened 2026-06-24
Fix IceBT transceiver fd leak and data race on aborted outgoing connection

### LINKED ISSUE #5443: IceBT TransceiverI::close() races connect completion — data race, fd/thread/bus-connection leak, debug-build abort

## Summary

`IceBT::TransceiverI::close()` reads `_connection` and closes `_stream` **without taking `_mutex`**, while `connectCompleted` / `connectFailed` set those members under `_mutex` from the BlueZ DBus dispatch/connect threads. Besides the data race, the ordering is unhandled: if `close()` runs first (connect timeout / communicator destroy) and `connectCompleted` arrives afterward, `setFd(fd)` installs a live socket into an already-closed/abandoned stream that nothing ever closes — leaking the fd, the `DBus::ConnectionI` (which owns a private system-bus connection plus a dedicated dispatch thread), and (in debug builds) aborting on the `StreamSocket` destructor assertion `_fd == INVALID_SOCKET`.

## The bug

`close()` takes no lock:

```cpp
// cpp/src/IceBT/TransceiverI.cpp:58-66
void
IceBT::TransceiverI::close()
{
    if (_connection)            // unsynchronized read of _connection
    {
        _connection->close();
    }
    _stream->close();           // unsynchronized close of the stream fd
}
```

while the connect completions mutate the same state under `_mutex`:

```cpp
// cpp/src/IceBT/TransceiverI.cpp:174-198
void IceBT::TransceiverI::connectCompleted(int fd, const ConnectionPtr& conn)
{
    lock_guard lock(_mutex);
    _connection = conn;
    _stream->setFd(fd);          // installs a live fd
    _stream->ready(IceInternal::SocketOperationConnect, true);
}

void IceBT::TransceiverI::connectFailed(std::exception_ptr ex)
{
    lock_guard lock(_mutex);
    _exception = ex;
    _stream->ready(IceInternal::SocketOperationConnect, true);
}
```

`close()` runs on an Ice thread-pool thread during connection-establishment teardown, while `connectCompleted`/`connectFailed` arrive from the BlueZ DBus dispatch / connect threads — so there is a data race on the `_connection` `shared_ptr` and on the stream fd (`setFd` vs `close`). Even with locking added, the late-completion ordering is not handled: a `connectCompleted` arriving after `close()` calls `setFd(fd)` on a stream that was already closed/abandoned. Nothing subsequently closes that fd or the `ConnectionI`, and the `StreamSocket` destructor asserts the fd is already closed:

```cpp
// cpp/src/IceBT/StreamSocket.cpp:25  (~StreamSocket asserts _fd == INVALID_SOCKET)
```

## Impact

An outgoing Bluetooth connection attempt that is aborted (connect timeout or communicator destroy) racing BlueZ `NewConnection` delivery leaks one fd, one dispatch thread, and one private system-bus connection per aborted connect, and aborts debug builds via the `StreamSocket` destructor assertion.

(Related, lower severity: `AcceptorI::close()` does not drain `_transceivers`, so transceivers queued by a late `newConnection` that are never `accept()`ed are destroyed without `close()` — the same assert/fd leak at shutdown.)

## Source references

- **(IceBT / close accesses connection/stream without _mutex)** — [`cpp/src/IceBT/TransceiverI.cpp#L58-L66`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceBT/TransceiverI.cpp#L58-L66)
- **(IceBT / connectCompleted/connectFailed under _mutex)** — [`cpp/src/IceBT/TransceiverI.cpp#L174-L198`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceBT/TransceiverI.cpp#L174-L198)
- **(IceBT / StreamSocket destructor assertion)** — [`cpp/src/IceBT/StreamSocket.cpp#L25`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceBT/StreamSocket.cpp#L25)

## Recommendation

Take `_mutex` in `close()` and record a closed/aborted flag while holding it; have `connectCompleted` check that flag and immediately close the just-delivered fd and the `ConnectionPtr` instead of installing them when the transceiver is already closed. Ensure the `ConnectionI`/fd are released on every abort path so the dispatch thread and bus connection are not leaked.

## Severity

**Medium-High** — data race plus fd/thread/bus-connection leak (and a debug-build abort) on an aborted outgoing Bluetooth connection that races BlueZ connection delivery.

bernardnormier:
Fixes #5443.

`IceBT::TransceiverI::close()` read `_connection` and closed `_stream` without holding `_mutex`, racing `connectCompleted`/`connectFailed`, which mutate those members under the lock from the connection's I/O thread.

- `close()` now takes `_mutex`, flags the transceiver closed, and moves out `_connection`/closes `_stream` under the lock. The connection is closed *outside* the lock, since its `close()` blocks on a D-Bus `DisconnectProfile` call and joins the connection's I/O thread.
- A connection completion that arrives after `close()` no longer installs the fd into the abandoned stream; it closes the socket instead, fixing the descriptor leak and a debug-build assertion failure.

As discussed in the issue, the late-completion path deliberately does not close the D-Bus connection: `connectCompleted` runs on the connection's own I/O thread, where a blocking close would deadlock (and self-join). That leaves a connection/thread leak on this rare race only; fully reclaiming it requires off-thread disposal, which isn't warranted for a patch release.

### DIFF
--- a/cpp/src/IceBT/TransceiverI.cpp
+++ b/cpp/src/IceBT/TransceiverI.cpp
@@ -58,11 +58,26 @@ IceBT::TransceiverI::closing(bool initiator, exception_ptr)
 void
 IceBT::TransceiverI::close()
 {
-    if (_connection)
+    ConnectionPtr connection;
     {
-        _connection->close();
+        lock_guard lock(_mutex);
+        //
+        // Flag the transceiver as closed so a connection completion that races this call
+        // disposes of the late-delivered fd and connection instead of installing them.
+        //
+        _closed = true;
+        connection = std::move(_connection);
+        _stream->close();
     }
-    _stream->close();
+
+    //
+    // Close the connection without holding _mutex: ConnectionI::close() blocks (it makes a
+    // DisconnectProfile D-Bus call and joins the connection's I/O thread).
+    //
+    if (connection)
+    {
+        connection->close();
+    }
 }
 
 IceInternal::SocketOperation
@@ -174,13 +189,26 @@ IceBT::TransceiverI::~TransceiverI() = default;
 void
 IceBT::TransceiverI::connectCompleted(int fd, const ConnectionPtr& conn)
 {
-    lock_guard lock(_mutex);
-    _connection = conn;
-    _stream->setFd(fd);
-    //
-    // Triggers a call to write() from a different thread.
-    //
-    _stream->ready(IceInternal::SocketOperationConnect, true);
+    {
+        lock_guard lock(_mutex);
+        if (!_closed)
+        {
+            _connection = conn;
+            _stream->setFd(fd);
+            //
+            // Triggers a call to write() from a different thread.
+            //
+            _stream->ready(IceInternal::SocketOperationConnect, true);
+            return;
+        }
+    }
+
+    // The transceiver was already closed before the connection completed. We must not install the fd into the abandoned
+    // stream (nothing would ever close it), so close the socket here instead. We deliberately don't close 'conn': we
+    // run on the connection's own I/O thread, where a blocking close would deadlock. This leaks the connection's I/O
+    // thread and system-bus connection, but only on the rare race between an aborted outgoing connection and its
+    // completion.
+    IceInternal::closeSocketNoThrow(fd);
 }
 
 void
--- a/cpp/src/IceBT/TransceiverI.h
+++ b/cpp/src/IceBT/TransceiverI.h
@@ -40,6 +40,7 @@ namespace IceBT
         std::string _addr;
         std::string _uuid;
         bool _needConnect;
+        bool _closed{false};
         std::exception_ptr _exception;
         std::mutex _mutex;
         void connectCompleted(int, const ConnectionPtr&);


cpp/src/IceBT/TransceiverI.cpp
```diff
@@ -174,13 +189,26 @@ IceBT::TransceiverI::~TransceiverI() = default;
...
+            _stream->ready(IceInternal::SocketOperationConnect, true);
+            return;
+        }
+    }
+
+    // The transceiver was already closed before the connection completed. We must not install the fd into the abandoned
+    // stream (nothing would ever close it), so close the socket here instead. We deliberately don't close 'conn': we
+    // run on the connection's own I/O thread, where a blocking close would deadlock. This leaks the connection's I/O
+    // thread and system-bus connection, but only on the rare race between an aborted outgoing connection and its
+    // completion.
```
pepone, line 210: Doesn't leaking the I/O thread means there is a chance that we end up calling std::terminate, we cannot exit the process without joining or detaching all threads. Note this is not a new concern the same would happen before if close was not called because `_connection` was null.

pepone (approved):

externl (approved):

cpp/src/IceBT/TransceiverI.cpp
```diff
@@ -58,11 +58,26 @@ IceBT::TransceiverI::closing(bool initiator, exception_ptr)
...
-        _connection->close();
+        lock_guard lock(_mutex);
+        //
+        // Flag the transceiver as closed so a connection completion that races this call
+        // disposes of the late-delivered fd and connection instead of installing them.
+        //
+        _closed = true;
+        connection = std::move(_connection);
+        _stream->close();
+    }
```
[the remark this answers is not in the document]
bernardnormier: It does not block, see the code.

cpp/src/IceBT/TransceiverI.cpp
bernardnormier: Good point to raise, but I don't think the `std::terminate` risk applies here. `std::terminate` is only triggered by destroying a *joinable* `std::thread` object — a still-running thread doesn't call it and doesn't prevent process exit. And in this path the `DBus::ConnectionI` is never destroyed anyway: it self-references through the dbus filter userdata (`freeConnection`), which is only released by `close()`. Since we skip `close()` here, the object (and its `std::thread`) is never destructed, so `~thread()` never runs.

The net effect is a leaked idle dbus poll thread on this rare abort race — not great, but not critical, and as you note it's pre-existing (the old `_connection == nullptr` path leaked it the same way).

PR #5703, opened 2026-06-25
Reject mid-stream TLS renegotiation in the Schannel transport

### LINKED ISSUE #5449: Schannel: renegotiation requiring network I/O aborts the connection with a mislabeled ConnectionLostException

## Summary

On Windows (Schannel), a renegotiation that cannot complete synchronously is converted into a fatal `ConnectionLostException`. `sslHandshake` legitimately returns `SocketOperationRead`/`SocketOperationWrite` when the resumed handshake needs network I/O (a genuine TLS 1.2 renegotiation always needs a round trip), but the code treats anything other than `SocketOperationNone` as failure — and clears the saved extra bytes before throwing, so resumption is impossible even in principle.

## The bug

```cpp
// cpp/src/Ice/SSL/SchannelTransceiverI.cpp:813-820, 905-924
IceInternal::SocketOperation op = sslHandshake(&extraBuffer);
_extraBuffer.b.clear();                    // cleared before the throw
if (op == IceInternal::SocketOperationNone) { _sslConnectionRenegotiating = false; continue; }
throw ConnectionLostException(__FILE__, __LINE__);   // any Read/Write result → fatal
```

## Impact

Any client-side renegotiation / post-handshake exchange that requires network I/O is converted into a `ConnectionLostException`, dropping the connection. The mislabeled exception also obscures the real cause.

## Source references

- **(Ice / renegotiation Read/Write result thrown as ConnectionLost)** — [`cpp/src/Ice/SSL/SchannelTransceiverI.cpp#L805-L820`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/Ice/SSL/SchannelTransceiverI.cpp#L805-L820)

## Recommendation

Treat `SocketOperationRead`/`SocketOperationWrite` from `sslHandshake` during renegotiation as "needs I/O" — return that operation and resume the handshake on the next call, preserving `_extraBuffer` — rather than throwing. Only treat a true error as fatal.

## Severity

**Medium** — routine client-side renegotiation requiring network I/O aborts the connection on Windows.

pepone:
Fixes #5449.

### Problem
On Windows (Schannel), a peer-initiated mid-stream TLS renegotiation aborted the client connection with a mislabeled `ConnectionLostException`. `sslHandshake` legitimately returns `SocketOperationRead`/`SocketOperationWrite` when a resumed renegotiation needs network I/O (a TLS 1.2 renegotiation always needs a round trip), but `read()`/`finishRead()` treated anything other than `SocketOperationNone` as fatal — and cleared the saved `_extraBuffer` bytes before throwing, so resumption was impossible.

### Why not make renegotiation work
Completing a renegotiation means driving the full SSL handshake (which needs both reads and writes) from the established data path. That does not fit the transport IO model: the post-handshake read path cannot drive writes, and on the IOCP path `finishRead()` is `void`, so it cannot defer for I/O at all. Supporting it would require re-entering a handshake IO state at the `ConnectionI` level.

The server side already rejects peer-initiated renegotiation outright (a CPU-asymmetric DoS vector on TLS 1.2). This makes the client consistent with it.

### Fix
Reject a peer-initiated renegotiation with a clear `ProtocolException` at the point it is detected in `decryptMessage` (symmetric with the existing server-side rejection), instead of attempting it and later aborting with a mislabeled `ConnectionLostException`. The now-dead renegotiation handling in `read()`/`finishRead()` and the unused `_extraBuffer`/`_sslConnectionRenegotiating` members are removed.

### Testing
:warning: Windows-only code; I could not build or exercise live TLS 1.2 renegotiation locally. This relies on CI (Windows build) and review. The change only removes paths and converts the renegotiation case into a clear rejection.

### DIFF
--- a/changelog.d/cpp/schannel-renegotiation.md
+++ b/changelog.d/cpp/schannel-renegotiation.md
@@ -1,0 +1,2 @@
+- Fixed the Schannel (Windows) SSL transport to reject mid-stream TLS renegotiation with a clear error instead of
+  aborting the connection with a mislabeled `ConnectionLostException`.
--- a/cpp/src/Ice/SSL/SchannelTransceiverI.cpp
+++ b/cpp/src/Ice/SSL/SchannelTransceiverI.cpp
@@ -458,12 +458,9 @@ Schannel::TransceiverI::sslHandshake(SecBuffer* initialBuffer)
         throw SecurityException(__FILE__, __LINE__, os.str());
     }
 
-    assert(!_peerCertificate || _sslConnectionRenegotiating);
-    if (_peerCertificate)
-    {
-        CertFreeCertificateContext(_peerCertificate);
-        _peerCertificate = nullptr;
-    }
+    // sslHandshake only runs for the initial handshake (mid-stream renegotiation is rejected), so the peer
+    // certificate has not been queried yet.
+    assert(!_peerCertificate);
 
     err = QueryContextAttributes(&_ssl, SECPKG_ATTR_REMOTE_CERT_CONTEXT, &_peerCertificate);
     if (err != SEC_E_OK && err != SEC_E_NO_CREDENTIALS)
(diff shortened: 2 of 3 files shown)


pepone:
Closing this â€” it's a regression, and the original code is correct.

`SEC_I_RENEGOTIATE` from `DecryptMessage` is not only an actual renegotiation: in **TLS 1.3** it's how Schannel signals normal **post-handshake messages** (session tickets / key updates). The existing code already handles these by running `sslHandshake`, which completes **synchronously** (`op == SocketOperationNone` â†’ `continue`), so normal TLS 1.3 connections work.

This PR rejected **all** `SEC_I_RENEGOTIATE`, which broke every TLS 1.3 connection at validation â€” CI failed the C++ Schannel SSL tests across the board (while C#/Java passed). See #5449, which is closed as not planned: the underlying audit finding was overstated/misframed and the in-practice behavior is correct.

PR #5706, opened 2026-06-25
Propagate a located servant's dispatch error in Swift ServantManager

### LINKED ISSUE #5555: [Audit-Low] Swift ServantManager converts the located servant's dispatch error locally — error-inspecting middleware never observes the throw (diverges from C#/JS)
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Swift source (pure-Swift runtime + Objective-C++ bridge) performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; verify the reasoning, reproducibility, and severity before acting on it. Verdict **PARTIAL** — half of the original claim was retired by the verification pass (see below).

## Summary

In `ServantManager.dispatch`'s servant-locator path, Swift **catches** the located servant's dispatch error and converts it to a response in place via `makeOutgoingResponse(error:)`. The C# and JS runtimes instead wrap the dispatch in try/finally (so `locator.finished` still runs) and let the error **propagate** out through the middleware pipeline. Since `ServantManager` is the innermost dispatcher, the Swift conversion short-circuits any installed middleware that inspects or transforms dispatch errors — the wire result is usually identical, but the observable error path differs.

## The bug

```swift
// swift/src/Ice/ServantManager.swift:195-218
if let servant = servant {
    let response: OutgoingResponse
    do {
        response = try await servant.dispatch(request)
    } catch {
        response = current.makeOutgoingResponse(error: error)   // converted locally — middleware never sees the throw
    }
    try locator.finished(curr: current, servant: servant, cookie: cookie)
    return response
}
```

C#: [`ServantManager.cs#L42-L61`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/csharp/src/Ice/Internal/ServantManager.cs#L42-L61); JS: [`ServantManager.js#L43-L58`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/js/packages/ice/src/Ice/ServantManager.js#L43-L58).

**Retired half (per the verification pass):** the original report also flagged a missing `request.inputStream.skipEncapsulation()` on the locator/not-found paths (which C#/JS do for batch-stream reuse). This does **not** apply to Swift — `DispatchAdapter.mm` calls C++ `readEncapsulation` and hands Swift an independent copied/moved encapsulation before dispatch ([`DispatchAdapter.mm#L19-L48`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/swift/src/IceImpl/DispatchAdapter.mm#L19-L48)), so batched requests don't share a live input stream.

## Impact

A servant resolved by a `ServantLocator` whose dispatch throws, with error-inspecting middleware installed: the middleware cannot observe or transform the original error. No wire-format effect.

## Source references

Pinned to commit `e248644d788c7e76e14af8676529201c54c1735d`: [`ServantManager.swift#L195-L218`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/swift/src/Ice/ServantManager.swift#L195-L218)

## Recommendation

Mirror C#/JS: run `locator.finished` via `defer`/try-finally and let the dispatch error propagate instead of converting it in place.

## Severity

**Low** — a behavioral divergence visible only to error-inspecting middleware on the servant-locator path.

pepone:
Fixes #5555.

In `ServantManager.dispatch`'s servant-locator path, the located servant's dispatch error was caught and converted to a response in place via `makeOutgoingResponse(error:)`. Because `ServantManager` is the innermost dispatcher, this short-circuited any error-inspecting middleware — it never observed the throw. The plain-servant path in the same dispatcher, and the C# and JS runtimes, instead let the error propagate (running `locator.finished` in a `finally`).

`ServantManager` now runs `locator.finished` once on both the success and error paths and re-throws the dispatch error, so error-inspecting middleware observes it — restoring consistency with the plain-servant path and C#/JS. No wire-format change for clients.

Added a probing test in `Ice/middleware` (a `ServantLocator` returning a throwing servant + an error-observing middleware): it fails before the fix (the middleware sees a converted response) and passes after (it observes the throw). Verified red→green; `swift format --strict` clean.

### DIFF
--- a/swift/src/Ice/ServantManager.swift
+++ b/swift/src/Ice/ServantManager.swift
@@ -203,7 +203,7 @@ final class ServantManager: Dispatcher {
                 } catch {
                     // The locator returned a servant, so finished must run once no matter what. Then let the
                     // dispatch error propagate (instead of converting it here) so error-inspecting middleware
-                    // observes it, like the plain-servant path above and the C# and JS implementations.
+                    // observes it.
                     try locator.finished(curr: current, servant: servant, cookie: cookie)
                     throw error
                 }
--- a/swift/src/Ice/ServantManager.swift
+++ b/swift/src/Ice/ServantManager.swift
@@ -201,10 +201,14 @@ final class ServantManager: Dispatcher {
                 do {
                     response = try await servant.dispatch(request)
                 } catch {
-                    response = current.makeOutgoingResponse(error: error)
+                    // The locator returned a servant, so finished must run once no matter what. Then let the
+                    // dispatch error propagate (instead of converting it here) so error-inspecting middleware
+                    // observes it, like the plain-servant path above and the C# and JS implementations.
+                    try locator.finished(curr: current, servant: servant, cookie: cookie)
+                    throw error
                 }
 
-                // If the locator returned a servant, we must execute finished once no matter what.
+                // Success path: the locator returned a servant, so finished must run once.
                 try locator.finished(curr: current, servant: servant, cookie: cookie)
 
                 return response
--- a/swift/test/Ice/middleware/AllTests.swift
+++ b/swift/test/Ice/middleware/AllTests.swift
@@ -72,8 +72,7 @@ func allTests(_ helper: TestHelper) async throws {
         }
 
         // The located servant always throws, so the client must see a dispatch error, and the middleware
-        // must have observed it as a throw. Before the fix, ServantManager converts the located servant's
-        // error to a response, so the middleware never sees the throw.
+        // must have observed it as a throw.
         try test(threw)
         try await test(log.observed)
 
--- a/swift/test/Ice/middleware/AllTests.swift
+++ b/swift/test/Ice/middleware/AllTests.swift
@@ -48,7 +48,7 @@ func allTests(_ helper: TestHelper) async throws {
     }
 
     // Verifies that a dispatch error from a servant returned by a ServantLocator propagates through the
-    // middleware pipeline, instead of being converted to a response inside ServantManager (issue #5555).
+    // middleware pipeline, instead of being converted to a response inside ServantManager.
     func testMiddlewareObservesLocatorDispatchError(_ communicator: Communicator, _ output: TextWriter)
         async throws
     {
--- a/swift/test/Ice/middleware/AllTests.swift
+++ b/swift/test/Ice/middleware/AllTests.swift
@@ -12,6 +12,7 @@ func allTests(_ helper: TestHelper) async throws {
     let output = helper.getWriter()
 
     try await testMiddlewareExecutionOrder(communicator, output)
+    try await testMiddlewareObservesLocatorDispatchError(communicator, output)
 
     // Verifies the middleware execute in installation order.
     func testMiddlewareExecutionOrder(_ communicator: Communicator, _ output: TextWriter) async throws {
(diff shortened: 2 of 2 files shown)


swift/src/Ice/ServantManager.swift
```diff
@@ -201,10 +201,14 @@ final class ServantManager: Dispatcher {
                 do {
                     response = try await servant.dispatch(request)
                 } catch {
-                    response = current.makeOutgoingResponse(error: error)
+                    // The locator returned a servant, so finished must run once no matter what. Then let the
+                    // dispatch error propagate (instead of converting it here) so error-inspecting middleware
+                    // observes it, like the plain-servant path above and the C# and JS implementations.
+                    try locator.finished(curr: current, servant: servant, cookie: cookie)
+                    throw error
```
pepone, line 208:
```suggestion
                    // The locator returned a servant, so finished must run once no matter what. Then let the
                    // dispatch error propagate (instead of converting it here) so error-inspecting middleware
                    // observes it.
                    try locator.finished(curr: current, servant: servant, cookie: cookie)
                    throw error
```

swift/test/Ice/middleware/AllTests.swift
```diff
@@ -46,6 +47,78 @@ func allTests(_ helper: TestHelper) async throws {
         oa.destroy()
     }
 
+    // Verifies that a dispatch error from a servant returned by a ServantLocator propagates through the
+    // middleware pipeline, instead of being converted to a response inside ServantManager (issue #5555).
+    func testMiddlewareObservesLocatorDispatchError(_ communicator: Communicator, _ output: TextWriter)
+        async throws
```
pepone, line 53:
```suggestion
    // Verifies that a dispatch error from a servant returned by a ServantLocator propagates through the
    // middleware pipeline, instead of being converted to a response inside ServantManager.
    func testMiddlewareObservesLocatorDispatchError(_ communicator: Communicator, _ output: TextWriter)
        async throws
```

swift/test/Ice/middleware/AllTests.swift
```diff
@@ -46,6 +47,78 @@ func allTests(_ helper: TestHelper) async throws {
...
+        do {
+            _ = try await p.getName()
+        } catch {
+            threw = true
+        }
+
+        // The located servant always throws, so the client must see a dispatch error, and the middleware
+        // must have observed it as a throw. Before the fix, ServantManager converts the located servant's
+        // error to a response, so the middleware never sees the throw.
+        try test(threw)
```
pepone, line 76:
```suggestion
        // The located servant always throws, so the client must see a dispatch error, and the middleware
        // must have observed it as a throw.
        try test(threw)
```

swift/src/Ice/ServantManager.swift
```diff
@@ -201,10 +201,14 @@ final class ServantManager: Dispatcher {
                 do {
                     response = try await servant.dispatch(request)
                 } catch {
```
externl, line 203: What about using a defer statement. Something like:

```swift
defer {
    try locator.finished(curr: current, servant: servant, cookie: cookie)
}
```

externl (approved):

bernardnormier (approved):
LGTM. The fix mirrors C#'s `try { dispatch } finally { finished }` semantics exactly (finished runs once on both paths; if finished itself throws, its error wins), and there's no wire-format change since the located path now propagates like the plain-servant path already does.

Non-blocking aside: Java's located-servant path still converts a `UserException` to a response in-place (ServantManager.java `handle`, while it re-throws Error/RuntimeException), so it has the same divergence narrowed to UserExceptions. Given ServantLocator + middleware is a rare combination, this is at most a low-priority follow-up — not something to address here.

swift/src/Ice/ServantManager.swift
pepone: 🤖 **Claude:**

`finished` is throwing (`func finished(curr:servant:cookie:) throws`), and Swift doesn't allow an error to propagate out of a `defer` body — so `defer { try locator.finished(...) }` won't compile ("errors cannot be thrown out of a defer body").

The ways to make it compile each change behavior:
- `try?` swallows a `finished` error (it should surface, like C#/JS);
- `try!` turns it into a crash;
- a `do/catch` inside the `defer` compiles, but a `defer` still can't re-throw to the caller, so `finished`'s error can't propagate — and on the dispatch-error path it can't win over the dispatch error the way C#'s `try { … } finally { finished }` does.

So we need a "finally whose throw propagates," which `defer` can't express — hence the two explicit `try locator.finished(...)` calls. If the duplication is the concern, a `Result` holds the dispatch outcome across a single `finished` call; happy to push that if preferred.

PR #5710, opened 2026-06-25
Invoke properties admin update callbacks outside the admin lock (C++, C#, Java)

### LINKED ISSUE #5669: Properties admin update callbacks are invoked while holding the admin lock (C++, C#, Java)
Split out from #5487 (item 7) for separate tracking.

### Problem

`NativePropertiesAdmin::setProperties` invokes the registered property-update callbacks **while holding the admin mutex**. This is a lock-order inversion hazard: a callback that acquires an application lock runs under the admin lock, so if another thread holds that application lock and then calls into the admin (e.g. `addUpdateCallback` / `removeUpdateCallback`), the two locks can be acquired in opposite orders.

This affects all three mappings, which all invoke the callbacks under the lock:

- **C++** — `cpp/src/Ice/NativePropertiesAdmin.cpp`, `setProperties` (whole method under `lock_guard{_mutex}`).
- **C#** — `csharp/src/Ice/NativePropertiesAdmin.cs`, `setProperties` (whole method under `lock (_mutex)`).
- **Java** — `java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/NativePropertiesAdmin.java`, `setProperties` (`synchronized` method).

### Severity / reachability

Low. The inversion requires an application to call `add`/`removeUpdateCallback` while holding a lock that its own update callback also acquires. `setProperties` itself is an Ice-dispatched operation, so it is not realistically invoked with an application lock held. Each mapping already copies the callback list before iterating, so a callback that (de)registers callbacks during the call is already handled.

### Note

This was intentionally **not** fixed in C++ alone (see #5665), to avoid a cross-mapping inconsistency. If addressed, the callbacks should be copied under the lock and invoked **after** releasing it, consistently across C++, C#, and Java.

externl:
`NativePropertiesAdmin::setProperties` invoked the registered property-update callbacks while holding the admin mutex, which is a lock-order inversion hazard: a callback that acquires an application lock runs under the admin lock, so if another thread holds that application lock and then calls into the admin (e.g. `addUpdateCallback` / `removeUpdateCallback`), the two locks can be acquired in opposite orders.

This copies the callbacks (and the change set passed to them) under the lock and invokes them after releasing it, consistently across all three mappings. C++ switches the whole-method `lock_guard` to a `unique_lock` and unlocks before the loop; C# and Java capture the callbacks into method-scope locals and run the invocation loop after the locked region exits. Each mapping already copied the callback list before iterating, so callbacks that (de)register during the call remain handled.

Fixes #5669.

### DIFF
--- a/cpp/src/Ice/NativePropertiesAdmin.cpp
+++ b/cpp/src/Ice/NativePropertiesAdmin.cpp
@@ -38,7 +38,7 @@ Ice::NativePropertiesAdmin::getPropertiesForPrefix(string prefix, const Current&
 void
 Ice::NativePropertiesAdmin::setProperties(PropertyDict props, const Current&)
 {
-    lock_guard lock{_mutex};
+    unique_lock lock{_mutex};
 
     PropertyDict old = _properties->getPropertiesForPrefix("");
     PropertyDict::const_iterator p;
@@ -157,8 +157,11 @@ Ice::NativePropertiesAdmin::setProperties(PropertyDict props, const Current&)
         changes.insert(changed.begin(), changed.end());
         changes.insert(removed.begin(), removed.end());
 
-        // Copy callbacks to allow callbacks to update callbacks
+        // Copy the callbacks and release the lock before invoking them. This allows the callbacks to update the
+        // callbacks, and avoids a lock-order inversion: a callback may acquire an application lock, while another
+        // thread may call into the admin (e.g. add/removeUpdateCallback) while holding that same lock.
         auto callbacks = _updateCallbacks;
+        lock.unlock();
         for (const auto& cb : callbacks)
         {
             cb(changes);
--- a/csharp/src/Ice/NativePropertiesAdmin.cs
+++ b/csharp/src/Ice/NativePropertiesAdmin.cs
@@ -22,6 +22,12 @@ public sealed class NativePropertiesAdmin : PropertiesAdminDisp_
 
     public override void setProperties(Dictionary<string, string> newProperties, Current current)
     {
+        // The callbacks and the change set to pass to them, captured under the lock and invoked after releasing it.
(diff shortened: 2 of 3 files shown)


pepone (approved):

bernardnormier (commented):
The fix looks correct across all three mappings — callbacks are copied under the lock and invoked after release, and the copy preserves the "a callback may (de)register callbacks" semantics. Worth noting this also fixes a latent **self-deadlock**, not just an inversion: the mutex is non-recursive, so under the old code a callback that called `add`/`removeUpdateCallback` (or re-entered `setProperties`) while the lock was held would have deadlocked outright. Might be worth mentioning in the commit message.

Two requests:

**1. Simplify the post-lock guard.** `callbacks` and `changes` are always assigned together (only in the `_updateCallbacks` non-empty branch), so `callbacks != null` ⟺ `changes != null`. Checking both reads as if they could diverge. Clearer to guard on one and assert the invariant:

C# (needs `using System.Diagnostics;` — as in ServantManager.cs):
```csharp
if (callbacks is not null)
{
    Debug.Assert(changes is not null);
    foreach (Action<Dictionary<string, string>> callback in callbacks)
    {
        callback(changes);
    }
}
```

Java:
```java
if (callbacks != null) {
    assert changes != null;
    for (Consumer<Map<String, String>> callback : callbacks) {
        callback.accept(changes);
    }
}
```

**2. CHANGELOG.** This is a user-facing deadlock fix but there's no `changelog.d` fragment — could you add one?

Otherwise LGTM.

bernardnormier:
Scratch my CHANGELOG request above — on reflection this race is rare enough in practice that it doesn't warrant an entry. Just the guard-simplification stands.

bernardnormier (approved):

PR #5716, opened 2026-06-25
Standardize NSError out-parameter nullability in the IceImpl Swift bridge

### LINKED ISSUE #5714: Standardize NSError out-parameter nullability in the IceImpl Swift bridge
## Summary

The Objective-C++ methods in `swift/src/IceImpl/*.mm` are **bridge methods called only from the Swift Ice layer**. Their trailing `error:` out-parameter is dereferenced unconditionally:

```objc
catch (...)
{
    *error = convertException(std::current_exception());
    return nil;
}
```

This is **correct**: Swift's `throws` bridging always passes the address of a local `NSError?`, so the outer pointer is never `nil`. No `if (error)` guard is needed (and tools like Copilot flagging a potential nil-deref here are false positives — see #5707).

## Problem

The nullability annotations on these out-parameters are inconsistent, and the dominant spelling contradicts both the calling convention and the guard-free dereference:

- A few methods use bare `error:(NSError**)error` (e.g. `Communicator.h`, `PropertiesAdmin.h`). Under `NS_ASSUME_NONNULL_BEGIN`, this means **outer `_Nonnull`**, inner unspecified — the accurate contract.
- Most methods explicitly declare `error:(NSError* _Nullable* _Nullable)error` (Connection.h, IceUtil.h, ObjectPrx.h, ObjectAdapter.h, Properties.h, and most of Communicator.h), marking the **outer pointer `_Nullable`** — which says the caller may pass `nil`, contradicting the unconditional `*error = ...`.

## Proposed fix

Standardize every `error:` out-parameter in the IceImpl bridge on **outer `_Nonnull`, inner `_Nullable`** — i.e. `NSError* _Nullable * _Nonnull error` (or simply the bare `NSError**` the assume-nonnull region already promotes to outer `_Nonnull`). Apply consistently across the `.h` declarations and `.mm` definitions.

This is an **annotation-only** change — no runtime behavior changes, since Swift never passes a nil error slot. The benefit is an accurate API contract and silencing this recurring class of false-positive "guard the error out-param" review comments.

Surfaced while reviewing #5707.

externl:
Standardizes every `error:` out-parameter in the IceImpl Swift bridge on `NSError* _Nullable* _Nonnull`, replacing the mixed bare `NSError**` and `NSError* _Nullable* _Nullable` spellings. The outer pointer is non-null because Swift's `throws` bridging always passes the address of a local `NSError?`, so the unconditional `*error = ...` in the catch blocks is correct and needs no nil guard.

Annotation-only; no runtime behavior change. Fixes #5714.

### DIFF
--- a/swift/src/IceImpl/Communicator.mm
+++ b/swift/src/IceImpl/Communicator.mm
@@ -300,7 +300,9 @@
     }
 }
 
-- (BOOL)addAdminFacet:(id<ICEDispatchAdapter>)dispatchAdapter facet:(NSString*)facet error:(NSError* _Nullable* _Nonnull)error
+- (BOOL)addAdminFacet:(id<ICEDispatchAdapter>)dispatchAdapter
+                facet:(NSString*)facet
+                error:(NSError* _Nullable* _Nonnull)error
 {
     try
     {
--- a/swift/src/IceImpl/Communicator.mm
+++ b/swift/src/IceImpl/Communicator.mm
@@ -52,7 +52,7 @@
     return self.communicator->isShutdown();
 }
 
-- (id)stringToProxy:(NSString*)str error:(NSError**)error
+- (id)stringToProxy:(NSString*)str error:(NSError* _Nullable * _Nonnull)error
 {
     try
     {
@@ -70,7 +70,7 @@
     }
 }
 
-- (nullable id)propertyToProxy:(NSString*)property error:(NSError* _Nullable* _Nullable)error
+- (nullable id)propertyToProxy:(NSString*)property error:(NSError* _Nullable * _Nonnull)error
 {
     try
     {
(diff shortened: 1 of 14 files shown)


externl:
Copilot's five inline comments are all the same point: the `.mm` definitions don't repeat the `nullable` *return* annotation declared in the headers.

These are out of scope for this PR, which standardizes only the `error:` *out-parameter* nullability (issue #5714). The return-type annotations are pre-existing — untouched by this diff — and an unannotated return type in the `.mm` is compatible with a `nullable` declaration, so there's no actual mismatch warning (a mismatch needs a *conflicting* annotation, not a missing one). It's arguably another instance of the false-positive class #5714 is meant to reduce.

That said, there's a reasonable broader cleanup: fully annotate every pointer in the `.mm` definitions to mirror the headers (return types included), which `-Wnullability-completeness` would prefer since these files already carry nullability qualifiers. Reviewers — would you prefer I keep this PR focused on the `error:` parameter, or fold that wider pass in here?

pepone (approved):

bernardnormier (approved):

bernardnormier (approved):

PR #2, opened 2015-06-29
Change 'riterator' to 'itarator' in EvictorBase.java

mkasprz:

Issue #5492, opened 2026-06-12
[low-severity, grouped] IceBT / IceDiscovery / IceLocatorDiscovery low-severity findings (5 items)


pepone:
> **AI-generated audit finding** — these are low-severity items from an automated **correctness** audit of the Ice C++ source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source by a second model (OpenAI **Codex GPT-5**). They have **not been triaged by a human yet**; please verify each before acting on it. Grouped to reduce noise. Links pinned to commit `e248644d788c7e76e14af8676529201c54c1735d`.

This issue collects **low-severity** findings in IceBT, IceDiscovery, and IceLocatorDiscovery.

### 1. (IceBT/DBus) Double `dbus_pending_call_unref` on `set_notify` failure
[`cpp/src/IceBT/DBus.cpp#L758-L791`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceBT/DBus.cpp#L758-L791) — when `dbus_pending_call_set_notify` fails, `init()` unrefs `_call` and throws; the `~AsyncResultI` destructor unrefs it again (refcount underflow / UAF inside libdbus), and the heap `self` is leaked. Don't unref in the error branch (or null `_call`).

### 2. (IceBT) Failed `RegisterProfile` leaves the profile service registered
[`cpp/src/IceBT/Engine.cpp#L765-L815`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceBT/Engine.cpp#L765-L815), [`cpp/src/IceBT/Engine.cpp#L479-L503`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/IceBT/Engine.cpp#L479-L503) — `registerProfileImpl` `addService`s before the remote call; on an error reply / throw it never `removeService`s (unlike `unregisterProfile`), leaking the `ServerProfile` and its callback chain on the long-lived DBus connection.

### 3. (IceDiscovery) Stale exception callbacks corrupt the failure count after a retry
[...]

PR #5719, opened 2026-06-25
Fix IceDiscovery stale-round failure counting and replica-group latency window (C++, C#, Java)

bernardnormier:
An IceDiscovery lookup request is retried in place — the **same `Request` object** is re-`invoke()`d when its round times out. Two issues stemmed from per-round state not being scoped to the round. Both are present identically in C++, C#, and Java (no JS IceDiscovery).

## 1. Stale-round failure counting (correctness)

`invoke()` resets `_failureCount = 0` for the new round, but the per-lookup exception callbacks (one per multicast interface) carried no round identifier. A *delayed* failure callback from an earlier round could increment the new round's `_failureCount` and drive it to `_lookupCount`, finishing the request with "not found" while the current round's datagrams were still outstanding → spurious lookup failure.

The object-identity guard in the exception handlers (`request != the one in the map`) can't catch this, because a retry reuses the same object with the same `_requestId`.

Fix: tag each round with a `_generation` counter (incremented in `invoke()`). The exception callbacks capture the value current when they were sent; `exception(generation)` ignores a callback whose round no longer matches before touching `_failureCount`. Responses still match on the stable `_requestId`, so valid late responses are unaffected — only failure *counting* becomes round-aware.

## 2. Replica-group latency window measured from the wrong point

`AdapterRequest` measures the replica-group aggregation window from `_start`, which was set once at construction and never reset. After a failed round, the window therefore included the earlier timeout, making the request wait an unnecessarily long extra window before aggregating replica endpoints.

Fix: reset `_start` when a new round begins (in `retry()`). An `assert(_latency == 0)` documents/guards the invariant that the latency timer can't already be running on the retry path — `_latency` is only set together with a non-empty `_proxies`, and `retry()` only proceeds when `_proxies` is empty.

This is primarily an efficiency fix (less extra latency). One caveat: because responses match on the stable `_requestId`, a delayed replica-group *reply* from a timed-out earlier round can land in the new round and trigger the window against the freshly-reset `_start`, producing a window too short to collect the current round's replicas — and thus a *smaller* (degraded, not wrong) replica set than the base code. This only occurs when replica RTT ≳ the round timeout, a high-latency regime where discovery is already marginal, so it's a reasonable trade-off.

## Reachability

Low. The exception callbacks come from the UDP multicast lookup invocations, which only fail on a local datagram-send error; item 1 additionally requires such a failure to straddle a retry timeout. Item 2 is an efficiency nit with no correctness impact. No CHANGELOG entry.

Addresses items 3–4 of #5492 (items 1–2 are in a separate PR; item 5 is tracked separately); leaving that issue open.

### DIFF
--- a/cpp/src/IceDiscovery/LookupI.cpp
+++ b/cpp/src/IceDiscovery/LookupI.cpp
@@ -30,6 +30,7 @@ IceDiscovery::Request::retry()
 void
 IceDiscovery::Request::invoke(const string& domainId, const vector<pair<LookupPrx, LookupReplyPrx>>& lookups)
 {
+    ++_invokeId;
     _lookupCount = lookups.size();
     _failureCount = 0;
     Ice::Identity id;
@@ -42,8 +43,15 @@ IceDiscovery::Request::invoke(const string& domainId, const vector<pair<LookupPr
 }
 
 bool
-IceDiscovery::Request::exception()
+IceDiscovery::Request::exception(size_t invokeId)
 {
+    // Ignore a delayed failure from an earlier invocation round: it must not count against the current round, which
+    // reset _failureCount and may still have outstanding lookups.
+    if (invokeId != _invokeId)
+    {
+        return false;
+    }
+
     // If all the invocations on all the lookup proxies failed, report it to the locator.
     if (++_failureCount == _lookupCount)
     {
@@ -69,7 +77,17 @@ AdapterRequest::AdapterRequest(const LookupIPtr& lookup, const std::string& adap
 bool
 AdapterRequest::retry()
 {
-    return _proxies.empty() && --_retryCount >= 0;
+    if (_proxies.empty() && --_retryCount >= 0)
+    {
+        // We only reach here with _proxies empty, i.e. no replica-group response arrived. _latency is set only
+        // together with inserting into _proxies, so the latency timer is not running.
+        assert(_latency == chrono::nanoseconds::zero());
+
+        // Restart the replica-group latency window so it's measured from this round rather than from request creation.
+        _start = chrono::steady_clock::now();
+        return true;
+    }
+    return false;
 }
 
 bool
@@ -127,7 +145,8 @@ AdapterRequest::invokeWithLookup(const string& domainId, const LookupPrx& lookup
         _id,
         lookupReply,
         nullptr,
-        [self = shared_from_this()](exception_ptr ex) { self->_lookup->adapterRequestException(self, ex); });
+        [self = shared_from_this(), invokeId = _invokeId](exception_ptr ex)
+        { self->_lookup->adapterRequestException(self, ex, invokeId); });
 }
 
 void
@@ -155,7 +174,8 @@ ObjectRequest::invokeWithLookup(const string& domainId, const LookupPrx& lookup,
         _id,
         lookupReply,
         nullptr,
-        [self = shared_from_this()](exception_ptr ex) { self->_lookup->objectRequestException(self, ex); });
+        [self = shared_from_this(), invokeId = _invokeId](exception_ptr ex)
+        { self->_lookup->objectRequestException(self, ex, invokeId); });
 }
 
 void
@@ -390,7 +410,7 @@ LookupI::objectRequestTimedOut(const ObjectRequestPtr& request)
 }
 
 void
-LookupI::adapterRequestException(const AdapterRequestPtr& request, exception_ptr ex)
+LookupI::adapterRequestException(const AdapterRequestPtr& request, exception_ptr ex, size_t invokeId)
 {
     lock_guard lock(_mutex);
     auto p = _adapterRequests.find(request->getId());
@@ -399,7 +419,7 @@ LookupI::adapterRequestException(const AdapterRequestPtr& request, exception_ptr
         return;
     }
 
-    if (request->exception())
+    if (request->exception(invokeId))
     {
         if (_warnOnce)
         {
@@ -448,7 +468,7 @@ LookupI::adapterRequestTimedOut(const AdapterRequestPtr& request)
 }
 
 void
-LookupI::objectRequestException(const ObjectRequestPtr& request, exception_ptr ex)
+LookupI::objectRequestException(const ObjectRequestPtr& request, exception_ptr ex, size_t invokeId)
 {
     lock_guard lock(_mutex);
     auto p = _objectRequests.find(request->getId());
@@ -457,7 +477,7 @@ LookupI::objectRequestException(const ObjectRequestPtr& request, exception_ptr e
         return;
     }
 
-    if (request->exception())
+    if (request->exception(invokeId))
     {
         if (_warnOnce)
         {
--- a/cpp/src/IceDiscovery/LookupI.cpp
+++ b/cpp/src/IceDiscovery/LookupI.cpp
@@ -30,7 +30,7 @@ IceDiscovery::Request::retry()
 void
 IceDiscovery::Request::invoke(const string& domainId, const vector<pair<LookupPrx, LookupReplyPrx>>& lookups)
 {
-    ++_invokeId;
+    ++_generation;
     _lookupCount = lookups.size();
     _failureCount = 0;
     Ice::Identity id;
@@ -43,11 +43,11 @@ IceDiscovery::Request::invoke(const string& domainId, const vector<pair<LookupPr
 }
 
 bool
-IceDiscovery::Request::exception(size_t invokeId)
+IceDiscovery::Request::exception(size_t generation)
 {
     // Ignore a delayed failure from an earlier invocation round: it must not count against the current round, which
     // reset _failureCount and may still have outstanding lookups.
-    if (invokeId != _invokeId)
+    if (generation != _generation)
     {
         return false;
     }
@@ -95,6 +95,7 @@ AdapterRequest::response(const ObjectPrx& proxy, bool isReplicaGroup)
 {
     if (isReplicaGroup)
     {
+        _proxies.insert(proxy);
         if (_latency == chrono::nanoseconds::zero())
         {
             _latency = chrono::duration_cast<chrono::nanoseconds>(chrono::steady_clock::now() - _start) *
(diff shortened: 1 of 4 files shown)


cpp/src/IceDiscovery/LookupI.h
```diff
@@ -35,6 +35,11 @@ namespace IceDiscovery
         int _retryCount;
         size_t _lookupCount{0};
         size_t _failureCount{0};
+
+        // Incremented on each invoke() (i.e. each retry round). The exception callbacks capture the value current when
+        // they were sent, so a delayed failure from an earlier round can be ignored instead of counting against the
+        // current round.
+        size_t _invokeId{0};
```
pepone: Nit (optional): `_invokeId` is really a per-round/generation counter — a single `invoke()` fans out N sends that all capture the same value. `_invocationRound` or `_generation` would convey that a bit more directly. Purely cosmetic; the doc comment already clarifies it.

cpp/src/IceDiscovery/LookupI.cpp
```diff
@@ -69,7 +77,17 @@ AdapterRequest::AdapterRequest(const LookupIPtr& lookup, const std::string& adap
...
 {
-    return _proxies.empty() && --_retryCount >= 0;
+    if (_proxies.empty() && --_retryCount >= 0)
+    {
+        // We only reach here with _proxies empty, i.e. no replica-group response arrived. _latency is set only
+        // together with inserting into _proxies, so the latency timer is not running.
+        assert(_latency == chrono::nanoseconds::zero());
+
+        // Restart the replica-group latency window so it's measured from this round rather than from request creation.
+        _start = chrono::steady_clock::now();
```
pepone, line 87: Re-anchoring `_start` here is correct for the common case, and the `assert(_latency == 0)` is a valid invariant. One subtlety worth a line in the PR description: since responses match on the stable `_requestId`, a delayed replica-group reply from a *timed-out* earlier round can trigger the window against this freshly-reset `_start` and produce a window too short to catch the current round's replicas — yielding a smaller endpoint set than base (degraded, not wrong; only when RTT ≳ timeout). So it's not strictly "no incorrect result." Fine to leave as-is, but worth acknowledging.

cpp/src/IceDiscovery/LookupI.cpp
```diff
@@ -69,7 +77,17 @@ AdapterRequest::AdapterRequest(const LookupIPtr& lookup, const std::string& adap
 bool
 AdapterRequest::retry()
 {
-    return _proxies.empty() && --_retryCount >= 0;
+    if (_proxies.empty() && --_retryCount >= 0)
+    {
+        // We only reach here with _proxies empty, i.e. no replica-group response arrived. _latency is set only
+        // together with inserting into _proxies, so the latency timer is not running.
```
pepone, line 83: Minor (C++ only): `response()` assigns `_latency` and schedules the timer *before* `_proxies.insert(...)`, so this comment's "set only together with inserting into `_proxies`" is slightly off in ordering (C#/Java insert first). The invariant still holds under the lock, but inserting into `_proxies` first would also make it exception-safe — today a throw from `schedule`/`insert` could leave `_latency != 0` with `_proxies` empty (no pending timer).

pepone (approved):
Reviewed both fixes across all three language ports (C++/C#/Java), plus the Ice core async-dispatch path and the sibling IceLocatorDiscovery plugin. Both fixes are correct and well-targeted. The notes below are non-blocking: one is a behavior trade-off worth a line in the PR description, the rest are pre-existing/out-of-scope items I'd suggest as separate follow-ups.

## Fix #1 — stale-round failure counting ✅

The premise holds and the fix is minimal and correct.

- **Bug confirmed:** base counts failures with no round identity, and the object-identity guard (`request != the one in the map`) can't catch a retry because the same `Request` object with the same `_requestId` is reused. A delayed round-N failure callback can drive round N+1's `_failureCount` to `_lookupCount` and fire a spurious `finished(null)` while N+1's datagrams are still outstanding.
- **Fix is sound:** `_invokeId` is captured by value at send time and checked before `++_failureCount`; responses stay matched on the stable `_requestId`, so valid late responses are unaffected — only failure *counting* becomes round-aware. That success/failure asymmetry is the right call.
- **Thread-safety:** `_invokeId` has no lock of its own — it's protected by `LookupI`'s mutex (the same lock guarding `_failureCount`/`_lookupCount`), held by all callers of `invoke()`/`exception()`. Same protection the pre-existing `_failureCount` already relied on, so no new synchronization assumption. The one invariant it rests on — that the async exception callback is never delivered inline under that lock — holds: immediate send failures route through `OutgoingAsyncBase::invokeExceptionAsync()`, which dispatches via `clientThreadPool()->execute(...)`, never inline. No deadlock risk.
- **Nits (optional):** `_invokeId` is really a per-round/generation counter (one `invoke()` fans out N sends sharing one value) — `_invocationRound`/`_generation` would read more clearly. No regression test is added; a unit-level test asserting `exception(staleId)` returns false without bumping `_failureCount` would be cheap and lock in the contract (the end-to-end race isn't deterministically reproducible).

## Fix #2 — replica-group latency window ✅ (one caveat)

- **Bug confirmed:** `_start` was set once at construction and never reset, so after a timed-out round the aggregation window measured in `response()` (`(now - _start) * latencyMultiplier`) included the prior round's full timeout — up to ~`RetryCount × Timeout` (≈900 ms with defaults) of unnecessary wait.
- **Fix is correct:** resetting `_start` at the start of a new round in `AdapterRequest::retry()` is the right, `AdapterRequest`-specific seam, and `assert(_latency == 0)` is a genuinely sound invariant — `_latency` goes non-zero only together with making `_proxies` non-empty (under the lock), and `retry()` enters that block only when `_proxies.empty()`, so the assert can never fire on that path. The C# `readonly`→mutable `_start` change is required (CS0191) and correctly applied.
- **⚠️ Caveat worth noting in the PR description:** the claim *"No incorrect result, just extra latency"* is slightly overstated. Because responses are matched on the stable `_requestId` (round-agnostic, and unchanged by fix #1, which only round-tags *failure* callbacks), a delayed replica-group **reply** from a timed-out earlier round can land in the new round and trigger the window against the freshly-reset `_start`:

  ```
  t0      round 1 sent (_start=t0); no reply within 300ms
  t0+300  timeout → retry(): _start reset to t0+300; round 2 sent
  t0+301  delayed ROUND-1 reply (RTT≈301ms) matches _requestId
          → _latency = (t0+301 − t0+300) = 1ms → 1ms window
  t0+302  window fires → aggregate {R1} only → request erased
  t0+601  round-2 replies arrive → request gone → ignored
  ```

  Base would have collected the fuller set (its over-long window happens to cover them); fix #2 returns a *less complete* replica set here. The returned proxy is still valid (degraded, not wrong), and it only triggers when replica RTT ≳ the timeout — a high-latency regime where discovery is already marginal — so this is low-severity and a reasonable trade-off. A line acknowledging it (or round-scoping the window measurement) would make the description accurate.

## Suggested follow-ups (separate PRs/issues, out of scope here)

1. **IceLocatorDiscovery has the identical stale-round failure-counting defect** that fix #1 addresses: `LocatorI::exception()` is `++_failureCount == _lookups.size() && _pending` with no round guard, and `runTimerTask` resets `_failureCount` and re-sends `findLocatorAsync` with callbacks carrying no round id (`cpp/src/IceLocatorDiscovery/PluginI.cpp:678`+`:645`/`:766`; `csharp/.../PluginI.cs:423`; `java/.../PluginI.java:386`). `LocatorI` is a singleton, so it has no object-identity guard at all. This PR scopes itself to IceDiscovery (items 3–4 of #5492), so this looks untracked.
2. **C#-only divergence in IceLocatorDiscovery:** the C# `runTimerTask` retry path (`PluginI.cs:477-516`) does **not** reset `_failureCount = 0`, whereas C++ (`PluginI.cpp:759`) and Java (`PluginI.java:447`) do — an independent latent over-counting bug even without delayed callbacks.
3. **Pre-existing Java latency-formula bug (not introduced here):** `LookupI.java` computes `_latency = (System.nanoTime() - _start) * _latencyMultiplier / 100000.0`, but ns→ms needs `/ 1000000.0`. The Java replica-group aggregation window is therefore ~10× longer than C++/C#. (C# `/10000.0` on 100 ns ticks is correct; C++ uses typed `chrono::nanoseconds`.)

---

externl (approved):

cpp/src/IceDiscovery/LookupI.h
bernardnormier: Done — renamed to `_generation` across C++/C#/Java in 5f6afbf2a4.

cpp/src/IceDiscovery/LookupI.cpp
bernardnormier: Good catch. Reordered C++ `response()` to insert into `_proxies` before assigning `_latency`/scheduling, matching C#/Java and making it exception-safe (a throw from `schedule` can no longer leave `_latency != 0` with `_proxies` empty). 5f6afbf2a4.

cpp/src/IceDiscovery/LookupI.cpp
bernardnormier: You're right, thanks. Updated the PR description: a delayed replica-group reply from a timed-out earlier round can fire the freshly-reset window and yield a smaller (degraded, not wrong) replica set when RTT ≳ timeout — so it's a reasonable trade-off rather than strictly 'no incorrect result.'

bernardnormier:
Thanks for the thorough review @pepone. Addressed the in-scope notes in 5f6afbf2a4 (rename `_invokeId` → `_generation`, reorder C++ `response()`) and updated the PR description with the replica-group latency caveat.

Filed the three out-of-scope follow-ups you identified as separate issues:
- #5725 — IceLocatorDiscovery has the same stale-round failure-counting defect (C++/C#/Java; singleton, so no object-identity guard)
- #5726 — IceLocatorDiscovery C# `runTimerTask` retry path doesn't reset `_failureCount` (C#-only divergence from C++/Java)
- #5727 — IceDiscovery Java replica-group latency window is ~10× too long (ns→ms divisor `100000.0` should be `1000000.0`)

PR #5723, opened 2026-06-26
Fix DataStorm getLastIds null dereference on multi-key element detach

### LINKED ISSUE #5722: [Audit-Low] DataStorm null dereference (crash) in SessionI::getLastIds when a multi-key reader/writer is destroyed

## Summary

A DataStorm node can crash with a null-pointer dereference when a **multi-key** reader or writer is destroyed and the affected topic later re-synchronizes with its peer (for example, after a session reconnect). The crash happens inside `DataStorm::SessionI::getLastIds` (`cpp/src/DataStorm/SessionI.cpp`).

## Background

DataStorm readers and writers can be *multi-key*: a single element subscribes to (or publishes under) several keys at once — see `makeMultiKeyReader` / `makeMultiKeyWriter`.

To deliver each sample exactly once across a disconnect/reconnect, a session tracks, per remote key, which local elements are attached and the id of the last sample each has seen. When a session (re)attaches to a peer, `SessionI::getLastIds` collects those "last seen" ids so the peer knows where to resume. It has two branches: one for filter subscriptions and one for key subscriptions.

## The defect

In `getLastIds`, the **filter** branch null-checks the per-element subscriber before using it, but the **key** branch does not:

```cpp
// key branch
for (const auto& [elementId, _] : subscriber.keys[keyOrFilterId].second)
{
    // subscriber.get(elementId) can be null -> dereference crash
    lastIds.emplace(elementId, subscriber.get(elementId)->getSubscriber(element)->lastId);
}
```

`subscriber.get(elementId)` returns null when an element id still appears in a key's subscriber set but its `ElementSubscribers` entry has already been removed. That this can happen is part of the code's own model — `SessionI::unsubscribeFromKey` explicitly handles `get(elementId) == nullptr` as a normal case. The key branch is simply missing the guard its sibling filter branch has.

## How the dangling id arises (root cause)

When an element detaches, `DataStorm::DataElementI::detachKey` (`cpp/src/DataStorm/DataElementI.cpp`) tells the session to unsubscribe using the element's **first** key id for *every* key, instead of the key actually being detached.

So a multi-key element subscribed under keys `K1` and `K2` detaches as two `unsubscribeFromKey(element, K1)` calls:
- the first removes the element's `ElementSubscribers` entry and decrements `K1`'s per-element counter;
- the second runs against `K1` again (its counter is decremented a second time, leaving a phantom negative value), while **`K2`'s counter for the element is never decremented** and keeps a dangling reference to the now-removed element.

A subsequent `getLastIds(..., K2, ...)` iterates `K2`'s subscriber set, hits the dangling element id, and dereferences null.

## When it triggers

The crash requires all of:
1. a multi-key element under keys `{K1, K2}` is destroyed (so it detaches), and
2. another element is still subscribed under one of those shared keys (so the key's set survives), and
3. the topic re-synchronizes afterward — a new element attaches under that key, or the session reconnects — so `getLastIds` runs over the stale entry.

Narrow, but reachable: it is a hard node crash on a destroy + reconnect path. No existing test covers a multi-key detach followed by `getLastIds`.

## Suggested fix

**Minimal:** null-check in the key branch, matching the filter branch — skip an element that is no longer present, since there is nothing to resume for it.

```cpp
for (const auto& [elementId, _] : subscriber.keys[keyOrFilterId].second)
{
    if (auto* elementSubscribers = subscriber.get(elementId))
    {
        if (auto* s = elementSubscribers->getSubscriber(element))
        {
            lastIds.emplace(elementId, s->lastId);
        }
    }
}
```

**Deeper (separate, more invasive):** fix `detachKey` to pass the actual per-key id so the multi-key per-key counters stay consistent. This also removes the phantom negative counter and a small bookkeeping leak left by the same bug.

## Severity

Low–Medium — a genuine null-dereference crash, but gated on the specific multi-key-detach-plus-reconnect scenario above.

pepone:
Fixes #5722.

`SessionI::getLastIds` dereferenced `subscriber.get(elementId)` and `getSubscriber(element)` without a null check in its **key** branch, while the sibling **filter** branch guards the identical lookup. The codebase's own model already allows the null — `SessionI::unsubscribeFromKey` treats `get(elementId) == nullptr` as a normal case.

**Root cause:** `DataElementI::detachKey` unsubscribes *every* key of an element with the element's **first** key id (`subscriber->id`), not the key actually being detached. So when a multi-key element under `{k1, k2}` detaches, `k1`'s counter is decremented twice and **`k2`'s entry is left dangling** (pointing at the now-removed element). A later `getLastIds(…, k2, …)` — driven from `TopicI::getElementSpecs` during (re)attach — then dereferences null and crashes the node.

**Fix:** guard the key branch like the filter branch. Skipping an element that is no longer present is correct — there is nothing to resume for it. (The deeper `detachKey` accounting fix — passing the actual per-key id — is tracked separately as the more invasive change.)

**Red → green test** (`DataStorm/partial`): a persistent multi-key reader attaches to a multi-key writer under both keys; the writer is destroyed (detach → stale key entry) and a second writer attaches, driving `getLastIds` over the stale entry. Verified locally:
- **Without the fix:** the reader process **SIGSEGVs** (`EXC_BAD_ACCESS` / `KERN_INVALID_ADDRESS`) with the faulting stack through `TopicI::getElementSpecs` → `SessionI::getLastIds`.
- **With the fix:** `testing multi-key writer detach then reattach... ok` (both plain and multicast configs).

### DIFF
--- a/changelog.d/datastorm/5722-multikey-getlastids-crash.md
+++ b/changelog.d/datastorm/5722-multikey-getlastids-crash.md
@@ -1,0 +1,2 @@
+- Fixed a crash in DataStorm (a null-pointer dereference during session (re)attachment) that could occur after a
+  multi-key reader or writer was destroyed.
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -1147,10 +1147,19 @@ SessionI::getLastIds(int64_t topicId, int64_t keyOrFilterId, const std::shared_p
         else
         {
             // Key subscription: report this element's lastId for each remote element it's subscribed to under this
-            // key. Key subscriptions are indexed by remote key id, so look the key up directly.
+            // key. Key subscriptions are indexed by remote key id, so look the key up directly. An element id can
+            // remain in the key map after its ElementSubscribers was removed (e.g. a multi-key element detaches
+            // under one key but its other keys are still counted), so skip ids with no live subscriber, like the
+            // filter branch above.
             for (const auto& [elementId, _] : subscriber.keys[keyOrFilterId].second)
             {
-                lastIds.emplace(elementId, subscriber.get(elementId)->getSubscriber(element)->lastId);
+                if (auto* elementSubscribers = subscriber.get(elementId))
+                {
+                    if (auto* s = elementSubscribers->getSubscriber(element))
+                    {
+                        lastIds.emplace(elementId, s->lastId);
+                    }
+                }
             }
         }
     }
--- a/cpp/test/DataStorm/partial/Reader.cpp
+++ b/cpp/test/DataStorm/partial/Reader.cpp
@@ -91,6 +91,28 @@ void ::Reader::run(int argc, char* argv[])
         test(aapl->lastBid == 13.0f); // AAPL's own bid, not GOOG's 101
         test(aapl->lastAsk == 14.0f); // AAPL's own ask, not GOOG's 102
     }
+
+    // A persistent multi-key reader attaches to writer 1 under both keys; writer 1's detach leaves a stale entry
+    // in the reader's per-key subscriber map, and writer 2's attach drives SessionI::getLastIds over it, which
+    // must not dereference null.
+    Topic<string, int> detachTopic(node, "multiKeyDetach");
+    detachTopic.setReaderDefaultConfig(config);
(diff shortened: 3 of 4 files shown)


externl (approved):
Looks good!

bernardnormier (approved):

PR #5728, opened 2026-06-26
Fix IceLocatorDiscovery stale-round failure counting (C++, C#, Java)

### LINKED ISSUE #5725: IceLocatorDiscovery has the same stale-round failure-counting defect as IceDiscovery
Follow-up from the review of #5719 (which fixes this defect in **IceDiscovery**). **IceLocatorDiscovery** has the identical issue.

A locator lookup request is retried in place: `runTimerTask` resets `_failureCount = 0` and re-sends `findLocatorAsync` to every lookup proxy, with per-lookup exception callbacks that carry **no round identifier**. A delayed failure callback from an earlier round can increment the new round's `_failureCount` and drive it to `_lookups.size()`, finishing the request early while the current round's datagrams are still outstanding.

Unlike IceDiscovery, `LocatorI` is a **singleton**, so there is no object-identity guard at all.

Locations:
- C++ `cpp/src/IceLocatorDiscovery/PluginI.cpp` — `LocatorI::exception()` (`++_failureCount == _lookups.size() && _pending`) and the `findLocatorAsync` callbacks in `getLocators`/`runTimerTask`.
- C# `csharp/src/IceLocatorDiscovery/PluginI.cs` — `exception()` around line 419.
- Java `java/src/com.zeroc.icediscovery/.../PluginI.java` — `exception()` around line 386.

Fix mirrors #5719: tag each round with a `_generation` counter captured by the exception callbacks, and ignore a callback whose generation no longer matches before touching `_failureCount`.

Reachability is low (same as #5719: requires a local datagram-send failure straddling a retry timeout). Found during review of #5719; part of #5492.

### LINKED ISSUE #5726: IceLocatorDiscovery C# runTimerTask retry path does not reset _failureCount
Follow-up from the review of #5719. C#-only divergence in **IceLocatorDiscovery**.

In `csharp/src/IceLocatorDiscovery/PluginI.cs`, the initial lookup path resets `_failureCount = 0` before sending (around line 352), but the retry path in `runTimerTask` (around line 477-516) re-sends `findLocatorAsync` to all lookups **without** resetting `_failureCount`. The carried-over count makes `++_failureCount == _lookups.Count` trip early on a subsequent round.

C++ (`PluginI.cpp`) and Java (`PluginI.java`) both reset `_failureCount = 0` on the retry path, so this is a C#-only latent over-counting bug — independent of the delayed-callback race tracked separately (and even without delayed callbacks).

Fix: reset `_failureCount = 0` in the C# `runTimerTask` retry branch to match C++/Java. (Best fixed together with the IceLocatorDiscovery stale-round defect.)

Found during review of #5719; part of #5492.

bernardnormier:
Follow-up to #5719, which fixed the same class of bug in **IceDiscovery**. These are the IceLocatorDiscovery counterparts (issues #5725 and #5726).

## 1. Stale-round failure counting (correctness)

`LocatorI` retries a locator lookup in place — the **singleton** is reused across rounds. `runTimerTask` resets `_failureCount = 0` and re-sends `findLocatorAsync` to every lookup proxy, but the per-lookup exception callbacks carried **no round identifier**. A *delayed* failure callback from an earlier round could increment the new round's `_failureCount` and drive it to `_lookups.size()`, finishing the request with the void locator while the current round's datagrams were still outstanding → spurious lookup failure.

Because `LocatorI` is a singleton, there is no per-request object-identity guard at all (unlike a per-`Request` object).

Fix: tag each round with a `_generation` counter (incremented where `_failureCount` is reset, in both the initial `invoke()` path and the `runTimerTask` retry path). The `findLocatorAsync` exception callbacks capture the value current when they were sent; `exception(generation)` ignores a callback whose round no longer matches before touching `_failureCount`. This mirrors the `_generation` fix applied to IceDiscovery in #5719.

## 2. C# `runTimerTask` missing `_failureCount` reset

Independently, the **C#** `runTimerTask` retry path re-sent `findLocatorAsync` to all lookups **without** resetting `_failureCount = 0` — C++ and Java already reset it. So even without a delayed callback, the carried-over count made `++_failureCount == _lookups.Count` trip early on a subsequent round. This PR adds the missing reset (alongside the `_generation` increment).

## Reachability

Low, same as #5719: the exception callbacks come from the UDP multicast lookup invocations, which only fail on a local datagram-send error; the stale-round race additionally requires such a failure to straddle a retry timeout. The C# missing-reset bug is also low-impact (requires a send failure on a retried round). No CHANGELOG entry.

Fixes #5725.
Fixes #5726.

### DIFF
--- a/cpp/src/IceLocatorDiscovery/PluginI.cpp
+++ b/cpp/src/IceLocatorDiscovery/PluginI.cpp
@@ -76,7 +76,7 @@ namespace
 
         vector<Ice::LocatorPrx> getLocators(const string&, const chrono::milliseconds&);
 
-        void exception(std::exception_ptr);
+        void exception(std::exception_ptr, size_t generation);
 
     private:
         void runTimerTask() final;
@@ -99,6 +99,10 @@ namespace
         bool _pending{false};
         int _pendingRetryCount{0};
         size_t _failureCount{0};
+        // Incremented at the start of each lookup round. The findLocatorAsync exception callbacks capture the value
+        // current when they were sent, so a delayed failure from an earlier round is ignored instead of counting
+        // against the current round.
+        size_t _generation{0};
         bool _warnOnce{true};
         vector<RequestPtr> _pendingRequests;
         std::mutex _mutex;
@@ -629,6 +633,7 @@ LocatorI::invoke(const optional<Ice::LocatorPrx>& locator, const RequestPtr& req
         {
             _pending = true;
             _failureCount = 0;
+            ++_generation;
             _pendingRetryCount = _retryCount;
             try
             {
@@ -647,7 +652,8 @@ LocatorI::invoke(const optional<Ice::LocatorPrx>& locator, const RequestPtr& req
                         _instanceName,
                         l.second,
                         nullptr,
-                        [self = shared_from_this()](exception_ptr ex) { self->exception(ex); });
+                        [self = shared_from_this(), generation = _generation](exception_ptr ex)
+                        { self->exception(ex, generation); });
                 }
                 _timer->schedule(shared_from_this(), _timeout);
             }
@@ -677,9 +683,15 @@ LocatorI::invoke(const optional<Ice::LocatorPrx>& locator, const RequestPtr& req
 }
 
 void
-LocatorI::exception(std::exception_ptr ex)
+LocatorI::exception(std::exception_ptr ex, size_t generation)
 {
     lock_guard lock(_mutex);
+    // Ignore a delayed failure from an earlier lookup round: it must not count against the current round, which reset
+    // _failureCount and may still have outstanding lookups.
+    if (generation != _generation)
+    {
+        return;
+    }
     if (++_failureCount == _lookups.size() && _pending)
     {
         //
@@ -762,13 +774,15 @@ LocatorI::runTimerTask()
                 }
             }
             _failureCount = 0;
+            ++_generation;
             for (const auto& l : _lookups)
             {
                 l.first->findLocatorAsync(
                     _instanceName,
                     l.second,
                     nullptr,
-                    [self = shared_from_this()](exception_ptr ex) { self->exception(ex); });
+                    [self = shared_from_this(), generation = _generation](exception_ptr ex)
+                    { self->exception(ex, generation); });
             }
             _timer->schedule(shared_from_this(), _timeout);
             return;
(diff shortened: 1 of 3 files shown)


pepone (approved):
Reviewed across all three language ports (C++/C#/Java) and cross-checked independently. This is the clean follow-up to the two IceLocatorDiscovery items flagged on #5719, and it adopts the `_generation` name. Both fixes are correct, thread-safe, and consistent — approving.

## Stale-round failure counting ✅

- **Bug confirmed in base:** `LocatorI` is a reused singleton with no object-identity guard, and the callbacks carried no round id, so a delayed `findLocatorAsync` failure from round N could increment round N+1's `_failureCount` to `_lookups.size()` and trip the all-failed branch (resolving pending requests on the void locator) while N+1's datagrams were still outstanding.
- **`_generation` guard is correct:** bumped at every *operational* `_failureCount` reset (the initial `invoke()` path and the `runTimerTask` retry path — the two callback-send sites), captured by value at send time, and checked in `exception()` *before* `++_failureCount`. The only two unpaired `_failureCount = 0` (`PluginI.cs:139`, `PluginI.java:158`) are constructors, where `_generation` defaults to 0 — no hole.
- **Thread-safety:** every `_generation` read/write is under the single `LocatorI` lock — including the Java retry, which runs in `_retryTask.run()` under `synchronized (LocatorI.this)` (same monitor as `synchronized exception()`/`invoke()`). The callbacks are always dispatched asynchronously (C# `await Task.Yield()` in `performFindLocatorAsync` + `ContinueWith(..., ice_scheduler())` on the retry path; Java `whenCompleteAsync(..., ice_executor())`; C++ AMI), never inline under the lock — so no self-deadlock — and the generation is a true by-value snapshot, not a live re-read.
- **Singleton edges check out:** the success path (`foundLocator` sets `_pending=false` without bumping generation) is safe via the `&& _pending` clause; a second `invoke()` while pending only queues into `_pendingRequests` without starting a round; and `_generation`, though long-lived and monotonic, can't realistically overflow (once per round).

## C# `runTimerTask` missing `_failureCount` reset ✅

Confirmed against base: the C# retry path re-sent `findLocatorAsync` without resetting `_failureCount` (C++/Java already did), so a carried-over count could trip the all-failed branch early even without a delayed callback. Fixed at `PluginI.cs:497`. Good to fold it in here.

## Non-blocking notes

- **No regression test** — consistent with #5719; the race isn't deterministically reproducible without a send-failure injection seam.
- **Pre-existing / out of scope:** `performFindLocatorAsync` reads `_instanceName` off-lock after `await Task.Yield()` while `foundLocator` can mutate it (`PluginI.cs:315`). Benign (atomic reference read; worst case a stale instance name) and unrelated to this change — noting only for the record.

LGTM.

externl (approved):

PR #5731, opened 2026-06-26
Fix DataStorm relayed-session identity collision on fan-in through a relay

### LINKED ISSUE #5724: [Audit-Medium] DataStorm: colliding relay session identities drop a disconnect notification, hanging a fan-in subscriber

## Summary

In a DataStorm "fan-in" relay topology — two or more subscriber applications reaching the same publisher through a shared relay node — a subscriber can hang indefinitely when the relay loses its connection to the publisher. The subscriber is never told its session ended, so its reads (e.g. `getNextUnread()`) block forever even though the publisher is gone.

## Background

DataStorm nodes establish *node sessions* with each other to exchange topic data. A node with no public endpoint (a typical client application) is reached through a relay: the relay hosts a forwarder and tracks, per peer node, the relayed sessions so it can notify them if that peer disconnects.

Each session is represented by a servant whose Ice identity DataStorm assigns from a **per-node counter**: `{name = "<n>", category = "s"}` for subscriber sessions (`"p"` for publisher sessions), where `<n>` is the node's local session count. The relay stores relayed sessions in `NodeSessionI::_sessions`, keyed by that identity.

## The defect

The per-node session counter has **no node-unique component**, so two *different* nodes independently produce the **same** identity for their first session — both get `{name = "1", category = "s"}`.

`NodeSessionI::_sessions` is keyed purely by this identity and updated with `insert_or_assign`. When two subscriber apps both relay a session to the same endpoint-less publisher, they go through the same relay-side `NodeSessionI` (there is one per target node), their entries collide on `{"1","s"}`, and **the second overwrites the first**.

When the relay→publisher connection then drops, `NodeSessionI::destroy()` walks `_sessions` and sends a disconnect notification to each remaining entry — but only one survived the collision. The overwritten subscriber is never notified. Because that subscriber's own connection (subscriber→relay) is still up, nothing else tells it the session ended, so its session never recovers and its reads block indefinitely.

## When it triggers

All of:
1. Two or more subscriber (or publisher) apps connect to the **same** publisher node that has **no public endpoint**, through the **same** relay node.
2. The colliding apps are fresh, so their first sessions get the same identity (`{"1","s"}`) — the common case, since the counter starts from the same value on every node.
3. The relay→publisher connection drops while a colliding subscriber's connection to the relay stays up.

The identity collision is deterministic; the hang then needs the relay→publisher link to drop.

## Suggested fix

Make relayed sessions distinguishable when their identities collide across nodes. Two directions:
- Give session identities a node-unique component (e.g. derived from the node's identity) so sessions from different nodes never collide. (Wider blast radius — the identity is how peers address the session servant.)
- Key the relay's relayed-session map on something source-distinct (e.g. include the originating connection) rather than the bare session identity, so two peers' sessions coexist. (More localized to the relay path.)

A regression test would set up two subscribers → one relay → an endpoint-less publisher, drop the relay→publisher connection, and assert that **both** subscribers observe the disconnect and recover.

## Severity

Medium — a subscriber hangs (no crash, no data corruption), but the trigger is a realistic deployment (a fan-in relay to an endpoint-less publisher) and the identity collision is deterministic.

pepone:
Fixes #5724.

## Problem

`NodeSessionI` tracks relayed publisher/subscriber sessions in `_sessions`, keyed by the session's Ice identity alone. That identity (e.g. `{"1","s"}`) comes from a per-node counter with no node-unique component, so two different source nodes reaching the same endpoint-less peer through one relay node mint the *same* identity. The second session's `insert_or_assign` then overwrites the first in the relay's map.

When the relay later loses its connection to the peer, `destroy()` notifies only the surviving entry, so the overwritten session is never told of the disconnect — its reader/writer blocks forever in `waitForNoWriters()` / `waitForNoReaders()`.

Only **endpoint-less** sources are affected: an endpoint-ful source advertises its own endpoint and so has an independent disconnect path that does not depend on the relay's `_sessions` map.

## Fix

Key `_sessions` by the composite `(source node identity, session identity)` instead, so sessions from different source nodes no longer collide. The session identity stays in the key because a single node can host both a publisher (`"p"`) and a subscriber (`"s"`) session.

`destroy()` is also hardened to catch per session, so one dead or stale session proxy can't stop the remaining sessions from being notified.

## Test

New red→green regression test `DataStorm/fanInRelay`: two subscriber apps and one endpoint-less publisher app connect through a relay; terminating the publisher must unblock both subscribers. Before the fix, one subscriber hangs.

### DIFF
--- a/cpp/src/DataStorm/NodeSessionI.cpp
+++ b/cpp/src/DataStorm/NodeSessionI.cpp
@@ -78,8 +78,13 @@ namespace
                     updateNodeAndSessionProxy(*subscriber, subscriberSessionForwarder, current);
 
                     // Keep track of the subscriber session with the NodeSession, the NodeSession will use this proxy
-                    // to inform the subscriber of the disconnection if the target publisher is disconnected.
+                    // to inform the subscriber of the disconnection if the target publisher is disconnected. Key the
+                    // entry by (subscriber node identity, session identity): the session identity alone is only
+                    // unique within a node, so two different subscriber nodes can produce the same one and overwrite
+                    // each other in the map.
                     nodeSession->addSession(
+                        subscriber->ice_getIdentity(),
+                        subscriberSession->ice_getIdentity(),
                         subscriberIsHostedOnRelay ? subscriberSession->ice_fixed(current.con) : *subscriberSession);
 
                     // Forward the call to the target Node.
@@ -112,8 +117,13 @@ namespace
                     updateNodeAndSessionProxy(*publisher, publisherSessionForwarder, current);
 
                     // Keep track of the publisher session with the NodeSession, the NodeSession will use this proxy
-                    // to inform the publisher of the disconnection if the target subscriber is disconnected.
+                    // to inform the publisher of the disconnection if the target subscriber is disconnected. Key the
+                    // entry by (publisher node identity, session identity): the session identity alone is only unique
+                    // within a node, so two different publisher nodes can produce the same one and overwrite each
+                    // other in the map.
                     nodeSession->addSession(
+                        publisher->ice_getIdentity(),
+                        publisherSession->ice_getIdentity(),
                         publisherIsHostedOnRelay ? publisherSession->ice_fixed(current.con) : *publisherSession);
                     // Forward the request to the target subscriber.
                     _node->confirmCreateSessionAsync(publisher, publisherSessionForwarder, response, exception);
@@ -211,8 +221,15 @@ NodeSessionI::destroy()
 
         for (const auto& [_, session] : _sessions)
         {
-            // Notify sessions of the disconnection, don't need to wait for the result.
-            session->disconnectedAsync(nullptr);
+            // Notify each session of the disconnection (we don't wait for the result). Catch per session so that a
+            // dead or stale session proxy can't prevent the remaining sessions from being notified.
+            try
+            {
+                session->disconnectedAsync(nullptr);
+            }
+            catch (const Ice::LocalException&)
+            {
+            }
         }
     }
     catch (const ObjectAdapterDestroyedException&)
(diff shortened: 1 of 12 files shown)


cpp/src/DataStorm/NodeSessionI.cpp
```diff
@@ -78,8 +78,13 @@ namespace
                     updateNodeAndSessionProxy(*subscriber, subscriberSessionForwarder, current);
 
                     // Keep track of the subscriber session with the NodeSession, the NodeSession will use this proxy
-                    // to inform the subscriber of the disconnection if the target publisher is disconnected.
+                    // to inform the subscriber of the disconnection if the target publisher is disconnected. Key the
+                    // entry by (subscriber node identity, session identity): the session identity alone is only
+                    // unique within a node, so two different subscriber nodes can produce the same one and overwrite
+                    // each other in the map.
```
bernardnormier: The comment in the header is sufficient. No need to duplicate this rationale multiple times.

cpp/src/DataStorm/NodeSessionI.cpp
```diff
@@ -112,8 +117,13 @@ namespace
                     updateNodeAndSessionProxy(*publisher, publisherSessionForwarder, current);
 
                     // Keep track of the publisher session with the NodeSession, the NodeSession will use this proxy
-                    // to inform the publisher of the disconnection if the target subscriber is disconnected.
+                    // to inform the publisher of the disconnection if the target subscriber is disconnected. Key the
+                    // entry by (publisher node identity, session identity): the session identity alone is only unique
```
bernardnormier: One more time!

cpp/src/DataStorm/NodeSessionI.h
```diff
@@ -63,8 +63,9 @@ namespace DataStormI
         // A map containing all publisher and subscriber sessions established with the session's target node via a
         // node forwarder.
         //
-        // The key is the session identity, and the value is the session proxy.
-        std::map<Ice::Identity, DataStormContract::SessionPrx> _sessions;
+        // The key is a pair of (source node identity, session identity); the session identity alone is only unique
+        // within a node, so two different source nodes can produce the same one. The value is the session proxy.
+        std::map<std::pair<Ice::Identity, Ice::Identity>, DataStormContract::SessionPrx> _sessions;
```
bernardnormier, line 71: It would be interesting to comment on why we use a map instead of a vector or list. We only ever insert_or_assign and iterate over `_sessions`. I am guessing it's because we sometimes replace a session proxy, but that's not obvious.

bernardnormier (approved):

externl (approved):

PR #5733, opened 2026-06-26
Add copilot-setup-steps.yml to provision the Copilot coding agent

externl:
This adds a `copilot-setup-steps.yml` workflow so the GitHub Copilot coding agent provisions a working build environment before it starts. The job mirrors the toolchain setup used by the Linux jobs in `ci.yml` (C++, PHP, Python, .NET, Java, Node, and Ruby) so the agent can build and test the various language mappings.

It runs on our dedicated `linux-latest-copilot` runner (which I just created) rather than the shared `ubuntu-*` pool, which is sometimes exhausted by CI and leaves Copilot waiting for a runner. Routing both the coding agent and code review onto the dedicated runner (the latter via the org-level Copilot runner-type setting) keeps them out of that contention (hopefully).

The primary goal is to mitigate: 

Screenshot 2026-06-26 at 1 36 13 PM

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -12,6 +12,9 @@ on:
     paths:
       - .github/workflows/copilot-setup-steps.yml
 
+permissions:
+  contents: read
+
 jobs:
   copilot-setup-steps:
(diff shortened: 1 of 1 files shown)


pepone (approved):
Looks good!

bernardnormier (approved):

PR #5734, opened 2026-06-26
Rename "interface classes" in Slice

### LINKED ISSUE #5559: Cleanup `interface MyClass` in Testing Code
I've come across a handful of Slice files containing the improperly named:
`interface MyClass { ... }`
These should be changed to `MyInterface` or something which accurately reflects the type.

InsertCreativityHere:
This PR fixes #5559.

We had lots of `interfaces` named `MyClass` (or similar), I guess leftover from operations on classes.
This PR renames every instance of `interface .*Class` to `interface .*Interface` instead. For correctness.


cpp/test/Ice/operations/Test.ice
```diff
@@ -267,21 +267,21 @@ module Test
     struct MyStruct1
     {
         string tesT; // Same name as the enclosing module
-        MyClass* myClass; // Same name as an already defined class
+        MyInterface* myInterface; // Same name as an already defined interface
         string myStruct1; // Same name as the enclosing struct
     }
 
-    class MyClass1
```
InsertCreativityHere, line 274: This PR also drops the `1` suffix on this type.
I think it was only called `MyClass1` because `MyClass` was already erroneously taken.
Now that the name is open again, it makes no sense to keep this type named `MyClass1`.

PR #5740, opened 2026-06-26
Fix DataStorm session leak from a pending retry task self-reference cycle

### LINKED ISSUE #5736: [Audit-Medium] DataStorm: a pending retry timer leaks a removed session (and its node/instance) via a self-reference cycle

## Summary

A DataStorm node leaks a session object every time a session to an **endpoint-less peer** is removed after the peer fails to reconnect. The session's pending retry timer holds a strong reference back to the session that nothing ever releases, forming an unbreakable reference cycle. Because each leaked session also keeps its `NodeI` and `Instance` alive, the leak is large per incident and **accumulates over the node's uptime** — it is not a one-time shutdown cost.

## Background

A node with no public endpoint (a typical client application) cannot be dialed back, so when a session to such a peer disconnects, `SessionI::retry` cannot actively reconnect. Instead it schedules a one-shot timer to remove the session if the peer does not reconnect in time:

```cpp
_retryTask = make_shared<IceInternal::InlineTimerTask>([self = shared_from_this()] { self->remove(); });
_instance->scheduleTimerTask(_retryTask, delay);
```

The task's lambda captures `self = shared_from_this()` — a **strong** `shared_ptr<SessionI>` — and the task is stored in the session's own `_retryTask` member.

## The defect

`_retryTask` is cancelled and nulled in exactly two places: `SessionI::connected()` (on a successful reconnect) and `SessionI::retry()` (when re-arming). Neither runs on the removal path. In particular:

- When the timer fires, `InlineTimerTask::runTimerTask()` is simply `{ _function(); }` — it does **not** clear `_function`, so the lambda (and its captured `self`) outlives the call.
- The lambda calls `self->remove()` → `NodeI::removeSubscriberSession` / `removePublisherSession`, which erases the session from the node's maps and calls `SessionI::destroyImpl()`.
- `destroyImpl()` marks the session destroyed but **never cancels or nulls `_retryTask`**.

After the timer drops its own reference (the task is one-shot), the only surviving strong reference to the `SessionI` is the `self` captured inside its own `_retryTask`'s lambda:

```
SessionI::_retryTask  ──►  InlineTimerTask  ──►  std::function lambda  ──►  self (the same SessionI)
```

A `shared_ptr` self-cycle is never reclaimed by reference counting, so the `SessionI` — together with its topics and (in the reconnect variant) the captured node proxy — leaks. Its destructor and `destroyImpl`'s remaining work never run. Via the session's strong `_parent` (the `NodeI`) and `_instance` (the `Instance`) members, the **entire node and instance leak too**.

## Where it leaks

1. **Normal operation (recurring, unbounded).** An endpoint-less peer disconnects and does not reconnect within the retry window. The timer fires `remove()` → `destroyImpl()`, the session is dropped from the maps, and the self-cycle leaks it. This happens for **every** endpoint-less peer that disconnects and stays gone, so a long-running node accumulates one leaked session (plus the topics/proxy it pins) per such event, without bound.

2. **Node destroyed while a session is mid-retry (one-time).** `NodeI::destroy()` clears the session maps **without** calling `destroyImpl()`, so a session whose retry timer is still pending at shutdown is dropped from the maps but its `_retryTask` is never cancelled — the same self-cycle leaks it. (The timer's worker thread is joined before the maps are cleared, so there is no use-after-free here, only a leak.)

## When it triggers

The endpoint-less branch always re-arms `_retryTask` and the removal path never clears it, so case (1) requires only:
1. A peer with **no public endpoint**.
2. Its session disconnects.
3. It does not reconnect within the retry window, so the removal timer fires.

These are routine conditions for a node serving transient endpoint-less clients.

## Suggested fix

Cancel and null `_retryTask` wherever a session is torn down, mirroring what `connected()` / `retry()` already do:
- In `SessionI::destroyImpl()` — this closes the recurring normal-operation leak.
- On the `NodeI::destroy()` path, before clearing the session maps — this closes the shutdown-mid-retry case.

A regression test can hold a `weak_ptr` to the `SessionI`, arm the endpoint-less retry timer (drop the peer so `_session == nullopt` with `_retryTask` pending), let the timer fire `remove()`, and assert the `weak_ptr` has expired (LeakSanitizer can serve as a fallback signal).

## Severity

Medium — a recurring, unbounded memory leak in normal operation (no crash, no data corruption, no use-after-free), with a small internal fix.

pepone:
Fixes #5736.

## Problem

When a `SessionI` toward an endpoint-less peer disconnects, `SessionI::retry` schedules a remove/reconnect task on `_retryTask` whose lambda captures `self = shared_from_this()`. Only `connected()` and `retry()` cleared `_retryTask`; **`destroyImpl` and `NodeI::destroy` did not**.

So on the session's teardown paths the cycle `_retryTask → InlineTimerTask → lambda → self` survives, and shared_ptr refcounting can never reclaim it. The session leaks — and through its strong `_parent`/`_instance` members, so do the node and the instance.

This is not shutdown-only. It happens in **normal operation**: every endpoint-less peer that disconnects and does not reconnect makes the wait-then-remove timer fire `remove() → destroyImpl`, leaking that session. A long-running node accumulates one such leak per event. It also leaks at shutdown when `NodeI::destroy` drops a session that is still mid-retry. (There is no use-after-free — the timer worker is joined before the maps are cleared; it is purely a leak.)

## Fix

Cancel and clear `_retryTask` on both teardown paths:
- in `SessionI::destroyImpl` — covers the recurring normal-operation removal;
- via a new `SessionI::cancelRetryTask`, in `NodeI::destroy` before it clears the session maps — covers the shutdown-mid-retry case (which never calls `destroyImpl`).

## Tests

No new test. The fix only cancels a pending timer on the teardown paths. I prototyped a regression test (a single-process two-node `weak_ptr` leak probe) but dropped it as disproportionately complex for the fix. The existing DataStorm tests still pass.


cpp/src/DataStorm/SessionI.cpp
```diff
@@ -767,6 +767,16 @@ SessionI::destroyImpl(const exception_ptr& ex)
     assert(!_destroyed);
     _destroyed = true;
 
+    // Cancel and clear any pending retry task. Its lambda captures a strong reference to this session
+    // (self = shared_from_this()), so leaving _retryTask set forms a reference cycle
+    // (_retryTask -> task -> lambda -> self) that keeps the session -- and, through its _parent/_instance members,
+    // the node and instance -- alive forever. connected()/retry() clear it on the live path; this covers teardown.
```
bernardnormier: I am fond of these long comments that describe the old buggy behavior: "so leaving _retryTask ...". 

The comment's focus should be on what we do now, like: 
> Cancel and clear any pending retry task to break the reference cycle (_retryTask -> task -> lambda -> self).

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -126,6 +126,18 @@ NodeI::destroy(bool ownsCommunicator)
             }
         }
     }
+    // Cancel any pending retry task before dropping the sessions. A session that is mid-retry holds its task via
```
bernardnormier: I find the details (A session that is mid-retry...) over the top.

bernardnormier (commented):
> Add a new single-process DataStorm/sessionLeak regression test covering both the destroyImpl and NodeI::destroy teardown paths.

I don't see the new test.

pepone:
Dropped it — the single-process `weak_ptr` leak probe was disproportionately complex for a fix that just cancels a timer on the teardown paths, so I removed it and noted that in the description. Happy to add a lighter test if you'd prefer one.

bernardnormier (approved):

externl (approved):

PR #5743, opened 2026-06-26
Add DataStorm test for reconnect with the writer actively publishing

pepone:
A coverage test from the DataStorm session-reconnection audit. No production code change.

## Gap

On a session reconnect, DataStorm resumes exactly-once sample delivery to a subscriber. `SubscriberSessionI::s` does this **without an `id > lastId` guard** — it relies on the `initialized == false` gate plus the topic lock to keep stale/duplicate samples out of `s()` while a session resyncs.

The existing `reliability` test only forces reconnects **after the writer has published its whole batch and gone idle**, so the reader always resyncs from retained history. The window where a *live* `s()` sample forward overlaps the resync — i.e. the writer is still publishing while the session re-establishes — is never exercised.

## Test

`DataStorm/writerActiveReconnect`: the writer publishes a known `0..999` sequence at a steady pace; the reader reads the whole sequence and closes its session connection every 50 samples. Because the closes land at `i <= 950` while the writer runs to `999`, the writer is **guaranteed to still be live-publishing during each of the ~20 reconnects**. The reader asserts every sample arrives **exactly once and in order** — a regression in the writer-active resync window would surface as a duplicate, a gap, or an out-of-order value.

Confirmed via session trace that ~20 resume cycles occur mid-stream (the session is resumed, not recreated, and re-initializes from the writer each time), so the test genuinely exercises the previously-uncovered path rather than passing vacuously. Stable across repeated runs.

The path is correct today, so this is a regression guard. It closes a coverage hole called out in the audit (§6 item 4 / the untested "`s()` exactly-once across writer-active reconnect" path).

### DIFF
--- a/cpp/msbuild/ice.test.slnx
+++ b/cpp/msbuild/ice.test.slnx
@@ -76,6 +76,14 @@
       <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
     </Project>
   </Folder>
+  <Folder Name="/DataStorm/writerActiveReconnect/">
+    <Project Path="../test/DataStorm/writerActiveReconnect/msbuild/reader/reader.vcxproj">
+      <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
+    </Project>
+    <Project Path="../test/DataStorm/writerActiveReconnect/msbuild/writer/writer.vcxproj">
+      <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
+    </Project>
+  </Folder>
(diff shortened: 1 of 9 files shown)


pepone:
Closing. This is essentially the `reliability` test's `int` case plus a 1 ms writer sleep -- same split topology, same `getSessionConnection().close()` reconnect mechanism, same exactly-once assertion. The only intended delta (keeping the writer publishing live so an `s()` forward races the resync) is hit only probabilistically, so it adds little robust coverage over `reliability` for a path that is already believed correct. Not worth the near-duplicate test surface.

PR #5744, opened 2026-06-26
Allow at most one syslog logger per process (C++)

bernardnormier:
## Summary

`SysLoggerI` calls the process-global `openlog`/`closelog` per instance and passes its own `_prefix.c_str()` as the ident. POSIX retains the ident pointer rather than copying it. With more than one `SysLoggerI` alive — two communicators in the same process, each with `Ice.UseSyslog=1` — the second `openlog` replaces the global ident with a pointer into the second instance's `_prefix`, and destroying that instance both runs `closelog()` (closing the connection shared by survivors) and leaves the global ident dangling. The next `syslog()` from a surviving logger then dereferences freed memory (use-after-free).

`cloneWithPrefix` would produce the same situation, but it is never reached at runtime in practice.

## Fix

Enforce a single active `SysLoggerI` per process:

- The constructor claims a static flag via `claimSyslog`, throwing `InitializationException` if a `SysLoggerI` is already active.
- The destructor `closelog()`s and clears the flag under the same static `_mutex` that serializes the `syslog()` calls, so teardown can't race a concurrent construction or log call.
- `cloneWithPrefix` now throws `FeatureNotSupportedException` — syslog cannot support a per-instance ident.

This means `Ice.UseSyslog` now supports one communicator per process; a second one configured with `Ice.UseSyslog=1` fails fast with `InitializationException` instead of corrupting the shared syslog state. A changelog entry documents this, and a follow-up doc issue will clarify the `Ice.UseSyslog` property page.

Fixes zeroc-ice/ice-audit#75

### DIFF
--- a/changelog.d/cpp/syslog-single-communicator.md
+++ b/changelog.d/cpp/syslog-single-communicator.md
@@ -1,0 +1,3 @@
+- Only one Ice syslog logger can be active at a time in a process. Creating a second one (for example, a second
+  communicator with `Ice.UseSyslog=1`) while the first is still active now throws `InitializationException` instead of
+  corrupting the process-global syslog state.
--- a/cpp/src/Ice/SysLoggerI.cpp
+++ b/cpp/src/Ice/SysLoggerI.cpp
@@ -10,6 +10,12 @@ using namespace std;
 using namespace Ice;
 using namespace IceInternal;
 
+// syslog's openlog/closelog and its ident are process-global, and openlog retains the ident pointer rather than
+// copying it. We therefore allow at most one SysLoggerI alive at a time, so the ident never dangles and closelog never
+// tears down a connection another logger is still using. The constructor claims this flag; the destructor releases it.
+std::mutex Ice::SysLoggerI::_mutex;
+bool Ice::SysLoggerI::_active = false;
+
 Ice::SysLoggerI::SysLoggerI(string prefix, string_view facilityString) : _facility(0), _prefix(std::move(prefix))
 {
     if (facilityString == "LOG_KERN")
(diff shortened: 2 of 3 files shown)


externl (approved):

pepone (approved):
LGTM, the fix is correct — single syslog per process is the right constraint given `openlog` retains the ident pointer, and claim/release are properly serialized by the static `_mutex` (`claimSyslog` runs before `openlog`, so a rejected second instance corrupts nothing).

On the `cloneWithPrefix` throw: it's actually the least impactful part. Ice never calls it internally on the syslog logger — communicator init constructs `SysLoggerI` directly with the program name, the only internal `cloneWithPrefix` caller is the `LoggerAdminI` delegation, and `Service.cpp`'s is the Windows EventLog wrapper. So it only fires on an explicit `getLogger()->cloneWithPrefix(...)`, which previously created a second `SysLoggerI` → the UAF being fixed. Fail-fast with `FeatureNotSupportedException` is strictly better there.

The more impactful change is the `InitializationException` on a second communicator, and the realistic trigger is **IceBox**: each service gets its own communicator by default, and IceBox doesn't pass the process logger into it (`initData.logger` is unset), so if `Ice.UseSyslog` reaches a service communicator (via `IceBox.InheritProperties` or a service config), the second `SysLoggerI` now throws and the service fails to load. That was a latent UAF before — fail-fast-vs-UAF again — but an IceBox+syslog deployment that appears to work today could start failing at startup. Worth naming the IceBox case explicitly in the changelog and the doc follow-up (the entry just says "a second communicator"), and maybe considering whether IceBox should share the process syslog logger with service communicators so it just works.

Minor: the C++ `SysLoggerI` also backs Python/Ruby/PHP/MATLAB (they use the C++ core), so the rule and both exceptions apply to those mappings too — not only C++.

Approving.

PR #5751, opened 2026-06-26
Add DataStorm test for relay-to-relay forwarding across a connection drop

pepone:
A coverage test from the DataStorm session-reconnection audit. No production code change.

## Gap

DataStorm relays are byte forwarders — a sample id is stable end-to-end across a multi-relay chain. But the `reliability` suite only ever closes **app-side** connections; **no test closes a non-app (relay-to-relay) connection**, so the forwarding path was never exercised under a mid-chain connection drop.

## Test

`DataStorm/relayHopDrop` builds a `writer → relay1 → relay2 → reader` chain in a single process and, while a background thread publishes a paced `0..499` sequence, repeatedly drops the `relay1↔relay2` hop (via the connection Ice caches for relay2's session to relay1). The reader asserts every sample arrives **exactly once and in order**.

Traced confirmation that it's meaningful: ~10 hop drops occur **during active forwarding** (data in flight), and the relay heals transparently each time — the end-to-end reader session is never even disrupted, and no sample is lost or duplicated. A regression that dropped or duplicated an in-flight sample on a closed relay hop would fail here.

### Note on scope

This started as a "2-hop resync" test, but tracing showed a relay-to-relay drop does **not** force the reader to resync: DataStorm's relay heals transparently (the relay reconnects on its first, immediate retry and re-forwards), so the end-to-end session stays connected. The test therefore verifies relay **forwarding integrity** across hop drops rather than a session resync — which is the relevant "byte-forwarder" property the audit flagged as untested.

## H3 helper

Adds a small bounded-read helper (the audit's "H3"): `getNextUnread` wrapped in a deadline that fails with context instead of blocking forever, so a lost-sample regression surfaces as a diagnostic failure rather than a silent CI timeout. Stable across repeated runs.

### DIFF
--- a/cpp/msbuild/ice.test.slnx
+++ b/cpp/msbuild/ice.test.slnx
@@ -44,6 +44,11 @@
       <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
     </Project>
   </Folder>
+  <Folder Name="/DataStorm/relayHopDrop/">
+    <Project Path="../test/DataStorm/relayHopDrop/msbuild/writer/writer.vcxproj">
+      <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
+    </Project>
+  </Folder>
   <Folder Name="/DataStorm/relayReconnect/">
     <Project Path="../test/DataStorm/relayReconnect/msbuild/reader/reader.vcxproj">
       <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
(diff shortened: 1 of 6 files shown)


pepone:
Closing. The single-process four-node relay-chain harness is disproportionately complex, and a relay-to-relay connection drop heals transparently -- the relay reconnects on its immediate first retry and re-forwards, so the end-to-end reader session is never disrupted and no sample is lost. The test therefore only confirms already-correct behavior with no bug at stake. The useful finding (relays heal transparently across a hop drop, so a relay-hop close does not force a downstream resync) is recorded in the audit notes.

PR #5765, opened 2026-06-29
Backport the SSL test certificate setup from main

pepone:
Backports main's SSL test-certificate setup to 3.7: a single `certs/` tree generated by
openssl/keytool/faketime + Docker, replacing the scattered per-language dirs and the
`zeroc-icecertutils` `makecerts.py`.

## What changed
- **Consolidated** the `IceSSL/configuration` certificates out of the four duplicated dirs
  (`cpp/`, `csharp/`, `java/`, `java-compat/test/IceSSL/certs`) into one `certs/configuration/`.
- **Common `--protocol ssl` certs** stay at the `certs/` root with their existing filenames
  (`cacert.pem`, `server.p12`, …) so driver references change *path only*.
- **New scripts**: `makecerts.sh` (common) + `makeconfiguration.sh` (configuration), driven by
  `makecerts-docker.sh` / `Dockerfile` (openssl, keytool, faketime, Bouncy Castle). All
  certificates are regenerated by these scripts.
- **Repointing**: a new `{certsdir}` driver substitution (`scripts/Util.py`) gives the
  `IceSSL/configuration` test the consolidated path; the C++, C#, Java, java-compat and Swift
  sources use it directly instead of `<test>/../certs`.

## Notes
- Kept 3.7's filenames, subject DNs, SANs, EKUs and pinned serials (`s_rsa_ca1`=1, `c_rsa_ca1`=2),
  so the DN/SAN/EKU/serial assertions are unchanged. New key pairs change the key-derived values, so
  the subject/authority key identifiers and SHA-1 thumbprints were recomputed in the C++ and C#
  `IceSSL/configuration` tests.
- Android keeps BKS keystores (generated with the Bouncy Castle provider) — no controller changes.
- The IceCertUtils CA database under `db/` is replaced by the openssl `db/ca4` that the OCSP
  responder (`revocationutil.py`) reads; the other `db/` dirs are dropped.

## Opened to exercise CI
A few platform behaviors can't be verified locally and are the reason for this PR:
- **Schannel** (Windows) `FindCert` THUMBPRINT/SUBJECTKEYID rendering.
- **SecureTransport** (macOS) `FindCert` and the `cn3`/`cn4`/`cn8` CN-only hostname checks.
- The **OCSP** live handshake (Darwin/Windows).

Plan and implementation were both reviewed with codex.


externl (approved):
Looks good!

PR #5766, opened 2026-06-29
Propagate a located servant's dispatch error in Java ServantManager

### LINKED ISSUE #5712: [Audit-Low] Java ServantManager converts a located servant's UserException to a response in-place — error-inspecting middleware never observes the throw
Follow-up to #5555 (Swift), fixed in #5706. The same divergence exists in the Java (com.zeroc) runtime, narrowed to user exceptions.

In `ServantManager.dispatch`'s servant-locator path, the async completion handler converts a `UserException` to a response in-place instead of letting it propagate:

```java
// java/src/com.zeroc.ice/.../ServantManager.java (the response.handle(...) lambda)
if (ex instanceof Error errorEx) { throw errorEx; }
if (ex instanceof RuntimeException runtimeEx) { throw runtimeEx; }
return current.createOutgoingResponse(ex);   // UserException converted here
```

Because `ServantManager` is the innermost dispatcher, a located servant whose dispatch completes exceptionally with a `UserException` is turned into a response here, so error-inspecting middleware never observes the throw — unlike the plain-servant path (which returns the stage as-is) and the C#/JS runtimes. `Error`/`RuntimeException` and synchronous throws already propagate (re-thrown after `finished`), so the divergence is limited to async `UserException` completions.

Low priority: `ServantLocator` + error-inspecting middleware is a rare combination, and there is no wire-format change for clients either way.

pepone:
Fixes #5712. Java (com.zeroc) follow-up to the Swift fix #5706 (for #5555).

## Problem

In `ServantManager.dispatch`'s servant-locator path, the async completion handler converted a `UserException` that completed a located servant's dispatch into a response **in-place**:

```java
if (ex instanceof Error errorEx) { throw errorEx; }
if (ex instanceof RuntimeException runtimeEx) { throw runtimeEx; }
return current.createOutgoingResponse(ex);   // UserException converted here
```

Because `ServantManager` is the innermost dispatcher, error-inspecting middleware never observed the throw — unlike the **plain-servant path** (line 25, which returns `servant.dispatch(request)` as-is) and the C#/JS runtimes. `Error`/`RuntimeException` and synchronous throws already propagated, so the divergence was limited to async `UserException` completions.

## Fix

Propagate the user exception instead of converting it:

```java
throw new CompletionException(ex);
```

The lambda can't throw a checked `UserException`, so it's wrapped in `CompletionException`. The higher-level consumer (`ConnectionI`, `response.whenComplete(... createOutgoingResponse(exception) ...)`) marshals the propagated exception, and `Current.createOutgoingResponse(Throwable)` **unwraps** `CompletionException` before marshalling — so the client receives the **same wire response** as before, while middleware between the connection and `ServantManager` now observes the dispatch error. This matches the plain-servant path.

## Test

New `test/Ice/middleware` case: a request is routed through a servant locator whose located servant fails its dispatch with a user exception (a custom dispatcher returning `CompletableFuture.failedFuture(...)` — a generated AMD servant would wrap the exception in `CompletionException`, which already propagates), with an error-observing middleware installed. It asserts the client sees the dispatch error **and** the middleware observed the throw.

Confirmed red→green: without the fix, the client still errors but the middleware no longer observes the throw (`ServantManager` converts the exception to a response). `:ice` + `:test` compile; checkstyle and OpenRewrite are clean.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ServantManager.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ServantManager.java
@@ -4,6 +4,7 @@ package com.zeroc.Ice;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 
 /** A dispatcher that manages servants and servant locators for an object adapter. */
@@ -70,15 +71,17 @@ final class ServantManager implements Object {
                         ex = finishedEx;
                     }
                     if (ex != null) {
-                        // We only marshal errors and runtime exceptions
-                        // (including CompletionException) at a higher level.
+                        // Propagate the dispatch error like the plain-servant path so error-inspecting
+                        // middleware observes it (instead of converting it to a response here). Errors and
+                        // runtime exceptions propagate directly; the remaining (checked) user exception is
+                        // wrapped so it can flow through the completion stage.
                         if (ex instanceof Error errorEx) {
                             throw errorEx;
                         }
                         if (ex instanceof RuntimeException runtimeEx) {
                             throw runtimeEx;
                         }
-                        return current.createOutgoingResponse(ex);
+                        throw new CompletionException(ex);
                     } else {
                         return r;
                     }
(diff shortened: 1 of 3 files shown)


bernardnormier (approved):
Looks good.

Keep in mind that servant locators are in effect deprecated, and we don't document them at https://docs.zeroc.com/ice/3.8/java/object-adapters. 

So we don't want to spend too much time on fine-tuning the servant locator / middleware interactions.

externl (approved):

PR #5767, opened 2026-06-29
Validate ThreadPriority properties (C#, Java)

bernardnormier:
A `ThreadPriority` property set to an unrecognized value was silently ignored and fell back to the normal priority. This PR rejects such values with a `PropertyException` instead, in both C# and Java.

### Java
- Unified the thread pool priority path with the existing `Ice.ThreadPriority` path: both now go through `Util.getThreadPriorityProperty` rather than the thread pool parsing the value as a plain integer via `getPropertyAsInt`.
- As a result, **all** thread priority properties (`Ice.ThreadPriority`, `Ice.ThreadPool.{Client,Server}.ThreadPriority`, and an object adapter's `<adapter>.ThreadPool.ThreadPriority`) share one validated code path:
  - unrecognized values throw `PropertyException`;
  - integer values are validated against the allowed range (1-10);
  - thread pool priorities now also accept the `MIN_PRIORITY`/`NORM_PRIORITY`/`MAX_PRIORITY` constant names, which previously only `Ice.ThreadPriority` accepted.
- Dropped the now-redundant `_hasPriority` flag (the helper always yields a valid priority, defaulting to `NORM_PRIORITY`), so the thread pool and the timer/endpoint-resolver threads now always set their priority explicitly — matching C#.

### C#
- Reworked `Util.stringToThreadPriority` into `getThreadPriorityProperty(Ice.Properties, string prefix)`, which reads the property value itself instead of having each caller fetch it, mirroring the Java helper. Unrecognized values now throw `PropertyException`.

### Behavioral note
The Java timer/endpoint-resolver and thread-pool threads now always call `setPriority` (to `NORM_PRIORITY` when unconfigured) instead of inheriting the creating thread's priority. In practice this is the same value (5) and matches long-standing C# behavior.

### Tests
Updated the C# and Java `threadPoolPriority` tests to cover the new validation (invalid string in C#, out-of-range integer in Java). Both pass.

### DIFF
--- a/java/test/src/main/java/test/Ice/threadPoolPriority/Client.java
+++ b/java/test/src/main/java/test/Ice/threadPoolPriority/Client.java
@@ -4,6 +4,8 @@ package test.Ice.threadPoolPriority;
 
 import com.zeroc.Ice.Communicator;
 import com.zeroc.Ice.ObjectPrx;
+import com.zeroc.Ice.Properties;
+import com.zeroc.Ice.PropertyException;
 
 import test.Ice.threadPoolPriority.Test.PriorityPrx;
 import test.TestHelper;
@@ -24,5 +26,16 @@ public class Client extends TestHelper {
             out.println("ok");
             priority.shutdown();
         }
+
+        out.print("testing invalid thread priority... ");
+        out.flush();
+        Properties properties = new Properties();
+        properties.setProperty("Ice.ThreadPool.Client.ThreadPriority", "11");
+        try (Communicator communicator = initialize(properties)) {
+            test(false);
+        } catch (PropertyException ex) {
+            // expected: 11 is outside the allowed range [1, 10]
+        }
+        out.println("ok");
     }
 }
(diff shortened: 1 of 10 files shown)


java/test/src/main/java/test/Ice/threadPoolPriority/Client.java
```diff
@@ -24,5 +26,16 @@ public void run(String[] args) {
             out.println("ok");
             priority.shutdown();
         }
+
+        out.print("testing invalid thread priority... ");
+        out.flush();
+        Properties properties = new Properties();
```
pepone, line 32: Maybe we can do `MAX_PRIORITY + 1`

pepone (approved):

externl (approved):

java/test/src/main/java/test/Ice/threadPoolPriority/Client.java
bernardnormier: Fixed as suggested

PR #5773, opened 2026-06-29
Fix latent UB in C++ WebSocket unmask offset computation

bernardnormier:
The C++ WebSocket transceiver computed the per-frame unmask offset by subtracting two iterators into the read buffer:

```cpp
for (auto n = static_cast<size_t>(_readStart - _readFrameStart); p < buf.i; ++p, ++n)
    *p ^= _readMask[n % 4];
```

`_readFrameStart` is captured at the frame's payload start, while `ConnectionI` reads the 14-byte Ice message header. `ConnectionI` then `resize`s `_readStream` to the full message size — a `::realloc` in `Buffer.cpp` — before reading the body, freeing the old allocation. For the first frame of any message larger than the read buffer's capacity (~240 bytes), `postRead` then subtracts pointers into two different allocations, one of them freed.

This is **latent undefined behavior**. On mainstream allocators it still produced the correct mask index — only `n % 4` is used, and since `malloc`/`realloc` align to ≥ 4 bytes the spurious `(newBuf - oldBuf)` term (and any `size_t` wraparound) vanishes mod 4 — so no incorrect bytes were ever observed in practice. But subtracting pointers across distinct allocations (and reading an iterator into a freed block) is UB by the standard, and breaks under pointer-provenance tooling (`-fsanitize=pointer-subtract`) and provenance-strict targets such as CHERI.

The fix tracks the mask offset as an integer (`_readFrameOffset`, the number of payload bytes consumed so far in the frame) rather than an iterator into a reallocatable buffer, mirroring the C# port. No cross-allocation pointer subtraction remains. Continuation-frame behavior is unchanged: the offset resets to 0 at each frame start, exactly as `_readFrameStart` did.

Tested with `Ice/operations` over the `ws` protocol (client/server, AMD, collocated) — all pass.

No CHANGELOG entry: no observable behavior change on supported platforms; this removes latent UB only.

### DIFF
--- a/cpp/src/Ice/WSTransceiver.cpp
+++ b/cpp/src/Ice/WSTransceiver.cpp
@@ -1365,7 +1365,7 @@ IceInternal::WSTransceiver::preRead(Buffer& buf)
                     }
                     _readState = ReadStatePayload;
                     assert(buf.i != buf.b.end());
-                    _readFrameStart = buf.i;
+                    _readFrameOffset = 0;
                     break;
                 }
                 case OP_CLOSE: // Connection close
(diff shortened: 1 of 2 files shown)


pepone (approved):
LGTM — nice catch on the latent UB.

The integer `_readFrameOffset` is equivalent to the old `_readStart - _readFrameStart` (payload bytes consumed so far in the frame) and removes the cross-allocation pointer subtraction. Confirmed all three `_readFrameStart` uses are covered, the early `_readStart == buf.i` return keeps the offset from drifting, and per-frame masking is preserved.

Optional nit: now that it's a plain `size_t`, you could initialize `_readFrameOffset(0)` in the constructor init list alongside `_readHeaderLength`/`_readPayloadLength` — it's always set in `preRead` before use, so not required.

bernardnormier:
> Optional nit: now that it's a plain `size_t`, you could initialize `_readFrameOffset(0)` in the constructor init list alongside `_readHeaderLength`/`_readPayloadLength` — it's always set in `preRead` before use, so not required.

Fixed, except I initialized in in the header, not in the constructors. We should replace all these ctor initializations to 0/false etc. Maybe for 3.9?

PR #5776, opened 2026-06-30
Add OOM null checks on IcePy dispatch/decode paths

pepone:
Several CPython allocations on IcePy decode/dispatch paths were used without a null check. Under memory pressure a failed allocation (`NULL` with a pending `MemoryError`) became a null dereference or, worse, silently wrong object state — and these paths run while processing incoming traffic.

Each fix null-checks the allocation and aborts cleanly, leaving the pending exception set, mirroring the neighboring already-checked allocations in the same files.

### `Types.cpp`
- `PrimitiveInfo::unmarshal` — each primitive case (`PyLong_FromLong` / `PyLong_FromLongLong` / `PyFloat_FromDouble` / `createString`) passed its result to `cb->unmarshaled(p.get(), ...)` without a null check. A `NULL` to `DataMember::unmarshaled` becomes `PyObject_SetAttrString(target, name, nullptr)`, which CPython interprets as *delete the attribute* — masking the OOM as wrong object state. Now checked, mirroring `StructInfo::unmarshal`.
- `ProxyInfo::unmarshal` and `SequenceInfo::createSequenceFromMemory` had the same unguarded pattern (`createProxy` / `PyTuple_New(2)`).

### `Current.cpp`
- `createCurrent` did not null-check `PyTuple_New(9)` or the context `PyDict_New()` before using them; both run on every incoming dispatch. Now return `nullptr`, consistent with the function's other allocation checks.

### `Operation.cpp`
- `TypedUpcall::dispatch` and `BlobjectUpcall::dispatch` stole `createCurrent`'s result into the argument tuple without checking it for `NULL`. Now checked (`throwPythonException`), so an allocation failure in `createCurrent` aborts the dispatch instead of placing `NULL` into the tuple.

### `PropertiesAdmin.cpp`
- The property-update callback skipped its populate block on `PyDict_New` failure and then placed the `NULL` `result` into the call tuple; the per-entry key and the call tuple itself were also unchecked. Now bails on `PyDict_New` failure and null-checks the key and the tuple. `result` is released into the tuple only after the tuple is confirmed non-null.

These are OOM-only triggers (no behavior change on the success path); verified from the CPython C-API control flow.

### DIFF
--- a/python/modules/IcePy/Current.cpp
+++ b/python/modules/IcePy/Current.cpp
@@ -15,6 +15,10 @@ IcePy::createCurrent(const Ice::Current& current)
     PyObject* currentType{lookupType("Ice.Current")};
 
     PyObjectHandle args{PyTuple_New(9)};
+    if (!args.get())
+    {
+        return nullptr;
+    }
 
     PyObject* adapter{wrapObjectAdapter(current.adapter)};
     if (!adapter)
@@ -78,6 +82,10 @@ IcePy::createCurrent(const Ice::Current& current)
     PyTuple_SetItem(args.get(), 5, getAttr(operationModeType, enumerator, false));
 
     PyObjectHandle ctx{PyDict_New()};
+    if (!ctx.get())
+    {
+        return nullptr;
+    }
     if (!contextToDictionary(current.ctx, ctx.get()))
     {
         return nullptr;
--- a/python/modules/IcePy/Operation.cpp
+++ b/python/modules/IcePy/Operation.cpp
@@ -2031,6 +2031,10 @@ IcePy::TypedUpcall::dispatch(PyObject* servant, pair<const byte*, const byte*> i
     // Create an object to represent Ice::Current. We need to append this to the argument tuple.
     //
     PyObjectHandle curr{createCurrent(current)};
+    if (!curr.get())
+    {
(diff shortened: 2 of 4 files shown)


externl (approved):
Looks good!

bernardnormier (approved):
Solid defensive hardening, and the per-file choice of unwind mechanism is right: `AbortMarshaling` on the `Types.cpp` unmarshal paths (caught and converted at the `Operation.cpp` boundary) and `throwPythonException()` for the `createCurrent` failures in `dispatch`, which run outside that handler. The `DataMember::unmarshaled` analysis is accurate — an unchecked `NULL` becomes `PyObject_SetAttrString(target, name, nullptr)`, i.e. silent attribute deletion / wrong object state, so the explicit check is the correct fix. The `PropertiesAdmin` change also closes a real pre-existing gap (the key was previously unchecked). No changelog is right for OOM-only hardening.

One coordination note: this overlaps #5780 in `nativePropertiesAdminAddUpdateCB`. Plan is to merge this PR first; #5780 then rebases on top and reworks that callback's error model — including the OOM paths here that currently `return` with a pending Python exception set while invoked from the C++ runtime, which is the same concern #5780 addresses. Approving on that basis.

PR #5779, opened 2026-06-30
Validate operation mode in incoming requests

pepone:
The operation mode is a one-byte field in the Ice request header, with only three valid values: `Normal` (0), `Nonmutating` (1, deprecated), and `Idempotent` (2). The server runtime previously accepted any byte value without validation:

- C++ and C# cast the byte straight to the `OperationMode` enum, producing an out-of-range enum value.
- Java and JS mapped an unknown byte to `null`/`undefined`.

This change validates the operation mode where the request header is unmarshaled and rejects an out-of-range value with a `MarshalException`, mirroring the existing facet-path validation in the same code. The C++ change also covers the Python, PHP, Ruby, MATLAB, and Swift mappings, which dispatch through the C++ core.

Valid values are unchanged — the deprecated `Nonmutating` (1) is still accepted; only values `>= 3` are rejected.

No test: triggering this requires a peer that writes a malformed request header, since an out-of-range mode byte cannot be produced through the public API (it would need raw-protocol fault injection).

### DIFF
--- a/changelog.d/general/incoming-request-invalid-operation-mode.md
+++ b/changelog.d/general/incoming-request-invalid-operation-mode.md
@@ -1,2 +1,0 @@
-- The Ice server runtime now validates the operation mode of an incoming request and rejects a request with an invalid
-  operation mode with a `MarshalException`.
--- a/changelog.d/general/incoming-request-invalid-operation-mode.md
+++ b/changelog.d/general/incoming-request-invalid-operation-mode.md
@@ -1,0 +1,2 @@
+- The Ice server runtime now validates the operation mode of an incoming request and rejects a request with an invalid
+  operation mode with a `MarshalException`.
--- a/cpp/src/Ice/IncomingRequest.cpp
+++ b/cpp/src/Ice/IncomingRequest.cpp
@@ -38,13 +38,7 @@ IncomingRequest::IncomingRequest(
 
     inputStream.read(_current.operation, false);
 
-    uint8_t mode;
-    inputStream.read(mode);
-    if (mode > static_cast<uint8_t>(OperationMode::Idempotent))
-    {
-        throw MarshalException{__FILE__, __LINE__, "received invalid operation mode"};
-    }
-    _current.mode = static_cast<OperationMode>(mode);
+    inputStream.read(_current.mode);
 
     int32_t sz = inputStream.readSize();
     while (sz--)
--- a/cpp/src/Ice/IncomingRequest.cpp
+++ b/cpp/src/Ice/IncomingRequest.cpp
@@ -40,6 +40,10 @@ IncomingRequest::IncomingRequest(
 
     uint8_t mode;
     inputStream.read(mode);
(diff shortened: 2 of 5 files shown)


externl (approved):

bernardnormier (commented):
The validation itself is correct and consistent across the four mappings (`> Idempotent` / `valueOf == null`), and `MarshalException` matches the facet-path check right above it. Two points before merging.

**1. Prefer the generated enum unmarshaling over a hand-rolled check in four languages.**

`OperationMode` is a Slice enum, and its generated stream unmarshaling already performs exactly this range check, in every mapping:

- C++ `StreamHelper<T, StreamHelperCategoryEnum>::read` → `readEnum(maxValue)` then rejects out-of-range with `MarshalException("enumerator out of range")` (`minValue=0, maxValue=2` for `OperationMode`).
- Java `OperationMode.validate(int)` / `ice_read` → throws `MarshalException("enumerator value N is out of range")` when `valueOf` returns null.
- C# / JS generate the same `validate` / `ice_read` pattern.

So instead of re-implementing the boundary by hand four times (which can drift between mappings), the header reader could route the mode through the generated unmarshaling — e.g. `inputStream.read(_current.mode)` in C++, `OperationMode.ice_read(istr)` in the others — and get the validation for free, from a single source of truth.

This is wire-compatible. The request/reply header is encoded with the 1.0 encoding, and under 1.0 an enum with fewer than 128 enumerators is encoded as a single byte equal to the value — identical to the raw `uint8_t` the senders write (`OutgoingAsync` `_os.write(static_cast<uint8_t>(_mode))`, `Reference` `s->write(static_cast<uint8_t>(_mode))`). So `readEnum`/`writeEnum` on the 1.0 header stream produce/consume exactly the same byte as the current hand-marshaling.

The cleanest version routes the mode through the generated enum marshaling on **both** sides (reader and writer), which removes the manual `static_cast` everywhere and keeps the two sides symmetric on one validated codec.

If you'd rather keep the change minimal, the current per-language check is acceptable as-is; this is a preference toward not duplicating logic the generated code already owns.

**2. Drop the changelog fragment.**

This is a silent-accept -> reject of malformed input, non-crash, only reachable by a peer that writes a bad request header (no public API can produce an out-of-range mode byte). That's hardening against a misbehaving peer rather than a user-facing behavior change, so it shouldn't get a CHANGELOG entry. Please remove `changelog.d/general/incoming-request-invalid-operation-mode.md`.

pepone:
Addressed in 510c486:

**1.** Routed the mode through the generated enum unmarshaling — `inputStream.read(_current.mode)` (C++), `OperationMode.ice_read` (Java), `OperationMode._read` (JS).

C# is the exception: its generated enum read is `(OperationMode)readEnum(2)` with no range check (and `readEnum` itself doesn't validate), so routing C# through it would silently re-accept out-of-range modes. Kept the explicit check there and filed #5785 for the broader gap — C# doesn't range-validate any Slice enum on unmarshal.

**2.** Dropped the changelog fragment.

bernardnormier (approved):

PR #5780, opened 2026-06-30
Fix IcePy properties-admin update callback to not throw into the Ice runtime

### LINKED ISSUE #5777: IcePy: NativePropertiesAdmin update callback throws AbortMarshaling out of the C++ runtime
### Summary

In the IcePy extension, the property-update notification callback registered by `Ice.NativePropertiesAdmin.addUpdateCallback` throws `AbortMarshaling` when the user's Python callback raises:

https://github.com/zeroc-ice/ice/blob/a97a2b8e57396b4aa707c508b26e46df9e1f9cb9/python/modules/IcePy/PropertiesAdmin.cpp#L100-L105

```cpp
PyObjectHandle obj{PyObject_Call(callback, callbackArgs.get(), nullptr)};
if (!obj.get())
{
    assert(PyErr_Occurred());
    throw AbortMarshaling();   // <-- escapes into the C++ Ice runtime
}
```

This is the **only** inconsistent use of `AbortMarshaling` in the extension (see sweep below), and it's wrong for this context.

### Why this is incorrect

`AbortMarshaling` (`Types.h`, `class AbortMarshaling {};` — an empty marker, **not** a `std::exception`) is the extension's internal signal for "a Python error occurred mid-(un)marshaling; unwind to the dispatch boundary." It is only ever caught by the 8 `catch (const AbortMarshaling&)` handlers in `Operation.cpp`, which wrap request marshaling/unmarshaling and clear/convert the pending `PyErr`.

This callback is **not** on a marshaling path. It is invoked by the C++ runtime from `Ice::NativePropertiesAdmin::setProperties`:

https://github.com/zeroc-ice/ice/blob/a97a2b8e57396b4aa707c508b26e46df9e1f9cb9/cpp/src/Ice/NativePropertiesAdmin.cpp#L163-L168

```cpp
auto callbacks = _updateCallbacks;
lock.unlock();
for (const auto& cb : callbacks)
{
    cb(changes);   // no try/catch
}
```

So when the Python callback raises:

1. `AbortMarshaling` propagates out of `setProperties` — there is **no** `AbortMarshaling` handler anywhere on this path. It aborts the remaining update callbacks and unwinds to the connection dispatch's generic `catch (...)`, which turns it into an `UnknownException` returned to the admin caller.
2. The pending Python exception set by `PyObject_Call` is **never cleared** — it leaks past the `AdoptThread` scope and can surface spuriously on the next Python API call on that thread.
3. It is inconsistent with the callback's own sibling failure paths, which all `return;` (e.g. `PyDict_New` failure). Only the `PyObject_Call` failure throws.

A property-update notification is best-effort: the property change has already been applied by the time callbacks run, and the C++ `std::function` callback contract here is "don't throw." A failing user callback should not abort the admin dispatch or the other callbacks.

### Recommended fix

Make the `PyObject_Call` failure path consistent with the others — do **not** throw `AbortMarshaling` out of a runtime-invoked callback. Clear the pending Python error and `return;`. Optionally surface it first (e.g. `PyErr_Print()`, or route through the communicator logger) so a buggy callback isn't silently swallowed, rather than dropping it as the other `return;` paths currently do.

### Sweep: other `AbortMarshaling` uses in IcePy

I checked every `throw AbortMarshaling()` in `python/modules/IcePy/`; this is the only misuse. All others are on the marshaling/unmarshaling path reached from an `Operation.cpp` `catch (const AbortMarshaling&)`:

| Site | Count | Context | Status |
|------|-------|---------|--------|
| `Types.cpp` | 108 | (un)marshal of primitives/proxies/sequences/values, driven by the `Ice::InputStream`/`OutputStream` that `Operation.cpp` constructs inside its try/catch | OK |
| `Operation.cpp` | 1 | inside the dispatch path itself | OK |
| `DefaultSliceLoader.cpp`, `PySliceLoader.cpp` | 1 each | `SliceLoader::newInstance`, invoked synchronously during value unmarshaling (same `InputStream` chain) | OK |
| **`PropertiesAdmin.cpp`** | **1** | **runtime-invoked update callback — not a marshaling path** | **this issue** |

The other files that register C++ callbacks with the runtime (`Connection.cpp`, `Communicator.cpp`, `ObjectAdapter.cpp`, `Proxy.cpp`) do not throw `AbortMarshaling`.

### Notes

- Pre-existing on `main`; not introduced by #5776 (that PR only tightened the null-checks/`return;` paths in the same lambda and left the throw in place).
- No memory-safety / remote-trigger angle: the trigger is a user-supplied Python update callback raising during an admin `setProperties` dispatch. This is an error-handling/robustness cleanup.

pepone:
The property-update callback registered with `NativePropertiesAdmin.addUpdateCallback` is invoked by the Ice runtime from `NativePropertiesAdmin::setProperties`. Previously, when the user's Python callback raised, the IcePy callback threw `AbortMarshaling` — the extension's internal marshaling-unwind sentinel, which has no handler on this path: it escaped into the connection dispatch's generic `catch (...)` and left a pending Python exception set.

This makes every failure path in the callback call `throwPythonException()`, which captures and clears the pending Python error and converts it to a C++ exception that unwinds out of `setProperties` — the same behavior as a throwing update callback in the C++, C#, and Java mappings (which propagate, aborting the dispatch and skipping the remaining callbacks). The OOM `return;` paths added in #5776, which also left a pending Python error set while invoked from the runtime, now propagate the same way.

Also drops the now-unused `Types.h` include.

No changelog: externally visible behavior is unchanged (callers already received an `UnknownException`); the fix is internal — stop throwing the internal sentinel, and stop leaking a pending Python error.

Closes #5777

### DIFF
--- a/python/modules/IcePy/PropertiesAdmin.cpp
+++ b/python/modules/IcePy/PropertiesAdmin.cpp
@@ -107,8 +107,7 @@ nativePropertiesAdminAddUpdateCB(NativePropertiesAdminObject* self, PyObject* ar
                     if (!obj.get())
                     {
                         assert(PyErr_Occurred());
-                        // The update callback raised. Propagate the exception out of setProperties (aborting the
-                        // dispatch and skipping any remaining callbacks), matching the C++, C#, and Java mappings.
+                        // The update callback raised.
                         throwPythonException();
                     }
                 });
--- a/python/modules/IcePy/PropertiesAdmin.cpp
+++ b/python/modules/IcePy/PropertiesAdmin.cpp
@@ -3,7 +3,6 @@
 #include "PropertiesAdmin.h"
 #include "Ice/DisableWarnings.h"
 #include "Thread.h"
-#include "Types.h"
 #include "Util.h"
 
 #include <algorithm>
@@ -83,7 +82,7 @@ nativePropertiesAdminAddUpdateCB(NativePropertiesAdminObject* self, PyObject* ar
                     PyObjectHandle result{PyDict_New()};
                     if (!result.get())
                     {
-                        return;
+                        throwPythonException();
                     }
 
                     for (const auto& [key, value] : dict)
@@ -93,14 +92,14 @@ nativePropertiesAdminAddUpdateCB(NativePropertiesAdminObject* self, PyObject* ar
                         if (!pyKey.get() || !pyValue.get() ||
                             PyDict_SetItem(result.get(), pyKey.get(), pyValue.get()) < 0)
                         {
-                            return;
+                            throwPythonException();
                         }
                     }
 
                     PyObjectHandle callbackArgs{PyTuple_New(1)};
                     if (!callbackArgs.get())
                     {
-                        return;
+                        throwPythonException();
                     }
                     PyTuple_SetItem(callbackArgs.get(), 0, result.release());
 
@@ -108,7 +107,9 @@ nativePropertiesAdminAddUpdateCB(NativePropertiesAdminObject* self, PyObject* ar
                     if (!obj.get())
                     {
                         assert(PyErr_Occurred());
-                        throw AbortMarshaling();
+                        // The update callback raised. Propagate the exception out of setProperties (aborting the
+                        // dispatch and skipping any remaining callbacks), matching the C++, C#, and Java mappings.
+                        throwPythonException();
                     }
                 });
 


externl (approved):

bernardnormier (changes requested):
Removing the `throw AbortMarshaling()` is right — `AbortMarshaling` is the extension's internal marshaling-unwind signal and has no business in a properties-admin callback. But the replacement (`PyErr_PrintEx` + continue) takes Python in the wrong direction: it makes Python the only mapping that *tolerates* a throwing update callback, and the changelog entry then advertises that divergence.

All three reference mappings share one contract — an update callback must not throw; if it does, it propagates out of `setProperties`, aborting the dispatch and skipping the remaining callbacks:

- C++ `NativePropertiesAdmin::setProperties`: `for (const auto& cb : callbacks) cb(changes);` — no catch.
- C# `NativePropertiesAdmin.setProperties`: same loop, with an explicit `// The callback should not throw any exception.`
- Java `NativePropertiesAdmin.setProperties`: `for (var callback : callbacks) callback.accept(changes);` — no catch.

The `__aexit__` precedent cited in the description isn't equivalent: that's a future-completion continuation with nowhere to propagate. This callback runs synchronously inside the `setProperties` dispatch — exactly where the other mappings do propagate.

Suggested change — propagate instead of swallow, so Python matches the others:

```cpp
if (!obj.get())
{
    assert(PyErr_Occurred());
    // Propagate like a throwing callback in the other mappings (aborts the setProperties
    // dispatch) rather than throwing the extension's internal AbortMarshaling.
    throwPythonException();
}
```

`throwPythonException()` captures-and-clears the pending Python error (`PyErr_GetRaisedException()`) and converts it to the corresponding C++ exception, which then unwinds through `setProperties` to the dispatch boundary — same result as a throwing C++/C#/Java callback, and no leaked pending exception.

With that change the externally visible behavior barely moves (callers already got `UnknownException`); the real fix is internal (stop throwing the internal sentinel, stop leaking a Python error). I'd drop the changelog fragment rather than keep the current "we tolerate throwing callbacks" wording.

pepone:
You're right — propagating matches the other mappings; `PyErr_PrintEx` + continue made Python the odd one out. Reworked in 58d867a: every failure path in the callback now `throwPythonException()` — the raised-callback case **and** the OOM `return` paths from #5776 — so they all unwind out of `setProperties` like a throwing update callback in C++/C#/Java. Dropped the changelog and the now-unused `Types.h` include. Rebased onto main (post-#5776).

python/modules/IcePy/PropertiesAdmin.cpp
```diff
@@ -93,22 +92,24 @@ nativePropertiesAdminAddUpdateCB(NativePropertiesAdminObject* self, PyObject* ar
...
                     PyTuple_SetItem(callbackArgs.get(), 0, result.release());
 
                     PyObjectHandle obj{PyObject_Call(callback, callbackArgs.get(), nullptr)};
                     if (!obj.get())
                     {
                         assert(PyErr_Occurred());
-                        throw AbortMarshaling();
+                        // The update callback raised. Propagate the exception out of setProperties (aborting the
+                        // dispatch and skipping any remaining callbacks), matching the C++, C#, and Java mappings.
+                        throwPythonException();
```
pepone:
```suggestion
                        // The update callback raised.
                        throwPythonException();
```

bernardnormier (approved):
Looks good!

PR #5782, opened 2026-06-30
Fix DataStorm DiscardPolicy::Priority crash for filtered/any-key readers

pepone:
A DataStorm reader configured with `DiscardPolicy::Priority` compared a sample's priority against `_connectedKeys[sample->key].back()->priority`. `_connectedKeys` maps a key to the connected publishers' subscriber list (sorted by priority).

A filtered or any-key reader attached to an any-key writer is registered under the **null** key, but its samples carry a real marshaled key. So `_connectedKeys[sample->key]` looked up the real key, found no entry, and `operator[]` default-constructed an **empty** vector — making `.back()` undefined behavior (typically a crash). This is a supported reader configuration, so no malformed input is required; it affects both `queue` (live samples) and `initSamples`.

The fix resolves the priority threshold via a small helper (`hasLowerPriorityThanConnected`) that:
- never calls `.back()` on an empty list (uses `find`, not `operator[]`, and checks `empty()`), and
- compares the sample's priority against the highest-priority connected publisher across **both** the sample's key bucket and the null-key bucket — an any-key/filter peer (null key) delivers every key, so it is a valid publisher for the sample's key. The sample is discarded only when some connected publisher has a strictly higher priority.

Behavior:
- **Keyed readers** have only their own-key bucket (no null entry), so the decision is unchanged.
- **Filter/any-key readers** no longer crash when the sample's key has no own bucket, and when such a reader is connected to both a keyed writer and an any-key writer for the same key, the threshold now correctly accounts for both.

No test: reproducing this needs a filtered/any-key reader with `DiscardPolicy::Priority` receiving samples from an any-key writer, which requires a coordinated multi-process reader/writer scenario in the DataStorm test harness — disproportionate for a guarded lookup. The change is verified by compilation and review.

### DIFF
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -734,25 +734,6 @@ DataReaderI::getNextUnread()
     return sample;
 }
 
-bool
-DataReaderI::hasLowerPriorityThanConnected(int priority, const shared_ptr<Key>& key) const
-{
-    // A sample with this key can come from a connected publisher registered under the key itself (a keyed peer) or
-    // under the null key (a filter or any-key peer, which delivers every key). Discard the sample when its priority is
-    // below the highest priority among all of them. Each list is sorted by ascending priority in addConnectedKey, so
-    // back() is the highest priority; don't discard when there are no connected publishers to compare against.
-    int threshold = priority;
-    for (const auto& k : {key, shared_ptr<Key>{}})
-    {
-        auto p = _connectedKeys.find(k);
-        if (p != _connectedKeys.end() && !p->second.empty())
-        {
-            threshold = std::max(threshold, p->second.back()->priority);
-        }
-    }
-    return priority < threshold;
-}
-
 void
 DataReaderI::initSamples(
     const vector<shared_ptr<Sample>>& samples,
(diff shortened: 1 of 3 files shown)


cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -734,6 +734,25 @@ DataReaderI::getNextUnread()
     return sample;
 }
 
+bool
```
bernardnormier: I would move this implementation after addConnectedKey, to follow the declaration order.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -734,6 +734,25 @@ DataReaderI::getNextUnread()
...
+DataReaderI::hasLowerPriorityThanConnected(int priority, const shared_ptr<Key>& key) const
+{
+    // A sample with this key can come from a connected publisher registered under the key itself (a keyed peer) or
+    // under the null key (a filter or any-key peer, which delivers every key). Discard the sample when its priority is
+    // below the highest priority among all of them. Each list is sorted by ascending priority in addConnectedKey, so
```
bernardnormier: I don't like this comment: this function doesn't discard anything. The discarding is performed by the caller.

bernardnormier (approved):

externl (approved):

PR #5784, opened 2026-06-30
Use IceSSL.Password when loading an encrypted PEM key on OpenSSL

pepone:
The OpenSSL engine defines a password callback (`Ice_SSL_opensslPasswordCallback`) and reads `IceSSL.Password`, but it never registered the callback on the SSL context. As a result `IceSSL.Password` was ignored when loading an **encrypted PEM private key**: OpenSSL fell back to prompting for the password on the terminal (blocking the process), or failing with a confusing error when no terminal is attached. Only the PKCS12 path consumed the password.

This registers the password callback on the context before loading the certificate and key, so `IceSSL.Password` is honored for encrypted PEM keys.

It also adds a regression test (`testEncryptedKey`, OpenSSL only) that configures a server with an encrypted PEM key + `IceSSL.Password` and verifies it loads, along with the `certs/configuration/ca1/server_encrypted_key.pem` it uses (regenerated by `makecerts.sh`). This coverage gap is why the bug went unnoticed: the only password-protected test material was PKCS12 (which uses the password directly), while every `_key.pem` was unencrypted.

Verified by building and running `IceSSL/configuration` in the Linux/OpenSSL CI builder image: the new test passes with the fix and blocks/fails without it.


bernardnormier (approved):
Good fix.

externl (approved):

Issue #4872, opened 2026-01-10
csharp/Ice/metrics test failure (Windows)


bernardnormier:
From https://github.com/zeroc-ice/ice/actions/runs/20869322679/job/59967583960?pr=4866 (a Python PR):

```
[ running client/server test - 01/10/26 00:38:43 ]
- Config: ssl,serialize,mx,x64
(dotnet D:\a\ice\ice\csharp\test\Ice\metrics\msbuild\server\net8.0\server.dll --Ice.Default.Host=127.0.0.1 --Test.BasePort=14000 --Ice.Warn.Connections=1 --Ice.Default.Protocol=ssl --Ice.ThreadPool.Server.Serialize=1 --Ice.IPv6=0 --Ice.Admin.Endpoints="tcp -h 127.0.0.1" --Ice.Admin.InstanceName=Server --IceMX.Metrics.Debug.GroupBy=id --IceMX.Metrics.Parent.GroupBy=parent --IceMX.Metrics.All.GroupBy=none --IceSSL.*** --IceSSL.DefaultDir=D:\a\ice\ice\certs/common/ca --IceSSL.CAs=ca_cert.pem --IceSSL.VerifyPeer=2 --IceSSL.CertFile=server.p12 --Ice.ThreadPool.Server.Size=1 --Ice.ThreadPool.Server.SizeMax=3 --Ice.ThreadPool.Server.SizeWarn=0 --Ice.PrintAdapterReady=1 --Ice.Trace.Network=2 --Ice.Trace.Retry=1 --Ice.Trace.Protocol=1 --Ice.LogFile=D:\a\ice\ice\csharp\test\Ice\metrics\server-011026-0038.log env={'PATH': 'D:\\a\\ice\\ice\\cpp\\msbuild\\packages\\bzip2.v143.1.0.6.10\\build\\native\\bin\\x64\\Release'})
[...]

PR #5787, opened 2026-06-30
Report metric values on Ice/metrics EndpointLookup assertion failures

pepone:
The `Ice/metrics` test asserts via `test(bool)`, which throws a bare `Exception` with no detail. When the EndpointLookup section fails intermittently (#4872), the log shows only `System.Exception` at a line number — no actual metric values — so the failure can't be diagnosed.

This adds a `test(bool, string)` overload and a `metricsToString` helper, and includes the actual count plus each metric's `current`/`total`/`failures` in the endpoint-lookup assertion messages. The repeated `getMetricsView(...)` calls in that section are collapsed into a single `lookups` local.

Diagnostics only — the assertions' intent is unchanged. Next time the section fails, the message will show e.g. `[{ id="ssl -h localhost -p 14000 -t 500" current=0 total=2 failures=0 }]`, which will confirm or refute the suspected retry/double-lookup.

Relates to #4872 (does not fix the intermittent failure itself).

### DIFF
--- a/csharp/test/Ice/metrics/AllTests.cs
+++ b/csharp/test/Ice/metrics/AllTests.cs
@@ -33,6 +33,14 @@ public class AllTests : Test.AllTests
         }
     }
 
+    // Formats a metrics array for assertion failure messages, to make intermittent metrics failures diagnosable.
+    private static string metricsToString(IceMX.Metrics[] metrics) =>
+        "[" +
+        string.Join(
+            ", ",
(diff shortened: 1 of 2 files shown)


bernardnormier (approved):
Nice diagnostics improvement. The `test(bool, string)` overload mirrors the existing `test(bool)` (Debug.Assert + throw), and collapsing the repeated `getMetricsView(...)` calls into a single `lookups` local is a small correctness win too — the failure message now reports the same snapshot that was asserted, rather than a possibly-different re-fetch. Assertion intent is unchanged.

PR #5788, opened 2026-06-30
Fix Schannel data loss when a renegotiation coincides with application data

pepone:
Fixes the high-severity audit finding zeroc-ice/ice-audit#58.

## Problem

On Windows (Schannel), `Schannel::TransceiverI::decryptMessage` copies decrypted application data into the caller's buffer as it loops over TLS records, advancing a local cursor `i` (the caller's `buf.i` is advanced by the *caller*). On `SEC_I_RENEGOTIATE` it returned `0` instead of `i - buffer.i` — the number of bytes it had already copied during that call (from draining `_readUnprocessed` and from earlier records in the same batch).

Both callers (`read()` and `finishRead()`) then performed the renegotiation handshake and `continue`d **without** advancing `buf.i` for those bytes, so the next `decryptMessage` wrote at the same `buf.i` and overwrote the already-extracted plaintext.

### Impact

A client connection where Schannel surfaces a TLS 1.3 `KeyUpdate` / `NewSessionTicket` (reported as `SEC_I_RENEGOTIATE`) in the same read as already-decrypted application data silently loses those application bytes — Ice protocol-stream corruption, typically surfacing later as an unmarshaling or protocol error that is hard to attribute back to the renegotiation. This is reachable by a conformant peer.

## Fix

- `decryptMessage`: return `i - buffer.i` on `SEC_I_RENEGOTIATE` instead of `0`.
- `read()`: advance `buf.i += decrypted` immediately after `decryptMessage`, so the bytes copied before the renegotiation survive the `continue`.
- `finishRead()`: advance `buf.i` per `decryptMessage` inside the loop (it previously advanced once, after the loop), and guard the loop with `decryptMessage`'s data precondition. The readiness signal now reflects total progress across the call.

After the fix, the plaintext extracted before the renegotiation is preserved and the post-handshake application data follows it contiguously and in order; `buf.i` is advanced by exactly the number of plaintext bytes delivered.

## Verification

- Compiles and links cleanly on `main` (Windows, `ice.dll`).
- Reviewed by tracing the state machine for both `read()` and `finishRead()`, covering the app-records-then-renegotiation and renegotiation-first cases. Confirmed no data loss/overwrite, no new asserts or infinite loops, and that `ready()` matches the previous behavior on the non-renegotiation path. The new `finishRead` loop guard is exactly `decryptMessage`'s data precondition, so it is strictly safer than the previous unconditional `while (true)`.

## Testing notes

A targeted regression test is hard to construct deterministically — it requires injecting a renegotiation record mid-batch behind already-decrypted application data, which depends on Schannel's record framing. The existing SSL test suite exercises the non-renegotiation paths.

### DIFF
--- a/changelog.d/cpp/schannel-renegotiate-data-loss.md
+++ b/changelog.d/cpp/schannel-renegotiate-data-loss.md
@@ -1,3 +1,3 @@
 - Fixed silent application-data loss in the Schannel-based IceSSL transport (Windows) when a TLS renegotiation (such as
   a TLS 1.3 KeyUpdate or NewSessionTicket) is received in the same read as already-decrypted application data. The
-  plaintext extracted before the renegotiation request was dropped, corrupting the Ice protocol stream.
+  plaintext extracted before the renegotiation request was dropped, breaking the affected connection.
--- a/changelog.d/cpp/schannel-renegotiate-data-loss.md
+++ b/changelog.d/cpp/schannel-renegotiate-data-loss.md
@@ -1,0 +1,3 @@
+- Fixed silent application-data loss in the Schannel-based IceSSL transport (Windows) when a TLS renegotiation (such as
+  a TLS 1.3 KeyUpdate or NewSessionTicket) is received in the same read as already-decrypted application data. The
+  plaintext extracted before the renegotiation request was dropped, corrupting the Ice protocol stream.
--- a/cpp/src/Ice/SSL/SchannelTransceiverI.cpp
+++ b/cpp/src/Ice/SSL/SchannelTransceiverI.cpp
@@ -605,7 +605,11 @@ Schannel::TransceiverI::decryptMessage(IceInternal::Buffer& buffer)
                 memcpy(_extraBuffer.i, extraBuffer->pvBuffer, extraBuffer->cbBuffer);
                 _extraBuffer.i += extraBuffer->cbBuffer;
             }
-            return 0;
+            // Return the application data already copied into the caller's buffer during this call (it may be non-zero
+            // because earlier records in the same batch were decrypted before this renegotiation request). The caller
+            // must advance buf.i by this amount before performing the re-handshake, otherwise the next decryptMessage
+            // overwrites these bytes and the plaintext is silently lost.
+            return i - buffer.i;
         }
         else if (err == SEC_I_CONTEXT_EXPIRED)
         {
@@ -829,6 +833,9 @@ Schannel::TransceiverI::read(IceInternal::Buffer& buf)
         }
 
         size_t decrypted = decryptMessage(buf);
+        // Account for the bytes decryptMessage copied into buf right away, so a renegotiation in the middle of the
+        // batch does not drop the plaintext already extracted before it.
+        buf.i += decrypted;
         if (_sslConnectionRenegotiating)
         {
             // The peer has requested a renegotiation.
@@ -854,8 +861,6 @@ Schannel::TransceiverI::read(IceInternal::Buffer& buf)
             }
             continue;
         }
-
-        buf.i += decrypted;
     }
     _delegate->getNativeInfo()->ready(
         IceInternal::SocketOperationRead,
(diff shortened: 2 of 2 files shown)


changelog.d/cpp/schannel-renegotiate-data-loss.md
```diff
@@ -0,0 +1,3 @@
+- Fixed silent application-data loss in the Schannel-based IceSSL transport (Windows) when a TLS renegotiation (such as
+  a TLS 1.3 KeyUpdate or NewSessionTicket) is received in the same read as already-decrypted application data. The
+  plaintext extracted before the renegotiation request was dropped, corrupting the Ice protocol stream.
```
bernardnormier: "Ice protocol stream" mixes layers: the bug corrupts the **transport** byte stream (the transceiver silently drops decrypted application bytes), and the Ice protocol *frames* are just the downstream casualty. Since the first sentence already says "application-data loss," I'd replace this with the user-visible effect — the connection breaks:

```suggestion
  plaintext extracted before the renegotiation request was dropped, breaking the affected connection.
```

bernardnormier (approved):
The fix is correct — I traced the state machine through both `read()` and `finishRead()`. Returning `i - buffer.i` on `SEC_I_RENEGOTIATE`, plus advancing `buf.i` before the re-handshake, preserves the plaintext extracted ahead of the renegotiation, and the non-renegotiation paths reduce to the previous behavior (when no app data precedes the record, `i - buffer.i == 0`, identical to today). The new `finishRead` loop guard matching `decryptMessage`'s precondition is also a nice tightening over the old `while (true)`. Scope is right: only Schannel surfaces renegotiation to the application — OpenSSL/SecureTransport consume post-handshake records inside `SSL_read`/`SSLRead`, so there's no parallel bug there.

Follow-up suggestion (separate PR, not blocking this one): add a cross-platform interop test that actually exercises this path. The same-engine harness can't reach it — Schannel-as-server can't be made to emit a post-handshake record *after* application data — but an OpenSSL peer can. Setup: a Linux Ice+OpenSSL server that calls `SSL_key_update()` (or `SSL_new_session_ticket()`) mid-stream, writing `[app data][KeyUpdate][app data]` back-to-back so they coalesce into a single Schannel read; a Windows Ice+Schannel client reads a known payload and asserts byte-for-byte integrity. The buggy build drops a record's worth of plaintext; the fixed build doesn't.

This tests the genuine `DecryptMessage` -> `SEC_I_RENEGOTIATE` path (no simulation). The Schannel side is a plain Ice client, so the only test-only piece is a peer-side KeyUpdate trigger on the OpenSSL (Linux) side. It'd be a manual cross-machine test to start, optionally promoted to a dedicated interop job later.

cpp/src/Ice/SSL/SchannelTransceiverI.cpp
```diff
@@ -925,10 +930,14 @@ Schannel::TransceiverI::finishRead(IceInternal::Buffer& buf)
     _delegate->finishRead(_readBuffer);
     if (_state == StateHandshakeComplete)
     {
-        size_t decrypted;
-        while (true)
+        std::byte* const start = buf.i;
+        // Decrypt the buffered data into buf, advancing buf.i as we go so that a renegotiation in the middle of the
+        // batch does not drop or overwrite the plaintext already extracted. The loop guard matches decryptMessage's
+        // precondition: there is buffered (_readBuffer) or unprocessed data left to decrypt.
+        while (buf.i != buf.b.end() && (_readBuffer.i != _readBuffer.b.begin() || !_readUnprocessed.b.empty()))
```
externl, line 941: This also fixes a latent bug: the old `while (true)` could re-enter `decryptMessage` with its buffers drained (partially-filled caller buffer + `[app][KeyUpdate]` batch), tripping its precondition assert in debug builds. Maybe worth noting in the commit message?

cpp/src/Ice/SSL/SchannelTransceiverI.cpp
```diff
@@ -605,7 +605,11 @@ Schannel::TransceiverI::decryptMessage(IceInternal::Buffer& buffer)
                 memcpy(_extraBuffer.i, extraBuffer->pvBuffer, extraBuffer->cbBuffer);
                 _extraBuffer.i += extraBuffer->cbBuffer;
             }
-            return 0;
+            // Return the application data already copied into the caller's buffer during this call (it may be non-zero
+            // because earlier records in the same batch were decrypted before this renegotiation request). The caller
+            // must advance buf.i by this amount before performing the re-handshake, otherwise the next decryptMessage
+            // overwrites these bytes and the plaintext is silently lost.
+            return i - buffer.i;
```
externl, line 616: Worth a comment: recovering app data after the `KeyUpdate` works only because `_readBuffer.i` is left unreset and `sslHandshake` reads the trailing record back out of `_readBuffer`. Resetting it here (like the `SEC_E_OK` path) would underflow the `memmove`. Pre-existing.

cpp/src/Ice/SSL/SchannelTransceiverI.cpp
```diff
@@ -829,6 +833,9 @@ Schannel::TransceiverI::read(IceInternal::Buffer& buf)
         }
 
         size_t decrypted = decryptMessage(buf);
+        // Account for the bytes decryptMessage copied into buf right away, so a renegotiation in the middle of the
+        // batch does not drop the plaintext already extracted before it.
+        buf.i += decrypted;
         if (_sslConnectionRenegotiating)
```
externl, line 843: Pre-existing: a conformant `KeyUpdate(update_requested)` that can't flush its response in one write becomes `ConnectionLostException` instead of resuming. Narrow, but pairs with the suggested interop test.

externl (approved):
Reviewed with Claude. Fix looks correct — preserving `buf.i` before the re-handshake is right, and the non-renegotiation paths are unchanged. A few non-blocking notes.

PR #5789, opened 2026-06-30
Escape '#' in slice2rb-generated Ruby string literals

bernardnormier:
`slice2rb` emitted Slice string constants and operation deprecation reasons into double-quoted Ruby string literals without escaping `#`. In a double-quoted Ruby string, `#{...}`, `#@ivar`, and `#$gvar` are interpolation sequences, so a Slice string containing this syntax was evaluated when the generated file was loaded — corrupting the constant's value, raising a load-time error, or executing the embedded expression.

This adds `#` to the set of characters escaped by `toStringLiteral` for both emitters (the same fix PHP already applies for `$`).

A constant exercising `#{`, `#@`, `#$`, and a bare `#` was added to the Ruby `defaultValue` test. Verified that the fixture crashes at load time without the fix (`NameError`) and passes with it.


cpp/src/slice2rb/RubyUtil.cpp
```diff
@@ -467,7 +467,7 @@ Slice::Ruby::CodeVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
         {
             // Get the deprecation reason if present, or default to an empty string.
             const string reason = op->getDeprecationReason().value_or("");
-            const string escapedReason = toStringLiteral(reason, "\a\b\f\n\r\t\v\x20\x1b", "", EC6UCN, 0);
+            const string escapedReason = toStringLiteral(reason, "\a\b\f\n\r\t\v\x20\x1b", "#", EC6UCN, 0);
```
externl, line 470: Optional, non-blocking: this deprecation-reason path is fixed but isn't covered by a test — only the constant path (`ConstHashString`) is. Same mechanism, so it's proven transitively; a test would need a deprecated operation with a `#{`-containing reason. Could also mirror the explanatory comment you added at the constant site (line 1009) here.

externl (approved):
Reviewed with Claude. Correct and complete — the two fixed sites are the only ones that emit an arbitrary Slice string into a double-quoted Ruby literal; the other double-quote emissions are all identifiers (which can't contain `#`). With `#` added, the emission now escapes the full set of Ruby double-quote-special chars (`\`, `"`, `#`). One optional note inline.

pepone (approved):

PR #5791, opened 2026-07-01
Fix OpenSSL-based IceSSL build against OpenSSL 4.0

pepone:
Motivated by the Debian OpenSSL 4.0 transition, which flagged that `libzeroc-ice` fails to build against OpenSSL 4.0 — Debian bug [#1138403](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1138403). (The fix is prepared here on `main` and will be cherry-picked to `3.8`, since 3.8.3 is what we upload to Debian sid.)

## Problem

OpenSSL 4.0 makes `ASN1_STRING` (and its typedefs `ASN1_IA5STRING`, `ASN1_OCTET_STRING`, …) **opaque** — code that accesses `->data` / `->length` / `->type` directly no longer compiles. The OpenSSL-based IceSSL transport does exactly that in `convertGeneralNames` (Subject Alternative Name parsing), so it fails to build against OpenSSL 4.0:

```
error: invalid use of incomplete type 'ASN1_IA5STRING' {aka 'struct asn1_string_st'}
```

## Fix

Replace the direct `ASN1_STRING` member accesses with the accessor functions `ASN1_STRING_get0_data`, `ASN1_STRING_length`, and `ASN1_STRING_type`. These have existed since OpenSSL 1.1.0 and behave identically on 3.x, so no version guards are needed. `GENERAL_NAME` itself is not opaque, so `gen->type` / `gen->d.*` are unchanged.

## Verification

| Test | OpenSSL | Result |
|---|---|---|
| Original pattern | 4.0.1 (fedora:rawhide) | ❌ `invalid use of incomplete type 'ASN1_IA5STRING'` |
| Fixed pattern | 4.0.1 | ✅ compiles |
| Fixed pattern | 3.6.3 (host) | ✅ compiles — no regression |

This is the only OpenSSL-4.0 blocker in the current transport: the SSL sources already build with `OPENSSL_API_COMPAT 30000` + `OPENSSL_NO_DEPRECATED`, use no ENGINE API, no SSLv3, and no other opaque-struct access.

### DIFF
--- a/changelog.d/cpp/openssl-4.0-asn1-opaque.md
+++ b/changelog.d/cpp/openssl-4.0-asn1-opaque.md
@@ -1,0 +1,3 @@
+- Fixed a build failure of the OpenSSL-based IceSSL transport against OpenSSL 4.0, which makes `ASN1_STRING`
+  opaque. Subject Alternative Name parsing now uses the `ASN1_STRING_get0_data`, `ASN1_STRING_length`, and
+  `ASN1_STRING_type` accessors instead of direct struct member access.
--- a/cpp/src/Ice/SSL/SSLUtil.cpp
+++ b/cpp/src/Ice/SSL/SSLUtil.cpp
@@ -496,6 +496,19 @@ namespace
         return result;
     }
 
+    string convertIA5StringToString(ASN1_IA5STRING* str)
+    {
+        if (str && ASN1_STRING_type(str) == V_ASN1_IA5STRING && ASN1_STRING_length(str) > 0)
+        {
+            const unsigned char* data = ASN1_STRING_get0_data(str);
+            if (data)
+            {
+                return string(reinterpret_cast<const char*>(data), ASN1_STRING_length(str));
+            }
+        }
+        return {};
+    }
+
     vector<pair<int, string>> convertGeneralNames(GENERAL_NAMES* gens)
     {
         vector<pair<int, string>> alt;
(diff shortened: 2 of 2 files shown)


externl (approved):
Looks good!

bernardnormier (approved):
LGTM. I verified the accessors are drop-in equivalents of the removed member accesses (same guards, same bytes, available in every OpenSSL version we build against), and that `convertGeneralNames` was indeed the only direct access to a struct that OpenSSL 4.0 makes opaque — the remaining direct accesses (`GENERAL_NAME`, `BUF_MEM`) are on structs that stay public.

One suggestion: the three IA5STRING cases are now identical 8-line blocks, each repeating the guard and the `reinterpret_cast`. A small helper next to `convertX509NameToString` (which GEN_DIRNAME already uses in this switch) would collapse each case to one line and keep a single cast:

```cpp
string convertIA5StringToString(ASN1_IA5STRING* str)
{
    if (str && ASN1_STRING_type(str) == V_ASN1_IA5STRING && ASN1_STRING_length(str) > 0)
    {
        const unsigned char* data = ASN1_STRING_get0_data(str);
        if (data)
        {
            return string(reinterpret_cast<const char*>(data), ASN1_STRING_length(str));
        }
    }
    return {};
}
```

```cpp
case GEN_EMAIL:
{
    p.second = convertIA5StringToString(gen->d.rfc822Name);
    break;
}
case GEN_DNS:
{
    p.second = convertIA5StringToString(gen->d.dNSName);
    break;
}
...
case GEN_URI:
{
    p.second = convertIA5StringToString(gen->d.uniformResourceIdentifier);
    break;
}
```

(While you're there, hoisting `const unsigned char* data = ASN1_STRING_get0_data(addr);` in the GEN_IPADD case would similarly un-wrap that condition and restore `data[j]` in the loop.)

pepone:
🤖 **Claude:** Applied both suggestions in c2c6d43 — extracted your `convertIA5StringToString` helper and collapsed the three IA5STRING cases (GEN_EMAIL/GEN_DNS/GEN_URI) to one-liners, and un-wrapped GEN_IPADD to fetch the octet data once.

One tweak on the IPADD hoist: I kept `const unsigned char* data = ASN1_STRING_get0_data(addr);` **inside** the guard (after the `addr` null / type / length checks) rather than before the condition, since `ASN1_STRING_get0_data` dereferences `addr` — same shape as your IA5STRING helper.

Verified old vs new `convertGeneralNames` produce identical output on a cert with all four SAN types (email/DNS/URI/IP), and `IceSSL/configuration` still passes.

Issue #5793, opened 2026-07-01
IAPEndpointInfo.protocol returns the transport name instead of the accessory protocol string


bernardnormier:
`Ice::IAPEndpointInfo::protocol` is documented as *"The protocol supported by the accessory"* — i.e. the External Accessory protocol string set via the `-p` endpoint option (stored in `iAPEndpointI::_protocol`, e.g. `com.zeroc.ice`).

However, `iAPEndpointI::getInfo()` populates that field with `protocol()`, which returns the transport name (`"iap"` or `"iaps"`), not the accessory protocol string:

https://github.com/zeroc-ice/ice/blob/main/cpp/src/Ice/ios/iAPEndpointI.mm#L128-L137

```cpp
return make_shared<Ice::IAPEndpointInfo>(
    _compress,
    _manufacturer,
    _modelNumber,
    _name,
    protocol(),   // transport name ("iap"), not _protocol
    type(),
    secure());
```

As a result, the accessory protocol string is not available through `IAPEndpointInfo`. It is still present in the endpoint's string form (`-p ...` via `options()`), so the impact is limited to the info API.

Expected: `IAPEndpointInfo::protocol` should carry `_protocol` (the `-p` value).

Found while adding tests for the iAP transport (#5792); the test intentionally does not assert this field.

PR #5800, opened 2026-07-01
Fix iOS iAP getInfo: nil-safe IAPConnectionInfo and correct IAPEndpointInfo protocol

### LINKED ISSUE #5794: iAPTransceiver::getInfo can construct std::string from a null UTF8String
`iAPTransceiver::getInfo()` builds an `IAPConnectionInfo` from `EAAccessory` fields via `[... UTF8String]`:

https://github.com/zeroc-ice/ice/blob/main/cpp/src/Ice/ios/iAPTransceiver.mm#L386-L399

```cpp
return make_shared<Ice::IAPConnectionInfo>(
    std::move(adapterName),
    std::move(connectionId),
    [_session.accessory.name UTF8String],
    [_session.accessory.manufacturer UTF8String],
    [_session.accessory.modelNumber UTF8String],
    [_session.accessory.firmwareRevision UTF8String],
    [_session.accessory.hardwareRevision UTF8String],
    [_session.protocolString UTF8String]);
```

If any of these Objective-C strings is `nil`, or `UTF8String` returns `nullptr` (a string not representable as UTF-8), the corresponding `std::string` is constructed from a null pointer, which is undefined behavior. Some `EAAccessory` fields are optional and may be `nil` in practice.

The same pattern was made nil-safe in `iAPEndpointI::connectorsAsync` in #5792; `getInfo()` should use the same nil-safe conversion (ideally a small shared helper). Raised as a follow-up so #5792 could stay focused on tests.

bernardnormier:
Fixes #5794 and #5793 — two bugs in the iOS iAP transport's `getInfo()` methods.

### #5794 — nil-safe `IAPConnectionInfo`
`iAPTransceiver::getInfo()` built each `IAPConnectionInfo` field by constructing a `std::string` directly from `[... UTF8String]`. When an `EAAccessory` field is `nil` (several are optional) or not representable as UTF-8, `UTF8String` returns `nullptr` and the `std::string` construction is undefined behavior. The fields now go through a nil-safe conversion helper.

### #5793 — correct `IAPEndpointInfo::protocol`
`iAPEndpointI::getInfo()` populated `IAPEndpointInfo::protocol` with `protocol()` — the transport name (`iap`/`iaps`) — instead of `_protocol`, the External Accessory protocol string set via the `-p` endpoint option. It now reports `_protocol`, matching the documented meaning ("The protocol supported by the accessory").

### Other changes
- Extracted the `NSString*`→`std::string` conversion (previously a local lambda in `connectorsAsync`) into a shared `IceObjC::nsToString` helper in a new `iAPUtil.h`.
- Added assertions for `IAPEndpointInfo::protocol` to the iAP endpoint test.

Both modified `.mm` files were syntax-checked against the iphonesimulator SDK. The iAP transport is iOS-only, so the test suite runs only on the iOS simulator/device.

### DIFF
--- a/cpp/src/Ice/ios/iAPEndpointI.mm
+++ b/cpp/src/Ice/ios/iAPEndpointI.mm
@@ -18,6 +18,7 @@
 #    include "iAPConnector.h"
 #    include "iAPEndpointI.h"
 #    include "iAPMatch.h"
+#    include "iAPUtil.h"
 
 #    include <CoreFoundation/CoreFoundation.h>
 
@@ -132,7 +133,7 @@ IceObjC::iAPEndpointI::getInfo() const noexcept
         _manufacturer,
         _modelNumber,
         _name,
-        protocol(),
+        _protocol,
         type(),
         secure());
 }
@@ -244,12 +245,6 @@ IceObjC::iAPEndpointI::connectorsAsync(
 
         NSString* protocol =
             _protocol.empty() ? @"com.zeroc.ice" : [[NSString alloc] initWithUTF8String:_protocol.c_str()];
-        // Converts an NSString to a std::string, yielding "" when the string is nil or not representable as UTF-8.
-        auto toString = [](NSString* s)
-        {
-            const char* utf8 = [s UTF8String];
-            return utf8 ? string{utf8} : string{};
-        };
 
         NSArray* array = [manager connectedAccessories];
         NSEnumerator* enumerator = [array objectEnumerator];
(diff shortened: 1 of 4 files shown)


externl (approved):

pepone (approved):
Reviewed the fix and cross-checked it with Codex (gpt-5.5) — both passes agree it's correct with no introduced issues in the modified paths.

**#5794 (nil-safe `IAPConnectionInfo`)** — right fix. The old code fed `[nil UTF8String]` (→ `nullptr`) straight into `std::string{}`, which is UB. `nsToString` covers both hazards: messaging `nil` returns `nil`, and `UTF8String` on a non-UTF-8-representable string also returns `nullptr` — both map to `""`. The inner pointer is copied immediately, so no lifetime concern, and the field order matches the `IAPConnectionInfo` ctor.

**#5793 (`IAPEndpointInfo::protocol`)** — reporting `_protocol` (the EA protocol from `-p`) instead of `protocol()` (the transport name `iap`/`iaps`) matches the field's documented meaning, "the protocol supported by the accessory." The added assertions genuinely guard it: `-p com.zeroc.ice` → `"com.zeroc.ice"`, bare `iap` → empty; the first would have failed pre-fix (it returned `"iap"`).

Also confirmed:
- The unqualified `nsToString(...)` calls resolve to `IceObjC::nsToString` since both call sites are out-of-line members of classes in `namespace IceObjC`.
- The refactor is behavior-preserving (`nsToString` is exactly the removed lambda), `inline` in a guarded header (no ODR issue), and both includes sit inside the existing `#if TARGET_OS_IPHONE != 0` guard.

**Optional, out of scope for this PR:** the same `UTF8String` nil-hazard pattern still exists in a few sibling spots that the new `nsToString` helper could clean up — `iAPTransceiver.mm` (`string{[domain UTF8String]}` in the `SocketException` path, and the `os << [... UTF8String]` in `_desc`) and `iAPConnector::toString()`. The `toString`/`_desc` ones stream the same optional `EAAccessory` fields (`modelNumber`/`manufacturer`) flagged in #5794, so they can still crash on a nil field. All pre-existing — just flagging for a possible follow-up sweep.

PR #5802, opened 2026-07-01
Fix slice2cpp invalid C++ literal for float constants in exponent form

### LINKED ISSUE #5477: slice2cpp: float constants in exponent form generate invalid C++ literals (1e+07.0F)
> **AI-generated audit finding** — this issue was opened from an automated **correctness** audit of the Ice C++ source performed by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), and independently re-derived against the source (including against generated `slice2cpp` output) by a second model (OpenAI **Codex GPT-5**). It has **not been triaged by a human yet**; please verify the reasoning, reproducibility, and severity before acting on it.

## Summary

`slice2cpp` emits float constants by appending `.0F` when the value string contains no `.`. But the parser stringifies a `double` with default `ostringstream` formatting, so any float ≥ 1e6 or < 1e-4 is rendered in scientific notation **without** a decimal point (e.g. `1e+07`). The emitter then produces `1e+07.0F` — not a valid C++ literal — and the generated header fails to compile.

## The bug

```cpp
// cpp/src/slice2cpp/Gen.cpp:123
else if (bp && bp->kind() == Builtin::KindFloat)
{
    out << value;
    if (value.find('.') == string::npos) { out << ".0"; }   // misses exponent form
    out << "F";
}
```

The value string comes from `ostringstream << double` (`Grammar.y:1540`, `Scanner.l:314`). The C# generator just appends `F` (`1e+07F` is valid C#), so only the C++ path is affected.

## Impact

`const float Big = 10000000.0;` or a data member `float f = 1e8;` generates `constexpr float Big = 1e+07.0F;` — the generated header does not compile.

## Source references

- **(slice2cpp / float exponent form gets .0F appended)** — [`cpp/src/slice2cpp/Gen.cpp#L123`](https://github.com/zeroc-ice/ice/blob/e248644d788c7e76e14af8676529201c54c1735d/cpp/src/slice2cpp/Gen.cpp#L123)

## Recommendation

Treat a value containing `e`/`E` as already floating-point (append only `F`, no `.0`), or render float constants with a format that always includes a decimal point.

## Severity

**Medium** — legal Slice float constants generate non-compiling C++.

bernardnormier:
## Description

`slice2cpp` emits a `float` constant by appending `.0F` when the value string contains no `.`. But the value is stringified from a `double` with default `ostringstream` formatting, so any value whose magnitude forces scientific notation (roughly ≥ 1e6 or < 1e-4, e.g. `1e8` → `"1e+08"`) is rendered *without* a decimal point. The emitter then produced `1e+08.0F` — an invalid C++ literal (`invalid suffix '.0F'`) — and the generated header failed to compile.

The fix appends `.0` only when the value is an integer literal, i.e. it has neither a decimal point nor an exponent. Values already in floating-point form (containing `.`, `e`, or `E`) get only the `F` suffix.

`const float Big = 1e8;` now generates `constexpr float Big = 1e+08F;` instead of `1e+08.0F`.

Only the C++ mapping is affected: the C# generator appends just `F`, and `1e+08F` is valid C#.

## Test

Added `const float ConstExpFloat = 1e8;` to the `Ice/defaultValue` Slice test (a C++-local fixture) with a value check in `AllTests.cpp`. Without the fix, the generated header does not compile.

Fixes #5477

### DIFF
--- a/changelog.d/cpp/slice2cpp-float-exponent-literal.md
+++ b/changelog.d/cpp/slice2cpp-float-exponent-literal.md
@@ -1,0 +1,3 @@
+- Fixed a `slice2cpp` bug where a `float` constant whose value is rendered in scientific notation (any magnitude
+  ≥ 1e6 or < 1e-4, e.g. `1e8`) generated an invalid C++ literal such as `1e+08.0F`, causing the generated header to
+  fail to compile.
--- a/cpp/src/slice2cpp/Gen.cpp
+++ b/cpp/src/slice2cpp/Gen.cpp
@@ -124,7 +124,9 @@ namespace
             else if (bp && bp->kind() == Builtin::KindFloat)
             {
                 out << value;
-                if (value.find('.') == string::npos)
+                // Append ".0" only when the value is an integer literal, i.e. it has neither a decimal point nor an
+                // exponent (values such as "1e+07" are already valid floating-point literals).
+                if (value.find_first_of(".eE") == string::npos)
                 {
                     out << ".0";
                 }
--- a/cpp/test/Ice/defaultValue/AllTests.cpp
+++ b/cpp/test/Ice/defaultValue/AllTests.cpp
@@ -46,6 +46,7 @@ allTests()
         test(v.i == ConstInt);
         test(v.l == ConstLong);
         test(v.f == ConstFloat);
+        test(ConstExpFloat == 1e8f);
         test(v.d == ConstDouble);
         test(v.str == ConstString);
         test(v.c1 == ConstColor1);
--- a/cpp/test/Ice/defaultValue/Test.ice
+++ b/cpp/test/Ice/defaultValue/Test.ice
@@ -45,6 +45,7 @@ module Test
     const int ConstInt = 3;
     const long ConstLong = 4;
     const float ConstFloat = 5.1;
+    const float ConstExpFloat = 1e8;
     const double ConstDouble = 6.2;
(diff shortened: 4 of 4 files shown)


externl (approved):

pepone (approved):
We should run a cross language check for this one if not already done.

bernardnormier:
@pepone I ran a cross-language check of `writeConstantValue` (the constant-emission path) in all nine mappings. **C++ is the only one affected.**

| Mapping | float handling | output for `1e8` | affected? |
|---|---|---|---|
| C++ | appended `.0` when value had no `.` (the bug) | `1e+08F` (after fix) | **yes — fixed** |
| C# | `value` + `F`/`D` | `1e+08F` | no |
| Java | `value` + `F` (float), raw (double) | `1e+08F` | no |
| JS | raw `value` | `1e+08` | no |
| Swift | raw `value` | `1e+08` | no |
| PHP | raw `value` | `1e+08` | no |
| Python | raw `value` | `1e+08` | no |
| Ruby | raw `value` | `1e+08` | no |
| MATLAB | raw `value` | `1e+08` | no |

Every other generator emits the value verbatim or appends only a type suffix (`F`/`D`/`L`) that doesn't require a decimal point — none inserts `.0`, so none can produce a malformed `1e+08.0F`-style literal. I confirmed the raw exponent form `1e+08` is accepted by Python, Ruby, Node, PHP, and the Swift compiler, and the `1e+08F` form is valid in Java and C#.

PR #5803, opened 2026-07-02
Add hardware-free IceBT endpoint tests and a Bluetooth CI probe

externl:
The IceBT transport has had no CI coverage because its RFCOMM data path and BlueZ/D-Bus control path both need real hardware. This adds:

- A hardware-free test (`cpp/test/IceBT/endpoints`) that parses `bt`/`bts` endpoint strings with the IceBT plug-in loaded but not initialized (`Ice.InitPlugins=0`) — no Bluetooth daemon or radios needed.
- Coverage that a `bts` endpoint layers SSL over an IceBT endpoint, and that malformed endpoints are rejected.
- Build/run gating so it compiles only on Linux with BlueZ/D-Bus and runs in the normal suite under the default tcp config (no `--host-bt`).
- A manual `bluetooth-probe` workflow (`workflow_dispatch` only) that probes whether GitHub runners can host an emulated Bluetooth stack (virtual HCI via `hci_vhci` + BlueZ) — a step toward running IceBT over Bluetooth in CI without hardware.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -239,6 +239,29 @@ jobs:
           echo "AppTargetFramework=net10.0" >> $env:GITHUB_ENV
         shell: pwsh
 
+      # Report whether the runner can host an emulated Bluetooth stack (virtual HCI + BlueZ) -- a
+      # prerequisite for eventually exercising the IceBT transport over Bluetooth in CI without radios.
+      - name: Probe Bluetooth runner capabilities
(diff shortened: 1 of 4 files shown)


externl:
Superseded by #5839, which folds in these hardware-free endpoint tests and the CI probe (verified identical) alongside the Android-emulator IceBT-over-Bluetooth harness. Closing in favor of the consolidated PR.

PR #5806, opened 2026-07-06
Fix IceStorm shutdown hang from lost wakeup in SubscriberLink::flush()

### LINKED ISSUE #5657: IceStorm: lost condition-variable wakeup in SubscriberLink::flush() can hang shutdown
### Summary
`SubscriberLink::flush()` can drain its event queue to empty without ever calling `_condVar.notify_one()`. If that happens during `IceStorm::ServiceI::stop()`, `Subscriber::shutdown()` waits forever on its predicate and the whole IceStorm process hangs. This is the same lost-wakeup class fixed for `Subscriber::error()` in #5655, on a sibling path that PR does not touch.

### Affected code (main@`37c84bde06`)
- [`SubscriberLink::flush()` — Subscriber.cpp#L302-L363](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/Subscriber.cpp#L302-L363): the only drain-capable `flush()` with **no** terminal notify.
- [`v.swap(_events)` — #L312](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/Subscriber.cpp#L312) empties `_events`; [the cost filter erases entries — #L334](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/Subscriber.cpp#L334); [`if (!v.empty())` — #L341](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/Subscriber.cpp#L341) is skipped when all entries were filtered, so the function returns with no notify.
- [`Subscriber::completed()` — #L630-L656](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/Subscriber.cpp#L630-L656): after `--_outstanding`, when `_events` is non-empty it skips the [notify at #L648](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/Subscriber.cpp#L648) and delegates to [`flush()` at #L654](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/Subscriber.cpp#L654).
- [`Subscriber::shutdown()` wait predicate — #L797](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/Subscriber.cpp#L797): `_condVar.wait(lock, [this]{ return _outstanding == 0 && _events.empty(); })`.
- Contrast — [`SubscriberOneway::flush()` terminal notify — #L203-L206](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/Subscriber.cpp#L203-L206): the analogous drain-capable flush *does* notify.

### Mechanism
A `SubscriberLink` (topic federation) created with non-zero cost has a forward in flight when shutdown begins (`_shutdown = true`, `shutdown()` waiting). The forward completes → `completed()` does `--_outstanding` (to 0); `_events` is non-empty (events queued during the forward), so it calls `flush()`. `SubscriberLink::flush()` then `swap`s `_events` out (now empty) and cost-filters every entry away (`cost > _rec.cost`), so `if (!v.empty())` is false, `_outstanding` stays 0, and it returns **without notifying**. State is now `_outstanding == 0 && _events.empty() && _shutdown` — the wait predicate is satisfied, but no `notify_one()` ever fires, so `shutdown()` blocks forever. Because `ServiceI::stop()` runs while holding the topic and topic-manager mutexes, the whole process hangs.

`SubscriberTwoway::flush()` is not affected: it only empties `_events` by *sending* (always `++_outstanding`), never by discarding.

### Trigger conditions
Cost-based topic link (`link <topic> <cost>` with cost > 0); published events whose `"cost"` request-context exceeds the link cost; queued while a forward is outstanding, during IceStorm shutdown. Narrow / config-gated, but user-reachable — not purely theoretical.

### Impact
IceStorm shutdown hang (the service never stops). Availability. Severity: **Medium**.

### Suggested fix
Add the standard terminal notify at the end of `SubscriberLink::flush()`, mirroring `SubscriberOneway::flush()`:
```cpp
if (_events.empty() && _outstanding == 0 && _shutdown)
{
    _condVar.notify_one();
}
```

### Provenance
Found during review of #5655. Analysis by Claude (Opus 4.8) and independently corroborated by a Codex cross-check and a separate adversarial verification agent (all three reached the same path). Static analysis only — not reproduced by a running test, and not yet human-triaged.

pepone:
Fixes #5657.

`SubscriberLink::flush()` can drain its event queue to empty without ever calling `_condVar.notify_one()`. It swaps
`_events` into a local vector, cost-filters that vector, and discards every event whose `"cost"` request-context value
exceeds the link cost. When every event is discarded the send block (`if (!v.empty())`) is skipped, so the function
returns with `_events` empty and `_outstanding` still `0` but with no notify.

If that drain runs as the last outstanding forward completes during shutdown (`Subscriber::completed()` → `flush()`),
the `shutdown()` wait predicate (`_outstanding == 0 && _events.empty()`) is satisfied but never signalled, so
`shutdown()` blocks forever. Because `ServiceI::stop()` runs while holding the topic and topic-manager mutexes, the
whole IceStorm process hangs.

This is the same lost-wakeup class fixed for `Subscriber::error()` in #5655, on a sibling drain path that PR did not
touch. The fix adds the terminal notify that `SubscriberOneway::flush()` already performs. `SubscriberTwoway::flush()`
is unaffected: it only empties `_events` by sending (always `++_outstanding`), never by discarding.

A changelog fragment is included under `changelog.d/service-icestorm/`.

No test: like #5655, this timing- and config-gated lost wakeup has no proportionate deterministic reproducer.

### DIFF
--- a/changelog.d/service-icestorm/5657.md
+++ b/changelog.d/service-icestorm/5657.md
@@ -1,0 +1,2 @@
+- Fixed a hang that could prevent IceStorm from shutting down cleanly when a cost-based topic link discarded all of
+  its remaining queued events (because their cost exceeded the link's cost) as its last outstanding request completed.
--- a/changelog.d/service-icestorm/5657.md
+++ b/changelog.d/service-icestorm/5657.md
@@ -1,2 +1,2 @@
-- Fixed a hang that could prevent IceStorm from shutting down cleanly when a cost-based topic link discarded all of
-  its remaining queued events (because their cost exceeded the link's cost) as its last outstanding request completed.
+- Fixed a hang that could prevent IceStorm from shutting down cleanly after a topic received messages whose
+  cost prevented their propagation on one of its links.
--- a/cpp/src/IceStorm/Subscriber.cpp
+++ b/cpp/src/IceStorm/Subscriber.cpp
@@ -372,6 +372,13 @@ namespace
                 error(true, current_exception());
             }
         }
+
+        // If every queued event was filtered out by cost, we emptied _events without forwarding anything and no
+        // completion callback will fire; wake a waiting shutdown() ourselves, as SubscriberOneway::flush() does.
+        if (_events.empty() && _outstanding == 0 && _shutdown)
+        {
+            _condVar.notify_one();
+        }
     }
 }
 


changelog.d/service-icestorm/5657.md
```diff
@@ -0,0 +1,2 @@
+- Fixed a hang that could prevent IceStorm from shutting down cleanly when a cost-based topic link discarded all of
+  its remaining queued events (because their cost exceeded the link's cost) as its last outstanding request completed.
```
bernardnormier:
```suggestion
- Fixed a hang that could prevent IceStorm from shutting down cleanly after a topic received messages whose
  cost prevented their propagation on one of its links.
```

bernardnormier (approved):
LGTM. I verified the lost wakeup: the cost filter in `SubscriberLink::flush()` was the only remaining path that can empty the queue without either an outstanding request or a notify, and `shutdown()` waits on exactly that predicate while `ServiceI::stop()` holds the topic and topic-manager mutexes. The new check matches the existing `SubscriberOneway::flush()` idiom and runs under the (recursive) mutex at every call site.

One suggestion below: reword the changelog fragment in user-level terms (topic, link, cost, propagation) instead of the internal queue/discard mechanics.

externl (approved):
Verified the notify predicate matches the shutdown() wait predicate and that no other drain path still lacks a notify. Inline copy over a shared helper matches #5655.

Issue #5807, opened 2026-07-06
IceStorm: observerRemoveSubscriber can regress the persisted LLU on a concurrent topic destroy (replica path)


pepone:
### Summary
`TopicImpl::observerRemoveSubscriber` — the replica-side apply path for a replicated "remove subscriber" log update —
commits `_lluMap.put` without checking `_destroyed`, the same missing guard fixed for `observerAddSubscriber` in #5658.
On a concurrent topic destroy it can overwrite the destroy's newer LLU with the older remove LLU, regressing the
persisted last-log-update on the replica. Lower severity than #5658: it deletes (rather than creates) records, so it
produces no orphan and no topic resurrection — just a transient, self-healing LLU regression.

### Affected code (main@`b71bc8ed0a`)
- [`TopicImpl::observerRemoveSubscriber` — TopicI.cpp#L1068-L1121](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/IceStorm/TopicI.cpp#L1068-L1121):
  takes `_subscribersMutex` ([#L1086](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/IceStorm/TopicI.cpp#L1086)),
  deletes subscriber records ([#L1096](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/IceStorm/TopicI.cpp#L1096)),
  and commits [`_lluMap.put` — #L1099](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/IceStorm/TopicI.cpp#L1099)
  with **no `_destroyed` check**.
- The window — [`TopicManagerImpl::observerRemoveSubscriber` — TopicManagerI.cpp#L560-L575](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/IceStorm/TopicManagerI.cpp#L560-L575):
  looks up the topic under `_mutex`, **releases `_mutex` at [#L573](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/IceStorm/TopicManagerI.cpp#L573)**,
[...]

PR #5808, opened 2026-07-06
Guard IceStorm observerAddSubscriber against a destroyed topic

### LINKED ISSUE #5658: IceStorm: TopicImpl::observerAddSubscriber commits subscriber records without a _destroyed guard (replica path)
### Summary
`TopicImpl::observerAddSubscriber` — the replica-side apply path for a replicated "add subscriber" log update — commits a `SubscriberRecord` to LMDB without checking `_destroyed`, unlike its sibling `observerDestroyTopic` and unlike the client-facing `subscribeAndGetPublisher`/`link` (which #5655 guards). A concurrent topic destroy can leave an orphan record for a destroyed topic — the same DB-corruption class #5655 addresses, on the replication path.

### Affected code (main@`37c84bde06`)
- [`TopicImpl::observerAddSubscriber` — TopicI.cpp#L992-L1045](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/TopicI.cpp#L992-L1045): takes `_subscribersMutex` (#L994) and commits via [`_subscriberMap.put` — #L1038](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/TopicI.cpp#L1038) with **no `_destroyed` check** in between.
- Guarded sibling — [`TopicImpl::observerDestroyTopic` — TopicI.cpp#L1111-L1119](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/TopicI.cpp#L1111-L1119): checks `_destroyed` right after the lock.
- [`TopicImpl::destroy` sets `_destroyed` under `_subscribersMutex` — #L773-L781](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/TopicI.cpp#L773-L781).
- Client paths #5655 guards: [`subscribeAndGetPublisher` — #L494](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/TopicI.cpp#L494) and [`link` — #L600](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/TopicI.cpp#L600).
- The window — [`TopicManagerImpl::observerAddSubscriber` — TopicManagerI.cpp#L542-L557](https://github.com/zeroc-ice/ice/blob/37c84bde06b664bc9766d3d6f3c74f04a5d8142d/cpp/src/IceStorm/TopicManagerI.cpp#L542-L557): looks up the topic under `_mutex`, **releases `_mutex` at #L555**, then calls `topic->observerAddSubscriber` at #L556 outside the lock.

### Mechanism
On a replica, `TopicManagerImpl::observerAddSubscriber` resolves the `TopicImpl` under `_mutex`, releases `_mutex`, then invokes `topic->observerAddSubscriber`. If a concurrent `observerDestroyTopic` sets `_destroyed` on that topic in the window between release and call, `observerAddSubscriber` (no guard) still commits a `SubscriberRecord` keyed by the now-destroyed topic, with no placeholder record. On restart, `TopicManagerImpl` expects the first record per topic to be a placeholder — debug builds assert/abort; release builds resurrect the destroyed topic with an orphan subscriber.

### Scope / caveats
- Pre-existing; identical on `main`, untouched by #5655.
- Replica/HA deployments only (the master→replica replication apply path), not client-facing — not remotely triggerable by a publisher/subscriber.
- Partial mitigation: if the destroy removes the topic from `_topics` *before* the lookup, a fresh `observerAddSubscriber` throws `ObserverInconsistencyException` before committing; the gap is specifically the lookup→release→call ordering.

### Impact
Persistent DB corruption (orphan/resurrected topic; restart abort in debug) on replicated IceStorm. Severity: **Medium**.

### Suggested fix
Add a `_destroyed` guard right after the lock in `observerAddSubscriber`. Note this is the **replication apply path**, not a client call — so the right handling likely differs from the client paths (which throw `ObjectNotExistException`): it may need to *skip/return* (and reconcile via the LLU/recovery machinery) rather than throw, to avoid disrupting replication. Exact handling needs the author's judgment.

### Provenance
Found during review of #5655. Analysis by Claude (Opus 4.8), corroborated by an independent adversarial verification agent; guard placement and the `_mutex`-release window verified against `main`. Static analysis only — not yet human-triaged.

pepone:
Adds a defensive `_destroyed` guard at the top of `TopicImpl::observerAddSubscriber` (the replica-side add-subscriber
apply path), matching the class idiom of checking `_destroyed` right after taking `_subscribersMutex` at every other
remote entry point (`observerDestroyTopic`, `subscribeAndGetPublisher`, `link`, `destroy`).

This is hardening, not a reachable bug fix. As analyzed in #5658, the master strictly serializes add-subscriber and
destroy for a topic: both hold the topic's `_subscribersMutex`, and `Observers::addSubscriber` waits synchronously for
every replica to apply the update before returning. So a destroy can't be in flight while an add is being applied on a
replica, and the lookup → release-`_mutex` → call window in `TopicManagerImpl::observerAddSubscriber` can never observe
`_destroyed` flip. No changelog fragment (no observable effect).

Closes #5658.

### DIFF
--- a/cpp/src/IceStorm/TopicI.cpp
+++ b/cpp/src/IceStorm/TopicI.cpp
@@ -1006,6 +1006,12 @@ TopicImpl::observerAddSubscriber(const LogUpdate& llu, const SubscriberRecord& r
 {
     lock_guard lock(_subscribersMutex);
 
+    // Ignore updates for a destroyed topic.
+    if (_destroyed)
+    {
+        return;
+    }
+
     auto traceLevels = _instance->traceLevels();
     if (traceLevels->topic > 0)
     {


externl (approved):
Verified both destroy and the guarded add hold _subscribersMutex across their LMDB work, and the skip is LLU-safe since the master serializes the two updates. Deferring the remove sibling to #5807 leaves no crash path (IceDB del() doesn't throw on a missing key).

---

**Edit — a deeper re-review turned up issues worth addressing before merge:**

- **Possible regression via observerInit reuse:** `observerInit` reuses an existing `TopicImpl` (TopicManagerI.cpp:484 `r->second->update(q.records)`) and nothing ever clears `_destroyed` (`update()` doesn't touch it). A destroyed-pending-reap instance revived by a new master's init keeps `_destroyed == true`, so the new guard silently drops legitimate replicated adds (record *and* LLU write) on that node — pre-PR they were persisted.
- **The described interleaving looks unreachable:** `Observers::addSubscriber`/`destroyTopic` hold `Observers::_mutex` across send **and** `wait()`, and the master holds `_subscribersMutex` across commit+broadcast, so a destroy can't dispatch concurrently with an in-flight add's unlocked window. The reachable path to a stale add appears to go through failed/timed-out invocations + recovery/observerInit — the PR body/changelog mechanism should probably be reworded (and the same caveat applies to #5807).
- **Minor:** the skip is silent at every trace level (the 'already subscribed' skip in the same function logs at `topic > 0`); the comment doesn't say why skipping the LLU put is safe; in release the restart symptom is a topic resurrected with *zero* subscribers (orphan consumed as pseudo-placeholder), not 'with the orphan subscriber'.

cpp/src/IceStorm/TopicI.cpp
```diff
@@ -1006,6 +1006,12 @@ TopicImpl::observerAddSubscriber(const LogUpdate& llu, const SubscriberRecord& r
 {
     lock_guard lock(_subscribersMutex);
 
+    // Skip the update if the topic was already destroyed; committing now would orphan a subscriber record.
```
bernardnormier:
```suggestion
    // Ignore updates for a destroyed topic.
```

bernardnormier (commented):
See https://github.com/zeroc-ice/ice/issues/5658#issuecomment-4893799894: the race described in #5658 is not reachable — the master sends the add-subscriber update while holding the topic's `_subscribersMutex` and waits synchronously for every replica's reply, so a destroy for the same topic can't be in flight concurrently with an add on a replica.

The guard itself is fine as defensive code: it matches the class idiom of checking `_destroyed` right after taking `_subscribersMutex` at every remote entry point, and doing nothing for a destroyed topic is the right behavior on the replication-apply path. Two adjustments:

1. The code comment should state the local invariant rather than the orphan mechanism, which is exactly what the analysis disproved (suggestion inline).
2. Since the fix has no observable effect, it shouldn't have a changelog entry — please drop the fragment (comment inline).

changelog.d/service-icestorm/5658.md
```diff
@@ -0,0 +1,2 @@
+- Fixed a race in replicated IceStorm where applying a replicated add-subscriber update concurrently with a topic
+  destroy could leave an orphaned subscriber record, resurrecting the destroyed topic the next time IceStorm restarted.
```
bernardnormier: Per the analysis in #5658, the symptom described here isn't reachable, so this change has no user-visible effect — please remove this fragment.

pepone: 🤖 **Claude:** Agreed — dropped the fragment. Verified your analysis: the master holds `_subscribersMutex` across the synchronous `Observers::addSubscriber` (`future.get()` per replica) and `destroy` needs that same mutex, so add and destroy are never concurrent on a replica and the guard has no observable effect.

cpp/src/IceStorm/TopicI.cpp
pepone: 🤖 **Claude:** Applied your suggestion. Reworded to state the local invariant; the guard stays as defensive code matching the `_destroyed`-after-lock idiom at the other remote entry points.

bernardnormier (approved):

Issue #5661, opened 2026-06-22
DataStorm: partial-update base (`_lastByKey`) coupled to history — unbounded growth and silent sampleCount=0 corruption


pepone:
> **AI-generated finding** — identified by Claude (Anthropic), model **Claude Opus 4.8** (`claude-opus-4-8`) while reviewing #5602. Every code-level claim below was verified against the source with file:line evidence, and cross-checked by **OpenAI Codex (GPT-5)**, which confirmed the core diagnosis and contributed three refinements (folded in below; full verdict in the comments). **Not yet triaged by a human** — verify the reasoning and severity before acting.

## Summary

#5602 (fixing #5473) replaces the single per-element `_last` sample in `DataReaderI`/`DataWriterI` with a per-key map `_lastByKey` so partial updates resolve against the previous sample of the *same* key. That correctness fix is right, but it inherits a **structural coupling**: `_lastByKey` — the *base* a partial update merges onto — is tied to history retention. It is updated only on the code path that keeps history, and is never pruned. That single coupling produces two distinct problems.

## Symptom 1 — unbounded growth

`_lastByKey` entries are only inserted/updated (`DataReaderI::queue`, `DataWriterI::publish`, and the `initSamples` rebuild), never `.erase()`/`.clear()`. Even when `_samples` *is* bounded (by a positive `sampleCount`, `sampleLifetime`, or `clearHistory`), `_lastByKey` is not — it retains a key's last sample after that key's samples have aged out of `_samples`. (`_samples` is not always bounded either: an unset or negative `sampleCount` with no lifetime and a non-clearing history grows too. But `_lastByKey` grows *independently* of all of that.)

- Single-key elements (`KeyDataReaderI`/`KeyDataWriterI`, one key): ≤1 entry — fine.
- Filtered/multi-key elements on a **bounded** key space: bounded — fine.
- Filtered/multi-key elements on an **unbounded/churning** key space (transient per-session/per-request keys): **unbounded memory growth** — one retained `Sample` + `Key` per distinct key ever seen, for the element's lifetime. The old single `_last` was O(1).

## Symptom 2 — `sampleCount == 0` silently breaks partial updates

[...]

PR #5809, opened 2026-07-06
Fix DataStorm partial-update crash when the receiver has no base for a key

pepone:
Fixes Symptom 2 of #5661 — a partial update crashing (or silently losing state) when the receiver has no base for
the key it applies to. The unbounded-growth concern (Symptom 1) remains tracked in #5661, retargeted to 3.9.0.

DataStorm resolves a partial update against the previous full value of the *same* key, held in the per-key base map
`_lastByKey`. That base is independent of history retention. When it is missing, the updater receives a
default-constructed value — a null `shared_ptr` for a class-typed value — and the user updater dereferences it →
crash; for scalar/struct values the fields not carried by the update are silently reset to their defaults.

This PR closes the cases where the receiver could be left without a base:

- **`sampleCount == 0` (keep no history).** `DataReaderI::queue`, `DataWriterI::publish`, and
  `DataReaderI::initSamples` recorded `_lastByKey` only *after* the `sampleCount == 0` early-return, so such an
  element never stored the base. They now record it before the early-return.
- **Late joiners of a writer that keeps no history, or whose history aged out (`sampleLifetime`).** The writer now
  bootstraps a late-joining reader from the per-key base (which is not subject to history retention).
- **Any-key / filtered readers that receive history for only some of a writer's keys.** An any-key or filtered
  reader fetches all keys in a single request. When a key's full value is not in the returned history — e.g. with
  the default `ClearHistory = OnAll`, which keeps only the newest sample overall — `KeyDataWriterI::getSamples` now
  appends the current base for every key the history does not already cover, instead of only bootstrapping when the
  whole response is empty.

Supporting changes in `KeyDataWriterI::getSamples` / `DataReaderI::initSamples` / `SessionI::subscriberInitialized`:

- **The init batch keeps a base for every key, even when the writer has more distinct keys than the reader's
  `sampleCount`.** `getSamples` reserves the newest sample of each key first, then fills any remaining `sampleCount`
  budget with the newest of the other samples; `initSamples` records these per-key bases in `_lastByKey` before
  capping its own visible history back to `sampleCount`, so a batch larger than `sampleCount` can't overflow the
  reader. A naive "keep the newest N overall" cap could evict a key's only base whenever another key had more
  history samples, re-introducing the null-base crash.
- The earliest delivered sample of each key is resolved to a full `Update` (using that sample's own resolved value)
  so the reader always has a full base before any later partial for that key — generalizing the previous
  single-sample transform to per-key.
- `subscriberInitialized` now advances the subscriber `lastId` monotonically and no longer asserts that each batch
  strictly follows the previous one: a multi-key subscriber shares one `ElementSubscriber` and is acked one batch
  per key with ids interleaved across keys.

The base is the already-resolved sample and is independent of `_samples`, so history queries and notifications are
unaffected; single-key elements still hold at most one base entry.

One handshake direction is not yet fully covered: per-key init batches delivered via the `initSamples` / `i()` path
are still dropped after the first batch per element (#5819), so for a multi-key keyed writer the bootstrap isn't
guaranteed in that direction until #5819 is fixed. The remaining null-base path — a partial delivered under a
discard policy that left no base (#5824) — ends at the same call site and is left as a follow-up; a reader-side
guard that drops (and traces) a baseless partial would turn it from a crash into degraded delivery.

### Tests

Adds partial-update cases to `test/DataStorm/partial`, each verified red→green:

- a `sampleCount == 0` reader (observed via `onSamples`, since it keeps no history);
- a late joiner of a no-history writer, and one whose base aged out via `sampleLifetime`;
- a late any-key reader after the writer trimmed a key's history (`ClearHistory = OnAll`), which SIGSEGVs without
  the fix;
- a `sampleCount`-limited any-key reader against a writer with more distinct keys than `sampleCount`, verifying the
  writer delivers a base for every key and the reader caps its own history without evicting any base — without the
  fix the reader SIGSEGVs on the surplus key's partial, or aborts on the `initSamples` cap assert in a debug build.

The full DataStorm C++ suite passes. A changelog fragment is included under `changelog.d/datastorm/`.

### DIFF
--- a/changelog.d/datastorm/5661.md
+++ b/changelog.d/datastorm/5661.md
@@ -1,5 +1,7 @@
 - Fixed a bug in DataStorm where a partial update was not merged with the key's previous value when the receiver did
-  not already have that value: an element configured with `sampleCount = 0` (keep no history), or a reader that joins
-  after the value was published to a writer that keeps no history or whose history has aged out (`sampleLifetime`).
-  Applications using class-typed values typically crashed; with other value types, the fields not carried by the
-  update were silently reset to their defaults.
+  not already have that value: an element configured with `sampleCount = 0` (keep no history), a reader that joins
+  after the value was published to a writer that keeps no history or whose history has aged out (`sampleLifetime`),
+  or an any-key/filtered reader that receives history for only some of a writer's keys (for example with
+  `ClearHistory = OnAll`). The current per-key value is now bootstrapped for every such key. Applications using
+  class-typed values typically crashed; with other value types, the fields not carried by the update were silently
+  reset to their defaults.
--- a/changelog.d/datastorm/5661.md
+++ b/changelog.d/datastorm/5661.md
@@ -1,3 +1,4 @@
-- Fixed a crash in DataStorm when partial updates were used with `sampleCount = 0` (keep no history): the previous
-  sample used as the base for merging the update was never recorded, so the partial update resolved against a null
-  base. The per-key base is now maintained independently of history retention.
+- Fixed a bug in DataStorm where a partial update was not merged with the key's previous value when the
+  reader or writer was configured with `sampleCount = 0` (keep no history). Applications using class-typed
+  values typically crashed; with other value types, the fields not carried by the update were silently
+  reset to their defaults.
--- a/changelog.d/datastorm/5661.md
+++ b/changelog.d/datastorm/5661.md
@@ -1,4 +1,5 @@
-- Fixed a bug in DataStorm where a partial update was not merged with the key's previous value when the
-  reader or writer was configured with `sampleCount = 0` (keep no history). Applications using class-typed
-  values typically crashed; with other value types, the fields not carried by the update were silently
-  reset to their defaults.
+- Fixed a bug in DataStorm where a partial update was not merged with the key's previous value when the receiver did
+  not already have that value: an element configured with `sampleCount = 0` (keep no history), or a reader that joins
+  after the value was published to a writer that keeps no history or whose history has aged out (`sampleLifetime`).
+  Applications using class-typed values typically crashed; with other value types, the fields not carried by the
+  update were silently reset to their defaults.
--- a/changelog.d/datastorm/5661.md
+++ b/changelog.d/datastorm/5661.md
@@ -1,0 +1,3 @@
+- Fixed a crash in DataStorm when partial updates were used with `sampleCount = 0` (keep no history): the previous
+  sample used as the base for merging the update was never recorded, so the partial update resolved against a null
+  base. The per-key base is now maintained independently of history retention.
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -1286,9 +1286,45 @@ KeyDataWriterI::getSamples(
     DataSamples samples;
     samples.id = _keys.empty() ? -_id : _id;
 
-    // If the caller sample queueing is disabled, there is no need to return any samples.
+    // Append the current per-key base value(s) so a reader with no other init sample can still resolve subsequent
+    // partial updates. A removed key has no value and is skipped; a partial base is sent as a full Update. A single-key
+    // writer stores its base under a null publish key, so the key is matched on the sample rather than the map key.
+    auto appendBaseSamples = [&]
+    {
+        vector<DataSample> bases;
+        for (const auto& [baseKey, baseSample] : _lastByKey)
+        {
+            if (baseSample->id <= lastId || !baseSample->hasValue() || (key && key != baseSample->key) ||
+                (sampleFilter && !sampleFilter->match(baseSample)))
+            {
+                continue;
+            }
+
+            DataSample ds = toSample(baseSample, getCommunicator(), _keys.empty());
+            if (baseSample->event == DataStorm::SampleEvent::PartialUpdate)
+            {
+                ds.tag = 0;
+                ds.event = DataStorm::SampleEvent::Update;
+                ds.value = baseSample->encodeValue(getCommunicator());
+            }
+            bases.push_back(std::move(ds));
+        }
+
+        // _lastByKey is ordered by key (pointer), not by sample id. The subscriber records the last delivered
+        // sample's id as its lastId, so deliver the bases in ascending id order to keep the newest last; otherwise a
+        // higher-id base could be re-sent as a duplicate on the next resync.
+        sort(bases.begin(), bases.end(), [](const DataSample& lhs, const DataSample& rhs) { return lhs.id < rhs.id; });
+        for (auto& base : bases)
+        {
+            samples.samples.push_back(std::move(base));
+        }
+    };
+
+    // A reader that keeps no history still needs the current per-key base to resolve partial updates. Send just the
+    // base, not the history.
     if (config->sampleCount && *config->sampleCount == 0)
     {
+        appendBaseSamples();
         return samples;
     }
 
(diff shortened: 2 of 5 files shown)


changelog.d/datastorm/5661.md
```diff
@@ -0,0 +1,3 @@
+- Fixed a crash in DataStorm when partial updates were used with `sampleCount = 0` (keep no history): the previous
+  sample used as the base for merging the update was never recorded, so the partial update resolved against a null
+  base. The per-key base is now maintained independently of history retention.
```
bernardnormier:
```suggestion
- Fixed a bug in DataStorm where a partial update was not merged with the key's previous value when the
  reader or writer was configured with `sampleCount = 0` (keep no history). Applications using class-typed
  values typically crashed; with other value types, the fields not carried by the update were silently
  reset to their defaults.
```

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -816,6 +816,11 @@ DataReaderI::initSamples(
     }
     _lastSendTime = valid.back()->timestamp;
 
+    // Record the resolved per-key bases before the sampleCount == 0 early-return below: the base is independent of
+    // history retention, so a reader that keeps no history must still resolve future partial updates against these
+    // values. valid is non-empty here (checked above), matching the previous placement of this assignment.
```
bernardnormier:
```suggestion
    // The per-key bases for partial updates are maintained even when the element keeps no history (sampleCount == 0).
```

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -935,6 +939,11 @@ DataReaderI::queue(
     }
     _lastSendTime = sample->timestamp;
 
+    // Record the sample as the per-key base for resolving future partial updates. This must happen before the
+    // sampleCount == 0 early-return below: the base is independent of history retention, so a reader that keeps no
+    // history (sampleCount == 0) must still resolve a subsequent partial update against this value.
```
bernardnormier:
```suggestion
    // The per-key base for partial updates is maintained even when the element keeps no history (sampleCount == 0).
```

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -1065,6 +1073,10 @@ DataWriterI::publish(const shared_ptr<Key>& key, const shared_ptr<Sample>& sampl
     }
     send(key, sample);
 
+    // Record the sample as the per-key base for resolving future partial updates. This must happen before the
+    // sampleCount == 0 early-return below: the base is independent of history retention.
```
bernardnormier:
```suggestion
    // The per-key base for partial updates is maintained even when the element keeps no history (sampleCount == 0).
```

cpp/test/DataStorm/partial/Reader.cpp
```diff
@@ -113,6 +115,43 @@ void ::Reader::run(int argc, char* argv[])
         sample = reader.getNextUnread(); // from writer 2: attach runs getLastIds over the stale entry
         test(sample.getValue() == 2);
     }
+
+    // A sampleCount=0 reader keeps no history but must still resolve a partial update against the previous full
+    // sample of the same key. Before the fix the per-key base was recorded only on the history-keeping path, so with
+    // sampleCount=0 the partial resolved against a null base and the class-typed updater dereferenced it.
```
bernardnormier:
```suggestion
    // A sampleCount=0 reader keeps no history but must still resolve a partial update against the previous full
    // sample of the same key.
```

bernardnormier (approved):
LGTM. I verified the three moves are behavior-preserving for `sampleCount > 0` (nothing between the old and new positions reads `_lastByKey`/`previousByKey`, and the partial-update resolution sites sit above the new positions), that late joiners are unaffected (attach-time init samples come solely from `_samples`), and the red→green test choreography is race-free.

Suggestions below: a changelog reword (the crash is only the class-typed manifestation, and the mechanism details belong here in the PR, not the entry) and comment trims. If you want to keep the invariant documented in one place, the `_lastByKey` doc comment in `DataElementI.h` is the natural home for "maintained independently of history retention".

externl (commented):
The three `_lastByKey` moves look correct, but the crash isn't fully closed:

- A reader that attaches *after* the `add` still gets no base (`KeyDataWriterI::getSamples` returns no init samples when either side has `sampleCount = 0`) and crashes on the next partial.
- Pre-PR a `sampleCount = 0` writer crashed at its first `partialUpdate`; post-PR it forwards the partial, so late-joining readers crash instead. Consider dropping a partial that has no base rather than calling the updater with null, or tracking this as a follow-up.
- The test covers only the `queue` site; `publish` and `initSamples` are unpinned (Copilot flagged the writer case too).
- `sampleCount = 0` elements now pin one never-pruned sample per key — worth a note in #5661.

externl (commented):
The new bootstrap commits address all my earlier points — one gap remains in the cap:

- When a topic has more keys than the reader's `sampleCount`, `finalize`'s first pass exhausts the budget before every key gets a slot, so the surplus keys' bases are evicted after `collectBases` collected them — and a later partial on one of those keys hits the same null-base crash. The comment says the cap keeps "a resolvable base to merge later partial updates" per key, so intent and code disagree here. Exempting one base per key from the cap looks safe: the reader records `_lastByKey` before pruning `_samples`, so delivering #keys samples can't overflow its history. `capTopic` tests budget == #keys; a 3-keys/`sampleCount=2` case would pin it.

bernardnormier (approved):
Re-approving after the latest commits — the per-key bootstrap, the per-key cap, and the monotonic `lastId` all look correct to me (I traced the cap selection: the kept samples of each key always form a contiguous newest suffix, so partial chains never resolve against a stale base).

Two points worth addressing:

1. The description says "This PR closes every case where the receiver could be left without a base", but for a multi-key keyed writer the per-key init batches can also arrive via the `initSamples`/`i()` path, which still drops every batch after the first one per element (#5819). Until #5819 is fixed, the bootstrap isn't guaranteed in that handshake direction, so I'd soften the claim and reference #5819 from the description.

2. The residual null-base paths — the acknowledged `sampleCount < number of keys` limit and the discard-policy hole (#5824) — both end at the same call site: the updater invoked with a null base, which still segfaults for class-typed values. A small reader-side guard that drops (and traces) a partial with no base would turn these from a crash into degraded delivery. Fine as a follow-up rather than this PR, but worth deciding explicitly.

pepone:
🤖 **Claude:** Confirmed and fixed in 232f28c — the gap is real, and closing it needs a two-sided change: the one-base-per-key exemption alone trips a reader-side assert.

**Writer** (`finalize`): the first cap pass now reserves the newest sample of *every* key, even past `sampleCount`, so no key's base is evicted after `collectBases` bootstraps it.

**Reader** (`initSamples`): your intuition holds — `_lastByKey` is recorded before `_samples` is pruned, so the bases survive. But the `_samples` cap itself still assumed `batch ≤ sampleCount`: once the writer delivers `#keys > sampleCount`, `assert(_samples.size() + valid.size() == maxCount)` fires (or `_samples` silently overflows under `NDEBUG`). Replaced it with a post-batch trim to `sampleCount`.

**Test**: extended `capTopic` to 3-keys/`sampleCount=2` as you suggested, partial-updating the key a naive "newest N" cap drops. Verified red→green both ways — writer fix but no reader fix → the reader aborts on that assert (`line 841`); both fixes → full DataStorm suite green. Whole chain cross-checked with codex (C1–C5 confirmed).

externl (approved):
Verified 232f28c closes the cap-starvation gap: the first pass now reserves the newest sample of every key unbounded, and the reader-side post-trim (replacing the pre-trim whose assert an oversized batch would have tripped) makes the larger batch safe; the 3-key/sampleCount=2 test pins it. Residual paths are tracked in #5819/#5824. Note clang-format wants the trailing comments realigned in Writer.cpp.

PR #5812, opened 2026-07-06
Fix IceGrid property-set diamond rejected as a circular dependency

### LINKED ISSUE #5662: IceGrid: property-set circular-dependency check over-rejects diamond references (getProperties)
> **AI-generated finding** — identified by Claude (Anthropic), model **Claude Opus 4.8** (`claude-opus-4-8`) while reviewing #5643. Verified against the source and `git` history (file:line + commits below). **Low severity** — a ~19-year-old over-rejection with a narrow trigger and no known reports. **Not yet triaged by a human** — verify before acting.

## Summary

`Resolver::getProperties(const Ice::StringSeq&, set<string>& resolved)` in `cpp/src/IceGrid/DescriptorHelper.cpp` detects circular property-set references with a **visited-only** set: it inserts each reference into `resolved` before recursing and **never erases** it. So it conflates "seen anywhere during this resolution" with "on the current path," and **over-rejects a legitimate diamond** — a non-leaf property set reachable via two independent parents — as a circular dependency.

This is the inverse of the correct, path-based **variable** cycle check just added in #5643 (insert before recursing, `erase` afterward).

## The code

```cpp
// cpp/src/IceGrid/DescriptorHelper.cpp:932
PropertyDescriptorSeq
Resolver::getProperties(const Ice::StringSeq& references, set<string>& resolved) const
{
    PropertyDescriptorSeq properties;
    for (const auto& reference : references)
    {
        if (resolved.find(reference) != resolved.end())
        {
            exception("detected circular dependency with property reference '" + reference + "'");
        }
        PropertySetDescriptor desc = getPropertySet(reference);
        if (!desc.references.empty())
        {
            resolved.insert(reference);                            // inserted, never erased
            PropertyDescriptorSeq q = getProperties(desc.references, resolved);
            properties.insert(properties.end(), q.begin(), q.end());
        }
        // ... append desc.properties ...
    }
    return properties;
}
```

## Repro — a valid DAG that is wrongly rejected

Property sets: `A → {B, C}`, `B → {D}`, `C → {D}`, `D → {E}` (D is a shared **non-leaf**).
Resolving `A`: `B`'s subtree inserts `D` into `resolved`; then `C`'s subtree sees `D` still present → `detected circular dependency with property reference 'D'`. But `A→B→D` / `A→C→D` is a legitimate diamond, not a cycle.

Note the `resolved.insert` only fires for a property set that has its **own** references, so a shared **leaf** property set is unaffected — the trigger requires the shared node to itself reference other property sets.

## Fix

Restore on-stack semantics by erasing after the recursive call (identical to #5643's variable check):

```cpp
if (!desc.references.empty())
{
    resolved.insert(reference);
    PropertyDescriptorSeq q = getProperties(desc.references, resolved);
    resolved.erase(reference);          // only references on the current path remain in `resolved`
    properties.insert(properties.end(), q.begin(), q.end());
}
```

Then `resolved` holds only the active resolution path: true cycles are still caught, diamonds resolve.

## History — not intentional

- Introduced 2006 by Benoit Foucher in `707df022c7` ("Added support for property sets") — visited-only from day one.
- The 2007 commit `abada90e3f8` (attributed to Bernard Normier in blame) is a tabs→spaces reindentation; the logic is untouched.
- Every change since is cosmetic (range-for modernization, etc.).

So the visited-only behavior is the original simplification, not a deliberate "a property set may only be referenced once" rule — there is no such comment or enforcement anywhere. It has simply never hit the diamond case in ~19 years.

## Severity

Low. A real over-rejection of a valid configuration, but a narrow trigger (a non-leaf property set shared by independent parents) and no known reports in ~19 years. Worth fixing to align with #5643's correct check; not urgent.

## Notes

- Surfaced during review of #5643 (which fixes the analogous variable cycle check). Filing separately so #5643 lands unblocked.
- A diamond regression test (two parents referencing a shared non-leaf property set) is straightforward to add.

pepone:
Fixes #5662.

`Resolver::getProperties` in `cpp/src/IceGrid/DescriptorHelper.cpp` tracked resolved property-set references in a
visited-only set: it inserted each reference before recursing but never erased it, conflating "seen anywhere during
this resolution" with "on the current path". A non-leaf property set reached through two independent parents (a
diamond) — or simply referenced more than once — was therefore wrongly rejected as a circular dependency.

Example: property sets `A → {B, C}`, `B → {D}`, `C → {D}`, `D → {E}`. Resolving `A` inserts `D` while walking
`B`'s subtree and never removes it, so `C`'s subtree sees `D` still present and throws
`detected circular dependency with property reference 'D'` — even though `A→B→D` / `A→C→D` is a legitimate DAG.

The fix erases the reference after its subtree is resolved, so `resolved` holds only the active resolution path. True
cycles are still detected; diamonds and repeated references resolve. This mirrors the variable cycle check in
`substitute()` added in #5643.

### Test

Adds a diamond regression test to `IceGrid/update` (`A → {B, C}`, `B → {D}`, `C → {D}`, `D → {E}`). Verified
red→green: without the fix, deploying the application throws `detected circular dependency with property reference 'D'`
(the regression fails at `AllTests.cpp` `test(false)`); with the fix the application deploys and the suite passes.

A changelog fragment is included under `changelog.d/service-icegrid/`.

### DIFF
--- a/changelog.d/service-icegrid/5662.md
+++ b/changelog.d/service-icegrid/5662.md
@@ -1,2 +1,2 @@
-- Fixed the IceGrid registry to accept a property set referenced through two independent parents (a diamond)
-  instead of incorrectly rejecting it as a circular dependency.
+- Fixed a bug where the IceGrid registry incorrectly rejected an application with a "circular dependency" error
+  when a property set was referenced by two other property sets.
--- a/changelog.d/service-icegrid/5662.md
+++ b/changelog.d/service-icegrid/5662.md
@@ -1,0 +1,2 @@
+- Fixed the IceGrid registry to accept a property set referenced through two independent parents (a diamond)
+  instead of incorrectly rejecting it as a circular dependency.
--- a/cpp/src/IceGrid/DescriptorHelper.cpp
+++ b/cpp/src/IceGrid/DescriptorHelper.cpp
@@ -954,9 +954,8 @@ Resolver::getProperties(const Ice::StringSeq& references, set<string>& resolved)
         PropertySetDescriptor desc = getPropertySet(reference);
         if (!desc.references.empty())
         {
-            // Track references only for the current resolution path so a shared non-leaf property set reached through
-            // two independent parents (a diamond) is not mistaken for a cycle. Erasing after recursion mirrors the
-            // variable cycle check in substitute().
+            // 'resolved' tracks only the current resolution path: a property set reached twice through independent
+            // parents is not a cycle.
             resolved.insert(reference);
             PropertyDescriptorSeq q = getProperties(desc.references, resolved);
             resolved.erase(reference);
--- a/cpp/src/IceGrid/DescriptorHelper.cpp
+++ b/cpp/src/IceGrid/DescriptorHelper.cpp
@@ -954,8 +954,12 @@ Resolver::getProperties(const Ice::StringSeq& references, set<string>& resolved)
         PropertySetDescriptor desc = getPropertySet(reference);
         if (!desc.references.empty())
         {
+            // Track references only for the current resolution path so a shared non-leaf property set reached through
+            // two independent parents (a diamond) is not mistaken for a cycle. Erasing after recursion mirrors the
+            // variable cycle check in substitute().
             resolved.insert(reference);
             PropertyDescriptorSeq q = getProperties(desc.references, resolved);
+            resolved.erase(reference);
             properties.insert(properties.end(), q.begin(), q.end());
         }
 
--- a/cpp/test/IceGrid/update/AllTests.cpp
+++ b/cpp/test/IceGrid/update/AllTests.cpp
@@ -686,6 +686,37 @@ allTests(Test::TestHelper* helper)
     }
 
     {
+        cout << "testing property set cycle detection... " << flush;
+
+        //
+        // A genuine property set reference cycle (B -> {C}, C -> {B}) must still be rejected as a circular
+        // dependency.
+        //
+        ApplicationDescriptor app;
+        app.name = "PropertySetCycleApp";
+
+        PropertySetDescriptor b;
+        b.references = {"C"};
+        app.propertySets["B"] = b;
+
+        PropertySetDescriptor c;
+        c.references = {"B"};
+        app.propertySets["C"] = c;
+
+        try
+        {
+            admin->addApplication(app);
+            test(false);
+        }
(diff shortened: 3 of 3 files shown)


cpp/src/IceGrid/DescriptorHelper.cpp
```diff
@@ -954,8 +954,12 @@ Resolver::getProperties(const Ice::StringSeq& references, set<string>& resolved)
         PropertySetDescriptor desc = getPropertySet(reference);
         if (!desc.references.empty())
         {
+            // Track references only for the current resolution path so a shared non-leaf property set reached through
+            // two independent parents (a diamond) is not mistaken for a cycle. Erasing after recursion mirrors the
+            // variable cycle check in substitute().
```
bernardnormier: The last clause cross-references `substitute()` (it can go stale independently of this loop) and narrates the change rather than the invariant. Suggestion:
```suggestion
            // 'resolved' tracks only the current resolution path: a property set reached twice through independent
            // parents is not a cycle.
```

changelog.d/service-icegrid/5662.md
```diff
@@ -0,0 +1,2 @@
+- Fixed the IceGrid registry to accept a property set referenced through two independent parents (a diamond)
+  instead of incorrectly rejecting it as a circular dependency.
```
bernardnormier: "Referenced through two independent parents (a diamond)" is graph vocabulary; the user-level story is the `<properties refid="..."/>` references and the error they saw:
```suggestion
- Fixed a bug where the IceGrid registry incorrectly rejected an application with a "circular dependency" error
  when a property set was referenced by two other property sets.
```

bernardnormier (approved):
LGTM. I verified the fix: the cycle check runs before the insert, so with the erase `resolved` holds exactly the open resolution path — true cycles and self-references still throw, every entry point builds a fresh set (no stale-state hazard when a cycle throws past the erase), and this is the only property-reference resolution site. Diamond duplicate emission is benign for the deployed config since `toPropertyDict` is last-wins over identical subsequences.

One ask: could you add a small negative case next to the diamond test asserting a genuine property-set reference cycle (e.g. `B -> {C}`, `C -> {B}`) is still rejected? The suite's only "circular dependency" assertion covers variable cycles, so nothing currently pins the invariant this fix touches — a misplaced erase would go undetected.

Two suggestions inline (comment wording, changelog wording).

externl (approved):

PR #5813, opened 2026-07-06
Treat Ice.ClassGraphDepthMax < 1 as unlimited in Swift

externl:
Swift was the only mapping that rejected `Ice.ClassGraphDepthMax` values less than 1, aborting communicator initialization with a `fatalError` where every other mapping treats such values as "unlimited".

- Coerce `Ice.ClassGraphDepthMax < 1` to `Int32.max`, matching C++, C#, Java, JS, and MATLAB.
- Removed the now-redundant `precondition` (the upper-bound check was a no-op since the property is already an `Int32`, and its message incorrectly said `>= 0`).
- Added a changelog fragment.

Fixes zeroc-ice/ice-audit#102.

### DIFF
--- a/changelog.d/swift/class-graph-depth-max-unlimited.md
+++ b/changelog.d/swift/class-graph-depth-max-unlimited.md
@@ -1,2 +1,2 @@
 - Fixed communicator initialization to treat `Ice.ClassGraphDepthMax` values less than 1 as unlimited, matching the
-  other language mappings. Previously, setting this property to 0 aborted the program with a `fatalError`.
+  other language mappings. Previously, setting this property to 0 or a negative value aborted the program.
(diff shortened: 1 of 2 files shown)


pepone (approved):

changelog.d/swift/class-graph-depth-max-unlimited.md
```diff
@@ -0,0 +1,2 @@
+- Fixed communicator initialization to treat `Ice.ClassGraphDepthMax` values less than 1 as unlimited, matching the
+  other language mappings. Previously, setting this property to 0 aborted the program with a `fatalError`.
```
bernardnormier:
```suggestion
  other language mappings. Previously, setting this property to 0 or a negative value aborted the program.
```

bernardnormier (approved):
LGTM. I verified the cross-mapping claim in the source: C++, C#, Java, JS, and MATLAB all coerce `Ice.ClassGraphDepthMax < 1` to int32 max, and Python/Ruby/PHP inherit the C++ coercion — Swift was the only mapping that aborted. Every Swift consumer of `classGraphDepthMax` is a plain comparison, so `Int32.max` is safely "unlimited", and the removed upper-bound check was indeed vacuous on an `Int32`.

One suggestion on the changelog fragment: the abort came from the `precondition` trap rather than a `fatalError` call, and negative values tripped it too.

PR #5814, opened 2026-07-06
Don't reuse a destroyed lingering TopicImpl in IceStorm observerInit

### LINKED ISSUE #5810: IceStorm: observerInit reuses a destroyed lingering TopicImpl without clearing _destroyed, leading to topic resurrection on a rejoined replica
### Summary

In a replicated deployment, `TopicManagerImpl::observerInit` can reuse a lingering, already-destroyed `TopicImpl` for a topic that is alive group-wide: it rewrites the database correctly but calls `TopicImpl::update()`, which never clears `_destroyed` (and doesn't re-register the servants that `destroyInternal` removed). The node then holds a live, correctly-synced topic backed by an in-memory `TopicImpl` still flagged destroyed — and the pre-existing `_destroyed` early return in `TopicImpl::observerDestroyTopic` turns a later replicated destroy of that topic into a silent no-op that leaks the topic's records, resurrecting the destroyed topic on this node at the next restart.

### Affected code (main@`b71bc8ed0a`)

- [`TopicImpl::destroy` sets `_destroyed`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0a917629757db1c9c53f0f88fac41f5f/cpp/src/IceStorm/TopicI.cpp#L794) but leaves the impl in `_topics`: removal is deferred to the [lazy reaper](https://github.com/zeroc-ice/ice/blob/b71bc8ed0a917629757db1c9c53f0f88fac41f5f/cpp/src/IceStorm/TopicI.cpp#L1223), and [`reap()`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0a917629757db1c9c53f0f88fac41f5f/cpp/src/IceStorm/TopicManagerI.cpp#L693-L707) only runs from `create`/`retrieve`/`retrieveAll`/`getContent`/`initMaster`.
- [`destroyInternal` also unregisters the topic's servants](https://github.com/zeroc-ice/ice/blob/b71bc8ed0a917629757db1c9c53f0f88fac41f5f/cpp/src/IceStorm/TopicI.cpp#L1217-L1234) (link/publisher/topic) and clears `_servant`.
- [`observerInit`'s "topic exists in both" branch reuses the lingering impl via `update()`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0a917629757db1c9c53f0f88fac41f5f/cpp/src/IceStorm/TopicManagerI.cpp#L472-L486); [`TopicImpl::update`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0a917629757db1c9c53f0f88fac41f5f/cpp/src/IceStorm/TopicI.cpp#L830-L884) rebuilds `_subscribers` only — it neither clears `_destroyed` nor re-registers the servants.
- [`TopicImpl::observerDestroyTopic` returns early when `_destroyed`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0a917629757db1c9c53f0f88fac41f5f/cpp/src/IceStorm/TopicI.cpp#L1123-L1142), skipping `destroyInternal` — i.e. skipping the deletion of the topic's records and the LLU update — while [`TopicManagerImpl::observerDestroyTopic` erases the topic from `_topics` regardless](https://github.com/zeroc-ice/ice/blob/b71bc8ed0a917629757db1c9c53f0f88fac41f5f/cpp/src/IceStorm/TopicManagerI.cpp#L526-L539).

### Mechanism

Formation of the zombie (3 nodes; M1 master at generation g, S1/S2 slaves):

1. A client destroys topic T on M1. `TopicImpl::destroy` sets `_destroyed`, commits the local erase + LLU, unregisters T's servants, and adds T to the reaper — then calls `observers()->destroyTopic`. The network fails at exactly this point: both replica invocations fail, both replicas are [erased from the observer set](https://github.com/zeroc-ice/ice/blob/b71bc8ed0a917629757db1c9c53f0f88fac41f5f/cpp/src/IceStorm/Observers.cpp#L166-L191), majority is lost and the call throws. T is destroyed on M1 only; the destroyed `TopicImpl` stays in M1's `_topics`.
2. M1 recovers into an inactive state. Client operations block while the node is inactive, so none of the `reap()` call sites run — the zombie survives.
3. S1/S2 detect the failure and elect a new master; T is still alive in their state (the destroy never reached them).
4. The partition heals and M1 rejoins as a replica. The new master's LLU wins, so M1 receives `observerInit`: its database is rewritten from the group content — including T's placeholder and records, which is correct — but T is found in `_topics` and the lingering impl is reused via `update()`. `_destroyed` remains `true`.

Consequences on the rejoined node:

1. **Replicated destroy → records leak → resurrection on restart.** If the group later destroys T, `TopicManagerImpl::observerDestroyTopic` invokes the zombie's `observerDestroyTopic`, which early-returns on `_destroyed` — T's placeholder and subscriber records are never deleted and the destroy's LLU is never persisted — and then erases T from `_topics` anyway. The database still contains T, so the next restart of this node reloads the topic the group destroyed. The node's LLU is also left stale, affecting the next election/sync.
2. **Unreachable servants.** The original `destroyInternal` removed T's topic/publisher/link servants and `update()` doesn't restore them, so T is not reachable via this node's adapters even though the node's database says it exists.
3. A `_destroyed` guard in `observerAddSubscriber` (proposed in #5808) would extend the exposure to the add path: a replicated add applied to the zombie would be silently skipped (record and LLU) while the master receives a success reply, silently diverging the replica.

### Impact

Replica/HA deployments only, and the formation window requires a destroy whose replication fails entirely (e.g. partition at that instant) followed by the node rejoining while the topic is still alive group-wide. Consequences are persistent (database keeps a destroyed topic across restarts) but the trigger is narrow. Severity: Low/Medium.

### Suggested fix

In `observerInit`'s "exists in both" branch, don't reuse a `TopicImpl` whose `_destroyed` is set: erase it and install a fresh impl from the content (`installTopic` already builds everything from the records). Alternatively `update()` could clear `_destroyed` and re-register the servants `destroyInternal` removed — but erase-and-reinstall looks simpler and keeps `_destroyed` monotonic per impl.

### Provenance

Found while triaging #5658 (see [the reachability analysis](https://github.com/zeroc-ice/ice/issues/5658#issuecomment-4893799894)) and reviewing #5808. Analysis by Claude (Fable 5); the zombie-formation chain was adversarially verified against `main`, the replicated-destroy consequence traced but not independently re-derived. Static analysis only — no reproduction attempted.

bernardnormier:
Fixes #5810.

## The bug

`TopicImpl::destroy` marks the impl destroyed and tears it down (deletes the topic's database records, unregisters its topic/publisher/link servants), but leaves it in `TopicManagerImpl::_topics` until the lazy reaper runs. If the destroy's replication fails entirely (e.g. a partition at that instant) and the node later rejoins as a slave while the topic is still alive group-wide, `observerInit`'s "topic exists in both" branch reused that destroyed remnant via `update()` — which neither clears `_destroyed` nor re-registers the servants. See #5810 for the full mechanism.

On the rejoined node the topic was then unreachable (no servants), and a later replicated destroy of the topic hit the `_destroyed` early return in `TopicImpl::observerDestroyTopic`, silently skipping the record deletion and LLU update: the node's database kept the destroyed topic and resurrected it at the next restart.

## The fix

In `observerInit`, when the local impl for a topic present in the group content is destroyed, erase it and install a fresh impl from the synced content (`installTopic`) instead of reusing it via `update()`. This keeps `_destroyed` monotonic per impl.

This is safe:

- A destroyed impl has already had its three servants (topic, publisher, link) removed — `destroyInternal` runs exactly once, guarded by `_destroyed` — so the reinstall cannot hit `AlreadyRegisteredException`.
- No database work is needed here: `observerInit` already rewrote the subscriber records and LLU before this loop.
- The stale reaper entry queued by the original destroy is harmless: `reap()` re-checks `destroyed()` on the current map entry, so it won't discard the fresh impl.
- The other `observerInit` scan (topic absent from the group content) already handles a destroyed impl correctly: `observerDestroyTopic` early-returns and the erase is all that's needed, since the database was already synced.

`initMaster` and `getContent` both call `reap()` before using `_topics`, so this was the only place a destroyed impl could be reused.

## Why no regression test

Reproducing this from the test harness is impractical, because the zombie exists only in the old master's memory — its database no longer contains the topic — so the formation sequence requires the old master's **process to survive** from the failed destroy all the way through rejoining as a slave. The harness's only partition primitive (stopping/starting replica processes) can't express that:

- Stopping both slaves makes the destroy's replication fail and forms the zombie, but the moment the slaves restart, the old master is reachable again and joins the election. The failed destroy incremented its LLU, so its state wins, and `initMaster` calls `reap()` first — which discards the zombie before it can be reused. The bug needs the old master's LLU to *lose*, i.e. the slaves must advance the generation while the old master is alive but unreachable — a real network partition.
- Restarting the old master instead clears the zombie trivially, since `_topics` is rebuilt from its database (which no longer has the topic).

The closest emulation would be `SIGSTOP`/`SIGCONT` on the master process so it stays alive but unresponsive while the slaves elect at a higher generation. That is POSIX-only, and because the kernel still completes TCP handshakes for a stopped process, every failure manifests as a timeout rather than a fast connection refusal — making the test slow and timing-sensitive. The observable consequence (leaked records after a later group-wide destroy) is also only visible via an offline database dump or another restart. That combination is a recipe for a CI flake, so this fix ships without an automated reproduction; it was verified by review of the affected paths plus a clean `IceStorm/rep1` run, which exercises `observerInit` on every replica rejoin.

### DIFF
--- a/changelog.d/service-icestorm/replica-zombie-topic-resurrection.md
+++ b/changelog.d/service-icestorm/replica-zombie-topic-resurrection.md
@@ -1,0 +1,3 @@
+- Fixed a replication bug where a replica rejoining the group could reuse the in-memory remnant of a topic it
+  had destroyed locally. The rejoined replica then ignored a later replicated destroy of that topic, keeping the
+  topic's records in its database and resurrecting the destroyed topic on restart.
--- a/cpp/src/IceStorm/TopicManagerI.cpp
+++ b/cpp/src/IceStorm/TopicManagerI.cpp
@@ -479,6 +479,14 @@ TopicManagerImpl::observerInit(const LogUpdate& llu, const TopicContentSeq& cont
         {
             installTopic(name, q.id, true, q.records);
         }
+        else if (r->second->destroyed())
+        {
+            // A topic destroyed on this replica can linger in _topics until the next reap. Since a destroyed
+            // TopicImpl has no servants and ignores replicated updates, we can't reuse it: install a fresh
+            // topic from the content instead.
+            _topics.erase(r);
+            installTopic(name, q.id, true, q.records);
+        }
         else
         {
             r->second->update(q.records);


pepone (approved):
LGTM.

The locking concern checks out: every `_topics` access in `TopicManagerImpl` is under `_mutex`, and the new `erase` + `installTopic` run inside the `observerInit` lock. `r->second->destroyed()` under `_mutex` is the same call `reap()` already makes, so it adds no new pattern and keeps the in-process order `_mutex → _subscribersMutex`.

Reinstall is safe: `destroyInternal` runs once (guarded by `_destroyed`) and frees the topic/publisher/link identities, so the fresh `installTopic` can't hit `AlreadyRegisteredException`, and the leftover reaper entry is harmless since `reap()` re-checks `destroyed()` on the current map entry.

Agree on skipping the test — the SIGSTOP/SIGCONT partition emulation would just flake in CI.

externl (approved):

PR #5815, opened 2026-07-06
Validate non-primitive sequence count before allocating in IcePy/IceRuby/IcePHP

externl:
IcePy, IceRuby, and IcePHP read a non-primitive sequence's element count with a bare `readSize()` and act on it immediately — IcePy/IceRuby pre-allocate a container of that many slots, IcePHP loops that many times. Since `readSize()` only rejects negatives, a peer can declare a count near `INT32_MAX` via the 5-byte size escape and trigger a multi-GB allocation from a tiny message. The primitive path and the C++/C#/Java readers already guard this via `readAndCheckSeqSize`.

This routes the non-primitive path through `readAndCheckSeqSize(elementType->wireSize())`, so a count that can't fit the buffer throws `MarshalException` up front. Dictionaries were already safe (empty map, insert per entry) and are unchanged.

Fixes zeroc-ice/ice-audit#98.

### DIFF
--- a/php/src/Types.cpp
+++ b/php/src/Types.cpp
@@ -1538,7 +1538,7 @@ IcePHP::SequenceInfo::unmarshal(
     array_init(&zv);
     AutoDestroy destroy(&zv);
 
-    int32_t sz = is->readSize();
+    int32_t sz = is->readAndCheckSeqSize(elementType->wireSize());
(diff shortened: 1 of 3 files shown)


pepone (approved):
Looks good, I will simplify the comments or drop them.

bernardnormier (approved):

PR #5839, opened 2026-07-07
Add Android IceBT-over-Bluetooth CI testing

externl:
Adds CI coverage for the IceBT (Bluetooth) transport by running real Ice test suites between two Android emulators over the emulator's userspace virtual Bluetooth controller (Netsim/Rootcanal) — no hardware. GitHub-hosted Linux runners can't host a Bluetooth stack, but two emulators do RFCOMM between themselves.

- **Hardware-free endpoint tests** (`cpp/test/IceBT/endpoints`) — parse `bt`/`bts` endpoint strings with the plug-in registered but not initialized (`Ice.InitPlugins=0`); run in the normal Linux suite under tcp.
- **Two-emulator harness support** (`scripts/Util.py`) — device-scoped adb + a per-emulator forward port, four two-device reachability fixes, and adb-driven device-setup helpers (boot-wait, priv-app install, BT enable, bond).
- **Bluetooth modes in the harness** (`scripts/Controller.py`) — `--bt-emulators`, `--bt-prepare`, `--bt-setup`, `--bt-bond` and `--bt-diagnostics` keep the emulator lifecycle and all the adb work in Python, so CI and a local run take the same path.
- **A new `android-bt` CI configuration** (`.github/actions/setup-bt` + `.github/workflows/ci.yml`) — the composite action builds the pre-bond helper, boots and bonds the emulator pair, and starts the server-side process controller; the matrix entry then runs the client through the shared test step, reusing the existing setup, build and ccache steps.
- **`java/test/android/btbond`** — IceBT needs secure RFCOMM, which only works between BR/EDR bonded devices, and bonding is normally an interactive pairing dialog with no adb equivalent. `createBond()` and `setPairingConfirmation()` are app-level APIs, and the latter requires `BLUETOOTH_PRIVILEGED`, which an `adb install`-ed app cannot hold — hence a small helper installed into `/system/priv-app` that opens a secure RFCOMM socket to trigger pairing and auto-confirms it. It runs once during setup and exits before the tests start.

The `test` action gains optional `workers` and `all` inputs, defaulting to today's values, because the Bluetooth job drives a single device and must not fan each suite out over configurations that don't apply to Bluetooth.

Exercises `Ice/operations` plus the marshalling core (`ami`, `enums`, `exceptions`, `facets`, `inheritance`, `objects`, `optional`, `proxy`) — the suites verified to pass over emulated RFCOMM. The job takes ~23 min alongside the rest of the matrix, comfortably inside the longest job, and is dominated by the shared C++ build rather than the tests themselves (~2 min of actual Bluetooth testing). C++↔Java cross-tests stay out of scope, since the C++ side needs a host Bluetooth stack. `BLUETOOTH.md` documents running the same thing locally.


java/test/android/btbond/src/main/java/com/zeroc/btbond/MainActivity.java
```diff
@@ -0,0 +1,191 @@
...
+    // (BLUETOOTH_PRIVILEGED), so setPairingConfirmation should succeed; setPin is a fallback that logs
+    // any SecurityException.
+    private final BroadcastReceiver pairingReceiver = new BroadcastReceiver() {
+        @Override
+        public void onReceive(Context ctx, Intent intent) {
+            BluetoothDevice dev = intent.getParcelableExtra(BluetoothDevice.EXTRA_DEVICE);
+            int variant = intent.getIntExtra(BluetoothDevice.EXTRA_PAIRING_VARIANT, -1);
+            Log.i(TAG, "PAIRING_REQUEST variant=" + variant + " from " + (dev == null ? "?" : dev.getAddress()));
+            try {
+                dev.setPairingConfirmation(true);
```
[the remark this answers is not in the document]
externl: Fixed in c854273. `setPairingConfirmation` is now checked for a `false` return as well as for a throw, and both fall through to the `setPin` fallback — previously a refused confirmation was logged as accepted and skipped the fallback entirely, which is exactly the case it exists for.

java/test/android/btbond/src/main/java/com/zeroc/btbond/MainActivity.java
```diff
@@ -0,0 +1,191 @@
...
+                    : adapter.listenUsingInsecureRfcommWithServiceRecord("btbond", uuid);
+                try {
+                    Log.i(TAG, "listening secure=" + secure + ", waiting up to 120s for client...");
+                    BluetoothSocket s = ss.accept(120000);
+                    try {
+                        Log.i(TAG, "accepted a connection");
+                        InputStream in = s.getInputStream();
+                        OutputStream out = s.getOutputStream();
+                        byte[] buf = new byte[64];
+                        int n = in.read(buf);
```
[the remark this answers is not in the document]
externl: Fixed in c854273. The server now reads the whole probe through a `readFully` helper before echoing, and reports OK only on a complete read. The buffer is also sized to the probe rather than 64 bytes.

java/test/android/btbond/src/main/java/com/zeroc/btbond/MainActivity.java
```diff
@@ -0,0 +1,191 @@
...
+                    : dev.createInsecureRfcommSocketToServiceRecord(uuid);
+                try {
+                    Log.i(TAG, "connecting secure=" + secure + "...");
+                    s.connect();
+                    Log.i(TAG, "connected");
+                    OutputStream out = s.getOutputStream();
+                    InputStream in = s.getInputStream();
+                    out.write("PING".getBytes()); out.flush();
+                    byte[] buf = new byte[64];
+                    int n = in.read(buf);
```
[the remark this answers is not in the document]
externl: Fixed in c854273 — the client reads the full probe through the same `readFully` helper before comparing, so a split write no longer reads as a bonding failure.

scripts/Util.py
```diff
@@ -2561,10 +2593,238 @@ def getControllerIdentity(self, current):
...
+
+    def reboot(self):
+        self._adbTolerant("reboot")
+        # `adb reboot` returns while the device is still up, so waitForBoot's wait-for-device would
+        # return at once and read sys.boot_completed from the system we just asked to go down. Wait
+        # for the device to actually drop off adb first.
+        try:
+            subprocess.run([*self.adb().split(), "wait-for-disconnect"], timeout=60, check=False)
+        except subprocess.TimeoutExpired:
+            pass
```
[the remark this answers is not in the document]
externl: Fixed in c854273. `adb reboot` now goes through `run()` so a rejected reboot raises, and a device that never drops off adb raises rather than falling through. Either case previously left `waitForBoot` reading the old boot's `sys.boot_completed`, so a `/system` app install that never took effect looked like a success.

scripts/Util.py
```diff
@@ -2561,10 +2593,239 @@ def getControllerIdentity(self, current):
...
+        raise RuntimeError(f"'{self.device}' did not become root within {timeout}s")
+
+    def bluetoothAddress(self):
+        return run(f"{self.adb()} shell settings get secure bluetooth_address").strip()
+
+    def enableBluetooth(self):
+        self._adbTolerant("root")
+        time.sleep(2)
+        self._adbTolerant("shell cmd bluetooth_manager enable")
+        self._adbTolerant("shell cmd bluetooth_manager wait-for-state:STATE_ON")
```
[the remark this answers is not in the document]
externl: Good catch — and the sequencing makes it worse than described: `enableBluetooth` runs immediately after `installSystemApp`'s reboot, so `bluetooth_address` would still hold the previous boot's value. The check was blind in the only situation it ever ran in.

Fixed in 317df02: `wait-for-state:STATE_ON` now goes through `run()` so it fails setup, `enable` stays tolerant for the already-enabled case, and the address check is gone rather than left as false reassurance.

Worth noting this command has been tolerant since it was written, so CI has never actually proven it succeeds — the next run is the first real evidence either way.

scripts/Util.py
```diff
@@ -2550,9 +2564,27 @@ def __init__(self, current):
...
-        run(f"{self.adb()} forward tcp:15001 tcp:15001")
+        port = self.hostPort()
+        print(f"forwarding host port {port} to the controller app (device={self.device})")
+        run(f"{self.adb()} forward tcp:{port} tcp:15001")
+
+    def bindProcessesToController(self):
+        # The controller app advertises its process proxies on the device port (15001). The host only
+        # reaches the device through a per-emulator adb forward, so rebind process proxies onto the
+        # controller connection instead of connecting to the advertised (unreachable) endpoint.
+        return True
```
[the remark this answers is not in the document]
externl: The observation is right but the consequence doesn't follow, so I've documented this rather than changed the routing.

`RemoteProcessController.getHost` does ask the app, and the app does return `127.0.0.1` on an emulator — so `--host-bt` genuinely doesn't reach the endpoints here. But that doesn't put the tests on loopback, because the Java IceBT acceptor never uses that value: `AcceptorI.java:64` calls `listenUsingRfcommWithServiceRecord()` on the local adapter, and line 106 publishes `bluetoothAdapter().getAddress()` — the real adapter address. The client reaches the server through that published proxy, not through `Ice.Default.Host`.

Two independent checks agree. The client runs on emulator-5554 and the server on emulator-5556, which have separate loopbacks and no TCP route between them — if the client were connecting to `127.0.0.1` it would find nothing, yet 9 suites pass. And a deliberately bogus `--host-bt` earlier in this branch's history also passed, for exactly this reason.

Added a comment at the `AndroidProcessController` explaining why `getHost` is deliberately not overridden, since the code as written invites precisely this reading.

scripts/Util.py
```diff
@@ -2638,7 +2898,8 @@ def startControllerApp(self, current, ident):
         except Exception:
             pass
         print("installing the controller application")
-        run("{} install -t -r {}".format(self.adb(), current.testcase.getMapping().getApk(current)))
+        # -g grants runtime permissions (e.g. BLUETOOTH_CONNECT/SCAN) the controller needs for bt tests.
+        run("{} install -t -r -g {}".format(self.adb(), current.testcase.getMapping().getApk(current)))
```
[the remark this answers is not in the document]
externl: `adb connect` does run for emulator serials, but it doesn't raise — I checked directly:

```
$ adb connect emulator-5554
failed to resolve host: 'emulator-5554': nodename nor servname provided, or not known
$ echo $?
0
```

Exit status 0, so `run()` returns normally and the install proceeds — which is why the controller app has been installing correctly in CI all along.

Still worth skipping, though: it's a pointless network call that prints an alarming line into the logs of a job that is working fine. Done in 9f2bee0, guarded on the `emulator-` prefix.

scripts/Util.py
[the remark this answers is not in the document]
externl: Correcting myself: the explanation I gave above was wrong, and I've replaced the comment it produced (19b075a3).

I claimed the IceBT acceptor publishes the local adapter's address. It does not — `AcceptorI.listen()` returns the endpoint unchanged and `EndpointI.toPublishedEndpoint()` returns `this`, so the published `-a` is whatever `_addr` already was; `bluetoothAdapter().getAddress()` only ever appears in `toString()`/`getInfo`, never in an endpoint.

Your premise was also wrong, though, so the conclusion still doesn't follow: `_startServerSide` calls `getProcessController(current)` **without a process argument**, and that dispatch's Android branch is `elif process and current.config.android`. With `process=None` it falls through to `LocalProcessController`, so `RemoteProcessController.getHost` — and the app's `127.0.0.1` — is never reached on this path at all.

What actually happens: `LocalProcessController.getHost` -> `Driver.getHost` -> `--host-bt`. So `--host-bt` on the server-side `Controller.py` is load-bearing (it becomes the server's `Ice.Default.Host`, which IceBT uses as the endpoint address when no `-a` is given), and `startServerSide` hands that value back for the client. The `--host-bt` on the `allTests.py` line is the inert one, which is why a bogus value there doesn't fail the run.

Net effect on the code is unchanged — `getHost` still shouldn't be overridden — but the comment now describes the real mechanism instead of a made-up one.

scripts/Util.py
```diff
@@ -2550,21 +2564,274 @@ def __init__(self, current):
...
+        peerAdb = AndroidProcessController.forDevice(peerDevice).adb()
+        if not re.fullmatch(r"[A-Fa-f0-9-]+", uuid):
+            raise RuntimeError(f"invalid service UUID: {uuid!r}")
+        peerAddress = run(f"{peerAdb} shell settings get secure bluetooth_address").strip()
+        if not peerAddress or peerAddress == "null":
+            raise RuntimeError(f"could not read the Bluetooth address of peer '{peerDevice}'")
+        activity = f"{package}/.MainActivity"
+        # Clear both logs first: the result line is matched out of logcat below, and one left over
+        # from an earlier attempt would otherwise be read as this attempt's result.
+        self._adbTolerant("logcat -c")
```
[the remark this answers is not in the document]
externl: Agreed, and already fixed in 19b075a3 — this review is anchored to 9f2bee0, which was the head when it started.

The two clears had their tolerance exactly inverted: ours (the buffer the verdict is read from) was tolerant, the peer's (which nothing reads) was fatal. That's now swapped, with a comment on each explaining why.

The result check was tightened too, since making the clear mandatory alone doesn't close the hole you describe — a substring search still lets an older line decide the outcome if btbond ever runs twice within a boot. It now pulls every `RESULT client <verdict>` with a regex and requires the *last* one to be OK.

scripts/Util.py
```diff
@@ -2550,21 +2564,308 @@ def __init__(self, current):
...
+        self._adbTolerant("shell cmd bluetooth_manager enable")
+        # Not tolerant: this is the only command that establishes the adapter actually reached
+        # STATE_ON. bluetooth_address can't stand in for it -- it is a persisted secure setting that
+        # stays populated while the adapter is off, and enableBluetooth runs just after
+        # installSystemApp's reboot, so it would still hold the value from the previous boot.
+        run(f"{self.adb()} shell cmd bluetooth_manager wait-for-state:STATE_ON")
+
+    def grantRuntimePermissions(self, package, permissions):
+        for permission in permissions:
+            self._adbTolerant(f"shell pm grant {package} {permission}")
```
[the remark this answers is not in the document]
externl: Fixed in 6e9bf42 — the grant now goes through `run()` and fails setup. Agreed on the reasoning: it runs right after the fatal `wait-for-state`, so adb is known-good there, and idempotency means the only failures left are real ones.

scripts/Util.py
```diff
@@ -2550,21 +2564,308 @@ def __init__(self, current):
...
+        # "RESULT client OK" left in it would be read as this attempt's result. The peer's is only
+        # tidiness -- nothing reads it -- so a transient failure there must not kill the bond.
+        run(f"{self.adb()} logcat -c")
+        self._adbTolerantFor(peerAdb, "logcat -c")
+        run(f"{peerAdb} shell am start -n {activity} --es mode server --es uuid {uuid}")
+        time.sleep(6)
+        run(f"{self.adb()} shell am start -n {activity} --es mode client --es peer {peerAddress} --es uuid {uuid}")
+        result = ""
+        verdict = ""
+        for _ in range(40):
```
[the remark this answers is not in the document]
externl: Fixed in 6e9bf42 — the poll is now 60×3s ≈ 180s, past the 150s watchdog with margin. Fair catch on a gap I created: the watchdog was added without revisiting the host-side budget, so a slow-but-in-budget bond could be rejected and the watchdog's terminal verdict was never collected.

scripts/Util.py
```diff
@@ -2550,21 +2564,308 @@ def __init__(self, current):
...
+            result = self._adbTolerant("logcat -d -s BTBOND")
+            # Take the LAST verdict, not any match: if the activity ran more than once, an earlier
+            # line must not decide the outcome -- an old FAIL followed by an OK (or the reverse)
+            # would otherwise be read by substring search as whichever we happened to look for.
+            verdicts = re.findall(r"RESULT client (\w+)", result)
+            if verdicts:
+                verdict = verdicts[-1]
+                break
```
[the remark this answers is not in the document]
externl: Partially addressed in 6e9bf42, deliberately short of a per-attempt correlation protocol.

What changed: after the first verdict sighting the host settles ~6s and re-reads, taking the last verdict of the final read — so a trailing verdict from a racing recreated worker is picked up in the common case.

Why not the full fix: correlating attempts means changing the helper again, and the last robustness change to its lifecycle (closing sockets in onDestroy) regressed CI in exactly this recreate scenario. The invariant the harness actually needs is already gated independently: `dumpsys bluetooth_manager` must show the peer bonded, and that reads device state, not logcat. A stale OK can't fake that, and the residual race (a verdict landing after the settle window) fails conservatively — loud, not silently green.

java/test/android/btbond/src/main/java/com/zeroc/btbond/MainActivity.java
```diff
@@ -0,0 +1,251 @@
+// Forces a BR/EDR bond between the two Android emulators (over Netsim/Rootcanal) before the IceBT
```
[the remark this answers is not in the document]
externl: Fixed in 3c5e648. Only MainActivity.java needed it — `settings.gradle` and `AndroidManifest.xml` in btbond also lack the header, but so do the controller app's, so those follow the existing convention.

scripts/Controller.py
```diff
@@ -189,6 +251,105 @@ def getHost(self, name, ipv6, c):
         adapter.activate()
         self.communicator.waitForShutdown()
 
+    @staticmethod
+    def emulatorPort(serial):
+        # "emulator-5554" -> 5554
+        try:
+            return int(serial.rsplit("-", 1)[1])
+        except (IndexError, ValueError):
+            raise RuntimeError(f"expected an emulator serial like 'emulator-5554', got '{serial}'")
```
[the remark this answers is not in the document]
externl: Fixed in 3c5e648 — the prefix is now checked alongside the port.

Worth noting the failure mode is nastier than "targets a nonexistent serial": adb names an emulator after the port it listens on, so `phone-5554` boots an emulator that adb knows as `emulator-5554`. The emulator is genuinely up and healthy while every later command addresses a serial that will never exist, so it surfaces as a 300s boot timeout rather than as bad input.

Checked `emulator-5554`, `emulator-5556`, `phone-5554`, `emulator-abc`, `emulator-`, `5554` and `""`.

java/test/android/btbond/gradle/wrapper/gradle-wrapper.jar
pepone, line 1: Can we avoid adding a new Gradle wrapper, and use the existing one?

cpp/test/IceBT/endpoints/Client.cpp
```diff
@@ -0,0 +1,123 @@
+// Copyright (c) ZeroC, Inc.
```
pepone, line 1: Linux can use a virtual BT device, did we explore that alternative?

java/test/android/btbond/src/main/java/com/zeroc/btbond/MainActivity.java
```diff
@@ -0,0 +1,253 @@
+// Copyright (c) ZeroC, Inc.
+
+// Forces a BR/EDR bond between the two Android emulators (over Netsim/Rootcanal) before the IceBT
```
pepone, line 3: this is just for `bts` or for `bt` too?

pepone: Answered my own question — it's both. IceBT always uses secure RFCOMM (`AcceptorI.java:62` even comments "We always listen using the 'secure' method", and the connector uses `createRfcommSocketToServiceRecord`, not the insecure variant), and secure RFCOMM only works between bonded devices. `bts` just layers Ice SSL on top of that same bonded transport. So the helper is needed for `bt` too, not only `bts`. No change needed.

pepone (approved):
Really solid work — this is a lot of careful adb/emulator plumbing and it reads cleanly. The endpoint-parsing test, the two-emulator harness, and the btbond helper all hang together well, and the review history shows the tricky bits (reboot detection, verdict matching, secure-RFCOMM bonding) were chased down properly. Approving.

Two non-blocking follow-ups, neither needs to hold this:

- `bond()` interpolates `peerAddress` into the `am start` command unquoted, while the serial and UUID right next to it are regex-validated before hitting `shell=True`. Worth the same guard for consistency:
  ```python
  if not re.fullmatch(r"([0-9A-Fa-f]{2}:){5}[0-9A-Fa-f]{2}", peerAddress):
      raise RuntimeError(f"invalid peer Bluetooth address: {peerAddress!r}")
  ```
- The btbond verdict race is already documented and backstopped by the `dumpsys` bond-state check, so fine to defer — but a nonce-scoped "server ready" signal plus a single per-attempt verdict would remove the remaining flakiness whenever you next touch the helper.

On my inline questions: bt vs bts — confirmed it's both (IceBT always uses secure RFCOMM, `bts` just layers SSL on top), so no change needed. The gradle-wrapper and Linux-vhci threads are the only ones I'd still like a line on.

java/test/android/btbond/gradle/wrapper/gradle-wrapper.jar
externl: Done in 08cf26c — btbond now uses `java/gradlew`, which is already on the same Gradle version (9.2.1). That drops 5 files, including the checked-in wrapper jar (-770 lines).

Verified with a clean build from an empty `build/` using the exact command CI runs.

Note I left the controller app's own wrapper alone — it predates this PR, so consolidating it felt out of scope here, but happy to do that too if you'd like them unified.

cpp/test/IceBT/endpoints/Client.cpp
externl: Yes — that was the first thing tried, and it's why this ended up on Android emulators instead.

PR #5803 added a "Probe Bluetooth runner capabilities" step that ran on the Linux legs (amd64 and arm64) and checked exactly that path:

```
for m in bluetooth l2cap rfcomm bnep hci_vhci; do sudo modprobe "$m"; done
ls -l /dev/vhci
grep -E 'CONFIG_BT(_RFCOMM|_HCIVHCI|_BNEP)?=' /boot/config-$(uname -r)
command -v btvirt   # from bluez-tests
```

Result on the GitHub-hosted runners (kernel 6.17.0-azure): every `modprobe` failed, no `/dev/vhci`, and `btvirt` isn't packaged in Ubuntu's bluez — it would need building from BlueZ source. The kernel simply doesn't ship the Bluetooth modules, so there's nothing for `hci_vhci`/`btvirt` to attach to and no RFCOMM stack for IceBT to use.

One caveat on that evidence: I can quote the probe itself from git history, but the original run's logs are gone with the closed PR's branch, so the result above is the conclusion recorded at the time rather than a log I can re-read today. Runner images do change — if you want, I can re-add the probe step for one run to re-confirm on the current image.

Worth noting this only blocks the **C++** side, which needs a host kernel BT stack. That's also why C++↔Java cross-tests stay out of scope here. A self-hosted runner with a real adapter (or a real Android device, per BLUETOOTH.md) would unblock it.

externl: Correcting myself on one point above: I said "the kernel simply doesn't ship the Bluetooth modules". That's not right, and it makes the path sound more dead than it is.

The modules aren't *absent* — they're just not *installed*. Ubuntu splits them out: `hci_vhci.ko` lives in `linux-modules-extra-<version>`, not the base `linux-modules`. And `linux-modules-extra-azure-6.17` exists for noble, matching the runners' 6.17.0-azure kernel; I also confirmed a `hci_vhci.ko.zst` shipping inside an azure kernel modules package, so these are built for azure, not stripped from it.

The probe only ran `modprobe`, which fails when the .ko isn't on disk. It never tried `apt-get install linux-modules-extra-$(uname -r)` first. So the conclusion I drew from it was too strong.

What would still need solving, and is the harder half:

- **`btvirt`** — still genuinely unpackaged (BlueZ ships it under `tools/`, built only with `--enable-testing`), so creating the virtual controllers means building it, or writing the small bit of code that opens `/dev/vhci` directly.
- **`bluetoothd`** — needed, since the C++ IceBT plug-in talks to BlueZ over D-Bus. That part *is* packaged.
- **Unknowns** — whether a hosted runner permits `modprobe` of these at all, and whether two vhci controllers will actually pair and carry RFCOMM. BlueZ's own test suite does roughly this, so it's plausible, but unproven here.

So: "install a package and modprobe" rather than "build a kernel", which is a much lower bar than I implied. Worth a cheap probe before anyone commits to it — one CI step doing the apt-get + modprobe + `ls /dev/vhci` would settle the kernel half in a single run. Happy to put that up separately if it's of interest, since it's the C++ side this PR can't reach.

externl: Tracking this as a follow-up PR rather than growing this one.

Plan: start with a throwaway probe step — `apt-get install linux-modules-extra-$(uname -r)`, `modprobe hci_vhci bluetooth rfcomm l2cap bnep`, `ls /dev/vhci` — which settles the kernel half in one run for almost nothing. If that comes back clean, the second step is `btvirt` (build from BlueZ source, or open `/dev/vhci` directly) plus `bluetoothd`, and then C++ IceBT tests between two virtual controllers.

That's the piece this PR can't reach: everything here exercises the Java implementation, so the C++ side still only has endpoint-parsing coverage.

.github/workflows/ci.yml
```diff
@@ -140,6 +167,9 @@ jobs:
             build_cpp_and_python: true
 
     runs-on: ${{ matrix.os }}
+    # Individual steps carry their own tighter limits; this is a backstop so a stall in a step that
+    # has none (SDK download, cache, artifact upload) can't sit on a runner for the 6h default.
+    timeout-minutes: 150
```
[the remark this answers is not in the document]
externl: Good catch — fixed in 0fb39c9 by scoping it, which is the option that matches the intent.

The job-level limit was only ever meant to restore the bound the standalone Bluetooth workflow had. Applying it matrix-wide imposed a new constraint on 17 jobs that previously had none, which wasn't intended. It's now `${{ matrix.job_timeout || 360 }}` with only android-bt setting 90 — the value its own workflow used — so everything else keeps GitHub's default.

Two things worth recording from checking this: the longest job actually observed on main is ~54 min, so the risk wasn't imminent, but you're right that the bound has to clear the sum of step limits on the longest path (release-on-macOS allows 90+45+30+45) rather than observed runtimes. And the consequence you flagged is the worse half — a job killed at the job level skips the `always()` diagnostics and artifact upload, so it fails in the least debuggable way possible.

cpp/test/IceBT/endpoints/Client.cpp
externl: One more correction to what I wrote above: I said `btvirt` isn't packaged and would need building from the BlueZ source tree. That's wrong — it ships in **`bluez-test-tools`** (`/usr/bin/btvirt`), which is in the archive for both architectures. The earlier probe only installed `bluez` and then checked `command -v btvirt`, so it looked absent.

That removes the build-from-source step entirely. #6089 now installs `bluez-test-tools`, runs `btvirt -l2` to create two virtual controllers, and adds a `cpp-bt` matrix entry that runs the C++ tests over `--protocol=bt` against them — so `setup-bluetooth` is only invoked by a job that actually tests Bluetooth.

.github/workflows/ci.yml
```diff
@@ -114,6 +114,36 @@ jobs:
...
+            build_cpp_and_python: true
+            build_android_controller: true
+            # A single device, and only the configuration under test -- --all would fan each suite
+            # out over configurations that don't apply to Bluetooth. Keep test_all a quoted string:
+            # a YAML boolean false is falsy in the `matrix.test_all || 'true'` default below.
+            test_workers: 1
+            test_all: "false"
+            # The job normally takes ~25 min; this is the bound the standalone workflow had, so a
+            # wedged emulator costs that rather than GitHub's 6h default.
+            job_timeout: 90
```
bernardnormier: The step limits on this job's longest path already sum to 100 minutes (setup-bt 30 + test 45 + bts 20 + diagnostics 5) before the build, so 90 doesn't clear it — the rule the comment on `timeout-minutes` below states. Concretely: a slow setup followed by a test step that runs into its own 45-minute timeout gets the job killed at 90 mid-step, skipping the `always()` diagnostics this PR goes to some trouble to guarantee.

```suggestion
            # The job normally takes ~25 min; bound it so a wedged emulator costs that rather than
            # GitHub's 6h default. Keeps clear of the sum of the step limits on the longest path
            # (setup-bt 30 + test 45 + bts 20 + diagnostics 5, plus the build).
            job_timeout: 120
```

.github/actions/setup-bt/action.yml
```diff
@@ -0,0 +1,88 @@
...
+        # devices to query when one of the steps below is what failed.
+        echo "BT_CLIENT=${{ inputs.client }}" >> "$GITHUB_ENV"
+        echo "BT_SERVER=${{ inputs.server }}" >> "$GITHUB_ENV"
+
+    - name: Build the btbond APK
+      shell: bash
+      run: |
+        # btbond needs an explicit keystore next to its build file -- unlike the controller app it
+        # can't fall back to AGP's default debug signing, which fails validateSigningDebug here.
+        keytool -genkeypair -v -storepass android -keypass android -alias androiddebugkey \
```
bernardnormier, line 41: `keytool -genkeypair` fails with "alias <androiddebugkey> already exists" when the keystore is already present. Fresh CI runners never hit this, but a re-run of the action (or a local run that already followed BLUETOOTH.md) does.

```suggestion
        rm -f java/test/android/btbond/debug.keystore
        keytool -genkeypair -v -storepass android -keypass android -alias androiddebugkey \
```

BLUETOOTH.md
```diff
@@ -1,87 +1,130 @@
...
-Where:
-- `--host-bt` is the Bluetooth address of the machine running the C++ server
-- `--host` is the IP address of the machine running the C++ server
-- `--id=server` identifies this controller as managing servers
+**1. Build `btbond`.** IceBT needs bonded devices. `btbond` is a privileged helper that
+auto-confirms pairing.
 
-Then from the `java` directory, start the Java/Android test suite to run against the C++ servers:
+```bash
+keytool -genkeypair -v -keystore java/test/android/btbond/debug.keystore -storepass android \
```
bernardnormier, line 72: Same as in setup-bt: a second local run fails on the existing keystore; removing it first makes the documented flow rerunnable.

```suggestion
rm -f java/test/android/btbond/debug.keystore
keytool -genkeypair -v -keystore java/test/android/btbond/debug.keystore -storepass android \
```

bernardnormier (approved):
Nice work — this is a lot of finicky harness machinery, and the comments explaining the failure modes (the onDestroy socket tradeoff, the logcat verdict races, the load-bearing server-side `--host-bt`) make it reviewable.

A few things I verified rather than took on faith:

- The `test.py` guard (`pkg-config --exists bluez dbus-1`) matches the condition that sets `IceBT_system_libs` in `config/Make.rules.Linux` exactly, so built and run stay in sync.
- All three `matrix.config == 'android'` conditions in ci.yml were updated to `startsWith`; no stragglers.
- The base `RemoteProcessController` defines `setup()`, so the re-setup after `restartControllerApp` is safe for the non-Android controllers.
- Only the empty-`--device` path sets `self.emulator`, so the new `adb kill-server` gating does keep external-serial runs from dropping the peer's forwards.

Two small suggestions inline: the job timeout doesn't clear the sum of the step limits (the rule its own comment states), and the `keytool` invocations fail on a re-run when `debug.keystore` already exists.

Nits, take or leave: `MainActivity` uses the deprecated one-arg `getParcelableExtra`, and `hostPort()` would map `emulator-5001` onto the 15001 default (theoretical — adb assigns 5554+).

scripts/Util.py
```diff
@@ -2550,21 +2564,321 @@ def __init__(self, current):
...
 
     def adb(self):
-        return "adb -d" if self.device == "usb" else "adb"
+        if self.device == "usb":
+            return "adb -d"
+        elif self.device:
+            # The serial is interpolated into shell commands run with shell=True; only allow the
+            # characters that appear in adb serials (emulator-NNNN, USB serials, host:port) so a
+            # stray value can't inject shell syntax.
+            if not re.fullmatch(r"[A-Za-z0-9._:-]+", self.device):
```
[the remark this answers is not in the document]
externl: Fixed in bacf53f — a real regression on my part, since `adb()` is on the path for every `--device`.

Both halves of your point were needed. Widening the allowlist alone isn't enough: `[` and `]` are shell globs, so an unquoted bracketed serial could still misfire once accepted. So `adb()` is now split — `adbArgs()` returns argv for the seven `subprocess.run` callers, and `adb()` joins it through `shlex.quote` for the f-string/`shell=True` ones.

Checked: `emulator-5554`, `usb`, `192.168.1.5:5555`, `[2001:db8::1]:5555`, `[fe80::1%eth0]:5555` all pass and come out correctly in both forms (plain serials stay unquoted); `x; rm -rf /`, `a$(id)` and backtick forms are still rejected.

.github/workflows/ci.yml
externl: Taken as suggested — 120 in bacf53f.

You're right that it fails the rule stated in the comment two lines below it, which I wrote and then didn't apply here: 30 + 45 + 20 + 5 is already 100 before the build. And the consequence is the bad one — a job killed at the job level skips the `always()` diagnostics step, so it would fail in exactly the case where the logs matter most.

.github/actions/setup-bt/action.yml
externl: Taken as suggested — fixed in bacf53f.

Confirmed rather than assumed: running `keytool` twice against the same keystore fails on the second attempt, and prefixing `rm -f` makes it succeed.

BLUETOOTH.md
externl: Fixed in bacf53f, same as the action. Worth having in both: the documented local flow is the one most likely to be run twice.

java/test/android/btbond/build.gradle
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) ZeroC, Inc.
+
+// btbond is a tiny standalone Android app used only by the Bluetooth CI harness to force a BR/EDR
+// bond between the two emulators before the IceBT tests run (IceBT uses secure RFCOMM, which
+// requires bonding). It has no Ice dependencies -- it drives android.bluetooth directly. It is
+// installed as a privileged system app so it can auto-confirm pairing.
+plugins {
+    id 'com.android.application' version '8.13.2'
```
InsertCreativityHere, line 8: This adds non-trivial Java code, so it would be good (and also just consistent) to use the `checkstyle` and `rewrite` plugins (mirroring what we do in the other `controller` 'test'). Otherwise the code will have 0 linting or formatting checked on it.

Now that we have two projects, might be worth refactoring out some common stuff? But also maybe 2 is still small enough that copy/pasting is totally fine.

.github/actions/setup-bt/action.yml
```diff
@@ -0,0 +1,91 @@
...
+        # btbond needs an explicit keystore next to its build file -- unlike the controller app it
+        # can't fall back to AGP's default debug signing, which fails validateSigningDebug here.
+        # Remove any existing one first: keytool fails with "alias already exists" otherwise, which
+        # a fresh runner never hits but a re-run does.
+        rm -f java/test/android/btbond/debug.keystore
+        keytool -genkeypair -v -storepass android -keypass android -alias androiddebugkey \
+          -keyalg RSA -keysize 2048 -validity 10000 -dname "CN=Android Debug,O=Android,C=US" \
+          -keystore java/test/android/btbond/debug.keystore
+        # btbond is a standalone Gradle project but has no wrapper of its own; use java's.
+        (cd java/test/android/btbond && ../../../gradlew --no-daemon assembleDebug)
```
InsertCreativityHere: So, right now, all of our Gradle projects are standalone and self-contained because they have their own gradle wrappers right next to them. The added `btbond` is actually _not_ standalone, since it lacks this wrapper.

For consistency with our other projects, we should probably just add the wrappers for `btbond` like we do for all the others. It also turns this command into the more natural command:
```suggestion
        # btbond is a standalone Gradle project but has no wrapper of its own; use java's.
        (cd java/test/android/btbond && ./gradlew --no-daemon assembleDebug)
```
There's also some version coupling this avoids between `btbond` and the rest of our Java. But, this probably doesn't matter here tbh.

InsertCreativityHere: Here on the command line, we explicitly set `--no-daemon` but... (1/2)

java/test/android/btbond/gradle.properties
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) ZeroC, Inc.
+
+# btbond has no Ice or AndroidX dependency and no Slice to compile, so none of the controller app's
+# iceVersion / cppPlatform / cppConfiguration / useAndroidX properties apply here.
+
+# Gradle build properties
+org.gradle.daemon = true
```
InsertCreativityHere, line 8: (2/2) ... here we ask for `daemon` to be `true` for this project.

I'm guessing the command-line flag wins here? But this seems contradictory here.
Or is the logic that we want `deamon` outside of CI? (if I run this manually?).

java/test/android/btbond/gradle.properties
```diff
@@ -0,0 +1,8 @@
...
+org.gradle.configureondemand=false
```
InsertCreativityHere: To be fair, this bad whitespace was also present in `controller`. I'll cleanup `controller` in a separate PR.
```suggestion
org.gradle.configureondemand = false
```

java/test/android/btbond/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+}
+
+android {
+    namespace = 'com.zeroc.btbond'
+    compileSdk = 36
+
+    defaultConfig {
+        applicationId = 'com.zeroc.btbond'
+        minSdk = 34
+        targetSdk = 36
```
InsertCreativityHere, line 24: Just a note, here we say that we want to hit Sdks between 34 and 36 (inclusive).

java/test/android/btbond/src/main/java/com/zeroc/btbond/MainActivity.java
```diff
@@ -0,0 +1,253 @@
...
+    private volatile BluetoothServerSocket serverSocket;
+    private volatile BluetoothSocket socket;
+
+    // Best-effort auto-accept of incoming pairing requests. The app runs as a privileged system app
+    // (BLUETOOTH_PRIVILEGED), so setPairingConfirmation should succeed; setPin is a fallback that logs
+    // any SecurityException.
+    private final BroadcastReceiver pairingReceiver = new BroadcastReceiver() {
+        @Override
+        public void onReceive(Context ctx, Intent intent) {
+            BluetoothDevice dev = intent.getParcelableExtra(BluetoothDevice.EXTRA_DEVICE);
```
InsertCreativityHere: Found with Copilot review: `getParceableExtra` is deprecated as of Sdk 33.
See: https://developer.android.com/reference/android/content/Intent#getParcelableExtra(java.lang.String)

This method should be a drop-in replacement:
https://developer.android.com/reference/android/content/Intent#getParcelableExtra(java.lang.String,%20java.lang.Class%3CT%3E)

java/test/android/btbond/src/main/java/com/zeroc/btbond/MainActivity.java
```diff
@@ -0,0 +1,253 @@
...
+        // regressed CI once; the leak is transient and self-heals, so leave it alone. The watchdog
+        // bounds a genuinely stuck run instead.
+        super.onDestroy();
+    }
+
+    // Everything here is deliberately fixed rather than configurable: only secure RFCOMM triggers
+    // pairing, which is the whole point of this app, so there is no insecure mode to select.
+    void run(String mode, String peer, String uuidStr) {
+        try {
+            BluetoothAdapter adapter = BluetoothAdapter.getDefaultAdapter();
```
InsertCreativityHere: Likewise, `getDefaultAdapter` was deprecated in Sdk 31, see:
https://developer.android.com/reference/android/bluetooth/BluetoothAdapter#getDefaultAdapter()

That page also lists the drop-in replacement to use here instead.

InsertCreativityHere (commented):
Adding some comments as a checkpoint while I finish reviewing.

java/test/android/btbond/build.gradle
externl: Added both in abe49cc, mirroring the controller app's setup.

Worth reporting what the first checkstyle run turned up: 9 violations (line length, `{` placement, multiple statements per line, missing blank line before `package`). All fixed. That's a good argument for your point — the file genuinely had no formatting checked on it.

On refactoring the common config: I've left it copy/pasted for now. The two projects share the checkstyle/rewrite blocks but differ in the parts that matter (btbond has no Slice, no Ice dependency, no certificate copying, and needs its own `lint` disables for the privileged-system-app permissions), so extracting a common file would carry about as much conditional logic as it removes at n=2. Happy to do it if you'd rather have it now.

.github/actions/setup-bt/action.yml
externl: This one is a direct conflict with pepone's review, so flagging rather than just flipping it back.

The wrapper was there originally. pepone asked to remove it ("Can we avoid adding a new Gradle wrapper, and use the existing one?"), so 08cf26c deleted it and pointed the build at `java/gradlew` — which is already on the same Gradle version (9.2.1). That dropped 5 files including a checked-in jar, ~770 lines.

Your argument for putting it back is the one I originally used for keeping it, and I don't think either of you is wrong — it's a question of which consistency matters more: every Gradle project being self-contained, or not adding a third copy of the wrapper to the repo.

I'd rather you two settle it than have me flip the file based on whoever commented last. Happy to restore it the moment there's agreement — it's a one-command revert of 08cf26c.

externl: Your second guess is right, and I've made the file say so in abe49cc: the daemon is wanted for local builds, where it makes repeated runs much faster, and CI passes `--no-daemon` on the command line, which wins. Nothing is left running on the runner.

java/test/android/btbond/gradle.properties
externl: Taken as suggested in abe49cc.

java/test/android/btbond/src/main/java/com/zeroc/btbond/MainActivity.java
externl: Fixed in abe49cc — now uses the `Class`-typed overload. Safe here: it landed in SDK 33 and our minSdk is 34.

java/test/android/btbond/src/main/java/com/zeroc/btbond/MainActivity.java
externl: Fixed in abe49cc — the adapter now comes from `getSystemService(BluetoothManager.class).getAdapter()`. Also fine at minSdk 34, since that route predates it.

Both of these are worth having beyond the deprecation: the module now has checkstyle wired up, so this file is actually linted going forward.

java/test/android/btbond/build.gradle
InsertCreativityHere: Nah, I suspected it might not be worth it at n=2. Fine to have copy/paste for now.

.github/workflows/ci.yml
```diff
@@ -114,6 +114,37 @@ jobs:
...
+            build_android_controller: true
+            # A single device, and only the configuration under test -- --all would fan each suite
+            # out over configurations that don't apply to Bluetooth. Keep test_all a quoted string:
+            # a YAML boolean false is falsy in the `matrix.test_all || 'true'` default below.
+            test_workers: 1
+            test_all: "false"
+            # The job normally takes ~25 min; bound it so a wedged emulator costs that rather than
+            # GitHub's 6h default. Keeps clear of the sum of the step limits on the longest path
+            # (setup-bt 30 + test 45 + bts 20 + diagnostics 5, plus the build).
+            job_timeout: 120
```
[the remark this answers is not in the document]
externl: Raised to 210 in b49c1626. You're right that 120 didn't clear it — and so was the previous round's 120 suggestion, which accounted for the four later steps but not the 90-minute build.

Flagging the tension rather than pretending it's resolved: a job cap that must exceed the sum of its own step caps is close to inert — 210 against GitHub's 360 default is not much protection. The step limits are what actually bound a wedged emulator, and they fail the individual step while leaving the `always()` diagnostics and artifact steps to run, which is the better failure mode. The job cap is now just a backstop for the steps that carry no limit of their own (SDK install, cache, artifact upload). I've said that in the comment so the next person doesn't re-tighten it.

.github/actions/setup-bt/action.yml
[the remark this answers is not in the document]
externl: Correct, and a fair catch — I added the plugins in the previous commit but left setup-bt running only `assembleDebug`, so neither gate ever ran. Fixed in b49c1626: it now runs `build checkstyle` and then `rewriteDryRun --no-parallel --no-configuration-cache`, matching the controller app.

Running those locally before pushing turned up two things worth recording:

- `rewriteDryRun` failed outright — every recipe in the shared `rewrite.yml` came back "does not exist", because the `rewrite(...)` artifacts the controller declares weren't in btbond's build file. Added them; it passes now.
- `build` also produces an unsigned *release* APK, and `find ... | head -1` picked that one ahead of the debug APK. So this change would have handed the installer an unsigned APK. The path is now named explicitly instead of globbed.

scripts/Util.py
```diff
@@ -2550,21 +2565,327 @@ def __init__(self, current):
...
+        name = self.device or self.avd or "device"
+        t = time.time()
+        while (time.time() - t) <= timeout:
+            try:
+                if run(f"{self.adb()} shell getprop sys.boot_completed").strip() == "1":
+                    return
+            except RuntimeError:
+                pass  # device offline mid-reboot
+            time.sleep(3)
+        raise RuntimeError(f"'{name}' not booted after {timeout}s")
```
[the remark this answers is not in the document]
externl: Fixed in e6382f1 — one deadline now covers both phases, so the advertised timeout is the actual bound.

Same commit also drops the job timeout from 210 to 60, on Joe's point that it should follow the data rather than the sum of step limits: successful runs of this job take 23–29 minutes (worst observed 29 across the last ten completed runs), so 60 gives 2× headroom while actually bounding a wedged runner. The comment records the trade explicitly: a run slow enough to hit 60 is pathological, and cutting it at the job level — losing the always() diagnostics — is accepted in exchange for not holding a runner for hours.

scripts/Util.py
```diff
@@ -2550,21 +2565,329 @@ def __init__(self, current):
...
+        while time.time() < deadline:
+            self._adbTolerant("root")
+            try:
+                subprocess.run([*self.adbArgs(), "wait-for-device"], timeout=60, check=False)
+            except subprocess.TimeoutExpired:
+                pass
+            try:
+                if run(f"{self.adb()} shell id -u").strip() == "0":
+                    self._adbTolerant("remount")
+                    return
```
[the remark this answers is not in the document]
externl: Fixed in 18156ed, though not quite as suggested — returning only once remount succeeds would break the first call, which exists to disable dm-verity and cannot make /system writable until the reboot that follows it.

So the check goes where it matters: the push loop now calls `rootRemount(requireWritable=True)`, which probes by touching a file under /system and keeps retrying until that works. remount stays tolerant on the dm-verity pass. Probing rather than parsing /proc/mounts because system-as-root images don't give /system its own entry.

Worth noting this is preventive rather than a live bug: in the last green run the only swallowed failure was `adb root` (benign — already root), with no remount failures and no push retries at all. The failure mode you describe would have been invisible until it consumed all three attempts, which is exactly why it's worth closing.

PR #5866, opened 2026-07-07
Assign DataStorm topic element ids from a single counter

### LINKED ISSUE #5820: [Audit-High] DataStorm: any-key and filtered reader element IDs collide in the negated wire namespace
> **AI-generated finding** — surfaced by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), in a multi-agent correctness audit of DataStorm: five subsystem reviewers followed by three independent adversarial "refute-it" verifiers (this one confirmed with full wire-path evidence). Code is pinned to `origin/main` at [`90dc6c7`](https://github.com/zeroc-ice/ice/commit/90dc6c7171184e513f40114b36d1aa2dbf130c49). **Not yet human-triaged, and not independently cross-checked by Codex** — please verify before acting.

## Summary

`TopicReaderI` assigns element ids from **two independent counters** but both kinds of element are presented to the peer as **negative** ids, so the first any-key reader and the first filtered reader on a topic collide.

- An any-key reader (`makeAnyKeyReader` → `MultiKeyReader(topic, {}, …)`) goes through `create()` with empty keys, drawing its id from `++_nextId`, yet is routed to `addFiltered(element, alwaysMatchFilter)`:
  https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/TopicI.cpp#L847
  https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/TopicI.cpp#L955
- A filtered reader (`createFiltered`) draws its id from the **separate** `++_nextFilteredId`:
  https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/TopicI.cpp#L934

Both are identified on the wire as **negative** element ids (`attachFilter` negates: `const int64_t filterId = -elementId;`):
https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/DataElementI.cpp#L343

So the first filtered reader (id 1 → wire `-1`) and the first any-key reader (id 1 → wire `-1`) are the **same** element id to the writer node.

## Trigger

On one topic, create the first `makeFilteredKeyReader(...)` and the first `makeAnyKeyReader(...)`, both connected to a matching writer.

## Impact

On the writer node the two readers collapse into a single subscriber entry — `Listener::addOrGet` and the writer session's `TopicSubscriber.add(-1, …)` merge them, and the surviving entry keeps whichever reader attached **first** (its filter, sample filter, name, priority). The any-key reader then silently misses every sample the other reader's filter rejects. Destroying **either** reader sends `detachElements({-1})` (`cpp/src/DataStorm/DataElementI.cpp:1138` for the empty-keys reader, and the `FilteredDataReaderI` destroy path), which removes the merged entry and **detaches the survivor** until it reconnects. The writer side is safe (writers use a single counter; keyed vs any-key writers occupy distinct signed keys). No test covers coexisting filtered + any-key readers.

## Fix direction

Use a **single per-topic element-id counter** for all reader elements — the sign convention already disambiguates keys (+id) from filters (−id) on the wire, so distinct raw values are all that is required. Equivalent minimal fix: draw empty-keys (any-key) readers from `_nextFilteredId` in `TopicReaderI::create`.

pepone:
Fixes #5820.

`TopicReaderI` drew any-key reader element ids from `_nextId` but filtered reader ids from a separate
`_nextFilteredId` counter. Both kinds of element are presented to the peer as **negated** ids (filter
subscriptions), so the first any-key reader and the first filtered reader on a topic collided on the wire: the
writer's `Listener::addOrGet` merged them into a single subscription keyed by `(topicId, -1)`, keeping only the
first-attached reader's filter/sample filter/name/priority — and `detachElements({-1})` from destroying either
reader removed the merged entry, silently detaching the survivor until its session reconnects.

## Fix

Topic elements now draw their ids from the single per-topic `_nextId` counter (`_nextFilteredId` removed), making
raw ids unique across keyed, any-key, and filtered elements. The wire sign convention — positive for key
subscriptions, negative for filter subscriptions — still distinguishes their kind, and nothing depends on filtered
ids being dense or drawn from a separate counter. Writers already used the single counter and are unaffected.

## Testing

New case in `test/DataStorm/events`: an any-key reader and a filtered reader coexisting on the same topic — both
receive a sample matching the filter, then the filtered reader is destroyed and the any-key reader must still
receive a subsequent sample. Without the fix the writer hangs at `waitForReaders(2)` because the two readers merge
into a single subscription. The full DataStorm suite passes.

### DIFF
--- a/changelog.d/datastorm/5820.md
+++ b/changelog.d/datastorm/5820.md
@@ -1,0 +1,4 @@
+- Fixed a bug in DataStorm where an any-key reader and a filtered reader created on the same topic could be assigned
+  the same element ID and merge into a single subscription on the writer: one of the readers silently missed the
+  samples rejected by the other reader's filter, and destroying either reader detached the other. Topic elements now
+  draw their IDs from a single counter.
--- a/cpp/src/DataStorm/TopicI.cpp
+++ b/cpp/src/DataStorm/TopicI.cpp
@@ -931,7 +931,7 @@ TopicReaderI::createFiltered(
     auto element = make_shared<FilteredDataReaderI>(
         this,
         std::move(name),
-        ++_nextFilteredId,
+        ++_nextId,
         filter,
         std::move(sampleFilterName),
         std::move(sampleFilterCriteria),
--- a/cpp/src/DataStorm/TopicI.h
+++ b/cpp/src/DataStorm/TopicI.h
@@ -179,8 +179,10 @@ namespace DataStormI
         // The number of connected listeners.
         size_t _listenerCount{0};
 
+        // Element ids are drawn from a single counter for all elements of a topic (keyed, any-key, and filtered).
+        // Any-key and filtered elements are both presented to the peer as negated ids (filter subscriptions), so
+        // distinct raw values are required across all of them to keep their subscriptions apart.
         std::int64_t _nextId{0};
-        std::int64_t _nextFilteredId{0};
         std::int64_t _nextSampleId{0};
     };
 
--- a/cpp/test/DataStorm/events/Reader.cpp
+++ b/cpp/test/DataStorm/events/Reader.cpp
@@ -404,6 +404,36 @@ void ::Reader::run(int argc, char* argv[])
             testSample(reader2, SampleEvent::Update, "elem3", "value");
         }
     }
+
+    // Coexisting any-key and filtered readers on the same topic: each keeps its own subscription. Both receive a
+    // sample matching the filter, and destroying the filtered reader leaves the any-key reader subscribed.
+    {
+        Topic<string, string> topic(node, "readerCoexistence");
+        Topic<string, int> barrier(node, "readerCoexistenceBarrier");
+
+        auto anyKeyReader = makeAnyKeyReader(topic, "", config);
+        {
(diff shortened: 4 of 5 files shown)


bernardnormier (approved):
LGTM. Verified that any-key and filtered elements share the negated-id wire namespace, so the separate counters did collide; nothing depends on filtered ids being dense. Nice deterministic test.

externl (approved):

externl (approved):
Traced the full element-id lifecycle: creation, advertisement, and every consumer (Listener subscribers map, TopicSubscriber elements/keys, subscribeTo/unsubscribeFrom, detachElements sign handling, getLastIds dict). All uses are opaque map keys plus the ±sign convention — nothing depended on separate counters or dense filtered ids, and there's no wire-format change since ids are locally assigned and echoed opaquely. The single counter also removes a latent lastIds-dict collision where +elementId and -elementId folded into one positive-keyed dict.

PR #5867, opened 2026-07-07
Fix incorrect and stale API doc comments in C++ Ice headers

pepone:
Fixes incorrect or stale API doc comments in the hand-written C++ Ice headers, found during a C++/C# API doc review. Comment-only changes (plus one declaration inside `#ifdef ICE_DOXYGEN`); no behavior change.

Incorrect behavior claims corrected:

- `Communicator::destroyAsync`: a null `completed` callback makes the call a no-op — the communicator is not destroyed.
- `Communicator::waitForShutdownAsync`: the callback is normally called by a background thread; "already shut down → called by the current thread" only holds after shutdown completion was observed.
- `getProcessLogger`/`setProcessLogger`: the process logger is not used by "all communicators without their own logger" — loggers set in `InitializationData` or configured through logger properties take precedence.
- `OutgoingResponse::exceptionId`: for non-Ice exceptions the ID is the demangled type name (or `"unknown"`), not `what()`.
- `ServantLocator::deactivate`: called when the object adapter is destroyed, not deactivated.
- `Connection::setAdapter`: documents the `std::logic_error` thrown on incoming connections.
- `Connection::throwException`: also throws while the connection is closing.
- `WSConnectionInfo::headers`: holds the upgrade *response* headers on outgoing connections.
- `OutputStream::startValue`: preserved slices are only re-marshaled with the sliced format.
- `OutputStream::startEncapsulation()`: a nested encapsulation uses the enclosing encapsulation's encoding/format, not the stream defaults.
- `InputStream::throwException`: a null factory falls back to the stream's Slice loader, not "generated helper functions".
- `Value::ice_getSlicedData`: removed the stale "preserved-slice base class" condition (slice preservation is unconditional with the sliced format since 3.8).
- `InitializationException`: thrown for initialization failures in general, not just communicator initialization (matches the C# doc).
- `AlreadyRegisteredException::kindOfObject`: removed "object" and "replica group", which no code produces.
- `Ice::Service::main`: documents the reserved `--pidfile FILE` option.
- `UTF8Buffer::getMoreBytes`: returns a pointer to the newly reserved space, not the beginning of the buffer; dropped the stale `MarshalException` tag (no implementation throws it).
- `ToStringMode::Compat`: only *non-printable* ASCII characters are escaped, matching the other two enumerators.
- `InputStream::readAndCheckSeqSize`: `minSize` is per element.

Markup/typo fixes: broken `@see` targets (`InitializeData::pluginFactories`, removed stale `initialize(int&, const char*[])` overload reference), garbled Doxygen-only `writeAll` signature (`>>`, missing `void` — inside `#ifdef ICE_DOXYGEN`, never compiled), stray `@return encoding`, Schannel prose referencing the pre-rename `*CertificateSelectionCallback` names, the SecureTransport client `trustedRootCertificates` example pulling the server snippet, "ClientAuthenticationOptions" copy-paste in the Server options aliases, and typos (`SetConsoleCtrlCHandler`, "colleted", "shared point", "type type", "a AlreadyRegisteredException", "message return by", `CompositeSliceLoaderPtr`, FSize "prior to unmarshaling").


cpp/include/Ice/SSL/ClientAuthenticationOptions.h
```diff
@@ -43,7 +43,7 @@ namespace Ice::SSL
         /// @param host The target host name.
         /// @return The client SSL credentials.
         ///
-        /// Example of setting `clientCertificateSelectionCallback`:
+        /// Example of setting `clientCredentialsSelectionCallback`:
         /// @snippet Ice/SSL/SchannelClientAuthenticationOptions.cpp clientCertificateSelectionCallback
```
pepone: We should rename the snippet `clientCertificateSelectionCallback` too.

cpp/include/Ice/Communicator.h
```diff
@@ -44,7 +43,7 @@ namespace Ice
 
         /// Destroys this communicator asynchronously.
         /// @param completed The callback to call when the destruction is complete. This function must not throw any
-        /// exception.
+        /// exception. When @p completed is null, this function does nothing: the communicator is not destroyed.
```
pepone: What is the convention in C++ comments null or nullptr?

```suggestion
        /// exception. When @p completed is nullptr, this function does nothing: the communicator is not destroyed.
```

cpp/include/Ice/Connection.h
```diff
@@ -153,7 +155,7 @@ namespace Ice
         /// Throws an exception that provides the reason for the closure of this connection. For example, this function
         /// throws CloseConnectionException when the connection was closed gracefully by the peer; it throws
         /// ConnectionAbortedException when the connection is aborted with #abort. This function does nothing if the
-        /// connection is not yet closed.
+        /// connection is not yet closing or closed.
```
pepone, line 156: The exceptions should be documented using `@throws`

cpp/include/Ice/ServantLocator.h
```diff
@@ -38,7 +38,7 @@ namespace Ice
         /// @param cookie The cookie that was returned by #locate.
         virtual void finished(const Current& curr, const ObjectPtr& servant, const std::shared_ptr<void>& cookie) = 0;
 
-        /// Notifies this servant locator that the object adapter in which it's installed is being deactivated.
+        /// Notifies this servant locator that the object adapter in which it's installed is being destroyed.
```
pepone, line 41: Is this update correct, who calls deactivate?

cpp/include/Ice/Communicator.h
bernardnormier: We use nullptr in C++. Maybe `@c nullptr` here?

bernardnormier: More importantly, I find the newly documented behavior very surprising, and inconsistent with how we use callbacks elsewhere. The expected behavior is destroyAsync destroys the communicator no matter what.

cpp/include/Ice/ServantLocator.h
bernardnormier: Called by ServerManager::destroy:
https://github.com/zeroc-ice/ice/blob/dfd3cabb321df1a5ba4e3dd9c4698b59e31b621e/cpp/src/Ice/ServantManager.cpp#L442

itself called by ObjectAdapterI::destroy:
https://github.com/zeroc-ice/ice/blob/dfd3cabb321df1a5ba4e3dd9c4698b59e31b621e/cpp/src/Ice/ObjectAdapterI.cpp#L310

So yes, it's correct.

cpp/include/Ice/StreamableTraits.h
```diff
@@ -76,7 +76,7 @@ namespace Ice
         VSize = 5,
 
         /// Fixed "size encoding" using 4 bytes followed by data.
-        /// Used by variable-size structs, and containers whose sizes can't be computed prior to unmarshaling.
+        /// Used by variable-size structs, and containers whose sizes can't be computed prior to marshaling.
```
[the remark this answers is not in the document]
bernardnormier: Need to fix the other language mappings doc

bernardnormier (approved):

cpp/include/Ice/Communicator.h
pepone: This is indeed very odd behavior. The more natural would be to either destroy the communicator anyway or throw if the callback is not present, but this just do nothing doesn't make much sense.

bernardnormier: See https://github.com/zeroc-ice/ice/pull/5885

bernardnormier: We can't throw on a null callback as we want destroyAsync to be noexcept.

externl (approved):

externl (approved):
Checked all 21 files hunk-by-hunk against the implementations — every corrected claim matches the code (setAdapter/throwException state machine, InputStream per-element size check, AlreadyRegisteredException kinds, ServantLocator.deactivate call path, Value slicing preservation, and the typo/@see/@snippet fixes), all Doxygen targets resolve, and the only non-comment line is inside the never-compiled ICE_DOXYGEN block.

One low nit: startValue's new wording says slices are re-marshaled 'only when this stream uses the sliced format', but the deciding factor is the current encapsulation's format (startEncapsulation can override the stream default, as the server reply path does) — still strictly better than the old unconditional claim.

cpp/include/Ice/SSL/ClientAuthenticationOptions.h
InsertCreativityHere: Yes we should also rename the snippet.

cpp/include/Ice/OutputStream.h
```diff
@@ -283,7 +284,7 @@ namespace Ice
         /// @tparam Te The types of the values in the tuple, starting at index @p I.
         /// @param tuple The tuple to marshal.
         // Declared here because the actual definition below breaks doxygen 1.13.2.
-        template<size_t I = 0, typename... Te >> writeAll(std::tuple<Te...> tuple);
+        template<size_t I = 0, typename... Te> void writeAll(std::tuple<Te...> tuple);
```
InsertCreativityHere, line 287: What is even the point of this `#ifdef ICE_DOXYGEN` section?

InsertCreativityHere (approved):
Looks good to me.

cpp/include/Ice/Initialize.h
```diff
@@ -112,13 +112,12 @@ namespace Ice
         return initialize(std::move(initData));
     }
 
-    /// Gets the per-process logger. This logger is used by all communicators that do not have their own specific logger
-    /// configured at the time the communicator is created.
+    /// Gets the per-process logger.
     /// @return The current per-process logger instance.
```
InsertCreativityHere, line 116: I'm fine with removing the double stating of "this logger is used by ...".

But maybe we should add an `@see` pointing to `setProcessLogger`, so a user browsing the docs will still see this explanation of what this function is all about.

cpp/include/Ice/OutputStream.h
InsertCreativityHere: Okay, I understand now. It's about our template functions.

cpp/include/Ice/Communicator.h
pepone: 🤖 **Claude:** Reworded to use `@c nullptr` and phrase it positively: the callback is called on completion when not null, and the communicator is destroyed regardless — matching #5885.

cpp/include/Ice/StreamableTraits.h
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Good catch. Weʼll sync the other mappings (Java/C#/Swift/JS `OptionalFormat`) in the per-language doc PRs so each stays clean, rather than crossing mappings in this C++ PR.

cpp/include/Ice/SSL/ClientAuthenticationOptions.h
pepone: 🤖 **Claude:** Done — renamed the snippet reference and both `//! [...]` markers to `clientCredentialsSelectionCallback`.

cpp/include/Ice/Connection.h
pepone: 🤖 **Claude:** Done — documented with `@throws CloseConnectionException` / `@throws ConnectionAbortedException`.

cpp/include/Ice/Initialize.h
pepone: 🤖 **Claude:** Added `@see setProcessLogger`.

PR #5868, opened 2026-07-07
Fix incorrect and stale API doc comments in the C# Ice library

pepone:
Fixes incorrect or stale XML doc comments in the C# Ice library, found during a C++/C# API doc review. Comment-only changes; no behavior change.

Incorrect behavior claims corrected:

- `ObjectPrx.ice_getContext`: never returns null — it returns a copy that is empty when there is no per-proxy context (also dropped the per-proxy/implicit conflation).
- `ObjectPrx.ice_identity` (interface): summary said "except for the per-proxy context" (copy-paste from `ice_context`).
- `ObjectPrxHelper`: not the "base class for all proxy helpers" — generated helpers derive from `ObjectPrxHelperBase`.
- `Connection.throwException`: `ConnectionClosedException` marks graceful closure, not abort; also throws while the connection is closing (matches the C++ doc).
- `Connection.setAdapter`: documents the `InvalidOperationException` thrown on incoming connections.
- `WSConnectionInfo.headers`: holds the upgrade *response* headers on outgoing connections.
- `OutputStream.writeValue`: deferral to `writePendingValues` is 1.0-only; with the 1.1 encoding the state is marshaled eagerly.
- `Value.ice_getSlicedData`: removed the stale "preserved-slice base class" condition (slice preservation is unconditional with the sliced format since 3.8).
- `Properties.parseIceCommandLineOptions`: the stale six-prefix list is replaced by the open-ended reserved-prefix wording (16 prefixes are converted today, including `--IceMX` and `--DataStorm`).
- `ServantLocator.deactivate`: called when the object adapter is destroyed, not deactivated.
- `ToStringMode.Compat`: only *non-printable* ASCII characters are escaped, matching the other two enumerators.
- `InputStream`/`OutputStream` class summaries: sealed classes described as "Interface for…"; `endEncapsulation` ends the *current* encapsulation; `clear()` no longer documents itself as "internally calls clear()".
- `Util.identityToString`: the mode governs non-ASCII handling too, not just non-printable ASCII (matches the C++ doc).
- `OptionalFormat.FSize`: "prior to marshaling", not unmarshaling.

Markup/typo fixes: `<returns>`/`<param>` tags nested inside `<summary>` on `ice_getIdentity`/`ice_identity` (they didn't render), `<exception name=...>` → `<exception cref=...>` (5 sites in `Proxy.cs`/`Properties.cs`), stray `>` after `</param>` in both `SliceTypeIdAttribute.cs` constructors, Doxygen-style `\\t` double backslashes in the `ToStringMode` docs (rendered literally in XML docs), a stray `>` in the `Properties` key-format example, "requests parameters" (×4), "nullptr" in `findDefaultServant`, and the `ServantLocator.finished` param doc copy-pasted from `locate`.

### DIFF
--- a/csharp/src/Ice/Connection.cs
+++ b/csharp/src/Ice/Connection.cs
@@ -63,6 +63,8 @@ public interface Connection
     /// the default object adapter of an outgoing connection is the communicator's default object adapter.
     /// </summary>
     /// <param name="adapter">The object adapter to associate with this connection.</param>
+    /// <exception cref="InvalidOperationException">Thrown when this connection is an incoming connection: you can
+    /// only call this method on outgoing (client) connections.</exception>
     /// <seealso cref="Communicator.getDefaultObjectAdapter"/>
     /// <seealso cref="getAdapter"/>
     void setAdapter(ObjectAdapter? adapter);
@@ -136,8 +138,8 @@ public interface Connection
     /// <summary>
     /// Throws an exception that provides the reason for the closure of this connection. For example,
     /// this method throws <see cref="CloseConnectionException"/> when the connection was closed gracefully by the peer;
-    /// It throws <see cref="ConnectionAbortedException"/> or <see cref="ConnectionClosedException"/>
-    /// when the connection is aborted. This method does nothing if the connection is not yet closed.
+    /// it throws <see cref="ConnectionAbortedException"/> when the connection is aborted with <see cref="abort"/>.
+    /// This method does nothing if the connection is not yet closing or closed.
     /// </summary>
     void throwException();
 }
@@ -339,7 +341,8 @@ public sealed class UDPConnectionInfo : IPConnectionInfo
 public sealed class WSConnectionInfo : ConnectionInfo
 {
     /// <summary>
-    /// The headers from the HTTP upgrade request.
+    /// The HTTP headers from the WebSocket upgrade handshake: the request headers for an incoming connection, and
+    /// the response headers for an outgoing connection.
     /// </summary>
     public readonly Dictionary<string, string> headers;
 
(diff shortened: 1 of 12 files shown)


csharp/src/Ice/Connection.cs
```diff
@@ -136,8 +138,8 @@ System.Threading.Tasks.Task flushBatchRequestsAsync(
     /// <summary>
     /// Throws an exception that provides the reason for the closure of this connection. For example,
     /// this method throws <see cref="CloseConnectionException"/> when the connection was closed gracefully by the peer;
-    /// It throws <see cref="ConnectionAbortedException"/> or <see cref="ConnectionClosedException"/>
-    /// when the connection is aborted. This method does nothing if the connection is not yet closed.
+    /// it throws <see cref="ConnectionAbortedException"/> when the connection is aborted with <see cref="abort"/>.
+    /// This method does nothing if the connection is not yet closing or closed.
     /// </summary>
     void throwException();
```
pepone: The convention in C# is to document exceptions using `<exception` tag.

bernardnormier (approved):

externl (approved):

externl (approved):
Verified every rewritten xmldoc against the C# implementations — setAdapter's InvalidOperationException, the throwException closing/closed taxonomy, WSTransceiver request-vs-response parsing, the 1.0-vs-1.1 writeValue behavior, PropertyException on all four Properties members, the 16 reserved prefixes, and createProxy's ParseException all check out; every cref target resolves and no behavior changes hide in the diff.

Optional: the ObjectPrx.ice_identity interface doc (what IDE tooltips show) could carry over the ArgumentException tag the ObjectPrxHelperBase impl doc already documents, since this PR rewrote exactly that block.

csharp/src/Ice/Connection.cs
pepone: 🤖 **Claude:** Done — documented with `<exception>` tags.

PR #5869, opened 2026-07-08
Surface DataStorm configuration and API misuse errors as catchable failures

### LINKED ISSUE #5823: [Audit-Medium] DataStorm: Node construction failure after Instance creation calls std::terminate
> **AI-generated finding** — surfaced by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), in a multi-agent correctness audit of DataStorm (five subsystem reviewers + three independent adversarial verifiers). Code is pinned to `origin/main` at [`90dc6c7`](https://github.com/zeroc-ice/ice/commit/90dc6c7171184e513f40114b36d1aa2dbf130c49). **Not yet human-triaged, and not independently cross-checked by Codex** — please verify before acting.

## Summary

If an exception escapes `Instance::init()` after the `Instance` (and its `Timer` / `CallbackExecutor` worker threads) has been constructed, the `Node` constructor's catch destroys only the communicator and rethrows — it never calls `Instance::destroy()`:
https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/Node.cpp#L40

Because every servant holds a `weak_ptr<Instance>`, the stack unwind drops the last reference and runs `~Instance`, which destroys the `Timer` whose `std::thread _worker` is still **joinable** (`Timer::destroy()`, which joins, was never called) → `std::terminate`. `CallbackExecutor` (thread started in its ctor, no destructor join) is a second independent terminate source.

## Trigger

A malformed proxy property is the easy path: `--DataStorm.Node.ConnectTo=<bad endpoint>` throws `ParseException` while constructing `LookupPrx{communicator, "DataStorm/Lookup:" + connectTo}` — evaluated as an argument **outside** `connect()`'s try block:
https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/NodeSessionManager.cpp#L82

A malformed `DataStorm.Node.Multicast.Proxy` is a second trigger.

## Impact

A configuration typo aborts the process via `std::terminate` instead of surfacing the catchable exception the `Node` constructor API promises (and, for a non-owned communicator, leaves it polluted — default object adapter still set, DataStorm adapters alive).

## Fix direction

In the `Node` ctor catch (or via an RAII guard around `Instance` creation), call `_instance->destroy(...)` — which stops/joins the timer and executor and clears the default object adapter — before rethrowing. Alternatively, parse `ConnectTo`/`Multicast.Proxy` inside a try that maps to the same catchable failure as the adapter-creation errors.

### LINKED ISSUE #5828: [Audit-Low] DataStorm: invalid DataStorm.Topic.* property values call std::terminate via noexcept lazy topic creation

## Summary

`Topic::hasWriters`, `hasReaders`, `setReaderDefaultConfig`, and `setWriterDefaultConfig` are declared `noexcept` and lazily construct the `TopicReaderI`/`TopicWriterI`, whose constructors call `parseConfig()` → `parseClearHistory`/`parseDiscardPolicy`, which throw `ParseException` on invalid property values:

https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/TopicI.cpp#L66

An exception crossing a `noexcept` boundary → `std::terminate`.

## Trigger

`--DataStorm.Topic.ClearHistory=onAdd` (case typo), a non-numeric `SampleCount`, or the very natural `--DataStorm.Topic.DiscardPolicy=None` — the C++ enum is `DiscardPolicy::None` but the parser only accepts `"Never"` — followed by a first call to one of the four `noexcept` methods above.

## Impact

A misconfiguration aborts the process with no diagnosable exception, instead of the catchable `ParseException` the common first calls (`makeSingleKeyReader`/`makeSingleKeyWriter`) would surface. The `None`-vs-`Never` naming mismatch actively invites the typo. Low-Medium.

## Fix direction

Parse the topic defaults eagerly from a throwing context (Node/TopicFactory construction), or drop `noexcept` from the lazily-creating `Topic` methods; and accept `"None"` as an alias for the `Never`/`None` discard policy.

### LINKED ISSUE #5829: [Audit-Low] DataStorm: Node::getSessionConnection is noexcept but stringToIdentity throws on malformed input

## Summary

`Node::getSessionConnection(std::string_view)` is declared `noexcept`:

https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/Node.cpp#L114

but it delegates to `NodeI::getSessionConnection`, which calls `Ice::stringToIdentity(id)` — that throws `ParseException` on an unescaped `/` or a bad escape. No catch intervenes.

## Impact

A malformed identity string passed to this otherwise fail-soft accessor (it returns `nullptr` for an unknown session or a moved-from node) instead crosses the `noexcept` boundary → `std::terminate`.

## Fix direction

Wrap the parse in try/catch and return `nullptr` on a malformed identity, or move the parse off the `noexcept` surface.

### LINKED ISSUE #5830: [Audit-Low] DataStorm: moved-from Node::shutdown/isShutdown/waitForShutdown dereference a null _instance

## Summary

`Node::shutdown`, `isShutdown`, and `waitForShutdown` dereference `_instance` without a null check:

https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/Node.cpp#L74

while the sibling accessors `getCommunicator` and `getSessionConnection`, the destructor, and move-assignment all guard against a null `_instance` (`Node.cpp:108-118`). The class clearly intends to tolerate the moved-from state, so this is an inconsistency.

## Trigger / impact

`Node b = std::move(a); a.shutdown();` → null dereference (→ `std::terminate`) on a moved-from object that the other members handle gracefully.

## Fix direction

Add the same `_instance ?` guard to `shutdown`/`isShutdown`/`waitForShutdown` (no-op / return `false` / return immediately), or document a consistent moved-from restriction across all members.

pepone:
Fixes #5823, fixes #5828, fixes #5829, fixes #5830.

Several DataStorm error paths aborted the process via `std::terminate` instead of reporting the error to the
application:

- **#5823** — A `Node` constructor failure after the `Instance` was created (e.g. an invalid
  `DataStorm.Node.ConnectTo` endpoint, or a malformed `DataStorm.Node.Multicast.Proxy`) destroyed only the
  communicator; the stack unwind then dropped the last `Instance` reference and destructed the still-joinable
  `Timer` and `CallbackExecutor` worker threads → `std::terminate`. The constructor catch now calls
  `Instance::destroy`, which joins both threads and handles the communicator/adapter teardown for both ownership
  modes; `Instance::destroy` now tolerates an instance whose `init()` did not complete (`_node` guard).
- **#5828** — The `noexcept` `Topic` methods (`hasWriters`, `hasReaders`, `setReaderDefaultConfig`,
  `setWriterDefaultConfig`) lazily created the topic reader/writer, whose constructor parsed the
  `DataStorm.Topic.*` properties and threw `ParseException` on an invalid value — across the `noexcept` boundary →
  `std::terminate`. The topic defaults are now parsed **eagerly by the `Node` constructor** (before any resource or
  thread is created), so a bad value surfaces there as a catchable exception and the lazy paths no longer parse.
  `DataStorm.Topic.DiscardPolicy` additionally accepts `None` as an alias for `Never`, matching the
  `DiscardPolicy::None` enumerator that invites the typo.
- **#5829** — `Node::getSessionConnection` is `noexcept` but `Ice::stringToIdentity` throws on a malformed
  identity. The parse failure now returns null, consistent with the accessor's fail-soft contract.
- **#5830** — `Node::shutdown`/`isShutdown`/`waitForShutdown` dereferenced a null `_instance` on a moved-from node
  while every sibling member guards it. They are now no-ops; `isShutdown` reports a moved-from node as shut down
  (it is inert and can never transition, keeping `waitForShutdown` ≡ "wait until `isShutdown()`" consistent).

Behavior note: `DataStorm.Topic.*` defaults are now read once at `Node` construction instead of at the first topic
reader/writer creation, so a misconfigured value fails the `Node` constructor even if the application never creates
a topic. Per-topic overrides via `setReaderDefaultConfig`/`setWriterDefaultConfig` are unaffected.

## Testing

New cases in `test/DataStorm/api`, each verified red→green:

- invalid `DataStorm.Node.ConnectTo` → `ParseException` from the `Node` constructor, with and without communicator
  ownership (without the fix the process aborts via `std::terminate` during unwinding);
- invalid `DataStorm.Topic.ClearHistory` → `ParseException`, and non-numeric `DataStorm.Topic.SampleCount` →
  `PropertyException`, both from the `Node` constructor;
- `DataStorm.Topic.DiscardPolicy=None` accepted;
- `getSessionConnection("a/b/c")` returns null (without the fix: `libc++abi: terminating due to uncaught exception
  of type Ice::ParseException`);
- `shutdown`/`isShutdown`/`waitForShutdown` on a moved-from node (without the fix: SIGSEGV).

The full DataStorm suite passes. A changelog fragment is included.

### DIFF
--- a/changelog.d/datastorm/5823.md
+++ b/changelog.d/datastorm/5823.md
@@ -1,8 +1,0 @@
-- Fixed several DataStorm error-handling bugs that aborted the process (`std::terminate`) instead of reporting the
-  error to the application: a `Node` constructor failure after a configuration error (such as an invalid
-  `DataStorm.Node.ConnectTo` endpoint) left worker threads running during unwinding; an invalid `DataStorm.Topic.*`
-  property value threw from the `noexcept` `Topic` methods that lazily create the topic reader or writer — the topic
-  defaults are now parsed by the `Node` constructor, so a bad value surfaces there as a catchable exception; and
-  `Node::getSessionConnection` threw on a malformed identity instead of returning null. In addition, the `Node`
-  shutdown members are now safe on a moved-from node, and the `DataStorm.Topic.DiscardPolicy` property accepts
-  `None` as an alias for `Never`, matching the `DiscardPolicy::None` enumerator.
--- a/changelog.d/datastorm/5823.md
+++ b/changelog.d/datastorm/5823.md
@@ -1,0 +1,8 @@
+- Fixed several DataStorm error-handling bugs that aborted the process (`std::terminate`) instead of reporting the
+  error to the application: a `Node` constructor failure after a configuration error (such as an invalid
+  `DataStorm.Node.ConnectTo` endpoint) left worker threads running during unwinding; an invalid `DataStorm.Topic.*`
+  property value threw from the `noexcept` `Topic` methods that lazily create the topic reader or writer — the topic
+  defaults are now parsed by the `Node` constructor, so a bad value surfaces there as a catchable exception; and
+  `Node::getSessionConnection` threw on a malformed identity instead of returning null. In addition, the `Node`
+  shutdown members are now safe on a moved-from node, and the `DataStorm.Topic.DiscardPolicy` property accepts
+  `None` as an alias for `Never`, matching the `DiscardPolicy::None` enumerator.
--- a/cpp/src/DataStorm/Instance.cpp
+++ b/cpp/src/DataStorm/Instance.cpp
@@ -277,8 +277,6 @@ Instance::destroy(bool ownsCommunicator)
     }
     if (_node)
     {
-        // _node is only set once init() has progressed far enough; it is null when the Node constructor destroys a
-        // partially-initialized instance after an init() failure.
         _node->destroy(ownsCommunicator);
     }
 
--- a/cpp/src/DataStorm/Instance.cpp
+++ b/cpp/src/DataStorm/Instance.cpp
@@ -14,6 +14,58 @@ using namespace std;
 using namespace DataStormI;
 using namespace Ice;
 
+namespace
+{
+    DataStorm::ClearHistoryPolicy parseClearHistory(const std::string& value)
+    {
+        if (value == "OnAdd")
+        {
+            return DataStorm::ClearHistoryPolicy::OnAdd;
+        }
+        else if (value == "OnRemove")
+        {
+            return DataStorm::ClearHistoryPolicy::OnRemove;
+        }
+        else if (value == "OnAll")
+        {
+            return DataStorm::ClearHistoryPolicy::OnAll;
+        }
+        else if (value == "OnAllExceptPartialUpdate")
+        {
+            return DataStorm::ClearHistoryPolicy::OnAllExceptPartialUpdate;
+        }
+        else if (value == "Never")
+        {
+            return DataStorm::ClearHistoryPolicy::Never;
+        }
+        else
+        {
+            throw ParseException(__FILE__, __LINE__, "Invalid clear history policy: " + value);
+        }
+    }
+
+    DataStorm::DiscardPolicy parseDiscardPolicy(const std::string& value)
+    {
+        // "None" is accepted as an alias for "Never" to match the DataStorm::DiscardPolicy::None enumerator name.
+        if (value == "Never" || value == "None")
+        {
+            return DataStorm::DiscardPolicy::None;
+        }
+        else if (value == "SendTime")
+        {
+            return DataStorm::DiscardPolicy::SendTime;
+        }
+        else if (value == "Priority")
+        {
+            return DataStorm::DiscardPolicy::Priority;
+        }
+        else
+        {
+            throw ParseException(__FILE__, __LINE__, "Invalid discard policy: " + value);
+        }
+    }
+}
+
 Instance::Instance(
     CommunicatorPtr communicator,
     function<void(function<void()> call)> customExecutor,
@@ -27,6 +79,21 @@ Instance::Instance(
     }
 
     PropertiesPtr properties = _communicator->getProperties();
+
+    // Parse the topic default configurations eagerly, before any resource is created: an invalid DataStorm.Topic.*
+    // property value then surfaces as a catchable exception from the Node constructor. Parsing these lazily when a
+    // topic reader or writer is first created would let the exception escape the noexcept Topic methods
+    // (hasWriters/hasReaders/setReaderDefaultConfig/setWriterDefaultConfig) and call std::terminate.
+    _defaultReaderConfig.clearHistory = parseClearHistory(properties->getIceProperty("DataStorm.Topic.ClearHistory"));
+    _defaultReaderConfig.sampleCount = properties->getIcePropertyAsInt("DataStorm.Topic.SampleCount");
+    _defaultReaderConfig.sampleLifetime = properties->getIcePropertyAsInt("DataStorm.Topic.SampleLifetime");
+    _defaultReaderConfig.discardPolicy =
+        parseDiscardPolicy(properties->getIceProperty("DataStorm.Topic.DiscardPolicy"));
+
+    _defaultWriterConfig.clearHistory = *_defaultReaderConfig.clearHistory;
+    _defaultWriterConfig.sampleCount = *_defaultReaderConfig.sampleCount;
+    _defaultWriterConfig.sampleLifetime = *_defaultReaderConfig.sampleLifetime;
+    _defaultWriterConfig.priority = properties->getIcePropertyAsInt("DataStorm.Topic.Priority");
 
     if (properties->getIcePropertyAsInt("DataStorm.Node.Server.Enabled") > 0)
     {
@@ -208,7 +275,12 @@ Instance::destroy(bool ownsCommunicator)
             _multicastAdapter->destroy();
         }
     }
-    _node->destroy(ownsCommunicator);
+    if (_node)
+    {
+        // _node is only set once init() has progressed far enough; it is null when the Node constructor destroys a
+        // partially-initialized instance after an init() failure.
+        _node->destroy(ownsCommunicator);
+    }
 
     _executor->destroy();
     _connectionManager->destroy();
--- a/cpp/src/DataStorm/Instance.h
+++ b/cpp/src/DataStorm/Instance.h
@@ -56,11 +56,15 @@ namespace DataStormI
             return _adapter;
         }
 
-        // The topic default configurations are parsed from the DataStorm.Topic.* properties by the constructor, so an
-        // invalid property value surfaces as a catchable exception from the Node constructor rather than from the
-        // noexcept Topic methods that lazily create the topic reader or writer.
-        [[nodiscard]] const DataStorm::ReaderConfig& getDefaultReaderConfig() const { return _defaultReaderConfig; }
-        [[nodiscard]] const DataStorm::WriterConfig& getDefaultWriterConfig() const { return _defaultWriterConfig; }
+        [[nodiscard]] const DataStorm::ReaderConfig& getDefaultReaderConfig() const noexcept
+        {
+            return _defaultReaderConfig;
+        }
+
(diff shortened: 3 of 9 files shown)


changelog.d/datastorm/5823.md
```diff
@@ -0,0 +1,8 @@
+- Fixed several DataStorm error-handling bugs that aborted the process (`std::terminate`) instead of reporting the
+  error to the application: a `Node` constructor failure after a configuration error (such as an invalid
+  `DataStorm.Node.ConnectTo` endpoint) left worker threads running during unwinding; an invalid `DataStorm.Topic.*`
+  property value threw from the `noexcept` `Topic` methods that lazily create the topic reader or writer — the topic
+  defaults are now parsed by the `Node` constructor, so a bad value surfaces there as a catchable exception; and
+  `Node::getSessionConnection` threw on a malformed identity instead of returning null. In addition, the `Node`
+  shutdown members are now safe on a moved-from node, and the `DataStorm.Topic.DiscardPolicy` property accepts
+  `None` as an alias for `Never`, matching the `DiscardPolicy::None` enumerator.
```
pepone: We should consider spliting this entry and simplify it.

We can use `datastorm/<issue-number>-<slug-fix>.md`  and have multiple changelog fragments one per fix.

cpp/src/DataStorm/Instance.cpp
```diff
@@ -208,7 +275,12 @@ Instance::destroy(bool ownsCommunicator)
             _multicastAdapter->destroy();
         }
     }
-    _node->destroy(ownsCommunicator);
+    if (_node)
+    {
+        // _node is only set once init() has progressed far enough; it is null when the Node constructor destroys a
+        // partially-initialized instance after an init() failure.
```
pepone: The code is clear enough, no need for this comment.

cpp/src/DataStorm/Instance.h
```diff
@@ -55,6 +56,12 @@ namespace DataStormI
             return _adapter;
         }
 
+        // The topic default configurations are parsed from the DataStorm.Topic.* properties by the constructor, so an
+        // invalid property value surfaces as a catchable exception from the Node constructor rather than from the
+        // noexcept Topic methods that lazily create the topic reader or writer.
+        [[nodiscard]] const DataStorm::ReaderConfig& getDefaultReaderConfig() const { return _defaultReaderConfig; }
+        [[nodiscard]] const DataStorm::WriterConfig& getDefaultWriterConfig() const { return _defaultWriterConfig; }
+
```
pepone: I would remove this comment to, we should consider make the methods noexcept which makes things clear.

cpp/src/DataStorm/Node.cpp
```diff
@@ -39,7 +39,14 @@ Node::Node(NodeOptions options)
     }
     catch (...)
     {
-        if (_ownsCommunicator)
+        if (_instance)
+        {
+            // Destroy the instance: this joins the timer and callback executor threads started by the Instance
+            // constructor (destroying them still running would call std::terminate) and takes care of the
+            // communicator and adapter teardown for both ownership modes.
```
pepone: This comment is too verbose, I would remove it, looking at the code is clear that instance must be destroyed.

cpp/src/DataStorm/Node.cpp
```diff
@@ -74,19 +81,26 @@ Node::~Node()
...
-    return _instance->isShutdown();
+    // A moved-from node is inert and can never transition to shutdown; report it as already shut down.
+    return _instance ? _instance->isShutdown() : true;
 }
 
 void
 Node::waitForShutdown() const noexcept
 {
-    _instance->waitForShutdown();
+    if (_instance)
```
pepone, line 96: ditto , this is just defensive

externl (changes requested):
The four fixes check out — I traced the ctor/init/destroy ordering (destroy only touches ctor-created members plus the now-guarded `_node`), the eager topic parse, and the `noexcept` chains for `getSessionConnection`/moved-from `Node`. One blocker before merge:

- The new moved-from-node test trips the required clang-tidy check: `test/DataStorm/api/Writer.cpp:36` calls `n.shutdown()` after `Node nm(std::move(n))` → `clang-analyzer-cplusplus.Move`. Since the moved-from use is the point of the test, wrap the calls in `NOLINTBEGIN/END(clang-analyzer-cplusplus.Move)` or `NOLINTNEXTLINE` (precedent: `cpp/test/DataStorm/types/Writer.cpp:27`).

Two optional low notes: if the `Instance` constructor throws after `setDefaultObjectAdapter` on a user-supplied communicator, the adapter/default-adapter setting still leak (pre-existing, narrowed not closed by this PR); and the new moved-from/fail-soft semantics live only in .cpp comments — the `Node.h` docs don't mention them.

externl (approved):
The four fixes check out — I traced the ctor/init/destroy ordering (destroy only touches ctor-created members plus the now-guarded `_node`), the eager topic parse, and the `noexcept` chains for `getSessionConnection` and the moved-from `Node` guards. All the DataStorm suites pass on every platform.

Minor, non-blocking: the new moved-from test trips clang-tidy at `test/DataStorm/api/Writer.cpp:36` (`clang-analyzer-cplusplus.Move`) — a `NOLINTNEXTLINE` there (as in `types/Writer.cpp:27`) will green the check.

changelog.d/datastorm/5823.md
bernardnormier: It's actually simpler. A single changelog fragment can have multiple "bullets".

bernardnormier (approved):
LGTM. One thing to fix before merging: per changelog.d/README.md, a fragment is named after the issue number only when the PR fixes a single public issue — this PR fixes four, so 5823.md should get a descriptive slug instead (e.g. node-config-error-handling.md).

changelog.d/datastorm/5823.md
pepone: 🤖 **Claude:** Renamed to `node-config-error-handling.md` with one bullet per fix (single fragment, per Bernard). Done in 838f1e6.

cpp/src/DataStorm/Instance.cpp
pepone: 🤖 **Claude:** Removed. Done in 838f1e6.

cpp/src/DataStorm/Instance.h
pepone: 🤖 **Claude:** Comment removed and both getters are now `noexcept`. Done in 838f1e6.

cpp/src/DataStorm/Node.cpp
pepone: 🤖 **Claude:** Removed. Done in 838f1e6.

cpp/src/DataStorm/Node.cpp
pepone: 🤖 **Claude:** Removed. Done in 838f1e6.

pepone:
🤖 **Claude:** Review comments addressed in 838f1e6:

- **clang-tidy** (@externl): wrapped the moved-from calls in `NOLINTBEGIN/END(clang-analyzer-cplusplus.Move)` — the moved-from use is the point of the test.
- **Changelog** (@bernardnormier): fragment renamed to `node-config-error-handling.md` and split into one bullet per fix.
- Comment trims + `noexcept` config getters per the inline comments.
- The `Instance`-constructor pollution note is now tracked as #5891 (residual throw sites after `setDefaultObjectAdapter` on a user-supplied communicator). No `Node.h` doc changes for the moved-from members: the guards are a safety net, not a supported usage.

PR #5870, opened 2026-07-08
Match the key when a DataStorm multi-key writer forwards a sample

### LINKED ISSUE #5825: [Audit-Low] DataStorm: multi-key writer forwards every key to every session, tripping the s() key assert

## Summary

`KeyDataWriterI::forward` passes `matchOne(_sample, _keys.empty())`; for a keyed **multi-key** writer `_keys.empty()` is false, so the per-key match is skipped and the writer forwards **every** key's sample to every attached session:

https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/DataElementI.cpp#L1380

A session whose reader subscribed only key `k1` therefore receives an `s()` for `k2`, whose keyId it never registered; the handler default-inserts a null key and asserts:

https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/SessionI.cpp#L1391 (then `assert(key)` at `SessionI.cpp:1397`)

## Trigger / impact

`makeMultiKeyWriter({k1, k2})` + `makeSingleKeyReader(k1)` + `writer.update(k2, …)`. Debug builds crash on the assert on every unmatched-key publish; release drops the sample at the per-element guard (correct outcome) after a wasted decode and leaves one permanent `(nullptr, {})` stub in the subscriber's key map. Also wastes bandwidth (full fan-out of unmatched keys). The init path is unaffected (`getSamples` filters by key).

## Fix direction

Guard the lookup (`keys.find` + early return) instead of `operator[]` + `assert`; consider passing `matchKey = true` in `KeyDataWriterI::forward` so unmatched keys are not forwarded at all.

pepone:
Fixes #5825.

`KeyDataWriterI::forward` only key-matched the subscribers of any-key writers (`matchOne(_sample, _keys.empty())`),
so a keyed **multi-key** writer forwarded every key's sample to every attached session — including sessions whose
readers subscribe only some of the writer's keys. On such a session the `s()` dispatch looked up the
never-registered key id with `operator[]`, default-inserting a null key: debug builds died on `assert(key)` on
every unmatched-key publish, and release builds dropped the sample only after the wasted forward and unmarshaling,
leaving a permanent null stub in the subscriber's key map.

## Fix

- `KeyDataWriterI::forward` now always matches the key when forwarding. Any-key readers keep an empty key set in
  their subscriber entry and still match everything; filtered readers are matched by their filter as before, so
  only genuinely unmatched keys stop being forwarded (which also saves the wasted bandwidth).
- The `s()` dispatch looks up the key id fail-soft: a sample whose key was never subscribed is traced and
  discarded instead of asserting — such samples can still arrive from peers without the writer-side fix.

The base `DataElementI::forward` (which never has `_sample` set outside `KeyDataWriterI::send`) is unchanged.

## Testing

New case in `test/DataStorm/events`: a multi-key writer `{elem1, elem2}` publishes `elem2` first while the peer
session's only reader subscribes `elem1`; the reader must receive only `elem1`'s sample. In a debug build without
the fix the reader aborts: `Assertion failed: (key), SessionI.cpp` (verified red→green with asserts enabled; the
default release build drops the sample and passes either way). The full DataStorm suite passes.

No changelog fragment: the crash is debug-only and there are no field reports.

### DIFF
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -1559,8 +1559,10 @@ KeyDataWriterI::forward(const ByteSeq& inParams, const Current& current) const
 {
     for (const auto& [_, listener] : _listeners)
     {
-        // If there's at least one subscriber interested in the update (check the key if any writer)
-        if (!_sample || listener.matchOne(_sample, _keys.empty()))
+        // Forward the sample if the listener has at least one subscriber interested in the update. The key is
+        // always matched: a multi-key writer's sessions don't necessarily subscribe to every key of the writer,
+        // and an unmatched key's sample would be wasted bandwidth at best (the receiver never subscribed its id).
+        if (!_sample || listener.matchOne(_sample, true))
         {
             // Forward the call using the listener's session proxy, don't need to wait for the result.
             listener.proxy
--- a/changelog.d/datastorm/5825.md
+++ b/changelog.d/datastorm/5825.md
@@ -1,0 +1,3 @@
+- Fixed a bug in DataStorm where a writer created for several keys sent every key's samples to all subscribed
+  sessions, including sessions whose readers subscribe to only some of the writer's keys: the unmatched samples
+  wasted bandwidth, and debug builds crashed on an assertion when receiving them.
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -1423,7 +1423,20 @@ SubscriberSessionI::s(int64_t topicId, int64_t elementId, DataSample dataSample,
                 shared_ptr<Key> key;
                 if (dataSample.keyValue.empty())
                 {
-                    key = topicSubscriber.keys[dataSample.keyId].first;
+                    auto q = topicSubscriber.keys.find(dataSample.keyId);
+                    if (q == topicSubscriber.keys.end())
+                    {
+                        // The session never subscribed to this key (a peer can forward a sample for a key none of
+                        // this session's readers are subscribed to); no subscriber can match it.
+                        if (_traceLevels->session > 2)
+                        {
+                            Trace out(_traceLevels->logger, _traceLevels->sessionCat);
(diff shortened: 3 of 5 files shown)


cpp/src/DataStorm/DataElementI.cpp
pepone, line 1: This PR needs a changelog fragment in changelog.d to describe how this affects DataStorm applications.

externl (approved):
Verified matchOne's key-set semantics end-to-end: subscriber->keys is populated with canonicalized per-topic Key instances by both attachKey and attachFilter for every matched bucket, so keyed, filtered, and any-key subscribers all still receive exactly what they subscribe to. The i()/initSamples path is per-attach key-filtered so it can never carry an unmatched key even from an unfixed peer, and the s() fail-soft return only skips the current topic subscriber. The new events case is a genuine red→green for the debug assert, and it's a net perf win (one set::find replaces a full unmatched ice_invokeAsync).

One low gap: no test pairs an any-key reader with a keyed multi-key writer receiving all keys — correctness there rests on code reading (one attachFilter per key bucket), so a one-block test would guard against future attach-machinery changes.

bernardnormier (approved):
Looks good

cpp/src/DataStorm/DataElementI.cpp
pepone: 🤖 **Claude:** Added `changelog.d/datastorm/5825.md` describing the application effects (unmatched samples wasted bandwidth; debug builds crashed on an assertion). Done in 43a7379.

pepone:
🤖 **Claude:** Updated in 43a7379 (rebased onto main — the events tests conflicted with the merged #5866 case):

- **Changelog** (@pepone): fragment added with the application-effect wording.
- **Any-key coverage** (@externl): the `unmatchedKey` case now also runs an any-key reader alongside the single-key
  reader — it receives both keys of the multi-key writer in publication order, guarding the live forward path
  (`matchOne(_sample, true)` with an empty key set) against future attach-machinery changes.

PR #5872, opened 2026-07-08
Detach the elements of a destroyed DataStorm topic when the session is disconnected

### LINKED ISSUE #5821: [Audit-Medium] DataStorm: topic destroy racing session disconnect leaks element listeners (assert / waitForNoWriters hang)
> **AI-generated finding** — surfaced by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), in a multi-agent correctness audit of DataStorm (five subsystem reviewers + three independent adversarial verifiers; found independently by two reviewers). Code is pinned to `origin/main` at [`90dc6c7`](https://github.com/zeroc-ice/ice/commit/90dc6c7171184e513f40114b36d1aa2dbf130c49). **Not yet human-triaged, and not independently cross-checked by Codex** — please verify before acting.

## Summary

A `Topic` destruction racing a connection loss can leave a session's element listeners attached forever, because both detach paths bail out.

`TopicI::destroy` sets `_destroyed` and releases the topic mutex **before** calling `disconnect()`:
https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/TopicI.cpp#L146

A concurrent `SessionI::disconnected` — which naturally blocks on that same topic mutex during destroy — then **skips** the destroyed topic in `runWithTopics` (the `isDestroyed()` check) and nulls `_session`; `TopicI::disconnect` → `session->disconnect(topicId, this)` then early-returns on `!_session`:
https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/SessionI.cpp#L956

## Trigger

Destroy a topic reader/writer while its session's connection is dropping (both critical sections need only run in the order: `TopicI::destroy` body → `SessionI::disconnected` → `TopicI::disconnect`). Elements legitimately outlive a destroyed topic — this is the same state the earlier detach fix (zeroc-ice/ice#5599) targets.

## Impact

Neither path runs `detachKey`/`detachFilter`, so the elements keep their `Listener` entries and `_listenerCount > 0`:

- **Debug builds** abort on `assert(_listenerCount == 0)` at the end of `TopicI::disconnect`.
- **Release builds**: surviving elements report `hasWriters()`/`hasReaders()` true with no peer, hang in `waitForNoWriters()`/`waitForNoReaders()`, never fire the `onConnected*` Disconnect callback, and pin the session servant via the listener `shared_ptr`s — until each element is individually destroyed.

The zeroc-ice/ice#5599 fix block (`SessionI.cpp:969-978`) only covers the still-`_session`-connected entry point, not this interleaving.

## Fix direction

In `SessionI::disconnect`, run the unsubscribe when the topic-subscriber entry exists rather than gating solely on `_session` (the peer session is gone either way; `unsubscribe` only mutates local state), or make `disconnected()`/`destroyImpl` detach elements of destroyed topics too.

pepone:
Fixes #5821.

`SessionI::disconnect` — called by `TopicI::destroy` via `TopicI::disconnect` — contains the cleanup that detaches
the elements of a destroyed topic directly, because the session's own detach paths (`runWithTopics`) skip destroyed
topics. That cleanup was gated behind an early return on `!_session`, so when a connection loss won the race
against the topic destruction — `disconnected()` skips the already-marked-destroyed topic, then nulls `_session` —
nothing detached the topic's elements. Their listener entries and the topic's listener count stayed behind:

- debug builds abort on `assert(_listenerCount == 0)` at the end of `TopicI::disconnect`;
- release builds keep `hasWriters()`/`hasReaders()` true for a peer that is gone, hang
  `waitForNoWriters()`/`waitForNoReaders()`, never fire the `onConnected*` disconnect callbacks, and pin the
  session servant through the listener `shared_ptr`s until each element is individually destroyed.

## Fix

The cleanup now runs whenever the session still holds the topic's subscriber entry, regardless of `_session`. The
entry is absent in every state where the cleanup already happened: `TopicI::detach` (driven by a disconnect that
saw the topic before it was destroyed) erases the session from the topic's listeners so this path isn't reached
again, a peer `detachTopic` erases the entry first, and `destroyImpl` clears the map.

Two narrower siblings of this race — a peer `detachTopic` or `destroyImpl` erasing the subscriber entry in the same
window (both require the peer/node teardown to land between `_destroyed = true` and `TopicI::disconnect`) — are
deliberately not covered here and are tracked separately; a uniform fix needs care because a double element detach
is not safe (`Listener::get` dereferences an unchecked `find()`).

## Testing

No new test: the race window is between `TopicI::destroy` marking the topic destroyed and `TopicI::disconnect`
taking the session lock, with the connection-close notification landing in between — not drivable deterministically
from the public API. The full DataStorm suite passes. A changelog fragment is included.

### DIFF
--- a/changelog.d/datastorm/5821.md
+++ b/changelog.d/datastorm/5821.md
@@ -1,0 +1,3 @@
+- Fixed a bug in DataStorm where destroying a topic while its session's connection was closing could leave the
+  topic's readers or writers permanently attached to the disconnected peer: `hasWriters()`/`hasReaders()` stayed
+  true, `waitForNoWriters()`/`waitForNoReaders()` hung, and the disconnect callbacks never fired.
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -961,15 +961,15 @@ void
 SessionI::disconnect(int64_t topicId, TopicI* topic)
 {
     lock_guard<mutex> lock(_mutex); // Called by TopicI::destroy
-    if (!_session)
-    {
-        return;
-    }
 
+    // No _session check: a session disconnected after the topic was marked destroyed skipped this topic when
+    // detaching (runWithTopics skips destroyed topics), so the subscriber entry must be cleaned up here even when
+    // the session is gone. The checks below cover every other state: the entry is absent when the peer detached the
+    // topic first or the session was destroyed.
     auto t = _topics.find(topicId);
     if (t == _topics.end())
     {
-        return; // Peer topic detached first.
+        return; // Peer topic detached first, or the session was destroyed.
     }
     auto& subscriber = t->second;
 


externl (approved):
The removed !_session gate was import-era boilerplate that became wrong once #5599 made SessionI::disconnect the sole element-detach path for destroyed topics; the cleanup block, its session→topic lock order, and its double-detach guards (the TopicI::_listeners swap gating the single caller, plus DataElementI::detachKey's listeners.find early-return) are unchanged and safe for every _session-null state I could construct.

One note: there's a third sibling of the races deferred to #5873 that isn't listed there — a reconnect landing in the destroy window lets the announceTopics reap (keyed on the incremented _sessionInstanceId) erase the topic's subscriber entry before TopicI::disconnect runs, so SessionI::disconnect early-returns at 't == _topics.end()' and hits the same leak/assert. Worth adding to #5873; even narrower than the two already there.

bernardnormier (approved):
Looks good.

pepone:
🤖 **Claude:** Verified the reconnect-reap sibling against the source (the `announceTopics` reap erases the
destroyed topic's subscriber entry without the local detach, since the name-based `runWithTopics` skips destroyed
topics and the stale instance id makes `reap()` drop the whole entry) — added it to #5873 as sibling 3.

PR #5880, opened 2026-07-08
Deliver every per-key DataStorm initialization batch to the matching readers

### LINKED ISSUE #5819: [Audit-High] DataStorm: per-key init-sample batches silently dropped (flow 1) or applied to the wrong reader (flow 2)

## Summary

## Impact

## Fix direction

pepone:
Fixes #5819.

A writer element covering several keys produces one initialization batch per key **and per attached reader
element**, all carrying only the writer element id — nothing on the wire identifies which reader or key a batch is
for. The `initSamples` dispatch mishandled this two ways:

- **Flow 1 — dropped batches**: every uninitialized subscriber was marked `initialized` on the *first* batch, so
  all later batches of the same round were skipped. A late-joining multi-key (or any-key) reader silently lost
  every key's initialization samples but one (reproduced at runtime in the issue).
- **Flow 2 — misrouted batches**: batches fanned out to every uninitialized reader element of the writer with
  `checkKey = samples.id < 0` — false for a keyed writer — so a reader element could queue **another reader's**
  samples, and its own batch was then discarded.

## Fix (receiver-side, wire-compatible)

`SessionI::initSamples` now:

- processes **every batch of the current initialization round**, tracking the subscribers initialized during the
  call (batches from an earlier round are still ignored — one `initSamplesAsync` carries a whole round, built by
  `TopicI::attachElementsAck`);
- **routes batches by key**: a subscriber is only initialized and delivered by a batch whose samples match its key
  subscriptions; an empty key set or a null key (any-key and filtered readers) matches every key;
- still marks the remaining subscribers initialized on an **empty batch** (required to enable the live sample
  path);
- advances `lastId` **monotonically** (per-key batches carry interleaved sample ids, matching the
  `subscriberInitialized` behavior from #5809) and passes `checkKey = true` when queuing.

The wire protocol cannot address an initialization batch to a specific reader element; the residual corner (an
empty batch of one round marking a subscriber whose non-empty batch arrives in a later round — a loss that also
existed before this fix) needs the initialization stream to become addressable, tracked as a 3.9.0 protocol
follow-up.

## Testing

Three new late-join cases in `test/DataStorm/events`, all verified red→green (red = hang, the batch is dropped):

- a late-joining **multi-key** reader receives both keys' initialization samples;
- a late-joining **any-key** reader receives both keys' samples (pins the null-key wildcard routing);
- two late-joining **single-key** readers each receive exactly their own key's sample and not the other's
  (pins the per-reader routing across separate attach rounds).

The choreography uses a probe reader to guarantee the topics are attached before the late element is created, so
the element attach is writer-initiated and the samples deterministically arrive via the `initSamples` request (the
other handshake direction — `ElementDataAck.samples` → `subscriberInitialized` — was fixed by #5809). Verified
stable across repeated runs of both test variants; the full DataStorm suite passes. A changelog fragment is
included.

### DIFF
--- a/changelog.d/datastorm/5819.md
+++ b/changelog.d/datastorm/5819.md
@@ -1,0 +1,5 @@
+- Fixed a bug in DataStorm where the initialization samples for a late-joining reader could be lost or delivered to
+  the wrong reader. A writer covering several keys, or several readers of the same writer, produces one
+  initialization batch per key and per reader; these batches were returned to the subscriber without identifying the
+  target reader, so the receiver could apply only the first batch, ignore the keys, or deliver a batch twice. The
+  writer now returns each batch as an acknowledgment addressed to the exact reader element.
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -154,7 +154,7 @@ DataElementI::attach(
     SessionPrx prx,
     const ElementDataAck& data,
     const chrono::time_point<chrono::system_clock>& now,
-    DataSamplesSeq& samples)
+    ElementDataAckSeq& acks)
 {
     // Called with the topic and session from TopicI::attachElementsAck locked.
     shared_ptr<Filter> sampleFilter;
@@ -178,10 +178,11 @@ DataElementI::attach(
         name = os.str();
     }
 
-    // Attach a key or filter. If the attachment is successful, compute the queued samples to send to the peer.
-    // - If this data element is a data reader, the computed samples will be empty.
-    // - If this data element is a data writer, the computed samples will contain the samples to send to the peer
-    //   based on the peer reader configuration.
+    // Attach a key or filter. If the attachment succeeds and this element is a data writer, emit its queued samples
+    // as an acknowledgment addressed to the exact peer reader element (data.id), mirroring the acknowledgment the
+    // announce-first attachElements path builds. A data reader has no samples to send. This addressed return
+    // acknowledgment replaces the old unaddressed initSamples leg, so the peer routes the batch to the reader that
+    // attached instead of guessing from the writer element id.
     if ((id > 0 &&
          attachKey(topicId, data.id, key, sampleFilter, session, std::move(prx), facet, id, name, priority)) ||
         (id < 0 &&
@@ -189,7 +190,22 @@ DataElementI::attach(
     {
         auto q = data.lastIds.find(_id);
         int64_t lastId = q != data.lastIds.end() ? q->second : 0;
-        samples.push_back(getSamples(key, sampleFilter, data.config, lastId, now));
+        DataSamples batch = getSamples(key, sampleFilter, data.config, lastId, now);
+
+        // Only a data writer produces samples: KeyDataWriterI::getSamples always stamps a nonzero batch id (+/-_id),
+        // even for an empty history, whereas a data reader's base getSamples returns an id-0 batch. Send even an empty
+        // writer batch, so the peer marks the reader initialized and enables its live sample path. The acknowledgment
+        // carries the raw positive element id (_id); batch.id is used only as the writer discriminator here. lastIds
+        // is left empty: on this terminating leg the peer's element is already attached, so it never reads them.
+        if (batch.id != 0)
+        {
+            acks.push_back(ElementDataAck{
+                .id = _id,
+                .config = _config,
+                .lastIds = {},
+                .samples = std::move(batch.samples),
+                .peerId = data.id});
+        }
     }
 
     auto samplesI =
--- a/cpp/src/DataStorm/DataElementI.h
+++ b/cpp/src/DataStorm/DataElementI.h
@@ -192,8 +192,9 @@ namespace DataStormI
         /// @param data The acknowledgment data associated with the remote element, describing its configuration or
         /// state.
         /// @param now The timestamp indicating when the attachment was requested.
-        /// @param samples Output parameter filled with the data samples in the publisher's queue. This parameter is
-        /// always empty when the method is called on the subscriber side.
+        /// @param acks Output parameter filled with the publisher's queued samples as an acknowledgment addressed to
+        /// the peer reader element (`data.id`). Empty when the method is called on the subscriber side, since a reader
+        /// has no samples to send.
         /// @return A function that initializes the reader with the prepared samples:
         /// - For a publisher, this method always returns a `nullptr` function.
         /// - For a subscriber, this method returns a function that initializes the reader with samples provided by the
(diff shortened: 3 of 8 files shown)


pepone:
I did an adversarial source-level cross-check of the new `initializedNow` logic. I believe the current patch introduces a real duplicate-delivery corner.

## The duplicate-delivery corner

Two same-key readers can be included in one `attachElementsAck`: announcements are per key/filter rather than per reader element, and `TopicI::attachElements` attaches all local not-yet-attached elements matching that key. The writer then creates one initialization batch per attached reader and aggregates them into one `initSamples` call.

For two fresh readers `rA1` and `rA2` on the same key, that can produce two identical batches `[B1, B2]` for the same writer element. With the proposed loop:

1. `B1` matches both readers, delivers to both, and inserts both into `initializedNow`.
2. On `B2`, the "initialized by an earlier round" check does not skip either reader, because both are present in `initializedNow`.
3. The key matches again, so `element->initSamples(...)` runs again for both readers.

The `lastId` comparison only controls updating the cursor; it does not guard delivery. I also could not find downstream sample-ID deduplication: `DataReaderI::initSamples` applies key/discard checks and then queues the samples and fires callbacks normally.

So, in this exact corner, the pre-fix code was accidentally correct: the first batch initialized both readers and the second was dropped. The PR changes that to two deliveries per reader.

## Why per-subscriber ID-set dedup is not the right general fix

Initialization batches are tailored per reader using its filter, history depth, and `lastId`; additionally, the earliest partial update can be encoded as a full update. Consequently, the same sample ID can legitimately have different payload representations in two reader-specific batches. If the wrong batch reaches a reader first, ID dedup can suppress the correctly tailored representation.

## Proposed fix: return the samples in a second, addressed `attachElementsAck`

A binary-compatible fix appears to already exist in the current protocol. `ElementDataAck` already contains both:

- `peerId`, identifying the exact remote reader element.
- `samples`, documented as the writer-to-reader initialization samples.

Existing receiver code in `TopicI::attachElementsAck` selects the local element using `data.peerId`, and `DataElementI::attach` calls `subscriberInitialized(..., data.samples, ..., element)` for that exact element even when the duplicate attachment itself fails.

The implementation shape would be:

```cpp
auto batch = getSamples(key, sampleFilter, data.config, lastId, now);
if (batch.id != 0) // writers use +/-_id; readers return {}
{
    acks.push_back(ElementDataAck{
        .id = _id,                 // raw positive writer ID
        .config = _config,
        .lastIds = {},
        .samples = std::move(batch.samples),
        .peerId = data.id          // exact remote reader ID
    });
}
```

Then:

- Change the ACK-taking `DataElementI::attach` output from `DataSamplesSeq` to `ElementDataAckSeq`.
- Make `TopicI::attachElementsAck` return a mirrored `ElementSpecAckSeq`.
- Rebuild the outer key/filter envelope using the same rules as `TopicI::attachElements`. In particular, re-encode the local writer key/filter for key↔filter matches; merely swapping the incoming value is incorrect.
- In `SessionI::attachElementsAck`, send the result with the existing `attachElementsAckAsync` instead of `initSamplesAsync`.

This should terminate naturally: the reader had already attached the pair before creating its original ACK, so processing the returned ACK does not create another attachment or response. Nevertheless, the addressed samples are still consumed. An empty writer history must still produce an addressed ACK so that only the intended reader is marked initialized.

## Properties of this approach

- No `Contract.ice` or generated-code change.
- No public ABI or wire-format change.
- No sample-ID deduplication.
- Preserves reader-specific filters, history limits, cursors, and sample encodings.
- Handles empty batches and differing `lastId` values correctly.
- New publisher → old subscriber works, because old receivers already support sample-bearing `ElementDataAck`.
- The message count remains unchanged: one addressed `attachElementsAck` replaces the current aggregate `initSamples` return call.

The existing `initSamples` receiver path should remain as a legacy fallback for old publishers. That mixed-version direction is inherently ambiguous, because the old publisher did not transmit the target reader ID.

## Recommendation

I recommend addressing this before merging #5880, and adding tests for:

- Two coalesced fresh same-key readers.
- Same-key readers with different filters/history depths.
- Different nonzero `lastId` values.
- Key↔filter and any-key combinations.
- Empty initialization batches.
- New-publisher/old-subscriber interoperability.

pepone (commented):
Deep review result: I recommend holding the merge until the new regression tests are synchronized deterministically. The production DataStorm change looks sound for current-to-current peers and addresses both failure modes in #5819.

1. High: release-test race in `cpp/test/DataStorm/events/Writer.cpp:449`

The test does:

```cpp
writer.waitForReaders(2);
writer.waitForNoReaders();
```

`waitForReaders(2)` observes the current listener count; it does not latch that two readers were previously attached. On optimized CI builds, both readers can attach, consume the correct initialization samples, and detach before the writer reacquires the topic mutex. The writer then sees zero and waits forever for the count to return to two.

This matches all three observed release failures: Ubuntu 24.04, Windows 2022, and Windows 2025. The Ubuntu stack is blocked in `DataElementI::waitForListeners(2)` from Writer.cpp:449, while the reader trace shows that both readers received their correctly addressed samples and detached. The preceding case at Writer.cpp:428-429 has the same vulnerable pattern.

Suggested fix: use an explicit ready/done topic handshake, as the later cases already do, or keep the readers alive until the writer acknowledges observing them.

2. Medium: the same-key regression case is nondeterministic

The comment at `cpp/test/DataStorm/events/Reader.cpp:536` says the two readers are not guaranteed to coalesce into one initialization round. If they attach in separate rounds, the test passes without exercising the duplicate-delivery corner that motivated this redesign. Please make coalescing deterministic through synchronization or a test hook, or add a lower-level protocol test that proves both readers participate in the same attachment/ACK round.

3. Mixed-version compatibility caveat

The wire format is unchanged, but old 3.8.0-3.8.2 subscribers assume strictly increasing IDs between initialization batches. A new publisher sending addressed per-key batches with interleaved IDs, such as A=[1,3] and B=[2], can still trip the old debug assertion or move the old release subscriber cursor backward. This is not a current-to-current defect, but if rolling upgrades are supported I recommend a new-publisher/old-subscriber test and either a subscriber-first upgrade note or the prerequisite receiver fix backported.

Implementation audit: I traced key/key, key/filter, filter/key, and filter/filter attachments; empty histories; per-reader history/sample filters; cursor handling; ACK termination; and live-sample ordering. I found no production correctness issue for current peers. The complete C++ build and all 10 DataStorm groups passed locally, as did 35 focused stress iterations. The PR description should also be updated because it still describes the superseded receiver-side `initSamples` approach rather than the addressed return-ACK implementation.

— Codex (model: gpt-5.6-sol; reasoning: max)

pepone:
Superseded by #6067, which fixes the same bug (#5819) with the negotiated protocol-epoch approach agreed in the [icerpc-dev discussion](https://github.com/icerpc/icerpc-dev/discussions/5): a receiver-side fix that stays wire-compatible with Ice 3.8.0-3.8.2 by default, plus an opt-in protocol epoch for exact per-reader initialization between upgraded nodes. Closing in favor of #6067.

PR #5882, opened 2026-07-08
Make DataStorm reader-side partial update resolution robust to discards and errors

### LINKED ISSUE #5824: [Audit-Low] DataStorm: reader discard policy drops the partial-update init base before it is recorded

## Summary

In `DataReaderI::initSamples` the `SendTime`/`Priority` discard `continue` runs **before** the per-key base is recorded, so a discarded init sample leaves no base for that key:

https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/DataElementI.cpp#L766

(the `previousByKey[sample->key] = sample` record follows at `DataElementI.cpp:791`).

## Trigger / impact

A reader with `DiscardPolicy::SendTime` whose element-wide `_lastSendTime` was advanced by another writer's samples: a second writer's init base for a different key arrives with an older timestamp → discarded → not recorded. A later, fresh-timestamped `PartialUpdate` for that key passes the discard check but now has no base → resolves against a default/null base (same failure class as zeroc-ice/ice#5809 A1: crash for class-typed values, silent default-reset otherwise). Requires the non-default `DiscardPolicy::SendTime` plus a second writer; narrow, hence Low.

## Fix direction

Seed `_lastByKey` even for discarded init samples — discard should affect app-visible delivery, not the value used as the partial-update resolution base — or document the limitation.

### LINKED ISSUE #5833: [Audit-Low] DataStorm: a throwing reader-side updater/decoder drops the sample and leaves the per-key base stale

## Summary

The reader resolves a `PartialUpdate` inline inside the oneway `SubscriberSessionI::s` dispatch, in `DataReaderI::queue`, with no surrounding try/catch:

https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/SessionI.cpp#L1391

An exception from `Decoder<UpdateValue>::decode` (e.g. a writer/reader updater **type mismatch** for a tag — which the header cannot enforce across processes) or from a user updater unwinds out of `queue()` before the sample is delivered and before `_lastByKey[sample->key] = sample` runs.

## Trigger / impact

The sample is dropped and the per-key base is not updated. `Ice.Warn.Dispatch` defaults to `1`, so a dispatch warning **is** logged (not silent). A later non-partial sample decodes and unconditionally resets the base, so divergence is not permanent in general — persistent loss only for an unbroken stream of partial updates (a type-mismatched decoder, which then throws on every partial anyway). Low.

## Fix direction

Wrap the updater/decoder invocation in `queue`/`initSamples`; on failure, log and fall back to `noOpUpdater` semantics (or drop the base so the next full `Update` resyncs). Document that reader-side updaters must not throw.

### LINKED ISSUE #5834: [Audit-Low] DataStorm: Reader::onSamples init callback documented as "always called" but skipped when there is no history

## Summary

`Reader::onSamples` documents that "the @p init callback is always called after this method returns" (`cpp/include/DataStorm/DataStorm.h`), but the implementation skips it when the sample history is empty:

https://github.com/zeroc-ice/ice/blob/90dc6c7171184e513f40114b36d1aa2dbf130c49/cpp/src/DataStorm/DataElementI.cpp#L990

(`if (init && !_samples.empty())`). The sibling callbacks `onConnectedKeys`/`onConnectedElements` do invoke their init unconditionally, even with an empty collection, making `onSamples` the behavioral outlier.

## Impact

An application that gates startup on the init callback (as it validly can for `onConnectedKeys`/`onConnectedWriters`) waits forever on a freshly created reader with no history.

## Fix direction

Either always queue `init` (with an empty vector, matching the sibling callbacks) or correct the documentation.

pepone:
Fixes #5824, fixes #5833, fixes #5834.

Three related hardening changes in the reader sample paths (`DataReaderI::queue` and `DataReaderI::initSamples`),
plus a documentation correction:

- **#5824** — A sample discarded by the reader's discard policy (`SendTime` or `Priority`) was dropped *before*
  the per-key base partial updates resolve against was recorded, so a later partial update on the key had no base.
  The discard now only affects the application-visible delivery: when the key has no base yet, the discarded
  sample still establishes it (only-if-absent, so a discarded lower-priority or older sample never replaces an
  existing base; partial updates are not recorded since they cannot be resolved).

- **Reader-side null-base guard** (the follow-up agreed in the #5809 review) — a partial update whose key has no
  base (for example from a peer writer that doesn't bootstrap the bases) invoked the updater with a null previous
  value, resolving the update against a default-constructed value — typically a crash for applications using
  class-typed values. Such samples are now traced and dropped: degraded delivery instead of a crash.

- **#5833** — A user updater or decoder that throws unwound out of the dispatch: the whole initialization batch
  was discarded — a late-joining reader lost every initialization sample, including full values preceding the
  failed one — and the base was left stale. The failure now drops only the offending sample with a warning; the
  base is left unchanged so the next full sample resynchronizes the key. `Topic::setUpdater`'s documentation now
  states that a throwing updater causes the sample to be dropped.

- **#5834** — resolved as a documentation fix: the header claimed the `onSamples` init callback "is always
  called", but the implementation intentionally skips it when the reader has no history, and the callbacks test
  asserts that behavior (`test(false)` in the init callback of an empty reader). The documentation now matches the
  implementation; no behavior change.

## Testing

New case in `test/DataStorm/partial`, verified red→green: a late-joining reader whose updater throws on one of the
initialization samples receives the other samples, and the following partial update resolves against the last
successfully applied value (without the fix the throw discards the entire batch and the reader hangs). The
discarded-base seeding and the null-base guard are not deterministically drivable from the test harness (they need
timestamp/attach-order control); both were verified by an adversarial codex audit of the diff, which also caught —
and this PR fixes — the all-samples-discarded corner where the seeded bases were lost with the early return.

The full DataStorm suite passes. A changelog fragment is included.

### DIFF
--- a/changelog.d/datastorm/5824.md
+++ b/changelog.d/datastorm/5824.md
@@ -1,0 +1,5 @@
+- Improved the robustness of DataStorm partial updates on the reader side: a sample discarded by the reader's
+  discard policy still establishes the value a later partial update on its key resolves against, a partial update
+  without such a value is now dropped with a trace instead of being applied to a default-constructed value (which
+  typically crashed applications using class-typed values), and an updater or decoder that throws now drops only
+  that sample with a warning instead of discarding the rest of the initialization samples.
--- a/changelog.d/datastorm/5824.md
+++ b/changelog.d/datastorm/5824.md
@@ -1,5 +1,0 @@
-- Improved the robustness of DataStorm partial updates on the reader side: a sample discarded by the reader's
-  discard policy still establishes the value a later partial update on its key resolves against, a partial update
-  without such a value is now dropped with a trace instead of being applied to a default-constructed value (which
-  typically crashed applications using class-typed values), and an updater or decoder that throws now drops only
-  that sample with a warning instead of discarding the rest of the initialization samples.
--- a/changelog.d/datastorm/reader-partial-update-robustness.md
+++ b/changelog.d/datastorm/reader-partial-update-robustness.md
@@ -1,0 +1,6 @@
+- Fixed a bug in DataStorm where a sample discarded by the reader's discard policy left later partial updates on
+  its key without a base value to resolve against.
+- A DataStorm partial update without a base value is now dropped with a trace, instead of being applied to a
+  default-constructed value — which typically crashed applications using class-typed values.
+- A DataStorm reader-side updater or decoder that throws now drops only that sample and logs a warning; previously
+  the exception could discard an entire batch of initialization samples.
--- a/cpp/include/DataStorm/DataStorm.h
+++ b/cpp/include/DataStorm/DataStorm.h
@@ -188,8 +188,9 @@ namespace DataStorm
 
         /// Calls the given function to provide the initial set of unread samples and when new samples are queued.
         /// If a function is already set, it will be replaced.
-        /// The @p init callback is always called after this method returns to provide the initial set of unread
-        /// samples. The @p queue callback is called when a new sample is received.
+        /// The @p init callback is called after this method returns to provide the initial set of unread samples;
+        /// it is only called when the reader has unread samples. The @p queue callback is called when a new sample
+        /// is received.
         /// @param init The function to call with the initial set of unread samples.
         /// @param queue The function to call when a new sample is received.
         void onSamples(
@@ -378,6 +379,8 @@ namespace DataStorm
         /// Sets an updater function for the given update tag. The function is called when a partial update is
         /// received or sent to compute the new value. The function is provided the latest value and the partial
         /// update. It should return the new value.
+        /// An updater that throws when applying a received partial update causes the sample to be dropped: a
+        /// warning is logged and the following samples are delivered normally.
         /// @param tag The update tag.
         /// @param updater The updater function.
         template<typename UpdateValue>
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -768,27 +768,79 @@ DataReaderI::initSamples(
             (_discardPolicy == DataStorm::DiscardPolicy::Priority &&
              hasLowerPriorityThanConnected(priority, sample->key)))
         {
+            // The discard only affects the application-visible delivery: when the key has no base yet, still record
+            // the discarded sample as the base partial updates resolve against, otherwise a later partial update on
+            // the key would have no base.
+            if (sample->event != DataStorm::SampleEvent::PartialUpdate &&
+                previousByKey.find(sample->key) == previousByKey.end())
+            {
+                try
+                {
+                    if (!sample->hasValue())
+                    {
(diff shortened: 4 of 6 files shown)


externl (approved):
The #5809 follow-up plus the two hardening fixes all check out against the post-change queue()/initSamples():

- #5824 discard-then-seed: a discarded sample seeds the base only-if-absent (`find()`, not `operator[]`), partials excluded, `_lastSendTime` untouched — a discarded sample never replaces an existing base.
- Null-base guard: base-less partial is traced and dropped instead of driving the updater with null; `lastId` is already advanced upstream, so it's self-healing degraded delivery, not an infinite refetch.
- #5833: per-sample try/catch leaves the base unchanged (the wrapper mutates a clone), preceding full samples survive, and the all-samples-discarded corner now moves the seeded bases into `_lastByKey` before the empty return. The throw test is a genuine red→green.

One optional note: `DataWriterI::publish` still passes null to the updater when the key has no base — the symmetric writer-side of this same pattern (app-controlled and synchronous, so not remotely triggerable). Possible follow-up, out of scope here.

changelog.d/datastorm/5824.md
```diff
@@ -0,0 +1,5 @@
+- Improved the robustness of DataStorm partial updates on the reader side: a sample discarded by the reader's
+  discard policy still establishes the value a later partial update on its key resolves against, a partial update
+  without such a value is now dropped with a trace instead of being applied to a default-constructed value (which
+  typically crashed applications using class-typed values), and an updater or decoder that throws now drops only
+  that sample with a warning instead of discarding the rest of the initialization samples.
```
externl: Can we shorten this?

bernardnormier: Or split it in multiple bullets?

bernardnormier (approved):
When a PR fixes multiple issues, the changelog fragment should have a simple unique name - don't name it after one of the issues.

changelog.d/datastorm/5824.md
pepone: 🤖 **Claude:** Renamed to `reader-partial-update-robustness.md` and split into one bullet per fix. Done in f4710be.

pepone:
🤖 **Claude:** Changelog restructured in f4710be — descriptive slug + one bullet per fix.

On the writer-side note (@externl): declining as working-as-documented rather than filing a follow-up.
`DataWriterI::publish` passing null to the updater when the key has no base is app-controlled and synchronous, and
the api test pins it as documented behavior — `partialUpdate` after `remove()` must still publish an empty-value
partial (`test/DataStorm/api/Writer.cpp`, the `getLast().getUpdateTag() == "partialupdate"` block). A writer-side
guard was tried during the #5809 work and reverted precisely because that test asserts the publish succeeds.

Issue #5879, opened 2026-07-08
[Audit-Low, grouped] JS: 5 low-severity runtime findings from the API doc audit


pepone:
Five small runtime defects found while verifying the JS API docs against the implementation, grouped per audit convention. Each is independently fixable with a one-to-few-line change.

### 1. `Ice.EnumBase` is declared but never exported
`EnumBase.d.ts` declares `class EnumBase` publicly and every enum type (hand-written and slice2js-generated) extends it, but [`EnumBase.js:10`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/EnumBase.js#L10) lacks `export` — the module's only export is `defineEnum`, so the spread in `Ice/index.js` never adds the class and `Ice.EnumBase` is `undefined` at runtime (module exports verified with node). TypeScript-legal code like `x instanceof Ice.EnumBase` throws. Fix: `export class EnumBase`.

### 2. `NoEndpointException` message starts with a literal `$"`
[`LocalExceptions.js:364`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/LocalExceptions.js#L364) — `` `$"No suitable endpoint available for proxy '${arg}'.` `` — C# string-interpolation residue inside a JS template literal; user-visible messages begin with `$"`.

### 3. The `OutputStream` constructor's `format` argument is dead in every form
[`OutputStream.js:484-521`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/OutputStream.js#L484-L521) — `arg2` is only examined when `arg1` is an `EncodingVersion`, and even then the constructor tail unconditionally reassigns `this._format` (instance default or `CompactFormat`), clobbering the value set from `arg2`. So `new OutputStream(Encoding_1_1, FormatType.SlicedFormat)` still produces a compact-format stream, contradicting the documented `format` parameter ([`OutputStream.d.ts:16-22`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/OutputStream.d.ts#L16-L22)). Fix: only apply the default when no format was supplied.

[...]

Issue #5875, opened 2026-07-08
[Audit-Medium] JS: default servant is bypassed when the identity is registered with a different facet


pepone:
### Summary
JavaScript analog of #5865: the default-servant fallback only runs when the request's identity is entirely absent from the Active Servant Map. If the identity is registered with a *different* facet, the default servant is never consulted and the client receives `FacetNotExistException` — diverging from the documented lookup order and from the C++ implementation.

### Mechanism
`ServantManager.findServant` consults `_defaultServantMap` only when the identity has no entry in `_servantMapMap`; when the identity exists it takes the `else` branch and returns null on facet miss ([`ServantManager.js:155-172`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/ServantManager.js#L155-L172)). Dispatch then finds `hasServant(current.id)` true and throws `FacetNotExistException` ([`ServantManager.js:66-70`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/ServantManager.js#L66-L70)).

C++ falls through to the default-servant lookup when *either* the identity *or* the facet is missing ([`ServantManager.cpp:204-218`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Ice/ServantManager.cpp#L204-L218)), and the JS doc describes the C++ order ([`ObjectAdapter.d.ts:100-114`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/ObjectAdapter.d.ts#L100-L114)).

### Trigger
```js
adapter.addFacet(facetServant, id, "x");
adapter.addDefaultServant(defaultServant, id.category);
```
A request for identity `id` with the default facet `""` dispatches to `defaultServant` in C++ but throws `FacetNotExistException` in JS.

### Impact
Same as #5865, in the JS mapping: servers mixing per-facet ASM registrations with default servants for the same category fail requests that should reach the default servant.

### Suggested fix
Mirror the C++ logic in `findServant`: fall back to `_defaultServantMap` when the facet lookup fails, not only when the identity is absent (same fix shape as #5865).

### Notes
[...]

PR #5883, opened 2026-07-08
Fix incorrect API doc comments and .d.ts declarations in the JS Ice library

pepone:
Fixes incorrect TSDoc comments and hand-written `.d.ts` declarations in the JS Ice library, found during the JavaScript API doc review (follow-up to #5867/#5868). Declaration/doc changes only — no runtime behavior change (the only `.js` edit is a one-word comment fix in `Protocol.js`); runtime bugs found by the same review are tracked separately in #5875-#5879.

Declared APIs that don't exist at runtime (calling them throws `TypeError`):

- `ObjectPrx.ice_toString()` → replaced with the actually-implemented `toString()`; also declared the implemented `hashCode()` (needed to use typed proxies as `HashMap` keys).
- `Logger.getPrefix()` → removed (no runtime logger implements it; implementing it for cross-mapping parity would be a follow-up).
- `UserException.ice_staticId()` → removed (slice2js generates `_ice_id` for exceptions; `ice_staticId` statics exist only for Value/proxy classes).

Declaration-vs-runtime mismatches corrected:

- `ObjectPrx.checkedCast` returns a native promise (it's an `async function`), not an `AsyncResult` with `cancel()`/`isSent()` — now `globalThis.Promise<ObjectPrx | null>`. Same one-line fix in `slice2js` for the generated proxies' `checkedCast` declaration.
- `ice_identity`/`ice_facet` return a plain `ObjectPrx`, not `this`.
- Missing nullability: `ice_getCachedConnection`, `getImplicitContext`, `proxyToString(prx)`, `setCloseCallback(callback)`, `InitializationData.properties/logger/sliceLoader`, ObjectAdapter `find`/`findFacet`/`findByProxy`/`findServantLocator`/`findDefaultServant`, `InputStream.endValue`/`readProxy`/`readOptionalProxy`/`readValue` callback, `OutputStream.startValue`/`writeString`/`writeProxy`, `Value.ice_getSlicedData`, `OutgoingResponse.exceptionId`/`exceptionDetails` (nullish for Ok/UserException).
- `OutputStream.writeOptional` returns a load-bearing `boolean`, not `void`; `OutgoingResponse` constructor's `replyStatus` is optional (defaults to `Ok`); `CompositeSliceLoader` constructor's array is optional; `RequestFailedException` now declares its real `(replyStatus, id?, facet?, operation?)` constructor as `protected` (previously TS exposed the inherited `(replyStatus, message?)` arity, which crashes at runtime); `Promise.resolve` no longer re-declares a shadowing `<T>`; `Protocol` gets a `private constructor()` (the runtime value is a plain object — `new Ice.Protocol()` throws); removed the redundant no-arg `Communicator`/`initialize` overloads; added the implemented-but-undeclared `Properties.getUnusedProperties()`.
- Removed `OutputStream.prepareWrite()` and `OutputStream.buffer` from the declarations: both return the internal `Buffer` class, not the declared `Uint8Array`, and the `.d.ts` files deliberately omit internal members elsewhere (e.g. `InputStream`'s `buffer` getter). `finished()` is the public API. Happy to re-add with an opaque type instead if you prefer.

Incorrect doc claims corrected (many are the same stale pre-3.8 patterns fixed in #5867/#5868):

- `writeValue` 1.0-only deferral, `writePendingValues`, "Ends the previous encapsulation", `startEncapsulation` overload saying "Writes the end", nested-encapsulation inheritance, `startValue` sliced-format-only caveat, `OptionalFormat.FSize` "prior to unmarshaling", `readAndCheckSeqSize` per-element, `ice_getSlicedData` "preserved-slice base class", `InitializationException` scope, `ToStringMode.Compat` non-printable-only, `Connection.throwException` "closing or closed" + "by the peer".
- JS-specific: `ObjectAdapter.createProxy` claimed adapter-ID/replica-group indirect proxies (JS always creates a direct proxy from published endpoints); `findDefaultServant` doc was a copy-paste of `findServantLocator`; `ServantLocator.deactivate` is called on adapter destroy; `EndpointInfo.secure()` claimed WSS-only (also true for ssl); `Connection.type()` listed "udp"; `close()` now documents rejection semantics; `getInfo()` documents the throw-after-closure; WS `localAddress`/`localPort` sentinels; `EndpointInfo.timeout` no-effect note; dead "@throws MarshalException if the stream has not been initialized with a communicator" on `readProxy`/`readOptionalProxy` (communicator-less streams are unconstructible in JS); `ice_getConnection` summary said "Gets the connection ID"; `createObjectAdapter` never throws synchronously (the promise rejects) and only bidir/routed adapters are supported; `getPropertyAsListWithDefault` claimed "empty list" instead of the default; `InitializationData.clone` is a shallow copy; `stringVersion`/`intVersion` pre-release forms; `getEndpoints` always returns an empty array; `HashMap` never compares values; `stringToIdentity` doesn't validate identities; `stringToEncodingVersion` returns doc said "object identity".
- Coverage/markup: `@throws` tags added for `stringToProxy`/`propertyToProxy` (`ParseException`), `setProperty` (`InitializationException`/`PropertyException`), `add`/`addFacet` (`AlreadyRegisteredException`); rewritten pre-3.8 `Communicator` class summary; `@return`→`@returns` (8×); stray `</param>`/`</returns>` C# residue; assorted typos ("Reads and encapsulation", double periods/spaces, "Ice run time", `{@link Communicator#destroy()}` parens, constructor `@returns` tags).

Verified with a differential `tsc` check over all 46 hand-written `.d.ts` files against the pre-change baseline: the only new diagnostic is a `Cannot find name 'ReplyStatus'` on the new `RequestFailedException` constructor line, the same pre-existing class of noise as the untouched `DispatchException` members (`ReplyStatus` is a build-generated declaration, absent in an unbuilt tree).

### DIFF
--- a/js/packages/ice/src/Ice/Logger.d.ts
+++ b/js/packages/ice/src/Ice/Logger.d.ts
@@ -4,7 +4,7 @@ declare module "@zeroc/ice" {
     namespace Ice {
         /**
          * Represents Ice's abstraction for logging and tracing. Applications can provide their own logger by
-         * implementing this abstraction and setting a logger on the communicator.
+         * implementing this abstraction and supplying a logger in {@link InitializationData}.
          * @see {@link InitializationData}
          */
         interface Logger {
@@ -35,12 +35,6 @@ declare module "@zeroc/ice" {
              * @see #warning
              */
             error(message: string): void;
-
-            /**
-             * Returns this logger's prefix.
-             * @returns The prefix.
-             */
-            getPrefix(): string;
 
             /**
              * Returns a clone of the logger with a new prefix.
--- a/js/packages/ice/src/Ice/ObjectPrx.d.ts
+++ b/js/packages/ice/src/Ice/ObjectPrx.d.ts
@@ -70,7 +70,14 @@ declare module "@zeroc/ice" {
              *
              * @returns A stringified proxy.
              */
-            ice_toString(): string;
+            toString(): string;
+
+            /**
+             * Returns the hash code of this proxy.
+             *
+             * @returns the hash code of this proxy.
+             */
+            hashCode(): number;
 
             /**
              * Creates a proxy that is identical to this proxy, except for the identity.
@@ -78,7 +85,7 @@ declare module "@zeroc/ice" {
              * @param id - The identity for the new proxy.
              * @returns A proxy with the new identity.
              */
-            ice_identity(id: Identity): this;
+            ice_identity(id: Identity): ObjectPrx;
 
             /**
              * Gets the identity embedded in this proxy.
@@ -150,10 +157,10 @@ declare module "@zeroc/ice" {
             /**
              * Creates a proxy that is identical to this proxy, except for the facet.
              *
-             * @param facet The facet for the new proxy.
+             * @param facet - The facet for the new proxy.
              * @returns A proxy with the new facet.
              */
-            ice_facet(facet: string): this;
+            ice_facet(facet: string): ObjectPrx;
 
             /**
              * Gets the facet for this proxy.
@@ -311,7 +318,8 @@ declare module "@zeroc/ice" {
             ice_isFixed(): boolean;
 
             /**
-             * Gets the connection ID of this proxy.
+             * Gets the connection for this proxy. If the proxy does not yet have an established connection, it first
+             * attempts to create a connection.
              *
              * @returns An asynchronous result that resolves to the connection used by this proxy.
              * @remarks You can call this function to establish a connection or associate the proxy with an existing
@@ -326,7 +334,7 @@ declare module "@zeroc/ice" {
              * @returns The cached connection for this proxy, or `null` if the proxy does not have an established
              * connection.
              */
-            ice_getCachedConnection(): Connection;
+            ice_getCachedConnection(): Connection | null;
 
             /**
              * Creates a proxy that is identical to this proxy, except for connection caching.
@@ -364,7 +372,7 @@ declare module "@zeroc/ice" {
              *
              * @param prx - The source proxy.
              * @param facet - An optional facet name.
-             * @returns A proxy with the requested type and facet, or null if the source proxy is null.
+             * @returns A proxy with the requested type and facet.
              */
             static uncheckedCast(prx: ObjectPrx, facet?: string): ObjectPrx;
 
@@ -384,14 +392,14 @@ declare module "@zeroc/ice" {
              * @param prx - The source proxy.
              * @param facet - An optional facet name.
              * @param context - The request context.
-             * @returns An asynchronous result resolving to a proxy with the requested type and facet, or `null` if the
+             * @returns A promise that resolves to a proxy with the requested type and facet, or `null` if the
              *          target object does not support the requested type.
              */
             static checkedCast(
                 prx: ObjectPrx | null,
                 facet?: string,
                 context?: Map<string, string>,
-            ): AsyncResult<ObjectPrx | null>;
+            ): globalThis.Promise<ObjectPrx | null>;
         }
     }
 }
--- a/js/packages/ice/src/Ice/OutgoingResponse.d.ts
+++ b/js/packages/ice/src/Ice/OutgoingResponse.d.ts
@@ -19,8 +19,8 @@ declare module "@zeroc/ice" {
             constructor(
                 outputStream: OutputStream,
                 replyStatus?: ReplyStatus,
-                exceptionId?: string,
-                exceptionDetails?: string,
+                exceptionId?: string | null,
+                exceptionDetails?: string | null,
             );
 
             /**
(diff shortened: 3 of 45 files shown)


js/packages/ice/src/Ice/ObjectPrx.d.ts
```diff
@@ -70,15 +70,22 @@ declare module "@zeroc/ice" {
...
+            hashCode(): number;
 
             /**
              * Creates a proxy that is identical to this proxy, except for the identity.
              *
              * @param id - The identity for the new proxy.
              * @returns A proxy with the new identity.
              */
-            ice_identity(id: Identity): this;
+            ice_identity(id: Identity): ObjectPrx;
```
pepone, line 88: Is this `this` -> `ObjectPrx` correct, my understanding is the method returns a proxy of the existing type `this` not `ObjectPrx`, is this because we overwrite the method in derived proxy types?

js/packages/ice/src/Ice/OutgoingResponse.d.ts
```diff
@@ -34,17 +34,18 @@ declare module "@zeroc/ice" {
...
 
             /**
-             * The exception ID of the response. It's empty when `replyStatus` is {@link ReplyStatus.Ok} or
-             * {@link ReplyStatus.UserException}. Otherwise, this ID is the value returned by
+             * The exception ID of the response. It's null or undefined when `replyStatus` is {@link ReplyStatus.Ok}
+             * or {@link ReplyStatus.UserException}. Otherwise, this ID is the value returned by
              * {@link LocalException#ice_id}. For other exceptions, this ID is the value returned by `Error#name`.
              */
-            exceptionId: string;
+            exceptionId: string | null | undefined;
```
pepone: Is in really null or undefined? doesn't the code deterministically use one or the other for ReplyStatus.Ok?

js/packages/ice/src/Ice/Properties.d.ts
```diff
@@ -132,11 +133,21 @@ declare module "@zeroc/ice" {
...
+             */
+            getUnusedProperties(): string[];
+
             /**
              * Sets a property. To unset a property, set it to the empty string.
              *
              * @param key - The property key.
              * @param value - The property value.
+             * @throws {@link InitializationException} - Thrown when the key is empty or contains only whitespace.
+             * @throws {@link PropertyException} - Thrown when the key uses a reserved Ice prefix (`Ice`, `IceSSL`,
```
pepone, line 149: `IceSSL` is not correct here, is it? for example is not marked as a property for "js" language in `config/PropertyNames.xml`

pepone:
🤖 **Claude:** CI failure was the test TypeScript hitting the now-correctly-nullable declarations (`ice_getCachedConnection()` and `getImplicitContext()` return `| null`). Fixed the 16 affected test files with non-null assertions (plus a small loop rewrite in `binding/Client.ts` to narrow properly); `gulp build` including the test typecheck now passes locally.

js/packages/ice/src/Ice/Logger.d.ts
```diff
@@ -36,12 +36,6 @@ declare module "@zeroc/ice" {
              */
             error(message: string): void;
 
-            /**
-             * Returns this logger's prefix.
-             * @returns The prefix.
-             */
-            getPrefix(): string;
```
pepone, line 43: This was not implemented in JS logger

js/packages/ice/src/Ice/ObjectPrx.d.ts
bernardnormier: It's correct: both ice_identity and ice_facet return a base proxy. 

The logic is you typically change type when you change identity or facet. We do this in all languages.

js/packages/ice/src/Ice/OutgoingResponse.d.ts
bernardnormier: We should certainly pick one or the other.

bernardnormier: Today's runtime produces both, deterministically per path: for Ok, the constructor never assigns these fields, so the getters return `undefined`; for the exception paths, `createOutgoingResponseWithException` passes `null` explicitly (`exceptionId = null` for user exceptions, and `exceptionId != null ? exception.stack : null` for the details). So the declaration matches the current behavior. If we normalize the constructor to assign `null` unconditionally, this can tighten to `string | null`.

js/packages/ice/src/Ice/Properties.d.ts
bernardnormier: The shipped JS runtime does reserve `IceSSL`: the generated `PropertyNames.js` includes `IceSSLProps` in `validProps`, so the Properties constructor consumes `--IceSSL.*` arguments and `setProperty` recognizes the prefix. Since `IceSSLProps.properties` is empty in JS, any `IceSSL.*` property then throws `PropertyException` — but the prefix itself is reserved, which is what this sentence documents.

bernardnormier (approved):
Verified the declaration changes against the runtime (ice_identity/ice_facet return a base ObjectPrx, checkedCast is an async function, OutgoingResponse Ok/exception sentinels, IceSSL reserved prefix, no Logger.getPrefix/UserException.ice_staticId at runtime) — all match.

js/packages/ice/src/Ice/OutgoingResponse.d.ts
pepone: 🤖 **Claude:** Normalized in ed30f5683b: the constructor now assigns `exceptionId ?? null` / `exceptionDetails ?? null` unconditionally, and the declarations tightened to `string | null`. This also fixes the `!== null` guard in `LoggerMiddleware.logDispatchFailed`, which an `undefined` value used to slip past.

Issue #5890, opened 2026-07-08
[Audit-Low, grouped] Python: 4 low-severity IcePy findings from the API doc audit


pepone:
Four small defects found while verifying the Python API docs against the implementation, grouped per audit convention.

### 1. `IcePy.identityToString` requires two arguments; the stub and docstring declare `toStringMode` optional
[`Communicator.cpp:1619`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Communicator.cpp#L1619) parses `"O!O"` with no optional marker, so `IcePy.identityToString(ident)` raises `TypeError`, while [`Init.cpp:86-97`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Init.cpp#L86-L97) and [`IcePy-stubs/__init__.pyi:693`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/python/IcePy-stubs/__init__.pyi#L693) declare `toStringMode=None`. (The `Ice.Util` wrapper always passes both, so only direct `IcePy` callers hit it.) Fix: accept `|O` or fix the stub/docstring.

### 2. Admin-facet APIs silently degrade for native facets — undocumented
[...]

Issue #5886, opened 2026-07-08
[Audit-Medium] Python: ice_id() raises AttributeError for every local exception


pepone:
### Summary
`Ice.Exception.ice_id()` returns `self._ice_id`, but `_ice_id` is only a class-level *annotation* — nothing assigns it for the hand-written local exceptions. Calling `ice_id()` on any of the 40 exception classes in `LocalExceptions.py` raises `AttributeError`; the method's docstring example (`::Ice::CommunicatorDestroyedException`) is itself a failing case.

### Mechanism
[`Exception.py:12`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/python/Ice/Exception.py#L12) declares `_ice_id: str` (annotation only) and [`:24`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/python/Ice/Exception.py#L24) returns it. No class in [`LocalExceptions.py`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/python/Ice/LocalExceptions.py) assigns `_ice_id`; slice2py emits the assignment only for Slice-defined exceptions ([`PythonUtil.cpp:1408-1410`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/slice2py/PythonUtil.cpp#L1408-L1410)).

### Trigger
```python
try:
    communicator.stringToProxy("x")  # after destroy, or any local-exception path
except Ice.LocalException as ex:
    print(ex.ice_id())  # AttributeError: ... has no attribute '_ice_id'
```
Verified by execution.

### Impact
Any generic error-handling/logging code calling `ice_id()` on a local exception crashes with `AttributeError` instead of producing the type ID. No Python test calls `ice_id()` on a local exception, so it went unnoticed.

### Suggested fix
Assign `_ice_id` per class (or derive it in `LocalException`, e.g. from the class name), and add a test.

### Notes
Confidence: high (reproduced). Found in the 2026-07-08 Python API doc audit (finding A.1).

---
*Filed from an automated AI audit of the Ice core API (Claude Fable 5, 2026-07-08); confirmed through a multi-agent adversarial refute pass and independent Codex cross-verification over the actual source. Not yet human-triaged.*

PR #5893, opened 2026-07-08
Fix incorrect API docstrings, annotations, and stubs in the Python Ice library

pepone:
Fixes incorrect API docstrings, type annotations, IcePy type stubs, and `PyDoc_STR` strings in the Python Ice mapping, found during the Python API doc review (follow-up to #5867/#5868/#5883). Docstring/annotation/string changes only — the runtime bugs found by the same review are tracked separately in #5886-#5890 and deliberately not fixed here.

Incorrect doc claims corrected (many mirror the wording already landed for C++/C#/JS):

- `Connection.throwException` "not yet closing or closed"; `setAdapter` documents the raise on incoming connections; `setCloseCallback` documents `None`-to-clear; `close` documents that awaiting the future raises the closure exception.
- `WSConnectionInfo.headers`: request headers for incoming, **response** headers for outgoing connections; `SSLConnectionInfo.peerCertificate` is a single PEM-encoded peer certificate (or empty), not "the certificate chain".
- `ServantLocator.deactivate` is called on adapter *destroy*; `locate`'s contract is now `Object | tuple[Object, object | None] | None` — a missing servant is returned as bare `None`, not `(None, cookie)` (which trips the dispatch warning path). The four test locators were updated accordingly (behaviorally equivalent).
- `InitializationException` "failure occurs during initialization"; `AlreadyRegisteredException` kind list trimmed ("object"/"replica group" never reach Python); `closedByApplication` documented on both connection exceptions; `DispatchException` no longer mentions middleware (Python has none).
- Process-logger docs no longer claim it is "used by all communicators..." (logger properties and `InitializationData.logger` take precedence) — in `ProcessLogger.py` and the `Init.cpp` PyDocs.
- `Future.cancel`/`running` no longer describe `concurrent.futures` semantics that don't exist here (cancellation succeeds from the running state); `Future.completed` documented; `InvocationFuture.sent` documents re-raising the invocation's failure; `Future.result` Raises references `BaseException` instead of the ambiguous bare `Exception`.
- `Properties()` no-arg creates an empty property set (Ice.Config/ICE_CONFIG loading happens only with `args`/`defaults`); `ToStringMode.Compat` escapes only non-printable ASCII; `stringVersion`/`intVersion` document pre-release forms; `stringToIdentity`/`identityToString` document the `LocalException` for empty-name identities; `ice_getSlicedData` drops the stale "preserved-slice base class" wording; `SliceInfo.bytes` excludes the slice header (the size is consumed before capture — the same stale sentence in C++ `SlicedData.h` will be fixed in the C++ follow-up); `batchRequestInterceptor` no longer says the interceptor "must" enqueue; `shutdownCompleted` "equivalent to calling waitForShutdown"; `Blobject` isn't sync-only; `ice_isA` doc references its actual parameter name.

Annotation/stub/PyDoc accuracy (each verified against the IcePy runtime):

- Nullability: `ice_getContext -> dict[str, str]` (never `None`), `ice_identity`/`ice_facet -> ObjectPrx` (not `Self`), `checkedCastAsync(proxy: ObjectPrx | None)`, `stringToProxy`/`propertyToProxy`/`getLocator`/`ice_getRouter`/`ice_getLocator`/`ice_getCompress`/`ice_getConnection` gain `| None`, `addAdminFacet` loses a phantom `| None`, `findAdminFacet -> Object | NativePropertiesAdmin | None` and `removeAdminFacet -> Object | None` (native facets), `setLocator`/`setCloseCallback` accept `None`, `ImplicitContext.put`/`remove -> str` (empty string, never `None`).
- Stub fixes: `Communicator.__init__` was declared a `@classmethod`; `waitForHold -> bool` (the Ice wrapper loops on it); `Operation.__init__` now matches the real constructor (missing `metadata` param, three misnamed/shifted params, no phantom default); `ExecutorCall.__call__` and `Properties.__init__`/`__str__` declared; `DispatchCallback.response(result)` arity; `ObjectPrx.__init__` param renamed `endpoint` → `proxyString`; `getProcessLogger -> Ice.Logger | Logger`.
- `SlicedData`/`SliceInfo` now carry class-level annotations (previously `pass` bodies — every attribute access was a type error), including the previously undocumented `instances`, which the runtime requires when re-marshaling; `UnknownSlicedValue` documents runtime-only construction and the compact-ID empty `unknownTypeId`.
- Matching `PyDoc_STR` corrections in `Connection.cpp`, `ConnectionInfo.cpp`, `EndpointInfo.cpp`, `Communicator.cpp`, `ObjectAdapter.cpp`, `Proxy.cpp`, `Init.cpp` (wrong signatures like `waitForHold -> None`, `getEndpoints -> None`, `ice_getEndpointSelection -> bool`, `ice_identity(id: str)`), plus six copy-paste "Force inheritance" comments.

Deliberately excluded (tracked elsewhere): `identityToString`'s optional-argument mismatch, the `args`/`initData` mutual-exclusion wording, and the admin-facet native exposure (#5890); the `getPropertyAsList` mismatched-quote claims (pending the cross-mapping code-vs-doc decision from the C++ audit). Two justified typing-metadata additions: `Value.py` gains the `__all__` its siblings have, and `InvocationCanceledException` gains the `@final` every other leaf exception in the file carries.

Verified with `pyright` (CI configuration, generated sources present): 0 errors; `ruff format`/`ruff check`: clean; changed IcePy files are clang-format-19 clean.


pepone:
🤖 **Claude:** The `Raises: LocalException` entries added here document the current behavior: these paths throw C++ `std::invalid_argument`/`std::logic_error`, which IcePy's `convertException` catch-all turns into a plain base `Ice.LocalException`. Mapping them to native Python error types (`ValueError`/`RuntimeError`) is backward-incompatible and tracked for 3.9 in #5894.

python/modules/IcePy/ObjectAdapter.cpp
```diff
@@ -1532,15 +1532,15 @@ static PyMethodDef AdapterMethods[] = {
...
     {"getLocator",
      reinterpret_cast<PyCFunction>(adapterGetLocator),
      METH_NOARGS,
-     PyDoc_STR("getLocator() -> Ice.LocatorPrx")},
+     PyDoc_STR("getLocator() -> Ice.LocatorPrx | None")},
     {"getEndpoints",
      reinterpret_cast<PyCFunction>(adapterGetEndpoints),
      METH_NOARGS,
-     PyDoc_STR("getEndpoints() -> None")},
+     PyDoc_STR("getEndpoints() -> tuple")},
```
bernardnormier: Should be `tuple[Endpoint, ...]` like getPublishedEndpoints below.

python/python/Ice/Future.py
```diff
@@ -35,7 +35,8 @@ def cancel(self) -> bool:
         """
         Attempts to cancel this future.
 
-        If this future is already running or has completed, it cannot be cancelled.
+        If this future has already completed, it cannot be cancelled and this method returns ``False``;
+        otherwise the future transitions to the cancelled state.
```
bernardnormier: "transitions to the cancelled state" isn't quite right for a repeat `cancel()` — an already-cancelled future doesn't transition, it just returns `True` again.

```suggestion
        If this future has already completed, it cannot be cancelled and this method returns ``False``.
        Otherwise, this method cancels the future (if it wasn't cancelled already) and returns ``True``.
```

The `Returns` description below (outside the diff) could then read: ``` ``True`` if this future is cancelled, ``False`` if it has completed. ``` — "is cancelled" covers the repeat-cancel case that "was cancelled" doesn't.

bernardnormier (approved):
Verified the behavioral claims against the runtime (locate contract, ice_identity/ice_facet returning the base proxy type, ImplicitContext returns, waitForHold, SliceInfo.bytes, Properties(), Operation stub) — all check out. The unknownTypeId wording is accurate for today's behavior; the underlying Python/Ruby divergence from the other mappings is now #5916.

python/modules/IcePy/ObjectAdapter.cpp
pepone: 🤖 **Claude:** Done in bf181f504c — `getEndpoints` now advertises `tuple[Endpoint, ...]`, matching `getPublishedEndpoints` and the stub.

python/python/Ice/Future.py
pepone: 🤖 **Claude:** Applied verbatim in bf181f504c, including the Returns line ("``True`` if this future is cancelled, ``False`` if it has completed").

Issue #5840, opened 2026-07-07
[Audit-Medium] IceGrid: loadServer racing server destruction strands the command — orphaned reply + node-wide loop under NodeI mutex


pepone:
> **AI-generated finding** — surfaced by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), in a multi-agent correctness audit of IceGrid (`cpp/src/IceGrid` + `cpp/test/IceGrid`): six subsystem reviewers followed by three independent adversarial "refute-it" verifiers; this finding was additionally corroborated by an independent Codex (gpt-5.5) read. Code is pinned to `origin/main` at [`a53cc349e0`](https://github.com/zeroc-ice/ice/commit/a53cc349e0379aa4f3a4c67160e4452982b6fd80). **Not yet human-triaged** — please verify before acting.

## Summary

A `loadServer` for a server that is currently being **destroyed** (`ServerI` state `Destroying`) is admitted, creates a fresh `_load` command whose AMD reply is never sent, and leaves the `Destroyed` servant registered in the object adapter. A subsequent registry retry then spins forever inside `NodeI::loadServer` while holding `NodeI::_mutex`, hanging all node dispatch.

`checkDestroyed()` throws only for state `Destroyed`, not `Destroying`:
https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/ServerI.cpp#L1108-L1109

`destroy()` holds `_mutex` only to snapshot `_adapters`, then runs `removeServer` + `removeRecursive` + adapter destruction **unlocked** — a wide window during which a `loadServer` can be dispatched.

The `Destroyed` case of `setStateNoSync` clears only `_destroy`, so the cleanup guard `if (_state == Destroyed && !_load)` is false when a `_load` was created during `Destroying`, and the servant is never removed / `_desc` never cleared:
https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/ServerI.cpp#L2695-L2714

`NodeI::loadServer` holds `NodeI::_mutex` across its whole find/create+load retry loop:
https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/NodeI.cpp#L811-L864

## Trigger

[...]

PR #5899, opened 2026-07-08
Don't clear ReplicaSessionManager::_thread in destroy()

### LINKED ISSUE #5842: [Audit-Medium] IceGrid: ReplicaSessionManager reads _thread shared_ptr without its mutex — data race with registry shutdown
> **AI-generated finding** — surfaced by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), in a multi-agent correctness audit of IceGrid (`cpp/src/IceGrid` + `cpp/test/IceGrid`): six subsystem reviewers followed by three independent adversarial "refute-it" verifiers. Code is pinned to `origin/main` at [`a53cc349e0`](https://github.com/zeroc-ice/ice/commit/a53cc349e0379aa4f3a4c67160e4452982b6fd80). **Not yet human-triaged** — please verify before acting.

## Summary

`ReplicaSessionManager::getSession()` reads the non-atomic `shared_ptr _thread` member **without holding `_mutex`**, while `ReplicaSessionManager::destroy()` writes `_thread = nullptr` **under `_mutex`**. Because `RegistryI::stop()` destroys the session before shutting down the communicator, server adapters are still dispatching when `_thread` is nulled → data race (UB), and the `_thread ? _thread->…` double-read can also null-deref.

Unlocked read (header inline):
https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/ReplicaSessionManager.h#L59-L62

Locked write in `destroy()`:
https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/ReplicaSessionManager.cpp#L360

`RegistryI::stop()` calls `_session->destroy()` before `_communicator->shutdown()`:
https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/RegistryI.cpp#L800

## Trigger

On a **slave**, a `LocatorRegistry::setAdapterDirectProxy` request (a server/adapter registering its endpoints) runs `_session.getSession()` on a dispatch thread:
https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/LocatorRegistryI.cpp#L407

concurrently with registry shutdown, which nulls `_thread` while adapters are still active.

## Impact

Unsynchronised read of a `shared_ptr` that another thread is writing is a data race (torn read / refcount corruption — undefined behavior). The ternary's double read of `_thread` can pass the null test and then dereference null → registry crash. `NodeSessionManager::destroy()` deliberately does **not** null its `_thread`; the asymmetry is the defect. (`getNodes` / `registerAllWellKnownObjects` read `_thread` only from single-threaded startup, so the dispatch-time reader above is the racing one.)

## Fix direction

Read `_thread` under `_mutex` in the accessors (snapshot to a local before use), or do not null it in `destroy()` (matching `NodeSessionManager`).

bernardnormier:
`ReplicaSessionManager::destroy()` cleared `_thread` under `_mutex`, but `getSession()` reads `_thread` without the mutex on dispatch threads. On a slave registry, `RegistryI::stop()` calls `_session->destroy()` *before* `_communicator->shutdown()`, so server adapters are still dispatching when `_thread` is nulled. A `LocatorRegistry::setAdapterDirectProxy` request (a server registering its adapter endpoints) runs `getSession()` on a dispatch thread concurrently with `destroy()`:

- unsynchronized read of a `shared_ptr` that another thread is writing is a data race (undefined behavior), and
- the `_thread ? _thread->getSession() : nullopt` ternary can pass the null test and then dereference a `_thread` that `destroy()` just nulled → crash.

### Fix

Stop clearing `_thread` in `destroy()`, matching `NodeSessionManager`, which deliberately never clears its thread. `_thread` is assigned exactly once, in `create()` under the lock, and is never modified afterward, so the lock-free reads in `getSession()` (and the startup-only `getNodes`/`registerAllWellKnownObjects`) are race-free. `destroy()` still terminates and joins the thread; it just doesn't null the member.

This keeps no cycle alive: the manager is owned by `RegistryI` via `unique_ptr` (nothing holds a `shared_ptr` to it), and the `Thread` refers back to the manager only through a raw reference, so retaining `_thread` simply keeps the already-joined `Thread` object alive until the manager is destroyed — exactly as `NodeSessionManager` already does.

While here, since `_thread` is now write-once:

- `destroy()` and `create(const InternalRegistryPrx&)` use `_thread` directly instead of capturing it into a local (the capture only existed because `destroy()` used to null it);
- `create(const InternalRegistryPrx&)` guards on `_communicator` (the actual destroyed indicator, cleared by `destroy()` under the lock) rather than `_thread`. Guarding on `_thread` would no longer detect a manager destroyed *after* initialization now that `_thread` stays set.

Fixes #5842


pepone (approved):
Approving — both fixes are correct, and I traced them against the live paths.

**#5840 (ServerI):** the `Destroying` window is genuinely reachable — `load()` passes `checkDestroyed()` (which throws only at `Destroyed`) while the internal `destroy()` runs without the lock, so a `_load` with a registered callback can survive into `Destroyed`. `LoadCommand::failed()` clears `_loadCB`, so the added `failed()` replies exactly once (no double-reply), and nulling `_load` lets the `_state == Destroyed && !_load` servant-removal guard run. Mirrors the `Destroying` case exactly. 👍

**#5842 (ReplicaSessionManager):** write-once `_thread` removes both the `shared_ptr` data race and the null-test-then-deref. After the wait, the new `!_communicator` guard provably implies `_thread` is set, so the direct `_thread->…` calls can't deref null. No ownership cycle or leak (manager owned by `RegistryI` via `unique_ptr`, `Thread` holds a raw back-ref, `destroy()` still terminates+joins) — consistent with `NodeSessionManager`.

One non-blocking note: since `_thread` is no longer nulled and `terminate()` doesn't clear the thread's cached `_session`, `getSession()` after `destroy()` now returns the stale (destroyed) session proxy instead of `nullopt`. `LocatorRegistryI::setAdapterDirectProxy` therefore makes one doomed call on it before the `LocalException` → `AdapterNotFoundException` — same observable result as before, just in a narrow shutdown window, and matching how `NodeSessionManager` already behaves. Fine as-is; could tighten later by having `SessionKeepAliveThread::getSession()` return `nullopt` once the thread is `Destroyed`.

bernardnormier:
I included the fix for #5840 in this PR by accident. It's now only #5842.

bernardnormier:
@pepone Thanks for the detailed review. I folded in your suggestion: `SessionKeepAliveThread::getSession()` now returns `nullopt` once the thread is `Destroyed` (e72798d), so after `destroy()` both `ReplicaSessionManager` and `NodeSessionManager` report no session instead of the stale proxy, and `LocatorRegistryI::setAdapterDirectProxy` skips the doomed call on the destroyed session.

PR #5900, opened 2026-07-09
Fix icestormdb to use the IceStorm service's subscribers key comparator

### LINKED ISSUE #5847: [Audit-High] icestormdb --import rebuilds the subscribers LMDB database without the service's key comparator — restored DB silently misroutes lookups
### Summary
`icestormdb` opens the `subscribers` database with LMDB's **default** (byte-wise) key ordering, but the IceStorm service opens the same database with a **custom** comparator that decodes and compares keys structurally. LMDB persists the B-tree in comparator order but not the comparator itself, so a database produced by `icestormdb --import` is laid out in an order the service cannot correctly search — keyed lookups miss whenever identity name lengths differ.

### Mechanism
- `icestormdb` opens `subscribers` with **no** comparator, both on import ([`IceStormDB.cpp:232-234`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/IceStormDB.cpp#L232-L234)) and export ([`IceStormDB.cpp:288-290`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/IceStormDB.cpp#L288-L290)). The `Dbi` comparator parameter defaults to `nullptr`, so `mdb_set_compare` is never called.
- The service opens it **with** `compareSubscriberRecordKey` ([`Instance.cpp:308`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Instance.cpp#L308)), which decodes each key and compares the `SubscriberRecordKey` struct ([`Util.cpp:62-80`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Util.cpp#L62-L80)) — i.e. lexicographic decoded-string order (via `Ice::Tuple`), not raw `memcmp`.
- Ice marshals strings size-prefixed (a length byte precedes the bytes), so decoded order ≠ memcmp order whenever identity name lengths differ (e.g. `topic.z` vs `topic.ab`).

The service does **keyed** operations on this map — `del` in `removeSubscribers` ([`TopicI.cpp:1258-1260`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TopicI.cpp#L1258-L1260)), `MDB_SET` cursor `find` of the placeholder in `destroyInternal`, and `find`/`put` in `create` ([`TopicManagerI.cpp:515-521`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TopicManagerI.cpp#L515-L521)) — so a wrong sort order produces genuine lookup misses, not just cosmetic reordering. The `llu` map is string-keyed with the default comparator on both sides, so it is unaffected.

### Trigger
Documented backup/restore (`icestormdb --export` then `--import`), then run IceStorm against the imported directory with topic/subscriber identities of heterogeneous name lengths. Startup looks fine (topic reload is a pure cursor scan), but later: `unsubscribe` → `del` misses → ghost subscriptions resurrect on every restart; `destroy` → placeholder `find` misses → records leak / topic resurrects; `create` → misplaced duplicate keys can trip the placeholder assert.

### Empirical reproduction
A standalone LMDB program that writes 40 Ice-style `[len][bytes]` keys with the default comparator, then reopens with a decoded-order comparator, returns `MDB_NOTFOUND` for **20 of 40** keys. The existing test (`cpp/test/IceStorm/icestormdb/test.py`) cannot catch this: equal-length topic names, zero subscribers, and it only exercises the startup cursor scan + byte-identical re-export.

### Suggested fix
Pass `IceStormInternal::compareSubscriberRecordKey` when opening `subscribers` in `IceStormDB.cpp` (both import and export), matching [`Instance.cpp:308`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Instance.cpp#L308).

---
*Filed from an automated AI correctness audit of `cpp/src/IceStorm` (Claude Fable 5, 2026-07-07), independently cross-verified by an adversarial Codex (gpt-5.5) read-only pass — both CONFIRMED this finding against source. Not yet human-triaged.*

### LINKED ISSUE #5857: [Audit-Low] icestormdb --export never checks the output stream — silent truncated backup on write error
### Summary
`icestormdb --export` checks that the output file opened, but never checks the stream after writing, and does not enable stream exceptions — so a write error (ENOSPC, quota, I/O) sets badbit without throwing and the tool returns 0 with a truncated backup.

### Mechanism
[`IceStormDB.cpp:320-327`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/IceStormDB.cpp#L320-L327): open failure is checked, but `fs.write`/`fs.close` results are not, and there is no `fs.exceptions(...)` anywhere in the file.

### Trigger
Disk-full or I/O error while exporting a backup.

### Impact
The backup tool reports success while writing a truncated file; the loss surfaces only when the import of that file fails later. (icegriddb shares this defect.)

### Suggested fix
Enable `ofstream::exceptions(badbit|failbit)` or check the stream state after write/close and exit non-zero on failure.

---

pepone:
Fixes #5847
Fixes #5857

`icestormdb` opened the `subscribers` LMDB database with LMDB's default byte ordering, while the IceStorm service opens it with the custom `compareSubscriberRecordKey` comparator. A database produced by `icestormdb --import` was therefore laid out in a different order than the service expects, so after restoring a backup the service could silently fail to find subscriber and topic records whose identities differ in length (keyed `del`/`find` miss — ghost subscriptions resurrect on restart, topic records leak).

The comparator previously lived in a translation unit that pulls in the whole service, so it is extracted into `SubscriberRecordKeyCompare.cpp`, compiled into **both** the service and `icestormdb` (a single source of truth), with its communicator installed on both before the database is opened. `icestormdb` now passes the comparator to its import and export opens.

Also makes `icestormdb --export` return a non-zero status on an output-stream write error instead of silently producing a truncated backup (#5857).

### Testing
Extended `test/IceStorm/icestormdb` to create topics with different-length names and exercise a keyed lookup after export → import → restart: it fails without the comparator fix (topics resurrect after a destroy) and passes with it. `single` and `federation` remain green.

<sub>🤖 Produced by an automated correctness audit (Claude Fable 5) and cross-verified by Codex; reviewed but flagged for human triage.</sub>

### DIFF
--- a/changelog.d/service-icestorm/icestormdb-subscribers-comparator.md
+++ b/changelog.d/service-icestorm/icestormdb-subscribers-comparator.md
@@ -1,0 +1,4 @@
+- Fixed `icestormdb` to open the subscribers database with the same key comparator as the IceStorm service.
+  Previously a database produced by `icestormdb --import` was ordered differently than the service expected, so
+  after restoring a backup the service could fail to find subscriber and topic records whose identities have
+  differing lengths.
--- a/cpp/src/IceStorm/IceStormDB.cpp
+++ b/cpp/src/IceStorm/IceStormDB.cpp
@@ -7,6 +7,7 @@
 #include "DBTypes.h"
 #include "Ice/Ice.h"
 #include "Ice/StringUtil.h"
+#include "Util.h"
 
 #include <fstream>
 #include <iterator>
@@ -145,6 +146,7 @@ run(const shared_ptr<Ice::Communicator>& communicator, const Ice::StringSeq& arg
         IceStorm::AllData data;
 
         IceDB::IceContext dbContext = {communicator};
+        IceStormInternal::setKeyComparatorCommunicator(communicator);
 
         if (import)
         {
@@ -231,7 +233,12 @@ run(const shared_ptr<Ice::Communicator>& communicator, const Ice::StringSeq& arg
 
                 IceDB::
                     Dbi<IceStorm::SubscriberRecordKey, IceStorm::SubscriberRecord, IceDB::IceContext, Ice::OutputStream>
-                        subscriberMap(txn, "subscribers", dbContext, MDB_CREATE);
+                        subscriberMap(
+                            txn,
+                            "subscribers",
+                            dbContext,
+                            MDB_CREATE,
+                            IceStormInternal::compareSubscriberRecordKey);
 
                 for (const auto& subscriber : data.subscribers)
                 {
@@ -287,7 +294,7 @@ run(const shared_ptr<Ice::Communicator>& communicator, const Ice::StringSeq& arg
 
                 IceDB::
                     Dbi<IceStorm::SubscriberRecordKey, IceStorm::SubscriberRecord, IceDB::IceContext, Ice::OutputStream>
-                        subscriberMap(txn, "subscribers", dbContext, 0);
+                        subscriberMap(txn, "subscribers", dbContext, 0, IceStormInternal::compareSubscriberRecordKey);
 
                 IceStorm::SubscriberRecordKey key;
                 IceStorm::SubscriberRecord record;
@@ -325,6 +332,11 @@ run(const shared_ptr<Ice::Communicator>& communicator, const Ice::StringSeq& arg
             }
             fs.write(reinterpret_cast<const char*>(stream.b.begin()), static_cast<streamsize>(stream.b.size()));
             fs.close();
+            if (fs.fail())
+            {
+                consoleErr << args[0] << ": could not write output file: " << IceInternal::errorToString(errno) << endl;
+                return 1;
+            }
         }
     }
     catch (const Ice::Exception& ex)
--- a/cpp/src/IceStorm/Instance.cpp
+++ b/cpp/src/IceStorm/Instance.cpp
@@ -301,6 +301,7 @@ PersistentInstance::PersistentInstance(
     try
     {
         dbContext.communicator = std::move(communicator);
+        IceStormInternal::setKeyComparatorCommunicator(dbContext.communicator);
 
         IceDB::ReadWriteTxn txn(_dbEnv);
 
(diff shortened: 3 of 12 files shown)


bernardnormier (commented):
The fix and the test look good.

Suggestion for a follow-up PR: remove the `dbContext` global from Util.cpp altogether. `Dbi` takes the marshaling context by value and keeps its own copy, so the global is only read when the maps are constructed — after this PR its sole ongoing consumer is the comparator, which already has its own `keyContext` in SubscriberRecordKeyCompare.cpp. `PersistentInstance` can pass a local context to the two map constructors, like `icestormdb` already does:

```cpp
// Instance.cpp
IceDB::IceContext dbContext{communicator};
IceStormInternal::setKeyComparatorCommunicator(communicator);

IceDB::ReadWriteTxn txn(_dbEnv);

_lluMap = LLUMap(txn, "llu", dbContext, MDB_CREATE);
_subscriberMap = SubscriberMap(txn, "subscribers", dbContext, MDB_CREATE, compareSubscriberRecordKey);
```

`destroy()` then only needs `setKeyComparatorCommunicator(nullptr)`, and the service is down to a single global (the one LMDB's function-pointer comparator forces on us).

pepone:
🤖 **Claude:** Tracked the dbContext-global removal as #5950, targeting 3.9.0.

bernardnormier (approved):

Issue #5858, opened 2026-07-07
[Audit-Low, grouped] IceStorm: 14 low-severity correctness/robustness findings


pepone:
Grouped low-severity findings from the 2026-07-07 IceStorm correctness audit. Each is small and independent; filed together to keep the tracker manageable. All verified against source via a multi-agent adversarial refute pass.

### B2 — destroy/re-create servant-vs-reaper ordering race
`destroyInternal` adds the topic name to the reaper ([`TopicI.cpp:1229`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TopicI.cpp#L1229)) before removing the topic servant (`:1238`); `destroy` and `create` hold disjoint locks. A concurrent `create(sameName)` in that window commits DB/replica state then throws `AlreadyRegisteredException` on `topicAdapter()->add`, leaving the new topic servant-less. **Fix:** remove the servant before the reaper-add. Narrow window.

### B7 — link() derefs a null getLinkProxy() (assert-only)
The peer's `getLinkProxy()` result is checked only with `assert` ([`TopicI.cpp:610-612`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TopicI.cpp#L610-L612); transient has no check, [`TransientTopicI.cpp:221-222`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TransientTopicI.cpp#L221-L222)); a null flows into `SubscriberLink`'s ctor deref ([`Subscriber.cpp:315-318`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Subscriber.cpp#L315-L318)) — release-build UB. Requires a non-conforming peer. **Fix:** `checkNotNull` the result in both impls.

### A3 — per-subscriber publisher path drops the reap signal
[...]

PR #5903, opened 2026-07-09
Fix several low-severity IceStorm correctness issues

pepone:
Part of #5858 (fixes 5 of the grouped items; the remainder need an API change, generated-code regeneration, or a feature and are proposed for 3.9.0).

- **Node-id stagger underflow** — `NodeI::start` computed the initial election-check delay with unsigned arithmetic that underflowed when a node id exceeded the node count (ids need not be contiguous), throwing from the timer and preventing the replica from starting. The delay is now clamped at zero.
- **`accept` null observer** — `NodeI::accept` stored the peer's observer proxy without a null check (unlike `ready`), risking undefined behavior in `Observers::init`. A null observer is now rejected.
- **destroy/re-create race** — `TopicImpl::destroyInternal` published the topic name to the reaper before removing its servant, so a concurrent `createTopic` for the same name could fail with `AlreadyRegisteredException` and leave the new topic without a servant. The servant is now removed first.
- **`Send.QueueSizeMax=0`** — was unvalidated and matched an empty queue, causing a pop on an empty queue (`DropEvents`) or removing every subscriber on the first event. `0` is now treated as unbounded.
- **`subscribers` admin command** — did not accept `instance/topic` qualified names, aborted on the first missing topic, and had a malformed message. It now uses `findTopic`, continues past a failing topic, and the message is fixed.

### Testing
`single`, `federation2`, and `icestormdb` remain green.

<sub>🤖 Produced by an automated correctness audit (Claude Fable 5) and cross-verified by Codex; reviewed but flagged for human triage.</sub>

### DIFF
--- a/cpp/src/IceStorm/Parser.cpp
+++ b/cpp/src/IceStorm/Parser.cpp
@@ -307,24 +307,24 @@ Parser::subscribers(const list<string>& args)
 {
     if (args.empty())
     {
-        error("subscribers' requires at least one argument (type `help' for more info) ");
+        error("`subscribers' requires at least one argument (type `help' for more info)");
         return;
     }
-    try
+    for (const auto& arg : args)
     {
-        for (const auto& arg : args)
+        try
         {
-            auto topic = _defaultManager->retrieve(arg);
+            auto topic = findTopic(arg);
             consoleOut << arg << ": subscribers:" << endl;
             for (const auto& subscriber : topic->getSubscribers())
             {
                 consoleOut << "\t" << _communicator->identityToString(subscriber) << endl;
             }
         }
-    }
-    catch (const std::exception&)
-    {
-        exception(current_exception());
+        catch (const std::exception&)
+        {
+            exception(current_exception());
+        }
     }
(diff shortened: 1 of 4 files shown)


cpp/src/IceStorm/Parser.cpp
```diff
@@ -307,24 +307,24 @@ Parser::subscribers(const list<string>& args)
 {
     if (args.empty())
     {
-        error("subscribers' requires at least one argument (type `help' for more info) ");
+        error("`subscribers' requires at least one argument (type `help' for more info)");
```
pepone:
```suggestion
        error("'subscribers' requires at least one argument (type 'help' for more info)");
```

bernardnormier (approved):
Verified each fix against the surrounding code: the null-observer check mirrors ready(), the subscribers command now goes through findTopic like the other commands, and the destroyInternal reorder closes the create()/reap()/installTopic window. Note the CI run was cancelled (~10m in, no code failures) and needs a re-run.

Issue #5844, opened 2026-07-07
[Audit-Low] IceGrid: batch of low-severity correctness findings (12 items + cosmetics)


pepone:
> **AI-generated findings** — surfaced by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), in a multi-agent correctness audit of IceGrid (`cpp/src/IceGrid` + `cpp/test/IceGrid`): six subsystem reviewers followed by three independent adversarial "refute-it" verifiers. Code is pinned to `origin/main` at [`a53cc349e0`](https://github.com/zeroc-ice/ice/commit/a53cc349e0379aa4f3a4c67160e4452982b6fd80). **Not yet human-triaged** — please verify before acting.

A batch of low-severity / cosmetic correctness findings grouped into one issue (higher-severity findings from the same audit are filed separately). Each is independent; feel free to split or cherry-pick.

## L1 — `#5645`'s `assert(request)` documents a false invariant
[`Allocatable.cpp:306-311`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/Allocatable.cpp#L306-L311). The release-drain `catch(const Ice::UserException&)` comments "canTryAllocate() never throws a UserException" and asserts `request` non-null. But a destroyed `AllocatableObjectEntry` throws `ObjectNotRegisteredException` from [`checkAllocatable()`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/AllocatableObjectCache.cpp#L385-L390) — a `UserException` **not** derived from `AllocationException`, so it is not caught by `tryAllocate` or `TypeEntry::canTryAllocate` and can reach line 306 with `request == nullptr` → `assert` (debug) / `request->cancel()` null-deref (NDEBUG). Reaching the throwing state is convoluted (a `(destroyed entry, nullptr)` hook still queued while its multi-object type holds a pending `allocateByType` request) and I could not construct a definitive trigger — filing as a robustness fix. **Fix:** make the guard real — `if (request) request->cancel(current_exception());` — matching the sibling at [line 299-301](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/Allocatable.cpp#L299-L301).

## L2 — `activate()` re-fetches and derefs the master node session unchecked
[...]

PR #5906, opened 2026-07-09
Fix _releasing leak and null request cancel in IceGrid Allocatable::release

bernardnormier:
This PR fixes two defects in `Allocatable::release` (items L1 and L6 of #5844).

### Null `request` dereference in the release-drain loop (L1)

The drain loop's `catch (const Ice::UserException&)` asserted that `request` is non-null, based on the assumption that `canTryAllocate()` never throws a `UserException`. That assumption is false:

`AllocatableObjectEntry::canTryAllocate()` → `AllocatableObjectCache::canTryAllocate()` → `TypeEntry::canTryAllocate()` → `Allocatable::tryAllocate()` → `allocate()` → `checkAllocatable()`, where a destroyed `AllocatableObjectEntry` throws `ObjectNotRegisteredException`. This exception is not derived from `AllocationException`, so it passes through `tryAllocate()`'s catch, and `TypeEntry::canTryAllocate()` only catches `SessionDestroyedException`. It can therefore reach the drain loop's catch with a null `request` (a queued can-try-allocate hook), firing the assert in debug builds and dereferencing a null pointer in release builds.

The fix replaces the assert with a null check, matching the sibling catch a few lines above.

### `_releasing` never cleared when an exception escapes `release()` (L6)

`release()` sets `_releasing = true` before draining queued allocation attempts, and cleared it only on the two normal return paths. Any exception escaping the drain — or `releasedNoSync()`, which runs remote node synchronization via `ServerEntry::sync()` — left `_releasing` permanently set: subsequent `release()` calls block forever on `_condVar` and new allocation requests queue without ever being drained.

Such an escape is possible: `NodeEntry::loadServer`/`destroyServer` only catch `NodeUnreachableException`, and `ServerEntry::syncImpl` only catches `NodeNotExistException`, so a synchronously-thrown `Ice::CommunicatorDestroyedException` from `loadServerAsync`/`destroyServerAsync` (e.g. when a session is released during registry shutdown) propagates all the way into `release()`.

The fix wraps everything that runs after `_releasing` is set in a `try/catch (...)` that clears the flag, notifies waiting threads, and rethrows. The two normal return paths keep their existing clear-and-notify, which must stay where it is: it runs in the same critical section that observes the request queue (empty, or re-allocated), so no allocation request can be queued behind a completed drain. A destructor- or catch-timed clear on those paths would open a window where a concurrent allocation is queued after the drain's final queue check and stranded until the next allocate/release cycle.

Neither fix changes behavior on any normal path, so there is no CHANGELOG entry.

### DIFF
--- a/cpp/src/IceGrid/Allocatable.cpp
+++ b/cpp/src/IceGrid/Allocatable.cpp
@@ -218,105 +218,122 @@ Allocatable::release(const shared_ptr<SessionI>& session, bool fromRelease)
         }
     }
 
-    // If we set _releasing above, clear it and notify waiting threads when we leave this function, no matter how.
-    ReleasingGuard releasingGuard(*this, hasRequests);
-
-    if (isReleased)
+    try
     {
-        releasedNoSync(session);
-    }
-
-    if (_parent)
-    {
-        _parent->release(session, fromRelease);
-    }
-    else if (!fromRelease)
-    {
-        if (hasRequests)
+        if (isReleased)
         {
-            while (true)
+            releasedNoSync(session);
+        }
+
+        if (_parent)
+        {
+            _parent->release(session, fromRelease);
+        }
+        else if (!fromRelease)
+        {
+            if (hasRequests)
             {
-                shared_ptr<AllocationRequest> request;
-                shared_ptr<Allocatable> allocatable;
+                while (true)
                 {
-                    lock_guard lock(_mutex);
-                    allocatable = dequeueAllocationAttempt(request);
-                    if (!allocatable)
+                    shared_ptr<AllocationRequest> request;
+                    shared_ptr<Allocatable> allocatable;
                     {
-                        assert(_requests.empty());
-                        assert(_count == 0);
-                        return;
-                    }
-                }
-
-                //
-                // Try to allocate the allocatable with the request or if
-                // there's no request, just notify the allocatable that it can
-                // be allocated again.
-                //
-                try
-                {
-                    if ((request && allocatable->allocate(request, true)) ||
-                        (!request && allocatable->canTryAllocate()))
-                    {
-                        while (true)
+                        lock_guard lock(_mutex);
+                        allocatable = dequeueAllocationAttempt(request);
+                        if (!allocatable)
                         {
+                            assert(_requests.empty());
+                            assert(_count == 0);
+                            _releasing = false;
+                            _condVar.notify_all();
+                            return;
+                        }
+                    }
+
+                    //
+                    // Try to allocate the allocatable with the request or if
+                    // there's no request, just notify the allocatable that it can
+                    // be allocated again.
+                    //
+                    try
+                    {
+                        if ((request && allocatable->allocate(request, true)) ||
+                            (!request && allocatable->canTryAllocate()))
+                        {
+                            while (true)
                             {
-                                lock_guard lock(_mutex);
-                                assert(_count);
-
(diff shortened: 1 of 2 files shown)


pepone (commented):
L1 is correct — nice catch that `canTryAllocate()` reaches the outer `catch` with a null `request` via `ObjectNotRegisteredException` (`checkAllocatable()` → not an `AllocationException`, so `tryAllocate`'s catch misses it; not a `SessionDestroyedException`, so `TypeEntry::canTryAllocate`'s catch misses it). The `if (request)` guard is right.

The `ReleasingGuard` also correctly fixes the exception-path leak, and I confirmed it can't self-deadlock (every inner lock is released before the destructor re-acquires `_mutex`).

One blocker, though: routing the **normal-path** clear through the guard breaks an atomicity that the old code relied on.

Pre-PR, at the empty-queue return the clear ran in the *same* critical section that had just observed `_requests` empty:

```cpp
lock_guard lock(_mutex);
allocatable = dequeueAllocationAttempt(request);
if (!allocatable)
{
    assert(_requests.empty());
    _releasing = false;      // same lock as the emptiness check
    _condVar.notify_all();
    return;
}
```

Now the check and the clear are in two different lock acquisitions (the `return` releases `_mutex`, then `~ReleasingGuard()` re-locks to clear). In that gap `_releasing` is still `true` with the lock free, so a concurrent `allocate(..., fromRelease=false)` from another session:

- `!_session && (fromRelease || !_releasing)` → `true && false` → skips allocation
- `_session == request->getSession()` → `null == session` → false
- falls to `queueAllocationAttempt(...)` and **queues the request**

The drain has already exited, so nothing drains it and the guard then clears `_releasing`. The request is stranded in `_requests` — the allocation hangs until the next allocate+release on the object (or its allocation timeout), even though the object is free. Pre-PR this couldn't happen because the emptiness check and the clear were atomic.

Suggestion: keep `_releasing = false; _condVar.notify_all();` under the lock at the two normal `return`s (preserving the atomic check-and-clear) and let the guard cover only the exception paths — e.g. deactivate it once a normal return has done the clear. That keeps the L6 exception-safety fix without reopening this window.

bernardnormier:
@pepone Good catch on the check-and-clear atomicity — the drain's final queue observation and the clearing of `_releasing` do need to happen in the same critical section, otherwise an allocation queued in the gap is stranded behind the finished drain, exactly as you described.

Fixed in 25c19c5a48, going one step further than the dismissable-guard suggestion: since the normal-path clears have to stay inline anyway, the RAII guard no longer bought much, so I dropped it entirely. The two normal returns keep their original in-lock clear-and-notify, and an enclosing `try/catch (...)` now clears the flag and rethrows on exceptional exits only (covering both the drain loop and `releasedNoSync()`, which was one of the original leak paths). The `IceGrid/allocation` suite passes, including the concurrent-allocation and stress tests.

pepone (approved):
Approve.

Atomic check-and-clear is restored — both drain returns clear `_releasing` + `notify_all()` under the same lock that observed the queue, so the stranding window is gone. Dropping the guard outright reads cleaner than deactivating it, and the enclosing `try/catch(...)` also covers the `releasedNoSync()` leak path. L1 null-`request` guard is intact.

One non-blocking note: on an *exceptional* drain abort the catch clears `_releasing` but doesn't re-drive requests already queued — they wait for the next allocate/release to make progress. Fine as-is (strictly better than the stuck-flag it replaces), just flagging it.

PR #5911, opened 2026-07-09
Don't drop detected server terminations on a pipe read error in the Activator

bernardnormier:
Fixes item L3 of #5844.

In the POSIX termination listener, once one server's status pipe reached EOF, its process was moved to the local `terminated` vector and erased from `_processes`. If a later fd in the same `select()` batch then failed its `read()`, the listener threw immediately: the `terminated` vector was destroyed and `server->terminated()` was never called for the already-erased processes, leaving those servers permanently "active" on the node (the listener restart re-scans `_processes`, which no longer contains them).

Two changes:

- A hard read failure is now captured in an `exception_ptr` and rethrown only after the detected terminations have been dispatched. The rethrow still lands in `runTerminationListener`'s catch, which logs the error and restarts the listener, exactly as before.
- `read()` is retried on `EINTR` instead of treating an interrupted call as a pipe failure — consistent with the `select()` call above, which already retries on `EINTR`/`EPROTO`.

Internal robustness with no realistic reported symptom, so no CHANGELOG entry. Windows uses per-process wait handles and has no shared failure path, so it is unaffected.

### DIFF
--- a/cpp/src/IceGrid/Activator.cpp
+++ b/cpp/src/IceGrid/Activator.cpp
@@ -1316,6 +1316,7 @@ Activator::terminationListener()
 
         vector<Process> terminated;
         bool deactivated = false;
+        exception_ptr readException;
         {
             lock_guard lock(_mutex);
 
@@ -1346,9 +1347,21 @@ Activator::terminationListener()
                 //
                 // Read the message over the pipe.
                 //
-                while ((rs = read(fd, &s, 16)) > 0) // NOLINT(clang-analyzer-unix.BlockInCriticalSection)
+                while (true)
                 {
-                    message.append(s, static_cast<size_t>(rs));
+                    rs = read(fd, &s, 16); // NOLINT(clang-analyzer-unix.BlockInCriticalSection)
+                    if (rs > 0)
+                    {
+                        message.append(s, static_cast<size_t>(rs));
+                    }
+                    else if (rs == -1 && errno == EINTR)
+                    {
+                        continue; // Interrupted by a signal: retry.
+                    }
+                    else
+                    {
+                        break; // EOF or error.
+                    }
                 }
 
                 //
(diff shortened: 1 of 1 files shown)


pepone (approved):
Approving — traced this end-to-end and it correctly closes the drop.

Capturing the read failure in an `exception_ptr` and rethrowing it after the `for (terminated)` dispatch loop means every already-erased server gets `terminated()` before the exception propagates. `SyscallException` derives from `Ice::Exception`, so it still lands in `runTerminationListener`'s `catch (const Exception&)` → log → restart, exactly as the old immediate throw did — just deferred.

A few things I checked:

- The read `throw` was the only throw in the detection lock scope (the other early `return` fires before any termination is detected, so `terminated` is empty). Nothing else between the erase and the dispatch loop throws.
- `deactivated` can't be shadowed by the new rethrow-first ordering: a hard read error leaves the failed process in `_processes`, so `deactivated` is necessarily false when `readException` is set.
- Breaking out early only defers the other ready fds — pipe EOF/error is level-triggered, so `select()` re-reports them on restart. Strictly better than before, which abandoned the collected terminations too.
- The `EINTR` retry is safe: `process.pipeFd` is `O_NONBLOCK` (set right after fork), so the read-under-lock returns `EAGAIN` rather than blocking, and `continue` only fires on transient `EINTR`; after the loop `rs` is always 0 or -1 and the `errno != EAGAIN || message.empty()` test is unchanged from before.

Agree with the no-CHANGELOG call (internal robustness, no realistic reported symptom) and the Windows exemption (per-process wait handles, no shared failure path).

cpp/src/IceGrid/Activator.cpp
```diff
@@ -1346,9 +1347,21 @@ Activator::terminationListener()
                 //
                 // Read the message over the pipe.
                 //
-                while ((rs = read(fd, &s, 16)) > 0) // NOLINT(clang-analyzer-unix.BlockInCriticalSection)
+                while (true)
                 {
-                    message.append(s, static_cast<size_t>(rs));
+                    rs = read(fd, &s, 16); // NOLINT(clang-analyzer-unix.BlockInCriticalSection)
```
pepone, line 1352: Not for this PR — but while we're in this read loop: the `16` here and the `char s[16]` above are the same magic number written twice, so they have to be kept in sync by hand (drift → truncated read or a stack overflow). Filed #5917 to switch these to `read(fd, s, sizeof(s))` across the three copies (here, the child-side read at ~L871, and Ice/Service.cpp).

PR #5915, opened 2026-07-09
Dispatch to the default servant when the identity is registered with a different facet

### LINKED ISSUE #5865: [Audit-Medium] C# default servant is bypassed when the identity is registered with a different facet
### Summary
In the C# mapping, the default-servant fallback only runs when the request's identity is entirely absent from the Active Servant Map. If the identity is registered with a *different* facet, the default servant is never consulted and the client receives `FacetNotExistException` — diverging from both the documented lookup order and the C++ implementation, which dispatch to the default servant.

### Mechanism
C# `ServantManager.findServant` consults `_defaultServantMap` only in the `else` branch of the identity lookup, i.e. only when `_servantMapMap` has no entry for the identity at all ([`ServantManager.cs:178-200`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/csharp/src/Ice/Internal/ServantManager.cs#L178-L200)). When the identity exists but the facet doesn't, it returns null; `dispatchAsync` then finds `hasServant(current.id)` true and throws `FacetNotExistException` ([`ServantManager.cs:69-76`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/csharp/src/Ice/Internal/ServantManager.cs#L69-L76)).

C++ falls through to the default-servant lookup when *either* the identity *or* the facet is missing ([`ServantManager.cpp:204-218`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Ice/ServantManager.cpp#L204-L218)).

Both mappings document the C++ order — "tries to find a servant for the identity and facet in the Active Servant Map; if this fails, … a default servant for the category" ([`ObjectAdapter.h:149-162`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/include/Ice/ObjectAdapter.h#L149-L162), [`ObjectAdapter.cs:484-503`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/csharp/src/Ice/ObjectAdapter.cs#L484-L503)).

### Trigger
```
adapter.addFacet(facetServant, id, "x");
adapter.addDefaultServant(defaultServant, id.category);
```
A request for identity `id` with the default facet `""` dispatches to `defaultServant` in C++ but throws `FacetNotExistException` in C# (servant locators, consulted between the two, are unaffected).

### Impact
Servers mixing per-facet ASM registrations with default servants for the same category behave differently in C# than in every mapping that follows the documented order; requests that should reach the default servant fail with `FacetNotExistException`.

### Suggested fix
In `findServant`, fall back to the `_defaultServantMap` lookup when the facet lookup fails, not only when the identity is absent — mirroring the C++ logic.

### Notes
Confidence: high. Found in the 2026-07-07 C++/C# API documentation audit (cross-mapping divergence B.1); doc claims and both implementations re-verified against source.

---
*Filed from an automated AI audit of the Ice core API (Claude Fable 5, 2026-07-07); confirmed through a multi-agent adversarial refute pass and independent Codex cross-verification over the actual source. Not yet human-triaged.*

### LINKED ISSUE #5875: [Audit-Medium] JS: default servant is bypassed when the identity is registered with a different facet

bernardnormier:
Fixes #5865.
Fixes #5875.

The default-servant fallback in `ServantManager.findServant` ran only when the request's identity was entirely absent from the Active Servant Map. When the identity was registered with a *different* facet, `findServant` returned null without consulting the default servant map, and the dispatch failed with `FacetNotExistException` — diverging from the documented lookup order ("tries to find a servant for the identity and facet in the Active Servant Map; if this fails, … a default servant for the category") and from the C++ implementation, which falls through to the default servant when either the identity or the facet lookup fails.

#5865 reported the C# mapping and #5875 the JS mapping, but the same faulty lookup was copied into every mapping with its own dispatch path: **C#, Java, JS, and Swift** are all fixed here, with the same restructure in each — perform the identity+facet lookup first, then fall back to the default-servant chain (category, then empty category) whenever it produced nothing. Python dispatches through the C++ core and is unaffected; Ruby, PHP, and MATLAB have no server side.

The existing `Ice/defaultServant` test in each fixed mapping — plus C++, as a guard on the reference behavior — now registers a facet servant on an identity while a default servant is active for its category and verifies that a request on the default facet dispatches to the default servant. All five pass; the new check fails on the pre-fix code.

One changelog fragment per fixed mapping.

Note: the same bug exists on the 3.7 branch (including `java-compat`); a hand-ported backport PR will follow, since the file layout and Swift internals differ there.

### DIFF
--- a/changelog.d/csharp/default-servant-facet-fallback.md
+++ b/changelog.d/csharp/default-servant-facet-fallback.md
@@ -1,0 +1,3 @@
+- Fixed the servant lookup for incoming requests: when the target identity is registered with a different facet, the
+  object adapter now dispatches the request to the default servant if one is registered, as documented, instead of
+  rejecting the request with `FacetNotExistException`.
--- a/changelog.d/java/default-servant-facet-fallback.md
+++ b/changelog.d/java/default-servant-facet-fallback.md
@@ -1,0 +1,3 @@
+- Fixed the servant lookup for incoming requests: when the target identity is registered with a different facet, the
+  object adapter now dispatches the request to the default servant if one is registered, as documented, instead of
+  rejecting the request with `FacetNotExistException`.
--- a/changelog.d/js/default-servant-facet-fallback.md
+++ b/changelog.d/js/default-servant-facet-fallback.md
@@ -1,0 +1,3 @@
+- Fixed the servant lookup for incoming requests: when the target identity is registered with a different facet, the
+  object adapter now dispatches the request to the default servant if one is registered, as documented, instead of
+  rejecting the request with `FacetNotExistException`.
--- a/changelog.d/swift/default-servant-facet-fallback.md
+++ b/changelog.d/swift/default-servant-facet-fallback.md
@@ -1,0 +1,3 @@
+- Fixed the servant lookup for incoming requests: when the target identity is registered with a different facet, the
+  object adapter now dispatches the request to the default servant if one is registered, as documented, instead of
+  rejecting the request with `FacetNotExistException`.
--- a/cpp/test/Ice/defaultServant/AllTests.cpp
+++ b/cpp/test/Ice/defaultServant/AllTests.cpp
@@ -118,6 +118,15 @@ allTests(Test::TestHelper* helper)
         }
     }
 
+    // The identity is registered with a facet servant only: a request on the default facet still dispatches to the
+    // default servant.
+    identity.category = "foo";
+    identity.name = "x";
+    oa->addFacet(std::make_shared<MyObjectI>(), identity, "theFacet");
+    prx = oa->createProxy<MyObjectPrx>(identity);
+    test(prx->getName() == "x");
+    oa->removeFacet(identity, "theFacet");
+
     oa->removeDefaultServant("foo");
     identity.category = "foo";
     prx = oa->createProxy<MyObjectPrx>(identity);
--- a/csharp/src/Ice/Internal/ServantManager.cs
+++ b/csharp/src/Ice/Internal/ServantManager.cs
@@ -186,7 +186,8 @@ internal sealed class ServantManager : Object
             {
                 m.TryGetValue(facet, out obj);
             }
-            else
+
+            if (obj is null)
             {
                 _defaultServantMap.TryGetValue(ident.category, out obj);
                 if (obj is null)
--- a/csharp/test/Ice/defaultServant/AllTests.cs
+++ b/csharp/test/Ice/defaultServant/AllTests.cs
@@ -115,6 +115,15 @@ public class AllTests : global::Test.AllTests
             }
         }
 
+        // The identity is registered with a facet servant only: a request on the default facet still dispatches
+        // to the default servant.
+        identity.category = "foo";
+        identity.name = "x";
+        oa.addFacet(new MyObjectI(), identity, "theFacet");
+        prx = Test.MyObjectPrxHelper.uncheckedCast(oa.createProxy(identity));
+        test(prx.getName() == "x");
+        oa.removeFacet(identity, "theFacet");
+
         oa.removeDefaultServant("foo");
         identity.category = "foo";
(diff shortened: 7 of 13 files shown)


pepone (approved):
Approving. The core fix is correct and consistent across C#, Java, JS, and Swift — each does the identity+facet ASM lookup first, then falls through to the default-servant chain (category, then "") when it yields nothing, matching the C++ reference in `ServantManager` and the documented lookup order. Only the intended behavior changes; the found-facet fast path and identity-absent path are untouched. Swift's `servantMapMap[id]?[facet]` optional-chaining correctly falls through both when the identity is absent and when it's present under a different facet.

For the record on the JS test (in case it looks odd): the adapter is intentionally bidirectional-only (`createObjectAdapter("")`) since JS has no incoming connections, and dispatch works through the `__echo` bounce-back — the same mechanism every other dispatch in that test already uses. The new lines are a valid regression check: pre-fix `foo/x` on the default facet throws `FacetNotExistException` (the identity is registered under `theFacet` only), post-fix it falls through to the default servant and `getName()` returns `"x"`.

Minor (non-blocking): both the default and facet servants are `MyObjectI`, so `getName() == "x"` proves the request was dispatched rather than which servant handled it — still a valid red/green guard since pre-fix reaches neither.

InsertCreativityHere (approved):
Looks good to me.

PR #5920, opened 2026-07-09
Fix Local Metadata not Being Correctly Applied to Optional Params in Python

### LINKED ISSUE #5801: slice2py: optional branch of typeToTypeHintString drops parameter metadata
### Description

In `cpp/src/slice2py/PythonUtil.cpp`, the `optional` branch of `CodeVisitor::typeToTypeHintString` recurses without forwarding the `localMetadata` argument:

```cpp
if (optional)
{
    if (isProxyType(type) || type->isClassType())
    {
        // We map optional proxies and classes like regular ones.
        return typeToTypeHintString(type, false, source, forMarshaling);          // drops localMetadata
    }
    else
    {
        return typeToTypeHintString(type, false, source, forMarshaling) + " | None"; // drops localMetadata
    }
}
```

As a result, metadata-driven type hints are lost for optional parameters/return values. For example:

```slice
sequence<int> IntSeq;

interface I
{
    void op(["python:numpy.ndarray"] optional(1) IntSeq values);
}
```

The generated hint for `values` omits the `numpy.typing.NDArray[...]` alternative, on both the proxy signature and the servant abstract method (and, consequently, the operation docstring). The non-optional case produces the expected hint.

### Fix

Forward `localMetadata` on the recursive calls in the `optional` branch.

### Notes

Pre-existing behavior, discovered during review of #5795 (which fixed the non-optional parameter-metadata case and made the proxy side consistent with the servant side). It was out of scope for that PR since the two sides already matched for optional types.

InsertCreativityHere:
This PR fixes #5801.
The code-path for optional params simply wasn't passing along the local metadata like it should of.


pepone (approved):

changelog.d/python/local-metadata-on-optional-parameters.md
```diff
@@ -0,0 +1 @@
+- Fixed a bug where local metadata that was directly applied to optional parameters was not being honored by `slice2py`.
```
bernardnormier: The metadata was honored at runtime (the operation table always carried it), so the bug was only in the generated type hints. Suggest making the entry say that:

```suggestion
- Fixed the type hints generated by `slice2py` for optional parameters: metadata applied directly to the parameter (such as `python:numpy.ndarray`) was ignored, so the hint omitted the mapped type.
```

bernardnormier (approved):

PR #5921, opened 2026-07-09
Support bcrypt password hashes in the Glacier2 Crypt permissions verifier

### LINKED ISSUE #5862: [Audit-Low] Glacier2 bcrypt password entries never verify on Linux/FreeBSD — crypt salt truncated at the last '$'
### Summary
`CryptPermissionsVerifier` derives the crypt "setting" by truncating the stored hash at its **last** `$`. That is correct for `$1$`/`$5$`/`$6$`/yescrypt but wrong for bcrypt (`$2a$`/`$2b$`/`$2y$`), whose 22-char salt has no trailing `$`. bcrypt entries therefore fail to verify for every login, and the failure looks like a wrong password.

### Mechanism
`salt = stored.substr(0, rfind('$') + 1)` ([`CryptPermissionsVerifier.cpp:176-207`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Glacier2/CryptPermissionsVerifier.cpp#L176-L207)). In `$2y$10$<22-char-salt><31-char-hash>` the last `$` is the one after the cost, so the setting passed to `crypt_r` is `"$2y$10$"` with the salt cut off. libxcrypt/FreeBSD `crypt` rejects a bcrypt setting without its salt (returns null or a `*0` failure token), so the length check at [`:203`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Glacier2/CryptPermissionsVerifier.cpp#L203) fails and `checkPermissions` returns false.

### Trigger
An admin puts a bcrypt hash (e.g. from `htpasswd -B`, or FreeBSD's default scheme) in `Glacier2.CryptPasswords` on Linux (libxcrypt) or FreeBSD.

### Impact
Those users can never create a session; the failure is a silent `permission denied` that looks like a wrong password.

### Suggested fix
Pass the **whole stored hash** as the crypt setting — the canonical `crypt(pw, stored)` verification idiom — which works for bcrypt and every other scheme (including the ones that work today). This also makes the code robust to future formats.

### Notes
Caveat: `icehashpassword.py` never emits bcrypt (sha512/256-crypt on Linux, PBKDF2 on macOS/Windows) and no in-repo doc promises bcrypt — hence Low. Long-standing. Confidence: high on the mechanism. Finding 8 of the 2026-07-07 Glacier2 correctness audit.

---

bernardnormier:
Fixes #5862.

On Linux, the Glacier2 Crypt permissions verifier derived the crypt "setting" by truncating the stored hash at its last `$`. That works for DES, `$1$`, `$5$`/`$6$`, and `$y$` hashes, but not for bcrypt (`$2a$`/`$2b$`/`$2y$`), whose 22-character salt has no trailing `$`: the setting passed to `crypt_r` lost its salt, the call failed, and the entry never verified.

This PR passes the whole stored hash as the setting instead — the canonical `crypt(password, storedHash)` idiom, since crypt only reads the salt prefix of its second argument. This works for every scheme supported by the system's crypt library, including bcrypt. The DES special case is subsumed: crypt uses only the first 2 characters as the DES salt, and the existing output-length + constant-time comparison still rejects any mismatch or failure token.

Also adds a comment documenting the platform-specific accepted hash formats (crypt-based on Linux, passlib-style PBKDF2 on macOS and Windows).

### DIFF
--- a/.vscode/cspell.json
+++ b/.vscode/cspell.json
@@ -25,6 +25,7 @@
         "icestorm",
         "istr",
         "jgoodies",
+        "libxcrypt",
         "LMDB",
         "nohup",
         "nullopt",
@@ -47,6 +48,7 @@
         "unregisters",
         "upcall",
         "upcalls",
+        "yescrypt",
         "zeroc"
     ],
     "flagWords": [
--- a/changelog.d/service-glacier2/crypt-verifier-bcrypt.md
+++ b/changelog.d/service-glacier2/crypt-verifier-bcrypt.md
@@ -1,0 +1,1 @@
+- The Glacier2 Crypt permissions verifier now supports bcrypt password hashes (`$2a$`, `$2b$`, `$2y$`) on Linux.
--- a/cpp/src/Glacier2/CryptPermissionsVerifier.cpp
+++ b/cpp/src/Glacier2/CryptPermissionsVerifier.cpp
@@ -165,6 +165,15 @@ namespace
     }
 #endif
 
+    // Verifies the password against the stored hash for the given user ID. The accepted hash formats are
+    // platform-specific:
+    //
+    // - Linux and FreeBSD: any format supported by the system's crypt library; with libxcrypt this includes DES,
+    //   $1$ (MD5), $5$/$6$ (sha-crypt), $2b$ (bcrypt), and $y$ (yescrypt).
+    // - macOS and Windows: passlib-style PBKDF2 only ($pbkdf2$, $pbkdf2-sha256$, $pbkdf2-sha512$).
+    //
+    // On Windows, the BCrypt* APIs used below belong to Microsoft's CNG (Cryptography API: Next Generation); they
+    // are unrelated to the bcrypt password-hashing scheme, which is not supported on Windows.
     bool CryptPermissionsVerifierI::checkPermissions(string userId, string password, string&, const Current&) const
     {
         auto p = _passwords.find(userId);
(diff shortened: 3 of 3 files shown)


pepone (approved):
Approve.

Passing the whole stored hash as the `crypt_r` setting is the right call — canonical `crypt(pw, stored)` idiom, DES stays subsumed (crypt takes the first 2 chars as salt), and the length + constant-time compare still gate. The comment documenting the per-platform accepted formats is a nice touch.

One scope note for the record: this now accepts *any* full-setting scheme the platform `crypt` recognizes, not just bcrypt — e.g. BSD extended DES (`_…`), which the old 13-char precheck rejected. That's in line with the "robust to future formats" intent, just flagging it's broader than bcrypt-only. (bcrypt cost is admin-controlled, so a high-cost entry makes logins expensive — expected, not a concern.)

Non-blocking follow-up: the `router` test already exercises `checkPermissions` end-to-end via `Glacier2.CryptPasswords`, but only with the formats `icehashpassword` emits (sha512-crypt on Linux, PBKDF2 on mac/win) — so **bcrypt** specifically is uncovered, which is why #5862 slipped through. A pinned bcrypt entry added to a password-file test would close it. Gotchas: gate on `__GLIBC__ || __FreeBSD__` (not "linux" — musl skips the crypt branch), and the skip-probe must call native `crypt` directly rather than go through `checkPermissions`, or the regression itself masquerades as "bcrypt unsupported" and skips.

PR #5923, opened 2026-07-09
Support session creation from secure WebSocket (wss) connections in Glacier2 and IceGrid

### LINKED ISSUE #5860: [Audit-Medium] Glacier2 createSessionFromSecureConnection rejects secure-WebSocket (wss) clients as "not ssl connection"
### Summary
`createSessionFromSecureConnection` does a direct `dynamic_pointer_cast` to `Ice::SSL::ConnectionInfo` that doesn't walk the `underlying` chain, so a `wss` (WebSocket-over-TLS) client — TLS-secured, with a peer certificate — is rejected with `PermissionDeniedException("not ssl connection")`. SSL-certificate session creation is impossible over the `wss` transport.

### Mechanism
```cpp
auto info = dynamic_pointer_cast<Ice::SSL::ConnectionInfo>(current.con->getInfo());
if (!info) { exception(make_exception_ptr(PermissionDeniedException("not ssl connection"))); return; }
```
[`SessionRouterI.cpp:668-673`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Glacier2/SessionRouterI.cpp#L668-L673). For a `wss` connection the stack is WS → SSL → TCP, and `getInfo()` returns a top-level `WSConnectionInfo` — a `final` class deriving from `ConnectionInfo`, **not** from `SSL::ConnectionInfo` ([`Connection.h:337-352`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/include/Ice/Connection.h#L337-L352)) — whose `underlying` holds the SSL info ([`WSTransceiver.cpp:838-843`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Ice/WSTransceiver.cpp#L838-L843)). The direct cast returns null. The adjacent `getIPConnectionInfo` helper walks `underlying` correctly ([`SessionRouterI.cpp:16-29`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Glacier2/SessionRouterI.cpp#L16-L29)); the SSL cast here (and at [`:349`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Glacier2/SessionRouterI.cpp#L349) for the `_con.peerCert` context) does not, so `_con.peerCert` is also silently dropped for `wss` clients.

### Trigger
A client connects over `wss` with a client certificate and calls `createSessionFromSecureConnection`.

### Impact
`PermissionDeniedException("not ssl connection")` on a TLS-secured, certificate-bearing connection — contradicting Router.ice ("authenticated through the SSL certificate(s) associated with the connection"). Certificate-authenticated sessions are unavailable over `wss`.

### Suggested fix
Walk the `underlying` chain to find the SSL info (a `getSSLConnectionInfo` helper mirroring `getIPConnectionInfo`), used at both `:668` and `:349`.

### Notes
Long-standing (same in 3.7). The identical pattern exists in IceGrid `RegistryI.cpp:1159` — a shared helper would fix both. Confidence: high; CONFIRMED by a Codex refute pass. Finding 6 of the 2026-07-07 Glacier2 correctness audit.

---

bernardnormier:
Fixes #5860.

The Glacier2 router and the IceGrid registry located the SSL connection info with a direct `dynamic_pointer_cast` on `Connection::getInfo()`. For a wss connection, `getInfo()` returns a `WSConnectionInfo` whose `underlying` holds the SSL info, so the cast returned null and `createSessionFromSecureConnection` (and, for IceGrid, `createAdminSessionFromSecureConnection`) failed with a `PermissionDeniedException`, even though the connection is TLS-secured and carries a client certificate. Glacier2 also skipped the `_con.peerCert` context entry for wss connections for the same reason.

Both services now walk the `underlying` chain to find the SSL connection info, with a `getSSLConnectionInfo` helper mirroring the existing `getIPConnectionInfo`. Glacier2's "not ssl connection" exception message also becomes "not an ssl connection", matching IceGrid.

Test coverage:
- The `Glacier2/ssl` test client now also creates a certificate session over a wss endpoint, and the server-side connection-context checks were generalized from a tcp/ssl boolean to a connection-type string, additionally asserting that `_con.peerCert` is set on secure connections.
- The secure-session blocks in the `IceGrid/session` and `IceGrid/replication` tests now run when the test protocol is `wss` in addition to `ssl`. This was required anyway: the `IceGrid/session` test previously asserted that session creation from a secure connection is denied for any protocol other than `ssl`, which codified the old behavior.

### DIFF
--- a/changelog.d/service-glacier2/wss-secure-session.md
+++ b/changelog.d/service-glacier2/wss-secure-session.md
@@ -1,0 +1,3 @@
+- The Glacier2 router now supports `createSessionFromSecureConnection` calls from clients connecting over secure
+  WebSocket (wss), and sets the `_con.peerCert` context entry for these connections when
+  `Glacier2.AddConnectionContext` is enabled.
--- a/changelog.d/service-icegrid/wss-secure-session.md
+++ b/changelog.d/service-icegrid/wss-secure-session.md
@@ -1,0 +1,2 @@
+- The IceGrid registry now supports `createSessionFromSecureConnection` and `createAdminSessionFromSecureConnection`
+  calls from clients connecting over secure WebSocket (wss).
--- a/cpp/src/Glacier2/SessionRouterI.cpp
+++ b/cpp/src/Glacier2/SessionRouterI.cpp
@@ -27,6 +27,20 @@ namespace
 
         return nullptr;
     }
+
+    shared_ptr<Ice::SSL::ConnectionInfo> getSSLConnectionInfo(const ConnectionInfoPtr& info)
+    {
+        for (ConnectionInfoPtr p = info; p; p = p->underlying)
+        {
+            auto sslInfo = dynamic_pointer_cast<Ice::SSL::ConnectionInfo>(p);
+            if (sslInfo)
+            {
+                return sslInfo;
+            }
+        }
+
+        return nullptr;
+    }
 }
 
 namespace Glacier2
@@ -346,7 +360,7 @@ CreateSession::CreateSession(shared_ptr<SessionRouterI> sessionRouter, string us
             }
         }
         {
-            auto info = dynamic_pointer_cast<Ice::SSL::ConnectionInfo>(current.con->getInfo());
+            auto info = getSSLConnectionInfo(current.con->getInfo());
             if (info && info->peerCertificate)
             {
                 _context["_con.peerCert"] = Ice::SSL::encodeCertificate(info->peerCertificate);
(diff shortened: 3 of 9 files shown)


pepone (approved):
Approve.

Walking the `underlying` chain for the SSL info is the right fix, and it's symmetric across both services. The cert gate is preserved — a wss client still has to present a valid certificate (and non-empty subject DN), so no downgrade vs ssl.

Worth noting the semantics for the record: this accepts *any* transport with an SSL layer somewhere in its chain, not just wss — which is correct, since a transport layered over SSL is secure by construction, and the cert check runs on that SSL layer regardless of the wrapper.

PR #5924, opened 2026-07-09
Reject invalid Glacier2.RoutingTable.MaxSize values at startup

### LINKED ISSUE #5863: [Audit-Low] Glacier2 RoutingTable.MaxSize ≤ 0 is unvalidated — UB/crash (negative) or per-invocation livelock (zero)
### Summary
`Glacier2.RoutingTable.MaxSize` is read as an `int` with no floor. A negative value dereferences an empty list (undefined behavior / crash) on the first client invocation; `0` immediately evicts every added proxy, producing a per-invocation client↔router livelock. Neither is validated or warned.

### Mechanism
`_maxSize` is read with no validation ([`RoutingTable.cpp:13`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Glacier2/RoutingTable.cpp#L13)). `add()` inserts the proxy then evicts while `static_cast<int>(_map.size()) > _maxSize` ([`RoutingTable.cpp:102-116`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Glacier2/RoutingTable.cpp#L102-L116)):
- `MaxSize=-1`: after the just-added entry is evicted (size 1 → 0), `0 > -1` is still true → `_queue.front()` on an **empty `std::list`** (UB) and `_map.erase` on a garbage iterator.
- `MaxSize=0`: every added proxy is evicted immediately; the client then re-drives `ice_add_proxy` on the `ObjectNotExistException` retry (which never counts against the retry limit) → add → evict → invoke → retry, forever, with no completion and no error.

### Trigger
Misconfiguration: `Glacier2.RoutingTable.MaxSize=-1` (crash) or `=0` (livelock, e.g. an operator expecting "unlimited"). Default is 1000.

### Impact
Router crash (negative) or a silent hot request loop that never completes (zero).

### Suggested fix
Reject `MaxSize < 1` at construction (fatal `InitializationException`), or clamp with a warning.

### Notes
Confidence: high; CONFIRMED by a Codex refute pass. Finding 11 of the 2026-07-07 Glacier2 correctness audit.

---

bernardnormier:
Fixes #5863.

`Glacier2.RoutingTable.MaxSize` was read without validation. A negative value made the routing table's eviction loop pop from an empty queue (undefined behavior, crash) on the first `ice_add_proxy`, and `0` evicted every proxy immediately after adding it, causing clients to retry `ice_add_proxy` forever.

The Glacier2 router now rejects `MaxSize < 1` at startup with `InitializationException` ("invalid value for Glacier2.RoutingTable.MaxSize: 0"), reported through the existing "Glacier2 initialization failed" startup error path.

The property is now read and validated once in `Instance` and injected into the `RoutingTable` constructor, instead of being re-read from the properties by every per-session routing table.

### DIFF
--- a/cpp/src/Glacier2/Instance.cpp
+++ b/cpp/src/Glacier2/Instance.cpp
@@ -17,8 +17,17 @@ Glacier2::Instance::Instance(
       _logger(_communicator->getLogger()),
       _clientAdapter(std::move(clientAdapter)),
       _serverAdapter(std::move(serverAdapter)),
-      _proxyVerifier(make_shared<ProxyVerifier>(_communicator))
+      _proxyVerifier(make_shared<ProxyVerifier>(_communicator)),
+      _routingTableMaxSize(_properties->getIcePropertyAsInt("Glacier2.RoutingTable.MaxSize"))
 {
+    if (_routingTableMaxSize < 1)
+    {
+        throw Ice::InitializationException(
+            __FILE__,
+            __LINE__,
+            "invalid value for Glacier2.RoutingTable.MaxSize: " + to_string(_routingTableMaxSize));
+    }
+
     //
     // If an Ice metrics observer is setup on the communicator, also enable metrics for Glacier2.
     //
--- a/cpp/src/Glacier2/Instance.h
+++ b/cpp/src/Glacier2/Instance.h
@@ -25,6 +25,7 @@ namespace Glacier2
 
         [[nodiscard]] std::shared_ptr<ProxyVerifier> proxyVerifier() const { return _proxyVerifier; }
         [[nodiscard]] std::shared_ptr<SessionRouterI> sessionRouter() const { return _sessionRouter; }
+        [[nodiscard]] int routingTableMaxSize() const { return _routingTableMaxSize; }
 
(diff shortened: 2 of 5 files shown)


pepone (approved):
Looks good. I verified:

- Rejecting `MaxSize < 1` in the `Instance` constructor prevents both failure modes from #5863: with
  `_maxSize >= 1`, the eviction loop in `RoutingTable::add` can no longer pop from an empty queue
  (negative values) or immediately evict the proxy it just added (zero).
- The `InitializationException` is caught by the existing handler in `Glacier2Router.cpp` and reported
  as "Glacier2 initialization failed" before startup aborts.
- `RouterI` is the only place that constructs a `RoutingTable`, and nothing else reads
  `Glacier2.RoutingTable.MaxSize`, so reading and validating the property once in `Instance` covers
  everything.
- The error message and exception type match the existing convention for this kind of validation
  (e.g. `IceGrid.Registry.NodeSessionTimeout` in `InternalRegistryI.cpp`).

A nice side effect of moving the read to `Instance`: a non-numeric value (e.g. `MaxSize=abc`) now fails
at startup with a `PropertyException` instead of failing at the first session creation.

Issue #5889, opened 2026-07-08
[Audit-Medium] Python: Connection.setCloseCallback rejects bound methods and other callables


pepone:
### Summary
`Connection.setCloseCallback` only accepts plain functions (and `None`): the implementation gates the argument with `isinstance(cb, types.FunctionType)`, so bound methods, `functools.partial` objects, callable instances, and builtin functions all raise `ValueError: callback must be None or a function`. C++ accepts any invocable.

### Mechanism
[`Connection.cpp:601-606`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Connection.cpp#L601-L606) — `lookupType("types.FunctionType")` + `PyObject_IsInstance`. Python semantics (verified): `isinstance(obj.method, types.FunctionType)` is `False` for bound methods, partials, callable objects, and builtins; only plain functions and lambdas pass. The stub also omits the accepted `None` (clears the callback): [`IcePy-stubs/__init__.pyi:212`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/python/IcePy-stubs/__init__.pyi#L212).

### Trigger
```python
connection.setCloseCallback(self.on_closed)   # ValueError
```

### Impact
The most natural callback form in object-oriented code (a bound method) is rejected; users must wrap in a lambda.

### Suggested fix
Use `PyCallable_Check` instead of the `FunctionType` isinstance check, and declare `Callable[[Connection], None] | None` in the stub.

### Notes
Confidence: high. Found in the 2026-07-08 Python API doc audit (finding A.4).

---
*Filed from an automated AI audit of the Ice core API (Claude Fable 5, 2026-07-08); confirmed through a multi-agent adversarial refute pass and independent Codex cross-verification over the actual source. Not yet human-triaged.*

PR #5927, opened 2026-07-09
Destroy the replaced close callback outside the connection lock

### LINKED ISSUE #5926: [Audit-Medium] C++: setCloseCallback deadlocks when the replaced callback's destructor re-enters the connection
### Summary

`Connection::setCloseCallback` destroys the callback it replaces **while holding the connection's mutex**. A callback destructor can run user code, and that code can call back into the connection. `ConnectionI::_mutex` is a plain `std::mutex`, so the re-entrant lock deadlocks the process.

### Mechanism

[`ConnectionI.cpp:764-779`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/cpp/src/Ice/ConnectionI.cpp#L764-L779) takes `std::lock_guard lock(_mutex)` and then assigns [`:777`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/cpp/src/Ice/ConnectionI.cpp#L777) `_closeCallback = std::move(callback);`. The assignment destroys the previous `std::function`, and therefore whatever it captured, still under the lock. `_mutex` is non-recursive ([`ConnectionI.h:418`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/cpp/src/Ice/ConnectionI.h#L418)).

In Ice for Python this is easy to reach: [`~CloseCallbackWrapper`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/python/modules/IcePy/Connection.cpp#L229-L235) reacquires the GIL and `Py_DECREF`s the Python callable, which can run an arbitrary `__del__`.

`finish()` also clears `_closeCallback`, but it holds no lock there, so that path is fine. `setCloseCallback` is the only affected site.

### Trigger

Reproduced against `4a1ed88e9fc27c82d7871879d0187733e69df0d3` (Python; a plain lambda, so it does not depend on any pending change to `setCloseCallback`):

```python
class Reentrant:
    def __init__(self, holder): self.h = holder
    def __del__(self):
        con = self.h.get("con")
        if con is not None:
            con.setCloseCallback(None)     # re-enters ConnectionI::setCloseCallback

con = prx.ice_getConnection()
holder = {"con": con}

def make(r):
    return lambda c: r                     # the lambda's closure owns the Reentrant
con.setCloseCallback(make(Reentrant(holder)))

con.setCloseCallback(lambda c: None)       # destroys the old callback under _mutex -> HANGS
```

A pure C++ trigger is a functor whose destructor calls back into the connection.

### Impact

Process-wide deadlock, not just the calling thread: the connection's mutex is never released.

The Python surface makes this more likely than it looks. A close callback is a natural place to hold a session object, and dropping the last reference to it during callback replacement runs its finalizer. Widening `setCloseCallback` to accept any callable (#5889) makes a callable instance with a `__del__` an equally direct trigger.

### Suggested fix

Move the previous callback out of the member under the lock and let it be destroyed after the lock is released:

```cpp
CloseCallback previous;
{
    std::lock_guard lock(_mutex);
    ...
    previous = std::exchange(_closeCallback, std::move(callback));
}
// previous destroyed here, outside the lock
```

A deterministic regression test is possible in `cpp/test/Ice/ami`: hold a re-entrant destructor in a `shared_ptr` captured by the callback, so it fires exactly once when the connection releases it. The test deadlocks against the current code and passes against the fix.

### Notes

Affects `main` and the `3.8` branch (identical code). Confidence: high — reproduced by execution, and the proposed fix verified against both a C++ and a Python repro.

---
*Found by an independent AI cross-check (Codex, `gpt-5.6-sol`) of the fix for #5889, on 2026-07-10. Unlike the other `ai-audit` issues, this one has been reproduced and the fix verified locally; it still warrants human review.*

pepone:
Fixes #5926.

`ConnectionI::setCloseCallback` assigned the new callback to `_closeCallback` while holding `_mutex`, so the callback being replaced was destroyed under that lock. A callback destructor can run user code, and that code can call back into the connection — `_mutex` is a plain `std::mutex`, so the re-entrant lock deadlocks the process.

This is reachable today from Ice for Python: `~CloseCallbackWrapper` reacquires the GIL and decrefs the Python callable, running its `__del__`. A lambda closing over an object whose finalizer touches the connection hangs. C++ callers can hit it with a functor whose destructor calls back into the connection.

The fix moves the previous callback into a local under the lock and lets it be destroyed after the `lock_guard` goes out of scope. `finish()` also clears `_closeCallback`, but holds no lock there, so it needs no change.

### Testing

New test in `cpp/test/Ice/ami`, which holds a re-entrant destructor in a `shared_ptr` captured by the callback so it fires exactly once. Verified it **deadlocks against the current code** and passes against the fix. The two Python repros in #5926 likewise go from hang to clean exit.

### Merge order

Please merge this **before** #5889 (`Accept any callable in Connection.setCloseCallback`): widening the accepted callable types makes a callable instance with a `__del__` an equally direct trigger for this deadlock.

### DIFF
--- a/cpp/src/Ice/ConnectionI.cpp
+++ b/cpp/src/Ice/ConnectionI.cpp
@@ -763,18 +763,25 @@ Ice::ConnectionI::flushBatchRequestsAsync(
 void
 Ice::ConnectionI::setCloseCallback(CloseCallback callback)
 {
-    std::lock_guard lock(_mutex);
-    if (_state >= StateClosed)
+    // Holds the callback this call replaces, so that it is destroyed after the lock is released. Destroying a
+    // callback can run arbitrary user code - a language mapping may attach a finalizer to it - and that code can
+    // call back into this connection, which would deadlock on the non-recursive _mutex.
+    CloseCallback previous;
+
     {
-        if (callback)
+        std::lock_guard lock(_mutex);
+        if (_state >= StateClosed)
         {
-            auto self = shared_from_this();
-            _threadPool->execute([self, callback = std::move(callback)]() { self->closeCallback(callback); }, self);
+            if (callback)
+            {
+                auto self = shared_from_this();
+                _threadPool->execute([self, callback = std::move(callback)]() { self->closeCallback(callback); }, self);
+            }
         }
-    }
-    else
-    {
-        _closeCallback = std::move(callback);
+        else
+        {
+            previous = std::exchange(_closeCallback, std::move(callback));
+        }
     }
 }
 
--- a/cpp/src/Ice/ConnectionI.cpp
+++ b/cpp/src/Ice/ConnectionI.cpp
@@ -26,6 +26,7 @@
 #include <algorithm>
 #include <iomanip>
 #include <stdexcept>
+#include <utility>
 
 #ifdef ICE_HAS_BZIP2
 #    include <bzlib.h>
--- a/changelog.d/cpp/5926.md
+++ b/changelog.d/cpp/5926.md
@@ -1,0 +1,3 @@
+- Fixed a deadlock in `Connection::setCloseCallback`. Setting or clearing a close callback destroyed the callback it
+  replaced while holding the connection's lock, so a callback whose destructor called back into the connection
+  deadlocked. This affected the language mappings that attach a finalizer to the callback, such as Ice for Python.
--- a/cpp/test/Ice/ami/AllTests.cpp
+++ b/cpp/test/Ice/ami/AllTests.cpp
@@ -1102,7 +1102,11 @@ allTests(TestHelper* helper, bool collocated)
                 struct Reenter
                 {
                     ConnectionPtr connection;
-                    ~Reenter() { connection->getInfo(); } // reenter the connection from the callback's destructor
+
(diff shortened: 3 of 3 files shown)


cpp/src/Ice/ConnectionI.cpp
```diff
@@ -763,18 +763,25 @@ Ice::ConnectionI::flushBatchRequestsAsync(
...
-            _threadPool->execute([self, callback = std::move(callback)]() { self->closeCallback(callback); }, self);
+            if (callback)
+            {
+                auto self = shared_from_this();
+                _threadPool->execute([self, callback = std::move(callback)]() { self->closeCallback(callback); }, self);
+            }
+        }
+        else
+        {
+            previous = std::exchange(_closeCallback, std::move(callback));
```
bernardnormier, line 784: This is the first use of `std::exchange` in the code base, and this file doesn't include `<utility>` — it compiles only because some other header drags it in transitively. Please add `#include <utility>` to the standard-library include block at the top of the file.

bernardnormier (approved):

PR #5928, opened 2026-07-09
Fix ice_getConnectionAsync for collocated proxies and the event loop adapter

### LINKED ISSUE #5888: [Audit-Medium] Python: ice_getConnectionAsync mishandles collocated proxies and bypasses the event-loop adapter
### Summary
Two defects in `ice_getConnectionAsync`: (1) for a collocated proxy — where the C++ core delivers a null `ConnectionPtr` — the returned future either never resolves or resolves to a `Connection` wrapper around a null pointer; (2) the future is not passed through `wrapFuture`, so under an asyncio `EventLoopAdapter` it is the only async proxy API that yields a non-asyncio future.

### Mechanism
1. `GetConnectionAsyncCallback` uses `_connection` as its "already completed" sentinel ([`Operation.cpp:2395-2414`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Operation.cpp#L2395-L2414)): when the null response arrives before `setFuture`, `if (_connection)` stays false and the future is never settled; when it arrives after, [`Operation.cpp:2417-2437`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Operation.cpp#L2417-L2437) calls `createConnection(null, ...)` which wraps the null pointer ([`Connection.cpp:835-845`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Connection.cpp#L835-L845)). The sync `ice_getConnection` correctly returns `None` for this case ([`Proxy.cpp:1100`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Proxy.cpp#L1100)). The core delivers the (unset) `_cachedConnection` for collocated proxies (`cpp/src/Ice/ProxyAsync.cpp:294-306`).
2. [`Proxy.cpp:1104-1137`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Proxy.cpp#L1104-L1137) returns the raw future; `ice_invokeAsync` and `ice_flushBatchRequestsAsync` wrap theirs via `IcePy::wrapFuture` ([`Proxy.cpp:1200`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Proxy.cpp#L1200), [`Operation.cpp:1450-1459`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Operation.cpp#L1450-L1459)).

### Trigger
`await prx.ice_getConnectionAsync()` in an asyncio application (non-asyncio future), or on a collocated proxy (hang / broken wrapper).

### Impact
Hangs or broken objects on the collocated path; asyncio integration broken for this one API.

### Suggested fix
Resolve the future with `None` for a null connection (mirroring the sync path), settle regardless of arrival order, and return `wrapFuture(...)` like the sibling async APIs.

### Notes
Confidence: high. Found in the 2026-07-08 Python API doc audit (finding A.3).

---

pepone:
Fixes #5888.

`GetConnectionAsyncCallback` used its `_connection` member as the "response already delivered" sentinel. A collocated proxy has no connection, so the core completes the operation with a null `ConnectionPtr` and the sentinel never becomes true:

- when the response arrived **before** `setFuture`, the future was never settled;
- when it arrived **after**, `createConnection` wrapped the null pointer and returned a `Connection` object that **segfaults the interpreter** on first use.

The synchronous `ice_getConnection` already returned `None` here. Note the issue understates this: it is a crash, not just a hang or a broken wrapper.

```
sync  ice_getConnection()      -> None
async ice_getConnectionAsync() -> <IcePy.Connection object at 0x...>
      con.toString()           -> exit code 139 (SIGSEGV)
```

The fix tracks the response with a dedicated `_responseReceived` flag and resolves the future with `None` when the connection is null, so both orderings settle and both mirror the synchronous operation.

`proxyIceGetConnectionAsync` also returned its future raw — the only `createFuture` site in the extension that skipped `IcePy::wrapFuture`. Under an asyncio `EventLoopAdapter` it was the one async proxy operation yielding a non-asyncio future. It now wraps like `ice_flushBatchRequestsAsync`, and the two method-table docstrings that omitted `None` from the return type are corrected.

### Testing

New block in `Ice/asyncio`, which already asserts `isinstance(f, asyncio.Future)` for sibling async calls. Verified it **fails against the unfixed extension** (`AssertionError` on the `isinstance` check) and passes with it. A remote proxy still returns a real `Connection`; a collocated one returns `None`. `Ice/{ami,proxy,operations,timeout,info,asyncio,exceptions,blobject,adapterDeactivation}` all pass.

### DIFF
--- a/changelog.d/python/5888.md
+++ b/changelog.d/python/5888.md
@@ -1,0 +1,5 @@
+- Fixed `ice_getConnectionAsync` on a collocated proxy. It resolved with an unusable `Connection` object that crashed
+  the interpreter as soon as it was used, or never resolved at all, instead of resolving with `None` like the
+  synchronous `ice_getConnection`.
+- Fixed `ice_getConnectionAsync` to return a future created by the configured event loop adapter, like the other
+  asynchronous proxy operations. Awaiting it in an asyncio application no longer yields a non-asyncio future.
--- a/python/modules/IcePy/Operation.cpp
+++ b/python/modules/IcePy/Operation.cpp
@@ -2396,9 +2396,9 @@ IcePy::GetConnectionAsyncCallback::setFuture(PyObject* future)
     //
     // Check if any callbacks have been invoked already.
     //
-    if (_connection)
+    if (_responseReceived)
     {
-        PyObjectHandle pyConn{createConnection(_connection, _communicator)};
+        PyObjectHandle pyConn{_connection ? createConnection(_connection, _communicator) : Py_NewRef(Py_None)};
         assert(pyConn.get());
         PyObjectHandle tmp{callMethod(future, "set_result", pyConn.get())};
         PyErr_Clear();
@@ -2425,10 +2425,11 @@ IcePy::GetConnectionAsyncCallback::response(const Ice::ConnectionPtr& conn)
         // The future hasn't been set yet, which means the request is still being invoked. Save the results for later.
         //
         _connection = conn;
+        _responseReceived = true;
         return;
     }
 
-    PyObjectHandle pyConn{createConnection(conn, _communicator)};
+    PyObjectHandle pyConn{conn ? createConnection(conn, _communicator) : Py_NewRef(Py_None)};
     PyObjectHandle tmp{callMethod(_future, "set_result", pyConn.get())};
     PyErr_Clear();
 
--- a/python/modules/IcePy/Operation.h
+++ b/python/modules/IcePy/Operation.h
@@ -38,6 +38,9 @@ namespace IcePy
         std::string _op;
         PyObject* _future{nullptr};
         Ice::ConnectionPtr _connection;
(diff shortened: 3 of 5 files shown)


bernardnormier (approved):
Looks good.

PR #5930, opened 2026-07-09
Accept the request context as a keyword argument in ice_invoke

### LINKED ISSUE #5887: [Audit-Medium] Python: ice_invokeAsync silently drops a context passed as keyword argument
### Summary
`proxy.ice_invokeAsync(op, mode, inParams, ctx={...})` sends the request with **no context**: the C function is registered `METH_KEYWORDS` but ignores its `kwds` argument. The sync `ice_invoke` is `METH_VARARGS` only, so the same keyword call raises `TypeError` — while the IcePy stub declares a keyword-capable `ctx` parameter on both.

### Mechanism
[`Proxy.cpp:1209-1213`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Proxy.cpp#L1209-L1213) — `proxyIceInvokeAsync(ProxyObject* self, PyObject* args, PyObject* /*kwds*/)` discards keywords; the positional-only parser then sees no `ctx` and the invocation goes out with `noExplicitContext` ([`Operation.cpp:1797-1841`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Operation.cpp#L1797-L1841)). Method table flags: [`Proxy.cpp:1419-1428`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Proxy.cpp#L1419-L1428). Stub: [`IcePy-stubs/__init__.pyi:553-558`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/python/IcePy-stubs/__init__.pyi#L553-L558).

### Trigger
```python
prx.ice_invokeAsync("op", Ice.OperationMode.Normal, params, ctx={"k": "v"})
# type-checks against the stub; request is sent WITHOUT the context
```

### Impact
Silent wrong behavior: per-invocation contexts (routing hints, session tokens) are dropped without any error for the dynamic-invocation API.

### Suggested fix
Parse keywords in `proxyIceInvokeAsync` (and either add keyword support to `ice_invoke` or mark `ctx` positional-only in the stub for both).

### Notes
Confidence: high. Found in the 2026-07-08 Python API doc audit (finding A.2).

---

pepone:
Fixes #5887.

`ice_invokeAsync` was registered `METH_VARARGS | METH_KEYWORDS` but its implementation discarded `kwds` and parsed the positional tuple only, so

```python
prx.ice_invokeAsync(op, mode, inParams, ctx={"k": "v"})
```

sent the request **with no context at all** — silently dropping routing hints and session tokens. The synchronous `ice_invoke` was registered `METH_VARARGS`, so the same call raised `TypeError` instead. Both stubs declare a keyword-capable `ctx`, and the slice2py-generated proxies already accept their `context` by keyword.

This threads `kwds` from the proxy methods down to the two blobject invocations, which now parse with `PyArg_ParseTupleAndKeywords` over the keyword names the stubs already publish, and registers `ice_invoke` with `METH_KEYWORDS` so the pair behaves the same. Positional calls parse exactly as before, so the change is purely additive.

These are the first `PyArg_ParseTupleAndKeywords` calls in the extension.

### Testing

The blobject test router now records the context it receives, so a dropped context fails an assertion rather than passing silently. Verified against the unfixed extension:

```
sync  keyword ctx -> TypeError: ObjectPrx.ice_invoke() takes no keyword arguments
async keyword ctx -> delivered {}          <-- silently dropped
```

and after the fix both deliver the context. `Ice/{blobject,ami,operations,proxy,asyncio,adapterDeactivation,servantLocator}` pass; pyright and ruff are clean.

### DIFF
--- a/python/modules/IcePy/Operation.h
+++ b/python/modules/IcePy/Operation.h
@@ -18,8 +18,8 @@ namespace IcePy
     extern PyTypeObject AsyncInvocationContextType;
 
     // Blobject invocations.
-    PyObject* iceInvoke(PyObject*, PyObject*);
-    PyObject* iceInvokeAsync(PyObject*, PyObject*);
+    PyObject* iceInvoke(PyObject*, PyObject*, PyObject* = nullptr);
+    PyObject* iceInvokeAsync(PyObject*, PyObject*, PyObject* = nullptr);
 
     /// Used as the callback for getConnectionAsync operation.
     class GetConnectionAsyncCallback
--- a/python/modules/IcePy/Operation.h
+++ b/python/modules/IcePy/Operation.h
@@ -18,8 +18,8 @@ namespace IcePy
     extern PyTypeObject AsyncInvocationContextType;
 
     // Blobject invocations.
-    PyObject* iceInvoke(PyObject*, PyObject*, PyObject* = nullptr);
-    PyObject* iceInvokeAsync(PyObject*, PyObject*, PyObject* = nullptr);
+    PyObject* iceInvoke(PyObject*, PyObject*, PyObject*);
+    PyObject* iceInvokeAsync(PyObject*, PyObject*, PyObject*);
 
     /// Used as the callback for getConnectionAsync operation.
     class GetConnectionAsyncCallback
--- a/changelog.d/python/5887.md
+++ b/changelog.d/python/5887.md
@@ -1,0 +1,3 @@
+- Fixed `ObjectPrx.ice_invoke` and `ObjectPrx.ice_invokeAsync` to accept the request context as a keyword argument, as
+  their signatures advertise. `ice_invokeAsync` previously sent the request without any context when `ctx` was passed by
+  keyword, and `ice_invoke` rejected the call with a `TypeError`.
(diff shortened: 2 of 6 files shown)


python/modules/IcePy/Operation.h
```diff
@@ -18,8 +18,8 @@ namespace IcePy
     extern PyTypeObject AsyncInvocationContextType;
 
     // Blobject invocations.
-    PyObject* iceInvoke(PyObject*, PyObject*);
-    PyObject* iceInvokeAsync(PyObject*, PyObject*);
+    PyObject* iceInvoke(PyObject*, PyObject*, PyObject* = nullptr);
+    PyObject* iceInvokeAsync(PyObject*, PyObject*, PyObject* = nullptr);
```
bernardnormier: Both callers now always pass `kwds`, so the defaults only make it possible for a future call site to silently drop the keyword arguments — the mistake this PR fixes. Better to make every caller provide it:

```suggestion
    PyObject* iceInvoke(PyObject*, PyObject*, PyObject*);
    PyObject* iceInvokeAsync(PyObject*, PyObject*, PyObject*);
```

bernardnormier (approved):

PR #5931, opened 2026-07-09
Accept any callable in Connection.setCloseCallback

### LINKED ISSUE #5889: [Audit-Medium] Python: Connection.setCloseCallback rejects bound methods and other callables

pepone:
Fixes #5889.

> [!IMPORTANT]
> Please merge **after** #5927. Widening the accepted callable types makes a callable instance with a `__del__` a direct trigger for the pre-existing `setCloseCallback` deadlock fixed there (#5926).

`setCloseCallback` gated its argument with an `isinstance` check against `types.FunctionType`, which only plain functions and lambdas satisfy. A bound method — the most natural callback in object-oriented code — plus `functools.partial` objects, callable instances and builtin functions were all rejected with `ValueError`, even though C++ accepts any invocable and the wrapper only ever calls the object.

It now gates with `PyCallable_Check`. This was the **last** `types.FunctionType` check in the extension; every other callback setter (`PropertiesAdmin`, `Executor`, `Thread`, `BatchRequestInterceptor`, `Communicator`, `Types`, `PySliceLoader`) already used `PyCallable_Check`, so the family is finally uniform — including its `TypeError` for a non-callable.

The stub and docstring also omitted the accepted `None`, which clears the callback.

### Compatibility

Widening a type check cannot break a call that previously worked. The only observable change for incorrect code is `ValueError` becoming `TypeError`, which matches the sibling setters; no test or caller depended on it (the suite's only `setCloseCallback` call passes a lambda).

`CloseCallbackWrapper` takes a strong reference via `Py_NewRef`, so accepting temporaries such as bound methods and `functools.partial` objects is safe — but dropping that reference is what exposes #5926, hence the merge order above.

### Testing

`Ice/ami` now asserts that a bound method, a `functools.partial`, a callable instance and `None` are all accepted, and that a non-callable raises `TypeError`. Verified all four forms are rejected with `ValueError` against the unfixed extension.

### DIFF
--- a/python/modules/IcePy/Connection.cpp
+++ b/python/modules/IcePy/Connection.cpp
@@ -142,15 +142,15 @@ Awaitable[None]
     A future that becomes available when the flush completes.)";
 
     constexpr const char* connectionSetCloseCallback_doc =
-        R"(setCloseCallback(callback: Callable[[Connection], None]) -> None
+        R"(setCloseCallback(callback: Callable[[Connection], None] | None) -> None
 
 Sets a close callback on the connection. The callback is called by the connection when it's closed.
 The callback is called from the Ice thread pool associated with the connection.
 
 Parameters
 ----------
-callback : Callable[[Connection], None]
-    The close callback callable.)";
+callback : Callable[[Connection], None] | None
+    The close callback callable, or None to remove the current callback.)";
 
     constexpr const char* connectionType_doc = R"(type() -> str
 
@@ -598,10 +598,9 @@ connectionSetCloseCallback(ConnectionObject* self, PyObject* args)
         return nullptr;
     }
 
-    PyObject* callbackType = lookupType("types.FunctionType");
-    if (cb != Py_None && !PyObject_IsInstance(cb, callbackType))
+    if (cb != Py_None && !PyCallable_Check(cb))
     {
-        PyErr_Format(PyExc_ValueError, "callback must be None or a function");
+        PyErr_Format(PyExc_TypeError, "callback must be None or a callable");
         return nullptr;
     }
 
--- a/python/python/IcePy-stubs/__init__.pyi
+++ b/python/python/IcePy-stubs/__init__.pyi
@@ -209,15 +209,15 @@ class Connection:
         """
         ...
 
-    def setCloseCallback(self, callback: Callable[[Connection], None]) -> None:
+    def setCloseCallback(self, callback: Callable[[Connection], None] | None) -> None:
         """
         Sets a close callback on the connection. The callback is called by the connection when it's closed.
         The callback is called from the Ice thread pool associated with the connection.
 
         Parameters
         ----------
-        callback : Callable[[Connection], None]
-            The close callback callable.
+        callback : Callable[[Connection], None] | None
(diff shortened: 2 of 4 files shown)


python/modules/IcePy/Connection.cpp
```diff
@@ -142,15 +142,15 @@ Awaitable[None]
...
 
 Sets a close callback on the connection. The callback is called by the connection when it's closed.
 The callback is called from the Ice thread pool associated with the connection.
 
 Parameters
 ----------
-callback : Callable[[Connection], None]
-    The close callback callable.)";
+callback : Callable[[Connection], None] | None
+    The close callback callable, or None to remove the current callback.)";
```
pepone, line 154: We should probably mention the TypeError in its doc comment.

python/python/IcePy-stubs/__init__.pyi
```diff
@@ -209,15 +209,15 @@ class Connection:
         """
         ...
 
-    def setCloseCallback(self, callback: Callable[[Connection], None]) -> None:
+    def setCloseCallback(self, callback: Callable[[Connection], None] | None) -> None:
         """
```
pepone, line 213: Here too document what it can raise

python/modules/IcePy/Connection.cpp
bernardnormier: I don't think it's necessary.

python/python/IcePy-stubs/__init__.pyi
bernardnormier: That's minor.

bernardnormier (approved):

PR #5933, opened 2026-07-10
Fix incorrect and stale API doc comments in the Java Ice library

pepone:
Corrects incorrect and stale API doc comments in the **com.zeroc.ice** mapping, mirroring the C++ (#5867) and C#
(#5868) doc fixes. Found by a doc audit that compared each Java javadoc against the implementation and against the
already-merged C++/C# wording. Only the modern com.zeroc.ice mapping is covered (as with the other per-language doc
PRs).

### Incorrect documented behavior

- **`SliceInfo.bytes`** — dropped "including the leading size integer"; the captured bytes exclude the 4-byte slice
  size (re-marshaling via `writeSlicedData` would otherwise double it). **The same doc bug exists in the other
  mappings and is *not* fixed here — follow-up needed:** C++ `cpp/include/Ice/SlicedData.h:33`, C#
  `csharp/src/Ice/SlicedData.cs:18` (the `bytes` param), and JS `js/packages/ice/src/Ice/UnknownSlicedValue.d.ts`.
  Neither #5867 nor #5868 touched it.
- **`ObjectPrx.ice_getContext`** — the returned map is never `null` (a fresh, possibly empty, map).
- **`Util.getProcessLogger`/`setProcessLogger`** — property-configured loggers (`Ice.LogFile`, syslog) take
  precedence, and a default process logger is replaced per communicator, so it is not "used by all communicators
  that do not have their own specific logger".
- **`Communicator.createObjectAdapter(String, SSLEngineFactory)`** — passing `null` is equivalent to
  `createObjectAdapter(String)`, not `createObjectAdapterWithEndpoints(String, String)` (copy-paste).
- **`ServantLocator.deactivate`** — the adapter is being *destroyed*, not deactivated (locators are only notified
  from `ServantManager.destroy` ← `ObjectAdapter.destroy`).
- **`WSConnectionInfo.headers`** — request headers on an incoming connection, response headers on an outgoing one.
- **`Connection.setAdapter`** — documents that it throws `UnsupportedOperationException` on incoming connections.
- **`OutputStream.writeValue`** — the "write an index, defer the state to `writePendingValues`" behavior is 1.0-only;
  the default 1.1 encoding marshals the state eagerly.
- **`Value.ice_getSlicedData`** — dropped the stale 3.7-era "preserved-slice base class" wording; preservation is
  unconditional in the sliced format.
- **`InitializationException`** — broadened "communicator initialization fails" to "a failure occurs during
  initialization" (matches C#; it's also thrown from property/adapter/logger/SSL setup).

### Doc-quality fixes

- `InputStream`/`OutputStream` class summaries no longer call the `final class` an "Interface".
- `OutputStream.endEncapsulation` — "current" (not "previous") encapsulation.
- `ToStringMode` — escape examples render with a single backslash (`\t`, `\unnnn`), not `\\t`.
- `ObjectAdapter.add` — `@throws AlreadyRegisteredException` now says "identity **and default facet**".
- `Instrumentation.ObserverUpdater` — removed a stray `=` that rendered mid-sentence.
- `PluginManager.initializePlugins` — markdown backticks → `{@code}` (they don't render in `/** */` javadoc).
- `ObjectPrx.equals` / `ProxyIdentityKey` / `ProxyIdentityFacetKey` — grammar.
- `Util.stringToIdentity` — the exception message now says "unescaped slash" (it fires on an extra `/`, not a `\`).

### Not included (tracked separately)

- The default-servant facet-fallback **code bug** that the audit reproduced in Java is already fixed cross-mapping by
  #5915 (no doc/code change needed here).
- Coverage-gap `@throws` additions (e.g. `setProperty`, `getDefaultLocator`, `createObjectAdapter`) and the
  Instrumentation "run-time"→"runtime" / "This method is called when…" rephrasing are cross-mapping cleanups left for
  follow-up, to keep this PR consistent with the C++/C# scope.

Full findings: `API_DOC_AUDIT_JAVA-2026-07-09.md` (ice-audit).

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
@@ -33,7 +33,7 @@ public interface Connection {
      * the default object adapter of an outgoing connection is the communicator's default object adapter.
      *
      * @param adapter the object adapter to associate with this connection
-     * @throws UnsupportedOperationException Thrown when this connection is an incoming connection: you can only call
+     * @throws UnsupportedOperationException if this connection is an incoming connection: you can only call
      *     this method on outgoing (client) connections.
      * @see #getAdapter
      * @see Communicator#setDefaultObjectAdapter
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
@@ -33,6 +33,8 @@ public interface Connection {
      * the default object adapter of an outgoing connection is the communicator's default object adapter.
      *
      * @param adapter the object adapter to associate with this connection
+     * @throws UnsupportedOperationException Thrown when this connection is an incoming connection: you can only call
+     *     this method on outgoing (client) connections.
      * @see #getAdapter
      * @see Communicator#setDefaultObjectAdapter
      */
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SliceInfo.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SliceInfo.java
@@ -10,7 +10,7 @@ public final class SliceInfo {
     /** The Slice compact type ID for this slice, or {@code -1} if the slice has no compact ID. */
     public final int compactId;
 
-    /** The encoded bytes for this slice, including the leading size integer. */
+    /** The encoded bytes for this slice. */
     public final byte[] bytes;
 
     /** The class instances referenced by this slice. */
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ToStringMode.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ToStringMode.java
@@ -24,8 +24,8 @@ public enum ToStringMode {
 
     /**
      * Characters with ordinal values greater than 127 are encoded as a sequence of UTF-8 bytes using octal escapes.
-     * Characters with ordinal values 127 and below are encoded as \t, \n (etc.) or an octal escape.
-     * Use this mode to generate strings compatible with Ice 3.6 and earlier.
+     * Non-printable ASCII characters with ordinal values 127 and below are encoded as \t, \n (etc.) or an octal
+     * escape. Use this mode to generate strings compatible with Ice 3.6 and earlier.
      */
     Compat(2);
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ToStringMode.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ToStringMode.java
@@ -10,21 +10,21 @@ package com.zeroc.Ice;
 public enum ToStringMode {
     /**
      * Characters with ordinal values greater than 127 are kept as-is in the resulting string.
-     * Non-printable ASCII characters with ordinal values 127 and below are encoded as \\t, \\n
-     * (etc.) or \\unnnn.
+     * Non-printable ASCII characters with ordinal values 127 and below are encoded as \t, \n
+     * (etc.) or &#92;unnnn.
      */
     Unicode(0),
 
     /**
      * Characters with ordinal values greater than 127 are encoded as universal character names in the resulting string:
-     * \\unnnn for BMP characters and \\Unnnnnnnn for non-BMP characters.
-     * Non-printable ASCII characters with ordinal values 127 and below are encoded as \\t, \\n (etc.) or \\unnnn.
+     * &#92;unnnn for BMP characters and \Unnnnnnnn for non-BMP characters.
+     * Non-printable ASCII characters with ordinal values 127 and below are encoded as \t, \n (etc.) or &#92;unnnn.
      */
     ASCII(1),
 
     /**
      * Characters with ordinal values greater than 127 are encoded as a sequence of UTF-8 bytes using octal escapes.
-     * Characters with ordinal values 127 and below are encoded as \\t, \\n (etc.) or an octal escape.
+     * Characters with ordinal values 127 and below are encoded as \t, \n (etc.) or an octal escape.
      * Use this mode to generate strings compatible with Ice 3.6 and earlier.
      */
     Compat(2);
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
@@ -149,7 +149,7 @@ public final class Util {
                     slash = pos;
                 } else {
                     // Extra unescaped slash found.
-                    throw new ParseException("unescaped backslash in identity string '" + s + "'");
+                    throw new ParseException("unescaped slash in identity string '" + s + "'");
                 }
             }
             pos++;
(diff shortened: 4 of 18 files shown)


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SliceInfo.java
```diff
@@ -10,7 +10,7 @@ public final class SliceInfo {
     /** The Slice compact type ID for this slice, or {@code -1} if the slice has no compact ID. */
     public final int compactId;
 
-    /** The encoded bytes for this slice, including the leading size integer. */
+    /** The encoded bytes for this slice. */
```
bernardnormier, line 13: We should fix this in all languages, in particular C++.

bernardnormier (approved):

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
```diff
@@ -33,6 +33,8 @@ public interface Connection {
      * the default object adapter of an outgoing connection is the communicator's default object adapter.
      *
      * @param adapter the object adapter to associate with this connection
+     * @throws UnsupportedOperationException Thrown when this connection is an incoming connection: you can only call
```
InsertCreativityHere: This is not the style of exception documentation we use in Java. I believe `Thrown when` is really only what we do in C++. `if ...` is the convention we use for Java.
```suggestion
     * @throws UnsupportedOperationException if this connection is an incoming connection: you can only call
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ToStringMode.java
```diff
@@ -10,21 +10,21 @@
...
-     * Non-printable ASCII characters with ordinal values 127 and below are encoded as \\t, \\n (etc.) or \\unnnn.
+     * &#92;unnnn for BMP characters and \Unnnnnnnn for non-BMP characters.
+     * Non-printable ASCII characters with ordinal values 127 and below are encoded as \t, \n (etc.) or &#92;unnnn.
      */
     ASCII(1),
 
     /**
      * Characters with ordinal values greater than 127 are encoded as a sequence of UTF-8 bytes using octal escapes.
-     * Characters with ordinal values 127 and below are encoded as \\t, \\n (etc.) or an octal escape.
+     * Characters with ordinal values 127 and below are encoded as \t, \n (etc.) or an octal escape.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Done in adc1a73967 — the Compat doc now says "Non-printable ASCII characters with ordinal values 127 and below…", matching the sibling entries and the C++ reference wording (verified against `StringUtil.encodeChar`, which keeps printable ASCII as-is in every mode).

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/SliceInfo.java
pepone: 🤖 **Claude:** Opened #5949 for the remaining mappings (C++, C#, JS, MATLAB) — verified in each stream implementation that the capture starts after the size integer is consumed. Python is already fixed on #5893 and Swift on #5941.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -149,7 +149,7 @@ public static Identity stringToIdentity(String s) {
                     slash = pos;
                 } else {
                     // Extra unescaped slash found.
-                    throw new ParseException("unescaped backslash in identity string '" + s + "'");
+                    throw new ParseException("unescaped slash in identity string '" + s + "'");
```
InsertCreativityHere, line 152: Why make this change?
I'm pretty sure we only care about backslashes as far as escaping is concerned.
Also note that we say "backslash" not just "slash" in our C# code too.
https://github.com/zeroc-ice/ice/blob/3d42d58a127952b2847b112d73d9e66878dfd93a/csharp/src/Ice/Util.cs#L85

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
pepone: 🤖 **Claude:** Done in 7d83449f57.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
pepone: 🤖 **Claude:** The throw fires on a *second unescaped forward slash* — the backslash counting above it only decides whether each `/` is escaped, per the `// Extra unescaped slash found.` comment. C++ reports `unescaped '/' in identity` (`cpp/src/Ice/Initialize.cpp:103`), so the old Java message named the wrong character. The C# message you linked has the same bug for the same condition — fixed in #5949.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Value.java
```diff
@@ -48,8 +48,8 @@ public String ice_id() {
     /**
      * Returns the sliced data associated with this instance.
      *
-     * @return If this value has a preserved-slice base class and has been sliced during unmarshaling, this returns the
-     *     sliced data; otherwise this returns {@code null}.
+     * @return The sliced data if this value was sliced during unmarshaling; otherwise {@code null}. Unknown slices are
```
InsertCreativityHere: To be more consistent with the comment we have for this in C++, C#, JS, and Python:
```suggestion
     * @return The sliced data if this value was sliced during unmarshaling, {@code null} otherwise. Unknown slices are
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ToStringMode.java
```diff
@@ -10,22 +10,22 @@
 public enum ToStringMode {
     /**
      * Characters with ordinal values greater than 127 are kept as-is in the resulting string.
-     * Non-printable ASCII characters with ordinal values 127 and below are encoded as \\t, \\n
-     * (etc.) or \\unnnn.
+     * Non-printable ASCII characters with ordinal values 127 and below are encoded as \t, \n
```
InsertCreativityHere, line 13: For the record, I checked the rest of the code, and this was the only file using this `\\` escape incorrectly. Everywhere else looks good!

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
InsertCreativityHere: Yes, I was misreading the code.
Glad it led to the bug getting fixed in the other languages at least!

InsertCreativityHere (approved):
Looks good to me!

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Value.java
pepone: 🤖 **Claude:** Applied to Java in 4956416638. C# turned out to be the one divergent mapping ("; otherwise null") — kept as-is since that is the .NET-conventional phrasing; #5956 tracks aligning the C# docs with the .NET conventions repo-wide. MATLAB never got the corrected wording at all (stale "preserved-slice base class" + no null/sliced-format sentence) and will be covered by a full MATLAB doc review pass.

PR #5902, opened 2026-07-09
Fix IceStorm subscriber and topic robustness issues

pepone:
Fixes #5851
Fixes #5852
Fixes #5853
Fixes #5854

- **#5851** — `TopicImpl::destroy` marked the topic destroyed *before* its database transaction committed, so a failed commit left an undestroyable topic that still delivered events. `_destroyed` is now set only after `destroyInternal` commits.
- **#5852** — A replica reconciling its subscribers matched them by identity only, so a subscriber that re-registered with a different proxy or QoS while the replica was out of sync kept its stale in-memory record (and could push it cluster-wide on promotion). The in-memory subscriber is now replaced when the record differs.
- **#5853** — `subscribeAndGetPublisher` and `observerAddSubscriber` registered the per-subscriber publisher servant *before* the database transaction; on a transaction failure the servant leaked, permanently blocking re-subscription of that identity. The subscriber is now destroyed on failure. (This also corrects `Subscriber::destroy` to catch the current `ObjectAdapterDestroyedException`, which the cleanup relies on.)
- **#5854** — Oneway/datagram subscribers never reset their retry counter on a successful send (only the twoway completion path did), so retries accumulated over the subscription's lifetime and eventually removed healthy subscribers. The counter is now reset on a successful oneway send.

### Testing
`single`, `createOrRetrieve`, and `federation` remain green.

<sub>🤖 Produced by an automated correctness audit (Claude Fable 5) and cross-verified by Codex; reviewed but flagged for human triage.</sub>

PR #5939, opened 2026-07-10
Reset the oneway subscriber retry counter on a successful send

### LINKED ISSUE #5854: [Audit-Low] IceStorm oneway/datagram subscribers never reset the retry counter on success — retries are lifetime-cumulative, eventual spurious removal
### Summary
`_currentRetry` is meant to count *consecutive* failed retries and is reset to 0 on each successful delivery — but the reset lives only in `completed()`, which oneway/datagram subscribers never invoke. So a oneway subscriber's retry counter only ever increments across the whole subscription lifetime, and with `retryCount = N` the (N+1)-th transient failure *ever* removes it permanently, whereas an identical twoway subscriber survives indefinitely.

### Mechanism
The only reset is `_currentRetry = 0` in `completed()` ([`Subscriber.cpp:656`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Subscriber.cpp#L656)). Oneway/datagram pass a `nullptr` response callback ([`Subscriber.cpp:186-195`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Subscriber.cpp#L186-L195)), so `completed()` never runs, and their success paths (`:206-249`) don't touch the counter; it is only incremented in `error()` (`:777`). `retryCount` is parsed and honored for oneway (a soft failure such as ConnectFailed takes the retry/offline path via `error()`'s `_currentRetry < _retryCount` gate at `:743`).

### Trigger
A oneway/datagram subscriber with QoS `retryCount = N > 0` that accumulates N+1 separate transient failures over its lifetime, with any number of successful deliveries in between.

### Impact
Silent permanent removal of exactly the failure-tolerant oneway subscribers. (Note: issue #164 documents `retryCount` as twoway-oriented, so this is partly a semantics/doc question — but IceStorm currently accepts and enforces `retryCount` on oneway QoS.)

### Suggested fix
Reset `_currentRetry` on a successful oneway send (the sent/`sentAsynchronously` path), or document/reject `retryCount` on non-twoway subscribers.

---

pepone:
Oneway and datagram subscribers never reset their consecutive-retry counter on a successful send — only the twoway completion path did — so retries accumulated over the subscription's lifetime and eventually removed healthy subscribers. Reset the counter on a successful oneway send (synchronous delivery in `flush` and the async `sentAsynchronously` path), matching the twoway subscriber.

Fixes #5854

### DIFF
--- a/cpp/src/IceStorm/Subscriber.cpp
+++ b/cpp/src/IceStorm/Subscriber.cpp
@@ -207,9 +207,15 @@ SubscriberOneway::flush()
             {
                 ++_outstanding;
             }
-            else if (_observer)
+            else
             {
-                _observer->delivered(1);
+                // The event was delivered synchronously, so the subscriber is reachable: reset the
+                // consecutive-retry count as the twoway subscriber does on a successful reply.
+                _currentRetry = 0;
+                if (_observer)
+                {
+                    _observer->delivered(1);
+                }
             }
         }
         catch (const std::exception&)
@@ -233,6 +239,9 @@ SubscriberOneway::sentAsynchronously()
     // Decrement the _outstanding count.
     --_outstanding;
     assert(_outstanding >= 0 && _outstanding < _maxOutstanding);
+    // The event was delivered, so the subscriber is reachable: reset the consecutive-retry count as the
+    // twoway subscriber does on a successful reply.
+    _currentRetry = 0;
     if (_observer)
     {
         _observer->delivered(1);


bernardnormier (approved):
No changelog fragment needed: the retryCount QoS retry behavior was effectively undocumented (in 3.8 and earlier — see #5951), so there is no documented expectation for an entry to correct. #5951 tracks documenting the actual behavior, including the consecutive-failure semantics this PR establishes.

PR #5940, opened 2026-07-10
Mark an IceStorm topic destroyed only after its transaction commits

### LINKED ISSUE #5851: [Audit-Low] IceStorm TopicImpl::destroy latches _destroyed before the LMDB transaction — a failed txn leaves an undestroyable topic that still delivers
### Summary
`TopicImpl::destroy()` sets `_destroyed = true` before running the LMDB deletion transaction. If that transaction throws (e.g. `MDB_MAP_FULL`), all durable side effects are rolled back or skipped but the in-memory flag stays latched and is never reset — leaving a topic that can never be destroyed, never re-created under the same name, yet still delivers published events.

### Mechanism
`destroy()` sets the flag at [`TopicI.cpp:794`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TopicI.cpp#L794) then calls `destroyInternal` ([`TopicI.cpp:805`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TopicI.cpp#L805)). `destroyInternal` performs the LMDB work first and rethrows on `LMDBException` ([`TopicI.cpp:1216-1221`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TopicI.cpp#L1216-L1221)); the reaper-add, servant removals, and per-subscriber destroy all happen only after `txn.commit()` and are skipped on failure. Nothing clears `_destroyed`.

### Impact
Retry `destroy()` → `ObjectNotExistException`; `create(name)` → `TopicExists` forever (the name was never added to the reaper, so it lingers in `_topics`); subscribe/link/unlink → ONE via the #5655 guards. But `publish` has no `_destroyed` check and keeps delivering to the intact `_subscribers`. In-memory state contradicts the DB/adapter/replicas. Bounded: a restart reloads the intact DB cleanly, and in replicated mode `observerInit` reinstalls the topic on resync — so it persists only until restart/re-election.

### Trigger
`destroy()` while the LMDB write fails (deletes still allocate copy-on-write pages, so `MDB_MAP_FULL` is reachable).

### Suggested fix
Set `_destroyed = true` only after the transaction commits (or reset it on failure before rethrowing).

---

pepone:
`TopicImpl::destroy` marked the topic destroyed *before* its database transaction committed, so a failed commit left an undestroyable topic that still delivered events. `_destroyed` is now set only after `destroyInternal` commits, so a failed commit leaves the topic intact and still destroyable.

Fixes #5851

### DIFF
--- a/cpp/src/IceStorm/TopicI.cpp
+++ b/cpp/src/IceStorm/TopicI.cpp
@@ -797,7 +797,6 @@ TopicImpl::destroy()
     {
         throw Ice::ObjectNotExistException{__FILE__, __LINE__};
     }
-    _destroyed = true;
 
     auto traceLevels = _instance->traceLevels();
     if (traceLevels->topic > 0)
@@ -806,9 +805,12 @@ TopicImpl::destroy()
         out << _name << ": destroy";
     }
 
-    // destroyInternal clears out the topic content.
+    // destroyInternal clears out the topic content. Mark the topic destroyed only once its database
+    // transaction has committed, so a failed commit leaves the topic intact and still destroyable.
     LogUpdate llu = {0, 0};
-    _instance->observers()->destroyTopic(destroyInternal(llu, true), _name);
+    LogUpdate destroyLLU = destroyInternal(llu, true);
+    _destroyed = true;
+    _instance->observers()->destroyTopic(destroyLLU, _name);
 
     _observer.detach();
 }


bernardnormier (approved):
No changelog entry needed here: an LMDB write failure such as MDB_MAP_FULL leaves the service degraded until a restart regardless, and the restart also clears the latched in-memory state since the failed transaction never committed. The fix is still worth merging — destroy now reports the failure instead of silently latching a half-destroyed topic — but the user-observable improvement is too small for an entry.

Issue #5938, opened 2026-07-10
[Audit-Low, grouped] Swift: 4 low-severity findings from the API doc audit


pepone:
Four small defects found while verifying the Swift API docs against the implementation, grouped per audit convention.

### 1. `ObjectAdapter.setLocator` silently swallows errors under a phantom `throws`
The protocol declares `func setLocator(_ loc: LocatorPrx?) throws` ([`ObjectAdapter.swift:259`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/swift/src/Ice/ObjectAdapter.swift#L259)), but the bridge catches `ObjectAdapterDeactivatedException`, `CommunicatorDestroyedException`, and `std::exception` and ignores all of them ([`ObjectAdapter.mm:143-166`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/swift/src/IceImpl/ObjectAdapter.mm#L143-L166)) — the method can never throw, and `setLocator` on a destroyed adapter silently succeeds where C++ throws. Either propagate the errors or drop the `throws`.

### 2. `InputStream.readEncapsulation` never advances the stream
It reads the size and encoding, rewinds 6 bytes, and returns a `Data` slice without advancing `pos` ([`InputStream.swift:54-68`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/swift/src/Ice/InputStream.swift#L54-L68)) — a second call returns the same bytes, and the returned value is a shared-storage slice with non-zero-based indices. C++ consumes the encapsulation ([`InputStream.cpp:336-353`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/cpp/src/Ice/InputStream.cpp#L336-L353)). Latent today (current callers read once), but a behavioral divergence for streaming-API users.

### 3. `CtrlCHandler` is constructible on iOS, where it silently does nothing
[...]

Issue #5936, opened 2026-07-10
[Audit-Medium] Swift: SSL peerCertificate crosses the bridge as an invalid non-optional when the peer presents no certificate


pepone:
### Summary
`SSLConnectionInfo.peerCertificate` is declared `SecCertificate?`, but the nil case is unreachable through the bridge: when the peer presents no certificate, the C++ side produces a NULL `SecCertificateRef` that is bridged into a **non-optional** Swift `SecCertificate` — an invalid object — instead of `nil`. Same defect class as the iAP nil-safety fix (#5800), which covered iAP only.

### Mechanism
The SecureTransport transceiver creates the info with `peerCertificate = nullptr` when there is no peer certificate ([`SecureTransportTransceiverI.cpp:493-503`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/cpp/src/Ice/SSL/SecureTransportTransceiverI.cpp#L493-L503)). The bridge forwards it unconditionally ([`Connection.mm:232-236`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/swift/src/IceImpl/Connection.mm#L232-L236)) through a parameter that sits inside the header's `NS_ASSUME_NONNULL` region and is therefore `_Nonnull` ([`include/Connection.h:8, 56`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/swift/src/IceImpl/include/Connection.h#L56)) — the bridge violates its own declared contract. The Swift factory accordingly imports the parameter **non-optional** ([`ConnectionInfoFactory.swift:61-64`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/swift/src/Ice/ConnectionInfoFactory.swift#L61-L64)), so a NULL CF reference becomes a non-optional Swift object before being stored into the optional property ([`Connection.swift:239`](https://github.com/zeroc-ice/ice/blob/4a1ed88e9fc27c82d7871879d0187733e69df0d3/swift/src/Ice/Connection.swift#L239)).

### Trigger
On macOS/iOS, a server (or a client with an SSL server that doesn't request client certificates) calls `connection.getInfo()` on an SSL connection whose peer presented no certificate; using the resulting `peerCertificate` operates on an invalid object.

### Suggested fix
[...]

Issue #5894, opened 2026-07-08
IcePy should map non-Ice C++ exceptions to native Python error types instead of plain LocalException


pepone:
### Summary
IcePy's exception conversion has a catch-all that turns any non-Ice C++ exception into an instance of the plain base `Ice.LocalException`:

```cpp
catch (const std::exception& ex)
{
    std::array args{IcePy::createString(ex.what())};
    return createPythonException(localExceptionTypeId, args); // "::Ice::LocalException"
}
```
([`Util.cpp` `convertException`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/python/modules/IcePy/Util.cpp#L653-L657))

As a result, argument-validation failures that every other mapping reports with the language's native error type surface in Python as the Ice base class — the only place where Ice raises a non-derived `LocalException`:

- `Ice.stringToIdentity("")` / `identityToString` with an empty name → C++ `checkIdentity` throws `std::invalid_argument` → Python `Ice.LocalException` (C++ raises `std::invalid_argument`, C# `ArgumentException`, JS `TypeError`).
- `Connection.setAdapter` on an incoming connection → C++ `std::logic_error` → Python `Ice.LocalException` (C# `InvalidOperationException`, JS `Error`).

Separately, Ice local exceptions with no Python counterpart (`PluginInitializationException`, `FileException`) degrade to the base class through the `fallbackToLocalException` path.

### Proposal (3.9)
- Map `std::invalid_argument` → `ValueError` and `std::logic_error` → `RuntimeError` in `convertException`, matching the native-error convention of the other mappings.
- Add the missing `PluginInitializationException`/`FileException` Python classes (or keep the base-class fallback only for genuinely unknown Ice local exceptions).

Changing raised exception types is backward-incompatible, so this is 3.9.0 material. The Python API docs updated in #5893 document the current 3.8 behavior (`Raises: LocalException`) and should be updated together with this change.

PR #5941, opened 2026-07-10
Fix incorrect and stale API doc comments in the Swift Ice library

pepone:
Fixes incorrect and stale DocC comments in the Swift Ice library, found during the Swift API doc review (follow-up to #5867/#5868/#5883/#5893). Documentation-only — no declaration or code changes; the runtime bugs found by the same review are tracked in #5936-#5938 and deliberately not fixed here.

Incorrect claims corrected (many are the same stale pre-#5867 patterns fixed in the other mappings):

- `Communicator.shutdownCompleted()` — "returns immediately when already shut down" replaced with the corrected resumption-threading wording; `ServantLocator.deactivate` is called on adapter *destroy*; WS `headers` are the *response* headers on outgoing connections; `Connection.throwException` also throws while *closing*; `setAdapter` documents the `LocalException` on incoming connections; `SSLConnectionInfo.peerCertificate` is a single peer certificate (the `nil` wording documents the field's contract — the currently-impossible nil crossing is bug #5936); `ToStringMode.Compat` escapes only *non-printable* ASCII; `OptionalFormat.FSize` "prior to marshaling"; `endEncapsulation` ends the *current* encapsulation (both streams); `writePendingValues`/`readPendingValues` describe the 1.0-only deferral and reference the real Swift APIs (the old text pointed at a nonexistent `writeValue()` and a Javadoc `{@link #readValue}`); `ice_getSlicedData` drops the stale "preserved-slice base class"; `SliceInfo.bytes` excludes the slice header; `IncomingRequest` is positioned at the in-parameters encapsulation (the old sentence was a C# copy-paste); `OutgoingResponse.exceptionId` is the *name* of the error's type, not a "full name"; `intVersion`/`stringVersion` document the pre-release forms.
- `AlreadyRegisteredException` was triply wrong: the class doc was physically truncated by a `//` line (DocC rendered "…more than once for the"), `kindOfObject` said "could not be **removed**", and the kind list included "object"/"replica group" which nothing produces — now the 7 actual kinds.
- Trap/terminate realities documented per the #5894 pattern: `getIceProperty`/`getIcePropertyAsList` no longer carry `- Throws:` tags on non-throwing functions — a note states that an unknown Ice property terminates the program (the bridge deliberately doesn't catch); `setProperty` documents the same for unknown reserved-prefix properties; preconditions documented on `identityToString`, `Connection.createProxy`, `proxyToProperty` (fixed proxies), `DispatchException.init`, and `CtrlCHandler.init` (second instance terminates).
- Missing/vague docs: the proxy `- Returns: The result of the invocation.` batch now states the real semantics (`ice_isA` true-if-implements, `ice_ids` alphabetical, `ice_invoke`'s `ok`/`outEncaps` tuple incl. oneway/batch, `ice_getConnection` nil for collocation); `ice_getInvocationTimeout` "(in milliseconds)"; `uncheckedCast` purely-local remark; `==`/`!=` documented for proxies and endpoints; `stringToProxy`/`propertyToProxy`/`stringToIdentity`/`addWithUUID`/`addFacetWithUUID`/`initializePlugins` Throws named; `Encoding_1_0`/`Encoding_1_1` documented; `Endpoint.getInfo()`/`endValue()` nil cases; `readEncapsulation` documents that the position is not advanced and the 6-byte header is included (today's behavior; a change is tracked in #5938); single-encapsulation restriction documented on both `startEncapsulation` overloads; `ImplicitContext.put`/`remove` empty-string convention; `Properties` thread-safety remark; `createProperties` Ice.Config/ICE_CONFIG loading conditions; `close()` rethrow note.
- Markup/consistency: stray third backtick breaking the ``ServantLocator`` link, malformed `- Parameter:` on the properties-update callback typealias, `CloseCallback` parameter prose, "Ice run time"→"runtime", imperative "Read an…"→"Reads an…", double spaces, redundant middleware sentence.

Deliberately excluded: the `addDefaultServant`/`find`/`findFacet` lookup-order docs (correct once #5915 merges), `setLocator` docs (#5938 owns the impl decision), and the `getPropertyAsList` mismatched-quotes sentences (pending the cross-mapping code-vs-doc decision). One codex-audit flag declined: `ice_ping`'s "tests whether the target object can be reached" is the corrected C++ reference wording — the oneway/batch nuance applies identically in C++ and should be refined there first if desired.

Verified with the CI-identical `swift format` check (in-place format produces no diff) and `swift format lint --strict`: clean. All changes were codex-audited against the Swift implementation, the ObjC++ bridge, and the C++ core; the three valid flags (createProperties ICE_CONFIG nuance, readPendingValues 1.1 qualification, sequential-encapsulation precondition) are incorporated.

### DIFF
--- a/swift/src/Ice/Connection.swift
+++ b/swift/src/Ice/Connection.swift
@@ -20,7 +20,7 @@ public enum CompressBatch: UInt8 {
 
 /// An `Ice.InputStream` extension to read `CompressBatch` enumerated values from the stream.
 extension InputStream {
-    /// Read an enumerated value.
+    /// Reads an enumerated value.
     ///
     /// - Returns: The enumerated value.
     public func read() throws -> CompressBatch {
@@ -31,7 +31,7 @@ extension InputStream {
         return val
     }
 
-    /// Read an optional enumerated value from the stream.
+    /// Reads an optional enumerated value from the stream.
     ///
     /// - Parameter tag: The numeric tag associated with the value.
     /// - Returns: The enumerated value.
@@ -69,9 +69,8 @@ extension OutputStream {
 public typealias HeaderDict = [String: String]
 
 /// The callback function given to ``Connection/setCloseCallback(_:)``.
-/// This callback is called by the connection when the connection is closed.
-///
-/// - Parameter _: The connection that was closed. It's never `nil`.
+/// This callback is called by the connection when the connection is closed. Its argument is the connection that
+/// was closed; it's never `nil`.
 public typealias CloseCallback = (Connection?) -> Void
 
 /// Represents a connection that uses the Ice protocol.
@@ -81,12 +80,14 @@ public protocol Connection: AnyObject, CustomStringConvertible, Sendable {
 
     /// Closes this connection gracefully once all outstanding invocations have completed. If closing the connection
     /// takes longer than the configured close timeout, the connection is aborted with a ``CloseTimeoutException``.
+    /// If the connection was lost or aborted, awaiting the result rethrows the exception that caused the closure.
     func close() async throws
 
     /// Creates a special proxy (a "fixed proxy") that always uses this connection.
     ///
     /// - Parameter id: The identity of the target object.
     /// - Returns: A fixed proxy with the provided identity.
+    /// - Precondition: `id.name` must not be empty.
     func createProxy(_ id: Identity) throws -> ObjectPrx
 
     /// Associates an object adapter with this connection. When a connection receives a request, it dispatches this
@@ -96,6 +97,8 @@ public protocol Connection: AnyObject, CustomStringConvertible, Sendable {
     /// the default object adapter of an outgoing connection is the communicator's default object adapter.
     ///
     /// - Parameter adapter: The object adapter to associate with this connection.
+    /// - Throws: A ``LocalException`` when this connection is an incoming connection: you can only call this method
+    ///   on outgoing (client) connections.
     func setAdapter(_ adapter: ObjectAdapter?) throws
 
     /// Gets the object adapter associated with this connection.
@@ -151,7 +154,7 @@ public protocol Connection: AnyObject, CustomStringConvertible, Sendable {
     /// Throws an exception that provides the reason for the closure of this connection. For example,
     /// this method throws ``CloseConnectionException`` when the connection was closed gracefully by the peer;
     /// it throws ``ConnectionAbortedException`` when the connection is aborted with ``abort()``.
-    /// This method does nothing if the connection is not yet closed.
+    /// This method does nothing if the connection is not yet closing or closed.
     func throwException() throws
 }
 
@@ -235,7 +238,7 @@ public final class TCPConnectionInfo: IPConnectionInfo {
 
 /// Provides access to the connection details of an SSL connection.
 public final class SSLConnectionInfo: ConnectionInfo {
-    /// The certificate chain.
+    /// The peer certificate, or `nil` if the peer did not provide one.
     public let peerCertificate: SecCertificate?
 
     internal init(underlying: ConnectionInfo, peerCertificate: SecCertificate?) {
@@ -277,7 +280,8 @@ public final class UDPConnectionInfo: IPConnectionInfo {
 
 /// Provides access to the connection details of a WebSocket connection.
 public final class WSConnectionInfo: ConnectionInfo {
-    /// The headers from the HTTP upgrade request.
+    /// The HTTP headers from the WebSocket upgrade handshake: the request headers for an incoming connection,
+    /// and the response headers for an outgoing connection.
     public let headers: HeaderDict
 
     internal init(underlying: ConnectionInfo, headers: HeaderDict) {
--- a/swift/src/Ice/Connection.swift
+++ b/swift/src/Ice/Connection.swift
@@ -151,10 +151,11 @@ public protocol Connection: AnyObject, CustomStringConvertible, Sendable {
     ///   - sndSize: The size of the send buffer.
     func setBufferSize(rcvSize: Int32, sndSize: Int32) throws
 
-    /// Throws an exception that provides the reason for the closure of this connection. For example,
-    /// this method throws ``CloseConnectionException`` when the connection was closed gracefully by the peer;
-    /// it throws ``ConnectionAbortedException`` when the connection is aborted with ``abort()``.
+    /// Throws an exception that provides the reason for the closure of this connection.
     /// This method does nothing if the connection is not yet closing or closed.
+    ///
+    /// - Throws: ``CloseConnectionException`` when the connection was closed gracefully by the peer;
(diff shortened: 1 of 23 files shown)


swift/src/Ice/Connection.swift
```diff
@@ -69,9 +69,8 @@ extension OutputStream {
 public typealias HeaderDict = [String: String]
 
 /// The callback function given to ``Connection/setCloseCallback(_:)``.
-/// This callback is called by the connection when the connection is closed.
-///
-/// - Parameter _: The connection that was closed. It's never `nil`.
+/// This callback is called by the connection when the connection is closed. Its argument is the connection that
+/// was closed; it's never `nil`.
 public typealias CloseCallback = (Connection?) -> Void
```
pepone, line 74: Why do we use `?` if is never nill?

swift/src/Ice/Connection.swift
```diff
@@ -151,7 +154,7 @@ public protocol Connection: AnyObject, CustomStringConvertible, Sendable {
     /// Throws an exception that provides the reason for the closure of this connection. For example,
     /// this method throws ``CloseConnectionException`` when the connection was closed gracefully by the peer;
     /// it throws ``ConnectionAbortedException`` when the connection is aborted with ``abort()``.
```
pepone: We should use `Throws:` for the exceptions like we do in C#/C++

swift/src/Ice/LocalExceptions.swift
```diff
@@ -16,10 +16,11 @@ public class DispatchException: LocalException, @unchecked Sendable {
     /// Creates a DispatchException.
     ///
     /// - Parameters:
-    ///   - replyStatus: The reply status raw value. It may not correspond to a valid ``ReplyStatus`` enum value.
+    ///   - replyStatus: The reply status raw value.
     ///   - message: The exception message.
     ///   - file: The file where the exception was thrown.
     ///   - line: The line where the exception was thrown.
+    /// - Precondition: `replyStatus` must be greater than `ReplyStatus.userException.rawValue`.
```
pepone:
```suggestion
    /// - Precondition: `replyStatus` must be greater than `ReplyStatus.userException`. ?
```

swift/src/Ice/Connection.swift
bernardnormier: No reason — the runtime always passes a non-nil connection (`cb(self)` in `ConnectionI.setCloseCallback`), so the optionality is a port artifact. Changing the typealias to `(Connection) -> Void` is source-breaking and, unlike C# (where nullability annotations are erased at the IL level), also binary-incompatible in Swift: `Optional` is a real type that participates in symbol mangling, the protocol witness table, and runtime type metadata. So the doc note is the right move for 3.8; #5953 tracks the API cleanup for 3.9.

bernardnormier (commented):
Two points from review; approving after the next push.

- Copilot's ICE_CONFIG comment is correct: `Properties::loadConfig` reads the environment variable when `Ice.Config` is unset, empty, or set to `"1"` (what a bare `--Ice.Config` becomes), so the new wording in both `createProperties` overloads ("when no `Ice.Config` property is set at all") is incomplete.
- The "Non-printable" qualification added to `ToStringMode.Compat` is behaviorally correct, but C++, C# and Java still share the old unqualified sentence — please add it to the cross-mapping follow-up list so the four mappings end up saying the same thing.

The rest checks out: I verified the `SliceInfo.bytes` capture window against the decoder, the `setAdapter` behavior, and the `ServantLocator.deactivate`-on-destroy claim, among others.

swift/src/Ice/Initialize.swift
```diff
@@ -115,7 +115,9 @@ public func createProperties() -> Properties {
     return PropertiesI(handle: ICEUtil.createProperties())
 }
 
-/// Creates a property set initialized from an argument array.
+/// Creates a property set initialized from an argument array. This function loads configuration files named by the
+/// `Ice.Config` property set through `args` or, when no `Ice.Config` property is set at all, by the `ICE_CONFIG`
+/// environment variable.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Confirmed against `Properties::loadConfig()` — the env var is also consulted when `Ice.Config` is empty or `1` (what a bare `--Ice.Config` becomes). Reworded to "not set, is empty, or is set to `1`". Done in 4996ec81fe.

swift/src/Ice/Initialize.swift
```diff
@@ -137,7 +139,9 @@ public func createProperties(_ args: [String], defaults: Properties? = nil) thro
     }
 }
 
-/// Creates a property set initialized from an argument array.
+/// Creates a property set initialized from an argument array. This function loads configuration files named by the
+/// `Ice.Config` property set through `args` or, when no `Ice.Config` property is set at all, by the `ICE_CONFIG`
+/// environment variable.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Same fix as the other overload — both `createProperties` docs now cover unset/empty/`1`. Done in 4996ec81fe.

swift/src/Ice/Connection.swift
pepone: 🤖 **Claude:** Rewrote as a `- Throws:` callout listing ``CloseConnectionException`` and ``ConnectionAbortedException``, mirroring the C++ `@throws` / C# `<exception>` tags. Done in 4996ec81fe.

swift/src/Ice/LocalExceptions.swift
pepone: 🤖 **Claude:** Applied — C++ and C# both say "greater than `ReplyStatus::UserException`" with no raw-value mention, and the runtime precondition message matches. Done in 4996ec81fe.

swift/src/Ice/Connection.swift
pepone: 🤖 **Claude:** No code change — the "never `nil`" note documents the 3.8 reality, and #5953 tracks the API-shape cleanup for 3.9 (dropping the `Optional` is source- and ABI-breaking in Swift).

pepone:
🤖 **Claude:** Both review points addressed:

- **ICE_CONFIG**: confirmed against `Properties::loadConfig()` and fixed in 4996ec81fe — both `createProperties` overloads now say the env var is used when `Ice.Config` is not set, is empty, or is set to `1` (the value a bare `--Ice.Config` becomes).
- **ToStringMode.Compat cross-mapping**: filed #5955. Note the current state is narrower than "C++, C# and Java share the old sentence": C++ (`StringUtil.h`) and C# (`Communicator.cs`) already carry "Non-printable" on all three enumerators from #5867/#5868. The gaps are the `Compat` entry only — the same one-word omission — in Java, JS, Python, and MATLAB; `Unicode`/`ASCII` are already correct everywhere.

The same push also applies the two inline suggestions from Jose (`- Throws:` tags on `throwException`, precondition wording without `.rawValue`).

bernardnormier (approved):
Looks good!

PR #5943, opened 2026-07-10
Destroy the IceStorm subscriber when its registration transaction fails

### LINKED ISSUE #5853: [Audit-Low] IceStorm: an LMDB failure after Subscriber::create orphans the per-subscriber servant — resubscribe blocked and replica sync can wedge
### Summary
`Subscriber::create` registers the per-subscriber publisher servant **before** the caller's LMDB transaction commits. If that transaction throws, nobody calls `Subscriber::destroy`, so the servant (under a deterministic identity) leaks — every later subscribe for that identity throws `AlreadyRegisteredException`, and on a replica every subsequent full sync fails the same way.

### Mechanism
`Subscriber::create` adds the servant at [`Subscriber.cpp:489-494`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Subscriber.cpp#L489-L494). The call sites do this before the txn and only rethrow on failure: `subscribeAndGetPublisher` ([`TopicI.cpp:543-563`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TopicI.cpp#L543-L563)) and `observerAddSubscriber` (`:1048-1069`). The subscriber is never pushed to `_subscribers` and `Subscriber::destroy` (the only servant-removal path, `:613-637`) never runs. `unsubscribe`/`removeSubscribers` can't clean it up either — the DB `del` finds nothing, so `found=false` and the cleanup is skipped ([`TopicI.cpp:1252-1301`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TopicI.cpp#L1252-L1301)).

### Trigger
One transient LMDB write failure during subscribe; then the client retries after the condition clears.

### Impact
Every retry throws `AlreadyRegistered → UnknownException` until restart. On a replica the same leak in `observerAddSubscriber` makes each later full sync fail (`observerInit → update → create → AlreadyRegistered`) — a recovery loop until restart.

### Suggested fix
On transaction failure, call `subscriber->destroy()` before rethrowing (or register the per-subscriber servant only after the commit).

---

pepone:
`subscribeAndGetPublisher` and `observerAddSubscriber` registered the per-subscriber publisher servant *before* the database transaction; on a transaction failure the servant leaked, permanently blocking re-subscription of that identity. The subscriber is now destroyed on failure.

This also corrects `Subscriber::destroy` to catch the current `ObjectAdapterDestroyedException` (its catch clause had been stale since the `Deactivated`/`Destroyed` split), which the cleanup relies on to stay silent on shutdown.

Fixes #5853

### DIFF
--- a/cpp/src/IceStorm/Subscriber.cpp
+++ b/cpp/src/IceStorm/Subscriber.cpp
@@ -626,9 +626,9 @@ Subscriber::destroy()
         {
             // Ignore
         }
-        catch (const Ice::ObjectAdapterDeactivatedException&)
+        catch (const Ice::ObjectAdapterDestroyedException&)
         {
-            // Ignore
+            // Ignore -- this can occur on shutdown.
         }
     }
 
--- a/cpp/src/IceStorm/TopicI.cpp
+++ b/cpp/src/IceStorm/TopicI.cpp
@@ -564,6 +564,9 @@ TopicImpl::subscribeAndGetPublisher(QoS qos, Ice::ObjectPrx obj)
     catch (const IceDB::LMDBException& ex)
     {
         logError(_instance->communicator(), ex);
+        // The subscriber was created (and its publisher servant registered) before the transaction; remove
+        // it so a retry with the same subscriber identity is not rejected with AlreadyRegisteredException.
+        subscriber->destroy();
         throw; // will become UnknownException in caller
     }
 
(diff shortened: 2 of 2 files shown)


bernardnormier (approved):
No changelog entry needed: the only trigger is an LMDB write failure, which leaves the service degraded until a restart regardless — and the restart also clears the leaked servant. The fix is still worth merging (the failed call now cleans up after itself, and the stale catch clause in Subscriber::destroy is a real fix for shutdown-time cleanup), but the user-observable improvement is too small for an entry.

PR #3102, opened 2024-11-07
Simplify InputStream constructor in JS

bernardnormier:
This is the JS equivalent of the previous InputStream simplification PRs.

PR #5945, opened 2026-07-10
Fix JS InputStream constructor handling of ArrayBuffer and Uint8Array view arguments

### LINKED ISSUE #5877: [Audit-Medium] JS: InputStream constructor mishandles ArrayBuffer arguments and Uint8Array views
### Summary
Two defects in the `InputStream` buffer-argument handling make the public constructor silently produce wrong streams: an `ArrayBuffer` argument yields an *empty* stream, and a `Uint8Array` that is a view into a larger buffer unmarshals the wrong bytes.

### Mechanism
1. The `ArrayBuffer` branch executes `this._buf = new Buffer(arg.bytes)` ([`InputStream.js:911-916`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/InputStream.js#L911-L916)); `ArrayBuffer` has no `bytes` property, so `Buffer` receives `undefined` and constructs an empty buffer (`b = null`, `_limit = 0`, [`Buffer.js:7-20`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/Buffer.js#L7-L20)). It should be `new Buffer(arg)`.
2. The `Uint8Array` branch executes `new Buffer(arg.buffer)` ([`InputStream.js:917-922`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/InputStream.js#L917-L922)), wrapping the *entire* underlying `ArrayBuffer`: the view's `byteOffset` and `byteLength` are discarded (`Buffer` starts at position 0 with `_limit = byteLength` of the whole ArrayBuffer). Only a view spanning its complete ArrayBuffer works; a `subarray()`/offset view — e.g. a slice of a larger network buffer — decodes from offset 0 of the wrong region.

The declaration advertises plain `buffer: Uint8Array` parameters ([`InputStream.d.ts:15-24`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/InputStream.d.ts#L15-L24)) with no whole-buffer restriction.

### Trigger
```js
const data = new Uint8Array(bigBuffer.buffer, 100, encapsSize); // view into a larger buffer
new Ice.InputStream(communicator, data);                        // reads from offset 0, not 100
new Ice.InputStream(communicator, bigBuffer.buffer);            // empty stream
```

### Impact
Streaming-API users deserializing from a view or an ArrayBuffer get `MarshalException`s or silently wrong data.

### Suggested fix
Use `new Buffer(arg)` for the ArrayBuffer branch, and honor `byteOffset`/`byteLength` for Uint8Array views (e.g. wrap with a DataView over the exact range, or copy the view).

### Notes
Confidence: high. Found in the 2026-07-08 JavaScript API doc audit (finding A.4) while verifying the constructor docs.

---

pepone:
Fixes #5877.

- An `ArrayBuffer` argument was passed to `Buffer` as `arg.bytes` (undefined), producing an empty stream; the stream now wraps the `ArrayBuffer` itself (regression from #3102).
- A `Uint8Array` view was unwrapped to its entire underlying `ArrayBuffer`, ignoring `byteOffset`/`byteLength`; the stream now reads the view's exact byte range (zero-copy when the view spans its whole buffer, copying the range otherwise).

The `InputStream` TypeScript constructor overloads now accept `Uint8Array | ArrayBuffer`.

Note: the constructor's exact-constructor checks still reject `Uint8Array` subclasses such as Node's `Buffer`, although the TypeScript signature structurally admits them; that pre-existing limitation is unchanged.

### DIFF
--- a/changelog.d/js/5877.md
+++ b/changelog.d/js/5877.md
@@ -1,0 +1,3 @@
+- Fixed the `InputStream` constructor's handling of buffer arguments. An `ArrayBuffer` argument previously produced
+  an empty stream, and a `Uint8Array` view that does not span its entire underlying buffer was read from the wrong
+  byte range.
--- a/js/packages/ice/src/Ice/InputStream.d.ts
+++ b/js/packages/ice/src/Ice/InputStream.d.ts
@@ -10,18 +10,20 @@ declare module "@zeroc/ice" {
              * Constructs an InputStream using a communicator and this communicator's default encoding version.
              *
              * @param communicator The communicator to use for unmarshaling tasks.
-             * @param buffer The encoded data.
+             * @param buffer The encoded data. When a `Uint8Array` is supplied, the stream reads the exact byte
+             * range of the view.
              */
-            constructor(communicator: Communicator, buffer: Uint8Array);
+            constructor(communicator: Communicator, buffer: Uint8Array | ArrayBuffer);
 
             /**
              * Constructs an InputStream using a communicator and encoding version.
              *
              * @param communicator The communicator to use for unmarshaling tasks.
              * @param encoding The encoding version used to encode the data to be unmarshaled.
-             * @param buffer The encoded data.
+             * @param buffer The encoded data. When a `Uint8Array` is supplied, the stream reads the exact byte
+             * range of the view.
              */
-            constructor(communicator: Communicator, encoding: EncodingVersion, buffer: Uint8Array);
+            constructor(communicator: Communicator, encoding: EncodingVersion, buffer: Uint8Array | ArrayBuffer);
 
             /**
              *  Releases any data retained by encapsulations.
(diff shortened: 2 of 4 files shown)


bernardnormier (approved):
Checked the 3.7 branch (js/src/Ice/Stream.js) for both bugs:

- **ArrayBuffer → empty stream: not present in 3.7.** The 3.7 constructor stores the ArrayBuffer itself (`args.bytes = arg`) and wraps it correctly — confirming this is a 3.8 regression from the #3102 rewrite.
- **Uint8Array view range: present in 3.7.** The 3.7 code does `args.bytes = arg.buffer`, ignoring `byteOffset`/`byteLength`, so a subarray view has always read the wrong byte range in every 3.x JS release. Long-standing bug rather than a regression; probably not worth a 3.7 backport.

PR #5946, opened 2026-07-10
Use the logger supplied in InitializationData over Ice.LogFile in JS

### LINKED ISSUE #5878: [Audit-Low] JS: Ice.LogFile overrides the logger supplied in InitializationData
### Summary
When `Ice.LogFile` is set, the JS runtime unconditionally replaces the logger the application supplied in `InitializationData` with a `FileLogger`. In C++ the property is only consulted when no logger was supplied — a user-provided logger always wins.

### Mechanism
[`InstanceExtensions.js:181-192`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/js/packages/ice/src/Ice/InstanceExtensions.js#L181-L192) — the `Ice.LogFile` check assigns `this._initData.logger = new FileLogger(...)` before the `logger === null` fallback, with no guard on an existing logger. C++ guards the whole property-logger block with `if (!_initData.logger)` ([`Instance.cpp:1045-1084`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Ice/Instance.cpp#L1045-L1084)).

### Trigger
Create a communicator with both `initData.logger = myLogger` and `--Ice.LogFile=app.log`: log output silently goes to the file, and `myLogger` is never used.

### Impact
Configuration-dependent loss of application log routing; diverges from the other mappings. Related, undocumented: in browsers `Ice.LogFile` throws `InitializationException` ("not supported in Web Browsers").

### Suggested fix
Only consult `Ice.LogFile` when `this._initData.logger === null`, matching C++.

### Notes
Confidence: high. Found in the 2026-07-08 JavaScript API doc audit (finding A.3).

---

pepone:
Fixes #5878.

`Ice.LogFile` unconditionally replaced the logger supplied in `InitializationData` with a `FileLogger`. The property is now only consulted when no logger was supplied, matching C++ (a user-provided logger always wins).

As in C++, this also means a communicator with a user-supplied logger no longer throws `InitializationException` in browsers when `Ice.LogFile` is set, since the property is never consulted.

### DIFF
--- a/js/packages/ice/src/Ice/InstanceExtensions.js
+++ b/js/packages/ice/src/Ice/InstanceExtensions.js
@@ -179,12 +179,15 @@ Instance.prototype.finishSetup = function (communicator) {
         );
 
         const programName = this._initData.properties.getIceProperty("Ice.ProgramName");
-        const logFile = this._initData.properties.getIceProperty("Ice.LogFile");
-        if (logFile.length > 0) {
-            if (FileLogger === null) {
-                throw new InitializationException("Ice.LogFile property is not supported in Web Browsers");
+        // Ice.LogFile is only consulted when the application didn't supply its own logger.
+        if (this._initData.logger === null) {
+            const logFile = this._initData.properties.getIceProperty("Ice.LogFile");
+            if (logFile.length > 0) {
+                if (FileLogger === null) {
+                    throw new InitializationException("Ice.LogFile property is not supported in Web Browsers");
+                }
+                this._initData.logger = new FileLogger(programName, logFile);
             }
-            this._initData.logger = new FileLogger(programName, logFile);
         }
 
         if (this._initData.logger === null) {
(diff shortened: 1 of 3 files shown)


js/packages/ice/src/Ice/InstanceExtensions.js
```diff
@@ -179,12 +179,15 @@ Instance.prototype.finishSetup = function (communicator) {
         );
 
         const programName = this._initData.properties.getIceProperty("Ice.ProgramName");
-        const logFile = this._initData.properties.getIceProperty("Ice.LogFile");
-        if (logFile.length > 0) {
-            if (FileLogger === null) {
-                throw new InitializationException("Ice.LogFile property is not supported in Web Browsers");
+        // Ice.LogFile is only consulted when the application didn't supply its own logger.
```
bernardnormier:
```suggestion
        // Ice.LogFile is only consulted when the application doesn't supply its own logger.
```

bernardnormier (approved):
Please also add a changelog fragment (changelog.d/js/5878.md) — the trigger is routine and the symptom user-visible. Suggested wording:

```md
- Fixed the communicator to use the logger supplied in `InitializationData` even when `Ice.LogFile` is set. The
  property previously replaced the supplied logger with a file logger, and in browsers caused communicator
  initialization to fail.
```

PR #5970, opened 2026-07-14
Fix icestormadmin error reporting when no topic manager is reachable

### LINKED ISSUE #5962: [Audit-Low] icestormadmin without configuration dies with a raw endpoint ParseException
With no topic manager configured, the Finder fallback builds `IceStorm/Finder:tcp -p <empty>…` ([`Admin.cpp:154-174`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Admin.cpp#L154-L174)), which throws `ParseException` at proxy construction, outside the try/catch that prints the intended "no manager proxies configured" message — so `icestormadmin` run without configuration dies with a raw endpoint parse error instead of a helpful diagnostic.

**Fix:** validate the port / move the proxy construction inside the guard. Cosmetic.

---
Split out of #5858 (item E4) for individual milestone triage.

*Filed from an automated AI correctness audit of `cpp/src/IceStorm` (Claude Fable 5, 2026-07-07); confirmed through a multi-agent adversarial refute pass over the actual source.*

pepone:
Without any configuration, `icestormadmin` built the `IceStorm/Finder` fallback proxy with an empty `IceStormAdmin.Port` (the property has no default) and constructed it outside the try block, so the tool died with a raw endpoint parse error — `no argument provided for -p option in endpoint ''tcp  -p -t 3000''` — instead of the intended `no manager proxies configured` message.

The tool now distinguishes the failure modes instead of reporting them all as "no manager proxies configured": that message is reserved for a truly unconfigured run, while a partially configured or unreachable Finder gets a specific error (the unreachable case follows `icegridadmin`'s "could not contact the default locator" pattern). The exit code is unchanged (1) in all cases.

Verified against a local build:

| Invocation | Before | After |
|---|---|---|
| no configuration | `no argument provided for -p option in endpoint ''tcp  -p -t 3000''` | `no manager proxies configured` |
| `--IceStormAdmin.Host=localhost` only | same parse error (empty port) | `IceStormAdmin.Host is set but IceStormAdmin.Port is not` |
| `--IceStormAdmin.Port=abc` | `invalid port value 'abc' in endpoint ''tcp  -p abc -t 3000''` | `invalid IceStormAdmin.Host or IceStormAdmin.Port value` |
| `--IceStormAdmin.Port=39999` (nothing listening) | `no manager proxies configured` (after the 3s timeouts, misleading) | `could not contact the IceStorm/Finder object:` + the connection exception |

Fixes #5962.

### DIFF
--- a/changelog.d/service-icestorm/5962.md
+++ b/changelog.d/service-icestorm/5962.md
@@ -1,0 +1,4 @@
+- Improved the `icestormadmin` diagnostics when no topic manager can be reached. Without any configuration the
+  tool now prints `no manager proxies configured` instead of failing with a raw endpoint parse error, and when
+  `IceStormAdmin.Host` or `IceStormAdmin.Port` is set, a missing port, a malformed value, and an unreachable
+  `IceStorm/Finder` object are now each reported as such.
--- a/changelog.d/service-icestorm/5962.md
+++ b/changelog.d/service-icestorm/5962.md
@@ -1,4 +1,5 @@
-- Improved the `icestormadmin` diagnostics when no topic manager can be reached. Without any configuration the
-  tool now prints `no manager proxies configured` instead of failing with a raw endpoint parse error, and when
-  `IceStormAdmin.Host` or `IceStormAdmin.Port` is set, a missing port, a malformed value, and an unreachable
-  `IceStorm/Finder` object are now each reported as such.
+- Fixed `icestormadmin` to report `no manager proxies configured` when it runs without any configuration.
+  Previously, it failed with a cryptic endpoint parse error.
+- Improved the `icestormadmin` error messages when `IceStormAdmin.Host` or `IceStormAdmin.Port` is set: a missing
+  port, an invalid host or port value, and an unreachable `IceStorm/Finder` object are now reported as distinct
+  errors.
--- a/cpp/src/IceStorm/Admin.cpp
+++ b/cpp/src/IceStorm/Admin.cpp
@@ -156,20 +156,37 @@ run(const shared_ptr<Ice::Communicator>& communicator, const Ice::StringSeq& arg
         string host = properties->getIceProperty("IceStormAdmin.Host");
         string port = properties->getIceProperty("IceStormAdmin.Port");
 
-        const int timeout = 3000; // 3s connection timeout.
-        ostringstream os;
-        os << "IceStorm/Finder";
-        os << ":tcp" << (host.empty() ? "" : (" -h \"" + host + "\"")) << " -p " << port << " -t " << timeout;
-        os << ":ssl" << (host.empty() ? "" : (" -h \"" + host + "\"")) << " -p " << port << " -t " << timeout;
+        if (!host.empty() || !port.empty())
+        {
+            // The IceStorm/Finder fallback is configured. A host alone is not sufficient to build the Finder
+            // endpoints: the port is required, it has no default.
+            if (port.empty())
+            {
+                consoleErr << args[0] << ": IceStormAdmin.Host is set but IceStormAdmin.Port is not" << endl;
+                return 1;
+            }
 
-        IceStorm::FinderPrx finder{communicator, os.str()};
-        try
-        {
-            defaultManager = finder->getTopicManager();
-        }
-        catch (const Ice::LocalException&)
-        {
-            // Ignore.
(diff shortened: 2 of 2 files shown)


changelog.d/service-icestorm/5962.md
```diff
@@ -0,0 +1,4 @@
+- Improved the `icestormadmin` diagnostics when no topic manager can be reached. Without any configuration the
+  tool now prints `no manager proxies configured` instead of failing with a raw endpoint parse error, and when
+  `IceStormAdmin.Host` or `IceStormAdmin.Port` is set, a missing port, a malformed value, and an unreachable
+  `IceStorm/Finder` object are now each reported as such.
```
bernardnormier: This packs all four scenarios into one long sentence — hard to follow. Two bullets, one per user-visible improvement, read better:

```suggestion
- Fixed `icestormadmin` to report `no manager proxies configured` when it runs without any configuration.
  Previously, it failed with a cryptic endpoint parse error.
- Improved the `icestormadmin` error messages when `IceStormAdmin.Host` or `IceStormAdmin.Port` is set: a missing
  port, an invalid host or port value, and an unreachable `IceStorm/Finder` object are now reported as distinct
  errors.
```

bernardnormier (approved):
LGTM. Verified the before/after table against the code paths: `IceStormAdmin.Port` indeed has no default, the old proxy construction sat outside the `try` so the `ParseException` escaped to `main`'s generic catch, and the Finder fallback (with its new errors) only engages when no manager proxy was configured and Host/Port is actually set. The unreachable-Finder case turning from a silent fall-through into a reported error is a clear improvement. One suggestion on the changelog fragment.

cpp/src/IceStorm/Admin.cpp
```diff
@@ -156,20 +156,37 @@ run(const shared_ptr<Ice::Communicator>& communicator, const Ice::StringSeq& arg
...
-            // Ignore.
+            // The IceStorm/Finder fallback is configured. A host alone is not sufficient to build the Finder
+            // endpoints: the port is required, it has no default.
+            if (port.empty())
+            {
+                consoleErr << args[0] << ": IceStormAdmin.Host is set but IceStormAdmin.Port is not" << endl;
+                return 1;
+            }
+
+            const int timeout = 3000; // 3s connection timeout.
```
bernardnormier: We should drop that:

https://docs.zeroc.com/ice/3.8/cpp/endpoint-syntax#EndpointSyntax-TCPEndpointSyntax

cpp/src/IceStorm/Admin.cpp
```diff
@@ -156,20 +156,37 @@ run(const shared_ptr<Ice::Communicator>& communicator, const Ice::StringSeq& arg
...
+            ostringstream os;
+            os << "IceStorm/Finder";
+            os << ":tcp" << (host.empty() ? "" : (" -h \"" + host + "\"")) << " -p " << port << " -t " << timeout;
+            os << ":ssl" << (host.empty() ? "" : (" -h \"" + host + "\"")) << " -p " << port << " -t " << timeout;
```
bernardnormier: Concretely (see the thread above): in 3.8 the endpoint `-t` option is parsed but has no runtime effect — `TcpConnector` only uses it for connector comparison, and the connect timeout actually comes from `Ice.Connection.Client.ConnectTimeout` (default 10s) — so the `3s connection timeout` comment was never true. Drop the option and the variable:

```suggestion
            ostringstream os;
            os << "IceStorm/Finder";
            os << ":tcp" << (host.empty() ? "" : (" -h \"" + host + "\"")) << " -p " << port;
            os << ":ssl" << (host.empty() ? "" : (" -h \"" + host + "\"")) << " -p " << port;
```

bernardnormier: One refinement: if the 3-second probe was intentional (an admin tool shouldn't hang 10s per endpoint on an absent Finder), the correct 3.8 replacement is to reduce the connect timeout instead — e.g. in `main`, next to the existing `Ice.Warn.Endpoints` setup:

```cpp
properties->setProperty("Ice.Connection.Client.ConnectTimeout", "3"); // seconds, default 10
```

That actually restores the behavior the `-t 3000` was pretending to provide.

Issue #5969, opened 2026-07-13
SliceInfo.typeId holds a stringified compact ID in MATLAB and C#, but is empty in C++ and Python


pepone:
For a value slice encoded with a compact type ID, the mappings disagree on what `SliceInfo.typeId` contains:

- **MATLAB**: `EncapsDecoder11.m:125-128` sets `current.typeId = int2str(current.compactId)` and passes it into `Ice.SliceInfo` (`:258`). The doc (`SliceInfo.m:15-17`) says "The Slice type ID for this slice" with no compact-ID caveat.
- **C#**: same behavior — `csharp/src/Ice/InputStream.cs:2562-2570` sets `typeId = compactId.ToString(CultureInfo.InvariantCulture)` and preserves it into `SliceInfo` (`:2729-2731`); the doc is silent.
- **JavaScript**: same behavior — `js/packages/ice/src/Ice/InputStream.js:515` sets `_current.typeId = _current.compactId.toString()`, and `skipSlice` copies it verbatim into the observed `SliceInfo` (`info.typeId = _current.typeId`, `:628`). The doc is silent (`UnknownSlicedValue.js:8` / `UnknownSlicedValue.d.ts:10`: "The Slice type ID for this slice.").
- **C++**: `typeId` is empty when `compactId != -1` — documented (`cpp/include/Ice/SlicedData.h:26-27` "It's empty when the compact ID is set") and asserted in the SliceInfo constructor (`SlicedData.h:63-70`, `cpp/src/Ice/InputStream.cpp:2042`).
- **Python**: exposes the C++ `SliceInfo` directly (`python/modules/IcePy/Types.cpp:320-333`) and documents the empty case (`python/python/Ice/SliceInfo.py:19`).
- **Java**: empty, like C++/Python. `EncapsDecoder11.startSlice` computes `_current.typeId = String.valueOf(_current.compactId)` (`java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InputStream.java:1924`), but the sole `SliceInfo` construction discards it: `new SliceInfo(_current.compactId == -1 ? _current.typeId : "", …)` (`InputStream.java:2036-2041`). Documented (`SliceInfo.java:7` "It's empty when `compactId` is set (not `-1`)") and asserted in the constructor (`SliceInfo.java:42,44`), mirroring C++. Java is self-consistent — no change needed.

[...]

Issue #5967, opened 2026-07-13
MATLAB: getImplicitContext crashes MATLAB under the default configuration


pepone:
With `Ice.ImplicitContext` unset (default `None`), calling any method on the object returned by `Communicator.getImplicitContext()` dereferences a null `shared_ptr` in the MEX extension and crashes the MATLAB process.

Repro (default configuration):

```matlab
c = Ice.Communicator();
c.getImplicitContext().getContext()   % segfault
```

Chain:

- C++ `getImplicitContext()` returns null for `None` (`cpp/src/Ice/Instance.cpp:1326-1331`, assert-null; default is `None` per `cpp/src/Ice/PropertyNames.cpp:129`).
- The MEX wraps the result unconditionally: `createShared<Ice::ImplicitContext>(p)` heap-allocates `new shared_ptr<T>(move(p))` even for null (`matlab/src/Communicator.cpp:83-95`, `matlab/src/Util.h:46`) — unlike the proxy path right below (`Util.h:50-53`), which returns nullptr for nullopt so `isNull` works.
- `Communicator.m:242-246` constructs and caches `Ice.ImplicitContext(impl)` with no `isNull` check (`isNull` would be false anyway — the wrapper pointer is non-null).
- All six MEX methods (`matlab/src/ImplicitContext.cpp:22,38,52,65,79,92`) do `deref<Ice::ImplicitContext>(self)->method()`; the first member access (`lock_guard` on `_mutex`) dereferences null `this`. Not a C++ exception, so the MEX try/catch cannot intercept it.

The method's own doc (`Communicator.m:237`) promises "empty array of Ice.ImplicitContext" — that return is unreachable today. Test gap: `matlab/test/Ice/operations/Twoways.m:1300` (and TwowaysAMI.m) only exercise `Shared`/`PerThread`, never the default.

Suggested fix: mirror the proxy path — return NULL from the MEX when the C++ implicit context is null, and return `Ice.ImplicitContext.empty` from `Communicator.getImplicitContext` on `isNull`, which also makes the existing doc correct.

Found by the MATLAB API doc audit (2026-07-13); codex-confirmed.

Issue #5955, opened 2026-07-13
ToStringMode.Compat doc comment is missing the "Non-printable" qualification in Java, JS, Python, and MATLAB


pepone:
The doc comment on the `Compat` enumerator of `ToStringMode` says:

> Characters with ordinal values 127 and below are encoded as \t, \n (etc.) or an octal escape.

Only *non-printable* ASCII characters with ordinal values 127 and below are escaped; printable characters are kept as-is. The `Unicode` and `ASCII` enumerators in the same files already say "Non-printable ASCII characters…", and C++ (`cpp/include/Ice/StringUtil.h`), C# (`csharp/src/Ice/Communicator.cs`), and Swift (#5941) carry the qualification on all three enumerators. Only the `Compat` entry was missed in:

- `java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ToStringMode.java:27`
- `js/packages/ice/src/Ice/ToStringMode.d.ts:26`
- `python/python/Ice/ToStringMode.py:29`
- `matlab/lib/+Ice/ToStringMode.m:24`

Cross-mapping follow-up requested in the #5941 review, so all mappings end up with the same sentence.

pepone:
This is now fixed on main through the broader API documentation cleanup PRs:

- Java: #5933
- JavaScript: #5883
- Python: #5893
- MATLAB: #5971

I verified that the Compat documentation in all four mappings now says “Non-printable ASCII characters with ordinal values 127 and below…”, consistent with C++, C#, and Swift.

Closing as resolved.

PR #5971, opened 2026-07-14
Fix incorrect and stale API doc comments in the MATLAB Ice library

pepone:
Fixes incorrect and stale help text in the MATLAB Ice library, found during the MATLAB API doc review (follow-up to #5867/#5868/#5883/#5893/#5933/#5941). Documentation-only — every changed line is a comment line (the two `InitializationData` property declarations only lost their trailing inline comments, moved into block-style docs). The code bugs found by the same review are tracked in #5967/#5968 and deliberately not fixed here.

Incorrect claims corrected (many are the same stale pre-#5867 patterns fixed in the other mappings):

- `OptionalFormat` — the `Size`/`VSize`/`FSize` descriptions were garbled to the point of stating the inverse (`FSize` "Fixed-length format with size" actually marks variable-length data preceded by a fixed 4-byte size); all three now mirror `StreamableTraits.h`, and the `Class` summary line matches its own property comment ("no longer supported").
- `ObjectPrx.ice_flushBatchRequests`/`Async` claimed to flush "for this **communicator**" — the implementation is the per-proxy flush (×3 spots incl. the methods list).
- The constructor example `'name:tcp -p localhost -p 4061'` was an invalid proxy string ("localhost" parses as the port and throws `ParseException`); now `-h localhost` (×2).
- `Ice.Properties` constructor — documents the reserved-prefix arg-parsing scope and the real `ICE_CONFIG` condition (used when there is no `--Ice.Config` argument and `Ice.Config` is not already set). Note: unlike C++'s default constructor, the no-arg `Ice.Properties()` goes through the same path and honors `ICE_CONFIG`.
- `getIcePropertyAsList` — unset returns the property's **default list**, not "an empty list"; mismatched quotes → "the default list" (here and in `getPropertyAsListWithDefault`); the `getIceProperty*`/`getPropertyAsInt*` family now documents the `Ice.PropertyException` throws that the C++ reference documents.
- `parseIceCommandLineOptions` listed 6 prefixes as exhaustive; the implementation recognizes 16 — now uses the reference's non-exhaustive formulation.
- `AlreadyRegisteredException.kindOfObject` said "could not be **removed**" — it is an add/registration conflict ("is already registered").
- `Connection.close` was described as a blocking call — it starts a graceful closure and returns a future; the close-timeout abort (`CloseTimeoutException`) is now documented. `throwException` — "not yet **closing** or closed", corrected exception examples (peer close vs application close vs abort), and a truthful methods-list entry. `Future.cancel` — pre-3.8 AsyncResult wording replaced (local cancel, `InvocationCanceledException`, no effect once sent); the `'sent'` state is documented as twoway-only.
- `WSConnectionInfo.headers` — MATLAB connections are always outgoing, so these are the HTTP **response** headers of the upgrade handshake, not the request headers.
- `Value.ice_getSlicedData` — drops the stale "preserved-slice base class" wording and documents the real contract (empty array when not sliced; unknown slices preserved only with the sliced format). `FormatType` — Slice 1.1-only scope + corrected Compact/Sliced descriptions.
- `ToStringMode` — doubled backslashes rendered literally in `help` (`\\t` → `\t` etc.), and `Compat` gains the missing "Non-printable" qualification — this is the MATLAB part of #5955.
- Plus: `identityToString` mode description (non-ASCII handling), `intVersion`/`stringVersion` pre-release forms (30950 / `3.9.0-alpha.0`), `ConnectionAbortedException` "closed"→"aborted" copy-paste, stale `ObjectNotExist`/`OperationNotExist` wording, compare functions' `int32`→`double` output type + empty-proxy ordering, `checkedCast`/`uncheckedCast` (purely-local remark, empty-input case), empty-return cases (`propertyToProxy`, `ice_getRouter`/`ice_getLocator`, `ice_getFacet`, `setDefaultRouter`/`Locator` clearing), `stringToProxy` `ParseException`, `getInfo`-on-closed rethrow + `toString` remains-usable, SSL `peerCertificate` PEM/empty note, `SliceInfo.compactId` -1 case, `CompositeSliceLoader` consultation order, the `Ice:<Name>Exception` identifier convention + `Ice.LocalException` fallback role, methods-list drift (`disableInactivityCheck`, `checkedCast`/`uncheckedCast`/`ice_staticId`, `ice_id`/`ice_staticId`), help on undocumented public overrides (endpoint-info `type`/`secure`/`datagram`, `newInstance`, `UnknownSlicedValue.ice_id`), and markup/consistency fixes (the mapping's only "Ice run time" occurrences, verb alignment, self-referential "See also eq", missing copyright in `SliceLoader.m`, indentation).

Deliberately excluded: the `SliceInfo.typeId` doc (pending the #5969 MATLAB+C# vs C++/Python code-vs-doc decision), `InputStream`/`OutputStream` per-method docs (H1-only state is deliberate per #4141 — worth a maintainer decision before documenting ~120 methods), `(Sealed)` on the three C++-`final` leaf exceptions and the public `OutputStream.buf` property (code changes), and error-contract notes where the C++ reference is itself silent.

Full findings: `API_DOC_AUDIT_MATLAB-2026-07-13.md` (ice-audit). All Medium+ findings were adversarially re-verified against the implementation and cross-checked with codex (8 CONFIRMED, 1 PARTIAL); the MATLAB code analyzer runs in CI (doc comments are analyzer-neutral).

### DIFF
--- a/matlab/lib/+Ice/Communicator.m
+++ b/matlab/lib/+Ice/Communicator.m
@@ -118,7 +118,7 @@ classdef Communicator < IceInternal.WrapperObject
             %
             %   Input Arguments
             %     str - The stringified proxy to convert into a proxy.
-            %       character vector | string scalar
+            %       character vector
             %
             %   Output Arguments
             %     r - The proxy, or an empty array if str is an empty string.
--- a/matlab/lib/+Ice/Communicator.m
+++ b/matlab/lib/+Ice/Communicator.m
@@ -14,7 +14,7 @@ classdef Communicator < IceInternal.WrapperObject
     %     getImplicitContext - Gets the implicit context associated with this communicator.
     %     getLogger - Gets the logger for this communicator.
     %     getProperties - Gets the Ice properties for this communicator.
-    %     identityToString - Convert an identity into a string.
+    %     identityToString - Converts an identity into a string.
     %     propertyToProxy - Converts a set of proxy properties into a proxy.
     %     proxyToProperty - Converts a proxy to a set of proxy properties.
     %     proxyToString - Converts a proxy into a string.
@@ -41,9 +41,10 @@ classdef Communicator < IceInternal.WrapperObject
             %
             %   Input Name-Value Arguments
             %     Properties - Properties object used to initialize the communicator properties. If args is non-empty,
-            %       any reserved properties specified in args override these properties.
+            %       any Ice options in args (arguments starting with -- and one of the reserved prefixes, such as Ice
+            %       or IceSSL) override these properties.
             %       Ice.Properties scalar
-            %     SliceLoader - Slice loader used to load Slice classes and exceptions.
+            %     SliceLoader - The Slice loader, used to unmarshal Slice classes and exceptions.
             %       Ice.SliceLoader scalar
             %
             %   Output Arguments
@@ -113,9 +114,11 @@ classdef Communicator < IceInternal.WrapperObject
             %STRINGTOPROXY Converts a stringified proxy into a proxy.
             %   Deprecated: Use the constructor of your proxy class instead.
             %
+            %   Throws an Ice.ParseException if str is not a valid proxy string.
+            %
             %   Input Arguments
             %     str - The stringified proxy to convert into a proxy.
-            %       character vector
+            %       character vector | string scalar
             %
             %   Output Arguments
             %     r - The proxy, or an empty array if str is an empty string.
@@ -170,7 +173,7 @@ classdef Communicator < IceInternal.WrapperObject
             %       character vector
             %
             %   Output Arguments
-            %     r - The proxy.
+            %     r - The proxy, or an empty array if the property is not set.
             %       Ice.ObjectPrx scalar | empty array of Ice.ObjectPrx
 
             arguments
@@ -212,7 +215,7 @@ classdef Communicator < IceInternal.WrapperObject
         end
 
         function r = identityToString(obj, id)
-            %IDENTITYTOSTRING Convert an identity into a string.
+            %IDENTITYTOSTRING Converts an identity into a string.
             %
             %   Input Arguments
             %     id - The identity to convert into a string.
@@ -304,7 +307,8 @@ classdef Communicator < IceInternal.WrapperObject
 
         function setDefaultRouter(obj, proxy)
             %SETDEFAULTROUTER Sets a default router for this communicator.
-            %   All newly created proxies will use this default router.
+            %   All newly created proxies will use this default router. This method has no effect on existing proxies.
+            %   Pass an empty array to remove the default router.
             %
             %   Input Arguments
             %     proxy - The default router to use for this communicator.
@@ -343,7 +347,8 @@ classdef Communicator < IceInternal.WrapperObject
 
         function setDefaultLocator(obj, proxy)
             %SETDEFAULTLOCATOR Sets a default locator for this communicator.
-            %   All newly created proxies will use this default locator.
+            %   All newly created proxies will use this default locator. This method has no effect on existing
+            %   proxies. Pass an empty array to remove the default locator.
             %
             %   Input Arguments
             %     proxy - The default locator to use for this communicator.
@@ -389,7 +394,7 @@ classdef Communicator < IceInternal.WrapperObject
             %       Ice.CompressBatch scalar
             %
             %   Output Arguments
-            %     r - A future that will be completed when the invocation completes.
+            %     r - A future that completes when all batch requests have been sent.
             %       Ice.Future scalar
 
             arguments
--- a/matlab/lib/+Ice/Future.m
+++ b/matlab/lib/+Ice/Future.m
@@ -9,7 +9,7 @@ classdef Future < IceInternal.WrapperObject
     %     State - The current state of the future.
     %
     %   Future Methods:
-    %     cancel - If the invocation is still pending, calling this method instructs the local Ice runtime to ignore its results.
+    %     cancel - Cancels this invocation locally if it is still pending.
     %     fetchOutputs - Blocks until the invocation completes and then returns the results or throws an exception.
     %     wait - Blocks until the invocation reaches a certain state, or a timeout expires.
 
@@ -34,6 +34,8 @@ classdef Future < IceInternal.WrapperObject
         Read (1, 1) logical = false
 
         %STATE The current state of the future. Its initial value is 'running' and its final value is 'finished'.
+        %   The 'sent' state is only reported for twoway invocations; other futures transition directly from 'running'
+        %   to 'finished'.
         %   'running' | 'sent' | 'finished'
         State (1, :) char = 'running'
     end
@@ -76,7 +78,7 @@ classdef Future < IceInternal.WrapperObject
             %     state - If provided, wait blocks until the future reaches the given state. If not provided, wait
             %     blocks until the state is 'finished'. Note that the future enters the 'finished' state when completed
             %     successfully or exceptionally.
-            %     'running' | 'sent' | 'finished' | empty array (default)
+            %     'running' | 'sent' | 'finished'
             %   timeout - If provided, wait blocks up to the given number of seconds while waiting for the future to
             %     reach the desired state. If the timeout is negative or not provided, wait blocks indefinitely.
             %     double | empty array (default)
@@ -126,8 +128,9 @@ classdef Future < IceInternal.WrapperObject
         end
 
         function cancel(obj)
-            %CANCEL If the invocation is still pending, calling this method instructs the local Ice runtime to ignore
-            %   its results.
+            %CANCEL Cancels this invocation locally if it is still pending. The future then completes with an
+            %   Ice.InvocationCanceledException. Cancellation has no effect on the request if it was already sent to
+            %   the server.
 
             if ~isempty(obj.impl_)
                 obj.iceCall('cancel');
--- a/matlab/lib/+Ice/Future.m
+++ b/matlab/lib/+Ice/Future.m
@@ -33,9 +33,8 @@ classdef Future < IceInternal.WrapperObject
         %   logical scalar
         Read (1, 1) logical = false
 
-        %STATE The current state of the future. Its initial value is 'running' and its final value is 'finished'.
-        %   The 'sent' state is only reported for twoway invocations; other futures transition directly from 'running'
-        %   to 'finished'.
+        %STATE The current state of the future. Its final value is 'finished'.
+        %   The 'sent' state is only reported for twoway invocations.
         %   'running' | 'sent' | 'finished'
         State (1, :) char = 'running'
     end
--- a/matlab/lib/+Ice/ObjectPrx.m
+++ b/matlab/lib/+Ice/ObjectPrx.m
@@ -12,6 +12,8 @@ classdef ObjectPrx < IceInternal.WrapperObject
     %         character vector
     %
     %   ObjectPrx Methods:
+    %     checkedCast - Creates an Ice.ObjectPrx from an existing proxy after confirming the target object's type
+    %       via a remote invocation.
     %     disp - Displays this proxy as a string.
     %     eq - Compares this proxy with another ObjectPrx for equality.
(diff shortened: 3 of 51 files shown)


matlab/lib/+Ice/OptionalFormat.m
```diff
@@ -8,10 +8,12 @@
     %     F2 - Fixed-length format (2 bytes).
     %     F4 - Fixed-length format (4 bytes).
     %     F8 - Fixed-length format (8 bytes).
-    %     Size - Variable-length format with size.
-    %     VSize - Variable-length format with variable size.
-    %     FSize - Fixed-length format with size.
-    %     Class - Class format.
+    %     Size - Size encoding using either 1 or 5 bytes. Used by enums, class identifiers, etc.
```
bernardnormier: I don't find this "class identifiers" example useful, on the contrary. A class identifier is not a distinct type that you can mark "optional".

matlab/lib/+Ice/OptionalFormat.m
```diff
@@ -29,16 +31,18 @@
         %F8 Fixed-length format (8 bytes).
         F8 = uint8(3)
 
-        %SIZE Variable-length format with size.
+        %SIZE Size encoding using either 1 or 5 bytes. Used by enums, class identifiers, etc.
```
bernardnormier: Ditto

matlab/lib/+Ice/ObjectPrx.m
```diff
@@ -69,6 +71,10 @@
     %     ice_router - Returns a proxy that is identical to this proxy, except for the router.
     %     ice_toString - Creates a stringified version of this proxy.
     %     ice_twoway - Returns a proxy that is identical to this proxy, but uses twoway invocations.
+    %     checkedCast - Creates an Ice.ObjectPrx from an existing proxy after confirming the target object's type
+    %       via a remote invocation.
+    %     ice_staticId - Returns the Slice type ID associated with this type.
+    %     uncheckedCast - Creates a new Ice.ObjectPrx from an existing proxy without any validation.
```
bernardnormier: This methods list is otherwise alphabetical. `checkedCast` belongs before `disp`, and `ice_staticId` between `ice_router` and `ice_toString`; `uncheckedCast` already sorts last, so it can stay. (If the intent was to group the static methods at the end, nothing in the list marks them as such, so alphabetical seems less surprising.)

matlab/lib/+Ice/Value.m
```diff
@@ -55,6 +62,11 @@ function ice_postUnmarshal(~)
     end
     methods (Static)
         function id = ice_staticId()
+            %ICE_STATICID Gets the Slice type ID of this type.
+            %
+            %   Output Arguments
+            %     id - The return value is always '::Ice::Object'.
+            %       character vector
```
bernardnormier: "The return value is always '::Ice::Object'." reads awkwardly as an output-argument description, and the summary verb differs from the `ObjectPrx.ice_staticId` doc added in this same PR. Suggest aligning both on the `ObjectPrx` phrasing:

```suggestion
            %ICE_STATICID Returns the Slice type ID associated with this type.
            %
            %   Output Arguments
            %     id - The Slice type ID ('::Ice::Object').
            %       character vector
```

If you take this, the methods-list entry at the top of the file (`ice_staticId - Gets the Slice type ID of this type.`) should follow suit.

matlab/lib/+Ice/Communicator.m
```diff
@@ -113,9 +114,11 @@ function destroy(obj)
             %STRINGTOPROXY Converts a stringified proxy into a proxy.
             %   Deprecated: Use the constructor of your proxy class instead.
             %
+            %   Throws an Ice.ParseException if str is not a valid proxy string.
+            %
             %   Input Arguments
             %     str - The stringified proxy to convert into a proxy.
-            %       character vector
+            %       character vector | string scalar
```
bernardnormier: Every other string-typed input in this PR keeps plain `character vector`, even though the `(1, :) char` coercion accepts string scalars everywhere — adding `| string scalar` only here makes `stringToProxy` look special when it isn't. Suggest keeping the uniform notation (a mapping-wide `| string scalar` pass could be done separately if desired):

```suggestion
            %       character vector
```

bernardnormier (approved):
LGTM. Doc-comment-only; I spot-checked the riskiest behavioral claims against the implementation (`'sent'` twoway-only per `InvocationFuture`, compare functions returning `double` with empty-proxy ordering, the `Ice:<Name>Exception` identifier convention and `Ice.LocalException` fallback in `Util.cpp`, `ClassSliceLoader`'s `Ice:ArgumentException`, 30950/`3.9.0-alpha.0`, and the `Future.wait` state values) and they all check out. A few consistency nits inline, none blocking.

matlab/lib/+Ice/OptionalFormat.m
bernardnormier: We should actually fix this in all mappings, not just MATLAB.

I also confirmed that only enums use this Size encoding.

matlab/lib/+Ice/Future.m
```diff
@@ -34,6 +34,8 @@
         Read (1, 1) logical = false
 
         %STATE The current state of the future. Its initial value is 'running' and its final value is 'finished'.
+        %   The 'sent' state is only reported for twoway invocations; other futures transition directly from 'running'
+        %   to 'finished'.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Done in b8baafaa4a — confirmed batch invocation futures are constructed directly in `Finished` (`ObjectPrx.cpp:40`, `_state(batch ? State::Finished : State::Running)`), so they never pass through `running`. Dropped both the transition claim and the "initial value is 'running'" claim, keeping only the twoway-only `'sent'` distinction.

matlab/lib/+Ice/UnknownSlicedValue.m
```diff
@@ -2,7 +2,7 @@
     %UNKNOWNSLICEDVALUE Represents an instance of an unknown class.
     %
     %   UnknownSlicedValue Methods:
-    %     ice_id - Returns the Slice type ID associated with this instance.
+    %     ice_id - Returns the Slice type ID of the unknown most-derived class (the type ID of the first slice).
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Done in b8baafaa4a — the summary now covers both representations (a Slice type ID, or a compact ID as a decimal string when the most-derived slice used a compact ID).

matlab/lib/+Ice/UnknownSlicedValue.m
```diff
@@ -13,6 +13,11 @@
     end
     methods
         function id = ice_id(obj)
+            %ICE_ID Returns the Slice type ID of the unknown most-derived class (the type ID of the first slice).
+            %
+            %   Output Arguments
+            %     id - The Slice type ID.
+            %       character vector
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Done in b8baafaa4a — documented the compact-ID case on the method. For the record this is cross-mapping-consistent: C++ and C# stringify the compact ID into `ice_id` the same way via `mostDerivedId`, which is a distinct code path from the `SliceInfo.typeId` divergence tracked in #5969.

matlab/lib/+Ice/OptionalFormat.m
pepone: 🤖 **Claude:** Done in b8baafaa4a — `Size` is now "Used by enums." (both the summary and the property comment). Confirmed in the Slice compiler that `Enum` is the only type whose `getOptionalFormat()` returns `Size` (`Parser.cpp`; `GetOptionalFormat` maps only the enum category to `OptionalFormat::Size` in `StreamHelpers.h`). Filed #5993 to make the same correction in the C++ `StreamableTraits.h` reference and the C#/Java/JS/Swift docs (Python/Ruby/PHP have no public `OptionalFormat` doc).

matlab/lib/+Ice/OptionalFormat.m
pepone: 🤖 **Claude:** Done in b8baafaa4a — same fix on the property comment.

matlab/lib/+Ice/ObjectPrx.m
pepone: 🤖 **Claude:** Done in b8baafaa4a — `checkedCast` now sorts before `disp`, `ice_staticId` between `ice_router` and `ice_toString`, and `uncheckedCast` stays last.

matlab/lib/+Ice/Value.m
pepone: 🤖 **Claude:** Done in b8baafaa4a — applied your suggestion; the methods-list entry at the top of the file now reads "Returns the Slice type ID associated with this type." too.

matlab/lib/+Ice/Communicator.m
pepone: 🤖 **Claude:** Done in b8baafaa4a — reverted to plain `character vector`.

matlab/lib/+Ice/Properties.m
```diff
@@ -217,12 +229,14 @@
 
         function r = getIcePropertyAsList(obj, key)
             %GETICEPROPERTYASLIST Gets an Ice property as a list of strings.
-             %  The strings must be separated by whitespace or comma. If the property is not set, an empty list is
+            %   The strings must be separated by whitespace or comma. If the property is not set, its default list is
             %   returned. The strings in the list can contain whitespace and commas if they are enclosed in single or
-            %   double quotes. If quotes are mismatched, an empty list is returned. Within single quotes or double
+            %   double quotes. If quotes are mismatched, the default list is returned. Within single quotes or double
             %   quotes, you can escape the quote in question with \, e.g. O'Reilly can be written as O'Reilly,
             %   "O'Reilly" or 'O\'Reilly'.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** You're right — I confirmed from source that the default isn't reliably returned: `splitString` pushes the accumulated element before returning `false` (`StringUtil.cpp:687-693`), and `getPropertyAsListWithDefault` substitutes the default only `if (result.size() == 0)` (`Properties.cpp:319`), so a non-empty mismatched value like `'foo` returns `["foo"]` despite the warning saying "returning default value". This is a C++-core bug (C# and Java are already correct — they return the default), not MATLAB-specific, and the MATLAB doc mirrors the reference header. Filed #5994 to fix the C++ core so it returns the default like C#/Java; keeping the MATLAB doc aligned with the reference rather than forking it.

matlab/lib/+Ice/Properties.m
```diff
@@ -242,7 +256,7 @@
             %GETPROPERTYASLISTWITHDEFAULT Gets a property as a list of strings.
             %   The strings must be separated by whitespace or comma. If the property is not set, the default value is
             %   returned. The strings in the list can contain whitespace and commas if they are enclosed in single or
-            %   double quotes. If quotes are mismatched, an empty list is returned. Within single quotes or double
+            %   double quotes. If quotes are mismatched, the default list is returned. Within single quotes or double
             %   quotes, you can escape the quote in question with \, e.g. O'Reilly can be written as O'Reilly,
             %   "O'Reilly" or 'O\'Reilly'.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Same as the `getIcePropertyAsList` thread — C++-core behavior (C# and Java are already correct), tracked in #5994.

matlab/lib/+Ice/Properties.m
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Confirmed — bare unquoted `O'Reilly` opens a quote at the apostrophe and parses to `OReilly` plus a mismatched-quote failure, so the first example form is indeed bogus. This exact text is in the C++ reference header (`Properties.h`) and every mapping, so I filed #5995 to fix it cross-mapping rather than diverging MATLAB alone.

matlab/lib/+Ice/Properties.m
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Same as above — shared reference-header text; tracked in #5995.

PR #5974, opened 2026-07-14
Fix MATLAB Ice.Future.fetchOutputs crash when called a second time after a failed invocation

### LINKED ISSUE #5968: MATLAB: calling fetchOutputs a second time after a failed invocation crashes MATLAB
`Ice.Future.fetchOutputs` documents "Can only be called once" and guards with `error('Ice:InvalidStateException', 'Outputs already read')` — but the guard is never armed when the invocation fails, and the second call crashes the MATLAB process instead of raising the documented error.

Flow (`matlab/lib/+Ice/Future.m:109-125`):

- The guard checks `obj.Read` (`:109-111`).
- On the exception path, the `catch` clears `impl_` and rethrows (`:119-122`) **before** `obj.Read = true` (`:125`, after the try/catch) — so `Read` stays false.
- A second `fetchOutputs` passes the guard, re-invokes `fetchFunc` (never cleared), and `calllib` converts the empty `impl_` to a NULL pointer (documented MATLAB behavior for `[]` on pointer parameters). `WrapperObject.iceCall`/`IceInternal.Util.call` have no isempty check — the one unguarded path (`delete`/`wait`/`cancel`/`get.State` all guard on `~isempty(obj.impl_)`).
- All four native fetch entry points dereference `self` unconditionally via `deref` (`matlab/src/Util.h:44`): `Ice_SimpleFuture_check` (`matlab/src/Future.cpp:178-194`), `Ice_InvocationFuture_results`/`_check`, `Ice_GetConnectionFuture_fetch` (`matlab/src/ObjectPrx.cpp:947,988,995,1039,1050`) → NULL dereference, segfault.

Affects every async API (invocations, `ice_getConnectionAsync`, `flushBatchRequestsAsync`, `Connection.close`). The trigger is local API misuse (e.g. a retry loop or duplicated error handling calling `fetchOutputs` twice on a failed future), but the cost is the whole MATLAB session.

Suggested fix: set `obj.Read = true` in the catch branch too (or guard `fetchOutputs` on `isempty(obj.impl_)`), so the documented contract is enforced on the exception path.

Found by the MATLAB API doc audit (2026-07-13); codex-confirmed.

pepone:
Fixes #5968.

### Problem
`Ice.Future.fetchOutputs` is documented "Can only be called once" and guards re-entry with
`error('Ice:InvalidStateException', ...)`. But `obj.Read = true` was set only *after* the `try/catch`, so when the
first call completes exceptionally (the fetch function throws), the `catch` clears `impl_` and rethrows before `Read`
is armed — `Read` stays `false`.

A second `fetchOutputs` then passes the guard and re-invokes the fetch function while `obj.impl_ == []`. MATLAB's
`calllib` converts the empty `self` argument to a NULL pointer, and the native fetch/check entry points
(`Ice_SimpleFuture_check`, `Ice_InvocationFuture_results`/`_check`, `Ice_GetConnectionFuture_fetch`) dereference `self`
unconditionally, crashing the whole MATLAB process instead of raising the documented exception. This affects every
async API: invocations, `ice_getConnectionAsync`, `flushBatchRequestsAsync`, and `Connection.close`.

### Fix
Set `obj.Read = true` immediately after the guard, before invoking the fetch function. The fetch function runs at most
once (it deletes the underlying C++ object), so arming `Read` up front makes any second call — whether the first
succeeded or threw — raise `Ice:InvalidStateException`, exactly as documented.

### Testing
Added a regression test to the `Ice/ami` suite: it drives an indirect invocation to exceptional completion
(`Ice.NoEndpointException`), then asserts a second `fetchOutputs` raises `Ice:InvalidStateException` rather than
crashing.

Verified the fix locally with MATLAB R2025b against a freshly-built `ice` MEX: on a future that completed
exceptionally, a second `fetchOutputs` now raises `Ice:InvalidStateException` (previously a hard process crash), while
the first call still surfaces the underlying `NoEndpointException`. CI exercises the full `ami` suite. Verified the
bug and reviewed the diff with codex.

### DIFF
--- a/changelog.d/matlab/5968.md
+++ b/changelog.d/matlab/5968.md
@@ -1,2 +1,0 @@
-- Fixed `Ice.Future.fetchOutputs`, which crashed the MATLAB process when called a second time after the first call
-  completed with an exception. As documented, any second call now raises `Ice:InvalidStateException`.
--- a/changelog.d/matlab/5968.md
+++ b/changelog.d/matlab/5968.md
@@ -1,0 +1,2 @@
+- Fixed `Ice.Future.fetchOutputs`, which crashed the MATLAB process when called a second time after the first call
+  completed with an exception. As documented, any second call now raises `Ice:InvalidStateException`.
--- a/matlab/lib/+Ice/Future.m
+++ b/matlab/lib/+Ice/Future.m
@@ -109,11 +109,14 @@ classdef Future < IceInternal.WrapperObject
             if obj.Read
                 error('Ice:InvalidStateException', 'Outputs already read');
             end
+            %
+            % Mark the outputs as read before invoking the fetch function. The fetch function deletes the
+            % underlying C++ object (clearing impl_ below), so it can only run once; arming Read here ensures a
+            % second call raises the exception above even when the first call completes exceptionally, rather
+            % than re-invoking the fetch function on a null object and crashing MATLAB.
+            %
+            obj.Read = true;
             if ~isempty(obj.fetchFunc)
-                %
-                % We assume the fetch function implementation also deletes the C++ object so that we
-                % can avoid another call into C++.
-                %
                 try
                     [varargout{1:obj.NumOutputArguments}] = obj.fetchFunc(obj);
                     obj.impl_ = [];
@@ -122,7 +125,6 @@ classdef Future < IceInternal.WrapperObject
                     rethrow(ex);
                 end
             end
-            obj.Read = true;
         end
 
         function cancel(obj)
--- a/matlab/lib/+Ice/Future.m
+++ b/matlab/lib/+Ice/Future.m
@@ -110,10 +110,8 @@ classdef Future < IceInternal.WrapperObject
                 error('Ice:InvalidStateException', 'Outputs already read');
             end
             %
-            % Mark the outputs as read before invoking the fetch function. The fetch function deletes the
-            % underlying C++ object (clearing impl_ below), so it can only run once; arming Read here ensures a
-            % second call raises the exception above even when the first call completes exceptionally, rather
-            % than re-invoking the fetch function on a null object and crashing MATLAB.
+            % The fetch function runs at most once: it deletes the underlying C++ object even when
+            % it throws. Set Read before calling it so a second call always takes the guard above.
             %
             obj.Read = true;
             if ~isempty(obj.fetchFunc)
--- a/matlab/test/Ice/ami/AllTests.m
+++ b/matlab/test/Ice/ami/AllTests.m
@@ -108,7 +108,7 @@ classdef AllTests
 
             fprintf('testing fetchOutputs after a failed invocation... ');
             % A second fetchOutputs call must raise InvalidStateException rather than crash the MATLAB
-            % process, even when the first call completed exceptionally. See issue #5968.
+            % process, even when the first call completed exceptionally.
             f = p.ice_adapterId('dummy').opAsync();
             try
                 f.fetchOutputs();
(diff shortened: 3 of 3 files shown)


matlab/test/Ice/ami/AllTests.m
```diff
@@ -106,6 +106,24 @@ function allTests(helper)
 
             fprintf('ok\n');
 
+            fprintf('testing fetchOutputs after a failed invocation... ');
+            % A second fetchOutputs call must raise InvalidStateException rather than crash the MATLAB
+            % process, even when the first call completed exceptionally. See issue #5968.
```
pepone:
```suggestion
            % A second fetchOutputs call must raise InvalidStateException rather than crash the MATLAB
            % process, even when the first call completed exceptionally.
```

changelog.d/matlab/5968.md
```diff
@@ -0,0 +1,2 @@
+- Fixed `Ice.Future.fetchOutputs`, which crashed the MATLAB process when called a second time after the first call
```
bernardnormier: I'd drop this changelog entry: no valid program calls `fetchOutputs` twice, so the fix only changes what an already-buggy program observes (the documented exception instead of a crash).

matlab/lib/+Ice/Future.m
```diff
@@ -109,11 +109,14 @@ function delete(obj)
             if obj.Read
                 error('Ice:InvalidStateException', 'Outputs already read');
             end
+            %
+            % Mark the outputs as read before invoking the fetch function. The fetch function deletes the
+            % underlying C++ object (clearing impl_ below), so it can only run once; arming Read here ensures a
+            % second call raises the exception above even when the first call completes exceptionally, rather
+            % than re-invoking the fetch function on a null object and crashing MATLAB.
+            %
```
bernardnormier, line 115: Most of this comment is justification for the reviewer ("rather than re-invoking the fetch function on a null object and crashing MATLAB") — that story is already in the PR description where it belongs. In the source, stating the local invariant is enough:

```suggestion
            % The fetch function runs at most once: it deletes the underlying C++ object even when
            % it throws. Set Read before calling it so a second call always takes the guard above.
```

bernardnormier (approved):
LGTM. Verified the invariant the fix relies on: all three native fetch/check entry points (`Ice_InvocationFuture_check`, `Ice_GetConnectionFuture_fetch`, `Ice_SimpleFuture_check`) delete the C++ object on both the success and exception paths, so the fetch function can only ever run once and arming `Read` before the `try` is the right fix — it covers every async surface through the shared `Future.m`, and the success path is unchanged. Two comments inline.

matlab/lib/+Ice/Future.m
```diff
@@ -109,11 +109,14 @@ function delete(obj)
             if obj.Read
                 error('Ice:InvalidStateException', 'Outputs already read');
             end
+            %
+            % Mark the outputs as read before invoking the fetch function. The fetch function deletes the
+            % underlying C++ object (clearing impl_ below), so it can only run once; arming Read here ensures a
+            % second call raises the exception above even when the first call completes exceptionally, rather
+            % than re-invoking the fetch function on a null object and crashing MATLAB.
```
InsertCreativityHere: I think the rest of the comment is fine, it's really just the last sentence that I think strays into "describing a past bug". But, feel free to rewrite how you want.
```suggestion
            % second call raises the exception above even when the first call completes exceptionally.
```

InsertCreativityHere (approved):
Looks good to me!

PR #5975, opened 2026-07-14
Fix MATLAB Communicator.getImplicitContext crash under the default configuration

### LINKED ISSUE #5967: MATLAB: getImplicitContext crashes MATLAB under the default configuration

pepone:
Fixes #5967.

### Problem
With `Ice.ImplicitContext` unset (default `None`), calling any method on the object returned by
`Communicator.getImplicitContext()` crashes the MATLAB process.

C++ `getImplicitContext()` returns a null `shared_ptr` for `None` (documented in `Ice/Communicator.h`). The MEX
wrapped that null unconditionally — `createShared<Ice::ImplicitContext>(p)` is `new shared_ptr<T>(move(p))`, a
**non-null** heap pointer holding a null `shared_ptr` — so `isNull(impl)` was false and `Communicator.m` cached a
live `Ice.ImplicitContext`. The six native `Ice_ImplicitContext_*` methods then `deref` that null `shared_ptr` and
dereference a null `this` on the first member access. This is a hard crash, not an Ice exception, so the MEX
try/catch cannot intercept it. The method's own doc already promised an *"empty array of Ice.ImplicitContext"*
return that was unreachable.

### Fix
Mirror the proxy path (`stringToProxy`/`propertyToProxy`, which already do the `isNull` → `.empty` dance):

- `matlab/src/Communicator.cpp`: return a NULL pointer from the MEX when the C++ implicit context is null
  (`*ctx = p ? createShared<Ice::ImplicitContext>(p) : nullptr;`). The generic `createShared` template is left
  unchanged — its other callers (logger, endpoints) are never null.
- `matlab/lib/+Ice/Communicator.m`: when `isNull(impl)`, return `Ice.ImplicitContext.empty`, making the documented
  empty-array return reachable.

### Testing
Added a default-`None` regression test to both the `Twoways` and `TwowaysAMI` blocks of the `Ice/operations` suite
(the two files that previously only exercised `Shared`/`PerThread`): it creates a communicator under the default
configuration and asserts `getImplicitContext()` returns an empty array instead of crashing.

Rebuilt the `ice` MEX and ran `Ice/operations` locally with MATLAB R2025b against a Python server — all cases pass,
including the new default-`None` assertions in both `Twoways` and `TwowaysAMI`. Also confirmed directly that under
the default configuration `getImplicitContext()` returns an empty array (no crash), while `Shared` still returns a
usable, cached context. Verified the bug and reviewed the diff with codex.

### DIFF
--- a/matlab/src/Communicator.cpp
+++ b/matlab/src/Communicator.cpp
@@ -85,7 +85,9 @@ extern "C"
         try
         {
             auto p = deref<Ice::Communicator>(self)->getImplicitContext();
-            *ctx = createShared<Ice::ImplicitContext>(p);
+            // p is null when Ice.ImplicitContext is None (the default); return a null pointer so the MATLAB side
+            // can map it to an empty array instead of wrapping a null shared_ptr that later crashes on deref.
+            *ctx = p ? createShared<Ice::ImplicitContext>(p) : nullptr;
             return createEmptyArray();
         }
         catch (...)
--- a/matlab/src/Communicator.cpp
+++ b/matlab/src/Communicator.cpp
@@ -85,8 +85,8 @@ extern "C"
         try
         {
             auto p = deref<Ice::Communicator>(self)->getImplicitContext();
-            // p is null when Ice.ImplicitContext is None (the default); return a null pointer so the MATLAB side
-            // can map it to an empty array instead of wrapping a null shared_ptr that later crashes on deref.
+            // p is null when Ice.ImplicitContext is None (the default); the MATLAB side maps a null pointer to an
+            // empty array.
             *ctx = p ? createShared<Ice::ImplicitContext>(p) : nullptr;
             return createEmptyArray();
         }
--- a/matlab/test/Ice/operations/Twoways.m
+++ b/matlab/test/Ice/operations/Twoways.m
@@ -1295,8 +1295,7 @@ classdef Twoways
             assert(isequal(r, ctx));
 
             %
-            % Under the default configuration (Ice.ImplicitContext=None), getImplicitContext returns an empty
-            % array rather than crashing the MATLAB process.
+            % Under the default configuration (Ice.ImplicitContext=None), getImplicitContext returns an empty array.
             %
             icNone = helper.initialize(Properties = helper.communicator().getProperties().clone());
             assert(isempty(icNone.getImplicitContext()));
--- a/matlab/test/Ice/operations/Twoways.m
+++ b/matlab/test/Ice/operations/Twoways.m
@@ -1296,7 +1296,7 @@ classdef Twoways
 
             %
             % Under the default configuration (Ice.ImplicitContext=None), getImplicitContext returns an empty
-            % array rather than crashing the MATLAB process. See issue #5967.
+            % array rather than crashing the MATLAB process.
             %
             icNone = helper.initialize(Properties = helper.communicator().getProperties().clone());
             assert(isempty(icNone.getImplicitContext()));
--- a/matlab/test/Ice/operations/Twoways.m
+++ b/matlab/test/Ice/operations/Twoways.m
@@ -1295,6 +1295,14 @@ classdef Twoways
             assert(isequal(r, ctx));
 
             %
+            % Under the default configuration (Ice.ImplicitContext=None), getImplicitContext returns an empty
+            % array rather than crashing the MATLAB process. See issue #5967.
+            %
+            icNone = helper.initialize(Properties = helper.communicator().getProperties().clone());
+            assert(isempty(icNone.getImplicitContext()));
+            icNone.destroy();
+
+            %
             % Test implicit context propagation
             %
             impls = {'Shared', 'PerThread'};
--- a/matlab/test/Ice/operations/TwowaysAMI.m
+++ b/matlab/test/Ice/operations/TwowaysAMI.m
@@ -1111,6 +1111,14 @@ classdef TwowaysAMI
             assert(isequal(r, ctx));
 
             %
+            % Under the default configuration (Ice.ImplicitContext=None), getImplicitContext returns an empty
+            % array rather than crashing the MATLAB process. See issue #5967.
+            %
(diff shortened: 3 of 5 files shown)


matlab/test/Ice/operations/Twoways.m
```diff
@@ -1294,6 +1294,14 @@ function twoways(helper, p)
             r = p2.opContext(ctx);
             assert(isequal(r, ctx));
 
+            %
+            % Under the default configuration (Ice.ImplicitContext=None), getImplicitContext returns an empty
+            % array rather than crashing the MATLAB process. See issue #5967.
```
pepone:
```suggestion
            % array rather than crashing the MATLAB process.
```

matlab/test/Ice/operations/TwowaysAMI.m
```diff
@@ -1110,6 +1110,14 @@ function twowaysAMI(helper, p)
             r = p2.opContext(ctx);
             assert(isequal(r, ctx));
 
+            %
+            % Under the default configuration (Ice.ImplicitContext=None), getImplicitContext returns an empty
+            % array rather than crashing the MATLAB process. See issue #5967.
```
pepone:
```suggestion
            % array rather than crashing the MATLAB process.
```

matlab/src/Communicator.cpp
```diff
@@ -85,7 +85,9 @@ extern "C"
         try
         {
             auto p = deref<Ice::Communicator>(self)->getImplicitContext();
-            *ctx = createShared<Ice::ImplicitContext>(p);
+            // p is null when Ice.ImplicitContext is None (the default); return a null pointer so the MATLAB side
+            // can map it to an empty array instead of wrapping a null shared_ptr that later crashes on deref.
```
bernardnormier: The "instead of wrapping a null shared_ptr that later crashes on deref" tail is justification for the reviewer — that story belongs in the PR description (where it already is). The durable part is the contract:

```suggestion
            // p is null when Ice.ImplicitContext is None (the default); the MATLAB side maps a null pointer to an
            // empty array.
```

matlab/test/Ice/operations/Twoways.m
```diff
@@ -1294,6 +1294,14 @@ function twoways(helper, p)
             r = p2.opContext(ctx);
             assert(isequal(r, ctx));
 
+            %
+            % Under the default configuration (Ice.ImplicitContext=None), getImplicitContext returns an empty
+            % array rather than crashing the MATLAB process.
+            %
```
bernardnormier, line 1299: "rather than crashing the MATLAB process" describes the old bug, not the behavior under test — the test should just state what it asserts:

```suggestion
            %
            % Under the default configuration (Ice.ImplicitContext=None), getImplicitContext returns an empty array.
            %
```

matlab/test/Ice/operations/TwowaysAMI.m
```diff
@@ -1110,6 +1110,14 @@ function twowaysAMI(helper, p)
...
             assert(isequal(r, ctx));
 
+            %
+            % Under the default configuration (Ice.ImplicitContext=None), getImplicitContext returns an empty
+            % array rather than crashing the MATLAB process.
+            %
+            icNone = helper.initialize(Properties = helper.communicator().getProperties().clone());
+            assert(isempty(icNone.getImplicitContext()));
+            icNone.destroy();
+
```
bernardnormier: This copy is byte-identical to the `Twoways.m` test and exercises nothing asynchronous — `getImplicitContext` is a synchronous call — so it adds no coverage. I'd drop it and keep only the `Twoways.m` one.

```suggestion
```

bernardnormier (approved):
LGTM. Verified the root cause end-to-end (C++ documents the nullptr return for `None`; `createShared` yields a non-null wrapper around a null `shared_ptr`, so the old `isNull` could never fire), that the fix follows the `stringToProxy`/`propertyToProxy` pattern and makes the already-documented empty-array return reachable, that the untyped `implicitContext` cache accepts the empty assignment, and that the other `createShared` callers can never receive null. Three comment/test trims inline.

InsertCreativityHere (approved):
Looks good to me!

PR #5976, opened 2026-07-14
Return an empty SliceInfo.typeId for compact-ID slices in Ice for MATLAB

pepone:
Part of #5969.

### Problem

### Fix
and blank `typeId` **only** at the single `Ice.SliceInfo` construction site when a compact ID is present —
`matlab/lib/+IceInternal/EncapsDecoder11.m`. Also update the `Ice.SliceInfo.typeId` doc to state it is empty when
`compactId` is set, matching C++/Java.

### Testing
compact-ID slice*: the only compact types in the slicing suite are client-known (so they are fully reconstructed,
producing no `SlicedData`), and every "unknown preserved" server operation returns string-ID types. A true
end-to-end test would require new, coordinated cross-language `.ice` types plus a preserving operation. The change
was verified by an adversarial multi-agent review and an independent codex review, both of which re-derived the
single construction site and confirmed compact-id factory resolution is untouched. Pure `.m` change (no rebuild);
CI runs the `slicing` suites. Happy to add a coordinated cross-language regression test as a follow-up if wanted.

### DIFF
--- a/changelog.d/matlab/5969.md
+++ b/changelog.d/matlab/5969.md
@@ -1,3 +1,3 @@
-- `Ice.SliceInfo.typeId` is now empty for a value slice decoded via a compact type ID, matching the C++, Java, and
-  Python mappings; the compact ID remains available through `Ice.SliceInfo.compactId`. Previously `typeId` held the
-  stringified compact ID.
+- `Ice.SliceInfo.typeId` is now empty for the slice of a class with a compact type ID, matching the C++, Java, and
+  Python mappings. Previously, `typeId` held the stringified compact ID. The compact ID itself remains available
+  through `Ice.SliceInfo.compactId`.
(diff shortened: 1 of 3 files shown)


changelog.d/matlab/5969.md
```diff
@@ -0,0 +1,3 @@
+- `Ice.SliceInfo.typeId` is now empty for a value slice decoded via a compact type ID, matching the C++, Java, and
+  Python mappings; the compact ID remains available through `Ice.SliceInfo.compactId`. Previously `typeId` held the
+  stringified compact ID.
```
bernardnormier: "for a value slice decoded via a compact type ID" is hard to parse — what this is about is a class with a compact type ID:

```suggestion
- `Ice.SliceInfo.typeId` is now empty for the slice of a class with a compact type ID, matching the C++, Java, and
  Python mappings. Previously, `typeId` held the stringified compact ID. The compact ID itself remains available
  through `Ice.SliceInfo.compactId`.
```

bernardnormier (approved):
LGTM. Verified this mirrors the Java reference exactly — Java also keeps the stringified compact ID internally (`InputStream.java:1924`) and blanks it only at the `SliceInfo` construction site (`InputStream.java:2037`) — that every `readSliceType` branch sets `compactId` so the `== -1` discriminator is reliable, and that factory resolution and `ice_id` are untouched. One wording suggestion on the changelog. Heads-up: #5971 edits the doc lines directly adjacent to this PR's `SliceInfo.m` changes, so whichever merges second will need a trivial conflict resolution.

PR #5977, opened 2026-07-14
Return an empty SliceInfo.typeId for compact-ID slices in Ice for .NET

pepone:
Part of #5969.

### Problem

### Fix
and blank `typeId` **only** at the single `SliceInfo` construction site when a compact ID is present —
`csharp/src/Ice/InputStream.cs`. Also update the `SliceInfo.typeId` doc (`csharp/src/Ice/SlicedData.cs`) to state it
is empty when the slice was decoded via a compact ID, matching C++/Java.

### Testing
compact-ID slice*: the only compact types in the slicing suite are client-known (fully reconstructed, no
`SlicedData`), and every "unknown preserved" server operation returns string-ID types. A true end-to-end test would
require new, coordinated cross-language `.ice` types plus a preserving operation. Verified by an adversarial
multi-agent review and an independent codex review (single construction site confirmed; factory resolution
untouched). The change is a trivial `string` ternary; a local full build is gated on first building `slice2cs`, so
CI performs the full C# compile and runs the `slicing` suites. Happy to add a coordinated cross-language regression
test as a follow-up if wanted.

### DIFF
--- a/changelog.d/csharp/5969.md
+++ b/changelog.d/csharp/5969.md
@@ -1,3 +1,3 @@
-- `Ice.SliceInfo.typeId` is now empty for a value slice decoded via a compact type ID, matching the C++, Java, and
-  Python mappings; the compact ID remains available through `Ice.SliceInfo.compactId`. Previously `typeId` held the
-  stringified compact ID.
+- `Ice.SliceInfo.typeId` is now empty for the slice of a class with a compact type ID, matching the C++, Java, and
+  Python mappings. Previously, `typeId` held the stringified compact ID. The compact ID itself remains available
+  through `Ice.SliceInfo.compactId`.
--- a/changelog.d/csharp/5969.md
+++ b/changelog.d/csharp/5969.md
@@ -1,0 +1,3 @@
+- `Ice.SliceInfo.typeId` is now empty for a value slice decoded via a compact type ID, matching the C++, Java, and
+  Python mappings; the compact ID remains available through `Ice.SliceInfo.compactId`. Previously `typeId` held the
+  stringified compact ID.
--- a/csharp/src/Ice/InputStream.cs
+++ b/csharp/src/Ice/InputStream.cs
@@ -2727,7 +2727,6 @@ public sealed class InputStream
                 b.position(end);
 
                 var info = new SliceInfo(
-                    // Empty type ID when the slice was decoded via a compact ID.
                     typeId: _current.compactId == -1 ? _current.typeId! : "",
                     compactId: _current.compactId,
                     bytes: bytes,
--- a/csharp/src/Ice/InputStream.cs
+++ b/csharp/src/Ice/InputStream.cs
@@ -2727,8 +2727,7 @@ public sealed class InputStream
                 b.position(end);
 
                 var info = new SliceInfo(
-                    // Empty type ID when the slice was decoded via a compact ID; the compact ID is carried
-                    // separately by SliceInfo.compactId (matches C++/Java/Python).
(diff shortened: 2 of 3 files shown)


csharp/src/Ice/InputStream.cs
```diff
@@ -2727,7 +2727,9 @@ internal override void skipSlice()
                 b.position(end);
 
                 var info = new SliceInfo(
-                    typeId: _current.typeId!,
+                    // Empty type ID when the slice was decoded via a compact ID; the compact ID is carried
+                    // separately by SliceInfo.compactId (matches C++/Java/Python).
```
pepone:
```suggestion
                    // Empty type ID when the slice was decoded via a compact ID.
```

changelog.d/csharp/5969.md
```diff
@@ -0,0 +1,3 @@
+- `Ice.SliceInfo.typeId` is now empty for a value slice decoded via a compact type ID, matching the C++, Java, and
+  Python mappings; the compact ID remains available through `Ice.SliceInfo.compactId`. Previously `typeId` held the
+  stringified compact ID.
```
bernardnormier: Same reword as the MATLAB fragment — "for a value slice decoded via a compact type ID" is hard to parse; this is about a class with a compact type ID:

```suggestion
- `Ice.SliceInfo.typeId` is now empty for the slice of a class with a compact type ID, matching the C++, Java, and
  Python mappings. Previously, `typeId` held the stringified compact ID. The compact ID itself remains available
  through `Ice.SliceInfo.compactId`.
```

csharp/src/Ice/SlicedData.cs
```diff
@@ -13,7 +13,8 @@ public sealed record class SlicedData(SliceInfo[] slices);
 /// <summary>
 /// Encapsulates the details of a class slice with an unknown type.
 /// </summary>
-/// <param name="typeId">The Slice type ID for this slice.</param>
+/// <param name="typeId">The Slice type ID for this slice. It's empty when the slice was decoded via a compact ID
+/// (<paramref name="compactId"/> is not -1).</param>
```
bernardnormier: The C++ reference states this more plainly ("It's empty when the compact ID is set (not `-1`)") — suggest aligning:

```suggestion
/// <param name="typeId">The Slice type ID for this slice. It's empty when the compact ID is set
/// (<paramref name="compactId"/> is not -1).</param>
```

csharp/src/Ice/InputStream.cs
```diff
@@ -2727,7 +2727,8 @@ internal override void skipSlice()
                 b.position(end);
 
                 var info = new SliceInfo(
-                    typeId: _current.typeId!,
+                    // Empty type ID when the slice was decoded via a compact ID.
+                    typeId: _current.compactId == -1 ? _current.typeId! : "",
```
bernardnormier: The ternary right above the `compactId:` argument already says this — I'd drop the comment:

```suggestion
                    typeId: _current.compactId == -1 ? _current.typeId! : "",
```

bernardnormier (approved):
LGTM. Same verification as the MATLAB companion (#5976): exact mirror of the Java reference, every `readSliceType` branch sets `compactId` so the `== -1` discriminator is reliable, this is the only `SliceInfo` construction site in `csharp/src`, and factory resolution plus the null-forgiving `typeId!` are untouched (the no-type-ID branch stores `""`, never null). Three wording suggestions inline — the "decoded via a compact ID" phrasing reads hard in all three spots.

PR #5978, opened 2026-07-14
Return an empty SliceInfo.typeId for compact-ID slices in Ice for JavaScript

pepone:
Part of #5969.

### Problem

### Fix
and blank `typeId` **only** at the single `SliceInfo` construction site when a compact ID is present —
`js/packages/ice/src/Ice/InputStream.js`. Also update the `SliceInfo.typeId` doc
(`js/packages/ice/src/Ice/UnknownSlicedValue.js` and `.d.ts`) to state it is empty when the compact ID is set,
matching C++/Java.

### Testing
compact-ID slice*: the only compact types in the slicing suite are client-known (fully reconstructed, no
`SlicedData`), and every "unknown preserved" server operation returns string-ID types. A true end-to-end test would
require new, coordinated cross-language `.ice` types plus a preserving operation (note: JS generated compact-id
classes auto-register into the global default slice loader, which further complicates a local unit test). Verified
by an adversarial multi-agent review and an independent codex review (single construction site confirmed; factory
resolution untouched); the changed files pass `node --check`. CI builds the package and runs the `slicing` suites.
Happy to add a coordinated cross-language regression test as a follow-up if wanted.

### DIFF
--- a/js/packages/ice/src/Ice/InputStream.js
+++ b/js/packages/ice/src/Ice/InputStream.js
@@ -625,7 +625,9 @@ class EncapsDecoder11 extends EncapsDecoder {
         // Preserve this slice.
         //
         const info = new SliceInfo();
-        info.typeId = this._current.typeId;
+        // When the slice was decoded via a compact type ID, expose an empty type ID; the compact ID is
+        // carried separately by info.compactId (matches C++/Java/Python).
+        info.typeId = this._current.compactId === -1 ? this._current.typeId : "";
         info.compactId = this._current.compactId;
         info.hasOptionalMembers = (this._current.sliceFlags & Protocol.FLAG_HAS_OPTIONAL_MEMBERS) !== 0;
         info.isLastSlice = (this._current.sliceFlags & Protocol.FLAG_IS_LAST_SLICE) !== 0;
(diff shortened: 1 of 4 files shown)


js/packages/ice/src/Ice/InputStream.js
```diff
@@ -625,7 +625,9 @@ class EncapsDecoder11 extends EncapsDecoder {
         // Preserve this slice.
         //
         const info = new SliceInfo();
-        info.typeId = this._current.typeId;
+        // When the slice was decoded via a compact type ID, expose an empty type ID; the compact ID is
+        // carried separately by info.compactId (matches C++/Java/Python).
```
pepone:
```suggestion
        // When the slice was decoded via a compact type ID, expose an empty type ID.
```

bernardnormier (commented):
The fix itself looks correct — it mirrors the Java reference (`_current.compactId == -1 ? _current.typeId : ""`) at the single `SliceInfo` construction site, and re-marshaling is unaffected since `OutputStream.startSlice` writes the compact ID when it's set and ignores `typeId`.

However, I don't think this fix warrants a changelog entry: it's a super-minor behavior alignment in `SliceInfo`, and it's very unlikely any application inspects this API. Please drop `changelog.d/js/5969.md`.

Since the C# and MATLAB companions (#5977, #5976) already merged, could you also remove `changelog.d/csharp/5969.md` and `changelog.d/matlab/5969.md` in this PR?

bernardnormier (approved):
Disregard my previous comment about the changelog fragments — keeping them for consistency with the already-merged C# and MATLAB companions. The fix itself LGTM (matches the Java reference, single construction site, re-marshaling unaffected).

PR #5980, opened 2026-07-14
Memoize failed states in Glacier2 address filter matching

### LINKED ISSUE #5954: Glacier2 address filter matching cost is exponential in the number of wildcard segments
### Summary
The wildcard backtracking introduced by #5897 (not yet in any release) retries each wildcard segment at every later alignment, so the worst-case matching cost grows exponentially with the number of wildcard segments in a rule. The 255-character host cap added in the same PR bounds the host length but not the exponent.

### Mechanism
`AddressRule::matchAddress` recursively retries the matchers that follow a wildcard (`ContainsString`, `ContainsNumberMatch`) at later alignments until one match reaches the end of the host. The same `(matcher index, position)` state can be reached through many different earlier alignments and is re-evaluated each time, so a rule with k wildcard segments against a host of length n can explore O(n^k) alignment combinations.

See [this review benchmark](https://github.com/zeroc-ice/ice/pull/5897#pullrequestreview-4674479307) with a 150-character host: realistic rules (`192.168.*`, `*.example.com`) verify in microseconds, but four wildcard segments take over a second and five or more exceed 15 s — while `RoutingTable::add` holds its mutex across `verify()`.

### Impact
A slow `verify()` stalls the session's routing table and occupies a dispatch thread. The wildcard count is operator-configured and the host must come through an authenticated session, so this is a configuration foot-gun rather than a remote DoS: realistic filter rules are unaffected.

### Suggested fix
Memoize failed `(matcher index, position)` states within a single host check. With hosts capped at 255 characters, this bounds the worst case at O(k·n²), which is negligible, and requires no change to the matching semantics.

Since the backtracking has not shipped in any release, this needs no CHANGELOG entry.

bernardnormier:
The wildcard backtracking introduced by #5897 retries each wildcard segment of an address filter rule at every later alignment of the host, so the worst-case matching cost grows exponentially with the number of wildcard segments in a rule. A slow `verify()` stalls the session's routing table and occupies a dispatch thread.

This PR records failed `(matcher index, position)` states during a host check and skips any state reached again through a different alignment of the preceding wildcards. Since each state is now evaluated at most once and hosts are capped at 255 characters, the worst-case cost is O(k·n²) for a rule with k matchers against a host of length n, with no change to the matching semantics.

The new test case exercises both outcomes through the memoized path: a 255-character host that fails a six-wildcard-segment rule only after every alignment is exhausted (previously minutes to hours of backtracking, now instant), and a host accepted by a multi-wildcard rule.

No CHANGELOG entry since the backtracking has not shipped in any release.

Fixes #5954.

### DIFF
--- a/cpp/src/Glacier2/ProxyVerifier.cpp
+++ b/cpp/src/Glacier2/ProxyVerifier.cpp
@@ -546,7 +546,8 @@ namespace Glacier2
                     return false;
                 }
 
-                if (!matchAddress(host, 0, 0))
+                vector<bool> failed(_addressRules.size() * (host.size() + 1), false);
+                if (!matchAddress(host, 0, 0, failed))
                 {
                     return false;
                 }
@@ -573,9 +574,15 @@ namespace Glacier2
         // Matches host against the matchers at position index and up, starting at position pos in host. The
         // matchers must match the remainder of the host in full. When they don't, this function retries the
         // matchers that can match at a later position (the matchers created for the portion of a rule that
-        // follows a wildcard) until every later alignment is exhausted.
-        [[nodiscard]] bool
-        matchAddress(const string& host, vector<AddressMatcher*>::size_type index, string::size_type pos) const
+        // follows a wildcard) until every later alignment is exhausted. The same (index, pos) state can be
+        // reached through many alignments of the preceding wildcards; failed is a matcher-count by
+        // (host length + 1) matrix that records failed states so each one is evaluated at most once, keeping
+        // the matching cost polynomial in the host length.
+        [[nodiscard]] bool matchAddress(
+            const string& host,
+            vector<AddressMatcher*>::size_type index,
+            string::size_type pos,
+            vector<bool>& failed) const
         {
             if (index == _addressRules.size())
             {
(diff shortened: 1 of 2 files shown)


pepone (approved):
Reviewed the memoization for correctness and traced it against the pre-#5897 backtracking. LGTM.

The key soundness property holds: `matchAddress(host, index, pos)`'s result is a pure function of `(index, pos)` for a fixed host — every `AddressMatcher` subclass is stateless (`match`/`retry` read only const members and mutate only the by-ref `pos`), so caching a *failed* state and skipping it on re-entry cannot change any accept/reject decision. Checked the supporting details too:
- `failed[state] = true` is set only after all `match`/`retry` alignments are exhausted, and failure is cached but success is not — semantics are identical to the un-memoized version.
- `state = index*(host.size()+1)+pos` stays in bounds: `index ∈ [0, size-1]` (the `index == size` case returns before `state` is computed) and no matcher ever advances `pos` past `host.size()`.
- `failed` is allocated per endpoint inside the loop, which is required since the host varies per endpoint.

The new test exercises the full-exhaustion path (255-char host vs. the six-wildcard rule) and an accept, both through the memoized path. 👍

One optional nit, no change needed: the description advertises `O(k·n²)`, but the per-match string ops (`substr` in `MatchesNumber::match`, `find` in `ContainsString`) add an `O(n)` factor, so the honest worst-case is a degree higher — still polynomial and instant for `n ≤ 255`. The code comment's "polynomial in the host length" is accurate as written.

PR #5981, opened 2026-07-14
Reject a proxy when any endpoint matches a Glacier2 address reject rule

bernardnormier:
A `Glacier2.Filter.Address.Reject` rule matched a proxy only when **every** endpoint of the proxy matched the rule, the same all-endpoints-must-match logic used for accept rules. For a reject rule this is backwards: a multi-endpoint proxy was rejected only when the rule covered all its endpoints, so a reject rule did not reliably keep out the hosts it names.

With this PR, a reject rule matches — and the proxy is rejected — as soon as any endpoint of the proxy matches the rule. Accept rules keep the existing semantics: a proxy is acceptable only when every endpoint matches.

`AddressRule::check` now delegates per-endpoint matching to a new `matchEndpoint` helper and aggregates according to the rule's mode (any-endpoint for reject, all-endpoints for accept).

Based on #5980; will rebase once it merges.


bernardnormier:
Closing pending code review; will reopen once reviewed.

PR #5983, opened 2026-07-14
Match Glacier2 address filter host names case-insensitively

bernardnormier:
Fixes a bug in the Glacier2 address filters, `Glacier2.Filter.Address.Accept` and `Glacier2.Filter.Address.Reject`: the filters compared host names case-sensitively, while DNS names are case-insensitive. The rule's address portion is now folded to ASCII lower case when the filter property is parsed, and the endpoint host is folded the same way before matching. IPv4 literals and the port portion of a rule are unaffected.

### DIFF
--- a/changelog.d/service-glacier2/address-filter-case-insensitive.md
+++ b/changelog.d/service-glacier2/address-filter-case-insensitive.md
@@ -1,0 +1,2 @@
+- Fixed a bug in the Glacier2 address filters, `Glacier2.Filter.Address.Accept` and `Glacier2.Filter.Address.Reject`:
+  the filters compared host names case-sensitively. Host names now match regardless of case, like DNS names.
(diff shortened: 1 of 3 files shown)


pepone (approved):
Confirmed the fix is correct — `toLower` only folds ASCII `A`–`Z`, so wildcards, groups, ranges, and digits in a rule are untouched, and address groups are numeric-only anyway.

Minor: the "IP literals … are unaffected" wording isn't quite right — an IPv6 literal like `FE80::1` does get lowercased to `fe80::1`. Harmless/correct (IPv6 hex is case-insensitive and both sides fold the same way), but might be worth rewording to "IPv4 literals and the port portion are unaffected."

bernardnormier:
Reworded the description to "IPv4 literals and the port portion are unaffected." IPv6 literals do get folded, but both the rule and the host fold the same way, so matching is unaffected — and IPv6 literals aren't supported in address filters anyway (tracked separately).

PR #5997, opened 2026-07-15
Clean up DataStorm adapters after constructor failure

### LINKED ISSUE #5891: DataStorm: Instance constructor failure leaves a user-supplied communicator polluted
> Follow-up to a review note on PR #5869 (which narrowed but deliberately did not close this pre-existing gap).

## Summary

When the `Instance` **constructor** throws after `setDefaultObjectAdapter` was called on a **user-supplied**
communicator, the communicator is left polluted: the default object adapter remains set and the already-created
DataStorm object adapters stay alive.

`Node::Node` only reaches its cleanup (`Instance::destroy`, which resets the default object adapter and destroys
the adapters) when `_instance` was assigned — i.e. when the `Instance` constructor completed and `init()` threw.
When the constructor itself throws, the catch falls back to `communicator->destroy()` for an owned communicator
(fine), but does nothing for a user-supplied one.

PR #5869 moved the `DataStorm.Topic.*` default parsing to the top of the constructor (before any adapter is
created), so those property errors are clean; the remaining throw sites after `setDefaultObjectAdapter` are:

- an invalid `DataStorm.Node.Multicast.Endpoints` (`invalid_argument` from the multicast adapter creation),
- an invalid numeric `DataStorm.Node.RetryDelay` / `RetryMultiplier` / `RetryCount` (`PropertyException`),
- the collocated adapter creation.

## Impact

Config-error path only, user-supplied communicator only: after catching the `Node` constructor exception, the
communicator still has a default object adapter set (a subsequent `Node` construction on it fails with
"communicator used to initialize a DataStorm node must not have a default object adapter") and the created
adapters keep their resources until the communicator is destroyed.

## Fix direction

Unwind the partial construction on failure — an RAII guard in the `Instance` constructor that resets the default
object adapter and destroys the created adapters, or move the adapter creation into `init()` so the existing
`Node` constructor cleanup covers it.

pepone:
## Summary
- create the DataStorm object adapters during Instance initialization so constructor failures use the existing destroy path
- parse retry and trace configuration before adapter resources are created
- make Instance destruction tolerate each partial adapter-initialization stage
- add regression coverage for retry-property and multicast-adapter failures with a user-supplied communicator

## Testing
- make -C cpp -j10 test/DataStorm/api_writer
- python3 allTests.py DataStorm/api

Fixes #5891

### DIFF
--- a/changelog.d/datastorm/5891.md
+++ b/changelog.d/datastorm/5891.md
@@ -1,0 +1,2 @@
+- Fixed a bug where a failed `DataStorm::Node` construction left a user-supplied communicator's default object
+  adapter set and its DataStorm object adapters alive, preventing the communicator from being reused.
--- a/changelog.d/datastorm/5891.md
+++ b/changelog.d/datastorm/5891.md
@@ -1,2 +1,2 @@
 - Fixed a bug where a failed `DataStorm::Node` construction left a user-supplied communicator's default object
-  adapter set and its DataStorm object adapters alive, preventing the communicator from being reused.
+  adapter set and its DataStorm object adapters alive.
--- a/changelog.d/datastorm/5891.md
+++ b/changelog.d/datastorm/5891.md
@@ -1,2 +1,0 @@
-- Fixed a bug where a failed `DataStorm::Node` construction left a user-supplied communicator's default object
-  adapter set and its DataStorm object adapters alive.
--- a/cpp/src/DataStorm/Instance.cpp
+++ b/cpp/src/DataStorm/Instance.cpp
@@ -66,10 +66,7 @@ namespace
     }
 }
 
-Instance::Instance(
-    CommunicatorPtr communicator,
-    function<void(function<void()> call)> customExecutor,
-    std::optional<Ice::SSL::ServerAuthenticationOptions> serverAuthenticationOptions)
+Instance::Instance(CommunicatorPtr communicator, function<void(function<void()> call)> customExecutor)
     : _communicator(std::move(communicator))
 {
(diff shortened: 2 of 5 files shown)


changelog.d/datastorm/5891.md
```diff
@@ -0,0 +1,2 @@
+- Fixed a bug where a failed `DataStorm::Node` construction left a user-supplied communicator's default object
+  adapter set and its DataStorm object adapters alive, preventing the communicator from being reused.
```
pepone:
```suggestion
  adapter set and its DataStorm object adapters alive.
```

bernardnormier (commented):
The fix looks correct — moving all communicator-mutating work into `init()` so the existing `Node` catch-path cleanup covers every failure stage is exactly the right approach, and the throw-site ordering in the constructor (all property parses before the `CallbackExecutor`/`Timer` creation) means a parse failure can't leak a started thread either. The new tests nicely verify both failure classes, including that the named server adapter from the failed construction was destroyed.

However, I'd drop the changelog entry. This is hardening/cleanup: realistically, when `Node` construction fails with your own communicator, you don't retry on that communicator — you exit the application and debug the configuration. So the leftover default object adapter isn't something a real application would notice.

bernardnormier (approved):
Approving — see my comment above regarding dropping the changelog entry.

pepone:
🤖 **Claude (on behalf of Jose):** Agreed — dropped the changelog entry (`changelog.d/datastorm/5891.md`). This is hardening/cleanup an application wouldn't notice in practice. Done in 355d453e07.

PR #5998, opened 2026-07-15
Evict replicas whose initialization fails during an IceStorm election

### LINKED ISSUE #5849: [Audit-Medium] IceStorm Observers::init has no eviction — one write-broken but reachable replica loops group formation cluster-wide
### Summary
`Observers::init` rethrows on the first slave failure without evicting it, unlike `Observers::wait` which evicts a failing observer and proceeds as long as a majority remains. A single replica that stays reachable (answers election RPCs) but whose `observerInit` persistently fails (e.g. LMDB `MDB_MAP_FULL`/disk error — reads still work) makes the coordinator recover and re-invite it every round, so the cluster never reaches Normal.

### Mechanism
`Observers::init` ([`Observers.cpp:68-120`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Observers.cpp#L68-L120)) logs and rethrows on any slave's init failure — no evict, no `_reaped`. Contrast `wait()` ([`Observers.cpp:166-203`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Observers.cpp#L166-L203)), which erases the failed observer and only throws below majority. The rethrow reaches `mergeContinue`'s catch around `initMaster` → `recovery()` ([`NodeI.cpp:595-610`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/NodeI.cpp#L595-L610)). The broken node never recovers itself on init failure (its election RPCs and reads still work), so it re-accepts and re-fails without backoff.

### Trigger
One replica with a persistent local write failure while remaining reachable.

### Impact
With a healthy majority available the cluster still never reaches Normal, so even **reads** hang everywhere (`startCachedRead` waits for Normal): one node's disk problem becomes recurring cluster-wide outage churn.

### Suggested fix
Evict minority init failures in `Observers::init` the same way `wait()` does (drop the failing observer, proceed while a majority succeeds).

---

pepone:
`Observers::init` rethrew on the first slave whose `observerInit` failed, which made the coordinator's `initMaster` fail and triggered `recovery()`. A replica that stays reachable (it keeps answering election RPCs) but persistently fails to apply the initial replica state — e.g. a local database write error such as LMDB `MDB_MAP_FULL` while reads still work — therefore made every election fail: the coordinator recovered and re-invited the broken replica each round, the group never reached the Normal state, and since `startCachedRead` waits for Normal, even reads hung on every replica. One node's disk problem became a cluster-wide outage.

`Observers::init` now mirrors what `Observers::wait` already does for failures during replicated updates: a slave whose initialization fails (either sending `initAsync` or completing it) is evicted and reported as reaped, and the election completes as long as a majority of the observers succeeds; below majority it still throws and the coordinator recovers. The node's next `check()` drains the reaped ids and removes those slaves from the group, exactly as for a mid-update eviction.

Two design points worth calling out for review:

- **The evicted replica still receives `ready()`** (it remains in `mergeContinue`'s `tmpSet`) and transitions to Normal without the initial content, so it resumes serving from its stale, un-applied state — entering Normal simply ungates the local cached-read paths (`startCachedRead` waits only for Normal; nothing validates the served generation against applied state). This matches the existing semantics for a slave evicted by `wait()` mid-update. It does **not**, however, keep the group quiescent: `ready()` never cancels the slave's repeated `TimeoutTask`, so once the coordinator's next `check()` reaps it from `_up`, its `areYouThere` poll returns false, it runs `recovery()` and becomes self-coordinated, and the coordinator's merge scan pulls the whole group into a fresh election — where its init fails and it is evicted again. The permanent outage therefore becomes a healthy group punctuated by periodic brief re-elections. Keeping the replica in `ready()` versus excluding it only changes what it does during each window (serving stale cached state versus blocking while stuck in Reorganization), not the churn itself. Eliminating the churn would need invite suppression/backoff for a known-broken replica — filed as a follow-up (#6012).
- **Cold start**: the full-participation gate in `mergeContinue` runs before `initMaster`, so with an eviction during init a cold-start cluster can reach Normal with N−1 slaves. Latest-data selection is unaffected — the evicted replica did participate in the election and its LLU was considered in the max-LLU computation; it only failed to apply the new state.

Tested with `IceStorm/rep1` (replicated) and `IceStorm/single` (all green). No new test: injecting a persistent local write failure into one replica isn't feasible in the existing harness, and the healthy-path election behavior is covered by `rep1`.

Fixes #5849.

### DIFF
--- a/changelog.d/service-icestorm/5849.md
+++ b/changelog.d/service-icestorm/5849.md
@@ -1,5 +1,5 @@
-- Fixed a highly available IceStorm availability issue: when a replica stayed reachable but persistently failed to
-  apply the initial replica state (for example on a local database write error), every election failed and the
-  whole group kept cycling through recovery, so no replica served requests. The coordinator now evicts a replica
+- Fixed an availability issue in a replicated IceStorm deployment: when a replica stayed reachable but persistently
+  failed to apply the initial replica state (for example on a local database write error), every election failed and
+  the whole group kept cycling through recovery, so no replica served requests. The coordinator now evicts a replica
   whose initialization fails and completes the election as long as a majority of the replicas succeeds, as it
   already did for failures during replicated updates.
--- a/changelog.d/service-icestorm/5849.md
+++ b/changelog.d/service-icestorm/5849.md
@@ -1,0 +1,5 @@
+- Fixed a highly available IceStorm availability issue: when a replica stayed reachable but persistently failed to
+  apply the initial replica state (for example on a local database write error), every election failed and the
+  whole group kept cycling through recovery, so no replica served requests. The coordinator now evicts a replica
+  whose initialization fails and completes the election as long as a majority of the replicas succeeds, as it
+  already did for failures during replicated updates.
--- a/cpp/src/IceStorm/Observers.cpp
+++ b/cpp/src/IceStorm/Observers.cpp
@@ -96,26 +96,52 @@ Observers::init(const set<GroupNodeInfo>& slaves, const LogUpdate& llu, const To
                 Ice::Trace out(_traceLevels->logger, _traceLevels->replicationCat);
                 out << "error calling init on " << slave.id << ", exception: " << ex;
             }
-            throw;
+
+            lock_guard<mutex> reapedLock(_reapedMutex);
+            _reaped.push_back(slave.id);
         }
     }
 
-    for (auto& o : observers)
+    // Like the replicated update operations (see wait()), the initialization proceeds as long as a majority of the
+    // slaves succeeds: a slave whose init fails is evicted and reported as reaped, and the node removes it from the
+    // group on its next check(). Failing the whole initialization instead would make every election fail when a
+    // single reachable replica persistently cannot apply the initial state (for example on a local database write
+    // error), keeping the group from ever reaching the normal state.
+    auto p = observers.begin();
+    while (p != observers.end())
     {
         try
         {
-            o.future.get();
+            p->future.get();
         }
         catch (const Ice::Exception& ex)
         {
             if (_traceLevels->replication > 0)
             {
                 Ice::Trace out(_traceLevels->logger, _traceLevels->replicationCat);
-                out << "init on " << o.id << " failed with exception " << ex;
+                out << "init on " << p->id << " failed with exception " << ex;
             }
-            throw;
+            int id = p->id;
+            p = observers.erase(p);
+
+            lock_guard<mutex> reapedLock(_reapedMutex);
+            _reaped.push_back(id);
+            continue;
         }
+        ++p;
     }
+
+    // If we don't have the majority of observers we raise.
+    if (observers.size() < _majority)
+    {
+        if (_traceLevels->replication > 0)
+        {
+            Ice::Trace out(_traceLevels->logger, _traceLevels->replicationCat);
+            out << "number of observers '" << observers.size() << "' is less than the majority '" << _majority << "'";
+        }
+        throw Ice::UnknownException(__FILE__, __LINE__, "too few observers");
+    }
+
     _observers = std::move(observers);
 }
 


changelog.d/service-icestorm/5849.md
```diff
@@ -0,0 +1,5 @@
+- Fixed a highly available IceStorm availability issue: when a replica stayed reachable but persistently failed to
+  apply the initial replica state (for example on a local database write error), every election failed and the
+  whole group kept cycling through recovery, so no replica served requests. The coordinator now evicts a replica
+  whose initialization fails and completes the election as long as a majority of the replicas succeeds, as it
+  already did for failures during replicated updates.
```
bernardnormier: "highly available IceStorm availability issue" reads awkwardly — suggestion:
```suggestion
- Fixed an availability issue in a replicated IceStorm deployment: when a replica stayed reachable but persistently
  failed to apply the initial replica state (for example on a local database write error), every election failed and
  the whole group kept cycling through recovery, so no replica served requests. The coordinator now evicts a replica
  whose initialization fails and completes the election as long as a majority of the replicas succeeds, as it
  already did for failures during replicated updates.
```

bernardnormier (commented):
The code change itself looks correct: it faithfully mirrors `wait()`/`check()` (same erase-and-reap pattern, same majority threshold, same below-majority `UnknownException` with the unchanged `initMaster` → `recovery()` handling), the locking matches the existing paths, and the cold-start reasoning holds — including that when the broken replica holds the max LLU, the coordinator syncs *from* it before init, so no data is lost.

However, I don't think the first design point's justification holds up. Keeping the evicted replica in `ready()` doesn't avoid the periodic cycle — the cycle happens either way:

1. The slave's `TimeoutTask` is `scheduleRepeated` at invitation-accept (`NodeI.cpp:767-771`) and `ready()` never cancels it, so it keeps polling the coordinator in the Normal state.
2. The coordinator's first `check()` (about `ElectionTimeout` later) drains the reaped ids and erases the evicted slave from `_up`.
3. `areYouThere` requires `_up.find(j) != _up.end()` (`NodeI.cpp:916-920`), so it now returns false → the evicted replica runs `recovery()` and forms a self-coordinated group.
4. The coordinator's merge scan finds it via `areYouCoordinator` → `MergeTask` → group-wide re-election → its `observerInit` fails again → evicted again — repeating every few tens of seconds with the default timeouts.

So with this fix the permanent outage becomes: the group serves normally, punctuated by periodic brief re-elections, with the broken replica serving *stale* reads during each window (`ready()` sets its `_generation`, so cached-read generation checks pass despite the un-applied state). That's still a major improvement and I think the right patch-level behavior — the churn is inherent unless we add invite suppression/backoff for a known-broken replica, which would be protocol surgery. But please double-check this analysis, and if it's right, correct the PR description (the evicted replica doesn't quietly serve "until a future election"; and excluding it from `ready()` would produce the same cycle, only with the replica blocking instead of serving stale reads during the window). A follow-up issue for invite backoff may be worth filing.

changelog.d/service-icestorm/5849.md
pepone: 🤖 **Claude:** Applied your suggested wording verbatim. Done in 0871d31b77.

pepone:
🤖 **Claude:** You're right, and I've corrected the PR description. I re-traced it and cross-checked with a second reviewer (codex, xhigh) — your chain holds at every step:

- `TimeoutTask` is `scheduleRepeated` in `invitation()` (`NodeI.cpp:767-770`) and is cancelled only in `recovery()` (`1009`) and `destroy()` (`1041`) — never in `ready()`. So the evicted slave keeps polling in Normal.
- `timeout()` has no `Normal`/`Reorganization` guard (`301-345`); once `check()` reaps it from `_up` (`185-198`), `areYouThere` (`916-919`) returns false → `recovery()` → self-coordinated → merge scan (`areYouCoordinator` is true even while `Inactive`) → group-wide re-election → init fails again → evicted again. The churn is inherent, and excluding it from `ready()` produces the same cycle (only with the replica blocking in Reorganization instead of serving during each window).

Two refinements from the cross-check, both now folded into the description:

1. My "cached-read generation checks pass" reasoning (which I lifted into the description) is wrong. `startCachedRead` waits only for `NodeStateNormal` and copies `_generation` with **no** comparison against applied/db state (`1130-1142`). The generation-validation helper `CachedReadHelper::observerPrecondition` (`NodeI.h:145`) has **no callers anywhere**. So the stale-serving isn't a passed check — entering Normal just ungates the cached paths outright.
2. "Serves stale reads" undersells it: Normal ungates everything behind `startCachedRead` — cached subscriber/publish (`Subscriber.cpp:30-43`) and mutating calls forwarded via `_coordinatorProxy` (`1084-1095`), not only reads.

Filed the invite-backoff follow-up as #6012 (also flags the dead `observerPrecondition`). Agreed this is the right patch-level behavior; the churn removal is protocol surgery for that issue.

bernardnormier (approved):

PR #6000, opened 2026-07-15
Drop invalid unquoted O'Reilly form from getPropertyAsList doc

### LINKED ISSUE #5995: getPropertyAsList doc example lists an unquoted O'Reilly that the parser rejects
The `getPropertyAsList` / `getPropertyAsListWithDefault` doc comment gives three ways to write the value `O'Reilly`:

> you can escape the quote in question with `\`, e.g. O'Reilly can be written as O'Reilly, "O'Reilly" or 'O\'Reilly'.

The first form — bare, unquoted `O'Reilly` — is wrong. `IceInternal::splitString` (`cpp/src/Ice/StringUtil.cpp:647-650`) treats the apostrophe as an opening quote: it is consumed (dropped), the rest is accumulated, and the trailing unmatched quote makes the parse fail. So bare `O'Reilly` does not yield the string `O'Reilly` — it parses to `OReilly` with a mismatched-quote failure. Only the quoted forms `"O'Reilly"` and `'O\'Reilly'` are valid.

**Fix** — drop the bare unquoted form from the example, e.g. `O'Reilly can be written as "O'Reilly" or 'O\'Reilly'.`

The identical example text appears in every mapping's `Properties` doc:

- C++ — `cpp/include/Ice/Properties.h`
- C# — `csharp/src/Ice/Properties.cs`
- Java — `java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java`
- JS — `js/packages/ice/src/Ice/Properties.d.ts`
- Python — `python/python/Ice/Properties.py`
- Swift — `swift/src/Ice/Properties.swift`
- MATLAB — `matlab/lib/+Ice/Properties.m`

(Ruby and PHP have no public `Properties` doc comment.) Found as a follow-up to the MATLAB API doc audit (#5971).

pepone:
The `getPropertyAsList` / `getPropertyAsListWithDefault` doc comment listed a bare, unquoted `O'Reilly` as one of three valid ways to write the value. That form is wrong: `IceInternal::splitString` (`cpp/src/Ice/StringUtil.cpp:647-650`) treats the apostrophe as an opening quote, so it is dropped and the trailing unmatched quote fails the parse — bare `O'Reilly` yields `OReilly`, not `O'Reilly`. Only the quoted forms `"O'Reilly"` and `'O\'Reilly'` are valid.

This drops the bare form from the example in every mapping's `Properties` doc (C++, C#, Java, JS, Python, Swift, MATLAB). Ruby and PHP have no public `Properties` doc comment. The MATLAB comment wrapped mid-list, so its two lines are reflowed to stay within the column limit.

Doc-comment-only change; no behavior change.

Fixes #5995

### DIFF
--- a/js/packages/ice/src/Ice/Properties.d.ts
+++ b/js/packages/ice/src/Ice/Properties.d.ts
@@ -114,7 +114,7 @@ declare module "@zeroc/ice" {
              * If the property is not set, the default list is returned. The strings in the list can contain whitespace
              * and commas if they are enclosed in single or double quotes. If quotes are mismatched, the default list
              * is returned. Within single or double quotes, the respective quote character can be escaped with a
-             * backslash. For example, O'Reilly can be written as "O'Reilly", or 'O\'Reilly'.
+             * backslash. For example, O'Reilly can be written as "O'Reilly" or 'O\'Reilly'.
              *
              * @param key - The property key.
              * @param value - The default value to use if the property is not set.
--- a/js/packages/ice/src/Ice/Properties.d.ts
+++ b/js/packages/ice/src/Ice/Properties.d.ts
@@ -85,7 +85,7 @@ declare module "@zeroc/ice" {
              * Gets a property as a list of strings. The strings must be separated by whitespace or comma. The strings
              * in the list can contain whitespace and commas if they are enclosed in single or double quotes. If quotes
              * are mismatched, an empty list is returned. Within single quotes or double quotes, you can escape the
-             * quote in question with a backslash, e.g. O'Reilly can be written as O'Reilly, "O'Reilly" or 'O\'Reilly'.
+             * quote in question with a backslash, e.g. O'Reilly can be written as "O'Reilly" or 'O\'Reilly'.
              *
              * @param key - The property key.
              * @returns The property value interpreted as a list of strings, or an empty list if the property is not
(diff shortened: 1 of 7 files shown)


js/packages/ice/src/Ice/Properties.d.ts
```diff
@@ -114,7 +114,7 @@ declare module "@zeroc/ice" {
              * If the property is not set, the default list is returned. The strings in the list can contain whitespace
              * and commas if they are enclosed in single or double quotes. If quotes are mismatched, the default list
              * is returned. Within single or double quotes, the respective quote character can be escaped with a
-             * backslash. For example, O'Reilly can be written as O'Reilly, "O'Reilly", or 'O\'Reilly'.
+             * backslash. For example, O'Reilly can be written as "O'Reilly", or 'O\'Reilly'.
```
bernardnormier: Leftover serial comma from the three-item list — the other sites read `"O'Reilly" or`:
```suggestion
             * backslash. For example, O'Reilly can be written as "O'Reilly" or 'O\'Reilly'.
```

bernardnormier (approved):
Verified against `splitString`: a quote char opens quoting even mid-word, so bare `O'Reilly` starts a single-quoted section that never closes and the parse fails — the bare form was never valid. A repo-wide grep confirms all 7 files with this doc comment are covered (PHP and Ruby have none). One tiny inline suggestion for a leftover serial comma in the JS `.d.ts`.

PR #6002, opened 2026-07-15
Correct OptionalFormat.Size doc comment: only enums use the Size format

### LINKED ISSUE #5993: OptionalFormat.Size doc says "Used by enums, class identifiers, etc." — only enums use the Size format
The doc comment on the `Size` optional format across mappings reads:

> "Size encoding" using either 1 or 5 bytes. Used by enums, class identifiers, etc.

Only **enums** marshal a tagged/optional member with the `Size` optional format. In the Slice compiler, `Enum::getOptionalFormat()` is the only type that returns `Size` (`cpp/src/Slice/Parser.cpp`); every other type returns `F1`/`F2`/`F4`/`F8`/`VSize`/`FSize`/`Class`. The runtime `GetOptionalFormat` template likewise maps only the enum category to `OptionalFormat::Size` (`cpp/include/Ice/StreamHelpers.h`). A "class identifier" is not a distinct, optional-markable Slice type, so the "class identifiers, etc." example is misleading. Raised by @bernardnormier in #5971, who confirmed only enums use this encoding.

Correct the wording to "Used by enums." in each mapping that still carries it:

- **C++** — `cpp/include/Ice/StreamableTraits.h:71`
- **C#** — `csharp/src/Ice/OptionalFormat.cs:35`
- **Java** — `java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/OptionalFormat.java:19`
- **JS** — `js/packages/ice/src/Ice/OptionalFormat.d.ts:30`
- **Swift** — `swift/src/Ice/OptionalFormat.swift:18`

Already fixed in **MATLAB** (`matlab/lib/+Ice/OptionalFormat.m`) in #5971. **Python, Ruby, and PHP** don't expose a public `OptionalFormat` doc comment (the type is only referenced internally in their C++ extensions — `IcePy`/`IceRuby`/`IcePHP` `Types.cpp`), so no change is needed there.

Found as a follow-up to the MATLAB API doc audit; the "class identifiers" wording was inherited from the C++ reference (`StreamableTraits.h`) and adopted verbatim by each mapping's doc audit.

pepone:
The doc comment on the `Size` optional format read:

> "Size encoding" using either 1 or 5 bytes. Used by enums, class identifiers, etc.

Only **enums** marshal a tagged/optional member with the `Size` optional format — `Enum::getOptionalFormat()` is the only Slice type that returns `Size`, and the runtime `GetOptionalFormat` template maps only the enum category to `OptionalFormat::Size`. A "class identifier" is not a distinct, optional-markable Slice type, so the "class identifiers, etc." example is misleading.

Corrects the wording to "Used by enums." in each mapping that still carried it (C++, C#, Java, JS, Swift). MATLAB was already fixed in #5971; Python/Ruby/PHP expose no public `OptionalFormat` doc comment.

Fixes #5993

### DIFF
--- a/cpp/include/Ice/StreamableTraits.h
+++ b/cpp/include/Ice/StreamableTraits.h
@@ -68,7 +68,7 @@ namespace Ice
         /// Fixed 8-byte encoding.
         F8 = 3,
 
-        /// "Size encoding" using either 1 or 5 bytes. Used by enums, class identifiers, etc.
+        /// "Size encoding" using either 1 or 5 bytes. Used by enums.
         Size = 4,
 
         /// Variable "size encoding" using either 1 or 5 bytes followed by data.
--- a/csharp/src/Ice/OptionalFormat.cs
+++ b/csharp/src/Ice/OptionalFormat.cs
@@ -32,7 +32,7 @@ public enum OptionalFormat
     F8 = 3,
 
     /// <summary>
-    /// "Size encoding" using either 1 or 5 bytes. Used by enums, class identifiers, etc.
+    /// "Size encoding" using either 1 or 5 bytes. Used by enums.
     /// </summary>
     Size = 4,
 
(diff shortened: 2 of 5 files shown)


InsertCreativityHere (approved):
Looks good to me!

And a good catch that this is enum-specific now!

PR #6003, opened 2026-07-15
Stop using the no-op endpoint -t timeout option

### LINKED ISSUE #5979: Stop using the no-op endpoint -t timeout option
Since Ice 3.8 removed per-endpoint timeouts, the `-t timeout` endpoint option is a no-op: the parsers still accept it and the stringification round-trips it for backward compatibility, but nothing consumes the value at runtime (`TcpConnector` uses it only for connector comparison; the connection establishment timeout comes from `Ice.Connection.Client.ConnectTimeout`).

We should not append this useless option anywhere ourselves. A scan of the non-test code found two remaining places:

1. `icestormadmin` appends `-t 3000` to the `IceStorm/Finder` fallback endpoints, with a comment claiming a 3s connection timeout:
https://github.com/zeroc-ice/ice/blob/8ed7832bcc4a49295575d5421c636fc743bd4b06/cpp/src/IceStorm/Admin.cpp#L162-L163
(being addressed in the #5970 review: https://github.com/zeroc-ice/ice/pull/5970#discussion_r3581059509)

2. The IceGrid GUI adapter-endpoints tooltip presents `-t` as a working example — "`tcp -t 10000 (sets a timeout of 10,000 milliseconds)`" — which misinforms users:
https://github.com/zeroc-ice/ice/blob/8ed7832bcc4a49295575d5421c636fc743bd4b06/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Application/AdapterEditor.java#L125

The endpoint parsers and the `options()`/`toString` round-trip of a parsed `-t` (including `-t infinite`) are intentional backward compatibility and stay as-is.

Note that removing the option is not always the whole fix: where a shorter timeout was actually intended — as in the `icestormadmin` case, where a 3s probe for a possibly-absent Finder beats the 10s default — the replacement is to reduce the connect timeout, e.g. `Ice.Connection.Client.ConnectTimeout=3` (seconds), rather than to silently fall back to the default.

pepone:
Fixes #5979.

The endpoint `-t <timeout>` option has had no runtime effect since Ice 3.8 — it only influences connector comparison, not the connect timeout (which comes from `Ice.Connection.Client.ConnectTimeout`). We shouldn't append or advertise it ourselves. Two remaining non-test places used it:

**icestormadmin** — #5970 already dropped the bogus `-t 3000` from the `IceStorm/Finder` fallback endpoints, but that left the connect timeout at the 10s default. This restores the intended fast probe by defaulting `Ice.Connection.Client.ConnectTimeout` to 3s, so an unreachable manager/Finder fails quickly instead of hanging. The set is guarded, so a user's explicit `--Ice.Connection.Client.ConnectTimeout=…` (command line or config file) still wins.

**IceGrid GUI** — the object-adapter endpoints tooltip presented `tcp -t 10000 (sets a timeout of 10,000 milliseconds)` as a working example, which misinforms users. Replaced with `tcp -p 10000 (listen on a specific port)`.

The endpoint parsers and the `-t` round-trip in `toString()`/`options()` (including `-t infinite`) are intentional backward compatibility and stay as-is.

### DIFF
--- a/changelog.d/service-icestorm/5979.md
+++ b/changelog.d/service-icestorm/5979.md
@@ -1,3 +1,0 @@
-- Lowered the `icestormadmin` connect timeout to 3 seconds so the tool no longer waits the full 10-second
-  default when a topic manager or the `IceStorm/Finder` object is unreachable. Set
-  `Ice.Connection.Client.ConnectTimeout` to override it.
--- a/changelog.d/service-icestorm/5979.md
+++ b/changelog.d/service-icestorm/5979.md
@@ -1,0 +1,3 @@
+- Lowered the `icestormadmin` connect timeout to 3 seconds so the tool no longer waits the full 10-second
+  default when a topic manager or the `IceStorm/Finder` object is unreachable. Set
+  `Ice.Connection.Client.ConnectTimeout` to override it.
--- a/cpp/src/IceStorm/Admin.cpp
+++ b/cpp/src/IceStorm/Admin.cpp
@@ -35,6 +35,12 @@ main(int argc, char* argv[])
 
         auto properties = Ice::createProperties(argc, argv, "IceStormAdmin");
         properties->setProperty("Ice.Warn.Endpoints", "0");
+        // Give up quickly when a topic manager or the IceStorm/Finder fallback is unreachable, instead of waiting
+        // the full default connect timeout (10s). Only lower it when the user hasn't configured it explicitly.
+        if (properties->getProperty("Ice.Connection.Client.ConnectTimeout").empty())
+        {
+            properties->setProperty("Ice.Connection.Client.ConnectTimeout", "3"); // seconds
+        }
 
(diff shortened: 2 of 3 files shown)


bernardnormier (approved):
The changes look good: the `getProperty().empty()` guard is the correct explicit-set test (command line and config file are both absorbed by `createProperties` at that point), "3" matches the seconds-based 3.8 property, and I verified the tooltip line was the last constructed `-t` in any mapping's non-test sources.

However, I'd drop the changelog entry. The connect timeout only matters when the connection attempt hangs (filtered or unreachable host). Typically you get an exception immediately when nothing listens on the target host/port — especially on a local network, the main IceStorm/icestormadmin use case. So the 3s vs 10s difference is an edge case users are unlikely to notice.

pepone:
🤖 **Claude:** Dropped the changelog fragment as suggested — done in 28a2c31e5f. Thanks for the review!

Issue #5864, opened 2026-07-07
[Audit-Low, grouped] Glacier2: 13 low-severity correctness/robustness findings


pepone:
### Summary
Thirteen low-severity correctness/robustness findings from the 2026-07-07 Glacier2 correctness audit, grouped for triage. Each is independently minor; none is a regression unless noted. Commit-pinned to `b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b`.

---

**1. PR #5648 parity gap: other AMI proxy calls lack the synchronous-CommunicatorDestroyedException guard.**
PR #5648 wrapped only `checkPermissionsAsync`/`authorizeAsync`. The same synchronous throw is unguarded on `_sessionManager->createAsync` ([`SessionRouterI.cpp:173`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Glacier2/SessionRouterI.cpp#L173), `:285`), `session->destroyAsync` inside `sessionCreated` ([`:453`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Glacier2/SessionRouterI.cpp#L453)), and `RouterI::destroy` ([`RouterI.cpp:82-87`](https://github.com/zeroc-ice/ice/blob/b71bc8ed0ad0baff8ecc784614f42ca9d15a5a0b/cpp/src/Glacier2/RouterI.cpp#L82-L87)). A sync CDE from the manager-create call escapes into the AMI response callback and reproduces the #5648 loss (dispatch never answered, `_pending` leaked). Narrow window (communicator destruction racing an in-flight create). *Fix:* extend the try/catch to these calls. *(Audit finding 9.)*

**2. `SessionControl::destroy` surfaces an undeclared `SessionNotExistException` as `UnknownUserException`.**
[...]

PR #6007, opened 2026-07-15
Guard Glacier2 session destroy calls against communicator destruction

bernardnormier:
Addresses items 1 and 4 of #5864.

**Item 1 (remainder):** `Communicator::destroy` drains in-flight AMI completions and fires connection close callbacks; a proxy invocation made from one of those callbacks throws `CommunicatorDestroyedException` synchronously. Two Glacier2 session-destroy calls sit on such paths and were unguarded:

- `RouterI::destroy`'s `_session->destroyAsync`, reachable from the client connection's close callback and from `finishCreateSession`'s late-create teardown. The exception is now caught and ignored — the session is destroyed along with the communicator.
- `CreateSession::sessionCreated`'s cleanup `session->destroyAsync`, which runs from the session manager's AMI response callback. Same treatment, consistent with its fire-and-forget semantics.

The `createAsync` calls listed in the finding are already covered: since #5984, both `createSession` implementations run inside `authorized`'s try/catch.

Also, `sessionDestroyException` no longer traces `CommunicatorDestroyedException`: it has exactly one cause (router shutdown with sessions still active), so the trace only added noise at `Glacier2.Trace.Session=1`.

**Item 4:** `RouterI::destroy` caught `ObjectAdapterDeactivatedException` around `remove(_controlId)`, but in 3.8 `remove` throws the sibling class `ObjectAdapterDestroyedException` — the catch could never match. It now catches the correct type.

### DIFF
--- a/cpp/src/Glacier2/RouterI.cpp
+++ b/cpp/src/Glacier2/RouterI.cpp
@@ -72,7 +72,7 @@ Glacier2::RouterI::destroy(function<void(exception_ptr)> error)
             catch (const NotRegisteredException&)
             {
             }
-            catch (const ObjectAdapterDeactivatedException&)
+            catch (const ObjectAdapterDestroyedException&)
             {
                 //
                 // Expected if the router has been shutdown.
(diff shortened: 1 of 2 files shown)


pepone (approved):
LGTM. Verified both items against the source:

**Item 4** — `remove()` → `removeFacet()` → `checkForDestruction()` throws only `ObjectAdapterDestroyedException`, and it's a `final` sibling of `ObjectAdapterDeactivatedException` (both directly under `LocalException`), so the old catch was genuinely dead. Correct type now.

**Item 1** — the guards do real work, not no-ops:
- `RouterI::destroy`: catching the sync CDE keeps `_routingTable->destroy()` running and stops the shutdown loop in `SessionRouterI::destroy` from aborting mid-iteration.
- `sessionCreated`: this is the one that matters — a sync CDE escaping the session-manager AMI response callback would reproduce #5648 (dispatch unanswered, `_pending` leaked). Swallowing it lets `unexpectedCreateSessionException` answer the dispatch. Confirmed `current_exception()` still resolves to the outer router-creation exception.
- `sessionDestroyException`: covers the async CDE path and drops a noise-only trace; catch order is correct.

Also confirmed the `createAsync` sites are already covered by #5984's `authorized` try/catch, so item 1 is fully closed.

PR #6008, opened 2026-07-15
Make Glacier2 SessionControl::destroy idempotent

bernardnormier:
Addresses item 2 of #5864.

`SessionControlI::destroy` called `SessionRouterI::destroySession` with no try/catch. When an application-initiated `SessionControl::destroy` races another teardown of the same session (the client connection closing, the client calling `Router::destroySession`, or a concurrent double-destroy), the losing dispatch finds the session entry already gone and `destroySession` throws `SessionNotExistException` — which `SessionControl::destroy` doesn't declare, so the application receives `UnknownUserException` and the filter/connection cleanup in the servant is skipped.

`SessionControlI::destroy` now catches `SessionNotExistException` and treats it as a no-op, making destroy idempotent: the race resolves silently and the local cleanup always runs.

Note the purely sequential case doesn't reach this: router-initiated teardown removes the SessionControl servant before invoking the application session's `destroyAsync`, so a `sessionControl->destroy()` call made from the session's own destroy implementation gets `ObjectNotExistException`, which the existing error handler already ignores.

### DIFF
--- a/cpp/src/Glacier2/SessionRouterI.cpp
+++ b/cpp/src/Glacier2/SessionRouterI.cpp
@@ -71,24 +71,31 @@ namespace Glacier2
 
         void destroy(const Current&) final
         {
-            _sessionRouter->destroySession(
-                _connection,
-                [defaultExceptionHandler = _sessionRouter->defaultSessionDestroyExceptionHandler()](exception_ptr e)
-                {
-                    try
+            try
+            {
+                _sessionRouter->destroySession(
+                    _connection,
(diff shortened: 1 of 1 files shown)


pepone (approved):
Correct and well-targeted. `destroySession` throws `SessionNotExistException` synchronously when the session was already removed by a racing teardown, and since `SessionControl::destroy` doesn't declare it, catching it here makes destroy idempotent and ensures the local `_filters->destroy()` / `_connection->close()` cleanup always runs. Placement is right (wraps only the throwing call), it's consistent with the connection-close teardown path that already swallows these races, and `Router::destroySession` correctly still propagates `SessionNotExistException` since it declares it. LGTM.

Issue #6010, opened 2026-07-15
getPropertyAsList returns the default value on mismatched quotes instead of failing


bernardnormier:
In all language mappings, `getPropertyAsList` / `getPropertyAsListWithDefault` / `getIcePropertyAsList` handle a property value with mismatched quotes by logging a warning and returning the default value (or an empty list). This is documented behavior, but it silently replaces a malformed configuration value with a completely different one — the caller cannot distinguish 'property not set' from 'property set to a malformed value'.

A malformed property value should fail with a `PropertyException` naming the property, consistent with how `getPropertyAsInt` rejects a non-integer value.

This is a documented-behavior change across every mapping, so it belongs in 3.9. Related: #6009 (Glacier2 filter properties currently work around this with custom parsing that rejects unmatched quotes at startup).

bernardnormier:
Tracked under #6058, which generalizes this to failing with a `PropertyException` on any invalid property value across all mappings. Keeping this open as the specific `getPropertyAsList` mismatched-quotes instance.

Issue #6009, opened 2026-07-15
Switch Glacier2 filter properties to standard list parsing


bernardnormier:
The Glacier2 session filter properties (`Glacier2.Filter.Category.Accept`, `Glacier2.Filter.AdapterId.Accept`, `Glacier2.Filter.Identity.Accept`) use custom space-delimited parsing in `cpp/src/Glacier2/Instance.cpp`, while every other Ice list property is read with `getIcePropertyAsList`, which also accepts ',' as a delimiter.

Switch these properties to the standard list parsing for consistency. This is a (minor) syntax change — a ',' in a filter value becomes a delimiter instead of part of the value — so it belongs in a feature release, not a patch release.

Note: `getPropertyAsListWithDefault` returns the default value with a warning on an unmatched quote. In 3.8.x, Glacier2 rejects an unmatched quote in these properties at startup; the switch should preserve that behavior rather than regress to warn-and-default — #6010 tracks fixing this in the property API itself.

PR #6011, opened 2026-07-15
Reject malformed Glacier2 filter properties at startup

bernardnormier:
Addresses item 8 of #5864.

The Glacier2 session filter properties were parsed leniently, so an admin typo produced a filter that silently differed from the configuration:

- An unmatched quote in `Glacier2.Filter.Category.Accept` or `Glacier2.Filter.AdapterId.Accept` was silently accepted (`splitString`'s failure return was ignored).
- An unmatched quote in `Glacier2.Filter.Identity.Accept` produced an identity containing the opening quote character, which never matches anything.
- A `Glacier2.Filter.Category.AcceptUser` value other than 0, 1, or 2 silently disabled the feature.

The router now refuses to start in all three cases, with an error naming the property. Valid configurations parse exactly as before — this only rejects input that previously mis-parsed silently.

We keep the custom space-delimited parsing rather than switching to `getIcePropertyAsList`: it returns the default value on an unmatched quote instead of failing, and it treats ',' as a delimiter, which would change the syntax of these properties in a patch release. #6009 tracks switching to the standard list parsing in 3.9, and #6010 tracks fixing the unmatched-quote behavior of the property API itself.

### DIFF
--- a/changelog.d/service-glacier2/filter-property-unmatched-quote.md
+++ b/changelog.d/service-glacier2/filter-property-unmatched-quote.md
@@ -1,3 +1,0 @@
-- The Glacier2 router now refuses to start when a session filter property contains an unmatched quote, or when
-  `Glacier2.Filter.Category.AcceptUser` is set to a value other than 0, 1, or 2. Previously, such values were
-  silently accepted and could produce filters that did not match the intended configuration.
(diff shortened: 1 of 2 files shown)


pepone (approved):
LGTM. Traced all three parsing paths and the fix is correct and well-scoped:

- `parseSpaceDelimitedProperty` now honors `splitString`'s return, which is `false` if and only if a quote is left unmatched — so the "unmatched quote" diagnostic is accurate.
- The identity path previously folded the opening quote into the identity (so it never matched anything); throwing and wrapping via the existing `parseFilterIdentities` try/catch is the right fix.
- The `AcceptUser` range check matches the only consumer (`FilterManager.cpp`, which acts on `>0`/`==1`/`==2`).

Confirmed no regression in the `staticFiltering` suite — every quoted test case is balanced (including `"a\"nother"`) and `AcceptUser` is only ever `2`, so nothing that previously started now fails. Since `Instance` is constructed at router startup, the throws correctly abort startup with a property-named message.

Only optional follow-up is the negative-test coverage Copilot flagged; fine to defer given the harness only expects successful router starts today.

PR #6013, opened 2026-07-15
Use sizeof(s) as the read count in pipe-draining loops

### LINKED ISSUE #5917: Replace magic-number read-buffer size with sizeof(s) in Activator/Service pipe draining
**Priority: low** (readability / latent-footgun cleanup, no observed bug.)

The POSIX pipe-draining code in the Activator (and one copy in `Ice/Service.cpp`) reads a variable-length message in fixed-size chunks using a bare `16` literal for both the buffer size and the `read()` count:

```cpp
char s[16];
ssize_t rs;
string message;
while ((rs = read(fd, &s, 16)) > 0)
{
    message.append(s, static_cast<size_t>(rs));
}
```

The `16` has no semantic meaning — it's just an arbitrary read-chunk size. The message can be any length; the child writes it whole (`write(fd, msg.c_str(), msg.size())` in `reportChildError`) and the reader drains it a chunk at a time into a `std::string`. So the magic number makes the intent harder to read than it should be.

More importantly, the buffer size (`char s[16]`) and the read count (`read(..., 16)`) are two independent literals that must be kept in sync by hand. If they ever drift, the result is either a silently truncated read (count < buffer) or a **stack buffer overflow** (count > buffer). Tying the count to the buffer removes both the magic number and the footgun:

```cpp
char s[16];
while ((rs = read(fd, s, sizeof(s))) > 0)
{
    message.append(s, static_cast<size_t>(rs));
}
```

`sizeof(s)` is self-documenting and can't diverge from the declaration. (Aside: `&s` is `char (*)[16]`; `s` or `&s[0]` is the conventional spelling for the `void*` argument.)

### Sites (same copy-pasted idiom)

- `cpp/src/IceGrid/Activator.cpp:1349` — termination-listener status-pipe read
- `cpp/src/IceGrid/Activator.cpp:871` — child-side error-pipe read after fork
- `cpp/src/Ice/Service.cpp:1516` — daemon startup pipe read

Predates any recent change (the idiom dates back to ~2009); noticed while reviewing #5911, which touched the first site.

### Suggested fix

Replace the literal read count with `sizeof(s)` at all three sites. Optionally bump the buffer to something larger (e.g. 256) to reduce syscalls, though the messages are short so it's not important.

bernardnormier:
The POSIX pipe-draining loops in the IceGrid Activator and in `Ice::Service` read into a `char s[16]` buffer using a separate bare `16` literal as the `read()` count. This PR ties the count to the buffer with `sizeof(s)`, so the two can't drift apart, and replaces the unconventional `&s` argument with `s`.

No behavior change.

Fixes #5917.

### DIFF
--- a/cpp/src/Ice/Service.cpp
+++ b/cpp/src/Ice/Service.cpp
@@ -1513,7 +1513,7 @@ Ice::Service::runDaemon(int argc, char* argv[], InitializationData initData)
             ssize_t rs;
             char s[16];
             string message;
-            while ((rs = read(fds[0], &s, 16)) > 0)
+            while ((rs = read(fds[0], s, sizeof(s))) > 0)
             {
                 message.append(s, static_cast<size_t>(rs));
             }
--- a/cpp/src/IceGrid/Activator.cpp
+++ b/cpp/src/IceGrid/Activator.cpp
@@ -868,7 +868,7 @@ Activator::activate(
         string message;
 
         // We keep the Activator mutex locked for the whole activate.
-        while ((rs = read(errorFds[0], &s, 16)) > 0) // NOLINT(clang-analyzer-unix.BlockInCriticalSection)
+        while ((rs = read(errorFds[0], s, sizeof(s))) > 0) // NOLINT(clang-analyzer-unix.BlockInCriticalSection)
         {
             message.append(s, static_cast<size_t>(rs));
         }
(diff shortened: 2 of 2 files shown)


pepone (approved):
Looks good. The change covers all three pipe-draining loops identified in #5917, removes the duplicated size literal, and uses the conventional buffer pointer without changing behavior. Approved.

PR #6018, opened 2026-07-15
Fix nullability of peerCertificate in Swift SSL bridge

### LINKED ISSUE #5936: [Audit-Medium] Swift: SSL peerCertificate crosses the bridge as an invalid non-optional when the peer presents no certificate

bernardnormier:
When the peer of an SSL connection presents no certificate, the C++ side of the Swift bridge forwards a NULL `SecCertificateRef` through a parameter declared `_Nonnull` (it sits in the header's `NS_ASSUME_NONNULL` region), and the Swift factory imports it as a non-optional `SecCertificate`. This PR annotates the bridge parameter `nullable` and makes the factory parameter `SecCertificate?`, so the chain honors its declared nullability down to the already-optional `SSLConnectionInfo.peerCertificate` property.

Contrary to the issue's description, the defect is not observable: Swift represents `Optional` of a class/CF type with the NULL pointer as `nil`, so the NULL reference stored into the optional property was already indistinguishable from `nil`. The change removes the nullability-contract violation (undefined behavior that currently happens to be benign) rather than fixing an application-visible bug — hence no CHANGELOG entry.

Also extends the `Ice/info` test: the test server now accepts SSL clients without a certificate (`IceSSL.VerifyPeer=1`) and reports the presence of the peer certificate in `getConnectionInfoAsContext`, and the client verifies both directions — a non-nil `peerCertificate` when the peer presents a certificate, and a nil one on the server side for a client that doesn't.

Fixes #5936.

### DIFF
--- a/swift/src/Ice/ConnectionInfoFactory.swift
+++ b/swift/src/Ice/ConnectionInfoFactory.swift
@@ -58,7 +58,7 @@ final class ConnectionInfoFactory: ICEConnectionInfoFactory {
         WSConnectionInfo(underlying: underlying as! ConnectionInfo, headers: headers)
     }
 
-    static func createSSLConnectionInfo(_ underlying: Any, peerCertificate: SecCertificate) -> Any {
+    static func createSSLConnectionInfo(_ underlying: Any, peerCertificate: SecCertificate?) -> Any {
         return SSLConnectionInfo(
             underlying: underlying as! ConnectionInfo, peerCertificate: peerCertificate)
     }
--- a/swift/src/IceImpl/include/Connection.h
+++ b/swift/src/IceImpl/include/Connection.h
@@ -53,7 +53,7 @@ ICEIMPL_API @protocol ICEConnectionInfoFactory
 
 + (id)createWSConnectionInfo:(id)underlying headers:(NSDictionary<NSString*, NSString*>*)headers;
 
-+ (id)createSSLConnectionInfo:(id)underlying peerCertificate:(SecCertificateRef)peerCertificate;
++ (id)createSSLConnectionInfo:(id)underlying peerCertificate:(nullable SecCertificateRef)peerCertificate;
 
 + (id)createIAPConnectionInfo:(BOOL)incoming
                   adapterName:(NSString*)adapterName
--- a/swift/test/Ice/info/AllTests.swift
+++ b/swift/test/Ice/info/AllTests.swift
@@ -212,6 +212,23 @@ func allTests(_ helper: TestHelper) async throws {
             try test(ctx["ws.Sec-WebSocket-Key"] != nil)
         }
 
+        if ["ssl", "wss"].contains(helper.getTestProtocol()) {
+            // Both peers presented a certificate.
+            let sslInfo = getSSLConnectionInfo(info)!
+            try test(sslInfo.peerCertificate != nil)
+            try test(ctx["peerCertificate"] == "true")
+
+            // The server reports a nil peer certificate for a client that doesn't present one.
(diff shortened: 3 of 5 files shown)


pepone (approved):
LGTM. Nice catch on the nullability-contract violation, and the two-direction test (present cert vs. `VerifyPeer=1` + no client cert) exercises the exact NULL-forwarding path. Confirmed the no-cert client genuinely presents nothing on SecureTransport — the client chain is built only from `CertFile`/`FindCert`, never auto-selected from the keychain.

One optional nit in `Server.swift`: `initialize(_ properties:)` already does the `InitializationData` wiring, so this can lose two lines:

```swift
let properties = try createTestProperties(args)
// Accept SSL clients that don't present a certificate.
properties.setProperty(key: "IceSSL.VerifyPeer", value: "1")
let communicator = try initialize(properties)
```

PR #6023, opened 2026-07-16
Restore IceStorm batch subscriber delivery

### LINKED ISSUE #5964: Bring back IceStorm batch subscriber delivery, accidentally removed by the C++17 port
IceStorm no longer delivers events to batch oneway / batch datagram subscribers in batches: the subscriber's proxy is silently converted to a plain oneway/datagram proxy at subscription time ([`Subscriber.cpp:469-478`](https://github.com/zeroc-ice/ice/blob/6dd2adada8ea0aa3f01e35bd4bce237e910bdaa9/cpp/src/IceStorm/Subscriber.cpp#L469-L478)), and the timer-driven flush (`SubscriberBatch`) is gone.

**The removal was accidental.** It originated in the 2020 Ice 4.0-era C++17 port (c6ab3b8193e304fc0c9b11f8a9d68acfd70ebd74, #716), when Ice 4.0 planned to drop batch support. We then changed course — 4.0 became IceRPC and 3.8 kept batch invocations — but #1629 brought that port branch to main wholesale, carrying the removal along. Everything around the feature survived, and nothing announced it:

- No changelog entry.
- The `IceStorm.Flush.Timeout` property and the `Instance::_flushInterval` read of it.
- The `flush-timeout` parameter of the default IceGrid templates.
- The documentation still describes both batch delivery modes.
- `test/IceStorm/federation2` still runs "testing federation with batch subscribers" (`-b`) — passing only because the downgrade is silent, so it now tests oneway delivery.

**Plan:** bring back batch delivery — keep the subscriber's batch proxy and flush it on the instance timer every `IceStorm.Flush.Timeout` milliseconds, reimplemented on the current `Subscriber` architecture (AMI outstanding-count bookkeeping, retry/reap semantics per #5939).

Supersedes #5959, which proposed removing the dead property instead.

pepone:
Fixes #5964.

Batch oneway and batch datagram subscribers stopped being delivered in batches after the 2020 C++17 port: the subscriber's proxy was silently downgraded to a plain oneway/datagram proxy at subscription time, and the timer-driven flush was removed. Everything around the feature survived — the `IceStorm.Flush.Timeout` property, the IceGrid templates, and the `federation2` batch test — so the downgrade went unnoticed.

This reinstates `SubscriberBatch` on the current subscriber architecture: it keeps the batch proxy, accumulates events, and flushes them together every `IceStorm.Flush.Timeout` milliseconds on a dedicated instance flusher timer. The flusher is destroyed in `Instance::destroy()` (after the subscribers are drained) rather than `Instance::shutdown()`, so a pending flush can still complete during teardown. A misconfigured negative `IceStorm.Flush.Timeout` is clamped to the default with a warning.

### Includes the `Timer::destroy()` deadlock fix (also up as #6022)

Restoring batch delivery surfaced a latent core deadlock in `Timer::destroy()` — that's the first commit on this branch, and it's up for review on its own in #6022. It's included here so the batch subscriber tests don't deadlock on shutdown in CI; once #6022 merges to `main`, this branch rebases and that commit drops out.

Verified: the full IceStorm test suite (single, federation2 batch phases across all subcases, rep/stress) and the core `Ice/timeout`, `Ice/idleTimeout`, `Ice/ami`, and `Ice/background` tests all pass.

### DIFF
--- a/cpp/src/IceStorm/Subscriber.cpp
+++ b/cpp/src/IceStorm/Subscriber.cpp
@@ -88,6 +88,29 @@ namespace
 // Each of the various Subscriber types.
 namespace
 {
+    class SubscriberBatch final : public Subscriber
+    {
+    public:
+        SubscriberBatch(
+            const shared_ptr<Instance>&,
+            const SubscriberRecord&,
+            const Ice::ObjectPrx&,
+            int,
+            Ice::ObjectPrx);
+
+        void flush() override;
+
+        // Called on the flush timer to deliver the queued events as a single batch.
+        void doFlush();
+
+    private:
+        // Called once the batch has been written to the transport.
+        void batchSent();
+
+        const Ice::ObjectPrx _obj;
+        const std::chrono::milliseconds _interval;
+    };
+
     class SubscriberOneway final : public Subscriber
     {
     public:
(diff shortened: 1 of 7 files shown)


pepone:
Review by Codex (OpenAI)

The overall approach looks good, and the focused `IceStorm/single`, `IceStorm/federation2`, and `Ice/timeout` tests pass. I found one correctness issue and one important test gap:

1. `SubscriberBatch::doFlush()` catches only `std::exception`. A configured batch-request interceptor can throw a non-standard exception, which `BatchRequestQueue` deliberately rethrows. In that case the exception escapes with `_outstanding == 1`; the timer logs and swallows it, subsequent batches are never scheduled, and `shutdown()` can wait forever. Please catch all exceptions and normalize unknown exceptions before passing them to `error()`, or update `error()` to classify them.

2. Please add a regression test that proves batching is actually occurring. The existing tests only check eventual delivery and passed while batch proxies were silently downgraded. A test could configure a sufficiently long `IceStorm.Flush.Timeout`, assert that no callbacks arrive before the timeout, and then verify delivery after the flush.

Apart from these two points, the timer lifetime and subscriber teardown changes look sound.

— Codex

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -134,6 +157,116 @@ namespace
...
+
+        auto self = static_pointer_cast<SubscriberBatch>(shared_from_this());
+        _obj->ice_flushBatchRequestsAsync(
+            [self](exception_ptr ex) { self->error(true, ex); },
+            [self](bool) { self->batchSent(); });
+    }
+    catch (const std::exception&)
+    {
+        error(true, current_exception());
+    }
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Fixed. `doFlush` now catches `(...)`, and `error()` has a `catch (...)` fallback that classifies an unknown exception as a non-hard failure. So a non-`std::exception` rethrown from a batch-request interceptor decrements the outstanding count and unwinds normally instead of hanging `shutdown()`.

cpp/src/IceStorm/Subscriber.cpp
```diff
@@ -134,6 +157,116 @@ namespace
...
+        return;
+    }
+
+    // Defer delivery: accumulate events and send them together when the flush timer fires. _outstanding doubles as a
+    // "flush already scheduled or in flight" guard, so at most one flush is pending at a time (_maxOutstanding is 1).
+    if (_outstanding == 0)
+    {
+        ++_outstanding;
+        auto self = static_pointer_cast<SubscriberBatch>(shared_from_this());
+        _instance->batchFlusher()->schedule([self] { self->doFlush(); }, _interval);
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Added a regression test to `IceStorm/single`: a plain oneway subscriber acts as a checkpoint — once it has received every event, the batch subscriber must still have 0, then receives them together after the flush. It fails against the silent-downgrade behavior and passes with the fix.

cpp/test/IceStorm/single/Subscriber.cpp
```diff
@@ -246,6 +275,38 @@ Subscriber::run(int argc, char** argv)
...
+
+        const int count = 20;
+        auto publisher = uncheckedCast<SinglePrx>(batchTopic->getPublisher()->ice_twoway());
+        for (int i = 0; i < count; ++i)
+        {
+            publisher->event(i);
+        }
+
+        onewayCounter->waitForCount(count);
+        test(batchCounter->count() == 0);
```
bernardnormier: `test(batchCounter->count() == 0)` is a hard assertion that publishing all 20 events and delivering them to the oneway subscriber completed within the 1s `IceStorm.Flush.Timeout` window that started when the first event reached the batch subscriber. On a loaded CI runner (sanitizer builds, parallel suites) a >1s stall here fails the test, and the failure would read as "batching is broken" rather than "machine was slow". Suggestion: keep the strict check but only enforce it when the run demonstrably stayed inside the window — no added wall-clock time, and `waitForCount` below still verifies delivery on every run:

```suggestion
        const int count = 20;
        auto publisher = uncheckedCast<SinglePrx>(batchTopic->getPublisher()->ice_twoway());
        const auto start = chrono::steady_clock::now();
        for (int i = 0; i < count; ++i)
        {
            publisher->event(i);
        }

        onewayCounter->waitForCount(count);
        // The zero check is only meaningful while this run is still within the 1s IceStorm.Flush.Timeout window: on
        // a slow machine the flush may already have fired. Skip the check rather than fail it.
        if (chrono::steady_clock::now() - start < 500ms)
        {
            test(batchCounter->count() == 0);
        }

File: cpp/test/IceStorm/single/Subscriber.cpp [hunk 4]
```diff
@@ -246,6 +275,38 @@ Subscriber::run(int argc, char** argv)
...
         p->waitForEvents();
     }
+
+    //
+    // Verify that a batch subscriber actually batches: its events must be coalesced and delivered together when the
+    // IceStorm.Flush.Timeout timer fires, not one at a time like a plain oneway subscriber. A plain oneway subscriber
+    // on the same topic serves as a checkpoint: once it has received every event, a batch subscriber must still have
+    // received nothing, because its events are queued until the next flush.
+    //
+    {
```
bernardnormier, line 286: The other checks in this test print their label *before* running (`testing <name> ... ` then `ok`), so a hang points at the check that died. Printing the whole line only at the end loses that: a `waitForCount` timeout here shows no marker. Suggestion — print the label up front:

```suggestion
    cout << "testing batch delivery ... " << flush;
    {
```

cpp/test/IceStorm/single/Subscriber.cpp
```diff
@@ -246,6 +275,38 @@ Subscriber::run(int argc, char** argv)
...
+        }
+
+        onewayCounter->waitForCount(count);
+        test(batchCounter->count() == 0);
+        batchCounter->waitForCount(count);
+
+        batchTopic->destroy();
+    }
+
+    cout << "testing batch delivery ... ok" << endl;
```
bernardnormier: …and with the label printed before the block (see comment above), this becomes:

```suggestion
    cout << "ok" << endl;
```

cpp/test/IceStorm/single/Subscriber.cpp
pepone: 🤖 **Claude:** Applied — added the `elapsed < 500ms` guard, with `waitForCount` still verifying delivery on every run. Confirmed it doesn't weaken the regression coverage: with the silent downgrade reintroduced, the guarded check still fires on a normal run (elapsed well under 500ms) and fails, so this keeps catching the downgrade while no longer flaking on a slow runner.

cpp/test/IceStorm/single/Subscriber.cpp
pepone: 🤖 **Claude:** Applied — the label now prints before the block, so a `waitForCount` timeout points here.

cpp/test/IceStorm/single/Subscriber.cpp
pepone: 🤖 **Claude:** Applied.

bernardnormier (approved):
Looks good!

PR #6024, opened 2026-07-16
Strip the friendlyName from the test PKCS12 files

### LINKED ISSUE #5973: Intermittent Schannel 0x8009030D failures on Windows CI caused by shared PFX key-container race
## Symptom

C# test suites running an `ssl` config on Windows occasionally fail on the first TLS call with:

```
Ice.SecurityException: SSL authentication failure.
 ---> System.ComponentModel.Win32Exception (0x8009030D): The credentials supplied to the package were not recognized
   at System.Net.SSPIWrapper.AcquireCredentialsHandle(ISSPIInterface secModule, String package, CredentialUse intent, SCH_CREDENTIALS* scc)
```

Latest occurrence: `csharp/Ice/servantLocator` in https://github.com/zeroc-ice/ice/actions/runs/29270468032 (.NET10 on windows-2025). The same suite passed the identical ssl config seconds earlier, and the same certificates worked in dozens of concurrent suites in the same job — each recurrence forces a log dive to re-identify the flake.

## Root cause

Windows parallel PFX import race, acknowledged by Microsoft as won't-fix (see dotnet/runtime#23221, dotnet/runtime#17654, dotnet/runtime#20356):

- The C# SSL engine imports `IceSSL.CertFile` with `X509KeyStorageFlags.UserKeySet` (`csharp/src/Ice/SSL/SSLEngine.cs`), which writes the private key to an on-disk key container; .NET deletes that container when the `X509Certificate2` is disposed or finalized.
- `certs/makecerts.sh` exports the test PKCS12 files with `-name "${alias}"`, which sets a `friendlyName` on the key bag. Windows uses that name as the key container name, so every concurrent test process importing e.g. `certs/common/ca/client.p12` shares a **single** container.
- Parallel test suites spawn short-lived client/server processes that all import the same two PFX files. A process exiting deletes the shared key file out from under a process that is mid-handshake, and the next `AcquireCredentialsHandle` fails with 0x8009030D.

C++ tests on Windows import the same PFX files through `PFXImportCertStore`, so they are in principle exposed to the same race.

## Fix

Strip the key `friendlyName` from the generated test PFX files so each import gets a unique GUID-named container — the same fix .NET applied to its own test data (dotnet/corefx-testdata@7d0aeccf). The JKS conversion in `makecerts.sh` consumes the p12 alias, so it needs to either convert from a named intermediate p12 or pass the alias explicitly.

Note: importing with `X509KeyStorageFlags.EphemeralKeySet` is not an option — Schannel rejects ephemeral private keys with the same 0x8009030D error (dotnet/runtime#110067, dotnet/runtime#84505).

pepone:
Fixes #5973.

## Root cause

Windows uses the friendlyName of a PKCS12 key bag as the name of the CryptoAPI key container it creates on import. `makecerts.sh` exported the test PKCS12 files with `-name "${alias}"`, so every process importing e.g. `certs/common/ca/client.p12` shared a single container named `client`. The test suites run many short-lived processes in parallel, and a process exiting deletes that shared container out from under a process that is still handshaking — the next `AcquireCredentialsHandle` then fails with `0x8009030D`.

## Fix

Export the PKCS12 files without a friendlyName, so Windows names each container after a freshly generated GUID instead. This is the same fix .NET applied to its own test data (dotnet/corefx-testdata@7d0aeccf).

`keytool` takes the JKS alias from the PKCS12 friendlyName, so the JKS files are converted from a temporary named copy, which keeps their aliases unchanged (`java/test/.../IceSSL/configuration/AllTests.java` looks up `client`/`server` by alias).

C++ imports the same files through `PFXImportCertStore` and benefits automatically.

## Scope

The 38 regenerated PKCS12 files were re-exported from the already-committed `_key.pem`/`_cert.pem` files, so **no certificate was re-issued**: no new keys, no shifted expiry dates, only the friendlyName attribute is gone. The `.jks` files are untouched. `ca_all.p12` is built by `keytool -importcert` and holds no private keys, so it is left as is.

## Verification

- All 76 certificates and every private key across the 38 files are byte-identical to their committed versions; zero friendlyNames remain.
- Running the new `export_pkcs12_and_jks` helper with `keytool` produces JKS alias `client`, matching the committed keystore.
- Reproduced the race through the exact import call in `csharp/src/Ice/SSL/SSLEngine.cs` (`new X509Certificate2(certFile, password, UserKeySet)`), with 8 concurrent processes × 40 imports per round:

| PKCS12 | failures |
| --- | --- |
| with friendlyName | 4, 0, 1, 0 / 320 per round |
| without friendlyName (this PR) | 0, 0, 0, 0 / 320 per round |

Failures surface as `CryptographicException: Keyset does not exist` — the same shared-container deletion that Schannel reports as `0x8009030D`. The ~0.4% intermittent rate matches how rarely this trips in CI.

> [!NOTE]
> Windows only reuses the recorded name *if it is not already in use*, so a developer machine that has run the test suite before already has `client`/`server`/`ca*` containers left over, and those leftovers mask the race locally. Clear them with `certutil -user -delkey <name>` before trying to reproduce.


bernardnormier (approved):
Looks good.

PR #6048, opened 2026-07-17
Throw MarshalException on a null getLinkProxy result in IceStorm link()

### LINKED ISSUE #5957: [Audit-Low] IceStorm link() dereferences a null getLinkProxy() result — assert-only check
The peer's `getLinkProxy()` result is checked only with `assert` ([`TopicI.cpp:610-612`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TopicI.cpp#L610-L612); the transient implementation has no check at all, [`TransientTopicI.cpp:221-222`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/TransientTopicI.cpp#L221-L222)); a null proxy flows into `SubscriberLink`'s constructor dereference ([`Subscriber.cpp:315-318`](https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceStorm/Subscriber.cpp#L315-L318)) — undefined behavior in release builds. Requires a non-conforming peer.

**Fix:** `checkNotNull` the result in both implementations. The open question is which exception to throw for a null return from a peer (hence not a mechanical patch).

---
Split out of #5858 (item B7) for individual milestone triage.

*Filed from an automated AI correctness audit of `cpp/src/IceStorm` (Claude Fable 5, 2026-07-07); confirmed through a multi-agent adversarial refute pass over the actual source.*

pepone:
Fixes an assert-only / unchecked null-proxy dereference in IceStorm topic linking.

`getLinkProxy` is expected to return the topic's non-null `TopicLink` proxy. Both `link()` implementations let a null result from a non-conforming peer flow into the `SubscriberLink` constructor, which dereferences it (`rec.obj->ice_invocationTimeout(...)`) — undefined behavior in release builds:

- **Persistent** `TopicImpl::link` — guarded only by `assert(link)`, which is elided under `NDEBUG`.
- **Transient** `TransientTopicImpl::link` — no check at all.

Both paths now check the result and throw `Ice::MarshalException`, matching Ice's own `throwNullProxyMarshalException` handling of an unexpected null proxy. Only reachable against a non-conforming/malicious peer; the normal non-null path is unchanged.

Reviewed with an independent Codex pass (data-flow trace to the `SubscriberLink` deref, exception hierarchy, and a search confirming these are the only two `getLinkProxy` consumers).

Closes #5957

### DIFF
--- a/cpp/src/IceStorm/TopicI.cpp
+++ b/cpp/src/IceStorm/TopicI.cpp
@@ -618,7 +618,12 @@ TopicImpl::link(const TopicPrx& topic, int cost)
 {
     auto internal = Ice::uncheckedCast<TopicInternalPrx>(topic);
     optional<TopicLinkPrx> link = internal->getLinkProxy();
-    assert(link);
+    if (!link)
+    {
+        // A conforming peer always returns a non-null TopicLink; a null result would be dereferenced when
+        // constructing the SubscriberLink.
+        throw Ice::MarshalException{__FILE__, __LINE__, "getLinkProxy returned a null proxy"};
+    }
 
     auto traceLevels = _instance->traceLevels();
     if (traceLevels->topic > 0)
--- a/cpp/src/IceStorm/TopicI.cpp
+++ b/cpp/src/IceStorm/TopicI.cpp
@@ -620,8 +620,8 @@ TopicImpl::link(const TopicPrx& topic, int cost)
     optional<TopicLinkPrx> link = internal->getLinkProxy();
     if (!link)
     {
-        // A conforming peer always returns a non-null TopicLink; a null result would be dereferenced when
-        // constructing the SubscriberLink.
+        // Defense in depth: getLinkProxy never returns a null proxy. Once proxies can be declared
+        // non-null in Slice (#5209), this check will be performed automatically during unmarshaling.
         throw Ice::MarshalException{__FILE__, __LINE__, "getLinkProxy returned a null proxy"};
     }
 
--- a/cpp/src/IceStorm/TransientTopicI.cpp
+++ b/cpp/src/IceStorm/TransientTopicI.cpp
@@ -231,8 +231,8 @@ TransientTopicImpl::link(optional<TopicPrx> topic, int cost, const Ice::Current&
     auto link = internal->getLinkProxy();
     if (!link)
     {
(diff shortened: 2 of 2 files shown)


bernardnormier (commented):
LGTM — verified both `getLinkProxy` consumers are covered and the throw precedes any state change in both paths.

cpp/src/IceStorm/TopicI.cpp
```diff
@@ -618,7 +618,12 @@ TopicImpl::link(const TopicPrx& topic, int cost)
 {
     auto internal = Ice::uncheckedCast<TopicInternalPrx>(topic);
     optional<TopicLinkPrx> link = internal->getLinkProxy();
-    assert(link);
+    if (!link)
+    {
+        // A conforming peer always returns a non-null TopicLink; a null result would be dereferenced when
+        // constructing the SubscriberLink.
```
bernardnormier: The peer here is always another IceStorm instance, so the "conforming peer" framing overstates it — this is really defense in depth against a bug in our own implementation.
```suggestion
        // Defense in depth: getLinkProxy never returns a null proxy. Once proxies can be declared
        // non-null in Slice (#5209), this check will be performed automatically during unmarshaling.
```

cpp/src/IceStorm/TransientTopicI.cpp
```diff
@@ -229,6 +229,12 @@ TransientTopicImpl::link(optional<TopicPrx> topic, int cost, const Ice::Current&
     checkNotNull(topic, __FILE__, __LINE__, current);
     auto internal = Ice::uncheckedCast<TopicInternalPrx>(*topic);
     auto link = internal->getLinkProxy();
+    if (!link)
+    {
+        // A conforming peer always returns a non-null TopicLink; a null result would be dereferenced when
+        // constructing the SubscriberLink.
```
bernardnormier: Same here.
```suggestion
        // Defense in depth: getLinkProxy never returns a null proxy. Once proxies can be declared
        // non-null in Slice (#5209), this check will be performed automatically during unmarshaling.
```

bernardnormier (approved):

PR #6049, opened 2026-07-17
Reject non-canonical IPv4 address forms in the Glacier2 address filters

bernardnormier:
The Glacier2 address filters match the host of an endpoint as a string, while an IPv4 address can be written in several alternative numeric forms (octal or hexadecimal parts, leading zeros, fewer than four parts, a single 32-bit number): `0x7f000001`, `2130706433`, and `127.1` are all alternative forms of `127.0.0.1`. A host in such a form matched neither the Accept nor the Reject rules written for the dotted quad, while the connection could still reach the address the rules describe.

An earlier revision of this PR converted these forms to the canonical dotted quad before matching. That requires reproducing the platform resolver's reading of each form, which is not portable: the same literal can name different addresses on different platforms (`0177.0.0.1` reads as octal `127.0.0.1` with glibc and Winsock, and as decimal `177.0.0.1` on macOS), and the platforms' handling of out-of-range values resists characterization.

With this PR, the router instead extends the existing oversized-host check: when address filters are configured, a proxy is rejected outright — regardless of the rules — when any of its endpoints has a host that writes an IPv4 address in a form other than the plain dotted quad. The check is purely structural (is every dot-separated part a numeral, and is the whole a four-part decimal quad), so it needs no knowledge of how a given platform reads a form, and it fails closed. A canonical dotted quad and a DNS name match the rules exactly as before.

The trailing dot of a fully-qualified DNS name is dropped from the host and from the rules before matching, so `backend.example.com.` and `backend.example.com` match the same rules in either position. A trailing dot on a numeric form (`127.0.0.1.`) is rejected outright: platform resolvers treat it as a DNS name, not an address.

Behavior change: when filters are configured, an endpoint that used an alternative IPv4 form deliberately (for example `Accept=127.1` matching `-h 127.1`) is now rejected; write the rule and the endpoint as the dotted quad. IPv6 hosts are not affected; IPv6 filtering is a separate, known limitation.

### DIFF
--- a/changelog.d/service-glacier2/address-filter-host-forms.md
+++ b/changelog.d/service-glacier2/address-filter-host-forms.md
@@ -1,8 +1,8 @@
 - When address filters are configured (`Glacier2.Filter.Address.Accept` or `Glacier2.Filter.Address.Reject`), the
-  Glacier2 router now rejects a proxy when any of its endpoints has a host that writes an IPv4 address in an
-  alternative numeric form: for example, `0x7f000001`, `2130706433`, and `127.1` are all alternative forms of
-  `127.0.0.1`. The filter rules match the host as a string, so a host in such a form matched neither the Accept
-  nor the Reject rules written for the usual dotted-quad form.
+  Glacier2 router now rejects a proxy when any of its endpoints has a host that writes an IPv4 address in a
+  non-canonical form: for example, `0x7f000001`, `2130706433`, and `127.1` are all non-canonical forms of
+  `127.0.0.1`. The filter rules match the host as a string, so previously a host in such a form matched neither
+  the Accept nor the Reject rules written for the usual dotted-quad form.
 
 - The Glacier2 address filters now ignore the trailing dot of a fully-qualified DNS name, in the endpoint host and
   in the Accept and Reject rules alike: a rule written for `backend.example.com` matches the host
--- a/changelog.d/service-glacier2/address-filter-host-forms.md
+++ b/changelog.d/service-glacier2/address-filter-host-forms.md
@@ -1,0 +1,10 @@
+- When address filters are configured (`Glacier2.Filter.Address.Accept` or `Glacier2.Filter.Address.Reject`), the
+  Glacier2 router now rejects a proxy when any of its endpoints has a host that writes an IPv4 address in an
+  alternative numeric form: for example, `0x7f000001`, `2130706433`, and `127.1` are all alternative forms of
+  `127.0.0.1`. The filter rules match the host as a string, so a host in such a form matched neither the Accept
+  nor the Reject rules written for the usual dotted-quad form.
+
+- The Glacier2 address filters now ignore the trailing dot of a fully-qualified DNS name, in the endpoint host and
+  in the Accept and Reject rules alike: a rule written for `backend.example.com` matches the host
+  `backend.example.com.`, and vice versa. Previously, the host and the rule matched only when both spelled the
+  trailing dot the same way.
--- a/changelog.d/service-glacier2/address-filter-host-normalization.md
+++ b/changelog.d/service-glacier2/address-filter-host-normalization.md
@@ -1,0 +1,5 @@
+- Improved the Glacier2 address filters, `Glacier2.Filter.Address.Accept` and `Glacier2.Filter.Address.Reject`: the
+  filters now normalize the host of each endpoint before matching it, so that all equivalent spellings of the same
+  host match the same rules. The normalization removes the trailing dot of a fully-qualified DNS name and converts
+  IPv4 address literals to canonical dotted-quad form. Write IPv4 rules in canonical form (`127.0.0.1`): a rule
+  spelled in a non-canonical form no longer matches.
--- a/changelog.d/service-glacier2/address-filter-host-normalization.md
+++ b/changelog.d/service-glacier2/address-filter-host-normalization.md
@@ -1,5 +1,5 @@
-- Improved the Glacier2 address filters, `Glacier2.Filter.Address.Accept` and `Glacier2.Filter.Address.Reject`: the
-  filters now normalize the host of each endpoint before matching it, so that all equivalent spellings of the same
-  host match the same rules. The normalization removes the trailing dot of a fully-qualified DNS name and converts
-  IPv4 address literals to canonical dotted-quad form. Write IPv4 rules in canonical form (`127.0.0.1`): a rule
-  spelled in a non-canonical form no longer matches.
+- Improved address filtering in the Glacier2 router. The router now normalizes the host of each endpoint before
+  matching it against the `Glacier2.Filter.Address.Accept` and `Glacier2.Filter.Address.Reject` rules, so that a
+  rule matches every equivalent spelling of the host it specifies. This normalization removes the trailing dot of
+  a fully-qualified DNS name and converts IPv4 address literals to canonical dotted-quad form (for example, `0x7f.1`
+  becomes `127.0.0.1`).
--- a/changelog.d/service-glacier2/address-filter-host-normalization.md
+++ b/changelog.d/service-glacier2/address-filter-host-normalization.md
@@ -1,5 +1,0 @@
-- Improved address filtering in the Glacier2 router. The router now normalizes the host of each endpoint before
-  matching it against the `Glacier2.Filter.Address.Accept` and `Glacier2.Filter.Address.Reject` rules, so that a
-  rule matches every equivalent spelling of the host it specifies. This normalization removes the trailing dot of
-  a fully-qualified DNS name and converts IPv4 address literals to canonical dotted-quad form (for example, `0x7f.1`
-  becomes `127.0.0.1`).
--- a/cpp/src/Glacier2/ProxyVerifier.cpp
+++ b/cpp/src/Glacier2/ProxyVerifier.cpp
@@ -518,7 +518,7 @@ namespace Glacier2
 
         // The leading parts are single bytes; the last part covers the remaining bytes.
         uint32_t address = 0;
-        for (vector<unsigned long>::size_type i = 0; i + 1 < parts.size(); ++i)
+        for (size_t i = 0; i + 1 < parts.size(); ++i)
         {
             if (parts[i] > 0xFF)
             {
@@ -651,7 +651,7 @@ namespace Glacier2
             {
                 host.pop_back();
             }
-            if (optional<string> canonical = canonicalIPv4Address(host))
+            else if (optional<string> canonical = canonicalIPv4Address(host))
             {
                 host = *canonical;
             }
--- a/cpp/src/Glacier2/ProxyVerifier.cpp
+++ b/cpp/src/Glacier2/ProxyVerifier.cpp
@@ -500,7 +500,7 @@ namespace Glacier2
             char* end;
             errno = 0;
             unsigned long value = strtoul(p, &end, 0);
-            if (errno == ERANGE || value > 0xFFFFFFFF)
+            if (errno == ERANGE || value > UINT32_MAX)
             {
                 return nullopt;
             }
--- a/cpp/src/Glacier2/ProxyVerifier.cpp
+++ b/cpp/src/Glacier2/ProxyVerifier.cpp
@@ -4,11 +4,8 @@
 #include "../Ice/ConsoleUtil.h"
 #include "Ice/StringUtil.h"
 
+#include <algorithm>
 #include <cctype>
-#include <cerrno>
-#include <cstdint>
-#include <cstdlib>
-#include <optional>
 #include <stdexcept>
 #include <string>
 #include <vector>
@@ -482,74 +479,95 @@ namespace Glacier2
         return true;
     }
 
-    // Parses host as an IPv4 address literal, accepting every form the resolver accepts (the inet_aton
-    // grammar): one to four '.'-separated parts, each decimal, octal (leading 0), or hexadecimal (leading 0x),
-    // with the last part supplying the remaining bytes of the address. Returns the address in canonical
-    // dotted-quad form, or nullopt when host is not an IPv4 address literal.
-    static optional<string> canonicalIPv4Address(const string& host)
+    // Splits host into its '.'-separated parts.
+    static vector<string> splitHost(const string& host)
     {
-        vector<unsigned long> parts;
-        const char* p = host.c_str();
+        vector<string> parts;
+        string::size_type start = 0;
         while (true)
         {
-            // Each part must start with a digit; strtoul alone would also accept whitespace and signs.
-            if (!isdigit(static_cast<unsigned char>(*p)))
-            {
-                return nullopt;
-            }
-            char* end;
-            errno = 0;
-            unsigned long value = strtoul(p, &end, 0);
-            if (errno == ERANGE || value > UINT32_MAX)
-            {
-                return nullopt;
-            }
-            parts.push_back(value);
-            if (*end == '\0')
+            string::size_type dot = host.find('.', start);
+            parts.push_back(host.substr(start, dot == string::npos ? string::npos : dot - start));
+            if (dot == string::npos)
             {
                 break;
             }
-            if (*end != '.' || parts.size() == 4)
-            {
-                return nullopt;
-            }
-            p = end + 1;
+            start = dot + 1;
         }
-
-        // The leading parts are single bytes; the last part covers the remaining bytes.
-        uint32_t address = 0;
-        for (size_t i = 0; i + 1 < parts.size(); ++i)
-        {
-            if (parts[i] > 0xFF)
-            {
-                return nullopt;
-            }
-            address |= static_cast<uint32_t>(parts[i]) << (8 * (3 - i));
-        }
-        unsigned long last = parts.back();
-        int lastSize = 5 - static_cast<int>(parts.size()); // in bytes
-        if (lastSize < 4 && last >= (1ul << (8 * lastSize)))
-        {
-            return nullopt;
-        }
-        address |= static_cast<uint32_t>(last);
-
-        ostringstream os;
-        os << ((address >> 24) & 0xFF) << '.' << ((address >> 16) & 0xFF) << '.' << ((address >> 8) & 0xFF) << '.'
-           << (address & 0xFF);
-        return os.str();
+        return parts;
     }
 
-    // Returns true when an endpoint of the proxy has a host longer than 255 characters. No legal host name or
-    // IP address is that long, and the bound keeps the cost of matching the address rules low.
-    static bool hasOversizedHost(const ObjectPrx& proxy)
+    // Returns true when every '.'-separated part of host is a numeral (digits, or hexadecimal digits with a
+    // 0x prefix): the host spells an IPv4 address in one of the many forms resolvers accept. This is a check
+    // on the spelling, not the value: a digit-only part covers the decimal and octal (leading zero) readings
+    // alike.
+    static bool isIPv4Numeral(const string& host)
+    {
+        for (const string& part : splitHost(host))
+        {
+            bool digitsOnly =
+                !part.empty() && all_of(part.begin(), part.end(), [](unsigned char c) { return isdigit(c) != 0; });
+            bool hex = part.size() > 2 && part[0] == '0' && (part[1] == 'x' || part[1] == 'X') &&
+                       all_of(part.begin() + 2, part.end(), [](unsigned char c) { return isxdigit(c) != 0; });
+            if (!digitsOnly && !hex)
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    // Returns true when host is the canonical spelling of an IPv4 address: four decimal parts, each between 0
+    // and 255, with no leading zero.
+    static bool isCanonicalIPv4Spelling(const string& host)
+    {
+        vector<string> parts = splitHost(host);
+        if (parts.size() != 4)
+        {
+            return false;
+        }
+        for (const string& part : parts)
+        {
+            bool digitsOnly =
+                !part.empty() && all_of(part.begin(), part.end(), [](unsigned char c) { return isdigit(c) != 0; });
+            if (!digitsOnly || part.size() > 3 || (part.size() > 1 && part[0] == '0') || stoi(part) > 255)
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    // Returns true when an endpoint of the proxy has a host that the address rules cannot match reliably; such
+    // a proxy is rejected outright:
+    // - A host longer than 255 characters. No legal host name or IP address is that long, and the bound keeps
+    //   the cost of matching the address rules low.
+    // - A non-canonical spelling of an IPv4 address, such as 0x7f000001, 2130706433, 0177.0.0.1, or 127.1.
+    //   The resolver accepts many aliases for the same address (with platform-dependent readings), so the
+    //   spelling would bypass rules written for the dotted quad.
+    // - An IPv4 address with a trailing dot: the resolver reads it as a DNS name, not as an address.
+    // A DNS name with a trailing dot is fine: the resolver treats 'name.' as 'name', and the matching strips
+    // the dot the same way.
+    static bool hasDisallowedHost(const ObjectPrx& proxy)
     {
         for (const auto& endpoint : proxy->ice_getEndpoints())
         {
             string host;
-            if (extractPart("-h ", endpoint->toString(), host) && host.size() > 255)
+            if (extractPart("-h ", endpoint->toString(), host))
             {
-                return true;
+                if (host.size() > 255)
+                {
+                    return true;
+                }
+                bool trailingDot = host.size() > 1 && host.back() == '.';
+                if (trailingDot)
+                {
+                    host.pop_back();
+                }
+                if (isIPv4Numeral(host) && (trailingDot || !isCanonicalIPv4Spelling(host)))
+                {
+                    return true;
+                }
             }
         }
         return false;
@@ -644,16 +662,11 @@ namespace Glacier2
             // Host names are case-insensitive; the rule was folded to lower case when parsed.
             host = toLower(host);
 
-            // Normalize the host so that a rule cannot be bypassed with an equivalent but different spelling
-            // of the same host: drop the trailing dot of a fully-qualified DNS name, and put an IPv4 address
-            // literal in canonical dotted-quad form.
+            // The resolver treats the fully-qualified name 'name.' as 'name'; strip the trailing dot so the
+            // spelling matches the rules written for 'name'.
             if (host.size() > 1 && host.back() == '.')
             {
                 host.pop_back();
-            }
-            else if (optional<string> canonical = canonicalIPv4Address(host))
-            {
-                host = *canonical;
             }
 
             string port;
@@ -801,6 +814,13 @@ namespace Glacier2
                 // Host names are case-insensitive; fold the rule to lower case and match it against the
                 // lower-cased host.
                 addr = toLower(addr);
+
+                // The trailing dot of a fully-qualified name is stripped from the host before matching; strip
+                // it from the rule the same way.
+                if (addr.size() > 1 && addr.back() == '.')
+                {
+                    addr.pop_back();
+                }
 
                 //
                 // The addr portion can contain alphanumerics, * and
@@ -1056,7 +1076,7 @@ Glacier2::ProxyVerifier::verify(const ObjectPrx& proxy)
 
     bool result = false;
 
-    if (Glacier2::hasOversizedHost(proxy))
+    if (Glacier2::hasDisallowedHost(proxy))
     {
         // Rejected regardless of the configured rules.
     }
--- a/cpp/src/Glacier2/ProxyVerifier.cpp
+++ b/cpp/src/Glacier2/ProxyVerifier.cpp
@@ -4,6 +4,11 @@
 #include "../Ice/ConsoleUtil.h"
 #include "Ice/StringUtil.h"
 
+#include <cctype>
+#include <cerrno>
+#include <cstdint>
+#include <cstdlib>
+#include <optional>
 #include <stdexcept>
 #include <string>
 #include <vector>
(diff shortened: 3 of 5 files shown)


cpp/src/Glacier2/ProxyVerifier.cpp
```diff
@@ -477,6 +482,64 @@ namespace Glacier2
...
+            if (*end != '.' || parts.size() == 4)
+            {
+                return nullopt;
+            }
+            p = end + 1;
+        }
+
+        // The leading parts are single bytes; the last part covers the remaining bytes.
+        uint32_t address = 0;
+        for (vector<unsigned long>::size_type i = 0; i + 1 < parts.size(); ++i)
```
InsertCreativityHere: Super minor, but makes this a little more digestible, and I think is what we usually do instead of writing the fully-qualified type.
```suggestion
        for (size_t i = 0; i + 1 < parts.size(); ++i)
```

changelog.d/service-glacier2/address-filter-host-normalization.md
```diff
@@ -0,0 +1,5 @@
+- Improved the Glacier2 address filters, `Glacier2.Filter.Address.Accept` and `Glacier2.Filter.Address.Reject`: the
+  filters now normalize the host of each endpoint before matching it, so that all equivalent spellings of the same
```
InsertCreativityHere: The colon here feels weird, you'd expect it to list the "filters" being talked about, but it's like, delayed?
I guess the colon is tied to "the improvement" not "the filters" but it doesn't read that way to me.
```suggestion
- Improved the Glacier2 address filters: `Glacier2.Filter.Address.Accept` and `Glacier2.Filter.Address.Reject`. These
  filters now normalize the host of each endpoint before matching it, so that all equivalent spellings of the same
```

changelog.d/service-glacier2/address-filter-host-normalization.md
```diff
@@ -0,0 +1,5 @@
...
+  host match the same rules. The normalization removes the trailing dot of a fully-qualified DNS name and converts
+  IPv4 address literals to canonical dotted-quad form. Write IPv4 rules in canonical form (`127.0.0.1`): a rule
+  spelled in a non-canonical form no longer matches.
```
InsertCreativityHere: Also, I think it tried too hard to make this last sentence short.
With the current wording, it's not clear that "YOU need to write rules in canonical form" nor is it clear "this has changed", instead it feels like a tack on to the sentence above; how Ice normalizes it.
```suggestion
  IPv4 address literals to canonical dotted-quad form. Now, IPv4 rules must be written in canonical form (`127.0.0.1`);
  rules spelled in a non-canonical form no longer match.
```

cpp/src/Glacier2/ProxyVerifier.cpp
```diff
@@ -580,6 +643,19 @@ namespace Glacier2
...
             host = toLower(host);
+
+            // Normalize the host so that a rule cannot be bypassed with an equivalent but different spelling
+            // of the same host: drop the trailing dot of a fully-qualified DNS name, and put an IPv4 address
+            // literal in canonical dotted-quad form.
+            if (host.size() > 1 && host.back() == '.')
+            {
+                host.pop_back();
+            }
+            if (optional<string> canonical = canonicalIPv4Address(host))
```
InsertCreativityHere: I'm pretty sure the answer is just "we de-facto don't support it", but DNS names can be fully numeric. And our `canonicalIPv4Address` will happily take these numerical DNS host-names and convert them to dotted-quad for rule checking purposes, since it's very permissive in matching. Is there any secondary security issue here because of this?

Doing `else-if` would at least _slightly hard against this... but not entirely.
```suggestion
            else if (optional<string> canonical = canonicalIPv4Address(host))
```

InsertCreativityHere (dismissed):
Looks good to me!

cpp/src/Glacier2/ProxyVerifier.cpp
bernardnormier: Done.

cpp/src/Glacier2/ProxyVerifier.cpp
bernardnormier: Good catch — I applied the `else if`, and it's more than a slight hardening: it actually makes the matching faithful to the resolver.

For a host without a trailing dot, there is no secondary issue: anything the inet_aton grammar accepts is consumed by the resolver as an address literal and never reaches DNS, so a fully-numeric DNS name of that shape is unreachable through `getaddrinfo` anyway — canonicalizing it matches exactly where a connection would go.

A trailing dot however disqualifies inet_aton parsing: `getaddrinfo` resolves `127.1` and `0x7f000001` numerically to 127.0.0.1, but sends `127.1.`, `0x7f.1.` and even `127.0.0.1.` to DNS as fully-qualified names (verified on macOS). So a dot-stripped host is by definition a DNS name, and the `else if` keeps it from being reinterpreted as an address literal. Without it, `-h 127.1.` would match an accept rule for `127.0.0.1` even though the connection actually performs a DNS lookup of `127.1.` — which resolves nowhere in public DNS, but wherever DNS says in a private setup.

changelog.d/service-glacier2/address-filter-host-normalization.md
bernardnormier: I agree the colon read oddly. I ended up reworking the whole entry: the filters are just configured values and don't do anything themselves, so the router is now the subject — "The router now normalizes the host of each endpoint before matching it against the ... rules".

changelog.d/service-glacier2/address-filter-host-normalization.md
bernardnormier: I dropped this sentence altogether. Before this change, matching was a literal string comparison, so a rule spelled in non-canonical form never matched the canonical host — it's very unlikely anyone wrote such rules, and telling users to change them suggests they previously worked. The entry now just describes the normalization, with an example (`0x7f.1` becomes `127.0.0.1`) that makes the expected rule spelling clear.

pepone (commented):
The parser is clean in isolation — I diff-tested `canonicalIPv4Address` against the system `inet_aton` over ~670 **in-range** inputs and the arithmetic is correct for the classic `inet_aton` grammar, with no overflow or UB (the only potentially-32-bit shift is short-circuited by `lastSize < 4`, and both LP64 and LLP64 reject `> UINT32_MAX`). The `else if` in the latest commit is right, and for a stronger reason than "slight hardening": trailing-dot numerics (`127.0.0.1.`, `2130706433.`) are not address literals to the numeric parser on macOS or Windows, so without it `2130706433.` would be reinterpreted as `127.0.0.1` when the resolver would send it to DNS. (It doesn't make the dot handling fully resolver-faithful — after stripping, `-h 127.0.0.1.` still matches `Accept=127.0.0.1` while the connection actually performs a DNS lookup — but that's inherent to matching a name-shaped host by string.)

The problem is what happens when the normalized string and the resolver disagree. Three issues below; **1 and 3 I'd fix before merge**, because each opens a hole that did not exist before this PR.

### Which resolver call the normalization has to agree with

Worth pinning down, because it's easy to validate against the wrong one. `IPEndpointI::resolve` calls `getAddresses(..., canBlock=false)` **first** (`cpp/src/Ice/IPEndpointI.cpp:453`), and `canBlock=false` sets `AI_NUMERICHOST` (`cpp/src/Ice/Network.cpp:571-573`); the blocking DNS resolve runs only if that returns empty. So in the default configuration, for a numeric literal it's the **`AI_NUMERICHOST` parse** — not `getaddrinfo(flags=0)` — that decides where Ice connects. (Two exceptions: with a network proxy configured Ice skips this path entirely at `IPEndpointI.cpp:448`, and the protocol family follows `Ice.IPv4`/`Ice.IPv6` at `Network.cpp:557`.)

### 1. Leading zeros: the PR opens an accept-rule escape on macOS

macOS parses a 4-part dotted host as *decimal, leading zeros allowed* (`inet_pton` is tried first), while this function — and glibc, and Windows — read a leading-zero part as **octal**. Measured on macOS 26.5 (arm64) under `AI_NUMERICHOST`:

| Endpoint `-h` host | macOS connects to | function → | Effect |
|---|---|---|---|
| `0177.0.0.1` | `177.0.0.1` | `127.0.0.1` | **`Accept=127.0.0.1` now admits a proxy that reaches `177.0.0.1`** |
| `0127.0.0.1` | `127.0.0.1` | `87.0.0.1` | `Reject=127.0.0.1` bypassed |
| `010.0.0.1` | `10.0.0.1` | `8.0.0.1` | `Reject=10.0.0.1` bypassed |
| `0192.0168.000.001` | `192.168.0.1` | *(nullopt → left as literal)* | `Reject=192.168.0.1` bypassed |

The first row is the one I'd flag hardest: it is **a regression, not a residual gap**. Before this PR, `0177.0.0.1` did not match `Accept=127.0.0.1` under exact string comparison, so the proxy was refused. Now it normalizes onto the allowlisted spelling and is admitted, while the connection goes to `177.0.0.1`. The PR widens an allowlist on macOS.

Only 4-part forms diverge — 1/2/3-part legacy forms use the octal parser on macOS too. That's why the new tests pass: in `127.000.000.001` every zero-padded component has the same value read as decimal or octal, so nothing in the suite distinguishes the two readings.

Confirmed on macOS only (I haven't tested another BSD). I ran the same probe on Windows 11 Pro (10.0.26200, MSVC x64): Windows reads these as octal and agrees with the function exactly, so Linux and Windows are fine here.

### 2. Oversized literals: resolver divergence on macOS *and* Windows

```
0x17f000001    AI_NUMERICHOST -> 127.0.0.1    canonicalIPv4Address -> nullopt
```

Both platforms hand back `127.0.0.1`; Ice takes that first result and connects to loopback. The function rejects the input (`value > UINT32_MAX`), so the host stays a literal string and `Reject=127.0.0.1` never fires. Unlike #1 this reproduces on Windows.

What the platforms do with oversized values is **not** the same, which matters for any fix:

| input | macOS | Windows |
|---|---|---|
| `0x17f000001` | `127.0.0.1` | `127.0.0.1` |
| `999999999999` | `212.165.15.255` | reject |
| `0x100000000` (2³²) | `0.0.0.0` | reject |
| `4294967296` (2³²) | `0.0.0.0` | reject |

macOS is consistent with plain mod-2³² truncation (`999999999999` mod 2³² is exactly `212.165.15.255`). Windows is not — it rejects that input and both spellings of 2³² while still accepting `0x17f000001`. I could not characterize the Windows rule from these probes, so please don't read this as "reject anything above 2³² and you match Windows": the repair that truncation suggests — canonicalize mod 2³² instead of rejecting — would map both 2³² spellings onto `0.0.0.0`, which Windows refuses to connect to, so the filter would start normalizing hosts that never resolve.

That we can't pin the rule down after several probes is itself an argument against re-deriving it.

### 3. Rules are never normalized — @Copilot's point, and it's broader than exact literals

Rule parsing only does `toLower(addr)` (`cpp/src/Glacier2/ProxyVerifier.cpp:803`). Endpoints are normalized, rules are not, so previously-working rules stop matching. Traced through the matcher: a rule `127.1` with no wildcard becomes a single `StartsWithString("127.1")`, and `matchAddress` requires it to consume the whole host (`ProxyVerifier.cpp:693`) — so it was exact matching before. Against a now-normalized `127.0.0.1` it fails at the fifth character.

- `Reject=127.1` + endpoint `-h 127.1` → previously rejected, now **accepted** (fail-open).
- `Accept=127.1` + endpoint `-h 127.1` → previously accepted, now rejected (fail-closed).

Normalizing exact literal rules helps but doesn't finish the job. Trailing-dot rules can be normalized the same way (`Reject=badhost.example.com.` → strip the dot), but numeric **wildcard** rules can't be translated at all:

- `Reject=0x7f*` + endpoint `-h 0x7f000001` → endpoint becomes `127.0.0.1`, the wildcard still reads `0x7f*` → accepted.

A cleaner fix that covers all of these without touching rule parsing: **match a reject rule against the union of the raw and canonicalized host**, rejecting if either matches. `Reject=127.1` matches the raw spelling; `Reject=127.0.0.1` matches the canonicalized one; `Reject=0x7f*` matches the raw one. Nothing that previously matched stops matching, and the new spellings are still caught.

Note the union is only correct for **reject** rules — for accept rules, admitting when *either* form matches would widen the allowlist, which is the failure in #1. Accept rules want the conservative direction: admit only on the canonicalized form, and decline to canonicalize when the reading is ambiguous.

Either way the changelog caveat dropped in the last commit is worth reinstating unless the union approach is adopted. The rationale for removing it was that mentioning it *"suggests they previously worked"* — but for the exactly-matching spelling they **did** work, which is what makes this a regression rather than a clarification.

### Suggested direction

Findings 1 and 2 both come from re-deriving a grammar that has to agree with a parser we don't control — one that differs across macOS, glibc and Winsock, and that we couldn't fully characterize on Windows. I'd stop re-deriving it and ask the platform:

```cpp
// Canonicalize via the same numeric parse Ice's first resolve call performs, so the matched
// host tracks where the connection would actually go. protocolSupport comes from the instance
// (Ice.IPv4/Ice.IPv6); AI_NUMERICHOST never performs a DNS lookup.
addrinfo hints{};
hints.ai_family = family;              // PF_INET / PF_INET6 / PF_UNSPEC, as Network.cpp:557 picks
hints.ai_flags = AI_NUMERICHOST;
// getaddrinfo -> inet_ntop -> dotted quad, with freeaddrinfo via a scope guard
```

Three things this needs that the current call site doesn't do:

1. **Canonicalize once per endpoint**, before the rule loop — not inside `matchEndpoint` (`ProxyVerifier.cpp:934`), which runs once per configured rule on attacker-supplied proxies, while the routing table holds its mutex (`RoutingTable.cpp:49`). As written that's an allocating resolver call per rule per endpoint per `addProxy`.
2. **Use the same protocol family Ice uses**, not a hardcoded `AF_INET`.
3. **Guard empty and NUL-containing hosts explicitly** (this applies to the current parser too).

What I would *not* claim for it: that it cannot diverge. In the default configuration (no network proxy, IPv4 enabled) it calls the same parser Ice's first resolve uses and therefore agrees with it; with a network proxy configured Ice skips that path entirely, so the filter is best-effort there and should be documented as such. The trade-off is that the canonical form becomes platform-dependent — `0127.0.0.1` → `127.0.0.1` on macOS, `87.0.0.1` on Linux/Windows. For a filter that's the right behavior, since it tracks the local router's actual destination, but it is a deliberate design choice rather than a free win.

**If a resolver call in the matching path isn't acceptable**, the minimal alternative is to keep the parser but **decline to canonicalize where the reading is platform-ambiguous**: a **four-part** host with a non-hex, zero-prefixed component whose octal and decimal values differ. That blocks `0177.0.0.1` (127 vs 177), `0127.0.0.1` (87 vs 127) and `010.0.0.1` (8 vs 10), which stay literal strings.

The scoping matters in both directions. Restricting it to four-part forms keeps `0177.1` and `017700000001` normalizing, since 1/2/3-part forms use the octal parser on macOS too and aren't ambiguous. Comparing values rather than just spotting a leading zero keeps `127.000.000.001` normalizing — every padded component there reads the same either way — so **the test added in this PR keeps passing**. `0x7f000001`, `2130706433` and `127.1` are untouched.

That removes the newly introduced allowlist escape in #1 without pretending to know which reading the local resolver picks. It does **not** address #2 — that divergence is in overflow handling, not leading zeros — and it can't cover Winsock quirks we haven't characterized.

Worth noting that the two rule directions want opposite handling **when normalization is uncertain**. (A canonical form taken from the resolver's actual destination serves both fine — the asymmetry is about what to do when you can't be sure of it.) Both should fail closed, which means opposite things:

- **Reject**: match against the union of the raw and canonicalized host, so an uncertain reading can only ever add a match.
- **Accept**: require the canonicalized form, and refuse to canonicalize ambiguous readings, so an uncertain reading can only ever remove a match.

Replacement is the wrong primitive for reject rules either way: rewriting the host to `127.0.0.1` is what makes `Reject=127.1` stop firing in #3.

Finding 3 needs fixing regardless of which parser you keep — it's a property of normalizing endpoints but not rules, not of how the normalization is computed.

bernardnormier:
Thanks — the review convinced us the premise was off, not just the parser. Canonicalizing means reproducing each platform resolver's reading of a form, and findings 1 and 2 show that reading isn't portable (and on Windows, not fully characterizable). The PR now does the opposite, in 7e66131: when address filters are configured, a proxy whose endpoint host writes an IPv4 address in any form other than the plain dotted quad is rejected outright, extending the existing oversized-host check. Nothing is ever rewritten onto a rule's spelling, and the rules only ever see the raw host.

How this lands on the findings:

1. Leading zeros: `0177.0.0.1` is rejected outright. The router makes no claim about which address it names, so the macOS decimal/octal divergence no longer matters.
2. Oversized literals: `0x17f000001`, `999999999999`, and both spellings of 2^32 are numeric forms that aren't the dotted quad, so they are rejected outright — no need to characterize Winsock's overflow rule.
3. Rules are never normalized, and now never need to be: alternative-form endpoints don't reach the rules. `Reject=127.1` and `Reject=0x7f*` keep matching nothing, but the hosts they were written against are rejected before matching; `Accept=127.1` now admits nothing (fail closed, the direction you prescribed for accept rules), and the changelog states that alternative forms are rejected when filters are configured.

The check is purely structural — every dot-separated part must be a numeral (digits, or 0x-prefixed hex) for the host to count as an IPv4 form, and canonical means four decimal parts up to 255 without leading zeros — so empty and NUL-containing hosts are handled inherently; there is no C-string parsing left. Your trailing-dot residual is closed as well: a trailing dot on a numeric form is rejected outright (the numeric parsers treat it as a DNS name), while a trailing-dot DNS name is supported — the dot is stripped from the host and from the rules alike before matching, with a test for a rule written as `badhost.example.com.`.

cpp/test/Glacier2/staticFiltering/test.py
```diff
@@ -412,6 +412,62 @@ def setup(self, current):
                     ],
                     [],
                 ),
+                (
```
InsertCreativityHere, line 415: One thing we aren't testing and I think we should be is `123.example.com`. DNS allows fully-numeric segments, and I could see future changes to this logic quietly breaking them.

cpp/src/Glacier2/ProxyVerifier.cpp
```diff
@@ -477,16 +479,95 @@ namespace Glacier2
         return true;
     }
 
-    // Returns true when an endpoint of the proxy has a host longer than 255 characters. No legal host name or
-    // IP address is that long, and the bound keeps the cost of matching the address rules low.
-    static bool hasOversizedHost(const ObjectPrx& proxy)
+    // Splits host into its '.'-separated parts.
+    static vector<string> splitHost(const string& host)
```
InsertCreativityHere, line 483: I will always advocate for using proper doc-comments on functions so they work nicer with extensions. Especially since it costs us only 1 extra slash.
```suggestion
    /// Splits @p host into its '.'-separated parts.
    static vector<string> splitHost(const string& host)
```

cpp/src/Glacier2/ProxyVerifier.cpp
```diff
@@ -477,16 +479,95 @@ namespace Glacier2
...
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    // Returns true when host is the canonical spelling of an IPv4 address: four decimal parts, each between 0
+    // and 255, with no leading zero.
+    static bool isCanonicalIPv4Spelling(const string& host)
```
InsertCreativityHere, line 522: Doc-comments. Also "zeros" should be plural. Sentences should read fine without comma-delimited parentheticals, and "four decimal parts with no leading zero." sounds pretty weird to me.
```suggestion
    /// Returns @c true when @p host is the canonical spelling of an IPv4 address: four decimal parts, each between 0
    /// and 255, with no leading zeros.
    static bool isCanonicalIPv4Spelling(const string& host)
```

cpp/src/Glacier2/ProxyVerifier.cpp
```diff
@@ -477,16 +479,95 @@ namespace Glacier2
...
+            start = dot + 1;
+        }
+        return parts;
+    }
+
+    // Returns true when every '.'-separated part of host is a numeral (digits, or hexadecimal digits with a
+    // 0x prefix): the host spells an IPv4 address in one of the many forms resolvers accept. This is a check
+    // on the spelling, not the value: a digit-only part covers the decimal and octal (leading zero) readings
+    // alike.
+    static bool isIPv4Numeral(const string& host)
```
InsertCreativityHere: Doc-comment, and also I think this could be worded much simpler:
```suggestion
    /// Returns @c true if @p host is an IPv4 address in one of the forms resolvers accept,
    /// i.e. it's '.'-separated, and each part is fully numeric, or a hexadecimal literal (with a 0x prefix).
    static bool isIPv4Numeral(const string& host)
```

changelog.d/service-glacier2/address-filter-host-forms.md
```diff
@@ -0,0 +1,10 @@
+- When address filters are configured (`Glacier2.Filter.Address.Accept` or `Glacier2.Filter.Address.Reject`), the
+  Glacier2 router now rejects a proxy when any of its endpoints has a host that writes an IPv4 address in an
+  alternative numeric form: for example, `0x7f000001`, `2130706433`, and `127.1` are all alternative forms of
+  `127.0.0.1`. The filter rules match the host as a string, so a host in such a form matched neither the Accept
+  nor the Reject rules written for the usual dotted-quad form.
```
InsertCreativityHere: Everywhere else we use the term non-canonical. Except here we say "alternative". I think we should stick with the more canonical term: "non-canonical". Also, we should be more clear about what is previous behavior vs current behavior. Like in the other entry, where we clearly say "Previously, ..."
```suggestion
- When address filters are configured (`Glacier2.Filter.Address.Accept` or `Glacier2.Filter.Address.Reject`), the
  Glacier2 router now rejects a proxy when any of its endpoints has a host that writes an IPv4 address in a
  non-canonical form: for example, `0x7f000001`, `2130706433`, and `127.1` are all non-canonical forms of
  `127.0.0.1`. The filter rules match the host as a string, so previously a host in such a form matched neither the
  Accept nor the Reject rules written for the usual dotted-quad form.
```

InsertCreativityHere (approved):
Looks good to me!

pepone (commented):
Switching from normalization to outright rejection is the right call, and it resolves the earlier findings cleanly: `0177.0.0.1`, `0x7f000001`, `127.1`, `127.000.000.001` are all rejected now, non-canonical endpoints never reach rule matching (so the reject-rule regression is gone), and stripping the trailing dot from rules as well as hosts handles the `backend.example.com.` case. Not reproducing the resolver's grammar is exactly the property that makes this portable.

Two issues remain — one form that slips *past* the new check, and one case where it fires when it shouldn't. Both are confirmed by building the PR-head `ProxyVerifier` and calling `verify()` directly.

## 1. Bare `0x` / `0X` slips through `hasDisallowedHost` and reaches loopback

`isIPv4Numeral` classifies a part as hex with:

```cpp
bool hex = part.size() > 2 && part[0] == '0' && (part[1] == 'x' || part[1] == 'X') && /* rest isxdigit */;
```

A bare `0x` or `0X` (exactly two chars, no digits after the prefix) fails `size() > 2`, so the part is treated as non-numeric, the whole host is read as a DNS name, and `hasDisallowedHost` returns false. But the resolver reads `0x` as `0`. Confirmed through `verify()` with `Glacier2.Filter.Address.Reject=127.0.0.1`:

```
-h 127.0.0.1    -> REJECT   (control: the rule works)
-h 0177.0.0.1   -> REJECT   (the rewrite correctly catches this)
-h 127.0x.0.1   -> ACCEPT   <-- getaddrinfo(AI_NUMERICHOST) reads this as 127.0.0.1
-h 127.0X.0.1   -> ACCEPT
```

So `-h 127.0x.0.1` is admitted past a `Reject=127.0.0.1` rule and Ice connects to loopback. The `0x` can sit in any zero octet, so private ranges go the same way: `10.0x.0.1` → `10.0.0.1`, `192.168.0x.1` → `192.168.0.1`. It's reachable with an ordinary proxy string — no crafting.

I checked this is the *only* such gap, not just one example: two independent fuzzers (one C++, one Python, ~11k and ~127k host spellings across every octet position) agree the complete set of "resolver reads it numerically but `isIPv4Numeral` returns false" is exactly bare `0x` / `0X`. The boundary is clean — `0xa`/`0x0` (hex body) are caught, `0xg`/`0a` (non-hex after `0`) are rejected by the resolver as names.

**Fix** — one character, `>` to `>=`:

```cpp
bool hex = part.size() >= 2 && part[0] == '0' && (part[1] == 'x' || part[1] == 'X') && /* rest isxdigit */;
```

`all_of` over the empty body is vacuously true, so `0x`/`0X` are flagged and the proxy is rejected. Re-running both fuzzers with this change drops the miss set to empty, so it closes the class rather than just the `127.0x.0.1` instance. Rejecting a filtered proxy whose host contains a bare `0x` is safe — there's no legitimate host of that shape. A test case (`0x.0.0.1` under `Reject=127.0.0.1`) would guard it.

Scope: the *reach* is confirmed on macOS; I haven't checked whether glibc or Winsock read bare `0x` as `0`. The defense gap itself is platform-independent (it's in `isIPv4Numeral`) and the fix is safe regardless, so it's worth applying either way.

## 2. `ProxySizeMax` alone triggers the non-canonical rejection

The new check also runs *outside* its documented scope. `verify()` returns early only when both rule vectors are empty — but the constructor pushes a `ProxyLengthRule` onto `_rejectRules` when `Glacier2.Filter.ProxySizeMax` is set, so with *only* a size limit configured and no address filter, `hasDisallowedHost` still runs. Confirmed through `verify()`:

```
Glacier2.Filter.ProxySizeMax=53      (no Accept/Reject address filter)
-h 127.1        -> REJECT    (the proxy is 24 chars, well under the size limit)
-h 127.0.0.1    -> ACCEPT
```

A short, otherwise-valid proxy using a compact IPv4 form is rejected purely because a size limit is configured. That contradicts the changelog's "when address filters are configured" scope and changes `ProxySizeMax`-only behavior. `_rejectRules.size()` can't be the gate for the numeric-form check, since the length rule lives there too — it needs a separate "an address filter was configured" flag, or the numeric-form check split out of `hasDisallowedHost` so it only runs when an address rule exists.

*(Finding 2 came out of an independent review of the same commit.)*

---

Neither blocks the rewrite being the right direction, but I'd fix both before merge: #1 lets a client reach a rejected address, and #2 rejects proxies no address rule was written for.

changelog.d/service-glacier2/address-filter-host-forms.md
bernardnormier: Applied — the entry now says "non-canonical" and spells out "previously", and I renamed the PR to match, so the term is now consistent across the code, the changelog, and the title.

cpp/test/Glacier2/staticFiltering/test.py
bernardnormier: The behavior is safe today — `example` isn't a numeral, so `isIPv4Numeral` reads the whole host as a DNS name and it goes to the rules as a string — but I don't see a way to guard it in this suite: an accepted proxy is actually pinged through the router, so a positive "this name passes the filter" case needs a host that resolves, and `123.example.com` doesn't. The reject-side variant (`Reject=123.example.com` against `-h 123.example.com`) wouldn't guard anything, since a rule rejection and an outright rejection are indistinguishable to the test client.

cpp/src/Glacier2/ProxyVerifier.cpp
bernardnormier: We deliberately use plain `//` comments in this file — it's an internal source file, and `///` doc-comments are the convention for the public headers. Keeping the current style here (applies to the other two suggestions as well).

cpp/src/Glacier2/ProxyVerifier.cpp
bernardnormier: Took the "zeros" fix, keeping the `//` style per the other thread.

cpp/src/Glacier2/ProxyVerifier.cpp
bernardnormier: Keeping this comment as is (modulo the bare-`0x` update from the other review): the suggested wording says the host *is* an IPv4 address in a form resolvers accept, but the function deliberately checks only the spelling — `999.999.999.999` passes it even though no resolver reads it as an address. That spelling-vs-value distinction is what keeps this check portable, so I'd rather the comment spell it out than read simpler.

bernardnormier:
Both fixed — and thanks for the fuzzer-backed confirmation that bare `0x`/`0X` is the complete miss set, which is what makes the one-character fix sufficient.

1. `isIPv4Numeral` now classifies a part as hex with `part.size() >= 2`, so a bare `0x`/`0X` part counts as a numeral (the `all_of` over the empty body is vacuously true) and `127.0x.0.1` is rejected outright. I added the test you suggested: `0x.0.0.1` under `Reject=127.0.0.1`.

2. The verifier now keeps an explicit "an address filter is configured" flag, set when `Glacier2.Filter.Address.Accept` or `Glacier2.Filter.Address.Reject` is non-empty, and `hasDisallowedHost` runs only when this flag is set — `_rejectRules` is no longer the gate, so `ProxySizeMax` alone no longer triggers the non-canonical rejection. New test: with only `Glacier2.Filter.ProxySizeMax` configured, `-h 127.1` is accepted. The flag gates all of `hasDisallowedHost`, including the pre-existing 255-character host bound: that bound exists to keep address-rule matching cheap, so with no address rules configured it has nothing to protect and no longer applies (`ProxySizeMax` still bounds the entire proxy string).

Pushing shortly, together with the remaining review items.

pepone (approved):

cpp/test/Glacier2/staticFiltering/test.py
InsertCreativityHere: Yes, there isn't a good way to test this here. I wasn't aware of the limitation of our test setup beforehand.

cpp/src/Glacier2/ProxyVerifier.cpp
InsertCreativityHere: I don't think this applies here. If a function is declared in a header file and defined in a source, then yes, we should only apply a comment to the header file. But these functions do not exist in a header file, they only exist here in the source file. There is nowhere else a doc-comment could go for these functions. The real rule, as far as I know is: place doc-comments on function declarations. And here, this _is_ the declaration.

More importantly, there is literally no downside to making these doc-comments, it is a trivial change to the code (adding a slash), and there is an upside to doing it (better support with the tooling and extensions).

cpp/src/Glacier2/ProxyVerifier.cpp
InsertCreativityHere: I don't understand this distinction between "spells an IPv4 address" and "is an IPv4" address. But the function is named "isIPv4Numeral". By your logic this name is incorrect, and it should be named "spellsIPv4Numeral". My comment is simply matching the name of the function.

Either way, if that is your only real hangup, feel free to say "true if host spells an IPv4 address" instead. But I still think my suggested change manages to communicate exactly the same information as the current comment, with half of the reading required.

InsertCreativityHere (approved):
Still looks good to me.

Issue #6039, opened 2026-07-17
DataStorm/reliability C++ test hangs in 2-node anyKeyReconnect case — reader session destroyed after retry limit (SessionCreationException)


pepone:
The C++ `DataStorm/reliability` test hung in CI for ~5 minutes and was killed by the test runner. This is a **different subtest and mechanism** than the previously-fixed #5359 (which was the `string`-topic block / relay announcement re-sync).

## Environment

- Platform: Windows (GitHub Actions, `windows-2025`), x64 / Release.
- Run: https://github.com/zeroc-ice/ice/actions/runs/29525598321/job/87713226512 (2026-07-16). Result: `287 succeeded and 1 failed`.
- Test: `cpp/test/DataStorm/reliability`, case **"reader/writer as client with 2 nodes"** (non-reversed entry in `props`, `test.py`). Topology:

  ```
  reader-app ──dials──▶ relay node1 (:14011)
  writer-app ──dials──▶ relay node2 (:14012) ──dials──▶ node1
  ```
- The hang is in the **last** block of `Reader.cpp`/`Writer.cpp` — `anyKeyReconnect` / *"reconnect with an any-key writer without duplicate samples"* — which was added later by #5604 (2026-06-24), after #5359 was closed.

## Symptom

`reader.exe` blocks forever in `Reader::run` (`getNextUnread()` in the `anyKeyReconnect` block); the runner's `waitSuccess(timeout=30)` (`scripts/Expect.py:679`) raises `Expect.TIMEOUT`:

```
  File "D:\a\ice\ice\scripts\Expect.py", line 578, in wait
    raise TIMEOUT("timed wait exceeded timeout")
Expect.TIMEOUT: timed wait exceeded timeout
```

The active part of the test runs in ~15 s (localhost, in-memory), then hangs. The reader's subscriber session to `writer-app` churns and is finally **destroyed permanently** — after which there is **no further session/data activity for ~4.5 min**, only node-level validate-connection heartbeats, until the runner kills the process:

```
20:14:07.509  reader: Session: s/3: retrying connecting to 'writer-app:tcp ... -p 14012' in 2000 (ms), retry 4/6
                ::DataStormContract::SessionCreationException
20:14:09.512  reader: Session: s/3: retrying connecting to 'writer-app:tcp ... -p 14012' in 4000 (ms), retry 5/6
                ::DataStormContract::SessionCreationException
[...]

PR #6050, opened 2026-07-17
Return NodeSessionI load info by value

### LINKED ISSUE #6043: IceGrid `NodeSessionI::getLoadInfo` returns a reference after releasing its lock (data race)
> [!NOTE]

## Summary

`NodeSessionI::getLoadInfo` locks `_mutex` but returns a `const LoadInfo&` to the mutex-protected `_load`. The lock releases on return, so callers read/copy the three floats **unlocked**, racing `keepAlive`'s `_load = load` write under `_mutex`. The lock inside `getLoadInfo` protects nothing.

**File:** `cpp/src/IceGrid/NodeSessionI.cpp:218`
```cpp
const LoadInfo&
NodeSessionI::getLoadInfo() const
{
    lock_guard lock(_mutex);
    return _load;          // reference escapes; lock released on return
}
```
Declared `[[nodiscard]] const LoadInfo& getLoadInfo() const;` at `cpp/src/IceGrid/NodeSessionI.h:46`. `keepAlive` writes `_load` under `_mutex` (`NodeSessionI.cpp`), and `LoadInfo` is three non-atomic floats (`slice/IceGrid/Admin.ice`) → concurrent read/write with no synchronization is UB and can yield a torn snapshot.

Both callers copy by value **after** the lock is gone (so both race): `cpp/src/IceGrid/NodeCache.cpp:272` (`return _session->getLoadInfo();`, return type `LoadInfo`) and `cpp/src/IceGrid/PluginFacadeI.cpp:140`. The sibling accessors `getNode`/`getInfo`/`getProxy` return references to immutable members — `getLoadInfo` is the only one returning a reference to a member that `keepAlive` mutates.

Reachability: `keepAlive` is driven by the node's periodic session refresh while the registry queries load during server placement — different thread-pool threads.

## Fix

Return `LoadInfo` by value, copying under the lock:
```cpp
LoadInfo NodeSessionI::getLoadInfo() const { lock_guard lock(_mutex); return _load; }
```
(also removes a latent dangling-reference trap). Update the header signature accordingly.

**Severity:** Low (torn/stale load snapshot skews load-balancing input; genuine UB but not a crash). **Binary compatibility:** internal IceGrid class — signature change is source-internal.

bernardnormier:
`NodeSessionI::getLoadInfo` locked `_mutex` but returned a `const LoadInfo&` to the protected `_load` member. The lock is released when the function returns, so both callers copied the three floats unlocked, racing `keepAlive`'s `_load = load` write under the mutex — the lock inside the accessor protected nothing. The accessor now copies `_load` under the lock and returns it by value.

The same bug exists on the 3.7 branch (only the lock idiom differs), hence the backport-3.7 label.

Fixes #6043

### DIFF
--- a/cpp/src/IceGrid/NodeSessionI.cpp
+++ b/cpp/src/IceGrid/NodeSessionI.cpp
@@ -214,7 +214,7 @@ NodeSessionI::getInfo() const noexcept
     return _info;
 }
 
-const LoadInfo&
+LoadInfo
 NodeSessionI::getLoadInfo() const
 {
     lock_guard lock(_mutex);
--- a/cpp/src/IceGrid/NodeSessionI.h
+++ b/cpp/src/IceGrid/NodeSessionI.h
@@ -43,7 +43,7 @@ namespace IceGrid
 
         [[nodiscard]] const NodePrx& getNode() const;
         [[nodiscard]] const std::shared_ptr<InternalNodeInfo>& getInfo() const noexcept;
-        [[nodiscard]] const LoadInfo& getLoadInfo() const;
+        [[nodiscard]] LoadInfo getLoadInfo() const;
         [[nodiscard]] NodeSessionPrx getProxy() const;
 
         [[nodiscard]] bool isDestroyed() const;


pepone (approved):
Approving — this is a correct fix for a real data race.

Returning `const LoadInfo&` from an accessor whose only synchronization is a `lock_guard` released on return meant both callers read `_load` after the mutex was dropped, racing `keepAlive`'s `_load = load` write. Returning by value copies `_load` while the lock is still held (return-value construction is sequenced before the `lock_guard` destructor), so the accessor now actually protects the read.

Both callers already consume the result into a by-value `LoadInfo`, so the signature change is transparent, and `LoadInfo` is just three floats — no cost concern.

The one red check (`cpp-win32-release` / windows-2022) is unrelated: the build passed and the sole failing test was `DataStorm/reliability` timing out (`Expect.TIMEOUT`) in the "reader/writer as client with 2 nodes" case — most probably a dup of #6039. A re-run should clear it.

LGTM 👍

PR #6051, opened 2026-07-17
Contain exceptions thrown by Connection::close callbacks

### LINKED ISSUE #6033: C++ `Connection::close` can invoke user callbacks from a `noexcept` API (may `std::terminate` / interrupt finalization)
> [!NOTE]

## Summary

The callback-taking `Connection::close` overload is declared `noexcept`, and its documentation says the callbacks "may be called synchronously (from the calling thread) … when you call `close` on a connection that is already closed." Those user callbacks are invoked with **no exception guard** at two sites, unlike the `setCloseCallback` family (which is guarded). A throwing callback therefore (1) escapes the `noexcept` boundary → `std::terminate()` on the synchronous path, and (2) can interrupt finalization on the async path.

**File:** `cpp/include/Ice/Connection.h:67`
```cpp
virtual void
close(std::function<void()> response, std::function<void(std::exception_ptr)> exception) noexcept = 0;
```

**Path 1 — synchronous, already-closed (`cpp/src/Ice/ConnectionI.cpp:486`):** inside the `noexcept` implementation, on an already-closed connection the supplied callback is called directly:
```cpp
if (response) { response(); }              // :555
...
if (exception) { exception(closeException); }   // :562
```
No surrounding try/catch → a throwing callback escapes `noexcept` → `std::terminate()`.

**Path 2 — finalization (`ConnectionI::finish`, `cpp/src/Ice/ConnectionI.cpp`):** the stored `_onClosedList` callbacks are invoked, then the state is advanced:
```cpp
pair.first();            // :1782
...
pair.second(_exception); // :1789
...
setState(StateFinished); // :1807
```
`finish` is not `noexcept` (it runs as a thread-pool work item, so the throw is caught and logged rather than fatal), but it propagates out **before** `_onClosedList.clear()`, the `_closeCallback` invocation, and `setState(StateFinished)` — so the connection never reaches `StateFinished`, is never reaped, and any `waitUntilFinished()` blocks indefinitely.

The intended pattern already exists for the other callback family: `ConnectionI::closeCallback` wraps the `setCloseCallback` callback in a try/catch and logs. The `close()` response/exception callbacks simply don't route through an equivalent guard.

## Fix

Contain callback exceptions at both invocation sites (the synchronous `close` path at `:555/:562` and the `finish` path at `:1782/:1789`) — log and continue, preserving the one-callback contract — mirroring `ConnectionI::closeCallback`. Regression test: install a throwing response/exception callback on both an already-closed connection and an asynchronously-finalizing one; assert the exception is contained and the connection reaches its terminal state.

**Severity:** Low-Medium (Path 1 is a process abort, and the docs actively invite synchronous callback execution). **Binary compatibility:** unaffected (implementation-only).

bernardnormier:
The response and exception callbacks passed to the callback-based `Connection::close` overload must not throw — this PR documents that requirement in the API reference. It is a doc clarification plus hardening, not a bug fix: before and after, throwing from these callbacks is a contract violation.

Ice now also handles a violation gracefully instead of catastrophically. Previously, a throwing callback escaped the `noexcept` `close` on the synchronous already-closed path (`std::terminate`), and on the finalization path it prevented the remaining close callbacks from running and the connection from ever reaching `StateFinished`, hanging `Communicator::destroy`. Both paths now route the callbacks through the same catch-and-log guard that the `setCloseCallback` family already used (`ConnectionI::closeCallback`, renamed to an `executeCallback` overload) — matching the established AMI convention where throwing callbacks are caught and logged.

The new test verifies both paths: a throwing callback during an asynchronous close does not prevent the other close callbacks from running or the connection from finishing, and a throwing callback invoked synchronously on an already-closed connection does not terminate the process.

No changelog entry: only applications that violate the documented contract observe any change.

Fixes #6033

### DIFF
--- a/cpp/include/Ice/Connection.h
+++ b/cpp/include/Ice/Connection.h
@@ -62,7 +62,8 @@ namespace Ice
         /// @remark The response and exception callbacks may be called synchronously (from the calling thread); in
         /// particular, this occurs when you call `close` on a connection that is already closed. The implementation
         /// always calls one of the two callbacks once; it never calls both. If closing the connection takes longer than
-        /// the configured close timeout, the connection is aborted with a CloseTimeoutException.
+        /// the configured close timeout, the connection is aborted with a CloseTimeoutException. The response and
+        /// exception callbacks must not throw any exception.
         virtual void
         close(std::function<void()> response, std::function<void(std::exception_ptr)> exception) noexcept = 0;
 
--- a/cpp/src/Ice/ConnectionI.cpp
+++ b/cpp/src/Ice/ConnectionI.cpp
@@ -552,14 +552,14 @@ Ice::ConnectionI::close(function<void()> response, function<void(std::exception_
         {
             if (response)
             {
-                response();
+                executeCallback(response);
             }
         }
         else
         {
             if (exception)
             {
-                exception(closeException);
+                executeCallback([&] { exception(closeException); });
             }
         }
     }
@@ -776,7 +776,9 @@ Ice::ConnectionI::setCloseCallback(CloseCallback callback)
             if (callback)
             {
                 auto self = shared_from_this();
-                _threadPool->execute([self, callback = std::move(callback)]() { self->closeCallback(callback); }, self);
+                _threadPool->execute(
+                    [self, callback = std::move(callback)]() { self->executeCallback(callback); },
+                    self);
             }
         }
         else
@@ -795,11 +797,17 @@ Ice::ConnectionI::disableInactivityCheck() noexcept
 }
 
 void
-Ice::ConnectionI::closeCallback(const CloseCallback& callback)
+Ice::ConnectionI::executeCallback(const CloseCallback& callback) noexcept
+{
+    executeCallback([self = shared_from_this(), &callback] { callback(self); });
+}
+
+void
+Ice::ConnectionI::executeCallback(const std::function<void()>& callback) noexcept
 {
     try
     {
-        callback(shared_from_this());
+        callback();
     }
     catch (const std::exception& ex)
     {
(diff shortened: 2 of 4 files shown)


pepone:
The new test isn't deterministic: nothing keeps the connection open, so the second `close` can find the connection already closed and run its callback synchronously instead of queuing it after the throwing one. And since both of the first close's callbacks throw, the promises can't tell whether `first` ran before `second` anyway.

Suggest keeping the connection open with a `sleepAsync` (like the "testing connection close" local case above) so both calls queue their callbacks, and recording the invocation order:

```cpp
auto con = p2->ice_getConnection();

vector<string> order;
mutex orderMutex;
auto record = [&](const char* event)
{
    lock_guard<mutex> lock(orderMutex);
    order.emplace_back(event);
};

promise<void> closed;
con->setCloseCallback(
    [&](const ConnectionPtr&)
    {
        record("closed");
        closed.set_value();
    });

// Keeps the connection open until both close calls below have queued their callbacks.
auto sleepFuture = p2->sleepAsync(100);

con->close(
    [&]
    {
        record("first");
        throw std::runtime_error("from response callback");
    },
    [&](exception_ptr)
    {
        record("first");
        throw std::runtime_error("from exception callback");
    });
con->close([&] { record("second"); }, [&](exception_ptr) { record("second"); });

sleepFuture.get();
closed.get_future().get();
test((order == vector<string>{"first", "second", "closed"}));
```

bernardnormier:
Good catch — applied your suggestion in 32e178b: a pending `sleepAsync` keeps the connection open so both `close` calls queue their callbacks, and the test now verifies the `first`/`second`/`closed` invocation order. I kept the final synchronous `close` on the already-closed connection, since it covers the other fixed path (the exception escaping the `noexcept` `close`).

pepone (approved):

PR #6052, opened 2026-07-17
Add precondition when marshaling a fixed proxy in Swift

### LINKED ISSUE #6035: Swift: marshaling a fixed proxy lets a C++ `FixedProxyException` escape into non-throwing Swift → crash
> [!NOTE]

## Summary

Marshaling a **fixed proxy** (`proxy.ice_fixed(connection)`) from Swift lets a C++ `FixedProxyException` unwind through non-throwing Swift/ObjC frames — undefined behavior, in practice a crash — where the C++ API produces a *catchable* exception. The Objective-C++ bridge method `ObjectPrx.ice_write` calls `out.write(_prx)` with no try/catch and no `NSError**` parameter. This is the direct, untreated twin of the `proxyToProperty` fix (#5707), which guarded the sibling `FixedReference::toProperty` throw.

**File:** `swift/src/IceImpl/ObjectPrx.mm:496`
```objc
- (void)ice_write:(id<ICEOutputStreamHelper>)os
    encodingMajor:(std::uint8_t)encodingMajor
    encodingMinor:(std::uint8_t)encodingMinor
{
    ...
    Ice::OutputStream out(communicator, encoding);
    out.write(_prx);            // :506 — can throw, no try/catch, and this method has no NSError** out-param
    ...
}
```

Chain (every hop verified):
1. Swift `OutputStream.write(_ v: ObjectPrx?)` is **non-throwing** (`swift/src/Ice/OutputStream.swift:492`) → `ObjectPrxI.ice_write(to:)` (also non-throwing, `swift/src/Ice/Proxy.swift`) → the ObjC `ice_write` above.
2. `out.write(_prx)` → `OutputStream::writeProxy` → `ObjectPrx::_write` → `_reference->streamWrite()`.
3. For a fixed reference, `FixedReference::streamWrite` throws `FixedProxyException` (`cpp/src/Ice/Reference.cpp:570`). Its sibling `FixedReference::toProperty` (immediately below) throws the same exception and **is** guarded, via `Communicator.mm:91` `proxyToProperty` (the #5707 fix with an `error:` param + `try/catch → convertException`).
4. Fixed proxies are reachable from public API: `proxy.ice_fixed(connection)` (`swift/src/Ice/Proxy.swift`). Passing one as an operation parameter, or embedding it in any Slice type that gets written, runs this path.

Because the throw happens in the ObjC++ `ice_write` (no local catch) and unwinds through non-throwing Swift frames, it is UB → `std::terminate`/crash.

## Fix

Give `ice_write` an `NSError**` out-param and wrap `out.write(_prx)` in `try { … } catch (...) { *error = convertException(std::current_exception()); }`, and propagate it through the Swift `OutputStream.write(ObjectPrx?)` / `ObjectPrxI.ice_write(to:)` path (as C++ does, via a throwable) — mirroring exactly what #5707 did for `proxyToProperty`. Regression test: marshal an `ice_fixed` proxy and assert a catchable error, not a crash.

**Severity:** Medium (crash where C++ gives a catchable exception; trigger is marshaling a fixed proxy). **Binary compatibility:** the ObjC bridge is internal (`IceImpl`); the Swift `OutputStream.write(ObjectPrx?)` signature would gain `throws`.

bernardnormier:
Marshaling a fixed proxy is not supported. In the Swift mapping, doing so let the C++ `FixedProxyException` thrown by `FixedReference::streamWrite` unwind through non-throwing Swift and Objective-C++ frames — undefined behavior, in practice a crash without a usable diagnostic.

The Swift mapping already maps the other fixed-proxy misuses (`ice_adapterId`, `ice_endpoints`, `ice_router`, ...) — which throw `FixedProxyException` in C++ — to preconditions. This PR gives `OutputStream.write(ObjectPrx?)` the same treatment: `precondition(!prxImpl.ice_isFixed(), "Cannot marshal a fixed proxy")`, so the misuse now fails fast with a clear message. No signatures change.

No test: a precondition failure aborts the process, so it is not testable in-process, like the existing fixed-proxy preconditions. No changelog entry: only applications that marshal a fixed proxy — a contract violation — observe any change.

Fixes #6035

### DIFF
--- a/swift/src/Ice/OutputStream.swift
+++ b/swift/src/Ice/OutputStream.swift
@@ -489,8 +489,10 @@ extension OutputStream {
     /// Writes a proxy to the stream.
     ///
     /// - Parameter v: The proxy to write.
+    /// - Precondition: `v` is not a fixed proxy.
     public func write(_ v: ObjectPrx?) {
         if let prxImpl = v as? ObjectPrxI {
+            precondition(!prxImpl.ice_isFixed(), "Cannot marshal a fixed proxy")
             prxImpl.ice_write(to: self)
         } else {
             // A nil proxy is represented by an Identity with empty name and category fields.


pepone (approved):
Correct, minimal, and idiomatic. The precondition sits at the single marshaling chokepoint (`ice_write(to:)` has one caller, and all proxy sequence/dict/member writes route through `OutputStream.write`), so it covers every path. The ObjC++ `ice_write` has no `@try/@catch`, so today a fixed proxy lets the C++ `FixedProxyException` unwind through non-throwing frames — UB/opaque crash; this makes it fail fast with a clear message. Matches the existing fixed-proxy precondition convention. Approve.

PR #6054, opened 2026-07-17
Remove unused self capture in Swift CtrlCHandler callback

### LINKED ISSUE #6047: Swift `CtrlCHandler.setCallback` retain cycle from an unnecessary `self` capture
> [!NOTE]

## Summary

The block passed to the C++ `CtrlCHandler::setCallback` captures `self` strongly, but the block body never uses `self` — only `callback` and `signal`. Since `_cppObject` is a by-value ivar of `self` that owns the block, this forms a retain cycle `self → _cppObject → std::function → block → self`, so the `ICECtrlCHandler` (and captured Swift block) leak until `setCallback(nil)` is called.

**File:** `swift/src/IceImpl/CtlCHandler.mm:16`
```objc
self->_cppObject.setCallback(
    [callback, self](int signal)        // :16 — `self` captured but never used
    {
        @autoreleasepool { callback(signal); }
    });
```

## Fix

Drop `self` from the capture list: `[callback](int signal){ ... }`.

**Severity:** Low (macOS-only; `CtrlCHandler` is typically a process-lifetime object, so the leak is bounded — but it is a genuine unintended strong capture). **Binary compatibility:** unaffected.

bernardnormier:
The lambda that `ICECtrlCHandler.setCallback` passes to the C++ `CtrlCHandler::setCallback` captured `self` without using it. The capture is a leftover from the one-shot `catchSignal` lambda it was copied from (#3596), which reset the callback from its body and therefore needed `self`.

Since `_cppObject` is a by-value ivar of `self` that owns the lambda, the unused capture formed a permanent retain cycle (`self → _cppObject → std::function → block → self`), leaking the `ICECtrlCHandler` and the captured Swift callback. Dropping `self` from the capture list breaks the cycle; lifetime remains sound because the lambda is owned by `_cppObject`, which cannot outlive `self`.

Fixes #6047

### DIFF
--- a/swift/src/IceImpl/CtlCHandler.mm
+++ b/swift/src/IceImpl/CtlCHandler.mm
@@ -13,7 +13,7 @@
     if (callback)
     {
         self->_cppObject.setCallback(
-            [callback, self](int signal)
+            [callback](int signal)
             {
                 // This callback executes in the C++ CtrlCHandler thread.
                 // We need an autorelease pool in case the callback creates autorelease objects.


pepone (approved):
Verified the fix end-to-end. The retain cycle is genuine: `ICECtrlCHandler` owns `_cppObject` as a by-value ivar, which stores the lambda in a `std::function`, and under ARC (enabled by default for this SwiftPM target) the `self` capture retained the handler — closing the loop `self → _cppObject → std::function → lambda → self` so its refcount never hit zero.

Removing the capture is safe:
- The lambda body only references `callback` and `signal`, never `self`, so it's behavior-preserving.
- Lifetime stays sound — the lambda is owned by `_cppObject`, which is destroyed during `self`'s `dealloc`, so it can't outlive `self` (and wouldn't touch it if it did).
- The `callback` block is captured independently and retained by the lambda, so it isn't released prematurely.

Clean, minimal fix with a great explanation. LGTM.

Issue #6058, opened 2026-07-17
Fail with PropertyException on an invalid property value instead of silently coercing or returning the default


bernardnormier:
Across the property accessors, a malformed property value is silently degraded rather than reported: a non-integer or out-of-range integer, a malformed `Ice.RetryIntervals` token (coerced to `0`), and a list value with mismatched quotes or only separators (returns the default or an empty list). In each case, the caller cannot distinguish "property not set" from "property set to a malformed value."

For 3.9, a malformed property value should fail with a `PropertyException` naming the property — consistent with how `getPropertyAsInt` already rejects a non-integer — uniformly across all language mappings. This is a documented-behavior change, so it belongs in 3.9.

Subsumes/relates to:
- #6010 — `getPropertyAsList` returns the default value on mismatched quotes instead of failing (specific instance, kept open and tracked under this issue).
- #6042 — malformed `Ice.RetryIntervals` token is coerced to `0`; the 3.8.3 fix warns on coercion, and this issue is the strict-mode (reject-at-startup) successor.
- #6009 — Glacier2 filter properties work around this today with custom parsing that rejects unmatched quotes at startup.
- #6040 — `getPropertyAsListWithDefault` separators-only handling.

PR #6059, opened 2026-07-17
Warn on an invalid Ice.RetryIntervals value in C++, C#, and Java

### LINKED ISSUE #6042: Malformed `Ice.RetryIntervals` token: C++/C#/Java silently coerce to 0, JS rejects (cross-language inconsistency)
> [!NOTE]

## Summary

A malformed token in `Ice.RetryIntervals` (e.g. `abc`, `1abc`, `1.5`) is silently coerced to `0` ("retry immediately") by C++, C#, and Java, whereas JS rejects it with a `ParseException` at initialization. The lenient languages give no diagnostic that the configuration is wrong.

- **C++** `cpp/src/Ice/Instance.cpp:1267` — `if (!(value >> v) || !value.eof()) { v = 0; }`
- **C#** `csharp/src/Ice/Internal/Instance.cs:827` — `catch (System.FormatException) { v = 0; }`
- **Java** `java/.../Ice/Instance.java:745` — `catch (NumberFormatException ex) { v = 0; }`
- **JS** `js/packages/ice/src/Ice/InstanceExtensions.js:284` — `StringUtil.toInt32(...)` throws → `throw new ParseException(...)` (`:286`), init fails.

Framing note (from review): this is a **robustness/consistency** issue, not a DoS. `0` is already the clamp all three lenient languages apply to any non-positive value, so a malformed token is indistinguishable from an intentional `0`, and the retry *count* stays bounded by the list length (see `cpp/src/Ice/OutgoingAsync.cpp`) — no unbounded/immediate-loop behavior.

## Fix

Pick one contract and apply it everywhere: either warn-and-coerce (log the malformed token) or reject at startup like JS. Either way, make the four mappings agree.

**Severity:** Low (silent config degradation + cross-language inconsistency). **Binary compatibility:** unaffected.

bernardnormier:
C++, C#, and Java silently coerced an invalid `Ice.RetryIntervals` token to `0` ("retry immediately"), so a malformed configuration value was indistinguishable from an intentional `0` and gave no diagnostic. They now log a warning when coercing, while leaving a legitimate `0` and the leading `-1` (no retries) silent.

"Invalid" now uniformly covers a non-integer token, an out-of-range value, and an unexpected negative value (any negative other than the leading `-1` sentinel — previously clamped to `0` with no warning). The C# change additionally coerces an out-of-range value to `0` like C++ and Java, instead of letting `int.Parse`'s `OverflowException` escape.

JS is unchanged: it continues to reject an invalid value with a `ParseException` at startup. Unifying the four mappings on that stricter reject-at-startup contract is a documented-behavior change tracked for 3.9 in #6058.

Fixes #6042

### DIFF
--- a/cpp/src/Ice/Instance.cpp
+++ b/cpp/src/Ice/Instance.cpp
@@ -1262,20 +1262,26 @@ IceInternal::Instance::initialize(const Ice::CommunicatorPtr& communicator)
                 istringstream value(retryValue);
 
                 int v;
-                if (!(value >> v) || !value.eof())
-                {
-                    v = 0;
-                }
+                bool parsed = static_cast<bool>(value >> v) && value.eof();
 
                 //
                 // If -1 is the first value, no retry and wait intervals.
                 //
-                if (v == -1 && _retryIntervals.empty())
+                if (parsed && v == -1 && _retryIntervals.empty())
                 {
                     break;
                 }
 
-                _retryIntervals.push_back(v > 0 ? v : 0);
+                // Any value that is not a non-negative integer (or the leading -1 handled above) is invalid and
+                // treated as 0.
+                if (!parsed || v < 0)
+                {
+                    Warning out(_initData.logger);
+                    out << "invalid value '" << retryValue << "' in property Ice.RetryIntervals, assuming 0";
+                    v = 0;
+                }
+
+                _retryIntervals.push_back(v);
             }
         }
 
--- a/csharp/src/Ice/Internal/Instance.cs
+++ b/csharp/src/Ice/Internal/Instance.cs
@@ -819,26 +819,38 @@ public sealed class Instance
                 for (int i = 0; i < retryValues.Length; i++)
                 {
                     int v;
+                    bool parsed;
 
                     try
                     {
                         v = int.Parse(retryValues[i], CultureInfo.InvariantCulture);
+                        parsed = true;
                     }
-                    catch (System.FormatException)
+                    catch (System.Exception ex) when (ex is System.FormatException or System.OverflowException)
                     {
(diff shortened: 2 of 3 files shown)


pepone (approved):
LGTM.

Two non-blocking suggestions:

1. In C#, the try/catch + `parsed` flag can be a single `TryParse`, with the same semantics (returns `false` and sets `v = 0` on both format and overflow errors):

```csharp
bool parsed = int.TryParse(retryValues[i], NumberStyles.Integer, CultureInfo.InvariantCulture, out int v);
```

2. "Invalid" is still not fully uniform for whitespace inside quoted tokens: with `Ice.RetryIntervals=' 5'`, Java warns and coerces to 0 while C++ and C# accept 5; with `'5 '`, C# accepts 5 while C++ and Java warn. Trimming each token before parsing would close this — also fine to leave it to the 3.9 unification (#6058).

bernardnormier:
Regarding the whitespace divergence (suggestion 2): it's only reachable with quoted tokens. `getIcePropertyAsList` already consumes unquoted whitespace as the token separator, so `Ice.RetryIntervals=  5 , 7` parses as `["5", "7"]` in all three mappings — there's nothing to trim. With `' 5'`, the space is preserved by design: the documented purpose of quoting in `getPropertyAsList` is to let a token contain whitespace. The remaining divergence is in each language's integer parsing of that token (Java rejects any whitespace, C++ skips leading whitespace, C# allows leading and trailing).

Trimming before parsing would work against the documented quoting semantics, so I'd rather not add it here. And in 3.9, if you go out of your way to write `' 5'`, you'll get an error, not a successful parse — so leaving this to #6058.

PR #6060, opened 2026-07-17
Fix loading the wrong config file with adjacent --Ice.Config arguments (C#, Java)

### LINKED ISSUE #6031: Adjacent `--Ice.Config` command-line arguments load the wrong config file (C# and Java)
> [!NOTE]

## Summary

When two `--Ice.Config` options are passed as **adjacent** command-line arguments (e.g. `--Ice.Config=a.cfg --Ice.Config=b.cfg`), the C# and Java runtimes load the values from `a.cfg` but set the `Ice.Config` property to `b.cfg` — the second file is never loaded. C++ loads the last file (`b.cfg`) as expected. The cause is a forward loop that removes the current element from the argument array and then increments the index, so the element that shifts into the freed slot is skipped.

**File:** `csharp/src/Ice/Properties.cs:621`
```csharp
private void loadArgs(ref string[] args)
{
    ...
    for (int i = 0; i < args.Length; i++)
    {
        if (args[i].StartsWith("--Ice.Config", StringComparison.Ordinal))
        {
            ...
            parseLine(line[2..]);
            loadConfigFiles = true;

            string[] arr = new string[args.Length - 1];
            Array.Copy(args, 0, arr, 0, i);
            if (i < args.Length - 1)
            {
                Array.Copy(args, i + 1, arr, i, args.Length - i - 1);
            }
            args = arr;   // element that was at i+1 is now at i; the i++ then skips it
        }
    }
```

Trace for `--Ice.Config=a.cfg --Ice.Config=b.cfg`: at `i=0` the first arg is parsed (`Ice.Config`→`a.cfg`) and removed; `args` becomes `["--Ice.Config=b.cfg"]`; `i` becomes `1`; `1 < 1` is false → the loop ends without ever seeing the second arg. `loadConfig()` then reads `Ice.Config`=`a.cfg` and loads **a.cfg**; the later `parseIceCommandLineOptions` pass sets `Ice.Config`=`b.cfg` but does not load it. Final state: the property says `b.cfg`, the loaded values come from `a.cfg`.

**File:** `java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java:591` — identical logic (`args = arr;` at `:608`).

C++ `loadArgs` (`cpp/src/Ice/Properties.cpp`) is correct: it iterates with an iterator and builds a fresh vector, calling `setProperty` for every match, so last-wins holds. The C# sibling `parseCommandLineOptions` also uses the safe build-a-new-list pattern — `loadArgs` is the lone deviation.

## Fix

Don't mutate the array being forward-scanned. Build a new list of the non-`--Ice.Config` args (mirroring C++ `loadArgs` / the existing `parseCommandLineOptions`), or decrement the index after removal. Regression test: pass two **adjacent** `--Ice.Config` args and assert the second file's values are loaded and the first file's are absent, in both C# and Java.

**Severity:** Medium-Low (correctness; the trigger is an unusual argv shape — two separate `--Ice.Config` options rather than the documented comma-separated `--Ice.Config=a,b`). **Binary compatibility:** unaffected.

bernardnormier:
When two `--Ice.Config` options were passed as adjacent command-line arguments (e.g. `--Ice.Config=a.cfg --Ice.Config=b.cfg`), the C# and Java runtimes loaded `a.cfg` but set the `Ice.Config` property to `b.cfg` — so the loaded values and the property disagreed, and the second file was never loaded. C++ loads the last file, as expected.

The cause was a forward loop in `loadArgs` that removed the current element from the argument array and then incremented the index, skipping the element that shifted into the freed slot. `loadArgs` now builds a fresh list of the non-`--Ice.Config` arguments — mirroring C++ `loadArgs` and the sibling `parseCommandLineOptions` — so the last file wins.

Added a regression case to the C# and Java properties tests: two adjacent `--Ice.Config` arguments load the second file's values, not the first, and the arguments are consumed.

Fixes #6031

### DIFF
--- a/changelog.d/csharp/config-args-adjacent.md
+++ b/changelog.d/csharp/config-args-adjacent.md
@@ -1,0 +1,2 @@
+- Fixed a bug where specifying the `--Ice.Config` command-line option more than once loaded the wrong configuration
+  file: Ice loaded the first file instead of the last.
--- a/changelog.d/java/config-args-adjacent.md
+++ b/changelog.d/java/config-args-adjacent.md
@@ -1,0 +1,2 @@
+- Fixed a bug where specifying the `--Ice.Config` command-line option more than once loaded the wrong configuration
+  file: Ice loaded the first file instead of the last.
--- a/csharp/src/Ice/Properties.cs
+++ b/csharp/src/Ice/Properties.cs
@@ -621,28 +621,26 @@ public sealed class Properties
     private void loadArgs(ref string[] args)
     {
         bool loadConfigFiles = false;
+        List<string> remainingArgs = [];
 
-        for (int i = 0; i < args.Length; i++)
+        foreach (string arg in args)
         {
-            if (args[i].StartsWith("--Ice.Config", StringComparison.Ordinal))
+            if (arg.StartsWith("--Ice.Config", StringComparison.Ordinal))
             {
-                string line = args[i];
+                string line = arg;
                 if (!line.Contains('=', StringComparison.Ordinal))
                 {
                     line += "=1";
                 }
                 parseLine(line[2..]);
                 loadConfigFiles = true;
-
-                string[] arr = new string[args.Length - 1];
-                Array.Copy(args, 0, arr, 0, i);
-                if (i < args.Length - 1)
-                {
-                    Array.Copy(args, i + 1, arr, i, args.Length - i - 1);
-                }
-                args = arr;
+            }
+            else
+            {
+                remainingArgs.Add(arg);
             }
         }
+        args = [.. remainingArgs];
 
         if (!loadConfigFiles)
         {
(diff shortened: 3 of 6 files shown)


pepone (approved):
LGTM. The refactor mirrors C++ `loadArgs` exactly: parse every `--Ice.Config` through `parseLine` (last wins), collect the rest, then `loadConfig` + `parseIceCommandLineOptions` — so the loaded file and the `Ice.Config` property can no longer disagree. Root cause (remove-then-`i++` skipping the shifted element) is right, and the scope is correct: C++ was already fine, and JS/Python/Ruby/PHP/Swift/MATLAB don't hit this path. Test is a clean red→green and also asserts the args are consumed.

Optional nit, non-blocking: the changelog says "more than once," but only *adjacent* duplicates actually misloaded — fine to leave as-is since the fix makes all cases last-wins.

PR #6061, opened 2026-07-17
Fix two C# property accessor divergences

### LINKED ISSUE #6032: C# `getPropertyAsIntWithDefault` throws `OverflowException` for out-of-range values, bypassing communicator-init cleanup
> [!NOTE]

## Summary

C# `getPropertyAsIntWithDefault` catches only `FormatException`, so a numeric-but-out-of-`Int32`-range property value (e.g. `99999999999`) makes `int.Parse` throw `System.OverflowException`, which propagates uncaught instead of the documented `PropertyException`. Worse, when this happens during `Instance.initialize()`, the phase's cleanup handler is typed `catch (Ice.LocalException)` — and `OverflowException` is not an `Ice.LocalException` — so `destroy()` is skipped and the partially-constructed communicator leaks. C++ (`getPropertyAsIntWithDefault` in `cpp/src/Ice/Properties.cpp`) throws `PropertyException` on overflow, as documented.

**File:** `csharp/src/Ice/Properties.cs:158`
```csharp
public int getPropertyAsIntWithDefault(string key, int value)
{
    ...
    try
    {
        return int.Parse(pv.value, CultureInfo.InvariantCulture);   // :169
    }
    catch (FormatException)                                          // :171 — only FormatException
    {
        throw new PropertyException($"property '{key}' has an invalid integer value: '{pv.value}'");
    }
}
```
`int.Parse` throws **two** exceptions: `FormatException` for non-numeric text and `OverflowException` (a `SystemException`, not derived from `FormatException`) for a well-formed number outside `Int32` range. The catch misses the latter. The XML doc contract promises `PropertyException` "if the property value is not a valid integer"; an out-of-range value is not a valid integer.

**Cleanup bypass — `csharp/src/Ice/Internal/Instance.cs`:** `initialize()` reads many `Ice.*` int properties after allocating managers/factories (e.g. `Ice.MessageSizeMax` `:676`, `Ice.BatchAutoFlushSize` `:692`, `Ice.ClassGraphDepthMax` `:708`, `Ice.IPv4/IPv6` `:761`). Its only handler is:
```csharp
catch (Ice.LocalException)   // Instance.cs:850
{
    destroy();               // :852 — frees the partially-built instance
    throw;
}
```
`PropertyException` **is** a `LocalException` (`LocalExceptions.cs:980`), so the intended path cleans up. `OverflowException` slips past the typed catch, `destroy()` is skipped, and the allocated managers/factories leak.

## Fix

Catch `OverflowException` alongside `FormatException` in `getPropertyAsIntWithDefault` (and the sibling `getIcePropertyAsInt` default-parse) and convert to `PropertyException`. Regression test: `communicator, --Ice.PreferIPv6Address=99999999999` should raise a `PropertyException` and run `destroy()`, not leak.

**Severity:** Medium-Low (contract break + init-time resource leak; requires an out-of-range Ice int property). **Binary compatibility:** unaffected.

### LINKED ISSUE #6040: C# `getPropertyAsListWithDefault` ignores the default for a separators-only value
> [!NOTE]

## Summary

C# `getPropertyAsListWithDefault` is the only mapping that does **not** substitute the caller's default when a present property parses to an empty list. C++, Java, and JS all do. So a stored, non-empty, separators-only value (whitespace/commas, e.g. `"   "` or `","`) makes C# return `[]` while the others return the default.

**File:** `csharp/src/Ice/Properties.cs:218`
```csharp
string[] result = Ice.UtilInternal.StringUtil.splitString(pv.value, ", \t\r\n");
if (result == null)                     // :232 — mismatched quotes → default
{
    ...
    return value;
}
else
{
    return result;                       // :240 — no `result.Length == 0` → default clause
}
```
The peers have the missing clause: C++ `if (result.size() == 0) result = value;` (`cpp/src/Ice/Properties.cpp:320`), Java `if (result.length == 0) result = value;` (`java/.../Ice/Properties.java:305`), JS (`js/.../Ice/Properties.js:118`).

Note the trigger is narrow: an *explicitly empty* value (`""`) is removed by `setProperty` in all four languages, so it returns the default everywhere. The divergence only shows for a **stored, non-empty, separators-only** value (which `setProperty` keeps, since it only checks length, and which `splitString` reduces to an empty list). A regression test must use `"   "` / `","`, not `""`.

## Fix

Add the missing clause before `return result;` in C# `getPropertyAsListWithDefault`:
```csharp
if (result.Length == 0) { return value; }
```
mirroring C++/Java/JS.

**Severity:** Low (cross-mapping consistency; trigger is an unusual config value). **Binary compatibility:** unaffected.

bernardnormier:
Two small, independent C# `Properties` accessor fixes where C# diverged from the other mappings, one commit each:

- **#6032** — `getPropertyAsIntWithDefault` caught only `FormatException`, so an out-of-range value made `int.Parse` throw an uncaught `OverflowException` instead of the documented `PropertyException`. Since `OverflowException` is not a `LocalException`, it also bypassed `Instance.initialize`'s `catch (LocalException)` cleanup, leaking the partially-constructed communicator. The method now catches `OverflowException` too and converts it to a `PropertyException`, matching C++.
- **#6040** — `getPropertyAsListWithDefault` returned an empty array when a present, non-empty value parsed to an empty list (a separators-only value such as `,`), instead of returning the caller's default like C++, Java, and JS. It now returns the default.

Added regression cases to the C# properties test for both.

Fixes #6032
Fixes #6040

### DIFF
--- a/csharp/src/Ice/Properties.cs
+++ b/csharp/src/Ice/Properties.cs
@@ -168,7 +168,7 @@ public sealed class Properties
             {
                 return int.Parse(pv.value, CultureInfo.InvariantCulture);
             }
-            catch (FormatException)
+            catch (System.Exception ex) when (ex is System.FormatException or System.OverflowException)
             {
                 throw new PropertyException($"property '{key}' has an invalid integer value: '{pv.value}'");
             }
--- a/csharp/src/Ice/Properties.cs
+++ b/csharp/src/Ice/Properties.cs
@@ -235,10 +235,7 @@ public sealed class Properties
                     $"mismatched quotes in property {key}'s value, returning default value");
                 return value;
             }
-            else
-            {
-                return result;
-            }
+            return result.Length == 0 ? value : result;
         }
     }
 
--- a/csharp/test/Ice/properties/Client.cs
+++ b/csharp/test/Ice/properties/Client.cs
@@ -225,6 +225,15 @@ public class Client : Test.TestHelper
         }
 
         {
+            Console.Out.Write("testing that a separators-only list value returns the default... ");
+            var properties = new Ice.Properties();
+            properties.setProperty("Foo", ",");
+            string[] result = properties.getPropertyAsListWithDefault("Foo", ["a", "b"]);
+            test(result.Length == 2 && result[0] == "a" && result[1] == "b");
+            Console.Out.WriteLine("ok");
+        }
+
+        {
             Console.Out.Write("testing Ice.ProgramName default... ");
             Console.Out.Flush();
             using var communicator = new Ice.Communicator();
--- a/csharp/test/Ice/properties/Client.cs
+++ b/csharp/test/Ice/properties/Client.cs
@@ -210,6 +210,21 @@ public class Client : Test.TestHelper
         }
 
         {
+            Console.Out.Write("testing that reading an out-of-range value as an int throws... ");
+            var properties = new Ice.Properties();
+            properties.setProperty("Foo", "99999999999");
+            try
+            {
+                properties.getPropertyAsIntWithDefault("Foo", 0);
+                test(false);
+            }
+            catch (Ice.PropertyException)
+            {
+            }
+            Console.Out.WriteLine("ok");
+        }
+
+        {
             Console.Out.Write("testing Ice.ProgramName default... ");
             Console.Out.Flush();
             using var communicator = new Ice.Communicator();


InsertCreativityHere (approved):
We should consider logging a warning or throwing an exception when property lists contains multiple separators in a row `[, ,,]`. This can only ever be a typo IMO.

pepone (approved):

bernardnormier:
> We should consider logging a warning or throwing an exception when property lists contains multiple separators in a row `[, ,,]`. This can only ever be a typo IMO.

Should be included in #6058.

PR #6063, opened 2026-07-17
Range-check the default value in Ruby and PHP getPropertyAsIntWithDefault

### LINKED ISSUE #6041: Ruby/PHP `getPropertyAsIntWithDefault` silently narrows an out-of-range default value
> [!NOTE]

## Summary

In the Ruby and PHP bindings, `getPropertyAsIntWithDefault` accepts a wide (64-bit) integer default and silently casts it to `int32_t`, so an out-of-range **default** is truncated with no error. (The parsed *property value* is fine — it goes through the C++ core, which validates and throws `PropertyException`; only the caller-supplied default is narrowed.)

**File:** `php/src/Properties.cpp:240`
```cpp
// TODO: Range check
int32_t val = _this->getPropertyAsIntWithDefault(propName, static_cast<int32_t>(def));  // :241 — def is zend_long (64-bit)
```
**File:** `ruby/src/IceRuby/Properties.cpp:159`
```cpp
int32_t d = static_cast<int32_t>(getInteger(def));   // getInteger returns 64-bit long
int32_t v = p->getPropertyAsIntWithDefault(k, d);
```
The C++ core signature takes a 32-bit default, so native callers can't introduce this at that layer; the narrowing is binding-specific, from accepting a wider dynamic integer and casting down. `props.getPropertyAsIntWithDefault("Unset.Key", 99999999999)` returns `1215752191` (mod 2³²) instead of raising.

## Fix

Range-check the default before narrowing (the PHP `// TODO: Range check` marks the spot) and raise on out-of-range, matching how the property value itself is validated.

**Severity:** Low (the truncated input is an app-supplied default, not attacker/config-controlled data; result is a wrong-but-defined value). **Binary compatibility:** unaffected.

bernardnormier:
The Ruby and PHP bindings accept a wide (64-bit) integer default for `getPropertyAsIntWithDefault` and silently cast it to `int32_t`, so an out-of-range default was truncated with no error — for example `getPropertyAsIntWithDefault("Unset.Key", 99999999999)` returned `1215752191` (mod 2³²). The parsed property value is unaffected: it goes through the C++ core, which validates and throws. Only the caller-supplied default was narrowed, from accepting a wider dynamic integer and casting it down.

The bindings now range-check the default before narrowing and raise on out-of-range — `RangeError` in Ruby, `InvalidArgumentException` in PHP — matching how the property value itself is validated.

Added a regression case to the Ruby and PHP properties tests.

Fixes #6041

### DIFF
--- a/php/src/Properties.cpp
+++ b/php/src/Properties.cpp
@@ -237,7 +237,7 @@ ZEND_METHOD(Ice_Properties, getPropertyAsIntWithDefault)
     string_view propName(name, nameLen);
     if (def < INT32_MIN || def > INT32_MAX)
     {
-        invalidArgument("the value argument is out of the range of a 32-bit integer");
+        invalidArgument("the default value is out of the range of a 32-bit integer");
         RETURN_NULL();
     }
     try
--- a/php/src/Properties.cpp
+++ b/php/src/Properties.cpp
@@ -235,9 +235,13 @@ ZEND_METHOD(Ice_Properties, getPropertyAsIntWithDefault)
     assert(_this);
 
     string_view propName(name, nameLen);
+    if (def < INT32_MIN || def > INT32_MAX)
+    {
+        invalidArgument("the value argument is out of the range of a 32-bit integer");
+        RETURN_NULL();
+    }
     try
     {
-        // TODO: Range check
         int32_t val = _this->getPropertyAsIntWithDefault(propName, static_cast<int32_t>(def));
         RETURN_LONG(val);
     }
--- a/php/test/Ice/properties/Client.php
+++ b/php/test/Ice/properties/Client.php
@@ -79,6 +79,15 @@ class Client extends TestHelper
         }
         echo "ok\n";
 
+        echo "testing that an out-of-range default value throws... ";
+        $properties = Ice\createProperties();
+        try {
+            $properties->getPropertyAsIntWithDefault("Unset.Key", 99999999999);
+            test(False);
+        } catch (\InvalidArgumentException $ex) {
+        }
+        echo "ok\n";
+
         echo "testing Ice.initialize with args... ";
         $args = ["--Foo.Bar=1", "--Foo.Bar=2", "--Ice.Trace.Network=3"];
         $communicator = Ice\initialize($args);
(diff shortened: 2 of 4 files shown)


pepone (approved):
LGTM. The fix matches the setBufferSize precedent in both bindings and the tests follow the file conventions.

A few non-blocking notes:

- The message says "the value argument", but the PHP reflected signature is
  `getPropertyAsIntWithDefault($name, $defaultValue)` — named-argument calls use `defaultValue:`,
  so PHP users are pointed at a parameter that doesn't exist. Since Ruby exposes no parameter name
  at all, both messages could read "the default value is out of the range of a 32-bit integer",
  name-free like the setBufferSize checks.

- In Ruby, `getLong` instead of `getInteger` would make the check independent of `sizeof(long)`
  and give a consistent RangeError up to ±2^63 (a default >= 2^63 currently raises TypeError
  "unable to convert value to an int" from inside getInteger). Same applies to setBufferSize,
  so feel free to ignore here.

- The same silent-narrowing class survives elsewhere in these two extensions:
  `ice_invocationTimeout` (php/src/Proxy.cpp:1127, still carrying the twin `// TODO: range check?`)
  and `ice_locatorCacheTimeout` (php/src/Proxy.cpp:470, ruby/src/IceRuby/Proxy.cpp:319/339) truncate
  a 64-bit timeout with no validation anywhere downstream — 2^32 + 5000 becomes a real 5s timeout,
  2^32 - 1 becomes -1 (cache forever). The unchecked `ToStringMode` casts and the
  `Ice\register` expires float→int conversion are in the same family. I can open a follow-up issue;
  a checked `getInt32` helper per binding would retire the (now two) hand-rolled copies.

bernardnormier:
Reworded both messages to the name-free form in ff7c432.

On `getLong`: leaving `getInteger` here to stay consistent with the setBufferSize precedent — if we switch, both call sites should move together, which fits better in the follow-up.

Opened #6066 for the other narrowing sites (`ice_invocationTimeout`, `ice_locatorCacheTimeout`, the `ToStringMode` casts, and the `Ice\register` expires conversion) with the checked `getInt32` helper as the suggested fix.

PR #6064, opened 2026-07-20
Bound the IceGrid Activator fork-child file-descriptor close loop

### LINKED ISSUE #5843: [Audit-Medium] IceGrid Activator: fork-child fd-close loop runs to INT_MAX under unlimited RLIMIT_NOFILE — node-wide activation stall
> **AI-generated finding** — surfaced by Claude (Anthropic), model **Claude Fable 5** (`claude-fable-5`), in a multi-agent correctness audit of IceGrid (`cpp/src/IceGrid` + `cpp/test/IceGrid`): six subsystem reviewers followed by three independent adversarial "refute-it" verifiers. Code is pinned to `origin/main` at [`a53cc349e0`](https://github.com/zeroc-ice/ice/commit/a53cc349e0379aa4f3a4c67160e4452982b6fd80). **Not yet human-triaged** — please verify before acting.

## Summary

In the POSIX fork child, the "close all inherited file descriptors" loop is bounded by `sysconf(_SC_OPEN_MAX)` with an `INT_MAX` fallback. When the soft file-descriptor limit is unlimited, the loop issues ~2.1 billion `close()` calls before `exec`, and because the parent blocks reading the child's error pipe under `Activator::_mutex` for the whole of `activate()`, the entire node's activation/deactivation stalls with it.

https://github.com/zeroc-ice/ice/blob/a53cc349e0379aa4f3a4c67160e4452982b6fd80/cpp/src/IceGrid/Activator.cpp#L799-L811

```cpp
int maxFd = static_cast<int>(sysconf(_SC_OPEN_MAX));
if (maxFd <= 0)
{
    maxFd = numeric_limits<int>::max();
}
for (int fd = 3; fd < maxFd; ++fd) { if (fd != fds[1] && fd != errorFds[1]) close(fd); }
```

## Trigger

A node whose soft `RLIMIT_NOFILE` is unlimited (e.g. systemd `LimitNOFILE=infinity`, common in containers): `sysconf(_SC_OPEN_MAX)` returns `-1` (or, on glibc under `RLIM_INFINITY`, `INT_MAX` directly) → `maxFd = INT_MAX`. Every `activate()` fork then runs the loop up to `INT_MAX`.

## Impact

The child spends a long time issuing ~2 billion `close()` syscalls before `exec`. `activate()` holds `Activator::_mutex` for its whole body (lock acquired near line 342) and the parent blocks reading `errorFds[0]` (near line 871) until the child closes `errorFds[1]` (whose `FD_CLOEXEC` is set only *after* the loop). So the whole activator — every activate / deactivate / getServerPid / kill — stalls: a per-activation hang / node-wide DoS on affected deployments.

## Fix direction

Use `close_range(3, ~0u, 0)` / `closefrom(3)` where available (with the `/proc/self/fd` fallback), or clamp `maxFd` to the soft `getrlimit(RLIMIT_NOFILE)` value and a sane cap instead of `INT_MAX`.

pepone:
Closes #5843

In the POSIX fork child, the IceGrid `Activator` closed inherited descriptors by looping from `3` up to `sysconf(_SC_OPEN_MAX)`, with an `INT_MAX` fallback. When the soft `RLIMIT_NOFILE` is very large — e.g. a unit with systemd `LimitNOFILE=infinity`, common in containers (docker/containerd have shipped this in their service files for years) — the loop bound becomes that limit. On Linux the kernel rejects any `RLIMIT_NOFILE` above `fs.nr_open`, so the soft limit is never literally `RLIM_INFINITY`; but systemd v240+ raises `fs.nr_open` to 1,073,741,816 at boot, so `LimitNOFILE=infinity` yields an effective soft limit of ~2^30 and the child issued on the order of a billion `close()` calls before `exec` — minutes per activation. Because `activate()` holds the `Activator` mutex for its whole body and the parent blocks reading the child's error pipe until the child sets `FD_CLOEXEC` (only *after* this loop), every activate/deactivate on the node stalled with it — a per-activation hang / node-wide stall on affected deployments.

The close loop runs *after* `setuid`/`setgid`, so an inherited descriptor left open in the child would also cross a privilege boundary.

## Fix

Extracted a file-local `closeInheritedDescriptors()` helper that keeps the two preserved descriptors open — the interrupt pipe write end `fds[1]` (kept open across `exec` for server-termination detection) and the error pipe write end `errorFds[1]` (closed on `exec` via `FD_CLOEXEC`) — and closes everything else `>= 3` using the best available mechanism:

1. **`close_range(2)`** where available (`SYS_close_range` — Linux 5.9+, FreeBSD 12.2+): closes the whole range in a single syscall regardless of a huge or unlimited limit, keeping the two descriptors open by closing the sub-ranges around them (each clamped to `>= 3`). On any failure (e.g. `ENOSYS` on an older kernel, or a seccomp `EPERM`) it closes nothing and falls through to the next tier. Every supported platform that can have a huge limit has `close_range`: the minimum supported Linux distributions (RHEL 9, Debian 12, Ubuntu 24.04, Amazon Linux 2023) all ship kernel 5.10+.
2. **Bounded loop** (fallback, when `close_range` is unavailable — a pre-5.9 Linux kernel outside the supported set, macOS, or a seccomp profile that rejects the syscall): close `3..maxFd`, where `maxFd` is the soft `RLIMIT_NOFILE` (the kernel never allocates a descriptor at or above it) capped at `1 << 20`. On these platforms the limit is modest, so the loop stays cheap.

`getrlimit(RLIMIT_NOFILE)` for the fallback's bound is queried in the **parent before `fork()`** and passed into the child, so the child helper does not call it (POSIX does not require `getrlimit` to be async-signal-safe). The child helper uses only async-signal-safe calls.

An earlier revision added a `/proc/self/fd` enumeration tier between the two above (the CPython/libuv three-tier shape). It was dropped in review: it only ever executed on platforms outside the support matrix (RHEL 8, Ubuntu 20.04, whose kernels predate `close_range`), and it was the most delicate part of the helper (`opendir`/`readdir` allocating in the fork child, close-while-reading over procfs). On every supported platform `close_range` handles the whole job, and the bounded loop covers macOS and the rare seccomp case.

## Notes

- Added a `changelog.d/service-icegrid/5843.md` fragment.
- No test: reproducing the stall needs an IceGrid node with a very large `RLIMIT_NOFILE` plus activation timing — flaky and disproportionate for a fork-child fd-hygiene fix (matches the sibling Activator fix #5911, which added none).
- Compiled with the project's build flags (`-std=c++20 -Wall -Wextra -Wconversion`); both the `close_range` and bounded-loop paths build clean. Reviewed with an independent Codex cross-check.
- 3.8.3-milestoned; to be cherry-picked to `3.8`, where the `CHANGELOG-3.8.md` entry is added.
- A related but less severe occurrence of the same pattern in `cpp/src/Ice/Service.cpp` (daemonize path) is filed as #6065.
- Follow-up for `main`: nothing in the C++ tree sets `SOCK_CLOEXEC`/`O_CLOEXEC` on the descriptors it opens, which is what makes this fork-child close-loop load-bearing for security rather than just hygiene. Filed as 6072.

### DIFF
--- a/changelog.d/service-icegrid/5843.md
+++ b/changelog.d/service-icegrid/5843.md
@@ -1,1 +1,2 @@
-- Fixed a node-wide activation stall in the IceGrid node when running with a very large file descriptor limit (such as `LimitNOFILE=infinity` under systemd).
+- Fixed a node-wide activation stall in the IceGrid node when running with a very large file descriptor limit
+  (such as `LimitNOFILE=infinity` under systemd).
--- a/changelog.d/service-icegrid/5843.md
+++ b/changelog.d/service-icegrid/5843.md
@@ -1,1 +1,1 @@
-- Fixed a node-wide activation stall in the IceGrid node when running with an unlimited file-descriptor limit.
+- Fixed a node-wide activation stall in the IceGrid node when running with a very large file descriptor limit (such as `LimitNOFILE=infinity` under systemd).
--- a/changelog.d/service-icegrid/5843.md
+++ b/changelog.d/service-icegrid/5843.md
@@ -1,0 +1,1 @@
+- Fixed a node-wide activation stall in the IceGrid node when running with an unlimited file-descriptor limit.
--- a/cpp/src/IceGrid/Activator.cpp
+++ b/cpp/src/IceGrid/Activator.cpp
@@ -90,12 +90,9 @@ namespace IceGrid
     }
 
     //
-    // Close every inherited file descriptor >= 3 in the fork child, keeping the
-    // two given descriptors open (the write ends of the interrupt and error
-    // pipes). This runs after setuid/setgid, so a leaked descriptor would cross
-    // a privilege boundary.
-    //
-    // Runs in the fork child, using only async-signal-safe calls.
+    // Close every inherited file descriptor >= 3, except keep1 and keep2. Runs in the fork child, using only
+    // async-signal-safe calls. maxFd bounds the fallback close loop used when close_range is unavailable. This runs
+    // after setuid/setgid, so a leaked descriptor would cross a privilege boundary.
     //
     void closeInheritedDescriptors(int keep1, int keep2, long maxFd)
     {
@@ -114,9 +111,11 @@ namespace IceGrid
         // never inverted, so the only realistic failure is the syscall being
         // unavailable (ENOSYS on an older kernel, or a seccomp EPERM), and on
         // failure it closes nothing; we fall through to the bounded loop below,
-        // so there is no need to inspect errno.
+        // so there is no need to inspect errno. Clamp the probe's start to 3 so
+        // that if both keeps landed below 3 it still cannot close stderr.
         //
-        if (syscall(SYS_close_range, static_cast<unsigned int>(hi) + 1, ~0u, 0u) == 0)
+        const unsigned int probeStart = static_cast<unsigned int>(hi + 1 > 3 ? hi + 1 : 3);
+        if (syscall(SYS_close_range, probeStart, ~0u, 0u) == 0)
         {
             if (lo > 3)
             {
@@ -777,17 +776,14 @@ Activator::activate(
     const char* pwdCStr = pwd.c_str();
 
     //
-    // Compute the last-resort descriptor-close bound before fork(), so the child
-    // has a finite bound to fall back on if it can use neither close_range nor
-    // /proc/self/fd (and so it need not call getrlimit() itself). Prefer the soft
-    // RLIMIT_NOFILE — the kernel never allocates a descriptor at or above it —
-    // capped so the loop stays bounded even under a huge or unlimited limit.
+    // Compute the fallback descriptor-close bound before fork(), so the child has a finite bound if it cannot use
+    // close_range (and doesn't need to call getrlimit() itself). Prefer the soft RLIMIT_NOFILE — the kernel never
+    // allocates a descriptor at or above it — capped so the loop stays bounded even under a huge or unlimited limit.
     //
     long maxFdToClose = 1 << 20;
     {
         rlimit fdLimit{};
-        if (getrlimit(RLIMIT_NOFILE, &fdLimit) == 0 && fdLimit.rlim_cur != RLIM_INFINITY &&
-            static_cast<long>(fdLimit.rlim_cur) < maxFdToClose)
+        if (getrlimit(RLIMIT_NOFILE, &fdLimit) == 0 && fdLimit.rlim_cur < static_cast<rlim_t>(maxFdToClose))
         {
             maxFdToClose = static_cast<long>(fdLimit.rlim_cur);
         }
--- a/cpp/src/IceGrid/Activator.cpp
+++ b/cpp/src/IceGrid/Activator.cpp
@@ -21,6 +21,7 @@
 
 #ifndef _WIN32
 #    include <csignal>
+#    include <dirent.h>
 #    include <pwd.h> // for getpwuid
 #    include <sys/resource.h>
 #    include <sys/syscall.h>
@@ -93,9 +94,13 @@ namespace IceGrid
     // Close every inherited file descriptor >= 3 in the fork child, keeping the
     // two given descriptors open (the write ends of the interrupt and error
     // pipes). This runs after setuid/setgid, so a leaked descriptor would cross
-    // a privilege boundary. It runs in the fork child, so it uses only
-    // async-signal-safe facilities: close_range/close and the maxFd bound that
-    // the caller computed before fork().
+    // a privilege boundary.
+    //
+    // Runs in the fork child. The close_range fast path and the final bounded
+    // loop use only async-signal-safe calls; the /proc/self/fd enumeration uses
+    // opendir/readdir (which allocate), but that path runs only on Linux, where
+    // glibc keeps the allocator usable in the child after fork — as the
+    // surrounding child code, which also allocates, already relies on.
     //
     void closeInheritedDescriptors(int keep1, int keep2, long maxFd)
     {
@@ -127,9 +132,31 @@ namespace IceGrid
 #    endif
 
         //
-        // Fallback for platforms without close_range (or an older kernel that
-        // returns ENOSYS): close descriptors one by one up to maxFd, the bound
-        // the caller computed from the soft RLIMIT_NOFILE before fork().
+        // Fallback for kernels/platforms without close_range (e.g. Linux before
+        // 5.9): enumerate the descriptors that are actually open from
+        // /proc/self/fd and close them, so the child neither iterates up to a
+        // huge or unlimited limit nor leaks a descriptor above one.
+        //
+        if (DIR* dir = opendir("/proc/self/fd"))
+        {
+            const int dirFd = dirfd(dir);
+            for (dirent* entry = readdir(dir); entry != nullptr; entry = readdir(dir))
+            {
+                const int fd = atoi(entry->d_name); // "." and ".." parse to 0 and are skipped below
+                if (fd >= 3 && fd != lo && fd != hi && fd != dirFd)
+                {
+                    close(fd);
+                }
+            }
+            closedir(dir);
+            return;
+        }
+
+        //
+        // Last resort (no /proc, e.g. some minimal containers or non-Linux
+        // platforms without close_range): close descriptors one by one up to
+        // maxFd, the bound the caller computed from the soft RLIMIT_NOFILE
+        // before fork().
         //
         for (long fd = 3; fd < maxFd; ++fd)
         {
@@ -764,16 +791,17 @@ Activator::activate(
     const char* pwdCStr = pwd.c_str();
 
     //
-    // Compute the bound for closing inherited descriptors before fork(), so the
-    // child (which must stay async-signal-safe) does not call getrlimit(). A
-    // finite soft RLIMIT_NOFILE is an exact bound — the kernel never allocates a
-    // descriptor at or above it — so the child closes the whole range; if it is
-    // unlimited we fall back to a sane cap so the child cannot loop up to INT_MAX.
+    // Compute the last-resort descriptor-close bound before fork(), so the child
+    // has a finite bound to fall back on if it can use neither close_range nor
+    // /proc/self/fd (and so it need not call getrlimit() itself). Prefer the soft
+    // RLIMIT_NOFILE — the kernel never allocates a descriptor at or above it —
+    // capped so the loop stays bounded even under a huge or unlimited limit.
     //
-    long maxFdToClose = 65536;
+    long maxFdToClose = 1 << 20;
     {
         rlimit fdLimit{};
-        if (getrlimit(RLIMIT_NOFILE, &fdLimit) == 0 && fdLimit.rlim_cur != RLIM_INFINITY)
+        if (getrlimit(RLIMIT_NOFILE, &fdLimit) == 0 && fdLimit.rlim_cur != RLIM_INFINITY &&
+            static_cast<long>(fdLimit.rlim_cur) < maxFdToClose)
         {
             maxFdToClose = static_cast<long>(fdLimit.rlim_cur);
         }
--- a/cpp/src/IceGrid/Activator.cpp
+++ b/cpp/src/IceGrid/Activator.cpp
@@ -22,6 +22,8 @@
 #ifndef _WIN32
 #    include <csignal>
 #    include <pwd.h> // for getpwuid
+#    include <sys/resource.h>
+#    include <sys/syscall.h>
 #    include <sys/wait.h>
 #    include <unistd.h>
 #else
@@ -85,6 +87,64 @@ namespace IceGrid
         // process.
         //
         _exit(EXIT_FAILURE);
+    }
+
+    //
+    // Close every inherited file descriptor >= 3 in the fork child, keeping the
+    // two given descriptors open (the write ends of the interrupt and error
+    // pipes). This runs after setuid/setgid, so a leaked descriptor would cross
+    // a privilege boundary; it must use only async-signal-safe facilities.
+    //
+    void closeInheritedDescriptors(int keep1, int keep2)
+    {
+        const int lo = keep1 < keep2 ? keep1 : keep2;
+        const int hi = keep1 < keep2 ? keep2 : keep1;
+
+#    ifdef SYS_close_range
+        //
+        // close_range (Linux 5.9+, FreeBSD 12.2+) closes a whole descriptor
+        // range in a single syscall, so the child never iterates up to a huge
+        // or unlimited RLIMIT_NOFILE. We keep lo and hi open by closing the
+        // ranges around them. Probe with the always-non-empty top range first;
+        // if the syscall exists close the lower ranges too and we are done,
+        // otherwise (e.g. an older kernel returns ENOSYS) fall through to the
+        // descriptor loop below.
+        //
+        if (syscall(SYS_close_range, static_cast<unsigned int>(hi) + 1, ~0u, 0u) == 0)
+        {
+            if (lo > 3)
+            {
+                syscall(SYS_close_range, 3u, static_cast<unsigned int>(lo) - 1, 0u);
+            }
+            if (hi > lo + 1)
+            {
+                syscall(SYS_close_range, static_cast<unsigned int>(lo) + 1, static_cast<unsigned int>(hi) - 1, 0u);
+            }
+            return;
+        }
+#    endif
+
+        //
+        // Fallback for platforms without close_range: close descriptors one by
+        // one up to the soft RLIMIT_NOFILE. The kernel does not allocate
+        // descriptors at or above that limit, so it is a suitable bound; when it
+        // is unlimited (or unavailable) we cap the loop so a single activation
+        // cannot issue billions of close() calls.
+        //
+        rlimit fdLimit{};
+        long maxFd = 65536;
+        if (getrlimit(RLIMIT_NOFILE, &fdLimit) == 0 && fdLimit.rlim_cur != RLIM_INFINITY &&
+            static_cast<long>(fdLimit.rlim_cur) < maxFd)
+        {
+            maxFd = static_cast<long>(fdLimit.rlim_cur);
+        }
+        for (long fd = 3; fd < maxFd; ++fd)
+        {
+            if (fd != lo && fd != hi)
+            {
+                close(static_cast<int>(fd));
+            }
+        }
     }
 
 #endif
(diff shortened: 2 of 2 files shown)


cpp/src/IceGrid/Activator.cpp
```diff
@@ -87,6 +89,64 @@ namespace IceGrid
...
+        // descriptors at or above that limit, so it is a suitable bound; when it
+        // is unlimited (or unavailable) we cap the loop so a single activation
+        // cannot issue billions of close() calls.
+        //
+        rlimit fdLimit{};
+        long maxFd = 65536;
+        if (getrlimit(RLIMIT_NOFILE, &fdLimit) == 0 && fdLimit.rlim_cur != RLIM_INFINITY &&
+            static_cast<long>(fdLimit.rlim_cur) < maxFd)
+        {
+            maxFd = static_cast<long>(fdLimit.rlim_cur);
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Fixed in e4527d2749. The fallback now closes the full soft `RLIMIT_NOFILE` when finite — the 65536 cap applies only when the limit is unlimited (no other bound). The primary `close_range` path closes the whole range regardless of fd numbering, so the above-limit / subsequently-lowered-limit case is covered there. Only the non-`close_range` fallback keeps that gap, and there is no async-signal-safe way to enumerate open fds on those platforms (`/proc/self/fd` would malloc post-fork).

cpp/src/IceGrid/Activator.cpp
```diff
@@ -87,6 +89,64 @@ namespace IceGrid
...
+        // Fallback for platforms without close_range: close descriptors one by
+        // one up to the soft RLIMIT_NOFILE. The kernel does not allocate
+        // descriptors at or above that limit, so it is a suitable bound; when it
+        // is unlimited (or unavailable) we cap the loop so a single activation
+        // cannot issue billions of close() calls.
+        //
+        rlimit fdLimit{};
+        long maxFd = 65536;
+        if (getrlimit(RLIMIT_NOFILE, &fdLimit) == 0 && fdLimit.rlim_cur != RLIM_INFINITY &&
+            static_cast<long>(fdLimit.rlim_cur) < maxFd)
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Fixed in e4527d2749. `getrlimit()` now runs in the parent before `fork()`, and the computed bound is passed into the child — so the child helper uses only `close_range`/`close`.

cpp/src/IceGrid/Activator.cpp
```diff
@@ -792,23 +852,12 @@ Activator::activate(
...
-        }
-
-        for (int fd = 3; fd < maxFd; ++fd)
-        {
-            if (fd != fds[1] && fd != errorFds[1])
-            {
-                close(fd);
-            }
-        }
+        closeInheritedDescriptors(fds[1], errorFds[1]);
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Added `changelog.d/service-icegrid/5843.md` in e4527d2749.

pepone:
🤖 **Claude:** Addressed the suppressed low-confidence note on `Activator.cpp:778` (finite `rlim_cur` copied verbatim → billions of `close()` calls on a no-`close_range` platform) in 858094e535.

The fallback no longer relies on that numeric bound. When `close_range` is unavailable, the child now enumerates the actually-open descriptors from `/proc/self/fd` and closes them — bounded by the real open-descriptor count, with no leak above any numeric bound. The plain loop (now capped at `1<<20`) is only a last resort when `/proc` is unavailable (e.g. a `/proc`-less container or a non-Linux platform without `close_range`, where the soft limit is modest). This is the standard `close_range` → `/proc/self/fd` → bounded-loop tiering used by CPython/libuv.

externl (approved):
Verified the diagnosis and the tiering — `sysconf(_SC_OPEN_MAX)` does return `INT_MAX` under an unlimited `RLIMIT_NOFILE`, and `_mutex` is held for all of `activate()`, so the stall is node-wide as described.

Two before merge:
- `close_range` sub-ranges aren't clamped to 3 — with `lo == 1, hi == 5` the middle call closes stderr.
- The `/proc/self/fd` loop closes while still reading the stream; procfs can skip an entry, and after `setuid` that leaks across the privilege boundary. Collect, then close.

Separately worth an issue against `main`: nothing in the C++ tree sets `SOCK_CLOEXEC`, which is what makes this helper load-bearing for security rather than just hygiene.

cpp/src/IceGrid/Activator.cpp
```diff
@@ -87,6 +90,86 @@ namespace IceGrid
...
+        //
+        if (syscall(SYS_close_range, static_cast<unsigned int>(hi) + 1, ~0u, 0u) == 0)
+        {
+            if (lo > 3)
+            {
+                syscall(SYS_close_range, 3u, static_cast<unsigned int>(lo) - 1, 0u);
+            }
+            if (hi > lo + 1)
+            {
+                syscall(SYS_close_range, static_cast<unsigned int>(lo) + 1, static_cast<unsigned int>(hi) - 1, 0u);
```
externl: Expanding the first bullet from my review.

The middle range starts at `lo + 1` and is never clamped. Its guard, `hi > lo + 1`, only checks the range is non-empty — it says nothing about where the range begins. The bottom range doesn't have this problem because it starts at a literal `3`.

So when `lo == 1, hi == 5`:

- top → `close_range(6, ~0)` ✓
- bottom → `lo > 3` is false, skipped ✓
- middle → `hi > lo + 1` is true → `close_range(2, 4)` — **closes stderr**

The other two tiers can't hit this; both filter per descriptor with `fd >= 3`.

Reachability: `lo` and `hi` are the pipe write ends, and `pipe()` hands out the lowest free numbers, so they land at 3+ only when 0/1/2 are all already open.

- Daemonized — `Service.cpp:1660` reopens 0/1/2 to `/dev/null`. Safe.
- Foreground from a normal shell — 0/1/2 open. Safe.
- Foreground with a standard descriptor closed, e.g. `icegridnode 0<&-`, or launched from a parent that closed one. `pipe()` returns 0 or 1, and `lo < 3`.

Narrow, but the failure is quiet rather than loud: the child closes fd 2 and then execs, so the server starts with stderr closed. The next descriptor it opens becomes fd 2, and anything writing to stderr after that writes into that file instead.

Clamping the start is enough:

```cpp
const int mid = lo + 1 > 3 ? lo + 1 : 3;
if (hi - 1 >= mid)
{
    syscall(SYS_close_range, static_cast<unsigned int>(mid), static_cast<unsigned int>(hi) - 1, 0u);
}
```

changelog.d/service-icegrid/5843.md
```diff
@@ -0,0 +1 @@
+- Fixed a node-wide activation stall in the IceGrid node when running with an unlimited file-descriptor limit.
```
bernardnormier: On Linux the soft limit can never actually be unlimited (the kernel rejects any `RLIMIT_NOFILE` above `fs.nr_open`), so "unlimited" describes the configuration (`LimitNOFILE=infinity`) rather than the effective limit:

```suggestion
- Fixed a node-wide activation stall in the IceGrid node when running with a very large file descriptor limit (such as `LimitNOFILE=infinity` under systemd).
```

bernardnormier (commented):
I agree with the fix and the 3.8.3 targeting. Verified the structural claims against the source: `activate()` holds `_mutex` across the blocking read of the error pipe, and the child sets `FD_CLOEXEC` on `errorFds[1]` only after the close loop, so a slow loop really does stall the whole activator. The `close_range` sub-range arithmetic around the two kept descriptors (including the adjacent/boundary cases), the probe-first-then-commit ordering, and moving `getrlimit` to the parent before `fork()` all look correct.

Two notes:

**The "unlimited"/INT_MAX framing is slightly overstated.** On Linux the kernel rejects any `RLIMIT_NOFILE` above `fs.nr_open`, so the soft limit can never be `RLIM_INFINITY` and the `sysconf(_SC_OPEN_MAX)` → INT_MAX path is unreachable in practice. What actually happens: systemd v240+ raises `fs.nr_open` to 1,073,741,816 at boot, so a unit with `LimitNOFILE=infinity` gets a soft limit of about 2^30 — roughly a billion `close()` calls, i.e. minutes per activation, not the ~2 billion in the description. The impact conclusion stands (this is the same configuration docker/containerd shipped in their service files for years); suggest adjusting the "unlimited" / "~2 billion" wording in the PR description, plus the inline suggestion on the changelog fragment.

**Optional simplification: consider dropping the `/proc/self/fd` tier.** It is the bulk of the new code and its most delicate part (`opendir`/`readdir` allocating in the fork child, the `dirFd` self-exclusion). What it buys only matters on pre-5.9 kernels without `close_range` (RHEL 8, Ubuntu 20.04): there the clamped loop closes at most 2^20 descriptors — around 0.1–0.3 s per activation, acceptable for a fork/exec path — and a descriptor above the cap would only exist if the node itself held over a million open descriptors. Keeping the CPython/libuv three-tier shape as-is is fine too; just flagging that two tiers already fix the reported stall.

bernardnormier:
Following up on the `/proc/self/fd` note above: per the [supported platforms for 3.8.2](https://docs.zeroc.com/ice/3.8/cpp/supported-platforms-for-ice-3-8-2), the minimum supported Linux distributions are RHEL 9, Debian 12, Ubuntu 24.04, and Amazon Linux 2023 — all on kernel 5.10 or later, so every supported Linux has `close_range`. The `/proc/self/fd` tier therefore only ever executes on platforms outside the support matrix (RHEL 8, Ubuntu 20.04), and I'd drop it rather than keep it as an option: on supported platforms `close_range` handles the whole job, and the clamped loop remains as the safety net for macOS and for the rare seccomp profile that rejects `close_range`.

cpp/src/IceGrid/Activator.cpp
pepone: 🤖 **Claude:** Fixed in d626d2dea8. The middle range now starts at `mid = lo + 1 > 3 ? lo + 1 : 3` with guard `hi - 1 >= mid`, so the `lo == 1, hi == 5` case issues `close_range(3, 4)` and no longer touches stderr. Verified across the boundary cases (`lo ≤ 2` → `mid = 3` keeps fd 2; `lo == 3, hi == 5` still closes fd 4; adjacent `lo, hi` skips the middle).

changelog.d/service-icegrid/5843.md
pepone: 🤖 **Claude:** Applied in d626d2dea8 — the fragment now reads "very large file descriptor limit (such as `LimitNOFILE=infinity` under systemd)".

pepone:
🤖 **Claude:** Addressed the remaining review points in d626d2dea8.

- **`close_range` middle range not clamped to 3 (@externl).** Fixed — the start is now clamped to 3 so a pipe write end at fd 0/1/2 can't cause the child to close stderr. Replied in-thread with the boundary-case verification.
- **Drop the `/proc/self/fd` tier (@bernardnormier, @externl).** Dropped. Per the [3.8.2 supported platforms](https://docs.zeroc.com/ice/3.8/cpp/supported-platforms-for-ice-3-8-2), the minimum Linux distributions (RHEL 9, Debian 12, Ubuntu 24.04, Amazon Linux 2023) are all kernel 5.10+, so `close_range` always exists where a huge limit is possible. The tier only ran on out-of-support platforms and was the most delicate part of the helper (`readdir` allocating in the fork child; the close-while-reading procfs hazard @externl flagged) — so removing it also resolves that second point. The helper is now two tiers: `close_range` → bounded loop (the loop remains the safety net for macOS and a seccomp `EPERM`).
- **"unlimited" / "~2 billion" framing (@bernardnormier).** Reworded the changelog fragment ("very large" + the `LimitNOFILE=infinity` example) and the PR description (systemd raises `fs.nr_open` to ~2^30, so it's ~a billion `close()` calls / minutes per activation, and the soft limit is never literally `RLIM_INFINITY`).
- **`SOCK_CLOEXEC` follow-up (@externl).** Filed as #6072 against `main`.

Both the `close_range` and bounded-loop paths build clean under `-std=c++20 -Wall -Wextra -Wconversion` (macOS build via `make`; the Linux `close_range` branch compile-tested separately since it's `#ifdef`'d out on macOS).

changelog.d/service-icegrid/5843.md
```diff
@@ -0,0 +1 @@
+- Fixed a node-wide activation stall in the IceGrid node when running with a very large file descriptor limit (such as `LimitNOFILE=infinity` under systemd).
```
bernardnormier: Nit: use 120 columns

cpp/src/IceGrid/Activator.cpp
```diff
@@ -87,6 +89,70 @@ namespace IceGrid
         _exit(EXIT_FAILURE);
     }
 
+    //
+    // Close every inherited file descriptor >= 3 in the fork child, keeping the
```
bernardnormier: Nit: use 120-columns for new comments.

cpp/src/IceGrid/Activator.cpp
```diff
@@ -87,6 +89,70 @@ namespace IceGrid
...
+    // two given descriptors open (the write ends of the interrupt and error
+    // pipes). This runs after setuid/setgid, so a leaked descriptor would cross
+    // a privilege boundary.
+    //
+    // Runs in the fork child, using only async-signal-safe calls.
```
bernardnormier: The parenthesis re-documents the call site (which already explains what the two write ends are for) and "in the fork child" appears twice. Suggest stating just the helper's own contract, covering all three parameters:

```suggestion
    // Close every inherited file descriptor >= 3, except keep1 and keep2.
    // Runs in the fork child, using only async-signal-safe calls. maxFd bounds
    // the fallback close loop used when close_range is unavailable.
```

cpp/src/IceGrid/Activator.cpp
```diff
@@ -710,6 +776,23 @@ Activator::activate(
     //
     const char* pwdCStr = pwd.c_str();
 
+    //
+    // Compute the last-resort descriptor-close bound before fork(), so the child
+    // has a finite bound to fall back on if it can use neither close_range nor
+    // /proc/self/fd (and so it need not call getrlimit() itself). Prefer the soft
+    // RLIMIT_NOFILE — the kernel never allocates a descriptor at or above it —
+    // capped so the loop stays bounded even under a huge or unlimited limit.
```
bernardnormier: The `/proc/self/fd` tier was dropped in review, so this comment references a mechanism that no longer exists. Also reworded the somewhat formal "it need not call" while at it.

```suggestion
    // Compute the fallback descriptor-close bound before fork(), so the child has
    // a finite bound if it cannot use close_range (and doesn't need to call
    // getrlimit() itself). Prefer the soft RLIMIT_NOFILE — the kernel never
    // allocates a descriptor at or above it — capped so the loop stays bounded
    // even under a huge or unlimited limit.
```

cpp/src/IceGrid/Activator.cpp
```diff
@@ -710,6 +776,23 @@ Activator::activate(
...
+    //
+    long maxFdToClose = 1 << 20;
+    {
+        rlimit fdLimit{};
+        if (getrlimit(RLIMIT_NOFILE, &fdLimit) == 0 && fdLimit.rlim_cur != RLIM_INFINITY &&
+            static_cast<long>(fdLimit.rlim_cur) < maxFdToClose)
```
bernardnormier: Optional nit: comparing in the `rlim_t` domain avoids narrowing to `long` before the comparison (on a platform with a 32-bit `long`, a soft limit of exactly 2^32 would wrap to 0 and silently disable the fallback), and it makes the explicit `RLIM_INFINITY` check redundant — infinity is the maximum `rlim_t` value, so it can never be `< 2^20`. Theoretical either way, since every supported platform has `close_range`.

```suggestion
        if (getrlimit(RLIMIT_NOFILE, &fdLimit) == 0 && fdLimit.rlim_cur < static_cast<rlim_t>(maxFdToClose))
```

bernardnormier (approved):
The `close_range` interval arithmetic checks out: the three sub-ranges exactly cover "everything >= 3 except the two keeps", the `lo > 3` / `hi - 1 >= mid` guards prevent inverted ranges, and the probe-first ordering means a failed probe closes nothing before the fallback loop runs. Querying `getrlimit` in the parent is the right call.

Two small suggestions inline. One more observation, fine to leave as is: the probe's start (`hi + 1`) isn't clamped to 3 the way `mid` is, so if *both* pipe write ends landed below 3 the probe would close stderr. That requires two standard descriptors closed before the pipes were created — strictly more contrived than the `lo < 3` case the `mid` clamp defends against — so take it or leave it.

changelog.d/service-icegrid/5843.md
pepone: 🤖 **Claude:** Wrapped in 9b6e2c3311 — the fragment now breaks at ≤120 columns, with the continuation line indented two spaces to match the sibling fragments.

cpp/src/IceGrid/Activator.cpp
pepone: 🤖 **Claude:** Reflowed the new comment blocks to 120 columns in 9b6e2c3311 (the `closeInheritedDescriptors` header and the pre-`fork()` bound comment) — folded into the two rewords below.

cpp/src/IceGrid/Activator.cpp
pepone: 🤖 **Claude:** Applied in 9b6e2c3311. Tightened to the three-line form covering all three parameters, dropped the duplicate "in the fork child" and the call-site parenthetical. Kept the "runs after setuid/setgid → crosses a privilege boundary" note (still within 120 cols) since it is the rationale for why the close matters.

cpp/src/IceGrid/Activator.cpp
pepone: 🤖 **Claude:** Fixed in 9b6e2c3311. The comment no longer references `/proc/self/fd` (the dropped tier); it now reads "a finite bound if it cannot use close_range … so the child does not need to call getrlimit() itself", per your wording, reflowed to 120.

cpp/src/IceGrid/Activator.cpp
pepone: 🤖 **Claude:** Applied in 9b6e2c3311. Now compares in the `rlim_t` domain — `fdLimit.rlim_cur < static_cast<rlim_t>(maxFdToClose)` — and dropped the explicit `RLIM_INFINITY` check, since infinity is the maximum `rlim_t` value and can never be `< 2^20`.

pepone:
🤖 **Claude:** Addressed the remaining review points in 9b6e2c3311.

- **Stale `/proc/self/fd` comment (@bernardnormier).** The pre-`fork()` bound comment no longer references the dropped tier — reworded per your suggestion, reflowed to 120.
- **`closeInheritedDescriptors` header comment (@bernardnormier).** Tightened to cover all three parameters; removed the duplicate "in the fork child" and the call-site parenthetical, kept the setuid/setgid privilege-boundary rationale.
- **`rlim_t`-domain comparison (@bernardnormier).** Compares `fdLimit.rlim_cur < static_cast<rlim_t>(maxFdToClose)`; dropped the now-redundant `RLIM_INFINITY` check.
- **Probe start not clamped to 3 (@bernardnormier, from your approval).** Took it — the probe now starts at `max(hi + 1, 3)`, so the contrived "both pipe write ends below 3" case can no longer close stderr.
- **120-column nits (@bernardnormier).** Changelog fragment wrapped, new comment blocks reflowed to 120.

Both the `close_range` and bounded-loop paths still build clean under `-std=c++20 -Wall -Wextra -Wconversion` (macOS native path via `make`; the `SYS_close_range` branch compile-checked separately since it is `#ifdef`'d out on macOS), and `clang-format-19` is clean.

cpp/src/IceGrid/Activator.cpp
```diff
@@ -87,6 +89,69 @@ namespace IceGrid
...
+            // write ends land at 0/1/2 only when a standard descriptor was
+            // already closed before exec), and an unclamped lo + 1 == 2 would
+            // close stderr.
+            //
+            const int mid = lo + 1 > 3 ? lo + 1 : 3;
+            if (hi - 1 >= mid)
+            {
+                syscall(SYS_close_range, static_cast<unsigned int>(mid), static_cast<unsigned int>(hi) - 1, 0u);
+            }
+            return;
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Declining. `close_range` here is a best-effort bulk close — the fallback loop is for when the mechanism is *unavailable* (probe fails: ENOSYS on an old kernel, or seccomp rejecting `close_range` outright), not to retry individual sub-ranges. A seccomp policy that permits the unbounded probe yet arg-rejects a bounded sub-range isn't a real configuration; anything that denies `close_range` denies the probe too, and we already fall through. Not worth a second-mechanism retry for that corner.

PR #6067, opened 2026-07-20
Negotiate a DataStorm protocol epoch for correct reader initialization

### LINKED ISSUE #5819: [Audit-High] DataStorm: per-key init-sample batches silently dropped (flow 1) or applied to the wrong reader (flow 2)

pepone:
Fixes #5819.

Supersedes #5880, which addressed the same bug by making the writer re-address its initialization batches through a
second `attachElementsAck`. That approach is not safe across 3.8 patch versions (the reasoning is in the
[icerpc-dev discussion](https://github.com/icerpc/icerpc-dev/discussions/5)). This PR takes the agreed alternative: a
receiver-side fix that stays wire-compatible with Ice 3.8.0-3.8.2, plus an opt-in protocol epoch that upgraded nodes
negotiate for an exact fix.

## The bug

A writer covering several keys, or several readers of one writer, produces one initialization batch per key and per
reader. The `initSamples` receiver marked *every* attached reader initialized on the first batch and dropped the rest,
so a late-joining multi-key or any-key reader lost all but one key's samples.

## Receiver-side fix (protocol epoch 0, wire-compatible with 3.8.0-3.8.2)

`SessionI::initSamples` now groups an initialization round's batches by writer element, then delivers the merged
samples to each attached reader filtered by the keys it subscribes to (via the reader's own `matchKey`), in global
sample-id order. This repairs the dropped and misrouted batches for **every** peer, old or new, and the sender wire is
byte-identical to 3.8.2. The one case the legacy `DataSamples` envelope cannot resolve — two readers on the same key
with different sample filters — is left to epoch 1.

## Exact fix (protocol epoch 1)

Two nodes running this release negotiate a protocol epoch during the twoway `Node` handshake — an
`optional(1) int protocolEpoch` on the three establishment operations, absent meaning epoch 0 — and initialize each
reader element by exact address with a new `initializeReaders` operation. Each node advertises the highest epoch it
supports and both use the lower of the two, so the negotiation is automatic and requires no configuration.

- **`DataStorm.Node.MaxProtocolEpoch`** (default `1`) caps the epoch a node advertises. Set it to `0` to make a node
  speak the legacy wire.
- **`DataStorm.Node.MinProtocolEpoch`** (default `0`) is the acceptance floor. Set it to `1` to refuse sessions with
  peers that do not support the exact fix; those sessions are rejected with a `NotSupported` dispatch exception, which
  every Ice version understands.

Relays forward the tagged epoch, so two upgraded nodes negotiate epoch 1 through an upgraded relay; an old relay strips
the tag and both endpoints safely fall back to epoch 0.

## Compatibility

Default-configured nodes stay fully interoperable with 3.8.0-3.8.2 (epoch-0 sessions, with the receiver-side fix
applied) and automatically use the exact fix between upgraded nodes. There is no coordinated-upgrade requirement unless
an operator opts into `MinProtocolEpoch=1`.

The session retry policy still treats the `NotSupported` rejection as a transient failure and retries it within the
retry budget; tightening that classification (so a deterministic protocol rejection is terminal) is a separate change,
tracked as a follow-up.

## Testing

The `test/DataStorm/events` suite covers late-joining multi-key, single-key, any-key, filtered, interleaved-order,
unwritten-key, and coalesced same-key readers. It now runs under both the default configuration (epoch 1, exercising
`initializeReaders`) and a `DataStorm.Node.MaxProtocolEpoch=0` configuration (epoch 0, exercising the receiver-side
`initSamples` fix). The full DataStorm suite passes, including the relay tests.

## Known limitation (pre-existing, tracked separately)

When a single node hosts two `Topic` instances with the *same name* both reading one writer, initialization can be
misrouted because DataStorm element and key ids are only unique within a topic instance. This is not specific to this
change — it predates it and affects the legacy path as well — and it is tracked separately in #6068.

### DIFF
--- a/changelog.d/datastorm/5819.md
+++ b/changelog.d/datastorm/5819.md
@@ -1,0 +1,10 @@
+- Fixed a bug in DataStorm where the initialization samples for a late-joining reader could be lost or delivered to
+  the wrong reader. A writer covering several keys, or several readers of the same writer, produces one initialization
+  batch per key and per reader; the receiver marked every attached reader initialized on the first batch and dropped
+  the rest, so a multi-key or any-key late reader lost all but one key's samples. The receiver now routes each writer's
+  batches to the readers that subscribe to their keys. Two DataStorm nodes running this release also negotiate a new
+  protocol epoch during session establishment and initialize each reader element by exact address, which removes the
+  remaining ambiguity when several readers subscribe to the same key with different sample filters. The negotiation is
+  backward compatible: a node still establishes epoch-0 sessions with Ice 3.8.0-3.8.2 peers and applies the
+  receiver-side fix on them. Set `DataStorm.Node.MinProtocolEpoch=1` to require the new epoch and refuse sessions with
+  older peers, and `DataStorm.Node.MaxProtocolEpoch` to cap the epoch a node advertises.
--- a/config/PropertyNames.xml
+++ b/config/PropertyNames.xml
@@ -399,6 +399,8 @@
 
     <section name="DataStorm" opt-in="false">
         <property name="Node.ConnectTo" languages="cpp" />
+        <property name="Node.MaxProtocolEpoch" default="1" languages="cpp" />
+        <property name="Node.MinProtocolEpoch" default="0" languages="cpp" />
         <property name="Node.Multicast" class="ObjectAdapter" languages="cpp" />
         <property name="Node.Multicast.Enabled" default="1" languages="cpp" />
         <property name="Node.Multicast.Proxy" class="Proxy" languages="cpp" />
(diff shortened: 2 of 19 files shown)


pepone:
Superseded by #6138, which takes the agreed approach from https://github.com/icerpc/icerpc-dev/discussions/5: break wire compatibility with Ice 3.8.0–3.8.2 (segregating nodes with a new `DataStorm/Lookup2` object identity) and address each reader element exactly by adding a `peerId` to `DataSamples`. That removes the protocol-epoch negotiation, the `initializeReaders` operation, and the receiver-side mixed-version mitigation added here.

PR #6075, opened 2026-07-21
Report the JS idle-timeout in seconds in the idle-check abort message

### LINKED ISSUE #5948: JS idle check reports the idle timeout in milliseconds with an "s" suffix
When the JS idle check aborts a connection, the `ConnectionAbortedException` message reports the timeout in milliseconds but labels it as seconds. For the default 60s idle timeout:

```
Connection aborted by the idle check because it did not receive any bytes for 60000s.
```

`IdleTimeoutTransceiverDecorator` converts the timeout to milliseconds for its timers ([IdleTimeoutTransceiverDecorator.js:8](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/IdleTimeoutTransceiverDecorator.js#L8)) and then passes that milliseconds value to `idleCheck` ([IdleTimeoutTransceiverDecorator.js:96](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/IdleTimeoutTransceiverDecorator.js#L96)), whose message template appends `s` ([ConnectionI.js:956](https://github.com/zeroc-ice/ice/blob/main/js/packages/ice/src/Ice/ConnectionI.js#L956)).

The other mappings report seconds here (see `idleTimeoutInSeconds` in the C# `ConnectionI.cs`). `idleCheck` should receive the timeout in seconds; only the timer scheduling needs the milliseconds value.

pepone:
Fixes #5948.

When the JS idle check aborts a connection, the `ConnectionAbortedException` message reported the idle timeout in **milliseconds** while labeling it with an `s` suffix. For the default 60s idle timeout:

```
Connection aborted by the idle check because it did not receive any bytes for 60000s.
```

`IdleTimeoutTransceiverDecorator` stored the idle timeout already converted to milliseconds and passed that same value to `Connection.idleCheck`, whose message template appends `s`.

## Fix

This mirrors the C++/C# mappings, where `idleCheck` receives the timeout in seconds and only the timer scheduling uses milliseconds (C++ stores `_idleTimeout` as `chrono::seconds` and converts to milliseconds only for the heartbeat timer):

- Keep a single `_idleTimeoutMs` field (milliseconds, used by both timer `schedule()` calls) and convert to seconds at the one call that needs it — `idleCheck(this._idleTimeoutMs / 1000)`. `ConnectionI.js` is unchanged — its `${idleTimeout}s` template is now correct.
- Strengthen the existing `Ice/idleTimeout` client test to assert the abort message reports the timeout in seconds (`...for 1s.`). The test previously checked only the exception type, so the bug was untested.

No changelog fragment: the incorrect suffix is a minor cosmetic issue not worth a changelog entry.

### DIFF
--- a/js/packages/ice/src/Ice/IdleTimeoutTransceiverDecorator.js
+++ b/js/packages/ice/src/Ice/IdleTimeoutTransceiverDecorator.js
@@ -5,8 +5,7 @@ export class IdleTimeoutTransceiverDecorator {
         console.assert(idleTimeout > 0);
 
         this._decoratee = decoratee;
-        this._idleTimeout = idleTimeout; // in seconds, as reported by the idle check
-        this._idleTimeoutMs = idleTimeout * 1000; // in milliseconds, for the timers
+        this._idleTimeoutMs = idleTimeout * 1000;
         this._timer = timer;
         this._connection = connection;
 
@@ -94,7 +93,7 @@ export class IdleTimeoutTransceiverDecorator {
         if (this._idleCheckEnabled) {
             this.cancelReadTimer();
             this._readTimerToken = this._timer.schedule(() => {
-                this._connection.idleCheck(this._idleTimeout);
+                this._connection.idleCheck(this._idleTimeoutMs / 1000);
             }, this._idleTimeoutMs);
         }
     }
--- a/js/packages/ice/src/Ice/IdleTimeoutTransceiverDecorator.js
+++ b/js/packages/ice/src/Ice/IdleTimeoutTransceiverDecorator.js
@@ -5,7 +5,8 @@ export class IdleTimeoutTransceiverDecorator {
         console.assert(idleTimeout > 0);
 
         this._decoratee = decoratee;
-        this._idleTimeout = idleTimeout * 1000; // Convert seconds to milliseconds
+        this._idleTimeout = idleTimeout; // in seconds, as reported by the idle check
+        this._idleTimeoutMs = idleTimeout * 1000; // in milliseconds, for the timers
         this._timer = timer;
         this._connection = connection;
 
@@ -94,7 +95,7 @@ export class IdleTimeoutTransceiverDecorator {
             this.cancelReadTimer();
             this._readTimerToken = this._timer.schedule(() => {
                 this._connection.idleCheck(this._idleTimeout);
-            }, this._idleTimeout);
+            }, this._idleTimeoutMs);
         }
     }
 
@@ -102,6 +103,6 @@ export class IdleTimeoutTransceiverDecorator {
         this.cancelWriteTimer();
         this._writeTimerToken = this._timer.schedule(() => {
             this._connection.sendHeartbeat();
-        }, this._idleTimeout / 2);
+        }, this._idleTimeoutMs / 2);
     }
 }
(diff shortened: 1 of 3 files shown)


externl (approved):

js/packages/ice/src/Ice/IdleTimeoutTransceiverDecorator.js
```diff
@@ -5,7 +5,8 @@ export class IdleTimeoutTransceiverDecorator {
         console.assert(idleTimeout > 0);
 
         this._decoratee = decoratee;
-        this._idleTimeout = idleTimeout * 1000; // Convert seconds to milliseconds
+        this._idleTimeout = idleTimeout; // in seconds, as reported by the idle check
+        this._idleTimeoutMs = idleTimeout * 1000; // in milliseconds, for the timers
```
bernardnormier: Prefer a single field over two fields holding the same quantity in different units — keep the milliseconds value the timers use, and convert at the one call that needs seconds (see the suggestion on `idleCheck` below).

```suggestion
        this._idleTimeoutMs = idleTimeout * 1000;
```

js/packages/ice/src/Ice/IdleTimeoutTransceiverDecorator.js
```diff
@@ -94,14 +95,14 @@ export class IdleTimeoutTransceiverDecorator {
             this.cancelReadTimer();
             this._readTimerToken = this._timer.schedule(() => {
                 this._connection.idleCheck(this._idleTimeout);
```
bernardnormier:
```suggestion
                this._connection.idleCheck(this._idleTimeoutMs / 1000);
```

js/packages/test/Ice/idleTimeout/Client.ts
```diff
@@ -137,6 +137,8 @@ async function testServerWithEnableDisableIdleCheck(
         } catch (ex) {
             test(ex instanceof Ice.ConnectionAbortedException);
             test(enabled);
+            // The idle check must report the timeout in seconds (1s), not milliseconds (1000s). See issue #5948.
```
bernardnormier: No issue references in tests — each test stands on its own; the first sentence already explains the assertion.

```suggestion
            // The idle check must report the timeout in seconds (1s), not milliseconds (1000s).
```

bernardnormier (commented):
The fix is correct — I checked every use in the decorator (`_idleTimeout` now flows only into `idleCheck`, `_idleTimeoutMs` into both `schedule` calls) and the `${idleTimeout}s` template in `ConnectionI.js`. Strengthening the existing assertion is the right scope, and I agree with skipping the changelog fragment.

One structural preference and one test-comment nit inline: rather than carrying two fields for the same quantity, keep a single `_idleTimeoutMs` field and convert at the one place that needs seconds.

js/packages/ice/src/Ice/IdleTimeoutTransceiverDecorator.js
pepone: 🤖 **Claude:** Done in 2e2268227d — dropped `_idleTimeout`, keeping the single `_idleTimeoutMs` field.

js/packages/ice/src/Ice/IdleTimeoutTransceiverDecorator.js
pepone: 🤖 **Claude:** Done in 2e2268227d — `idleCheck` now converts `_idleTimeoutMs / 1000`.

js/packages/test/Ice/idleTimeout/Client.ts
pepone: 🤖 **Claude:** Done in 2e2268227d — removed the issue reference from the test comment.

bernardnormier (approved):
Looks good. Just update the PR description to match the code.

PR #6076, opened 2026-07-21
Add coverage-gap @throws and align Instrumentation wording in the Java Ice API docs

### LINKED ISSUE #5935: Java API doc follow-ups: coverage-gap @throws and Instrumentation wording (deferred from #5933)
Follow-up doc-quality items for the **com.zeroc.ice** mapping that were deliberately left out of #5933 to keep that
PR aligned with the C++ (#5867) / C# (#5868) scope. Surfaced by the Java API doc audit. All paths under
`java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/`.

### 1. Instrumentation wording (Java-specific)

The `Instrumentation` javadoc still uses the pre-3.8 style; C++ standardized on "runtime" and the "Notifies…" voice.

- **"Ice run-time" → "Ice runtime"** — `Instrumentation/CommunicatorObserver.java` (10 occurrences),
  `Instrumentation/ObserverUpdater.java` (1).
- **"This method is called when…" / "This method should…" → "Notifies…"** style — `Instrumentation/Observer.java`,
  `Instrumentation/CommunicatorObserver.java`.

### 2. Coverage-gap `@throws`

Public methods that throw documented Ice exceptions without a corresponding `@throws` tag:

- **`Communicator.getDefaultLocator`** (`Communicator.java:418`) — missing `@throws CommunicatorDestroyedException`.
  Its siblings `getDefaultRouter` (`:394`), `getDefaultObjectAdapter`, and `getPluginManager` all document it; the
  same `referenceFactory()`/`Instance` throwing path applies.
- **`Communicator.createObjectAdapter` / `createObjectAdapterWithEndpoints`** (`Communicator.java:199-283`) —
  missing `@throws InitializationException` (requires-configuration, invalid `ProxyOptions`, oversized
  `MessageSizeMax`, router-with-endpoints) and `@throws AlreadyRegisteredException` (duplicate adapter name). C#
  documents `InitializationException`.
- **`PluginManager.getPlugin` / `addPlugin`** (`PluginManager.java:33`, `:42`) — undocumented
  `CommunicatorDestroyedException`.
- **`Communicator.createAdmin`** (`Communicator.java:492`) — undocumented `CommunicatorDestroyedException` and
  `IllegalArgumentException` (invalid admin identity).
- **`ObjectPrx.ice_identity`** interface decl (`ObjectPrx.java:213`) — missing `@throws IllegalArgumentException` for
  an empty name (thrown by `_ObjectPrxI`).
- **`Properties.setProperty`** (`Properties.java:341`) — missing `@throws PropertyException` (unknown Ice property /
  disabled opt-in prefix) and `@throws InitializationException` (empty key).
- **`Connection.getInfo` / `setBufferSize`** (`Connection.java:111`, `:119`) — both rethrow the stored close
  exception once the connection is closed; undocumented.
- **`DispatchException`** ctors (`DispatchException.java:19/30/39`) — the `@param replyStatus` doesn't state the
  "must be greater than `ReplyStatus.UserException`" expectation. (Java, unlike C++, doesn't validate it — so this is
  a doc note, and there's a separate question of whether Java should validate to match C++.)

### Cross-mapping note

Several `@throws` gaps above are **also** missing in C++/C# (`setProperty`, `getInfo`/`setBufferSize`,
`createObjectAdapter`'s `InitializationException`) per the audit — worth a coordinated pass if picked up. The
Instrumentation rephrasing is Java-only.

Refs: Java doc audit report `API_DOC_AUDIT_JAVA-2026-07-09.md` (ice-audit), sections C/D/E; PR #5933.

> AI-assisted follow-up (Claude, Anthropic) from the Java API doc audit; not yet human-triaged.

pepone:
Follow-up doc-quality items for the **com.zeroc.ice** mapping deferred from #5933 (to keep that PR aligned with the C++ #5867 / C# #5868 scope). Javadoc-only; each `@throws` was verified against the implementation. Closes #5935.

### 1. Instrumentation wording (Java-specific)

- **`Instrumentation/CommunicatorObserver.java`** — all "Ice run-time" → "Ice runtime" (10 occurrences), and the pre-3.8 "This method should return X…" summaries rephrased into the modern C++ voice from `cpp/include/Ice/Instrumentation.h`: the imperative "Gets an observer for…" for the getters, plus a proper "Sets the observer updater." summary on `setObserverUpdater`.
- **`Instrumentation/Observer.java`** — `attach`/`detach` now use "Notifies the observer that an instrumented object was created/destroyed." (matches C++).
- **`Instrumentation/ObserverUpdater.java`** — the 1 remaining "run-time" → "runtime".

> Note: for `CommunicatorObserver` I used the C++-accurate "Gets…" getter voice rather than "Notifies…" (the issue's shorthand) — these methods return observers, they don't notify. "Notifies…" is applied only to `Observer.attach`/`detach`, where it's correct.

### 2. Coverage-gap `@throws`

| Method | Added | Verified in |
|---|---|---|
| `Communicator.getDefaultLocator` | `CommunicatorDestroyedException` | `referenceFactory()` path; matches sibling `getDefaultRouter` |
| `Communicator.createObjectAdapter` / `…WithEndpoints` (4 overloads) | `InitializationException`, `AlreadyRegisteredException` | `ObjectAdapter.java`, `ObjectAdapterFactory.java:108` |
| `Communicator.createAdmin` | `CommunicatorDestroyedException`, `IllegalArgumentException` | `Instance.createAdmin` |
| `PluginManager.getPlugin` / `addPlugin` | `CommunicatorDestroyedException` | `PluginManagerI.java:67,81` |
| `ObjectPrx.ice_identity` | `IllegalArgumentException` (empty name) | `_ObjectPrxI.ice_identity` |
| `Properties.setProperty` | `InitializationException` (empty key), `PropertyException` (unknown / opt-in prefix) | `Properties.java:348,356,362` |
| `Connection.getInfo` / `setBufferSize` | `LocalException` (rethrown close reason) | `ConnectionI.java:949,957` |
| `DispatchException` (3 ctors) | `@param replyStatus` now states it must be greater than `ReplyStatus.UserException` | matches C++/C# `LocalExceptions` |

### Cross-mapping note

Some of these gaps (`setProperty`, `getInfo`/`setBufferSize`, `createObjectAdapter`'s `InitializationException`) are **also** missing in C++/C# per the audit and could use a coordinated pass. The Instrumentation rephrasing is Java-only.

Refs: Java doc audit report, sections C/D/E; PR #5933.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
@@ -205,7 +205,8 @@ public final class Communicator implements AutoCloseable {
      *
      * @param name the object adapter name
      * @return the new object adapter
-     * @throws InitializationException if the object adapter configuration is invalid
+     * @throws CommunicatorDestroyedException if the communicator has been destroyed
+     * @throws InitializationException if the object adapter's configuration is invalid
      * @throws AlreadyRegisteredException if an object adapter with the same name already exists
      * @see #createObjectAdapterWithEndpoints
      * @see Properties
@@ -230,7 +231,8 @@ public final class Communicator implements AutoCloseable {
      *     calling {@link #createObjectAdapter(String)}.
      * @return the new object adapter
      * @throws IllegalArgumentException if the provided name is empty and sslEngineFactory is non-null
-     * @throws InitializationException if the object adapter configuration is invalid
+     * @throws CommunicatorDestroyedException if the communicator has been destroyed
+     * @throws InitializationException if the object adapter's configuration is invalid
      * @throws AlreadyRegisteredException if an object adapter with the same name already exists
      * @see #createObjectAdapterWithEndpoints
      * @see Properties
@@ -251,7 +253,8 @@ public final class Communicator implements AutoCloseable {
      * @param name the object adapter name
      * @param endpoints the endpoints of the object adapter
      * @return the new object adapter
-     * @throws InitializationException if the object adapter configuration is invalid
+     * @throws CommunicatorDestroyedException if the communicator has been destroyed
+     * @throws InitializationException if the object adapter's configuration is invalid
      * @throws AlreadyRegisteredException if an object adapter with the same name already exists
      * @see #createObjectAdapter
      * @see Properties
@@ -275,7 +278,8 @@ public final class Communicator implements AutoCloseable {
      *     through {@code Ice.SSL} configuration properties. Passing {@code null} is equivalent to
      *     calling {@link #createObjectAdapterWithEndpoints(String, String)}.
      * @return the new object adapter
-     * @throws InitializationException if the object adapter configuration is invalid
+     * @throws CommunicatorDestroyedException if the communicator has been destroyed
+     * @throws InitializationException if the object adapter's configuration is invalid
      * @throws AlreadyRegisteredException if an object adapter with the same name already exists
      * @see #createObjectAdapter
      * @see Properties
@@ -497,7 +501,8 @@ public final class Communicator implements AutoCloseable {
      * @return a proxy to the main ("") facet of the Admin object
      * @throws CommunicatorDestroyedException if the communicator has been destroyed
      * @throws IllegalArgumentException if the admin identity is not valid
-     * @throws InitializationException if this method is called more than once
+     * @throws InitializationException if the Admin object is already created, the Admin facility is
+     *     disabled, or {@code Ice.Admin.Endpoints} is not set
      * @see #getAdmin
      */
     public ObjectPrx createAdmin(ObjectAdapter adminAdapter, Identity adminId) {
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
@@ -205,6 +205,8 @@ public final class Communicator implements AutoCloseable {
      *
      * @param name the object adapter name
      * @return the new object adapter
+     * @throws InitializationException if the object adapter configuration is invalid
+     * @throws AlreadyRegisteredException if an object adapter with the same name already exists
      * @see #createObjectAdapterWithEndpoints
      * @see Properties
      */
@@ -228,6 +230,8 @@ public final class Communicator implements AutoCloseable {
      *     calling {@link #createObjectAdapter(String)}.
      * @return the new object adapter
      * @throws IllegalArgumentException if the provided name is empty and sslEngineFactory is non-null
+     * @throws InitializationException if the object adapter configuration is invalid
+     * @throws AlreadyRegisteredException if an object adapter with the same name already exists
      * @see #createObjectAdapterWithEndpoints
      * @see Properties
      */
@@ -247,6 +251,8 @@ public final class Communicator implements AutoCloseable {
      * @param name the object adapter name
      * @param endpoints the endpoints of the object adapter
      * @return the new object adapter
+     * @throws InitializationException if the object adapter configuration is invalid
+     * @throws AlreadyRegisteredException if an object adapter with the same name already exists
      * @see #createObjectAdapter
      * @see Properties
      */
@@ -269,6 +275,8 @@ public final class Communicator implements AutoCloseable {
      *     through {@code Ice.SSL} configuration properties. Passing {@code null} is equivalent to
      *     calling {@link #createObjectAdapterWithEndpoints(String, String)}.
      * @return the new object adapter
+     * @throws InitializationException if the object adapter configuration is invalid
+     * @throws AlreadyRegisteredException if an object adapter with the same name already exists
      * @see #createObjectAdapter
      * @see Properties
      */
@@ -412,6 +420,7 @@ public final class Communicator implements AutoCloseable {
      * Gets the default locator of this communicator.
      *
      * @return the default locator of this communicator
+     * @throws CommunicatorDestroyedException if the communicator has been destroyed
      * @see #setDefaultLocator
      * @see Locator
      */
@@ -486,6 +495,8 @@ public final class Communicator implements AutoCloseable {
      *     after creating and activating this adapter.
      * @param adminId the identity of the Admin object
      * @return a proxy to the main ("") facet of the Admin object
+     * @throws CommunicatorDestroyedException if the communicator has been destroyed
+     * @throws IllegalArgumentException if the admin identity is not valid
      * @throws InitializationException if this method is called more than once
      * @see #getAdmin
      */
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
@@ -107,6 +107,8 @@ public interface Connection {
      * Returns the connection information.
      *
      * @return the connection information
+     * @throws LocalException if the connection is closed; the exception provides the reason for the
+     *     closure (see {@link #throwException}).
      */
     ConnectionInfo getInfo();
 
@@ -115,6 +117,8 @@ public interface Connection {
      *
      * @param rcvSize the size of the receive buffer
      * @param sndSize the size of the send buffer
+     * @throws LocalException if the connection is closed; the exception provides the reason for the
+     *     closure (see {@link #throwException}).
      */
     void setBufferSize(int rcvSize, int sndSize);
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/DispatchException.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/DispatchException.java
@@ -16,9 +16,15 @@ public class DispatchException extends LocalException {
      *     than {@link ReplyStatus#UserException}
      * @param message the detail message
      * @param cause the cause
+     * @throws IllegalArgumentException if {@code replyStatus} does not fit in a byte or is not
+     *     greater than {@link ReplyStatus#UserException}
      */
     public DispatchException(int replyStatus, String message, Throwable cause) {
         super(createMessage(message, replyStatus), cause);
+        if (replyStatus <= ReplyStatus.UserException.value() || replyStatus > 255) {
+            throw new IllegalArgumentException(
+                "The reply status of a DispatchException must fit in a byte and be greater than ReplyStatus.UserException.");
+        }
         this.replyStatus = replyStatus;
     }
 
@@ -28,6 +34,8 @@ public class DispatchException extends LocalException {
      * @param replyStatus the reply status as an int (see {@link ReplyStatus}); it must be greater
      *     than {@link ReplyStatus#UserException}
      * @param message the detail message
+     * @throws IllegalArgumentException if {@code replyStatus} does not fit in a byte or is not
+     *     greater than {@link ReplyStatus#UserException}
      */
     public DispatchException(int replyStatus, String message) {
         this(replyStatus, message, null);
@@ -38,6 +46,8 @@ public class DispatchException extends LocalException {
      *
      * @param replyStatus the reply status as an int (see {@link ReplyStatus}); it must be greater
      *     than {@link ReplyStatus#UserException}
+     * @throws IllegalArgumentException if {@code replyStatus} does not fit in a byte or is not
+     *     greater than {@link ReplyStatus#UserException}
      */
     public DispatchException(int replyStatus) {
         this(replyStatus, null, null);
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/DispatchException.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/DispatchException.java
@@ -12,7 +12,8 @@ public class DispatchException extends LocalException {
     /**
      * Constructs a DispatchException with the specified reply status, message, and cause.
      *
-     * @param replyStatus the reply status as an int (see {@link ReplyStatus})
+     * @param replyStatus the reply status as an int (see {@link ReplyStatus}); it must be greater
+     *     than {@link ReplyStatus#UserException}
      * @param message the detail message
      * @param cause the cause
      */
@@ -24,7 +25,8 @@ public class DispatchException extends LocalException {
     /**
      * Constructs a DispatchException with the specified reply status and message.
      *
-     * @param replyStatus the reply status as an int (see {@link ReplyStatus})
+     * @param replyStatus the reply status as an int (see {@link ReplyStatus}); it must be greater
+     *     than {@link ReplyStatus#UserException}
      * @param message the detail message
      */
     public DispatchException(int replyStatus, String message) {
@@ -34,7 +36,8 @@ public class DispatchException extends LocalException {
     /**
      * Constructs a DispatchException with the specified reply status.
      *
-     * @param replyStatus the reply status as an int (see {@link ReplyStatus})
+     * @param replyStatus the reply status as an int (see {@link ReplyStatus}); it must be greater
+     *     than {@link ReplyStatus#UserException}
      */
     public DispatchException(int replyStatus) {
         this(replyStatus, null, null);
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instrumentation/Observer.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instrumentation/Observer.java
@@ -7,14 +7,12 @@ package com.zeroc.Ice.Instrumentation;
  */
 public interface Observer {
     /**
-     * This method is called when the instrumented object is created or when the observer is
-     * attached to an existing object.
+     * Notifies the observer that an instrumented object was created.
      */
     void attach();
 
     /**
-     * This method is called when the instrumented object is destroyed and as a result the observer
-     * detached from the object.
+     * Notifies the observer that an instrumented object was destroyed.
      */
     void detach();
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instrumentation/ObserverUpdater.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instrumentation/ObserverUpdater.java
@@ -3,7 +3,7 @@
 package com.zeroc.Ice.Instrumentation;
 
 /**
- * The observer updater interface. This interface is implemented by the Ice run-time and an instance of this interface
+ * The observer updater interface. This interface is implemented by the Ice runtime and an instance of this interface
  * is provided by the Ice communicator on initialization to the {@link CommunicatorObserver} object set with the
  * communicator initialization data. The Ice communicator calls {@link CommunicatorObserver#setObserverUpdater} to
  * provide the observer updater. This interface can be used by add-ins implementing the {@link CommunicatorObserver}
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
@@ -300,6 +300,7 @@ public interface ObjectPrx {
      *
      * @param connection the fixed proxy connection
      * @return a fixed proxy bound to the given connection
+     * @throws IllegalArgumentException if the connection is null or not a valid Ice connection
      */
     ObjectPrx ice_fixed(Connection connection);
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
@@ -209,6 +209,7 @@ public interface ObjectPrx {
      *
      * @param newIdentity the identity for the new proxy
      * @return a proxy with the new identity
+     * @throws IllegalArgumentException if the name of the new identity is empty
      */
     ObjectPrx ice_identity(Identity newIdentity);
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/PluginManager.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/PluginManager.java
@@ -12,6 +12,7 @@ public interface PluginManager {
      * initialization, the communicator calls {@link Plugin#destroy} on the plug-ins that have already been initialized.
      *
      * @throws InitializationException if the plug-ins have already been initialized
+     * @throws PluginInitializationException if a plug-in fails to initialize
      */
     void initializePlugins();
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/PluginManager.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/PluginManager.java
@@ -28,6 +28,7 @@ public interface PluginManager {
      *
      * @param name the plug-in's name
      * @return the plug-in
+     * @throws CommunicatorDestroyedException if the communicator has been destroyed
      * @throws NotRegisteredException if no plug-in is found with the given name
      */
     Plugin getPlugin(String name);
@@ -37,6 +38,7 @@ public interface PluginManager {
      *
      * @param name the plug-in's name
      * @param pi the plug-in
+     * @throws CommunicatorDestroyedException if the communicator has been destroyed
      * @throws AlreadyRegisteredException if a plug-in already exists with the given name
      */
     void addPlugin(String name, Plugin pi);
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
@@ -336,6 +336,9 @@ public final class Properties {
      *
      * @param key the property key
      * @param value the property value
+     * @throws InitializationException if the key is empty
+     * @throws PropertyException if the key is an unknown Ice property, or is in an opt-in property
+     *     prefix that has not been explicitly enabled
      * @see #getProperty
      */
     public void setProperty(String key, String value) {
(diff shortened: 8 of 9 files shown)


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
```diff
@@ -336,6 +336,9 @@ public synchronized Map<String, String> getPropertiesForPrefix(String prefix) {
      *
      * @param key the property key
      * @param value the property value
+     * @throws InitializationException if the key is empty
+     * @throws PropertyException if the key is an unknown Ice property, or is in an opt-in property
+     *     prefix that has not been explicitly enabled
```
InsertCreativityHere: Minor, but it sounds weird to say the "property" is "in a property prefix". It's really that this property is part of an opt-in property array, and that we identify the array via its prefix.

Maybe something more like that? But feel free to tweak based on the above explanation.
```suggestion
     * @throws PropertyException if the key is an unknown Ice property, or belongs to an opt-in 
     *     property prefix that has not been explicitly enabled
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -205,6 +205,8 @@ public String identityToString(Identity ident) {
      *
      * @param name the object adapter name
      * @return the new object adapter
+     * @throws InitializationException if the object adapter configuration is invalid
```
bernardnormier: `ObjectAdapterFactory.createObjectAdapter` throws `CommunicatorDestroyedException` as its first check — worth documenting on all four overloads since this PR adds it to the other communicator methods.

```suggestion
     * @throws CommunicatorDestroyedException if the communicator has been destroyed
     * @throws InitializationException if the object adapter configuration is invalid
```

bernardnormier (commented):
I verified every `@throws` in the table against the implementation and the Instrumentation wording against `Instrumentation.h` — all accurate, including the "Gets…" vs "Notifies…" call on the getters. Two coverage findings inline, same category as the gaps this PR closes:

1. The four `createObjectAdapter` overloads also throw `CommunicatorDestroyedException` — it's the first check in `ObjectAdapterFactory.createObjectAdapter` — but don't document it, while `getDefaultLocator`/`createAdmin`/`getPlugin`/`addPlugin` now do.
2. `createAdmin`'s existing `InitializationException` line covers only one of its three triggers.

Plus a trivial punctuation-consistency trim on the two new Connection `@throws`.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -228,6 +230,8 @@ public ObjectAdapter createObjectAdapter(String name) {
      *     calling {@link #createObjectAdapter(String)}.
      * @return the new object adapter
      * @throws IllegalArgumentException if the provided name is empty and sslEngineFactory is non-null
+     * @throws InitializationException if the object adapter configuration is invalid
```
bernardnormier:
```suggestion
     * @throws CommunicatorDestroyedException if the communicator has been destroyed
     * @throws InitializationException if the object adapter configuration is invalid
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -247,6 +251,8 @@ public ObjectAdapter createObjectAdapter(String name, SSLEngineFactory sslEngine
      * @param name the object adapter name
      * @param endpoints the endpoints of the object adapter
      * @return the new object adapter
+     * @throws InitializationException if the object adapter configuration is invalid
```
bernardnormier:
```suggestion
     * @throws CommunicatorDestroyedException if the communicator has been destroyed
     * @throws InitializationException if the object adapter configuration is invalid
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -269,6 +275,8 @@ public ObjectAdapter createObjectAdapterWithEndpoints(String name, String endpoi
      *     through {@code Ice.SSL} configuration properties. Passing {@code null} is equivalent to
      *     calling {@link #createObjectAdapterWithEndpoints(String, String)}.
      * @return the new object adapter
+     * @throws InitializationException if the object adapter configuration is invalid
```
bernardnormier:
```suggestion
     * @throws CommunicatorDestroyedException if the communicator has been destroyed
     * @throws InitializationException if the object adapter configuration is invalid
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -486,6 +495,8 @@ private CommunicatorFlushBatch _iceI_flushBatchRequestsAsync(CompressBatch compr
      *     after creating and activating this adapter.
      * @param adminId the identity of the Admin object
      * @return a proxy to the main ("") facet of the Admin object
+     * @throws CommunicatorDestroyedException if the communicator has been destroyed
+     * @throws IllegalArgumentException if the admin identity is not valid
      * @throws InitializationException if this method is called more than once
```
bernardnormier: `Instance.createAdmin` also throws `InitializationException` when the Admin facility is disabled and when `Ice.Admin.Endpoints` is not set (Instance.java:298-309) — since this PR touches this `@throws` block, it can cover all three triggers.

```suggestion
     * @throws InitializationException if the Admin object is already created, the Admin facility is
     *     disabled, or {@code Ice.Admin.Endpoints} is not set
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
```diff
@@ -107,6 +107,8 @@ public interface Connection {
      * Returns the connection information.
      *
      * @return the connection information
+     * @throws LocalException if the connection is closed; the exception provides the reason for the
+     *     closure (see {@link #throwException}).
```
bernardnormier, line 111: Trivial: the other `@throws` entries (including the ones added in this PR) have no trailing period.

```suggestion
     * @throws LocalException if the connection is closed; the exception provides the reason for the
     *     closure (see {@link #throwException})
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
```diff
@@ -115,6 +117,8 @@ public interface Connection {
      *
      * @param rcvSize the size of the receive buffer
      * @param sndSize the size of the send buffer
+     * @throws LocalException if the connection is closed; the exception provides the reason for the
+     *     closure (see {@link #throwException}).
```
bernardnormier, line 121:
```suggestion
     * @throws LocalException if the connection is closed; the exception provides the reason for the
     *     closure (see {@link #throwException})
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/PluginManager.java
```diff
@@ -28,6 +28,7 @@ public interface PluginManager {
      *
```
InsertCreativityHere, line 29: There seems to be a missing `@throws PluginInitializationException` for the `initializePlugins` method in this file. Is there a reason to not document this exception like the others?

Note that despite it's name, `PluginInitializationException` is not a subclass of `InitializationException`.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
```diff
@@ -209,6 +209,7 @@ CompletableFuture<Object.Ice_invokeResult> ice_invokeAsync(
      *
      * @param newIdentity the identity for the new proxy
      * @return a proxy with the new identity
+     * @throws IllegalArgumentException if the name of the new identity is empty
      */
     ObjectPrx ice_identity(Identity newIdentity);
```
InsertCreativityHere, line 215: `ice_fixed` throws `IllegalArgumentException` if you provide a null or invalid connection, but it is also still missing an `@throws` tag. Is there a reason to not document this exception?

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
InsertCreativityHere: This sounds like it should possessive to me, but I think it's still grammatically correct without it.
```suggestion
     * @throws InitializationException if the object adapter's configuration is invalid
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
InsertCreativityHere:
```suggestion
     * @throws InitializationException if the object adapter's configuration is invalid
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
InsertCreativityHere:
```suggestion
     * @throws InitializationException if the object adapter's configuration is invalid
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
InsertCreativityHere:
```suggestion
     * @throws InitializationException if the object adapter's configuration is invalid
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
```diff
@@ -412,6 +420,7 @@ public void setDefaultRouter(RouterPrx router) {
      * Gets the default locator of this communicator.
      *
      * @return the default locator of this communicator
+     * @throws CommunicatorDestroyedException if the communicator has been destroyed
```
InsertCreativityHere, line 427: I don't know about adding this `@throws`.

Nearly every single method in this file throws `CommunicatorDestroyedException`. So why choose to only document it on this one seemingly random method? If we're going to document it on `getDefaultLocator` we should do it for other methods too.

That being said, I'm not sure there's much value in adding this `@throws` to every single of the many methods for which it's true.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
InsertCreativityHere: There's actually a whole bunch of reasons why `InitializationException` can be thrown:
- `throw new InitializationException("Admin already created");`
- `throw new InitializationException("Admin is disabled");`
- `throw new InitializationException("Ice.Admin.Endpoints is not set");`
- `throw new InitializationException("Locator knows nothing about server '" + serverId + "'");`

Not saying we should list out all the reasons! But maybe switch "if this method is called more than once" to something more general...

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
InsertCreativityHere: The period is actually correct here. The rule in Java is that you use periods iff your comment is a sentence.
So if it only said "if the connection is closed", no period, since that isn't a stand-alone sentence.

But this comment rises to the level of sentence, and should use a period.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
InsertCreativityHere: Same as above, a period is correct here because this is a complete sentence.

There is no consistency issue with the other comments. See the rules for Java-doc-comments here (https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html). The meat of the rules are only stated in `@param`'s documentation, but the rules hold in general, even for `@throws`.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instrumentation/ObserverUpdater.java
```diff
@@ -3,7 +3,7 @@
 package com.zeroc.Ice.Instrumentation;
 
 /**
- * The observer updater interface. This interface is implemented by the Ice run-time and an instance of this interface
+ * The observer updater interface. This interface is implemented by the Ice runtime and an instance of this interface
```
InsertCreativityHere, line 6: I agree with this change!

But, just a note, we still have 3 places where we use the "run-time" spelling in `Instrumentation.h`.
Not for this PR, but we should probably remove these uses from C++ like we're doing here too.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/DispatchException.java
```diff
@@ -12,7 +12,8 @@ public class DispatchException extends LocalException {
     /**
      * Constructs a DispatchException with the specified reply status, message, and cause.
      *
-     * @param replyStatus the reply status as an int (see {@link ReplyStatus})
+     * @param replyStatus the reply status as an int (see {@link ReplyStatus}); it must be greater
+     *     than {@link ReplyStatus#UserException}
```
InsertCreativityHere, line 16: We don't actually enforce this anywhere I see. Should we reject `replyStatus`s that don't uphold this? The other constructors delegate to this one, so we'd only need to do it in one single location.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Added `@throws CommunicatorDestroyedException` to all four `createObjectAdapter` overloads (it is the first check in `ObjectAdapterFactory.createObjectAdapter`). Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Reworded to cover all three triggers: `Admin` already created, `Admin` disabled, and `Ice.Admin.Endpoints` not set. Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
pepone: 🤖 **Claude:** Reworded to `belongs to an opt-in property prefix that has not been explicitly enabled`. Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/PluginManager.java
pepone: 🤖 **Claude:** Good catch — `initializePlugins` does throw `PluginInitializationException` when a plug-in fails to initialize (`PluginManagerI.java:32-35`), and it is indeed not a subclass of `InitializationException`. Added the `@throws`. Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
pepone: 🤖 **Claude:** Added `@throws IllegalArgumentException if the connection is null or not a valid Ice connection`, matching `_ObjectPrxI.ice_fixed`. Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Switched to the possessive `object adapter's configuration` on all four overloads. Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Replaced the single-trigger wording with all three reasons: `Admin` already created, `Admin` disabled, and `Ice.Admin.Endpoints` not set. Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/DispatchException.java
pepone: 🤖 **Claude:** Agreed — added the check in the single delegating constructor. It now throws `IllegalArgumentException` when `replyStatus` is not greater than `ReplyStatus.UserException` or does not fit in a byte, matching the C++ (`LocalExceptions.cpp`) and C# (`LocalExceptions.cs`) mappings. Verified it rejects `Ok`/`UserException`/`>255` and accepts `2..255`, and that no internal caller is affected — the wire path masks the status to 0-255 and only constructs a `DispatchException` for statuses above `UserException`. Done in 8a40c8727f.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
```diff
@@ -336,6 +336,9 @@ public synchronized Map<String, String> getPropertiesForPrefix(String prefix) {
      *
      * @param key the property key
      * @param value the property value
+     * @throws InitializationException if the key is empty
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Leaving this as-is. The `key == null` branch is a defensive guard for a programming error rather than a documented contract, and the other `@throws` entries in this file describe only the meaningful (empty-key) case. Happy to spell out the null case too if you would prefer.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ObjectPrx.java
```diff
@@ -209,6 +209,7 @@ CompletableFuture<Object.Ice_invokeResult> ice_invokeAsync(
      *
      * @param newIdentity the identity for the new proxy
      * @return a proxy with the new identity
+     * @throws IllegalArgumentException if the name of the new identity is empty
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Same reasoning as the `setProperty` thread — leaving the null-`name` case undocumented, since it is a defensive guard for a programming error rather than part of the documented contract. The empty-name case is the meaningful one for callers.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instrumentation/Observer.java
```diff
@@ -7,14 +7,12 @@
  */
 public interface Observer {
     /**
-     * This method is called when the instrumented object is created or when the observer is
-     * attached to an existing object.
+     * Notifies the observer that an instrumented object was created.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** This wording is a deliberate match of the canonical C++ docs (`Ice/Instrumentation.h`: "Notifies the observer that an instrumented object was created."), which is the goal of this PR. The re-attach-to-an-existing-object path you mention does exist, but the C++ mapping documents it the same way, so I am keeping Java aligned. Any refinement should be a cross-mapping change led by C++.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instrumentation/Observer.java
```diff
@@ -7,14 +7,12 @@
...
      */
     void attach();
 
     /**
-     * This method is called when the instrumented object is destroyed and as a result the observer
-     * detached from the object.
+     * Notifies the observer that an instrumented object was destroyed.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Same as the `attach` thread — this matches the canonical C++ wording ("Notifies the observer that an instrumented object was destroyed."). The still-alive-detach case is real, but refining it should be a cross-mapping change starting from C++.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: 🤖 **Claude:** Keeping this one: the sibling `getDefaultRouter` already documents `@throws CommunicatorDestroyedException`, so this restores symmetry between the two accessors rather than singling out a random method. I agree we should not sprinkle it across every method — this is scoped to matching the existing `getDefaultRouter` doc.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
pepone: 🤖 **Claude:** Keeping the period here. As Austin notes below, this entry is a complete sentence ("... is closed; the exception provides the reason for the closure ..."), so the trailing period is correct per the Javadoc convention, whereas the fragment-style entries elsewhere correctly omit it.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Connection.java
pepone: 🤖 **Claude:** Same as the `getInfo` thread — this is a complete sentence, so keeping the period per the Javadoc convention.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Instrumentation/ObserverUpdater.java
pepone: 🤖 **Claude:** Agreed — the three remaining `run-time` uses in C++ `Instrumentation.h` are worth cleaning up in a follow-up; leaving them out of this Java PR as you suggest.

pepone: 🤖 **Claude:** Filed as #6082 for the C++ cleanup.

bernardnormier (approved):
Verified the feedback commit: the DispatchException enforcement matches C++ (LocalExceptions.cpp throws std::invalid_argument for replyStatus <= UserException, same placement) and C#'s documented ArgumentOutOfRangeException, with the byte-fit check as the Java analog of taking a ReplyStatus enum — and the only internal constructions (ProxyOutgoingAsyncBase) sit in branches that already guarantee a valid wire status. The new ice_fixed and initializePlugins tags also match the implementations. No changelog is consistent with how we treat enforcement of a documented contract.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
InsertCreativityHere: The fact that `getDefaultRouter` documents this exception is not written in stone, and does not, on it's own, force our hand on whether to document it on all other sites.

For example, even with all of the `@throws CommunicatorDestroyedException` that have been added here in this PR, all of these methods also throw this same exception on the same paths, but do not document it: `stringToProxy, propertyToProxy, createObjectAdapterWithRouter, setDefaultObjectAdapter, setDefaultRouter, setDefaultLocator, _iceI_flushBatchRequestsAsync, getAdmin, addAdminFacet, removeAdminFacet, findAdminFacet, findAllAdminFacets`

Why document this exception on a fairly un-special methods like `getDefaultRouter` and `getDefaultLocator`, but not any of the above? We should at least be consistent here, documenting it everywhere it's meaningful, or only on the constructors, but certainly not just these 2 `getDefault` methods and none of the other equally important ones.

InsertCreativityHere (approved):
Looks good to me.

Still think we should be a little more consistent about whether or not to document `CommunicatorDestroyedException` on the various methods of `Communicator`. Not clear that it's worth the added verbosity to me, but if it is, we shouldn't _only_ document it on a couple random seeming methods.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Communicator.java
pepone: Ok, I would drop the addition and open a issue to clean this up, probably in other language mappings too. I think it is best to list it everywhere, I would not expect that a method throws exceptions it doesn't list other than system exceptions.

pepone: 🤖 **Claude:** Reverted the `getDefaultLocator` addition in 227f269a5e — agreed it should not be documented on just these two `getDefault*` methods while the equally-affected siblings are not. Opened #6094 to adopt a consistent policy across `Communicator` (and the C++/C# mappings).

PR #6083, opened 2026-07-21
Discard DataStorm partial updates that have no base value

pepone:
A partial update resolves against the key's current value, but `publish`, `queue`, and `initSamples` recorded every sample—including a value-less `Remove`—as the per-key base. Resolving a later partial update against such a base cloned a value the sample doesn't have: applications using class-typed values typically crashed (`ice_clone` on a nil handle), and other value types silently resurrected the removed key from a default-constructed value.

With this change:

- A `Remove` clears the key's per-key base, on the writer and on the reader (live samples, init samples, and the discard-policy base seeding).
- A partial update for a key with no base value—no full value was written yet, or the key was removed—is discarded: the writer logs a warning and publishes nothing, matching the reader-side drop introduced by #5882. The `partialUpdate` API docs now state this contract.
- `Cloner<std::shared_ptr<T>>` (Ice values) clones a null value to null instead of dereferencing it, and the `setUpdater` wrapper and `noOpUpdater` treat a value-less previous sample as absent: a legitimate null class value (`writer.add(nullptr)`) no longer crashes the resolution path.
- `SampleT::_value` and the updater wrapper's local value are value-initialized: a remove sample's value was previously indeterminate for scalar value types (and those indeterminate bytes were encoded on the wire).

The `partial` test gains two scenarios: a writer discarding a partial update published after its own `remove`, and a reader discarding a partial update from one writer that arrives after another writer's `remove` of the same key (previously both crashed for class-typed topics, reproduced red→green). The `api` test's sample block is updated to the new contract.

Fixes zeroc-ice/ice-audit#103.

### DIFF
--- a/changelog.d/datastorm/partial-update-no-base.md
+++ b/changelog.d/datastorm/partial-update-no-base.md
@@ -1,7 +1,7 @@
-- Fixed a bug in DataStorm where a partial update for a key with no current value—no full value was written for the
-  key, or the key was removed—was resolved against an invalid base value: applications using class-typed values
-  typically crashed, and with other value types the update was silently applied to a default-constructed value,
-  resurrecting removed keys. A remove now clears the key's value, and a partial update for a key with no value is
-  discarded, by both writers and readers.
+- Fixed a bug in DataStorm where a partial update for a key with no current value was resolved against an invalid
+  base value: applications using class-typed values typically crashed, and with other value types the update was
+  silently applied to a default-constructed value, resurrecting removed keys. A key has no current value when no full
+  value was written for it yet, or when it was removed. A remove now clears the key's value; publishing a partial
+  update for a key with no value now throws `std::logic_error` on the writer and is discarded on the reader.
 - Fixed a bug in DataStorm where a remove sample's value was indeterminate for scalar value types: `getValue()` on
   the remove sample returned arbitrary data instead of a default-constructed value.
--- a/changelog.d/datastorm/partial-update-no-base.md
+++ b/changelog.d/datastorm/partial-update-no-base.md
@@ -3,5 +3,5 @@
   silently applied to a default-constructed value, resurrecting removed keys. A key has no current value when no full
   value was written for it yet, or when it was removed. A remove now clears the key's value; publishing a partial
   update for a key with no value now throws `std::logic_error` on the writer and is discarded on the reader.
-- Fixed a bug in DataStorm where a remove sample's value was indeterminate for scalar value types: `getValue()` on
-  the remove sample returned arbitrary data instead of a default-constructed value.
+- Fixed a bug in DataStorm where calling `getValue()` on a sample carrying the `Remove` event returned indeterminate
+  data instead of the documented default-constructed value, when the topic's value type is a scalar type.
--- a/changelog.d/datastorm/partial-update-no-base.md
+++ b/changelog.d/datastorm/partial-update-no-base.md
@@ -1,0 +1,7 @@
+- Fixed a bug in DataStorm where a partial update for a key with no current value—no full value was written for the
+  key, or the key was removed—was resolved against an invalid base value: applications using class-typed values
+  typically crashed, and with other value types the update was silently applied to a default-constructed value,
+  resurrecting removed keys. A remove now clears the key's value, and a partial update for a key with no value is
+  discarded, by both writers and readers.
+- Fixed a bug in DataStorm where a remove sample's value was indeterminate for scalar value types: `getValue()` on
+  the remove sample returned arbitrary data instead of a default-constructed value.
--- a/changelog.d/datastorm/partial-update-no-base.md
+++ b/changelog.d/datastorm/partial-update-no-base.md
@@ -1,7 +1,6 @@
-- Fixed a bug in DataStorm where a partial update for a key with no current value was resolved against an invalid
-  base value: applications using class-typed values typically crashed, and with other value types the update was
-  silently applied to a default-constructed value, resurrecting removed keys. A key has no current value when no full
-  value was written for it yet, or when it was removed. A remove now clears the key's value; publishing a partial
-  update for a key with no value now throws `std::logic_error` on the writer and is discarded on the reader.
-- Fixed a bug in DataStorm where calling `getValue()` on a sample carrying the `Remove` event returned indeterminate
-  data instead of the documented default-constructed value, when the topic's value type is a scalar type.
+- Fixed the handling of partial updates for keys with no value. A key has no value when it was removed, or when no
+  full value was written for it yet. Previously, such a partial update crashed applications using class-typed
+  values; with other value types, it silently resurrected removed keys. Now a writer throws `std::logic_error`
+  when publishing such an update, and a reader discards incoming ones.
+- Fixed a bug where calling `getValue()` on a `Remove` sample could return indeterminate data instead of the
+  documented default value.
--- a/cpp/include/DataStorm/DataStorm.h
+++ b/cpp/include/DataStorm/DataStorm.h
@@ -767,8 +767,10 @@ namespace DataStorm
         /// function generates a SampleEvent::PartialUpdate sample with the given partial update value.
         /// The UpdateValue template parameter must match the UpdateValue type used to register the updater with
         /// the Topic::setUpdater method.
-        /// A partial update resolves against the key's current value. When the key has no value—no full value was
-        /// written yet, or the key was removed—the partial update is discarded and nothing is published.
+        /// A partial update resolves against the key's current value, so the key must have a current value when the
+        /// returned function is called: a full value was written for the key and the key was not since removed.
+        /// Calling the returned function for a key with no current value is an application error that throws
+        /// std::logic_error and publishes nothing.
         /// @param tag The partial update tag.
         template<typename UpdateValue>
         [[nodiscard]] std::function<void(const UpdateValue&)> partialUpdate(const UpdateTag& tag);
@@ -822,8 +824,10 @@ namespace DataStorm
         /// function generates a SampleEvent::PartialUpdate sample with the given partial update value.
         /// The UpdateValue template parameter must match the UpdateValue type used to register the updater with
         /// the Topic::setUpdater method.
-        /// A partial update resolves against the key's current value. When the key has no value—no full value was
-        /// written yet, or the key was removed—the partial update is discarded and nothing is published.
+        /// A partial update resolves against the key's current value, so the key must have a current value when the
+        /// returned function is called: a full value was written for the key and the key was not since removed.
+        /// Calling the returned function for a key with no current value is an application error that throws
+        /// std::logic_error and publishes nothing.
         /// @param tag The partial update tag.
         template<typename UpdateValue>
         [[nodiscard]] std::function<void(const Key&, const UpdateValue&)> partialUpdate(const UpdateTag& tag);
--- a/cpp/include/DataStorm/DataStorm.h
+++ b/cpp/include/DataStorm/DataStorm.h
@@ -1639,8 +1639,11 @@ namespace DataStorm
                                            const std::shared_ptr<DataStormI::Sample>& next,
                                            const Ice::CommunicatorPtr& communicator)
             {
-                // Callers discard partial updates that have no usable base before invoking the updater; assert that
-                // here, and keep the default-value fallback as defense in depth for release builds.
+                // Every updater call site ensures the previous sample exists and has a value before invoking the
+                // updater (the writer throws otherwise, the reader drops the sample), so this assert holds and the
+                // clone below always runs. The guarded branch is not a safe release-build fallback for a broken
+                // invariant: a default-constructed base is null for class-typed values, which the user's updater
+                // would dereference just as if the guard were absent.
                 assert(previous && previous->hasValue());
                 Value value{};
                 if (previous && previous->hasValue())
--- a/cpp/include/DataStorm/DataStorm.h
+++ b/cpp/include/DataStorm/DataStorm.h
@@ -10,6 +10,7 @@
 #include "Node.h"
 #include "Types.h"
 
+#include <cassert>
 #include <regex>
 
 #if defined(__clang__)
@@ -1634,8 +1635,9 @@ namespace DataStorm
                                            const std::shared_ptr<DataStormI::Sample>& next,
                                            const Ice::CommunicatorPtr& communicator)
             {
-                // A value-less previous sample is not a usable base: fall back to a default value. This is defense
-                // in depth: callers discard partial updates that have no usable base before invoking the updater.
+                // Callers discard partial updates that have no usable base before invoking the updater; assert that
+                // here, and keep the default-value fallback as defense in depth for release builds.
+                assert(previous && previous->hasValue());
                 Value value{};
                 if (previous && previous->hasValue())
                 {
--- a/cpp/include/DataStorm/DataStorm.h
+++ b/cpp/include/DataStorm/DataStorm.h
@@ -766,6 +766,8 @@ namespace DataStorm
         /// function generates a SampleEvent::PartialUpdate sample with the given partial update value.
         /// The UpdateValue template parameter must match the UpdateValue type used to register the updater with
         /// the Topic::setUpdater method.
+        /// A partial update resolves against the key's current value. When the key has no value—no full value was
+        /// written yet, or the key was removed—the partial update is discarded and nothing is published.
         /// @param tag The partial update tag.
         template<typename UpdateValue>
         [[nodiscard]] std::function<void(const UpdateValue&)> partialUpdate(const UpdateTag& tag);
@@ -819,6 +821,8 @@ namespace DataStorm
         /// function generates a SampleEvent::PartialUpdate sample with the given partial update value.
         /// The UpdateValue template parameter must match the UpdateValue type used to register the updater with
         /// the Topic::setUpdater method.
+        /// A partial update resolves against the key's current value. When the key has no value—no full value was
+        /// written yet, or the key was removed—the partial update is discarded and nothing is published.
         /// @param tag The partial update tag.
         template<typename UpdateValue>
         [[nodiscard]] std::function<void(const Key&, const UpdateValue&)> partialUpdate(const UpdateTag& tag);
@@ -1630,8 +1634,10 @@ namespace DataStorm
                                            const std::shared_ptr<DataStormI::Sample>& next,
                                            const Ice::CommunicatorPtr& communicator)
             {
-                Value value;
-                if (previous)
+                // A value-less previous sample is not a usable base: fall back to a default value. This is defense
+                // in depth: callers discard partial updates that have no usable base before invoking the updater.
+                Value value{};
+                if (previous && previous->hasValue())
                 {
                     value = Cloner<Value>::clone(
                         std::static_pointer_cast<DataStormI::SampleT<Key, Value, UpdateTag>>(previous)->getValue());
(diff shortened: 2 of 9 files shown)


pepone:
Follow-up for 3.9.0: #6084 — `SampleT::decode` skips empty encoded values (custom encoders), so such full samples are rejected as partial-update bases by the `hasValue()` guard added here. Deliberately out of scope for this fix since it changes decode/`hasValue()` semantics.

cpp/include/DataStorm/DataStorm.h
```diff
@@ -1630,8 +1634,10 @@ namespace DataStorm
                                            const std::shared_ptr<DataStormI::Sample>& next,
                                            const Ice::CommunicatorPtr& communicator)
             {
-                Value value;
-                if (previous)
+                // A value-less previous sample is not a usable base: fall back to a default value. This is defense
+                // in depth: callers discard partial updates that have no usable base before invoking the updater.
+                Value value{};
+                if (previous && previous->hasValue())
```
pepone, line 1649: Would be nice to pair this with an assert so that it fails loudly on debug.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -1154,8 +1180,15 @@ DataWriterI::publish(const shared_ptr<Key>& key, const shared_ptr<Sample>& sampl
     {
         assert(!sample->hasValue());
         auto p = _lastByKey.find(key);
-        _parent->getUpdater(
-            sample->tag)(p == _lastByKey.end() ? nullptr : p->second, sample, _parent->instance()->getCommunicator());
+        if (p == _lastByKey.end())
+        {
+            // No base to resolve the partial update against (the key was never set, or was removed): discard the
+            // update rather than apply it to a default-constructed value, like the reader side does.
```
pepone: Is this comment accurate, doesn't the reader skip it too?

cpp/test/DataStorm/partial/Writer.cpp
```diff
@@ -266,6 +266,50 @@ void ::Writer::run(int argc, char* argv[])
...
+    twoWritersTopic.setUpdater<float>("price", [](StockPtr& stock, float price) { stock->price = price; });
+    cout << "testing partial update after another writer's remove... " << flush;
+    {
+        auto writer1 = makeSingleKeyWriter(twoWritersTopic, "AAPL", "writer1");
+        auto writer2 = makeSingleKeyWriter(twoWritersTopic, "AAPL", "writer2");
+        writer1.waitForReaders();
+        writer2.waitForReaders();
+        writer1.add(make_shared<Stock>(12.0f, 13.0f, 14.0f));
+        writer2.add(make_shared<Stock>(100.0f, 101.0f, 102.0f));
+        writer1.remove();
```
pepone, line 311: The test assumes the order between writer events is maintain over the wire, is does safe assumption or will this result in a flaky test?

cpp/include/DataStorm/DataStorm.h
pepone: 🤖 **Claude:** Added `assert(previous && previous->hasValue())` here, and the same assert in `TopicI.cpp`'s `noOpUpdater` (same invariant, same call sites). All three updater call sites guard the base before invoking — the reader paths check `hasValue()` explicitly and the writer's `_lastByKey` only ever holds valued samples — so the assert fails loudly on debug while the default-value fallback stays for release. Done in f710528241

cpp/src/DataStorm/DataElementI.cpp
pepone: 🤖 **Claude:** The comment was accurate but ambiguously worded: the reader does discard such updates too (`DataReaderI::queue` and the init-samples path drop a partial update with no usable base). "like the reader side does" dangled after "apply it to a default-constructed value", inviting the wrong reading. Reworded into two sentences so the comparison unambiguously attaches to the discard. Done in f710528241

cpp/test/DataStorm/partial/Writer.cpp
pepone: 🤖 **Claude:** Safe, not flaky. Both writers are in the same node and DataStorm sessions are per peer node, so both deliver over the same session connection; the sequential publishes are queued on it in program order. On the receiving side, `Instance::init` sets `DataStorm.Node.Server.ThreadPool.Serialize=1` precisely "to ensure that samples are processed in the order they are received", noting it is "especially important for PartialUpdate samples". The rest of the suite already relies on this same-connection ordering (e.g. the add → partial(15) → partial(18) sequence asserted at the top of Reader.cpp), so this test adds no new assumption. Extended the scenario comment to state the guarantee. Done in f710528241

changelog.d/datastorm/partial-update-no-base.md
```diff
@@ -0,0 +1,7 @@
+- Fixed a bug in DataStorm where a partial update for a key with no current value—no full value was written for the
+  key, or the key was removed—was resolved against an invalid base value: applications using class-typed values
+  typically crashed, and with other value types the update was silently applied to a default-constructed value,
+  resurrecting removed keys. A remove now clears the key's value, and a partial update for a key with no value is
+  discarded, by both writers and readers.
+- Fixed a bug in DataStorm where a remove sample's value was indeterminate for scalar value types: `getValue()` on
+  the remove sample returned arbitrary data instead of a default-constructed value.
```
externl: Two nits on this fragment:

- `changelog.d/README.md` maps `datastorm/` to the "Ice Service Changes > DataStorm" section, and the collation script generates that heading — so "in DataStorm" is redundant here.
- The em dashes are unspaced, so `value—no full value` reads as a hyphenated compound rather than an aside. Every other em dash in `changelog.d/` is spaced (`js/5876.md`, `service-icestorm/5848.md`, `datastorm/reader-partial-update-robustness.md`). Since the aside also splits the subject from its verb, it's cleaner as its own sentence.

```suggestion
- Fixed a bug where a partial update for a key with no current value was resolved against an invalid base value:
  applications using class-typed values typically crashed, and with other value types the update was silently
  applied to a default-constructed value, resurrecting removed keys. A key has no current value when no full value
  was written for it yet, or when it was removed. A remove now clears the key's value, and a partial update for a
  key with no value is discarded, by both writers and readers.
- Fixed a bug where a remove sample's value was indeterminate for scalar value types: `getValue()` on the remove
  sample returned arbitrary data instead of a default-constructed value.
```

Note that 7 of the 10 existing `changelog.d/datastorm/` fragments do open with "Fixed a bug in DataStorm", so this makes the new one the odd one out until the others are swept. Happy to leave it as-is if you'd rather keep the section internally consistent for now.

externl (commented):
Verified the base-clearing is consistent across the three sites, and that mixed-version deployments are safe both ways — `toSample` marshals `_encodedValue` (the update payload), so a writer's local merge never reaches the wire.

- The `initSamples` half of the fix is untested: every new scenario calls `waitForReaders()` first, so all samples go through `queue()` and `initSamples` only ever sees an empty batch. Seconds Copilot's r3631451625.
- `DataElementI.cpp:1183` checks only `p == _lastByKey.end()`, but the updater it calls asserts `previous && previous->hasValue()`. The reader guards spell out both halves — worth matching.
- "Defense in depth for release builds" overstates the fallback: for class types `Value value{}` is null, so the user's updater still derefs null in release. `noOpUpdater` is the same shape — `setValue` sets `_hasValue = true` unconditionally, so release silently resurrects the key from a default value.
- The writer warning at `:1189` is unconditional and doesn't name the key; the reader's equivalent is trace-gated and names the sample id. `Key::toString()` exists.
- `DataStorm.h:770`/`:825` repeat the unspaced em dashes from the changelog, and these ones render into the published API docs.

Nits: the discard-policy branch deliberately doesn't clear the base for a discarded remove — correct for both policies, but the comment only explains the seeding exclusion; `collectBases`' `!hasValue()` check at `:1508` is now dead.

Finding 1 is the one I'd want before merge, the rest is polish.

externl (approved):

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -1154,8 +1180,16 @@ DataWriterI::publish(const shared_ptr<Key>& key, const shared_ptr<Sample>& sampl
...
         auto p = _lastByKey.find(key);
-        _parent->getUpdater(
-            sample->tag)(p == _lastByKey.end() ? nullptr : p->second, sample, _parent->instance()->getCommunicator());
+        if (p == _lastByKey.end())
+        {
+            // No base to resolve the partial update against (the key was never set, or was removed): discard the
+            // update rather than apply it to a default-constructed value. The reader side discards such updates
+            // the same way.
+            Warning out(_traceLevels->logger);
+            out << this << ": discarded partial update: no base value for the key";
```
bernardnormier: Should this be an error rather than a warning-and-discard? A writer-side missing base is always the application's own sequencing bug: `_lastByKey` holds only this writer's own published samples, so another writer's remove can never clear it — the two-writers test shows exactly that (writer2's partial still resolves after writer1's remove; only the reader discards it). Unlike the reader, there's no distributed-race scenario where this discard is innocent.

And this call path already reports errors to the application by exception: the updater call below is deliberately not wrapped in try/catch, so a throwing user updater or a `Decoder` failure already propagates synchronously out of the application's `partialUpdate(...)` call. Something like `throw std::logic_error("cannot apply partial update: the key has no value")` would use the same channel, match the existing `std::logic_error("no sample")` idiom in DataStorm.h, and surface the bug at the offending call site instead of in a log the application may never read. Backward compatibility isn't a concern: the pre-fix behavior on this path was a crash or silent key resurrection, so no correct application can be relying on it.

The reader-side discard should stay as-is — there a missing base is often nobody's bug, and there's no application frame on the stack to throw into.

If adopted, the `partialUpdate` doc comments, the api/partial writer tests, and the first changelog entry need the corresponding writer-throws wording. If you'd rather keep the warning, consider including the key (or at least the tag) in the message — for an any-key writer, the key is what the user needs to locate the bug.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -971,8 +984,9 @@ DataReaderI::queue(
 
         // The discard only affects the application-visible delivery: when the key has no base yet, still record the
         // discarded sample as the base partial updates resolve against, otherwise a later partial update on the key
-        // would have no base.
-        if (sample->event != DataStorm::SampleEvent::PartialUpdate && _lastByKey.find(sample->key) == _lastByKey.end())
+        // would have no base. A remove carries no value and cannot serve as a base.
+        if (sample->event != DataStorm::SampleEvent::PartialUpdate && sample->event != DataStorm::SampleEvent::Remove &&
```
bernardnormier, line 997: Now that a remove erases the map entry, the reader can no longer distinguish "removed" from "never seen", so this branch can resurrect a removed key. With `DiscardPolicy::SendTime` and two writers: writer A's remove (ts=10) is delivered and erases the base; writer B's stale full update (ts=5) arrives late, is correctly not delivered, but re-seeds the base here; writer B's next partial update then resolves against that stale value and is delivered — a partial update surfacing for a key the application last saw removed, with no intervening Add/Update. Same applies to the `initSamples` copy of this branch.

Before this PR the identical sequence hit the crash being fixed, so this is strictly an improvement, and distinguishing the two states would need a remove tombstone — likely not worth it. Just flagging it as an explicit accepted trade-off; a sentence here would keep someone from rediscovering it as a bug later.

bernardnormier (commented):
The fix checks out end-to-end: all three updater call sites are guarded before invocation, the discard-policy branches return before the new erase (so a stale remove can't clear a legitimate base), the writer's reconnect path needed no change (`collectBases` already skips value-less bases), and the wire format is unchanged — removes still encode a value, now a deterministic one, so old readers decode a clean default and new readers skip decoding entirely.

A few notes in addition to the inline comments:

- The writer-side guard checks only `p == _lastByKey.end()`, while the reader paths also check `!p->second->hasValue()`. The erase-on-remove invariant makes the extra check unreachable on the writer today, but both sides maintain the same map with the same semantics, so mirroring the defensive check keeps them reasoning identically.
- Of Copilot's coverage notes, the `initSamples` one is the one worth acting on: the batch-local `previousByKey` erase is new logic no test currently executes — both new scenarios use already-connected readers, so every sample goes through `queue`.
- Pre-existing and optional: `initSamples` assigns `_lastByKey = std::move(previousByKey)` both in the `valid.empty()` early return and in the fallthrough right after it; hoisting the assignment above the check would remove the duplication.

changelog.d/datastorm/partial-update-no-base.md
bernardnormier: "a remove sample's value was indeterminate" is written from the implementation's point of view — "remove sample" isn't a term a user necessarily knows. The user-visible contract lives on `getValue()`, whose doc already promises a default value when the sample event is Remove, so lead with that:

```suggestion
- Fixed a bug in DataStorm where calling `getValue()` on a sample carrying the `Remove` event returned indeterminate
  data instead of the documented default-constructed value, when the topic's value type is a scalar type.
```

(Independent of @externl's suggestion above — whichever lands second needs a trivial rebase.)

cpp/src/DataStorm/DataElementI.cpp
pepone: 🤖 **Claude:** Adopted — the writer now throws `std::logic_error("cannot apply a partial update to a key that has no value")` instead of warn-and-discard (870c7b55a6). Updated the `partialUpdate` doc comments, the api/partial writer tests (they assert the throw now), and the changelog. Reader-side discard left as-is. Kept the message generic rather than naming the key: the throw is synchronous at the `partialUpdate` call site so the app already has the key in-frame, and the single-key path publishes with a null key anyway.

cpp/src/DataStorm/DataElementI.cpp
pepone: 🤖 **Claude:** Added the accepted-trade-off note here and on the `initSamples` copy (3effe09b91). Agreed a remove tombstone isn't worth it.

changelog.d/datastorm/partial-update-no-base.md
pepone: 🤖 **Claude:** Applied your wording (3effe09b91).

pepone: 🤖 **Claude:** Em dashes spaced/split (the first entry was reworded for the throw regardless). Kept the "in DataStorm" opener for consistency with the 7/10 existing `datastorm/` fragments — happy to sweep them all together separately. (870c7b55a6 / 3effe09b91)

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -841,7 +844,17 @@ DataReaderI::initSamples(
             }
         }
         valid.push_back(sample);
-        previousByKey[sample->key] = sample;
+
+        // A remove clears the per-key base: the key has no value anymore, so a later partial update for the key
+        // has no base to resolve against and is discarded.
+        if (sample->event == DataStorm::SampleEvent::Remove)
+        {
+            previousByKey.erase(sample->key);
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Added a late-joining reader scenario to the `partial` test (add → remove → re-add → partial, delivered as init samples); confirmed via temporary instrumentation that it executes the `initSamples` base-clearing erase (3effe09b91). A cross-writer "peer partial dropped" variant isn't constructible without flakiness: init samples are delivered per-writer-element, each with its own `_lastByKey` snapshot (`SessionI::initSamples`), and `finalize` converts the first per-key init sample to a full value, so an in-batch partial always has a base. The reader-side drop stays covered by the existing live two-writer test.

cpp/include/DataStorm/Types.h
```diff
@@ -186,10 +186,13 @@ namespace DataStorm
...
     };
 
-    /// Cloner template specialization to clone shared Ice values using ice_clone.
+    /// Cloner template specialization to clone shared Ice values using ice_clone. A null value clones to null.
     template<typename T> struct Cloner<std::shared_ptr<T>, typename std::enable_if_t<std::is_base_of_v<Ice::Value, T>>>
     {
-        static std::shared_ptr<T> clone(const std::shared_ptr<T>& value) noexcept { return value->ice_clone(); }
+        static std::shared_ptr<T> clone(const std::shared_ptr<T>& value) noexcept
+        {
+            return value ? value->ice_clone() : nullptr;
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Added (57e9141530): a class-valued topic with no updater registered, `add(nullptr)` then a partial update — the no-op updater clones the null base to null. Verified red→green (crashes on the null deref without the `Cloner` fix).

cpp/include/DataStorm/InternalT.h
```diff
@@ -403,7 +403,9 @@ namespace DataStormI
 
     private:
         bool _hasValue;
-        Value _value;
+        // Value-initialized so a value-less sample (such as a remove) holds a determinate default value: scalar
+        // types would otherwise be indeterminate, and encodeValue encodes this member for remove samples.
+        Value _value{};
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Not adding a dedicated case. A `getValue() == 0` assertion on a scalar remove can't reliably distinguish the pre-fix indeterminate value (which could read as 0 by chance), so it wouldn't be a real regression guard. The value-initialization is type-generic, and the api test already asserts the documented default on a remove sample.

pepone:
🤖 **Claude:** Pushed `870c7b55a6`, `3effe09b91`, `57e9141530` addressing the review — inline replies on the specific threads. The review-body points that didn't have a thread:

- **Mirror the reader's `!hasValue()` on the writer guard** (@externl, @bernardnormier): done — `publish` now checks `p == end || !p->second->hasValue()`. Unreachable today (a remove erases the entry, so `_lastByKey` never holds a value-less sample), but both sides now consume `_lastByKey` with identical semantics.
- **"Defense in depth for release builds" overstates it** (@externl): reworded both the `setUpdater` wrapper and `noOpUpdater` comments. The guarded branch isn't a safe release fallback — for class values the default base is null (still dereferenced by the user updater) and `noOpUpdater`'s `setValue(nullptr)` resurrects the key. The comment now just documents the asserted invariant.
- **`collectBases` `!hasValue()` is now dead** (@externl): left it in place, as the same defensive pattern you asked me to add on the writer guard above. Happy to strip both if you'd rather go minimal.
- **Hoist the duplicated `_lastByKey = std::move(previousByKey)`** (@bernardnormier): done — kept the single assignment *after* the `onSamples` queueing so behavior under a throwing enqueue is unchanged.

Full DataStorm suite green (10/10 tests, unicast + multicast).

changelog.d/datastorm/partial-update-no-base.md
```diff
@@ -0,0 +1,7 @@
+- Fixed a bug in DataStorm where a partial update for a key with no current value was resolved against an invalid
+  base value: applications using class-typed values typically crashed, and with other value types the update was
+  silently applied to a default-constructed value, resurrecting removed keys. A key has no current value when no full
+  value was written for it yet, or when it was removed. A remove now clears the key's value; publishing a partial
+  update for a key with no value now throws `std::logic_error` on the writer and is discarded on the reader.
+- Fixed a bug in DataStorm where calling `getValue()` on a sample carrying the `Remove` event returned indeterminate
+  data instead of the documented default-constructed value, when the topic's value type is a scalar type.
```
bernardnormier: Suggestion: shorter sentences, drop "in DataStorm" (implied by the section), and fold the scalar-type condition into "could".

```suggestion
- Fixed the handling of partial updates for keys with no value. A key has no value when it was removed, or when no
  full value was written for it yet. Previously, such a partial update crashed applications using class-typed
  values; with other value types, it silently resurrected removed keys. Now a writer throws `std::logic_error`
  when publishing such an update, and a reader discards incoming ones.
- Fixed a bug where calling `getValue()` on a `Remove` sample could return indeterminate data instead of the
  documented default value.
```

bernardnormier (approved):

PR #6085, opened 2026-07-21
Reject malformed hosts and non-IP endpoints in the Glacier2 address filters

bernardnormier:
Follow-up to #6049. When address filters are configured, the Glacier2 router now rejects a proxy outright when any of its endpoints:

- is not IP-based (such as a Bluetooth endpoint), or uses a transport unknown to the router — the address rules describe IP hosts and ports, so an endpoint they cannot interpret is rejected rather than admitted unchecked;
- has an empty host — Ice resolves it to the loopback address, yet it matches no address rule, so a reject rule for loopback would not catch it;
- has a host containing a control character or a space — the resolver stops at an embedded null while the rules see the whole string, and a space can split an injected option (such as ` -p `) out of the rendered endpoint, letting the endpoint the router connects to differ from the one the rules matched.

The `hasDisallowedHost` check introduced by #6049 now reads the host from the typed `IPEndpointInfo` (walking `underlying` for ssl/ws/wss) instead of parsing the endpoint's string representation. That rendering was never meant to carry this check: an empty host emits no `-h` option at all, and an embedded null or space renders ambiguously.

The new static filtering test cases verify that a proxy pairing a legitimate accepted endpoint with a host-less or non-IP endpoint flips to rejected, so such an endpoint cannot ride in behind a legitimate one.

### DIFF
--- a/cpp/src/Glacier2/ProxyVerifier.cpp
+++ b/cpp/src/Glacier2/ProxyVerifier.cpp
@@ -6,6 +6,8 @@
 
 #include <algorithm>
 #include <cctype>
+#include <memory>
+#include <optional>
 #include <stdexcept>
 #include <string>
 #include <vector>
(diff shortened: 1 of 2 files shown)


pepone (approved):
**Approve** ✅

Traced the gate-vs-matcher-vs-router data flow and the new tests — the change is correct and lands at the right depth.

- Sourcing the host from the typed `IPEndpointInfo` (walking `underlying` for ssl/ws/wss) instead of the rendered string is the right call — `getInfo()` is `noexcept` and returns the configured host with no DNS resolution, so it's safe on the dispatch path.
- The new rejections are sound and fail-closed: non-IP / unknown-transport endpoints → `nullopt` → reject; empty, space, and control-char hosts are cut before they can diverge from what the router actually connects to.
- The `>255` and IPv4-numeral / trailing-dot checks are preserved verbatim — strict superset, nothing previously rejected is now admitted.
- Both new static-filtering cases assert the right outcomes: a host-less (`tcp -p 12010`) or opaque (`opaque -t 99 -v ABCD`) endpoint flips an otherwise-accepted proxy to rejected, so it can't ride in behind a legitimate one.

LGTM.

InsertCreativityHere (approved):

PR #6098, opened 2026-07-22
Use DocC symbol links consistently in Swift doc comments

### LINKED ISSUE #6095: Use DocC symbol links consistently in Swift doc comments
The Swift doc comments in `swift/src/Ice` are inconsistent about how they reference other symbols: some use DocC symbol links (``` ``Symbol`` ```), while others use plain code voice (`` `Symbol` ``). Both render as code, but only the double-backtick form resolves to a link on the generated documentation page and is validated by DocC when the documentation is built.

For example, exception names in `- Throws:` lines currently use both styles:

```swift
// Properties.swift
/// - Throws: ``PropertyException`` when the property value is not a valid integer.

// Communicator.swift
/// - Throws: `AlreadyRegisteredException` when a facet with the same name is already registered.
```

The current split in `swift/src/Ice` is roughly 2:1 in favor of single backticks for exception names, while method/type cross-references (for example ``` ``Connection/setCloseCallback(_:)`` ```) mostly use symbol links already.

We should standardize on DocC symbol links (``` ``Symbol`` ```) for every reference to a resolvable symbol — exception types in `- Throws:` clauses, types, and methods — and keep single backticks for non-symbol code fragments such as parameter names, property names like `Ice.Admin.Endpoints`, and literals like `nil`.

bernardnormier:
This PR standardizes symbol references in the Swift doc comments on DocC symbol links (``` ``Symbol`` ```), in two parts:

### Hand-written docs (`swift/src/Ice`)

Every reference to a resolvable symbol now uses a symbol link: exception types in `- Throws:` clauses, types, methods (e.g. ``getAdmin()``, ``locate(_:)``), and enumerators (``ReplyStatus/ok``). Module-qualified prose references such as `Ice.InputStream` become bare links (``InputStream``) since the docs build inside the Ice module.

Single backticks remain for non-symbol code fragments: parameter names, literals (`nil`, `true`), configuration property names (`Ice.Admin.Endpoints`) and their values, type expressions (`Dispatcher?`), symbols DocC cannot resolve (`Sendable`, members of internal types), and `Ice.initialize`, whose three overloads share the `initialize(_:)` name and would make a bare symbol link ambiguous.

### Generated docs (`slice2swift`)

- `writeOpDocSummary` used to emit the `@throws` exception name as plain text (`- Throws: ObjectNotFoundException Thrown when ...`). It now resolves the name — the same lookup the other language mappings use — and formats it with the existing link formatter: a symbol link for same-module references, and the qualified name in monospace for cross-module references (which DocC cannot link).
- The hardcoded `makeProxy`/`checkedCast` doc comments emit ``ParseException``/``LocalException`` links when the generated code is part of the Ice module itself, and keep `Ice.ParseException`/`Ice.LocalException` in monospace elsewhere.

Verified by regenerating code for `Ice/Locator.ice` (same-module links), `IceGrid/Registry.ice` (monospace for Ice-typed helpers), and a synthetic interface with multiple `@throws` tags (linked list items).

Fixes #6095

### DIFF
--- a/cpp/src/slice2swift/Gen.cpp
+++ b/cpp/src/slice2swift/Gen.cpp
@@ -571,7 +571,8 @@ Slice::TypesVisitor::visitStructStart(const StructPtr& p)
     _out << eb;
 
     _out << sp;
-    _out << nl << "/// An `Ice.InputStream` extension to read `" << docName << "` structured values from the stream.";
+    _out << nl << "/// An " << iceDocLink("InputStream", swiftModule) << " extension to read `" << docName
+         << "` structured values from the stream.";
     _out << nl << "public extension " << getUnqualified("Ice.InputStream", swiftModule);
     _out << sb;
 
@@ -622,7 +623,8 @@ Slice::TypesVisitor::visitStructStart(const StructPtr& p)
     _out << eb;
 
     _out << sp;
-    _out << nl << "/// An `Ice.OutputStream` extension to write `" << docName << "` structured values from the stream.";
+    _out << nl << "/// An " << iceDocLink("OutputStream", swiftModule) << " extension to write `" << docName
+         << "` structured values to the stream.";
     _out << nl << "public extension " << getUnqualified("Ice.OutputStream", swiftModule);
     _out << sb;
 
@@ -701,7 +703,8 @@ Slice::TypesVisitor::visitSequence(const SequencePtr& p)
 
     _out << sp;
     _out << nl << "/// Helper class to read and write `" << unescapedName << "` sequence values from";
-    _out << nl << "/// `Ice.InputStream` and `Ice.OutputStream`.";
+    _out << nl << "/// " << iceDocLink("InputStream", swiftModule) << " and " << iceDocLink("OutputStream", swiftModule)
+         << ".";
     _out << nl << "public struct " << unescapedName << "Helper";
     _out << sb;
 
@@ -845,7 +848,8 @@ Slice::TypesVisitor::visitDictionary(const DictionaryPtr& p)
 
     _out << sp;
     _out << nl << "/// Helper class to read and write `" << unescapedName << "` dictionary values from";
-    _out << nl << "/// `Ice.InputStream` and `Ice.OutputStream`.";
+    _out << nl << "/// " << iceDocLink("InputStream", swiftModule) << " and " << iceDocLink("OutputStream", swiftModule)
+         << ".";
     _out << nl << "public struct " << unescapedName << "Helper";
     _out << sb;
 
@@ -1002,7 +1006,8 @@ Slice::TypesVisitor::visitEnum(const EnumPtr& p)
     _out << eb;
 
     _out << sp;
-    _out << nl << "/// An `Ice.InputStream` extension to read `" << docName << "` enumerated values from the stream.";
+    _out << nl << "/// An " << iceDocLink("InputStream", swiftModule) << " extension to read `" << docName
+         << "` enumerated values from the stream.";
     _out << nl << "public extension " << getUnqualified("Ice.InputStream", swiftModule);
     _out << sb;
 
@@ -1037,7 +1042,8 @@ Slice::TypesVisitor::visitEnum(const EnumPtr& p)
     _out << eb;
 
     _out << sp;
-    _out << nl << "/// An `Ice.OutputStream` extension to write `" << docName << "` enumerated values to the stream.";
+    _out << nl << "/// An " << iceDocLink("OutputStream", swiftModule) << " extension to write `" << docName
+         << "` enumerated values to the stream.";
     _out << nl << "public extension " << getUnqualified("Ice.OutputStream", swiftModule);
     _out << sb;
 
@@ -1167,8 +1173,7 @@ Slice::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     _out << nl << "///    - proxyString: The proxy string to parse.";
     _out << nl << "///    - type: The type of the new proxy.";
     _out << nl << "/// - Returns: A new proxy with the requested type.";
-    _out << nl << "/// - Throws: " << (swiftModule == "Ice" ? "``ParseException``" : "`Ice.ParseException`")
-         << " if the proxy string is invalid.";
+    _out << nl << "/// - Throws: " << iceDocLink("ParseException", swiftModule) << " if the proxy string is invalid.";
     _out << nl << "public func makeProxy(communicator: Ice.Communicator, proxyString: String, type: " << prx
          << ".Protocol) throws -> " << prx;
     _out << sb;
@@ -1193,8 +1198,7 @@ Slice::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     _out << nl << "///   - context: The optional context dictionary for the remote invocation.";
     _out << nl
          << "/// - Returns: A proxy with the requested type or nil if the target object does not support this type.";
-    _out << nl << "/// - Throws: " << (swiftModule == "Ice" ? "``LocalException``" : "`Ice.LocalException`")
-         << " if a communication error occurs.";
+    _out << nl << "/// - Throws: " << iceDocLink("LocalException", swiftModule) << " if a communication error occurs.";
     _out << nl << "public func checkedCast" << spar << ("prx: " + getUnqualified("Ice.ObjectPrx", swiftModule))
          << ("type: " + prx + ".Protocol") << ("facet: Swift.String? = nil")
          << ("context: " + getUnqualified("Ice.Context", swiftModule) + "? = nil") << epar << " async throws -> " << prx
@@ -1238,7 +1242,8 @@ Slice::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     // InputStream extension
     //
     _out << sp;
-    _out << nl << "/// Extension to `Ice.InputStream` class to support reading proxies of type";
+    _out << nl << "/// Extension to " << iceDocLink("InputStream", swiftModule)
+         << " class to support reading proxies of type";
     _out << nl << "/// `" << prx << "`.";
     _out << nl << "public extension " << getUnqualified("Ice.InputStream", swiftModule);
     _out << sb;
--- a/cpp/src/slice2swift/Gen.cpp
+++ b/cpp/src/slice2swift/Gen.cpp
@@ -1167,7 +1167,8 @@ Slice::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     _out << nl << "///    - proxyString: The proxy string to parse.";
     _out << nl << "///    - type: The type of the new proxy.";
     _out << nl << "/// - Returns: A new proxy with the requested type.";
-    _out << nl << "/// - Throws: `Ice.ParseException` if the proxy string is invalid.";
+    _out << nl << "/// - Throws: " << (swiftModule == "Ice" ? "``ParseException``" : "`Ice.ParseException`")
+         << " if the proxy string is invalid.";
     _out << nl << "public func makeProxy(communicator: Ice.Communicator, proxyString: String, type: " << prx
          << ".Protocol) throws -> " << prx;
     _out << sb;
@@ -1192,7 +1193,8 @@ Slice::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     _out << nl << "///   - context: The optional context dictionary for the remote invocation.";
     _out << nl
          << "/// - Returns: A proxy with the requested type or nil if the target object does not support this type.";
-    _out << nl << "/// - Throws: `Ice.LocalException` if a communication error occurs.";
+    _out << nl << "/// - Throws: " << (swiftModule == "Ice" ? "``LocalException``" : "`Ice.LocalException`")
+         << " if a communication error occurs.";
     _out << nl << "public func checkedCast" << spar << ("prx: " + getUnqualified("Ice.ObjectPrx", swiftModule))
          << ("type: " + prx + ".Protocol") << ("facet: Swift.String? = nil")
          << ("context: " + getUnqualified("Ice.Context", swiftModule) + "? = nil") << epar << " async throws -> " << prx
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -280,17 +280,28 @@ Slice::Swift::writeOpDocSummary(IceInternal::Output& out, const OperationPtr& p,
     if (!docExceptions.empty())
     {
         out << nl << "/// - Throws:";
-        for (const auto& docException : docExceptions)
+        for (const auto& [name, lines] : docExceptions)
         {
             if (docExceptions.size() > 1)
             {
                 out << nl << "///   -";
             }
-            out << " " << docException.first;
-            if (!docException.second.empty())
+
+            // Try to locate the exception's definition using the name given in the comment, so we can reference it
+            // with a symbol link.
+            if (ExceptionPtr ex = p->container()->lookupException(name, false))
+            {
+                out << " " << swiftLinkFormatter(name, p, ex);
+            }
+            else
+            {
+                out << " `" << name << "`";
+            }
+
+            if (!lines.empty())
             {
                 out << " ";
-                writeDocLines(out, docException.second, false, "     ");
+                writeDocLines(out, lines, false, "     ");
             }
         }
     }
--- a/cpp/src/slice2swift/SwiftUtil.cpp
+++ b/cpp/src/slice2swift/SwiftUtil.cpp
@@ -378,6 +378,14 @@ Slice::Swift::swiftLinkFormatter(const string& rawLink, const ContainedPtr& sour
     }
 }
 
+string
+Slice::Swift::iceDocLink(const string& name, const string& swiftModule)
+{
+    // DocC does not support cross-module linking, so we can only emit a symbol link when the generated code is part
+    // of the Ice module itself.
+    return swiftModule == "Ice" ? "``" + name + "``" : "`Ice." + name + "`";
+}
+
 void
 Slice::Swift::validateSwiftMetadata(const UnitPtr& unit)
 {
--- a/swift/src/Ice/Communicator.swift
+++ b/swift/src/Ice/Communicator.swift
@@ -11,7 +11,7 @@ import Foundation
 /// You create a communicator with `Ice.initialize`, and it's usually the first object you create when programming
 /// with Ice. You can create multiple communicators in a single program, but this is not common.
 public protocol Communicator: AnyObject, Sendable {
-    /// Destroys this communicator. This method calls ``shutdown()`` implicitly. Calling `destroy` destroys all
+    /// Destroys this communicator. This method calls ``shutdown()`` implicitly. Calling ``destroy()`` destroys all
     /// object adapters and closes all outgoing connections. This method waits for all outstanding dispatches to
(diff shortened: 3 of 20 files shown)


externl (approved):

swift/src/Ice/Object.swift
```diff
@@ -83,7 +83,7 @@ public struct ObjectTraits: SliceTraits {
     public static let staticId = "::Ice::Object"
 }
 
-/// `DefaultObject` provides a default implementation of the 4 ``Object`` operations
+/// ``DefaultObject`` provides a default implementation of the 4 ``Object`` operations
 /// (`ice_id`, `ice_ids`, `ice_isA`, `ice_ping`) for a given Slice interface.
```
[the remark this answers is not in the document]
bernardnormier: This is on purpose. We're talking about the Slice operations here.

cpp/src/slice2swift/Gen.cpp
```diff
@@ -1167,7 +1167,8 @@ Slice::TypesVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
     _out << nl << "///    - proxyString: The proxy string to parse.";
```
InsertCreativityHere, line 1173: These look fine, but should we do the same to the other ``/// ... `Ice.xxx`` comments in this file?
Because these 2 are not the only ones. Although I think they were the most important.

swift/src/Ice/Communicator.swift
```diff
@@ -11,7 +11,7 @@ import Foundation
 /// You create a communicator with `Ice.initialize`, and it's usually the first object you create when programming
 /// with Ice. You can create multiple communicators in a single program, but this is not common.
 public protocol Communicator: AnyObject, Sendable {
-    /// Destroys this communicator. This method calls ``shutdown()`` implicitly. Calling `destroy` destroys all
+    /// Destroys this communicator. This method calls ``shutdown()`` implicitly. Calling ``destroy()`` destroys all
```
InsertCreativityHere: I think the single-backticks are more idiomatic here, because in this case `destroy()` _is the function we're documenting_. Not another function we're linking to.

I asked Claude with 0 priming, and it agreed that self-linking is unidiomatic (but will compile ofc).
Suggesting we either use "code-voice" (what it calls the single backticks), or just say "this method" instead.

```suggestion
    /// Destroys this communicator. This method calls ``shutdown()`` implicitly. Calling `destroy()` destroys all
```

swift/src/Ice/Communicator.swift
```diff
@@ -164,11 +164,12 @@ public protocol Communicator: AnyObject, Sendable {
...
     /// - Parameter compress: Specifies whether or not the queued batch requests should be compressed before being
     ///   sent over the wire.
-    /// - Throws: `CommunicatorDestroyedException` when the communicator has been destroyed.
+    /// - Throws: ``CommunicatorDestroyedException`` when the communicator has been destroyed.
     func flushBatchRequests(_ compress: CompressBatch) async throws
 
     /// Adds the Admin object with all its facets to the provided object adapter. If `Ice.Admin.ServerId`
-    /// is set and the provided object adapter has a Locator, `createAdmin` registers the Admin's Process facet with
+    /// is set and the provided object adapter has a Locator, ``createAdmin(adminAdapter:adminId:)`` registers the
+    /// Admin's Process facet with
```
InsertCreativityHere: Same here, `createAdmin` is the function we're documenting. Having it link to itself is pointless.
We should either keep single backticks or do the Apple-style of saying "this method" instead.

swift/src/Ice/Communicator.swift
```diff
@@ -177,18 +178,18 @@ public protocol Communicator: AnyObject, Sendable {
     ///   - adminId: The identity of the Admin object.
     /// - Returns: A proxy to the main ("") facet of the Admin object.
     /// - Throws:
-    ///   - `InitializationException` when `createAdmin` is called more than once.
-    ///   - `CommunicatorDestroyedException` when the communicator has been destroyed.
+    ///   - ``InitializationException`` when ``createAdmin(adminAdapter:adminId:)`` is called more than once.
```
InsertCreativityHere: And same here. Pointless self-reference.

swift/src/Ice/Connection.swift
```diff
@@ -18,7 +18,7 @@ public enum CompressBatch: UInt8 {
     }
 }
 
-/// An `Ice.InputStream` extension to read `CompressBatch` enumerated values from the stream.
```
InsertCreativityHere, line 21: This change looks good.

But the fact that we _do_ properly link the `Ice.InputStream` here means we should also be doing this in `Gen.cpp`, instead of right now where all these `Ice.InputStream` links were untouched, and for some reason, only the exception links got this treatment.

swift/src/Ice/CurrentExtensions.swift
```diff
@@ -17,7 +17,7 @@ extension Current {
         }
     }
 
-    /// Creates an OutgoingResponse with reply status `ok`.
+    /// Creates an OutgoingResponse with reply status ``ReplyStatus/ok``.
```
InsertCreativityHere: This is redundantly worded. The generated docs include the enum name, so this is going to say:
`... with reply status ReplyStatus.ok`. We should just drop the leading reply status...

```suggestion
    /// Creates an OutgoingResponse with ``ReplyStatus/ok``.
```

----

For example, this the rendered page for linking `OperationMode.idempotent`:
Image

swift/src/Ice/CurrentExtensions.swift
```diff
@@ -37,7 +37,7 @@ extension Current {
         return OutgoingResponse(ostr)
     }
 
-    /// Creates an empty OutgoingResponse with reply status `ok` and an empty payload.
+    /// Creates an empty OutgoingResponse with reply status ``ReplyStatus/ok`` and an empty payload.
```
InsertCreativityHere: Same here:
```suggestion
    /// Creates an empty OutgoingResponse with ``ReplyStatus/ok`` and an empty payload.
```

swift/src/Ice/CurrentExtensions.swift
```diff
@@ -48,10 +48,12 @@ extension Current {
         return OutgoingResponse(ostr)
     }
 
-    /// Creates an OutgoingResponse with the specified payload with reply status `ok` or `user exception`.
+    /// Creates an OutgoingResponse with the specified payload with reply status ``ReplyStatus/ok`` or
```
InsertCreativityHere: Same:
```suggestion
    /// Creates an OutgoingResponse with the specified payload with ``ReplyStatus/ok`` or
```

swift/src/Ice/InputStream.swift
```diff
@@ -178,8 +178,8 @@ public final class InputStream {
     }
 
     /// Indicates that unmarshaling is complete, except for any class instances. The application must call this method
-    /// only if the stream actually contains class instances. With the 1.0 encoding, calling `readPendingValues`
-    /// triggers the calls to consumers provided to the class-read methods (`read(cb:)` and `read(_:cb:)`) to inform
+    /// only if the stream actually contains class instances. With the 1.0 encoding, calling ``readPendingValues()``
```
InsertCreativityHere: Same note here about a self-link for `readPendingValues`.

swift/src/Ice/OutgoingResponse.swift
```diff
@@ -5,12 +5,13 @@ import Foundation
 /// Represents the response to an incoming request. It's returned by ``Dispatcher/dispatch(_:)``.
 public struct OutgoingResponse {
     /// The exception ID of the response.
-    /// It's nil when ``replyStatus`` is `ok` or `userException`. Otherwise, this ID is the value returned by `ice_id` for
-    /// Ice local exceptions. For other errors, this ID is the name of the error's type.
+    /// It's nil when ``replyStatus`` is ``ReplyStatus/ok`` or ``ReplyStatus/userException``. Otherwise, this ID is
```
InsertCreativityHere: Should put these in code-formatting too to be consistent...

```suggestion
    /// It's `nil` when ``replyStatus`` is ``ReplyStatus/ok`` or ``ReplyStatus/userException``. Otherwise, this ID is
```

swift/src/Ice/OutgoingResponse.swift
```diff
@@ -5,12 +5,13 @@ import Foundation
...
+    /// the value returned by ``LocalException/ice_id()`` for Ice local exceptions. For other errors, this ID is the
+    /// name of the error's type.
     public let exceptionId: String?
 
     /// The full details of the exception marshaled into the response.
-    /// It's nil when ``replyStatus`` is `ok` or `userException`.
+    /// It's nil when ``replyStatus`` is ``ReplyStatus/ok`` or ``ReplyStatus/userException``.
```
InsertCreativityHere: Same:

```suggestion
    /// It's `nil` when ``replyStatus`` is ``ReplyStatus/ok`` or ``ReplyStatus/userException``.
```

swift/src/Ice/ServantLocator.swift
```diff
@@ -12,8 +12,8 @@ public protocol ServantLocator: Sendable {
     ///
     /// - Remark: The caller (the object adapter) does not insert the returned servant into its Active Servant Map.
     ///   This must be done by the servant locator implementation, if this is desired.
-    /// - Remark: If you call `locate` from your own code, you must also call ``finished(curr:servant:cookie:)``
-    ///   when you have finished using the servant, provided that `locate` returned a non-nil servant.
+    /// - Remark: If you call ``locate(_:)`` from your own code, you must also call ``finished(curr:servant:cookie:)``
+    ///   when you have finished using the servant, provided that ``locate(_:)`` returned a non-nil servant.
```
InsertCreativityHere: Same note about the `locate` self-reference.

swift/src/Ice/ServantLocator.swift
```diff
@@ -31,8 +31,9 @@ public protocol ServantLocator: Sendable {
     ///   - servant: The servant that was returned by ``locate(_:)``.
     ///   - cookie: The cookie that was returned by ``locate(_:)``.
     /// - Throws: The implementation can throw any exception, including ``UserException``.
-    ///   The Ice runtime will marshal this exception in the response. If both the dispatch and `finished` throw
-    ///   an exception, the exception thrown by `finished` prevails and is marshaled back to the client.
+    ///   The Ice runtime will marshal this exception in the response. If both the dispatch and
+    ///   ``finished(curr:servant:cookie:)`` throw an exception, the exception thrown by
+    ///   ``finished(curr:servant:cookie:)`` prevails and is marshaled back to the client.
```
InsertCreativityHere: Same note about the `finished` self-reference.

InsertCreativityHere (commented):
Always good to see more proper links!

But there are some leftover `Ice.InputStream` and `Ice.OutputStream` in the generated code which weren't given the same treatment.

And this PR inserts alot of self-links: a function's doc-comment linking to itself.
My gut said this is non-idiomatic, and after having my Claude review the DocC guidance and usage on GitHub, it agreed. Saying these should either use code-formatting (single back-ticks) or just say "this method". idiomaticity aside, these links are literally pointless.

swift/src/Ice/CurrentExtensions.swift
bernardnormier: I'm not sure this is redundant — "reply status" names the `OutgoingResponse.replyStatus` property being set, while the link names its value; they only echo each other because the type is named after the property. Apple's own docs use this exact structure, e.g. Date.FormatStyle: "The default date style is ``DateStyle/abbreviated``", which renders as "The default date style is Date.FormatStyle.DateStyle.abbreviated". And since DocC symbol links don't support custom link text, the alternative is losing either the link or the property reference. Your rendered example reads fine to me too: "the request's operation mode is OperationMode.idempotent".

cpp/src/slice2swift/Gen.cpp
bernardnormier: Done — I added an `iceDocLink` helper to SwiftUtil and switched all the `Ice.xxx` doc strings in this file over to it: these two exceptions plus the seven `Ice.InputStream`/`Ice.OutputStream` references (struct/enum stream extensions, sequence/dictionary helpers, and the proxy InputStream extension).

swift/src/Ice/Connection.swift
bernardnormier: Done — Gen.cpp now applies the same treatment to all its `Ice.InputStream`/`Ice.OutputStream` doc strings: a symbol link when the generated code is part of the Ice module itself, and monospace `Ice.Xxx` otherwise.

swift/src/Ice/OutgoingResponse.swift
bernardnormier: Fixed.

swift/src/Ice/OutgoingResponse.swift
bernardnormier: Fixed.

swift/src/Ice/Communicator.swift
bernardnormier: Fixed.

swift/src/Ice/Communicator.swift
bernardnormier: Fixed — now "this method" (and I gave the same treatment to the getAdmin self-references below, which weren't flagged).

swift/src/Ice/Communicator.swift
bernardnormier: Fixed.

swift/src/Ice/InputStream.swift
bernardnormier: Fixed — went with code voice here since this comment already says "this method" twice.

swift/src/Ice/ServantLocator.swift
bernardnormier: Fixed — code voice for the `locate(_:)` self-references; the cross-references to `finished` remain symbol links.

swift/src/Ice/ServantLocator.swift
bernardnormier: Fixed — now "this method".

swift/src/Ice/CurrentExtensions.swift
InsertCreativityHere: I think I should of used more precise words. I think that either:
1) We should get rid of the _directly redundant_ words and say:
```` Creates an OutgoingResponse with ``ReplyStatus/ok`` ````
2) Or we should expand it into a real sentence and say:
```` Creates an OutgoingResponse with a reply status of ``ReplyStatus/ok`` ````

I have no issue with using "reply status" and "ReplyStatus/ok" in the same sentence!
I just think that the current wording is weird. That's all :)

Note that the Apple example is option 2.
They don't say "The default date style DateStyle/abbreviated", which is what this comment does.
We also do option 2 throughout our code already too. A little further down in this file we say:
> /// Creates an OutgoingResponse with a reply status other than ``ReplyStatus/ok``.

And I had no comment on that because it's totally fine! Same for the `OperationMode` wording. It was just an example of the output. Not as an example of something I thought was bad. I should of been a little more clear about that...

File: cpp/src/slice2swift/SwiftUtil.cpp [hunk 15]
```diff
@@ -280,17 +280,28 @@ Slice::Swift::writeOpDocSummary(IceInternal::Output& out, const OperationPtr& p,
...
                 out << nl << "///   -";
             }
-            out << " " << docException.first;
-            if (!docException.second.empty())
+
+            // Try to locate the exception's definition using the name given in the comment, so we can reference it
+            // with a symbol link.
+            if (ExceptionPtr ex = p->container()->lookupException(name, false))
+            {
+                out << " " << swiftLinkFormatter(name, p, ex);
```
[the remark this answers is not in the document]
bernardnormier: No changelog fragment for this type of fix.

InsertCreativityHere (approved):
Looks good to me!

Replied to the specific `ReplyStatus` comment, which I still think could be improved.
I should of been clear that my issue was with the wording sounding off,
not with the link having the same name in as the field!

swift/src/Ice/CommunicatorI.swift
```diff
@@ -289,8 +289,8 @@ extension Communicator {
     /// already been initialized.
     ///
     /// - Throws:
-    ///   - `InitializationException` if the plug-ins have already been initialized.
-    ///   - `CommunicatorDestroyedException` if the communicator has been destroyed.
+    ///   - ``InitializationException`` if the plug-ins have already been initialized.
```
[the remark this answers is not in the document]
bernardnormier: This is a self-reference: it's `initializePlugins()`'s own doc comment. Per the review discussion, self-references use code voice, not symbol links.

swift/src/Ice/ServantLocator.swift
```diff
@@ -12,8 +12,8 @@ public protocol ServantLocator: Sendable {
     ///
     /// - Remark: The caller (the object adapter) does not insert the returned servant into its Active Servant Map.
     ///   This must be done by the servant locator implementation, if this is desired.
-    /// - Remark: If you call `locate` from your own code, you must also call ``finished(curr:servant:cookie:)``
-    ///   when you have finished using the servant, provided that `locate` returned a non-nil servant.
+    /// - Remark: If you call `locate(_:)` from your own code, you must also call ``finished(curr:servant:cookie:)``
+    ///   when you have finished using the servant, provided that `locate(_:)` returned a non-nil servant.
```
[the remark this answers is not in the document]
bernardnormier: These are self-references within `locate(_:)`'s own doc comment — code voice is intentional per the review discussion; only cross-references get symbol links.

swift/src/Ice/InputStream.swift
```diff
@@ -178,8 +178,8 @@ public final class InputStream {
     }
 
     /// Indicates that unmarshaling is complete, except for any class instances. The application must call this method
-    /// only if the stream actually contains class instances. With the 1.0 encoding, calling `readPendingValues`
-    /// triggers the calls to consumers provided to the class-read methods (`read(cb:)` and `read(_:cb:)`) to inform
+    /// only if the stream actually contains class instances. With the 1.0 encoding, calling `readPendingValues()`
```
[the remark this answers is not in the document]
bernardnormier: Self-reference — this is `readPendingValues()`'s own doc comment, so code voice is intentional per the review discussion.

swift/src/Ice/Communicator.swift
```diff
@@ -11,7 +11,7 @@ import Foundation
 /// You create a communicator with `Ice.initialize`, and it's usually the first object you create when programming
 /// with Ice. You can create multiple communicators in a single program, but this is not common.
 public protocol Communicator: AnyObject, Sendable {
-    /// Destroys this communicator. This method calls ``shutdown()`` implicitly. Calling `destroy` destroys all
+    /// Destroys this communicator. This method calls ``shutdown()`` implicitly. Calling `destroy()` destroys all
```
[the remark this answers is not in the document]
bernardnormier: These are self-references — `destroy()` in its own doc comment, and likewise `getAdmin()` at line 188. Code voice is intentional per the review discussion; only cross-references get symbol links.

swift/src/Ice/CurrentExtensions.swift
bernardnormier: Reworded per your option 2 — this one now reads "with a reply status of ``ReplyStatus/ok``".

swift/src/Ice/CurrentExtensions.swift
bernardnormier: Fixed.

swift/src/Ice/CurrentExtensions.swift
bernardnormier: Fixed — this one reads "with the specified payload and a reply status of ``ReplyStatus/ok`` or ``ReplyStatus/userException``" to avoid the double "with".

PR #6099, opened 2026-07-22
Document exceptions with an Exceptions section in hand-written MATLAB doc comments

### LINKED ISSUE #6096: Document exceptions with an Exceptions section in hand-written MATLAB doc comments
`slice2matlab` documents the exceptions an operation can throw with a dedicated `Exceptions` section at the end of the help block, after `Input Arguments`/`Output Arguments`:

```matlab
%   Exceptions
%     Ice.SomeException - <description>
```

The hand-written MATLAB code in `matlab/lib` doesn't follow this convention consistently. `Ice.Communicator` and `Ice.Properties` used to document exceptions with a "Throws an Ice.X if ..." sentence in the description block instead (now converted to `Exceptions` sections). The remaining hand-written classes either mention exceptions in prose or don't document them at all. For example:

```matlab
% Connection.m
%GETINFO Returns the connection information. If the connection is closed, this method throws the exception
%   that caused the closure.

% Future.m
%FETCHOUTPUTS Blocks until the invocation completes and then returns the results or throws an exception.
```

We should review the doc comments of the hand-written classes in `matlab/lib` (`Ice.Connection`, `Ice.Future`, `Ice.ObjectPrx`, `Ice.ObjectAdapter`, ...) and document the exceptions they throw with an `Exceptions` section, matching the generated code.

bernardnormier:
This PR documents the exceptions thrown by the hand-written MATLAB classes with an `Exceptions` section, matching the convention used by the generated code and by the already-converted `Ice.Communicator`/`Ice.Properties`:

- `Ice.Connection` — `getInfo` (the closure exception, as `Ice.LocalException`) and `throwException` (the three close/abort exceptions), both converted from prose.
- `Ice.ObjectPrx` — the constructor (`Ice.ParseException`, `Ice.CommunicatorDestroyedException`).
- `Ice.stringToIdentity` — `Ice.ParseException`.
- `Ice.Communicator` — adds the missing `Ice.ParseException` to `propertyToProxy`, found by checking each `Exceptions` section against the corresponding C++ function.

Each entry was verified against the implementation or the documented exceptions of the same operation in the other language mappings. `InputStream`/`OutputStream` methods have no per-method help blocks, so documenting their exceptions is out of scope for this sweep.

Fixes #6096

### DIFF
--- a/matlab/lib/+Ice/Communicator.m
+++ b/matlab/lib/+Ice/Communicator.m
@@ -179,6 +179,7 @@ classdef Communicator < IceInternal.WrapperObject
             %       Ice.ObjectPrx scalar | empty array of Ice.ObjectPrx
             %
             %   Exceptions
+            %     Ice.ParseException - If the property value is not a valid proxy string.
             %     Ice.CommunicatorDestroyedException - If the communicator has been destroyed.
 
             arguments
--- a/matlab/lib/+Ice/Connection.m
+++ b/matlab/lib/+Ice/Connection.m
@@ -193,12 +193,15 @@ classdef Connection < IceInternal.WrapperObject
         end
 
         function r = getInfo(obj)
-            %GETINFO Returns the connection information. If the connection is closed, this method throws the exception
-            %   that caused the closure.
+            %GETINFO Returns the connection information.
             %
             %   Output Arguments
             %     r - The connection information.
             %       Ice.ConnectionInfo scalar
+            %
+            %   Exceptions
+            %     Ice.LocalException - If the connection is closed. This method throws the exception that caused the
+            %       closure.
 
             arguments
                 obj (1, 1) Ice.Connection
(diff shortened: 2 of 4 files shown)


pepone (approved):
The five new entries all check out — I traced each one through the MATLAB glue and the C++ core, and the section format matches slice2matlab's output exactly.

One thing to consider: `Fixes #6096` auto-closes the issue, but `Ice.Future.fetchOutputs` — the issue's other verbatim example — is skipped with no exclusion note (`Future.m` still has the prose-only help block). Either convert it here too, or reword to "Part of #6096" so the issue stays open.

Minor, optional: `throwException` now reads as a closed three-type list while `getInfo` — which rethrows the same stored closure exception — uses the open-ended `Ice.LocalException`; both can also surface e.g. `Ice.ConnectTimeoutException` or `Ice.SocketException`. This matches the C++/C# docs though, so fine as-is. (Side note: the C++ `throwException` doc omits `ConnectionClosedException` — the MATLAB list here is the more complete one.)

externl (approved):

bernardnormier:
> One thing to consider: `Fixes #6096` auto-closes the issue, but `Ice.Future.fetchOutputs` — the issue's other verbatim example — is skipped with no exclusion note (`Future.m` still has the prose-only help block). Either convert it here too, or reword to "Part of #6096" so the issue stays open.

I did not update `fetchOutputs` on purpose - it's not an oversight. An exception block with Ice.Exception (meaning any Ice.Exception) is more confusing than helpful.

PR #6100, opened 2026-07-22
Document ParseException in the propertyToProxy doc comments

bernardnormier:
`Communicator::propertyToProxy` parses the property value as a stringified proxy and throws `ParseException` when it's invalid — the same path as `stringToProxy` — but only the CommunicatorDestroyedException was documented. This PR adds the missing `@throws`/`<exception>`/`Raises` entry in C++, C#, Java, and Python, using each mapping's `stringToProxy` phrasing.

The JS and Swift doc comments (and MATLAB, as part of #6099) already document it.

### DIFF
--- a/cpp/include/Ice/Communicator.h
+++ b/cpp/include/Ice/Communicator.h
@@ -107,6 +107,7 @@ namespace Ice
         /// @tparam Prx The type of the proxy to return.
         /// @param property The base property name.
         /// @return The proxy, or nullopt if the property is not set.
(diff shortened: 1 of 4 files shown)


pepone (approved):
LGTM.

One note: the throws contract is still incomplete — unlike `stringToProxy`, `propertyToProxy` also runs proxy sub-property validation and can throw `PropertyException` (e.g. `MyProxy.Bogus=1` or `MyProxy.InvocationTimeout=abc`), in all four mappings. Fine as a follow-up, no need to hold this PR.

Issue #6068, opened 2026-07-20
DataStorm misroutes initialization when a node hosts multiple same-name topics reading one writer


pepone:
## Summary

When a node hosts two or more `Topic` instances with the **same name** that read from the same remote writer,
DataStorm misroutes reader initialization: samples are delivered to the wrong reader (legacy path) or a reader's
`getNextUnread()` blocks (epoch-1 path in #6067). The root cause is that reader element ids are only unique within a
`TopicI` instance, so two same-name reader topics assign colliding ids.

This is not an exotic pattern. DataStorm topics are value objects, and independent components of an application are
expected to create their own `Topic` instance for a shared name. The write direction is unaffected — two writer topics
get distinct topic ids, so a reader sees them as distinct `(remoteTopic, writer)` pairs — but the read analog (two
components each reading the same topic) triggers the bug.

## Reproduction

On one node, create two same-name topics, each with a single-key reader on a different key, both reading a single
multi-key writer on a second node:

```cpp
Topic<string, string> topicA(node, "twoSameName");
Topic<string, string> topicB(node, "twoSameName");

// (writer side, other node)
auto writer = makeMultiKeyWriter(topic, {"elemA", "elemB"}, "", config);
writer.add("elemA", "valueA");
writer.add("elemB", "valueB");

// reader side: make each real reader the *second* reader of its topic, so both get element id 2
auto readerA = makeSingleKeyReader(topicA, "elemA", "", config); // element id 2 in topicA
auto readerB = makeSingleKeyReader(topicB, "elemB", "", config); // element id 2 in topicB

readerB.getNextUnread(); // wrong value on the legacy path; blocks on the epoch-1 path
```

## Diagnosis

Session trace (`DataStorm.Trace.Session=3`, `DataStorm.Trace.Data=3`) shows the ids colliding across the two topic
instances:

```
initializing reader 'e2' from 'e1' on 'e2:[k5:elemA]@20:twoSameName'
initializing reader 'e2' from 'e1' on 'e2:[k5:elemB]@21:twoSameName'
```

Both readers are `e2`, both keys are `k5`; only the local topic id (`@20` vs `@21`) differs. `TopicI::_nextId` starts
at 0 for each topic instance
([TopicI.h](https://github.com/zeroc-ice/ice/blob/main/cpp/src/DataStorm/TopicI.h)), so the second reader of each topic
is element id 2. Initialization is addressed by the reader element id (`data.id` → `readerId`), and the receiver's
[...]

PR #6138, opened 2026-07-23
Address DataStorm reader initialization to the exact reader element

### LINKED ISSUE #5819: [Audit-High] DataStorm: per-key init-sample batches silently dropped (flow 1) or applied to the wrong reader (flow 2)

### LINKED ISSUE #5881: DataStorm: make the initialization sample stream addressable per reader element
> Follow-up to #5819 / PR #5880.

## Summary

A DataStorm initialization batch (`DataSamples`) carries only the **writer element id** — nothing on the wire
identifies which **reader element** (or key) the batch is for, even though the sender computes one batch per key
and per attached reader element with pair-specific content (`lastId` cutoff, config, sample filter).

PR #5880 fixes the resulting sample loss and misrouting receiver-side by routing batches to subscribers by key
intersection. That heuristic is correct for the common cases but cannot be exact:

- **Empty batches are unaddressable.** An empty batch (the writer had no queued samples for that pair) must mark a
  subscriber initialized to enable the live sample path, but the receiver cannot tell *which* subscriber the batch
  was for. A subscriber attached in round N can be marked initialized by another pair's empty batch from round N−1
  that is dispatched late, and its own non-empty batch is then discarded as a stale round — the initialization
  samples are lost (this loss predates PR #5880).
- **Key routing is content-based.** A batch is routed by the keys its samples happen to carry, not by the pair it
  was computed for; two readers overlapping on a key both receive (and filter) each other's batches, with
  pair-specific `lastId` cutoffs applied to the wrong reader in edge cases.

## Proposal (protocol change → 3.9.0)

Add a peer (reader) element id to `DataSamples`, mirroring `ElementDataAck.peerId`:

```slice
struct DataSamples
{
    long id;
    long peerId;   // NEW: the reader element this batch initializes
    DataSampleSeq samples;
}
```

The receiver then dispatches each batch exactly once to its target element — no `initialized` heuristics, no
content-based routing, and empty batches initialize exactly their target. The `initialized`/`initializedNow`
machinery in `SessionI::initSamples` reduces to a per-pair check.

Requires an encoding/protocol version bump or a new operation, hence 3.9.0.

pepone:
Fixes #5819. Closes #5881 (duplicate).

Supersedes #6067, which fixed the same bug while preserving wire compatibility with Ice 3.8.0–3.8.2 through a negotiated protocol epoch. Per the [icerpc-dev discussion](https://github.com/icerpc/icerpc-dev/discussions/5) we decided to break compatibility with 3.8.0–3.8.2 instead — DataStorm is a young product with no known production deployments — which lets us drop all of the epoch machinery in favor of the simple, exact fix.

## The bug

A writer covering several keys, or several readers of one writer, produces one initialization batch per key under the same writer element id, and nothing on the wire identifies the target reader. `SessionI::initSamples` marked *every* attached reader initialized on the first batch and dropped the rest, so a late-joining multi-key, any-key, or filtered reader lost all but one key's initialization samples, and a sibling reader could be initialized with another reader's samples.

## The fix

`DataSamples` now carries a `peerId` identifying the reader element it initializes (as proposed in #5881, mirroring `ElementDataAck.peerId`). The publisher merges the per-key batches sharing a writer–reader pair into one globally ordered, de-duplicated batch, and the receiver initializes exactly the reader named by `peerId`. No new operation, no epoch negotiation, no receiver-side heuristics.

## Breaking change

This changes the DataStorm session wire, so 3.8.3 nodes no longer interoperate with 3.8.0–3.8.2 nodes. To keep incompatible nodes from establishing sessions, the `Lookup` servant identity changes from `DataStorm/Lookup` to `DataStorm/Lookup2`: all cross-node contact bootstraps through the `Lookup` servant, so a node only discovers peers whose `Lookup` servant uses the matching identity. Upgrade all the DataStorm nodes in a deployment together.

## Testing

`test/DataStorm/events` gains late-joining scenarios: multi-key reader, two single-key readers of a multi-key writer, coalesced same-key readers, filtered reader, unwritten-key (empty-batch) reader, interleaved sample ids, and any-key writer. Verified red→green — the new scenarios hang on `origin/main` (a reader waits forever for the dropped samples) and pass with the fix. The full `DataStorm/*` suite passes, including the relay tests.

## Known limitation (pre-existing, tracked separately)

When a single node hosts two `Topic` instances with the *same name* both reading one writer, initialization can still be misrouted because DataStorm reader element ids are only unique within a topic instance, so `peerId` cannot tell the two readers apart. This predates this change and affects the legacy path as well; it is tracked in #6068.

### DIFF
--- a/changelog.d/datastorm/5819.md
+++ b/changelog.d/datastorm/5819.md
@@ -1,0 +1,7 @@
+- Fixed a bug where a late-joining reader of a writer that spans several keys — a multi-key, any-key, or filtered
+  reader — silently lost all but one key's initialization samples, and where such a reader could be initialized with
+  another reader's samples.
+- DataStorm nodes running this release no longer interoperate with nodes running Ice 3.8.0 through 3.8.2. Fixing the
+  reader-initialization bug above required a change to the DataStorm session protocol, and nodes now use the object
+  identity `DataStorm/Lookup2` for discovery so that a node only establishes sessions with peers running Ice 3.8.3 or
+  later. Upgrade all the DataStorm nodes in a deployment together.
--- a/changelog.d/datastorm/5819.md
+++ b/changelog.d/datastorm/5819.md
@@ -1,7 +1,6 @@
-- Fixed a bug where a late-joining reader of a writer that spans several keys — a multi-key, any-key, or filtered
-  reader — silently lost all but one key's initialization samples, and where such a reader could be initialized with
-  another reader's samples.
+- Fixed the initialization of late-joining readers. A reader of a writer that spans several keys — a multi-key,
+  any-key, or filtered reader — silently lost all but one key's initialization samples. Such a reader could also be
+  initialized with another reader's samples.
 - DataStorm nodes running this release no longer interoperate with nodes running Ice 3.8.0 through 3.8.2. Fixing the
-  reader-initialization bug above required a change to the DataStorm session protocol, and nodes now use the object
-  identity `DataStorm/Lookup2` for discovery so that a node only establishes sessions with peers running Ice 3.8.3 or
-  later. Upgrade all the DataStorm nodes in a deployment together.
+  reader-initialization bug above required a change to the DataStorm session protocol, and a node only establishes
+  sessions with peers running Ice 3.8.3 or later. Upgrade all the DataStorm nodes in a deployment together.
--- a/cpp/src/DataStorm/Contract.ice
+++ b/cpp/src/DataStorm/Contract.ice
@@ -61,12 +61,17 @@ module DataStormContract
     /// A queue of {@link DataSample}
     ["cpp:type:std::deque<DataSample>"] sequence<DataSample> DataSampleSeq;
 
-    /// Represents a collection of data samples produced by a specific writer.
+    /// Represents a collection of data samples produced by a specific writer to initialize a specific reader.
     ["cpp:custom-print"]
     struct DataSamples
     {
         /// The unique identifier for the writer.
         long id;
+
+        /// The unique identifier for the reader element these samples initialize.
+        /// The sender produces one `DataSamples` per writer-reader pair, with its samples merged across the reader's
+        /// keys, so the receiver initializes exactly the reader element identified by `peerId`.
+        long peerId;
 
         /// The sequence of samples produced by the writer.
         DataSampleSeq samples;
@@ -460,7 +465,9 @@ module DataStormContract
 
     /// The lookup interface is used by DataStorm nodes to announce their topic readers and writers to other connected
     /// nodes. When multicast is enabled, the lookup interface also broadcasts these announcements.
-    /// Each DataStorm node hosts a lookup servant with the identity `DataStorm/Lookup`.
+    /// Each DataStorm node hosts a lookup servant with the identity `DataStorm/Lookup2`. Ice 3.8.3 changed this
+    /// identity from `DataStorm/Lookup` (used by Ice 3.8.0 through 3.8.2) to prevent nodes running the incompatible
+    /// pre-3.8.3 reader-initialization protocol from establishing sessions with each other.
     interface Lookup
     {
         /// Announce a topic reader.
--- a/cpp/src/DataStorm/Contract.ice
+++ b/cpp/src/DataStorm/Contract.ice
@@ -68,13 +68,11 @@ module DataStormContract
         /// The unique identifier for the writer.
         long id;
 
-        /// The unique identifier for the reader element these samples initialize.
-        /// The sender produces one `DataSamples` per writer-reader pair, with its samples merged across the reader's
-        /// keys, so the receiver initializes exactly the reader element identified by `peerId`.
-        long peerId;
-
         /// The sequence of samples produced by the writer.
         DataSampleSeq samples;
+
+        /// The unique identifier for the reader element these samples initialize.
+        long peerId;
     }
     sequence<DataSamples> DataSamplesSeq;
 
@@ -465,9 +463,7 @@ module DataStormContract
 
     /// The lookup interface is used by DataStorm nodes to announce their topic readers and writers to other connected
     /// nodes. When multicast is enabled, the lookup interface also broadcasts these announcements.
-    /// Each DataStorm node hosts a lookup servant with the identity `DataStorm/Lookup2`. Ice 3.8.3 changed this
-    /// identity from `DataStorm/Lookup` (used by Ice 3.8.0 through 3.8.2) to prevent nodes running the incompatible
-    /// pre-3.8.3 reader-initialization protocol from establishing sessions with each other.
+    /// Each DataStorm node hosts a lookup servant with the same identity (currently, `DataStorm/Lookup2`).
     interface Lookup
     {
         /// Announce a topic reader.
--- a/cpp/src/DataStorm/Instance.h
+++ b/cpp/src/DataStorm/Instance.h
@@ -22,11 +22,7 @@ namespace DataStormI
     class NodeI;
     class CallbackExecutor;
 
-    // The object identity of the Lookup servant. Ice 3.8.3 changed this from "DataStorm/Lookup" (used by Ice
-    // 3.8.0-3.8.2) to segregate nodes speaking the incompatible pre-3.8.3 reader-initialization protocol: a node only
-    // discovers peers whose Lookup servant uses the matching identity, so a 3.8.3 node never establishes a session
-    // with a 3.8.0-3.8.2 node. This identity must be used everywhere the Lookup servant is registered or addressed. See
-    // issue #5819.
+    // The object identity shared by all Lookup servants and proxies.
     inline Ice::Identity lookupIdentity() { return {.name = "Lookup2", .category = "DataStorm"}; }
 
     class Instance final : public std::enable_shared_from_this<Instance>
--- a/cpp/src/DataStorm/Instance.h
+++ b/cpp/src/DataStorm/Instance.h
@@ -25,8 +25,9 @@ namespace DataStormI
     // The object identity of the Lookup servant. Ice 3.8.3 changed this from "DataStorm/Lookup" (used by Ice
     // 3.8.0-3.8.2) to segregate nodes speaking the incompatible pre-3.8.3 reader-initialization protocol: a node only
     // discovers peers whose Lookup servant uses the matching identity, so a 3.8.3 node never establishes a session
-    // with a 3.8.0-3.8.2 node. This constant must be used everywhere the Lookup identity is needed. See issue #5819.
-    inline const Ice::Identity lookupIdentity{.name = "Lookup2", .category = "DataStorm"};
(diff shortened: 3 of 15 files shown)


cpp/src/DataStorm/Contract.ice
```diff
@@ -460,7 +465,9 @@ module DataStormContract
 
     /// The lookup interface is used by DataStorm nodes to announce their topic readers and writers to other connected
     /// nodes. When multicast is enabled, the lookup interface also broadcasts these announcements.
-    /// Each DataStorm node hosts a lookup servant with the identity `DataStorm/Lookup`.
+    /// Each DataStorm node hosts a lookup servant with the identity `DataStorm/Lookup2`. Ice 3.8.3 changed this
+    /// identity from `DataStorm/Lookup` (used by Ice 3.8.0 through 3.8.2) to prevent nodes running the incompatible
+    /// pre-3.8.3 reader-initialization protocol from establishing sessions with each other.
```
bernardnormier: Simplification: this is not the right spot to discuss DataStorm protocol versioning.

```suggestion
    /// Each DataStorm node hosts a lookup servant with the same identity (currently, `DataStorm/Lookup2`).
```

cpp/src/DataStorm/Contract.ice
```diff
@@ -61,13 +61,18 @@ module DataStormContract
...
         /// The unique identifier for the writer.
         long id;
 
+        /// The unique identifier for the reader element these samples initialize.
+        /// The sender produces one `DataSamples` per writer-reader pair, with its samples merged across the reader's
+        /// keys, so the receiver initializes exactly the reader element identified by `peerId`.
+        long peerId;
+
         /// The sequence of samples produced by the writer.
         DataSampleSeq samples;
```
bernardnormier: One-line doc like the other `peerId` fields, and `peerId` moved last to match `ElementSpec`, `ElementDataAck`, and `ElementSpecAck` (local fields first, peer fields last). Since this release breaks the session wire anyway, the reordering is free.

```suggestion
        /// The sequence of samples produced by the writer.
        DataSampleSeq samples;

        /// The unique identifier for the reader element these samples initialize.
        long peerId;
```

cpp/src/DataStorm/Instance.h
```diff
@@ -22,6 +22,13 @@ namespace DataStormI
     class NodeI;
     class CallbackExecutor;
 
+    // The object identity of the Lookup servant. Ice 3.8.3 changed this from "DataStorm/Lookup" (used by Ice
+    // 3.8.0-3.8.2) to segregate nodes speaking the incompatible pre-3.8.3 reader-initialization protocol: a node only
+    // discovers peers whose Lookup servant uses the matching identity, so a 3.8.3 node never establishes a session
+    // with a 3.8.0-3.8.2 node. This identity must be used everywhere the Lookup servant is registered or addressed. See
+    // issue #5819.
```
bernardnormier: Simplification: what matters here is that the identity is shared by all Lookup servants and proxies; the versioning story lives in the changelog.

```suggestion
    // The object identity shared by all Lookup servants and proxies.
```

bernardnormier (commented):
A few small comments below. Also:
- This PR should carry the `datastorm` label.
- Doc follow-up for 3.8.3: https://docs.zeroc.com/ice/3.8/cpp/datastorm-node mentions the lookup identity once, in the `DataStorm.Node.Multicast.Proxy` description. It needs the new identity, and since the identity is now enforced at startup.

We could add a table that shows the identity for Ice 3.8.0-2, and Ice 3.8.3+.

changelog.d/datastorm/5819.md
```diff
@@ -0,0 +1,7 @@
+- Fixed a bug where a late-joining reader of a writer that spans several keys — a multi-key, any-key, or filtered
+  reader — silently lost all but one key's initialization samples, and where such a reader could be initialized with
+  another reader's samples.
+- DataStorm nodes running this release no longer interoperate with nodes running Ice 3.8.0 through 3.8.2. Fixing the
+  reader-initialization bug above required a change to the DataStorm session protocol, and nodes now use the object
+  identity `DataStorm/Lookup2` for discovery so that a node only establishes sessions with peers running Ice 3.8.3 or
+  later. Upgrade all the DataStorm nodes in a deployment together.
```
bernardnormier: Suggestion: split the first bullet into shorter sentences, and drop the object identity from the second bullet — it's an internal detail; the interop consequence carries the message.

```suggestion
- Fixed the initialization of late-joining readers. A reader of a writer that spans several keys — a multi-key,
  any-key, or filtered reader — silently lost all but one key's initialization samples. Such a reader could also be
  initialized with another reader's samples.
- DataStorm nodes running this release no longer interoperate with nodes running Ice 3.8.0 through 3.8.2. Fixing the
  reader-initialization bug above required a change to the DataStorm session protocol, and a node only establishes
  sessions with peers running Ice 3.8.3 or later. Upgrade all the DataStorm nodes in a deployment together.
```

externl (commented):
Traced the attach → ack → `initSamples` path on both sides: the `peerId` routing and the merge hold up. `ElementData.id` is never negated, so the `getId()` match works for keyed, any-key and filtered readers alike, and sample ids come from the per-topic `_nextSampleId`, so the dedup can't drop distinct samples.

Minor: `mergeInitSamples` should use `stable_sort`. `unique` keeps the first of an equal-id run, and `finalize` resolves the earliest sample of each key to a full value *per batch*, so an arbitrary survivor could leave the reader with an unresolvable partial update.

bernardnormier:
> Two things worth addressing:
> 
> * **Mixed-version deployments fail silently.** A 3.8.3↔3.8.2 pair gets `ObjectNotExist`

This is on purpose. 3.8.2 and 3.8.3 are essentially isolated from each other.

cpp/test/DataStorm/events/Reader.cpp
```diff
@@ -458,6 +458,250 @@ void ::Reader::run(int argc, char* argv[])
...
+        auto reader = makeMultiKeyReader(topic, {"elemA", "elemB"}, "", config);
+        reader.waitForUnread(3);
+        map<string, vector<string>> byKey;
+        for (const auto& sample : reader.getAllUnread())
+        {
+            byKey[sample.getKey()].push_back(sample.getValue());
+        }
+        test(byKey.size() == 2);
+        test((byKey["elemA"] == vector<string>{"valueA1", "valueA2"}));
+        test((byKey["elemB"] == vector<string>{"valueB1"}));
```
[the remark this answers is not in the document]
pepone: Done — the test now reads the three samples in sequence and asserts the global sample-id order (A1, B1, A2), so a broken or removed sort is caught. (01c2b838a3)

cpp/test/DataStorm/events/Reader.cpp
```diff
@@ -458,6 +458,250 @@ void ::Reader::run(int argc, char* argv[])
...
+        auto doneWriter = makeSingleKeyWriter(done, "done");
+        doneWriter.waitForReaders();
+        doneWriter.update(0);
+    }
+
+    // Two readers on the same key, created back-to-back so they attach to the writer element in one initialization
+    // round, must each receive the key's initialization sample exactly once. The live update "valueB" is published
+    // only after both attach and drain, so a duplicated initialization sample would surface as a second "valueA"
+    // ahead of it. This depends on the two readers coalescing into one round (very likely with back-to-back creation,
+    // though not guaranteed); when they do not, the case still passes.
```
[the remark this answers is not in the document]
pepone: Reworded to what the test actually guarantees. Two readers on the same key are distinct reader elements (distinct `peerId`), so they merge as separate pairs rather than one batch — the exact-once, correctly-routed initialization the test asserts is deterministic regardless of round coalescing, so I kept it rather than adding a determinism hook. (01c2b838a3)

cpp/src/DataStorm/Contract.ice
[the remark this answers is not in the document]
pepone: Fixed — replaced the comment with a version-agnostic one-liner (per Bernard's suggestion below). (01c2b838a3)

pepone: Applied. (01c2b838a3)

cpp/src/DataStorm/Contract.ice
pepone: Applied — `peerId` moved last with a one-line doc; wire regenerated. (01c2b838a3)

cpp/src/DataStorm/Instance.h
pepone: Applied. (01c2b838a3)

changelog.d/datastorm/5819.md
pepone: Applied. (01c2b838a3)

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -484,84 +533,78 @@ SessionI::initSamples(int64_t topicId, DataSamplesSeq samplesSeq, const Current&
...
+                if (_traceLevels->session > 2)
+                {
+                    Trace out(_traceLevels->logger, _traceLevels->sessionCat);
+                    out << _id << ": initializing samples from 'e" << samples.id << "' on '" << element << "'";
+                }
+
+                elementSubscriber.initialized = true;
+                if (samples.samples.empty())
+                {
+                    return;
```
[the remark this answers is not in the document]
pepone: Good catch — `initialized` is now set after `samplesI` is built (and before the early return for an empty batch), restoring the transactional ordering so a decode/construct failure leaves the reader uninitialized for a redelivery. (8c88432f04)

pepone:
Thanks for the reviews — all inline threads addressed and resolved. Also from the review bodies:

- @externl `mergeInitSamples` now uses `stable_sort`, so `unique` keeps the first (base-resolved) copy of an equal-id run. (01c2b838a3)
- Added the `datastorm` label.
- Doc follow-up (the `DataStorm.Node.Multicast.Proxy` identity, now enforced at startup, plus a per-version table) filed as #6155.

changelog.d/datastorm/5819.md
```diff
@@ -0,0 +1,6 @@
+- Fixed the initialization of late-joining readers. A reader of a writer that spans several keys — a multi-key,
+  any-key, or filtered reader — silently lost all but one key's initialization samples. Such a reader could also be
+  initialized with another reader's samples.
+- DataStorm nodes running this release no longer interoperate with nodes running Ice 3.8.0 through 3.8.2. Fixing the
+  reader-initialization bug above required a change to the DataStorm session protocol, and a node only establishes
+  sessions with peers running Ice 3.8.3 or later. Upgrade all the DataStorm nodes in a deployment together.
```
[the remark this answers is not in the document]
pepone: Valid, but the `DataStorm.Node.Multicast.Proxy` identity migration is documented in the docs follow-up #6155, and a mismatched value fails loudly at startup naming the required identity. Per an earlier review the changelog stays free of the object identity, so this is left to the docs.

bernardnormier (approved):
Looks good.

externl (approved):
Looks good!

Issue #6088, opened 2026-07-22
DataStorm: SessionI subscriber maps key on raw TopicI* — a lingering entry is a use-after-free, not a leak


pepone:
> Structural/hardening follow-up surfaced while fixing #5873. No confirmed live crash on current `main` — after
> #5872/#5873 the known teardown windows are closed — but the session's subscriber bookkeeping keys on raw
> `TopicI*` and its safety rests on an unchecked invariant that has already been violated three times (#5821,
> #5872, #5873). The raw-pointer representation turns any future violation into a use-after-free rather than a
> benign stale entry.

Code links pinned to `origin/main` at [`2c95df8`](https://github.com/zeroc-ice/ice/blob/2c95df83e46157613191bbb6e5357690cd51efab).

## The load-bearing invariant

`SessionI` tracks which local topics subscribe to each remote topic with two maps keyed by **raw** `TopicI*`:

- [`_topics`](https://github.com/zeroc-ice/ice/blob/2c95df83e46157613191bbb6e5357690cd51efab/cpp/src/DataStorm/SessionI.h#L446) — `map<int64_t, TopicSubscribers>` (remote topic id → subscribers), and
- [`TopicSubscribers::_subscribers`](https://github.com/zeroc-ice/ice/blob/2c95df83e46157613191bbb6e5357690cd51efab/cpp/src/DataStorm/SessionI.h#L255) — `map<TopicI*, TopicSubscriber>`.

Neither map owns the topic. A `TopicI` is kept alive only **transitively**, through the element subscriptions:
each `ElementSubscriber` holds a `shared_ptr<DataElementI>` ([`SessionI.h#L118`](https://github.com/zeroc-ice/ice/blob/2c95df83e46157613191bbb6e5357690cd51efab/cpp/src/DataStorm/SessionI.h#L118)), and every `DataElementI` owns
its parent through a `const shared_ptr<TopicI>` ([`DataElementI.h#L328`](https://github.com/zeroc-ice/ice/blob/2c95df83e46157613191bbb6e5357690cd51efab/cpp/src/DataStorm/DataElementI.h#L328)).

So the raw `TopicI*` used as a map key is only valid while this invariant holds:

1. **every `TopicI*` in `_topics` is pinned by at least one live element subscription**, and
2. **the entry is removed from `_topics` before the topic is destroyed and deleted.**

Both halves are fragile.

## Why (1) does not always hold

[...]

Issue #5873, opened 2026-07-08
DataStorm: peer detachTopic and destroyImpl drop a destroyed topic's subscriber entry without detaching its elements


pepone:
> Follow-up to #5821 / PR #5872, surfaced by a Codex cross-check of that fix. Same defect shape, narrower trigger
> windows.

## Summary

PR #5872 fixed the main path: when a **connection loss** races `Topic::destroy`, `SessionI::disconnect` (called by
`TopicI::disconnect`) now detaches the destroyed topic's elements even when `_session` is already null, keyed on
the surviving `_topics` subscriber entry.

Two sibling paths still drop that subscriber entry **without** running the local element detach, so
`TopicI::disconnect` finds nothing to clean and the same leak occurs (stale element listener entries,
`_listenerCount > 0` → debug assert in `TopicI::disconnect`, release `waitForNoWriters()`/`waitForNoReaders()`
hang, pinned session servant):

1. **Peer `detachTopic`** (`SessionI::detachTopic`): `runWithTopics` skips the destroyed topic (so `TopicI::detach`
   → `unsubscribe` never runs), then `_topics.erase(id)` drops the entry wholesale.
2. **`SessionI::destroyImpl`**: same shape — `runWithTopics` skips destroyed topics, then `_topics.clear()`.

## Trigger windows

Both need the erase to land between `TopicI::destroy` marking `_destroyed` and `TopicI::disconnect` acquiring the
session mutex:

- (1) requires the **peer** to destroy its matching topic (sending `detachTopic`) essentially simultaneously with
  the local destroy — both teardown paths in a microsecond window;
- (2) requires the session to be destroyed (node shutdown / session removal) in that window; at shutdown the
  practical impact is mostly the debug asserts, since everything is being torn down anyway.

Much narrower than the connection-loss race fixed by PR #5872, hence split out.

## Fix considerations

A naive "second pass unsubscribing destroyed-topic subscribers" is **not safe**: `isDestroyed()` can flip between
`runWithTopics`' check and the second pass, producing a double `element->detachKey`, and
`DataElementI::Listener::get` dereferences an unchecked `find()` — UB when the listener still has other
subscriptions. Options:

- retain destroyed-topic subscriber entries in `detachTopic`/`destroyImpl` (erase only the non-destroyed ones) and
  let `TopicI::disconnect` → `SessionI::disconnect` clean them, accepting a bounded stale-entry residue in the
  inverse TOCTOU corner; or
[...]

PR #6139, opened 2026-07-23
Draw DataStorm topic element, key, tag, and filter ids from a node-wide counter

### LINKED ISSUE #6068: DataStorm misroutes initialization when a node hosts multiple same-name topics reading one writer

pepone:
Fixes #6068.

A node can host several `Topic` instances with the same name reading the same remote writer (topics are value
objects, so independent components each create their own instance for a shared name). Reader initialization and
sample delivery are addressed by the local element id and resolved by the local key id, and the peer echoes both
back as opaque handles that the receiver resolves against **every** same-name topic (`runWithTopics` fans a
sample/ack over all local topics subscribed to the one remote topic). Element and key ids were numbered
per-`TopicI`, so two same-name topics assigned colliding ids: a reader received the wrong key's value (legacy
path) or blocked (epoch-1 path in #6067).

### Note on the diagnosis

The issue proposed node-scoping **element ids only**, stating "key ids can stay per-topic … only element ids need
to be node-unique." That is not sufficient: the reproduction test confirms **key ids collide the same way**. In
`TopicI::attachElements`, the peer-echoed local key id (`spec.peerId`) is resolved with `_keyFactory->get(peerId)`
against every same-name topic, so a key id valid in one topic resolves to a *different* key in another. Fixing only
element ids leaves the reader subscribed to both writer keys and still misrouted. This PR node-scopes element,
key, tag, and filter ids.

### Fix

- Element ids are drawn from a node-wide counter on `Instance` (`Instance::nextElementId`, used by `TopicI`).
- Key, tag, and filter ids are drawn from a second node-wide counter shared by the per-topic factories
  (`KeyFactoryT`/`TagFactoryT`/`FilterManagerT`), reached from the `Topic` constructor via a private
  `Node::getIdCounter()`. The counter starts at 1, preserving the reserved match-all filter id 1.

All of these are node-local handles the peer only echoes back (cross-node matching is by value), so renumbering
them changes the values, not the protocol: mixed 3.8.x meshes interoperate unchanged and no protocol epoch is
required. Independent of #6067's epoch work, which can land separately.

### Testing

`cpp/test/DataStorm/events` gains a scenario with two same-name reader topics reading one multi-key writer, each
reader on a different key; each must receive only its own key's value. Verified red→green (without the fix the
reader receives the other key's value).

The test uses the live-delivery path rather than a late-joining initialization path on purpose: the late-join
variant additionally trips a pre-existing same-name-topic teardown crash (a use-after-free from the session
subscriber maps keying on raw `TopicI*`, tracked in #6088 and fixed by #5873/#6087, neither on `main` yet). This
fix only changes id values, not teardown, and the full DataStorm suite passes with it.

### DIFF
--- a/changelog.d/datastorm/6068.md
+++ b/changelog.d/datastorm/6068.md
@@ -1,0 +1,5 @@
+- Fixed a bug in DataStorm where a node hosting two or more `Topic` instances with the same name that read from the
+  same remote writer misrouted reader initialization and sample delivery: a reader received the wrong key's value or
+  blocked. Topic element, key, tag, and filter ids are now drawn from a node-wide counter instead of a per-topic one,
+  so they stay unique across same-name topics. Element ids are node-local handles the peer only echoes back, so this
+  is wire-compatible with earlier 3.8.x nodes.
--- a/cpp/src/DataStorm/TopicI.h
+++ b/cpp/src/DataStorm/TopicI.h
@@ -179,10 +179,8 @@ namespace DataStormI
         // The number of connected listeners.
         size_t _listenerCount{0};
 
-        // Element ids are drawn from a single counter for all elements of a topic (keyed, any-key, and filtered).
-        // Any-key and filtered elements are both presented to the peer as negated ids (filter subscriptions), so
-        // distinct raw values are required across all of them to keep their subscriptions apart.
-        std::int64_t _nextId{0};
+        // Element ids are drawn from a node-wide counter on the Instance (see Instance::nextElementId), not from a
+        // per-topic counter, so they are unique across all the node's topics.
         std::int64_t _nextSampleId{0};
     };
 
--- a/cpp/include/DataStorm/DataStorm.h
+++ b/cpp/include/DataStorm/DataStorm.h
@@ -1520,12 +1520,17 @@ namespace DataStorm
     Topic<Key, Value, UpdateTag>::Topic(const Node& node, std::string name) noexcept
         : _name(std::move(name)),
           _topicFactory(node._factory),
-          _keyFactory(DataStormI::KeyFactoryT<Key>::createFactory()),
-          _tagFactory(DataStormI::TagFactoryT<UpdateTag>::createFactory()),
+          // The key, tag, and filter ids are drawn from a node-wide counter (rather than per-topic) so they are
+          // unique across all the node's topics; the peer echoes them back as opaque handles that this node
+          // resolves against every same-name topic, so per-topic ids would collide and misroute.
+          _keyFactory(DataStormI::KeyFactoryT<Key>::createFactory(node.getIdCounter())),
+          _tagFactory(DataStormI::TagFactoryT<UpdateTag>::createFactory(node.getIdCounter())),
           _keyFilterFactories(std::make_shared<DataStormI::FilterManagerT<DataStormI::KeyT<Key>>>()),
           _sampleFilterFactories(
               std::make_shared<DataStormI::FilterManagerT<DataStormI::SampleT<Key, Value, UpdateTag>>>())
     {
+        _keyFilterFactories->setIdCounter(node.getIdCounter());
+        _sampleFilterFactories->setIdCounter(node.getIdCounter());
         RegexFilter<Key, Key>::add(_keyFilterFactories);
         RegexFilter<Sample<Key, Value, UpdateTag>, Value>::add(_sampleFilterFactories);
         _sampleFilterFactories->set("_event", makeSampleEventFilter(*this));
--- a/cpp/include/DataStorm/InternalT.h
+++ b/cpp/include/DataStorm/InternalT.h
@@ -8,6 +8,8 @@
 #include "InternalI.h"
 #include "Types.h"
 
+#include <atomic>
+
 #if defined(__clang__)
 #    pragma clang diagnostic push
 #    pragma clang diagnostic ignored "-Wshadow-field-in-constructor"
@@ -156,6 +158,11 @@ namespace DataStormI
 
         void init() { _deleter = Deleter{std::enable_shared_from_this<AbstractFactoryT<K, V>>::shared_from_this()}; }
 
+        // Draw element ids from the given node-wide counter instead of this factory's own, so ids are unique across
+        // all the node's topics rather than per-topic (see Instance::getIdCounter). Must be called before any
+        // element is created.
+        void setIdCounter(std::shared_ptr<std::atomic<std::int64_t>> counter) noexcept { _nextId = std::move(counter); }
+
         template<typename F, typename... Args>
         [[nodiscard]] std::shared_ptr<typename V::BaseClassType> create(F&& value, Args&&... args)
         {
@@ -208,7 +215,7 @@ namespace DataStormI
             }
 
             auto k =
-                std::shared_ptr<V>(new V(std::forward<F>(value), std::forward<Args>(args)..., ++_nextId), _deleter);
+                std::shared_ptr<V>(new V(std::forward<F>(value), std::forward<Args>(args)..., ++(*_nextId)), _deleter);
             _elements[k->get()] = k;
             _elementsById[k->getId()] = k;
             return k;
@@ -237,7 +244,9 @@ namespace DataStormI
         mutable std::mutex _mutex;
         std::map<K, std::weak_ptr<V>> _elements;
         std::map<std::int64_t, std::weak_ptr<V>> _elementsById;
-        std::int64_t _nextId{1};
+        // Defaults to this factory's own counter (starting at 1, so the first id is 2 and id 1 stays reserved for
+        // the match-all filter); setIdCounter replaces it with the shared node-wide counter.
+        std::shared_ptr<std::atomic<std::int64_t>> _nextId{std::make_shared<std::atomic<std::int64_t>>(1)};
     };
 
     template<typename K> class KeyT final : public Key, public AbstractElementT<K>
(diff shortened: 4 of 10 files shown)


externl (commented):
Verified the reserved ids still hold: `AlwaysMatchFilter::getId()` is a protocol constant, the shared counter starting at 1 reproduces the old per-factory `_nextId{1}`, and 0 (the inline-key sentinel in `DataSample.keyId`) is never drawn. `setIdCounter` also lands before `RegexFilter::add` and propagates through `FilterManagerT::set`, so filters added later via `setKeyFilter`/`setSampleFilter` are covered.

The fix is broader than the description claims. Node-unique ids also neutralize the spurious `detachElements` that the same-name fan-out generates: the non-owning topic resolves the echoed key id to `nullptr` and pushes the peer element id into `removedIds`, and on the peer `subscriber.remove()` is now a no-op instead of possibly tearing down a live subscription.

Note the tests conflict with #6138 — both append at the same spot in `events/Reader.cpp` and `events/Writer.cpp`. `Instance.h` automerges.

externl (approved):

cpp/src/DataStorm/TopicI.h
```diff
@@ -179,10 +179,8 @@ namespace DataStormI
         // The number of connected listeners.
         size_t _listenerCount{0};
 
-        // Element ids are drawn from a single counter for all elements of a topic (keyed, any-key, and filtered).
-        // Any-key and filtered elements are both presented to the peer as negated ids (filter subscriptions), so
-        // distinct raw values are required across all of them to keep their subscriptions apart.
-        std::int64_t _nextId{0};
+        // Element ids are drawn from a node-wide counter on the Instance (see Instance::nextElementId), not from a
+        // per-topic counter, so they are unique across all the node's topics.
         std::int64_t _nextSampleId{0};
```
bernardnormier, line 184: This comment describes a member that no longer exists, and it now sits above `_nextSampleId`, which it doesn't describe. `Instance::nextElementId` already documents the design.

```suggestion
        std::int64_t _nextSampleId{0};
```

changelog.d/datastorm/6068.md
```diff
@@ -0,0 +1,5 @@
+- Fixed a bug in DataStorm where a node hosting two or more `Topic` instances with the same name that read from the
+  same remote writer misrouted reader initialization and sample delivery: a reader received the wrong key's value or
+  blocked. Topic element, key, tag, and filter ids are now drawn from a node-wide counter instead of a per-topic one,
+  so they stay unique across same-name topics. Element ids are node-local handles the peer only echoes back, so this
+  is wire-compatible with earlier 3.8.x nodes.
```
bernardnormier, line 5: Suggestion: drop "in DataStorm" (implied by the section), drop "or blocked" (that failure mode is from the epoch path of #6067, which never shipped), keep the entry on the symptom rather than the id-counter mechanism — and drop the wire-compatibility sentence, which #6138 makes moot (3.8.3 nodes no longer talk to earlier 3.8.x at all).

```suggestion
- Fixed a bug where a node hosting two or more `Topic` instances with the same name, reading from the same remote
  writer, could deliver samples to the wrong reader: a reader could receive another key's value.
```

bernardnormier (commented):
Given #6138 breaks the DataStorm session wire for 3.8.3 anyway, wire compatibility no longer constrains this fix — which raises a design question: should we fix the addressing instead of the ids?

Topic ids are already node-unique, and each element exchange corresponds to a specific (sender topic instance, receiver topic instance) pair — the sender always knows which peer topic instance it is responding to, because it received that instance's announcement under its unique topic id. But `attachElements`, `attachElementsAck`, and `initSamples` carry only the sender's topic id, so the receiver fans over all its same-name topics (`runWithTopics`) and resolves the echoed element and key ids against each of them. That fan-out is the actual bug.

If these three operations also carried the destination topic id — the same pattern as `DataSamples.peerId` in #6138, one level up — the receiver would route to exactly one topic instance and per-topic ids would stay unambiguous by construction:

- no node-wide counters, and no changes to the public headers (`DataStorm.h`, `InternalT.h`, `Node.h`);
- every peer-echoed handle is scoped by construction, current and future, instead of enumerating the id kinds (the issue initially missed that key ids collide too);
- live sample delivery is unaffected: subscriber lists are built at attach time per topic pair, so once attach routes exactly, the existing delivery path stays as is.

The cost is a rework of the session routing rather than this PR's mechanical change. If you judge that too risky for 3.8.3 the counter approach works — but this is the one release where the protocol change is free, and I'd rather not spend it on making the fan-out survivable when we can remove the ambiguity outright.

If the counter approach stays: the per-topic→node-wide rationale is currently repeated in six places (`Topic` ctor, `InternalT.h` ×2, `Instance.h` ×2, `Node.h`) — I'd keep the full story on `Instance` and reduce the rest to pointers. Also, `createFactory`'s `idCounter = nullptr` default has no remaining caller — make the parameter required, or pass the counter to `FilterManagerT`'s constructor instead of the setter with its "must be called before" contract.

pepone:
Superseded by #6165, which fixes #6068 with the destination-topic-id addressing approach @bernardnormier suggested in the review here instead of node-wide id counters: each `attachElements`/`attachElementsAck`/`initSamples` is routed to the exact destination topic instance rather than fanned over every same-name topic, so per-topic ids stay unambiguous by construction. No public-header changes, and it composes with #6138's reader-element addressing.

PR #6141, opened 2026-07-23
Fix typos and grammar in the Swift mapping

externl:
Comment and doc-comment typo fixes across the Swift mapping — misspellings (`unmarhshaled`), duplicated words (`with with`), subject-verb agreement, and `its`/`it's`. No behavior changes.

Three edits are worth a closer look, since they are not pure comment fixes:

- `CompileSlicePlugin.swift` — the "Missing config file" error message had mismatched quoting (a `'` opened and a backtick closed it, plus a stray ` '` before the period). Now uses matching single quotes.
- `ControllerI.swift` — renamed the local `documentsDirectory` to `cachesDirectory`, and the matching `fatalError` text. The code asks for `.cachesDirectory`, so the old name and message described the wrong location. Local variable only.
- `Endpoint.swift` — `OpaqueEndpointInfo.rawBytes` was documented as "the raw encoding of the opaque endpoint"; it is a `ByteSeq`, so it now reads "the raw bytes".

The remaining changes are comments and doc comments only. Where a comment is mirrored in the other language mappings, only the Swift copy is touched here — the siblings are handled in the companion PRs.

### DIFF
--- a/swift/Plugins/CompileSlice/CompileSlicePlugin.swift
+++ b/swift/Plugins/CompileSlice/CompileSlicePlugin.swift
@@ -49,7 +49,7 @@ enum PluginError: Error {
             return "Missing slice compiler: '\(path)'."
         case .missingConfigFile(let path, let target):
             return
-                "Missing config file '\(path)` for target `\(target)`. '. This file must be included in your sources."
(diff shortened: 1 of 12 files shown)


externl:

PR #6142, opened 2026-07-23
Fix typos and grammar in the JavaScript mapping

externl:
Comment and doc-comment typo fixes across the JavaScript mapping — `that`/`than`, `its`/`it's`, missing articles, a stale `_sendStream` field reference, and `shutdown` used as a verb. No behavior changes.

Four edits go beyond plain comment fixes and are worth a look:

- `StringUtil.js` — `"has invalid ordinal value" + c` was missing the trailing space, so the message read `...ordinal value3`. This is a user-visible exception message.
- `Properties.js` — the comment block describing the property-prefix match had the comparison backwards (it said the property must start with the key, and described the key as shorter). Rewritten to match what the code does; the code is unchanged.
- `idleTimeout/Client.ts` — renamed `testServerWithEnableDisableIdleCheck` to `testEnableDisableIdleCheck` and adjusted the test banner: the helper drives the client-side idle check, not the server's. Test-local; both call sites updated.
- `InputStream.js` — dropped a comment line that duplicated a sentence already present three lines above it.

Where a comment is mirrored in the other language mappings, only the JavaScript copy is touched here — the siblings are handled in the companion PRs.

### DIFF
--- a/js/packages/ice/src/Ice/ConnectionI.js
+++ b/js/packages/ice/src/Ice/ConnectionI.js
@@ -994,7 +994,7 @@ export class ConnectionI {
             }
 
             // We send a heartbeat to the peer to generate a "write" on the connection. This write in turns creates
-            // a read on the peer, and resets the peer's idle check timer. When _sendStream is not empty, there is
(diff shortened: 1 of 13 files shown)


externl:

PR #6144, opened 2026-07-23
Fix typos in the Slice compilers and Slice definitions

externl:
Comment and doc-comment typo fixes in the Slice parser, the ten `slice2*` generators, and the Slice definitions under `slice/`. No behavior changes.

Two things to look at:

- `Slice/MetadataValidation.cpp` — the `deprecatedInfo.extraValidation` lambda assignment ended with `,` instead of `;`, chaining it to the following `knownMetadata.emplace` through the comma operator. Same semantics, but it reads as a typo; changed to `;`.
- `Slice/Scanner.l` and `Slice/Grammar.y` — the `fowards` to `forwards` fix is applied to both the flex/bison sources and their checked-in generated output, so the two stay in sync.

The `slice/` changes are doc comments on the public Slice contract — they affect generated API documentation in every mapping, so they may deserve a closer read than the rest.

### DIFF
--- a/cpp/src/Slice/DocCommentParser.cpp
+++ b/cpp/src/Slice/DocCommentParser.cpp
@@ -285,12 +285,12 @@ namespace
         openingEnd = (openingEnd == string::npos ? line.size() : openingEnd);
         auto count = openingEnd - pos;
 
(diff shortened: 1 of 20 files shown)


externl:

PR #6146, opened 2026-07-23
Fix typos and grammar in the C# mapping

externl:
Comment, doc-comment, and message-string typo fixes across the C# runtime and test suite. No behavior changes.

The non-comment edits worth reading:

- `Internal/BZip2.cs` — the guard is `decompressedSize <= headerSize`, but the exception said `"compressed size <= header size"`. Now says `decompressed`.
- `Internal/Instance.cs` — `"Both IPV4 and IPv6..."` to `IPv4`.
- `SSL/RFC2253.cs` — `"expected end of data"` to `"unexpected end of data"`, matching the neighbouring messages and the C++/Java copies.
- `IceLocatorDiscovery/PluginI.cs` — a concatenated warning was missing the space between `InstanceName'` and `is not`.
- `ZeroC.Ice.Slice.Tools/SliceDependTask.cs` — two MSBuild log messages said the source is "older than" the depend file/target; both guards are `sourceInfo.LastWriteTime > …`, i.e. newer. Log text only.
- `BUILDING.md` — "The Glacier2 test also require" to "requires" (there is exactly one Glacier2 test).

Everything else is comments and doc comments. Where a comment is mirrored in the other mappings, only the C# copy is touched here.


externl:

PR #6148, opened 2026-07-23
Fix typos and grammar in the Python, Ruby, PHP, and MATLAB mappings

externl:
Comment, doc-comment, and message-string typo fixes across the C-extension bindings and the MATLAB toolbox. No behavior changes.

The non-comment edits:

- `IcePy/EndpointInfo.cpp` and `IcePy-stubs/__init__.pyi` — `"Provides access to a TCP endpoint information"` drops the article, in the `tp_doc` strings and the matching stubs. These are the docstrings users see from `help()`, and the stub file is required to stay in sync with the C source.
- `IcePy/Properties.cpp` — `"defaults must be None or a Ice.Properties"` to `an`.
- `IceRuby/Endpoint.cpp` — `"argument must be a endpoint"` to `an endpoint`.
- `php/src/Util.cpp` — a type-mismatch message named the received class where it meant the expected one.
- `matlab/lib/+Ice/TCPEndpointInfo.m` — same article fix as the Python docstrings, matching the other MATLAB endpoint-info classes.

Everything else is comments.

### DIFF
--- a/matlab/lib/+Ice/InputStream.m
+++ b/matlab/lib/+Ice/InputStream.m
@@ -565,7 +565,7 @@ classdef InputStream < handle
                     format = bitand(v, uint8(7)); % First 3 bits.
                     tag = uint32(bitshift(v, -3));
                     if tag > 30
-                        % We check for '> 30' instead of '> 29' because 30 is special sentinel tag, handled by the next block.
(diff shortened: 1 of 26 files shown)


externl:

PR #6150, opened 2026-07-23
Update the slice2py and slice2php man pages to match the compilers

externl:
The man pages document options the compilers no longer accept, and omit ones they do. Verified against the option tables in the sources rather than from the help text.

- `slice2py.1` — `--all`, `--prefix`, `--no-package`, and `--build-package` are gone; `--build` and `--list-generated` (both taking `modules|index|all`) were undocumented. See the `addOpt` list and usage block in `cpp/src/slice2py/PythonUtil.cpp`.
- `slice2php.1` — `--no-namespace` is gone. See `cpp/src/slice2php/Main.cpp`.

Split out of the typo PRs because this rewrites documentation content rather than fixing spelling.

### DIFF
--- a/man/man1/slice2php.1
+++ b/man/man1/slice2php.1
@@ -86,11 +86,6 @@ generate any code.
 .br
 Generate code for all Slice definitions, including those from included files.
 
-.TP
-.BR \-\-no\-namespace\fR
-.br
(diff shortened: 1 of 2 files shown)


externl:

PR #6151, opened 2026-07-23
Remove the duplicate ZeroC.Ice DistPackage entry

externl:
`ZeroC.Ice` is listed twice in the `DistPackage` item group in `csharp/msbuild/ice.proj`, with identical `ProjectDir` metadata.

Harmless either way — the `Publish` target consumes it via `%(DistPackage.ProjectDir)` batching, so the repeat is a no-op — but it should not ride along in a typo PR.

### DIFF
--- a/csharp/msbuild/ice.proj
+++ b/csharp/msbuild/ice.proj
@@ -16,9 +16,6 @@
         <DistPackage Include="ZeroC.Glacier2">
             <ProjectDir>$(MSBuildThisFileDirectory)..\src\Glacier2</ProjectDir>
         </DistPackage>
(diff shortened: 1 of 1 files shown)


pepone (approved):

InsertCreativityHere:
This PR is correct, but, just a note, it breaks the usual ordering we use for our projects.
Normally we do: `Glacier2`, `Ice`, `IceBox`, etc...

But now, the order in this solution file has `Ice` and `Glacier2` flipped.
We should probably fix the ordering to match what we do for all our other projects.
For an example, check the `ice.dist.slnx` or `ice.slnx` files next to the one this PR touched

PR #6153, opened 2026-07-23
Fix typos and grammar across the repository

externl:
Comment, doc-comment, and message-string typo fixes across every language mapping, the Slice compilers, the test suites, and the build tooling — 295 files, +674/-703. No behavior changes.

Found with `codespell` over all tracked text files, a duplicated-word scan, and a single-occurrence dictionary check, then a per-directory sweep for the classes a spell checker structurally cannot catch: `its`/`it's`, `that`/`than`, `a`/`an`, subject-verb agreement, and comments that contradict the code beside them. Every edit was re-reviewed afterwards; 19 were corrected or reverted when that second pass found the fix itself was wrong — for example `generated in in the target language` is not a doubled word (the first `in` is a stranded preposition), `the middleware execute` is a deliberate plural collective used throughout the public API docs, and one past-tense comment was explaining why the assertion on the following line holds.

Roughly 227 of the 286 files are comment-only. The rest touch a user-visible string, an identifier, or documentation content — those are listed below, since that is where review attention is worth spending.

- `Glacier2/ProxyVerifier.cpp` — the address-filter description joined ranges with `", or"`, producing `10.0.0.5, or10.0.0.9`.
- `Glacier2/SessionRouterI.cpp` — local `sessionForwader` renamed (declaration and two uses).
- `Ice/SSL/RFC2253.{cpp,cs,java}` — `"expected end of data"` to `"unexpected end of data"`; the guard is `pos >= data.size()`. The Java copy also leaked the internal field name as `end of state.data`.
- `IceGrid/PlatformInfo.cpp` — `GetLogicalProcessInformation` to `GetLogicalProcessorInformation`, matching the Win32 symbol resolved a few lines above.
- `IceGrid/Parser.cpp` — one help-table entry indented a column past the other 38.
- `csharp Internal/BZip2.cs` — the guard is `decompressedSize <= headerSize` but the message said `"compressed size"`.
- `csharp/java Instance` — `"Both IPV4 and IPv6"` to `IPv4`.
- `java ReferenceFactory.java` — `"-e option in in proxy string"`.
- `java TwowayOnlyException.java` — `"datagram, of batchDatagram"` to `or`.
- `java Object.java` — `{@link ice_isA}` and siblings were unresolvable without `#`.
- IceGridGUI — an unterminated `<html>` tag, `"Enter 0 to returns"`, `"Unable reload Application"`, `"Which application do you to display"`, and `Subject Alternate` to `Subject Alternative` (RFC 5280).
- `js StringUtil.js` — `"has invalid ordinal value" + c` was missing the trailing space, rendering as `...value3`.
- `js Properties.js` — the prefix-match comment had the comparison backwards; code unchanged.
- `js idleTimeout/Client.ts` — test helper renamed; it drives the client-side idle check, not the server's.
- `swift CompileSlicePlugin.swift` — the "Missing config file" message had mismatched quoting.
- `swift ControllerI.swift` — local `documentsDirectory` renamed to `cachesDirectory`; the code requests `.cachesDirectory`.
- `IcePy EndpointInfo.cpp` + `IcePy-stubs/__init__.pyi` — `tp_doc` docstrings and their stubs, kept in sync.
- `Slice/MetadataValidation.cpp` — a lambda assignment ended with `,` instead of `;`, chaining to the next statement via the comma operator. Same semantics.
- `Slice/Scanner.l` + `Grammar.y` — fixes applied to both the flex/bison sources and their checked-in generated output.
- `man/man1/icestormdb.1` — a `.TP` macro missing its leading dot rendered as literal text.
- `man/man1/slice2py.1`, `slice2php.1` — these documented options the compilers no longer accept (`--all`, `--prefix`, `--no-package`, `--build-package`, `--no-namespace`) and omitted `--build`/`--list-generated`. Verified against the option tables in `PythonUtil.cpp` and `slice2php/Main.cpp`.
- `README.md` — a broken relative link, `java/test//android/controller/`.

`cpp/test/IceGrid/deployer/` — renamed `Overrided` to `Overridden` across `AllTests.cpp` and the six XML descriptors (41 occurrences, 8 identifier forms). These variable, parameter, and property names are defined by the test's own descriptors and appear nowhere in `cpp/src/IceGrid`, `slice/IceGrid`, or any shipped schema, so they are safe to rename. Verified afterwards that every `getProperty("…")` in the test still resolves to a `name="…"` in a descriptor, and every `${…}` reference to a declaration.

Deliberately left alone: the French text in `cpp/test/IceUtil/unicode/coeur.utf8`, external API names such as `apple.laf.useScreenMenuBar`, and the joke name in `RFC2253.h`.

Two changes were intentionally kept out of this PR, since neither is a typo fix; both have since merged separately: #6151 and #6152.

clang-format (llvm 19, matching CI) is clean on all 130 modified C++ files.


pepone (approved):
A couple of remaining wording issues:

1. In cpp/src/Ice/WSTransceiver.cpp, the comment should say “we assume we did not initiate the close,” not “the connection.” _closingInitiator = false changes which peer is treated as initiating the close handshake. The same wording also remains in the Java and C# implementations.

2. In java/src/IceGridGUI/src/main/resources/metrics.cfg, the tooltip still reads: “Outstanding event are event which were sent by not yet delivered.” Suggested wording: “Outstanding events are events that were sent but not yet delivered.”

Otherwise, the executable changes look mechanically safe, and git diff --check passes.

cpp/include/DataStorm/Types.h
```diff
@@ -181,7 +181,7 @@ namespace DataStorm
         /// Clones the given value. This helper is used when processing partial update to clone the previous value
         /// and compute the new value with the partial update. The default implementation performs a plain C++ copy
         /// with the copy constructor.
-        /// @param value The value to encode
+        /// @param value The value to clone
         /// @return The cloned value
```
InsertCreativityHere, line 185: Since this is a typos/grammar PR...
The other functions I can see end these with periods, like the `encode` function this one was clearly copy/pasted from. So, should probably also end with periods here for consistency.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -124,7 +124,7 @@ DataElementI::attach(
         name = os.str();
     }
 
-    // Attach the key or filter, and if attach success compute the ACK data to send to the peer.
+    // Attach the key or filter, and if the attach succeeds compute the ACK data to send to the peer.
```
InsertCreativityHere: I find this very hard to read without the comma. I do think it is technically correct though...
```suggestion
    // Attach the key or filter, and if the attach succeeds, compute the ACK data to send to the peer.
```

cpp/src/DataStorm/NodeI.h
```diff
@@ -119,7 +119,7 @@ namespace DataStormI
         // A map of all publisher sessions, indexed by the identity of the peer node.
         std::map<Ice::Identity, std::shared_ptr<PublisherSessionI>> _publishers;
 
-        // A proxy to a colocated publisher session object that forwards requests to all active publisher sessions.
+        // A proxy to a collocated publisher session object that forwards requests to all active publisher sessions.
```
InsertCreativityHere, line 122: The last two straggling `colocated` that have snuck by all these years... good catch!

cpp/src/Ice/SSL/SecureTransportEngine.cpp
```diff
@@ -34,8 +34,8 @@ namespace
     //
     // Retrieve the name of a cipher, SSLCipherSuite includes duplicated values for TLS/SSL
     // protocol ciphers, for example SSL_RSA_WITH_RC4_128_MD5/TLS_RSA_WITH_RC4_128_MD5
-    // are represented by the same SSLCipherSuite value, the names return by this method
-    // doesn't include a protocol prefix.
+    // are represented by the same SSLCipherSuite value, the names returned by this method
+    // don't include a protocol prefix.
```
InsertCreativityHere: Since this is a grammar PR: I feel like a colon _feels_ more right here than a comma...

```suggestion
    // are represented by the same SSLCipherSuite value: the names returned by this method
    // don't include a protocol prefix.
```

cpp/src/Ice/SSL/SecureTransportTransceiverI.cpp
```diff
@@ -229,7 +229,7 @@ Ice::SSL::SecureTransport::TransceiverI::initialize(IceInternal::Buffer& readBuf
                     continue; // Call SSLHandshake to resume the handshake.
                 }
             }
-            // Let it fall through, this will raise a SecurityException with the SSLCopyPeerTrust error.
+            // Let it fall through, this will raise a ProtocolException with the SSLCopyPeerTrust error.
```
InsertCreativityHere: Some python snuck in here I guess?
```suggestion
            // Let it fall through, this will throw a ProtocolException with the SSLCopyPeerTrust error.
```

cpp/src/Ice/ConnectionFactory.cpp
```diff
@@ -426,7 +426,7 @@ IceInternal::OutgoingConnectionFactory::getConnection(
             throw Ice::CommunicatorDestroyedException(__FILE__, __LINE__);
         }
 
-        // Search for an existing connections matching one of the given endpoints.
+        // Search for existing connections matching one of the given endpoints.
```
InsertCreativityHere: Seems like this got fixed the wrong way. The return value is a _single_ `ConnectionI`, not multiple.
So this should of gotten fixed to be:

```suggestion
        // Search for an existing connection matching one of the given endpoints.
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/InputStream.java
```diff
@@ -473,7 +473,7 @@ public int readAndCheckSeqSize(int minSize) {
         }
 
         // If there isn't enough data to read on the stream for the sequence (and possibly enclosed sequences),
-        // something is wrong with the marshaled data: it's claiming having more data that what is possible to read.
+        // something is wrong with the marshaled data: it's claiming having more data than what is possible to read.
```
InsertCreativityHere: This sentence is wrong, but the fix doesn't fix it.
```suggestion
        // something is wrong with the marshaled data: it's claiming to have more data than what is possible to read.
```

cpp/src/DataStorm/TopicI.cpp
```diff
@@ -781,7 +781,7 @@ TopicI::forward(const ByteSeq& inParams, const Current& current) const
     // Forwarder proxy must be called with the mutex locked!
     for (const auto& [_, listener] : _listeners)
     {
-        // Forward the call to all listeners using its session proxy, passing nullptr for the callbacks because we
+        // Forward the call to all listeners using their session proxy, passing nullptr for the callbacks because we
```
InsertCreativityHere: The grammar still sounds off. If this is plural with `their`, should this become:
```suggestion
        // Forward the call to all listeners using their session proxies, passing nullptr for the callbacks because we
```
Or am I missing something about the implementation here?

cpp/src/IceGrid/Parser.cpp
```diff
@@ -2403,7 +2403,7 @@ Parser::showWarranty()
 
 //
 // With older flex version <= 2.5.35 YY_INPUT second
-// paramenter is of type int&, in newer versions it
+// parameter is of type int&, in newer versions it
 // changes to size_t&
 //
```
InsertCreativityHere, line 2408: I know this is a grammar PR, but... this comment still sounds weird and wrong to me, and we should just delete it. Flex '2.5.35' is 18 years old now.

cpp/src/IceGrid/Parser.h
```diff
@@ -85,7 +85,7 @@ namespace IceGrid
 
         //
         // With older flex version <= 2.5.35 YY_INPUT second
-        // paramenter is of type int&, in newer versions it
+        // parameter is of type int&, in newer versions it
         // changes to size_t&
         //
```
InsertCreativityHere: Same, we should just delete this super outdated comment.

cpp/src/IceStorm/Parser.cpp
```diff
@@ -362,7 +362,7 @@ Parser::showBanner()
 
 //
 // With older flex version <= 2.5.35 YY_INPUT second
-// paramenter is of type int&, in newer versions it
+// parameter is of type int&, in newer versions it
 // changes to size_t&
 //
```
InsertCreativityHere: We should just delete this very old comment.

cpp/src/IceStorm/Parser.h
```diff
@@ -31,7 +31,7 @@ namespace IceStorm
 
         //
         // With older flex version <= 2.5.35 YY_INPUT second
-        // paramenter is of type int&, in newer versions it
+        // parameter is of type int&, in newer versions it
         // changes to size_t&
         //
```
InsertCreativityHere: We should just delete this very old comment.

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -256,11 +256,11 @@ Slice::Preprocessor::preprocess(const string& languageArg)
         char* buf = mcpp_get_mem_buffer(Out);
 
         //
-        // First try to open temporay file in tmp directory.
+        // First try to open temporary file in tmp directory.
```
InsertCreativityHere: I think this could be further improved:
```suggestion
        // First try to open a temporary file in the 'tmp' directory.
```

cpp/src/slice2js/Gen.cpp
```diff
@@ -928,7 +928,7 @@ Slice::ImportVisitor::writeImports(
 
         if (jsModule == jsImportedModule || jsImportedModule.empty())
         {
-            // For Slice modules mapped to the same JavaScript module, or Slice files that doesn't use "js:module".
+            // For Slice modules mapped to the same JavaScript module, or Slice files that don't use "js:module".
             // We import them using their Slice include relative path.
```
InsertCreativityHere: Because this is a grammar PR... this is actually just 1 sentence. Not 2.
```suggestion
            // For Slice modules mapped to the same JavaScript module, or Slice files that don't use "js:module",
            // we import them using their Slice include relative path.
```

bernardnormier (approved):
Looks good. One nit: the description says the "Overrided" values in the IceGrid deployer tests were deliberately left alone, but the diff renames them to "Overridden" — consistently, in both the XML descriptors and AllTests.cpp, so the change itself is fine. Please reconcile the description with the diff.

externl:
Good catch — thanks. The description was written before the rename and I never reconciled it.

Updated: the `Overrided` → `Overridden` rename is now described in place of the old "left alone" claim, with the reasoning (test-local names, absent from `cpp/src/IceGrid`, `slice/IceGrid`, and any shipped schema) and the reference checks I ran afterwards.

While in there I also refreshed two other stale bits: the file/line counts (286/+618/-626 → 295/+665/-673, they had drifted across the review fixes and the rename) and the closing note, since #6151 and #6152 have both merged since.

cpp/test/Ice/metrics/AllTests.cpp
```diff
@@ -101,7 +101,7 @@ namespace
             }
 
             // Ensure that the previous updates were committed, the setProperties call returns before
-            // notifying the callbacks so to ensure all the update callbacks have be notified we call
+            // notifying the callbacks so to ensure all the update callbacks have been notified we call
```
InsertCreativityHere: Really should be some kind of punctuation to split this up. Either a comma or maybe even a new sentence.
```suggestion
            // notifying the callbacks, so to ensure all the update callbacks have been notified we call
```

cpp/test/IceSSL/configuration/AllTests.cpp
```diff
@@ -911,7 +911,7 @@ testFindCert(const string& factoryRef, const string& defaultDir, const Ice::Prop
         initData.properties->setProperty("IceSSL.KeychainPassword", "password");
         initData.properties->setProperty("IceSSL.FindCert", clientFindCertProperties[i]);
 
-// Use TrustOnly to ensure the peer has pick the expected certificate.
+// Use TrustOnly to ensure the peer has picked the expected certificate.
```
InsertCreativityHere: Missing space in the front.
```suggestion
        // Use TrustOnly to ensure the peer has picked the expected certificate.
```
Note that we have this same comment on an ifdef further down, and even there we space it right. So this isn't just a special case for the ifdef.

scripts/ws_ping_probe.py
```diff
@@ -109,7 +109,7 @@ def ping_pong(host, port, payload=b"", timeout=10.0):
     Upgrade, send a PING with the given payload, and return the PONG payload.
 
     Non-PONG frames the server emits first (e.g. its Ice connection-validation
-    message) are skipped. Raises if the server sends CLOSE or never ponngs.
+    message) are skipped. Raises if the server sends CLOSE or never pongs.
```
InsertCreativityHere: judging from the rest of this comment, this should probably be capitalized:

```suggestion
    message) are skipped. Raises if the server sends CLOSE or never PONGs.
```

slice/Ice/Metrics.ice
```diff
@@ -18,7 +18,7 @@ module IceMX
     /// A dictionary of strings to integers.
     dictionary<string, int> StringIntDict;
 
-    /// The base class for metrics. A metrics object represents a collection of measurements associated to a given a
+    /// The base class for metrics. A metrics object represents a collection of measurements associated to a given
```
InsertCreativityHere: Should this be "with" instead of "to"? Both grammatically correct, but "with" _feels_ more normal to me.
```suggestion
    /// The base class for metrics. A metrics object represents a collection of measurements associated with a given
```

scripts/generate-cpp-code-coverage.sh
```diff
@@ -10,7 +10,7 @@ export LDFLAGS="-fprofile-instr-generate"
 export LLVM_PROFILE_FILE="${PWD}/coverage/%m.profraw"
 export OPTIMIZE=no
 
-# Use llvm from homebrew if we're on macOS
+# The llvm tools used below come from the Xcode Command Line Tools, so this script is macOS only
```
InsertCreativityHere: I agree with expanding this into a real sentence, but then we should make it a real sentence.
```suggestion
# The llvm tools used below come from the Xcode Command Line Tools, so this script is macOS only.
```

cpp/include/Ice/Endpoint.h
```diff
@@ -123,7 +123,7 @@ namespace Ice
         }
     };
 
-    /// Provides access to a TCP endpoint information.
+    /// Provides access to TCP endpoint information.
```
InsertCreativityHere: So, I definitely don't like the old wording, but while this fixes the grammar, I think it fixes it "in the wrong direction". From the code, I think this comment was really trying to say:
```suggestion
    /// Provides access to a TCP endpoint's information.
```
Because this information is for a single specific TCP endpoint.

The new description ("Provides access to TCP endpoint information") sounds more like something I could query for what TCP endpoints are in use, or active, or available, etc.

If you disagree, feel free to keep the current change. Still an improvement.
But, my gut when reading the old comment, is that the original writer just forgot the `'s`.

----

This comment applies to all the various EndpointInfo changes in this PR.
I'm going to be lazy and not point out each and every one of them : v)

InsertCreativityHere (commented):
Definitely an improvement. I think there's some places that could be further improved.

externl:
Thanks — all applied. Details on the two I want to flag:

**EndpointInfo wording.** You're right, and the file itself settles it: `Endpoint::getInfo` right above these classes already reads "Returns **this endpoint's** information." — so the possessive was the original intent and I fixed it in the wrong direction. Changed to `Provides access to a TCP endpoint's information.` at all 32 sites (C++, C#, Java, JS `.js`+`.d.ts`, MATLAB, IcePy `tp_doc` + the `.pyi` stubs, Swift), with `a`/`an` per initial.

Worth noting for a future pass: the sibling `ConnectionInfo` classes use a third form — "Provides access to the connection details of a TCP connection." I left those alone since they're outside this PR, but the family is now split across two phrasings if you want them unified.

**The flex comments.** Deleted all four (`IceGrid/Parser.{cpp,h}`, `IceStorm/Parser.{cpp,h}`). Heads-up that the comment was the only thing explaining why `getInput` has both an `int&` and a `size_t&` overload — with flex 2.5.35 that far back, the `int&` overload is presumably dead and could go too, but that's a code change so I left it out of a typo PR.

**`Search for existing connections`** — good catch, `findConnection` returns a single `ConnectionIPtr`. The original read "an existing connections", and I resolved the disagreement the wrong way; now singular.

Everything else taken as suggested. clang-format is clean on the 17 touched C++ files, no new over-120-column lines, and the `.py`/`.pyi`/shell edits still parse.

externl:
All eight applied.

**`claiming having`** — good catch that this was a partial fix. I'd only corrected the Java copy; C++, C# and JS were flagged here, and **Swift** had it too. All four now read `claiming to have`, matching Java and MATLAB, so all six mappings agree.

**RFC2253** — agreed, and the message was wrong in a second way. That branch is guarded by `pos < size && !isDigit(data[pos])`, so a character is present and simply isn't a digit; "expected end of data" describes neither the condition nor the failure. Now `invalid attribute type (missing digit after '.')`, matching the `(missing =)` style used a few lines up. Applied to all three mappings, and no test asserts on the old text.

The rest as suggested:

- `DataElementI.cpp` — split into two sentences and named the subject. Note this comment appears twice (lines 682 and 1588); both updated.
- `GraphView.java` — split into two sentences: `Remove empty series that are no longer in use. Always keep one series, so that new values can still be added.` That matches the guard, which only removes when `row.series.size() > 1`.
- Glacier2 router test — `once it has received the 3 callbacks`.
- DataStorm reliability test — `The session has been reestablished twice without activity on the "element" key.`

Rebased onto the latest branch state along the way. clang-format clean, no new over-120-column lines.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -677,10 +677,10 @@ DataElementI::forward(const ByteSeq& inParams, const Current& current) const
 {
     for (const auto& [_, listener] : _listeners)
     {
-        // If we are forwarding a sample check if at least once of the listeners is interested in the sample.
+        // If we are forwarding a sample check if at least one of the listeners is interested in the sample.
```
InsertCreativityHere: This comment is still grammatically incorrect.
Not sure what the right fix is without reading the code.

slice/IceGrid/Admin.ice
```diff
@@ -349,9 +349,9 @@ module IceGrid
         void instantiateServer(string application, string node, ServerInstanceDescriptor desc)
             throws AccessDeniedException, ApplicationNotExistException, DeploymentException;
 
-        /// Gets an application descriptor.
+        /// Gets information about an application.
```
InsertCreativityHere, line 352: The new wording is good, but if we update the comment here, we should update the identical comment in `cpp/include/IceGrid/PluginFacade.h` right? Since right now it looks like all the comments are in sync between this file and that one.

slice/Ice/Metrics.ice
```diff
@@ -32,7 +32,7 @@ module IceMX
         /// The number of objects currently observed by this metrics.
         int current = 0;
 
-        /// The sum of the lifetime of each observed objects. This does not include the lifetime of objects which are
+        /// The sum of the lifetime of each observed object. This does not include the lifetime of objects which are
         /// currently observed, only the objects observed in the past.
```
InsertCreativityHere: There should be a plural somewhere in this sentence, because this is observing multiple objects (according to the other doc-comments at least).
```suggestion
        /// The sum of the lifetimes of each observed object. This does not include the lifetimes of objects which are
/// currently observed, only the objects observed in the past.
```

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/SessionKeeper.java
```diff
@@ -825,7 +825,7 @@ public synchronized void setIsDefault(boolean isDefault) {
         private boolean _isDefault;
     }
 
-    // FocusListener implementation that unselect the text
+    // FocusListener implementation that unselects the text
     // of a text component after focus gained.
```
InsertCreativityHere: This sentence still sounds weird to me.

csharp/test/Ice/metrics/AllTests.cs
```diff
@@ -219,7 +219,7 @@ public void
             }
 
             // Ensure that the previous updates were committed, the setProperties call returns before
-            // notifying the callbacks so to ensure all the update callbacks have be notified we call
+            // notifying the callbacks so to ensure all the update callbacks have been notified we call
             // a second time, this will block until all the notifications from the first update have
```
InsertCreativityHere: This sentence needs some more punctuation. Also, the last line fits on the one above it.
```suggestion
            // notifying the callbacks, so to ensure all the update callbacks have been notified we call
            // a second time. This will block until all the notifications from the first update have completed.
```

csharp/src/Ice/Properties.cs
```diff
@@ -240,7 +240,7 @@ public string[] getPropertyAsListWithDefault(string key, string[] value)
     }
 
     /// <summary>
-    /// Get all properties whose keys begins with prefix.
+    /// Get all properties whose keys begin with prefix.
```
InsertCreativityHere: Maybe make `prefix` use a `paramref` here?
Took me longer than I'd like to realize this was a parameter name to the operation that was too far below to see.

cpp/src/Glacier2/SessionRouterI.cpp
```diff
@@ -1104,7 +1104,7 @@ SessionRouterI::startCreateSession(const shared_ptr<CreateSession>& cb, const Co
         //
         // No session exists yet, so we will try to create one. To
         // avoid that other threads try to create sessions for the
-        // same connection, we add our endpoints to _pending.
+        // same connection, we add our connection to _pending.
```
InsertCreativityHere, line 1107: This sentence is still weirdly worded.

InsertCreativityHere (approved):
Looks good to me!

externl:
Thanks for the approval — all seven applied.

- **`PluginFacade.h` sync** — good catch, and you were right that the two files track each other. Updated both the summary and the `@return` ("The application descriptor" → "The application information"), so the pair matches again. I also diffed every doc line this PR changed in `Admin.ice` against `PluginFacade.h` to see whether any others had drifted out of sync — `getApplicationInfo` was the only one.
- **`Metrics.ice` totalLifetime** — pluralized both "lifetime"s, as suggested.
- **`Properties.cs`** — both bare `prefix` mentions in the summary are now `<paramref name="prefix" />`.
- **`DataElementI.cpp:680`** — the missing comma was the main problem, and the trailing "…interested in the sample" repeated the subject. Now: "If we are forwarding a sample, check whether at least one of the listeners is interested in it."
- **`SessionRouterI.cpp`** — "To avoid that other threads try to create sessions" was the awkward part, and it predates this PR. Now "To prevent other threads from creating a session for the same connection…".
- **`SessionKeeper.java`** — "after focus gained" → "after it gains focus".
- **metrics test comment** — applied your version, and made the **same** edit to the C++ copy at `cpp/test/Ice/metrics/AllTests.cpp:104`. It had only received the comma in the last round, so the two would otherwise have drifted apart.

clang-format clean, no new over-120-column lines.

PR #3442, opened 2025-01-29
Drop support for ICE_BIN_DIST

pepone:

PR #6158, opened 2026-07-23
Add type hints to the Python test scripts

externl:
Adds type hints to the test framework under `scripts/` — `Util.py`, `Expect.py`, `Component.py` and the component `*Util.py` modules — and enforces them with pyright in CI.

- `pyrightconfig.json` now includes `scripts`, with a `scripts` execution environment so the framework modules resolve each other's imports. The files not yet annotated are listed under `ignore`.
- The Pyright CI job runs `pyright --project ..` from `python/` so it actually checks `scripts`; the previous `pyright .` silently skipped everything outside `python/`.
- Fixes `IceStormAdmin.getExe`, which called `Component.component.useBinDist` — removed in #3442 — and so raised `AttributeError` on ppc. It now matches its sibling classes.

No changelog fragment — no user-facing impact.


scripts/Component.py
```diff
@@ -54,18 +56,20 @@ def getPhpExtension(self, mapping, current):
         else:
             return "ice.so"
 
-    def getNugetPackageVersionFile(self, mapping):
+    def getNugetPackageVersionFile(self, mapping: Util.Mapping) -> str:
         if isinstance(mapping, Util.CSharpMapping):
             return os.path.join(Util.toplevel, "csharp", "msbuild", "zeroc.ice.net.nuspec")
         else:
+            # Only the Windows nuget packages are named after the platform toolset.
+            assert isinstance(Util.platform, Util.Windows)
```
InsertCreativityHere: It is of course, impossible to follow the current code : v)
But we should probably make a note that this path means we're looking for the C++ nugets.

scripts/Expect.py
```diff
@@ -12,38 +14,61 @@
...
 
 __all__ = ["Expect", "EOF", "TIMEOUT"]
 
-win32 = sys.platform == "win32"
-if win32:
+if sys.platform == "win32":
     import ctypes
 
+win32 = sys.platform == "win32"
+
```
InsertCreativityHere: I don't understand this change at all.
Now we check the `sys.platform` twice in a row instead of checking it once like we did before?
Why?

scripts/Expect.py
```diff
@@ -73,26 +98,27 @@ def escape(s, escapeNewlines=True):
...
     if win32:
         taskkill("/F /T /PID {0}".format(p.pid))
     else:
         os.kill(p.pid, signal.SIGKILL)
 
 
-def terminateProcess(p, hasInterruptSupport=True):
-    if win32:
+def terminateProcess(p: subprocess.Popen[bytes], hasInterruptSupport: bool = True) -> None:
+    if sys.platform == "win32":
```
InsertCreativityHere, line 129: And here too, we re-check `platform` instead of checking the global?

Is this just an exercise in "globals bad"?
Because I can't imagine there's some race-condition or that `platform` is going to change.

InsertCreativityHere (approved):
I skimmed it, and it looks fine enough. Code is too convoluted to really double-check anything, and it did make some changes beyond just adding asserts and type-hints, so guess I'll just trust it to of done it correctly!

Just our testing code, so not too high a risk anyways.

scripts/Component.py
externl: Added a note: the non-C# branch is the C++ nuget package, which is only built on Windows and named after the platform toolset (e.g. `zeroc.ice.v143.nuspec`).

scripts/Expect.py
externl: Fair — it looked gratuitous and I should have explained it. Type checkers only narrow the platform for a *literal* `sys.platform == "win32"` comparison; testing the `win32` variable does not narrow, and `ctypes.windll` only exists on Windows, so the import guard has to compare `sys.platform` directly or the `ctypes.windll` use fails to type check.

Ive added a comment saying exactly that, and restored the original ordering so `win32` is defined first and used everywhere else.

scripts/Expect.py
externl: Same reason as the import above (not a globals-vs-locals thing, and no race): this branch is the one that calls `ctypes.windll`, which typeshed only declares on Windows, so the check has to be the literal `sys.platform` comparison for it to type check. Added a one-line comment pointing back at the note on the import.

pepone (approved):
Approving. This is a clean, almost-entirely-mechanical type-hints pass; I verified the runtime-affecting bits are behavior-preserving.

**Checked:**
- All 9 modules byte-compile; `IceGridUtil`/`IceStormUtil` import fine.
- The `_Base = Process if TYPE_CHECKING else object` trick on `IceGridProcess`/`IceStormProcess` keeps the runtime base as `object` — I diffed the actual MROs and `getParentProps` bindings against `main` and they're identical (`Server.getProps` for the Server-based classes, `Process.getProps` via `Client.getProps` for the Client-based ones).
- `.servers`/`.clients` → `getServers()`/`getClients()` and `current.testcase` → `current.getTestCase()` are just assert-non-None wrappers; `init()` always runs first and the two guarded call sites still guard, so no new failure mode.
- The `isinstance(..., FunctionType)` → `isinstance(..., re.Pattern)` inversions in `trace()`/`getOutput()` and the `getAll().gen()` rewrite preserve semantics and ordering.
- The `IceStormAdmin.getExe` fix is correct: `useBinDist` was removed in #3442, so the old path raised `AttributeError` on ppc — the replacement matches its six siblings.

**Minor, non-blocking:** `RemoteProcess.waitSuccess` now uses `getattr(ex, "reason", "")` — a slight behavior change (re-raises instead of masking with `AttributeError` when a `UserException` has no `reason`); fine, arguably better.

I didn't re-run the 580-case output diff or pyright locally — trusting your verification and CI for those.

PR #6159, opened 2026-07-23
Fail gracefully when a middleware factory throws during dispatch pipeline creation

bernardnormier:
An object adapter creates its dispatch pipeline lazily, at the first dispatch, by calling the middleware factories installed with `use`. A middleware factory should never throw. However, if one did throw, none of the mappings handled it well:

- **C++**: `dispatchPipeline()` is `noexcept`, so the program terminated.
- **C#**: the `Lazy<Object>` cached the exception and rethrew it at internal call sites, which could leave client invocations unanswered.
- **Java**: the object adapter cached a partial dispatch pipeline, and subsequent dispatches ran without the middleware that had already been applied.
- **JS**: the exception escaped into the connection's incoming message processing, and a subsequent dispatch could cache a partial pipeline.

With this PR, all four mappings handle a middleware factory exception the same way: the object adapter logs an error with the exception's details and installs a pipeline that fails every dispatch with a generic `UnknownException` that does not reveal the failure's cause. The failure is permanent for the object adapter — the pipeline creation is not retried.

Swift needs no change: its middleware factories are non-throwing closures returning a non-optional `Dispatcher`, and its pipeline construction already commits only on success.

Each mapping's middleware test gets a new case verifying both the generic exception and that the failure is permanent, plus changelog fragments for the four mappings.

### DIFF
--- a/cpp/src/Ice/ObjectAdapterI.cpp
+++ b/cpp/src/Ice/ObjectAdapterI.cpp
@@ -55,10 +55,7 @@ namespace
     public:
         void dispatch(IncomingRequest&, std::function<void(OutgoingResponse)>) final
         {
-            throw UnknownException{
-                __FILE__,
-                __LINE__,
-                "the object adapter could not create its dispatch pipeline"};
+            throw UnknownException{__FILE__, __LINE__, "the object adapter could not create its dispatch pipeline"};
         }
     };
 }
--- a/cpp/src/Ice/ObjectAdapterI.cpp
+++ b/cpp/src/Ice/ObjectAdapterI.cpp
@@ -48,6 +48,19 @@ namespace
     inline EndpointIPtr toEndpointI(const EndpointPtr& endp) { return dynamic_pointer_cast<EndpointI>(endp); }
 
     string emptyName{};
+
+    // Installed as the dispatch pipeline when the creation of the actual pipeline fails.
+    class FailedDispatchPipeline final : public Object
+    {
+    public:
+        void dispatch(IncomingRequest&, std::function<void(OutgoingResponse)>) final
+        {
+            throw UnknownException{
+                __FILE__,
+                __LINE__,
+                "the object adapter could not create its dispatch pipeline"};
+        }
+    };
 }
 
 Ice::ObjectAdapter::~ObjectAdapter() = default; // avoid weak vtable
@@ -535,11 +548,27 @@ Ice::ObjectAdapterI::dispatchPipeline() const noexcept
     lock_guard lock(_mutex);
     if (!_dispatchPipeline)
     {
-        _dispatchPipeline = _servantManager;
-        while (!_middlewareFactoryStack.empty())
+        try
         {
-            _dispatchPipeline = _middlewareFactoryStack.top()(std::move(_dispatchPipeline));
-            _middlewareFactoryStack.pop();
+            ObjectPtr dispatchPipeline = _servantManager;
+            while (!_middlewareFactoryStack.empty())
+            {
+                dispatchPipeline = _middlewareFactoryStack.top()(std::move(dispatchPipeline));
+                _middlewareFactoryStack.pop();
+            }
+            _dispatchPipeline = std::move(dispatchPipeline);
+        }
+        catch (const std::exception& ex)
+        {
+            Error out(_instance->initializationData().logger);
+            out << "failed to create the dispatch pipeline of object adapter '" << _name << "':\n" << ex;
+            _dispatchPipeline = make_shared<FailedDispatchPipeline>();
+        }
+        catch (...)
+        {
+            Error out(_instance->initializationData().logger);
+            out << "failed to create the dispatch pipeline of object adapter '" << _name << "': unknown c++ exception";
+            _dispatchPipeline = make_shared<FailedDispatchPipeline>();
         }
     }
     return _dispatchPipeline;
--- a/cpp/src/Ice/ObjectAdapterI.cpp
+++ b/cpp/src/Ice/ObjectAdapterI.cpp
@@ -563,12 +563,14 @@ Ice::ObjectAdapterI::dispatchPipeline() const noexcept
             Error out(_instance->initializationData().logger);
             out << "failed to create the dispatch pipeline of object adapter '" << _name << "':\n" << ex;
             _dispatchPipeline = make_shared<FailedDispatchPipeline>();
+            _middlewareFactoryStack = {};
         }
         catch (...)
         {
             Error out(_instance->initializationData().logger);
             out << "failed to create the dispatch pipeline of object adapter '" << _name << "': unknown c++ exception";
             _dispatchPipeline = make_shared<FailedDispatchPipeline>();
+            _middlewareFactoryStack = {};
         }
     }
     return _dispatchPipeline;
--- a/cpp/test/Ice/middleware/AllTests.cpp
+++ b/cpp/test/Ice/middleware/AllTests.cpp
@@ -3,8 +3,21 @@
 #include "TestHelper.h"
 #include "TestI.h"
 
+#include <stdexcept>
+
 using namespace std;
 using namespace Test;
+
+class NullLogger final : public Ice::Logger, public enable_shared_from_this<NullLogger>
+{
+public:
+    void print(const string&) final {}
+    void trace(const string&, const string&) final {}
+    void warning(const string&) final {}
+    void error(const string&) final {}
+    string getPrefix() final { return "NullLogger"; }
+    Ice::LoggerPtr cloneWithPrefix(string) final { return shared_from_this(); }
+};
 
 class Middleware final : public Ice::Object
 {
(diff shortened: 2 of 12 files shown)


cpp/src/Ice/ObjectAdapterI.cpp
```diff
@@ -48,6 +48,19 @@ namespace
...
 
     string emptyName{};
+
+    // Installed as the dispatch pipeline when the creation of the actual pipeline fails.
+    class FailedDispatchPipeline final : public Object
+    {
+    public:
+        void dispatch(IncomingRequest&, std::function<void(OutgoingResponse)>) final
+        {
+            throw UnknownException{
```
bernardnormier: Another option would be to throw `ObjectNotExistException`:
 - this provides no internal details to the client
 - this allows the client to retry with another replica (if the OA is replicated)

js/packages/ice/src/Ice/ObjectAdapter.js
```diff
@@ -266,12 +267,22 @@ export class ObjectAdapter {
...
+                while (this._middlewareStack.length > 0) {
+                    const middleware = this._middlewareStack.pop();
+                    dispatchPipeline = middleware(dispatchPipeline);
+                }
+                this._dispatchPipeline = dispatchPipeline;
+            } catch (ex) {
+                this._instance
+                    .initializationData()
+                    .logger.error(
+                        `failed to create the dispatch pipeline of object adapter '${this._name}':\n${ex}\n${ex.stack}`,
```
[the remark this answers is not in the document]
bernardnormier: Fixed in af98c0acb5: the log line now formats Error values with their stack and other thrown values via String(ex).

js/packages/test/Ice/middleware/Server.ts
```diff
@@ -90,4 +106,49 @@ export class Server extends TestHelper {
...
+        try {
+            const echo = new Test.EchoPrx(communicator, `__echo:${this.getTestEndpoint()}`);
+            const adapter = await communicator.createObjectAdapter("");
+            await echo.setConnection();
+            echo.ice_getCachedConnection()!.setAdapter(adapter);
+
+            adapter.add(new MyObjectI(), new Ice.Identity("test"));
+            adapter.use(() => {
+                throw new Error("middleware factory exception");
+            });
```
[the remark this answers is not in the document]
bernardnormier: This test intentionally verifies only the client-observable behavior: every dispatch on this object adapter fails with UnknownException. Whether the implementation retries the pipeline creation internally is not part of the contract this test pins down.

java/test/src/main/java/test/Ice/middleware/AllTests.java
```diff
@@ -146,6 +149,47 @@ public static void allTests(TestHelper helper) {
...
+        initData.logger = new NullLogger();
+
+        try (Communicator communicator = Util.initialize(initData)) {
+            ObjectAdapter oa = communicator.createObjectAdapter("");
+            ObjectPrx obj = oa.add(new MyObjectI(), new Identity("test", ""));
+
+            oa.use(
+                next -> {
+                    throw new IllegalStateException("middleware factory exception");
+                });
```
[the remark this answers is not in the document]
bernardnormier: This test intentionally verifies only the client-observable behavior: every dispatch on this object adapter fails with UnknownException. Whether the implementation retries the pipeline creation internally is not part of the contract this test pins down.

csharp/test/Ice/middleware/AllTests.cs
```diff
@@ -43,6 +44,48 @@ private static void testMiddlewareExecutionOrder(Communicator communicator, Text
...
+        output.Flush();
+
+        // Use a separate communicator with a null logger: the pipeline creation failure is logged as an error.
+        var initData = new InitializationData { logger = new NullLogger() };
+        using Communicator communicator = Util.initialize(initData);
+
+        ObjectAdapter oa = communicator.createObjectAdapter("");
+        oa.add(new MyObjectI(), new Identity { name = "test" });
+
+        oa.use(next => throw new InvalidOperationException("middleware factory exception"));
```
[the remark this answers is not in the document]
bernardnormier: This test intentionally verifies only the client-observable behavior: every dispatch on this object adapter fails with UnknownException. Whether the implementation retries the pipeline creation internally is not part of the contract this test pins down.

cpp/test/Ice/middleware/AllTests.cpp
```diff
@@ -68,9 +81,53 @@ testMiddlewareExecutionOrder(const Ice::CommunicatorPtr& communicator)
...
+    // Use a separate communicator with a null logger: the pipeline creation failure is logged as an error.
+    Ice::InitializationData initData;
+    initData.logger = make_shared<NullLogger>();
+    Ice::CommunicatorHolder communicatorHolder{Ice::initialize(initData)};
+    const Ice::CommunicatorPtr& communicator = communicatorHolder.communicator();
+
+    Ice::ObjectAdapterPtr oa = communicator->createObjectAdapter("");
+    oa->add(make_shared<MyObjectI>(), Ice::Identity{"test", ""});
+
+    oa->use([](Ice::ObjectPtr) -> Ice::ObjectPtr { throw runtime_error{"middleware factory exception"}; });
```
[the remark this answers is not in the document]
bernardnormier: This test intentionally verifies only the client-observable behavior: every dispatch on this object adapter fails with UnknownException. Whether the implementation retries the pipeline creation internally is not part of the contract this test pins down.

cpp/src/Ice/ObjectAdapterI.cpp
pepone: I was thinking if we can add a new InternalError ReplyStatus and throw a DispatchException with this reply status. But I guess UnknownException is the Ice way.

pepone (approved):
Approving — this is a clean, symmetric fix. The core change (build the pipeline into a local and commit only on success) correctly eliminates the partial-pipeline caching in Java/JS and the cached-exception rethrow in C#, and keeps C++'s `dispatchPipeline()` `noexcept`. I traced the synchronous throw from `FailedDispatchPipeline` through each connection's dispatch path and confirmed it's caught and converted into a response in all four mappings, so the client reliably gets a generic `UnknownException` with the cause kept to the server log. Good coverage too — each test asserts the generic exception, no cause leak, and permanence.

A couple of non-blocking notes:

- The client-facing message casing differs (C++ lowercase/no period vs. managed sentence-case). Presumably intentional per each mapping's convention, but flagging in case you want them uniform.
- Middleware constructed before the throwing factory are discarded without cleanup. Fine given "factories should never throw," but maybe worth a comment.

Neither needs to block. 👍

cpp/src/Ice/ObjectAdapterI.cpp
bernardnormier: Yes: InternalError maps to UnknownException is Ice.

js/packages/test/Ice/middleware/Server.ts
```diff
@@ -66,6 +80,8 @@ export class Server extends TestHelper {
                 .use((next: Ice.Object) => new Middleware(next, "B", inLog, outLog))
                 .use((next: Ice.Object) => new Middleware(next, "C", inLog, outLog));
             await communicator.waitForShutdown();
+
+            await this.testMiddlewareFactoryException(communicator, args);
```
externl: This runs after `await communicator.waitForShutdown()`, which makes it depend on a few things that aren't obvious from reading it:

- `invokingCommunicator` is communicator `#1`, which `shutdown()` has already torn down — every object adapter on it is destroyed by that point. Outgoing invocations still work, but that's a subtle property to rely on silently.
- It only reaches the second adapter because `echo.setConnection()` on the second communicator overwrites the echo server's single relay connection (`cpp/test/Ice/echo/BlobjectI.h` holds one `_connection`). Two communicators are competing for one global slot, and nothing here says so.
- The output comes out in the wrong order: `"testing middleware factory exception... ok"` prints inside the `try`, while `"testing middleware execution order... ok"` prints in the `finally`, so the second test reports first.

Moving this call to the top of `run()`, before the bidir setup and `serverReady()`, avoids all three — the test becomes self-contained (own communicator, claims the relay slot, runs, destroys it) and doesn't need `invokingCommunicator` at all. The client is blocked on `serverReady()` throughout, so there's no race.

externl (approved):
The core change looks right — building the pipeline into a local and committing only on success cleanly fixes the partial-pipeline caching in Java/JS and the cached-exception rethrow in C#, and it keeps `dispatchPipeline()` `noexcept` in C++. I traced the synchronous throw from `FailedDispatchPipeline` through each mapping's connection and collocated dispatch paths and confirmed it's caught and turned into a response everywhere.

One note on the placement of the new JS test — see inline.

bernardnormier:
> * Middleware constructed before the throwing factory are discarded without cleanup. Fine given "factories should never throw," but maybe worth a comment.

What do you mean? I don't see how you cleanup a constructed middleware. The overall partially constructed pipeline is cleaned-up since we assign the `FailedDispatchPipeline` into it.

pepone:
> The overall partially constructed pipeline is cleaned-up since we assign the FailedDispatchPipeline into it.

You are right, we drop the references that is enough.

js/packages/test/Ice/middleware/Server.ts
bernardnormier: Moving this to the top of `run()` doesn't work well in practice: the test driver consumes the server's stdout while waiting for the "server ready" marker, so everything this test prints before `serverReady()` is swallowed — and the server shouldn't delay its readiness signal on a sub-test anyway. So I kept it after `waitForShutdown()`, where the relay slot is genuinely free, and fixed the three issues you raised: it now invokes through its own communicator (the `invokingCommunicator` parameter is gone), the execution-order verification moved out of `finally` so the output prints in execution order, and the relay-connection takeover is now stated explicitly at the `setConnection()` call.

cpp/src/Ice/ObjectAdapterI.cpp
externl: I think `UnknownException` is more correct. `ObjectNotExistException` would be like a web server returning 404 when it should return a 500.

externl: I think `UnknownException` is more correct. `ObjectNotExistException` would be like a web server returning 404 when it should return a 500.

PR #6167, opened 2026-07-24
Fix unmarshaling of optional bare Object* proxies in MATLAB

### LINKED ISSUE #6125: MATLAB optional bare `Object*` unmarshaling checks the wrong `nargin` — every optional bare proxy fails
> [!NOTE]

## Summary

`InputStream.readProxyOpt` checks `nargin == 1` to detect the untyped (no-`cls`) call, but MATLAB method `nargin` counts `obj`, so the bare optional-`Object*` call (`readProxyOpt(tag)`) has `nargin == 2`. Both the value-present and value-absent branches therefore reference the omitted `cls` → runtime error, so **every optional bare `Object*` fails to unmarshal**.

**Files:** `matlab/lib/+Ice/InputStream.m:673-687`
```matlab
if nargin == 1              % never true for the bare (obj,tag) form
    r = obj.readProxy();
else
    r = obj.readProxy(cls); % cls undefined -> error
...
else
    emptyFunc = str2func(strcat(cls, '.empty')); % cls undefined -> error
```
Generator emits `readProxyOpt(tag)` (nargin 2) for bare optional `Object*` and `readProxyOpt(tag, 'Foo.FooPrx')` (nargin 3) for typed (`cpp/src/slice2matlab/Gen.cpp:3164-3169,3186`). The sibling `readProxy(obj, cls)` correctly uses `nargin == 2` — `readProxyOpt` copied those constants without accounting for its extra `tag` param, so every check is off by one. Typed optional proxies work (nargin 3). It is the only `read*Opt` method taking an optional `cls`.

## Fix

Test `nargin == 2` for the untyped `(obj, tag)` form, and add present/absent optional bare-`Object*` cases to the MATLAB optional test schema (currently absent).

**Severity:** Medium (High within MATLAB) — a complete, silent MATLAB-only hole in a Slice construct the other mappings support.

bernardnormier:
`InputStream.readProxyOpt` detected its untyped call with `nargin == 1`, but a MATLAB method's `nargin` counts `obj`, so the form the generator emits for `optional(N) Object*` — `readProxyOpt(tag)` — arrives as `nargin == 2`. The typed form, `readProxyOpt(tag, 'Foo.FooPrx')`, is `nargin == 3`.

`nargin == 1` was therefore never true, and both the value-present and value-absent branches fell through to the `cls` path with `cls` unbound:

```
testing optional proxies without a proxy type... Not enough input arguments.
Error in Ice.InputStream/readProxyOpt (line 679)
                    r = obj.readProxy(cls);
```

So every optional bare `Object*` failed to unmarshal, whether or not the sender set the value. The sibling `readProxy(obj, cls)` correctly uses `nargin == 2`; `readProxyOpt` reused those constants without accounting for its extra `tag` parameter. It is the only `read*Opt` method taking an optional `cls`.

### Test

The optional test schema has no bare `Object*` — only `MyInterface*` — so this path had no coverage at all. The new `bareOptionalProxies` helper marshals a set and an unset optional proxy and reads both back through the untyped form. It is a local `OutputStream`/`InputStream` round-trip, modeled on the `skipUnknownOptionals` helper already in that file, so it needs no Slice change and no Python server change.

Verified it fails without the fix (at the line above) and passes with it. Full MATLAB suite: 16/16.

Closes #6125. A follow-up will add `optional(11) Object* oprx;` to `MultiOptional` for end-to-end coverage across mappings and drop this local test.

### DIFF
--- a/changelog.d/matlab/6125.md
+++ b/changelog.d/matlab/6125.md
@@ -1,0 +1,2 @@
+- Fixed the unmarshaling of optional `Object*` parameters and fields. Reading such a proxy always failed with a MATLAB
+  error, whether or not the sender set the optional value.
--- a/matlab/lib/+Ice/InputStream.m
+++ b/matlab/lib/+Ice/InputStream.m
@@ -673,12 +673,12 @@ classdef InputStream < handle
         function r = readProxyOpt(obj, tag, cls)
             if obj.readOptional(tag, Ice.OptionalFormat.FSize)
                 obj.skip(4);
-                if nargin == 1
+                if nargin == 2
                     r = obj.readProxy();
                 else
                     r = obj.readProxy(cls);
                 end
-            elseif nargin == 1
+            elseif nargin == 2
                 r = Ice.ObjectPrx.empty;
             else
                 emptyFunc = str2func(strcat(cls, '.empty'));
--- a/matlab/test/Ice/optional/AllTests.m
+++ b/matlab/test/Ice/optional/AllTests.m
@@ -11,6 +11,10 @@ classdef AllTests
             AllTests.skipUnknownOptionals(communicator);
             fprintf('ok\n');
 
+            fprintf('testing optional proxies without a proxy type... ');
+            AllTests.bareOptionalProxies(communicator);
+            fprintf('ok\n');
+
             ref = ['initial:', helper.getTestEndpoint()];
             initial = InitialPrx(communicator, ref);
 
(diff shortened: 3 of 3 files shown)


pepone (approved):
Verified the `nargin` accounting: MATLAB counts `obj`, so the untyped call is 2 and the typed one is 3 — matching what the generator emits (`Gen.cpp:3168` and `:3186`). The untyped branch lands on `readProxy`'s `Ice.ObjectPrx` arm, and the old `nargin == 1` was indeed unreachable.

Also checked the two things that could have made this incomplete:

- No other `read*Opt` takes an optional `cls`, and sweeping `nargin` across `matlab/lib` turned up nothing else off by one.
- Returning `Ice.ObjectPrx.empty` rather than `Ice.Unset` for the absent case is the documented mapping contract (`Gen.cpp:175`), and `writeProxyOpt` matches it — so the test asserts the right thing.

Deferring the `MultiOptional` field is the right call: `matlab/test/Ice/optional/Test.ice` has to stay in sync with the Python server's copy. Tag 11 is free.

The `amzn2023-aarch64` failure is the 6h-cap hang in a C++ job, unrelated to this change.

Nit: set the 3.8.3 milestone so this isn't missed for the cherry-pick — `origin/3.8` has identical code at identical lines, so it applies clean.

InsertCreativityHere (approved):
Looks good to me!

PR #6171, opened 2026-07-24
Fix a string-map leak and the Future timed wait in Ice for MATLAB

### LINKED ISSUE #6118: MATLAB low-severity batch: `createStringMap` leaks two string arrays per call; `Future` timed wait uses `system_clock` + unchecked float→duration
> [!NOTE]

## Summary

Two low-severity MATLAB findings.

### 1. `createStringMap` leaks two string arrays per call
`matlab/src/Util.cpp:176-197`: the non-empty path builds `params[0]`/`params[1]` via `cellArrayToString` and passes them to `mexCallMATLAB("dictionary", …)` but never `mxDestroyArray`s them (`mexCallMATLAB` does not take ownership of `prhs`). The empty path (`:170-175`) does destroy them. Callers are frequent getters — `getProperties`, `getPropertiesForPrefix`, WS-header/context getters — so each call leaks both string arrays (all keys + all values).

### 2. `Future` timed wait: wall clock + unchecked float→duration
`matlab/src/Future.cpp:29-35` uses `system_clock` for the deadline (a wall-clock step lengthens/shortens the wait — use `steady_clock`) and `static_cast<long long>(timeout * 1000)` with no validation, so `NaN`/`Inf`/huge `timeout` (forwarded verbatim from `+Ice/Future.m:96`) is UB / overflow.

## Fix

`mxDestroyArray(params[0]); mxDestroyArray(params[1]);` after the `dictionary` call; use `steady_clock` and validate/clamp `timeout` (reject NaN/Inf, saturate large values).

**Severity:** Low.

bernardnormier:
Fixes the two unrelated MATLAB defects reported together in #6118.

Fixes #6118.

### `createStringMap` leaked its arguments

`createStringMap` passed two string arrays to `mexCallMATLAB("dictionary", …)` without destroying them. `mexCallMATLAB` does not take ownership of `prhs`, and Ice for MATLAB is loaded with `loadlibrary` rather than called as a MEX function — so there is no MEX-call boundary for MATLAB to reclaim temporaries at, and the arrays leaked permanently. The empty-map branch already destroyed its argument.

Reachable from `Communicator.proxyToProperty`, `ImplicitContext.getContext`, `ObjectPrx.ice_getContext`, `Properties.getPropertiesForPrefix`, and `Connection.getInfo` on a WebSocket connection.

Measured with `getPropertiesForPrefix` on a three-entry map, reporting `MemUsedMATLAB` growth:

| calls | before | after |
| ----- | ------ | ----- |
| 20,000 | +39.9 MB | +2.7 MB |
| 40,000 | +81.9 MB | +2.7 MB |
| 60,000 | +124.0 MB | +2.7 MB |
| 80,000 | +166.0 MB | +2.1 MB |

Linear at ~2.1 KB per call, versus flat.

Note the issue also names `getProperties`, which does not leak: it returns a `Properties` object, not a dictionary, and never reaches `createStringMap`.

### `Future::waitForState` timed wait

The deadline came from `system_clock`, so a wall-clock adjustment lengthened or shortened the wait; it now uses `steady_clock`. The delay was computed as `static_cast<long long>(timeout * 1000)` with no validation, which is undefined for `NaN`, for infinity, and for values beyond the range of `long long`. An infinite timeout is now an unlimited wait, `NaN` is rejected, and the delay is capped at 1e9 seconds — over 30 years, so indistinguishable from an unlimited wait, but comfortably inside what the conversion and the addition to `now()` can represent.

Checked against the built MEX:

```
timeout 0.5  -> ok=0 elapsed=0.51s
timeout 0    -> ok=0 elapsed=0.00s
timeout NaN  -> Ice:CppException: timeout must be a number
timeout 1e16 -> RETURNED ok=0 elapsed=19.48s   (blocked until the TCP connect failed)
```

The 1e16 case is the one that matters: it saturates and blocks rather than expiring. The timed branch itself is already covered by `Ice/ami`, which suspends the adapter and asserts that `wait('finished', 0.1)` returns false.

### Changelog

Only the leak gets a fragment. The `wait` change is an internal cleanup: the clock swap only bites if the system clock steps during a bounded wait, and the validation only matters for arguments no working program passes.

Full MATLAB suite: 16/16.

### DIFF
--- a/changelog.d/matlab/6118.md
+++ b/changelog.d/matlab/6118.md
@@ -1,0 +1,3 @@
+- Fixed a memory leak in `Communicator.proxyToProperty`, `ImplicitContext.getContext`, `ObjectPrx.ice_getContext`,
+  `Properties.getPropertiesForPrefix`, and `Connection.getInfo` on a WebSocket connection. Each call leaked memory
+  proportional to the size of the dictionary it returned.
--- a/matlab/lib/+Ice/Future.m
+++ b/matlab/lib/+Ice/Future.m
@@ -79,7 +79,8 @@ classdef Future < IceInternal.WrapperObject
             %     successfully or exceptionally.
             %     'running' | 'sent' | 'finished'
             %   timeout - If provided, wait blocks up to the given number of seconds while waiting for the future to
-            %     reach the desired state. If the timeout is negative or not provided, wait blocks indefinitely.
+            %     reach the desired state. If the timeout is negative, infinite, or not provided, wait blocks
+            %     indefinitely. NaN is not a valid timeout value.
             %     double | empty array (default)
             %
             %    Output Arguments
--- a/matlab/src/Future.cpp
+++ b/matlab/src/Future.cpp
@@ -4,6 +4,9 @@
 #include "Util.h"
 #include "ice.h"
 
+#include <algorithm>
+#include <cmath>
+
 using namespace std;
 using namespace IceMatlab;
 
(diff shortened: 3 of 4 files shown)


pepone (approved):
Both fixes look right to me.

The destroy-after-`mexCallMATLAB` pattern matches the rest of the file — `createIdentity`, `createEncodingVersion`, `destroyParams`, and `createStringMap`'s own empty-dictionary branch all do it, so the non-empty branch was the only outlier. I checked the other `mexCallMATLAB` sites: they either pass caller-owned arrays or already destroy their inputs, so this closes the leak class rather than one instance of it. The changelog's list of entry points matches the five call sites, and dropping `getProperties` from the issue's list is correct.

On the timed wait: `-Inf` is absorbed by the existing `timeout < 0` test, and the cap works out to 1e18 ns against `steady_clock`'s int64 range, so the addition can't overflow. Worth noting `steady_clock` is a real fix even on macOS, where the condition variable re-anchors to `CLOCK_REALTIME` internally — the predicate overload re-checks the steady deadline each iteration, so a clock step only causes a spurious wake.

Two small things, neither blocking:

- `Future.m` documents "If the timeout is negative or not provided, wait blocks indefinitely" — `Inf` now does the same and `NaN` is an error. A one-line doc touch would keep it in sync.
- I'd skip the boundary-value tests Copilot asked for. Those are arguments no working program passes, the timed branch is already covered by `Ice/ami`, and asserting `Ice:CppException` on `NaN` would pin an internal error-mapping detail.

Added the 3.8.3 milestone to match #6118.

PR #6172, opened 2026-07-24
Document the per-key current-value state of DataStorm writers

pepone:
Follows up #5661.

A DataStorm writer keeps one current value per key it has written and not since removed — the base that partial updates and late-joining readers resolve against. This is the writer's *current state*, not retained history, so `sampleCount`/`sampleLifetime` (history controls) don't bound it; `remove(key)` retires a key and releases its state on the writer and its connected readers.

This documents that model on `MultiKeyWriter` and clarifies the `remove()` docs on both writer classes. Docs-only, no behavior change.

### DIFF
--- a/cpp/include/DataStorm/DataStorm.h
+++ b/cpp/include/DataStorm/DataStorm.h
@@ -499,6 +499,12 @@ namespace DataStorm
     };
 
     /// The key reader to read the data element associated with a given set of keys.
+    ///
+    /// Like a multi-key writer, a multi-key reader retains the current value of every key it has received and not since
+    /// seen removed, so that it can resolve later partial updates against it. This per-key state is independent of the
+    /// reader's @c sampleCount and @c sampleLifetime history settings, and is released when the key's writer removes
+    /// the key. A reader connected to writers over an unbounded set of keys therefore accumulates one current value per
+    /// key.
     /// @headerfile DataStorm/DataStorm.h
     template<typename Key, typename Value, typename UpdateTag = std::string>
     class MultiKeyReader : public Reader<Key, Value, UpdateTag>
@@ -647,6 +653,11 @@ namespace DataStorm
     }
 
     /// The filtered reader to read data elements whose key match a given filter.
+    ///
+    /// A filtered reader retains the current value of every key it has received and not since seen removed, so that it
+    /// can resolve later partial updates against it. This per-key state is independent of the reader's @c sampleCount
+    /// and @c sampleLifetime history settings, and is released when the key's writer removes the key. A reader matching
+    /// an unbounded set of keys therefore accumulates one current value per key.
     /// @headerfile DataStorm/DataStorm.h
     template<typename Key, typename Value, typename UpdateTag = std::string>
     class FilteredKeyReader : public Reader<Key, Value, UpdateTag>
--- a/cpp/include/DataStorm/DataStorm.h
+++ b/cpp/include/DataStorm/DataStorm.h
@@ -775,7 +775,9 @@ namespace DataStorm
         template<typename UpdateValue>
         [[nodiscard]] std::function<void(const UpdateValue&)> partialUpdate(const UpdateTag& tag);
 
-        /// Removes the data element. This generates a SampleEvent::Remove sample.
+        /// Removes the data element. This generates a SampleEvent::Remove sample and releases the key's current value
+        /// on the writer and on its connected readers, so a later partial update has no value to resolve against and is
+        /// rejected until a new full value is written.
         void remove() noexcept;
 
     private:
@@ -783,6 +785,13 @@ namespace DataStorm
     };
 
     /// The key writer to write data elements associated with a given set of keys.
+    ///
+    /// A multi-key writer retains the current value of every key it has written and not since removed, so that later
+    /// partial updates and late-joining readers can resolve against it. This per-key state is the writer's current data
+    /// set, not retained history: it is independent of the @c sampleCount and @c sampleLifetime history settings, which
+    /// bound the retained samples but never the current value of a live key. An open-key writer (one constructed with
+    /// an empty key vector) that writes to an unbounded set of keys therefore accumulates one current value per key;
+    /// call remove(const Key&) to retire a key and release its state once the key is no longer in use.
     /// @headerfile DataStorm/DataStorm.h
     template<typename Key, typename Value, typename UpdateTag = std::string>
     class MultiKeyWriter : public Writer<Key, Value, UpdateTag>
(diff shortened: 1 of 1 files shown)


cpp/include/DataStorm/DataStorm.h
```diff
@@ -499,6 +499,12 @@ namespace DataStorm
     };
 
     /// The key reader to read the data element associated with a given set of keys.
+    ///
+    /// Like a multi-key writer, a multi-key reader retains the current value of every key it has received and not since
+    /// seen removed, so that it can resolve later partial updates against it. This per-key state is independent of the
+    /// reader's @c sampleCount and @c sampleLifetime history settings, and is released when the key's writer removes
```
bernardnormier: Doc-comments in this file (and the Ice headers) use markdown backticks rather than `@c`:
```suggestion
    /// reader's `sampleCount` and `sampleLifetime` history settings, and is released when the key's writer removes
```

cpp/include/DataStorm/DataStorm.h
```diff
@@ -647,6 +653,11 @@ namespace DataStorm
     }
 
     /// The filtered reader to read data elements whose key match a given filter.
+    ///
+    /// A filtered reader retains the current value of every key it has received and not since seen removed, so that it
+    /// can resolve later partial updates against it. This per-key state is independent of the reader's @c sampleCount
+    /// and @c sampleLifetime history settings, and is released when the key's writer removes the key. A reader matching
```
bernardnormier:
```suggestion
    /// can resolve later partial updates against it. This per-key state is independent of the reader's `sampleCount`
    /// and `sampleLifetime` history settings, and is released when the key's writer removes the key. A reader matching
```

cpp/include/DataStorm/DataStorm.h
```diff
@@ -775,14 +786,23 @@ namespace DataStorm
...
     private:
         const std::shared_ptr<DataStormI::TagFactoryT<UpdateTag>> _tagFactory;
     };
 
     /// The key writer to write data elements associated with a given set of keys.
+    ///
+    /// A multi-key writer retains the current value of every key it has written and not since removed, so that later
+    /// partial updates and late-joining readers can resolve against it. This per-key state is the writer's current data
+    /// set, not retained history: it is independent of the @c sampleCount and @c sampleLifetime history settings, which
+    /// bound the retained samples but never the current value of a live key. An open-key writer (one constructed with
```
bernardnormier: "open-key writer" is new terminology — the established name in this file is *any-key* (`makeAnyKeyWriter`, "Creates an any-key writer..."). Plus the backtick conversion:
```suggestion
    /// set, not retained history: it is independent of the `sampleCount` and `sampleLifetime` history settings, which
    /// bound the retained samples but never the current value of a live key. An any-key writer (one constructed with
```

cpp/include/DataStorm/DataStorm.h
```diff
@@ -832,7 +852,9 @@ namespace DataStorm
         template<typename UpdateValue>
         [[nodiscard]] std::function<void(const Key&, const UpdateValue&)> partialUpdate(const UpdateTag& tag);
 
-        /// Removes the data element. This generates a TopicEvent::Remove sample.
+        /// Removes the data element. This generates a TopicEvent::Remove sample and retires the key: its current value
```
bernardnormier: `TopicEvent` doesn't exist — the enum is `SampleEvent` (pre-existing typo, but this PR rewrites the line):
```suggestion
        /// Removes the data element. This generates a SampleEvent::Remove sample and retires the key: its current value
```

bernardnormier (commented):
The model documented here matches the implementation (writer-side `std::logic_error`, reader-side discard, per-key base independent of the history settings). A few wording fixes below.

cpp/include/DataStorm/DataStorm.h
```diff
@@ -499,6 +499,12 @@ namespace DataStorm
     };
 
     /// The key reader to read the data element associated with a given set of keys.
+    ///
+    /// Like a multi-key writer, a multi-key reader retains the current value of every key it has received and not since
```
externl: Nit: I don't think we should document the behavior of  MultiKeyReader based on MultiKeyWriter, especially since it comes after it in the file.

externl (approved):

cpp/include/DataStorm/DataStorm.h
pepone: 🤖 **Claude:** Done in c5f6551. Note the same line also changed "when the key's writer removes the key" to "when the reader receives the key's remove sample" — see the Copilot thread.

cpp/include/DataStorm/DataStorm.h
pepone: 🤖 **Claude:** Done in c5f6551. Note the same line also changed "when the key's writer removes the key" to "when the reader receives the key's remove sample" — see the Copilot thread.

cpp/include/DataStorm/DataStorm.h
pepone: 🤖 **Claude:** Done in c5f6551 — any-key it is.

cpp/include/DataStorm/DataStorm.h
pepone: 🤖 **Claude:** Done in c5f6551.

cpp/include/DataStorm/DataStorm.h
pepone: 🤖 **Claude:** Agreed — done in c5f6551. The paragraph now stands on its own, with no reference to `MultiKeyWriter`.

cpp/include/DataStorm/DataStorm.h
```diff
@@ -499,6 +499,12 @@ namespace DataStorm
...
+    /// seen removed, so that it can resolve later partial updates against it. This per-key state is independent of the
+    /// reader's @c sampleCount and @c sampleLifetime history settings, and is released when the key's writer removes
+    /// the key. A reader connected to writers over an unbounded set of keys therefore accumulates one current value per
+    /// key.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Confirmed and fixed in c5f6551. Sample filters are enforced writer-side only (`KeyDataWriterI::forward` → `Listener::matchOne`); `DataReaderI::queue` never re-applies them, so a reader whose `_event` filter excludes `Remove` keeps its per-key base indefinitely.

Reworded to tie the release to receipt rather than adding a filter caveat: readers now say "released when the reader receives the key's remove sample", and both `remove()` docs say "on the readers that receive the sample".

One correction: the gate is per-session, not per-reader — if another subscriber on the same session matches the remove, it is forwarded and the filtered reader clears its base too. So the retention is conditional.

bernardnormier (approved):
The reworked wording addresses all the threads — thanks.

PR #6173, opened 2026-07-24
Test DataStorm partial updates after reconnect

pepone:
## Summary

- force a session reconnect after the reader consumes a full value
- publish a partial update without resending the full value
- verify the reader retains its pre-disconnect base and receives no duplicate initialization sample
- follow the partial with a full sentinel so a dropped partial fails immediately instead of hanging

Related to #5661.

## Testing

- `make -C cpp -j8`
- `python3 allTests.py DataStorm/reliability` (all 16 topology variants passed)

### DIFF
--- a/cpp/msbuild/ice.test.slnx
+++ b/cpp/msbuild/ice.test.slnx
@@ -52,6 +52,14 @@
       <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
     </Project>
   </Folder>
+  <Folder Name="/DataStorm/partialReconnect/">
+    <Project Path="../test/DataStorm/partialReconnect/msbuild/reader/reader.vcxproj">
+      <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
+    </Project>
+    <Project Path="../test/DataStorm/partialReconnect/msbuild/writer/writer.vcxproj">
+      <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
+    </Project>
+  </Folder>
   <Folder Name="/DataStorm/relayReconnect/">
     <Project Path="../test/DataStorm/relayReconnect/msbuild/reader/reader.vcxproj">
       <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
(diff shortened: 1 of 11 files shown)


externl (approved):

bernardnormier (approved):
Coverage note, non-blocking: this pins the no-gap resume (nothing published during the outage). The other half of the reconnect contract — a reader with an established base reconnecting across a *gap* (writer history no longer covers the outage) and getting re-seeded with a full-value `Update` instead of a replay — has no test either here or in `DataStorm/partial` (those scenarios all join fresh with `lastId = 0`). Worth a companion scenario, here or in `partial`, e.g. a `sampleCount=0` writer publishing a full update + partial during the outage, with the reader asserting one full `Update` carrying the resolved value followed by a live partial.

pepone:
Added the companion scenario, but as its own suite (`cpp/test/DataStorm/partialReconnect`) rather than here or in `partial`.

The gap needs samples published while the session is down, and only the writer can do that — but a writer can't reach its own data session's connection. `Sample::getSession()` is the only handle and a writer never receives its own samples, so the best it can do is close the connection behind a *barrier* session. That works only where both happen to share one: in `reliability`'s 2-relay-node case the writer holds connections to both relays, and in `partial`'s multicast case the reader connects back, so in both the data session keeps delivering and the reader gets the two originals live instead.

Reversing the roles doesn't help. The reader can identify the right connection, but then the writer has no sound trigger to publish inside the outage — the first retry is immediate (`SessionI.cpp:867`) and `waitForNoReaders()` can miss a 1→0→1 transition and block (`DataElementI.cpp:573-585`). And there is no gap without a real outage: `lastId` advances on receipt, before any discard (`SessionI.cpp:1628`).

So the new suite pins the topology — multicast off, writer as client with no server endpoint, one connection. A `sampleCount=0` writer publishes a full value plus a partial into the gap; the reader asserts a single `Update` carrying the resolved value, then a live partial resolving against it. 50/50 consecutive clean runs, and all 11 DataStorm suites pass.

One caveat worth knowing: if the re-seed regresses, the reader simply never receives it, so the failure surfaces as a suite timeout rather than an assertion — there is no timed read in the reader API. The existing `partial` late-joiner scenarios behave the same way.

PR #6175, opened 2026-07-24
Decode DataStorm full values with an empty custom encoding

### LINKED ISSUE #6084: DataStorm: SampleT::decode skips empty encoded values, making them unusable as partial-update bases
> Surfaced by a Codex cross-check during the review of #6083. No crash is possible—#6083's guards make the
> affected samples inert—but the resulting behavior deviates from the documented `Decoder` contract.

## Summary

`SampleT::decode` skips decoding when the encoded value is empty, leaving the sample value-less
(`_hasValue == false`):

https://github.com/zeroc-ice/ice/blob/main/cpp/include/DataStorm/InternalT.h#L396-L404

This is inconsistent with the `Decoder` contract: the default `DataStorm::Decoder<T>::decode` explicitly defines
empty input as a default-constructed value (`Types.h`, `v = T()`), and custom `Decoder` specializations never get
the chance to apply their own convention because `SampleT::decode` never calls them for empty input.

The default Ice encoding never produces an empty byte sequence, so this only affects topics using a custom
`Encoder` specialization that encodes some values to zero bytes—a pattern the `Encoder`/`Decoder` extension
points explicitly support (e.g. a compact encoding where the default value is represented by no bytes).

## Effect (since #6083)

A received full `Add`/`Update` whose custom encoding is empty:

- reports `hasValue() == false`, and
- is rejected as a partial-update base by the reader's `hasValue()` guard, so subsequent partial updates on the
  key are silently discarded (trace: "no base value for the key") even though the writer legitimately published a
  full value for the key.

`getValue()` happens to return the right value for the default `Decoder` (the sample value is value-initialized),
but a custom `Decoder` specialization mapping empty input to something other than `T()` is ignored.

## Proposed fix

In `SampleT::decode`, decode full samples even when `_encodedValue` is empty, marking the sample value-bearing
(`_hasValue = true`) and letting the `Decoder` (default or specialized) define the empty-input semantics. Keep the
#6083 behavior of not decoding `Remove` samples. Such samples then work as partial-update bases like any other
full value.

pepone:
A full `Add`/`Update` value that a custom `Encoder` encodes to an empty byte sequence was left value-less by `SampleT::decode`, so it could not serve as a base for a subsequent partial update — the reader discarded it for want of a base (trace: "no base value for the key"). Decode such values through their `Decoder` like any other full value, letting the `Decoder` define the meaning of empty input. Remove samples remain value-less.

Only custom `Encoder` specializations that emit zero bytes are affected; the default Ice encoding never produces an empty sequence. Follow-up to #6083, deliberately deferred from it.

Includes a `partial` test (`Counter`, whose default value custom-encodes to zero bytes) that hangs without the fix and passes with it.

Fixes #6084.


externl (approved):

cpp/include/DataStorm/InternalT.h
```diff
@@ -393,12 +393,19 @@ namespace DataStormI
...
             {
-                _hasValue = true;
-                _value = DecoderT<Value>::decode(communicator, _encodedValue);
-                _encodedValue.clear();
+                return;
             }
+
+            // Decode full samples even when the encoded value is empty. A custom Encoder specialization may encode
+            // a value to zero bytes, and the Decoder (default or specialized) defines the meaning of empty input;
+            // the resulting value-bearing sample can then serve as a partial-update base like any other full value.
```
bernardnormier: These comments mostly explain the old bug and the callers — the bug history is already in the PR description. Two one-liners stating the local facts are enough:
```suggestion
            // A remove sample carries no value.
            if (event == DataStorm::SampleEvent::Remove)
            {
                return;
            }

            // A custom Encoder may encode a value to zero bytes; the Decoder defines the meaning of empty input.
```

bernardnormier (approved):
The fix is right: every `sample->decode()` call site already skips `PartialUpdate` and `Remove`, so decoding unconditionally only affects full values, and the empty sequence is a legal encoding — the framing lives in `DataSample::value`, one level up from the codec. Just one suggestion to slim down the comments.

changelog.d/datastorm/6084.md
```diff
@@ -0,0 +1,4 @@
+- Fixed a bug where a full value that a custom `Encoder` encodes to an empty byte sequence was decoded as
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Correct — this PR fixes the single public issue #6084, so the fragment takes the issue number. Renamed to `changelog.d/datastorm/6084.md` (content unchanged) in 39c10f4533.

cpp/include/DataStorm/InternalT.h
pepone: 🤖 **Claude:** Agreed — the bug history and the caller argument belong in the PR description, not the code. Applied your suggestion verbatim in 39c10f4533.

PR #6176, opened 2026-07-24
Clean up the test scripts after the type hints pass

externl:
Finishes the type hints started in #6158: pyright now checks every file under `scripts/`, and the ignore list in `pyrightconfig.json` is gone.

- Type hint the remaining framework — `LocalDriver.py`, `Controller.py`, `NetworkProxy.py` — the shared test definitions under `scripts/tests`, and the standalone tools.
- Use real Ice types instead of `Any` for the communicator, the object adapter and the process-controller proxy map, via a `TYPE_CHECKING` import.
- Take a `Sequence` rather than a `list` for the test cases, clients, servers and output filters a caller passes in: `list` is invariant, so a `list[ClientServerTestCase]` could not be passed for a `list[TestCase]`.
- Add `Mapping.requireByPath` for the test definitions, and `LocalDriver.crossMappings` for the two places that built the cross testing mapping list by hand.
- Fix signatures that did not describe the code: `BaseProxy.createConnection` took no arguments although it is called with two, `BaseConnection.request` and `response` returned `None` although callers use the result, `Driver.run` returns `int | None`, and `IceGridTestCase`'s application is optional.
- Fix `NetworkProxy.terminate`, which raised `NameError` from its `finally` rather than closing the socket if one could not be created, and `BaseProxy.run`, which skipped `notify()` when there was no socket to close and so left the constructor waiting out its full 60 second timeout.
- Validate in `ControllerDriver.getCurrent` what `Current` actually requires — the server side resolved against the cross mapping, and the client side — rather than the server side of the test case's own mapping.
- Remove the unused `EOF` class, an empty branch in `splitCommand`, and the leftover Python 2 branches in `NetworkProxy`.
- Remove unused imports from the Python sources — `Ice/Util.py`, `Ice/_LoggerI.py` and four tests. Nothing checks for these: `ruff.toml` selects only the isort rules, and pyright's `reportUnusedImport` cannot be enabled until slice2py stops emitting imports its output does not use (#6225).
- Name the first parameter of the class methods `cls`.


pepone (commented):
Nice finishing pass. One small fix, then LGTM.

I checked this doesn't change test behavior: enumerated every mapping's test suites, test cases, resolved client/server processes and outfilter counts on `4a4621edd0` vs this branch — 927 lines, identical. Pyright reports 0 errors under `scripts/`, ruff `check`/`format` clean.

The two deltas that looked risky are both fine:

- `runClientServerTestCase`: `[""]` → `[None]` is safe, since `getServerTestCase` does `cross if isinstance(cross, Mapping) else self.getMapping()`, so `""` and `None` are interchangeable, and both `if cross` guards became `is not None` consistently.
- `executor.submit(..., [""] → [])`: `submit` iterates `for cross in crossMappings: if cross:`, so the falsy entry was already a no-op.

## The one thing to fix

`Controller.py`, `ControllerDriver.Current.__init__` — the assert's justification isn't right:

```python
serverTestCase = self.testcase.getServerTestCase(cross)
clientTestCase = self.testcase.getClientTestCase()
# getCurrent rejects a test case without a server side before building a Current.
assert serverTestCase is not None and clientTestCase is not None
```

`getCurrent` validates `tc.getServerTestCase()` with **no cross argument**, i.e. against the test case's own mapping. `Current.__init__` then calls `getServerTestCase(crossMapping)`, which resolves against the cross mapping and can legitimately be `None` — that's exactly the case `LocalDriver.runClientServerTestCase` handles with `if not server: continue`. And `getCurrent` never checks `getClientTestCase()` at all, yet the assert covers it.

When it does fire, the `AssertionError` escapes `ControllerI.runTestCase` and the driver gets an `UnknownException` instead of the declared `TestCaseNotExistException`. Before this PR the same situation gave `None._startServerSide(...)` → `AttributeError` inside `TestCaseI.startServerSide`, caught and reported as `TestCaseFailedException` with a traceback. Narrow path (LocalDriver filters `if not server: continue` before calling the remote controller, so it needs the remote's mapping set to differ), but it's a step back in error reporting.

Fix is one line in `getCurrent` — validate what `Current` actually requires:

```python
if not tc or not tc.getServerTestCase(crossMapping) or not tc.getClientTestCase():
    raise Test_Common.TestCaseNotExistException("unknown testcase {0}".format(testcase))
```

Then the assert is justified.

## Worth keeping

- `NetworkProxy.terminate`: the old `finally: connectToSelf.close()` raised `UnboundLocalError` when `socket()` failed, masking the real error.
- `BaseProxy.run`: if `socket.socket()` itself threw, `self.socket.close()` raised inside the `except` and `cond.notify()` never ran, so the constructor sat out its full 60s wait before reporting. Both real fixes.
- The `list()` copies for `outfilters`/`clients`/`servers` are required once the params are `Sequence`, and they close a latent footgun: `Server.__init__` does `self.outfilters.append(re.compile("[^\n]+ ready"))`, which under the old aliasing mutated the caller's list. Not a live bug today (each shared list feeds one process per module execution, and `loadTestSuites` uses `runpy.run_module` so the module re-executes per mapping), but good to have.
- The `"."` extraPath is justified — `scripts/tests/IceSSL/configuration.py` does `from scripts.tests.IceSSL import revocationutil`, and `loadTestSuites` puts the repo root on `sys.path`. Comment matches the code.

## Nits

- `getSupportedArgs`/`usage` in both `Controller.py` and `LocalDriver.py`, plus `LocalDriver.isWorkerThread`/`isInterrupted`, are the only non-`__init__` defs left without return types in the newly-checked files. Pyright lets them slide because the base declares them, but the `Util.py` equivalents are annotated.
- `runClientServerTestCase`/`runTestCase` introduce a local `testcase = current.getTestCase()` but the `writeln` still formats `current.testcase`. Use the local one.
- `ControllerDriver.Current` could redeclare `serverTestCase: TestCase`. The base declares it `Any`, so the subclass keeps `Any` and the `TestCaseI._startServerSide`/`_stopServerSide` calls go unchecked — the `assert ... is not None` buys nothing type-wise as it stands.
- `_serverIndex(process) -> object` in `tests/IceDiscovery/simple.py`: `object` is an odd return type for something only fed to `.format()`; `str(process.args[0]) -> str` reads better. Also inconsistent with `faultTolerance.py`, which inlines the same `assert not callable(process.args)` without a helper — pick one shape.
- Drive-by while you're in there: `revocationutil.py` still uses the deprecated `self.thread.setDaemon(True)`; `self.thread.daemon = True` is the current spelling.

externl:
Fixed in dfb7282aa37 — took the suggestion as written. Enumerating every reachable test case found 237 mapping/cross combinations where the own-mapping server side resolves but the cross one is `None`, so the assert really could fire; the added `getClientTestCase()` check rejects nothing today (0 of 179).

Note `LocalDriver.startServerSide` calls `runTestCase` outside its `try`, so `TestCaseNotExistException` isn't caught there — that case surfaces as a recorded failure with a traceback rather than `RuntimeError("test failed")`. Happy to catch it too if you'd rather.

Nits all in 7f4b76d760a. `_serverIndex` is gone, which restores main's original expression verbatim.

pepone (approved):
All addressed, LGTM.

Nice catch on the protocol one — I'd looked at the guard and stopped at "updateCurrent overwrites it anyway", and missed that the shared config's "tcp" default gets published as Ice.Default.Protocol where main published nothing. `protocol or ""` is the right shape; every reader only tests it for truthiness or equality, so it's indistinguishable from the None main stored.

> Happy to catch it too if you'd rather.

Leave it. `TestCaseNotExistException` means the driver and controller disagree about which test cases exist — harness misconfiguration, not a test failure. Collapsing it into `RuntimeError("test failed")` would hide the exception type, and the recorded traceback names it. Moving `runTestCase` inside that `try` would also break the `except Exception: current.serverTestCase.destroy()` cleanup, which assumes the proxy exists. Nothing leaks as it stands: no server was started, and `serverTestCase` keeps the `None` the base now initializes.

> The rule can be moved into the scripts execution environment if the python/ half is unwanted.

Keep it repo-wide. Scoping a hygiene rule to one directory invites the same miss on the python/ side — and that miss is the whole point here, since Copilot's stated reason for the `IO` finding was wrong: `ruff.toml` selects only `I`, so F401 never fires. A file with an unused import gets `I001` and nothing else. Fixing the gap beats fixing the instance.

I checked the one library-code removal: `AsyncIOEventLoopAdapter` is used only by `Communicator.py`, which has its own import, isn't re-exported, and the module still loads via the adjacent `.Communicator` import — no registration side effect lost.

Re-verified on 8caf54521f: pyright clean under `scripts/`, ruff clean, and the test enumeration (suites, cases, resolved processes, outfilter counts) is identical to cd3d0493f8.

externl:
Backed `reportUnusedImport` out again in ffec11ed7a4. With the Slice code built it reports 235 errors across 113 files, all generated — slice2py emits `IcePy`, `TYPE_CHECKING` and the `_Xxx_t` type markers whether or not a fragment uses them. I had measured the rule in a tree with no generated code, which is exactly the population that breaks it.

Opened #6225 for the generator work. Also reverted the `python/` import removals the rule had found, so this is back to `pyrightconfig.json` plus `scripts/` only.

PR #6178, opened 2026-07-24
Avoid pointer arithmetic past the end of the buffer in InputStream bounds checks

bernardnormier:
`InputStream::skip` and the four encapsulation bounds checks (`startEncapsulation`, `skipEmptyEncapsulation`, `readEncapsulation`, `skipEncapsulation`) computed `i + size` — with a peer-controlled size of up to `INT32_MAX` — before comparing the result to `b.end()`. Forming a pointer past one-past-the-end is undefined behavior.

In practice the check still throws as intended on every configuration we build: 64-bit platforms can't wrap, and on 32-bit Windows the addition could only wrap with a buffer above `0x80000001`, which requires a large-address-aware process — our Win32 builds are not, so the address space tops out at 2GB. The fix removes undefined behavior rather than a reachable out-of-bounds read.

This PR rewrites these five checks to compare sizes using pointer subtraction (`b.end() - i`), the well-defined idiom already used by all the other bounds checks in `InputStream`. Behavior is unchanged: the same inputs throw the same `MarshalException`.

### DIFF
--- a/cpp/include/Ice/InputStream.h
+++ b/cpp/include/Ice/InputStream.h
@@ -625,7 +625,7 @@ namespace Ice
         /// @param size The number of bytes to skip.
         void skip(size_type size)
         {
-            if (i + size > b.end())
+            if (size > static_cast<size_type>(b.end() - i))
             {
                 throwUnmarshalOutOfBoundsException(__FILE__, __LINE__);
             }
--- a/cpp/src/Ice/InputStream.cpp
+++ b/cpp/src/Ice/InputStream.cpp
@@ -243,7 +243,8 @@ Ice::InputStream::startEncapsulation()
     {
         throwUnmarshalOutOfBoundsException(__FILE__, __LINE__);
     }
-    if (i - sizeof(std::int32_t) + sz > b.end())
+    // sz includes the 4 bytes of the size itself, which we've already read.
+    if (sz - 4 > b.end() - i)
     {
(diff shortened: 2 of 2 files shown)


pepone (approved):
LGTM.

Verified the rewrite is equivalent to the original on 64-bit for all buffer sizes and `sz` up to `INT32_MAX`, and that `sz >= 6` (checked just above) keeps `sz - 4` from underflowing. Both operands stay signed (`int32_t` widened to `ptrdiff_t`), so there's no wrap on either 32- or 64-bit. `git grep` confirms these were the only 5 sites left using pointer addition — the file is now uniform on the `b.end() - i` idiom.

One nit on the description rather than the code: wrapping needs `i - 4 + sz >= 2^32`, which requires the buffer above `0x80000001` — i.e. a 32-bit process with a >2GB user address space. We don't set `LargeAddressAware` and MSVC defaults x86 to `/LARGEADDRESSAWARE:NO`, so the old check still throws on the Win32 configuration we ship and test. Worth rewording so this doesn't read as a live OOB read; it's UB and worth fixing regardless.

Re the two Copilot comments, I'd skip both:

- A test can't be red→green anywhere our CI runs — 64-bit behavior is unchanged by construction, and Win32 CI has a 2GB address space where the old code also throws.
- No user-visible behavior change on any default-configured platform, and `changelog.d/README.md` says such changes get no fragment.

PR #6179, opened 2026-07-24
Validate the invocation timeout passed to ice_invocationTimeout in C#

### LINKED ISSUE #6102: C# invocation-timeout `TimeSpan` narrows `long`→`int` in the timer loop → timer-thread crash / process termination
> [!NOTE]

## Summary

`ice_invocationTimeout(TimeSpan)` accepts an arbitrary `TimeSpan` with no validation and schedules it on the Ice `Timer`, which stores the deadline as `long` ms but then narrows the wait to `int`: `Monitor.Wait(_mutex, (int)(first.scheduledTime - now))`. For a timeout above `Int32.MaxValue` ms (~24.9 days) the cast wraps negative; `Monitor.Wait` throws `ArgumentOutOfRangeException` for any negative value other than `-1`. That throw is **not** inside the `Timer.Run` task try/catch — it escapes the bare timer thread, and in .NET an unhandled exception on any thread **terminates the process** (taking the invocation-timeout and RetryQueue timers with it). It fires immediately on invoke, not after 24.9 days.

**Files:** `csharp/src/Ice/Internal/Timer.cs:189-190`
```csharp
_wakeUpTime = first.scheduledTime;
Monitor.Wait(_mutex, (int)(first.scheduledTime - now));   // long -> int narrowing, inside lock, no try
```
scheduled via `csharp/src/Ice/Internal/OutgoingAsync.cs:469-473` (`(long)invocationTimeout.TotalMilliseconds`); public entry `csharp/src/Ice/Proxy.cs:1047-1057` / `Reference.cs:141-146` (no range check). Secondary modes: a delta of exactly `0xFFFFFFFF` casts to `-1` (`Timeout.Infinite`) and silently stalls the timer; a positive sub-millisecond timeout truncates to `0` and times out immediately.

## Fix

Wait in bounded chunks no larger than `Int32.MaxValue`; validate/clamp the supported range at `ice_invocationTimeout`; define sub-millisecond behavior. (Proxy properties parse the timeout as `int`, so this is reachable only through the `TimeSpan` API, not remotely.)

**Severity:** Medium (local-API-only, but a valid-looking API call terminates the whole process).

bernardnormier:
In C#, `ice_invocationTimeout(TimeSpan)` accepted any `TimeSpan` value. For a timeout greater than `int.MaxValue` milliseconds (about 24.9 days), the Ice timer thread narrowed the wait delay to `int` in `Monitor.Wait(_mutex, (int)(first.scheduledTime - now))` once that deadline became the earliest pending one; the wrapped negative value made `Monitor.Wait` throw `ArgumentOutOfRangeException` outside any try block, and the unhandled exception on the timer thread terminated the process.

This PR makes two changes:

- `ice_invocationTimeout(TimeSpan)` now throws `ArgumentOutOfRangeException` for timeouts outside the range of an `int` in milliseconds, and rounds fractional timeouts down to a whole number of milliseconds, so the stored timeout always matches the applied one (a sub-millisecond timeout previously scheduled with a 0 ms delay and timed out immediately). Values less than or equal to zero keep their existing "no timeout" semantics, consistent with the other language mappings and the existing tests.
- The invocation timeout is now stored internally as an `int` in milliseconds (`invocationTimeoutMs`) instead of a `TimeSpan`, matching how it is parsed from properties, compared, stringified, and scheduled on the timer. The public API is unchanged: `ice_invocationTimeout(int)`, `ice_invocationTimeout(TimeSpan)`, and `ice_getInvocationTimeout()` keep their signatures, with the `TimeSpan` overload converting once at the API boundary.

Fixes #6102.

### DIFF
--- a/changelog.d/csharp/6102.md
+++ b/changelog.d/csharp/6102.md
@@ -1,0 +1,4 @@
+- Fixed `ice_invocationTimeout(TimeSpan)`: setting an invocation timeout greater than `int.MaxValue` milliseconds
+  (about 24.9 days) could crash the process with an unhandled exception in the Ice timer thread. Timeouts outside
+  the range of an `int` in milliseconds are now rejected with an `ArgumentOutOfRangeException`, and fractional
+  timeouts are rounded down to a whole number of milliseconds.
--- a/csharp/src/Ice/Internal/CollocatedRequestHandler.cs
+++ b/csharp/src/Ice/Internal/CollocatedRequestHandler.cs
@@ -87,7 +87,7 @@ public class CollocatedRequestHandler : RequestHandler
             }
 
             outAsync.attachCollocatedObserver(_adapter, requestId);
-            if (!synchronous || !_response || _reference.getInvocationTimeout() > TimeSpan.Zero)
+            if (!synchronous || !_response || _reference.getInvocationTimeoutMs() > 0)
             {
                 // Don't invoke from the user thread if async or invocation timeout is set
                 _adapter.getThreadPool().execute(
--- a/csharp/src/Ice/Internal/DefaultsAndOverrides.cs
+++ b/csharp/src/Ice/Internal/DefaultsAndOverrides.cs
@@ -82,7 +82,7 @@ public sealed class DefaultsAndOverrides
             {
                 throw new InitializationException($"invalid value for Ice.Default.InvocationTimeout: {value}");
             }
-            defaultInvocationTimeout = TimeSpan.FromMilliseconds(value);
+            defaultInvocationTimeoutMs = value;
         }
 
         val = properties.getIceProperty("Ice.Default.EncodingVersion");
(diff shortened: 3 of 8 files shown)


externl (approved):
Could we keep `TimeSpan` as the internal representation and just validate at the API boundary? `ice_invocationTimeout(TimeSpan)` would throw for out-of-range values and normalize to whole milliseconds (`TimeSpan.FromMilliseconds((long)newTimeout.TotalMilliseconds)`) — that fixes both the crash and the sub-millisecond case, and leaves `Reference`, `ReferenceFactory`, `DefaultsAndOverrides`, `CollocatedRequestHandler`, and `OutgoingAsync` untouched.

Either way, a positive sub-millisecond timeout now means "no timeout" instead of timing out immediately. That matches the C++ mapping, but it silently drops a timeout the caller asked for — worth stating in the doc comment ("a timeout of less than one millisecond is equivalent to no timeout") rather than leaving it implied by "rounded down".

Not for this PR: the narrowing at `Timer.cs:190` is still there, and `Timer.schedule` still takes a `long`, so a future caller could re-arm the same crash.

bernardnormier:
> Could we keep `TimeSpan` as the internal representation and just validate at the API boundary?

This was my initial fix but I found it inelegant. All these TimeSpan in the internal API just to cast it to milliseconds in the end. I find the new version much cleaner. A longer term fix could be to replace this internal Timer by a native Timer that accepts TimeSpan. Definitely out of the scope of this fix.

bernardnormier:
> Not for this PR: the narrowing at `Timer.cs:190` is still there, and `Timer.schedule` still takes a `long`, so a future caller could re-arm the same crash.

What do you mean?

bernardnormier:
Closing: will rework on a separate branch / PR.

PR #6181, opened 2026-07-24
Read connection-related properties at communicator initialization

### LINKED ISSUE #6103: C++ `ConnectionI` constructor is `noexcept` but reads throwing integer property parsers → `std::terminate`
> [!NOTE]

## Summary

The `ConnectionI` constructor is declared `noexcept` but reads mutable integer properties (`Ice.Warn.Connections`, `Ice.Warn.Datagrams`, `Ice.Compression.Level`) whose parser throws `PropertyException` on non-integer text. An exception escaping a `noexcept` function calls `std::terminate`. Property *values* are validated lazily at read time, not at set time, so a config typo or a runtime `setProperty` deterministically aborts the process at the next connection. On a pure client none of these three is read during `initialize()`, so a bad value passes init and kills the process at the first connection. (Same family as #6033 and #6058.)

**File:** `cpp/src/Ice/ConnectionI.cpp:1881-1911` (ctor `noexcept`)
```cpp
_warn(_instance->initializationData().properties->getIcePropertyAsInt("Ice.Warn.Connections") > 0),   // :1900
_warnUdp(_instance->initializationData().properties->getIcePropertyAsInt("Ice.Warn.Datagrams") > 0),   // :1901
...
compressionLevel = properties->getIcePropertyAsInt("Ice.Compression.Level");                            // :1911
```
`getPropertyAsIntWithDefault` throws `PropertyException` on invalid text (`cpp/src/Ice/Properties.cpp:266-286`). A throwing `Ice.UDP.SndSize` parse is also reachable from the same ctor via `BatchRequestQueue` (`BatchRequestQueue.cpp:57`); and user observer/logger calls inside `ConnectionI::setState` (`:2191, 2195`) sit outside its try/catch and are reached from six `noexcept` callbacks.

## Fix

Remove `noexcept` (and let creation fail with a catchable exception), or cache/validate these values at initialization so the reads cannot throw.

**Severity:** Medium (deterministic `std::terminate` from a plausible operator typo / runtime `setProperty`; not attacker-reachable).

bernardnormier:
The `ConnectionI` constructor read `Ice.Warn.Connections`, `Ice.Warn.Datagrams`, and `Ice.Compression.Level`, and the `BatchRequestQueue` it creates read `Ice.UDP.SndSize`. Property values are validated lazily at read time, so a malformed value (for example `Ice.Warn.Connections=x`) passed communicator initialization and then made these reads throw at the first connection. In C++ the `ConnectionI` constructor is `noexcept`, so the escaping `PropertyException` aborted the process with `std::terminate`; in the other mappings the failure surfaced as a confusing per-connection exception.

`Instance` now reads, validates, and caches these values during communicator initialization — following the existing pattern for `Ice.MessageSizeMax` and `Ice.BatchAutoFlushSize` — and the connection code uses the cached values. A malformed value now fails `initialize()` with a clear exception, in every mapping, before any connection is established. The out-of-range handling of `Ice.Compression.Level` (clamp to [1, 9]) is unchanged; only where the properties are read has moved.

This change applies to C++, C#, Java, and JS (which only reads `Ice.Warn.Connections`). As a side effect, changing one of these properties at run time with `setProperty` no longer affects subsequently created connections — the value set at initialization is used for the communicator's lifetime, consistently across mappings.

Fixes #6103.

### DIFF
--- a/cpp/src/Ice/BatchRequestQueue.cpp
+++ b/cpp/src/Ice/BatchRequestQueue.cpp
@@ -12,8 +12,6 @@ using namespace IceInternal;
 
 namespace
 {
-    const int udpOverhead = 20 + 8;
-
     class BatchRequestI final : public Ice::BatchRequest
     {
     public:
@@ -53,8 +51,7 @@ BatchRequestQueue::BatchRequestQueue(const InstancePtr& instance, bool datagram)
     _maxSize = instance->batchAutoFlushSize();
     if (_maxSize > 0 && datagram)
     {
-        const Ice::InitializationData& initData = instance->initializationData();
-        int32_t udpSndSize = initData.properties->getPropertyAsIntWithDefault("Ice.UDP.SndSize", 65535 - udpOverhead);
+        int32_t udpSndSize = instance->udpSndSize();
         if (udpSndSize < _maxSize)
         {
             _maxSize = udpSndSize;
--- a/cpp/src/Ice/ConnectionFactory.cpp
+++ b/cpp/src/Ice/ConnectionFactory.cpp
@@ -1496,7 +1496,7 @@ IceInternal::IncomingConnectionFactory::IncomingConnectionFactory(
               : instance->initializationData().properties->getPropertyAsInt(adapter->getName() + ".MaxConnections")),
       _endpoint(endpoint),
       _adapter(adapter),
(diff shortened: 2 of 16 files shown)


bernardnormier:
Closing: we should not change when these properties are read in a patch release, especially for such a minor bug.

PR #6183, opened 2026-07-24
Validate the delay passed to the C# timer and the invocation timeout

### LINKED ISSUE #6102: C# invocation-timeout `TimeSpan` narrows `long`→`int` in the timer loop → timer-thread crash / process termination

bernardnormier:
In C#, `ice_invocationTimeout(TimeSpan)` accepted any `TimeSpan` value. For a timeout greater than `int.MaxValue` milliseconds (about 24.9 days), the Ice timer thread narrowed the wait delay to `int` in `Monitor.Wait(_mutex, (int)(first.scheduledTime - now))` once that deadline became the earliest pending one; the wrapped negative value made `Monitor.Wait` throw `ArgumentOutOfRangeException` outside any try block, and the unhandled exception on the timer thread terminated the process.

The root issue is that `Timer.schedule` accepted delays the timer cannot wait. `schedule` now takes a `TimeSpan` delay, performs the milliseconds conversion itself, and validates the delay against the range the timer supports (0 to `int.MaxValue` milliseconds), throwing `ArgumentOutOfRangeException` in the caller's thread — no caller can schedule a delay that would crash the timer thread. A `Debug.Assert` in the timer loop records why the narrowing to `int` is now safe.

`ice_invocationTimeout(TimeSpan)` also validates its argument against the same bound, so an unusable invocation timeout is rejected when it is set rather than at the next invocation. Values less than or equal to zero keep their existing "no timeout" semantics, consistent with the other language mappings and the existing tests, and a positive timeout is now documented to round down to the nearest millisecond.

Also adds a `Debug.Assert(interval >= 0)` in `RetryQueue.add` documenting the caller contract (0 is a valid interval, used for immediate retries).

Fixes #6102.

### DIFF
--- a/changelog.d/csharp/6102.md
+++ b/changelog.d/csharp/6102.md
@@ -1,0 +1,3 @@
+- Fixed `ice_invocationTimeout(TimeSpan)`: setting an invocation timeout greater than `int.MaxValue` milliseconds
+  (about 24.9 days) could crash the process with an unhandled exception in the Ice timer thread. Such timeouts are
+  now rejected with an `ArgumentOutOfRangeException`.
--- a/csharp/src/Ice/Internal/OutgoingAsync.cs
+++ b/csharp/src/Ice/Internal/OutgoingAsync.cs
@@ -469,7 +469,7 @@ public abstract class ProxyOutgoingAsyncBase : OutgoingAsyncBase, TimerTask
                 TimeSpan invocationTimeout = proxy_.iceReference().getInvocationTimeout();
                 if (invocationTimeout > TimeSpan.Zero)
                 {
-                    instance_.timer().schedule(this, (long)invocationTimeout.TotalMilliseconds);
+                    instance_.timer().schedule(this, invocationTimeout);
                 }
             }
             else
--- a/csharp/src/Ice/Internal/RetryQueue.cs
+++ b/csharp/src/Ice/Internal/RetryQueue.cs
@@ -67,6 +67,7 @@ public class RetryQueue
 
     public void add(ProxyOutgoingAsyncBase outAsync, int interval)
     {
+        Debug.Assert(interval >= 0);
         lock (_mutex)
         {
             if (_instance == null)
@@ -75,7 +76,7 @@ public class RetryQueue
             }
             var task = new RetryTask(_instance, this, outAsync);
             outAsync.cancelable(task); // This will throw if the request is canceled.
-            _instance.timer().schedule(task, interval);
+            _instance.timer().schedule(task, TimeSpan.FromMilliseconds(interval));
             _requests.Add(task, null);
         }
     }
--- a/csharp/src/Ice/Internal/Timer.cs
+++ b/csharp/src/Ice/Internal/Timer.cs
@@ -36,8 +36,15 @@ public sealed class Timer
         _thread.Join();
     }
 
-    public void schedule(TimerTask task, long delay)
+    public void schedule(TimerTask task, TimeSpan delay)
     {
+        if (delay < TimeSpan.Zero || delay > TimeSpan.FromMilliseconds(int.MaxValue))
+        {
+            throw new ArgumentOutOfRangeException(
+                nameof(delay),
+                $"The timer delay must be between 0 and {int.MaxValue} milliseconds.");
+        }
+
         lock (_mutex)
         {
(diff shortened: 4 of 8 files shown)


pepone (approved):
Approving. Validating in `schedule` is the right place — no caller can hand the timer a delay it can't wait on, and the range is the timer's constraint rather than something each caller has to know.

Checked the new invariant: `Debug.Assert(first.scheduledTime - now <= int.MaxValue)` holds because `Time.currentMonotonicTimeMillis()` is `Stopwatch.ElapsedMilliseconds`, so the delta can only shrink after scheduling. Same for `Debug.Assert(interval >= 0)` in `RetryQueue.add` — `Instance.cs` already clamps negative `Ice.RetryIntervals` entries to 0.

One suggestion: reject a positive sub-millisecond timeout for the same reason as the upper bound. The timer's unit is whole milliseconds, so 0.5ms is as unrepresentable as 30 days — and today it doesn't round down to "no timeout", it schedules a zero-delay token that expires immediately.

```csharp
if (newTimeout > TimeSpan.Zero &&
    (newTimeout < TimeSpan.FromMilliseconds(1) || newTimeout > TimeSpan.FromMilliseconds(int.MaxValue)))
{
    throw new ArgumentOutOfRangeException(
        nameof(newTimeout),
        $"A positive invocation timeout must be between 1 and {int.MaxValue} milliseconds.");
}
```

That collapses the contract into one closed interval, and makes the round-down sentence harmless — with a 1ms floor, rounding down can never reach 0. In `ice_invocationTimeout` only; `schedule`'s `[0, int.MaxValue]` is right as-is, since `RetryQueue` and the discovery plugins schedule zero delays.

For reference, the mappings don't agree on sub-ms today: C++ truncates in the setter (`Proxy.h:158`), so it stores 0ms and means "no timeout", while Java and C# test the unrounded value and time out immediately.

Filed as #6189 — no need to hold this PR for it.

Also `Proxy.cs:1048`: the rounding sentence is only on the interface — the implementation doc still says just "The new invocation timeout." Mirror it, per Copilot.

externl (approved):

bernardnormier:
Thanks for the careful review. I mirrored the rounding sentence onto the implementation doc in 1272d4936c.

Agreed on the 1ms floor for positive timeouts — it makes the contract a single closed interval and fixes the immediate-expiry surprise for sub-ms values. I'll leave that to #6189 as you suggested, since it should probably land in all the mappings together given the C++/Java/C# divergence you found.

PR #6184, opened 2026-07-24
Let PropertyException propagate from the ConnectionI constructor

### LINKED ISSUE #6103: C++ `ConnectionI` constructor is `noexcept` but reads throwing integer property parsers → `std::terminate`

bernardnormier:
In C++, the `ConnectionI` constructor is declared `noexcept` but reads integer properties (`Ice.Warn.Connections`, `Ice.Warn.Datagrams`, `Ice.Compression.Level`, and `Ice.UDP.SndSize` through the `BatchRequestQueue` it creates) whose parser throws `PropertyException` on a value that cannot be parsed. Property values are validated lazily at read time, so an invalid value passed communicator initialization and then aborted the process with `std::terminate` at the first connection.

This PR removes `noexcept` from the constructor. All three `ConnectionI::create` call sites already catch `LocalException`: an outgoing connection attempt now reports the `PropertyException` to the invocation, an incoming connection attempt logs a warning and refuses the connection, and the UDP bind-time path fails object-adapter creation with the exception. When the properties are read is unchanged.

The new test in `Ice/properties` establishes an outgoing connection with `Ice.Warn.Connections` set to an invalid value and expects `PropertyException`; without the fix, the client aborts with `SIGABRT`.

Fixes #6103.

### DIFF
--- a/cpp/src/Ice/ConnectionI.cpp
+++ b/cpp/src/Ice/ConnectionI.cpp
@@ -1878,7 +1878,7 @@ Ice::ConnectionI::ConnectionI(
     const EndpointIPtr& endpoint,
     const shared_ptr<ObjectAdapterI>& adapter,
     std::function<void(const ConnectionIPtr&)> removeFromFactory,
-    const ConnectionOptions& options) noexcept
+    const ConnectionOptions& options)
     : _communicator(std::move(communicator)),
       _instance(instance),
       _transceiver(transceiver),
--- a/cpp/src/Ice/ConnectionI.h
+++ b/cpp/src/Ice/ConnectionI.h
@@ -249,7 +249,7 @@ namespace Ice
             const IceInternal::EndpointIPtr&,
             const std::shared_ptr<ObjectAdapterI>&,
             std::function<void(const ConnectionIPtr&)>,
-            const ConnectionOptions&) noexcept;
+            const ConnectionOptions&);
 
         static ConnectionIPtr create(
             const Ice::CommunicatorPtr&,
--- a/cpp/src/Ice/IdleTimeoutTransceiverDecorator.cpp
+++ b/cpp/src/Ice/IdleTimeoutTransceiverDecorator.cpp
@@ -70,15 +70,18 @@ IdleTimeoutTransceiverDecorator::initialize(Buffer& readBuffer, Buffer& writeBuf
 
 IdleTimeoutTransceiverDecorator::~IdleTimeoutTransceiverDecorator()
 {
-    // Since ConnectionI is noexcept, decoratorInit always sets _heartbeatTimerTask.
-    assert(_heartbeatTimerTask);
-    _timer->cancel(_heartbeatTimerTask);
+    // _heartbeatTimerTask remains null when the connection creation fails before decoratorInit is called.
+    if (_heartbeatTimerTask)
+    {
+        _timer->cancel(_heartbeatTimerTask);
+    }
     disableIdleCheck();
 }
 
 void
 IdleTimeoutTransceiverDecorator::close()
 {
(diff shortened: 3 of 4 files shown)


pepone (approved):
The change looks right. All three `ConnectionI::create` call sites catch `LocalException` and close the transceiver, so the new throw path doesn't leak an fd, and `nextConnector` still unwinds `_pending` correctly. The `IdleTimeoutTransceiverDecorator` dtor fix is the right catch — `decoratedTransceiver` in `create` is the decorator's only owner, so a ctor throw destroys it before `decoratorInit`.

Two things:

**The server side now fails silently.** `IncomingConnectionFactory::message` logs the refusal only `if (_warn)`, which defaults to 0. With `Ice.Compression.Level=x` a server accepts and drops every connection forever with nothing in the log. Not a blocker for 3.8.3, but I'd follow up by reading these three properties once in `Instance` so a typo fails `initialize()` — same reasoning as the comment at `Instance.cpp:1188`, and it drops three property lookups per connection.

**`Fixes #6103` is broader than this PR.** The issue also covers the observer/logger calls in `setState` (~`:2195`) that sit outside its try/catch and are reached from six `noexcept` members — a throwing user observer or logger still terminates. Worth splitting that out before this auto-closes the issue.

CI: the only red job is `ios on macos-26`, failing in `Ice/timeout` (swift, `ws,ipv6,serialize,mx`) with a controller `ConnectionLostException` — unrelated.

bernardnormier:
Thanks for the review. Both follow-ups are now filed:

- #6192 splits the `setState` observer/logger part out of #6103, with the mechanism spelled out and re-verified (the three unguarded application-code call sites and the six `noexcept` members that reach them), so merging this PR no longer buries it.
- #6193 tracks the server-side silent refusal, milestoned 3.9.0. I included your Instance-eager-read suggestion but also a significant caveat: it changes when these properties are read, and that undocumented timing is something applications plausibly rely on — the test added by this PR does. So the eager validation may not be a good idea even in 3.9.0; the issue also floats a narrower alternative of logging `PropertyException` refusals unconditionally.

PR #6186, opened 2026-07-25
Reject invalid discovery timeout properties at initialization

### LINKED ISSUE #6185: IceDiscovery and IceLocatorDiscovery do not validate their timeout properties consistently
The discovery plugins read several duration properties without consistent validation:

- `IceDiscovery.Timeout`: not validated in any mapping (C++ `LookupI.cpp`, C# `LookupI.cs`, Java `LookupI.java`). A negative value is passed directly to the timer.
- `IceLocatorDiscovery.Timeout`: C# and Java clamp a negative value to 300 ms; C++ (`PluginI.cpp`) performs no check.
- `IceLocatorDiscovery.RetryDelay`: C# and Java clamp a negative value to 0; C++ performs no check.

By contrast, all three mappings validate `IceDiscovery.LatencyMultiplier` and throw `PropertyException` when it is less than 1.

With a negative timeout, the C++ and Java timers schedule the task in the past and fire it immediately — discovery requests effectively time out instantly. In C#, the Ice timer now rejects negative delays (see #6183), so a negative `IceDiscovery.Timeout` makes `Timer.schedule` throw `ArgumentOutOfRangeException` during a lookup.

The three mappings should validate these properties the same way at plugin initialization — either reject invalid values with `PropertyException`, like `IceDiscovery.LatencyMultiplier`, or clamp them, like the existing C#/Java `IceLocatorDiscovery` code — and pick one behavior for all of them.

arimu1:
Fixes #6185

## Summary

- Validate `IceDiscovery.Timeout`, `IceLocatorDiscovery.Timeout`, and `IceLocatorDiscovery.RetryDelay` in C++, C#, and Java by throwing `PropertyException` when the value is negative, matching the existing `IceDiscovery.LatencyMultiplier` behavior.
- **Prior behavior:** `IceDiscovery.Timeout` was previously passed through to the timer unvalidated. (C++ `IceInternal::Timer` rejects a negative delay with `std::invalid_argument`; Java/`ScheduledThreadPoolExecutor` and the C# timer can schedule an immediate fire.) `IceLocatorDiscovery.Timeout` / `RetryDelay` previously clamped negatives to 300 ms / 0 ms in C++, C#, and Java.
- **Behavior change:** those LocatorDiscovery properties now reject negatives at plugin initialization instead of clamping. If maintainers prefer clamp-everywhere consistency, I can switch all three mappings to that approach.

## Test plan

- [x] `./gradlew :icediscovery:compileJava :icelocatordiscovery:compileJava --rerun-tasks` (JDK 17)
- [ ] C# `dotnet build` on `IceDiscovery` / `IceLocatorDiscovery` — blocked locally by disk space during NuGet restore
- [ ] C++ build — not run (no local Ice C++ build environment configured)
- [ ] Integration tests — not run

### DIFF
--- a/changelog.d/cpp/6185.md
+++ b/changelog.d/cpp/6185.md
@@ -1,4 +1,4 @@
 - Rejected a negative `IceDiscovery.Timeout` property value with `PropertyException` at plugin initialization,
-  instead of silently scheduling a timer that fires immediately.
+  instead of letting it reach the timer.
 - `IceLocatorDiscovery.Timeout` and `IceLocatorDiscovery.RetryDelay` now reject negative values with
   `PropertyException` at plugin initialization instead of silently clamping them to 300 ms and 0 ms respectively.
--- a/changelog.d/cpp/6185.md
+++ b/changelog.d/cpp/6185.md
@@ -1,4 +1,0 @@
-- Rejected a negative `IceDiscovery.Timeout` property value with `PropertyException` at plugin initialization,
-  instead of letting it reach the timer.
-- `IceLocatorDiscovery.Timeout` and `IceLocatorDiscovery.RetryDelay` now reject negative values with
-  `PropertyException` at plugin initialization instead of silently clamping them to 300 ms and 0 ms respectively.
--- a/changelog.d/cpp/6185.md
+++ b/changelog.d/cpp/6185.md
@@ -1,3 +1,4 @@
-- Rejected negative `IceDiscovery.Timeout`, `IceLocatorDiscovery.Timeout`, and `IceLocatorDiscovery.RetryDelay`
-  property values with `PropertyException` at plugin initialization, instead of silently scheduling a timer that
-  fires immediately.
+- Rejected a negative `IceDiscovery.Timeout` property value with `PropertyException` at plugin initialization,
+  instead of silently scheduling a timer that fires immediately.
+- `IceLocatorDiscovery.Timeout` and `IceLocatorDiscovery.RetryDelay` now reject negative values with
+  `PropertyException` at plugin initialization instead of silently clamping them to 300 ms and 0 ms respectively.
--- a/changelog.d/cpp/6185.md
+++ b/changelog.d/cpp/6185.md
@@ -1,0 +1,3 @@
+- Rejected negative `IceDiscovery.Timeout`, `IceLocatorDiscovery.Timeout`, and `IceLocatorDiscovery.RetryDelay`
+  property values with `PropertyException` at plugin initialization, instead of silently scheduling a timer that
+  fires immediately.
--- a/changelog.d/csharp/6185.md
+++ b/changelog.d/csharp/6185.md
@@ -1,4 +1,0 @@
-- Rejected a negative `IceDiscovery.Timeout` property value with `PropertyException` at plugin initialization,
-  instead of letting it reach the timer.
-- `IceLocatorDiscovery.Timeout` and `IceLocatorDiscovery.RetryDelay` now reject negative values with
-  `PropertyException` at plugin initialization instead of silently clamping them to 300 ms and 0 ms respectively.
--- a/changelog.d/csharp/6185.md
+++ b/changelog.d/csharp/6185.md
@@ -1,0 +1,4 @@
+- Rejected a negative `IceDiscovery.Timeout` property value with `PropertyException` at plugin initialization,
+  instead of letting it reach the timer.
+- `IceLocatorDiscovery.Timeout` and `IceLocatorDiscovery.RetryDelay` now reject negative values with
+  `PropertyException` at plugin initialization instead of silently clamping them to 300 ms and 0 ms respectively.
--- a/changelog.d/java/6185.md
+++ b/changelog.d/java/6185.md
@@ -1,4 +1,0 @@
-- Rejected a negative `IceDiscovery.Timeout` property value with `PropertyException` at plugin initialization,
-  instead of letting it reach the timer.
-- `IceLocatorDiscovery.Timeout` and `IceLocatorDiscovery.RetryDelay` now reject negative values with
-  `PropertyException` at plugin initialization instead of silently clamping them to 300 ms and 0 ms respectively.
--- a/changelog.d/java/6185.md
+++ b/changelog.d/java/6185.md
@@ -1,0 +1,4 @@
+- Rejected a negative `IceDiscovery.Timeout` property value with `PropertyException` at plugin initialization,
+  instead of letting it reach the timer.
+- `IceLocatorDiscovery.Timeout` and `IceLocatorDiscovery.RetryDelay` now reject negative values with
+  `PropertyException` at plugin initialization instead of silently clamping them to 300 ms and 0 ms respectively.
--- a/cpp/src/IceDiscovery/LookupI.cpp
+++ b/cpp/src/IceDiscovery/LookupI.cpp
@@ -193,6 +193,13 @@ LookupI::LookupI(LocatorRegistryIPtr registry, const LookupPrx& lookup, const Ic
       _domainId(properties->getIceProperty("IceDiscovery.DomainId")),
       _timer(IceInternal::getInstanceTimer(lookup->ice_getCommunicator()))
 {
+    if (_timeout < chrono::milliseconds::zero())
+    {
+        throw PropertyException{
+            __FILE__,
+            __LINE__,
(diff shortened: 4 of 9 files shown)


arimu1:
I have read the CLA Document and I hereby sign the CLA

arimu1:
recheck

pepone:
Thanks for the patch — the guards look right, and CI covered the C++/C# builds you couldn't run locally.

One request before you change anything: please wait for @bernardnormier to weigh in. He filed #6185 and offered both reject and clamp there, so the throw-vs-clamp decision and the exception type are his call. Everything below is what I'd ask for, but don't start on it until he confirms.

My take on your open question: rejecting is the right choice, not clamp-everywhere. It matches the `IceDiscovery.LatencyMultiplier` guard, a negative timeout is unambiguously a configuration error, and nothing in-tree sets one of these properties to a negative value.

**Reject a timeout of 0 as well.** Both plugins send the lookup and then schedule the timeout (`cpp/src/IceDiscovery/LookupI.cpp:321`, `cpp/src/IceLocatorDiscovery/PluginI.cpp:649`), so `Timeout=0` leaves no waiting window at all. Ice already draws this line: where a non-positive timeout has a meaning it is accepted silently (`ConnectTimeout`, `CloseTimeout`, `IdleTimeout`, `InactivityTimeout` all just skip their timer — `ConnectionI.cpp:374`, `:3549`, `:1934`, `:1950`), but where it has none, the value is range-checked at startup — `Ice.Default.InvocationTimeout` accepts only `-1` or `>= 1` and rejects `0` in all three mappings (`cpp/src/Ice/DefaultsAndOverrides.cpp:57`, `DefaultsAndOverrides.cs:81`, `DefaultsAndOverrides.java:65`). A discovery lookup has no "no timeout" state — the timer is what completes the request and drives the retries — so it belongs in the second group:

- `IceDiscovery.Timeout` and `IceLocatorDiscovery.Timeout`: require `> 0`, with the message `property 'X.Timeout' must be greater than 0`.
- `IceLocatorDiscovery.RetryDelay`: keep `>= 0` as you have it — `0` means retry immediately, just as `Ice.Default.LocatorCacheTimeout = 0` means don't cache.

So in C++: `if (_timeout <= chrono::milliseconds::zero())`, and `if (_timeout <= 0)` in C# and Java. No `-1` sentinel needed: waiting forever for a multicast reply isn't a useful configuration.

Two changelog fixes:

**The entry names an exception the application never sees.** `initialize()` doesn't let `PropertyException` out — the plugin manager rewraps it as `PluginInitializationException` (`cpp/src/Ice/PluginManagerI.cpp:45`, `csharp/src/Ice/PluginManagerI.cs:59`, `PluginManagerI.java:34`; the last two keep the `PropertyException` as the cause). The entry should name the exception the application catches.

**Use a single `changelog.d/general/6185.md`.** The three fragments are byte-identical, and the change reaches every mapping that supports these plugins — Python, Ruby, PHP, MATLAB and Swift register the C++ plugin factories, so their users are affected too and would see nothing under a C++-only heading.

```markdown
- The discovery plugins now validate their timeout properties when the plugin is initialized:
  `IceDiscovery.Timeout` and `IceLocatorDiscovery.Timeout` must be greater than 0, and
  `IceLocatorDiscovery.RetryDelay` must not be negative. An invalid value fails communicator initialization with a
  `PluginInitializationException` caused by a `PropertyException`. Previously, `IceLocatorDiscovery.Timeout` and
  `IceLocatorDiscovery.RetryDelay` silently clamped a negative value to 300 ms and 0 ms, and a negative
  `IceDiscovery.Timeout` reached the timer, where every discovery lookup then failed or timed out immediately.
```

I wouldn't act on Copilot's test requests: the existing `LatencyMultiplier` guard has no test in any mapping, and a test here would have to expect `PluginInitializationException`, not `PropertyException`.

Heads-up on merge order: this conflicts with #6183 in both C# files — that PR converts `_timeout` to `TimeSpan` on the same lines and keeps the 300 ms clamp. Whichever lands second gets rebased, and the clamp there becomes the throw.

bernardnormier:
Thanks @arimu1 for the contribution, and @pepone for the detailed analysis. My calls on the open questions:

**Reject, don't clamp.** Rejecting at plugin initialization is the right behavior — it matches the existing `IceDiscovery.LatencyMultiplier` guard, and a negative timeout is a configuration error the user should hear about, not something we paper over. No clamp-everywhere variant needed.

**Also reject `0` for the two Timeouts.** As @pepone explained, both plugins send the lookup and then schedule the timeout task, and that timer is what completes the request and drives retries — so `Timeout=0` has no meaningful semantics, unlike the connection timeouts where `0`/non-positive means "no timer". Please make it:

- `IceDiscovery.Timeout` and `IceLocatorDiscovery.Timeout`: require `> 0`, with the message `property 'X.Timeout' must be greater than 0`.
- `IceLocatorDiscovery.RetryDelay`: keep `>= 0` as you have it.

**Drop the changelog fragments.** Please delete all three files. Our bar for a changelog entry is that the change affects valid programs during normal operations. Setting a negative timeout is an invalid configuration, and how invalid config fails (clamp vs. exception at startup) isn't behavior anyone may rely on — so this fix is consistency hardening, not news. (#6184 merged without an entry on the same basis.) This also moots the `PropertyException` vs `PluginInitializationException` wording question.

**No new tests needed** — the existing `LatencyMultiplier` guard isn't tested either, and I don't want tests pinning the wrapped exception type.

One logistical note: #6183 merged today, so please rebase on main — in both C# files `_timeout` is now a `TimeSpan`, so the guards there become `_timeout <= TimeSpan.Zero` (and `_retryDelay < 0` stays an `int` comparison).

arimu1:
Thanks @bernardnormier and @pepone — applied your rulings:

1. **Reject, don't clamp** — unchanged.
2. **Timeouts must be `> 0`** for `IceDiscovery.Timeout` and `IceLocatorDiscovery.Timeout` (C++/C#/Java), message exactly `property 'X.Timeout' must be greater than 0`. `IceLocatorDiscovery.RetryDelay` stays `>= 0`.
3. **Removed** the three `changelog.d/{cpp,csharp,java}/6185.md` fragments.

Also rebased onto current `main`. I couldn't build C++ locally on this Mac; relying on CI for C++/C# as before.

pepone (approved):
Matches the rulings exactly — guards in all six files, `> 0` for both Timeouts, `>= 0` for `RetryDelay`, changelog fragments gone, rebased on main. CI is green on `aaef8d0`, including clang-format, checkstyle, rewrite, dotnet-format and the full C++/C#/Java test runs.

Checked the blast radius: nothing in the tree sets any of the three to `0` or a negative — `scripts/tests/IceDiscovery/simple.py` and `scripts/tests/IceGrid/simple.py` use 50, the three `IceGrid/simple` `AllTests` use 100, and the defaults are 300/300/2000. IceGrid propagates only `IceLocatorDiscovery.InstanceName` to spawned servers (`cpp/src/IceGrid/ServerI.cpp:3033-3038`), never the timeouts, so nothing deployed inherits a bad value. C++/C#/Java are the whole surface — Python, PHP, Ruby, MATLAB and Swift push the C++ plugin factories.

Approving. Three optional nits, none blocking:

**`csharp/src/IceDiscovery/LookupI.cs:229`** — use the same form as the sibling file: assign `_timeout` first, then guard `_timeout <= TimeSpan.Zero`. `_timeout` is `readonly` (line 549), so constructor assignment is fine and the local goes away.

**`cpp/src/IceLocatorDiscovery/PluginI.cpp:87,89`** — `_timeout` and `_retryDelay` are write-once now that the clamps are gone; make them `const chrono::milliseconds` to match `IceDiscovery::LookupI` (`LookupI.h:148`). `_retryCount` stays non-const, it's still clamped.

**Line wrapping** — with the shorter messages, three of the throws fit on one line under the 120-column limit: `csharp/src/IceLocatorDiscovery/PluginI.cs:119` (105 cols), `IceDiscovery/LookupI.java:242` (98), `IceLocatorDiscovery/PluginI.java:141` (109). The two `RetryDelay` ones are 120 and 124, so those stay wrapped.

`RetryCount` still silently clamps a negative to 0 in all three mappings, now sitting between two guards that throw. Outside the stated scope for this PR, so I filed it as #6286 — @bernardnormier, fold it in here instead if you prefer and close that one.

bernardnormier (approved):
Looks good. #6286 is actually a duplicate of #6190; let's not fix this issue as part of this PR.

Issue #5781, opened 2026-06-30
DataStorm: TopicSubscriber reap leaves stale entries in the keys/filters maps


pepone:
## Summary

When a DataStorm session reconnects, `SessionI` reaps the topic subscriber via `subscriber.reap(_sessionInstanceId)` ([`SessionI.cpp#L333`](https://github.com/zeroc-ice/ice/blob/a97a2b8e57396b4aa707c508b26e46df9e1f9cb9/cpp/src/DataStorm/SessionI.cpp#L333)), which removes stale entries from the subscriber's `_elements`/`_subscribers`. It does **not** reap the per-key `keys` map or the `filters` map ([`SessionI.h#L182`](https://github.com/zeroc-ice/ice/blob/a97a2b8e57396b4aa707c508b26e46df9e1f9cb9/cpp/src/DataStorm/SessionI.h#L182)), so element ids for elements that are not re-attached after the reconnect remain in `keys[keyId].second` indefinitely. This is the existing in-code `// TODO: reap keys / filters`:

[`SessionI.cpp#L332-L335`](https://github.com/zeroc-ice/ice/blob/a97a2b8e57396b4aa707c508b26e46df9e1f9cb9/cpp/src/DataStorm/SessionI.cpp#L332-L335)
```cpp
// Reap unused keys and filters from the topic subscriber
subscriber.reap(_sessionInstanceId);

// TODO: reap keys / filters
```

## Impact

No correctness or crash impact today. `getLastIds` — the only consumer that dereferenced these entries — now skips ids with no live subscriber ([`SessionI.cpp#L1172-L1182`](https://github.com/zeroc-ice/ice/blob/a97a2b8e57396b4aa707c508b26e46df9e1f9cb9/cpp/src/DataStorm/SessionI.cpp#L1172), added in #5723 and #5604), so stale entries are simply ignored.

The residual is memory hygiene: stale `keys`/`filters` entries accumulate across reconnects and are never released. In a long-running node with high element churn over many reconnects, the per-key maps grow without bound (bounded only by the set of distinct element ids ever seen per key). Low severity.

## Suggested fix

[...]

PR #5597, opened 2026-06-15
Remove dead code in C++

bernardnormier:
This PR removes unused internal C++ code.

PR #6196, opened 2026-07-27
Use weak ownership for DataStorm topic subscribers

### LINKED ISSUE #6088: DataStorm: SessionI subscriber maps key on raw TopicI* — a lingering entry is a use-after-free, not a leak

pepone:
Fixes #6088.

SessionI now keys each remote-topic subscriber map by weak TopicI ownership identity. Every map consumer locks the weak key before accessing the topic mutex or state and discards expired entries. This avoids both dangling dereferences and collisions when a later topic reuses the same address, without introducing a Topic-to-Session-to-Topic ownership cycle.

The topic attach, detach, and disconnect paths now pass shared topic identity into SessionI, and DataElementI exposes its retained parent for subscriber lookup. This keeps the existing destroyed-topic cleanup from #5872 and #6087 intact: destroyed topics that are still alive remain available for cleanup, while actually expired topics are skipped safely.

This also removes SessionI::_topicLock. Its only reader was removed by #5597, leaving the member write-only and not exception-safe.

The related reconnect key bookkeeping in #5781 is intentionally not included: it has separate refcount semantics and needs its own fix.

No changelog fragment is included because this is defensive hardening for an invariant violation with no currently confirmed normal-operation trigger.

Testing:

- make -C cpp -j8 DataStorm
- python3 allTests.py --filter=DataStorm (11/11 passed)

### DIFF
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -158,16 +158,9 @@ SessionI::announceTopics(TopicInfoSeq topics, bool, const Current&)
         auto p = _topics.begin();
         while (p != _topics.end())
         {
-            auto detach = [this, topicId = p->first](TopicI* topic, TopicSubscriber& subscriber)
+            auto detach = [this, topicId = p->first](const shared_ptr<TopicI>& topic)
             {
-                if (subscriber.getAll().empty())
-                {
-                    // No element subscriptions: nothing to detach. The element subscribers are also what keeps the
-                    // topic alive (each element holds its parent topic), so the topic must not be dereferenced here.
-                    return;
-                }
                 unique_lock<mutex> topicLock(topic->getMutex());
-                _topicLock = &topicLock;
                 if (topic->isDestroyed())
                 {
                     // The topic is being destroyed: TopicI::disconnect detaches its elements only while the
@@ -181,7 +174,6 @@ SessionI::announceTopics(TopicInfoSeq topics, bool, const Current&)
                     // session from the topic's listeners.
                     topic->detach(topicId, shared_from_this());
                 }
-                _topicLock = nullptr;
             };
             if (p->second.reap(_sessionInstanceId, detach))
             {
@@ -226,7 +218,7 @@ SessionI::attachTopic(TopicSpec spec, const Current&)
                 // If the topic spec has tags, decode them and add them to the subscriber.
                 if (!spec.tags.empty())
                 {
-                    auto& subscriber = _topics.at(spec.id).getSubscriber(topic.get());
+                    auto& subscriber = _topics.at(spec.id).getSubscriber(topic);
                     for (const auto& tag : spec.tags)
                     {
                         subscriber.tags[tag.id] =
@@ -273,10 +265,15 @@ SessionI::detachTopic(int64_t id, const Current&)
         return;
     }
 
-    for (auto& [topic, _] : t->second.getSubscribers())
+    for (auto& [topicRef, _] : t->second.getSubscribers())
     {
+        auto topic = topicRef.lock();
+        if (!topic)
+        {
+            continue;
+        }
+
         unique_lock<mutex> topicLock(topic->getMutex());
-        _topicLock = &topicLock;
         if (topic->isDestroyed())
         {
             // The topic is being destroyed: TopicI::disconnect detaches its elements only while the subscriber
@@ -292,7 +289,6 @@ SessionI::detachTopic(int64_t id, const Current&)
             }
             topic->detach(id, shared_from_this());
         }
-        _topicLock = nullptr;
     }
 
     // Erase the topic
@@ -310,7 +306,7 @@ SessionI::attachTags(int64_t topicId, ElementInfoSeq tags, bool initialize, cons
 
     runWithTopics(
         topicId,
-        [&](TopicI* topic, TopicSubscriber& subscriber)
+        [&](const shared_ptr<TopicI>& topic, TopicSubscriber& subscriber)
         {
             if (_traceLevels->session > 2)
             {
@@ -341,7 +337,7 @@ SessionI::detachTags(int64_t topicId, LongSeq tags, const Current&)
 
     runWithTopics(
         topicId,
-        [&](TopicI* topic, TopicSubscriber& subscriber)
+        [&](const shared_ptr<TopicI>& topic, TopicSubscriber& subscriber)
         {
             if (_traceLevels->session > 2)
             {
@@ -367,7 +363,7 @@ SessionI::announceElements(int64_t topicId, ElementInfoSeq elements, const Curre
 
     runWithTopics(
         topicId,
-        [&](TopicI* topic, TopicSubscriber&)
+        [&](const shared_ptr<TopicI>& topic, TopicSubscriber&)
         {
             if (_traceLevels->session > 2)
             {
@@ -406,7 +402,7 @@ SessionI::attachElements(int64_t topicId, int64_t peerTopicId, ElementSpecSeq el
     runWithTopics(
         topicId,
         peerTopicId,
-        [&](TopicI* topic, TopicSubscriber& subscriber)
+        [&](const shared_ptr<TopicI>& topic, TopicSubscriber& subscriber)
         {
             if (_traceLevels->session > 2)
             {
@@ -455,7 +451,7 @@ SessionI::attachElementsAck(int64_t topicId, int64_t peerTopicId, ElementSpecAck
     runWithTopics(
         topicId,
         peerTopicId,
-        [&](TopicI* topic, TopicSubscriber&)
+        [&](const shared_ptr<TopicI>& topic, TopicSubscriber&)
         {
             if (_traceLevels->session > 2)
             {
@@ -499,7 +495,7 @@ SessionI::detachElements(int64_t id, LongSeq elements, const Current&)
 
     runWithTopics(
         id,
-        [&](TopicI* topic, TopicSubscriber& subscriber)
+        [&](const shared_ptr<TopicI>& topic, TopicSubscriber& subscriber)
         {
             if (_traceLevels->session > 2)
             {
@@ -550,7 +546,7 @@ SessionI::initSamples(int64_t topicId, int64_t peerTopicId, DataSamplesSeq initi
         runWithTopics(
             topicId,
             peerTopicId,
-            [&](TopicI* topic, TopicSubscriber& subscriber)
+            [&](const shared_ptr<TopicI>& topic, TopicSubscriber& subscriber)
             {
                 ElementSubscribers* elementSubscribers = subscriber.get(initializationBatch.id);
                 if (!elementSubscribers)
@@ -772,7 +768,9 @@ SessionI::disconnectedImpl(const ConnectionPtr& connection, exception_ptr ex)
         // The analyzer falsely flags the lambda's captures as undefined: it loses track of the closure once it
         // is stored in runWithTopics' std::function parameter.
         // NOLINTNEXTLINE(clang-analyzer-core.NullDereference)
-        runWithTopics(topicId, [topicId, self](TopicI* topic, TopicSubscriber&) { topic->detach(topicId, self); });
+        runWithTopics(
+            topicId,
+            [topicId, self](const shared_ptr<TopicI>& topic, TopicSubscriber&) { topic->detach(topicId, self); });
     }
 
     // The peer's wire disconnected() op is not a connection close, so unregister the session that connected()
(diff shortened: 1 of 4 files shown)


pepone:
> [!NOTE]
> Review by Claude Code (Claude Fable 5), posted at Jose's request.

I reviewed the diff against the full sources at cb3f0a4, focusing on the ownership chains and every mutation path of the subscriber map.

**Verdict: correct and well executed.** The weak-key design is the right pattern for this problem, and every semantic change I checked against the old code is either equivalent or a strict improvement. The findings below are minor.

### What I verified

- **The weak-key pattern is sound.** `owner_less` ordering is based on the control block, which an expired `weak_ptr` still retains, so map ordering invariants survive topic destruction; a later topic reusing the same address gets a distinct control block, so the aliasing variant is closed too. Both hazards from #6088 are eliminated: expired → erased, alive-but-destroyed → skipped by `isDestroyed()`.
- **The load-bearing ownership invariant holds.** `ElementSubscribers` keys are owning `shared_ptr<DataElementI>`, and `DataElementI::_parent` owns the topic — so a weak key can only expire once its `TopicSubscriber` has no element subscriptions left. The erase-without-detach paths (`reap`, `runWithTopics`) therefore never skip element cleanup, and the destroyed-topic cleanup from #5872/#6087 is preserved.
- **Removing the `getAll().empty()` early-return in the `announceTopics` detach lambda is right, and slightly better than before.** That guard existed only because a raw pointer to an element-less topic might dangle. With `lock()` proving liveness, a live topic with an element-less stale subscriber now gets a real `topic->detach()`, dropping the session from the topic's `_listeners` — a stale listener the old code had no safe way to remove.
- **The `reap()` rewrite is behavior-equivalent.** The merged condition covers exactly the union of the old whole-object-stale and per-entry-stale paths, plus the new expired-entry cleanup; the return value is unchanged.
- **Iterator safety.** The erase-while-iterating loops use the standard idiom, and no callback can mutate the `SubscriberMap` under iteration: `unsubscribe` (reached via `TopicI::detach`) only touches element-level state, and `removeSubscriber` is only reachable via `TopicI::destroy` → `SessionI::disconnect`, which needs the session mutex — excluded while any `runWithTopics` caller holds it. `disconnectedImpl`'s outer `_topics` loop is unaffected since only inner-map entries get erased.
- **The new `shared_from_this()` calls in `TopicI` are safe.** `attach`/`detach`/`disconnect` all run on live, shared_ptr-owned topics; `disconnect()` is only called from `destroy()`, never from the destructor (`~TopicI` is `assert(_destroyed)` only), so no `bad_weak_ptr` risk.
- **Lock ordering is unchanged** (session mutex → topic mutex; `TopicI::destroy` drops the topic lock before calling back into the session). `_topicLock` was indeed write-only since #5597, and it wasn't exception-safe — good removal. `_sessionInstanceId{0}` closes a (practically unreached) uninitialized read. No stray raw `TopicI*` remains except the `DataElementI` constructor parameter, which is fine.

### Minor points

1. **Latent hazard worth a code comment.** The id-based `runWithTopics` overloads can now release the *last* topic reference while holding the session mutex (the temporary from `p->first.lock()`; the raw-pointer code never held a reference). Today that's harmless — `~TopicI`/`~DataElementI` are assert-only, and the loop body releases the topic mutex before the `shared_ptr`. But the `retained` mechanism in the name-based overload exists precisely because topic teardown once called back into the session; if the destructors ever regain real teardown, these overloads silently acquire the same hazard without the escape hatch. A one-line comment near the `lock()` would keep that from being rediscovered the hard way.
2. **`getSubscriber` release-mode behavior.** The old code ended in `.at()`, so a broken precondition threw `out_of_range` in release builds; the new find + `assert` + deref is UB in release if the invariant is ever violated. Given the PR's theme is surviving invariant violations, consider keeping a throwing fallback — though the single-lookup form is defensible.
3. **Optional simplification:** `owner_less<>` is transparent, so `find(topic)` accepts the `shared_ptr` directly — the `weak_ptr<TopicI>{topic}` temporaries are only genuinely required for `erase` (C++17 `map::erase` takes `key_type` only). Keeping them everywhere for symmetry is also fine.
4. **Tests:** no new regression test, which seems like the right call — deterministically forcing the race needs allocator address reuse plus disconnect/destroy/reconnect timing, and a bespoke harness would be disproportionate. Existing DataStorm suite passes 11/11.

pepone:
Applied the verified review follow-ups in 809a9d7886:

- Restored defined release-build behavior in `TopicSubscribers::getSubscriber` by using `map::at`, which throws `std::out_of_range` if the invariant is violated.
- Added comments to both id-based `runWithTopics` overloads documenting that the locked `shared_ptr` can be the final owning reference and that `TopicI` destruction must remain callback-free while the session mutex is held.

I left the transparent `find(topic)` conversions unchanged because they are an optional style simplification, and did not add a regression test because deterministic reproduction would require allocator address reuse and timing control or test-only hooks.

Validation completed successfully:

- `clang-format --dry-run --Werror cpp/src/DataStorm/SessionI.cpp cpp/src/DataStorm/SessionI.h`
- `python3 allTests.py DataStorm/api`
- `python3 allTests.py DataStorm/reliability`

externl (approved):
Verified the weak-key rekey: only element-less subscriber entries could dangle (elements pin their parent topic), and `reap` preserves the whole-map-stale and destroyed-topic paths.

- No callback in the new erase-while-iterating loops mutates the subscriber map re-entrantly.
- `out << topic` still prints `id:name` — the std `shared_ptr` inserter delegates to `.get()`, ADL picks up the `TopicI*` overload.
- Nit: `getSubscriber` lost its debug assert, so debug builds now get an uncaught `out_of_range` instead.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -1431,50 +1431,70 @@ SessionI::runWithTopics(
...
     {
-        for (auto& [topic, subscriber] : t->second.getSubscribers())
+        auto& subscribers = t->second.getSubscribers();
+        auto p = subscribers.begin();
+        while (p != subscribers.end())
         {
-            unique_lock<mutex> lock(topic->getMutex());
-            if (topic->isDestroyed())
+            // The shared_ptr returned by lock() can be the last owning reference. TopicI's destructor must remain
+            // callback-free because the caller holds the session mutex.
```
bernardnormier: "TopicI's destructor must remain callback-free" states a requirement on remote code; better to state the local fact and let `~TopicI` speak for itself:
```suggestion
            // lock() can return the last owning reference, so the topic can be destroyed at the end of
            // this iteration, while the session mutex is held.
```

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -1431,50 +1431,70 @@ SessionI::runWithTopics(
...
     auto t = _topics.find(topicId);
     if (t != _topics.end())
     {
-        for (auto& [topic, subscriber] : t->second.getSubscribers())
+        auto& subscribers = t->second.getSubscribers();
+        auto p = subscribers.begin();
+        while (p != subscribers.end())
         {
+            // The shared_ptr returned by lock() can be the last owning reference. TopicI's destructor must remain
+            // callback-free because the caller holds the session mutex.
```
bernardnormier: Same here:
```suggestion
            // lock() can return the last owning reference, so the topic can be destroyed at the end of
            // this iteration, while the session mutex is held.
```

bernardnormier (approved):
The weak-key conversion is sound: the control block held by the weak key rules out address-reuse collisions, every consumer locks before dereferencing, the paths that skip expired entries without erasing them clear the map right after, and the restructured `reap` is behavior-equivalent plus expired-topic reaping. Dropping the `announceTopics` empty-elements guard is right — it only existed because an unpinned topic must not be dereferenced. Agreed on no changelog fragment. Two identical suggestions on the new comment.

PR #6197, opened 2026-07-27
Fix slice2cpp and slice2java crash on a doc-comment tag with no description

### LINKED ISSUE #6123: Empty `@param`/`@throws` descriptions crash slice2cpp (and slice2java on `@throws`) — invalid empty doc-list access
> [!NOTE]

## Summary

`DocCommentParser` intentionally creates **empty** line-lists for a bare `@param`/`@throws` tag (a description-less tag is legal, parser-accepted Slice), but `slice2cpp` and `slice2java` access `.front()` / assert non-empty without a guard, so a bare doc tag **crashes the compiler** (SIGSEGV / SIGABRT) with no diagnostic.

**Parser:** `cpp/src/Slice/DocCommentParser.cpp:546,596` create `{}` entries; text is appended only when non-empty (`:732`).

**slice2cpp — crashes on in/out `@param` AND `@throws`** (`cpp/src/slice2cpp/Gen.cpp`): `writeDocLines` (`:233-250`) does `assert(l != cend()); out << *l;` with no emptiness guard; `writeOpDocParams` (`:382`) and `writeOpDocExceptions` (`:437`) call it unguarded; `createOpOutParamsDoc` (`:417`) does `outParamDoc.front().insert(...)`. Codex: bare out-param `@param` → exit 139; bare `@throws` → abort.

**slice2java — crashes on `@throws` only** (`cpp/src/slice2java/Gen.cpp`): `writeExceptionDocComment` (`:1930`) → `writeDocCommentLines` (`:1960`, `assert(!lines.empty()); ... l.front()`). Its parameter sites are already guarded (`!…empty()` at `:966,1033,1162,1232,2132`), so the "Java params also crash" implication does **not** hold — Java crashes on exceptions only. Codex: bare `@throws` → exit 134.

**Safe (already guard):** slice2js (`if (lines.size() > 0)`), slice2matlab (`if (!docLines.empty())`), slice2cs, slice2py, slice2rb.

## Fix

Add the `if (!lines.empty())` guard (which the other 4 back-ends already have) to `slice2cpp/Gen.cpp:233` (`writeDocLines`) + `:417` (`createOpOutParamsDoc`) and `slice2java/Gen.cpp:1930` (`writeExceptionDocComment`), or reject description-less tags in `DocCommentParser`.

**Severity:** Medium (hard compiler crash, no diagnostic, on legal-but-sparse Slice doc comments — a build-time DoS).

externl:
Fixes #6123.

A `@param`, `@return`, or `@throws` tag with no description is valid Slice, and `DocCommentParser` stores an empty
line list for it. Three code-generation sites read the first element of that list without checking, so a named tag
with no description corrupted the heap and aborted the compiler in a default build, where the asserts guarding
those reads are compiled out by `-DNDEBUG`.

- Guarded the empty list at the three call sites that can see one — the `@param` and `@throws` writers in
  `slice2cpp` and the `@throws` writer in `slice2java` — writing the separator space inside the guard, so a
  description-less tag emits no trailing whitespace. The helpers keep their asserts: every caller handles the
  empty list explicitly.
- A `@param` for an `out` parameter with no description now contributes a bullet with just the parameter name,
  instead of dropping the parameter from the generated list.
- Added `bareCommentTagTest` to `WarningInvalidComment.ice`, covering description-less `@param` (in and out),
  `@return`, and `@throws`. It produces no new warnings, so the `.err` change is only a line-number shift. The
  harness runs `slice2cpp` only, so the `slice2java` half has no automated coverage.


cpp/test/Slice/errorDetection/WarningInvalidComment.ice
```diff
@@ -71,6 +71,14 @@ module Test
         /// @throws CommentDummy It should be okay for the description of an exception to start on one line,
         ///     and then continue onto the next line. Whitespace should be stripped correctly too.
         bool splitMultilineCommentTest(int param) throws CommentDummy;
+
+        /// This tests that doc-comment tags with no description at all are accepted. The description is optional, so
+        /// none of these produce a warning; they used to make the language mappings read from an empty list.
+        /// @param param
+        /// @param myOut
+        /// @return
```
InsertCreativityHere, line 78: I wonder if there is really any point to allowing an empty `@return` tag... The other tags at least name the parameter (which does show up in the generated API reference with it.

But the empty `@return` is adding nothing. Either way, outside the scope of this PR for sure.

cpp/test/Slice/errorDetection/WarningInvalidComment.ice
```diff
@@ -71,6 +71,14 @@ module Test
         /// @throws CommentDummy It should be okay for the description of an exception to start on one line,
         ///     and then continue onto the next line. Whitespace should be stripped correctly too.
         bool splitMultilineCommentTest(int param) throws CommentDummy;
+
+        /// This tests that doc-comment tags with no description at all are accepted. The description is optional, so
+        /// none of these produce a warning; they used to make the language mappings read from an empty list.
```
InsertCreativityHere: I wouldn't mention that these "used to make the language mappings...". And I think the part before could be said simpler too:
```suggestion
        /// This tests that doc-comment tags with no descriptions are accepted, without producing any warnings.
```

cpp/src/slice2java/Gen.cpp
```diff
@@ -1964,7 +1964,13 @@ Slice::JavaVisitor::writeDocCommentLines(Output& out, const StringList& lines)
     // subsequent lines. We assume the caller prepended a leading " * " for the first
     // line if necessary.
     //
-    assert(!lines.empty());
+    // 'lines' is empty when the doc comment has a tag with no description, such as a bare '@throws Ex'. The caller
```
InsertCreativityHere: I do think this fix should be in `writeExceptionDocComment` instead. Specifically, `writeDocCommentLines(out, lines);` should become:
```
if (!lines.empty())
{
    writeDocCommentLines(out, lines);
}
```
I don't see any other spots where this bug would cause a crash at.

----

The reason we had `assert(!lines.empty);` here is to makes sure we explicitly handle an empty description that in all callers of this function. Otherwise, even if we don't crash, we could end up accidentally generating something like: `/** */`: where we skip writing the inside because, instead of skipping writing the entire thing.

changelog.d/java/6123.md
```diff
@@ -0,0 +1,2 @@
+- Fixed a crash in `slice2java` when compiling a doc comment with a `@throws` tag that has no description. Such tags
+  are valid: the description is optional.
```
InsertCreativityHere: Just to be clear, writing a bare `@throws` is rejected as an invalid doc-comment.
What actually causes the crash is `@throws Foo`: a valid tag, just with no description after it.

InsertCreativityHere (commented):
I think we should keep the `assert` in `slice2java` and move the check somewhere else.
The rest looks fine though.

pepone (approved):
Verified locally: built `slice2cpp`/`slice2java` at 863bac1a90 and at its parent, and ran both on the new fixture. The
parent segfaults (`slice2cpp` exit 139 — release build, the assert is compiled out by `-DNDEBUG`) and aborts
(`slice2java` exit 134); with the fix both exit 0. Regenerating all 34 Slice files under `slice/` with each build gives
byte-identical output (58 C++ / 317 Java files), so nothing in our own doc comments changes. The other seven compilers
don't have the bug.

+1 on keeping the `assert` in `slice2java`. `writeExceptionDocComment` is the only caller of that overload that can pass
an empty list — every other call site already checks for an empty list first — so the narrow fix is sufficient.

One nit: the separator space is written before the description, so a bare tag now emits a line ending in a space
(`/// @param a `, `/// @throws Test2::E1 `, ` * @throws Test.Err `). Our generated headers have no trailing whitespace
today. Moving the space inside the guard fixes that, and gives the `slice2java` shape above for free:

```cpp
out << nl << "/// @throws " << scopedName;
if (!lines.empty())
{
    out << ' ';
    writeDocLines(out, lines, false);
}
```

Two things for follow-up issues rather than this PR:

- A bare `@return` still drops the `` - `returnValue` `` bullet, so `int op(int a, out int b, out int c)` documents a
  three-element tuple with two bullets and the first one names the second element. Same shape as the out-parameter case
  you fixed here, but pre-existing — it already happens when `@return` is simply omitted — and `DocComment` can't
  distinguish "no tag" from "tag with no description" since `_returns` is a plain `StringList`.
- On the missing `slice2java` coverage: `slice2java`'s stderr matches `slice2cpp`'s on 53 of the 54 `errorDetection`
  fixtures (only `WarningInvalidMetadata.ice` differs, as expected). Running the other back-ends over the same fixtures
  and checking just the exit status would have caught this crash for all nine compilers.

externl:
Reworked per review: both asserts restored, the guards moved to the three tag call sites (`writeExceptionDocComment` plus the two `slice2cpp` sites), with the separator space inside the guard so description-less tags emit no trailing whitespace. Also applied the fixture-comment and changelog rewordings.

InsertCreativityHere (approved):
Looks good to me!

PR #6200, opened 2026-07-27
Fix IcePy servant reference leaks from ServantWrapper::getObject

### LINKED ISSUE #6127: IcePy leaks a servant reference from `ServantWrapper::getObject()` at three adapter call sites
> [!NOTE]

## Summary

`IcePy::ServantWrapper::getObject()` returns a **new** reference (`Py_NewRef`), but three callers treat it as borrowed and never adopt it, leaking one Python servant reference per located request and one per returned facet per call.

**Files:** `python/modules/IcePy/Operation.cpp:2496-2500` (`return Py_NewRef(_servant);`). Leaking callers in `python/modules/IcePy/ObjectAdapter.cpp`:
- `:210` `ServantLocatorWrapper::finished` — `PyObject_CallMethod(_locator, "finished", "OOO", ..., wrapper->getObject(), ...)`; `"O"` INCREFs into the args tuple, the new ref is never DECREF'd → leak per located request (normal + exception dispatch both call `finished`).
- `:837` `removeAllFacets` and `:1009` `findAllFacets` — `PyDict_SetItemString(dict, facet, wrapper->getObject())`; `SetItemString` does not steal → leak per facet.

Root cause: this overload's contract diverges from its siblings — `LoggerWrapper::getObject()` and `ValueReader::getObject()` return *borrowed* refs, and callers that wrap the borrowed ones in `Py_NewRef` are correct precisely because those are borrowed. The other ~11 `return wrapper->getObject();` sites hand the value straight back to the Python C-API (correct for a new ref). Only these three leak.

## Fix

Adopt each returned new reference in a `PyObjectHandle` before passing the borrowed pointer onward (or make `ServantWrapper::getObject()` return a borrowed reference to match its siblings).

**Severity:** Medium — the `finished()` leak grows one servant reference per ServantLocator-dispatched request, keeping every located servant alive forever (a slow but genuine leak in a common server pattern).

externl:
Fixes #6127.

`ServantWrapper::getObject()` returns a new reference, but three callers treated it as borrowed, leaking one
servant reference per servant-locator-dispatched request and one per facet returned by
`findAllFacets`/`removeAllFacets`.

- Adopt the returned reference in a `PyObjectHandle` at the three call sites: `ServantLocatorWrapper::finished`
  (the `"O"` format increfs on its own) and the two facet loops (`PyDict_SetItemString` increfs on its own).
- Documented the new-reference contract on the `getObject` declarations of `ServantWrapper` and
  `ServantLocatorWrapper`, whose siblings (`LoggerWrapper`, `ValueReader`) return borrowed references.

No test: the leak is only observable through `sys.getrefcount` deltas; verified with a refcount harness over all
three sites plus a non-leaking control, flat after the fix.

### DIFF
--- a/changelog.d/python/6127.md
+++ b/changelog.d/python/6127.md
@@ -1,0 +1,3 @@
+- Fixed a memory leak in Ice for Python: the object adapter leaked a reference to the servant on each request
+  dispatched through a servant locator, and a reference to each servant returned by `findAllFacets` and
+  `removeAllFacets`.
--- a/python/modules/IcePy/Logger.h
+++ b/python/modules/IcePy/Logger.h
@@ -23,6 +23,8 @@ namespace IcePy
         void error(const std::string&) final;
         std::string getPrefix() final;
         Ice::LoggerPtr cloneWithPrefix(std::string) final;
+
+        /// Returns a borrowed reference to the Python logger.
         PyObject* getObject();
 
     private:
--- a/python/modules/IcePy/ObjectAdapter.cpp
+++ b/python/modules/IcePy/ObjectAdapter.cpp
@@ -53,6 +53,7 @@ namespace IcePy
 
         void deactivate(string_view) final;
 
+        /// Returns a new reference to the Python servant locator.
         PyObject* getObject();
 
     private:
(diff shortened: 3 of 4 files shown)


pepone (approved):
Verified the caller set is complete. There are 15 call sites for the two new-reference
`getObject()` overloads (`ServantWrapper` and `ServantLocatorWrapper`):

- `Communicator.cpp:920, 1024` and `ObjectAdapter.cpp:744, 791, 876, 916, 963, 1045, 1084,
  1155, 1194` return the value straight out of an `extern "C" PyObject*` method slot, which
  is where a new reference belongs.
- `Communicator.cpp:973` assigns into a `PyObjectHandle`, and `operator=(PyObject*)` steals
  without increfing.
- The three sites fixed here are the only ones that dropped the reference.

The borrowed siblings are fine too: both `LoggerWrapper::getObject()` callers wrap in
`Py_NewRef`, and the three `ValueReader::getObject()` callers use it as borrowed.

One nit: `LoggerWrapper::getObject()` (`Logger.h:26`) is now the only one of the four with no
annotation — `ValueReader`'s has `// Borrowed reference.` at `Types.h:514`. Worth adding
while you're documenting the contract, so the answer is in the header rather than the `.cpp`:

```cpp
/// Returns a borrowed reference to the Python logger.
PyObject* getObject();
```

PR #6204, opened 2026-07-27
Add an optional bare Object* to the Ice/optional test schema

### LINKED ISSUE #6168: Add an optional bare Object* to the Ice/optional test schema
### Summary

`MultiOptional` in the `Ice/optional` test schema has an optional proxy of an interface type (`optional(10) MyInterface* j`) but none of the bare `Object*` form. That gap hid #6125, where MATLAB's `InputStream.readProxyOpt` could never take its untyped branch — the bug was invisible because no test ever exercised it.

Tag 11 is unused in every mapping's copy of `MultiOptional` (all seven go straight from `optional(10) MyInterface* j` to `optional(12) ByteSeq bs`), so the field slots into the existing gap with no renumbering:

```slice
optional(11) Object* oprx;
```

### Scope

`Ice/optional` is on the cross-test list in `scripts/Component.py`, so the schema is shared: cpp, csharp, php, python, ruby, swift and matlab each keep a copy of `Test.ice`. Adding the field to only some of them would pass CI today — unset optionals marshal nothing, and MATLAB clients only ever run against Python servers — but it would leave the copies divergent and break the first `--cross=cpp` run against a MATLAB client. Prefer adding it everywhere.

Per mapping:

- add `optional(11) Object* oprx;` to `MultiOptional` in `test/Ice/optional/Test.ice`
- set and assert it alongside the existing `j` assertions in `AllTests`, including the all-unset cases

### Also

Remove the `bareOptionalProxies` helper added to `matlab/test/Ice/optional/AllTests.m` by #6167. It was a local stream round-trip standing in for schema coverage; once `oprx` exists it is redundant.

This also gives the other mappings their first coverage of an optional bare `Object*` — the schema gap is not MATLAB-specific, MATLAB just happened to have a bug behind it.

bernardnormier:
`MultiOptional` in the `Ice/optional` test schema had an optional proxy of an interface type (`optional(10) MyInterface* j`) but no bare `Object*` field. That gap hid #6125: MATLAB's untyped `readProxyOpt` branch was never exercised by any test.

This PR adds `optional(11) Object* oprx;` to `MultiOptional` in all eight copies of the test schema (C++, C#, Java, MATLAB, PHP, Python, Ruby, Swift) — tag 11 was unused in every copy, so no renumbering — and sets/asserts the field alongside the existing `j` assertions, including the all-unset round-trip and the alternating clear-half rounds (tag 11 is odd, so it lands in the opposite half from `j`).

The `bareOptionalProxies` stream-level helper added to the MATLAB test by #6167 is removed: it was a stand-in for schema coverage and is now redundant.

Verified by running `Ice/optional` in C++, C#, Java, Python, Ruby, PHP, and Swift, plus a C++/C# cross run; the MATLAB mapping is not buildable on this platform and relies on CI.

Fixes #6168.

### DIFF
--- a/cpp/test/Ice/optional/AllTests.cpp
+++ b/cpp/test/Ice/optional/AllTests.cpp
@@ -314,6 +314,7 @@ allTests(Test::TestHelper* helper, const Ice::SliceLoaderPtr& sliceLoader)
     mo1->h = string("test");
     mo1->i = MyEnum::MyEnumMember;
     mo1->j = MyInterfacePrx(communicator, "test");
+    mo1->oprx = Ice::ObjectPrx(communicator, "test");
     mo1->bs = ByteSeq();
     (*mo1->bs).push_back(byte{5});
     mo1->ss = StringSeq();
@@ -371,6 +372,7 @@ allTests(Test::TestHelper* helper, const Ice::SliceLoaderPtr& sliceLoader)
     test(mo3->h == string("test"));
     test(mo3->i = MyEnum::MyEnumMember);
     test(mo3->j = MyInterfacePrx(communicator, "test"));
+    test(mo3->oprx == mo1->oprx);
     test(mo3->bs == mo1->bs);
     test(mo3->ss == mo1->ss);
     test(mo3->iid == mo1->iid);
@@ -431,6 +433,7 @@ allTests(Test::TestHelper* helper, const Ice::SliceLoaderPtr& sliceLoader)
     test(!mo4->h);
     test(!mo4->i);
     test(!mo4->j);
+    test(!mo4->oprx);
     test(!mo4->bs);
     test(!mo4->ss);
     test(!mo4->iid);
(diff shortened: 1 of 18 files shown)


pepone (commented):
We should update JavaScript too:

https://github.com/bernardnormier/ice/blob/add7b75cbb31142fbbf2ede5e88c720f729a8848/js/packages/test/Ice/optional/Test.ice#L73-L74

externl (approved):

bernardnormier:
Good catch — added in the latest commit: `oprx` with tag 11 in the JavaScript `Test.ice` and the mirror assertions in `Client.ts` (unset, set, round-trip, and both alternating-half cases, following the same split as the other fields). The JS `Ice/optional` suite passes locally.

pepone (approved):

pepone (approved):
A few follow-ups, none blocking:

`csharp/test/Ice/optional/Test.ice:93` has `optional(30) Object* plainProxy;` where the eight other copies have `optional(30) Serializable ser;`. No test code references it, so it never gave us the coverage `oprx` adds — but it also leaves tag 30 incompatible across the copies (proxy is FSize, ByteSeq is VSize), so the first `--cross` run that sets `ser` against a C# peer fails with `Invalid optional tag '30': unexpected format`. Since this PR is about keeping the copies in sync, let's drop `plainProxy` here or in a follow-up.

The description still says "all eight copies ... (C++, C#, Java, MATLAB, PHP, Python, Ruby, Swift)" and omits JavaScript from the mappings you ran — worth updating before merge, it becomes the commit message.

Drive-by, pre-existing: `matlab/test/Ice/optional/AllTests.m:276` repeats `assert(isempty(mo9.j)); % optional proxy`. Judging by its position and the other mappings it should be `assert(mo9.fs == Ice.Unset);`.

PR #6207, opened 2026-07-27
Resolve a DataStorm partial update against each reader's own value

### LINKED ISSUE #6174: DataStorm: receive-side sample shared across co-subscribed readers is resolved in place, corrupting divergent-base readers
> **AI-generated finding** — identified by Claude (Anthropic), model **Claude Opus 4.8** (`claude-opus-4-8`) while analyzing the sample-dispatch path behind #5661. Every code-level claim below was verified against the source with file:line evidence, and independently cross-checked by **OpenAI Codex (GPT-5.6)**, which confirmed the diagnosis (verdict in the comments). **Not yet triaged by a human** — verify the reasoning and severity before acting.

## Summary

When a reader session receives a sample, it decodes it into a **single** `Sample` object and delivers that same object to **every** co-subscribed reader element. `DataReaderI::queue` resolves a partial update **in place** on that shared object, gated by `hasValue()`, so only the *first* co-subscribed element actually resolves the partial — the others reuse its result. When two co-subscribed elements hold **different bases** for the key, the element processed later silently receives a value resolved against the *other* element's base, and stores that poisoned value as its own base until a full update resynchronizes the key.

## Mechanism (verified against `origin/main`)

1. **One sample, shared across elements.** A writer forwards one encoded sample to a session when *any* of its subscribers match (`Listener::matchOne`, `DataElementI.h:91`). The reader session decodes it into one `shared_ptr<Sample>` (`SessionI.cpp:1573`) and queues **that same pointer** to every co-subscribed local element (`SessionI.cpp:1583-1596`), gating only on `initialized` + key membership — there is no per-element sample-filter or discard check in this loop.
2. **In-place resolution gated by `hasValue()`.** `DataReaderI::queue` resolves a partial only `if (!sample->hasValue())` (`DataElementI.cpp:1042`); the updater clones *this element's* base and writes the result back into the shared sample via `setValue` (`DataStorm.h:1655`).
3. **First reader resolves, the rest reuse.** The first co-subscribed element flips `hasValue()` to `true`; each later element then skips resolution and stores the same shared, already-resolved object as its own `_lastByKey` base (`DataElementI.cpp:1100`) and delivers it to its callback.

## When it is safe

Sharing is correct **iff all co-subscribed elements hold the same base value** for the key when the partial arrives — the overwhelming common case: same-config readers, a single writer, in-order delivery. Different **init timing** also converges, because the init/live handoff (`lastId`) guarantees a late subscriber's init batch reflects the writer's current full value with no live sample missed. Decoding a full update is deterministic, so decode-once-and-share of a full value is always safe. **Only partial resolution against a per-element base is order-sensitive.**

## When it corrupts — divergent bases

Divergent bases are reachable via per-element `DiscardPolicy` (`_discardPolicy` is per-element config, `DataElementI.cpp:704`, applied in `queue` at `:1006`; a discard does not overwrite an existing base, and a discarded sample seeds the base only when none exists, `:1023`). Concrete, legal, reachable scenario:

- Two **keyed** readers on one node for key `K`, same (default) facet: **R_A** = `DiscardPolicy::None`, **R_B** = `DiscardPolicy::Priority`. Both share the session and co-subscribe to each writer element for `K`.
- Two publishers of `K`: **W_hi** (priority 10), **W_lo** (priority 1).
- `W_hi` writes full `V_hi` → both accept → `base_A = base_B = V_hi`.
- `W_lo` writes full `V_lo` → R_A (`None`) accepts → `base_A = V_lo`; R_B (`Priority`, threshold 10 > 1) **discards** and, already having a base, is not re-seeded → `base_B = V_hi`. **Bases now diverge.**
- `W_hi` writes a **partial** `P` (priority 10) → shared to both, both accept. Whichever element the subscriber map (`std::map` keyed by `shared_ptr` address) visits **first** resolves `P` against its base and writes the result into the shared sample; the other **reuses** it. One reader ends at `V_lo+P`, the other at `V_hi+P` — exactly **one is wrong**, and that reader's `_lastByKey` base is poisoned until a full update resyncs.

This is **deterministic**, not a data race: the map order is address-ordered but stable within a process, and dispatch is serialized under locks. A `DiscardPolicy::SendTime` reader vs a `None` reader under send reordering diverges the same way.

## Second consequence — it defeats the no-base discard (#6083)

A reader with **no** base for a key normally drops a partial update rather than resolving against a default-constructed base (`DataElementI.cpp:1046-1057`, the #6083 contract). But if a co-subscriber that *does* have a base resolves the shared partial first, the no-base reader sees `hasValue() == true`, **bypasses that guard**, and stores/delivers the resolved value — **resurrecting** a key it should have no value for. So the aliasing also silently undermines the #6083 discard contract in the multi-reader case.

## Not a regression

Pre-existing. The shared-sample / resolve-once-in-place structure predates the per-key `_lastByKey` work: before `4e4ad2fb50` (#5602) the dispatch already delivered one shared sample and `queue` resolved under `!hasValue()` against the single element-wide `_last`. #5602 changed `_last` → `_lastByKey` but did not touch the `SessionI` dispatch, so it neither introduced nor fixed this.

## Severity

High data-integrity impact, medium likelihood: silent, persistent, and deterministic, but requiring ≥2 co-subscribed same-facet readers with **asymmetric discard behaviour** and multiple/reordered publishers. Correctness only — not a memory-safety or security issue.

## Recommended fix

Stop mutating the shared receive-side sample during **per-element partial resolution**: clone the sample per element (or resolve into a fresh `Sample`) before applying the updater and before storing it as the element's base. Keep decode-once-and-share for **full** values — deterministic, therefore safe. Filed against 3.8.3 since it fits the current DataStorm fix batch.

pepone:
Closes #6174

A subscriber session decoded a received sample once and queued that same object to every element subscribed to the
writer element that published it. `DataReaderI::queue` resolves a partial update in place on the sample, against the
element's own current value for the key, so the first element served wrote its result into the shared object and every
other element reused it. Elements that hold different values for the key — for instance because only some of them use a
discard policy — therefore delivered, and kept as their own value, an update resolved against another element's value.
It also bypassed the discard of a partial update that has no base value (#6083), resurrecting a key the element should
have had no value for.

`SubscriberSessionI::s` now creates a sample per element for a partial update, and keeps the single decoded sample only
for the events that no element mutates. That is the one place that knows a wire sample is about to fan out over several
elements, so the fix stays at the source of the sharing rather than in each consumer.

Decoding a full value shares the same defect, which came up while auditing the fix: `SampleT::decode` marked the sample
as having a value *before* calling the `Decoder`, so a `Decoder` that threw left the shared sample claiming a value it
never got, and the elements decoding it afterwards took the default-constructed value for the published one. The sample
is now marked only once the `Decoder` returned a value.

## Tests

Two cases at the end of `cpp/test/DataStorm/partial/{Writer,Reader}.cpp`:

- two writers (priority 10 and 1) and two readers of the same key, one with `DiscardPolicy::Priority` and one without,
  so the readers hold different values for the key when a partial update arrives;
- two co-subscribed readers and a full value the `Decoder` rejects, which both must drop.

Each fails for its own reason without its half of the fix, and the full `DataStorm` suite passes.

### DIFF
--- a/changelog.d/datastorm/6174.md
+++ b/changelog.d/datastorm/6174.md
@@ -2,6 +2,3 @@
   the current value of whichever reader was served first, instead of against each reader's own current value. Readers
   that hold different values for a key, for example because only some of them use a discard policy, now each resolve
   the update against their own value.
-- Fixed a bug where a full value that the reader's `Decoder` rejected was dropped only by the first reader of the key
-  on the node. The other readers of that key took a default-constructed value for the published one. They now all
-  drop the value.
--- a/changelog.d/datastorm/6174.md
+++ b/changelog.d/datastorm/6174.md
@@ -1,0 +1,7 @@
+- Fixed a bug where a partial update delivered to several readers of the same key on one node was resolved against
+  the current value of whichever reader was served first, instead of against each reader's own current value. Readers
+  that hold different values for a key, for example because only some of them use a discard policy, now each resolve
+  the update against their own value.
+- Fixed a bug where a full value that the reader's `Decoder` rejected was dropped only by the first reader of the key
+  on the node. The other readers of that key took a default-constructed value for the published one. They now all
+  drop the value.
--- a/cpp/include/DataStorm/InternalT.h
+++ b/cpp/include/DataStorm/InternalT.h
@@ -400,11 +400,6 @@ namespace DataStormI
             }
 
             // A custom Encoder may encode a value to zero bytes; the Decoder defines the meaning of empty input.
-            //
-            // The sample is only marked as having a value once the Decoder returned one. A received sample is shared
-            // by all the elements subscribed to the writer element that published it, so marking it beforehand would
-            // let a Decoder that throws leave the sample claiming a value it never got, and the elements that decode
-            // it after the first one would take that default-constructed value for the published one.
             _value = DecoderT<Value>::decode(communicator, _encodedValue);
             _hasValue = true;
             _encodedValue.clear();
--- a/cpp/include/DataStorm/InternalT.h
+++ b/cpp/include/DataStorm/InternalT.h
@@ -400,8 +400,13 @@ namespace DataStormI
             }
 
             // A custom Encoder may encode a value to zero bytes; the Decoder defines the meaning of empty input.
-            _hasValue = true;
+            //
+            // The sample is only marked as having a value once the Decoder returned one. A received sample is shared
+            // by all the elements subscribed to the writer element that published it, so marking it beforehand would
+            // let a Decoder that throws leave the sample claiming a value it never got, and the elements that decode
+            // it after the first one would take that default-constructed value for the published one.
             _value = DecoderT<Value>::decode(communicator, _encodedValue);
+            _hasValue = true;
             _encodedValue.clear();
         }
 
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -1610,24 +1610,39 @@ SubscriberSessionI::s(int64_t topicId, int64_t elementId, DataSample dataSample,
                 }
                 assert(key);
 
-                auto sample = topic->getSampleFactory()->create(
-                    _id,
-                    elementSubscribers->name,
-                    dataSample.id,
-                    dataSample.event,
-                    key,
-                    topicSubscriber.tags[dataSample.tag],
-                    dataSample.value,
-                    dataSample.timestamp);
+                auto createSample = [&]
+                {
+                    return topic->getSampleFactory()->create(
+                        _id,
+                        elementSubscribers->name,
+                        dataSample.id,
+                        dataSample.event,
+                        key,
+                        topicSubscriber.tags[dataSample.tag],
+                        dataSample.value,
+                        dataSample.timestamp);
+                };
+
+                // A full value is decoded into the sample the same way whatever element receives it, so the elements
+                // subscribed to this writer element share a single sample, decoded once. A partial update is not: each
+                // element resolves it against its own current value for the key and the resolved value is written into
+                // the sample, so each element gets a sample of its own. Sharing one would let the value resolved by
+                // whichever element is served first surface in all the others.
+                shared_ptr<Sample> sharedSample;
+                if (dataSample.event != DataStorm::SampleEvent::PartialUpdate)
+                {
+                    sharedSample = createSample();
+                }
 
                 for (auto& [element, elementSubscriber] : elementSubscribers->getSubscribers())
                 {
                     if (elementSubscriber.initialized &&
                         (dataSample.keyId <= 0 || elementSubscriber.keys.find(key) != elementSubscriber.keys.end()))
                     {
+                        auto elementSample = sharedSample ? sharedSample : createSample();
                         elementSubscriber.lastId = dataSample.id;
                         element->queue(
-                            sample,
+                            elementSample,
                             elementSubscribers->priority,
                             shared_from_this(),
                             current.facet,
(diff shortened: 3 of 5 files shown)


changelog.d/datastorm/6174.md
```diff
@@ -0,0 +1,7 @@
+- Fixed a bug where a partial update delivered to several readers of the same key on one node was resolved against
+  the current value of whichever reader was served first, instead of against each reader's own current value. Readers
+  that hold different values for a key, for example because only some of them use a discard policy, now each resolve
+  the update against their own value.
+- Fixed a bug where a full value that the reader's `Decoder` rejected was dropped only by the first reader of the key
+  on the node. The other readers of that key took a default-constructed value for the published one. They now all
+  drop the value.
```
bernardnormier: A `Decoder` that fails to decode a value means the application is broken (a codec bug, or an `Encoder`/`Decoder` mismatch between nodes), so this bullet describes better handling of an application error — not something a working application ever observes. Per our convention that gets no entry; the first bullet already carries this PR's user-facing news.
```suggestion
```

cpp/include/DataStorm/InternalT.h
```diff
@@ -400,8 +400,13 @@ namespace DataStormI
             }
 
             // A custom Encoder may encode a value to zero bytes; the Decoder defines the meaning of empty input.
-            _hasValue = true;
+            //
+            // The sample is only marked as having a value once the Decoder returned one. A received sample is shared
+            // by all the elements subscribed to the writer element that published it, so marking it beforehand would
+            // let a Decoder that throws leave the sample claiming a value it never got, and the elements that decode
+            // it after the first one would take that default-constructed value for the published one.
```
bernardnormier: This narrates the pre-fix bug; with the fix in place, computing the value and then marking it present is the natural order and doesn't need defending.
```suggestion
```
The underlying cleanup is worth recording where it would land — at the member declarations:
```cpp
        // TODO: make _value a std::optional and remove _hasValue.
        bool _hasValue;
```
That's a 3.9 change (`SampleT` is public-header surface), hence a TODO rather than part of this PR.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -1610,24 +1610,39 @@ SubscriberSessionI::s(int64_t topicId, int64_t elementId, DataSample dataSample,
...
+                        topicSubscriber.tags[dataSample.tag],
+                        dataSample.value,
+                        dataSample.timestamp);
+                };
+
+                // A full value is decoded into the sample the same way whatever element receives it, so the elements
+                // subscribed to this writer element share a single sample, decoded once. A partial update is not: each
+                // element resolves it against its own current value for the key and the resolved value is written into
+                // the sample, so each element gets a sample of its own. Sharing one would let the value resolved by
+                // whichever element is served first surface in all the others.
```
bernardnormier: The first two sentences carry the design rationale; the last one is the old bug.
```suggestion
                // A full value is decoded into the sample the same way whatever element receives it, so the elements
                // subscribed to this writer element share a single sample, decoded once. A partial update is not: each
                // element resolves it against its own current value for the key and the resolved value is written into
                // the sample, so each element gets a sample of its own.
```

bernardnormier (approved):
Both fixes check out against `DataReaderI::queue`: the per-element sample stops the first-served reader's resolved value (or its no-base bypass) from leaking into the others, and the decode reorder means every element re-attempts and drops an undecodable full value. The priority/no-discard test genuinely forces two readers to hold different bases. Three cleanups below.

cpp/include/DataStorm/InternalT.h
pepone: 🤖 **Claude:** Comment removed in 59acea76ca. Filed the cleanup as #6231 (3.9.0) rather than an in-code TODO.

Tracing it turned up a second, separable thing: the rest of that member comment — `encodeValue` encoding `_value` for remove samples — is dead weight. `toSample` encodes every sample unconditionally, but no receiver reads a remove's value (`DataElementI::queue` skips the `decode()` call, and `SampleT::decode` returns early anyway), so every `remove()` runs the application's `Encoder<Value>` against a value it never published and ships the result. Filed as #6232.

cpp/src/DataStorm/SessionI.cpp
pepone: 🤖 **Claude:** Applied your wording in 59acea76ca.

externl (approved):

PR #6208, opened 2026-07-27
Create the server thread pool before the incoming connection factories

### LINKED ISSUE #6188: C++ `ConnectionI::create` can throw after the constructor, tripping `~ConnectionI` assertions (UDP adapter + invalid `Ice.ServerIdleTime`)
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced while reviewing #6184 and verified by re-reading the source on `main` (`f25f15cfdc`). The exception path was **reproduced at runtime** against a local release build (trace below). The debug-build assertion failure is derived from source, not observed on a debug build. Please confirm before acting.

## Summary

`Instance::serverThreadPool()` is lazily initialized and parses `Ice.ServerIdleTime` on first call (`cpp/src/Ice/Instance.cpp:456`) — the only read site of that property in the tree. Property *values* are validated at read time, so an invalid value passes `Ice::initialize()`.

For a **UDP object adapter**, `IncomingConnectionFactory::initialize` creates the connection directly (`cpp/src/Ice/ConnectionFactory.cpp:1556`), and the first call to `adapter->getThreadPool()` therefore happens *inside* `ConnectionI::create`, **after the `ConnectionI` constructor has already returned** (`cpp/src/Ice/ConnectionI.cpp:1968`):

```cpp
Ice::ConnectionIPtr connection(new ConnectionI(...));                            // :1940 — succeeds
...
const_cast<ThreadPoolPtr&>(connection->_threadPool) = adapter->getThreadPool();  // :1968 — throws PropertyException
```

When that throws, `connection` is the sole owner — a datagram endpoint suppresses both `_inactivityTimerTask` (`_inactivityTimeout` is forced to zero) and the `IdleTimeoutTransceiverDecorator` — so unwinding destroys the connection and runs `~ConnectionI` with `_state == StateNotInitialized`, tripping `assert(_state == StateFinished)` (`cpp/src/Ice/ConnectionI.cpp:1979`).

TCP adapters are not affected: `createAcceptor` calls `getThreadPool()` before any `ConnectionI` exists (`cpp/src/Ice/ConnectionFactory.cpp:1727`).

## Reproducer

```cpp
initData.properties = Ice::createProperties();
initData.properties->setProperty("Ice.ServerIdleTime", "not-an-int");
initData.properties->setProperty("Ice.Trace.Network", "2");
Ice::CommunicatorHolder ich{Ice::initialize(std::move(initData))};
ich->createObjectAdapterWithEndpoints("Test", "udp -h 127.0.0.1 -p 0");
```

Release build (macOS, `libIce` built with `NDEBUG`):

```
Network: attempting to bind to udp socket
   local address = 127.0.0.1:0
RESULT: caught Ice::PropertyException: property 'Ice.ServerIdleTime' has an invalid integer value: 'not-an-int'
```

The bind trace pins the throw to inside the `if (_transceiver)` branch of `IncomingConnectionFactory::initialize`, whose next statement is `ConnectionI::create`. Swapping the endpoint to `tcp -h 127.0.0.1 -p 0` prints `listening for tcp connections` first and throws from `createAcceptor` instead — no `ConnectionI` involved.

## Impact

- **Release builds (`NDEBUG`): correct.** The `PropertyException` propagates cleanly out of `createObjectAdapter`, which is the desired outcome.
- **Debug builds: assertion failure / abort** in `~ConnectionI` instead of that exception.

The same applies to an invalid `Ice.ThreadPool.Server.*`, which `ThreadPool::create` parses on the same lazy path.

For completeness, the other statements in `ConnectionI::create` that run after the constructor — the `make_shared` calls, and `CreateIoCompletionPort` in the IOCP `Selector::initialize` (`cpp/src/Ice/Selector.cpp:64`) — reach the same destructor, but all of them need resource exhaustion to fail. The property path above is the only deterministic trigger.

## Fix

Either is sufficient; the first is preferable:

- **Validate the lazily-read properties at communicator initialization.** `Ice.ServerIdleTime` here, and the connection properties from #6103, follow the pattern already used for `Ice.Warn.UnusedProperties` at `cpp/src/Ice/Instance.cpp:1190` — *"Read here (not in the noexcept destroy) so an invalid value is reported when the communicator is created."* One change closes this path and the residue of #6103.
- **Or** relax `~ConnectionI`'s assertions / make `ConnectionI::create` strongly exception-safe, so a post-construction failure is not an assertion.

**Severity:** Low — deterministic, but the impact is confined to debug builds and the only way in is a configuration typo.

Related: #6103, #6033, #6058.

bernardnormier:
The server thread pool is created on demand, and its creation reads `Ice.ServerIdleTime` and the `Ice.ThreadPool.Server.*` properties. For a UDP object adapter in C++, that first read happened inside `ConnectionI::create`, after the `ConnectionI` constructor had returned: an invalid property value threw `PropertyException` at that point, and unwinding destroyed the connection with `_state == StateNotInitialized`, tripping the `assert(_state == StateFinished)` in `~ConnectionI` on debug builds. Release builds already reported the `PropertyException` cleanly, as do Java and C#, where the thread pool is acquired inside the connection constructor.

The object adapter now resolves its dispatch thread pool before creating the incoming connection factories, in all three mappings: the on-demand creation of the server thread pool — and any exception from it — happens before any acceptor or connection exists. The pool is still created by the same `createObjectAdapter` call as before, just a few steps earlier (for TCP, before the acceptor binds and listens instead of after); adapters without endpoints don't create the server thread pool, same as before.

This also removes a platform inconsistency: on iOS, `IncomingConnectionFactory::initialize` defers the acceptor creation until the app is foregrounded, so even a TCP adapter used to create the server thread pool outside `createObjectAdapter` — potentially on a notification callback, which is also where an invalid property value would have surfaced. The pool now exists from `createObjectAdapter` on every platform and transport.

With the thread pool guaranteed to exist by the time a connection is created, the try/catch around the thread pool wiring in the C# and Java `ConnectionI` constructors had no remaining live path (their `ThreadPool.initialize` doesn't throw); it is removed. C++ keeps its handling: `ThreadPool::initialize` registers with the selector, which can fail on Windows.

Each mapping's `Ice/adapterDeactivation` test now verifies that creating a UDP object adapter with an invalid `Ice.ServerIdleTime` throws `PropertyException`.

Fixes #6188.

### DIFF
--- a/cpp/src/Ice/ObjectAdapterI.cpp
+++ b/cpp/src/Ice/ObjectAdapterI.cpp
@@ -1035,20 +1035,20 @@ Ice::ObjectAdapterI::initialize(optional<RouterPrx> router)
             }
             else
             {
-                // Make sure the dispatch thread pool is created before we create the incoming connection
-                // factories: the on-demand creation of the server thread pool can throw (for example, when a
-                // thread pool property has an invalid value), and it must not fail once we start creating
-                // connections.
-                getThreadPool();
-            }
+                // An adapter with endpoints accepts incoming connections, and each connection needs the
+                // adapter's dispatch thread pool. Create it before the first connection factory: its
+                // on-demand creation reads the thread pool properties and can throw, and it must not fail
+                // once connections exist.
+                [[maybe_unused]] auto _ = getThreadPool();
 
-            for (const auto& endpoint : endpoints)
-            {
-                for (const auto& expanded : endpoint->expandHost())
+                for (const auto& endpoint : endpoints)
                 {
-                    auto factory = make_shared<IncomingConnectionFactory>(_instance, expanded, shared_from_this());
-                    factory->initialize();
-                    _incomingConnectionFactories.push_back(factory);
+                    for (const auto& expanded : endpoint->expandHost())
+                    {
+                        auto factory = make_shared<IncomingConnectionFactory>(_instance, expanded, shared_from_this());
+                        factory->initialize();
+                        _incomingConnectionFactories.push_back(factory);
+                    }
                 }
             }
         }
--- a/cpp/src/Ice/ObjectAdapterI.cpp
+++ b/cpp/src/Ice/ObjectAdapterI.cpp
@@ -1024,15 +1024,6 @@ Ice::ObjectAdapterI::initialize(optional<RouterPrx> router)
             // fill in the real port number.
             //
             vector<EndpointIPtr> endpoints = parseEndpoints(properties->getProperty(_name + ".Endpoints"), true);
-            for (const auto& endpoint : endpoints)
-            {
-                for (const auto& expanded : endpoint->expandHost())
-                {
-                    auto factory = make_shared<IncomingConnectionFactory>(_instance, expanded, shared_from_this());
-                    factory->initialize();
-                    _incomingConnectionFactories.push_back(factory);
-                }
-            }
             if (endpoints.empty())
             {
                 TraceLevelsPtr tl = _instance->traceLevels();
@@ -1040,6 +1031,24 @@ Ice::ObjectAdapterI::initialize(optional<RouterPrx> router)
                 {
                     Trace out(_instance->initializationData().logger, tl->networkCat);
                     out << "created adapter '" << _name << "' without endpoints";
+                }
+            }
+            else
+            {
+                // Make sure the dispatch thread pool is created before we create the incoming connection
+                // factories: the on-demand creation of the server thread pool can throw (for example, when a
+                // thread pool property has an invalid value), and it must not fail once we start creating
+                // connections.
+                getThreadPool();
+            }
+
+            for (const auto& endpoint : endpoints)
+            {
+                for (const auto& expanded : endpoint->expandHost())
+                {
+                    auto factory = make_shared<IncomingConnectionFactory>(_instance, expanded, shared_from_this());
+                    factory->initialize();
+                    _incomingConnectionFactories.push_back(factory);
                 }
             }
         }
--- a/cpp/test/Ice/adapterDeactivation/AllTests.cpp
+++ b/cpp/test/Ice/adapterDeactivation/AllTests.cpp
@@ -321,6 +321,25 @@ allTests(Test::TestHelper* helper)
     }
     cout << "ok" << endl;
 
+    cout << "testing UDP object adapter creation with an invalid thread pool property... " << flush;
+    {
+        Ice::InitializationData initData;
+        initData.properties = Ice::createProperties();
+        initData.properties->setProperty("Ice.ServerIdleTime", "invalid");
+        Ice::CommunicatorHolder ich{Ice::initialize(initData)};
+        try
+        {
+            ich->createObjectAdapterWithEndpoints("UdpAdapter", "udp -h 127.0.0.1 -p 0");
+            test(false);
+        }
+        catch (const Ice::PropertyException&)
+        {
+            // Expected: the server thread pool is created on demand by the UDP adapter, and its creation reads
+            // Ice.ServerIdleTime.
+        }
+    }
+    cout << "ok" << endl;
+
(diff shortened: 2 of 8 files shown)


pepone (approved):
Approach looks right — it closes the deterministic path and the iOS inconsistency at the same time. Two things on the C++ side.

**Build break.** `getThreadPool()` is `[[nodiscard]]` (`ObjectAdapterI.h:93`), so the discarded call doesn't compile under `-Werror`:

```
src/Ice/ObjectAdapterI.cpp:1042:32: error: ignoring return value of
'IceInternal::ThreadPoolPtr Ice::ObjectAdapterI::getThreadPool() const',
declared with attribute 'nodiscard' [-Werror=unused-result]
```

That's all 18 red jobs — .NET, android, matlab and iOS die at `ObjectAdapterI.o` too, so the new tests in this PR never got compiled. Idiom used elsewhere in the tree (`DescriptorHelper.cpp:713`, `IceBT/Util.cpp:68`):

```cpp
[[maybe_unused]] auto _ = getThreadPool();
```

**The `else` branch.** It reads as if creating the thread pool is conditional on tracing — that branch only exists to log "created adapter without endpoints". The real condition is that an adapter without endpoints accepts no incoming connections and must not create a server thread pool; that's the behavior being preserved, but the comment doesn't say so. Pulling the factory loop into the same branch makes it self-evident (the loop is a no-op when `endpoints` is empty anyway):

```cpp
if (endpoints.empty())
{
    TraceLevelsPtr tl = _instance->traceLevels();
    if (tl->network >= 2)
    {
        Trace out(_instance->initializationData().logger, tl->networkCat);
        out << "created adapter '" << _name << "' without endpoints";
    }
}
else
{
    // An adapter with endpoints accepts incoming connections, and each connection needs the adapter's
    // dispatch thread pool. Create it before the first connection factory: its on-demand creation reads
    // the thread pool properties and can throw, and must not fail once connections exist.
    [[maybe_unused]] auto _ = getThreadPool();

    for (const auto& endpoint : endpoints)
    {
        for (const auto& expanded : endpoint->expandHost())
        {
            auto factory = make_shared<IncomingConnectionFactory>(_instance, expanded, shared_from_this());
            factory->initialize();
            _incomingConnectionFactories.push_back(factory);
        }
    }
}
```

Same trace-driven `else` in `ObjectAdapter.cs:1159` and `ObjectAdapter.java:1118` — worth the same treatment.

externl (approved):

bernardnormier:
Thanks — both taken in the latest commit, and sorry for pushing a branch whose C++ side never compiled; the red wall of 18 jobs was exactly the `[[nodiscard]]` discard.

Applied your restructure in all three languages: the thread-pool creation and the connection-factory loop now live together in the with-endpoints branch, with your comment stating the real invariant (an adapter without endpoints accepts no connections and creates no server thread pool) instead of the previous phrasing that made the pool look tied to tracing. Used the `[[maybe_unused]] auto _` idiom as suggested.

Verified locally: C++, C# and Java build clean and `Ice/adapterDeactivation` passes in all three — including the new invalid `Ice.ServerIdleTime` cases that CI never got to compile.

bernardnormier:
The two red jobs (`ios on macos-26`, `static on ubuntu-24.04`) were caused by the new C++ test: it creates its communicator from a bare `InitializationData`, and only the shared Ice library registers IceUDP automatically — on the two static-link configurations the `udp` endpoint failed to parse (`ParseException: invalid object adapter endpoint`) before reaching the code under test.

Fixed in 7531544e8c with the established min-build idiom (`if (IceInternal::isMinBuild()) { initData.pluginFactories = {Ice::udpPluginFactory()}; }`). Verified locally on both the shared and static macOS configs — the failing case now passes on static, where it genuinely exercises the fix: the endpoint parses, adapter creation reaches the thread-pool creation, and `PropertyException` surfaces as intended.

PR #6220, opened 2026-07-28
Hold the writer's adapter to keep the DataStorm partialReconnect gap open

### LINKED ISSUE #6211: DataStorm/partialReconnect is flaky on Windows: the test races the session reattach it depends on
> **AI-generated analysis** — produced by Claude (Anthropic), model **Claude Opus 5** (`claude-opus-5`), while triaging the CI failure on #6196. Every code-level claim below was verified against the source with file:line evidence and against the CI trace logs, and independently cross-checked by **OpenAI Codex (GPT-5.6)** (verdict in the comments). **Not yet triaged by a human.**

## Summary

`DataStorm/partialReconnect`, added by #6173, fails intermittently on Windows. The assertion it makes is only reachable when the test wins a ~1–2 ms race against DataStorm's automatic session reattach, and nothing in the test synchronizes that race.

**This is a test defect. No product defect is implicated** — every interleaving produces a correct resolved value; only the number of application-visible samples changes.

## Symptom

```
cpp\test\DataStorm\partialReconnect\Reader.cpp:44:
assertion `sample.getValue() == "second-tail"' failed
```

The reader aborts (`cpp/test/include/TestHelper.h:143`), so it never creates the `resumed` barrier writer (`Reader.cpp:47`). The writer is already blocked in `getNextUnread` on its matching reader (`Writer.cpp:48`), which waits indefinitely (`cpp/src/DataStorm/DataElementI.cpp:750`). The test framework then reports the writer as hanging and fails with `Expect.TIMEOUT` after burning ~6 minutes. The timeout is collateral; the assertion is the failure.

## Observed occurrences

Mostly Windows (`windows-2022` and `windows-2025`), release config, but **not Windows-only** — see the correction comment below for a `ubuntu-24.04-arm` occurrence that surfaced as a `cancelled` job. Within ~18 hours of the test landing on `main`:

| Run | Branch |
| --- | --- |
| [30278178220](https://github.com/zeroc-ice/ice/actions/runs/30278178220) | `fix/datastorm-6084-empty-encoded-decode` |
| [30278336020](https://github.com/zeroc-ice/ice/actions/runs/30278336020) | `main` (push) |
| [30281993772](https://github.com/zeroc-ice/ice/actions/runs/30281993772) | `fix/datastorm-6088-topic-lifetime` (#6196) |
| [30290978557](https://github.com/zeroc-ice/ice/actions/runs/30290978557) | `fix-csharp-multicast-cancellation-endpoint-parse` |
| [30307542462](https://github.com/zeroc-ice/ice/actions/runs/30307542462) | `main` (push) |
| [30368819149](https://github.com/zeroc-ice/ice/actions/runs/30368819149) | `fix-service-startup-diagnostics` — `windows-2025` **and** `ubuntu-24.04-arm` (reported as `cancelled`) |

None of these branches touch the code path under test; two are pushes to `main`.

Windows loses this race by far the most often, but the race itself is not platform-specific: injecting a 1 s stall between `close().get()` and the two publications reproduces the exact assertion 3/3 on macOS (see #6220).

## Mechanism

The test intends the writer to publish a full value and a partial update *while the session is down*, so that the writer's history (`sampleCount = 0`) cannot replay them and the reader must be re-seeded from the writer's per-key base as a single resolved `Update`.

To reach that state, `writer.update("second")` and `writer.partialUpdate<string>("append")("-tail")` (`Writer.cpp:42-43`) must both execute before the session re-establishes and re-attaches. Nothing enforces this:

1. **`connection->close().get()` is not a barrier.** Ice fulfills the close future *before* invoking DataStorm's registered close callback (`cpp/src/Ice/ConnectionI.cpp:1752`, `:1804`), so when `close().get()` returns, the writer's session may not have processed the detach at all.
2. **The reattach starts immediately.** `SessionI::retry` hardcodes a zero delay for the first attempt — `auto delay = _retryCount == 0 ? 0ms : _instance->getRetryDelay(_retryCount - 1);` (`cpp/src/DataStorm/SessionI.cpp:867`, comment: *"Always retry immediately on the first attempt"*), and `_retryCount` is reset on every disconnect (`SessionI.cpp:787`). Over loopback the full reattach handshake completes in ~1–2 ms.
3. **`DataStorm.Node.RetryDelay` cannot widen the window**, because it only applies from the second attempt onward.

On Windows the application thread's wakeup from `close().get()` regularly exceeds that window, so the reattach wins and both samples are delivered as ordinary live samples.

Three interleavings are possible, and only the third satisfies the assertion:

| Reattach lands | Reader sees |
| --- | --- |
| before both updates | live `Update "second"`, then live `PartialUpdate` → `"second-tail"` |
| between the two updates | init `"second"`, then live `PartialUpdate` → `"second-tail"` |
| after both updates | one init `Update "second-tail"` ✅ |

## Trace evidence

`szN` is the initialization-batch sample count (`cpp/src/DataStorm/TraceUtil.cpp:114`); `queuing sample` is emitted by `SubscriberSessionI::s`, the *live* delivery operation (`cpp/src/DataStorm/SessionI.cpp:1565`), not the initialization path.

Failing run — empty init batch, then both samples live:

```
writer  p/1: initializing elements '[e1->e1:sz0]@0' on topic '0:partialUpdateGap'
reader  s/1: queuing sample '2[k2]' from 'e1@0' …      <- Update "second"  (live)
reader  s/1: queuing sample '3[k2]' from 'e1@0' …      <- PartialUpdate    (live)
```

Passing run on the same platform — the re-seed the test is written for:

```
writer  p/1: initializing elements '[e1->e1:sz1]@0' on topic '0:partialUpdateGap'
reader  s/1: queuing sample '4[k2]' from 'e1@0' …      <- only the later "-live" partial
```

The empty batch is correct: at reattach the writer's per-key base was still sample 1, and `collectBases` skips a base whose `id <= lastId` (`cpp/src/DataStorm/DataElementI.cpp:1545`), with the reader's `lastId` at 1.

Sibling tests using the same close-then-publish pattern (`DataStorm/reliability`) do not flake because they keep history, so replay covers either interleaving. `partialReconnect` is the only test whose assertion requires the samples to have landed in the gap.

## Fix

The gap can be held open deterministically from the test, with no product change. DataStorm installs its node adapter as the communicator's default adapter (`cpp/src/DataStorm/Instance.cpp:138`) and exposes the communicator (`cpp/include/DataStorm/Node.h:97`), and `ObjectAdapter::hold` / `waitForHold` / `activate` are reversible (`cpp/include/Ice/ObjectAdapter.h:52-62`).

Flipping the topology so the **writer** is the listener lets the writer drive the whole sequence locally and in order:

```
adapter->hold(); adapter->waitForHold();   // the reader's reattach cannot be dispatched
connection->close().get();
writer.waitForNoReaders();                 // real barrier: listener count reaches 0
writer.update("second");
writer.partialUpdate<string>("append")("-tail");
adapter->activate();                       // reader reconnects and is re-seeded from the base
```

Gating on the reader's adapter instead does not work: the reader would have no way to learn when to re-activate, because the barrier channel is exactly what is down. `DataStorm/reliability` already exercises the reader-as-client / writer-as-listener topology, so the flip is well-trodden.

pepone:
Fixes #6211.

`DataStorm/partialReconnect` asserts that a reader resuming across a gap in the writer's history is re-seeded from the writer's per-key base as a single resolved update. Reaching that state required the writer to publish both gap samples before the session re-established itself, and nothing enforced that ordering: `SessionI::retry` always uses a zero delay for the first attempt (`SessionI.cpp:867`), and Ice fulfills the close future before invoking DataStorm's close callback (`ConnectionI.cpp:1752`, `:1804`). On Windows the reattach regularly won the race, both samples were delivered live, and the reader asserted on `Reader.cpp:44`.

The writer now holds its node adapter before closing the session connection and activates it again once both samples are published. A held adapter stops reading requests, so the reader cannot re-establish the session while the gap is being created.

This requires the writer to be the listening node, so the topology is flipped. Only the node whose adapter is held can decide when the gap ends — the barrier that would otherwise carry that decision runs over the very session being interrupted. `DataStorm/reliability` already exercises the reader-as-client / writer-as-listener arrangement, including passing a `Writer` as the test-framework server.

No product code changes: every interleaving already produced a correct resolved value, only the number of application-visible samples differed. No changelog entry, as this is test-only.

## Testing

Verified that the fix is what makes the outcome deterministic, rather than the platform being faster. Injecting a one-second stall between `close().get()` and the two publications simulates the Windows scheduling delay:

| Writer body | Injected stall | Result on macOS |
| --- | --- | --- |
| before this change | 1s | 3/3 fail — `Reader.cpp:44: assertion 'sample.getValue() == "second-tail"' failed`, the exact CI signature |
| with the adapter gate | 1s | 3/3 pass |

Also confirmed from the traces that the test still exercises the path it is written for, rather than passing by skipping it — the writer's second initialization batch carries the re-seeded base and samples 2 and 3 are never delivered live:

```
writer  p/1: initializing elements '[e1->e1:sz1]@0' on topic '0:partialUpdateGap'
reader  s/1: queuing sample '1[k2]' …    <- base, live
reader  s/1: initializing samples from 'e1' …   <- resolved "second-tail"
reader  s/1: queuing sample '4[k2]' …    <- the later "-live" partial, live
```

- `python3 allTests.py --filter=DataStorm/partialReconnect` — 8/8 runs passed
- `python3 allTests.py --filter=DataStorm` — 11/11 passed

### DIFF
--- a/cpp/test/DataStorm/partialReconnect/Writer.cpp
+++ b/cpp/test/DataStorm/partialReconnect/Writer.cpp
@@ -33,15 +33,25 @@ void ::Writer::run(int argc, char* argv[])
         writer.waitForReaders();
         writer.add("base");
 
-        // Wait until the reader has consumed the base value, then close the session connection and publish while it is
-        // down. The reader cannot learn of the closure before close completes, and reconnecting takes a full session
-        // handshake, so both samples land in the gap.
+        // Wait until the reader has consumed the base value, then hold this node's adapter before closing the session
+        // connection. A held adapter stops reading requests, so the reader cannot re-establish the session until the
+        // adapter is activated again. Without it the gap is not observable: the session reconnect is attempted
+        // immediately, and the reader resumes before the two samples below are published.
         auto sample = makeSingleKeyReader(barrier, "ready").getNextUnread();
+        auto adapter = node.getCommunicator()->getDefaultObjectAdapter();
+        adapter->hold();
+        adapter->waitForHold();
+
         auto connection = node.getSessionConnection(sample.getSession());
         test(connection);
         connection->close().get();
+
+        // Publish once the session is gone, so both samples land in the gap.
+        writer.waitForNoReaders();
         writer.update("second");
         writer.partialUpdate<string>("append")("-tail");
+
+        adapter->activate();
 
         // Wait until the reader has consumed the value it resumed with, so this partial update is delivered live
         // rather than folded into the initialization batch.
--- a/cpp/test/DataStorm/partialReconnect/test.py
+++ b/cpp/test/DataStorm/partialReconnect/test.py
@@ -8,11 +8,16 @@
 # current per-key base instead, delivered as a single full update carrying the resolved value. The scenarios in
 # DataStorm/partial cover the same re-seeding for readers that join fresh, with a last sample id of 0.
 #
-# The writer opens the gap by closing the session connection and publishing immediately afterwards, which is only
-# reliable while the two nodes share a single connection. This test therefore pins the topology: multicast is disabled
-# and the writer runs as a client with no server endpoint, so the reader can never connect back to it. Given more than
-# one connection between the nodes - multicast, or a relay chain - DataStorm can route the writer's publisher session
-# and the barrier's subscriber session over different connections, and closing one would leave the other delivering.
+# The writer opens the gap by holding its node adapter and closing the session connection, then publishes and
+# activates the adapter again. A held adapter stops reading requests, so the reader cannot re-establish the session
+# while the samples are published. This requires the writer to be the node that listens, because only the node whose
+# adapter is held can decide when the gap ends: the barrier that would otherwise carry that decision runs over the
+# session being interrupted.
+#
+# The topology is pinned for the same reason the gap is: multicast is disabled and the reader has no server endpoint,
+# so a single connection carries every session between the two nodes. Given more than one connection - multicast, or a
+# relay chain - DataStorm can route the writer's publisher session and the barrier's subscriber session over different
+# connections, and closing one would leave the other delivering.
 #
 
 from DataStormUtil import Reader, Writer
@@ -24,15 +29,15 @@ traceProps = {
     "DataStorm.Trace.Data": 2,
 }
 
-# The reader listens and never connects out; the writer only connects in. One connection carries every session between
+# The writer listens and never connects out; the reader only connects in. One connection carries every session between
 # them, so closing it interrupts the writer's data session.
-readerProps = {
+writerProps = {
     "DataStorm.Node.Multicast.Enabled": 0,
     "DataStorm.Node.Server.Endpoints": "tcp -p {port1}",
     "DataStorm.Node.ConnectTo": "",
 }
 
-writerProps = {
+readerProps = {
     "DataStorm.Node.Multicast.Enabled": 0,
     "DataStorm.Node.Server.Enabled": 0,
     "DataStorm.Node.ConnectTo": "tcp -p {port1}",
@@ -43,8 +48,8 @@ TestSuite(
     [
         ClientServerTestCase(
             name="partial update after reconnect across a history gap",
-            client=Writer(props=writerProps),
-            server=Reader(props=readerProps),
+            client=Reader(props=readerProps),
+            server=Writer(props=writerProps),
             traceProps=traceProps,
         )
     ],


bernardnormier (approved):
The fix is correct and well-validated. I verified the mechanism against the source:

- `getDefaultObjectAdapter()` is guaranteed non-null and is the node's server adapter: `Instance::init` creates the `DataStorm.Node.Server` adapter and sets it as the communicator default, and the Node constructor rejects communicators that already have one.
- `hold()` really blocks re-establishment: the incoming connection factory unregisters the acceptor, so the reader's reconnect sits un-accepted in the OS backlog until `activate()` — connection validation can't complete during the gap.
- No deadlock closing a held connection: `initiateShutdown` re-registers the connection for the closing handshake and schedules the close timer, so `close().get()` completes regardless of hold state.
- Holding before closing leaves no window for the reader to slip back in, and `waitForNoReaders()` after `close().get()` covers the close future resolving before the writer-side detach.
- Writer-as-server is already exercised by DataStorm/reliability in several arrangements.

The fault-injection validation (1s stall reproducing the exact CI assertion 3/3 before, 3/3 pass after) plus the trace check that samples 2 and 3 are never delivered live is exactly the right evidence for a determinism fix.

One note: this race isn't Windows-only — run 30368819149 (PR #6205, ubuntu-24.04-arm) hit the identical signature today: reader assert at Reader.cpp:44 followed by the writer hanging on the barrier. This PR fixes that flake too; worth noting the arm occurrence on #6211.

pepone:
Thanks — and your arm sighting was right; I initially dismissed it because the job summary reports `cancelled` rather than `failure`.

Confirmed from that job's `test-logs-release-ubuntu-24.04-arm` artifact: it contains a `partialReconnect/writer-072826-1611-coredump.log` whose main thread is blocked in `DataReaderI::getNextUnread` on the `Topic<string, int>` barrier — the collateral hang after the reader aborts — and the arm trace logs carry the identical signature (`initializing elements '[e1->e1:sz0]@0'` on the writer, `queuing sample '2[k2]'` delivered live on the reader). The reader asserted, the writer hung, the job was killed, and the live log was truncated before any assertion text was emitted.

I've corrected #6211: added the arm occurrence, replaced the Windows-only claim, and preserved the coredump stack and trace excerpts inline since the artifact expires 2026-10-26. I also noted there that a `cancelled` job is not evidence of "did not fail" for this flake.

The framing I'd suggest for "not Windows-only" is the fault injection in this PR rather than the platform list: a 1 s stall between `close().get()` and the two publications reproduces the exact assertion 3/3 on macOS. Windows and arm are just where the natural wakeup delay is long enough to lose the race unaided.

PR #6222, opened 2026-07-28
Serialize the client thread pool of the communicators a node creates

### LINKED ISSUE #6195: DataStorm: sample dispatch ordering is unprotected on locally-initiated connections (the client thread pool is never configured)
> **AI-generated finding** — identified by Claude (Anthropic), model **Claude Opus 5** (`claude-opus-5`), while reviewing #6191 / #6187. Every code-level claim below was verified against the source with file:line evidence, and independently cross-checked by **OpenAI Codex (GPT-5.6)**, which confirmed the finding and its severity (verdict in the comments). **Not yet triaged by a human** — verify the reasoning and severity before acting.

## Summary

DataStorm depends on requests within one session being **dispatched** in the order they were sent. `Instance.cpp:116-118` states it outright:

> Use a serialized thread pool to ensure that samples are processed in the order they are received. This is especially important for PartialUpdate samples which depend on the order to compute the value.

That guarantee is configured on the **server adapter's** thread pool, so it covers only sessions whose traffic arrives on connections the node **accepted**. For sessions whose traffic arrives on a connection the node **initiated**, dispatch runs on `Ice.ThreadPool.Client`, which DataStorm never configures. Ordering there holds today only because that pool defaults to a single thread — by accident of the default, not by design.

## Mechanism (verified against `origin/main`)

1. **A connection's thread pool is bound at creation and never changes.** `ConnectionI::create` picks `instance->clientThreadPool()` when there is a connector (outgoing) and `adapter->getThreadPool()` otherwise (`ConnectionI.cpp:1960-1970`). Ice documents the invariant explicitly at `ConnectionI.cpp:1026-1029`: *"We never change the thread pool with which we were initially registered, even if we add or remove an object adapter."*
2. **Associating an adapter does not move the connection.** DataStorm calls `setDefaultObjectAdapter` (`Instance.cpp:138`) so outgoing connections can dispatch bidirectional requests, but `setAdapterFromAdapter` only sets `_adapter` for servant lookup (`ConnectionI.cpp:1053-1065`). Those dispatches therefore run on the **client** pool.
3. **The serialized pool is server-side only.** `DataStorm.Node.Server.ThreadPool.Serialize=1` is set inside the `DataStorm.Node.Server.Enabled > 0` branch (`Instance.cpp:110-118`). Any `<adapter>.ThreadPool.*` property makes `ObjectAdapterI` create a dedicated pool (`ObjectAdapterI.cpp:941-947`); that pool serves the adapter's accepted connections. With the server disabled, the node creates a nameless adapter with no such property (`Instance.cpp:136`).
4. **Nothing configures the client pool.** DataStorm sets no `Ice.ThreadPool.Client.*` property. Defaults: `Size=1`, `SizeMax` defaults to `Size` (`ThreadPool.cpp:189-197`), `Serialize=0` (`config/PropertyNames.xml:27-31`).
5. **Without `Serialize`, same-connection requests can execute concurrently.** Once the pool can grow, `ConnectionI` marks IO complete to elect a new leader thread *before* running the upcall (`ConnectionI.cpp:1426-1430`), and the handler is disabled for the duration of the upcall only when `_serialize` is set (`ThreadPool.cpp:738-747`). `cpp/test/Ice/hold` exists to exercise exactly this reordering.
6. **DataStorm's own locking does not restore the order.** `SessionI::_mutex` serializes mutation but is not FIFO, and there is no sequence gate before `elementSubscriber.lastId` is overwritten and the sample queued (`SessionI.cpp:1623-1630`); partial resolution then reads the mutable per-key base (`DataElementI.cpp:1042`, `:1091`).

## Who is affected

- **A reader node with `DataStorm.Node.Server.Enabled=0`** (the `Node.ConnectTo` client topology): every session request it receives arrives on its own outgoing connection, so all of it is dispatched on the client pool.
- **Peer-to-peer / multicast nodes that do have a server adapter**: the subset of sessions whose traffic arrives on the connection *that node* initiated. In that topology there are two connections per node pair, one per direction (see #6191), so roughly half the sessions land on the unprotected side.
- Collocated sessions are out of scope.

Worth noting for triage: the writer side of a `ConnectTo` pair is not the direct victim — samples flow to the reader and are dispatched there. The victim is a **reader** receiving samples over a connection it initiated.

## Reachability

`DataStorm::Node` can be constructed on an application-supplied communicator (`DataStorm/Node.h`, `Node.cpp`), so `Ice.ThreadPool.Client.Size` / `SizeMax` are under application control. An application that enlarges the client pool for its own traffic silently removes DataStorm's ordering guarantee on the affected sessions, with no warning and no visible configuration link to DataStorm. A caller-supplied `Ice::InitializationData::executor` widens this further — Ice does not require an executor to run upcalls serially.

## Not a regression

Present since the DataStorm import (`8e3d0b07ec`), which set `DataStorm.Node.Server.ThreadPool.SizeMax=1`; #4575 replaced that with `Serialize=1` so the pool could grow while staying ordered. Neither revision ever configured the client pool.

## Severity

**Medium.** Impact is high when triggered — silently wrong partial-update results and a poisoned per-key base that persists until a full update resynchronizes the key — but it requires a non-default multi-threaded client thread pool (or an unordered custom executor) together with a reader session on a locally-initiated connection. No default configuration is affected.

## Recommended fix

Make the requirement explicit instead of relying on `Ice.ThreadPool.Client.Size == 1`:

- When DataStorm creates the communicator itself (`Node.cpp:29`), set `Ice.ThreadPool.Client.Serialize=1` next to the existing server-side setting.
- When the application supplies the communicator, DataStorm should not silently change the application's dispatch behaviour — document the requirement, and consider warning at node startup if the client pool can grow beyond one thread.

`Ice.Connection.Client.MaxDispatches=1` would also serialize dispatch, but it is a heavier restriction and the default is 100.

pepone:
Fixes #6195.

## The gap

A node depends on the requests of a session being dispatched in the order they were sent — a partial update is
applied to the value left by the preceding sample — and it configures its **server adapter's** thread pool
accordingly (`Instance.cpp:118`). But a session's traffic is dispatched on the thread pool of the connection it
arrives on, and a connection's pool is fixed when the connection is created (`ConnectionI.cpp:1962`);
`setDefaultObjectAdapter` does not move it. So the sessions carried by a connection the node itself *opened* are
dispatched on `Ice.ThreadPool.Client`, which the node never configured. Their order held only because that pool
holds a single thread by default.

The whole `Node.ConnectTo` topology runs this way, and with `DataStorm.Node.Server.Enabled=0` the node can only
connect out, so 100% of its session traffic lands there.

## The fix

Seed the communicators the node creates with `Ice.ThreadPool.Client.Serialize=1`.

The client thread pool reads its configuration when the communicator creates it (`Instance.cpp:1538`, and
`ThreadPool.cpp:151` for `.Serialize`), so the property cannot be set afterwards — `Instance::init` is too late.
That leaves the two places where the node creates a communicator: `Node.cpp` and the `Node(argc, argv)`
constructor, which now builds its properties with `Ice::Properties(argc, argv, defaultProperties())` and passes
them through `InitializationData`, instead of calling `Ice::initialize(argc, argv)`.

It is the same parsing path, so `--Ice.Config`, `ICE_CONFIG`, `--DataStorm.*` parsing and the argv reshuffle are
unchanged, and the node properties are `defaults`, so a configuration file or a command-line argument still
overrides them. One behaviour difference: `Ice.ProgramName` is no longer seeded from `argv[0]`. `Instance`
derives it from the executable instead (`Instance.cpp:971-976`, basename per `Instance.cpp:132-136`), so a node
started as `./writer` now reports `writer`.

Two deliberate choices:

- **A default, not an override.** An application that sets `Ice.ThreadPool.Client.Serialize` itself keeps its own
  value.
- **A supplied communicator is left untouched.** The node cannot reconfigure a pool that already exists, and
  silently changing an application's communicator is not something a library should do. The requirement is
  documented on `NodeOptions::communicator` instead, and referenced from the `Node(CommunicatorPtr)` overload.

## Blast radius

`Ice.ThreadPool.Client` serves every outgoing connection on the communicator, and the application can reach that
communicator through `Node::getCommunicator()` — so this is worth being precise about.

**Ordering changes only where the pool can grow.** On the POSIX selectors the entire serialize path sits inside
`if (_sizeMax > 1)` (`ThreadPool.cpp:738`), and `SizeMax` defaults to `Size`, which defaults to 1. A
default-configured communicator therefore ignores the setting completely, and it changes behaviour only for one
configured to let the client pool grow — which is exactly the configuration in which the ordering is otherwise
lost.

**On Windows it is not free at the default size.** `ioCompleted` returns `_serialize && ...`
(`ThreadPool.cpp:802`) outside that size guard, and IOCP is the platform that consumes the return value
(`ThreadPool.h:255-265`): it moves `finishMessage` — which posts the next asynchronous read
(`ThreadPool.cpp:872-899`) — past the dispatch upcall. A one-thread pool cannot reorder anything, so this buys no
ordering there; it costs the connection its read/dispatch overlap, and every node pays that on its outgoing
connections. #6279 tracks giving the IOCP return the same `_sizeMax > 1` guard as POSIX, which would make the
default-size case free on Windows too.

Where the pool *can* grow, serialization means a connection does not read while it dispatches, on either
platform.

## Not doing

- **No startup warning for a supplied communicator.** It would have to re-derive the pool geometry from
  properties the pool already latched at `Ice::initialize` time and which stay mutable afterwards, so it would
  both false-positive and false-negative; it cannot see a reordering `InitializationData::executor` at all; and
  `Instance` does not know whether the communicator was application-supplied. Documentation is the honest answer.
- `Ice.Connection.Client.MaxDispatches=1` — also serializes, but it is a back-pressure knob, heavier, and it
  would throttle the application's own traffic on a shared communicator.
- Sequencing samples on the wire so DataStorm stops depending on transport dispatch order. That is the
  structural fix and a protocol change, well beyond what a non-default-configuration finding justifies.

## Testing

`cpp/test/DataStorm/config` now asserts that a node's communicator reports `Serialize == 1` — for a node built
from `argc`/`argv` and for one built from `NodeOptions`, the two paths that create a communicator — and that a
node constructed with an explicit `--Ice.ThreadPool.Client.Serialize=0` reports `0`, pinning the
default-not-override semantics.

Reverting the change makes it fail:

```
testing client thread pool configuration... failed!
config/Writer.cpp:27: assertion `...getIceProperty("Ice.ThreadPool.Client.Serialize") == "1"' failed
```

To be clear about what this covers: it pins the *configuration wiring*, which is the part that can silently
regress — a later simplification back to `Ice::initialize(argc, argv)` would set the property too late or not at
all. It does not exercise the reordering itself; that would need a multi-threaded client pool plus a locally
initiated session and would be inherently racy. The ordering guarantee `Serialize` provides is Ice's, and
`cpp/test/Ice/hold` already exercises it.

Full DataStorm suite with `OPTIMIZE=no`: 11 tests, 11 succeeded.

### DIFF
--- a/changelog.d/datastorm/6195.md
+++ b/changelog.d/datastorm/6195.md
@@ -1,7 +1,5 @@
 - Fixed a node applying the samples of a session out of order when its communicator was configured to let
-  `Ice.ThreadPool.Client` grow beyond one thread — a partial update could resolve against the wrong base value.
-  The sessions carried by a connection a node opened are dispatched on the client thread pool, which the node did
-  not configure; their order held only because that pool holds a single thread by default. A node now defaults
-  `Ice.ThreadPool.Client.Serialize` to 1 on the communicators it creates. An application that configures the
-  property itself keeps its own value, and a communicator supplied to a node is left untouched and must be
-  configured for ordered dispatch before it is created.
+  `Ice.ThreadPool.Client` grow beyond one thread — a partial update could resolve against the wrong base value. A
+  node now defaults `Ice.ThreadPool.Client.Serialize` to 1 on the communicators it creates. An application that
+  configures the property itself keeps its own value, and a communicator supplied to a node is left untouched and
+  must be configured for ordered dispatch before it is created.
--- a/changelog.d/datastorm/6195.md
+++ b/changelog.d/datastorm/6195.md
@@ -1,7 +1,7 @@
-- A node now serializes the client thread pool of the communicators it creates. A node depends on the requests it
-  receives on a connection being dispatched in the order they were sent, and the sessions carried by a connection
-  the node opened are dispatched on that pool; previously they were ordered only because the pool holds a single
-  thread by default. The setting has no effect on a single-threaded pool, and the `Node(argc, argv)` constructor
-  applies it as a default, so an application that configures `Ice.ThreadPool.Client.Serialize` itself keeps its
-  own value. A communicator supplied to a node is left untouched and must be configured for ordered dispatch
-  before it is created.
+- Fixed a node applying the samples of a session out of order when its communicator was configured to let
+  `Ice.ThreadPool.Client` grow beyond one thread — a partial update could resolve against the wrong base value.
+  The sessions carried by a connection a node opened are dispatched on the client thread pool, which the node did
+  not configure; their order held only because that pool holds a single thread by default. A node now defaults
+  `Ice.ThreadPool.Client.Serialize` to 1 on the communicators it creates. An application that configures the
+  property itself keeps its own value, and a communicator supplied to a node is left untouched and must be
+  configured for ordered dispatch before it is created.
--- a/changelog.d/datastorm/6195.md
+++ b/changelog.d/datastorm/6195.md
@@ -1,0 +1,7 @@
+- A node now serializes the client thread pool of the communicators it creates. A node depends on the requests it
+  receives on a connection being dispatched in the order they were sent, and the sessions carried by a connection
+  the node opened are dispatched on that pool; previously they were ordered only because the pool holds a single
+  thread by default. The setting has no effect on a single-threaded pool, and the `Node(argc, argv)` constructor
+  applies it as a default, so an application that configures `Ice.ThreadPool.Client.Serialize` itself keeps its
+  own value. A communicator supplied to a node is left untouched and must be configured for ordered dispatch
+  before it is created.
--- a/cpp/include/DataStorm/Node.h
+++ b/cpp/include/DataStorm/Node.h
@@ -101,10 +101,6 @@ namespace DataStorm
         Node& operator=(Node&& node) noexcept;
 
         /// Returns the Ice communicator associated with the node.
-        /// @remark For a communicator the node created, the node defaults `Ice.ThreadPool.Client.Serialize` to 1 —
-        /// the ordered dispatch it requires. An application that sets this property itself keeps its own value and
-        /// is then responsible for the ordering. The setting applies to every connection the communicator opens: a
-        /// connection does not read while it dispatches.
         [[nodiscard]] Ice::CommunicatorPtr getCommunicator() const noexcept;
 
         /// Returns the Ice connection associated with a session given a session identifier. Session identifiers are
@@ -115,8 +111,7 @@ namespace DataStorm
         [[nodiscard]] Ice::ConnectionPtr getSessionConnection(std::string_view ident) const noexcept;
 
     private:
-        /// Returns the properties a node applies to the communicators it creates. This constructor passes them as
-        /// defaults, so a configuration file or a command-line argument overrides them.
+        /// Returns the properties a node applies to the communicators it creates.
         static Ice::PropertiesPtr defaultProperties();
 
         template<typename ArgvT> NodeOptions createNodeOptions(int& argc, ArgvT argv)
@@ -124,10 +119,10 @@ namespace DataStorm
             // The node defaults have to be in place before the communicator is created: the client thread pool
             // reads its configuration when the communicator creates it.
             Ice::InitializationData initData;
-            initData.properties = defaultProperties();
+            initData.properties = std::make_shared<Ice::Properties>(argc, argv, defaultProperties());
 
             NodeOptions options;
-            options.communicator = Ice::initialize(argc, argv, std::move(initData));
+            options.communicator = Ice::initialize(std::move(initData));
             options.nodeOwnsCommunicator = true;
             return options;
         }
--- a/cpp/include/DataStorm/Node.h
+++ b/cpp/include/DataStorm/Node.h
@@ -6,6 +6,7 @@
 #include "Config.h"
 #include "InternalI.h"
 
+#include <algorithm>
 #include <functional>
 
 namespace DataStorm
@@ -27,6 +28,12 @@ namespace DataStorm
     struct NodeOptions
     {
         /// The Ice communicator used by the node. If nullptr, the node creates its own communicator.
+        /// @remark A DataStorm node requires the requests it receives on a connection to be dispatched in the order
+        /// they were sent; a partial update is applied to the value left by the preceding sample. A node configures
+        /// the communicators it creates accordingly, but it does not change the configuration of a communicator
+        /// supplied here. Such a communicator must leave `Ice.ThreadPool.Client` single-threaded, or set
+        /// `Ice.ThreadPool.Client.Serialize` to 1, before it is created. An executor set through
+        /// Ice::InitializationData must likewise preserve the dispatch order.
         Ice::CommunicatorPtr communicator{nullptr};
 
         /// Specifies whether or not the node owns the communicator.
@@ -60,7 +67,8 @@ namespace DataStorm
         /// Constructs a DataStorm node with the specified communicator.
         /// A node is the main DataStorm object. It is required to construct topics.
         /// @param communicator The communicator used by the node. If nullptr, the node creates its own communicator.
-        /// @remark This constructor sets the nodeOwnsCommunicator option to false.
+        /// @remark This constructor sets the nodeOwnsCommunicator option to false. The communicator must be
+        /// configured for ordered dispatch, as described in NodeOptions::communicator.
         explicit Node(Ice::CommunicatorPtr communicator);
 
         /// Constructs a DataStorm node with an Ice communicator initialized from command-line arguments.
@@ -95,6 +103,8 @@ namespace DataStorm
         Node& operator=(Node&& node) noexcept;
 
         /// Returns the Ice communicator associated with the node.
+        /// @remark A communicator the node created is configured for the ordered dispatch the node requires: it
+        /// sets `Ice.ThreadPool.Client.Serialize`, which applies to every connection the communicator opens.
         [[nodiscard]] Ice::CommunicatorPtr getCommunicator() const noexcept;
 
         /// Returns the Ice connection associated with a session given a session identifier. Session identifiers are
@@ -105,10 +115,31 @@ namespace DataStorm
         [[nodiscard]] Ice::ConnectionPtr getSessionConnection(std::string_view ident) const noexcept;
 
     private:
+        /// Returns the properties a node applies to the communicators it creates. This constructor passes them as
+        /// defaults, so a configuration file or a command-line argument overrides them.
+        static Ice::PropertiesPtr defaultProperties();
+
         template<typename ArgvT> NodeOptions createNodeOptions(int& argc, ArgvT argv)
         {
+            // Equivalent to Ice::initialize(argc, argv), except that the property set starts from the node
+            // defaults. They have to be in place before the communicator is created: the client thread pool reads
+            // its configuration when the communicator creates it.
+            auto properties = std::make_shared<Ice::Properties>(argc, argv, defaultProperties());
+            if (properties->getProperty("Ice.ProgramName").empty() && argc > 0)
+            {
+                Ice::StringSeq args = Ice::argsToStringSeq(argc, argv);
+                std::string programName = args[0];
+                // Replace any backslashes in this value with forward slashes, in case this value is used by the
+                // event logger.
+                std::replace(programName.begin(), programName.end(), '\\', '/');
+                properties->setProperty("Ice.ProgramName", std::move(programName));
+            }
+
+            Ice::InitializationData initData;
+            initData.properties = std::move(properties);
+
             NodeOptions options;
-            options.communicator = Ice::initialize(argc, argv);
+            options.communicator = Ice::initialize(std::move(initData));
             options.nodeOwnsCommunicator = true;
             return options;
         }
--- a/cpp/include/DataStorm/Node.h
+++ b/cpp/include/DataStorm/Node.h
@@ -6,7 +6,6 @@
 #include "Config.h"
 #include "InternalI.h"
 
-#include <algorithm>
 #include <functional>
 
 namespace DataStorm
@@ -31,9 +30,8 @@ namespace DataStorm
         /// @remark A DataStorm node requires the requests it receives on a connection to be dispatched in the order
         /// they were sent; a partial update is applied to the value left by the preceding sample. A node configures
         /// the communicators it creates accordingly, but it does not change the configuration of a communicator
-        /// supplied here. Such a communicator must leave `Ice.ThreadPool.Client` single-threaded, or set
-        /// `Ice.ThreadPool.Client.Serialize` to 1, before it is created. An executor set through
-        /// Ice::InitializationData must likewise preserve the dispatch order.
+        /// supplied here. Such a communicator must set `Ice.ThreadPool.Client.Serialize` to 1 before it is created.
+        /// An executor set through Ice::InitializationData must likewise preserve the dispatch order.
         Ice::CommunicatorPtr communicator{nullptr};
 
         /// Specifies whether or not the node owns the communicator.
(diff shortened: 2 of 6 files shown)


cpp/include/DataStorm/Node.h
```diff
@@ -27,6 +28,12 @@ namespace DataStorm
     struct NodeOptions
     {
         /// The Ice communicator used by the node. If nullptr, the node creates its own communicator.
+        /// @remark A DataStorm node requires the requests it receives on a connection to be dispatched in the order
+        /// they were sent; a partial update is applied to the value left by the preceding sample. A node configures
+        /// the communicators it creates accordingly, but it does not change the configuration of a communicator
+        /// supplied here. Such a communicator must leave `Ice.ThreadPool.Client` single-threaded, or set
+        /// `Ice.ThreadPool.Client.Serialize` to 1, before it is created. An executor set through
+        /// Ice::InitializationData must likewise preserve the dispatch order.
```
externl: Let's standardize on `Ice.ThreadPool.Client.Serialize=1` here and drop the "or leave `Ice.ThreadPool.Client` single-threaded" alternative. One rule is easier to state and to check, it doesn't rest on `SizeMax` defaulting to `Size` defaulting to 1, and it's what the node already applies to the communicators it creates.

On async dispatch: there is none in DataStorm today — no `["amd"]` in `slice/DataStorm`, and every `SessionI` servant op is a synchronous `void f(..., const Ice::Current&)`. Both options depend on that, though: under AMD neither orders completion, since both only serialize up to the point the dispatch returns.

externl (commented):
- "No-op in the default configuration" holds on POSIX only. `ioCompleted()`'s `return _serialize && ...` (`ThreadPool.cpp:802`) is outside the `_sizeMax > 1` guard and drives the IOCP-only deferral at `ThreadPool.h:260`, so on Windows `finishMessage` and the next async read are deferred until the upcall finishes even with a one-thread pool. Ordering is unaffected, but every node loses read/dispatch overlap there.
- The one red check is the known flake #6211, whose fix landed after this run. Rebase and re-run — worth a clean Windows signal given the point above.
- `Node.h:124-142` verbatim-copies `Ice::initialize(int&, ArgvT)` (`Initialize.h:97-113`), backslash comment included, and it will drift. An `Ice::initialize(int& argc, ArgvT argv, InitializationData initData)` overload would remove the duplication, and the new `<algorithm>` include from a public header with it.
- Agree with Copilot's first comment: the test covers only `Node(argc, argv)`, not the `NodeOptions`/`Node()` branch at `Node.cpp:43-48`.

pepone:
🤖 **Claude:** All four addressed in 6c18214401, on top of a rebase onto main.

**IOCP.** Confirmed, and the PR was wrong about it. On POSIX `IOScope::completed()` discards `ioCompleted()`'s return (`ThreadPool.h:215`), so the size guard is the whole story there; on IOCP it consumes it (`ThreadPool.h:255-265`), and `finishMessage` is what re-arms the read via `startAsync` (`ThreadPool.cpp:872-899`) — so the next `WSARecv` waits for the upcall at any pool size. The body, the changelog fragment and the comment in `Node.cpp` now say so, and #6279 tracks giving the IOCP return the same `_sizeMax > 1` guard as POSIX.

**Red check.** Was `partialReconnect/Reader.cpp:44: assertion `sample.getValue() == "second-tail"` failed` — #6211. Rebased onto main, which now has #6220.

**Duplication.** Added `Ice::initialize(int& argc, ArgvT argv, InitializationData initData)`; `initData.properties` supplies the `defaults` for `Ice::Properties(argc, argv, defaults)`. The existing `initialize(int&, ArgvT)` is now a one-line call to it, `createNodeOptions` is down to three lines, and `<algorithm>` is out of the public header. That refactor reaches every C++ program, so the full C++ suite was run: 106 tests, 106 succeeded.

**NodeOptions coverage.** The config test now also asserts `Serialize == 1` on a `Node{NodeOptions{}}`.

Also took your `Node.h:36` wording — one rule, `Ice.ThreadPool.Client.Serialize=1`, no single-threaded alternative.

cpp/include/DataStorm/Node.h
pepone: 🤖 **Claude:** Done — the remark now states the single rule, `Ice.ThreadPool.Client.Serialize=1`, and drops the single-threaded alternative.

cpp/test/DataStorm/config/Writer.cpp
```diff
@@ -20,6 +20,27 @@ void ::Writer::run(int argc, char* argv[])
 {
     Node node(argc, argv);
 
+    cout << "testing client thread pool configuration... " << flush;
+    {
+        // A node serializes the client thread pool of the communicator it creates: the sessions carried by a
+        // connection the node opened are dispatched there, and they must be dispatched in order.
+        test(node.getCommunicator()->getProperties()->getIceProperty("Ice.ThreadPool.Client.Serialize") == "1");
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Added — the config test now also asserts `Serialize == 1` on a `Node{NodeOptions{}}`, covering the `Node.cpp` branch.

cpp/include/DataStorm/Node.h
```diff
@@ -95,6 +103,8 @@ namespace DataStorm
         Node& operator=(Node&& node) noexcept;
 
         /// Returns the Ice communicator associated with the node.
+        /// @remark A communicator the node created is configured for the ordered dispatch the node requires: it
+        /// sets `Ice.ThreadPool.Client.Serialize`, which applies to every connection the communicator opens.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Reworded — the remark now says the node *defaults* the property to 1, that an application setting it keeps its own value and owns the ordering, and that a connection does not read while it dispatches.

changelog.d/datastorm/6195.md
```diff
@@ -0,0 +1,7 @@
+- A node now serializes the client thread pool of the communicators it creates. A node depends on the requests it
+  receives on a connection being dispatched in the order they were sent, and the sessions carried by a connection
+  the node opened are dispatched on that pool; previously they were ordered only because the pool holds a single
+  thread by default. The setting has no effect on a single-threaded pool, and the `Node(argc, argv)` constructor
+  applies it as a default, so an application that configures `Ice.ThreadPool.Client.Serialize` itself keeps its
+  own value. A communicator supplied to a node is left untouched and must be configured for ordered dispatch
+  before it is created.
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Rewritten symptom-first — it now leads with a node applying a session's samples out of order, and the partial update resolving against the wrong base value, before the serialization default.

bernardnormier (commented):
I'd rather not add another `Ice::initialize` overload — `initialize(InitializationData)` is the intended entry point for advanced use-cases, none of the other language mappings has a combined (args, initData) form, and 3.8 deliberately dropped the 3.7-era three-arg overload. The node can get the same behavior from the existing public API:

```cpp
template<typename ArgvT> NodeOptions createNodeOptions(int& argc, ArgvT argv)
{
    initData.properties = std::make_shared<Ice::Properties>(argc, argv, defaultProperties());

    NodeOptions options;
    options.communicator = Ice::initialize(std::move(initData));
    options.nodeOwnsCommunicator = true;
    return options;
}
```

`Properties(argc, argv, defaults)` gives the same precedence (command line and configuration file override the node defaults). The only difference is `Ice.ProgramName` — no longer seeded from `argv[0]`, but `Instance::initialize` derives it from the executable path when it's unset, which is fine for a node. This also drops the `Initialize.h` change and the cpp changelog fragment.

changelog.d/datastorm/6195.md
```diff
@@ -0,0 +1,7 @@
+- Fixed a node applying the samples of a session out of order when its communicator was configured to let
+  `Ice.ThreadPool.Client` grow beyond one thread — a partial update could resolve against the wrong base value.
+  The sessions carried by a connection a node opened are dispatched on the client thread pool, which the node did
+  not configure; their order held only because that pool holds a single thread by default. A node now defaults
+  `Ice.ThreadPool.Client.Serialize` to 1 on the communicators it creates. An application that configures the
+  property itself keeps its own value, and a communicator supplied to a node is left untouched and must be
+  configured for ordered dispatch before it is created.
```
bernardnormier: The third sentence narrates the internal mechanism (which thread pool dispatches what, and why it happened to work); the entry reads fine without it.

```suggestion
- Fixed a node applying the samples of a session out of order when its communicator was configured to let
  `Ice.ThreadPool.Client` grow beyond one thread — a partial update could resolve against the wrong base value. A
  node now defaults `Ice.ThreadPool.Client.Serialize` to 1 on the communicators it creates. An application that
  configures the property itself keeps its own value, and a communicator supplied to a node is left untouched and
  must be configured for ordered dispatch before it is created.
```

cpp/include/DataStorm/Node.h
```diff
@@ -95,6 +101,10 @@ namespace DataStorm
         Node& operator=(Node&& node) noexcept;
 
         /// Returns the Ice communicator associated with the node.
+        /// @remark For a communicator the node created, the node defaults `Ice.ThreadPool.Client.Serialize` to 1 —
+        /// the ordered dispatch it requires. An application that sets this property itself keeps its own value and
+        /// is then responsible for the ordering. The setting applies to every connection the communicator opens: a
+        /// connection does not read while it dispatches.
```
bernardnormier: This remark is not needed: the function just returns the communicator. The ordering requirement is already documented where the caller decides — `NodeOptions::communicator` and the `Node(communicator)` constructor.

```suggestion
```

cpp/include/DataStorm/Node.h
```diff
@@ -105,10 +115,19 @@ namespace DataStorm
         [[nodiscard]] Ice::ConnectionPtr getSessionConnection(std::string_view ident) const noexcept;
 
     private:
+        /// Returns the properties a node applies to the communicators it creates. This constructor passes them as
+        /// defaults, so a configuration file or a command-line argument overrides them.
```
bernardnormier: The second sentence documents the callers' behavior, and there is no constructor here.

```suggestion
        /// Returns the properties a node applies to the communicators it creates.
```

cpp/src/DataStorm/Node.cpp
```diff
@@ -16,6 +16,24 @@ NodeShutdownException::what() const noexcept
...
+
+    // A node depends on the requests it receives on a connection being dispatched in the order they were sent: a
+    // partial update is applied to the value left by the preceding sample. The server adapter gets that from its
+    // own serialized thread pool (see Instance::init), but the sessions carried by a connection the node opened
+    // are dispatched on the client thread pool, which the node does not otherwise configure.
+    //
+    // Ordering is at stake only for a communicator configured to let the client pool grow; a single-threaded pool
+    // cannot reorder anything. The setting is not free at that default size on Windows though: the IOCP path
+    // defers the next asynchronous read to the end of the upcall whatever the pool size, so a connection stops
+    // reading while it dispatches. See #6279.
```
bernardnormier: This comment is quite long — the platform-cost analysis and the pool-size discussion are PR-description material (and already there). The local invariant is enough:

```suggestion
    // A node requires the requests it receives on a connection to be dispatched in the order they were sent: a
    // partial update is applied to the value left by the preceding sample. The server adapter gets this from its
    // own serialized thread pool (see Instance::init); the sessions carried by connections the node opens are
    // dispatched on the client thread pool.
```

pepone:
🤖 **Claude:** Taken as written, in 4b52ec4678 — `Initialize.h` is back to main and the cpp changelog fragment is gone. Verified the three premises before reverting: 3.7 did have `initialize(int&, char*[], const InitializationData&, int)` (`Initialize.h:427` on `origin/3.7`) and 3.8 dropped it; no other mapping has a combined form (C# has `initialize(InitializationData?)` and `initialize(ref string[])`, Java `initialize(InitializationData)` / `initialize()` / `initialize(String[], List<String>)` / `initialize(String[])`); and `Instance.cpp:971-976` derives `Ice.ProgramName` when unset.

On `Ice.ProgramName` — it is the executable **basename**, not the path: `getProgramName()` strips the directory at `Instance.cpp:132-136`. A node started as `./writer` reported `./writer` before and reports `writer` now. Noted in the PR description; agreed it does not matter for a node.

The three doc/comment suggestions are applied verbatim. Full DataStorm suite with `OPTIMIZE=no`: 11 tests, 11 succeeded.

changelog.d/datastorm/6195.md
pepone: 🤖 **Claude:** Applied in 4b52ec4678.

cpp/include/DataStorm/Node.h
pepone: 🤖 **Claude:** Applied in 4b52ec4678.

cpp/include/DataStorm/Node.h
pepone: 🤖 **Claude:** Applied in 4b52ec4678.

cpp/src/DataStorm/Node.cpp
pepone: 🤖 **Claude:** Applied in 4b52ec4678.

bernardnormier (approved):
Looks good!

externl (approved):
Looks good!

PR #6223, opened 2026-07-28
Discard batch auto-flush failures instead of dropping the promise

### LINKED ISSUE #6107: JavaScript: batch auto-flush calls `ice_flushBatchRequests()` unawaited/uncaught → unhandled rejection can crash Node
> [!NOTE]

## Summary

`BatchRequestQueue.finishBatchRequest` auto-flushes by calling `proxy.ice_flushBatchRequests()` without `await` or `.catch()`. The surrounding `try/finally` only catches synchronous throws; an **asynchronous** flush failure (connection loss mid-flush, marshal error in the send path) rejects an unobserved promise → `unhandledRejection`, which terminates a Node ≥ 15 process by default. Flush failures happen precisely when a batching app hits network trouble.

**File:** `js/packages/ice/src/Ice/BatchRequestQueue.js:27-41`
```js
if (this._maxSize > 0 && this._batchStream.size >= this._maxSize) {
    proxy.ice_flushBatchRequests(); // Auto flush — promise dropped, not awaited or caught
}
```
`ice_flushBatchRequests` returns a `ProxyFlushBatch` promise (`ObjectPrxExtensions.js:274-275`).

## Fix

`proxy.ice_flushBatchRequests().catch(ex => { /* record on the queue / log */ });` — C++/C# queue the flush and surface errors through the connection rather than an unrelated caller.

**Severity:** Medium (can crash a Node process on flush failure).

pepone:
`BatchRequestQueue.finishBatchRequest` auto-flushes by calling `ice_flushBatchRequests()` and
dropping the returned promise. The surrounding `try`/`finally` only catches synchronous throws, so a
flush that fails *asynchronously* — typically because the connection cannot be established — left an
unobserved rejection behind. Node.js terminates the process on an unhandled rejection, so a batching
application could be killed by a network failure.

Reproduced against the unmodified runtime: queueing batched requests past `Ice.BatchAutoFlushSize`
against an endpoint with no listener exits the process with
`Unhandled Rejection ... Ice.ConnectionRefusedException`. With the fix the same script runs to
completion.

The fix attaches a rejection handler that discards `LocalException`, the delivery failure the other
mappings also discard — C++ passes a null exception callback (`cpp/src/Ice/BatchRequestQueue.cpp:97`),
C# discards the `Task` and Java discards the `CompletableFuture`. Anything else means a bug in the
runtime, so it is rethrown and the resulting unhandled rejection still terminates the process; this
mirrors `OutgoingConnectionFactory.flushAsyncBatchRequests`. Logging instead would be new
JavaScript-only behaviour, and if it is wanted it should be decided across mappings rather than in a
patch release. Application calls to `ice_flushBatchRequests()` are untouched and still reject
normally.

### Test

Added to the existing `Ice/operations` batch-oneway test, which already configures
`Ice.BatchAutoFlushSize=100` and relies on auto-flush firing. It queues 11 requests of 10 KiB at an
endpoint with no listener, which triggers exactly one auto-flush, then awaits an explicit flush.
Both flushes are queued on the same request handler and a connection failure completes them in
order, so once the explicit flush rejects the auto-flush has necessarily settled — no sleep needed.

Under Node this is red before the fix and green after, because the test harness turns an unhandled
rejection into `process.exit(1)` (`js/packages/test/Common/run.js:19-23`). Forcing the flush to
reject with a `TypeError` instead makes it red again, which is the intended behaviour for a bug. A
browser run is inert either way, since nothing there promotes an unhandled rejection to a failure;
that is acceptable for a Node-specific defect.

Fixes #6107

### DIFF
--- a/js/packages/ice/src/Ice/BatchRequestQueue.js
+++ b/js/packages/ice/src/Ice/BatchRequestQueue.js
@@ -1,5 +1,6 @@
 // Copyright (c) ZeroC, Inc.
 
+import { LocalException } from "./LocalException.js";
 import { Protocol } from "./Protocol.js";
 import { OutputStream } from "./OutputStream.js";
 
@@ -29,10 +30,14 @@ export class BatchRequestQueue {
 
         try {
             if (this._maxSize > 0 && this._batchStream.size >= this._maxSize) {
-                // Auto flush. We don't wait for the flush to complete, and we discard any failure, like the other
-                // language mappings. The no-op rejection handler is required because an unhandled promise rejection
-                // terminates the process in Node.js.
-                proxy.ice_flushBatchRequests().catch(() => {});
+                // Auto flush. We don't wait for the flush to complete, and we discard local exceptions. Any other
+                // error is a bug, and we rethrow it: the resulting unhandled promise rejection terminates the
+                // process in Node.js.
+                proxy.ice_flushBatchRequests().catch(ex => {
+                    if (!(ex instanceof LocalException)) {
+                        throw ex;
+                    }
+                });
             }
 
             console.assert(this._batchMarker < this._batchStream.size);
--- a/js/packages/ice/src/Ice/BatchRequestQueue.js
+++ b/js/packages/ice/src/Ice/BatchRequestQueue.js
@@ -29,7 +29,10 @@ export class BatchRequestQueue {
 
         try {
             if (this._maxSize > 0 && this._batchStream.size >= this._maxSize) {
-                proxy.ice_flushBatchRequests(); // Auto flush
+                // Auto flush. We don't wait for the flush to complete, and we discard any failure, like the other
+                // language mappings. The no-op rejection handler is required because an unhandled promise rejection
+                // terminates the process in Node.js.
+                proxy.ice_flushBatchRequests().catch(() => {});
             }
 
             console.assert(this._batchMarker < this._batchStream.size);
--- a/js/packages/test/Ice/operations/BatchOneways.ts
+++ b/js/packages/test/Ice/operations/BatchOneways.ts
@@ -69,12 +69,14 @@ export async function batchOneways(prx: Test.MyInterfacePrx, helper: TestHelper)
     await batch.ice_flushBatchRequests();
     await batch.ice_ping();
 
-    // An auto-flush that cannot be delivered fails silently: the batched requests are lost, like any other oneway,
-    // and the failure is not reported to the caller that happened to fill the batch. Nothing listens on this
-    // endpoint, so the flush triggered by Ice.BatchAutoFlushSize cannot connect.
+    // Nothing listens on this endpoint.
     const unreachable = Test.MyInterfacePrx.uncheckedCast(
         new Test.MyInterfacePrx(prx.ice_getCommunicator(), `test:${helper.getTestEndpoint(1)}`).ice_batchOneway(),
     );
+
+    // Filling the batch past Ice.BatchAutoFlushSize triggers an auto-flush that cannot connect. It fails silently:
+    // the batched requests are lost, like any other oneway, and the failure is not reported to the caller that
+    // happened to fill the batch.
     for (let i = 0; i < 11; ++i) {
         await unreachable.opByteSOneway(bs1);
     }
--- a/js/packages/test/Ice/operations/BatchOneways.ts
+++ b/js/packages/test/Ice/operations/BatchOneways.ts
@@ -2,9 +2,9 @@
 
 import { Ice } from "@zeroc/ice";
 import { Test } from "./Test.js";
-import { test } from "../../Common/TestHelper.js";
+import { TestHelper, test } from "../../Common/TestHelper.js";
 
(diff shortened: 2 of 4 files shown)


js/packages/ice/src/Ice/BatchRequestQueue.js
```diff
@@ -29,7 +29,10 @@ export class BatchRequestQueue {
 
         try {
             if (this._maxSize > 0 && this._batchStream.size >= this._maxSize) {
-                proxy.ice_flushBatchRequests(); // Auto flush
+                // Auto flush. We don't wait for the flush to complete, and we discard any failure, like the other
+                // language mappings. The no-op rejection handler is required because an unhandled promise rejection
+                // terminates the process in Node.js.
+                proxy.ice_flushBatchRequests().catch(() => {});
```
pepone:
```suggestion
                // Auto flush. We don't wait for the flush to complete, and we discard any failure, The no-op rejection handler
                // is required because an unhandled promise rejection terminates the process in Node.js.
                proxy.ice_flushBatchRequests().catch(() => {});
```

externl (approved):
Two non-blocking comments:

- `.catch(() => {})` swallows non-`LocalException` errors too — `OutgoingConnectionFactory.js:122` rethrows those. Rethrowing here would recreate the unhandled rejection, so there's no clean alternative, but worth a word in the comment.
- The regression is asserted only indirectly ("the process didn't exit"), and is inert in a browser. Referencing #6107 in the test comment would keep the intent findable.

js/packages/ice/src/Ice/BatchRequestQueue.js
pepone: 🤖 **Claude:** Applied in c93fea1d2e, with one change on top of your suggestion: `LocalException` is discarded, anything else is rethrown (see the reply to @externl below), so the handler is no longer a no-op and the comment says why. Fixed the comma splice too.

pepone:
🤖 **Claude:** Thanks @externl — on your first point, we went with the rethrow rather than a word in the comment. A rejection that is not a `LocalException` means a bug in the runtime, and hiding it is worse than the crash, so the handler now discards local exceptions only and lets everything else through, mirroring `OutgoingConnectionFactory.flushAsyncBatchRequests`. Done in c93fea1d2e.

Verified against `Ice/operations` under Node: green with the fix, red with the original `proxy.ice_flushBatchRequests();` (`Unhandled Rejection ... ConnectionRefusedException`, exit 1), and red again when the flush is forced to reject with a `TypeError` — which is the point.

On the second: leaving the test comment as is. It describes the scenario and the expected behaviour, which is the convention we follow for test comments, and the issue link is already carried by the changelog entry and this PR. Agreed the assertion is indirect and browser-inert; that is inherent to a defect whose symptom is process termination.

js/packages/test/Ice/operations/BatchOneways.ts
```diff
@@ -68,4 +68,22 @@ export async function batchOneways(prx: Test.MyInterfacePrx) {
...
+
+    // An auto-flush that cannot be delivered fails silently: the batched requests are lost, like any other oneway,
+    // and the failure is not reported to the caller that happened to fill the batch. Nothing listens on this
+    // endpoint, so the flush triggered by Ice.BatchAutoFlushSize cannot connect.
+    const unreachable = Test.MyInterfacePrx.uncheckedCast(
+        new Test.MyInterfacePrx(prx.ice_getCommunicator(), `test:${helper.getTestEndpoint(1)}`).ice_batchOneway(),
+    );
+    for (let i = 0; i < 11; ++i) {
+        await unreachable.opByteSOneway(bs1);
+    }
```
bernardnormier: This comment describes the auto-flush behavior but sits on the proxy setup; the auto-flush fires in the loop below.

```suggestion
    // Nothing listens on this endpoint.
    const unreachable = Test.MyInterfacePrx.uncheckedCast(
        new Test.MyInterfacePrx(prx.ice_getCommunicator(), `test:${helper.getTestEndpoint(1)}`).ice_batchOneway(),
    );

    // Filling the batch past Ice.BatchAutoFlushSize triggers an auto-flush that cannot connect. It fails silently:
    // the batched requests are lost, like any other oneway, and the failure is not reported to the caller that
    // happened to fill the batch.
    for (let i = 0; i < 11; ++i) {
        await unreachable.opByteSOneway(bs1);
    }
```

js/packages/test/Ice/operations/BatchOneways.ts
```diff
@@ -68,4 +68,22 @@ export async function batchOneways(prx: Test.MyInterfacePrx) {
...
+    try {
+        // Both flushes are queued on the same request handler, which completes them in order, so once this one
+        // fails the auto-flush above has already settled.
```
bernardnormier: The request-handler ordering rationale is PR-description material; the test only needs the fact that matters for its validity.

```suggestion
        // This flush fails the same way; by the time it does, the auto-flush above has settled too.
```

bernardnormier (approved):

PR #6224, opened 2026-07-28
Deliver a sliced 1.0 exception that has no patch entry

### LINKED ISSUE #6124: JavaScript Encoding 1.0: a pending value dereferences a null `_patchMap`, dropping a legitimate user exception
> [!NOTE]

## Summary

In the JavaScript Encoding 1.0 decoder, `_patchMap` starts `null` and `EncapsDecoder10.readInstance` tests the looked-up entry only against `undefined` — but the null-map case yields `l = null`, which is `!== undefined`, so the code enters the block and dereferences `null` → a raw `TypeError` instead of an Ice `MarshalException`.

**File:** `js/packages/ice/src/Ice/InputStream.js:385-389`
```js
const l = this._patchMap === null ? null : this._patchMap.get(index);
if (l !== undefined) {                 // true for null
    console.assert(l.length > 0);      // TypeError: Cannot read properties of null
```
**Trigger (corrected):** nil references alone do **not** trigger it (they add no pending instance and the sender marshals zero pending values). The reachable case is a valid derived user exception whose *unknown* derived slice carries the only class reference: the receiver `skipSlice()`s it without calling `addPatchEntry` (so `_patchMap` stays null), then `readPendingValues` → `readInstance` hits the null map. C++/C#/Java/Swift all guard an absent patch map (`_patchMap.find` on an empty `std::map` etc.).

## Fix

Null-safe lookup (`this._patchMap?.get(index)`) or tighten the guard to `l !== undefined && l !== null`, preserving the normal missing-instance marshaling check.

**Severity:** Low-Medium (legacy 1.0 client only; peer-influenced; surfaces as a rejected promise with the wrong error type, not a process crash).

pepone:
`EncapsDecoder10.readInstance` looked up the class graph depth with

```js
const l = this._patchMap === null ? null : this._patchMap.get(index);
if (l !== undefined) {
    console.assert(l.length > 0);
```

`_patchMap` is lazily created, so it is still `null` when none of the slices the reader understands
referenced a class instance. The lookup then yielded `null`, which passes the `!== undefined` guard,
and evaluating `l.length` threw a raw `TypeError` out of the Ice runtime. (`console.assert` is not
the culprit — the throw happens while evaluating its argument, and the `for (const entry of l)` on
the next line would throw too. The optimized browser bundle strips `console.assert`, and still
fails.)

This is reachable over the 1.0 encoding: when the most derived slice of a user exception is unknown,
the reader slices it off without recording a patch entry, but the sender still queued the instance
that slice referenced, so `readPendingValues` reads it with nothing to patch it into. Because
`OutgoingAsync.throwUserException` only unwinds the encapsulation for a `UserException`, the
`TypeError` propagates verbatim: the application's `await proxy.op()` rejects with a `TypeError`
rather than the base exception it declared, and it isn't even an `Ice.LocalException`, so it slips
past the usual error handling.

The fix uses an optional lookup, so an absent patch map yields `undefined`. The orphan instance is
unmarshaled and discarded and the sliced exception is delivered — exactly what C++, C#, Java and
Swift already do (`csharp/src/Ice/InputStream.cs:2412` has the equivalent `_patchMap != null &&`
guard; C++ uses a non-optional `std::map`). The separate "index for class received, but no instance"
check in `readPendingValues` is unaffected.

This was the only one of the 13 `_patchMap` references that conflated "null map" with "no entry".

### Test

Added to `Ice/stream`, which already round-trips `MyException` through `OutputStream`/`InputStream`
and already uses `Ice.Encoding_1_0`. The test marshals a derived exception that is deliberately never
registered with the Slice loader, so any reader slices it off.

The scenario has two invisible preconditions, both verified by mutation to matter:

- the sliced-off slice must carry a **non-null** class reference — with a null one the sender queues
  no instance and the path is never reached (this is why the comment calls it out);
- the base slice's own class member must be nil, otherwise a patch entry is registered and the map
  is non-null — guarded by the `ex1.c === null` assertion.

`ex1.ice_id() === "::Test::MyException"` guards against the test quietly becoming a no-op if a
`::Test::UnknownDerived` is ever added to `stream/Test.ice` and thereby registered.

Verified red→green against a pristine `origin/main` runtime: unfixed it fails with
`TypeError: Cannot read properties of null (reading 'length')`; fixed it passes. A control
round-trip of a self-referential class graph confirms the populated-patch-map path is unchanged.

Fixes #6124

### DIFF
--- a/changelog.d/js/6124.md
+++ b/changelog.d/js/6124.md
@@ -1,0 +1,3 @@
+- Fixed a bug where receiving a user exception with the 1.0 encoding could fail with a `TypeError` instead of
+  delivering the exception, when the reader sliced off a derived slice that held the only reference to a class
+  instance.
--- a/changelog.d/js/6124.md
+++ b/changelog.d/js/6124.md
@@ -1,3 +1,2 @@
-- Fixed a bug where receiving a user exception with the 1.0 encoding could fail with a `TypeError` instead of
-  delivering the exception, when the reader sliced off a derived slice that held the only reference to a class
-  instance.
+- Fixed a bug where receiving a user exception of an unknown type (but a known base type) could fail with a
+  `TypeError`. This only affects the 1.0 encoding.
--- a/js/packages/ice/src/Ice/InputStream.js
+++ b/js/packages/ice/src/Ice/InputStream.js
@@ -382,10 +382,10 @@ class EncapsDecoder10 extends EncapsDecoder {
         // keep the biggest one.
         //
         this._classGraphDepth = 0;
-        const l = this._patchMap?.get(index);
-        if (l !== undefined) {
-            console.assert(l.length > 0);
-            for (const entry of l) {
+        const patchEntries = this._patchMap?.get(index);
+        if (patchEntries !== undefined) {
+            console.assert(patchEntries.length > 0);
+            for (const entry of patchEntries) {
                 if (entry.classGraphDepth > this._classGraphDepth) {
                     this._classGraphDepth = entry.classGraphDepth;
                 }
--- a/js/packages/ice/src/Ice/InputStream.js
+++ b/js/packages/ice/src/Ice/InputStream.js
@@ -382,7 +382,7 @@ class EncapsDecoder10 extends EncapsDecoder {
         // keep the biggest one.
         //
         this._classGraphDepth = 0;
-        const l = this._patchMap === null ? null : this._patchMap.get(index);
+        const l = this._patchMap?.get(index);
         if (l !== undefined) {
             console.assert(l.length > 0);
             for (const entry of l) {
--- a/js/packages/test/Ice/stream/Client.ts
+++ b/js/packages/test/Ice/stream/Client.ts
@@ -516,6 +516,58 @@ export class Client extends TestHelper {
         }
 
         {
+            // With the 1.0 encoding, a reader slices off the derived part of an exception it doesn't know, and then
+            // still reads the class instances the sender queued after it. Here the sliced-off slice holds the only
+            // reference to such an instance, so the reader has nothing to patch it into: it must discard the
+            // instance and deliver the base exception it does know.
+            //
+            // This exception is deliberately never registered with the Slice loader, and it must carry a non-null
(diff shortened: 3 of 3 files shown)


js/packages/ice/src/Ice/InputStream.js
```diff
@@ -382,7 +382,7 @@ class EncapsDecoder10 extends EncapsDecoder {
         // keep the biggest one.
         //
         this._classGraphDepth = 0;
-        const l = this._patchMap === null ? null : this._patchMap.get(index);
+        const l = this._patchMap?.get(index);
         if (l !== undefined) {
             console.assert(l.length > 0);
             for (const entry of l) {
```
bernardnormier: Let's also rename `l` while touching this line — it's a terrible variable name.

```suggestion
        const patchEntries = this._patchMap?.get(index);
        if (patchEntries !== undefined) {
            console.assert(patchEntries.length > 0);
            for (const entry of patchEntries) {
```

changelog.d/js/6124.md
```diff
@@ -0,0 +1,3 @@
+- Fixed a bug where receiving a user exception with the 1.0 encoding could fail with a `TypeError` instead of
+  delivering the exception, when the reader sliced off a derived slice that held the only reference to a class
+  instance.
```
bernardnormier: Simpler wording — the trigger detail (which slice held which reference) is PR material, and the encoding version is kept in its own sentence:

```suggestion
- Fixed a bug where receiving a user exception of an unknown type (but a known base type) could fail with a
  `TypeError`. This only affects the 1.0 encoding.
```

bernardnormier (approved):

externl (approved):
Looks good.

js/packages/ice/src/Ice/InputStream.js
pepone: Renamed. I left the other `l` occurrences in `addPatchEntry` and `readPendingValues` alone, to keep the diff on the fix — happy to sweep them in a follow-up.

changelog.d/js/6124.md
pepone: Applied as suggested.

PR #6227, opened 2026-07-28
Track python's dependencies with the uv ecosystem

externl:
Dependabot treats `uv` as its own ecosystem, separate from `pip`, and `pip` does not read `uv.lock`.

- Switch the `/python` entry from `pip` to `uv`, so the dev group pinned in `uv.lock` — `pyright` most of all, which decides whether the Pyright job passes — actually gets bumped.
- Leave `/python/docs` on `pip`; it is a plain `requirements.txt` with no uv involved.

Depends on #6176, which is what adds `python/uv.lock`. Merging this first is harmless — the uv ecosystem reads `pyproject.toml` too — but it only does anything useful once the lockfile is there.


pepone (approved):
Looks good

PR #1715, opened 2024-01-23
Update IceUtilInternal::Random implementation to use std::random

pepone:

PR #6233, opened 2026-07-28
Generate random bytes with the OS CSPRNG instead of std::random_device

### LINKED ISSUE #6206: 'IceUtil/uuid' Intermittent Test Failure
The `IceUtil/uuid` failed today with this:
```
2026-07-27T18:04:14.3700753Z *** [12/86] Running cpp/IceUtil/uuid tests ***
2026-07-27T18:04:14.3700903Z [ running client test - 07/27/26 18:03:41 ]
2026-07-27T18:04:14.3701127Z (/ice/cpp/test/IceUtil/uuid/build/x64/shared/client env={'LD_LIBRARY_PATH': '/ice/cpp/lib64'})
2026-07-27T18:04:14.3701394Z Generating 300000 UUIDs using 3 threads... ******* iteration 1900
2026-07-27T18:04:14.3701697Z ******* Duplicate item: 00000000-0000-4000-8000-000000000000
2026-07-27T18:04:14.3701850Z failed!
2026-07-27T18:04:14.3701991Z test/IceUtil/uuid/Client.cpp:54: assertion `ok.second' failed
2026-07-27T18:04:14.3702122Z 
2026-07-27T18:04:14.3702173Z test in IceUtil/uuid failed:
2026-07-27T18:04:14.3702305Z Traceback (most recent call last):
2026-07-27T18:04:14.3702465Z   File "/ice/cpp/../scripts/Util.py", line 2029, in run
2026-07-27T18:04:14.3702628Z     self.runWithDriver(current)
2026-07-27T18:04:14.3702796Z   File "/ice/cpp/../scripts/Util.py", line 2047, in runWithDriver
2026-07-27T18:04:14.3702980Z     current.driver.runTestCase(current)
2026-07-27T18:04:14.3703186Z   File "/ice/cpp/../scripts/LocalDriver.py", line 674, in runTestCase
2026-07-27T18:04:14.3703378Z     current.testcase._runClientSide(current)
2026-07-27T18:04:14.3703558Z   File "/ice/cpp/../scripts/Util.py", line 2002, in _runClientSide
2026-07-27T18:04:14.3703731Z     self.runClientSide(current)
2026-07-27T18:04:14.3703895Z   File "/ice/cpp/../scripts/Util.py", line 1881, in runClientSide
2026-07-27T18:04:14.3704151Z     self._runClient(current, client)
2026-07-27T18:04:14.3704315Z   File "/ice/cpp/../scripts/Util.py", line 2014, in _runClient
2026-07-27T18:04:14.3704477Z     client.run(current)
2026-07-27T18:04:14.3704612Z   File "/ice/cpp/../scripts/Util.py", line 1397, in run
2026-07-27T18:04:14.3704798Z     process.waitSuccess(exitstatus=exitstatus, timeout=30)
2026-07-27T18:04:14.3704997Z   File "/ice/cpp/../scripts/Expect.py", line 758, in waitSuccess
2026-07-27T18:04:14.3705173Z     self.testExitStatus(exitstatus)
2026-07-27T18:04:14.3705344Z   File "/ice/cpp/../scripts/Expect.py", line 795, in testExitStatus
2026-07-27T18:04:14.3705523Z     test(self.exitstatus, exitstatus)
2026-07-27T18:04:14.3705681Z   File "/ice/cpp/../scripts/Expect.py", line 781, in test
2026-07-27T18:04:14.3705933Z     raise RuntimeError("unexpected exit status: expected: {0}, got {1}\n".format(expected, result))
2026-07-27T18:04:14.3706199Z RuntimeError: unexpected exit status: expected: 0, got -6
```
https://github.com/zeroc-ice/ice/actions/runs/30276264943/job/90011010735

I remember seeing an identical failure about a month ago, where the UUID was all `0`'s except for the `4` in the middle.

externl:
The IceUtil/uuid test has been failing intermittently on CI since mid-June with a duplicate `00000000-0000-4000-8000-000000000000` UUID (#6206). The root cause is not in Ice or the test: some GitHub `ubuntu-24.04` jobs now land on Azure `Standard_D4ads_v7` runners (AMD EPYC 9V45, Zen 5), where 32-bit RDSEED frequently returns 0 while reporting success (CVE-2025-62626). libstdc++ implements `std::random_device` with raw RDSEED on x86-64, so `IceInternal::generateRandom` could fill entire buffers with zeros — on an affected runner, [this probe](https://github.com/externl/ice/actions/runs/30375189802/job/90328941105) measured 4.1M zero-with-success draws out of 15M, and 66% of 16-byte fills came back all zero. Reading the OS CSPRNG avoids trusting the CPU instruction directly, on any hardware or firmware.

- `IceInternal::generateRandom` now uses `getrandom()` on Linux, `arc4random_buf()` on macOS/BSD, and `rand_s` on Windows; other platforms fail the build with `#error`
- `IceInternal::random` and `shuffle` seed their `mt19937` from the new `createMT19937()` helper instead of `std::random_device`
- Glacier2's per-session callback category is drawn directly from `generateRandom`, with modulo-bias rejection
- Documented that `Ice::generateUUID` can throw `std::system_error`, and added a changelog fragment

Fixes #6206

### DIFF
--- a/changelog.d/cpp/6206.md
+++ b/changelog.d/cpp/6206.md
@@ -1,4 +1,3 @@
-- Fixed `Ice::generateUUID` occasionally returning duplicate, all-zero UUIDs on x86-64 systems with defective
-  hardware entropy, such as AMD Zen 5 CPUs without fixed firmware (CVE-2025-62626). Ice now generates random
-  bytes — including those behind `Ice::generateUUID` and `addWithUUID` — with the operating system's CSPRNG
-  instead of `std::random_device`.
+- Fixed `Ice::generateUUID` returning duplicate, all-zero UUIDs on systems with defective hardware entropy.
+  Ice now generates random bytes — including those behind `Ice::generateUUID` and `addWithUUID` — with the
+  operating system's CSPRNG instead of `std::random_device`.
--- a/changelog.d/cpp/6206.md
+++ b/changelog.d/cpp/6206.md
@@ -1,0 +1,4 @@
+- Fixed `Ice::generateUUID` occasionally returning duplicate, all-zero UUIDs on x86-64 systems with defective
+  hardware entropy, such as AMD Zen 5 CPUs without fixed firmware (CVE-2025-62626). Ice now generates random
+  bytes — including those behind `Ice::generateUUID` and `addWithUUID` — with the operating system's CSPRNG
+  instead of `std::random_device`.
--- a/changelog.d/cpp/6206.md
+++ b/changelog.d/cpp/6206.md
@@ -1,3 +1,2 @@
-- Fixed `Ice::generateUUID` returning duplicate UUIDs on systems with defective hardware entropy.
-  Ice now generates random bytes — including those behind `Ice::generateUUID` and `addWithUUID` — with the
+- Ice now generates random bytes — including those behind `Ice::generateUUID` and `addWithUUID` — with the
   operating system's CSPRNG instead of `std::random_device`.
--- a/changelog.d/cpp/6206.md
+++ b/changelog.d/cpp/6206.md
@@ -1,3 +1,3 @@
-- Fixed `Ice::generateUUID` returning duplicate, all-zero UUIDs on systems with defective hardware entropy.
+- Fixed `Ice::generateUUID` returning duplicate UUIDs on systems with defective hardware entropy.
   Ice now generates random bytes — including those behind `Ice::generateUUID` and `addWithUUID` — with the
   operating system's CSPRNG instead of `std::random_device`.
--- a/cpp/src/Glacier2/RouterI.cpp
+++ b/cpp/src/Glacier2/RouterI.cpp
@@ -38,7 +38,6 @@ Glacier2::RouterI::RouterI(
     {
         Identity ident = {"dummy", ""};
 
-        // We use ASCII 33-126 (from ! to ~, w/o space); bytes >= 188 are rejected to avoid modulo bias.
         ident.category.reserve(20);
         while (ident.category.size() < 20)
         {
@@ -47,8 +46,10 @@ Glacier2::RouterI::RouterI(
             for (char c : buf)
             {
                 auto b = static_cast<unsigned char>(c);
+                // Reject bytes >= 188 (= 2 * 94) to avoid modulo bias.
                 if (b < 188)
                 {
+                    // Map the byte to ASCII 33-126 (from ! to ~, w/o space).
                     ident.category.push_back(static_cast<char>(33 + b % 94));
                     if (ident.category.size() == 20)
                     {
--- a/cpp/src/Glacier2/RouterI.cpp
+++ b/cpp/src/Glacier2/RouterI.cpp
@@ -1,6 +1,7 @@
 // Copyright (c) ZeroC, Inc.
 
 #include "RouterI.h"
+#include "../Ice/Random.h"
 #include "FilterManager.h"
 #include "ForwardObserver.h"
 #include "Glacier2/Session.h"
@@ -39,8 +40,7 @@ Glacier2::RouterI::RouterI(
     {
         Identity ident = {"dummy", ""};
 
-        random_device rd;
-        mt19937 gen(rd());
+        mt19937 gen = IceInternal::createMT19937();
         uniform_int_distribution<> dist(33, 126); // We use ASCII 33-126 (from ! to ~, w/o space).
         for (unsigned int i = 0; i < 20; ++i)
         {
--- a/cpp/src/Glacier2/RouterI.cpp
+++ b/cpp/src/Glacier2/RouterI.cpp
@@ -7,8 +7,6 @@
 #include "Glacier2/Session.h"
 #include "RoutingTable.h"
 
-#include <random>
-
 using namespace std;
 using namespace Ice;
 using namespace Glacier2;
@@ -40,11 +38,24 @@ Glacier2::RouterI::RouterI(
     {
         Identity ident = {"dummy", ""};
 
-        mt19937 gen = IceInternal::createMT19937();
-        uniform_int_distribution<> dist(33, 126); // We use ASCII 33-126 (from ! to ~, w/o space).
-        for (unsigned int i = 0; i < 20; ++i)
+        // We use ASCII 33-126 (from ! to ~, w/o space); bytes >= 188 are rejected to avoid modulo bias.
+        ident.category.reserve(20);
+        while (ident.category.size() < 20)
         {
-            ident.category.push_back(static_cast<char>(dist(gen)));
+            char buf[32];
+            IceInternal::generateRandom(buf, sizeof(buf));
+            for (char c : buf)
+            {
+                auto b = static_cast<unsigned char>(c);
+                if (b < 188)
+                {
+                    ident.category.push_back(static_cast<char>(33 + b % 94));
+                    if (ident.category.size() == 20)
+                    {
+                        break;
+                    }
+                }
+            }
         }
 
         const_cast<optional<ObjectPrx>&>(_serverProxy) = _instance->serverObjectAdapter()->createProxy(ident);
--- a/cpp/src/Ice/Random.cpp
+++ b/cpp/src/Ice/Random.cpp
@@ -11,7 +11,6 @@
 #include <cstdint>
 #include <cstdlib>
 #include <cstring>
-#include <stdexcept>
 #include <system_error>
 
 #if defined(__linux__)
@@ -51,7 +50,8 @@ IceInternal::generateRandom(char* buffer, size_t size)
     arc4random_buf(buffer, size);
 #elif defined(__linux__)
     // getrandom with flags == 0 deliberately blocks until the kernel's entropy pool is initialized; don't
-    // "fix" this with GRND_NONBLOCK.
+    // "fix" this with GRND_NONBLOCK. There is also deliberately no /dev/urandom fallback here: getrandom is
+    // available in every supported glibc (2.25+), so a failure is a real error, not a missing feature.
     size_t index = 0;
     while (index < size)
     {
@@ -64,7 +64,7 @@ IceInternal::generateRandom(char* buffer, size_t size)
             }
             if (n == 0)
             {
-                throw runtime_error("getrandom returned no data");
+                throw system_error(EIO, generic_category(), "getrandom returned no data");
             }
             throw system_error(errno, generic_category(), "getrandom failed");
         }
@@ -94,7 +94,7 @@ IceInternal::generateRandom(char* buffer, size_t size)
             close(fd);
             if (n == 0)
             {
-                throw runtime_error("unexpected EOF reading /dev/urandom");
+                throw system_error(EIO, generic_category(), "unexpected EOF reading /dev/urandom");
             }
             throw system_error(err, generic_category(), "cannot read /dev/urandom");
         }
--- a/cpp/src/Ice/Random.cpp
+++ b/cpp/src/Ice/Random.cpp
@@ -1,29 +1,110 @@
 // Copyright (c) ZeroC, Inc.
 
+#ifdef _WIN32
+#    define _CRT_RAND_S
+#endif
+
 #include "Random.h"
 
+#include <array>
 #include <cassert>
+#include <cstdint>
+#include <cstring>
+#include <system_error>
+
+#if defined(__APPLE__) || defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__)
+#    include <cstdlib>
+#elif defined(__linux__)
+#    include <cerrno>
+#    include <sys/random.h>
+#elif !defined(_WIN32)
+#    include <cerrno>
+#    include <fcntl.h>
+#    include <unistd.h>
+#endif
 
 using namespace std;
 
+// The random bytes must come from the operating system's CSPRNG, not from std::random_device. On x86-64,
+// libstdc++ implements std::random_device with the raw RDSEED CPU instruction, and on AMD Zen 5 CPUs without
+// fixed firmware, RDSEED frequently returns 0 while reporting success (CVE-2025-62626); this produced
+// duplicate all-zero UUIDs. The OS conditions hardware entropy sources instead of trusting their output.
 void
 IceInternal::generateRandom(char* buffer, size_t size)
 {
-    // We use the random_device directly here to get cryptographic random numbers when possible.
-    thread_local static std::random_device rd;
-    uniform_int_distribution<unsigned int> distribution(0, 255);
-    for (size_t i = 0; i < size; ++i, ++buffer)
+#if defined(_WIN32)
+    // rand_s is backed by the same system CSPRNG as BCryptGenRandom and requires no additional link
+    // dependency.
+    size_t index = 0;
+    while (index < size)
     {
-        *buffer = static_cast<char>(distribution(rd));
+        unsigned int r = 0;
+        errno_t err = rand_s(&r);
+        if (err != 0)
+        {
+            throw system_error(err, generic_category(), "rand_s failed");
+        }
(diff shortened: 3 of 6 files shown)


pepone (commented):
Reopened this — I went through the libstdc++ side on #6206 and there is no upstream mitigation filed or in flight, so option 1 is the fix we control. The approach here is right, and I checked the three platform paths against their contracts. Four things I'd fix before merging, two suggestions, then some notes.

## Should fix

**1. `rand_s` needs `<cstdlib>` included explicitly.** The Windows branch is the only one that doesn't include the header its API comes from — it currently compiles through a transitive include (`<random>`, via `Random.h`). Add `#include <cstdlib>` to the unconditional include block. The `#define _CRT_RAND_S` at the top of the file already precedes it, which is what matters.

**2. `open("/dev/urandom", O_RDONLY)` is missing `O_CLOEXEC`.** IceGrid's `Activator` forks and execs servers, so this descriptor would leak into every activated process. `O_RDONLY | O_CLOEXEC`.

**3. The Linux loop can spin on `n == 0`.**

```cpp
ssize_t n = getrandom(buffer + index, size - index, 0);
if (n < 0) { ... }
index += static_cast<size_t>(n);
```

`getrandom` shouldn't return 0 for a non-zero request, but the loop shouldn't be *able* to spin forever if it does. Treat `n <= 0` as the error case (retrying only `n < 0 && errno == EINTR`), matching the `/dev/urandom` branch below it.

**4. The `/dev/urandom` branch reports a stale `errno` on EOF.** `read` returning 0 falls into the error path and throws `system_error(errno, ...)` with whatever `errno` happened to hold from an earlier unrelated call. Worth distinguishing the two, e.g. throw a `runtime_error("unexpected EOF reading /dev/urandom")` for `n == 0`. Same branch: `open` isn't retried on `EINTR`.

## Suggestions

**5. Generate the Glacier2 category with `generateRandom` directly.** `RouterI.cpp` now builds a fresh 2.5 KB `mt19937` and consumes 32 bytes of OS entropy per session, only to draw 20 values from it. Twenty bytes straight from `generateRandom` is simpler and doesn't route a routing token through a non-cryptographic engine. One catch if you do it: mapping bytes into 33..126 with `% 94` is biased (256 = 2·94 + 68), so reject bytes >= 188.

**6. This warrants a changelog fragment** — `changelog.d/cpp/<slug>.md`. It changes the documented source of `Ice::generateUUID` and it protects deployments running on unpatched Zen 5 hardware, not just CI. Something like:

```
- Ice now generates random bytes — including the ones behind `Ice::generateUUID` and `addWithUUID` — with the
  operating system's CSPRNG instead of `std::random_device`, which on some x86-64 systems reads a CPU instruction
  that can silently return zeros.
```

## Notes, no action needed

**7. `getrandom(..., 0)` blocks until the pool is initialized.** That's the right choice here, but it's the kind of thing someone later "fixes" to `GRND_NONBLOCK`. Worth a sentence in the comment saying the block is deliberate.

**8. `generateRandom` now throws `std::system_error`.** `Ice::generateUUID` is public API, so it's worth a line in `Ice/UUID.h`. Not a regression — `std::random_device` could throw too.

**9. Worth stating explicitly, because it answers the obvious objection to this PR:** #1715 removed a `/dev/urandom` implementation with a static mutex, a cached fd, and a `pthread_atfork` child handler to reset it after fork. None of that comes back here — `getrandom` and `rand_s` are stateless syscalls, `arc4random_buf` handles fork internally, and the fallback opens and closes per call. The complexity that justified the 2024 change isn't what's being reintroduced.

**10. Scope is right.** Python, PHP, Ruby and MATLAB have no UUID generator of their own and inherit this through the C++ core (`python/modules/IcePy/ObjectAdapter.cpp:609-636`, `php/src/Util.cpp:698-706`), so this fixes five mappings. Java, C#, Swift and JavaScript generate their own and are unaffected by *this* bug — I've filed the independent weaknesses I found in them as #6260 (JavaScript UUIDs) and #6261 (C#/Java WebSocket masking).

**11. No new test needed.** `IceUtil/uuid` is the regression check, and an OS CSPRNG isn't meaningfully unit-testable beyond that.

One performance observation, not a concern: this is now one syscall per call, i.e. one per UUID. `arc4random_buf` buffers in userspace, Linux 6.11+ serves `getrandom` from the vDSO, and `addWithUUID` isn't on a hot path — but it's a change from 16 `RDSEED`s with no syscall, so worth knowing if anyone profiles adapter-heavy startup.

externl:
All applied: unconditional `<cstdlib>`, `O_CLOEXEC` + EINTR retry on `open`, `n <= 0` handled in both read loops with a distinct EOF error, category drawn directly from `generateRandom` with bias rejection, changelog fragment `cpp/6206.md`, blocking-`getrandom` comment, and `@throws` on `generateUUID`.

pepone (commented):
Re-reviewed at 4de94904e. All eight items are applied, and applied correctly — I re-checked the bias math and the two rewritten loops rather than taking the summary at face value.

**Verified**

- `<cstdlib>` is unconditional and still after `#define _CRT_RAND_S`, so `rand_s` no longer depends on a transitive include.
- `open` retries `EINTR` and sets `O_CLOEXEC`.
- Linux loop: `n == 0` throws instead of spinning, and the `EINTR` retry is correctly guarded by `n < 0` so it can't be reached with a stale `errno`.
- `/dev/urandom`: EOF is distinguished from a real error, `errno` is captured before `close(fd)`, and both throw paths close the descriptor.
- Category: 188 = 2·94, so bytes 0..187 map uniformly onto 33..126 — the rejection threshold is right. A 32-byte draw yields ~23.5 accepted bytes, so it finishes on the first pass with overwhelming probability.
- `changelog.d/cpp/6206.md` is in the right section with the right `<id>` — `changelog.d/README.md` asks for the issue number when a PR fixes a single public issue.
- Blocking-`getrandom` comment and the `@throws` line on `generateUUID`.

`std::random_device` is now gone from `cpp/src` and `cpp/include` entirely. Five test files still use it (`test/Glacier2/router`, `test/Glacier2/attack`, `test/IceGrid/allocation`, `test/IceStorm/stress`) — harmless, but they're the ones that would behave oddly on affected hardware if anyone ever chases a flake there.

**One thing to fix**

`Ice/UUID.h` documents `@throws std::system_error`, but the two new degenerate paths throw plain `std::runtime_error` (`"getrandom returned no data"`, `"unexpected EOF reading /dev/urandom"`). Those aren't `system_error`, so a caller that catches exactly what the doc promises misses them. I'd keep a single exception type on the public path rather than widen the doc:

```cpp
throw system_error(EIO, generic_category(), "getrandom returned no data");
```

**Two suggestions**

1. **Say that the missing Linux fallback is deliberate.** Every non-`EINTR` `getrandom` failure throws, so Linux never reaches the `/dev/urandom` branch below it. That's the right call — `getrandom` has been in glibc since 2.25 and every platform we support is well past that — but the code doesn't say so, and having the fallback sitting right there invites someone to wire it up later for an `ENOSYS` that can't happen. One sentence next to the blocking-`getrandom` comment would settle it.

2. **The changelog says "all-zero UUIDs", which isn't quite what anyone sees.** `UUID.cpp:57-61` forces the version and variant bits, so the degenerate value is `00000000-0000-4000-8000-000000000000` — which is exactly what #6206 reported. Since the point of the entry is that an affected user recognizes their symptom, I'd either name that value or just say "duplicate UUIDs".

**One for a follow-up, not this PR**

`SessionRouterI.cpp:1143` registers the new category with `assert(rc.second)`. `NDEBUG` is defined in every optimized build, so a duplicate category is not diagnosed at all — the insert fails, `_routersByCategoryHint` is left pointing at the *other* session's entry, and destroying either session then erases the shared mapping at `:882-887`. This PR is what makes a duplicate essentially unreachable, so nothing here is wrong; it's just that the one check that would have made #6206 fail loudly instead of misrouting is still an assert.

It isn't a two-line change, which is why I'd keep it out of this PR: the router is already registered by connection at `:1137` and its close callback isn't installed until `:1148`, so a bare `throw` there leaks a half-registered session. It needs either a check ahead of the connection insert or a rollback on the failure path, plus the router destruction the `_destroy` branch above it does.

**Correcting my earlier note**

I said the extra syscall wasn't on a hot path. That's true for `addWithUUID`, but `WSTransceiver.cpp:1868` draws a fresh 32-bit mask for *every outgoing frame* on a client-side ws/wss connection, so `generateRandom` is now on the per-frame write path. It may well be a wash or a win — the old path made four `random_device` draws per mask, and on x86-64 that's a CPU entropy instruction each, which isn't cheap either, while `getrandom` is vDSO-served on Linux 6.11+ and `arc4random_buf` is pure userspace. But neither of us has measured it, and it's the one place in this change where the cost is per-message rather than per-session.

CI is still mostly pending; `debian12-arm64` is green, which is the build that compiles the `getrandom` branch.

pepone (approved):
Approving — nothing left here is blocking.

The `@throws std::system_error` / plain `runtime_error` mismatch and the two nits are worth picking up while you're on the branch, but none of them gate the merge. The `SessionRouterI` category assert is a separate follow-up, not this PR's job.

externl:
Applied: both degenerate paths now throw `system_error(EIO, ...)` so the `@throws` doc is accurate, the Linux branch comment states the missing `/dev/urandom` fallback is deliberate (glibc 2.25+), and the changelog now says just "duplicate UUIDs". Agreed the `SessionRouterI` duplicate-category handling is a separate follow-up.

changelog.d/cpp/6206.md
```diff
@@ -0,0 +1,3 @@
+- Fixed `Ice::generateUUID` returning duplicate UUIDs on systems with defective hardware entropy.
+  Ice now generates random bytes — including those behind `Ice::generateUUID` and `addWithUUID` — with the
+  operating system's CSPRNG instead of `std::random_device`.
```
bernardnormier: Let's frame this as an improvement rather than a fix "on systems with defective hardware entropy" — the OS CSPRNG is simply the right source on any hardware, and the RDSEED story can stay in the PR/issue.

```suggestion
- Ice now generates random bytes — including those behind `Ice::generateUUID` and `addWithUUID` — with the
  operating system's CSPRNG instead of `std::random_device`.
```

cpp/src/Ice/Random.h
```diff
@@ -9,13 +9,18 @@
 
 namespace IceInternal
 {
+    // Fills the buffer with cryptographically strong random bytes from the operating system's CSPRNG.
+    // See Random.cpp for why this must not draw from std::random_device.
```
bernardnormier: The first sentence stands on its own; the rationale already lives in Random.cpp, right where someone tempted to change the implementation will be looking.

```suggestion
    // Fills the buffer with cryptographically strong random bytes from the operating system's CSPRNG.
```

cpp/src/Glacier2/RouterI.cpp
```diff
@@ -39,12 +38,24 @@ Glacier2::RouterI::RouterI(
...
         {
-            ident.category.push_back(static_cast<char>(dist(gen)));
+            char buf[32];
+            IceInternal::generateRandom(buf, sizeof(buf));
+            for (char c : buf)
+            {
+                auto b = static_cast<unsigned char>(c);
+                if (b < 188)
+                {
+                    ident.category.push_back(static_cast<char>(33 + b % 94));
```
bernardnormier: This comment sits on the setup but explains the two lines below; split it so each half sits on the line it explains.

```suggestion
        ident.category.reserve(20);
        while (ident.category.size() < 20)
        {
            char buf[32];
            IceInternal::generateRandom(buf, sizeof(buf));
            for (char c : buf)
            {
                auto b = static_cast<unsigned char>(c);
                // Reject bytes >= 188 (= 2 * 94) to avoid modulo bias.
                if (b < 188)
                {
                    // Map the byte to ASCII 33-126 (from ! to ~, w/o space).
                    ident.category.push_back(static_cast<char>(33 + b % 94));
```

cpp/src/Ice/UUID.cpp
```diff
@@ -2,8 +2,8 @@
 
 #include "Ice/UUID.h"
 
-// We use a high quality random number generator
-// (std::random_device) to generate "version 4" UUIDs, as described in
+// We use the operating system's CSPRNG (see Random.h) to generate
+// "version 4" UUIDs, as described in
```
bernardnormier: Nit: reflow on 120 columns.

```suggestion
// We use the operating system's CSPRNG (see Random.h) to generate "version 4" UUIDs, as described in
```

cpp/src/Ice/Random.cpp
```diff
@@ -1,29 +1,123 @@
...
+#include <cstring>
+#include <system_error>
+
+#if defined(__linux__)
+#    include <cerrno>
+#    include <sys/random.h>
+#elif !defined(_WIN32) && !defined(__APPLE__) && !defined(__FreeBSD__) && !defined(__OpenBSD__) && !defined(__NetBSD__)
+#    include <cerrno>
+#    include <fcntl.h>
+#    include <unistd.h>
```
bernardnormier: With the `/dev/urandom` branch removed (see below), these includes go too.

```suggestion
```

cpp/src/Ice/Random.cpp
```diff
@@ -1,29 +1,123 @@
...
+            close(fd);
+            if (n == 0)
+            {
+                throw system_error(EIO, generic_category(), "unexpected EOF reading /dev/urandom");
+            }
+            throw system_error(err, generic_category(), "cannot read /dev/urandom");
+        }
+        index += static_cast<size_t>(n);
     }
+    close(fd);
```
bernardnormier: Every supported platform takes one of the three branches above, so this block never compiles on any platform we build or test — let's not add more never-compiled code. An `#error` tells a future porter exactly what to provide, instead of silently handing them an untested code path.

```suggestion
#else
#    error "unsupported platform: no OS CSPRNG binding"
```

externl:
All five applied: the `#else` branch is now `#error` (and its includes are gone), `Random.h` keeps just the one-line doc, the UUID.cpp comment is reflowed, and the category comment is split as suggested.

bernardnormier (approved):
Looks good.

pepone (approved):
Re-reviewed the three commits since 4de94904e. All clean — approval stands.

**The `@throws` mismatch is fixed.** Both degenerate paths now throw
`system_error(EIO, generic_category(), ...)`, so the doc on `Ice::generateUUID` is accurate.
`<stdexcept>` went out with the last `runtime_error`, and `system_error` is now the only
exception type on the public path.

**I checked the `#error` rather than assuming it**, since it's the one change here that fails
at porting time instead of loudly in CI. The only exotic platform macros anywhere in `cpp/src`
(`__hpux`, `_AIX`) are in Bison-generated parser files, not real ports — so Windows /
Apple+BSD / Linux genuinely covers everything we build. Android takes the `__linux__` branch
(bionic has `getrandom` since API 28) and both `android` and `android-bt` are green; iOS is
`__APPLE__`, green. It's also a behavioral no-op: on Linux every non-`EINTR` failure already
threw before reaching the fallback, so that branch was dead on the only platform that could
have used it.

**Includes are clean after the removal.** `<cerrno>`/`<fcntl.h>`/`<unistd.h>` went with the
branch, and everything left is still used — `<cstdlib>`/`<cstring>` for `rand_s` + `memcpy`,
`<array>`/`<cstdint>` for `createMT19937`, `<cassert>` for `random`, `<system_error>` for the
throws.

**One thing I hadn't checked in the earlier rounds.** Since I corrected myself about
`WSTransceiver.cpp:1868` being on the per-frame write path, I went back to confirm a
`getrandom` failure there unwinds rather than terminating: `WSTransceiver.h` declares nothing
`noexcept`, so it propagates normally. No `std::terminate` hazard on the highest-frequency
caller.

CI: one red, `release on ubuntu-24.04-arm` —
`test/Ice/inactivityTimeout/AllTests.cpp:68: assertion 'connection2 == connection' failed`.
Known flake, not a compile failure, and not in a random-consuming test.

PR #6234, opened 2026-07-28
Type hint the rest of the Python sources

externl:
Follows #6176. Pyright's `include` now covers the whole repository rather than just `python` and `scripts`, and everything outside `python/` type checks clean — 249 errors across 32 files.

- Put `scripts` on the top-level search path: the `test.py` files under `cpp/test` and friends import the framework by name, the way `loadTestSuites` runs them.
- Type hint `config/makeprops.py`, the densest file left at 74 errors. Its generated file handles are `None` until `openFiles` runs, so every write and close went unchecked; `createProperty`'s overrides named their last parameter differently from the base, which makes the override incompatible by keyword; and a section's name reached a dict key as `str | None`. Regenerating `PropertyNames` for C++, Java and C# gives output identical to what is checked in.
- Type hint the 30 `test.py` files under `cpp/test` and `php/test`.
- `IceGridTestCase` and `RelayReconnectTestCase` took `list`s their callers could not satisfy, since `list` is invariant and both are given a list of a subclass. They take a `Sequence`, as elsewhere in the framework.
- `IceGridRegistry` gains the `createDb` flag `IceStorm` already has, so `icegriddb` can keep `setup` from wiping the database it imported instead of replacing the method with a lambda. `icestormdb` does the same through the existing flag.
- `Glacier2Router`'s `passwords` is optional; `dynamicFiltering` passes `None` and both readers test it for truthiness.
- Write `makeprops.py`'s diagnostics to stderr. All ten were `print(sys.stderr, "...")`, which passes the stream as a value to print rather than as its destination, so each message went to stdout with the stream's repr in front of it.
- Fix things the annotations exposed: `Slice/headers` rebound `slice2cpp` from the translator to its command line, leaving a nested call to `.run()` on a `str`; `staticFiltering` reset the suite's `OrderedDict` by rebinding it to a plain dict; `repstress` read `.group(1)` straight off a match that can be `None`; and `stress` carried a dead `pubSub()` that built a test case it discarded and passed dicts where a process wants arguments.

### DIFF
--- a/config/makeprops.py
+++ b/config/makeprops.py
@@ -9,9 +9,10 @@ import shutil
 import signal
 import sys
 from enum import StrEnum, auto
-from typing import override
+from typing import TextIO, override
 from xml.sax import make_parser
 from xml.sax.handler import ContentHandler, feature_namespaces
+from xml.sax.xmlreader import AttributesImpl
 
 
 class Language(StrEnum):
@@ -37,49 +38,49 @@ class PropertyArray:
         self.prefixOnly = prefixOnly
         self.isOptIn = isOptIn
         self.isClass = isClass
-        self.properties = []
+        self.properties: list[str] = []
 
-    def addProperty(self, property):
+    def addProperty(self, property: str) -> None:
         self.properties.append(property)
 
 
 class PropertyHandler(ContentHandler):
-    def __init__(self, language):
+    def __init__(self, language: Language):
         # The language we are generating properties for
         self.language = language
         # The section we are currently parsing
-        self.parentNodeName = None
+        self.parentNodeName: str | None = None
         # Dictionary of section names to attr dicts
         self.propertyArrayDict: dict[str, PropertyArray] = dict()
 
-    def cleanup(self):
+    def cleanup(self) -> None:
         # Needs to be overridden in derived class
         pass
 
-    def openFiles(self):
+    def openFiles(self) -> None:
         # Needs to be overridden in derived class
         pass
 
-    def closeFiles(self):
+    def closeFiles(self) -> None:
         # Needs to be overridden in derived class
         pass
 
     def createProperty(
         self,
-        propertyName,
-        usesRegex,
-        defaultValue,
-        deprecated,
-        propertyArrayName,
-    ):
+        propertyName: str,
+        usesRegex: bool,
+        defaultValue: str,
+        deprecated: bool,
+        propertyArrayName: str | None,
+    ) -> str:
+        # Needs to be overridden in derived class
+        raise NotImplementedError()
+
+    def writePropertyArray(self, propertyArray: PropertyArray) -> None:
         # Needs to be overridden in derived class
         pass
 
-    def writePropertyArray(self, propertyArray: PropertyArray):
-        # Needs to be overridden in derived class
-        pass
-
-    def writeProperties(self):
+    def writeProperties(self) -> None:
         self.openFiles()
 
         for propertyArray in self.propertyArrayDict.values():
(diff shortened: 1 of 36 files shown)


scripts/IceGridUtil.py
```diff
@@ -183,7 +184,15 @@ def shutdown(self, current: Driver.Current) -> None:
 
 
 class IceGridRegistry(ProcessFromBinDir, Server):
-    def __init__(self, name: str, portnum: int = 20, ready: str = "AdminSessionManager", *args: Any, **kargs: Any):
+    def __init__(
+        self,
+        name: str,
+        portnum: int = 20,
+        ready: str = "AdminSessionManager",
+        createDb: bool = True,
```
pepone, line 192: Make this keyword-only (`*args, createDb: bool = True`). It sits in front of `*args`, which forwards positionally to `Server.__init__` — so a 4th positional used to mean `waitForShutdown` and now means `createDb`. No caller does that today, but nothing warns if one starts.

config/makeprops.py
```diff
@@ -125,21 +128,21 @@ def parseProperty(self, attrs):
             propertyArrayName,
         )
 
-    def validateKnownAttributes(self, validAttrs, attrs):
+    def validateKnownAttributes(self, validAttrs: list[str], attrs: AttributesImpl) -> None:
         if "name" not in attrs:
-            print(sys.stderr, "missing name attribute")
+            print("missing name attribute", file=sys.stderr)
```
pepone, line 132: This warning can't be observed: `parseProperty` asserts on the same missing name at l.105, and the `section` branch does at l.195. Pick one — make it fatal here and drop both asserts, or keep warning and let the callers cope. Same area: the `class` branch at l.176 turns a missing name into the literal string `"None"`.

config/makeprops.py
```diff
@@ -584,55 +612,55 @@ def writePropertyArray(self, propertyArray):
...
             defaultValue=f'"{defaultValue}"',
             deprecated="true" if deprecated else "false",
-            propertyArray=f"PropertyNames.{propertyArray}Props" if propertyArray else "null",
+            propertyArray=f"PropertyNames.{propertyArrayName}Props" if propertyArrayName else "null",
         )
         return line
 
     @override
-    def moveFiles(self, location):
+    def moveFiles(self, location: str) -> None:
```
pepone, line 622: Pre-existing, but this path is dead: the JS file lives at `js/packages/ice/src/Ice/PropertyNames.js`, so `makeprops.py` exits 1 on every run and never regenerates it. C++/Java/C# get moved first, which is why the checked-in files still match. Filed separately rather than folding it in here.

cpp/test/DataStorm/relayReconnect/test.py
```diff
@@ -64,9 +68,18 @@ class RelayReconnectTestCase(ClientServerTestCase):
     # middleRelay: the relay that joins the two halves, started last.
     # writerRelay/readerRelay: the relays whose logs confirm the writer/reader announcement arrived before the middle
     # relay starts.
-    def __init__(self, nodes, middleRelay, writerRelay, readerRelay, reader, writer, **kwargs):
+    def __init__(
+        self,
+        nodes: Sequence[Node],
+        middleRelay: Process,
```
pepone, line 74: These are always `Node`s — elements of `chain`/`interior`. Typing them `Node` and annotating `self.logs: dict[Node, str] = {}` (l.100) closes the one spot left in this file inferred as `Unknown`.

pepone (approved):
Approving; comments are non-blocking.

Checked both claims in the description: regenerating `PropertyNames` for C++/Java/C# leaves the tree clean, and loading every `cpp`/`php` `test.py` through the framework gives an identical suite/test-case inventory to main.

One thing across the whole PR: a number of the new comments explain the type checker rather than the code (`# writeProperties calls openFiles before anything reaches here.` x4, `# Spelled out rather than concatenated: a list[str] is not a list[ExpectPattern].`). Fine one at a time, heavy at 35 files.

externl:
Trimmed in b20e2ed908f — 26 added comment lines down to 12. The ones that went were restating the assert on the next line or explaining why pyright wanted something.

What stays says something the code does not: that `start()` does not call `setup()`, the `getParentProps` delegation through the MRO, and why passlib is not a declared dependency.

Also dropped `include` from the config while I was in there. It had grown to name every top-level directory containing a Python file, so the whole repository is checked instead — same 316 files.

PR #6241, opened 2026-07-28
Prevent crash dump processing from hanging CI jobs

bernardnormier:
The "Process Linux crash dumps" step can hang until the 6-hour job-level timeout: the processing scripts run external debuggers with no timeout, and a wedged `gdb` (or `cdb`) silently stalls the job. This happened on a recent CI run where the step sat for hours after a test failure left a core dump behind.

Changes:

- `process_linux_crash_dumps.py`: add timeouts to the `coredumpctl list` (60s), `coredumpctl dump` (300s), and `gdb` (300s) invocations. A timed-out `gdb` leaves a partial `backtrace.txt`, and processing continues with the remaining dumps. Also pass `set debuginfod enabled off` to `gdb`: batch mode auto-answers "y" to the debuginfod prompt on Ubuntu 24.04, which can stall on per-library network downloads.
- `process_windows_crash_dumps.py`: add a 300s timeout to the `cdb` invocation, which downloads symbols from the Microsoft symbol server.
- `ci.yml`: add `timeout-minutes: 15` to both crash dump processing steps as a backstop, and pass `-v` so the scripts report per-dump progress instead of running silently.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -379,10 +379,12 @@ jobs:
 
       - name: Process Linux crash dumps
         if: runner.os == 'Linux' && always()
+        timeout-minutes: 15
         run: |
           python3 scripts/process_linux_crash_dumps.py \
               --workspace $GITHUB_WORKSPACE \
-              --reports LinuxDumpReports
+              --reports LinuxDumpReports \
+              -v
         shell: bash
 
       - name: Upload Linux crash dump reports
@@ -403,6 +405,7 @@ jobs:
 
       - name: Process Windows crash dumps
         if: runner.os == 'Windows' && always()
+        timeout-minutes: 15
         shell: pwsh
         run: |
           $dumpDir   = Join-Path $env:GITHUB_WORKSPACE 'LocalDumps'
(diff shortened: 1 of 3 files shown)


externl (approved):

pepone (approved):
LGTM.

The interaction that matters works: the upload steps are `if: ... && always()`, so when the 15-minute backstop kills the processing step, the partial reports (extracted core, truncated `backtrace.txt`) still get uploaded. `-iex "set debuginfod enabled off"` is the right form too, since `-iex` runs before the inferior is loaded, which is what's needed to suppress the download on the main executable. Both call sites in `ci.yml` are covered, and there are no others.

A few follow-up ideas, none of them blocking:

- On Windows, `.symfix` is what points cdb at the Microsoft symbol server, so the 300s we're now bounding is mostly symbol download. `parse_image_paths` only reads `Image path:` lines from `lmv`, and those come from the dump's module list rather than from PDBs. Worth checking whether dropping `.symfix; .reload` produces the same `modules.txt` with no network round trip; that would remove the hang rather than bound it, the way the debuginfod change does on Linux.
- Worst case is 300s extract + 300s gdb per dump, so two slow dumps already exceed the 15-minute step budget. With debuginfod off, gdb on a local core rarely needs more than a few seconds, so 60-120s would make the per-dump timeouts the effective bound instead of the step timeout.
- Setting `DEBUGINFOD_URLS=""` in gdb's environment would also cover gdb builds that don't know the `set debuginfod` command.
- After a `coredumpctl list` timeout we return `[]`, and `main` then logs "No core dumps found" right after an ERROR line saying the opposite.

bernardnormier:
Thanks — all four ideas are tracked in #6268 so they survive the merge; the `.symfix` one needs a Windows CI experiment to confirm `modules.txt` is PDB-independent (and to weigh losing system-DLL symbolization in the backtraces), so none of them holds this PR.

PR #1694, opened 2024-01-19
Upgrade PHP extension to the new C++ mapping

pepone:
This PR upgrades the PHP extension to use the new C++ mapping.

PR #6242, opened 2026-07-28
Fix two DEBUG macro checks that never fired

externl:
Neither build system defines a bare `DEBUG`. Make adds `-DNDEBUG` for optimized builds and only `-g` otherwise, and MSBuild defines `_DEBUG` for Debug and `NDEBUG` for Release. These were the only two bare-`DEBUG` checks in the tree, and both were dead on every platform and configuration.

- `php/src/Operation.cpp`: the out-parameter `Z_ISREF` checks in `prepareRequest` were gated on `DEBUG`, so they had never run. Switched to `NDEBUG`, which is what governs `assert()` and what `Operation.cpp:778`, `Types.cpp`, `DefaultSliceLoader.cpp` and `Config.h` already use.
- `cpp/src/Ice/ResourceConfig.h`: `VER_DEBUG` was gated on `DEBUG`, so it was always `0` and the 31 `.rc` files using `FILEFLAGS VER_DEBUG` never set `VS_FF_DEBUG` — Windows debug binaries were not marked as debug builds. Switched to `_DEBUG`, matching `ICE_LIBNAME` six lines above and what MSBuild passes to `rc.exe`.

Found while auditing the codebase for dead code.

### DIFF
--- a/cpp/src/Ice/ResourceConfig.h
+++ b/cpp/src/Ice/ResourceConfig.h
@@ -18,7 +18,7 @@
 #    define ICE_LIBNAME(NAME) NAME ICE_SO_VERSION ""
 #endif
 
-#ifndef DEBUG
+#ifndef _DEBUG
 #    define VER_DEBUG 0
 #else
 #    define VER_DEBUG VS_FF_DEBUG
--- a/php/src/Operation.cpp
+++ b/php/src/Operation.cpp
@@ -376,7 +376,7 @@ IcePHP::TypedInvocation::prepareRequest(
     }
 
     // The operation's configuration (zend_function) forces out parameters to be passed by reference.
-#ifdef DEBUG
+#ifndef NDEBUG
     for (int i = static_cast<int>(_op->inParams.size()); i < _op->numParams; ++i)
     {
         assert(Z_ISREF(args[i]));


bernardnormier (approved):

pepone (approved):
Both checks verified.

**`ResourceConfig.h`** — `cpp/msbuild/ice.cpp.props` defines `_DEBUG` on the `ResourceCompile` item (not just `ClCompile`) for `Configuration=Debug`, and sets nothing for rc in Release, so `VER_DEBUG` resolves correctly both ways. 30 of the 31 `.rc` files using `FILEFLAGS VER_DEBUG` build through projects importing that props file.

**`Operation.cpp`** — worth noting this assert has never executed since it was added in #1694, and PHP is set up for every `ci.yml` config including the two `OPTIMIZE=no` debug jobs, so it goes live on merge. I checked it holds rather than assuming: swapped the assert for a non-aborting probe and ran the PHP suite — 358 out-parameter checks across `Ice/operations`, `Ice/optional`, `Ice/objects` and `Ice/scope`, all `Z_ISREF`, and all 10 suites green.

The index range is right too, which isn't obvious from the diff: `convertParams` numbers `ParamInfo::pos` per-list so out-params restart at 0, but `OperationI::function()` lays out `arg_info` as in-params then out-params, so `[inParams.size(), numParams)` are the by-ref slots. The preceding `argc` check keeps the indexing in bounds.

One follow-up, pre-existing and not for this PR: `python/modules/IcePy/msbuild/icepy.vcxproj` is the one outlier — it imports `python/msbuild/ice.props`, and neither props file sets ResourceCompile defines, so `IcePy.rc` keeps `VER_DEBUG = 0` in Debug. Same gap already stops its own `#ifdef _DEBUG` from renaming the pyd to `IcePy_d`.

PR #1944, opened 2024-03-15
Replace Int64 by std::int64_t

pepone:

PR #6246, opened 2026-07-28
Support a trailing required-argument option in a short option cluster

### LINKED ISSUE #6121: C++ `Options`: a trailing required-argument option in a mixed short-option cluster silently loses its argument
> [!NOTE]

## Summary

`IceInternal::Options::parse` uses one `argDone` flag per argv entry. A no-argument option character in a cluster sets it; a **required-argument** character at the *end* of the same cluster leaves it unchanged, so the separated-argument handler is skipped and the option's argument is silently dropped (and consumed as a positional/input file).

**File:** `cpp/src/Ice/Options.cpp:598-663`. Traced `-dI incdir` (`d`=NoArg, `I`=NeedArg): `d` sets `argDone=true`; `I` is the last cluster char so the `if (p != size-1)` attached-value branch is false and nothing runs; the `if (!argDone)` separated-arg block is then skipped → `I` unset, `incdir` misclassified as an input file. `-d -I incdir` and `-dIincdir` both work. `slice2cpp -dI incdir File.ice` reproduces (Codex: `incdir` treated as input file).

This is the shared parser for every Slice compiler and the IceGrid/IceStorm/IceBox admin tools; it is also compiled into `slice2py`'s `PythonUtil.cpp` and `slice2rb`'s `Ruby.cpp`, so the misparse reaches the Python/Ruby in-process Slice loaders.

## Fix

Track "the last clustered option still needs an argument" independently of "some option was processed" (don't reuse a single `argDone`): when a trailing clustered `NeedArg` option has no attached text, leave the loop and let the separated-argument path consume the next argv element. (Latent sibling: short `OptArg` options are also mishandled in clusters — no Slice tool registers one today.)

**Severity:** Low-Medium (silent, no diagnostic; standard getopt semantics violated across a widely shared parser).

bernardnormier:
## Summary

`IceInternal::Options::parse` — shared by the Slice compilers and the Ice command-line tools (icegridadmin, icegridregistry, icebox, ...) — mishandled a short option cluster ending with an option that requires an argument: `slice2cpp -dI dir file.ice` silently ignored `-I` and parsed `dir` as an extra Slice file. The cause is the per-argument `argDone` flag, which conflates "an option in this cluster was processed" with "this argument needs no separated option argument": the no-argument `-d` set it, which skipped the separated-argument handling for the trailing `-I`. Both `-d -I dir` and `-dIdir` already worked.

The parser now resets `argDone` when the last character of a cluster is a required-argument option, giving the usual getopt semantics: `-dI dir` is equivalent to `-d -I dir`. As a corollary, a cluster with a missing argument (`slice2cpp -dI`) now fails with the usual "option requires an argument" error instead of silently ignoring the option.

The new `cpp/test/IceUtil/inputUtil` test (named after the C# and Java tests of the same name) covers the short option cluster parsing. This parser is C++-only: the C# and Java `Options` classes only provide `split`.

Fixes #6121.


pepone (approved):
Approve.

Verified independently: linked the PR's `Client.cpp` against `libIce.a` with each version of
`Options.cpp` — red on main at `Client.cpp:42`, green here. Also ran a differential fuzz of the two
parsers over ~8k argv combinations: every differing case involves a cluster ending in a `NeedArg`
option, nothing else moves, and no command line that previously parsed *correctly* changes meaning.

Nice side effect worth knowing about: it also removes a spurious `` `-o:' option cannot be repeated ``
— pre-fix a trailing `NoRepeat` `NeedArg` option was added to `seenNonRepeatableOpts` without ever
being set, so `-do -o out.cpp` failed as a repeat.

Two non-blocking suggestions:

1. Changelog section — this is cross-mapping, not C++. `slice2java -dI dir` changed too, and
   `Ice.loadSlice` registers the same `d` + `D`/`U`/`I` shape in Python
   (`python/modules/IcePy/Slice.cpp:46-50`) and Ruby. `changelog.d/cpp/` renders as
   `### C++ Changes`, which those users won't read. Precedent for `general/`: the `ice2slice`
   removal and the MCPP update are both under General Changes. Suggest
   `changelog.d/general/6121.md`.

2. `Options::split()` is still untested in C++ — this file is the natural home, as a follow-up. The
   old `cpp/test/IceUtil/inputUtil/Client.cpp` had a `testing string to command line arguments... `
   section covering it; it went away as collateral in a5d54806db (#1944) along with the `Int64`
   tests that gave the directory its name. The C#/Java `inputUtil` tests are the surviving ports of
   that section, and `split` is still live in `PluginManagerI.cpp:272` and
   `IceBox/ServiceManagerI.cpp:31`.

When closing #6121: its "latent sibling: short `OptArg` options are also mishandled" note doesn't
apply — `ArgType` only has `NeedArg` and `NoArg`. Nothing left latent.

bernardnormier:
Thanks — the differential fuzz over the two parsers is a great way to bound the blast radius, and good spot on the spurious `-o:' option cannot be repeated` failure; I hadn't realized the pre-fix code also poisoned `seenNonRepeatableOpts` for the never-set option.

Both suggestions taken in the latest commit:

- The fragment now lives under `changelog.d/general/` — you're right that `### C++ Changes` is the wrong shelf for `slice2java` and `Ice.loadSlice` users.
- The `Options::split` section is back in `inputUtil/Client.cpp`: the pre-#1944 "testing string to command line arguments" tests, updated to the current API, plus the trailing-`\c`-in-unterminated-quote case that #6201 added to the C# and Java twins (C++ was already correct on that path; now it's pinned). Runs green locally.

And noted for the #6121 close: no latent `OptArg` sibling, since `ArgType` only has `NeedArg` and `NoArg`.

PR #6250, opened 2026-07-28
Fix wrong-category eviction in the Java Logger admin facet

### LINKED ISSUE #6108: Java `LoggerAdminI` stores absolute list indexes into one shared queue → wrong-category eviction / retention violation
> [!NOTE]

## Summary

`LoggerAdminI` keeps log and trace messages in **one** shared `LinkedList` and tracks the oldest of each category by **absolute integer position** (`_oldestLog`, `_oldestTrace`). Removing an older message of one category shifts the other category's element down by one without updating its stored index, so the stale index later evicts the wrong category. C++ uses stable `list::iterator` and C# uses `LinkedListNode`; the Java port mistranslated those stable references into absolute indexes.

**File:** `java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerAdminI.java:167-212`
```java
_queue.remove(_oldestLog);   // positional List.remove(int) — shifts every later element, incl. traces
// _oldestTrace not adjusted
```
Concrete sequence (KeepLogs=1, KeepTraces=1): enqueue log, trace, log, trace → removing the first log shifts the trace from index 1 to 0, and the stale `_oldestTrace = 1` then evicts the new **log** instead — cascading wrong-category eviction and cap violation. Both asserts (`:186, :204`) can fire under `-ea`; with asserts off it silently corrupts what `getLog`/`attachRemoteLogger` return.

## Fix

Use stable nodes/iterators (as C++/C# do), or separate per-category queues, or update both indexes after every removal.

**Severity:** Medium-Low (wrong diagnostic retention; no crash with asserts off).

bernardnormier:
## Summary

The Java `LoggerAdminI` tracked the oldest log and the oldest trace message by absolute position (`_oldestLog`, `_oldestTrace`) in the single queue shared by both categories. Removing a message shifts the positions of all later messages, so the other category's stale index could evict the wrong message. With the caps set to 1, the sequence print, trace, print, trace left the queue with two trace messages and no log message at all. The C++ and C# implementations use stable `list` iterators / `LinkedListNode` references and don't have this bug; the Java port mistranslated them into indexes.

The fix removes the cached positions entirely: when a category is at its cap, the queue is scanned from the head and the first message of that category is removed. With no cross-category state, wrong-category eviction is impossible by construction, and the scan is bounded by the queue's configured capacity.

The new eviction test runs in all three mappings; it fails against the previous Java code and passes against the (already correct) C++ and C# implementations.

Fixes #6108.

### DIFF
--- a/changelog.d/java/6108.md
+++ b/changelog.d/java/6108.md
@@ -1,0 +1,4 @@
+- Fixed the eviction of old log messages by the Logger admin facet. Once the facet had accumulated
+  `Ice.Admin.Logger.KeepLogs` log messages and `Ice.Admin.Logger.KeepTraces` trace messages, the arrival of a new
+  message could evict a message of the wrong category — for example, a new trace message could evict a warning
+  message. As a result, `getLog` and attached remote loggers could receive fewer or older messages than expected.
--- a/cpp/test/Ice/admin/AllTests.cpp
+++ b/cpp/test/Ice/admin/AllTests.cpp
@@ -533,6 +533,33 @@ allTests(Test::TestHelper* helper)
 
         com->destroy();
     }
+    {
+        // With KeepLogs=1 and KeepTraces=1, a new message evicts the previous message of the same category,
+        // regardless of the messages of the other category in between.
+        PropertyDict props;
+        props["Ice.Admin.Endpoints"] = "tcp -h " + defaultHost;
+        props["Ice.Admin.InstanceName"] = "Test";
+        props["NullLogger"] = "1";
+        props["Ice.Admin.Logger.KeepLogs"] = "1";
+        props["Ice.Admin.Logger.KeepTraces"] = "1";
+        optional<RemoteCommunicatorPrx> com = factory->createCommunicator(props);
+
+        com->print("print1");
+        com->trace("testCat", "trace1");
+        com->print("print2");
+        com->trace("testCat", "trace2");
+
+        auto logger = com->getAdmin()->ice_facet<LoggerAdminPrx>("Logger");
+        string prefix;
+
+        LogMessageSeq logMessages = logger->getLog(LogMessageTypeSeq(), StringSeq(), -1, prefix);
+        test(logMessages.size() == 2);
+        auto p = logMessages.begin();
+        test(p++->message == "print2");
+        test(p->message == "trace2");
+
+        com->destroy();
+    }
     cout << "ok" << endl;
 
     cout << "testing custom facet... " << flush;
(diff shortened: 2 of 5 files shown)


pepone (approved):
LGTM.

I checked the equivalence claim rather than taking it on faith: transcribed the C++/C# stable-iterator / `LinkedListNode` algorithm into a model and fuzzed it against the new head-scan — 300k random sequences, caps 0–5, up to 40 messages, compared by element identity — 0 divergences. The scan finds the same element by construction, since an eviction only ever removes a message of its own category, so the stable iterator always denotes the first message of that category.

The test is genuinely red→green: pre-fix Java leaves `[trace1, trace2]`, the fix leaves `[print2, trace2]`, which is what C++/C# already produce. Same fuzz: the new code never exceeds the caps; the old code exceeded them in ~54% of 200k random small-cap sequences.

Two non-blocking notes:

1. The description says "both asserts in `log()` can fire with `-ea`". I couldn't reproduce that — 2M random sequences (caps 1–8, up to 60 messages) against the old algorithm, and neither assert fired, nor did `List.remove(int)` ever go out of bounds. The claim comes from #6108, where it was also stated without a scenario. Suggest dropping the clause; the silent wrong-category eviction the changelog fragment describes is the real symptom and stands on its own.

2. Filed #6259 for something adjacent I noticed while reading `log()`: in C++ and Java the remote-logger fan-out loop sits inside the retention `if`, so `KeepLogs=0` / `KeepTraces=0` also stops live delivery to attached remote loggers; C# does the fan-out outside the guard and keeps delivering. Pre-existing, out of scope here.

Added `backport-3.7` — 3.7 has the same indexes in both `java/` and `java-compat/`.

The one red CI job is `IceGrid/replication` on windows-2022 debug failing in a `ConnectTimerTask` connect timeout, unrelated.

bernardnormier:
Thanks — fuzzing the transcribed C++/C# algorithm against the head-scan is a stronger equivalence argument than the by-construction one in the description, and the ~54% cap-exceedance number is a nice measure of how live the old bug was.

Dropped the "both asserts in `log()` can fire with `-ea`" clause from the description as you suggest — it was inherited from #6108 without a scenario, and your 2M-sequence run not firing either assert (nor an out-of-bounds `remove`) is good evidence it was an overclaim. The silent wrong-category eviction stands on its own.

Good catch on the fan-out-inside-the-retention-guard inconsistency too; #6259 is the right place for it. And thanks for the `backport-3.7` label — agreed the same indexes are in both `java/` and `java-compat/` there.

PR #6254, opened 2026-07-28
Catch Throwable from connection close callbacks in Java

### LINKED ISSUE #6101: Java: unchecked close-callback exceptions skip `StateFinished` and strand connection finalization (deterministic destroy hang)
> [!NOTE]

## Summary

Java catches only `LocalException` around the user close callback in `ConnectionI.finish()`. A `RuntimeException` or `Error` from the user-supplied `CloseCallback` escapes **before** `_closeCallback` is cleared, `StateFinished` is set, observers are detached, and the connection is removed from its factory. `waitUntilFinished()` then never returns, so adapter destruction and `Communicator.destroy()` hang deterministically. C++ catches `std::exception` and `...`; C# catches `System.Exception`; JS catches all thrown values — Java is the only mapping with a narrow catch. (Related to the C++-side #6033.)

**File:** `java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectionI.java:898-923`
```java
if (_closeCallback != null) {
    try {
        _closeCallback.closed(this);
    } catch (LocalException ex) {          // :901 — LocalException only
        _logger.error("connection callback exception:\n" + ex + '\n' + _desc);
    }
    _closeCallback = null;                 // :904 — skipped on a RuntimeException/Error
}
// ... setState(StateFinished) at :911, observer detach at :916, _removeFromFactory at :922 — all skipped
```
`finish()` runs from a `ThreadPool` work item whose only guard logs and continues, so the thread survives but `finish()` is never re-run; the connection stays `< StateFinished` forever and `waitUntilFinished()` (`:184-192`) blocks. An `Error` additionally kills the event-handler thread.

## Fix

Catch `Throwable` around the callback (log it) and perform the state transition / observer detach / factory removal in a `finally`, mirroring C++/C#/JS.

**Severity:** High (deterministic, unrecoverable hang of `Communicator.destroy()` from a user callback throwing a `RuntimeException` — a common bug class; Java-only divergence from the cross-language "log and continue" contract).

bernardnormier:
In the Java mapping, the connection close callback was invoked under `catch (LocalException)`, so a `RuntimeException` (or `Error`) thrown by a user-supplied `CloseCallback` escaped `ConnectionI.finish` before `StateFinished` was set and before the connection was removed from its factory. The connection then never finished, and `Communicator.destroy()` hung forever in `waitUntilFinished`. I reproduced this hang with a standalone program: the exception propagates from the callback through `ConnectionI.finish` into the thread pool, which logs it and carries on, leaving the connection stranded.

The fix widens the guard to `catch (Throwable)` at both call sites — `finish()` and the immediate dispatch in `setCloseCallback` when the connection is already closed — matching the convention used elsewhere in the Java runtime for user callbacks, and the behavior of the other mappings (C++ catches `...`, C# catches `System.Exception`, JS catches everything).

A new case in the `Ice/ami` test sets a throwing close callback on an active connection and closes it. Without the fix, the test suite hangs at communicator destruction; with the fix, the exception is logged and the connection finishes normally. Note the `ice_ping` in the test is load-bearing: after the previous case closes the connection, `ice_getConnection` returns the same closed connection, and a callback set on it takes the immediate-dispatch path instead of the `finish()` path.

Close callbacks are not supposed to throw, so this is a robustness/consistency fix rather than the High-severity issue reported; there is no CHANGELOG entry.

Fixes #6101.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectionI.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectionI.java
@@ -301,7 +301,7 @@ public final class ConnectionI extends EventHandler implements Connection, Cance
                         public void run() {
                             try {
                                 callback.closed(ConnectionI.this);
-                            } catch (LocalException ex) {
+                            } catch (Throwable ex) {
                                 _logger.error("connection callback exception:\n" + ex + '\n' + _desc);
                             }
                         }
@@ -898,7 +898,7 @@ public final class ConnectionI extends EventHandler implements Connection, Cance
         if (_closeCallback != null) {
             try {
                 _closeCallback.closed(this);
-            } catch (LocalException ex) {
+            } catch (Throwable ex) {
                 _logger.error("connection callback exception:\n" + ex + '\n' + _desc);
             }
             _closeCallback = null;
--- a/java/test/src/main/java/test/Ice/ami/AllTests.java
+++ b/java/test/src/main/java/test/Ice/ami/AllTests.java
@@ -955,6 +955,19 @@ public class AllTests {
                 cb.check();
             }
             {
+                // Setting a close callback on an already-closed connection invokes the callback
+                // immediately; an Error thrown by the callback is logged, not propagated.
+                p.ice_ping(); // Establishes a new working connection: the previous one is closed.
+                Connection con = p.ice_getConnection();
+                con.close();
+                Callback cb = new Callback();
+                con.setCloseCallback(c -> {
+                    cb.called();
+                    throw new AssertionError("close callback");
+                });
+                cb.check();
+            }
+            {
                 //
                 // Remote case.
                 //
--- a/java/test/src/main/java/test/Ice/ami/AllTests.java
+++ b/java/test/src/main/java/test/Ice/ami/AllTests.java
@@ -941,6 +941,20 @@ public class AllTests {
                 cb.check();
             }
             {
+                // A close callback that throws: the exception is logged and the connection still
+                // finishes (a stranded connection would hang Communicator.destroy at the end of the
+                // test).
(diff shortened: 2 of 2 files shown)


java/test/src/main/java/test/Ice/ami/AllTests.java
```diff
@@ -940,6 +940,20 @@ public static void allTests(TestHelper helper, boolean collocated) {
                 }
                 cb.check();
             }
+            {
+                // A close callback that throws: the exception is logged and the connection still
+                // finishes (a stranded connection would hang Communicator.destroy at the end of the
+                // test).
+                p.ice_ping(); // Establishes a new working connection: the previous one is closed.
+                Connection con = p.ice_getConnection();
```
pepone, line 948: I guess you would get rid of the `ice_ping` after you merge #6256

bernardnormier: Yes

pepone (approved):

InsertCreativityHere (approved):
Looks good to me!

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectionI.java
```diff
@@ -301,7 +301,7 @@ public synchronized void setCloseCallback(final CloseCallback callback) {
                         public void run() {
                             try {
                                 callback.closed(ConnectionI.this);
-                            } catch (LocalException ex) {
+                            } catch (Throwable ex) {
```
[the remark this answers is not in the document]
bernardnormier: Added in 8c05cccf8b: the new case closes the connection first — `close()` blocks until the connection reaches the closed state, so `setCloseCallback` deterministically takes the immediate-invocation branch — and the callback throws an `AssertionError` to pin the `Throwable`-vs-`Exception` distinction. `Ice/ami` passes.

PR #6266, opened 2026-07-29
Move minWireSize from the enum type to EnumHelper

externl:
`minWireSize` was defined on the enum type rather than on `EnumHelper`. slice2js emits `<Type>._helper` in every helper position, so the stream helpers never saw it.

- `SequenceHelper.read` passes it to `readAndCheckSeqSize`, so for `sequence<SomeEnum>` the size check computed `sz * undefined === NaN` and never fired.
- The NaN then persisted: the reset condition is also false against NaN, so the guard stayed disabled for every later sequence on the same `InputStream`.
- Decoding was never affected. `minSize` feeds only the bounds check, not the element count read from the wire, so well-formed sequences read identically before and after; malformed ones now fail up front instead of on buffer underrun.
- Moved the property to `EnumHelper`, where the five readers in `StreamHelpers.js` actually look.
- Dictionaries and the fixed-length branches are unaffected; neither reaches an enum's `minWireSize`.

### DIFF
--- a/js/packages/ice/src/Ice/EnumBase.js
+++ b/js/packages/ice/src/Ice/EnumBase.js
@@ -46,6 +46,12 @@ class EnumHelper {
     readOptional(is, tag) {
         return this._enumType._readOpt(is, tag);
     }
+
+    // SequenceHelper.read consults this to sanity-check a sequence's declared element count against the
+    // remaining buffer. An enum marshals as a size, which occupies at least one byte.
+    get minWireSize() {
+        return 1;
+    }
 }
 
 export function defineEnum(enumerators) {
@@ -67,8 +73,6 @@ export function defineEnum(enumerators) {
             maxValue = value;
         }
     }
-
-    Object.defineProperty(type, "minWireSize", { get: () => 1 });
 
     type._write = function (os, v) {
         os.writeEnum(v);


pepone (approved):
Verified. Approving.

`EnumHelper` is the only object that ever lands in a helper position for an enum —
`getHelper` has emitted `<Enum>._helper` since the original JS mapping (`.__helper`
until 006bdff840), so the type-level `minWireSize` was dead from day one. Of the five
readers in `StreamHelpers.js`, only `SequenceHelper.read` was reachable: the fixed-seq
and fixed-dict branches are unreachable for enums because `Enum::isVariableLength()`
is true.

Confirmed locally, before → after:

    MyEnum._helper.minWireSize    undefined  →  1
    _minSeqSize after a seq read  NaN        →  3
    guard on the next sequence    no-op      →  MarshalException

and a well-formed `sequence<MyEnum>` decodes identically either way.

Two notes:

- "every later sequence on the same `InputStream`" is bounded — `clear()` resets
  `_startSeq` to -1, which un-poisons the guard, and `ConnectionI` clears the read
  stream between messages. So the window is one message, not the connection.
- No test needed. Malformed input threw `MarshalException` before this too, just later
  on buffer underrun, so a regression test would have to assert on `_minSeqSize`.

3.7 has the same bug — same `EnumHelper`, same `._helper` codegen — and the fix ports
as-is, so I'm tagging `backport-3.7`.

The red `android-bt` job fails in *Setup Bluetooth*; unrelated.

PR #6267, opened 2026-07-29
Remove dead code from the Slice compilers

externl:
Dead code in the Slice front-end and back-ends, none of it reachable from any of the nine translators.

- `FileTracker::error()` and its `_source` field, and `Preprocessor::getFileName()` and its `_shortFileName` field — the two methods have no callers, and both fields are written and never read.
- `ImportScope` in `slice2py/PythonUtil.h` — never used; the code tracks import scope with the `_runtimeImports`/`_typingImports` maps instead.
- The `TypeList` and `StringPairList` aliases, each with a single occurrence. `StringPairList` was also at global scope, so it leaked into every translation unit including `SwiftUtil.h`.
- Five `break` statements after `return` in `slice2js`'s `getValue`, which the `KindString` case above them already omits.
- `splitScopedName`'s `allowEmpty` parameter — all three call sites take the default, making the `else if` an unconditional `else`.
- `writeDependencies`' declaration in `Util.h` — all three calls are inside `SliceUtil.cpp`, so it moves to an anonymous namespace.

Generated output is unchanged: running all nine translators over a Slice file exercising enums, sequences, dictionaries, structs, classes and interfaces produces byte-identical output before and after, across all 83 files.


cpp/src/Slice/Preprocessor.cpp
```diff
@@ -110,19 +110,12 @@ Slice::Preprocessor::create(const string& path, const string& fileName, const ve
 Slice::Preprocessor::Preprocessor(string path, const string& fileName, const vector<string>& args)
     : _path(std::move(path)),
       _fileName(fullPath(fileName)),
-      _shortFileName(fileName),
```
InsertCreativityHere, line 113: Indeed dead code.
Use to be used by the old `printMakefileDependencies` function which was removed in #4259.

cpp/src/slice2py/PythonUtil.h
```diff
@@ -21,16 +21,6 @@ namespace Slice::Python
         Dispatch
     };
 
-    /// Represents the scope of an import statement—either required at runtime or only used for type hints.
```
InsertCreativityHere, line 24: Fun fact, this was always dead code.
It was added by the big python code-gen refactoring PR #4213,
but even in that PR it was already dead.

cpp/src/Slice/Preprocessor.cpp
```diff
@@ -110,19 +110,12 @@ Slice::Preprocessor::create(const string& path, const string& fileName, const ve
...
       _args(args)
 {
 }
 
 Slice::Preprocessor::~Preprocessor() { close(); }
 
-string
-Slice::Preprocessor::getFileName()
```
InsertCreativityHere, line 121: Was only ever called by `slice2py` as a hacky way to support the `pkgdir` metadata.
These 2 callers were also removed in the python refactoring of #4213.

cpp/src/Slice/SliceUtil.cpp
```diff
@@ -373,24 +373,26 @@ Slice::printGeneratedHeader(IceInternal::Output& out, string_view path, string_v
     out << nl << commentStyle << " <auto-generated>Generated from Slice file '" << path << "'.</auto-generated>";
 }
 
-void
-Slice::writeDependencies(const string& dependencies, const string& dependFile)
+namespace
```
InsertCreativityHere: We place anonymous namespaces like this at the top of files for visibility, instead of having them randomly woven through the source. This looks fine, but should be moved to the top of this file.

cpp/src/Slice/SliceUtil.cpp
```diff
@@ -423,7 +425,7 @@ Slice::argvToArgs(int argc, char* argv[])
 // Split a scoped name into its components and return the components as a list of (unscoped) identifiers.
 //
 vector<string>
-Slice::splitScopedName(const string& scoped, bool allowEmpty)
```
InsertCreativityHere, line 426: Alot of this function is actually transitively dead.

`allowEmpty` existed to handle trailing `::` in the provided identifiers. The reason it's a dead parameter is because we _never_ pass identifiers with trailing `::` anymore. Stating this with an assert removes alot of unnecessary checks from this function. This is the simplest Claude got it down to:
```
Slice::splitScopedName(const string& scoped)
{
    assert(scoped.compare(0, 2, "::") == 0); // must be a fully-scoped identifier
    assert(scoped.back() != ':'); // must end with an identifier (not just "::")

    vector<string> ids;
    string::size_type next = 2; // skip the leading "::"
    string::size_type pos;
    while ((pos = scoped.find("::", next)) != string::npos)
    {
        ids.push_back(scoped.substr(next, pos - next));
        next = pos + 2;
    }
    ids.push_back(scoped.substr(next));
    return ids;
}
```

cpp/src/Slice/SliceUtil.cpp
InsertCreativityHere: This was checking for a trailing `::`, which never happens anymore. So it is always true.

InsertCreativityHere: Same here, always true now.

InsertCreativityHere (approved):
Looks good to me, just some small things.

pepone (approved):
LGTM.

Verified no behavior change beyond the single-file check in the description: built both sides and ran all nine translators over every `.ice` file in the repo — 509 files, 4612 invocations, 30692 output files, all byte-identical (exit codes and stderr included) once the embedded worktree path is normalized. That sweep also covers `--depend`/`--depend-xml`/`--depend-json` and `--list-generated` for both consumers, plus the malformed-input error/cleanup paths.

Two notes on the reasoning, not the code:

- `FileTracker::error()` wasn't merely uncalled, it was unreachable: both `--list-generated` consumers — `slice2java/Main.cpp:249` and `slice2matlab/Main.cpp:290`, worth noting it's two, not just Java — `cleanup()` and return on failure before `dumpxml()` runs.
- `splitScopedName`'s surviving `else { ids.emplace_back(""); }` is dead as well. I instrumented it and got 0 hits across all 509 files: neither `scoped()` nor `mappedScoped("::", true)` ever ends in `::`. Leaving it is the right call for a no-behavior-change PR, just flagging it.

Three non-blocking suggestions:

1. `Preprocessor.cpp:123` — `getBaseName()` declares an unused `string suffix;` (copy-paste from `checkInputFile()`, where it is used). Same file, same theme; no compiler flags it because `std::string` isn't trivially destructible.

   ```diff
    Slice::Preprocessor::getBaseName()
    {
        string base(_fileName);
   -    string suffix;
        string::size_type pos = base.rfind('.');
   ```

2. The new anonymous namespace in `SliceUtil.cpp` sits ~300 lines above its only three callers. Moving it just above `DependencyGenerator::writeMakefileDependencies` would keep it next to what uses it.

3. Out of scope but adjacent: `python/modules/IcePy/Slice.cpp:6` includes `Slice/Preprocessor.h` without using the class.

No changelog entry needed — internal only, no user-visible impact.

externl:
All three applied: anonymous namespace moved to the top of the file, `splitScopedName` simplified as suggested, and the unused `suffix` local dropped.

The rewrite also removes the `else { ids.emplace_back(""); }` you both flagged as transitively dead. Ran `slice2swift` — the only caller — over all 508 `.ice` files in the repo with asserts enabled: no failures, and generated output is unchanged.

Left `python/modules/IcePy/Slice.cpp` alone as out of scope; happy to take it in a follow-up.

PR #4499, opened 2025-10-02
Do not restart acceptor on accept error

pepone:
The C++ side for #4498

Updated to include C#/Java as this is a small change.

PR #3850, opened 2025-04-14
Remove ThreadHookPlugin

bernardnormier:

PR #6269, opened 2026-07-29
Remove some dead code in C++ 

externl:
Dead code in the core C++ runtime.

- `IceInternal::removeWhitespace` — its only callers were `slice2freeze` and `slice2freezej`, deleted in `99c3941638f` when Freeze moved to its own repository.
- `Instance::setThreadHook`, orphaned by "Remove ThreadHookPlugin" (#3850).
- Four never-read fields: `Instance::_implicitContext` (the live path is `_implicitContextKind` plus `_sharedImplicitContext`) and `Instance::_metricsAdmin` with the forward declarations that existed only for it, `OpenSSLTransceiverI::_cipher`, and `IncomingConnectionFactory::_acceptorStopped`, write-only since #4499.
- `class TcpEndpoint;` and `class UdpEndpoint;` — forward declarations for classes that do not exist anywhere; the real ones are `TcpEndpointI` and `UdpEndpointI`.
- The `__osf__` disjunct in `Network.h`, for Tru64.
- `iAPEndpointIPtr` converted from `typedef` to `using`.

### DIFF
--- a/cpp/include/Ice/StringUtil.h
+++ b/cpp/include/Ice/StringUtil.h
@@ -104,11 +104,6 @@ namespace IceInternal
     ICE_API std::string toUpper(std::string_view);
     ICE_API bool isAlpha(char);
     ICE_API bool isDigit(char);
-
-    //
-    // Remove all whitespace from a string
-    //
-    ICE_API std::string removeWhitespace(std::string_view);
 }
 
 #endif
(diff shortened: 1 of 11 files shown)


bernardnormier (approved):
Looks good

pepone (approved):

PR #6273, opened 2026-07-29
Make android-bt setup survive transient device state

externl:
Two intermittent `android-bt` setup failures, found in a sweep of the last 77 CI runs (52 green, 23 cancelled, 2 failed). One on a PR, one on `main`; neither related to the change under test.

- Wait for btbond's `listening` marker instead of sleeping six seconds, then settle briefly. The client was connecting before the server registered its SDP record. A client-side SDP miss lands in `cleanup_rfc_slot(slot, BTSOCK_ERROR_SDP_DISCOVERY_FAILURE)` and closes the app fd, so Java's `readAll` returns -1 and `connect()` throws exactly `read failed, socket might closed or timeout, read ret: -1` — which reads like an I/O fault rather than a missing service. The settle is because the marker is not quite readiness: `listenUsingRfcommWithServiceRecord` returns once the stack hands back the channel number, before the record is added.
- Fail fast if the server reports a verdict. Otherwise a server that dies outright ("adapter not enabled") burns the whole wait and then starts a client with nothing to talk to.
- Retry `bluetoothAddress()`. On `main`, `settings get secure bluetooth_address` came back `cmd: Can't find service: settings` after the bond had already succeeded — the service is reached over binder and was briefly unregistered.
- Stop asking for the address three times. `bond()` already reads it and returns it, so `runBluetoothPrepare` uses that, and `bond()` calls `bluetoothAddress()` rather than duplicating the command.
- Keep the last 40 log lines, minus stack frames, from both devices. The old 500-character cap was less than one stack trace — which is why neither failure could be classified from CI output.


pepone (approved):
Approve. The diagnosis is right, and the mechanism is provable from AOSP rather than inferred.

A client-side SDP miss lands in `handle_discovery_comp`, which calls `cleanup_rfc_slot(slot, BTSOCK_ERROR_SDP_DISCOVERY_FAILURE)` (`btif_sock_rfc.cc:1143-1152`) and closes the app fd. Java's `readAll` then returns -1 and throws exactly `read failed, socket might closed or timeout, read ret: -1` (`BluetoothSocket.java:1397-1403`), reached from `connect()` via `readInt`/`waitSocketSignal`. So "client connects before the SDP record exists" produces precisely the observed message — worth stating in the description, since it currently calls the race inferred.

One gap: the `listening` marker doesn't imply the server is connectable. `listenUsingRfcommWithServiceRecord` returns as soon as `bindListen`'s `readInt` gets the channel (`BluetoothSocket.java:809`), which native writes in `send_app_scn` on `BTA_JV_GET_SCN_EVT` (`btif_sock_rfc.cc:1029`) — before it queues `BTA_JvCreateRecordByUser` (`do_in_main_thread`, `bta_jv_api.cc:200`). The record is only added on the later `BTA_JV_CREATE_RECORD_EVT` (`btif_sock_rfc.cc:1082`), then `BTA_JvRfcommStartServer`. After the marker appears, registration and the marker race — and this drops the 6s that used to cover it.

That window is a couple of main-thread turns, so slow activity launch remains the likelier cause and this does fix that. But a short settle is free insurance, and it also makes the stale-marker case harmless: if the peer's tolerant `logcat -c` fails and an old `listening` matches, you degrade to the old behavior instead of racing.

```python
for _ in range(30):
    peerLog = self._adbTolerantFor(peerAdb, "logcat -d -s BTBOND")
    # A server that failed outright is terminal. Take the last verdict rather than any match:
    # one log can hold both the marker and a later failure.
    verdicts = re.findall(r"RESULT server (\w+)", peerLog)
    if verdicts and verdicts[-1] != "OK":
        raise RuntimeError(f"btbond's server on '{peerDevice}' failed\n{peerLog[-500:]}")
    if "listening" in peerLog:
        # listenUsingRfcommWithServiceRecord returns once the stack hands back the channel number,
        # which it does before registering the SDP record (AOSP btif_sock_rfc.cc: send_app_scn on
        # BTA_JV_GET_SCN_EVT, create_server_sdp_record only on BTA_JV_CREATE_RECORD_EVT). Settle so
        # the marker is not read as readiness it does not yet imply.
        time.sleep(2)
        break
    time.sleep(1)
else:
    ...
```

Without the verdict check a server that fails fast ("adapter not enabled") burns the full 30s, warns, then starts a client with nothing to talk to. Note it makes a stale `RESULT server FAIL` fatal, so it leans on the same tolerant clear — another reason to keep the settle.

Minor, not blocking: the 30s fallback eats a quarter of the server's 120s `accept()` window (`MainActivity.java:186`), and `createBond` may take 60s of what's left — thinner margin than it looks.

externl:
Both applied in 3415e9b, with the AOSP mechanism written into the description — you are right that "inferred" undersold it once the path from `BTSOCK_ERROR_SDP_DISCOVERY_FAILURE` to that exact string is traceable.

The `send_app_scn`-before-`BTA_JvCreateRecordByUser` ordering is the part I had wrong: I took the marker as readiness when it only means the channel number came back, and then removed the sleep that had been covering the gap. Settle added, and the verdict check with it — a fast server failure burning the full wait and then launching a doomed client was a worse outcome than the bug I set out to fix.

Kept your last-verdict rule for the same reason it is used on the client side, so a stale `FAIL` followed by an `OK` is not read as this attempt's. Reused the frame-stripping tail for the raise rather than a character cap.

On the fallback eating the accept window: it now only runs when the server logs neither a marker nor a verdict, which is a hung activity rather than a failed one — so it should be rare enough to leave at 30s. Happy to trim it if you would rather have the margin.

Separately, pyright caught `reason` as possibly unbound in `bluetoothAddress` — it counts a zero-iteration `for`. Fixed; ruff had passed it, which is what I checked first.

externl:
Trimmed to 15s in the latest push. Worst case against the server's 120s `accept()` goes from 30s of headroom to 45s, and a healthy launch logs the marker in about a second, so it costs nothing in the normal path.

Issue #6117, opened 2026-07-23
Java low-severity batch: `Properties.load` reg-query pipe deadlock + `%VAR%` loops, TCP fd leaks, oversized invocation-timeout `Duration`


pepone:
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI static-review audit and verified by re-reading the source on `main` (`1579b6928e`). It was not benchmarked, reproduced at runtime, or exploited, so the severity is a static-review estimate. Please confirm before acting.

## Summary

Low-severity Java findings.

### 1. `Properties.load` reg-query pipe deadlock + `%VAR%` infinite loops
`java/src/com.zeroc.ice/.../Properties.java:458-503`: on Windows the config import runs `reg query` and calls `process.waitFor()` *before* draining stdout — a large key exceeds the OS pipe buffer and deadlocks startup. And `%VAR%` expansion (`:485-503`) loops forever on a self-reference (`FOO=%FOO%`) or a mutual reference (`FOO=%BAR%`, `BAR=%FOO%`). Fix: drain the stream concurrently with `waitFor`; bound the expansion / detect cycles.

### 2. TCP fd leaks on some failure paths
`TcpAcceptor.java:70-84` discards `_fd` without closing when `getAddressForServer` (DNS) throws; `Network.doAccept` (`Network.java:281-287`) and `createTcpSocket` (`:107-117`) throw on a post-open socket-option failure without closing the fd. (`doConnect` closes on every path; the acceptor's own option failures also close via helpers — those do not leak.)

### 3. Oversized invocation-timeout `Duration`
`ProxyOutgoingAsyncBase.java:177-188`: any `Duration` is accepted unvalidated; `toMillis()` throws `ArithmeticException` mid-invoke for durations beyond `Long.MAX_VALUE` ms (bypassing the `LocalException` catch), and `Reference.hashCode()` (`Reference.java:413`) calls `toMillis()` too, so an oversized-timeout proxy detonates in any hash-based collection. A positive sub-millisecond `Duration` truncates to 0 and times out immediately.

## Fix

Concurrent stream drain + cycle-bounded env expansion; close the fd on the leaking failure paths; validate/clamp the invocation-timeout `Duration` at `ice_invocationTimeout`.

**Severity:** Low (Windows/registry niche for #1; fd exhaustion over time for #2; pathological-but-accepted input for #3).

pepone:
[...]

PR #6285, opened 2026-07-29
Fix Java registry config loading and TCP fd leaks

bernardnormier:
Addresses items 1 and 2 of #6117. Item 3 is fixed by #6235.

## Item 1 — `Properties.load` with a Windows registry key

- The `reg query` output is now drained before `Process.waitFor`. The previous order deadlocked startup when the
  key held more configuration than the OS pipe buffer: `reg` blocked writing while `load` waited for it to exit.
  stderr is discarded at the OS level (`ProcessBuilder.Redirect.DISCARD`), so it cannot fill a pipe either. The
  rework also destroys the `reg` process when reading its output fails, and decodes that output using the
  platform's native encoding instead of byte-by-byte Latin-1.
- `%var%` references in a `REG_EXPAND_SZ` value are now expanded in a single pass: an expanded value is not
  rescanned. This fixes two bugs of the old rescan-from-the-start approach. First, a variable whose value
  contains `%var%` references (self- or mutual) looped the expansion forever. Second, a variable whose value
  contains a literal `%` silently corrupted the rest of the string: the introduced `%` paired with the next
  `%var%` reference, which was then swallowed as an empty-named variable — with `PCT=50%` and `A=alpha`, the
  input `%PCT%%A%` expanded to `50A%`; it now expands to `50%alpha`. `ExpandEnvironmentStringsW`, which the C++
  runtime uses for the same registry values, does not rescan either. An unknown variable still expands to the
  empty string, as before (this is pre-existing behavior that differs from `ExpandEnvironmentStringsW`, which
  leaves the reference in place).

## Item 2 — TCP file descriptor leaks on failure paths

- `Network.doAccept` now closes the accepted channel when setting its socket options fails. The existing catch
  translates the failure into a `SocketException` for the caller, but it left the accepted fd behind. Reachable
  when the peer resets the connection right after the handshake — what TCP health probes and port scanners
  produce — with the option call then failing on BSD-derived stacks; a listening server leaked one fd per
  occurrence.
- The `TcpAcceptor` constructor closes the server socket before discarding it when a later step throws — in
  practice `getAddressForServer` on a host that fails to resolve, or `setTcpBufSize` on a malformed
  `Ice.TCP.RcvSize` value, whose `PropertyException` is thrown before reaching the option helper that would
  have closed the socket. The `Network` option helpers close the socket themselves before throwing, and their
  close is idempotent, so the constructor's close is safe on every path.

The audit's third leak candidate, `Network.createTcpSocket`, is left unchanged: the socket is not connected yet
when its options are set, so no peer can invalidate it and the failure is not reachable in practice.

No changelog entry: the registry config path is a niche feature and the fixes are hardening; the fd leaks only
occur on failure paths. No tests: the registry path is Windows-only subprocess plumbing, and the fd leaks would
require injecting socket-option failures.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Network.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Network.java
@@ -283,6 +283,7 @@ public final class Network {
             socket.setTcpNoDelay(true);
             socket.setKeepAlive(true);
         } catch (IOException ex) {
+            closeSocketNoThrow(fd);
             throw new SocketException(ex);
         }
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
@@ -456,17 +456,20 @@ public final class Properties {
             && (file.startsWith("HKCU\\") || file.startsWith("HKLM\\"))) {
             try {
                 java.lang.Process process = Runtime.getRuntime().exec(new String[]{"reg", "query", file});
-                process.waitFor();
-                if (process.exitValue() != 0) {
-                    throw new InitializationException("Could not read Windows registry key '" + file + "'");
-                }
 
+                // Drain stdout before waiting for the process: reg blocks when its output fills the pipe buffer.
                 java.io.InputStream is = process.getInputStream();
                 StringWriter sw = new StringWriter();
                 int c;
                 while ((c = is.read()) != -1) {
                     sw.write(c);
                 }
+
+                process.waitFor();
+                if (process.exitValue() != 0) {
+                    throw new InitializationException("Could not read Windows registry key '" + file + "'");
+                }
+
                 String[] result = sw.toString().split("\n");
 
                 for (String line : result) {
@@ -482,27 +485,25 @@ public final class Properties {
                     if (pos != -1) {
                         String name = line.substring(0, pos).trim();
                         line = line.substring(pos + 13, line.length()).trim();
-                        while (true) {
-                            int start = line.indexOf('%', 0);
-                            int end = line.indexOf('%', start + 1);
 
-                            // If there isn't more %var% break the loop
+                        // Expand each %var% in a single pass; an expanded value is not re-scanned, so a variable
+                        // whose value contains %var% references cannot loop the expansion.
+                        StringBuilder expanded = new StringBuilder();
+                        int index = 0;
+                        while (true) {
+                            int start = line.indexOf('%', index);
+                            int end = start == -1 ? -1 : line.indexOf('%', start + 1);
                             if (start == -1 || end == -1) {
+                                expanded.append(line, index, line.length());
                                 break;
                             }
 
-                            String envKey = line.substring(start + 1, end);
-                            String envValue = System.getenv(envKey);
-                            if (envValue == null) {
-                                envValue = "";
-                            }
-
-                            envKey = "%" + envKey + "%";
-                            do {
-                                line = line.replace(envKey, envValue);
-                            } while (line.indexOf(envKey) != -1);
+                            String envValue = System.getenv(line.substring(start + 1, end));
+                            expanded.append(line, index, start);
+                            expanded.append(envValue == null ? "" : envValue);
+                            index = end + 1;
                         }
-                        setProperty(name, line);
+                        setProperty(name, expanded.toString());
                         continue;
                     }
                 }
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Properties.java
@@ -6,7 +6,7 @@ import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.PushbackInputStream;
-import java.io.StringWriter;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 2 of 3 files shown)


pepone (dismissed):
Approve with minor comments — all three fixes are correct. I traced each claim in the description against the source and ran the two `%var%` implementations side by side rather than eyeballing them; details below.

Only one thing I'd actually change before merge (the stderr drain); the rest is description wording and notes for later.

## Verification

- **`Network.doAccept`** — safe. The only caller is `TcpAcceptor.accept()`, which never receives the channel when `doAccept` throws, so there's no double-close or use-after-close. `SSL/AcceptorI` and `WSAcceptor` delegate to `TcpAcceptor` and inherit both fixes.
- **`TcpAcceptor` ctor** — the safety argument holds: `setBlock`, `setReuseAddress(ServerSocketChannel)` and `setRecvBufferSize`/`getRecvBufferSize(ServerSocketChannel)` all close before throwing, and `SelectableChannel.close()` is idempotent, so the added close is a no-op on those paths. If `createTcpServerSocket()` throws, `_fd` is still `null` and the guard skips it. The resulting shape matches both `UdpTransceiver` constructors.
- **Drain-before-`waitFor`** — correct fix for a real deadlock.
- **`%var%` expansion** — I ran both versions over 41 inputs. 6 differ, and every difference is either the intended fix or an improvement.

## The description undersells the change on two points

**1. `getAddressForServer` isn't the only leaking path in the constructor.** `setTcpBufSize(ServerSocketChannel, …)` reads `Ice.TCP.RcvSize` through `getPropertyAsIntWithDefault`, which throws `PropertyException` on a non-integer value *before* reaching `setRecvBufferSize` — the helper that would have closed the socket. So a malformed `Ice.TCP.RcvSize` leaked the listening socket too, and this PR fixes that as well.

**2. The expansion rewrite fixes a silent value-corruption bug, not just the infinite loop.** With `PCT=50%` and `A=alpha`, the input `%PCT%%A%` produced:

| | old | new |
|---|---|---|
| `%PCT%%A%` | `50A%` 💥 | `50%alpha` ✅ |

The old code re-scanned from index 0 after each substitution, so the `%` introduced by expanding `%PCT%` paired up with the following `%A%` and was swallowed as an empty-named variable. Any value whose expansion contains a literal `%` was silently mangled. Worth a line in the description — it's the one change here that could have quietly produced a wrong value rather than an obvious failure.

For completeness, the other differences are the intended ones — `%SELF%` and `%M1%`/`%M2%` no longer hang (they now yield the literal `%SELF%` / `%M2%`), and `%NEST%` is no longer expanded recursively. There's also one degenerate class: odd-length runs of `%`, e.g. `%%%A%%` gives `%A` before and `alpha%` after. Neither is "correct" for malformed input, and the new one at least doesn't drop the `%A%` expansion, so I'm not asking for a change — just noting it so it isn't a surprise later.

## Comments

**1. stderr is still undrained — the same deadlock class technically remains.** `reg` gets a separate stderr pipe that nobody reads. If it filled, `reg` would block on write, never exit, and the stdout loop would never see EOF — exactly the failure mode being fixed. In practice `reg`'s stderr is one short line, but nothing documents a bound on it, so I'd treat it as unproven rather than safe. One line closes it out:

```java
java.lang.Process process = new ProcessBuilder("reg", "query", file)
    .redirectError(ProcessBuilder.Redirect.DISCARD)
    .start();
```

`Runtime.exec` is a thin wrapper over `ProcessBuilder` anyway.

**2. The `ExpandEnvironmentStringsW` claim in the description is a bit strong.** The two sentences are in tension: `ExpandEnvironmentStringsW` leaves an *undefined* `%VAR%` in place, it doesn't blank it. So Java and C++ still diverge on undefined variables — Java blanking them is pre-existing behavior this PR correctly preserves as a non-goal. Worth narrowing to "does not rescan either", so nobody later reads it as "these paths are now equivalent".

## Nits / out of scope

- `TcpAcceptor.listen()` still does a bare `_fd = null`. It's safe — `Network.doBind` closes first — but it now sits a few methods above a constructor that closes defensively for the same reason.
- Pre-existing, in the block you're already rewriting: the process streams are never closed and the child is never destroyed on the throw path; and `sw.write(c)` on raw bytes is effectively a Latin-1 decode of `reg`'s console-codepage output, so non-ASCII registry values are mangled. (`Charset.defaultCharset()` isn't the fix — it's UTF-8 since JDK 18; `native.encoding` is closer.) `is.readAllBytes()` would also collapse the drain loop to one line.

## Follow-up (separate issue, not blocking this PR)

There's a related fd leak still open, adjacent to item 2 of #6117 but outside this diff:

`TcpAcceptor.accept()` hands the accepted channel to `StreamSocket`, whose constructor calls `init()` → `Network.setTcpBufSize(SocketChannel, …)`, which reads **`Ice.TCP.SndSize`** and throws `PropertyException` before touching the socket. `StreamSocket` then does `assert (!_fd.isOpen()); throw ex;` — it *asserts* the helper closed the fd rather than closing it itself — and `IncomingConnectionFactory` catches the `LocalException` with `transceiver` still `null`, so nothing closes it.

The acceptor starts up fine because the server-socket overload of `setTcpBufSize` only reads `Ice.TCP.RcvSize`. So `Ice.TCP.SndSize=abc` gives a listening-but-broken server that leaks one accepted fd per inbound connection. Low severity — it needs a malformed config value and the server rejects every connection anyway — but the `assert` is documenting exactly the invariant that path breaks.

## Tests

Agree with skipping them. `com.zeroc.ice` has no JUnit harness (only `tools/slice-tools` does) and the `TestHelper` suites can't reach package-private internals, so there's no cheap way to cover a pure string helper here. The regression risk was concentrated in the expansion edge cases, which is why I ran the differential check rather than relying on inspection — it came back clean.

bernardnormier:
Thanks for the differential check on the expansion — that's a stronger verification than the paths I traced by hand. All addressed in d3346af57b plus a description update:

- **stderr** — switched to `ProcessBuilder` with `Redirect.DISCARD`, as suggested.
- **Description** — now credits both undersold points: the `Ice.TCP.RcvSize`/`PropertyException` leak path in the `TcpAcceptor` constructor, and the silent `%`-swallowing corruption with your `%PCT%%A%` example. The `ExpandEnvironmentStringsW` claim is narrowed to the non-rescanning behavior, with an explicit note that blanking unknown variables is pre-existing Java behavior that differs from the Win32 API.
- **Nits** — took them all while in the block: the drain is now `readAllBytes` in try-with-resources, decoded with `native.encoding` (guaranteed present on our Java 17 floor), with `process.destroy()` in a finally for the throw path; and `listen()` now closes defensively like the constructor.
- **Follow-up** — filed as #6288. I checked the other mappings: C++ and C# share the gap (`init()` runs outside the guard in both; C++'s `~NativeInfo` doesn't close the fd, so it's a hard leak there too), JS doesn't read the buffer-size properties. Details in the issue.

pepone (approved):
Approve. The second commit takes every point from the last round. I re-verified the new block by running it rather than reading it.

## Verified

- **Drain before `waitFor`** — red/green on JDK 17: the pre-PR shape hangs on 200 KB of child stdout, the post-PR shape reads all of it.
- **`Redirect.DISCARD`** — 200 KB on stderr no longer blocks, with and without concurrent stdout. The old code would have hung there even with stdout drained, so this closed a second hole rather than just tidying one.
- **`destroy()` in the `finally`** — a no-op after a successful `waitFor`, and the non-zero-exit path still reaches the `InitializationException` branch.
- **`native.encoding` on our Java 17 floor** — correct. JEP 400 points at JDK 18, but the property landed in 17: non-null on both Temurin 17.0.19 and Oracle 17.0.12.
- **Android** — `Redirect.DISCARD` and `readAllBytes` are API 33 and minSdk is 34, and the branch is Windows-guarded regardless.
- **`listen()`** — equivalent-or-better. `doBind` already closes on its only throwing path, so the change is purely defensive; `close()` is idempotent and still nulls `_fd`.

## One nit — description wording, not code

> decodes that output using the platform's native encoding instead of byte-by-byte Latin-1, so non-ASCII registry values are no longer mangled

This is stronger than what the code delivers. `native.encoding` is the ANSI code page (`GetACP`), but console tools write *redirected* output in the console output code page (`GetConsoleOutputCP`, typically OEM), and `reg.exe` is specifically reported to do this. So non-ASCII can still decode wrong — and in a configuration where the console page happens to be Latin-1-compatible, the old byte-by-byte decode was the right one.

Practically a wash, since registry-backed property values are ASCII in any realistic config, so I wouldn't change the code. I'd just drop the "no longer mangled" claim. I could not test this on Windows.

PR #6287, opened 2026-07-30
Discard DataStorm session creation replies from superseded attempts

### LINKED ISSUE #6282: `Datastorm/reliability` Test Failure: Writer Resends its Entire Contained Queue
Observed once so far, on a `release, windows-2025` runner, on an unrelated PR:
https://github.com/zeroc-ice/ice/actions/runs/30468754312/job/90633372861

The crux of the failure is:
```
2026-07-29T18:42:34.1979977Z duplicate or rewound sample: 0 expected:100
2026-07-29T18:42:34.1980281Z failed!
2026-07-29T18:42:34.1980687Z D:\a\ice\ice\cpp\test\DataStorm\reliability\Reader.cpp:164: assertion `false' failed
```
Here is the code where the assertion tripped:
```
        // After the reconnect the reader must continue from 100: the writer resumes from the reader's last received
        // sample rather than re-sending its whole retained queue (which would re-deliver 0..99).
        for (int i = 0; i < 100; ++i)
        {
            auto sample = reader.getNextUnread();
            if (sample.getValue() != i + 100) // We get into this loop when we should not
            {
                cerr << "duplicate or rewound sample: " << sample.getValue() << " expected:" << (i + 100) << endl;
                test(false);
            }
        }
```

pepone:
Fixes #6282.

A session creation attempt that fails is retried with a bounded budget (`DataStorm.Node.RetryCount`, backing off from `DataStorm.Node.RetryDelay`). Replies were not tied to the attempt that produced them, so a late reply from an attempt that had already been superseded was charged against the attempt in progress: it cancelled and rescheduled that attempt's retry task and consumed one of its retries.

A relay between two nodes makes such replies routine. `AlreadyConnected` means the peer already has a session with this node; `NodeI::retrySubscriberSessionCreation` treated it as success only while this session was connected, so after a later close it was reclassified as a failure of the current attempt. In the reported trace a burst of them exhausted the six-retry budget in **2 ms** — none of the nominal 15.5 s of backoff elapsed — and destroyed a session that was reconnecting normally:

```
15.071  s/1: retrying connecting to 'writer-app:…' in 500 (ms),  retry 2/6
15.072  s/1: retrying connecting to 'writer-app:…' in 1000 (ms), retry 3/6
15.072  s/1: retrying connecting to 'writer-app:…' in 2000 (ms), retry 4/6
15.073  s/1: retrying connecting to 'writer-app:…' in 4000 (ms), retry 5/6
15.073  s/1: retrying connecting to 'writer-app:…' in 8000 (ms), retry 6/6
15.073  s/1: connection to 'writer-app:…' failed and the retry limit has been reached
15.073  s/1: destroyed session
15.076  s/3: created session (peer = 'writer-app:…')
```

The replacement session had no record of what the reader had received, so `getLastIds` reported nothing, the writer's `lastId` was 0, and it resumed from the start of its retained history. The reader received samples it had already consumed. #6039 was the same event taking its other branch: there the destroyed session was never re-created, so the reader hung instead.

## The fix

Each attempt now carries an identifier. `SessionI::sessionCreationFailed` ignores a reply whose identifier no longer matches the attempt in progress, so late replies can no longer cancel a retry task, consume a retry, or remove the session.

Two details matter more than the token itself:

**The decision is atomic.** Classifying the failure and deciding what to do with it happen together under the session mutex. They cannot be split: inspecting the session state can itself disconnect the session (`checkSession` detects an already-closed connection and disconnects), which changes the very state the decision depends on. A caller that checked separately could have its *own* failure discarded and be left with no attempt in flight, no retry scheduled, and no removal — reintroducing #6039's hang. `checkSession` is therefore split into a locked `checkSessionImpl`, and a disconnect it discovers still falls through to `retryImpl` in the same critical section.

**The identifier advances only where something else takes responsibility for recovery** — when the session connects, and when a failure is accounted for. Deliberately *not* on a bare disconnect: disconnecting schedules nothing, so invalidating the in-flight attempt there would discard the failure that would have scheduled a retry.

`NodeI`'s two retry helpers reduce to a single call, and no longer evaluate session state themselves.

## Testing

- `python3 allTests.py --filter=DataStorm` — 11/11, four runs after the final corrections (and four before them)

**No regression test.** The failing interleaving needs a late reply held across a disconnect, which survives only because it returns through a relay's own connections. Reproducing that deterministically would require either compiling DataStorm's internal `Contract.ice` into a test and implementing a controllable relay — the contract is unexported (`nm -gU libDataStorm.dylib | grep -c DataStormContract` → 0) and has churned recently, so it would rot — or adding a test-only seam to the product, for which there is no precedent in DataStorm. Neither seemed proportionate; I would rather say so than add a test that does not exercise the path. The new session trace line (`ignoring a session creation failure from attempt N, the current attempt is M`) makes the behaviour observable in a CI trace.

## Review

Cross-reviewed by OpenAI Codex (GPT-5.6) in two rounds, both initially rejecting the change.

The first round rejected composing the decision from separate `connectAttempt()` / `checkSession()` / `retry()` calls, for the stranded-session reason above — verified from source before reworking. The second round accepted the atomic rework (the lock-ordering change in `checkSession`, dropping its retry loop, and omitting the disconnect-time advance were all confirmed safe) but caught two mechanical defects, both fixed here:

- the attempt token was captured as `int` at three sites while the state was `std::int64_t`, so once the token exceeded `int` a current reply could never compare equal and would be discarded;
- one outer catch re-read the session's current token instead of the one belonging to the failed call, which could charge an older failure to a newer attempt — the original defect class.

It also noted the replaced code called `rethrow_exception(nullptr)` unconditionally, which is undefined when the exception is null; the new code guards it.

Its remaining suggestion — checked arithmetic for signed overflow of the token — is not adopted. The token is only compared for equality, and wrapping an `int64_t` is not reachable.

## Not addressed here

- `connected()` is still invoked from the confirmation callback without validating the attempt (`NodeI.cpp`). It cannot produce the stranded state this PR fixes, but it carries a stale-success risk that predates this change; worth its own issue.
- `NodeI::createPublisherSession` can be called with a null session and its `LocalException` catch already dereferences it. Unchanged and not made worse.

### DIFF
--- a/changelog.d/datastorm/6282.md
+++ b/changelog.d/datastorm/6282.md
@@ -1,4 +1,4 @@
-- Fixed a session reconnect failure that could re-deliver a writer's retained samples. A late reply from a superseded
-  session creation attempt was charged against the attempt in progress, so a burst of them could exhaust the reconnect
-  retry budget in milliseconds and destroy a session that was reconnecting normally. The replacement session resumed
-  from the start of the writer's history, and the reader received samples it had already consumed.
+- Fixed a bug where a DataStorm reader could receive samples it had already consumed. When a session reconnected
+  to its peer — typically through a DataStorm node — stale replies from an earlier connection attempt could
+  exhaust the session's reconnect retries and destroy it; the replacement session then received the writer's
+  retained samples from the beginning.
--- a/changelog.d/datastorm/6282.md
+++ b/changelog.d/datastorm/6282.md
@@ -1,0 +1,4 @@
+- Fixed a session reconnect failure that could re-deliver a writer's retained samples. A late reply from a superseded
+  session creation attempt was charged against the attempt in progress, so a burst of them could exhaust the reconnect
+  retry budget in milliseconds and destroy a session that was reconnecting normally. The replacement session resumed
+  from the start of the writer's history, and the reader received samples it had already consumed.
--- a/cpp/src/DataStorm/NodeI.cpp
+++ b/cpp/src/DataStorm/NodeI.cpp
@@ -194,6 +194,12 @@ NodeI::createSessionAsync(
     assert(instance);
 
     shared_ptr<PublisherSessionI> session;
+
+    // Identifies the attempt started here, so a reply that arrives after it has been superseded is discarded rather
+    // than charged against a later attempt's retry budget. It is read once, where the attempt starts, and never
+    // re-read for a failure that has already happened.
+    int64_t connectAttempt = 0;
+
     try
     {
         NodePrx s = *subscriber;
@@ -220,6 +226,11 @@ NodeI::createSessionAsync(
 
         unique_lock<mutex> lock(_mutex);
         session = createPublisherSessionServant(*subscriber);
+
+        // Identifies the attempt started here, so a reply that arrives after it has been superseded is discarded
+        // rather than charged against a later attempt's retry budget.
+        connectAttempt = session->connectAttempt();
+
         s->ice_getConnectionAsync(
             [=, self = shared_from_this()](const auto& connection) mutable
             {
@@ -271,8 +282,8 @@ NodeI::createSessionAsync(
                                 connection,
                                 instance->getTopicFactory()->getTopicWriters());
                         },
-                        [self, subscriber, session](auto ex)
-                        { self->retryPublisherSessionCreation(*subscriber, session, ex); });
+                        [self, subscriber, session, connectAttempt](auto ex)
+                        { self->retryPublisherSessionCreation(*subscriber, session, ex, connectAttempt); });
                 }
                 catch (const CommunicatorDestroyedException&)
                 {
@@ -284,12 +295,12 @@ NodeI::createSessionAsync(
                 }
                 catch (const LocalException&)
                 {
-                    self->retryPublisherSessionCreation(*subscriber, session, current_exception());
+                    self->retryPublisherSessionCreation(*subscriber, session, current_exception(), connectAttempt);
                 }
             },
-            [self = shared_from_this(), session, subscriber, exception](exception_ptr ex)
+            [self = shared_from_this(), session, subscriber, exception, connectAttempt](exception_ptr ex)
             {
-                self->retryPublisherSessionCreation(*subscriber, session, ex);
+                self->retryPublisherSessionCreation(*subscriber, session, ex, connectAttempt);
                 exception(make_exception_ptr(SessionCreationException{SessionCreationError::Internal}));
             });
     }
@@ -307,7 +318,7 @@ NodeI::createSessionAsync(
     {
         assert(session);
         exception(make_exception_ptr(SessionCreationException{SessionCreationError::Internal}));
-        retryPublisherSessionCreation(*subscriber, session, current_exception());
+        retryPublisherSessionCreation(*subscriber, session, current_exception(), connectAttempt);
     }
 }
 
@@ -391,6 +402,10 @@ NodeI::createSubscriberSession(
     // The publisher session is null when we are creating a new session, in response to a topic reader announcement. It
     // is not null when we are attempting to reconnect an existing session.
 
+    // Identifies the attempt started here, so a reply that arrives after it has been superseded is discarded
+    // rather than charged against a later attempt's retry budget.
+    const int64_t connectAttempt = session ? session->connectAttempt() : 0;
+
     try
     {
         subscriber = getNodeWithExistingConnection(instance, subscriber, subscriberConnection);
@@ -398,11 +413,11 @@ NodeI::createSubscriberSession(
         subscriber->initiateCreateSessionAsync(
             _proxy,
             nullptr,
-            [self = shared_from_this(), session, subscriber](exception_ptr ex)
+            [self = shared_from_this(), session, subscriber, connectAttempt](exception_ptr ex)
             {
                 if (session)
                 {
-                    self->retryPublisherSessionCreation(subscriber, session, ex);
+                    self->retryPublisherSessionCreation(subscriber, session, ex, connectAttempt);
                 }
                 // Else node is shutting down, or the session was established by a concurrent call.
             });
@@ -419,7 +434,7 @@ NodeI::createSubscriberSession(
     {
         if (session)
         {
-            retryPublisherSessionCreation(subscriber, session, current_exception());
+            retryPublisherSessionCreation(subscriber, session, current_exception(), connectAttempt);
         }
         // Else node is shutting down, or the session was established by a concurrent call.
     }
@@ -437,6 +452,10 @@ NodeI::createPublisherSession(
 
     auto traceLevels = instance->getTraceLevels();
 
+    // Identifies the attempt started here, so a reply that arrives after it has been superseded is discarded
+    // rather than charged against a later attempt's retry budget.
+    int64_t connectAttempt = session ? session->connectAttempt() : 0;
+
     try
     {
         auto p = getNodeWithExistingConnection(instance, publisher, publisherConnection);
@@ -445,6 +464,7 @@ NodeI::createPublisherSession(
         if (!session)
         {
             session = createSubscriberSessionServant(publisher);
+            connectAttempt = session->connectAttempt();
             if (session->checkSession())
             {
                 // The session is already connected.
@@ -465,8 +485,8 @@ NodeI::createPublisherSession(
             uncheckedCast<SubscriberSessionPrx>(session->getProxy()),
             false,
             nullptr,
-            [self = shared_from_this(), publisher, session](exception_ptr ex)
-            { self->retrySubscriberSessionCreation(publisher, session, ex); });
+            [self = shared_from_this(), publisher, session, connectAttempt](exception_ptr ex)
+            { self->retrySubscriberSessionCreation(publisher, session, ex, connectAttempt); });
     }
     catch (const CommunicatorDestroyedException&)
     {
@@ -480,7 +500,7 @@ NodeI::createPublisherSession(
     }
     catch (const LocalException&)
     {
-        retrySubscriberSessionCreation(publisher, session, current_exception());
+        retrySubscriberSessionCreation(publisher, session, current_exception(), connectAttempt);
         throw SessionCreationException{SessionCreationError::Internal};
     }
 }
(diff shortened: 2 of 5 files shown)


externl (approved):
Looks good.

Agreed on the `connected()` gap you flag under "Not addressed here" — worth its own issue. A stale success from a superseded attempt is only ignored while `_session` is set, so when the session is disconnected it can install an already-dead connection, and a bare disconnect schedules nothing. Whether that strands the session depends on the connection manager invoking the callback for a connection that is already closed.

changelog.d/datastorm/6282.md
```diff
@@ -0,0 +1,4 @@
+- Fixed a session reconnect failure that could re-deliver a writer's retained samples. A late reply from a superseded
+  session creation attempt was charged against the attempt in progress, so a burst of them could exhaust the reconnect
+  retry budget in milliseconds and destroy a session that was reconnecting normally. The replacement session resumed
+  from the start of the writer's history, and the reader received samples it had already consumed.
```
bernardnormier: The entry leads with the mechanism ("superseded session creation attempt", "retry budget"). Let's lead with what a DataStorm user observes and compress the cause:
```suggestion
- Fixed a bug where a DataStorm reader could receive samples it had already consumed. When a session reconnected
  to its peer — typically through a DataStorm node — stale replies from an earlier connection attempt could
  exhaust the session's reconnect retries and destroy it; the replacement session then received the writer's
  retained samples from the beginning.
```

cpp/src/DataStorm/SessionI.h
```diff
@@ -303,10 +303,10 @@ namespace DataStormI
         /// session.
         [[nodiscard]] bool handleDisconnected(const Ice::ConnectionPtr&, std::exception_ptr);
 
-        [[nodiscard]] bool retry(DataStormContract::NodePrx, std::exception_ptr);
         void destroyImpl(const std::exception_ptr&);
 
-        // The implementations of disconnected and retry; called with the session mutex locked.
+        // The implementations of checkSession, disconnected and retry; called with the session mutex locked.
+        [[nodiscard]] bool checkSessionImpl();
         [[nodiscard]] bool disconnectedImpl(const Ice::ConnectionPtr&, std::exception_ptr);
```
bernardnormier: Please remove `[[nodiscard]]` here: it gives the wrong impression that the returned bool is the point of calling, while `disconnectedImpl` does its work through side effects — `checkSessionImpl` legitimately ignores the result (via the `[[maybe_unused]]` workaround). It's right on `retryImpl` and `sessionCreationFailed`, where ignoring `false` would strand the session removal.
```suggestion
        bool disconnectedImpl(const Ice::ConnectionPtr&, std::exception_ptr);
```

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -1016,39 +1070,38 @@ SessionI::getConnection() const
...
-            }
-            return true;
+            _connection->throwException();
         }
-        else
+        catch (const LocalException&)
         {
+            // Ignore the result: either this call disconnected the session, or another thread already did. Both leave
+            // the session disconnected, which is what the caller needs to know.
+            [[maybe_unused]] bool disconnected = disconnectedImpl(_connection, current_exception());
```
bernardnormier: With `[[nodiscard]]` removed from `disconnectedImpl` (see SessionI.h), this collapses to a plain call:
```suggestion
            // Either this call disconnected the session, or another thread already did. Both leave the session
            // disconnected, which is what the caller needs to know.
            disconnectedImpl(_connection, current_exception());
```

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -194,6 +194,12 @@ NodeI::createSessionAsync(
     assert(instance);
 
     shared_ptr<PublisherSessionI> session;
+
+    // Identifies the attempt started here, so a reply that arrives after it has been superseded is discarded rather
+    // than charged against a later attempt's retry budget. It is read once, where the attempt starts, and never
+    // re-read for a failure that has already happened.
+    int64_t connectAttempt = 0;
```
bernardnormier: Nit: this comment appears four times in this file (here and at lines 230, 405 and 455), and the declarations in NodeI.h and SessionI.h already explain the parameter. I'd drop all four copies.
```suggestion
    int64_t connectAttempt = 0;
```

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -220,6 +226,11 @@ NodeI::createSessionAsync(
 
         unique_lock<mutex> lock(_mutex);
         session = createPublisherSessionServant(*subscriber);
+
+        // Identifies the attempt started here, so a reply that arrives after it has been superseded is discarded
+        // rather than charged against a later attempt's retry budget.
+        connectAttempt = session->connectAttempt();
```
bernardnormier: Same as line 198.
```suggestion
        connectAttempt = session->connectAttempt();
```

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -391,18 +402,22 @@ NodeI::createSubscriberSession(
     // The publisher session is null when we are creating a new session, in response to a topic reader announcement. It
     // is not null when we are attempting to reconnect an existing session.
 
+    // Identifies the attempt started here, so a reply that arrives after it has been superseded is discarded
+    // rather than charged against a later attempt's retry budget.
+    const int64_t connectAttempt = session ? session->connectAttempt() : 0;
```
bernardnormier: Same as line 198.
```suggestion
    const int64_t connectAttempt = session ? session->connectAttempt() : 0;
```

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -437,6 +452,10 @@ NodeI::createPublisherSession(
 
     auto traceLevels = instance->getTraceLevels();
 
+    // Identifies the attempt started here, so a reply that arrives after it has been superseded is discarded
+    // rather than charged against a later attempt's retry budget.
+    int64_t connectAttempt = session ? session->connectAttempt() : 0;
```
bernardnormier: Same as line 198.
```suggestion
    int64_t connectAttempt = session ? session->connectAttempt() : 0;
```

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -489,27 +509,13 @@ void
...
-    catch (...)
-    {
-        // Let Session::retry handle the exception.
-    }
-
-    if (!session->retry(node, ex))
+    // SessionI::sessionCreationFailed classifies the failure and decides what to do with it in one step under the
+    // session mutex. Evaluating the session state here instead would not be race-free: checking it can itself
+    // disconnect the session, which changes the very state the decision depends on.
+    if (!session->sessionCreationFailed(node, ex, connectAttempt))
```
bernardnormier: Nit: this restates `sessionCreationFailed`'s own doc comment (SessionI.h), here and in `retryPublisherSessionCreation` below. I'd drop both copies.
```suggestion
    if (!session->sessionCreationFailed(node, ex, connectAttempt))
```

cpp/src/DataStorm/NodeI.cpp
```diff
@@ -536,27 +542,13 @@ void
...
```
bernardnormier: Same as line 515.
```suggestion
    if (!session->sessionCreationFailed(node, ex, connectAttempt))
```

bernardnormier (approved):
The attempt-token protocol checks out — I traced every discard path to a live recovery owner (a connected session, a scheduled retry task, or a completed removal), and holding the lock across the disconnect decision preserves the old `checkSession` loop's semantics.

Two things beyond the inline comments:

- `SessionI::disconnected(const Ice::ConnectionPtr&, std::exception_ptr)` (SessionI.cpp:699, declared at SessionI.h:296) lost its last caller with the `checkSession` rewrite, so it's now dead code — please remove it.
- The windows-2022 release failure is unrelated infrastructure (the Windows SDK installer failed during runner setup); it just needs a re-run.

changelog.d/datastorm/6282.md
pepone: 🤖 **Claude:** Done in d92055537f — applied as suggested.

cpp/src/DataStorm/SessionI.h
pepone: 🤖 **Claude:** Done in d92055537f. Agreed: `handleDisconnected` is the only caller that acts on the result, and the `[[maybe_unused]]` workaround was the tell.

cpp/src/DataStorm/SessionI.cpp
pepone: 🤖 **Claude:** Done in d92055537f.

cpp/src/DataStorm/NodeI.cpp
pepone: 🤖 **Claude:** Done in d92055537f — dropped all four copies.

cpp/src/DataStorm/NodeI.cpp
pepone: 🤖 **Claude:** Done in d92055537f.

cpp/src/DataStorm/NodeI.cpp
pepone: 🤖 **Claude:** Done in d92055537f.

cpp/src/DataStorm/NodeI.cpp
pepone: 🤖 **Claude:** Done in d92055537f.

cpp/src/DataStorm/NodeI.cpp
pepone: 🤖 **Claude:** Done in d92055537f — dropped both copies.

cpp/src/DataStorm/NodeI.cpp
pepone: 🤖 **Claude:** Done in d92055537f.

pepone:
🤖 **Claude:** All nine inline comments applied in d92055537f, plus the dead-code removal.

`SessionI::disconnected(const Ice::ConnectionPtr&, std::exception_ptr)` is gone — confirmed dead: on `main` its only caller was `checkSession`'s unlock/retry loop (`SessionI.cpp:1057`), which this PR replaced with a direct `disconnectedImpl` call.

One correction on the CI failure. It isn't the Windows SDK installer — runner setup succeeded, and the `Test release on windows-2022` step failed with `289 succeeded and 1 failed`. The failure is `Ice/inactivityTimeout` (Java, `ws,compress,ipv6,serialize,mx,x64`), a known flake. Unrelated either way, and the push has started a fresh run.

Verified locally: `make -C cpp -j10 srcs tests`, clang-format-19 clean, `allTests.py --filter=DataStorm` 11/11.

PR #6291, opened 2026-07-30
Make ice_getCachedConnection return null instead of throwing during a failed connection establishment

### LINKED ISSUE #6278: ice_getCachedConnection can throw during connection-establishment failure cleanup
`ObjectPrx::ice_getCachedConnection` is documented to return the cached connection if there is one and null otherwise. In C++ it is also declared `noexcept` (`cpp/include/Ice/Proxy.h`). It can nevertheless throw.

The proxy's `RequestHandlerCache` caches a request handler at the start of the first invocation — before any connection exists — so while a connect is in flight the cached handler is a `ConnectRequestHandler`, and `getCachedConnection` obtains the connection from it. `ConnectRequestHandler::getConnection` rethrows the stored connect failure:

- `ConnectRequestHandler::setException` publishes `_exception` first, then fails the queued requests; the handler is evicted from the proxy's cache (`clearCachedRequestHandler`) only as those requests process their failure.
- A concurrent `ice_getCachedConnection` call in this window finds the handler still cached and reaches the `rethrow_exception(_exception)` branch in `ConnectRequestHandler::getConnection` (`cpp/src/Ice/ConnectRequestHandler.cpp`).

In C++ the exception escapes through the `noexcept` frame and calls `std::terminate`. C# and Java have the same publish-then-evict ordering with a throwing getter; there the call throws the connect exception instead of returning null, contradicting the documented contract.

`RequestHandlerCache::getCachedConnection` is the only caller of `RequestHandler::getConnection`, so the throwing branch has no other consumer — no caller can legitimately receive the exception.

Proposed fix: make `ConnectRequestHandler.getConnection` return null instead of rethrowing (C++, C#, Java), matching the documented behavior: return the connection, in whatever state it is, if there is one, otherwise return null — never throw.

bernardnormier:
`ice_getCachedConnection` is documented to return the cached connection if there is one and null otherwise, and in C++ it is also declared `noexcept`. It could nevertheless throw: while a connection establishment is failing, `ConnectRequestHandler.setException` publishes the exception before the queued requests evict the handler from the proxy's request-handler cache, so a concurrent `ice_getCachedConnection` call in that window reached the rethrow branch in `ConnectRequestHandler.getConnection`. In C++, the exception escaped through the `noexcept` frame and terminated the application; in C# and Java, the call threw the connection-establishment exception where the documentation promises null.

The window is not always brief. When no request is queued on the handler at `setException` time — the establishment fails between `getRequestHandler()` and `sendAsyncRequest()`, including a synchronous failure inside `getRequestHandler()` — nothing evicts the handler: it stays in the cache, and `ice_getCachedConnection` kept throwing, until the pending `sendAsyncRequest` throws and clears it.

## The fix

`ConnectRequestHandler.getConnection` now returns the connection in whatever state it is — null when the establishment failed — instead of rethrowing the stored exception (C++, C#, Java). A comment on `RequestHandler.getConnection` records the contract: implementations never throw; a null return is how the absence of a connection is reported. In C++, the contract is also compiler-enforced: `RequestHandler::getConnection`, its three implementations, and `RequestHandlerCache::getCachedConnection` are now `noexcept`.

## Why the ripple effect is contained

The throwing branch had no legitimate consumer — the entire call chain is a straight line, verified in each mapping:

- `RequestHandlerCache.getCachedConnection` is the **only** caller of `RequestHandler.getConnection`. The other implementations of that method (`FixedRequestHandler` / `ConnectionRequestHandler`, and `CollocatedRequestHandler`) already never throw; `ConnectRequestHandler`'s was the only throwing one.
- The proxy method `ice_getCachedConnection` is in turn the **only** caller of `RequestHandlerCache.getCachedConnection`.

So the change's entire observable surface is the return value of `ice_getCachedConnection` during the failed-establishment window: it now returns null — the same value it returns once the failed handler is evicted from the cache.

JavaScript has the same code shape but is not affected: `setException` and the cache eviction run synchronously in the same event-loop turn, so user code cannot observe the window. The JS code is left unchanged.

No tests: hitting the window requires a thread to call `ice_getCachedConnection` while the runtime is between publishing the connection-establishment failure and completing the queued requests, and there is no deterministic way to hold the runtime in that state from the test harness.

Fixes #6278.

### DIFF
--- a/cpp/src/Ice/ConnectRequestHandler.cpp
+++ b/cpp/src/Ice/ConnectRequestHandler.cpp
@@ -62,20 +62,10 @@ Ice::ConnectionIPtr
 ConnectRequestHandler::getConnection()
 {
     lock_guard lock(_mutex);
-    //
-    // First check for the connection, it's important otherwise the user could first get a connection
-    // and then the exception if he tries to obtain the proxy cached connection multiple times (the
-    // exception can be set after the connection is set if the flush of pending requests fails).
-    //
-    if (_connection)
-    {
-        return _connection;
-    }
-    else if (_exception)
-    {
-        rethrow_exception(_exception);
-    }
-    return nullptr;
+    // Return the connection in whatever state it is; when the connection establishment failed, return null. Like all
+    // getConnection implementations, this function never throws: a null return is how the absence of a connection is
+    // reported.
+    return _connection;
 }
 
 void
--- a/cpp/src/Ice/ConnectRequestHandler.cpp
+++ b/cpp/src/Ice/ConnectRequestHandler.cpp
@@ -59,10 +59,10 @@ ConnectRequestHandler::asyncRequestCanceled(const OutgoingAsyncBasePtr& outAsync
 }
 
 Ice::ConnectionIPtr
-ConnectRequestHandler::getConnection()
+ConnectRequestHandler::getConnection() noexcept
 {
     lock_guard lock(_mutex);
-    // Return the connection in whatever state it is; when the connection establishment failed, return null. Like all
+    // Return the connection in whatever state it is; if the connection establishment failed, return null. Like all
     // getConnection implementations, this function never throws: a null return is how the absence of a connection is
     // reported.
     return _connection;
--- a/csharp/src/Ice/Internal/ConnectRequestHandler.cs
+++ b/csharp/src/Ice/Internal/ConnectRequestHandler.cs
@@ -65,7 +65,7 @@ internal class ConnectRequestHandler : RequestHandler, Reference.GetConnectionCa
     {
         lock (_mutex)
         {
-            // Return the connection in whatever state it is; when the connection establishment failed, return null.
+            // Return the connection in whatever state it is; if the connection establishment failed, return null.
             // Like all getConnection implementations, this method never throws: a null return is how the absence of a
             // connection is reported.
             return _connection;
--- a/csharp/src/Ice/Internal/ConnectRequestHandler.cs
+++ b/csharp/src/Ice/Internal/ConnectRequestHandler.cs
@@ -65,20 +65,10 @@ internal class ConnectRequestHandler : RequestHandler, Reference.GetConnectionCa
     {
         lock (_mutex)
         {
-            //
-            // First check for the connection, it's important otherwise the user could first get a connection
-            // and then the exception if he tries to obtain the proxy cached connection multiple times (the
-            // exception can be set after the connection is set if the flush of pending requests fails).
-            //
-            if (_connection != null)
-            {
-                return _connection;
-            }
-            else if (_exception != null)
-            {
-                throw _exception;
-            }
-            return null;
+            // Return the connection in whatever state it is; when the connection establishment failed, return null.
+            // Like all getConnection implementations, this method never throws: a null return is how the absence of a
+            // connection is reported.
+            return _connection;
         }
     }
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectRequestHandler.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectRequestHandler.java
@@ -51,9 +51,9 @@ final class ConnectRequestHandler
 
     @Override
     public synchronized ConnectionI getConnection() {
-        // Return the connection in whatever state it is; when the connection establishment failed,
-        // return null. Like all getConnection implementations, this method never throws: a null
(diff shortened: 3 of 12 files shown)


csharp/src/Ice/Internal/ConnectRequestHandler.cs
```diff
@@ -65,20 +65,10 @@ public Ice.ConnectionI getConnection()
...
-            if (_connection != null)
-            {
-                return _connection;
-            }
-            else if (_exception != null)
-            {
-                throw _exception;
-            }
-            return null;
+            // Return the connection in whatever state it is; when the connection establishment failed, return null.
```
InsertCreativityHere:
```suggestion
            // Return the connection in whatever state it is; if the connection establishment failed, return null.
```

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/ConnectRequestHandler.java
```diff
@@ -51,16 +51,10 @@ public void asyncRequestCanceled(OutgoingAsyncBase outAsync, LocalException ex)
...
-        // pending requests fails).
-        if (_connection != null) {
-            return _connection;
-        } else if (_exception != null) {
-            throw (LocalException) _exception.fillInStackTrace();
-        }
-        return null;
+        // Return the connection in whatever state it is; when the connection establishment failed,
+        // return null. Like all getConnection implementations, this method never throws: a null
+        // return is how the absence of a connection is reported.
```
InsertCreativityHere: Minor nit, and we can use 120 columns, no need to stop at 100.
```suggestion
        // Return the connection in whatever state it is; if the connection establishment failed, return null.
        // Like all getConnection implementations, this method never throws: a null return is how the absence of a
        // connection is reported.
```

cpp/src/Ice/ConnectRequestHandler.cpp
```diff
@@ -62,20 +62,10 @@ Ice::ConnectionIPtr
...
-    if (_connection)
-    {
-        return _connection;
-    }
-    else if (_exception)
-    {
-        rethrow_exception(_exception);
-    }
-    return nullptr;
+    // Return the connection in whatever state it is; when the connection establishment failed, return null. Like all
```
InsertCreativityHere:
```suggestion
    // Return the connection in whatever state it is; if the connection establishment failed, return null. Like all
```

InsertCreativityHere (approved):
Looks good, just a nit on the wording of the comment.

Also yep, the only caller for any of the `RequestHandler::getConnection`s is indeed just `RequestHandlerCache::getCachedConnection`, and in turn `ObjectPrx::ice_getCachedConnection`... So no other side-effects.

pepone (approved):
Approve.

Verified the call chain independently in all three mappings: `RequestHandlerCache::getCachedConnection` is the only caller of `RequestHandler::getConnection`, and `ice_getCachedConnection` the only caller of that. `FixedRequestHandler` and `CollocatedRequestHandler` already never throw, so the observable surface really is just the return value during the failed-establishment window.

Two non-blocking notes:

1. The description understates the window slightly. When `_requests` is empty at `setException` time — the connect fails between `getRequestHandler()` and `sendAsyncRequest()` at `OutgoingAsync.cpp:436-437`, including a synchronous failure inside `getRequestHandler()` — nothing evicts the handler at all. It stays cached until that pending `sendAsyncRequest` throws and clears it. Doesn't change the fix, just makes the window wider than "while the queued requests process their failure."

2. In C++ we can make the contract compiler-enforced rather than a comment: `noexcept` on `RequestHandler::getConnection` plus the three overrides, and on `RequestHandlerCache::getCachedConnection`. `ObjectPrx::ice_getCachedConnection` is already `noexcept`, so the promise exists at the boundary — this is what propagates it down so an implementation can't silently break it again. I built it: no symbol change (Itanium doesn't mangle exception specs), no vtable change, and none of the classes are `ICE_API`-exported. Fine either way on whether it rides along here or lands separately.

bernardnormier:
@pepone Both notes taken in 5e3db742cb: the description now covers the empty-`_requests` case — the failed handler staying cached until the pending `sendAsyncRequest` clears it — and `RequestHandler::getConnection`, its three implementations, and `RequestHandlerCache::getCachedConnection` are now `noexcept`, so the contract is compiler-enforced rather than comment-enforced.

bernardnormier:
@InsertCreativityHere Applied all three wording suggestions in 5e3db742cb (with the Java comment reflowed to 120 columns). Thanks for double-checking the call chain.

Issue #6192, opened 2026-07-27
An exception thrown by a custom logger or connection observer escapes ConnectionI::setState and terminates the process


bernardnormier:
Split out from #6103, whose main finding — `PropertyException` escaping the `noexcept` `ConnectionI` constructor — was fixed by #6184. This issue covers the remaining part, re-verified against the source on `main` (2d036c3fcb).

## Summary

`ConnectionI::setState` calls application-provided code — the configured `Logger` and the instrumentation observers — without an exception barrier, and it is invoked from six `noexcept` members. An exception thrown by one of these application callbacks therefore unwinds into a `noexcept` frame and calls `std::terminate`, aborting the process with no indication of which callback threw.

The application calls made without a barrier, all in `cpp/src/Ice/ConnectionI.cpp`:

- `Logger::warning` via `Warning out(_logger)` when a connection is aborted by an unexpected exception ([`setState(State, exception_ptr)`, lines 2033–2045](https://github.com/zeroc-ice/ice/blob/2d036c3fcb77bfc4d7422ebf7f6d3dfd699f652d/cpp/src/Ice/ConnectionI.cpp#L2033-L2045)).
- `Logger::error` via `Error out(_logger)` — ironically inside the `catch` handler that guards the state-transition switch ([lines 2189–2193](https://github.com/zeroc-ice/ice/blob/2d036c3fcb77bfc4d7422ebf7f6d3dfd699f652d/cpp/src/Ice/ConnectionI.cpp#L2189-L2193)).
- The observer block ([lines 2195–2242](https://github.com/zeroc-ice/ice/blob/2d036c3fcb77bfc4d7422ebf7f6d3dfd699f652d/cpp/src/Ice/ConnectionI.cpp#L2195-L2242)): `CommunicatorObserver::getConnectionObserver`, `ConnectionObserver::attach`/`detach` (through `_observer.attach`), and `ConnectionObserver::failed` — all implemented by the application when instrumentation is enabled.

The six `noexcept` members that call `setState` directly: `abort`, `idleCheck`, `inactivityCheck`, `connectTimedOut`, `closeTimedOut`, and `sendHeartbeat` (`cpp/src/Ice/ConnectionI.h:145,227-239`). Most run on timer or thread-pool threads, so the application has no frame of its own on the stack and no way to catch anything.

[...]

PR #6292, opened 2026-07-30
Document that Logger and instrumentation observer implementations must not throw

bernardnormier:
Adds a `@remark` to the `Logger` class, the `Ice::Instrumentation::Observer` base class (covering all derived observer interfaces), and `CommunicatorObserver`, stating that the Ice runtime calls these interfaces from contexts where exceptions cannot be handled — such as destructors and `noexcept` functions — and that implementations must not throw: a thrown exception can terminate the process.

Neither interface previously said anything about exceptions, so an application had no way to know that a throwing logger or observer is fatal. The requirement itself is not new: for example, the `Trace`/`Warning`/`Error` helpers flush to the logger from their implicitly-`noexcept` destructors, so this has always been the de facto contract throughout the runtime.

This addresses the documentation part of #6192. Whether to additionally enforce the contract with a central exception barrier is deferred to 3.9.0 with that issue.

No changelog entry: documentation-only, and only applications whose own logger or observer implementations throw are affected.

### DIFF
--- a/cpp/include/Ice/Instrumentation.h
+++ b/cpp/include/Ice/Instrumentation.h
@@ -103,6 +103,9 @@ namespace Ice::Instrumentation
     };
 
     /// The base class for Ice observers.
+    /// @remark The Ice runtime calls observers from contexts where exceptions cannot be handled, such as `noexcept`
+    /// functions. Implementations of this interface and its derived interfaces must not throw exceptions: a thrown
+    /// exception can terminate the process.
     /// @headerfile Ice/Ice.h
     class Observer
     {
@@ -240,6 +243,9 @@ namespace Ice::Instrumentation
     /// objects. This interface should be implemented by add-ins that wish to observe Ice objects in order to collect
     /// statistics. An instance of this interface can be provided to the Ice runtime through the Ice communicator
     /// initialization data.
+    /// @remark The Ice runtime calls this interface from contexts where exceptions cannot be handled, such as
+    /// `noexcept` functions. An implementation of this interface must not throw exceptions: a thrown exception can
+    /// terminate the process.
     /// @headerfile Ice/Ice.h
     class CommunicatorObserver
     {
--- a/cpp/include/Ice/Instrumentation.h
+++ b/cpp/include/Ice/Instrumentation.h
@@ -103,9 +103,9 @@ namespace Ice::Instrumentation
     };
 
     /// The base class for Ice observers.
-    /// @remark The Ice runtime calls observers from contexts where exceptions cannot be handled, such as `noexcept`
-    /// functions. Implementations of this interface and its derived interfaces must not throw exceptions: a thrown
-    /// exception can terminate the process.
+    /// @remark The Ice runtime calls observers from contexts where exceptions cannot be handled, such as destructors
+    /// and `noexcept` functions. Implementations of this interface and its derived interfaces must not throw
+    /// exceptions: a thrown exception can terminate the process.
     /// @headerfile Ice/Ice.h
     class Observer
     {
@@ -244,8 +244,8 @@ namespace Ice::Instrumentation
     /// statistics. An instance of this interface can be provided to the Ice runtime through the Ice communicator
     /// initialization data.
     /// @remark The Ice runtime calls this interface from contexts where exceptions cannot be handled, such as
-    /// `noexcept` functions. An implementation of this interface must not throw exceptions: a thrown exception can
-    /// terminate the process.
+    /// destructors and `noexcept` functions. An implementation of this interface must not throw exceptions: a thrown
+    /// exception can terminate the process.
     /// @headerfile Ice/Ice.h
     class CommunicatorObserver
     {
--- a/cpp/include/Ice/Instrumentation.h
+++ b/cpp/include/Ice/Instrumentation.h
@@ -244,8 +244,8 @@ namespace Ice::Instrumentation
     /// statistics. An instance of this interface can be provided to the Ice runtime through the Ice communicator
     /// initialization data.
     /// @remark The Ice runtime calls this interface from contexts where exceptions cannot be handled, such as
-    /// destructors and `noexcept` functions. An implementation of this interface must not throw exceptions: a thrown
-    /// exception can terminate the process.
+    /// `noexcept` functions. An implementation of this interface must not throw exceptions: a thrown exception can
+    /// terminate the process.
     /// @headerfile Ice/Ice.h
     class CommunicatorObserver
     {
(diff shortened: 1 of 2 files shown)


externl (approved):
Looks good.

Should the same remark go on the other mappings? C# and Java `Logger` say nothing about exceptions either (`csharp/src/Ice/Logger.cs:7-10`, `java/.../Ice/Logger.java:5-8`). #6192 is C++-specific because `std::terminate` is, but "don't throw from a logger or observer" is the same contract everywhere — there it escapes mid-transition and leaves the connection state out of sync rather than aborting, which is the outcome the issue describes for the non-`noexcept` C++ paths.

pepone (commented):
Verified the premise: `Trace::~Trace()` → `flush()` → `_logger->trace()` (`cpp/src/Ice/LoggerUtil.cpp:37-49`) and `LoggerOutput::~LoggerOutput()` (`cpp/include/Ice/LoggerUtil.h:95`) do reach the logger from implicitly-`noexcept` destructors, so the contract is real and undocumented. Two things on the wording.

**1. The `Logger` remark over-reaches.** It says "an implementation of *this interface* must not throw", which covers all six virtuals — but `SysLoggerI::cloneWithPrefix` deliberately throws:

```cpp
// cpp/src/Ice/SysLoggerI.cpp:170-174
throw FeatureNotSupportedException(__FILE__, __LINE__, "the Ice syslog logger does not support cloneWithPrefix");
```

So an Ice-shipped logger violates the contract as written. It also contradicts the 3.9.0 plan in #6192, which exempts `getPrefix`/`cloneWithPrefix` — correctly: their only non-implementation call sites are `LoggerAdminI.cpp:347`, `:413`, `:640`, all admin-facet dispatch that can handle exceptions. Suggest scoping to the logging functions:

```suggestion
    /// @remark The Ice runtime calls #print, #trace, #warning and #error from contexts where exceptions cannot be
    /// handled, such as destructors and `noexcept` functions. Implementations of these functions must not throw
    /// exceptions: a thrown exception can terminate the process.
```

That also keeps the doc consistent with what 3.9.0 would actually enforce.

**2. The `Observer` remark should mention destructors too.** It says only "such as `noexcept` functions", but `ObserverHelperT::~ObserverHelperT()` calls `_observer->detach()` (`cpp/include/Ice/ObserverHelper.h:33-39`) — that's the exit path for essentially every instrumented object. Match the `Logger` wording: "such as destructors and `noexcept` functions".

Minor, take it or leave it: the derived observer interfaces don't carry the remark and Doxygen doesn't inherit `@remark`, so someone landing on `ConnectionObserver` sees nothing. The base-class text does say "and its derived interfaces", so it's covered for anyone who follows the link — just not repeating it six times has a cost.

Rest looks good: `@remark` singular is the right tag for hand-written headers (103 uses vs `@remarks` almost entirely in generated code), tag order matches `IncomingRequest.h:16-18`, lines fit under 120, `ObserverUpdater` correctly left alone since Ice implements it, and skipping the changelog is right for docs-only.

bernardnormier:
@pepone Both taken in ff04b82ce0: the `Logger` remark now names `print`/`trace`/`warning`/`error` (your wording), and both observer remarks say "destructors and `noexcept` functions" — `CommunicatorObserver` had the same gap as `Observer`, so I aligned the two. On the derived interfaces: leaving the remark on the base only; the "and its derived interfaces" sentence covers it for anyone who follows the link, and six copies of the same remark would be maintenance noise.

bernardnormier:
@externl Agreed the contract is the same everywhere — but I'd rather keep this PR C++-only: it's the doc companion to #6192, and the consequence wording ("destructors and `noexcept` functions", process termination) is C++-specific. C# and Java deserve their own pass with wording that matches what actually happens there (the exception escapes mid-transition and leaves the connection state out of sync). Future PR.

pepone (approved):
Both asks landed, and I re-verified them against the source rather than the reply.

The `Logger` narrowing is right: `getPrefix`/`cloneWithPrefix` are only called from `LoggerAdminI.cpp:347`, `:413`, `:640` — admin-facet dispatch that can handle exceptions — so `SysLoggerI::cloneWithPrefix` throwing (`SysLoggerI.cpp:173`) no longer contradicts the doc, and the text matches what #6192 would enforce in 3.9.0. The `Observer` destructor mention is right too: `ObserverHelperT::~ObserverHelperT` → `_observer->detach()` (`cpp/include/Ice/ObserverHelper.h:33-39`).

One thing on the new `CommunicatorObserver` remark. The `noexcept` half checks out on two independent paths: `Instance::destroy() noexcept` (`Instance.h:130`) calls `setObserverUpdater(nullptr)` (`Instance.cpp:1674`), and `setState` → `getConnectionObserver` (`ConnectionI.cpp:2203`) is reached from all six `noexcept` members the issue names. But the "destructors" half I can't substantiate for this interface — I walked every call site (`ConnectionI.cpp:631`/`:2203`, `ObserverMiddleware.cpp:22`, `ThreadPool.cpp:1005`, `ConnectionFactory.cpp:933`, `ObserverHelper.cpp:19,30,41,53`, `Instance.cpp:249,1509,1674`, `IPEndpointI.cpp:480,595`) and none is in a destructor; `~ConnectionI`/`~Instance`/`~ThreadPool` are assert-only. Destructors are the `Observer::detach` path, not the `CommunicatorObserver` getters, so aligning the two remarks copied a claim that only holds for one. It over-warns rather than under-warns, so genuinely take it or leave it:

```suggestion
    /// @remark The Ice runtime calls this interface from contexts where exceptions cannot be handled, such as
    /// `noexcept` functions. An implementation of this interface must not throw exceptions: a thrown exception can
    /// terminate the process.
```

On keeping this C++-only: agreed, the consequence wording really is mapping-specific. But there's no tracker for the C#/Java pass — #6192 is the only hit and it's C++-only under 3.9.0. Worth filing a follow-up before this merges so it doesn't evaporate.

Rest is clean. `#print`/`#trace`/`#warning`/`#error` is the established convention in these headers (`Communicator.h` uses `#member` throughout) and resolves in class scope; all added lines fit under 120. Scope is correctly drawn — `ConnectionI::executeCallback(const CloseCallback&) noexcept` already has its own try/catch barrier (`ConnectionI.cpp:806-808`), so close callbacks rightly get no such warning.

bernardnormier:
Took the suggestion in eb440c1015 — your call-site walk is right: destructors are the `Observer::detach` path, and the `CommunicatorObserver` functions are reached from `noexcept` functions only, so the aligned wording claimed more than that interface's call sites support. And the C#/Java docs pass is now tracked by #6334, so it won't evaporate.

Issue #6294, opened 2026-07-30
Generalize OutgoingAsync exception handling beyond Ice exceptions in C# and Java


bernardnormier:
In C++, the invocation completion machinery is `exception_ptr`-based end to end, so an invocation can complete with *any* exception — including a non-Ice exception thrown by an application callback on an internal thread (for example, an instrumentation observer's `retried()` during a queued retry, see #6111).

In C# and Java, the `OutgoingAsync` exception plumbing is typed to Ice exceptions — `exceptionImpl(Ice.Exception)` / `handleInvokeException(Ice.Exception, ...)` in C#, `finished(LocalException)` / `completed(LocalException)` in Java — even though the underlying completion primitives (`Task`, `CompletableFuture`) can carry any exception. A non-Ice exception on an internal path therefore cannot complete the invocation at all.

The 3.8.x mitigation for #6111 contains such an exception in `RetryTask` (`try`/`finally` around `retry()`, plus logging in Java): `Communicator.destroy` no longer hangs, but the invocation's `Task`/`CompletableFuture` never completes. Wrapping in `Ice.UnknownException` was considered and rejected: `UnknownException` means "the server replied with the `UnknownException` reply status" — the wrong semantics for a locally generated failure.

Proposal for 3.9: generalize the exception paths to `System.Exception` (C#) and `Throwable`/`RuntimeException` (Java), so invocations complete with the original exception — mirroring the C++ `exception_ptr` model. The `RetryTask` containment then reduces to the C++ shape, and the two runtimes stop special-casing a corner of a generally mis-typed pipeline.

PR #6295, opened 2026-07-30
Prevent a non-Ice exception from stranding a retry task and hanging Communicator.destroy

### LINKED ISSUE #6111: A throwing instrumentation observer (`retried()`) strands a retry task and hangs `Communicator.destroy()` (C++/C#/Java)
> [!NOTE]

## Summary

Each retry task calls `retry()` and only *then* removes itself from the retry queue, with no `finally`/RAII. `retry()` invokes the user `CommunicatorObserver`'s `InvocationObserver.retried()`. If that user callback throws a **non-Ice** exception, it escapes before the task is removed; the timer/executor absorbs it, but `Communicator.destroy()` waits for the retry-task set to become empty and therefore hangs forever, with the invocation permanently stranded.

**Files:**
- C++ [`RetryQueue.cpp:23-35`](https://github.com/zeroc-ice/ice/blob/1579b6928e0a1c62fc54215e24a4b5f0bb117f97/cpp/src/Ice/RetryQueue.cpp#L23-L35) — `_outAsync->retry(); _queue->remove(shared_from_this());` (no finally); `retried()` at `OutgoingAsync.cpp:428/479`, outer catch is `catch (const Exception&)` only; destroy waits at `RetryQueue.cpp:98-119`.
- C# [`RetryQueue.cs:16-27`](https://github.com/zeroc-ice/ice/blob/1579b6928e0a1c62fc54215e24a4b5f0bb117f97/csharp/src/Ice/Internal/RetryQueue.cs#L16-L27) — same shape; outer catch `Ice.Exception` only; destroy waits at `RetryQueue.cs:101-104`.
- Java [`RetryTask.java:14-25`](https://github.com/zeroc-ice/ice/blob/1579b6928e0a1c62fc54215e24a4b5f0bb117f97/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/RetryTask.java#L14-L25) — worst: `cancel()` claims the task first, so after a throw it can never be reclaimed; the exception is swallowed into the executor's Future and not even logged.

The exposure generalizes to `_childObserver.failed()`/`detach()` on the same retry path.

## Fix

Remove the task in `finally`/RAII (preserving the requirement that removal happen only after the retry attempt finishes), and wrap the whole `runTimerTask`/`run` so a non-Ice observer throw cannot strand it.

**Severity:** Low (requires a user observer that throws a non-Ice exception — misuse — but the failure is an unbounded `Communicator.destroy()` hang, and Java doesn't even log it).

bernardnormier:
A non-Ice exception escaping `ProxyOutgoingAsyncBase.retry()` left the retry task in the retry queue forever: `Communicator.destroy()` waits for the queue to empty and hung permanently, and in Java the exception wasn't even logged (the scheduled executor swallows task exceptions into an unread `Future`).

This is defense-in-depth, not support for new behavior. Every sanctioned flow on the retry path — connection failures, `NoEndpointException`, retry exhaustion — raises an Ice exception, which was already caught and completed the invocation; the new code is unreachable for those flows. What can actually reach it:

- an instrumentation observer callback (`InvocationObserver.retried()`, `failed()`, `getRemoteObserver()`) that throws a non-Ice exception — forbidden since #6292, but nothing enforces the contract until the `noexcept` work planned with #6192;
- an allocation failure (`std::bad_alloc` / `OutOfMemoryException` / `OutOfMemoryError`) in the retry machinery — rare but not misuse;
- a latent runtime bug under `sendAsyncRequest`.

For all three, a silent permanent hang of `destroy()` is the worst possible outcome.

## The fix

- **C++**: `invokeImpl` now catches any exception on the retry path (`catch (...)` instead of `catch (const Exception&)`) and completes the invocation with it — the C++ completion machinery is `exception_ptr`-based end to end, so the caller's future receives the original exception. As a result `retry()` never throws, which is now documented on its declaration. (`noexcept` was deliberately not added in 3.8.x to keep the exported function's contract unchanged; it can come with the #6192 `noexcept` work in 3.9.)
- **C# and Java**: the completion machinery only accepts Ice exceptions, and wrapping a locally-thrown exception in `UnknownException` would misuse its meaning ("the server replied with the UnknownException reply status"), so these runtimes get containment instead: the retry task removes itself from the queue in a `finally` block, preserving the requirement that removal happens only after the retry attempt finishes. In C#, the timer's existing catch-all logs the escaped exception; in Java, the task logs it (`catch (Throwable)`, as in #6254) since the executor would swallow it silently. The invocation's `Task`/`CompletableFuture` still doesn't complete in this case — generalizing the C#/Java exception plumbing so it can, mirroring the C++ model, is #6294 (3.9).

One residual in C++: an observer that throws from `failed()` *during* the exceptional completion — a second contract violation — can still escape; that's inherent until observers are `noexcept` (3.9, #6192).

No changelog entry: defense-in-depth for contract-violating callbacks and allocation failure. No tests: exercising the path requires a deliberately-throwing observer, and a regression would manifest as a suite timeout rather than a clean failure.

Verified with the `Ice/retry` and `Ice/ami` suites in C++, C#, and Java. JavaScript is unaffected: its runtime has no instrumentation observers on the retry path.

Fixes #6111.

### DIFF
--- a/cpp/include/Ice/OutgoingAsync.h
+++ b/cpp/include/Ice/OutgoingAsync.h
@@ -155,7 +155,11 @@ namespace IceInternal
         bool exception(std::exception_ptr) override;
 
         void retryException();
+
+        // Retries the invocation; when the retry attempt fails, completes the invocation with the exception. This
+        // function never throws.
         void retry();
+
         void abort(std::exception_ptr);
 
         std::shared_ptr<ProxyOutgoingAsyncBase> shared_from_this()
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/RetryTask.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/RetryTask.java
@@ -14,13 +14,18 @@ class RetryTask implements Runnable, CancellationHandler {
     @Override
     public void run() {
         if (cancel()) {
-            _outAsync.retry();
-
-            // NOTE: this must be called last, destroy() blocks until all task
-            // are removed to prevent the client thread pool to be destroyed
-            // (we still need the client thread pool at this point to call
-            // exception callbacks with CommunicatorDestroyedException).
-            _queue.remove(this);
+            try {
+                _outAsync.retry();
+            } catch (Throwable ex) {
+                // The scheduled executor silently swallows exceptions thrown by its tasks.
+                _instance.initializationData().logger.error("unexpected exception from retry task:\n" + Ex.toString(ex));
+            } finally {
+                // NOTE: this must be called last, destroy() blocks until all task
+                // are removed to prevent the client thread pool to be destroyed
+                // (we still need the client thread pool at this point to call
+                // exception callbacks with CommunicatorDestroyedException).
+                _queue.remove(this);
+            }
         }
     }
 
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/RetryTask.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/RetryTask.java
@@ -18,7 +18,8 @@ class RetryTask implements Runnable, CancellationHandler {
                 _outAsync.retry();
             } catch (Throwable ex) {
                 // The scheduled executor silently swallows exceptions thrown by its tasks.
-                _instance.initializationData().logger.error("unexpected exception from retry task:\n" + Ex.toString(ex));
+                String msg = "unexpected exception from retry task:\n" + Ex.toString(ex);
+                _instance.initializationData().logger.error(msg);
             } finally {
                 // NOTE: this must be called last, destroy() blocks until all task
                 // are removed to prevent the client thread pool to be destroyed
--- a/cpp/src/Ice/OutgoingAsync.cpp
+++ b/cpp/src/Ice/OutgoingAsync.cpp
@@ -481,7 +481,7 @@ ProxyOutgoingAsyncBase::invokeImpl(bool userThread)
             }
         }
     }
-    catch (const Exception&)
+    catch (...)
     {
         // If called from the user thread we re-throw, the exception will be caught by the caller and handled using
         // abort.
--- a/csharp/src/Ice/Internal/RetryQueue.cs
+++ b/csharp/src/Ice/Internal/RetryQueue.cs
@@ -15,15 +15,20 @@ public class RetryTask : TimerTask, CancellationHandler
 
     public void runTimerTask()
     {
-        _outAsync.retry();
-
-        //
-        // NOTE: this must be called last, destroy() blocks until all task
-        // are removed to prevent the client thread pool to be destroyed
-        // (we still need the client thread pool at this point to call
-        // exception callbacks with CommunicatorDestroyedException).
-        //
-        _retryQueue.remove(this);
+        try
+        {
+            _outAsync.retry();
+        }
+        finally
+        {
+            //
+            // NOTE: this must be called last, destroy() blocks until all task
+            // are removed to prevent the client thread pool to be destroyed
+            // (we still need the client thread pool at this point to call
+            // exception callbacks with CommunicatorDestroyedException).
+            //
+            _retryQueue.remove(this);
+        }
     }
 
     public void asyncRequestCanceled(OutgoingAsyncBase outAsync, Ice.LocalException ex)


cpp/include/Ice/OutgoingAsync.h
```diff
@@ -155,7 +155,11 @@ namespace IceInternal
         bool exception(std::exception_ptr) override;
 
         void retryException();
+
+        // Retries the invocation; when the retry attempt fails, completes the invocation with the exception. This
+        // function never throws.
```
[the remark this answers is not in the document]
InsertCreativityHere: Yes, it is still theoretically possible for the exception handling itself to throw (`exceptionImpl` and `invokeExceptionAsync`), but it's not 100% clear under what exact conditions.

InsertCreativityHere (approved):
Looks good to me!

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/RetryTask.java
```diff
@@ -14,13 +14,18 @@ class RetryTask implements Runnable, CancellationHandler {
...
-            // NOTE: this must be called last, destroy() blocks until all task
-            // are removed to prevent the client thread pool to be destroyed
-            // (we still need the client thread pool at this point to call
-            // exception callbacks with CommunicatorDestroyedException).
-            _queue.remove(this);
+            try {
+                _outAsync.retry();
+            } catch (Throwable ex) {
+                // The scheduled executor silently swallows exceptions thrown by its tasks.
+                _instance.initializationData().logger.error("unexpected exception from retry task:\n" + Ex.toString(ex));
```
InsertCreativityHere: This line is over the 120 column limit. Maybe something like this to split it up?
```suggestion
                String msg = "unexpected exception from retry task:\n" + Ex.toString(ex);
                _instance.initializationData().logger.error(msg);
```

InsertCreativityHere (approved):

pepone (approved):
Re the open question on `OutgoingAsync.h:160` — the conditions are concrete, and one of the two is reachable:

**`exceptionImpl` — reachable.** It calls `_childObserver.failed()` / `_observer.failed()` / `detach()` (`OutgoingAsync.cpp:231,235`), and `ObserverHelperT` forwards straight to the application observer, which isn't `noexcept` (`Instrumentation.h:107-120`). Those calls run *inside* the new `catch (...)`, so a throw escapes `invokeImpl` → escapes `retry()` → `RetryTask::runTimerTask` (`RetryQueue.cpp:24-34`) never reaches `remove()` → `destroy()` waits forever (`RetryQueue.cpp:118`). The trigger is the same buggy observer this PR is defending against: one that throws from `retried()` throws from `failed()` too. #6111 calls out `_childObserver.failed()`/`detach()` on this path explicitly.

**`invokeExceptionAsync` — not reachable here.** It can only propagate `CommunicatorDestroyedException` from the client thread pool, and `Instance::destroy` drains the retry queue (`Instance.cpp:1662-1665`) before destroying that pool (`:1707-1710`).

So the containment gap is a single, well-defined path. Four lines in `RetryQueue.cpp` close it and give all three mappings the same shape:

```cpp
try
{
    _outAsync->retry(); // Retry again the invocation.
}
catch (...)
{
    _queue->remove(shared_from_this());
    throw; // Logged by the timer thread.
}
_queue->remove(shared_from_this());
```

Safe with respect to `assert(_requests.find(task) != _requests.end())` (`RetryQueue.cpp:125`): a running one-shot task is already out of the timer's task map (`Timer.cpp:94-105`), so `Timer::cancel` fails for it and neither `cancel()` nor `destroy()` can erase it first; `add()` holds the queue mutex across schedule+insert, so there's no zero-delay race. With that in, the `retry()` doc comment becomes true as written.

The rest of the change is right: `handleException(std::exception_ptr) noexcept` (`OutgoingAsync.h:56`) plus `getExceptionId`'s `what()`/`"unknown"` fallback make completing with a non-Ice exception safe, the user-thread path is unchanged, C#'s timer logs the escape (`Timer.cs:232`), and Java's executor really does swallow it (`Instance.Timer.afterExecute` ignores its `Throwable`). Agree with no tests — the C#/Java failure mode is a stranded future, which a test can only observe as a timeout.

Separately, not for this PR: all three timers call the *thread* observer before entering the task, outside any guard (`Instance.cpp:270-277`, `Timer.cs:203-215`, `Instance.java:44-48`). A throw there strands the retry task the same way — and in Java kills the pool worker — regardless of any guard inside `RetryTask`. Worth a note on #6111.

bernardnormier:
Taking a different disposition on the `exceptionImpl` path: it's deliberate, not a gap. #6292 documents that observer implementations must not throw — the Ice runtime calls them from contexts where exceptions cannot be handled. `retry()`'s contract is written against conforming observers: the widened `catch (...)` contains everything the *machinery* can throw, plus the single misbehaving callback (`retried()`) that #6111 describes. The scenario you traced requires the observer to violate that contract twice — throw from `retried()`, then throw again from `failed()` while the first violation is being handled. Defending against that leads to the regress your own follow-up note illustrates: the thread observer is called by all three timers outside any guard, so a contract-violating observer can strand a retry task (and kill a Java pool worker) no matter what `RetryTask` does. The contract is the containment; I'd rather keep the code shape simple than guard one of the N unguardable call sites.

Your `invokeExceptionAsync` verification (unreachable — the retry queue drains before the client pool is destroyed) is a good confirmation that the machinery side is airtight. The thread-observer note is now recorded as #6331, linked to the 3.9 observer no-throw enforcement discussion (#6192).

Austin's line-length suggestion is applied in 5c6e8265b7.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/RetryTask.java
bernardnormier: Applied in 5c6e8265b7, thanks.

PR #6296, opened 2026-07-30
Fix race between concurrent ObjectAdapter destroy calls in Java

### LINKED ISSUE #6271: Concurrent ObjectAdapter.destroy calls can resurrect a destroyed adapter in Java
Found by @pepone while reviewing #6253; pre-existing and out of that PR's scope.

`ObjectAdapter.deactivate()` returns early only when `_state > StateDeactivating`, and then writes `_state = StateDeactivated` unconditionally (`ObjectAdapter.java:254`). Two concurrent `destroy()` calls on an active adapter can interleave so that the second call's `deactivate()` overwrites `StateDestroyed` after the first call completed, resurrecting a fully destroyed adapter — the second `destroy()` then NPEs on `_instance.outgoingConnectionFactory()` (or on any other field the completed destroy nulled). The window is wide when `updateLocatorRegistry` is involved, since deactivation blocks on the locator during it.

bernardnormier:
`ObjectAdapter.deactivate()` only waited out `StateActivating`, so a second concurrent `deactivate()` entered the deactivation body while the first was still blocked in `updateLocatorRegistry`, and later unconditionally wrote `StateDeactivated` — overwriting `StateDestroyed` if the first thread's `destroy()` had already completed. The second `destroy()` then failed with a `NullPointerException` on the fields the completed destroy had cleared, and its cleanup path restored `StateDeactivated`, leaving the destroyed adapter destroyable again.

The wait loop now also covers `StateDeactivating`, matching the C++ and C# implementations, so only one thread ever runs the deactivation body; a concurrent caller wakes once the state has advanced past `StateDeactivating` and returns through the existing guard. Also added the same assertion before the final state write that C++ and C# have.

Fixes #6271

### DIFF
--- a/changelog.d/java/6271.md
+++ b/changelog.d/java/6271.md
@@ -1,0 +1,3 @@
+- Fixed a race condition in `ObjectAdapter`: when two threads called `destroy` (or `deactivate`) concurrently on the
+  same object adapter, the second call could mark a fully destroyed adapter as merely deactivated, and a subsequent
+  `destroy` call would then fail with a `NullPointerException`.
--- a/changelog.d/java/6271.md
+++ b/changelog.d/java/6271.md
@@ -1,3 +1,3 @@
-- Fixed a race condition in `ObjectAdapter`: when two threads called `destroy` (or `deactivate`) concurrently on the
-  same object adapter, the second call could mark a fully destroyed adapter as merely deactivated, and a subsequent
-  `destroy` call would then fail with a `NullPointerException`.
+- Fixed a race condition in `ObjectAdapter`: when `destroy` raced with another `destroy` or with `deactivate` on
+  the same object adapter, the destroyed adapter could be marked as merely deactivated, and a subsequent `destroy`
+  call would then fail with a `NullPointerException`.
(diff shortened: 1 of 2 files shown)


changelog.d/java/6271.md
```diff
@@ -0,0 +1,3 @@
+- Fixed a race condition in `ObjectAdapter`: when two threads called `destroy` (or `deactivate`) concurrently on the
+  same object adapter, the second call could mark a fully destroyed adapter as merely deactivated, and a subsequent
+  `destroy` call would then fail with a `NullPointerException`.
```
pepone: Two concurrent `deactivate` calls on their own are harmless — both just write `StateDeactivated`, and the locator update and `factory.destroy()` are idempotent. The resurrection needs a `destroy` on at least one side.

```suggestion
- Fixed a race condition in `ObjectAdapter`: when `destroy` raced with another `destroy` or with `deactivate` on
  the same object adapter, the destroyed adapter could be marked as merely deactivated, and a subsequent `destroy`
  call would then fail with a `NullPointerException`.
```

pepone (approved):
Approve. Matches the C++ and C# wait loop, so only one thread can enter the deactivation body. The assert can't fire either: `activate` and `hold` both go through `checkForDeactivation`, and the `finally` rollback in `destroy` requires `StateDestroying`, which is unreachable while another thread is still deactivating.

One nit on the changelog below.

changelog.d/java/6271.md
bernardnormier: Right — two concurrent deactivates are harmless on their own. Applied your wording verbatim in ae8b4b37fa.

PR #6297, opened 2026-07-30
Replace Glacier2's duplicate-category assert with a runtime check

externl:
Follow-up to [pepone's review of #6233](https://github.com/zeroc-ice/ice/pull/6233#pullrequestreview-4810201595). Glacier2 guards the uniqueness of each session's randomly generated server proxy category with only an `assert`, which optimized builds compile out: a duplicate silently corrupts `_routersByCategory`, callbacks for the two colliding sessions misroute, and destroying either session erases the shared entry.

With #6233 a duplicate is essentially unreachable, so this is defense in depth: a duplicate now fails cleanly at session creation.

- Move the category registration ahead of the connection registration and check the insert result: a duplicate destroys the router (as the `_destroy` branch does) and throws, leaving no half-registered state.
- Throw `UnknownException` rather than `CannotCreateSessionException`: a duplicate is an internal error, not a refusal, and the default `LoggerMiddleware` logs Unknown replies so the operator sees it.
- Reuse `defaultSessionDestroyExceptionHandler()` in the `_destroy` branch in place of an identical inline lambda.

Fixes zeroc-ice/ice-audit#127.

### DIFF
--- a/cpp/src/Glacier2/SessionRouterI.cpp
+++ b/cpp/src/Glacier2/SessionRouterI.cpp
@@ -1129,21 +1129,29 @@ SessionRouterI::finishCreateSession(const ConnectionPtr& connection, const share
 
     if (_destroy)
     {
-        router->destroy([self = shared_from_this()](exception_ptr e) { self->sessionDestroyException(e); });
+        router->destroy(defaultSessionDestroyExceptionHandler());
 
         throw CannotCreateSessionException("router is shutting down");
     }
 
-    _routersByConnectionHint = _routersByConnection.insert(_routersByConnectionHint, {connection, router});
-
     if (_instance->serverObjectAdapter())
     {
+        // The category is randomly generated, so a collision with an existing session's category should never
+        // occur. Check anyway, and register the category before the connection so that a failure leaves nothing
+        // to undo.
         string category = router->serverProxy()->ice_getIdentity().category;
         assert(!category.empty());
-        auto rc = _routersByCategory.insert({category, router});
-        assert(rc.second);
-        _routersByCategoryHint = rc.first;
+        auto [pos, inserted] = _routersByCategory.insert({category, router});
+        if (!inserted)
+        {
+            router->destroy(defaultSessionDestroyExceptionHandler());
+
+            throw CannotCreateSessionException("duplicate server proxy category");
+        }
+        _routersByCategoryHint = pos;
     }
+
+    _routersByConnectionHint = _routersByConnection.insert(_routersByConnectionHint, {connection, router});
 
     connection->setCloseCallback(
         [self = shared_from_this()](const ConnectionPtr& c)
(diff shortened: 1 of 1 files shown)


cpp/src/Glacier2/SessionRouterI.cpp
```diff
@@ -1129,22 +1129,30 @@ SessionRouterI::finishCreateSession(const ConnectionPtr& connection, const share
...
-        auto rc = _routersByCategory.insert({category, router});
-        assert(rc.second);
-        _routersByCategoryHint = rc.first;
+        auto [pos, inserted] = _routersByCategory.insert({category, router});
+        if (!inserted)
+        {
+            router->destroy(defaultSessionDestroyExceptionHandler());
+
+            throw CannotCreateSessionException("duplicate server proxy category");
+        }
```
pepone, line 1152: Worth logging this. A duplicate now means the CSPRNG is broken — the same failure class #6233 addressed — but the only signal goes to the client, who can't act on it, while the operator who can sees nothing. `unexpectedCreateSessionException` already follows the log-the-detail/reply-generically pattern; unguarded by the trace level, since this is a can't-happen.

```suggestion
        if (!inserted)
        {
            Error out(_instance->logger());
            out << "duplicate server proxy category: the random number generator is not producing unique values";

            router->destroy(defaultSessionDestroyExceptionHandler());

            throw CannotCreateSessionException("duplicate server proxy category");
        }
```

pepone (approved):
LGTM. Verified the duplicate path leaves nothing behind: `_pending` is erased first, the failed insert keeps the incumbent entry, and neither the connection map nor the close callback has been touched yet. `RouterI::destroy` covers the `SessionControl` servant — `_control` is only created under `if (createSession)`, and a null session proxy short-circuits to `exception()` rather than `sessionCreated`, so `_control != nullptr` implies `_session != nullopt`. The reorder is unobservable: the whole function is under `_mutex` and every reader of both maps takes it. `sessionCreated` handles the new throw with exactly one reply and still runs the pending callbacks.

Agree on no test and no changelog fragment for an unreachable path.

cpp/src/Glacier2/SessionRouterI.cpp
externl: I don't think this is really worthwhile.

bernardnormier: I agree the operator should see this — a duplicate means Glacier2's own invariant is broken. But we get the logging for free by picking the right exception: `UnknownException` is a public `DispatchException` subclass whose message marshals verbatim to the client, and the `LoggerMiddleware` installed by default (`Ice.Warn.Dispatch`, default 1) logs Unknown* replies unconditionally, with the exception details. And since this is an internal bug — not a refusal of the caller's request — `CannotCreateSessionException` is the wrong type anyway: that's the documented refusal exception (what a `PermissionsVerifier` rejection produces), which client code can legitimately react to.

So the minimal change:

```cpp
throw UnknownException{__FILE__, __LINE__, "duplicate client category"};
```

("server proxy category" is internal vocabulary — the caller knows this string as the client category, per `getCategoryForClient` in `Glacier2/Router.ice`.)

bernardnormier (approved):

cpp/src/Glacier2/SessionRouterI.cpp
externl: Done in 4ad0376d99e — `UnknownException{__FILE__, __LINE__, "duplicate client category"}`.

PR #6298, opened 2026-07-30
Deliver messages to remote loggers when Logger admin retention is 0

### LINKED ISSUE #6259: Setting Ice.Admin.Logger.KeepLogs/KeepTraces to 0 also stops live delivery to attached remote loggers in C++ and Java
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced while reviewing #6250 and confirmed by re-reading `log()` in all three implementations on `main` (`516cb08400`). It was not reproduced at runtime, so the severity is a static-review estimate.

## Summary

`LoggerAdminI::log` does two independent things: it appends the message to the retention queue, and it computes the list of attached remote loggers that want to receive the message. In C++ and Java the second step is nested **inside** the `if` that guards the first, so when the retention limit for a category is `0`, messages of that category are not merely left out of the queue — they are never delivered to attached remote loggers either.

| Mapping | Retention guard | Remote-logger fan-out |
| --- | --- | --- |
| C++ | `cpp/src/Ice/LoggerAdminI.cpp:456-530` | `:516-529` — **inside** the guard |
| Java | `java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerAdminI.java:171-229` | `:215-228` — **inside** the guard |
| C# | `csharp/src/Ice/Internal/LoggerAdminI.cs:200-257` | `:261-274` — **outside** the guard |

`log()` returns that destination list, and the logger wrapper only enqueues a remote call when the list is non-empty — `cpp/src/Ice/LoggerAdminI.cpp:647-660` and `java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/LoggerAdminLoggerI.java:168-184`. So with `Ice.Admin.Logger.KeepLogs=0`, a C++ or Java application delivers no print/warning/error message to any attached `RemoteLogger`, even one that explicitly asked for those message types. With `Ice.Admin.Logger.KeepTraces=0`, the same happens to trace messages. C# keeps delivering in both cases.

## Which behavior is correct

`slice/Ice/RemoteLogger.ice` documents `attachRemoteLogger`'s `messageTypes`/`traceCategories` as the filter on what the remote logger receives, and `messageMax` as bounding only the historical messages passed to `RemoteLogger::init`. Retention is not described as a delivery filter, which points at C# being right and the C++/Java nesting being accidental.

If the intended meaning of `KeepLogs=0` is instead "turn this category off entirely", then C# is the one to change and the property should say so. Either way the three mappings should not disagree.

## Trigger and impact

Both properties default to `100`, so this only affects a deployment that explicitly sets one of them to `0` — plausible for a "stream traces live, don't retain them" configuration. The failure is silent: no error, the remote logger simply never receives those messages.

## Suggested fix

Move the fan-out loop out of the retention `if` in C++ and Java, matching the C# structure. Note that no test in any mapping covers an attached remote logger with either retention property set to `0`.

bernardnormier:
In the C++ and Java `LoggerAdminI::log`, the loop that computes which attached remote loggers should receive a message was nested inside the `if` that guards the retention queue update. With `Ice.Admin.Logger.KeepLogs=0` or `Ice.Admin.Logger.KeepTraces=0`, messages of the corresponding category were therefore never delivered to attached remote loggers — silently. The retention properties are only supposed to bound what the facet keeps for `getLog` and `RemoteLogger::init`; C# already had the loop outside the guard and behaved correctly.

This PR moves the fan-out loop out of the retention guard in C++ and Java, matching the C# structure, and adds a retention=0 case to the admin test in all three mappings (C# included, to guard against future divergence): with both properties set to `0`, `getLog` returns nothing but an attached remote logger still receives print and trace messages live. I verified the new test fails against the unfixed C++ library.

Fixes #6259

### DIFF
--- a/changelog.d/cpp/6259.md
+++ b/changelog.d/cpp/6259.md
@@ -1,0 +1,5 @@
+- Fixed the delivery of log messages to remote loggers attached to the Logger admin facet when
+  `Ice.Admin.Logger.KeepLogs` or `Ice.Admin.Logger.KeepTraces` is set to `0` (their default value is `100`). These
+  properties control only how many messages the facet retains for `getLog` and `RemoteLogger::init`; previously,
+  setting one of them to `0` also prevented the live delivery of messages of the corresponding category to attached
+  remote loggers.
--- a/changelog.d/java/6259.md
+++ b/changelog.d/java/6259.md
@@ -1,0 +1,5 @@
+- Fixed the delivery of log messages to remote loggers attached to the Logger admin facet when
+  `Ice.Admin.Logger.KeepLogs` or `Ice.Admin.Logger.KeepTraces` is set to `0` (their default value is `100`). These
+  properties control only how many messages the facet retains for `getLog` and `RemoteLogger.init`; previously,
+  setting one of them to `0` also prevented the live delivery of messages of the corresponding category to attached
+  remote loggers.
--- a/cpp/src/Ice/LoggerAdminI.cpp
+++ b/cpp/src/Ice/LoggerAdminI.cpp
@@ -510,21 +510,21 @@ namespace
                     }
                 }
             }
+        }
 
-            //
-            // Queue updated, now find which remote loggers want this message
-            //
-            for (const auto& q : _remoteLoggerMap)
+        //
+        // Queue updated, now find which remote loggers want this message
+        //
+        for (const auto& q : _remoteLoggerMap)
+        {
+            const Filters& filters = q.second;
+
+            if (filters.messageTypes.empty() || filters.messageTypes.count(logMessage.type) != 0)
             {
-                const Filters& filters = q.second;
-
-                if (filters.messageTypes.empty() || filters.messageTypes.count(logMessage.type) != 0)
+                if (logMessage.type != LogMessageType::TraceMessage || filters.traceCategories.empty() ||
+                    filters.traceCategories.count(logMessage.traceCategory) != 0)
                 {
-                    if (logMessage.type != LogMessageType::TraceMessage || filters.traceCategories.empty() ||
-                        filters.traceCategories.count(logMessage.traceCategory) != 0)
-                    {
-                        remoteLoggers.push_back(q.first);
-                    }
+                    remoteLoggers.push_back(q.first);
                 }
             }
         }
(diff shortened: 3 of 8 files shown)


pepone (approved):
Verified the diagnosis and the fix against the source, and red/green'd all three mappings locally.

**C# is the right reference here.** The issue left open whether `KeepLogs=0` was meant as "turn the category off entirely". It wasn't: `LoggerAdminLoggerI::print/trace/warning/error` call `_localLogger` unconditionally before ever consulting retention, so these properties never gated local output either. Together with the Slice docs scoping `messageMax` to `init` only, the C++/Java nesting is clearly accidental.

**Red/green** — worktree at PR HEAD, each mapping built, then the library fix reverted in isolation and the test re-run:

| | with fix | fix reverted |
|---|---|---|
| C++ `Ice/admin` | passes, 0.66s | aborts at 21s (20s wait timeout, then assert) |
| Java `Ice/admin` | passes, 0.60s | hangs at `testing logger facet...` |
| C# `Ice/admin` | passes, 0.49s | n/a |

**Completeness**: C++, C#, and Java are the only `LoggerAdminI` implementations, and the properties are declared for exactly those three (`config/PropertyNames.xml:61-62`). Swift/Python/Ruby/PHP/MATLAB drive the C++ core and inherit the fix; JS has none.

Lock scope is preserved in both mappings (C++ function-scope `lock_guard`, Java method stays `synchronized`), clang-format-19 is clean, and omitting a C# changelog fragment is right since only its test changed.

One non-blocking suggestion: the Java and C# `RemoteLoggerI.wait(int)` helpers wait unbounded, unlike C++ which bounds at 20s and returns `false`. That's why the Java red pass hung rather than failed — a future regression here would burn the CI job timeout with no diagnostic instead of failing fast. Pre-existing helper limitation, but this PR adds the first Java assertion whose only failure mode is a hang. Worth bounding those two, here or separately.

bernardnormier:
@pepone Thanks for the red/green verification — that table is exactly the confirmation this fix needed. Took the wait-bounding suggestion here rather than separately: 4bc9e661c0 gives the C# and Java `wait` helpers the C++ shape (20-second bound, missing-count diagnostic on stderr, `false` return with every call site asserting the result), so a future regression fails fast instead of hanging the CI job. Both `Ice/admin` suites pass sub-second with the new helpers.

PR #6300, opened 2026-07-30
Fix Ruby getImplicitContext crash under the default configuration

### LINKED ISSUE #6104: Ruby default `getImplicitContext` returns a null-backed wrapper whose methods crash (sibling of #5967)
> [!NOTE]

## Summary

`Communicator#getImplicitContext` unconditionally wraps the C++ result, but under the default `Ice.ImplicitContext=None` the C++ `getImplicitContext()` returns a **null** shared pointer. The Ruby wrapper stores the null pointer and every method (`getContext`, `setContext`, `get`, `put`, `containsKey`, `remove`) immediately dereferences it → segfault. This is the Ruby sibling of #5967 (same crash, MATLAB); Python already guards this exact case and returns `nil`.

**Files:** `ruby/src/IceRuby/Communicator.cpp:386-397`
```cpp
Ice::ImplicitContextPtr implicitContext = p->getImplicitContext();
return createImplicitContext(implicitContext);          // no null check
```
`ruby/src/IceRuby/ImplicitContext.cpp:29-35` etc. deref `p->...` with no guard; C++ default returns null at `cpp/src/Ice/Instance.cpp:1332-1337`. Compare the Python guard at `python/modules/IcePy/Communicator.cpp:1086-1087` (`implicitContext ? createImplicitContext(...) : Py_None`).

## Fix

Return `Qnil` when the native pointer is null (mirroring Python).

**Severity:** Medium (deterministic interpreter crash, but only when an app calls `getImplicitContext` without configuring `Shared`/`PerThread`).

bernardnormier:
With the default `Ice.ImplicitContext=None`, the C++ `getImplicitContext()` returns a null `shared_ptr`. The Ruby wrapper stored the null pointer unconditionally, and every method on the returned `ImplicitContext` object (`getContext`, `setContext`, `get`, `put`, `containsKey`, `remove`) dereferenced it, crashing the Ruby interpreter. This is the Ruby sibling of the MATLAB crash fixed by #5975.

`Communicator#getImplicitContext` now returns `nil` when the communicator has no implicit context, matching the existing Python guard.

Following the precedent of #5975, the implicit-context section of the `Ice/operations` test now includes a default-configuration case asserting that `getImplicitContext` returns `nil`. The test fails against the unfixed extension.

Fixes #6104

### DIFF
--- a/changelog.d/ruby/6104.md
+++ b/changelog.d/ruby/6104.md
@@ -1,0 +1,3 @@
+- Fixed a crash in `Communicator#getImplicitContext`: with the default `Ice.ImplicitContext=None` configuration, this
+  method returned a broken `ImplicitContext` object whose methods crashed the Ruby interpreter. It now returns `nil`
+  when the communicator has no implicit context.
--- a/ruby/src/IceRuby/Communicator.cpp
+++ b/ruby/src/IceRuby/Communicator.cpp
@@ -390,7 +390,7 @@ IceRuby_Communicator_getImplicitContext(VALUE self)
     {
         Ice::CommunicatorPtr p = getCommunicator(self);
         Ice::ImplicitContextPtr implicitContext = p->getImplicitContext();
-        return createImplicitContext(implicitContext);
+        return implicitContext ? createImplicitContext(implicitContext) : Qnil;
     }
     ICE_RUBY_CATCH
     return Qnil;
--- a/ruby/test/Ice/operations/Twoways.rb
+++ b/ruby/test/Ice/operations/Twoways.rb
@@ -1227,6 +1227,14 @@ def twoways(helper, communicator, p)
     #
     # Test implicit context propagation
     #
+
+    # Under the default Ice.ImplicitContext=None, there is no implicit context.
+    initData = Ice::InitializationData.new
+    initData.properties = communicator.getProperties().clone()
+    ic = Ice::initialize(initData)
+    test(ic.getImplicitContext().nil?)
+    ic.destroy()
+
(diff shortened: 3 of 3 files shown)


pepone (approved):
Approve.

Verified locally rather than by reading — built the extension from this branch (Ruby 4.0.6, macOS arm64):

- Pre-fix, under the default config `getImplicitContext()` returns an `Ice::ImplicitContextI` wrapping a null `shared_ptr`, and `getContext()` on it is a hard `[BUG] Segmentation fault at 0x0`.
- Post-fix it returns `nil`. `Ice.ImplicitContext=Shared` behaves as before.
- The new assertion flips `false` → `true` and fails cleanly (no segfault) against the unfixed extension, so it's a genuine red→green test. `ruby/Ice/operations` passes.

`createImplicitContext` has exactly one caller, so no other path can produce a null-backed wrapper — the unguarded derefs in `ImplicitContext.cpp` are unreachable and don't need guards of their own. PHP's `getImplicitContext` is `not implemented`, so no sibling bug there.

One nit, non-blocking: the fresh communicator in the test reads as isolation but isn't — `getProperties().clone()` inherits `Ice.ImplicitContext` if it's ever set on the driver, so the assertion would fail the same way. `test(communicator.getImplicitContext().nil?)` on the communicator already in scope is equivalent. Fine to keep for symmetry with the MATLAB test.

bernardnormier:
@pepone Thanks for the red/green verification. On the nit: agreed the clone doesn't actually isolate — keeping the shape anyway for symmetry with the MATLAB test, as you suggested. No code change.

Issue #6301, opened 2026-07-30
IcePy: ConnectionInfo doesn't expose connectionId, and sequence<float> marshaling skips the float range check


bernardnormier:
Two Python siblings of Ruby findings fixed under #6116.

## 1. `ConnectionInfo` doesn't expose `connectionId`

C++ `Ice::ConnectionInfo` has a `connectionId` member, and C++, C#, Java (and Ruby, once #6116 is fixed) all expose it. IcePy's `ConnectionInfo` type never sets or declares it (`python/modules/IcePy/ConnectionInfo.cpp`), so a Python application has no way to retrieve the connection ID from `Connection.getInfo()`.

## 2. `sequence<float>` marshaling skips the float range check

The scalar float marshaling path validates the value (`python/modules/IcePy/Types.cpp:703-708`): a finite double outside the `float` range raises `ValueError: invalid value for float type: out of range`, while infinity and NaN are deliberately allowed. The optimized `sequence<float>` path (`Types.cpp:1779-1806`) performs `static_cast<float>(PyFloat_AsDouble(item))` with no such check, so a large finite double in a `sequence<float>` element silently marshals as `Infinity`. The finite-but-out-of-range `double` → `float` conversion is also formally undefined behavior in C++; producing infinity is just what IEEE hardware happens to do.

The fix is to apply the same check as the scalar path to the sequence path, preserving the infinity/NaN pass-through.

InsertCreativityHere:
Both points were fixed by #6373.

PR #6302, opened 2026-07-30
Fix four low-severity issues in the Ruby extension

### LINKED ISSUE #6116: Ruby low-severity batch: communicator/slice-loader map leak, Connection `eql?` without `hash`, `ConnectionInfo.connectionId` always nil, FloatSeq range check
> [!NOTE]

## Summary

Four low-severity Ruby findings.

### 1. Communicator / slice-loader maps never fully erased
`ruby/src/IceRuby/Communicator.cpp:18-25,154-186`: `_communicatorMap` entries survive in every path (the `erase` at `:164` is dead code) and `_sliceLoaderMap` is left behind by block-form initialization, deterministically pinning native communicators/loaders (and a custom slice loader as a GC root) for process lifetime. Note the non-erasure is partly load-bearing for `IceRuby_ObjectPrx_mark`, so a fix must keep GC marking correct.

### 2. `Connection#eql?` without matching `hash`
`ruby/src/IceRuby/Connection.cpp:334-335` defines `==`/`eql?` (native-pointer identity) but no `hash`, so `Object#hash` (identity) is used, violating Ruby's `eql? ⇒ hash` contract; fresh wrappers per `ice_getConnection` (`Proxy.cpp:727`) then miss Hash/Set membership. Proxy and Endpoint both define `hash`; Connection is the only mismatch.

### 3. `ConnectionInfo#connectionId` always nil
`ruby/src/IceRuby/Connection.cpp:349` declares the attr but `createConnectionInfo` never assigns `@connectionId`. Consider removing the vestigial attr (IcePy does not expose it either).

### 4. FloatSeq marshal skips the scalar range check
`ruby/src/IceRuby/Types.cpp:1412-1428` casts `double→float` in the optimized sequence path without the `numeric_limits<float>::max()` check the scalar path has (`:555-559`), so a large finite double in a `sequence<float>` silently marshals as `Infinity`.

## Fix

Erase map entries on communicator destroy (preserving GC marking); add a native-pointer `hash` to Connection; assign or remove `connectionId`; add the float-range check to the sequence path.

**Severity:** Low.

bernardnormier:
One fix per item of #6116:

**1. Communicator / slice-loader map leak — and a hang at exit.** The `_communicatorMap`/`_sliceLoaderMap` entries now get erased in the communicator wrapper's GC free function. They cannot be erased in `destroy()`: `IceRuby_ObjectPrx_mark` looks up (and asserts) the communicator wrapper in the map for every live proxy, so the entry must survive as long as any proxy does. The free function is the correct point — it only runs once nothing marks the wrapper, which implies no proxy for that communicator is alive. This releases the native communicator and slice loader that were previously pinned for the process lifetime, and it fixes a deterministic hang at exit: with a custom `sliceLoader`, the pinned entry was released only during the C++ runtime's static destruction, which runs after the Ruby VM is gone — `~RubySliceLoader` then calls `rb_gc_unregister_address` on a dead VM, segfaults, and Ruby's fatal-signal handler wedges. Any program that doesn't call the Ruby-level `destroy()` was affected, including programs using the documented block form of `Ice::initialize`: the block form destroys the communicator through the C++ API and so never ran the Ruby-level erase. Our test suite always destroys communicators explicitly through `TestHelper`, which is why CI never saw the hang. Also removed the dead lookup before the map insert: a live entry's key (a `shared_ptr`) keeps the C++ communicator alive, so `Ice::initialize` can never return a pointer that is still a key.

**2. `Connection#hash`.** Added, hashing the native pointer — exactly what `eql?` compares — restoring Ruby's `eql?` ⇒ `hash` contract so connections work as `Hash` keys and `Set` elements. Proxy and Endpoint already had `hash`; Connection was the only mismatch. `eql?` also now returns `false` for a non-Connection argument instead of raising `TypeError`: Ruby calls `eql?` with arbitrary keys on a `Hash` collision, so a well-behaved `Hash` key must accept them.

**3. `ConnectionInfo#connectionId`.** Now assigned from the C++ `ConnectionInfo`; it was declared but never set, so it was always `nil`. Assigning (rather than removing the attr, as the issue suggested) is backward compatible and matches C++, C#, and Java, which all expose `connectionId`; Python's missing exposure is tracked in #6301.

**4. `sequence<float>` range check.** The optimized sequence marshaling path now applies the same check as a plain `float` parameter: a finite value outside the `float` range raises `TypeError` instead of silently marshaling as `Infinity` (the finite-out-of-range `double` → `float` conversion is also undefined behavior in C++). Infinity and NaN still pass through, as for plain floats. Both checks now spell the lower bound with `numeric_limits<float>::lowest()`. The same gap exists in IcePy and is tracked in #6301.

Tests: `Ice/objects` now exercises the block form of `Ice::initialize` with a custom slice loader and exits normally — against the unfixed extension, the client hangs at exit. `Ice/info` covers `connectionId` (default and with `ice_connectionId`), `eql?`/`hash`/Hash-membership for two wrappers of the same connection, and `eql?` with a non-Connection argument; `Ice/operations` covers out-of-range `sequence<float>` elements of both signs. All three suites fail against the unfixed extension and pass with the fixes.

The changelog fragment covers items 1–3; item 4 only changes how an out-of-range input fails.

Fixes #6116

### DIFF
--- a/changelog.d/ruby/6116.md
+++ b/changelog.d/ruby/6116.md
@@ -1,0 +1,4 @@
+- Added a `hash` method to `Connection`, consistent with `eql?`. Previously, two `Connection` objects representing
+  the same connection could compare equal but hash differently, so connections could not be used reliably as `Hash`
+  keys or `Set` elements.
+- `ConnectionInfo#connectionId` now returns the ID of the connection. Previously, it was always `nil`.
--- a/changelog.d/ruby/6116.md
+++ b/changelog.d/ruby/6116.md
@@ -1,3 +1,5 @@
+- Fixed a hang at program exit when a communicator is created with a custom `sliceLoader` and destroyed through the
+  block form of `Ice::initialize`, or not destroyed at all.
 - Added a `hash` method to `Connection`, consistent with `eql?`. Previously, two `Connection` objects representing
   the same connection could compare equal but hash differently, so connections could not be used reliably as `Hash`
   keys or `Set` elements.
--- a/ruby/src/IceRuby/Communicator.cpp
+++ b/ruby/src/IceRuby/Communicator.cpp
@@ -28,7 +28,12 @@ namespace
 extern "C" void
 IceRuby_Communicator_free(void* p)
 {
-    delete static_cast<Ice::CommunicatorPtr*>(p);
+    // The wrapper is collected only once no other object (such as a proxy) marks it. Erase the map
+    // entries so that the native communicator and slice loader can be released.
+    auto communicator = static_cast<Ice::CommunicatorPtr*>(p);
+    _communicatorMap.erase(*communicator);
+    _sliceLoaderMap.erase(*communicator);
+    delete communicator;
 }
 
 static const rb_data_type_t IceRuby_CommunicatorType = {
@@ -158,11 +163,6 @@ IceRuby_initialize(int argc, VALUE* argv, VALUE /*self*/)
             &IceRuby_CommunicatorType,
             new Ice::CommunicatorPtr(communicator));
 
-        CommunicatorMap::iterator p = _communicatorMap.find(communicator);
-        if (p != _communicatorMap.end())
-        {
-            _communicatorMap.erase(p);
-        }
         _communicatorMap.insert(CommunicatorMap::value_type(communicator, reinterpret_cast<const VALUE&>(result)));
 
         _sliceLoaderMap[communicator] = sliceLoader;
--- a/ruby/src/IceRuby/Connection.cpp
+++ b/ruby/src/IceRuby/Connection.cpp
@@ -188,13 +188,10 @@ IceRuby_Connection_equals(VALUE self, VALUE other)
 {
     ICE_RUBY_TRY
     {
-        if (NIL_P(other))
+        // eql? returns false for a non-Connection argument: Ruby calls it with arbitrary keys on a Hash collision.
+        if (NIL_P(other) || callRuby(rb_obj_is_kind_of, other, _connectionClass) != Qtrue)
         {
             return Qfalse;
-        }
-        if (callRuby(rb_obj_is_kind_of, other, _connectionClass) != Qtrue)
-        {
-            throw RubyException(rb_eTypeError, "argument must be a connection");
         }
         Ice::ConnectionPtr* p1 = reinterpret_cast<Ice::ConnectionPtr*>(DATA_PTR(self));
         Ice::ConnectionPtr* p2 = reinterpret_cast<Ice::ConnectionPtr*>(DATA_PTR(other));
--- a/ruby/src/IceRuby/Connection.cpp
+++ b/ruby/src/IceRuby/Connection.cpp
@@ -204,6 +204,19 @@ IceRuby_Connection_equals(VALUE self, VALUE other)
     return Qnil;
 }
 
+extern "C" VALUE
+IceRuby_Connection_hash(VALUE self)
+{
+    ICE_RUBY_TRY
+    {
+        Ice::ConnectionPtr* p = reinterpret_cast<Ice::ConnectionPtr*>(DATA_PTR(self));
+        // Hash the native pointer, which is exactly what eql? compares.
+        return INT2FIX(std::hash<Ice::ConnectionPtr>{}(*p));
+    }
+    ICE_RUBY_CATCH
+    return Qnil;
+}
+
 // ConnectionInfo
 
 extern "C" void
@@ -300,6 +313,7 @@ IceRuby::createConnectionInfo(const Ice::ConnectionInfoPtr& p)
     rb_ivar_set(info, rb_intern("@underlying"), createConnectionInfo(p->underlying));
     rb_ivar_set(info, rb_intern("@incoming"), p->incoming ? Qtrue : Qfalse);
     rb_ivar_set(info, rb_intern("@adapterName"), createString(p->adapterName));
+    rb_ivar_set(info, rb_intern("@connectionId"), createString(p->connectionId));
     return info;
 }
 
@@ -333,6 +347,7 @@ IceRuby::initConnection(VALUE iceModule)
     rb_define_method(_connectionClass, "inspect", CAST_METHOD(IceRuby_Connection_toString), 0);
     rb_define_method(_connectionClass, "==", CAST_METHOD(IceRuby_Connection_equals), 1);
     rb_define_method(_connectionClass, "eql?", CAST_METHOD(IceRuby_Connection_equals), 1);
+    rb_define_method(_connectionClass, "hash", CAST_METHOD(IceRuby_Connection_hash), 0);
 
     //
     // ConnectionInfo.
--- a/ruby/src/IceRuby/Types.cpp
+++ b/ruby/src/IceRuby/Types.cpp
@@ -553,7 +553,7 @@ IceRuby::PrimitiveInfo::marshal(VALUE p, Ice::OutputStream* os, ValueMap*, bool)
             }
             assert(TYPE(val) == T_FLOAT);
             double d = static_cast<double>(RFLOAT_VALUE(val));
-            if (isfinite(d) && (d > numeric_limits<float>::max() || d < -numeric_limits<float>::max()))
+            if (isfinite(d) && (d > numeric_limits<float>::max() || d < numeric_limits<float>::lowest()))
             {
                 throw RubyException(rb_eTypeError, "value is out of range for a float");
             }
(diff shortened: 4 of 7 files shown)


externl (approved):
Looks good!

pepone (approved):
Verified locally: built the branch and ran the Ruby tests against the C++ servers. Both new tests are genuine red→green (they fail with `ruby/src` reverted to base) and the full Ruby suite is 19/19.

The four fixes look right: `connectionId` is set for every info subclass including `underlying`; `Connection#hash` matches what `equals` compares, and a live wrapper holds a `shared_ptr` so an address can't be recycled under it; the float check correctly lets ±Inf/NaN through and rejects only finite out-of-range values; and the removed pre-insert `find`/`erase` was indeed dead, since the map key keeps the C++ communicator alive.

One thing worth adding to the description: item 1 fixes a crash, not just a leak. On main, a program that passes a custom `sliceLoader` in `InitializationData` and never calls `destroy()` segfaults at exit — `_sliceLoaderMap`'s static destructor runs after the Ruby VM is torn down, so `~RubySliceLoader` calls `rb_gc_unregister_address()` on a dead VM:

```
exit → __cxa_finalize_ranges
  → std::map<CommunicatorPtr, SliceLoaderPtr>::~map()      # Communicator.cpp:25
    → Ice::CompositeSliceLoader::~CompositeSliceLoader()
      → IceRuby::RubySliceLoader::~RubySliceLoader()       # RubySliceLoader.cpp:16
        → SIGSEGV in libruby (then hangs in Ruby's crash reporter)
```

| custom slice loader, communicator… | main | this PR |
| --- | --- | --- |
| dropped and GC'd, never destroyed | SIGSEGV at exit | clean |
| held until exit, never destroyed | SIGSEGV at exit | clean |
| `destroy()` called | clean | clean |

Our own tests always destroy the communicator, which is why CI never hit it. If you do want a test for item 1 after all, it's observable without relying on GC timing of the native side: `ObjectSpace.each_object(MySliceLoader).count` is 5 on main and 0 with the fix.

I also checked the free-function safety, since erasing `_sliceLoaderMap` from a `RUBY_TYPED_FREE_IMMEDIATELY` dfree does end up calling `rb_gc_unregister_address()` during the sweep. That's a list unlink plus `xfree` with the GVL held and marking already finished — survives `GC.stress` with `GC.verify_internal_consistency`. No threading concern either: the maps are only touched from Ruby threads, since `lookupSliceLoader` is reached only from `unmarshalResults`/`unmarshalException`.

Separate and pre-existing, not from this PR: `_communicatorMap` stores a raw `VALUE` that isn't updated when GC moves the wrapper. After `GC.verify_compaction_references(expand_heap: true, toward: :empty)`, `proxy.ice_getCommunicator()` returns an unrelated object (a `Prism::Node` in my run) and then segfaults; `IceRuby_ObjectPrx_mark` would be marking a moved slot. Same on main, so it's out of scope here, but it deserves its own issue.

Non-blocking nits:

- `Types.cpp:1427` raises `TypeError` while the byte/short element checks in the same function raise `RangeError`. Matching the scalar float path is defensible, just flagging the split.
- The `eql?`/`hash` block in `AllTests.rb` lands between the `info.adapterName` and `tcpinfo.remotePort` assertions, splitting the connection-info checks; it reads better after them.
- The new sequence float test covers only the positive overflow, while the scalar tests next to it cover both signs.

pepone (approved):
All four fixes look right. I built the extension at `02a9feab25` and ran the Ruby suite over tcp/ssl/ws: 19/19. Reverting only `ruby/src/IceRuby/` to the parent makes `Ice/info` (AllTests.rb:108) and `Ice/operations` (Twoways.rb:243) fail, so items 2–4 are genuinely red→green.

Nothing below blocks the merge.

## Item 1 is undersold — it fixes a hang at exit, and it is testable

The description says "No automated test for this item — the observable effect is GC-timing dependent". It isn't GC-timing dependent: it's static-destruction order, and it's deterministic. On the parent commit this program hangs forever at exit:

```ruby
require 'Ice'
class MyLoader; def newInstance(t); nil; end; end
initData = Ice::InitializationData.new
initData.sliceLoader = MyLoader.new
Ice::initialize(initData) { |c| c.stringToProxy("hello:tcp -p 1") }
puts "end of script"
```

`end of script` prints, then the process never exits (I had to SIGKILL it). `sample` on the hung process:

```
exit -> __cxa_finalize_ranges
  std::map<shared_ptr<Ice::Communicator>, shared_ptr<Ice::SliceLoader>>::~map()   <- _sliceLoaderMap
    Ice::CompositeSliceLoader::~CompositeSliceLoader()
      IceRuby::RubySliceLoader::~RubySliceLoader()
        _sigtramp -> sigsegv -> rb_bug_for_fatal_signal -> rb_vm_get_sourceline   <- SEGV, handler wedges
```

`~RubySliceLoader` calls `rb_gc_unregister_address` (`ruby/src/IceRuby/RubySliceLoader.cpp:16`), and the static destructor of `_sliceLoaderMap` runs after the Ruby VM is gone. Same hang for a communicator that is simply never destroyed. With this PR: exits 0 in 0.2s, zero leaked loaders.

The trigger is a custom `sliceLoader` **plus** the block form of `Ice::initialize` — the form `ruby/README.md:36` demonstrates. The existing custom-loader test misses it only because `TestHelper#init` always calls `communicator.destroy()` in an `ensure` (`ruby/test/TestHelper.rb:66-70`), which takes the pre-existing erase at `Communicator.cpp:249`.

So two suggestions:

- Add a changelog bullet for item 1. A valid program using the documented idiom hangs at exit; that clears the "users of valid programs can notice during normal operations" bar.
- Add a test. Anything that uses the block form with a custom slice loader and then exits normally is red (hang) on the parent and green here — a small variant alongside `ruby/test/Ice/objects` would do it.

## Smaller

- `ruby/test/Ice/operations/Twoways.rb:241-246` only covers the positive arm. The scalar float test just above covers both `3.402823466E38*2` and its negation; the new sequence test leaves the new `numeric_limits<float>::lowest()` branch uncovered. One more `begin`/`rescue`.
- `ruby/test/Ice/info/AllTests.rb:108-116` sits between `test(info.adapterName.length == 0)` and `test(tcpinfo.remotePort == port)`, splitting the contiguous `tcpinfo` assertions. Reads better moved after them.
- `IceRuby_Connection_equals` raises `TypeError` for a non-Connection argument (`ruby/src/IceRuby/Connection.cpp:191-196`) instead of returning false. Pre-existing, but item 2 is precisely about making `Connection` a well-behaved Hash key, and Ruby calls `eql?` across key types on a hash collision — so a colliding foreign key makes lookup raise. Fine here or as a follow-up.

## Checked, no issue

- The marking argument for erasing in `dfree` holds: `IceRuby_ObjectPrx_mark` (`Proxy.cpp:22-30`) marks the wrapper, so it is freed only once no Ruby proxy for that communicator is live. Ruby exposes no `Connection#createProxy`, so a bare `Connection` cannot reintroduce one. `GC.compact` with a live proxy is clean.
- Removing the find+erase before the insert is genuinely dead code — a live entry's `shared_ptr` key blocks address reuse.
- The double erase (`destroy` then free) is a harmless no-op, and the `destroy`-side erase is still load-bearing: it releases the Ruby slice loader promptly instead of at GC.
- `INT2FIX` on a 64-bit `size_t` drops a bit, but that matches `ObjectPrx#hash`/`Endpoint#hash`, and equal connections still hash equally.
- `@connectionId` propagates correctly through the `underlying` chain and matches `cpp/include/Ice/Connection.h:200`.
- Item 4 correctly gets no changelog entry: `.github/copilot-instructions.md:47-49` excludes fixes that only change how invalid API arguments fail, "even when the previous behavior was a crash". `TypeError` over `RangeError` is also the right call — it matches the scalar float path exactly.

bernardnormier:
Great catch on the exit hang — you're right that it's deterministic static-destruction order, not GC timing, and digging into it made it worse-than-described in a useful way: the block form of `Ice::initialize` destroys the communicator through `Ice::CommunicatorHolder`, i.e. the C++ API, so it never ran the Ruby-level erase — a program that does everything right (documented block form, communicator properly destroyed) still hung. All taken in 80d7fe3ee9:

- **Test** — `Ice/objects` now runs the block form with the existing `CustomSliceLoader` after the harness block and exits normally. Verified red/green locally: with `Communicator.cpp` reverted to the merge base, the client prints its last line and hangs forever; with the fix it exits in under a second.
- **Changelog** — new bullet for the exit hang (symptom only), and the description's item 1 now describes the hang and drops the incorrect GC-timing disclaimer.
- **`Connection#eql?`** — took the hash-collision point here rather than as a follow-up: returns `false` for a non-Connection argument, with an assertion pinning it.
- **Smaller** — negative-overflow arm added to the sequence float test; the `eql?`/`hash` block moved after the contiguous `tcpinfo` assertions.

On the `TypeError`-vs-`RangeError` split in the element checks: keeping `TypeError`, per your own second pass — it matches the scalar float path exactly.

The `_communicatorMap` compaction bug is yours already as #6327 (I briefly filed a duplicate, now closed).

Issue #6231, opened 2026-07-28
DataStorm: replace SampleT's _hasValue/_value pair with a std::optional


pepone:
`DataStormI::SampleT` tracks whether it holds a value with a separate `bool`, in `cpp/include/DataStorm/InternalT.h`:

```cpp
    private:
        bool _hasValue;
        // Value-initialized so a value-less sample (such as a remove) holds a determinate default value: scalar
        // types would otherwise be indeterminate, and encodeValue encodes this member for remove samples.
        Value _value{};
```

Keeping the two in sync is manual, and it is spread over four constructors, both `setValue` overloads and `decode`.
Nothing in the class enforces it, and `getValue()` returns `const Value&` whether or not the sample has a value, so a
caller that skips `hasValue()` silently gets the default-constructed value instead of the published one.

That is not hypothetical: `SampleT::decode` used to set `_hasValue = true` before calling the `Decoder`, so a `Decoder`
that threw left the sample claiming a value it never received. Fixed in #6207, but by convention rather than by
construction.

`std::optional<Value>` makes the invariant structural: `hasValue()` becomes `_value.has_value()`, `getValue()` can
assert on it, and the value-initialization above goes away because there is no longer a default-constructed value to
reason about.

The second half of that comment — `encodeValue` encoding this member for remove samples — is tracked separately in #6232, since
it stands on its own.

3.9.0 rather than a patch release: `InternalT.h` is included by the public `DataStorm/DataStorm.h`, so `SampleT` is
instantiated in application code and its layout is not ours to change in 3.8.x.

PR #6313, opened 2026-07-31
Publish DataStorm partial updates with an empty custom encoding

### LINKED ISSUE #6306: DataStorm: partial update with zero-byte encoded payload is silently replaced by full-value bytes
While reviewing the partial update encoding path, I noticed that `SampleT::encode` uses "empty encoded value" as a *not yet encoded* sentinel:

https://github.com/zeroc-ice/ice/blob/a2a727827000dd62afca24b7f4d9ded67c68e44a/cpp/include/DataStorm/InternalT.h#L379-L386

For samples created by the `partialUpdate` updater lambdas, the constructor stores the user-encoded **update** payload in `_encodedValue` and sets `_hasValue = true` after the updater runs:

https://github.com/zeroc-ice/ice/blob/a2a727827000dd62afca24b7f4d9ded67c68e44a/cpp/include/DataStorm/DataStorm.h#L1415-L1424

If a custom `UpdateValue` encoder legitimately produces **zero bytes** (e.g., an empty string/sequence with a custom encoder, or a stateless "toggle" update type), `encode()` later sees `_encodedValue.empty()` and replaces the payload with `encodeValue(communicator)`, which encodes the **full resolved value** (since `_hasValue` is true). The sample is still forwarded with `SampleEvent::PartialUpdate`, so receivers decode full-value bytes with the partial-update decoder — silent data corruption (or a decode error).

This is the same bug class that was fixed for regular values in #6084 / #6175, where zero-byte custom encodings were explicitly acknowledged as legitimate. The fix there doesn't cover the partial-update payload path.

Suggested fix: track "encoded" state explicitly (e.g., a `bool _encoded` or `std::optional<Ice::ByteSeq>`) instead of using `empty()` as the sentinel. Note #6231 (refactoring `SampleT` to use `std::optional`) would provide a natural home for this.

Found during a code review of DataStorm at a2a7278270; no test written, but the code path is reachable with any custom update encoder that can produce 0 bytes.

pepone:
Fixes #6306.

A partial update whose custom `Encoder` produces an empty byte sequence was published with the encoding of the sample's full resolved value instead of its own payload, while the sample event remained `PartialUpdate`. Readers then decoded the full value's bytes as a partial update value, silently corrupting the update. Same bug class as #6084 (fixed by #6175), which established that zero-byte custom encodings are legitimate.

`SampleT::encode` used the emptiness of `_encodedValue` as its "not yet encoded" marker. The sample now tracks the encoded state with a dedicated `_hasEncodedValue` flag, so a legitimately empty payload is forwarded as-is. The flag is reset in `decode()`, preserving the existing re-encode behavior for decoded samples.

The `std::optional<Ice::ByteSeq>` refactoring suggested in the issue is left to #6231, which is scoped for 3.9.0; this fix uses a plain flag so it can be cherry-picked to 3.8.

Changes:
- `cpp/include/DataStorm/InternalT.h`: track encoded state explicitly instead of treating an empty payload as not yet encoded. Header-only change to a class instantiated only in application code; no library ABI impact.
- `cpp/test/DataStorm/partial`: new test section publishing a zero-byte-encoded partial update; it fails without the fix (the reader resolves 5 + 5 = 10 from the corrupted full-value payload instead of 5 + 0 = 5) and passes with it.
- `changelog.d/datastorm/6306.md`: changelog fragment.

The full DataStorm test suite (11 tests) passes on macOS.

### DIFF
--- a/changelog.d/datastorm/6306.md
+++ b/changelog.d/datastorm/6306.md
@@ -1,0 +1,4 @@
+- Fixed a bug where a partial update that a custom `Encoder` encodes to an empty byte sequence was published with
+  the encoding of the sample's full resolved value instead of the partial update's own payload, while the sample
+  event remained `PartialUpdate`. Readers then decoded the full value's bytes as a partial update value. The sample
+  now tracks whether its payload is encoded instead of treating an empty payload as not yet encoded.
--- a/cpp/include/DataStorm/InternalT.h
+++ b/cpp/include/DataStorm/InternalT.h
@@ -320,7 +320,8 @@ namespace DataStormI
             Ice::ByteSeq value,
             std::int64_t timestamp)
             : Sample(std::move(session), std::move(origin), id, event, key, tag, std::move(value), timestamp),
-              _hasValue(false)
+              _hasValue(false),
+              _hasEncodedValue(true)
         {
         }
 
@@ -330,7 +331,8 @@ namespace DataStormI
 
         SampleT(Ice::ByteSeq value, const std::shared_ptr<Tag>& tag)
             : Sample(DataStorm::SampleEvent::PartialUpdate, tag),
-              _hasValue(false)
+              _hasValue(false),
+              _hasEncodedValue(true)
         {
             _encodedValue = std::move(value);
         }
@@ -378,9 +380,10 @@ namespace DataStormI
 
         [[nodiscard]] const Ice::ByteSeq& encode(const Ice::CommunicatorPtr& communicator) final
         {
-            if (_encodedValue.empty())
+            if (!_hasEncodedValue)
             {
                 _encodedValue = encodeValue(communicator);
+                _hasEncodedValue = true;
             }
             return _encodedValue;
         }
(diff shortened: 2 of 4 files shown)


pepone (commented):
Codex review

No blocking correctness findings for `main`. The encoded-state flag is handled consistently across construction, encoding, and successful decoding, and the regression test correctly distinguishes the intended `5 + 0 = 5` result from the previous corrupted `5 + 5 = 10` behavior.

Validation performed:
- `python3 allTests.py DataStorm/partial --show-durations` passed both direct and multicast variants.
- `git diff --check` passed.
- GitHub CI is green.

One backport caveat: `_hasEncodedValue` changes the layout of the header-defined `DataStormI::SampleT`. That is allowed on `main`, but issue #6231 explicitly defers this kind of layout change to 3.9. For a 3.8 backport, consider a layout-preserving variant that treats `PartialUpdate` as already encoded even when `_encodedValue` is empty.

— Codex

pepone:
Moving this to 3.9.0 and closing for now.

The fix adds a member to `DataStormI::SampleT`, which is defined in a public header and instantiated in application
code, so its layout isn't ours to change in 3.8.x — the same reason #6231 is scheduled for 3.9.0.

The branch is preserved at
https://github.com/pepone/ice/tree/fix/datastorm-6306-partial-update-empty-encoding (a35ee84).

PR #6314, opened 2026-07-31
Hash the proxy request context in key order in Ice for JavaScript

### LINKED ISSUE #6114: JavaScript: equal proxies can hash differently when their context entries were inserted in different orders
> [!NOTE]
> Originally filed as a four-item low-severity batch. Because the items land in different releases,
> they have been split: sub-item 2 → #6217, sub-item 3 → #6218 (3.9.0), sub-item 4 → #6219 plus
> #6112. This issue now covers the context-hash item only. See the comment below for details.

## Summary

`Reference.hashCode` folds the context `Map` in iteration order — which for a JS `Map` is insertion
order — at `Reference.js:142-152`:

```js
if (this._context !== null && this._context !== undefined) {
    for (const [key, value] of this._context) {
        h = HashUtil.addString(h, key);
        h = HashUtil.addString(h, value);
    }
}
```

but `Reference.equals` compares contexts order-independently via `MapUtil.equals`
(`Reference.js:310`). `Ice.HashMap` requires both to agree: it short-circuits on the stored hash
before ever consulting the comparator (`HashMap.js:88-97`). Two equal proxies whose context entries
were inserted in different orders therefore miss on `get`/`has`/`delete` — and `set` inserts a
*duplicate* entry rather than replacing the existing one.

Measured: a proxy with context `{a: "1", b: "2"}` hashes to 2087644261 and the same proxy with
`{b: "2", a: "1"}` hashes to 2087676901, while `equals` returns `true` for the pair.

JavaScript is the only mapping with this asymmetry:

- C++ folds a **sorted** container — the `hashAdd` overload takes `std::map` (`cpp/src/Ice/HashUtil.h:56-63`).
- Java delegates to `Map.hashCode()`, contractually the order-independent sum of entry hashes (`Reference.java:407`).
- C# hashes only the entry count, with an explicit comment (`csharp/src/Ice/Internal/Reference.cs:349`).

`ObjectPrx.hashCode()` is public and documented (`ObjectPrx.d.ts`), and `Ice.HashMap`'s declared key
type explicitly sanctions `{ equals, hashCode }` keys, so proxy-keyed maps are a supported pattern
rather than an accident. The runtime's own proxy-keyed tables (`LocatorManager`, `RouterManager`) are
not affected in practice, because the proxies they store come from properties or a single
`setDefaultLocator` call, so their context ordering is deterministic.

**File:** `js/packages/ice/src/Ice/Reference.js:142-152`

## Fix

Make the context contribution order-independent by sorting the entries by key before the existing
fold. Hashing only `_context.size` (as C# does) would also restore the invariant, but it collapses
every proxy differing only by context — `prx.ice_context(perUserContext)`, precisely the pattern
that motivates the fix — into a single bucket.

Note that a naive "XOR the entry hashes together" is not a safe substitute: XOR-ing keys and values
independently loses the key↔value association, and pure XOR cancels equal entry hashes.

**Severity:** Low. Pre-existing since `c8d32e0487` (2016), present in every 3.7/3.8 release — not a
regression.

pepone:
Fixes #6114

`Reference.hashCode` folded the context `Map` in iteration order — which for a JS `Map` is insertion
order — while `Reference.equals` compares contexts order-independently through `MapUtil.equals`.
`Ice.HashMap` short-circuits on the stored hash before consulting the comparator, so two equal
proxies whose context entries were added in different orders missed on `get`/`has`/`delete`, and
`set` inserted a duplicate entry instead of replacing the existing one.

Fold the entries in key order instead, which is what the C++ mapping gets for free by folding a
`std::map`.

JavaScript was the only mapping with this asymmetry: Java delegates to `Map.hashCode()`, and C#
hashes only the entry count.

Contexts that genuinely differ still hash differently — including ones differing only in the
key/value association, such as `{a: "1", b: "2"}` versus `{a: "2", b: "1"}`. The proxy's stored
context keeps its own iteration order, so nothing changes on the wire.

The added assertion in `Ice/proxy` fails without the fix.

### DIFF
--- a/js/packages/ice/src/Ice/Reference.js
+++ b/js/packages/ice/src/Ice/Reference.js
@@ -144,9 +144,11 @@ export class Reference {
         h = HashUtil.addNumber(h, this._mode);
         h = HashUtil.addHashable(h, this._identity);
         if (this._context !== null && this._context !== undefined) {
-            for (const [key, value] of this._context) {
+            // Fold the entries in key order. equals compares contexts without regard to order, so the hash code must
+            // not depend on the order in which the entries were inserted into the Map.
+            for (const key of [...this._context.keys()].sort()) {
                 h = HashUtil.addString(h, key);
-                h = HashUtil.addString(h, value);
+                h = HashUtil.addString(h, this._context.get(key));
             }
         }
         h = HashUtil.addString(h, this._facet);
--- a/js/packages/test/Ice/proxy/Client.ts
+++ b/js/packages/test/Ice/proxy/Client.ts
@@ -730,6 +730,19 @@ export class Client extends TestHelper {
         test(!compObj.ice_context(null!).equals(compObj.ice_context(ctx2)));
         test(!compObj.ice_context(ctx1).equals(compObj.ice_context(ctx2)));
 
+        // Two proxies with the same context entries are equal and hash the same, whatever the order in which the
+        // entries were added to the context.
+        const ctx3 = new Map([
+            ["ctx1", "v1"],
+            ["ctx2", "v2"],
+        ]);
+        const ctx4 = new Map([
+            ["ctx2", "v2"],
+            ["ctx1", "v1"],
+        ]);
+        test(compObj.ice_context(ctx3).equals(compObj.ice_context(ctx4)));
+        test(compObj.ice_context(ctx3).hashCode() === compObj.ice_context(ctx4).hashCode());
+
         let compObj1 = new Ice.ObjectPrx(communicator, "foo:" + defaultProtocol + " -h 127.0.0.1 -p 10000");
         let compObj2 = new Ice.ObjectPrx(communicator, "foo:" + defaultProtocol + " -h 127.0.0.1 -p 10001");
         test(!compObj1.equals(compObj2));


externl (approved):
Looks good!

bernardnormier (approved):
What about not hashing the context at all, or only hashing its size like in C#?

I would also drop the changelog fragment. It's a super obscure bug: using proxies as keys in a hash map is unusual, adding context to proxies is unusual too, and the combination is ultra unusual.

pepone:
Sorting keeps proxies that differ only by their context in distinct buckets, and `prx.ice_context(perUserCtx)` is exactly the pattern this fix is about. With size-only hashing, 2000 proxies differing only by a one-entry `{user: uN}` context collapse to a single hash: 2000 `set` + 2000 `get` went from 8.5 ms to 184 ms in my measurement, and it degrades quadratically from there.

C# is the outlier rather than the norm here — C++ folds a sorted `std::map`, Java delegates to `Map.hashCode()`, and Python/Ruby delegate to `std::hash<Ice::ObjectPrx>`. `hashCode` is off the invocation path, so the sort costs nothing where it matters.

Changelog fragment dropped.

PR #6317, opened 2026-07-31
Complete DataStorm forwarded session requests when the node session is gone

### LINKED ISSUE #6307: DataStorm: NodeForwarder AMD operations never complete the dispatch when the node session is destroyed
The `NodeForwarder` AMD implementations only invoke the response/exception callbacks inside `if (auto nodeSession = _nodeSession.lock())`:

https://github.com/zeroc-ice/ice/blob/a2a727827000dd62afca24b7f4d9ded67c68e44a/cpp/src/DataStorm/NodeSessionI.cpp#L36-L129

`initiateCreateSessionAsync`, `createSessionAsync`, and `confirmCreateSessionAsync` all follow this pattern (as does `updateNodeAndSessionProxy` returning the un-forwardable proxy case, which is fine because callers handle it). When the node session has been destroyed and the weak pointer can't be locked, the methods return **without calling `response()` or `exception()`**.

Impact:

- The peer's `initiateCreateSession` / `createSession` / `confirmCreateSession` dispatch never completes. With no invocation timeout configured (the default), the caller's outstanding async invocation hangs until the connection closes.
- More importantly, the caller's failure path never runs, so session-creation retry accounting (`sessionCreationFailed`, retry scheduling per #6287) doesn't fire for that attempt.

Suggested fix: add an `else` branch that completes the dispatch, e.g. `exception(make_exception_ptr(Ice::ObjectNotExistException{__FILE__, __LINE__}))`, so the caller gets a prompt failure and can retry through the normal path.

pepone:
Fixes #6307.

The `NodeForwarder` AMD implementations (`initiateCreateSessionAsync`, `createSessionAsync`, `confirmCreateSessionAsync`) only completed the dispatch inside `if (auto nodeSession = _nodeSession.lock())`. When the node session was already destroyed, they returned without calling `response` or `exception`, so the peer's session creation dispatch never completed: with no invocation timeout configured (the default), the caller waited forever and its session creation retry accounting (`sessionCreationFailed`) never ran for that attempt.

Each operation now completes the dispatch with `Ice::ObjectNotExistException` when the node session is gone. This is the same failure the caller gets once `NodeSessionI::destroy` removes the forwarder from the object adapter — the destroyed node session means the target node is no longer reachable through this forwarder — and `SessionI::retryImpl` treats it as retryable, so the caller accounts for the failure and retries through the normal announcement path.

Changes:
- `cpp/src/DataStorm/NodeSessionI.cpp`: add the missing `else` branches.
- `changelog.d/datastorm/6307.md`: changelog fragment.

No new test: the window requires the node session to be destroyed between the dispatch of a forwarded session creation request and the `_nodeSession.lock()` call, which isn't deterministically reachable from the public API. The full DataStorm test suite (11 tests, including the relay tests that exercise the `NodeForwarder` paths) passes on macOS.

### DIFF
--- a/changelog.d/datastorm/6307.md
+++ b/changelog.d/datastorm/6307.md
@@ -1,0 +1,4 @@
+- Fixed a bug where a session creation request forwarded through a relay node was never answered when the relay's
+  session for the target node was destroyed while the request was being forwarded. The requesting node's session
+  creation attempt then hung instead of failing and retrying: the forwarder now completes such requests with
+  `ObjectNotExistException`, the same failure the requester gets once the forwarder is unregistered.
--- a/cpp/src/DataStorm/NodeSessionI.cpp
+++ b/cpp/src/DataStorm/NodeSessionI.cpp
@@ -54,6 +54,14 @@ namespace
                     exception(make_exception_ptr(SessionCreationException{SessionCreationError::NodeShutdown}));
                 }
             }
+            else
+            {
+                // The node session is destroyed: the target node is no longer reachable through this forwarder.
+                // Complete the dispatch with the same failure the caller gets once the forwarder is removed from the
+                // object adapter, so the caller accounts for the failure and can retry, instead of waiting forever
+                // for a response.
+                exception(make_exception_ptr(ObjectNotExistException{__FILE__, __LINE__}));
+            }
         }
 
         void createSessionAsync(
@@ -92,6 +100,11 @@ namespace
                     exception(make_exception_ptr(SessionCreationException{SessionCreationError::NodeShutdown}));
                 }
             }
+            else
+            {
+                // See initiateCreateSessionAsync.
+                exception(make_exception_ptr(ObjectNotExistException{__FILE__, __LINE__}));
+            }
         }
 
         void confirmCreateSessionAsync(
@@ -126,6 +139,11 @@ namespace
                 {
                     exception(make_exception_ptr(SessionCreationException{SessionCreationError::NodeShutdown}));
                 }
+            }
+            else
+            {
+                // See initiateCreateSessionAsync.
+                exception(make_exception_ptr(ObjectNotExistException{__FILE__, __LINE__}));
             }
         }
 


changelog.d/datastorm/6307.md
```diff
@@ -0,0 +1,4 @@
+- Fixed a bug where a session creation request forwarded through a relay node was never answered when the relay's
+  session for the target node was destroyed while the request was being forwarded. The requesting node's session
+  creation attempt then hung instead of failing and retrying: the forwarder now completes such requests with
+  `ObjectNotExistException`, the same failure the requester gets once the forwarder is unregistered.
```
bernardnormier, line 4: Sessions are an internal implementation detail of DataStorm, and the exception isn't application-visible either (the node retries internally). Suggest rewording in the documentation's vocabulary:

```suggestion
- Fixed a bug where a node could fail to connect to a peer reached through a relay node when the relay lost its own
  connection to that peer at the wrong moment. The writers and readers of the two nodes then never connected to each
  other. The node now retries the connection.
```

bernardnormier (approved):
One nit on the PR body (worth fixing before merge since it becomes the commit message): `sessionCreationFailed` doesn't exist on main — it only appeared in interim commits of the #6287 branch. The current accounting path is `retryPublisherSessionCreation`/`retrySubscriberSessionCreation` → `SessionI::retry`.

externl (approved):
Looks good!

pepone:
🤖 **Claude:**

@bernardnormier `sessionCreationFailed` is on `main` — it's in the merged form of #6287, not only in that branch's interim commits. Merge commit [`a2a7278`](https://github.com/zeroc-ice/ice/commit/a2a727827000dd62afca24b7f4d9ded67c68e44a), merged 2026-07-31 08:07 UTC, about 12 hours before your review. On `main` today: declaration at `cpp/src/DataStorm/SessionI.h:347`, definition at `cpp/src/DataStorm/SessionI.cpp:805`, callers at `cpp/src/DataStorm/NodeI.cpp:505` and `:535`.

The accounting path is `NodeI::retryPublisherSessionCreation`/`retrySubscriberSessionCreation` → `SessionI::sessionCreationFailed` → `SessionI::retryImpl`; there's no `SessionI::retry` on `main`. So I'm keeping the description as is — a checkout of `main` from earlier that day would show exactly what you describe.

changelog.d/datastorm/6307.md
pepone: 🤖 **Claude:** Applied verbatim in [`fdbacad`](https://github.com/zeroc-ice/ice/commit/fdbacad15fc664f277618d971acded92f8f5e256), pushed directly to `main` since this PR had already merged.

PR #6320, opened 2026-07-31
Derive the WebSocket masking key and handshake nonce from a CSPRNG

### LINKED ISSUE #6261: C# and Java: the WebSocket masking key and handshake nonce come from a non-cryptographic PRNG
> [!NOTE]
> Found while investigating #6206. Independent of that erratum: these two mappings never used a strong
> source here.

## Summary

RFC 6455 §5.3 requires the client's per-frame masking key to be "unpredictable" and "derived from a strong source of entropy", because masking is what stops a malicious script from steering the bytes a non-conforming intermediary sees on the wire. C# and Java derive both the masking key and the handshake nonce from a general-purpose PRNG.

**C#** — `csharp/src/Ice/Internal/WSTransceiver.cs`:

```csharp
_rand = new Random();                    // :718
...
_rand.NextBytes(key);                    // :64    Sec-WebSocket-Key
...
_rand.NextBytes(_writeMask);             // :1678  per-frame mask
```

**Java** — `java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/WSTransceiver.java`:

```java
_rand = new Random();                    // :428
...
_rand.nextBytes(key);                    // :62    Sec-WebSocket-Key
...
_rand.nextBytes(_writeMask);             // :1181  per-frame mask
```

`java.util.Random` is a 48-bit LCG whose full state is recoverable from two consecutive `nextInt` outputs; `System.Random` is a subtractive generator with comparable properties. Neither is suitable, and in both cases the same instance produces the handshake nonce and every subsequent mask, so observing one leaks the others.

C++ uses `IceInternal::generateRandom` for both values (`cpp/src/Ice/WSTransceiver.cpp:235,1868`) — which is exactly the primitive #6233 is hardening. JavaScript wraps the browser/Node WebSocket implementation and never masks by hand; Python, Ruby, PHP, MATLAB and Swift use the C++ core.

## Fix

- C#: `System.Security.Cryptography.RandomNumberGenerator.Fill(span)` — no instance to hold, no seeding.
- Java: a `java.security.SecureRandom` in place of `java.util.Random`. `nextBytes` on a shared `SecureRandom` is thread-safe, so the field can stay as it is.

Both are drop-in at the four call sites above.

### Explicitly out of scope

Whether Ice should mask at all in deployments with no intermediary — masking is mandatory for clients under RFC 6455 regardless, so this stays a conformance fix.

*Filed from an AI-assisted correctness review of the random-number paths, cross-checked against the source. Not yet human-triaged.*

pepone:
Fixes #6261.

RFC 6455 section 5.3 requires the client's per-frame masking key to be unpredictable and derived from a strong
source of entropy. The C# and Java WebSocket transports derived both the `Sec-WebSocket-Key` handshake nonce and
every masking key from a general-purpose PRNG (`System.Random` / `java.util.Random`), and from the *same* instance,
so observing one value leaked the others.

- **C#**: `RandomNumberGenerator.Fill`. It is static, so the `_rand` field and its initialization are gone.
- **Java**: `SecureRandom` in place of `java.util.Random`; `nextBytes` is a drop-in.

C++ already used `IceInternal::generateRandom` for both values, and JavaScript wraps the native WebSocket
implementation and never masks by hand, so no other mapping is affected.

### Verification

Both mappings build clean with no new warnings, `dotnet format --verify-no-changes` and the Java `checkstyleMain` /
`rewriteDryRun` tasks pass, and `Ice/operations` over `--protocol=ws` is green in C# and Java — that exercises the
handshake nonce and the per-frame masking path.

No test: whether the bytes come from a CSPRNG has no observable behavior to assert, and the WebSocket functional
path is already covered.

### Performance

`SecureRandom` is kept as a per-transceiver field, so incoming (server) transceivers construct one they never use,
and `nextBytes` now runs per outgoing frame. Measured locally: `new SecureRandom()` is 1.7 µs, once per connection
against a full HTTP upgrade round trip, and `nextBytes(4)` costs ~265 ns more than `Random.nextBytes(4)`. Both are
negligible, so the field structure is unchanged.

### DIFF
--- a/csharp/src/Ice/Internal/WSTransceiver.cs
+++ b/csharp/src/Ice/Internal/WSTransceiver.cs
@@ -61,7 +61,7 @@ internal sealed class WSTransceiver : Transceiver
                     // encoded with Base64.
                     //
                     byte[] key = new byte[16];
-                    _rand.NextBytes(key);
+                    RandomNumberGenerator.Fill(key);
                     _key = System.Convert.ToBase64String(key);
                     @out.Append(_key + "\r\n\r\n"); // EOM
 
@@ -715,7 +715,6 @@ internal sealed class WSTransceiver : Transceiver
         _writeMask = new byte[4];
         _key = "";
         _pingPayload = [];
-        _rand = new Random();
     }
 
     private void handleRequest(Buffer responseBuffer)
@@ -1675,7 +1674,7 @@ internal sealed class WSTransceiver : Transceiver
             // and apply the mask.
             //
             _writeBuffer.b.put(1, (byte)(_writeBuffer.b.get(1) | FLAG_MASKED));
-            _rand.NextBytes(_writeMask);
+            RandomNumberGenerator.Fill(_writeMask);
             _writeBuffer.b.put(_writeMask);
         }
     }
@@ -1741,8 +1740,6 @@ internal sealed class WSTransceiver : Transceiver
     private bool _writePending;
 
     private byte[] _pingPayload;
-
-    private Random _rand;
 
     //
     // WebSocket opcodes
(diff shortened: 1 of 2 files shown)


externl (approved):
Looks good!

bernardnormier (approved):
I would drop both changelog fragments — this is internal hardening / spec compliance rather than a user-visible fix. The masking key protects intermediaries (cache poisoning), not application data, and the handshake nonce isn't a secret; nothing an application observes changes, and there's no action for users to take.

PR #6322, opened 2026-07-31
Mark a DataStorm reader initialized only after its samples are decoded

### LINKED ISSUE #6308: DataStorm: subscriberInitialized marks the reader initialized before decoding samples, losing init history if a decoder throws
`SessionI::subscriberInitialized` marks the element subscriber initialized and advances `lastId` **before** decoding the inline keys and building the samples:

https://github.com/zeroc-ice/ice/blob/a2a727827000dd62afca24b7f4d9ded67c68e44a/cpp/src/DataStorm/SessionI.cpp#L1432-L1492

`elementSubscriber->initialized = true` is set at line 1451 and `lastId` is advanced at line 1472, but the per-sample key decoding (`topic->getKeyFactory()->decode(...)`) and sample construction happen after, at lines 1474–1494. If a user-provided key or value decoder throws there, the exception propagates out of the dispatch, yet the subscriber is already marked initialized with an advanced `lastId` — the init samples are lost permanently: the reader is treated as initialized, the writer won't resend (its `getLastIds` filtering sees the advanced id), and no redelivery occurs.

This is inconsistent with the deliberate ordering in the `initSamples` path, which was fixed to do the opposite:

https://github.com/zeroc-ice/ice/blob/a2a727827000dd62afca24b7f4d9ded67c68e44a/cpp/src/DataStorm/SessionI.cpp#L620-L631

There the comment states the reader is marked initialized *only after* its samples are decoded and built, and `initialized = true` / `lastId = samplesI.back()->id` come last.

Suggested fix: restructure `subscriberInitialized` the same way — decode keys and build the sample vector first, and only then set `initialized` / advance `lastId` (and deliver via `initSamples`).

pepone:
`SessionI::subscriberInitialized` marked the element subscriber initialized and advanced its last
received sample id *before* decoding the sample keys and building the samples. When decoding a key threw
— a custom `Decoder<Key>` specialization, or the default decoder on a malformed inline key — the
subscriber was left initialized with an advanced last id, so the writer filtered these samples out of
every later initialization and the reader lost them for good, including across a reconnect.

The subscriber is now marked initialized, and its last id advanced, only after its samples are decoded
and built — the ordering `SessionI::initSamples` already uses and documents.

### Notes

- **Reachability.** The keys of the initialization samples are decoded only when `key == nullptr`, which
  `TopicI::attachElementsAck` passes on exactly one branch (`spec.peerId <= 0 && spec.id < 0`): an
  any-key writer paired with an any-key or filtered-key reader, and only on the `attachElementsAck`
  path. It does not need a custom decoder — the default `Decoder<K>` unmarshals the peer-supplied
  `keyValue` and throws `MarshalException` on malformed bytes.

- **Why the loss is permanent.** `ElementSubscribers::addSubscriber` resets `initialized` on re-attach
  but preserves `lastId`, and `disconnectedImpl` never clears `_topics`. `getLastIds` reports the
  inflated id to the writer, whose `getSamples` skips both the history and the current per-key base at
  `id <= lastId`. The ack is a oneway call, so the dispatch failure produces no reply at all — the only
  symptom anywhere is the receiver-side `Ice.Warn.Dispatch` warning.

- **Behavior change, deliberate.** A reader whose key decoding fails now stays uninitialized, so it also
  stops receiving live samples from that writer (`SubscriberSessionI::s` gates on `initialized`) until
  the session initializes it again. That is the trade `initSamples` already made; the previous state —
  initialized with no per-key base — silently discarded every later partial update for those keys
  anyway.

- **No test.** A red→green test needs a throwing decoder plus a forced reconnect plus a pinned
  single-connection topology, and its redness rests on two invariants it cannot assert: that
  initialization arrives via the ack path rather than `initSamples`, and that the throw fires on that
  exact decode.

- The trace emitted before the decode loop now says `initializing` rather than `initialized`, matching
  the sibling path, since it precedes the step that can fail.

### Not addressed here

The same commit-before-delivery window remains one frame up: `TopicI::attachElementsAck` commits each
element's initialization inside the spec loop but runs the collected `initCallbacks` only after the loop,
so a throw in a later spec discards an earlier reader's already-built samples after that reader was
marked initialized. Filed as #6324.

Fixes #6308

### DIFF
--- a/changelog.d/datastorm/6308.md
+++ b/changelog.d/datastorm/6308.md
@@ -1,5 +1,0 @@
-- Fixed a bug where a reader permanently lost a writer's initialization samples when decoding one of their keys
-  threw. The reader was marked initialized, with its last received sample id advanced, before the samples were
-  decoded, so the writer treated them as already received and never offered them again, not even after a reconnect.
-  The reader now stays uninitialized in that case, so the writer can still offer these samples when the session
-  initializes the reader again.
--- a/changelog.d/datastorm/6308.md
+++ b/changelog.d/datastorm/6308.md
@@ -1,0 +1,5 @@
+- Fixed a bug where a reader permanently lost a writer's initialization samples when decoding one of their keys
+  threw. The reader was marked initialized, with its last received sample id advanced, before the samples were
+  decoded, so the writer treated them as already received and never offered them again, not even after a reconnect.
+  The reader now stays uninitialized in that case, so the writer can still offer these samples when the session
+  initializes the reader again.
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -1446,9 +1446,8 @@ SessionI::subscriberInitialized(
     if (_traceLevels->session > 1)
     {
         Trace out(_traceLevels->logger, _traceLevels->sessionCat);
-        out << _id << ": initialized '" << element << "' from 'e" << elementId << '@' << topicId << "'";
+        out << _id << ": initializing '" << element << "' from 'e" << elementId << '@' << topicId << "'";
     }
-    elementSubscriber->initialized = true;
 
     // If the samples collection is empty, the element subscriber's lastId remains unchanged:
     // - If no samples have been received, lastId is 0.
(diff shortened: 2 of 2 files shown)


externl (approved):
Looks good!

bernardnormier (approved):
The code change looks good.

However, I would drop the changelog fragment. The decoding of a key should never throw: a `Decoder<Key>` that throws indicates a buggy custom decoder or malformed data, not something that occurs during normal operations. That makes this fix internal hardening, which we don't advertise in the CHANGELOG. The full story is well captured in the PR description, which is where it belongs.

Issue #6324, opened 2026-07-31
DataStorm: reader initialization state is committed before the samples are delivered


pepone:
`SessionI::subscriberInitialized` marks an element subscriber initialized and advances its `lastId`, but it does not deliver the samples it builds. `DataElementI::attach` wraps them in a closure and returns it:

https://github.com/zeroc-ice/ice/blob/8e8e65c0d5d3edacb1be9e5ed066ff5f7a0e698d/cpp/src/DataStorm/DataElementI.cpp#L201-L208

and `TopicI::attachElementsAck` collects those closures and runs them only once the whole spec loop has completed:

https://github.com/zeroc-ice/ice/blob/8e8e65c0d5d3edacb1be9e5ed066ff5f7a0e698d/cpp/src/DataStorm/TopicI.cpp#L591-L596

So for spec *k* the subscriber state is committed at iteration *k*, while delivery happens after iteration *N*. Anything that throws in iterations *k+1..N* unwinds the loop and destroys the pending closures, and there is no `try`/`catch` anywhere on the path. Reader *k* is then marked `initialized` with `lastId` advanced past samples it never received; `getLastIds` reports that id to the peer writer, whose `getSamples` skips both the history and the current per-key base at `id <= lastId`, so those samples are never offered again — not even after a reconnect, since `ElementSubscribers::addSubscriber` resets `initialized` but preserves `lastId`.

That is #6308, one frame up. #6322 fixes the ordering inside `subscriberInitialized` itself, but does not close this window.

Throw sites reachable in a later iteration, none of which need application code:

- `TopicI.cpp:476` — `_keyFilterFactories->decode` on peer-supplied criteria. The built-in `_regex` key filter is registered for every topic with a streamable key, and its factory constructs a `std::regex` from the criteria, which throws `std::regex_error`. Regex dialects differ across libstdc++/libc++/MSVC, so a pattern that the sender accepts locally can throw on the receiver.
- `TopicI.cpp:544` — `_keyFactory->decode` on the peer's key bytes; the default `Decoder<K>` unmarshals and throws `MarshalException` on malformed input.
- `TopicI.cpp:497` and `TopicI.cpp:561` — `filter->match`, an application lambda.
- The user `Encoder<Key>`/`Encoder<Value>` and sample filters invoked from `getSamples` on the writer side of the same loop.

[...]

Issue #6213, opened 2026-07-28
DataStorm: an unresolvable partial-update tag is silently applied as the previous value


pepone:
> **AI-generated finding** — identified by Claude (Anthropic), model **Claude Opus 5** (`claude-opus-5`), while planning the fixes for #6128 / #5781 / #6195. Every code-level claim below was verified against the source on `main` (`cde286fb14`) with file:line evidence. **Not yet triaged by a human** — verify the reasoning and severity before acting. I could not demonstrate a trigger in a default configuration; see [Reachability](#reachability).

## Summary

A `PartialUpdate` sample carries a tag id, and the receiving session maps it back to a `Tag` through
`TopicSubscriber::tags`. That lookup is done with `operator[]`, so a tag id the session has not been told about
does not fail — it *inserts* a null `Tag` and hands it to the sample. `TopicI::getUpdater` then finds no updater
for a null tag and falls back to `noOpUpdater`, which copies the **previous** value into the sample.

The result is not a dropped sample and not an error. The reader delivers a sample that looks like a normal update
but carries the old value, and that sample becomes the key's new base — so the key silently stops tracking the
writer until the next full sample resynchronizes it. Nothing is logged at any trace level.

## Mechanism (verified against `origin/main`, `cde286fb14`)

1. **A full sample sends tag id 0; a partial update sends the updater's tag id.**
   `DataElementI.cpp:29`: `.tag = sample->tag ? sample->tag->getId() : 0`.

2. **The receiver resolves the tag id with `operator[]`, at all three sample paths.**
   `SessionI.cpp:606` (`initSamples`), `:1410` (`subscriberInitialized`), `:1640` (`SubscriberSessionI::s`), all
   of the form `subscriber.tags[sample.tag]`. `tags` is
   `std::map<std::int64_t, std::shared_ptr<Tag>>` (`SessionI.h:184`), so a miss inserts `{id, nullptr}` and yields
   a null `Tag`. The map is otherwise written only by `attachTopic` (`:224`), `attachTags` (`:324`) and
   `detachTags` (`:350`) — this is the only insert-on-read.

   Side effect, harmless but worth noting: because every full sample looks up tag id `0`, each subscriber
   permanently holds a junk `{0, nullptr}` entry.

3. **A null tag resolves to the no-op updater.**
   `TopicI::getUpdater` (`TopicI.cpp:630-639`) does `_updaters.find(tag)` and returns `noOpUpdater` when there is
[...]

PR #6343, opened 2026-08-03
Fix three low-severity issues in the DataStorm session

### LINKED ISSUE #6310: DataStorm: tag lookups in sample construction use map::operator[], inserting null Tag entries
Three sample-construction paths in `SessionI.cpp` look up the partial-update tag with `map::operator[]`, which inserts a null `Tag` entry into `TopicSubscriber::tags` when the tag id is unknown:

- `initSamples`: https://github.com/zeroc-ice/ice/blob/a2a727827000dd62afca24b7f4d9ded67c68e44a/cpp/src/DataStorm/SessionI.cpp#L623
- `subscriberInitialized`: https://github.com/zeroc-ice/ice/blob/a2a727827000dd62afca24b7f4d9ded67c68e44a/cpp/src/DataStorm/SessionI.cpp#L1492
- `SubscriberSessionI::s` (live samples): https://github.com/zeroc-ice/ice/blob/a2a727827000dd62afca24b7f4d9ded67c68e44a/cpp/src/DataStorm/SessionI.cpp#L1726

Besides quietly growing the `tags` map with null entries for ids that were never announced (they then persist for the lifetime of the topic subscriber), this contrasts with the key lookup path, where `findKey` was deliberately made non-inserting. Non-partial-update samples take these paths too (their `tag` is 0), so every such sample can insert/hit a null entry for id 0.

Related: #6213 tracks the broader issue of what to do when a partial update's tag can't be resolved; whatever behavior is chosen there, these lookups should use `find()` rather than `operator[]`.

### LINKED ISSUE #6311: DataStorm: SubscriberSessionI::reconnect can leak SessionCreationException to the timer thread, logging "uncaught exception"
`SubscriberSessionI::reconnect` runs on the IceInternal timer thread (scheduled by the session retry task) and calls `NodeI::createPublisherSession` directly:

https://github.com/zeroc-ice/ice/blob/a2a727827000dd62afca24b7f4d9ded67c68e44a/cpp/src/DataStorm/SessionI.cpp#L1764-L1772

`createPublisherSession` can throw `SessionCreationException` (e.g., when proxy fixup or session lookup fails during shutdown races). The exception propagates into `IceInternal::Timer::run`, which catches it and logs `IceInternal::Timer: uncaught exception: ...` to `consoleErr`. No crash — but it produces alarming stderr noise during otherwise-normal shutdown/retry races.

By contrast, `PublisherSessionI::reconnect` → `createSubscriberSession` doesn't throw on this path.

Suggested fix: catch `SessionCreationException` in `SubscriberSessionI::reconnect` (or inside `createPublisherSession`'s retry entry point) and route it to the normal `retry()` failure accounting instead of letting it escape to the timer thread.

Found during a code review of DataStorm at a2a7278270. Low severity — cosmetic/log noise only.

### LINKED ISSUE #6326: DataStorm: attachTags clears the subscriber's tag map before decoding the replacement tags
`SessionI::attachTags` clears the subscriber's tag map when `initialize` is set, then repopulates it one decode at a time:

https://github.com/zeroc-ice/ice/blob/8e8e65c0d5d3edacb1be9e5ed066ff5f7a0e698d/cpp/src/DataStorm/SessionI.cpp#L317-L326

`TagFactory::decode` runs the application's `Decoder<UpdateTag>`, or the default decoder, which unmarshals the peer's bytes and throws `MarshalException` on malformed input. A throw part-way through leaves the map empty or partially replaced, the dispatch aborts, and nothing retries `attachTags` — the tags are gone for the life of that topic subscriber.

The failure is silent and shows up later as wrong data rather than as an error. An unresolved tag id yields a null `Tag` (see #6310), `TopicI::getUpdater` then falls back to `noOpUpdater`:

https://github.com/zeroc-ice/ice/blob/8e8e65c0d5d3edacb1be9e5ed066ff5f7a0e698d/cpp/src/DataStorm/TopicI.cpp#L630-L640

so every later partial update carrying that tag silently resolves to a copy of the previous value. In a debug build it trips `assert(previous && previous->hasValue())` in `noOpUpdater` instead.

Suggested fix: decode into a local map and swap it in only once every tag has decoded, so a failure leaves the previous tags in place.

Same class as #6308: state committed before a call that can run application code. Related: #6310 (these lookups should use `find()` rather than `operator[]`) and #6213 (what to do when a partial update's tag cannot be resolved).

Found while preparing the fix for #6308 (PR #6322), and cross-checked against the source with an independent Codex review. Not yet human-triaged, and the milestone is left unset for triage.

pepone:
Three small fixes to the DataStorm session, one commit each. All three are failure-path only: with an application
that doesn't throw and a peer that sends decodable bytes, none of them changes what the code does.

### Resolve sample tags with a non-inserting lookup

The three sample-construction paths read `TopicSubscriber::tags` with `map::operator[]`, which inserts a null tag for
every id it is asked about. A sample with no update tag carries the tag id 0, so every such sample left an entry
behind. `TopicSubscriber::findTag` now mirrors the `findKey` written for the same reason. A miss still yields
`nullptr`, so the samples these paths build are unchanged — this deliberately does not pre-empt #6213, which is
about what *should* happen when a partial update's tag can't be resolved.

### Keep a session reconnect failure out of the timer thread

`SubscriberSessionI::reconnect` runs on the Ice timer thread and called `NodeI::createPublisherSession` without
catching `SessionCreationException`. Only two throws are reachable from there: `NodeShutdown`, which must not retry,
and `Internal`, which `NodeI` already handed to `retrySubscriberSessionCreation` before rethrowing. So the exception
carried nothing left to act on and simply escaped into `IceInternal::Timer::run`'s catch-and-print.

The throws are kept — `NodeI::initiateCreateSession` maps them to the `initiateCreateSession` reply — and only this
call site catches. After this change every call to `createPublisherSession` in the tree is inside a catch.

### Skip an update tag the application's decoder cannot decode

`SessionI::attachTags` cleared the tag map and then decoded into it one tag at a time, so a decoder that threw part
way left the subscriber with a truncated set, and with none at all when the peer sent the full set. The throw also
escaped `runWithTopics`, so the remaining subscribers of that topic id got no tags either. The peer only sends the
tags again when the session reconnects, and until then every partial update carrying an unresolved tag resolved to a
copy of the previous value, silently.

Letting the exception escape reaches nobody: `TopicI::forward` invokes the listener session proxies with null
response and exception callbacks ("we don't need to check the result"), and `SessionI::attachTopic` calls
`attachTagsAsync` the same way, so the only effect is a dispatch warning on the receiving node. Each tag is now
decoded under its own catch — a tag the decoder rejects is skipped with a warning, and the tags that decode are
kept. `DataElementI` already handles a throwing updater or decoder this way when resolving a partial update.

A skipped tag stays unresolved, so a partial update carrying it still resolves to the key's current value. That is
#6213, unchanged by this PR.

### Notes

- #6326 is covered in `cpp/test/DataStorm/partial`: an update tag type whose `Decoder` rejects one of the writer's
  two tags, and the reader still applies the partial update published with the tag it can decode. One caveat — the
  tags travel in a single batch whose order follows the writer's `map<shared_ptr<Tag>, Updater>`, so against the
  previous code the test fails only when the undecodable tag happens to be decoded first: 4 runs out of 5 locally,
  green 8 out of 8 with the fix.
- No test for the other two. #6310 has no symptom to observe — both forms hand an identical `nullptr` downstream, so
  a probe would be green before the fix. #6311's only symptom is a line on `consoleErr` from Ice's own timer.
- No changelog fragment. All three are failure-path hardening: reaching any of them takes a buggy custom decoder
  or an encoder/decoder mismatch between processes, not normal operation.
- `SessionI::attachTopic` decodes a topic spec's tags the same way. It doesn't clear first, so a failure there leaves
  a superset rather than destroying the map, and the throw aborts `attachTopic` after the topic is attached and
  before its elements are — the commit-before-application-code class tracked by #6324. Left alone here.

Verified with the full `cpp/DataStorm` suite (11/11) on 838c8c9.

Fixes #6310
Fixes #6311
Fixes #6326

### DIFF
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -314,7 +314,11 @@ SessionI::attachTags(int64_t topicId, ElementInfoSeq tags, bool initialize, cons
                 out << _id << ": attaching tags '[" << tags << "]@" << topicId << "' on topic '" << topic << "'";
             }
 
-            auto decoded = initialize ? map<int64_t, shared_ptr<Tag>>{} : subscriber.tags;
+            // An initializing call replaces the subscriber's tags, so it decodes into a temporary and installs it once
+            // every tag is decoded; the subscriber keeps its current tags until then. Any other call adds to the tags
+            // the subscriber already holds, and decodes into them directly.
+            map<int64_t, shared_ptr<Tag>> replacement;
+            auto& decoded = initialize ? replacement : subscriber.tags;
             for (const auto& tag : tags)
             {
                 try
@@ -323,17 +327,20 @@ SessionI::attachTags(int64_t topicId, ElementInfoSeq tags, bool initialize, cons
                 }
                 catch (const std::exception& ex)
                 {
-                    // The tag factory runs the application's decoder. Skip a tag it can't decode: letting the
-                    // exception escape would drop the tags that did decode, and abort runWithTopics before the
-                    // remaining subscribers of this topic get any. The peer resends its tags on reconnect; until
-                    // then a partial update carrying the skipped tag resolves to the key's current value.
+                    // The tag factory runs the application's decoder. Skip a tag it can't decode: the peer resends
+                    // its tags on reconnect; until then a partial update carrying the skipped tag resolves to the
+                    // key's current value.
                     Warning out(_traceLevels->logger);
                     out << "skipped update tag " << tag.id << " on topic '" << topic << "': the tag could not be "
                         << "decoded:\n"
                         << ex.what();
                 }
             }
-            subscriber.tags = std::move(decoded);
+
+            if (initialize)
+            {
+                subscriber.tags = std::move(replacement);
+            }
         });
 }
 
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -314,15 +314,26 @@ SessionI::attachTags(int64_t topicId, ElementInfoSeq tags, bool initialize, cons
                 out << _id << ": attaching tags '[" << tags << "]@" << topicId << "' on topic '" << topic << "'";
             }
 
-            if (initialize)
-            {
-                subscriber.tags.clear();
-            }
-
+            auto decoded = initialize ? map<int64_t, shared_ptr<Tag>>{} : subscriber.tags;
             for (const auto& tag : tags)
             {
-                subscriber.tags[tag.id] = topic->getTagFactory()->decode(_instance->getCommunicator(), tag.value);
+                try
+                {
+                    decoded[tag.id] = topic->getTagFactory()->decode(_instance->getCommunicator(), tag.value);
+                }
+                catch (const std::exception& ex)
+                {
+                    // The tag factory runs the application's decoder. Skip a tag it can't decode: letting the
+                    // exception escape would drop the tags that did decode, and abort runWithTopics before the
+                    // remaining subscribers of this topic get any. The peer resends its tags on reconnect; until
+                    // then a partial update carrying the skipped tag resolves to the key's current value.
+                    Warning out(_traceLevels->logger);
+                    out << "skipped update tag " << tag.id << " on topic '" << topic << "': the tag could not be "
+                        << "decoded:\n"
+                        << ex.what();
+                }
             }
+            subscriber.tags = std::move(decoded);
         });
 }
 
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -620,7 +620,7 @@ SessionI::initSamples(int64_t topicId, int64_t peerTopicId, DataSamplesSeq initi
                         sample.id,
                         sample.event,
                         key,
-                        subscriber.tags[sample.tag],
+                        subscriber.findTag(sample.tag),
                         sample.value,
                         sample.timestamp));
                 }
@@ -1489,7 +1489,7 @@ SessionI::subscriberInitialized(
                 sample.id,
                 sample.event,
                 key ? key : keyFactory->decode(_instance->getCommunicator(), sample.keyValue),
-                subscriber.tags[sample.tag],
+                subscriber.findTag(sample.tag),
                 sample.value,
                 sample.timestamp));
             assert(samplesI.back()->key);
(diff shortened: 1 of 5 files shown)


externl (approved):
Looks good.

cpp/src/DataStorm/SessionI.cpp
```diff
@@ -314,15 +314,26 @@ SessionI::attachTags(int64_t topicId, ElementInfoSeq tags, bool initialize, cons
...
+                try
+                {
+                    decoded[tag.id] = topic->getTagFactory()->decode(_instance->getCommunicator(), tag.value);
+                }
+                catch (const std::exception& ex)
+                {
+                    // The tag factory runs the application's decoder. Skip a tag it can't decode: letting the
+                    // exception escape would drop the tags that did decode, and abort runWithTopics before the
+                    // remaining subscribers of this topic get any. The peer resends its tags on reconnect; until
+                    // then a partial update carrying the skipped tag resolves to the key's current value.
```
bernardnormier: The middle of this comment describes what the previous code did wrong rather than the present behavior. Suggest keeping only the forward-looking part:
```suggestion
                    // The tag factory runs the application's decoder. Skip a tag it can't decode: the peer resends
                    // its tags on reconnect; until then a partial update carrying the skipped tag resolves to the
                    // key's current value.
```

cpp/src/DataStorm/SessionI.h
```diff
@@ -185,6 +185,16 @@ namespace DataStormI
                 return p == keys.end() ? nullptr : p->second.first;
             }
 
+            /// Returns the local update tag matching a remote tag id, or `nullptr` when this session holds no tag for
+            /// that id. Like findKey, it never inserts: every sample resolves its tag through this method, and a
+            /// sample without an update tag carries the id 0, so an inserting lookup would leave a null entry behind
+            /// for every id it is asked about.
```
bernardnormier: A lookup that doesn't insert is the expected behavior, so it doesn't need defending — and "an inserting lookup would leave a null entry behind" describes the old code. Suggest keeping the contract plus the one non-obvious fact:
```suggestion
            /// Returns the local update tag matching a remote tag id, or `nullptr` when this session holds no tag for
            /// that id. Misses are routine, since a sample without an update tag carries the tag id 0.
```

bernardnormier (approved):
The three code changes look good, and the new test is a nice addition.

However, I would drop the changelog fragment: an update tag decoder that throws signals a severe encoder/decoder mismatch between processes, not something that occurs during normal operations — same category as the two fixes in this PR that you kept out of the CHANGELOG.

Related: I filed #6389 for making the handling of a throwing application decoder uniform — this PR's `attachTags` change follows the rule; the key-decoding paths don't yet.

Issue #6177, opened 2026-07-24
DataStorm: clearHistory=Never with the default sampleCount/sampleLifetime retains every written sample


bernardnormier:
## Summary

A topic writer's sample history is bounded by three settings (per-writer `Config`, or the `DataStorm.Topic.*` defaults):

- `sampleCount` — default `-1` (unlimited)
- `sampleLifetime` — default `0` (disabled)
- `clearHistory` — default `OnAll` (every new sample clears the history)

With the defaults, the history never holds more than the latest sample. But setting `clearHistory = ClearHistoryPolicy::Never` while keeping the other two defaults produces a writer whose history is never trimmed: no event clears it, no count caps it, no age reaps it. Every sample published over the element's lifetime is retained, and memory grows without bound in an otherwise healthy application. The documentation doesn't warn about this: `ClearHistoryPolicy::Never` is described only as "Never clear the sample history."

## This configuration is never needed

The intuitive use case for `Never` — late joiners must receive the full data set — does not require unbounded history. A late-joining reader is initialized with the writer's retained history *plus* the current base of every key not covered by that history, so a writer with the default `OnAll` policy already delivers the complete current state of every key to every late joiner. Unbounded history adds only one thing: replay of the intermediate update *events*.

That event-replay semantic is only safe when the write volume is bounded (a short-lived writer) — and a bounded workload gets the same replay from a finite `sampleCount`. A long-running writer with the unbounded configuration leaks; a short-lived one doesn't need "unlimited". There is no workload that both needs the unbounded form and survives it. (The history is also not durable — it lives only as long as the writer element — so the retained "event log" only exists for the writer's process lifetime anyway.)

## Proposal

Make the statically-unbounded configuration invalid: fail when the merged writer configuration has `clearHistory == Never` and neither `sampleCount` nor `sampleLifetime` sets a bound. The check must apply to the *merged* configuration — the combination can arise from layered defaults (e.g. `DataStorm.Topic.ClearHistory=Never` set on the node with the other properties untouched), so the error surfaces where the merged configuration is first used.

Rejecting a currently-accepted configuration is a behavior change, likely feature-release material.

Side note: `OnAdd` and `OnAllExceptPartialUpdate` can also grow without bound when the clearing event never occurs in the workload (e.g. `OnAdd` with a writer that adds a key once and then only updates it). That is workload-dependent rather than statically detectable, so for those a documentation warning on `ClearHistoryPolicy` is the practical remedy.

[...]

Issue #6309, opened 2026-07-31
DataStorm: a sample-filtered reader cannot resolve partial updates whose base samples the filter excluded


pepone:
## Summary

A sample filter is evaluated on the **writer**, against the **resolved** value of a partial update — but what the
writer sends is still the unresolved delta. A reader whose filter excluded the samples that would have given it a
base for the key therefore receives an update it cannot apply.

## Mechanism

`DataWriterI::publish` resolves a partial update against the writer's own `_lastByKey` **before** anything is sent:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/DataElementI.cpp#L1231-L1250

The updater writes the merged value through `setValue`, which sets `_value`/`_hasValue` but leaves `_encodedValue`
holding the original delta:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/include/DataStorm/InternalT.h#L357-L361

and `encode()` returns that cached delta rather than re-encoding the merged value:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/include/DataStorm/InternalT.h#L379-L386

`KeyDataWriterI::send` marshals the sample through `toSample`, which uses `encode()`, and only then does the
collocated forwarder filter per listener:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/DataElementI.cpp#L1647-L1672

`Listener::matchOne` evaluates the sample filter against the in-memory sample, which by then carries the merged
value:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/DataElementI.h#L91-L104

So the filter decision is made on a value the reader may never have been sent, and the delta goes out with
`event = PartialUpdate` and the update tag intact.

## Two failure modes

**1. No base — the update is dropped.** Every prior `Add`/`Update` for the key was excluded by the filter, so the
reader has no `_lastByKey` entry and discards the sample:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/DataElementI.cpp#L1062-L1077

The only symptom is a trace record, emitted only when data tracing is enabled.

[...]

PR #6345, opened 2026-08-03
Document the DataStorm sample history and sample filter limitations

pepone:
Documentation only — no code change.

### `ClearHistoryPolicy::Never`

The enumerator was documented as "Never clear the sample history" and nothing more. With `sampleCount` and
`sampleLifetime` left at their defaults, nothing else bounds the history either: the `DataStorm.Topic.SampleCount`
and `DataStorm.Topic.SampleLifetime` defaults that `mergeConfigs` applies are `-1` and `0`, and `DataWriterI::publish`
trims only for a count or lifetime greater than zero. An element configured that way accumulates samples for as long
as it lives. The note says so, and points at the two settings that bound it.

### The sample filter / partial update interaction

A writer sends a reader that uses a sample filter only the samples that reader matches, so the reader can receive a
partial update for a key it never received a full value for. It has nothing to resolve the update against and
discards it. Neither `Topic::setSampleFilter` nor either `partialUpdate` overload mentioned this. #6309 tracks the
behavior fix; this documents the current behavior.

### On #6177

#6177 proposes rejecting the merged configuration `clearHistory=Never` with `sampleCount=-1` and `sampleLifetime=0`
outright. I don't think we should:

- It is the only way to express what `Writer::getAll()` documents — "all the written samples kept in the writer
  history".
- On the reader side that history *is* the unread queue that `getAllUnread`/`getNextUnread`/`hasUnread` read. Under
  the `OnAll` default `DataReaderI::queue` clears the queue on every arrival, so a reader that must not drop samples
  has no other option.
- Our own suite uses it, in 17 files under `cpp/test/DataStorm`, including one with the comment "Keep all the samples
  in the history."
- The check can't live where the issue says it must: `Topic::setWriterDefaultConfig` is `noexcept`, so throwing at the
  merge point calls `std::terminate`. `cpp/test/DataStorm/types/Writer.cpp` drives exactly that path with exactly
  that configuration.

A warning log is out for the same reason — it would fire on our own writer suites and on every legitimate use. The
issue's own premise that a late joiner already receives the full current state under the default `OnAll` policy holds
(`getSamples` appends `collectBases`, and `_lastByKey` is maintained before any trimming), which is what makes
documenting it sufficient.

Refs #6177
Refs #6309

### DIFF
--- a/cpp/include/DataStorm/DataStorm.h
+++ b/cpp/include/DataStorm/DataStorm.h
@@ -389,8 +389,9 @@ namespace DataStorm
 
         /// Sets a key filter factory. The given factory function must return a filter function that returns `true` if
         /// the key matches the filter criteria, `false` otherwise.
-        /// Register all key filters before creating any reader or writer for this topic: the set of filters is not
-        /// synchronized, so modifying it once the topic is in use races with the Ice threads that use the topic.
+        /// Register all key filter factories before creating any reader or writer for this topic: the set of
+        /// factories is not synchronized, so modifying it once the topic is in use races with the Ice threads that
+        /// use the topic.
         /// @param name The name of the key filter.
         /// @param factory The filter factory function.
         template<typename Criteria>
@@ -400,13 +401,15 @@ namespace DataStorm
 
         /// Sets a sample filter factory. The given factory function must return a filter function that returns `true`
         /// if the sample matches the filter criteria, `false` otherwise.
-        /// Register all sample filters before creating any reader or writer for this topic: the set of filters is not
-        /// synchronized, so modifying it once the topic is in use races with the Ice threads that use the topic.
-        /// A sample filter interacts with partial updates: a writer sends a reader that uses one only the samples the
-        /// reader matches, so the reader can receive a partial update for a key it never received a full value for. It
-        /// has nothing to resolve that update against and discards it, until it receives a full value for the key. A
-        /// partial update that matches after non-matching samples likewise resolves against the reader's own last
-        /// matching value rather than against the writer's current value.
+        /// Register all sample filter factories before creating any reader or writer for this topic: the set of
+        /// factories is not synchronized, so modifying it once the topic is in use races with the Ice threads that
+        /// use the topic.
+        /// A sample filter interacts with partial updates: a writer sends only the samples a reader's filter matches,
+        /// so the reader can receive a partial update for a key whose full value it never received. The reader has
+        /// nothing to resolve such updates against and discards them until it receives a full value for the key. And
+        /// when the filter rejects some samples for a key but a later partial update matches, the reader applies that
+        /// update to the last value it received for the key — not to the value the writer computed the update
+        /// against — so the reader's value can silently diverge from the writer's.
         /// @param name The name of the sample filter.
         /// @param factory The filter factory function.
         template<typename Criteria>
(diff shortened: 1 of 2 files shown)


externl (approved):

cpp/include/DataStorm/DataStorm.h
```diff
@@ -402,6 +402,11 @@ namespace DataStorm
         /// if the sample matches the filter criteria, `false` otherwise.
         /// Register all sample filters before creating any reader or writer for this topic: the set of filters is not
         /// synchronized, so modifying it once the topic is in use races with the Ice threads that use the topic.
```
bernardnormier: What's registered is the factory, not the filter:
```suggestion
        /// Register all sample filter factories before creating any reader or writer for this topic: the set of
        /// factories is not synchronized, so modifying it once the topic is in use races with the Ice threads that
        /// use the topic.
```
The `setKeyFilter` counterpart (line 392) has the same filters-vs-factories wording and deserves the same edit — it's outside this diff, so no suggestion there.

cpp/include/DataStorm/DataStorm.h
```diff
@@ -402,6 +402,11 @@ namespace DataStorm
...
+        /// A sample filter interacts with partial updates: a writer sends a reader that uses one only the samples the
+        /// reader matches, so the reader can receive a partial update for a key it never received a full value for. It
+        /// has nothing to resolve that update against and discards it, until it receives a full value for the key. A
+        /// partial update that matches after non-matching samples likewise resolves against the reader's own last
+        /// matching value rather than against the writer's current value.
```
bernardnormier:
```suggestion
        /// A sample filter interacts with partial updates: a writer sends only the samples a reader's filter matches,
        /// so the reader can receive a partial update for a key whose full value it never received. The reader has
        /// nothing to resolve such updates against and discards them until it receives a full value for the key. And
        /// when the filter rejects some samples for a key but a later partial update matches, the reader applies that
        /// update to the last value it received for the key — not to the value the writer computed the update
        /// against — so the reader's value can silently diverge from the writer's.
```

cpp/include/DataStorm/DataStorm.h
```diff
@@ -782,6 +787,9 @@ namespace DataStorm
         /// returned function is called: a full value was written for the key and the key was not since removed.
         /// Calling the returned function for a key with no current value is an application error that throws
         /// std::logic_error and publishes nothing.
+        /// A reader that uses a sample filter receives only the samples it matches, so it can have no current value
+        /// for the key even though the writer does; it discards the partial update in that case. See
+        /// Topic::setSampleFilter.
```
bernardnormier:
```suggestion
        /// A reader that uses a sample filter receives only the samples its filter matches: such a reader can lack
        /// a current value for the key even though the writer has one, and it discards partial updates until it
        /// receives a full value for the key. See Topic::setSampleFilter.
```

cpp/include/DataStorm/DataStorm.h
```diff
@@ -848,6 +856,9 @@ namespace DataStorm
...
```
bernardnormier: Same as the other overload:
```suggestion
        /// A reader that uses a sample filter receives only the samples its filter matches: such a reader can lack
        /// a current value for the key even though the writer has one, and it discards partial updates until it
        /// receives a full value for the key. See Topic::setSampleFilter.
```

cpp/include/DataStorm/Types.h
```diff
@@ -48,7 +48,11 @@ namespace DataStorm
         /// Clear the sample history when a new sample which is not a partial update is received.
         OnAllExceptPartialUpdate,
 
-        /// Never clear the sample history.
+        /// Never clear the sample history. The history is then bounded only by Config::sampleCount and
+        /// Config::sampleLifetime, both unbounded by default, so an element configured this way and left with the
+        /// default bounds accumulates samples for as long as it lives: every sample it publishes, for a writer, and
+        /// every sample it receives that the application has not yet read, for a reader. Set Config::sampleCount or
+        /// Config::sampleLifetime on a long-lived element that writes or receives a high volume of samples.
```
bernardnormier:
```suggestion
        /// Never clear the sample history. Only Config::sampleCount and Config::sampleLifetime then bound the
        /// history, and both are unbounded by default: with the defaults, a writer accumulates every sample it
        /// publishes and a reader accumulates every sample the application has not yet read, for as long as the
        /// element lives. Set Config::sampleCount or Config::sampleLifetime on a long-lived element that writes or
        /// receives a high volume of samples.
```

bernardnormier (commented):
The documentation is accurate and worth having, but I don't want this PR to close #6177 — please change "Fixes #6177" to "Refs #6177".

This is a fundamental design issue: changing one config setting must not give you a writer that is guaranteed to leak. Let's continue the discussion on the issue.

PR #6360, opened 2026-08-03
Make the Ruby extension safe under GC heap compaction

### LINKED ISSUE #6327: Ruby: _communicatorMap stores a raw VALUE that GC compaction invalidates
> [!NOTE]
> Raised by an AI cross-review of #6302 and confirmed at runtime with the repro below (Ruby 4.0.6, arm64-darwin).
> It is independent of that PR — it reproduces identically on `main` and on the PR branch.

## Summary

`ruby/src/IceRuby/Communicator.cpp:22-23` keeps the C++ communicator → Ruby wrapper mapping in a plain C++ container
holding a raw `VALUE`:

```cpp
using CommunicatorMap = map<Ice::CommunicatorPtr, VALUE>;
CommunicatorMap _communicatorMap;
```

The wrapper is created at `:166` and looked up by `lookupCommunicator` at `:563`. Nothing tells the GC about that
`VALUE`, so when compaction moves the wrapper the map keeps pointing at the old slot and `lookupCommunicator` returns
whatever now lives there.

The contrast within the same extension is `RubySliceLoader`, which does it correctly: it registers its `VALUE` with
`rb_gc_register_address`, and registered roots are marked and pinned. A raw `VALUE` in a C++ container is neither
marked, pinned, nor updated.

## Repro

```ruby
require 'Ice'

communicators = (0...5).map { Ice.initialize }
GC.verify_compaction_references(expand_heap: true, toward: :empty)

communicators.each_with_index do |communicator, i|
    proxy = Ice::ObjectPrx.new(communicator, "hello:tcp -h 127.0.0.1 -p #{10000 + i}")
    puts "#{i}: #{proxy.ice_getCommunicator().class}"
end

communicators.each(&:destroy)
```

Expected: `Ice::CommunicatorI` five times. Actual, on `main`:

```
issue_repro.rb:8:in 'block in <main>': undefined method 'class' for an instance of Prism::RationalNode (NoMethodError)
```

`ice_getCommunicator` handed back an unrelated object from the vacated slot. An earlier variant of the same script
segfaulted outright:

```
error_highlight/core_ext.rb:47: [BUG] Segmentation fault at 0x0000000000000004
```

`GC.verify_compaction_references` is the documented way to shake out this class of extension bug: it moves every
*movable* object, so the fact that the wrapper moves proves it is not pinned. Plain `GC.compact` did not move it in my
runs — it is heap-layout dependent, not immune.

## Why it matters

The stale `VALUE` is not only returned to user code by `ObjectPrx#ice_getCommunicator`; it is also what
`IceRuby_ObjectPrx_mark` marks on every GC (`ruby/src/IceRuby/Proxy.cpp:23-30`):

```cpp
volatile VALUE communicator = lookupCommunicator((*proxy)->ice_getCommunicator());
assert(!NIL_P(communicator));
rb_gc_mark(communicator);
```

So after a compaction the proxy marks a slot it does not own, while the real wrapper goes unmarked and can be freed
with a live proxy still referencing it. Exposure is any application that calls `GC.compact` or sets
`GC.auto_compact = true`.

**Files:** `ruby/src/IceRuby/Communicator.cpp:22-23,166,563-571`, `ruby/src/IceRuby/Proxy.cpp:23-30`

## Fix

The mapping needs a representation the GC maintains. Two directions, both compatible with the map cleanup added in
#6302:

- Hold the C++ → wrapper mapping in a Ruby-level `ObjectSpace::WeakMap` keyed by the communicator address as an
  `Integer`. GC updates its references across compaction, and weak values keep the wrapper collectible.
- Or give each proxy a direct reference to its communicator wrapper in its own data struct and mark it with
  `rb_gc_mark`, which pins. That also removes the map lookup and the assert from the mark function — though the map
  still needs a compaction-safe form for the window before the first proxy exists.

Note that `rb_gc_register_address` is not usable here: it roots the wrapper, so `IceRuby_Communicator_free` would
never run and the leak fixed by #6302 would come back.

### LINKED ISSUE #6359: Ruby: the Slice type registry stores raw VALUEs that GC compaction invalidates
> [!NOTE]
> Found while writing a regression test for #6327 — same class of bug, much broader scope. Confirmed at runtime
> with the repro below (Ruby 4.0.5, arm64-darwin).

## Summary

The Slice type registry in `ruby/src/IceRuby/Types.cpp` captures Ruby objects from `slice2rb`-generated code at
definition time and stores them as raw `VALUE`s in C++ members the GC cannot see:

- `ClassInfo::rubyClass`, `ClassInfo::typeObj`
- `ProxyInfo::rubyClass`, `ProxyInfo::typeObj`
- `StructInfo::rubyClass`
- `EnumInfo::rubyClass` and every entry of `EnumInfo::enumerators` (`std::map<std::int32_t, VALUE>`)
- `ExceptionInfo::rubyClass`

Nothing marks, pins, or updates these `VALUE`s. The generated classes are defined in Ruby code, so unlike classes
created through `rb_define_class` (which the VM registers as pinned roots), they are referenced only from constants —
and constant references are updated, not pinned, when GC compaction moves an object. After a compaction the registry
points at vacated heap slots, and every marshal/unmarshal that goes through it dispatches on whatever now lives there.

## Repro

Add a compaction to `ruby/test/Ice/proxy/AllTests.rb`, right after the client starts (or run the equivalent in any
program that loads Slice-generated code):

```ruby
GC.verify_compaction_references(expand_heap: true, toward: :empty)
```

The client segfaults a few sections later, e.g. at `testing proxy hierarchy...` — `aPrx == cPrx` dispatches on the
proxy returned by `cPrx.opA(cPrx)`, which `ProxyInfo::unmarshal` created from a stale `rubyClass` for the moved
`Test::APrx`:

```
testing proxy hierarchy... AllTests.rb:836: [BUG] Segmentation fault at 0x0000000000000000

libruby.4.0.dylib        rb_id_table_lookup
libruby.4.0.dylib        search_method0
libruby.4.0.dylib        callable_method_entry_or_negative
libruby.4.0.dylib        rb_callable_method_entry
libruby.4.0.dylib        rb_vm_search_method_slowpath
...
libruby.4.0.dylib        opt_equality
```

`GC.verify_compaction_references` moves every movable object, so the repro is deterministic; plain `GC.compact` or
`GC.auto_compact = true` hit the same bug heap-layout permitting.

## Why it matters

Exposure is the same as #6327: any application that calls `GC.compact` or enables `GC.auto_compact`. After one
compaction the entire marshaling machinery is poisoned — unmarshaled proxies, classes, exceptions, and enumerators
are instances of (or compared against) stale class `VALUE`s, producing segfaults or unrelated objects.

## Fix

Unlike #6327's communicator wrapper, which must remain collectible, these registry entries live for the life of the
process by design — the C++ code already assumes the captured objects are immortal. Registering each captured `VALUE`
with `rb_gc_register_mark_object` at definition time makes that assumption explicit: the object becomes a GC root and
is pinned, so it can neither be collected nor moved.

**Files:** `ruby/src/IceRuby/Types.h:193,213-216,257-259,379-387,419-424,439-445`, `ruby/src/IceRuby/Types.cpp`
(`EnumInfo::EnumInfo`, `StructInfo::StructInfo`, `ClassInfo::create`/`define`, `ProxyInfo::create`/`define`,
`IceRuby_defineException`)

bernardnormier:
Fixes #6327.
Fixes #6359.

The Ruby extension stores raw `VALUE`s in C++ containers and data members the GC cannot see, mark, or update. GC heap compaction (`GC.compact` or `GC.auto_compact = true`) moves objects to new heap slots, leaving those stored `VALUE`s pointing at vacated memory — so the extension would hand out unrelated objects or crash the interpreter.

Two distinct lifetimes require two distinct fixes:

- **`_communicatorMap`** (#6327) — maps each C++ communicator to its Ruby wrapper, and backs both `ObjectPrx#ice_getCommunicator` and the GC mark function of every proxy. The wrapper must remain *collectible* (rooting it would reintroduce the leak fixed by #6302), so the fix is a `dcompact` callback on the wrapper's `rb_data_type_t`: when compaction moves the wrapper, the callback re-points the map entry via `rb_gc_location`.

- **The Slice type registry** (#6359) — `ClassInfo`/`ProxyInfo` (`rubyClass`, `typeObj`), `StructInfo`/`ExceptionInfo` (`rubyClass`), and `EnumInfo` (`rubyClass`, `enumerators`) capture objects from `slice2rb`-generated code at definition time and hold them for the life of the process. Since these captures are already required to be immortal, they are now registered with `rb_gc_register_mark_object`, which roots and pins them.

The new test in `ruby/test/Ice/proxy` forces a full compaction with `GC.verify_compaction_references`, which moves every movable object, making the failure deterministic. Without the first fix it fails on the spot; without the second, the client segfaults in later test sections when unmarshaled proxies dispatch on a moved generated class. The communicator under test is referenced from an array element because an object referenced only from a local variable is pinned by conservative stack marking and never moves.

Verified on Ruby 4.0.5 (arm64-darwin): the full Ruby test suite passes, and a stress run confirms communicator wrappers are still collected once unreferenced (no #6302 regression).

### DIFF
--- a/ruby/test/Ice/proxy/AllTests.rb
+++ b/ruby/test/Ice/proxy/AllTests.rb
@@ -351,8 +351,8 @@ def allTests(helper, communicator)
     others = [Ice.initialize]
     begin
         GC.verify_compaction_references(expand_heap: true, toward: :empty)
-    rescue NotImplementedError, ArgumentError
-        # Compaction is not supported on this platform, or Ruby < 3.2 doesn't accept these keyword arguments.
+    rescue NotImplementedError
+        # Compaction is not supported on this platform.
     end
     otherProxy = Ice::ObjectPrx.new(others[0], "test:default -p 12010")
     test(otherProxy.ice_getCommunicator() == others[0])
--- a/ruby/test/Ice/proxy/AllTests.rb
+++ b/ruby/test/Ice/proxy/AllTests.rb
@@ -344,6 +344,19 @@ def allTests(helper, communicator)
     print "testing ice_getCommunicator... "
     STDOUT.flush
     test(base.ice_getCommunicator() == communicator)
+
+    # The communicator returned by ice_getCommunicator must remain valid after GC heap compaction moves the
+    # communicator object. The new communicator is referenced from an array element because an object referenced
+    # from a local variable is pinned by conservative stack marking and cannot move.
+    others = [Ice.initialize]
+    begin
+        GC.verify_compaction_references(expand_heap: true, toward: :empty)
+    rescue NotImplementedError, ArgumentError
+        # Compaction is not supported on this platform, or Ruby < 3.2 doesn't accept these keyword arguments.
+    end
+    otherProxy = Ice::ObjectPrx.new(others[0], "test:default -p 12010")
+    test(otherProxy.ice_getCommunicator() == others[0])
+    others[0].destroy
     puts "ok"
 
     print "testing proxy methods... "
--- a/changelog.d/ruby/gc-compaction.md
+++ b/changelog.d/ruby/gc-compaction.md
@@ -1,0 +1,3 @@
+- Fixed a bug where GC heap compaction (`GC.compact` or `GC.auto_compact = true`) broke the Ice extension: after a
+  compaction, the proxies, classes, exceptions, and enumerators returned by Ice invocations could be instances of
+  unrelated classes, and the Ruby interpreter could crash.
--- a/ruby/BUILDING.md
+++ b/ruby/BUILDING.md
@@ -36,7 +36,7 @@ This build compiles Ice for Ruby directly from the source tree and requires a pr
 
 ### Prerequisites
 
-1. **Ruby 3.0 or higher**
+1. **Ruby 3.4 or higher**
 
 2. **Ice for C++ source build**
 
--- a/ruby/ice.gemspec
+++ b/ruby/ice.gemspec
@@ -30,7 +30,7 @@ eos
                      Dir['scripts/*']
   spec.homepage    = 'https://zeroc.com'
   spec.license     = 'GPL-2.0-only'
-  spec.required_ruby_version = '>= 3.0.0'
+  spec.required_ruby_version = '>= 3.4.0'
   spec.require_paths = ['lib', 'dist/lib']
   spec.extensions = ['extconf.rb']
   spec.rdoc_options = ['--exclude=ext/IceRuby/.*\.o$' '--exclude=IceRuby\.(bundle|so)$' '--exclude=scripts/slice2rb$']
--- a/ruby/src/IceRuby/Communicator.cpp
+++ b/ruby/src/IceRuby/Communicator.cpp
@@ -36,11 +36,24 @@ IceRuby_Communicator_free(void* p)
     delete communicator;
 }
 
+extern "C" void
+IceRuby_Communicator_compact(void* p)
+{
+    // Keep the map entry pointing at the wrapper's current location when GC compaction moves it.
+    auto communicator = static_cast<Ice::CommunicatorPtr*>(p);
+    CommunicatorMap::iterator q = _communicatorMap.find(*communicator);
+    if (q != _communicatorMap.end())
+    {
+        q->second = rb_gc_location(q->second);
+    }
+}
+
 static const rb_data_type_t IceRuby_CommunicatorType = {
     .wrap_struct_name = "Ice::Communicator",
     .function =
         {
             .dfree = IceRuby_Communicator_free,
+            .dcompact = IceRuby_Communicator_compact,
         },
     .flags = RUBY_TYPED_FREE_IMMEDIATELY,
 };
--- a/ruby/src/IceRuby/Operation.cpp
+++ b/ruby/src/IceRuby/Operation.cpp
@@ -25,7 +25,7 @@ namespace IceRuby
     class ParamInfo final : public UnmarshalCallback
     {
     public:
-        void unmarshaled(VALUE, VALUE, void*) final;
+        void unmarshaled(VALUE, VALUE, void*, StreamUtil*) final;
 
         TypeInfoPtr type;
         bool optional;
@@ -146,7 +146,7 @@ IceRuby::Operation::~Operation() = default;
 // ParamInfo implementation.
 //
 void
-IceRuby::ParamInfo::unmarshaled(VALUE val, VALUE target, void* closure)
+IceRuby::ParamInfo::unmarshaled(VALUE val, VALUE target, void* closure, StreamUtil*)
 {
     assert(TYPE(target) == T_ARRAY);
     static_assert(sizeof(long) == sizeof(void*), "long and void* must have the same size");
--- a/ruby/src/IceRuby/Types.cpp
+++ b/ruby/src/IceRuby/Types.cpp
@@ -1836,7 +1836,7 @@ IceRuby::DictionaryInfo::unmarshal(
             {
                 // For string keys create a frozen string to ensure that the key used
                 // in the map matches the one keep in the closure
-                keyCB->key = rb_str_new_frozen(keyCB->key);
+                keyCB->key = callRuby(rb_str_new_frozen, keyCB->key);
             }
             callRuby(rb_hash_aset, hash, keyCB->key, Qnil);
 
--- a/ruby/src/IceRuby/Types.cpp
+++ b/ruby/src/IceRuby/Types.cpp
@@ -161,6 +161,11 @@ IceRuby::StreamUtil::StreamUtil() = default;
 
 IceRuby::StreamUtil::~StreamUtil()
 {
+    for (VALUE& value : _pinned)
+    {
+        rb_gc_unregister_address(&value);
+    }
+
     //
     // Make sure we break any cycles among the ValueReaders in preserved slices.
     //
@@ -195,6 +200,13 @@ IceRuby::StreamUtil::add(const shared_ptr<ValueReader>& reader)
 }
 
 void
+IceRuby::StreamUtil::pin(VALUE value)
+{
+    _pinned.push_back(value);
+    rb_gc_register_address(&_pinned.back());
+}
+
+void
 IceRuby::StreamUtil::updateSlicedData()
 {
     for (set<shared_ptr<ValueReader>>::iterator p = _readers.begin(); p != _readers.end(); ++p)
@@ -216,11 +228,13 @@ IceRuby::StreamUtil::setSlicedDataMember(VALUE obj, const Ice::SlicedDataPtr& sl
     {
         _slicedDataType = callRuby(rb_path2class, "Ice::SlicedData");
         assert(!NIL_P(_slicedDataType));
+        rb_gc_register_mark_object(_slicedDataType);
     }
     if (_sliceInfoType == Qnil)
     {
         _sliceInfoType = callRuby(rb_path2class, "Ice::SliceInfo");
         assert(!NIL_P(_sliceInfoType));
+        rb_gc_register_mark_object(_sliceInfoType);
     }
 
     volatile VALUE sd = callRuby(rb_class_new_instance, 0, static_cast<VALUE*>(0), _slicedDataType);
@@ -1803,6 +1817,15 @@ IceRuby::DictionaryInfo::unmarshal(
                 keyCB->key = rb_str_new_frozen(keyCB->key);
             }
             callRuby(rb_hash_aset, hash, keyCB->key, Qnil);
+
+            //
+            // The closure below holds a raw VALUE copy of the key until the value is
+            // unmarshaled; pin the key so that GC compaction cannot move it in the
+            // meantime.
+            //
+            StreamUtil* util = reinterpret_cast<StreamUtil*>(is->getClosure());
+            assert(util);
+            util->pin(keyCB->key);
         }
         //
         // The callback will set the dictionary entry with the unmarshaled value,
@@ -2622,7 +2645,13 @@ IceRuby::ReadValueCallback::ReadValueCallback(
       _target(target),
       _closure(closure)
 {
+    //
+    // Mark the target as in use for the lifetime of this wrapper.
+    //
+    rb_gc_register_address(&_target);
 }
+
+IceRuby::ReadValueCallback::~ReadValueCallback() { rb_gc_unregister_address(&_target); }
 
 void
 IceRuby::ReadValueCallback::invoke(const shared_ptr<Ice::Value>& p)
--- a/ruby/src/IceRuby/Types.cpp
+++ b/ruby/src/IceRuby/Types.cpp
@@ -40,6 +40,11 @@ static ProxyInfoMap _proxyInfoMap;
 typedef map<string, ExceptionInfoPtr, std::less<>> ExceptionInfoMap;
 static ExceptionInfoMap _exceptionInfoMap;
 
+namespace
+{
+    StreamUtil* streamUtil(Ice::InputStream* is) { return reinterpret_cast<StreamUtil*>(is->getClosure()); }
+}
+
 namespace IceRuby
 {
     class InfoMapDestroyer
@@ -157,14 +162,11 @@ addExceptionInfo(const string& id, const ExceptionInfoPtr& info)
 VALUE IceRuby::StreamUtil::_slicedDataType = Qnil;
 VALUE IceRuby::StreamUtil::_sliceInfoType = Qnil;
 
-IceRuby::StreamUtil::StreamUtil() = default;
+IceRuby::StreamUtil::StreamUtil() : _held(Qnil) { rb_gc_register_address(&_held); }
 
 IceRuby::StreamUtil::~StreamUtil()
 {
-    for (VALUE& value : _pinned)
-    {
-        rb_gc_unregister_address(&value);
-    }
+    rb_gc_unregister_address(&_held);
 
     //
     // Make sure we break any cycles among the ValueReaders in preserved slices.
@@ -199,11 +201,22 @@ IceRuby::StreamUtil::add(const shared_ptr<ValueReader>& reader)
     _readers.insert(reader);
 }
 
-void
-IceRuby::StreamUtil::pin(VALUE value)
+long
+IceRuby::StreamUtil::hold(VALUE value)
 {
-    _pinned.push_back(value);
-    rb_gc_register_address(&_pinned.back());
+    if (NIL_P(_held))
+    {
+        _held = callRuby(rb_ary_new);
+    }
+    callRuby(rb_ary_push, _held, value);
+    return RARRAY_LEN(_held) - 1;
+}
+
+VALUE
+IceRuby::StreamUtil::getHeldValue(long index) const
+{
+    assert(!NIL_P(_held) && index >= 0 && index < RARRAY_LEN(_held));
+    return RARRAY_AREF(_held, index);
 }
 
 void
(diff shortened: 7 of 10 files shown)


externl (commented):
Two more of the same kind were missed: `StreamUtil::_slicedDataType` and `_sliceInfoType` (`Types.cpp:157-158`, filled in at `:215-223` from `rb_path2class`). They are cached for the life of the process and nothing registers them, so compaction moves `Ice::SlicedData` / `Ice::SliceInfo` and the statics point at vacated slots — `rb_class_new_instance(..., _slicedDataType)` then builds an object from garbage. One line each where they are set; `rb_gc_register_address` matches what the file already does at `:2385`, `:2486` and `:2773`.

The new test won't catch it — `Ice/proxy` never unmarshals preserved slices.

A question on scope, since I could not settle it from the source. Both fixes here cover VALUEs the extension holds indefinitely. Marshaling also parks VALUEs in C++ for the length of one operation: `ValueMap` and `PrintObjectHistory::objects` are keyed by VALUE, and `ReadValueCallback::_target` and `DictionaryInfo::KeyCallback::key` hold one across a delayed callback. Unmarshaling allocates, allocation can trigger a GC, and under `auto_compact` that GC can compact — so is the marshaling path relying on conservative stack scanning to pin those, or is this a case you have already ruled out?

Nit: the test's `"test:default -p 12010"` hardcodes a port the harness doesn't use (`Test.BasePort` is 14000-based). Harmless since the proxy is never invoked, but misleading.

bernardnormier:
Thanks — all addressed in 1666ffc.

**The two statics**: now registered right after the lazy `rb_path2class` lookup. I used `rb_gc_register_mark_object` rather than `rb_gc_register_address` for consistency with the other immortal registry captures in this PR; for process-lifetime statics the two are equivalent.

**The scope question**: good instinct — the answer is split.

The marshaling path is safe, but via pinning rather than conservative stack scanning: every `ValueMap` insertion is paired with constructing a `ValueWriter`, whose constructor calls `rb_gc_register_address(&_object)`, and registered addresses are marked with pinning — so map keys cannot move while their writer lives. `PrintObjectHistory` is benign: a cycle's ancestor is always a `VALUE` argument in a live C++ frame (conservatively pinned), so cycle detection holds; the worst case is a shared DAG node printed twice.

The unmarshaling side did have real gaps, close to where you pointed:

- `ReadValueCallback::_target` was unregistered. Class-instance targets happen to be pinned by `ValueReader`'s registration, but array, hash, and struct targets receiving deferred patches are only heap-reachable — movable. It now registers `_target` in its constructor and unregisters in its destructor.
- Dictionary keys — with a twist: `KeyCallback::key` itself is only live per loop iteration (stack-covered). The long-lived holder is the raw `VALUE` smuggled through the patch *closure* pointer, read back when the value class is patched. The existing `hash[key] = Qnil` pre-insert keeps the key alive but not in place. Keys are now pinned via a new `StreamUtil::pin()` (register/unregister scoped to the stream's lifetime, including error paths).

To exercise this, the Ice/objects test's custom slice loader can now call `GC.verify_compaction_references(expand_heap: true, toward: :empty)` on every instantiation — Ruby code running mid-unmarshal, between container creation and patch application — and re-runs `getB1`, `opValueMap`, and `opM` under that flag. Reverting the pinning fixes makes it fail at the `opM` struct-key lookup. Interestingly, string keys survive compaction (the frozen copy ends up effectively pinned), so the struct-keyed `LMap` is what actually catches the stale-key corruption.

**The port nit**: `Test.BasePort` defaults to 12010 (`ruby/test/TestHelper.rb`), and this file already hardcodes 12010 throughout the stringification tests, so I left it as is.

ruby/test/Ice/proxy/AllTests.rb
```diff
@@ -344,6 +344,19 @@ def allTests(helper, communicator)
...
+
+    # The communicator returned by ice_getCommunicator must remain valid after GC heap compaction moves the
+    # communicator object. The new communicator is referenced from an array element because an object referenced
+    # from a local variable is pinned by conservative stack marking and cannot move.
+    others = [Ice.initialize]
+    begin
+        GC.verify_compaction_references(expand_heap: true, toward: :empty)
+    rescue NotImplementedError, ArgumentError
+        # Compaction is not supported on this platform, or Ruby < 3.2 doesn't accept these keyword arguments.
+    end
```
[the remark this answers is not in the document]
bernardnormier: The fix doesn't depend on the compaction check running, and the new Ice/objects test uses the same guard — I'd rather keep one convention (Ruby 3.2+ kwargs, skip on older Rubies) in both tests than add a legacy-keyword retry branch.

ruby/test/Ice/proxy/AllTests.rb
```diff
@@ -344,6 +344,19 @@ def allTests(helper, communicator)
...
     test(base.ice_getCommunicator() == communicator)
+
+    # The communicator returned by ice_getCommunicator must remain valid after GC heap compaction moves the
+    # communicator object. The new communicator is referenced from an array element because an object referenced
+    # from a local variable is pinned by conservative stack marking and cannot move.
+    others = [Ice.initialize]
+    begin
+        GC.verify_compaction_references(expand_heap: true, toward: :empty)
+    rescue NotImplementedError, ArgumentError
+        # Compaction is not supported on this platform, or Ruby < 3.2 doesn't accept these keyword arguments.
```
pepone: We list 3.4 as the minimal supported platform, we can probably remove this

pepone (commented):
The correctness work is right and the tests are good. One thing I'd change before merging: the per-item registrations in the second commit are expensive.

## Verified

Ruby 4.0.5, arm64-darwin, comparing this branch against its merge base.

`rb_gc_register_mark_object` does pin — `ObjectSpace.dump` addresses across `GC.verify_compaction_references(expand_heap: true, toward: :empty)`:

| | main | this branch |
|---|---|---|
| `Test::B` (Slice class) | **moved** | pinned |
| `Test::StructKey` | **moved** | pinned |
| `Test::EBase` | **moved** | pinned |
| plain Ruby `class Foo` | moved | moved |

The static sweep is complete. Every other static class `VALUE` in the extension — `_proxyClass`, `_endpointClass`, `_connectionClass`, `_loggerClass`, `_propertiesClass`, `_operationClass`, `_implicitContextClass`, `iceModule`, `_typeInfoClass`, `_exceptionInfoClass` and the `*InfoClass` ones — comes from `rb_define_class_under`/`rb_define_module`, which roots and pins through `rb_vm_add_root_module`; `Ice::ObjectPrx` and the `Ice` module never move on either build. `_slicedDataType`/`_sliceInfoType` were the only two obtained by lookup, so that gap is now closed.

Your marshaling-side argument holds: both `ValueMap` insertions (`Types.cpp:358`, `:2011`) are immediately preceded by a `ValueWriter` for the same `VALUE`, and its constructor registers and therefore pins the key.

`IceRuby_Communicator_compact` is the right mechanism — the wrapper stays collectible, so #6302 does not come back, and it needs no lock.

12 of 13 Ruby suites pass; `Ice/optional` fails identically on main here (stale prebuilt C++ server), unrelated to this PR. Compacting from inside the slice loader mid-unmarshal is a good test.

Also, your reply on the port is right: `Test.BasePort` defaults to 12010 (`ruby/test/TestHelper.rb:20,43`).

## The registrations are O(N) per operation

`StreamUtil::pin` (`Types.cpp:1817-1827`) registers one address per dictionary key, and the `ReadValueCallback` constructor (`:2645-2653`) one per pending class patch. Ruby's registered-address list is walked on every GC and searched linearly on every unregister — it is meant for a handful of process-lifetime statics.

`opValueMap` — `dictionary<string, Value>`, loopback to the C++ `Ice/objects` server, 3 runs each, within 3%:

| n | main | this branch | |
|---|---|---|---|
| 1,000 | 0.003 s | 0.011 s | 3.7× |
| 4,000 | 0.024 s | 0.151 s | 6.3× |
| 16,000 | 0.291 s | 2.19 s | 7.5× |
| 32,000 | 0.848 s | 8.36 s | 9.9× |

`opValueSeq` — `sequence<Value>`, which exercises `ReadValueCallback` but not `pin` — 16k: 0.275 s → 0.685 s; 32k: 0.693 s → 2.279 s. Both registrations contribute, and the factor grows with n. It is the same pathology #6365 just fixed from the other direction.

Suggestion: hold the values in one Ruby Array per `StreamUtil`, registered once, and recover them by index. Array elements are updated across compaction and track the moved object — I checked — so this is more correct than pinning, since it lets the objects move, and costs one registration per stream instead of N. The part that needs your judgement is that an index alone does not fit `DictionaryInfo::unmarshaled`'s current signature, so the closure has to identify the owning array as well.

## Two smaller things

`ReadValueCallback` should delete its copy and move operations. It now has a user-declared destructor calling `rb_gc_unregister_address(&_target)` but an implicit copy constructor that copies `_target` without registering it — the shape #6365 just fixed in `ExceptionReader`. Only constructed through `make_shared` today, so it is latent; `StreamUtil` has the same exposure now that it has a destructor. Both disappear if you take the array approach.

`PrintObjectHistory::objects` stays a `std::map<VALUE,int>` keyed by movable objects. Your read is right that the worst case is a shared DAG node printed twice, but it is a residual compaction gap — worth a follow-up issue rather than anything here.

externl (dismissed):
Looks good, and thanks for chasing the unmarshaling side — the `ReadValueCallback` and dictionary-key gaps are better finds than the statics I pointed at. `deque` is the right call for `_pinned`, since `push_back` preserves references to existing elements. My port nit was wrong: `Test.BasePort` does default to 12010 in `ruby/test/TestHelper.rb`.

One consistency point: `ReadValueCallback` now registers a GC root tied to `&_target`, but doesn't delete copy and move. `ValueWriter` and `ValueReader` delete all four at `Types.h:464-468`, with the comment that says exactly why — and `ReadValueCallback` is shared_ptr-only in the same way.

bernardnormier:
Reworked in 24565db as suggested — thanks for the measurements.

**The holding array.** Each `StreamUtil` now registers a single Ruby array (one `rb_gc_register_address` per stream) and parks VALUEs in it with `hold(VALUE) -> long`; `getHeldValue(long)` reads back the current reference after any compaction. On the signature question: rather than making the closure identify the owning array, I changed the signature — `UnmarshalCallback::unmarshaled` now takes the stream's `StreamUtil*` as a fourth argument. `DictionaryInfo::unmarshal` passes the key's index through the closure when the value type uses classes, and the raw key as before when it doesn't (that path completes while the loop frame is live, so the key is stack-pinned); `DictionaryInfo::unmarshaled` branches on the same `usesClasses()` predicate to decode. `ReadValueCallback` stores the util and the target's index.

**Copy/move.** With the GC-root registration gone, `ReadValueCallback` has no destructor and is back to rule of zero, so there's nothing to delete — the concern dissolves rather than needing the four deletions. `StreamUtil`, which still registers `&_held`, now deletes copy and move with the same comment style as `ValueWriter`/`ValueReader`.

All 19 Ruby test suites pass, and I verified the compaction test is still load-bearing: reverting the dictionary path to raw-key passing makes Ice/objects fail at the struct-key lookup. I didn't reproduce the scaled benchmark, but both per-item registrations are gone — unmarshaling now adds one `rb_ary_push` per deferred item and never touches the registered-address list — so it would be great if you could re-run yours.

Also removed the pre-3.2 `ArgumentError` handling from both tests and bumped ruby/BUILDING.md to Ruby 3.4, and filed #6372 for the `PrintObjectHistory` residual gap.

ruby/test/Ice/proxy/AllTests.rb
bernardnormier: Done in 24565db — removed the ArgumentError branch here and in the Ice/objects test (NotImplementedError stays for platforms without compaction), and bumped ruby/BUILDING.md from Ruby 3.0 to 3.4 to match.

externl (approved):
Looks good — the holding array is a better shape than the pinning it replaces, and the `usesClasses()` predicate that picks index-vs-raw is the same member on the same `DictionaryInfo` in both directions, so the two can't drift apart.

One thing: the gemspec bump doesn't belong here. Nothing shipped needs it — `dcompact` and `rb_gc_location` are Ruby 2.7 APIs, well under the existing `>= 3.0.0` floor. The only 3.2-ism is `GC.verify_compaction_references(expand_heap:, toward:)` in the tests, which aren't in the gem and which the previous revision handled with the `rescue ArgumentError` this one dropped. Raising the floor to 3.4 also drops 3.3, still supported upstream — user-facing packaging that wants its own rationale and changelog entry.

externl (approved):
Looks good.

The gemspec bump doesn't belong here. Nothing shipped needs it — `dcompact` and `rb_gc_location` are Ruby 2.7 APIs, well under the existing `>= 3.0.0` floor. The only 3.2-ism is `GC.verify_compaction_references(expand_heap:, toward:)` in the tests, which aren't in the gem and which the previous revision handled with the `rescue ArgumentError` this one dropped. Raising the floor to 3.4 also drops 3.3, still supported upstream — user-facing packaging that wants its own rationale and changelog entry.

pepone:
Ruby 3.4 requirement was documented in https://docs.zeroc.com/ice/3.8/cpp/supported-platforms-for-ice-3-8-2

We should fix gemspec, but that could be in a follow up PR.

pepone (approved):
Re-ran the benchmark on 5ca3a1d. The regression is gone — alternating A/B against the merge base, same server, two rounds:

| | merge base | this branch |
|---|---|---|
| `opValueMap` n=1,000 | 0.003 s | 0.003 s |
| `opValueMap` n=4,000 | 0.021–0.022 s | 0.022 s |
| `opValueMap` n=16,000 | 0.232–0.233 s | 0.235 s |
| `opValueSeq` n=16,000 | 0.276 s | 0.277 s |
| `opValueMap` n=32,000 | 1.014–1.046 s | 1.042–1.057 s |
| `opValueSeq` n=32,000 | 0.954–0.960 s | 0.964–0.966 s |

Within 1–2%, which is noise here. The previous revision was 9.9× at n=32,000. I alternated builds rather than reusing my earlier base figures because the absolute numbers drift on this machine between sessions.

I also checked the compaction test is load-bearing on both halves, not just the one you mentioned: reverting the dictionary key path to raw-`VALUE` passing fails `Ice/objects`, and so does reverting `ReadValueCallback` to a raw `_target`. Full suite is 12/13 here, with `Ice/optional` failing identically on the merge base for an unrelated reason in my setup.

The rework itself reads correct to me. `usesClasses()` is a pure function of the type graph on a `const valueType`, so the encode and decode sides can't disagree under nesting, recursion or optional framing. The signature change reached every implementor, and both top-level unmarshal entry points install the `StreamUtil` before anything dereferences it. `ReadValueCallback`'s raw `_util` is safe: `_callbacks` takes ownership before the stream gets its non-owning pointer, and `InputStream` destruction only destroys patch records rather than invoking them. Holding array retention is O(holds) and never cleared, which is new against main but no longer-lived than the roots the previous revision registered.

One thing worth fixing while you're here: `rb_str_new_frozen` at `Types.cpp:1839` is called without `callRuby`, and it is the only allocating Ruby call in the file that isn't. That matters because `callRuby` puts the `rb_protect` boundary *inside* the C++ frames, so a raise comes back as a `RubyException` and destructors run; a bare call longjmps straight to MRI's tag, past every C++ frame, and no destructor runs. `~StreamUtil` would then be skipped and `&_held` — a stack address — would stay on the global registered list with the frame reused under it.

The line predates this PR, and it only raises on allocation failure, so this is a nit rather than a defect. What the PR changes is the consequence: skipping `~StreamUtil` used to leak some `ValueReader` cycles, and now it also leaves a dangling GC root, which is the #6365 failure mode. One-word fix while the code is fresh.

bernardnormier:
@externl The minimum supported Ruby since Ice 3.8.0 is Ruby 3.4 (see the supported platforms page pepone linked above). The gemspec's `>= 3.0.0` was stale metadata predating that, not a support statement this PR is tightening — so I'm keeping the 3.4 floor here rather than splitting it out.

bernardnormier:
Thanks for re-running the benchmark and for checking both revert directions. Nit fixed in 7e6de03 — the `rb_str_new_frozen` call now goes through `callRuby`.

PR #6361, opened 2026-08-03
Fix `slice2py` ImportVisitor to Check Metadata on Fields

### LINKED ISSUE #6263: slice2py: sequence metadata on a data member doesn't emit the runtime factory import
Sequence metadata is accepted both on the sequence *definition* and on a *type reference* (`MetadataApplicationContext::DefinitionsAndTypeReferences`). When it is applied to a data member, `slice2py` emits the custom field factory but not the import that factory needs, so constructing the type raises `NameError`.

```slice
module X
{
    sequence<short> ShortSeq;
    ["python:array.array"] sequence<short> ShortSeqDef;

    struct AtMember { ["python:array.array"] ShortSeq a; }      // metadata on the type reference
    struct AtDefinition { ShortSeqDef a; }                      // metadata on the definition
}
```

`X/AtDefinition.py` gets `from array import array`. `X/AtMember.py` does not, although both emit the same field:

```python
a: array[int] = field(default_factory=lambda: array('h'))
```

```
>>> AtDefinition()
AtDefinition(a=array('h'))
>>> AtMember()
NameError: name 'array' is not defined. Did you forget to import 'array'?
```

pyright reports it too, as `"array" is not defined (reportUndefinedVariable)`.

`ImportVisitor::visitDataMember` calls `addRuntimeImportForSequence(sequence, parent)` without the member's metadata ([`PythonUtil.cpp:644`](https://github.com/zeroc-ice/ice/blob/main/cpp/src/slice2py/PythonUtil.cpp#L644)), so `getSequenceMetadata` sees only the definition's metadata and `directive` comes back empty — none of the factory-import branches run. The operation call sites do pass it (`PythonUtil.cpp:718`, `PythonUtil.cpp:734`); `CodeVisitor::visitDataMember` also uses the member metadata when it writes the field, which is why the factory is emitted but the import is not.

Same for `["python:numpy.ndarray"]` and `["python:memoryview:...]` on a data member.

InsertCreativityHere:
Previously, `slice2py` was not taking local metadata on fields into account when generating imports for fields, now it does!
Also realized while reviewing that the `forMarshaling` field on a related function was unused. We just forwarded it along but never actually checked it...

Fixes #6263.


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -619,15 +619,15 @@ Slice::Python::ImportVisitor::visitDataMember(const DataMemberPtr& p)
     // import for type hints.
     if (auto sequence = dynamic_pointer_cast<Sequence>(type))
     {
-        addRuntimeImportForSequence(sequence, parent);
+        addRuntimeImportForSequence(sequence, parent, p->getMetadata());
```
InsertCreativityHere, line 622: The fix for #6263.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -619,15 +619,15 @@ Slice::Python::ImportVisitor::visitDataMember(const DataMemberPtr& p)
...
     }
     else if (dynamic_pointer_cast<Struct>(type) || dynamic_pointer_cast<Enum>(type))
     {
         addRuntimeImport(type, parent);
     }
     else
     {
-        addTypingImport(type, parent, true);
```
InsertCreativityHere, line 630: This trailing bool parameter was dead code.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -706,8 +704,7 @@ Slice::Python::ImportVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
             {
                 addTypingImport("collections.abc", "Mapping", p);
             }
-            addTypingImport(ret, p, false);
-            addTypingImport(ret, p, true);
```
InsertCreativityHere, line 710: Since the bool was dead, we were just calling this function twice for no reason.
One call is sufficient.

cpp/src/slice2py/PythonUtil.h
```diff
@@ -291,11 +291,11 @@ namespace Slice::Python
         /// Add the runtime imports for the given Sequence definition.
         /// @param sequence The Sequence definition being imported.
         /// @param source The Slice definition that requires the import.
-        /// @param localMetadata Any additional metadata associated with the import. Such has parameter metadata.
+        /// @param localMetadata Any additional metadata associated with the import.
         void addRuntimeImportForSequence(
             const SequencePtr& sequence,
             const ContainedPtr& source,
-            const MetadataList& localMetadata = MetadataList());
```
InsertCreativityHere, line 298: Now we always pass this in. And we always should be passing it in, so I removed the default.

cpp/src/slice2py/PythonUtil.h
```diff
@@ -291,11 +291,11 @@ namespace Slice::Python
         /// Add the runtime imports for the given Sequence definition.
         /// @param sequence The Sequence definition being imported.
         /// @param source The Slice definition that requires the import.
-        /// @param localMetadata Any additional metadata associated with the import. Such has parameter metadata.
```
InsertCreativityHere, line 294: These extra words added nothing and were grammatically wrong to boot.

externl (approved):
Looks good.

pepone (approved):
Verified the fix: built `slice2py` from this branch and from `main` and diffed the generated code.

Red → green on all three directives applied to a field (struct, class and exception):

| field metadata | main | this PR |
| --- | --- | --- |
| `["python:array.array"]` | `NameError: name 'array' is not defined` | `AtMember(a=array('h'))` |
| `["python:numpy.ndarray"]` | `NameError: name 'numpy' is not defined` | `NumpyAtMember(a=array([], dtype=int32))` |
| `["python:memoryview:...]` | `NameError: name 'myBoolSeq' is not defined` | `MvAtMember(a=array('b'))` |

It also correctly drops a stale import: a field with `["python:list"]` over a sequence defined `["python:array.array"]` no longer emits an unused `from array import array`.

And `forMarshaling` was indeed dead — running both binaries over the same 79 `.ice` files (`slice/` + `python/test/`) gives byte-identical output, stderr and exit code.

Two asks.

**1. `PythonUtil.cpp:2351` has the same bug**, one line away. The docstring `Attributes` section drops the field metadata:

```c++
<< typeToTypeHintString(field->type(), field->isOptional(), p, false);
```

so the docstring contradicts the annotation right below it:

```python
    Attributes
    ----------
    a : list[int]                                              # wrong
    ...
    a: array[int] = field(default_factory=lambda: array('h'))
```

Every other docstring call site passes the metadata. Please add `, field->getMetadata()`.

**2. `typecode` belongs in the existing `words` array**, not a new `ignoreWords` section — that's where every previous addition went.

Nit: the test only exercises `python:array.array` on a field. The memoryview path was equally broken, and struct `M` doesn't catch it because `BoolSeq2` carries the same metadata at its definition — pointing `M`'s member at `BoolSeq1` covers it for free.

Unrelated and pre-existing, I'll open an issue: a dictionary whose value type is a custom-mapped sequence generates `dict[str, array[int]]` and `Sequence[int] | Buffer` with no import at all — pyright flags `"array" is not defined` and `"Buffer" is not defined`. Same root cause, `addTypingImport` recurses into the key and value types without consulting `getSequenceMetadata`.

InsertCreativityHere:
1) Yes, a good catch. I added the local-metadata to this call-site. I checked and indeed, everywhere else where we call this function we were passing metadata (excluding nested invocations which are different).

2) This is just an artifact of us not properly using 'cspell' in this repository, the distinction between 'words' and 'ignoreWords' is that 'words' should really just be for actual real words, not covered for one-reason-or-another. Whereas ignoreWords is for things that aren't real words, but cspell should tolerate anyways. Functionally: words are used to suggest auto-completes, ignoreWords isn't. And since `typecode` is neither a real-world, nor something we use often enough to warrant suggestion, it belongs in 'ignoreWords'. Realistically, there are probably _many_ words which should be moved from `words` to `ignoreWords`, but that's a separate job that this PR. This PR just didn't want to continue worsening an already existent problem.

----

I changed the `M` struct to use `BoolSeq1` instead of `BoolSeq2`. Also unrelated to this PR, but no reason not to fix now I guess. Not clear if there was a reason to repeating the metadata at both sites; None that I can see...

As for the remaining dictionary recursion issue, sure, go ahead and open an issue.
I'm going to open a follow-up PR to deal with the existing sequence-recursion issue anyways.

PR #6362, opened 2026-08-03
Read the IceBT buffer-size properties before creating sockets

### LINKED ISSUE #6351: Bluetooth socket leaks when IceBT.RcvSize or IceBT.SndSize has an invalid value
IceBT has the same defect that #6288 reported for TCP: the buffer-size properties are read while a socket is in flight with no owner, so an invalid `IceBT.RcvSize` or `IceBT.SndSize` value leaks the socket. Found by @pepone while reviewing the TCP fix (see the approval review on the fix PR for #6288).

### C++

`IceBT::AcceptorI::newConnection()` (`cpp/src/IceBT/AcceptorI.cpp`) hands the accepted fd to `IceBT::StreamSocket`, whose `init()` reads `IceBT.RcvSize`/`IceBT.SndSize` before anything owns the fd (`cpp/src/IceBT/StreamSocket.cpp`):

```cpp
void
IceBT::StreamSocket::init(SOCKET fd)
{
    IceInternal::setBlock(fd, false);

    int32_t rcvSize = _instance->properties()->getIcePropertyAsInt("IceBT.RcvSize");
    int32_t sndSize = _instance->properties()->getIcePropertyAsInt("IceBT.SndSize");

    setBufferSize(fd, rcvSize, sndSize);
}
```

A `PropertyException` thrown there propagates to the caller in `Engine.cpp`, which swallows it:

```cpp
try
{
    newConnection(fd->v);
}
catch (...)
{
    // Ignore.
}
```

So with an invalid value configured, the server leaks one fd per inbound Bluetooth connection, with no diagnostic at all. `StreamSocket::setFd()` has the same ordering on the outgoing path.

### Java (Android)

`AcceptorI.accept()` pops the accepted `BluetoothSocket` off `_pending` and passes it to `TransceiverI`, whose `init()` reads the same two properties (`java/src/com.zeroc.icebt/src/main/java/com/zeroc/IceBT/TransceiverI.java`); a `PropertyException` thrown there leaves the popped socket with no owner.

### Suggested fix

Same shape as the TCP fix: read the properties in the acceptor/connector constructors, before any socket exists, and pass the validated sizes through — so an invalid value fails at object adapter creation (server side) or connection establishment (client side) instead of leaking per-connection.

bernardnormier:
Fixes #6351.

The `IceBT.RcvSize`/`IceBT.SndSize` properties were read while a socket was in flight with no owner, so an invalid value leaked the socket on every connection — same defect the TCP transports had before #6335, with the same fix shape: read the properties before any socket exists, via a `BTBufSize` class mirroring `TcpBufSize`.

### C++

- `AcceptorI` reads the properties in its constructor, so an invalid value now fails at object adapter creation. Each accepted socket is configured (non-blocking + buffer sizes) in the `StreamSocket` constructor.
- On the outgoing path — where the fd arrives later, in the Engine's connect callback — `StreamSocket::setFd` now takes ownership of the fd with `setNewFd` *before* the steps that can throw, and `connectCompleted` routes any exception into `_exception`, the same channel `connectFailed` already uses. `initialize()` rethrows it, so connection establishment fails with the real error and `close()` releases the fd.

Previously, an invalid value on the outgoing path escaped into the Engine's catch-all `NewConnection` handler: the fd leaked, the error was swallowed without a diagnostic, and — because the ready signal was never raised — the application saw a `ConnectTimeoutException` several seconds later instead of the `PropertyException`. The throw also skipped `ClientProfile::newConnection`'s reference cleanup, leaking the profile's circular reference.

### Java (Android)

`AcceptorI` and `ConnectorI` read the properties in their constructors (`BTBufSize`, keeping the existing 128 KB default) and pass the validated sizes to `TransceiverI`, so the `BluetoothSocket` popped off the pending stack can no longer be orphaned by a `PropertyException`.

No changelog entry: the only trigger is an invalid property value, consistent with #6335.

### DIFF
--- a/cpp/src/IceBT/AcceptorI.cpp
+++ b/cpp/src/IceBT/AcceptorI.cpp
@@ -155,7 +155,8 @@ IceBT::AcceptorI::newConnection(int fd)
 {
     lock_guard lock(_mutex);
 
-    _transceivers.push(make_shared<TransceiverI>(_instance, make_shared<StreamSocket>(_instance, fd), nullptr, _uuid));
+    _transceivers.push(
+        make_shared<TransceiverI>(_instance, make_shared<StreamSocket>(_instance, fd, _bufSize), nullptr, _uuid));
 
     //
     // Notify the thread pool that we are ready to "read". The thread pool will invoke accept()
@@ -178,7 +179,8 @@ IceBT::AcceptorI::AcceptorI(
       _addr(std::move(addr)),
       _uuid(std::move(uuid)),
       _name(std::move(name)),
-      _channel(channel)
+      _channel(channel),
+      _bufSize(_instance->properties())
 {
     string s = IceInternal::trim(_addr);
     if (s.empty())
--- a/cpp/src/IceBT/AcceptorI.h
+++ b/cpp/src/IceBT/AcceptorI.h
@@ -5,6 +5,7 @@
 
 #include "../Ice/Acceptor.h"
 #include "../Ice/TransceiverF.h"
+#include "BTBufSize.h"
 #include "Config.h"
 #include "EngineF.h"
 #include "InstanceF.h"
@@ -42,6 +43,7 @@ namespace IceBT
         const std::string _uuid;
         const std::string _name;
         const int _channel;
+        const BTBufSize _bufSize;
         std::string _path;
 
         std::mutex _mutex;
--- a/cpp/src/IceBT/BTBufSize.h
+++ b/cpp/src/IceBT/BTBufSize.h
@@ -1,0 +1,33 @@
+// Copyright (c) ZeroC, Inc.
+
+#ifndef ICE_BT_BUF_SIZE_H
+#define ICE_BT_BUF_SIZE_H
+
+#include "Ice/Properties.h"
+
+#include <cstdint>
+
+namespace IceBT
+{
+    // The Bluetooth send and receive buffer sizes, as configured by the IceBT.SndSize and IceBT.RcvSize properties.
+    class BTBufSize
+    {
+    public:
+        // Reads the IceBT.RcvSize and IceBT.SndSize properties. Throws PropertyException if a value is not a valid
+        // integer.
+        explicit BTBufSize(const Ice::PropertiesPtr& properties)
+            : _rcvSize(properties->getIcePropertyAsInt("IceBT.RcvSize")),
+              _sndSize(properties->getIcePropertyAsInt("IceBT.SndSize"))
+        {
+        }
+
+        [[nodiscard]] std::int32_t rcvSize() const noexcept { return _rcvSize; }
+        [[nodiscard]] std::int32_t sndSize() const noexcept { return _sndSize; }
+
+    private:
+        std::int32_t _rcvSize;
+        std::int32_t _sndSize;
+    };
+}
+
+#endif
--- a/cpp/src/IceBT/StreamSocket.cpp
+++ b/cpp/src/IceBT/StreamSocket.cpp
@@ -2,7 +2,6 @@
 
 #include "StreamSocket.h"
 #include "Ice/LoggerUtil.h"
-#include "Ice/Properties.h"
 #include "IceBT/EndpointInfo.h"
 #include "Instance.h"
 #include "Util.h"
@@ -11,15 +10,21 @@ using namespace std;
 using namespace Ice;
 using namespace IceBT;
 
-IceBT::StreamSocket::StreamSocket(InstancePtr instance, SOCKET fd)
+IceBT::StreamSocket::StreamSocket(InstancePtr instance, SOCKET fd, const BTBufSize& bufSize)
     : IceInternal::NativeInfo(fd),
       _instance(std::move(instance))
 {
-    if (fd != INVALID_SOCKET)
-    {
-        init(fd);
-    }
+    assert(fd != INVALID_SOCKET);
+    IceInternal::setBlock(fd, false);
+    setBufferSize(fd, bufSize.rcvSize(), bufSize.sndSize());
     _desc = fdToString(fd);
+}
+
+IceBT::StreamSocket::StreamSocket(InstancePtr instance)
+    : IceInternal::NativeInfo(INVALID_SOCKET),
+      _instance(std::move(instance)),
+      _desc(fdToString(INVALID_SOCKET))
+{
 }
 
 IceBT::StreamSocket::~StreamSocket() { assert(_fd == INVALID_SOCKET); }
@@ -230,18 +235,10 @@ void
 IceBT::StreamSocket::setFd(SOCKET fd)
 {
     assert(fd != INVALID_SOCKET);
-    init(fd);
+    // Take ownership of the fd first: if a later step throws, close() releases the fd.
     setNewFd(fd);
-    _desc = fdToString(fd);
-}
-
-void
-IceBT::StreamSocket::init(SOCKET fd)
-{
     IceInternal::setBlock(fd, false);
-
-    int32_t rcvSize = _instance->properties()->getIcePropertyAsInt("IceBT.RcvSize");
-    int32_t sndSize = _instance->properties()->getIcePropertyAsInt("IceBT.SndSize");
-
-    setBufferSize(fd, rcvSize, sndSize);
+    BTBufSize bufSize{_instance->properties()};
+    setBufferSize(fd, bufSize.rcvSize(), bufSize.sndSize());
+    _desc = fdToString(fd);
 }
--- a/cpp/src/IceBT/StreamSocket.cpp
+++ b/cpp/src/IceBT/StreamSocket.cpp
@@ -235,10 +235,28 @@ void
 IceBT::StreamSocket::setFd(SOCKET fd)
 {
     assert(fd != INVALID_SOCKET);
-    // Take ownership of the fd first: if a later step throws, close() releases the fd.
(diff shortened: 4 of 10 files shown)


externl (commented):
If I'm reading this right, the incoming path still leaks the fd. `AcceptorI::newConnection` passes it to `StreamSocket`'s constructor, which adopts it via `NativeInfo(fd)` and then calls `setBlock` and `setBufferSize` — both of which throw `SocketException` on failure. At that point the object was never constructed, so there's no `StreamSocket` to `close()`, and nothing closes an fd implicitly (`~NativeInfo` is `= default`, `~StreamSocket` only asserts). `Engine.cpp:116-123` then swallows the exception with `catch (...) { // Ignore. }`, so the descriptor BlueZ handed over at `:100` would stay open with no diagnostic — the same shape as the outgoing path, just with `SocketException` as the trigger rather than `PropertyException`. Is that in scope here, or am I missing something that closes it?

bernardnormier:
Right — the incoming path is covered, but the close happens inside the network helpers rather than in a destructor, which is easy to miss when reading only the constructor and destructors.

The full path, taking `setBufferSize` as the example:

1. BlueZ hands over the fd, and the Engine's `NewConnection` handler calls `newConnection(fd->v)` (`Engine.cpp`).
2. `AcceptorI::newConnection` constructs `StreamSocket(_instance, fd, _bufSize)`; the `NativeInfo(fd)` base adopts the fd.
3. The constructor calls `IceInternal::setBlock(fd, false)`, then `setBufferSize`, which calls `IceInternal::setRecvBufferSize` / `setSendBufferSize` (plus the corresponding getters), and finally `fdToString`.
4. Every one of these bottoms out in the same failure convention — `closeSocketNoThrow(fd)` first, then `throw SocketException`:
   - `setBlock`, `setRecvBufferSize`, `getRecvBufferSize`, `setSendBufferSize`, `getSendBufferSize` in `cpp/src/Ice/Network.cpp`;
   - `fdToLocalAddress` / `fdToRemoteAddress` in `cpp/src/IceBT/Util.cpp`, reached via `fdToString`.
5. So by the time the `SocketException` leaves the half-constructed `StreamSocket`, the fd is already closed. The exception then unwinds through `AcceptorI::newConnection` into the Engine's `catch (...)`, which drops the connection — silently, but with nothing left to release.

That convention is also why the outgoing path did leak and needed the `setFd` change: the `PropertyException` came from property parsing — code outside the close-then-throw convention — while the fd had no owner. Same story for the incoming leak this PR fixes: the trigger was the property read in the constructor, not the socket calls.

Agreed the Engine's catch-all keeps per-connection failures silent; that's pre-existing behavior and orthogonal to the buffer-size properties, so I'd leave it alone in this PR.

pepone (commented):
One blocking issue in the C++ outgoing path; the rest looks good.

### The `setNewFd`-first reordering in `StreamSocket::setFd` turns three "already closed" paths into double closes

```cpp
setNewFd(fd);                                              // stream records the fd
IceInternal::setBlock(fd, false);                          // closes fd, then throws
BTBufSize bufSize{_instance->properties()};                // does NOT close  <- the bug being fixed
setBufferSize(fd, bufSize.rcvSize(), bufSize.sndSize());   // closes fd, then throws
_desc = fdToString(fd);                                    // closes fd, then throws
```

This is exactly the close-then-throw convention you documented in your reply above, and it cuts the other way here. `setBlock`, `setRecvBufferSize`, `setSendBufferSize`, `getRecvBufferSize` and `getSendBufferSize` (`cpp/src/Ice/Network.cpp:1162-1250`), and IceBT's `fdToLocalAddress` / `fdToRemoteAddress` reached via `fdToString` (`cpp/src/IceBT/Util.cpp:101,117`), all call `closeSocketNoThrow(fd)` before throwing.

So when one of them throws, the new `catch` stores `_exception`, `ready()` makes `Selector::ready` promote `_newFd` into `_fd` (`cpp/src/Ice/Selector.cpp:405-411`) and hand that number to `epoll_ctl` / `kevent`, `initialize()` rethrows, and `TransceiverI::close()` -> `StreamSocket::close()` closes the *same descriptor a second time* — with a full thread-pool round trip in between for another thread to claim the number.

Pre-PR those calls threw before `setNewFd`, so the stream never recorded the fd and `close()` was a no-op. The hazard is new.

### Suggested fix

Do what #6335 did for TCP (`TcpConnector` holds a `const TcpBufSize _bufSize`) and what this PR's own Java side already does: read the sizes in `IceBT::ConnectorI`'s constructor and pass them down. Then every throwing step in `setFd` closes its own fd, and `setNewFd` can go genuinely last — note that base already had it before `fdToString`, which was slightly wrong for the same reason:

```cpp
void
IceBT::StreamSocket::setFd(SOCKET fd)
{
    assert(fd != INVALID_SOCKET);
    // Each of these closes the fd before throwing, so don't record it until they all succeed.
    IceInternal::setBlock(fd, false);
    setBufferSize(fd, _bufSize.rcvSize(), _bufSize.sndSize());
    string desc = fdToString(fd);

    setNewFd(fd);
    _desc = std::move(desc);
}
```

Keep the `try` / `catch` in `connectCompleted` — that part is a real improvement (the application gets the actual error instead of a `ConnectTimeoutException`, and `ClientProfile::newConnection`'s reference cleanup runs). Just reword its comment: the failing call already released the fd.

### The rest checks out

- `AcceptorI::_bufSize(_instance->properties())` — `_bufSize` is declared after `_instance` (`AcceptorI.h:40,46`), so the initialization order is fine.
- Incoming path is leak-free, per your analysis above: the property read moved to the acceptor constructor, and the remaining constructor steps close their own fd.
- Java: `BTBufSize` mirrors `TcpBufSize` (including `dfltBufSize`), preserves the 128 KB default, and `AcceptorI` / `ConnectorI` read at construction. I compiled the module's sources against `ice-3.9.0-alpha.0.jar` — clean.
- No changelog fragment — consistent with #6335.

### Pre-existing, worth separate issues rather than this PR

1. C++ defaults `IceBT.RcvSize` / `IceBT.SndSize` to 0 (system default — the property table entry is `""`), while Java forces 128 KB.
2. `TransceiverI::getInfo` and `TransceiverI::setBufferSize` call the same close-on-error helpers with `_fd` already installed, leaving a stale `_fd` for a later `close()`.
3. `connectFailed` has no `_closed` guard, unlike `connectCompleted`. `ThreadPool::finish` clears the ready callback (`cpp/src/Ice/ThreadPool.cpp:356-361`), so a late `connectFailed` after a connect timeout would hit `assert(_readyCallback)`.

externl (approved):
You're right, and my comment was wrong — `setBlock` and `setRecvBufferSize` do `closeSocketNoThrow(fd)` before throwing (`Network.cpp:1149`, `:1215`), so the fd is already closed by the time the `SocketException` reaches the Engine's catch-all. No leak on that path.

bernardnormier:
Good catch on the double close — fixed in 45bebba.

I took the ordering fix but kept the property read inside `setFd` rather than threading a `BTBufSize` from `ConnectorI` through `TransceiverI`. The read now happens first, in its own try/catch that closes the fd on failure — it's the one step outside the close-then-throw convention, and at that point the fd isn't recorded anywhere, so a catch-all close is unconditionally correct. The remaining calls (`setBlock`, `setBufferSize`, `fdToString`) each close their own fd, and `setNewFd` moves genuinely last — which also fixes the pre-existing `fdToString`-after-`setNewFd` wrinkle you noted in the base code. Net effect: every failure path closes the fd exactly once and never records it, with no new state anywhere.

Also reworded the `connectCompleted` catch comment as you suggested, and added an `assert(_fd == INVALID_SOCKET)` at the top of `setFd`.

Agreed the three pre-existing findings are separate from this PR.

bernardnormier:
Filed the three pre-existing findings: #6375 (C++/Java default mismatch), #6376 (stale `_fd` in `getInfo`/`setBufferSize`), #6377 (missing `_closed` guard in `connectFailed`).

pepone (approved):
Verified 45bebba — the double close is resolved.

Walking the failure paths in the new `setFd`:

- `BTBufSize` throws — explicit `closeSocketNoThrow` + rethrow, fd not recorded.
- `setBlock` / `setBufferSize` / `fdToString` throw — each closes its own fd, and none is recorded.
- Nothing after `setNewFd` can throw (the `std::string` move-assign is `noexcept`).

So the fd is closed exactly once on every path and `_newFd` is only ever set on success. `StreamSocket::close()` is then a no-op after a failure, and `Selector::ready`'s `newFd()` returns false, so no closed descriptor reaches `epoll_ctl` / `kevent` either.

Keeping the property read in `setFd` behind its own try/catch instead of threading a `BTBufSize` down from `ConnectorI` works — it leaves C++ reading the properties per connection where Java and `TcpConnector` read them once at construction, but that's a consistency preference, not a correctness one, and your version adds no new state.

Nice catch folding in the pre-existing `fdToString`-after-`setNewFd` ordering while you were there. Thanks for filing #6375, #6376 and #6377.

LGTM.

bernardnormier:
> Keeping the property read in `setFd` behind its own try/catch instead of threading a `BTBufSize` down from `ConnectorI` works — it leaves C++ reading the properties per connection where Java and `TcpConnector` read them once at construction, but that's a consistency preference, not a correctness one, and your version adds no new state.

Yes, it's because I was not keen on keeping a BTSize field on TransceiverI or StreamSocket only for outgoing connections. We don't do that anywhere else. I find the current shape simpler.

PR #6363, opened 2026-08-03
Document the empty byte sequence special cases of dynamic invocation

### LINKED ISSUE #6356: Doc comments of ice_invoke/ice_invokeAsync don't document the empty in-params special case
The doc comments of the dynamic invocation APIs describe the in-parameters argument as "an encapsulation containing the encoded in-parameters for the operation" (or just "the encoded in-parameters"). Taken literally, this requires a valid encapsulation, which is at least 6 bytes. In reality, every mapping also accepts an empty buffer: the Ice runtime then marshals the empty encapsulation itself. This is the canonical way to dynamically invoke an operation that takes no in-parameters — the `Ice/invoke` tests rely on it — but no doc comment mentions it.

The doc comments should document this special case in all mappings that provide `ice_invoke` / `ice_invokeAsync`:

- **C++**: the six `ice_invoke` / `ice_invokeAsync` overloads in `cpp/include/Ice/Proxy.h`. Runtime special case: `InvokeOutgoingAsyncT::invoke` in `cpp/src/Ice/ProxyAsync.cpp` (and the sync/batch paths in `cpp/src/Ice/OutgoingAsync.cpp`).
- **C#**: the four `inEncaps` doc sites in `csharp/src/Ice/Proxy.cs`. The runtime accepts an empty (or even null) `byte[]`.
- **Java**: the four `@param inParams` sites in `com.zeroc.Ice.ObjectPrx`. Runtime special case in `ProxyIceInvoke`.
- **Swift**: `ice_invoke` / `ice_invokeAsync` in `swift/src/Ice/Proxy.swift` (an empty `Data` is accepted; the call bridges to the C++ runtime).
- **Python**: the IcePy docstrings (`python/modules/IcePy/Proxy.cpp`) are currently signature-only; if parameter documentation is added, it should cover this case too.

JavaScript, Ruby, PHP, and MATLAB have no public dynamic invocation API.

Found while adding a test for #6355, which invokes a parameterless operation with an empty `vector<byte>`.

bernardnormier:
Fixes #6356.

Documents the empty-byte-sequence special cases of the dynamic invocation APIs in C++, C#, Java, and Swift. All the documented behavior was verified against the runtimes.

- **`inParams`/`inEncaps`** (all `ice_invoke`/`ice_invokeAsync` overloads): you can pass an empty byte sequence when the operation takes no in-parameters; the Ice runtime then marshals an empty encapsulation. This is the canonical way to dynamically invoke a parameterless operation.
- **Blobject `outEncaps`** (C++, C#, Java — the server-side counterpart): the implementation can leave it empty when the operation returns no results; the runtime marshals an empty encapsulation into the reply (`makeOutgoingResponse` / `createOutgoingResponse`).
- **Client-side `outParams` for oneway, datagram, and batch proxies**: the invocation completes with `true` and an *empty* `outParams` (not an encapsulation) as soon as the request is accepted by the transport or added to the batch. Swift already documented this; C++, C#, and Java now do too.

Also rewords the Java `Blobject`/`BlobjectAsync` `@return` instructions in the imperative ("To report a successful completion, set…" / "To report a user exception, either throw it directly or…") instead of narrating the operation in the past tense, and rewraps the touched javadoc to the full 120 columns enforced by checkstyle.

Python is intentionally not covered: the IcePy docstrings are currently signature-only, so there is no parameter documentation to extend (as noted in the issue).

### DIFF
--- a/cpp/include/Ice/Object.h
+++ b/cpp/include/Ice/Object.h
@@ -93,7 +93,8 @@ namespace Ice
     public:
         /// Dispatches an incoming request.
         /// @param inEncaps An encapsulation containing the encoded in-parameters for the operation.
-        /// @param outEncaps An encapsulation containing the encoded result for the operation.
+        /// @param outEncaps An encapsulation containing the encoded result for the operation. You can leave it empty
+        /// when the operation returns no results; the Ice runtime then marshals an empty encapsulation.
         /// @param current The Current object of the incoming request.
         /// @return `true` if the dispatch completes successfully, `false` if the dispatch completes with a user
         /// exception encoded in @p outEncaps.
@@ -133,7 +134,8 @@ namespace Ice
         /// @param response The response callback. It accepts:
         /// - `returnValue` `true` if the operation completed successfully, `false` if it completed with a user
         ///   exception encoded in @p outEncaps.
-        /// - `outEncaps` An encapsulation containing the encoded result.
+        /// - `outEncaps` An encapsulation containing the encoded result. You can pass an empty byte sequence when
+        ///   the operation returns no results; the Ice runtime then marshals an empty encapsulation.
         /// @param exception The exception callback.
         /// @param current The Current object of the incoming request.
         virtual void ice_invokeAsync(
--- a/cpp/include/Ice/Proxy.h
+++ b/cpp/include/Ice/Proxy.h
@@ -449,10 +449,14 @@ namespace Ice
         /// Invokes an operation.
         /// @param operation The name of the operation to invoke.
         /// @param mode The operation mode (normal or idempotent).
-        /// @param inParams An encapsulation containing the encoded in-parameters for the operation.
+        /// @param inParams An encapsulation containing the encoded in-parameters for the operation. You can pass an
+        /// empty byte sequence when the operation takes no in-parameters; the Ice runtime then marshals an empty
+        /// encapsulation.
         /// @param outParams An encapsulation containing the encoded result.
         /// @param context The request context.
         /// @return `true` if the operation completed successfully, `false` if it completed with a user exception.
+        /// @remark When this proxy is a oneway, datagram, or batch proxy, this function returns `true` and an empty
+        /// @p outParams as soon as the request is accepted by the transport or added to the batch.
         bool ice_invoke(
             std::string_view operation,
             Ice::OperationMode mode,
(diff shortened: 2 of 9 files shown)


pepone (approved):
> [!NOTE]
> This is an AI-assisted review. Every claim below was verified by re-reading the runtime sources at the PR head commit (8364e6d); nothing was benchmarked or executed. Please triage accordingly.

LGTM. I traced each documented claim to the code that implements it rather than relying on the PR description, and all three special cases check out in all four mappings.

**Empty in-params → empty encapsulation**

- C++ `InvokeOutgoingAsyncT::invoke` — `if (inParams.first == inParams.second) _os.writeEmptyEncapsulation(_encoding);` (`cpp/src/Ice/ProxyAsync.cpp:158`)
- C# `InvokeOutgoingAsyncT.invoke` — `if (inParams is null || inParams.Length == 0)` (`csharp/src/Ice/Proxy.cs:1579`)
- Java `ProxyIceInvoke.writeParamEncaps` — same null/length-0 check
- Swift bridges to the C++ path

**Blobject empty out-encapsulation → empty encapsulation**

- C++ `Blobject::dispatch` and `BlobjectArray::dispatch` check `outEncaps.empty()`; `makeOutgoingResponse(bool, pair, current)` independently falls back to `writeEmptyEncapsulation` when `size == 0` (`cpp/src/Ice/OutgoingResponse.cpp:124`), which is what covers `BlobjectArrayAsync`
- C# and Java `createOutgoingResponse(ok, encapsulation)` both normalize null to an empty array, then branch on length

**Oneway / datagram / batch → `true` + empty out-params**

- C++ `InvokePromiseOutgoing::handleSent` sets `R{true, {}}`; the batch path reaches `responseImpl(true, false)` with an empty `_is`, which also yields `{true, {}}`. The sync overload then does `outParams.swap(resultOutParams)`, so a caller-populated `outParams` is correctly cleared
- C# `InvokeTaskCompletionCallback.handleInvokeSent` sets `Object_Ice_invokeResult(true, [])`, and `getResult` returns `[]` when the reference is not twoway
- Java `ProxyIceInvoke.markCompleted` and `waitForResponse` both produce `new Object.Ice_invokeResult(true, new byte[0])`

Two things I expected to be gaps and weren't: `BlobjectArray`/`BlobjectArrayAsync` inherit the new text via `@copydoc`, and the C++ lambda `ice_invokeAsync` overloads correctly do *not* get the "returns true and empty out-params" remark — they already carry a different and correct remark saying the response callback is never invoked for oneway/datagram/batch, which matches `LambdaInvoke::handleSent`. No line-length regressions either; the one 141-column line in `Proxy.cs` predates this PR.

Three comments, none blocking:

**1. Copilot's Swift comment is right — `swift/src/Ice/Proxy.swift:413`**

```swift
///   For oneway and batch proxies, `ok` is always `true` and `outEncaps` is empty.
```

Datagram proxies take the same non-twoway `else` branch (`Proxy.swift:443-467`) and return `(true, Data())`. This is now the only place saying "oneway and batch" while C++, C#, and Java all say "oneway, datagram, or batch" — so it is both an accuracy gap and an inconsistency introduced by this PR.

**2. C# wording — `csharp/src/Ice/Object.cs:141`**

`Blobject.ice_invoke`'s `outParams` is an `out byte[]`, so definite assignment means an implementation cannot "leave it empty" — it has to assign something. Suggest "set it to an empty array" instead. (The runtime also tolerates `null`, but not documenting that seems right.)

**3. Optional: the Python server side is coverable**

The stated reason for skipping Python holds for the proxy side — `python/modules/IcePy/Proxy.cpp` docstrings really are signature-only. But `python/python/Ice/Blobject.py:21` has full numpy-style parameter docs, and `BlobjectUpcall::response` (`python/modules/IcePy/Operation.cpp:2229`) has the same `sz > 0` fallback. So the server-side half of this change could cover Python even though the proxy half cannot. Fine as a follow-up.

externl (approved):
Looks good.

bernardnormier:
All three addressed in 3fcfccc: the Swift return doc now says "oneway, datagram, and batch" like the other mappings, the C# `Blobject.ice_invoke` doc says "set it to an empty array" (you're right that definite assignment rules out "leave it empty"), and I took the Python server-side suggestion here rather than as a follow-up — `Blobject.ice_invoke`'s docstring now documents the empty-byte-sequence return. The Python proxy side stays out, as its docstrings are signature-only.

PR #6365, opened 2026-08-03
Fix GC root leak when unmarshaling user exceptions in Ruby

bernardnormier:
`ExceptionReader` holds the unmarshaled Ruby exception in a `VALUE _ex` member and protects it from the GC with `rb_gc_register_address(&_ex)`. The registration was in the wrong place: `_read` runs *after* `DefaultSliceLoader::newExceptionInstance` has copied the reader into the C++ exception storage via `make_exception_ptr`. As a result:

- The primary constructor left `_ex` uninitialized and unregistered.
- The copy constructor registered the copy's `_ex` while it still held a garbage value.
- `_read` then registered the same address a second time.
- The destructor unregistered only once.

Net effect: every Slice user exception a Ruby client unmarshaled permanently leaked one registered address pointing into freed C++ exception storage. The GC then read that freed memory on every collection, and because `rb_gc_unregister_address` searches the global registration list linearly, each received exception made every subsequent exception and GC slower.

The fix makes the invariant per-instance: every constructor registers the new instance's `_ex` address exactly once (initialized to `Qnil` in the primary constructor), the destructor unregisters it, and `_read` just assigns. The copy constructor must remain available — the C++ exception machinery copies the reader into the exception storage and requires the type to be copy-constructible — so the assignment operator is deleted instead. This differs from the sibling `ValueReader`/`ValueWriter`, which the runtime only handles through `shared_ptr` and which therefore delete copy and move outright.

Measured with a loopback client/server unmarshaling 200,000 user exceptions (after a 100,000-exception warmup, Ruby 4.0.5, arm64-darwin):

| | RSS growth | wall time |
|---|---|---|
| before | 34.7 MB | 64.3 s |
| after | 31.5 MB | 15.6 s |

The remaining RSS growth is ordinary Ruby heap expansion from the allocation churn; the eliminated ~3 MB is the leaked registration nodes, and the 4x wall-time difference is the linear search over the ever-growing registration list.

### DIFF
--- a/changelog.d/ruby/exception-reader-gc-root.md
+++ b/changelog.d/ruby/exception-reader-gc-root.md
@@ -1,0 +1,2 @@
+- Fixed a bug where receiving a Slice user exception leaked memory. A program that received many user exceptions
+  became progressively slower and could eventually crash.
--- a/ruby/src/IceRuby/Types.cpp
+++ b/ruby/src/IceRuby/Types.cpp
@@ -2748,7 +2748,10 @@ IceRuby::ExceptionInfo::printMembers(VALUE value, IceInternal::Output& out, Prin
 //
 // ExceptionReader implementation.
 //
-IceRuby::ExceptionReader::ExceptionReader(const ExceptionInfoPtr& info) : _info(info) {}
+IceRuby::ExceptionReader::ExceptionReader(const ExceptionInfoPtr& info) : _info(info), _ex(Qnil)
+{
+    rb_gc_register_address(&_ex);
+}
 
 IceRuby::ExceptionReader::ExceptionReader(const ExceptionReader& reader) : _info(reader._info), _ex(reader._ex)
 {
@@ -2779,9 +2782,7 @@ void
 IceRuby::ExceptionReader::_read(Ice::InputStream* is)
 {
     is->startException();
-
-    const_cast<VALUE&>(_ex) = _info->unmarshal(is);
-    rb_gc_register_address(&_ex);
+    _ex = _info->unmarshal(is);
 }
 
 bool
--- a/ruby/src/IceRuby/Types.h
+++ b/ruby/src/IceRuby/Types.h
@@ -507,9 +507,14 @@ namespace IceRuby
     class ExceptionReader final : public Ice::UserException
     {
     public:
+        // Each constructor registers a GC root tied to the address of the new instance's _ex member, and the
+        // destructor unregisters it. The copy constructor must remain available: the exception machinery copies the
+        // reader into the exception storage.
         ExceptionReader(const ExceptionInfoPtr&);
         ExceptionReader(const ExceptionReader&);
         ~ExceptionReader();
+
+        ExceptionReader& operator=(const ExceptionReader&) = delete;
 
         const char* ice_id() const noexcept final;
         void ice_throw() const final;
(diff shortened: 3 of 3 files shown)


pepone (approved):
Approving. The diagnosis is correct and the fix is minimal and complete. I reproduced your numbers locally.

Built the branch in a worktree and ran a loopback probe against the C++ `Ice/exceptions` server (`throwAasA`, Ruby 4.0.5, arm64-darwin, 100k warmup + 100k measured):

| | wall | RSS growth |
|---|---|---|
| base (`HEAD~1`) | 27.83 s | 16.9 MB |
| PR | 7.61 s | 15.4 MB |

`Ice/exceptions` and `Ice/slicing` pass. `Ice/optional` fails identically on base and PR in my environment (stale prebuilt C++ server), so it's unrelated.

Mechanism confirmed against source: `DefaultSliceLoader.cpp:42` calls `make_exception_ptr`, which takes its argument by value and separately copy-constructs into the `__cxa` storage — so pre-patch the copy constructor read and registered an indeterminate `_ex` (formally UB, not just a garbage value). `InputStream.cpp:1536`/`1830` then call `ex._read(_stream)` on that stored object, registering the same address a second time. One destructor, two registrations, one leak per exception. The post-patch invariant holds on every in-tree path.

Two non-blocking comments:

**Spell out all five special members, like the siblings.** `ValueReader` (Types.h:483-487), `ValueWriter` (Types.h:454-458) and `RubySliceLoader` (RubySliceLoader.h:18-22) each list copy/move constructor and copy/move assignment under one comment. Here we delete only copy assignment and rely on the user-declared copy constructor plus destructor to suppress the moves. Correct today, but a later `ExceptionReader(ExceptionReader&&) = default;` would silently reintroduce an unregistered `_ex`. Adding both explicitly still builds — `make_exception_ptr` copies from an lvalue, so a deleted move constructor is never a candidate:

```cpp
ExceptionReader(ExceptionReader&&) = delete;
ExceptionReader& operator=(const ExceptionReader&) = delete;
ExceptionReader& operator=(ExceptionReader&&) = delete;
```

**No test is the right call.** The symptom has no functional signal and `Ice/exceptions` already covers the behavior.

Two pre-existing issues I noticed while reviewing, both out of scope here — I can file them:

- `ExceptionReader::_read` never calls `is->endException()`. IcePy (`Types.cpp:3324`) and IcePHP (`Types.cpp:3252`) both do, and 3.7's Ruby did too; it was dropped in the 3.8 refactor. It leaves `EncapsDecoder11::_current` unpopped and the slices uncleared. No observable impact today since the `InputStream` is single-use per exception, so it's a consistency fix rather than a bug.
- `ValueWriter`'s constructor registers `&_object` and then calls throwable Ruby code (`Types.cpp:2362-2369`; `callRuby` translates Ruby errors into `RubyException`). If it throws, the destructor never runs and the root leaks — the same class of bug this PR fixes. Hard to reach from `ClassInfo::marshal`, where validation guarantees `ICE_TYPE` exists, but reachable via the `SlicedData` `@instances` path at `Types.cpp:357`.

3.8.3 is the right milestone. 3.7 has the same `_read`-registers pattern but uses `throw ExceptionReader(info)` (Operation.cpp:90) with guaranteed prvalue elision and no user-declared copy constructor — one object, one registration, balanced. The leak looks specific to the 3.8 `SliceLoader`/`make_exception_ptr` refactor, so no 3.7 backport.

pepone:
Filed the two pre-existing findings from this review: #6369 (missing `endException`) and #6370 (`ValueWriter` registers its GC root before code that can throw).

externl (approved):
Looks good.

PR #6368, opened 2026-08-03
Simplify the executor termination waits in the Java mapping

### LINKED ISSUE #6358: Simplify the executor termination waits in the Java mapping
The Java mapping waits for executor termination with a redundant two-call loop and an arbitrary timeout:

```java
while (!_executor.isTerminated()) {
    // A very long time.
    _executor.awaitTermination(100000, TimeUnit.SECONDS);
}
```

`awaitTermination` already returns whether the executor terminated, so this can be the standard "wait forever" idiom:

```java
while (!_executor.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS)) {}
```

The behavior is identical: block until the executor terminates and propagate `InterruptedException`. `ExecutorService.close()` is not a substitute here — on interrupt it calls `shutdownNow()` and swallows the `InterruptedException`, which would break the interrupt contract of communicator destruction.

Occurrences:

- `EndpointHostResolver.joinWithThread`
- `Instance.destroy` (the `_timer` wait)
- `test/Ice/interrupt/AllTests.java` (three occurrences, with a 1,000-second timeout)

bernardnormier:
Fixes #6358.

Replaces the redundant two-call executor-termination waits — a `while (!executor.isTerminated()) { executor.awaitTermination(<arbitrary timeout>); }` loop — with the standard single "wait forever" call:

```java
executor.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
```

`awaitTermination` returns immediately when the executor has already terminated, blocks until it does otherwise, and propagates `InterruptedException` — so the behavior is identical, without the redundant `isTerminated()` check or the arbitrary 100000-second (27-hour) timeout the loop used to re-arm. `ExecutorService.close()` is not a substitute: on interrupt it calls `shutdownNow()` and swallows the `InterruptedException`, which would break the interrupt contract of communicator destruction.

This actually restores the original single-call form written in the 2014 interrupt-handling code (`b36ae21c88`); the loop was introduced later, by hand, during the Ice 3.7 "Java 8 mapping" rewrite (`e6cbf802f2`), and every commit since only reformatted it.

Occurrences updated: `EndpointHostResolver.joinWithThread`, `Instance.destroy` (the `_timer` wait), and three in `test/Ice/interrupt/AllTests.java`.

Internal cleanup only, so no changelog entry.

### DIFF
--- a/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/EndpointHostResolver.java
+++ b/java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/EndpointHostResolver.java
@@ -120,11 +120,7 @@ class EndpointHostResolver {
     void joinWithThread() throws InterruptedException {
         // Wait for the executor to terminate.
         try {
-            while (!_executor.isTerminated()) {
-                // A very long time.
-                _executor.awaitTermination(100000, TimeUnit.SECONDS);
-            }
-
+            _executor.awaitTermination(Long.MAX_VALUE, TimeUnit.NANOSECONDS);
         } finally {
             if (_observer != null) {
                 _observer.detach();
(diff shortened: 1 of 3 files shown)


pepone (approved):

InsertCreativityHere (approved):
Looks good to me!

PR #6373, opened 2026-08-04
Fix Python Missing `connectionId` and `sequence<float>` Marshaling Bug

InsertCreativityHere:
In Python, 'ConnectionInfo' didn't have a way to get the 'connectionId' from a 'ConnectionInfo' object, unlike
every other language mapping. Now it does.

Also fixed a bug where marshaling a decimal value outside the range of a Slice float used to be silently changed
to 'inf'; now when this happens we raise a 'ValueError' instead, stating the value is out-of-range.

This fixes both pieces of #6301.


changelog.d/python/missing-connection-id-getter.md
```diff
@@ -0,0 +1,2 @@
+- `Ice.ConnectionInfo` now provides a `connectionId` getter. Previously, the connection ID was not
+  available through `Connection.getInfo`.
```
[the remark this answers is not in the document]
InsertCreativityHere: In this case, the issue that was opened really should of been 2 separate issues.
It consists of 2 unrelated points, the only thing tying them together is that they were discovered at the same time.
This is not a great reason to tie their CHANGELOG entries together though.

python/modules/IcePy/Types.cpp
```diff
@@ -692,21 +692,21 @@ IcePy::PrimitiveInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap*, bo
         }
         case PrimitiveInfo::KindFloat:
         {
-            double dval = PyFloat_AsDouble(p); // Attempts to perform conversion.
```
InsertCreativityHere, line 695: Every other case names this variable just `val`. I get that this is a `double`... but I see no reason to have a special name. I changed it to match all the other cases.

python/modules/IcePy/Types.cpp
```diff
@@ -692,21 +692,21 @@ IcePy::PrimitiveInfo::marshal(PyObject* p, Ice::OutputStream* os, ObjectMap*, bo
...
+            double val = PyFloat_AsDouble(p); // Attempts to perform conversion.
             if (PyErr_Occurred())
             {
                 wrapRaisedExceptionInValueError("invalid value for float type");
                 throw AbortMarshaling();
             }
 
             // Check if the value is within float range (infinity/nan are allowed)
-            if (!((dval <= numeric_limits<float>::max() && dval >= -numeric_limits<float>::max()) || !isfinite(dval)))
+            if (isfinite(val) && (val > numeric_limits<float>::max() || val < numeric_limits<float>::lowest()))
```
InsertCreativityHere, line 703: This check is just over-complicated. Simplified by-hand, via De-Morgan's laws.

python/modules/IcePy/Types.cpp
```diff
@@ -1789,7 +1789,7 @@ IcePy::SequenceInfo::marshalPrimitiveSequence(const PrimitiveInfoPtr& pi, PyObje
                     throw AbortMarshaling();
                 }
 
-                auto val = static_cast<float>(PyFloat_AsDouble(item));
+                double val = PyFloat_AsDouble(item);
```
InsertCreativityHere, line 1792: This and the code below it is the actual fix for the `sequence<float>`; we were missing this necessary check for each `float`'s range.

bernardnormier (approved):
Both fixes look good — the `sequence<float>` check matches the Ruby fix for the same issue exactly.

Two suggestions on the changelog fragments:
- I would drop `float-sequence-range-check.md`: an out-of-range element is invalid input, and this fix only changes how it's rejected. The identical Ruby fix (under the same parent issue) has no entry for the same reason.
- In the remaining fragment, "getter" is the implementation view — for a Python user, `connectionId` is an attribute.

pepone (approved):
Looks good. I verified the fix locally: the new `opFloatDoubleS` out-of-range test fails with the previous marshaling code and passes with this PR.

One minor suggestion: `TestI.py` now adds `connectionId` to the context returned by `getConnectionInfoAsContext`, but `AllTests.py` never checks it. Consider asserting it next to the `adapterName` check:

```python
test(ctx["connectionId"] == "")
```

(incoming connections always have an empty connection ID), or dropping the `TestI.py` line.

InsertCreativityHere:
@pepone

I added the suggested check.
Originally I considered not adding this at all, but worried it could be a future foot-gun. We do this for every other field of the info object, so it'd be weird and inconsistent to omit `connectedId` just because it happened to be missing originally.

InsertCreativityHere:
@bernardnormier

I updated the wording from "getter" to "attribute", indeed this is the more correct terminology.

For the other point though, I think the note is warranted.
> this fix only changes how it's rejected

This is not correct. Beforehand, we _didn't_ reject these values. They would silently convert to `NaN`, which is a perfectly legal value. No error. For users who send overly large floats, this would succeed before, only now does it fail with a `ValueError`. Since this is a large, user-facing change, I really think we ought to document it in the CHANGELOG, instead of being silent about this more strict validation.

bernardnormier:
> > this fix only changes how it's rejected
> 
> This is not correct. Beforehand, we _didn't_ reject these values. They would silently convert to `NaN`, which is a perfectly legal value. No error. For users who send overly large floats, this would succeed before, only now does it fail with a `ValueError`. Since this is a large, user-facing change, I really think we ought to document it in the CHANGELOG, instead of being silent about this more strict validation.

Fine, add a changelog fragment if you like. It's still super minor in my view--and we don't want to overwhelm users with minor details in the eventual CHANGELOG.

InsertCreativityHere:
> and we don't want to overwhelm users with minor details in the eventual CHANGELOG.

I agree with this, but also don't want to leave users in the dark in case they run into this change.
And, I'm not saying it's not a real concern, but I think no matter what, our users will be overwhelmed with our CHANGELOG at this point...

PR #6381, opened 2026-08-04
Cancel the reap task when unregistering a communicator's last ID

### LINKED ISSUE #6366: PHP: Ice\unregister leaks the communicator when it was registered with an expiration time
## Summary

In the PHP extension, `Ice\unregister` does not cancel the reap timer task of the communicator it unregisters. When the communicator was registered with an expiration time, this leaves the communicator's `ActiveCommunicator ↔ ReapCommunicatorTimerTask` ownership cycle intact, so the communicator is never destroyed — its C++ `Ice::Communicator` survives until process exit, producing a "communicator not destroyed during global destruction." warning (and leaking the communicator and its resources).

## Mechanism

Registering with an expiration time creates a `ReapCommunicatorTimerTask` that holds an owning `ActiveCommunicatorPtr`, stores it in `ActiveCommunicator::reapTask`, and schedules it on the timer ([`Communicator.cpp:1048-1057`](https://github.com/zeroc-ice/ice/blob/dd1b1fbf41d1c6d7c42ba266e3ab236dfca87d11/php/src/Communicator.cpp#L1048-L1057)). This is a deliberate reference cycle: `ActiveCommunicator` → `reapTask` → `ReapCommunicatorTimerTask` → `_activeCommunicator` → `ActiveCommunicator`.

Every path that stops reaping is careful to break the cycle by cancelling the task and clearing `reapTask`:

- the reap task itself, when the communicator has expired ([`Communicator.cpp:136-140`](https://github.com/zeroc-ice/ice/blob/dd1b1fbf41d1c6d7c42ba266e3ab236dfca87d11/php/src/Communicator.cpp#L136-L140));
- `Communicator::destroy` ([`Communicator.cpp:244-247`](https://github.com/zeroc-ice/ice/blob/dd1b1fbf41d1c6d7c42ba266e3ab236dfca87d11/php/src/Communicator.cpp#L244-L247));
- `Ice\register`, before scheduling a replacement task ([`Communicator.cpp:1041-1046`](https://github.com/zeroc-ice/ice/blob/dd1b1fbf41d1c6d7c42ba266e3ab236dfca87d11/php/src/Communicator.cpp#L1041-L1046)).

`Ice\unregister` ([`Communicator.cpp:1063-1096`](https://github.com/zeroc-ice/ice/blob/dd1b1fbf41d1c6d7c42ba266e3ab236dfca87d11/php/src/Communicator.cpp#L1063-L1096)) removes the ID from `ActiveCommunicator::ids` and erases the `_registeredCommunicators` entry, but never touches `reapTask`. If the unregistered ID was the communicator's only registration and the reap task's expiration is far enough away that it doesn't fire, nothing ever breaks the cycle and the communicator leaks.

## Reproduction

```php
<?php
$c = Ice\initialize();
Ice\register($c, "A", 40000); // expiry far in the future, so the reap task won't fire
Ice\unregister('A');
// $c dropped; no explicit destroy
```

At exit:

```
!! ... error: communicator not destroyed during global destruction.
```

A plain registered-without-expiry communicator, or one destroyed explicitly, cleans up normally — the leak is specific to the register-with-expiry + unregister sequence. This is exercised today by `php/test/Ice/registeredCommunicator` (`register("Hello7", 40000)` followed by `unregister('Hello7')`), which prints the warning on every run.

## Suggested fix

Cancel and clear the reap task in `Ice\unregister`, under the `_registeredCommunicatorsMutex` it already holds, mirroring `destroy`:

```cpp
if (ac->reapTask)
{
    _timer->cancel(ac->reapTask);
    ac->reapTask = nullptr;
}
```

Note this is correct only because a communicator's reap task is shared across all its registrations (the most recent expiration wins); unregistering any ID cancels reaping for the communicator, which is consistent with the existing single-`reapTask`-per-communicator model.

bernardnormier:
Fixes #6366.

Registering a communicator with an expiration time creates a deliberate ownership cycle: `ActiveCommunicator` → `reapTask` → `ReapCommunicatorTimerTask` → `ActiveCommunicator`. Every path that stops reaping breaks this cycle (the reap task itself on expiration, `Communicator::destroy`, and `Ice\register` before scheduling a replacement task) — except `Ice\unregister`, which removed the registration but left the cycle intact, keeping the communicator alive until process exit.

`Ice\unregister` now cancels the reap task and clears `reapTask`, under the `_registeredCommunicatorsMutex` it already holds — but only when it removes the communicator's **last** registered ID. This differs from the sketch in #6366, which suggested cancelling unconditionally: a communicator registered under several IDs shares one reap task, so an unconditional cancel would silently disable expiration for the remaining registrations. The leak can only occur once no registration holds the communicator in the map, so the `ids.empty()` guard covers it fully.

### Test

The existing `php/test/Ice/registeredCommunicator` test already exercises this exact sequence (`register("Hello7", 40000)` followed by `unregister`), and printed `!! error: communicator not destroyed during global destruction.` on every run. With this fix the warning is gone. The full PHP test suite passes.

### DIFF
--- a/changelog.d/php/6366.md
+++ b/changelog.d/php/6366.md
@@ -1,0 +1,2 @@
+- Fixed a leak: unregistering a communicator that was registered with an expiration time left this communicator
+  alive until the process exited, along with its connections and other resources.
--- a/php/src/Communicator.cpp
+++ b/php/src/Communicator.cpp
@@ -1062,6 +1062,10 @@ ZEND_FUNCTION(Ice_unregister)
 
     string id(s, sLen);
 
+    // Declared before the lock: releasing the last reference destroys the communicator, which we must not do
+    // while holding _registeredCommunicatorsMutex.
+    ActiveCommunicatorPtr ac;
+
     lock_guard lock(_registeredCommunicatorsMutex);
 
     RegisteredCommunicatorMap::iterator p = _registeredCommunicators.find(id);
@@ -1076,7 +1080,7 @@ ZEND_FUNCTION(Ice_unregister)
     //
     // Remove the ID from the ActiveCommunicator's list of registered IDs.
     //
-    ActiveCommunicatorPtr ac = p->second;
+    ac = p->second;
     vector<string>::iterator q = find(ac->ids.begin(), ac->ids.end(), id);
     assert(q != ac->ids.end());
     ac->ids.erase(q);
--- a/php/src/Communicator.cpp
+++ b/php/src/Communicator.cpp
@@ -1083,6 +1083,15 @@ ZEND_FUNCTION(Ice_unregister)
 
     _registeredCommunicators.erase(p);
 
+    if (ac->ids.empty() && ac->reapTask)
+    {
+        // This was the communicator's last registration: cancel the reap task to break the
+        // ActiveCommunicator <-> ReapCommunicatorTimerTask ownership cycle, which would otherwise keep the
+        // communicator alive until the task fires.
+        _timer->cancel(ac->reapTask);
+        ac->reapTask = nullptr;
+    }
+
     RETURN_TRUE;
 }
 


pepone (approved):
The fix is correct, and the `ids.empty()` guard is the right call over the unconditional cancel sketched in #6366 — the reap task is per-communicator, not per-ID, so cancelling on a non-final unregister would silently disable expiration for the surviving registrations.

Verified locally against 84f5eeaa3606509f09d6f4998e247fe9771b538e (`php -n -d extension=php/lib/ice.so`, PHP 8.5.9 on macOS):

| Case | main | with this PR |
|---|---|---|
| `register($c, "C", 40000)` + `unregister("C")` (#6366) | warns | **clean** |
| `register($c, "A", 40000)`, never unregistered | warns | warns |
| `register($c, "B")` (no expiry), never unregistered | clean | — |
| never registered | clean | — |

The second row is a separate leak, not something this PR needs to cover — filed as #6392.

One non-blocking comment on who ends up destroying the communicator.

With the reap task gone, the remaining owners of the `ActiveCommunicator` are the per-request `CommunicatorMap` entries (via `CommunicatorInfoI::ac`, released in `communicatorRequestShutdown`) and the local `ac` in `Ice_unregister`. When the unregistering request never obtained the communicator — a script that unregisters an ID it never `Ice\find`s, i.e. the normal cross-request case — that local is the last owner, so `~ActiveCommunicator` runs at the end of `Ice_unregister` and calls `Ice::Communicator::destroy()` **while `_registeredCommunicatorsMutex` is still held**: `ac` is declared after the `lock_guard`, so it's destroyed first.

Every other teardown path deliberately avoids that. `Ice_Communicator::destroy` scopes the lock in a nested block and calls `c->destroy()` after it, and `runTimerTask` hoists the pointer out specifically to "destroy the communicator outside the lock". Here a blocking destroy (connection closes, retry queue drain) stalls every other `Ice\register`/`find`/`unregister` plus the reap timer thread.

Not introduced by this PR — it already applied to `expires == 0` registrations — but the PR extends it to the `expires > 0` case, which is exactly where there is more to tear down. Suggest declaring the pointer before the lock so it outlives it:

```cpp
// Declared before the lock: releasing the last reference destroys the communicator, which we must not do
// while holding _registeredCommunicatorsMutex.
ActiveCommunicatorPtr ac;
lock_guard lock(_registeredCommunicatorsMutex);
...
ac = p->second;
```

One smaller observation, pre-existing and covered by #6392: the `if (reapTask) { _timer->cancel(reapTask); }` in `~ActiveCommunicator` is dead code — a non-null `reapTask` means the task owns the `ActiveCommunicator`, so the destructor is unreachable with it set. The cycle is only ever broken by explicitly clearing `reapTask`, which is why this fix is needed in the first place.

bernardnormier:
@pepone good catch on the destruction order — `ac` was declared after the `lock_guard`, so in the cross-request unregister case the last reference dropped (and `Communicator::destroy()` blocked) while `_registeredCommunicatorsMutex` was held. Applied your suggestion in c43bfc50f7: `ac` is now declared before the lock, so the destroy runs after the lock is released, like the other teardown paths. Verified with the `registeredCommunicator` test.

PR #6382, opened 2026-08-04
Fix crash when the exception class lookup fails in PHP

### LINKED ISSUE #6371: IcePHP: createPHPException dereferences a null class entry when \Ice\LocalException is not defined
## Summary

`createPHPException` in `php/src/Util.cpp` guards its `\Ice\LocalException` fallback with nothing but an `assert`, so in a release build a null class entry is passed straight to `object_init_ex` and the process crashes.

```cpp
zend_class_entry* cls = nameToClass(result);
if (!cls)
{
    if (fallbackToLocalException)
    {
        cls = nameToClass("\\Ice\\LocalException");
        assert(cls);                                  // <-- compiled out with NDEBUG
    }
    else
    {
        ostringstream os;
        os << "unable to create PHP exception class for type ID " << typeId;
        runtimeError(os.str());
        return nullptr;
    }
}

if (object_init_ex(ex, cls) != SUCCESS)               // <-- cls may be null here
```

`php/src/Util.cpp:376-398`. The non-fallback branch handles the same condition properly by raising a `RuntimeException` and returning `nullptr`; only the fallback branch relies on the assert.

## Trigger

Any local exception raised while the Ice PHP library files are not on the include path — i.e. the extension is loaded but `Ice.php` was never required. The class lookup for the specific exception fails, the fallback lookup for `\Ice\LocalException` fails too, and `object_init_ex(ex, nullptr)` dereferences null.

```php
<?php
// note: no require 'Ice.php'
$c = Ice\initialize();
$c->destroy();
$c->stringToProxy("test");   // raises CommunicatorDestroyedException internally
```

```
$ php -d extension=php/lib/ice.so repro.php
$ echo $?
139
```

```
* frame #0: object_init_ex + 20
  frame #1: ice.so`(anonymous namespace)::createPHPException(_zval_struct*, char const*, bool) + 268
  frame #2: ice.so`IcePHP::convertException(_zval_struct*, std::exception_ptr) + 1752
  frame #3: ice.so`IcePHP::throwException(std::exception_ptr) + 40
  frame #4: ice.so`zim_Ice_Communicator_stringToProxy(...) + 1532
```

With `Ice.php` required the same script throws `Ice\CommunicatorDestroyedException` normally, so this is a misconfiguration-only path — but the failure mode should be a PHP error, not a segfault.

## Suggested fix

Give the fallback branch the same treatment as the other one: if `\Ice\LocalException` also resolves to null, raise via `runtimeError` and return `nullptr`. Callers already handle a null return.

Found while reviewing #6367; unrelated to that PR.

bernardnormier:
Fixes #6371.

The Ice exception classes of the PHP mapping are defined in PHP code, so they don't resolve when the extension is loaded without the Ice language library (a misconfiguration, but one that should produce a language-level error rather than a segfault).

### PHP

`createPHPException` guarded its `\Ice\LocalException` fallback lookup with nothing but an `assert`, so a release build passed a null class entry to `object_init_ex`. The fallback branch now takes the same error path as the non-fallback branch: `runtimeError` (which throws PHP's built-in `RuntimeException`) and a null return that callers already handle.

Fixing that exposed two latent bugs on the same conversion path, both zvals declared as uninitialized stack memory and destroyed on failure: `throwException`'s `zex` (while `convertException` leaves the zval untouched on failure, so the `Z_ISUNDEF` guard read an uninitialized type byte and threw garbage — "Need to supply an object when throwing an exception") and the `id` zval in `convertException`'s `RequestFailedException` branch (`zval_ptr_dtor` on uninitialized memory when `createIdentity` fails). Both are now initialized with `ZVAL_UNDEF`, matching the existing caller in Operation.cpp.

With these fixes, the repro from #6371 goes from SIGSEGV to an uncaught `RuntimeException: unable to create PHP exception class for type ID ::Ice::CommunicatorDestroyedException`.

This PR fixes the exception conversion path only — the path #6371 reported, and the one every failure funnels through. The mapping's other class lookups (`ice_getIdentity`, `ice_getEncodingVersion`, ...) are still guarded by asserts alone and can segfault the same way; they are tracked by #6394.

### Python

`createPythonException` has the same shape of defect: its `Ice.LocalException` fallback lookup has no null guard, and the failure path leaks the `args` references. However, the path is not reachable in practice: IcePy assumes the `Ice` package is importable throughout (roughly forty unguarded `lookupType` calls), so with the package missing, the first proxy operation crashes long before any exception conversion runs. Guarding two sites would not make the mapping fail gracefully, so the Python side is left unchanged here and the whole family is tracked by #6395. This PR only removes an inaccurate comment (see below).

### Ruby and MATLAB

Audited for the same defect: Ruby's `createRubyException` is safe — `callRuby(rb_path2class, ...)` throws a `RubyException` on lookup failure instead of returning a sentinel. MATLAB's `createMatlabException` fails through `mexCallMATLAB`, which raises a MATLAB-level error. No code change in either; this PR only removes a nonsensical comment that had been copied into the PHP, Python, and Ruby versions of this function ("should only ever be called on Ice local exceptions which don't use 'xxx:identifier'" — local exceptions are not defined in Slice in 3.8, so identifier metadata cannot apply to them).

### No changelog entry

The crash requires running the extension without its language library — a broken configuration in which nothing works; the fix only improves the failure mode. The full PHP, Python, and Ruby test suites pass.

### DIFF
--- a/php/src/Util.cpp
+++ b/php/src/Util.cpp
@@ -367,27 +367,24 @@ namespace
     zend_class_entry* createPHPException(zval* ex, const char* typeId, bool fallbackToLocalException = false)
     {
         // Convert the exception's typeId to its mapped PHP type by replacing "::Ice::" with "\Ice\".
-        // This function should only ever be called on Ice local exceptions which don't use 'php:identifier'.
         string result = typeId;
         assert(result.find("::Ice::") == 0);
         result.replace(0, 7, "\\Ice\\");
         assert(result.find(':') == string::npos); // Assert that there weren't any intermediate scopes.
 
         zend_class_entry* cls = nameToClass(result);
+        if (!cls && fallbackToLocalException)
+        {
+            cls = nameToClass("\\Ice\\LocalException");
+        }
         if (!cls)
         {
-            if (fallbackToLocalException)
-            {
-                cls = nameToClass("\\Ice\\LocalException");
-                assert(cls);
-            }
-            else
-            {
-                ostringstream os;
-                os << "unable to create PHP exception class for type ID " << typeId;
-                runtimeError(os.str());
-                return nullptr;
-            }
+            // The classes are defined in PHP code: they don't resolve when the application didn't load the Ice
+            // library files.
+            ostringstream os;
+            os << "unable to create PHP exception class for type ID " << typeId;
+            runtimeError(os.str());
+            return nullptr;
         }
 
         if (object_init_ex(ex, cls) != SUCCESS)
@@ -518,6 +515,7 @@ void
 IcePHP::throwException(std::exception_ptr ex)
 {
     zval zex;
+    ZVAL_UNDEF(&zex);
     convertException(&zex, ex);
     if (!Z_ISUNDEF(zex))
     {
--- a/php/src/Util.cpp
+++ b/php/src/Util.cpp
@@ -454,6 +454,7 @@ IcePHP::convertException(zval* zex, std::exception_ptr ex)
         setLongMember(zex, "replyStatus", static_cast<zend_long>(e.replyStatus()));
 
         zval id;
+        ZVAL_UNDEF(&id);
         if (!createIdentity(&id, e.id()))
         {
             zval_ptr_dtor(&id);
--- a/python/modules/IcePy/Util.cpp
+++ b/python/modules/IcePy/Util.cpp
@@ -550,30 +550,27 @@ namespace
     createPythonException(const char* typeId, std::array<PyObject*, N> args, bool fallbackToLocalException = false)
     {
         // Convert the exception's typeId to its mapped Python type by replacing "::Ice::" with "Ice.".
-        // This function should only ever be called on Ice local exceptions which don't use 'python:identifier'.
         string result = typeId;
         assert(result.find("::Ice::") == 0);
         result.replace(0, 7, "Ice.");
         assert(result.find(':') == string::npos); // Assert that there weren't any intermediate scopes.
 
         PyObject* type{IcePy::lookupType(result)};
+        if (!type && fallbackToLocalException)
+        {
+            type = IcePy::lookupType("Ice.LocalException");
+        }
         if (!type)
         {
-            if (fallbackToLocalException)
+            // The classes are defined in Python code: they don't resolve when the Ice module is not importable.
+            for (PyObject* pArg : args)
             {
-                type = IcePy::lookupType("Ice.LocalException");
+                Py_DECREF(pArg);
             }
-            else
-            {
-                for (PyObject* pArg : args)
-                {
-                    Py_DECREF(pArg);
-                }
 
-                ostringstream os;
-                os << "unable to create Python exception class for type ID " << typeId;
-                return PyObject_CallFunction(PyExc_Exception, "s", os.str().c_str());
-            }
+            ostringstream os;
+            os << "unable to create Python exception class for type ID " << typeId;
+            return PyObject_CallFunction(PyExc_Exception, "s", os.str().c_str());
         }
         IcePy::PyObjectHandle pArgs{PyTuple_New(N)};
         for (size_t i = 0; i < N; ++i)
--- a/python/modules/IcePy/Util.cpp
+++ b/python/modules/IcePy/Util.cpp
@@ -556,21 +556,23 @@ namespace
         assert(result.find(':') == string::npos); // Assert that there weren't any intermediate scopes.
 
         PyObject* type{IcePy::lookupType(result)};
-        if (!type && fallbackToLocalException)
-        {
-            type = IcePy::lookupType("Ice.LocalException");
-        }
         if (!type)
         {
-            // The classes are defined in Python code: they don't resolve when the Ice module is not importable.
-            for (PyObject* pArg : args)
+            if (fallbackToLocalException)
             {
-                Py_DECREF(pArg);
+                type = IcePy::lookupType("Ice.LocalException");
             }
+            else
+            {
+                for (PyObject* pArg : args)
+                {
+                    Py_DECREF(pArg);
+                }
 
-            ostringstream os;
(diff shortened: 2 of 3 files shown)


pepone (commented):
Looks good. I built the branch and ran both repros: the PHP one goes from SIGSEGV to `Uncaught RuntimeException: unable to create PHP exception class for type ID ::Ice::CommunicatorDestroyedException`, and the Python one to `Exception: unable to create Python exception class ...`. Success and normal-fallback paths are unchanged, and all seven `createPHPException` callers null-check. The `ZVAL_UNDEF(&zex)` fix is right — it matches the `Operation.cpp` caller. I also checked the Ruby and MATLAB claims and they hold: a failing fallback `rb_path2class` escapes `createRubyException` into `convertException`'s outer `catch (const RubyException& e) { return e.ex; }`, and `mexCallMATLAB` raises a fatal MATLAB error.

A few things before merging.

**The same uninitialized-zval bug is 60 lines below the one you fixed.** `php/src/Util.cpp:456`, in the `RequestFailedException` branch of `convertException`:

```cpp
zval id;
if (!createIdentity(&id, e.id()))
{
    zval_ptr_dtor(&id);   // destroys uninitialized stack memory
```

Worth folding in here — one line, `ZVAL_UNDEF(&id);`.

**In PHP the segfault is still one call away.** `\Ice\Identity` and `\Ice\EncodingVersion` are also defined in `Ice.php`, and their lookups are guarded by `assert` alone. On this branch, release build, no `Ice.php` required:

```php
$p = $c->stringToProxy("foo:tcp -h localhost -p 10000");
$p->ice_toString();            // fine
$p->ice_getIdentity();         // SIGSEGV — Util.cpp:189, createIdentity
$p->ice_getEncodingVersion();  // SIGSEGV — Util.cpp:78, createVersion
```

Same shape in `getVersion` (`Util.cpp:134`) and the proxy setters (`Proxy.cpp:148`, `Proxy.cpp:630`). A small `nameToClass`-or-`runtimeError` helper would cover all of them; otherwise let's file a follow-up, because as it stands the "language-level error rather than a segfault" guarantee doesn't hold for the mapping.

**The Python half is hardening rather than a reachable-crash fix.** IcePy assumes `Ice` is importable throughout — roughly forty unguarded `lookupType` calls. With the `Ice` import blocked, `stringToProxy` on a live communicator segfaults at `createProxy` → `lookupType("Ice.ObjectPrx")` (`Proxy.cpp:1479`) before any exception conversion happens; the path you fixed is only reachable via the destroy-then-call trick in the repro. And `IcePy::createIdentity` (`Util.cpp:720`) has the identical null-callable defect on the *same* conversion path — the `RequestFailedException` args are built at `Util.cpp:629` before `createPythonException` runs — so `ObjectNotExistException` still crashes. Keep the change, but I'd soften "a worse defect" in the description.

Nit: `Py_XDECREF` in the new cleanup loop (`Util.cpp:568`) — `createString` and `createIdentity` can both return null.

I disagree with Copilot's two test asks. A subprocess harness that loads the extension without its language library is a lot of machinery for a misconfiguration path; no test and no changelog entry are the right calls here.

bernardnormier:
Thanks for the thorough review — all points addressed in c4283c71f9:

- **Folded in the second uninitialized zval**: `ZVAL_UNDEF(&id)` in the `RequestFailedException` branch.
- **The other PHP class lookups**: went with the follow-up — the full family is larger than the proxy getters (also `extractIdentity`, `identityToString` ×2, `Ice\initialize`'s `InitializationData`, and the Types.cpp sites, each with different failure semantics), so it outgrows this PR. Filed #6394 with your repro and the nameToClass-or-raise helper as the suggested shape.
- **The Python half is withdrawn** rather than kept: since it's hardening on a path that's unreachable without the importlib trick, and the mapping-wide guarantee needs the ~40-site sweep anyway, Bernard preferred keeping this PR scoped to #6371 (which is PHP-only). Your findings — including the `IcePy::createIdentity` defect and the `Py_XDECREF` point — are captured in #6395. Only the comment removal remains on the Python side.
- **PR title and description updated** accordingly (no more "a worse defect", and the description now says explicitly that this fixes the exception conversion path only, pointing at #6394 for the rest).

Agreed on declining Copilot's subprocess-harness test asks.

pepone (approved):
All addressed — and the `createPHPException` restructure is better than what I suggested: instead of bolting a guard onto the fallback branch, both null cases now converge on one `runtimeError` + `nullptr` path.

Verified on a build of `c4283c71f`:

- The #6371 repro goes from **SIGSEGV to** `Uncaught RuntimeException: unable to create PHP exception class for type ID ::Ice::CommunicatorDestroyedException`.
- Full PHP suite: **20/20**.
- The comment removal is justified: there is no `slice/Ice/LocalException*.ice`, so local exceptions are not Slice-defined in 3.8 and identifier metadata cannot apply to them. (The only `*:identifier` under `slice/Ice` is a `Metrics.ice` data member.) Touching `python/` and `ruby/` purely to delete a comment in a now-PHP-scoped PR is a bit incongruous, but it is a real cleanup of a falsehood that had been copied into all three.

Not blocking, and equally present on `main` — one for #6394 rather than this PR. Its current framing ("class lookups") is too narrow for the "language-level error rather than a segfault" goal: without `Ice.php`, calling any operation on a proxy segfaults before reaching a single class lookup.

```php
$c = Ice\initialize();
$c->stringToProxy("nonexistent:tcp -h 127.0.0.1 -p 1")->ice_ping();  // SIGSEGV
```

`php/src/Proxy.cpp:1531-1534` — `ProxyInfoPtr info = _this->info; assert(info); info->getOperation(name->val);`. `info` is null because `::Ice::Object`'s `ProxyInfo` is registered by `Ice.php` through `IcePHP_defineProxy`; the fault address `0x68` is the `operations` map offset. Same defect shape as #6371, but a `nameToClass`-or-raise helper won't cover it. Added to #6394.

PR #6387, opened 2026-08-04
Ignore a Bluetooth connection failure delivered after close

### LINKED ISSUE #6377: IceBT TransceiverI::connectFailed lacks the _closed guard
`TransceiverI::connectCompleted` checks `_closed` and disposes of the late-delivered fd when the transceiver was already closed, but `connectFailed` has no such guard: it unconditionally stores the exception and calls `_stream->ready(SocketOperationConnect, true)`.

`ThreadPool::finish` clears the ready callback, so a `connectFailed` that arrives after the transceiver was closed (for example after a connect timeout) reaches `NativeInfo::ready()` with no callback installed and fails the `assert(_readyCallback)` — undefined behavior in release builds.

Noted by @pepone while reviewing #6362.

bernardnormier:
Fixes #6377.

`TransceiverI::connectFailed` now checks `_closed` under `_mutex`, mirroring `connectCompleted`: a failure delivered after the transceiver was closed is ignored.

Without the guard, the sequence is a crash in release builds:

1. A connect timeout (default 10 seconds) closes the connection: `ConnectionI::setState(StateClosed)` closes the transceiver — setting `_closed` — and the thread pool then clears the stream's ready callback (without a custom `Ice.Executor`, `ThreadPool::finish` runs `finished()` before `setReadyCallback(nullptr)`, so `_closed` is always set first; with an executor that queues the `finish(true)` call there is a pre-existing residual window where the callback is cleared before `_closed` is set — the same window that already defeats the `connectCompleted` guard, see the review discussion).
2. Nothing cancels the in-flight BlueZ `ConnectProfile` call — for a never-completed connect there is no `Connection` to close — so its result is always delivered eventually.
3. BlueZ routinely takes longer than the connect timeout to give up on an out-of-range or unresponsive device. The late failure reaches `connectFailed`, which called `_stream->ready(...)` on a stream whose ready callback was already cleared: `assert(_readyCallback)` in debug builds, a call through a null `shared_ptr` in release builds.

So "connect timeout against a dead peer, then crash" is the expected sequence, not a narrow race.

This is not a regression: the same unguarded `connectFailed` exists in 3.7 (with the same callback-clearing mechanism and the same `assert` in `NativeInfo::ready`), and in 3.8.0 `connectCompleted` was unguarded as well — #5701 introduced `_closed` and guarded the completed path only. A 3.7 backport is worth considering; it won't cherry-pick cleanly since 3.7's `connectFailed` has a different signature and locking, but the guard is equally simple there.

### DIFF
--- a/changelog.d/cpp/6377.md
+++ b/changelog.d/cpp/6377.md
@@ -1,0 +1,2 @@
+- Fixed a crash in the IceBT transport: a Bluetooth connection attempt that failed after the connection was
+  closed - for example following a connect timeout - could crash the program.
--- a/changelog.d/cpp/6377.md
+++ b/changelog.d/cpp/6377.md
@@ -1,2 +1,2 @@
-- Fixed a crash in the IceBT transport: a Bluetooth connection attempt that failed after the connection was
-  closed - for example following a connect timeout - could crash the program.
+- Fixed a crash in the IceBT transport. A Bluetooth connection attempt that failed after its connection was
+  closed — for example after a connect timeout — crashed the program.
--- a/cpp/src/IceBT/TransceiverI.cpp
+++ b/cpp/src/IceBT/TransceiverI.cpp
@@ -226,12 +226,18 @@ void
 IceBT::TransceiverI::connectFailed(std::exception_ptr ex)
 {
     lock_guard lock(_mutex);
-    //
-    // Save the exception - it will be rethrown in initialize().
-    //
-    _exception = ex;
-    //
-    // Wake up the thread pool, which resumes the connection establishment by calling initialize().
-    //
-    _stream->ready(IceInternal::SocketOperationConnect, true);
+
+    // Ignore a failure delivered after the transceiver was closed: there is no connection establishment left to
+    // resume.
+    if (!_closed)
+    {
+        //
+        // Save the exception - it will be rethrown in initialize().
+        //
+        _exception = ex;
+        //
+        // Wake up the thread pool, which resumes the connection establishment by calling initialize().
+        //
+        _stream->ready(IceInternal::SocketOperationConnect, true);
+    }
 }


externl (approved):

changelog.d/cpp/6377.md
```diff
@@ -0,0 +1,2 @@
+- Fixed a crash in the IceBT transport: a Bluetooth connection attempt that failed after the connection was
+  closed - for example following a connect timeout - could crash the program.
```
pepone: The other fragments use em dashes for asides, and the bullet repeats itself ("Fixed a crash … could crash").

```suggestion
- Fixed a crash in the IceBT transport. A Bluetooth connection attempt that failed after its connection was
  closed — for example after a connect timeout — crashed the program.
```

pepone (approved):
Approved. Verified the chain: `NativeInfo::ready()` derefs `_readyCallback` unconditionally (`Network.h:152-156`) and `ThreadPool::finish` clears it immediately after `finished()` (`ThreadPool.cpp:353-361`). Also confirmed it's not a regression — 3.7 has both paths unguarded, and 3.8 guards only `connectCompleted` (#5701).

One correction to the description:

> `ThreadPool::finish` runs `finished()` before `setReadyCallback(nullptr)`, so `_closed` is always set first; the guard therefore covers the whole window.

That holds only without `Ice.Executor`. With an executor that queues the call:

1. `setState(StateClosed)` calls `_threadPool->finish(this, false)`; `Selector::finish` returns `closeNow` unchanged (`Selector.cpp:394`), so `_transceiver->close()` at `ConnectionI.cpp:2176` is skipped.
2. The lambda runs `finished(current, true)`. For an async connect `_connectionStartFailed` is set, so the early return at `ConnectionI.cpp:1596` doesn't apply.
3. `executeFromThisThread` (`ThreadPool.cpp:404-410`) hands `finish(true)` to the executor, which may queue it and return.
4. The lambda then clears the ready callback — before `finish(true)` reaches `_transceiver->close()` and sets `_closed`.

So `_closed == false` and a null callback can coexist there, and it's a genuine data race on `_readyCallback` too — otherwise serialized by `TransceiverI::_mutex`. This is pre-existing and defeats the `connectCompleted` guard identically, so it's not a blocker. I just don't want the description on record claiming the window is fully closed, since that will shape the 3.7 backport.

Two follow-ups, both out of scope here:

- `AcceptorI::newConnection` (`AcceptorI.cpp:164`) has the same unguarded `ready()` — filed as #6441. `AcceptorI::close()` only unregisters the D-Bus profile, and `DBus::Connection::handleMessage` copies `_services` under the lock then dispatches without it (`DBus.cpp:1060-1071`), so an in-flight `NewConnection` still lands after `closeAcceptor()` cleared the callback. Trigger: adapter deactivated while a peer is connecting.
- Worth considering making `NativeInfo::ready()` a no-op when `_readyCallback` is null. If a cleared callback is legitimately reachable, the assert buys nothing, and that covers both IceBT callbacks, the acceptor, and the executor window in one place. Noted on #6441 rather than filed separately.

Changelog nit inline.

changelog.d/cpp/6377.md
bernardnormier: Applied, and updated the PR description to note the executor window (per your review) — thanks.

PR #6403, opened 2026-08-05
Fix `--interface` parsing order and remove an unreachable test filter

### LINKED ISSUE #6402: Test driver: `--interface` parsed before its attribute exists; `Slice/*` filter regex typo
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI dead-code audit and verified by re-reading the source on `main` (`deff5566cde`); same on the `3.8` branch. Test scripts only, no runtime impact.

## Summary

Two small defects in the test driver scripts.

### 1. `--interface` is parsed before its attribute exists

`Driver.__init__` calls the reflective `parseOptions` (`scripts/Util.py:3892`) and only afterwards assigns `self.interface = ""` (`:3919`). `parseOptions` routes options via `hasattr`, so during the base-class parse `--interface` — a Driver-declared option (`:3838`) — always falls through to `remaining`. It works today only because `LocalDriver`/`ControllerDriver` re-parse the leftovers after `Driver.__init__` returns; any driver that does not re-parse would silently ignore `--interface`.

### 2. `"Slice/*"` filter is a regex typo

`scripts/Component.py:124` lists `"Slice/*"` in the Swift iOS test filters, alongside proper regexes (`"Ice/.*"`, `"typescript/.*"`). As a regex it means `Slice` followed by zero or more `/`, which happens to prefix-match today's test ids but over-matches in general. Intended: `"Slice/.*"`.

## Fix

Move the `self.interface` default above the `parseOptions` call, and correct the filter to `"Slice/.*"`.

**Severity:** Low (test tooling only).

externl:
Fixes #6402.

- `Driver.__init__` assigned `self.interface` only after its `parseOptions` call, so the reflective parse never matched `--interface`; it worked only because `LocalDriver`/`ControllerDriver` re-parse the leftovers. The default now sits with the other option-backed attributes, before the parse.
- Removed the `SwiftMapping` branch in `Component.py`'s `getFilters`. It required `buildPlatform` to be `iphoneos`/`iphonesimulator`, but the method returns for exactly those platforms several branches earlier, so it never ran.

### DIFF
--- a/scripts/Component.py
+++ b/scripts/Component.py
@@ -121,7 +121,7 @@ class Ice(Util.Component):
             "iphoneos",
         ]:
             return (
-                ["Ice/.*", "IceSSL/configuration", "Slice/*"],
+                ["Ice/.*", "IceSSL/configuration", "Slice/.*"],
                 ["Ice/properties", "Ice/udp"],
             )
         return ([], [])
--- a/scripts/Component.py
+++ b/scripts/Component.py
@@ -116,14 +116,6 @@ class Ice(Util.Component):
             )
         elif isinstance(mapping, Util.JavaScriptMapping):
             return ([], ["typescript/.*"])
-        elif isinstance(mapping, Util.SwiftMapping) and config.buildPlatform in [
-            "iphonesimulator",
-            "iphoneos",
-        ]:
-            return (
-                ["Ice/.*", "IceSSL/configuration", "Slice/.*"],
-                ["Ice/properties", "Ice/udp"],
-            )
         return ([], [])
 
     def canRun(self, testId: str, mapping: Util.Mapping, current: Util.Driver.Current) -> bool:
--- a/scripts/Util.py
+++ b/scripts/Util.py
@@ -3883,6 +3883,7 @@ class Driver:
         self.rlanguages: list[str] = []
         self.failures: list[Any] = []
         self.keepLogs = False
+        self.interface = ""
         self.configs: dict[Mapping, Mapping.Config]
 
         logDir = os.path.join(toplevel, "logs")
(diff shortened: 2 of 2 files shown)


scripts/Component.py
```diff
@@ -121,7 +121,7 @@ def getFilters(self, mapping: Util.Mapping, config: Util.Mapping.Config) -> tupl
             "iphoneos",
         ]:
             return (
-                ["Ice/.*", "IceSSL/configuration", "Slice/*"],
+                ["Ice/.*", "IceSSL/configuration", "Slice/.*"],
```
pepone: This branch is unreachable: `getFilters` opens with `if config.buildPlatform in ["iphoneos", "iphonesimulator"]` (:74), which returns for every mapping, so `isinstance(mapping, SwiftMapping) and buildPlatform in [...]` is never reached. Suggest deleting lines 119-126 rather than fixing the regex in them.

Confirmed by this PR's own `ios on macos-26` run: `swift/Ice/properties` and `swift/Ice/udp` ran (this branch lists both as rfilters) and `swift/Slice/escape` didn't (only `Ice/.*` + `IceSSL/configuration` applied).

Dropping the branch removes the `Slice/*` string entirely, so the title and description would need a trim. If the Swift-iOS list is actually wanted, making it reachable is a real behavior change — separate PR.

pepone (approved):
Verified the `--interface` half against a base worktree: with a driver that doesn't re-parse the leftovers, `interface` goes `''` → `'1.2.3.4'`; `LocalDriver` and `ControllerDriver` are unchanged either way (`'1.2.3.4'`, endpoints `tcp -h 1.2.3.4`). Good fix.

One note on the `Slice/*` half below.

externl:
Good catch — the early return at `Component.py:74` covers both iOS platforms unconditionally, so the `SwiftMapping` branch could never be reached. Dropped lines 119-126 rather than fixing the regex inside them, and retitled accordingly.

Agreed that making that filter list effective is a behaviour change and belongs in its own PR if the Swift-iOS exclusions are actually wanted.

PR #6408, opened 2026-08-05
Document why create-xcode-project-targets is absent

### LINKED ISSUE #6404: Make.rules.Darwin: document why create-xcode-project-targets is absent
> [!NOTE]
> **AI-assisted finding — corrected after review.** The original report claimed the Xcode rules in `config/Make.rules.Darwin` were unreachable dead code. That was wrong: see @pepone's review on #6408.

## What is actually true

`make-projects` (`config/Make.project.rules:921-922`) dispatches twice per project:

- `create-<type>-project-targets` — undefined for `xcode`, expands to nothing
- `make-<type>-project` — `make-xcode-project` **is** defined and is invoked, building the iOS Test Controller (`cpp/test/ios/controller/Makefile.mk:3` sets `_type = xcode`)

Only the first dispatch is a no-op, and harmlessly so: `xcodebuild` compiles the sources, so an Xcode project has no per-source targets to create. Deleting the macros drops all four `test/ios/controller[...]` rules.

## Fix

Add a comment recording why there is no `create-xcode-project-targets`, so it does not read as a defect to the next reader. No code change.

**Severity:** Info (documentation).

externl:
Supersedes the original deletion in this PR, which was wrong — see @pepone's review.

`make-projects` (`config/Make.project.rules:921-922`) dispatches twice per project. Only the first dispatch (`create-xcode-project-targets`) is undefined for Xcode projects; the second (`make-xcode-project`) is defined here and does build the iOS Test Controller. Removing these macros dropped all four `test/ios/controller[...]` rules.

Rather than delete live code, this records why the first dispatch has no Xcode counterpart: `xcodebuild` compiles the sources, so there are no per-source targets to create.

Closes #6404.

### DIFF
--- a/config/Make.rules.Darwin
+++ b/config/Make.rules.Darwin
@@ -162,8 +162,9 @@ get-iostest-program-targets = $2/$1.bundle/$1.dylib $2/$1.bundle/Info.plist
 #
 # make-projects dispatches twice per project: once to create-<type>-project-targets and once to
 # make-<type>-project. Xcode projects deliberately define only the second: xcodebuild compiles the
-# sources itself, so there are no per-source targets to create and create-xcode-project-targets is
-# intentionally absent (make-projects falls back to expanding nothing for it).
+# sources itself, so there are no per-source targets to create. create-xcode-project-targets was an
+# empty macro until 9a02628927 removed it; $(or ...) still selects that name, and $(call) on an
+# undefined macro expands to nothing. It does not fall back to create-project-targets.
 #
 
 #
--- a/config/Make.rules.Darwin
+++ b/config/Make.rules.Darwin
@@ -160,6 +160,13 @@ get-iostest-program-targets = $2/$1.bundle/$1.dylib $2/$1.bundle/Info.plist
 ## Xcode projects
 
 #
+# make-projects dispatches twice per project: once to create-<type>-project-targets and once to
+# make-<type>-project. Xcode projects deliberately define only the second: xcodebuild compiles the
+# sources itself, so there are no per-source targets to create and create-xcode-project-targets is
+# intentionally absent (make-projects falls back to expanding nothing for it).
+#
+
+#
 # $(call make-xcode-project-with-config,$1=project,$2=platform,$3=config)
 #
 define make-xcode-project-with-config


pepone (commented):
These macros are still reachable — this removes live build code.

`make-projects` (`config/Make.project.rules:921-922`) dispatches twice:

```make
make-projects = $(foreach p,$1,$(eval $(call $(or $(if $($p_type),create-$($p_type)-project-targets),$3,create-project-targets),$p,$4)))\
                $(if $4,,$(foreach p,$1,$(eval $(call $(or $(if $($p_type),make-$($p_type)-project),$2,make-project),$p))))
```

Line 921 dispatches to `create-xcode-project-targets`, which is indeed undefined and expands to nothing. But line 922 dispatches to `make-$($p_type)-project` = `make-xcode-project`, which is defined here and does get invoked: `cpp/test/ios/controller/Makefile.mk:3` sets `_type = xcode`, and `cpp/Makefile:46-48` includes it and calls `make-projects` with an empty `$4`.

Comparing make's rule database on this branch vs `main`, using the iOS CI flags:

```shell
PLATFORMS='macosx iphonesimulator' CONFIGS='static shared' make -C cpp -p -n distclean | grep -c 'controller\['
# main:        4
# this branch: 0
```

`main` generates `test/ios/controller[iphonesimulator-static]` with the `xcodebuild … build` recipe, plus `test/ios/controller:: test/ios/controller[iphonesimulator-static]`. This branch generates none of it.

That rule is what builds the C++ Test Controller app:

- `.github/workflows/ci.yml:86` runs the iOS job with `PLATFORMS='macosx iphonesimulator' CONFIGS='static shared'`, and `.github/actions/build/action.yml:41` runs `make` at the repo root (`all` → `cpp` → `cpp/Makefile:60` `all:: srcs tests`).
- `scripts/Util.py:3526` (`getIOSAppFullPath`) only runs `xcodebuild -showBuildSettings` to read `TARGET_BUILD_DIR` — it does not build. Nothing else under `scripts/` builds that project.
- `cpp/BUILDING.md:173`/`:179` document `make PLATFORMS=iphonesimulator CONFIGS=all` as a supported workflow, feeding the `allTests.py --controller-app` flow at `:227`.

`make -C cpp -n srcs` can't detect this — the controller belongs to `tests`/`all`, not `srcs`, and the default `macosx` platform doesn't intersect `controller_platforms`.

#6404 is right that `create-xcode-project-targets` doesn't exist, so the first dispatch is a no-op for xcode projects. That's harmless — xcodebuild does the compiling, so the project needs no per-source targets — and worth a comment rather than a deletion.

externl:
You're right — thanks for catching this. I only probed the `create-xcode-project-targets` dispatch and never checked the second one, so I concluded the whole section was unreachable.

Reproduced your check: `0` controller rules on the deletion branch vs `4` on main. The PR now adds a comment explaining why `create-xcode-project-targets` has no Xcode counterpart instead of deleting the macros; rule count is back to 4.

config/Make.rules.Darwin
```diff
@@ -159,6 +159,13 @@ get-iostest-program-targets = $2/$1.bundle/$1.dylib $2/$1.bundle/Info.plist
 
 ## Xcode projects
 
+#
+# make-projects dispatches twice per project: once to create-<type>-project-targets and once to
+# make-<type>-project. Xcode projects deliberately define only the second: xcodebuild compiles the
+# sources itself, so there are no per-source targets to create and create-xcode-project-targets is
+# intentionally absent (make-projects falls back to expanding nothing for it).
```
pepone:
```suggestion
# make-projects dispatches twice per project: once to create-<type>-project-targets and once to
# make-<type>-project. Xcode projects deliberately define only the second: xcodebuild compiles the
# sources itself, so there are no per-source targets to create. create-xcode-project-targets was an
# empty macro until 9a02628927 removed it; $(or ...) still selects that name, and $(call) on an
# undefined macro expands to nothing. It does not fall back to create-project-targets.
```

"falls back" reads as if `$(or ...)` picks `create-project-targets` when the xcode one is missing. It doesn't -- it selects `create-xcode-project-targets` and stops; the emptiness comes from `$(call)` on an undefined macro:

```
selected-name=[create-xcode-project-targets]
call-result=[]
```

Since the comment exists to stop the next reader repeating the original mistake, worth being exact. The `9a02628927` reference also backs up "intentionally": it was an explicitly empty `define`/`endef` until that commit deleted it, an empty definition and no definition being identical to `$(call)`.

pepone (approved):
Thanks for turning this around. The comment is the right outcome -- one wording nit inline, non-blocking.

PR #6420, opened 2026-08-05
Build the unrecognized-layout message as one string

externl:
`set()` with two quoted arguments makes a list, so the not-found message for an unrecognized installation layout rendered with a stray `;` — `... is not in a recognized ;installation layout ...`. `string(CONCAT)` as suggested in the #6352 follow-up.


pepone (approved):
Correct, and it is the only one of its kind — the other six `Ice_NOT_FOUND_MESSAGE` assignments are single strings, and the `message(FATAL_ERROR "a" "b")` calls elsewhere concatenate, so nothing else in the package has the problem.

Rendered on the path this message actually serves — a RHEL host with the runtime package but not the devel package, which is #6440:

```
  Could not locate the Ice installation:
  /usr/lib64/cmake/Ice-3.9
  is not in a recognized installation layout, or the installation is missing
  Ice/Ice.h.
```

No `;`. A normal installation still resolves, with `Ice::IceGrid` linking `Ice::Glacier2`.

LGTM.

Issue #6378, opened 2026-08-04
slice2py: sequence metadata on a nested sequence doesn't emit the type-hint import


pepone:
`slice2py` resolves sequence metadata at every nesting level when it writes a type hint, but only at the top level when it decides which imports to emit. A custom-mapped sequence nested inside a dictionary or another sequence therefore gets a type hint referencing `array`, `numpy` or `Buffer` with no import at all — neither at runtime nor under `TYPE_CHECKING`.

This is the same class of bug as #6263, but a different code path: #6263 is about metadata applied *to a field*, this one is about a sequence *nested inside another constructed type*. It is unaffected by #6361.

```slice
module X
{
    ["python:array.array"] sequence<short> ShortSeq;
    dictionary<string, ShortSeq> ShortSeqDict;
    sequence<ShortSeq> SeqOfSeq;

    struct Box { ShortSeqDict d; }
    struct Nested { SeqOfSeq s; }

    interface I { ShortSeqDict op(ShortSeqDict d); }
}
```

`X/Box.py` — `array` is never imported:

```python
from X.ShortSeqDict import _X_ShortSeqDict_t

from dataclasses import dataclass
from dataclasses import field


@dataclass
class Box:
    """
    Notes
    -----
        The Slice compiler generated this dataclass from Slice struct ``::X::Box``.
    """
    d: dict[str, array[int]] = field(default_factory=dict)
```

`X/Nested.py` is the same with `s: list[array[int]]`.

`X/I.py` — neither `array` nor `Buffer` is imported, although both appear in every signature:

```python
if TYPE_CHECKING:
    from Ice.Current import Current
    from collections.abc import Awaitable
    from collections.abc import Mapping
    from collections.abc import Sequence

...
    def op(self, d: Mapping[str, Sequence[int] | Buffer], context: dict[str, str] | None = None) -> dict[str, array[int]]:
```

pyright:

```
Box.py      21  "array" is not defined
I.py        41  "Buffer" is not defined
I.py        41  "array" is not defined
I.py        44  "Buffer" is not defined
I.py        44  "array" is not defined
I.py        98  "array" is not defined
I.py        98  "Buffer" is not defined
I.py        98  "Buffer" is not defined
Nested.py   21  "array" is not defined
```

[...]

Issue #6264, opened 2026-07-29
slice2py: Buffer type hint emitted without an import for nested numeric sequences


pepone:
A marshaling-direction type hint for a numeric sequence includes `collections.abc.Buffer`, but the import for it is only registered when the sequence is the *top-level* type of a parameter or return value. Nested in a dictionary or another sequence, the hint is emitted with no import.

```slice
module N
{
    sequence<int> IntSeq;
    dictionary<string, IntSeq> DictOfIntSeq;

    interface J1 { void opd(DictOfIntSeq d); }
}
```

`N/J1.py`:

```python
def opd(self, d: Mapping[str, Sequence[int] | Buffer], context: dict[str, str] | None = None) -> None:
```

with no `Buffer` anywhere in the imports. pyright reports `"Buffer" is not defined (reportUndefinedVariable)`. Harmless at runtime because `from __future__ import annotations` keeps annotations unevaluated, but it is a type-check error, and with `reportUnusedImport` now enabled the Python job has no slack for it.

`sequence<IntSeq>` as a parameter reproduces it the same way. The interface must contain no other top-level numeric-sequence parameter, otherwise that one registers `Buffer` and masks the gap.

`ImportVisitor::visitInterfaceDefStart` only calls `addRuntimeImportForSequence` when the parameter or return type is itself a `Sequence` ([`PythonUtil.cpp:718`](https://github.com/zeroc-ice/ice/blob/main/cpp/src/slice2py/PythonUtil.cpp#L718), [`:734`](https://github.com/zeroc-ice/ice/blob/main/cpp/src/slice2py/PythonUtil.cpp#L734)); for a `Dictionary` it adds only `collections.abc.Mapping`. `addTypingImport` recurses through element types ([`:1015`](https://github.com/zeroc-ice/ice/blob/main/cpp/src/slice2py/PythonUtil.cpp#L1015), [`:1019`](https://github.com/zeroc-ice/ice/blob/main/cpp/src/slice2py/PythonUtil.cpp#L1019)) but never reaches `addRuntimeImportForSequence`, while `CodeVisitor::typeToTypeHintString` recurses with `forMarshaling` propagated and emits the `Buffer` alias at any depth ([`:309`](https://github.com/zeroc-ice/ice/blob/main/cpp/src/slice2py/PythonUtil.cpp#L309)).

InsertCreativityHere:
This was fixed by #6422.

PR #6422, opened 2026-08-05
Fix `slice2py` Not Generating Imports for Nested Collections

InsertCreativityHere:
This PR fixes both #6264 and #6378 which were 2 symptoms of the same singular bug:
The `ImportVisitor` wasn't recursively emitting imports for the `element` type of `sequence`s or the `value` type of `dictionary`s, only for the 'top-level' types of fields and parameters/returns were checked.

I updated the `ImportVisitor` to recursively check all the inner types of collections, fixing both bugs.


cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -614,20 +614,23 @@ Slice::Python::ImportVisitor::visitDataMember(const DataMemberPtr& p)
 
     // Add imports required for data member types.
 
-    // For fields with a type that is a Struct, we need to import it as a RuntimeImport, to
-    // initialize the field in the constructor. For other contained types, we only need the
-    // import for type hints.
     if (auto sequence = dynamic_pointer_cast<Sequence>(type))
     {
-        addRuntimeImportForSequence(sequence, parent, p->getMetadata());
+        // For fields with a type that is a sequence, we check if it will be using a special mapping.
```
InsertCreativityHere, line 619: `addRuntimeImportForSequence` was renamed to `addSequenceImports`. This function adds both `typing` and `runtime` imports, not just runtime imports.

Also added a call to `addTypingImport` which recurses into the sequence's element type.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -696,31 +699,13 @@ Slice::Python::ImportVisitor::visitInterfaceDefStart(const InterfaceDefPtr& p)
...
         if (ret)
         {
-            if (auto sequence = dynamic_pointer_cast<Sequence>(ret))
-            {
-                addRuntimeImportForSequence(sequence, p, op->getMetadata());
-            }
-            else if (dynamic_pointer_cast<Dictionary>(ret))
-            {
-                addTypingImport("collections.abc", "Mapping", p);
-            }
```
InsertCreativityHere, line 706: This logic was moved into `addTypingImport`, so that we add these imports to every type we visit.

Right now, since they're in the `visitInterfaceDefStart` method, they are only added once, when we visit the parameter's type, which skips over nested types: the bug.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -770,22 +755,17 @@ Slice::Python::ImportVisitor::visitConst(const ConstPtr& p)
...
     const MetadataList& localMetadata)
 {
     auto metadata = getSequenceMetadata(sequence, localMetadata);
     auto directive = metadata ? metadata->directive() : "";
 
-    // Whether the sequence is a field of the source type. Fields need runtime imports for their default factory,
-    // and are annotated with the sequence's own type. Everywhere else the sequence appears in an operation
-    // signature, where it only needs type hints, in the marshaling direction as well as the unmarshaling one.
-    auto isField = dynamic_pointer_cast<ClassDef>(source) || dynamic_pointer_cast<Struct>(source) ||
-                   dynamic_pointer_cast<Exception>(source);
```
InsertCreativityHere, line 785: Discerning if this is for a `field` like this was always flaky, and now doesn't work at all, where we visit nested types. So instead, we just pass in whether or not this was called from `visitDataMember` or `visitInterfaceDefStart`.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -834,15 +814,10 @@ Slice::Python::ImportVisitor::addRuntimeImportForSequence(
...
         {
-            // Otherwise, we have no idea what the type is so we just Any
+            // Otherwise, we have no idea what the type is so we just use 'Any'.
             addTypingImport("typing", "Any", source);
         }
     }
-    else
-    {
-        // This is required to import the sequence element type in case it is not a built-in type.
-        addTypingImport(sequence, source);
```
InsertCreativityHere, line 844: With the new structure, this isn't necessary anymore.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -979,33 +954,41 @@ Slice::Python::ImportVisitor::addTypingImport(
 }
 
 void
-Slice::Python::ImportVisitor::addTypingImport(const SyntaxTreeBasePtr& definition, const ContainedPtr& source)
```
InsertCreativityHere, line 982: This function always took a `TypePtr`, no need for it to `SyntaxTreeBasePtr`... it also now needs to take any `localMetadata` which could influence the mapping of sequences and dictionaries.

cpp/src/slice2py/PythonUtil.cpp
```diff
@@ -979,33 +954,41 @@ Slice::Python::ImportVisitor::addTypingImport(
...
+Slice::Python::ImportVisitor::addTypingImport(
+    const TypePtr& type,
+    const ContainedPtr& source,
+    const MetadataList& localMetadata)
 {
-    if (auto builtin = dynamic_pointer_cast<Builtin>(definition))
+    if (auto builtin = dynamic_pointer_cast<Builtin>(type))
     {
-        if (builtin->kind() == Builtin::KindValue)
+        if (builtin->kind() == Builtin::KindObjectProxy)
```
InsertCreativityHere, line 964: Just swapped the `if` and `else` here, to mirror the order we have in the other type-hint emitting logic.

pepone (approved):
Looks good. I verified it empirically:

- Built `slice2py` from this branch and from the merge-base, and diffed the generated output for all Slice files under `python/test` and `slice/Ice` (825 generated files). The only differences are the added imports in the new test types — every pre-existing file is byte-identical, including the numpy/memoryview/tuple custom-mapping tests.
- pyright over the #6264/#6378 repros plus the new test types: 69 undefined-name errors with the merge-base compiler, none with this branch, and no unused imports. A dict-of-numeric-sequence in return-only position still uses `Buffer` in the servant-side hint, so the `InterfaceDef`-gated import is never dead. Since the pyright CI job generates the test stubs, the new test types double as a regression gate for this fix.
- Nested numpy isn't covered by the new tests, but I checked it separately and the generated hints/imports match. Maybe worth a nested case in `TestNumPy.ice` for symmetry — your call.

Two comment nits, take or leave:

- `Test.ice`: "Numeric and re-mapped sequences cannot be used as parameter types" reads like a language restriction. They can be — the test just must not use them as top-level parameters, or the imports get registered without recursing and mask the bug.
- `AllTests.py`: "relies on imports for its field type hints" — the hints are never evaluated at runtime (`from __future__ import annotations`); the default constructor exercises the field default factories, and the hints are covered by the pyright job.

cpp/src/slice2py/PythonUtil.h
```diff
@@ -288,13 +288,15 @@ namespace Slice::Python
         }
 
     private:
-        /// Add the runtime imports for the given Sequence definition.
+        /// Add any necessary imports for the given Sequence definition.
         /// @param sequence The Sequence definition being imported.
         /// @param source The Slice definition that requires the import.
+        /// @param isFieldType Whether the provided sequence is the type of a field.
```
bernardnormier, line 294:
```suggestion
        /// @param isFieldType Whether the sequence is the declared type of a field. Only fields need runtime
        /// imports for the sequence's mapped type: their default factory constructs it. In any other position
        /// (parameter, return type, or nested inside another collection) the mapped type appears only in
        /// type hints.
```

cpp/src/slice2py/PythonUtil.h
```diff
@@ -288,13 +288,15 @@ namespace Slice::Python
...
         /// @param localMetadata Any additional metadata associated with the import.
```
bernardnormier:
```suggestion
        /// @param localMetadata The metadata applied where the sequence is used (field, parameter, or
        /// operation return type). Metadata applied at the point of use takes precedence over metadata
        /// applied to the sequence definition.
```

cpp/src/slice2py/PythonUtil.h
```diff
@@ -321,13 +323,17 @@ namespace Slice::Python
...
-        /// Adds a typing import for the package containing the given Slice definition.
+        /// Adds all the necessary typing imports for the given Slice type.
         ///
         /// Typing imports are generated inside an `if TYPE_CHECKING:` block, so they are only used for type hints.
         ///
-        /// @param definition The definition to import the containing package.
+        /// @param type The type to generate imports for.
         /// @param source The Slice definition that requires this import.
-        void addTypingImport(const SyntaxTreeBasePtr& definition, const ContainedPtr& source);
+        /// @param localMetadata The local metadata to consider when generating imports.
```
bernardnormier:
```suggestion
        /// @param localMetadata The metadata applied where the type is used (field, parameter, or
        /// operation return type). Metadata applied at the point of use takes precedence over metadata
        /// applied to the type definition.
```

python/test/Ice/custom/Test.ice
```diff
@@ -126,4 +128,50 @@ module Test
...
+    class NestedC
+    {
+        ShortSeq1Dict d;
+        ShortSeq1Seq s;
+    }
+
+    // These operations use numeric and re-mapped sequences, but only nested inside dictionaries.
+    // This tests that the import visitor correctly recurses through the dictionaries' value types when running.
+    // Numeric and re-mapped sequences cannot be used as parameter types: this would cause the import to be generated
+    // without recursing.
```
bernardnormier:
```suggestion
    // These operations use numeric and custom-mapped sequences nested inside dictionaries: they check that
    // the import visitor recurses into dictionary value types when it generates imports. The sequences must
    // not also appear as top-level parameter or return types in this interface, because any top-level use
    // registers their imports even without recursion.
```

python/test/Ice/custom/Test.ice
```diff
@@ -126,4 +128,50 @@ module Test
...
+        ByteSeq2Dict opByteSeq2Dict(ByteSeq2Dict v1);
+
+        NestedS opNestedS(NestedS v);
+        NestedC opNestedC(NestedC v);
+    }
+
+    // These operations use numeric and re-mapped sequences, but only nested inside other sequences.
+    // This tests that the import visitor correctly recurses through the sequences' element types when running.
+    // Numeric and re-mapped sequences cannot be used as parameter types: this would cause the import to be generated
+    // without recursing.
```
bernardnormier:
```suggestion
    // These operations use numeric and custom-mapped sequences nested inside other sequences: they check that
    // the import visitor recurses into sequence element types when it generates imports. The sequences must
    // not also appear as top-level parameter or return types in this interface, because any top-level use
    // registers their imports even without recursion.
```

bernardnormier (approved):
Looks good. A few doc-comment and comment wording suggestions.

InsertCreativityHere:
I updated all the mentioned comments based on both reviews, but edited down some of the more verbose/repeat sounding parts.

PR #6424, opened 2026-08-05
Remove dead code from the IceBT D-Bus wrapper

### LINKED ISSUE #6423: IceBT: ~185 lines of the private D-Bus wrapper have been dead since its initial commit
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI dead-code audit and verified by re-reading the source on `main` (`aea66c680fa`). Not compile-verified locally: IceBT is Linux/BlueZ-only and the audit ran on macOS. Linux CI compiles it on every PR.

## Summary

IceBT carries its own minimal libdbus C++ wrapper in `cpp/src/IceBT/DBus.h` / `DBus.cpp`. `DBus.h` is internal to `cpp/src/` and has exactly one consumer besides its own `.cpp` — `Engine.cpp` — so anything Engine does not call is unreachable. Roughly 185 lines are in that state.

Git history shows this was **dead from the wrapper's initial commit** ("initial commit of C++ Bluetooth transport"), not residue from a later refactor: `callAsync` never had a callback-passing caller, `clone()` never had a root caller, and `setCallback` has never been referenced outside the wrapper in any commit since. It reads as a general-purpose D-Bus binding written for consumers that never arrived.

### 1. The `AsyncCallback` completion layer (~60 lines)

`DBus.h:422-427,442,491`; `DBus.cpp:761,820,849,884,970`. All seven `callAsync` call sites in `Engine.cpp` (`:51`, `:575`, `:617`, `:696`, `:822`, `:833`, `:1165`) pass a single argument, so the `const AsyncCallbackPtr& = nullptr` default is always null, `_callback` is never non-null, and `setCallback` is never invoked. Nothing derives from `AsyncCallback`. Engine performs async D-Bus calls by polling the returned `AsyncResult`, never via callback.

**Note for whoever fixes this:** the libdbus notify hook (`pendingCallCompletedCallback`, `DBus.cpp:755`/`766`/`896`) is load-bearing — it drives completion for the polling path. Only the user-facing callback layer above it is dead.

### 2. The `Value::clone()` hierarchy (~30 lines)

`DBus.h:130` (pure virtual) plus overrides at `:161`, `:216`, `:246`, `:281`, `:342`. Every call site (`:216`, `:249`, `:250`, `:286`, `:347`) is inside another `clone()` override, so the recursion has no root caller. (`EndpointFactoryI::clone` in `EndpointI.cpp` is unrelated and live.)

### 3. Seven uncalled virtuals (~60 lines)

`isMethodReturn`, `getDestination`, `checkTypes`, `isPending`, `isComplete`, `removeFilter`, `getSessionBus` — each is exactly one declaration in `DBus.h` plus one definition in `DBus.cpp`, with no call anywhere. `getSessionBus` is the clearest: Bluetooth lives on the system bus and Engine only ever calls `getSystemBus`. `removeFilter` is dead because the single `addFilter` is for a process-lifetime object.

### 4. Smaller items (~35 lines)

- 15 unused `shared_ptr` aliases (`DBus.h:79,90,178-202`) — the underlying value classes stay live via the generic read/write switches; only the aliases are dead.
- `ArrayValue::toStringMap` (`DBus.h:305`) — Engine builds its own `VariantMap` via `extractProperties` instead.
- `ConnectionI::connection()` (`DBus.cpp`) and the `ExceptionI(const DBusError&)` overload (`DBus.cpp:34`), which no throw site binds to.

Adjacent Engine-side leftovers: `Engine::_mutex` (all lock sites belong to the nested `BluetoothService`, which has its own), the write-only `_adapterName` and `_defaultAdapterAddress`, and the uncalled `deviceExists` chain.

## Fix

Delete the above, preserving the notify hook.

**Severity:** Low (dead code; no behavior change).

externl:
Fixes #6423.

`DBus.h` is internal to `cpp/src` and `Engine.cpp` is its only consumer, so anything Engine does not call is unreachable. All of this has been dead since the wrapper's initial commit.

- The `Value::clone()` hierarchy — every call site was inside another `clone()` override, so the recursion had no root caller. (`EndpointFactoryI::clone` is unrelated and untouched.)
- Five uncalled virtuals: `isMethodReturn`, `getDestination`, `checkTypes`, `isPending`, `isComplete`.
- `ArrayValue::toStringMap`, `ConnectionI::connection`, and the `ExceptionI(const DBusError&)` overload.
- Engine-side leftovers: the unused `Engine::_mutex` (the lock sites belong to the nested `BluetoothService`, which has its own), the write-only `_defaultAdapterAddress`, and the uncalled `deviceExists` chain.

Scoped to self-contained removals per @bernardnormier's review. The entangled pieces — the `AsyncCallback` completion layer, `removeFilter`, `getSessionBus`, and the `shared_ptr` aliases — are left alone for the eventual wrapper refactor.

**Draft pending CI.** IceBT is Linux/BlueZ-only, so this does not compile on macOS and Linux CI is the first compiler to see it.

### DIFF
--- a/cpp/src/IceBT/DBus.cpp
+++ b/cpp/src/IceBT/DBus.cpp
@@ -1,6 +1,7 @@
 // Copyright (c) ZeroC, Inc.
 
 #include "DBus.h"
+#include <map>
 
 #include <dbus/dbus.h>
 
--- a/cpp/src/IceBT/DBus.cpp
+++ b/cpp/src/IceBT/DBus.cpp
@@ -1,11 +1,12 @@
 // Copyright (c) ZeroC, Inc.
 
 #include "DBus.h"
-#include <map>
 
 #include <dbus/dbus.h>
 
+#include <cassert>
 #include <condition_variable>
+#include <map>
 #include <mutex>
 #include <stack>
 #include <thread>
--- a/cpp/src/IceBT/DBus.cpp
+++ b/cpp/src/IceBT/DBus.cpp
@@ -31,8 +31,6 @@ namespace
     public:
         ExceptionI(const ErrorWrapper& w) { init(w.err); }
 
-        ExceptionI(const DBusError& err) { init(err); }
-
         ExceptionI(const string& s) : Exception(s) {}
 
     private:
@@ -154,12 +152,6 @@ namespace
             return t == DBUS_MESSAGE_TYPE_METHOD_CALL;
         }
 
-        [[nodiscard]] bool isMethodReturn() const override
-        {
-            const int t = ::dbus_message_get_type(const_cast<DBusMessage*>(_message));
-            return t == DBUS_MESSAGE_TYPE_METHOD_RETURN;
-        }
-
         [[nodiscard]] string getPath() const override
         {
             const char* s = ::dbus_message_get_path(const_cast<DBusMessage*>(_message));
@@ -178,12 +170,6 @@ namespace
             return s ? string(s) : string();
         }
 
-        [[nodiscard]] string getDestination() const override
-        {
-            const char* s = ::dbus_message_get_destination(const_cast<DBusMessage*>(_message));
-            return s ? string(s) : string();
-        }
-
         void write(const ValuePtr& v) override
         {
             DBusMessageIter iter;
@@ -199,17 +185,6 @@ namespace
             {
                 writeValue(value, &iter);
             }
-        }
-
-        [[nodiscard]] bool checkTypes(const vector<TypePtr>& types) const override
-        {
-            string msgSig = ::dbus_message_get_signature(_message);
-            string sig;
-            for (const auto& type : types)
-            {
-                sig += type->getSignature();
-            }
-            return sig == msgSig;
         }
 
         ValuePtr read() override
@@ -792,18 +767,6 @@ namespace
 
         ~AsyncResultI() { ::dbus_pending_call_unref(_call); }
 
-        bool isPending() const override
-        {
-            lock_guard lock(_mutex);
-            return _status == StatusPending;
-        }
-
-        bool isComplete() const override
-        {
-            lock_guard lock(_mutex);
-            return _status == StatusComplete;
-        }
-
         MessagePtr waitUntilFinished() const override
         {
             unique_lock lock(_mutex);
@@ -1050,8 +1013,6 @@ namespace
 
             _thread = std::thread(&ConnectionI::run, this);
         }
-
-        DBusConnection* connection() { return _connection; }
 
         DBusHandlerResult handleMessage(DBusMessage* m)
         {
--- a/cpp/src/IceBT/DBus.h
+++ b/cpp/src/IceBT/DBus.h
@@ -5,7 +5,6 @@
 
 #include <cassert>
 #include <cstdint>
-#include <map>
 #include <memory>
 #include <sstream>
 #include <string>
@@ -127,8 +126,6 @@ namespace IceBT::DBus
     public:
         [[nodiscard]] virtual TypePtr getType() const = 0;
 
-        [[nodiscard]] virtual ValuePtr clone() const = 0;
-
         [[nodiscard]] virtual std::string toString() const = 0;
 
     protected:
@@ -157,8 +154,6 @@ namespace IceBT::DBus
         PrimitiveValue(E val) : v(std::move(val)), kind(K) {}
 
         [[nodiscard]] TypePtr getType() const final { return Type::getPrimitive(kind); }
-
-        [[nodiscard]] ValuePtr clone() const final { return make_shared<PrimitiveValue>(v); }
 
         [[nodiscard]] std::string toString() const final
         {
@@ -213,8 +208,6 @@ namespace IceBT::DBus
 
         [[nodiscard]] TypePtr getType() const override { return _type; }
 
-        [[nodiscard]] ValuePtr clone() const override { return make_shared<VariantValue>(v ? v->clone() : nullptr); }
-
         [[nodiscard]] std::string toString() const override { return v ? v->toString() : "nil"; }
 
         ValuePtr v;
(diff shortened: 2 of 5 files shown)


bernardnormier (commented):
My recommendation is to trim this down to the obvious, self-contained dead code and leave the entangled pieces alone. The D-Bus wrapper reads as a general-purpose binding, and excising it piece by piece leaves it neither minimal nor coherent — the proper cleanup there is a wholesale refactor, not spot removals.

Fine to remove (self-contained, touch nothing else):
- the `clone()` hierarchy
- the uncalled virtuals (`isMethodReturn`, `getDestination`, `checkTypes`, `isPending`, `isComplete`)
- `toStringMap`, `ConnectionI::connection()`, the `ExceptionI(const DBusError&)` overload
- the Engine-side leftovers: the `deviceExists` chain, `_defaultAdapterAddress`, `Engine::_mutex`

Would revert:
- the `AsyncCallback` completion layer — removing it rewires `callAsync`'s signature, the `AsyncResultI` constructor, and `replyReceived`; that's a redesign of the wrapper's completion story, which belongs in the eventual refactor
- `removeFilter` — coupled to the connection/filter lifecycle (which has its own quirks: `close()` clears `_services` but not `_filters`)
- `getSessionBus` and the `connect(bool)` → `connect()` signature change — same reasoning, it's an entry point of the wrapper, not a stray
- the 15 `Ptr` aliases — unused, but `using XxxPtr = std::shared_ptr<Xxx>;` is our general style for shared_ptr, so I'd keep them regardless

externl:
Agreed — trimmed to your keep-list. Reverted the `AsyncCallback` layer (so `callAsync`, the `AsyncResultI` constructor and `replyReceived` are untouched), `removeFilter`, `getSessionBus` and the `connect(bool)` signature, and restored the 15 `Ptr` aliases.

Now a pure deletion: −122 lines, no signature changes anywhere. The only knock-on is dropping `#include <map>` from `DBus.h`, whose sole user was `toStringMap`.

Also noting the `close()` asymmetry you mention — `_services` cleared but not `_filters` — is not touched here; worth its own look during the refactor.

bernardnormier (approved):
Looks good

pepone (approved):
> [!NOTE]
> AI-assisted review. The clang-tidy diagnosis was verified against the failing job log and reproduced locally with clang-tidy; the dead-code checks were verified by re-reading the source at 6c595b7.

The removals all check out — approving. One thing to sort out before merge: this turns the `clang-tidy` job red, for a reason that isn't visible in the diff.

## clang-tidy failure

```
src/IceBT/Engine.h:56:9: error: function 'communicator' should be marked [[nodiscard]]
src/IceBT/Engine.h:59:9: error: function 'initialized' ...
src/IceBT/Engine.h:61:9: error: function 'getDefaultAdapterAddress' ...
src/IceBT/Engine.h:62:9: error: function 'adapterExists' ...
src/IceBT/Engine.h:72:9: error: function 'getDevices' ...
   [modernize-use-nodiscard,-warnings-as-errors]
```

None of those five declarations are touched here. The trigger is dropping `mutable std::mutex _mutex;`: `modernize-use-nodiscard` skips const member functions of any class that has a mutable field, so `Engine`'s const methods were silently exempt until now. Repro:

```cpp
struct WithMutable { mutable int m = 0; int f() const { return m; } };  // not flagged
struct NoMutable   { int n = 0;         int f() const { return n; } };  // flagged
```

**Fix:** add `[[nodiscard]]` to those five declarations in `Engine.h`. Every call site already uses the result (`Instance.cpp:11,32`, `AcceptorI.cpp:86,191,204`, `ConnectorI.cpp:18`, `PluginI.cpp:97`), and the neighbouring classes already do this (`Instance.h:20`, `PluginI.h:31`). `BluetoothService` keeps its own mutable mutex, so its const methods stay exempt — no further errors expected.

The other three red jobs are unrelated flakes: `debug on macos-26` = C# `Ice/inactivityTimeout`, `release on windows-2022` = Java `Ice/idleTimeout`, plus `matlab-analyzer`.

## The dead-code claims hold

- `DBus.h` is included only by `DBus.cpp` and `Engine.cpp`, and neither it nor `Engine.h` is in `cpp/include/IceBT` — no public API/ABI change.
- No repo-wide references remain to any removed name. The only surviving `clone` in IceBT is the unrelated `EndpointFactoryI::clone`, and no `Value`/`Message`/`AsyncResult` subclass exists outside these two files, so no override is orphaned.
- `DBusError` now appears only as `ErrorWrapper::err` and `init()`'s parameter, and `ErrorWrapper` has no conversion from `DBusError` — so dropping that overload can't silently rebind a call to the `ErrorWrapper` one (which would `dbus_error_free` the error).
- `_status` / `StatusPending` / `StatusComplete` are still used (`DBus.cpp:760,774,791,821,826`) after `isPending`/`isComplete` go.
- Every `lock_guard lock(_mutex)` in the old `Engine.cpp` is inside `BluetoothService` (all at or below line 1209; the class spans 202–1233 and keeps its own mutex). The `IceBT::Engine::` definitions start at 1236 and never touched it.
- `_defaultAdapterAddress` really was write-only: declaration plus one `.clear()`, never read, never assigned.
- The include moves are right — `DBus.cpp` does use `map` (`map<string, ServicePtr>` at 1021/1082), and no other IceBT translation unit uses `map` or `mutex` beyond `AcceptorI.h`, which already includes its own.

## Nits

1. `DBus.h` no longer uses `assert` but still includes `<cassert>`. Don't just drop it — `Engine.cpp` has ~20 asserts and no `<cassert>` of its own, so it would need one first if you want the same treatment `<map>` got.
2. `DBus.cpp` puts `#include <map>` under `"DBus.h"` rather than in the std block below it, and `TransceiverI.h` appends `<mutex>` to the project-header block without the blank line `AcceptorI.h` uses. clang-format won't reorder across blank lines. `Engine.cpp` got it right.
3. `Connection::callAsync`'s comment still says the AsyncResult "can be used to determine completion status" — that's `waitUntilFinished`/`getReply` only now.

externl:
Thanks — reproduced the `modernize-use-nodiscard` interaction locally with your probe before fixing, and it behaves exactly as you describe: the `mutable` field was the only thing exempting `Engine`'s const methods. Added `[[nodiscard]]` to all five; every call site already used the result.

Nits all addressed:

1. `<cassert>` moved to `DBus.cpp` and `Engine.cpp` (49 and 26 asserts respectively, neither had it) and dropped from `DBus.h`.
2. `DBus.cpp`'s `<map>` moved into the standard-header block, and `TransceiverI.h` now has the blank line before `<mutex>` like `AcceptorI.h`.
3. `callAsync` comment updated — it now describes waiting for and obtaining the reply, plus the callback, with no mention of completion status.

PR #6426, opened 2026-08-05
Check the IcePy stub against the IcePy module

externl:
IcePy's documentation exists twice: Sphinx reads the docstrings the module ships, pyright and the IDEs read `IcePy-stubs/__init__.pyi`. Nothing links them, so they drift silently — they had.

- Add `scripts/checkIcePyStub.py`, run after the Pyright job, which already builds the extension. It imports IcePy and compares the docstrings actually shipped rather than parsing the C++, so there is no name mapping to get wrong. It compares signatures as well as prose: autodoc cannot introspect a C extension and takes the signature from the first line of each docstring, so that line drifts just as easily.
- `Connection.close` never said that awaiting the future re-raises whatever closed the connection, so the published documentation omitted it.
- `DispatchCallback.response` was documented as taking `*args`, but the implementation parses exactly one object.
- `createProperties` and `getProcessLogger` named return types their implementations do not return — `getProcessLogger` may return a Python logger directly rather than a wrapper, as its own comment says.
- `ice_invoke` and `ice_invokeAsync` omitted the default for `ctx`.
- `getEndpoint` left `Endpoint` unqualified, `ice_identity` called its parameter `id`, and `setAdapter` described the exception it raises differently in each copy.

All nine are resolved in favour of the stub, which was consistently the better maintained side. Only the types re-exported raw reach the published documentation (`Ice.Connection` *is* `IcePy.Connection`); the `createProperties`, `getProcessLogger` and `ice_invoke` fixes reach `help()` and IDE hover.

The alternative, generating the stub from the module (#4947), is a bigger change than it looks; I have left the numbers on that issue.

### DIFF
--- a/python/modules/IcePy/Connection.cpp
+++ b/python/modules/IcePy/Connection.cpp
@@ -65,7 +65,8 @@ Starts a graceful closure of this connection once all outstanding invocations ha
 Returns
 -------
 Awaitable[None]
-    A future that becomes available when the connection is closed.)";
+    A future that becomes available when the connection is closed. If the connection was lost or
+    aborted, awaiting this future raises the exception that caused the connection's closure.)";
 
     constexpr const char* connectionCreateProxy_doc = R"(createProxy(identity: Ice.Identity) -> Ice.ObjectPrx
 
@@ -112,8 +113,8 @@ adapter : Ice.ObjectAdapter | None
 Raises
 ------
 LocalException
-    If this connection is an incoming (server) connection: setAdapter can only be called on outgoing
-    connections.)";
+    When this connection is an incoming (server) connection: only outgoing (client) connections
+    support setting the object adapter.)";
 
     constexpr const char* connectionGetAdapter_doc = R"(getAdapter() -> Ice.ObjectAdapter | None
 
@@ -205,7 +206,7 @@ Returns
 Ice.ConnectionInfo
     The connection information.)";
 
-    constexpr const char* connectionGetEndpoint_doc = R"(getEndpoint() -> Endpoint
+    constexpr const char* connectionGetEndpoint_doc = R"(getEndpoint() -> Ice.Endpoint
 
 Gets the endpoint from which the connection was created.
 
--- a/python/modules/IcePy/Connection.cpp
+++ b/python/modules/IcePy/Connection.cpp
@@ -113,7 +113,7 @@ adapter : Ice.ObjectAdapter | None
 Raises
 ------
 LocalException
-    When this connection is an incoming (server) connection: only outgoing (client) connections
+    If this connection is an incoming (server) connection: only outgoing (client) connections
     support setting the object adapter.)";
 
     constexpr const char* connectionGetAdapter_doc = R"(getAdapter() -> Ice.ObjectAdapter | None
--- a/scripts/checkIcePyStub.py
+++ b/scripts/checkIcePyStub.py
@@ -13,9 +13,9 @@ The signature is compared as well as the prose. autodoc cannot introspect a C ex
 the first line of each docstring -- so that line is a signature the stub also spells out, and it can
 drift just as easily.
 
-Run it after building Ice for Python:
+Run it from the repository root after building Ice for Python:
 
-    PYTHONPATH=python python3 checkIcePyStub.py
+    PYTHONPATH=python/python python3 scripts/checkIcePyStub.py
 """
 
 from __future__ import annotations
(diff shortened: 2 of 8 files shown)


externl:
All four fixed in abaf3caf4e0. The two checker gaps were real: a public name missing from the module and a deleted signature line both passed before. Both now fail — verified by injecting each drift. The `Returns` prose in `createProperties` and `getProcessLogger` now matches their signatures.

pepone (approved):
Built the branch and red/green'd the checker: green as-is, and reverting the four IcePy `.cpp` files to main reproduces exactly the nine divergences. The tool does what it says.

**clang-format is red.** `Proxy.cpp:1423` is 124 columns and `:1429` is 123, against `ColumnLimit: 120`. clang-format's own fix splits `ice_invokeAsync` mid-value (`| None = ` / `"None) -> "`); splitting at the comma instead is a fixed point under clang-format 19:

```cpp
     PyDoc_STR("ice_invoke(operation: str, mode: Ice.OperationMode, inParams: bytes, "
               "ctx: dict[str, str] | None = None) -> tuple[bool, bytes]")},
...
     PyDoc_STR("ice_invokeAsync(operation: str, mode: Ice.OperationMode, inParams: bytes, "
               "ctx: dict[str, str] | None = None) -> Awaitable[tuple[bool, bytes]]")},
```

**Two small things worth fixing here:**

1. `Init.cpp` — the `createProperties` signature line now says `defaults: Ice.Properties | None = None`, but the `Parameters` section three lines below still says `defaults : Properties | None`. `propertiesInit` requires the `Ice.Properties` wrapper and reads its `_impl`, so the prose is the wrong one. The check only compares the signature line, not the numpydoc body, so it can't catch this.

2. `checkIcePyStub.py:112` truncates both sides to 100 characters, but the first difference is at character 194 for `Connection.close` and 642 for `Connection.setAdapter` — so the report prints two identical-looking strings and hides what actually drifted. A `difflib.unified_diff` over `splitlines()` would make it actionable.

**The description slightly oversells the reach.** `conf.py` documents `PACKAGES = ("Ice", "Glacier2", "IceBox", "IceGrid", "IceMX", "IceStorm")`, not `IcePy`. The premise holds for the 17 types `IcePyTypes.py` re-exports raw — `Ice.Connection` *is* `IcePy.Connection`, so the `Connection.close` sentence genuinely was missing from the published docs. But `Ice.createProperties` and `Ice.getProcessLogger` are Python functions with their own docstrings, and `ice_invoke`/`ice_invokeAsync` aren't published at all (no `:inherited-members:`) — those fixes reach only `help()` and IDE hover.

**Filed as follow-ups, nothing needed here:** #6443 for the check's blind spots — the live one is that `stubEntries()` collects only `FunctionDef`, so the stub's 25 annotated data attributes are never compared and two already differ from the module. #6442 for a class this check can't see by construction: both copies agree and both are wrong about the callable. #6444 for the 3.9 direction — worth noting that making IcePy private doesn't remove the duplication by itself, since `Ice.Connection` stays a C type Sphinx reads docstrings from; it does make the compiled copy disposable, so sourcing docs from the stub in `conf.py` would close it out far more cheaply than wrapping the 17 types.

Approach is right for 3.8 — #4947's numbers make stubgen not patch-sized, and docstring codegen means new build machinery in a patch release.

externl:
All three in a37826e8358 — the comma split verifies clean under clang-format 19.1.7, and drift now prints as a unified diff (a one-word change deep in `setAdapter` renders as exactly that line). Also qualified the description's reach per your note. Thanks for filing the follow-ups.

python/modules/IcePy/Connection.cpp
```diff
@@ -112,8 +113,8 @@ adapter : Ice.ObjectAdapter | None
 Raises
 ------
 LocalException
-    If this connection is an incoming (server) connection: setAdapter can only be called on outgoing
-    connections.)";
+    When this connection is an incoming (server) connection: only outgoing (client) connections
+    support setting the object adapter.)";
```
InsertCreativityHere: Everywhere in this file (and generally in our Python source code), we use `if...` for exceptions, not `when...`. The fact we use `When` in the `IcePy-stubs` is a mistake.

scripts/checkIcePyStub.py
InsertCreativityHere, line 1: Only skimmed the file, but looks reasonable.

I do think it should be located somewhere else.
Top-level in the `python` folder makes it seem more important than it is.
Maybe in the `scripts/` folder? Maybe in `python/docs` since it's docs-related?
Dealer's choice.

InsertCreativityHere (approved):
Looks good to me.

python/modules/IcePy/Connection.cpp
externl: Fixed in 57a0ab3 — "If ..." now, in the module docstring and the stub together (the checker compares them verbatim).

scripts/checkIcePyStub.py
externl: Moved to `scripts/checkIcePyStub.py` in 57a0ab3; the workflow and the script's own stub path updated with it.

PR #6427, opened 2026-08-05
Add missing ice_* method redeclarations to Python's Ice.ObjectPrx

### LINKED ISSUE #6364: Python: ice_invoke and ice_invokeAsync are missing from Ice.ObjectPrx
`Ice.ObjectPrx` (`python/python/Ice/ObjectPrx.py`) subclasses the C extension type
`IcePy.ObjectPrx` and redeclares each inherited `ice_*` method with type hints and a
numpydoc docstring that delegates to `super()`. That file is the documented surface of
the Python proxy API — the Sphinx pages are generated with `:members:` and no
`:inherited-members:` (see `python/docs/conf.py`), so a method that isn't redeclared
there never shows up in the API reference.

`ice_invoke` and `ice_invokeAsync` were never redeclared. They work at runtime, and
they're typed in `python/python/IcePy-stubs/__init__.pyi`, but:

- their only documentation is the signature-only `PyDoc_STR` in `python/modules/IcePy/Proxy.cpp`
- they're absent from the generated Python API docs
- they're invisible to anyone reading `ObjectPrx.py` as the definition of the proxy API

They should be added alongside the other `ice_*` methods, with full doc comments. Those
doc comments should also cover the empty in-params special case described in #6356.

Three other public methods are missing for the same reason and could be fixed at the same
time: `ice_getConnectionAsync`, `ice_flushBatchRequestsAsync`, and `ice_toString`.

externl:
`Ice.ObjectPrx` redeclares each inherited `IcePy.ObjectPrx` method with type hints and a numpydoc docstring, and only redeclared methods appear in the generated API reference. Five methods were never redeclared. Closes #6364.

- Adds `ice_invoke`, `ice_invokeAsync`, `ice_getConnectionAsync`, `ice_flushBatchRequestsAsync`, and `ice_toString` to `Ice.ObjectPrx`, delegating to `IcePy.ObjectPrx`.
- Documents the empty in-params special case (#6356) in the `ice_invoke`/`ice_invokeAsync` doc comments.
- `ice_invoke`/`ice_invokeAsync` keep the `ctx` parameter name from the C extension's keyword list, so existing keyword calls keep working.

### DIFF
--- a/python/python/Ice/ObjectPrx.py
+++ b/python/python/Ice/ObjectPrx.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
     from .IcePyTypes import Connection, Endpoint
     from .Identity import Identity
     from .Locator import LocatorPrx
+    from .OperationMode import OperationMode
     from .Router import RouterPrx
 
     T = TypeVar("T", bound=ObjectPrx)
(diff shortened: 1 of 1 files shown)


bernardnormier (approved):
Looks good

pepone (approved):
Checked the doc comments against the repo's Python numpydoc conventions and against the other mappings — consistent on both counts. The empty-in-params sentence localizes the #6356 wording correctly for `bytes`, `ice_getConnectionAsync` faithfully mirrors the current `ice_getConnection` and the C++ comment (closed-or-closing, fixed proxy, collocation remark), `ice_toString` matches C++ verbatim, and keeping `ctx` is the right call given `blobject/Client.py` passes it by keyword.

One thing to flag: the `ice_invokeAsync` sentence

> When this proxy is a oneway, datagram, or batch proxy, the flag is always `True` and the encapsulation is empty.

is true for the sync `ice_invoke`, but not yet for the async one — IcePy completes the awaitable with `None` on every non-twoway proxy:

```
oneway         sync : (True, b'')     oneway         async: None
batchOneway    sync : (True, b'')     batchOneway    async: None
datagram       sync : (True, b'')     datagram       async: None
batchDatagram  sync : (True, b'')     batchDatagram  async: None
```

C++, Java, and Swift all synthesize the tuple at the future boundary and IcePy doesn't, so this is a Python-mapping gap rather than a doc error. Filed as #6438 for 3.8.3 — the doc as written matches the Java/Swift contract and becomes accurate once that lands, so no change needed here.

Unrelated, but it came up while comparing: your `ice_getConnectionAsync` wording ("completes when the connection is available") is more accurate than C++'s "when the invocation completes", which C++ itself contradicts two overloads earlier ("no request is sent to get a connection"). Filed as #6439 for C++ and MATLAB — feel free to fold it in here if you'd rather not carry a separate PR.

Issue #6329, opened 2026-07-31
`--depend` Makefile output does not escape make-significant characters in paths


pepone:
> [!NOTE]
> **AI-assisted finding.** Surfaced while fixing #6122 and reproduced at runtime against `main` (`c203ba3c8b`) with
> the command below, so this is not a static-review estimate.

## Summary

`Slice::DependencyGenerator::writeMakefileDependencies` (`cpp/src/Slice/SliceUtil.cpp:715-742`) writes the target and
every prerequisite verbatim:

```cpp
os << target << ": \\" << std::endl;
for (const auto& dependency : dependencies)
{
    os << " " << dependency;
```

Make splits a rule's prerequisite list on whitespace, and treats `#` as a comment introducer and `$` as the start of a
variable reference. A Slice file reached through a path containing any of those characters produces a rule that means
something other than what was intended — silently, with the compiler exiting 0.

This is the third dependency-output writer with an escaping gap; the JSON and XML ones were fixed in #6122 / #6323.

## Reproducing

```console
$ mkdir -p 'sp/my project'
$ printf '#pragma once\nmodule Test { class A {} }\n' > 'sp/my project/A.ice'
$ printf '#include "A.ice"\nmodule Test { interface I { void op(A a); } }\n' > 'sp/my project/Main.ice'
$ slice2cpp --depend 'sp/my project/Main.ice'
Main.h: \
 /tmp/sp/my project/Main.ice \
 /tmp/sp/my project/A.ice
```

Make reads four prerequisites here — `/tmp/sp/my`, `project/Main.ice`, `/tmp/sp/my`, `project/A.ice` — none of which
exist, so the rule is either wrong or fatal depending on the makefile.

## Impact

Low in practice: this only bites a build tree whose absolute path contains a space, `#` or `$`, and it is the C++ and
scripting-language makefile builds that consume `--depend`. It is nevertheless a silent wrong answer rather than an
error.

Affects every compiler that offers `--depend`: `slice2cpp`, `slice2cs`, `slice2js`, `slice2php`, `slice2py`,
`slice2swift`. (`slice2java` only offers `--depend-xml`.)

## Fix

Escape make-significant characters when writing the target and the prerequisites: `$` → `$$`, `#` → `\#`, and space →
`\ `. Note that GNU make's support for escaped spaces in prerequisites is partial, so it may be worth deciding
explicitly whether spaces are supported or should be diagnosed as an error instead.

**Severity:** Low.

[...]

PR #6428, opened 2026-08-05
Fail slice2cpp_generate at configure time when called from another directory

### LINKED ISSUE #6419: slice2cpp_generate silently generates nothing when called from another directory
`slice2cpp_generate(<target>)` can be called from a different directory than the one the target was
created in — `add_library(feature ...)` in `lib/CMakeLists.txt`, `slice2cpp_generate(feature)` in the
parent. `target_sources()` accepts that, but `add_custom_command(OUTPUT ...)` is directory-scoped, so
the rule that generates the files is attached to the calling directory while the target that consumes
them lives elsewhere.

The generated sources are then added to the target with nothing to produce them, and the build fails on
the missing `.h`/`.cpp` rather than on the misuse.

Either document that the call has to happen in the target's own directory, or detect it and say so:
`get_target_property(<var> <target> SOURCE_DIR)` compared against `CMAKE_CURRENT_SOURCE_DIR` is enough
to catch it at configure time, alongside the alias and non-compiling-target guards already there.

externl:
`slice2cpp_generate` attaches its `add_custom_command(OUTPUT)` rules to the calling directory, so calling it from a directory other than the one that created the target left the generated sources with nothing to produce them, and the failure only surfaced at build time as a missing generated header. Closes #6419.

- Rejects the call at configure time when the target's `SOURCE_DIR` is not the calling directory, alongside the existing alias and target-type guards.
- Documents the requirement in the file's header comment.

### DIFF
--- a/cpp/config/slice2cpp.cmake
+++ b/cpp/config/slice2cpp.cmake
@@ -72,8 +72,11 @@ function(slice2cpp_generate target)
   get_target_property(target_source_dir ${target} SOURCE_DIR)
   if(NOT target_source_dir STREQUAL CMAKE_CURRENT_SOURCE_DIR)
     message(FATAL_ERROR
-      "slice2cpp_generate: '${target}' was created in '${target_source_dir}'; call slice2cpp_generate "
-      "from that directory.")
+      "slice2cpp_generate: '${target}' was created in '${target_source_dir}', but this call is in "
+      "'${CMAKE_CURRENT_SOURCE_DIR}'. The rule that compiles the Slice files belongs to the directory "
+      "that calls slice2cpp_generate, so nothing would ever generate the sources it adds to "
+      "'${target}'. Move this call next to the add_library() or add_executable() that created "
+      "'${target}'.")
   endif()
 
   if(arg_UNPARSED_ARGUMENTS)
--- a/cpp/config/slice2cpp.cmake
+++ b/cpp/config/slice2cpp.cmake
@@ -1,7 +1,8 @@
 # Copyright (c) ZeroC, Inc.
 
 # Compiles the Slice (.ice) files in a target's sources and adds the generated C++ to the target.
-# Only sees the sources present when it is called.
+# Only sees the sources present when it is called, and must be called from the directory that
+# created the target.
 #
 #   slice2cpp_generate(<target>
 #     [INCLUDE_DIRS <dir>...]     # extra -I directories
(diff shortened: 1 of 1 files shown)


pepone (approved):
Verified against the case matrix from the earlier rounds.

The guard fires on exactly the shape #6419 describes — target created in `lib/`, `slice2cpp_generate` called from the parent:

```
before: configures, then fails at build with
        clang++: error: no such file or directory: '.../generated/feature/Thing.cpp'
after:  CMake Error ... slice2cpp_generate: 'feature' was created in '.../src/lib';
        call slice2cpp_generate from that directory.
```

and does not fire where it should not. Both of these still build:

- a target created and generated inside the same `add_subdirectory`
- a target created by a `.cmake` file `include()`d from the directory that calls `slice2cpp_generate` — `include()` leaves `CMAKE_CURRENT_SOURCE_DIR` alone, so the comparison holds

The 21-case matrix is unchanged at 20/21 on Ninja, the one non-pass being the space-in-source-path case that is #6329.

## One ask, on the wording

Worth spending a couple more lines here, because the guard's whole value is replacing a baffling build-time failure with a message that teaches — and the user who hits this is not being careless. They wrote

```cmake
add_subdirectory(lib)
slice2cpp_generate(feature)
```

because that is how every `target_*` command behaves. From the `target_link_libraries` documentation:

> Added in version 3.13: The `<target>` doesn't have to be defined in the same directory as the `target_link_libraries` call.

`add_custom_command` never got that relaxation — "A target created in the same directory (CMakeLists.txt file) that specifies any output of the custom command as a source file is given a rule to generate the file" — so the user is applying a rule that holds everywhere else and does not hold here.

Against that expectation, a message that states a fact and issues an instruction reads as an arbitrary Ice restriction. The two guards just above it both name a condition the caller recognizes as their own mistake — "is an alias of", "is a `INTERFACE_LIBRARY`, which cannot compile Slice" — and this one could do the same. It also prints only the target's directory; the call site is left in CMake's `Call Stack` trailer for the reader to match up.

```suggestion
    message(FATAL_ERROR
      "slice2cpp_generate: '${target}' was created in '${target_source_dir}', but this call is in "
      "'${CMAKE_CURRENT_SOURCE_DIR}'. The rule that compiles the Slice files belongs to the directory "
      "that calls slice2cpp_generate, so nothing would ever generate the sources it adds to "
      "'${target}'. Move this call next to the add_library() or add_executable() that created "
      "'${target}'.")
```

which renders as:

```
  slice2cpp_generate: 'feature' was created in
  '<src>/lib',
  but this call is in
  '<src>'.
  The rule that compiles the Slice files belongs to the directory that calls
  slice2cpp_generate, so nothing would ever generate the sources it adds to
  'feature'.  Move this call next to the add_library() or add_executable()
  that created 'feature'.
```

Mismatch, reason, fix — and both directories on screen. `message()` concatenates its arguments, so the multi-line form is safe.

Approving; take the wording or leave it.

externl:
Took the suggested wording as-is.

PR #6430, opened 2026-08-05
Document that find_package(Ice) requires CONFIG mode

### LINKED ISSUE #6418: Document that find_package(Ice) needs CONFIG, or CMake's bundled FindIce module wins
CMake ships its own `FindIce.cmake` module, which targets Ice 3.7 and earlier: it locates Ice by
searching for the executables and libraries by name and derives the version by running
`slice2cpp --version`. It never delegates to the config package Ice installs.

Module mode is tried first by default, so a consumer writing

```cmake
find_package(Ice REQUIRED)
```

silently gets that module rather than `IceConfig.cmake`, and none of the imported targets, components or
`slice2cpp_generate` this package provides. The `CONFIG` (or `NO_MODULE`) keyword is required:

```cmake
find_package(Ice REQUIRED CONFIG)
```

`zeroc-ice/ice-demos` gets this right in `cpp/cmake/common.cmake`, but nothing states it, and the
failure is confusing: `find_package` appears to succeed, then `Ice::Ice` does not exist.

Worth saying explicitly wherever the CMake package is documented — and it also bounds the fix for
#6348: a versioned request only reaches the version file in CONFIG mode.

externl:
`find_package` uses module mode by default, which runs the `FindIce` module bundled with CMake and never reaches `IceConfig.cmake`, so none of this package's imported targets or `slice2cpp_generate` exist. The `CONFIG` keyword selects config mode, which loads this package. Closes #6418.

- Adds `cpp/config/README.md` documenting the CMake package in the same style as the other build tooling: usage and the `CONFIG` requirement, Windows setup from the NuGet package, imported targets and components, result variables, versioned requests, and `slice2cpp_generate`.
- `IceConfig.cmake`: header shows `find_package(Ice REQUIRED CONFIG)` and says why CONFIG is required.
- `slice2cpp.cmake`: the usage example starts with the `find_package(Ice REQUIRED CONFIG)` call, and the shared-directory example is renamed to `weather_api`/`Weather.ice` to match the README.

The README is not installed alongside the `.cmake` files: no CMake package on my machine ships documentation into `lib/cmake/<Pkg>/`, and our other tooling keeps its README next to the sources.


cpp/config/README.md
```diff
@@ -0,0 +1,106 @@
...
+| IceBox              | `Ice::IceBox`              |            |
+| IceDiscovery        | `Ice::IceDiscovery`        |            |
+| IceGrid             | `Ice::IceGrid`             |            |
+| IceLocatorDiscovery | `Ice::IceLocatorDiscovery` |            |
+| IceStorm            | `Ice::IceStorm`            |            |
+| IceBT               | `Ice::IceBT`               | Linux only |
+
+Components present in the installation are always available; listing them in
+`find_package(Ice CONFIG COMPONENTS ...)` additionally makes them required. The package also
+defines `Ice::slice2cpp`, the imported executable `slice2cpp_generate` runs.
```
bernardnormier:
```suggestion
Components present in the installation are always available; listing them in
`find_package(Ice CONFIG COMPONENTS ...)` additionally makes them required. The package also
defines `Ice::slice2cpp`, the imported executable `slice2cpp_generate` runs. `Ice::slice2cpp` is
not a component: it is always defined and cannot be listed in `COMPONENTS`.
```

cpp/config/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Result Variables
+
+| Variable                   | Description                                        |
+| -------------------------- | -------------------------------------------------- |
+| `Ice_VERSION`              | The Ice version, e.g. `3.9.0`                      |
+| `Ice_SO_VERSION`           | The shared library version, e.g. `39`              |
+| `Ice_PREFIX`               | The installation prefix                            |
+| `Ice_INCLUDE_DIR`          | The directory containing the Ice header files      |
+| `Ice_SLICE_DIR`            | The directory containing the Ice Slice files       |
```
bernardnormier: As written, a reader cannot tell whether they must add this directory to `INCLUDE_DIRS` themselves. `slice2cpp_generate` always passes it to slice2cpp, so worth saying so here, e.g. "The directory containing the Ice Slice files; always passed to slice2cpp by `slice2cpp_generate`".

cpp/config/README.md
```diff
@@ -0,0 +1,106 @@
...
+| `Ice_SLICE2CPP_EXECUTABLE` | Path to the `slice2cpp` compiler                   |
```
bernardnormier: Suggest annotating this one too: this variable is for configure-time use, where generator expressions are unavailable; build-time custom commands should prefer `$<TARGET_FILE:Ice::slice2cpp>`, which also adds the dependency on the compiler.

bernardnormier (commented):
A few points, mostly on the README (whose content checks out against the package sources):

- Please reflow README.md to 120 columns.
- The description says "Maybe we'll be able to find an even better spot in the future" — what do you have in mind? The docs site, or a more discoverable place in the repo?
- Is there a reason not to install this README together with the other files? The install rule picks up `*.cmake` only, so as-is the README is visible only when browsing the repo. Installing it would put the usage docs right where a user of an installed Ice would look — next to `IceConfig.cmake`.

The Compiling Slice Files section keeps an important piece of information — where the files are generated — in the middle of a dense paragraph, and I found the layout sentence ("Generated files mirror the layout of the `.ice` files under the `INCLUDE_DIRS` directory that reaches them") hard to follow. Here is a proposed rework: it states the output locations up front, explains the layout starting from the `#include` directives it exists to serve, spells out what `INCLUDE_SCOPE` does, and renames the example target and Slice file (`demo` reads oddly next to the `Demo/` prefix, and a greeter is too obviously trivial to need a shared Slice directory):

````markdown
## Compiling Slice Files

`slice2cpp_generate` compiles the `.ice` files already present in a target's sources and adds the generated C++ to the
target. Call it from the directory that created the target, after all of the target's `.ice` sources have been added.

```cmake
slice2cpp_generate(<target>
  [INCLUDE_DIRS <dir>...]     # extra -I directories; also drive the generated file layout (see below)
  [OPTIONS <option>...]       # extra slice2cpp options, e.g. -DFOO or --header-ext hpp
  [HEADER_OUTPUT_DIR <dir>]   # put the headers here instead of with the sources
  [INCLUDE_DIR <dir>]         # slice2cpp --include-dir
  [INCLUDE_SCOPE <scope>]     # PRIVATE (default) or PUBLIC
)
```

`slice2cpp_generate` always passes `Ice_SLICE_DIR` to slice2cpp, before the `INCLUDE_DIRS` directories, so
`#include <Ice/...>` directives resolve without any configuration.

Generated files are placed under `<current binary dir>/generated/<target>`; with `HEADER_OUTPUT_DIR`, the headers are
placed under `HEADER_OUTPUT_DIR[/INCLUDE_DIR]` instead. When a Slice file includes another, e.g.
`#include <Demo/Weather.ice>`, the generated code includes the corresponding generated header by the same path,
`Demo/Weather.h`. To make these directives resolve, each generated file is placed at its `.ice` file's path relative to
the `INCLUDE_DIRS` directory containing it: with `INCLUDE_DIRS ../slice`, the file `../slice/Demo/Weather.ice`
generates `Demo/Weather.h` under the header output directory.

The directories containing the generated headers are added to the target's include directories with the scope given by
`INCLUDE_SCOPE`. With `PRIVATE` (the default), only this target's own sources can include the generated headers. With
`PUBLIC`, the directories become part of the target's usage requirements: when another target depends on this one
through `target_link_libraries`, they go on that target's include path as well, so its sources can include the
generated headers. `PUBLIC` applies within the build tree only: these directories are omitted from the target's
installed interface, and installing the generated headers is the caller's responsibility.

For example, the following defines a library whose Slice files include files from a shared directory, and publishes
its generated headers to its own consumers under a `Demo/` prefix:

```cmake
add_library(weather_api Weather.ice)
slice2cpp_generate(weather_api
  INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/../slice
  OPTIONS -DENABLE_EXTRAS
  HEADER_OUTPUT_DIR ${CMAKE_BINARY_DIR}/include
  INCLUDE_DIR Demo
  INCLUDE_SCOPE PUBLIC)
```
````

If you take the rename, the matching example in the slice2cpp.cmake header comment should follow suit to stay in sync.

**Review by pepone — COMMENTED** [review]
The `CONFIG` point is right and worth documenting. I staged an install tree and ran both modes
against CMake 4.4.2; Bernard has the README structure covered, so just what I found on top.

**`README.md:3-5` writes the call the page tells you not to make** — "`find_package(Ice)` provides
imported targets…". Should be `find_package(Ice CONFIG)`.

**"a plain `find_package(Ice)`" undersells it, and the symptoms aren't named.** "Plain" reads as
*the bare call is the trap* — a reviewer here took it exactly that way. In fact every call without
`CONFIG`/`NO_MODULE` runs CMake's module, whatever else you pass, and the two failures look nothing
alike: without `COMPONENTS` it hard-fails, with `COMPONENTS` it succeeds and defines a working
`Ice::Ice`. The second is the form this README's own Imported Targets section teaches, so dropping
`CONFIG` from it lands there. Suggested block:

> [!IMPORTANT]
> The `CONFIG` (or `NO_MODULE`) keyword is required. `find_package` has two modes: module mode,
> which runs a `FindIce` module — and CMake bundles one of its own, written for Ice 3.7 — and
> config mode, which loads the `IceConfig.cmake` installed by this package. Module mode is tried
> first, and CMake falls back to config mode only when no `FindIce` module exists anywhere; its
> own always does. So *every* call without `CONFIG` runs CMake's module, whatever else you pass —
> `REQUIRED`, `COMPONENTS`, a version. None of the targets or functions documented below come
> from it.
>
> It shows up two ways:
>
> - `find_package(Ice REQUIRED)` fails with `Failed to find all Ice components`, while reporting
>   your real Ice version — CMake's module cannot succeed without `COMPONENTS`.
> - `find_package(Ice REQUIRED COMPONENTS Ice)` succeeds and defines its own `Ice::Ice`, then the
>   build stops at `Unknown CMake command "slice2cpp_generate"`.

Quoting both error strings verbatim also makes the page findable by the error. (Mechanism, for
the record: `FindIce` lists `Ice_LIBRARY` in `REQUIRED_VARS` and only ever fills it from
`Ice_FIND_COMPONENTS` — `FindIce.cmake:717,762,818-825` — so with no components it always fails,
however correct the installation. With components it creates `Ice::Ice` at
`FindIce.cmake:843-853`, minus `Ice::slice2cpp`, `Ice::DataStorm`, `cxx_std_17` and
`Threads::Threads`.)

**Slice files reached through `INCLUDE_DIRS` are not compiled** (`slice2cpp.cmake:30-32`) — they
need their own target, which you then link. The example is built around a shared directory and
doesn't say so, so `INCLUDE_DIRS ../slice` generates `Demo/Weather.h` with nothing defining its
symbols. Not in Bernard's rewrite either; worth a sentence in whichever lands.

Smaller:

- `Ice_VERSION` includes the pre-release suffix by design — `3.9.0-alpha.0` / `39a0` today
  (`IceVersion.cmake:8-9`). Worth saying, for anyone doing `VERSION_GREATER`.
- `Ice::icebox_EXE` is the other Windows-only imported target (`IceTargets.cmake:78`).
- Versioned requests aren't mentioned. `IceConfigVersion.cmake` gives `SameMinorVersion` plus
  ranges; module mode only does a plain minimum compare against `slice2cpp --version`.

Longer term this is Kitware's to fix, and there's precedent: CMP0167 removed `FindBoost` in CMake
3.30 precisely because upstream ships `BoostConfig.cmake`. CMake's `FindIce` still targets 3.7 —
version list stops at `3.7.0`, still hunts for Freeze and IcePatch2 — and has no deprecation note
or config-mode delegation, though `FindTIFF`/`FindGLEW` show the `find_package(… CONFIG QUIET)`
delegation pattern in-tree. Worth an upstream issue; the README stays useful either way, since a
policy only helps projects that raise `cmake_minimum_required`.

CI: all twelve red checks are the 2026-08-06 Actions outage (`Failed to resolve action download
info. Error: Service Unavailable`), before any repo step ran. Just needs a re-run.

**Comment by externl** [comment]
Pushed the review changes.

**On installing the README next to the .cmake files — I'd rather not**, and the convention is fairly clear:

- Of the 15 CMake packages installed on this machine, all 54 shipped files are `.cmake`. Not one ships a README, `.md`, `.rst` or `.txt` into `lib/cmake/<Pkg>/`.
- Those config files aren't self-documenting either — fmt is 4 comment lines out of 31, harfbuzz 2 of 62, nghttp3 0 of 3. They're generated artifacts.
- Docs live in the project's repo or docs site instead. protobuf documents `protobuf_generate` in `docs/cmake_protobuf_generate.md`, which is not installed with the package.
- Our own tooling already works this way: the MSBuild, Gradle and SwiftPM READMEs sit next to their sources. The one that ships is `ZeroC.Ice.Slice.Tools.Cpp`, and it ships to the NuGet package root, where the registry renders it as the landing page — not next to installed files on disk.
- `CMakePackageConfigHelpers` says nothing about installing docs.

**Everything else taken.** Reflowed to 120 columns; took the `Compiling Slice Files` rework and the `weather_api`/`Weather.ice` rename, with the `slice2cpp.cmake` example renamed to match; all three inline suggestions; added `Ice::icebox_EXE`, the pre-release suffix note, the INCLUDE_DIRS-files-are-not-compiled sentence, and versioned requests.

**One deliberate departure**, @pepone: I cut the error strings. You're right that every call without `CONFIG` runs the module whatever else is passed, and that's now stated plainly — but the symptoms turned out to be layout-dependent rather than a fixed pair. I could reproduce both: against my staged install it fails on `Ice_SLICE_DIR`, and once that resolves it does succeed and then dies at `slice2cpp_generate`, exactly as you described. Documenting which one you get means documenting the bundled module's search behaviour, which isn't ours to describe or keep current. The README now says what our package needs and stops there.

CI reds are the 2026-08-06 Actions outage, before any repo step ran — needs a re-run.

File: cpp/config/IceConfig.cmake [hunk 4]
```diff
@@ -1,6 +1,12 @@
 # Copyright (c) ZeroC, Inc.
 
 # CMake package configuration file for Ice.
+#
+#   find_package(Ice REQUIRED CONFIG)
+#
+# CONFIG (or NO_MODULE) is required: it selects config mode, which loads this file. Without it
+# find_package uses module mode, which runs the FindIce module bundled with CMake and doesn't
+# load this file.
```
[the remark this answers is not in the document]
externl: I'd rather not have these kinds of references as they're easy to break.

bernardnormier:
> **On installing the README next to the .cmake files — I'd rather not**, and the convention is fairly clear:

The real issue here is how do user find this README.md?

bernardnormier (approved):
Looks good.

One optional placement nit: the versioned-requests paragraph ("A versioned request is answered by the installed version...") sits inside the **Windows** subsection, but it's platform-neutral — a reader who skips the Windows section will miss it. Consider moving it to the end of Usage (before the Windows subsection) or into Result Variables next to the `Ice_VERSION` discussion.

externl:
Moved the versioned-requests paragraph to the end of Usage, above the Windows subsection — it had ended up trailing that section when I added it.

On discoverability: nothing in the repo links to it yet, which is the gap you're pointing at. The nearest fix is a link from `cpp/README.md`, where a C++ user already goes, and the docs site beyond that. Happy to add the link here or leave it for a follow-up — say which you prefer.

externl:
Regarding where to put the docs. As with most things C++ is a mess and there's no one standard from what I can tell. The two most common are GitHub (source) and docs. 

- gRPC has theirs in [GitHub](https://github.com/grpc/grpc/tree/master/src/cpp#to-start-using-grpc-c), that you find  by reading their [Quick start](https://grpc.io/docs/languages/cpp/quickstart/) page (see if you can spot the link).
- QT uses their [docs](https://doc.qt.io/qt-6/cmake-get-started.html).
- Catch2  also uses [GitHub](https://github.com/catchorg/Catch2/blob/devel/docs/cmake-integration.md).

I lean towards keeping it in GitHub with links from the website and docs.

pepone (approved):
Agreed on cutting the error strings — documenting the bundled module's search behaviour isn't ours to keep current.

**The Windows copy example needs CMake 3.26, but the package requires 3.21** (`README.md:61` vs `:11`). `cmake -E copy -t` is `versionadded:: 3.26`; on 3.21–3.25 `-t` is taken as a source file and the build fails. The destination-last form works on 3.21 and the combined list is never empty, since `ICE_RUNTIME_DLLS` always carries bzip2:

```cmake
COMMAND ${CMAKE_COMMAND} -E copy
  $<TARGET_RUNTIME_DLLS:client>
  $<GENEX_EVAL:$<TARGET_PROPERTY:Ice::Ice,ICE_RUNTIME_DLLS>>
  $<TARGET_FILE_DIR:client>
```

**`Ice_WIN32_PLATFORM` doesn't default to `x64`** (`README.md:52-53`) — it follows the build architecture, `Win32` for a 32-bit build (`IceTargets.cmake:5-11`). Worth saying it auto-matches, so a Win32 build doesn't look like it needs setting.

**`Ice::icebox_EXE` is not always defined** (`README.md:88-90`) — "Both are always defined" contradicts the "on Windows" a line above (`IceTargets.cmake:47-79`).

**`$<TARGET_FILE:Ice::slice2cpp>` in `COMMAND` adds no dependency** (`README.md:112-113`). It's an imported target, so there is no target-level dependency to add, and a target-level one wouldn't re-run the command anyway — "List target names with the `DEPENDS` option to add such file-level dependencies". That's what the package itself does at `slice2cpp.cmake:323`, so the sentence should point at `DEPENDS`.

On discoverability: I'd add the `cpp/README.md` link in this PR rather than a follow-up — it's one line in the existing link list at the bottom, and without it the page ships unreachable.

externl:
All five taken.

- Copy example is destination-last; verified `-t` is `versionadded 3.26` against the package's 3.21 floor. The same `-t` snippet is in ice-demos, which also requires 3.21 — I'll fix that there separately.
- `Ice_WIN32_PLATFORM` now says it defaults to the build architecture and only needs setting to cross the two.
- "Both are always defined" is gone; neither executable can be listed in `COMPONENTS`.
- The `Ice_SLICE2CPP_EXECUTABLE` note points at `DEPENDS` for the re-run dependency instead of claiming `$<TARGET_FILE>` provides it.
- `cpp/README.md` links the package README from its link row.

PR #6432, opened 2026-08-05
Fix two IceRuby marshaling issues

### LINKED ISSUE #6369: Ruby: ExceptionReader::_read never calls InputStream::endException
> [!NOTE]
> Raised by an AI cross-review of #6365 and verified against source. It is independent of that PR — it is present on
> `main` both before and after it.

## Summary

`IceRuby::ExceptionReader::_read` calls `InputStream::startException` but never the paired `endException`
(`ruby/src/IceRuby/Types.cpp:2774-2780`):

```cpp
void
IceRuby::ExceptionReader::_read(Ice::InputStream* is)
{
    is->startException();

    const_cast<VALUE&>(_ex) = _info->unmarshal(is);
    rb_gc_register_address(&_ex);
}
```

Every sibling mapping pairs them, and Ruby itself did until the 3.8 `SliceLoader` refactor:

- IcePy: `python/modules/IcePy/Types.cpp:3316-3325`
- IcePHP: `php/src/Types.cpp:3246-3253`
- Ice 3.7 Ruby: `ruby/src/IceRuby/Types.cpp:2934-2942`, which called `is->endException(_info->preserve)` and stored the
  resulting `SlicedData`.

`endException()` forwards to `EncapsDecoder::endInstance()` (`cpp/include/Ice/InputStream.h:183-188`). For the 1.1
encoding that is `readSlicedData()`, clearing `indirectionTables` and `slices`, and popping `_current`
(`cpp/src/Ice/InputStream.cpp:1865-1872`); for the 1.0 encoding it resets `_sliceType` to `NoSlice`
(`cpp/src/Ice/InputStream.cpp:1575-1592`). Skipping it leaves the decoder mid-instance.

## Impact

None that is user-visible today, which is why this is a consistency fix rather than a bug report:

- The `InputStream` in `IceRuby::TypedInvocation::unmarshalException` is a local, single-use stream
  (`ruby/src/IceRuby/Operation.cpp:539-560`), so the unpopped `_current` and the retained slices die with the decoder
  immediately afterwards.
- `EncapsDecoder11::throwException` does not call `readPendingValues()` after `ex._read()`
  (`cpp/src/Ice/InputStream.cpp:1830-1834`), so nothing observes the unpopped frame.
- On the 1.0 path `readPendingValues()` is called (`cpp/src/Ice/InputStream.cpp:1536-1541`), but `readInstance()`
  assigns `_sliceType = ValueSlice` unconditionally (`cpp/src/Ice/InputStream.cpp:1678`), so the stale
  `ExceptionSlice` is overwritten before the `startInstance` assert sees it.

The cost of leaving it is that Ruby is the one mapping using the `startException`/`endException` API incorrectly, and
that any future change relying on balanced decoder state would break here first.

**Files:** `ruby/src/IceRuby/Types.cpp:2774-2780`

## Fix

Call `is->endException()` after the unmarshal, matching IcePy and IcePHP. Ruby's `ExceptionReader` has no
`_slicedData` member any more, so the returned `SlicedDataPtr` is simply discarded.

No changelog entry — there is no user-visible change.

### LINKED ISSUE #6370: Ruby: ValueWriter registers a GC root before running code that can throw
> [!NOTE]
> Raised by an AI cross-review of #6365 and verified against source. It is independent of that PR — it is present on
> `main` both before and after it.

## Summary

`IceRuby::ValueWriter`'s constructor registers a GC root and then runs code that can throw
(`ruby/src/IceRuby/Types.cpp:2354-2371`):

```cpp
IceRuby::ValueWriter::ValueWriter(VALUE object, ValueMap* valueMap, const ClassInfoPtr& formal)
    : _object(object), _map(valueMap), _formal(formal)
{
    rb_gc_register_address(&_object);
    if (!_formal || !_formal->interface)
    {
        volatile VALUE cls = CLASS_OF(object);
        volatile VALUE type = callRuby(rb_const_get, cls, rb_intern("ICE_TYPE"));
        ...
    }
}
```

`callRuby` exists precisely to turn a Ruby error into a C++ `RubyException` (`ruby/src/IceRuby/Util.h:143-148`), so
`rb_const_get` raising `NameError` on a class with no `ICE_TYPE` constant throws out of the constructor body. The
object is never fully constructed, so `~ValueWriter` (`ruby/src/IceRuby/Types.cpp:2373`) never runs and
`rb_gc_unregister_address` is never called. The registration survives, pointing at the freed `ValueWriter` storage —
the same defect class #6365 fixes in `ExceptionReader`, where the GC then reads freed memory on every collection and
`rb_gc_unregister_address`'s linear scan of the global list gets slower for everyone.

## Reachability

Narrow. Operation arguments are type-checked before marshaling (`ruby/src/IceRuby/Operation.cpp:412`), and
`ClassInfo::validate` (`ruby/src/IceRuby/Types.cpp:1940-1952`) already establishes that `ICE_TYPE` exists; nested
class members are validated the same way by `StructInfo` (`:970`), `SequenceInfo` (`:1178`), `DictionaryInfo`
(`:1737`) and `ValueWriter::writeMembers` (`:2442`).

The one unvalidated construction site is the preserved-slices path, which takes objects straight out of the Ruby
`SlicedData`'s `@instances` array with a null formal type (`ruby/src/IceRuby/Types.cpp:345-357`):

```cpp
VALUE o = RARRAY_AREF(instances, j);
...
writer = make_shared<ValueWriter>(o, valueMap, nullptr);
```

`_formal` is null there, so the `ICE_TYPE` lookup always runs. Under a normal unmarshal/re-marshal round trip that
array holds Slice class instances, so reaching the throw takes an application that builds or edits a `SlicedData` by
hand. That makes this latent rather than field-reachable — but it is a one-line fix.

**Files:** `ruby/src/IceRuby/Types.cpp:2354-2371`, `ruby/src/IceRuby/Types.cpp:345-357`

## Fix

Register last, so the constructor cannot throw after taking the root:

```cpp
    if (!_formal || !_formal->interface)
    {
        ...
    }
    rb_gc_register_address(&_object);
```

That matches the two classes in the extension that already get this right — `ValueReader`
(`ruby/src/IceRuby/Types.cpp:2458-2464`) and `RubySliceLoader` (`ruby/src/IceRuby/RubySliceLoader.cpp:10-16`), where
the registration is the constructor's final action.

No changelog entry — the trigger requires a hand-built `SlicedData`, and there are no field reports.

bernardnormier:
## Summary

Two small IceRuby fixes from the cross-review of the earlier ExceptionReader GC fix:

- `ExceptionReader::_read` now calls `InputStream::endException()` after unmarshaling the exception, pairing the `startException()` call like IcePy, IcePHP, and Ruby itself prior to the 3.8 SliceLoader refactor. Without it the decoder is left mid-instance; nothing observes that state today, so this is a consistency fix with no user-visible change.
- `ValueWriter`'s constructor now registers its GC root as the last statement. The `ICE_TYPE` lookup runs through `callRuby` and can throw `RubyException` (reachable with a null formal type via a hand-built `SlicedData`); throwing after `rb_gc_register_address` would skip the destructor and leak a registration pointing at freed storage. This matches `ValueReader` and `RubySliceLoader`, which already register last.

No changelog entries: neither change affects valid programs during normal operations.

Fixes #6369.
Fixes #6370.

### DIFF
--- a/ruby/src/IceRuby/Types.cpp
+++ b/ruby/src/IceRuby/Types.cpp
@@ -2436,10 +2436,6 @@ IceRuby::ValueWriter::ValueWriter(VALUE object, ValueMap* valueMap, const ClassI
       _map(valueMap),
       _formal(formal)
 {
-    //
-    // Mark the object as in use for the lifetime of this wrapper.
-    //
-    rb_gc_register_address(&_object);
     if (!_formal || !_formal->interface)
     {
         volatile VALUE cls = CLASS_OF(object);
@@ -2448,6 +2444,9 @@ IceRuby::ValueWriter::ValueWriter(VALUE object, ValueMap* valueMap, const ClassI
         _info = dynamic_pointer_cast<ClassInfo>(getType(type));
         assert(_info);
     }
+
+    // Mark the object as in use for the lifetime of this wrapper. We register last, in case the code above throws.
+    rb_gc_register_address(&_object);
 }
 
 IceRuby::ValueWriter::~ValueWriter() { rb_gc_unregister_address(&_object); }
@@ -2861,6 +2860,7 @@ IceRuby::ExceptionReader::_read(Ice::InputStream* is)
 {
     is->startException();
     _ex = _info->unmarshal(is);
+    is->endException();
 }
 
 bool


pepone (approved):
Both changes look right. Verified locally rather than by inspection alone.

**`endException()`** — Ruby was the only mapping not pairing it: `Ice::UserException::_read` (`cpp/src/Ice/UserException.cpp:47-52`), IcePy (`python/modules/IcePy/Types.cpp:3320-3325`) and IcePHP (`php/src/Types.cpp:3248-3253`) all do. I checked the "no user-visible change" claim against both decoders:

- 1.1: `skipSlice` does not preserve exception slices (`cpp/src/Ice/InputStream.cpp:2023-2025`), so `readSlicedData()` short-circuits to `nullptr` and the only effect is popping `_current` — which also makes `throwException`'s `assert(!_current)` hold on a reused stream.
- 1.0: `_sliceType` goes `ExceptionSlice` -> `NoSlice`, and on the `usesClasses` path `readInstance()` assigns `_sliceType = ValueSlice` unconditionally (`InputStream.cpp:1678`), so nothing downstream sees the difference.

**`ValueWriter` registering last** — no GC window is opened by the reorder: `_object` is initialized in the member-init list, and the `object` parameter keeps the value reachable and pinned through the conservative machine-stack scan for the whole body. The throw also now precedes `valueMap->insert`, so nothing half-registered survives.

Reachability confirmed: `getSlicedDataMember` passes a null formal (`ruby/src/IceRuby/Types.cpp:392`), so the `ICE_TYPE` lookup always runs, and `Ice::SlicedData`/`Ice::SliceInfo` are plain Ruby classes with only `attr_reader`s — an app can build one with `allocate` + `instance_variable_set`.

Probed it with a Ruby client against the C++ `slicing/objects` server, planting a plain `Object.new` in a hand-built `SlicedData` and calling `checkPBSUnknown`, 30k times on both the PR build and base `a4622f63e1`: both raise a clean `NameError: uninitialized constant ICE_TYPE` and survive `GC.verify_compaction_references` + `GC.start`; the pre-fix build just retains ~500 KB more RSS. So the leak is real but silent — which matches the "latent, not field-reachable" framing. Agree with no test and no changelog entry for either.

Sweep: after this PR all six `rb_gc_register_address` sites in `ruby/src` register as the constructor's last action, and every stream `start*`/`end*` pair in the extension is balanced.

Full Ruby suite is green on the PR build except `Ice/optional`, which fails identically with `ruby/src` reverted to base — stale prebuilt C++ test server on my machine, not this PR.

Issue #6434, opened 2026-08-05
Remove the unused batch support from the echo test servers


bernardnormier:
The echo test servers (`cpp/test/Ice/echo`, `csharp/test/Ice/echo`, and the Java equivalent) implement batch forwarding — `startBatch`/`flushBatch` operations backed by a `_batchProxy`/`_startBatch` state machine in each `BlobjectI` — and the JS tests' `Echo` Slice interface declares the same operations (for example `js/packages/test/Ice/facets/Test.ice`).

However, no test in any mapping ever calls `startBatch`: the only occurrences in the repository are the Slice declarations, the generated code, and the three server implementations. The batch-forwarding machinery is unexercised in all mappings.

Noticed while fixing item 4 of #6137 (the C# `startBatch` had an inverted assertion, which could never fire precisely because the method has no callers).

Suggested cleanup: remove `startBatch`/`flushBatch` from the echo `Test.ice` files and the `BlobjectI` implementations in the three server mappings, and from the JS test Slice — or, if batch coverage through the echo server is considered valuable, add a JS test that actually exercises it.

PR #6435, opened 2026-08-05
Fix weakened, inverted, and unreachable test assertions

### LINKED ISSUE #6137: Test-suite defect batch: weakened/inverted assertions (3 mask real regressions: SSL hostname, Timer guards, stream bounds)
> [!NOTE]

## Summary

Eight test-suite defects — weakened, inverted, or unreachable assertions — across C++, C#, Java, and Swift. **Items 1–3 mask real product regressions:** a regression in IceSSL hostname verification, in the `Timer` schedule guards, or in the empty-stream bounds check would still let these tests report ok. Items 4–8 reduce coverage or are latent helper bugs.

All line numbers are on `main` at `1579b6928e`.

### 1. IceSSL hostname-verification bypass is not asserted (C++)

`cpp/test/IceSSL/configuration/AllTests.cpp:597-604` — the `ca1/server_cn2` case (target host matches the certificate CN, DNS altName does **not** match, `IceSSL.CheckCertName=1`) pings and catches `Ice::SecurityException` with no `test(false)` after the ping:

```cpp
try
{
    server->ice_ping();     // <-- no test(false) here
}
catch (const Ice::SecurityException&)
{
    // Expected.
}
```

If `CheckCertName` stopped rejecting the altName mismatch, `ice_ping()` would succeed and the test would still pass. The identical omission is in the `ca1/server_cn3` `__APPLE__` branch (`:620-627`). The `ca1/server_cn4` case (`:645-652`) has the correct shape — `ice_ping(); test(false);` inside the `try`.

### 2. `Timer` schedule guards are not asserted, and one guard is unreachable (C++)

`cpp/test/IceUtil/timer/Client.cpp` — two schedule-rejection checks catch the expected `std::invalid_argument` without a `test(false)` after the call, so a `Timer` that stopped rejecting would pass:

- **`:176-184`** — "the same task cannot be scheduled more than once" (`cpp/src/Ice/Timer.h:86`, `"task is already scheduled"`). This check also races: the task is first scheduled with a 100 ms delay, so if the one-shot fires before the second `schedule` call the task is no longer scheduled and the duplicate is legitimately accepted. Use a long delay plus a `cancel` so the window cannot close.
- **`:266-279`** — schedule after `timer->destroy()` (`cpp/src/Ice/Timer.h:68`, `"timer destroyed"`).

The third block, **`:251-265`**, has a **dead catch**. `DestroyTask::runTimerTask` calls `_timer->destroy()` from the timer thread, which throws `std::runtime_error("a timer task cannot destroy the timer")` (`cpp/src/Ice/Timer.cpp:16-21`) — so the timer is *not* destroyed. The one-shot task was already erased when it ran, so the following re-`schedule` succeeds, `invalid_argument` is never thrown, and the block asserts nothing at all.

### 3. Empty-stream bounds check is swallowed (Swift)

`swift/test/Ice/stream/Client.swift:46-51`:

```swift
do {
    inS = Ice.InputStream(communicator: communicator, bytes: Data())
    do {
        _ = try inS.read() as Bool
        try test(false)
    } catch {}          // <-- also catches the TestFailed thrown by test(false)
}
```

The bare `catch {}` catches everything, including the `TestFailed` error that `try test(false)` throws — so an unexpectedly **successful** read from an empty stream is indistinguishable from the expected `MarshalException`. Narrow it to `catch is Ice.MarshalException {}`, as the C++ counterpart does.

### 4. C# echo `startBatch` asserts the wrong direction

`csharp/test/Ice/echo/BlobjectI.cs:11` asserts `Debug.Assert(_batchProxy != null)`, but the C++ and Java siblings assert the opposite — `assert(!_batchProxy)` (`cpp/test/Ice/echo/BlobjectI.cpp:21`) and `assert (_batchProxy == null)` (`java/test/src/main/java/test/Ice/echo/BlobjectI.java:16`). `startBatch` runs *before* any batch proxy exists, so the C# direction is inverted. Dormant: nothing calls `startBatch` in the C# test, and `flushBatch`'s `!= null` assertion is correct.

### 5. C#/Java property test never reaches the non-numeric parse path

`csharp/test/Ice/properties/Client.cs:212-224` and its Java sibling `java/test/src/main/java/test/Ice/properties/Client.java:205-214` set the dotless property `Foo` to `"bar"` and then call `getIcePropertyAsInt("Foo")`. That accessor rejects `Foo` as an unknown **Ice** property before it ever parses `"bar"`, so the `PropertyException` the test catches is the wrong one and "trying to read a non-numeric value as an int throws" is never actually exercised. C++ uses the general accessor and does reach the parse path (`cpp/test/Ice/properties/Client.cpp:228-239`, `getPropertyAsInt`). The adjacent out-of-range test is unaffected — it uses `getPropertyAsIntWithDefault`, so int parsing is still covered there.

### 6. C# `Custom<T>` equality is asymmetric and inconsistent with its hash

`csharp/test/Ice/seqMapping/Custom.cs:22-57` — `Equals` walks `_list` against the other sequence's enumerator but never checks that the other enumerator is exhausted at the end, so a sequence equals any longer sequence that starts with it (`[1,2].Equals([1,2,3])` is `true`, the reverse is `false`). `GetHashCode()` returns `base.GetHashCode()` (reference identity) while `Equals` is by value, which is inconsistent for any hashed container.

### 7. Swift AMI local-exception check asserts nothing

`swift/test/Ice/ami/AllTests.swift:60-64` — an invocation on `p.ice_adapterId("dummy")` that is expected to throw `Ice.NoEndpointException` has no `try test(false)` after it:

```swift
do {
    let indirect = p.ice_adapterId("dummy")
    try await indirect.op()     // <-- no try test(false) here
} catch is Ice.NoEndpointException {}
```

The oneway `opWithResult` check immediately below does it correctly.

### 8. Swift `TestI.sleep` ignores its argument

`swift/test/Ice/ami/TestI.swift:70-72` — `func sleep(ms: Int32)` always sleeps 100 ms and never reads `ms`. Every other language mapping honours the parameter. No observable difference today (the only caller passes 100), but any new test using a different delay would silently get 100 ms.

## Fix

- **Items 1, 2, 7:** add the missing `test(false)` / `try test(false)` immediately after each call that is expected to throw.
- **Item 2 (also):** give the duplicate-schedule check a long delay plus a `cancel` so it cannot race, and make the self-destroy block (`:251-265`) assert something real — or drop it.
- **Item 3:** narrow the catch to `Ice.MarshalException`.
- **Item 4:** flip the assertion to `_batchProxy == null`.
- **Item 5:** use the general property accessor so the parse path is reached.
- **Item 6:** check `!e.MoveNext()` after the loop and give `Custom<T>` a value-based `GetHashCode`.
- **Item 8:** honour `ms`.

**Severity:** Low (test-only — no shipped code is affected). Items 1–3 are the priority: they can hide an IceSSL hostname-verification, `Timer` schedule-guard, or stream bounds-check regression respectively.

bernardnormier:
## Summary

Fixes the eight test-suite defects from the audit batch — weakened, inverted, or unreachable assertions across C++, C#, Java, and Swift. The first three could have masked real product regressions; their strengthened assertions now re-verify the guards they protect (all suites pass, so none has regressed).

- **IceSSL configuration (C++)**: the `ca1/server_cn2` and `ca1/server_cn3` (`__APPLE__`) hostname-verification cases now assert that `ice_ping` fails; previously a `CheckCertName` regression would have passed silently. `ca1/server_cn4` already had the correct shape.
- **IceUtil timer (C++)**: the duplicate-schedule and schedule-after-destroy checks now assert the rejection (`test(false)` after the call). The duplicate-schedule check is also race-free: the task is scheduled with a long delay and cancelled afterwards, so it cannot fire before the duplicate `schedule` call. The self-destroy block's follow-up was dead (the re-schedule legitimately succeeds since the task already ran and the timer survived the in-task `destroy`); it now asserts something real — the timer remains usable after the rejected destroy. The destroy-from-task guard itself was already asserted inside `DestroyTask`.
- **stream (Swift)**: the empty-stream read check caught everything, including the `TestFailed` thrown by `try test(false)`; it now catches only `Ice.MarshalException`, like the C++ counterpart.
- **ami (Swift)**: the local-exception check asserts that the invocation on an unknown adapter ID throws, and `TestI.sleep` honors its `ms` argument instead of always sleeping 100 ms.
- **echo (C#)**: `startBatch` asserted `_batchProxy != null`; the C++ and Java siblings assert the opposite, and `startBatch` runs before any batch proxy exists. Flipped to `== null`. (The method currently has no callers in any mapping — the unused echo batch support is tracked separately in #6434.)
- **properties (C#, Java)**: the non-numeric-value test called `getIcePropertyAsInt("Foo")`, which rejects `Foo` as an unknown Ice property before parsing anything; it now uses `getPropertyAsInt` like C++, so the parse path is actually exercised.
- **seqMapping (C#)**: `Custom<T>.Equals` never checked that the other enumerator was exhausted, so a sequence equaled any longer sequence with a matching prefix; and `GetHashCode` was reference-based while `Equals` is by value. Both fixed.

Test-only changes; no changelog entry.

Fixes #6137.

### DIFF
--- a/csharp/test/Ice/seqMapping/Custom.cs
+++ b/csharp/test/Ice/seqMapping/Custom.cs
@@ -19,40 +19,7 @@ public class Custom<T> : IEnumerable<T>
 
     public void Add(T element) => _list.Add(element);
 
-    public override bool Equals(object obj)
-    {
-        try
-        {
-            var tmp = (Custom<T>)obj;
-            IEnumerator<T> e = tmp.GetEnumerator();
-            foreach (T element in _list)
-            {
-                if (!e.MoveNext())
-                {
-                    return false;
-                }
-                if (element == null)
-                {
-                    if (e.Current != null)
-                    {
-                        return false;
-                    }
-                }
-                else
-                {
-                    if (!element.Equals(e.Current))
-                    {
-                        return false;
-                    }
-                }
-            }
-            return !e.MoveNext();
-        }
-        catch (Exception)
-        {
-            return false;
-        }
-    }
+    public override bool Equals(object obj) => obj is Custom<T> other && _list.SequenceEqual(other._list);
 
     public override int GetHashCode()
     {
--- a/csharp/test/Ice/seqMapping/Custom.cs
+++ b/csharp/test/Ice/seqMapping/Custom.cs
@@ -17,7 +17,7 @@ public class Custom<T> : IEnumerable<T>
         set => _list[index] = value;
     }
 
-    public void Add(T elmt) => _list.Add(elmt);
+    public void Add(T element) => _list.Add(element);
 
     public override bool Equals(object obj)
     {
@@ -25,13 +25,13 @@ public class Custom<T> : IEnumerable<T>
         {
             var tmp = (Custom<T>)obj;
             IEnumerator<T> e = tmp.GetEnumerator();
-            foreach (T elmt in _list)
+            foreach (T element in _list)
             {
                 if (!e.MoveNext())
                 {
                     return false;
                 }
-                if (elmt == null)
+                if (element == null)
                 {
                     if (e.Current != null)
                     {
@@ -40,13 +40,13 @@ public class Custom<T> : IEnumerable<T>
                 }
                 else
                 {
-                    if (!elmt.Equals(e.Current))
+                    if (!element.Equals(e.Current))
                     {
                         return false;
                     }
                 }
             }
-            return true;
+            return !e.MoveNext();
         }
         catch (Exception)
         {
@@ -54,7 +54,15 @@ public class Custom<T> : IEnumerable<T>
         }
     }
 
-    public override int GetHashCode() => base.GetHashCode();
+    public override int GetHashCode()
+    {
+        var hash = new HashCode();
+        foreach (T element in _list)
+        {
+            hash.Add(element);
+        }
+        return hash.ToHashCode();
+    }
 
     private readonly List<T> _list = new List<T>();
 }
--- a/cpp/test/IceSSL/configuration/AllTests.cpp
+++ b/cpp/test/IceSSL/configuration/AllTests.cpp
@@ -597,6 +597,7 @@ testCertificateVerification(
         try
         {
             server->ice_ping();
+            test(false);
         }
         catch (const Ice::SecurityException&)
         {
@@ -620,6 +621,7 @@ testCertificateVerification(
         try
         {
             server->ice_ping();
+            test(false);
         }
         catch (const Ice::SecurityException&)
         {
--- a/cpp/test/IceUtil/timer/Client.cpp
+++ b/cpp/test/IceUtil/timer/Client.cpp
@@ -171,19 +171,20 @@ Client::run(int, char*[])
             task->waitForRun();
             task->clear();
             //
-            // Verify that the same task cannot be scheduled more than once.
+            // Verify that the same task cannot be scheduled more than once. The long delay ensures the task is
+            // still scheduled when we schedule it again.
             //
-            timer->schedule(task, chrono::milliseconds(100));
+            timer->schedule(task, chrono::hours(1));
             try
             {
                 timer->schedule(task, chrono::seconds::zero());
+                test(false);
             }
             catch (const invalid_argument&)
             {
                 // Expected.
             }
-            task->waitForRun();
-            task->clear();
+            timer->cancel(task);
         }
 
         {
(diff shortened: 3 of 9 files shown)


csharp/test/Ice/seqMapping/Custom.cs
```diff
@@ -17,21 +17,21 @@ public T this[int index]
         set => _list[index] = value;
     }
 
-    public void Add(T elmt) => _list.Add(elmt);
+    public void Add(T element) => _list.Add(element);
 
     public override bool Equals(object obj)
```
pepone: Can't we just use SequenceEqual?

```
public override bool Equals(object? obj) =>
    obj is Custom<T> other && _list.SequenceEqual(other._list);
```

pepone (approved):

externl (approved):
Looks good!

csharp/test/Ice/seqMapping/Custom.cs
bernardnormier: Done in 338ef775dd — kept the non-nullable `object obj` signature (the test projects don't enable nullable reference types), and the `is` pattern replaces the cast-plus-catch. `GetHashCode` keeps its per-element loop: `HashCode` has no sequence overload, and `HashCode.Add` uses `EqualityComparer<T>.Default`, matching `SequenceEqual`. Verified with the seqMapping suite (twoway, AMI, collocated).

PR #6445, opened 2026-08-06
Make slice2cpp_generate's mirror and slice2cpp agree about paths

### LINKED ISSUE #6416: slice2cpp_generate: the mirrored layout and slice2cpp disagree about paths
`slice2cpp_generate` mirrors generated files under the `INCLUDE_DIRS` directory that reaches them, so
the `#include` directives slice2cpp emits resolve. That mirror is computed with CMake string operations
while slice2cpp computes its own include paths differently, and the two disagree in three ways.

**A symlinked include root.** slice2cpp follows symlinks and canonicalises its include roots
(`cpp/src/Slice/SliceUtil.cpp`, `cpp/src/slice2cpp/Gen.cpp`), while the CMake side only makes the path
absolute and takes a lexical relative path. With sources addressed through `real/` and `INCLUDE_DIRS`
pointing at `link -> real`, CMake generates a flat `X.h` while the generated `Greeter.h` emits
`#include <pkg/X.h>`, and the build fails with that header missing.

**`changeInclude`'s prefix comparison has no path-boundary test**, so an include root that is a string
prefix of a sibling directory — `/a/b` against a file under `/a/bc` — can be treated as covering it.

**A `;` in a path** is a CMake list separator, so a path containing one splits where it should not.

The first is the one that turns up in practice: a symlinked source or install tree is ordinary,
particularly on macOS and in containerised builds.

### LINKED ISSUE #6532: changeInclude treats a string-prefix sibling directory as a covering include root
`Slice::changeInclude` (`cpp/src/Slice/SliceUtil.cpp`) selects the include root that covers an included file by raw string prefix, with no path-boundary test, and then unconditionally skips one character for the separator. An include directory that is a string prefix of a sibling directory is treated as covering the sibling's files, and the skip chops the emitted path mid-name:

```shell
mkdir -p api apix
echo '// dummy' > apix/A.ice
echo '#include <apix/A.ice>' > B.ice
slice2cpp -Iapi -I. B.ice
grep 'A\.h' B.h
```

```text
#include </A.h>
```

The mangled candidate wins because the shortest-remainder selection prefers it over the correct `apix/A` from `-I.`. It takes an `-I` set containing a string-prefix sibling of a directory holding Slice files, so it is rarely hit, but the result is a generated header that cannot compile.

slice2php and slice2rb emit `require` paths through the same helper and mangle the same way.

externl:
`slice2cpp_generate` must place each generated header at the path the generated code includes it by. That path is decided by slice2cpp, but the function computes it with its own arithmetic, and the two disagreed on symlinks: slice2cpp resolves them, the CMake arithmetic did not. Closes #6416.

Concretely, with a shared Slice API consumed through a stable symlink (`api/current -> v2`) and the file also reachable by its real path:

```cmake
add_library(station STATIC WeatherStation.ice ../api/v2/Acme/Weather.ice station.cpp)
slice2cpp_generate(station INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/../api/current)
```

The `-I` fixes what the compiler emits: it resolves `current -> v2`, sees `Weather.ice` covered at `Acme/Weather.ice`, and generates `#include <Acme/Weather.h>` into `WeatherStation.h`. The wrapper answered the same coverage question lexically — `v2` is not under `current` as strings — so it placed `Weather.h` flat, and the build dies inside a generated header the user never wrote:

```text
generated/station/WeatherStation.h:12:10: fatal error: 'Acme/Weather.h' file not found
```

Both spellings are correct paths to the same file, and nothing in the error mentions a symlink. Mixed spellings are the normal case on macOS, where Homebrew's stable paths are all symlinks into `Cellar/`.

- `slice2cpp.cmake`: canonicalize the include roots and each Slice file with `file(REAL_PATH)` before computing the mirror, except on Windows — matching slice2cpp's own symlink handling, which is also skipped there.
- `SliceUtil.cpp`: `changeInclude`'s prefix match now requires a path boundary, so `-I /a/b` no longer covers files under `/a/bc` and emits a mangled include. This is a pre-existing bug in released slice2cpp, and slice2php and slice2rb share the helper; it gets the changelog entry. Closes #6532.
- Adds a boundary regression case to the `cpp/test/Slice/headers` suite.

The issue's third item, a `;` in a path, needs no guard: the value has already split before the function sees it, and CMake itself rejects the split source path.


externl:
Dropped the `;` guards — they were backwards, and I only caught it by testing the premise.

CMake does not escape semicolons anywhere: a real on-disk `a;b` directory splits into two elements coming back from `file(GLOB)` or a target's `SOURCES`. So by the time `slice2cpp_generate` runs, the damage is done and no element contains a `;`:

| input | list length | guard fired |
|---|---|---|
| `"/tmp/a;b"` — what a user writes, and what GLOB/SOURCES yield | 2 | no |
| `"/tmp/a\;b"` — escaped, works correctly | 1 | yes |

It never caught the failure it was written for, and rejected the only form that works. Unnecessary too: for sources CMake fails first at `add_library` with "Cannot find source file", and for `INCLUDE_DIRS` the split is undetectable from inside the function. CMake's documentation is explicit that lists do not support elements containing `;`.

The two real fixes are unchanged and re-verified against a fresh install: with `INCLUDE_DIRS` pointing through a symlink, `main` still fails on `'pkg/X.h' file not found` while this builds clean and mirrors to `pkg/X.h`, and the non-symlinked control is identical.

externl:
Reworked per discussion: instead of canonicalizing the re-implemented mirror, the mirror is gone. slice2cpp grew `--print-include-paths` (the `changeInclude` answer per file, pure path math — no generation, no file reads), and `slice2cpp_generate` asks it once per target at configure time via `Ice_SLICE2CPP_EXECUTABLE`. One implementation, queried, on every platform — the Windows conditional went with the rest.

Verified against a staged install with the rebuilt compiler: the option answers correctly for covered/uncovered files, both symlink directions, the sibling-prefix boundary case, multi-dot names, and a not-yet-existing file; the symlinked-root scenario builds and mirrors to `pkg/X.h` with the non-symlinked control identical; `HEADER_OUTPUT_DIR`/`INCLUDE_DIR` relocation, duplicate listings, and `Foo.v1.ice` → `Foo.v1.h` all behave; and a pre-option compiler preset via `Ice_SLICE2CPP_EXECUTABLE` fails configure with a message naming the executable and option.

externl:
Backed out the `--print-include-paths` rework — we'd rather not grow slice2cpp's CLI for this. The fix stays in CMake: `file(REAL_PATH)` on the include roots and Slice files, mirroring the compiler's own symlink handling, including its Windows carve-out. Branch reset to that state and the scenario set re-verified against a staged install: the symlinked shared-API case mirrors `Acme/Weather.h` and builds, both symlink directions and the non-symlinked control pass, and `HEADER_OUTPUT_DIR`/`INCLUDE_DIR` relocation, `Foo.v1.ice` naming, and duplicate listings behave.

externl:
Took both Copilot comments — they check out against the repo's conventions.

- `changelog.d/cpp/6416.md` added: this is a user-visible fix to a public issue, which is exactly what `changelog.d/README.md` says warrants a fragment. Two bullets, one per fix.
- Boundary regression test added to `cpp/test/Slice/headers` — which turns out to be precisely the `changeInclude` symlink suite, so it's the right home. The new scenario (`-Iapi -I.` with the file under `apix/`) asserts `#include <apix/A.h>`; the pre-fix compiler emits `#include </A.h>` on it, so it fails without the `SliceUtil.cpp` change. `cpp/allTests.py --filter=Slice/headers` passes.

One correction for the record: the review summary says the PR rejects semicolons in paths — it did in an earlier revision, and no longer does; the description explains why no guard is possible or needed.

externl:
Dropped the changelog fragment: `slice2cpp_generate`'s mirrored layout is new in this release, so the symptom never shipped. (The `changeInclude` boundary fix does touch released slice2cpp behavior, but only for a prefix-sibling `-I` layout nobody has reported — not worth an entry on its own.)

cpp/test/Slice/headers/test.py
```diff
@@ -141,6 +141,22 @@ def runTest(args: str) -> None:
             raise RuntimeError("failed!")
         self.clean()
 
+        #
```
InsertCreativityHere, line 144: Completely unrelated to your PR, but while you're here, can you fix the typo `// dumy file` on line 75 of this file? ;-;

InsertCreativityHere (approved):
Looks reasonable to me!

cpp/test/Slice/headers/test.py
externl: Fixed.

pepone (approved):
`changeInclude` boundary fix and its test look good.

One note: move the changelog fragment to `changelog.d/general/`. It documents a `changeInclude` fix that also changes slice2php and slice2rb output, and says so ("the Slice compilers'", `require`), but `cpp/` files it under C++ Changes where PHP and Ruby users won't see it. `general/6122.md` is the same shape.

externl:
Moved to `changelog.d/general/`.

Issue #6413, opened 2026-08-05
Nothing in CI configures a project against the Ice CMake package


pepone:
Nothing in the repository consumes the Ice CMake package: there is no `find_package(Ice` anywhere in
the tree, and no CI job configures a project against an installed Ice.

The package is therefore only exercised by users. That is how #6348 shipped — *every* versioned
`find_package(Ice <version>)` in CONFIG mode was rejected, because the package shipped no version
file — and #6349 and #6350 are the same shape: defects that a single consumer configure would have
surfaced immediately.

A job that installs Ice to a prefix and then configures and builds a small consumer against it would
cover the whole class. Worth exercising:

- an unversioned `find_package(Ice CONFIG REQUIRED)` and a versioned one
- `COMPONENTS`, including a component with its own dependency (`IceGrid` links `Ice::Glacier2`)
- `slice2cpp_generate` on a `.ice` file, with a hand-written source including the generated header
- at least one multi-config generator, since several past defects only appear in `RelWithDebInfo`
  (#6349)
- a second configure with no changes, which must rebuild nothing

The demos in `zeroc-ice/ice-demos` are close to this already, but they build against a released Ice,
so they cannot catch a regression before it ships.

externl:
> The demos in zeroc-ice/ice-demos are close to this already, but they build against a released Ice,
so they cannot catch a regression before it ships.

We do build against the nightlies but actually testing in the Ice repo would be nice. However, I think the same would be said for all of our packaging.

pepone:
Maybe a release testing job is what we are missing here?

Issue #6412, opened 2026-08-05
Guard against IceVersion.cmake drifting from Ice/Config.h


externl:
The Ice CMake package bakes its version into `cpp/config/IceVersion.cmake` (#6353), hand-maintained like the other version strings in the tree. A missed bump makes `find_library(NAMES Ice${Ice_SO_VERSION})` search for the wrong library names on Windows — Ice reports not found with no hint that the version file is stale — and on other platforms configuration succeeds while `Ice_VERSION` misreports what is installed.

Nothing catches this today: no script rewrites the version strings at release time, and no CI job configures a project against the Ice CMake package (there is no `find_package(Ice` under `.github/workflows/`), which is how #6348 went unnoticed. A CI check that `IceVersion.cmake` agrees with `ICE_STRING_VERSION` and `ICE_SO_VERSION` in `cpp/include/Ice/Config.h` closes the gap cheaply; a `find_package(Ice)` smoke job would close it properly.

Raised by @pepone while reviewing #6353.

externl:
Fixed indirectly by https://github.com/zeroc-ice/ice/pull/6465. Now we only use `Ice/Config.h`.

Issue #6440, opened 2026-08-06
RPM ships the Ice CMake package in the runtime package instead of the devel package


pepone:
On RHEL and Fedora the Ice CMake package configuration ships in the **runtime** package rather than the development one, so a machine that merely runs an Ice application advertises a CMake package it cannot satisfy.

```
packaging/rpm/ice.spec:415  %files -n lib<prefix>ice<mm>-c++        # runtime, no Requires
  ...                         %{_libdir}/libIce.so.*  (and the other runtime sonames)
              :430            %{_libdir}/cmake/*/*.cmake            # <- developer-facing

packaging/rpm/ice.spec:449  %files -n lib<prefix>ice-c++-devel
              Requires:       lib<prefix>ice<mm>-c++, <prefix>ice-slice
              :463            %{_includedir}/Ice ...  %{_bindir}/slice2cpp
```

The dependency only runs one way, which is correct for a runtime package: installing `-devel` pulls the
runtime, never the reverse. The runtime package is also what arrives transitively as a dependency of any
Ice-based program. So `/usr/lib64/cmake/Ice-<mm>/IceConfig.cmake` is present on machines that have no
headers, no `slice2cpp` and no Slice files — not a corrupted installation, just a host with an Ice
application on it.

A developer on such a host gets a confusing failure, because `find_package` locates a configuration file
and then fails inside it:

```
CMake Error at CMakeLists.txt:3 (find_package):
  Found package configuration file:

    /usr/lib64/cmake/Ice-3.9/IceConfig.cmake

  but it set Ice_FOUND to FALSE so package "Ice" is considered to be NOT FOUND.
  Reason given by package:

  Could not locate the Ice installation: /usr/lib64/cmake/Ice-3.9 is not in a
  recognized ;installation layout, or the installation is missing Ice/Ice.h.
```

That reads as though Ice is broken, when the actual problem is that `libzeroc-ice-c++-devel` is not
installed.

Debian does not have this: `libzeroc-ice-dev` ships the headers, the `.so` symlinks, `slice2cpp` **and**
`usr/lib/*/cmake/*/*.cmake` together, and `Depends:` on `libzeroc-ice3.9` and `zeroc-ice-slice`, so a
partial developer installation cannot occur.

Moving `%{_libdir}/cmake/*/*.cmake` into `lib<prefix>ice-c++-devel`, alongside the headers and
`slice2cpp`, matches Debian and removes the failure mode: a runtime-only host then has no configuration
[...]

PR #6465, opened 2026-08-06
Derive the CMake package version from the installed Ice/Config.h

### LINKED ISSUE #6414: Ice version is duplicated between IceVersion.cmake and Ice/Config.h with no consistency check
The Ice version and SO version are stated in two places that nothing compares.

`cpp/config/IceVersion.cmake` carries `_ice_package_version` and `_ice_package_so_version`, while
`cpp/include/Ice/Config.h` carries `ICE_STRING_VERSION` and `ICE_SO_VERSION`. Both are updated by
hand at release time; `config/version.env` and `config/Make.rules` are updated separately again.

If `IceVersion.cmake` is missed in a release bump, the failures are quiet:

- **Windows**: `find_library(NAMES ${component}${Ice_SO_VERSION})` searches for `Ice39a0.dll` against a
  package shipping `Ice39.dll`, so the component is not found and Ice reports as **not found** — with
  no indication that the version file is stale rather than the installation incomplete.
- **Unix**: configuration succeeds and only the reported `Ice_VERSION` disagrees with the headers being
  compiled, so a versioned `find_package` can reject the very installation it names.

Before the version file was introduced, `Ice_SO_VERSION` was parsed out of the installed `Ice/Config.h`
and could not disagree with the binaries.

Comparing the two at configure time — reading `Ice/Config.h`, which is already located by then, and
failing loudly on a mismatch — is cheap insurance. Nothing else in the tree would catch it: no script
rewrites these strings, and no CI job configures against the package (#6348 discussion, and the CI gap
filed separately).

externl:
`IceVersion.cmake` duplicated the version already stated in `Ice/Config.h`, hand-bumped separately at release time, and nothing compared the two — a missed bump made Windows search for the wrong DLL names and Unix misreport `Ice_VERSION`. Before #6353 the package parsed the header; going back to that removes the drift class instead of detecting it. Closes #6414.

- `IceVersion.cmake` reads `ICE_STRING_VERSION` and `ICE_SO_VERSION` from the installed header and sets `Ice_VERSION` and `Ice_SO_VERSION` from them, in place of the hand-maintained literals. It is the only reader; `IceConfig.cmake` and `IceConfigVersion.cmake` include it as they already did.
- Both are left empty unless the header supplies both, so the version file and the config file agree on what counts as a readable installation.
- An installation whose version cannot be read is reported `PACKAGE_VERSION_UNSUITABLE`, so CMake passes over it and carries on to the next candidate, for unversioned requests as well as versioned ones.
- Nothing in the package needs bumping at release time any more.

This also makes #6412 moot: there is no literal left to drift. The `find_package(Ice)` CI smoke job it suggests is #6413.


pepone (commented):
The direction is right. Deriving the version from the header the binaries were built with removes the drift class
instead of policing it, and it makes #6412 moot. On a healthy installation nothing changes: I configured against fake
install trees in both the `lib/cmake/Ice-<mm>` and the Debian multiarch layout and got identical `Ice_VERSION`,
`Ice_SO_VERSION` and identical answers for all twelve version requests I tried (plain, `EXACT`, ranges on both
endpoints, pre-release, bare major).

One structural ask and two small fixes. Nothing here breaks a working installation.

On the approach itself: reading the version out of an installed header is ordinary CMake — 58 of the 162 `Find*.cmake`
modules CMake ships do it. What is unusual is doing it in a *ConfigVersion* file; of the 32 installed on my machine, 31
hardcode `set(PACKAGE_VERSION "x.y.z")` and none read a header. That is not a taste preference: those files are all
*generated* by `write_basic_package_version_file()` during the package's own CMake build, so the version is right by
construction.

The Ice equivalent is a token substituted at install time, and that plumbing is simple and already precedented in both
build systems (`matlab/Makefile:79-87`, `matlab/msbuild/ice.proj:120`). What it does not buy is one less version to
bump. The SO version is free — `soversion` and `IceSoVersion` are both already `39a0`, exactly what `ICE_SO_VERSION`
says. The string version is not: make has `3.9a0`, MSBuild has `3.9a0` and `3.9.0-alpha0`, and this package needs
`Config.h`'s `3.9.0-alpha.0`. Substituting it means adding a variable to `config/Make.rules` *and*
`config/ice.version.props`, and a third spelling of the pre-release suffix in a tree that already has two. `alpha0`
against `alpha.0` is exactly the near-miss nobody catches, and it would leave `Ice_VERSION` disagreeing with what
`glacier2router --version` prints and what the DLL version resources report. Reading `Config.h` binds the package to
the same literal the compiler sees, which is a stronger tie than any substitution can give. Worth a line in the header
comment, because the next reader will wonder why this is not the generated file everyone else ships.

## Do the parse once, not twice — the two copies already disagree

`IceConfigVersion.cmake` accepts an installation that carries `ICE_STRING_VERSION` alone; `IceConfig.cmake:43` requires
`ICE_SO_VERSION` as well. A header with the first and not the second is accepted by the version file — which commits
CMake to that candidate — and then rejected by the config file. That costs the PR its own best property: with a
truncated installation ahead of an intact one on `CMAKE_PREFIX_PATH`, `find_package(Ice 3.9 CONFIG)` gives up instead
of moving on.

Keeping `IceVersion.cmake` and changing its body from a hardcoded `set()` to the header read gives both files one
definition of the version *and* one definition of what a readable installation is. Both files already include it, so
`IceConfig.cmake` and `IceConfigVersion.cmake` go back to their `main` shape and the diff gets *smaller* than the
current one:

```cmake
# The version of this Ice installation, read from the installed Ice/Config.h - the header the
# binaries were built with, so it cannot disagree with the installation.
#
#   _ice_package_version      ICE_STRING_VERSION, pre-release suffix included
#   _ice_package_so_version   ICE_SO_VERSION, which forms the Windows library names
#
# Both are left empty when the header is missing, unreadable or carries neither define.

include("${CMAKE_CURRENT_LIST_DIR}/IcePrefix.cmake")

set(_ice_package_version "")
set(_ice_package_so_version "")

if(DEFINED Ice_INCLUDE_ROOT AND EXISTS "${Ice_INCLUDE_ROOT}/Ice/Config.h")
  file(STRINGS "${Ice_INCLUDE_ROOT}/Ice/Config.h" _ice_config_h REGEX "^#define ICE_(STRING|SO)_VERSION ")
  if(_ice_config_h MATCHES "#define ICE_STRING_VERSION \"([^\"]+)\"")
    set(_ice_package_version "${CMAKE_MATCH_1}")
  endif()
  if(_ice_config_h MATCHES "#define ICE_SO_VERSION \"([^\"]+)\"")
    set(_ice_package_so_version "${CMAKE_MATCH_1}")
  endif()
  unset(_ice_config_h)
endif()

# All or nothing: a header carrying only one of the two is a broken installation, and both readers
# must agree on that, or the version file accepts a candidate the config file then rejects.
if(NOT _ice_package_version OR NOT _ice_package_so_version)
  set(_ice_package_version "")
  set(_ice_package_so_version "")
endif()
```

Sharing the extraction is not by itself enough — that last block is the part that closes the disagreement. Measured
with a truncated installation ahead of an intact one, request `3.9`:

```
main          Ice_FOUND=1   (accepts the truncated one; its version came from the literal, not the header)
PR            Ice_FOUND=0
shared file   Ice_FOUND=1   (rejects the truncated one on version grounds, falls through to the intact one)
```

Everything else is unchanged: same result on a healthy installation, and the same answer for all twelve version
requests.

## `Ice_SO_VERSION` is no longer reset on the broken-installation path

Moving the assignments below the `NOT DEFINED Ice_PREFIX` return means neither is cleared on that path. `Ice_VERSION`
happens to be overwritten by CMake from `PACKAGE_VERSION`; `Ice_SO_VERSION` is not a variable CMake manages, so a
caller's earlier value survives into a failed `find_package`:

```cmake
set(Ice_SO_VERSION "999-STALE")
find_package(Ice CONFIG QUIET)     # against an installation with no headers
# main  ->  Ice_FOUND=0  Ice_SO_VERSION=39a0
# PR    ->  Ice_FOUND=0  Ice_SO_VERSION=999-STALE
```

The comment carried over from `main` still says the assignments are "set unconditionally", which after the move is the
one thing they are not. Keeping the two empty `set()`s above the prefix check restores both the behaviour and the
comment.

## One behaviour change to be aware of, not a blocker

For a *versioned* request against an installation that has the cmake files but no header, the message changes:

```
main   Could not locate the Ice installation: <dir> is not in a recognized installation layout,
       or the installation is missing Ice/Ice.h.
PR     Could not find a configuration file for package "Ice" that is compatible with requested
       version "3.9".  ... IceConfig.cmake, version: unknown
```

CMake rejects the candidate before `IceConfig.cmake` runs, so `Ice_NOT_FOUND_MESSAGE` is never set, and the reader is
told they have the wrong version when in fact they have the right version and no headers.

I do not think this matters, because the state that produces it is going away: #6440 moves the cmake files into
`lib<prefix>ice-c++-devel`, which owns the headers and the `.so` symlinks and is unversioned, so a runtime-only host
will have no configuration file at all and CMake will say so. Debian already ships them in `-dev`. What is left is a
genuinely partial installation, where "version: unknown" is a fair answer.

It is also not a one-way trade. Because the candidate is now *rejected* rather than accepted-and-then-failed, CMake
continues to the next one: with a broken install ahead of a good one on `CMAKE_PREFIX_PATH`,
`find_package(Ice 3.9 CONFIG)` succeeds under the PR and fails outright on `main`.

The only thing I would change is the comment at `IceConfigVersion.cmake:11-14`. "A missing or unreadable header reports
no version, and every versioned request fails" reads as a cost being accepted; it is the behaviour you want, and the
comment should say why.

## Nits

- `IceConfigVersion.cmake:31-32` — `unset(Ice_PREFIX)` / `unset(Ice_INCLUDE_ROOT)` are dead. CMake evaluates the
  version file in a nested scope; I added a `set()` to that file and it was invisible to the caller. The comment's
  reason ("this file must leave only `PACKAGE_*` behind") misstates the scoping rather than describing it.
- `IceConfigVersion.cmake:6-7` — "so the file is maintained by hand" is now only true of the logic, not of the version.
- `IceConfig.cmake:33` (114 chars) and `IceConfigVersion.cmake:22` (108) run past the 100 the rest of `cpp/config`
  holds to (`IceConfig` 100, `IcePrefix` 100, the deleted `IceVersion` 99). Falls out of the first point anyway.

## Checked and fine

- The regexes are robust against the trailing `// "A.B.C", with A=major...` comments, which contain quotes of their
  own, and neither `ICE_INT_VERSION` nor `#  define ICE_LIBNAME(NAME) NAME ICE_SO_VERSION "D.lib"` false-matches.
- "A missing or unreadable header reports no version" holds for the ordinary permission case: `if(EXISTS)` is itself
  read-permission-sensitive, so a header the configuring user cannot read short-circuits the guard rather than
  reaching `file(STRINGS)`, which would have been a hard error.
- The `CACHE STRING` entries and the `if(NOT DEFINED Ice_VERSION)` guard that #6353 deliberately removed are correctly
  not reintroduced.
- No dangling references to the deleted file anywhere in the tree.
- No changelog fragment, consistent with #6353 and #6420.

CI is 48/48, but per #6413 nothing in the tree configures against the Ice CMake package, so green says nothing about
this change — this PR is squarely the class #6413 exists to cover.

externl:
Took all three.

`IceVersion.cmake` is back, doing the read for both files, with the all-or-nothing block. Confirmed the fall-through with a truncated installation ahead of an intact one, request `3.9`: main selects the truncated one, this now selects the intact one.

`Ice_SO_VERSION` is cleared on the broken-installation path again — your repro now reports empty instead of `999-STALE`.

Comments reworded: why this reads the header rather than shipping the generated file, and the no-version case stated as intended behaviour. Dead `unset()`s and the "maintained by hand" line are gone, and the new file stays within 100 columns.

Matrix unchanged on a healthy installation.

externl:
Added `PACKAGE_VERSION_UNSUITABLE` for the unreadable-header case, which extends the fall-through you liked to unversioned requests. Measured with the truncated installation ahead of an intact one:

```
                     find_package(Ice 3.9)   find_package(Ice)
before               falls through           fails outright
after                falls through           falls through
```

The reason string goes in `PACKAGE_VERSION` so it shows against the rejected candidate, the way the generated files carry `(32bit)`:

```
IceConfig.cmake, version: unknown, Ice/Config.h is missing or unreadable
```

One consequence worth your call, since it partly revisits your second point. Once the version file rejects the candidate, `IceConfig.cmake` is never loaded, so nothing of ours runs and a caller's `Ice_SO_VERSION` survives again — your `999-STALE` repro is back to `999-STALE` for an installation with no headers. That is inherent to UNSUITABLE rather than fixable in the config file. It looks like the better trade to me: falling through to a usable installation beats clearing a variable the caller should not be reading after `Ice_FOUND` comes back false, and for what it is worth none of the 15 config packages installed on my machine clear result variables on a failure path. Say the word and I will drop it.

Both early returns in `IceConfig.cmake` are now unreachable through `find_package` for the same reason; I left them as defense for a direct `include()` or a missing version file.

externl:
Simplified further since the last round: `IceVersion.cmake` now sets `Ice_VERSION` and `Ice_SO_VERSION` itself rather than exporting `_ice_package_*` for `IceConfig.cmake` to copy. That drops the temporaries, both copy assignments, both `unset()`s and the explicit clearing block — `IceConfig.cmake` ends up shorter than it is on `main`, and the stale-value property comes for free wherever the config file still runs, since the variables are always assigned before the first early return.

Also restored the original `IceConfigVersion.cmake` header comment. With the `include()` already there, that file is a one-line diff, and the explanation belongs in `IceVersion.cmake` where the reading happens.

Two things that came out of checking rather than assuming:

- Clearing result variables on a failure path is not a convention. Of the 15 config packages installed on my machine, 10 handle not-found and 3 use early returns; none clear a result variable. CMake's contract is that they are unspecified once `<Pkg>_FOUND` is false.
- `PACKAGE_VERSION` and friends do not leak: I set them in the caller before `find_package` and they came back untouched, confirming the version file runs in a nested scope — which is also why the `unset()`s you flagged were dead.

pepone (approved):
Approving. All three of my earlier points are addressed, and letting `IceVersion.cmake` assign the variables
directly rather than exporting temporaries came out better than what I sketched.

Two things I would change, both small. Everything below them is confirmation, not an ask.

## 1. Reword the reason string

`IceConfigVersion.cmake:23` is now the *only* diagnostic a broken installation produces — once the version
file rejects the candidate, both messages in `IceConfig.cmake` are unreachable through `find_package`. So four
different failures arrive as one string, and for the third it is false:

| installation | actually wrong | user is told |
|---|---|---|
| not a recognized layout, or no `Ice.h` | layout / no include tree | `Ice/Config.h is missing or unreadable` |
| include tree present, no `Config.h` | accurate | `Ice/Config.h is missing or unreadable` |
| `Config.h` present **and readable**, only `ICE_STRING_VERSION` | missing `ICE_SO_VERSION` | ← false |
| `Config.h` unreadable | accurate | `Ice/Config.h is missing or unreadable` |

CMake's framing around it is what makes the wording matter. `find_package(Ice REQUIRED CONFIG)` — no version
anywhere — against a runtime-only RHEL box:

```
  Could not find a configuration file for package "Ice" that is compatible
  with requested version "".
    /usr/lib64/cmake/Ice-3.9/IceConfig.cmake, version: unknown, Ice/Config.h is missing or unreadable
      The version found is not compatible with the version requested.
```

A user who never requested a version is told twice they have a version problem. Leading with the diagnosis
recovers most of it, since our string prints before CMake's trailer:

```cmake
set(PACKAGE_VERSION "unknown - broken installation, cannot read the Ice version from Ice/Config.h")
```

True for all four rows. I re-ran the fall-through matrix with it — unchanged, all eight cells still reach the
intact installation.

## 2. `IceVersion.cmake:9-10`

"carries neither define" understates the all-or-nothing block, which also empties both when exactly one is
present. "does not carry both defines" matches the code. (Copilot's third suppressed comment, and it is the
one it got right.)

## Your call, no ask

- `IceConfig.cmake:16` — `include(IcePrefix.cmake)` is now redundant; line 17 pulls in `IceVersion.cmake`,
  which includes it again. Idempotent, so it is duplicate work, not a defect (`IcePrefix` runs three times per
  successful `find_package`). Dropping it makes the dependency implicit, so keeping it is defensible.
- `IceConfig.cmake:27` — that return precedes the `unset(Ice_INCLUDE_ROOT)` on line 42, so on the direct
  `include()` path you kept these returns for, it leaves that internal in the caller's scope.

## Keep `PACKAGE_VERSION_UNSUITABLE` — you asked

Keep it. It buys the fall-through for every broken shape, not just the one you measured — I ran all four
against a truncated installation ahead of an intact one, and `main` either picks the broken one or fails
outright in all eight cells, where this reaches the intact one in all eight.

The `Ice_SO_VERSION` objection does not survive contact with `main`: on the same failed `find_package`, `main`
leaves `Ice_SO_VERSION=39a0` from the literal — a confident wrong answer for an installation it just reported
as not found. Neither branch clears anything. Your reading of the contract and your survey both hold.

**Do not act on Copilot's other two suppressed comments** — I checked both against CMake's implementation and
they are wrong. `CheckVersion` loads the version file before testing whether a version was requested, and
`CheckVersionFile` computes `!unsuitable && (okay || requested_version.empty())`, so `UNSUITABLE` does reject
an unversioned request, exactly as your comment claims. And `if(EXISTS)` resolves to `access(path, R_OK)`, so
a mode-000 `Config.h` fails the guard and `file(STRINGS)` never runs.

## FYI — RPM co-installation, argues for #6440

Reading the version from the prefix's header assumes that header belongs to those binaries. On RPM it may not:
`ice.spec:430` ships the cmake files in the **versioned** runtime package, `:465` ships headers in the
**unversioned** devel package, so `Ice-3.8/` and `Ice-3.9/` can coexist and both resolve to the one installed
`/usr/include`. With `libice38-c++` + `libice39-c++` + one 3.9 devel:

```
find_package(Ice 3.9)   main -> Ice-3.9/IceConfig.cmake     PR -> Ice-3.8/IceConfig.cmake
find_package(Ice 3.8)   main -> "found 3.8.3", hands over 3.9 headers    PR -> not found
```

Neither is right, and the PR has the better half — `main`'s 3.8 answer is the drift bug in packaging form.
What it loses is which config *directory* wins; the user still gets 3.9 headers and libs, driven by 3.8's
`IceTargets.cmake`. Nearly a no-op today. The fix is #6440, not this PR — Debian, NuGet and Homebrew are
already immune. Both are 3.8.3; worth landing together.

## Checked and fine

Healthy installation identical to `main` across fourteen version requests. Reproduced #6414 on a fake Windows
NuGet layout with the literal drifted from the shipped binaries: `main` gets `IMPLIB-NOTFOUND` plus a hard
error at `IceTargets.cmake:208`, this finds the right `Ice39.lib`/`.dll`. Regexes still exclude
`ICE_INT_VERSION` and the indented `ICE_LIBNAME` line. No dangling `_ice_package_*`. Nothing leaks from the
version file's nested scope. No changelog fragment, matching #6420 and #6353. The 101-char line is inside the
102 that file already carries on `main` — withdrawing that one from last round.

## CI

47/48. `release on ubuntu-24.04-arm` fails on `DataStorm/partial` hanging in the multicast case — unrelated,
already on `main` (deff5566cd), and in my hands. Nothing in the build or test suite reads `cpp/config/*.cmake`
at all, which is #6413's point and why green would not have told us anything either.

externl:
Both taken. The reason string now leads with the diagnosis and is true for all four rows of your table, and the IceVersion.cmake comment says "does not carry both defines". Re-checked: message correct for the truncated case, fall-through unchanged, healthy installation unchanged.

Leaving the two optional ones — the redundant `IcePrefix` include stays for the explicit dependency, and the `Ice_INCLUDE_ROOT` leak is confined to the direct-`include()` path those returns exist for.

On the RPM finding: agreed, and we should ship the cmake files only with the devel package, which is #6440. Landing this now rather than waiting, since it is strictly better than `main` in that configuration.

PR #6466, opened 2026-08-06
Give each sample-filtered DataStorm reader a node-unique session facet

### LINKED ISSUE #6347: DataStorm: sample filtering is defeated when same-name topic readers collide on the session facet
## Summary

A reader that declares a sample filter is isolated from other readers by a per-reader session facet. The facet is
derived from the reader's **element** id, which restarts at 1 in every topic object. Two same-name topic readers in
one process therefore produce two filtered readers with the same facet, they share a single writer-side `Listener`,
and each one receives every sample the *other* one's filter matched.

Sample filtering is not merely leaky in this configuration — it is completely defeated in both directions.

## Mechanism

A reader that configures a sample filter is given a facet named after its element id:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/DataElementI.cpp#L1324-L1330

(and identically for a filtered-key reader at
[`DataElementI.cpp#L1691-L1697`](https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/DataElementI.cpp#L1691-L1697)).

The writer groups subscribers into listeners keyed by session and facet:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/DataElementI.h#L57-L73

But element ids are drawn from a counter owned by each topic:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/TopicI.h#L182-L186

and several topic readers may share a name, each stored separately by the factory:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/TopicFactoryI.cpp#L36-L44

So the first filtered reader of topic object A and the first filtered reader of same-name topic object B are both
element 1, and both take the facet `fa1`. They collapse into one `{session, "fa1"}` listener holding two
subscribers.

From there the leak is unavoidable:

1. `Listener::matchOne` forwards the sample if **any** subscriber in the listener matches — it is a per-listener
   test, not a per-subscriber one:
   https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/DataElementI.h#L86-L104
2. One invocation goes out on facet `fa1`.
3. The receiver dispatches to every element subscribed to that writer element, gated only on `initialized` and the
   key — the sample filter is never re-evaluated, because the reader side never constructs one:
   https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/SessionI.cpp#L1741-L1758
4. The only remaining gate is the facet string compare, which both readers pass:
   https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/DataElementI.cpp#L997-L1005

## Reproduction

Confirmed by experiment, not by inspection. Two single-key readers on key `k`, each with a sample filter accepting
values whose first character is its own criteria or `z`; one single-key writer publishing `aaa`, `bbb`, `zzz`.

Expected: reader A sees `aaa, zzz`; reader B sees `bbb, zzz`.

**Two same-name topic objects** — each reader is element 1, so both get facet `fa1`:

```
PROBE readerA got: aaa,bbb  readerB got: aaa,bbb
```

Both filters are defeated: each reader received the sample only the other one matched, and neither filter excluded
anything.

**Control — the same two readers on a single topic object**, so they are elements 1 and 2 and get facets `fa1` and
`fa2`:

```
PROBE readerA got: aaa,zzz  readerB got: bbb,zzz
```

Correct. The facet collision is the whole cause; sample filtering is otherwise sound.

## Impact

An application that creates more than one `Topic` object with the same name — a supported and tested configuration
(see the "sample routing across same-name reader topics" cases in `cpp/test/DataStorm/events`) — and gives each a
sample-filtered reader silently receives samples it asked to be filtered out. The samples are delivered as ordinary
samples through the normal callbacks and the unread queue, with no warning and no trace record, so the reader
cannot tell them from samples it did match.

The collision needs the filtered readers to land on the same element id, which is the common case when each topic
object's first element is its filtered reader. Readers created after other elements on the same topic get distinct
ids and are unaffected.

This does not cross a process boundary: the facet is scoped to a session, so all colliding readers live in the same
node. It is a correctness defect within one application, not a disclosure across a trust boundary.

## Suggested fix

Make the facet unique across the node rather than across the topic. `TopicI` already carries an id that the factory
allocates from a per-node counter:

https://github.com/zeroc-ice/ice/blob/838c8c94edb5270e539ad99e0d25453aaa398879/cpp/src/DataStorm/TopicFactoryI.cpp#L43

so a facet built from the topic id and the element id — for example `fa<topicId>-<elementId>` — restores the
one-listener-per-filtered-reader invariant the current code intends.

Worth considering alongside it: `Listener::matchOne`'s any-subscriber semantics are what turn a facet collision
into a data leak rather than a merely redundant send. Making the forwarding decision per subscriber, or asserting
that a facet-bearing listener holds exactly one subscriber, would make a future regression of this kind loud
instead of silent.

## Related

- #6309 — the same writer-side filtering machinery; its proposed fix relies on a filtered listener being
  addressable on its own, which this defect currently breaks.
- #6213, #6306 — other DataStorm sample-path defects found in the same review.

---

Found while scoping the fix for #6309. Analysis by Claude (Opus 5), cross-checked with Codex (gpt-5.6-sol), which
independently identified the id-collision path; the leak was then confirmed and isolated with the probe and control
runs above against 838c8c9. Not yet human-triaged, so the milestone is left unset.

pepone:
Fixes #6347.

A reader that declares a sample filter is isolated from the other readers of its session by a facet the writer
addresses it under. The facet was built from the reader's element id alone, and element ids restart at 1 in every
topic object — so two same-name topic objects in one node produced two sample-filtered readers with the facet
`fa1`. They collapsed into a single writer-side listener, which forwards a sample as soon as *any* of its
subscribers matches, and the only reader-side gate is the facet compare. Each reader received the union of the
two filters' matches rather than only its own.

The facet is now qualified with the topic id, which the topic factory allocates from a per-node counter, so it
identifies exactly one reader element of the node. The construction also moves into the `DataReaderI` constructor,
where the sample filter it depends on is recorded; the two subclasses were running identical copies of it.

New test in `cpp/test/DataStorm/events`: two same-name topics, each with a sample-filtered reader on the writer's
key, one filtering on `"a"` and the other on `"b"`, against a writer publishing `a`, `b` and `ab`. Without the fix
reader A's second sample is `b` rather than `ab`. All 108 C++ suites pass.

## Not in this PR

This makes the facet unique within a node; it does not make it survive a relay hop. A relay node forwards session
invocations without the facet, so a sample-filtered reader behind a relay receives nothing live — filed separately
as #6467, since it is a different defect in a different place and this fix does not depend on it.

Verifying this fix also turned up #6468: an unfiltered reader sharing a key and session with a sample-filtered
reader receives every matching sample twice, because a reader without a facet accepts an invocation on any facet.
Also left out.

### DIFF
--- a/changelog.d/datastorm/6347.md
+++ b/changelog.d/datastorm/6347.md
@@ -1,0 +1,3 @@
+- Fixed a bug affecting nodes that create two or more `Topic` instances with the same name. A reader with a sample
+  filter on one of these topics also received the samples another such reader's filter matched, and its own filter
+  no longer excluded anything.
--- a/changelog.d/datastorm/6347.md
+++ b/changelog.d/datastorm/6347.md
@@ -1,3 +1,3 @@
-- Fixed a bug affecting nodes that create two or more `Topic` instances with the same name. A reader with a sample
-  filter on one of these topics also received the samples another such reader's filter matched, and its own filter
-  no longer excluded anything.
+- Fixed a bug affecting nodes that create two or more `Topic` instances with the same name. A reader with a
+  sample filter on one of these topics received every sample that matched any such reader's filter, instead of
+  only the samples its own filter matched.
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -727,6 +727,15 @@ DataReaderI::DataReaderI(
     {
         _config->sampleFilter =
             FilterInfo{.name = std::move(sampleFilterName), .criteria = std::move(sampleFilterCriteria)};
+
+        // Sample filtering is evaluated on the writer, which groups the subscribers of a session by facet and
+        // forwards a sample to a facet as soon as one of its subscribers matches. A sample-filtered reader
+        // therefore needs a facet of its own, or it also receives the samples another reader's filter matched.
+        // Element ids restart at 1 in every topic and a node can hold several same-name topics, so the facet is
+        // qualified with the topic id, which the topic factory allocates per node.
+        ostringstream os;
+        os << "fa" << topic->getId() << '-' << _id;
+        _config->facet = os.str();
     }
 }
 
@@ -1320,14 +1329,6 @@ KeyDataReaderI::KeyDataReaderI(
         Trace out(_traceLevels->logger, _traceLevels->dataCat);
         out << this << ": created key reader";
     }
-
-    // If sample filtering is enabled, ensure the updates are received using a session facet specific to this reader.
-    if (_config->sampleFilter)
-    {
-        ostringstream os;
-        os << "fa" << _id;
-        _config->facet = os.str();
-    }
 }
 
 void
@@ -1686,14 +1687,6 @@ FilteredDataReaderI::FilteredDataReaderI(
     {
         Trace out(_traceLevels->logger, _traceLevels->dataCat);
         out << this << ": created filtered reader";
-    }
-
-    // If sample filtering is enabled, ensure the updates are received using a session facet specific to this reader.
-    if (_config->sampleFilter)
-    {
-        ostringstream os;
-        os << "fa" << _id;
-        _config->facet = os.str();
     }
 }
 
--- a/cpp/test/DataStorm/events/Reader.cpp
+++ b/cpp/test/DataStorm/events/Reader.cpp
@@ -759,6 +759,26 @@ void ::Reader::run(int argc, char* argv[])
         doneWriter.waitForReaders();
         doneWriter.update(0);
     }
+
+    // Two same-name topics on this node, each with a sample-filtered reader on the one writer's key. Each reader is
+    // the first element of its own topic, so the two topics number them identically. Each reader must receive only
+    // the samples its own filter matches: the writer publishes "a", "b" and "ab", and the reader filtering on "a"
+    // must see "a" and "ab" while the reader filtering on "b" must see "b" and "ab".
+    {
+        Topic<string, string> topicA(node, "sameNameSampleFilter");
+        Topic<string, string> topicB(node, "sameNameSampleFilter");
+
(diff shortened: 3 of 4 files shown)


changelog.d/datastorm/6347.md
```diff
@@ -0,0 +1,3 @@
+- Fixed a bug affecting nodes that create two or more `Topic` instances with the same name. A reader with a sample
+  filter on one of these topics also received the samples another such reader's filter matched, and its own filter
+  no longer excluded anything.
```
bernardnormier: "its own filter no longer excluded anything" is not quite right — samples matching neither filter were still excluded. What each reader actually received was the union of the same-facet readers' matches:
```suggestion
- Fixed a bug affecting nodes that create two or more `Topic` instances with the same name. A reader with a
  sample filter on one of these topics received every sample that matched any such reader's filter, instead of
  only the samples its own filter matched.
```

bernardnormier (approved):
You can drop the interoperability paragraph from the description — the Lookup2 change already breaks DataStorm wire compatibility between 3.8.2 and 3.8.3, so the cross-version behavior of the facet format does not matter.

externl (approved):
Looks good.

PR #6469, opened 2026-08-06
Clean up the C++ MSBuild build system

pepone:
Cleanup and fixes for the C++ MSBuild build system, following a review of `cpp\msbuild`.

## Fixes

- **`EventLoggerMsg.h` generation raced across parallel configurations**: `mc` wrote the header into the shared project directory and then `move`d it into the per-config generated include directory, so concurrent configurations (`/p:BuildAllConfigurations=yes`) could collide. The command now uses `mc -h` to write the header directly to its destination (and creates the directory first, removing the ordering dependency on the Slice compilation target). This also collapses the four identical per-config `Command`/`Outputs` blocks into one.
- **`SliceCompileClean` left generated headers behind**: it deleted headers from `OutputDir`, but the library projects emit them to `HeaderOutputDir` (`include\generated\...`). Clean now deletes from `HeaderOutputDir` when it is set.
- **Wrong build dependency in `ice.test.slnx`**: `IceGrid/session/verifier` depended on `IceGrid/allocation/verifier` (an unrelated test project) instead of `testcommon`.
- **`Clean` never cleaned the Slice tools solution**: `CleanDist` now cleans `ToolsSolution` since `BuildDist` builds it.
- **Paths with spaces broke `Exec` commands**: the nuget.exe invocations are now quoted, and `Pack`'s bare relative paths (which resolve against the process working directory) are anchored to `$(MSBuildThisFileDirectory)`.
- **The NuGet package's debugger environment replaced `PATH`**: `ZeroC.Ice.Cpp.targets` now appends `;%PATH%` instead of discarding the inherited value.

## Cleanup

- Removed the `Publish` target: it pushed the package to the local NuGet global-packages folder to support running the demos against a local build, which is no longer a supported configuration; it is not documented in any BUILDING instructions and nothing invokes it. (It was also subtly broken: `TrimStart('global-packages: ')` trims a *character set*, not a prefix, corrupting global-packages paths such as `e:\nuget\packages\`.)
- Removed the dead `SymbolServer` property (leftover from the removed `DownloadSymbols` target) and the unused `Microsoft.Cpp.Default.props` import from `ice.proj`.
- Removed the redundant `BuildDependency` entries on `slice.vcxproj` from the `slice2*` projects in `ice.slnx`; their `ProjectReference`s already enforce the build order (`slice2rb` has relied on this all along).
- Removed the duplicated `TargetName` property groups from `ice.test.props` (identical to the ones `ice.cpp.props` already sets) and collapsed the four `TargetName` groups in `ice.cpp.props` into one.
- Removed no-op compiler settings: `MinimalRebuild` (the underlying `/Gm` option no longer exists), `RuntimeTypeInfo=true` (the default), and the `_CONSOLE` define (unused, and meaningless for the DLL/static-library builds it was applied to).
- `RemovePackages` uses the `RemoveDir` task instead of shelling out to `rmdir`.
- Replaced the hard-coded version in `ZeroC.Ice.Cpp.nuspec` with a `0.0.0` placeholder; the actual version is always set via `nuget pack -Version`.
- Unified the `NuGetExe`/`GetNuGet` spelling in `config/ice.common.targets` and fixed the import to use the file's actual name (`ice.common.targets`), which only worked thanks to case-insensitive file systems.
- Dropped the obsolete `ToolsVersion` attribute and a stray `\` in import paths.

## Testing

On Windows (VS 2026), from a clean worktree:
- `msbuild /m ice.proj /t:BuildDist /p:Platform=x64 /p:Configuration=Release` — clean build, 0 warnings/errors; verified the generated `EventLoggerMsg.h` and the `39a0` target names.
- `msbuild ice.proj /t:Pack` — package builds with the correct version and layout (`build/native`, `tasks`, `tools`, `slice`, `cmake`).
- `msbuild /m ice.proj /t:Clean` — `bin` fully cleaned and no generated headers left under `include\generated` (previously they survived).

### DIFF
--- a/cpp/msbuild/ice.slnx
+++ b/cpp/msbuild/ice.slnx
@@ -107,29 +107,13 @@
   <Project Path="../src/Slice/msbuild/slice.vcxproj">
     <BuildDependency Project="../src/IceUtil/msbuild/iceutil/iceutil.vcxproj" />
   </Project>
-  <Project Path="../src/slice2cpp/msbuild/slice2cpp.vcxproj">
-    <BuildDependency Project="../src/Slice/msbuild/slice.vcxproj" />
-  </Project>
-  <Project Path="../src/slice2cs/msbuild/slice2cs.vcxproj">
-    <BuildDependency Project="../src/Slice/msbuild/slice.vcxproj" />
-  </Project>
-  <Project Path="../src/slice2java/msbuild/slice2java.vcxproj">
-    <BuildDependency Project="../src/Slice/msbuild/slice.vcxproj" />
-  </Project>
-  <Project Path="../src/slice2js/msbuild/slice2js.vcxproj">
-    <BuildDependency Project="../src/Slice/msbuild/slice.vcxproj" />
-  </Project>
-  <Project Path="../src/slice2matlab/msbuild/slice2matlab.vcxproj">
-    <BuildDependency Project="../src/Slice/msbuild/slice.vcxproj" />
-  </Project>
-  <Project Path="../src/slice2php/msbuild/slice2php.vcxproj">
-    <BuildDependency Project="../src/Slice/msbuild/slice.vcxproj" />
-  </Project>
-  <Project Path="../src/slice2py/msbuild/slice2py.vcxproj">
-    <BuildDependency Project="../src/Slice/msbuild/slice.vcxproj" />
-  </Project>
+  <Project Path="../src/slice2cpp/msbuild/slice2cpp.vcxproj" />
+  <Project Path="../src/slice2cs/msbuild/slice2cs.vcxproj" />
+  <Project Path="../src/slice2java/msbuild/slice2java.vcxproj" />
+  <Project Path="../src/slice2js/msbuild/slice2js.vcxproj" />
+  <Project Path="../src/slice2matlab/msbuild/slice2matlab.vcxproj" />
+  <Project Path="../src/slice2php/msbuild/slice2php.vcxproj" />
+  <Project Path="../src/slice2py/msbuild/slice2py.vcxproj" />
   <Project Path="../src/slice2rb/msbuild/slice2rb.vcxproj" />
-  <Project Path="../src/slice2swift/msbuild/slice2swift.vcxproj">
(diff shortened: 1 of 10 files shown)


cpp/msbuild/ice.slnx
```diff
@@ -107,29 +107,13 @@
...
-  <Project Path="../src/slice2matlab/msbuild/slice2matlab.vcxproj">
-    <BuildDependency Project="../src/Slice/msbuild/slice.vcxproj" />
-  </Project>
-  <Project Path="../src/slice2php/msbuild/slice2php.vcxproj">
-    <BuildDependency Project="../src/Slice/msbuild/slice.vcxproj" />
-  </Project>
-  <Project Path="../src/slice2py/msbuild/slice2py.vcxproj">
-    <BuildDependency Project="../src/Slice/msbuild/slice.vcxproj" />
-  </Project>
+  <Project Path="../src/slice2cpp/msbuild/slice2cpp.vcxproj" />
```
bernardnormier, line 110: Follow-up suggestion: the remaining `BuildDependency` entries exist only because the executable/DLL projects don't declare the libraries they link with — linking relies on the auto-link pragmas, and ordering on these hand-maintained solution edges. We could instead give each project a `ProjectReference` to the libraries it links with (as the slice2* projects already do) and drop the `BuildDependency` graph entirely — a new project then can't forget its solution edges.

bernardnormier (approved):
Looks good.

externl (approved):
Looks good!

cpp/msbuild/ice.slnx
pepone: Filed as #6500 — good idea. One wrinkle the survey turned up: of the 59 edges in `ice.slnx`, 14 point at `slice2cpp` (an exe — those order codegen, not linking) and one is `icegridadmin` → `icebox.vcxproj`, which can't be a link edge at all and looks like it was meant to be `iceboxlib`. So 44 convert cleanly and the tool-ordering ones need `LinkLibraryDependencies=false`. The bigger win is `ice.test.slnx`: 216 edges across 213 projects, none using `ProjectReference`.

PR #6470, opened 2026-08-06
Add 'to_s' Alias for Slice Classes and Structs in Ruby

### LINKED ISSUE #6390: Ruby: to_s on a Slice class or struct instance prints nothing useful
In the Ruby mapping, calling `to_s` on a Slice class or struct instance — directly, via `puts`, or via string interpolation — returns Ruby's default `#<Test::MyClass:0x0000000124704d40>`, which conveys nothing beyond the type name.

Both constructs already have a useful `inspect`:

- `Ice::Value#inspect` returns the Ice stringification of the instance (all fields, with back references for shared instances).
- Generated structs use Ruby's default `inspect`, which lists the fields: `#<Test::Point:0x... @x=1, @y=2>`.

The rest of the mapping does better: proxies define both `to_s` and `inspect` (both return the stringified proxy), generated enums return the enumerator name from `to_s`, and generated exceptions return the Slice type id.

The fix is to alias `to_s` to `inspect`:

- on `Ice::Value` (one line in `Value.rb`), and
- in the code slice2rb generates for structs (structs are plain generated classes with no common base class).

InsertCreativityHere:
This PR fixes #6390, by adding an alias `to_s -> inspect` for all Slice classes and structs.

The issue proposed adding this alias to structs via `slice2rb`, but I don't think this is necessary.
Every struct we generate already mixes in the `Struct.rb` file, so adding it just to this file should be sufficient,
and I'd argue, cleaner as well.


bernardnormier (approved):

externl (approved):

InsertCreativityHere:
Waiting to merge this PR until #6482.
Our code to stringify a `struct` is broken on main, and the test this PR added accidentally exercised this...

Issue #6467, opened 2026-08-06
DataStorm: a relay node drops the session facet, so sample filtering delivers nothing behind a relay


pepone:
## Summary

A sample-filtered reader receives its samples on a dedicated session facet. A relay node forwards session
invocations by rebuilding the target proxy from the request's identity and **never copies the facet**, so a
relayed sample arrives with an empty facet and every sample-filtered reader behind the relay discards it.

Live sample delivery to a sample-filtered reader is therefore completely broken whenever the writer and the
reader reach each other through a relay node. The reader receives nothing, with no error and no trace record
unless data tracing is enabled.

## Mechanism

A reader that configures a sample filter is addressed under a facet:

https://github.com/zeroc-ice/ice/blob/ea2a76a9224246da4fba5793d6f2fe72bd3f74cb/cpp/src/DataStorm/DataElementI.cpp#L1324-L1330

The writer keeps one listener per `{session, facet}` and applies the facet to that listener's session proxy:

https://github.com/zeroc-ice/ice/blob/ea2a76a9224246da4fba5793d6f2fe72bd3f74cb/cpp/src/DataStorm/DataElementI.h#L76-L82

so the outgoing `s` invocation carries the facet. When the two nodes are connected through a relay, that
invocation is dispatched by the relay's `SessionForwarder`, which reconstructs the target proxy from the identity
alone:

https://github.com/zeroc-ice/ice/blob/ea2a76a9224246da4fba5793d6f2fe72bd3f74cb/cpp/src/DataStorm/NodeSessionManager.cpp#L26-L50

`createProxy` yields a proxy with the default (empty) facet, and `current.facet` is not carried over. The
operation, mode, in-parameters and context are all forwarded; only the facet is lost.

On the receiving node, the reader compares the invocation's facet against its own and skips the sample:

https://github.com/zeroc-ice/ice/blob/ea2a76a9224246da4fba5793d6f2fe72bd3f74cb/cpp/src/DataStorm/DataElementI.cpp#L997-L1005

## Reproduction

Confirmed by experiment against `ea2a76a922`, not by inspection. A relay node with a writer application and a
reader application attached to it, neither with a public endpoint (the topology of
`cpp/test/DataStorm/fanInRelay`). One topic with a `contains` sample filter and a single key; the reader node
creates a sample-filtered reader on that key **and** an unfiltered reader on the same key as a control. A second
unfiltered topic carries a barrier value the writer publishes immediately after the sample, so the check runs only
[...]

PR #6473, opened 2026-08-06
Deliver a sample once to a DataStorm reader that has no sample filter

### LINKED ISSUE #6468: DataStorm: an unfiltered reader receives a duplicate of every sample a sample-filtered reader on the same key matches
## Summary

A reader with no facet accepts an invocation on **any** facet — the reader-side gate rejects a sample only when
the reader's own facet is set and differs. An unfiltered reader that shares a key and a session with a
sample-filtered reader therefore receives each matching sample twice: once from the writer's unfaceted listener,
and once from the faceted listener that exists for the filtered reader.

## Mechanism

The writer keeps one listener per `{session, facet}` and forwards a sample to every listener that has at least
one matching subscriber:

https://github.com/zeroc-ice/ice/blob/ea2a76a9224246da4fba5793d6f2fe72bd3f74cb/cpp/src/DataStorm/DataElementI.cpp#L1658-L1672

With an unfiltered reader and a sample-filtered reader on the same key of the same session, that is two listeners:
`{session, ""}` and `{session, "fa…"}`. A sample the filter matches goes out twice, once per listener.

On the receiving node, `SubscriberSessionI::s` hands the sample to every element subscribed to that writer
element, passing the invocation's facet along:

https://github.com/zeroc-ice/ice/blob/ea2a76a9224246da4fba5793d6f2fe72bd3f74cb/cpp/src/DataStorm/SessionI.cpp#L1765-L1782

and the reader's gate skips a sample only if the reader itself has a facet:

https://github.com/zeroc-ice/ice/blob/ea2a76a9224246da4fba5793d6f2fe72bd3f74cb/cpp/src/DataStorm/DataElementI.cpp#L997-L1005

The filtered reader correctly ignores the unfaceted copy. The unfiltered reader has no facet, so the condition is
false for both copies and it queues each of them.

## Reproduction

Confirmed by experiment against `ea2a76a922`, not by inspection. One topic with a `contains` sample filter, one
single-key writer, and two readers on the same key in one node: one unfiltered, one with a sample filter matching
`"a"`. The writer publishes `a` then `b`.

Expected: the unfiltered reader sees `a, b`.

```
PROBE plain reader got: a,a
```

`a` was delivered twice; `b` follows behind it. The sample is duplicated for every value the other reader's filter
matches, so a stream where most values match is nearly doubled for the unfiltered reader.

## Impact

Medium. The duplicate is delivered as an ordinary sample through the normal callbacks and the unread queue, with
the same sample id, event and timestamp as the original, so the application cannot distinguish it from a real
sample. It also consumes a slot of the reader's `sampleCount` budget.

A duplicated `PartialUpdate` is worse: `DataReaderI::queue` has no duplicate-id guard and the two copies arrive
as separate invocations, so the second one is resolved against the base the first one already installed and the
delta is applied twice. Confirmed with a second probe — an `add` updater (`value += delta`), a filter matching
everything, `add(10)` then `partialUpdate("add")(5)`:

```
PROBE plain got: 10 10 15 20
```

Expected `10 15`. The unfiltered reader ends at 20, a value the writer never published.

It needs an unfiltered and a sample-filtered reader on the same key in one node, both reading the same writer
element. Readers on different keys or in different nodes are unaffected.

## Suggested fix

Make the reader-side gate symmetric: a reader without a facet should accept only an unfaceted invocation, i.e.
compare `_config->facet.value_or("")` against the invocation facet rather than skipping the check when the
reader has no facet.

That is a one-line change but it should be weighed against the alternative of not sending the duplicate at all:
the writer knows which subscribers belong to the unfaceted listener, so the `matchOne` decision for the faceted
listener could exclude subscribers that are already served by the unfaceted one. The reader-side gate is the
smaller and safer of the two.

## Related

- #6347 — the same facet mechanism, colliding across same-name topic readers.

---

Found while verifying the fix for #6347. Analysis by Claude (Opus 5), cross-checked with Codex (gpt-5.6-sol),
which independently identified the asymmetric facet gate; the duplicate delivery was then confirmed with the probe
above against `ea2a76a922`. Not yet human-triaged, so the milestone is left unset.

pepone:
Fixes #6468.

A sample-filtered reader is addressed under a session facet of its own, so a writer that has both an unfiltered
and a sample-filtered reader on one key forwards a matching sample to two destinations: the unfaceted one and the
facet. The receiving session hands each copy to every element subscribed to that writer element, and the
reader-side gate rejected a sample only when the reader *itself* had a facet — so a reader without one accepted
both copies.

The gate now compares both sides, so a reader with no facet accepts only an unfaceted sample. Every reader is
attached to exactly one destination — its own facet, or the unfaceted one — so this rejects only the copies
addressed elsewhere.

Measured before the fix, on the same key of one writer:

- full values: the unfiltered reader read `a,a` where the writer published `a` then `b`;
- a partial update: with an `add` updater and `add(10)` then `partialUpdate("add")(5)`, it read `10 10 15 20`
  instead of `10 15` — the delta resolved twice, against the base the first copy had already installed, leaving
  the reader on a value the writer never published.

New test in `cpp/test/DataStorm/events`, next to the other sample-filter cases: an unfiltered and a
sample-filtered reader on one key, against a writer publishing `a` (matches) then `b` (does not). Without the fix
the unfiltered reader's second read is the duplicate `a` rather than `b`. All 108 C++ suites pass.

## Notes

Independent of #6466 — that one changes the facet's format and where it is built, this one changes how the facet
is compared on receipt. Neither depends on the other and the hunks do not overlap.

Behind a relay node the duplicate survives this fix, because #6467 strips the facet in transit and both copies
arrive unfaceted. That path behaves identically before and after this change; it is fixed by #6467, not here.

The writer still puts both copies on the wire — the duplicate is discarded on receipt rather than never sent.
Suppressing the send would mean teaching `matchOne` about subscribers that another listener already serves, which
is a larger change than the defect warrants.

### DIFF
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -996,8 +996,7 @@ DataReaderI::queue(
 {
     // The writer forwards a sample once per destination facet, and this reader is attached to exactly one of them:
     // the facet it configured for its sample filter, or the unfaceted destination when it has no sample filter.
-    // Both sides are compared, so a reader without a facet accepts only an unfaceted sample: testing the facet
-    // only when this reader has one would also give it the copy addressed to a sample-filtered reader of the key.
+    // Accept only the copy addressed to this reader's destination.
     if (_config->facet ? *_config->facet != facet : !facet.empty())
     {
         if (_traceLevels->data > 2)
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -994,7 +994,11 @@ DataReaderI::queue(
     const chrono::time_point<chrono::system_clock>& now,
     bool checkKey)
 {
-    if (_config->facet && *_config->facet != facet)
+    // The writer forwards a sample once per destination facet, and this reader is attached to exactly one of them:
+    // the facet it configured for its sample filter, or the unfaceted destination when it has no sample filter.
+    // Both sides are compared, so a reader without a facet accepts only an unfaceted sample: testing the facet
+    // only when this reader has one would also give it the copy addressed to a sample-filtered reader of the key.
+    if (_config->facet ? *_config->facet != facet : !facet.empty())
     {
         if (_traceLevels->data > 2)
         {
--- a/changelog.d/datastorm/6468.md
+++ b/changelog.d/datastorm/6468.md
@@ -1,0 +1,3 @@
+- Fixed a bug where a reader without a sample filter received a duplicate of every sample that a sample-filtered
+  reader of the same key and writer matched. The duplicate was indistinguishable from a real sample and, for a
+  partial update, applied the update to the reader's value twice.
--- a/cpp/test/DataStorm/events/Reader.cpp
+++ b/cpp/test/DataStorm/events/Reader.cpp
@@ -405,6 +405,23 @@ void ::Reader::run(int argc, char* argv[])
         }
     }
 
+    // An unfiltered reader and a sample-filtered reader on the same key of the same writer. The writer forwards a
+    // matching sample to the filtered reader's facet as well as to the unfaceted destination, and the unfiltered
+    // reader must receive it once: the writer publishes "a" and then "b", so this reader reads "a" then "b".
+    {
+        Topic<string, string> topic(node, "unfilteredWithSampleFilter");
+
+        auto plain = makeSingleKeyReader(topic, "elem", "", config);
+        auto filtered = makeSingleKeyReader(topic, "elem", Filter<string>("contains", "a"), "", config);
+
+        test(plain.getNextUnread().getValue() == "a");
+        test(plain.getNextUnread().getValue() == "b");
+        test(!plain.hasUnread());
+
+        test(filtered.getNextUnread().getValue() == "a");
+        test(!filtered.hasUnread());
+    }
+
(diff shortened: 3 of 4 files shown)


cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -994,7 +994,11 @@ DataReaderI::queue(
     const chrono::time_point<chrono::system_clock>& now,
     bool checkKey)
 {
-    if (_config->facet && *_config->facet != facet)
+    // The writer forwards a sample once per destination facet, and this reader is attached to exactly one of them:
+    // the facet it configured for its sample filter, or the unfaceted destination when it has no sample filter.
+    // Both sides are compared, so a reader without a facet accepts only an unfaceted sample: testing the facet
+    // only when this reader has one would also give it the copy addressed to a sample-filtered reader of the key.
```
bernardnormier: The last sentence describes the old bug: "testing the facet only when this reader has one" is the pre-fix gate in prose, so once this merges the reader has to reconstruct code that no longer exists to follow it. That rationale is already in the PR description, where it belongs. ("Both sides are compared" also just restates the line below.) The first two sentences state the standing invariant and are worth keeping:

```suggestion
    // The writer forwards a sample once per destination facet, and this reader is attached to exactly one of them:
    // the facet it configured for its sample filter, or the unfaceted destination when it has no sample filter.
    // Accept only the copy addressed to this reader's destination.
```

bernardnormier (approved):
Looks good.

PR #6475, opened 2026-08-06
Preserve the session facet when a DataStorm relay forwards an invocation

### LINKED ISSUE #6467: DataStorm: a relay node drops the session facet, so sample filtering delivers nothing behind a relay

pepone:
Fixes #6467.

A reader that declares a sample filter is addressed by the writer under a session facet of its own, and
`DataReaderI::queue` discards a live sample that does not arrive under that facet. The relay's `SessionForwarder`
rebuilt the target proxy from the request identity with `connection->createProxy(id)` and never carried
`current.facet` over, so every live sample the filter matched arrived unfaceted and was discarded. Sample
filtering delivered nothing whenever the writer and the reader met through a relay node — the documented topology
for nodes with no public endpoint.

Initialization was unaffected, because an initialization batch is routed to the reader element rather than to its
facet. A sample-filtered reader behind a relay therefore appeared to start correctly and then never updated, with
no error and no trace record unless data tracing was enabled.

The fix carries the facet over to the forwarded proxy. It is part of the request's addressing, like the identity,
operation, mode, parameters and context the forwarder already replays; `ice_facet("")` is a no-op, so every
unfaceted invocation — which is all of them except live samples to a sample-filtered reader — is unchanged.

`SessionForwarder` is the only DataStorm forwarder that reconstructs a proxy for an invocation that can carry a
facet. The other forwarders (`NodeI::forwardToSubscribers`/`forwardToPublishers`, `TopicI`'s forwarder,
`NodeSessionManager::forward`, the collocated `ForwarderManager`, `NodeForwarder`) either replay a stored session
proxy or handle operations that never carry one.

## Test

New `cpp/test/DataStorm/relaySampleFilter` suite: a writer application and a reader application, neither with a
public endpoint, meeting through a relay node. The reader declares a sample filter; the writer publishes `a`, `b`
and `ac` **after** the reader attached, so they travel the live path rather than the initialization batch, and the
reader must end up with exactly `a` and `ac`.

A second, unfiltered topic carries a barrier value published last, so the check distinguishes "not delivered yet"
from "never delivered". That is a sound fence: every hop dispatches a connection's requests one at a time — the
relay's adapter gets a per-adapter thread pool because DataStorm sets `DataStorm.Node.Server.ThreadPool.Serialize`
(`ObjectAdapterI` creates one whenever any `<adapter>.ThreadPool.*` property is set), and the endpoint-less
applications dispatch on the client thread pool, which DataStorm serializes too.

Without the fix the reader receives nothing and the assertion fails. All 109 C++ suites pass, and the new suite
was run 20 times without a flake.

## Notes

Independent of #6466 and #6473 — the forwarder treats the facet as an opaque string and neither parses nor builds
it, so the facet format change in #6466 needs nothing here.

This does not introduce the duplicate delivery of #6473 on the relay path: before this fix both copies of a
matching sample arrived unfaceted and an unfiltered reader on the same key already accepted both. Verified by
running the same probe with and without this change — the unfiltered reader reads `a a b ac ac` either way. #6473
is what fixes that, on the relay path as well as directly.

### DIFF
--- a/changelog.d/datastorm/6467.md
+++ b/changelog.d/datastorm/6467.md
@@ -1,0 +1,2 @@
+- Fixed a bug where a reader with a sample filter received no updates when it read the writer through a relay
+  node. The reader was still initialized correctly, so it appeared to start normally and then never updated.
--- a/cpp/msbuild/ice.test.slnx
+++ b/cpp/msbuild/ice.test.slnx
@@ -68,6 +68,14 @@
       <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
     </Project>
   </Folder>
+  <Folder Name="/DataStorm/relaySampleFilter/">
+    <Project Path="../test/DataStorm/relaySampleFilter/msbuild/reader/reader.vcxproj">
+      <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
+    </Project>
+    <Project Path="../test/DataStorm/relaySampleFilter/msbuild/writer/writer.vcxproj">
+      <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
+    </Project>
+  </Folder>
   <Folder Name="/DataStorm/reliability/">
     <Project Path="../test/DataStorm/reliability/msbuild/reader/reader.vcxproj">
       <BuildDependency Project="../test/Common/msbuild/testcommon.vcxproj" />
--- a/cpp/src/DataStorm/NodeSessionManager.cpp
+++ b/cpp/src/DataStorm/NodeSessionManager.cpp
@@ -31,10 +31,13 @@ namespace
                 if (auto session = _nodeSessionManager->getSession(
                         Identity{.name = current.id.name.substr(pos + 1), .category = ""}))
                 {
-                    // Forward the call to the target session, don't need to wait for the result.
+                    // Forward the call to the target session, don't need to wait for the result. The facet is part
+                    // of the addressing and is carried over: a writer addresses a sample-filtered reader under a
+                    // facet of its own, and the reader discards a sample that does not arrive under it.
                     Identity id{.name = current.id.name.substr(0, pos), .category = current.id.category.substr(0, 1)};
                     session->getConnection()
                         ->createProxy(std::move(id))
+                        ->ice_facet(current.facet)
                         ->ice_invokeAsync(
                             current.operation,
                             current.mode,
--- a/cpp/test/DataStorm/relaySampleFilter/Makefile.mk
+++ b/cpp/test/DataStorm/relaySampleFilter/Makefile.mk
@@ -1,0 +1,9 @@
+# Copyright (c) ZeroC, Inc.
+
+$(project)_programs        = reader writer
+$(project)_dependencies    = DataStorm Ice TestCommon
+
+$(project)_reader_sources  = Reader.cpp
+$(project)_writer_sources  = Writer.cpp
+
+tests += $(project)
--- a/cpp/test/DataStorm/relaySampleFilter/Reader.cpp
+++ b/cpp/test/DataStorm/relaySampleFilter/Reader.cpp
@@ -1,0 +1,48 @@
+// Copyright (c) ZeroC, Inc.
+
+#include "DataStorm/DataStorm.h"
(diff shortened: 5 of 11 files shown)


bernardnormier (approved):
Looks good.

externl (approved):

PR #6476, opened 2026-08-06
Clean up the Python MSBuild build system

pepone:
Cleanup and fixes for the Ice for Python MSBuild build system, following the same review as #6469.

## Fixes

- **`SliceCompileClean` was a no-op**: it deleted `@(PythonGenerated)`, an item that has been defined nowhere since the Python code-generation refactor (#4213), so `Clean` never removed the generated `.py` files. The target now cleans the five fully-generated packages (`Glacier2`, `IceBox`, `IceGrid`, `IceMX`, `IceStorm`) with `**\*.py` globs, which preserves the tracked `py.typed` markers. The generated modules mixed into the `Ice` package are deliberately left alone — they are enumerated only in the `python/python` `.gitignore` — and a comment documents this.
- **Paths with spaces broke the `Exec` commands**: the nuget.exe and slice2py.exe invocations now quote the tool path, output directory, include path, and each Slice file. Because `%(RecursiveDir)` ends with a backslash, quoting the test output directory would produce a trailing `\"` escape, so the `TestSliceCompile` `OutputDir` trims the trailing backslash.
- **Dead line removed**: the `SliceCompile` target ran `MakeDir` on `$(SliceCompileOutputDir)`, a property defined nowhere.

## Cleanup

- Deleted `python/config/Ice.Python.props`: referenced nowhere, and broken — it imports `cpp/config/ice.cpp.props`, a path that does not exist.
- Deleted `python/msbuild/ice.tests.props`: referenced nowhere; it contained only empty `Build`/`Clean` targets and a stale `UseBinDist`/`ICE_HOME` conditional.
- Fixed the `ice.props` import of `ice.cpp.props`, which was spelled `Ice.Cpp.props` and worked only thanks to case-insensitive file systems.
- The `Build` target now uses the existing `$(IceToolsPath)` property instead of repeating the slice2py path inline.
- Removed the unused `Microsoft.Cpp.Default.props` import from `ice.proj` (orchestration-only project), a stray `\` in the `ice.common.targets` import path, and the obsolete `ToolsVersion` attribute; quoted the `SliceCompile` target condition.

## Testing

On Windows (VS 2026), from a clean worktree:
- `msbuild /m ice.proj /t:Build /p:Platform=x64 /p:Configuration=Release /p:PythonHome=...` — full build (C++ dist targets, IcePy, dist and test Slice compilation), 0 warnings/errors.
- `msbuild ice.proj /t:Clean` — the five generated packages contain zero `.py` files afterwards and every `py.typed` survives (previously Clean removed nothing).


externl (approved):

bernardnormier (approved):
Looks good.

PR #6477, opened 2026-08-06
Clean up the MATLAB MSBuild build system

pepone:
Cleanup and fixes for the Ice for MATLAB MSBuild build system, following the same review as #6469.

## Fixes

- **`SliceCompileClean` was a no-op**: it deleted `%(OutputDir)\%(Filename).m`, but slice2matlab generates MATLAB package directories (`+Ice`, `+Test`, ...) with one `.m` file per Slice *type*, so nothing ever matched and `Clean` left all generated code behind. Since every output directory (`lib\generated` and the per-test `generated` dirs) contains only generated code, the target now removes the output directories.
- **The test Slice compilation ran with an empty `$(Platform)`** unless the caller passed `/p:Platform=x64` explicitly (as CI happens to do): `ice.proj` invoked `ice.matlab.test.proj` without passing `Properties`, and the parent's `Platform` is a project-level property, which does not flow to the child — so `IceToolsPath` resolved to `cpp\bin\\Release`. The test-project invocations now pass `Platform`/`Configuration` explicitly, and `ice.matlab.props` gained a `Platform` default (`x64`) alongside its existing `Configuration` default.
- **`Package` never cleaned up its temp files**: `Delete Files="toolbox.prj"` pointed at the project directory, but the file is generated in `..\toolbox\`; `info.xml` was not deleted at all. Both are now deleted at their real location.
- **Paths with spaces broke the `Exec` commands**: the nuget.exe and slice2matlab.exe invocations now quote the tool path, output directory, and each Slice file.

## Cleanup

- Deleted `matlab/msbuild/ice.tests.props`: dead — its `SliceOutputDir` property is referenced only by the file itself, and nothing in the repository invokes it.
- `RootDir` was computed as `GetFullPath($(MSBuildThisFileDirectory)..\msbuild)`, a self-referential round-trip whose only effect is trimming the trailing backslash; it is now a plain `TrimEnd`.
- The unconditional `<Platform>x64</Platform>` in `ice.proj` is now a conditioned default, like `Configuration` (command-line properties always won anyway).
- `matlab/lib/msbuild/ice.proj` and `ice.matlab.test.proj` declared `DefaultTargets="Build"` but have no `Build` target — invoking them without `/t:` failed; they now default to `SliceCompile`.
- Removed the unused `Microsoft.Cpp.Default.props` import from `ice.proj`, the obsolete `ToolsVersion` attributes, and the meaningless `DefaultTargets` attribute on the `ice.matlab.targets` targets file.

## Testing

On Windows (VS 2026, MATLAB R2025b), from a clean worktree:
- `msbuild /m ice.proj /t:Build /p:Configuration=Release` — full build (C++ dist targets, MATLAB solution, lib and test Slice compilation), 0 warnings/errors. This invocation — without `/p:Platform` — failed before the `Platform` propagation fix.
- `msbuild ice.proj /t:Clean` — `lib\generated` and all test `generated` directories are removed (previously Clean removed nothing).
- `msbuild ice.proj /t:Package` — produces `ice-3.9.0.0-R2025b-win.mltbx`; the generated `toolbox.prj` and `info.xml` are cleaned up afterwards, and the `@RootDir@` replacement in `toolbox.prj` is byte-identical to the old computation.


externl (approved):

bernardnormier (approved):
Looks good.

One spot the paths-with-spaces fix missed (on a line outside the diff, so no inline suggestion): in `ice.matlab.test.proj`, the `Ice/optional` item's `AdditionalOptions` carries `-I$(MSBuildThisFileDirectory)..\test\Ice\optional`, which lands in the slice2matlab command unquoted — from a checkout path with spaces, the dist compile now works but this one test compile still breaks. Suggest:

```xml
<AdditionalOptions>-I&quot;$(MSBuildThisFileDirectory)..\test\Ice\optional&quot;</AdditionalOptions>
```

Issue #6480, opened 2026-08-06
`Value.inspect` Always Fails due to Undefined `ICE_TYPE`


InsertCreativityHere:
This is a sister-issue to #6429 where the same underlying problem exists, but it's not limited to exceptions, it also affects Value types too.

Classes and proxies seem to be okay though.

InsertCreativityHere:
Fixed by #6482.

Issue #6429, opened 2026-08-05
Ruby: inspect raises NameError on every Slice user exception


pepone:
> [!NOTE]
> Found while reviewing #6386 (flagged in [the review](https://github.com/zeroc-ice/ice/pull/6386#pullrequestreview-4863224081)
> as "filing separately"). Long-standing — the same shape is present on 3.8 and 3.7 — so it is not a regression from
> that PR or from #6360.

## Summary

`inspect` raises `NameError` on every Slice-generated user exception:

```ruby
Ice::loadSlice(["-I.", "Test.ice"]) # ruby/test/Ice/objects
Test::EDerived.new.inspect
# => NameError: uninitialized constant Test::EDerived::ICE_TYPE
```

So `p myException` — and anything else that stringifies through `inspect`, such as an exception escaping to the
top level of `irb` — blows up instead of producing the member dump `ExceptionInfo::print` is meant to produce.
`to_s` is unaffected and returns the Slice type id.

## Root cause

`Ice::UserException#inspect` (`ruby/ruby/Ice/Exception.rb:17`) calls `Ice::__stringifyException`, which reads an
`ICE_TYPE` constant from the exception's class (`ruby/src/IceRuby/Types.cpp:3070` at a4622f63e1):

```cpp
volatile VALUE cls = CLASS_OF(exc);
volatile VALUE type = callRuby(rb_const_get, cls, rb_intern("ICE_TYPE"));
```

But `ICE_TYPE` is only ever defined by `defineProxy` (`Types.cpp:3013`) and `defineClass` (`Types.cpp:3037`) —
never by `IceRuby_defineException`, and `slice2rb` does not emit it either. Verified against a `main` build:

```ruby
defined?(Test::EDerived::ICE_TYPE) # => nil         (exception: missing)
defined?(Test::A1::ICE_TYPE)       # => "constant"  (class: present)
```

## Consequences

- `IceRuby_stringifyException` is the only caller of `ExceptionInfo::print` (`Types.cpp:3075`), and the
  `rb_const_get` above raises before reaching it — so the member-dump formatting for user exceptions is
  unreachable dead code.
- The fix is either to define `ICE_TYPE` in `defineException` (matching classes and proxies) or to have
  `__stringifyException` look up the `ExceptionInfo` by type id instead.

The broader stringification inconsistency across generated constructs is tracked for 3.9 in #6391, whose table
documents this row; this issue tracks the concrete defect so it can be fixed independently.

[...]

PR #6482, opened 2026-08-06
Fix `inspect` for Generated Exception and Struct Classes in Ruby

InsertCreativityHere:
Previously, due to a missing 'ICE_TYPE' constant, these methods would just always raise a 'NameError'...
Now we properly define the constant, so the `inspect` implementations work as-is.

This fixes both #6429 and #6480, which were identical issues, one for structs and one for exceptions. But the same underlying bug.


externl (approved):
LGTM!

pepone (approved):
Changes look good. A few minor things:

**`volatile` isn't needed.** In both new blocks `result` is passed to `rb_define_const` and then returned, so it's live across the only GC-capable call and the conservative stack/register scan will find it. The sibling `defineEnum`/`defineSequence`/`defineDictionary` return `createType(info)` bare, and `IceRuby_TypeInfo_defineProxy` does the same `rb_define_const(type, "ICE_TYPE", self)` with no guard. Plain `VALUE result` would be more consistent. (Not worth blocking over — it's harmless either way.)

**Changelog fragment wording:**

```
- Fixed the stringification of Slice user exceptions and structs: calling `inspect` on a Slice-generated exception or
  struct instance returns a description, instead of raising `NameError`.
```

One possible follow-up, out of scope here: `StructInfo::print` goes straight to `out.sb()`, so struct `inspect` output has no type name, unlike `exception ::Test::C` and `object #0 (::Test::C)`.

PR #6486, opened 2026-08-07
Clean up the C# MSBuild build system

pepone:
Cleanup and fixes for the Ice for C# MSBuild build system, following the same review as #6469.

## Fixes

- **`Publish` mangled the NuGet global-packages path**: `TrimStart('global-packages: ')` trims a *character set*, not a prefix, so a global-packages folder on a drive whose letter is in that set (e.g. `e:\nuget\packages\`) came out as `\nuget\packages\`. Unlike the C++ `Publish` target (removed in #6469 because it was undocumented), the C# one is documented in `csharp/BUILDING.md`, so it is fixed with a regex that strips everything up to and including the `global-packages:` label. Since `dotnet nuget locals` prints the path with a trailing backslash — which would escape the closing quote of the now-quoted `--source` argument — a `.` is appended after a trailing backslash (trimming it instead would corrupt a drive-root path such as `E:\`, a valid `NUGET_PACKAGES` override).
- **Paths with spaces broke the `Exec` commands**: the `dotnet nuget push` package path and `--source` argument, and the `dotnet pack` solution path, are now quoted.

## Cleanup

- Deleted `csharp/msbuild/ice.test.props`: dead — nothing imports it (the test projects import `ice.common.props` directly). Its content was never applied anywhere: a legacy `PlatformTarget`/`Prefer32Bit` group, per-configuration overrides, and six completely empty `PropertyGroup`s for platforms (`Win32`, `x64`) that the C# build doesn't even have.
- Simplified the solution `Properties` metadata in `ice.proj`: the `PackageVersion` conditional was always true (`PackageVersion` is unconditionally set from `IcePackageVersion`, which always has a value), so it is now part of the base `Properties`; also removed a stray trailing `;`.
- Deduplicated the debug settings in `ice.common.props`: `DebugType=portable` and `DebugSymbols=true` were set in the base group and then re-set identically in the per-configuration groups, which now contain only `Optimize`.
- Dropped the obsolete `ToolsVersion` and meaningless `DefaultTargets` attributes from the props/targets files, a stray `\` in the `ice.proj` import paths, a literal tab in the middle of `ice.proj`, and quoted the `SliceCompile` target condition in `ZeroC.Ice.Slice.Tools.targets`.

## Testing

On Windows, from a clean worktree (after building `slice2cs` from the C++ tree):
- `dotnet msbuild msbuild/ice.proj /t:Build` — full solution (134 projects), 0 warnings/errors.
- `dotnet msbuild msbuild/ice.proj /t:Pack` — all dist packages produced.
- `dotnet msbuild msbuild/ice.proj /t:Publish` — packages land in the local global-packages folder with the expected layout.
- `dotnet msbuild msbuild/ice.proj /t:Clean` — succeeds.


externl (approved):
Looks good.

bernardnormier (approved):
Looks good.

PR #6501, opened 2026-08-07
Remove dead code in the Ice for MATLAB decoder and encoder

### LINKED ISSUE #6461: MATLAB: dead IceInternal.Buffer branches, unused Encoding_1_0 and write-only properties
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI dead-code audit and verified by re-reading the source on `main` (`82b79108e2b`).

## Summary

Unreferenced code in the MATLAB `+IceInternal` package.

- **`Buffer.m`** — the only construction site anywhere is `+Ice/OutputStream.m:17`, `IceInternal.Buffer()`, always zero-argument, so the `if nargin == 1` constructor branch is dead. `expand` has no callers at all. (The `.reset(` hits elsewhere resolve to `Ice.InputStream.reset` or `IceInternal.ReadEncaps.reset`; `Ice.InputStream.buf` is a plain `uint8` array, not a `Buffer`.)
- **`Protocol.m:15`** — `Encoding_1_0` has no users; `Encoding_1_1` is used at `+Ice/OutputStream.m:9`.
- **`EncapsEncoder10.m`** — `valueIdIndex` is set to `0` in the constructor and never read; `registerValue` uses `v.iceInternal_` instead.
- **`EncapsDecoder11_InstanceData.m`** — `typeIdIndex` is assigned four times and never read, and `next` is assigned once and never read (`EncapsDecoder11` always allocates a fresh instance and pops via `previous`). *Leave `EncapsEncoder11_InstanceData.next` alone — that one is read.*

**Severity:** Low.

bernardnormier:
Clears the `+IceInternal` dead-code items in #6461, plus two adjacent cleanups that fall out of them.

### As filed in #6461

- **`Buffer.m`** — the only construction site (`+Ice/OutputStream.m`) always calls `Buffer()` with no argument, so the `nargin == 1` unmarshaling branch was dead. Removed it and the unused `data` parameter. `expand` had no callers; removed.
- **`Protocol.m`** — `Encoding_1_0` was never read (`Encoding_1_1` is used by `OutputStream`); removed.
- **`EncapsEncoder10.m`** — `valueIdIndex` was set to `0` in the constructor and never read (`registerValue` uses `v.iceInternal_`); removed.
- **`EncapsDecoder11.m` / `EncapsDecoder11_InstanceData.m`** — the per-decoder `typeIdIndex` property was assigned on four paths in `startSlice` and never read; removed. Also removed `next` (see below).

### Beyond the audit

- **`EncapsDecoder10.m`** — the 1.0 decoder had the *same* write-only `typeIdIndex` property, which #6461 did not mention. Removed it too. (The type-id map counter it duplicated is tracked by the base `EncapsDecoder` via `length(typeIdMap)`; C#'s 1.0 decoder has no such per-decoder field either.)
- **`EncapsDecoder.readTypeId`** — with `typeIdIndex` gone from both decoders, both callers discard the returned index, so `readTypeId` now returns only the type id instead of `[typeIdIndex, typeId]`.

### Why `next` is removable in MATLAB but not in C#

The MATLAB instance-data frame is a port of C#'s `InputStream.InstanceData`, which links frames both ways. In C#, `next` is a frame-reuse cache: `push` does `_current = _current.next ?? new InstanceData(_current)`, so a popped frame is reused on the next push instead of being reallocated, and the constructor's `previous.next = this` wires that cache up.

The MATLAB port copied the links but not the reuse. Both push sites unconditionally allocate (`obj.current = EncapsDecoder11_InstanceData(obj.current)`) and nothing ever reads `next`, so the field was written but never used. Rather than port the C# optimization — instance-data churn on this path isn't a measured cost in MATLAB — this removes the unused `next`. `previous`, which backs the push/pop stack (read in `endInstance`), stays.

No changelog fragment: internal dead-code removal, nothing a user observes.

### Testing

Full MATLAB suite 16/16. `objects`, `optional`, and `slicing` exercise both the 1.0 encoder and the 1.0/1.1 decoder paths touched here.

Fixes #6461.


externl (approved):
Looks good.

One you may or may not want: `Buffer.reset` is dead as well. The only `Buffer` is `OutputStream.m:17`, which touches `buf`, `capacity`, `push`, `pushByte`, `resize` and `size` but never `reset` — the other `.reset(` hits in MATLAB are `InputStream`'s encaps cache and `InputStream.reset`. Its `nargin == 2` branch is the same unmarshaling path you removed from the constructor. Your call whether it's worth keeping as a buffer API.

pepone (approved):

PR #6503, opened 2026-08-07
Load a PKCS12 certificate with an empty password on macOS

### LINKED ISSUE #6489: SSL transport: a certificate with an empty IceSSL.Password can never be loaded on macOS
On macOS, `loadKeychainItems` only sets `params.passphrase` when the password is non-empty:

```cpp
// cpp/src/Ice/SSL/SecureTransportUtil.cpp:248
UniqueRef<CFStringRef> passphraseHolder;
if (!passphrase.empty())
{
    passphraseHolder.reset(toCFString(passphrase));
    params.passphrase = passphraseHolder.get();
}
```

`SecItemImport` rejects a PKCS#12 blob with a null passphrase unconditionally — it returns `errSecPassphraseRequired` (-25260, *"Passphrase is required for import/export"*) before looking at the file contents. So when `IceSSL.CertFile` points at a PKCS#12 and `IceSSL.Password` is empty or unset, the import always fails on macOS, regardless of how the file was produced.

Measured on macOS 15 (Darwin 25.5.0) against a PKCS#12 with an empty password:

| `params.passphrase` | Result |
| --- | --- |
| `NULL` (current behavior) | FAIL -25260 `errSecPassphraseRequired` |
| `""` | OK |

The iOS path is unaffected — it always sets `kSecImportExportPassphrase`, possibly to an empty string (`SecureTransportUtil.cpp:652-654`).

### Suggested fix

Always set `params.passphrase`. The existing guard appears to be there so that unencrypted PEM private keys import cleanly, and forcing an empty string does not regress them — verified with a fresh keychain per case, importing `certs/configuration/ca1/server_key.pem` as `kSecItemTypePrivateKey`:

```
passphrase = NULL  (current)     OK
passphrase = ""    (proposed)    OK
```

### Note

This is necessary but not sufficient for loading `certs/configuration/ca1/*_password_less.p12`. Those are written by `openssl pkcs12 -export -passout pass:`, and Apple's PKCS#12 parser rejects them for a separate reason (it accepts only one of the two RFC 7292 empty-password encodings). That is tracked separately; this issue is only about the unconditional `-25260`.

pepone:
Closes #6489

## The bug

`SecItemImport` rejects a PKCS#12 blob with `errSecPassphraseRequired` (-25260) whenever `params.passphrase` is null, before it looks at the file. `loadKeychainItems` left it unset when `IceSSL.Password` was empty, so an empty-password certificate could never load on macOS. The first commit always passes the passphrase, even an empty one.

## Why that alone is not enough

RFC 7292 gives two incompatible answers for an empty password: appendix B.1 makes it two `0x00` bytes, appendix B.2 step 3 a zero-length block. They derive different keys, so a file written under one reading fails MAC verification under the other.

`openssl`, `keytool` and python-`cryptography` write the B.1 form. OpenSSL, Schannel, Java and .NET read both. The two Apple platforms each read only one, and not the same one — each rejects the other with `errSecAuthFailed` (-25293), measured by calling `SecPKCS12Import` on the fixture bytes:

| Fixture | Empty password encoding | Loads on |
| --- | --- | --- |
| `{client,server}_password_less.p12` | two `0x00` bytes, B.1 (`openssl -passout pass:`) | every platform except macOS |
| `{client,server}_null_password.p12` | zero-length block, B.2 step 3 (`make_passwordless_pkcs12.py`) | every platform except iOS |

Both carry the same algorithms (SHA-1 MAC/2048, PBE-SHA1-3DES), so the encoding is the only variable. There is no single file that loads on both, and `openssl` cannot write the B.2 form — hence the generator script and the second fixture.

## Tests

C# and Java parse PKCS#12 themselves and run both fixtures. C++ selects per platform, since `ICE_USE_SECURE_TRANSPORT` covers macOS *and* iOS:

```cpp
#if defined(ICE_USE_SECURE_TRANSPORT_MACOS)
    testPasswordLessCert(factory, defaultProps, "null_password");
#elif defined(ICE_USE_SECURE_TRANSPORT_IOS)
    testPasswordLessCert(factory, defaultProps, "password_less");
#else   // OpenSSL and Schannel read both
    testPasswordLessCert(factory, defaultProps, "null_password");
    testPasswordLessCert(factory, defaultProps, "password_less");
#endif
```

Java runs it only for the PKCS12 keystore type — JKS cannot be password-less, `keytool` requires six characters.

Red/green verified through Ice's own engine with a rebuild between: `-25260` before, adapter activates after. iOS was checked on the 26.5 simulator, not a device.

## Notes for the reviewer

- The key must go in a `pkcs8ShroudedKeyBag`. With a plain `keyBag`, `SecPKCS12Import` returns `noErr` but no identity, and Java's `KeyStore` reports no aliases. Recorded in the script docstring so it is not retried.
- PBE-SHA1-3DES and a SHA-1 MAC are forced by Apple, the same reason `makecerts.sh` already passes `-legacy`. This is what the three CodeQL alerts flag; the script only writes public test fixtures.
- `certs/Dockerfile` gains `python3` and `python3-cryptography`. `pyrightconfig.json` has no `include` list, so it covers `certs/`.
- Existing certificates are untouched — only the two new files are added, so no CA, CRL, serial, or validity changes.


externl (commented):
I don't think we should ship empty-password support in this form. A user who creates a passwordless PKCS#12 the normal way — `openssl pkcs12 -export -passout pass:` — still can't load it on macOS after this PR: openssl writes the two-byte form, which macOS keeps rejecting. The only tool that produces a macOS-loadable file is this repo's generator script, and we can't point users at that. If it can't be done with openssl, it's not worth supporting.

One thing worth checking before scoping down: `params.passphrase` also accepts a `CFDataRef`. If a data passphrase reaches the KDF as raw bytes, retrying a failed empty-password import with the two `0x00` bytes would make openssl-generated files load on macOS — that would be real support, with no script. If that doesn't pan out, I'd keep just the `SecItemImport` passphrase fix and drop the script, the `*_null_password.p12` fixtures, and the macOS test branch, with #6489 closed as not-planned.

bernardnormier (commented):
I agree with the scoping objection: as-is, a passwordless PKCS#12 produced the normal way (`openssl pkcs12 -export -passout pass:`) still doesn't load on macOS, and the only tool that produces a loadable file is this repo's script.

However, the `CFDataRef` question has a positive answer. Measured on macOS with `SecItemImport` called exactly as `loadKeychainItems` calls it (`kSecFormatPKCS12`, import into a fresh keychain):

| fixture | `CFString ""` | `CFData {00 00}` | `CFData {}` (empty) | passphrase unset |
| --- | --- | --- | --- | --- |
| `client_password_less.p12` (openssl, B.1) | −25264 | noErr, 1 identity | −25264 | −25260 |
| `client_null_password.p12` (script, B.2) | noErr, 1 identity | −25264 | noErr, 1 identity | −25260 |

A `CFDataRef` passphrase reaches the key derivation as raw password bytes. So when `IceSSL.Password` is empty, importing with `CFString ""` and retrying an `errSecPkcs12VerifyFailure` (−25264) failure with a two-byte `CFData {0x00, 0x00}` makes macOS accept both empty-password encodings, like OpenSSL, Schannel, Java, and .NET — and in particular files produced by plain `openssl -passout pass:`.

With that retry, the generator script, the `*_null_password.p12` fixtures, the Dockerfile/README additions, and the per-platform test selection can all be dropped: every platform tests the existing `*_password_less.p12` fixtures, macOS included (iOS already reads the two-byte form and needs no change). The changelog entry stays and becomes accurate for openssl-produced files.

pepone:
Superseded by #6529 — closing rather than force-pushing, since almost nothing on this branch survives the change of direction.

@externl was right. This branch left `openssl pkcs12 -export -passout pass:` — the normal way to produce a password-less PKCS#12 — still failing on macOS, and the only file it could actually load was one our own generator script produced.

The `CFDataRef` route suggested in that review works. `params.passphrase` accepts a `CFData`, which reaches the key derivation verbatim rather than being formatted as a NULL-terminated BMPString, so retrying a failed empty-password import with the two `0x00` bytes of RFC 7292 appendix B.1 loads openssl-written files. Measured on macOS 26 with Ice’\s exact `SecItemImport` call:

| `params.passphrase` | `openssl -passout pass:` | .NET `Export(Pkcs12, null)` |
| --- | --- | --- |
| `NULL` (today) | -25260 | -25260 |
| `CFString ""` | -25264 MAC fail | loads |
| `CFData 00 00` | loads | -25264 MAC fail |

#6529 keeps the retry rather than dropping it, so macOS ends up reading both encodings, which is what OpenSSL and Java already do. It is ~10 lines in `SecureTransportUtil.cpp` plus a test on the existing `*_password_less.p12` fixtures, and it drops the generator script, the extra fixtures, the Dockerfile python dependency and the cross-language test churn entirely — along with the CodeQL, Pyright and OpenRewrite failures that came with them.

Issue #6448, opened 2026-08-06
IceRuby: version helpers never return false, leaving their callers' error paths unreachable


externl:
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI dead-code audit and verified by re-reading the source on `main` (`82b79108e2b`).

## Summary

Two version helpers in `ruby/src/IceRuby/Util.cpp` can never report failure, leaving their callers' error paths unreachable.

- `setVersion` has a single unconditional `return true`.
- `getVersion` has `return false` statements at `:33` and `:41` placed immediately after `throw RubyException(...)`, so they are themselves dead; the function can only return `true` or throw.

Consequently `if (!setVersion<T>(obj, version)) { return Qnil; }` (`:55-58`) and `if (!getVersion<...>(p, v)) { return false; }` (`:392-395`) are permanently false. Knock-on: `IceRuby::getEncodingVersion` can never return false, so the trailing `return Qnil;` at `Proxy.cpp:437` is unreachable too.

## Fix

Either make the helpers `void` and drop the guards, or have them return false instead of throwing — whichever matches the intended contract.

**Severity:** Low.

InsertCreativityHere:
This was fixed by #6506.

Issue #6446, opened 2026-08-06
IceRuby: unreferenced callRubyVoid overloads, ProxyInfo::isA, createNumSeq and PrimitiveInfo()


externl:
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI dead-code audit and verified by re-reading the source on `main` (`82b79108e2b`).

## Summary

Unreferenced symbols in the IceRuby extension.

- **Three `callRubyVoid` overloads** (`Util.h`, the `RFV_1`/`RFV_2`/`RFV_4` forms). Only two `callRubyVoid` calls exist in the whole extension — `Operation.cpp:272` (arity 1) and `Util.cpp:418` (arity 4) — so the 2-, 3- and 5-argument forms are never instantiated. *Near-miss: the 7-argument `callRuby` looks equally dead but is live at `Util.cpp:318`.*
- **`ProxyInfo::isA`** (`Types.cpp:2393`) is non-virtual and its only calls are its own recursion; `ProxyInfo::validate` never calls it. The `ClassInfo::isA` overload at `:2228` is a different function and is live. IcePHP's equivalent **is** called (`Communicator.cpp`), and IcePy has no `isA` at all, so this is dead only in Ruby. Follow-on: `ProxyInfo::interfaces` then becomes write-only.
- **`createNumSeq`** (`Util.cpp:221`, declared `Util.h:92`) — declaration plus definition, no callers.
- **`PrimitiveInfo()`** default constructor (`Types.cpp:443`, `Types.h:197`) — all eight instantiations use the `Kind` constructor, and this one leaves `kind` uninitialised.

**Severity:** Low.

InsertCreativityHere:
Fixed by #6506.

PR #6506, opened 2026-08-07
Remove Various Dead Code from IceRuby

InsertCreativityHere:
This PR implements #6448, changing IceRuby to take advantage of the fact the functions in it always return `true` or throw.
They never return `false`, which allows for some simplification.

It also removes most of the dead code found in #6446.
While working on it, I identified an additional dead field `ProxyInfo::interfaces`, but removing it affects public APIs and even the code-generator. Instead, I'll open another issue to remove this code for 3.9.0.

This PR also fixes a small bug caught along the way: `getEncodingVersion` could throw, but was not called within an `ICE_RUBY_TRY` section.

### DIFF
--- a/ruby/src/IceRuby/Proxy.cpp
+++ b/ruby/src/IceRuby/Proxy.cpp
@@ -423,17 +423,15 @@ IceRuby_ObjectPrx_ice_getEncodingVersion(VALUE self)
 extern "C" VALUE
 IceRuby_ObjectPrx_ice_encodingVersion(VALUE self, VALUE v)
 {
-    Ice::EncodingVersion val;
-    if (getEncodingVersion(v, val))
+    ICE_RUBY_TRY
     {
-        ICE_RUBY_TRY
-        {
-            Ice::ObjectPrx p = getProxy(self);
-            return createProxy(p->ice_encodingVersion(val), rb_class_of(self));
-        }
-        ICE_RUBY_CATCH
-    }
+        Ice::EncodingVersion val;
+        getEncodingVersion(v, val);
 
+        Ice::ObjectPrx p = getProxy(self);
+        return createProxy(p->ice_encodingVersion(val), rb_class_of(self));
+    }
+    ICE_RUBY_CATCH
     return Qnil;
 }
 
--- a/ruby/src/IceRuby/Util.h
+++ b/ruby/src/IceRuby/Util.h
@@ -136,9 +136,9 @@ namespace IceRuby
     VALUE createEncodingVersion(const Ice::EncodingVersion&);
 
     //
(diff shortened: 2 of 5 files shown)


ruby/src/IceRuby/Util.h
```diff
@@ -136,9 +130,9 @@ namespace IceRuby
     VALUE createEncodingVersion(const Ice::EncodingVersion&);
 
     //
-    // Extracts the members of an encoding version.
+    // Extracts the members of an encoding version. May raise RubyException.
```
InsertCreativityHere, line 133: This was already true, we just failed to document that it could throw - something we do for all other methods in this file that can throw.

ruby/src/IceRuby/Proxy.cpp
```diff
@@ -423,17 +423,15 @@ IceRuby_ObjectPrx_ice_getEncodingVersion(VALUE self)
 extern "C" VALUE
 IceRuby_ObjectPrx_ice_encodingVersion(VALUE self, VALUE v)
 {
-    Ice::EncodingVersion val;
-    if (getEncodingVersion(v, val))
```
InsertCreativityHere, line 427: The bug that was fixed, `getEncodingVersion` could throw, but was outside the `TRY/CATCH`.

pepone (approved):
The removals are correct, and the `ICE_RUBY_TRY` fix is a better catch than the description suggests: on `main`,
`ice_encodingVersion(42)` aborts the interpreter; here it raises `TypeError`. Nothing blocking, two asks.

**1. Remove the `callRubyVoid` overloads too, so #6446 can close.**

**2. File the 3.9.0 issue, point the TODO at it, and fold `base` in.**

```cpp
        // TODO: base and interfaces are write-only; removing them is a 3.9.0 change, see #NNNN.
        const ProxyInfoPtr base;
        const ProxyInfoList interfaces;
```

(`// TODO:` with the colon, per `RubySliceLoader.h:26` and `Types.h:29`.)

`base` is worse off than `interfaces`: slice2rb hardcodes `nil` for it (`RubyUtil.cpp:343`), all 39 `defineProxy`
calls pass `nil`, and `define()` only assigns when non-nil — so it's permanently null, never read. And the real
constraint is generated-code compatibility rather than public API: arity 3 is baked into every emitted `.rb`, which
blocks dropping the *parameters*, not the fields. Both fields could go today with `defineProxy` ignoring `b` and `i`.

InsertCreativityHere:
I removed everything to do with the 2 unused fields except for their parameters.
See: #6543 which pins their removal for 3.9.0, where it'll be okay to break from the generated code.

PR #6508, opened 2026-08-07
Fix Python ice_invokeAsync result for oneway, datagram, and batch proxies

### LINKED ISSUE #6438: Python: ice_invokeAsync completes with None instead of (True, b'') for oneway, datagram, and batch proxies
`Ice.ObjectPrx.ice_invokeAsync` completes its awaitable with `None` for oneway, datagram, batch oneway, and batch datagram proxies, instead of the `(True, b"")` result that every other mapping produces — and that the method's own type annotation promises.

### Reproducer

```python
import Ice

with Ice.initialize() as communicator:
    adapter = communicator.createObjectAdapterWithEndpoints("Probe", "tcp -h 127.0.0.1")
    adapter.activate()
    base = adapter.add(Ice.Object(), Ice.stringToIdentity("probe"))

    ow = base.ice_oneway()
    print(ow.ice_invoke("ice_ping", Ice.OperationMode.Normal, b""))                  # (True, b'')
    print(ow.ice_invokeAsync("ice_ping", Ice.OperationMode.Normal, b"").result())    # None
```

All four non-twoway proxy kinds are affected (the udp cases with an `udp -h 127.0.0.1` adapter):

```
twoway         sync : (True, b'\x06\x00\x00\x00\x01\x01')
twoway         async: (True, b'\x06\x00\x00\x00\x01\x01')
oneway         sync : (True, b'')
oneway         async: None
batchOneway    sync : (True, b'')
batchOneway    async: None
datagram       sync : (True, b'')
datagram       async: None
batchDatagram  sync : (True, b'')
batchDatagram  async: None
```

### Cause

The C++ callback-based `ice_invokeAsync` intentionally does not invoke the response callback for oneway, datagram, or batch proxies — a successful invocation completes through the sent callback, or through nothing at all for a batch (see the `@remark` on `ice_invokeAsync` in `cpp/include/Ice/Proxy.h`). Each mapping is therefore responsible for synthesizing the result at the future/awaitable boundary. IcePy doesn't:

- `python/modules/IcePy/Operation.cpp:1574` — `AsyncInvocation::sent`: `if (!_twoway) { ... set_result(Py_None); }`
- `python/modules/IcePy/Operation.cpp:1453` — `AsyncInvocation::invoke`: batch oneway/datagram complete immediately with `set_result(Py_None)`

`AsyncBlobjectInvocation::handleResponse`, which builds the `(bool, bytes)` tuple, only runs on the twoway response path.

The other mappings all perform the adaptation:

- C++ (future overload): `cpp/src/Ice/ProxyAsync.cpp:226-231` synthesizes `{true, {}}`
- Java: `ProxyIceInvoke.java:78-86` — `Ice_invokeResult(true, new byte[0])`
- Swift: `swift/src/Ice/Proxy.swift:443-468` — `(true, Data())` in both the batch and the non-twoway sent path

Synchronous `ice_invoke` is unaffected: it goes through the C++ synchronous path and correctly returns `(True, b'')`.

### Impact

The declared type is wrong in three places, all of which say the awaitable holds a `(bool, bytes)` tuple:

- `python/python/Ice/ObjectPrx.py` (the redeclaration added by #6427)
- `python/python/IcePy-stubs/__init__.pyi:583`
- the `PyDoc_STR` signature at `python/modules/IcePy/Proxy.cpp:1430`

A caller that writes the natural `ok, out = await p.ice_invokeAsync(...)` gets a `TypeError` on any non-twoway proxy. `python/test/Ice/blobject/RouterI.py:22-31` works around it today by routing the oneway forwarding case through synchronous `ice_invoke`.

### Suggested fix

Complete the future with `(True, b'')` rather than `None` on the non-twoway and batch paths of `AsyncBlobjectInvocation`.

Note that the `set_result(Py_None)` calls live in the shared `AsyncInvocation` base, which `AsyncTypedInvocation` also uses — and there `None` is the correct result (a oneway `ice_pingAsync` should yield `None`). So the fix needs a hook the blobject subclass can override, not a change to the base behavior.

Worth adding coverage in `python/test/Ice/blobject` at the same time: it currently exercises `ice_invokeAsync` only on a twoway proxy.

Once fixed, the `ice_invokeAsync` doc comment added in #6427 — "When this proxy is a oneway, datagram, or batch proxy, the flag is always `True` and the encapsulation is empty" — becomes accurate; it matches the Java and Swift wording and is deliberately being kept as-is.

bernardnormier:
`ObjectPrx.ice_invokeAsync` completed its future with `None` for oneway, datagram, batch oneway, and batch datagram proxies, instead of the `(True, b"")` result that the synchronous `ice_invoke`, the method's type annotation, and the other language mappings all produce.

The affected completion paths live in the shared `AsyncInvocation` base class, where `None` is the correct result for typed invocations (a oneway `ice_pingAsync` yields `None`). This PR makes the completion value a virtual hook: the base class still returns `None`, and `AsyncBlobjectInvocation` overrides it to return `(True, b"")`.

It also removes the workaround in the `Ice/blobject` test router, which forwarded oneway requests through the synchronous `ice_invoke` because of this bug. The oneway forwarding path now exercises `ice_invokeAsync`, and the client additionally checks the result of `ice_invokeAsync` on oneway and batch oneway proxies.

Fixes #6438.


externl (approved):
Looks good!

pepone (approved):
Changes look good.

One suggestion: `noResponseResult()` can just `return Py_None;`. Since 3.12 `Py_None` is immortal, so `Py_NewRef` is no longer required — `AsyncTypedInvocation::handleResponse` already relies on that one screen up, with `r = PyObjectHandle{Py_None};`.

Minor, in `RouterI.py`: `BlobjectI.ice_invoke` (L112-119) still has the shape you just removed from the async blobject — both branches call the same `ice_invoke`, and `except Ice.Exception: raise` is a no-op.

```python
if "_fwd" in current.ctx and current.ctx["_fwd"] == "o":
    proxy = proxy.ice_oneway()
return proxy.ice_invoke(current.operation, current.mode, bytes, current.ctx)
```

bernardnormier:
Thanks — both applied in b189717424: `noResponseResult()` now returns `Py_None` directly, and `BlobjectI.ice_invoke` uses the simplified shape you suggested. The `Ice/blobject` test passes in both configurations.

PR #6509, opened 2026-08-07
Remove obsolete Ruby compatibility shims

### LINKED ISSUE #6447: IceRuby: compatibility shims for Ruby versions below the supported minimum
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI dead-code audit and verified by re-reading the source on `main` (`82b79108e2b`).

## Summary

`ruby/ice.gemspec:33` requires Ruby >= 3.4.0, so several compatibility shims can no longer be taken.

- **`Types.cpp:17-25`** — a `T_ZOMBIE`/`RHASH_SIZE` fallback whose own comment says it is "required for RHASH_SIZE to work properly with Ruby 1.8.x". Both are unconditionally defined by modern Ruby headers. *Keep the real `RHASH_SIZE` uses; only the fallback block is dead.*
- **`Util.cpp:282`, `:305-311`** — the `#else` arms of `RUBY_BLOCK_CALL_FUNC_TAKES_BLOCKARG` and `RB_BLOCK_CALL_FUNC_STRICT`, both annotated "Defined Ruby >= 2.1". Removing them lets `ICE_RUBY_HASH_FOREACH_CALLBACK` collapse to plain `rb_block_call_func_t`.
- **`Config.h`** — an `#ifdef _WIN32` undef block; `extconf.rb` exits with "MinGW is not supported with Ice for Ruby", there is no Ruby vcxproj, and `BUILDING.md` documents Linux/macOS only.

## Fix

Confirm the minimum supported Ruby, then drop the branches that predate it.

**Severity:** Low.

bernardnormier:
Ice for Ruby requires Ruby >= 3.4 and builds only on Linux and macOS, so several old compatibility shims are dead code:

- `Types.cpp`: the `T_ZOMBIE`/`RHASH_SIZE` fallback for Ruby 1.8/1.9 — both are unconditionally defined by the headers of all supported Ruby versions.
- `Util.cpp`: the pre-2.1 arms of the `RUBY_BLOCK_CALL_FUNC_TAKES_BLOCKARG` and `RB_BLOCK_CALL_FUNC_STRICT` conditionals. The hash-foreach callback now matches `rb_block_call_func_t` exactly, so the `ICE_RUBY_HASH_FOREACH_CALLBACK` typedef and the cast at the call site are gone as well.
- `Config.h`: the `_WIN32` macro-undef block (`extconf.rb` rejects MinGW builds).
- `config/Make.rules`: the MinGW branch in the `ruby_libdir` computation.

Fixes #6447.

### DIFF
--- a/ruby/config/Make.rules
+++ b/ruby/config/Make.rules
@@ -30,7 +30,7 @@ endif
 
 # Ruby linker flags
 ruby_ldflags    := $(call ruby-call,$$(LIBRUBYARG))
-ruby_libdir     := $(call ruby-call,$(if $(findstring MINGW,$(shell uname)),$$(bindir),$$(libdir)))
+ruby_libdir     := $(call ruby-call,$$(libdir))
 ifneq ($(ruby_libdir),)
     ruby_ldflags := -L$(call ruby-call,$$(libdir)) $(ruby_ldflags)
 endif
--- a/ruby/config/Make.rules
+++ b/ruby/config/Make.rules
@@ -32,7 +32,7 @@ endif
 ruby_ldflags    := $(call ruby-call,$$(LIBRUBYARG))
 ruby_libdir     := $(call ruby-call,$$(libdir))
 ifneq ($(ruby_libdir),)
-    ruby_ldflags := -L$(call ruby-call,$$(libdir)) $(ruby_ldflags)
+    ruby_ldflags := -L$(ruby_libdir) $(ruby_ldflags)
 endif
 
 #
--- a/ruby/src/IceRuby/Config.h
+++ b/ruby/src/IceRuby/Config.h
@@ -55,18 +55,6 @@
 #    pragma GCC diagnostic pop
 #endif
 
-//
-// The Ruby header file win32/win32.h defines a number of macros for
-// functions like shutdown() and close() that wreak havoc.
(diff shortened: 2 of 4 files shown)


ruby/config/Make.rules
```diff
@@ -30,7 +30,7 @@ endif
 
 # Ruby linker flags
 ruby_ldflags    := $(call ruby-call,$$(LIBRUBYARG))
-ruby_libdir     := $(call ruby-call,$(if $(findstring MINGW,$(shell uname)),$$(bindir),$$(libdir)))
+ruby_libdir     := $(call ruby-call,$$(libdir))
 ifneq ($(ruby_libdir),)
     ruby_ldflags := -L$(call ruby-call,$$(libdir)) $(ruby_ldflags)
```
pepone: `ruby_libdir` is unconditionally `libdir` now, so this recomputes it — spelling it out was only necessary while `ruby_libdir` could be `bindir`. Also saves a redundant `ruby -e` subprocess.

```suggestion
    ruby_ldflags := -L$(ruby_libdir) $(ruby_ldflags)
```

pepone (approved):
Looks good, just one inline suggestion.

Issue #6510, opened 2026-08-07
DataStorm: remove the unused getInstanceCount chain


bernardnormier:
The DataStorm `getInstanceCount` chain is dead code:

- the pure virtual `DataReader::getInstanceCount` in `cpp/include/DataStorm/InternalI.h`
- the override in `cpp/src/DataStorm/DataElementI.h` and its definition in `DataElementI.cpp`
- the backing `_instanceCount{0}` field in `DataElementI.h`, which is never written, so the getter can only ever return 0

Applications cannot reach it either: `DataStorm::Reader` holds `_impl` privately and exposes no wrapper for this operation.

However, this cannot be removed in a 3.8.x patch release. `InternalI.h` is installed and included by the public `DataStorm.h`, and `getInstanceCount` sits in the middle of the `DataReader` vtable, ahead of methods the public `Reader` template calls inline (`getAllUnread`, `getNextUnread`, `onSamples`, ...). Removing it would shift their vtable slots and break applications built against earlier 3.8 headers when running with an updated library.

Remove the whole chain in 3.9.

Split out of #6491, which grouped this with unrelated IceGrid dead code; see that issue for the original audit details.

PR #6511, opened 2026-08-07
Remove dead code in IceGrid

### LINKED ISSUE #6491: IceGrid and DataStorm: undefined static members, unreferenced accessors and unreachable branches
> [!NOTE]

## Summary

Dead code in IceGrid and DataStorm. Grouped as one issue since these are all small and in adjacent components.

### Declared but never defined

- **`IceGrid/Database.h:220-224`** — five `static const std::string` members (`_applicationDbName`, `_objectDbName`, `_internalObjectDbName`, `_adapterDbName`, `_replicaGroupDbName`) that are **declared and defined nowhere in the tree**, so they are not even linkable. Vestige of the pre-LMDB (Freeze) storage layer; the real names live in an anonymous namespace in `Database.cpp:32-39`. `git log -S` last touches them in "Bug 3231 — alternative storage for IceGrid and IceStorm".
- **`IceGrid/DescriptorHelper.h:94`** — `CommunicatorHelper::printPropertySet` is declared with no definition anywhere. Its neighbour `printObjectAdapter` on the line above is defined and called.

### Unreferenced

- **`IceGrid/NodeI.{h:142,cpp:770}`** — private `getApplicationServers`, declaration plus definition, no callers. Its backing field `_serversByApplication` is used elsewhere, so only the accessor is dead.
- **`IceGrid/ServerAdapterI.h:41`** — `const std::string _replicaId` is not in the constructor's member-init list and is never read. Leftover from the old `<adapter name>.ReplicaId` support.
- **`IceGrid/AllocatableObjectCache.h:93`** — `_allocatablesByType` is declared and never touched; the adjacent `_types` map of `TypeEntry` serves its purpose.

### Unreachable

- **`IceGrid/AdminI.cpp:565`** — `return 0;` immediately after `throw NodeNotExistException(name);`. (The `return 0` at `:560` is in a different catch block and is live.)
- **`IceGrid/Database.cpp:1920-1923`** — the `!replace` branch of `addInternalObject`. The single call site, `NodeSessionI.cpp:34`, passes `true` explicitly, and `Database.h` is internal to the icegridnode/icegridregistry binaries, so the `bool = false` default at `Database.h:162` is never used either.
- **`DataStorm`** — the whole `getInstanceCount` chain: the pure virtual at `cpp/include/DataStorm/InternalI.h:182`, the override at `DataElementI.h:342`, the definition at `DataElementI.cpp:734`, and the `_instanceCount{0}` field at `DataElementI.h:382`. **`_instanceCount` is never written**, so the getter can only ever return 0, and `DataStorm::Reader` holds `_impl` privately and exposes no wrapper, so users cannot reach it either.

### Stale comment

- **`IceGrid/Database.cpp:1483`** — a `COMPILERFIX` note about `result.insert()` "not compiling on Sun". Sun Studio has not been supported since well before 3.7; the loop below it can fold back into a one-line `insert`.

**Severity:** Low.

bernardnormier:
This removes dead code from IceGrid:

- `Database.h`: five `static const std::string` DB-name members that were declared but defined nowhere (vestige of the pre-LMDB storage layer; the real names live in an anonymous namespace in `Database.cpp`).
- `Database.h`/`Database.cpp`/`NodeSessionI.cpp`: `addInternalObject` loses its `replace` parameter — the only caller passes `true`, so the `!replace` branch and the `bool = false` default were unreachable.
- `Database.cpp`: folded the `getAllAdapters` append loop back into the `result.insert(...)` one-liner it was standing in for, and removed the obsolete Sun Studio `COMPILERFIX` comment.
- `NodeI.h`/`NodeI.cpp`: removed the uncalled `getApplicationServers` accessor (its backing `_serversByApplication` field remains in use).
- `DescriptorHelper.h`: removed the `printPropertySet` declaration, which has no definition.
- `ServerAdapterI.h`: removed `_replicaId`, never initialized and never read.
- `AllocatableObjectCache.h`: removed the never-touched `_allocatablesByType` map.
- `AdminI.cpp`: removed two unreachable `return 0;` statements after `throw` in `getNodeProcessorSocketCount` (the second was not flagged by the issue; a sweep of `cpp/src/IceGrid` found no other statement-after-throw occurrences).

The DataStorm `getInstanceCount` item from the issue is split out to #6510: `InternalI.h` ships via the public `DataStorm.h` and the method sits mid-vtable, so its removal is not ABI-compatible and is deferred to 3.9.

Fixes #6491.


pepone (approved):
LGTM.

A few more of the same kind, all non-blocking:

**`DescriptorHelper.h:268` — `NodeHelper::instantiate`** is the same defect as the `printPropertySet` you removed, 174 lines below it in the same header: declared, never defined, never called. Not quite a one-liner though — dropping the declaration makes clang start reporting `-Wunused-private-field` for `_instantiated`, which is only read from `assert()` and so has no use under `-DNDEBUG`.

**`SessionServantManager.h:42` — `void connectionClosed(const Ice::ConnectionPtr&);`** declared, never defined, no callers. (`ReapThread.cpp:174` is `ReapThread`'s.)

**Unused `getCommunicator()` accessors** in `AllocatableObjectCache.h:69`, `ObjectCache.h:46` and `ServerCache.h:139`. Your call — `NodeCache`'s and `Database`'s are used, so these three are the unused members of a symmetric family.

**`Database.cpp` `getAllAdapters`**, while you're in there: the opening `result` / `ids` / `swap` dance is a vestige of the same era as the `COMPILERFIX` comment. `Cache::getAll` returns by value, so it collapses to `vector<string> result = _adapterCache.getAll(expression);`.

bernardnormier:
Thanks — all four applied in the latest commit. For `NodeHelper::instantiate`, I also removed `_instantiated` together with its constructor initializer and the asserts that were its only readers, to avoid the `-Wunused-private-field` warning you mentioned. The three `getCommunicator()` accessors are gone too; the underlying `_communicator` fields remain since they're still used within each cache's implementation.

PR #6512, opened 2026-08-07
Remove the unreachable Windows event-log registry code in Ice::Service

### LINKED ISSUE #6492: Ice::Service: the Windows event-log registry chain is unreachable
> [!NOTE]
> Windows-only: not compiled or verified on macOS/Linux. Needs a Windows CI run to confirm.

## Summary

The Windows event-log registry code in `cpp/src/Ice/Service.cpp` is unreachable, roughly 100 lines inside the `#ifdef _WIN32` block.

- **`SMEventLoggerI::addKeys` (`:177-248`)** and **`removeKeys` (`:250-262`)** have no callers anywhere.
- **`createKey` (`:400-408`)** is called only from inside those two functions (`:187`, `:257`), so it is transitively dead.
- **`setModuleHandle` (`:384`)** and the static **`HMODULE _module` (`:412`, defined `:415`)** — `_module` is read only at `:205-206`, inside the dead `addKeys`.

This is a vestige of the removed `Ice::Service` `--install`/`--uninstall` support.

## The public API question

The chain is still *fed* from outside: `Ice::Service::setModuleHandle` (`cpp/include/Ice/Service.h:105`, implemented `Service.cpp:823-826`) forwards to `SMEventLoggerI::setModuleHandle`, and `DLLMain.cpp:24` calls it on `DLL_PROCESS_ATTACH`. So that path still executes — it just stores a handle nothing reads.

Removing the internal chain is safe. Removing the public `Ice::Service::setModuleHandle` as well would be an API break and wants a separate decision.

**Severity:** Low.

bernardnormier:
Fixes #6492.

`SMEventLoggerI::addKeys` and `removeKeys` have had no callers since the `Ice::Service` `--install`/`--uninstall` support was removed, which transitively makes `createKey`, `SMEventLoggerI::setModuleHandle` and the static `_module` handle unreachable. The rest of `SMEventLoggerI` is live and untouched — it's still the logger used when running as a Windows service — and `mangleSource` stays, since the constructor uses it.

With `SMEventLoggerI::setModuleHandle` gone, the `@private` `Ice::Service::setModuleHandle` has nothing left to store, so it goes too, along with its only caller in `DLLMain.cpp`. That leaves `DLLMain.cpp` as pure `_CRT_INIT` plumbing for a custom DLL entry point that doesn't exist: a repo-wide search for `ice_DLL_Main`, `EntryPointSymbol` and `/ENTRY:` across every build file finds only the function's own definition, there is no `.def` file, and `EntryPointSymbol` last appeared in the VS2013-era projects. The linker has been using the default `_DllMainCRTStartup` all along, so `ice_DLL_Main` was compiled but never called. The file is removed, together with the exclusions the make, Python and Ruby builds carried for it.

No behavior change: `addKeys` is what would have registered `EventMessageFile` under `HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application\<source>`, and nothing has called it for years — so no changelog entry.


externl (approved):
Looks good!

pepone (approved):
LGTM. One nit:

`python/setup.py:106` — the comment above the win32 exclude list still says "Skip SysLoggerI, OpenSSL and SecureTransport on Windows" but the list also excludes `Service`. Worth syncing since this line was touched anyway.

The `setModuleHandle` / `ice_DLL_Main` removal is a source/ABI break for any out-of-tree Windows caller, but the API was `@private` and inert, so fine for the next major.

bernardnormier:
> so fine for the next major.

We plan to cherry-pick this fix onto the 3.8 branch. It's not "for the next major release".

PR #6513, opened 2026-08-07
Remove Dead Java Code

### LINKED ISSUE #6494: Java: unused typeIdToClass, always-false amd parameter, unwired IceGridGUI fields and dead ProGuard rules
> [!NOTE]

## Summary

Dead code in the Java mapping. Note Java's compiler does not warn about unused methods and the project's OpenRewrite lints are style-only, so none of this is caught automatically.

- **`Ice/Util.java:363`** — `public static String typeIdToClass(String id)`. Its last caller was removed by #3872 during the SliceLoader refactor; the replacement inlines the same conversion at `DefaultPackageSliceLoader.java:41`. (The `_typeIdToClass` fields in the SliceLoader classes are unrelated private caches and are live.) This is public API, so removal is a 3.9-window decision.
- **`Ice/CollocatedRequestHandler.java:259-284`** — the `amd` parameter of the private `dispatchException`/`handleException` is `false` at every call site (`:172`, `:200`, `:211`, and `:260` which forwards its own always-false value), so the `if (amd)` branch and its `invokeCompletedAsync()` call are unreachable. **This is a Java-specific divergence:** the C# equivalent passes a real variable. `sendResponse`'s own `amd` parameter *is* live — only the two exception methods are affected.
- **`IceGridGUI/.../Application/AdapterEditor.java:554-555`** — `_currentStatus` and `_currentEndpoints` are constructed and never touched: no reads, no `setText`, never added to a layout. The `LiveDeployment` sibling declares identically-named fields that *are* used; these are a copy-paste leftover.
- **`IceGridGUI/icegridgui.pro`** — three dead ProGuard rules: a `-keep` for `com.zeroc.IceSSL.**` (`:92-94`), a package that does not exist (SSL is `com.zeroc.Ice.SSL`, already covered by the `com.zeroc.Ice.**` keep at `:88`), and two `-dontnote` rules for `com.zeroc.Ice.IceMX.*` (`:51-52`), a package that has never existed — the real one is `com.zeroc.IceMX`.

**Severity:** Low.

InsertCreativityHere:
This PR fixes #6494: it just removes various dead things from our Java code.

It also adds a couple extra rules to our linting setup, that helped catch even more dead code than what the issue flagged.

### DIFF
--- a/java/src/IceGridGUI/icegridgui.pro
+++ b/java/src/IceGridGUI/icegridgui.pro
@@ -48,8 +48,8 @@
 -dontnote com.zeroc.Ice.MetricsMap
 -dontnote com.zeroc.Ice.ObjectPrx
 -dontnote com.zeroc.Ice.PluginManagerI
--dontnote com.zeroc.Ice.IceMX.Observer
--dontnote com.zeroc.Ice.IceMX.ObserverFactory
+-dontnote com.zeroc.IceMX.Observer
+-dontnote com.zeroc.IceMX.ObserverFactory
 
 -dontnote com.zeroc.IceBox.ServiceManagerI
 -dontnote com.zeroc.IceGridGUI.Coordinator
@@ -89,9 +89,7 @@
 -keep public class com.zeroc.IceGrid.** {
   public *;
 }
--keep public class com.zeroc.IceSSL.** {
-  public *;
-}
+
 -keep interface com.zeroc.IceGrid.**
 -keep class com.zeroc.IceMX.**
 -keep interface com.zeroc.IceMX.**
--- a/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Application/CommunicatorChildEditor.java
+++ b/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Application/CommunicatorChildEditor.java
@@ -27,7 +27,6 @@ abstract class CommunicatorChildEditor extends Editor {
                 _target.destroy(); // just removes the child
 
                 try {
-                    // @SuppressWarnings("unchecked")
                     childList.tryAdd(descriptor);
                 } catch (UpdateFailedException e) {
                     // Restore ephemeral
@@ -48,7 +47,6 @@ abstract class CommunicatorChildEditor extends Editor {
                 }
 
                 // Success
-                // @SuppressWarnings("unchecked")
                 _target = childList.findChildWithDescriptor(descriptor);
                 root.updated();
                 if (refresh) {
--- a/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Application/PropertySetParent.java
+++ b/java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Application/PropertySetParent.java
@@ -7,7 +7,7 @@ import com.zeroc.IceGrid.PropertySetDescriptor;
 interface PropertySetParent {
     void tryAdd(String id, PropertySetDescriptor descriptor) throws UpdateFailedException;
 
-    void tryRename(String oldId, String oldUnresolveId, String newUnresolvedId)
+    void tryRename(String oldId, String oldUnresolvedId, String newUnresolvedId)
         throws UpdateFailedException;
 
     void insertPropertySet(PropertySet nps, boolean fireEvent) throws UpdateFailedException;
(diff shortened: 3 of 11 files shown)


java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/CollocatedRequestHandler.java
```diff
@@ -169,7 +169,7 @@ private void dispatchAll(OutputStream os, int requestId, int requestCount) {
                 try {
                     _adapter.incDirectCount();
                 } catch (ObjectAdapterDestroyedException ex) {
-                    handleException(ex, requestId, false);
```
InsertCreativityHere, line 172: The bools passed to these functions were always `false`. I just removed them and hardcoded the logic.

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Application/CommunicatorChildEditor.java
```diff
@@ -27,7 +27,6 @@ protected boolean applyUpdate(boolean refresh) {
                 _target.destroy(); // just removes the child
 
                 try {
```
InsertCreativityHere, line 29: Randomly looked through a couple IceGridGUI files and found these.

java/src/IceGridGUI/src/main/java/com/zeroc/IceGridGUI/Application/PropertySetParent.java
```diff
@@ -7,7 +7,7 @@
 interface PropertySetParent {
     void tryAdd(String id, PropertySetDescriptor descriptor) throws UpdateFailedException;
 
-    void tryRename(String oldId, String oldUnresolveId, String newUnresolvedId)
```
InsertCreativityHere, line 10: A typo in the parameter name here. Just randomly spotted while skimming.

java/test/src/main/java/test/Slice/escape/Client.java
```diff
@@ -78,6 +78,7 @@ public _assert _notify(
     // This section of the test is present to ensure that the C++ types are named correctly. It is
     // not expected to run.
     @SuppressWarnings({"unused", "null"})
```
InsertCreativityHere, line 80: This is the `escape` test, where we test all kinds of random variable names.
Shouldn't be linting them though.

externl (approved):

java/src/IceGridGUI/icegridgui.pro
```diff
@@ -48,8 +48,8 @@
 -dontnote com.zeroc.Ice.MetricsMap
 -dontnote com.zeroc.Ice.ObjectPrx
 -dontnote com.zeroc.Ice.PluginManagerI
--dontnote com.zeroc.Ice.IceMX.Observer
--dontnote com.zeroc.Ice.IceMX.ObserverFactory
+-dontnote com.zeroc.IceMX.Observer
+-dontnote com.zeroc.IceMX.ObserverFactory
```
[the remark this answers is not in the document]
InsertCreativityHere: It seems pretty clear to me that this was the intended behavior and we just made a typo.
If anyone has more specific knowledge, feel free to say so.
But I plan to keep my fix as-is, not remove this.

java/src/com.zeroc.ice/src/main/java/com/zeroc/Ice/Util.java
```diff
@@ -354,39 +354,6 @@ public static <T> InvocationFuture<T> getInvocationFuture(CompletableFuture<T> f
         return (InvocationFuture<T>) f;
     }
 
-    /**
-     * Translates a Slice type id to a Java class name.
-     *
-     * @param id The Slice type id, such as {@code ::Module::Type}.
-     * @return The equivalent Java class name, or null if the type id is malformed.
-     */
-    public static String typeIdToClass(String id) {
```
InsertCreativityHere, line 363: Technically, this is "public API", and removing it is API-breaking.

But, no one in their right mind is calling this function, and for all intents-and-purposes, it's internal, like everything else in `Util.java`.

pepone (approved):
Changes look good. The `rewrite` CI job is failing though — `rewriteDryRun` wants to change `test/Slice/escape/Client.java` (the new `RemoveUnusedLocalVariables` recipe doesn't honor the `CHECKSTYLE:OFF` marker, plus a `BlankLines` hunk on the `CHECKSTYLE:ON` line). Can you take a look?

Issue #6451, opened 2026-08-06
IcePHP: unused AutoReleaseString, dangling $Ice_Protocol_1_0 and stray unused includes


externl:
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI dead-code audit and verified by re-reading the source on `main` (`82b79108e2b`).

## Summary

Small unreferenced items in the PHP mapping.

- **`Util.h:127-139`** — class `AutoReleaseString` has no users anywhere; its siblings `AutoDestroy` and `AutoEfree` are both live.
- **`php/lib/Ice.php:15`** — `global $Ice_Protocol_1_0;` is the only occurrence of that name in the repository; nothing in `php/src` registers such a symbol. Contrast `$Ice_Encoding_1_0`/`$Ice_Encoding_1_1` two lines below, which are assigned at `:40-41`.
- **`Communicator.cpp`** — unused `<condition_variable>`, `<thread>` and `<valarray>` includes, left over from the old reaper thread now replaced by `IceInternal::Timer`.
- Minor strays: an unused local plus a duplicated `getInfo()` call (`Connection.cpp:204`), the unused `InvocationPtr` alias (`Operation.cpp:104`), a declared-and-unused `ParamInfoList::iterator` (`Operation.cpp:394`), an always-true sub-term (`Types.cpp:2599`), and a redundant `assert` after an equivalent guard (`Proxy.cpp:678`).

**Severity:** Low.

InsertCreativityHere:
Fixed by #6514.

Issue #6450, opened 2026-08-06
IcePHP: PHP 5-era Z_ADDREF_P and ZEND_MN fallbacks in Config.h


externl:
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI dead-code audit and verified by re-reading the source on `main` (`82b79108e2b`).

## Summary

`php/BUILDING.md:22` requires PHP 8.0 or higher, so two fallbacks in `php/src/Config.h` can never be taken.

- **`:88-93`** — the `#ifndef Z_ADDREF_P` block. `Z_ADDREF_P` is unconditionally defined by PHP 7.0+. Its inner `ZVAL_ADDREF` no longer exists in PHP 8 at all, so the `#error "Unknown PHP version"` beneath it is unreachable as well.
- **`:95-97`** — the `#ifndef ZEND_MN` fallback; `ZEND_MN` is unconditionally defined in `Zend/zend_API.h`. *Keep the three genuine `ZEND_MN` uses in `Connection.cpp`, `Proxy.cpp` and `Endpoint.cpp`.*

Note the neighbouring `ice_to_string_arginfo` version branch is **not** dead — with a PHP 8.0 minimum its `< 80200` arm is reachable.

**Severity:** Low.

InsertCreativityHere:
Fixed by #6514.

Issue #6449, opened 2026-08-06
IcePHP: Proxy connection fields are never assigned, so both destructor branches are dead


externl:
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI dead-code audit and verified by re-reading the source on `main` (`82b79108e2b`).

## Summary

`php/src/Proxy.cpp` declares two `zval*` members, `connection` (`:69`) and `cachedConnection` (`:70`), that are initialised to `0` in the constructor (`:1436-1437`) and **never assigned anywhere**. Both destructor branches (`:1447-1451`) are therefore permanently false, and removing the fields empties `~Proxy()` entirely.

This is a vestige of removed per-proxy connection caching: `ice_getConnection` (`:1271`) and `ice_getCachedConnection` (`:1296`) both call `createConnection(...)` fresh on every invocation.

**Severity:** Low.

InsertCreativityHere:
Fixed by #6514.

PR #6514, opened 2026-08-07
Removed Various Dead PHP Code

InsertCreativityHere:
This PR implements #6451, #6449, and #6450.
All of which were just about removing dead code from our PHP source.


php/lib/Ice.php
```diff
@@ -12,7 +12,6 @@
 global $IcePHP__t_double;
 global $IcePHP__t_string;
 
-global $Ice_Protocol_1_0;
```
InsertCreativityHere, line 15: Should of been removed when we removed all the public API for protocol versions, but was left behind.

php/src/Config.h
```diff
@@ -85,15 +85,4 @@ ZEND_END_ARG_INFO()
 #    define ICE_G(v) (ice_globals.v)
 #endif
```
InsertCreativityHere, line 87: Apparently, these two macros are always defined with newer Ruby versions like the ones we're targeting.

So the blocks are dead code.

php/src/Proxy.cpp
```diff
@@ -66,8 +65,6 @@ namespace IcePHP
         Ice::ObjectPrx proxy;
         ProxyInfoPtr info;
         CommunicatorInfoPtr communicator;
-        zval* connection;
-        zval* cachedConnection;
```
InsertCreativityHere, line 70: These two fields were never read from.

php/src/Types.cpp
```diff
@@ -2594,19 +2594,16 @@ IcePHP::ProxyInfo::getId() const
 bool
 IcePHP::ProxyInfo::validate(zval* zv, bool throwException)
 {
-    if (Z_TYPE_P(zv) != IS_NULL)
+    if (Z_TYPE_P(zv) != IS_NULL && (Z_TYPE_P(zv) != IS_OBJECT || Z_OBJCE_P(zv) != proxyClassEntry))
     {
-        if (Z_TYPE_P(zv) != IS_OBJECT || (Z_TYPE_P(zv) == IS_OBJECT && Z_OBJCE_P(zv) != proxyClassEntry))
```
InsertCreativityHere, line 2599: Not dead, but this `if` could be simplified since it contained both `!= IS_OBJECT` and `== IS_OBJECT` in different branches...

PR #6520, opened 2026-08-10
Replace hand-maintained BuildDependency edges with ProjectReference items

### LINKED ISSUE #6500: Declare C++ link dependencies with ProjectReference instead of hand-maintained BuildDependency edges
Follow-up to https://github.com/zeroc-ice/ice/pull/6469#discussion_r3736548332.

The C++ projects don't declare the libraries they link with. Linking works only because the public headers carry auto-link pragmas:

```cpp
// cpp/include/Ice/Config.h:52
#    pragma comment(lib, ICE_LIBNAME("Ice")) // Automatically link with Ice[D].lib
```

(same in `IceGrid/Config.h`, `Glacier2/Config.h`, `IceStorm/Config.h`, `IceBox/Config.h`, `DataStorm/Config.h`, `IceDiscovery/IceDiscovery.h`, `IceLocatorDiscovery/IceLocatorDiscovery.h`, and `cpp/src/IceStorm/Service.h`).

Because the pragma supplies the `.lib` but no build order, the ordering lives entirely in hand-maintained `BuildDependency` edges in the solution files:

| Solution | Projects | `BuildDependency` edges |
| --- | --- | --- |
| `cpp/msbuild/ice.slnx` | 32 | 59 |
| `cpp/msbuild/ice.test.slnx` | 213 | 216 |

Nothing verifies these edges. A new project that omits them still compiles and links on a warm tree, and only fails as an ordering flake on a clean parallel build. #6469 fixed one that was simply wrong — `IceGrid/session/verifier` declared a dependency on `IceGrid/allocation/verifier`, an unrelated test project, instead of on `testcommon` — and it had gone unnoticed.

## Proposal

Declare the dependencies in the project that has them, the way the `slice2*` compilers already do:

```xml
<ItemGroup>
  <ProjectReference Include="..\..\IceUtil\msbuild\iceutil\iceutil.vcxproj">
    <Project>{4d1a5110-3176-44ba-8bbb-57bf56519b9f}</Project>
  </ProjectReference>
  <ProjectReference Include="..\..\Slice\msbuild\slice.vcxproj">
    <Project>{57cd6ac2-0c9d-4648-9e9d-5df60c90f18a}</Project>
  </ProjectReference>
</ItemGroup>
```

`slice2rb` has relied on this alone all along, and #6469 removed the redundant `slice.vcxproj` edges from the other `slice2*` projects, so the pattern is already load-bearing. A project then carries its own dependencies and can't forget a solution edge it never had to write.

## Not every edge is a link edge

The `BuildDependency` graph cannot be dropped wholesale. Of the 59 edges in `ice.slnx`:

- **44** point at a library — genuine link dependencies, convertible as-is.
- **14** point at `slice2cpp.vcxproj`, which is an `Application`. These order Slice compilation, not linking. A plain `ProjectReference` would try to link the compiler's output, so they need `<LinkLibraryDependencies>false</LinkLibraryDependencies>` and `<ReferenceOutputAssembly>false</ReferenceOutputAssembly>`, or they stay as `BuildDependency`.
- **1** is `icegridadmin` → `icebox.vcxproj`, also an `Application`. `icegridadmin` can't link `icebox.exe`; this looks like it was meant to be `iceboxlib`, which `icegridadmin` does not currently list. Worth confirming while converting.

## Scope

- `cpp/msbuild/ice.slnx` + the ~23 projects under `cpp/src/**/msbuild/`.
- `cpp/msbuild/ice.test.slnx` + the 213 test projects, none of which use `ProjectReference` today. This is the bulk of the work and the part that benefits most, since test projects are added routinely.

Verification needs a clean-tree parallel build on Windows (`msbuild /m ice.proj /t:BuildDist`, and the same for the test solution) — the failure mode this guards against only shows up when the tree is cold and the graph is walked in parallel.

pepone:
Fixes #6500.

Converts all 59 `BuildDependency` edges in `cpp/msbuild/ice.slnx` and all 218 in `cpp/msbuild/ice.test.slnx` into GUID-free `ProjectReference` items declared by each `.vcxproj`, so a project carries its own build dependencies instead of relying on solution edges nothing verifies. The solution files now contain only plain project listings.

- The 14 edges to `slice2cpp.vcxproj` (an Application) order Slice compilation rather than linking, so those references set `LinkLibraryDependencies=false` and `ReferenceOutputAssembly=false`.
- `icegridadmin` declared a dependency on `icebox.vcxproj`, an executable it cannot link. It now references `iceboxlib.vcxproj`, matching the `IceBox/IceBox.h` include in `IceGrid/Parser.cpp` whose auto-link pragma pulls in `IceBox[D].lib`.
- Every converted test edge was checked against the sources before conversion: they are all genuine link dependencies established by auto-link pragmas (`test/include/TestHelper.h` for `testcommon`, plus the `Ice/library` `alltests`/`consumer`/`gencode` chain and the `Ice/background` `testtransport` plugin, which the clients link directly).
- The existing `slice2*` references drop their redundant `<Project>` GUID metadata; references resolve by path.

Verified on Windows with cold-tree parallel builds of both solutions (`msbuild /m ice.proj /t:BuildDist` and `msbuild /m ice.test.slnx`, x64 Release) from a fresh checkout — the failure mode this guards against only shows up when the tree is cold and the graph is walked in parallel. Both builds completed with zero warnings.

### DIFF
--- a/cpp/msbuild/ice.slnx
+++ b/cpp/msbuild/ice.slnx
@@ -3,110 +3,29 @@
     <Platform Name="Win32" />
     <Platform Name="x64" />
   </Configurations>
-  <Project Path="../src/DataStorm/msbuild/datastorm/datastorm.vcxproj">
-    <BuildDependency Project="../src/Ice/msbuild/ice/ice.vcxproj" />
-    <BuildDependency Project="../src/slice2cpp/msbuild/slice2cpp.vcxproj" />
-  </Project>
-  <Project Path="../src/dsnode/msbuild/dsnode.vcxproj">
-    <BuildDependency Project="../src/DataStorm/msbuild/datastorm/datastorm.vcxproj" />
-  </Project>
-  <Project Path="../src/Glacier2/msbuild/glacier2router.vcxproj">
-    <BuildDependency Project="../src/Glacier2Lib/msbuild/glacier2/glacier2.vcxproj" />
-    <BuildDependency Project="../src/Ice/msbuild/ice/ice.vcxproj" />
-  </Project>
-  <Project Path="../src/Glacier2Lib/msbuild/glacier2/glacier2.vcxproj">
-    <BuildDependency Project="../src/Ice/msbuild/ice/ice.vcxproj" />
-    <BuildDependency Project="../src/slice2cpp/msbuild/slice2cpp.vcxproj" />
-  </Project>
-  <Project Path="../src/Ice/msbuild/ice/ice.vcxproj">
-    <BuildDependency Project="../src/slice2cpp/msbuild/slice2cpp.vcxproj" />
-  </Project>
-  <Project Path="../src/IceBox/msbuild/icebox/icebox.vcxproj">
-    <BuildDependency Project="../src/IceBox/msbuild/iceboxlib/iceboxlib.vcxproj" />
-  </Project>
-  <Project Path="../src/IceBox/msbuild/iceboxadmin/iceboxadmin.vcxproj">
-    <BuildDependency Project="../src/Ice/msbuild/ice/ice.vcxproj" />
-    <BuildDependency Project="../src/IceBox/msbuild/iceboxlib/iceboxlib.vcxproj" />
-  </Project>
-  <Project Path="../src/IceBox/msbuild/iceboxlib/iceboxlib.vcxproj">
-    <BuildDependency Project="../src/Ice/msbuild/ice/ice.vcxproj" />
-    <BuildDependency Project="../src/slice2cpp/msbuild/slice2cpp.vcxproj" />
-  </Project>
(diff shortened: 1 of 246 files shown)


bernardnormier (approved):
Verified the conversion mechanically: all 277 edges reappear as `ProjectReference` items in the right projects, no pre-existing reference was lost, and every reference to an executable carries the no-link metadata. Also confirmed that `icegridadmin`'s new reference to `iceboxlib` (replacing the old edge to the `icebox` executable) matches the source: `IceGrid/Parser.cpp` includes `IceBox/IceBox.h`, whose auto-link pragma pulls in `IceBox[D].lib`.

Please cherry-pick this to the 3.8 branch as usual (and move #6500 to a 3.8.x milestone) — otherwise every future change touching these files will conflict on the cherry-pick.

Issue #6389, opened 2026-08-04
DataStorm: a throwing application decoder should always be logged locally with context


bernardnormier:
When an application-provided decoder or updater throws, this typically corresponds to an encoder/decoder mismatch with a process running different code — a severe error, but a legitimate one that DataStorm must expect. DataStorm cannot report this error to the peer: the session protocol is fire-and-forget (`TopicI::forward` and `SessionI::attachTopic` invoke the listener session proxies with null response and exception callbacks, and `attachElementsAck` is oneway), and the peer could not act on the report anyway — resolving the mismatch requires redeploying an application.

The realistic mechanism is local logging, and it should be uniform: catch the throw at the DataStorm layer, log a warning that names the topic and the element involved, and degrade predictably.

Today this rule is applied in some paths but not others.

**Where we do this:**

- Sample value decoding and a throwing updater on the read path: `DataElementI` catches, logs `dropped sample <id>: the value could not be decoded` / `the partial update could not be applied`, drops the sample, and lets the next full sample resynchronize the key.
- Update tag decoding in `SessionI::attachTags` (after #6343): the undecodable tag is skipped with a warning and the tags that decode are kept.

**Where we don't:**

- Key decoding in `SessionI::subscriberInitialized` (the `attachElementsAck` path): the exception escapes the oneway dispatch. The only diagnostic is a generic `Ice.Warn.Dispatch` warning that names neither the topic nor the key, and the reader then quietly receives nothing from that writer. Staying uninitialized is the correct behavior (see #6322) — the problem is that nothing actionable is logged.
- Key decoding in `SessionI::initSamples`: same escape through the dispatch, same generic warning.

These two paths should catch around the key decoding and log a contextual warning like the paths above do, while keeping the reader uninitialized as today.

Related: #6213 tracks what should happen when a partial update carries a tag that was skipped.

pepone:
All application-decoder call sites now catch and log a contextual warning:

- Sample value decoding and throwing updaters on the read path (pre-existing)
- Update tag decoding in `attachTags` (#6343) and `attachTopic` (#6507)
- Key decoding in `initSamples` and `subscriberInitialized` (#6522)
- The nine element-attachment sites — keys, key filters, and sample filter criteria in `TopicI`/`DataElementI`, plus the live inline-key decode in `SubscriberSessionI::s` (#6541)

A throwing sample filter *predicate* remains tracked separately in #6596.

Issue #6507, opened 2026-08-07
DataStorm: an update tag that fails to decode during attachTopic hangs both nodes


pepone:
## Summary

A subscriber decodes the publisher's update tags with the application's tag factory. On the `attachTags` path
that decode is guarded (#6343); on the `attachTopic` path it is not. When the tags arrive in the initial topic
spec instead of a later `attachTags` call, a throwing decoder escapes the attach dispatch: the topic never
finishes attaching, the oneway ack means the peer is never told, and both nodes wait forever — the writer in
`waitForReaders`, the reader in `getNextUnread`.

## Mechanism

`SessionI::attachTopic` decodes every tag carried in the peer's topic spec, unguarded:

https://github.com/zeroc-ice/ice/blob/87c74740720fb5f2bc154b2fc7250c819bb031f6/cpp/src/DataStorm/SessionI.cpp#L218-L227

The sibling decode in `SessionI::attachTags` was guarded by #6343, which treats a tag it cannot decode as
skipped and logs a warning:

https://github.com/zeroc-ice/ice/blob/87c74740720fb5f2bc154b2fc7250c819bb031f6/cpp/src/DataStorm/SessionI.cpp#L317-L337

Which of the two sees the tags depends on the attach sequence. When the peer's topic is already known, the tags
arrive through `attachTags` and are skipped cleanly. When they ride the initial spec, they reach the unguarded
loop in `attachTopic` and the throw aborts the whole dispatch — no element of the topic is attached.

The unguarded decode dates to the DataStorm import (`8e3d0b07ec`). #6343 did not touch `attachTopic`; it guarded
the other site and added the test that exposes this one.

## Evidence

Observed in CI, not by inspection: `release on ubuntu-24.04-arm` in
[run 31189830839](https://github.com/zeroc-ice/ice/actions/runs/31189830839/job/92903605621), `DataStorm/partial`,
`Writer/Reader multicast enabled` case, hanging in `testing update tag that fails to decode...` until the 30s
wait expired.

The writer's coredump puts it in the reader barrier, so the reader was never registered:

```
#5  DataStormI::DataElementI::waitForListeners (this=..., count=1)
#6  DataStorm::Writer<std::string, Counter, Op>::waitForReaders (count=1) at DataStorm.h:1286
#7  Writer::run (...) at test/DataStorm/partial/Writer.cpp:513
```

The reader's log names the escaping exception and the dispatch it escaped:

```
[...]

PR #6522, opened 2026-08-10
Log a warning when a DataStorm key decoder throws during initialization

pepone:
Refs #6389 — left open for the remaining unguarded sites.

DataStorm catches a throwing application decoder and logs a contextual warning on the sample read path
(`DataElementI`) and when decoding update tags (`SessionI::attachTags`, since #6343), but not when decoding a key
during a session's initialization. There the exception escaped a oneway dispatch, so the only diagnostic was a
generic `Ice.Warn.Dispatch` message naming neither the topic nor the element, and the reader then quietly received
nothing from that writer.

`SessionI::initSamples` and `SessionI::subscriberInitialized` now catch around the key decoding and log
`discarded the initialization samples for '<element>': the key could not be decoded:` followed by `what()`. The
element renders through `DataElementI::toString`, which already ends in `@<topicId>:<topicName>`, so one token names
both the element and the topic — `attachTags` spells the topic out separately only because a tag id carries no
topic.

The affected reader stays uninitialized, as #6322 established. What changes is the blast radius: the throw used to
abort the whole dispatch, taking the sibling elements with it.

- In `initSamples`, one undecodable key abandoned the outer `initializationBatches` loop, so the batches addressed
  to the *other* readers in the same request were never applied.
- In `subscriberInitialized`, the throw unwound through `DataElementI::attach` into `TopicI::attachElementsAck`,
  discarding every `initCb` accumulated so far — those run only after the attach loop completes — and skipping the
  `detachElementsAsync` for `removedIds` back in `SessionI::attachElementsAck`.

Each site now abandons only its own element.

## Notes

The inline-key decode is reachable only against an **any-key or filtered writer**: `toSample` marshals the key
inline exactly when the writer element has no keys, so `keyId == 0` implies such a peer.

In `subscriberInitialized` the decode is reached only for a **local reader**. A peer *reader* element acks with
empty `samples` — the base `DataElementI::getSamples` returns `{}` and only `KeyDataWriterI::getSamples` yields
samples — so the early `samples.empty()` return fires first. The starved siblings are therefore the other local
readers in that ack, not the peer's readers.

`initSamples` is guarded by `if (q->second.initialized) return;`, so leaving it there is exactly "still
uninitialized". `subscriberInitialized` has no such guard, so its comment says the subscriber keeps its previous
state.

No test. The logging itself is not assertable, and the containment is only observable through batch/ack ordering,
which follows a `std::map` keyed by `shared_ptr` — a harness built on that would fail before the fix only some of
the time, and by hanging.

Ten sites still run an application decoder unguarded inside a dispatch, and would each need a caller-specific
degradation decision rather than a blanket catch: `SessionI.cpp:225` (#6507), `SessionI.cpp:1776` (the live `s()`
key decode), `TopicI.cpp:167/224/366/417/476/544`, and `DataElementI.cpp:110/164`. #6389 stays open for them. All
four `sample->decode` value sites are already caught.

Verified on macOS: `make -C cpp srcs` and `tests` clean, 12/12 `cpp/DataStorm` suites pass.

### DIFF
--- a/changelog.d/datastorm/6389.md
+++ b/changelog.d/datastorm/6389.md
@@ -1,0 +1,3 @@
+- Fixed a bug where a key that the application's `Decoder` rejects during a session's initialization also left the
+  other readers of that session uninitialized. Only the reader the samples are addressed to stays uninitialized now,
+  and DataStorm logs a warning naming that reader instead of letting the failure escape as a generic dispatch error.
--- a/changelog.d/datastorm/6389.md
+++ b/changelog.d/datastorm/6389.md
@@ -1,3 +1,0 @@
-- Fixed a bug where a key that the application's `Decoder` rejects during a session's initialization also left the
-  other readers of that session uninitialized. Only the reader the samples are addressed to stays uninitialized now,
-  and DataStorm logs a warning naming that reader instead of letting the failure escape as a generic dispatch error.
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -614,7 +614,22 @@ SessionI::initSamples(int64_t topicId, int64_t peerTopicId, DataSamplesSeq initi
                     }
                     else
                     {
-                        key = topic->getKeyFactory()->decode(_instance->getCommunicator(), sample.keyValue);
+                        try
+                        {
+                            key = topic->getKeyFactory()->decode(_instance->getCommunicator(), sample.keyValue);
+                        }
+                        catch (const std::exception& ex)
+                        {
+                            // The key factory runs the application's decoder. Abandon the batch and leave the reader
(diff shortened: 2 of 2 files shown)


externl (approved):

bernardnormier (commented):
Verified both containment claims in context: the `initSamples` catch returns from the per-batch lambda, and the empty return from `subscriberInitialized` leaves the rest of the ack unaffected.

Two asks:

1. Please drop the changelog fragment: a throwing `Decoder` is an application bug, so a working application never observes this change — same disposition as the earlier decoder-handling fixes, which merged without an entry.
2. Please change "Fixes #6389" to a non-closing reference — let's keep #6389 open for the remaining unguarded sites rather than file a new issue.

bernardnormier (approved):

PR #6526, opened 2026-08-10
Skip an undecodable update tag carried in a DataStorm topic spec

### LINKED ISSUE #6507: DataStorm: an update tag that fails to decode during attachTopic hangs both nodes

pepone:
Closes #6507.

`SessionI::attachTopic` decoded the peer's update tags with the application's tag factory without a guard, so a
decoder that threw aborted the whole attach dispatch: no element of the topic was attached, the oneway ack meant the
peer was never told, and both nodes waited forever — the writer in `waitForReaders`, the reader in `getNextUnread`.

The sibling decode in `attachTags` was already guarded, so whether an undecodable tag was skipped or deadlocked the
pair depended only on whether the tag rode the initial topic spec or a later `attachTags` call. This moves the
guarded decode into a `decodeTags` helper and calls it from both sites, so the two cannot diverge again.

The `partial` test now creates the writer topic and its update tags ahead of the reader topic, so the tags reliably
reach the reader in the topic spec instead of racing between the two paths. Against the unpatched library it hangs at
`testing update tag that fails to decode...` with the reported `failed to dispatch attachTopic` warning; with the fix
it passes, as do all 12 DataStorm suites.

### DIFF
--- a/cpp/src/DataStorm/SessionI.cpp
+++ b/cpp/src/DataStorm/SessionI.cpp
@@ -218,12 +218,7 @@ SessionI::attachTopic(TopicSpec spec, const Current&)
                 // If the topic spec has tags, decode them and add them to the subscriber.
                 if (!spec.tags.empty())
                 {
-                    auto& subscriber = _topics.at(spec.id).getSubscriber(topic);
-                    for (const auto& tag : spec.tags)
-                    {
-                        subscriber.tags[tag.id] =
-                            topic->getTagFactory()->decode(_instance->getCommunicator(), tag.value);
-                    }
+                    decodeTags(topic, spec.tags, _topics.at(spec.id).getSubscriber(topic).tags);
                 }
 
                 // Provide the local tags to the remote topic by calling attachTagsAsync.
@@ -318,24 +313,7 @@ SessionI::attachTags(int64_t topicId, ElementInfoSeq tags, bool initialize, cons
             // every tag is decoded; the subscriber keeps its current tags until then. Any other call adds to the tags
             // the subscriber already holds, and decodes into them directly.
             map<int64_t, shared_ptr<Tag>> replacement;
-            auto& decoded = initialize ? replacement : subscriber.tags;
-            for (const auto& tag : tags)
-            {
-                try
-                {
-                    decoded[tag.id] = topic->getTagFactory()->decode(_instance->getCommunicator(), tag.value);
-                }
-                catch (const std::exception& ex)
-                {
-                    // The tag factory runs the application's decoder. Skip a tag it can't decode: the peer resends
-                    // its tags on reconnect; until then a partial update carrying the skipped tag resolves to the
-                    // key's current value.
-                    Warning out(_traceLevels->logger);
-                    out << "skipped update tag " << tag.id << " on topic '" << topic << "': the tag could not be "
-                        << "decoded:\n"
-                        << ex.what();
-                }
-            }
+            decodeTags(topic, tags, initialize ? replacement : subscriber.tags);
 
             if (initialize)
             {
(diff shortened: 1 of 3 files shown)


bernardnormier (approved):
Verified both call sites keep their semantics through the helper (the initialize replacement-swap in `attachTags` picks the target map at the call site), and the test reordering makes the topic-spec path deterministic.

Please drop the changelog fragment — same disposition as the previous decoder-handling fixes, which merged without an entry: a decoder only throws on a severe deployment error (mismatched definitions between the two sides) or data corruption, neither of which is part of normal operations.

externl (approved):
Looks good.

PR #6527, opened 2026-08-10
Remove dead code from the Ice C# assembly

### LINKED ISSUE #6495: C#: unreferenced getBool and isLinux, unused BzAction.Flush and an always-true null check
> [!NOTE]

## Summary

The C# assemblies are the cleanest mapping: `AnalysisMode=All` with `TreatWarningsAsErrors` means every unused `private`/`internal` member already fails the build, and there is no conditional compilation at all. What remains is in the analyzer blind spots — `public` members of documented-internal namespaces, unused enum members, and always-true conditions.

- **`Ice/Internal/ByteBuffer.cs:195`** — `getBool()` has no callers; `InputStream.readBool()` reads `_buf.b.get() == 1` directly. Its siblings `putBool`, `getBoolSeq` and `putBoolSeq` are all live. `public` on a `public` class, but in the docfx-excluded `Ice.Internal` namespace.
- **`Ice/Internal/AssemblyUtil.cs:13`** — `isLinux` is never read. Both siblings are live: `isWindows` at `ConnectionI.cs`, `Network.cs`, `BZip2.cs`, `UdpTransceiver.cs`, and `isMacOS` at `BZip2.cs`.
- **`Ice/Internal/BZip2.cs:39`** — the `BzAction.Flush` enum member is unused; only `Run` (`:143`) and `Finish` (`:155`) are ever passed. *Do not touch the `BzStatus` negative members — they look unreferenced but are interpolated into exception messages, so their names are live at runtime.*
- **`Ice/ConnectionI.cs:945`** — a nested `if (startCB is not null)` inside a block already guarded by the same condition, with no intervening assignment. CA1508 flags it as always true; the body collapses to `++upcallCount;`.

**Severity:** Low.

pepone:
Fixes #6495

### DIFF
--- a/csharp/src/Ice/ConnectionI.cs
+++ b/csharp/src/Ice/ConnectionI.cs
@@ -942,10 +942,7 @@ public sealed class ConnectionI : Internal.EventHandler, CancellationHandler, Co
                         {
                             startCB = _startCallback;
                             _startCallback = null;
-                            if (startCB is not null)
-                            {
-                                ++upcallCount;
-                            }
+                            ++upcallCount;
                         }
                     }
                     else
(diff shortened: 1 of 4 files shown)


externl (approved):
Looks good.

PR #6537, opened 2026-08-10
Ignore an incoming Bluetooth connection delivered after the acceptor is closed

### LINKED ISSUE #6441: IceBT AcceptorI::newConnection lacks a closed guard
`IceBT::AcceptorI::newConnection` calls `ready(SocketOperationRead, true)` (`cpp/src/IceBT/AcceptorI.cpp:164`) with no guard against the acceptor having already been closed — the acceptor-side counterpart of #6377.

`AcceptorI::close()` (`AcceptorI.cpp:40-52`) only unregisters the D-Bus profile. `DBus::Connection::handleMessage` copies `_services` under `_mutex` and then dispatches outside it (`cpp/src/IceBT/DBus.cpp:1060-1071`), so a `NewConnection` invocation already in flight when `removeService()` runs still reaches `AcceptorI::newConnection`.

`IncomingConnectionFactory::finished` calls `closeAcceptor()` from the same `ThreadPool::finish` work item that then clears the ready callback (`cpp/src/Ice/ThreadPool.cpp:353-361`). The late `newConnection` therefore reaches `NativeInfo::ready()` with no callback installed: `assert(_readyCallback)` in debug builds, a call through a null `shared_ptr` in release.

Trigger: an object adapter is deactivated, or the communicator shut down, while a peer is connecting.

Second consequence of the same sequence: the late connection's fd is wrapped in a `TransceiverI` and pushed onto `_transceivers`, but `accept()` is never called for it. `~AcceptorI` is defaulted, so that transceiver is destroyed with a live fd — `assert(_fd == INVALID_SOCKET)` in `~StreamSocket` (`cpp/src/Ice/StreamSocket.cpp:68`), and a leaked fd in release builds. This applies to any transceiver left queued at teardown, not just this race.

A guard mirroring #6387 fixes the crash. Alternatively, making `NativeInfo::ready()` a no-op when `_readyCallback` is null closes the whole class at the core — if a cleared callback is legitimately reachable, the assert has no diagnostic value.

Noted while reviewing #6387.

bernardnormier:
Fixes #6441.

`IceBT::AcceptorI::close()` only unregistered the D-Bus profile, and `DBus::Connection::handleMessage` dispatches outside the lock, so a `NewConnection` invocation already in flight can still reach `AcceptorI::newConnection` after the acceptor is closed. Once the thread-pool work item queued during `IncomingConnectionFactory` shutdown has cleared the ready callback, the late `newConnection` called `ready()` with no callback installed: `assert(_readyCallback)` in debug builds, a call through a null `shared_ptr` in release builds.

This PR adds a `_closed` flag to `AcceptorI`, mirroring the `TransceiverI` guard from #6387:

- `newConnection` now checks `_closed` under `_mutex`; when the acceptor is closed, it closes the delivered fd and returns instead of queuing a transceiver and calling `ready()`. The ordering is sound because `close()` runs inline from `setState(StateClosed)` under the factory mutex (on Linux, `Selector::finish` returns `closeNow` unchanged), while the ready callback is cleared afterwards by a work item whose `finished()` must first acquire that same factory mutex: a `newConnection` that acquires the acceptor's `_mutex` before `close()` completes its `ready()` call while the callback is still installed, and one that acquires it after sees `_closed`.
- `close()` also drains the queued transceivers and closes them. A transceiver queued but never accepted was previously destroyed with a live fd — `assert(_fd == INVALID_SOCKET)` in `~StreamSocket` in debug builds, a leaked fd in release builds. Acceptor-side transceivers have no Bluetooth `Connection` object, so closing them just closes the socket and cannot block.

Unlike #6387, there is no residual `Ice.Executor` window here: the acceptor close is never executor-deferred, so the `_closed` guard covers the whole shutdown path. The startup-side window (a `NewConnection` dispatched between `listen()` and ready-callback installation) is tracked separately in #6550.

### DIFF
--- a/changelog.d/cpp/6441.md
+++ b/changelog.d/cpp/6441.md
@@ -1,0 +1,2 @@
+- Fixed a crash in the IceBT transport. An incoming Bluetooth connection delivered while its object adapter was
+  being deactivated — or the communicator destroyed — could crash the program.
--- a/cpp/src/IceBT/AcceptorI.cpp
+++ b/cpp/src/IceBT/AcceptorI.cpp
@@ -40,6 +40,22 @@ IceBT::AcceptorI::getNativeInfo()
 void
 IceBT::AcceptorI::close()
 {
+    stack<IceInternal::TransceiverPtr> transceivers;
+    {
+        lock_guard lock(_mutex);
+        _closed = true;
+        _transceivers.swap(transceivers);
+    }
+
+    //
+    // Close queued transceivers that were never accepted, otherwise their sockets would leak.
+    //
+    while (!transceivers.empty())
+    {
+        transceivers.top()->close();
+        transceivers.pop();
+    }
+
     if (!_path.empty())
     {
         try
@@ -154,6 +170,15 @@ void
 IceBT::AcceptorI::newConnection(int fd)
 {
     lock_guard lock(_mutex);
+
+    if (_closed)
+    {
+        //
+        // The acceptor was closed: nothing will ever accept this connection, so close the socket here.
+        //
+        IceInternal::closeSocketNoThrow(fd);
+        return;
+    }
 
     _transceivers.push(
         make_shared<TransceiverI>(_instance, make_shared<StreamSocket>(_instance, fd, _bufSize), nullptr, _uuid));
--- a/cpp/src/IceBT/AcceptorI.cpp
+++ b/cpp/src/IceBT/AcceptorI.cpp
@@ -52,7 +52,13 @@ IceBT::AcceptorI::close()
     //
     while (!transceivers.empty())
     {
-        transceivers.top()->close();
+        try
+        {
+            transceivers.top()->close();
+        }
+        catch (...)
+        {
+        }
         transceivers.pop();
     }
 
--- a/cpp/src/IceBT/AcceptorI.h
+++ b/cpp/src/IceBT/AcceptorI.h
@@ -47,6 +47,7 @@ namespace IceBT
         std::string _path;
 
         std::mutex _mutex;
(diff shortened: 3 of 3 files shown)


externl (approved):
Looks good.

pepone (approved):
Two non-blocking points.

- The startup end of the same window is still open: `createAcceptor` installs the ready callback at `ConnectionFactory.cpp:1725`, after `listen()` at `:1718` has already published the BlueZ profile, so a `NewConnection` dispatched in between hits the same null `_readyCallback`. Filed as #6550 — it also invalidates the premise in #6538's last bullet that the callback is always set before any use.

- The `Ice.Executor` caveat in the description doesn't apply to the acceptor: `closeAcceptor()` runs inline from `setState(StateClosed)` under `IncomingConnectionFactory::_mutex` (`Selector::finish` returns `closeNow` unchanged on Linux), and the work item has to take that same mutex in `finished()` before clearing the callback. Unlike `ConnectionI::finished`, it is never executor-deferred, so `_closed` is airtight here.

bernardnormier:
Both points taken:

- Thanks for filing #6550 — agreed, the startup end of the window is out of reach of this guard, and it does invalidate the always-set-before-use premise in #6538's last bullet.
- You're right about the caveat: `closeAcceptor()` runs inline from `setState(StateClosed)` under the factory mutex, and the work item can only clear the ready callback after taking that same mutex in `finished()`, so the executor never opens a window here. I amended the PR description accordingly — the ordering rationale now describes the inline close, and the `Ice.Executor` caveat paragraph is replaced with a note that, unlike #6387, the guard covers the whole shutdown path.

PR #6539, opened 2026-08-10
Don't defer reads when Serialize is enabled on a single-threaded pool

### LINKED ISSUE #6279: Ice.ThreadPool.Client.Serialize defers the next read on IOCP even with a single-threaded pool
`Ice.ThreadPool.Client.Serialize` (and the server equivalent) is a no-op on POSIX when the pool cannot grow, but not on IOCP, where it costs a connection its read/dispatch overlap at any pool size.

## POSIX

The whole serialize path sits inside the size guard in `ThreadPool::ioCompleted` (`cpp/src/Ice/ThreadPool.cpp:738`):

```cpp
if (_sizeMax > 1)
{
#if !defined(ICE_USE_IOCP)
    --_inUseIO;
    if (!_destroyed)
    {
        if (_serialize && current._handler.get() != _workQueue.get())
        {
            _selector.disable(current._handler.get(), current.operation);
        }
    }
```

and `IOScope::completed()` discards the return value (`cpp/src/Ice/ThreadPool.h:215`). `SizeMax` defaults to `Size`, which defaults to 1, so a default-configured pool ignores the setting entirely.

## IOCP

`ioCompleted` ends with (`cpp/src/Ice/ThreadPool.cpp:802`):

```cpp
return _serialize && current._handler.get() != _workQueue.get();
```

which is **outside** the `_sizeMax > 1` guard, and here the return value is used (`cpp/src/Ice/ThreadPool.h:255-265`):

```cpp
// Call finishMessage once IO is completed only if serialization is not enabled.
// Otherwise, finishMessage will be called when the event handler is done with
// the message (it will be called from ~ThreadPoolMessage below).
if (_message._current.ioCompleted())
{
    _finish = false;
    _message._finish = true;
}
```

So `finishMessage` moves from the end of the IO scope to `~ThreadPoolMessage`, after the dispatch upcall. IOCP's `finishMessage` (`cpp/src/Ice/ThreadPool.cpp:872-899`) is what re-arms the read:

```cpp
else if (!current._handler->startAsync(current.operation))
```

The next `WSARecv` is therefore not posted until the upcall returns.

## Why the size guard belongs there too

With `_sizeMax == 1` the pool has one thread, so it cannot dispatch two messages concurrently and cannot reorder them — there is nothing for serialization to do. Deferring the read buys no ordering; it only prevents the kernel from filling the receive buffer while the dispatch runs, which on IOCP does not need a thread. Adding `_sizeMax > 1` to the IOCP return would align the two platforms and make the setting genuinely free at the default size.

Found while reviewing #6222, which defaults DataStorm node communicators to `Ice.ThreadPool.Client.Serialize=1`. That default is a no-op on POSIX at the default pool size, but on Windows it currently costs every node the read/dispatch overlap on its outgoing connections.

bernardnormier:
Fixes #6279.

On Windows, `ThreadPool::ioCompleted` returned `_serialize && current._handler.get() != _workQueue.get()` regardless of the pool size, and the IOCP `IOScope` uses this return value to defer `finishMessage` — which posts the next `WSARecv` — until after the dispatch. A single-threaded pool cannot dispatch two messages concurrently, so serialization buys no ordering there; deferring the read only costs the connection its read/dispatch overlap. On POSIX the entire serialize path already sits inside the `_sizeMax > 1` guard (and the return value is discarded), so this change adds the same guard to the return value, aligning the two platforms.

The C# thread pool copied the async-model design and had the same behavior — on all platforms, since C# always uses the async model. Same one-line fix there. Java follows the POSIX selector model and already guards its serialize path with `_sizeMax > 1`, and JS has no thread pool.

This is not a regression: the behavior is identical in 3.7.10 and at the tip of the 3.7 branch, in both C++ and C#.

### DIFF
--- a/changelog.d/cpp/6279.md
+++ b/changelog.d/cpp/6279.md
@@ -1,0 +1,2 @@
+- Improved performance on Windows when thread pool serialization (`Serialize=1`) is enabled on a thread pool with a
+  single thread.
--- a/changelog.d/csharp/6279.md
+++ b/changelog.d/csharp/6279.md
@@ -1,0 +1,2 @@
+- Improved performance when thread pool serialization (`Serialize=1`) is enabled on a thread pool with a single
+  thread.
--- a/cpp/src/Ice/ThreadPool.cpp
+++ b/cpp/src/Ice/ThreadPool.cpp
@@ -799,7 +799,7 @@ IceInternal::ThreadPool::ioCompleted(ThreadPoolCurrent& current)
         }
     }
 
-    return _serialize && current._handler.get() != _workQueue.get();
+    return _sizeMax > 1 && _serialize && current._handler.get() != _workQueue.get();
 }
 
 #if defined(ICE_USE_IOCP)
--- a/csharp/src/Ice/Internal/ThreadPool.cs
+++ b/csharp/src/Ice/Internal/ThreadPool.cs
@@ -550,7 +550,7 @@ public sealed class ThreadPool : System.Threading.Tasks.TaskScheduler
                 }
             }
         }
-        return _serialize;
+        return _sizeMax > 1 && _serialize;
     }
 
     public bool startMessage(ThreadPoolCurrent current)


externl (approved):
Looks good.

pepone (approved):

PR #6541, opened 2026-08-10
Contain a throwing DataStorm decoder during element attachment

pepone:
Refs #6389.

#6522 guarded the two key decodings on the initialization paths. This guards the nine that remain on the element
attachment paths, where the failure is worse than a missing diagnostic: the throw abandons the whole request, so
the elements around the rejected value are left unattached and unacknowledged and the peer never initializes them.
Readers that had nothing to do with the rejected value silently receive nothing.

A key, a key filter and a sample filter criteria all run the application's `Decoder`. The nine sites:

| Site | Decodes | Now |
| --- | --- | --- |
| `TopicI::getElementSpecs` ×2 | a peer key, a peer key filter | skip that announcement entry, keep matching the rest |
| `TopicI::attachElements` ×2 | an inline key, a peer key filter | skip that spec, keep attaching and acking the rest |
| `TopicI::attachElementsAck` ×2 | an inline key, a peer key filter | skip that spec, keep the `initCb`s and `removedIds` |
| `DataElementI::attach` ×2 | a sample filter criteria | leave that element unattached, ack nothing for it |
| `SubscriberSessionI::s` | an inline key | discard that sample |

`SessionI.cpp:225` is left alone — that one is #6507.

## Why per-item skip rather than all-or-nothing

`getElementSpecs` is pure, but the other five `TopicI` sites mutate as they iterate: each spec's
`DataElementI::attach` runs `attachKey`, which inserts into `_listeners`, populates `subscriber->keyIds`/`keys` and
queues the `onConnectedElements` callbacks. None of that unwinds. Returning an empty result would therefore leave
this node believing it is attached to peer elements the peer never hears about — a half-open attach that survives
until the session drops. In `attachElementsAck` the same throw also discards the `initCb`s accumulated so far (they
run only after the loop) and the `removedIds` that drive `detachElementsAsync`.

A partial spec set is indistinguishable on the wire from one where no local element matched, so skipping is not
observable to the peer as an error — it is the same shape as a normal partial match.

## Why a throw is not a null return

`getElementSpecs`, `attachElements` and `attachElementsAck` already fall back to `alwaysMatchFilter` when a key
filter factory returns **null**. That is fail-open for "no factory registered under that name". A throw is
different — the codec exists and failed — and failing open there would attach elements the peer's filter meant to
exclude, delivering samples it never subscribed to. The guards skip instead, with a comment at each site since it
reads as inconsistent with the line above it.

## Logging

Each site logs a `Warning` naming the topic or element, matching #6522 and the `attachTags` guard from #6343. The
session protocol is fire-and-forget, so a local warning is the only way to report the failure — the oneway dispatch
means no exception reaches the peer, and before this the only trace was a generic `Ice.Warn.Dispatch` naming
neither the topic nor the element:

```
-! warning: failed to dispatch attachElementsAck to p/1 over 127.0.0.1:50926<->127.0.0.1:12020:
   undecodable key
```

`SubscriberSessionI::s` is on the live path, so a persistently throwing decoder warns once per sample. That is
already the precedent one frame down the same call path — `DataReaderI::queue`'s `matchKey` guard warns per sample
too.

## Tests

New `cpp/test/DataStorm/decoderFailures` suite. `Decoder<TestKey>` throws on one value, and the type doubles as a
key filter and sample filter criteria. Four scenarios, each **verified red on `origin/main` and green with this
change** — every one hangs before the fix, because the reader it is asserting on never attaches:

1. **any-key reader, multi-key writer** — the writer answers with one spec per key, each carrying its key inline;
   the spec for the good key must still attach → `TopicI::attachElements` key decode
2. **any-key writer, keyed readers** — the readers' keys travel to the writer inline in the element specs →
   `TopicI::getElementSpecs` key decode
3. **key-filtered reader** whose criteria the writer cannot decode → `TopicI::getElementSpecs` filter decode
4. **sample-filtered reader** whose criteria the writer cannot decode → `DataElementI::attach`

Confirmed by the warnings the run emits, those four scenarios drive four of the nine guards. **Not covered by a
test**: the `attachElements` filter decode, both `attachElementsAck` decodes, and `SubscriberSessionI::s`. The
first three need a topology I could not construct without contorting the suite; the `s()` one is not observable —
every same-name topic in the loop rejects the same sample, so the sample is dropped with or without the guard and
only the warning differs.

No changelog fragment: a throwing `Decoder` is an application bug, so a working application never observes this,
same disposition as #6522.

Verified on macOS: `make -C cpp srcs` and `tests` clean, all 20 `cpp/DataStorm` suites pass, and the new suite ran
6× with no flake. clang-format-19 and ruff clean.

### DIFF
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -116,8 +116,8 @@ DataElementI::attach(
             // The sample filter factory runs the application's decoder. Leave this element unattached rather than
             // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
             // Returning without an ack confines the failure to this element: the caller attaches the other elements
-            // of the request and acks them, so the peer still initializes them. Print the element id rather than the
-            // element, whose keys stream through the application's formatter — application code inside this catch.
+            // of the request and acks them, so the peer still initializes them. The warning prints the element id:
+            // streaming the element itself would run the application's key formatter inside this very catch.
             Warning out(_traceLevels->logger);
             out << "did not attach 'e" << _id << "' to a peer element: its sample filter '" << info->name
                 << "' could not be decoded:\n"
@@ -186,8 +186,8 @@ DataElementI::attach(
             // The sample filter factory runs the application's decoder. Leave this element unattached rather than
             // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
             // Returning no initialization callback confines the failure to this element: the caller attaches and
-            // initializes the other elements of the acknowledgement. Print the element id rather than the element,
-            // whose keys stream through the application's formatter — application code inside this catch.
+            // initializes the other elements of the acknowledgement. The warning prints the element id: streaming
+            // the element itself would run the application's key formatter inside this very catch.
             Warning out(_traceLevels->logger);
             out << "did not attach 'e" << _id << "' to a peer element: its sample filter '" << info->name
                 << "' could not be decoded:\n"
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -107,7 +107,22 @@ DataElementI::attach(
     shared_ptr<Filter> sampleFilter;
     if (auto info = data.config->sampleFilter)
     {
-        sampleFilter = _parent->getSampleFilterFactories()->decode(getCommunicator(), info->name, info->criteria);
+        try
+        {
+            sampleFilter = _parent->getSampleFilterFactories()->decode(getCommunicator(), info->name, info->criteria);
+        }
+        catch (const std::exception& ex)
+        {
+            // The sample filter factory runs the application's decoder. Leave this element unattached rather than
+            // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
+            // Returning without an ack confines the failure to this element: the caller attaches the other elements
+            // of the request and acks them, so the peer still initializes them.
+            Warning out(_traceLevels->logger);
+            out << "did not attach '" << this << "' to a peer element: its sample filter '" << info->name
+                << "' could not be decoded:\n"
+                << ex.what();
+            return;
+        }
     }
 
     string facet = data.config->facet.value_or(string{});
@@ -161,7 +176,22 @@ DataElementI::attach(
     shared_ptr<Filter> sampleFilter;
     if (auto info = data.config->sampleFilter)
     {
-        sampleFilter = _parent->getSampleFilterFactories()->decode(getCommunicator(), info->name, info->criteria);
+        try
+        {
+            sampleFilter = _parent->getSampleFilterFactories()->decode(getCommunicator(), info->name, info->criteria);
+        }
+        catch (const std::exception& ex)
+        {
+            // The sample filter factory runs the application's decoder. Leave this element unattached rather than
+            // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
+            // Returning no initialization callback confines the failure to this element: the caller attaches and
+            // initializes the other elements of the acknowledgement.
+            Warning out(_traceLevels->logger);
+            out << "did not attach '" << this << "' to a peer element: its sample filter '" << info->name
+                << "' could not be decoded:\n"
+                << ex.what();
+            return nullptr;
+        }
     }
 
     string facet = data.config->facet.value_or(string{});
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -116,9 +116,10 @@ DataElementI::attach(
             // The sample filter factory runs the application's decoder. Leave this element unattached rather than
             // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
             // Returning without an ack confines the failure to this element: the caller attaches the other elements
-            // of the request and acks them, so the peer still initializes them.
+            // of the request and acks them, so the peer still initializes them. Print the element id rather than the
+            // element, whose keys stream through the application's formatter — application code inside this catch.
             Warning out(_traceLevels->logger);
-            out << "did not attach '" << this << "' to a peer element: its sample filter '" << info->name
+            out << "did not attach 'e" << _id << "' to a peer element: its sample filter '" << info->name
                 << "' could not be decoded:\n"
                 << ex.what();
             return;
@@ -185,9 +186,10 @@ DataElementI::attach(
             // The sample filter factory runs the application's decoder. Leave this element unattached rather than
             // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
             // Returning no initialization callback confines the failure to this element: the caller attaches and
-            // initializes the other elements of the acknowledgement.
+            // initializes the other elements of the acknowledgement. Print the element id rather than the element,
+            // whose keys stream through the application's formatter — application code inside this catch.
             Warning out(_traceLevels->logger);
-            out << "did not attach '" << this << "' to a peer element: its sample filter '" << info->name
+            out << "did not attach 'e" << _id << "' to a peer element: its sample filter '" << info->name
                 << "' could not be decoded:\n"
                 << ex.what();
             return nullptr;
--- a/cpp/src/DataStorm/DataElementI.cpp
+++ b/cpp/src/DataStorm/DataElementI.cpp
@@ -115,9 +115,8 @@ DataElementI::attach(
         {
             // The sample filter factory runs the application's decoder. Leave this element unattached rather than
             // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
-            // Returning without an ack confines the failure to this element: the caller attaches the other elements
-            // of the request and acks them, so the peer still initializes them. The warning prints the element id:
-            // streaming the element itself would run the application's key formatter inside this very catch.
+            // The warning prints the element id: streaming the element itself would run the application's key
+            // formatter inside this very catch.
             Warning out(_traceLevels->logger);
             out << "did not attach 'e" << _id << "' to a peer element: its sample filter '" << info->name
                 << "' could not be decoded:\n"
@@ -185,9 +184,8 @@ DataElementI::attach(
         {
             // The sample filter factory runs the application's decoder. Leave this element unattached rather than
             // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
-            // Returning no initialization callback confines the failure to this element: the caller attaches and
-            // initializes the other elements of the acknowledgement. The warning prints the element id: streaming
-            // the element itself would run the application's key formatter inside this very catch.
+            // The warning prints the element id: streaming the element itself would run the application's key
+            // formatter inside this very catch.
             Warning out(_traceLevels->logger);
             out << "did not attach 'e" << _id << "' to a peer element: its sample filter '" << info->name
                 << "' could not be decoded:\n"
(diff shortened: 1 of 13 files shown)


cpp/src/DataStorm/TopicI.cpp
```diff
@@ -473,7 +536,24 @@ TopicI::attachElementsAck(
...
+                            // The key filter factory runs the application's decoder. Skip this spec and keep
+                            // processing the remaining ones: the elements attached by the earlier specs have their
+                            // initialization callbacks queued below, and abandoning the loop would drop those along
+                            // with the ids collected in removedIds. Falling back to alwaysMatchFilter the way a null
+                            // return does would attach elements the peer's filter meant to exclude.
+                            Warning out(_traceLevels->logger);
+                            out << "skipped the acknowledged elements of key filter '" << spec.name << "' on topic '"
+                                << this << "': the filter could not be decoded:\n"
+                                << ex.what();
+                            continue;
```
pepone: 🤖 **Claude:** This `continue` also skips the `for (const auto& data : spec.elements)` loop below, so the ids the peer acked for elements this node no longer holds never reach `removedIds` and no `detachElements` is sent — the peer stays attached to elements that are gone. Leaving `filter` null and testing `filter && filter->match(key)` skips the attach without skipping the bookkeeping.

cpp/src/DataStorm/TopicI.cpp
```diff
@@ -541,7 +621,22 @@ TopicI::attachElementsAck(
...
+                    {
+                        // The key factory runs the application's decoder. Skip this spec and keep processing the
+                        // remaining ones: the elements attached by the earlier specs have their initialization
+                        // callbacks queued below, and abandoning the loop would drop those along with the ids
+                        // collected in removedIds.
+                        Warning out(_traceLevels->logger);
+                        out << "skipped the acknowledged elements announced under a key on topic '" << this
+                            << "': the key could not be decoded:\n"
+                            << ex.what();
+                        continue;
```
pepone: 🤖 **Claude:** Same here — `key` stays null, so `else if (key && filter->match(key))` keeps this spec's `removedIds` collection instead of skipping it.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -107,7 +107,22 @@ DataElementI::attach(
...
+            sampleFilter = _parent->getSampleFilterFactories()->decode(getCommunicator(), info->name, info->criteria);
+        }
+        catch (const std::exception& ex)
+        {
+            // The sample filter factory runs the application's decoder. Leave this element unattached rather than
+            // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
+            // Returning without an ack confines the failure to this element: the caller attaches the other elements
+            // of the request and acks them, so the peer still initializes them.
+            Warning out(_traceLevels->logger);
+            out << "did not attach '" << this << "' to a peer element: its sample filter '" << info->name
```
externl: `this` streams the element's keys through the application's formatter (`toString()` walks `_keys` with the app's key formatting) — app code inside the very catch that contains app-code failures, in both attach overloads. Printing `'e' << _id` like the SessionI warning does avoids the rethrow.

externl (approved):

cpp/src/DataStorm/TopicI.cpp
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Done in 6219a294 — the catches now leave the filter/key null instead of skipping the spec, so the loop still records the ids of the elements this node no longer holds and `detachElements` is sent for them. The undecodable elements themselves are deliberately not added to `removedIds`: making the peer detach its half on a local decode failure is a peer-visible partial-ack semantics change, deferred to the 3.9.0 ack-continuation work noted in #6594.

pepone: 🤖 **Claude:** Done in 6219a294 — `filter` stays null and the attach condition tests `filter && filter->match(key)`; the null-return `alwaysMatchFilter` fallback moved inside the try so the throw path cannot fail open.

cpp/src/DataStorm/TopicI.cpp
pepone: 🤖 **Claude:** Done in 6219a294 — `key` stays null and the attach condition tests `key && filter->match(key)`, so this spec's `removedIds` collection runs.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -161,7 +176,22 @@ DataElementI::attach(
...
+        {
+            // The sample filter factory runs the application's decoder. Leave this element unattached rather than
+            // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
+            // Returning no initialization callback confines the failure to this element: the caller attaches and
+            // initializes the other elements of the acknowledgement.
+            Warning out(_traceLevels->logger);
+            out << "did not attach '" << this << "' to a peer element: its sample filter '" << info->name
+                << "' could not be decoded:\n"
+                << ex.what();
+            return nullptr;
```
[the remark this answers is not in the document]
pepone: 🤖 **Claude:** Declining the explicit failure propagation: signalling the decode failure so the peer detaches its half changes what the peer observes on an application codec bug — the partial-ack continuation semantics #6324/#6594 deferred to 3.9.0. What this PR guarantees is local and consistent with the TopicI guards after 6219a294: the failed element attaches nothing and acks nothing, and with #6594 it also commits nothing. The half-open peer residual is documented in #6594 ("Not addressed here") and #6596.

cpp/src/DataStorm/DataElementI.cpp
pepone: 🤖 **Claude:** Fixed in 71318b4f — both warnings now print `'e' << _id` instead of streaming the element, with a comment stating why.

cpp/src/DataStorm/TopicI.cpp
```diff
@@ -164,7 +165,22 @@ TopicI::getElementSpecs(int64_t topicId, const ElementInfoSeq& infos, const shar
...
+            shared_ptr<Key> key;
+            try
+            {
+                key = _keyFactory->decode(_instance->getCommunicator(), info.value);
+            }
+            catch (const std::exception& ex)
+            {
+                // The key factory runs the application's decoder. Skip the peer key it can't decode and keep matching
+                // the remaining ones: the elements the peer announced under the other keys still attach, and a spec
+                // set missing this key is indistinguishable on the wire from one where no local element matched it.
```
bernardnormier: Nothing here is called a "spec set" — this loop fills the `ElementSpecSeq` returned to the peer.
```suggestion
                // the remaining ones: the elements the peer announced under the other keys still attach, and the
                // returned specs are indistinguishable on the wire from ones where no local element matched this key.
```

cpp/src/DataStorm/TopicI.cpp
```diff
@@ -221,7 +237,23 @@ TopicI::getElementSpecs(int64_t topicId, const ElementInfoSeq& infos, const shar
...
             {
-                peerFilter = _keyFilterFactories->decode(_instance->getCommunicator(), info.name, info.value);
+                try
+                {
+                    peerFilter = _keyFilterFactories->decode(_instance->getCommunicator(), info.name, info.value);
+                }
+                catch (const std::exception& ex)
+                {
+                    // The key filter factory runs the application's decoder. Skip the peer filter it can't decode
+                    // rather than falling back to alwaysMatchFilter the way a null return does above: a null return
```
bernardnormier: The `alwaysMatchFilter` fallback this refers to is below the catch, not above.
```suggestion
                    // rather than falling back to alwaysMatchFilter the way a null return does below: a null return
```

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -107,7 +107,23 @@ DataElementI::attach(
...
+        {
+            sampleFilter = _parent->getSampleFilterFactories()->decode(getCommunicator(), info->name, info->criteria);
+        }
+        catch (const std::exception& ex)
+        {
+            // The sample filter factory runs the application's decoder. Leave this element unattached rather than
+            // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
+            // Returning without an ack confines the failure to this element: the caller attaches the other elements
+            // of the request and acks them, so the peer still initializes them. Print the element id rather than the
+            // element, whose keys stream through the application's formatter — application code inside this catch.
```
bernardnormier: The last sentence is phrased as an instruction about the statement below it. Stating the constraint directly:
```suggestion
            // of the request and acks them, so the peer still initializes them. The warning prints the element id:
            // streaming the element itself would run the application's key formatter inside this very catch.
```

bernardnormier (commented):
The guard logic all checks out for me — the null-filter/null-key handling in `attachElementsAck` keeps the `removedIds` bookkeeping while attaching nothing, and the throw-vs-null-return distinction is right. A few comments on the comments, plus one optional note:

These comment blocks run 4–5 lines each, and about half of each is justification of the change ("Returning without an ack confines the failure to this element: the caller attaches the other elements of the request and acks them...") rather than a constraint on the code — that half is PR-description material. The sentences that earn their place in-source are the ones that block a bad future refactor: the "a throw is not a null return, don't fall back to alwaysMatchFilter" distinction, and "Skipping the whole spec would drop that bookkeeping along with the attachments" in the ack guards. Trimming each comment to those would read better; take it or leave it.

cpp/src/DataStorm/DataElementI.cpp
```diff
@@ -161,7 +177,23 @@ DataElementI::attach(
...
+        {
+            sampleFilter = _parent->getSampleFilterFactories()->decode(getCommunicator(), info->name, info->criteria);
+        }
+        catch (const std::exception& ex)
+        {
+            // The sample filter factory runs the application's decoder. Leave this element unattached rather than
+            // attaching it without the filter, which would send the peer the samples its filter meant to exclude.
+            // Returning no initialization callback confines the failure to this element: the caller attaches and
+            // initializes the other elements of the acknowledgement. Print the element id rather than the element,
+            // whose keys stream through the application's formatter — application code inside this catch.
```
bernardnormier: Same here.
```suggestion
            // initializes the other elements of the acknowledgement. The warning prints the element id: streaming
            // the element itself would run the application's key formatter inside this very catch.
```

cpp/src/DataStorm/TopicI.cpp
pepone: 🤖 **Claude:** Applied in f4747c9142.

cpp/src/DataStorm/TopicI.cpp
pepone: 🤖 **Claude:** Applied in f4747c9142.

cpp/src/DataStorm/DataElementI.cpp
pepone: 🤖 **Claude:** Applied in f4747c9142.

cpp/src/DataStorm/DataElementI.cpp
pepone: 🤖 **Claude:** Applied in f4747c9142.

pepone:
🤖 **Claude:** Took the optional trim in 827f550036 — the nine guard comments are cut to the sentences that block a refactor: the throw-vs-null `alwaysMatchFilter` distinction, the `removedIds` bookkeeping in the ack guards, and the element-id-not-element warning constraint. The confines-the-failure justification halves are gone; their substance was already in the PR description.

bernardnormier (approved):
Looks good.

PR #6542, opened 2026-08-10
Use Sphinx cross-reference roles consistently in Python docstrings

### LINKED ISSUE #6342: Python docstrings don't use Sphinx cross-reference roles consistently
The hand-written Python docstrings are inconsistent in how they refer to other methods of the API:

- A few files use proper Sphinx roles, which render as links: `:meth:` (Communicator.py, Locator.py, LoggerAdmin.py, RemoteLogger.py, Process.py, Router.py), `:func:` (ObjectAdapter.py, ServantLocator.py), `:class:`.
- Most references elsewhere are double-backtick inline literals (e.g. ````ice_collocationOptimized```` in ObjectPrx.py), which render as code text, not links.
- "See Also" sections use bare names. The docs build uses sphinx.ext.napoleon, which—unlike numpydoc—does not auto-link bare names in "See Also" sections, so these render as plain text.

We should pick one convention (presumably the Sphinx roles, since only they produce working cross-references with our napoleon-based build) and apply it across the Python docstrings in one pass.

Noted while reviewing the ObjectPrx.py doc updates in https://github.com/zeroc-ice/ice/pull/6337.

bernardnormier:
This PR makes the hand-written Python docstrings use Sphinx cross-reference roles consistently, so that references to other API members render as links in the generated API reference.

- References now use fully-qualified roles (`:meth:`Ice.Class.method``, `:class:`Ice.Class``), which is what resolves with our autodoc setup. Same-class self-references use the `~` form (`:meth:`~Ice.Communicator.shutdown``) so the rendered text stays short. `:func:` is reserved for module-level functions such as `Ice.loadSlice`.
- Double-backtick literals that name API methods (e.g. ``ice_collocationOptimized`` in `ObjectPrx.py`) are converted to roles. Property names, environment variables, and parameter names remain literals.
- Unqualified `:class:`Awaitable``/`:class:`Callable`` (which resolved to nothing) now point to `collections.abc` and link to the Python documentation via intersphinx.
- Also fixed the `AlreadyRegisteredException.kindOfObject` summary, which napoleon misparsed as a type declaration because it ended with a colon.
- The IcePy C docstrings render on the same `Ice.*` pages, so they get the same treatment: qualified roles in `Connection.cpp` and `Init.cpp`, and a rewording of the `WSConnectionInfo.headers` getter docstring, which napoleon misparsed the same way as `kindOfObject`.

The generated files (from slice2py) already emit qualified roles and are unchanged.

Verified with a nitpicky Sphinx build: no new unresolved-reference warnings, and the converted references render as links.

Fixes #6342.

### DIFF
--- a/python/modules/IcePy/Connection.cpp
+++ b/python/modules/IcePy/Connection.cpp
@@ -100,7 +100,7 @@ Associates an object adapter with this connection.
 
 When a connection receives a request, it dispatches this request using its associated object adapter.
 If the associated object adapter is ``None``, the connection rejects any incoming request with an
-:class:`ObjectNotExistException`.
+:class:`Ice.ObjectNotExistException`.
 
 The default object adapter of an incoming connection is the object adapter that created this connection;
 the default object adapter of an outgoing connection is the communicator's default object adapter.
@@ -229,8 +229,8 @@ sndSize : int
     constexpr const char* connectionThrowException_doc = R"(throwException() -> None
 
 Raises an exception that provides the reason for the closure of this connection. For example,
-this function raises :class:`CloseConnectionException` when the connection was closed gracefully by the peer;
-it raises :class:`ConnectionAbortedException` when the connection is aborted with :func:`abort`.
+this function raises :class:`Ice.CloseConnectionException` when the connection was closed gracefully by the peer;
+it raises :class:`Ice.ConnectionAbortedException` when the connection is aborted with :meth:`~Ice.Connection.abort`.
 This function does nothing if the connection is not yet closing or closed.)";
 }
 
--- a/python/modules/IcePy/Connection.cpp
+++ b/python/modules/IcePy/Connection.cpp
@@ -228,9 +228,9 @@ sndSize : int
 
     constexpr const char* connectionThrowException_doc = R"(throwException() -> None
 
-Raises an exception that provides the reason for the closure of this connection. For example,
-this function raises :class:`Ice.CloseConnectionException` when the connection was closed gracefully by the peer;
-it raises :class:`Ice.ConnectionAbortedException` when the connection is aborted with :meth:`~Ice.Connection.abort`.
+Raises an exception that provides the reason for the closure of this connection. For example, this function
+raises :class:`Ice.CloseConnectionException` when the connection was closed gracefully by the peer; it raises
+:class:`Ice.ConnectionAbortedException` when the connection is aborted with :meth:`~Ice.Connection.abort`.
 This function does nothing if the connection is not yet closing or closed.)";
 }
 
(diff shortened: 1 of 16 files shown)


pepone (approved):
Verified the rendered output: 45 → 3 unlinked cross-references, no new Sphinx warnings. Suggestions below, all non-blocking.

**1. `Communicator.py:692` drops a working link.** `:class:`IcePy.NativePropertiesAdmin`` currently resolves to `Ice.NativePropertiesAdmin.html`; as a double-backtick literal it renders as plain text. Suggest the public name (`Ice.NativePropertiesAdmin is IcePy.NativePropertiesAdmin`), 119 cols:

```
        ``Properties`` facet is returned as an :class:`Ice.NativePropertiesAdmin`, which is not an :class:`Ice.Object`.
```

**2. `LocalExceptions.py:358` — "such as" turns an exhaustive list into an open one.** `cpp/include/Ice/LocalExceptions.h:578-579` documents these seven as the complete set; "one of" keeps it closed and still avoids the colon napoleon misparsed.

**3. Consider `~` on the `Ice.`-qualified roles.** 38 of them are same-class self-references, so `Communicator.destroy` now reads "This function calls `Ice.Communicator.shutdown()` implicitly" where it read "calls `shutdown()`" — those already rendered as links before this PR. `:meth:`~Ice.Communicator.shutdown`` keeps the qualified target and restores the short text; I built it both ways and the warning count and resolved-link set are identical. You already use this form for `~collections.abc.Awaitable`.

**4. `Communicator.py:331-333` — re-wrap, "Calling this function" is orphaned on its own line.**

**5. The IcePy C docstrings render on the `Ice.*` pages and are untouched.** `Connection.cpp:103,232-233` still use unqualified roles, and `ConnectionInfo.cpp:254` has the same colon misparse you fixed for `kindOfObject` — `Ice.WSConnectionInfo.headers` renders a bogus `Type: The HTTP headers from the WebSocket upgrade handshake` field. Fix here or file a follow-up.

Also worth dropping the "napoleon does not auto-link bare names" bullet from the description: the bare `ice_collocationOptimized` in that See Also section already rendered as a link.

externl (approved):
Looks good.

bernardnormier:
All five addressed in 48042ba978 — thanks for building it both ways, that made the `~` call easy:

- **NativePropertiesAdmin**: restored as a link with your suggested line, using the public `Ice.NativePropertiesAdmin` name.
- **kindOfObject**: "such as" → "one of".
- **`~` self-references**: applied to the same-class sites — 38, matching your count. Cross-class references keep the full text.
- **createObjectAdapterWithEndpoints**: rewrapped.
- **IcePy C docstrings**: fixed here rather than a follow-up, since they render on the same pages. The three exception roles in `Connection.cpp` are now `Ice.`-qualified, `:func:`abort`` became `:meth:`~Ice.Connection.abort``, and the `headers` getter lost its second colon. Also qualified three `:func:`loadSlice`` roles in `Init.cpp` for consistency.

Description updated as well: the bare-names bullet is gone and the IcePy docstring work is now mentioned.

PR #6545, opened 2026-08-10
Fix RPM builds that define nameprefix

### LINKED ISSUE #6528: RPM: builds that define `nameprefix` fail in `%prep`
The RPM spec supports rebranding the packages with a custom prefix, applied consistently to every `%package -n`, `Requires:` and scriptlet as `%{?nameprefix}`. But a build that actually defines it fails in `%prep`, before anything is compiled:

```shell
rpmbuild -bb --define "nameprefix zeroc-" ice.spec
```

`packaging/rpm/ice.spec:328` unpacks with:

```
%setup -q -n %{name}-%{archive_tag}
```

`Name:` is `%{?nameprefix}ice`, so `%setup` looks for a `zeroc-ice-main` directory, while the GitHub source archive extracts to `ice-main`. The prefix has no bearing on the tarball layout, so `%{name}` is the wrong thing to derive the directory from:

```
%setup -q -n ice-%{archive_tag}
```

`%{name}` is used in only one other place, `Source0`'s `#/%{name}-%{version}.tar.gz`, where it just names the downloaded file and is harmless.

ZeroC's own packages are built without a prefix (`packaging/rpm/build-package.sh` never defines it), so the published RPMs are unaffected — this only breaks customers rebuilding the spec under their own name. Present on `main` and on `3.8` (`packaging/rpm/ice.spec:355`), since #3537.

bernardnormier:
The `%setup` directive derived the source directory name from the prefixed package name (`%{name}-%{archive_tag}`), while the source archive always extracts to an `ice-` directory, so any build defining `nameprefix` failed in `%prep`.

Verified on AlmaLinux 9: with the original spec, `rpmbuild -bp --define "nameprefix zeroc-"` fails in `%prep`; with this fix, `%prep` succeeds both with and without `nameprefix` defined.

Fixes #6528

### DIFF
--- a/changelog.d/packaging/6528.md
+++ b/changelog.d/packaging/6528.md
@@ -1,0 +1,1 @@
+- Fixed RPM builds that rebrand the packages with the `nameprefix` macro: such builds previously failed during `%prep`.
--- a/packaging/rpm/ice.spec
+++ b/packaging/rpm/ice.spec
@@ -325,7 +325,7 @@ network programming interfaces and allows you to focus your efforts on
 your application logic.
 
 %prep
-%setup -q -n %{name}-%{archive_tag}
+%setup -q -n ice-%{archive_tag}
 
 %build
 #


pepone (approved):
Tested on AlmaLinux 9 with and without `nameprefix` — everything works as expected.

PR #6546, opened 2026-08-10
Don't use the OpenSSL BIO's fd after the delegate transceiver lost its fd

### LINKED ISSUE #6393: OpenSSL TransceiverI can write to a stale descriptor after a buffer-size helper closes the delegate's fd
`OpenSSL::TransceiverI::initialize` snapshots the delegate's descriptor number into a socket BIO (`BIO_new_socket(fd, 0)`, OpenSSLTransceiverI.cpp:77) and installs it into `_ssl`. When a buffer-size helper fails on the live fd (`getInfo`/`setBufferSize`), it closes the fd, and after #6388 the delegate correctly forgets it (`NativeInfo::clearFd()`) — but nothing invalidates the copy held by the BIO.

As a result, after such a failure on an ssl/wss connection:

- `close()` unconditionally calls `SSL_shutdown` (OpenSSLTransceiverI.cpp:282), which sends close_notify through the BIO on the stale number.
- A send on the still-Active connection reaches `SSL_write` (OpenSSLTransceiverI.cpp:332) the same way.

If another thread was assigned the same descriptor number in the meantime, these write TLS bytes into an unrelated descriptor — the same hazard class as #6376, one layer above the transceiver-level fix in #6388 (and present before it too; #6388 removes the double close but can't reach the BIO).

SecureTransport and Schannel are unaffected: both route I/O through the delegate, which re-reads its `_fd`. The trigger has the same rarity as #6376: `getsockopt`/`setsockopt` must fail on a live socket. A fix needs the fd invalidation to reach the SSL layer — e.g. skip `SSL_shutdown` and fail `SSL_write`/`SSL_read` when the delegate's fd is `INVALID_SOCKET`.

Noted while reviewing #6388.

bernardnormier:
`OpenSSL::TransceiverI::initialize` snapshots the delegate's fd into a socket BIO installed in the SSL connection object. When a buffer-size helper fails and closes the fd, the delegate forgets it (`NativeInfo::clearFd()`, #6388), but nothing invalidates the copy held by the BIO — so a later `close()` still sends close_notify through the stale number, and `write()`/`read()` on a still-Active connection reach `SSL_write`/`SSL_read` the same way, potentially hitting an unrelated descriptor that reused the number.

This change closes the window in two layers:

- Entry guards: `close()` skips the shutdown notification and `write()`/`read()` throw `ConnectionLostException` when the delegate's fd is `INVALID_SOCKET`.
- BIO invalidation at the point of loss: `getInfo()` and `setBufferSize()` — the two operations whose delegate calls can clear the fd — invalidate the BIO's cached fd (`BIO_set_fd(..., INVALID_SOCKET, BIO_NOCLOSE)`) when that happens. This covers I/O issued from inside OpenSSL itself, where entry guards cannot help — in particular handshake records written by `SSL_accept`/`SSL_connect` after a `verifyCallback → getInfo()` failure — and every other `getInfo` caller. Subsequent BIO I/O fails with `EBADF` and no retry flag.

`getInfo()` also now duplicates the peer certificate only after the delegate call succeeds, so the `X509_dup` no longer leaks when that call throws.

Verified on Ubuntu 24.04: libIce builds cleanly and the IceSSL test suite passes.

No changelog entry: the trigger requires `getsockopt`/`setsockopt` to fail on a live socket, same as #6388 which also shipped without one.

Fixes #6393

### DIFF
--- a/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
+++ b/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
@@ -539,15 +539,24 @@ OpenSSL::TransceiverI::getInfo(bool incoming, string adapterName, string connect
     // adapterName is the name of the object adapter currently associated with this connection, while _adapterName
     // represents the name of the object adapter that created this connection (incoming only).
 
+    Ice::ConnectionInfoPtr delegateInfo;
+    try
+    {
+        delegateInfo = _delegate->getInfo(incoming, std::move(adapterName), std::move(connectionId));
+    }
+    catch (...)
+    {
+        invalidateBIOFd();
+        throw;
+    }
+
     X509* peerCertificate = nullptr;
     if (_peerCertificate)
     {
         peerCertificate = X509_dup(_peerCertificate);
     }
 
-    return make_shared<ConnectionInfo>(
-        _delegate->getInfo(incoming, std::move(adapterName), std::move(connectionId)),
-        peerCertificate);
+    return make_shared<ConnectionInfo>(std::move(delegateInfo), peerCertificate);
 }
 
 void
@@ -558,7 +567,27 @@ OpenSSL::TransceiverI::checkSendSize(const IceInternal::Buffer&)
 void
 OpenSSL::TransceiverI::setBufferSize(int rcvSize, int sndSize)
 {
-    _delegate->setBufferSize(rcvSize, sndSize);
+    try
+    {
+        _delegate->setBufferSize(rcvSize, sndSize);
+    }
+    catch (...)
+    {
+        invalidateBIOFd();
+        throw;
+    }
+}
+
+void
+OpenSSL::TransceiverI::invalidateBIOFd() const
+{
+    if (_ssl && _delegate->getNativeInfo()->fd() == INVALID_SOCKET)
+    {
+        // The socket BIO installed in _ssl caches the fd number. The delegate no longer holds this fd, so invalidate
+        // the BIO's copy: later TLS I/O — including handshake records written from inside SSL_accept/SSL_connect —
+        // fails with EBADF instead of touching a reused descriptor.
+        BIO_set_fd(SSL_get_wbio(_ssl), INVALID_SOCKET, BIO_NOCLOSE);
+    }
 }
 
 int
--- a/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
+++ b/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
@@ -165,7 +165,9 @@ OpenSSL::TransceiverI::initialize(IceInternal::Buffer& readBuffer, IceInternal::
                         break;
                     }
 
-                    if (IceInternal::connectionLost() || IceInternal::getSocketErrno() == 0)
+                    // A handshake I/O failure after the delegate lost its fd is a connection loss.
+                    if (_delegate->getNativeInfo()->fd() == INVALID_SOCKET || IceInternal::connectionLost() ||
+                        IceInternal::getSocketErrno() == 0)
                     {
                         throw ConnectionLostException(__FILE__, __LINE__, IceInternal::getSocketErrno());
                     }
--- a/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
+++ b/cpp/src/Ice/SSL/OpenSSLTransceiverI.cpp
@@ -279,14 +279,19 @@ OpenSSL::TransceiverI::close()
 {
     if (_ssl)
     {
-        int err = SSL_shutdown(_ssl);
-
-        //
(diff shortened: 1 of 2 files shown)


pepone (commented):
The three guards are correct, but the handshake path is still open: `verifyCallback` builds its `ConnectionInfo` argument with `getInfo()` (OpenSSLTransceiverI.cpp:583), which reaches `TcpTransceiver::getInfo` → `getRecvBufferSize`, then swallows the exception and returns 1 so OpenSSL finishes the flight. On an outgoing connection `SSL_connect` writes the client Certificate/Finished records through the BIO's stale fd right after the server certificate is verified, so #6393's hazard survives — and an entry guard before `SSL_accept`/`SSL_connect` wouldn't help, since the fd is cleared inside the call.

Invalidate the BIO's fd where the delegate loses it instead: `BIO_set_fd(SSL_get_wbio(_ssl), INVALID_SOCKET, BIO_NOCLOSE)` in this class's `getInfo()`/`setBufferSize()` overrides covers the handshake and every other caller in one place — writes then fail with `EBADF` and no retry flag. Keep the three entry guards for the clearer `ConnectionLostException`.

bernardnormier:
Good catch on the handshake path — fixed in 86374ee565 with your approach: `getInfo()` and `setBufferSize()` wrap their delegate calls and, when the delegate has lost its fd, invalidate the BIO's copy with `BIO_set_fd(SSL_get_wbio(_ssl), INVALID_SOCKET, BIO_NOCLOSE)` before rethrowing. That runs inside `getInfo()` itself, so it covers the `verifyCallback` flight, `initialize()`'s own `verifyPeer(getInfo(...))` call, and the admin/metrics callers in one place. The three entry guards stay as you suggested.

This also supersedes Copilot's inline comment, which described the same hazard — except its "propagate the stored exception from the callback" part: `verifyCallback` still returns 1 by design, since with the BIO invalidated the handshake write fails with `EBADF` and the hazard is gone; the stored verification exception still surfaces through the existing rethrow in `initialize()` when the handshake gets that far.

While touching `getInfo()`, the `X509_dup` now happens after the delegate call, so the duplicated certificate no longer leaks when that call throws.

Re-verified in the ubuntu:24.04 container: libIce builds and the IceSSL configuration suite passes.

pepone (approved):

externl (commented):
A few things from an adversarial pass:

- `initialize()` is only transitively protected: safe because every current clearFd site routes through the two wrapped calls, but a future clearFd caller outside them reintroduces the stale fd inside `SSL_accept`/`SSL_connect` while read/write/close stay guarded. One `invalidateBIOFd()` at the top of the handshake loop makes it self-protecting.
- Same lost-fd condition surfaces as `ConnectionLostException` post-handshake but plain `SocketException` mid-handshake: `EBADF` isn't in `connectionLost()`, so `initialize()`'s `SSL_ERROR_SYSCALL` branch takes the generic path. Since the guards deliberately picked `ConnectionLostException`, that branch could throw it too when the delegate fd is `INVALID_SOCKET`.
- PR-text nit: on Linux the trigger can't fire on a live socket (`getsockopt` fails only `EBADF`/`ENOTSOCK`, `setsockopt` clamps); the reachable trigger is BSD `ENOBUFS` via `Connection::setBufferSize`. Worth framing as hardening rather than a live-Linux bug.
- Layering note for a code comment: the entry guards are the removable half and `invalidateBIOFd` the load-bearing one — the fd can be cleared *inside* a guarded `SSL_read`/`SSL_write` via renegotiation's verify callback, so the guards alone are unsafe.

externl (approved):
Looks good — see the comments in my previous review.

bernardnormier:
Took the exception-type point: `initialize()`'s `SSL_ERROR_SYSCALL` branch now throws `ConnectionLostException` when the delegate's fd is `INVALID_SOCKET`, matching the entry guards.

On the trigger framing: the OpenSSL transport only builds where neither Schannel nor SecureTransport apply (the `#else` in `cpp/include/Ice/SSL/Config.h`), which among supported platforms means Linux - a BSD `ENOBUFS` path isn't reachable in any supported configuration. Agreed the live-socket trigger isn't reachable on Linux either; that's why this ships without a changelog entry.

On the loop-top `invalidateBIOFd()`: it wouldn't close a real hole - the fd is cleared *inside* `SSL_accept`/`SSL_connect` (verify callback → `getInfo()`), and that same call keeps writing; only the invalidation inside `getInfo()`/`setBufferSize()` covers it, as your layering note says. It would only guard hypothetical future `clearFd` callers between loop iterations, so I left it out.

On the layering comment: the comment on `invalidateBIOFd` already records the load-bearing fact - it covers I/O issued from inside OpenSSL, where entry guards can't help. I'd rather not add commentary about which half is removable.

Issue #6115, opened 2026-07-23
Python low-severity batch: Future callback timing, wall-clock waits, ctor arg combination, `adapterDealloc` GIL-held join, `waitForDeactivate` missing try/catch


pepone:
> [!NOTE]
> **AI-assisted finding — needs human triage.** Surfaced by an AI static-review audit and verified by re-reading the source on `main` (`1579b6928e`). It was not benchmarked, reproduced at runtime, or exploited, so the severity is a static-review estimate. Please confirm before acting.

## Summary

Five low-severity Python findings.

### 1. Future callback timing inconsistency
`python/python/Ice/Future.py:110-114`: a callback added *after* completion runs directly (`fn(self)`), so its exception propagates to the `add_done_callback` caller; a callback added *before* completion runs through `_callCallbacks`, which catches and logs. Same in `InvocationFuture.py:95-99` (`add_sent_callback`). CPython's own `Future.add_done_callback` wraps even the immediate call. 

### 2. Timed waits use the wall clock
`python/python/Ice/Future.py:253-260` uses `time.time()` for `_wait`, so a backward/forward wall-clock step lengthens/shortens the wait. Use `time.monotonic()`.

### 3. Constructor args documented mutually exclusive can be combined
`python/python/Ice/Communicator.py:91-103`: passing both `args` and `initData` silently replaces the user's `initData` with an args-derived one (discarding its properties/logger/executor/sliceLoader), while the event-loop adapter is still taken from the discarded object; passing both `eventLoop` and `initData` silently ignores `eventLoop`.

### 4. `adapterDealloc` joins `std::async` futures while holding the GIL (native)
`python/modules/IcePy/ObjectAdapter.cpp:281,285` `delete`s the hold/deactivate futures (joining the task) with no `AllowThreads`; a still-blocked `waitForHold` task can hang the interpreter at adapter GC. `communicatorDealloc` gets the ordering right.

### 5. `waitForDeactivate` non-main path lacks try/catch (native)
`python/modules/IcePy/ObjectAdapter.cpp:500-504` calls `waitForDeactivate()` with no try/catch, unlike its `waitForHold` twin (`:410-421`); a C++ throw would escape an `extern "C"` frame → `std::terminate`.

## Fix

Wrap the immediate done/sent callback in try/except-log; use `time.monotonic()`; validate the mutually-exclusive ctor args; release the GIL around the future joins; add the missing try/catch.

**Severity:** Low.

pepone:
[...]

PR #6547, opened 2026-08-11
Catch and log exceptions raised by immediately-invoked future callbacks

pepone:
Part of #6115 (finding 1).

`Ice.Future.add_done_callback` and `Ice.InvocationFuture.add_sent_callback` treat a callback differently depending on
when it is registered:

- registered while the future is still running, the callback later runs through `Future._callCallbacks` /
  `InvocationFuture.set_sent`, which catch and log any exception it raises;
- registered after the future completed (or after the request was sent), the callback runs immediately as `fn(self)` /
  `fn(self._sentSynchronously)` and its exception propagates to whoever called `add_done_callback` /
  `add_sent_callback`.

Which of the two paths a callback takes is a race in any real application, so an application can't rely on either
behavior. CPython has the same immediate-call path in `concurrent.futures.Future.add_done_callback` and wraps it in
`try/except Exception` + `LOGGER.exception`.

This PR routes the immediate call through the same catch-and-log helper as the deferred one. The sent-callback loop
moved into a new `InvocationFuture._callSentCallbacks` so both paths share it.

## Testing

`python/test/Ice/ami/AllTests.py` gets a "testing callback exception handling" block: a done callback registered before
completion, a done callback registered after completion, and a sent callback registered after the request was sent all
raise; the test checks that nothing propagates and that the three exceptions reach the `Ice.Future` logger. Against the
unfixed code, the second one escapes `add_done_callback` and fails the test.

### DIFF
--- a/changelog.d/python/future-callback-exception.md
+++ b/changelog.d/python/future-callback-exception.md
@@ -1,3 +1,3 @@
 - Fixed `Ice.Future.add_done_callback` and `Ice.InvocationFuture.add_sent_callback` in Ice for Python: when the future
-  was already completed - or the request already sent - an exception raised by the callback propagated to the caller
-  instead of being caught and logged like an exception raised by a callback registered earlier.
+  was already completed - or the request already sent - an exception raised by the callback propagated to the caller.
+  Such an exception is now caught and logged with Python's `logging` module.
--- a/changelog.d/python/future-callback-exception.md
+++ b/changelog.d/python/future-callback-exception.md
@@ -1,0 +1,3 @@
+- Fixed `Ice.Future.add_done_callback` and `Ice.InvocationFuture.add_sent_callback` in Ice for Python: when the future
+  was already completed - or the request already sent - an exception raised by the callback propagated to the caller
+  instead of being caught and logged like an exception raised by a callback registered earlier.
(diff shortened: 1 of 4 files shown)


python/python/Ice/InvocationFuture.py
```diff
@@ -151,11 +153,14 @@ def set_sent(self, sentSynchronously: bool):
...
+        self._callSentCallbacks(callbacks, sentSynchronously)
+
+    def _callSentCallbacks(self, callbacks: list[Callable[[bool], None]], sentSynchronously: bool):
         for callback in callbacks:
             try:
                 callback(sentSynchronously)
-            except Exception as ex:
-                logging.getLogger("Ice.Future").exception("sent callback raised exception: %s", ex)
+            except Exception:
+                logging.getLogger("Ice.Future").exception("sent callback raised exception")
```
pepone, line 163: This is consistent with what we do in `Future.py` and it includes the exception type, message, and traceback.

externl (approved):
Looks good.

changelog.d/python/future-callback-exception.md
```diff
@@ -0,0 +1,3 @@
+- Fixed `Ice.Future.add_done_callback` and `Ice.InvocationFuture.add_sent_callback` in Ice for Python: when the future
+  was already completed - or the request already sent - an exception raised by the callback propagated to the caller
+  instead of being caught and logged like an exception raised by a callback registered earlier.
```
bernardnormier:
```suggestion
- Fixed `Ice.Future.add_done_callback` and `Ice.InvocationFuture.add_sent_callback` in Ice for Python: when the future
  was already completed - or the request already sent - an exception raised by the callback propagated to the caller.
  Such an exception is now caught and logged with Python's `logging` module.
```

bernardnormier (approved):

PR #6549, opened 2026-08-11
Reject mutually exclusive Ice.Communicator initialization arguments

pepone:
Part of #6115 (finding 3).

The `Ice.Communicator` constructor and `Ice.initialize` document — through their overloads and their docstrings — that
`initData` is mutually exclusive with `args` and `eventLoop`. Combining them was accepted and silently dropped one of
the arguments:

- `Communicator(args, initData=initData)` took the event loop adapter from `initData`, then replaced `initData` itself
  with `InitializationData(properties=Properties(args))` — discarding the caller's properties, logger, executor, Slice
  loader and so on.
- `Communicator(eventLoop=eventLoop, initData=initData)` ignored `eventLoop`.
- `Ice.initialize(args, initData=initData)` ignored `args` and `eventLoop` outright.

Both now raise `ValueError`, which is what the Ice Python mapping raises for a bad argument elsewhere (`Ice.Proxy`,
`IcePy.Properties`, `IcePy.ObjectAdapter`, ...).

## Testing

`python/test/Ice/properties/Client.py` — a server-less client test — checks the three combinations. No in-tree caller
passes both.

### DIFF
--- a/changelog.d/python/communicator-mutually-exclusive-args.md
+++ b/changelog.d/python/communicator-mutually-exclusive-args.md
@@ -1,0 +1,4 @@
+- The `Ice.Communicator` constructor and `Ice.initialize` in Ice for Python now raise `ValueError` when the `initData`
+  argument is combined with the `args` or `eventLoop` argument. These arguments are mutually exclusive; previously the
+  extra argument was silently ignored, except for `Ice.Communicator(args, initData=initData)`, which replaced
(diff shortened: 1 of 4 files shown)


pepone:
Closing — wrong direction. The Python mapping relies on type checking for this: pyright already rejects all four combinations, which is why every call in the new test needed `# type: ignore`. See #6553 for the details.

